PR #874, opened 2016-09-09
Web config

thomasfangyi:
Create a web server for configuring `Parameters`


docs/source/configuration.rst
```diff
@@ -0,0 +1,73 @@
...
+Once the user has successfully configured the parameters through the web
+interface and saved the configuration file, the configuration needs to be
+loaded to Firedrake when using firedrake. To load a parameter file to a
+Parameters instance, use
+:meth:`~firedrake.parameters.Parameters.import_params` and pass the filename as
+the parameter. For example
+
+.. code-block:: python
+
+    parameters.import_params("parameters.json")
```
wence-: Use `Parameters.load` and `Parameters.save` (also mention the latter here).

docs/source/parameters.rst
```diff
@@ -0,0 +1,128 @@
...
+parameter accordingly with first inserting into the dict. This is done by
+:func:`~firedrake.parameters.KeyType.get_type`.
+
+However, if the user wishes to add a custom-configured TypedKey, the following
+syntax could be used at the time of first insertion.
+
+.. code-block:: python
+
+    parameters[TypedKey("foo", type=IntType(), help="A sample key",
+    visible_level=1, depends="baz")] = "bar"
```
wence-: Indentation.

docs/source/parameters.rst
```diff
@@ -0,0 +1,128 @@
...
+
+Note that if the user wishes to use pre-defined ``Keytypes``, it is necessary
```
wence-: Link to ``KeyTypes`` with ```:class:`Keytypes <~.KeyType>````

firedrake/web_config.py
```diff
@@ -0,0 +1,123 @@
...
+    import importlib
+    parser = argparse.ArgumentParser(
+        description="Run a web server to configure the module")
+    parser.add_argument("module_name", help="The name of the module, it must \
+                        contain an attribute of type `firedrake.parameters.Parameters` \
+                        of name `parameters`")
+    args = parser.parse_args()
+    module_name = args.module_name
+    module = importlib.import_module(module_name)
+    web_config(module.parameters)
```
wence-: This file should go.

firedrake/gui_config.py
```diff
@@ -0,0 +1,408 @@
...
+    button_save = Button(mainframe, text="Save to File", command=save_json)
+    button_save.grid(column=3, row=row_count, sticky=S)
+    button_quit = Button(mainframe, text="Save and Quit",
+                         command=save_and_quit)
+    button_quit.grid(column=4, row=row_count, sticky=E)
+
+    mainframe.bind('<Configure>', configure_frame)
+    canvas.bind('<Configure>', configure_canvas)
+
+    root.mainloop()
```
wence-: Pull out generic functionality, and then drop the Tk interface.

firedrake/parameters.py
```diff
@@ -4,29 +4,428 @@
...
+        """Specify whether a key is dependent on another key, the depended key
+            must be in the same Parameters class and must be of BoolType"""
+        if hasattr(self, "_depends"):
+            return self._depends
+        else:
+            return None
+
+    @depends.setter
+    def depends(self, key):
+        self._depends = key
```
wence-: Move type stuff to (say) `firedrake/config_types.py` (better names appreciated).

firedrake/parameters.py
```diff
@@ -51,6 +450,124 @@ def set_update_function(self, callable):
...
+        """Should PyOP2 print the size of caches at program exit?"""
+    parameters["pyop2_options"].get_key("print_summary").help = \
+        """Should PyOP2 print a summary of timings at program exit?"""
+    parameters["pyop2_options"].get_key("matnest").help = \
+        """Should matrices on mixed maps be built as nests? (Default yes)"""
+    # Other
+
 
 parameters = Parameters()
 """A nested dictionary of parameters used by Firedrake"""
```
wence-: Below, dogfood configuration types and use `TypedKey` when building assembly cache options.

requirements-ext.txt
```diff
@@ -8,3 +8,4 @@ pytest-xdist
 pytest-benchmark
 ipython
 matplotlib
+docrepr
```
wence-: Need `flask` too?

docs/source/configuration.rst
wence-: Can we add a section on configuring the look of the web page (branding)?

docs/source/configuration.rst
```diff
@@ -65,9 +65,36 @@ Once the user has successfully configured the parameters through the web
...
+
+To change the header and footer, there are two options.
+
+Option 1: Edit the ``header.html`` and ``footer.html`` in the ``templates``
+folder.
+
+After installation, the templates files for the web interface will be placed
+into the ``templates`` folder. To change the header and footer, simply replace
+``header.html`` and ``footer.html`` with your custom page with the same file
+name.
```
wence-: This shouldn't be an option :).

stephankramer:
Some feedback after trying to apply this for the configuration of Thetis. This was discussed previously in the Thetis dev. meeting:

- the FloatType is very permissive allowing any expression `x` for which `float(x)` can be evaluated. This on itself is fine, but then I would expect some sanitizing:
```
>>> parameters['assembly_cache']['max_factor'] = True
>>> print parameters['assembly_cache']['max_factor']
True
```
whereas BoolType is not permissive at all:
```
>>> parameters['reorder_meshes'] = 1
  File "<stdin>", line 1, in <module>
  File "/home/skramer/firedrake/src/firedrake/firedrake/parameters.py", line 148, in __setitem__
    self.__setitem__(self.get_key(key), value)
  File "/home/skramer/firedrake/src/firedrake/firedrake/parameters.py", line 145, in __setitem__
    + str(value))
ValueError: Invalid value for key reorder_meshes:1
```
but does allow the strings `"True"` or `"False"` which are again stored without coercion:
```
>>> parameters['reordering_meshes'] = "False"
>>> if parameters['reordering_meshes']:
...   print "Reordering anyway"
...
"Reordering anyway"
```
- related, the role of the parse() method on the KeyType (FloatType,StrType, etc.) objects is a little unclear: currently it seems to only be called to parse strings from a .json file but then in most of the implementations it seems to rely on validate() being directly able to validate that same string
- it seems the KeyType objects may acquire state that depends on the value they have been fed. If multiple keys share the same type this may give unexpected results. Firstly, a _last_error attribute is set if validate() returns False - it could be argued this is safe as _last_error should only be checked directly after a validate() call. Then there's lots going on in the OrType with _curr_type that I can't make much sense of (in particular in the parse() function). What is the use case of setting curr_type actually?
- for types that are instantiated with additional parameters - bounds for any NumericType, options for StrType and types for OrType - it would be useful if the `__str__` method would provide that information:
```
>>> parameters[TypedKey('foo', val_type=StrType("one", "two", "three"))] = "one"
>>> print parameters.get_key('foo').type
str
```

Some more wishlist items regarding the parameters dictionary:
- it would be good if the parameters dictionary were frozen after instantiation. This is to avoid parameters being ignored if their key is misspelled. Also in a typical use case for Firedrake "models", option names may frequently change as the model code changes, but options are typically set in separate "setup" files that may be out-of-date.

A way to implement this would be to follow the `FrozenClass` model:
```
class Parameters(object):
   _isfrozen = False
   def __init__(self):
   ...
   def __setitem__(self, key, value):
      if self._isfrozen and key is not in self.keys():
          raise KeyError('...')
```
I think a good model for options dictionaries in general would be to subclass from `Parameters` and set all of its expected keys with some default values:
```
class FiredrakeParameters(Parameters):
   def __init__(self, *args, **kwargs):
       super(FiredrakeParameters, self).__init__(self, *args, **kwargs)
       self[TypedKey("reorder_meshes", val_type=BoolType(), help="""...""")] = True
       ...
       _isfrozen = True

parameters = FiredrakeParameters()
```
Note that in general options dictionaries are not global variables (unlike Dolfin/Firedrake's `parameters`). So instantiating more than one of them can make sense.
It was discussed already, that some parts of the framework (in particular dolfin-adjoint) may be inserting new keys "at runtime" - this could be solved by simply not freezing the relevant sub-dictionaries.

- More controversially (as this is something we want for the Thetis options dictionary and I'm not sure this should be offered at the Firedrake level), we would like to access options via attributes, i.e.:
```
    parameter.key = value
    print parameters.key
```
This could be easily implemented by overloading the `__setattr__` and `__getattr__` methods. There is the possibility for name clashes between existing attributes but this can be worked around. Note that `__getattr__(attr)` only gets called when `attr` does not exist as a class or object attribute. For `__setattr__` we would have to check whether there is an associated attribute already:
```
    def __getattr__(self, attr):
         return self[attr]
    def __setattr__(self, attr, value):
          if attr in self.keys():
            self[attr] = value
          elif not self._isfrozen or hasattr(self, attr):
             super(Parameters, self).__setattr__(self, attr, value)
          else:
             raise AttributeError('...')
   def __setitem__(self, key, value):
         # we have to prevent keys being inserted that already exist as attributes
         # as this would block access to them:
         if attr is not in self.keys() and hasattr(self, key):
            raise KeyError('Key name {} cannot be used in parameters dictionary'.format(key))
         ...
```
Note that this means ```parameters.key = value``` cannot be used to insert new keys, only to change the value of existing ones. If `key` does not yet exist and the dictionary is not yet "frozen" it will always add `key` as a new attribute. Once it is frozen an AttributeError is raised, so setting the _isfrozen flag blocks adding both new keys and attributes.

wence-:
Unfortunately, I think no-one has a pressing need for this.  So we'll close for now, maybe to be resurrected in the future.

PR #881, opened 2016-09-19
Test nonlinear stokes

wence-:


wence-:
Needs firedrakeproject/tsfc#57.

miklos1:
Slightly odd though that Travis didn't pick this up.

PR #846, opened 2016-08-02
recursive prolongation and injection

alsgregory:
alteration:

- can now prolong and inject up more than one level of a hierarchy.
- offers support for 'generalised (or 'skipping') hierarchies' (see slack/general and firedrake-mlmc/generalised_hierarchy.py) by prolonging and injecting the full hierarchy instead of just the refined hierarchy.

PR #882, opened 2016-09-19
Better solvers

wence-:
Makes GMG more magic (no more need for NLVSHierarchy).  Also tidies up solver parameters setting, fixing a thetis issue.  Allows GMG inside fieldsplits (previously not supported) by symbolically splitting and coarsening/refining the application context.

@alsgregory This deprecates more of the hierarchy interface: you only need to explicitly build a `MeshHierarchy` and then just build functions/functionspaces on the mesh you want.  `inject`/`prolong`/`restrict` just take two `Function`s (need not be from a `FunctionHierarchy`).  As such, only meshes have level information attached.

@rckirby This doesn't compose right with matrix-free yet, because we'd need to think harder about how the solver application context gets transferred (or attached to) implicit matrices.  But I think it should not be too hard, if we can then implement a fast `MatGetDiagonal` on implicit matrices (should be possible by doing some symbolic/kernel manipulation), then this is a good step in the direction of matrix-free MG.

I have added a demo demonstrating some functionality.

@knepley this may be interesting for teaching purposes (if your students can't wrap their brains round PetscFE)

Currently needs the (referenced in .travis.yml) petsc4py branch.


firedrake/mg/interface.py
```diff
@@ -1,106 +1,132 @@
...
+        assert len(functions) == len(fs_hierarchy)
+        for f, V in zip(functions, fs_hierarchy):
+            assert f.function_space() == V
+        return tuple(functions)
+    else:
+        return tuple([firedrake.Function(f) for f in fs_hierarchy])
+
+
+def FunctionSpaceHierarchy(mesh_hierarchy, *args, **kwargs):
+    return tuple(firedrake.FunctionSpace(mesh, *args, **kwargs) for mesh in mesh_hierarchy)
```
alsgregory: Could we have some warnings in here to remind user of the depreciation, at least until I've had a chance to remove uses of FSH in some of my firedrake-mlmc stuff?

firedrake/solving_utils.py
```diff
@@ -104,91 +145,128 @@ def __init__(self, problems, mat_type, pmat_type, appctx=None):
...
         if matfree or pmatfree:
-            # We will want the newton state for some preconditioners
-            for c, x in zip(appctxs, self._xs):
-                c["state"] = x
-
+            # A split context will already get the full state.
+            # TODO, a better way of doing this.
+            # Now we don't have a temporary state inside the snes
+            # context we could just require the user to pass in the
+            # full state on the outside (or maybe just the velocity field).
```
wence-: Good point, will remove.

firedrake/mg/interface.py
wence-: Yep.

demos/multigrid/geometric_multigrid.py.rst
```diff
@@ -0,0 +1,248 @@
...
+
+  a = (inner(grad(u), grad(v)) - p * div(v) + div(u) * q)*dx
+
+  L = inner(Constant((0, 0)), v) * dx
+
+  x, y = SpatialCoordinate(mesh)
+
+  t = conditional(y < 0.5, y - Constant(1.0/4.0), y - Constant(3.0/4.0))
+  l = Constant(1.0/6.0)
+  gbar = conditional(Or(And(Constant(1.0/4.0) - l/2 < y,
```
miklos1: I don't quite see the point of wrapping all these numbers with `Constant` in this example, especially since they are inlined in the expression so you won't be able to change their values later. (Reminder: these are functions constant in space values which are passed in to assembly kernels rather than constant values hardcoded in the kernel.)

firedrake/functionspace.py
```diff
@@ -225,7 +225,8 @@ def MixedFunctionSpace(spaces, name=None, mesh=None):
 
         def rec(eles):
             for ele in eles:
-                if ele.num_sub_elements() > 0:
+                if ele.num_sub_elements() > 0 \
```
miklos1: There was some related change in UFL I had to make when merging our UFL upstream. IIRC `TensorProductElement` now also stores its subelements this way.

rckirby:
"Stuff the solver context into the Python context of an implicit matrix"
may be too naive to hope it would work for GMG?

For getting the diagonal, there are a few options:
1.) A_{ii} = <A e_i, e_i>, so you could in theory extract the diagonal
entries via actions.  You can minimize the number of actions by coloring
the basis functions appropriately and extracting multiple diagonal entries
concurrently.  I'm not sure what the chromatic number of the right graph
would be, but it's likely tolerable in 2d and Bad in 3d (plus the cost of
coloring)
2.) UFL/TSFC/PyOP2 magic -- build a kernel that just loops over the
diagonal and assembles into a petsc vector.  That will be more efficient at
run-time and might not take much more development time than coloring.
3.) As a stop-gap, we could assemble the matrix to extract its diagonal,
and then throw the aij matrix away.  That would make things "work" but be
deeply against the spirit of matrix-free.

wence-:
I think this is now good to go.  Please comment.

firedrake/solving_utils.py
dorugeber: poke

wence-:
Now hopefully incorporating changes from #846 as well.

wence-:
Travis is timing out in the mac build (which is why this is failing).  We could dial back the tests it does...

alsgregory:
@wence- See comment on error occurring with skipping hierarchies #846

PR #885, opened 2016-09-21
Safety check for user PYTHONPATH

dham:
Having a `PYTHONPATH` variable pointing at all sorts of extraneous things is a frequent cause of user pain. This makes having `PYTHONPATH` set a fatal error at install or upgrade time. The user can override this with the `--honour-pythonpath` option, and this override will be remembered for this installation.

### DIFF
--- a/.drone.yml
+++ b/.drone.yml
@@ -45,7 +45,7 @@ build:
 
     # Install firedrake
     - cd $FIREDRAKE_HOME
-    - $DRONE_BUILD_DIR/scripts/firedrake-install --disable-ssh --minimal-petsc --slepc --adjoint --slope --package-branch firedrake $DRONE_COMMIT
+    - $DRONE_BUILD_DIR/scripts/firedrake-install --disable-ssh --minimal-petsc --slepc --adjoint --slope --honour-pythonpath --package-branch firedrake $DRONE_COMMIT
 
     # Activate the virtualenv
     - . firedrake/bin/activate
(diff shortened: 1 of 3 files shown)


.drone.yml
```diff
@@ -45,7 +45,7 @@ build:
 
     # Install firedrake
     - cd $FIREDRAKE_HOME
-    - $DRONE_BUILD_DIR/scripts/firedrake-install --disable-ssh --minimal-petsc --slepc --adjoint --slope --package-branch firedrake $DRONE_COMMIT
+    - $DRONE_BUILD_DIR/scripts/firedrake-install --disable-ssh --minimal-petsc --slepc --adjoint --slope --honour-pythonpath --package-branch firedrake $DRONE_COMMIT
```
miklos1: Why does Drone need this?

scripts/firedrake-install
```diff
@@ -86,6 +86,8 @@ honoured.""",
                         help="Minimise the set of petsc dependencies installed. This creates faster build times (useful for build testing).")
     parser.add_argument("--honour-petsc-dir", "--honour_petsc_dir", action="store_true",
                         help="Usually it is best to let Firedrake build its own PETSc. If you wish to use another PETSc, set PETSC_DIR and pass this option.")
+    parser.add_argument("--honour-pythonpath", "--honour_pythonpath", action="store_true",
+                        help="Pointing to external Python packages is usually a user error. Set this option if you know that you want PYTHONPATH set.")
```
miklos1: You specify `dest="pythonpath"` below, but not here. How does that work?

dham: Because of the hacky way that we save configuration. For install, it's fine to just set args.honour_pythonpath. For upgrade you have to take into account a possible previous saved value.

miklos1: So `args.pythonpath` won't be undefined if you install without `--honour-pythonpath`?

dham: `args.pythonpath` will be undefined at *install* but it is never accessed. `args.pythonpath` is only accessed on upgrade.

PR #890, opened 2016-09-22
Change custom extrusion example

dorugeber:
The previous snippet was nonsense!  Did it ever work?  If so, it must have been chance that every vertex in the base mesh was a vertex 0 of some cell.  I have no idea what happened at the top of the mesh, since the top cells of the top layer were definitely never accessed.


dorugeber:
custom

For fun, this is how it would have looked before, assuming the numbering worked out right.  From some angles it looks reasonable, from others it's clearly borked.

PR #892, opened 2016-09-23
During testing, switch demo plotting output to pdfs

dham:
Currently, `make test` will repeatedly pop up plots and stall while processing the demos. This patch causes the test harness to switch to a file backend. In fact, because we never call the right methods, no files are written.


miklos1:
Where do the PDF files go?

dham:
There are no pdf files. Switching to the file backend prevents the popups, but because we never call the methods to actually write to file, nothing happens. Effectively this neuters the plot commands.

PR #894, opened 2016-09-24
All new installs in developer mode.

dham:
Remove the developer mode and install all new Firedrake installs in
developer mode. This is because developer mode only has advantages.

This change is backwards compatible in the sense that existing
non-developer mode installs should continue to work as before.

### DIFF
--- a/scripts/firedrake-install
+++ b/scripts/firedrake-install
@@ -100,8 +100,6 @@ honoured.""",
                             formatter_class=RawDescriptionHelpFormatter)
 
     group = parser.add_mutually_exclusive_group()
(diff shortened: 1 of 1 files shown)


scripts/firedrake-install
```diff
@@ -146,6 +144,7 @@ honoured.""",
     if args.package_branch:
         branches = {package: branch for package, branch in args.package_branch}
 
+    args.developer = True # All new installations are developer mode, since there is no advantage to non-developer mode installs.
```
miklos1: Why this is not failing flake8? We aren't ignoring E261.

PR #901, opened 2016-10-05
Custom callbacks

dorugeber:
This adds the ability for the user to call custom functionality before Jacobian and residual assembly in nonlinear problems (e.g. if the user has non-UFL operations to perform).  We're using this in the Monge-Ampère quasi-Newton work, for example.

If we think the `init` is going to grow out of control, we should find a way to rewrite this sooner rather than later.

Travis showed X but all the Linux builds passed.

### DIFF
--- a/firedrake/solving_utils.py
+++ b/firedrake/solving_utils.py
@@ -167,6 +167,10 @@ class _SNESContext(object):
     :arg appctx: Any extra information used in the assembler.  For the
         matrix-free case this will contain the Newton state in
         ``"state"``.
+    :arg pre_jacobian_callback: User-defined function called immediately
+        before Jacobian assembly
+    :arg pre_function_callback: User-defined function called immediately
+        before residual assembly
 
     The idea here is that the SNES holds a shell DM which contains
     this object as "user context".  When the SNES calls back to the
@@ -174,7 +178,7 @@ class _SNESContext(object):
     get the context (which is one of these objects) to find the
     Firedrake level information.
     """
-    def __init__(self, problem, mat_type, pmat_type, appctx=None):
+    def __init__(self, problem, mat_type, pmat_type, appctx=None, pre_jacobian_callback=None, pre_function_callback=None):
         from firedrake.assemble import allocate_matrix, create_assembly_callable
         if pmat_type is None:
             pmat_type = mat_type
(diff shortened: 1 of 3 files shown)


dorugeber:
I'm also suspicious that I had to write (identical) docstrings in two places...

wence-:
In and of itself this is sort of fine.  However, it sort of raises the issue that we don't have a more general way to do this "right".  IOW, if I also want to do something after assembly, I need another keyword argument.  In dolfin-land the way this works is that you can define a subclass of `NonlinearProblem` that implements `f` and `J` methods that take the output tensor and the current state and fill the output tensor with values.  We could try and mimic something like that.  

We'd need to pick apart what lives in the `_SNESContext` and what lives in the `NonlinearVariationalProblem` and then figure out what ought to go where.  The idea would be something like the `snescontext` provides the `form_function` and `form_jacobian` callbacks as before, but rather than directly calling `assemble`, calls back to `NLVP.F/J/Jp` as appropriate.  The default implementation can just do what happens right now (call `assemble`) but the user could override.

firedrake/solving_utils.py
```diff
@@ -130,14 +130,18 @@ class _SNESContext(object):
     :arg appctx: Any extra information used in the assembler.  For the
         matrix-free case this will contain the Newton state in
         ``"state"``.
+    :arg pre_jacobian_callback: A user-defined function that will
+        be called immediately before Jacobian assembly
+    :arg pre_function_callback: A user-defined function that will
+        be called immediately before residual assembly
```
wence-: What arguments do these functions take.  What are they allowed to do?

dorugeber:
Added an improved docstring to `variational_solver.py`.  Re-review plz.

wence-:
Modulo the test issue I think this is probably OK.

dorugeber:
Test added!

tests/regression/test_custom_callbacks.py
```diff
@@ -0,0 +1,47 @@
...
+        else:
+            beta.assign(2.0)
+
+    problem = NonlinearVariationalProblem(F, u)
+
+    solver = NonlinearVariationalSolver(problem,
+                                        pre_jacobian_callback=update_alpha,
+                                        pre_function_callback=update_beta)
+
+    solver.solve()
```
wence-: If you use a single iteration, can you assert that you get u.dat.data == 1.5 (and then solving again you get 2?)

dorugeber:
Added a one-iteration version to test intermediate state.

dorugeber:
Yes, I fully agree.

PR #907, opened 2016-10-10
Move UFL interpolation code to TSFC

miklos1:
Since code extensively uses TSFC internals, it often needs to change when TSFC changes. To avoid having to file two pull requests at each such occasion in the future, this pull request moves this code into TSFC.

Requires firedrakeproject/tsfc#62.


dham:
Yes. Not controversial.

PR #909, opened 2016-10-13
Real function space

dham:
The long-awaited R-space. Most of the heavy lifting has already gone into PyOP2. This mostly just makes the Firedrake wrappers behave correctly. There is also a new manual section explaining how this works.


firedrake/functionspaceimpl.py
```diff
@@ -818,3 +823,64 @@ def ComponentFunctionSpace(parent, component):
...
+        return None
+
+    def top_nodes(self):
+        ":class:`RealFunctionSpace` objects have no bottom nodes."
+        return None
+
+    def exterior_facet_boundary_node_map(self, method):
+        """":class:`RealFunctionSpace` objects have no exterior facet boundary
+        node map."""
+        return None
```
wence-: To allow field splitting to work with real spaces (especially inside bigger problems) we need to enable the `.dm` property correctly.  And also the `._ises` property.  This will need to be special-cased.  So you should implement the `_dm` method that makes this DM.  The way I do it is that PyOP2 supplies the DM and the `field_ises` via the `dof_dset` of the function space.  This is presumably not implemented for `GlobalDataSet` objects.

firedrake/functionspaceimpl.py
```diff
@@ -818,3 +823,64 @@ def ComponentFunctionSpace(parent, component):
...
+        self.rank = 0
+
+        self._node_count = 1
+
+        self.comm = mesh.comm
+
+        """:class:`RealFunctionSpace` objects have no node set."""
+        self.node_set = None
+        """:class:`RealFunctionSpace` objects have no dof set."""
+        self.dof_dset = op2.GlobalDataSet(self.make_dat())
```
wence-: If I make two `RealFunctionSpace` objects, they currently don't share a `dof_dset`.  This may be a problem.

wence-: It's not clear to me, in the pyop2 implementation, why a global dataset needs to remember the global used to make it (and not just the shape, etc...).

firedrake/functionspace.py
```diff
@@ -245,7 +247,7 @@ def rec(eles):
     spaces = tuple(s.topological for s in flatten(spaces))
     # Error checking
     for space in spaces:
-        if type(space) is impl.FunctionSpace:
+        if type(space) in (impl.FunctionSpace, impl.RealFunctionSpace):
```
miklos1: Why, when `RealFunctionSpace` inherits `FunctionSpace`?

.travis.yml
```diff
@@ -56,7 +56,7 @@ install:
   - export CC=mpicc
   - mkdir tmp
   - cd tmp
-  - ../scripts/firedrake-install --disable-ssh --minimal-petsc ${SLEPC} --adjoint --slope --install thetis --install gusto ${PACKAGE_MANAGER}
+  - ../scripts/firedrake-install --disable-ssh --minimal-petsc ${SLEPC} --adjoint --slope  --install thetis --install gusto ${PACKAGE_MANAGER} --package-branch PyOP2 globals_petsc_changes
```
wence-: This branch is already merged, right?

firedrake/functionspaceimpl.py
```diff
@@ -822,3 +827,75 @@ def ComponentFunctionSpace(parent, component):
...
+    This class should not be directly instantiated by users. Instead,
+    FunctionSpace objects will transform themselves into
+    :class:`RealFunctionSpace` objects as appropriate.
+
+    """
+
+    def __init__(self, mesh, element, name):
+
+        self.name = name
+        self._index = None
```
wence-: Unnecessary attribute.

.travis.yml
miklos1: Also this is `.travis.yml`...

firedrake/functionspaceimpl.py
```diff
@@ -822,3 +827,75 @@ def ComponentFunctionSpace(parent, component):
...
+        self._mesh = mesh
+        self.dim = 1
+        self.rank = 0
+
+        self._node_count = 1
+
+        self.comm = mesh.comm
+
+        """:class:`RealFunctionSpace` objects have no node set."""
+        self.node_set = None
```
wence-: Most of these are argument independent, so I would make them class attributes:

```python
class RealFunctionSpace(FunctionSpace):

    finat_element = None
    dim = 1
    rank = 0
    shape = ()
    node_set = None

    def __init__(self, mesh, element, name):
        self._ufl_element = element
        self.name = name
        self.comm = mesh.comm
        self._mesh = mesh
        self.dof_dset = op2.GlobalDataSet(self.make_dat())
```

Note as well we now have `finat_element`, not `fiat_element`.

firedrake/tsfc_interface.py
```diff
@@ -199,6 +202,23 @@ def compile_form(form, name, parameters=None, inverse=False):
     return kernels
 
 
+def _real_mangle(form):
+    """If the form contains arguments in the Real function space, replace these with literal 1 before passing to tsfc."""
+
+    a = form.arguments()
+    reals = map(lambda x: x.ufl_element().family() == "Real", a)
+    if not a or not any(reals):
```
wence-: the second condition is subsumed by the first, since `any(()) => False`.

PR #915, opened 2016-10-19
Fix indexing errors in hedgehog extrusion on quads

wence-:
We never noticed on triangles since for P1 the coordinate dimension
matches the number of dofs.


dorugeber:
Oh... I vaguely recall us having to fix this for radial extrusion some time ago :(

miklos1:
This seems to fix my random failure issue.

Issue #914, opened 2016-10-17
Support baij and aij as separate matrix types


florianwechsung:
Hi, 

I am trying to solve the linear elasticity equations using the GAMG preconditioner. However, while my code works when I use hypre, is doesn't like gamg very much. The following simplified examples using the laplacian illustrate the problem:

```
from firedrake import *
parameters["pyop2_options"]["block_sparse"] = False
mesh = UnitSquareMesh(20, 20)
V = VectorFunctionSpace(mesh, "CG", 1)
u = TrialFunction(V)
v = TestFunction(V)

a = inner(grad(u), grad(v))*dx
L = (v[0]+v[1])*dx

u = Function(V)
solve(a == L, u, solver_parameters={"ksp_type": "cg", "mat_type" : "aij", "pc_type": "gamg"})
```

This fails with PETSc telling me `Require AIJ matrix type`. However, the following two examples work.

When I use hypre instead of gamg:

```
from firedrake import *
parameters["pyop2_options"]["block_sparse"] = False
mesh = UnitSquareMesh(20, 20)
V = VectorFunctionSpace(mesh, "CG", 1)
u = TrialFunction(V)
v = TestFunction(V)

a = inner(grad(u), grad(v))*dx
L = (v[0]+v[1])*dx

u = Function(V)
solve(a == L, u, solver_parameters={"ksp_type": "cg", "mat_type" : "aij", "pc_type": "hypre"})
```

Or when I solve in 1D instead of 2D:

```
from firedrake import *
parameters["pyop2_options"]["block_sparse"] = False
mesh = UnitSquareMesh(20, 20)
V = FunctionSpace(mesh, "CG", 1)
u = TrialFunction(V)
v = TestFunction(V)

a = inner(grad(u), grad(v))*dx
L = v*dx

u = Function(V)
solve(a == L, u, solver_parameters={"ksp_type": "cg", "mat_type" : "aij", "pc_type": "gamg"})
```

I was told on the slack channel that `parameters["pyop2_options"]["block_sparse"] = False` should result in a matrix of type aij for PETSc, but this doesn't seem to happen here.

wence-:
Sorry, I lied to you about the name of the option, it should be `"block_sparsity"`.  Suggestions for how to make this more natural to control from Firedrake welcome.  We currently support:

```
mat_type: {aij, matfree, nest}
```

[...]

PR #918, opened 2016-10-19
Near null space

florianwechsung:
First suggestion for implementation of the near null space.


florianwechsung:
addresses #914

firedrake/linear_solver.py
```diff
@@ -80,8 +82,14 @@ def __init__(self, A, P=None, solver_parameters=None,
             if P is not None:
                 transpose_nullspace._apply(self.P, transpose=True)
 
+        if near_nullspace is not None:
+            near_nullspace._apply(self.A)
+            if P is not None:
+                near_nullspace._apply(self.P)
```
wence-: I think you need to pass `near=True` here.  Otherwise this will be treated as a nullspace, no?

firedrake/linear_solver.py
```diff
@@ -123,6 +131,8 @@ def solve(self, x, b):
             self.nullspace._apply(self._W.dof_dset.field_ises)
         if len(self._W) > 1 and self.transpose_nullspace is not None:
             self.transpose_nullspace._apply(self._W.dof_dset.field_ises, transpose=True)
+        if len(self._W) > 1 and self.near_nullspace is not None:
+            self.near_nullspace._apply(self._W.dof_dset.field_ises)
```
wence-: Similarly here, say `near=True`.

firedrake/nullspace.py
```diff
@@ -257,7 +273,7 @@ def _apply(self, matrix_or_ises, transpose=False):
                 raise RuntimeError("Shape of matrix (%d, %d) does not match size of nullspace %d" %
                                    (rows, cols, len(self)))
             # Hang the expanded nullspace on the big matrix
-            self._apply_monolithic(matrix, transpose=transpose)
+            self._apply_monolithic(matrix, transpose=transpose, near=near)
             return
         ises = matrix_or_ises
```
wence-: In the `near=True` case, you near to do the same thing as currently, composing the split space with the IS.  But you should use the key `"nearnullspace"`.  So do something like:

```python
key = "nearnullspace" if near else "nullspace"
...
   is_.compose(key, ...)
```

tests/regression/test_nullspace.py
```diff
@@ -119,6 +119,43 @@ def test_nullspace_mixed():
...
+    ns = [n0, n1, n2]
+    n_interp = [interpolate(n, V) for n in ns]
+    n_normalized = [interpolate(n*(1.0/sqrt(n.dat.inner(n.dat))), V) for n in n_interp]
+    nsp = VectorSpaceBasis(vecs=n_normalized)
+
+    w = Function(V)
+    solve(lhs(F) == rhs(F), w, bcs=bcs, solver_parameters={'ksp_monitor': True, 'ksp_rtol': 1e-8, 'ksp_atol': 1e-8, 'ksp_type': 'cg', 'pc_type': 'gamg'}, near_nullspace=nsp)
+    w = Function(V)
+    solve(lhs(F) == rhs(F), w, bcs=bcs, solver_parameters={'ksp_monitor': True, 'ksp_rtol': 1e-8, 'ksp_atol': 1e-8, 'ksp_type': 'cg', 'pc_type': 'gamg'})
+    assert sqrt(assemble(inner(w-w_exact, w-w_exact)*dx)) < 1e-7
```
wence-: How do I determine if this worked?  I presume when you pass the near null space the convergence is much better?  Can you assert that, maybe by comparing the convergence against "known good" values.  You can see an example of how to do that at https://github.com/firedrakeproject/firedrake/blob/master/tests/regression/test_matrix_free.py#L59.  Where I test that two ostensibly equivalent solves produce the same convergence history.

florianwechsung:
updated my code with the suggestions above

tests/regression/test_nullspace.py
```diff
@@ -119,6 +119,73 @@ def test_nullspace_mixed():
...
+        wo = f.read()
+
+    with open(w_nns_log, "r") as f:
+        f.readline()
+        w = f.read()
+
+    print "Without near nullspace"
+    print wo
+    print "With near nullspace"
+    print w
```
wence-: Please don't print in the tests.

florianwechsung: May I ask why not? If I understand pytest correctly output of successful tests isn't shown anyway.

wence-: It's sort of a bad code smell.  If you want to test that these values are somehow "right", then do so.  If you just want to see them when the test fails, you can drop into the debugger on failure and inspect the values.

florianwechsung: Allright, took the print statements out!

wence-:
Aha, now the build fails because you changed a private API that dolfin-adjoint uses.  This requires a matching change in dolfin-adjoint here https://bitbucket.org/dolfin-adjoint/dolfin-adjoint/src/e9e1ab44105c3dfa984dac2c52f42e4faf8ea293/dolfin_adjoint/compatibility.py?at=master&fileviewer=file-view-default#compatibility.py-119

florianwechsung:
I opened a pull request in dolfin_adjoint, see [bitbucket](https://bitbucket.org/dolfin-adjoint/dolfin-adjoint/pull-requests/22/added-near_nullspace-option/diff)

wence-:
Thanks.  I think this looks fine, although all this pulling apart of boolean options is a little ugly...

firedrake/linear_solver.py
```diff
@@ -71,17 +73,23 @@ def __init__(self, A, P=None, solver_parameters=None,
         self.ksp.setDMActive(False)
 
         if nullspace is not None:
-            nullspace._apply(self.A)
+            nullspace._apply(self.A, near=False)
             if P is not None:
-                nullspace._apply(self.P)
+                nullspace._apply(self.P, near=False)
```
wence-: No need here (and elsewhere) to provide explicit defaults.

firedrake/nullspace.py
```diff
@@ -269,7 +285,8 @@ def _apply(self, matrix_or_ises, transpose=False):
             # Compose appropriate nullspace with IS for schur complement
             if ises is not None:
                 is_ = ises[i]
-                is_.compose("nullspace", basis.nullspace(comm=self.comm))
+                key = "near_nullspace" if near else "nullspace"
```
wence-: This line could live outside the loop, but meh.

tests/regression/test_nullspace.py
```diff
@@ -119,6 +119,68 @@ def test_nullspace_mixed():
...
+    w_exact.interpolate(Expression(("x[0]*x[1]", "x[0]*x[1]")))
+    # div(sigma(w_exact)) = (mu + lmbda, mu + lmbda)
+    f = Constant((mu + lmbda, mu + lmbda))
+    F = inner(sigma(u), grad(v))*dx + inner(f, v)*dx
+
+    bcs = [DirichletBC(V, w_exact, (1, 2, 3, 4))]
+
+    n0 = Constant((1, 0))
+    n1 = Constant((0, 1))
+    n2 = Expression(("x[1] - 0.5", "-(x[0]-0.5)"))
```
florianwechsung: Updated n2 but had to use 
```
    w_exact = Function(V)
    w_exact.interpolate(as_vector([x*y, x*y]))
```
as I got an exception otherwise.

wence-: Ah, I see, you're using it in the boundary condition too.

PR #919, opened 2016-10-20
Codecov

wence-:
Posts (hopefully) code coverage stats.

### DIFF
--- a/.codecov.yml
+++ b/.codecov.yml
@@ -1,2 +1,8 @@
 comment:
   layout: "header, diff, changes"
+
+coverage:
+  status:
+    project:
+      default:
+        target: 85
--- a/.codecov.yml
+++ b/.codecov.yml
@@ -1,0 +1,2 @@
+comment:
+  layout: "header, diff, changes"
--- a/.coveragerc
+++ b/.coveragerc
@@ -1,0 +1,19 @@
+[run]
+branch = True
+
+[report]
+# Regexes for lines to exclude from consideration
+exclude_lines =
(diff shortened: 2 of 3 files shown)


wence-:
Seems to work...

miklos1:
This is just coverage of the firedrake component by the Firedrake test suite, right? Would it be easy to report how much of other packages are covered by the Firedrake test suite?

miklos1:
Is it line based coverage? Can you switch it to branch based?

wence-:
Easy (we just have to add them to the list of `--cov` commands).  However, the magic source tracking in the GUI won't work I think.

wence-:
I think you can have branch coverage (http://coverage.readthedocs.io/en/coverage-4.0.3/branch.html#branch).  Just need to figure out how to enable it.

wence-:
So the coverage stuff will easily measure the coverage of everything.  However, that doesn't hook up with the GUI.  So you have to look in the build log to see anything useful.  So I'm tempted to remove that part.  But keep the branch based coverage setting.

miklos1:
Yeah, I'm not interested in looking at the build log.

wence-:
OK, so for the online stuff I'll only do coverage of firedrake itself.  I can add a `make coverage` target that does the necessary for local testing.

mairieli:
Hi @wence- ! I've noticed that you changed the configuration of a bot in this PR. I am researching how to develop better GitHub bots. Would you be willing to participate in my research, answering a few questions?

PR #925, opened 2016-11-01
Status command

dham:
Introduce the `firedrake-status` command which reports the current firedrake configuration and the status of all of the component branches.

### DIFF
--- a/scripts/firedrake-status
+++ b/scripts/firedrake-status
@@ -61,7 +61,7 @@ status = {}
 for dir in sorted(os.listdir(firedrake_env + "/src")):
     if dir.startswith("h5py"):
         continue
-    
+
     try:
         os.chdir(firedrake_env + "/src/" + dir)
     except OSError as e:
@@ -83,8 +83,8 @@ for dir in sorted(os.listdir(firedrake_env + "/src")):
             modified = True
 
         status[dir] = {"branch": branch,
-                       "revision" : revision,
-                       "modified" : modified}
+                       "revision": revision,
+                       "modified": modified}
 config["component_status"] = status
 log.info("Firedrake configuration: ")
 log.info(pformat(config))
--- a/scripts/firedrake-status
+++ b/scripts/firedrake-status
@@ -5,6 +5,7 @@ import logging
 import sys
 import os
 import subprocess
+from collections import OrderedDict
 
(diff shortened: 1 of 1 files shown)


scripts/firedrake-status
```diff
@@ -0,0 +1,90 @@
...
+            modified = False
+        except subprocess.CalledProcessError as e:
+            modified = True
+
+        status[dir] = {"branch": branch,
+                       "revision": revision,
+                       "modified": modified}
+config["component_status"] = status
+log.info("Firedrake configuration: ")
+log.info(pformat(config))
```
wence-: Other things we might consider showing here:

- Output of `pip freeze` (shows all packages installed in virtualenv).
- value of `PYTHONPATH` ?
- values of `PETSC_DIR` and `PETSC_ARCH` (in case user used `--honour-petsc-dir`).
- Output of `uname -a` (to get OS information)

Anything else?

dham: `pip freeze` is horribly verbose. Maybe we should only dump that into the log file and not to screen. It's not very useful for the use case of someone wanting to see the state of their own system - but I concede that it's very useful for the debug case.

wence-: OK, so this command is intended to have two uses then?

1. Run `firedrake-status` to see if you're up to date.
2. Run `firedrake-status` to report information about your system when reporting a bug.

If this is so, is it possible to make the screen output more friendly?  Plausibly also say if any given component is not up to date wrt upstream (git fetch in each repo...)?

dham: I don't think it's to see if you're up to date. It's for the "now what branches were everything on" case - and similar "what state is this tree in" issues.

wence-: OK.

scripts/firedrake-status
```diff
@@ -0,0 +1,134 @@
...
+print "Firedrake Configuration:"
+for key, val in config["options"].iteritems():
+    print "    {}: {}".format(key, val)
+
+print "Additions:"
+if config["additions"]:
+    for a in config["additions"]:
+        print "   " + a
+else:
+    print "    None"
```
wence-: When we couldn't import `firedrake_configuration` this prints as:

```
Firedrake Configuration:
Additions:
    None
```

Should we print something like:

```
Firedrake Configuration:
    No install configuration found
...
```

Or is that too confusing?

PR #934, opened 2016-11-04
Forcibly upgrade all installations to developer mode

dham:
This change to firedrake-update detects if an install is not in developer mode and, if this is the case, upgrades it to developer mode. This is achieved by:

a. Setting `args.clean`. This uninstalls all the relevant firedrake components.
b. Setting `options["developer"]`. This causes all the relevant components to install in developer mode and records developer mode in the configuration file.

### DIFF
--- a/scripts/firedrake-install
+++ b/scripts/firedrake-install
@@ -1205,16 +1205,32 @@ else:
 
     # Forcibly upgrade old installations to developer mode.
     if not options["developer"]:
+        forced_developer = True
         args.clean = True
         options["developer"] = True
+    else:
+        forced_developer = False
 
     if args.clean:
-        clean_obsolete_packages()
-        for package in packages:
-            pip_uninstall(package)
-        if args.rebuild:
-            pip_uninstall("petsc")
-            pip_uninstall("petsc4py")
+        try:
+            clean_obsolete_packages()
+            for package in packages:
+                pip_uninstall(package)
+            if args.rebuild:
+                pip_uninstall("petsc")
+                pip_uninstall("petsc4py")
+        except:
+            if forced_developer:
+                log.error("*************************************************************")
+                log.error("* Uninstalling packages failed while attemping to upgrade   *")
+                log.error("* your installation to developer mode.                      *")
+                log.error("*                                                           *")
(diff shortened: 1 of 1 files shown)


scripts/firedrake-install
```diff
@@ -123,7 +123,7 @@ honoured.""",
     parser.add_argument("--honour-pythonpath", "--honour_pythonpath", action="store_true",
                         help="Pointing to external Python packages is usually a user error. Set this option if you know that you want PYTHONPATH set.")
     parser.add_argument("--rebuild-script", "--rebuild_script", action="store_true",
-                        help="Only rebuild the firedrake-install script. Use this option if your firedrake-install script is broken and a fix has been released in upstream Firedrake. You will need to specify any other options which you wish to be honoured by your new update script (eg. --developer or --sudo).")
+                        help="Only rebuild the firedrake-install script. Use this option if your firedrake-install script is broken and a fix has been released in upstream Firedrake. You will need to specify any other options which you wish to be honoured by your new update script (--sudo).")
```
miklos1: Do we still have sudo?

dham: It's currently still an option which we honour. I don't know whether we should...

dham:
@miklos1: I have tested it by forcing an install into non-developer mode and then running this upgrade. It appeared to work. I'm not sure what else I could do.

scripts/firedrake-install
miklos1: I cannot find any place where we actually use this option in the install/update script.

dham: Fine then I'll get rid of it. I assumed that since it was still there someone was using it. It's kind of not really relevant to this PR.

tkarna:
FYI I tried this on a non-developer mode install and got error:
```
subprocess.CalledProcessError: Command '['/work/02454/karnat/sources/python-gcc-O/firedrake/bin/pip', 'uninstall', '-y', 'COFFEE==0.1.0']' returned non-zero exit status 1
```

Ran the same command manually, it worked fine. After that the update was successful and all the repos are now in developer mode.

wence-:
The travis push failure is because I merged the backend-purge in the interim and therefore that commit is not expected to pass (the merge, OTOH, does pass)/

PR #936, opened 2016-11-07
move options_prefix docstring to correct position

dorugeber:
Fixes bork in 6875323


dorugeber:
updated to move code block to end

PR #940, opened 2016-11-09
Shortcut install options

dham:
This makes --install thetis and --install gusto work as discussed in the meeting on Tuesday.


scripts/firedrake-install
```diff
@@ -204,7 +213,9 @@ else:
                         help="Produce a verbose log of the update process in firedrake-update.log. If you have problem running this script, please include this log in any bug report you file.")
     parser.add_argument("--verbose", "-v", action="store_true", help="Produce more verbose debugging output.")
     parser.add_argument("--install", action="append", dest="packages", default=config["options"].get("packages", []),
-                        help="Additional packages to be installed. The address should be in format vcs+protocol://repo_url/#egg=pkg&subdirectory=pkg_dir")
+                        help="Additional packages to be installed. The address should be in format vcs+protocol://repo_url/#egg=pkg&subdirectory=pkg_dir. Some additional packages have shortcut install options for more information see --install_help.")
```
wence-: Does one need a newline somewhere in here so that it's not too long?

scripts/firedrake-install
```diff
@@ -136,7 +143,9 @@ honoured.""",
                         type=str, action="store",
                         help="Name of the virtualenv to create (default is 'firedrake')")
     parser.add_argument("--install", action="append", dest="packages",
-                        help="Additional packages to be installed. The address should be in format vcs+protocol://repo_url/#egg=pkg&subdirectory=pkg_dir")
+                        help="Additional packages to be installed. The address should be in format vcs+protocol://repo_url/#egg=pkg&subdirectory=pkg_dir . Some additional packages have shortcut install options for more information see --install_help.")
```
wence-: Newline in here too?  Or does the arg parser wrap automatically?

scripts/firedrake-install
```diff
@@ -254,6 +280,8 @@ class directory(object):
 
 options = config["options"]
 
+# Apply short cut package names.
+options["packages"] = [firedrake_apps.get(p, (None, p))[1] for p in options["packages"]]
```
dham: Exactly.

scripts/firedrake-install
dham: As above. The parser does the wrapping.

wence-:
Hmm, it appears I can't cancle my "good" review :).

One other thought.  Do we at least want to test that the installation of thetis and gusto via this mechanism does not break, by running `firedrake-install` with `--thetis --gusto` on travis/drone.  One of the tests could then be to try importing those packages.  That way, we don't claim that the package works, but at least we're happy that it is installable this way.

tkarna:
@wence- testing the install seems like a good idea if at all possible. Of course the import may fail if thetis/gusto master is out-of-date (e.g. thetis tries to set a firedrake parameter that's been removed).

scripts/firedrake-install
```diff
@@ -202,7 +212,9 @@ else:
                         help="Produce a verbose log of the update process in firedrake-update.log. If you have problem running this script, please include this log in any bug report you file.")
     parser.add_argument("--verbose", "-v", action="store_true", help="Produce more verbose debugging output.")
     parser.add_argument("--install", action="append", dest="packages", default=config["options"].get("packages", []),
-                        help="Additional packages to be installed. The address should be in format vcs+protocol://repo_url/#egg=pkg&subdirectory=pkg_dir")
+                        help="Additional packages to be installed. The address should be in format vcs+protocol://repo_url/#egg=pkg&subdirectory=pkg_dir. Some additional packages have shortcut install options for more information see --install_help.")
+    parser.add_argument("--install_help", action="store_true",
```
miklos1: `--install-help`?

PR #944, opened 2016-11-16
Add a DOLFIN-compatible HDF5File.

pefarrell:
This is used for adjoint checkpoints (among other things).

### DIFF
--- a/firedrake/checkpointing.py
+++ b/firedrake/checkpointing.py
@@ -299,7 +299,7 @@ class HDF5File(object):
     to disk in parallel (using HDF5) and reloading them on the same
     number of processes and a :func:`~.Mesh` constructed identically.
 
-    :arg filename: filename (including suffix ".h5) of checkpoint file.
+    :arg filename: filename (including suffix .h5) of checkpoint file.
     :arg file_mode: the access mode (one of :data:`~.FILE_READ`,
          :data:`~.FILE_CREATE`, or :data:`~.FILE_UPDATE`)
     :arg comm: communicator the writes should be collective
@@ -328,47 +328,23 @@ class HDF5File(object):
             self.mode = FILE_UPDATE
 
         self._filename = filename
-        self._timestamps = []#np.array((0,), dtype = float)
-        self.new_file()
 
-    def set_timestamp(self, t):
-        """Set the timestamp for storing.
-
-        :arg t: The timestamp value.
-        """
-        self._timestamps.append(t)        
-        if self.mode == FILE_READ:
-            return
-        attrs = self.attributes("/")
-        timestamps = attrs.get("stored_timestamps", [])
-        #timestamps.append(self._timestamps[-1])
-        attrs["stored_timestamps"] = np.concatenate((timestamps, [self._timestamps[-1]]))
-
-    def new_file(self):
-        """
-        Open a new on-disk file for writing checkpoint data.
-        """
-        self.close()
-        name = self._filename
-
-        import os
-        exists = os.path.exists(name)
+        exists = os.path.exists(filename)
         if self.mode == FILE_READ and not exists:
-            raise IOError("File '%s' does not exist, cannot be opened for reading" % name)
-        mode = self.mode
-        if mode == FILE_UPDATE and not exists:
-            mode = FILE_CREATE
+            raise IOError("File '%s' does not exist, cannot be opened for reading" % filename)
+        if file_mode == FILE_UPDATE and not exists:
+            file_mode = FILE_CREATE
 
         # Create the directory if necessary
-        dirname = os.path.dirname(name)
+        dirname = os.path.dirname(filename)
         try:
             os.makedirs(dirname)
         except OSError:
             pass
 
-        self._vwr = PETSc.ViewerHDF5().create(name, mode=mode,
+        self._vwr = PETSc.ViewerHDF5().create(filename, mode=file_mode,
                                               comm=self.comm)
-        if self.mode == FILE_READ:
+        if file_mode == FILE_READ:
             nprocs = self.attributes('/')['nprocs']
             if nprocs != self.comm.size:
                 raise ValueError("Process mismatch: written on %d, have %d" %
@@ -379,9 +355,6 @@ class HDF5File(object):
     @property
     def vwr(self):
         """The PETSc Viewer used to store and load function data."""
-        if hasattr(self, '_vwr'):
-            return self._vwr
-        self.new_file()
         return self._vwr
 
     @property
@@ -391,6 +364,24 @@ class HDF5File(object):
             return self._h5file
         self._h5file = h5i.get_h5py_file(self.vwr)
         return self._h5file
+
+    def _set_timestamp(self, t):
+        """Set the timestamp for storing.
+
+        :arg t: The timestamp value.
+        """
+        if self.mode == FILE_READ:
+            return
+        attrs = self.attributes("/")
+        timestamps = attrs.get("stored_timestamps", [])
+        attrs["stored_timestamps"] = np.concatenate((timestamps, [t]))
+
+    def get_timestamps(self):
+        """Get the timestamps this HDF5File knows about."""
+
+        attrs = self.attributes("/")
+        timestamps = attrs.get("stored_timestamps", [])
+        return timestamps
 
     def close(self):
         """Close the checkpoint file (flushing any pending writes)"""
@@ -410,27 +401,23 @@ class HDF5File(object):
 
         :arg function: The function to store.
         :arg path: the path to store the function under.
-        :arg timestamp: timestamp to add in front of path
-
-        This function is timestamp-aware and stores in the appropriate folder
-        if :meth:`set_timestamp` has been called.
+        :arg timestamp: timestamp associated with function, or None for
+                        stationary data
         """
         if self.mode is FILE_READ:
             raise IOError("Cannot store to checkpoint opened with mode 'FILE_READ'")
         if not isinstance(function, firedrake.Function):
             raise ValueError("Can only store functions")
-        
+
         if timestamp is None:
             name = os.path.basename(path)
         else:
             suffix = "/%.15e" % timestamp
             stampedpath = path + suffix
             name = os.path.basename(stampedpath)
-            attr = self.attributes(stampedpath)
-            attr["timestamp"] = timestamp
-            self.set_timestamp(timestamp)
+
         group = os.path.dirname(path)
-        
+
         with function.dat.vec_ro as v:
             self.vwr.pushGroup(group)
             oname = v.getName()
@@ -438,6 +425,11 @@ class HDF5File(object):
             v.view(self.vwr)
             v.setName(oname)
             self.vwr.popGroup()
+
+        if timestamp is not None:
+            attr = self.attributes(stampedpath)
+            attr["timestamp"] = timestamp
+            self._set_timestamp(timestamp)
 
     def read(self, function, path, timestamp=None):
         """Store a function from the checkpoint file.
--- a/firedrake/checkpointing.py
+++ b/firedrake/checkpointing.py
@@ -328,7 +328,7 @@ class HDF5File(object):
             self.mode = FILE_UPDATE
 
         self._filename = filename
-        self._timestamps = []
+        self._timestamps = []#np.array((0,), dtype = float)
         self.new_file()
 
     def set_timestamp(self, t):
@@ -339,8 +339,10 @@ class HDF5File(object):
         self._timestamps.append(t)        
         if self.mode == FILE_READ:
             return
-        attr = self.attributes("/")["stored_timestamps"]
-        attr.append(self._timestamps[-1])
+        attrs = self.attributes("/")
+        timestamps = attrs.get("stored_timestamps", [])
+        #timestamps.append(self._timestamps[-1])
+        attrs["stored_timestamps"] = np.concatenate((timestamps, [self._timestamps[-1]]))
 
     def new_file(self):
         """
--- a/firedrake/checkpointing.py
+++ b/firedrake/checkpointing.py
@@ -290,6 +290,7 @@ class DumbCheckpoint(object):
             free_comm(self.comm)
             del self.comm
 
+
 class HDF5File(object):
 
     """An object to mimic the DOLFIN HDF5File class.
--- a/firedrake/checkpointing.py
+++ b/firedrake/checkpointing.py
@@ -4,9 +4,10 @@ from pyop2.mpi import COMM_WORLD, dup_comm, free_comm
 from firedrake import hdf5interface as h5i
 import firedrake
 import numpy as np
+import os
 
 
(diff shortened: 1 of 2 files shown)


tests/output/test_hdf5file_checkpoint.py
```diff
@@ -0,0 +1,101 @@
...
+def test_attributes(f, dumpfile):
+    mesh = f.function_space().mesh()
+    dumpfile = mesh.comm.bcast(dumpfile, root=0)
+    with HDF5File(dumpfile, file_mode="w", comm=mesh.comm) as h5:
+        with pytest.raises(KeyError):
+            attrs = h5.attributes("/foo")
+            attrs["nprocs"] = 1
+            with pytest.raises(KeyError):
+                attrs = h5.attributes("/bar")
+                attrs["nprocs"]
```
wence-: Should this `with` be nested inside the previous one?

firedrake/checkpointing.py
```diff
@@ -288,3 +289,161 @@ def __del__(self):
         if hasattr(self, "comm"):
             free_comm(self.comm)
             del self.comm
+
+
+class HDF5File(object):
+
+    """An object to mimic the DOLFIN HDF5File class.
```
dham: The role of the object is to facilitate checkpointing or similar. The fact that it's DOLFIN compatible is secondary for most users.

firedrake/checkpointing.py
```diff
@@ -288,3 +289,161 @@ def __del__(self):
...
+    to disk in parallel (using HDF5) and reloading them on the same
+    number of processes and a :func:`~.Mesh` constructed identically.
+
+    :arg filename:
+    :arg file_mode: the access mode (one of :data:`~.FILE_READ`,
+         :data:`~.FILE_CREATE`, or :data:`~.FILE_UPDATE`)
+    :arg comm: communicator the writes should be collective
+         over.
+
+    The argument names are carefully chosen to match DOLFIN's HDF5File.
```
dham: As above, if we adopt the convention of referring to DOLFIN compatibility all the way through our documentation, we would end up cluttering the large majority of our docstrings with this information, since most Firedrake objects are deliberately compatible with DOLFIN or some other part of FEniCS.

firedrake/checkpointing.py
pefarrell: Sure. That information was probably better incorporated in the pull request text.

In any case, I wanted to justify the choice of kwarg names, as it's not firedrake-standard or even consistent (e.g. filename vs file_mode); unfortunately it's necessary to use the same code with both dolfin and firedrake backends.

tests/output/test_hdf5file_checkpoint.py
pefarrell: Thanks for pointing that out.

wence-:
So this new object is very similar to the existing `DumbCheckpoint`, modulo some name changes.  It would be nice if the two could be merged.  What's different:

- `HDF5File.write/read` don't look in `Function.name` for lookup, instead only using the `path`.  I think this is probably a better design than what I did, that requires you provide a function with the correct name to load into.

- `HDF5File.new_file` is slightly misleadingly named: it creates the file that you built the object with the first time you call it, then recreates it again (potentially truncating) on subsequent calls.  I added this functionality in `DumbCheckpoint` because Tuomas wanted the option of dumping to a single massive file, or having a single object backed by multiple files.  In retrospect, I think it makes more sense to actually create a new file object, rather than have DumbCheckpoint manage everything.

- `DumbCheckpoint` has some rudimentary support for storing timestep data (with some attributes as well), this doesn't exist at all in `HDF5File`.  Thetis doesn't actually use this data, but Gusto does (although it could easily be changed to manage the attributes itself).  Having said that, does this extra functionality step on any existing part of the putative dolfin-compatible API?  If not, could one just merge the functionality together?

As a followup to this, assuming that merging the superset functionality of `DumbCheckpoint` into `HDF5File` is straightforward, we'll want to update the docs to change the names of things: I can deal with the two downstream users.

firedrake/checkpointing.py
```diff
@@ -288,3 +289,159 @@ def __del__(self):
...
+
+class HDF5File(object):
+
+    """An object to facilitate checkpointing.
+
+    This checkpoint object is capable of writing :class:`~.Function`\s
+    to disk in parallel (using HDF5) and reloading them on the same
+    number of processes and a :func:`~.Mesh` constructed identically.
+
+    :arg filename:
```
wence-: Description missing.

pefarrell:
I agree it would be nice to merge the functionality. I thought I would go for the minimal perturbation approach first, and then see where the discussion went.

Regarding support for timestep data: I just checked. DOLFIN has some support for this in its own HDF5File implementation (I've never used it; I didn't know about it until now). I suggest we just implement that interface too. It is `HDF5File.write(function, name, timestamp)` where timestamp is a double. Do you agree?

wence-:
Yes, I think that's right.  So let's have:

```python
HDF5File.write(function, path, timestamp=None)
```

What about reading?  Matching on timestamp is going to be ugly, so I think when writing the timestamp we need to write some attributes.

Let's say if the timestamp is `None` and path is `/foo/bar` we write to `/foo/bar`.

If the timestamp is not None, how about we create a group: `/foo/bar` and write to:

`/foo/bar/0`, `/foo/bar/1`, etc...

On `/foo/bar` we have an attribute saying how many timesteps there are, and on each `/foo/bar/N` we have an attribute saying what the timestamp is.

Now reading from something without a timestamp should just access `/foo/bar`.  What about with a timestamp?  If I just write `HDF5File.read(..., "/foo/bar")`, and there's actually `/foo/bar/0` and `/foo/bar/1`, then it's presumably undefined what I meant.  So I presumably need a way to query about timesteps and then say `HDF5File.read(..., "/foo/bar", idx=idx)`.  Amusingly, I note that although dolfin allows you to write timestamped data, you can only ever read the first timestep.

APaganini:
Just a question on the usage: do you expect fixed timesteps? My point is:

- on the one hand, Patrick mentioned that `timestamp` is a double, from which I infer that it represents time and could take the values 0.3, 0.4, 0.7, 1.6, and so forth.

- on the other hand, you would like to read by index `idx`, which sounds like an integer that stands for the iteration count

It timesteps are not constant, the relation`timestamp` <-> `idx` might be misinterpreted.


Assuming that my understanding is correct, we should

- either save and read by`timestamp` (and be careful by reading)
- or use indeces, in which case the name `timestamp` could be misleading.

pefarrell:
I think that if we save by a float key we should query by a float key.

wence-:
I'm concerned that if we save into `/foo/bar/0.1` can we guarantee that we can reconstruct the key?

pefarrell:
We can if we canonicalise. The string representation "%.15e" is bijective with double-precision floating point numbers. (I've had to face exactly this problem in defcon, where I store information in HDF5 files by floating point key.)

wence-:
OK, I think that's good enough for me.  However, I still think it would be good if the attribute information on `/foo/bar` holds information that allows me to determine, e.g. the "last" timestep.  Imagine I'm picking up a checkpoint to continue a time-dependent sim.  I'd like to know what the time is this dump corresponds to.  I guess I can cast the str keys of `/foo/bar` to float and then find the max, but that feels sort of ugly.  Thoughts?

APaganini:
How about storing an array that contains all intermediate times?
You can use it to find
- the final time, 
- how many time iterations have been performed, 
- which timesteps have been used.

pefarrell:
Yes, I think keeping an attribute with a list of all the keys would be a good thing to do.

pefarrell:
OK, we've modified the HDF5File to support getting and setting by timestamp.

tests/output/test_hdf5file_checkpoint.py
```diff
@@ -0,0 +1,125 @@
...
+            h5.write(f, "/function")
+
+def test_multiple_timestamps(f, dumpfile):
+    with HDF5File(dumpfile, "w") as h5:
+        h5.write(f, "/solution", timestamp=0.1)
+        h5.write(f, "/solution", timestamp=0.2)
+
+        timestamps = h5.get_timestamps()
+
+        assert np.allclose(timestamps, [0.1, 0.2])
```
wence-: Can this test write two different values at the timesteps and assert that you can read them back in?

tests/output/test_hdf5file_checkpoint.py
```diff
@@ -0,0 +1,125 @@
...
+            attrs = h5.attributes("/bar")
+            attrs["nprocs"]
+
+        h5.write(mesh.coordinates, "/coords")
+        attrs = h5.attributes("/coords")
+        attrs["dimension"] = mesh.coordinates.dat.cdim
+
+        assert attrs["dimension"] == mesh.coordinates.dat.cdim
+
+def test_store_read_only_ioerror(f, dumpfile):
```
wence-: Need an extra blank line here.

tests/output/test_hdf5file_checkpoint.py
```diff
@@ -0,0 +1,125 @@
...
+    # Make file
+    with HDF5File(dumpfile, "w") as h5:
+        pass
+    with HDF5File(dumpfile, "r") as h5:
+        with pytest.raises(IOError):
+            h5.write(f, "/function")
+
+def test_multiple_timestamps(f, dumpfile):
```
wence-: Likewise here.

tests/output/test_hdf5file_checkpoint.py
```diff
@@ -0,0 +1,125 @@
...
+        h5.write(f, "/solution", timestamp=0.1)
+        h5.write(f, "/solution", timestamp=0.2)
+
+        timestamps = h5.get_timestamps()
+
+        assert np.allclose(timestamps, [0.1, 0.2])
+
+
+
+if __name__ == "__main__":
```
wence-: And one fewer blank lines here.

tests/output/test_hdf5file_checkpoint.py
wence-: That tests that writing and reading from a single timestamp works, no?  I was meaning test that writing and reading from two different timestamps under the same group gives the expected two different solutions.

wence-:
flake8 is still unhappy (see, for example https://jenkins.ese.ic.ac.uk:8000/firedrakeproject/firedrake/73).  If you `pip install flake8` and run `make lint` in the firedrake source tree you can see the errors locally.

tests/output/test_hdf5file_checkpoint.py
APaganini: ok, I'll do that

APaganini: this is now tested in line 69

wence-:
This currently fails tests on mac (test_attributes).  It appears to be an actual problem, but it's slightly subtle.  This is what I think is going on:

The `HDF5File` object holds a PETSc viewer (which we use to write data).  As well as that, we get the HDF5 file ID from the PETSc viewer, and turn it into an h5py file.  The C reference count of the hdf5 file is now two (one from PETSc, one from h5py).  When the `HDF5File` object goes out of scope, we delete the references to the PETSc viewer and the h5py file (so the refcount is zero and the file is closed).

However, If we get at the attributes of the h5py file (via `HDF5File.attributes`)  and hold on to them past the end of the context manager in which the file is open, then we get errors:

e.g. this:

```python
def test():
   with HDF5File("foo.h5", file_mode="w") as h5:
       attrs = h5.attributes("/")
test()
```

Results in an error.

This appears to be because h5py manages refcounts in a strange way that doesn't play nicely with the fact that its original reference to the file is not "first class".

In the dumbcheckpoint class we don't see this issue because the attribute manager is never explicitly exposed.  We could allow read only access to the attributes by returning `dict(self.h5file[path].attrs)` in `attributes`, but then we need to add a `set_attribute` function.

Ugh, I hate refcounting.

Thoughts?

pefarrell:
One possible take on this is that, since we're wrapping a HDF5 file via PETSc, we should use PETSc's PetscViewerHDF5WriteAttribute/PetscViewerHDF5ReadAttribute. (I copied the h5py stuff from DumbCheckpoint, but I agree it causes a problem.) What do you think?

wence-:
That would be OK, although the only "array-like" attribute you can write through that interface is a string.  That's probably OK for these purposes though.  My reason for doing attribute access through h5py was mostly laziness, since it was much easier to drive from python.  However, as you note, it does create issues.  The other option would be (given that we're only using HDF5 as a dumb datastore) to upgrade to h5py v2.6 which has support for parallel collective writes and do all the dumping purely though h5py.  That might be a little more work?

I'd be happy to access all the attributes through the PETSc interface and just ditch having an exposed h5file property.

pefarrell:
Another possibility: could we have attributes() make a proxy object that holds a weakref to the h5py file? That way a user who keeps attributes() around will not keep a strong reference and thus not cause the crash?

wence-:
I think that could work, but I realise the problem is more pervasive in some sense.  In that holding on to any sub-group of `h5file` after `HDF5File` goes out of scope results in this problem.  This is also a problem with the `DumbCheckpoint` object.

pefarrell:
It sounds like the Right Way to do this is to improve PETSc's HDF5 wrapping --- or cut it out completely. Can we cut it out completely? (You mentioned h5py >2.6 but I don't understand what helps there.)

wence-:
h5py < 2.6 doesn't support collective writes.  So you don't have a "high performance" write path.  OTOH 2.6 does support it: https://github.com/h5py/h5py/blob/master/examples/collective_io.py

Maybe it doesn't matter as long as each process writes to its own part of the dataset (other than for speed).  Let's assume it doesn't matter for now.  I think we can do this by doing:

```python
from firedrake import *
import h5py

mesh = UnitSquareMesh(4, 5)

g = Function(V)
g.interpolate(SpatialCoordinate(mesh)[0])

f = h5py.File("foo.h5", mode="w")

with g.dat.vec_ro as v:
    dset = f.create_dataset("foo", shape=(v.getSize(), ), dtype=g.dat.dtype)

    dset[slice(*v.getOwnershipRange())] = v.array_r

f.close()
```

Or, if h5py is built with MPI and is 2.6 you can use the MPI-IO driver:

```python
from firedrake import *
import h5py
mesh = UnitSquareMesh(4, 5)

g = Function(V)
g.interpolate(SpatialCoordinate(mesh)[0])

f = h5py.File("foo.h5", mode="w", driver="mpio", comm=g.comm)

with g.dat.vec_ro as v:
    dset = f.create_dataset("foo", shape=(v.getSize(), ), dtype=g.dat.dtype)

    with dset.collective:
        dset[slice(*v.getOwnershipRange())] = v.array_r

f.close()
```

pefarrell:
I've had a first go at this. It works for me, with h5py compiled with MPI support. It might not work with h5py compiled without MPI support, I don't have that installed on my machine.

wence-:
Just getting back to this now.  It seems like if you don't have MPI support enabled then this approach fails.  Fortunately, we control installation of h5py, so I can update the firedrake-install script to do that. I'll also raise the appropriate error if MPI support is not enabled.

pefarrell:
Great, that solves all problems nicely.

wence-:
Superseded by #952

PR #945, opened 2016-11-16
Disable linux builders

dham:
The Travis linux builders are constantly timing out, so it's time to get a little more assertive about moving to Drone. The .yml isn't fully cleaned up as a pure Mac builder because this is intended to be temporary until we can get Mac over onto drone too.


wence-:
Tick, I think.

PR #948, opened 2016-11-25
Matrix improvements

wence-:
Add ability to get matrix info from matfree matrices, and reduce memory footprint when no boundary conditions are required.

Also, make it possible to specify on a per-solve basis whether to build AIJ or BAIJ matrices.

Doesn't propagate into sub-matrices of nest matrices yet (but that can be controlled globally with `parameters["default_sub_matrix_type"]`.

### DIFF
--- a/docs/source/solving-interface.rst
+++ b/docs/source/solving-interface.rst
@@ -457,6 +457,40 @@ precondition the Krylov system solving for :math:`S`, you can also use
 a `least squares commutator <LSC_>`_, see the relevant section of the
 `PETSc manual pages <fieldsplit_>`_ for more details.
 
+Specifying assembled matrix types
++++++++++++++++++++++++++++++++++
+
+Firedrake supports assembling linear operators in a number of
+different formats.  Either as assembled sparse matrices, or as
+matrix-free operators that only provide matrix-vector products.  Since
+matrix-free actions require special preconditioning, they have
+:doc:`their own section in the manual <matrix-free>`.  Even in the
+sparse matrix case there are a few options.  Firedrake can build
+nested block matrices, or monolithic sparse matrices.  The latter
+admit a wider range of preconditioners, but are memory-inefficient
+when using fieldsplit preconditioning.  Even monolithic matrices have
+choices, specifically, if there is a block structure, whether that
+should be exploited or not.  Again the trade-off is between memory
+efficiency (in the block case) and access to a slightly smaller range
+of preconditioners.
+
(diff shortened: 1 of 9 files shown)


docs/source/solving-interface.rst
```diff
@@ -457,6 +457,40 @@ precondition the Krylov system solving for :math:`S`, you can also use
...
+The default matrix type can be set with the global parameter
+``parameters["default_matrix_type"]``.  In the case where the matrix
+is assembled as a nested matrix, there is a choice as to the type of
+the blocks (they may be "aij" or "baij").  The default choice can be
+controlled with ``parameters["default_sub_matrix_type"]``.  For
+finer-grained control over the matrix type, one can provide it when
+calling :func:`~.assemble` through the ``mat_type`` and
+``sub_mat_type`` keyword arguments.  When using variational solvers,
+the matrix type is controlled through use of the ``solver_parameters``
+dictionary by specifying the ``"mat_type"`` entry.
```
miklos1: If I have a nested matrix, can I make some of submatrices matrix-free while others assembled?

wence-: right now, no.  We don't have fine-enough grained control over the pieces.  I'm inclined to not do that..

docs/source/solving-interface.rst
```diff
@@ -457,6 +457,40 @@ precondition the Krylov system solving for :math:`S`, you can also use
 a `least squares commutator <LSC_>`_, see the relevant section of the
 `PETSc manual pages <fieldsplit_>`_ for more details.
 
+Specifying assembled matrix types
++++++++++++++++++++++++++++++++++
+
+Firedrake supports assembling linear operators in a number of
```
dham: "the assembly of"

rckirby:
The typical use case for some blocks assembled and others not is that one wants an algebraic preconditioner (e.g. AMG, LU) on a sub-block in a field split.  In that case, using an "AssembledPC" on that block will force the assembly inside the preconditioner and leave the other blocks alone.

I think it would be painful to have a structure that supported this at the matrix rather than PC level, and can't think of a compelling use case, either.

docs/source/solving-interface.rst
wence-: As Rob points out, you can use mat_type "matfree" and control what goes on will all the pieces as part of the PC.  You might think that we would be able to control what comes back from MatGetSubMatrix, but the appropriate options prefix is not set at the point when it is called.

PR #951, opened 2016-12-01
Indexed vfs

wence-:
Fix silly bugs in indexed VFS.  Needs OP2/PyOP2#512.

### DIFF
--- a/firedrake/functionspacedata.py
+++ b/firedrake/functionspacedata.py
@@ -95,6 +95,8 @@ def get_node_set(mesh, nodes_per_entity):
         node_set = op2.ExtrudedSet(node_set, layers=mesh.layers)
 
     assert global_numbering.getStorageSize() == node_set.total_size
+    if not extruded and node_set.total_size >= (1 << 28):
+        raise RuntimeError("Problems with more than %d nodes per process unsupported", (1 << 28))
     return node_set
 
 
--- a/firedrake/functionspacedata.py
+++ b/firedrake/functionspacedata.py
@@ -503,10 +503,10 @@ class FunctionSpaceData(object):
                         # And in the generated code we can then
                         # extract the information to discard the
                         # correct entries.
-                        val = 2 ** (30 - bc.function_space().component)
                         # bcids is sorted, so use searchsorted to find indices
                         idx = numpy.searchsorted(bcids, bc.nodes)
-                        negids[idx] += val
+                        # Set appropriate bit
+                        negids[idx] |= (1 << (30 - bc.function_space().component))
                         decorate = True
                 node_list_bc = numpy.arange(self.node_set.total_size,
                                             dtype=numpy.int32)
--- a/tests/regression/test_vfs_component_bcs.py
+++ b/tests/regression/test_vfs_component_bcs.py
@@ -157,6 +157,44 @@ def test_cant_subscript_3_cmpt(m):
         V.sub(3)
 
 
+def test_stokes_component_all():
+    mesh = UnitSquareMesh(10, 10)
+
+    # Define function spaces
+    V = VectorFunctionSpace(mesh, "CG", 2)
+    Q = FunctionSpace(mesh, "CG", 1)
+    W = V * Q
+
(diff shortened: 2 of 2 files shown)


florianwechsung:
I think there is still a problem with this: when mixing component wise boundary conditions with full boundary conditions (i.e. for the entire vector space). 

```
from firedrake import *
import numpy as np
mesh = UnitSquareMesh(2, 2)
Q = FunctionSpace(mesh, "CG", 1)
W = V * Q

bc0 = DirichletBC(W.sub(0), Constant((0,0)), [3, 4])
bc1 = DirichletBC(W.sub(0), Constant((1,0)), 1)
bcs_full = [bc0, bc1]

bc0 = DirichletBC(W.sub(0).sub(0), 0, [3, 4])
bc1 = DirichletBC(W.sub(0).sub(1), 0, [3, 4])
bc2 = DirichletBC(W.sub(0).sub(0), 1, 1)
bc3 = DirichletBC(W.sub(0).sub(1), 0, 1)
bcs_cmp = [bc0, bc1, bc2, bc3]

bc0 = DirichletBC(W.sub(0), Constant((0,0)), [3, 4])
bc1 = DirichletBC(W.sub(0).sub(0), 1, 1)
bc2 = DirichletBC(W.sub(0).sub(1), 0, 1)
bcs_mixed = [bc0, bc1, bc2]

(u, p) = TrialFunctions(W)
(v, q) = TestFunctions(W)
a = inner(grad(u), grad(v))*dx + div(v)*p*dx + q*div(u)*dx

def ToNP(A):
    return A.M.handle.convert(PETSc.Mat.Type.DENSE).getDenseArray()
A_full = ToNP(assemble(a, bcs=bcs_full, mat_type = "aij"))
A_cmp = ToNP(assemble(a, bcs=bcs_cmp, mat_type = "aij"))
A_mixed = ToNP(assemble(a, bcs=bcs_mixed, mat_type = "aij"))
print np.allclose(A_full, A_cmp)
print np.allclose(A_mixed, A_full)
```

I'll check if it only occurs when using mixed function spaces or always.

wence-:
You can access by just calling `__getitem__` on the matrix:  `A.M.handle[:, :]`

wence-:
I think I know what's going on, the logic for deciding whether to throw things away in the generated code is slightly wrong.

We do:

`drop_full_row = 0 if rmap.vector_index is not None else 1`

In your case, we have a `rmap.vector_index` that is not None, so `drop_full_row` is False.  But, the map entry for a block, because we've also set some "drop both components" is something like:

```[-1610612737, -1610612738,          -3]```

So we correctly determine to drop both components for the first two entries, but then for the last entry, we have:

```c
                        if ( block_row < 0 ) {
                            tmp = -(block_row + 1);
                            discard = 1;
                            block_row = tmp & ~0x70000000;
                        }
                        for ( int k = 0; k < 2; k++ ) {
                            if ( discard && (0 || ((tmp & (1 << (30 - k))) != 0)) ) {
                                rowmap[j*2 + k] = -1;
                            } else {
                                rowmap[j*2 + k] = (block_row)*2 + k;
                            }
                        }
```

Let's walk through this:

```
block_row = -3
tmp = ~-3 = 2
discard = 1
block_row = (~-3) & ~0x70000000 = 2
```

So `discard` is True, so we'll hit the first branch, but `tmp` has no high bits set, so the second half of the condition will always evaluate to false.

So I think we need to make that condition evaluate:

```c
if (discard && (0 || !(tmp & 0x70000000) || ...)) {
...
}
```
So if we're discarding, and there are no high bits set, then discard the whole row/column.

Do you want to have a go (this is in pyop2/sequential.py around line 350.

florianwechsung:
Brilliant, that fixed it. I have opened a pull request on PyOP2, should I add a test to firedrake checking this functionality once the pr has been merged?

wence-:
I've merged the PyOP2 change and added your test.  Hopefully this will go green.

florianwechsung:
brilliant, looks good!

PR #954, opened 2016-12-05
Add script to move firedrake plus license.

nickaj:
This adds a script which allows one to move an intact firedrake install around their filesystem
without breaking it. Also updates the license (as it is a derivative work) and readme.

### DIFF
--- a/scripts/clone.py
+++ b/scripts/clone.py
@@ -1,0 +1,350 @@
+#!/usr/bin/env python
+from __future__ import with_statement
+import logging
+import optparse
+import os
+import os.path
+import re
+import shutil
+import subprocess
+import sys
+import itertools
+import time
+
+logger = logging.getLogger()
+
+env_bin_dir = 'bin'
+if sys.platform == 'win32':
+    env_bin_dir = 'Scripts'
+
+class UserError(Exception):
+    pass
+
+def in_directory(path, tree):
+    "Make both paths absolute."
+    directory = os.path.join(os.path.realpath(tree), '')
+    file = os.path.realpath(path)
+
+    #return true, if the common prefix of both is equal to directory
+    #e.g. /a/b/c/d.rst and directory is /a/b, the common prefix is /a/b
+    return os.path.commonprefix([path, tree]) == tree
+
+
+def _dirmatch(path, matchwith):
+    """Check if path is within matchwith's tree.
+
+    >>> _dirmatch('/home/foo/bar', '/home/foo/bar')
+    True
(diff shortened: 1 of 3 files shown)


scripts/clone.py
```diff
@@ -0,0 +1,350 @@
...
+if sys.platform == 'win32':
+    env_bin_dir = 'Scripts'
+
+class UserError(Exception):
+    pass
+
+def in_directory(path, tree):
+    "Make both paths absolute."
+    directory = os.path.join(os.path.realpath(tree), '')
+    file = os.path.realpath(path)
```
wence-: unused variable?

scripts/clone.py
```diff
@@ -0,0 +1,350 @@
...
+    >>> _dirmatch('/home/foo/bar2', '/home/foo/bar')
+    False
+    >>> _dirmatch('/home/foo/bar2/etc', '/home/foo/bar')
+    False
+    """
+    matchlen = len(matchwith)
+    if (path.startswith(matchwith)
+        and path[matchlen:matchlen + 1] in [os.sep, '']):
+        return True
+    return False
```
wence-: AKA `return condition`

scripts/clone.py
```diff
@@ -0,0 +1,350 @@
...
+    logger.debug("stdout from sed: %s" % stdout)
+    logger.debug("errcode from sed: %s" % errcode)
+    
+
+def clone_virtualenv(src_dir, dst_dir, do_copy=True):
+    """ Central Function: Move the whole directory tree, except .pyc files """
+    
+    # It is possible that you want to go through the process more than once, but the second time, you don't want to do the copy.
+    # This is risky, as if do_copy is False, you will skip the checks on file existence etc.
+    if do_copy==True:
```
wence-: `if do_copy`

wence-:
I haven't read through this properly.  What I would like is that this is *absolutely* failsafe.  That is, if, for whatever reason, the script falls over halfway through, does it leave my virtualenv in an unusable state?  That would be bad.  I presume this is a one-off thing: I create my virtualenv somehow, then I run this script (how?).  Once I've done that, I'm ready to use the new "boxed" virtualenv.  So presumably my original virtualenv is always "read-only".

wence-:
Is the current status that we are going to do this as a stand-alone script (not in the firedrake repo?).

nickaj:
Yes, please could you drop this PR and generate a new repo under firedrakeproject that I can issue PRs against (so someone from the team reviews the code).

wence-:
Done, see https://github.com/firedrakeproject/relocate-venv

PR #955, opened 2016-12-06
Interpolation ast

tj-sun:
This branch changes interpolation.py so that it returns kernels with COFFEE ast, rather than a string.


FabioLuporini:
Looks OK

tj-sun:
Cannot locally replicate the "external library" errors in jenkins

wence-:
That's because you haven't updated your firedrake installation recently.  But it's OK, because it's not your change that's producing the problem (as the merge commit demonstrates).

PR #960, opened 2016-12-07
Shorten names of cell and facet sets for use as profiling tags

mlange05:
Goes alongside OP2/Pyop2#516 to provide kernel-specific profiling.

### DIFF
--- a/firedrake/mesh.py
+++ b/firedrake/mesh.py
@@ -77,7 +77,7 @@ class _Facets(object):
             else:
                 base = self.mesh._base_mesh.exterior_facets.set
             return op2.ExtrudedSet(base, layers=self.mesh.layers)
-        return op2.Set(size, "%s_%s_facets" % (self.mesh.name, self.kind),
+        return op2.Set(size, "%sFacets" % self.kind.capitalize()[:3],
                        comm=self.mesh.comm)
 
     @utils.cached_property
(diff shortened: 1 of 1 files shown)


firedrake/mesh.py
```diff
@@ -77,7 +77,8 @@ def set(self):
             else:
                 base = self.mesh._base_mesh.exterior_facets.set
             return op2.ExtrudedSet(base, layers=self.mesh.layers)
-        return op2.Set(size, "%s_%s_facets" % (self.mesh.name, self.kind),
+        return op2.Set(size, "%s%sFacets" % (self.mesh.name,
+                                             self.kind.capitalize()[:3]),
```
wence-: I recommend removing `mesh.name` from these format strings, because I think they are basically always `None`.

wence-: Or rather, they're either "plexmesh" or the name of a mesh file.  I don't think that's useful information.

mlange05: Agreed. I also re-instated the "ParLoop" part of the tag in the PyOP2 PR, so that we now get the following output from `-log_view`:
```
ParLoopExecute       240 ...
ParLoopCells         240 ...
ParLoopHaloEnd       240 ...
ParLoopReductionBegin     240 ...
ParLoopReductionEnd     240 ...
ParLoopIntFacets     240 ...
ParLoopExtFacets     240 ...
```

PR #961, opened 2016-12-07
modified to_expression to handle a single string

APaganini:
added 2 lines to the function `expression.to_expression`
so that we can mimic dolfin and use a string for a
scalar Dirichlet boundary condition,  e.g.,
`DirichletBC(V, "0.1", "on_boundary")`


firedrake/expression.py
```diff
@@ -220,10 +220,14 @@ def to_expression(val, **kwargs):
...
-        expr = "%s" % val
+    
+    if isinstance(val, str):
+        expr = [val]
+    else:
+        try:
+            expr = ["%s" % v for v in val]
+        except TypeError:
+            # Not iterable
+            expr = "%s" % val
```
miklos1:
```python
if isinstance(val, collections.Iterable) and not isinstance(val, str):
    expr = ["%s" % v for v in val]
else:
   expr = "%s" % val
```
maybe a bit clearer?

firedrake/expression.py
```diff
@@ -220,10 +220,14 @@ def to_expression(val, **kwargs):
     :arg \*\*kwargs: keyword arguments passed to the
          :class:`Expression` constructor (which see).
     """
-    try:
-        expr = ["%s" % v for v in val]
-    except TypeError:
-        # Not iterable
-        expr = "%s" % val
+
```
wence-: Please remove the whitespace here.

APaganini:
ok, in which folder should the test go?

wence-:
In `tests/regression`.  It's to do with bcs, so you could add it to `tests/regression/test_bcs.py` I think.

tests/regression/test_bcs.py
```diff
@@ -1,403 +0,0 @@
-import pytest
```
wence-: Why do you delete this file?

tests/regression/test_to_expression.py
```diff
@@ -0,0 +1,44 @@
...
+    v1 = expression.to_expression(0.)
+    v2 = expression.to_expression([0.])
+    v3 = expression.to_expression("0.")
+    v4 = expression.to_expression(["0."])
+
+    f1 = interpolate(v1, V)
+    f2 = interpolate(v2, V)
+    f3 = interpolate(v3, V)
+    f4 = interpolate(v4, V)
+
```
wence-: This is a bad test, because functions are initialised to zero, and so you're not testing that the expression interpolation did anything (since you're interpolating zero). Also, they could all be producing the same wrong answer (since you're checking them only against each other)

tests/regression/test_bcs.py
APaganini: I renamed it `test_to_expression.py` because it's not a test on BCs.
DirichletBCs calls the function `expression.to_expression`, and this
is the function I modified, and that should therefore be tested.

tests/regression/test_to_expression.py
APaganini: I didn't know that functions are initialized to zero. The goal of the test
is to check that you can write the same value in different fashions and
that all do the same thing. I'll add a line to test that they do the right thing.

tests/regression/test_bcs.py
wence-: Well, this file was definitely testing BC functionality (e.g. bc.apply and so forth).  Moreover, you didn't rename it, because all of the tests in this file are now gone (and to reappear in test_to_expression.py).

APaganini: ok that was I mistake, I didn't noticed that `test_bcs.py` already existed. I'll put it back.

wence-:
This causes a failing test: `test_bcs.py` line 48.  Because now the value "foo" is no longer considered an invalid value.  This was always a pretty weak test, you could change it to `object()` I think.

wence-:
> and about this comment by Lawrence

> > This causes a failing test: test_bcs.py line 48. Because now the value foo is no longer considered an invalid value. This was always a pretty weak test, you could change it to object() I think.
> I don't know what to do... How do you test whether the expression is a valid C snippet? What fails if I type SIN instead of sin?

You get, later, a `CompilationError`.  Let's just remove this part of the test, so it reads:

```python
@pytest.mark.parametrize('v', [(0, 0)])
def test_init_bcs_illegal(mesh, v):
    "Initialise a DirichletBC with illegal values."
    with pytest.raises(RuntimeError):
        DirichletBC(FunctionSpace(mesh, "CG", 1), v, 0)
```

wence-:
Merged with some minor cleanups.

PR #962, opened 2016-12-07
Apaganini/subtract vector

APaganini:
Added `__sub__`and `__isub__' to `firedrake/firedrake/vectory.py`and
added a test in `firedrake/tests/regression/test_vector.py`

### DIFF
--- a/firedrake/vector.py
+++ b/firedrake/vector.py
@@ -101,7 +101,7 @@ class Vector(object):
         try:
             self.dat += other.dat
         except AttributeError:
-            self += other
+            self.dat += other
         return self
 
     def __sub__(self, other):
@@ -118,7 +118,7 @@ class Vector(object):
         try:
             self.dat -= other.dat
         except AttributeError:
-            self -= other
+            self.dat -= other
         return self
 
     def apply(self, action):
--- a/firedrake/vector.py
+++ b/firedrake/vector.py
@@ -103,6 +103,23 @@ class Vector(object):
         except AttributeError:
             self += other
         return self
+    
+    def __sub__(self, other):
+        """Add other to self"""
+        diff = self.copy()
+        try:
+            diff.dat -= other.dat
+        except AttributeError:
+            diff -= other
+        return diff
+
+    def __isub__(self, other):
+        """Add other to self"""
+        try:
+            self.dat -= other.dat
+        except AttributeError:
+            self -= other
+        return self
 
     def apply(self, action):
         """Finalise vector assembly. This is not actually required in
--- a/firedrake/vector.py
+++ b/firedrake/vector.py
@@ -103,7 +103,7 @@ class Vector(object):
         except AttributeError:
             self += other
         return self
-    
+
     def __sub__(self, other):
         """Add other to self"""
         diff = self.copy()
--- a/tests/regression/test_vector.py
+++ b/tests/regression/test_vector.py
@@ -94,6 +94,15 @@ def test_subtraction(f):
     assert (w.array() == 0.).all()
 
 
+def test_isub(f):
+    f.interpolate(Expression("2"))
+    v = f.vector()
+    y = Vector(v)
+    v -= y
+
+    assert (v.array() == 0.).all()
+
+
 def test_scale(f):
     f.interpolate(Expression("3"))
     v = f.vector()
--- a/tests/regression/test_vector.py
+++ b/tests/regression/test_vector.py
@@ -90,8 +90,10 @@ def test_addition(f):
     v = f.vector()
     y = Vector(v)
     w = v + y
-
     assert (w.array() == 4.).all()
+
+    w = v + 3.
+    assert (w.array() == 5.).all()
 
 
 def test_iadd(f):
@@ -99,7 +101,6 @@ def test_iadd(f):
     v = f.vector()
     y = Vector(v)
     v += y
-
     assert (v.array() == 4.).all()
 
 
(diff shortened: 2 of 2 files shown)


dorugeber:
Can you not kill that other test file please?

APaganini:
@dorugeber what do you mean?

thomasgibson:
The file `tests/regression/test_bcs.py` appears to have been killed. Was that intentional?

APaganini:
not it's a mistake, I'll put it back. I'm sorry

wence-:
When we merge #961, if we then merge #962 on top, we'll now revert the changes you made to improve `to_expression`.  The mistake was to base this branch on top of your expressions branch, rather than master.  Please instead do the following:

```
git reset --hard master
git cherry-pick e07178e
git cherry-pick cf5b8ae
git push -f
```

wence-:
OK, your master wasn't where I thought it was.  Try now (assuming that the origin remote points to firedrakeproject/firedrake)

```
git rebase --onto origin/master a485c60 apaganini/subtract_vector
git diff origin/master
```

This should should only changes to firedrake/vector.py and tests/regression/test_vector.py

wence-:
I like http://learngitbranching.js.org/ to understand what's going on.

APaganini:
it returns
```diff --git a/firedrake/vector.py b/firedrake/vector.py
index a7b7bfc..8a466ba 100644
--- a/firedrake/vector.py
+++ b/firedrake/vector.py
@@ -104,6 +104,23 @@ class Vector(object):
             self += other
         return self
 
+    def __sub__(self, other):
+        """Add other to self"""
+        diff = self.copy()
+        try:
+            diff.dat -= other.dat
+        except AttributeError:
+            diff -= other
+        return diff
+
+    def __isub__(self, other):
+        """Add other to self"""
+        try:
+            self.dat -= other.dat
+        except AttributeError:
+            self -= other
+        return self
+
     def apply(self, action):
         """Finalise vector assembly. This is not actually required in
         Firedrake but is provided for Dolfin compatibility."""
diff --git a/tests/regression/test_vector.py b/tests/regression/test_vector.py
index ff10374..a83c3fb 100644
--- a/tests/regression/test_vector.py
+++ b/tests/regression/test_vector.py
@@ -85,6 +85,15 @@ def test_axpy(f):
     assert (v.array() == 14.0).all()
 
 
+def test_subtraction(f):
+    f.interpolate(Expression("2"))
+    v = f.vector()
+    y = Vector(v)
+    w = v - y
+
+    assert (w.array() == 0.).all()
+
+
 def test_scale(f):
     f.interpolate(Expression("3"))
     v = f.vector()
```

wence-:
OK great, please `git push -f` that state.

wence-:
Can you please also add a test of the `__isub__` functionality?  (that's the one that does `x -= y` to make an in-place modification of x (but doesn't change y)).  That new part of the code isn't currently tested.

APaganini:
in which branch should I work? apaganini/subract_vector?

wence-:
Yes, please add a new commit on this branch that adds an appropriate test for `__isub__`.

APaganini:
I don't know how to test the exception `AttributeError`

APaganini:
and about this comment by Lawrence

> This causes a failing test: `test_bcs.py` line 48. Because now the value `foo` is no longer considered an invalid value. This was always a pretty weak test, you could change it to object() I think.

I don't know what to do... How do you test the expression is a valid `C` snippet? What fails if I type `SIN` instead of `sin`?

wence-:
> I don't know how to test the exception AttributeError

If you do:

```python
f = Function(V)

fv = f.vector()

fv -= 2.0
```

Then `f.dat.data` should be -2.0, this hits the exception catching branch.

APaganini:
summing/subtracting  a `vector`and a `float`is not supported by this implementation
(note that `__add_`_ and `__iadd__` as well can't treat it because they enter an infinite recursion)

wence-:
I'm really confused now.  You have two open pull requests, but you keep on mixing fixes for one into fixes for the other.  Your latest commit 300ca44 was presumably meant for #961.

wence-:
Thanks for noticing the bug in the exception catching path!

APaganini:
well, I thought that #961 and #962 went together, but I probably misunderstood...
Do you want me to revert the change in `test_bcs.py`?

wence-:
We fixed the branch issue (you can see this graphically if you run `gitk --all`), so now #961 and #962 introduce completely separate  features/tests.

Please don't `git revert` it, because then if we merge this branch after #961, then the patch will be reverted there too.

Instead:

```
git checkout apaganini/firedrake.expression
git cherry-pick 300ca44
git push
git checkout apaganini/subtract_vector
git reset --hard HEAD~
git push -f
```

This picks the test_bcs fix from this branch (where it doesn't want to live) onto the head of firedrake.expression (where it does), updates that. Then we switch back to this branch, drop the test_bcs fix from this branch (because now it's coming in in firedrake.expression instead) and update the remote as well.

In passing, I recommend that if you're working on multiple branches simultaneously and you find yourself losing track of which one you're currently on when you're in a terminal that you set up a PROMPT_COMMAND that will remind you.  For example, see here https://git-scm.com/book/en/v2/Git-in-Other-Environments-Git-in-Bash if you use bash.

APaganini:
Many thanks for the clean explanation, for the solution of this git-issue, and for your patience.

wence-:
Thanks, merged.

PR #971, opened 2016-12-15
add FInAT to zenodo

dorugeber:
I tentatively propose "a smarter library of finite elements", rather than "FInAT is not a tabulator" (which feels out of place with the others)


scripts/firedrake-zenodo
```diff
@@ -17,6 +17,7 @@ descriptions = OrderedDict([
     ("tsfc", "The Two Stage Form Compiler"),
     ("COFFEE", "A Compiler for Fast Expression Evaluation"),
     ("ufl", "The Unified Form Language"),
+    ("FInAT", "a smarter library of finite elements"),
```
miklos1: Inconsistent capitalisation: all other description start with a capital letter.

dorugeber: No they don't.

PR #986, opened 2017-01-16
Future imports

miklos1:


miklos1:
At this point we may start considering whether we want `flake8-future-import` in the default Firedrake install.

PR #987, opened 2017-01-18
Correction for documentation of FunctionSpace

juan42:
dof_count and node_count do return the number of _local_ degrees of freedom

### DIFF
--- a/firedrake/functionspaceimpl.py
+++ b/firedrake/functionspaceimpl.py
@@ -391,7 +391,7 @@ class FunctionSpace(object):
 
     @property
     def node_count(self):
-        """The number of global nodes in the function space.  If the
+        """The number of local nodes in the function space.  If the
         :class:`FunctionSpace` has :attr:`rank` 0, this is equal to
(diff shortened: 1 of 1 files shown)


firedrake/functionspaceimpl.py
```diff
@@ -391,15 +391,15 @@ def __mul__(self, other):
 
     @property
     def node_count(self):
-        """The number of global nodes in the function space.  If the
+        """The number of local nodes in the function space.  If the
```
dham: I think there might be a degree of confusion about nomenclature here. I think "local" means "per element".

dham: How about "The number of nodes of this function space on this process."

miklos1: How about "The number of nodes of this function space _visible_ on this process."

wence-:
Maybe add a parenthesised ("includes halo nodes") ?

wence-:
Please fix the trailing whitespace.

PR #988, opened 2017-01-19
A few Python 3 compatibility fixes

miklos1:


firedrake/bcs.py
```diff
@@ -57,7 +58,7 @@ def __init__(self, V, g, sub_domain, method="topological"):
         self._original_arg = self.function_arg
         if sub_domain == "on_boundary":
             self.sub_domain = \
-                map(int, V.mesh().topology.exterior_facets.unique_markers)
+                list(map(int, V.mesh().topology.exterior_facets.unique_markers))
```
wence-: FWIW, I would prefer a `tuple` here.

firedrake/slate/slac/compiler.py
```diff
@@ -164,7 +165,7 @@ def compile_expression(slate_expr, tsfc_parameters=None):
                        oriented=oriented,
                        subdomain_id="otherwise",
                        domain_number=0,
-                       coefficient_map=range(len(slate_expr.coefficients())),
+                       coefficient_map=list(range(len(slate_expr.coefficients()))),
```
wence-: Prefer `tuple` here too?

firedrake/slate/slac/kernel_builder.py
```diff
@@ -125,20 +126,20 @@ def prepare_tsfc_kernels(temps, tsfc_parameters=None):
...
         integrals = expr.form.integrals()
         mapper = RemoveRestrictions()
-        integrals = map(partial(map_integrand_dags, mapper), integrals)
+        integrals = list(map(partial(map_integrand_dags, mapper), integrals))
         prefix = "subkernel%d_" % len(kernel_exprs)
 
         # Now we split integrals by type: interior_facet and all other cases
         # First, the interior_facet case:
-        interior_facet_intergrals = filter(lambda x: x.integral_type() == "interior_facet", integrals)
+        interior_facet_intergrals = list(filter(lambda x: x.integral_type() == "interior_facet", integrals))
```
wence-: No need to coerce to list here, because we only use the operable once?

firedrake/slate/slac/kernel_builder.py
```diff
@@ -125,20 +126,20 @@ def prepare_tsfc_kernels(temps, tsfc_parameters=None):
...
 
         # Now we reconstruct all interior_facet integrals to be of type: exterior_facet
         # This is because locally over each cell, SLATE views them as being "exterior"
         # with respect to the cell.
         interior_facet_intergrals = [it.reconstruct(integral_type="exterior_facet")
                                      for it in interior_facet_intergrals]
         # Now for the rest:
-        other_integrals = filter(lambda x: x.integral_type() != "interior_facet", integrals)
+        other_integrals = list(filter(lambda x: x.integral_type() != "interior_facet", integrals))
```
wence-: Here you need to coerce to list because `Form` doesn't handle the `iterable` case properly.

miklos1:
No plans yet for preventing Py2isms coming back. `pylint --py3k` can catch though some of the easy problems.

Before I filed this pull request, I had managed to have a Py3 installation, but it didn't work mostly because of the string schism (`str`, `unicode` -> `bytes`, `str`).

firedrake/bcs.py
miklos1: The previous (Py2) behaviour of `map` was to return a `list`, but I'm happy to change this to a `tuple`.

firedrake/slate/slac/kernel_builder.py
miklos1: Sure, if it is only used once, it will work. However, if one adds another usage later, the failure mode is quite unpleasant: you get what you expect at the first iteration, and then you get no elements for any further iteration attempts.

firedrake/bcs.py
wence-: Sure, I was just noting that we'd presumably prefer if the `sub_domain` property were immutable.

wence-:
Assuming that `pylint --py3k` passes on this branch, let's go ahead and enable it in testing.

wence-:
OK, go ahead.

firedrake/slate/slac/kernel_builder.py
thomasgibson: I think I fixed this in #998

PR #991, opened 2017-01-21
Update in the Benney-Luke demo

kalogirou:
I have added a small paragraph in the end of the BL demo to say that we have used the same weak formulations to study a different problem. I have also updated the references by adding our new paper and modifying some of our previous ones.

### DIFF
--- a/demos/demo_references.bib
+++ b/demos/demo_references.bib
@@ -25,6 +25,14 @@
                   {http://www1.maths.leeds.ac.uk/~matak/lms-cup2015.pdf}
 }
 
+@Manual{Gidel:2017,
+  author =       {Gidel, F. and Bokhove, O. and Kalogirou, A.},
+  title =        {Variational modelling of extreme waves through oblique interaction of solitary waves: application to {Mach} reflection},
+  journal =      {Nonlinear Processes in Geophysics},
+  year =         2017,
+  note = {accepted for publication}
+}
+
 @Article{Arnold:2000,
   author =       {Arnold, D. N. and Falk, R. S. and Winther, R.},
   title =        {Multigrid in {H(div)} and {H(curl)}},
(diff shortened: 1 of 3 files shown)


demos/demo_references.bib
```diff
@@ -25,6 +25,14 @@ @InBook{2015:lmscup
                   {http://www1.maths.leeds.ac.uk/~matak/lms-cup2015.pdf}
 }
 
+@Manual{Gidel:2017,
```
dham: Shouldn't this be an article?

docs/source/_static/firedrake-apps.bib
```diff
@@ -1,3 +1,11 @@
+@Manual{Gidel2017,
```
dham: As above.

demos/demo_references.bib
kalogirou: Yes, but we still don't have volume and page information which are essential in the bibtex entry for articles.

dorugeber: Pretty sure that's not true.  Various bibtex guides say [for article] "Required fields: author, title, journal, year. Optional fields: volume, number, pages, month, note."

kalogirou: Ok then I apologise. The ref. Gidel:2017 is an article which is accepted by Nonlinear Processes Geophysics.

docs/source/_static/firedrake-apps.bib
kalogirou: Yes, this should be changed to article.

demos/demo_references.bib
kalogirou: Update: The paper has been published in Volume 24, pages 43-60 (doi:10.5194/npg-24-43-2017). Can you make the change?

docs/source/_static/firedrake-apps.bib
kalogirou: Please include the final publication info, as above. Thanks.

demos/demo_references.bib
dham: Hi Anna, this pull request is from kalogirou:master. Only you have the access to make this change!

wence-:
Thanks!

Issue #496, opened 2015-03-31
Poor serial scaling in simple 2D mesh generation


jrmaddison:
Generation of meshes, via UnitSquareMesh, is sub-linear for meshes larger than around a million nodes.

``` python
from firedrake import *
import numpy
import os
import subprocess
import time

def poll(pid):
  proc_ps = subprocess.Popen(("ps u -p %i" % pid).split(), stdout = subprocess.PIPE)
  proc_ps.wait()
  header, data = proc_ps.stdout.readlines()
  header, data = header.split(), data.split()
  rss = int(data[header.index("RSS")])
  vsz = int(data[header.index("VSZ")])

  return rss, vsz

ln, lt = None, None
for p in xrange(0, 50):
  N = int(2 ** (0.25 * p))
  t0 = time.time()
  mesh = UnitSquareMesh(N, N)
  n, t = mesh.num_vertices(), time.time() - t0
  rss, vsz = poll(os.getpid())
  info("N = %i, t = %.6fs (scaling = %.3f), RSS = %i, VSZ = %i" % (n, t, numpy.NAN if lt is None else numpy.log(t / lt) / numpy.log(n / ln), rss, vsz))
  ln, lt = n, t

```

leads to:

``` bash
pyop2:INFO N = 4, t = 0.004819s (scaling = nan), RSS = 66568, VSZ = 382740
UnitSquareMesh.py:29: RuntimeWarning: divide by zero encountered in double_scalars
  info("N = %i, t = %.6fs (scaling = %.3f), RSS = %i, VSZ = %i" % (n, t, numpy.NAN if lt is None else numpy.log(t / lt) / numpy.log(n / ln), rss, vsz))
pyop2:INFO N = 4, t = 0.001086s (scaling = -inf), RSS = 66596, VSZ = 382740
pyop2:INFO N = 4, t = 0.001083s (scaling = -inf), RSS = 66632, VSZ = 382740
pyop2:INFO N = 4, t = 0.001109s (scaling = inf), RSS = 66676, VSZ = 382740
pyop2:INFO N = 9, t = 0.001184s (scaling = 0.081), RSS = 66724, VSZ = 382740
pyop2:INFO N = 9, t = 0.001207s (scaling = inf), RSS = 66772, VSZ = 382740
pyop2:INFO N = 9, t = 0.001487s (scaling = inf), RSS = 66820, VSZ = 382740
[...]

Issue #505, opened 2015-04-28
UnitCubeMesh(1, 1, 40) causes crash


chromy:
I ran in to a slightly odd bug, the code:

``` python
import firedrake
firedrake.utility_meshes.UnitCubeMesh(1, 1, 40)
```

causes this crash for me:

```
In [3]: firedrake.utility_meshes.UnitCubeMesh(1, 1, 40)
---------------------------------------------------------------------------
Error                                     Traceback (most recent call last)
<ipython-input-3-c5785ca29434> in <module>()
----> 1 firedrake.utility_meshes.UnitCubeMesh(1, 1, 40)

/homes/hjd11/proj/Firedrake/firedrake/firedrake/utility_meshes.pyc in UnitCubeMesh(nx, ny, nz, reorder)
    506     * 6: plane z == 1
    507     """
--> 508     return CubeMesh(nx, ny, nz, 1, reorder=reorder)
    509
    510

/homes/hjd11/proj/Firedrake/firedrake/firedrake/utility_meshes.pyc in CubeMesh(nx, ny, nz, L, reorder)
    486     * 6: plane z == L
    487     """
--> 488     return BoxMesh(nx, ny, nz, L, L, L, reorder=reorder)
    489
    490

/homes/hjd11/proj/Firedrake/firedrake/firedrake/utility_meshes.pyc in BoxMesh(nx, ny, nz, Lx, Ly, Lz, reorder)
    438     boundary.setDimension(2)
    439     boundary.createCubeBoundary([0., 0., 0.], [Lx, Ly, Lz], [nx, ny, nz])
--> 440     plex = PETSc.DMPlex().generate(boundary)
    441
    442     # Apply boundary IDs

/homes/hjd11/.virtualenvs/firedrake/local/lib/python2.7/site-packages/petsc4py/lib/PETSc.so in petsc4py.PETSc.DMPlex.generate (src/petsc4py.PETSc.c:215426)()

Error: error code 4
[0] DMPlexGenerate() line 1080 in /tmp/hjd11/pip-ZiAskM-build/src/dm/impls/plex/plexgenerate.c
[0] DMPlexGenerate_CTetgen() line 834 in /tmp/hjd11/pip-ZiAskM-build/src/dm/impls/plex/plexgenerate.c
[0] TetGenTetrahedralize() line 21576 in /tmp/hjd11/pip-ZiAskM-build/arch-python-linux-x86_64/externalpackages/ctetgen/ctetgen.c
[...]

PR #993, opened 2017-01-30
No ctetgen triangle

wence-:
Baby steps towards supporting 64 bit ints (triangle doesn't support them, so you can't `--download-triangle --with-64-bit-indices` for PETSc).  However, making these regular meshes is not so hard...


wence-:
Also fixes #496, #505.

PR #994, opened 2017-01-30
Notebook updates

wence-:
Link notebooks into webpages.

Testing will come separately.  We will use [nbval](https://github.com/computationalmodelling/nbval), but I am waiting for a PR there to allow testing cells that raise exceptions.


wence-:
Possible bone of contention, we're storing the *executed* notebooks now, so that they can be rendered statically in [nbviewer](https://nbviewer.jupyter.org).  But this makes them quite a bit larger.

wence-:
Comments on wording?

Issue #995, opened 2017-01-30
SEGFAULT on loading exodus mesh with more than one process


juan42:
Executing this code with "mpiexec -n 2 scripy.py"
```
from firedrake import *
mesh = Mesh('disk_out_ref.e')
```
fails with the following error
```
[1]PETSC ERROR: ------------------------------------------------------------------------
[1]PETSC ERROR: Caught signal number 11 SEGV: Segmentation Violation, probably memory access out of range
[1]PETSC ERROR: Try option -start_in_debugger or -on_error_attach_debugger
[1]PETSC ERROR: or see http://www.mcs.anl.gov/petsc/documentation/faq.html#valgrind
[1]PETSC ERROR: or try http://valgrind.org on GNU/linux and Apple Mac OS X to find memory corruption errors
[1]PETSC ERROR: configure using --with-debugging=yes, recompile, link, and run 
[1]PETSC ERROR: to get more information on the crash.
application called MPI_Abort(MPI_COMM_WORLD, 59) - process 1
```
This happens to any type of exodus mesh i have available. The plain C version for DMPlex works fine.
A sample mesh is available [here](https://cloud.tf.uni-kiel.de/index.php/s/GliLS7LNZo1suKz).

wence-:
The boundary marking code needs to be guarded.  Try this patch:

```diff
diff --git a/firedrake/mesh.py b/firedrake/mesh.py
index 0501d5c..a8a2b95 100644
--- a/firedrake/mesh.py
+++ b/firedrake/mesh.py
@@ -203,12 +203,13 @@ def _from_exodus(filename, comm):
     """
     plex = PETSc.DMPlex().createExodusFromFile(filename, comm=comm)
 
-    boundary_ids = plex.getLabelIdIS("Face Sets").getIndices()
-    plex.createLabel("boundary_ids")
-    for bid in boundary_ids:
-        faces = plex.getStratumIS("Face Sets", bid).getIndices()
-        for f in faces:
-            plex.setLabelValue("boundary_ids", f, bid)
+    if plex.hasLabel("Face Sets"):
+        boundary_ids = plex.getLabelIdIS("Face Sets").getIndices()
+        plex.createLabel("boundary_ids")
+        for bid in boundary_ids:
+            faces = plex.getStratumIS("Face Sets", bid).getIndices()
+            for f in faces:
+                plex.setLabelValue("boundary_ids", f, bid)
 
     return plex
 
```

wence-:
[...]

PR #997, opened 2017-01-31
Guard the boundary marking code for exodus files.

juan42:
Loading an exodus mesh file resulted in a memory error in PETSc. This fixes issue #995.


wence-:
Can you provide a small (couple of elements) exodus mesh that we can add to the test suite so we can test this code path?

juan42:
Sure. I will add a test including a small exodus mesh from a recent cubit version.

wence-:
such test data lives in `tests/meshes/` right now.

tests/regression/test_exodus_mesh.py
```diff
@@ -0,0 +1,23 @@
+import pytest
```
thomasgibson: Add `from __future__ import absolute_import, print_function, division` to satisfy the flake8 requirements

juan42:
The tests on travis/drone fail because 
```
E   [0] This method requires ExodusII support. Reconfigure using --download-exodusii
```

tests/regression/test_exodus_mesh.py
```diff
@@ -8,7 +8,10 @@
 
 @pytest.fixture
 def exodus_mesh():
-    return Mesh(join(cwd, "..", "meshes", "brick.e"))
+    try:
+        return Mesh(join(cwd, "..", "meshes", "brick.e"))
+    except PETSc.Error as e:
+        assert e.ierr == 56
```
miklos1: You should rather make an xfail in this case.

wence-:
Fixed the petsc import and merged.

wence-:
Thanks!

Issue #975, opened 2016-12-20
Slate: next steps


thomasgibson:
This issue is for tracking progress on the Slate package:

- [x] Base language in working condition, with some features ready to go (localized kernels for dense linear algebra)

**TODO:**

Organization (pointed out by Lawrence):
- [x] Make `TensorBase` an abstract base class and properly document required subclass methods/properties by annotating with `@abstractmethod`/`@abstractproperty`.
- [x] Address some inconsistencies with attributes: `.tensor` vs `.tensors`

Use Slate in a python preconditioner (inspired by Rob Kirby):
- [x] The framework is there for constructing a preconditioner for RT/BDM mixed systems via hybridization. It's now a matter of designing a `PCBase` (from the matrix free package) preconditioner that does the following:
  - Takes a mixed-system and breaks the velocity space;
  - Introduces the Lagrange multipliers defined on the trace space;
  - Constructs the Slate expression executing element-wise elimination of unknowns to arrive at a system for the multipliers only;
  - Solves the global reduced system for the multipliers;
  - Performs the element-wise reconstruction of the original unknowns in Slate using the multipliers.
    - To make this step more efficient, we need to extend PyOP2 to write into mixed dats.

I believe is a significant step towards addressing #108

Other additions:
- [x] Support extruded cells.
  - Note that we will need to revise how Slate checks mesh layer info once variable column height is implemented in PyOP2
- [x] Introduce a `solve` node in Slate that uses matrix factorizations (provided by Eigen) to solve elemental systems rather than directly inverting the local dense matrices.
- [x] Introduce kernel caching.
- [x] Compose Slate with the matrix-free interface.
- [x] Multiple domains/subdomains 
- [ ] Strong boundary condition support at the Slate level.
- [x] Support tensor blocking.
   - [ ] Entiry indexing: row-sum (or diagonal) mass-lumping.
- [ ] Operator that extracts the diagonal of a matrix.
- [ ] Should `TensorBase` inherit from `gem.node.Node`?

If y'all have other suggestions, feel free to post and/or edit these notes.

PR #998, opened 2017-01-31
Slate extensions: support for extruded problems and various edits

thomasgibson:
This is the second PR which addresses some of the comments detailed in issue #975 

Minor organization changes/updates:
 - It was suggested the base Slate language should use meta-class inheritance and documents some methods. This has now been done. The base language itself remains largely unchanged.

- The were some comments on the use of `.tensor` `.tensors` and `.operands` when accessing components of a particular Slate object. Now `.operands` is the standard way to access the children of a Slate node.

- Documentation has been added/reformatted/edited.

Major additions:

- Slate can now deal with forms defined on both regular/base meshes and extruded meshes. This required some rewriting primarily in the compiler directory `slac`. The changes are:
  - `tsfc_driver.py` was added to provide a standard mechanism for Slate to compile forms associated with terminal `slate.Tensor` objects and transform integral types where appropriate. `ContextKernels` are returned and used in `compiler.py`.
  - `compiler.py` has been rewritten to loop over `ContextKernels` which provide information about the terminal `slate.Tensor`, the original (per-transformed) integral types and all of TSFC `SplitKernel`s for that particular form.
  - Helper functions in `compiler.py` help construct the appropriate code-blocks to evaluate all types of facet integrals: `interior`/`exterior` facets, as well as all forms of extruded facets: `interior_facet_vert`, `interior_facet_horiz`, `exterior_facet_vert`, `exterior_facet_bottom` and `exterior_facet_top`.
  - `kernel_builder.py` has been modified but is functionally equivalent to the previous one. It's fulfilling the same role as before in `compiler.py`

- Tests were added/revised. I relaxed some of the existing tests, as they can be time-consuming.


firedrake/slate/slac/compiler.py
```diff
@@ -36,17 +40,29 @@
...
 
     :arg slate_expr: a :class:'TensorBase' expression.
     :arg tsfc_parameters: an optional `dict` of form compiler parameters to
-                          be passed onto TSFC during the compilation of ufl forms.
+                          be passed onto TSFC during the compilation of
+                          ufl forms.
     """
     if not isinstance(slate_expr, TensorBase):
-        raise ValueError("Expecting a `slate.TensorBase` expression, not a %r" % slate_expr)
+        raise ValueError("Expecting a `TensorBase` expr, not %r" % slate_expr)
```
wence-: I think the plan here was to remember the compiled kernels on the `TensorBase` object, so that if we see it again, we don't have to go through this long symbolic compilation step.

That would look like this:

```python

class TensorBase(...):

    def __init__(self, ...):
          self._kernels = None
   ...

def compile_expression(expr, ...):
    if not isinstance(expr, TensorBase):
        raise ValueError(...)
    if expr._kernels is not None:
        return expr._kernels
    ...

   kernels = (SplitKernel(idx, kinfo),)
   expr._kernels = kernels
   return kernels
```

firedrake/slate/slac/compiler.py
```diff
@@ -144,82 +215,295 @@ def compile_expression(slate_expr, tsfc_parameters=None):
...
+        )
+        index = top.indices
+
+        for cindex in set(bottom.kinfo.coefficient_map
+                          + top.kinfo.coefficient_map):
+            c = exp.coefficients()[cindex]
+            cl.extend(builder.coefficient(c))
+
+        incl.extend(set(bottom.kinfo.kernel._include_dirs +
+                        top.kinfo.kernel._include_dirs))
```
thomasgibson: I'm mainly just being careful about getting all the information here. Or as you put it: belt and braces. I will have a look at `gem.utils` for the `OrderedSet` you mentioned above.

firedrake/slate/slate.py
```diff
@@ -549,38 +590,40 @@ def __init__(self, tensor, coefficient):
             "coefficient function space."
         )
         super(Action, self).__init__()
-        self.tensor = tensor
-        self._acting_coefficient = coefficient
+        self.operands = tensor, coefficient
```
thomasgibson: Should I treat `operands` then to strictly mean the participating Slate objects? Technically the `coefficient` is an operand. I can just add back in the attribute `acting_coefficient` if that's nicer.

wence-: I think so, following the gem model, `operands` are closed over gem objects and "non-child-data" don't live on `operands`.

I prefer `actee` to `acting_coefficient` (since the coefficient isn't acting, the tensor is...)

firedrake/slate/slac/compiler.py
```diff
@@ -290,13 +583,17 @@ def eigen_matrixbase_type(shape):
              library syntax.
     """
     if len(shape) == 0:
-        raise NotImplementedError("Scalar-valued expressions cannot be declared as an Eigen::MatrixBase object.")
+        raise NotImplementedError(
+            "Scalar exprs cannot be declared as an Eigen::MatrixBase object."
```
thomasgibson: I have tried this before, and I couldn't quite get it to work (I did precisely as you said). So at the moment I am raising `NotImplementedError` if one tries to assemble a zero-rank tensor.

firedrake/slate/slac/compiler.py
wence-: You already do the equivalent elsewhere in `slate.py` with:

`tuple(OrderedDict.fromkeys(iterable))`

Note that a `dict` is a set-like object.

firedrake/assemble.py
```diff
@@ -146,9 +150,17 @@ def _assemble(f, tensor=None, bcs=None, form_compiler_parameters=None,
...
+    representing the result. This will be a :class:`float` for 0-forms, a
+    :class:`.Function` for 1-forms/rank-1 Slate tensors and a :class:`.Matrix`
+    for 2-forms/rank-2 Slate tensors.
+
+    .. warning::
+
+    If f is a 0-form wrapped as a Slate tensor, this will fail. 0-form tensors
+    cannot be expressed as an `Eigen::MatrixBase` object. However, rank-0
+    tensors may be used in a Slate expression, so long as the resulting rank
+    of the expression is 1 or 2.
```
wence-: I saw that the deep in the compiler you raise a `NotImplementedError` in the rank-0 case.  I think it's friendlier to someone reading the code if the argument checking happens high up.

firedrake/slate/slac/tsfc_driver.py
```diff
@@ -0,0 +1,98 @@
...
+
+    assert isinstance(tensor, Tensor), (
+        "Only terminal tensors have forms associated with them!"
+    )
+
+    mapper = RemoveRestrictions()
+    integrals = map(partial(map_integrand_dags, mapper),
+                    tensor.form.integrals())
+
+    transformed_integrals = transform_integrals(integrals)
```
thomasgibson: I assert that there can only be positive restrictions (if there are any restrictions at all). This is checked in `slate.py` by the `CheckRestrictions` multifunction. But I agree with revising the cell-facet info in the near future to allow for more general forms

thomasgibson:
Updates have been pushed.

firedrake/slate/slate.py
```diff
@@ -548,39 +592,23 @@ def __init__(self, tensor, coefficient):
             "Argument function space must be the same as the "
             "coefficient function space."
         )
-        super(Action, self).__init__()
-        self.tensor = tensor
-        self._acting_coefficient = coefficient
+        super(Action, self).__init__(tensor,
+                                     assembled_coeffs=(coefficient,))
+        self.actee = coefficient
```
miklos1: You have a double accounting of coefficient. In fact, this class is the only one (unless I missaw something) which actually uses `assembled_coeffs` which suggests that it shouldn't exist on the parent class.

tests/regression/test_slate_assemble_tensors.py
```diff
@@ -4,20 +4,9 @@
...
-        return UnitIntervalMesh(1)
-    elif cell == triangle:
-        return UnitSquareMesh(1, 1)
-    elif cell == tetrahedron:
-        return UnitCubeMesh(1, 1, 1)
-    elif cell == quadrilateral:
-        return UnitSquareMesh(1, 1, quadrilateral=True)
-    else:
-        raise ValueError("%s cell not recognized" % cell)
+    return UnitSquareMesh(2, 2, quadrilateral=request.param)
```
miklos1: Why are you testing fewer cell types?

firedrake/slate/slac/tsfc_driver.py
```diff
@@ -0,0 +1,98 @@
...
+    cxt_kernels = []
+    counter = 0
+    for orig_it_type, integrals in transformed_integrals.items():
+        prefix = "subkernel%d_%s_%s_to_" % (counter,
+                                            tensor.__str__(),
```
thomasgibson: `test_slac.py` seems to disagree. But perhaps it's not thorough enough.

tests/regression/test_slate_assemble_tensors.py
thomasgibson: Short answer: it was a time-consuming and a long-winded test. I can revert this if the time it takes to run tests isn't a factor for us.

miklos1: Fine, if it takes much time without much benefit, then you may reduce it.

firedrake/slate/slate.py
```diff
@@ -589,41 +592,23 @@ def __init__(self, tensor, coefficient):
             "Argument function space must be the same as the "
             "coefficient function space."
         )
-        super(Action, self).__init__()
-        self.operands = tensor, coefficient
+        super(Action, self).__init__(tensor,
+                                     assembled_coeffs=(coefficient,))
```
wence-: I think the superclass shouldn't know about this extra coefficient junk.  But rather than you should override `coefficients` here:

```python
def coefficients(self):
    return tuple(OrderedDict.fromkeys(super(Action, self).coefficients() + (self.actee, )))
```

Similarly with `ufl_domains`.  If you're worried about code reuse in `join_domains` and similar, make free functions:

`merge_coefficients` and `join_domains` and just call them with the appropriate arguments in the instance methods.

firedrake/slate/slac/compiler.py
```diff
@@ -559,7 +582,8 @@ def metaphrase_slate_to_cpp(expr, temps, prec=None):
         return parenthesize(result, expr.prec, prec)
 
     elif isinstance(expr, Action):
-        tensor, c = expr.operands
+        tensor, = expr.operands
+        c = expr.actee
         # Extra check
         assert isinstance(c, Coefficient)
```
wence-: You already assert that `actee` is a `Function` in the `Action` contructor.  Probably no need to do so again here.

firedrake/slate/slac/tsfc_driver.py
miklos1: Right, it will give the same result if you compile the same slate expression object, but it will generate  a different code if you reconstruct the expression provided that some of the `Tensor` form wrapper leaf nodes are reconstructed as well. Perhaps this is good enough though.

wence-: Hmm, `transformed_integrals` is a `defaultdict(list)`, so the iteration order is not necessarily consistent.  You could make it an `OrderedDict` at the cost of a little more code in `transform_integrals`:

```python

def transform_integrals(...):
   transformed = collections.OrderedDict()

   for it in integrals:
       trans_itgs = transformed.setdefault(it.integral_type(), list())
       ... append to trans_itgs.
```

thomasgibson: @miklos1 The non-idempotent bit of the code-generation if you reconstruct is that `tensor.__str__()` will change since its `id` will be updated. Do you think this will be an issue?

miklos1: Probably not. I think it can stay like this for now.

firedrake/slate/slate.py
```diff
@@ -612,3 +611,39 @@ def __repr__(self):
...
+    :arg assembled_coeffs: a `tuple` of assembled data provided
+                           as a `firedrake.Function`.
+    """
+    coefficient_list = []
+    extra_data = assembled_coeffs or ()
+    for op in operands:
+        coefficient_list.extend(op.coefficients())
+
+    return tuple(OrderedDict.fromkeys(tuple(coefficient_list) +
+                                      extra_data))
```
miklos1: I think you can do this with less code.

`assembled_coeffs=None` version:
```python
return tuple(OrderedDict.fromkeys(chain(*[op.coefficients() for op in operands])))
```

When `assembled_coefficient` is specified:
```python
return tuple(OrderedDict.fromkeys(chain(*([op.coefficients() for op in operands] + [assembled_coefficient]))))
```
or if you don't want such a long line:
```python
coefficientz = [op.coefficients() for op in operands] + [assembled_coefficient]
return tuple(OrderedDict.fromkeys(chain(*coefficientz)))
```

firedrake/slate/slate.py
```diff
@@ -612,3 +611,39 @@ def __repr__(self):
...
+    """Returns a set of non-repeated ufl_domain objects.
+
+    :arg operands: a `tuple` of Slate tensors.
+    :arg assembled_coeffs: a `tuple` of assembled data provided
+                           as a `firedrake.Function`.
+    """
+    collected_domains = []
+    extra_data = assembled_coeffs or ()
+    for obj in operands + extra_data:
+        collected_domains.extend(obj.ufl_domains())
```
miklos1: You can probably do roughly the same here.

firedrake/slate/slate.py
```diff
@@ -625,13 +627,12 @@ def merge_coefficients(operands, assembled_coeffs=None):
...
                            as a `firedrake.Function`.
     """
-    coefficient_list = []
-    extra_data = assembled_coeffs or ()
-    for op in operands:
-        coefficient_list.extend(op.coefficients())
+    if assembled_coeffs is None:
+        coeffs = [op.coefficients() for op in operands]
+    else:
+        coeffs = [op.coefficients() for op in operands] + [assembled_coeffs]
```
miklos1: I think you could just inline this function now, if I counted correctly, it is only called once with `assembled_coeffs` and once without it.

firedrake/slate/slate.py
```diff
@@ -641,9 +642,10 @@ def merge_domains(operands, assembled_coeffs=None):
...
     """
-    collected_domains = []
-    extra_data = assembled_coeffs or ()
-    for obj in operands + extra_data:
-        collected_domains.extend(obj.ufl_domains())
+    if assembled_coeffs is None:
+        collected_domains = [op.ufl_domain() for op in operands]
+    else:
+        collected_domains = [obj.ufl_domain() for obj in operands +
+                             assembled_coeffs]
```
miklos1: Same here.

thomasgibson:
Okay, now I believe it's ready for another round of checks.

thomasgibson:
Conflicts resolved from recent merge.

firedrake/slate/slac/compiler.py
thomasgibson: @wence- I finally solved this problem. I still do the `Eigen::MatrixBase<double, 1, 1>` hack, but @miklos1 pointed out some useful information I needed to know at the PyOP2 level (specifically for assembling 0-forms). Next commit will add zero-form tests in Slate and should be ready for review.

miklos1:
Maybe you can now reconsider whether you need a 1x1 Eigen matrix or you 
could use an Eigen scalar instead (if such a thing exists).

thomasgibson:
I would if I could find any info on it. Currently I find this: https://eigen.tuxfamily.org/dox/TopicScalarTypes.html ...

firedrake/slate/slac/kernel_builder.py
```diff
@@ -106,102 +186,50 @@ def prepare_coefficients(expression):
...
         aux_exprs = []
-    else:
-        aux_exprs = aux_exprs
 
-    if isinstance(expression, Tensor):
-        if expression not in temps.keys():
-            temps[expression] = ast.Symbol("T%d" % len(temps))
+    if isinstance(expr, Tensor):
+        if expr not in temps:
+            temps[expr] = ast.Symbol("T%d" % len(temps))
```
wence-: AKA `temps.setdefault(expr, ast.Symbol(...))`

firedrake/slate/slac/kernel_builder.py
```diff
@@ -106,102 +186,50 @@ def prepare_coefficients(expression):
...
+        aux_exprs.append(expr)
         # Pass in the acting tensor to extract any necessary temporaries
-        prepare_temps_and_aux_exprs(expression.tensor, temps=temps, aux_exprs=aux_exprs)
-
-    elif isinstance(expression, (UnaryOp, BinaryOp)):
-        for operand in expression.operands:
-            prepare_temps_and_aux_exprs(operand, temps=temps, aux_exprs=aux_exprs)
+        generate_expr_data(expr.operands[0], temps=temps, aux_exprs=aux_exprs)
 
+    elif isinstance(expr, (UnaryOp, BinaryOp)):
```
wence-: This could have just been `isinstance(expr, TensorOp)` now, I think.  But doesn't matter.

firedrake/slate/slate.py
```diff
@@ -240,62 +286,84 @@ def _output_string(self, prec=None):
...
+        coeffs = [op.coefficients() for op in self.operands]
+        return tuple(OrderedDict.fromkeys(chain(*coeffs)))
 
     def ufl_domains(self):
         """Returns the integration domains of the integrals associated with
         the tensor.
         """
-        return self.tensor.ufl_domains()
+        collected_domains = [op.ufl_domain() for op in self.operands]
+        return join_domains(collected_domains)
```
wence-: Does this want to be `join_domains(itertools.chain(*[op.ufl_domains() for op in self.operands]))` that DTRT if the operands have multiple domains, no?

firedrake/slate/slac/compiler.py
```diff
@@ -37,94 +41,169 @@
...
+            inc.extend(incl)
+            statements.append(stmt)
+
+        elif it_type in ["exterior_facet_bottom", "exterior_facet_top"]:
+            # These kernels will only be called if we are on
+            # the top or bottom layers of the extruded mesh.
+            builder.require_mesh_layers()
+            stmt, incl = extruded_top_bottom_facet(cxt_kernel, builder,
+                                                   coordsym, mesh_layer_sym)
+            inc.extend(incl)
```
thomasgibson: Ah yes, I see. Perhaps I should use:
```
dirs = list(set(incl) - set(inc))
inc.extend(dirs)
```
or something similar.

firedrake/slate/slac/tsfc_driver.py
```diff
@@ -0,0 +1,102 @@
...
+    # the tensor name for code idempotency reasons, but is not
+    # strictly required.
+    prefix = prefix or "subkernel%s_" % tensor.__str__()
+    mapper = RemoveRestrictions()
+    integrals = map(partial(map_integrand_dags, mapper),
+                    tensor.form.integrals())
+
+    transformed_integrals = transform_integrals(integrals)
+    cxt_kernels = []
+    for orig_it_type, integrals in transformed_integrals.items():
```
thomasgibson: Good call. Will update

firedrake/slate/slac/compiler.py
```diff
@@ -135,92 +214,300 @@ def compile_expression(slate_expr, tsfc_parameters=None):
...
+        bottom_calls.append(ast.FunCall(btm.kinfo.kernel.name,
+                                        tensor, coordsym, *clist))
+
+    else_stmt = ast.Block(top_calls + bottom_calls, open_scope=True)
+    inter_stmt = ast.If(ast.Eq(mesh_layer_sym, nlayers - 1),
+                        (ast.Block(bottom_calls, open_scope=True),
+                         else_stmt))
+    stmt = ast.If(ast.Eq(mesh_layer_sym, 0),
+                  (ast.Block(top_calls, open_scope=True),
+                   inter_stmt))
```
thomasgibson: I'll make a note of this in #975

firedrake/slate/slate.py
```diff
@@ -317,8 +315,8 @@ def ufl_domains(self):
         """Returns the integration domains of the integrals associated with
         the tensor.
         """
-        collected_domains = [op.ufl_domain() for op in self.operands]
-        return join_domains(collected_domains)
+        collected_domains = [op.ufl_domains() for op in self.operands]
+        return tuple(OrderedDict.fromkeys(chain(*collected_domains)))
```
miklos1: Is `join_domains` the same as this?

thomasgibson: After looking in `ufl.domain`, I believe that they do the same thing...but now I realize that they are doing one additional check: `join_domains` also checks that domains with the same id are compatible. Probably should revert this back to `join_domains`

firedrake/slate/base.py
```diff
@@ -0,0 +1,116 @@
...
+        return not self.__eq__(other)
+
+    @cached_property
+    def _hash_id(self):
+        """Returns a hash id."""
+        return hash(self._key)
+
+    def __hash__(self):
+        """Generates a hash for the TensorBase object."""
+        return self._hash_id
```
miklos1: I'm not convinced that there are sufficient grounds for splitting the base class. Although this base class does not define the expression operators, it is still sufficiently Slate specific.

C.f. `gem.node.Node` is used twice: for GEM (`gem.gem.Node`) and for the Impero AST (`gem.impero.Node`).

PR #1005, opened 2017-02-08
Jenkins

dham:
This pull request introduces basic Jenkins testing on Mac.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -4,6 +4,7 @@ pipeline {
     TEST_FILES = "tests/extrusion/test_facet_integrals_2D.py tests/extrusion/test_mixed_bcs.py tests/extrusion/test_steady_advection_2D_extr.py tests/multigrid/test_two_poisson_gmg.py tests/output tests/regression/test_facet_orientation.py tests/regression/test_matrix_free.py tests/regression/test_nested_fieldsplit_solves.py tests/regression/test_nullspace.py tests/regression/test_point_eval_api.py tests/regression/test_point_eval_cells.py tests/regression/test_point_eval_fs.py tests/regression/test_solving_interface.py tests/regression/test_steady_advection_2D.py"
     PATH = "/usr/local/bin:/usr/bin:/bin"
     CC = "mpicc"
+    FIREDRAKE_CI_TESTS = "1"
   }
   stages {
     stage('Clean') {
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -1,0 +1,42 @@
+pipeline {
+  agent any
+  environment {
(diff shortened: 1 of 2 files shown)


Jenkinsfile
```diff
@@ -0,0 +1,54 @@
+pipeline {
+  agent any
+  environment {
+    TEST_FILES = "tests/extrusion/test_facet_integrals_2D.py tests/extrusion/test_mixed_bcs.py tests/extrusion/test_steady_advection_2D_extr.py tests/multigrid/test_two_poisson_gmg.py tests/output tests/regression/test_facet_orientation.py tests/regression/test_matrix_free.py tests/regression/test_nested_fieldsplit_solves.py tests/regression/test_nullspace.py tests/regression/test_point_eval_api.py tests/regression/test_point_eval_cells.py tests/regression/test_point_eval_fs.py tests/regression/test_solving_interface.py tests/regression/test_steady_advection_2D.py"
```
wence-: Note for future, do full testing when we have more build hardware.

Jenkinsfile
```diff
@@ -0,0 +1,54 @@
...
+        }
+      }
+    }
+    stage('Build') {
+      steps {
+        sh 'mkdir tmp'
+        dir('tmp') {
+          timestamps {
+            sh 'pip install virtualenv'
+            sh '../scripts/firedrake-install --disable-ssh --minimal-petsc ${SLEPC} --adjoint --slope --install thetis --install gusto ${PACKAGE_MANAGER}'
```
wence-: This installs firedrake master (unless some magic has happened that I don't understand) I think.

Jenkinsfile
```diff
@@ -0,0 +1,54 @@
...
+      }
+    }
+    stage('Test'){
+      steps {
+        dir('tmp') {
+          timestamps {
+            sh """
+. ./firedrake/bin/activate
+pip install pytest-cov pytest-xdist
+cd firedrake/src/firedrake; py.test --cov firedrake --short -v ${TEST_FILES}
```
wence-: Could add the coverage upload (I think this page says how: https://docs.codecov.io/docs/supported-ci-providers)

Jenkinsfile
dham: You're right. I squashed the magic which does this right. I'll fix it in the morning.

scripts/firedrake-install
```diff
@@ -1159,23 +1160,26 @@ if mode == "install":
...
-            raise NotImplementedError("Unhandled CI type")
-        os.chdir("firedrake")
-        if pull_request != "false":
-            check_call(["git", "fetch", "origin", "pull/" + pull_request + "/merge"])
-            check_call(["git", "checkout", "FETCH_HEAD"])
-        else:
-            check_call(["git", "checkout", commit])
-        os.chdir("..")
+    if jenkins:
+        check_call(["ln", "-s", "../../../", "firedrake"])
```
wence-: Can you also guard this with the same logic as used for drone? i.e. ensure that we are testing firedrake itself.  Otherwise I think this DTRT if using jenkins to test (say) gusto/thetis.

scripts/firedrake-install
```diff
@@ -1159,23 +1160,26 @@ if mode == "install":
...
-            raise NotImplementedError("Unhandled CI type")
-        os.chdir("firedrake")
-        if pull_request != "false":
-            check_call(["git", "fetch", "origin", "pull/" + pull_request + "/merge"])
-            check_call(["git", "checkout", "FETCH_HEAD"])
-        else:
-            check_call(["git", "checkout", commit])
-        os.chdir("..")
+    if jenkins and ci_testing_firedrake:
+        check_call(["ln", "-s", "../../../", "firedrake"])
```
dham: @wence: Pending the last build test, I claim this is done. (the earlier fail appears to have been a drone crash. Rerunning now)

PR #1007, opened 2017-02-11
Use FInAT elements for node allocation

### LINKED ISSUE #967: Use FInAT elements in function spaces
For the allocation of nodes / degrees of freedom. Requires FInAT/FInAT#20.

miklos1:
Resolves #967. Requires FInAT/FInAT#23. FIAT elements are still used when the dual is required.

### DIFF
--- a/firedrake/slate/slac/compiler.py
+++ b/firedrake/slate/slac/compiler.py
@@ -479,7 +479,7 @@ def auxiliary_information(builder):
             # compute its shape to declare as an Eigen::MatrixBase object
             temp = ast.Symbol("C%d" % i)
             V = actee.function_space()
-            node_extent = V.fiat_element.space_dimension()
+            node_extent = V.finat_element.space_dimension()
             dof_extent = np.prod(V.ufl_element().value_shape())
             typ = eigen_matrixbase_type(shape=(dof_extent * node_extent,))
             aux_statements.append(ast.Decl(typ, temp))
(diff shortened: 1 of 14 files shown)


miklos1:
The branch test on Drone fails `tests_firedrake/poisson_bc/test_poisson_bc.py::test_poisson_bc` because the convergence is slightly below the expected. The PR test passed. Is this a known problem?

thomasgibson:
Yes, this has happened to me a few times. It happens a bit sporadically.

firedrake/slate/slac/compiler.py
```diff
@@ -479,7 +479,7 @@ def auxiliary_information(builder):
             # compute its shape to declare as an Eigen::MatrixBase object
             temp = ast.Symbol("C%d" % i)
             V = actee.function_space()
-            node_extent = V.fiat_element.space_dimension()
+            node_extent = V.finat_element.space_dimension()
             dof_extent = np.prod(V.ufl_element().value_shape())
```
thomasgibson: We noted earlier that is calculation is wrong and is fixed in #1010

wence-:
The failure is because that test is actually random (it makes a random perturbation direction).  We should just relax the tolerance somewhat.

firedrake/slate/slac/compiler.py
miklos1: The point here is to use the FInAT element rather than the FIAT element.

miklos1:
I guess this should be ready to merge then (subject to review).

wence-:
I think this is fine.

PR #1008, opened 2017-02-11
Parallel code-generation and cell orientations in Slate

thomasgibson:
This PR fixes a bug in the code-generation bit of Slate when running problems in parallel. In addition, I also added support for problems that require `cell_orientations` (such as hybridizing on a spherical domain).

### DIFF
--- a/firedrake/slate/slac/compiler.py
+++ b/firedrake/slate/slac/compiler.py
@@ -103,13 +103,16 @@ def compile_expression(slate_expr, tsfc_parameters=None):
     mesh_layer_sym = ast.Symbol("layer")
     inc = []
 
+    declared_temps = set()
     for cxt_kernel in builder.context_kernels:
         exp = cxt_kernel.tensor
         t = builder.get_temporary(exp)
 
-        # Declare and initialize the temporary
-        statements.append(ast.Decl(eigen_matrixbase_type(exp.shape), t))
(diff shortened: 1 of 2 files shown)


thomasgibson:
Closing this PR. These small fixes will be a part of the PR featuring the hybridization preconditioner and eigen factorizations.

wence-:
It's easier for us to review small changes.  Do these changes not make sense on their own (i.e. not in the context of factorisations and PC)?

thomasgibson:
The `cell_orientations` update was a tiny addition to the compiler and only needed a few lines here and there. The parallel fix requires being very careful about iterating over `dict` objects, but was corrected by using `OrderedDict()` instances. I will go ahead and reopen this and propose these changes separately. Sorry!

thomasgibson:
After further review, I think this PR can be closed and replaced with a PR proposing a merge of the branch https://github.com/firedrakeproject/firedrake/tree/aux-temps into master. I propose this because it modifies how temporaries are declared while also addressing the parallel bug (which is related to the temporaries). Cell orientations can be added to the branch as it is a trivial addition (or it can just be pulled out of this PR). The new branch allocates temporaries for inverses that are reused (determined by reference count). Is this a reasonable suggestion or should we stick with this PR?

wence-:
OK, opened as #1010.

PR #1009, opened 2017-02-14
Int64

wence-:
Allows installing firedrake using petsc with 64 bit indices.

`firedrake-install --petsc-int-type int64` selects.

`firedrake-update --petsc-int-type int32` changes the int type and forces a rebuild.

Updating retaining the int type doesn't rebuild.

Requires OP2/PyOP2#521.


.drone.yml
```diff
@@ -2,6 +2,11 @@ workspace:
   base: /drone
   path: src
 
+matrix:
+  PETSC_INT_TYPE:
+    - int32
+    - int64
```
wence-: Ensures that drone tests with both int types.

firedrake/dmplex.pyx
```diff
@@ -1936,11 +1942,18 @@ def exchange_cell_orientations(
...
         PETSc.Section new_section
-        np.int32_t *new_values = NULL
+        PetscInt *new_values = NULL
         PetscInt i, c, cStart, cEnd, l, r
 
+    try:
+        try:
+            dtype = MPI.__TypeDict__[np.dtype(IntType).char]
+        except AttributeError:
+            dtype = MPI._typedict[np.dtype(IntType).char]
```
miklos1: So when is which API correct?

scripts/firedrake-install
```diff
@@ -316,9 +332,15 @@ if options["minimal_petsc"]:
     # If you change these minimal PETSc options, you will have to
     # manually invalidate the Travis cache of PETSc and petsc4py.  See
     # https://docs.travis-ci.com/user/caching/#Clearing-Caches.
-    petsc_opts = """--download-chaco --download-hdf5 --download-hypre --download-eigen"""
+    if options["petsc_int_type"] == "int64":
+        petsc_opts = """--download-metis --download-parmetis --download-hdf5 --download-hypre --download-eigen --with-64-bit-indices"""
```
miklos1: So with 64 bits we don't need Chaco but we need METIS and ParMETIS?

scripts/firedrake-install
```diff
@@ -316,9 +332,15 @@ if options["minimal_petsc"]:
...
+    else:
+        petsc_opts = """--download-chaco --download-hdf5 --download-hypre --download-eigen"""
 else:
-    petsc_opts = """--download-chaco --download-metis --download-parmetis --download-scalapack --download-hypre --download-mumps --download-netcdf --download-hdf5 --download-exodusii --download-eigen"""
+    if options["petsc_int_type"] == "int64":
+        petsc_opts = """--download-metis --download-parmetis --download-hypre --download-netcdf --download-hdf5 --download-exodusii --download-eigen --with-64-bit-indices"""
+    else:
+        petsc_opts = """--download-chaco --download-metis --download-parmetis --download-scalapack --download-hypre --download-mumps --download-netcdf --download-hdf5 --download-exodusii --download-eigen"""
```
miklos1: Do Chaco, ScaLAPACK and MUMPS not work with 64 bits?

firedrake/dmplex.pyx
wence-: A change between `mpi4py` version 1.3.1 and 2.0 (the next released version).

scripts/firedrake-install
wence-: Correct.  Chaco is only designed for 32bit int type.  If you think about it, why would you need 64bit ints for a direct solver?

PR #1010, opened 2017-02-15
Add auxiliary temporaries for tensors

wence-:
Supersedes #1008 

This PR introduces auxiliary temporaries for inverses (depending on how often they appear in the DAG). It also corrects a small bug that occurs when running Slate in parallel. This was corrected by using `OrderedDict` objects to hold temporary information.

In addition, a few lines were added to support `cell_orientations` arguments when necessary.


firedrake/slate/slate.py
```diff
@@ -185,6 +186,15 @@ def __eq__(self, other):
     def __ne__(self, other):
         return not self.__eq__(other)
 
+    @abstractproperty
+    def _complexity(self):
+        """Returns an integer totaling the relevant complexity
+        of the tensor expression. This is done by summing the
+        complexity of each tensor in the expression. This
+        value is a function of operator precedence if the
+        expression is not a terminal tensor.
```
wence-: Hmmm, this is why I am concerned by this code smell.  Why does operator precedence (say) have anything to do with expression complexity?

tests/regression/test_slate_linear_algebra.py
```diff
@@ -79,6 +79,26 @@ def test_add_the_negative(mesh):
...
+    v = TestFunction(V)
+
+    A = Tensor(u*v*dx)
+    B = Tensor(2.0*u*v*dx)
+
+    # This is a very silly way to write the vector of ones
+    foo = (B.T*A.inv).T*g + (-A.inv.T*B.T).inv*f + B.inv*(A.T).T*f
+    assert np.allclose(assemble(foo).dat.data, np.ones(V.node_count))
+
+
```
wence-: Can you add test of the cell_orientations functionality too?  That code path is currently not hit by the test suite.

firedrake/slate/slac/kernel_builder.py
```diff
@@ -36,12 +40,25 @@ def __init__(self, expression, tsfc_parameters=None):
...
+
+        # Collect the reference count of operands in auxiliary expressions
+        self._ref_counts = collect_reference_count(expr_dags)
+        # Since the most complicated expressions get caught first, we
+        # reorder by expression complexity to create temporaries for the
+        # inner-most instances first (using reference count later to decide
+        # if it's worth the extra memory)
+        self.aux_temps = OrderedDict.fromkeys(sorted(aux_exprs,
+                                                     key=lambda x:
+                                                     x._complexity))
```
thomasgibson: I wanted to start inside and work my way outside to avoid running into a case where I need to generate an expression for a temp and there are unknowns inside that haven't been declared yet. However, I just removed this and simply used `reversed` (I work my way down an expression by starting from top and move to the roots). Seems to work. Will amend.

firedrake/slate/slac/compiler.py
```diff
@@ -452,67 +479,86 @@ def facet_integral_loop(cxt_kernel, builder, coordsym, cellfacetsym):
...
+                aux_statements.append(ast.Assign(temp, result))
+
+                # Update declared temps
+                declared_temps[exp] = temp
+
+        elif isinstance(exp, Action):
+            # Action computations are relatively inexpensive, so
+            # we don't waste memory space on creating temps for
+            # these expressions. However, we must create a temporary
+            # for the actee coefficient (if we haven't already).
```
thomasgibson: Yes. Do you think I should create temps for actions? I did previously but decided against it. It's a fairly straightforward thing to add back in.

miklos1:
@thomasgibson: Can you give a proper title to this PR? (Not sure if you can edit the description as well.)

firedrake/slate/slac/kernel_builder.py
```diff
@@ -199,35 +202,23 @@ def generate_expr_data(expr, temps=None, aux_exprs=None):
...
+    for tensor in traverse_dags([expr]):
+        if isinstance(tensor, Tensor):
+            temps.setdefault(tensor, ast.Symbol("T%d" % len(temps)))
+
+        elif isinstance(tensor, TensorOp):
+            # For Action, we need to declare a temporary later on for the
+            # acting coefficient. For inverses, we may declare additional
+            # temporaries (depending on reference count).
+            if isinstance(tensor, (Action, Inverse)):
+                aux_exprs.setdefault(tensor)
```
miklos1: What is this doing?

firedrake/slate/slate.py
```diff
@@ -61,6 +61,7 @@ class TensorBase(with_metaclass(ABCMeta)):
     def __init__(self):
         """Constructor for the TensorBase abstract class."""
         self._kernels = None
+        self.operands = ()
```
miklos1: You could make this a class attribute.

firedrake/slate/slac/utils.py
```diff
@@ -133,3 +135,43 @@ def visit_Symbol(self, o, *args, **kwargs):
...
+    """Returns a mapping from operand to reference count.
+
+    :arg expr_dags: an iterable of Slate nodes that make up the
+                    DAG of a given Slate expression.
+    """
+    result = collections.Counter(expr_dags)
+    for tensor in traverse_dags(expr_dags):
+        result.update(tensor.operands)
+
+    return result
```
miklos1: It is probably beyond the scope of this pull request, so I'm just telling for later consideration: if the slate root class (`TensorBase`) were derived from `gem.node.Node` (not to be confused with `gem.gem.Node`) and followed the same conventions, then you could just use the corresponding utility functions from `gem.node` rather than having to implement them again.

I suggested this at the beginning, but it was dismissed then for the difficulty of inheriting both `ufl.form.Form` and `gem.node.Node`. However, slate tensors have not been derived classes of `ufl.form.Form` for quite long, so this objection does not exist anymore.

firedrake/slate/slate.py
```diff
@@ -567,7 +568,7 @@ class Action(TensorOp):
     :arg function: a :class:`firedrake.Function` object.
     """
 
-    prec = 2
+    prec = 3
```
miklos1: I probably missed something earlier, but what are these precedences actually used for?

firedrake/slate/slac/compiler.py
```diff
@@ -452,67 +479,86 @@ def facet_integral_loop(cxt_kernel, builder, coordsym, cellfacetsym):
...
             actee, = exp.actee
+            if actee not in declared_temps:
+                # Declare a temporary for the coefficient
+                ctemp = ast.Symbol("auxT%d" % len(declared_temps))
+                V = actee.function_space()
+                node_extent = V.fiat_element.space_dimension()
+                dof_extent = np.prod(V.ufl_element().value_shape())
+                typ = eigen_matrixbase_type(shape=(dof_extent * node_extent,))
```
miklos1: OK, we discussed this is wrong.

firedrake/slate/slac/kernel_builder.py
thomasgibson: It's generating an ordered, set-like object of expressions that need extra reasoning. `Action`s require creating temporaries for `firedrake.Function`s and `Inverse` may need a temporaries if it's reused multiple times.

firedrake/slate/slate.py
thomasgibson: Pretty-printing and parenthesizing an expression when the C++ code-snippet is generated

firedrake/slate/slac/utils.py
thomasgibson: I will add this as a TODO/Have a think about it.

firedrake/slate/slac/kernel_builder.py
miklos1: > It's generating an ordered, set-like object of expressions that need extra reasoning.

So the oddity is that here you call `setdefault` without actually giving a default value. A few lines above you do specify a value.

> Actions require creating temporaries for firedrake.Functions

I guess you do this because actees are non-child data of an `Action`. Still it smells a bit like shoving different things into the same box.

> and Inverse may need a temporaries if it's reused multiple times.

Doesn't this rule apply to all expressions?

firedrake/slate/slate.py
miklos1: I see you increment most of these precedence values. It scales better if you use the Martin/UFL approach (see `ufl.precedence`).

miklos1: Wait, aren't you generating a COFFEE AST? Then COFFEE should handle the parenthesising... (?)

wence-: All of the eigen tensor expressions are created as a `FlatBlock` (since they're method calls).  Hence they want parenthesising here.

firedrake/slate/slac/kernel_builder.py
thomasgibson: > So the oddity is that here you call `setdefault` without actually giving a default value. A few lines above you do specify a value.

Right, it is a bit weird. However I really do need to preserve ordering. And I don't want to double count an expression (hence set-like). I put nothing as a default value because it hasn't been decided at this point whether a temporary (`ast.Symbol`) should be created. Should I add a symbol here anyways (even if I never actually use it)?

> I guess you do this because actees are non-child data of an Action. Still it smells a bit like shoving different things into the same box.

Yeah I can see that. I cannot avoid creating a temp for a coefficient though. And I don't create the same coefficient temp if it has already been declared. Is the additional O(n) storage worth it, you think?

> Doesn't this rule apply to all expressions?

It may, but wouldn't it bloat the memory space? I chose inverses in particular because they are much more expensive and storing it may be worthwhile. In the future, I will do some profiling and optimizing the cpp code. It should be a relatively easy to extend this temp declaration system to other expressions.

thomasgibson:
Edits have been pushed. I believe it's ready now for another review. Since the Slate tests have been adding up, I moved them to a separate test directory.

firedrake/slate/slac/kernel_builder.py
```diff
@@ -219,6 +220,6 @@ def generate_expr_data(expr):
             # acting coefficient. For inverses, we may declare additional
             # temporaries (depending on reference count).
             if isinstance(tensor, (Action, Inverse)):
-                aux_exprs.setdefault(tensor)
+                aux_exprs.append(tensor)
```
miklos1: If you aren't sure that there will be no duplications, you may want to add an `aux_exprs = list(OrderedDict.fromkeys(aux_exprs))` line after the for loop.

firedrake/slate/slate.py
```diff
@@ -60,8 +60,9 @@ class TensorBase(with_metaclass(ABCMeta)):
 
     def __init__(self):
         """Constructor for the TensorBase abstract class."""
+        # NOTE: This attribute is for caching kernels after
+        # an expression has been compiled.
         self._kernels = None
```
miklos1: Perhaps if you insert a blank line after this attribute that helps as a visual guide to the reader that this comment applies only to the immediately following line, and not for the any further lines in the same block. Alternatively, if you can get away with a shorter comment, then you could have it inline, in that case no blank lines needed as visual guides.

But probably it's even better if you just give this attribute a clearer name, e.g. `_kernel_cache` or even `_metakernel_cache`, maybe `_metakernels` or something similar.

firedrake/slate/slate.py
```diff
@@ -620,3 +607,21 @@ def __repr__(self):
...
+    [UnaryOp],
+    [Add, Sub],
+    [Mul, Action]
+]
+
+# Here we establish the precedence class attribute for a given
+# Slate tensor. Unary and binary operations will have higher
+# precedence over terminal Tensor objects. Similarly, tensor
+# action and multiplication will have higher precedence over
+# addition/subtraction
```
miklos1: What does it mean for a terminal to have a precedence? Terminals are not operations, they are atomic, so they never need surrounding parenthesis.

miklos1: The first sentence of this comment is fine, but the rest is at best misplaced. The following code does not authoritatively define precedence, it just assigns the levels to the classes. The definition happens above. Moreover, one must always be cautious when wanting to repeat in comment exactly what the code does: it often does not help much, but there is danger that when the code is changed in the future, someone will forget to update the comments as well.

thomasgibson: Well they are given `precedence = 0`, so maybe `None` (or just omit completely) is better. And terminal tensors never get parenthesized.

miklos1: If you want/need to assign precedence levels to them, shouldn't terminals appear to be the most strongly binding? That is, they never need parenthesis around them, because they "bind" more strongly anyway than their surrounding.

Also please be aware that Python3 is stricter when comparing different types. `1 < None` is fine in Python2, but is an error in Python3.

wence-: Note that the compiler never asks for the precedence of a terminal (as expected, since there is no need).  Hence better to not even put a precedence property on terminals?

thomasgibson: Right, I just realized that after checking again. Probably best to just omit `Tensor`

miklos1: If it's not used, then better just skip terminals and stick to operations when defining precedence levels.

thomasgibson:
Merge conflict resolved. Should be ready (subject to further review)

PR #1011, opened 2017-02-15
Remove xcode check from firedrake-install

dham:
Using a brand new, clean, Mac Mini I have just confirmed @francispoulin's previous observation that we don't actually need XCode -  the command line tools suffice.

### DIFF
--- a/scripts/firedrake-install
+++ b/scripts/firedrake-install
@@ -955,10 +955,6 @@ if travis:
 osname = platform.uname()[0]
 if osname == "Darwin":
     if options["package_manager"]:
-        try:
-            check_call(["xcodebuild", "-version"])
-        except:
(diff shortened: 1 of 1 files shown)


scripts/firedrake-install
```diff
@@ -955,10 +955,6 @@ if travis:
 osname = platform.uname()[0]
 if osname == "Darwin":
     if options["package_manager"]:
-        try:
-            check_call(["xcodebuild", "-version"])
-        except:
-            quit("Xcode not found. Please install xcode from the App Store and try again.")
```
miklos1: Don't you want to check that the command line tools are installed?

dham: We do that already on immediate next lines.

miklos1: Ah, I see, OK.

PR #1019, opened 2017-02-22
Fix firedrake-install for Python 3

miklos1:
This patch lets Firedrake correctly install with Python 3. The installation, of course, will not work, because firedrake is not Python 3 compatible.

Additionally, there are several small Python 3 compatibility fixes in this patch.


scripts/firedrake-install
```diff
@@ -343,7 +343,13 @@ def check_call(arguments, env=None):
...
 def check_output(args, env=None):
     try:
-        return subprocess.check_output(args, stderr=subprocess.STDOUT, env=env)
+        result = subprocess.check_output(args, stderr=subprocess.STDOUT, env=env)
+        if isinstance(result, str):
+            # Python 2
+            return result
+        else:
+            # Python 3
+            return result.decode()
```
miklos1: `subprocess.check_output` returns `bytes` not `str` in Python 3.

scripts/firedrake-install
```diff
@@ -717,7 +723,7 @@ def build_and_install_h5py():
     else:
         try:
             petsc_dir = check_output(python +
-                                     ["-c", "import petsc; print petsc.get_petsc_dir()"]).strip()
+                                     ["-c", "import petsc; print(petsc.get_petsc_dir())"]).strip()
```
miklos1: `print` function in Python 3, superfluous parenthesis in Python 2, but works in both.

scripts/firedrake-install
```diff
@@ -1221,10 +1227,11 @@ if mode == "install":
...
         v = sys.version_info
-        easyinstall = file("../lib/python%d.%d/site-packages/easy-install.pth" % (v.major, v.minor), "r").readlines()
-        new_packages = [os.getcwd() + "/" + p + "\n" for p in packages]
-        file("../lib/python%d.%d/site-packages/easy-install.pth" % (v.major, v.minor), "w").writelines(
-            easyinstall[:1] + new_packages + easyinstall[1:])
+        if v.major <= 2:
+            easyinstall = open("../lib/python%d.%d/site-packages/easy-install.pth" % (v.major, v.minor), "r").readlines()
+            new_packages = [os.getcwd() + "/" + p + "\n" for p in packages]
+            open("../lib/python%d.%d/site-packages/easy-install.pth" % (v.major, v.minor), "w").writelines(
+                easyinstall[:1] + new_packages + easyinstall[1:])
```
miklos1: Patch not required in Python 3, moreover, it would fail, since `easy-install.pth` does not exist.

tests/regression/test_expressions.py
```diff
@@ -523,7 +524,6 @@ def test_math_functions(expr, value):
     actual = Function(V)
 
     actual.assign(eval(expr))
-    from math import *          # noqa: F401
```
miklos1: Illegal in Python 3.

firedrake/tsfc_interface.py
```diff
@@ -84,7 +84,7 @@ def _cache_store(cls, key, val):
             # No need for a barrier after this, since non root
             # processes will never race on this file.
             with gzip.open(tempfile, 'wb') as f:
-                cPickle.dump(val, f)
+                cPickle.dump(val, f, 0)
```
miklos1: Default protocol number is 0 in Python 2, and 3 in Python 3, and I forgot what was actually wrong with this one...

scripts/firedrake-install
```diff
@@ -59,7 +59,7 @@ class FiredrakeConfiguration(dict):
 
 def deep_update(this, that):
     import collections
-    for k, v in that.iteritems():
+    for k, v in that.items():
```
wence-: Oh, you can't use `from six import iteritems` because it may not be installed yet?

scripts/firedrake-install
wence-: FWIW, this is a pretty cryptic comment...  What is the bug?

scripts/firedrake-install
miklos1: Yes, quite possibly no `six` here.

tests/regression/test_expressions.py
wence-: I think this doesn't DTRT?  Because those math functions are shadowed by the ones imported from ufl (via the `from firedrake import *`).  So maybe there's no need for `from math import *`, because UFL constant folds anyway.

scripts/firedrake-install
miklos1: `FileNotFoundError` (an exception which does not even exist in Python 2)

tests/regression/test_zero_forms.py
```diff
@@ -113,7 +114,6 @@ def test_math_functions(mesh, expr, value, typ, fs_type):
 
     actual = assemble(eval(expr)*dx)
 
-    from math import *          # noqa: F401
```
wence-: Likewise here?

tests/slate/test_scalar_tensors.py
```diff
@@ -44,7 +45,6 @@ def test_functions(mesh, expr, value, typ, fs_type):
 
     actual = assemble(Tensor(eval(expr)*dx))
 
-    from math import *          # noqa: F401
```
wence-: And here.

scripts/firedrake-install
miklos1: Oh, the original bug... I don't remember, this code snippet has been laying around in this file for a while.

miklos1: A few lines up a comment says "Work around easy-install.pth bug."

wence-: yeah, I don't know what that bug is...

firedrake/tsfc_interface.py
miklos1: Problem here is that some `__init__`/`__new__` overrides don't work well with higher protocol versions.

tests/regression/test_expressions.py
miklos1: Fixed.

tests/regression/test_zero_forms.py
miklos1: Fixed.

tests/slate/test_scalar_tensors.py
miklos1: Fixed.

wence-:
I think this is fine.  I guess if we want python 3 support, firedrake-install needs to make a `venv`, not a `virtualenv` in that case...

Given that py3k doesn't currently work, do we want to provide a warning and exit if `sys.executable` is py3?

miklos1:
Making `firedrake-install` work with `venv` requires further work, but `virtualenv` exists and works in Python 3 as well.

Issue #108, opened 2013-12-05
Local solver kernels


colinjcotter:
Request: kernels that assemble local matrices (including mixed function spaces), and then apply direct solver (BLAS) to local matrix/RHS system, and write output to chosen field.

This would be great for reconstructing the mass flux from the density advection to feed to the other advection schemes in the dynamical core project, but I have a possible workaround that is less efficient but will get it working.

miklos1:
@colinjcotter: What is the state of this? Is it implemented already?

colinjcotter:
It is not, we are discussing to implement something like this to use
hybridisation. The main thing is that it needs to be able to operate on
mixed function spaces.

colinjcotter:
My comment about on 23rd Dec 2014 only applies to the Helmholtz case, as it turns out, where pressure can be eliminated analytically. For Poisson, and more complicated things, it is necessary to assemble a local mixed function space block and invert it. We'll need to sketch out a plan for that.

wence-:
Coming as part of @thomasgibson's work.

thomasgibson:
Related to #975

wence-:
Basic functionality is there, please raise specific slate issues.

PR #1026, opened 2017-03-08
Hybridization for RT/BDM mixed problems

thomasgibson:
Addresses #975 and #108. This PR contains the Slate-based preconditioner `HybridizationPC` which takes mixed problems and solves them via hybridization (a subset of static condensation techniques).

I also made corrections to `slac/compiler.py` regarding cell orientations. I caught this bug while running some experiments on spheres. It should be fine now.

Once I get extruded trace elements completed, I will amend the PC in a future PR to support extruded problems.


firedrake/slate/preconditioners.py
```diff
@@ -0,0 +1,285 @@
...
+        self.unbroken_solution = Function(V)
+        self.unbroken_rhs = Function(V)
+
+        # Create the symbolic Schur-reduction
+        Atilde = Tensor(new_form)
+        gammar = TestFunction(TraceSpace)
+        n = FacetNormal(V.mesh())
+
+        # Vector trial function will have a non-empty ufl_shape
+        sigma = next(f for f in TrialFunctions(V_d) if bool(f.ufl_shape))
```
wence-: I note that even though this part of the code is agnostic to which order the spaces come in, the `apply` method assumes that it's `V*P`.

firedrake/slate/preconditioners.py
```diff
@@ -0,0 +1,285 @@
...
+                self.ksp.solve(b, x)
+
+        # Assemble the pressure and velocity (in that order)
+        # using the Lagrange multipliers
+        self._assemble_pressure()
+        self._assemble_velocity()
+
+        # Project the broken solution into non-broken spaces
+        sigma_h, u_h = self.broken_solution.split()
+        sigma_u, u_u = self.unbroken_solution.split()
```
wence-: Inconsistency with naming before.  Also with the naming of the spaces.  Perhaps:

```python
sigma_d, u_d = self.broken_solution.split()
sigma, u = self.unbroken_solution.split()
```

Warning, implicit assumption that you have `V*P`.

firedrake/slate/preconditioners.py
```diff
@@ -0,0 +1,285 @@
...
+        new_form = map_integrand_dags(replacer, context.a)
+
+        # Create the space of approximate traces.
+        # The vector function space will have a non-empty value_shape
+        W = next(v for v in V if bool(v.ufl_element().value_shape()))
+        if W.ufl_element().family() in ["Raviart-Thomas", "RTCF"]:
+            tdegree = W.ufl_element().degree() - 1
+
+        else:
+            tdegree = W.ufl_element().degree()
```
wence-: Question: what if the user is solving a Tensor-Valued x Vector-Valued problem?  Is that even a legitimate thing?

firedrake/slate/preconditioners.py
```diff
@@ -0,0 +1,285 @@
...
+        """
+        from firedrake import project
+
+        # Transfer non-broken x into a firedrake function
+        with self.unbroken_rhs.dat.vec as v:
+            x.copy(v)
+
+        # Transfer unbroken_rhs into broken_rhs
+        field0, field1 = self.unbroken_rhs.split()
+        bfield0, bfield1 = self.broken_rhs.split()
```
wence-: Also, implicit assumption of `V*P`.

firedrake/slate/preconditioners.py
```diff
@@ -0,0 +1,285 @@
...
+        self._assemble_velocity = create_assembly_callable(
+            sigma_sol,
+            tensor=sigma_h,
+            form_compiler_parameters=context.fc_params)
+
+        # Set up the projector for projecting the broken solution
+        # into the unbroken finite element spaces
+        # NOTE: Tolerance here matters!
+        sigma_b, _ = self.broken_solution.split()
+        sigma_u, _ = self.unbroken_solution.split()
```
wence-: Implicit assumption of `V*P`.  Also, continuing inconsistency in what you call these things.

firedrake/slate/preconditioners.py
```diff
@@ -0,0 +1,285 @@
...
+        raise NotImplementedError(
+            "The transpose application of this PC"
+            "is not implemented."
+        )
+
+    def view(self, pc, viewer=None):
+        super(HybridizationPC, self).view(pc, viewer)
+        viewer.printfASCII("Hybridizing mixed system:\n")
+        viewer.pushASCIITab()
+        viewer.printfASCII("KSP solver for the multipliers:\n")
```
wence-: Maybe write something like.  Solves `-K.T P^-1 K` using local elimation for P?  Can we bikeshed on this?

firedrake/slate/preconditioners.py
```diff
@@ -0,0 +1,285 @@
...
+
+    vecs = nullspace.getVecs()
+    tmp = Function(V)
+    tmp_b = Function(V_d)
+    tnsp_tmp = Function(TraceSpace)
+    forward_action = forward * tmp_b
+    new_vecs = []
+    for v in vecs:
+        with tmp.dat.vec as t:
+            v.copy(t)
```
wence-: Note to self.  Add a `.vec_wo` property to dats.

firedrake/slate/slac/compiler.py
```diff
@@ -479,7 +479,7 @@ def facet_integral_loop(cxt_kernel, builder, coordsym, cellfacetsym,
         tensor = eigen_tensor(exp, t, index)
 
         if kinfo.oriented:
-            clist.append(cell_orientations)
+            clist.insert(0, cell_orientations)
```
wence-: Why did the location of the cell orientations change?

firedrake/slate/preconditioners.py
thomasgibson: >Also, this only works for hybridising H(div) x L2 mixed problems, right? Or will it work if I use taylor-hood?

This PC assumes H(div) x L2. I doubt this will work for Taylor-Hood without modifications. This is because two sets of Lagrange multipliers will need to be introduced (I think) since we're breaking two continuous elements (P2-P1). This means the Schur-complement system will be mixed and we run into the "PyOP2 can't write into mixed dats" problem.

> what if the user is solving a Tensor-Valued x Vector-Valued problem? Is that even a legitimate thing?

I don't know. Can we think of such a formulation where we can apply hybridization? Perhaps some other form of static condensation, but that's not in the scope of this PC, I think. At least for this PR.

firedrake/slate/preconditioners.py
```diff
@@ -0,0 +1,285 @@
...
+        # Set up the functions for the original, hybridized
+        # and schur complement systems
+        self.broken_solution = Function(V_d)
+        self.broken_rhs = Function(V_d)
+        self.trace_solution = Function(TraceSpace)
+        self.unbroken_solution = Function(V)
+        self.unbroken_rhs = Function(V)
+
+        # Create the symbolic Schur-reduction
+        Atilde = Tensor(new_form)
```
thomasgibson: By "input form" you mean the original form contained in the `context`? The input form is not broken. I need the broken mixed operator to perform the forward elimination.

firedrake/slate/preconditioners.py
```diff
@@ -0,0 +1,285 @@
...
+            new_vecs = get_trace_nullspace_vecs(K * Atilde.inv, nullsp,
+                                                V, V_d, TraceSpace)
+            tr_nullsp = PETSc.NullSpace().create(vectors=new_vecs,
+                                                 comm=pc.comm)
+            Smat.setNullSpace(tr_nullsp)
+
+        # Set up the KSP for the system of Lagrange multipliers
+        ksp = PETSc.KSP().create(comm=pc.comm)
+        ksp.setOptionsPrefix(prefix + "trace_")
+        ksp.setTolerances(rtol=1e-13)
```
thomasgibson: Ah, this is a remnant of previous experimentation. There's no strong reason. I will remove this and amend.

firedrake/slate/preconditioners.py
```diff
@@ -0,0 +1,285 @@
...
+        # Now we construct the local tensors for the reconstruction stage
+        # TODO: Add support for mixed tensors and these variables
+        # become unnecessary
+        split_forms = split_form(new_form)
+        A = Tensor(next(sf.form for sf in split_forms
+                        if sf.indices == (0, 0)))
+        B = Tensor(next(sf.form for sf in split_forms
+                        if sf.indices == (1, 0)))
+        C = Tensor(next(sf.form for sf in split_forms
+                        if sf.indices == (1, 1)))
```
thomasgibson: Your suggestion sounds safer, I will change that accordingly. Yes, I was assuming the system would have this symmetry.

firedrake/slate/preconditioners.py
thomasgibson: To address the implicit assumption of `V * P`, should I just inspect the function space (similar to how I grab the vector test functions)?

firedrake/slate/slac/compiler.py
thomasgibson: My previous testing didn't catch that the assembly kernels take the `cell_orientation` arg before any coefficient args. This fixes it up and makes sure the args are passed to the assembly kernels in the right order.

firedrake/slate/preconditioners.py
wence-: Well, this is what I was wondering.  Since the broken form is only used in the forward half inside `K.T A.inv K`, and the fact that wrapping `Tensor` round it "breaks" it, might be enough?

firedrake/slate/preconditioners.py
wence-: Note that the default rtol for the petsc KSP object is 1e-5.

firedrake/slate/preconditioners.py
wence-: I think there are a few options:

1. default to `V*P` documenting (and via `view(pc)`)  with a solver option, say `-trace_field_order V,P` or `-trace_field_order P,V`

2. Require the user to specify (using same options)

3. Require use to specify (via `appctx`)

4. Determine automagically (sort of as you were doing, but only do it once!)

firedrake/formmanipulation.py
```diff
@@ -11,6 +11,28 @@
 from firedrake.ufl_expr import Argument
 
 
+class ArgumentReplacer(MultiFunction):
+    """Replaces an argument with a new argument defined on
+    a new function space.
+    """
```
miklos1: I suspect you could just use `ufl.replace` instead.

firedrake/slate/preconditioners.py
```diff
@@ -0,0 +1,285 @@
...
+        """
+        from ufl.algorithms.map_integrands import map_integrand_dags
+        from firedrake import (FunctionSpace, TrialFunction,
+                               TrialFunctions, TestFunction, Function,
+                               BrokenElement, MixedElement,
+                               FacetNormal, Constant, DirichletBC,
+                               Projector)
+        from firedrake.assemble import (allocate_matrix,
+                                        create_assembly_callable)
+        from firedrake.formmanipulation import ArgumentReplacer, split_form
```
miklos1: Why are all these imports here and not at the beginning of file?

firedrake/slate/preconditioners.py
```diff
@@ -0,0 +1,285 @@
...
+        broken_elements = [BrokenElement(Vi.ufl_element()) for Vi in V]
+        elem = MixedElement(broken_elements)
+        V_d = FunctionSpace(V.mesh(), elem)
+        arg_map = {test: TestFunction(V_d),
+                   trial: TrialFunction(V_d)}
+
+        # Replace the problems arguments with arguments defined
+        # on the new discontinuous spaces
+        replacer = ArgumentReplacer(arg_map)
+        new_form = map_integrand_dags(replacer, context.a)
```
miklos1: I would say `new_form = ufl.replace(context.a, arg_map)`.

firedrake/slate/preconditioners.py
miklos1: Cast to `bool` is superfluous I think.

miklos1: Do you more than one non-scalar trial function, but intentionally keep just the first one? If not, better write this:
`sigma, = (f for f in TrialFunctions(V_d) if f.ufl_shape)`
which implicitly asserts that there is only one such trial function.

firedrake/slate/preconditioners.py
miklos1: Alternatively, one might write:
```python
split_forms = dict(split_form(new_form))
A = Tensor(split_forms[(0, 0)])
B = Tensor(split_forms((1, 0)])
C = Tensor(split_forms((1, 1)])
```

firedrake/slate/slate.py
```diff
@@ -211,6 +211,8 @@ def _output_string(self):
 
     def __str__(self):
         """Returns a string representation."""
+        if isinstance(self, Tensor):
+            return self._output_string()
         return self._output_string(self.prec)
```
miklos1: > Could one attach prec = None to the Tensor class, and not have this special-casing?

:+1:

tests/slate/test_hybrid_poisson_sphere.py
```diff
@@ -0,0 +1,66 @@
...
+
+    V = FunctionSpace(mesh, hdiv_space, 1)
+    U = FunctionSpace(mesh, "DG", 0)
+    W = V * U
+
+    f = Function(U)
+    expr = Expression("x[0]*x[1]*x[2]")
+    f.interpolate(expr)
+
+    u_exact = Function(U).interpolate(Expression("x[0]*x[1]*x[2]/12.0"))
```
miklos1: Old school C string `Expression`s

tests/slate/test_hybrid_poisson_sphere.py
```diff
@@ -0,0 +1,66 @@
...
+def run_hybrid_poisson_sphere(MeshClass, refinement, hdiv_space):
+    mesh = MeshClass(refinement_level=refinement)
+    mesh.init_cell_orientations(Expression(("x[0]", "x[1]", "x[2]")))
+
+    V = FunctionSpace(mesh, hdiv_space, 1)
+    U = FunctionSpace(mesh, "DG", 0)
+    W = V * U
+
+    f = Function(U)
+    expr = Expression("x[0]*x[1]*x[2]")
```
miklos1: Old school C string `Expression`s

tests/slate/test_orientations.py
```diff
@@ -1,40 +0,0 @@
...
-    refC = assemble(flux_ds)
-
-    assert np.allclose(A.M.values, refA.M.values, rtol=1e-13)
-    assert np.allclose(B.dat.data, refB.dat.data, rtol=1e-13)
-    assert np.allclose(C.dat.data, refC.dat.data, rtol=1e-13)
-
-
-if __name__ == '__main__':
-    import os
-    pytest.main(os.path.abspath(__file__))
```
miklos1: Was this a broken test?

firedrake/formmanipulation.py
wence-: Probably, although that does some form preprocessing....

tests/slate/test_hybrid_poisson_sphere.py
dorugeber: are great ;)

firedrake/slate/preconditioners.py
```diff
@@ -0,0 +1,285 @@
...
+        else:
+            tdegree = W.ufl_element().degree()
+
+        # NOTE: Once extruded is ready, we will need to be aware of this
+        # and construct the appropriate trace space for the HDiv element
+        TraceSpace = FunctionSpace(V.mesh(), "HDiv Trace", tdegree)
+
+        # NOTE: For extruded, we will need to add "on_top" and "on_bottom"
+        trace_conditions = [DirichletBC(TraceSpace, Constant(0.0),
+                                        "on_boundary")]
```
thomasgibson: This enforces that the trace variables don't contribute anything on the exterior boundary, as it should. It's our only "BC" in this case. Any strong BCs on the pressure term will already appear in the mixed formulation as a natural (weak) BC.

tests/slate/test_hybrid_poisson_sphere.py
thomasgibson: I will modernize the tests then

tests/slate/test_orientations.py
thomasgibson: It wasn't "complete." I didn't catch the bug I mentioned above concerning orientations. However, I will reintroduce these tests and update this one. I just need to include some coefficient information in the forms.

firedrake/slate/preconditioners.py
thomasgibson: As a remark: when I recover using all the blocks `A, B, C, D` I lose some precision when measuring the error in `test_slate_hybridization.py`. Instead of `1e-11` relative error between hybridized and non-hybridized solutions, I'm hovering at around `1e-7`. I've tried changing the solver parameters, but opted to stay with what's currently there.

After running some resolution tests (not tests in this PR but on my machine), I still get expected convergence and accuracy results. I think this is something worth discussing at some point.

wence-: Hmmm, that is a big difference...

thomasgibson: Wow. Please disregard my last comment. I just found out what went wrong. I forgot to change a minus sign in the recovery expression... everything is fine now.

firedrake/slate/preconditioners.py
thomasgibson: Some of the tools I need (`DirichletBC`, `Projector`/`project`, and everything from `assemble.py` to name a few) cause `ImportError`s if I try to import them from the beginning of the file. So I opted to import most of what I need here.

I don't think I'm doing anything too radical here. I'm importing a lot so I understand the possible eye-sore. If you think it's a big problem, I can move the imports which don't cause trouble.

firedrake/slate/slate.py
```diff
@@ -242,6 +242,7 @@ class Tensor(TensorBase):
     """
 
     operands = ()
+    prec = None
```
miklos1: Why is this needed?

thomasgibson: If you want to debug Slate and, for what ever reason, want to `print Tensor(foo)`, then this must be specified. Otherwise an `AttributeError` occurs. It's primarily used in the compiler to parenthesize, but it's also used for pretty printing.

miklos1: What is the backtrace? Perhaps you just need to add `Tensor` with the _most binding_ precedence. I suspect you are now comparing `None` to `int`s which does not work in Python 3.

thomasgibson:
Okay, I believe it's ready for another look-over.

thomasgibson:
Excellent. @wence- when you get the chance, could you have a look?

firedrake/slate/preconditioners.py
```diff
@@ -0,0 +1,370 @@
...
+        opts = PETSc.Options()
+        fact_type = opts.getString(prefix + "fieldsplit_schur_fact_type",
+                                   "default")
+
+        # Generate reconstruction calls
+        self._reconstruction_calls(split_mixed_op, split_trace_op, fact_type)
+
+        # Set up the projectors
+        self.data_projector = Projector(self.unbroken_rhs.split()[self.vidx],
+                                        self.broken_rhs.split()[self.vidx])
```
wence-: The projection into the broken space can be done element-wise?  So you could set `-ksp_type preonly -pc_type bjacobi -sub_pc_type ilu`?

thomasgibson: I believe it can. Should we allow this to be specified by the user through options? Or just hard-code these into the projector?

PR #1027, opened 2017-03-09
Notebook testing

wence-:
Use [nbval](https://github.com/computationalmodelling/nbval) to validate the outputs of the tutorial notebooks.

Also changed: allow point location and function evaluation to specify a tolerance for the `inside_check`.  Necessary to allow plotting in one of the notebooks.  @miklos1 I think this is fine, please check it.


wence-:
Note that I've only changed the tolerance for plotting purposes, rather than unilaterally.

thomasgibson:
Drone failed due to a timeout in `example-stokes-bc-control.ipynb` (SNES timeout). This occurs in cells 7 and 9 and not marked as `xfail`. Should we be concerned with that or just mark as `xfail`?

PR #1030, opened 2017-03-16
replaced numpy.rank with numpy.ndim in firedrake/interpolation

APaganini:
The goal is to avoid a numpy Deprecation Warning.


miklos1:
First line of commit message is too long (see the odd line wrapping in GitHub), and committer email isn't registered at GitHub, so it won't identify that it's your commit.

Otherwise the content is fine.

thomasgibson:
Perhaps we should wait for the Drone build to finish pending?

PR #1034, opened 2017-03-30
Jenkins parallel tests

dham:
This makes Jenkins on Linux work. Merging this PR will cause the Jenkins tests to run properly for all branches.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -31,9 +31,9 @@ pipeline {
           timestamps {
             sh """
 . ./firedrake/bin/activate
-firedrake-clean
-pip install pytest-cov pytest-xdist
-cd firedrake/src/firedrake; py.test --cov firedrake --short -v ${TEST_FILES}
+python firedrake-clean
+python pip install pytest-cov pytest-xdist
+cd firedrake/src/firedrake; python py.test --cov firedrake --short -v ${TEST_FILES}
 """
           }
         }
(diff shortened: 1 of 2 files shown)


wence-:
Can we add the coverage checking?  See https://docs.codecov.io/docs/supported-ci-providers

Jenkinsfile
```diff
@@ -43,10 +61,26 @@ cd firedrake/src/firedrake; py.test --cov firedrake --short -v ${TEST_FILES}
...
+    }
+    stage('Codecov'){
+      steps {
+        dir('tmp') {
+          timestamps {
+            sh '''
 . ./firedrake/bin/activate
-cd firedrake/src/dolfin-adjoint; py.test -v tests_firedrake
-"""
+pip install flake8
```
wence-: No need to install flake8 here?

dham: Oh. Duh. copy paste fail.

wence-:
It's unclear to me where the mac build went (are we still waiting for the new mac mini?).  Additionally, I would expect the merge commit to be tested here (it was previously), but now it looks like the github status is only spewing back a single commit status?

dham:
Note that the PR is not getting built by Jenkins due to some overzealous Jenkins security on branches which change the Jenkinsfile.

wence-:
Ah, makes sense.

Issue #407, opened 2014-12-02
The difference between w.split() and split(w) is irritating


dham:
I'm not sure this can be fixed but we definitely need to document it very, very clearly.

tkarna:
Would it make sense to rename one of those functions to avoid confusion? As they have the same name it is intuitive to assume they do the same thing, like with numpy.max(a) and a.max().

dham:
Unfortunately the wierd behaviour of split is something we inherit from FEniCS so we would gratuitously break compatibility if we changed the names.

wence-:
Fixed by #1035.

PR #1035, opened 2017-03-30
Add error checking inside derivative for u.split()

wence-:
If one defines a form with u.split(), rather than split(u), and then
attempts to take a derivative wrt u, bad things happen.  In
particular, UFL does not notice that the form has dependence on u, and
therefore the resulting form is always zero.  Since this is an easy
mistake to make, we should catch it.  I extract the coefficients from
the form and check if any of them are in the set of u.split().  If so,
raise a ValueError suggesting a fix.  No false positives occur in the
test suite, so I'm going with a exception, rather than a warning,
because no-one reads those anyway.

### DIFF
--- a/firedrake/ufl_expr.py
+++ b/firedrake/ufl_expr.py
@@ -3,7 +3,7 @@ import ufl
 import ufl.argument
 from ufl.assertions import ufl_assert
 from ufl.split_functions import split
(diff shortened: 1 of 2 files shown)


wence-:
Plausibly fixes #407.

PR #1036, opened 2017-03-30
Nested options

wence-:
Support specifying solver parameters using nested options dicts.  Useful for complex configurations, and using the same set of parameters in different subblocks.

### DIFF
--- a/demos/matrix_free/rayleigh-benard.py.rst
+++ b/demos/matrix_free/rayleigh-benard.py.rst
@@ -42,7 +42,7 @@ Along with gravity, which points down. ::
       inner(grad(u), grad(v))*dx
       + inner(dot(grad(u), u), v)*dx
       - inner(p, div(v))*dx
-      - Ra*Pr*inner(T*g, v)*dx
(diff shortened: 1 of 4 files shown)


dham:
I have one query. Is the current concatenation rule the Right Thing or should components be joined with an underscore thereby removing the trailing underscore on the first component. The latter would seem more natural to me.

A third option would be to add the underscore if not present. Maybe that's the one which would minimise user error.

wence-:
I prefer either

1. straight concatenation (do what I say)
2. appending `_` only if it doesn't exist (do what I meant)

Right now, I mimicked the current code for when one provides an `options_prefix=` argument to a solve call (where I do straight concatenation).  I guess if we do automagic here, we should also do it in the options_prefix handling.

dham:
I think I'm in favour of the magic, simply because I foresee this being a frequent user error. (I suspect users don't use options_prefix= so much, but I agree with the point about consistency).

wence-:
Done in 224a473 with MOAR docs.  Please check if explanations make sense.

PR #1039, opened 2017-04-07
Make framework exceptions nonfatal.

dham:
If the framework munging fails for any reason, log the exception
but keep calm and carry on.

### DIFF
--- a/scripts/firedrake-install
+++ b/scripts/firedrake-install
@@ -1260,7 +1260,12 @@ os.environ["VIRTUAL_ENV"] = virtual_env_vars["base"]
 
 os.chdir(firedrake_env)
 
-framework = fix_osx_virtualenv_python()
+try:
+    framework = fix_osx_virtualenv_python()
+except BaseException as e:
+    log.exception(e)
+    framework = False
+
 
 if mode == "install":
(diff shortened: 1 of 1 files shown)


scripts/firedrake-install
```diff
@@ -1184,7 +1184,12 @@ os.environ["VIRTUAL_ENV"] = virtual_env_vars["base"]
 
 os.chdir(firedrake_env)
 
-framework = fix_osx_virtualenv_python()
+try:
+    framework = fix_osx_virtualenv_python()
+except BaseException as e:
+    log.exception(e)
+    framework = False
```
wence-: This is OK, although I wonder, depending on which step fails, whether we can end up in a broken state but still say "success"?

dham: Hmm. That's also possible.

miklos1:
Rebased on `master`.

wence-:
Alternate proposal, we could remove this code entirely, people can still use matplotlib, they just need to select a different backend (as described in the matplotlib faq).  I do that in 3ef94216951.

wence-:
replaced by removing framework munging.

PR #1040, opened 2017-04-08
add helmholtz tests for higher-order spheres

dorugeber:
Scalar Helmholtz tests work at lowest- and higher-order.  Mixed Helmholtz tests work at lowest-order and for higher-order icos, but fail for higher-order cubed-sphere (and the parallel higher-order cubed-sphere test).

P.S. How do I just xfail one parameter choice out of a set?  (UnitCubedSphereMesh in the 4th set of tests)


miklos1:
Superseded by #1041.

PR #1045, opened 2017-04-11
Temporary workaround for --download-eigen issue

wence-:
Please check the logic carefully.

Idea:

- Check if petsc is the "buggy" version
- If yes, apply patch
- If no, check if HEAD~ matches origin/firedrake
- If yes, the remote has obtained petsc update (with fix), so pull that.
- If not, don't update.

So now, when we push an upstream fix, then the repository is reset and this patch "never existed".

### DIFF
--- a/scripts/firedrake-install
+++ b/scripts/firedrake-install
@@ -439,6 +439,77 @@ def git_url(plain_url, protocol):
         raise ValueError("Unknown git protocol: %s" % protocol)
 
 
+def update_and_apply_petsc_fix():
+    HEAD1 = check_output(["git", "rev-parse", "HEAD~"]).strip()
+    # Fetch from upstream
+    check_call(["git", "fetch", "origin"])
+    # Check if we've applied horrific monkey-patching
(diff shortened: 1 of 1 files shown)


scripts/firedrake-install
```diff
@@ -439,6 +439,77 @@ def git_url(plain_url, protocol):
...
+        try:
+            p = subprocess.Popen(["git", "am"], stdout=subprocess.PIPE,
+                                 stdin=subprocess.PIPE, stderr=subprocess.STDOUT)
+            p.communicate(input=patch)
+        except subprocess.CalledProcessError as e:
+            log.debug(e.output)
+            raise
+    else:
+        # Back to normal operation (neither HEAD nor HEAD~ are buggy)
+        check_call(["git", "pull"])
```
miklos1: But the new HEAD may be buggy now...

miklos1: How about this logic:
```python
if HEAD1 == magic_sha:
    # same as before
else:
    # ordinary pull
    if HEAD == magic_sha:
        # same as before

wence-:
Like that?

PR #1047, opened 2017-04-18
Update download page

dorugeber:
Since we no longer have caching (right?), remove obsolete mention of cachetools and psutil.  Add mention of, and link to, FInAT.


miklos1:
What kind of caching does we not have? We still seem to use `cachetools`...

dorugeber:
Fail, let's try that again.  Now just removing psutil, and also from the `requirements-ext.txt` file.

PR #1049, opened 2017-04-19
Extruded trace testing and hybridization

thomasgibson:
Just adding some generic tests that make sure the trace element works on extruded meshes. Also modifies the hybridization solver to support extruded mixed problems.

I removed a kwarg that was not only a misnomer, but turns out to not be that useful. The hybridization solver will only eliminate the vector unknown first when doing local reconstructions. When we are able to write into mixed dats, local reconstructions will simplify into one clean expression.

### DIFF
--- a/firedrake/slate/preconditioners.py
+++ b/firedrake/slate/preconditioners.py
@@ -115,17 +115,13 @@ class HybridizationPC(PCBase):
         # boundary. Extruded cells will have both horizontal and vertical
         # facets
         if mesh.cell_set._extruded:
-            zero_trace_condition = [DirichletBC(TraceSpace, Constant(0.0),
-                                                "on_boundary"),
-                                    DirichletBC(TraceSpace, Constant(0.0),
-                                                "bottom"),
-                                    DirichletBC(TraceSpace, Constant(0.0),
-                                                "top")]
+            trace_bcs = [DirichletBC(TraceSpace, Constant(0.0), "on_boundary"),
+                         DirichletBC(TraceSpace, Constant(0.0), "bottom"),
+                         DirichletBC(TraceSpace, Constant(0.0), "top")]
             K = Tensor(gammar('+') * ufl.dot(sigma, n) * ufl.dS_h +
                        gammar('+') * ufl.dot(sigma, n) * ufl.dS_v)
         else:
-            zero_trace_condition = [DirichletBC(TraceSpace, Constant(0.0),
-                                                "on_boundary")]
+            trace_bcs = [DirichletBC(TraceSpace, Constant(0.0), "on_boundary")]
             K = Tensor(gammar('+') * ufl.dot(sigma, n) * ufl.dS)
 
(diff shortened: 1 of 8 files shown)


firedrake/slate/preconditioners.py
```diff
@@ -69,34 +66,27 @@ def initialize(self, pc):
...
         # Create the space of approximate traces.
-        # TODO: Once extruded and tensor product trace elements
-        # are ready, this logic will be updated.
         W = V[self.vidx]
-        hdiv_family = W.ufl_element().family()
-        if hdiv_family == "Raviart-Thomas":
-            tdegree = W.ufl_element().degree() - 1
-
-        elif hdiv_family == "Brezzi-Douglas-Marini":
+        if W.ufl_element().family() == "Brezzi-Douglas-Marini":
```
thomasgibson: An option here that I've been thinking of is determining the degree of the trace space based on the number of facet degrees of freedom.

For example, the lowest order RT element has 1 dof per facet, therefore a trace space of piece-wise constants should be used. A BDM1 element has 2 dofs per facet, so we need to use piece-wise linears on the facet field. Is there a convenient way to do this?

wence-: You could ask the finat element from the dofs on the facet?

thomasgibson: Update: I've leaving this as it. After our discussion today, I feel it's best to just leave this logic here (subject to editing if necessary). Note that if `BDMCF` existed in FIAT, the logic for BDM applies in the same way.

firedrake/slate/preconditioners.py
```diff
@@ -124,6 +124,12 @@ def initialize(self, pc):
             trace_bcs = [DirichletBC(TraceSpace, Constant(0.0), "on_boundary")]
             K = Tensor(gammar('+') * ufl.dot(sigma, n) * ufl.dS)
 
+        # If boundary conditions are contained in the ImplicitMatrixContext:
+        if self.cxt.row_bcs:
+            raise NotImplementedError(
+                "Strong BCs not currently handled. Try imposing them weakly."
+            )
+
```
thomasgibson: Strong boundary conditions coming through is not a straightforward: "apply bcs to function then project into the broken space." I have tried many approaches and I simply cannot get this right.

The dofs associated with a non-broken HDiv space are topologically associated with the boundary facets, while the broken HDiv space only has "cell" dofs in this case. There doesn't seem to be a straightforward mapping yet which allows one to safely bring the bcs into the broken space. I have tried the naive "project them in" approach and it still doesn't get it right.

For now, I raise this exception until I figure out what needs to be done. Any help would be greatly appreciated.

wence-:
You need `--download-scalapack` at least, maybe also `--download-metis --download-parmetis`

thomasgibson:
Thanks @wence-. Updated. I will also push some minor clean ups soon (mainly removing line-breaks).

PR #1050, opened 2017-04-22
Fix libspatialindex upgrade path

miklos1:
Tried to test syncing our libspatialindex fork with upstream, however, `git pull` fails because of the local changes applied to the source tree.

### DIFF
--- a/scripts/firedrake-install
+++ b/scripts/firedrake-install
@@ -872,15 +872,14 @@ def build_and_install_libspatialindex():
     log.info("Installing libspatialindex\n")
     if os.path.exists("libspatialindex"):
         log.info("Updating the git repository for libspatialindex")
-        os.chdir("libspatialindex")
-        check_call(["git", "fetch"])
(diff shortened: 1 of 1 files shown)


scripts/firedrake-install
```diff
@@ -871,7 +871,16 @@ def build_and_install_glpsol():
...
+        log.info("Updating the git repository for libspatialindex")
+        os.chdir("libspatialindex")
+        check_call(["git", "fetch"])
+        git_sha = check_output(["git", "rev-parse", "HEAD"])
+        git_sha_new = check_output(["git", "rev-parse", "@{u}"])
+        libspatialindex_changed = git_sha != git_sha_new
+        if libspatialindex_changed:
+            check_call(["git", "reset", "--hard"])
+            check_call(["git", "pull"])
+        os.chdir("..")
```
wence-: use the `with directory("libspatialindex")` context manager, otherwise fine.

PR #1053, opened 2017-04-24
Variable number of cell layers for extruded meshes

wence-:
## Summary

This brings in the ability to handle "generalised" z-layer extruded meshes.

We relax the requirement that every cell column contains the same number of cells and starts and stops at the same height.  Where cell columns have neighbours, they must still match topologically (no hanging nodes anywhere).

So that the extra "external" facets that are thus introduced do not have to be treated specially, we instead treat them as interior facets, but guarantee never to iterate over them.  The user must therefore arrange that these facets have zero area.

## Missing functionality (raises `NotImplementedError`)

1. Point evaluation.  The cell location algorithm on extruded meshes doesn't work right now with variable numbers of cell layers

2. SLATE.  Actually, it's only some of the facet integrals, but we need to figure out how to arrange that we don't accidentally integrate on the zero area interior facets.

3. matplotlib plotting (related to 1.).

## Discussion points

Some of the new code is just a slight generalisation of the existing code in `dmplex.pyx`.  I can refactor so that there's only the more general code (it's slightly slower, but not by much), at the expense of increasing the complication of the "base" code a little.  OTOH, this duplication is probably bad.

Domain decomposition is done on the 2D mesh (before knowing how many cells are going to be in each column).  This has the potential to result in suboptimal load balance.  I think for now, we leave it, but note this issue and come up with a way to weight the distribution by a putative number of cells in each column.


wence-:
With respect to boundary nodes, right now I identify the circled nodes as on the top and bottom boundaries:
current bc nodes

But the PyOP2 codegen in the matrix also includes the nodes in this picture that are surrounded by a rectangle.  In both cases, I don't include those surrounded by a diamond.  Which if I were on an unstructured mesh I would expect, no?

possible bcs?

So this needs some more thought, I think.

Including the squared nodes is quite easy, the diamond surrounded ones are harder (I need a mask per column...)

firedrake/extrusion_numbering.pyx
```diff
@@ -0,0 +1,762 @@
...
+    |        |
+    0--------3
+
+The bottom boundary nodes are::
+
+   [0, 3, 6, 9]
+
+whereas the top are::
+
+   [2, 5, 8, 10]
```
wence-: I think this is wrong.  On an unstructured mesh I'd expect the "bottom" boundary nodes to be `[0, 3, 4, 6, 7, 9]` and the "top" to be `[2, 5, 7, 8, 10]`.

Additionally, I think I need to include any nodes that are on, e.g., the 7-8 facet.

firedrake/extrusion_numbering.pyx
```diff
@@ -0,0 +1,762 @@
...
+
+We just need to ensure that when we visit the second cell column for
+the bottom nodes we do not consider the node 4 to be at the bottom of
+the mesh, even though it looks to be at the bottom of the column for
+that cell.
```
wence-: As above, this is probably not right.

firedrake/extrusion_numbering.pyx
```diff
@@ -0,0 +1,762 @@
...
+    CHKERR(DMPlexRestoreTransitiveClosure(dm.dm, 0, PETSC_TRUE, NULL, &closure))
+    if mesh.comm.size == 1:
+        return layer_extents
+
+    # OK, now we have the correct extents for owned points, but
+    # potentially incorrect extents for ghost points, so do a SF Bcast
+    # over the point SF to get it right.
+    # TODO: this only works because we have a cell halo that contains
+    # all the cells that touch the vertices we own.  If this changes,
+    # e.g we don't grow the halo overlap, then we need to fix it.
```
wence-: That would be `PetscSFReduceBeginEnd(..., op=custom_min_max_max_min)` followed by `PetscSFBcastBeginEnd` I think.

firedrake/extrusion_numbering.pyx
```diff
@@ -0,0 +1,762 @@
...
+    dimension = dm.getDimension()
+
+    for i in range(dimension + 1):
+        pStart, pEnd = dm.getDepthStratum(i)
+        for p in range(pStart, pEnd):
+            if variable:
+                layers = layer_extents[p, 1] - layer_extents[p, 0]
+                ndof = layers*nodes[i, 0] + (layers - 1)*nodes[i, 1]
+            else:
+                ndof = nodes[i]
```
wence-: This branching is the only thing different to the `dmplex.create_section` implementation, ripe for merging.

firedrake/extrusion_numbering.pyx
```diff
@@ -0,0 +1,762 @@
...
+
+    layer_extents = mesh.layer_extents
+    offset = 0
+    if label is not None:
+        CHKERR(DMGetLabel(dm.dm, <char *>label, &clabel))
+        CHKERR(DMLabelCreateIndex(clabel, hStart, hEnd))
+    pStart, pEnd = dm.getChart()
+    CHKERR(ISGetIndices((<PETSc.IS?>mesh._plex_renumbering).iset, &renumbering))
+    for p in range(pStart, pEnd):
+        point = renumbering[p]
```
wence-: This is the translation from DMPlex point numbering to Firedrake numbering.  By writing this generically we don't have to build a Section that gives us the renumbering for each entity type.

firedrake/extrusion_numbering.pyx
```diff
@@ -0,0 +1,762 @@
...
+    dimension = dm.getDimension()
+    stratum_bounds = numpy.zeros((dimension + 1, 2), dtype=IntType)
+    for i in range(dimension + 1):
+        stratum_bounds[i, :] = dm.getDepthStratum(i)
+
+    for i, lbl in enumerate(["op2_core", "op2_non_core",
+                             "op2_exec_halo", "op2_non_exec_halo"]):
+        if dm.getStratumSize(lbl, 1) < 1:
+            continue
+        indices = dm.getStratumIS(lbl, 1).indices
```
wence-: Use `ISGetRestoreIndices`.

firedrake/extrusion_numbering.pyx
```diff
@@ -0,0 +1,762 @@
...
+        for i in range(nclosure):
+            entity = cell_closures[cell, i]
+            CHKERR(PetscSectionGetDof(global_numbering.sec, entity, &ndofs))
+            if ndofs > 0:
+                CHKERR(PetscSectionGetOffset(global_numbering.sec, entity, &off))
+                # The cell we're looking at the entity through is
+                # higher than the lowest cell the column touches, so
+                # we need to offset by the difference from the bottom.
+                if variable:
+                    off += offset[flat_index[k]]*(layer_extents[c, 0] - layer_extents[entity, 0])
```
wence-: This branch is the only difference from the `dmplex.get_cell_nodes` implementation.  Ripe for merging.

firedrake/extrusion_numbering.pyx
```diff
@@ -0,0 +1,762 @@
...
+                PetscSectionGetOffset(cell_numbering.sec, support[i], &cell)
+                for j in range(ndof):
+                    dof = cell_nodes[cell, j]
+                    if variable:
+                        # This facet iterates from higher than the
+                        # cell numbering of the cell, so we need
+                        # to add on the difference.
+                        dof += offset[j]*(layer_extents[point, 2] - layer_extents[support[i], 0])
+                    facet_nodes[facet, ndof*i + j] = dof
+            facet += 1
```
wence-: This code, with a little bit of refactoring, would also do to replace the `dmplex.get_facet_nodes`.  And we can do away with carrying around the facet renumbering.

firedrake/extrusion_numbering.pyx
```diff
@@ -0,0 +1,762 @@
...
+            entity = cell_closure[cell, entity_mask[i, 1]]
+            offset = entity_mask[i, 2]
+            # Now we need to check if this dof is actually on the boundary.
+            if bottom:
+                # On the bottom, we need the cell column we're looking
+                # through to start at the same height as the entity
+                # column.
+                if layer_extents[c, 0] == layer_extents[entity, 0]:
+                    indices[idx] = dof
+                    idx += 1
```
wence-: As described above, I think this reasoning is incorrect.

firedrake/extrusion_utils.py
```diff
@@ -29,12 +28,12 @@ def make_extruded_coords(extruded_topology, base_coords, ext_coords,
...
         void kernel(double **base_coords, double **ext_coords,
-                    int **layer, double *layer_height)
+                    double *layer_height, int layer)
 
     The kernel iterates over the cells of the mesh and receives as
     arguments the coordinates of the base cell (to read), the
-    coordinates on the extruded cell (to write to), the layer number
-    of each cell and the fixed layer height.
+    coordinates on the extruded cell (to write to), the fixed layer
+    height, and the current cell layer.
```
wence-: Need to change docs as well.

firedrake/function.py
```diff
@@ -496,9 +496,12 @@ def at(self, arg, *args, **kwargs):
         if not arg.shape:
             arg = arg.reshape(-1)
 
+        mesh = self.function_space().mesh()
+        if mesh.cell_set._extruded and not mesh.cell_set.constant_layers:
```
wence-: Better way of checking this condition?

firedrake/mesh.py
```diff
@@ -808,41 +843,73 @@ def interior_facets(self):
...
+            nodes_per_entity = sum(nodes[:, i]*(self.layers - i) for i in range(2))
+            return super(ExtrudedMeshTopology, self).create_section(nodes_per_entity)
+        else:
+            return extnum.create_section(self, nodes_per_entity)
+
+    def node_classes(self, nodes_per_entity):
+        """Compute node classes given nodes per entity.
+
+        :arg nodes_per_entity: number of function space nodes per topological entity.
+        :returns: the number of nodes in aech of core, owned, exec, and non exec classes.
```
wence-: typo.

firedrake/extrusion_utils.py
```diff
@@ -150,17 +149,29 @@ def make_extruded_coords(extruded_topology, base_coords, ext_coords,
...
+        for i in entity_dofs[(b, v)]:
+            # This line is fairly magic.
+            # It works because an interval has two points.
+            # We pick up the DoFs from the bottom point,
+            # then the DoFs from the interior of the interval,
+            # then finally the DoFs from the top point.
+            flat_entity_dofs[b][i] = (entity_dofs[(b, 0)][2*i] +
+                                      entity_dofs[(b, 1)][i] +
+                                      entity_dofs[(b, 0)][2*i+1])
+    return flat_entity_dofs
```
miklos1: We must have this code somewhere already...

firedrake/mesh.py
```diff
@@ -808,41 +843,73 @@ def interior_facets(self):
...
+        """
+        if self.cell_set.constant_layers:
+            nodes = np.asarray(nodes_per_entity)
+            nodes_per_entity = sum(nodes[:, i]*(self.layers - i) for i in range(2))
+            return super(ExtrudedMeshTopology, self).node_classes(nodes_per_entity)
+        else:
+            return extnum.node_classes(self, nodes_per_entity)
```
wence-: Should return tuple for consistency.

firedrake/mesh.py
```diff
@@ -1304,6 +1416,24 @@ def ExtrudedMesh(mesh, layers, layer_height=None, extrusion_type='uniform', kern
     import firedrake.function as function
 
     mesh.init()
+    try:
+        layers = np.asarray(layers, dtype=IntType).reshape(mesh.cell_set.total_size, 2)
+        layers[:, 1] += 1 + layers[:, 0]
+    except ValueError:
+        layers = np.asarray(layers, dtype=IntType)
+        if layers.shape:
+            raise ValueError("Must provide scalar layer value, not %s", layers)
```
wence-: Provide more useful error message in case of mismatching shape.

firedrake/functionspacedata.py
```diff
@@ -89,68 +91,72 @@ def get_node_set(mesh, nodes_per_entity):
...
     node_set = op2.Set(node_classes, halo=halo_mod.Halo(dm), comm=mesh.comm)
     # Don't need it any more, explicitly destroy.
     dm.destroy()
-    extruded = bool(mesh.layers)
+    extruded = mesh.cell_set._extruded
     if extruded:
-        node_set = op2.ExtrudedSet(node_set, layers=mesh.layers)
+        # FIXME! This is a LIE! But these sets should not be extruded
+        # anyway, only the code gen in PyOP2 is busted.
+        node_set = op2.ExtrudedSet(node_set, layers=2)
```
miklos1: Yes, when data is accessed through a `Map`, then it matters if the source `Set` (which should be the same as the iteration `Set`) is extruded, but extrusion does not really have a meaning for the destination `Set`, even though parts of the PyOP2 codegen rely on that bit.

firedrake/functionspacedata.py
```diff
@@ -426,6 +480,24 @@ def exterior_facet_boundary_node_map(self, V, method):
...
+    def boundary_nodes(self, V, sub_domain, method):
+        if method not in {"topological", "geometric"}:
+            raise ValueError("Don't know how to extract nodes with method '%s'", method)
+        if sub_domain == "bottom":
+            return V.bottom_nodes(method)
+        elif sub_domain == "top":
+            return V.top_nodes(method)
+        else:
+            key = (entity_dofs_key(V.finat_element.entity_dofs()), as_tuple(sub_domain), method)
+            return get_boundary_nodes(V.mesh(), key, V)
```
miklos1: How are `top_botton_boundary_nodes` different from calling `boundary_nodes` with `sub_domain == "bottom"` or `sub_domain == "top"`?

firedrake/functionspaceimpl.py
```diff
@@ -493,29 +493,38 @@ def exterior_facet_boundary_node_map(self, method):
...
+        """Return the boundary nodes for this :class:`~.FunctionSpace`.
+
+        :arg sub_domain: the mesh marker selecting which subset of facets to consider.
+        :arg method: the method for determining boundary nodes.
+        :returns: A numpy array of the unique function space nodes on
+           the selected portion of the boundary.
+
+        See also :class:`~.DirichletBC` for details of the arguments.
+        """
+        return self._shared_data.boundary_nodes(self, sub_domain, method)
```
miklos1: Ah, so we pass `self` (the function space) to the function space data object, which then can call back to `self.bottom_nodes` or `self.top_nodes` for example. Which in turn calls back to `top_bottom_boundary_nodes` on function space data. Why is this back and forth control necessary?

firedrake/output.py
```diff
@@ -114,26 +114,38 @@ def get_topology(coordinates):
...
+        offsets = 0
     else:
-        cell_layers = mesh.layers - 1
+        if mesh.cell_set.constant_layers:
+            cell_layers = mesh.cell_set.layers - 1
+            offsets = numpy.arange(cell_layers, dtype=IntType) * scale
+            offsets = numpy.tile(offsets.reshape(-1, 1), (num_cells, 1))
+            num_cells *= cell_layers
+        else:
+            layers = mesh.cell_set.layers_array[:num_cells, ...]
```
miklos1: What is it that is beyond `num_cells`? Halo?

firedrake/extrusion_utils.py
wence-: We did, that's why f7c48b7848 removes the usage and turns it into a function.  Unless I missed other usages.

firedrake/functionspaceimpl.py
wence-: Probably not necessary, should splat `top_bottom_...` into `boundary_nodes` and remove other bits.

firedrake/output.py
wence-: yes.

firedrake/extrusion_numbering.pyx
thomasgibson: >On an unstructured mesh I'd expect the "bottom" boundary nodes to be `[0, 3, 4, 6, 7, 9]` and the "top" to be `[2, 5, 7, 8, 10]`.

I agree with you. My concern is how should we handle nodes that are, sort of, both "bottom" and "top" (such as node 7). Will this be a problem?

wence-: No.  It's the same idea as when you have corner nodes that are on both the "top" and "side" of a domain.

wence-:
OK, the boundary conditions on the top and bottom are harder than I anticipated.  They would become marginally easier if the restriction were added that neighbouring cell columns can only differ in height at either end by a single cell, I think.  Is that a reasonable restriction?

wence-:
Note also, the zero-sized facets do seem to cause problems for operators.  We could instead do a `ds_tv` and `ds_bv` measure setup for the "vertical" top and bottom facets.  That might plausibly make things easier in some regards.

firedrake/functionspacedata.py
```diff
@@ -184,32 +190,62 @@ def get_dof_offset(mesh, key, entity_dofs, ndof):
...
+                support_indices.extend(sorted(esd[key]))
+            if sum(ent) == dim - 1:
+                facet_points.append(p)
+            closure_indices.extend(sorted(ecd[ent][key]))
+            p += 1
+    closure_section.setUp()
+    support_section.setUp()
+    closure_indices = PETSc.IS().createGeneral(closure_indices, comm=PETSc.COMM_SELF)
+    support_indices = PETSc.IS().createGeneral(support_indices, comm=PETSc.COMM_SELF)
+    facet_points = PETSc.IS().createGeneral(facet_points, comm=PETSc.COMM_SELF)
```
wence-: Plausibly these arrays don't need to be `IS`es, and can just be numpy arrays.

firedrake/functionspacedata.py
```diff
@@ -441,7 +547,6 @@ def get_map(self, V, entity_set, map_arity, bcs, name, offset, parent):
         # V is only really used for error checking and "name".
         assert len(V) == 1, "get_map should not be called on MixedFunctionSpace"
         entity_node_list = self.entity_node_lists[entity_set]
-        cache = self.map_caches[entity_set]
```
wence-: Removing duplicated variable assignment.

firedrake/functionspaceimpl.py
wence-: Gone now.

firedrake/extrusion_numbering.pyx
```diff
@@ -0,0 +1,1015 @@
...
+                cbottom = layer_extents[c, 0]
+                coffset = csection.getOffset(cell)
+                foffset = section.getOffset(facet)
+                for i in range(top - bottom - 1):
+                    top_masks[foffset + i, j] = ctop[coffset + i + bottom - cbottom]
+                    bottom_masks[foffset + i, j] = cbottom[coffset + i + bottom - cbottom]
+            facet += 1
+    bottom = PETSc.IS().createGeneral(bottom_masks, comm=PETSc.COMM_SELF)
+    top = PETSc.IS().createGeneral(top_masks, comm=PETSc.COMM_SELF)
+    return op2.ExtrudedSet.EntityMask(section=section, bottom=bottom, top=top)
```
wence-: Problem here.  By the time we get in to PyOP2, we don't know we've got a map that's come from a concatenated thing. So how do we deal with that.  Do we treat the two cells as a "macro cell" in some sense, and glue together the masks for the interior facets.  Or do we actually remember this information.

firedrake/mesh.py
```diff
@@ -1304,6 +1426,22 @@ def ExtrudedMesh(mesh, layers, layer_height=None, extrusion_type='uniform', kern
     import firedrake.function as function
 
     mesh.init()
+    layers = np.asarray(layers, dtype=IntType)
+    if layers.shape:
+        if layers.shape != (mesh.cell_set.total_size, 2):
+            raise ValueError("Must provide single layer number or array of shape (%d, 2), not %s",
+                             mesh.cell_set.total_size, layers.shape)
```
tkarna: bad formatting, missing `%`

wence-:
I think this is now ready to be looked at properly.  I've updated for owner-computes, and implemented the necessary magic to make sure `dS_v` integrals get treated correctly in the presence of boundary conditions.

It's quite a lot of code, so it would be good if someone can just sanity check if they can vaguely follow what's going on.

firedrake/extrusion_numbering.pyx
```diff
@@ -0,0 +1,1035 @@
...
+            if ndof <= 0:
+                continue
+            facet = cell_closure[cell, fstart + i]
+            if top:
+                # Is the facet exposed when viewed through this cell?
+                if layer_extents[c, 1] == layer_extents[facet, 3]:
+                    continue
+                # Count number of exposed layers.
+                initial_offset = layer_extents[facet, 3] - layer_extents[c, 0] - 1
+                exposed_layers = layer_extents[facet, 1] - layer_extents[facet, 3]
```
wence-: Number of facets above the facet iteration column.

firedrake/extrusion_numbering.pyx
```diff
@@ -0,0 +1,1035 @@
...
+        int f, i, j, dof, facet, idx
+        int nfacet, nlocal, layers
+        PetscInt local_facet
+        bint all_facets
+
+    facet_dim = V.mesh().facet_dimension()
+    if method == "topological":
+        boundary_dofs = V.finat_element.entity_closure_dofs()[facet_dim]
+    elif method == "geometric":
+        boundary_dofs = entity_support_dofs(V.finat_element, facet_dim)
```
wence-: Use new method, rather than free function.

firedrake/extrusion_numbering.pyx
```diff
@@ -0,0 +1,1035 @@
...
+    # We iterate over the base cell and do all the entities in the extruded cell above it,
+    # therefore we need a mapping from the standard firedrake entity ordering to this one.
+    fiat_cell = as_fiat_cell(mesh.ufl_cell())
+    points = eutils.entity_reordering(fiat_cell)
+    # Some masks incorporate others (since the closure of the entity
+    # is not just the entity itself), we'll use this to minimise the
+    # number of mask bits that are on.
+    flips = numpy.zeros(points.shape[0], dtype=numpy.int64)
+    for k, v in eutils.entity_closures(fiat_cell).items():
+        flips[k] = ~sum(2**x for x in v)
```
wence-: So the idea here is that we need to maintain, on each cell of the extruded mesh, information that indicates which topological entities are "exposed" in that cell (either to the top or the bottom of the mesh).  That's one bit of information per topological entity.  Fortunately, all the cells we deal with have at most 27 topological entities (the hexahedron).  So I could use a 32bit int to encode this.  But wait, interior facets need twice as much, so I use a 64bit int.

Moreover, to minimise the amount of duplicate masking we do in the generated code, we notice that we only need to record the "largest" entity that is exposed (since the closure of that will pick up all the others).  So `flips[k]` stores the bitmask that turns off all entities in the closure of k.

firedrake/extrusion_numbering.pyx
```diff
@@ -0,0 +1,1035 @@
...
+        c = cell_closure[cell, nclosure-1]
+        cell_bottom = layer_extents[c, 0]
+        cell_top = layer_extents[c, 1]
+        CHKERR(PetscSectionSetDof(section.sec, cell, cell_top - cell_bottom - 1))
+        for layer in range(cell_bottom, cell_top - 1):
+            p = 0
+            # Iteration order.  Normal FIAT ordering for the entities
+            # in the base cell.  Then, on each entity, we look at the
+            # bottom vertex of the interval, then the top vertex, then
+            # the cell.
```
wence-: IOW, `point[p]` is the entity at the bottom of the extruded interval, `point[p+1]` is the entity at the top, and `point[p+2]` is the full extruded interval.

firedrake/extrusion_numbering.pyx
```diff
@@ -0,0 +1,1035 @@
...
+
+    label = label.encode()
+    dm = mesh._plex
+    pStart, pEnd = dm.getChart()
+    fStart, fEnd = dm.getHeightStratum(1)
+
+    cell_numbering = mesh._cell_numbering
+    renumbering = mesh._plex_renumbering.indices
+
+
```
wence-: Too much whitespace.

firedrake/extrusion_numbering.pyx
```diff
@@ -0,0 +1,1035 @@
...
+    for p in range(nfacet):
+        CHKERR(PetscSectionSetDof(section.sec, p, layers[p, 1] - layers[p, 0] - 1))
+    section.setUp()
+    top_mask = numpy.zeros(section.getStorageSize(), dtype=numpy.int64)
+    bottom_mask = numpy.zeros_like(top_mask)
+
+    fiat_cell = as_fiat_cell(mesh.ufl_cell())
+    # Number of topological entities on the cell (excluding cell itself)
+    nentities = sum(map(len, fiat_cell.get_topology().values())) - 1
+    assert nentities < 31, "Need fewer than 31 entities"
```
wence-: Because we're going to slam two masks together for interior facets.

firedrake/extrusion_numbering.pyx
```diff
@@ -0,0 +1,1035 @@
...
+    CHKERR(DMGetLabel(dm.dm, <const char *>label, &lbl))
+    CHKERR(DMLabelCreateIndex(lbl, fStart, fEnd))
+    for p in range(pStart, pEnd):
+        point = renumbering[p]
+        if point < fStart or point >= fEnd:
+            continue
+        CHKERR(DMLabelHasPoint(lbl, point, &flg))
+        if not flg:
+            continue
+        ent_bottom = layer_extents[point, 2]
```
wence-: OK, so we're on an appropriate facet.

firedrake/extrusion_numbering.pyx
```diff
@@ -0,0 +1,1035 @@
...
+        CHKERR(DMPlexGetSupport(dm.dm, point, &support))
+        for j in range(supportSize):
+            c = support[j]
+            CHKERR(PetscSectionGetOffset(cell_numbering.sec, c, &cell))
+            cell_bottom = layer_extents[c, 0]
+            CHKERR(PetscSectionGetOffset(cell_section.sec, cell, &cell_offset))
+            CHKERR(PetscSectionGetOffset(section.sec, facet, &facet_offset))
+            for i in range(ent_bottom, ent_top - 1):
+                top_mask[facet_offset + i - ent_bottom] |= cell_top_mask[cell_offset + i - cell_bottom] << (nentities*j)
+                bottom_mask[facet_offset + i - ent_bottom] |= cell_bottom_mask[cell_offset + i - cell_bottom] << (nentities*j)
```
wence-: So the facet masks are just the concatenation of the cell masks in the support of the facet.  Note that I only need to maintain facet masks for "iteration" facets, because I've guaranteed already that I will never assemble over the exposed facets, and so will never have to do any of these horrible tricks.

The first cell's mask is in the bottom `nentities` bits, the second cell's in the top `nentities` bits.  So in fact for most cell types we could get away with 32bits, but I don't care enough.

firedrake/extrusion_utils.py
wence-: Still need to change the documentation.

firedrake/extrusion_utils.py
```diff
@@ -149,17 +153,83 @@ def make_extruded_coords(extruded_topology, base_coords, ext_coords,
...
+
+    :arg cell: a FIAT cell.
+    """
+    indices = entity_indices(cell)
+    closure = {}
+    for e, ents in cell.sub_entities.items():
+        for ent, vals in ents.items():
+            idx = indices[(e, ent)]
+            closure[idx] = list(map(indices.get, vals))
+    return closure
```
wence-: I think fiat doesn't give me this already.

firedrake/function.py
```diff
@@ -504,9 +504,12 @@ def at(self, arg, *args, **kwargs):
         if not arg.shape:
             arg = arg.reshape(-1)
 
+        mesh = self.function_space().mesh()
+        if mesh.variable_layers:
+            raise NotImplementedError("Point evaluation not implemented for variable layers")
```
wence-: To do this requires some much more serious refactoring of the point evaluation code.

firedrake/functionspacedata.py
wence-: I didn't manage to unpick where this is busted.

firedrake/functionspacedata.py
```diff
@@ -81,67 +83,71 @@ def get_node_set(mesh, nodes_per_entity):
...
     :returns: A dict mapping mesh entity sets to numpy arrays of
         function space nodes.
     """
     # set->node lists are specific to the sorted entity_dofs.
-    cell_node_list = get_cell_node_list(mesh, entity_dofs, global_numbering)
-    interior_facet_node_list = get_facet_node_list(mesh, "interior_facets", cell_node_list)
-    exterior_facet_node_list = get_facet_node_list(mesh, "exterior_facets", cell_node_list)
+    cell_node_list = get_cell_node_list(mesh, entity_dofs, global_numbering, offsets)
+    interior_facet_node_list = get_facet_node_list(mesh, "interior_facets", cell_node_list, offsets)
+    exterior_facet_node_list = get_facet_node_list(mesh, "exterior_facets", cell_node_list, offsets)
```
wence-: Arguably we should postpone creation of these things until first use.

firedrake/functionspacedata.py
```diff
@@ -175,32 +181,77 @@ def get_dof_offset(mesh, key, entity_dofs, ndof):
...
+            for key in sorted(ecd[ent].keys()):
+                closure_section.setDof(p, len(ecd[ent][key]))
+                vals = numpy.asarray(sorted(ecd[ent][key]), dtype=IntType)
+                closure_indices.extend(vals + cell*offset)
+                if esd is not None:
+                    support_section.setDof(p, ncell*len(esd[ent][key]))
+                    vals = numpy.asarray(sorted(esd[ent][key]), dtype=IntType)
+                    support_indices.extend(vals + cell*offset)
+                if sum(ent) == dim - 1:
+                    facet_points.append(p)
```
wence-: Remember which are the (macro-)cell facet points, so that we can extract them later.

firedrake/functionspacedata.py
```diff
@@ -175,32 +181,77 @@ def get_dof_offset(mesh, key, entity_dofs, ndof):
...
+                p += 1
+    closure_section.setUp()
+    support_section.setUp()
+    closure_indices = numpy.asarray(closure_indices, dtype=IntType)
+    support_indices = numpy.asarray(support_indices, dtype=IntType)
```
wence-: So these are CSR arrays that index which dofs are on the boundary.  The CSR encoding is in the section, `section.getDof(p) > 0` says that dofs on entity `p` are on the boundary, the offset into the array is given by `section.getOffset(p)`.

firedrake/functionspacedata.py
```diff
@@ -217,6 +268,59 @@ def get_work_function_cache(mesh, ufl_element):
...
+        idx = {"bottom": -2, "top": -1}[sub_domain]
+        section, indices, facet_points = V.cell_boundary_masks[method]
+        facet = facet_points[idx]
+        dof = section.getDof(facet)
+        off = section.getOffset(facet)
+        mask = indices[off:off+dof]
+        nodes = cell_node_list[..., mask]
+        if sub_domain == "top":
+            nodes = nodes + offset[mask]*(mesh.cell_set.layers - 2)
+        return numpy.unique(nodes)
```
wence-: So for constant layers, we just redo existing code.

firedrake/functionspacedata.py
```diff
@@ -217,6 +268,59 @@ def get_work_function_cache(mesh, ufl_element):
...
+    :arg V: The FunctionSpace to select from.
+    :arg entity_dofs: The flattened entity dofs.
+    :returnsL: A numpy array of the (unique) boundary nodes.
+    """
+    _, sub_domain, method = key
+    cell_node_list = V.cell_node_list
+    offset = V.offset
+    if mesh.variable_layers:
+        if method == "geometric":
+            raise NotImplementedError("Generic entity_support_dofs not implemented.")
```
wence-: This is no longer true.

firedrake/functionspacedata.py
```diff
@@ -217,6 +268,59 @@ def get_work_function_cache(mesh, ufl_element):
...
+    nodes = V.exterior_facet_boundary_node_map(method).values_with_halo
+    if sub_domain != "on_boundary":
+        indices = mesh.exterior_facets.subset(sub_domain).indices
+        nodes = nodes.take(indices, axis=0)
+    if not V.extruded:
+        return numpy.unique(nodes)
+    else:
+        offset = V.exterior_facet_boundary_node_map(method).offset
+        return numpy.unique(numpy.concatenate([nodes + i * offset
+                                               for i in range(mesh.cell_set.layers - 1)]))
```
wence-: This is just code movement from `bcs.py`.

firedrake/functionspacedata.py
```diff
@@ -400,15 +506,17 @@ def exterior_facet_boundary_node_map(self, V, method):
...
                      facet_dat(op2.WRITE),
                      local_facet_dat(op2.READ))
 
-        if self.extruded:
-            offset = self.offset[boundary_dofs[0]]
-        else:
-            offset = None
         val = op2.Map(facet_set, self.node_set,
                       nodes_per_facet,
                       facet_dat.data_ro_with_halos,
```
wence-: `exterior_facet_boundary_node_map` is only used in the computation of boundary nodes for bcs in the non-variable-layers case.  Could probably refactor to use some cython code instead.

firedrake/functionspacedata.py
```diff
@@ -530,28 +655,18 @@ def get_map(self, V, entity_set, map_arity, bcs, name, offset, parent,
...
-                        b.append(i+off)
-                    for i in top:
-                        t.append(i)
-                        t.append(i+off)
-                    bt_masks[method] = (b, t)
+            if kind == "interior_facet":
+                boundary_masks = self.interior_facet_boundary_masks
             else:
-                bt_masks = self.bt_masks
+                boundary_masks = self.cell_boundary_masks
```
wence-: A better fix for the interior facet boundary condition issue.

firedrake/extrusion_utils.py
```diff
@@ -149,17 +153,83 @@ def make_extruded_coords(extruded_topology, base_coords, ext_coords,
...
-                              extruded_topology.cell_set.total_size).reshape(layers-1, extruded_topology.cell_set.total_size).T.ravel(), dtype=np.int32)
     height = op2.Global(1, layer_height, dtype=float)
     op2.par_loop(kernel,
                  ext_coords.cell_set,
                  base_coords.dat(op2.READ, base_coords.cell_node_map()),
                  ext_coords.dat(op2.WRITE, ext_coords.cell_node_map()),
-                 layer(op2.READ, layer_fs.cell_node_map()),
-                 height(op2.READ))
+                 height(op2.READ),
+                 pass_layer_arg=True)
```
miklos1: I see, you use this, so no need to generate a layer number `Dat`.

firedrake/mesh.py
```diff
@@ -738,6 +775,12 @@ def __init__(self, mesh, layers):
         self._cell_numbering = mesh._cell_numbering
         self._entity_classes = mesh._entity_classes
         self._subsets = {}
+        self._ufl_cell = ufl.TensorProductCell(mesh.ufl_cell(), ufl.interval)
+
+        self.cell_set = op2.ExtrudedSet(mesh.cell_set, layers=layers)
+        if layers.shape:
+            # Bootstrapping issue, this call needs the cell_set to be created.
```
wence-: Because it accesses `layer_extents` which needs the `cell_set`.

firedrake/mesh.py
wence-: Also the docstring is wrong after owner-computes.

firedrake/mesh.py
```diff
@@ -816,11 +891,52 @@ def make_offset(self, entity_dofs, ndofs):
...
+        """The layer extents for all mesh points.
+
+        For variable layers, the layer extent does not match those for cells.
+
+        :returns: numpy array of layer extents (in PyOP2 format
+            :math:`[start, stop)`), of shape ``(num_cells, 4)`` where
+            the first two extents are used for allocation and the last
+            two for iteration.
+        """
+        extents = extnum.layer_extents(self)
```
wence-: Add `assert self.variable_layers`, might not be possible due to bootstrapping issues?

firedrake/utility_meshes.py
```diff
@@ -28,7 +28,7 @@
            'TorusMesh', 'CylinderMesh']
```
wence-: I just add the `partitioning` keyword to a load of meshes.

tests/extrusion/test_variable_layers_poisson.py
```diff
@@ -0,0 +1,65 @@
+from firedrake import *
+import numpy
+from pyop2.datatypes import IntType
+
+
+def test_steady_advection_variable_layers():
```
wence-: Rename, this is a poisson test, not an advection test.

firedrake/extrusion_utils.py
```diff
@@ -149,17 +153,83 @@ def make_extruded_coords(extruded_topology, base_coords, ext_coords,
...
+    """Return a dict mapping topological entities on a cell to their integer index.
+
+    This provides an iteration ordering for entities on extruded meshes.
+
+    :arg cell: a FIAT cell.
+    """
+    d = collections.defaultdict(functools.partial(next, itertools.count()))
+    for key in sorted(itertools.chain(*(entity for eid in cell.sub_entities.values() for entity in eid.values()))):
+        d[key]
+    return d
```
miklos1: This feels a bit convoluted. Is it not the same as
```python
{e: i for i, e in enumerate(sorted(cell.sub_entities[cell.get_dimension()]))}
```
?

firedrake/extrusion_utils.py
```diff
@@ -149,17 +153,83 @@ def make_extruded_coords(extruded_topology, base_coords, ext_coords,
...
+    counter = collections.Counter()
+
+    topos = (c.get_topology() for c in cell.cells)
+
+    indices = entity_indices(cell)
+    ordering = numpy.zeros(len(indices), dtype=IntType)
+    for i, ent in enumerate(itertools.product(*(itertools.chain(*points(t)) for t in topos))):
+        ordering[i] = indices[ent, counter[ent]]
+        counter[ent] += 1
+    return ordering
```
miklos1: So `ent` is really a tuple dimension here...

firedrake/function.py
```diff
@@ -504,9 +504,12 @@ def at(self, arg, *args, **kwargs):
...
         # Immersed not supported
-        tdim = self.function_space().mesh().ufl_cell().topological_dimension()
-        gdim = self.function_space().mesh().ufl_cell().geometric_dimension()
+        tdim = mesh.ufl_cell().topological_dimension()
+        gdim = mesh.ufl_cell().geometric_dimension()
```
miklos1: I think UFL `Mesh` has `.topological_dimension()` and `.geometric_dimension()` directly.

firedrake/functionspaceimpl.py
miklos1: Slightly dodgy that the function space (`self`) is still passed in to construct something that is supposedly shared by this and other function spaces.

firedrake/functionspacedata.py
```diff
@@ -417,6 +525,24 @@ def exterior_facet_boundary_node_map(self, V, method):
...
+            entity_dofs = eutils.flat_entity_dofs(V.finat_element.entity_dofs())
+            key = (entity_dofs_key(entity_dofs), sub_domain, method)
+            return get_top_bottom_boundary_nodes(V.mesh(), key, V)
+        else:
+            if sub_domain == "on_boundary":
+                sdkey = sub_domain
+            else:
+                sdkey = as_tuple(sub_domain)
+            key = (entity_dofs_key(V.finat_element.entity_dofs()), sdkey, method)
+            return get_boundary_nodes(V.mesh(), key, V)
```
miklos1: ... although the only things actually used from the function space (`V`) are:

- `V.extruded` => `self.extruded`
- `V.mesh()` => `self.mesh`
- `V.finat_element.entity_dofs()`: The FInAT element was used to construct the shared data, so this sharing should be safe, but neither the FInAT element nor its entity dofs are saved on `self` to disentangle this.

firedrake/functionspacedata.py
miklos1: See [this](https://github.com/firedrakeproject/firedrake/compare/node-set-never-extruded) and [this](https://github.com/OP2/PyOP2/compare/fix-extruded-set-ugly).

firedrake/functionspacedata.py
miklos1: :+1:

firedrake/extrusion_numbering.pyx
```diff
@@ -0,0 +1,1035 @@
...
+            p = closure[2*ci]
+            # Allocation bounds
+            # Each entity column is from bottom of lowest to top of highest
+            layer_extents[p, 0] = min(layer_extents[p, 0], cell_extents[cell, 0])
+            layer_extents[p, 1] = max(layer_extents[p, 1], cell_extents[cell, 1])
+            # Iteration bounds
+            # Each entity column is from top of lowest to bottom of highest
+            layer_extents[p, 2] = max(layer_extents[p, 2], cell_extents[cell, 0])
+            layer_extents[p, 3] = min(layer_extents[p, 3], cell_extents[cell, 1])
+    CHKERR(DMPlexRestoreTransitiveClosure(dm.dm, 0, PETSC_TRUE, NULL, &closure))
```
miklos1: So at this point `layers_extent[:, [0, 1]] == layers_extent[:, [2, 3]]`, but then you do a halo exchange on the latter.

wence-: Ummm, not necessarily.

firedrake/mesh.py
```diff
@@ -816,11 +891,52 @@ def make_offset(self, entity_dofs, ndofs):
...
+        if np.any(extents[:, 3] - extents[:, 2] <= 0):
+            raise NotImplementedError("Vertically disconnected cells unsupported")
```
miklos1: How does this mean vertical disconnection?

firedrake/mesh.py
miklos1: It does not help to make this a `cached_property` since if it can be provided, then it is always needed.

firedrake/mesh.py
miklos1: Insert content of `Mesh.layer_extents` here.

firedrake/extrusion_numbering.pyx
```diff
@@ -0,0 +1,1035 @@
...
+    if xin[1] > xout[1]:
+        xout[1] = xin[1]
+    if xin[2] > xout[2]:
+        xout[2] = xin[2]
+    if xin[3] < xout[3]:
+        xout[3] = xin[3]
+
+
+@cython.wraparound(False)
+def layer_extents(mesh):
```
miklos1: Receive `plex`, `cell_numbering`, `cell_extents`, and `comm` instead. That way you don't need to construct the `cell_set` to call this.

firedrake/extrusion_numbering.pyx
miklos1: Oh, I see, `min`/`max` are swapped...

firedrake/mesh.py
wence-: Consider this mesh:
```
4

3       +----+
        |    |
2       +----+

1  +----+
   |    |
0  +----+
```

The cell layer extents will be:

```
[0, 2, 0, 2]
[2, 4, 2, 4]
```

The interior facet layer extent will be:

```
[0, 4, 2, 2]
```

Because the latter two entries are respectively `max(0, 2)` and min(2, 4)`.  A difference `<=0` implies that the column of entities has a gap in it because the starting iteration layer is above the stop iteration layer.

firedrake/mesh.py
```diff
@@ -319,13 +324,15 @@ class MeshTopology(object):
...
-    def __init__(self, plex, name, reorder, distribute):
+    def __init__(self, plex, name, reorder, distribute, partitioning=None):
         """Half-initialise a mesh topology.
 
         :arg plex: :class:`DMPlex` representing the mesh topology
         :arg name: name of the mesh
         :arg reorder: whether to reorder the mesh (bool)
         :arg distribute: whether to distribute the mesh to parallel processes
+        :kwarg partitioning: provide a cell partition for running in
+            parallel (useful for debugging, but not necessary).
```
miklos1: This could be merged with `distribute`, no?
- `None`: default behaviour
- `False`: do not distribute mesh
- `True`: distribute mesh (automatically)
- otherwise: distribute mesh according to given partition

wence-: Good point.

firedrake/functionspacedata.py
wence-: Also the facet->dof map and various other bits.

firedrake/extrusion_utils.py
wence-: Almost, you need to get the one value from the cell.sub_entities, thanks.

firedrake/functionspacedata.py
wence-: This could be done like so, but it's quite magic.

```patch
diff --git a/firedrake/functionspacedata.py b/firedrake/functionspacedata.py
index 83408aa..bc336aa 100644
--- a/firedrake/functionspacedata.py
+++ b/firedrake/functionspacedata.py
@@ -17,7 +17,7 @@ vs VectorElement) can share the PyOP2 Set and Map data.
 import numpy
 import finat
 from decorator import decorator
-from functools import reduce
+from functools import reduce, partial
 
 from coffee import base as ast
 
@@ -144,11 +144,18 @@ def get_entity_node_lists(mesh, key, entity_dofs, global_numbering, offsets):
     """
     # set->node lists are specific to the sorted entity_dofs.
     cell_node_list = get_cell_node_list(mesh, entity_dofs, global_numbering, offsets)
-    interior_facet_node_list = get_facet_node_list(mesh, "interior_facets", cell_node_list, offsets)
-    exterior_facet_node_list = get_facet_node_list(mesh, "exterior_facets", cell_node_list, offsets)
-    return {mesh.cell_set: cell_node_list,
-            mesh.interior_facets.set: interior_facet_node_list,
-            mesh.exterior_facets.set: exterior_facet_node_list}
+    interior_facet_node_list = partial(get_facet_node_list, mesh, "interior_facets", cell_node_list, offsets)
+    exterior_facet_node_list = partial(get_facet_node_list, mesh, "exterior_facets", cell_node_list, offsets)
+
+    class magic(dict):
+        def __missing__(self, key):
+            return self.setdefault(mesh.interior_facets.set,
+                                   {mesh.interior_facets.set: interior_facet_node_list,
+                                    mesh.exterior_facets.set: exterior_facet_node_list}[key]())
+
+    retval = magic()
+    retval[mesh.cell_set] = cell_node_list
+    return retval
 
 
 @cached
```

docs/source/extruded-meshes.rst
```diff
@@ -202,6 +203,57 @@ constructed by Firedrake.
...
+     |        |
+     x--------x
+
+To simplify the implementation, we never iterate over the interior
+facets that only have cells on one side.  When you construct the mesh,
+you should arrange that these facets have zero area, by squashing the
+coordinates together.  In addition, we require that the resulting
+extruded mesh does not contain topologically disconnected columns:
+offset cells must, at least, share a vertex with some other cell.
+
```
wence-: Done.

firedrake/extrusion_numbering.pyx
```diff
@@ -0,0 +1,1035 @@
...
+        pStart, pEnd = dm.getDepthStratum(i)
+        for p in range(pStart, pEnd):
+            if variable:
+                layers = layer_extents[p, 1] - layer_extents[p, 0]
+                ndof = layers*var_nodes[i, 0] + (layers - 1)*var_nodes[i, 1]
+            else:
+                ndof = const_nodes[i]
+            CHKERR(PetscSectionSetDof(section.sec, p, ndof))
+    section.setUp()
+    return section
```
wence-: Done.

firedrake/extrusion_numbering.pyx
```diff
@@ -0,0 +1,1035 @@
...
+                # we need to offset by the difference from the bottom.
+                if variable:
+                    off += offset[flat_index[k]]*(layer_extents[c, 0] - layer_extents[entity, 0])
+                for j in range(ceil_ndofs[i]):
+                    cell_nodes[cell, flat_index[k]] = off + j
+                    k += 1
+
+    CHKERR(PetscFree(ceil_ndofs))
+    CHKERR(PetscFree(flat_index))
+    return cell_nodes
```
wence-: Done

firedrake/extrusion_numbering.pyx
wence-: Done.

firedrake/functionspaceimpl.py
wence-: The thing that is required is the facet->dof map, I can try refactoring further.

firedrake/output.py
```diff
@@ -113,26 +113,38 @@ def get_topology(coordinates):
...
     else:
-        cell_layers = mesh.layers - 1
+        if mesh.variable_layers:
+            layers = mesh.cell_set.layers_array[:num_cells, ...]
+            cell_layers = layers[:, 1] - layers[:, 0] - 1
 
-    connectivity = numpy.repeat(values, cell_layers, axis=0)
+            def vrange(cell_layers):
+                return numpy.repeat(cell_layers - cell_layers.cumsum(),
+                                    cell_layers) + numpy.arange(cell_layers.sum())
```
wence-: A vectorised `arange`.

firedrake/utility_meshes.py
wence-: Now changed to `distribute`.

firedrake/extrusion_utils.py
wence-: Done.

firedrake/functionspacedata.py
```diff
@@ -81,70 +80,80 @@ def get_node_set(mesh, nodes_per_entity):
...
+    interior_facet_node_list = partial(get_facet_node_list, mesh, "interior_facets", cell_node_list, offsets)
+    exterior_facet_node_list = partial(get_facet_node_list, mesh, "exterior_facets", cell_node_list, offsets)
+
+    class magic(dict):
+        def __missing__(self, key):
+            return self.setdefault(key,
+                                   {mesh.cell_set: lambda: cell_node_list,
+                                    mesh.interior_facets.set: interior_facet_node_list,
+                                    mesh.exterior_facets.set: exterior_facet_node_list}[key]())
+
```
wence-: This is a magic dict that calls the provided function when requesting a missing key.  This way we postpone creating interior_facet_node_list and exterior_facet_node_list until actually required.

wence-:
OK, barring jenkins complaints, I claim this is now good to go.

wence-:
FWIW, cloc says it adds 570 LOC in firedrake and 250 LOC in pyop2, so not soo bad.

Issue #772, opened 2016-05-06
Custom start index for VTK output file numbering


tkarna:
Currently VTK output files always start at index 0, i.e. the first file is `field_0.vtu`. However if you need to continue a previous simulation you should be able to set a custom start index.

I guess we could add a custom start index to `File` constructor. Or add a keyword arg in write method

``` python
File.write(..., index=4)
```

that overrides the default counter, similarly to `time`. This requires that the users keeps track of the counter however.

In terms of continuing a simulation, I'm not sure how to handle the pvd file though. Creating a new `File` object will re-create it, therefore wiping out all the previous time steps, which is the right behavior in most cases. For my applications this is not a major issue though.

wence-:
How about `File("foo.pvd", restart=count)`  Which would open the file for "restarted" output, providing the initial value of the counter.  Now we can just call `File.write(...)` as normal and the correct thing will happen (appending to the end of the existing pvd file).

wence-:
IOW, I prefer the custom start index, rather than an extra argument to the write method.

wence-:
One could even (ugh) do the following: `File(..., restart=True)` and I would open the file, figure out how many `DataSet` entries have been written and then set the counter appropriately.

tkarna:
`File("foo.pvd", restart=count)` sounds like a good idea. I guess in this case you could just append to the pvd file instead of regenerating it, or (maybe better) just writing the `0, ..., count-1` entries there first.

The latter form `File(..., restart=True)` assumes that you want to restart from the last written index which might not be true.

colinjcotter:
I also like the custom start index.

wence-:
Fixed by #1059.

PR #1059, opened 2017-05-02
Allow to restart pvd output from a defined count.

juan42:
This is a work in progress to support restarting and resuming of simulations with the ability to use the same output directories for the pvd files. Cf. #772.

- [x] Custom numbering for restarts.
- [x] Tests

### DIFF
--- a/firedrake/output.py
+++ b/firedrake/output.py
@@ -232,7 +232,7 @@ class File(object):
     _footer = ('</Collection>\n'
                '</VTKFile>\n')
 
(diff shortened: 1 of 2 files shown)


wence-:
Would be nice to have tests (as the PR notes).  Otherwise looks fine I think.  Do you think you can add a test or two?

wence-:
I'm happy to not have the restart from a given timestep (but rather just the minimal functionality).

wence-:
Thanks, I fixed up the writing of the pvd file in the case where the restart was not the last entry, and one more test.

PR #1060, opened 2017-05-02
Switch COFFEE optimisation level to 'Ov'

miklos1:
Alongside with switching TSFC mode to `'coffee'` in firedrakeproject/tsfc#98.


miklos1:
Done.

PR #1061, opened 2017-05-08
reuse function in vertex-based limiter

dorugeber:
assemble centroid calculation RHS into same function

### DIFF
--- a/firedrake/slope_limiter/vertex_based_limiter.py
+++ b/firedrake/slope_limiter/vertex_based_limiter.py
@@ -31,6 +31,7 @@ class VertexBasedLimiter(Limiter):
 
         # Storage containers for cell means, max and mins
         self.centroids = Function(self.P0)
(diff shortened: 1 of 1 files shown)


firedrake/slope_limiter/vertex_based_limiter.py
```diff
@@ -75,8 +76,8 @@ def _update_centroids(self, field):
         """
         Update centroid values
         """
-        b = assemble(TestFunction(self.P0) * field * dx)
-        self.centroid_solver.solve(self.centroids, b)
+        assemble(TestFunction(self.P0) * field * dx, tensor=self.centroids_rhs)
+        self.centroid_solver.solve(self.centroids, self.centroids_rhs)
```
wence-: AKA:
```python
v = TestFunction(self.P0)
assemble((field*v)/CellVolume(v.ufl_domain())*dx, tensor=self.centroids)
```

PR #1063, opened 2017-05-08
Mesh adaptation

taupalosaurus:
Tentative simple mesh adaptation interface. 

This is still work in progress:
 - There are probably many checks required to block use in parallel and for extruded meshes.
 - I haven't started to include interpolation in the code - we need to re-discuss what we want.
 - The test will have to be improved, for now it's mostly intended to run manually to check everything is fine.
 - Some code will be removed when petsc4py is updated.


firedrake/meshadapt.py
```diff
@@ -0,0 +1,47 @@
...
+    vStart, vEnd = plex.getDepthStratum(0)
+    nbrVer = vEnd - vStart
+    
+    dmCoords = mesh.topology._plex.getCoordinateDM()
+    dmCoords.setDefaultSection(coordSection)    
+
+    with mesh.coordinates.dat.vec_ro as coords:
+        mesh.topology._plex.setCoordinatesLocal(coords)
+    with metric.dat.vec_ro as vec:
+    	dmplex.reorder_metric(plex, vec, coordSection)
```
wence-: You asked for a read-only view of the metric, but then update it in place.  Don't do that!  Instead, return a new reordered Vec in `reorder_metric`.

Plausibly for simplification, you could also use the generic reordering to reorder coordinates to send into the DM as well?  Then this would look like

```python
def adapt(mesh, metric):
    assert mesh.coordinates in P1
    assert metric in P1
    with metric.dat.vec_ro as v:
        reordered_metric = to_petsc_numbering(v, metric.function_space())
    with mesh.coordinates.dat.vec_ro as v:
        reordered_coords = to_petsc_numbering(v, mesh.coordinates.function_space())
    dm = mesh._plex
    dm.setCoordinates(reordered_coords)
    newdm = dm.adapt(reordered_metric)
    # distribute=True needs to be set once parallel redistribution works
    return Mesh(newdm, distribute=False)
```

firedrake/meshadapt.py
```diff
@@ -0,0 +1,47 @@
...
+        newplex = dmplex.petscAdapt(mesh.topology._plex, vec)
+
+    newmesh = Mesh(newplex)
+
+    return newmesh
+
+
+
```
wence-: I think the proposed API included a `MeshTransfer` (or similar, we can bike shed the name) object as well.  That way, we can wrap up any optimisations to do with transferring fields behind an interface.

```python
class MeshTransfer(object):
    def __init__(self, donor, target):
         self.donor = donor
         self.target = target
   
     def interpolate(self, donor_field, target_field):
         do_whatever...
```

etc...

miklos1:
Git fail: half of the commits are not recognised as yours; half of the commits are merge commits.

You should probably rebase. Also fix commit authorship and/or GitHub configuration.

firedrake/meshadapt.py
wence-: OK, we want an `Adaptor` that takes an old mesh and has, as an attribute, the new mesh, along with methods for interpolating between the meshes (which takes options to optionally program the interpolation).

wence-:
PETSc/petsc4py pointers updated, please remove the now unnecessary wrappers here.

firedrake/meshadapt.py
taupalosaurus: I tried something in my latest commit. Is is the kind of things we are thinking of ?

dham: Sort of. Let's talk about this in the meeting this afternoon.

taupalosaurus:
Hopefully the new interface looks more like what we discussed.

Things that are yet to do:
  - cleaning/documenting - how do I get rid of the unwanted imports in meshadapt.py ?
  - writing proper tests
  - proper testing (debugging) in parallel
  - modify install scripts

wence-:
> cleaning/documenting - how do I get rid of the unwanted imports in meshadapt.py ?

I don't know what this means.  You just remove the imports for stuff you don't need, no?

firedrake/meshadapt.py
```diff
@@ -0,0 +1,113 @@
...
+class BaseAdaptation(with_metaclass(abc.ABCMeta)):
+    """
+    Abstract top-level class for mesh adaptation
+    """
+
+    def __init__(self, mesh):
+        """
+        """
```
miklos1: Don't waste lines for empty docstrings. Remove it, or write an actual docstring.

firedrake/meshadapt.py
```diff
@@ -0,0 +1,113 @@
...
+        self.mesh = mesh
```
miklos1: Not apparently clear why you need a mesh on the interface class (you don't use that here), but maybe OK.

firedrake/meshadapt.py
```diff
@@ -0,0 +1,113 @@
...
+    @abc.abstractproperty
+    def newmesh(self):
+        pass
+
+    @abc.abstractmethod
+    def transfer_solution(self, f, **kwargs):
+        pass
+
+
+class AAdaptation(BaseAdaptation):
```
miklos1: Unless A-adaptivity is a thing, this looks like a very poor name.

miklos1: OK, docstring says below that this is _anisotropic_ mesh adaptivity, which confirms that this is a poor name. Please type out `Anisotropic` instead.

firedrake/meshadapt.py
```diff
@@ -1,34 +1,67 @@
...
+    Abstract top-level class for mesh adaptation
+    """
+
+    def __init__(self, mesh):
+        """
+        """
+        self.mesh = mesh
+
+    @abc.abstractproperty
+    def newmesh(self):
```
dham: should be `new_mesh` for consistency with the rest of Firedrake.

firedrake/meshadapt.py
```diff
@@ -0,0 +1,113 @@
...
```
miklos1: `newmesh` sounds like a poor name.

firedrake/dmplex.pyx
```diff
@@ -2103,3 +2103,36 @@ def halo_end(PETSc.SF sf, dat, MPI.Datatype dtype, reverse):
...
+    start, end = vec.getOwnershipRange()
+    for p in range(*section.getChart()):
+        dof = section.getDof(p)
+        if dof > 0:
+            off = section.getOffset(p)
+            if off >= 0:
+                for d in range(dof):
+                    for k in range(dim):
+                        oarray[idx] = varray[dim*off + d*off + k - start]
+                        idx += 1
```
wence-: If you want this to compile down to C code, you need to do:

```python
CHKERR(PetscSectionGetDof(section.sec, p, &dof))
CHKERR(PetscSectionGetOffset(section.sec, p, &off))
```

In the appropriate places.

Also, you should type-annotate the numpy arrays a `np.ndarray[PetscReal, mode="c", ndim=1]`

You can run `cython -a dmplex.pyx` and open the resulting `dmplex.html`.  You want the inner loop of your function to not have any yellow background.  Look at how the other functions in this file are annotated.

firedrake/dmplexinc.pxi
```diff
@@ -30,6 +30,8 @@ cdef extern from "petscdmplex.h":
     int DMPlexRestoreTransitiveClosure(PETSc.PetscDM,PetscInt,PetscBool,PetscInt *,PetscInt *[])
     int DMPlexDistributeData(PETSc.PetscDM,PETSc.PetscSF,PETSc.PetscSection,MPI.MPI_Datatype,void*,PETSc.PetscSection,void**)
 
+    int DMPlexAdapt(PETSc.PetscDM, PETSc.PetscVec, const char [], PETSc.PetscDM*);
+
```
wence-: This hunk should go.

firedrake/meshadapt.py
```diff
@@ -0,0 +1,113 @@
...
+        pass
+
+    @abc.abstractmethod
+    def transfer_solution(self, f, **kwargs):
```
miklos1: This `**kwargs` looks very suspicious, especially without a docstring. What is the API, really?

firedrake/meshadapt.py
```diff
@@ -0,0 +1,113 @@
...
+    """
+    Adapt the mesh to a prescribed metric field.
+
+    :arg mesh: the base mesh to adapt
+    :arg metric: a metric tensor field (a Function of a TensorFunctionSpace)
+
+    :return: a new mesh adapted to the metric
+    """
+    adaptor = AAdaptation(mesh, metric)
+    return adaptor.newmesh
```
miklos1: I see, your class is basically a [Poltergeist](https://sourcemaking.com/antipatterns/poltergeists).

tests/test_adapt.py
```diff
@@ -0,0 +1,19 @@
...
+
+metric.interpolate(Constant([[100.,0],[0,100.]]))
+x,y = SpatialCoordinate(mesh)
+metric.interpolate(as_tensor([[1+500*x,0],[0,1+500*y]]))
+
+newmesh = adapt(mesh, metric)
+
+
+f = Function(VectorFunctionSpace(newmesh, 'CG', 1)).interpolate(SpatialCoordinate(newmesh))
+File("mesha.pvd").write(f)
```
wence-: This is still not a test file.

firedrake/meshadapt.py
wence-: This is basically saying that every `Adaptation` needs at least a mesh.

firedrake/meshadapt.py
wence-: Probably with an underscore, maybe `adapted_mesh`?

firedrake/meshadapt.py
dham: No. This adapt function only exists for FEniCS compatibility.

We decided that the adaptation operation needs to be a class because it potentially carries internal state (even if the current very simple implementation does not do so). It's not a Poltergeist.

firedrake/meshadapt.py
wence-: I think we want to take a `*fields` argument, so I can pass any number of fields in.  The `kwargs` are method-specific parameters, which may be different depending on the concrete classes, so we just specify here that we can take arbitrary parameters, and the concrete classes name the ones they actually support.

firedrake/meshadapt.py
```diff
@@ -0,0 +1,113 @@
...
+
+
+class AAdaptation(BaseAdaptation):
+    """
+    Object that performs anisotropic mesh adaptation
+    """
+
+    def __init__(self, mesh, metric):
+        """
+        """
```
wence-: The documentation system we glues together the docstring for `__init__` and the class.  So you can separate them like this, or you can shove it all in `__init__`.

firedrake/meshadapt.py
dham: Transfer operations will have their own parameters. These will depend on what transfer operation is being undertaken. The `**kwargs` here is essentially like `solver_options` on `solve`.

firedrake/meshadapt.py
```diff
@@ -0,0 +1,113 @@
...
+        # PETSc adapt routine expects that the right coordinate section is set
+        # hence the following bloc of code
+        entity_dofs = np.zeros(dim+1, dtype=np.int32)
+        entity_dofs[0] = self.mesh.geometric_dimension()
+        coordSection = plex.createSection([1], entity_dofs, perm=self.mesh.topology._plex_renumbering)
+        dmCoords = plex.getCoordinateDM()
+        dmCoords.setDefaultSection(coordSection)
+
+        with self.mesh.coordinates.dat.vec_ro as coords:
+            plex.setCoordinatesLocal(coords)
```
wence-:
```
with self.mesh.coordinates.dat.vec_ro as coords:
    reordered_coords = dmplex.to_petsc_numbering(coords, self.coordinates.function_space())
plex.setCoordinates(reordered_coords)
```

Not also that this is the *global* coordinate Vec, not the local one.

firedrake/meshadapt.py
```diff
@@ -0,0 +1,113 @@
...
+        with self.metric.dat.vec_ro as vec:
+            reordered_metric = dmplex.to_petsc_numbering(vec, self.metric.function_space())
+
+        newplex = plex.adapt(reordered_metric)
+        new_mesh = Mesh(newplex)
+        return new_mesh
```
wence-: AKA `return Mesh(plex.adapt(reordered_metric))`

But I note that you appear to be missing the label for the boundary.  Is that intentional?

firedrake/meshadapt.py
```diff
@@ -0,0 +1,113 @@
...
+import firedrake.utils as utils
+import firedrake.functionspace as functionspace
+import firedrake.function as function
+import firedrake.mesh as fmesh
+
+
+__all__ = ['adapt']
+
+
+class BaseAdaptation(with_metaclass(abc.ABCMeta)):
```
wence-: FWIW, I prefer if the bikeshed is painted `BaseAdaptor` and `AnisotropicAdaptor`, rather than `BaseAdaptation` and `AnisotropicAdaptation`.

firedrake/meshadapt.py
```diff
@@ -0,0 +1,113 @@
...
+
+from firedrake.mesh import Mesh
+import firedrake.dmplex as dmplex
+import firedrake.utils as utils
+import firedrake.functionspace as functionspace
+import firedrake.function as function
+import firedrake.mesh as fmesh
+
+
+__all__ = ['adapt']
```
wence-: You want to expose `AAdaptation` (or whatever it is renamed to) as well here.

firedrake/meshadapt.py
```diff
@@ -0,0 +1,113 @@
...
+        if f.ufl_element().family() == 'Lagrange' and f.ufl_element().degree() == 1:
+            fnew.dat.data[:] = f.at(self.newmesh.coordinates.dat.data)
+        elif f.ufl_element().family() == 'Lagrange':
+            degree = f.ufl_element().degree()
+            C = functionspace.VectorFunctionSpace(self.newmesh, 'CG', degree)
+            interp_coordinates = function.Function(C)
+            interp_coordinates.interpolate(self.newmesh.coordinates)
+            fnew.dat.data[:] = f.at(interp_coordinates.dat.data)
+        else:
+            raise NotImplementedError("Can only interpolate CG fields")
```
wence-: The point evaluation machinery can evaluate any function, but the problem is putting it back into basis coefficients for vector-valued spaces, right?  In any case, this function works equally well with DG fields.

So maybe you want to check if `f.ufl_element().mapping() == "identity"`.

Also, I would not special-case P1.  Why not just write:

```python
element = f.ufl_element()
if element.mapping != "identity":
    raise NotImplementedError("Can only transfer affine mapped functions")
if type(element) is ufl.MixedElement:
    raise NotImplementedError("Cannot transfer mixed functions, try transfering f.split()")
if element.value_shape(): # vector/tensor element
    element = element.sub_elements()[0]
element = VectorElement(element)
C = functionspace.FunctionSpace(self.newmesh, element)

points = function.Function(C)
points.interpolate(SpatialCoordinate(self.newmesh))

fnew.dat.data[:] = f.at(points.dat.data_ro)
```

Untested.

firedrake/meshadapt.py
```diff
@@ -0,0 +1,113 @@
...
+
+        newplex = plex.adapt(reordered_metric)
+        new_mesh = Mesh(newplex)
+        return new_mesh
+
+    def transfer_solution(self, f, method=None):
+        """
+        Transfers a solution field from the old mesh to the new mesh
+
+        :arg f: function defined on the old mesh that one wants to transfer
```
wence-: Document what the `method` argument does.

wence-:
You need to add `--download-pragmatic` to the list of minimal petsc options in `firedrake-install` for the test suite to pass.

firedrake/meshadapt.py
miklos1: This reminds me... `Base` in class names is usually used as a suffix rather than a prefix. The common prefix word is `Abstract`. That is, `AdaptorBase` or `AbstractAdaptor` feels familiar, but `BaseAdaptor` feels odd. C# also has convention of using an `I` prefix in interfaces (e.g. `IAdaptor`), but we don't do that here.

firedrake/meshadapt.py
taupalosaurus: The change you suggest does not work (in serial, the mesh transfered to pragmatic is wrong) whereas current version works. I am not entirely sure I understand why your version would be better ?

firedrake/meshadapt.py
taupalosaurus: So, I do not know what the vertex-wise value of a DG field means - if we do what you suggest for DG fields, we basically turn the DG field into a CG one, no ?
I am not a big fan of putting such a bad interpolation method in Firedrake - but if everybody else think we should do it, so be it.

wence-:
Closing for now, since this does not seem like it will ever be ready.

PR #1064, opened 2017-05-09
save cubed-sphere radius

dorugeber:
Extending 2092c15, and minor variable rename


dorugeber:
In the mg code, Lawrence is using the attribute to work out which mesh he's on, so you would need to find another way to do that.

miklos1:
OK, then.

wence-:
I think that different name is only being used to compute `gdim`.

dorugeber:
Feel free to fixup.

PR #1069, opened 2017-05-11
Install updates

wence-:
Moves eigen download to be managed ourselves.

Also:

- remove framework munging on macos
- remove travis/drone hacks
- includes output of running `env`
- always produces the log file (so we don't have to re-run on failure)
- uses `directory` context manager more.

### DIFF
--- a/scripts/firedrake-install
+++ b/scripts/firedrake-install
@@ -693,7 +693,7 @@ def build_and_install_petsc():
 
 def build_and_install_h5py():
     import shutil
-    log.info("Installing h5py\n")
+    log.info("Installing h5py")
     # Clean up old downloads
     if os.path.exists("h5py-2.5.0"):
         log.info("Removing old h5py-2.5.0 source")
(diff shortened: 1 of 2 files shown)


scripts/firedrake-install
```diff
@@ -790,6 +663,33 @@ def pip_uninstall(package):
             raise InstallError("pip claims it uninstalled %s more than 10 times.  Something is probably broken.", package)
 
 
+def build_and_install_petsc():
+    import hashlib
+    tarball = "eigen-3.3.3.tgz"
+    url = "http://bitbucket.org/eigen/eigen/get/3.3.3.tar.gz"
```
miklos1: Let's hope we won't need a `pip uninstall h5py` style download loop...

scripts/firedrake-install
```diff
@@ -790,6 +663,33 @@ def pip_uninstall(package):
...
+    sha = hashlib.sha256()
+    expect = "94878cbfa27b0d0fbc64c00d4aafa137f678d5315ae62ba4aecddbd4269ae75f"
+    if not os.path.exists(tarball):
+        log.info("Downloading Eigen from '%s' to '%s'" % (url, tarball))
+        check_call(["curl", "-L", "-o", tarball, url])
```
miklos1: `curl` is not really a documented dependency yet, is it? Doesn't Python have some downloader in its standard library?

scripts/firedrake-install
```diff
@@ -790,6 +663,33 @@ def pip_uninstall(package):
...
+    else:
+        log.info("Eigen tarball already downloaded to '%s'" % tarball)
+    log.info("Checking Eigen tarball integrity")
+    with open(tarball, "rb") as f:
+        while True:
+            data = f.read(65536)
```
miklos1: Magic number?

scripts/firedrake-install
```diff
@@ -790,6 +663,33 @@ def pip_uninstall(package):
...
+            if not data:
+                break
+            sha.update(data)
```
miklos1: There ought to be a more concise way to say this loop, but perhaps not as memory efficient?

wence-: well, I know that the file is only a couple of megs, so I could just `sha.update(f.read())`, but meh.

scripts/firedrake-install
wence-: 64k

scripts/firedrake-install
wence-: I think that we will want to rewrite this stuff in the future.  But note that if I just download a newer version (by changing the url, tarball name and expected hash) then all that happens is I leave the old guy around.

scripts/firedrake-install
dorugeber: Heh.  The download instructions are to `curl` the install script, then run it, so the user *probably* has it by this point.  However, it's true that `curl` is _not_ part of a default Ubuntu 16.04 install.

miklos1: Some people just use `wget` instead of `curl`.

PR #1070, opened 2017-05-11
Generate point evaluation code with the help of TSFC

miklos1:
This only affects the generation of that code which evaluates a function at a given reference point. The new code generation will employ TSFC so that this will generalises for the point evaluation of UFL expressions, not just single functions. The point evaluation of UFL expressions, however, will not be available yet with this pull request, since:

1. No API exists for that functionality.
2. Data structures and wrappers cannot deal with more than one coefficient at the moment.

I would like to note that point 2 would not be an issue for things like point evaluations of derivatives of functions.

Immediate advantages:
- Better code generation for "structured" elements thanks to FInAT.
Same code generation for all other elements.
- Piola transformation applied correctly on nonaffine geometries. Current `master` applies them with an affine approximation which calculates the Jacobian at the right place, but assumes that the second derivative of the coordinate space is zero.

Regressions:
- The Jacobian calculated during the Newton iteration that finds the reference coordinates is no longer reused.

Requires FInAT/FInAT#28 and firedrakeproject/tsfc#121.

### DIFF
--- a/firedrake/pointeval_utils.py
+++ b/firedrake/pointeval_utils.py
@@ -1,10 +1,12 @@
 from __future__ import absolute_import, print_function, division
-from six.moves import map, range
+from six.moves import map
 
 import collections
 import numpy
 import sympy
 from pyop2.datatypes import IntType, as_cstr
+
+from coffee import base as ast
 
 
 def operands_and_reconstruct(expr):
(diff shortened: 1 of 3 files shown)


firedrake/pointeval_utils.py
```diff
@@ -1,247 +1,115 @@
...
+    assert coordinates.ufl_domain() == domain
+    expression = tsfc.ufl_utils.replace_coordinates(expression, coordinates)
+
+    # Initialise kernel builder
+    builder = firedrake_interface.KernelBuilderBase()
+    x_arg = builder._coefficient(coordinates, "x")
+    f_arg = builder._coefficient(coefficient, "f")
+
+    # Point evaluation of mixed coefficients not supported here
+    # expression = ufl_utils.split_coefficients(expression, builder.coefficient_split)
```
wence-: Raise `NotImplementedError`?

miklos1:
A follow-up could be to ask TSFC to generate the code for updating the guess in the Newton iteration when doing mapping from physical to reference coordinates. That would allow us to finally get rid of ~`codesnippets.py`~ `firedrake_geometry.h`, and also the custom evaluation code generator from the FIAT element.

PR #1071, opened 2017-05-11
Allow one to specify boundary conditions when using the Projector

thomasgibson:
It would be nice to be able to specify any strong boundary conditions on the function space you want to project into using the ``Projector``. This PR allows one to do just that. Have a look at it to make sure the approach is correct.

### DIFF
--- a/firedrake/projection.py
+++ b/firedrake/projection.py
@@ -38,8 +38,7 @@ def project(v, V, bcs=None, mesh=None,
     then ``v`` is projected into a new :class:`.Function` and that
     :class:`.Function` is returned.
 
-    The ``bcs``, ``mesh`` and ``form_compiler_parameters`` are
-    currently ignored."""
+    The ``mesh`` and ``form_compiler_parameters`` are currently ignored."""
     from firedrake import function
 
     if isinstance(V, functionspaceimpl.WithGeometry):
@@ -113,6 +112,8 @@ class Projector(object):
     :arg v: the :class:`ufl.Expr` or
          :class:`.Function` to project
     :arg v_out: :class:`.Function` to put the result in
+    :arg bcs: an optional set of :class:`.DirichletBC` objects to apply
+              on the target function space.
     :arg solver_parameters: parameters to pass to the solver used when
          projecting.
     """
@@ -133,12 +134,10 @@ class Projector(object):
             except TypeError:
                 bcs = (bcs,)
             if any(bc.function_space() != v_out.function_space() for bc in bcs):
-                raise ValueError(
-                    "bcs must be enforced on the space of the result function."
-                )
+                raise ValueError("bcs must be enforced on the space of the result function.")
         self.bcs = bcs
 
-        if not self._same_fspace:
+        if not self._same_fspace or self.bcs:
             V = v_out.function_space()
 
             p = ufl_expr.TestFunction(V)
@@ -147,7 +146,7 @@ class Projector(object):
             a = ufl.inner(p, q)*ufl.dx
             L = ufl.inner(p, v)*ufl.dx
 
-            problem = vs.LinearVariationalProblem(a, L, v_out,
+            problem = vs.LinearVariationalProblem(a, L, v_out, bcs=self.bcs,
                                                   constant_jacobian=constant_jacobian)
 
             if solver_parameters is None:
@@ -162,10 +161,7 @@ class Projector(object):
         """
         Apply the projection.
         """
-        if self._same_fspace:
+        if self._same_fspace and not self.bcs:
             self.v_out.assign(self.v)
         else:
             self.solver.solve()
-        if self.bcs:
-            for bc in self.bcs:
-                bc.apply(self.v_out)
--- a/firedrake/projection.py
+++ b/firedrake/projection.py
@@ -128,13 +128,6 @@ class Projector(object):
                              v_out.function_space())
         self.v = v
         self.v_out = v_out
-        if bcs:
-            try:
-                bcs = tuple(bcs)
-            except TypeError:
-                bcs = (bcs,)
-            if any(bc.function_space() != v_out.function_space() for bc in bcs):
-                raise ValueError("bcs must be enforced on the space of the result function.")
         self.bcs = bcs
 
         if not self._same_fspace or self.bcs:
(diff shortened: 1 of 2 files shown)


tests/regression/test_projection.py
```diff
@@ -249,28 +249,38 @@ def test_projector_bcs(tensor):
...
-                                           "42.0", "42.0")),
-                         (1, 2, 3, 4))
-
-    exact = Function(V_ho)
-    bc.apply(exact)
+        bcs = [DirichletBC(V_ho, Expression(("0.5", "0.5",
+                                             "0.5", "0.5")), (1, 3)),
+               DirichletBC(V_ho, Expression(("-0.5", "-0.5",
+                                             "-0.5", "-0.5")), (2, 4))]
+        expr = Expression(['cos(x[0]*pi*2)*sin(x[1]*pi*2)']*4)
```
miklos1: I think it would be a bad idea to go around and replace existing uses of `Expression`s in the test suite, because then we would lose test coverage for that feature. However, I suggest to use UFL expressions in all new code, including test code, primarily because they are easier to get right because of the more helpful error messages.

wence-: I agree, don't fix old code everywhere, but don't introduce new "bad" code.

thomasgibson: Okay I will update the test

tests/regression/test_projection.py
```diff
@@ -246,29 +246,35 @@ def test_projector_expression():
...
-        expr = Expression('cos(x[0]*pi*2)*sin(x[1]*pi*2)')
+        fct = cos(x[0]*pi*2)*sin(x[1]*pi*2)
+        v = Function(V).interpolate(fct)
     elif tensor == 'vector':
         V = VectorFunctionSpace(mesh, "CG", 1)
         V_ho = VectorFunctionSpace(mesh, "CG", 5)
-        bcs = [DirichletBC(V_ho, Expression(("0.5", "0.5")), (1, 3)),
-               DirichletBC(V_ho, Expression(("-0.5", "-0.5")), (2, 4))]
-        expr = Expression(['cos(x[0]*pi*2)*sin(x[1]*pi*2)']*2)
+        bcs = [DirichletBC(V_ho, Constant((0.5, 0.5)), (1, 3)),
```
miklos1: Do you really need to make it a `Constant`, can't you just use the tuple?

thomasgibson: If I just use tuples, I get a `ValueError` telling me that I have not provided "a valid DirichletBC expression", which is why I use `Constant`/UFL expressions. Can you give me an example of what you exactly mean? I may just not be using the right input.

miklos1: OK, then.

firedrake/projection.py
```diff
@@ -127,8 +128,16 @@ def __init__(self, v, v_out, solver_parameters=None, constant_jacobian=True):
...
+        if bcs:
+            try:
+                bcs = tuple(bcs)
+            except TypeError:
+                bcs = (bcs,)
+            if any(bc.function_space() != v_out.function_space() for bc in bcs):
+                raise ValueError("bcs must be enforced on the space of the result function.")
+        self.bcs = bcs
+
+        if not self._same_fspace or self.bcs:
```
wence-: If they are the same function space and you have bcs, is it not the case that the "right" thing is:

`out.assign(in); bc.apply(out)` ?

firedrake/projection.py
```diff
@@ -127,8 +128,16 @@ def __init__(self, v, v_out, solver_parameters=None, constant_jacobian=True):
...
         self.v_out = v_out
-
-        if not self._same_fspace:
+        if bcs:
+            try:
+                bcs = tuple(bcs)
+            except TypeError:
+                bcs = (bcs,)
+            if any(bc.function_space() != v_out.function_space() for bc in bcs):
+                raise ValueError("bcs must be enforced on the space of the result function.")
```
wence-: Do we not do this kind of checking inside `LinearVariationalSolver`?

wence-: Yeah, we do.

firedrake/projection.py
thomasgibson: I thought about that. But I wasn't sure if it's really the right thing to do. Wouldn't exterior BCs influence the function values between the interior and exterior nodes? I'll experiment with it and update if it turns out to be the right thing.

wence-: No, because the projection is just the identity, right?  Especially in the case where the two function spaces are the same, you're just saying "copy these values, except on the boundary where you use the boundary values"

dorugeber: No, I think @thomasgibson is right.  If you have a P1 function space in 1D, and you solve <u, v> = <0, v>, u = 1 on the boundary, you don't get 1, 0, 0, 0, ..., 0, 0, 1, but rather 1, -, +, -, ..., +, -, 1

dorugeber: (the ..., 0, 0, 1 solution can't be right, as <u, v> != 0 for the test function v one node in from the boundary)

For some reason I was thinking about this in 2015 as well; not the first time it's come up.

wence-: Oh, yeah, sorry.

wence-: (I hate boundary conditions...)

thomasgibson: (You and me both...)

tests/regression/test_projection.py
```diff
@@ -243,6 +243,57 @@ def test_projector_expression():
...
+    projector = Projector(v, ret, bcs=bcs, solver_parameters={"ksp_type": "preonly",
+                                                              "pc_type": "lu"})
+    projector.project()
+
+    ref = Function(V_ho)
+    # The method "project" already handles bcs.
+    # So we use this as a reference to make sure the
+    # projector does the same thing (up to machine precision).
+    project(v, ref, bcs=bcs, solver_parameters={"ksp_type": "preonly",
+                                                "pc_type": "lu"})
```
wence-: This is going to be a bad test soon, because I would like to replace the internals of `project` with the construction of a `Projector`...

thomasgibson: Okay, I can just solve a Galerkin projection problem instead if that's what you had in mind.

Issue #1068, opened 2017-05-10
nullspace in hybridisation PC not correct


wence-:
This appears not to be right in a bunch of cases (at the very least `PeriodicSquareMesh`, icosahedral meshes and cubed sphere meshes).  But it's most noticeable on cubed sphere.

There are two things that are going wrong:

1. The created nullspace vectors do not have l2 norm 1.
2. In the cubed sphere case, the created nullspace vector is not constant (as I expect it should be)

As a result, the nullspace is invalid (which may explain why you're not getting useful results in some cases).

Here's some test code:

```python
from firedrake import *
from firedrake.slate.preconditioners import create_schur_nullspace
import numpy as np

mesh = UnitCubedSphereMesh(2)
mesh.init_cell_orientations(SpatialCoordinate(mesh))

n = FacetNormal(mesh)

V = FunctionSpace(mesh, "RTCF", 1)
Q = FunctionSpace(mesh, "DG", 0)

W = V*Q
sigma, u = TrialFunctions(W)
tau, v = TestFunctions(W)
a = (inner(sigma, tau) + div(sigma)*v + div(tau)*u)*dx

W_d = FunctionSpace(mesh, MixedElement([BrokenElement(Vi.ufl_element()) for Vi in W]))

atilde = Tensor(replace(a, dict(zip(a.arguments(), (TestFunction(W_d), TrialFunction(W_d))))))

Vt = FunctionSpace(mesh, "HDiv Trace", 0)
gamma = TestFunction(Vt)

sigma, _ = TrialFunctions(W_d)

K = Tensor(gamma('+') * dot(sigma, n) * dS)

A = assemble(a, mat_type="aij")
nullspace = MixedVectorSpaceBasis(W, [W[0], VectorSpaceBasis(constant=True)])
nullspace._build_monolithic_basis()
A.petscmat.setNullSpace(nullspace._nullspace)

from firedrake.slate.preconditioners import create_schur_nullspace
Snullsp = create_schur_nullspace(A.petscmat, K * atilde.inv,
                                 W, W_d, Vt,
                                 COMM_WORLD)

v = Snullsp.getVecs()[0]
print "Computed nullspace of S (min, max, norm)", v.array_r.min(), v.array_r.max(), v.norm()

S = K * atilde.inv * K.T

u, s, v = np.linalg.svd(assemble(S, mat_type="aij").M.values)

singular_vector = v[-1]
[...]

PR #1073, opened 2017-05-17
Fix null space computation for hybridization

thomasgibson:
This addresses the bug in https://github.com/firedrakeproject/firedrake/issues/1068. Previously I thought that the null space basis vectors could be generated by applying the same forward eliminations to each basis vector. However, this turns out to be incorrect.

Instead, I arrived at an expression for the nullspace vectors for the Schur complement operator in terms of the null space vectors of the original operator (projected into the broken space):
```
A * x_z + K.T * lambda_z = 0
K * x_z                  = 0
```
Using the first equation, this gives us `lambda_z = -(K * K.T).inv * K * A * x_z` (obtained by multiplying the first equation by `K`), so I instead pass the operator `-(K * K.T).inv * K * A` to generate the null space vectors, then normalize.  This gives the expected constant vectors with norm 1 in the example https://github.com/firedrakeproject/firedrake/issues/1068. It also gives the norm-1 constant vectors with the right magnitude for isocahedral sphere meshes.

Please have a look to make sure everything is sound.

### DIFF
--- a/firedrake/slate/preconditioners.py
+++ b/firedrake/slate/preconditioners.py
@@ -142,7 +142,7 @@ class HybridizationPC(PCBase):
         Smat = self.S.petscmat
 
         # Nullspace for the multiplier problem
-        nullspace = create_schur_nullspace(P, K * Atilde.inv,
+        nullspace = create_schur_nullspace(P, -(K * K.T).inv * K * Atilde,
                                            V, V_d, TraceSpace,
                                            pc.comm)
         if nullspace:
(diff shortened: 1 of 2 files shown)


firedrake/slate/preconditioners.py
```diff
@@ -142,7 +142,7 @@ def initialize(self, pc):
         Smat = self.S.petscmat
 
         # Nullspace for the multiplier problem
-        nullspace = create_schur_nullspace(P, K * Atilde.inv,
+        nullspace = create_schur_nullspace(P, -(K * K.T).inv * K * Atilde,
```
thomasgibson: The operator `K` exhibits a semi-orthogonality property (they're not square). Numerically `K * K.T` is producing a scalar multiple of the identity. If I omit the local tensor inversion in this expression, I get the same null space vectors. So perhaps it's safe to leave it as `-K * Atilde`?

wence-: I think that should be OK.  Can you add some tests?

thomasgibson: Sure, I'll update

thomasgibson:
When you're happy with this, let me know and I'll merge tomorrow

PR #1075, opened 2017-05-19
Remove remaining sticky FFC bits

miklos1:
No more `firedrake_geometry.h` and other non-sense. Switch generation of the Newton update code in the physical to reference coordinates mapping to TSFC-based tools.


firedrake/pointquery_utils.py
```diff
@@ -286,90 +74,84 @@ def X_isub_dX(topological_dimension):
...
+        # Translate to COFFEE
+        ir = impero_utils.preprocess_gem(ir)
+        return_variable = gem.Variable('dX', (dim,))
+        assignments = [(gem.Indexed(return_variable, (i,)), e)
+                       for i, e in enumerate(ir)]
+        impero_c = impero_utils.compile_gem(assignments, ())
+        body = tsfc.coffee.generate(impero_c, {}, parameters["precision"])
+        body.open_scope = False
+
+        return body
```
miklos1: This is the third place I see something like this, which suggests a potential refactoring opportunity at some later occasion...

Issue #1074, opened 2017-05-17
SLATE compilation error


stephankramer:
I am trying to use SLATE to implement a preconditioner for a Schur system A11-A10 A00^{-1} A01, where A00 is a DG mass matrix that assembles the Schur system explicitly (on @wence- 's suggestion here: https://github.com/thetisproject/thetis/pull/97#discussion_r116945058). The relevant code is:

        mass = Tensor(dot(v, u)*dx)
        fs = dict(formmanipulation.split_form(a))
        A01 = Tensor(fs[(0, 1)])
        A10 = Tensor(fs[(1, 0)])
        A11 = Tensor(fs[(1, 1)])
        S = A11 - A10*mass.inv*A01

        Smat = assemble(S, form_compiler_parameters=ctx.fc_params)
        Smat.force_evaluation()

here `a` is the form of the coupled SWE system (only linear terms). This fails in the last step with a compilation error because the generated code contains a subroutine:

    static inline void compile_slate (double  A4[6][6] , double **  coords , double *  w_0 , double **  w_1 , double **  w_2_0 , double **  w_2_1 , double *  w_3 , double **  w_4_0 , double **  w_6 , double **  w_5 , double **  w_6 , double **  w_7 , double *  w_8 )

which contains a `w_6` argument twice.

Unfortunately I haven't been able to reproduce in a standalone test yet. To reproduce in thetis, checkout the assembledschur-slate branch of thetis, and run `py.test -k pressurep -svx test/swe2d/test_standing_wave.py`.

thomasgibson:
I can reproduce this and I'll have a close look at it. Thanks for letting us know

thomasgibson:
Just an update: I believe I fixed the issue in https://github.com/firedrakeproject/firedrake/tree/slate-coefficient-fix.

I had to make a minor change in `thetis/assembledschur.py` because `cxt` was being called out of scope. Making `cxt` a class attribute fixed that. With my firedrake branch, the test seems to be passing.

I will clean up the branch and submit a PR into Firedrake master soon!

thomasgibson:
Fixed in https://github.com/firedrakeproject/firedrake/pull/1077.

PR #1077, opened 2017-05-23
Fix mixed coefficient handling in Slate

thomasgibson:
This PR addresses a bug described in https://github.com/firedrakeproject/firedrake/issues/1074. The changes here corrects how Slate handles mixed coefficients by providing a mapping from coefficient to it's split components.

### DIFF
--- a/firedrake/slate/slac/compiler.py
+++ b/firedrake/slate/slac/compiler.py
@@ -15,6 +15,7 @@ all low-level numerical linear algebra operations are performed using
 this templated function library.
 """
 from __future__ import absolute_import, print_function, division
+from six import iteritems
 from six.moves import range
 
 from coffee import base as ast
@@ -80,9 +81,6 @@ def compile_expression(slate_expr, tsfc_parameters=None):
     # simply reuse the produced kernel.
     if slate_expr._metakernel_cache is not None:
         return slate_expr._metakernel_cache
-
-    # Initialize coefficients, shape and statements list
-    expr_coeffs = slate_expr.coefficients()
 
     # We treat scalars as 1x1 MatrixBase objects, so we give
     # the right shape to do so and everything just falls out.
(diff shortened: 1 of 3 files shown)


firedrake/slate/slac/compiler.py
```diff
@@ -246,12 +244,12 @@ def compile_expression(slate_expr, tsfc_parameters=None):
...
 
     # Coefficient information
-    for c in expr_coeffs:
+    for c, csym_info in iteritems(builder.coefficient_map):
         if isinstance(c, Constant):
             ctype = "%s *" % SCALAR_TYPE
         else:
             ctype = "%s **" % SCALAR_TYPE
-        args.extend([ast.Decl(ctype, csym) for csym in builder.coefficient(c)])
+        args.extend([ast.Decl(ctype, csym) for csym in csym_info])
```
wence-: So previously you pulled the coefficients out of the expression and then called `builder.coefficient` here.  Now you make the coefficient_map in the constructor of the kernel builder and then iterate over that.  But now `coefficient_map` is a public API member.  Why not iterate over the coefficients of the expression and call `builder.coefficient` as before?

firedrake/slate/slac/kernel_builder.py
```diff
@@ -42,8 +42,18 @@ def __init__(self, expression, tsfc_parameters=None):
...
+            if type(coefficient.ufl_element()) == MixedElement:
+                csym_info = []
+                for j, _ in enumerate(coefficient.split()):
+                    csym_info.append(ast.Symbol("w_%d_%d" % (i, j)))
+            else:
+                csym_info = (ast.Symbol("w_%d" % i),)
+
+            coefficient_map[coefficient] = tuple(csym_info)
+
+        self.coefficient_map = coefficient_map
```
wence-: If you don't want to pollute the constructor with this code, but don't want the `prepare_coefficients` free function (I agree it only appears here, so it makes sense to inline), you could make `coefficient_map` a `@cached_property` or `@property`, probably the former.

firedrake/slate/slac/kernel_builder.py
```diff
@@ -139,8 +145,11 @@ def _finalize_kernels_and_update(self):
 
         for splitkernel in splitkernels:
             oriented = oriented or splitkernel.kinfo.oriented
+
```
wence-: Why?

thomasgibson:
Updates pushed

tests/slate/test_assemble_tensors.py
```diff
@@ -100,17 +100,12 @@ def test_assemble_matrix_into_tensor(mesh):
...
     f = Function(W)
+    g, h = f.split()
     f.assign(1)
-    u = TrialFunction(V)
-    v = TestFunction(V)
-    T = Tensor(f[0] * u * v * dx)
-    ref = assemble(f[0] * u * v * dx)
-
-    assert np.allclose(assemble(T).M.values, ref.M.values)
+    assert np.allclose(assemble(Tensor((g + f[0] + h + f[1])*dx)), 4.0)
```
wence-: Oh, I hadn't spotted that there was already a test.  I guess you could have kept the existing one and added this?

thomasgibson:
Unless I missed something, I believe this is ready to go.

PR #1079, opened 2017-05-23
Mg transfer fix

wence-:
Fixes bug reported on mailing list by Wei Pan.


miklos1:
What is changing here?

wence-:
Better?

miklos1:
Yes.

PR #1081, opened 2017-05-30
Appropriate handling of the HDiv residual when using hybridization

thomasgibson:
This PR fixes a problem when hybridizing a mixed problem with a nontrivial expression in the HDiv part of the right-hand side.

The problem before was that the incoming HDiv residual in the `apply` method of the preconditioner treated the data as a vector. The residual at this stage is, in fact, a covector with test functions. A typical projection is NOT the right approach in this case. This PR adds a new `ksp` which numerically inverts an HDiv-mass matrix system in order to construct the "broken" HDiv residual. It functions like any other `ksp` and can be tuned via `solver_parameters` (as demonstrated in the testing suite).

I also changed the naming of a few variables for clarity purposes.


firedrake/slate/preconditioners.py
```diff
@@ -149,27 +183,19 @@ def initialize(self, pc):
...
-        data_params = {"ksp_type": "preonly",
-                       "pc_type": "bjacobi",
-                       "sub_pc_type": "ilu"}
-        self.data_projector = Projector(self.unbroken_rhs.split()[self.vidx],
-                                        self.broken_rhs.split()[self.vidx],
-                                        solver_parameters=data_params)
-
         # NOTE: Tolerance is very important here and so we provide
         # the user a way to specify projector tolerance
         opts = PETSc.Options()
```
wence-: Can we, while we're here, set this projector up as a KSP, so we can control all the options?  Or else, add an `options_prefix` setting to `Projector`/`project`.

firedrake/slate/preconditioners.py
```diff
@@ -251,22 +276,35 @@ def apply(self, pc, x, y):
...
+
         # Compute the rhs for the multiplier system
         self._assemble_Srhs()
 
         # Solve the system for the Lagrange multipliers
         with self.schur_rhs.dat.vec_ro as b:
-            with self.trace_solution.dat.vec as x:
-                self.ksp.solve(b, x)
+            with self.trace_solution.dat.vec as x_trace:
+                self.trace_ksp.solve(b, x_trace)
```
wence-: You should add viewer calls to the other configurable objects: to whit the forward dual->primal KSP and the reconstruction project call.

firedrake/slate/preconditioners.py
wence-: Although if you're going to replace this step with a Fortin operator, maybe don't both for now.

firedrake/slate/preconditioners.py
```diff
@@ -89,10 +89,44 @@ def initialize(self, pc):
...
+
+        # Set up the KSP for the hdiv residual projection
+        hd_ksp = PETSc.KSP().create(comm=pc.comm)
+        hd_ksp.setOptionsPrefix(prefix + "hdiv_residual_")
+
+        # HDiv mass operator
+        p = TrialFunction(V[self.vidx])
+        q = TestFunction(V[self.vidx])
+        mass = ufl.dot(p, q)*ufl.dx
+        # TODO: Bcs?
```
thomasgibson: I agree, something will probably need to be done. I'll leave this for another PR, as I currently raise a `NotImplementedError` in the presence of strong BCs. Once I hook up BCs, I think during assemble, we simply need to pass through `bcs=self.cxt.row_bcs`. Does that seem reasonable?

wence-: it sounds reasonable, I don't know if it's *right*.

firedrake/slate/preconditioners.py
thomasgibson: Done in one of my recently pushed commits

PR #1086, opened 2017-06-15
Vec wo

wence-:
Use the new `vec_wo` context manager for cases where we know we're not going to read from the vec (especially, for example, in residual evaluation).

Needs OP2/PyOP2#527.

### DIFF
--- a/firedrake/linear_solver.py
+++ b/firedrake/linear_solver.py
@@ -147,7 +147,11 @@ class LinearSolver(solving_utils.ParametersMixin):
             b = b_bc
         with self.inserted_options():
             with b.dat.vec_ro as rhs:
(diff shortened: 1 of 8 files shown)


firedrake/linear_solver.py
```diff
@@ -147,7 +147,11 @@ def solve(self, x, b):
             b = b_bc
         with self.inserted_options():
             with b.dat.vec_ro as rhs:
-                with x.dat.vec as solution:
+                if self.ksp.getInitialGuessNonzero():
+                    acc = x.dat.vec
+                else:
+                    acc = x.dat.vec_wo
+                with acc as solution:
```
wence-: We only read from this vec if the KSP is using a non-zero initial guess.

PR #1089, opened 2017-06-16
Fix extruded bcs

### LINKED ISSUE #1088: Buggy application of strong boundary conditions on extruded meshes
Consider the following, contrived, example:

```python
mesh = UnitIntervalMesh(3)
extmesh = ExtrudedMesh(mesh, layers=2)
u = TrialFunction(V)
v = TestFunction(V)
a = u('-')*v('-')*dS_h
bcs = DirichletBC(V, 0, "top")

A = assemble(a, bcs=bcs).M.values

A[:, bcs.nodes]
=> 
array([[ 0.        ,  0.        ,  0.        ,  0.        ],
       [ 0.        ,  0.        ,  0.        ,  0.        ],
       [ 1.        ,  0.        ,  0.        ,  0.        ],
       [ 0.        ,  0.        ,  0.        ,  0.        ],
       [ 0.        ,  0.08333333,  0.        ,  0.        ],
       [ 0.        ,  1.        ,  0.        ,  0.        ],
       [ 0.        ,  0.        ,  0.        ,  0.        ],
       [ 0.        ,  0.        ,  0.08333333,  0.        ],
       [ 0.        ,  0.        ,  1.        ,  0.        ],
       [ 0.        ,  0.        ,  0.        ,  0.        ],
       [ 0.        ,  0.        ,  0.        ,  0.        ],
       [ 0.        ,  0.        ,  0.        ,  1.        ]])

A[bcs.nodes, :]
=>
array([[ 0.        ,  0.        ,  1.        ,  0.        ,  0.        ,  0.        ,  0.        ,  0.        ,  0.        ,  0.        ,  0.        ,  0.        ],
       [ 0.        ,  0.        ,  0.        ,  0.        ,  0.08333333,  1.        ,  0.        ,  0.        ,  0.        ,  0.        ,  0.        ,  0.        ],
       [ 0.        ,  0.        ,  0.        ,  0.        ,  0.        ,  0.        ,  0.        ,  0.08333333,  1.        ,  0.        ,  0.        ,  0.        ],
       [ 0.        ,  0.        ,  0.        ,  0.        ,  0.        ,  0.        ,  0.        ,  0.        ,  0.        ,  0.        ,  0.        ,  1.        ]])
```

This is because the determination of the top and bottom masks doesn't "double up" the cell masks for the interior facet integrals:

```
V.interior_facet_node_map().top_mask
=> 
{'geometric': array([ 0, -1,  0, -1,  0,  0,  0,  0], dtype=int32),
 'topological': array([ 0, -1,  0, -1,  0,  0,  0,  0], dtype=int32)}
```
Whereas it should be:

```
{'geometric': array([ 0, -1,  0, -1,  0,  -1,  0,  -1], dtype=int32),
 'topological': array([ 0, -1,  0, -1,  0,  -1,  0,  -1], dtype=int32)}
```

wence-:
Fixes #1088 in a hacky way.  A better fix will come in as part of the variable layers work.

### DIFF
--- a/firedrake/functionspacedata.py
+++ b/firedrake/functionspacedata.py
@@ -16,6 +16,7 @@ vs VectorElement) can share the PyOP2 Set and Map data.
 
 from __future__ import absolute_import, print_function, division
 from six.moves import map, range
+from six import iteritems
 
 import numpy
 import finat
@@ -426,7 +427,8 @@ class FunctionSpaceData(object):
         self.map_caches["boundary_node"][method] = val
         return val
 
-    def get_map(self, V, entity_set, map_arity, bcs, name, offset, parent):
+    def get_map(self, V, entity_set, map_arity, bcs, name, offset, parent,
+                kind=None):
         """Return a :class:`pyop2.Map` from some topological entity to
         degrees of freedom.
 
@@ -538,13 +540,28 @@ class FunctionSpaceData(object):
             else:
                 new_entity_node_list = entity_node_list
 
+            if kind == "interior_facet" and self.bt_masks is not None:
+                bt_masks = {}
+                off = map_arity // 2
+                for method, (bottom, top) in iteritems(self.bt_masks):
+                    b = []
+                    t = []
+                    for i in bottom:
+                        b.append(i)
+                        b.append(i+off)
+                    for i in top:
+                        t.append(i)
+                        t.append(i+off)
+                    bt_masks[method] = (b, t)
+            else:
+                bt_masks = self.bt_masks
             val = op2.Map(entity_set, self.node_set,
                           map_arity,
                           new_entity_node_list,
                           ("%s_"+name) % (V.name),
                           offset=offset,
                           parent=parent,
-                          bt_masks=self.bt_masks)
+                          bt_masks=bt_masks)
 
             if decorate:
                 val = op2.DecoratedMap(val, vector_index=True)
--- a/firedrake/functionspaceimpl.py
+++ b/firedrake/functionspaceimpl.py
@@ -455,7 +455,8 @@ class FunctionSpace(object):
                             bcs,
                             "interior_facet_node",
                             offset,
-                            parent)
+                            parent,
+                            kind="interior_facet")
         map.factors = (self.mesh().interior_facets.facet_cell_map,
                        self.cell_node_map())
         return map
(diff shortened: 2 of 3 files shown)


firedrake/functionspaceimpl.py
```diff
@@ -455,7 +455,8 @@ def interior_facet_node_map(self, bcs=None):
                             bcs,
                             "interior_facet_node",
                             offset,
-                            parent)
+                            parent,
+                            kind="interior_facet")
```
wence-: Ugh!

firedrake/functionspacedata.py
```diff
@@ -538,13 +540,28 @@ def get_map(self, V, entity_set, map_arity, bcs, name, offset, parent):
...
+                for method, (bottom, top) in iteritems(self.bt_masks):
+                    b = []
+                    t = []
+                    for i in bottom:
+                        b.append(i)
+                        b.append(i+off)
+                    for i in top:
+                        t.append(i)
+                        t.append(i+off)
+                    bt_masks[method] = (b, t)
```
wence-: There must be a cleaner way to do this.

miklos1: I'm not sure I understand this. Surely, top and bottom BCs have no effect on most interior horizontal facet integrals, except for the lowest interior horizontal facet layer, where bottom masks are to be applied for the bottom cell, and for the highest interior horizontal facet layer, where top masks are to applied for the top cell. But here in the mask you seem to apply both top and bottom masks for both cells. What's going on here?

wence-: the bt_masks argument to the `Map` constructor is a dict of pairs.  The first entry in the pair is the mask for the bottom, the second entry is the mask for the top.  Previously, the code assumed that for interior horizontal facets, the pair is the same as the cell pair.  Of course, this is wrong, because now we only mask out the dofs for the + cell.  This reconstructs the pair to apply the mask to everything on the bottom or top in the full map.

miklos1: Wouldn't it be better to do this duplication in PyOP2?

wence-: after creation, we don't know we need to do the duplication.  Notice the disgusting new `kind` argument to this function.

miklos1: But PyOP2 knows very well whether the integral is `interior_facet_horiz` or not...

wence-: huh, how?

wence-: oh wait, you confused me and I didn't correct.  This covers the case of `interior_facet_vert`, not `interior_facet_horiz`.

miklos1: Ah, I see then...

Issue #884, opened 2016-09-20
FunctionSpace.dim is consistent with dolfin


pefarrell:
In dolfin, `FunctionSpace.dim()` gives the global number of degrees of freedom of the function space.

In firedrake, it is currently a property that does ... something else (number of dofs per mesh node, I think).

pefarrell:
A note: if you want to compute it, it's `V.dof_dset.layout_vec.getSize()`.

PR #1093, opened 2017-06-21
API compatibility

wence-:
Some more API compatibility.  Mostly to support new firedrake-adjoint.

### DIFF
--- a/firedrake/bcs.py
+++ b/firedrake/bcs.py
@@ -8,6 +8,7 @@ from ufl.algorithms.analysis import has_type
 import pyop2 as op2
 from pyop2.profiling import timed_function
 from pyop2 import exceptions
+from pyop2.utils import as_tuple
 
 import firedrake.expression as expression
 import firedrake.function as function
@@ -157,6 +158,13 @@ class DirichletBC(object):
         self.function_arg = val
         self._original_arg = self.function_arg
         self._original_val = val
+
+    @utils.cached_property
+    def domain_args(self):
+        """The sub_domain the BC applies to."""
+        if isinstance(self.sub_domain, str):
+            return (self.sub_domain, )
+        return (as_tuple(self.sub_domain), )
 
     @utils.cached_property
     def nodes(self):
--- a/firedrake/constant.py
+++ b/firedrake/constant.py
@@ -87,6 +87,10 @@ class Constant(ufl.Coefficient):
             return self.dat.data_ro
         return self.dat.data_ro[component]
 
+    def values(self):
+        """Return a (flat) view of the value of the Constant."""
+        return self.dat.data_ro.reshape(-1)
+
     def function_space(self):
         """Return a null function space."""
         return None
--- a/firedrake/extrusion_utils.py
+++ b/firedrake/extrusion_utils.py
@@ -58,7 +58,7 @@ def make_extruded_coords(extruded_topology, base_coords, ext_coords,
                 ext_coords[2*d+1][%(base_coord_dim)d] = *layer_height * (layer[0][0] + 1);
             }
         }""" % {'base_map_arity': base_coords.cell_node_map().arity,
-                'base_coord_dim': base_coords.function_space().dim},
+                'base_coord_dim': base_coords.function_space().value_size},
             "uniform_extrusion_kernel")
     elif extrusion_type == 'radial':
         kernel = op2.Kernel("""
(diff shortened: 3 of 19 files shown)


firedrake/bcs.py
```diff
@@ -159,6 +160,13 @@ def set_value(self, val):
         self._original_val = val
 
     @utils.cached_property
+    def domain_args(self):
+        """The sub_domain the BC applies to."""
+        if isinstance(self.sub_domain, str):
+            return (self.sub_domain, )
+        return (as_tuple(self.sub_domain), )
```
wence-: Arguably this is wrong, but it works for the use-cases we need.

Additionally, we should also provide a copy constructor for DirichletBC.

firedrake/constant.py
```diff
@@ -87,6 +87,10 @@ def evaluate(self, x, mapping, component, index_values):
             return self.dat.data_ro
         return self.dat.data_ro[component]
 
+    def values(self):
+        """Return a (flat) view of the value of the Constant."""
+        return self.dat.data_ro.reshape(-1)
+
     def function_space(self):
         """Return a null function space."""
         return None
```
wence-: Arguably, we should go back and make `Constant(...) => Function(FunctionSpace(mesh, "Real", 0))`.  But the implementation of the latter does not support vector and tensor-valued things, whereas the former does.

firedrake/function.py
```diff
@@ -232,7 +234,8 @@ def __init__(self, function_space, val=None, name=None, dtype=ScalarType):
             raise NotImplementedError("Can't make a Function defined on a "
                                       + str(type(function_space)))
 
-        if isinstance(val, CoordinatelessFunction):
+        if isinstance(val, (Function, CoordinatelessFunction)):
+            val = val.topological
             if val.function_space() != V.topological:
                 raise ValueError("Function values have wrong function space.")
             self._data = val
```
wence-: Allow creating a new function sharing data with an existing one.

firedrake/functionspaceimpl.py
```diff
@@ -391,19 +391,25 @@ def __mul__(self, other):
...
         """The number of degrees of freedom (includes halo dofs) of this
         function space on this process. Cf. :attr:`node_count`."""
-        return self.node_count*self.dim
+        return self.node_count*self.shape_dim
+
+    def dim(self):
+        """The global number of degrees of freedom for this function space.
+
+        See also :attr:`dof_count` and :attr:`node_count`."""
+        return self.dof_dset.layout_vec.getSize()
```
wence-: Fixing #884.

firedrake/ufl_expr.py
```diff
@@ -138,15 +138,24 @@ def derivative(form, u, du=None, coefficient_derivatives=None):
 
     See also :func:`ufl.derivative`.
     """
-    if set(extract_coefficients(form)) & set(u.split()):
+    # TODO: What about Constant?
+    if len(u.split()) > 1 and set(extract_coefficients(form)) & set(u.split()):
```
wence-: This check is ugly.  If `f` is not a mixed function, then `f.split()` is a singleton tuple, which contains a new function that shares data with `f`.  Whereas if `f` is a `Constant`, then `f.split()` is a singleton tuple which is `f`.

firedrake/ufl_expr.py
```diff
@@ -138,15 +138,24 @@ def derivative(form, u, du=None, coefficient_derivatives=None):
...
             number = max(a.number() for a in args) if args else -1
             du = Argument(V, number + 1)
+        elif isinstance(u, firedrake.Constant):
+            if u.ufl_shape != ():
+                raise ValueError("Real function space of vector elements not supported")
+            mesh = form.ufl_domain()
+            V = firedrake.FunctionSpace(mesh, "Real", 0)
+            args = form.arguments()
+            number = max(a.number() for a in args) if args else -1
+            du = Argument(V, number + 1)
```
wence-: Should be able to differentiate wrt `Constant`.  Again, suggests `Constant` should just be `Function(FunctionSpace(mesh, "Real", 0))`.

firedrake/extrusion_utils.py
```diff
@@ -58,7 +58,7 @@ def make_extruded_coords(extruded_topology, base_coords, ext_coords,
                 ext_coords[2*d+1][%(base_coord_dim)d] = *layer_height * (layer[0][0] + 1);
             }
         }""" % {'base_map_arity': base_coords.cell_node_map().arity,
-                'base_coord_dim': base_coords.function_space().dim},
+                'base_coord_dim': base_coords.function_space().shape_dim},
```
miklos1: `value_size`?

firedrake/functionspaceimpl.py
miklos1: :+1:

firedrake/extrusion_utils.py
wence-: `.ufl_element().value_size()`.  Yes, that would work, but breaks if coordinates become intrinsically vector-valued (unlikely).

miklos1: `.ufl_element().value_size()` is not quite the same as you mentioned. What I wanted to suggest is that `value_size` might be a more consistent/familiar name than `shape_dim`.

wence-: Oh, I see, yes.

PR #1097, opened 2017-07-03
Allow LVS jacobian to be invalidated by user

dorugeber:
Sometimes, setting `constant_jacobian=False` in the LVP is overkill, e.g. if the underlying mesh remains constant for several timesteps before being adapted, or similar.  What is presented here is a bit nicer than periodically remaking the entire solver object.

Presumably not necessary to implement this for NVP/NLS (though would be simple to implement), since `constant_jacobian` is automatically `False`.


dorugeber:
-- minor, but feel free to suggest better names, e.g. `force_matrix_reassemble()`, ...

PR #1102, opened 2017-07-17
Make use of the labels for interior facets defined in gmsh

florianwechsung:
I'm happy to add a test as well, but that would require adding a `.msh` file to the repo.


dham:
Hi Florian,

Please do add a test. This sort of change is quite susceptible to being broken by future developments. It's fine to put a (smallish) .msh file in the repo.

wence-:
For tests we put them in `tests/meshes/...` (please include the geo), and you can then find the mesh file with something like:

```python
cwd = abspath(dirname(__file__))


@pytest.fixture
def mesh():
    return Mesh(join(cwd, "..", "meshes", "cell-sets.msh"))
```

wence-:
I'll close this, because we're doing it in #1110.  Thanks!

PR #1103, opened 2017-07-18
An indentation fix

miklos1:


firedrake/pointquery_utils.py
```diff
@@ -127,7 +127,7 @@ def to_reference_coordinates(ufl_coordinate_element):
         if max_extent:
             def predicate(index):
                 return index.extent <= max_extent
-        ir = gem.optimise.unroll_indexsum(ir, predicate=predicate)
```
miklos1: If `max_extent` counts as false, then predicate is not defined, which cases an error.

PR #1109, opened 2017-07-21
Fix periodic quad mesh warning

### LINKED ISSUE #1090: Periodic square mesh produces a warning about DQ element
I am setting up a case on square quadrilateral mesh. It's a simple Poisson equation but my target test case is meant be periodic in the x direction so I went with `PeriodicSquareMesh` object from `utility_meshes`.
```python
mesh = PeriodicSquareMesh(N, N, 1.0, direction='x', quadrilateral=True)
```

The strange thing is that no matter which element I am trying in `FunctionSpace`:
```python
V = FunctionSpace(mesh, 'Q', 2)
# V = FunctionSpace(mesh, 'DQ', 2)
# V = FunctionSpace(mesh, 'CG', 2)
```
Firedrake produces the following warning
```
UFL:WARNING Discontinuous Lagrange element requested on quadrilateral, creating DQ element
```
It produced two additional warnings if I decide to save my solution.

The problem disappears if I use quadrilateral `UnitSquareMesh`.

dham:
Actually use the correct element type when constructing periodic meshes. Fixes #1090.

### DIFF
--- a/firedrake/utility_meshes.py
+++ b/firedrake/utility_meshes.py
@@ -487,7 +487,8 @@ cells in each direction are not currently supported")
 
     m = TorusMesh(nx, ny, 1.0, 0.5, quadrilateral=quadrilateral, reorder=reorder,
                   comm=comm)
-    coord_fs = VectorFunctionSpace(m, 'DG', 1, dim=2)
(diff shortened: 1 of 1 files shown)


robertsawko:
Thanks for sorting it out.

PR #1110, opened 2017-07-21
Cleanup mesh markers

wence-:
Cleans up some of the mesh marker code, subsumes #1102.  @florianwechsung can you confirm this branch DTRT for you.  Also, if you can provide a (small) test that would be great.

Also brings in PDE test of octahedral sphere mesh (which exposes possible issues with the exact placement of the nodes)


firedrake/dmplex.pyx
```diff
@@ -913,6 +917,35 @@ def get_facet_ordering(PETSc.DM plex, PETSc.Section facet_numbering):
...
+    CHKERR(DMGetLabel(dm.dm, FACE_SETS_LABEL, &label))
+    for f in range(nfacet):
+        CHKERR(DMLabelGetValue(label, facets[f], &val))
+        if val != -1:
+            all_default = PETSC_FALSE
+        ids[f] = val
+    if all_default:
+        return None
+    else:
+        return ids
```
wence-: This was previously done in python, but only for exterior facets.  Now that we mark interior facets too, it's best to do it in C.

firedrake/dmplex.pyx
```diff
@@ -927,16 +960,14 @@ def get_facets_by_class(PETSc.DM plex, label,
...
         PetscBool has_point, is_class
         DMLabel lbl_facets, lbl_class
         np.ndarray[PetscInt, ndim=1, mode="c"] facets
 
-    label_chr = <char*>label
     dim = plex.getDimension()
     fStart, fEnd = plex.getHeightStratum(1)
     pStart, pEnd = plex.getChart()
-    CHKERR(DMGetLabel(plex.dm, label, &lbl_facets))
+    CHKERR(DMGetLabel(plex.dm, <const char*>label, &lbl_facets))
```
wence-: Cleanup.

firedrake/mesh.py
```diff
@@ -65,9 +65,8 @@ def __init__(self, mesh, classes, kind, facet_cell, local_facet_number, markers=
 
         # assert that markers is a proper subset of unique_markers
         if markers is not None:
-            for marker in markers:
-                assert (marker in unique_markers), \
-                    "Every marker has to be contained in unique_markers"
+            assert set(markers) <= set(unique_markers), \
+                "Every marker has to be contained in unique_markers"
```
wence-: Make it clearer what we are doing.

firedrake/bcs.py
```diff
@@ -58,11 +58,7 @@ def __init__(self, V, g, sub_domain, method="topological"):
         self.function_arg = g
         self.comm = V.comm
         self._original_arg = self.function_arg
-        if sub_domain == "on_boundary":
-            self.sub_domain = \
-                tuple(map(int, V.mesh().topology.exterior_facets.unique_markers))
-        else:
-            self.sub_domain = sub_domain
+        self.sub_domain = sub_domain
```
wence-: The markers may not mark the entire boundary, so we deal with it when constructing the nodes.

firedrake/bcs.py
```diff
@@ -170,25 +166,24 @@ def domain_args(self):
...
-            return np.unique(
-                fs.exterior_facet_boundary_node_map(
-                    self.method).values_with_halo.take(
-                    fs._mesh.exterior_facets.subset(self.sub_domain).indices,
-                    axis=0))
+            values = V.exterior_facet_boundary_node_map(
+                self.method).values_with_halo
+            if self.sub_domain != "on_boundary":
+                values = values.take(mesh.exterior_facets.subset(self.sub_domain).indices,
+                                     axis=0)
```
wence-: Filter indices if the subdomain is not "on_boundary".

firedrake/dmplex.pyx
```diff
@@ -23,6 +23,10 @@ cdef extern from "mpi-compat.h":
 include "dmplexinc.pxi"
 
 
+FACE_SETS_LABEL = "Face Sets"
+CELL_SETS_LABEL = "Cell Sets"
```
wence-: OAOO

florianwechsung:
I added a test in 0233327b1a76923f4c9ab12a2fd3a32e65dfe642, however it fails on this branch in `mesh.py:67`

```
        if markers is not None:
            assert set(markers) <= set(unique_markers), \
                "Every marker has to be contained in unique_markers"
```

because `markers` additionally contains lots of `-1` entries (I assume that those interior facets that are not explicitly labeled in gmsh are labeled with `-1` by default).

wence-:
>I added a test in 0233327, however it fails on this branch in mesh.py:67

>        if markers is not None:
>            assert set(markers) <= set(unique_markers), \
>                "Every marker has to be contained in unique_markers"
> because markers misses the 1label and additionally contains lots of -1 entries (I assume that those interior facets that are not explicitly labeled in gmsh are labeled with -1 by default).

Yes.  These labels are sparse in the mesh representation, but we make them dense when converting into markers for facets.  The idea is that you have to mark everything because there isn't an obvious way of expressing an integral over the complement of some set of markers.

With this defaulting in place, I can make `-1` be "the complement to all explicitly labelled things", but that's a bit non-intuitive.  WDYT?

wence-:
OTOH, I see how it is not easy to explicitly mark all the interior facets. So if you're only interested in integrating over this marked facet for diagnostic purposes, (and don't need to iterate over the complement) then that's easier to handle.

florianwechsung:
I need the marked surface on two occasions: to calculate a scalar integral over the surface and to define a right-handside for an adjoint equation. So I don't need to integrate over the complement at any time. The `-1=unmarked` default seems sensible to me!

wence-:
Thanks, I copied your test over which seems to work.  I made a slightly smaller mesh.

I wonder what the best name for the "complement" marker is.  Right now you write:

```python
f = ...*dS(defaultmarker)
```

But perhaps

```python
f = ...*dS(unmarked)
```

would be better?

florianwechsung:
I agree, that's more intuitive.

firedrake/mesh.py
```diff
@@ -508,70 +494,47 @@ def cell_closure(self):
...
-
-            # Determine union of boundary IDs.  All processes must
-            # know the values of all ids, for collective reasons.
-            comm = self.comm
+    def _facets(self, kind):
+        if kind not in ["interior", "exterior"]:
+            raise ValueError("Unknown facet type '%s'" % kind)
+
+        dm = self._plex
+        facets, classes = dmplex.get_facets_by_class(dm, (kind + "_facets").encode(),
```
wence-: str-to-bytes to go to C.

firedrake/mg/impl.pyx
```diff
@@ -526,12 +527,12 @@ def filter_exterior_facet_labels(PETSc.DM plex):
...
     # zero-sized), but may not always have boundary_ids or
     # boundary_faces.
-    has_bdy_ids = plex.hasLabel("boundary_ids")
+    has_bdy_ids = plex.hasLabel(dmplex.FACE_SETS_LABEL)
     has_bdy_faces = plex.hasLabel("boundary_faces")
 
     CHKERR(DMGetLabel(plex.dm, <char*>"exterior_facets", &exterior_facets))
     if has_bdy_ids:
-        CHKERR(DMGetLabel(plex.dm, <char*>"boundary_ids", &boundary_ids))
+        CHKERR(DMGetLabel(plex.dm, <char*>dmplex.FACE_SETS_LABEL, &boundary_ids))
```
wence-: This will need `.encode()` I think.

firedrake/mesh.py
```diff
@@ -34,6 +34,10 @@
 }
 
 
+unmarked = -1
+"""A mesh marker that selects all entities that are not explicitly marked."""
```
wence-: When I write `dS(foo)` to select a subdomain, `foo` must be an integer.  So I use `-1` (the default for unmarked entities in DMLabels) to indicate the complement of all marked entities here.

PR #1128, opened 2017-08-04
Owner computes

wence-:
Switch to an owner computes model.  This shrinks the halos.  We now only exchange data that is in the closure of the support of boundary facets.

Not ready for merge yet, because the PETSc change necessary is not merged right now, but opening for discussion.

Fixes subtle issues to do with assembling over interior facets in parallel when the field in question is not discontinuous.

### DIFF
--- a/firedrake/dmplex.pyx
+++ b/firedrake/dmplex.pyx
@@ -879,8 +879,21 @@ def get_cell_markers(PETSc.DM plex, PETSc.Section cell_numbering,
     if not plex.hasLabel(CELL_SETS_LABEL):
         return np.empty(0, dtype=IntType)
     vals = plex.getLabelIdIS(CELL_SETS_LABEL).indices
-    if subdomain_id not in vals:
+    comm = plex.comm.tompi4py()
+
+    def merge_ids(x, y, datatype):
+        return x.union(y)
+
+    op = MPI.Op.Create(merge_ids, commute=True)
+
+    all_ids = np.asarray(sorted(comm.allreduce(set(vals), op=op)),
+                         dtype=IntType)
+    op.Free()
+    if subdomain_id not in all_ids:
         raise ValueError("Invalid subdomain_id %d not in %s" % (subdomain_id, vals))
+
+    if subdomain_id not in vals:
+        return np.empty(0, dtype=IntType)
 
     indices = plex.getStratumIS(CELL_SETS_LABEL, subdomain_id).indices
     cells = np.empty(indices.shape[0], dtype=IntType)
--- a/firedrake/dmplex.pyx
+++ b/firedrake/dmplex.pyx
@@ -636,182 +636,83 @@ def mark_entity_classes(PETSc.DM plex):
     """Mark all points in a given Plex according to the PyOP2 entity
     classes:
 
-    core      : owned and not in send halo
-    non_core  : owned and in send halo
-    exec_halo : in halo, but touch owned entity
-    non_exec_halo : in halo and only touch halo entities
+    core   : owned and not in send halo
+    owned  : owned and in send halo
+    ghost  : in halo
 
     :arg plex: The DMPlex object encapsulating the mesh topology
     """
     cdef:
-        PetscInt p, pStart, pEnd, cStart, cEnd, vStart, vEnd, fStart, fEnd
-        PetscInt c, ncells, f, nfacets, ci, nclosure, vi, dim
-        PetscInt depth, non_core, exec_halo, nroots, nleaves
-        PetscInt v_per_cell
-        PetscInt *cells = NULL
-        PetscInt *facets = NULL
-        PetscInt *vertices = NULL
+        PetscInt pStart, pEnd, cStart, cEnd
+        PetscInt c, ci, p
+        PetscInt nleaves
         PetscInt *closure = NULL
+        PetscInt nclosure
         PetscInt *ilocal = NULL
         PetscBool non_exec
         PetscSFNode *iremote = NULL
         PETSc.SF point_sf = None
-        PETSc.IS cell_is = None
-        PETSc.IS facet_is = None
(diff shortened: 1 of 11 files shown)


firedrake/dmplex.pyx
```diff
@@ -636,182 +636,83 @@ def mark_entity_classes(PETSc.DM plex):
...
-                    non_exec = PETSC_FALSE
-        if non_exec:
-            # If we still think we're non-exec, check for halo cells
-            CHKERR(DMPlexGetTransitiveClosure(plex.dm, facets[f],
-                                              PETSC_FALSE,
-                                              &nclosure,
-                                              &closure))
+            CHKERR(DMLabelHasPoint(lbl_ghost, p, &is_ghost))
+            if is_ghost:
+                is_owned = PETSC_TRUE
```
wence-: So a point in the closure is a ghost, so therefore all the points in the closure must be either owned or ghost.

firedrake/dmplex.pyx
```diff
@@ -636,182 +636,83 @@ def mark_entity_classes(PETSc.DM plex):
...
-            continue
-        CHKERR(DMLabelHasPoint(lbl_non_exec, facets[f], &has_point))
-        if has_point:
-            # Remove facet from exec-halo label
-            CHKERR(DMLabelClearValue(lbl_exec, facets[f], 1))
-
+                p = closure[2*ci]
+                CHKERR(DMLabelHasPoint(lbl_ghost, p, &is_ghost))
+                if not is_ghost:
+                    CHKERR(DMLabelSetValue(lbl_owned, p, 1))
```
wence-: All those points in the closure that are not ghosts must be owned.

firedrake/halo.py
```diff
@@ -33,71 +35,75 @@ def _get_mtype(dat):
...
+    def comm(self):
+        return self.dm.comm.tompi4py()
 
-        .. note::
+    @utils.cached_property
+    def local_to_global_numbering(self):
+        lsec = self.dm.getDefaultSection()
+        gsec = self.dm.getDefaultGlobalSection()
+        return dmplex.make_global_numbering(lsec, gsec)
+        return self._gnn2unn
```
wence-: Remove.

firedrake/halo.py
```diff
@@ -33,71 +35,75 @@ def _get_mtype(dat):
...
-    def verify(self, *args):
-        """No-op"""
-        pass
-
-    @property
-    def global_to_petsc_numbering(self):
-        """Return a mapping from global (process-local) to universal
-    (process-global) numbers"""
-        return self._gnn2unn
+        dmplex.halo_end(self.sf, dat, mtype, True)
```
wence-: Basically just adapting to the new abstract base class in PyOP2.

firedrake/mesh.py
```diff
@@ -372,8 +372,10 @@ def __init__(self, plex, name, reorder, distribute):
         def callback(self):
             """Finish initialisation."""
             del self._callback
-            if self.comm.size > 1:
+            if self.comm.size > 1 and distribute:
+                dmplex.set_adjacency_callback(self._plex)
                 self._plex.distributeOverlap(1)
+                dmplex.clear_adjacency_callback(self._plex)
```
wence-: Grow the overlap with our custom callback.

firedrake/bcs.py
```diff
@@ -167,21 +168,31 @@ def nodes(self):
...
+        # Now determine which full set of indices (including those in
+        # ghost region).  Simple-minded GtoL right now, could be
+        # improved probably
+        # FIXME: need to figure out if all of this is necessary.
+        d = op2.Dat(V.dof_dset.set, dtype=np.int32)
+        d.data_with_halos[local_indices] = 1
+        d.global_to_local_begin(op2.READ)
+        d.global_to_local_end(op2.READ)
+        indices, = np.where(d.data_ro_with_halos == 1)
+        return indices.astype(IntType)
```
miklos1: It is not obvious to me why this is necessary...

firedrake/dmplexinc.pxi
```diff
@@ -63,7 +65,7 @@ cdef extern from "petscsf.h":
     int PetscSFReduceBegin(PETSc.PetscSF,MPI.MPI_Datatype,const void*, void*,MPI.MPI_Op)
     int PetscSFReduceEnd(PETSc.PetscSF,MPI.MPI_Datatype,const void*, void*,MPI.MPI_Op)
 
-cdef extern from "petscbt.h":
+cdef extern from "petscbt.h" nogil:
```
miklos1: Does `nogil` makes things faster, or it helps avoiding deadlocks? What is the effect?

firedrake/function.py
```diff
@@ -489,6 +489,8 @@ def at(self, arg, *args, **kwargs):
         """
         # Need to ensure data is up-to-date for reading
         self.dat._force_evaluation(read=True, write=False)
+        self.dat.global_to_local_begin(op2.READ)
+        self.dat.global_to_local_end(op2.READ)
```
miklos1: So `_force_evaluation` no longer forces halos to be up-to-date?

firedrake/halo.py
miklos1: Yes, unreachable code...

firedrake/function.py
wence-: Yes, because the computation phase guarantees only that owned points are correct once the computation has finished.  We need this global to local phase to get up to date ghosts.

firedrake/dmplexinc.pxi
wence-: It means that this function guarantees it doesn't invoke the python interpreter.  Cython emits some "refnanny" pre-/post-ambles to functions that might grab the gil.  I would like the new `use adjacency` function not to do this, which requires annotating that function, and its callees as `nogil`.

PR #1131, opened 2017-08-14
Fix coefficient mapping for Slate subkernels

thomasgibson:
A small, but important change. Basically forms with several coefficients weren't properly being organized and led to nasty errors in the generated code (discovered by Colin). The cause: coefficients weren't being mapped properly to their corresponding assembly kernels. This appears to fix it up.

Some additional minor changes were also made (rearranging code blocks, deleting extra whitespace, etc.)

### DIFF
--- a/firedrake/slate/slac/compiler.py
+++ b/firedrake/slate/slac/compiler.py
@@ -136,14 +136,16 @@ def compile_expression(slate_expr, tsfc_parameters=None):
 
             # If tensor is mixed, there will be more than one SplitKernel
             incl = []
+            split_forms = cxt_kernel.form_indices_map
             for splitkernel in cxt_kernel.tsfc_kernels:
                 index = splitkernel.indices
                 kinfo = splitkernel.kinfo
 
                 # Generate an iterable of coefficients to pass to the subkernel
-                # if any are required
-                clist = [c for ci in kinfo.coefficient_map
-                         for c in builder.coefficient(exp.coefficients()[ci])]
+                # if any are required.
+                sform = split_forms[index]
+                clist = [c for i in kinfo.coefficient_map
+                         for c in builder.coefficient(sform.coefficients()[i])]
 
                 if kinfo.oriented:
                     clist.insert(0, cell_orientations)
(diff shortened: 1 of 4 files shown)


firedrake/slate/slac/compiler.py
```diff
@@ -136,14 +136,16 @@ def compile_expression(slate_expr, tsfc_parameters=None):
...
                 kinfo = splitkernel.kinfo
 
                 # Generate an iterable of coefficients to pass to the subkernel
-                # if any are required
-                clist = [c for ci in kinfo.coefficient_map
-                         for c in builder.coefficient(exp.coefficients()[ci])]
+                # if any are required.
+                sform = split_forms[index]
+                clist = [c for i in kinfo.coefficient_map
+                         for c in builder.coefficient(sform.coefficients()[i])]
```
thomasgibson: The `ContextKernel` contains the relevant mapping between the `kinfo.indices` and the associated split form. Then the `kinfo.coefficient_map` provides the map to the relevant coefficients of the split form. This way, the local assembly kernel should have the right coefficient arguments.

firedrake/slate/slac/tsfc_driver.py
```diff
@@ -47,10 +60,12 @@ def compile_terminal_form(tensor, prefix=None, tsfc_parameters=None):
...
     for orig_it_type, integrals in transformed_integrals.items():
         subkernel_prefix = prefix + "%s_to_" % orig_it_type
-        kernels = tsfc_compile(Form(integrals),
+        form = Form(integrals)
+        kernels = tsfc_compile(form,
                                subkernel_prefix,
                                parameters=tsfc_parameters)
         cxt_k = ContextKernel(tensor=tensor,
+                              form_indices_map=dict(split_form(form)),
                               original_integral_type=orig_it_type,
```
wence-: This stuff is some serious action at a distance.  Inside `tsfc_compile`, we call `split_form` and unwind some coefficient numbering from the split forms into the original expression.  That information is then immediately thrown away, because we don't need it any more.

Here you're sort of "recreating" it, by remembering the split forms in the `ContextKernel`.

firedrake/slate/slac/tsfc_driver.py
```diff
@@ -47,10 +60,12 @@ def compile_terminal_form(tensor, prefix=None, tsfc_parameters=None):
     cxt_kernels = []
     for orig_it_type, integrals in transformed_integrals.items():
         subkernel_prefix = prefix + "%s_to_" % orig_it_type
-        kernels = tsfc_compile(Form(integrals),
+        form = Form(integrals)
+        kernels = tsfc_compile(form,
                                subkernel_prefix,
                                parameters=tsfc_parameters)
         cxt_k = ContextKernel(tensor=tensor,
```
wence-: I note in passing that this is weird.  I would have expected that for each tensor there is one and only one `ContextKernel`, but it appears not.  There are `len(transformed_integrals)` `ContextKernel`s, all sharing the same `tensor`

firedrake/slate/slac/compiler.py
```diff
@@ -462,14 +466,17 @@ def facet_integral_loop(cxt_kernel, builder, coordsym, cellfacetsym,
...
 
         # Generate an iterable of coefficients to pass to the subkernel
-        # if any are required
-        clist = [c for ci in kinfo.coefficient_map
-                 for c in builder.coefficient(exp.coefficients()[ci])]
+        # if any are required.
+        # Split form for this particular splitkernel
+        sform = split_forms[index]
+        clist = [c for i in kinfo.coefficient_map
+                 for c in builder.coefficient(sform.coefficients()[i])]
```
wence-: OK, so rather than indexing into the original tensor expression's coefficients, you instead magically remember the split form, and ask it for its coefficients.  I think a much better thing to do is for this mapping to live in the `ContextKernel`, then you can map from the coefficients of the `tensor` directly to the relevant sub-bit.

firedrake/slate/slac/tsfc_driver.py
thomasgibson: Yeah. When I originally introduced the `ContextKernel`, it was mainly meant to help organize the incoming TSFC kernels corresponding to a particular form. And it happens to be the case that a `Tensor` can have many integrals contained in `Tensor.form` (true also if the form is mixed), hence why multiple context kernels share the same Slate tensor. I'm debating whether or not to change it in this PR. I think it could work just fine with one `ContextKernel` per `Tensor`, but I need to think about how to organize the various integral types. Thoughts?

wence-: So I think (discussed) that what you want is a bucket which contains tsfc kernels that correspond to the different integral types.  So why not make that?

Then the coefficient_mapping you need can be created by composing the coefficient mapping from the tensor's coefficients to the form that provides each of these integral type's coefficients with the tsfc kernel's coefficient map.  No need to split the form (which is good, because you're no longer repeating work).

thomasgibson: Updated. The `ContextKernel` keeps the local form's coefficients and the `kinfo.coefficient_map`s take care of the rest.

thomasgibson:
I think this is safe to close since #1134 is being considered over this one.

PR #1132, opened 2017-08-16
Make FunctionSpace hashable

miklos1:


miklos1:
@thomasgibson: Thanks, I made this patch more complete now.

PR #1134, opened 2017-08-29
Slate: compiler updates and better code generation

thomasgibson:
This PR supersedes #1131 as the fix is included.

## Summary:
1. Removes a significant amount of code-duplication, specifically when generating function calls to TSFC subkernels. Some code was re-factored as well, particularly in `kernel_builder.py` and `compiler.py`.

2. `interior_facet_horiz` is split into `interior_facet_horiz_top` and `interior_facet_horiz_bottom` for reasons which are clear from a Slate point of view and motivated by item 3.

3. The `KernelBuilder` is split into a base class and the primary builder `LocalKernelBuilder` is added to `compiler.py`. The `LocalKernelBuilder` manages the function calls to TSFC kernels via `builder.assembly_calls`. The TSFC kernels are organized by integral type.

      - `compile_expression` extracts all the function calls from `builder.assembly_calls` and performs some basic reasoning. This includes loop-fusion for interior/exterior facet calls, and grouping mesh-level-dependent calls (exterior top/bottom and interior horizontal facets)

4. Loop-fusion for coefficient temporaries in `Action` nodes. The coefficients are grouped by node and dof extent.

5. More aggressive temporary creation. All expressions have a reference count (obtained from the builder) which is used to determine whether an additional temporary should be introduced. Currently, I have set it such that nodes (excluding `Negative` and `Transpose`) with a reference count greater than 1 are given a temporary. Whether or not this is the best approach is up for discussion.


firedrake/slate/slac/compiler.py
```diff
@@ -41,18 +45,167 @@
...
+    @property
+    def integral_type(self):
+        """Returns the integral type associated with a Slate kernel.
+
+        Note that Slate kernels (for now) are always of type 'cell' since
+        these are localized kernels for element-wise linear algebra. This
+        may change in the future if we want Slate to be used for LDG/CDG
+        finite element discretizations.
+        """
+        return "cell"
```
wence-: Best not to add "we might do this in the future" stuff into the docstring, you could rather write:

```
integral_type = "cell"
"Integral type associated with a Slate kernel."

File: firedrake/slate/slac/compiler.py [hunk 2]
```diff
@@ -41,18 +45,167 @@
...
+                            "interior_facet_horiz_bottom",
+                            "exterior_facet_bottom",
+                            "exterior_facet_top"]
+        return any(cxt_k.original_integral_type in mesh_layer_types
+                   for cxt_k in self.context_kernels)
+
+    def construct_ast(self, *args):
+        """Constructs the final kernel AST.
+
+        :arg args: An iterable of macro kernel functions, which call subkernels
```
wence-: You don't provide an iterable (a single thing), but rather just the macro kernels.

firedrake/slate/slac/compiler.py
```diff
@@ -41,18 +45,167 @@
...
+                   and perform elemental linear algebra.
+
+        Returns: The complete kernel AST as a COFFEE `ast.Node`
+        """
+        macro_kernels = list(args)
+        kernel_ast = self.templated_subkernels
+        kernel_ast.extend(macro_kernels)
```
wence-: This modifies `self.templated_subkernels` in place, is that intentional?

Also confusing is that there's somehow no guarantee that the provided macro kernels only call subkernels that this kernel builder object already knows about.

firedrake/slate/slac/compiler.py
```diff
@@ -70,154 +223,255 @@ def compile_expression(slate_expr, tsfc_parameters=None):
...
+        chker = {"interior_facet": 1,
+                 "exterior_facet": 0,
+                 "interior_facet_vert": 1,
+                 "exterior_facet_vert": 0}
+        int_calls = list(chain(*[builder.assembly_calls[it_type]
+                                 for it_type in chker
+                                 if chker[it_type] == 1]))
+        ext_calls = list(chain(*[builder.assembly_calls[it_type]
+                                 for it_type in chker
+                                 if chker[it_type] == 0]))
```
wence-: Ugh, prefer `if it_type in set(...)`.

firedrake/slate/slac/compiler.py
```diff
@@ -263,26 +516,26 @@ def compile_expression(slate_expr, tsfc_parameters=None):
...
-                           include_dirs=inc,
+                           include_dirs=include_dirs,
                            headers=['#include <Eigen/Dense>',
                                     '#define restrict __restrict'])
 
-    assert len(slate_expr.ufl_domains()) == 1, (
-        "No support for multiple domains yet!"
-    )
+    if len(slate_expr.ufl_domains()) > 1:
+        raise NotImplementedError("Multiple domains not implemented.")
```
wence-: I guess you could move this check earlier.

firedrake/slate/slac/compiler.py
```diff
@@ -295,294 +548,33 @@ def compile_expression(slate_expr, tsfc_parameters=None):
...
-                # Update declared temporaries with the coefficient
-                declared_temps[actee] = ctemp
-        else:
-            raise NotImplementedError(
-                "Auxiliary expr type %s not currently implemented." % type(exp)
-            )
+    # all kernel body statements must be wrapped up as a coffee.base.Block
+    assert isinstance(statements, ast.Block), (
+        "Body statements must be wrapped in an ast.Block"
+    )
```
wence-: I still don't like all this whitespace, why not just `assert XXX, "body statements ..."`?

firedrake/slate/slac/compiler.py
```diff
@@ -610,7 +602,7 @@ def metaphrase_slate_to_cpp(expr, temps, prec=None):
     """
     # If the tensor is terminal, it has already been declared.
     # Coefficients in action expressions will have been declared by now,
-    # as well as any inverses with high reference count.
+    # as well as any other nodes with high reference count.
```
wence-: What is "high"?

firedrake/slate/slac/kernel_builder.py
```diff
@@ -1,76 +1,70 @@
...
 
         :arg expression: a :class:`TensorBase` object.
         :arg tsfc_parameters: an optional `dict` of parameters to provide to
                               TSFC when constructing subkernels associated
                               with the expression.
         """
         assert isinstance(expression, TensorBase)
+
+        # Collect reference counts
+        ref_counts = collect_reference_count([expression])
```
wence-: No need for a comment like this, if the code on the following line doesn't collect reference counts then it is highly misnamed.

firedrake/slate/slac/kernel_builder.py
```diff
@@ -1,76 +1,70 @@
...
+            elif isinstance(tensor, TensorOp):
+                # Actions will always require a coefficient temporary.
+                if isinstance(tensor, Action):
+                    actee, = tensor.actee
+                    action_coefficients.append(actee)
+
+                # Operations which have "high" reference count will have
+                # auxiliary temporaries created. Negative and Transpose
+                # operations will not have extra temporaries.
+                if ref_counts[tensor] > 1 and not isinstance(tensor, (Negative, Transpose)):
```
wence-: Is it obvious why you don't make a temporary for a transposed matrix?

firedrake/slate/slac/utils.py
```diff
@@ -137,6 +136,19 @@ def visit_Symbol(self, o, *args, **kwargs):
...
+    """Counts the total number of nodes in a given Slate expression.
+
+    :arg expr: A Slate expression.
+    """
+    count = 1
+
+    for op in expr.operands:
+        count += count_operands(op)
+
+    return count
```
wence-: This doesn't dagify, is that intentional?

firedrake/slate/slate.py
```diff
@@ -118,6 +118,15 @@ def subdomain_data(self):
...
         """
 
+    @cached_property
+    def is_mixed(self):
+        """Returns `True` if the tensor has mixed arguments and `False` otherwise.
+        """
+        if any(len(arg.function_space()) > 1 for arg in self.arguments()):
+            return True
+
+        return False
```
wence-: AKA `return any...`

```python
if condition:
    return True
return False
```

should be shortened to `return condition`.  If you really care about booleanness, then say `return bool(condition)`, but no need to here.

firedrake/slate/slac/tsfc_driver.py
```diff
@@ -47,10 +60,12 @@ def compile_terminal_form(tensor, prefix=None, tsfc_parameters=None):
     cxt_kernels = []
     for orig_it_type, integrals in transformed_integrals.items():
         subkernel_prefix = prefix + "%s_to_" % orig_it_type
-        kernels = tsfc_compile(Form(integrals),
+        form = Form(integrals)
+        kernels = tsfc_compile(form,
                                subkernel_prefix,
                                parameters=tsfc_parameters)
         cxt_k = ContextKernel(tensor=tensor,
+                              coefficients=form.coefficients(),
```
wence-: Is there a test of this fix included?

firedrake/slate/slac/kernel_builder.py
thomasgibson: Well, in Eigen `A.transpose()` returns a proxy object without actually computing the transpose. Assignments will trigger an evaluation: `T = A.transpose()`. However, now that I think about it, if a transpose appears in n places, (a transpose proxy object appears n times in the expression), then surely the transpose would need to physically be evaluated at some point? So this means n transpose evaluations?

OK. I think I'll go ahead and amend this and include the transpose.

firedrake/slate/slac/compiler.py
thomasgibson: Well it's a value currently set to 1. Meaning nodes with a reference count greater than 1 will have a temporary created for it. Whether this is the best way to do this sort of auxiliary temp creation is still open for discussion I think. This seems to work just fine based on my experiments.

firedrake/slate/slac/compiler.py
thomasgibson: Old habits... I'll go through the code and tidy it up.

firedrake/slate/slac/compiler.py
thomasgibson: You mean something like:
```python
int_calls = list(chain(*[builder.assembly_calls[it_type]
                         for it_type in ("interior_facet", "interior_facet_vert")]))
```
?

wence-: Oh, I guess yes, I hadn't realised you were iterating over the integral types in chker and filtering based on the value.

firedrake/slate/slac/tsfc_driver.py
thomasgibson: Yes, one has recently been added in `tests/slate/test_assemble_tensors.py`

firedrake/slate/slac/utils.py
thomasgibson: I didn't have DAGs in mind when I wrote this. This function is used to sort the expressions requiring auxiliary temporaries. I want to order them in such a way that I don't run into an instance where I'm declaring a temporary for an expression which *may* depend on another temporary that hasn't been declared yet.

I just learned today that what I actually want is to *topologically sort* the expressions. So I will replace this function is a top-sort algorithm.

firedrake/slate/slac/compiler.py
```diff
@@ -41,18 +45,147 @@
...
+            else:
+                coords = coordinates
+
+            for split_kernel in cxt_kernel.tsfc_kernels:
+                indices = split_kernel.indices
+                kinfo = split_kernel.kinfo
+
+                # TSFC No-op kernels do not contain a kernel body. If this is
+                # the case, we don't create a function call.
+                if kinfo.kernel:
```
wence-: When is it the case that `compile_form` returns an iterable of kernels from tsfc that contains an empty kernel?

firedrake/slate/slac/utils.py
```diff
@@ -137,6 +136,37 @@ def visit_Symbol(self, o, *args, **kwargs):
...
+    The list is generated by examining the dependencies of each
+    expression and moving nodes that have all its dependencies
+    resolved. A node is resolved once all the nodes in its dependencies
+    have been moved into the sorted list.
+
+    :arg exprs: An iterable of Slate nodes.
+    """
+    # Don't include the expr itself in its dependencies
+    dependency_graph = {expr: list(set(traverse_dags([expr])) - {expr})
+                        for expr in exprs}
```
wence-: I think this wants to be an `OrderedDict`.

firedrake/slate/slac/utils.py
```diff
@@ -137,6 +136,37 @@ def visit_Symbol(self, o, *args, **kwargs):
         return SymbolWithFuncallIndexing(o.symbol, o.rank, o.offset)
 
 
+def topological_sort(exprs):
+    """Returns a topologically sorted list of Slate expressions.
+    The list is generated by examining the dependencies of each
+    expression and moving nodes that have all its dependencies
+    resolved. A node is resolved once all the nodes in its dependencies
+    have been moved into the sorted list.
```
wence-: This just described you implementation of the topo sort.  Better to just say what graph it is you are sorting.

firedrake/slate/slac/compiler.py
thomasgibson: This only really occurs when using the `HybridizationPC`. Suppose `a` is the mixed bilinear form describing the mixed poisson operator. When generating the reconstruction calls for the flux and scalar unknowns, the form is split to generate the relevant local `Tensor` objects. Calling `split_form(a)` will zero-pad the scalar-mass bit (since it doesn't exist in the Poisson case) and when passing that through TSFC you get a no-op. The kernel bodies for these are empty.

I implemented it this way because I wanted to keep the PC somewhat general. Whether you're solving a mixed Poisson/Helmholtz, or one reverses the typical ordering of the function spaces (HDiv * L^2 or L^2 * HDiv), the reconstruction expressions work for all cases, except some terms may be "zero."

Ideally, the best way to handle this would be to create some sort of `Visitor` that removes all the `noop` (zero-valued tensors) nodes in an expression and then reconstruct the expression with those nodes removed. However, I haven't implemented such a thing and I'll need some time to figure that out. I figured this is good enough for now.

firedrake/slate/slac/utils.py
```diff
@@ -137,6 +136,37 @@ def visit_Symbol(self, o, *args, **kwargs):
...
+                    break
+            else:
+                acyclic = True
+                del dependency_graph[tensor]
+                sorted_exprs.append(tensor)
+
+        if not acyclic:
+            raise RuntimeError("A cyclic dependency found.")
+
+    return sorted_exprs
```
wence-: How about:
```python
def topo_sort(exprs, depgraph):
    depgraph = OrderedDict(((expr, set(traverse_dags([expr])) - {expr}) for expr in exprs))
    schedule = []
    while depgraph:
        for expr in depgraph:
            if not depgraph[expr].intersection(depgraph):
                schedule.append(expr)
                break
        else:
            raise ValueError("Expression graph is cyclic")
        depgraph.pop(expr)
    return schedule
```

firedrake/slate/slac/compiler.py
wence-: I don't understand, you make `cxt_kernel.tsfc_kernels` like this:

`cxt_kernel.tsfc_kernels = compile_form(form)`

if `form` evaluates to a no-op, then `compile_form` returns an empty tuple, so the iteration `for split_kernel in cxt_kernel.tsfc_kernels` will never execute the body.

e.g.
```python
mesh = UnitSquareMesh(1, 1)
V = FunctionSpace(mesh, BrokenElement(FiniteElement("RT", triangle, 1)))

Q = FunctionSpace(mesh, "DG", 0)

W = V*Q

u, p = TrialFunctions(W)

v, q = TestFunctions(W)

a = (dot(u, v) + div(u)*q + div(v)*p)*dx

from firedrake.formmanipulation import split_form

splits = dict(split_form(a))
compile_form(splits[(1, 1)], "foo") => ()
```

Or am I completely missing something?

firedrake/slate/slac/utils.py
thomasgibson: Thanks, I like this version more. I'll update.

firedrake/slate/slac/compiler.py
thomasgibson: No you're right, I am confusing a time when I was messing around with TSFC split kernels using methods imported directly from TSFC, not `tsfc_interface.py`. I'll remove the special casing here.

However, the "zero-tensor" thing is still an issue. While no assembly calls are made, a `Tensor` temporary with the right shape but initialised as zero is still created and used in the generated code.

wence-: That suggests that you need to prune earlier when you compile the form and notice that it's zero.

thomasgibson:
Last major modification: originally I fused coefficient temporary loops by grouping coefficients by function space. While it works in most cases, it doesn't fuse loops for coefficients in different spaces despite having the same node/dof extent.

That last update organizes coefficient loops by node/dof extent strictly (not just examining the function space).

firedrake/slate/slac/compiler.py
thomasgibson: Yes, but once I have an expression, and I know a particular operand is zero, I'm not sure how to remove the node and reconstruct the expression. If `O` is a zero tensor, I want to transform: `A * O + B.T * D.inv * B - O` into `B.T * D.inv * B`. I need to have a think about how to do this in code.

wence-: OK, I see.  I think this is the wrong time to do that.  Here's an idea (not for this PR).  You split the expression compilation into multiple passes.  In the first pass, you visit the terminals of the DAG and  compile the form using tsfc, you then reconstruct a new type of terminal with this compiled form.  At this point when you reconstruct the DAG, you can put any zero simplification in.

Now you have a DAG where there are no forms needed to compile and the next pass just needs to glue to kernel calls together.

This is morally quite similar to how tsfc translates from UFL.  First, replace finite element basis function evaluation with their finat representation, then translate the gem to code.

firedrake/slate/slac/utils.py
```diff
@@ -137,6 +138,47 @@ def visit_Symbol(self, o, *args, **kwargs):
...
+    schedule = []
+    while dependency_graph:
+        for expr in list(dependency_graph.keys()):
+            if not dependency_graph[expr].intersection(dependency_graph):
+                schedule.append(expr)
+                break
+        else:
+            raise RuntimeError("A cyclic dependency is found.")
+
+        del dependency_graph[expr]
```
miklos1: This algorithm is, I guess, no worse than the original here, but _come on_: DFS is also simple and much faster.

firedrake/slate/slac/utils.py
```diff
@@ -137,41 +136,61 @@ def visit_Symbol(self, o, *args, **kwargs):
...
-
-    :arg expr_dags: an iterable of Slate nodes that make up the
-                    DAG of a given Slate expression.
-    """
-    result = collections.Counter(expr_dags)
-    for tensor in traverse_dags(expr_dags):
-        result.update(tensor.operands)
-
-    return result
+    return expanded_dag
```
wence-: Is it obvious why you replace the `traverse_dags` function with this renamed function that appears to nothing different except to only visit a single expression (so treats all DAGS as single-root), and isn't iterable, but instead makes and returns a list?

firedrake/slate/slac/kernel_builder.py
```diff
@@ -27,52 +25,53 @@ def __init__(self, expression, tsfc_parameters=None):
...
-
-        # Collect terminals and expressions
+        # Collect terminals, expressions, and reference counts
         temps = OrderedDict()
         action_coeffs = OrderedDict()
-        tensor_ops = []
         seen_coeff = set()
-        for tensor in traverse_dags([expression]):
-            if isinstance(tensor, Tensor):
+        expression_dag = expand_dag(expression)
```
wence-: FWIW, the name `expand_dag` has negative connotations, you're not actually expanding the dag, you're just eagerly evaluating the traversal of the dag.

firedrake/slate/slac/utils.py
thomasgibson: I renamed it to better reflect what I'm actually using it for.

Also I felt that it wasn't necessary for me to always cast the argument to a list (`traverse_dags([foo])`) when I only ever send in one expression at a time. It made sense to me to rewrite the function to take a single argument. If you prefer the old name/implementation, then I can revert it.

firedrake/slate/slac/kernel_builder.py
thomasgibson: I see. So I guess the old name is better? I guess I just misunderstood.

firedrake/slate/slac/compiler.py
```diff
@@ -65,159 +195,253 @@ def compile_expression(slate_expr, tsfc_parameters=None):
...
+                block_btm = ast.Block(ext_btm, open_scope=True)
+                statements.append(ast.If(ast.Eq(mesh_layer_sym, layer),
+                                         (block_btm,)))
+
+            if ext_top:
+                statements.append(ast.FlatBlock("/* Top calls */\n"))
+                layer = num_layers - 1
+                block_top = ast.Block(ext_top, open_scope=True)
+                statements.append(ast.If(ast.Eq(mesh_layer_sym, layer),
+                                         (block_top,)))
```
wence-: This kind of repetition of the same code looks to ripe for simplification.  What about:

```python
bottom = ast.Block(int_top + ext_btm, open_scope=True)
top = ast.Block(int_btm + ext_top, open_scope=True)
rest = ast.Block(int_btm + int_top, open_scope=True)
statements.append(ast.If(ast.Eq(mesh_layer_sym, 0), 
                         bottom,
                         ast.If(ast.Eq(mesh_layer_sym, num_layers - 1),
                                top,
                                rest)))
```

wence-: Having an empty statement in the generated code might look a little less pretty, but the compiler will elide it, so it'll be fine.

firedrake/slate/slac/compiler.py
```diff
@@ -65,159 +195,253 @@ def compile_expression(slate_expr, tsfc_parameters=None):
...
+        if_ext = ast.Eq(ast.Symbol(cell_facet_sym, rank=(it_sym,)), 0)
+        if_int = ast.Eq(ast.Symbol(cell_facet_sym, rank=(it_sym,)), 1)
+        if int_calls and ext_calls:
+            else_if = ast.If(if_ext, (ast.Block(ext_calls, open_scope=True),))
+            body = ast.If(if_int, (ast.Block(int_calls, open_scope=True),
+                                   else_if))
+        elif int_calls:
+            body = ast.If(if_int, (ast.Block(int_calls, open_scope=True),))
+        elif ext_calls:
+            body = ast.If(if_ext, (ast.Block(ext_calls, open_scope=True),))
```
wence-: Again here I would just write:
```python
if int_calls:
    body = ast.Block(ast.If(if_int, ...))
if ext_calls:
   body = ast.Block(ast.If(...))
```

firedrake/slate/slac/compiler.py
```diff
@@ -41,18 +45,144 @@
...
+        """
+        mesh_layer_types = ["interior_facet_horiz_top",
+                            "interior_facet_horiz_bottom",
+                            "exterior_facet_bottom",
+                            "exterior_facet_top"]
+        return any(cxt_k.original_integral_type in mesh_layer_types
+                   for cxt_k in self.context_kernels)
+
+
 def compile_expression(slate_expr, tsfc_parameters=None):
```
wence-: Even with all the new comments, this function is now very long (almost 350 lines).  It could probably be split into pieces and be easier to follow.

One logical splitting might be to follow the splitting in the generated macro kernel:

1. A pass to declare temporaries
2. A pass to initialise temporaries
3. Generate the linear algebra expression
4. Glue everything together.

WDYT?

thomasgibson: Thanks for pointing this out. Yeah, I've been wondering if it would be better to split everything up. The outline you just presented is very similar to what I've imagined. Item 3 is already its own method. How about I separate 1 and 2 (possibly combine them together if it can be done cleanly) into their own methods and let `compile_expression` glue everything together?

thomasgibson: @wence- I went ahead an split `compile_expression` into separate stages. It was primarily code re-factorization. Let me know how you feel about this latest change.

thomasgibson:
This branch is now up-to-date. I believe I've addressed all previous comments. Let me know if this is ready for merging or needs more changes.

thomasgibson:
Review please :)

firedrake/slate/slac/kernel_builder.py
```diff
@@ -1,19 +1,39 @@
...
+                     the vector temporary for assignment.
+:param shape: A singleton with an integer describing the shape of
+              the coefficient temporary.
+:param coefficient: The :class:`ufl.Coefficient` containing the
+                    relevant data to be placed into the temporary.
+"""
+
+
+class LocalKernelBuilder(object):
+    """The primary helper class for constructing cell-local linear
```
thomasgibson: I decided to not split the builder class after all, that's the only real change the last couple of commits did.

PR #1144, opened 2017-09-18
Higher order octahedral meshes

colinjcotter:
Higher order octahedral meshes

This fixes a previous xfail noted by Lawrence, that there was a
convergence loss for higher order meshes near the pole.

This turned out to be because the coordinate mapping from the
octahedron to the sphere always maps the top four triangles to
pi/2 sectors. The shape of the edges doesn't matter for P1
coordinates, but higher order coordinates lead to a curved edge
around the cell and these cells never converge to flat triangles
as the mesh is refined. I fixed this by making a second
coordinate field that is interpolated from the equivalent P1
mesh, and then pushed out radially to the sphere. This mesh does
not have nodes aligned along lines of constant latitude, so the final
step is to blend between the two meshes. For z/R<z0, we keep the
latitudinally aligned mesh. For z/r>z0 we blend between the two
meshes using a C_infty ramping function, so that near the pole,
the mesh looks like the second coordinate field.

### DIFF
--- a/firedrake/utility_meshes.py
+++ b/firedrake/utility_meshes.py
@@ -958,13 +958,14 @@ def OctahedralSphereMesh(radius, refinement_level=0, degree=1,
     r = ufl.sqrt(Xlow[0]**2 + Xlow[1]**2 + Xlow[2]**2)
     Xradial = m.coordinates.copy(deepcopy=True).interpolate(Xlow/r)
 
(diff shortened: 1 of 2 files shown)


colinjcotter:
Can I merge now?

wence-:
Go ahead.

PR #1148, opened 2017-09-21
fix slepc in installer

juan42:


dham:
I think this is not totally upgrade safe. We want existing installs to transition automatically to the new repo. See elsewhere in the install script for how we have done this previously.

PR #1155, opened 2017-10-16
Add Slate node for assembled data

thomasgibson:
This PR adds a new class: `AssembledVector`. This object associates firedrake functions with local element vectors. With this new class, the `Action` node is now obsolete and the `Mul` node can now handle these "actions." I also removed the `Sub` node, as it was basically just duplicate code which could be eliminated by writing A - B as A + -B.

TL;DR: Slate is officially closed under all binary operations. No need for the hacky `Action` node.

### DIFF
--- a/firedrake/linear_solver.py
+++ b/firedrake/linear_solver.py
@@ -119,7 +119,7 @@ class LinearSolver(solving_utils.ParametersMixin):
             bc.apply(b)
         from firedrake.assemble import _assemble
         if isinstance(self.A.a, slate.TensorBase):
-            return _assemble(self.A.a * b)
+            return _assemble(self.A.a * slate.AssembledVector(b))
         else:
             return _assemble(ufl.action(self.A.a, b))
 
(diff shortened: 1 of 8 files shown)


miklos1:
I hope you don't require users to write A + -B instead of A - B.

miklos1:
How is `AssembledVector` any different from a `Tensor` made of a function?

thomasgibson:
> I hope you don't require users to write A + -B instead of A - B.

Of course not. Look at the `__sub__` method.

> How is `AssembledVector` any different from a `Tensor` made of a function?

The most obvious difference is that you can't create a `Tensor` from a function. If you want to combine the two and let `Tensor` handle both `form` and `function`, this would require some special casing in the class which might be a bit ugly. Do you prefer this over what's currently proposed?

miklos1:
> Do you prefer this over what's currently proposed?

No, thanks, it's clear now.

PR #1160, opened 2017-11-08
Doc update

wence-:
Add a load of new bib entries to the firedrake-apps bib.  Also update firedrake citing info for newly appeared papers.

Additionally, address @dorugeber's comment about splitting out the "citing firedrake" page separately by doing so.  I haven't made an attempt to really improve the text...


dorugeber:
Looks good.

PR #1166, opened 2017-11-15
Better error checking for mismatching meshes

### LINKED ISSUE #1165: Code paths for mismatching meshes need to be tightened
Because we assume that error checking is done in firedrake, we turn off some of pyop2's error checking (oops!).  In particular if some of the coefficients in a form live on a different mesh topology to the integration domain.  As a result, we can get segfaults when accessing invalid data.

wence-:
Fixes #1165.

### DIFF
--- a/firedrake/assemble.py
+++ b/firedrake/assemble.py
@@ -1,6 +1,7 @@
 import numpy
 import ufl
 from collections import defaultdict
+from itertools import chain
 
 from pyop2 import op2
 from pyop2.base import collecting_loops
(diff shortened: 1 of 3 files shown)


firedrake/assemble.py
```diff
@@ -186,13 +187,19 @@ def _assemble(f, tensor=None, bcs=None, form_compiler_parameters=None,
...
+    topology = f.ufl_domains()[0].topology
     for m in f.ufl_domains():
         # Ensure mesh is "initialised" (could have got here without
         # building a functionspace (e.g. if integrating a constant)).
         m.init()
-        if m.topology != f.ufl_domains()[0].topology:
+        if m.topology != topology:
             raise NotImplementedError("All integration domains must share a mesh topology.")
 
+    for o in chain(f.arguments(), f.coefficients()):
```
miklos1: I think both `f.arguments()` and `f.coefficients()` return a tuple, so you could just write `f.arguments() + f.coefficients()`.

PR #1167, opened 2017-11-15
Make all xfails strict

wence-:
Means that an XPASS turns into a test failure.

### DIFF
--- a/tests/regression/test_real_space.py
+++ b/tests/regression/test_real_space.py
@@ -231,7 +231,6 @@ def test_real_space_mixed_assign():
     assert np.allclose(q.dat.data_ro, 2.0)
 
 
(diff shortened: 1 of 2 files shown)


tests/regression/test_real_space.py
```diff
@@ -231,15 +231,14 @@ def test_real_space_mixed_assign():
     assert np.allclose(q.dat.data_ro, 2.0)
 
 
-@pytest.mark.xfail(reason="Error checking in PyOP2 accesses property of None")
```
miklos1: This is the recently usual XPASS, I guess...

wence-:
It should, and I realise it can...

wence-:
Done.

PR #1170, opened 2017-11-29
Static condensation for CG problems

thomasgibson:
I think this is ready to be reviewed. The issue with strong Dirichlet boundary conditions still needs to be addressed in the near future.

This PR adds a Slate-implementation of static condensation for CG problems. I've included some tests and minor edits as well.


firedrake/slate/preconditioners/static_condensation.py
```diff
@@ -0,0 +1,240 @@
...
+                g = interpolate(bc.function_arg, V_facet)
+            else:
+                g = bc.function_arg
+            bcs.append(DirichletBC(V_facet, g, bc.sub_domain))
+
+        if bcs:
+            msg = ("Currently strong bcs are not handled correctly. "
+                   "The solver may still converge with this PC if an "
+                   "appropriate iterative method is used. ")
+            log(WARNING, msg)
```
thomasgibson: I chose to log a warning over raising an exception because methods using this PC (with Dirichlet conditions) still converge (consistently in 4 GMRES iterations). We can have this "feature" exposed or just raise a `NotImplementedError` until we can apply strong boundary conditions in Slate. Thoughts?

thomasgibson:
Closing this PR for now. If we want something like this in the future, then I'll revive this.

PR #1171, opened 2017-12-01
Block method for Slate expressions

thomasgibson:
I've added a `.block()` method which pulls out blocks of a Slate expression. I use the argument indices to specify a particular block.

### DIFF
--- a/firedrake/slate/slac/compiler.py
+++ b/firedrake/slate/slac/compiler.py
@@ -498,6 +498,33 @@ def metaphrase_slate_to_cpp(expr, temps, prec=None):
 
         return parenthesize(result, expr.prec, prec)
 
+    elif isinstance(expr, slate.IndexedTensor):
+        tensor, = expr.operands
+        idx = expr._idx
+        if expr.rank == 1:
+            # Vector segment
+            idx, = idx
+            shape = tensor.shapes[0][idx]
+            start = sum(tensor.shapes[0][:idx])
+            result = "(%s).segment<%d>(%d)" % (metaphrase_slate_to_cpp(tensor,
+                                                                       temps,
+                                                                       expr.prec),
+                                               shape, start)
+        else:
+            # Matrix block
+            ridx, cidx = idx
+            rshape = tensor.shapes[0][ridx]
+            rstart = sum(tensor.shapes[0][:ridx])
+            cshape = tensor.shapes[1][cidx]
+            cstart = sum(tensor.shapes[1][:cidx])
+            result = "(%s).block<%d, %d>(%d, %d)" % (metaphrase_slate_to_cpp(tensor,
+                                                                             temps,
+                                                                             expr.prec),
+                                                     rshape, cshape,
+                                                     rstart, cstart)
+
+        return parenthesize(result, expr.prec, prec)
+
     else:
         raise NotImplementedError("Type %s not supported.", type(expr))
 
--- a/firedrake/slate/slac/compiler.py
+++ b/firedrake/slate/slac/compiler.py
@@ -504,15 +504,23 @@ def slate_to_cpp(expr, temps, prec=None):
             ridx, cidx = indices
         except ValueError:
             ridx, = indices
-            cidx = (0,)
+            cidx = 0
+        rids = as_tuple(ridx)
+        cids = as_tuple(cidx)
+
+        # Check if indices are non-contiguous
+        if not all(all(ids[i] + 1 == ids[i + 1] for i in range(len(ids) - 1))
+                   for ids in (rids, cids)):
+            raise NotImplementedError("Non-contiguous blocks not implemented")
+
         rshape = expr.shape[0]
-        rstart = sum(tensor.shapes[0][:min(as_tuple(ridx))])
+        rstart = sum(tensor.shapes[0][:min(rids)])
         if expr.rank == 1:
             cshape = 1
             cstart = 0
         else:
             cshape = expr.shape[1]
-            cstart = sum(tensor.shapes[1][:min(as_tuple(cidx))])
+            cstart = sum(tensor.shapes[1][:min(cids)])
 
         result = "(%s).block<%d, %d>(%d, %d)" % (slate_to_cpp(tensor,
                                                               temps,
--- a/firedrake/slate/slac/compiler.py
+++ b/firedrake/slate/slac/compiler.py
@@ -497,7 +497,7 @@ def slate_to_cpp(expr, temps, prec=None):
 
         return parenthesize(result, expr.prec, prec)
 
-    elif isinstance(expr, slate.IndexedTensor):
+    elif isinstance(expr, slate.Block):
         tensor, = expr.operands
         idx = expr._idx
         try:
--- a/firedrake/slate/slac/compiler.py
+++ b/firedrake/slate/slac/compiler.py
@@ -139,8 +139,7 @@ def generate_kernel_ast(builder, statements, declared_temps):
     # Generate the complete c++ string performing the linear algebra operations
     # on Eigen matrices/vectors
     statements.append(ast.FlatBlock("/* Linear algebra expression */\n"))
-    cpp_string = ast.FlatBlock(metaphrase_slate_to_cpp(slate_expr,
-                                                       declared_temps))
+    cpp_string = ast.FlatBlock(slate_to_cpp(slate_expr, declared_temps))
     statements.append(ast.Incr(result_sym, cpp_string))
 
     # Generate arguments for the macro kernel
@@ -221,7 +220,7 @@ def auxiliary_temporaries(builder, declared_temps):
     for exp in builder.aux_exprs:
         if exp not in declared_temps:
             t = ast.Symbol("auxT%d" % len(declared_temps))
-            result = metaphrase_slate_to_cpp(exp, declared_temps)
+            result = slate_to_cpp(exp, declared_temps)
             tensor_type = eigen_matrixbase_type(shape=exp.shape)
             statements.append(ast.Decl(tensor_type, t))
             statements.append(ast.FlatBlock("%s.setZero();\n" % t))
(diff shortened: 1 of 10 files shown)


firedrake/slate/slate.py
```diff
@@ -296,6 +305,107 @@ def _key(self):
...
+
+        assert len(idx) == tensor.rank, (
+            "Number of indices must match the tensor rank."
+        )
+
+        if not tensor.is_mixed:
+            raise ValueError("Can only index block (mixed) tensors")
+
+        fs_extent = len(tensor.arguments()[0].function_space())
+        if any(i not in range(0, fs_extent) for i in idx):
```
wence-: This seems wrong.  I think you want:

```python
for arg, i in zip(tensor.arguments(), idx):
    if not 0 <= i < len(arg.function_space()):
        raise ValueError(...)
```

firedrake/slate/slate.py
```diff
@@ -296,6 +305,107 @@ def _key(self):
...
+
+        if tensor.rank == 0:
+            raise ValueError("Cannot index scalars.")
+
+        assert len(idx) == tensor.rank, (
+            "Number of indices must match the tensor rank."
+        )
+
+        if not tensor.is_mixed:
+            raise ValueError("Can only index block (mixed) tensors")
```
wence-: I think it makes sense to be able to index non-mixed (assuming that the indices are all 0).  This makes writing code outside much more natural (you don't have to special-case outside).  

You code do:

```python
def __new__(cls, tensor, idx):
    ...
    if not tensor.is_mixed:
        assert all(i == 0 for i in idx)
        return tensor
    return super().__new__(cls)
```

firedrake/slate/slate.py
```diff
@@ -296,6 +305,107 @@ def _key(self):
...
+        """Constructor for the IndexedTensor class."""
+        if not isinstance(tensor, TensorBase):
+            raise TypeError("Can only index Slate expressions.")
+
+        if tensor.rank == 0:
+            raise ValueError("Cannot index scalars.")
+
+        assert len(idx) == tensor.rank, (
+            "Number of indices must match the tensor rank."
+        )
```
wence-: This case also catches indexing a "scalar" with more than zero indices.  So again, I think it makes sense to allow `IndexedTensor(scalar, ())`

firedrake/slate/slate.py
```diff
@@ -332,15 +442,12 @@ def __init__(self, form):
         if r not in (0, 1, 2):
             raise NotImplementedError("No support for tensors of rank %d." % r)
 
-        # Checks for positive restrictions on integrals
-        integrals = form.integrals()
-        mapper = CheckRestrictions()
-        for it in integrals:
-            map_integrand_dags(mapper, it)
+        # Remove any negative restrictions and replace with zero
+        _form = map_integrand_dags(RemoveNegativeRestrictions(), form)
```
wence-: You don't need `form` after this call, so just reassign `form = map_integrand_dags(..., form)`

firedrake/slate/slac/compiler.py
```diff
@@ -477,24 +476,51 @@ def metaphrase_slate_to_cpp(expr, temps, prec=None):
...
-                               metaphrase_slate_to_cpp(B, temps, expr.prec))
+                               slate_to_cpp(B, temps, expr.prec))
+
+        return parenthesize(result, expr.prec, prec)
+
+    elif isinstance(expr, slate.IndexedTensor):
+        tensor, = expr.operands
+        idx = expr._idx
+        if expr.rank == 1:
+            # Vector segment
```
wence-: WAT?!  How does eigen not treat vectors just like matrices :P.

wence-:
So this doesn't solve the (difficult) problem of deciding which sub-blocks actually need to be assembled to provide a particular indexed block.  But perhaps that make sense because you're usually needing all of them and then pulling them apart?

thomasgibson:
Correct. In all cases I've looked at, all blocks are used in computations. I think it would be nice to detect which bit needs assembly and only compute what is necessary, but I haven't gotten around to implementing it just yet. I'll update soon addressing your comments!

firedrake/slate/slac/compiler.py
thomasgibson: Ah, I just remembered that all of these `Eigen::MatrixBase` objects (which includes vectors as well) have a `block` method. So I'm pretty sure I can squash this into one stream of code for all cases.

thomasgibson:
I've pushed some changes. Let me know if this looks alright

wence-:
My only real concern is that going forward, we might want to index slate expressions by quadrature-point or basis function indices.  So I think it would be good to think about what we want the indexing space of slate expressions to be.  Should it be the same as UFL forms?  This isn't quite that, since here the index space is of arguments, whereas for forms it is the `ufl_shape` of the expression.

I'm not opposed to the idea of indexing tensors, I just want to have a discussion about what type of indexing we want, and whether or not this approach precludes indexing over different spaces in the future.

Perhaps we need to having different index types? (That feels ugly though).

thomasgibson:
Once build issues are resolved, I think this is ready. As we've discussed, I decided to remove indexing for now but provide the same feature as a `.block()` method. This takes argument indices just like the `ExtractSubBlock.split()` method in `formmanipulation.py`.

This works for extracting blocks over one or multiple test/trial spaces. I've tested in this my own implementations of static condensation for CG problems and fully formed hybridized mixed/DG problems. I added some tests in Firedrake, but it's currently not possible to test assembly of individual blocks spanning multiple spaces. 'Cause, you know, mixed mats/dats.

firedrake/slate/slate.py
```diff
@@ -140,6 +141,42 @@ def inv(self):
...
+                raise ValueError("Trial indices found for a 1-tensor. "
+                                 "If you're attempting to extract a block "
+                                 "over multiple test spaces, provide a tuple.")
+            arg_indices = (to_tuple(test_idx),)
+
+        else:
+            assert self.rank == 2, "Rank %d not supported." % self.rank
+            arg_indices = (to_tuple(test_idx), to_tuple(trial_idx))
+
+        return Block(self, arg_indices)
```
wence-: I'd prefer to be liberal in what one receives here.  I think you should be allowed to `block` a 0-tensor, by passing `tensor.block(()) => tensor`.  This way you don't have to write special-case code outside.

Similarly, you could (like `ExtractSubBlock`) just accept any number of indices, but only handle the ones you care about.

firedrake/slate/slate.py
```diff
@@ -296,6 +333,116 @@ def _key(self):
...
+                       for i in idx)
+            return tensor
+
+        return super().__new__(cls)
+
+    def __init__(self, tensor, arg_indices):
+        """Constructor for the Block class."""
+        assert len(arg_indices) == tensor.rank, (
+            "Number of indices must match the tensor rank."
+        )
```
wence-: I continue to be sad at the extra line breaks.  Just write `assert foo, "bar"`.  We don't have a strict 80 character line limit.

firedrake/slate/slate.py
```diff
@@ -296,6 +333,116 @@ def _key(self):
...
+        for arg, idx in zip(tensor.arguments(), arg_indices):
+            assert all(0 <= i < len(arg.function_space())
+                       for i in idx), "Index out of bounds."
+
+        super(Block, self).__init__()
+        self.operands = (tensor,)
+        self._blocks = dict(zip((0, 1), arg_indices))
```
wence-: I realise a better way to write this is `dict(enumerate(arg_indices))`

firedrake/slate/slate.py
```diff
@@ -296,6 +333,116 @@ def _key(self):
...
+        self._arg_indices = arg_indices
+
+    @cached_property
+    def _split_arguments(self):
+        """Splits the function space and stores the component
+        spaces determined by the indices.
+        """
+        from firedrake.functionspace import (FunctionSpace,
+                                             MixedFunctionSpace)
```
wence-: No need for a line break here.

firedrake/slate/slate.py
wence-: But if you're going to do all this type-checking, best to do it in `Block`, since then you get it when either the user calls `Tensor.block`, or else when they call `Block` directly.

tests/slate/test_assemble_tensors.py
```diff
@@ -170,6 +170,76 @@ def test_mixed_argument_tensor(mesh):
...
+    A20ref = assemble(gammar('+')*jump(u, n=n)*dS)
+    A22ref = assemble(lambdar*gammar*ds)
+
+    assert np.allclose(A00.M.values, A00ref.M.values, rtol=1e-14)
+    assert np.allclose(A01.M.values, A01ref.M.values, rtol=1e-14)
+    assert np.allclose(A02.M.values, A02ref.M.values, rtol=1e-14)
+    assert np.allclose(A10.M.values, A10ref.M.values, rtol=1e-14)
+    assert np.allclose(A11.M.values, A11ref.M.values, rtol=1e-14)
+    assert np.allclose(A20.M.values, A20ref.M.values, rtol=1e-14)
+    assert np.allclose(A22.M.values, A22ref.M.values, rtol=1e-14)
```
wence-: Can you test that taking a `block` resulting in a mixed matrix, and then taking a block of that (into a non-mixed matrix) and then assembling that works too?

wence-: I note in passing that if you made the API consistent between form-splitting and tensor blocking, this test could be much reduced in length.

firedrake/slate/slate.py
```diff
@@ -140,6 +141,42 @@ def inv(self):
     def T(self):
         return Transpose(self)
 
+    def block(self, test_idx, trial_idx=None):
```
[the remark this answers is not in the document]
thomasgibson: I chose `block` because that seemed to be the logical choice for name here. I want "a block of a matrix over argument indices (...)" I'm not opposed to renaming it, but there was reason and thought into this.

I think `split` could work. So this probably means this method should return several `Block` objects corresponding to the function arguments provided?

firedrake/slate/slate.py
thomasgibson: Okay I'll update this!

firedrake/slate/slate.py
thomasgibson: Sorry... it's a habit.

thomasgibson:
OK I've updated this branch. The API here should be more similar to form splitting. Test suite now includes more robust block testing.

PR #1173, opened 2017-12-01
Complex Plotting

benSepanski:
Changes made so that plotting (via either pyplot or paraview) visualizes real and imaginary parts of complex functions separately. Changes made to plot.py and output.py

### DIFF
--- a/firedrake/plot.py
+++ b/firedrake/plot.py
@@ -187,7 +187,7 @@ def prepare_functions(function, real=False):
             imaginary_part.rename(function.name() + " (imaginary part)")
             functions = (real_part, imaginary_part)
     else:
-        functions = (function)
+        functions = (function, )
     return functions
 
 def plot(function_or_mesh,
@@ -241,7 +241,7 @@ def plot(function_or_mesh,
             plots.append(p)
         return plots
     else:
-        function = functions[0]
+        function, = functions
     
     try:
         import matplotlib.pyplot as plt
--- a/firedrake/plot.py
+++ b/firedrake/plot.py
@@ -175,11 +175,26 @@ def plot_mesh(mesh, axes=None, **kwargs):
     axes.set_aspect("equal")
     return axes
 
+def prepare_functions(function, real=False):
+    function_data = function.vector().array()
+    if function_data.dtype.kind == "c":
+        real_part = Function(function.function_space(), function_data.real)
+        if real:
+            functions = (real_part,)
+        else:
+            imaginary_part = Function(function.function_space(), function_data.imag)
+            real_part.rename(function.name() + " (real part)")
+            imaginary_part.rename(function.name() + " (imaginary part)")
+            functions = (real_part, imaginary_part)
+    else:
+        functions = (function)
+    return functions
 
 def plot(function_or_mesh,
          num_sample_points=10,
          axes=None,
          plot3d=False,
+         real=False,
          **kwargs):
     """Plot a Firedrake object.
 
@@ -216,25 +231,18 @@ def plot(function_or_mesh,
                             type(function_or_mesh))
         return _plot_mult(functions, num_sample_points, axes=axes, **kwargs)
     # Single Function...
-  
-    #if complex, plot real and imaginary parts separately  
-    function_data = function_or_mesh.vector().array()
-    if function_data.dtype == np.complex128:
-        real_function = Function(function_or_mesh.function_space(), 
-                                 function_data.real) 
-        imag_function = Function(function_or_mesh.function_space(), 
-                                 function_data.imag)
-        
-        real_plot = complex_plot(real_function, num_sample_points, axes=axes, 
-                                 plot3d=plot3d, **kwargs)
-        real_plot.set_title(real_plot.get_title() + 'real part')
-        imag_plot = complex_plot(imag_function, num_sample_points, axes=axes, 
-                                 plot3d=plot3d, **kwargs)
-        imag_plot.set_title(imag_plot.get_title() + 'imaginary part')
-        return [real_plot, imag_plot]
-
-    # else plot as normal
     function = function_or_mesh
+    functions = prepare_functions(function, real)
+    if not real:
+        plots = []
+        for function in functions:
+            p = plot(function,  num_sample_points, axes=axes, plot3d=plot3d, real=True, **kwargs)
+            p.set_title(function.name())
+            plots.append(p)
+        return plots
+    else:
+        function = functions[0]
+    
     try:
         import matplotlib.pyplot as plt
         from matplotlib import cm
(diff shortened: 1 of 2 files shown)


wence-:
Thanks!  I have some comments on the implementation.  Let's just consider the paraview output first.

I think we can get away with not replicating quite so much code.  In the complex case, the only thing that needs to change is that when writing the function data array one spits out both the real and imaginary parts.  I think this can be done by changing `write_array` and `write_array_descriptor`.

We just have to be careful to only write the real part of the coordinate field.

I think this patch from master does the right thing:

```diff
diff --git a/firedrake/output.py b/firedrake/output.py
index f702d880..384a572a 100644
--- a/firedrake/output.py
+++ b/firedrake/output.py
@@ -167,36 +167,56 @@ def get_byte_order(dtype):
             ">": "BigEndian"}[dtype.byteorder]
 
 
-def write_array(f, ofunction):
-    array = ofunction.array
-    numpy.uint32(array.nbytes).tofile(f)
-    if get_byte_order(array.dtype) == "BigEndian":
-        array = array.byteswap()
-    array.tofile(f)
-
-
-def write_array_descriptor(f, ofunction, offset=None, parallel=False):
+def prepare_ofunction(ofunction, real):
     array, name, _ = ofunction
-    shape = array.shape[1:]
-    ncmp = {0: "",
-            1: "3",
-            2: "9"}[len(shape)]
-    typ = {numpy.dtype("float32"): "Float32",
-           numpy.dtype("float64"): "Float64",
-           numpy.dtype("int32"): "Int32",
-           numpy.dtype("int64"): "Int64",
-           numpy.dtype("uint8"): "UInt8"}[array.dtype]
-    if parallel:
-        f.write(('<PDataArray Name="%s" type="%s" '
-                 'NumberOfComponents="%s" />' % (name, typ, ncmp)).encode('ascii'))
+    if array.dtype.kind == "c":
+        if real:
+            arrays = (array.real, )
+            names = (name, )
+        else:
+            arrays = (array.real, array.imag)
+            names = (name + "(real part)", name + "(imaginary part")
     else:
-        if offset is None:
-            raise ValueError("Must provide offset")
-        f.write(('<DataArray Name="%s" type="%s" '
-                 'NumberOfComponents="%s" '
-                 'format="appended" '
-                 'offset="%d" />\n' % (name, typ, ncmp, offset)).encode('ascii'))
-    return 4 + array.nbytes     # 4 is for the array size (uint32)
+        arrays = (array, )
+        names = (name, )
+    return arrays, names
+
+
+def write_array(f, ofunction, real=False):
+    arrays, _ = prepare_ofunction(ofunction, real=real)
+    for array in arrays:
+        numpy.uint32(array.nbytes).tofile(f)
+        if get_byte_order(array.dtype) == "BigEndian":
+            array = array.byteswap()
+        array.tofile(f)
+
+
+def write_array_descriptor(f, ofunction, offset=None, parallel=False, real=False):
+    arrays, names = prepare_ofunction(ofunction, real)
+    nbytes = 0
+    for array, name in zip(arrays, names):
+        shape = array.shape[1:]
+        ncmp = {0: "",
+                1: "3",
+                2: "9"}[len(shape)]
+        typ = {numpy.dtype("float32"): "Float32",
+               numpy.dtype("float64"): "Float64",
+               numpy.dtype("int32"): "Int32",
+               numpy.dtype("int64"): "Int64",
+               numpy.dtype("uint8"): "UInt8"}[array.dtype]
+        if parallel:
+            f.write(('<PDataArray Name="%s" type="%s" '
+                     'NumberOfComponents="%s" />' % (name, typ, ncmp)).encode('ascii'))
+        else:
+            if offset is None:
+                raise ValueError("Must provide offset")
+            offset += nbytes
+            nbytes += (4 + array.nbytes)  # 4 is for the array size (uint32)
+            f.write(('<DataArray Name="%s" type="%s" '
+                     'NumberOfComponents="%s" '
+                     'format="appended" '
+                     'offset="%d" />\n' % (name, typ, ncmp, offset)).encode('ascii'))
+    return nbytes
 
 
 def get_vtu_name(basename, rank, size):
@@ -431,7 +451,7 @@ class File(object):
                      'NumberOfCells="%d">\n' % (num_points, num_cells)).encode('ascii'))
             f.write(b'<Points>\n')
             # Vertex coordinates
-            offset += write_array_descriptor(f, coordinates, offset=offset)
+            offset += write_array_descriptor(f, coordinates, offset=offset, real=True)
             f.write(b'</Points>\n')
 
             f.write(b'<Cells>\n')
@@ -452,7 +472,7 @@ class File(object):
             # Appended data must start with "_", separating whitespace
             # from data
             f.write(b'_')
-            write_array(f, coordinates)
+            write_array(f, coordinates, real=True)
             write_array(f, connectivity)
             write_array(f, offsets)
             write_array(f, types)
@@ -476,7 +496,7 @@ class File(object):
 
             f.write(b'<PPoints>\n')
             # Vertex coordinates
-            write_array_descriptor(f, coordinates, parallel=True)
+            write_array_descriptor(f, coordinates, parallel=True, real=True)
             f.write(b'</PPoints>\n')
 
             f.write(b'<PCells>\n')
```

Do you agree?

This way, we don't replicate the code for writing vtus.

I suspect a similar thing can be done for the matplotlib case too.

benSepanski:
I agree, that seems to work. I have edited the matplotlib file to remove duplicate code in a similar manner. I did notice that on matplotlib, the default of plotting a function doesn't output a title. I went ahead and forced it to output <function.name() + " (real/imaginary part)"> so that people can tell which one is which. Is there a preferred way to tell the plots apart?

firedrake/plot.py
```diff
@@ -175,11 +175,26 @@ def plot_mesh(mesh, axes=None, **kwargs):
...
+        real_part = Function(function.function_space(), function_data.real)
+        if real:
+            functions = (real_part,)
+        else:
+            imaginary_part = Function(function.function_space(), function_data.imag)
+            real_part.rename(function.name() + " (real part)")
+            imaginary_part.rename(function.name() + " (imaginary part)")
+            functions = (real_part, imaginary_part)
+    else:
+        functions = (function)
```
wence-: You need a trailing comma here: `(function, )`

firedrake/plot.py
```diff
@@ -217,6 +232,17 @@ def plot(function_or_mesh,
...
+    functions = prepare_functions(function, real)
+    if not real:
+        plots = []
+        for function in functions:
+            p = plot(function,  num_sample_points, axes=axes, plot3d=plot3d, real=True, **kwargs)
+            p.set_title(function.name())
+            plots.append(p)
+        return plots
+    else:
+        function = functions[0]
```
wence-: Better to restructure like this: `function, = functions`. That way we get an error if this list ever has more than one entry.

dham:
I'm happy with the overall design of this. I agree this approach is the right one.

benSepanski:
Requested changes made

wence-:
Thanks!  Can you just fix the flake8 errors in the changed files:  See https://jenkins.ese.ic.ac.uk:1080/blue/organizations/jenkins/firedrakeproject%2Ffiredrake/detail/PR-1173/3/pipeline

In particular:
```
firedrake/plot.py:178:1: E302 expected 2 blank lines, found 1
firedrake/plot.py:193:1: E302 expected 2 blank lines, found 1
firedrake/plot.py:239:31: E241 multiple spaces after ','
firedrake/plot.py:245:1: W293 blank line contains whitespace

firedrake/output.py:214:41: E261 at least two spaces before inline comment
firedrake/output.py:219:18: W291 trailing whitespace
```

You can see these locally by running `make lint` in the firedrake source directory once you've activated the virtualenv.

benSepanski:
I fixed the flake8 errors above.
When I run make lint I get some errors on files I haven't modified:

firedrake/utility_meshes.py:1128:9: E741 ambiguous variable name 'I'
firedrake/mesh.py:235:9: E722 do not use bare except'
firedrake/mesh.py:239:13: E722 do not use bare except'
firedrake/bcs.py:110:17: E722 do not use bare except'

wence-:
Yes, no need to worry about those.  The problem is that those errors have appeared and been fixed in master since this branch was changed.

PR #1175, opened 2017-12-06
value packing

tj-sun:
Necessary changes to accommodate value packing in tsfc.

### DIFF
--- a/firedrake/assemble.py
+++ b/firedrake/assemble.py
@@ -487,14 +487,15 @@ def _assemble(f, tensor=None, bcs=None, form_compiler_parameters=None,
 
             coords = m.coordinates
             args = [kernel, itspace, tensor_arg,
-                    coords.dat(op2.READ, get_map(coords))]
+                    coords.dat(op2.READ, get_map(coords)[op2.i[0]])]
             if needs_orientations:
                 o = m.cell_orientations()
-                args.append(o.dat(op2.READ, get_map(o)))
+                args.append(o.dat(op2.READ, get_map(o)[op2.i[0]]))
             for n in coeff_map:
                 c = coefficients[n]
                 for c_ in c.split():
-                    args.append(c_.dat(op2.READ, get_map(c_)))
+                    m_ = get_map(c_)
+                    args.append(c_.dat(op2.READ, m_ and m_[op2.i[0]]))
             if needs_cell_facets:
                 assert integral_type == "cell"
                 extra_args.append(m.cell_to_facets(op2.READ))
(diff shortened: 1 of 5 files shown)


firedrake/assemble.py
```diff
@@ -487,14 +487,17 @@ def get_map(x, bcs=None, decoration=None):
...
-                args.append(o.dat(op2.READ, get_map(o)))
+                args.append(o.dat(op2.READ, get_map(o)[op2.i[0]]))
             for n in coeff_map:
                 c = coefficients[n]
                 for c_ in c.split():
-                    args.append(c_.dat(op2.READ, get_map(c_)))
+                    if get_map(c_):
+                        args.append(c_.dat(op2.READ, get_map(c_)[op2.i[0]]))
+                    else:
+                        args.append(c_.dat(op2.READ, None))
```
miklos1: Slightly nicer:
```python
m_ = get_map(c_)
args.append(c_.dat(op2.READ, m_ and m_[op2.i[0]]))
```

firedrake/interpolation.py
```diff
@@ -194,9 +194,16 @@ def _interpolator(V, dat, expr, subset):
...
-        args.append(co.dat(op2.READ, co.cell_node_map()))
+        args.append(co.dat(op2.READ, co.cell_node_map()[op2.i[0]]))
     for coefficient in coefficients:
-        args.append(coefficient.dat(op2.READ, coefficient.cell_node_map()))
+        m = coefficient.cell_node_map()
+        if indexed:
+            if m:
+                args.append(coefficient.dat(op2.READ, m[op2.i[0]]))
+            else:
+                args.append(coefficient.dat(op2.READ, None))
```
miklos1: Same here.

firedrake/assemble.py
tj-sun: Good idea. Amended.

firedrake/interpolation.py
tj-sun: Amended,

miklos1:
Oh, just one request: squash history please!

miklos1:
OK, but your squashed history has an ugly commit message, can you fix that, please?

miklos1:
Yes, this is a better commit message, thanks.

tj-sun:
### Performance
[link](https://docs.google.com/spreadsheets/d/1xawM8Q5cDfcs5meoxSdbTzY8RYWFbIVbQdW2DqvkbkU/edit?usp=sharing)

This is assembly of various forms, of various degrees, multiplied by coefficients of various degrees. E.g. for mass form, the code is
```python
def mass(q, p, mesh, nf):
    V = FunctionSpace(mesh, 'CG', q)
    P = FunctionSpace(mesh, 'CG', p)
    u = TrialFunction(V)
    v = TestFunction(V)
    it = dot(v, u)
    f = [Function(P).assign(1.0) for _ in range(nf)]
    return reduce(inner, f + [it])*dx
```
Measurement is for 5 runs, with one warm up run before. Default parameters.

So it appears there isn't much of a difference, and value packing is slightly faster.

PR #1177, opened 2017-12-12
Reinstate flake8 E722 (don't use bare except)

wence-:
Fix up the few places we were being lazy.


scripts/zenodo-bibtex.py
```diff
@@ -77,7 +77,8 @@ def get_bibtex(ids):
     assert firedrake_tag[18] == "."
     digits = map(str, range(10))
     assert all([a in digits for a in (firedrake_tag[10:18] + firedrake_tag[19:])])
-except:
+except:                    # noqa: E722
+    # We really want to swallow everything here
```
miklos1: if you really want to swallow everything, you could just write `except Exception:` I think

wence-:
> Not sure why the build failed though... Jenkins issue?

Bug in patch.  Now fixed?

PR #1178, opened 2017-12-13
Label facet subdomains

thomasgibson:
This PR extends the `cell_to_facets` map on `MeshTopology` to also include labeling of exterior facet subdomains.

As a consequence, Slate can handle exterior facet integrals over specific subdomains (addresses an issue in #975).

### DIFF
--- a/firedrake/dmplex.pyx
+++ b/firedrake/dmplex.pyx
@@ -782,15 +782,16 @@ def cell_facet_labeling(exterior_facets,
                         PETSc.Section cell_numbering,
                         np.ndarray[PetscInt, ndim=2, mode="c"] cell_closures):
     """Computes a labeling for the facet numbers on a particular cell
-    (interior and exterior facet labels). The i-th local facet is
-    represented as:
+    (interior and exterior facet labels with subdomain markers). The
+    i-th local facet is represented as:
 
     cell_facets[c, i]
 
     If `cell_facets[c, i, 0]` is :data:`0`, then the local facet
     :data:`i` is an exterior facet, otherwise if the result is :data:`1`
     it is interior. `cell_facets[c, i, 1]` returns the subdomain marker
-    for the particular facet.
+    for the local facet. This is `-1` for interior facets, and some
+    positive integer for exterior facets.
 
     :arg exterior_facets: A :class:`_Facets` object of `kind` "exterior".
     :arg plex: The DMPlex object representing the mesh topology.
@@ -818,11 +819,13 @@ def cell_facet_labeling(exterior_facets,
     exterior_markers = exterior_facets.markers
     if exterior_markers is None:
         # No subdomains to collect (for example, mesh is a sphere)
-        collect_subdomains = False
+        exterior_subdomains = False
     else:
-        collect_subdomains = True
+        exterior_subdomains = True
+        # Get unique markers
+        exterior_markers = np.unique(exterior_markers)
         sub_domain_ids = dict((sid, plex.getStratumIS(FACE_SETS_LABEL, sid))
-                              for sid in set(exterior_markers))
+                              for sid in exterior_markers)
 
     for c in range(cstart, cend):
         CHKERR(PetscSectionGetOffset(cell_numbering.sec, c, &cell))
@@ -835,10 +838,10 @@ def cell_facet_labeling(exterior_facets,
                 cell_facets[cell, fi, 0] = <np.int8_t>is_interior
 
                 # Collect subdomains (if any)
-                if collect_subdomains:
+                if exterior_subdomains:
                     # If on an exterior facet, search for the subdomain marker
                     if not is_interior:
-                        for marker in sub_domain_ids:
+                        for marker in exterior_markers:
                             # Facet point in subdomain
                             if point in sub_domain_ids[marker].array:
                                 cell_facets[cell, fi, 1] = <np.int8_t>marker
--- a/firedrake/dmplex.pyx
+++ b/firedrake/dmplex.pyx
@@ -815,8 +815,14 @@ def cell_facet_labeling(exterior_facets,
     CHKERR(DMGetLabel(plex.dm, int_label, &label))
     CHKERR(DMLabelCreateIndex(label, fstart, fend))
 
-    sub_domain_ids = dict((sid, plex.getStratumIS(FACE_SETS_LABEL, sid))
-                          for sid in set(exterior_facets.markers))
+    exterior_markers = exterior_facets.markers
+    if exterior_markers is None:
+        # No subdomains to collect (for example, mesh is a sphere)
+        collect_subdomains = False
+    else:
+        collect_subdomains = True
+        sub_domain_ids = dict((sid, plex.getStratumIS(FACE_SETS_LABEL, sid))
+                              for sid in set(exterior_markers))
 
     for c in range(cstart, cend):
         CHKERR(PetscSectionGetOffset(cell_numbering.sec, c, &cell))
@@ -827,12 +833,15 @@ def cell_facet_labeling(exterior_facets,
                 # This is a facet point
                 DMLabelHasPoint(label, point, &is_interior)
                 cell_facets[cell, fi, 0] = <np.int8_t>is_interior
-                # If on an exterior facet, search for the subdomain marker
-                if not is_interior:
-                    for marker in sub_domain_ids:
-                        # Facet point in subdomain
-                        if point in sub_domain_ids[marker].array:
-                            cell_facets[cell, fi, 1] = <np.int8_t>marker
+
+                # Collect subdomains (if any)
+                if collect_subdomains:
+                    # If on an exterior facet, search for the subdomain marker
+                    if not is_interior:
+                        for marker in sub_domain_ids:
+                            # Facet point in subdomain
+                            if point in sub_domain_ids[marker].array:
+                                cell_facets[cell, fi, 1] = <np.int8_t>marker
                 fi += 1
 
     CHKERR(DMLabelDestroyIndex(label))
--- a/firedrake/dmplex.pyx
+++ b/firedrake/dmplex.pyx
@@ -777,8 +777,7 @@ def label_facets(PETSc.DM plex, label_boundary=True):
 
 @cython.boundscheck(False)
 @cython.wraparound(False)
-def cell_facet_labeling(exterior_facets,
-                        PETSc.DM plex,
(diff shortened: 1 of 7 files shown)


tests/slate/test_subdomains.py
```diff
@@ -0,0 +1,124 @@
...
+    x, y, z = SpatialCoordinate(mesh_extr)
+    f = project(as_vector([z, y, x]), DG)
+
+    ds_vsd = ds_v(1) + ds_v(2) + ds_v(3) + ds_v(4)
+    form = dot(f[0]*f[1]*f[2]*u, n)*ds_vsd
+
+    A = assemble(Tensor(form)).dat.data
+    ref = assemble(form).dat.data
+
+    assert np.allclose(A, ref, rtol=1e-14)
```
wence-: Can you add a test where you do something on only part of the boundary?  e.g. a hybridised problem with mixed neumann and dirichlet conditions?

firedrake/dmplex.pyx
```diff
@@ -819,7 +835,16 @@ def cell_facet_labeling(PETSc.DM plex,
...
+                cell_facets[cell, fi, 0] = <np.int8_t>is_interior
+
+                # Collect subdomains (if any)
+                if exterior_subdomains:
+                    # If on an exterior facet, search for the subdomain marker
+                    if not is_interior:
+                        for marker in exterior_markers:
+                            # Facet point in subdomain
+                            if point in sub_domain_ids[marker].array:
+                                cell_facets[cell, fi, 1] = <np.int8_t>marker
```
wence-: This is terrible.

1. you've put python calls in the inner loop
2. For every point, you search the whole array, turning this into an (at best) NlogN, but probably N^2 code path.

Just write:
```python
cell_facets[cell, fi, 0] = <np.int8_t>is_interior
if not is_interior:
    CHKERR(DMLabelGetValue(label, point, &marker)
    cell_facets[cell, fi, 1] = <np.int8_t>marker
```

wence-: Also, it is legitimate for some exterior facets to be unlabelled (DMLabelGetValue returns -1).  Similarly, I see for now you don't handle interior facet subdomains (although that is also allowed).

firedrake/slate/slac/kernel_builder.py
```diff
@@ -161,21 +168,16 @@ def _setup(self):
...
 
                 if kinfo.oriented:
                     args.insert(0, self.cell_orientations_sym)
 
-                if kinfo.integral_type in ["interior_facet",
-                                           "exterior_facet",
-                                           "interior_facet_vert",
-                                           "exterior_facet_vert"]:
+                if kint_type in ["interior_facet", "exterior_facet",
+                                 "interior_facet_vert", "exterior_facet_vert"]:
```
wence-: FWIW, it's best not to introduce unnecessary whitespace changes in functionality patches.  I was trying to understand what changed here (the answer was nothing).

firedrake/slate/slac/compiler.py
```diff
@@ -342,15 +350,25 @@ def tensor_assembly_calls(builder):
...
-        ext_calls = list(chain(*[builder.assembly_calls[it_type]
+        ext_calls = list(chain(*[assembly_calls[it_type]
                                  for it_type in ("exterior_facet",
                                                  "exterior_facet_vert")]))
 
+        # Generate logical statements for handling *exterior facet*
+        # integrals on subdomains (currently this is the only supported type)
+        if assembly_calls["subdomains_exterior_facet"]:
+            subdomain_calls = assembly_calls["subdomains_exterior_facet"]
+            for k in subdomain_calls:
```
wence-: Why not just `for k in assembly_calls["subdomains_exterior_facet"]` ?

firedrake/slate/slac/kernel_builder.py
```diff
@@ -54,7 +54,7 @@ class LocalKernelBuilder(object):
     mesh_layer_sym = ast.Symbol("layer")
 
     # Supported integral types
-    supported_integral_types = [
+    integral_keys = [
```
wence-: Again, why rename this variable in this changeset?  Especially to something that is less descriptive.

firedrake/slate/slac/compiler.py
wence-: Oh, I see, because you index later on.  I think a clearer way to write this would be to collect a list of 2-tuples of `(subdomain, integral_call)` and then do:

```python
from gem.utils import groupby

for (subdomain, calls) in groupby(assembly_calls["subdomains_exterior_facet"], lambda x: x[0]):
    ext_calls.append(ast.If(ast.Eq(ast.Symbol(..., subdomain)), (as.Block(calls, ...), )))))
```

tests/slate/test_subdomains.py
thomasgibson: Sure, I'll cook up a test problem soon.

firedrake/slate/slac/compiler.py
thomasgibson: >I think a clearer way to write this would be to collect a list of 2-tuples of `(subdomain, integral_call)` and then do:
>
>```python
>from gem.utils import groupby
>
>for (subdomain, calls) in groupby(assembly_calls["subdomains_exterior_facet"], lambda x: x[0]):
>     ext_calls.append(ast.If(ast.Eq(ast.Symbol(..., subdomain)), (as.Block(calls, ...), )))))
>```

Ah, I like this.

firedrake/dmplex.pyx
thomasgibson: >This is terrible.

Yeah I don't like it either. That's why we have peer-reviewed code though, right? Someone might know something I'm missing here.

>Just write:
>```python
>cell_facets[cell, fi, 0] = <np.int8_t>is_interior
>if not is_interior:
 >   CHKERR(DMLabelGetValue(label, point, &marker)
 >   cell_facets[cell, fi, 1] = <np.int8_t>marker
>```

Great, this is helpful. I'm assuming the `label` here is NOT the same one used to determine facet types (exterior/interior). So something like:
```python
cdef:
    ...
    PetscInt marker
    DMLabel sub_domain
    ...
CHKERR(DMGetLabel(plex.dm, FACE_SETS_LABEL.encode(), &sub_domain))
...
CHKERR(DMLabelGetValue(sub_domain, point, &marker))
cell_facets[cell, fi, 1] = <np.int8_t>marker
```
Or is there a way I can reuse the label to do this? Note that if the mesh is spherical, then `CHKERR(DMGetLabel(plex.dm, FACE_SETS_LABEL.encode(), &sub_domain))` together with `CHKERR(DMLabelGetValue(sub_domain, point, &marker))` causes a segmentation violation. This error can also be reproduced if you simply run:
```python
from firedrake import *
mesh = UnitIcosahedralSphereMesh(2)
mesh.init_cell_orientations(Expression(("x[0]", "x[1]", "x[2]")))
plex = mesh._plex
plex.getLabelIdIS("Face Sets").allGather()
```
Is this a bug/expected behaviour or did I do something wrong? If I don't special case sphere meshes, then at some point this causes tests on spheres to break due to a `segv` error. Perhaps by checking:
```python
if plex.getLabelIdIS("exterior_facets").size == 0:
    facet_set = "interior_facets"
else:
    facet_set = FACE_SETS_LABEL
CHKERR(DMGetLabel(plex.dm, facet_set.encode(), &sub_domain))
...
```
Basically, I guess I want to check if the mesh has a boundary. This was just one way I thought I could do this. Perhaps I don't need to? I'll push an update anyway and we'll go from there.
>Also, it is legitimate for some exterior facets to be unlabelled (DMLabelGetValue returns -1). Similarly, I see for now you don't handle interior facet subdomains (although that is also allowed).

OK. I see no reason why I shouldn't just handle interior facet subdomains then. So I'll make some modifications.

firedrake/dmplex.pyx
```diff
@@ -781,45 +781,60 @@ def cell_facet_labeling(PETSc.DM plex,
...
-                # This is a facet point
+                # Get interior/exterior label
                 DMLabelHasPoint(label, point, &is_interior)
-                cell_facets[cell, fi] = <np.int8_t>is_interior
+                cell_facets[cell, fi, 0] = <np.int8_t>is_interior
+
+                # Get subdomain marker
+                CHKERR(DMLabelGetValue(sub_domain, point, &marker))
+                cell_facets[cell, fi, 1] = <np.int8_t>marker
+
```
wence-: OK, I see.  You want to do this:

```python
DMLabel exterior = NULL, subdomain = NULL
CHKERR(DMGetLabel(plex.dm, "exterior_facets".encode(), &exterior))
CHKERR(DMGetLabel(plex.dm, FACE_SETS_LABEL.encode(), &subdomain))

for c in range(cstart, cend):
    CHKERR(PetscSectionGetOffset(cell_numbering.sec, c, &cell))
    fi = 0
    for p in range(nclosure):
        point = cell_closures[cell, p]
        if fstart <= point < fend:
            # I use the exterior facets label, because it is smaller.
            DMLabelHasPoint(exterior, point, &is_exterior)
            cell_facets[cell, fi, 0] = <np.int8_t>(not is_exterior)
            if subdomain != NULL:
                 CHKERR(DMLabelGetValue(subdomain, point, &marker))
                 cell_facets[cell, fi, 1] = <np.int8_t>marker
             else:
                 cell_facets[cell, fi, 1] = -1
             fi += 1
```

tests/slate/test_subdomains.py
thomasgibson: I've added a test problem which has both Dirichlet and Neumann conditions on different parts of the boundary.

firedrake/slate/slac/compiler.py
```diff
@@ -342,15 +352,29 @@ def tensor_assembly_calls(builder):
...
                                  for it_type in ("exterior_facet",
                                                  "exterior_facet_vert")]))
 
+        # Generate logical statements for handling exterior/interior facet
+        # integrals on subdomains.
+        # Currently only facet integrals are supported.
+        for sd_type in ("subdomains_exterior_facet", "subdomains_interior_facet"):
+            stmts = []
+            for sd, sd_calls in groupby(assembly_calls[sd_type], lambda x: x[0]):
+                calls = [sd_call[1] for sd_call in sd_calls]
```
wence-: AKA, `_, calls = zip(*sd_calls)`

thomasgibson:
Arg, I didn't want to dismiss the review. I just pushed just minor cosmetic changes and it automatically dismissed it.

Issue #197, opened 2014-03-20
Enforcing strong boundary conditions in petrov-galerkin methods


colinjcotter:
When computing hydrostatic pressure, the test space is different from the trial space but (if the boundary conditions are correctly set) then the system is still solvable. 

See: `tests/extrusion/test_extrusion_hydro.py` in the hydrostatic branch.

PR #1180, opened 2017-12-15
Linear solver bc change

wence-:


firedrake/linear_solver.py
```diff
@@ -101,27 +100,36 @@ def __init__(self, A, P=None, solver_parameters=None,
...
+        u.dat.zero()
+        for bc in self.A.bcs:
+            bc.apply(u)
+        update()
+        # blift contains -A u_bc
+        blift += b
+        for bc in self.A.bcs:
+            bc.apply(blift)
+        # blift is now b - A u_bc, and satisfies the boundary conditions
+        return blift
```
thomasgibson: OK, so when boundary conditions are present, you populate `u` with the `bc` data and assemble the update into the rhs. The only real difference here is that you're always applying the bcs in case they have changed.

firedrake/linear_solver.py
```diff
@@ -131,21 +139,17 @@ def solve(self, x, b):
...
-        if len(self._W) > 1 and self.near_nullspace is not None:
-            self.near_nullspace._apply(self._W.dof_dset.field_ises, near=True)
+        if len(self.trial_space) > 1 and self.nullspace is not None:
+            self.nullspace._apply(self.trial_space.dof_dset.field_ises)
+        if len(self.test_space) > 1 and self.transpose_nullspace is not None:
+            self.transpose_nullspace._apply(self.test_space.dof_dset.field_ises,
+                                            transpose=True)
+        if len(self.trial_space) > 1 and self.near_nullspace is not None:
+            self.near_nullspace._apply(self.trial_space.dof_dset.field_ises, near=True)
+
```
thomasgibson: So now that both the test and trial spaces are recorded, `solve` should do the right thing for Petrov-Galerkin discretizations with bcs, right? Would it be worthwhile to add a small test solving a simple PDE using a PG-method with bcs?

firedrake/linear_solver.py
wence-: OK, so if there are boundary conditions, then the modified right hand side is `b_orig - A u_gamma` where `u_gamma` is zero in the bulk and satisfies the boundary conditions on the boundary.

So we would do this modification before, but precompute `A u_gamma` once.

But that is unsafe, because what if the values in `u_gamma` change?  So the change here is that we compute `A u_gamma` every time we solve.

firedrake/linear_solver.py
wence-: No, at least not for general PG I think.  Think about what that means: the test and trial spaces are different.

The current interface only allows providing the bc for the row (trial) space.  For PG, you need to identify which nodes are killed in the test space given boundary conditions on the trial space.  But this is not a trivial identification (for straight Galerkin, it's easy, because the nodes are the same).

Now, I admit that I have not thought hard about what the maths (and therefore the code) looks like for this case.  

c.f. also the discussion in #197.

dham: I thought that with PG one always had a structural identity between the test and trial space, with only the basis function values being different between test and trial.

PR #1183, opened 2018-01-09
ensemble parallelism

colinjcotter:
Some tools for ensemble parallelism.

Currently only implements allreduce but other communications 
can be added in due course.


firedrake/ensemble.py
```diff
@@ -0,0 +1,63 @@
+from pyop2.mpi import MPI
+
+__all__ = ("CommManager", )
+
+
+class CommManager(object):
```
wence-: Let's bikeshed this name.  I picked `CommManager`, but I don't think it's good.  Do we want to call this `Ensemble`?  I think so.

Then, what attributes does it have?

In the normal way, it should have a `Ensemble.comm` attribute, but which communicator should this refer to.  In lunch discussion with @dham I think we concluded that it should be the "spatial communicator"?

Then, the global (input) communicator might be in the `gcomm` attribute.

Right now, we implement `allreduce`, we will also want `isend` and `irecv` probably (with the same semantics as the MPI equivalents).  These should take a source/target "ensemble rank", so this object should expose as `rank` the index of this member in the ensemble (that's `ecomm.rank` I think).

Did I miss things?

colinjcotter: It would be good to add isend and irecv.

Do we have to do them for this PR? I don't need them right now for my usecase, and don't feel too confident about writing them as I don't know the MPI lingo.

firedrake/ensemble.py
```diff
@@ -0,0 +1,65 @@
...
+        self.comm = comm.Split(color=(rank // M), key=rank)
+        """The communicator for spatial parallelism, contains a
+        contiguous chunk of M processes from :attr:`comm`"""
+
+        self.ecomm = comm.Split(color=(rank % M), key=rank)
+        """The communicator for ensemble parallelism, contains all
+        processes in :attr:`comm` which have the same rank in
+        :attr:`scomm`."""
+
+        self.erank = self.ecomm.rank
```
dham: I'm not convinced about the abbreviations here. I think as a matter of policy that readability is more important than brevity. I'd vote for `global_comm`, `ensemble_comm`. I'm also not convinced we need `self.erank`. Surely self.ensemble_comm.rank is clearer.

tests/regression/test_ensembleparallelism.py
```diff
@@ -0,0 +1,62 @@
+from firedrake import *
+import pytest
+
+
+@pytest.mark.parallel(nprocs=6)
+def test_ensemble_allreduce():
+    manager = CommManager(COMM_WORLD, 2)
```
dham: Tests need updating for name changes in the code.

jshipton:
I've fixed this but I don't think I have permissions to push - can someone sort this out for me please?

dham:
@jshipton, I've added you to @firedrakeproject/firedrake-contributors so you should now be able to push to all @firedrakeproject repositories.

colinjcotter:
Is it ok to merge this now?

PR #1184, opened 2018-01-11
No need for fortran bindings for PETSc, fixes #1099

### LINKED ISSUE #1099: install script ignores petsc configuration
hi,
i want to skip downloading eigen during the firedrake install; to this end i use

`curl -O https://raw.githubusercontent.com/firedrakeproject/firedrake/master/scripts/firedrake-install`
`PETSC_CONFIGURE_OPTIONS="--download-eigen=0" python firedrake-install`

as suggested before. however, i still see `Downloading Eigen from 'https://bitbucket.org/eigen/eigen/get/3.3.3.tar.gz' to 'eigen-3.3.3.tgz'` in the terminal, which indicates that this configuration didn't work

here is the install log:
[firedrake-install.txt](https://github.com/firedrakeproject/firedrake/files/1137800/firedrake-install.txt)

thanks for the help

wence-:
Although we need a fortran *compiler*, to install mumps and similar, we don't actually need PETSc to builds its fortran bindings (which takes lost of memory, because they're massive files).

### DIFF
--- a/scripts/firedrake-install
+++ b/scripts/firedrake-install
@@ -321,6 +321,8 @@ If you really want to use your own Python packages, please run again with the
 
 petsc_opts = "--download-eigen=%s/src/eigen-3.3.3.tgz " % firedrake_env
 
+petsc_opts += "--with-fortran-bindings=0 "
+
 if options["minimal_petsc"]:
     if options["petsc_int_type"] == "int64":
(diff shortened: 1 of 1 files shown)


wence-:
They did, and as a happy side effect reduced the jenkins petsc build time by around 3 minutes.

PR #1186, opened 2018-01-18
tsfc to loopy

tj-sun:
I think this is finally ready for review.

This should land together with https://github.com/firedrakeproject/tsfc/pull/159 and https://github.com/OP2/PyOP2/pull/547.

~~This PR is NOT ready to land yet, but I'm creating it to keep track of progress. There are several problems to fix, some of which require improving loopy. However, I do plan to explore vectorization with this while working through these problems in the meantime.~~

**TODO**
- [x] loopy kernel -> C is slow for kernels with large arrays.

[link](https://gitlab.tiker.net/inducer/loopy/issues/118)
~~It takes 36 seconds to generate 35 MB of C code, which doesn't have many instructions, but has some large const arrays.~~
This is solved by https://gitlab.tiker.net/inducer/loopy/merge_requests/210, plot is updated.

- [x] loopy kernel -> C is slow due to instruction scheduling.

~~ It takes 102 seconds to generate holzapfel kernel currently. I have observed this to be ~10 seconds before, but that is a mistake due to either forgetting to turn off loopy cache, or something major changed since Christmas (unlikely). ~~

This is solved by https://gitlab.tiker.net/inducer/loopy/merge_requests/229

- [x] need to be able to declare an global constant array to be static.

[link](https://gitlab.tiker.net/inducer/loopy/issues/117)
It's a bit tricky to generate `static` keyword in loopy without messing up code generation for opencl. So currently this branch generates const arrays outside the kernel:
```C
const double t = {...}
void kernel(...)
{
  ...
}
```
instead of this:
```C
void kernel(...)
{
  static const double t = {...}
  ...
}
```
~~I think this is also what we want for GPU, but we need to declare the global constants to be static for C, so that they are not visible outside (not sure in what cases this problem hurts us in the context of firedrake).~~
This is sorted by [211](https://gitlab.tiker.net/inducer/loopy/merge_requests/211).

- [x] quirky gcc error

Fixed by https://github.com/OP2/PyOP2/commit/5b73d76750732440ec651a17ca768fa55358c097
Bug report filed https://gcc.gnu.org/bugzilla/show_bug.cgi?id=84477

This shows up in the test `extrusion/test_steady_advection_3D_extr.py::test_near_to_far[True]`

Consider [code1](https://gist.github.com/tj-sun/45eb2ed176112e9dac4e81eac24b7312#file-code1) and [code2](https://gist.github.com/tj-sun/45eb2ed176112e9dac4e81eac24b7312#file-code2), the only difference is the exchange of loop order in line 330 and 332, and both are valid code. But gcc 5.4 cannot compile code1, with error message "internal compiler error: in compute_live_loop_exits, at tree-ssa-loop-manip.c:255". 

This issue can be reproduced with the simplified code [simple](https://gist.github.com/tj-sun/45eb2ed176112e9dac4e81eac24b7312#file-simple). Any of the following will fix the error: 
- compile without O3
- compile without -ffast-math
- initialize some of the array
- remove one of the loop after the loop in question
- use gcc 7 or icc (haven't tried other version of gcc)

- [x] install with loopy

~~Simply adding loopy to requirement-git.txt in tsfc and pyop2 doesn't seem to do the job. [link](https://jenkins.ese.ic.ac.uk:1080/blue/organizations/jenkins/firedrakeproject%2Ffiredrake/detail/tsfc2loopy/6/pipeline/?start=0#step-28-log-0)~~
Sorted by d027d2bc93120aa06e962a6291cc9460f88831fa

- [x] ~~need to be able to enforce alignments in array declaration, e.g `__attribute__((aligned(64)))`.~~

Fixed by [link](https://gitlab.tiker.net/inducer/loopy/merge_requests/215)

**Performance**

Assembly of the following bilinear forms: mass, helmholtz, elasticity, hyperelasticity, poisson, mixed poisson.
Test and trial functions on CG function space of degree `p` (= 1, 2, 3, 4), inner product with `n` (= 0, 1, 2, 3) functions on CG function space of degree `q` (= 1, 2, 3, 4), on UnitSquareMesh(31, 31) and UnitCubeMesh(9, 9, 9).

Kernel compile time (form -> C) vs size of C file:
compile_plot

The top right ones are the mixed poisson forms (which is the one mentioned in the first todo point). Also note holzapfel is a lot worse as mentioned.

Kernel compile time vs lines of code (for loopy kernels, i.e. "new", since COFFEE generates a lot more lines for array declaration)
compile_plot_loc

Run time for different forms (5 times average, with warming up)
run_plot
Not much differences here.

- [ ] port patch preconditioner to code generation with loopy


inducer:
> Simply adding loopy to requirement-git.txt in tsfc and pyop2 doesn't seem to do the job.

Looks like it slurped in PyOpenCL, I assume by accident. Does your build logic look at loopy's `requirements.txt`? Loopy does not have a dependency on pyopencl, although it *can* obviously work with it. (It doesn't declare a dep in `setup.py`, but *does* include it in `requirements.txt`. If that's an issue, that can probably be changed.

wence-:
>> Simply adding loopy to requirement-git.txt in tsfc and pyop2 doesn't seem to do the job.

> Looks like it slurped in PyOpenCL, I assume by accident. Does your build logic look at loopy's requirements.txt? Loopy does not have a dependency on pyopencl, although it can obviously work with it. It doesn't declare a dep in setup.py, but does include it in requirements.txt. If that's an issue, that can probably be changed.

Yeah, for packages we clone we check for `requirements-{ext/git}.txt` then `requirements.txt`, install those manually and then `pip install` (which would satisfy further dependencies specified in setup.py).

firedrake/assemble_expressions.py
```diff
@@ -591,25 +522,126 @@ def assemble_expression(expr, subset=None):
...
+    elif isinstance(expr, ufl.classes.NE):
+        op = "!="
+    elif isinstance(expr, ufl.classes.LT):
+        op = "<"
+    elif isinstance(expr, ufl.classes.LE):
+        op = "<="
+    elif isinstance(expr, ufl.classes.GT):
+        op = ">"
+    elif isinstance(expr, ufl.classes.GE):
+        op = ">="
```
miklos1: Shorter version: `op = expr._name`

tj-sun: Done

tj-sun:
I think this is ready for a new round of review. Some updates:

- use loopy backend for all wrappers and (almost) all kernels, e.g. expression, interpolation.
- `slate` and `point evaluation` still generates coffee kernel, but the wrapper is generated as a loopy kernel, so effectively there is no more string pasting in `sequential.py`.
- rewrite utility kernels (e.g. prolongation etc in multigrid) as loopy kernel.
- depreciate `op2.IterationSpace`

This is not ready to go in until `kernel callable` is merged into loopy. However, registration of kernel with wrapper happens in pyop2, so I expect no change here (and in tsfc) when that happens.

TODO
- [ ] update documentation
- [ ] hashing of loopy kernel (with registered callables) is not perfect currently in loopy (thus the warning of cache key collision), will need to ensure that's fixed in `kernel callable`.

firedrake/mesh.py
```diff
@@ -1086,24 +1090,20 @@ def spatial_index(self):
...
-        kernel = kernel.replace("nodes_per_cell", str(nodes_per_cell))
-
-        par_loop(kernel, ufl.dx, {'f': (self.coordinates, READ),
-                                  'f_min': (coords_min, RW),
-                                  'f_max': (coords_max, RW)})
+        domain = "{{[d, i]: 0 <= d < {0} and 0 <= i < {1}}}".format(gdim, nodes_per_cell)
+        instructions = """
+        for d, i
+            f_min[0, d] = fmin(f_min[0, d], f[i, d])
+            f_max[0, d] = fmax(f_max[0, d], f[i, d])
```
wence-: Why do we need this indexing `[0, d]` for `f_min`. And also, does loopy not spell `min/max` like that (without the `f`?)

firedrake/parloops.py
```diff
@@ -74,38 +72,50 @@ def _form_kernel(kernel, measure, args, **kwargs):
...
+def par_loop(kernel_domains, instructions, measure, args, kernel_kwargs=None, **kwargs):
     r"""A :func:`par_loop` is a user-defined operation which reads and
     writes :class:`.Function`\s by looping over the mesh cells or facets
     and accessing the degrees of freedom on adjacent entities.
 
-    :arg kernel: is a string containing the C code to be executed.
+    :arg kernel_domains: the iteration domain for the kernel, in Integer Set
+        Libarary (ISL) syntax. function_name.dofs can be used as a magic word.
+    :arg instructions: is a string containing the instruction for the kernel
+        in loo.py syntax.
```
wence-: Can you add some links here?

firedrake/parloops.py
```diff
@@ -74,38 +72,50 @@ def _form_kernel(kernel, measure, args, **kwargs):
...
-                kargs.append(ast.Decl("double *", ast.Symbol(var, (ndof,))))
-        lkernel = lkernel.replace(var+".dofs", str(ndof))
+            # FIXME: shape for facets [2][ndof]?
+            kargs.append(loopy.GlobalArg(var, dtype=dtype, shape=(ndof, cdim)))
+        kernel_domains = kernel_domains.replace(var+".dofs", str(ndof))
 
-    body = ast.FlatBlock(lkernel)
+    if kernel_domains is "":
+        kernel_domains = "[] -> {[]}"
+    kargs.append(...)
```
wence-: Why do we need this? I would have thought we explicitly state all kernel arguments. Or is this so that loopy can infer types of local variables?

firedrake/parloops.py
```diff
@@ -74,38 +72,50 @@ def _form_kernel(kernel, measure, args, **kwargs):
...
+                    cdim = func.dat.cdim
+                    dtype = func.dat.dtype
             if measure.integral_type() == 'interior_facet':
                 ndof *= 2
-            if measure is direct:
-                kargs.append(ast.Decl("double", ast.Symbol(var, (ndof,))))
-            else:
-                kargs.append(ast.Decl("double *", ast.Symbol(var, (ndof,))))
-        lkernel = lkernel.replace(var+".dofs", str(ndof))
+            # FIXME: shape for facets [2][ndof]?
```
wence-: We would need to reshape (or change the way we do indexing, I think).

firedrake/pointeval_utils.py
```diff
@@ -109,26 +109,42 @@ def predicate(index):
...
     }
+    int layers[2] = {0, 0};
+    if (f->extruded != 0) {
+        int nlayers = f->n_layers;
+        layers[1] = cell %% nlayers + 2;
+        cell = cell / nlayers;
+    }
 
-    wrap_evaluate(result, reference_coords.X, f->coords, f->coords_map, f->f, f->f_map%(nlayers)s, cell);
+    wrap_evaluate(result, reference_coords.X, cell, cell+1%(layers)s, f->coords, f->f, %(map_args)s);
```
wence-: This kind of code-pasting continues to be ugly. I think we should seriously consider (not for this PR) how we are going to make these parts maintainable going forward. In particular, to me it makes sense that we write some of these things as loopy kernels (possibly opaquely) and then use the new function-composition to glue them together.

firedrake/pointquery_utils.py
```diff
@@ -197,15 +202,25 @@ def compile_coordinate_element(ufl_coordinate_element, contains_eps, parameters=
...
     return return_value;
 }
+
+int to_reference_coords_xtr(void *result_, struct Function *f, int cell, int layer, double *x)
+{
+    int return_value;
+    int layers[2] = {0, layer+2};  // +2 because the layer loop goes to layers[1]-1, which is nlayers-1
+    %(non_extr_comment_out)swrap_to_reference_coords(result_, x, &return_value, cell, cell+1, layers, f->coords, f->coords_map);
+    return return_value;
+}
```
wence-: Can you explain why you had to make a separate function like this for extruded vs. not. I notice that we didn't have this beforehand.

firedrake/slate/preconditioners.py
```diff
@@ -98,19 +98,21 @@ def initialize(self, pc):
...
-        for (int i=0; i<%d; ++i) {
-        for (int j=0; j<%d; ++j) {
-        vec_out[i][j] += vec_in[i][j]/w[i][j];
-        }}""" % shapes
+        par_loop(domain, instructions, ufl.dx, {"w": (self.weight, INC)})
+
+        instructions = """
+        for i, j
+            vec_out[i,j] = vec_out[i,j] + vec_in[i,j]/w[i,j]
+        end
```
wence-: It's not super important for these kernels, but I guess it would be good to ensure the loop priority is somehow right. Does loopy's scheduler prioritise "stride-1" over something else? @inducer ?

firedrake/slate/slac/compiler.py
```diff
@@ -156,7 +156,8 @@ def generate_kernel_ast(builder, statements, declared_temps):
     result_sym = ast.Symbol("T%d" % len(declared_temps))
     result_data_sym = ast.Symbol("A%d" % len(declared_temps))
     result_type = "Eigen::Map<%s >" % eigen_matrixbase_type(shape)
-    result = ast.Decl(SCALAR_TYPE, ast.Symbol(result_data_sym, shape))
+    # result = ast.Decl(SCALAR_TYPE, ast.Symbol(result_data_sym, shape))
```
wence-: Please don't leave commented code like this lying around.

firedrake/slate/slac/compiler.py
```diff
@@ -213,7 +214,7 @@ def generate_kernel_ast(builder, statements, declared_temps):
                              qualifiers=["const"]))
 
     # Macro kernel
-    macro_kernel_name = "compile_slate"
+    macro_kernel_name = "pyop2_kernel_compile_slate"
```
wence-: Is it still the case that all of these kernels need to be written with "pyop2_kernel_" for loopy to recognise them?

firedrake/slope_limiter/vertex_based_limiter.py
```diff
@@ -37,26 +37,30 @@ def __init__(self, space):
...
-for(int i = 0; i < maxq.dofs; i++) {
-    maxq[i][0] = fmax(maxq[i][0],q[0][0]);
-    minq[i][0] = fmin(minq[i][0],q[0][0]);
-}
-                             """
+        domain = "{[i]: 0 <= i < maxq.dofs}"
+        instructions = """
+        for i
+            maxq[i, 0] = fmax(maxq[i, 0], q[0, 0])
+            minq[i, 0] = fmin(minq[i, 0], q[0, 0])
```
wence-: The double indexing was an artifact of old PyOP2 codegen, right?

firedrake/utility_meshes.py
```diff
@@ -1360,25 +1336,27 @@ def PartiallyPeriodicRectangleMesh(nx, ny, Lx, Ly, direction="x", quadrilateral=
...
-            new_coords[i][1] = old_coords[i][2]*Ly[0];
-            }"""
+
+    domain = "{[i, j]: 0 <= i < old_coords.dofs and 0 <= j < new_coords.dofs}"
+    instructions = """
+    <float64> Y = 0
+    <float64> pi = 3.141592653589793
+    for i
+        Y = Y + old_coords[i, 1]
+    end
```
wence-: Can we write this as the reduction it is? `<float64> Y = sum(i, old_coords[i, 1])`?

wence-: (Minor issue/comment)

requirements-ext.txt
```diff
@@ -1,8 +1,8 @@
 sympy
 cachetools
-pytest
-pytest-xdist
-pytest-benchmark
+pytest==3.9.3
+pytest-xdist==1.24.1
+pytest-benchmark==3.1.1
```
wence-: I think I am fixing these.

tests/extrusion/test_rhs_side_bcs.py
```diff
@@ -21,7 +21,7 @@ def run_test(x, degree, quadrilateral, parameters={}, test_mode=False):
     for bc in bcs:
         bc.apply(u)
     v = Function(V)
-    v.interpolate(Expression("x[0] < 0.05 ? 10.0 : x[0] > 0.95 ? 42.0 : 0.0"))
+    v.interpolate(Expression("if(x[0] < 0.05, 10, if(x[0] > 0.95, 42, 0))"))
```
wence-: OK, so `Expression` is now not C string, but loopy. This needs to be documented, or else we need to figure out a way to not do that.

tests/extrusion/test_steady_advection_3D_extr.py
```diff
@@ -79,7 +79,8 @@ def test_right_to_left(mesh, DG0, W):
     velocity = Expression(("-1.0", "0.0", "0.0"))
     u0 = project(velocity, W)
 
-    inflowexpr = Expression("x[1] > 0.25 && x[1] < 0.75 ? 1.0 : 0.5")
+    # inflowexpr = Expression("x[1] > 0.25 && x[1] < 0.75 ? 1.0 : 0.5")
```
wence-: Please don't leave old code in comments.

firedrake/parloops.py
wence-: Thinking about it some more, I think that we ought to be able to deduce the common domains automagically, and should provide the option to say what they are.

How about an interface that is something like:

```
# Example nonsense instructions
instructions = """
for i, j, k
    A[i, j] = A[i, j] + A[i, j] * B[i, k]
end
"""
par_loop(instructions, dx, {"A": (a, RW, ("i", "j")), "B": (b, R, ("i", "k")})
```

And we can deduce the index extents from the `dim` of the `dat` in each `Function`, right?

if we optionally want to override the domains, we can say:

```
par_loop(instructions, dx, {...}, domains="{ [i, j, k] : ...}")
```

Achievable?

firedrake/slate/preconditioners.py
inducer: There is no built-in preference. Loop order is user-specified.

tests/extrusion/test_rhs_side_bcs.py
inducer: If feasible, I'd recommend not using strings to pass information to loopy. Instead, simply build the equivalent expression in pymbolic.

firedrake/parloops.py
inducer: > instructions: is a string containing the instructions

* If feasible, I'd recommend not using strings to pass information to loopy.
* Terminology-wise, I am still thinking of moving to "statement" over "instruction". If it happens, it'll be in a backward-compatible manner, but you may still want to write `statements` rather than `instruction` in your code.

tests/extrusion/test_rhs_side_bcs.py
wence-: This is "user-facing" code, rather than auto-generated, so I would prefer not to have users writing pymbolic expressions if possible. Previously it was for pasting raw C strings into kernels.

firedrake/parloops.py
wence-: Here again, this is meant to be "user-facing". I think it is friendlier for them to write the "string" expression than the equivalent pymbolic, but you could convince me otherwise if it's really easy.

inducer:
> This is "user-facing" code, rather than auto-generated, so I would prefer not to have users writing pymbolic expressions if possible. Previously it was for pasting raw C strings into kernels. 

OIC

inducer:
> Here again, this is meant to be "user-facing". I think it is
> friendlier for them to write the "string" expression than the
> equivalent pymbolic, but you could convince me otherwise if it's
> really easy.

I had not realized this was user-facing. I agree strings are probably
friendlier, though I would give some consideration to allowing IR in
addition to strings, much like loopy itself.

tests/extrusion/test_rhs_side_bcs.py
dham: We already have a deprecation policy on C strings. It's in the manual: https://www.firedrakeproject.org/interpolation.html#c-string-expressions

We only keep C string expressions for FEniCS compatibility (and if they are going to contain something that isn't actually C then we're not compatible and we should just drop them).

I am not aware of any live use cases for C string expressions that can't be done in UFL. If they do exist, then we should extend UFL to fix them. In extremis the user can fall back to a PyOP2 kernel.

wence-: OK, so my suggestion, if we can't keep C string compatibility:

1. replace `Expression` with a function that raises an error with "how to use UFL expressions"
2. expunge `Expression` from everywhere in Firedrake.

miklos1:
Dropping support for C string expressions drops a FEniCS compatibility feature, however, if C expressions cannot be supported with loopy, it might be the right thing to do...

dham:
Lawrence and I came to the same conclusion on slack. I'm going to start the process now.

inducer:
A C pass-through-expression type thing *can* be done, even if I'm not the world's biggest fan of the idea. Alternatively, we could try to parse them and convert them to more regular IR, although I don't think that'll ever be robustly compatible with the pre-existing functionality. My main question would be how those C expressions are used, and if that use case justifies the amount of effort required.

dham:
We're going to drop the C expression functionality. It's a misfeature which existed only for FEniCS compatibility. As far as I am aware it has no legitimate uses. Writing UFL expressions is just better in every respect.

scripts/firedrake-install
```diff
@@ -437,13 +437,19 @@ def git_clone(url):
     try:
         if options["disable_ssh"]:
             raise InstallError("Skipping ssh clone because --disable-ssh")
-        check_call(["git", "clone", "-q", git_url(plain_url, "ssh")])
+        if name == "loopy":
+            check_call(["git", "clone", "-q", "--recursive", git_url(plain_url, "ssh")])
+        else:
+            check_call(["git", "clone", "-q", git_url(plain_url, "ssh")])
```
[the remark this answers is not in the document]
tj-sun: Agreed and amended. Checked that petsc and petsc4py do not have submodules.

scripts/firedrake-install
```diff
@@ -1064,6 +1073,12 @@ if mode == "install":
     build_and_install_libspatialindex()
     pipinstall.append("--no-deps")
     for p in packages:
+        if p == "loopy":
+            # install loopy dependencies in setup.py
+            pipinstall.pop()
+            install(p + "/")
+            pipinstall.append("--no-deps")
+            continue
```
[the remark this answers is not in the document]
tj-sun: Yes!

tests/regression/test_periodic_interval_advection.py
```diff
@@ -22,7 +22,7 @@ def run_test(degree):
     l2error = []
     # Advect a sine wave with a constant, unit velocity for 200
     # timesteps (dt = 5e-5)
-    for n in range(5, 11):
+    for n in range(5, 10):
```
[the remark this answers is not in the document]
tj-sun: There are floating point precision issues for the last `n`:

> regression/test_periodic_interval_advection.py::test_periodic_1d_advection[DG3] FAILED        [100%]

>  assert np.all(convergence > threshold)
> E       assert False
> E        +  where False = <function all at 0x7fba7b8f6158>(array([3.97610266, 4.14458185, 4.107893  , 4.06709071, 2.45697382]) > 3.9)
> E        +    where <function all at 0x7fba7b8f6158> = np.all

> print(l2error)
> [4.09884710e-07 2.60456711e-08 1.47262475e-09 8.54069248e-11
 5.09538244e-12 9.28012800e-13]

firedrake/mesh.py
tj-sun: I think loopy treats `max` and `min` as reductions rather than converting them to function calls.

firedrake/utility_meshes.py
tj-sun: Also, I was thinking the reduction in the tsfc generated assembly kernels could be written in similar ways. Maybe one step at a time?

requirements-ext.txt
tj-sun: Yes, tested. Thanks!

tests/extrusion/test_steady_advection_3D_extr.py
tj-sun: Oops.

firedrake/mesh.py
inducer: [Relevant discussion](https://github.com/inducer/loopy/issues/73). tl;dr: This will be (but currently isn't) `minimum(x,y)`, to be consistent with numpy naming.

tests/multigrid/test_embedded_transfer.py
```diff
@@ -105,6 +105,7 @@ def solver(V, space, solver_parameters):
     return solver
 
 
+@pytest.mark.skip("preconditioners/patch.py are not updated to loopy codegen yet")  # FIXME: fix this before merging
```
dham: How are we going to fix this?

firedrake/assemble.py
```diff
@@ -330,8 +330,8 @@ def _assemble(f, tensor=None, bcs=None, form_compiler_parameters=None,
         def mat(testmap, trialmap, i, j):
             m = testmap(test.function_space()[i])
             n = trialmap(trial.function_space()[j])
-            maps = (m[op2.i[0]] if m else None,
-                    n[op2.i[1 if m else 0]] if n else None)
+            maps = (m if m else None,
```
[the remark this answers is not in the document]
wence-: indexing was a way of saying "pass data, not pointers" in kernel arguments. Now all arguments are passed as data, so there is no need for this.

firedrake/assemble_expressions.py
```diff
@@ -506,6 +435,10 @@ def operator(self, o):
         return o._ufl_expr_reconstruct_(*operands)
 
 
+class Bag():
```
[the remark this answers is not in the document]
wence-: An object on which one can set arbitrary properties. AKA a PropertyDict. Aside, should subclass `object`.

firedrake/extrusion_utils.py
```diff
@@ -47,116 +47,113 @@ def make_extruded_coords(extruded_topology, base_coords, ext_coords,
...
-                    ext_coords[2*d][c] = base_coords[d][c];
-                    ext_coords[2*d+1][c] = base_coords[d][c];
-                }
-                ext_coords[2*d][%(base_coord_dim)d] = *layer_height * (layer);
-                ext_coords[2*d+1][%(base_coord_dim)d] = *layer_height * (layer + 1);
-            }
-        }""" % {'base_map_arity': base_coords.cell_node_map().arity,
-                'base_coord_dim': base_coords.function_space().value_size},
-            "uniform_extrusion_kernel")
+inline void pyop2_kernel_uniform_extrusion(double *ext_coords,
```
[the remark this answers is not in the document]
wence-: They can, but need not. I think we should rewrite these to be loopy kernels though.

firedrake/interpolation.py
```diff
@@ -263,64 +263,72 @@ def compile_c_kernel(expression, to_pts, to_element, fs, coords):
 
     names = {v[0] for v in expression._user_args}
```
wence-: With `Expression` dead, is this is a code path we can now ever go down?

firedrake/assemble_expressions.py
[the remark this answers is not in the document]
dham: Subclassing `object` is optional in Python 3 and has no effect.

firedrake/mesh.py
```diff
@@ -84,7 +84,16 @@ def __init__(self, mesh, classes, kind, facet_cell, local_facet_number, markers=
 
         self.facet_cell = facet_cell
 
-        self.local_facet_number = local_facet_number
+        if isinstance(self.set, op2.ExtrudedSet):
+            dset = op2.DataSet(self.set.parent, self._rank)
+        else:
+            dset = op2.DataSet(self.set, self._rank)
```
wence-: This is because `datasets` can't be made on extruded sets, right?

firedrake/mesh.py
```diff
@@ -1151,7 +1151,7 @@ def _c_locator(self, tolerance=None):
     int locator(struct Function *f, double *x)
     {
         struct ReferenceCoords reference_coords;
-        return locate_cell(f, x, %(geometric_dimension)d, &to_reference_coords, &reference_coords);
+        return locate_cell(f, x, %(geometric_dimension)d, &to_reference_coords, &to_reference_coords_xtr, &reference_coords);
```
wence-: It remains unclear to me why one needs a separate function for mapping to reference coordinates on extruded meshes (especially given that it was not needed beforehand).

firedrake/mg/kernels.py
```diff
@@ -334,7 +335,7 @@ def inject_kernel(Vf, Vc):
         %(to_reference)s
         %(evaluate)s
 
-        void inject_kernel(double *R, const double *X, const double *f, const double *Xf)
+        void pyop2_kernel_inject(double *R, const double *X, const double *f, const double *Xf)
```
wence-: Is it still the case that all of these functions need to start with `pyop2_kernel_`?

firedrake/pointquery_utils.py
```diff
@@ -197,15 +202,25 @@ def compile_coordinate_element(ufl_coordinate_element, contains_eps, parameters=
...
-    wrap_to_reference_coords(result_, x, &return_value, f->coords, f->coords_map%(nlayers)s, cell);
+    %(extr_comment_out)swrap_to_reference_coords(result_, x, &return_value, cell, cell+1, f->coords, f->coords_map);
     return return_value;
 }
+
+int to_reference_coords_xtr(void *result_, struct Function *f, int cell, int layer, double *x)
+{
+    int return_value;
+    int layers[2] = {0, layer+2};  // +2 because the layer loop goes to layers[1]-1, which is nlayers-1
+    %(non_extr_comment_out)swrap_to_reference_coords(result_, x, &return_value, cell, cell+1, layers, f->coords, f->coords_map);
```
wence-: This is disgustingly hacky.

tests/multigrid/test_embedded_transfer.py
wence-: I am going to do it.

wence-:
Let's carry on in #1408, where I think I have a merge candidate that passes all tests on Jenkins.

PR #1188, opened 2018-01-24
Pefarrell/custom prolongation

florianwechsung:
This is a suggestion for how one could add the capability that the user can provide their own prolongation, restriction and injection operators. Happy to change this.


wence-:
I was thinking a little bit about this on the train.  I wonder if this is the right place to stash this information.  Short term it's probably OK, but really this configuration is problem (i.e. variational solve) specific, rather than function space specific.

So should this instead somehow live on the `NonlinearVariationalProblem`? which is accessible to the `create_restriction/...` routines through the DM's application context.

What do you think?

I'm worrying about the case where you want to solve two different problems (with different transfer operators) but defined on the same function space.

wence-:
Doing this slightly differently with a context manager.

PR #1192, opened 2018-02-16
Add ability to configure overlap for parallel meshes

wence-:
This is necessary for some schwarz-like solvers in parallel, where the
solver needs more mesh overlap than is necessary for building the
global problem.

@colinjcotter, @pefarrell, @florianwechsung 

You need something like this to run any of the patch smoother code in parallel.  Otherwise on the overlaps you'll be inverting the wrong operator.


wence-:
Added a test, and fixed a stupid bug.

PR #1195, opened 2018-02-19
Fix trace conditions on periodic meshes and revise nullspace handling for hybridization

thomasgibson:
This fixes an issue when using the `HybridizationPC` on an extruded periodic mesh. When the trace conditions are applied, vanishing conditions on the trace space get applied to an empty domain marker (`DirichletBC(trace_space, 0, ())`, which is just wrong). Without modifying the trace condition logic too much, I believe I corrected this problem.

Another change here drops the automating of the nullspace computation in the hybridization PC. Instead, the user provides a nullspace basis for the trace space via the `appctx` on solver objects. I like the idea of automating the nullspace handling, but I am 100% sure the current approach doesn't work in general. A minimal example would be to hybridize the Poisson equation on an extruded periodic interval mesh (mixed with homogenous strong flux conditions on top/bottom parts). The current approach doesn't give the right basis. It strangely works for spherical problems though. I'm open to suggestions.


firedrake/slate/preconditioners.py
```diff
@@ -211,12 +212,9 @@ def initialize(self, pc):
         self.S.force_evaluation()
         Smat = self.S.petscmat
 
-        # Nullspace for the multiplier problem
-        nullspace = create_schur_nullspace(P, -K * Atilde,
-                                           V, V_d, TraceSpace,
-                                           pc.comm)
+        nullspace = self.app_cxt.get("hybridization_trace_nullspace", None)
```
wence-: So the only thing I don't like about this is that the user doesn't get access to the trace function space outside (they never see it).

Why not instead allow the user to provide a callback for making the null space:

```
def nullspace(V):
   return ...

appctx["nullspace"] = ...
```

And then you do:

```
nullspace = self.app_cxt.get("...", None)
if nullspace is not None:
    nullspace = nullspace(TraceSpace)
```

firedrake/slate/preconditioners.py
```diff
@@ -170,9 +171,9 @@ def initialize(self, pc):
             if "on_boundary" in neumann_subdomains:
                 measures.append(ds)
             else:
-                measures.append(ds(tuple(neumann_subdomains)))
+                measures.extend([ds(sd) for sd in neumann_subdomains])
                 dirichlet_subdomains = set(mesh.exterior_facets.unique_markers) - neumann_subdomains
-                trace_subdomains.append(sorted(dirichlet_subdomains))
+                trace_subdomains.extend([sd for sd in dirichlet_subdomains])
```
wence-: You lost the sorting, I think that is important.  So just write: `trace_subdomains.extend(sorted(dirichlet_subdomains))`

firedrake/slate/preconditioners.py
```diff
@@ -45,6 +45,7 @@ def initialize(self, pc):
         prefix = pc.getOptionsPrefix() + "hybridization_"
         _, P = pc.getOperators()
         self.cxt = P.getPythonContext()
+        self.app_cxt = self.cxt.appctx
```
wence-: I know you like `cxt` over `ctx`, but this feels somewhat wilful :).

In fact, why does this need to be a member variable at all, you use it once, in this function, so just write `self.cxt.appctx` at that point, no?

thomasgibson:
For some reason, the GMG demo failed on Jenkins. I just forced a rebuild. Hopefully, it's just a fluke.

PR #1207, opened 2018-03-27
Modify logic for binary operations on Slate tensors

thomasgibson:
I ran into this problem while working on some HDG stuff. There are instances when using `tensor` blocking and multiplying against an `AssembledVector`, for example, causes a `ValueError` because the contracted spaces are not the "same."

This occurs because calling `block` on a tensor reconstructs function spaces such that they agree with the argument indices of the underlying UFL form, or other Slate objects. (Similar to what is done when you call the form splitter). Blocking does not carry over the information that a particular function space is an indexed subspace of a mixed space. Therefore, I run into instances where I hit my own error-checking when performing a Slate mat-vec, even though it mathematically makes sense.

This reproduces the error:
```python
from firedrake import *


mesh = UnitIcosahedralSphereMesh(3)

V = VectorFunctionSpace(mesh, "DG", 1)
U = FunctionSpace(mesh, "DG", 1)
T = FunctionSpace(mesh, "DGT", 1)
W = V * U * T

w, psi, gammar = TestFunctions(W)
u, phi, lambdar = TrialFunctions(W)

a = inner(w, u)*dx + inner(psi, phi)*dx + gammar('+')*lambdar('+')*dS

A = Tensor(a)

w = Function(W)
_, _, lambda_hyb = w.split()

A02 = A.block((0, 2))

# Slate coefficient vector for lambda.
# Note: same function space as the trial space for A02 (but is indexed)
Lambda = AssembledVector(lambda_hyb)
foo = A02 * Lambda
```

I'm not sure how to correct this without modifying the current logic for binary operations on Slate expressions. It wasn't invasive and this change fixes the issue. I'm open to suggestions.

### DIFF
--- a/firedrake/slate/slate.py
+++ b/firedrake/slate/slate.py
@@ -757,7 +757,8 @@ class Add(BinaryOp):
             raise ValueError("Illegal op on a %s-tensor with a %s-tensor."
                              % (A.shape, B.shape))
 
-        assert A.arg_function_spaces() == B.arg_function_spaces(), (
+        assert all([space_equivalence(fsA, fsB) for fsA, fsB in
+                    zip(A.arg_function_spaces(), B.arg_function_spaces())]), (
             "Function spaces associated with operands must match."
         )
 
@@ -796,7 +797,10 @@ class Mul(BinaryOp):
             raise ValueError("Illegal op on a %s-tensor with a %s-tensor."
                              % (A.shape, B.shape))
 
-        assert A.arg_function_spaces()[-1] == B.arg_function_spaces()[0], (
+        fsA = A.arg_function_spaces()[-1]
+        fsB = B.arg_function_spaces()[0]
+
+        assert space_equivalence(fsA, fsB), (
             "Cannot perform argument contraction over middle indices. "
             "They must be in the same function space."
         )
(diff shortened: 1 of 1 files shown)


firedrake/slate/slate.py
```diff
@@ -821,6 +825,24 @@ def arguments(self):
...
+    :arg A: A function space.
+    :arg B: Another function space.
+
+    Returns `True` if they have matching meshes, elements, and rank. Otherwise,
+    `False` is returned.
+    """
+
+    equiv = False
+    assert A.mesh() == B.mesh(), "Mismatching meshes!"
+    if A.rank == B.rank and A.ufl_element() == B.ufl_element():
```
wence-: Under what circumstances do the elements match, but the rank not?

firedrake/slate/slate.py
```diff
@@ -821,6 +825,24 @@ def arguments(self):
...
+
+    :arg A: A function space.
+    :arg B: Another function space.
+
+    Returns `True` if they have matching meshes, elements, and rank. Otherwise,
+    `False` is returned.
+    """
+
+    equiv = False
+    assert A.mesh() == B.mesh(), "Mismatching meshes!"
```
wence-: I would allow checking on mismatching meshes, but say false.

firedrake/slate/slate.py
```diff
@@ -821,6 +825,24 @@ def arguments(self):
...
+    if A.rank == B.rank and A.ufl_element() == B.ufl_element():
+        equiv = True
+
+    return equiv
```
wence-: You can do beta-reduction:

`return A.mesh() == B.mesh() and A.ufl_element() == B.ufl_element() and A.rank == B.rank`.

firedrake/slate/slate.py
thomasgibson: This is just me being paranoid and making *extra* sure. But now that I think about it, rank is probably a superfluous check.

thomasgibson:
Good to go?

wence-:
OK.

PR #1210, opened 2018-03-28
Matrix-free support for operators assembled from Slate

thomasgibson:
This adapts the `ImplicitMatrixContext` to support taking actions of Slate expressions on vectors. That way, we can solve the trace systems in hybridized solvers in a matrix-free manner.

### DIFF
--- a/firedrake/slate/preconditioners.py
+++ b/firedrake/slate/preconditioners.py
@@ -171,7 +171,8 @@ class HybridizationPC(PCBase):
                 measures.append(ds)
             else:
                 measures.extend((ds(sd) for sd in sorted(neumann_subdomains)))
-                dirichlet_subdomains = set(mesh.exterior_facets.unique_markers) - neumann_subdomains
+                markers = [int(x) for x in mesh.exterior_facets.unique_markers]
+                dirichlet_subdomains = set(markers) - neumann_subdomains
                 trace_subdomains.extend(sorted(dirichlet_subdomains))
 
             for measure in measures:
--- a/firedrake/slate/preconditioners.py
+++ b/firedrake/slate/preconditioners.py
@@ -200,7 +200,7 @@ class HybridizationPC(PCBase):
             tensor=self.schur_rhs,
             form_compiler_parameters=self.ctx.fc_params)
 
-        mat_type = PETSc.Options().getString(prefix + "S_mat_type", "aij")
+        mat_type = PETSc.Options().getString(prefix + "mat_type", "aij")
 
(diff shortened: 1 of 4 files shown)


firedrake/slate/preconditioners.py
```diff
@@ -292,8 +297,9 @@ def update(self, pc):
         """Update by assembling into the operator. No need to
         reconstruct symbolic objects.
         """
-        self._assemble_S()
-        self.S.force_evaluation()
+        if self.mat_type == "aij":
+            self._assemble_S()
+            self.S.force_evaluation()
```
wence-: This is wrong.  You need to assemble in the matrix-free case too.  That way, the PETSc KSP notices if the operator changes.

tests/slate/test_darcy_hybridized_mixed.py
```diff
@@ -63,3 +63,8 @@ def test_darcy_flow_hybridization(degree, hdiv_family):
 
     assert sigma_err < 1e-8
     assert u_err < 1e-8
+
+
+if __name__ == '__main__':
+    import os
+    pytest.main(os.path.abspath(__file__))
```
wence-: Don't bother adding this line.  It's cargo-culted rom something that florian used to do, I keep meaning to remove them.

tests/slate/test_matrix_free_hybridization.py
```diff
@@ -0,0 +1,61 @@
...
+    bc2 = DirichletBC(W[0], as_vector([0.0, sin(5*y)]), 2)
+    bcs = [bc1, bc2]
+
+    matfree_params = {'mat_type': 'matfree',
+                      'ksp_type': 'preonly',
+                      'pc_type': 'python',
+                      'pc_python_type': 'firedrake.HybridizationPC',
+                      'hybridization': {'ksp_type': 'cg',
+                                        'ksp_rtol': 1e-8,
+                                        'S_mat_type': 'matfree'}}
```
wence-: Do you not need to select a `pc_type` of `none` here?

firedrake/slate/preconditioners.py
```diff
@@ -199,13 +200,17 @@ def initialize(self, pc):
             tensor=self.schur_rhs,
             form_compiler_parameters=self.ctx.fc_params)
 
+        self.mat_type = PETSc.Options().getString(prefix + "S_mat_type", "aij")
```
wence-: The PC only makes one matrix, I would prefer if this were just called `mat_type`, rather than `S_mat_type`.

PR #1213, opened 2018-04-04
Coordinate derivative

florianwechsung:
Ufl can now (or soon) calculate shape derivatives! It's as easy as writing `derivative(J, SpatialCoordinate(mesh))`

This PR contains tests + some tiny changes to make sure the appropriate Test-/TrialFunction is created.

This needs https://bitbucket.org/fenics-project/ufl/pull-requests/90/coordinate-derivative/

### DIFF
--- a/firedrake/ufl_expr.py
+++ b/firedrake/ufl_expr.py
@@ -144,7 +144,24 @@ def derivative(form, u, du=None, coefficient_derivatives=None):
         raise ValueError("Taking derivative of form wrt u, but form contains coefficients from u.split()."
                          "\nYou probably meant to write split(u) when defining your form.")
     if du is None:
-        if isinstance(u, firedrake.Function):
+        u_is_spatialcoordinate = isinstance(u, ufl.SpatialCoordinate)
+        u_is_mesh_coordinate = u is form.ufl_domain().coordinates
+
+        if u_is_spatialcoordinate or u_is_mesh_coordinate:
+            if u_is_mesh_coordinate:
+                coords = u
+                u = SpatialCoordinate(u.ufl_domain())
+            else:
+                coords = u.ufl_domain().coordinates
+            V = coords.function_space()
+            args = form.arguments()
+            number = max(a.number() for a in args) if args else -1
+            du = Argument(V, number + 1)
+            cds = {coords: du}
+            if coefficient_derivatives is not None:
+                cds.update(coefficient_derivatives)
+            coefficient_derivatives = cds
+        elif isinstance(u, firedrake.Function):
             V = u.function_space()
             args = form.arguments()
             number = max(a.number() for a in args) if args else -1
@@ -154,11 +171,6 @@ def derivative(form, u, du=None, coefficient_derivatives=None):
                 raise ValueError("Real function space of vector elements not supported")
             mesh = form.ufl_domain()
             V = firedrake.FunctionSpace(mesh, "Real", 0)
-            args = form.arguments()
-            number = max(a.number() for a in args) if args else -1
-            du = Argument(V, number + 1)
-        elif isinstance(u, ufl.SpatialCoordinate):
-            V = u.ufl_domain().coordinates.function_space()
             args = form.arguments()
             number = max(a.number() for a in args) if args else -1
             du = Argument(V, number + 1)
--- a/firedrake/ufl_expr.py
+++ b/firedrake/ufl_expr.py
@@ -3,6 +3,7 @@ import ufl.argument
 from ufl.assertions import ufl_assert
 from ufl.split_functions import split
 from ufl.algorithms import extract_arguments, extract_coefficients
+from ufl.geometry import SpatialCoordinate
 
 import firedrake
 from firedrake import utils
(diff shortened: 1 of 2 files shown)


wence-:
Is it possible to allow the user to write either: `derivative(J, SpatialCoordinate(mesh))` or `derivative(J, mesh.coordinates)`?

Related, if I write a form that explicitly contains `mesh.coordinates`, and compute `derivative(J, SpatialCoordinate(mesh))`, do you changes DTRT?

florianwechsung:
1. This could be added, if we can figure out by looking at a `Coefficient` whether it is actually the coordinate field of a mesh and then repalce it with `SpatialCoordinate(mesh)`

2. I'm not sure its obvious what "the right thing to do" is in that case. Currently, if you use mesh.coordinates in a form, it is treated as if it was any other Function defined on mesh, with the coordinate field interpolated into it. To illustrate: this test passes
```
mesh = UnitSquareMesh(60, 60)
X = SpatialCoordinate(mesh)
u = Function(V)
u.interpolate(X)

J1 = inner(u, u) * dx
computed1 = assemble(derivative(J1, X)).dat.data
J2 = inner(mesh.coordinates, mesh.coordinates) * dx
computed2 = assemble(derivative(J2, X)).dat.data
assert np.allclose(computed1, computed2, rtol=1e-14)    

J3 = inner(X, X) * dx
computed3 = assemble(derivative(J3, X)).dat.data
assert not np.allclose(computed1, computed3, rtol=1e-14)    
```

This means that if you want to calculate the shape derivative for something like \int x*y*sin(x) dx, then you need to use `x, y = SpatialCoordinate(mesh)` for that.

wence-:
Can you fix the pep8 failures in this patch?  Type `make lint` in the source tree and fix any errors.

dham:
I think that we do want to identify `mesh.coordinates` and `SpatialCoordinate`, and that this could be done by object identity. i.e. it has to exactly be the object `mesh.coordinate`. I think that this is legitimate because `mesh.coordinate` really is the coordinate field, while `u` above is just some random field which happens to take the same value.

florianwechsung:
Fair point. From what I see, this can be achieved in two ways:

1. On firedrake side: in the case of a `CoordinateDerivative` (or even always), replace `mesh.coordinates` with `SpatialCoordinates(mesh)`. That would involve looping over all `Coefficients` `u` in a form, checking if they are equal to `u.ufl_domain().coordinates`, and in that case replacing them with the proper `SpatialCoordinate`.
2. On UFL side: currently, if the `CoordinateRuleSet` stumbles upon a `Coefficient` `u`, it returns zero. We could add a check to see whether `u==X.ufl_domain().coordinates`, and in that case return the proper Test-/TrialFunction. That would be firedrake specific though, as in dolfin, `Mesh.coordinates` is a function that returns a numpy array.

Or is there a third option that I'm missing?

florianwechsung:
Excellent idea, that works. I have made a small change in ufl and updated the firedrake side accordingly. 
the new test shows the behaviour

firedrake/ufl_expr.py
```diff
@@ -138,11 +139,29 @@ def derivative(form, u, du=None, coefficient_derivatives=None):
...
+                coords = u.ufl_domain().coordinates
+            V = coords.function_space()
+            args = form.arguments()
+            number = max(a.number() for a in args) if args else -1
+            du = Argument(V, number + 1)
+            cds = {coords: du}
+            if coefficient_derivatives is not None:
+                cds.update(coefficient_derivatives)
+            coefficient_derivatives = cds
+        elif isinstance(u, firedrake.Function):
```
wence-: Sorry for letting this languish so long.

I think we need to provide the coefficient_derivatives in all cases where a coordinate derivative is taken.

How about this logic?

```python
def derivative(f, u, du=None, coefficient_derivatives=None):
   ...
   is_dX = isinstance(u, SpatialCoordinate) or u is f.ufl_domain().coordinates

   def argument(V):
       n = max(a.number() for a in args) if args else -1
       return Argument(V, n + 1)

   if is_dX:
       coords = f.ufl_domain().coordinates
       u = SpatialCoordinate(f.ufl_domain())
       V = coords.function_space()
       if du is None:
           du = argument(V)
       cds = {coords: du}
       if coefficient_derivatives is not None:
           cds.update(coefficient_derivatives)
       coefficient_derivatives = cds
    elif isinstance (u, ...):
        ...
```

WDYT?

florianwechsung: You're absolutely right, fixed it.

tests/regression/test_coordinatederivative.py
```diff
@@ -0,0 +1,149 @@
...
+    V = FunctionSpace(mesh, "CG", 1)
+    u = Function(V)
+
+    J = u * u * dx
+    with pytest.raises(ValueError):
+        computed = assemble(Constant(2.0) * derivative(J, X))
+    with pytest.raises(ValueError):
+        computed = assemble(derivative(Constant(2.0) * derivative(J, X), X))
+    with pytest.raises(ValueError):
+        computed = assemble(Constant(2.0) * derivative(derivative(J, X), X))
```
thomasgibson: No need to assign these to anything. Just write:
```python
with pytest.raises(ValueError):
    assemble(foo)
```

wence-:
Can you fix the flake8 failures (run `make lint` until clean)

wence-:
And merged, thanks!

PR #1220, opened 2018-04-18
Update multigrid support

wence-:
Add support for grid transfer on arbitrary cell shapes.  And also between function spaces of different degree.  Needs firedrakeproject/tsfc#160

Kernels are implemented like so:

## Prolongation
Spin over fine nodes, evaluate the coarse function at the fine nodes
Requires converting fine reference cell node locations to coarse cell locations (done by pulling back to physical space then pushing forward to reference space).

## Restriction
Spin over fine nodes, spread fine coefficient out amongst coarse basis functions.
Again, requires converting fine reference cell node locations to coarse cell locations.

## CG Injection
Spin over coarse nodes, evaluate fine function at those nodes.

## DG injection
Spin over coarse cells, solve <u_c, v_c> = <f, v_c> locally on a cell.
Done by doing a macro-cell integration over the fine cells and then locally restricting that residual.  This uses a midpoint jacobian approximation for the pushforward to coarse reference space.  I conjecture that it does not overly matter, because this only needs to give me a "good enough" state.


wence-:
This borks something in our solver, we are currently debugging, so not ready to merge.

wence-:
We determined that the error was in user code (not here).  So I think this is now good to go (assuming clean jenkins).

PR #1222, opened 2018-04-23
Expose Eigen factorizations for local solves

thomasgibson:
This PR reworks the handling of matrix inverses and introduces a `Solve` node in Slate. Larger matrices will always have a factorization declared and used beforehand. This avoids explicitly storing the inverse and instead just applies the inverse of the defined factorization.

### DIFF
--- a/firedrake/slate/slac/compiler.py
+++ b/firedrake/slate/slac/compiler.py
@@ -48,8 +48,7 @@ def compile_expression(slate_expr, tsfc_parameters=None):
 
     :arg slate_expr: a :class:'TensorBase' expression.
     :arg tsfc_parameters: an optional `dict` of form compiler parameters to
-                          be passed onto TSFC during the compilation of
-                          ufl forms.
+        be passed onto TSFC during the compilation of ufl forms.
 
     Returns: A `tuple` containing a `SplitKernel(idx, kinfo)`
     """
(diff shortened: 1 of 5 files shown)


firedrake/slate/slac/compiler.py
```diff
@@ -209,32 +207,47 @@ def generate_kernel_ast(builder, statements, declared_temps):
...
-    for exp in builder.aux_exprs:
+
+    # These are either already declared terminals or expressions
+    # which do not require an extra temporary/expression
+    terminals = (slate.Tensor, slate.AssembledVector,
+                 slate.Negative, slate.Transpose)
+    statements = []
+
+    sorted_exprs = [exp for exp in topological_sort(builder.expression_dag)
+                    if (builder.ref_counter[exp] > 1 and not isinstance(exp, terminals))
```
wence-: Parenthesise like `if ((builder.ref_counter[exp] > 1 and not ...) or isinstance(exp, slate.Factorization))` to make it clearer?

thomasgibson:
I pushed some minor refactoring/edits. Should be good to go now.

firedrake/slate/slate.py
```diff
@@ -635,14 +738,19 @@ def __init__(self, A):
         assert A.shape[0] == A.shape[1], (
             "The inverse can only be computed on square tensors."
         )
+
+        if A.shape > (4, 4) and not isinstance(A, Factorization):
+            A = Factorization(A, decomposition="PartialPivLU")
+
```
wence-: This works because eigen turns `Factorisation(A).inverse()*foo` into "apply factorisation of `A` to `foo`, right?

thomasgibson: Correct. In Eigen:
```cpp
A.inverse() * b;
```
is mathematically equivalent to
```cpp
Eigen::PartialPivLU< ... > dec(A);
dec.inverse() * b;  // same as dec.solve(b)
```
In the latter case, the inverse of `A` is not explicitly computed; this just applies the already declared factorization.

PR #1227, opened 2018-05-01
Add disk caching for slate kernels

wence-:


thomasgibson:
So the test suite is failing. In particular, `tests/slate/test_slac.py` fails when asserting
```python
assert kernel_1a[0].kinfo.kernel._ast == kernel1[0].kinfo.kernel._ast
assert _kernels[0].kinfo.kernel._ast == kernel1[0].kinfo.kernel._ast
```
Initial inspection doesn't seem to show any differences in the generated code, but the `ast`s apparently are different...? However,
```python
assert kernel_1a[0].kinfo.kernel == kernel1[0].kinfo.kernel
```
evaluates to `True`.

wence-:
Fixed that issue, and the merge conflict.

PR #1228, opened 2018-05-04
mg; Reimplement skipping refinement hierarchies

wence-:
Turns out this wasn't so difficult to support now that everything is
nicely refactored.

I think @florianwechsung and @APaganini want this for fireshape.


wence-:
And now I think it might even work too!

PR #1245, opened 2018-06-27
Add support for Argyris, Bell, Hermite, and Morley elements

wence-:
Mostly tests, just need to adapt to new kernel interface that additional can provide a vertex-discretised local cell size for basis function rescaling.

### DIFF
--- a/firedrake/assemble.py
+++ b/firedrake/assemble.py
@@ -393,7 +393,7 @@ def _assemble(f, tensor=None, bcs=None, form_compiler_parameters=None,
             loops.append(zero_tensor)
         else:
             zero_tensor()
-        for indices, (kernel, integral_type, needs_orientations, subdomain_id, domain_number, coeff_map, needs_cell_facets, pass_layer_arg) in kernels:
+        for indices, (kernel, integral_type, needs_orientations, subdomain_id, domain_number, coeff_map, needs_cell_facets, pass_layer_arg, needs_cell_sizes) in kernels:
             m = domains[domain_number]
             subdomain_data = f.subdomain_data()[m]
             # Find argument space indices
@@ -499,6 +499,9 @@ def _assemble(f, tensor=None, bcs=None, form_compiler_parameters=None,
                     coords.dat(op2.READ, get_map(coords)[op2.i[0]])]
             if needs_orientations:
                 o = m.cell_orientations()
+                args.append(o.dat(op2.READ, get_map(o)[op2.i[0]]))
+            if needs_cell_sizes:
+                o = m.cell_sizes
                 args.append(o.dat(op2.READ, get_map(o)[op2.i[0]]))
             for n in coeff_map:
                 c = coefficients[n]
--- a/firedrake/assemble.py
+++ b/firedrake/assemble.py
@@ -393,7 +393,17 @@ def _assemble(f, tensor=None, bcs=None, form_compiler_parameters=None,
             loops.append(zero_tensor)
         else:
             zero_tensor()
-        for indices, (kernel, integral_type, needs_orientations, subdomain_id, domain_number, coeff_map, needs_cell_facets, pass_layer_arg, needs_cell_sizes) in kernels:
+        for indices, kinfo in kernels:
+            kernel = kinfo.kernel
+            integral_type = kinfo.integral_type
+            domain_number = kinfo.domain_number
+            subdomain_id = kinfo.subdomain_id
+            coeff_map = kinfo.coefficient_map
+            pass_layer_arg = kinfo.pass_layer_arg
+            needs_orientations = kinfo.oriented
+            needs_cell_facets = kinfo.needs_cell_facets
+            needs_cell_sizes = kinfo.needs_cell_sizes
+
             m = domains[domain_number]
             subdomain_data = f.subdomain_data()[m]
             # Find argument space indices
--- a/firedrake/bcs.py
+++ b/firedrake/bcs.py
@@ -45,7 +45,7 @@ class DirichletBC(object):
     def __init__(self, V, g, sub_domain, method="topological"):
         # First, we bail out on zany elements.  We don't know how to do BC's for them.
         import finat
-        if isinstance(V.finat_element, (finat.Hermite, finat.Argyris, finat.Morley)):
+        if isinstance(V.finat_element, (finat.Hermite, finat.Argyris, finat.Morley, finat.Bell)):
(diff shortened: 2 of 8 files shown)


rckirby:
This also disallows Dirichlet BC on the new elements since we don't know
how to support them yet.

firedrake/assemble.py
```diff
@@ -393,7 +393,7 @@ def thunk(bcs):
             loops.append(zero_tensor)
         else:
             zero_tensor()
-        for indices, (kernel, integral_type, needs_orientations, subdomain_id, domain_number, coeff_map, needs_cell_facets, pass_layer_arg) in kernels:
+        for indices, (kernel, integral_type, needs_orientations, subdomain_id, domain_number, coeff_map, needs_cell_facets, pass_layer_arg, needs_cell_sizes) in kernels:
```
miklos1: This is now really begging for something more descriptive than an n-tuple... (`namedtuple`?)

firedrake/bcs.py
```diff
@@ -43,6 +43,11 @@ class DirichletBC(object):
     '''
 
     def __init__(self, V, g, sub_domain, method="topological"):
+        # First, we bail out on zany elements.  We don't know how to do BC's for them.
+        import finat
+        if isinstance(V.finat_element, (finat.Hermite, finat.Argyris, finat.Morley)):
```
miklos1: Bell is not zany?

firedrake/mesh.py
```diff
@@ -1032,6 +1032,27 @@ def coordinates(self, value):
 
         raise AttributeError(message)
 
+    @utils.cached_property
+    def cell_sizes(self):
+        """A :class`~.Function` in the :math:`P^1` space containing the local mesh size.
+
+        This is computed by the :math:`L^2` projection of the local mesh element size."""
```
miklos1: Thanks, this now clarifies the P1 mesh size, and why that needs to pre-calculated like `cell_orientations`...

tests/regression/test_helmholtz_zany.py
```diff
@@ -0,0 +1,58 @@
...
+
+    # Analytical solution
+    f.project(Expression("cos(x[0]*pi*2)*cos(x[1]*pi*2)"))
+    return sqrt(assemble(dot(x - f, x - f) * dx)), x, f
+
+
+@pytest.mark.parametrize(('el', 'deg', 'convrate'),
+                         [('Hermite', 3, 3.8),
+                          ('Bell', 5, 4.8),
+                          ('Argyris', 5, 5)])
```
miklos1: Morley not good for Helmholtz?

rckirby:
On Tue, Jul 3, 2018 at 11:07 AM, Miklós Homolya <notifications@github.com>
wrote:

> *@miklos1* approved this pull request.
> ------------------------------
>
> In firedrake/assemble.py
> <https://github.com/firedrakeproject/firedrake/pull/1245#discussion_r199863311>
> :
>
> > @@ -393,7 +393,7 @@ def thunk(bcs):
>              loops.append(zero_tensor)
>          else:
>              zero_tensor()
> -        for indices, (kernel, integral_type, needs_orientations, subdomain_id, domain_number, coeff_map, needs_cell_facets, pass_layer_arg) in kernels:
> +        for indices, (kernel, integral_type, needs_orientations, subdomain_id, domain_number, coeff_map, needs_cell_facets, pass_layer_arg, needs_cell_sizes) in kernels:
>
> This is now really begging for something more descriptive than an
> n-tuple... (namedtuple?)
>
@wence?

> ------------------------------
>
> In firedrake/bcs.py
> <https://github.com/firedrakeproject/firedrake/pull/1245#discussion_r199863488>
> :
>
> > @@ -43,6 +43,11 @@ class DirichletBC(object):
>      '''
>
>      def __init__(self, V, g, sub_domain, method="topological"):
> +        # First, we bail out on zany elements.  We don't know how to do BC's for them.
> +        import finat
> +        if isinstance(V.finat_element, (finat.Hermite, finat.Argyris, finat.Morley)):
>
> Bell is not zany?
>

Good catch.


> ------------------------------
>
> In firedrake/mesh.py
> <https://github.com/firedrakeproject/firedrake/pull/1245#discussion_r199863742>
> :
>
> > @@ -1032,6 +1032,27 @@ def coordinates(self, value):
>
>          raise AttributeError(message)
>
> +    @utils.cached_property
> +    def cell_sizes(self):
> +        """A :class`~.Function` in the :math:`P^1` space containing the local mesh size.
> +
> +        This is computed by the :math:`L^2` projection of the local mesh element size."""
>
> Thanks, this now clarifies the P1 mesh size, and why that needs to
> pre-calculated like cell_orientations...
> ------------------------------
>
> In tests/regression/test_helmholtz_zany.py
> <https://github.com/firedrakeproject/firedrake/pull/1245#discussion_r199864237>
> :
>
> > +
> +    # Compute solution
> +    assemble(a)
> +    assemble(L)
> +    x = Function(V)
> +    solve(a == L, x, solver_parameters={'ksp_type': 'preonly', 'pc_type': 'lu'})
> +
> +    # Analytical solution
> +    f.project(Expression("cos(x[0]*pi*2)*cos(x[1]*pi*2)"))
> +    return sqrt(assemble(dot(x - f, x - f) * dx)), x, f
> +
> +
> +@pytest.mark.parametrize(('el', 'deg', 'convrate'),
> +                         [('Hermite', 3, 3.8),
> +                          ('Bell', 5, 4.8),
> +                          ('Argyris', 5, 5)])
>
> Morley not good for Helmholtz?
>

No, it is not a suitable C^0 nonconforming element.  Comment added.

tests/regression/test_helmholtz_zany.py
```diff
@@ -0,0 +1,58 @@
...
+import numpy as np
+import pytest
+
+from firedrake import *
+
+
+def helmholtz(x, el_type, degree, mesh=None):
+    # Create mesh and define function space
+    if mesh is None:
+        mesh = UnitSquareMesh(2 ** x, 2 ** x)
```
miklos1: Would it make sense to also test these elements on less regular meshes?

firedrake/bcs.py
wence-: Oh yeah, should add that.

firedrake/assemble.py
wence-: So in tsfc_interface it is a named tuple, I should destructure by name here.

firedrake/mesh.py
wence-: Yes, I'll add some explanatory comments on the tsfc side.

Issue #1247, opened 2018-07-14
Difficulty reading in 2-meshes in 3-space


pefarrell:
I'm trying to solve a problem on a global ocean domain, a triangular mesh of a manifold in three spatial dimensions. I'm using the `Earth_Global_filtered_200km_30min_metric` demo of the Shingle project (https://github.com/shingleproject/Shingle). To save anyone installing shingle, I've put a gzipped copy of the mesh at http://people.maths.ox.ac.uk/~farrellp/Earth_Global_filtered_200km_30min_metric.msh.gz .

The mesh looks fine in gmsh. However, when I try to use it, I get an error:

```
from firedrake import *

mesh = Mesh("Earth_Global_filtered_200km_30min_metric.msh", dim=3)

print(mesh.topological_dimension())
print(mesh.geometric_dimension())

V = VectorFunctionSpace(mesh, "CG", 1, dim=2)
```

yields

```
2
3
Traceback (most recent call last):
  File "test.py", line 8, in <module>
    V = VectorFunctionSpace(mesh, "CG", 1, dim=2)
  File "/home/pefarrell/local/firedrake/firedrake-dev-20180616/src/firedrake/firedrake/functionspace.py", line 169, in VectorFunctionSpace
    sub_element = make_scalar_element(mesh, family, degree, vfamily, vdegree)
  File "/home/pefarrell/local/firedrake/firedrake-dev-20180616/src/firedrake/firedrake/functionspace.py", line 40, in make_scalar_element
    mesh.init()
  File "/home/pefarrell/local/firedrake/firedrake-dev-20180616/src/firedrake/firedrake/mesh.py", line 993, in init
    self._callback(self)
  File "/home/pefarrell/local/firedrake/firedrake-dev-20180616/src/firedrake/firedrake/mesh.py", line 1332, in callback
    (self.num_vertices(), geometric_dim))
  File "firedrake/dmplex.pyx", line 851, in firedrake.dmplex.reordered_coords
ValueError: cannot reshape array of size 374816 into shape (187408,3)
```

This is strange, because the .msh file does have a 187408x3 array of coordinates in it; I suspect that the z-coordinate has been lost in the process of reading the .msh.

Incidentally, it would be nice if the mesh constructor didn't need the explicit `dim=3` argument.

pefarrell:
Yep, that fixes it. Can that be made automatic if `dim=3` is passed?

PR #1248, opened 2018-07-16
Mesh options

wence-:
Set the correct option for reading embedded Gmsh meshes when `dim` is passed into `Mesh` constructor.


pefarrell:
Fixes issue #1247 for me. Thanks!

PR #1249, opened 2018-07-16
slate: Add approximate flop counts to kernels

wence-:
A slate expression has a flop count consisting of the flops to
assemble the terminal data (approximately determined from the TSFC
kernels), and the flops to compute the expression on assembled data.
This latter is just matrix algebra, so easy to do.

@rckirby use this to get flops counts for slate kernels doing static condensation.

### DIFF
--- a/firedrake/slate/slac/kernel_builder.py
+++ b/firedrake/slate/slac/kernel_builder.py
@@ -1,6 +1,8 @@
+import numpy as np
 from coffee import base as ast
 
 from collections import OrderedDict, Counter, namedtuple
+from functools import singledispatch
 
 from firedrake.slate.slac.utils import (traverse_dags, eigen_tensor, Transformer)
 from firedrake.utils import cached_property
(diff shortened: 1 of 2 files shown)


rckirby:
Thanks, @wence- 
Are we sure we're counting facet flops right?  You multiply an interior facet integral by the number of facets a reference cell has.  I think that double-counts in some sense because those facets are shared by two cells?

wence-:
In particular, interior facet integrals have been rewritten into exterior facet integrals before being fed to tsfc: they're exterior to the domain of the cell.

rckirby:
Ok, maybe I misunderstood -- are these computed by visiting facets of cells
rather than in a separate loop over the facets?

thomasgibson:
Correct. Everything is from the perspective of the cell. At the moment, Slate kernels know nothing about the adjacent cell. It can, however, distinguish between a facet that's interior or on the domain boundary (including sub-domains). However, all facet integrals are treated as "exterior to the cell."

firedrake/slate/slac/kernel_builder.py
```diff
@@ -128,6 +130,84 @@ def dimension(e):
...
+        nfacets = self.expression.ufl_domain().ufl_cell().num_facets()
+        for ctx in self.context_kernels:
+            itype = ctx.original_integral_type
+            for k in ctx.tsfc_kernels:
+                kinfo = k.kinfo
+                if itype == "cell":
+                    flops += kinfo.kernel.num_flops
+                elif itype.startswith("interior_facet"):
+                    # Executed once per facet (approximation)
+                    flops += kinfo.kernel.num_flops * nfacets
```
thomasgibson: Just noting here that this is a bit tricky if the integral is defined over facet sub-domains. In which case, not all facets will be included in the local computation.

wence-: Yes, there's no real way to get this exactly right, short of runtime counting in some way.

PR #1253, opened 2018-07-20
Zenodo updates

wence-:
Update the zenodo stuff to chase tags and releases and provide a consolidated "single DOI for everything" release on Zenodo.  This is open for comments on extra information to include/things to change.

Things to do:

- [ ] Allow user to specify extra information to include (could be more DOIs, etc...) in free-form text
- [ ] Automatically picking up the title provided by the user

### User does:
```
$ firedrake-zenodo --output components.json --message "My cool paper"
```

### Developer does:

```
$ firedrake-zenodo --input components.json
-> Release tag TAG
# ... Make coffee for zenodo to catch up
$ firedrake-zenodo --meta-release TAG --message "My cool paper"
```

### User can then do:

```
$ firedrake-zenodo --bibtex TAG
```
as before, and gets a single bibtex entry.

## Meta records

A metarecord looks something like this: https://sandbox.zenodo.org/record/223259

The description contains DOIs, the one file is a json that is machine parseable that we could use to glue into a putative `firedrake-install --zenodo-tag TAG` in the future (So wow, much reproducible).

You get bibtex that looks like this:

```
@misc{zenodo/Firedrake_20171121.0,
 author = {zenodo/Firedrake_20171121.0},
 title  = {{Software used in 'This was fun'}},
 year   = {2018},
 month  = {jul},
 doi    = {10.5072/zenodo.223259},
 url    = {https://doi.org/10.5072/zenodo.223259},
}
```


wence-:
I think this now does all the things we want.  Please review workflow description.  Does it make sense.  @dorugeber, @dham

wence-:
Dealt with encoding issues (in case end user wants to upload a tarball).  And added some checksumming of the upload.

I think this is good to fly.

PR #1257, opened 2018-08-01
Enable the use of MG etc for the algebraic solve interface. Fixes #1256.

### LINKED ISSUE #1256: Add GMG support to algebraic solve interface
Right now, one can only say `-pc_type mg` for variational solvers.  But pyadjoint often wants to use the algebraic version (assembled operators).  We should be able to support this, but need to figure out what bits of information we should pass around (and how!).

c.f. https://bitbucket.org/dolfin-adjoint/pyadjoint/issues/59/cannot-use-firedrake-mg-firedrake_adjoint

pefarrell:
Enable the use of sophisticated solvers via DM contexts with the algebraic solve interface `solve(A, u, b)`.

Fixes firedrake #1256 and dolfin-adjoint #59 (https://bitbucket.org/dolfin-adjoint/pyadjoint/issues/59/cannot-use-firedrake-mg-firedrake_adjoint).

### DIFF
--- a/firedrake/solving.py
+++ b/firedrake/solving.py
@@ -223,12 +223,8 @@ def _la_solve(A, x, b, **kwargs):
                              near_nullspace=near_nullspace,
                              options_prefix=options_prefix)
 
-    from firedrake import dx, inner, Constant
-    v = ufl.algorithms.extract_arguments(A.a)[0]
-    L = inner(Constant(ufl.zero(v.ufl_shape)), v)*dx # linear MG doesn't need RHS form,
-                                                     # supply zero
-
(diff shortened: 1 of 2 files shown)


wence-:
Thanks! I added support for a literal 0 as the RHS in an LVP, which simplifies things.

One minor snag is that this setup doesn't work for the reused `LinearSolver` case.  While that doesn't matter for dolfin-adjoint, it's a bit of a pain.  The only snag I see is that at the point of creation, you don't know the `x` you're solving for, so you can't make an `LVP`.

pefarrell:
Great, thanks.

PR #1258, opened 2018-08-01
install: Implement firedrake-install --doi XXX

wence-:
Now that we have a new firedrake-zenodo setup, we can automate
installing a particular archived version.

### DIFF
--- a/scripts/firedrake-install
+++ b/scripts/firedrake-install
@@ -7,6 +7,7 @@ import os
 import shutil
 from glob import glob
 from argparse import ArgumentParser, RawDescriptionHelpFormatter
+import argparse
 from collections import OrderedDict
 import atexit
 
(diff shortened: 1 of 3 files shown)


scripts/firedrake-install
```diff
@@ -249,23 +318,6 @@ where shortname is one of the names given below:
...
-console = logging.StreamHandler()
-console.setLevel(logging.INFO)
-formatter = logging.Formatter('%(message)s')
-console.setFormatter(formatter)
-logging.getLogger().addHandler(console)
-
-log = logging.getLogger()
-
-log.info("Running %s" % " ".join(sys.argv))
-
```
wence-: Code movement.

scripts/firedrake-install
```diff
@@ -436,8 +488,7 @@ def git_clone(url):
         log.info("Using existing petsc installation\n")
         return name
     log.info("Cloning %s\n" % name)
-    if name in branches:
-        branch = branches[name]
+    branch = branches.get(name.lower(), branch)
```
wence-: Simplify logic.

wence-: Also, make package branch arguments case insensitive

scripts/firedrake-install
```diff
@@ -100,6 +118,53 @@ jenkins = "JENKINS_URL" in os.environ
...
+        log.error("Unable to download 'components.json'")
+        log.error("Response was {}".format(download.json()))
+        sys.exit(1)
+    # component response has checksum as "md5:HEXDIGEST", strip the md5.
+    if hashlib.md5(download.content).hexdigest() != componentjson["checksum"][4:]:
+        log.error("Download failed checksum, expecting {expect}, got {got}".format(
+            expect=componentjson["checksum"][4:],
+            got=hashlib.md5(download.content).hexdigest()))
+        sys.exit(1)
+    componentjson = download.json()
```
wence-: So we download the archived `components.json` file that `firedrake-zenodo --create-meta-release` made.  This has the commits of all the individual components we should use.

scripts/firedrake-install
```diff
@@ -100,6 +118,53 @@ jenkins = "JENKINS_URL" in os.environ
...
+    branches = {}
+    for record in componentjson["components"]:
+        commit = record["commit"]
```
wence-: I use the commit, rather than the tag, because the latter might have been moved.

wence-:
Not done here, because I assume that the repositories are not going away, is downloading the archived tarballs from zenodo, rather than checking out the reference commit from the repo.

wence-:
That checking is easy to do. Will add something. We don't need to even fail fast after download. Just try and resolve the doi and pull out the commit data and verify against known good values.

wence-:
Done in 35de95bf

PR #1265, opened 2018-08-15
Fix obscure bug with solve(A, x, b) if A had multiple DirichletBCs ap…

pefarrell:
As bc.apply(A) was called in turn, each bc would replace the previous ones,
rather than appending to A.bcs, if they applied to different components of the solution space.

### DIFF
--- a/firedrake/functionspaceimpl.py
+++ b/firedrake/functionspaceimpl.py
@@ -319,7 +319,8 @@ class FunctionSpace(object):
         # FIXME: Think harder about equality
         return self.mesh() is other.mesh() and \
             self.dof_dset is other.dof_dset and \
-            self.ufl_element() == other.ufl_element()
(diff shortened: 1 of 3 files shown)


pefarrell:
As an example, the following code prints only one BC object attached to the matrix with master, but two with my changes:

```
from firedrake import *

mesh = UnitSquareMesh(10, 10)

A = assemble(inner(grad(TrialFunction(V)), grad(TestFunction(V)))*dx)

for i in [0, 1]:
    bc = DirichletBC(V.sub(i), 0, "on_boundary") # different components
    bc.apply(A)

print(A.bcs)
```

wence-:
Ooops!  Can you add a quick test?  Basically turn your comment into a test that does `assert A.bcs == bcs_I_applied`

PR #1267, opened 2018-08-15
Fix firedrake-zenodo and add some documentation to it

dorugeber:
Also changed the authentication token environment variable names.


wence-:
Thanks!

PR #1270, opened 2018-08-20
fix expression dimension in test_slate_infrastructure.py

tj-sun:
When interpolating `"cos(x[1]*pi*2)"` in `boundary_load` on intervals, the generated code is
```
// ...
 static inline void expression_kernel (double  A[2] , double*  x_ )
{
  const double  X[2][2]  = {{1.0,-2.47782643405e-17},
  {-2.47782643405e-17,1.0}};
  double  x[1] ;
  const double  pi  = 3.141592653589793;
  {
    for (int  k  = 0; k < 2; k += 1)
    {   
      for (unsigned int  d  = 0; d < 1; d += 1)
      {
        x[d] = 0;   
        for (unsigned int  i  = 0; i < 2; i += 1)
        {
          x[d] += X[k][i] * x_[(i * 1) + d];
        }
      }
      A[k] = cos(x[1]*pi*2);
    }
  }
}
// ...
```
so we have a out of bound access at `x[1]`.

Another issue, apart from my fix may or may not make mathematical sense, is that we shouldn't actually generate code for this test. What I observe is, if I just run

> `py.test slate/test_slate_infrastructure.py::test_integral_information` 

no code is generated. But if I run

> `py.test slate/test_slate_infrastructure.py`

code for interpolation is generated at the test `slate/test_slate_infrastructure.py::test_integral_information[cg2-mesh2]`.

Presumably something triggered the evaluation of the queue, but I haven't figured out why. Can raise a separate issue if it is indeed problematic.

### DIFF
--- a/tests/slate/test_slate_infrastructure.py
+++ b/tests/slate/test_slate_infrastructure.py
@@ -1,6 +1,7 @@
 import pytest
 from firedrake import *
 from firedrake.formmanipulation import ExtractSubBlock
+import math
 
 
 @pytest.fixture(scope='module', params=[interval, triangle, quadrilateral])
@@ -58,7 +59,8 @@ def stiffness(function_space):
 @pytest.fixture
 def load(function_space):
     f = Function(function_space)
-    f.interpolate(Expression("cos(x[0]*pi*2)"))
+    x = SpatialCoordinate(function_space.mesh())
+    f.interpolate(cos(x[0])*math.pi*2)
     v = TestFunction(function_space)
     return Tensor(f * v * dx)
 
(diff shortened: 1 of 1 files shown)


tests/slate/test_slate_infrastructure.py
```diff
@@ -66,15 +66,21 @@ def load(function_space):
 @pytest.fixture
 def boundary_load(function_space):
     f = Function(function_space)
-    f.interpolate(Expression("cos(x[1]*pi*2)"))
+    if function_space.mesh().cell_dimension == 1:
```
thomasgibson: Good catch. That was totally my bad. I wrote this test ages ago and I guess I just wasn't careful. Thanks for fixing this

thomasgibson:
`slate/test_slate_infrastructure.py::test_integral_information` is testing the symbolic objects. Perhaps the code generation occurs when running the whole script because the fixtures get evaluated (`interpolate` causes code-gen)?

tj-sun:
I guess what I'm confused about is why running a single test (`test_integral_information`) doesn't generate code, but running the same test together with other tests generates code at somewhere in the middle. I would expect the tests to behave as if they are independent.

I *think* I didn't see this behaviour before rebating to master, but I'm not entirely sure.

tests/slate/test_slate_infrastructure.py
```diff
@@ -66,15 +66,21 @@ def load(function_space):
...
+        f.interpolate(Expression("cos(x[0]*pi*2)"))
+    else:
+        f.interpolate(Expression("cos(x[1]*pi*2)"))
```
dham: The reason that this generated illegal code rather than producing a nice error is that you are using the obsolete and unsafe C string interpolation. Please do this the nice way using Python objects. See: https://www.firedrakeproject.org/interpolation.html

tj-sun: Done.

PR #1272, opened 2018-08-23
PCBase for other mat_type than matfree

tlroy:
These changes allow the use of the preconditioners.PCBase class even when `mat_type` and `pmat_type` are not `matfree`. 

The `get_appctx` static method in PCBase can be used to obtain `appctx` instead of using `getPythonContext`. The current state is now always in `appctx`, not only for `matfree`.

In `solving_utils._SNESContext constructor`, `_jac`, `_assemble_jac`, `is_mixed`, `_pjac`, `_assemble_pjac`, and `_F` have been changed into `@cached_property`s so that they are no longer made in the constructor (`__init__`), but instead on first use.


firedrake/solving_utils.py
```diff
@@ -70,10 +71,13 @@ def __init__(self, problem, mat_type, pmat_type, appctx=None,
                  pre_jacobian_callback=None, pre_function_callback=None,
                  options_prefix=None):
         from firedrake.assemble import allocate_matrix, create_assembly_callable
+        self.allocate_matrix = allocate_matrix
+        self.create_assembly_callable = create_assembly_callable
```
wence-: These don't need to by instance attributes, you can instead import the relevant function in the appropriate `@cached_property`.

firedrake/solving_utils.py
```diff
@@ -82,67 +86,39 @@ def __init__(self, problem, mat_type, pmat_type, appctx=None,
...
-            # TODO, a better way of doing this.
-            # Now we don't have a temporary state inside the snes
-            # context we could just require the user to pass in the
-            # full state on the outside.
-            appctx.setdefault("state", self._x)
+        # TODO, a better way of doing this.
+        # Now we don't have a temporary state inside the snes
+        # context we could just require the user to pass in the
+        # full state on the outside.
+        appctx.setdefault("state", self._x)
```
wence-: Maintain the `# A split context` comment?

firedrake/solving_utils.py
```diff
@@ -82,67 +86,39 @@ def __init__(self, problem, mat_type, pmat_type, appctx=None,
...
             self.Jp = None
-            self._pjac = self._jac
 
-        self._F = function.Function(self.F.arguments()[0].function_space())
-        self._assemble_residual = create_assembly_callable(self.F,
-                                                           tensor=self._F,
-                                                           form_compiler_parameters=fcp)
+        self._assemble_residual = self.create_assembly_callable(self.F,
+                                                                tensor=self._F,
+                                                                form_compiler_parameters=self.fcp)
```
wence-: This could also be a `@cached_property`.

firedrake/solving_utils.py
```diff
@@ -312,3 +288,34 @@ def compute_operators(ksp, J, P):
             assert P.handle == ctx._pjac.petscmat.handle
             ctx._assemble_pjac()
             ctx._pjac.force_evaluation()
+
+    @cached_property
+    def _jac(self):
+        return self.allocate_matrix(self.J, bcs=self._problem.bcs,
```
wence-: Import `allocate_matrix` here.

wence-:
So the idea behind doing all this dance is that `ctx.split` should have a runtime independent of the problem size.  Can you check if this is the case?

Build a `LVP` on a mixed space, get the context out and split it using `master`, and this branch, with increasing mesh sizes, and see what happens.

tlroy:
Tested runtime of ctx.split for this branch vs master. This branch is independent of mesh size, while master is not. Tested this on poisson_mixed from the demos. Table of runtime (sec):
size nxn | thomasbranch | master
4 | 0.022497654 | 0.0258696079
8 | 0.0253133774 | 0.0253436565
16 | 0.0306050777 | 0.027369976
32 | 0.0234675407 | 0.0279653072
64 | 0.0226764679 | 0.0385830402
128 | 0.0303628445 | 0.0654873848
256 | 0.0242142677 | 0.2034685612
512 | 0.0255651474 | 0.8202588558
1024 | 0.0252854824 | 3.278832674

wence-:
Wonderful, thanks!

pefarrell:
@tlroy , are you planning on chasing this up? You should merge master into your branch and fix the other automated checks.

tlroy:
@wence- jenkins seems to work now. Is there anything else I need to do?

wence-:
Great, thanks!

PR #1278, opened 2018-09-03
Parloop on top or bottom

tommbendall:
Very small pull request to pass keyword arguments through the Firedrake parloop to the underlying pyop2 parloop -- intended so that the Firedrake parloop can be run on the top or bottom layers of an extruded mesh.


firedrake/parloops.py
```diff
@@ -100,7 +100,7 @@ def _form_kernel(kernel, measure, args, **kwargs):
                         "par_loop_kernel", **kwargs)
 
 
-def par_loop(kernel, measure, args, **kwargs):
+def par_loop(kernel, measure, args, kernel_kwargs={}, **kwargs):
```
miklos1: This line makes a 'potentially dangerous' alarm go off. Its use later is safe, but note that the empty dictionary here is a global value, so if you were to modify `kernel_kwargs`, then your changes were kept for the next function call.

tommbendall: Thanks @miklos1 for your advice! Is there a better way to separate out `kwargs` into those to be passed to forming the kernel and those to be sent to the `pyop2` kernel? How about:
```
def par_loop(kernel, measure, args, **kwargs):
    kernel_kwargs = {}  # kwargs to be passed to _form_kernel
    pyop2_kwargs = {}  # kwargs to be passed to underlying pyop2 kernel
    for kwarg, value in kwargs.iteritems():
        if kwarg == 'iterate':
            pyop2_kwargs[kwarg] = value
        else:
            kernel_kwarg[kwarg] = value
```

wence-: Say:
```python
def par_loop(kernel, measure, args, kernel_kwargs=None, **kwargs):
   if kernel_kwargs is None:
       kernel_kwargs = {}
   ...
```

The problem Miklos is describing is demonstrated by this:
```
In [1]: def foo(mydict, k={}):
   ...:     k.update(mydict)
   ...:     print(k)
   ...:     

In [2]: foo({'a': 'b'})
{'a': 'b'}

In [3]: foo({})
{'a': 'b'}

In [4]: 
```

Note how the second call to `foo` remembered the update from the first.  This can cause surprises!

tommbendall: Oh I see, thanks!

PR #1287, opened 2018-09-17
Implement strongly imposed Dirichlet BCs for Hermite elements in 1D.

pefarrell:


pefarrell:
It would also be useful to be able to strongly impose Dirichlet BCs on the derivative of the solution at the endpoints. I know I can code it by hand by subclassing `DirichletBC` and overriding its `nodes`, but could firedrake support this more directly?

pefarrell:
No, I can't think of a nice way to express this that extends naturally to higher dimensions. I'll stick to doing it by hand. Thanks!

dham:
@pefarrell: what is it that you think needs to be expressible in higher dimensions that is difficult?
@wence: Well we have two options for C^0. We can either do the analogous thing to what happens if you apply DirichletBC to a DG space: silently do nothing because there are no relevant DoFs to constrain. Or we can raise an exception.

In terms of how we write things, it sort of depends whether we think we also need to be able to deal with C^2. If we think that we're only ever going to support C^1 then maybe we should just have a subclass such as `DirichletGradientBC`. That's no worse namespace pollution than `VectorFunctionSpace`.

To what extent are C^2 finite elements a thing?

wence-:
C^2 is pretty esoteric unless you go splines. I will not that in higher dimensions unless you have axis aligned boundaries, dirichlet bcs are often not really possible, because you end up having to overconstrain so linear combination of basis functions. @rckirby and I have an example in the recent C^1 elements paper.

rckirby:
One should think about use cases.  For Neumann conditions for scalar
elliptic problems, one just sets du/dn strongly in the variational form.
But there are issues with trying to do that in general (as @Lawrence
Mitchell <lawrence.mitchell@imperial.ac.uk> points out).  Strong BC make
sense for Hermite in 1d, and they also (in a nonconforming sense) make
sense for Morley in 2d.  One can impose Dirichlet BC by setting the vertex
values and Neumann by setting the edge midpoint normal deriviative DOFs.
That's inexact, but keeps with the spirit of Morley and seems to work.

Enforcing BC strongly (for values or derivatives) of other elements
(Hermite, Argyris, Bell) is a complete mess, and I'm glad to keep punting
on them.

PR #1296, opened 2018-09-28
R tensorproduct

tkarna:
Adds support for foo x Real extruded function spaces.

### DIFF
--- a/firedrake/mesh.py
+++ b/firedrake/mesh.py
@@ -611,15 +611,15 @@ class MeshTopology(object):
         return op2.Dat(dataset, cell_facets, dtype=cell_facets.dtype,
                        name="cell-to-local-facet-dat")
 
-    def create_section(self, nodes_per_entity):
+    def create_section(self, nodes_per_entity, real_tensorproduct=False):
         """Create a PETSc Section describing a function space.
 
         :arg nodes_per_entity: number of function space nodes per topological entity.
         :returns: a new PETSc Section.
         """
-        return dmplex.create_section(self, nodes_per_entity)
+        return dmplex.create_section(self, nodes_per_entity, on_base=real_tensorproduct)
 
-    def node_classes(self, nodes_per_entity):
+    def node_classes(self, nodes_per_entity, real_tensorproduct=False):
         """Compute node classes given nodes per entity.
 
         :arg nodes_per_entity: number of function space nodes per topological entity.
@@ -645,7 +645,7 @@ class MeshTopology(object):
         """
         return [len(entity_dofs[d][0]) for d in sorted(entity_dofs)]
 
-    def make_offset(self, entity_dofs, ndofs):
+    def make_offset(self, entity_dofs, ndofs, real_tensorproduct=False):
         """Returns None (only for extruded use)."""
         return None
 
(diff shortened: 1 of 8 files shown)


tkarna:
Requires firedrakeproject/tsfc#173

tests/extrusion/test_real_tensorproduct.py
```diff
@@ -0,0 +1,130 @@
...
+    mesh = extmesh(2, 2, 8)
+    fspace = FunctionSpace(mesh, hfamily, hdegree, vfamily=vfamily, vdegree=vdegree)
+
+    u = TrialFunction(fspace)
+    v = TestFunction(fspace)
+
+    xyz = SpatialCoordinate(mesh)
+    if functype == 'sin':
+        expr = sin(pi*xyz[2])
+        solution_expr = Constant(2.0/pi)
```
dham: This is a bit icky. Why not just put the expression and solution in the pytest parametrisation?

tkarna: Could do that too.

tkarna: I can modify this to use string expressions if it's better.

wence-: Use parametrised fixtures

```python
@pytest.fixture
def V(extmesh):
    mesh = extmesh(2, 2, 8)
    return FunctionSpace(mesh, "CG", 1, vfamily="Real", vdegree=1)

@pytest.fixture(params=["linear", "sin"]):
def variant(request):
    return request.param

@pytest.fixture
def expr(variant, V):
    x, y, z = SpatialCoordinate(V.ufl_domain())
    if variant == "linear":
        return z
    else:
       return sin(pi*z)

@pytest.fixture
def solution(variant):
    return {"linear": Constant(0.5), "sin": Constant(2.0/pi)}[variant]

@pytest.fixture
def tolerance(variant):
    return {"linear": 1e-15, "sin": 1e-5}[variant]

def test_foo(V, expr, solution, tolerance):
   ...
```

tkarna: Huh, okay. Will do.

firedrake/mesh.py
```diff
@@ -902,10 +906,11 @@ def make_offset(self, entity_dofs, ndofs):
             entity_offset[b] += len(entities[0])
 
         dof_offset = np.zeros(ndofs, dtype=IntType)
-        for (b, v), entities in entity_dofs.items():
-            for dof_indices in entities.values():
-                for i in dof_indices:
-                    dof_offset[i] = entity_offset[b]
```
wence-: I wonder if it works to `return None` in the real case.  That is more "correct"

tkarna: Doesn't seem to work out of the box.

wence-: Might work in TJ's new reworking of the pyop2 codegen (I guess this is where things fail?).  Let's keep it in the back of our minds.

tkarna: OK

wence-:
I think this is fine. Will just wait for Jenkins to run through now that I've merged the TSFC branch.

PR #1301, opened 2018-10-03
Teach firedrake how to prolong/restrict/inject functions in Real function spaces

pefarrell:
This is necessary when coefficients are in R (as they are injected into the coarse space upon rediscretisation).


wence-:
Merged, thanks!

PR #1302, opened 2018-10-03
A preconditioner for explicitly specifiying the form to use as approximate Schur complements

pefarrell:
This adds an `ExplicitSchurPC` that gives users a hook to specify a bilinear form to be used to assemble a Schur complement approximation.

It's driven by subclassing `ExplicitSchurPC` and overriding the `form` method to supply the form and boundary conditions.

Under the hood, it's just a subclass of `AssembledPC` that gets the form from its `form` method instead of the Python context. Some small refactoring of `AssembledPC` was necessary to make this possible.

To see how this can be used, I changed the Stokes fieldsplit solve in the geometric multigrid demo. I verified that it gives exactly the same results as the old code, up to roundoff errors (KSP residual norm  `1.576276679138e-05` at iteration 22 vs `1.576276688964e-05` with the old code).

### DIFF
--- a/demos/multigrid/geometric_multigrid.py.rst
+++ b/demos/multigrid/geometric_multigrid.py.rst
@@ -167,8 +167,15 @@ remembering to tell PETSc to use pivoting in the factorisation. ::
                                                "ksp_monitor": True,
                                                "pmat_type": "aij"})
 
-Next we'll use a schur complement solver, using geometric multigrid to
+Next we'll use a Schur complement solver, using geometric multigrid to
 invert the velocity block. ::
+
+  class Mass(ExplicitSchurPC):
+
+      def form(self, test, trial, pc):
+          a = inner(test, trial)*dx
+          bcs = None
+          return (a, bcs)
 
   parameters = {
       "ksp_type": "gmres",
@@ -179,19 +186,17 @@ invert the velocity block. ::
       "fieldsplit_0_ksp_type": "preonly",
       "fieldsplit_0_pc_type": "mg",
       "fieldsplit_1_ksp_type": "preonly",
-      "fieldsplit_1_pc_type": "bjacobi",
-      "fieldsplit_1_sub_pc_type": "icc",
+      "fieldsplit_1_pc_type": "python",
+      "fieldsplit_1_pc_python_type": "__main__.Mass",
+      "fieldsplit_1_schur_pc_type": "bjacobi",
+      "fieldsplit_1_schur_sub_pc_type": "icc",
   }
 
-We provide an auxiliary operator so that we can precondition the schur
-complement inverse with a pressure mass matrix. ::
-
-  Jp = a + p*q*dx
   u = Function(Z)
-  solve(a == L, u, bcs=bcs, Jp=Jp, solver_parameters=parameters)
+  solve(a == L, u, bcs=bcs, solver_parameters=parameters)
 
 Finally, we'll use coupled geometric multigrid on the full problem,
-using schur complement "smoothers" on each level.  On the coarse grid
+using Schur complement "smoothers" on each level.  On the coarse grid
 we use a full factorisation with LU for the velocity block, whereas on
 the finer levels we use incomplete factorisations for the velocity
 block.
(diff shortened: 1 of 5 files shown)


wence-:
This broke some tests because in certain cases the application context was not transferred to split DMs.  In particular, if you said `pc_fieldsplit_0_fields: 0,1` (say), then we called `create_subdm` and didn't transfer the appctx for the `0,1` fields over to the subdm.

I /think/ I have fixed this (it fixes the breakage on Jenkins).  But have not tested things at all.

A trivial test to ensure things work right:  Solve a Stokes problem (with BCs) in a 3x3 system where the 3rd field solves an uncoupled poisson problem (also with BCs).

I hope that now works right, but haven't tested yet.

demos/multigrid/geometric_multigrid.py.rst
```diff
@@ -167,9 +167,16 @@ remembering to tell PETSc to use pivoting in the factorisation. ::
...
+Next we'll use a Schur complement solver, using geometric multigrid to
 invert the velocity block. ::
 
+  class Mass(ExplicitSchurPC):
+
+      def form(self, test, trial, pc):
+          a = inner(test, trial)*dx
+          bcs = None
+          return (a, bcs)
+
```
wence-: You could now address the note lower down, and push the mass matrix into the coarse grid fieldsplit in the monolithic case.

firedrake/mg/ufl_utils.py
```diff
@@ -122,6 +122,9 @@ def coarsen_bc(bc, self, coefficient_mapping=None):
 @coarsen.register(firedrake.functionspaceimpl.FunctionSpace)
 @coarsen.register(firedrake.functionspaceimpl.WithGeometry)
 def coarsen_function_space(V, self, coefficient_mapping=None):
+    if hasattr(V, "_coarse"):
+        return V._coarse
+    fine = V
```
wence-: UGH, this is ugly, but the cargo-culting appears to work.

firedrake/preconditioners/assembled.py
```diff
@@ -72,3 +94,12 @@ def view(self, pc, viewer=None):
...
             self.pc.view(viewer)
+
+
+class ExplicitSchurPC(AssembledPC):
+
+    _prefix = "schur_"
+
+    def form(self, test, trial, pc):
+        """Return (a, bcs)"""
+        raise NotImplementedError
```
wence-: Prefer

```python
import abc
class ExplicitSchurPC(AssembledPC, metaclass=abc.ABCMeta):

    @abstractmethod
    def form(self, test, trial, pc):
        ....
```

As well, would it make sense to mimic all the other methods for these classes and reorder so that the signature is:

```python
def form(self, pc, test, trial)
```

Can you also elaborate in the docstring on what the arguments are?

Arguably, the name is a little off, in that this doesn't really have anything to do with Schur complements.  I can't think of a better one straight away though.

demos/multigrid/geometric_multigrid.py.rst
```diff
@@ -167,9 +167,16 @@ remembering to tell PETSc to use pivoting in the factorisation. ::
                                                "ksp_monitor": True,
                                                "pmat_type": "aij"})
 
-Next we'll use a schur complement solver, using geometric multigrid to
+Next we'll use a Schur complement solver, using geometric multigrid to
 invert the velocity block. ::
 
+  class Mass(ExplicitSchurPC):
```
wence-: Can you describe what it is you're doing here?  This comes out of nowhere for the demo, it seems to me.

wence-:
The issues are mostly minor/documentation.

firedrake/preconditioners/assembled.py
pefarrell: I've added the `abc.abstractmethod` stuff, changed the arguments order, and added more complete docstrings.

demos/multigrid/geometric_multigrid.py.rst
pefarrell: I've added a brief explanation that the Schur complement is spectrally equivalent to the viscosity-weighted mass matrix, and that we will supply it in this way.

demos/multigrid/geometric_multigrid.py.rst
pefarrell: Done. I also used the mass matrix in the smoothers on the finer levels, not sure if that's what you wanted.

wence-:
I think I fixed the breakage in ctx.split, let's wait for jenkins.

wence-:
One test is still failing which I think is due to my changes. Will debug tomorrow.

pefarrell:
Great, thanks!

PR #1303, opened 2018-10-04
mg: Default to inheriting distribution_parameters

wence-:
The normal case is that we want refined meshes to have the same
overlap as coarse meshes.  So make that the default, with the
possibility to override.


pefarrell:
Yep, I confirm this branch fixes the problem with the H(div) solver in parallel, even without explicitly passing the distribution parameters to the `MeshHierarchy`. Thanks!

PR #1306, opened 2018-10-08
Get the P1PC working with the new preconditioning structure

pefarrell:


dham:
Is this tested somewhere? A very cursory grep of the tests directory appeared to indicate not.

pefarrell:
I've lifted Lawrence's test from the old ssc repository as `tests/regression/test_p1pc.py`.

PR #1307, opened 2018-10-08
fix for setting abs, sin and cos

jshipton:
For some reason python doesn't like setting these in one go. There may be a better fix...


wence-:
If binds tighter than comma is the reason I think

dham:
I claim I've fixed this the way Wence suggested on master. Complain if this is not true.

PR #1308, opened 2018-10-09
Some fixes to do with VectorFunctionSpace(mesh, "R", 0).

pefarrell:
Fixes some crashes when you try to solve nonlinear algebraic systems with firedrake. There are other bugs, but this is a necessary first step to run anything.

### DIFF
--- a/firedrake/functionspaceimpl.py
+++ b/firedrake/functionspaceimpl.py
@@ -839,12 +839,14 @@ class RealFunctionSpace(FunctionSpace):
     """
 
     finat_element = None
-    rank = 0
-    shape = ()
-    value_size = 1
 
     def __init__(self, mesh, element, name):
         self._ufl_element = element
+
+        self.shape = element.value_shape()
+        self.rank = len(self.shape)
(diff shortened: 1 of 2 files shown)


pefarrell:
Ah, it seems that putting all the components to 1 in `_real_mangle` has the effect that all the residuals are added to the first component of the residual, and the rest are zero; and that only the first component of the solution is ever updated:

```
from firedrake import *

mesh = UnitIntervalMesh(2)

V = VectorFunctionSpace(mesh, "R", 0, dim=2)
z = Function(V)

(a, b) = split(z)
(c, d) = split(TestFunction(V))

g = 2 # the (wrong) solution depends on this
F = inner(a - 1, c)*dx + inner(b - g, d)*dx
solve(F == 0, z, options_prefix="")
print(z.dat.data)
```

Running with `-ksp_monitor_solution -snes_monitor` shows that only the first component is ever updated.

I guess the approach taken here to circumvent the crash in `_real_mangle` is wrong: I guess I have to make TSFC understand Real function spaces.

PR #1309, opened 2018-10-10
Embedded prolongation

wence-:
Uses @pefarrell's code to do transfer for all spaces we support by embedding the non-natively supported ones in a DG space. As we implement more support for this stuff in FInAT, the set of spaces that we need to support using this method will shrink.

To map back and forth between the space and embedding DG space there exists the option either to use a Fortin projection (the default, cheap), or iterative mass solves (more expensive).

A test is added inverting the H(div) and H(curl) Riesz maps using the multigrid scheme from Arnold, Falk, and Winther (2000).

### DIFF
--- a/firedrake/mg/embedded.py
+++ b/firedrake/mg/embedded.py
@@ -1,0 +1,300 @@
+import firedrake
+import ufl
+from firedrake.petsc import PETSc
+
+
+__all__ = ("EmbeddedDGTransfer", )
+
+
+native = frozenset(["Lagrange", "Discontinuous Lagrange", "Real", "Q", "DQ"])
+
+
+def EmbeddedDGTransfer(element, use_fortin_interpolation=True):
+    """Return an object that provides grid transfers.
+
+    This object provides transfers of functions between levels in a
+    mesh hierarchy for element types that are not supported directly
+    by Firedrake's multilevel implementation.  It computes transfers
+    by embedding the element in an appropriate DG space, transferring
+    the resulting DG function, and then mapping back.
+
+    :arg element: The UFL element that will be transferred.
+    :arg use_fortin_interpolation: If True, use Fortin operator
+       as an approximate mass inverse, otherwise, use an actual mass
+       solve.
+    """
+    if type(element) is firedrake.MixedElement:
+        return MixedTransfer(element, use_fortin_interpolation=use_fortin_interpolation)
+    else:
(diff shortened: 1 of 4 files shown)


firedrake/mg/embedded.py
```diff
@@ -0,0 +1,300 @@
...
+        :arg V: function space to compute weights for.
+        :returns: A PETSc Vec.
+        """
+        key = V.dim()
+        try:
+            return self._V_dof_weights[key]
+        except KeyError:
+            f = firedrake.Function(V)
+            firedrake.par_loop(
+                """for (int i = 0; i < A.dofs; i++)
```
pefarrell: Can you add a comment explaining what this `par_loop` is doing?

tests/multigrid/test_embedded_transfer.py
```diff
@@ -0,0 +1,112 @@
+import pytest
+from firedrake import *
+
+
+solver_parameters = {
+    "mat_type": "aij",
+    "snes_type": "ksponly",
+    "ksp_type": "cg",
```
pefarrell: Should this be `fcg`  for the case `exact=True`, as you're doing a solve with CG inside the transfer?

wence-: Do you mean `use_fortin_interpolation=False`? I suspect it should be `fcg` in that case, because yes, the V-cycle is no longer a linear operator.

firedrake/mg/embedded.py
wence-: This is just spinning over all the cells (this process owns) and adding one to every dof that the cell can see.  After halo exchange this has computed the dof weights.

tests/multigrid/test_embedded_transfer.py
pefarrell: Yes, sorry, I meant `use_fortin_interpolation=False` (I used to have `exact=True` in my code, called it that, then edited one side of the equation but not the other)

firedrake/mg/embedded.py
pefarrell: Yes, I meant in the code.

wence-: Done in ecbbeb8b

tests/multigrid/test_embedded_transfer.py
wence-: Done in f26d11ac.

PR #1314, opened 2018-10-22
Add bounding box location call

cjlh:
Baby steps towards parallel point location.


firedrake/spatialindex.pyx
```diff
@@ -1,9 +1,11 @@
 from __future__ import absolute_import, print_function, division
```
wence-: While we're here (changing this file), we can remove the `__future__` import, since we don't support Python 2 any more.

PR #1315, opened 2018-10-24
Test Bernstein element

miklos1:


wence-:
Rebased on master to hopefully run through Jenkins.

wence-:
Humph, jenkins just died instead.

PR #1316, opened 2018-10-26
Change default MPI packages for linux

jandrej:
This changes the default MPI packages from OpenMPI to MPICH for linux in the install script.

This was tested on Ubuntu 18.04 and passed `make alltest`


dham:
Note that the test pass here is a bit spurious because Jenkins doesn't have permission to run apt, so mpich is not actually installed on the build slave.

PR #1318, opened 2018-10-27
tests: Fix pytest cleanup failure in output tests

wence-:
Seems like the tmpdir fixture doesn't like the parallel run we were
doing. So fabricate our own.


wence-:
This should, I hope, fix the test hangs we are seeing.

PR #1323, opened 2018-11-05
Fix for Stokes notebook

jwallwork23:
Fix for Stokes flow notebook example, as recommended by @wence-

### DIFF
--- a/firedrake/functionspaceimpl.py
+++ b/firedrake/functionspaceimpl.py
@@ -514,6 +514,9 @@ class FunctionSpace(object):
         """
         return self._shared_data.boundary_nodes(self, sub_domain, method)
 
+    def collapse(self):
+        return FunctionSpace(self.mesh(), self.ufl_element())
+
 
 class MixedFunctionSpace(object):
     r"""A function space on a mixed finite element.


firedrake/functionspaceimpl.py
```diff
@@ -514,6 +514,9 @@ def boundary_nodes(self, sub_domain, method):
         """
         return self._shared_data.boundary_nodes(self, sub_domain, method)
 
+    def collapse(self):
+        return FunctionSpace(self.mesh(), self.ufl_element())
```
wence-: This is wrong, you need:

```python
def collapse(self):
   from firedrake import FunctionSpace
   return FunctionSpace(self.mesh(), self.ufl_element())
```

This is so you get a `WithGeometry` appropriately.

jwallwork23: Gotcha. Thanks, updated below

wence-: I presume this still works for the notebook. Can you add a simple test that this does the "correct" thing?

That is, collapsing subspaces should return something that is `==` to just making the subspace "directly"

jwallwork23: The notebook works on my machine with this, yes.

Will write a test now

jwallwork23: Maybe I misunderstood something, but my simple test below fails, with
```
    def test_function_space_collapse(cg1):
>       assert cg1 == cg1.collapse()
E       assert WithGeometry(...), dim=2), 1)) == FunctionSpace(...1), name=None)
```

wence-: Ah, that means, I think you need to implement `collapse` in `WithGeometry` as:

```python
def collapse(self):
    return type(self)(self.topological.collapse(), self.mesh())
```

jwallwork23: Cheers Lawrence, test passes now

PR #1327, opened 2018-11-12
Multiple transferoperators

wence-:


firedrake/dmhooks.py
```diff
@@ -64,11 +69,21 @@ def set_function_space(dm, V):
...
+    indices = []
+
+    Vtmp = V
+    while Vtmp.parent is not None:
+        indices.append(Vtmp.index)
+        Vtmp = Vtmp.parent
+    element = V.ufl_element()
+
+    info = (weakref.ref(mesh), element, tuple(reversed(indices)))
+    dm.setAttr("__fs_info__", info)
```
wence-: FIxing, hopefully, problems to do with not finding functionspaces on DMs. We must have a mesh lying around, so store that and the necessary information to make the FS from it (instead of a weakref to the FS).

firedrake/linear_solver.py
```diff
@@ -149,14 +149,13 @@ def solve(self, x, b):
...
-                    acc = x.dat.vec_wo
-                with acc as solution:
-                    self.ksp.solve(rhs, solution)
+        if self.ksp.getInitialGuessNonzero():
+            acc = x.dat.vec
+        else:
+            acc = x.dat.vec_wo
+
+        with self.inserted_options(), b.dat.vec_ro as rhs, acc as solution:
+            self.ksp.solve(rhs, solution)
```
wence-: No functional changes.

firedrake/mg/interface.py
```diff
@@ -31,7 +31,8 @@ def prolong(coarse, fine):
         if len(Vc) != len(Vf):
             raise ValueError("Mixed spaces have different lengths")
         for in_, out in zip(coarse.split(), fine.split()):
-            prolong(in_, out)
+            myprolong, _, _ = firedrake.dmhooks.get_transfer_operators(in_.function_space().dm)
+            myprolong(in_, out)
```
wence-: Don't hardcode the transfer operators here so that this does the right thing when we have custom ones.

firedrake/mg/ufl_utils.py
```diff
@@ -138,11 +138,22 @@ def coarsen_function_space(V, self, coefficient_mapping=None):
...
+    Vf = V
     V = firedrake.FunctionSpace(mesh, V.ufl_element())
+
+    from firedrake.dmhooks import get_transfer_operators, push_transfer_operators
+    transfer = get_transfer_operators(Vf.dm)
+    push_transfer_operators(V.dm, *transfer)
+    if len(V) > 1:
+        for V_, Vc_ in zip(Vf, V):
+            transfer = get_transfer_operators(V_.dm)
+            push_transfer_operators(Vc_.dm, *transfer)
```
wence-: Here is another place where we coarsen function spaces (not via DMCoarsen) so we must attach transfer operators here. This should be refactored so that there is not code duplication with the similar stuff in dmhooks.py

firedrake/variational_solver.py
```diff
@@ -227,17 +229,16 @@ def solve(self, bounds=None):
...
-
+        with self._problem.u.dat.vec as u:
+            u.copy(work)
+            with ExitStack() as stack:
+                # Ensure options database has full set of options (so monitors
+                # work right)
+                for ctx in chain((self.inserted_options(), dmhooks.appctx(dm, self._ctx)),
+                                 self._transfer_operators):
+                    stack.enter_context(ctx)
+                self.snes.solve(None, work)
```
wence-: Ugh, of course python people don't allow you to write:

```python
with *list_of_context_managers:
   ...
```

wence-:
I think these are now correct.

colinjcotter:
Is there a way to directly test the transfer operators other than using them in a preconditioner i.e. checking that they actually do what they are supposed to?

wence-:
Yes, I should hook them up with the transfer testing that is done for the existing spaces.

wence-:
> Is there a way to directly test the transfer operators other than using them in a preconditioner i.e. checking that they actually do what they are supposed to?

Done.

PR #1328, opened 2018-11-12
Changes to handle multigrid on barycentrically refined meshes

wence-:


tests/multigrid/test_grid_transfer.py
```diff
@@ -1,6 +1,488 @@
...
+import firedrake
+from fractions import Fraction
+from firedrake.mg import impl
+from firedrake.petsc import *
+import numpy
+
+def closure(dm, p):
+    return dm.getTransitiveClosure(p, useCone=True)
+
+def bary(cdm):
```
wence-: Where should this live? Not here.

wence-: Removed.

tests/multigrid/test_grid_transfer.py
```diff
@@ -114,29 +132,38 @@ def run_prolongation(hierarchy, vector, space, degrees):
...
-            restrict(actual, tmp)
-            actual = tmp
-            assert numpy.allclose(expect.dat.data_ro, actual.dat.data_ro)
+            coarse_dual = Function(Vc)
+            fine_primal = Function(Vf)
+            restrict(fine_dual, coarse_dual)
+            prolong(coarse_primal, fine_primal)
+            coarse_functional = functional(coarse_primal, coarse_dual)
+            fine_functional = functional(fine_primal, fine_dual)
+
```
wence-: This test changes because on non-nested hierarchies, this is the thing that is being preserved.

Issue #1238, opened 2018-06-02
Segfault in Eigen when using slate to assemble the inverse of a DG mass matrix


pefarrell:
Please consider the following code:

```
from firedrake import *

mesh = UnitSquareMesh(2, 2)
V = FunctionSpace(mesh, "Discontinuous Lagrange", 3) # fails
#V = FunctionSpace(mesh, "Discontinuous Lagrange", 2) # works

massinv = assemble(Tensor(inner(TrialFunction(V), TestFunction(V))*dx).inv)
massinv.force_evaluation()
```

On my workstation, running this code yields a segfault:

```
Thread 1 "python" received signal SIGSEGV, Segmentation fault.
0x00007fffd630579d in Eigen::internal::gemm_pack_lhs<double, long, Eigen::internal::blas_data_mapper<double, long, 0, 0>, 24, 8, 0, false, true>::operator()(double*, Eigen::internal::blas_data_mapper<double, long, 0, 0> const&, long, long, long, long) () from /scratch2/farrellp/install-scripts/firedrake/firedrake-dev-20180430/.cache/pyop2/f6bb6d68e82fd2964da2b1383aa0cf7c.so
(gdb) bt
#0  0x00007fffd630579d in Eigen::internal::gemm_pack_lhs<double, long, Eigen::internal::blas_data_mapper<double, long, 0, 0>, 24, 8, 0, false, true>::operator()(double*, Eigen::internal::blas_data_mapper<double, long, 0, 0> const&, long, long, long, long) () from /scratch2/farrellp/install-scripts/firedrake/firedrake-dev-20180430/.cache/pyop2/f6bb6d68e82fd2964da2b1383aa0cf7c.so
#1  0x00007fffd63063df in Eigen::internal::triangular_solve_matrix<double, long, 2, 6, false, 0, 0>::run(long, long, double const*, long, double*, long, Eigen::internal::level3_blocking<double, double>&) ()
   from /scratch2/farrellp/install-scripts/firedrake/firedrake-dev-20180430/.cache/pyop2/f6bb6d68e82fd2964da2b1383aa0cf7c.so
#2  0x00007fffd62ffec7 in wrap_compile_slate () from /scratch2/farrellp/install-scripts/firedrake/firedrake-dev-20180430/.cache/pyop2/f6bb6d68e82fd2964da2b1383aa0cf7c.so
[...]

Issue #1221, opened 2018-04-23
Slate: high-degree problems surpasses Eigen's stack allocation limit for fixed matrix types


thomasgibson:
I'm logging this issue here for now and will gradually update as solutions are found.

This came up while running some HDG tests in three-dimensions. I've added a code snippet of the Slate-generated code for the static condensation of a three-dimensional HDG problem (degree 5 on tetrahedra).
```cpp
 static inline void compile_slate (double  A2[84][84] , double *  coords , double *  w_0 , int8_t *  arg_cell_facets )
{
  int8_t (*cell_facets)[2] = (int8_t (*)[2])arg_cell_facets;
  /* Declare and initialize */
  Eigen::Matrix<double, 308, 308, Eigen::RowMajor>  T0 ;
  T0.setZero();
  /* Assemble local tensors */
  subkernel0_cell_to_00_cell_integral_otherwise(T0.block<168, 168>(0, 0), coords);
  subkernel0_cell_to_01_cell_integral_otherwise(T0.block<168, 56>(0, 168), coords);
  subkernel0_cell_to_10_cell_integral_otherwise(T0.block<56, 168>(168, 0), coords);
  subkernel0_cell_to_11_cell_integral_otherwise(T0.block<56, 56>(168, 168), coords);
  /* Loop over cell facets */
  
  for (unsigned int  i0  = 0; i0 < 4; i0 += 1)
  {
    if (cell_facets[i0][0] == 0) {
      subkernel0_exterior_facet_to_02_exterior_facet_integral_otherwise(T0.block<168, 84>(0, 224), coords, w_0, &i0);
      subkernel0_exterior_facet_to_10_exterior_facet_integral_otherwise(T0.block<56, 168>(168, 0), coords, w_0, &i0);
      subkernel0_exterior_facet_to_11_exterior_facet_integral_otherwise(T0.block<56, 56>(168, 168), coords, w_0, &i0);
      subkernel0_exterior_facet_to_12_exterior_facet_integral_otherwise(T0.block<56, 84>(168, 224), coords, w_0, &i0);
      
    }
     if (cell_facets[i0][0] == 1) {
      subkernel0_interior_facet_to_02_exterior_facet_integral_otherwise(T0.block<168, 84>(0, 224), coords, w_0, &i0);
      subkernel0_interior_facet_to_10_exterior_facet_integral_otherwise(T0.block<56, 168>(168, 0), coords, w_0, &i0);
[...]

PR #1335, opened 2018-11-15
Pefarrell/supermesh mixed mass matrix

wence-:
Provide facility to assemble a mass matrix between two spaces on non-matching meshes.

This doesn't hook up libsupermeshes' intersection finder, but assumes that you can somehow provide a list of candidate cells.

`project(f, V)` where `V` is on a different mesh to `f` for this case also works.

We use this for DG injection on non-nested meshes in multigrid.

We haven't attempted to hook up with the general assembly routine (which would require more thinking about what the right interface looks like).


firedrake/dmplexinc.pxi
```diff
@@ -4,6 +4,7 @@ cimport mpi4py.MPI as MPI
 cdef extern from "petsc.h":
    ctypedef long PetscInt
    ctypedef double PetscReal
+   ctypedef double PetscScalar
```
wence-: More copying from petsc4py.

firedrake/projection.py
```diff
@@ -1,22 +1,69 @@
...
 from firedrake import function
-import firedrake.variational_solver as vs
+from pyop2.utils import as_tuple
 
 
 __all__ = ['project', 'Projector']
 
-# Store the solve function to use in a variable so external packages
-# (dolfin-adjoint) can override it.
-_solve = solving.solve
```
wence-: No longer relevant, since pyadjoint doesn't do this.

firedrake/projection.py
```diff
@@ -35,72 +81,129 @@ def project(v, V, bcs=None, mesh=None,
...
     ``V`` and ``V`` is returned. If `V` is a :class:`.FunctionSpace`
     then ``v`` is projected into a new :class:`.Function` and that
-    :class:`.Function` is returned.
+    :class:`.Function` is returned."""
 
-    The ``mesh`` and ``form_compiler_parameters`` are currently ignored."""
-    from firedrake import function
+    val = Projector(v, V, bcs=bcs, solver_parameters=solver_parameters,
+                    form_compiler_parameters=form_compiler_parameters).project()
+    val.rename(name)
```
wence-: Just use the `Projector` object.

firedrake/projection.py
```diff
@@ -35,72 +81,129 @@ def project(v, V, bcs=None, mesh=None,
...
 
-        if v.rank() == 0:
-            fs = functionspace.FunctionSpace(V.mesh(), 'DG', deg+1)
-        elif v.rank() == 1:
-            fs = functionspace.VectorFunctionSpace(V.mesh(), 'DG',
-                                                   deg+1,
-                                                   dim=shape[0])
+    def project(self):
+        self.target.assign(self.source)
+        return self.target
```
wence-: Used in the case where the source and target are both functions in the same space.

firedrake/projection.py
```diff
@@ -35,72 +81,129 @@ def project(v, V, bcs=None, mesh=None,
...
+        element = self.target.function_space().finat_element
+        return element.entity_closure_dofs() == element.entity_dofs()
+
+    @cached_property
+    def A(self):
+        u = firedrake.TrialFunction(self.target.function_space())
+        v = firedrake.TestFunction(self.target.function_space())
+        a = firedrake.inner(u, v)*firedrake.dx
+        if self.is_DG and self.bcs is None:
+            a = firedrake.Tensor(a).inv
```
wence-: This is maybe bad, because for high order spaces (especially tensor-valued ones), we overflow Eigen's stack limits. Could remove it.

firedrake/projection.py
```diff
@@ -35,72 +81,129 @@ def project(v, V, bcs=None, mesh=None,
...
+                def solve(self, x, b):
+                    self.A.force_evaluation()
+                    with x.dat.vec_wo as x_, b.dat.vec_ro as b_:
+                        self.A.M.handle.mult(b_, x_)
+                    return x
+
+            return Dummy(self.A)
+        else:
+            solver = firedrake.LinearSolver(self.A, solver_parameters=self.solver_parameters)
+            return solver
```
wence-: That would also simplify this code, so probably one should do that.

firedrake/projection.py
```diff
@@ -35,72 +81,129 @@ def project(v, V, bcs=None, mesh=None,
...
+
+    @property
+    def apply_massinv(self):
+        if not self.constant_jacobian:
+            firedrake.assemble(self.A.a, tensor=self.A, bcs=self.bcs,
```
wence-: I think this does one too many assemblies, because it makes `A` and then reassembles it.

firedrake/projection.py
```diff
@@ -35,72 +81,129 @@ def project(v, V, bcs=None, mesh=None,
...
+    @cached_property
+    def mixed_mass(self):
+        from firedrake.supermeshing import assemble_mixed_mass_matrix
+        return assemble_mixed_mass_matrix(self.source.function_space(),
+                                          self.target.function_space())
+
+    @property
+    def rhs(self):
+        with self.source.dat.vec_ro as u, self.residual.dat.vec_wo as v:
+            self.mixed_mass.mult(u, v)
```
wence-: RHS is built by hitting the source vector with the mass matrix from the old space to the target space.  This is just integrating against a test function, but not supermeshing every time.

firedrake/projection.py
```diff
@@ -35,72 +81,129 @@ def project(v, V, bcs=None, mesh=None,
...
+        return self.residual
+
+
+def Projector(v, v_out, bcs=None, solver_parameters=None,
+              form_compiler_parameters=None, constant_jacobian=True):
```
wence-: functional interface for ease of dispatch.

firedrake/projection.py
```diff
@@ -35,72 +81,129 @@ def project(v, V, bcs=None, mesh=None,
...
+    def rhs(self):
+        self.assembler()
+        return self.residual
+
+
+class SupermeshProjector(ProjectorBase):
+    @cached_property
+    def mixed_mass(self):
+        from firedrake.supermeshing import assemble_mixed_mass_matrix
+        return assemble_mixed_mass_matrix(self.source.function_space(),
```
wence-: Needs to be reassembled if meshes are moving I think.

firedrake/supermeshimpl.pyx
```diff
@@ -0,0 +1,82 @@
...
+                               const double *, const double *,
+                               const double *, double *)
+
+
+cdef extern from "petscmat.h" nogil:
+    int MatSetValuesLocal(PETSc.PetscMat, PetscInt, const PetscInt[], PetscInt, const PetscInt[],
+                          const PetscScalar[], PetscInt)
+
+
+def assemble_mixed_mass_matrix(V_A, V_B, candidates,
```
wence-: Just a cell loop, calling a compiled kernel.

firedrake/supermeshimpl.pyx
```diff
@@ -0,0 +1,82 @@
...
+    simplices_C = numpy.empty(MAGIC[gdim], dtype=ScalarType)
+
+    vertices_A = mesh_A.coordinates.dat.data_ro_with_halos
+    vertices_B = mesh_B.coordinates.dat.data_ro_with_halos
+    V_A_cell_node_map = V_A.cell_node_map().values_with_halo
+    V_B_cell_node_map = V_B.cell_node_map().values_with_halo
+    num_dof_A = V_A.cell_node_map().arity
+    num_dof_B = V_B.cell_node_map().arity
+    for cell_A in range(num_cell_A):
+        for cell_B in candidates(cell_A):
```
wence-: candidates is a python function call, so maybe not superfast?

tests/regression/test_assemble_mixed_mass_matrix.py
```diff
@@ -0,0 +1,67 @@
...
+    print("nrm: %s" % nrm)
+    assert nrm < 1.0e-10
+
+
+if __name__ == "__main__":
+
+    class ThereMustBeABetterWay(object):
+        param = 3
+
+    test_assemble_mixed_mass_matrix(mesh(ThereMustBeABetterWay()), "CG", "CG", 1, 1)
```
wence-: This bit should go.

tests/regression/test_assemble_mixed_mass_matrix.py
```diff
@@ -0,0 +1,67 @@
...
+    M = assemble_mixed_mass_matrix(V_A, V_B)
+
+    M_ex = assemble(inner(TrialFunction(V_A), TestFunction(V_B)) * dx)
+    M_ex.force_evaluation()
+    M_ex = M_ex.M.handle
+
+    M_ex.axpy(-1.0, M)
+    nrm = M_ex.norm(PETSc.NormType.NORM_INFINITY)
+    print("nrm: %s" % nrm)
+    assert nrm < 1.0e-10
```
wence-: This test tests that the supermesh assembly does the right thing when both meshes are the same mesh.

tests/regression/test_assemble_mixed_mass_matrix.py
```diff
@@ -0,0 +1,67 @@
+from firedrake import *
```
wence-: Should move to `tests/supermesh/...`

tests/supermesh/test_galerkin_projection.py
```diff
@@ -0,0 +1,90 @@
...
+
+    assert norm < 1.0e-12
+
+
+if __name__ == "__main__":
+
+    class ThereMustBeABetterWay(object):
+        param = 2
+
+    test_galerkin_projection(mesh(ThereMustBeABetterWay()), lambda x: x, "CG", "CG", 2, 3)
```
wence-: This bit should go.

tests/supermesh/test_galerkin_projection.py
```diff
@@ -0,0 +1,90 @@
...
+        "pc_factor_mat_solver_type": "mumps",
+    }
+
+    f_B_project = project(f_A, V_B, solver_parameters=solver_parameters)
+
+    diff = Function(V_B)
+    diff.assign(f_B_prolong - f_B_project)
+    norm = sqrt(assemble(inner(diff, diff)*dx))
+
+    assert norm < 1.0e-12
```
wence-: This tests that supermesh projection when the target mesh is nested works.

tests/supermesh/test_nonnested_project.py
```diff
@@ -0,0 +1,84 @@
...
+
+    c = Function(Vc)
+    c.interpolate(SpatialCoordinate(cmesh)**2)
+    expect = assemble(c*dx)
+
+    actual = project(c, Vf)
+
+    actual = assemble(actual*dx)
+
+    assert numpy.allclose(expect, actual)
```
wence-: This tests non-nested projection. Should also test projecting from fine to coarse.

firedrake/supermeshing.py
```diff
@@ -0,0 +1,432 @@
...
+from pyop2.sparsity import get_preallocation
+from pyop2.compilation import load
+from pyop2.mpi import COMM_SELF
+from pyop2.utils import get_petsc_dir
+
+
+__all__ = ["assemble_mixed_mass_matrix"]
+
+
+class BlockMatrix(object):
```
wence-: Used to provide `matvec` for vector-/tensor-valued spaces, rather than assembling the bigger operator.

firedrake/supermeshing.py
```diff
@@ -0,0 +1,432 @@
...
+    """
+
+    if len(V_A) > 1 or len(V_B) > 1:
+        raise NotImplementedError("Sorry, only implemented for non-mixed spaces")
+
+    if V_A.ufl_element().mapping() != "identity" or V_B.ufl_element().mapping() != "identity":
+        msg = """
+Sorry, only implemented for affine maps for now. To do non-affine, we'd need to
+import much more of the assembly engine of UFL/TSFC/etc to do the assembly on
+each supermesh cell.
```
wence-: Argues for composition of tsfc kernels in a better way, plus dual evaluation in finat, etc...

firedrake/supermeshing.py
```diff
@@ -0,0 +1,432 @@
...
+
+    # for cell_A in range(mesh_A.num_cells()):
+    #     print("likely(%s) = %s" % (cell_A, likely(cell_A)))
+
+    # Preallocate sparsity pattern for mixed mass matrix from likely() function:
+    # For each cell_A, find dofs_A.
+    #   For each cell_B in likely(cell_B),
+    #     Find dofs_B.
+    #     For dof_B in dofs_B:
+    #         nnz[dof_B] += len(dofs_A)
```
wence-: Could redo these comments to describe algorithmic steps, rather than pseudo-code?

firedrake/supermeshing.py
```diff
@@ -0,0 +1,432 @@
...
+    if V_A.value_size > 1:
+        V_A = firedrake.FunctionSpace(mesh_A, V_A.ufl_element().sub_elements()[0])
+    if V_B.value_size > 1:
+        V_B = firedrake.FunctionSpace(mesh_B, V_B.ufl_element().sub_elements()[0])
+
+    assert V_A.value_size == 1
+    assert V_B.value_size == 1
+
+    # for cell_A in range(mesh_A.num_cells()):
+    #     print("likely(%s) = %s" % (cell_A, likely(cell_A)))
```
wence-: remove.

firedrake/supermeshing.py
```diff
@@ -0,0 +1,432 @@
...
+    #
+    mat = PETSc.Mat().create(comm=mesh_A.comm)
+    mat.setType(PETSc.Mat.Type.AIJ)
+    rsizes = tuple(n * rdim for n in nrows)
+    csizes = tuple(c * cdim for c in ncols)
+    mat.setSizes(size=(rsizes, csizes),
+                 bsize=(rdim, cdim))
+    mat.setPreallocationNNZ((dnnz, onnz))
+    mat.setLGMap(rmap=rset.lgmap, cmap=cset.lgmap)
+    # TODO: Boundary conditions not handled.
```
wence-: If we wanted to assemble with bcs, would need to fix a few more things.

firedrake/supermeshing.py
```diff
@@ -0,0 +1,432 @@
...
+    M_SS = M_SS.M.handle[:, :]
+
+    node_locations_A = utils.physical_node_locations(V_S_A).dat.data_ro_with_halos
+    node_locations_B = utils.physical_node_locations(V_S_B).dat.data_ro_with_halos
+    num_nodes_A = node_locations_A.shape[0]
+    num_nodes_B = node_locations_B.shape[0]
+
+    to_reference_kernel = to_reference_coordinates(mesh_A.coordinates.ufl_element())
+
+    supermesh_kernel_str = """
```
wence-: Ugh!

firedrake/supermeshing.py
```diff
@@ -0,0 +1,432 @@
...
+    #include "libsupermesh-c.h"
+    #include <petsc.h>
+    %(to_reference)s
+    %(evaluate_S)s
+    %(evaluate_A)s
+    %(evaluate_B)s
+
+    #define PrintInfo(...) do { if (PetscLogPrintInfo) printf(__VA_ARGS__); } while (0)
```
wence-: Can turn on "debug" output by running with `-info` on command line.

firedrake/supermeshing.py
```diff
@@ -0,0 +1,432 @@
...
+    #     For cell_B in likely(cell_A):
+    #         mesh_S = supermesh(cell_A, cell_B)
+    #         if mesh_S is empty: continue
+    #         For cell_S in mesh_S:
+    #             evaluate basis functions of cell_A at dofs(A) of cell_S -> R_AS matrix
+    #             scale precomputed mass matrix to get M_SS
+    #                   (or mixed mass matrix if V_A, V_B have different finite elements)
+    #             evaluate basis functions of cell_B at dofs(B) of cell_S -> R_BS matrix
+    #             compute out = R_BS^T @ M_SS @ R_AS with dense matrix triple product
+    #             stuff out into relevant part of M_AB (given by outer(dofs_B, dofs_A))
```
wence-: Again with the comment algorithms.

firedrake/supermeshing.py
```diff
@@ -0,0 +1,432 @@
...
+    if orig_value_size == 1:
+        return mat
+    else:
+        (lrows, grows), (lcols, gcols) = mat.getSizes()
+        lrows *= orig_value_size
+        grows *= orig_value_size
+        lcols *= orig_value_size
+        gcols *= orig_value_size
+        size = ((lrows, grows), (lcols, gcols))
+        context = BlockMatrix(mat, orig_value_size)
```
wence-: Make a python matrix that is `mass \otimes I` where `I` is an appropriate identity matrix.

firedrake/projection.py
```diff
@@ -35,72 +81,136 @@ def project(v, V, bcs=None, mesh=None,
...
-    L = ufl.inner(p, v) * ufl.dx(domain=V.mesh())
-
-    # Default to 1e-8 relative tolerance
-    if solver_parameters is None:
-        solver_parameters = {'ksp_type': 'cg', 'ksp_rtol': 1e-8}
-    else:
-        solver_parameters.setdefault('ksp_type', 'cg')
-        solver_parameters.setdefault('ksp_rtol', 1e-8)
+            solver_parameters = solver_parameters.copy()
+        solver_parameters.setdefault("ksp_type", "cg")
```
wence-: To maintain compatibility, need to also say:
`solver_parameters.setdefault("ksp_rtol", 1e-8)`

firedrake/projection.py
thomasgibson: I would also like to fix #1238. One possible option is to use dynamic matrix types where you no longer need to rely on stack limits and instead, these larger matrices are heap allocated. I raised an issue that @wence-  described in #1221. This is still a topic that needs further discussion. The compiler for Slate needs to detect when heap allocation is necessary (i.e. if the matrix is large enough to surpass Eigen's stack limit, we switch to dynamic types). I'm open to suggestions for what the best way to do this is.

PR #1336, opened 2018-11-16
Static condensation preconditioners

thomasgibson:
Reworks Slate-based preconditioners and adds a more generic static condensation interface: `SCPC`. This also provides some insight into developing more preconditions using Slate, which is summarized in `SCBase`.

### DIFF
--- a/firedrake/slate/static_condensation/hybridization.py
+++ b/firedrake/slate/static_condensation/hybridization.py
@@ -1,10 +1,7 @@
-"""This module provides custom python preconditioners utilizing
-the Slate language.
-"""
 import ufl
 import numpy as np
 
-from firedrake.preconditioners import PCBase
+from firedrake.slate.static_condensation.sc_base import SCBase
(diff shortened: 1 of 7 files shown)


firedrake/slate/static_condensation/hybridization.py
```diff
@@ -1,21 +1,19 @@
...
 from firedrake.slate.slate import Tensor, AssembledVector
 from pyop2.profiling import timed_region, timed_function
 from pyop2.utils import as_tuple
 
+
 __all__ = ['HybridizationPC']
 
 
-class HybridizationPC(PCBase):
+class HybridizationPC(SCBase):
```
thomasgibson: The only reason I have not made `HybridizationPC` a subclass of `SCPC` is that I have not figured out how to transform the two-field UFL problem into a complete 3-field hybrid problem. I mentioned this in the Slack channel (general chat).

dham:
This looks fine as far as I can see, though it's not quite my native territory. Can we have manual and/or demos soon?

PR #1343, opened 2018-11-29
install script from complex branch

dham:
This merges the install script from the complex branch into master. This has two changes:

1. Complex can be installed using the default install script. It still requires a complex-enabled branch. This is designed to prevent a currently observed user-firearm-foot issue.

2. Nic's cleaner petsc options mangling code is included.

### DIFF
--- a/scripts/firedrake-install
+++ b/scripts/firedrake-install
@@ -55,7 +55,7 @@ class FiredrakeConfiguration(dict):
                            "honour_petsc_dir",
                            "show_petsc_configure_options",
                            "slepc", "slope", "packages", "honour_pythonpath",
-                           "petsc_int_type", "cache_dir"]
+                           "petsc_int_type", "cache_dir", "complex"]
 
 
 def deep_update(this, that):
@@ -244,6 +244,8 @@ honoured.""",
     parser.add_argument("--cache-dir", type=str,
                         action="store",
                         help="Directory to use for disk caches of compiled code (default is the .cache subdirectory of the Firedrake installation).")
+    parser.add_argument("--complex", action="store_true",
+                        help="Installs the complex version of Firedrake; rebuilds PETSc in complex mode.")
     parser.add_argument("--documentation-dependencies", action="store_true",
                         help="Install the dependencies required to build the documentation")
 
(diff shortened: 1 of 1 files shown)


scripts/firedrake-install
```diff
@@ -407,30 +422,33 @@ If you really want to use your own Python packages, please run again with the
 
 petsc_opts = "--download-eigen=%s/src/eigen-3.3.3.tgz " % firedrake_env
 
-petsc_opts += "--with-fortran-bindings=0 "
+full_opts = '--download-chaco --download-metis --download-parmetis \
+--download-scalapack --download-hypre --download-mumps --with-zlib --download-netcdf --download-pnetcdf --download-hdf5 --download-exodusii --with-fortran-bindings=0'
```
wence-: Prefer set of full options here? that way adding is easier.

Do we want to add more additional LU solver types (e.g. superlue_dist, pastix, suitesparse [for umfpack]) by default?

scripts/firedrake-install
```diff
@@ -407,30 +422,33 @@ If you really want to use your own Python packages, please run again with the
...
+if options["petsc_int_type"] == 'int64':
+    split_opts -= {'--download-chaco', '--download-scalapack', '--download-mumps'}
+    split_opts |= {'--with-64-bit-indices'}
+if options["complex"]:
+    split_opts -= {'--download-hypre'}
+    split_opts |= {'--with-scalar-type=complex'}
 
 if "PETSC_CONFIGURE_OPTIONS" not in os.environ:
-    os.environ["PETSC_CONFIGURE_OPTIONS"] = petsc_opts
+    os.environ["PETSC_CONFIGURE_OPTIONS"] = " ".join(split_opts)
```
wence-: Although we say this in `--help`, it's somehow not matched up with everything else (where you pass stuff as an option). Do we want to change `firedrake-install` such that you say `firedrake-install --additional-petsc-options ...` instead of setting an environment variable?

scripts/firedrake-install
```diff
@@ -1113,6 +1145,8 @@ if mode == "install":
     else:
         git_clone("git+https://github.com/firedrakeproject/firedrake.git")
 
+    write_configuration(config)
```
wence-: We do this early because now firedrake/setup.py depends on being able to import and read `firedrake_configuration`.

scripts/firedrake-install
wence-: Finally, I added a `--show-petsc-configure-options` command to `firedrake-install` a while back, but it doesn't exit early (so we can't use it to show what you need to pass to petsc's configure). Should we do that properly while we're messing around here?

Should we also be explicit about all the options that petsc's setup.py puts in? https://github.com/firedrakeproject/petsc/blob/firedrake/setup.py#L94

dham: Is there a clean way to extract the options from petsc's setup.py? I don't particularly just want to copy the options into Firedrake...

wence-: you can run `config` with dry_run=True and it prints them out and we could get that.

OTOH, if we do that we should go ahead I think and manage the petsc install ourselves.

scripts/firedrake-install
```diff
@@ -407,30 +422,33 @@ If you really want to use your own Python packages, please run again with the
 
 petsc_opts = "--download-eigen=%s/src/eigen-3.3.3.tgz " % firedrake_env
```
thomasgibson: This gets ignored completely. Not that it makes a huge difference right now since there are still issues with using Eigen and complex-mode Firedrake together. Eigen not installing when running complex-mode is one of them.

PR #1347, opened 2018-12-07
patch multiplicative test

florianwechsung:
added test that checks that multiplicative star relaxation gives the same as sor on cg1 function spaces


wence-:
Thanks!

PR #1348, opened 2018-12-07
Apply boundary conditions on matrices via local to global maps

wence-:
Not yet ready to go (needs some tidying of the caching, and checking that lazy matrix assembly works).

But see the exciting code reduction!


firedrake/functionspacedata.py
```diff
@@ -454,134 +453,71 @@ def boundary_nodes(self, V, sub_domain, method):
...
+                nodes.append(bc.nodes * V.value_size + bc.function_space().component)
+            elif unblocked:
+                tmp = bc.nodes * V.value_size
+                for i in range(V.value_size):
+                    nodes.append(tmp + i)
+            else:
+                nodes.append(bc.nodes)
+        nodes = numpy.unique(numpy.concatenate(nodes))
+        indices[nodes] = -1
+        return PETSc.LGMap().create(indices, bsize=bsize, comm=lgmap.comm)
```
wence-: It's so simple, we just unroll the indices for the L2G map and set the boundary ones negative. This works transparently in extruded and everywhere else because bc.nodes reports the boundary nodes for all nodes in the dof set.

wence-:
I will wait until after the loopy changes. I need to push some code through PETSc.

wence-:
This will come via a new PR once #1408 lands.

PR #1352, opened 2018-12-17
Raise error if incorrectly invoking a Python PC

thomasgibson:
Small QOL change that will provide the user with an error message when trying to use a Python-based PC without first specifying "mat_type" or "pmat_type" to be "matfree". This is preferrable over what happens now (user gets a seg-fault error!). If you feel there is a better way to do this, please let me know!

I also noticed that some doc-strings were missing in one of the static condensation PCs.

### DIFF
--- a/demos/ma-demo/ma-demo.py.rst
+++ b/demos/ma-demo/ma-demo.py.rst
@@ -173,9 +173,9 @@ Finally, we'd like to see some output to check things are working, and
 to limit the KSP solver to 20 iterations. ::
 
   #
-     "ksp_monitor": True,
+     "ksp_monitor": None,
      "ksp_max_it": 20,
-     "snes_monitor": True
+     "snes_monitor": None
      }
 
 We then put all of these options into the iterative solver, ::
--- a/demos/matrix_free/navier_stokes.py.rst
+++ b/demos/matrix_free/navier_stokes.py.rst
@@ -53,7 +53,7 @@ in a ``try/except`` block. ::
 
   try:
       solve(F == 0, up, bcs=bcs, nullspace=nullspace,
-            solver_parameters={"snes_monitor": True,
+            solver_parameters={"snes_monitor": None,
                                "ksp_type": "gmres",
                                "mat_type": "aij",
                                "pc_type": "lu",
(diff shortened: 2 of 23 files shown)


thomasgibson:
Good call, you're absolutely right. I'll have a think about it and update this.

wence-:
How about this:

A PC can say whether it needs a Python amat or pmat (defaulting to False and True) respectively. This is checked in the `setUp` method of PCBase, which calls the user-provided initialize/update functions. This moves the checking and claim of what we need to the relevant PC.

```
diff --git a/firedrake/preconditioners/assembled.py b/firedrake/preconditioners/assembled.py
index c0e63b58..468fa09f 100644
--- a/firedrake/preconditioners/assembled.py
+++ b/firedrake/preconditioners/assembled.py
@@ -18,6 +18,8 @@ class AssembledPC(PCBase):
 
     _prefix = "assembled_"
 
+    needs_python_pmat = False
+
     def initialize(self, pc):
         from firedrake.assemble import allocate_matrix, create_assembly_callable
 
diff --git a/firedrake/preconditioners/base.py b/firedrake/preconditioners/base.py
index cbfc0531..2d4ae81d 100644
--- a/firedrake/preconditioners/base.py
+++ b/firedrake/preconditioners/base.py
@@ -24,6 +24,14 @@ class PCBase(object, metaclass=abc.ABCMeta):
         self.initialized = False
         super(PCBase, self).__init__()
 
+    needs_python_amat = False
+    """Set this to True if the A matrix needs to be Python (matfree)"""
+
+    needs_python_pmat = True
+    """Set this to False if the P matrix needs to be Python (matfree).
+
+    If the preconditioner also works with assembled matrices, then use False here."""
+
     @abc.abstractmethod
     def update(self, pc):
         """Update any state in this preconditioner."""
@@ -39,6 +47,14 @@ class PCBase(object, metaclass=abc.ABCMeta):
 
         Subclasses should probably not override this and instead
         implement :meth:`update` and :meth:`initialize`."""
+        A, P = pc.getOperators()
+        Atype = A.getType()
+        Ptype = P.getType()
+        if self.needs_python_amat and Atype != PETSc.Mat.Type.PYTHON:
+            raise ValueError("PC needs amat to have type python, but it is %s" % Atype)
+        if self.needs_python_pmat and Ptype != PETSc.Mat.Type.PYTHON:
+            raise ValueError("PC needs pmat to have type python, but it is %s" % Atype)
+
         if self.initialized:
             self.update(pc)
         else:
diff --git a/firedrake/preconditioners/patch.py b/firedrake/preconditioners/patch.py
index b34ae434..5fc38ad2 100644
--- a/firedrake/preconditioners/patch.py
+++ b/firedrake/preconditioners/patch.py
@@ -249,6 +249,7 @@ class PlaneSmoother(object):
 
 
 class PatchPC(PCBase):
+    needs_python_pmat = False
     def initialize(self, pc):
         A, P = pc.getOperators()
         if P.getType() == "python":
```

thomasgibson:
Great suggestion. I'll give this a try!

thomasgibson:
Sorry for the late update. I've gone ahead and modified the base class for all Python preconditioners to handle the appropriate checks. This should address all the mentioned concerns. If I made a mistake anywhere, please let me know!

PR #1353, opened 2018-12-20
When coarsening a SNESContext, coarsen its nullspaces too.

pefarrell:
At the moment, when using geometric multigrid the fine grid knows about the nullspace, but coarser grids don't. With this patch, the nullspaces are coarsened when a `SNESContext` is coarsened, and the coarsened nullspaces are applied to the coarsened Jacobians.

### DIFF
--- a/firedrake/mg/ufl_utils.py
+++ b/firedrake/mg/ufl_utils.py
@@ -216,7 +216,9 @@ def coarsen_nlvp(problem, self, coefficient_mapping=None):
 @coarsen.register(firedrake.VectorSpaceBasis)
 def coarsen_vectorspacebasis(basis, self, coefficient_mapping=None):
     coarse_vecs = [self(vec, self, coefficient_mapping=coefficient_mapping) for vec in basis._vecs]
-    return firedrake.VectorSpaceBasis(coarse_vecs, constant=basis._constant)
+    vsb = firedrake.VectorSpaceBasis(coarse_vecs, constant=basis._constant)
+    vsb.orthonormalize()
+    return vsb
 
 
 @coarsen.register(firedrake.MixedVectorSpaceBasis)
@@ -227,11 +229,10 @@ def coarsen_mixedvectorspacebasis(mspbasis, self, coefficient_mapping=None):
     for basis in mspbasis._bases:
         if isinstance(basis, firedrake.VectorSpaceBasis):
             coarse_bases.append(self(basis, self, coefficient_mapping=coefficient_mapping))
-            continue
-        if basis.index is not None:
+        elif basis.index is not None:
             coarse_bases.append(coarse_V.sub(basis.index))
-            continue
-        raise RuntimeError("MixedVectorSpaceBasis can only contain vector space bases or indexed function spaces")
+        else:
+            raise RuntimeError("MixedVectorSpaceBasis can only contain vector space bases or indexed function spaces")
 
     return firedrake.MixedVectorSpaceBasis(coarse_V, coarse_bases)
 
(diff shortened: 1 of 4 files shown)


wence-:
This looks plausible. Is it easy to add a test that this does the right thing?

pefarrell:
I tested it by running a problem with multigrid and a nullspace under gdb, setting a breakpoint on `MatNullSpaceRemove`, and verifying that it was indeed called in the coarse grid solve. I'm not sure if that is automatable in the test suite. Maybe with `pygdb2`, https://pypi.org/project/pygdb2/ ?

pefarrell:
Question: should we `orthonormalize` the basis first? I don't see any guarantee that the coarsening of an orthonormal basis yields an orthonormal coarse basis.

wence-:
> Question: should we orthonormalize the basis first? I don't see any guarantee that the coarsening of an orthonormal basis yields an orthonormal coarse basis.

I think we should orthonormalize. I think a fine grid basis which is exactly representable on the coarse grid will be orthonormal, but otherwise not.

wence-:
> I tested it by running a problem with multigrid and a nullspace under gdb, setting a breakpoint on MatNullSpaceRemove, and verifying that it was indeed called in the coarse grid solve. I'm not sure if that is automatable in the test suite. Maybe with pygdb2, https://pypi.org/project/pygdb2/ ?

I had thought "here is a simple problem where if I don't coarsen the null space I get the wrong convergence" as a test. Is that easy?

firedrake/mg/ufl_utils.py
```diff
@@ -213,6 +213,29 @@ def coarsen_nlvp(problem, self, coefficient_mapping=None):
...
+    coarse_bases = []
+
+    for basis in mspbasis._bases:
+        if isinstance(basis, firedrake.VectorSpaceBasis):
+            coarse_bases.append(self(basis, self, coefficient_mapping=coefficient_mapping))
+            continue
+        if basis.index is not None:
+            coarse_bases.append(coarse_V.sub(basis.index))
+            continue
+        raise RuntimeError("MixedVectorSpaceBasis can only contain vector space bases or indexed function spaces")
```
wence-: Style, prefer:
```
if ...:
   ...
elif ...:
   ...
else:
    raise RuntimeError(...)
```

firedrake/mg/ufl_utils.py
```diff
@@ -241,6 +264,17 @@ def coarsen_snescontext(context, self, coefficient_mapping=None):
...
+
+    if context._nullspace:
+        coarse._nullspace = self(context._nullspace, self, coefficient_mapping=coefficient_mapping)
+        coarse.set_nullspace(coarse._nullspace, problem.J.arguments()[0].function_space()._ises, transpose=False, near=False)
+    if context._nullspace_T:
+        coarse._nullspace_T = self(context._nullspace_T, self, coefficient_mapping=coefficient_mapping)
+        coarse.set_nullspace(coarse._nullspace_T, problem.J.arguments()[0].function_space()._ises, transpose=True, near=False)
+    if context._near_nullspace:
+        coarse._near_nullspace = self(context._near_nullspace, self, coefficient_mapping=coefficient_mapping)
+        coarse.set_nullspace(coarse._near_nullspace, problem.J.arguments()[0].function_space()._ises, transpose=False, near=True)
```
wence-: `coarsen` handles `None` fine, as does set_nullspace, so perhaps:

```python
ises = problem.J.arguments()[0].function_space()._ises
kwargs = {"_nullspace": {"transpose": False, "near": False},
                 "_nullspace_T": {"transpose": True, "near": False},
                 "_near_nullspace": {"transpose": False, "near": True}}
for attr, kwarg in kwargs.items():
   setattr(coarse, attr) = self(getattr(context, attr), self, coefficient_mapping=coefficient_mapping)
   coarse.set_nullspace(getattr(context, attr), ises, **kwarg)
```

If that is too magic, perhaps at least pull the duplicated access of the is access into an assignment to a local variable.

firedrake/solving_utils.py
```diff
@@ -296,11 +300,19 @@ def form_jacobian(snes, X, J, P):
...
             assert P.handle == ctx._pjac.petscmat.handle
             ctx._assemble_pjac()
             ctx._pjac.force_evaluation()
 
+        if ctx._nullspace:
+            ctx.set_nullspace(ctx._nullspace, problem.J.arguments()[0].function_space()._ises, transpose=False, near=False)
+        if ctx._nullspace_T:
+            ctx.set_nullspace(ctx._nullspace_T, problem.J.arguments()[0].function_space()._ises, transpose=True, near=False)
+        if ctx._near_nullspace:
+            ctx.set_nullspace(ctx._near_nullspace, problem.J.arguments()[0].function_space()._ises, transpose=False, near=True)
```
wence-: `set_nullspace` handles `None` fine, so no need for the condition.

firedrake/mg/ufl_utils.py
pefarrell: OK, thanks!

firedrake/solving_utils.py
pefarrell: Done.

pefarrell:
I've added a test that fails if the nullspace is not coarsened, but passes on this branch.

pefarrell:
Yep, Jenkins has passed. OK to merge?

wence-:
👍

PR #1355, opened 2019-01-10
Slate patch: fix assignments for local vector temporaries

thomasgibson:
This small patch fixes a problem where some local temporaries for firedrake `Coefficient`s where not being assigned in the correct outer dof-extent loops in Slate-generated kernels.

### DIFF
--- a/tests/slate/test_variational_prb.py
+++ b/tests/slate/test_variational_prb.py
@@ -1,0 +1,88 @@
+import pytest
+import numpy as np
+from firedrake import *
+
+
+@pytest.mark.parametrize('degree', range(1, 4))
+def test_lvp_equiv_hdg(degree):
+    """Runs an HDG problem and checks that passing
+    a Slate-defined problem into a variational problem
+    produces the same result for the traces as solving
+    the whole system using built-in and tested solvers
+    and preconditioners.
+    """
+
+    mesh = UnitSquareMesh(5, 5)
+    x = SpatialCoordinate(mesh)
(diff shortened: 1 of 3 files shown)


wence-:
Test?

thomasgibson:
Green lights on everything. I think it's ready to go if you're happy with it.

tests/slate/test_variational_prb.py
```diff
@@ -0,0 +1,88 @@
+import pytest
+import numpy as np
+from firedrake import *
+
+
+@pytest.mark.parametrize('degree', range(1, 4))
```
wence-: Maybe only `[1, 2]`? You don't expect the degree to make a difference to whether or not this bug exhibits, right?

tests/slate/test_variational_prb.py
```diff
@@ -0,0 +1,88 @@
...
+    lvp = LinearVariationalProblem(S, r_lambda, t)
+    solver = LinearVariationalSolver(lvp, solver_parameters={'ksp_type': 'preonly',
+                                                             'pc_type': 'lu',
+                                                             'pc_factor_mat_solver_type': 'mumps'})
+    solver.solve()
+
+    assert np.allclose(uhat_ref.dat.data, t.dat.data, rtol=1.E-12)
+
+
+if __name__ == '__main__':
```
wence-: No need to have these bits. I should remove them from the entire test suite so people stop cargo-culting them :)

PR #1357, opened 2019-01-12
Smaclachlan/embedded error message

ScottMacLachlan:
Adding a check when inject/restrict/prolong are called to throw a better error message when no native operators are available.

Note: this duplicates "native" as is also set in embedded.py.  Should this list be stored in just one place?

### DIFF
--- a/firedrake/mg/interface.py
+++ b/firedrake/mg/interface.py
@@ -22,7 +22,7 @@ def check_arguments(coarse, fine):
         raise ValueError("Can't transfer between functions from different hierarchies")
     if coarse.ufl_shape != fine.ufl_shape:
         raise ValueError("Mismatching function space shapes")
-    if coarse.ufl_element().family() not in native or fine.ufl_element().family() not in native:
+    if (not isinstance(coarse.ufl_element(),firedrake.MixedElement) and not isinstance(fine.ufl_element(),firedrake.MixedElement)) and (coarse.ufl_element().family() not in native or fine.ufl_element().family() not in native):
         raise ValueError("Function space doesn't allow native transfer operators, consider using EmbeddedDGTransfer instead")
 
 
--- a/firedrake/mg/interface.py
+++ b/firedrake/mg/interface.py
@@ -22,7 +22,7 @@ def check_arguments(coarse, fine):
         raise ValueError("Can't transfer between functions from different hierarchies")
     if coarse.ufl_shape != fine.ufl_shape:
         raise ValueError("Mismatching function space shapes")
-    if coarse.ufl_element().family() not in native:
+    if coarse.ufl_element().family() not in native or fine.ufl_element().family() not in native:
         raise ValueError("Function space doesn't allow native transfer operators, consider using EmbeddedProlongation instead")
 
 def prolong(coarse, fine):
--- a/firedrake/mg/interface.py
+++ b/firedrake/mg/interface.py
@@ -7,7 +7,9 @@ from . import kernels
 
 __all__ = ["prolong", "restrict", "inject"]
 
+
 native = frozenset(["Lagrange", "Discontinuous Lagrange", "Real", "Q", "DQ"])
+
 
 def check_arguments(coarse, fine):
     cfs = coarse.function_space()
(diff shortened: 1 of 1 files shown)


firedrake/mg/interface.py
```diff
@@ -8,6 +8,9 @@
 __all__ = ["prolong", "restrict", "inject"]
 
 
+native = frozenset(["Lagrange", "Discontinuous Lagrange", "Real", "Q", "DQ"])
```
wence-: I think I would prefer a single source of truth for this support. The more natural place is for it to be in embedded.py I think. To avoid circular import problems, you can do `from .embedded import native` at the top of the `check_arguments` function.

firedrake/mg/interface.py
```diff
@@ -21,6 +24,8 @@ def check_arguments(coarse, fine):
         raise ValueError("Can't transfer between functions from different hierarchies")
     if coarse.ufl_shape != fine.ufl_shape:
         raise ValueError("Mismatching function space shapes")
+    if coarse.ufl_element().family() not in native or fine.ufl_element().family() not in native:
+        raise ValueError("Function space doesn't allow native transfer operators, consider using EmbeddedProlongation instead")
```
wence-: I think the relevant thing is `EmbeddedDGTransfer`?

firedrake/mg/interface.py
ScottMacLachlan: Fixed in latest commit

firedrake/mg/interface.py
ScottMacLachlan: Good catch - fixed in latest commit

ScottMacLachlan:
Looking at the checks, this seemed to break on a mixed function space, which I've been able to reproduce for a simple Stokes test with monolithic MG, e.g., as in demos/multigrid/geometric_multigrid.py.  Do we need a more subtle hook to check within a mixed space?

ScottMacLachlan:
I think isinstance was correct - at least, I was able to make that work for me.  The logic is now that neither fine nor coarse function space is of mixed type and that either fine or coarse is not in native list.  That seems somewhat overly constrained, since it seems nonsensical to have one function space be mixed and one not, but I wasn't sure how else to resolve it...

ScottMacLachlan:
Looking at what failed, it seems that we don't properly catch TensorProductElements.  That makes sense from the logic (they aren't in native nor instances of MixedElements), but what's the correct behaviour here - should they also be excluded from the check (because, presumably, they'd get caught when building interpolation for one of the spaces in the tensor product?)?

wence-:
Apologies that this has taken so long. I think we want logic that is the inverse of what is implemented in EmbeddedDGTransfer (isinstance(foo, MixedElement) picks up Vector/TensorElement too, which is bad)

```python
native = ...
c_element = coarse.ufl_element()
f_element = fine.ufl_element()
def valid(element):
    return type(element) is not firedrake.MixedElement and element.family() in native

if not (valid(c_element) and valid(f_element)):
    raise ValueError("Unsupported native transfer, maybe you need EmbeddedDGTransfer?")
```
?

FWIW, you can check this locally (like the linting) with:

```
pytest tests/multigrid
```

ScottMacLachlan:
Hrm.  The above fails on  test_multiple_custom_transfer_monolithc in tests/test_custom_transfer.py (and in lots of other places...).  Digging in, I see that type(c_element) returns <class 'ufl.finiteelement.mixedelement.MixedElement'> and not firedrake.MixedElement. In contrast, isinstance(c_element,firedrake.MixedElement) returns true.

Is there a reason to use type in place of isinstance?

wence-:
Hi Scott, let me try and find some time to look at this properly on Friday (rather than guessing at problems!)

wence-:
@ScottMacLachlan Apologies for the radio silence. It turns out doing this check in a nice way was more effort than I anticipated. One needs to unpick much more of the element to figure out if it is suitable for the builtin ones. This is still on my radar, and hopefully I can patch it up at some point soon.

ScottMacLachlan:
@wence- No worries.  I'm happy to leave this on the back burner for now...

wence-:
I think this is too complicated to do cleanly for now, sorry! Prod again if it continues to bite.

PR #1360, opened 2019-01-22
Form bcs

ksagiyam:
A cleaner FormBC implementation


firedrake/bcs.py
```diff
@@ -15,21 +15,14 @@
 import firedrake.utils as utils
 
 
-__all__ = ['DirichletBC', 'homogenize']
+__all__ = ['DirichletBC', 'homogenize', 'FormBC']
 
 
-class DirichletBC(object):
-    r'''Implementation of a strong Dirichlet boundary condition.
+class BCObject(object):
```
wence-: Prefer `BCBase` or `AbstractBC` (I think in the rest of firedrake we typically use `FooBase` for these abstract base classes).

firedrake/bcs.py
```diff
@@ -295,3 +315,72 @@ def homogenize(bc):
...
+        self.J = J
+        self.Jp = Jp
+        # f (= F, J or Jp) is set right before calling "asesmble(...),"
+        # telling the function which one of residual, Jacobian, or
+        # Preconditionar is to be assembled.
+        self.f = None
+
+    @property
+    def formeq(self):
+        return self._formeq
```
wence-: You already set the public property `formeq` in the constructor, so no need for this.

firedrake/bcs.py
```diff
@@ -295,3 +315,72 @@ def homogenize(bc):
...
+
+    @formeq.setter
+    def formeq(self, v):
```
wence-: Should it be allowed to change these in the public interface?

firedrake/bcs.py
```diff
@@ -295,3 +315,72 @@ def homogenize(bc):
...
+    def J(self, v):
+        self._J = v
+
+    @property
+    def Jp(self):
+        return self._Jp
+
+    @Jp.setter
+    def Jp(self, v):
+        self._Jp = v
```
wence-: I suspect all of these setters should go, and we just use the public constructor to make things.

firedrake/solving.py
```diff
@@ -215,6 +203,13 @@ def _la_solve(A, x, b, **kwargs):
 
     bcs, solver_parameters, nullspace, nullspace_T, near_nullspace, \
         options_prefix = _extract_linear_solver_args(A, x, b, **kwargs)
+
+    from firedrake.bcs import FormBC
+    if bcs is not None:
+        for bc in bcs:
+            if isinstance(bc, FormBC) and not (isinstance(bc.formeq.lhs, ufl.Form) and isinstance(bc.formeq.rhs, ufl.Form)):
```
wence-: Prefer to check this in the `FormBC` constructor. Then you have to sprinkle fewer `isinstance` checks here. You "know" the value is good.

firedrake/solving.py
```diff
@@ -319,14 +314,14 @@ def _extract_args(*args, **kwargs):
...
     if bcs is None:
         return ()
     try:
         bcs = tuple(bcs)
     except TypeError:
         bcs = (bcs,)
     for bc in bcs:
-        if not isinstance(bc, DirichletBC):
-            raise TypeError("Provided boundary condition is a '%s', not a DirichletBC" % type(bc).__name__)
+        if not isinstance(bc, DirichletBC) and not isinstance(bc, FormBC):
```
wence-: AKA `if not isinstance(bc, BCObject)` (or whatever if we rename it).

firedrake/solving_utils.py
```diff
@@ -337,7 +360,13 @@ def compute_operators(ksp, J, P):
     @cached_property
     def _jac(self):
         from firedrake.assemble import allocate_matrix
-        return allocate_matrix(self.J, bcs=self._problem.bcs,
+        from firedrake.bcs import FormBC
+        if self._problem.bcs is not None:
+            for bc in self._problem.bcs:
+                if isinstance(bc, FormBC):
+                    bc.f = bc.J
```
wence-: Why do we have to copy from one property of the `bc` object to another?

firedrake/variational_solver.py
```diff
@@ -9,20 +9,25 @@
...
+__all__ = ["VariationalProblem",
+           "LinearVariationalProblem",
            "LinearVariationalSolver",
            "NonlinearVariationalProblem",
            "NonlinearVariationalSolver"]
 
 
-class NonlinearVariationalProblem(object):
-    r"""Nonlinear variational problem F(u; v) = 0."""
+class VariationalProblem(object):
```
wence-: Does it make sense to make a single `VariationalProblem` class, and then refactor the `NonlinearVariationalProblem` and `LinearVariationalProblem` into functional interfaces that just provide the correct arguments? (reducing the number of different classes).

Places where this might end up breaking things are in some of the multigrid coarsening/refining where we possibly do different things based on LVP vs NLVP.

tests/formbc/test_poisson_FormBC.py
```diff
@@ -0,0 +1,84 @@
...
+
+    solve(a == L, u, bcs=[bc1, bc2, bc3, bc4], solver_parameters={'ksp_type': 'gmres', 'ksp_atol': 1e-12, 'ksp_rtol': 1e-20, 'ksp_divtol': 1e8})
+
+    f.interpolate(cos(x * pi * 2) * cos(y * pi * 2))
+    err = sqrt(assemble(dot(u - f, u - f) * dx))
+
+    assert(err < 0.05)
+
+
+if __name__ == "__main__":
```
wence-: Thomas has just diligently gone through and removed allow of these bits from the test files. There's no need to have them there (you weren't to know this!).

tests/formbc/test_poisson_FormBC.py
```diff
@@ -0,0 +1,84 @@
...
+    bc4 = DirichletBC(V, cos(2 * pi * x), 4)
+
+    u = Function(V)
+
+    solve(a == L, u, bcs=[bc1, bc2, bc3, bc4], solver_parameters={'ksp_type': 'gmres', 'ksp_atol': 1e-12, 'ksp_rtol': 1e-20, 'ksp_divtol': 1e8})
+
+    f.interpolate(cos(x * pi * 2) * cos(y * pi * 2))
+    err = sqrt(assemble(dot(u - f, u - f) * dx))
+
+    assert(err < 0.05)
```
wence-: Given that we are solving an equation on the boundary now, and this is not an exact solution. Should we perform a convergence test to check we're getting the correct order?

firedrake/bcs.py
```diff
@@ -295,3 +315,72 @@ def homogenize(bc):
...
+    def __init__(self, V, eq, sub_domain, J=None, Jp=None, method="topological"):
+        super().__init__(V, sub_domain, method="topological")
+        self.formeq = eq
+        self.F = None
+        self.J = J
+        self.Jp = Jp
+        # f (= F, J or Jp) is set right before calling "asesmble(...),"
+        # telling the function which one of residual, Jacobian, or
+        # Preconditionar is to be assembled.
+        self.f = None
```
wence-: This feels like bad design, because it is stateful, and likely to be the cause of difficult to find bugs.

firedrake/assemble.py
```diff
@@ -214,11 +216,23 @@ def _assemble(f, tensor=None, bcs=None, form_compiler_parameters=None,
         kernels = tsfc_interface.compile_form(f, "form", parameters=form_compiler_parameters, inverse=inverse)
         integral_types = [integral.integral_type() for integral in f.integrals()]
 
+        if bcs is not None:
+            for bc in bcs:
+                if isinstance(bc, FormBC):
+                    integral_types += [integral.integral_type() for integral in bc.f.integrals()]
```
wence-: Could we have a `FormBC` provide an `integrals()` method which returns the integrals, with a `DirichletBC` returning an empty list? Then we don't need this instance check.

firedrake/assemble.py
```diff
@@ -214,11 +216,23 @@ def _assemble(f, tensor=None, bcs=None, form_compiler_parameters=None,
...
     is_mat = rank == 2
     is_vec = rank == 1
 
+    # FormBCs requre recursive calls of _assemble(...), and kernels keep
+    # being appended to a sigle list ("loops") at all levels of recursion.
+    # The first level call of _assemble(...) is identified by "loops is None."
+    first_call = False
+    if loops is None:
+        loops = []
+        first_call = True
```
wence-: I think we should think about how we can refactor this loop collection so that it is less messy.

Thinking about it, since it is an internal implementation detail, we could always be in "collecting loops mode" in here. And then the top-level `assemble` could collect the loops and immediately evaluate them.

We could then make this function `yield` loops (i.e. become a generator) and the recursive call to assemble collecting the loops can be replaced by `yield from _assemble(...)`

wence-: This would also simplify the code below, because the `thunk` call would just become the list of loops we need to evaluate (I think...)

firedrake/bcs.py
```diff
@@ -295,3 +315,72 @@ def homogenize(bc):
...
+        the top and bottom boundaries of the extruded mesh respectively.
+    :arg method: the method for determining boundary nodes. The default is
+        "topological", indicating that nodes topologically associated with a
+        boundary facet will be included. The alternative value is "geometric",
+        which indicates that nodes associated with basis functions which do not
+        vanish on the boundary will be included. This can be used to impose
+        strong boundary conditions on DG spaces, or no-slip conditions on HDiv spaces.
+    '''
+
+    def __init__(self, V, eq, sub_domain, J=None, Jp=None, method="topological"):
```
wence-: We should be able to determine the `V` from `eq`, no? Should we therefore remove that argument?

docs/source/boundary_conditions.rst
```diff
@@ -310,3 +310,190 @@ solved with boundary conditions as follows:
...
+   a) ``F`` is assembled using unmodified indirection maps taking no
+      account of the boundary conditions. This results in an assembled
+      residual which is correct on the non-boundary condition nodes but
+      contains spurious values in the boundary condition entries.
+
+   b) The entries of ``F`` corresponding to boundary condition nodes
+      are set to zero.
+
+   c) ``F_{Form}`` is assembled to populate the entries corresponding to 
+      the form boundary condition nodes that have been zeroed in b).
```
wence-: OK, so the big change to assembly is that split assembly of operators and residuals that are subject to equational boundary conditions into two pieces. We do this with `DirichletBC` too, except that the subsequent "assembly" that we do to fill in the bits is trivial.

Does this suggest an implementation where the presence of BCs in a variational problem/assembly turns (internally) into a 2-tuple of `(before_action, after_action)`? This would allow us to redo this code to avoid the action-at-a-distance of setting `bc.f` in FormBCs.

firedrake/variational_solver.py
```diff
@@ -34,43 +39,128 @@ def __init__(self, F, u, bcs=None, J=None,
...
+                raise ValueError("Provided preconditioner is not a bilinear form")
+            self.is_linear = False
+
+        # Boundary (FormBC)
+        if self.bcs is not None:
+            for bc in self.bcs:
+                if isinstance(bc, FormBC):
+                    # linear
+                    if isinstance(bc.formeq.lhs, ufl.Form) and isinstance(bc.formeq.rhs, ufl.Form):
+                        bc.J = bc.formeq.lhs
```
wence-: Careful, this smells like bad code to me, because you're updating the `FormBC` object in place. In particular, I think this means you can't use the same `FormBC` object with two different `VariationalProblems`, since the `bc.J`, etc... will be set by the most recently constructed one.

firedrake/bcs.py
ksagiyam: You are right.  formeq should not be changed.

tests/formbc/test_poisson_FormBC.py
ksagiyam: This is actually still the exact solution.  We just apply boundary conditions differently: u=g (DirichletBC) or (u-g)*ds=0 (FormBC).

wence-: Well, the discrete solution converges to the exact solution at second order. So do we want convergence to make sure that we do get this correct order of convergence?

firedrake/assemble.py
```diff
@@ -400,13 +417,14 @@ def vec(testmap, i):
     # In collecting loops mode, we collect the loops, and assume the
     # boundary conditions provided are the ones we want.  It therefore
     # is only used inside residual and jacobian assembly.
-    loops = []
 
     def thunk(bcs):
         if collect_loops:
-            loops.append(zero_tensor)
+            if first_call:
```
dham: This becomes:
```
if zero_tensor:
    yield zero_tensor_parloop
```

firedrake/assemble.py
```diff
@@ -546,53 +571,100 @@ def get_map(x):
...
+                            for j in range(shape[1]):
+                                # Set diagonal entries on bc nodes to 1 if the current
+                                # block is on the matrix diagonal and its index matches the
+                                # index of the function space the bc is defined on.
+                                if i != j:
+                                    continue
+                                if fs.component is None and fs.index is not None:
+                                    # Mixed, index (no ComponentFunctionSpace)
+                                    if fs.index == i:
+                                        loops.append(tensor[i, j].set_local_diagonal_entries(nodes))
```
dham: becomes:
```
yield tensor[i, j].set_local_diagonal_entries(nodes)
```

firedrake/assemble.py
ksagiyam: Thanks! I will make changes based on this suggestion.

firedrake/variational_solver.py
ksagiyam: I will move this part to the constructor, too.

docs/source/boundary_conditions.rst
ksagiyam: For now I will make an alternative change to avoid bc.f, but will come back to this idea when I have to consider more general cases where boundary equations themselves require boundary conditions (on edges, say).

PR #1363, opened 2019-01-23
Some updates to the auxiliary operator preconditioner

pefarrell:
This branch renames `ExplicitSchurPC` (options prefix `schur_`) to `AuxiliaryOperatorPC` (options prefix `aux_`).

I've also added code to set up a `SNESContext` and a `DM` on the assembled PC that underpins this. This allows for the use of the full panoply of solvers available in Firedrake on auxiliary operators supplied via this route. I have added a test in `tests/regression` that segfaults on master and runs successfully with this branch.

### DIFF
--- a/firedrake/preconditioners/assembled.py
+++ b/firedrake/preconditioners/assembled.py
@@ -6,7 +6,7 @@ from firedrake.petsc import PETSc
 from firedrake.ufl_expr import TestFunction, TrialFunction
 from firedrake.dmhooks import get_function_space, get_appctx, push_appctx, pop_appctx
 
-__all__ = ("AssembledPC", "ExplicitSchurPC")
+__all__ = ("AssembledPC", "AuxiliaryOperatorPC")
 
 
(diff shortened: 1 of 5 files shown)


firedrake/preconditioners/assembled.py
```diff
@@ -107,12 +121,12 @@ def view(self, pc, viewer=None):
...
 
 
-class ExplicitSchurPC(AssembledPC):
+class AuxiliaryOperatorPC(AssembledPC):
     """A preconditioner that builds a PC on a specified form.
     Mainly used for describing approximations to Schur complements.
     """
 
-    _prefix = "schur_"
+    _prefix = "aux_"
```
wence-: Not in this change, but I might make this prefix public, and therefore more obviously overridable, so you can say, for example:

```python
class MassPC(AuxiliaryOperatorPC):
    prefix = "mass_"
```

And use a more "natural" prefix for the configuration.

PR #1365, opened 2019-01-28
Drop support for c-string expressions

thomasgibson:
In preparation for the inevitable loo.py merge, Firedrake is now dropping support for c-string `Expression`s. The class is still around, but will now raise a `ValueError` when attempting to create/use these. Tests and demos have been updated accordingly.


thomasgibson:
According to the Jenkins build, pyadjoint needs to be updated as well.

PR #1370, opened 2019-02-04
meshes: Implement diagonal="crossed" for RectangleMesh

wence-:


wence-:
@pefarrell does this DTRT for you?

pefarrell:
Yes! I have tried it out in serial and parallel and it looks great. Thanks!

PR #1371, opened 2019-02-04
Deprecate options translation

wence-:
It makes it difficult to switch off boolean options in the natural
way:
```
  {"pc_use_foo": False}
```
Presently does not work if the default value for "pc_use_foo" is
True (since our translation removes this option from the dictionary
entirely).

We can argue about when this should change, I vote for sooner rather than later.

### DIFF
--- a/firedrake/petsc.py
+++ b/firedrake/petsc.py
@@ -193,12 +193,13 @@ class OptionsManager(object):
             for k, v in self.parameters.items():
                 key = self.options_prefix + k
                 if type(v) is bool:
-                    warning("""Firedrake will stop translating True/False options soon.\n"""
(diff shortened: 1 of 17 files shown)


firedrake/petsc.py
```diff
@@ -188,10 +188,18 @@ def set_from_options(self, petsc_obj):
     def inserted_options(self):
         """Context manager inside which the petsc options database
     contains the parameters from this object."""
+        from firedrake.logging import warning
         try:
             for k, v in self.parameters.items():
                 key = self.options_prefix + k
                 if type(v) is bool:
```
wence-: Once the deprecation period is up, we just remove the condition and this turns into:
```
for k, v in self.parameters.items():
    self.options_object[self.options_prefix + k] = v
```

PR #1376, opened 2019-02-20
Form bcs revised

ksagiyam:
Add fieldsplit capability to EquationBC and relevant examples.  Simplify assemble.py.  We are still unable to apply boundary conditions to an EquationBC itself,  so for now we have to have an ad hoc implementation of ImplicitMatrixContext.mult() and ImplicitMatrixContext.multTranspose().


ksagiyam:
move to PR #1429

PR #1377, opened 2019-02-27
Simple Dockerfile

dham:
This creates a basic dockerfile which installs Firedrake. This is initially being introduced to support pyadjoint's CI system, however we could also provide this as an alternative installation route.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,6 +1,5 @@
-# DockerFile for a firedrake + jupyter container
+# DockerFile for a firedrake
 
-# Use a jupyter notebook base image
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -0,0 +1,39 @@
+# DockerFile for a firedrake + jupyter container
```
wence-: No jupyter here?

Dockerfile
```diff
@@ -0,0 +1,39 @@
...
+
+# Use a jupyter notebook base image
```
wence-: Likewise

PR #1379, opened 2019-03-01
Test case for NaNs in Conditionals

miklos1:
To protect against regressions on firedrakeproject/tsfc#183.

Should also mention that it requires firedrakeproject/tsfc#186.


wence-:
Thanks!

PR #1383, opened 2019-03-07
Wence/feature patch facets

wence-:
Implement interior face terms for PCPatch. Will need a PETSc PR which is currently open.

### DIFF
--- a/firedrake/mesh.py
+++ b/firedrake/mesh.py
@@ -588,9 +588,14 @@ class MeshTopology(object):
                                    self._cell_numbering,
                                    self.cell_closure)
 
-        return _Facets(self, classes, kind,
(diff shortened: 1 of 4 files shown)


wence-:
Rebased and squashed all of this. Need to merge the petscp4y change and update the firedrake petsc pointer I think.

wence-:
OK, this needs https://bitbucket.org/petsc/petsc/pull-requests/1459/fix-semantics-of-patch-callback-functions/diff, hopefully that is all now.

dham:
Should I be worried about the fact that this passes tests despite the fact that the upstream change hasn't landed yet? Looks like a hole in coverage.

wence-:
No, because it had edited the branch of petsc we were testing with.

PR #1389, opened 2019-03-10
Parallel point location

cjlh:
These changes update the cell location interface to accept a list of points to locate, rather than a single point.

A new PointCloud class has been created that can be used to store a list of points and to locate those points within the distributed mesh -- the results are given as an array of (rank, cell) pairs. A function may then be evaluated at these points on the mesh using the `evaluate` function (WIP).


firedrake/pointcloud.py
```diff
@@ -0,0 +1,299 @@
...
+        # Iterate through all points. If they have not been located locally,
+        # iterate through each point request reponse to find the element.
+        # Sometimes an element can be reported as found by more than one
+        # process -- in this case, choose the process with the lower rank.
+        for rank, requested_points in local_candidates.items():
+            for j, point in enumerate(requested_points):
+                i = local_candidate_indices[rank][j]
+                curr_rank = located_elements[i][0]
+                result = recv_results[rank][j][0]
+                if (result != -1 and (curr_rank == -1 or rank < curr_rank)):
```
wence-: I guess we could arbitrarily decide to choose the max rank, rather than the min rank, then we just write:
```
located_elements[i] = (max(rank, curr_rank), result)
```
With no need for the `if`.

wence-: I think we never used the `point` here, so perhaps:
```python
for rank, result in recv_results.items():
   indices = local_candidate_indices[rank]
   found, = np.where(result != -1)
   for idx in found:
      i = indices[idx]
      loc_rank = located_elements[i, 0]
      if loc_rank == -1 or rank < loc_rank:
          located_elements[i, :] = (rank, result[idx])
```

firedrake/pointcloud.py
```diff
@@ -0,0 +1,299 @@
...
+    set to at least debug.
+    """
+    if logging.logger.isEnabledFor(logging.DEBUG):
+        PETSc.Sys.syncFlush(*args, **kwargs)
+
+
+class PointCloud(object):
+    """Store a group of points and evaluate the (cell, rank) pair corresponding their
+    location in the distributed mesh.
+    """
```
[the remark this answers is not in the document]
cjlh: I'm happy with the description you've given here if others are too -- I agree it makes more sense.

firedrake/pointcloud.py
```diff
@@ -0,0 +1,299 @@
...
+    def __init__(self, mesh, points, *args, **kwargs):
+        """Initialise the PointCloud.
+
+        :arg mesh: A mesh object.
+        :arg points: A list of points in the point cloud.
```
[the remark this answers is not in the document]
cjlh: Sounds good to me.

firedrake/pointcloud.py
```diff
@@ -0,0 +1,299 @@
...
+        """
+        self.mesh = mesh
+        self.points = np.asarray(points)
+        _, dim = points.shape
+        assert dim == mesh.geometric_dimension()
```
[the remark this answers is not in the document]
cjlh: That looks good to me.

firedrake/pointcloud.py
```diff
@@ -0,0 +1,299 @@
...
+
+        # Create lists containing the minimum and maximum bounds of each process, where
+        # the index in each list is the rank of the process.
+        min_bounds = []
+        max_bounds = []
+        arr_size = int(len(local) / 2)
+        for i in range(0, self.mesh.comm.size):
+            j = arr_size * 2 * i
+            min_bounds.append(global_[j:j + arr_size])
+            max_bounds.append(global_[j + arr_size:j + arr_size * 2])
```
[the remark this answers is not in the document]
cjlh: Thanks. I believe this should be:

```
min_bounds, max_bounds = global_.reshape(self.mesh.comm.size, 2, len(local) // 2).swapaxes(0, 1)
```

Without the calls to `flatten`, as `spatial_index.from_regions` expects a 2-dimensional array. The arrays must then also be made contiguous.

firedrake/pointcloud.py
```diff
@@ -0,0 +1,299 @@
...
+             rank.
+        """
+        # Create lists to hold send and receive communication request objects.
+        recv_reqs = []
+        send_reqs = []
+
+        for rank, recv_buffers in recv_buffers_dict.items():
+            # Create all receive requests.
+            for i in range(0, len(recv_buffers)):
+                req = self.mesh.comm.Irecv(recv_buffers[i], source=rank, tag=i)
```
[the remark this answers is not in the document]
cjlh: I think this becomes an issue when we use the same function for sending/receiving points (in this case `buffer` would be an array of n-dimensional arrays). As far as I can tell, mpi4py only allows size-1 arrays to be sent/received.

Noted on the naming -- thanks.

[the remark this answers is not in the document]
cjlh: I think this becomes an issue when we use the same function for sending/receiving points (in this case `buffer` would be an array of n-dimensional arrays). As far as I can tell, mpi4py only allows size-1 arrays to be sent/received.

Noted on the naming -- thanks.

firedrake/pointcloud.py
cjlh: I have went with your second suggestion here -- with the first, if two processes report locating the same point on different elements, would `located_elements` not then be updated with the new result regardless of which rank is higher or am I misunderstanding? Cheers.

cjlh:
Thanks for the comments @wence-. I have implemented and pushed most of the changes you suggested -- happy to discuss further.

firedrake/pointcloud.py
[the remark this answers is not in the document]
wence-: Ah. So flatten makes contiguous arrays, you can then reshape afterwards.

firedrake/pointcloud.py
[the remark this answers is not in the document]
wence-: I think we can send n-dimensional objects with mpi4py. But I could have misremembered.

firedrake/pointcloud.py
```diff
@@ -0,0 +1,299 @@
...
+    """
+    if logging.logger.isEnabledFor(logging.DEBUG):
+        PETSc.Sys.syncFlush(*args, **kwargs)
+
+
+class PointCloud(object):
+    """Store a group of points and evaluate the (cell, rank) pair corresponding their
+    location in the distributed mesh.
+    """
+    def __init__(self, mesh, points, *args, **kwargs):
```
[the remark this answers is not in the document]
dham: I think that way around makes sense.

firedrake/pointcloud.py
```diff
@@ -0,0 +1,283 @@
...
+
+class PointCloud(object):
+    """Store points for repeated location in a mesh. Facilitates lookup and evaluation
+    at these point locations.
+    """
+    def __init__(self, mesh, points, *args, **kwargs):
+        """Initialise the PointCloud.
+
+        :arg mesh: A mesh object.
+        :arg points: An array of points.
```
dham: What's a point? I think you mean an N x mesh.geometric_dimension() array of point locations.

firedrake/pointcloud.py
```diff
@@ -0,0 +1,283 @@
...
+        self.points = np.asarray(points)
+        _, dim = points.shape
+        if dim != mesh.geometric_dimension():
+            raise ValueError("Points must be %d-dimensional, (got %d)" %
+                             (mesh.geometric_dimension(), dim))
+
+        # Build spatial index of processes for point location.
+        self.processes_index = self._build_processes_spatial_index()
+
+    @cached_property
```
dham: If this is a cached property, what is the mechanism for blowing the cache if the user moves the coordinate locations on the mesh?

firedrake/pointcloud.py
```diff
@@ -0,0 +1,283 @@
...
+        6. Perform sparse communication round to return results.
+        7. Process responses to obtain final results array. If multiple processes report
+           containing the same point, choose the process with the lower rank.
+
+        :returns: An array of (rank, cell number) pairs; (-1, -1) if not found.
+        """
+        rank_cell_pairs = self._locate_mesh_elements()
+        return np.array(rank_cell_pairs, dtype=IntType)
+
+    def evaluate(self, function):
```
dham: I'm not convinced that having an evaluate method here is wise. What happens if I want to use this in the Functional for an optimisation problem? What does the symbol look like for the derivative of this operator with respect to the point cloud? What does its adjoint look like?

I think a PointCloud is analogous to a FunctionSpace. We probably need PointFunction. Then you would have an `UFL.Delta(expr, cloud)` which would evaluate expr on the point cloud and gives us the right place to do all the derivatives and adjoints. 

I guess it would be OK to have an evaluate method as long as what it does is `return assemble(UFL.Delta(expr, self))`

firedrake/pointcloud.py
dham: Ideally to be safe, the `PointCloud` would register a callback on the mesh such that the cache gets blown every time `mesh.coordinate.dat.data` gets touched.

firedrake/pointcloud.py
```diff
@@ -0,0 +1,283 @@
...
+        """Evaluate a function at the located points.
+
+        :arg function: The function to evaluate.
```
dham: Either way I'm pretty sure you mean a UFL expression, not a function.

firedrake/pointcloud.py
wence-: OK, let's punt on this for now (and remove the `evaluate` method). Instead a baby step in the right direction will be to make `Function.at` use the locations from here (rather than locating itself).

firedrake/pointcloud.py
wence-: I suppose that means we need to resurrect object versioning. I will add that to my lazy-evaluation purge.

cjlh:
Thank you for the comments @dham, and for making the distinction between what I referred to as an array of "points" and N x mesh.geometric_dimension() array of point locations. I have corrected this and removed the evaluation method from the class.

firedrake/pointcloud.py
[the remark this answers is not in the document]
cjlh: I have now pushed the code to correct this and send the data as one message, cheers.

wence-:
Reuben to look at in October.

firedrake/pointcloud.py
```diff
@@ -0,0 +1,299 @@
...
+        return np.array(rank_cell_pairs, dtype=IntType)
+
+    def _build_processes_spatial_index(self):
+        """Build a spatial index of processes using the bounding boxes of each process.
+        This will be used to determine which processes may hold a given point.
+
+        :returns: A libspatialindex spatial index structure.
+        """
+        min_c = self.mesh.coordinates.dat.data_ro.min(axis=0)
+        max_c = self.mesh.coordinates.dat.data_ro.max(axis=0)
```
lrtfm, line 91: Should it be 
```
min_c = self.mesh.coordinates.dat.data_ro_with_halos.min(axis=0)
 max_c = self.mesh.coordinates.dat.data_ro_with_halos.max(axis=0)
```

ReubenHill:
Should this be closed? At a glance I think my `VertexOnlyMesh` work supersedes this

PR #1390, opened 2019-03-12
Wrap libsupermesh's intersection finder routines.

pefarrell:
Interface the intersection finder routines for identifying containment relationships between meshes where we don't have a MeshHierarchy structure.

### DIFF
--- a/firedrake/supermeshimpl.pyx
+++ b/firedrake/supermeshimpl.pyx
@@ -142,8 +142,8 @@ def intersection_finder(mesh_A, mesh_B):
 
     vertices_A = mesh_A.coordinates.dat.data_ro_with_halos
     vertices_B = mesh_B.coordinates.dat.data_ro_with_halos
-    vertex_map_A = mesh_A.coordinates.cell_node_map().values_with_halo.astype(long)
-    vertex_map_B = mesh_B.coordinates.cell_node_map().values_with_halo.astype(long)
+    vertex_map_A = mesh_A.coordinates.cell_node_map().values_with_halo.astype(numpy.long)
+    vertex_map_B = mesh_B.coordinates.cell_node_map().values_with_halo.astype(numpy.long)
     nnodes_A = mesh_A.num_vertices()
     nnodes_B = mesh_B.num_vertices()
     dim_A = mesh_A.geometric_dimension()
@@ -159,8 +159,8 @@ def intersection_finder(mesh_A, mesh_B):
 
     libsupermesh_tree_intersection_finder_query_output(&nindices)
 
-    indices = numpy.empty((nindices,), dtype=long)
-    indptr  = numpy.empty((mesh_A.num_cells() + 1,), dtype=long)
+    indices = numpy.empty((nindices,), dtype=numpy.long)
+    indptr  = numpy.empty((mesh_A.num_cells() + 1,), dtype=numpy.long)
 
     libsupermesh_tree_intersection_finder_get_output(&ncells_A, &nindices, <long*>indices.data, <long*>indptr.data)
 
--- a/firedrake/supermeshimpl.pyx
+++ b/firedrake/supermeshimpl.pyx
@@ -120,6 +120,14 @@ def intersection_finder(mesh_A, mesh_B):
         long nnodes_A, nnodes_B, ncells_A, ncells_B
         int dim_A, dim_B, loc_A, loc_B
 
+    dim = mesh_A.geometric_dimension()
+    assert dim == mesh_B.geometric_dimension()
+    assert dim == mesh_A.topological_dimension()
+    assert dim == mesh_B.topological_dimension()
+
+    assert mesh_A.coordinates.function_space().ufl_element().degree() == 1
+    assert mesh_B.coordinates.function_space().ufl_element().degree() == 1
+
     vertices_A = mesh_A.coordinates.dat.data_ro_with_halos
     vertices_B = mesh_B.coordinates.dat.data_ro_with_halos
     vertex_map_A = mesh_A.coordinates.cell_node_map().values_with_halo.astype(long)
--- a/firedrake/supermeshimpl.pyx
+++ b/firedrake/supermeshimpl.pyx
@@ -59,6 +59,7 @@ def assemble_mixed_mass_matrix(V_A, V_B, candidates,
         numpy.ndarray[PetscReal, ndim=2, mode="c"] simplex_A, simplex_B
         numpy.ndarray[PetscReal, ndim=3, mode="c"] simplices_C
         compiled_call library_call = (<compiled_call *><uintptr_t>lib)[0]
+
     num_cell_A = V_A.mesh().cell_set.size
     num_cell_B = V_B.mesh().cell_set.size
 
@@ -68,6 +69,17 @@ def assemble_mixed_mass_matrix(V_A, V_B, candidates,
     mesh_B = V_B.mesh()
     vertex_map_A = mesh_A.coordinates.cell_node_map().values_with_halo
     vertex_map_B = mesh_B.coordinates.cell_node_map().values_with_halo
+
+    if mesh_A.comm.size > 1:
+        compatible = False
+        assert mesh_B._parallel_compatible is not None, "Whoever made mesh_B should explicitly mark mesh_A as having a compatible parallel layout."
+        for _mesh_A in mesh_B._parallel_compatible:
+            if mesh_A is _mesh_A():
+                compatible = True
+                break
+
+        if not compatible:
+            assert ValueError("Whoever made mesh_B should explicitly mark mesh_A as having a compatible parallel layout.")
 
     num_vertices = vertex_map_A.shape[1]
     gdim = mesh_A.geometric_dimension()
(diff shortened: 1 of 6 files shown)


firedrake/supermeshimpl.pyx
```diff
@@ -98,3 +103,50 @@ def assemble_mixed_mass_matrix(V_A, V_B, candidates,
...
+    dim_A = mesh_A.geometric_dimension()
+    dim_B = mesh_B.geometric_dimension()
+    ncells_A = mesh_A.num_cells()
+    ncells_B = mesh_B.num_cells()
+    loc_A = vertex_map_A.shape[1]
+    loc_B = vertex_map_B.shape[1]
+
+    libsupermesh_tree_intersection_finder_set_input(&nnodes_A, &dim_A, &ncells_A, &loc_A,
+                                                    &nnodes_B, &dim_B, &ncells_B, &loc_B,
+                                                    <const double*>vertices_A.data, <const long*>vertex_map_A.data, <const double*>vertices_B.data, <const long*>vertex_map_B.data)
```
wence-: Needs to check that mesh_A.comm and mesh_B.comm are congruent, and pass things into libsupermesh appropriately. If libsupermesh only works over COMM_WORLD then options are (in increasing order of preference)

1. Check that we're running on COMM_WORLD, raise NotImplementedError if not
2. Fix libsupermesh and pass the communicator in.

firedrake/supermeshing.py
```diff
@@ -78,29 +78,30 @@ def likely(cell_A):
...
-call libsupermesh's intersection finder here to compute the likely cell
-coverings that we fetch from the hierarchy.
-"""
 
-            raise NotImplementedError(msg)
+            # No mesh hierarchy structure, call libsupermesh for
+            # intersection finding
+            intersections = intersection_finder(mesh_A, mesh_B)
+            likely = intersections.__getitem__
+
```
wence-: No blank line here.

firedrake/supermeshimpl.pyx
```diff
@@ -98,3 +103,50 @@ def assemble_mixed_mass_matrix(V_A, V_B, candidates,
...
+
+    libsupermesh_tree_intersection_finder_query_output(&nindices)
+
+    indices = numpy.empty((nindices,), dtype=long)
+    indptr  = numpy.empty((mesh_A.num_cells() + 1,), dtype=long)
+
+    libsupermesh_tree_intersection_finder_get_output(&ncells_A, &nindices, <long*>indices.data, <long*>indptr.data)
```
wence-: What assumptions are there on the parallel decomposition of mesh_A and mesh_B. What happens if they don't coincide? Is the oracle still correct?

If it is not correct, what is the outcome? Silently wrong answer? Error condition?

firedrake/supermeshimpl.pyx
```diff
@@ -98,3 +103,50 @@ def assemble_mixed_mass_matrix(V_A, V_B, candidates,
...
+    vertex_map_A = mesh_A.coordinates.cell_node_map().values_with_halo.astype(long)
+    vertex_map_B = mesh_B.coordinates.cell_node_map().values_with_halo.astype(long)
+    nnodes_A = mesh_A.num_vertices()
+    nnodes_B = mesh_B.num_vertices()
+    dim_A = mesh_A.geometric_dimension()
+    dim_B = mesh_B.geometric_dimension()
+    ncells_A = mesh_A.num_cells()
+    ncells_B = mesh_B.num_cells()
+    loc_A = vertex_map_A.shape[1]
+    loc_B = vertex_map_B.shape[1]
```
wence-: What shape of elements does this handle?

wence-: Does it work if the elements are embedded?

wence-: Does it work if the mesh is not affine-mapped/bilinearly mapped.

firedrake/supermeshimpl.pyx
wence-: Possible option: provide metadata in the form of a weak reference to mesh_A and mesh_B in a magic slot that says "I guarantee that mesh_B has a decomposition that matches mesh_A". By default, this slot is empty => error. Pragmatic (or whatever) interfacing the remeshing can fill this slot in mesh_B with a weakref to mesh_A and we check this and satisfy ourselves that this is OK.

wence-: This also fixes the communicator issue above.

wence-: (But leave those checks in).

firedrake/supermeshimpl.pyx
pefarrell: It does not work if the mesh is embedded. I have added asserts to check that this is not the case.

It does not work if the mesh is not mapped with degree 1. I have added asserts.

Regarding shape of elements: it works for any polygon/polyhedron (it just relies on bounding box information). The interface does assume that all cells are the same kind of polytope. I have added a test for quads.

firedrake/supermeshing.py
pefarrell: Done.

firedrake/supermeshimpl.pyx
pefarrell: This intersection finding is done purely locally (no parallelism or communication), so libsupermesh doesn't see or need a communicator.

What do you mean by 'congruent' -- that they are the same (up to duplication)? I don't think this is necessary for the correct operation of the algorithm. For example, suppose that the target mesh was distributed across N cores, while the source mesh was associated with a coarse problem computed on COMM_SELF on each process independently. You can sensibly do a Galerkin projection from source to target, and need to calculate the intersections as done here, but the communicators aren't congruent (in the sense above).

What's actually required is that each core sees all the data it needs in the source mesh to compute its piece of the target function.

firedrake/supermeshimpl.pyx
pefarrell: There is a (strong) constraint on the parallel decomposition: each process needs to see all relevant source data for the computation of its piece of the target.

I don't know how to implement the magic slot idea you describe. Could you expand a little?

pefarrell: I have pushed a first implementation of the magic slot design you suggest.

firedrake/supermeshimpl.pyx
wence-: OK, thanks.

firedrake/supermeshimpl.pyx
```diff
@@ -64,6 +70,17 @@ def assemble_mixed_mass_matrix(V_A, V_B, candidates,
...
+    if mesh_A.comm.size > 1:
+        compatible = False
+        assert mesh_B._parallel_compatible is not None, "Whoever made mesh_B should explicitly mark mesh_A as having a compatible parallel layout."
+        for _mesh_A in mesh_B._parallel_compatible:
+            if mesh_A is _mesh_A():
+                compatible = True
+                break
+
+        if not compatible:
+            assert ValueError("Whoever made mesh_B should explicitly mark mesh_A as having a compatible parallel layout.")
```
wence-: Does adding the check here break projection in the mesh hierarchy case? (we haven't said that the meshes are parallel compatible?)

wence-: Since this function is an internal implementation one, we could move the checking into the interface code. That is in `intersection_finder` (which you already have), or else in the `supermesh.assemble_mixed_mass_matrix`?

firedrake/supermeshimpl.pyx
wence-: I noticed that the embedding dimension check is already handled in the calling code (sorry!).

wence-: I guess it doesn't hurt to check again in case someone calls this directly.

firedrake/supermeshimpl.pyx
```diff
@@ -98,3 +115,69 @@ def assemble_mixed_mass_matrix(V_A, V_B, candidates,
...
+
+    vertices_A = mesh_A.coordinates.dat.data_ro_with_halos
+    vertices_B = mesh_B.coordinates.dat.data_ro_with_halos
+    vertex_map_A = mesh_A.coordinates.cell_node_map().values_with_halo.astype(long)
+    vertex_map_B = mesh_B.coordinates.cell_node_map().values_with_halo.astype(long)
```
wence-: What is the actual Fortran type you need? Prefer explicit `numpy.int32/numpy.int64` if possible?

pefarrell: The actual Fortran type is `c_long`. I think it's machine-dependent how many bits that has? I agree this type stuff is a mess, though.

firedrake/supermeshimpl.pyx
pefarrell: Yes, it does break the nested case.

I suppose by creating a `MeshHierarchy` one is implicitly asserting that the parallel layouts are compatible. So I'll leave these checks only in `intersection_finder`. I prefer to leave them there because someone might call `intersection_finder` without wanting to assemble the mixed mass matrix.

pefarrell: Should I make a MeshHierarchy set `_parallel_compatible`? Or is that for meshing tools, like pragmatic?

firedrake/supermeshimpl.pyx
pefarrell: Yes, I think other code may want to call `intersection_finder` directly.

firedrake/supermeshimpl.pyx
pefarrell: `numpy.long`?

wence-: I guess so, yes.

pefarrell: OK, done.

firedrake/supermeshimpl.pyx
wence-: We decided to not do this last for now.

firedrake/supermeshimpl.pyx
wence-: Thanks.

PR #1391, opened 2019-03-15
Monkeypatch UFL so that users never put literals in their forms

wence-:
Doesn't fix the slow "rebuilding form" issues, but does at least make sure that they don't recompile when they change `t`.

### DIFF
--- a/firedrake/ufl_expr.py
+++ b/firedrake/ufl_expr.py
@@ -247,3 +247,8 @@ def FacetNormal(mesh):
     """
     mesh.init()
     return ufl.FacetNormal(mesh)
+
+
+for cls in [ufl.classes.IntValue, ufl.classes.FloatValue, ufl.classes.ComplexValue]:
+    # Make sure users never put literals in their forms.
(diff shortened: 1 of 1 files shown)


wence-:
@stephankramer this might DTRT.

firedrake/ufl_expr.py
```diff
@@ -247,3 +247,8 @@ def FacetNormal(mesh):
     """
     mesh.init()
     return ufl.FacetNormal(mesh)
+
+
+for cls in [ufl.classes.IntValue, ufl.classes.FloatValue, ufl.classes.ComplexValue]:
+    # Make sure users never put literals in their forms.
+    setattr(cls, "__new__", lambda cls, value: firedrake.Constant(value))
```
wence-: This breaks too much stuff: we can't be sure UFL isn't relying on these classes internally, for example.

wence-:
Turns out this is less simple than I thought, so I don't think we can transparently upcast everything from literals into Constants.

PR #1395, opened 2019-03-20
preconditioners: don't require python pmat by default

wence-:
Results in confusing user errors when they implement something that
should work.

Additionally, report the name of the PC in the error message to make
things easier to track down.

### DIFF
--- a/firedrake/preconditioners/assembled.py
+++ b/firedrake/preconditioners/assembled.py
@@ -18,8 +18,6 @@ class AssembledPC(PCBase):
 
     _prefix = "assembled_"
 
-    needs_python_pmat = False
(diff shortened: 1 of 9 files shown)


thomasgibson:
I see the test, but is this PR missing some changes to `preconditioners/base.py`?

wence-:
Oops.

thomasgibson:
The new test you added is failing. I just looked at it briefly in the log, but it appears that it's not catching the error you're expecting.

wence-:
Yes, I saw. it passes for me. I will look tomorrow.

wence-:
Fixed the test

PR #1397, opened 2019-03-20
Add symlinks for mpicc, mpicxx, mpif90, mpiexec inside the firedrake venv

pefarrell:
Example: if one runs

`firedrake-install --mpicc mpicc.mpich`

then a symlink will be created `firedrake/bin/mpicc` -> `$(which mpicc.mpich)`.

This is so that the installer can e.g. specify the use of MPICH on Ubuntu 18.04 and users can transparently continue to use `mpiexec` and friends as before.

### DIFF
--- a/scripts/firedrake-install
+++ b/scripts/firedrake-install
@@ -374,6 +374,18 @@ class directory(object):
         log.debug("New path '%s'" % self.olddir)
 
 
+class environment(object):
+    def __init__(self, **env):
+        self.old = os.environ.copy()
+        self.new = env
+
+    def __enter__(self):
+        os.environ.update(self.new)
+
+    def __exit__(self, *args):
+        os.environ = self.old
+
+
 options = config["options"]
 
 # Apply short cut package names.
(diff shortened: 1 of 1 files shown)


wence-:
Question 1: should we require that if the user provides _one of_ `--mpicc`, etc..., then they need to provide all of them? (So they don't accidentally get in a mess).

Question 2: should we report (at the end of the install) that we created these symlinks.

Question 3: should we mention this feature on the install documentation page?

dham:
> Question 2: should we report (at the end of the install) that we created these symlinks.

I'm not sure this is necessary, but the creation of the symlinks should go in the log for debugging purposes. I have done this.

> 
> Question 3: should we mention this feature on the install documentation page?

meh. That's more detail than we otherwise have in that page.

pefarrell:
Q1 and Q2 have been addressed; I agree with David that Q3 is unnecessary.

I confirm that this branch allows me to build the firedrake stack entirely against MPICH on Ubuntu 18.04, even without root access and without `update-alternatives`.

The build above failed with some JSON error in something to do with Zenodo. This doesn't look related to the PR. Could someone with the rights to do so start the tests again?

dham:
The merge commit is passing tests, which is the important one. I'm just going ahead.

wence-:
FWIW, I think that zenodo just had a temporary blip.

PR #1400, opened 2019-03-22
be more careful about --no-deps

dham:
Currently we end up leaving `--no-deps` on the our set of pip install instructions, which means that the documentation dependencies get installed without, in turn, their dependencies.

### DIFF
--- a/scripts/firedrake-install
+++ b/scripts/firedrake-install
@@ -397,6 +397,18 @@ python = ["%s/bin/python" % firedrake_env]
 # Use the pip from the venv
 pip = python + ["-m", "pip"]
 pipinstall = pip + ["install", "--no-binary", ",".join(wheel_blacklist)]
+
+# This context manager should be used whenever arguments should be temporarily added to pipinstall.
+class pipargs(object):
+    def __init__(self, *args):
+        self.args = args
+    def __enter__(self):
+        self.save = pipinstall.copy()
+        pipinstall.extend(self.args)
+
+    def __exit__(self, *args, **kwargs):
+        pipinstall = self.save
+
 pyinstall = python + ["setup.py", "install"]
 
 if "PYTHONPATH" in os.environ and not args.honour_pythonpath:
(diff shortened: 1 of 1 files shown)


scripts/firedrake-install
```diff
@@ -1265,11 +1266,16 @@ else:
...
+    pipinstall.pop()
 
 if options["slepc"]:
+    pipinstall.append("--no-deps")
     build_and_install_slepc()
+    pipinstall.pop()
 if options["slope"]:
+    pipinstall.append("--no-deps")
     build_and_install_slope()
+    pipinstall.pop()
```
wence-: As you note, this is rather fragile. How about:

```python
class pipargs(object):
    def __init__(self, *args):
        self.args = args
    def __enter__(self):
        self.save = pipinstall.copy()
        pipinstall.extend(self.args)

    def __exit__(self, *args, **kwargs):
        pipinstall = self.save

...

with pipargs("--no-deps"):
    build_and_install_slope()
```

etc...

WDYT?

scripts/firedrake-install
```diff
@@ -1160,21 +1172,20 @@ if mode == "install":
...
     build_and_install_h5py()
     build_and_install_libspatialindex()
     build_and_install_libsupermesh()
-    pipinstall.append("--no-deps")
-    for p in packages:
-        install(p+"/")
-        sys.path.append(os.getcwd() + "/" + p)
+    with pipargs("--no-deps"):
+        for p in packages:
+            install(p+"/")
```
wence-: Sympy now requires `mpmath` as a separate package. We install sympy when installing FIAT, but with `--no-deps`, this breaks the sympy installation (hence the present build error).

dham: Did this just happen? I don't understand why this change changes anything.

wence-: Not just now, I think (sympy >= 1.0, we're presently on 1.3). Like you, I don't understand why this change would break things in this way.

dham: So what's the fix?

wence-: No clue (I got as far as noticing why the build had failed).

dham: OK I'll come back to this later.

PR #1408, opened 2019-04-04
loopy codegen

wence-:
The firedrake side of the migration to loopy code generation. Again, I have squashed things so we have a chance of stuff being bisectable. Supersedes #1186.

### DIFF
--- a/firedrake/__init__.py
+++ b/firedrake/__init__.py
@@ -82,10 +82,6 @@ from firedrake.logging import *
 set_log_level(WARNING)
 set_log_handlers(comm=COMM_WORLD)
 
-# Turn off loopy caching since we cache the JITModule already
-import loopy
-loopy.set_caching_enabled(False)
-
 check()
 del check
 
--- a/firedrake/__init__.py
+++ b/firedrake/__init__.py
@@ -82,6 +82,10 @@ from firedrake.logging import *
 set_log_level(WARNING)
 set_log_handlers(comm=COMM_WORLD)
 
+# Turn off loopy caching since we cache the JITModule already
+import loopy
+loopy.set_caching_enabled(False)
+
 check()
 del check
 
--- a/firedrake/assemble_expressions.py
+++ b/firedrake/assemble_expressions.py
@@ -9,13 +9,16 @@ from ufl.mathfunctions import MathFunction
 from ufl.core.ufl_type import ufl_type as orig_ufl_type
 from ufl import classes
 
-import coffee.base as ast
+import loopy
+import pymbolic.primitives as p
 from pyop2 import op2
 from pyop2.profiling import timed_function
 
 from firedrake import constant
 from firedrake import function
 from firedrake import utils
+
+from functools import singledispatch, reduce
 
 
 def ufl_type(*args, **kwargs):
@@ -37,18 +40,6 @@ def ufl_type(*args, **kwargs):
     return decorator
 
 
-def _ast(expr):
-    r"""Convert expr to a COFFEE AST."""
-
-    try:
-        return expr.ast
-    except AttributeError:
-        for t, f in _ast_map.items():
-            if isinstance(expr, t):
-                return f(expr)
-        raise TypeError("No ast handler for %s" % str(type(expr)))
-
-
 class DummyFunction(ufl.Coefficient):
 
     r"""A dummy object to take the place of a :class:`.Function` in the
@@ -61,6 +52,7 @@ class DummyFunction(ufl.Coefficient):
         ufl.Coefficient.__init__(self, function.ufl_function_space())
 
         self.argnum = argnum
+        self.name = "fn_{0}".format(argnum)
         self.function = function
 
         # All arguments in expressions are read, except those on the
@@ -81,18 +73,7 @@ class DummyFunction(ufl.Coefficient):
 
     @property
     def arg(self):
-        argtype = self.function.dat.ctype + "*"
-        name = "fn_%r" % self.argnum
-
-        return ast.Decl(argtype, ast.Symbol(name))
-
-    @property
-    def ast(self):
-        # Constant broadcasts across functions if it's a scalar
-        if isinstance(self.function, constant.Constant) and \
-           len(self.function.ufl_element().value_shape()) == 0:
-            return ast.Symbol("fn_%d" % self.argnum, (0, ))
-        return ast.Symbol("fn_%d" % self.argnum, ("dim",))
+        return loopy.GlobalArg(self.name, dtype=self.function.dat.dtype, shape=loopy.auto)
 
 
 class AssignmentBase(Operator):
@@ -121,18 +102,12 @@ class AssignmentBase(Operator):
         return "%s(%s)" % (self.__class__.__name__,
                            ", ".join(repr(o) for o in self.ufl_operands))
 
-    @property
-    def ast(self):
-
-        return self._ast(_ast(self.ufl_operands[0]), _ast(self.ufl_operands[1]))
-
 
 @ufl_type(num_ops=2, is_abstract=False, is_index_free=True, is_shaping=False)
 class Assign(AssignmentBase):
 
     r"""A UFL assignment operator."""
     _symbol = "="
-    _ast = ast.Assign
     __slots__ = ("ufl_shape",)
 
     def _visit(self, transformer):
@@ -188,7 +163,6 @@ class IAdd(AugmentedAssignment):
 
     r"""A UFL `+=` operator."""
     _symbol = "+="
-    _ast = ast.Incr
     __slots__ = ()
 
 
@@ -197,7 +171,6 @@ class ISub(AugmentedAssignment):
 
     r"""A UFL `-=` operator."""
     _symbol = "-="
-    _ast = ast.Decr
     __slots__ = ()
 
 
@@ -206,7 +179,6 @@ class IMul(AugmentedAssignment):
 
     r"""A UFL `*=` operator."""
     _symbol = "*="
-    _ast = ast.IMul
     _identity = IntValue(1)
     __slots__ = ()
 
@@ -216,35 +188,8 @@ class IDiv(AugmentedAssignment):
 
     r"""A UFL `/=` operator."""
     _symbol = "/="
-    _ast = ast.IDiv
     _identity = IntValue(1)
     __slots__ = ()
-
-
-class Power(ufl.algebra.Power):
-
-    r"""Subclass of :class:`ufl.algebra.Power` which prints pow(x,y)
-    instead of x**y."""
-
-    def __str__(self):
-        return "pow(%s, %s)" % self.ufl_operands
-
-    @property
-    def ast(self):
-        return ast.FunCall("pow", _ast(self.ufl_operands[0]), _ast(self.ufl_operands[1]))
-
-
-class Ln(ufl.mathfunctions.Ln):
-
-    r"""Subclass of :class:`ufl.mathfunctions.Ln` which prints log(x)
-    instead of ln(x)."""
-
-    def __str__(self):
-        return "log(%s)" % str(self.ufl_operands[0])
-
-    @property
-    def ast(self):
-        return ast.FunCall("log", _ast(self.ufl_operands[0]))
 
 
 class ComponentTensor(ufl.tensors.ComponentTensor):
@@ -254,10 +199,6 @@ class ComponentTensor(ufl.tensors.ComponentTensor):
     def __str__(self):
         return str(self.ufl_operands[0])
 
-    @property
-    def ast(self):
-        return _ast(self.ufl_operands[0])
-
 
 class Indexed(ufl.indexed.Indexed):
     r"""Subclass of :class:`ufl.indexed.Indexed` which only prints the first
@@ -265,10 +206,6 @@ class Indexed(ufl.indexed.Indexed):
 
     def __str__(self):
         return str(self.ufl_operands[0])
-
-    @property
-    def ast(self):
-        return _ast(self.ufl_operands[0])
 
 
 class ExpressionSplitter(ReuseTransformer):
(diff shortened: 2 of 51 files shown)


firedrake/pointeval_utils.py
```diff
@@ -109,26 +109,42 @@ def predicate(index):
 
     code = {
         "geometric_dimension": cell.geometric_dimension(),
-        "extruded_arg": ", %s nlayers" % as_cstr(IntType) if extruded else "",
-        "nlayers": ", f->n_layers" if extruded else "",
+        # "extruded_arg": ", %int* layers" if extruded else "",
```
wence-: Commented code.

firedrake/pointeval_utils.py
```diff
@@ -109,26 +109,42 @@ def predicate(index):
...
     }
+    int layers[2] = {0, 0};
+    if (f->extruded != 0) {
+        int nlayers = f->n_layers;
+        layers[1] = cell %% nlayers + 2;
+        cell = cell / nlayers;
+    }
 
-    wrap_evaluate(result, reference_coords.X, f->coords, f->coords_map, f->f, f->f_map%(nlayers)s, cell);
+    wrap_evaluate(result, reference_coords.X, cell, cell+1%(layers)s, f->coords, f->f, %(map_args)s);
```
wence-: OK, so this is code that pyop2 makes, and it has changed argument order I guess.

firedrake/pointquery_utils.py
```diff
@@ -157,8 +158,12 @@ def compile_coordinate_element(ufl_coordinate_element, contains_eps, parameters=
...
         "dX_norm_square": dX_norm_square(cell.topological_dimension()),
         "X_isub_dX": X_isub_dX(cell.topological_dimension()),
-        "extruded_arg": ", %s nlayers" % as_cstr(IntType) if extruded else "",
-        "nlayers": ", f->n_layers" if extruded else "",
+        "extruded_arg": ", int const *__restrict__ layers" if extruded else "",
+        # "layer_arg": ", int layer" if extruded else "",
+        # "layers_init":"int layers[2] = [layer, layer+1]" if extruded else "",
+        "extr_comment_out": "//" if extruded else "",
+        "non_extr_comment_out": "//" if not extruded else "",
+        # "layers": ", layers" if extruded else "",
```
wence-: More commented code.

firedrake/pointquery_utils.py
```diff
@@ -197,15 +202,25 @@ def compile_coordinate_element(ufl_coordinate_element, contains_eps, parameters=
...
-    wrap_to_reference_coords(result_, x, &return_value, f->coords, f->coords_map%(nlayers)s, cell);
+    %(extr_comment_out)swrap_to_reference_coords(result_, x, &return_value, cell, cell+1, f->coords, f->coords_map);
     return return_value;
 }
+
+int to_reference_coords_xtr(void *result_, struct Function *f, int cell, int layer, double *x)
+{
+    int return_value;
+    int layers[2] = {0, layer+2};  // +2 because the layer loop goes to layers[1]-1, which is nlayers-1
+    %(non_extr_comment_out)swrap_to_reference_coords(result_, x, &return_value, cell, cell+1, layers, f->coords, f->coords_map);
```
wence-: This is disgusting. Why do we have to comment the code out in these different cases?

firedrake/preconditioners/patch.py
```diff
@@ -31,105 +35,57 @@ def __init__(self, rset, cset):
...
-    def __init__(self, dset):
+
+class LocalDatPack(LocalPack, DatPack):
+    def __init__(self, needs_mask, *args, **kwargs):
+        self.needs_mask = needs_mask
+        super().__init__(*args, **kwargs)
+
+    def _mask(self, map_):
+        if self.needs_mask:
+            return Comparison(">=", map_, Literal(numpy.int32(0)))
```
wence-: This is so the dat packing code handles negative map indices correctly.

firedrake/preconditioners/patch.py
```diff
@@ -179,17 +135,19 @@ def matrix_funptr(form, state):
...
+        arg = mat(op2.INC, (entity_node_map, entity_node_map))
         arg.position = 0
         args.append(arg)
-        statedat = DenseDat(dofset)
-        statearg = statedat(op2.READ, entity_node_map[op2.i[0]])
+        statedat = LocalDat(dofset)
+        state_entity_node_map = op2.Map(iterset,
+                                        toset, arity,
+                                        values=numpy.zeros(iterset.total_size*arity, dtype=IntType))
+        statearg = statedat(op2.READ, state_entity_node_map)
```
wence-: The new loopy codegen passes maps only once if they are identical (object identity). Because the state arg has a different map to the matrix, we need to make the "same" map again.

firedrake/preconditioners/patch.py
```diff
@@ -602,8 +603,12 @@ def Ffacet_op(pc, point, vec, mat, facetIS, facet_dofmap, facet_dofmapWithAll):
...
                     nfacet = len(facets)
                     dofs = facet_dofmap.ctypes.data
-                    Fint_facet_kernel.funptr(0, nfacet, facets.ctypes.data, mat.handle,
-                                             dofs, dofs, *facet_Fop_args)
+                    if facet_Fop_state_data_slot is not None:
+                        assert facet_dofmapWithAll is not None
+                        facet_Fop_data_args[facet_Fop_state_data_slot] = vec.array_r.ctypes.data
+                        facet_Fop_map_args[facet_Fop_state_map_slot] = facet_dofmapWithAll.ctypes.data
+                    Fint_facet_kernel.funptr(0, nfacet, facets.ctypes.data, mat.handle, *facet_Fop_data_args,
+                                             dofs, *facet_Fop_map_args)
```
wence-: I need to think about how to refactor `assemble` to give us a bunch of this "for free", but have not done so yet.

firedrake/slate/slac/compiler.py
```diff
@@ -156,7 +156,8 @@ def generate_kernel_ast(builder, statements, declared_temps):
     result_sym = ast.Symbol("T%d" % len(declared_temps))
     result_data_sym = ast.Symbol("A%d" % len(declared_temps))
     result_type = "Eigen::Map<%s >" % eigen_matrixbase_type(shape)
-    result = ast.Decl(SCALAR_TYPE, ast.Symbol(result_data_sym, shape))
+    # result = ast.Decl(SCALAR_TYPE, ast.Symbol(result_data_sym, shape))
+    result = ast.Decl(SCALAR_TYPE, ast.Symbol(result_data_sym), pointers=[("restrict",)])
```
wence-: Commented code.

firedrake/slope_limiter/vertex_based_limiter.py
```diff
@@ -37,26 +37,30 @@ def __init__(self, space):
...
+        <float64> alpha = 1
+        <float64> qavg = qbar[0, 0]
+        for i
+            <float64> _alpha1 = fmin(alpha, fmin(1, (qmax[i] - qavg)/(q[i] - qavg)))
+            <float64> _alpha2 = fmin(alpha, fmin(1, (qavg - qmin[i])/(qavg - q[i])))
+            alpha = if(q[i] > qavg, _alpha1, if(q[i] < qavg, _alpha2, alpha))
+        end
+        for ii
+            q[ii] = qavg + alpha * (q[ii] - qavg)
+        end
```
wence-: Is the scheduling deterministic @inducer? i.e. if I compile this stuff in parallel, will all processes agree on the instruction order?

firedrake/tsfc_interface.py
```diff
@@ -112,13 +116,15 @@ def __init__(self, form, name, parameters, number_map, interface):
         if self._initialized:
             return
 
-        tree = tsfc_compile_form(form, prefix=name, parameters=parameters, interface=interface)
+        assemble_inverse = parameters.get("assemble_inverse", False)
+        coffee = coffee or assemble_inverse
```
wence-: We should remove assemble_inverse, and I will do so in later code cleanup.

firedrake/mesh.py
```diff
@@ -1156,7 +1156,7 @@ def _c_locator(self, tolerance=None):
     int locator(struct Function *f, double *x)
     {
         struct ReferenceCoords reference_coords;
-        return locate_cell(f, x, %(geometric_dimension)d, &to_reference_coords, &reference_coords);
+        return locate_cell(f, x, %(geometric_dimension)d, &to_reference_coords, &to_reference_coords_xtr, &reference_coords);
```
dham: Another mysterious second case...

firedrake/assemble_expressions.py
```diff
@@ -591,25 +522,115 @@ def assemble_expression(expr, subset=None):
...
+    if (isinstance(expr.function, constant.Constant)
+            and len(expr.function.ufl_element().value_shape()) == 0):
+        # Broadcast if constant
+        return p.Variable(expr.name).index((0,))
+    return p.Variable(expr.name).index(context.indices)
+
+
+@loopy_instructions.register(ufl.constantvalue.Zero)
+def loopy_inst_zero(expr, context):
+    return 0
```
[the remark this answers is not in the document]
dham: Let's just try it and see ;)

firedrake/assemble_expressions.py
```diff
@@ -591,25 +522,115 @@ def assemble_expression(expr, subset=None):
...
+@loopy_instructions.register(ufl.algebra.Power)
+def loopy_inst_power(expr, context):
+    children = [loopy_instructions(o, context) for o in expr.ufl_operands]
+    return p.Power(*children)
+
+
+@loopy_instructions.register(ufl.algebra.Abs)
+def loopy_inst_abs(expr, context):
+    child, = [loopy_instructions(o, context) for o in expr.ufl_operands]
+    return p.Variable("abs")(child)
```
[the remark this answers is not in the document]
wence-: Pymbolic doesn't have a separate node name for function calls, so any `Variable` implements `__call__` to make a function call.

firedrake/assemble_expressions.py
```diff
@@ -516,21 +449,19 @@ def expression_kernel(expr, args):
...
+    inames = isl.make_zero_and_vars(["d"])
+    domain = (inames[0].le_set(inames["d"])) & (inames["d"].lt_set(inames[0] + fs.dof_dset.cdim))
+
+    context = Bag()
+    context.within_inames = frozenset(["d"])
+    context.indices = (p.Variable("d"),)
+
+    insn = loopy_instructions(expr, context)
+    data = [arg.arg for arg in args]
+    knl = loopy.make_function([domain], [insn], data, name="expression", silenced_warnings=["summing_if_branches_ops"])
```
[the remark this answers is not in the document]
kaushikcfd: Its not needed, as `make_function`(callee) does not take a `lang_version`, but `make_program`(caller kernel) does.

firedrake/assemble_expressions.py
```diff
@@ -591,25 +522,115 @@ def assemble_expression(expr, subset=None):
...
+
+@loopy_instructions.register(ufl.constantvalue.ScalarValue)
+def loopy_inst_scalar(expr, context):
+    return expr._value
+
+
+@loopy_instructions.register(ufl.algebra.Product)
+@loopy_instructions.register(ufl.algebra.Sum)
+@loopy_instructions.register(ufl.algebra.Division)
+def loopy_inst_binary(expr, context):
```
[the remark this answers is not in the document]
dham: I think @tj-sun was confused by misleading UFL documentation which incorrectly suggests that multiple operands are possible.

firedrake/__init__.py
```diff
@@ -82,6 +82,10 @@
 set_log_level(WARNING)
 set_log_handlers(comm=COMM_WORLD)
 
+# Turn off loopy caching since we cache the JITModule already
+import loopy
+loopy.set_caching_enabled(False)
```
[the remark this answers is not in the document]
wence-: Done.

firedrake/assemble_expressions.py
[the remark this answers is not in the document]
wence-: OK, thanks.

firedrake/assemble_expressions.py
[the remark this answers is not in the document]
wence-: Turns out no, but I added a comment explaining why not

firedrake/assemble_expressions.py
```diff
@@ -591,25 +522,115 @@ def assemble_expression(expr, subset=None):
...
+def loopy_inst_compare(expr, context):
+    left, right = [loopy_instructions(o, context) for o in expr.ufl_operands]
+    op = expr._name
+    return p.Comparison(left, op, right)
+
+
+@loopy_instructions.register(ComponentTensor)
+@loopy_instructions.register(Indexed)
+def loopy_inst_component_tensor(expr, context):
+    return loopy_instructions(expr.ufl_operands[0], context)
```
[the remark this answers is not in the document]
wence-: I figured it out, the indexing is handled elsewhere.

firedrake/function.py
```diff
@@ -444,8 +445,11 @@ def _constant_ctypes(self):
         c_function = _CFunction()
         c_function.n_cols = mesh.num_cells()
         if mesh.layers is not None:
+            # TODO: assert constant layer. Can we do variable though?
```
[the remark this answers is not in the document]
wence-: No need, because `mesh.layers` raises `ValueError` when called on a mesh with variable layers.

wence-:
Assuming the current round of builds passes, and others are happy, I think this is good to merge (along with OP2/PyOP2#559).

firedrake/slope_limiter/vertex_based_limiter.py
tj-sun: Oops, is this the reason that this fails in parallel when vectorised?

wence-: No idea, because I don't know what the failure mode is.

tj-sun: So it fails when we both vectorise and run in parallel, but not when we do only one of the two. And this is the only failure with vectorisation, which is quite weird. I will try to get to the end of this.

wence-: But what does "fail" mean. If the kernels are different on different processes you get a CompilationError. But it doesn't sound like you have that.

tj-sun: Indeed.
So it looks the problem is with vectorising vertex based limiters in general. When we batch elements, e.g. batch = cells 0, 1, 2, 3, cell 1 does not see the updates of the vertices of cell 0 because the values are loaded in batches before the updates.
Is this algorithm intrinsically serial?
And I'm not sure why the test passes with np=1, probably just coincidence.

wence-: Ah, ok. This is almost certainly because RW is a lie as an access descriptor. Instead it should be MIN or MAX. We will try and fix it this week.

PR #1409, opened 2019-04-04
Dockerise Jenkins

dham:
Cause Jenkins to run in a container. This will facilitate access to further build resources.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -1,12 +1,19 @@
 pipeline {
   agent {
-    label 'linux'
+    docker {
+        image 'firedrakeproject/firedrake-env:latest'
+        label 'firedrakeproject'
+        args '-v /var/run/docker.sock:/var/run/docker.sock'
+        alwaysPull true
+        }
   }
   environment {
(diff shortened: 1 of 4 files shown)


wence-:
Aside, should the "full" `Dockerfile` build on top of this one (so that we don't repeat the apt-get install line twice?)

Jenkinsfile
```diff
@@ -1,12 +1,19 @@
...
+        args '-v /var/run/docker.sock:/var/run/docker.sock'
+        alwaysPull true
+        }
   }
   environment {
     PATH = "/usr/local/bin:/usr/bin:/bin"
     CC = "mpicc"
     FIREDRAKE_CI_TESTS = "1"
     PYTHONHASHSEED = "12453221"
+    DOCKER_CREDENTIALS = credentials('f52ccab9-5250-4b17-9fb6-c3f1ebdcc986')
```
wence-: What can these credentials do? Does it matter that everyone can see them?

dham: Everyone can't directly see them. That's what credentials does.

Issue #1415, opened 2019-04-15
Matrix-free fieldsplits involving R blocks deadlock in parallel


pefarrell:
Firedrake cannot unfortunately assemble linear systems involving R blocks (to use LU). I therefore resort to Schur complement preconditioners that split onto the R block, use LU for the top-left block, and apply GMRES without any preconditioner to the (1x1) Schur complement system.

This strategy works successfully in serial but deadlocks in parallel. Consider the following code to solve the Moore-Spence system identifying the first pitchfork bifurcation point of the Euler elastica:

```
from defcon import *
from firedrake import *
from firedrake.petsc import PETSc
from slepc4py import SLEPc


msh = IntervalMesh(1000, 1)
V = FunctionSpace(msh, "CG", 1)
R = FunctionSpace(msh, "R", 0)

#elastica residual
def residual(theta, lmbda, ttheta):
    return inner(grad(theta), grad(ttheta))*dx - lmbda**2*sin(theta)*ttheta*dx

th = Function(V)
x = SpatialCoordinate(msh)[0]
tth = TestFunction(V)
lm = Constant(3.142)


# Using guess for parameter lm, solve for state theta (th)
A = residual(th, lm, tth)
bcs = [DirichletBC(V, 0.0, "on_boundary")]
solve(A==0, th, bcs=bcs)

#Now solve eigenvalue problem for $F_u(u, \lambda)\phi = r\phi$
#Want eigenmode phi with minimal eigenvalue r
B = derivative(residual(th, lm, TestFunction(V)),th, TrialFunction(V))

petsc_M = assemble(inner(TestFunction(V), TrialFunction(V))*dx, bcs=bcs).M.handle
petsc_B = assemble(B, bcs=bcs).M.handle

num_eigenvalues = 1

opts = PETSc.Options()
opts.setValue("eps_target_magnitude", None)
opts.setValue("eps_target", 0)
opts.setValue("st_type", "sinvert")

es = SLEPc.EPS().create(comm=COMM_WORLD)
es.setDimensions(num_eigenvalues)
es.setOperators(petsc_B, petsc_M)
es.setProblemType(SLEPc.EPS.ProblemType.GHEP)
es.setFromOptions()
es.solve()

ev_re, ev_im = petsc_B.getVecs()
r = es.getEigenpair(0, ev_re, ev_im)
eigenmode = Function(V)
eigenmode.vector().set_local(ev_re)

Z = MixedFunctionSpace([V,R,V])

[...]

PR #1420, opened 2019-04-17
Fix issue #1415.

pefarrell:
Add a regression test ensuring that fieldsplits onto R blocks don't deadlock in parallel.

### DIFF
--- a/tests/regression/test_moore_spence.py
+++ b/tests/regression/test_moore_spence.py
@@ -1,0 +1,105 @@
+import pytest
+from firedrake import *
+from firedrake.petsc import PETSc
+from slepc4py import SLEPc
+
+
+@pytest.mark.parallel(nprocs=2)
+def test_moore_spence():
+    msh = IntervalMesh(1000, 1)
+    V = FunctionSpace(msh, "CG", 1)
+    R = FunctionSpace(msh, "R", 0)
+
+    # elastica residual
+    def residual(theta, lmbda, ttheta):
+        return inner(grad(theta), grad(ttheta))*dx - lmbda**2*sin(theta)*ttheta*dx
+
+    th = Function(V)
+    x = SpatialCoordinate(msh)[0]
+    tth = TestFunction(V)
+    lm = Constant(3.142)
+
+    # Using guess for parameter lm, solve for state theta (th)
+    A = residual(th, lm, tth)
+    bcs = [DirichletBC(V, 0.0, "on_boundary")]
+    solve(A == 0, th, bcs=bcs)
+
+    # Now solve eigenvalue problem for $F_u(u, \lambda)\phi = r\phi$
+    # Want eigenmode phi with minimal eigenvalue r
+    B = derivative(residual(th, lm, TestFunction(V)), th, TrialFunction(V))
+
+    petsc_M = assemble(inner(TestFunction(V), TrialFunction(V))*dx, bcs=bcs).M.handle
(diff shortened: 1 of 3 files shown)


wence-:
Can you try this patch instead?
```diff
diff --git a/firedrake/matrix_free/operators.py b/firedrake/matrix_free/operators.py
index dc878086..cf26900e 100644
--- a/firedrake/matrix_free/operators.py
+++ b/firedrake/matrix_free/operators.py
@@ -1,4 +1,7 @@
 import operator
+from collections import OrderedDict
+from mpi4py import MPI
+import numpy
 
 import firedrake
 from firedrake.ufl_expr import adjoint, action
@@ -23,27 +26,32 @@ def find_sub_block(iset, ises):
         ``ises``.
     """
     found = []
-    sfound = set()
     comm = iset.comm
+    target_indices = iset.indices
+    comm = iset.comm.tompi4py()
+    candidates = OrderedDict(enumerate(ises))
     while True:
         match = False
-        for i, iset_ in enumerate(ises):
-            if i in sfound:
-                continue
-            lsize = iset_.getLocalSize()
-            if lsize > iset.getLocalSize():
-                continue
-            indices = iset.indices
-            tmp = PETSc.IS().createGeneral(indices[:lsize], comm=comm)
-            if tmp.equal(iset_):
+        for i, candidate in list(candidates.items()):
+            candidate_indices = candidate.indices
+            candidate_size, = candidate_indices.shape
+            target_size, = target_indices.shape
+            # Does the local part of the candidate IS match a prefix
+            # of the target indices?
+            lmatch = (candidate_size <= target_size
+                      and numpy.array_equal(target_indices[:candidate_size], candidate_indices))
+            if comm.allreduce(lmatch, op=MPI.LAND):
+                # Yes, this candidate matched, so remove it from the
+                # target indices, and list of candidate
+                target_indices = target_indices[candidate_size:]
                 found.append(i)
-                sfound.add(i)
-                iset = PETSc.IS().createGeneral(indices[lsize:], comm=comm)
+                ises.pop(i)
+                # And keep looking for the remainder in the remaining candidates.
                 match = True
-                continue
         if not match:
             break
-    if iset.getSize() > 0:
+    if comm.allreduce(len(target_indices), op=MPI.SUM) > 0:
+        # We didn't manage to hoover up all the target indices, not a match
         raise LookupError("Unable to find %s in %s" % (iset, ises))
     return found
```

It should be slightly cheaper as well in parallel.

tests/regression/test_moore_spence.py
```diff
@@ -0,0 +1,105 @@
+import pytest
+from firedrake import *
+from firedrake.petsc import PETSc
+from slepc4py import SLEPc
```
wence-: This test fails because the builders don't currently install slepc. I can fix that, but we should still fail gracefully if the user didn't install slepc. Rather than importing at the top level, can you do:

```
    try:
        from slepc4py import SLEPc
    except ImportError:
        pytest.skip(msg="SLEPc unavailable, skipping eigenvalue test")
```

inside the test itself.

pefarrell:
That patch is slightly off (should be `candidates.pop()`, not `ises.pop()`?). With this modification, the test now passes. Thanks!

wence-:
> That patch is slightly off (should be candidates.pop(), not ises.pop()?). With this modification, the test now passes. Thanks!

Ah yes, thanks!

PR #1429, opened 2019-04-26
Equation bcs

ksagiyam:
This branch allows for application of boundary conditions in equation form (EquationBC).  It also deprecates "bc.apply(A)," removing "thunk()."


firedrake/matrix_free/operators.py
```diff
@@ -161,27 +226,53 @@ def mult(self, mat, X, Y):
...
+                bbc.zero(self._y)
+
+            next(iter_assemble_actionT)()
+
+            self._xbc += self._x
+
+            # recursion
+            for bbc in bc.bcs:
+                if isinstance(bbc, EquationBCSplit):
+                    multTransposePart(bbc)
```
wence-: There's a lot of hand-crafted recursion on all of these bc objects. Can we find a common pattern? It looks like you want `for bc in EquationBC` to iterate over all recursive bcs.

firedrake/solving.py
```diff
@@ -137,6 +138,7 @@ def _solve_varproblem(*args, **kwargs):
         options_prefix = _extract_args(*args, **kwargs)
 
     appctx = kwargs.get("appctx", {})
+
```
wence-: Can we squish all of these unnecessary whitespace changes?

firedrake/solving.py
```diff
@@ -215,6 +218,19 @@ def _la_solve(A, x, b, **kwargs):
...
+    from firedrake.bcs import EquationBC
+
+    def rassert_linear(bcs):
+        for bc in bcs:
+            if isinstance(bc, EquationBC):
+                if bc.is_linear:
+                    rassert_linear(bc.bcs)
+                else:
+                    raise RuntimeError("EquationBCs must also be linear when solving linear system.")
```
wence-: Again, this looks like if you could iterate over the bcs of an `EquationBC` sanely, then you wouldn't have to hand-craft this recursion. Perhaps it it implemented:

```
class DirichletBC(...):
    is_linear = True
    def __iter__(self):
        yield self

class EquationBC(...):
    def __iter__(self):
        yield self
        yield from itertools.chain(*self.bcs)
```

Then this code would be something like:

```
bcs = _extract_bcs(bcs)
assert all(bc.is_linear for bc in bcs), "..."
```

tests/regression/test_slepc.py
```diff
@@ -33,8 +33,8 @@ def test_laplace_physical_ev(parallel=False):
     # This shifts the "1.0" Eigenvalues out of the spectrum
     # of interest (which is around 0.0 in this case).
     vals = np.repeat(1E8, len(bc.nodes))
-    A.setValuesLocalRCV(bc.nodes.reshape(-1, 1),
-                        bc.nodes.reshape(-1, 1),
+    A.setValuesLocalRCV(bc.nodes.reshape(-1, 1).astype('int32'),
+                        bc.nodes.reshape(-1, 1).astype('int32'),
```
wence-: This is incorrect (should use `pyop2.datatypes.IntType`), but also, we should ensure that the `nodes` that a BC object returns are _already_ of that type.

wence-:
Another comment. I don't understand the difference between `EquationBC` and `EquationBCSplit`. It seems like the user makes an `EquationBC`, but that's turned into an `EquationBCSplit` everywhere internally in a kind of weird way: the calling code has to know how to unpick things, rather than asking the `EquationBC` to do something.

firedrake/bcs.py
```diff
@@ -95,6 +80,173 @@ def __init__(self, V, g, sub_domain, method="topological"):
...
+        # Ex:
+        #                                      key.. one = 1 to avoid 0 * sqrt(...)
+        # facet  = ((1, ), )       ->     (one + 1) * sqrt(1)
+        # edge   = ((1, 2), )      ->     (one + 1) * sqrt(1) + (one + 2) * sqrt(2)
+        # vertex = ((1, 2, 4), )   ->     (one + 1) * sqrt(1) + (one + 2) * sqrt(2) + (one + 4) * sqrt(3)
+        #
+        # Multiple facets: (1, 2, 4) := ((1, ), (2, ), (4, )) returns ((1, 2, 4), )
+        #
+        # One facet and two edges:
+        # ((1,), (1, 3), (1, 4)) returns ((2, 2 + 4 * sqrt(2), 2 + 5 * sqrt(2)), )
```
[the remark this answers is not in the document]
ksagiyam: I intended this change to be backward-compatible, but I agree that using float is strange.
I just modified `bcs.py` and `functionspacedata.py` to reflect your suggestion and pushed. Does it look better?

firedrake/bcs.py
```diff
@@ -95,6 +80,173 @@ def __init__(self, V, g, sub_domain, method="topological"):
...
+        if isinstance(sub_d, str):
+            bcnodes = self._function_space.boundary_nodes(sub_d, self.method)
+            if isinstance(self._function_space.finat_element, finat.Hermite) and \
+               self._function_space.mesh().topological_dimension() == 1:
+                bcnodes = bcnodes[::2]  # every second dof is the vertex value
+        else:
+            sub_d = as_tuple(sub_d)
+            sub_d = [as_tuple(sub_d[i]) for i in range(len(sub_d))]
+            bcnodes = np.empty(0, dtype=int)
+            for iboun in range(len(sub_d)):
```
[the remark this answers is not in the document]
ksagiyam: True! Fixed.

firedrake/bcs.py
```diff
@@ -95,6 +80,173 @@ def __init__(self, V, g, sub_domain, method="topological"):
...
+                        self._function_space.mesh().topological_dimension() == 1:
+                        _bcnodes1 = _bcnodes1[::2]  # every second dof is the vertex value
+                    _bcnodes = np.intersect1d(_bcnodes, _bcnodes1)
+                bcnodes = np.concatenate((bcnodes, _bcnodes))
+        else:
+            bcnodes = self._function_space.boundary_nodes(self.sub_domain, self.method)
+            if isinstance(self._function_space.finat_element, finat.Hermite) and \
+                self._function_space.mesh().topological_dimension() == 1:
+                bcnodes = bcnodes[::2]  # every second dof is the vertex value
+        """
```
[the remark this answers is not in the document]
ksagiyam: Fixed.

firedrake/bcs.py
```diff
@@ -95,6 +80,173 @@ def __init__(self, V, g, sub_domain, method="topological"):
...
+                _bcnodes = self._function_space.boundary_nodes(sub_d[iboun][0], self.method)
+                if isinstance(self._function_space.finat_element, finat.Hermite) and \
+                   self._function_space.mesh().topological_dimension() == 1:
+                    _bcnodes = _bcnodes[::2]  # every second dof is the vertex value
+                for i in range(1, len(sub_d[iboun])):
+                    # intersection of facets
+                    # Edge conditions have only been tested with Lagrange elements.
+                    # Need to expand the list.
+                    if not isinstance(self._function_space.finat_element, finat.Lagrange):
+                        raise TypeError("Currently, edge conditions have only been tested with Lagrange elements")
```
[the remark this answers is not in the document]
ksagiyam: This algorithm extracts the edge (or vertex) nodes for H^1 functions by taking intersections of facet nodes.
We need this to apply edge Dirichlet BCs to a facet EquationBC.
For instance if we have an EquationBC on subdomain 1 and a DirichletBC on subdomain 3, we need to apply an edge DirichletBC at the intersection of 1 and 3 when we assemble the facet EquationBC on subdomain 1.

If we want to apply edge EquationBCs for facet EquationBCs, we need something better as we would also need edge iteration sets, but I think for now this can deal with many problems.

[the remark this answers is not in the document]
ksagiyam: This part looks better now, too.

firedrake/bcs.py
```diff
@@ -297,3 +396,93 @@ def homogenize(bc):
...
+            self.J = J or ufl_expr.derivative(self.F, self.u)
+            if Jp is not None and not isinstance(Jp, (ufl.Form, slate.slate.TensorBase)):
+                raise TypeError("Provided BC preconditioner is a '%s', not a Form or Slate Tensor" % type(Jp).__name__)
+            if Jp is not None and len(Jp.arguments()) != 2:
+                raise ValueError("Provided BC preconditioner is not a bilinear form")
+            self.Jp = Jp or self.J
+            self.is_linear = self.is_linear or False
+
+
+class EquationBCSplit(BCBase):
```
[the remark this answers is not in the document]
ksagiyam: Removed.

tests/regression/test_bcs.py
```diff
@@ -57,6 +57,7 @@ def test_assemble_bcs_wrong_fs(V, measure):
     W = FunctionSpace(V.mesh(), "CG", 2)
     with pytest.raises(RuntimeError):
         A = assemble(dot(u, v)*measure, bcs=[DirichletBC(W, 32, 1)])
+        print(A)  # F841
```
wence-: Prefer no `print(A)` and squash the warning either by not assigning the result to a variable, or else `# noqa: F401` (I think).

firedrake/bcs.py
```diff
@@ -297,3 +396,93 @@ def homogenize(bc):
...
+            if J is not None and not isinstance(J, (ufl.Form, slate.slate.TensorBase)):
+                raise TypeError("Provided BC Jacobian is a '%s', not a Form or Slate Tensor" % type(J).__name__)
+            if J is not None and len(J.arguments()) != 2:
+                raise ValueError("Provided BC Jacobian is not a bilinear form")
+            self.J = J or ufl_expr.derivative(self.F, self.u)
+            if Jp is not None and not isinstance(Jp, (ufl.Form, slate.slate.TensorBase)):
+                raise TypeError("Provided BC preconditioner is a '%s', not a Form or Slate Tensor" % type(Jp).__name__)
+            if Jp is not None and len(Jp.arguments()) != 2:
+                raise ValueError("Provided BC preconditioner is not a bilinear form")
+            self.Jp = Jp or self.J
```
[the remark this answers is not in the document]
ksagiyam: I have a vague idea of deriving `EquationBC` and `NonlinearVariationalProblem' from the same ancestor maybe called `Equation`, as these two both have 'F`, `J`, `Jp`, and now `bcs`.  
I think these two are mathematically very similar, and I think we can make the codes more monolithic if we regard them as the same thing. For instance, we currently have recursions over EquationBCs and deal with NonlinearVariationalProblem separately, but we can probably bring these operations together.  
In that case we can place this argument checking in the ancestor class.
I need to think more about this.  What would you think?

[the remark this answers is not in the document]
wence-: I was thinking in the short term just to pull that checking code out into a free function that can be used in two places. Rather than a larger-scale refactor.

I think it is worthwhile thinking about this latter change for the slightly longer term, but I would prefer to get these changes into a reasonable state sooner rather than later: I have some code that I want to base on top of this which is waiting for the merge.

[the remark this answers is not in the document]
ksagiyam: Ok! I'll do it now.

firedrake/solving.py
ksagiyam: This is cool.  Added __iter__ in `BCBase` and in `NonlinearVariationalProblem` and removed recursions wherever we can "flatten" bcs.

firedrake/matrix_free/operators.py
```diff
@@ -85,8 +85,26 @@ def __init__(self, a, row_bcs=[], col_bcs=[],
...
 
-        self.row_bcs = row_bcs
-        self.col_bcs = col_bcs
+        # Collect all DirichletBC instances including
+        # DirichletBCs applied to an EquationBC.
+        from firedrake.bcs import DirichletBC, EquationBCSplit
+
+        # all bcs (DirichletBC, EquationBCSplit)
+        self.bcs = row_bcs
+        self.bcs_col = col_bcs
```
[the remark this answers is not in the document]
ksagiyam: I think `self.bcs` here can include all kinds of bcs just like `problem.bcs`.
'self.row_bcs' and `self.col_bcs` should really be called 'self.row_dbcs' and `self.col_dbcs`, but these seem to be used in preconditionar(?), so I decided not to change this time.
Would there be possible confusion if we used `self.bcs`?

tests/regression/test_bcs.py
ksagiyam: Fixed!

tests/regression/test_slepc.py
ksagiyam: It once somehow complained about automatic conversion from 'int64' to 'int32', but it's gone now, so astype() is just removed.

ksagiyam:
`EquationBC` object has `F`, `J`, and `Jp`.
We cannot directly pass `EquationBC` objects to `_assemble()`, as this assembly function does not know if it is assembling `J` or `Jp` unless we set a flag for this (it knows it is assembling `F` by checking the rank).
That's why we introduced `EquationBCSplit` which only has `J` or `Jp` (or `F`).

firedrake/matrix_free/operators.py
```diff
@@ -85,16 +86,28 @@ def __init__(self, a, row_bcs=[], col_bcs=[],
...
+
+        # DirichletBCs
+        def collect_DirichletBCs(bcs):
+            for bc in bcs:
+                for b in bc:
+                    if isinstance(b, DirichletBC):
+                        yield b
```
wence-: AKA

```
tuple(bc for bc itertools.chain(*bcs) if isinstance(bc, DirichletBC))
```

firedrake/matrix_free/operators.py
```diff
@@ -118,11 +131,50 @@ def __init__(self, a, row_bcs=[], col_bcs=[],
...
+
+        # Each par_loop is to run with appropriate masks on self._y
+
+        self.list_assemble_actionT = []
+        self.list_assemble_actionT.append(create_assembly_callable(self.actionT,
+                                                                   tensor=self._x,
+                                                                   bcs=None,
+                                                                   form_compiler_parameters=self.fc_params))
+        for bc in self.bcs:
+            for b in bc:
```
wence-: Again:
```python
for bc in itertools.chain(*self.bcs):
```

firedrake/matrix_free/operators.py
ksagiyam: Thanks, fixed.

firedrake/matrix_free/operators.py
ksagiyam: fixed.

firedrake/matrix_free/operators.py
wence-: This code is still quite impenetrable to me. I think it is doing something like:
```python
for action, obj in zip(assembly_actions, (self, all_eq_bcs)):
    copy Y -> self._y
    for bc in obj.bcs:
       bc.zero(self._y)
    action()
    self._xbc += self._x
```

So for every object that has a form, we're lifting the action of the form to the result. The copy has to happen each time, because ... (I don't know this). Can you write out the maths?

In any case, this crafted recursion suggests we should have convenience:

```python
eq_bcs = tuple(bc for bc in itertools.chain(*list_of_bcs) if isinstance(bc, EquationBCSplit))
```

firedrake/matrix_free/operators.py
```diff
@@ -244,24 +318,44 @@ def createSubMatrix(self, mat, row_is, col_is, target=None):
...
-        for bc in self.row_bcs:
-            for i, r in enumerate(row_inds):
-                if bc.function_space().index == r:
-                    row_bcs.append(DirichletBC(Wrow.split()[i],
-                                               bc.function_arg,
-                                               bc.sub_domain,
-                                               method=bc.method))
-
-        if Wrow == Wcol and row_inds == col_inds and self.row_bcs == self.col_bcs:
+        def rcollect_bcs(bc, inds, row_inds, col_inds, Ws):
```
wence-: I guess you're using `r` to mean "recursive". But that's an implementation detail I think. Why not just call this `collect_bcs`? Or perhaps `collect_sub_bcs`?

firedrake/solving.py
```diff
@@ -215,15 +212,17 @@ def _la_solve(A, x, b, **kwargs):
 
     bcs, solver_parameters, nullspace, nullspace_T, near_nullspace, \
         options_prefix = _extract_linear_solver_args(A, x, b, **kwargs)
+
+    for bc in _extract_bcs(bcs):
+        if not all(b.is_linear for b in bc):
```
wence-: Realised one could write:
```python
for bc in itertools.chain(*_extract_bcs(bcs)):
    if not bc.is_linear:
       ...
```

firedrake/solving_utils.py
```diff
@@ -106,19 +107,53 @@ def __init__(self, problem, mat_type, pmat_type, appctx=None,
...
+            # copy if bc objects themselves have .bcs;
+            # see bcs.py.
+            def create_bc_tree(ebc, form_type):
+                ebcsplit = EquationBCSplit(ebc, getattr(ebc, form_type))
+                for bbc in ebc.bcs:
+                    if isinstance(bbc, DirichletBC):
+                        ebcsplit.add(bbc)
+                    elif isinstance(bbc, EquationBC):
+                        ebcsplit.add(create_bc_tree(bbc, form_type))
+                return ebcsplit
```
wence-: Lots of the "management" of EquationBC objects is scattered around at point of use. This kind of defeats much of the point of having an object for it.

Can we generalise the functionality of this type of code (I note it looks very similar to `get_ebc_tree` above), and make these methods of `EquationBC` (or at least have utility free functions in `bcs.py`).

firedrake/solving_utils.py
```diff
@@ -229,10 +266,38 @@ def split(self, fields):
...
+                        bc_F = replace(splitter.split(bc.F, argument_indices=(field, )), {bc.u: u})
+                        bc_J = replace(splitter.split(bc.J, argument_indices=(field, field)), {bc.u: u})
+                        bc_Jp = None if bc.Jp is None else replace(splitter.split(bc.Jp, argument_indices=(field, field)), {bc.u: u})
+                        ebc = EquationBC(bc_F == 0,
+                                         subu,
+                                         bc.sub_domain,
+                                         method=bc.method,
+                                         bcs=None,
+                                         J=bc_J,
+                                         Jp=None if bc.Jp_eq_J else bc_Jp,
```
wence-: Just make this `bc_Jp` and the assignment of `bc_Jp` something like `bc_Jp = None if (bc.Jp_eq_J or bc.Jp is None) else ...` ?

firedrake/solving_utils.py
```diff
@@ -229,10 +266,38 @@ def split(self, fields):
                         W = V.sub(field_renumbering[index])
                     if cmpt is not None:
                         W = W.sub(cmpt)
-                    bcs.append(type(bc)(W,
-                                        bc.function_arg,
-                                        bc.sub_domain,
-                                        method=bc.method))
+                    if isinstance(bc, DirichletBC):
+                        return DirichletBC(W,
```
wence-: This breaks user-defined subclasses of `DirichletBC`. TBF, the old code also did. But `DirichletBC` objects have a `reconstruct` method which one should use:
```
return bc.reconstruct(V=W, g=bc.function_arg, sub_domain=bc.sub_domain, method=bc.method)
```

firedrake/solving_utils.py
```diff
@@ -229,10 +266,38 @@ def split(self, fields):
...
+                                         bcs=None,
+                                         J=bc_J,
+                                         Jp=None if bc.Jp_eq_J else bc_Jp,
+                                         V=W,
+                                         is_linear=bc.is_linear)
+                        for bbc in bc.bcs:
+                            bc_temp = rcollect_bcs(bbc)
+                            # Due to the "if index", bc_temp can be None
+                            if bc_temp is not None:
+                                ebc.add(bc_temp)
```
wence-: Again, it looks like simple iteration over the bcs:
```python
for bc in itertools.chain(*problem.bcs):
   ...
```
Would be preferable to this recursion.

wence-: Or if not this, then we should think about how these kind of "tree-visitor" patterns can be encapsulated in easier to follow code.

firedrake/solving_utils.py
```diff
@@ -334,8 +394,15 @@ def compute_operators(ksp, J, P):
...
             inject(fine._x, ctx._x)
-            for bc in ctx._problem.bcs:
-                bc.apply(ctx._x)
+
+            def rapply(bcs, u):
+                for bc in bcs:
+                    if isinstance(bc, DirichletBC):
+                        bc.apply(u)
+                    elif isinstance(bc, EquationBC):
+                        rapply(bc.bcs, u)
```
wence-: Can this make use of the `__iter__` methods on these objects now? If not, seems like something has gone wrong.

firedrake/solving_utils.py
```diff
@@ -106,19 +107,53 @@ def __init__(self, problem, mat_type, pmat_type, appctx=None,
         self.F = problem.F
         self.J = problem.J
 
-        if mat_type != pmat_type or problem.Jp is not None:
+        # For Jp to equal J, bc.Jp must equal bc.J for all EquationBC objects.
+        Jp_eq_J = all(b.Jp_eq_J for b in problem)
```
wence-: Is this right? Do you mean `for b in problem.bcs`?

firedrake/variational_solver.py
```diff
@@ -35,37 +54,30 @@ def __init__(self, F, u, bcs=None, J=None,
...
+        # Argument checking
+        check_pde_args(self)
 
         # Store form compiler parameters
         self.form_compiler_parameters = form_compiler_parameters
         self._constant_jacobian = False
 
+    def __iter__(self):
+        yield self
+        yield from chain(*self.bcs)
```
wence-: This seems like an odd choice. Normally iterating over a thing should give you back things of that type. But here we get back the problem, and then a bunch of boundary conditions.

firedrake/variational_solver.py
```diff
@@ -225,8 +237,10 @@ def solve(self, bounds=None):
         # Make sure appcontext is attached to the DM before we solve.
         dm = self.snes.getDM()
         # Apply the boundary conditions to the initial guess.
-        for bc in self._problem.bcs:
-            bc.apply(self._problem.u)
+        from firedrake.bcs import DirichletBC
+        for b in self._problem:
+            if isinstance(b, DirichletBC):
+                b.apply(self._problem.u)
```
wence-: So I would rather than `NLVProblem` were not iterable, and this were just `for bc in chain(*self._problem.bcs):`

firedrake/variational_solver.py
```diff
@@ -281,7 +295,12 @@ def __init__(self, a, L, u, bcs=None, aP=None,
             F = ufl_expr.action(J, u) - L
 
         super(LinearVariationalProblem, self).__init__(F, u, bcs, J, aP,
-                                                       form_compiler_parameters=form_compiler_parameters)
+                                                       form_compiler_parameters=form_compiler_parameters,
+                                                       is_linear=True)
+        # check all equations are linear
+        for e in self:
+            assert(e.is_linear)
```
wence-: Prefer `assert e.is_linear, "message"`

firedrake/matrix_free/operators.py
ksagiyam: Fixed.

firedrake/solving.py
ksagiyam: Fixed.

wence-:
> EquationBC object has F, J, and Jp.
We cannot directly pass EquationBC objects to _assemble(), as this assembly function does not know if it is assembling J or Jp unless we set a flag for this (it knows it is assembling F by checking the rank).
That's why we introduced EquationBCSplit which only has J or Jp (or F).

OK, so the user makes an `EquationBC` which really is an "equation" to apply on the boundary. But then the internal logic all acts on these split things which are "forms + boundary conditions".

Aside, I would prefer if the construction of these things didn't expose an `add` method to add more bcs at a later date.

firedrake/solving_utils.py
ksagiyam: So, as I mentioned eariler, I think NLVP class and EquationBC class have a lot in common by nature, and I thought it makes sense to add __iter__ method in the same way in NLVP class.

wence-: While I agree that they are quite similar, iterating over the `EquationBC` gives back a bunch of things that are "like BCs". Whereas iterating over a `NLVP` gives back an `NLVP` and then a bunch of things "like BCs". Which feels less natural to me.

ksagiyam: So, my idea is to make a parent class for those two, say `Equation` class, which contains, e.g. `F`, `J`, `Jp`, `bc`, and `__iter__` method (but not in this PR at least).  The motivation for this is that even a 3D PDE can be regarded as an `EquationBC` of a fictitious 4D problem, which I believe simplifies the code by removing the need to separately deal with the 'domain' and the 'boundary.'  For instance, here, I can check `Jp_eq_J` in a monolithic way, which I think is more natural.  
`NLVP` currently has more things, so I would like to discuss more if we can really do this, but this was basically what I was thinking.

firedrake/solving_utils.py
ksagiyam: Fixed.

firedrake/variational_solver.py
ksagiyam: Fixed.

firedrake/solving_utils.py
ksagiyam: I missed this one.  Now fixed!

firedrake/solving_utils.py
ksagiyam: Thinking about this, I realised that `bc.Jp` is never `None`.  Simplified now.

firedrake/solving_utils.py
ksagiyam: Added a method `reconstruct` to EquationBC/EquationBCSplit.  It removed these tree-visiting recursions.

firedrake/solving_utils.py
ksagiyam: This part is also greatly simplified by `reconstruct` method.  I also changed the `reconstruct` method of `DirichletBC` class accordingly (this change should be backward compatible).

firedrake/matrix_free/operators.py
ksagiyam: Thanks! I cleaned this up using `zip` and added some explanation of the algorithm in the comment.  It turned out that we actually can avoid many copies here.

firedrake/assemble.py
```diff
@@ -381,167 +399,164 @@ def vec(testmap, i):
...
+            raise NotImplementedError("subdomain_data only supported with cell integrals.")
+
+        # Extract block from tensor and test/trial spaces
+        # FIXME Ugly variable renaming required because functions are not
+        # lexical closures in Python and we're writing to these variables
+        tsbc = []
+        trbc = []
+        if is_mat and result_matrix.block_shape > (1, 1):
+            # Unwind ComponentFunctionSpace to check for matching BCs
+            def collect_dbc(bcs):
```
thomasgibson: `collect_dbc` is defined twice. Can we try to merge the definition somehow and define it once? Perhaps if `collect_dbc` takes two arguments: `result_matrix`, and `bc`. Then the check for `block_shape` can be added to the definition of the function itself:
```python
def collect_dbc(result_matrix, bcs):
    if bcs is not None:
        if result_matrix.block_shape > (1, 1):
            for bc in bcs:
                fs = bc.function_space()
                if fs.component is not None:
                    fs = fs.parent
                if fs.index == i:
                    tsbc.append(bc)
                if fs.index == j:
                    if isinstance(bc, DirichletBC):
                        trbc.append(bc)
                    elif isinstance(bc, EquationBCSplit):
                        collect_dbc(bc.bcs)
                    else:
                        raise NotImplementedError("Unsupported bc type")
        else:
            if bcs is not None:
                for bc in bcs:
                    tsbc.append(bcs)
                    if isinstance(bc, DirichletBC):
                        trbc.append(bc)
                    elif isinstance(bc, EquationBCSplit):
                        collect_dbc(bc.bcs)

if is_mat:
    collect_dbc(bcs)
```

Might be a better way to refactor this, but just a suggestion.

firedrake/assemble.py
```diff
@@ -381,167 +399,164 @@ def vec(testmap, i):
...
+        # Decoration for applying to matrix maps in extruded case
+        decoration = None
+        itspace = m.measure_set(integral_type, subdomain_id,
+                                all_integer_subdomain_ids)
+        if integral_type == "cell":
+            itspace = sdata or itspace
+            if subdomain_id not in ["otherwise", "everywhere"] and sdata is not None:
+                raise ValueError("Cannot use subdomain data and subdomain_id")
+
+            def get_map(x):
```
thomasgibson: I think `get_map` can also be parameterized by the arguments `x` and `integral_type`.

firedrake/bcs.py
```diff
@@ -257,32 +423,169 @@ def apply(self, r, u=None):
...
+    def reconstruct(self, V, subu, u, field):
+        _F = self._F.reconstruct(field=field, V=V, subu=subu, u=u)
+        _J = self._J.reconstruct(field=field, V=V, subu=subu, u=u)
+        _Jp = self._Jp.reconstruct(field=field, V=V, subu=subu, u=u)
+        if all([_F is not None, _J is not None, _Jp is not None]):
+            return EquationBC(_F, _J, _Jp, Jp_eq_J=self.Jp_eq_J, is_linear=self.is_linear)
+
+
+class EquationBCSplit(BCBase):
+
```
thomasgibson: Although this is only used internally, might be useful to have a docstring attached that describes the function of this object.

firedrake/matrix_free/operators.py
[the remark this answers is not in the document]
thomasgibson: Some preconditioners use this. One option would be to rename here and change all existing preconditioners to use the updated attributes. However, I think this is fine as it currently is.

firedrake/variational_solver.py
```diff
@@ -15,12 +15,28 @@
...
+
+
 class NonlinearVariationalProblem(object):
     r"""Nonlinear variational problem F(u; v) = 0."""
 
     def __init__(self, F, u, bcs=None, J=None,
                  Jp=None,
-                 form_compiler_parameters=None):
+                 form_compiler_parameters=None,
+                 is_linear=False):
```
thomasgibson: Seems a bit strange to have an `is_linear` argument. Wouldn't it be possible to simply check for instances of `LinearVariationalProblem`? Otherwise, let's add `is_linear` to the docstring and specify why one might set this to be `True`

firedrake/variational_solver.py
```diff
@@ -281,7 +289,12 @@ def __init__(self, a, L, u, bcs=None, aP=None,
             F = ufl_expr.action(J, u) - L
 
         super(LinearVariationalProblem, self).__init__(F, u, bcs, J, aP,
-                                                       form_compiler_parameters=form_compiler_parameters)
+                                                       form_compiler_parameters=form_compiler_parameters,
+                                                       is_linear=True)
+        # check all equations are linear
+        for bc in self.bcs:
+            assert bc.is_linear, "All equations including ones defined in EquationBC objects have to be linear"
```
thomasgibson: Would `assert isinstance(bc, LinearVariationalProblem)` be sufficient?

firedrake/assemble.py
wence-: It's defined differently based on the integral_type right now.

firedrake/variational_solver.py
ksagiyam: `bc` is currently not an instance of `LinearVariationalProblem`, but an instance of `EquationBC.`  I have an idea of extracting common components of `EquationBCSplit` and `NonlinearVariationalProblem` (such as `F`, `J`, `Jp`, and `bcs`), which will also simplify this part.  I would like to discuss this idea some time, but as it requires many structural change, I'm not going to do it for this PR.

firedrake/bcs.py
```diff
@@ -94,6 +84,201 @@ def __init__(self, V, g, sub_domain, method="topological"):
...
+        return []
+
+    def increment_bc_depth(self):
+        # Increment _bc_depth by 1
+        self._bc_depth += 1
+        for bc in itertools.chain(*self.bcs):
+            bc._bc_depth += 1
+
+
+class DirichletBC(BCBase):
```
nbouziani: Is DirichletBC not a special case of EquationBCSplit when the operator is identity ? I do not understand why is there a need of a specific class for Dirichlet BC.

Is it to avoid solving the trivial equation u = BCoperator(u) = g ?

firedrake/assemble.py
ksagiyam: Fixed.

firedrake/variational_solver.py
ksagiyam: Thanks! The form style checks (`A == b` or `F == 0`) were actually missing in `NonlinearVariationalProblem.`  Cleaned up.

firedrake/bcs.py
ksagiyam: Added!

firedrake/matrix_free/operators.py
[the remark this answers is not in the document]
wence-: OK, let's do this for now. But note to future that we should consider some more consistent renaming.

ksagiyam:
Sounds great.  Thanks!

firedrake/bcs.py
wence-: Effectively yes. We can apply certain other optimisations (e.g. maintaining symmetry) in the DirichletBC case.

wence-:
Thanks!

PR #1430, opened 2019-04-29
Serendipity element

cyruscycheng21:
This branch has working tests for the serendipity elements in 2D and 3D. The serendipity elements are the 0-forms on the final column of the Periodic Table of Finite Elements.


tests/regression/test_steady_advection_2D.py
```diff
@@ -67,7 +69,7 @@ def test_left_to_right(mesh, DG0, W):
     run_left_to_right(mesh, DG0, W)
 
 
-@pytest.mark.parallel
+#@pytest.mark.parallel
```
dham: I think this is probably a left-over from previous testing.

tests/regression/test_helmholtz_serendipity.py
```diff
@@ -0,0 +1,101 @@
...
+    f = -div(grad(uex)) + uex
+
+    a = (inner(grad(u), grad(v)) + inner(u, v))*dx(degree=12)
+    L = inner(f, v)*dx(degree=12)
+
+    params = {"snes_type": "ksponly",
+              "ksp_type": "preonly",
+              "pc_type": "lu",
+              "snes_rtol": 1e-16,
+              "snes_atol": 1e-25}
```
dham: We should remove lines 45 and 46 because they aren't used when the snes_type is is-only.

tests/regression/test_helmholtz.py
```diff
@@ -77,6 +77,19 @@ def test_firedrake_helmholtz_scalar_convergence_on_quadrilaterals(testcase, conv
     assert (np.array([np.log2(l2err[i]/l2err[i+1]) for i in range(len(l2err)-1)]) > convrate).all()
 
 
+@pytest.mark.parametrize(('testcase', 'convrate'),
+                         [((1, (4, 6)), 1.9),
+                          ((2, (3, 6)), 2.9),
+                          ((3, (2, 4)), 3.9),
+                          ((4, (2, 4)), 4.7)])
+def test_firedrake_helmholtz_scalar_convergence_on_quadrilaterals_dpc(testcase, convrate):
```
thomasgibson: There is already a separate file that tests the serendipity element. I don't think we need this (another note: the name is incorrect. this is not a test for dPc)

tests/regression/test_helmholtz_serendipity.py
```diff
@@ -0,0 +1,99 @@
...
+
+
+@pytest.mark.parametrize(('testcase', 'convrate'),
+                         [((1, (4, 6)), 1.9),
+                          ((2, (3, 6)), 2.9),
+                          ((3, (2, 4)), 3.9),
+                          ((4, (2, 4)), 4.7),
+                          ((5, (2, 4)), 5.7),
+                          ((6, (2, 4)), 6.7)])
+def test_firedrake_helmholtz_scalar_convergence_on_quadrilaterals_s(testcase, convrate):
```
thomasgibson: I think the test is fine, except that going all the way to degree 6 seems to be a bit overkill. Maybe stop at 4?

Jenkinsfile~
```diff
@@ -0,0 +1,123 @@
...
+sudo docker push firedrakeproject/firedrake-env:latest
+sudo docker build --no-cache --build-arg PETSC_CONFIGURE_OPTIONS -t firedrakeproject/firedrake-vanilla:latest -f docker/Dockerfile.vanilla .
+sudo docker push firedrakeproject/firedrake-vanilla:latest
+sudo docker build --no-cache --build-arg PETSC_CONFIGURE_OPTIONS -t firedrakeproject/firedrake:latest -f docker/Dockerfile.firedrake .
+sudo docker push firedrakeproject/firedrake:latest
+'''
+      }
+    }
+  }
+}
```
wence-: This backup file should be removed.

tests/regression/test_helmholtz_serendipity.py
wence-: There was a reason.

wence-:
FIAT changes are in. Check that everything runs with master.

thomasgibson:
We will need https://github.com/firedrakeproject/tsfc/pull/198

dham:
> We will need [firedrakeproject/tsfc#198](https://github.com/firedrakeproject/tsfc/pull/198)

I've just merged it.

thomasgibson:
OK. Everything appears to be working now

PR #1432, opened 2019-05-03
patch: Support cell orientations

wence-:
@colinjcotter I think this works for you?


colinjcotter:
Thanks, when I run this, I get a segfault without any other traceback or output. But this might be due to not finding the points as you mentioned on Slack?

colinjcotter:
Oh, I can try a pure patch PC without MG

colinjcotter:
Yes, good, I confirm that this works just using patch pc as a pc without mg.

PR #1433, opened 2019-05-07
dmhooks: Fix movement of transfer operators

wence-:
Move them when coarsening the DM, not the function space. Avoids
leaving cycles around that the GC can't collect.


wence-:
It turns out that although this is a step in the right direction, it's not complete, so hold on merging for now. (I can't request changes on my own PR).

wence-:
Not ready yet, and needs to be completely refactored.

Issue #1436, opened 2019-05-16
Compiler options for Eigen in Slate


taupalosaurus:
slate/slac/compiler.py:227: include_dirs.extend(["%s/include/eigen3/" % d for d in PETSC_DIR])

This assumes that Eigen is installed by and in PETSc, which is not necessarily the case. Eigen can be passed as a configure option. This can happen notably when developing an external module used in PETSc that also uses Eigen.

I am not sure what the best way to retrieve the include directory is. Maybe there's a PETSc command that is able to spit it out, or a environment variable that contains that information. 
What can be done for sure is parsing PETSc configure log, which recalls the include and lib dirs of external packages at the end.

taupalosaurus:
Best option I can come up with so far: find the correct includes during the Firedrake install process, store it in an environment variable of the venv, and use it kernels. It might even be extended to other petsc external packages. 

I'm happy to try that if the idea is approved.

wence-:
Can you use the approach here https://github.com/FluidityProject/fluidity/blob/master/petsc_makefile ?

taupalosaurus:
Fixed in PR #1441

PR #1441, opened 2019-05-21
Extract eigen include dir from PETSc conf

taupalosaurus:
Draft fix for #1436.

I'm not very happy with it yet but at least we've got a base to discuss.

Needed improvement:
  - I assume that $PETSC_ARCH is set, which pyOP2 doesn't seem to. I'm not too sure why.

Possible improvements:
  - using Python (multiline) regexps would shorten the patch but may be overkill.

### DIFF
--- a/firedrake/slate/slac/compiler.py
+++ b/firedrake/slate/slac/compiler.py
@@ -240,8 +240,6 @@ def generate_kernel_ast(builder, statements, declared_temps):
                         return include_dir
         except FileNotFoundError:
             raise FileNotFoundError("Could not find %s/lib/petsc/conf/petscvariables." % PETSC_ARCH if PETSC_ARCH else PETSC_DIR)
-
-        file.close()
         raise ValueError(""" Could not find Eigen configuration in %s/lib/petsc/conf/petscvariables.
                              Check PETSc was built with the correct options.""" % PETSC_ARCH)
 
--- a/firedrake/slate/slac/compiler.py
+++ b/firedrake/slate/slac/compiler.py
@@ -51,6 +51,19 @@ except ValueError:
     PETSC_DIR = get_petsc_dir()
     PETSC_ARCH = None
 
+EIGEN_INCLUDE_DIR = None
+if COMM_WORLD.rank == 0:
+    with open(os.path.join(PETSC_ARCH, "lib", "petsc", "conf", "petscvariables") if PETSC_ARCH
+              else os.path.join(PETSC_DIR, "lib", "petsc", "conf", "petscvariables")) as file:
+        for line in file:
+            if line.find("EIGEN_INCLUDE") == 0:
+                EIGEN_INCLUDE_DIR = line[18:].rstrip()
+                break
+    if EIGEN_INCLUDE_DIR is None:
+        raise ValueError(""" Could not find Eigen configuration in %s/lib/petsc/conf/petscvariables.
+                             Check if PETSc was built with the correct options.""" % PETSC_ARCH)
+EIGEN_INCLUDE_DIR = COMM_WORLD.bcast(EIGEN_INCLUDE_DIR, root=0)
+
 cell_to_facets_dtype = np.dtype(np.int8)
 
 
@@ -230,26 +243,7 @@ def generate_kernel_ast(builder, statements, declared_temps):
     # Now we wrap up the kernel ast as a PyOP2 kernel and include the
     # Eigen header files
     include_dirs = builder.include_dirs
-
-    def extract_eigen_include_dir(PETSC_DIR, PETSC_ARCH):
-        include_dir = None
-        if COMM_WORLD.rank == 0:
-            with open(os.path.join(PETSC_ARCH, "lib", "petsc", "conf", "petscvariables") if PETSC_ARCH
-                      else os.path.join(PETSC_DIR, "lib", "petsc", "conf", "petscvariables")) as file:
-                for line in file:
-                    if line.find("EIGEN_INCLUDE") == 0:
-                        include_dir = line[18:].rstrip()
-                        break
-            if include_dir is None:
-                raise ValueError(""" Could not find Eigen configuration in %s/lib/petsc/conf/petscvariables.
-                                     Check PETSc was built with the correct options.""" % PETSC_ARCH)
-        include_dir = COMM_WORLD.bcast(include_dir, root=0)
-        if include_dir is None:
-            raise ValueError("Mayday")
-        return include_dir
-
-    include_dir = extract_eigen_include_dir(PETSC_DIR, PETSC_ARCH)
-    include_dirs.append(include_dir)
+    include_dirs.append(EIGEN_INCLUDE_DIR)
     op2kernel = op2.Kernel(kernel_ast,
                            macro_kernel_name,
                            cpp=True,
--- a/firedrake/slate/slac/compiler.py
+++ b/firedrake/slate/slac/compiler.py
@@ -232,8 +232,8 @@ def generate_kernel_ast(builder, statements, declared_temps):
 
     def extract_eigen_include_dir(PETSC_DIR, PETSC_ARCH):
         try:
-            with open(os.path.join(PETSC_ARCH, "lib/petsc/conf/petscvariables") if PETSC_ARCH
-                      else os.path.join(PETSC_DIR, "lib/petsc/conf/petscvariables")) as file:
+            with open(os.path.join(PETSC_ARCH, "lib", "petsc", "conf", "petscvariables") if PETSC_ARCH
+                      else os.path.join(PETSC_DIR, "lib", "petsc", "conf", "petscvariables")) as file:
                 for line in file:
                     if line.find("EIGEN_INCLUDE") == 0:
                         include_dir = line[18:].rstrip()
--- a/firedrake/slate/slac/compiler.py
+++ b/firedrake/slate/slac/compiler.py
@@ -43,12 +43,11 @@ import numpy as np
 __all__ = ['compile_expression']
 
 
-# Can't I assume that PETSC_ARCH is set ? or should I patch pyOP2 to exit when ARCH is not set ?
 try:
     PETSC_DIR, PETSC_ARCH = get_petsc_dir()
 except ValueError:
-    raise ValueError("PETSC_ARCH environment variable needs to be set but is not.")
-
+    PETSC_DIR = get_petsc_dir()
+    PETSC_ARCH = None
 
 cell_to_facets_dtype = np.dtype(np.int8)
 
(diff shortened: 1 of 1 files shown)


firedrake/slate/slac/compiler.py
```diff
@@ -43,7 +43,12 @@
 __all__ = ['compile_expression']
 
 
-PETSC_DIR = get_petsc_dir()
+# Can't I assume that PETSC_ARCH is set ? or should I patch pyOP2 to exit when ARCH is not set ?
```
wence-: No,  if PETSc was installed with a prefix install, then `PETSC_ARCH` is not needed. So you should set `PETSC_ARCH` to "" in that case.

firedrake/slate/slac/compiler.py
```diff
@@ -224,7 +229,23 @@ def generate_kernel_ast(builder, statements, declared_temps):
     # Now we wrap up the kernel ast as a PyOP2 kernel and include the
     # Eigen header files
     include_dirs = builder.include_dirs
-    include_dirs.extend(["%s/include/eigen3/" % d for d in PETSC_DIR])
+
+    def extract_eigen_include_dir(PETSC_ARCH):
+        try:
+            file = open("%s/lib/petsc/conf/petscvariables" % (PETSC_ARCH))
+        except:
+            raise FileNotFoundError("Could not find $PETSC_DIR/$PETSC_ARCH/lib/petsc/conf/petscvariables, where $PETSC_DIR/$PETSC_ARCH=%s." % PETSC_ARCH)
```
wence-: Prefer context managers:

```
try:
   with open(...) as f:
       ...
except FileNotFoundError:
   ...
```

wence-: Also prefer `os.path.join` over string interpolation.

firedrake/slate/slac/compiler.py
```diff
@@ -224,7 +229,23 @@ def generate_kernel_ast(builder, statements, declared_temps):
     # Now we wrap up the kernel ast as a PyOP2 kernel and include the
     # Eigen header files
     include_dirs = builder.include_dirs
-    include_dirs.extend(["%s/include/eigen3/" % d for d in PETSC_DIR])
+
+    def extract_eigen_include_dir(PETSC_ARCH):
+        try:
+            file = open("%s/lib/petsc/conf/petscvariables" % (PETSC_ARCH))
```
wence-: Assumes current working directory is `PETSC_DIR`, I think.

taupalosaurus: No, pyOP2 puts $PETSC_DIR+PETSC_ARCH in `PETSC_ARCH`. Don't ask me why...
Does it even make sense to use pyop2 for that at all ?

firedrake/slate/slac/compiler.py
taupalosaurus: Ok so in that case, where would the conf files be ? $PREFIX/petsc/conf ?
How do I retrieve the prefix ?

wence-: `get_petsc_dir()` should provide it.

firedrake/slate/slac/compiler.py
taupalosaurus: So just `os.path.join(PETSC_ARCH, "/lib/conf...")` or join on every sub-directory ?

wence-: `os.path.join(PETSC_ARCH, "lib", "conf", ...)` (you can pass `*directories`)

taupalosaurus:
I think it's all fixed.

firedrake/slate/slac/compiler.py
```diff
@@ -224,7 +229,24 @@ def generate_kernel_ast(builder, statements, declared_temps):
...
+            with open(os.path.join(PETSC_ARCH, "lib", "petsc", "conf", "petscvariables") if PETSC_ARCH
+                      else os.path.join(PETSC_DIR, "lib", "petsc", "conf", "petscvariables")) as file:
+                for line in file:
+                    if line.find("EIGEN_INCLUDE") == 0:
+                        include_dir = line[18:].rstrip()
+                        return include_dir
+        except FileNotFoundError:
+            raise FileNotFoundError("Could not find %s/lib/petsc/conf/petscvariables." % PETSC_ARCH if PETSC_ARCH else PETSC_DIR)
+
+        file.close()
```
wence-: The context manager closes the file for you.

firedrake/slate/slac/compiler.py
```diff
@@ -224,7 +229,24 @@ def generate_kernel_ast(builder, statements, declared_temps):
...
+        raise ValueError(""" Could not find Eigen configuration in %s/lib/petsc/conf/petscvariables.
+                             Check PETSc was built with the correct options.""" % PETSC_ARCH)
+
+    include_dir = extract_eigen_include_dir(PETSC_DIR, PETSC_ARCH)
```
wence-: Please only do this once (so at module level) rather than once per kernel invocation: I really don't want to parse a file on disk every time. To avoid hitting filesystems too hard on supercomputers please also do it as:
```python
if COMM_WORLD.rank == 0:
   ...
   COMM_WORLD.bcast(...)
else:
   ...
```

firedrake/slate/slac/compiler.py
```diff
@@ -224,7 +229,24 @@ def generate_kernel_ast(builder, statements, declared_temps):
     # Now we wrap up the kernel ast as a PyOP2 kernel and include the
     # Eigen header files
     include_dirs = builder.include_dirs
-    include_dirs.extend(["%s/include/eigen3/" % d for d in PETSC_DIR])
+
+    def extract_eigen_include_dir(PETSC_DIR, PETSC_ARCH):
+        try:
+            with open(os.path.join(PETSC_ARCH, "lib", "petsc", "conf", "petscvariables") if PETSC_ARCH
+                      else os.path.join(PETSC_DIR, "lib", "petsc", "conf", "petscvariables")) as file:
```
wence-: Prefer:
```python
filepath = os.path.join(PETSC_ARCH or PETSC_DIR, ...)
with open(filepath, "r") as file:
    for line in file:
        if line.find("EIGEN_INCLUDE") == 0:
            ...return ...
    raise RuntimeError("Did not find Eigen include directory in PETSc config %s, do you need --download-eigen?" % filepath)
```

Give that you catch and then immediately re-raise the same exception (filenotfound), there seems no need to catch that one.

firedrake/slate/slac/compiler.py
taupalosaurus: So there must be things I'm still missing about how Firedrake works.
Isn't the compiler only called at most once in a run ? or even not if the kernel is in the cache ?
In master, the kernel explicitly includes $PETSC_DIR/includes...Eigen/Dense.h. How is that better in terms of filesystem access ?

firedrake/slate/slac/compiler.py
taupalosaurus: I was initially trying to make a more explicit error message - left it there after the first refactoring. Is it worth trying to explain why we're trying to open the file ?

firedrake/slate/slac/compiler.py
wence-: Once per run per kernel.

wence-: Presently this hits the file system additionally to determine the same (include) information once every time you call generate_kernel_ast. I would like not to do that.

taupalosaurus: So I do it on one proc only now. I have left it in the concerned function though - do you (really) want to add a global variable containing this information that is used nowhere else ?

wence-: Yes. Because hitting the file system unnecessarily on a supercomputer can take 30minutes.

taupalosaurus: Ok. I've just realised the function was actually called several times. Made the change.

taupalosaurus:
I'm not sure I understand what causes the tests to break. Jenkins seems to complain about `extrusion` but these work just fine on my laptop. On my laptop however some Slate tests do not pass, but I'm not sure how it could be my fault...

wence-:
What is the error message on your laptop?

taupalosaurus:
There are several error messages, one twice in test_mimetic.py and then 3 NaNs:
http://pastebin.notk.org/pastebin.php?show=f5bc19e8c

wence-:
Do they happen on master?

firedrake/slate/slac/compiler.py
```diff
@@ -43,7 +45,24 @@
 __all__ = ['compile_expression']
 
 
-PETSC_DIR = get_petsc_dir()
+try:
+    PETSC_DIR, PETSC_ARCH = get_petsc_dir()
+except ValueError:
+    PETSC_DIR = get_petsc_dir()
```
wence-: This should be `PETSC_DIR, = get_petsc_dir()`.

firedrake/slate/slac/compiler.py
```diff
@@ -43,7 +45,24 @@
...
+    PETSC_ARCH = None
+
+EIGEN_INCLUDE_DIR = None
+if COMM_WORLD.rank == 0:
+    with open(os.path.join(PETSC_ARCH, "lib", "petsc", "conf", "petscvariables") if PETSC_ARCH
+              else os.path.join(PETSC_DIR, "lib", "petsc", "conf", "petscvariables")) as file:
```
wence-: Please write:
```
os.path.join(PETSC_ARCH or PETSC_DIR, "lib", "petsc", "conf", "petscvariables")
```

taupalosaurus: That is a cool trick !

taupalosaurus:
Tests trigger the same errors in Master... are these known bugs ?

wence-:
> Tests trigger the same errors in Master... are these known bugs ?

No, did you run `firedrake-clean` before trying?

taupalosaurus:
> did you run firedrake-clean before trying

I even did a `firedrake-update --rebuild` then `firedrake-clean`, I'm afraid the same tests fail in the same way. Maybe I'm doing something wrong. But that should probably go in a new issue.

firedrake/slate/slac/compiler.py
```diff
@@ -224,7 +242,7 @@ def generate_kernel_ast(builder, statements, declared_temps):
     # Now we wrap up the kernel ast as a PyOP2 kernel and include the
     # Eigen header files
     include_dirs = builder.include_dirs
-    include_dirs.extend(["%s/include/eigen3/" % d for d in PETSC_DIR])
+    include_dirs.append(EIGEN_INCLUDE_DIR)
```
wence-: This was already the case, but while we're here, can we fix this.
Say:
```python
include_dirs = list(builder.include_dirs)
```
So that a builder if reused doesn't make a longer-and-longer list of include_dirs

firedrake/slate/slac/compiler.py
```diff
@@ -43,7 +45,23 @@
...
+EIGEN_INCLUDE_DIR = None
+if COMM_WORLD.rank == 0:
+    with open(os.path.join(PETSC_ARCH or PETSC_DIR, "lib", "petsc", "conf", "petscvariables")) as file:
+        for line in file:
+            if line.find("EIGEN_INCLUDE") == 0:
+                EIGEN_INCLUDE_DIR = line[18:].rstrip()
+                break
+    if EIGEN_INCLUDE_DIR is None:
+        raise ValueError(""" Could not find Eigen configuration in %s/lib/petsc/conf/petscvariables.
+                             Check if PETSc was built with the correct options.""" % PETSC_ARCH)
```
wence-: This message is wrong if PETSC_ARCH is none. Reuse the bit from above like so:
```python
filepath = os.path.join(PETSC_ARCH or PETSC_DIR, "lib", "petsc", "conf", "petscvariables")
with open(filepath, "r") as f:
    for line in f:
        if line.find("EIGEN_INCLUDE") == 0:
            EIGEN_INCLUDE_DIR = line[18:].rstrip()
            break
if EIGEN_INCLUDE_DIR is None:
    raise ValueError("Could not find Eigen configuration in %s. Did you build PETSc with Eigen?" % filepath)
```

firedrake/slate/slac/compiler.py
wence-: Thanks.

PR #1448, opened 2019-05-30
Construct non-nested mesh hierarchies from CAD data with successively better approximation of the geometry

pefarrell:
Accurately representating the domain geometry is important for many problems. This branch introduces an `OpenCascadeMeshHierarchy` object that constructs a hierarchy of meshes (e.g. for use in multigrid) directly from CAD data (in STEP format). For example, the code

    mh = OpenCascadeMeshHierarchy("tests/meshes/t_twist.step", mincoarseh=3, maxcoarseh=3, levels=4)

makes a mesh hierarchy with 4 levels, with the finer ones offering a better approximation of the geometry than the coarse grid. Here are pictures of the coarsest and finest grids:

ref-0
ref-4

Essentially, the code constructs a coarse mesh with gmsh, then applies a standard `MeshHierarchy` to that coarse grid, then projects each surface vertex of each refined grid to its appropriate CAD surface. I need to control the construction of the coarse grid as I need to carefully colour every boundary so as to be able to identify the associated CAD surface later. (This could possibly be relaxed so that the user can supply the coarse grid, but the user would need to do the same surface colouring.)

To enable multigrid on these non-nested hierarchies, the standard multigrid transfer operators had to be slightly generalised. I believe that this generalisation also enables the use of multigrid on manifolds (I hope @colinjcotter can confirm this).

I have modified the `firedrake-install` script so that it installs the necessary OpenCascade C++ library with apt, and compiles the python bindings if `--opencascade` is passed. I don't have access to a machine running OSX, so I haven't changed the relevant `brew` commands; I hope someone with a Mac can help here.

The STEP files included for the tests were generated by Eder Medina.


pefarrell:
I should also remark that one can construct higher-order meshes by passing `order=2` etc. to `OpenCascadeMeshHierarchy`. If CAD data is available this can circumvent the problem that none of PETSc's mesh readers support reading in higher-order meshes.

firedrake/mg/opencascade_mh.py
```diff
@@ -0,0 +1,196 @@
...
+            for edge in edges:
+                s = str(edge)  # FIXME: is there a more elegant way to get the OCC id?
+                for other_edge in other_edges:
+                    other_s = str(other_edge)
+                    if s == other_s:
+                        intersecting_edges.append(edge)
+
+            if len(intersecting_edges) == 0:
+                warnings.warn("face: %s other_face: %s intersecting_edges: %s" % (face, other_face, intersecting_edges))
+                warnings.warn("Warning: no intersecting edges in CAD, even though vertices on both faces?")
```
wence-: I tried to think of a case where this might happen, but only in 4D?

firedrake/mg/opencascade_mh.py
```diff
@@ -0,0 +1,196 @@
...
+            except OSError:
+                pass
+
+            if verbose:
+                stdout = None
+            else:
+                stdout = subprocess.DEVNULL
+
+            gmsh = subprocess.Popen(["gmsh", "-3", geopath], stdout=stdout)
+            gmsh.wait()
```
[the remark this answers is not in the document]
pefarrell: I don't think so. pygmsh, as far as I understand it, is a tool for doing CAD in gmsh (using either its built-in kernel, or as an interface to opencascade). It addresses a different problem.

scripts/firedrake-install
```diff
@@ -1313,6 +1352,10 @@ with environment(**compiler_env):
     with pipargs("--no-deps"):
         if options["slepc"]:
             build_and_install_slepc()
+
+        if options["opencascade"]:
+            build_and_install_pythonocc()
+
```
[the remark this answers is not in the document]
pefarrell: pythonocc explicitly guarantees (https://github.com/tpaviot/pythonocc-core) that any version of pythonocc in the 0.18.x range (since 2017) will compile against any oce in the 0.18.x range. However, I agree it's probably a good idea to pin to a specific version, once I figure out how ...

firedrake/mg/opencascade_mh.py
pefarrell: I've seen it once with CAD data that wasn't well-formed.

firedrake/mg/opencascade_mh.py
```diff
@@ -0,0 +1,196 @@
...
+import numpy
+import subprocess
+import os
+import warnings
+
+
+__all__ = ("OpenCascadeMeshHierarchy",)
+
+
+def OpenCascadeMeshHierarchy(stepfile, mincoarseh, maxcoarseh, levels, comm=COMM_WORLD, distribution_parameters=None, callbacks=None, order=1, mh_constructor=MeshHierarchy, cache=True, verbose=True):
```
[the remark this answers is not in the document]
pefarrell: I think if someone wants to mesh a stepfile without making a mesh hierarchy, they can just do it in gmsh directly and use the current pipeline. This has the (significant) advantage that the user can colour their meshes however they wish, and can more finely control the mesh size distribution (this branch only supports gmsh's most elementary way). 

In the main there's only work to be done if the user wants a mesh hierarchy where the geometry approximation improves as the mesh is refined.

The one use case where your proposal is better is if someone wants a higher-order mesh but not a mesh hierarchy; in this case they can just do `OpenCascadeMeshHierarchy(..., levels=0, order=2).` This seems straightforward enough to me as-is, though.

scripts/firedrake-install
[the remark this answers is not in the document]
pefarrell: Version now pinned.

pefarrell:
One other comment: this work requires a relatively recent version of gmsh. (For concreteness, I know that version 4.0.1 doesn't work, but 4.3.0 does.) Should I check for this? Should we install gmsh in the installer?

pefarrell:
I have noticed a problem with the higher-order stuff. Consider the following diff:

```
diff --git a/tests/multigrid/test_opencascade_volume.py b/tests/multigrid/test_opencascade_volume.py
index 71a0501b..fcc30298 100644
--- a/tests/multigrid/test_opencascade_volume.py
+++ b/tests/multigrid/test_opencascade_volume.py
@@ -38,10 +38,9 @@ def compute_err(mh, v_true):
 
 @pytest.mark.parallel(nprocs=2)
 def test_volume(stepdata):
-
     (stepfile, h) = stepdata
     try:
-        mh = OpenCascadeMeshHierarchy(stepfile, mincoarseh=h, maxcoarseh=h, levels=3, cache=False, verbose=False)
+        mh = OpenCascadeMeshHierarchy(stepfile, mincoarseh=h, maxcoarseh=h, levels=3, cache=False, verbose=False, order=2)
         v_true = get_volume(stepfile)
     except ImportError:
         pytest.skip(msg="OpenCascade unavailable, skipping test")
```

It fails with

```
        if mesh.comm.size > 1 and mesh._grown_halos:
>           raise RuntimeError("Cannot refine parallel overlapped meshes "
                               "(make sure the MeshHierarchy is built immediately after the Mesh)")
E           RuntimeError: Cannot refine parallel overlapped meshes (make sure the MeshHierarchy is built immediately after the Mesh)
```

This only happens if `order > 1`. Is there a way to achieve the same effect as lines 36--39 of `opencascade_mh.py` without growing the halo?

wence-:
> This only happens if order > 1. Is there a way to achieve the same effect as lines 36--39 of opencascade_mh.py without growing the halo?

Can you do that interpolation and projecting after you've made the mesh hierarchy, and then return a `BaseHierarchy(high_order_meshes, ....)`?

pefarrell:
Any further thoughts on this pull request?

colinjcotter:
I intend to add a test for MG on the sphere - probably tomorrow?


all the best

--cjc

wence-:
> I intend to add a test for MG on the sphere - probably tomorrow?
> 

Do you still want to do this, or do you want to wait until after it is merged.

colinjcotter:
Here is a script that uses this feature. What's the best way to turn it into a test?

We could check that it does indeed converge in 3 iterations?

cheers
--cjc

colinjcotter:
rom firedrake import *

R = 1.0

base = IcosahedralSphereMesh(radius=R, refinement_level=0)
nref = 5
mh = MeshHierarchy(base, nref)
for mesh in mh:
    x = SpatialCoordinate(mesh)
    mesh.init_cell_orientations(x)

mesh = mh[-1]

u = TrialFunction(V)
v = TestFunction(V)

a = (u*v + inner(u,v))*dx


f1 = exp((x+y+z)/R)*x*y*z/R**3
F = v*f1*dx

lu_params = {
    "ksp_monitor": None,
    "ksp_type": "gmres",
    "pc_type": "lu",
    "mat_type": 'aij'
}

mg_params = {"mat_type": "matfree",
             "snes_type": "ksponly",
             "ksp_type": "gmres",
             "ksp_rtol": 1.0e-8,
             "ksp_atol": 0.0,
             "ksp_max_it": 1000,
             "ksp_monitor": None,
             "ksp_converged_reason": None,
             "ksp_norm_type": "unpreconditioned",
             "pc_type": "mg",
             "mg_coarse_ksp_type": "preonly",
             "mg_coarse_pc_type": "python",
             "mg_coarse_pc_python_type": "firedrake.AssembledPC",
             "mg_coarse_assembled_pc_type": "lu",
             "mg_coarse_assembled_pc_factor_mat_solver_type": "mumps",
             "mg_levels_pc_type": "python",
             "mg_levels_pc_python_type": "firedrake.AssembledPC",
             "mg_levels_assembled_pc_type": "bjacobi",
             "mg_levels_assembled_pc_type": "ilu",
             "mg_levels_ksp_type": "richardson",
             "mg_levels_ksp_max_it": 1}

w = Function(V)

mg = True
if mg:
    Prob = LinearVariationalProblem(a, F, w)
    Solver = LinearVariationalSolver(Prob,
                                     solver_parameters=mg_params)
else:
    Prob = LinearVariationalProblem(a, F, w)
    Solver = LinearVariationalSolver(Prob, solver_parameters=lu_params)
Solver.solve()

wence-:
Yes, wrapped in a `test_convergence_on_sphere` or similar.

colinjcotter:
How do you check for the number of iterations?

wence-:
```python
solver.snes.ksp.getIterationNumber()
```

wence-:
I think the only thing that is missing is installation of opencascade OCE on MacOS? Is there a magic incantation? It seems that python-occ doesn't support the release opencascade, but needs this community edition. Homebrew science supplies "taps", but is out of date/not maintained?

scripts/firedrake-install
```diff
@@ -853,6 +858,39 @@ def build_and_install_libsupermesh():
...
+    if os.path.exists("pythonocc-core"):
+        log.info("Updating the git repository for pythonocc-core")
+        with directory("pythonocc-core"):
+            check_call(["git", "fetch"])
+            git_sha = check_output(["git", "rev-parse", "HEAD"])
+            git_sha_new = check_output(["git", "rev-parse", "@{u}"])
+            changed = git_sha != git_sha_new
+            if changed:
+                check_call(["git", "reset", "--hard"])
+                check_call(["git", "pull"])
```
wence-: I think we don't want to do this, because `firedrake-update` will now override the carefully checked out version?

colinjcotter:
OK, I pushed a test, which will require merging from master to work.

scripts/firedrake-install
pefarrell: I have removed this, thanks for pointing it out.

pefarrell:
I have pushed Colin's test to this branch.

pefarrell:
I have fixed Colin's test.

Can anyone with a mac please help with installing opencascade via homebrew?

dham:
In principle, yes. However the chance of it actually happening before the end of term next week is vanishing.

wence-:
Still just waiting for Mac build instructions.

dham:
Turns out the MacOS build instructions don't work because the Python package fails to find the cmake files.

dham:
Next piece of joy: setting the relevant environment variable and trying to build the python package crashes my whole laptop.

florianwechsung:
I have pushed some changes to fw/occhighorderpar:

- high order now seems to work in parallel
- one can also use 2D geometries and not just 3D

I don't mind whether those are merged into this branch or handled later in a separate PR

pefarrell:
I've merged your changes into this PR, @florianwechsung .

florianwechsung:
I have made some more changes in 976113831f42f0e3bde96f7e4bc3425cc7d5c580, any objections to merging those in @pefarrell ?

florianwechsung:
I made some more changes in [fw/moreocc](https://github.com/firedrakeproject/firedrake/tree/fw/moreocc). The dimension is now figured out automatically, the interface to gmsh a little bit more flexible, and one can preserve nestedness of the hierarchy by choosing a high order coarse mesh and then simply prolonging the coordinate field into the fine mesh. Two small things:

- I noticed that with the current code, even if the refined meshes are snapped to the boundary, `MeshHierarchy.nested` is still set to `true`, even though that clearly does not have to be the case and in fact @pefarrell had to change the transfer kernels to handle this non-nestedness. Should this be changed?

pefarrell:
Thanks. I've merged these changes. The code now sets the `nested` variable appropriately, depending on whether it projects to CAD or not.

pefarrell:
> Next piece of joy: setting the relevant environment variable and trying to build the python package crashes my whole laptop.

I removed the call to `make -j` in building pythonocc; it should now work on your laptop, @dham .

florianwechsung:
have added an example. in practice one would do something more involved like boundary layers or specifying a spatially varying function that gives the element size. the usage is the same though, as one just pastes the appropriate gmsh string.

florianwechsung:
When I try to use opencascade in an installation that didn't include occ bindings from the start, the error message says that I should run `firedrake-update --opencascade`. However, that doesn't seem to work:

```
ImportError: To use OpenCascadeMeshHierarchy, you must install firedrake with the OpenCascade python bindings (firedrake-update --opencascade).
-------------------------------------------------------
Primary job  terminated normally, but 1 process returned
a non-zero exit code.. Per user-direction, the job has been aborted.
-------------------------------------------------------
--------------------------------------------------------------------------
mpiexec detected that one or more processes exited with non-zero status, thus causing
the job to be terminated. The first process to do so was:

  Process name: [[20642,1],0]
  Exit code:    1
--------------------------------------------------------------------------
(firedrake-dev-20190711) colossus:pipe% firedrake-update --opencascade
Running /home/wechsung/bin/firedrake-dev-20190711/bin/firedrake-update --opencascade
usage: firedrake-update [-h] [--no-update-script] [--rebuild]
                        [--rebuild-script] [--slepc]
                        [--honour-petsc-dir | --petsc-int-type {int32,int64}]
                        [--honour-pythonpath] [--clean] [--verbose]
                        [--install PACKAGES] [--install-help]
                        [--cache-dir CACHE_DIR] [--documentation-dependencies]
firedrake-update: error: unrecognized arguments: --opencascade
```

wence-:
Current position, since Mac build is a disaster (yes?), please add, in firedrake-install/update a check that if you asked for `--download-opencascade` and are running on mac that "sorry, please contact us to get this working".

pefarrell:
I've made the change @wence- has suggested to the install script. It now quits if `--opencascade` is passed on OSX.

pefarrell:
Any further changes required?

firedrake/mg/opencascade_mh.py
```diff
@@ -0,0 +1,250 @@
...
+    coarse = Mesh(mshpath, distribution_parameters=distribution_parameters, comm=comm)
+    return coarse
+
+
+def push_coordinates_to_plex(mesh):
+    plex = mesh._plex
+    x = mesh.coordinates
+    with x.dat.vec_ro as x_:
+        plex.setCoordinates(x_)
+        plex.setCoordinatesLocal(x_)
```
wence-: Two things here:

1. `x_` is a global vector, so you only need to do `plex.setCoordinates` (it has some invalidation mechanism to ensure that the coordinatesLocal are reconstructed later) I think

2. The data layout is, I think, not the same for plex's view of the coordinates and ours (they have differently numbered sections (AKA cell_node_map)), so I don't know how this works!

scripts/firedrake-install
```diff
@@ -1049,6 +1080,9 @@ if mode == "install" or not args.update_script:
             brew_install("libtool")
             brew_install("mercurial")
             brew_install("boost")
+            if args.opencascade:
+                check_call(["brew", "tap", "brewsci/science"])
+                brew_install("oce")
```
wence-: This dead code should go.

firedrake/mg/opencascade_mh.py
```diff
@@ -0,0 +1,250 @@
...
+    project_to_cad = project_mesh_to_cad_2d if dim == 2 else project_mesh_to_cad_3d
+
+    if project_refinements_to_cad:
+        for mesh in mh:
+            project_to_cad(mesh, cad)
+            push_coordinates_to_plex(mesh)
+        mh.nested = False
+
+    if order > 1:
+        VFS = VectorFunctionSpace
```
wence-: Does this now work in parallel, or did the assertion of https://github.com/firedrakeproject/firedrake/pull/1448/commits/cfba05a86070faa78089e60e53bb434a209cfb97 get lost somewhere?

wence-: Ah, I see that https://github.com/firedrakeproject/firedrake/pull/1448/commits/ee5f9eeea285e0468a8850f9d4fce469f8ff6226 fixes it?

firedrake/mg/opencascade_mh.py
pefarrell: You're right, I can delete the call to `setCoordinatesLocal` and the tests still pass. I've committed this.

firedrake/mg/opencascade_mh.py
pefarrell: Yes, high-order construction now works in parallel and is tested in `tests/multigrid/test_opencascade_volume.py:test_area`.

firedrake/mg/opencascade_mh.py
wence-: Do you have any insight into point 2?

wence-: (I mean, clearly it does work, but I wonder if it does so by accident, magic, or something else).

pefarrell: Alas, I'm afraid I don't. @florianwechsung, do you have any insight?

florianwechsung: Mh no I have no idea. In fact, are we sure it DTRT? Where are the coordinates in plex used apart from when calling refine (which happens before this)? All the assembly and transfers use the coordinates specified in `mesh.coordinates`, right?

wence-: The setup is something like:

1. Plex has topology + geometry (via DMGetCoordinateDM)
2. When we build a `Mesh` we share (effectively) the topology and build a new geometry, by creating a VFS and _copying_ the coordinates from the plex into the VFS (via a permutation map)
3. Now we ignore the plex coordinates.

When plex refines, it uses the coordinates it knows about the its geometry. These are values (in a Vec) plus data layout (in a section).

When this code presently does `push_coords_to_plex`, it updates the values (with what Firedrake knows), but _not_ the data layout. I think if we want to replace the data layout too, we need to do:

```python
sec = new_coords.function_space().dm.getDefaultSection()
plex_sec = PETSc.Section().create(comm=new_coords.comm)
plex_sec.setNumFields(1)
plex_sec.setFieldComponents(gdim)
plex_sec.setPermutation(mesh._plex_renumbering)
for p in range(*sec.getChart()):
    dof = sec.getDof(p)
    plex_sec.setDof(p, dof*gdim)
    plex_sec.setFieldDof(p, 0, dof*gdim)
plex_sec.setUp()
plex.getCoordinateDM().setDefaultSection(plex_sec)
```
plex.setCoordinatesLocal(

wence-: untested, not even sure if all of those petsc4py functions exist.

pefarrell: I've deleted the `push_coordinates_to_plex` code, as discussed.

wence-:
I have fixed the merge conflicts in firedrake-install. I think this is now good to go, and will wait for Jenkins to agree.

PR #1449, opened 2019-06-05
Updating Tests to use inner(u, v)

benSepanski:
Primarily, these are edits in `firedrake/tests` so that each `TestFunction` is conjugated (usually by using inner). I also changed a few of the tests which used `conditional` by having them take the real part before doing a comparison. (e.g. `real(x[0]) < 0.2`).

There are also a few changes made in `firedrake/firedrake` that are either minor bug fixes or things lint complained about when I ran `make alltest`.


dham:
Oh wow. Thanks for doing this Ben, I've been trying to find the time for literally the last year (I was working on this at FEniCS '18 in June 2018)! This is amazing.

benSepanski:
Thanks! Dr. Kirby mentioned to me it might be better to update the tests to use inner( , ) on master, then merge them into complex from there. If you want, I can migrate the relevant changes over to my master branch, and make a pull request there.

dham:
> Thanks! Dr. Kirby mentioned to me it might be better to update the tests to use inner( , ) on master, then merge them into complex from there. If you want, I can migrate the relevant changes over to my master branch, and make a pull request there.

I'm perfectly happy for you to do it from the complex side. We do need to make sure that these tests pass when the complex branch is built in real mode, though.

benSepanski:
How do I build the complex branch in real mode?

dham:
Just don't pass`--complex` to `firedrake-install`.
Alternatively you can run `firedrake-update --real` on a complex firedrake to rebuild in real mode. However since the rebuild rebuilds PETSc, switching back and forward takes quite a time so you might choose to have two firedrake complex branch installs, one in real mode and one in complex.

benSepanski:
Several tests are failing because they are calling `project` on a `"DG"` FunctionSpace. Most of these pass if I set `use_slate_for_inverse=False`. Is that the right thing to do, or will that error eventually be fixed?

dham:
Yes, that’s a reasonable change to make. Slate doesn’t work in complex yet (Thomas is in Durham this week working on some necessary changes before that can happen). Switching to the non-slate inverse is a perfectly sensible intermediate step and it will be easy for us to back out once slate supports complex. Thanks immensely for your work on this, these are things I have failed to achieve for the last year.

benSepanski:
I'm happy to work on it! I just pushed some more test fixes. There are a few I'm still having trouble with. I'm putting a summary of the problem below (these are based on runs completed on complex firedrake built in real mode).

- `regression/`

    - `test_auxiliary_dm.py` crashes with no output for me.
    - `test_subdomain.py` doesn't terminate (or at least takes >1hr)
    - `test_coordinatederivative.py` ends with a slate error (about including `Eigen/Dense`) that I couldn't figure out how to fix
    - `test_vector.py` has an error about a `petsc4py.PETSc.Scatter` object not having an attribute called `createWithData`
    - `test_nonlinear_stokes_hdiv.py`: Here I had trouble fixing the arity errors. Either there's some easy fix I'm missing, or something about having `inner(u, u)` being added to things like `inner(u, v)` is provoking errors.

- `extrusion/`

    - `test_subdomain_extruded.py` doesn't terminate (or at least takes >1hr)

- `multigrid/`

    - The following tests would crash mid-test with no output: `test_custom_transfer.py`, `test_grid_transfer.py`, `test_embedded_transfer.py`, `test_multi_space_transfer.py`, and `test_nested_split.py`

- `slate/`

    - So... most of the tests fail here. Something felt wrong about trying to get the `slate/` tests to not use slate, so I've mostly left these alone (except for fixing any arity problems I could find).

- `supermesh/`

    - `test_nonnested_project_no_hierarchy.py` doesn't terminate (or at least takes >1hr)

I think that covers everything that's going wrong right now. I'll double check with another big test run, but I'm running on a virtualbox on a laptop so it takes quite awhile to run all the tests.

dham:
To update on this, I've just merged master onto this branch and pushed it as the loopy-complex branch (so as not to immediately break things for anyone using this branch). Once I convince myself that that's not completely broken, I plan to merge that into the complex branch, which will also merge this PR. At that stage, we will have complex loopy and it will no longer be necessary to use old PyOP2 and tsfc branches.

PR #1453, opened 2019-06-14
Interpolation operator

ThomasGregory:


tests/regression/test_interpolate.py
```diff
@@ -174,3 +175,89 @@ def test_lvalue_rvalue():
...
+    P1 = FunctionSpace(mesh, "CG", degree)
+    P2 = FunctionSpace(mesh, "CG", degree + 1)
+
+    expr = x[0]**degree + x[1]**degree
+    x_P1 = interpolate(expr, P1)
+    interpolator = Interpolator(TestFunction(P1), P2)
+    x_P2 = interpolator.interpolate(x_P1)
+    x_P2_direct = interpolate(expr, P2)
+
+    assert np.allclose(x_P2.dat.data, x_P2_direct.dat.data)
```
wence-: Do we need to test such a large range of degrees? Would it make more sense to test fewer degrees
But instead different meshes? e.g. triangles and tets?

tests/regression/test_interpolate.py
```diff
@@ -174,3 +175,89 @@ def test_lvalue_rvalue():
...
+    cg1 = FunctionSpace(mesh, "CG", 1)
+    dg1 = FunctionSpace(mesh, "DG", 1)
+
+    v = TestFunction(dg1)
+    u_cg = assemble(TestFunction(cg1) * dx)
+    interpolator = Interpolator(TestFunction(cg1), dg1)
+    v_adj = interpolator.interpolate(assemble(v * dx), transpose=True)
+
+    assert np.allclose(u_cg.dat.data, v_adj.dat.data)
+
```
wence-: Blank line missing

tests/regression/test_interpolate.py
```diff
@@ -174,3 +175,89 @@ def test_lvalue_rvalue():
     u.assign(1.0)
     u.interpolate(u + 1.0)
     assert np.allclose(u.dat.data_ro, 2.0)
+
+
+@pytest.mark.parametrize("degree", range(1, 6))
+def test_interpolator_spectral(degree):
+    mesh = UnitSquareMesh(10, 10, quadrilateral=True)
+    x = SpatialCoordinate(mesh)
+    fe1 = FiniteElement("CG", mesh.ufl_cell(), degree, variant="spectral")
```
wence-: `"spectral"` is now the default
Do we want to test equispaced and spectral?

firedrake/interpolation.py
```diff
@@ -87,24 +158,32 @@ def make_interpolator(expr, V, subset, access):
...
-        raise ValueError("Don't know how to interpolate a %r" % expr)
+        # Slice the expression and pass in the right number of values for
+        # each component function space of this function
+        d = 0
+        for fs, dat, dim in zip(V, f.dat, dims):
+            idx = d if fs.rank == 0 else slice(d, d+dim)
+            loops.extend(_interpolator(fs, dat,
+                                       SubExpression(expr, idx, fs.ufl_element().value_shape()),
+                                       subset, arguments, access))
+            d += dim
```
wence-: This change should not occur (this branch is dead code since the purging of
C-string expressions.

firedrake/interpolation.py
```diff
@@ -53,26 +58,92 @@ class Interpolator(object):
...
+                        callable.handle.multTranspose(x, out)
+
+            return result
+        else:
+            if output:
+                output.assign(callable)
+                return output
+            if isinstance(self.V, firedrake.Function):
+                self.V.assign(callable)
+                return self.V
```
wence-: One of the points of the interpolator object (for reuse) is to avoid
unnecessary copies of temporary data and par-loops.
This looks to be a regression in that it requires that we write into a "temporary"
and then copy into self.V. Why could we not write into self.V straight away?

firedrake/interpolation.py
```diff
@@ -53,26 +58,92 @@ class Interpolator(object):
...
+                result = output or firedrake.Function(self.V)
+                with function.dat.vec_ro as x:
+                    with result.dat.vec_wo as out:
+                        callable.handle.mult(x, out)
+            else:
+                result = output or firedrake.Function(self.args[0].function_space())
+                with function.dat.vec_ro as x:
+                    with result.dat.vec_wo as out:
+                        callable.handle.multTranspose(x, out)
+
```
wence-: Could be written as
```python
if transpose:
   mul = callable.handle.multTranspose
   V = self.args[0].function_space()
else:
   mul = callable.handle.mult
   V = self.V
result = output or firedrake.Function(V)
with function.dat.vec_ro as x, result.dat.vec_wo as out:
    mul(x, out)
```

firedrake/interpolation.py
```diff
@@ -53,26 +58,92 @@ class Interpolator(object):
...
         :returns: The resulting interpolated :class:`.Function`.
         """
-        return self.callable()
+        if transpose and not self.nargs:
+            raise ValueError("Cannot currently apply transpose interpolation with no argument.")
+        if self.nargs != len(function):
+            raise ValueError("Passed %d Functions to interpolate, expected %d"
+                             % (len(function), self.nargs))
+        try:
+            callable = self.frozen_callable
```
wence-: `callable` is sort of the wrong name for this thing, right?
That would suggest is a function that I can call that produces the result,
but it's actually the result.

firedrake/interpolation.py
```diff
@@ -53,26 +58,92 @@ class Interpolator(object):
...
+           interpolation operator.
+
         :returns: The resulting interpolated :class:`.Function`.
         """
-        return self.callable()
+        if transpose and not self.nargs:
+            raise ValueError("Cannot currently apply transpose interpolation with no argument.")
+        if self.nargs != len(function):
+            raise ValueError("Passed %d Functions to interpolate, expected %d"
+                             % (len(function), self.nargs))
```
wence-: What is going on here? It looks like below you require function to be of length 1

firedrake/interpolation.py
```diff
@@ -53,26 +58,92 @@ class Interpolator(object):
...
+           interpolate.
+        :kwarg output: Optional. A :class:`.Function` to contain the output.
+        :kwarg transpose: Set to true to apply the transpose (adjoint) of the
+           interpolation operator.
+
         :returns: The resulting interpolated :class:`.Function`.
         """
-        return self.callable()
+        if transpose and not self.nargs:
+            raise ValueError("Cannot currently apply transpose interpolation with no argument.")
```
wence-: Can we remove the double negative?

firedrake/interpolation.py
```diff
@@ -53,26 +58,92 @@ class Interpolator(object):
        arguments (such that they won't be collected until the
        :class:`Interpolator` is also collected).
     """
-    def __init__(self, expr, V, subset=None, access=op2.WRITE):
-        self.callable = make_interpolator(expr, V, subset, access)
+    def __init__(self, expr, V, subset=None, freeze_expr=False, access=op2.WRITE):
+        self.callable, args = make_interpolator(expr, V, subset, access)
+        self.args = args
```
wence-: Can we call these `arguments`?

tests/regression/test_interpolate.py
ThomasGregory: Increased variety of meshes being tested, maximum degree tested has been reduced on some of the tests.

tests/regression/test_interpolate.py
ThomasGregory: Added

tests/regression/test_interpolate.py
ThomasGregory: Equispaced now included on quads and triangles.

firedrake/interpolation.py
dham: Resolved

firedrake/interpolation.py
dham: No. Sometimes the required number of functions is 0

firedrake/interpolation.py
dham: This copy only occurs in the case of a frozen interpolation. In that case, it's the minimal work (the only thing that happens is we copy the frozen value).

thomasgibson:
The build looks good now. What's the status of this? Are there still plans to update this or is it ready to go?

firedrake/interpolation.py
```diff
@@ -137,7 +218,7 @@ def _interpolator(V, dat, expr, subset, access):
             raise NotImplementedError("Interpolation onto another mesh not supported.")
         if expr.ufl_shape != V.shape:
             raise ValueError("UFL expression has incorrect shape for interpolation.")
-        ast, oriented, needs_cell_sizes, coefficients, _ = compile_ufl_kernel(expr, to_pts, coords, coffee=False)
+        ast, oriented, needs_cell_sizes, coefficients, _ = compile_ufl_kernel(expr, to_pts, V.ufl_element(), coords, coffee=False)
```
thomasgibson: Based on what @miklos1 pointed out in https://github.com/firedrakeproject/tsfc/pull/192, the additional argument `V.ufl_element()` can be removed? Since this is getting ignored.

PR #1462, opened 2019-06-21
Fix R tensorproduct VectorFunctionSpace

tkarna:
Fixes VectorFunctionSpace for R tensorproduct case.

### DIFF
--- a/firedrake/functionspace.py
+++ b/firedrake/functionspace.py
@@ -123,8 +123,11 @@ def FunctionSpace(mesh, family, degree=None, name=None, vfamily=None,
 
     # Support foo x Real tensorproduct elements
     real_tensorproduct = False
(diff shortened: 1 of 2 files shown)


firedrake/functionspace.py
```diff
@@ -123,8 +123,11 @@ def FunctionSpace(mesh, family, degree=None, name=None, vfamily=None,
 
     # Support foo x Real tensorproduct elements
     real_tensorproduct = False
-    if isinstance(element, ufl.TensorProductElement):
-        a, b = element.sub_elements()
+    scalar_element = element
+    if isinstance(element, ufl.VectorElement):
```
wence-: I guess you could also allow it for `ufl.TensorElement`?

tkarna: Sure, done.

PR #1463, opened 2019-06-21
L2pullbacks

celdred:
This adds proper L2 pullbacks (ie integral-preserving, designed for n-forms), by defining new elements at the UFL level and introducing an "L2 Piola" mapping. These elements then share the same FIAT/FiNAT classes as the original elements, since from the perspective of the tabulator the pullback doesn't matter.

Requires https://github.com/firedrakeproject/tsfc/pull/194 and https://github.com/FEniCS/ufl/pull/4.

### DIFF
--- a/tests/regression/test_l2pullbacks.py
+++ b/tests/regression/test_l2pullbacks.py
@@ -1,0 +1,242 @@
+import pytest
+import numpy as np
+from firedrake import *
+
+
+def get_complex(family, hdegree, vdegree=None):
+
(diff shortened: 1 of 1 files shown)


wence-:
I've opened the UFL PR: https://bitbucket.org/fenics-project/ufl/pull-requests/114

miklos1:
The two PRs this depends on are already merged.

wence-:
I just squashed everything, let's see that it runs against current master.

PR #1466, opened 2019-06-25
functionspace: Allow TensorFunctionSpace on vector/tensor-valued things

wence-:


wence-:
Needs https://bitbucket.org/fenics-project/ufl/pull-requests/113/pullback-tensor-vector-valued/diff to actually allow anything to work.

PR #1469, opened 2019-06-27
Create DQ elements instead of DG elements in a few places.

pefarrell:
Firedrake makes some DG spaces internally in a few places. If one is using quads, these lead to warnings about creating DQ elements instead, even though it's not one's fault. This pull request changes some of these constructions to build DQ elements if appropriate.

### DIFF
--- a/firedrake/mg/embedded.py
+++ b/firedrake/mg/embedded.py
@@ -56,14 +56,15 @@ class SingleTransfer(object):
     def __init__(self, element, use_fortin_interpolation=True):
         degree = element.degree()
         cell = element.cell()
+        family = "DG" if cell.is_simplex() else "DQ"
         shape = element.value_shape()
         if len(shape) == 0:
-            DG = ufl.FiniteElement("DG", cell, degree)
+            DG = ufl.FiniteElement(family, cell, degree)
         elif len(shape) == 1:
             shape, = shape
-            DG = ufl.VectorElement("DG", cell, degree, dim=shape)
+            DG = ufl.VectorElement(family, cell, degree, dim=shape)
         else:
-            DG = ufl.TensorElement("DG", cell, degree, shape=shape)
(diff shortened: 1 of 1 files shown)


firedrake/mg/embedded.py
```diff
@@ -56,14 +56,15 @@ def __new__(cls, element, use_fortin_interpolation=True):
     def __init__(self, element, use_fortin_interpolation=True):
         degree = element.degree()
         cell = element.cell()
+        family = "DQ" if cell in (ufl.quadrilateral, ufl.hexahedron) else "DG"
```
dham: Does this actually do the right thing in 3D? Won't the cell in 3D be a tensor product cell? I suspect slightly more sophisticated logic is needed.

pefarrell: OK, I've restricted these changes to 2D. (I'm not familiar enough with the tensor product stuff in firedrake to make the right changes, as I've never used it.)

wence-:
To fix this, we should implement `UFLCell.is_hypercube` to determine whether or not to use `DQ`

firedrake/output.py
```diff
@@ -357,7 +357,10 @@ def _prepare_output(self, function, cg):
         if cg:
             family = "Lagrange"
         else:
-            family = "Discontinuous Lagrange"
+            if function.function_space().mesh().ufl_cell() in (ufl.quadrilateral,):
+                family = "DQ"
+            else:
+                family = "Discontinuous Lagrange"
```
wence-: It turns out the right thing to do is use "DG" if the cell is a simplex, otherwise "DQ" (UFL handles that case correctly). Can you update accordingly? You can kill the output.py one, because I think I fixed that when merging #1499.

pefarrell: Great, thanks for the pointer. Done.

PR #1470, opened 2019-06-28
Mimetic Spectral Elements

celdred:
This pull request adds mimetic spectral elements (described at https://www.sciencedirect.com/science/article/pii/S0021999113006414 and https://arxiv.org/abs/1111.4304), both the primal and dual complexes. This is done by adding the relevant 1D H1 element for the dual complex: Extended Gauss Legendre (EGL); and the L2 edge elements associated with GLL and EGL. These are non-Ciarlet finite elements that use a basis that histopolates (http://people.math.sfu.ca/~nrobidou/public_html/prints/histogram/histogram.pdf) rather than interpolates. The nD deRham complex on hypercubes is then accessible using the variant keyword (either “mse” or “dualmse”) with the relevant elements (DQ, Q, RTCF, etc.).

It requires https://github.com/FEniCS/ufl/pull/5 , https://github.com/FEniCS/fiat/pull/29, https://github.com/FInAT/FInAT/pull/54 and https://github.com/firedrakeproject/tsfc/pull/195.

It also requires correct (integral-preserving) L2 pullbacks, which are implemented at https://github.com/firedrakeproject/firedrake/pull/1463.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -26,7 +26,7 @@ pipeline {
         sh 'mkdir tmp'
         dir('tmp') {
           timestamps {
-            sh '../scripts/firedrake-install --package-branch ufl l2pullbacks --package-branch tsfc l2pullbacks --disable-ssh --minimal-petsc --slepc --documentation-dependencies --install thetis --install gusto --install icepack --install pyadjoint --no-package-manager || (cat firedrake-install.log && /bin/false)'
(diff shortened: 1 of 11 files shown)


celdred:
It is a heavy test load- they take approximately 2 minutes on my laptop with a hot cache. A cold cache (like the testing environment will have) is longer. I wanted to be comprehensive and make sure everything was working properly.

I am happy to reduce the # of tests to 1-2 choices of finite element degree per type of problem (helmholtz mse, hemholtz mixed mse, etc.) and/or reduce the number of types of problems.

Would it be possible to leave the others just commented out (rather than deleted) so if needed they can be easily restored and run? Or is there a better mechanism in pytest to mark optional testing (that can maybe be enabled by a command-line argument)?

wence-:
Please don't comment them out. Please instead mark the extra tests with a `pytest.mark` (you can do this on parameters too for parameterised tests). And then in `tests/conftest.py` we by default exclude those tests with a given mark.

dham:
This seems unlikely to land.

PR #1471, opened 2019-07-05
solving: Default to LU + MUMPS where possible

wence-:
This makes the default user experience friendlier.

### DIFF
--- a/firedrake/solving_utils.py
+++ b/firedrake/solving_utils.py
@@ -1,13 +1,15 @@
 import numpy
 
-import itertools
+from itertools import chain
 
 from pyop2 import op2
+from firedrake_configuration import get_config
 from firedrake import function, dmhooks
 from firedrake.exceptions import ConvergenceError
 from firedrake.petsc import PETSc
 from firedrake.formmanipulation import ExtractSubBlock
 from firedrake.utils import cached_property
+from firedrake.logging import warning
 
 
 def _make_reasons(reasons):
@@ -19,6 +21,101 @@ KSPReasons = _make_reasons(PETSc.KSP.ConvergedReason())
 
 
 SNESReasons = _make_reasons(PETSc.SNES.ConvergedReason())
+
+
+if get_config()["options"]["petsc_int_type"] == "int32":
+    DEFAULT_KSP_PARAMETERS = {"mat_type": "aij",
+                              "ksp_type": "preonly",
+                              "ksp_rtol": 1e-7,
+                              "pc_type": "lu",
+                              "pc_factor_mat_solver_type": "mumps",
+                              "mat_mumps_icntl_14": 200}
+else:
+    DEFAULT_KSP_PARAMETERS = {"mat_type": "aij",
+                              "ksp_type": "preonly",
+                              "ksp_rtol": 1e-7,
+                              "pc_type": "lu",
+                              "pc_factor_mat_solver_type": "superlu_dist"}
+
+
+def set_defaults(solver_parameters, arguments, *, ksp_defaults={}, snes_defaults={}):
+    """Set defaults for solver parameters.
+
+    :arg solver_parameters: dict of user solver parameters to override/extend defaults
+    :arg arguments: arguments for the bilinear form (need to know if we have a Real block).
+    :arg ksp_defaults: Default KSP parameters.
+    :arg snes_defaults: Default SNES parameters."""
+    if solver_parameters:
+        # User configured something, try and set sensible direct solve
+        # defaults for missing bits.
+        parameters = solver_parameters.copy()
+        for k, v in snes_defaults.items():
+            parameters.setdefault(k, v)
+        keys = frozenset(parameters.keys())
+        ksp_defaults = ksp_defaults.copy()
+        skip = set()
+        if "pc_type" in keys:
+            skip.update({"pc_factor_mat_solver_type", "mat_mumps_icntl_14",
+                         # Might reasonably expect to get petsc default
+                         "ksp_type"})
+        if parameters.get("mat_type") in {"matfree", "nest"}:
+            # Non-LU defaults.
+            ksp_defaults["ksp_type"] = "gmres"
+            ksp_defaults["pc_type"] = "jacobi"
+        for k, v in ksp_defaults.items():
+            if k not in skip:
+                parameters.setdefault(k, v)
+        return parameters
+
+    # OK, we're in full defaults mode now.
+    parameters = dict(chain.from_iterable(
+        d.items() for d in (ksp_defaults, snes_defaults)))
+
+    if any(V.ufl_element().family() == "Real"
+           for a in arguments for V in a.function_space()):
+        test, trial = arguments
+        if test.function_space() != trial.function_space():
+            # Don't know what to do here. How did it happen?
+            raise ValueError("Can't generate defaults for non-square problems with real blocks")
+
+        fields = []
+        reals = []
+        for i, V_ in enumerate(test.function_space()):
+            if V_.ufl_element().family() == "Real":
+                reals.append(i)
+            else:
+                fields.append(i)
+        if len(fields) == 0:
+            # Just reals, GMRES
+            opts = {"ksp_type": "gmres",
+                    "pc_type": "none"}
+            parameters.update(opts)
+        else:
+            warning("Real block detected, generating Schur complement elimination PC")
+            # Full Schur complement eliminating onto Real blocks.
+            opts = {"mat_type": "matfree",
+                    "ksp_type": "fgmres",
+                    "pc_type": "fieldsplit",
+                    "pc_fieldsplit_type": "schur",
+                    "pc_fieldsplit_schur_fact_type": "full",
+                    "pc_fieldsplit_0_fields": ",".join(map(str, fields)),
+                    "pc_fieldsplit_1_fields": ",".join(map(str, reals)),
+                    "fieldsplit_0": {
+                        "ksp_type": "preonly",
+                        "pc_type": "python",
+                        "pc_python_type": "firedrake.AssembledPC",
+                        "assembled": DEFAULT_KSP_PARAMETERS,
+                    },
+                    "fieldsplit_1": {
+                        "ksp_type": "gmres",
+                        "pc_type": "none",
+                    }}
+            parameters.update(opts)
+        return parameters
+    else:
+        # We can assemble an AIJ matrix, factor it with a sparse
+        # direct method.
+        return parameters
 
 
 def check_snes_convergence(snes):
@@ -390,7 +487,7 @@ class _SNESContext(object):
             manager = dmhooks.get_transfer_manager(fine._x.function_space().dm)
             manager.inject(fine._x, ctx._x)
 
-            for bc in itertools.chain(*ctx._problem.bcs):
+            for bc in chain(*ctx._problem.bcs):
                 if isinstance(bc, DirichletBC):
                     bc.apply(ctx._x)
 
--- a/firedrake/variational_solver.py
+++ b/firedrake/variational_solver.py
@@ -279,7 +279,7 @@ class LinearVariationalProblem(NonlinearVariationalProblem):
 
     def __init__(self, a, L, u, bcs=None, aP=None,
                  form_compiler_parameters=None,
-                 constant_jacobian=True):
+                 constant_jacobian=False):
         r"""
         :param a: the bilinear form
         :param L: the linear form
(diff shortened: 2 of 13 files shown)


firedrake/solving_utils.py
```diff
@@ -21,6 +23,41 @@ def _make_reasons(reasons):
...
 
 
+if get_config()["options"]["petsc_int_type"] == "int32":
+    DEFAULT_KSP_PARAMETERS = {"mat_type": "aij",
+                              "ksp_type": "preonly",
+                              "pc_type": "lu",
+                              "pc_factor_mat_solver_type": "mumps"}
+else:
+    # FIXME: which factorisation package to use in int64?
+    DEFAULT_KSP_PARAMETERS = {}
```
pefarrell: The MUMPS manual suggests it can be built in fully 64-bit mode, and in "partial" 64-bit mode (where only things that might be big are 64-bit integers). Does PETSc not support it?

firedrake/solving_utils.py
```diff
@@ -21,6 +23,41 @@ def _make_reasons(reasons):
...
+    """Set defaults for solver parameters.
+
+    :arg solver_parameters: dict of user solver parameters to override/extend defaults
+    :arg arguments: arguments for the bilinear form (need to know if we have a Real block).
+    :arg defaults: Any default sets of parameters to supply."""
+    parameters = dict(chain.from_iterable(d.items() for d in defaults))
+    if any(isinstance(V, fsi.RealFunctionSpace)
+           for V in chain.from_iterable(a.function_space() for a in arguments)):
+        # There's a Real block, override with defaults with nest + jacobi.
+        parameters.update({"pc_type": "jacobi", "mat_type": "nest"})
```
pefarrell: I think in this case we should warn the user and generate a set of options that takes a full Schur fieldsplit onto the R-blocks and uses LU for the remainder (à la the Moore-Spence tests in firedrake). This is the most robust solver we can automatically pick, in the absence of assembled R blocks.

firedrake/solving_utils.py
```diff
@@ -21,6 +23,41 @@ def _make_reasons(reasons):
...
+    solver_parameters = solver_parameters or {}
+    mat_type = solver_parameters.get("mat_type")
+    pmat_type = solver_parameters.get("pmat_type")
+    # Allow anything, interpret "matfree" as matrix_free.
+    matfree = mat_type == "matfree"
+    pmatfree = pmat_type == "matfree"
+    # No preconditioner by default for matrix-free
+    if matfree or pmatfree:
+        parameters.update({"pc_type": "none"})
```
pefarrell: Should we automatically set `ksp_type: "fgmres"` and `"ksp_monitor_true_residual": None`, too? It will default to `preonly/none` if the user just passes `{"mat_type": "matfree"}`, which probably is undesirable.

firedrake/variational_solver.py
```diff
@@ -95,7 +95,18 @@ def dm(self):
 class NonlinearVariationalSolver(OptionsManager):
     r"""Solves a :class:`NonlinearVariationalProblem`."""
 
-    def __init__(self, problem, **kwargs):
+    DEFAULT_SNES_PARAMETERS = {"snes_type": "newtonls"}
```
pefarrell: I think we should also default to `"snes_monitor": None` (in case it's not the default) and `"snes_linesearch_type": "basic"` (for compatibility with FEniCS; the PETSc `"bt"` default breaks lots of codes that otherwise would work).

firedrake/solving_utils.py
wence-: PETSc buildsystem says:
```python
    # Mumps does NOT work with 64 bit integers without a huge number of hacks we ain't making
    self.precisions       = ['single','double']
    self.requires32bitint = 1;  # 1 means that the package will not work with 64 bit integers
```

firedrake/solving_utils.py
wence-: True. I realised when doing this in general that setting up defaults for all the different cases is somewhat fiddly. I'm inclined to do:

1. No options, no real blocks -> LU + preonly
2. No options, real blocks -> either Jacobi or auto-generate fieldsplit options + LU
3. Anything else: no defaults, user knows what they're doing.

firedrake/variational_solver.py
wence-: I'm happy to add the linesearch change. The monitoring is a more controversial one: PETSc doesn't turn on monitors by default, although FEniCS does, Firedrake presently doesn't, and this may annoy existing users. Additionally defaulting to `"snes_monitor": None` has the unfortunate side-effect that users will struggle to turn it off: they need to remove it from the dictionary somehow, but if we're shoving it in via defaults...

firedrake/solving_utils.py
pefarrell: Fair enough. I guess the next best choice would be `superlu_dist`, but that would require changing the default PETSc configure options too.

firedrake/solving_utils.py
pefarrell: I agree (with choice fieldsplit + LU for point 2).

firedrake/variational_solver.py
pefarrell: I think they can turn it off with `"snes_monitor_cancel": None`?

pefarrell: Oh, that only works if the monitor has been set up in code, not via options. (Maybe that's the way to go, then.)

firedrake/solving_utils.py
dham: Can we detect the fieldsplit reliably enough? E.g. how do we determine which block to eliminate?

dham: If you have a mixed system then even in 2D LU can get big fast and/or defeat the MUMPS memory management heuristics. Is this really a robust option?

colinjcotter:
Also isn't fieldsplit plus LU on the blocks only marginally better than unpreconditioned GMRES for e.g. mixed Poisson?

This feels like a fool's errand ...

wence-:
> Also isn't fieldsplit plus LU on the blocks only marginally better than unpreconditioned GMRES for e.g. mixed Poisson? This feels like a fool's errand ...

The proposal is:

1. I can assemble an AIJ matrix -> use LU

2. I can't assemble an AIJ matrix (Real constraints) -> use fieldsplit + LU on the bit I can assemble + sherman-morrison.

firedrake/solving_utils.py
wence-: I'm happy to add superlu_dist to the default set of installed packages (it knows about malloc!). In terms of "big", 2D problems with O(1-10e6) dofs are quite happily solved in reasonable time using MUMPs on a workstation.

In terms of defaults, note that in all our tutorials we explicitly say "use LU until you have your numerics/model right and then think about preconditioning". But then we don't default to that option, so you have to learn a load of magic incantations first.

pefarrell:
Yes, if you have n real variables (n small, typically n = 1) then full Schur fieldsplit + LU for (1, 1) block + n iterations of GMRES on the Schur complement solves the problem exactly (in exact arithmetic).

pefarrell:
> If you have a mixed system then even in 2D LU can get big fast and/or defeat the MUMPS memory management heuristics. Is this really a robust option?

David makes a good point. I would suggest also defaulting to `"mat_mumps_icntl_14": 200` (or similar).

rckirby:
I also have frequent headaches trying to use mumps.  Maybe we dodge then if
we we can use the default LU (maybe with a special parameter to keep it
from breaking on saddle points) in serial processor mode -- that's usually
not terrible.

Is there any experience with pastix?  Lawrence and I used it in a paper to
some effect, but I'm not sure how robust it is across a range of problems.

wence-:
OK, so which of these best describes the conclusion of these discussions?

1. Yes, do this (with the better mumps options + fieldsplit/LU for problems with real blocks)
2. Yes, do this, but use superlu_dist instead of mumps (malloc! 64bit!)
3. No, continue with the current state of play.

colinjcotter:
I think that for mixed problems we shouldn’t try to provide an LU default (except for R).

On 10 Jul 2019, at 16:58, Lawrence Mitchell <notifications@github.com<mailto:notifications@github.com>> wrote:


OK, so which of these best describes the conclusion of these discussions?

  1.  Yes, do this (with the better mumps options + fieldsplit/LU for problems with real blocks)
  2.  Yes, do this, but use superlu_dist instead of mumps (malloc! 64bit!)
  3.  No, continue with the current state of play.

pefarrell:
I vote for option #1. With `"mat_mumps_icntl_14": 200` MUMPS is almost bulletproof.

@colinjcotter No one is advocating fieldsplits when not necessary (they are necessary with real blocks because of technical limitations).

rckirby:
If I don’t have to remember that mumps option and it’s robust, it’s fine with me.

firedrake/solving_utils.py
wence-: #1474 changes petsc defaults and installs superlu_dist by default, so if that lands, I will do that here.

firedrake/solving_utils.py
wence-: I've changed to (hopefully) do the 3-part proposal above (with fieldsplit + Schur elimination for problems with Real blocks).

firedrake/solving_utils.py
```diff
@@ -21,6 +24,78 @@ def _make_reasons(reasons):
...
+                    "fieldsplit_0": {
+                        "ksp_type": "preonly",
+                        "pc_type": "python",
+                        "pc_python_type": "firedrake.AssembledPC",
+                        "assembled": DEFAULT_KSP_PARAMETERS,
+                    },
+                    "fieldsplit_1": {
+                        "ksp_type": "gmres",
+                        "pc_type": "jacobi",
+                    }}
```
wence-: @pefarrell does this look right?

firedrake/variational_solver.py
```diff
@@ -95,7 +95,19 @@ def dm(self):
 class NonlinearVariationalSolver(OptionsManager):
     r"""Solves a :class:`NonlinearVariationalProblem`."""
 
-    def __init__(self, problem, **kwargs):
+    DEFAULT_SNES_PARAMETERS = {"snes_type": "newtonls",
+                               "snes_linesearch_type": "basic"}
```
wence-: Added basic linesearch type.

wence-:
If we're happy with the proposal of options, I think this is now right. If not, I'll leave it until agreement is reached on what we want.

firedrake/solving_utils.py
pefarrell: No, it wasn't quite right, for two reasons:

a) The catch for checking if any function spaces involved were R0 was wrong (wasn't tripping even when R0 was present)
b) Jacobi needs `MatGetDiagonal`, which isn't implemented.

I changed the condition to look at the element family, which now works, and changed `"jacobi"` for `"none"`.

I've changed `tests/regression/test_moore_spence.py` (which solves an augmented system of PDEs for a bifurcation point, which is a real value) to use the default solver parameters, and it now passes.

pefarrell:
It would be good to have a test that just solved a problem with R-blocks.

wence-:
I think I finally fixed all the corner cases. Let's hope tests pass.

wence-:
Ready for review. All tests are passing now.

firedrake/variational_solver.py
```diff
@@ -321,30 +316,9 @@ def __init__(self, a, L, u, bcs=None, aP=None,
...
-               space of the operator.
-        :kwarg transpose_nullspace: as for the nullspace, but used to
-               make the right hand side consistent.
-        :kwarg options_prefix: an optional prefix used to distinguish
-               PETSc options.  If not provided a unique prefix will be
-               created.  Use this option if you want to pass options
-               to the solver from the command line in addition to
-               through the ``solver_parameters`` dict.
-        :kwarg appctx: A dictionary containing application context that
-               is passed to the preconditioner if matrix-free.
```
wence-, line 340: Reinstate docstring part of this as class docstring.

firedrake/solving_utils.py
```diff
@@ -21,6 +23,101 @@ def _make_reasons(reasons):
...
+                              "pc_type": "lu",
+                              "pc_factor_mat_solver_type": "mumps",
+                              "mat_mumps_icntl_14": 200}
+else:
+    DEFAULT_KSP_PARAMETERS = {"mat_type": "aij",
+                              "ksp_type": "preonly",
+                              "ksp_rtol": 1e-7,
+                              "pc_type": "lu",
+                              "pc_factor_mat_solver_type": "superlu_dist"}
+
```
ksagiyam, line 39: I remember we discussed this in the meeting, but I forget. Why do we use `superlu` here instead of `mumps`?

wence-: mumps doesn't have a 64bit mode.

PR #1474, opened 2019-07-09
mesh: Allow a local partition to have no cells

wence-:
PETSc has changed the default partitioner to Parmetis, which produces
slightly worse partitions for small meshes. As a result, sometimes in
the tests we get meshes with no local cells.

### DIFF
--- a/firedrake/mesh.py
+++ b/firedrake/mesh.py
@@ -9,6 +9,7 @@ from ufl.classes import ReferenceGrad
 import enum
 import numbers
 
+from mpi4py import MPI
 from pyop2.datatypes import IntType
 from pyop2 import op2
 from pyop2.base import DataSet
@@ -418,13 +419,18 @@ class MeshTopology(object):
 
         dim = plex.getDimension()
 
+        # Allow empty local meshes on a process
         cStart, cEnd = plex.getHeightStratum(0)  # cells
         if cStart == cEnd:
-            raise RuntimeError("Mesh must have at least one cell on every process")
-        cell_nfacets = plex.getConeSize(cStart)
+            nfacets = -1
+        else:
+            nfacets = plex.getConeSize(cStart)
+
+        # TODO: this needs to be updated for mixed-cell meshes.
+        nfacets = self.comm.allreduce(nfacets, op=MPI.MAX)
 
         self._grown_halos = False
-        self._ufl_cell = ufl.Cell(_cells[dim][cell_nfacets])
+        self._ufl_cell = ufl.Cell(_cells[dim][nfacets])
 
         # A set of weakrefs to meshes that are explicitly labelled as being
         # parallel-compatible for interpolation/projection/supermeshing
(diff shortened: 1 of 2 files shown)


wence-:
The other option would be to explicitly override the PETSc defaults and use CHACO. But that is not available for 64bit ints anyway.

dham:
ParMetis is rather seriously not free software. It's free for educational and government use only. Do we want that to be a default?

ksagiyam:
The changes look nice!

wence-:
> ParMetis is rather seriously not free software. It's free for educational and government use only. Do we want that to be a default?

So we presently (in firedrake-install) always do `--download-parmetis/--download-metis`. Possibly because MUMPS needs them (or perhaps it can use ptscotch).

https://github.com/firedrakeproject/firedrake/blob/master/scripts/firedrake-install#L450

The PETSc change that provokes the ordering change is here: https://bitbucket.org/petsc/petsc/commits/56bf5a817defc15c3

I can try seeing if removing the parmetis stuff and just using --download-ptscotch will work.

wence-:
Does metis have the same nonsense license, or is it sane?

wence-:
> Does metis have the same nonsense license, or is it sane?

Metis >= 5.0.3 is released under Apache version 2.0, which I think is fine?

wence-:
I've refactored the install mechanism, and default to chaco or scotch (64 bit ints). This will need various changes in petsc4py: https://bitbucket.org/petsc/petsc4py/pull-requests/123/partitioner-add-missing-type-names/diff, and me to update the PETSc pointer (which I am planning on doing anyway, which required the minimal changes).

@pefarrell any other PCs I should add to the "full-fat" petsc?

wence-:
OK, ptscotch needs `bison`: how do I add this to the environment? Presumably I need to edit a dockerfile, but then how does that get updated?

wence-:
The minimal change, for me to be able to update the PETSc pointer, is to just explicitly set the default partitioner type to CHACO. I'm inclined to do that (at least) because there are various updates we would like.

The proposed refactoring of petsc defaults and so forth is a bit larger. In particular, we now explicitly require user-consent (via `--with-parmetis`) to install Parmetis. To install ptscotch on Jenkins requires that the base docker image has `bison` installed, but I don't think I know how to update the image.

firedrake/mesh.py
```diff
@@ -405,8 +406,10 @@ def add_overlap():
             # it, we grow the halo.
             partitioner = plex.getPartitioner()
             if IntType.itemsize == 8:
-                # Default to Parmetis on 64bit ints (Chaco is 32 bit int only)
-                partitioner.setType(partitioner.Type.PARMETIS)
+                # Default to PTSCOTCH on 64bit ints (Chaco is 32 bit int only)
+                partitioner.setType(partitioner.Type.PTSCOTCH)
```
wence-: Pick parmetis in this case if config options say it is available.

wence-:
Addressed comments from meeting. This is now waiting for the docker base image to be rebuilt with bison before trying afresh.

wence-:
Let's hope jenkins is clean now.

wence-:
Now passing (with relevant updated PETSc branches), happy with this?

dham:
Just about happy. One final query - do we also need bison on Mac?

wence-:
Let me check on my new (not yet firedrake-installed) mac later today.

wence-:
(probably)

wence-:
Bison/flex come as part of Xcode command line tools!

dham:
Go for it (obviously fixing Jenkinsfile)

Issue #1467, opened 2019-06-26
Boundary conditions on VectorFunctionSpace fail when using fieldsplit


halawiye:
I'm working on a problem that requires me to apply a `DirichletBC` to one component of a `VectorFunctionSpace` that is in itself a component of a `MixedFunctionSpace`. Due to the nature of the problem, I have to use different preconditioners/solvers for the different blocks, which I have achieved in the past using PETSc's PCFIELDSPLIT. However, when I try to combine these two aspects, firedrake does not seem to be able to correctly reconstruct the function spaces after the fieldsplit, causing various instances of `IndexError`, `MapValueError` and `RuntimeError`.

Here is a minimal example I was able to produce that demonstrates one such problem:
```                                                                                               
 from firedrake import *
 
 mesh = UnitSquareMesh(32,32)
 X = SpatialCoordinate(mesh)
 
 V = VectorFunctionSpace(mesh, "CG", 1)
 VV = MixedFunctionSpace([V,V])
 
 f = interpolate(as_vector([1.0, sin(X[0]*X[1]*2*pi)]), V)
 
 uu = Function(VV)
 vv = TestFunction(VV)
 
 L = 0 
 for u, v in zip(split(uu), split(vv)):
     L += inner(grad(u), grad(v))*dx - inner(f,v)*dx
 
 bcs = [ 
     #DirichletBC(VV.sub(0), Constant((0.0, 0.0)), "on_boundary"),
     DirichletBC(VV.sub(0).sub(0), 0.0, "on_boundary"),
     DirichletBC(VV.sub(0).sub(1), 0.0, "on_boundary"),
     #DirichletBC(VV.sub(1), Constant((0.0, 0.0)), "on_boundary")
     DirichletBC(VV.sub(1).sub(0), 0.0, "on_boundary"),
     DirichletBC(VV.sub(1).sub(1), 0.0, "on_boundary"),
     ]   
 
 sp = { 
     "mat_type" : "matfree",
     "snes_type" : "newtonls",
     "snes_monitor" : None,
     "snes_converged_reason" : None,
     "ksp_type" : "fgmres",
     "ksp_monitor_true_residual" : None,
     "ksp_max_it" : 10, 
     "pc_type" : "fieldsplit",
     "pc_fieldsplit_type": "schur",
     "pc_fieldsplit_schur_fact_type": "full",
     "pc_fieldsplit_0_fields" : "0",
     "pc_fieldsplit_1_fields" : "1",
[...]

PR #1475, opened 2019-07-09
Fix `DirichletBC.reconstruct`

ksagiyam:
This PR fixes issue #1467 regarding application of Dirichlet BCs on V.sub(i).sub(j), by fixing `DirichletBC.reconstruct` method that was returning `self` where it should not.

A relevant test has also been added.

### DIFF
--- a/firedrake/bcs.py
+++ b/firedrake/bcs.py
@@ -319,6 +319,8 @@ class DirichletBC(BCBase):
            V.parent == fs.parent and \
            V.index == fs.index and \
            V.component == fs.component and \
+           (V.parent is None or V.parent.parent == fs.parent.parent) and \
+           (V.parent is None or V.parent.index == fs.parent.index) and \
            g == self._original_arg and \
            sub_domain == self.sub_domain and method == self.method:
             return self
--- a/firedrake/bcs.py
+++ b/firedrake/bcs.py
@@ -318,7 +318,6 @@ class DirichletBC(BCBase):
         if V == fs and \
            V.parent == fs.parent and \
            V.index == fs.index and \
-           V.component == fs.component and \
            (V.parent is None or V.parent.parent == fs.parent.parent) and \
            (V.parent is None or V.parent.index == fs.parent.index) and \
            g == self._original_arg and \
--- a/tests/equation_bcs/test_bcs.py
+++ b/tests/equation_bcs/test_bcs.py
@@ -1,0 +1,64 @@
+import pytest
+
+from firedrake import *
+
+
+@pytest.mark.parallel
+def test_bc_on_sub_sub_domain():
+
+    # Solve a vector poisson problem
+
+    mesh = UnitSquareMesh(32, 32)
+
+    V = VectorFunctionSpace(mesh, "CG", 4)
+    VV = MixedFunctionSpace([V, V])
+
+    x, y = SpatialCoordinate(mesh)
+
+    f = Function(V)
+    f.interpolate(as_vector([-8.0 * pi * pi * cos(x * pi * 2) * cos(y * pi * 2),
+                             -8.0 * pi * pi * cos(x * pi * 2) * cos(y * pi * 2)]))
+
+    # interpolate the exact solution on gg[i][j]
+    gg = [[None, None], [None, None]]
+    for i in [0, 1]:
+        for j in [0, 1]:
+            gg[i][j] = Function(VV.sub(i).sub(j))
+            gg[i][j].interpolate(cos(2 * pi * x) * cos(2 * pi * y))
+
+    uu = Function(VV)
+    vv = TestFunction(VV)
+
+    F = 0
+    for u, v in zip(split(uu), split(vv)):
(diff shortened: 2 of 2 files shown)


tests/equation_bcs/test_bcs.py
```diff
@@ -0,0 +1,64 @@
+import pytest
+
+from firedrake import *
+
+
+@pytest.mark.parallel
```
wence-: Does this test need to be parallel? Prefer not, since they are much slower.

tests/equation_bcs/test_bcs.py
```diff
@@ -0,0 +1,64 @@
...
+def test_bc_on_sub_sub_domain():
+
+    # Solve a vector poisson problem
+
+    mesh = UnitSquareMesh(32, 32)
+
+    V = VectorFunctionSpace(mesh, "CG", 4)
```
wence-: Is degree 4 necessary to provoke the error, or will 1 do?

tests/equation_bcs/test_bcs.py
```diff
@@ -0,0 +1,64 @@
...
+        "ksp_rtol": 1.e-10,
+        "ksp_atol": 1.e-10,
+        "ksp_max_it": 200000,
+        "pc_type": "fieldsplit",
+        "pc_fieldsplit_type": "schur",
+        "pc_fieldsplit_schur_fact_type": "full",
+        "fieldsplit_0_ksp_type": "gmres",
+        "fieldsplit_0_ksp_rtol": 1.e-12,
+        "fieldsplit_1_ksp_type": "gmres",
+        "fieldsplit_1_ksp_rtol": 1.e-12,
```
wence-: To avoid having to do millions of iterations, do:
```python
parameters = {
  "mat_type": "matfree",
  "snes_type": "ksponly",
  "ksp_type": "preonly",
  "pc_type": "fieldsplit",
  "pc_fieldsplit_type": "additive",
  "fieldsplit_ksp_type": "preonly",
  "fieldsplit_pc_type": "python",
  "fieldsplit_pc_python_type": "firedrake.AssembledPC",
  "fieldsplit_assembled_pc_type": "lu"
}
```

I think

firedrake/bcs.py
```diff
@@ -319,6 +319,8 @@ def reconstruct(self, field=None, V=None, g=None, sub_domain=None, method=None,
            V.parent == fs.parent and \
            V.index == fs.index and \
            V.component == fs.component and \
+           (V.parent is None or V.parent.parent == fs.parent.parent) and \
+           (V.parent is None or V.parent.index == fs.parent.index) and \
```
wence-: Thinking about this, this is a more general statement about the equality of function spaces, no?

Should we not be writing this as:
```python
if V == fs and g == ...:
   return self
```
And implement `V == fs` correctly on functionspaces?

ksagiyam: So, `V.component == ...` is nicely removed from here as it has actually been in `FunctionSpace.__eq__` already.
Moving `V.parent ==` causes recursions (`FunctionSpace.__eq__` <->`MixedFunctionSpace.__eq__` ).
Shouldn't `V` still be regarded as equal to `V` even if it is embedded in some `MixedFunctionSpace`, `W`, so that `V == W.sub(0)` holds?
I think the reason we have to check `parents` here is that `DirichletBC._indices` has to be consistent to whether we have `parent` or `parent.parent`, so isn't it fair to leave these checks here?

tests/equation_bcs/test_bcs.py
ksagiyam: No, it does not.  Fixed.

tests/equation_bcs/test_bcs.py
ksagiyam: Fixed! Let's use 1.

tests/equation_bcs/test_bcs.py
ksagiyam: Thanks!

firedrake/bcs.py
wence-: OK, makes sense.

tests/equation_bcs/test_bcs.py
wence-: I don't see these changes.

ksagiyam: Sorry, just pushed.

firedrake/bcs.py
```diff
@@ -328,6 +329,10 @@ def reconstruct(self, field=None, V=None, g=None, sub_domain=None, method=None,
     def function_arg(self, g):
         '''Set the value of this boundary condition.'''
         if isinstance(g, function.Function) and g.function_space() != self._function_space:
+            print(g.function_space())
+            print(self._function_space)
+            import sys
+            sys.stdout.flush()
```
wence-: Remove this debugging output

wence-: Can you squash it?

ksagiyam: Done.

PR #1476, opened 2019-07-11
Fix/cleanup transfer operators

wence-:
# Overview
This changes the way solver application contexts are attached to (and removed from) DMs inside solvers/PCs. It fixes various dangling references and memory leaks where solvers would not get fully collected. It also makes the setup more robust to repeated solver calls.

# Background

When we enter a solver, PETSc might make DMs on field decompositions or coarse DMs for multigrid. These are created by callbacks into Firedrake, which return the appropriate subproblem and transfer the sub-application context. Because these calls might be expensive, they are only run on `SNESSetUp` by PETSc.

Previously, we just left the application contexts hanging around on the subproblems, and it mostly worked, because usually you just use one solver multiple times. But this isn't safe: what if you have multiple solvers on the same space? Moreover, the subproblems are not cleaned up by the Python garbage collector (because petsc4py has custom garbage collection that can't handle cycles).

# New design

Mimicing what happens inside PETSc to some degree, when a subproblem is first created, we remember the call that attaches the sub-application context in a `Hook` object. These hooks are gathered on the ultimate ancestral (outer solver) `DM` (this might not be optimal, but I ran out of brain). When we finish a solve, a set of `teardown` hooks are run, that remove all the application contexts from the various subproblems. When we subsequently run the solve, `setup` hooks are run.

This allows multiple solvers to share the same DM, and means that everything gets nicely cleaned up. When we exit the solve, there are no cycles left lying around.

# Issues

1. This is still not perfect, the code is still rather spaghetti-like, and I should probably go through and completely refactor, but that needs more thinking.

2. I couldn't manage to get hooks to reinject coarse grid coefficients on multiple solves working properly. But this doesn't break existing functionality, since we don't do that properly right now

3. We really should have different DMs for different parts of a mixed space that happen to be the same individual space (related to 2.) but I haven't done that either.

# Summary

This is not yet perfect, but it is a step in a positive direction to making this part of the code cleaner and more robust.

### DIFF
--- a/firedrake/dmhooks.py
+++ b/firedrake/dmhooks.py
@@ -39,6 +39,7 @@ advance.
 
 import weakref
 import numpy
+from functools import partial
 
 import firedrake
 from firedrake.petsc import PETSc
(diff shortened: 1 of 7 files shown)


wence-:
I think this is safe to do, because it makes no regressions the branch https://github.com/firedrakeproject/firedrake/tree/wence/feature-retransfer-coefficients is more complete, but also not quite ready for primetime. Plausibly we should just wait for that and merge it instead.

I need to write some tests that exercise all the corner cases. It would be helpful, if people have some, to just paste them in.

Issue #1477, opened 2019-07-12
setup.py not installing all subpackages


JaroslavHron:
doing direct installation into prefix directory like
`python3 -s setup.py --no-user-cfg install --prefix=/usr/local/pkg/firedrake`
will not install some of the subpackages since some are missing in the `packages` list - there shoud be:
```
setup( ..., 
packages=["firedrake", "firedrake.mg", "firedrake.slope_limiter",
                "firedrake.matrix_free", "firedrake_configuration",
                "firedrake_citations", "firedrake.slate", "firedrake.slate.slac",
                "firedrake.slate.static_condensation", "firedrake.preconditioners"],
...)
```
or with `setuptools` one might use `find_packages()`
```
from setuptools import setup, find_packages

setup( ..., packages=find_packages(), ....)
```

PR #1478, opened 2019-07-12
update setup.py to install all subpackages

JaroslavHron:
all subpackage has to be explicitly mentioned in setup to be installed - see #1477


wence-:
Thanks. This looks right, I think. As an aside, is there a reason that you are doing the install this way (with prefix install) rather than using `firedrake-install`? The latter allows much easier updates of the installation (with `firedrake-update`), and isolates the python environment from any other packages.

JaroslavHron:
this was issue I run into when setting up firedrake installation on our cluster using `spack` -https://spack.io/ - we use it for most of the other software installs, like fenics

it still has some rough corners and needs some cleanup - if anyone is interested see https://github.com/JaroslavHron/spack_repo

wence-:
Thanks, this looks like the right change.

I note that tracking what options we might require for PETSc could be tricky, in #1474 I update firedrake-install show that `firedrake-install --show-petsc-configure-options` just prints the required options and exits.

wence-:
Thanks!

PR #1480, opened 2019-07-24
finalise delayed-bcs deprecation

ksagiyam:
As a consequence of removing `thunk`, passing `bcs` to `solve(A, x, b)` with an assembled linear operator `A` does not make sense any more.  This PR makes the following changes:

-make `_la_solve` throw an error message when this happens
-remove relevant tests
-change documentation

### DIFF
--- a/docs/source/solving-interface.rst
+++ b/docs/source/solving-interface.rst
@@ -179,29 +179,11 @@ not supply them in solve call):
    b = assemble(L)
    solve(A, x, b)
 
-alternately, we can supply boundary conditions in
-:py:func:`~firedrake.solving.solve` as before:
+.. warning::
 
-.. code-block:: python
-
-  A = assemble(a)
-  b = assemble(L)
-  solve(A, x, b, bcs=[bc1, bc2])
-
-If boundary conditions have been supplied both in the assemble and
-solve calls, then those provided for the solve take precedence, for
-example, in the following, the system is solved only applying ``bc1``:
-
-.. code-block:: python
-
-  A = assemble(a, bcs=[bc1, bc2])
-  b = assemble(L)
-  solve(A, x, b, bcs=[bc1])
-
-Note that after the call to solve, ``A`` will be an assembled system
-with only ``bc1`` applied, hence subsequent calls to ``solve`` that do
-not change the boundary conditions again will not require a further
-re-assembly.
+   It is no longer possible to apply or change boundary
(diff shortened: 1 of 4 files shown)


docs/source/solving-interface.rst
```diff
@@ -179,29 +179,11 @@ not supply them in solve call):
...
-  A = assemble(a, bcs=[bc1, bc2])
-  b = assemble(L)
-  solve(A, x, b, bcs=[bc1])
-
-Note that after the call to solve, ``A`` will be an assembled system
-with only ``bc1`` applied, hence subsequent calls to ``solve`` that do
-not change the boundary conditions again will not require a further
-re-assembly.
+   Capability to apply/change boundary conditions after assembling
+   the linear operator ``A`` has been deprecated; pass any
```
dham: Should read: "It is no longer possible to apply or change boundary conditions after assembling the matrix ``A``; pass ..."

firedrake/solving.py
```diff
@@ -192,17 +191,16 @@ def _la_solve(A, x, b, **kwargs):
...
 
         .. code-block:: python
 
            A = assemble(a, bcs=[bc1])
-           solve(A, x, b, bcs=[bc2])
-
-        the boundary conditions in `bc2` will be applied to the problem
-        while `bc1` will be ignored.
+           solve(A, x, b)
+           # solve(A, x, b, bcs=[bc2])  # Deprecated
```
dham: I think we should just remove this line.

firedrake/solving.py
```diff
@@ -213,11 +211,9 @@ def _la_solve(A, x, b, **kwargs):
     bcs, solver_parameters, nullspace, nullspace_T, near_nullspace, \
         options_prefix = _extract_linear_solver_args(A, x, b, **kwargs)
 
-    for bc in _extract_bcs(bcs):
-        if not bc.is_linear:
-            raise RuntimeError("EquationBCs must also be linear when solving linear system.")
     if bcs is not None:
-        A.bcs = bcs
+        raise RuntimeError("Apply `bcs` when assembling the bilinear form `A`.")
+
```
dham: This error is a bit cryptic.

ksagiyam:
Thank you, David.  Those are fixed now.

docs/source/solving-interface.rst
ksagiyam: Thank you! Fixed.

PR #1482, opened 2019-07-25
Notebook updates

dham:
Merge in the updated notebooks we created for CSE. Also create a docker image with the notebooks running.


ksagiyam:
`06-pde-constrained-optimisation.ipynb` (the diff was too large to open...):
In the variational formulation the boundary integral over $\Gamma_{circ}$ should vanish as we have Dirichlet BC there.  
The pressure condition on $\Gamma_{out}$ is different from the one given above for the strong form.  Does the pressure condition make all boundary integral over $\Gamma_{out}$ vanish?

docs/notebooks/10-sum-factorisation.ipynb
```diff
@@ -0,0 +1,2014 @@
...
+   "source": [
+    "# Optimal complexity finite element assembly"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "## Sum factorisation\n",
+    "Suppose we are assemblying the 2-form of the Laplace operator on a hexahedral element, e.g. the Q3 element:"
```
ksagiyam: assembling

Issue #1468, opened 2019-06-26
Crossed meshes not supported on CylinderMesh


ScottMacLachlan:
For symmetry reasons, I'd like to use a crossed triangular mesh on a PartiallyPeriodicRectangleMesh, but this isn't supported in the underlying CylinderMesh.  Is it possible to add this functionality?  Thanks!

dham:
Hi Scott, I'm not sure when I could commit resources to this, but I think I can tell you how to do it if you want to have a go. Basically lines 1257 to 1275 of `utility_meshes.py` hand code the mesh. If you added an option to the mesh constructor, and used it to switch onto a crossed mesh construction at this point, then everything else should just fall through.

The one thing you need to be aware of is that in specifying a mesh cell in firedrake, the vertices should be in ascending order (i.e. not clockwise or anticlockwise, but ascending index order).

ScottMacLachlan:
Thanks, David!

If I'm reading this right, there's two steps:
1) Augment the list of vertices to include the midpoints of the quads (since these will be vertices in the crossed mesh that aren't vertices in the quad mesh).  Probably easiest to append these to the end of the list, just to not break the way that "cells" is formed for the quad mesh on line 1273.
2) Write an alternative to line 1276 that replaces each quad with 4 triangles.  The only magic here would seem to be correctly indexing the right midpoint for each quad.  (I.e., can I use some indexing magic here, or just a loop over the quads?)

Does that make sense?  Am I missing anything?  Thanks!

wence-:
Hi Scott,

Option 2 also needs to produce the extra vertices.

I think you can probably do option 1 by cargoculting the equivalent `diagonal="crossed"` code from `RectangleMesh` (starting at line 353).

pefarrell:
I've pushed an initial attempt in `pefarrell/crossed-cylinder-mesh`.

miklos1:
Closing because #1483 is merged. If this is not correct, feel free to reopen.

PR #1483, opened 2019-08-02
Add support for diagonal="crossed" in CylinderMesh (and hence PartiallyPeriodicRectangleMesh)

pefarrell:
This allows for the retention of symmetry in partially periodic meshes, cf #1468 .

### DIFF
--- a/firedrake/utility_meshes.py
+++ b/firedrake/utility_meshes.py
@@ -331,7 +331,7 @@ def RectangleMesh(nx, ny, Lx, Ly, quadrilateral=False, reorder=None,
         COMM_WORLD).
     :kwarg diagonal: For triangular meshes, should the diagonal got
         from bottom left to top right (``"right"``), or top left to
-        bottom right (``"left"``).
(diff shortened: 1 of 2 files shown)


firedrake/utility_meshes.py
```diff
@@ -1269,11 +1284,51 @@ def CylinderMesh(nr, nl, radius=1, depth=1, longitudinal_direction="z",
...
+        # |  4  |
+        # | / \ |
+        # 0-----1
+
+        offset = np.arange(nl, dtype=np.int32)*nr
+        origquads = np.row_stack(tuple(ring_cells + i for i in offset))
+        cells = np.zeros((origquads.shape[0]*4, 3), dtype=np.int32)
+        cellidx = 0
+        newvertices = range(len(origvertices), len(origvertices) + len(extras))
+        for (origquad, extravertex) in zip(origquads, newvertices):
```
pefarrell: I imagine a fancy vectorised numpy loop would be faster, but I'm not as good a programmer as @wence- :-)

PR #1488, opened 2019-08-14
Bdmc

cyruscycheng21:
This PR introduces the BDMC elements to Firedrake, and is linked to three other PRs in other firedrakeproject packages:
FIAT: https://github.com/firedrakeproject/fiat/pull/18
FInAT: https://github.com/FInAT/FInAT/pull/55
tsfc: https://github.com/firedrakeproject/tsfc/pull/197

The Firedrake component adds three unit tests for the new elements.


pefarrell:
What's the current status of this? BDMCF/DPC works fantastically well for an application of mine and I'd love to see it merged.

wence-:
Was the conclusion that these elements are broken in some way at lowest order?

pefarrell:
Yes. I've added a test for the H(div) and H(curl) Riesz maps. They converge at all orders k=1,2,3 for RTC[F/E], and for k=2,3 for BDMC[F/E], but not for BDMC[F/E] with k=1.

This is the solution chosen for the MMS:

exact

This is the solution for the H(div) Riesz map at k=1:

bdmcf1

@dham suggested it was a local ordering bug.

tests/regression/test_bdmc_riesz_map.py
```diff
@@ -0,0 +1,70 @@
...
+    errors = []
+    for N in [10, 20, 40]:
+        errors.append(error(N, problem, degree))
+
+    convergence_orders = lambda x: numpy.log2(numpy.array(x)[:-1] / numpy.array(x)[1:])
+    conv = convergence_orders(errors)
+    print("errors: ", errors)
+    print("convergence order: ", conv)
+
+    assert numpy.allclose(conv, degree, atol=0.1)
```
tommbendall, line 70: I've been looking into this -- I think I was expecting the convergence rate to actually be `degree+1`, so I think this is indicating that there are problems with all orders of `BDMCE/F` space?

wence-: Do you still think this is the case?

tommbendall: Sorry, I missed this message. I think the test and the elements were working when I last pushed to the branch

JDBetteridge (approved):
The empty review

Reviewed in meeting

PR #1490, opened 2019-08-23
purge lazy

wence-:
Needs to land simultaneously with OP2/PyOP2#568


firedrake/matrix.py
```diff
@@ -121,21 +114,9 @@ def assemble(self):
...
-            If you just need a handle on the :class:`pyop2.Mat` object it's
-            wrapping, use :attr:`_M` instead."""
-        # User wants to see it, so force the evaluation.
-        self._M._force_evaluation()
-        return self._M
-
     def force_evaluation(self):
         "Ensures that the matrix is fully assembled."
-        self._M._force_evaluation()
+        pass
```
miklos1: Probably retained for compatibility, although most calls to this elsewhere are being removed.

tests/benchmarks/test_solver_overheads.py
```diff
@@ -1,149 +0,0 @@
...
-
-    u = Function(V)
-    v = TestFunction(V)
-
-    F = u*v*dx - v*dx
-    if bcs:
-        bcs = DirichletBC(V, 0, 1)
-    else:
-        bcs = None
-    benchmark(lambda: solve(F == 0, u, bcs=bcs))
```
miklos1: Two benchmark files are being removed, not clear why.

wence-: The pytestplugin we use for this has bitrotted. Really I should have done it in a separate pr I guess

firedrake/matrix.py
wence-: Removed these calls.

wence-:
One of the two jenkins things passed, despite them testing identical commits, and I can't replicate the hang on my desktop. Any suggestions?

wence-:
I rebased again and fingers crossed

wence-:
These tests continue to hang. My hypothesis is still mumps issues, so I will expunge (temporarily) mumps from all the tests and cross my fingers.

wence-:
So that finally killed the issues, and this is now passing.

PR #1491, opened 2019-08-26
Ask scalar type from PETSc instead of just assuming "double"

miklos1:
This patch is cut from the `complex` branch, however, this part should be safe to merge.

Removes all references to `SCALAR_TYPE` from TSFC, instead we ask PETSc for the scalar type, and tell that to TSFC.

### DIFF
--- a/firedrake/evaluate.h
+++ b/firedrake/evaluate.h
@@ -18,11 +18,11 @@ struct Function {
 	int n_layers;
 
 	/* Coordinate values and node mapping */
-	double *coords;
+	PetscScalar *coords;
 	PetscInt *coords_map;
 
 	/* Field values and node mapping */
-	double *f;
+	PetscScalar *f;
 	PetscInt *f_map;
 
 	/* Spatial index */
@@ -37,7 +37,7 @@ struct Function {
 typedef int (*inside_predicate)(void *data_,
 				struct Function *f,
 				int cell,
-				double *x);
+				PetscScalar *x);
 
 typedef int (*inside_predicate_xtr)(void *data_,
 				struct Function *f,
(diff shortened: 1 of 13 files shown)


firedrake/evaluate.h
```diff
@@ -18,11 +18,11 @@ struct Function {
 	int n_layers;
 
 	/* Coordinate values and node mapping */
-	double *coords;
+	PetscScalar *coords;
```
wence-: Note that the _implementation_ that uses this stuff will not be complex-safe (since we need to say `PetscRealPart` inside everywhere).

firedrake/tsfc_interface.py
```diff
@@ -181,6 +182,8 @@ def compile_form(form, name, parameters=None, inverse=False, split=True, interfa
         parameters = default_parameters["form_compiler"].copy()
         parameters.update(_)
 
+    parameters['scalar_type'] = utils.ScalarType_c
```
wence-: Would it have made sense to set this in the `default_parameters["form_compiler"]` dict, rather than having to remember where-ever we use the parameters.

miklos1: I am making an attempt to do this.

firedrake/evaluate.h
miklos1: I am not that familiar with complex, would replacing `PetscScalar` with `PetscRealPart` just work in both real and complex cases?  I think I will leave this as it is now...

wence-: One needs to do something:

`PetscScalar *coords` and then replace any use of `coords[...]` with `PetscRealPart(coords[...])`, and possibly replace temporary variables that are declared as either `double` or `PetscScalar` with `PetscReal` I think.

PR #1497, opened 2019-09-05
ExtrudedMeshHierarchyUniform for full coarsening (not just semicoarsening)

pefarrell:
`ExtrudedMeshHierarchy` only does semi-coarsening (it keeps the number of layers in the vertical fixed).

When one wants to solve problems on hex meshes on O(1) aspect ratio domains, this isn't ideal. This pull request adds an `ExtrudedMeshHierarchyUniform` that does full coarsening in all dimensions. It also extends the code for identifying the cellwise inclusions appropriately (this was developed by @wence- and @JaroslavHron, not me).

I also add a test to make sure it does something sensible.

### DIFF
--- a/firedrake/mg/impl.pyx
+++ b/firedrake/mg/impl.pyx
@@ -122,7 +122,8 @@ def fine_to_coarse_nodes(Vf, Vc, np.ndarray[PetscInt, ndim=2, mode="c"] fine_to_
         coarse_offset = Vc.offset
         fine_offset = Vf.offset
         layers = Vc.mesh().layers - 1
-
+        fine_layers = Vf.mesh().layers - 1
+        
     fine_cells = fine_to_coarse_cells.shape[0]
     coarse_per_fine = fine_to_coarse_cells.shape[1]
     coarse_per_cell = coarse_map.shape[1]
@@ -137,14 +138,15 @@ def fine_to_coarse_nodes(Vf, Vc, np.ndarray[PetscInt, ndim=2, mode="c"] fine_to_
             for j in range(fine_per_cell):
                 node = fine_map[i, j]
                 if extruded:
-                    for layer in range(layers):
+                    for fine_layer in range(fine_layers):
+                        coarse_layer = fine_layer // 2
                         for k in range(coarse_per_cell):
-                            fine_to_coarse_map[node + fine_offset[j]*layer, k] = (coarse_map[coarse_cell, k] +
-                                                                                  coarse_offset[k]*layer)
+                            fine_to_coarse_map[node + fine_offset[j]*fine_layer, k] = coarse_map[coarse_cell, k] + coarse_offset[k]*coarse_layer
                 else:
                     for k in range(coarse_per_cell):
                         fine_to_coarse_map[node, coarse_per_cell*l + k] = coarse_map[coarse_cell, k]
 
+        
     return fine_to_coarse_map
 
 
--- a/firedrake/mg/impl.pyx
+++ b/firedrake/mg/impl.pyx
@@ -121,9 +121,13 @@ def fine_to_coarse_nodes(Vf, Vc, np.ndarray[PetscInt, ndim=2, mode="c"] fine_to_
     if extruded:
         coarse_offset = Vc.offset
         fine_offset = Vf.offset
-        layers = Vc.mesh().layers - 1
+        coarse_layers = Vc.mesh().layers - 1
         fine_layers = Vf.mesh().layers - 1
-        
+
+        ratio = fine_layers / coarse_layers
+        assert ratio == int(ratio)
+        ratio = int(ratio)
+
     fine_cells = fine_to_coarse_cells.shape[0]
     coarse_per_fine = fine_to_coarse_cells.shape[1]
     coarse_per_cell = coarse_map.shape[1]
@@ -139,14 +143,13 @@ def fine_to_coarse_nodes(Vf, Vc, np.ndarray[PetscInt, ndim=2, mode="c"] fine_to_
                 node = fine_map[i, j]
                 if extruded:
                     for fine_layer in range(fine_layers):
-                        coarse_layer = fine_layer // 2
+                        coarse_layer = fine_layer // ratio
                         for k in range(coarse_per_cell):
                             fine_to_coarse_map[node + fine_offset[j]*fine_layer, k] = coarse_map[coarse_cell, k] + coarse_offset[k]*coarse_layer
                 else:
                     for k in range(coarse_per_cell):
                         fine_to_coarse_map[node, coarse_per_cell*l + k] = coarse_map[coarse_cell, k]
 
-        
     return fine_to_coarse_map
 
 
--- a/firedrake/mg/impl.pyx
+++ b/firedrake/mg/impl.pyx
@@ -124,9 +124,8 @@ def fine_to_coarse_nodes(Vf, Vc, np.ndarray[PetscInt, ndim=2, mode="c"] fine_to_
         coarse_layers = Vc.mesh().layers - 1
         fine_layers = Vf.mesh().layers - 1
 
-        ratio = fine_layers / coarse_layers
-        assert ratio == int(ratio)
-        ratio = int(ratio)
+        ratio = fine_layers // coarse_layers
+        assert ratio * coarse_layers == fine_layers # check ratio is an int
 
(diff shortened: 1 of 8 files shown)


firedrake/mg/mesh.py
```diff
@@ -201,6 +201,36 @@ def ExtrudedMeshHierarchy(base_hierarchy, layers, kernel=None, layer_height=None
                          refinements_per_level=base_hierarchy.refinements_per_level,
                          nested=base_hierarchy.nested)
 
+def ExtrudedMeshHierarchyUniform(base_hierarchy, base_layers, height, kernel=None,
```
pefarrell: Note the change in API: I want to specify the final height of the domain, not the layer height (as this will vary by multigrid level).

pefarrell:
I can see your perspective.

In my dream world:

- make `ExtrudedMeshHierarchy` take in a vertical refinement factor, defaulting to 2 (i.e. change the current default behaviour, as semicoarsening is less common than full coarsening)
- make `ExtrudedMeshHierarchy` take in the `height`, rather than `layer_height`, as `layer_height` now varies with refinement
- update the tests etc and notify the gusto/thetis/icepack developers to change their code too

What do you think?

dham:
I agree. I think the only question that comes out of that is whether we should also change the API for straight extrusion to match.

pefarrell:
Great. Perhaps some subset of @colinjcotter , @jshipton , @tkarna or @danshapero can comment.

firedrake/mg/mesh.py
```diff
@@ -201,6 +201,36 @@ def ExtrudedMeshHierarchy(base_hierarchy, layers, kernel=None, layer_height=None
...
+                          extrusion_type='uniform', gdim=None,
+                          mesh_builder=firedrake.ExtrudedMesh):
+    """Build a hierarchy of extruded meshes by extruding a hierarchy of meshes.
+
+    :arg base_hierarchy: the unextruded base mesh hierarchy to extrude.
+    :arg mesh_builder: function used to turn a :class:`~.Mesh` into an
+       extruded mesh. Used by pyadjoint.
+
+    See :func:`~.ExtrudedMesh` for the meaning of the remaining parameters.
```
wence-: This docstring doesn't really tell me the difference between this version and just an `ExtrudedMeshHierarchy`. How about, if we unify (as suggested by @dham) the interface:

```python
def ExtrudedMeshHierarchy(base_hierarchy, height, *layers, *, extrusion_type='uniform', gdim=None, mesh_builder=firedrake.ExtrudedMesh):
   """Build a hierarchy of extruded meshes by extruding a hierarchy of meshes.

   :arg base_hierarchy: The unextruded base mesh hierarchy to extrude.
   :arg height: Height of the extruded mesh.
   :arg layers: number of layers for each extruded mesh. Either pass a single integer (for same number of layers on every level), or else as many as there are meshes in the base hierarchy.

    .. note::

       If different numbers of layers are selected on each level, each coarse mesh must have the same number (or fewer levels) as its fine mesh counterpart, and the number of layers on successive levels must be integer-divisible.
"""
    try:
        layers, = layers
        layers = itertools.repeat(layers)
    except ValueError:
        if len(layers) != len(base_hierarchy):
            raise ValueError("Must either provide single layer for all meshes or layers for each mesh")
        if any(c >= f for c, f in zip(layers, layers[1:])):
            raise ValueError("Coarser meshes cannot have more levels")
        if any(f / c != f // c for c, f in zip(layers, layers[1:])):
            raise ValueError("Refinement ratio in extruded direction must be an integer")
    meshes = [mesh_builder(mesh, layer, ...) for mesh, layer in zip(base_hierarchy._meshes, layers)]
```

wence-: Although I note this does a really weird thing if `refinements_per_level` of the base hierarchy is not 1. I need to think about that.

firedrake/mg/mesh.py
```diff
@@ -201,6 +201,36 @@ def ExtrudedMeshHierarchy(base_hierarchy, layers, kernel=None, layer_height=None
...
+    if not isinstance(base_hierarchy, HierarchyBase):
+        raise ValueError("Expecting a HierarchyBase, not a %r" % type(base_hierarchy))
+    if any(m.cell_set._extruded for m in base_hierarchy):
+        raise ValueError("Meshes in base hierarchy must not be extruded")
+
+    meshes = [mesh_builder(m, base_layers * 2**idx, kernel=kernel,
+                           layer_height=height/(base_layers * 2**idx),
+                           extrusion_type=extrusion_type,
+                           gdim=gdim)
+              for idx, m in enumerate(base_hierarchy._meshes)]
```
wence-: If refinements_per_level of the base hierarchy is not 1, then this does uniform refinement at the right rate, I think.

firedrake/mg/impl.pyx
```diff
@@ -121,7 +121,12 @@ def fine_to_coarse_nodes(Vf, Vc, np.ndarray[PetscInt, ndim=2, mode="c"] fine_to_
     if extruded:
         coarse_offset = Vc.offset
         fine_offset = Vf.offset
-        layers = Vc.mesh().layers - 1
+        coarse_layers = Vc.mesh().layers - 1
+        fine_layers = Vf.mesh().layers - 1
+
+        ratio = fine_layers / coarse_layers
+        assert ratio == int(ratio)
+        ratio = int(ratio)
```
wence-: Use
```python
ratio = fine_layers // coarse_layers
assert ratio * coarse_layers == fine_layers
```
?

colinjcotter:
I never liked that ExtrudedMesh took layer_height and not domain height - it becomes even more important with hierarchies to have domain height as the parameter (saves us having to do maths).

firedrake/mg/mesh.py
pefarrell: I don't like that one would have to do something like

`
base = UnitSquareMesh(10, 10, quadrilateral=True)
basemh = MeshHierarchy(base, 2)
mh = ExtrudedMeshHierarchy(basemh, height=1, layers=[10 * 2**i for i in range(2)])
`

It seems ugly to have to specify the number of refinements twice (once in the base hierarchy, and once in the list of layers). This is doubly so when the ratio has to be an integer anyway. I think we should just specify the ratio and compute the layers for each refinement in the base hierarchy.

wence-: Well, the ratio need not be 2 (I think). If we are going to allow coarsening in the extruded direction, I think it makes sense to support not just uniform refinement (or no refinement). e.g. if you want to do Poisson on thin domains, it makes sense to go from N layers to 1 layer in one step, and then solve 2D problems.

Why not offer:

```
EMH(..., base_layer=N, refinement_ratio=P, layers=[list_of_layers])
```

And check that user either provides base_layer and refinement_ratio (the latter defaulting to 1 as presently?) or list_of_layers?

pefarrell:
OK, I've made a first pass at unifying `ExtrudedMeshHierarchy` and `ExtrudedMeshHierarchyUniform`.

pefarrell:
If we do agree to change the API for `ExtrudedMesh` (to specify height rather than layer_height), I'd rather it happen in another pull request.

firedrake/mg/mesh.py
pefarrell: The code now implements the API @wence- suggests.

firedrake/mg/impl.pyx
pefarrell: Done

firedrake/mg/mesh.py
```diff
@@ -173,10 +173,10 @@ def MeshHierarchy(mesh, refinement_levels,
                          refinements_per_level, nested=True)
 
 
-def ExtrudedMeshHierarchy(base_hierarchy, layers, kernel=None, layer_height=None,
-                          extrusion_type='uniform', gdim=None,
+def ExtrudedMeshHierarchy(base_hierarchy, height, base_layer=-1, refinement_ratio=2, layers=None,
+                          kernel=None, extrusion_type='uniform', gdim=None,
```
dham: Please add documentation for the new arguments to the docstring. I realise that not all the existing arguments are documented, but that's not really an excuse for making things worse.

pefarrell: Agreed, thanks for pointing it out. I have added docstrings.

wence-:
Can you fix the lint errors, then I will merge.

pefarrell:
Sorry, done.

PR #1498, opened 2019-09-06
add rule for Max

colinjcotter:
Add rule so that Loopy can handle ufl.Max.

### DIFF
--- a/firedrake/assemble_expressions.py
+++ b/firedrake/assemble_expressions.py
@@ -607,6 +607,18 @@ def loopy_inst_abs(expr, context):
     return p.Variable("abs")(child)
 
 
+@loopy_instructions.register(ufl.classes.MaxValue)
+def loopy_inst_max(expr, context):
+    children = [loopy_instructions(o, context) for o in expr.ufl_operands]
+    return p.Variable("max")(*children)
+
+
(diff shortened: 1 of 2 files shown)


firedrake/assemble_expressions.py
```diff
@@ -607,6 +607,12 @@ def loopy_inst_abs(expr, context):
     return p.Variable("abs")(child)
 
 
+@loopy_instructions.register(ufl.classes.MaxValue)
+def loopy_inst_max(expr, context):
+    children = [loopy_instructions(o, context) for o in expr.ufl_operands]
+    return p.Variable("max")(*children)
```
dham: Presumably you want to add `min` as well.

firedrake/assemble_expressions.py
```diff
@@ -607,6 +607,18 @@ def loopy_inst_abs(expr, context):
...
+
+
+@loopy_instructions.register(ufl.classes.MinValue)
+def loopy_inst_min(expr, context):
+    children = [loopy_instructions(o, context) for o in expr.ufl_operands]
+    return p.Variable("min")(*children)
```
wence-: Tests please!

PR #1499, opened 2019-09-16
Add the ability to write higher order elements to VTK

wraith1995:
In accordance with [this announcement](https://blog.kitware.com/modeling-arbitrary-order-lagrange-finite-elements-in-the-visualization-toolkit/), this pull makes it possible to write higher order elements to vtu files in a manner that VTK/Paraview can understand. 

To do this, a new file, paraview_reordering.py, handles the annoyingly trivial task of permuting basis elements. This could be made more dependable and easier to maintain if we added a vtk dependency and then replaced various marked functions in this file with calls to vtk's python bindings.

output.py can still write linear elements if they are given, but if a non-linear element is detected, it will (conservatively) find an element that contains ALL involved elements (from the geometry and functions), project/interpolate on to these elements, and then write the data out. This is based on an interpretation of the unstructured grid format that might be wrong; as far as I can tell, the point data arrays (i.e where function dofs are written) must be in correspondence to points (the dofs for the coordinates), making everything isoparametric. This appears to be consistent with lower order, but non-linear elements in VTK.

I've not added any new tests; I have tested by comparing higher order output to both refined linear output and non-refined linear output:
```
from firedrake import *
m1 = UnitSquareMesh(2, 2)
m2 = UnitSquareMesh(10, 10)
V1 = FunctionSpace(m1, "Lagrange", 4)
V2 = FunctionSpace(m1, "Lagrange", 1)
V3 = FunctionSpace(m2, "Lagrange", 1)
x,y = SpatialCoordinate(m1)
xp, yp = SpatialCoordinate(m2)
exp = x*x*x*x + y*y*y*y
expp = xp*xp*xp*xp + yp*yp*yp*yp
#interpolation of these does not work it seems so fuck it.
f1 = interpolate(exp, V1)
f2 = interpolate(exp, V2)
f3 = interpolate(expp, V3)
File("TTest1.pvd").write(f1)
File("TTest2.pvd").write(f2)
File("TTest3.pvd").write(f3)
```
These build confidence though obviously this is far from complete assurance. Other people trying this out would be welcome as I might be missing some obvious holes.


pefarrell:
I suggest (i) renaming `oder` -> `order` in a bunch of places; (ii) renaming `paraview_reordering_bs.py` to `paraview_reordering.py`.

wraith1995:
@pefarrell Done.

Oh, also, note to everyone who uses this: you might have to use the 'tesselate' filter if data looks strange.

pefarrell:
This looks really cool.

I made a 4th order mesh of `tests/meshes/disk.step` with a very coarse mesh size and plotted it in paraview with this branch. The result clearly shows that paraview renders the boundary of the domain as 4 piecewise linear segments, instead of 1. After tesselation (the semi-transparent border around the boundary), the rendering of the boundary improves further.

disk

florianwechsung:
@pefarrell I think you need to increase the "Nonlinear Subdivision Level" in paraview.
In "Properties" start typing "nonlinear" and then you get the option. It should eventually look smooth like this: 
image

pefarrell:
Whoa, cool!

disk-better

wraith1995:
Hmmm... Maybe we should update the [plotting docs](https://www.firedrakeproject.org/visualisation.html) in this pull or another. Maybe @florianwechsung  and @pefarrell  can turn some beta-testing experience into this? I've not used this feature much and I think they are about to get much more practice with it.

That all being said, I think that I can provide a bit of context: in both of those options, I think that Paraview is just building a linear approximation to the original data. This is important to keep in mind because you might want to apply other Paraview algos to your visualization. Based on the code, I think isocontouring and clipping probably use the original higher order data (I'm not entirely sure). Besides this, I think that any other algo will use a linear approximation. Consequently, it might be wise apply algos to the tessellated data as opposed to the original data.(Note: you can save the tessellated data out if you want and it really should be thought of as an approximation that Paraview keeps around and uses when it does not know how to handle the higher order data.)

wraith1995:
I should also note that I suspect some of the Paraview code is [partially based on this paper](https://ieeexplore.ieee.org/document/1634311/citations?tabFilter=papers#citations), which is based on [this report](https://pdfs.semanticscholar.org/bed4/8789fa4edf82ae7f29e8c9c577516ae735b3.pdf). I believe this is the case because (1) of the names on the papers and changes and (2) because these papers originally worked around VTK.

pefarrell:
Teo, the paper that you cited refers to a "mesh with linear geometry and cubic attribute basis functions" (caption of Figure 2). Does that suggest that it's possible _not_ to do isoparametric in VTK somehow?

firedrake/output.py
```diff
@@ -66,6 +85,32 @@ def is_linear(V):
     nvertex = V.ufl_domain().ufl_cell().num_vertices()
     return V.finat_element.space_dimension() == nvertex
 
+#  When I wrote this function, I desperately wanted to write haskell code.
```
dham: A very droll comment, but probably not appropriate in production code!

firedrake/output.py
```diff
@@ -66,6 +85,32 @@ def is_linear(V):
...
+    :returns: A ufl element containing elem1 and elem2.
+    """
+    cell1 = elem1.cell()
+    cell2 = elem2.cell()
+    if cell1 != cell2:
+        raise Exception("Sup for spaces does not exist!")
+    maxDegree = 0
+    degree1 = elem1.degree()
+    degree2 = elem2.degree()
+    if isinstance(degree1, int):
```
dham: You can cut this to one line as `maxDegree = max(maxDegree, max(degree1))`

firedrake/output.py
```diff
@@ -75,19 +120,24 @@ def get_topology(coordinates):
         :class:`OFunction`\s.
     """
     V = coordinates.function_space()
+
+    use_lag = not(is_linear(V))
```
dham: `use_lag` is a bit obscure. How about `nonlinear`.

firedrake/output.py
```diff
@@ -109,52 +159,68 @@ def get_topology(coordinates):
...
-    else:
-        scale = cell.num_vertices()
-
     # Repeat up the column
     num_cells = mesh.cell_set.size
     if not mesh.cell_set._extruded:
         cell_layers = 1
         offsets = 0
     else:
+        # NOTE: This could be made more efficiet in the linear case.
```
dham: Typo: efficient.

firedrake/paraview_reordering.py
```diff
@@ -0,0 +1,449 @@
...
+
+
+def invert(list1, list2):
+    r"""Given two maps (lists) from [0..N] to nodes, finds a permutations between them.
+    :arg list1: a list of nodes.
+    :arg list2: a second list of nodes.
+    :returns a list of integers, l, such that list1[x] = list2[l[x]]
+    """
+    if len(list1) != len(list2):
+        raise Exception("Dimension of Paraview basis and Element basis unequal.")
```
dham: This should be `ValueError`. `Exception` should never be raised (it's just the base class).

firedrake/paraview_reordering.py
```diff
@@ -0,0 +1,449 @@
...
+
+def vtk_hex_point_index_from_ijk(i, j, k, order=None):
+    r"""
+    See vtkLagrangeHexahedron::PointIndexFromIJK
+    """
+    ibdy = (i == 0 or i == order[0])
+    jbdy = (j == 0 or j == order[1])
+    kbdy = (k == 0 or k == order[2])
+
+    itrue = i != 0
```
dham: This is excessive factorisation. Just use `i!=0` in place each time.

firedrake/paraview_reordering.py
```diff
@@ -0,0 +1,449 @@
...
+        ret = 4 if ktrue else 0
+        if itrue:
+            ret += 2 if jtrue else 1
+        else:
+            ret += 3 if jtrue else 0
+        return ret
+    offset = 8
+    if nbdy == 2:  # edge
+        if not ibdy:  # on the i axis
+            temp0 = i - 1
```
dham: I don't think any of these temporaries are doing anything for us. Just add onto the offset.

firedrake/paraview_reordering.py
```diff
@@ -0,0 +1,449 @@
...
+def vtk_quad_index_from_ij(i, j, order):
+    r"""
+    See vtkLagrangeQuadrilateral::PointIndexFromIJK
+    """
+    ibdy = (i == 0 or i == order[0])
+    jbdy = (j == 0 or j == order[1])
+
+    nbdy = int(ibdy) + int(jbdy)
+
+    itrue = i != 0
```
dham: as above.

firedrake/paraview_reordering.py
```diff
@@ -0,0 +1,449 @@
...
+    jtrue = j != 0
+    if (nbdy == 2):
+        return qsynatx(itrue, qsynatx(jtrue, 2, 1), qsynatx(jtrue, 3, 0))
+
+    offset = 4
+    if nbdy == 1:
+        if not ibdy:
+            temp1 = i - 1
```
dham: as above.

firedrake/paraview_reordering.py
```diff
@@ -0,0 +1,449 @@
...
+    def find_same(val, lst, tol=0.0000001):
+        for (idx, x)in enumerate(lst):
+            if np.linalg.norm(val - x) < tol:
+                return idx
+        raise Exception("Basis of element incompatible with paraview basis: node {0} is not in {1}".format(val, lst))
+    # finds map from idx of list 1 to list of list 2.
+    perm = [find_same(x, list2) for x in list1]
+    #  check for bijectivity:
+    if len(set(perm)) != len(list1):
+        raise Exception("Unable to establish permutation between Paraview basis and given element basis.")
```
wence-:
```python
N = len(list1)
pt1, o1 = zip(*sorted(zip(map(tuple, list1), range(N))))
pt2, o2 = zip(*sorted(zip(map(tuple, list2), range(N))))
if not np.allclose(pt1, pt2):
    raise ValueError(...)

return numpy.asarray(o2)[numpy.argsort(o2)
```

NlogN, rather than N^3. And you get to feel like you were writing Haskell.

firedrake/paraview_reordering.py
```diff
@@ -0,0 +1,449 @@
...
+    return perm
+
+
+"""
+The following functions are translations of funtions in the VTK source;we use them find the order of nodes in the lagrange basises that Paraview uses. We don't document them fully, but link back to the vtk source.
+These come from VTK version 8.2.0.
+We could simplify all of this by depending on VTK's python bindings.
```
wence-: `pip install vtk` appears to work, the only question for me is whether it plays badly with any global vtk install

firedrake/output.py
```diff
@@ -418,9 +473,18 @@ def _write_vtu(self, *functions):
...
             is_cg(mesh.coordinates.function_space())
 
-        coordinates = self._prepare_output(mesh.coordinates, continuous)
+        # Since Points define nodes for both the mesh and function, we must
+        # interpolate/project ALL involved elements onto a single larger
+        # finite element.
+        maxElem = mesh.coordinates.function_space().ufl_element()
+        for f in functions:
+            newElem = f.function_space().ufl_element()
+            maxElem = get_sup_element(maxElem, newElem,
```
wence-: Can we write `get_sup_element` to just eat `*elements` and then write:

```python
element = get_sup_element(*(f.ufl_element() for f in functions), ...)
```

firedrake/output.py
```diff
@@ -66,6 +85,32 @@ def is_linear(V):
...
+    degree2 = elem2.degree()
+    if isinstance(degree1, int):
+        maxDegree = max(degree1, maxDegree)
+    else:
+        maxDegree = max(maxDegree, *degree1)
+    if isinstance(degree2, int):
+        maxDegree = max(degree2, maxDegree)
+    else:
+        maxDegree = max(maxDegree, *degree2)
+    return(ufl.FiniteElement("CG" if both_continous else "DG", cell1, maxDegree))
```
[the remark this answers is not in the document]
wence-: From my comment below about accepting a list of elements:
```python
import itertools
def get_sup_element(*elements, *, continuous=False):
    try:
        cell, = set(e.cell() for e in elements)
    except ValueError:
        raise ValueError("All cells must be identical")
    degree = max(itertools.chain(*(as_tuple(e.degree()) for e in elements))
    return ufl.FiniteElement("CG" if continuous else "DG", cell, degree)
```

firedrake/paraview_reordering.py
```diff
@@ -0,0 +1,449 @@
...
+    # finds map from idx of list 1 to list of list 2.
+    perm = [find_same(x, list2) for x in list1]
+    #  check for bijectivity:
+    if len(set(perm)) != len(list1):
+        raise Exception("Unable to establish permutation between Paraview basis and given element basis.")
+    return perm
+
+
+"""
+The following functions are translations of funtions in the VTK source;we use them find the order of nodes in the lagrange basises that Paraview uses. We don't document them fully, but link back to the vtk source.
```
pefarrell: basises -> bases

wraith1995:
@pefarrel - As I said before, it is possible that VTK supports this and I've missed it. However, I find it more likely either the paper was not completely implemented as stated or they used a workaround analogous to what I'm doing. Anyway, the isoparametric constraint is annoying so I'm looking for a more authoritative answer (s[tarting on the VTK discourse](https://discourse.vtk.org/t/vtk-unstructured-grid-and-higher-order-lagrange-elements/1777)) and I'll report back if I learn anything.

I should also note this is fairly consistent with my experience with VTK. Here is another similar example: for hex elements, VTK can work with any triplet of orders, but the function  vtkLagrangeHexahedron::GetOrder enforces that all orders are the same because VTK currently must infer the order from the number of points needed to define the element! Thus, my implementation of this functionality is forced to turn a Q\times I with orders (4,2) into (4,4). This is just a limitation of the rather bad practice of dumping raw arrays of data without much information about the associated basis.

To others, revisions/comments incoming over the next 24ish hours.

firedrake/output.py
```diff
@@ -75,19 +120,24 @@ def get_topology(coordinates):
...
     cell = mesh.ufl_cell()
     values = V.cell_node_map().values
-    # Non-simplex cells need reordering
+    value_shape = values.shape
+    basis_dim = value_shape[1]
+    offsetMap = V.cell_node_map().offset
+    # Non-simplex cells and non-linear cells need reordering
     # Connectivity of bottom cell in extruded mesh
-    if cells[cell] == VTK_QUADRILATERAL:
+    if cells[cell, use_lag] == VTK_QUADRILATERAL:
```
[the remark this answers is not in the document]
wraith1995: They do as far as I can tell, but I think we should keep the linear options around for three reasons: 1. Older versions of Paraview (before 5.6 I think), won't understand the new cell types and you might not want to force people to update. 2. The linear elements are a useful tool for debugging this code or visualizations in general: they can be used as a potential source of truth.

firedrake/paraview_reordering.py
```diff
@@ -0,0 +1,449 @@
+from tsfc.fiatinterface import create_element
+import numpy as np
+
+
+def firedrake_local_to_cart(element):
+    r"""Gets the list of nodes for an element (provided they exist.)
+    :arg element: a ufl element.
+    :returns: a list of arrays of floats where each array is a node.
+    """
+    fiat_element = create_element(element, vector_is_mixed=False)
```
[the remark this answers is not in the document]
wraith1995: It is equispaced. Fixed via changing updating get_sup_element to enforce variant="equispaced"

firedrake/paraview_reordering.py
wraith1995: I don't think it will. 1. Anecdotally, I have no issues with a global install and python bindings. 2. The python bindings come with their own binary blobs (dynamic libs) that are, unless you got out of your way, only used by the python bindings (just look at https://pypi.org/project/vtk/#files).

firedrake/output.py
```diff
@@ -7,7 +6,10 @@
 import weakref
 from pyop2.mpi import COMM_WORLD, dup_comm
 from pyop2.datatypes import IntType
-
+from .paraview_reordering import vtk_lagrange_tet_reorder,\
+    vtk_lagrange_hex_reorder, vtk_lagrange_interval_reorder,\
+    vtk_lagrange_triangle_reorder, vtk_lagrange_quad_reorder,\
+    vtk_lagrange_wedge_reorder
```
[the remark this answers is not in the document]
wraith1995: Ubuntu is easy - as noted by you below, pip install vtk works. When I had a mac about 2 years ago, I never had any issues with this.

wraith1995:
Alright, everything has been resolved except:

- Do we keep the linear elements around.

- Do we use `pip install vtk` (This does not install vtk and I don't think it should conflict with an existing vtk install.)

- Update the docs. I can do this, but not right now (maybe by Thursday).

docs/source/visualisation.rst
```diff
@@ -7,7 +7,7 @@ Visualising the results of simulations
 
 Having run a simulation, it is likely that we will want to look at the
 results.  To do this, Firedrake supports saving data in VTK_
-format, suitable for visualisation in Paraview_ (amongst others).
```
dham: @colinjcotter is going to be very unhappy with you removing his "amongst".

colinjcotter: did I write that? I don't remember doing it

dham: "amongst" is your style...

colinjcotter:
I like to put the cat amongst the pigeons

firedrake/output.py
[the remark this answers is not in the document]
dham: We really don't want to add a compiled ubuntu/homebrew dependency because this will muck up supercomputer installs. However if we can just pip install, this is fine.

docs/source/visualisation.rst
wraith1995: I blame ispell. However, I kept visualisation for some reason so we might as well be consistent.

firedrake/output.py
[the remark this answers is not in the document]
wraith1995: Alright. I will add the pip install, revise the pull and make jenkins happy with it.

wraith1995:
@dham So, I was able to get things working with `pip install vtk`, but possibly in a potentially unsatisfactory way. If this is the case, let's just roll back to right after spelling corrections and be done there.

The main issue/attempted solution is noted in the comments of the code, but I'll repeat it here: `pip install vtk` downloads dynamic libs meant for exclusive use via python bindings, but some of these libs depend on things that you wouldn't have on a headless machine. However, the parts that we need depend on fairly typical things. Via manipulating importlib, I can extract only the parts that we need and everything works (Jenkins and my machine are okay at least).

pefarrell:
@wraith1995, I have a question. Consider the following code:

```
from firedrake import *

mesh.coordinates.dat.data[:] -= 0.5
V = FunctionSpace(mesh, "CG", 2)

(x, y) = SpatialCoordinate(mesh)
u = interpolate(x**2 + y**2, V)

File("quadratic.pvd").write(u)
```

When I load this in paraview, if I increase the number of nonlinear subdivisions the picture looks right. However, I can't get the isocontours to work. For example, if I increase the number of nonlinear subdivisions, then apply the isocontour filter, and increase the number of nonlinear subdivisions, I get

contours

They should be circles, but are not.

If I try tesselate, it only adds a few triangles, and I can't figure out how to get it to refine further.

tessellate

(The contours are wrong on this tesselation, of course).

Do you know how to get paraview to draw the correct contours?

colinjcotter:
This doesn't answer your question but I always cheat and reduce the number of colours to show contours.

wraith1995:
@pefarrell This is a bit stupid and it occurs to me I should update the docs to reflect this: basically, the options that are obviously shown by Paraview in the Tessellate menu don't cause further refinement for this data set. To trigger more, you must use a non-obvious option, which I've shown how to set here (look at the side panel and right below chord error):
 
Screenshot from 2019-09-25 17-18-01

The number being set is the field error, which is analogous to the coord error, but measures field values at the positions used to measure chord error. The availability of this option is[ documented here](https://kitware.github.io/paraview-docs/latest/python/paraview.simple.Tessellate.html), but I should add it to the docs that we provide.

firedrake/paraview_reordering.py
```diff
@@ -0,0 +1,236 @@
...
+    # TODO: Surely there is an easier way that I've forgotten?
+    carts = [np.array(list(phi.get_point_dict().keys())[0])
+             for phi in fiat_element.dual_basis()]
+    return carts
```
wence-: style: just return the list above directly.

firedrake/paraview_reordering.py
```diff
@@ -0,0 +1,236 @@
...
+
+
+def invert(list1, list2):
+    r"""Given two maps (lists) from [0..N] to nodes, finds a permutations between them.
+    :arg list1: a list of nodes.
+    :arg list2: a second list of nodes.
+    :returns a list of integers, l, such that list1[x] = list2[l[x]]
```
wence-: Formatting, `:returns:` with a trailing colon.

firedrake/paraview_reordering.py
```diff
@@ -0,0 +1,236 @@
...
+The follow functions call the two types of VTK functions and use them to produce maps from dof indicies to node locations. These functions are then wrapped to be applied to ufl_elements after the next block commment.
+"""
+
+
+def vtk_interval_local_coord(i, order):
+    r"""
+    See vtkLagrangeCurve::PointIndexFromIJK.
+    """
+    if i == 0:
+        return (0.0)
```
wence-: Style: no parentheses.

firedrake/paraview_reordering.py
```diff
@@ -0,0 +1,236 @@
...
+    elif i == order:
+        return 1.0
+    else:
+        return i / order
+
+
+def bar_to_cart_3d(bar):
```
wence-: `s/bar/bary/` I suspect.

firedrake/paraview_reordering.py
```diff
@@ -0,0 +1,236 @@
...
+        loc_to_cart[idx] = cart
+    return(loc_to_cart)
+
+
+def bar_to_cart_2d(bar):
+    v0 = np.array([0, 0])
+    v1 = np.array([1, 0])
+    v2 = np.array([0, 1])
+    mat = np.array([v1, v2, v0])
+    return(np.dot(bar, mat))
```
wence-: You could generalise these functions by:
```
N = len(bar)
mat = numpy.vstack([numpy.zeros(N), numpy.eye(N)])
```

firedrake/paraview_reordering.py
```diff
@@ -0,0 +1,236 @@
...
+    """
+    sizes = tuple([o + 1 for o in orders])
+    size = np.product(sizes)
+    loc_to_cart = np.empty(size, dtype="object")
+    quad = vtkLagrangeQuadrilateral()
+    for loc in np.ndindex(sizes):
+        idx = quad.PointIndexFromIJK(loc[0], loc[1], orders)
+        cart = np.array([c / o for (c, o) in zip(loc, orders)])
+        loc_to_cart[idx] = cart
+    return(loc_to_cart)
```
wence-: Style: `return loc_to_cart`

firedrake/paraview_reordering.py
wence-: Also style: `return np.dot(bar, mat))`

firedrake/paraview_reordering.py
```diff
@@ -0,0 +1,236 @@
...
+    totalSize = triSize * (orders[2] + 1)
+    loc_to_cart = np.empty(totalSize, dtype="object")
+    wedge = vtkLagrangeWedge()
+    for loc in np.ndindex(sizes):
+        if loc[0] + loc[1] > orders[0]:
+            continue
+        idx = wedge.PointIndexFromIJK(loc[0], loc[1], loc[2], orders)
+        cart = np.array([c / o for (c, o) in zip(loc, orders)])
+        loc_to_cart[idx] = cart
+    return(loc_to_cart)
```
wence-: Style

firedrake/paraview_reordering.py
```diff
@@ -0,0 +1,236 @@
...
+uses.
+"""
+
+
+def vtk_lagrange_interval_reorder(ufl_element):
+    degree = ufl_element.degree()
+    vtk_local = [vtk_interval_local_coord(x, degree) for x in range(degree + 1)]
+    firedrake_local = firedrake_local_to_cart(ufl_element)
+    inv = invert(vtk_local, firedrake_local)
+    return inv
```
wence-: Style, just `return invert(vtk_local, firedrake_local)`

firedrake/paraview_reordering.py
```diff
@@ -0,0 +1,236 @@
...
+    firedrake_local = firedrake_local_to_cart(ufl_element)
+    return invert(vtk_local, firedrake_local)
+
+
+def vtk_lagrange_quad_reorder(ufl_element):
+    degree = as_tuple(ufl_element.degree())[0]  # should be uniform
+    vtk_local = vtk_quad_local_to_cart((degree, degree))
+    firedrake_local = firedrake_local_to_cart(ufl_element)
+    inv = invert(vtk_local, firedrake_local)
+    return (inv)
```
wence-: Style

docs/source/visualisation.rst
```diff
@@ -81,6 +82,73 @@ possible to override this by passing the keyword argument ``time``.
...
+
+If instead we want projection, we use
+
+.. code-block:: python
+
+   projected = File("proj_output.pvd", project_output=True)
+   projected.write(f)
+
+.. note::
+   This feature requires Paraview version 5.5.0 or better. If you must use an older version of Paraview, you must manually interpolate mesh coordinates and field coordinates to a piecewise linear function space, represented with either a Lagrange or discontinuous Lagrange basis.
```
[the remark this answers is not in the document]
wraith1995: I think we should add something like that - I think that `target_element` makes more sense? Or maybe `target_cg` and `target_degree` ?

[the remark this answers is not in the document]
wraith1995: I went with the latter, but don't feel strongly about changing it if you so desire.

wence-:
Merged (with some minor fixes), thanks!

PR #1510, opened 2019-10-04
Wence/fix/dmhooks breadcrumbs

wence-:


wence-:
This arose from a thetis test fail, @tkarna can you confirm all is good?

PR #1516, opened 2019-10-10
Adapt to new pymbolic parsing rules

wence-:
Ternary if statements now written in python syntax

  foo if True else bar

rather than

  if(True, foo, bar)

This removes the need for the workaround in PyOP2 (filtering warnings).


wence-:
These passed, but Jenkins didn't manage to notify GitHub.

One thing that is missing, I think, is that we need to upgrade `pymbolic` in firedrake-update.

wence-:
> One thing that is missing, I think, is that we need to upgrade pymbolic in firedrake-update.

Ah it's OK, because we get pip to resolve the loopy dependencies, so things will be upgraded that way.

PR #1518, opened 2019-10-17
Allow user to pass bindings for coefficients in assembly

wence-:
Allows separation of symbolic UFL form from the data it will be
assembled with. Useful for avoiding use of ufl.replace in adjoint
annotation and replay.

@jrmaddison requested this, or something similar, so that in tlm_adjoint, the storage (which might want to be thrown away) can be separated from the symbolic form. With this interface, we can do a one-time `ufl.replace` on the form when taping, and then have a fast assembly with coefficients by providing a map `{symbolic_coefficient: real_coefficient}`. James reports this makes a large (positive) improvement to the speed: previously he was doing this himself by having fake coefficients which were mutated to add data at the correct moment.

### DIFF
--- a/firedrake/assemble.py
+++ b/firedrake/assemble.py
@@ -82,15 +82,9 @@ def assemble(f, tensor=None, bcs=None, form_compiler_parameters=None,
     boundary condition values.
     """
 
-    if "nest" in kwargs:
-        nest = kwargs.pop("nest")
-        from firedrake.logging import warning, RED
-        warning(RED % "The 'nest' argument is deprecated, please set 'mat_type' instead")
-        if nest is not None:
-            mat_type = "nest" if nest else "aij"
-
     collect_loops = kwargs.pop("collect_loops", False)
     allocate_only = kwargs.pop("allocate_only", False)
+    bindings = kwargs.pop("bindings", None)
     if len(kwargs) > 0:
         raise TypeError("Unknown keyword arguments '%s'" % ', '.join(kwargs.keys()))
 
@@ -101,7 +95,8 @@ def assemble(f, tensor=None, bcs=None, form_compiler_parameters=None,
                           sub_mat_type=sub_mat_type, appctx=appctx,
                           assemble_now=not collect_loops,
                           allocate_only=allocate_only,
-                          options_prefix=options_prefix)
+                          options_prefix=options_prefix,
+                          bindings=bindings)
         loops = tuple(loops)
         if collect_loops and not allocate_only:
             # Will this be useful?
@@ -170,7 +165,8 @@ def _assemble(f, tensor=None, bcs=None, form_compiler_parameters=None,
               options_prefix=None,
               assemble_now=False,
               allocate_only=False,
-              zero_tensor=True):
+              zero_tensor=True,
+              bindings={}):
     r"""Assemble the form or Slate expression f and return a Firedrake object
     representing the result. This will be a :class:`float` for 0-forms/rank-0
     Slate tensors, a :class:`.Function` for 1-forms/rank-1 Slate tensors and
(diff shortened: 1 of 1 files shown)


jrmaddison:
This is a source of irritation for small problems, where the cost of ufl.replace can be significant (see https://github.com/FEniCS/ufl/pull/14).

This is a particular issue in an adjoint calculation for the case where a forward equation is used multiple times, and in the worst case you must call ufl.replace once per adjoint term (or else differentiate the residual once per term), repeatedly, each time the adjoint equation is solved.

ksagiyam:
I think it looks good.

jrmaddison:
The following construct:

```
from firedrake import *

import ufl

mesh = UnitIntervalMesh(10)
space = FunctionSpace(mesh, "Lagrange", 1)
test, trial = TestFunction(space), TrialFunction(space)

m_c = ufl.classes.Coefficient(space)

m_v = Function(space, name="m_v")
m_v.assign(1.0)

A = assemble(inner(test, m_c * trial) * dx, bindings={m_c: m_v})

b = assemble(inner(test, Constant(1.0)) * dx)

F = Function(space, name="F")
solver = LinearSolver(A)
solver.solve(F, b)
```

leads to an error

```
  File "test.py", line 14, in <module>
    A = assemble(inner(test, m_c * trial) * dx, bindings={m_c: m_v})
  File "[...]/firedrake/firedrake/src/firedrake/firedrake/assemble.py", line 100, in assemble
    loops = tuple(loops)
  File "[...]/firedrake/firedrake/src/firedrake/firedrake/assemble.py", line 239, in _assemble
    for coeff in f.coefficients()):
  File "[...]/firedrake/src/firedrake/firedrake/assemble.py", line 239, in <genexpr>
    for coeff in f.coefficients()):
AttributeError: 'Coefficient' object has no attribute 'function_space'
```

The original Form, stored on the Matrix.a attribute, refers to the incorrect Coefficient.

wence-:
@jrmaddison sorry, I realise I've dropped the ball here.

How about if we wrap this up in a context manager, so you say:

```python
with form_bindings(form, bindings):
   ...
```

And this arranges that all the assemble calls that happen inside this block get the right coefficients.

jrmaddison:
I think that still doesn't resolve the issue of the deferred boundary condition lifting.

This specific issue could be resolved by lifting at the time of the bilinear form assembly and caching the RHS term, but that wouldn't fix other cases where Matrix.a is used after assembly.

Note that the current approach can lead to inconsistencies, e.g.

```
from firedrake import *
import numpy as np

mesh = UnitIntervalMesh(10)
space = FunctionSpace(mesh, "Lagrange", 1)
test, trial = TestFunction(space), TrialFunction(space)

F = Function(space, name="F")
F.assign(2.0)

bc = DirichletBC(space, 2.0, "on_boundary")

A = assemble(inner(test, F * trial) * dx, bcs=bc)
b = assemble(inner(test, Constant(1.0)) * dx)
solver = LinearSolver(A, solver_parameters={"ksp_type": "preonly",
                                            "pc_type": "cholesky"})

F.assign(1.0)  # Introduce an inconsistency between A and A.a

u = Function(space, name="u")
solver.solve(u, b)  # Boundary condition applied incorrectly

u_ref = Function(space, name="u_ref")
solve(inner(test, F * trial) * dx == inner(test, Constant(1.0)) * dx,
      u_ref, bc, solver_parameters={"ksp_type": "preonly",
                                    "pc_type": "cholesky"})

# Non-zero error
print(f"error norm = "
      f"{np.sqrt(assemble(inner(u - u_ref, u - u_ref) * dx)):.16e}")
```

wence-:
Given that dolfin are also throwing out backward compat (and this has never been an API compatibility we had anyway), perhaps we should think about how to redesign the linear solver interface so this all works "as expected".

PR #1520, opened 2019-10-18
wence/feature/assemble diagonal

wence-:
Allows assembly of the diagonal of an element tensor. Useful for matrix-free multigrid. Needs firedrakeproject/tsfc#200. Hooks up `getDiagonal` for `matfree` matrices so that one can do:
```
-pc_type mg
-mat_type matfree
-mg_levels_pc_type jacobi
``` 

@sv2518 this may be relevant eventually.


wence-:
Added tests, let's wait for jenkins.

PR #1522, opened 2019-10-22
change plotting API to match matplotlib

### LINKED ISSUE #1201: Colour bar handling for plotting tools
As a feature request it would be nice to be able to:
  1) pass a flag to firedrake.plot that turns off the colour bar (cbar=False, for example)
  2) pass an axes object for the colour bar (i.e. the cax argument in matplotlib.pyplot.colorbar) (cax=Ax_object, for example)

danshapero:
Adds functions that more closely mimic the matplotlib API. This makes it easier for the user to control things like colorbar placement and labeling. For example it would resolve #1201 .

The only significant departure from matplotlib is that the triplot function colors the boundary edges differently depending on their ID. You can then add a legend to the plot to see the correspondence.

I can add features I've missed (quad mesh plots, 3D plots) but wanted to get feedback first.


firedrake/plot.py
```diff
@@ -7,6 +9,137 @@
...
+    if (mesh.geometric_dimension() != 2) or (mesh.topological_dimension() != 2):
+        raise NotImplementedError("Plotting meshes only implemented for 2D")
+
+    if mesh.ufl_cell().cellname() == 'quadrilateral':
+        raise NotImplementedError('Plotting meshes only implemented for '
+                                  'triangles')
+
+    axes = axes if axes is not None else plt.gca()
+
+    mesh.init()  # Apparently this doesn't happen automatically?
```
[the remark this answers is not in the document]
danshapero: Woops, I blindly copied that over from icepack and I don't think it's necessary any longer.

danshapero:
@wence- This is ready for another look, I think it covers all or most of the existing features.

wence-:
Can you merge master please? The most recent fail was because of eigen migrating to gitlab.

docs/notebooks/01-spd-helmholtz.ipynb
```diff
@@ -1111,7 +1114,8 @@
    "cell_type": "markdown",
    "metadata": {},
    "source": [
-    "Let's have a look at the solution, remember we can `plot` both meshes and Functions."
+    "Let's have a look at the solution.\n",
+    "The built-in plotting functions correspond with those from matplotlib, like `tripcolor`, `tricontour`, `tricontourf`."
```
wence-: Link to matplotlib docs? (e.g. https://matplotlib.org/3.1.1/api/_as_gen/matplotlib.pyplot.tripcolor.html)

firedrake/plot.py
```diff
@@ -1,373 +1,279 @@
...
+        axes = figure.add_subplot(projection='3d')
 
-    def display_plot(index):
-        axes.clear()
-        return plot(functions[index], num_sample_points, axes=axes, **kwargs).figure
+    if len(function.ufl_shape) == 1:
+        mesh = function.ufl_domain()
+        element = function.ufl_element().sub_elements()[0]
+        Q = FunctionSpace(mesh, element)
+        function = interpolate(sqrt(inner(function, function)), Q)
```
wence-: Similarly here.

firedrake/plot.py
```diff
@@ -1,373 +1,279 @@
...
+
+    axes = kwargs.pop("axes", None)
+    if axes is None:
+        figure = plt.figure()
+        axes = figure.add_subplot()
+
+    coords = function.ufl_domain().coordinates.dat.data_ro
+    vals = np.asarray(function.at(coords, tolerance=1e-10))
+    C = np.linalg.norm(vals, axis=1)
+    return axes.quiver(*(coords.T), *(vals.T), C, *args, **kwargs)
```
wence-: For the quiver plots, we just interpolate the function at the coordinates and don't do any sub-sampling. I wonder if we should do the same thing for the triangulations (especially now that paraview output supports high-order viz)?

WDYT?

docs/notebooks/01-spd-helmholtz.ipynb
danshapero: Done! I used the URL that goes to the current version of matplotlib (which of course will change) rather than specifically 3.1.1.

firedrake/plot.py
danshapero: I'm not sure I understand the suggestion -- are you thinking we should use `function.at` on the mesh coordinates instead of subsampling for functions like tricontour, tricontourf, etc? Right now these functions call out to `_two_dimension_triangle_func_val` and then to `_calculate_values`. If it's ok with you I'd rather use this PR to change the API to match matplotlib, and then we can worry about simplifying the internal workings of the code later.

Specifically for quiver, subsampling is a bad idea because the spacing of the arrows can be so dense that the plot is impossible to parse visually.

dham: At least we should not be using `function.at`.  `at` discards all the mesh topology information and does geometric searches for elements. This is structurally slow. Instead one should use:
`vals = interpolate(function, coords.function_space())`

danshapero: I thought the use of `.at` there was a little fishy -- I fixed it as per your suggestion.

firedrake/plot.py
```diff
@@ -1,373 +1,280 @@
...
         figure = plt.figure()
-        axes = figure.add_subplot(111, projection=projection, **kwargs)
+        axes = figure.add_subplot()
 
-    lines = Lines(vertices, colors=colors)
+    if len(function.ufl_shape) == 1:
+        mesh = function.ufl_domain()
+        element = function.ufl_element().sub_elements()[0]
+        Q = FunctionSpace(mesh, element)
+        function = interpolate(sqrt(inner(function, function)), Q)
```
ReubenHill: That this is what's actually plotted should be noted in the docstring for the functions plot using `_plot_2d_field`

firedrake/plot.py
```diff
@@ -1,373 +1,280 @@
...
-    :arg kwargs: Additional keyword arguments passed to
-        ``matplotlib.plot``.
+    method = getattr(axes, method_name)
+    return method(triangulation, vals, *args, **kwargs)
+
+
+def tricontourf(function, *args, **kwargs):
+    r"""Create a filled contour plot of a finite element field
+
+    :arg function: the function to plot
```
ReubenHill: Unless the capability of this is wider than I understand, this should read
```python
r"""Create a filled contour plot of a Firedrake :class:`~.Function` 

:arg function: the Firedrake :class:`~.Function` to plot
```

firedrake/plot.py
```diff
@@ -1,373 +1,280 @@
...
-    Jupyter Notebook
+def tricontour(function, *args, **kwargs):
+    r"""Create a contour plot of a finite element field
 
-    :arg functions: 1D Functions to be plotted
-    :arg num_sample_points: Number of sample points per element, ignore if
-        degree < 4 where Bezier curve is used for an exact plot
-    :arg kwargs: Additional key word arguments to be passed to
-        ``matplotlib.plot``
+    :arg function: the function to plot
```
ReubenHill: Similarly here

firedrake/plot.py
```diff
@@ -1,373 +1,280 @@
...
-        from ipywidgets import interact, IntSlider
-    except ImportError:
-        raise RuntimeError("Not in notebook")
+    return _plot_2d_field("tricontour", function, *args, **kwargs)
 
+
+def trisurf(function, *args, **kwargs):
+    r"""Create a 3D surface plot of a finite element field
+
+    :arg function: the function to plot
```
ReubenHill: And here

firedrake/plot.py
```diff
@@ -1,373 +1,280 @@
...
+    :return:
```
ReubenHill: Missing note of what is returned

firedrake/plot.py
```diff
@@ -1,373 +1,280 @@
...
+        axes = figure.add_subplot(projection='3d')
 
-    def display_plot(index):
-        axes.clear()
-        return plot(functions[index], num_sample_points, axes=axes, **kwargs).figure
+    if len(function.ufl_shape) == 1:
+        mesh = function.ufl_domain()
+        element = function.ufl_element().sub_elements()[0]
+        Q = FunctionSpace(mesh, element)
+        function = interpolate(sqrt(inner(function, function)), Q)
```
ReubenHill: See previous note about documenting what this actually plots

firedrake/plot.py
```diff
@@ -1,373 +1,280 @@
...
 
-    :arg points: Points to be plotted
-    :arg axes: Axes to be plotted on
-    :arg kwargs: Additional key word arguments passed to plot
+
+def tripcolor(function, *args, **kwargs):
+    r"""Create a pseudo-color plot of a finite element field
+
+    :arg function: the function to plot
+    :return: matplotlib Collection object
```
ReubenHill: See previous notes about specifying that this plots a Firedrake function (assuming that I've understood capability correctly).

firedrake/plot.py
```diff
@@ -1,373 +1,280 @@
...
+    """
+    return _plot_2d_field("tripcolor", function, *args, **kwargs)
+
+
+def quiver(function, *args, **kwargs):
+    r"""Make a quiver plot of a vector field
+
+    :arg function: the vector field to plot
```
ReubenHill: And here

docs/source/visualisation.rst
```diff
@@ -251,38 +251,55 @@ product of elements of degree 4 and 4.
...
-keyword argument ``interactive=True`` when calling the function 
-:func:`plot <firedrake.plot.plot>` will generate an interactive slider for
-selecting the figures. 
-
-For 2D functions, both surface plots and contour plots are supported. By
-default, the :func:`plot <firedrake.plot.plot>` will return a surface plot in the
-colour map of coolwarm. Contour plotting could be enabled by passing the keyword
-argument ``contour=True``.
+To install matplotlib_, please look at the installation instructions of
+matplotlib.
```
ReubenHill: No quiver plot documentation?

firedrake/plot.py
```diff
@@ -1,373 +1,280 @@
...
+    V = function.ufl_domain().coordinates.function_space()
+    vals = interpolate(function, V).dat.data_ro
+    C = np.linalg.norm(vals, axis=1)
+    return axes.quiver(*(coords.T), *(vals.T), C, *args, **kwargs)
+
+
+def plot(function, *args, bezier=False, num_sample_points=10, **kwargs):
+    r"""Plot a Firedrake object.
+
+    :arg function_or_mesh: The :class:`~.Function` to plot
```
ReubenHill: This can be updated now to note limitations. Presumably "Firedrake object" is now too broad a term.

There's also a bunch of undocumented arguments here that should be documented

firedrake/plot.py
```diff
@@ -560,102 +413,75 @@ def _bezier_calculate_points(function):
...
 
-def bezier_plot(function, axes=None, **kwargs):
+def _bezier_plot(function, axes, **kwargs):
     """Plot a 1D function on a function space with order no more than 4 using
     Bezier curve within each cell, return a matplotlib axes
 
     :arg function: 1D function for plotting
-    :arg axes: Axes for plotting, if None, a new one will be created
+    :arg axes: Axes for plotting
     :arg kwargs: additional key work arguments to plot
```
ReubenHill: Worth updating this to note it's a firedrake function as elsewhere. Also if it's a matplotlib `Axes` object, that can be specifically noted/linked to

firedrake/plot.py
```diff
@@ -1,373 +1,280 @@
...
+            return Poly3DCollection, Poly3DCollection
+
+    raise ValueError("Geometric dimension must be either 2 or 3!")
+
+
+def triplot(mesh, axes=None, interior_kw={}, boundary_kw={}):
+    r"""Plot a mesh with a different color for each boundary segment
+
+    :arg mesh: mesh to be plotted
+    :arg axes: axis object on which to plot mesh
```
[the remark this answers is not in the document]
ReubenHill: The `Axes` object note would ideally link to the matplotlib API documentation

firedrake/plot.py
ReubenHill: Also missing documentation for `args` and `kwargs`

firedrake/plot.py
ReubenHill: Also missing documentation for `args` and `kwargs`

firedrake/plot.py
ReubenHill: Also missing documentation for `args` and `kwargs`

firedrake/plot.py
ReubenHill: Also missing documentation for `args` and `kwargs`

tests/output/test_plotting.py
```diff
@@ -0,0 +1,169 @@
+from firedrake import *
+import matplotlib.pyplot as plt
+from mpl_toolkits.mplot3d import Axes3D
+
+
```
ReubenHill: It's wonderful that this has tests, given that it's a plotting API. Could you perhaps add tests for `plot`, e.g. that it only works when it should?

firedrake/plot.py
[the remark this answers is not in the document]
danshapero: You're right about the naming (axes/ax) -- the current Firedrake plotting API uses "axes", so I stuck to that convention rather than change it to the one used in matplotlib. I can see your case for changing it to "ax" though. @wence- , any thoughts?

firedrake/plot.py
```diff
@@ -1,373 +1,280 @@
...
+    :arg interior_kw: keyword arguments to matplotlib collection for mesh interior
+    :arg boundary_kw: keyword arguments to matplotlib collection for mesh boundary
```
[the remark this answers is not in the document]
danshapero: So a typical use case for this is that you want the line thickness to be different for the boundary as for the interior. I had to do a bit of indirection in order to not have total spaghetti code here, but it gets passed to LineCollection, PolyCollection, etc. I'll change the wording to clarify.

firedrake/plot.py
```diff
@@ -1,373 +1,280 @@
...
     coordinates = mesh.coordinates
-    ele = coordinates.function_space().ufl_element()
-    if ele.degree() != 1:
+    element = coordinates.function_space().ufl_element()
+    if element.degree() != 1:
         # Interpolate to piecewise linear.
-        from firedrake import VectorFunctionSpace, interpolate
-        V = VectorFunctionSpace(mesh, ele.family(), 1)
+        V = VectorFunctionSpace(mesh, element.family(), 1)
         coordinates = interpolate(coordinates, V)
```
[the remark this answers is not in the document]
danshapero: I'm not sure I understand your example -- since f is zero-initialized the mesh will have no interior. I don't know what the right way to plot that would be. This code was to handle cases where the mesh has degree 2 or more. Plotting curved meshes is for a later PR ;)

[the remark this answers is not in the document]
ReubenHill: In the example I was trying to find a minimal version of something like changing mesh coordinates as detailed [here](https://www.firedrakeproject.org/mesh-coordinates.html) with the aim of working out what the point of the highlighted line is. But nothing in this PR is broken - plotting the mesh in the existing API also doesn't gives me an empty box (with a dot in the centre) too so you've at least maintained existing behaviour! Marking as resolved.

firedrake/plot.py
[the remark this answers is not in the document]
ReubenHill: Thinking about it, we've already deviated from matplotlib's `triplot` API by specifying the `Axes` as a keyword argument instead of the default first argument. I therefore lean towards maintaining the firedrake naming to avoid confusion and to maintain consistent naming within our own API. The documentation for `Firedrake.triplot` makes it clear what this refers to.

@danshapero if you agree then feel free to resolve this.

danshapero:
Note that this PR breaks backward compatibility with the existing `plot` function. Before, `plot` would take in meshes and functions of any dimension. In keeping with the matplotlib API, `plot` in this PR will only plot a 1D function. If a user passes in a mesh or 2D function, the error message for plot will direct them to triplot, tripcolor, etc.

PR #1524, opened 2019-10-23
Pointwise operator

nbouziani:
PointwiseOperator is the Firedrake bit of the ExternalOperator concept introduced in UFL (c.f. the associated pull request)

It handles the practical implementation thereof, especially it implements some of the subclasses covered by this concept : 

- PointexprOperator: Operator defined through an externally defined expression (e.g. lambda expression) passed in. This evaluation of this operator is done by merely evaluate this expression pointwise.

- PointsolveOperator:  Solve an externally defined algebraic relationship pointwise. For instance one can think to an implicit constitutive law that is part of a given PDE-based model.

There are more classes of function that stem from the ExternalOperator concept, but this pull request ONLY concern the architecture implemented to handle the AbstractPointwiseOperator (abstract base class) and 2 applications of it (the `point_solve` and the `point_expr`)


firedrake/pointwise_operators.py
```diff
@@ -0,0 +1,485 @@
...
+from firedrake import utils, functionspaceimpl
+from pyop2.datatypes import ScalarType
+from ufl.log import error
+import sympy as sp
+import numpy as np
+from scipy import optimize
+
+
+class AbstractPointwiseOperator(Function, ExternalOperator, metaclass=ABCMeta):
+    r"""Asbtract base class from wich stem from all the firedrake practical implementation of the
```
dham: Typo: Abstract

dham: also "which"

dham: the second "from" is spurious
Firedrake has a capital F
implementations (plural)

firedrake/pointwise_operators.py
```diff
@@ -0,0 +1,485 @@
...
+            fs = ofs
+
+        Function.__init__(self, fs, val, name, dtype, count=self._count)  # count has been initialised in ExternalOperator.__init__
+
+        self._ufl_function_space = aa
+        self._original_function_space = fs
+        self.operator_data = operator_data
+
+        if extop_id is not None:
+            self.extop_id = extop_id
```
dham: What does this do? I am a little scared of the caching implications of operator IDs.

firedrake/pointwise_operators.py
```diff
@@ -0,0 +1,485 @@
...
+
+    @abstractmethod
+    def compute_derivatives(self):
+        """apply the derivatives on operator_data"""
+
+    @abstractmethod
+    def evaluate(self):
```
dham: Random thought: do we do dependency analysis to make sure point wise operators get evaluated in the right order? Think about point wise operators with point wise operators in their arguments.

firedrake/pointwise_operators.py
nbouziani: It is a deprecated feature that I do not use anymore. I am wondering why I did not remove it, I will do it.

firedrake/pointwise_operators.py
nbouziani: I do not think we need to do it. The order in which the ExtOps get evaluated is based on the order on which they appear in the coefficients list of the form. And for external operator nesting, they will appear from bottom to top (cf. ufl/algorithms/analysis.py: extract_type function) so if I have two external operators p1 and p2 and a coefficient u and I consider the form p2(p1(u)), the coefficient list will be [u, p1, p2] and therefore p1 will be evaluated before p2.

firedrake/pointwise_operators.py
nbouziani: Oops, I changed that.

nbouziani: Oops

wence-:
Can this one be closed then?

PR #1525, opened 2019-10-28
wence/feature/embedded transfer default

wence-:
Provided transfer operators for all fields by default. With a new interface to override if you need something special.

@thomasgibson for the gtmg stuff you might need to adopt this new interface.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -31,6 +31,20 @@ pipeline {
         }
       }
     }
+    stage('Setup') {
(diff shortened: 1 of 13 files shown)


wence-:
I finally figured out where the parallel test hang bugs are. I think this is good to merge, please review.

dham:
I still don't understand this PR, but it appears to fix Jenkins so I'm merging it by executive fiat.

PR #1531, opened 2019-11-05
Parallel test for supermesh projection on periodic meshes

jwallwork23:
Is this a useful additional test? I copied the format from `test_nonnested_project.py`.


tests/supermesh/test_periodic.py
```diff
@@ -59,3 +60,38 @@ def test_periodic(shapify, direction, space):
     g.project(g_p)
 
     assert errornorm(f, g) < 1e-8
+
+
+@pytest.mark.parallel(nprocs=3)
+def test_periodic_parallel(shapify, direction, space):
```
wence-: Does this need to be in parallel.

jwallwork23: I was just figuring out how to use project in parallel and thought it might be useful. Can get rid of it if not.

wence-: If it's just to document some aspect, it's probably better in the documentation. We're trying to avoid adding tests that don't really add new coverage (because the test suite continues to take a long time).

jwallwork23: Okay, no worries!

PR #1539, opened 2019-11-11
Some install script fixes

benSepanski:
I think these are leftover from the merge

### DIFF
--- a/scripts/firedrake-install
+++ b/scripts/firedrake-install
@@ -54,9 +54,8 @@ class FiredrakeConfiguration(dict):
                            "minimal_petsc", "mpicc", "mpicxx", "mpif90", "mpiexec", "disable_ssh",
                            "honour_petsc_dir", "with_parmetis",
                            "slepc", "packages", "honour_pythonpath",
-                           "petsc_int_type", "cache_dir", "complex"]
-                           "opencascade",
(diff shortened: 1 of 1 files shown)


scripts/firedrake-install
```diff
@@ -492,15 +491,13 @@ if options["petsc_int_type"] == "int32":
     # Serial mesh partitioner
     petsc_options.add("--download-chaco")
     if not options["minimal_petsc"]:
-        # AMG
-        petsc_options.add("--download-ml")
```
dham: Why is it necessary to remove this line?

benSepanski: When I tried to install with ml, I got the following error in my `firedrake-install.log`

*******************************************************************************
             UNABLE to CONFIGURE with GIVEN OPTIONS    (see configure.log for details):
    -------------------------------------------------------------------------------
    Cannot use ml with complex numbers it is not coded for this capability
    *******************************************************************************
Would it be better to still install ml if real PETSc is built?

dham: Yes, It's the same procedure as for hypre, you take the configure option out of the set in the complex case.

benSepanski: Okay, I pushed a fix

dham:
Many thanks!

PR #1547, opened 2019-11-20
Add support for File("filename.pvd").write(mesh)

pefarrell:
Test with 

```
from firedrake import *
mesh = UnitCubeMesh(10, 10, 10)
File("output.pvd").write(mesh)
```


firedrake/output.py
```diff
@@ -466,9 +466,18 @@ def _prepare_output(self, function, max_elem):
...
     def _write_vtu(self, *functions):
         from firedrake.function import Function
+        from ufl import Mesh
+
+        # Check if the user has requested to write out a plain mesh
+        if len(functions) == 1 and isinstance(functions[0], ufl.Mesh):
+            from firedrake.functionspace import FunctionSpace
+            mesh = functions[0]
+            V = FunctionSpace(mesh, "CG", 1)
+            functions = [Function(V)]
```
wence-: I guess you could write the coordinate field instead of an empty function. Ideally, we would not write a function at all, but perhaps that is a bigger refactoring?

pefarrell: I agree that that would be the ideal, but it was a bigger refactoring than I was willing to do.

As for writing out the coordinate field: I don't feel strongly. You can get the coordinate field out via the calculator in paraview, but maybe it's useful. I chose a function of all zeroes with the vague idea that it would compress better.

PR #1550, opened 2019-11-20
Install VTK from source on non-x86_64 systems

dacreman:
The VTK library is pip installed as a dependency of firedrake. However it is distributed as pre-built binary wheels for x86_64 systems only.  This PR adds a build_and_install_vtk function which builds VTK from source if the architecture is not x86_64 (otherwise the function does nothing). 

This change has been tested on Isambard (the GW4 ARM64 system) and includes logic for handling ARM64 and Cray specific options to the VTK build.


scripts/firedrake-install
```diff
@@ -1006,6 +1006,114 @@ def build_and_install_slepc():
...
+    version="8.2.0",
+    author="Kitware",
+    author_email="kitware@kitware.com",
+    description="The Visualization Toolkit (VTK)",
+    long_description="The Visualization Toolkit (VTK)",
+    long_description_content_type="text/markdown",
+    url="https://vtk.org",
+    packages=setuptools.find_packages(),
+    classifiers=[
+        "Operating System :: ARM64",
```
dham: Above we only check for non x86_64, but now we assume that this automatically means ARM. Is that really safe?

scripts/firedrake-install
```diff
@@ -1006,6 +1006,114 @@ def build_and_install_slepc():
         log.info("No need to rebuild slepc4py")
 
 
+def build_and_install_vtk():
```
dham: I'm a bit nervous about having this whole package definition in firedrake-install. Would it be better to have a git repo containing a Python package which does this instead.

wence-:
Hi @dacreman. As discussed on Monday, I think it makes more sense if we don't put this code in firedrake-install, but instead split it out to build the wheels we need. We can then host those wheels somewhere and just point firedrake-install at them as appropriate.

wence-:
@dacreman do you have any idea if you'll have time to make the changes we talked about?

wence-:
This was fixed in #1672 by providing custom wheels in these cases.

PR #1558, opened 2019-12-03
Paraviewfix

wraith1995:
This pull fixes two problems in the current vis system:

1.  The old invert function is not robust enough so a simpler scheme is used. Credit to @florianwechsung  for stumbling upon this.
2.  There was a minor typo in the docs.

In the course of fixing this, I found that if I make vector data with Firedrake and then attempt to view it with Paraview, Paraview segfaults. I'm looking into this...

### DIFF
--- a/firedrake/paraview_reordering.py
+++ b/firedrake/paraview_reordering.py
@@ -46,6 +46,8 @@ def invert(list1, list2):
                 return idx
         raise ValueError("Unable to establish permutation between Paraview basis and given element's basis.")
     perm = [find_same(x, list2) for x in list1]
+    if len(set(perm)) != len(perm):
+        raise ValueError("Unable to establish permutation between Paraview basis and given element's basis.")
     return perm
 
 
--- a/firedrake/paraview_reordering.py
+++ b/firedrake/paraview_reordering.py
@@ -2,6 +2,7 @@ from tsfc.fiatinterface import create_element
 import numpy as np
 from pyop2.utils import as_tuple
 import importlib
+import functools
 """
 This requires an explentation.
 Vtk has some .so deps that might not be present (e.g. libsm.so (X11 Sessions))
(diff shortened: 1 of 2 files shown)


florianwechsung:
I can confirm that this fixes the issue I observed, thanks @wraith1995

firedrake/paraview_reordering.py
```diff
@@ -39,13 +39,16 @@ def invert(list1, list2):
     """
     if len(list1) != len(list2):
         raise ValueError("Dimension of Paraview basis and Element basis unequal.")
-    N = len(list1)
-    pt1, o1 = zip(*sorted(zip(map(tuple, list1), range(N))))
-    pt2, o2 = zip(*sorted(zip(map(tuple, list2), range(N))))
-    if not np.allclose(pt1, pt2):
```
wence-: Could we just adjust the tolerance in the allclose call?

Can you provide a test that exercises this?

wraith1995: Adding a tolerance doesn't work because the problem occurs prior to this; also, I think allclose uses a fairly large tolerance by default. The problem is that we get the paraview nodes with a bit of floating point error in either direction (for degrees >= 5), which leads to situations like this in the sorting:
```
>>> (0.0, 0.1999999999999, 0.8) < (0.0, 0.2, 0.2)
True
>>> (0.0, 0.2, 0.8) < (0.0, 0.2, 0.2)
False
>>> sorted([(0.0, 0.1999999999999, 0.8), (0.0, 0.2, 0.2)])
[(0.0, 0.1999999999999, 0.8), (0.0, 0.2, 0.2)]
>>> sorted([(0.0, 0.2, 0.8), (0.0, 0.2, 0.2)])
[(0.0, 0.2, 0.2), (0.0, 0.2, 0.8)]
```

I will add a test sometime today.

wence-: OK, I see, thanks.

PR #1559, opened 2019-12-03
Add firedrake removal instructions

ReubenHill:
Adds basic instructions for removing firedrake by deleting the install directory to the https://www.firedrakeproject.org/download.html page.

Note that I assume the user knows where their firedrake installation is. I believe this to be reasonable given the context of the rest of the download page.

### DIFF
--- a/docs/source/download.rst
+++ b/docs/source/download.rst
@@ -6,7 +6,7 @@ Firedrake is installed using its install script::
   curl -O https://raw.githubusercontent.com/firedrakeproject/firedrake/master/scripts/firedrake-install
   python3 firedrake-install
 
-Running ``firedrake-install`` with no arguments will install firedrake in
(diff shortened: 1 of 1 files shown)


ReubenHill:
@dham there are capitalisation inconsistencies throughout the download page - maybe worth fixing them all now?

dham:
> @dham there are capitalisation inconsistencies throughout the download page - maybe worth fixing them all now?

Well, there was one other case. I have now fixed it.

ReubenHill:
Titles don't respect title case except for the one I added - I'll push changes in 1 min

dham:
> Titles don't respect title case except for the one I added - I'll push changes in 1 min

As far as I can see the titles are all correctly cased. Titles should be cased like sentences.

ReubenHill:
If that is the convention used here then I will not push any changes

Issue #1544, opened 2019-11-18
Firedrake/Dolfin-adjoint integration


dham:
This issue sketches out the plan for the sprint to integrate Firedrake and Dolfin-adjoint more closely.

## The issues

Dolfin-adjoint subclasses some Firedrake objects. This causes issues if users manage to use unsubclassed objects. They can do this either by importing `firedrake` without realising they also have to import `firedrake_adjoint`, or in more subtle ways by subclassing Firedrake objects.

## The proposed solution

We intend to refactor Dolfin-adjoint and Firedrake as follows:

* Dolfin-adjoint becomes a hard Firedrake dependency.
* Dolfin-adjoint ceases to import Firedrake.
* Dolfin-adjoint subclasses become mixins, and Firedrake always inherits from them.
* Dolfin-adjoint wrapper functions which annotate Firedrake operations (e.g. `solve`) become decorators which are applied in the Firedrake code.
* Annotatiation is switched off by default. This prevents unsuspecting users who are not solving an inverse problem from leaking memory. Users wishing to solve inverse problems will need to explicitly press "record" to start taping.

## Potential issues

* Are there still places where Dolfin-adjoint still needs to call Firedrake?
* Will going through Dolfin-adjoint wrappers cause a measurable performance hit (hopefully no, because it just amounts to checking the value of a global variable).
* How much does this break the commonality of code between `firedrake_adjoint` and `fenics_adjoint`. Given the FEniCS API changes in the pipeline, is this even an issue?

dham:
`Function.split()` currently does not get annotated. This is a Bad Thing.

stephankramer:
Another thing that really needs fixing is `interpolate` it is currently not annotated at all (either the function or the method). It "works" when it's used as initialisation from an analytical expression (e.g. SpatialCoordinate dependent) as pyadjoint automatically checkpoints the function when it's first used and it's not as a block_variable on tape yet - but fails if the interpolate changes some intermediate state of a function, or when the function depends on other functions.

A similar story is true for reading from a `DumbCheckPoint`. Usually this is done as initialisation, and it works (on replay the checkpoint just sits in memory rather then being read from the actual hdf5 file) but when it is read into a function that's already been used as dependency previously (e.g. reading in a time series), this will not work.

stephankramer:
[...]

PR #1567, opened 2019-12-18
Self adjoint

nbouziani:
Following the dolfin-adjoint integration (cf. #1544). This pull request covers all the changes on the firedrake side, namely:

- Addition of Mixins for the Firedrake objects previously overloaded in pyadjoint, these mixins have methods which are used to decorate the methods of the corresponding Firedrake object that need to be annotated, and other new methods so far specific to the pyadjoint world.

- Incorporation of these Mixins and their methods in the corresponding Firedrake class (e.g. Function, DirichletBC)  and function (solve and assemble)

- The addition of a new module `adjoint` containing all the Mixins and pyadjoint Blocks specific to Firedrake (e.g. MeshInputBlock)

- Appropriate changes for subsidiary files (Jenkins, Docker, requirement, firedrake-install)


firedrake/adjoint/constant.py
```diff
@@ -0,0 +1,146 @@
...
+            if annotate:
+                other = args[0]
+                if not isinstance(other, OverloadedType):
+                    other = create_overloaded_object(ndarray(other))
+
+                block = ConstantAssignBlock(self, other)
+                tape = get_working_tape()
+                tape.add_block(block)
+
+            ret = assign(self, *args, **kwargs)
```
wence-: Prefer:
```python
annotate = annotate_tape(kwargs)
ret = assign(self, *args, **kwargs)
if not annotate:
    return ret
...
```

firedrake/adjoint/constant.py
```diff
@@ -0,0 +1,146 @@
...
+    def _ad_mul(self, other):
+        return self._constant_from_values(self.values() * other)
+
+    def _ad_add(self, other):
+        return self._constant_from_values(self.values() + other.values())
+
+    def _ad_dot(self, other, options=None):
+        return sum(self.values() * other.values())
+
+    @staticmethod
```
wence-: static method, yet `self` in argument list.

firedrake/adjoint/dirichletbc.py
```diff
@@ -0,0 +1,42 @@
...
+    @staticmethod
+    def _ad_annotate_init(init):
+        def wrapper(self, *args, **kwargs):
+            FloatingType.__init__(self,
+                                  *args,
+                                  block_class=DirichletBCBlock,
+                                  _ad_args=args,
+                                  _ad_floating_active=True,
+                                  **kwargs)
+            init(self, *args, **kwargs)
```
wence-: This is the concrete `__init__` right? That does stuff with projection for some dirichlet bc calls. Should those be annotated (as, it seems, is happening here?)

firedrake/adjoint/function.py
```diff
@@ -0,0 +1,243 @@
...
+            func = copy(self, *args, **kwargs)
+
+            if annotate:
+                if kwargs.pop("deepcopy", False):
+                    block = FunctionAssignBlock(func, self)
+                    tape = get_working_tape()
+                    tape.add_block(block)
+                    block.add_output(func.create_block_variable())
+                else:
+                    # TODO: Implement. Here we would need to use floating types.
```
wence-: `raise NotImplementedError`

firedrake/adjoint/function.py
```diff
@@ -0,0 +1,243 @@
...
+            return func
+
+        return wrapper
+
+    @staticmethod
+    def _ad_annotate_assign(assign):
+
+        def wrapper(self, other, *args, **kwargs):
+            """To disable the annotation, just pass :py:data:`annotate=False` to this routine, and it acts exactly like the
+            Dolfin assign call."""
```
wence-: Docstring incorrect.

firedrake/adjoint/function.py
```diff
@@ -0,0 +1,243 @@
...
+    @no_annotations
+    def adj_update_value(self, value):
+        self.original_block_variable.checkpoint = value._ad_create_checkpoint()
+
+    @no_annotations
+    def _ad_mul(self, other):
+        from firedrake import Function
+
+        r = Function(self.function_space())
+        Function.assign(r, self * other)
```
wence-: Why not `r.assign(self*other)`?

firedrake/adjoint/function.py
```diff
@@ -0,0 +1,243 @@
...
+
+    @no_annotations
+    def _ad_add(self, other):
+        from firedrake import Function
+
+        r = Function(self.function_space())
+        Function.assign(r, self + other)
+        return r
+
+    def _ad_dot(self, other, options=None):
```
wence-: `@no_annotations` required?

firedrake/adjoint/function.py
```diff
@@ -0,0 +1,243 @@
...
+            m_v = m
+        m_a = m_v.gather()
+
+        return m_a.tolist()
+
+    def _ad_copy(self):
+        from firedrake import Function
+
+        r = Function(self.function_space())
+        Function.assign(r, self)
```
wence-: `@no_annotations`?

firedrake/adjoint/function.py
```diff
@@ -0,0 +1,243 @@
...
+
+    def _ad_imul(self, other):
+        vec = self.vector()
+        vec *= other
+
+    def _ad_iadd(self, other):
+        vec = self.vector()
+        # FIXME: PETSc complains when we add the same vector to itself.
+        # So we make a copy.
+        vec += other.vector().copy()
```
wence-: Disgusting:
```python
ovec = other.vector()
if ovec.dat == vec.dat:
    vec *= 2
else:
   vec += ovec
```

firedrake/adjoint/function.py
```diff
@@ -0,0 +1,243 @@
...
+            r0 = r(vec[i], r0)
+        return r0
+
+    def _applyUnary(self, f):
+        vec = self.vector()
+        npdata = vec.get_local()
+        for i in range(len(npdata)):
+            npdata[i] = f(npdata[i])
+        vec.set_local(npdata)
+        vec.apply("insert")
```
wence-: Line should be removed (no needed).

firedrake/adjoint/function.py
```diff
@@ -0,0 +1,243 @@
...
+
+    def _applyBinary(self, f, y):
+        vec = self.vector()
+        npdata = vec.get_local()
+        npdatay = y.vector().get_local()
+        for i in range(len(npdata)):
+            npdata[i] = f(npdata[i], npdatay[i])
+        vec.set_local(npdata)
+        vec.apply("insert")
```
wence-: Remove (could then remove the compat stuff in `vector.py`?)

firedrake/adjoint/projection.py
```diff
@@ -0,0 +1,34 @@
...
+
+        To disable the annotation of this function, just pass :py:data:`annotate=False`. This is useful in
+        cases where the solve is known to be irrelevant or diagnostic for the purposes of the adjoint
+        computation (such as projecting fields to other function spaces for the purposes of
+        visualisation)."""
+
+        annotate = annotate_tape(kwargs)
+        with stop_annotating():
+            output = project(*args, **kwargs)
+        output = create_overloaded_object(output)
```
wence-: Non-annotated outputs should not be overloaded, I think.

firedrake/adjoint/solving.py
```diff
@@ -0,0 +1,58 @@
...
+
+from firedrake.adjoint.blocks import SolveBlock
+
+# Type dependencies
+
+# TODO: Clean up: some inaccurate comments. Reused code. Confusing naming with dFdm when denoting the control as c.
+
+
+def annotate_solve(solve):
+    """This solve routine wraps the real Dolfin solve call. Its purpose is to annotate the model,
```
wence-: Docstring wrong.

firedrake/adjoint/variational_solver.py
```diff
@@ -0,0 +1,48 @@
...
+            self._ad_args = args
+            self._ad_kwargs = kwargs
+
+        return wrapper
+
+    @staticmethod
+    def _ad_annotate_solve(solve):
+        def wrapper(self, **kwargs):
+            """To disable the annotation, just pass :py:data:`annotate=False` to this routine, and it acts exactly like the
+            Dolfin solve call. This is useful in cases where the solve is known to be irrelevant or diagnostic
```
wence-: Docstring wrong.

firedrake/functionspaceimpl.py
```diff
@@ -54,6 +54,10 @@ def _split(self):
 
     mesh = ufl.FunctionSpace.ufl_domain
 
+    @property
```
wence-: Use cached_property

firedrake/functionspaceimpl.py
```diff
@@ -345,6 +349,10 @@ def __ne__(self, other):
     def __hash__(self):
         return hash((self.mesh(), self.dof_dset, self.ufl_element()))
 
+    @property
```
wence-: Use cached_property

firedrake/mesh.py
```diff
@@ -979,18 +980,20 @@ def _order_data_by_cell_index(self, column_list, cell_data):
...
 
-class MeshGeometry(ufl.Mesh):
+class MeshGeometry(ufl.Mesh, MeshGeometryMixin):
     """A representation of mesh topology and geometry."""
 
     def __new__(cls, element):
         """Create mesh geometry object."""
         utils._init()
         mesh = super(MeshGeometry, cls).__new__(cls)
+        mesh = super(MeshGeometry, cls).__new__(cls)
```
wence-: I don't understand.

firedrake/mesh.py
```diff
@@ -1025,6 +1028,7 @@ def topological(self):
         return self._topology
 
     @utils.cached_property
+    @MeshGeometryMixin._ad_annotate_coordinates_function
```
wence-: Does this play nicely with the cached_property? I noticed some comments above in the block file.

firedrake/nullspace.py
```diff
@@ -83,6 +84,7 @@ def orthonormalize(self):
                 vec.array -= alpha
             vec.normalize()
         self.check_orthogonality()
+        self._ad_orthogonalized = True
```
wence-: Should set this in orthogonalize too, otherwise you're really claiming orthonormal.

firedrake/variational_solver.py
```diff
@@ -83,6 +86,13 @@ def __init__(self, F, u, bcs=None, J=None,
         self.form_compiler_parameters = form_compiler_parameters
         self._constant_jacobian = False
 
+        # used in NonlinearVariationalSolverMixin._ad_annotate_solve
+        self._ad_F = F
+        self._ad_u = u
+        self._ad_bcs = bcs
+        self._ad_J = J
+        self._ad_kwargs = {'Jp': Jp, 'form_compiler_parameters': form_compiler_parameters, 'is_linear': is_linear}
```
wence-: Since we now control the mixin class, can we not just use the properties that already exist? This sets up a contract between mixin of a related class to the `NLVP` that kind of smells bad.

requirements-git.txt
```diff
@@ -3,3 +3,4 @@ git+https://github.com/firedrakeproject/fiat.git#egg=fiat
 git+https://github.com/FInAT/FInAT.git#egg=finat
 git+https://github.com/firedrakeproject/tsfc.git#egg=tsfc
 git+https://github.com/OP2/PyOP2.git#egg=pyop2
+git+https://github.com/dolfin-adjoint/pyadjoint.git#egg=pyadjoint
```
wence-: Just point at the right branch here (rather than the jenkinsfile?)

firedrake/mesh.py
nbouziani: Not really, this has caused slate's tests to fail because the caching was not done correctly and new coordinates were created for every call of the method. The reason is that `utils.cached_property` will not cache the method but the wrapper returned by the mixin's decorator. The name of the cached function will then be `wrapper`, that is why the name was hacked in the block file. This is a recurrent problem when dealing with composition of decorators.

wence-: We should backport `cached_property` from functools in Py3.8 which handles this behaviour correctly. Can you try with OP2/PyOP2#573 ?

firedrake/variational_solver.py
ksagiyam: Or could we just create `NonlinearVariationalProblemMixin` that only for these lines?

ksagiyam: fixed. I just created `NonlinearVariationalProblemMixin` for now to be consistent with what we do for the other classes.

firedrake/adjoint/blocks.py
```diff
@@ -0,0 +1,108 @@
+from dolfin_adjoint_common.compat import compat
+from dolfin_adjoint_common import blocks
+from pyadjoint.block import Block
+
+
+class Backend:
+    @property
```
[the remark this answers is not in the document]
ksagiyam: fixed

firedrake/adjoint/blocks.py
```diff
@@ -0,0 +1,108 @@
...
+    def backend(self):
+        import firedrake
+        return firedrake
+
+    @property
+    def compat(self):
+        import firedrake
+        return compat(firedrake)
```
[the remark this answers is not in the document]
ksagiyam: fixed

firedrake/adjoint/blocks.py
```diff
@@ -0,0 +1,108 @@
...
+    pass
+
+
+class MeshInputBlock(Block):
+    """
+    Block which links a MeshGeometry to its coordinates, which is a firedrake
+    function.
+    """
+    def __init__(self, mesh):
+        super(MeshInputBlock, self).__init__()
```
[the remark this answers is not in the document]
ksagiyam: fixed

firedrake/adjoint/function.py
ksagiyam: fixed

firedrake/nullspace.py
nbouziani: Done

firedrake/mesh.py
nbouziani: I removed the line, it is a typo

firedrake/adjoint/assembly.py
```diff
@@ -0,0 +1,34 @@
...
+            if annotate:
+                block = AssembleBlock(form)
+
+                tape = get_working_tape()
+                tape.add_block(block)
+
+                block.add_output(output.block_variable)
+        else:
+            # Assembled a vector or matrix
+            output.form = form
```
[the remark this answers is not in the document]
nbouziani: Done

firedrake/adjoint/blocks.py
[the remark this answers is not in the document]
nbouziani: It should be the same problem than for the Mesh._coordinates_function comment since the `utils.cached_property` decorator is called on another decorator, shouldn't it?

firedrake/nullspace.py
nbouziani: Done

firedrake/mesh.py
nbouziani: I removed the line, it is a typo.

firedrake/mesh.py
nbouziani: Ok, I replied in the OP2 PR.

firedrake/adjoint/dirichletbc.py
dham: I think it's at least not harmful to annotate the projections.

firedrake/adjoint/function.py
dham: I don't think so, because this only gets called inside pyadjoint and at that stage annotations are guaranteed to be off.

firedrake/adjoint/function.py
dham: as above.

firedrake/adjoint/projection.py
dham: No, we overload in all cases. The alternative is very unsafe.

PR #1570, opened 2019-12-18
Add instructions for dealing with hanging parallel tests

ReubenHill:
This adds instructions for dealing with hanging parallel tests, as observed by me on MacOS. I have also removed the warning about failing parallel tests that @wence- says is outdated.

Note that I have written the suggested file modification in the same style as commands to run. If you are aware of a more appropriate way of displaying the suggested modification please do say.

### DIFF
--- a/docs/source/download.rst
+++ b/docs/source/download.rst
@@ -61,9 +61,6 @@ hosts database at ``/etc/hosts`` to include the entries::
 where ``LOCALHOSTNAME`` is the name returned by running the `hostname` 
 command. Should the local host name change, this may require updating.
 
-Note also that there is currently an issue with the test harness on MacOS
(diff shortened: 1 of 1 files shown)


docs/source/download.rst
```diff
@@ -49,9 +49,18 @@ venv_ as above and then run::
   cd firedrake/src/firedrake
   make alltest
 
-Note that there is currently an issue with the test harness on MacOS
-which causes the parallel tests to fail. Firedrake will still work in
-parallel, it is only the test system which is affected.
+Note that there is a known issue where parallel tests hang without
```
wence-: Can we put this in a `..note :: ` block? (I know the old one wasn't).

PR #1585, opened 2020-01-24
Remove pragmatic from install script for now

jwallwork23:
Pragmatic is causing install issues


jwallwork23:
Note you can still install with pragmatic. Just add `--download-pragmatic` to `PETSC_CONFIGURE_OPTIONS` flag.

PR #1586, opened 2020-01-24
Fixes for 3 small bugs found in install/update

JDBetteridge:
Found 2 bugs in install when using `icc`:

- PETSc uses wrong compilers sometimes in it's own build procedure
- Old versions of CMake don't correctly detect FORTRAN compiler

Found one bug in update:

- `mpiexec` is using value from `args` but I think this needs to be read from `options` instead, this bug will only occur on update.

See inline comments for details

### DIFF
--- a/scripts/firedrake-install
+++ b/scripts/firedrake-install
@@ -902,7 +902,7 @@ def build_and_install_libsupermesh():
                             "-DMPI_C_COMPILER=" + cc,
                             "-DMPI_CXX_COMPILER=" + cxx,
                             "-DMPI_Fortran_COMPILER=" + f90,
-                            "-DCMAKE_Fortran_COMPILER" + f90])
+                            "-DCMAKE_Fortran_COMPILER=" + f90])
                 check_call(["make"])
                 check_call(["make", "install"])
     else:
@@ -925,7 +925,7 @@ def build_and_install_pythonocc():
                             "-DMPI_C_COMPILER=" + cc,
                             "-DMPI_CXX_COMPILER=" + cxx,
                             "-DMPI_Fortran_COMPILER=" + f90,
-                            "-DCMAKE_Fortran_COMPILER" + f90])
+                            "-DCMAKE_Fortran_COMPILER=" + f90])
                 check_call(["make", "install"])
     else:
         log.info("No need to rebuild pythonocc-core")
--- a/scripts/firedrake-install
+++ b/scripts/firedrake-install
@@ -502,6 +502,9 @@ else:
 
 if options.get("mpiexec") is not None:
     petsc_options.add("--with-mpiexec={}".format(options["mpiexec"]))
+    petsc_options.add("--with-cc={}".format(options["mpicc"]))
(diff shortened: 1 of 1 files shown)


scripts/firedrake-install
```diff
@@ -502,6 +502,9 @@ else:
 
 if options.get("mpiexec") is not None:
     petsc_options.add("--with-mpiexec={}".format(options["mpiexec"]))
+    petsc_options.add("--with-cc={}".format(options["mpicc"]))
+    petsc_options.add("--with-cxx={}".format(options["mpicxx"]))
+    petsc_options.add("--with-fc={}".format(options["mpif90"]))
```
JDBetteridge: All (or none) of `args.mpicc and args.mpicxx and args.mpif90 and args.mpiexec` must be specified when invoking `firedrake-install`, is there any harm in passing them all to PETSc configure?

This ensures the correct compilers are used for all the built packages.

scripts/firedrake-install
```diff
@@ -898,7 +901,8 @@ def build_and_install_libsupermesh():
                             "-DCMAKE_INSTALL_PREFIX=" + firedrake_env,
                             "-DMPI_C_COMPILER=" + cc,
                             "-DMPI_CXX_COMPILER=" + cxx,
-                            "-DMPI_Fortran_COMPILER=" + f90])
+                            "-DMPI_Fortran_COMPILER=" + f90,
+                            "-DCMAKE_Fortran_COMPILER" + f90])
```
JDBetteridge: Old (but still installed on LTS Ubuntu, HPC etc...) versions of CMake fail to detect the right FORTRAN compiler, this is an issue for instance when using the Intel compiler, which requires a GCC compiler on path for C++11 support required elsewhere. CMake will pick up the GCC FORTRAN and then fail to build.

Is there any harm with manually specifying the FORTRAN compiler here?

scripts/firedrake-install
```diff
@@ -920,7 +924,8 @@ def build_and_install_pythonocc():
                             "-DPYTHON_EXECUTABLE=" + python[0],
                             "-DMPI_C_COMPILER=" + cc,
                             "-DMPI_CXX_COMPILER=" + cxx,
-                            "-DMPI_Fortran_COMPILER=" + f90])
+                            "-DMPI_Fortran_COMPILER=" + f90,
+                            "-DCMAKE_Fortran_COMPILER" + f90])
```
JDBetteridge: See above comment

scripts/firedrake-install
```diff
@@ -1250,7 +1255,7 @@ for opt in ["mpicc", "mpicxx", "mpif90"]:
     os.symlink(src, dest)
 
 if options.get("mpiexec"):
-    mpiexec_loc = shutil.which(args.mpiexec)
+    mpiexec_loc = shutil.which(options.get("mpiexec"))
```
JDBetteridge: `firedrake-update` will fail here if you previously installed with manually specified system MPI, since `args.mpiexec` is not present, but `options['mpiexec']` is

PR #1589, opened 2020-01-24
This fixes the problem Gusto is having with creating a MixedFunctionS…

jshipton:
Further to the discussion on Slack, this fixes the problem Gusto is having with creating a `MixedFunctionSpace` on an `ExtrudedMesh` based on a `PeriodicIntervalMesh`.

The problem arose from a recent change to the `ExtrudedMesh` class which now creates the `TensorProductElement` using `helement=mesh._coordinates.ufl_element().sub_elements()[0]`. Previously just `hfamily` and `hdegree` were taken from `mesh._coordinates.ufl_element()` and passed into `VectorFunctionSpace` which then created the `TensorProductElement` using `topology.ufl_cell()` to provide the cell.

My 'fix' is to reconstruct `helement`, passing in `cell=topology.ufl_cell()`.

I am not sure what's the right thing to do for radial or custom extrusion.

I would like to add a test for this - I can't see that any of the extruded tests use a periodic mesh. Should I add a new test in the extruded directory?

### DIFF
--- a/firedrake/mesh.py
+++ b/firedrake/mesh.py
@@ -1479,6 +1479,8 @@ def ExtrudedMesh(mesh, layers, layer_height=None, extrusion_type='uniform', kern
     helement = mesh._coordinates.ufl_element().sub_elements()[0]
     if extrusion_type == 'radial_hedgehog':
         helement = helement.reconstruct(family="DG", variant="equispaced")
+    elif extrusion_type == 'uniform':
+        helement = helement.reconstruct(cell=topology.ufl_cell().sub_cells()[0])
     velement = ufl.FiniteElement("Lagrange", ufl.interval, 1)
     element = ufl.TensorProductElement(helement, velement)
 


jshipton:
I'm wondering now if @dham was right when he suggested that this should be fixed when the periodic mesh is constructed - I don't know enough about the code but it seems like the 2Dness is a hang over from the circular mesh and should be got rid of as early as possible...

dham:
I think I might have been right. This change looks like a bandaid affixed to an already broken situation.

jshipton:
> I think I might have been right. This change looks like a bandaid affixed to an already broken situation.

I'm willing to accept that you were right. Happy to fix. Shall I continue here or start afresh? Would you like a test for the Periodic 2D + extruded + mixed function space combination and where do you think it fits in?

dham:
It's a one line PR so you might as well start over. In terms of a test, I think you could just create a new file in the `tests/extrusion` directory. If I understand the issue in Gusto correctly, all the test would have to do is extrude a periodic interval and create a mixed function space on it. The test passes if that doesn't raise an exception!

jshipton:
Yes, that's all that needs testing.

PR #1591, opened 2020-01-28
Ensure that PeriodicIntervalMesh coordinate function space has an int…

jshipton:
…erval cell, rather than the 2D interval left over from CircleManifoldMesh and add a test. Note: test is broken in 2D

Weird thing: `error("Sub elements must live on the same cell.")` in the UFL MixedElement class is only triggered in the very special case demonstrated in the test - i.e. when a function space created using a `BrokenElement` is passed in to `MixedFunctionSpace`.

I'm not sure what to do with the 2D case - is there an equivalent of `interval`?


dham:
The 2D cells are `triangle` and `quadrilateral`. Which one you use depends on whether the user passed `quadrilateral=True` to the constructor.

PR #1596, opened 2020-01-30
Add a mark_plex callback to label facets of the plex before building periodic meshes

pefarrell:
Useful for colouring periodic meshes.

Adds a test to make sure we can colour interior facets of periodic meshes.

I'm happy to bikeshed on the name `mark_plex`.


pefarrell:
Update: this doesn't work in parallel, so hold off.

PR #1597, opened 2020-01-31
Allow transfer managers to be set from options

pefarrell:
The `SNESContext` now looks up the options dictionary with its options path to see if the user has specified a transfer manager there. This allows one to set e.g.

```
solver_parameters = {"ksp_type": "cg", "pc_type": "mg", "mg_transfer_manager": __name__ + ".manager"}
```

to use the object `manager` in the current module to govern intergrid transfers. One can also specify a class name or function and it will be constructed/called with no arguments to make a transfer manager.

This allows for the setting of transfer managers on auxiliary problems defined via `AuxiliaryOperatorPC` (before this PR there was no way for the user to specify that a custom transfer manager should be used).

### DIFF
--- a/firedrake/mg/embedded.py
+++ b/firedrake/mg/embedded.py
@@ -18,12 +18,6 @@ class Op(IntEnum):
     INJECT = 2
 
 
-def is_native(element):
-    if isinstance(element.cell(), ufl.TensorProductCell):
-        return reduce(and_, map(is_native, element.sub_elements()))
-    return element.family() in native
-
-
 class TransferManager(object):
     class Cache(object):
         """A caching object for work vectors and matrices.
@@ -71,11 +65,18 @@ class TransferManager(object):
         self.use_averaging = use_averaging
         self.caches = {}
 
+    def is_native(self, element):
+        if element in self.native_transfers.keys():
+            return True
+        if isinstance(element.cell(), ufl.TensorProductCell):
+            return reduce(and_, map(self.is_native, element.sub_elements()))
+        return element.family() in native
+
     def _native_transfer(self, element, op):
         try:
             return self.native_transfers[element][op]
         except KeyError:
-            if is_native(element):
+            if self.is_native(element):
                 ops = firedrake.prolong, firedrake.restrict, firedrake.inject
                 return self.native_transfers.setdefault(element, ops)[op]
         return None
@@ -220,7 +221,7 @@ class TransferManager(object):
 
         source_element = Vs.ufl_element()
         target_element = Vt.ufl_element()
-        if is_native(source_element) and is_native(target_element):
+        if self.is_native(source_element) and self.is_native(target_element):
             return self._native_transfer(source_element, transfer_op)(source, target)
         if type(source_element) is ufl.MixedElement:
             assert type(target_element) is ufl.MixedElement
@@ -282,7 +283,7 @@ class TransferManager(object):
 
         source_element = Vf.ufl_element()
         target_element = Vc.ufl_element()
-        if is_native(source_element) and is_native(target_element):
+        if self.is_native(source_element) and self.is_native(target_element):
             return self._native_transfer(source_element, Op.RESTRICT)(gf, gc)
         if type(source_element) is ufl.MixedElement:
             assert type(target_element) is ufl.MixedElement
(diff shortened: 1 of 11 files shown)


firedrake/mg/ufl_utils.py
```diff
@@ -284,6 +277,31 @@ def coarsen_snescontext(context, self, coefficient_mapping=None):
...
+            V = val.function_space()
+            coarseneddm = V.dm
+            parentdm = get_parent(coarseneddm)
+
+            # Need to make sure parentdm has a __setup_hooks__,
+            # this can happen if e.g. a BC value is assigned on Z.sub(0)
+            stack = parentdm.getAttr("__setup_hooks__")
+            if stack is None:
+                hooks = SetupHooks()
+                push_attr("__setup_hooks__", parentdm, hooks)
```
wence-: I am concerned  that this has the potential to leave the `hooks` object dangling on the parentdm. The top-level usage of `SetupHooks` comes via the `with add_hooks` context manager which ensures that things are cleaned up.

firedrake/solving_utils.py
```diff
@@ -146,8 +146,28 @@ def __init__(self, problem, mat_type, pmat_type, appctx=None,
...
-            from firedrake import TransferManager
-            self._transfer_manager = TransferManager(use_averaging=True)
+            opts = PETSc.Options()
+            prefix = self.options_prefix or ""
+            if opts.hasName(prefix + "mg_transfer_manager"):
+                managername = opts[prefix + "mg_transfer_manager"]
+            elif opts.hasName(prefix + "fas_transfer_manager"):
+                managername = opts[prefix + "fas_transfer_manager"]
+            else:
+                managername = None
```
wence-: Can you add a docstring here explaining the options? I also wonder in what circumstances the options prefix is `None` at this point (I think it is never `None` if the `_SNESContext` was built by `NLVS`).

firedrake/mg/embedded.py
```diff
@@ -71,11 +65,18 @@ def __init__(self, *, native_transfers=None, use_averaging=True):
         self.use_averaging = use_averaging
         self.caches = {}
 
+    def is_native(self, element):
+        if element in self.native_transfers.keys():
+            return True
+        if isinstance(element.cell(), ufl.TensorProductCell):
+            return reduce(and_, map(self.is_native, element.sub_elements()))
+        return element.family() in native
```
wence-: OK.

firedrake/mg/interface.py
```diff
@@ -91,8 +91,8 @@ def restrict(fine_dual, coarse_dual):
         if len(Vc) != len(Vf):
             raise ValueError("Mixed spaces have different lengths")
         for in_, out in zip(fine_dual.split(), coarse_dual.split()):
-            _, myrestrict, _ = firedrake.dmhooks.get_transfer_operators(in_.function_space().dm)
-            myrestrict(in_, out)
+            manager = firedrake.dmhooks.get_transfer_manager(in_.function_space().dm)
```
wence-: I think the way things are now coded we can get the transfer manager once for the outer space (because there is really only one). But maybe not, and this is safer.

firedrake/mg/ufl_utils.py
```diff
@@ -284,6 +277,31 @@ def coarsen_snescontext(context, self, coefficient_mapping=None):
...
+    from firedrake.dmhooks import add_hook, get_parent, SetupHooks, push_attr
+    from itertools import chain
+    for val in chain(coefficient_mapping.values(), (bc._original_val for bc in problem.bcs)):
+        if isinstance(val, firedrake.function.Function):
+            V = val.function_space()
+            coarseneddm = V.dm
+            parentdm = get_parent(coarseneddm)
+
+            # Need to make sure parentdm has a __setup_hooks__,
+            # this can happen if e.g. a BC value is assigned on Z.sub(0)
```
wence-: I need to think about what this means in terms of design. Probably all of it needs another round of rethinking now that we know more... :(

firedrake/mg/ufl_utils.py
pefarrell: I think I was getting the parent from the wrong object. I think it is fixed now.

firedrake/solving_utils.py
pefarrell: Docstring added.

firedrake/mg/ufl_utils.py
pefarrell: I don't think it's as bad as we initially feared.

wence-:
I added two places where the options_prefix was available for new snes contexts. Hopefully jenkins will report happiness.

wence-:
Thanks!

PR #1601, opened 2020-02-04
Add hdf5 and netcdf dir env var

tommbendall:
This simply adds paths for netcdf and hdf5 variables (used for installing Gusto)


wence-:
Thanks.

PR #1603, opened 2020-02-07
Add reference for IMR paper

jwallwork23:
"Anisotropic Goal-Oriented Mesh Adaptation in Firedrake" conference proceedings reference


wence-:
Thanks.

PR #1611, opened 2020-02-16
Work around homebrew linker issues.

dham:
Homebrew configures Python to add -L/usr/local/lib to the C compiler flags.
Because mpicc sees this as a user configuration, this means that any MPI library
installed by a homebrew will be used in preference to the one associated with
mpicc. This causes build fails if the user has installed open-mpi.

This change manually adds the petsc lib directory to CFLAGS. This will be checked
before /usr/local/lib and hence a petsc installed mpi will be preferred over a
homebrew installed one.


wence-:
I don't know how to test this works, but seems sane.

PR #1626, opened 2020-03-16
pass solver_parameters etc to NonlinearVariationalSolver

florianwechsung:
This fixes https://github.com/dolfin-adjoint/pyadjoint/issues/6 

I have also pushed a test to https://github.com/florianwechsung/pyadjoint/tree/fw/options-nlvs.
Since this fails with the current firedrake, I suggest I open a PR on the pyadjoint side once this is merged here. Or is there a better way?

### DIFF
--- a/firedrake/adjoint/blocks.py
+++ b/firedrake/adjoint/blocks.py
@@ -104,7 +104,7 @@ class NonlinearVariationalSolveBlock(GenericSolveBlock):
         self.solver_params = solver_params.copy()
         self.solver_kwargs = solver_kwargs
 
-        super().__init__(lhs, rhs, func, bcs, **kwargs)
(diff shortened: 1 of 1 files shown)


firedrake/adjoint/blocks.py
```diff
@@ -104,7 +104,7 @@ def __init__(self, equation, func, bcs, problem_J, solver_params, solver_kwargs,
         self.solver_params = solver_params.copy()
         self.solver_kwargs = solver_kwargs
 
-        super().__init__(lhs, rhs, func, bcs, **kwargs)
+        super().__init__(lhs, rhs, func, bcs, **{**solver_kwargs,**kwargs})
```
wence-: Looks great other than the lint fail: space after final comma required.

wence-: Fixed.

PR #1628, opened 2020-03-17
plotting functions on surface meshes

danshapero:
This PR makes it possible to plot functions defined on surfaces embedded into 3D using matplotlib. This required a minor change to the internal plotting API. But the only thing users will see is that now they can call `trisurf` for functions defined on the surface of a sphere, torus, etc.


wence-:
This actually passed on jenkins, so will merge.

Issue #1632, opened 2020-03-23
Adjoint annotation breaks split method


stephankramer:
When running the following code:
```
from firedrake import *
from firedrake_adjoint import *
mesh=UnitSquareMesh(1,1) 
V=FunctionSpace(mesh,"CG",1) 
Z=V*V 
z=Function(V) 
z=Function(Z) 
u,v=z.split() 
u.project(Constant(1.)) 
print(z.dat.data[:])
```
the output is:
```
(array([0., 0., 0., 0.]), array([0., 0., 0., 0.]))
```
instead of the expected:
```
(array([1., 1., 1., 1.]), array([0., 0., 0., 0.]))
```
which is what you get when leaving out the `from firedrake_adjoint import *` line.

The reason seems to be that in the adjoint annotation wrapper for the split method the output from the core split method is wrapped in new Function instances, at which point it looses its connection to the parent `.dat`.

This breaks the Thetis adjoint, as we use the above steps to set initial conditions for velocity and elevation in 2d.

nbouziani:
Yes indeed it is at this point that the connection is lost. Specifying the core split functions as the value argument instead of the main argument when defining the new Function instances enables to keep this connection, that is `Function(u[0].function_space(), u[0])` instead of `Function(u[0])`. 

This has been fixed (cf. https://github.com/firedrakeproject/firedrake/pull/1635) and the associated pyadjoint branch with a new test for that (https://github.com/dolfin-adjoint/pyadjoint/pull/9)

wence-:
@stephankramer can confirm this works?

stephankramer:
Thanks! Works correctly now. Fixes adjoint tests in Thetis

stephankramer:
Sorry, just closed it - but just hit a related issue: it still doesn't work with VectorFunctionSpaces. The following fails:
```
from firedrake import *
from firedrake_adjoint import *
mesh = UnitSquareMesh(1,1)
V = VectorFunctionSpace(mesh, "CG", 1)
u = Function(V)
u.split()
```
I think `num_sub_spaces`(line 50 of firedrake/adjoint/function.py) wants to be something else...does `len(output)` not work in all cases?

nbouziani:
Yes, I think `len(output)` should work in all cases since `num_sub_spaces` is only used to iterate over `output`

stephankramer:
[...]

PR #1635, opened 2020-03-27
Fix adjoint annotation for split (cf. #1632)

nbouziani:
This branch fixes the adjoint annotation split issue #1632

### DIFF
--- a/firedrake/adjoint/function.py
+++ b/firedrake/adjoint/function.py
@@ -52,7 +52,8 @@ class FunctionMixin(FloatingType):
                 output = split(self, *args, **kwargs)
 
             if annotate:
-                output = tuple(firedrake.Function(output[i],
+                output = tuple(firedrake.Function(output[i].function_space(),
+                               output[i],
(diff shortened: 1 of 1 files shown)


nbouziani:
A new test has been added in pyadjoint (cf.https://github.com/dolfin-adjoint/pyadjoint/pull/9 )

firedrake/adjoint/function.py
```diff
@@ -52,7 +52,8 @@ def wrapper(self, *args, **kwargs):
                 output = split(self, *args, **kwargs)
 
             if annotate:
-                output = tuple(firedrake.Function(output[i],
+                output = tuple(firedrake.Function(output[i].function_space(),
+                               output[i],
```
dham: Can we fix the indentation here please? I realise that this is a pre-existing issue but this is very confusing. The arguments to the `Function` constructor should be indented to the opening bracket of that constructor, not the opening bracket of the `tuple`.

nbouziani: Done!

PR #1640, opened 2020-04-01
Replace .geo instead of geo. 

pefarrell:
A user with username containing 'geo' complained that `OpenCascadeMeshHierarchy` was looking in the wrong place for the generated coarse mesh. This commit fixes it.

### DIFF
--- a/firedrake/mg/opencascade_mh.py
+++ b/firedrake/mg/opencascade_mh.py
@@ -72,8 +72,8 @@ def make_coarse_mesh(stepfile, cad, element_size, dim, comm=COMM_WORLD, distribu
 
     curdir = os.path.dirname(stepfile) or os.getcwd()
     stepname = os.path.basename(os.path.splitext(stepfile)[0])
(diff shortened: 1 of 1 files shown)


firedrake/mg/opencascade_mh.py
```diff
@@ -73,7 +73,7 @@ def make_coarse_mesh(stepfile, cad, element_size, dim, comm=COMM_WORLD, distribu
     curdir = os.path.dirname(stepfile) or os.getcwd()
     stepname = os.path.basename(os.path.splitext(stepfile)[0])
     geopath = os.path.join(curdir, "coarse-%s.geo" % (stepname))
-    mshpath = geopath.replace("geo", "msh")
+    mshpath = geopath.replace(".geo", ".msh")
```
wence-, line 76: Why not `mshpath = os.path.join(curdir, "coarse-%s.msh" % stepname)` which is safer again?

pefarrell: Good point!

PR #1646, opened 2020-04-07
Added comment about garbage collection

JDBetteridge:
Documents how to turn of Python's garbage collector for future reference.

### DIFF
--- a/docs/source/parallelism.rst
+++ b/docs/source/parallelism.rst
@@ -42,6 +42,32 @@ bandwidth* of the machine.  You can measure how the achieved memory
 bandwidth changes depending on the number of processes used on your
 machine using STREAMS_.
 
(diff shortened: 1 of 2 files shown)


docs/source/parallelism.rst
```diff
@@ -42,6 +42,32 @@ bandwidth* of the machine.  You can measure how the achieved memory
...
+The garbage collector can be turned back on with this line:
+
+.. code-block:: python
+
+    gc.enable()
+
+More information can be found in
+`this <https://github.com/firedrakeproject/firedrake/issues/1617>`_
+issue.
+
```
dham, line 70: This is basically good. Can we add a warning (using `.. warning::`) that disabling the garbage collector may cause memory leaks?

PR #1647, opened 2020-04-07
Fix annotation of split() of non-mixed function.

### LINKED ISSUE #1632: Adjoint annotation breaks split method

stephankramer:
Fixes #1632

### DIFF
--- a/firedrake/adjoint/function.py
+++ b/firedrake/adjoint/function.py
@@ -47,7 +47,6 @@ class FunctionMixin(FloatingType):
 
         def wrapper(self, *args, **kwargs):
             annotate = annotate_tape(kwargs)
-            num_sub_spaces = self.ufl_element().num_sub_elements()
             with stop_annotating():
                 output = split(self, *args, **kwargs)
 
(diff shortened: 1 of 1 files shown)


stephankramer:
Tests here: https://github.com/dolfin-adjoint/pyadjoint/pull/10

PR #1650, opened 2020-04-07
Pefarrell/compile expression dual evaluation

wence-:
Co-depends on firedrakeproject/tsfc#206. One should institute some caching of the interpolation kernels (as is done for form compilation) but that will come in a next step since it needs more thought (and UFL changes too). A workaround for reuse that is fine is to use `Interpolator`s

### DIFF
--- a/firedrake/interpolation.py
+++ b/firedrake/interpolation.py
@@ -179,7 +179,11 @@ def make_interpolator(expr, V, subset, access):
 
 @utils.known_pyop2_safe
 def _interpolator(V, tensor, expr, subset, arguments, access):
-    to_element = create_element(V.ufl_element(), vector_is_mixed=False)
+    try:
+        to_element = create_element(V.ufl_element(), vector_is_mixed=False)
+    except KeyError:
+        # FInAT only elements
+        raise NotImplementedError("Don't know how to create FIAT element for %s" % V.ufl_element())
 
     if access is op2.READ:
         raise ValueError("Can't have READ access for output function")


firedrake/interpolation.py
```diff
@@ -209,11 +198,16 @@ def _interpolator(V, tensor, expr, subset, arguments, access):
     if not isinstance(expr, firedrake.Expression):
         if expr.ufl_domain() and expr.ufl_domain() != V.mesh():
             raise NotImplementedError("Interpolation onto another mesh not supported.")
-        if expr.ufl_shape != V.shape:
-            raise ValueError("UFL expression has incorrect shape for interpolation.")
```
wence-, line 213: This is checked above.

tests/slate/test_darcy_hybridized_mixed.py
```diff
@@ -46,14 +46,10 @@ def test_darcy_flow_hybridization(degree, hdiv_family):
...
-                 'ksp_type': 'gmres',
-                 'pc_fieldsplit_schur_fact_type': 'FULL',
-                 'fieldsplit_0': {'ksp_type': 'cg',
-                                  'pc_factor_shift_type': 'INBLOCKS'},
-                 'fieldsplit_1': {'ksp_type': 'cg',
-                                  'pc_factor_shift_type': 'INBLOCKS'}}
+    sc_params = {'mat_type': 'aij',
+                 'ksp_type': 'preonly',
+                 'pc_type': 'lu',
+                 'pc_factor_mat_solver_type': 'mumps'}
```
miklos1, line 52: Looks like an unrelated change, but whatever...

pefarrell: Ah, it was related, albeit tangentially. Now that we can interpolate the boundary data, the solver does so, and the code was giving slightly different answers when solved with an iterative solver with a loose tolerance. Using a direct solver fixed this.

miklos1: I see, thanks for the explanation.

PR #1652, opened 2020-04-09
Fix MeshTopology geometric dimension

ReubenHill:
Adds the cell geometric dimension to the `ufl_cell` created when creating a `MeshTopology` object. Otherwise the `MeshTopology`'s `ufl_cell` field defaults to the same topological dimension as geometric dimension.

This is necessary to ensure correct creation of `CoordinatelessFunction`s on the mesh topology: UFL uses the UFL cell's geometric dimension information when creating a `CoordinatelessFunction` (rather than looking at the `dim` parameter added when creating the `FunctionSpace` for the `CoordinatelessFunction`). See `ufl.Coefficient.__init__(self, function_space.ufl_element())` on line 62 of `function.py`.

### DIFF
--- a/firedrake/mesh.py
+++ b/firedrake/mesh.py
@@ -1193,7 +1193,7 @@ values from f.)"""
 
         if self.ufl_cell() not in (ufl.Cell('triangle', 3),
                                    ufl.Cell("quadrilateral", 3),
-                                   ufl.TensorProductCell(ufl.Cell('interval'), ufl.Cell('interval'),
+                                   ufl.TensorProductCell(ufl.Cell('interval', geometric_dimension=2), ufl.Cell('interval'),
                                                          geometric_dimension=3)):
             raise NotImplementedError('Only implemented for triangles and quadrilaterals embedded in 3d')
 
--- a/firedrake/mesh.py
+++ b/firedrake/mesh.py
@@ -424,7 +424,9 @@ class MeshTopology(object):
             partitioner.setFromOptions()
             plex.distribute(overlap=0)
 
-        dim = plex.getDimension()
+        tdim = plex.getDimension()
+        # assume the embedding space is the same as the geometric space
+        gdim = plex.getCoordinateDim()
 
         # Allow empty local meshes on a process
         cStart, cEnd = plex.getHeightStratum(0)  # cells
@@ -437,7 +439,7 @@ class MeshTopology(object):
         nfacets = self.comm.allreduce(nfacets, op=MPI.MAX)
 
         self._grown_halos = False
-        self._ufl_cell = ufl.Cell(_cells[dim][nfacets])
+        self._ufl_cell = ufl.Cell(_cells[tdim][nfacets], geometric_dimension=gdim)
 
         # A set of weakrefs to meshes that are explicitly labelled as being
         # parallel-compatible for interpolation/projection/supermeshing
@@ -470,7 +472,7 @@ class MeshTopology(object):
                                                                  reordering)
 
                 # Derive a cell numbering from the Plex renumbering
-                entity_dofs = np.zeros(dim+1, dtype=IntType)
+                entity_dofs = np.zeros(tdim+1, dtype=IntType)
                 entity_dofs[-1] = 1
 
                 self._cell_numbering = self.create_section(entity_dofs)
--- a/firedrake/output.py
+++ b/firedrake/output.py
@@ -38,20 +38,20 @@ ufl_wedge = ufl.TensorProductCell(ufl.Cell("triangle"),
 ufl_hex = ufl.TensorProductCell(ufl.Cell("quadrilateral"),
                                 ufl.Cell("interval"))
 cells = {
-    (ufl.Cell("interval"), False): VTK_INTERVAL,
-    (ufl.Cell("interval"), True): VTK_LAGRANGE_CURVE,
-    (ufl.Cell("triangle"), False): VTK_TRIANGLE,
-    (ufl.Cell("triangle"), True): VTK_LAGRANGE_TRIANGLE,
-    (ufl.Cell("quadrilateral"), False): VTK_QUADRILATERAL,
-    (ufl.Cell("quadrilateral"), True): VTK_LAGRANGE_QUADRILATERAL,
-    (ufl_quad, True): VTK_LAGRANGE_QUADRILATERAL,
-    (ufl_quad, False): VTK_QUADRILATERAL,
-    (ufl.Cell("tetrahedron"), False): VTK_TETRAHEDRON,
-    (ufl.Cell("tetrahedron"), True): VTK_LAGRANGE_TETRAHEDRON,
-    (ufl_wedge, False): VTK_WEDGE,
-    (ufl_wedge, True): VTK_LAGRANGE_WEDGE,
-    (ufl_hex, False): VTK_HEXAHEDRON,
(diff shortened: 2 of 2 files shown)


ReubenHill:
Where should the fix be? Inside `CoordinatelessFunction`? 

In UFL I came across this whilst trying to create a mesh with topologically zero dimensional elements which caused UFL to fail on line `245` of `ufl/domain.py` (function `affine_mesh`). Note that it gets the geometric dimension of the cell on line `243` of `ufl/domain.py`.

wence-:
I'm not sure. But a mesh topology that advertises a cell with a different embedding dimension than topological dimension smells funny.

ksagiyam:
I think there are actually three mesh abstraction levels:
1. topological one (with no concept of geometric dimension)
2. topological one + gdim
3. topological one + coordinates
Type 3 is `MeshGeometry`, which is least flexible. We currently call type 1 `MeshTopology,` which is already enough for `CoordinatelessFunction`. Type 2 can be regarded as type 3 with flexible coordinates; it basically does not distinguish some spherical surface meshes from some ellipsoidal surface meshes.
Type 1 mesh could tell us the minimum geometric dimension it can live in (e.g. by looking at a spherical surface mesh, we can tell that it has to live in three- or higher-dimensional space), but not the exact gdim, so I would like to regard type 2 as different from type1.

I think we need to discuss which abstraction levels (type1, type2 or both) Firedrake wants to have.

miklos1:
Like @wence-, I don't like this idea either.  When you create a `VectorFunctionSpace`, you can specify any dimension you like, although it defaults to the geometric dimension of the underlying mesh, because that's what most people need most of the time.  Rather than leaking geometric information into a `MeshTopology`, the idea was that in case of `gdim != tdim`, you do
```python
coordinate_fs = VectorFunctionSpace(mesh_topology, "CG", 1, dim=gdim)
coordinates = CoordinatelessFunction(coordinate_fs)
```
or something like this. (As in, I'm not sure if the above code is valid Firedrake, but you get the idea...)

ReubenHill:
If `MeshTopology` is actually @ksagiyam 's "type 1" then perhaps the cell geometric dimension should really be `None`? 
I agree that we need to sort out what the abstractions are.

miklos1:
Yes, MeshTopology is type 1 according to @ksagiyam, VectorFunctionSpace on a MeshTopology is more than type 2 but less than type 3, and MeshGeometry is type 3.

I don't mind setting `geometric_dimension`, though I doubt UFL support that, and you may also break assumptions elsewhere in Firedrake.

dham:
I think this probably actually is the right approach. `MeshTopology` is not really type 1, because it contains a Plex, and that Plex knows about geometric dimension. This means that `MeshTopology` is really type 2. It also builds a UFL cell, and those know about geometric dimension too. I think that knowing the correct geometric dimension and not using it just because the class name has "Topology" in it is a perverse thing to do. Remember that we actually make this distinction not because the Topology doesn't have access to information about the geometry, but rather because the class hierarchy would otherwise be circular.

wence-:
> In UFL I came across this whilst trying to create a mesh with topologically zero dimensional elements which caused UFL to fail on line 245 of ufl/domain.py (function affine_mesh). Note that it gets the geometric dimension of the cell on line 243 of ufl/domain.py.

Is the right fix here not to initialise the `ufl.Coefficient` in `CoordinatelessFunction` with an appropriate UFL function space (rather than an element)? I think the PointMeshTopology has a cell with tdim==gdim==0 and a `VectorFunctionSpace(mesh, "DG", 0, dim=0)` "domain"

ReubenHill:
I've pushed some of changes that need to be made to get this passing tests. As I've mentioned on slack, I don't like what this has done to extruded meshes. Their `ufl_cell` can now have two geometric dimensions `TensorProductCell(Cell('interval', 2), interval, geometric_dimension=3)` which seems confusing (though may be a useful descriptor).

ReubenHill:
@wence- That only makes sense if `MeshTopology` is indeed type 1 which @dham and I are not sure it is. For it to be type 1 we would need to remove the plex and be able to tell UFL that the cell's geometric dimension is `None`.

wence-:
Well, the plex is a topological thing on which has been hung geometric information (a plex has a coordinateDM which is like our meshtopology/meshgeometry separation). I don't think we need to remove the plex from a meshtopology. But I think we need to think about what we're going to tell UFL at various points along the way. UFL doesn't know anything concrete about topologies, so we're fine ignoring any plex information. What do we need to inform UFL about. I think, abstractly topological information, which means `tdim == gdim` everywhere in the TopologicalDomain. Then one augments that with some geometric information.

If, on the other hand, UFL doesn't know about topology (and "fakes" it) maybe we should think about adding that abstraction level.

wence-:
A side note, previously we thought 1D was always the hard thing to get right, now it appears you don't even need that many Ds.

firedrake/mesh.py
```diff
@@ -1191,7 +1193,7 @@ def init_cell_orientations(self, expr):
 
         if self.ufl_cell() not in (ufl.Cell('triangle', 3),
                                    ufl.Cell("quadrilateral", 3),
-                                   ufl.TensorProductCell(ufl.Cell('interval'), ufl.Cell('interval'),
+                                   ufl.TensorProductCell(ufl.Cell('interval', geometric_dimension=2), ufl.Cell('interval'),
                                                          geometric_dimension=3)):
```
miklos1, line 1197: Yes, this is precisely the thing that, I think, we shall not want to have.

ReubenHill:
- The UFL `AbstractCell` and `AbstractDomain` classes each have `geometric_dimension` and `topological_dimension` properties. 
 - When creating a UFL `Cell(AbstractCell)` the `geometric_dimension == topological_dimension` if `geometric_dimension = None` which it does by default. As far as I can tell this is the only time this occurs.
 - On `AbstractDomain`s, such as `Mesh(AbstractDomain)`, the geometric dimension is taken as given by the `AbstractCell`.

So @wence-  I think that your assumption that, if not specified, `geometric_dimension == topological_dimension` is correct from a UFL point of view. There is, however, no way of making it clear that this is "incomplete" topological-only information, bar it being held in a `MeshTopology`. In which case your solution to specify the UFL domain carefully when creating the CoordinatelessFunction seems reasonable. Ideally it would be possible to guard against this methinks to make sure we aren't making the same mistake elsewhere

wence-:
Proposal: the proximate error is because UFL tries to turn a finite element into a UFL domain (mesh) in `Coefficient.__init__` which we call in `CoordinatelessFunction.__init__`. Instead, we could just build the "right" mesh by hand, something like:

```python
element = ufl.VectorElement("Lagrange", cell, 1, dim=cell.topological_dimension())
mesh = ufl.Mesh(element)
V = ufl.FunctionSpace(mesh, function_space.ufl_element())
ufl.Coefficient.__init__(V)
```

wence-:
Proposal: the proximate error is because UFL tries to turn a finite element into a UFL domain (mesh) in `Coefficient.__init__` which we call in `CoordinatelessFunction.__init__`. Instead, we could just build the "right" mesh by hand, something like:

```python
element = ufl.VectorElement("Lagrange", cell, 1, dim=cell.topological_dimension())
mesh = ufl.Mesh(element)
V = ufl.FunctionSpace(mesh, function_space.ufl_element())
ufl.Coefficient.__init__(V)
```

ReubenHill:
I have created a new PR #1667 with the suggested changes. Closing this.

PR #1653, opened 2020-04-09
add mesh of the unit disk to utility meshes

danshapero:
This PR adds a mesh of the unit disk by mapping from the rectangle [-1, 1] x [-1, 1]. The minimum and maximum angles are around pi / 6.75 and pi / 2.5 respectively, and the ratio of the largest triangle area to the smallest approaches 2.

I haven't added the option for higher-degree coordinate spaces like for the icosahedral sphere mesh, but I can do that if need be.

### DIFF
--- a/firedrake/utility_meshes.py
+++ b/firedrake/utility_meshes.py
@@ -675,8 +675,6 @@ def UnitDiskMesh(refinement_level=0, distribution_parameters=None, comm=COMM_WOR
     # mark boundary facets
     plex.createLabel(dmplex.FACE_SETS_LABEL)
     plex.markBoundaryFaces("boundary_faces")
-    coords = plex.getCoordinates()
-    coord_sec = plex.getCoordinateSection()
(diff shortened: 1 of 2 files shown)


firedrake/utility_meshes.py
```diff
@@ -644,6 +644,57 @@ def CircleManifoldMesh(ncells, radius=1, distribution_parameters=None, comm=COMM
     return m
 
 
+def UnitDiskMesh(refinement_level=0, distribution_parameters=None, comm=COMM_WORLD):
+    """Generate a mesh of the unit disk in 2D
+
+    :kwarg refinement_level: optional number of refinements (0 is a diamond)
+    :kwarg reorder: (optional), should the mesh be reordered?
```
pefarrell, line 651: I don't think the flag takes a `reorder` flag?

danshapero: Woops, I was using the same docstring from other functions in the same module. Just added a reorder flag.

PR #1657, opened 2020-04-14
include arxiv urls in bibtex

dham:
This change introduces a mechanism for ensuring that all of the publications on the publications page of the website have a URL or DOI associated with them. The mechanism is a new script `firedrake-preprocess-bibtex`. This adds urls to the bibtex file for arxiv papers, and throws an error for any other bibtex entry which has neither a URL or a DOI. The script also has a verification mode which is used to ensure a test fail if a bibtex entry without a URL or DOI is inserted. As a side effect, because the bibtex files are parsed and rewritten, we get some uniformity of formatting among entries.

### DIFF
--- a/scripts/firedrake-preprocess-bibtex
+++ b/scripts/firedrake-preprocess-bibtex
@@ -16,7 +16,6 @@ parser.add_argument("bibtex-file", help="The BibTeX file to process")
 parser.add_argument("--validate", action='store_true',
                     help="Instead of rewriting the bibtex file, raise an exception if anything would have changed.")
 
-
 args = parser.parse_args()
 
 filename = sys.argv[1]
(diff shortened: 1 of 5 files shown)


dham:
The reformatting causes a one-off large diff on the bibtex files. On the bright side, enforced formatting means that future diffs will be small.

scripts/firedrake-preprocess-bibtex
```diff
@@ -0,0 +1,55 @@
...
+except ImportError:
+    raise ImportError("Failed to import bibtexparser. Run:\n firedrake-update --documentation-dependencies")
+
+
+parser = ArgumentParser(description="""Ensure BibTeX entries for inclusion in the Firedrake website have a
+URL or DOI, and impose clean formatting.""",
+                        formatter_class=RawDescriptionHelpFormatter)
+parser.add_argument("bibtex-file", help="The BibTeX file to process")
+parser.add_argument("--validate", action='store_true',
+                    help="Instead of rewriting the bibtex file, raise an exception if anything would have changed.")
```
wence-, line 17: It appears (check Jenkins) that option flags and positional arguments don't play well together. e.g. you can't write (with this setup) `firedrake-preprocess-bibtex --validate foo.bib`

scripts/firedrake-preprocess-bibtex
```diff
@@ -0,0 +1,55 @@
...
+parser = bibtexparser.bparser.BibTexParser()
+parser.common_strings = True
+parser.ignore_nonstandard_types = False
+
+with open(filename) as bibtex_file: 
+    bib_database = parser.parse_file(bibtex_file)
+
+for entry in bib_database.entries:
+    if not "url" in entry and \
+       not "doi" in entry:
```
wence-, line 33: Lint fail. Prefer `if "url" not in entry and "doi" not in entry:`

PR #1662, opened 2020-04-23
Support mat_type dense

pefarrell:
Depends on https://github.com/OP2/PyOP2/pull/578 .

### DIFF
--- a/firedrake/matrix.py
+++ b/firedrake/matrix.py
@@ -31,7 +31,7 @@ class MatrixBase(object, metaclass=abc.ABCMeta):
         self.mat_type = mat_type
         """Matrix type.
 
-        Matrix type used in the assembly of the PETSc matrix: 'aij', 'baij', or 'nest',
+        Matrix type used in the assembly of the PETSc matrix: 'aij', 'baij', 'dense' or 'nest',
         or 'matfree' for matrix-free."""
 
     @property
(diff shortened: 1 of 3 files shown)


firedrake/matrix.py
```diff
@@ -96,6 +96,7 @@ def __init__(self, a, bcs, mat_type, *args, **kwargs):
         # sets self._a, self._bcs, and self._mat_type
         super(Matrix, self).__init__(a, bcs, mat_type)
         options_prefix = kwargs.pop("options_prefix")
+        kwargs["mat_type"] = mat_type
         self.M = op2.Mat(*args, **kwargs)
```
wence-, line 100: You can just say `op2.Mat(*args, mat_type=mat_type, **kwargs)`

pefarrell: Great, thanks!

wence-:
Tests? Projections would be fine, but preferably one scalar, one vector, one mixed.

pefarrell:
I decided to add a mixed case to the test to be sure to be sure.

tests/regression/test_mat_type_dense.py
```diff
@@ -0,0 +1,38 @@
...
+    sp = {"snes_type": "ksponly",
+          "ksp_type": "preonly",
+          "pc_type": "lu",
+          "mat_type": mat_type}
+
+    solve(F == 0, u, bcs, solver_parameters=sp)
+    return u
+
+
+@pytest.mark.parametrize(('kind',), [('vector',), ('mixed',)])
```
wence-, line 33: FWIW for future reference, I think you can write:
```
@pytest.mark.parametrize("kind", ["vector", "mixed"])
```

instead here.

pefarrell: Ah, I was cargo-culting from other tests. Thanks!

wence-:
Thanks.

PR #1663, opened 2020-04-23
Implement p-multigrid

pefarrell:
This commit implements `firedrake.PMGPC`, a preconditioner for p-multigrid.

Internally, this sets up a DM with a custom coarsen routine that p-coarsens the problem. This DM is passed to an internal PETSc PC of type MG and with options prefix 'pmg_'. The relaxation to apply on every p-level is described by 'pmg_mg_levels_', and the coarse solve by 'pmg_mg_coarse_'. Geometric multigrid or any other solver in firedrake may be applied to the coarse problem; an example chaining p-MG, GMG and AMG is given in the tests.

The p-coarsening is implemented in the `coarsen_element` routine.  This takes in a `ufl.FiniteElement` and either returns a new, coarser element, or raises a `ValueError` (if the supplied element should be the coarsest one of the hierarchy).

The default coarsen_element is to perform power-of-2 reduction of the polynomial degree. For mixed systems a `NotImplementedError` is raised, as I don't know how to make a sensible default for this.  It is expected that many (most?) applications of this preconditioner will subclass :class:`PMGPC` to override `coarsen_element`.


firedrake/preconditioners/pmg.py
```diff
@@ -0,0 +1,300 @@
...
+        shape = ele.value_shape()
+
+        if len(shape) == 0:
+            new_ele = FiniteElement(family, ele.cell(), degree // 2)
+        elif len(shape) == 1:
+            new_ele = VectorElement(family, ele.cell(), degree // 2, dim=shape[0])
+        else:
+            new_ele = TensorElement(family, ele.cell(), degree // 2, shape=shape,
+                                    symmetry=ele.symmetry())
+        return new_ele
```
wence-, line 78: AKA `return ele.reconstruct(degree=degree // 2)`

firedrake/preconditioners/pmg.py
```diff
@@ -0,0 +1,300 @@
...
+        section = odm.getDefaultSection()
+        attach_hooks(pdm, level=len(elements)-1, sf=sf, section=section)
+        # Now overwrite some routines on the DM
+        pdm.setRefine(None)
+        pdm.setCoarsen(self.coarsen)
+        pdm.setCreateInterpolation(self.create_interpolation)
+        set_function_space(pdm, get_function_space(odm))
+
+        parent = get_parent(odm)
+        add_hook(parent, setup=partial(push_parent, pdm, parent), teardown=partial(pop_parent, pdm, parent),
```
wence-, line 115: Should we `assert parent is not None`?

wence-: Don't recall if that is necessary.

firedrake/preconditioners/pmg.py
```diff
@@ -0,0 +1,300 @@
...
+        cV = firedrake.FunctionSpace(fV.mesh(), cele)
+        cdm = cV.dm
+        cu = firedrake.Function(cV)
+
+        parent = get_parent(fdm)
+        add_hook(parent, setup=partial(push_parent, cdm, parent), teardown=partial(pop_parent, cdm, parent),
+                 call_setup=True)
+
+        mapper = ArgumentReplacer({test: firedrake.TestFunction(cV),
+                                   trial: firedrake.TrialFunction(cV)})
```
wence-, line 168: I think UFL's replace can now handle this case, can you check (and then purge `ArgumentReplacer` if so)?

firedrake/preconditioners/pmg.py
```diff
@@ -0,0 +1,300 @@
...
+from firedrake.preconditioners.base import PCBase
+from firedrake.preconditioners.low_order import ArgumentReplacer
+from firedrake.dmhooks import attach_hooks, get_appctx, push_appctx, pop_appctx
+from firedrake.dmhooks import add_hook, get_parent, push_parent, pop_parent
+from firedrake.dmhooks import get_function_space, set_function_space
+from firedrake.solving_utils import _SNESContext
+import firedrake
+
+
+class PMGPC(PCBase):
```
wence-, line 19: Does this subsume the functionality offered by `P1PC`? If yes, can you replace that with the obvious from here?

pefarrell: Yes, it does, at the cost of an API change. I have done this. The default PETSc settings are slightly different because e.g. PCMG does pre- and post-smoothing by default. The iteration counts are now beautifully regular (constant in p for a given mesh).

firedrake/preconditioners/pmg.py
pefarrell: Purged.

firedrake/preconditioners/pmg.py
pefarrell: I added the assertion, I don't think it does any harm.

firedrake/preconditioners/pmg.py
pefarrell: Much simpler, thanks.

rckirby:
I am using ufl.replace on pieces of test/trial functions from a mixed space and it’s fine.

firedrake/preconditioners/pmg.py
```diff
@@ -0,0 +1,290 @@
...
+        # We would like to use GMG for the coarse solve, at least
+        # sometimes. But PETSc will use this p-DM's getRefineLevels()
+        # instead of the getRefineLevels() of the MeshHierarchy to
+        # decide how many levels it should use for PCMG applied to
+        # the p-MG's coarse problem. So we need to set an option
+        # for the user, if they haven't already; I don't know any
+        # other way to get PETSc to know this at the right time.
+        opts = PETSc.Options(pc.getOptionsPrefix() + "pmg_")
+        if "mg_coarse_pc_mg_levels" not in opts:
+            opts["mg_coarse_pc_mg_levels"] = odm.getRefineLevel() + 1
```
wence-, line 127: You could `ppc.setFromOptions` and then `ppc.setMGNumLevels()` before `ppc.setUp()` I think (does that work?)

firedrake/preconditioners/pmg.py
```diff
@@ -0,0 +1,290 @@
...
+        fV = test.function_space()
+        fu = fctx._problem.u
+
+        cele = self.coarsen_element(fV.ufl_element())
+        cV = firedrake.FunctionSpace(fV.mesh(), cele)
+        cdm = cV.dm
+        cu = firedrake.Function(cV)
+
+        parent = get_parent(fdm)
+        add_hook(parent, setup=partial(push_parent, cdm, parent), teardown=partial(pop_parent, cdm, parent),
```
wence-, line 154: assert parent is not None here too?

firedrake/preconditioners/pmg.py
```diff
@@ -0,0 +1,290 @@
...
+    else:
+        rlgmap, clgmap = mat.local_to_global_maps
+        rlgmap = Pk.local_to_global_map(Pk_bcs, lgmap=rlgmap)
+        clgmap = P1.local_to_global_map(P1_bcs, lgmap=clgmap)
+        unroll = any(bc.function_space().component is not None
+                     for bc in chain(Pk_bcs, P1_bcs) if bc is not None)
+        matarg = mat(op2.WRITE, (Pk.cell_node_map(), P1.cell_node_map()),
+                     lgmaps=(rlgmap, clgmap), unroll_map=unroll)
+        op2.par_loop(prolongation_transfer_kernel(Pk, P1), mesh.cell_set,
+                     matarg)
```
wence-, line 287: This suggests we should provide a better API for doing this kind of thing. But not here.

firedrake/preconditioners/pmg.py
pefarrell: It's not the number of MG levels on `ppc` that I want to control; it's the number of multigrid levels on its inner coarse solver's PC. That isn't constructed until `setUp`, I think. I tried calling `setUp`, getting it out, calling `setMGLevels()`, and calling `setUp` again, but couldn't get it to work.

firedrake/preconditioners/pmg.py
wence-: It feels like `make_interpolator` from `interpolation.py` almost does this.

firedrake/preconditioners/pmg.py
wence-: OK, I see.

firedrake/preconditioners/pmg.py
pefarrell: Done

firedrake/preconditioners/pmg.py
wence-: Probably needs too much refactoring.

wence-:
Thanks!

PR #1666, opened 2020-04-25
Tsslac version1.1 optimisation

narciuch:
Remove ComponentTensors in GEM DAG which results in fewer unnecessary temporaries being created in Slate. 

Note that as of the recent merge with tsslac_version1.1, an AssertionError has appeared when running test_blocks() and test_layers(). The error is 
```
in replace_indices_flexiblyindexed
    assert isinstance(child, Terminal)
```


miklos1:
> an AssertionError has appeared

So what type is that non-terminal that is being flexibly indexed?

sv2518:
> Note that as of the recent merge with tsslac_version1.1, an AssertionError has appeared when running test_blocks() and test_layers(). The error is
> 
> ```
> in replace_indices_flexiblyindexed
>     assert isinstance(child, Terminal)
> ```
I use FlexiblyIndexed nodes to generate Indexed things with offsets for the translation Slate blocks, so FlexiblyIndexed do not solely act on Terminals anymore. Just remove the assertion.

narciuch:
Removing this outdated assertion works to fix this issue.
@sv2518 will remove this assertion in the translation-of-comptensors branch in tsfc

miklos1:
Just for your information: originally, I only used `FlexiblyIndexed` for indexing input and output buffers with views and shapes.  That's why the assertion is the way it is.  You can, of course, extend this, but you also have to take care of correct code generation for those new cases.

sv2518:
Do you still want me to merge this or can we close this, @narciuch ?

PR #1638, opened 2020-03-31
Vertex Only Mesh

ReubenHill:
Adds disconnected vertex only meshes, upon which a DGP0 function space can be created. 

Linked to https://github.com/firedrakeproject/tsfc/pull/207 and https://github.com/FEniCS/fiat/pull/41 and https://github.com/FEniCS/ufl/pull/30

PR #1667, opened 2020-04-27
Fix CoordinatelessFunction UFL Function Space

ReubenHill:
**EDIT: Up to date description**

Following #1652 , this ensures that the correct UFL mesh and UFL function space are made when making a `CoordinatelessFunction`. To do this
  1. A `ufl.classes.Mesh` representing the mesh topology is created is created when creating a `MeshTopology`.
     - Note that, by convention, the the UFL cells which correspond to a particular `MeshTopology` (accessed via the `.topological` property of a firedrake mesh) have geometric dimension equal to their topological dimension. This remains true should a mesh be a manifold immersed in a higher dimensional space.
  2. A `ufl.classes.FunctionSpace` representing a function space on a mesh topology is created when creating a `FunctionSpace`.
  3. The `ufl.classes.FunctionSpace` created for each `FunctionSpace` is then used _explicitly_ when calling `ufl.Coefficient.__init__` in `CoordinatelessFunction`.

Prior to this PR in `CoordinatelessFunction`, the line `ufl.Coefficient.__init__(self, function_space.ufl_element())` would create a UFL mesh _implicitly_ using the `function_space.ufl_element()` which, as stated above, has topological dimension equal to the geometric dimension by convention even for immersed manifold meshes. This led to confusion that is now avoided by explicitly creating a `ufl.classes.Mesh` when creating a  `MeshTopology` and correcponding `ufl.classes.FunctionSpace` when creating a `FunctionSpace`.

Todo:
 - [x] Note in a comment the convention for gdim == tdim in `MeshTopology`

**EDIT: Out of date original description (from which the below discussion flowed):**

Following #1652 , this ensures that the correct UFL mesh and UFL function space are made when making a `CoordinatelessFunction`. This is of particular importance during mesh generation. 

See for example  lines `1369` to `1383` in `firedrake/mesh.py`:

```python
    def callback(self):
        """Finish initialisation."""
        del self._callback
        # Finish the initialisation of mesh topology
        self.topology.init()

        coordinates_fs = functionspace.VectorFunctionSpace(self.topology, "Lagrange", 1,
                                                           dim=geometric_dim)

        coordinates_data = dmplex.reordered_coords(plex, coordinates_fs.dm.getDefaultSection(),
                                                   (self.num_vertices(), geometric_dim))

        coordinates = function.CoordinatelessFunction(coordinates_fs,
                                                      val=coordinates_data,
                                                      name="Coordinates")
```
When creating the `function_space` geometric dimension information is given (`dim=geometric_dim`). Now when the `CoordinatelessFunction` is created, the internal UFL mesh and UFL function space have the correct geometric dimension.

Prior to this in `CoordinatelessFunction`, the line `ufl.Coefficient.__init__(self, function_space.ufl_element())` would create a UFL mesh with geometric dimension equal to the firedrake mesh's topological dimension which was incorrect for immersed manifold meshes. For more, see the discussion in #1652 .

### DIFF
--- a/firedrake/mesh.py
+++ b/firedrake/mesh.py
@@ -424,7 +424,7 @@ class MeshTopology(object):
             partitioner.setFromOptions()
             plex.distribute(overlap=0)
 
-        dim = plex.getDimension()
+        tdim = plex.getDimension()
 
         # Allow empty local meshes on a process
         cStart, cEnd = plex.getHeightStratum(0)  # cells
@@ -437,8 +437,8 @@ class MeshTopology(object):
         nfacets = self.comm.allreduce(nfacets, op=MPI.MAX)
 
         self._grown_halos = False
-        cell = ufl.Cell(_cells[dim][nfacets])
-        self._ufl_mesh = ufl.Mesh(ufl.VectorElement("Lagrange", cell, 1, dim=dim))
+        cell = ufl.Cell(_cells[tdim][nfacets])
+        self._ufl_mesh = ufl.Mesh(ufl.VectorElement("Lagrange", cell, 1, dim=cell.topological_dimension()))
         # A set of weakrefs to meshes that are explicitly labelled as being
         # parallel-compatible for interpolation/projection/supermeshing
         # To set, do e.g.
@@ -470,7 +470,7 @@ class MeshTopology(object):
                                                                  reordering)
 
                 # Derive a cell numbering from the Plex renumbering
-                entity_dofs = np.zeros(dim+1, dtype=IntType)
+                entity_dofs = np.zeros(tdim+1, dtype=IntType)
                 entity_dofs[-1] = 1
 
                 self._cell_numbering = self.create_section(entity_dofs)
@@ -527,14 +527,14 @@ class MeshTopology(object):
         Each row contains ordered cell entities for a cell, one row per cell.
         """
         plex = self._plex
-        dim = plex.getDimension()
+        tdim = plex.getDimension()
 
         # Cell numbering and global vertex numbering
         cell_numbering = self._cell_numbering
         vertex_numbering = self._vertex_numbering.createGlobalSection(plex.getPointSF())
 
         cell = self.ufl_cell()
-        assert dim == cell.topological_dimension()
+        assert tdim == cell.topological_dimension()
         if cell.is_simplex():
             import FIAT
             topology = FIAT.ufc_cell(cell).get_topology()
--- a/firedrake/mesh.py
+++ b/firedrake/mesh.py
@@ -831,7 +831,8 @@ class ExtrudedMeshTopology(MeshTopology):
         self._cell_numbering = mesh._cell_numbering
         self._entity_classes = mesh._entity_classes
         self._subsets = {}
-        self._ufl_cell = ufl.TensorProductCell(mesh.ufl_cell(), ufl.interval)
+        cell = ufl.TensorProductCell(mesh.ufl_cell(), ufl.interval)
+        self._ufl_mesh = ufl.Mesh(ufl.VectorElement("Lagrange", cell, 1, dim=cell.topological_dimension()))
         if layers.shape:
             self.variable_layers = True
             extents = extnum.layer_extents(self._plex,
--- a/firedrake/mesh.py
+++ b/firedrake/mesh.py
@@ -437,8 +437,8 @@ class MeshTopology(object):
         nfacets = self.comm.allreduce(nfacets, op=MPI.MAX)
 
         self._grown_halos = False
-        self._ufl_cell = ufl.Cell(_cells[dim][nfacets])
-
+        cell = ufl.Cell(_cells[dim][nfacets])
+        self._ufl_mesh = ufl.Mesh(ufl.VectorElement("Lagrange", cell, 1, dim=dim))
         # A set of weakrefs to meshes that are explicitly labelled as being
         # parallel-compatible for interpolation/projection/supermeshing
         # To set, do e.g.
@@ -514,7 +514,11 @@ class MeshTopology(object):
 
     def ufl_cell(self):
         """The UFL :class:`~ufl.classes.Cell` associated with the mesh."""
-        return self._ufl_cell
+        return self._ufl_mesh.ufl_cell()
+
+    def ufl_mesh(self):
+        """The UFL :class:`~ufl.classes.Mesh` associated with the mesh."""
+        return self._ufl_mesh
 
     @utils.cached_property
     def cell_closure(self):
--- a/firedrake/mesh.py
+++ b/firedrake/mesh.py
@@ -437,6 +437,13 @@ class MeshTopology(object):
         nfacets = self.comm.allreduce(nfacets, op=MPI.MAX)
 
         self._grown_halos = False
+        # Note that the geometric dimension of the cell is not set here
+        # despite it being a property of a UFL cell. It will default to
+        # equal the topological dimension.
+        # Firedrake mesh topologies, by convention, which specifically
+        # represent a mesh topology (as here) have geometric dimension
+        # equal their topological dimension. This is reflected in the
+        # corresponding UFL mesh.
         cell = ufl.Cell(_cells[tdim][nfacets])
         self._ufl_mesh = ufl.Mesh(ufl.VectorElement("Lagrange", cell, 1, dim=cell.topological_dimension()))
         # A set of weakrefs to meshes that are explicitly labelled as being
@@ -513,11 +520,29 @@ class MeshTopology(object):
         return self
 
     def ufl_cell(self):
-        """The UFL :class:`~ufl.classes.Cell` associated with the mesh."""
+        """The UFL :class:`~ufl.classes.Cell` associated with the mesh.
+
+        .. note::
+
+            By convention, the UFL cells which specifically
+            represent a mesh topology have geometric dimension equal their
+            topological dimension. This is true even for immersed manifold
+            meshes.
+
+        """
         return self._ufl_mesh.ufl_cell()
 
     def ufl_mesh(self):
-        """The UFL :class:`~ufl.classes.Mesh` associated with the mesh."""
+        """The UFL :class:`~ufl.classes.Mesh` associated with the mesh.
+
+        .. note::
+
+            By convention, the UFL cells which specifically
+            represent a mesh topology have geometric dimension equal their
+            topological dimension. This convention will be reflected in this
+            UFL mesh and is true even for immersed manifold meshes.
+
+        """
         return self._ufl_mesh
 
     @utils.cached_property
--- a/firedrake/function.py
+++ b/firedrake/function.py
@@ -59,7 +59,7 @@ class CoordinatelessFunction(ufl.Coefficient):
                                            functionspaceimpl.MixedFunctionSpace)), \
             "Can't make a CoordinatelessFunction defined on a " + str(type(function_space))
 
-        ufl.Coefficient.__init__(self, function_space.ufl_element())
+        ufl.Coefficient.__init__(self, function_space.ufl_function_space())
 
         self.comm = function_space.comm
         self._function_space = function_space
--- a/firedrake/functionspace.py
+++ b/firedrake/functionspace.py
@@ -38,11 +38,11 @@ def make_scalar_element(mesh, family, degree, vfamily, vdegree):
        :meth:`.MeshGeometry.init`) as appropriate.
     """
     mesh.init()
-    if isinstance(family, ufl.FiniteElementBase):
-        return family
-
     topology = mesh.topology
     cell = topology.ufl_cell()
+    if isinstance(family, ufl.FiniteElementBase):
+        return family.reconstruct(cell=cell)
+
 
     if isinstance(cell, ufl.TensorProductCell) \
        and vfamily is not None and vdegree is not None:
--- a/firedrake/functionspace.py
+++ b/firedrake/functionspace.py
@@ -43,7 +43,6 @@ def make_scalar_element(mesh, family, degree, vfamily, vdegree):
     if isinstance(family, ufl.FiniteElementBase):
         return family.reconstruct(cell=cell)
 
-
     if isinstance(cell, ufl.TensorProductCell) \
        and vfamily is not None and vdegree is not None:
         la = ufl.FiniteElement(family,
--- a/firedrake/functionspaceimpl.py
+++ b/firedrake/functionspaceimpl.py
@@ -290,7 +290,7 @@ class FunctionSpace(object):
             self.shape = element.value_shape()[:1]
         else:
             self.shape = ()
-        self._ufl_element = element
+        self._ufl_function_space = ufl.FunctionSpace(mesh.ufl_mesh(), element)
         self._shared_data = sdata
         self._mesh = mesh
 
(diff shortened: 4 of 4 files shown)


wence-:
FWIW, this is not what I suggested here https://github.com/firedrakeproject/firedrake/pull/1652#issuecomment-617263476

I had thought that our conclusion was that we wouldn't have gdim information in the mesh inside a coordinatelessfunction.

ReubenHill:
> FWIW, this is not what I suggested here [#1652 (comment)](https://github.com/firedrakeproject/firedrake/pull/1652#issuecomment-617263476)
> 
> I had thought that our conclusion was that we wouldn't have gdim information in the mesh inside a coordinatelessfunction.

Oh so the proposal is that the coordinates element of the UFL mesh is _explicitly_ stated to be an L1 vector element with dimension equal to the topological dimension of the mesh? This at least stops UFL asking for the incorrect geometric dimension information but it still fails should the topological dimension be zero for the reasons given before.

Whilst this stops us doing something _wrong_ in UFL it would force me into allowing `ufl.VectorElement` to accept `dim=0` for my main job (see #1638 ) and I don't know if that's something that mathematically makes sense (though am very happy to be told it does!)

ReubenHill:
I've pushed a40af6af2277b5bcf378bfb7aac6a29842d40845 which _I think_ does what you want - but see my previous comment for why I am confused.

ReubenHill:
What's wrong with incorporating the geometric dimension information at this point? We already gave it when creating the `coordinates_fs` in Firedrake. Have I just got completely the wrong end of the stick?

ReubenHill:
Proposal:

```python
class CoordinatelessFunction(ufl.Coefficient):
    ...
    def __init__(self, function_space, val=None, name=None, dtype=ScalarType):
        ...

        # build UFL mesh with explicitly created UFL coordinates vector element 
        # (where, notably, dim = geometric_dimension) then build a function 
        # space on this mesh.
        gdim = function_space.ufl_element().num_sub_elements()
        e = ufl.VectorElement("Lagrange", function_space.mesh().ufl_cell(), 1, dim=gdim)
        mesh = ufl.Mesh(e)
        V = ufl.FunctionSpace(mesh, function_space.ufl_element())
        ufl.Coefficient.__init__(self, V)
```

wence-:
Does this gdim pollute anywhere else? I think not, so it's probably OK?

ReubenHill:
I think you are correct but I'm concerned: are there cases where `function_space.ufl_element().num_sub_elements()` could be less than 1 in a `function_space` given to `CoordinatelessFunction`? I'm assuming it's always a `VectorFunctionSpace` I think? For now I'll make the change and see how and if Jenkins complains.

ReubenHill:
Can confirm this introduces other problems, at least in the 0D case.

```python
# vm is some vertex only mesh with topological dimension 0
V = FunctionSpace(vm, "DG", 0)                                                                                                                                                                           
assert V.ufl_element().num_sub_elements() == 0 # <- passes
f = Function(V) # <- fails in Function init when calling self._data = CoordinatelessFunction(V.topological, val=val, name=name, dtype=dtype)
```

wence-:
I think you are getting confused as to the point of a coordinatelessfunction. It is _not_ an object used just to hold the coordinate field. It is used to hold any data that isn't yet attached to a mesh with geometric information. Hence you cannot infer anything about geometric dimension from the function space that it's built on.

A `VectorElement` with `dim=0` I think makes sense since that is saying you have something living in `\mathbb{R}^0` which is what your point cloud is before embedding in `\mathbb{R}^n`.

If you apply this patch in UFL:
```diff
diff --git a/ufl/finiteelement/mixedelement.py b/ufl/finiteelement/mixedelement.py
index f2a00c5a..cc43b923 100644
--- a/ufl/finiteelement/mixedelement.py
+++ b/ufl/finiteelement/mixedelement.py
@@ -56,9 +56,13 @@ class MixedElement(FiniteElementBase):
 
         # Check that all elements use the same quadrature scheme TODO:
         # We can allow the scheme not to be defined.
-        quad_scheme = elements[0].quadrature_scheme()
-        if not all(e.quadrature_scheme() == quad_scheme for e in elements):
+        quad_schemes = set(e.quadrature_scheme for e in elements)
+        if len(quad_schemes) > 1:
             error("Quadrature scheme mismatch for sub elements of mixed element.")
+        elif len(quad_schemes) == 1:
+            quad_scheme, = quad_schemes
+        else:
+            quad_scheme = None
 
         # Compute value sizes in global and reference configurations
         value_size_sum = sum(product(s.value_shape()) for s in self._sub_elements)
@@ -292,8 +296,8 @@ class VectorElement(MixedElement):
         # Initialize element data
         MixedElement.__init__(self, sub_elements, value_shape=value_shape,
                               reference_value_shape=reference_value_shape)
-        # FIXME: Storing this here is strange, isn't that handled by
-        # subclass?
+        FiniteElementBase.__init__(self, sub_element.family(), cell, degree, quad_scheme,
+                                   value_shape, reference_value_shape)
         self._family = sub_element.family()
         self._degree = sub_element.degree()
         self._sub_element = sub_element
```

Then you can at least make a `VectorElement` with dim=0.

Then something like this in Firedrake?
```diff
diff --git a/firedrake/function.py b/firedrake/function.py
index 8b83868b..19332dcd 100644
--- a/firedrake/function.py
+++ b/firedrake/function.py
@@ -59,7 +59,7 @@ class CoordinatelessFunction(ufl.Coefficient):
                                            functionspaceimpl.MixedFunctionSpace)), \
             "Can't make a CoordinatelessFunction defined on a " + str(type(function_space))
 
-        ufl.Coefficient.__init__(self, function_space.ufl_element())
+        ufl.Coefficient.__init__(self, function_space.ufl_function_space())
 
         self.comm = function_space.comm
         self._function_space = function_space
diff --git a/firedrake/functionspaceimpl.py b/firedrake/functionspaceimpl.py
index 4e4ded86..9d4405a9 100644
--- a/firedrake/functionspaceimpl.py
+++ b/firedrake/functionspaceimpl.py
@@ -290,7 +290,7 @@ class FunctionSpace(object):
             self.shape = element.value_shape()[:1]
         else:
             self.shape = ()
-        self._ufl_element = element
+        self._ufl_function_space = ufl.FunctionSpace(mesh.ufl_mesh(), element)
         self._shared_data = sdata
         self._mesh = mesh
 
@@ -389,7 +389,10 @@ class FunctionSpace(object):
         return self._mesh
 
     def ufl_element(self):
-        return self._ufl_element
+        return self.ufl_function_space().ufl_element()
+
+    def ufl_function_space(self):
+        return self._ufl_function_space
 
     def __len__(self):
         return 1
@@ -534,11 +537,12 @@ class MixedFunctionSpace(object):
         super(MixedFunctionSpace, self).__init__()
         self._spaces = tuple(IndexedFunctionSpace(i, s, self)
                              for i, s in enumerate(spaces))
-        self._ufl_element = ufl.MixedElement(*[s.ufl_element() for s
-                                               in spaces])
+        mesh, = set(s.mesh() for s in spaces)
+        self._ufl_function_space = ufl.FunctionSpace(mesh.ufl_mesh(),
+                                                     ufl.MixedElement(*[s.ufl_element() for s in spaces]))
         self.name = name or "_".join(str(s.name) for s in spaces)
         self._subspaces = {}
-        self._mesh = spaces[0].mesh()
+        self._mesh = mesh
         self.comm = self.node_set.comm
 
     # These properties are so a mixed space can behave like a normal FunctionSpace.
@@ -557,7 +561,11 @@ class MixedFunctionSpace(object):
 
     def ufl_element(self):
         r"""The :class:`~ufl.classes.Mixedelement` this space represents."""
-        return self._ufl_element
+        return self.ufl_function_space().ufl_element()
+
+    def ufl_function_space(self):
+        return self._ufl_function_space
+
 
     def __eq__(self, other):
         if not isinstance(other, MixedFunctionSpace):
diff --git a/firedrake/mesh.py b/firedrake/mesh.py
index b9ecf46b..f262ce2c 100644
--- a/firedrake/mesh.py
+++ b/firedrake/mesh.py
@@ -437,8 +437,8 @@ class MeshTopology(object):
         nfacets = self.comm.allreduce(nfacets, op=MPI.MAX)
 
         self._grown_halos = False
-        self._ufl_cell = ufl.Cell(_cells[dim][nfacets])
-
+        cell = ufl.Cell(_cells[dim][nfacets])
+        self._ufl_mesh = ufl.Mesh(ufl.VectorElement("Lagrange", cell, 1, dim=dim))
         # A set of weakrefs to meshes that are explicitly labelled as being
         # parallel-compatible for interpolation/projection/supermeshing
         # To set, do e.g.
@@ -514,7 +514,10 @@ class MeshTopology(object):
 
     def ufl_cell(self):
         """The UFL :class:`~ufl.classes.Cell` associated with the mesh."""
-        return self._ufl_cell
+        return self._ufl_mesh.ufl_cell()
+
+    def ufl_mesh(self):
+        return self._ufl_mesh
 
     @utils.cached_property
     def cell_closure(self):
```

ksagiyam:
As this is related, I think it is worth discussing the change I made in the `fs_filter` branch.
The `ufl.Filtered` wants to use `CoordinatelessFunction` to represent dof selection, so I had to introduce something that compares to `CoordinatelessFunction` in `ufl`.
Basically, I introduced what is for now called 
_ufl.TopologicalCoefficient_
_ufl.TopologicalFunctionSpace_
_ufl.TopologicalMesh_
and, consequently, related Firedrake classes subclass these as :
_class CoordinatelessFunction(ufl.TopologicalCoefficient)_
_class FunctionSpace(ufl.TopologicalFunctionSpace)_
_class MeshTopology(ufl.TopologicalMesh)_

As, here, `CoordiantelessFunction` subclasses `ufl.TopologicalCoefficient`, and not `ufl.Coefficient` (which requires `ufl.Functionspace` and `ufl.Mesh`), I think this new inheritance simplifies 
what we are trying to do here.

ReubenHill:
Note the UFL `FunctionSpace` constructor needs a UFL domain as its first argument. Simply using the UFL mesh causes UFL to error in many cases (see test outputs) since the elements don't match. I've also got some mysterious solve failures that I haven't worked out how to debug yet.

ReubenHill:
@wence I can't see an obvious way around this. In `functionspace.py` we have 
```python
def FunctionSpace(mesh, family, degree=None, name=None, vfamily=None,
                  vdegree=None):
    """Create a :class:`.FunctionSpace`.

    :arg mesh: The mesh to determine the cell from.
    :arg family: The finite element family.
    :arg degree: The degree of the finite element.
    :arg name: An optional name for the function space.
    :arg vfamily: The finite element in the vertical dimension
        (extruded meshes only).
    :arg vdegree: The degree of the element in the vertical dimension
        (extruded meshes only).

    The ``family`` argument may be an existing
    :class:`ufl.FiniteElementBase`, in which case all other arguments
    are ignored and the appropriate :class:`.FunctionSpace` is returned.
    """
    element = make_scalar_element(mesh, family, degree, vfamily, vdegree)

    # Support FunctionSpace(mesh, MixedElement)
    if type(element) is ufl.MixedElement:
        return MixedFunctionSpace(element, mesh=mesh, name=name)

    # Support foo x Real tensorproduct elements
    real_tensorproduct = False
    scalar_element = element
    if isinstance(element, (ufl.VectorElement, ufl.TensorElement)):
        scalar_element = element.sub_elements()[0]
    if isinstance(scalar_element, ufl.TensorProductElement):
        a, b = scalar_element.sub_elements()
        real_tensorproduct = b.family() == 'Real'

    # Check that any Vector/Tensor/Mixed modifiers are outermost.
    check_element(element)

    # Otherwise, build the FunctionSpace.
    topology = mesh.topology
    if element.family() == "Real":
        new = impl.RealFunctionSpace(topology, element, name=name)
    else:
        new = impl.FunctionSpace(topology, element, name=name, real_tensorproduct=real_tensorproduct)
    if mesh is not topology:
        return impl.WithGeometry(new, mesh)
    else:
        return new
```

When one calls `new = impl.FunctionSpace(topology, element, name=name, real_tensorproduct=real_tensorproduct)` one tries to create a function space with an element who's ufl cell often contains geometric dimension information on the mesh topology who's cells of course don't. 
`self._ufl_function_space = ufl.FunctionSpace(mesh.ufl_mesh(), element)` in `class FunctionSpace` therefore necessarily fails since the cell don't match.
See for example `tests/output/test_pvd_output.py::test_append[sphere[tri]] ` to see this happening.

Is this doomed to fail or does `element = make_scalar_element(mesh, family, degree, vfamily, vdegree)` produce the wrong element?

ReubenHill:
Rebased and now hopefully working.

ReubenHill:
Ready for review

firedrake/mesh.py
```diff
@@ -437,8 +437,15 @@ def add_overlap():
...
-
+        # Note that the geometric dimension of the cell is not set here
+        # despite it being a property of a UFL cell. It will default to
+        # equal the topological dimension.
+        # Firedrake mesh topologies, by convention, which specifically
+        # represent a mesh topology (as here) have geometric dimension
+        # equal their topological dimension. This is reflected in the
+        # corresponding UFL mesh.
+        cell = ufl.Cell(_cells[tdim][nfacets])
+        self._ufl_mesh = ufl.Mesh(ufl.VectorElement("Lagrange", cell, 1, dim=cell.topological_dimension()))
```
wence-, line 448: Does this break the `point` cell case where you need `VectorElement("DG", cell, 0, ...)` ? Or will you do that in a later round?

firedrake/mesh.py
```diff
@@ -513,8 +520,22 @@ def topological(self):
         return self
 
     def ufl_cell(self):
-        """The UFL :class:`~ufl.classes.Cell` associated with the mesh."""
-        return self._ufl_cell
+        """The UFL :class:`~ufl.classes.Cell` associated with the mesh.
+
+        Note that, by convention, the UFL cells which specifically
```
wence-, line 525: Use
```
.. note:
   
   Indented block
```

For notes.

firedrake/mesh.py
```diff
@@ -827,7 +848,8 @@ def __init__(self, mesh, layers):
         self._cell_numbering = mesh._cell_numbering
         self._entity_classes = mesh._entity_classes
         self._subsets = {}
-        self._ufl_cell = ufl.TensorProductCell(mesh.ufl_cell(), ufl.interval)
+        cell = ufl.TensorProductCell(mesh.ufl_cell(), ufl.interval)
+        self._ufl_mesh = ufl.Mesh(ufl.VectorElement("Lagrange", cell, 1, dim=cell.topological_dimension()))
```
wence-, line 852: I guess an extruded point is an interval, so this will fall out OK.

firedrake/mesh.py
ReubenHill: It seems to work fine with
```python
        cell = ufl.Cell("vertex")
        self._ufl_mesh = ufl.Mesh(ufl.VectorElement("DG", cell, 1, dim=cell.topological_dimension()))
```
in the `VertexOnlyMeshTopology` introduced in #1638 (which needs updating).

firedrake/mesh.py
ReubenHill: I haven't added tests for extrusion with points in #1638  yet - thanks for the heads up. If there are issues I'll revisit in #1638 .

firedrake/mesh.py
ReubenHill: Updated in amended commit 91e48d79b715b47662488044c90a33625d54e63d

firedrake/mesh.py
wence-: What I meant was that you presumably still need to special-case `point` here (but you will do it in #1638, so that is fine).

ReubenHill: Well that's what the `VertexOnlyMeshTopology` does (similarly to and `ExtrudedMeshTopology`. Interestingly 
```python
        cell = ufl.Cell("vertex")
        self._ufl_mesh = ufl.Mesh(ufl.VectorElement("Lagrange", cell, 1, dim=cell.topological_dimension()))
```
works fine and lets you create function spaces but I think this is because UFL doesn't have any checks on what elements are valid. If you try to make that element in FIAT it will fail.

ReubenHill:
ready to merge

PR #1670, opened 2020-04-28
Streamplot

### LINKED ISSUE #1634: visualize vector fields with streamplot
The current plotting interface includes a quiver function like the [matplotlib version](https://matplotlib.org/gallery/images_contours_and_fields/quiver_demo.html#sphx-glr-gallery-images-contours-and-fields-quiver-demo-py). Quiver plots are not always visually informative, especially when the mesh has a very high density.

A better approach would be to implement an analogue of [streamplot](https://matplotlib.org/3.1.1/gallery/images_contours_and_fields/plot_streamplot.html#sphx-glr-gallery-images-contours-and-fields-plot-streamplot-py). This would involve (1) an ODE integrator to generate the streamlines and (2) some kind of sampling strategy to pick the seed points.

There are several research papers on picking the seed points:
* Turk and Banks, 1996. Image-guided streamline placement
* Verma, Kao, Pang, 2000. A flow-guided streamline seeding strategy
* Merbarki, Alliez, Devilliers, 2005. Farthest point seeding for efficient placement of streamlines

I can foresee a possible hiccup in how to handle discontinuous vector fields. A dumb approach is to project into CG. Maybe there's a smarter approach for H(div) or H(curl)-conforming spaces.

danshapero:
Resolves #1634 . This patch adds:

* a `streamline` function to generate integral curves of vector fields
* a helper class `Streamplotter` to accumulate all the calculated streamlines and, most importantly, to calculate whether a point is too close to these streamlines
* a `streamplot` function that picks some seed points and uses the above to generate a streamline plot

The streamline generator uses an adaptive Runge-Kutta 1/2 scheme. The adaptivity needs a length scale, for which we use the `mesh.cell_size` member. This is really convenient but it means this is restricted to triangular and not quad meshes for now.

The streamlines are constrained to have a minimum length and stay a certain distance apart. When one streamline can pass very near itself -- for example spiraling around an attracting fixed point -- things get difficult. If you're too stringent in keeping a streamline far away from itself, you can terminate early because each proposed point is, obviously, close to the previous point. To fix this, I made the streamline integration work on chunks of some length that exceeds the spacing. Updating the grid that tracks distance to the streamlines is lagged by one chunk, which means you can have a maximum of one chunk length of self-proximity. With a reasonable chunk length, the streamlines stay far enough apart from themselves and others while also growing long enough. Hopefully this explains why there's so much code here.


pefarrell:
Lawrence and I discussed replacing the use of `CellSize` in `mesh.cell_size` (which only works on simplices) with `CellDiameter` (which works on more element types). If you make that change, can you apply this to quadrilateral elements also?

danshapero:
Yes, I just tried replacing CellSize with CellDiameter and running the streamplot test on a quad mesh and everything worked just fine.

firedrake/plot.py
```diff
@@ -292,6 +294,321 @@ def quiver(function, **kwargs):
...
+
+    streamplotter = Streamplotter(function, resolution, min_length, max_time,
+                                  tolerance, loc_tolerance)
+
+    # TODO: better way of seeding start points
+    start_points = []
+    shape = streamplotter._grid.shape
+    for i in range(shape[0] - 1):
+        for j in range(shape[1] - 1):
+            start_points.append(streamplotter._grid_point((i, j)))
```
wence-, line 566: This is surely just a `np.meshgrid`?

firedrake/plot.py
```diff
@@ -292,6 +294,321 @@ def quiver(function, **kwargs):
...
+
+    # Randomly shuffle the start points
+    import randomgen
+    generator = randomgen.MT19937(seed).generator
+    start_points = generator.permutation(np.array(start_points))
+
+    for x in start_points:
+        streamplotter.add_streamline(x)
```
wence-, line 574: Given that you add all the points, what is the rationale for permuting them? Is it that streamlines are merged in the order they're seen?

Minor, just do:
```python
for x in generator.permutation(...):
    streamplotter.add_streamline(x)
```

firedrake/plot.py
```diff
@@ -292,6 +294,321 @@ def quiver(function, **kwargs):
...
+
+    # Colors are determined by the speed, thicknesses by arc length
+    speeds = []
+    widths = []
+    for streamline in streamplotter.streamlines:
+        velocity = np.array(function.at(streamline, tolerance=loc_tolerance))
```
wence-, line 580: The fact that function.at doesn't return a numpy array seems to be annoyance. Should we fix that (maybe separately). WDYT?

firedrake/plot.py
```diff
@@ -292,6 +294,321 @@ def quiver(function, **kwargs):
...
+    :arg tolerance: dimensionless tolerance for the RK12 adaptive integration
+    :arg loc_tolerance: tolerance for point location
+    :yield x, v, dt: the position, velocity, and timestep
+    """
+    mesh = function.ufl_domain()
+    cell_sizes = mesh.cell_sizes
+
+    x = np.array(point)
+    v1 = direction * function.at(x, tolerance=loc_tolerance)
+    r = cell_sizes.at(x, tolerance=loc_tolerance)
```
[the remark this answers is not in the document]
danshapero: All that you really need is some length scale to guide the adaptivity, so I don't think it matters one way or the other for the numerics.

At one point I did write this using a `cell_sizes` object by interpolating into DG0, but in order to avoid recreating the cell sizes for every streamline, I had to wrap the generator in a whole other class and the code was more complicated. If you think it's warranted I can go back to doing things that way.

firedrake/plot.py
danshapero: I think returning a numpy array makes more sense, especially given that you can preallocate it easily, but that seems like something for another PR.

firedrake/plot.py
danshapero: The algorithm tries every grid point to initialize a streamline, but many of them will get "covered" by an earlier streamline and thus not used at all. I've found that randomizing the insertion order usually starts the streamlines far apart at first, which gives you longer streamlines. The Right Thing To Do is to use farthest point seeding, but that requires spatial data structures we don't have if you want it to run in less than O(n^2). Randomized approaches happen to work pretty well and could be improved on using [Poisson disk sampling](https://www.jasondavies.com/poisson-disc/).

danshapero:
I think I addressed everything. Once it's good then I'll rebase from master to fix the conflict. I can do another rebase to make the history sensible or squash it all down to one commit.

The approach for seeding points right now is pretty blunt but I think it's at least tolerably good. I have some ideas for how to improve that by finding the fixed points of the vector field first, but I'll save that for later.

firedrake/plot.py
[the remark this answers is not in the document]
wence-: Ah, ok, that's fine.

firedrake/plot.py
wence-: So we have some r-trees lying around, but I don't know what spatial data structures you need.

firedrake/plot.py
wence-: OK.

wence-:
Can you rebase/squash and resolve the conflicts, then I think this is good.

danshapero:
All squashed and I've switched quiver to streamplot in the Stokes demo notebooks.

wence-:
Thanks!

PR #1671, opened 2020-04-29
Show durations of slowest tests

dham:
Add the timings of the slowest 200 tests to the test output so as to guide the culling of tests.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -54,7 +54,7 @@ python -m pip list
                 sh '''
 . ./firedrake/bin/activate
 cd firedrake/src/firedrake
-python -m pytest -n 11 --cov firedrake -v tests
+python -m pytest --durations=200 -n 11 --cov firedrake -v tests
 '''
               }
             }


wence-:
So the top 4 are all serendipity tests. It turns out the tabulation of serendipity elements is _glacial_ since it always goes through sympy.

This is called everywhere because functionspace construction calls entity_support_dofs which does basis evaluation on every subentity.

We can get around the basis evaluation there by just `raise NotImplementedError` in an appropriate place in FInAT.

But really the Serendipity tabulation should be fixed. For example this snippet:
```
from firedrake import *
from tsfc.finatinterface import create_element
from finat.quadrature import make_quadrature
mesh = ExtrudedMesh(mesh, 1)

V = FunctionSpace(mesh, "S", 5)

e = V.finat_element
quad = make_quadrature(e.cell, 8)

basis = e.basis_evaluation(1, quad.point_set, ((2, 1), 0))
```

Takes 200 seconds on my laptop.

dham:
Yuck. OTOH, S5 is an insane element to use anyway. Maybe we should just not test it.

wence-:
This gets it down to under a second:

```patch
diff --git a/FIAT/serendipity.py b/FIAT/serendipity.py
index e5304ad..45dd29b 100644
--- a/FIAT/serendipity.py
+++ b/FIAT/serendipity.py
@@ -6,7 +6,7 @@
 #
 # Modified by David A. Ham (david.ham@imperial.ac.uk), 2019
 
-from sympy import symbols, legendre, Array, diff
+from sympy import symbols, legendre, Array, diff, lambdify
 import numpy as np
 from FIAT.finite_element import FiniteElement
 from FIAT.lagrange import Lagrange
@@ -101,7 +101,7 @@ class Serendipity(FiniteElement):
         super(Serendipity, self).__init__(ref_el=ref_el, dual=None, order=degree, formdegree=formdegree)
 
         self.basis = {(0,)*dim: Array(s_list)}
-
+        self.basis_evals = {(0,)*dim: lambdify(variables, Array(s_list))}
         topology = ref_el.get_topology()
         unflattening_map = compute_unflattening_map(topology)
         unflattened_entity_ids = {}
@@ -153,17 +153,14 @@ class Serendipity(FiniteElement):
             alphas = mis(dim, o)
             for alpha in alphas:
                 try:
-                    polynomials = self.basis[alpha]
+                    polynomials = self.basis_evals[alpha]
                 except KeyError:
                     polynomials = diff(self.basis[(0,)*dim], *zip(variables, alpha))
                     self.basis[alpha] = polynomials
-                T = np.zeros((len(polynomials), len(points)))
-                for i in range(len(points)):
-                    subs = {v: points[i][k] for k, v in enumerate(variables[:dim])}
-                    for j, f in enumerate(polynomials):
-                        T[j, i] = f.evalf(subs=subs)
+                    polynomials = self.basis_evals.setdefault(alpha, lambdify(variables, polynomials))
+                points = np.asarray(points)
+                T = np.asarray(polynomials(*(points[:, i] for i in range(points.shape[1]))))
                 phivals[alpha] = T
-
         return phivals
 
     def entity_dofs(self):
```

rckirby:
It's not completely insane, just wants to be accelerated (and hit with a
good preconditioned!).  It also seems like testing basis functions
correctness and performance might properly be a finat issue.

Also, for firedrake purposes (including assembly, etc), S4 is the lowest
order quad element with internal dofs, so the degree could be lowered from
5 to 4 without much loss of feature coverage.

wence-:
Btw, is there a reason that the serendipity element doesnt implement a dual basis?

rckirby:
I didn't (not) implement that....

In the papers, they are usually constructed "directly" in terms of formula
(part of why they are a bit slow).  At least for the scalar case, I think
there is a Kronecker property with a point set, but it's bonkers to
construct them by inverting a Vandermonde system on a prime basis like the
rest of FIAT.  I guess (and this would help for interpolation and Dirichlet
BC) one could go put in a dual basis, just not use it for construction?

wence-:
If you want interpolation into these spaces, I think we do want a dual.

rckirby:
It's always good to have interpolation and strong boundary conditions if
you can get them.  I guess one needs to figure out the enumeration of
points that would go along with the particular choice of basis functions
given and produce a list of point evaluation functionals somewhere in FIAT?

PR #1674, opened 2020-05-05
Pointwise adjoint operator

nbouziani:
This PR is an update of the deprecated `PointwiseOperator` PR. It is the Firedrake bit of the ExternalOperator concept introduced in UFL.

It handles the practical implementation thereof, especially it implements some of the subclasses covered by this concept :

 - PointexprOperator: Operator defined through an externally defined expression (e.g. lambda expression) passed in. This evaluation of this operator is done by merely evaluate this expression pointwise.

 - PointsolveOperator: Solve an externally defined algebraic relationship pointwise. For instance one can think to an implicit constitutive law that is part of a given PDE-based model.

 - PointnetOperator: Represents neural networks. Evaluation is associated to a forward-pass and gradient computation to a gradient of the output wrt inputs. So far, only PyTorch has been covered (it should be straightforward to extend this to TensorFlow).

There are more classes of function that could be derived from the AbstractPointwiseOperator base class.

This PR also cover the implementation of the adjoint capability of pointwise operators enabling to tackle more generally all the PDE-constrained optimization problems containing pointwise operators in the PDE-constraint and/or in the cost function.

Examples of applications cover:
  - Implicit constitutive relations (e.g. Glen's flow law in glaciology)
  - Regularizer for inverse problems (e.g. seismic inversion)
  - Subgrid parametrization

This PR is associated to several other PRs:

**UFL**: https://github.com/firedrakeproject/ufl/pull/16
**Pyadjoint**: https://github.com/dolfin-adjoint/pyadjoint/pull/15
**TSFC**: https://github.com/firedrakeproject/tsfc/pull/201
**PyOP2**: https://github.com/OP2/PyOP2/pull/579


firedrake/assemble.py
```diff
@@ -241,6 +242,22 @@ def _assemble(f, tensor=None, bcs=None, form_compiler_parameters=None,
...
+                d = new_coefficients[k]
+                # Check if we need to reconstruct new ExtOps
+                if d == c._extop_master and not (len(v) == 1 and v[0] == (0,)*len(v[0])):
+                    # ExternalOperators can generate other Extops (e.g when differentiating)
+                    # The kernel contains information about which Extop has been created during form compiling,
+                    # we still need to construct this dependency when needed, i.e. when the form is already
+                    # compiled and therefore the differentiation bit of the code is not hit.
+                    new_coefficients.remove(c)
+                    d._add_dependencies(v)
+                    new_coefficients.extend([e for e in d._extop_dependencies if e.derivatives in v])
```
wence-, line 259: So it seems like everywhere you have to special-case that this "coefficient" is not a coefficient, but rather an expression that depends on a bunch of coefficients. Why not just say "I am an expression that depends on some coefficients"?

wence-: Or is the problem that inside the kernel there really is just one coefficient, but it must be computed in some funny way?

firedrake/pointwise_operators.py
```diff
@@ -0,0 +1,813 @@
...
+        derivatives = tuple(dj + int(idx == j) for j, dj in enumerate(self.derivatives))
+        dNdq = self._ufl_expr_reconstruct_(*self.ufl_operands, derivatives=derivatives)
+        dNdq = dNdq.evaluate()
+        dNdq_adj = transpose(dNdq)
+        result = firedrake.assemble(dNdq_adj)
+        return result
+
+    @utils.cached_property
+    def _split(self):
+        return tuple(Function(V, val) for (V, val) in zip(self.function_space(), self.topological.split()))
```
wence-, line 97: forgets the operator-ness, is this right?

firedrake/pointwise_operators.py
```diff
@@ -0,0 +1,813 @@
...
+                    # if f(x,y) = 0 =>  df/dx + dy/dx*df/dy = 0
+                    dfds0 = sp.diff(fexpr, symb[0])
+                    dfdsi = sp.diff(fexpr, symb[i])
+
+                    # -> Sympy idiff does not work for high dimension
+                    # -> We could solve the system "dfds0*ds0/dsi = -dfdsi" symbolically and then
+                    # for the n-th derivative just apply n-1th time the derivative on the result.
+                    # However again sympy.linsolve() does not handle high dimensional case
+                    dfds0l = sp.lambdify(symb, dfds0, modules='numpy')
+                    dfdsil = sp.lambdify(symb, dfdsi, modules='numpy')
```
wence-, line 285: Please add `dummify=True` at the very least. Question, should we sanitise input, since this just calls `exec`

nbouziani: Ok thank you. I don't really know, I would say yes if we can.

Because `fexpr` is by assumption supposed to be a SymPy expression passed in the `point_solve` helper function by the user himself, then we just compose it with other SymPy operations (`sp.diff` and `sp.lambdify`). So any problems with the input should come from an incorrect SymPy expression passed in the `point_solve`, which the user can correct (I think).

firedrake/pointwise_operators.py
nbouziani: It's been a long time but I think we don't want `_split` to return subcomponents as being other pointwise operators themselves. `AbstractPointwiseOperator` is directly inheriting from `Function` so I think this is right.

firedrake/assemble.py
nbouziani: Yes exactly, it is just one coefficient which requires additional procedures to deal with it, it is a coefficient enriched with additional methods.

For example, it is a coefficient that can generate other coefficients: `grad(p(u)))` where `p` is a pointwise operator will produce when hitting the form compiler: `p'(u)*grad(u)`, the form now contains `p'` and not `p`.

nbouziani:
All the tests seem to be passing except the 2 tests involving neural nets:

- Firedrake: `tests/pointwiseoperator/test_neural_net.py`
- Pyadjoint : `tests/firedrake_adjoint/test_pointwise_operators.py`

The reason is the  `import torch` (for PyTorch), and there is no instructions anywhere saying that we should install this library.

How am I supposed to deal with this? Should I add an option in the install script to enable to specify if we want to have `torch` installed on the fly?

tests/pointwiseoperator/test_point_solve.py
```diff
@@ -0,0 +1,291 @@
...
+    p2 = p(b, a, a)
+    dp2db = p2._ufl_expr_reconstruct_(b, a, a, derivatives=(1, 0, 0))
+    a3 = dp2db*dx
+
+    a4 = 2*b*dx  # dp2/db
+
+    assemble_a3 = assemble(a3)
+    assemble_a4 = assemble(a4)
+
+    assert abs(assemble_a3 - assemble_a4) < 1.0e-7
```
sv2518, line 114: Maybe use math.isclose?

firedrake/adjoint/assembly.py
```diff
@@ -19,10 +20,19 @@ def wrapper(*args, **kwargs):
             if not annotate:
                 return output
 
+            tape = get_working_tape()
+
+            coeff_form = form.coefficients()
+            for coeff in coeff_form:
```
sv2518, line 26: You do not use coeff_form later, so maybe merge line 25 and 26 in one?

sv2518: Same for block_variable in 31

firedrake/adjoint/blocks.py
```diff
@@ -247,3 +249,38 @@ def recompute_component(self, inputs, block_variable, idx, prepared):
...
+        N, ops = inputs[-1], inputs[:-1]
+        return N._ufl_expr_reconstruct_(*ops)
+
+    def evaluate_adj_component(self, inputs, adj_inputs, block_variable, idx, prepared=None):
+        if self.point_op == block_variable.output:
+            # We are not able to calculate derivatives wrt initial guess.
+            return None
+
+        q_rep = block_variable.saved_output
+        N = prepared
```
sv2518, line 274: Can you not just use `prepared`?

firedrake/adjoint/solving.py
```diff
@@ -54,6 +51,18 @@ def wrapper(*args, **kwargs):
...
 
+            if isinstance(args[0], ufl.equation.Equation):
+                coeff_form = args[0].lhs.coefficients()
+                for coeff in coeff_form:
+                    if isinstance(coeff, ufl.ExternalOperator):
+                        block_extops = PointwiseOperatorBlock(coeff, *args, **sb_kwargs)
+                        tape.add_block(block_extops)
+
+                        block_variable = coeff.block_variable
+                        block_extops.add_output(block_variable)
```
sv2518, line 62: Same here.

firedrake/adjoint/solving.py
```diff
@@ -54,6 +51,18 @@ def wrapper(*args, **kwargs):
                 block_variable = args[1].function.create_block_variable()
             block.add_output(block_variable)
 
+            if isinstance(args[0], ufl.equation.Equation):
+                coeff_form = args[0].lhs.coefficients()
+                for coeff in coeff_form:
```
sv2518, line 56: Same here.

tests/pointwiseoperator/test_point_solve.py
nbouziani: Done, thank you

firedrake/adjoint/assembly.py
nbouziani: Done

firedrake/adjoint/blocks.py
nbouziani: I have replaced `prepared` by `N` as it seems more intuitive to see what is happening

dham:
Superseded by ExternalOperator.

PR #1675, opened 2020-05-06
Update adjoint notebook

jwallwork23:
Ahead of the MPECDT Firedrake tutorial next week, I thought it might be worth making a few of small changes to the adjoint notebook:
* No longer any need to tell the user to explicitly install pyadjoint.
* Bracket formatting.
* Idea for a little exercise so attendees have something to do once they have read and run the notebook.


jwallwork23:
I added another notebook which shows how to extract adjoint solutions from solve blocks.

docs/notebooks/11-extract-adjoint-solutions.ipynb
```diff
@@ -0,0 +1,2030 @@
...
+   "outputs": [],
+   "source": [
+    "stop_annotating();"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "Given the definition of $J$, we should exect its derivative w.r.t. $nu$ to be zero:"
```
dham, line 1034: It's a really unfortunate choice of functional if it doesn't depend on the control at all! Could we not use the value of u at the outflow instead?

jwallwork23: Sorry, I'm not sure I understand. Do you mean we should change the control or the functional? Without realising it, I've used one of the two controls and a very similar functional as in one of the dolfin-adjoint demos: http://www.dolfin-adjoint.org/en/latest/documentation/tutorial.html. The other control used there is the initial condition.

dham: Well you've chosen a combination of functional and control such that the functional doesn't depend on the control. That sort of defeats the purpose of solving an adjoint problem! In fact, your functional only depends on the boundary conditions, so it doesn't even really depend on the PDE solution. That's why I suggested changing the functional to be the outflow velocity, because that depends on the PDE solution and therefore on the control. This would result in a nonzero computed gradient.

jwallwork23: Sorry, I just realised that I wrote 'upstream' in the demo when I meant 'downstream'! The functional is evaluated at the RH boundary, which is the outflow. Is that right, or did you mean something else for the functional?

dham: In that case I don't understand the argument that the derivative is expected to be zero.

jwallwork23: Okay, I've removed the parts with UFL `derivative` calls for simplicity of presentation and added a sentence discussing the negative gradient of J w.r.t. control.

jwallwork23:
Apologies @dham, just realised my force push overwrote your minor fixes. Please could you push them again?

jwallwork23:
Thanks! I recall you also removed the comment about installing with dolfin-adjoint, so have pushed that, too.

PR #1686, opened 2020-05-11
wence/fix/mixed real tensorproduct

wence-:
Need to handle the real_tensorproduct case inside mixed/component spaces.


colinjcotter:
This works for my issue and the test looks good.

PR #1687, opened 2020-05-11
Update how PlaneSmoother computes its coordinates

pefarrell:
At the moment `PlaneSmoother` just gets the coordinates of a point from the Plex. This has the downside that it does the wrong thing when the PETSc's opinion of the coordinates and Firedrake's diverge, as when applied to a periodic problem.

This PR fixes this by always using Firedrake's opinion of the coordinates. It also adds a test to show how to make a periodic mesh hierarchy for solving periodic problems with GMG and how to use `PlaneSmoother` to do line relaxation on a problem with a large aspect ratio.

### DIFF
--- a/firedrake/preconditioners/patch.py
+++ b/firedrake/preconditioners/patch.py
@@ -13,6 +13,7 @@ from functools import partial
 import numpy
 from ufl import VectorElement, MixedElement
 from tsfc.kernel_interface.firedrake_loopy import make_builder
+import weakref
 
 import ctypes
 from pyop2 import op2
@@ -568,11 +569,21 @@ def select_entity(p, dm=None, exclude=None):
 class PlaneSmoother(object):
     @staticmethod
     def coords(dm, p):
-        coordsSection = dm.getCoordinateSection()
-        coordsDM = dm.getCoordinateDM()
-        dim = coordsDM.getDimension()
-        coordsVec = dm.getCoordinatesLocal()
-        return dm.getVecClosure(coordsSection, coordsVec, p).reshape(-1, dim).mean(axis=0)
+        mesh = dm.getAttr("__firedrake_mesh__")
+
+        data = mesh.coordinates.dat.data_ro
+        coordsDM = mesh.coordinates.function_space().dm
+        coordsSection = coordsDM.getDefaultSection()
+
+        closure_of_p = [x for x in dm.getTransitiveClosure(p, useCone=True)[0] if coordsSection.getDof(x) > 0]
+
+        gdim = data.shape[1]
+        bary = numpy.zeros(gdim)
+        for p_ in closure_of_p:
+            (dof, offset) = (coordsSection.getDof(p_), coordsSection.getOffset(p_))
+            bary += data[offset:offset+dof].reshape(gdim)
+        bary /= len(closure_of_p)
+        return bary
 
     def sort_entities(self, dm, axis, dir, ndiv):
         # compute
@@ -658,6 +669,7 @@ class PatchBase(PCSNESBase):
 
         mesh = J.ufl_domain()
         self.plex = mesh._plex
+        self.plex.setAttr("__firedrake_mesh__", weakref.proxy(mesh))
         self.ctx = ctx
 
         if mesh.cell_set._extruded:
--- a/firedrake/preconditioners/patch.py
+++ b/firedrake/preconditioners/patch.py
@@ -592,8 +592,10 @@ class PlaneSmoother(object):
 
         mesh = dm.getAttr("__firedrake_mesh__")
         ele = mesh.coordinates.function_space().ufl_element()
-        if ele.family() in ["Discontinuous Lagrange", "DQ"]:
-            # Ah, we got a periodic mesh. We need to interpolate to CGk
+        V = mesh.coordinates.function_space()
+        if V.finat_element.entity_dofs() == V.finat_element.entity_closure_dofs():
+            # We're using DG or DQ for our coordinates, so we got
+            # a periodic mesh. We need to interpolate to CGk
             # with access descriptor MAX to define a consistent opinion
             # about where the vertices are.
             CGkele = ele.reconstruct(family="Lagrange")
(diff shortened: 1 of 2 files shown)


firedrake/preconditioners/patch.py
```diff
@@ -568,11 +571,41 @@ def select_entity(p, dm=None, exclude=None):
...
+            # Ah, we got a periodic mesh. We need to interpolate to CGk
+            # with access descriptor MAX to define a consistent opinion
+            # about where the vertices are.
+
+            coords = dm.getAttr("__firedrake_interpolated_coords__")
+            if coords is None:
+                CGkele = ele.reconstruct(family="Lagrange")
+                # Can't do
+                # CGk = FunctionSpace(mesh, CGkele)
+                # because equality fails between a mesh and a weakref.proxy.
```
wence-, line 588: I think this comment is misleading:
```
mesh = ...
mesh2 = weakref.proxy(mesh)
mesh2 == mesh
=> True
``` 
(at least for me)

But weakref.proxies are not hashable, so the interpolation fails for me. I think.

firedrake/preconditioners/patch.py
```diff
@@ -568,11 +571,41 @@ def select_entity(p, dm=None, exclude=None):
...
+                CGk = FunctionSpace(mesh.coordinates.function_space().mesh(), CGkele)
+                coords = interpolate(mesh.coordinates, CGk, access=op2.MAX)
+                dm.setAttr("__firedrake_interpolated_coords__", coords)
```
wence-, line 591: Why do you need to save this?

wence-: The reason is that this creates a refcycle (which we're trying to avoid). You can hang it off the mesh if you like. Notice that this breaks if the mesh moves. Given that this setup only gets called once per solver object creation, I would rather just not cache this at all.

firedrake/preconditioners/patch.py
```diff
@@ -568,11 +571,41 @@ def select_entity(p, dm=None, exclude=None):
...
+        else:
+            coords = mesh.coordinates
+
+        coordsV = coords.function_space()
+        data = coords.dat.data_ro
```
wence-, line 596: I think this will fail in parallel because if `p` is a non-ghost cell, the closure may contain points that are `ghost` which will then break.

firedrake/preconditioners/patch.py
```diff
@@ -658,6 +691,7 @@ def initialize(self, obj):
 
         mesh = J.ufl_domain()
         self.plex = mesh._plex
+        self.plex.setAttr("__firedrake_mesh__", weakref.proxy(mesh))
```
wence-, line 694: Add comment. Maybe add a `destroy` method to the PC class that removes this attribute.

wence-: Petsc will call that destroy when the PC goes out of scope, so you can do teardown then.

tests/regression/test_line_smoother_periodic.py
```diff
@@ -0,0 +1,98 @@
...
+                                      "pc_type": "telescope",
+                                      "pc_telescope_subcomm_type": "contiguous",
+                                      "telescope_pc_type": "lu",
+                                      "telescope_pc_factor_mat_solver_type": "superlu_dist"}}
+
+    params = {**base, **params}
+
+    problem = NonlinearVariationalProblem(F, u, bcs=bc)
+    solver = NonlinearVariationalSolver(problem, solver_parameters=params)
+    solver.solve()
```
wence-, line 96: Do you need the multigrid aspect to test the plane smoother, or is it sufficient to do a 1-level method. We're trying to reduce the total testing loading. If you need the multilevel to check some other things that's fine.

firedrake/preconditioners/patch.py
pefarrell: The `coords` method is called once per plex point. I didn't want to interpolate the entire coordinate field once per plex point. I could store a weakref, but then someone would have to store a strong reference for it not to be garbage-collected?

As for the mesh moving --- I think everything in the `PlaneSmoother` construction has to be rebuilt then. It's probably safest for the user just to reconstruct the entire solver object?

wence-: Can you move the creation of this to the `sort_entities` function then. Make the coordinates, pass them as an argument here and then delete them.

tests/regression/test_line_smoother_periodic.py
pefarrell: I have reduced the size of the mesh (coarse grid is 3x3, fine grid is 6x6). Most of the runtime is in importing firedrake (at least on my machine).

firedrake/preconditioners/patch.py
pefarrell: Great point, done.

firedrake/preconditioners/patch.py
pefarrell: Thanks for pointing this out. I've switched it to use `data_ro_with_halos`.

firedrake/preconditioners/patch.py
pefarrell: Good point. I've moved it to `sort_entities`.

firedrake/preconditioners/patch.py
pefarrell: Thanks, fixed.

firedrake/preconditioners/patch.py
```diff
@@ -568,11 +571,41 @@ def select_entity(p, dm=None, exclude=None):
...
     def coords(dm, p):
-        coordsSection = dm.getCoordinateSection()
-        coordsDM = dm.getCoordinateDM()
-        dim = coordsDM.getDimension()
-        coordsVec = dm.getCoordinatesLocal()
-        return dm.getVecClosure(coordsSection, coordsVec, p).reshape(-1, dim).mean(axis=0)
+        mesh = dm.getAttr("__firedrake_mesh__")
+
+        ele = mesh.coordinates.function_space().ufl_element()
+        assert ele.family() in ["Lagrange", "Discontinuous Lagrange"]
```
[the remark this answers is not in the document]
pefarrell: OK, I've used this as the test of whether to build the interpolated coordinates or not.

PR #1689, opened 2020-05-14
Added missingDiagonal for matrix-free

IvanYashchuk:
Together with https://bitbucket.org/petsc/petsc4py/pull-requests/149 will allow using `matfree` matrices with PETSc TS.


pefarrell:
Personally, I think `TS` should be changed so that it doesn't require this silly method. But the change looks reasonable per se.

wence-:
Thanks.

PR #1691, opened 2020-05-17
wence/feature/expression assembler

wence-:
Rewrite expression assembler to use TSFC technology. Also expunges some of the more egregious corner cases we supported.

Needs firedrakeproject/tsfc#216.

### DIFF
--- a/firedrake/assemble_expressions.py
+++ b/firedrake/assemble_expressions.py
@@ -1,54 +1,109 @@
-import weakref
-
-import ufl
-from ufl.algorithms import ReuseTransformer
-from ufl.corealg.map_dag import MultiFunction, map_expr_dag
-from ufl.constantvalue import ConstantValue, Zero, IntValue
-from ufl.core.multiindex import MultiIndex
-from ufl.core.operator import Operator
-from ufl.mathfunctions import MathFunction
-from ufl.core.ufl_type import ufl_type as orig_ufl_type
-from ufl import classes
-from collections import defaultdict
 import itertools
+import weakref
+from collections import OrderedDict, defaultdict
 
+import gem
 import loopy
-import pymbolic.primitives as p
+import ufl
+from gem.impero_utils import compile_gem, preprocess_gem
+from gem.node import traversal as gem_traversal
 from pyop2 import op2
-from pyop2.profiling import timed_function
+from pyop2.sequential import Arg
+from tsfc import ufl2gem
+from tsfc.loopy import generate
+from ufl.algorithms.apply_algebra_lowering import LowerCompoundAlgebra
+from ufl.classes import Index, MultiIndex, ConstantValue
+from ufl.corealg.map_dag import map_expr_dags
+from ufl.corealg.multifunction import MultiFunction
+from ufl.corealg.traversal import unique_pre_traversal as ufl_traversal
 
-from firedrake import constant
-from firedrake import function
-from firedrake import utils
-
-from functools import singledispatch
+import firedrake
+from firedrake.utils import ScalarType, cached_property, known_pyop2_safe
 
 
-def ufl_type(*args, **kwargs):
-    r"""Decorator mimicing :func:`ufl.core.ufl_type.ufl_type`.
+def extract_coefficients(expr):
+    return tuple(e for e in ufl_traversal(expr) if isinstance(e, ufl.Coefficient))
 
-    Additionally adds the class decorated to the appropriate set of ufl classes."""
-    def decorator(cls):
-        orig_ufl_type(*args, **kwargs)(cls)
-        classes.all_ufl_classes.add(cls)
-        if cls._ufl_is_abstract_:
-            classes.abstract_classes.add(cls)
+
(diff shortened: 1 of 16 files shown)


firedrake/function.py
```diff
@@ -282,7 +282,9 @@ def copy(self, deepcopy=False):
         return type(self)(self.function_space(), val=val)
 
     def __getattr__(self, name):
-        return getattr(self._data, name)
+        val = getattr(self._data, name)
+        setattr(self, name, val)
+        return val
```
wence-, line 287: These are just helping in the latency bound regime a little bit.

firedrake/assemble_expressions.py
```diff
@@ -1,645 +1,366 @@
...
-from firedrake import constant
-from firedrake import function
-from firedrake import utils
+import firedrake
+from firedrake.utils import ScalarType, cached_property, known_pyop2_safe
 
-from functools import singledispatch
 
+def extract_coefficients(expr):
+    return tuple(e for e in ufl_traversal(expr) if isinstance(e, ufl.Coefficient))
```
sv2518, line 25: Why do you not use expr.coefficients()?

firedrake/assemble_expressions.py
```diff
@@ -1,645 +1,366 @@
...
+    args = []
+    for expr in exprs:
+        retval = coefficients.pop(expr.lvalue)
+        args.append(loopy.GlobalArg(retval.name, shape=retval.shape, dtype=expr.lvalue.dat.dtype))
+        for c in expr.rcoefficients:
+            if c.dat == expr.lvalue.dat:
+                continue
+            var = coefficients[c]
+            args.append(loopy.GlobalArg(var.name, shape=var.shape, dtype=c.dat.dtype))
+    knl = generate(impero_c, args, 16, scalar_type, kernel_name="expression_kernel")
```
sv2518, line 312: The precision argument in tsfc.loopys `generate` does not get used, I think. Maybe we can drop the argument?

wence-: It is indirectly via `ctx.epsilon` in the translation of `gem.Constant`. I'd be happy to drop it though, or just to infer it from `numpy.finfo`.

firedrake/assemble_expressions.py
wence-: `.coefficients()` isn't implemented for `ufl.Expr` objects. If you mean the `Assign.coefficients` property, that uses this function.

wence-: This is a micro-optimisation in that I could use `extract_coefficients` but that takes 25µs whereas this takes 17µs on my machine.

firedrake/assemble_expressions.py
wence-: I refactored tsfc to just handle precision from scalar_type everywhere, so now this is gone.

tests/regression/test_auxiliary_dm.py
```diff
@@ -28,7 +28,7 @@ def mesh(self):
 
     def bcs(self, Z):
         bcs = [DirichletBC(Z.sub(0), self.analytical_solution(Z.ufl_domain()), "on_boundary"),
-               DirichletBC(Z.sub(1), Constant((0, 0)), "on_boundary")]
+               DirichletBC(Z.sub(1), zero(), "on_boundary")]
```
wence-, line 31: This kind of things were wrong. We're assigning to an RT space here, so dof-wise we only have one value, not two. If someone had written `Constant((0, 1))` they would have seen something completely wrong. Now that latter is an error.

firedrake/assemble_expressions.py
```diff
@@ -1,645 +1,424 @@
...
+    def __init__(self):
+        self.varmapping = OrderedDict()
+        MultiFunction.__init__(self)
+        ufl2gem.Mixin.__init__(self)
+
+    # Override shape-based things
+    # Need to inspect GEM shape not UFL shape, due to Coefficients changing shape.
+    def sum(self, o, *ops):
+        shape, = set(o.shape for o in ops)
+        indices = tuple(gem.Index() for _ in shape)
```
sv2518, line 38: Maybe `indices = gem.indices(len(expr.shape))`?

firedrake/assemble_expressions.py
```diff
@@ -1,645 +1,424 @@
...
+        return gem.ComponentTensor(gem.MathFunction('abs', gem.Indexed(expr, indices)),
+                                   indices)
+
+    def conditional(self, o, condition, then, else_):
+        assert condition.shape == ()
+        shape, = set([then.shape, else_.shape])
+        indices = gem.indices(len(shape))
+        return gem.ComponentTensor(gem.Conditional(condition, gem.Indexed(then, indices),
+                                                   gem.Indexed(else_, indices)),
+                                   indices)
```
miklos1, line 68: Kind of late comment, but this pattern of `gem.ComponentTensor(Foo(gem.Indexed(bar, indices), ...), indices)` comes up so often that it would be useful to capture it as helper function.  The helper function could probably go to `ufl2gem.Mixin` (and also used there) and then imported here.

Issue #1697, opened 2020-05-20
install script checks for mpich


colinjcotter:
Install script checks for mpich package but we don't use it.

PR #1698, opened 2020-05-20
Wence/update install

wence-:
Addresses #1697. I am checking that the docker file still builds, but am not testing that it works (how best to do that).


wence-:
Tested the docker build (which worked, so I will drop that commit).

PR #1701, opened 2020-05-21
Fix _ad_assign_numpy and _ad_to_list methods of Constant

stephankramer:
These are called when using one of scipy's optimisation routines with Constants
as control, to pack adjoint values into numpy arrays. Their current version seem to be copied from fenics_adjoint and do not work.

### DIFF
--- a/firedrake/adjoint/constant.py
+++ b/firedrake/adjoint/constant.py
@@ -85,7 +85,7 @@ class ConstantMixin(OverloadedType):
     @staticmethod
     def _ad_assign_numpy(dst, src, offset):
         l = dst.ufl_element().value_size()
-        dst.assign(numpy.reshape(src[offset:offset + l], dst.ufl_shape))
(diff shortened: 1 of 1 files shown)


firedrake/adjoint/constant.py
```diff
@@ -85,16 +84,14 @@ def _ad_dot(self, other, options=None):
 
     @staticmethod
     def _ad_assign_numpy(dst, src, offset):
-        dst.assign(Constant(numpy.reshape(src[offset:offset + dst.value_size()], dst.ufl_shape)))
-        offset += dst.value_size()
+        l = dst.ufl_element().value_size()
+        dst.assign(numpy.reshape(src[offset:offset + l], dst.ufl_shape))
```
wence-, line 88: This was probably wrong before, but almost certainly this assignment wants to be a in a `with no_annotations` block or however it's spelt.

stephankramer:
Thanks, fixed that. Now also tested in https://github.com/dolfin-adjoint/pyadjoint/pull/19

PR #1707, opened 2020-05-23
wence/fix/pointwise expressions

wence-:
Fixes some cases that the firedrake test suite did not cover. Needs a pyop2 change as well OP2/PyOP2#585

### DIFF
--- a/firedrake/assemble_expressions.py
+++ b/firedrake/assemble_expressions.py
@@ -210,7 +210,7 @@ class Assign(object):
                 return tuple(type(self)(s, self.rvalue) for s in self.lvalue.split())
             else:
                 if indices == set([None]):
-                    if len(set(spaces) | {V} - {None}) != 1:
+                    if len((set(spaces) | {V}) - {None}) != 1:
                         # Check that there were no unindexed coefficients
                         raise ValueError("Saw unindexed coefficients in rvalue, "
                                          "perhaps you meant to index the lvalue with .sub(...)")
--- a/firedrake/assemble_expressions.py
+++ b/firedrake/assemble_expressions.py
@@ -117,7 +117,10 @@ class IndexRelabeller(MultiFunction):
 
 class CoefficientSplitter(MultiFunction):
     def coefficient(self, o):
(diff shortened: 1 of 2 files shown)


firedrake/assemble_expressions.py
```diff
@@ -169,6 +190,10 @@ def __init__(self, lvalue, rvalue):
             raise ValueError("lvalue for pointwise assignment must be a coefficient")
         self.lvalue = lvalue
         self.rvalue = ufl.as_ufl(rvalue)
+        n = len(self.lvalue.function_space())
+        if n > 1:
+            self.splitter = Memoizer(_split)
+            self.splitter.n = n
```
wence-, line 196: Will only need to split mixed spaces.

firedrake/assemble_expressions.py
```diff
@@ -358,17 +386,36 @@ def pointwise_expression_kernel(exprs, scalar_type):
...
-            if c.dat == expr.lvalue.dat:
+        for c, arg in zip(expr.coefficients, expr.args):
+            try:
+                var = coefficients.pop(c)
+            except KeyError:
                 continue
-            var = coefficients[c]
+            plargs.append(arg)
             args.append(loopy.GlobalArg(var.name, shape=var.shape, dtype=c.dat.dtype))
+    assert len(coefficients) == 0
```
wence-, line 398: Merging of arguments across the grouped expressions (basically constants only come in once)

firedrake/assemble_expressions.py
```diff
@@ -358,17 +386,36 @@ def pointwise_expression_kernel(exprs, scalar_type):
...
+        for a in self.args:
+            data = a.data()
+            if data is None:
+                raise ReferenceError
+            a.data = a.data()
+        return self.args
+
+    def __exit__(self, *args, **kwargs):
+        for a in self.args:
+            a.data = weakref.ref(a.data)
```
wence-, line 418: Ugly mutation, but much cheaper than rebuilding new Arg objects.

PR #1709, opened 2020-05-25
Add SLEPc to full-featured Docker image

danshapero:
This should make it easier to give demos of how to use Firedrake for eigenvalue problems.


wence-:
This actually passed but jenkins didn't report, will merge.

PR #1710, opened 2020-05-26
install: Build SLEPc like PETSc

wence-:
Fixes build failures on jenkins (e.g. see #1709)

### DIFF
--- a/scripts/firedrake-install
+++ b/scripts/firedrake-install
@@ -731,8 +731,10 @@ def install(package):
     log.info("Installing %s" % package)
     if package == "petsc/":
         build_and_install_petsc()
+    elif package == "slepc/":
+        build_and_install_slepc()
     # The following outrageous hack works around the fact that petsc and co. cannot be installed in developer mode.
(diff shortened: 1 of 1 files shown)


scripts/firedrake-install
```diff
@@ -980,39 +978,35 @@ def build_and_install_libspatialindex():
 
 def build_and_install_slepc():
     petsc_dir, petsc_arch = get_petsc_dir()
-
+    slepc_dir, slepc_arch = get_slepc_dir()
+    assert petsc_arch == slepc_arch
```
dham, line 982: Is it obvious that these should match?

wence-: Well, the definition of `get_slepc_dir` just returns `petsc_arch` in the second slot (SLEPc doesn't have a separate idea of the arch, it just matches what petsc uses). This is belt-and-braces assertion.

wence-:
Zenodo API canary failed (but passes locally, so I blame network), every other part is green.

PR #1719, opened 2020-06-03
Make p-multigrid transfer operators matrix-free

pefarrell:
The current implementation of transfer operators assembles the prolongation matrix. For high polynomial degree this is quite dense, and assembling it is a bad idea. This pull request instead implements matrix-free transfer operators. This is much more efficient in memory and allows us to solve problems we couldn't before.

The p-multigrid tests pass, with the same iteration counts and minor differences in residuals due to roundoff.

### DIFF
--- a/firedrake/preconditioners/pmg.py
+++ b/firedrake/preconditioners/pmg.py
@@ -1,4 +1,5 @@
 from functools import partial
+from itertools import chain
 import numpy as np
 
 from ufl import MixedElement, VectorElement, TensorElement, replace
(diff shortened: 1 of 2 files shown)


wence-:
Is it worth keeping both versions around, with the user an option to pick. For example P2->P1 I imagine is better with a matrix but P16->P8 probably not. This would also facilitate galerkin operators if one wanted those.

Aside, it's clear that we need better interfaces for building these kernels (the prolongation is fine via interpolate machinery, but the current restriction kernel is kind of ugly: one must surely be able to do this fully matrix-free rather than the partially matrix-free that is going on now).

pefarrell:
I'm happy to keep the previous implementation. I agree that it probably makes sense to use it at lower degrees. What kind of interface do you have in mind for someone to select which to use, and for which levels in the hierarchy?

wence-:
Via options? `mg_levels_transfer_mat_type {aij,matfree}` ?

wence-:
I guess the problem might be the callback doesn't provide a DM with the right options prefix?

pefarrell:
OK, done. The DMs are under my control, so I can set the options prefix to be whatever I want.

PR #1724, opened 2020-06-05
tests: Speed up some equation bc tests with better solver options

wence-:
Something is bad with the matfree versions.

I ran out effort trimming more tests. I am concerned with the matfree versions because I think the new solver options I provide should be exact inverses, but are not.

### DIFF
--- a/tests/equation_bcs/test_equation_bcs.py
+++ b/tests/equation_bcs/test_equation_bcs.py
@@ -207,12 +207,8 @@ def test_EquationBC_poisson_matrix(eq_type, mat_type, porder, with_bbc):
     # aij
 
     solver_parameters = {'mat_type': mat_type,
-                         'ksp_type': 'gmres',
-                         'ksp_atol': 1e-10,
-                         'ksp_rtol': 1e-10,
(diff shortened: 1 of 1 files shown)


tests/equation_bcs/test_equation_bcs.py
```diff
@@ -339,21 +331,18 @@ def test_EquationBC_mixedpoisson_matfree_fieldsplit(eq_type, mat_type, porder):
...
-                         'ksp_divtol': 1e8,
+                         'ksp_type': 'fgmres',
+                         'ksp_atol': 1e-11,
+                         'ksp_max_it': 200,
                          'pc_type': 'fieldsplit',
                          'pc_fieldsplit_type': 'schur',
                          'pc_fieldsplit_schur_fact_type': 'full',
-                         'fieldsplit_0_ksp_type': 'gmres',
-                         'fieldsplit_0_ksp_rtol': 1.e-12,
+                         'fieldsplit_0_ksp_type': 'cg',
```
wence-, line 340: It's concerning I need this, and can't get away with `preonly` for the `ksp_type` here, since the PC should be exact.

Issue #1723, opened 2020-06-05
Error when assembling ufl.derivative with mixed function space


IvanYashchuk:
Here's the code that fails at assembly
```python
from firedrake import *
import ufl
mesh = UnitSquareMesh(10, 10)
V = FunctionSpace(mesh, "P", 1)
W = V * V
w = Function(W)
dw = TestFunction(W)
w0, _ = split(w)
dw0, _ = split(dw)
j = inner(grad(w0), grad(w0))*dx
J = ufl.derivative(j, w0, dw0)
# this now raises an error
assemble(J)
```
The following error is raised
```
# ~/devdir/firedrake/src/firedrake/firedrake/tsfc_interface.py in _read_from_disk(cls, key, comm)
#      80         if val is None:
# ---> 81             raise KeyError("Object with key %s not found" % key)
#      82         return cls._cache.setdefault(key, pickle.loads(val))
# KeyError: 'Object with key 0da115d0286568e9535ebd02399dac10 not found'
# During handling of the above exception, another exception occurred:
# ...
# UFLException: Expecting scalar coefficient in this branch.
```
The above code works in FEniCS.
Using `J = inner(grad(w0), grad(dw0))*dx` works as expected, so there is some bug specifically related to using `ufl.derivative` on splitted arguments.
Also `firedrake.derivative` is broken in this case
```python
~/firedrake/src/firedrake/firedrake/ufl_expr.py in derivative(form, u, du, coefficient_derivatives)
    140     # TODO: What about Constant?
    141     u_is_x = isinstance(u, ufl.SpatialCoordinate)
--> 142     if not u_is_x and len(u.split()) > 1 and set(extract_coefficients(form)) & set(u.split()):
    143         raise ValueError("Taking derivative of form wrt u, but form contains coefficients from u.split()."
    144                          "\nYou probably meant to write split(u) when defining your form.")

AttributeError: 'Indexed' object has no attribute 'split'
```

IvanYashchuk:
Using `J = ufl.algorithms.expand_derivatives(ufl.derivative(j, w0, dw0))` assembles without error.

wence-:
Can you try this patch? The problem is some missing zero simplification in the form splitter leaving us with a nonsense J[1] form. By calling expand derivatives beforehand everything works its way out. Alternately, here I just handle that case.

```diff
diff --git a/firedrake/formmanipulation.py b/firedrake/formmanipulation.py
index 472c4e0e..b56e13d5 100644
[...]

PR #1729, opened 2020-06-09
Fix for derivative wrt to part of mixed function

IvanYashchuk:
This PR includes changes proposed by @wence- to fix an error reported in #1723 
I've added a test that fails on master and passes on this PR's branch.

### DIFF
--- a/firedrake/ufl_expr.py
+++ b/firedrake/ufl_expr.py
@@ -139,7 +139,8 @@ def derivative(form, u, du=None, coefficient_derivatives=None):
     """
     # TODO: What about Constant?
     u_is_x = isinstance(u, ufl.SpatialCoordinate)
-    if not u_is_x and len(u.split()) > 1 and set(extract_coefficients(form)) & set(u.split()):
+    uc, = extract_coefficients(u)
+    if not u_is_x and len(uc.split()) > 1 and set(extract_coefficients(form)) & set(uc.split()):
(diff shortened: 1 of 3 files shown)


tests/regression/test_split.py
```diff
@@ -58,3 +58,20 @@ def test_split_function_derivative():
...
+
+    x = Function(W)
+    u, p = split(x)
+    v, q = TestFunctions(W)
+
+    I = inner(grad(u), grad(u))*dx
+    F = derivative(I, u, v)
+    J = derivative(F, u, TrialFunctions(W)[0])
+
+    assert assemble(F) and assemble(J)
```
wence-, line 77: Can we check that we got the right values?

firedrake/ufl_expr.py
```diff
@@ -139,7 +139,8 @@ def derivative(form, u, du=None, coefficient_derivatives=None):
     """
     # TODO: What about Constant?
     u_is_x = isinstance(u, ufl.SpatialCoordinate)
-    if not u_is_x and len(u.split()) > 1 and set(extract_coefficients(form)) & set(u.split()):
+    uc, = extract_coefficients(u)
```
wence-, line 142: This extraction fails for coordinate derivatives `if u_is_x`, can you handle that?

IvanYashchuk:
How to make Jenkins run the tests for the latest commit?

wence-:
Can you just rebase on top of master and force push? Then hopefully jenkins will notice again.

PR #1730, opened 2020-06-09
wence/mixed matrix assembly

wence-:
Needs OP2/PyOP2#587. Allows assembly directly into mixed matrices. In particular, widens the applicability of slate to problems with more than one lagrange multiplier.

I don't have a test of that functionality, but check some things now.


firedrake/assemble.py
```diff
@@ -169,239 +160,240 @@ def thunk():
...
-              allocate_only=False,
-              zero_tensor=True,
-              diagonal=False):
-    r"""Assemble the form or Slate expression f and return a Firedrake object
-    representing the result. This will be a :class:`float` for 0-forms/rank-0
-    Slate tensors, a :class:`.Function` for 1-forms/rank-1 Slate tensors and
-    a :class:`.Matrix` for 2-forms/rank-2 Slate tensors.
+def get_matrix(form, mat_type, sub_mat_type, *, bcs=None,
+               options_prefix=None, tensor=None, appctx=None,
+               form_compiler_parameters=None):
```
sv2518, line 165: Maybe add documentation?

firedrake/bcs.py
```diff
@@ -271,6 +271,9 @@ class DirichletBC(BCBase, DirichletBCMixin):
     @DirichletBCMixin._ad_annotate_init
     def __init__(self, V, g, sub_domain, method="topological"):
         super().__init__(V, sub_domain, method=method)
+        if len(V) > 1:
+            raise ValueError("Cannot apply boundary conditions on mixed spaces directly.\n"
```
sv2518, line 275: Why not?

sv2518, line 275: Why not?

firedrake/assemble.py
```diff
@@ -169,239 +160,240 @@ def thunk():
...
+    matfree = mat_type == "matfree"
```
sv2518, line 166: Maybe add documentation?

firedrake/assemble.py
```diff
@@ -169,239 +160,240 @@ def thunk():
...
-        mat_type = parameters.parameters["default_matrix_type"]
-    if mat_type not in ["matfree", "aij", "baij", "nest", "dense"]:
-        raise ValueError("Unrecognised matrix type, '%s'" % mat_type)
-    if sub_mat_type is None:
-        sub_mat_type = parameters.parameters["default_sub_matrix_type"]
-    if sub_mat_type not in ["aij", "baij"]:
-        raise ValueError("Invalid submatrix type, '%s' (not 'aij' or 'baij')", sub_mat_type)
+    if isinstance(form, slate.TensorBase):
+        # FIXME: inherit from slate form somehow.
+        integral_types = {"cell"}
```
sv2518, line 184: In the KernelBuilder of slac, the integral type is inherited from TSFC and TSFC inherits it from the UFL form if I understand it, right. Maybe something like `[integral.integral_type for integral in form.form.integrals()]` is what you need here?

firedrake/bcs.py
wence-: One needs to index the space (`V.sub(...)`) I can add that to the message. This valueerror is migrating from inside assembly to dirichletbc creation.

firedrake/assemble.py
wence-: Well, a slate kernel only iterates over cells, so it can only make contributions through cell_node_maps (hence `cell` here). This is being used to determine the sparsity pattern of the resulting global matrix, so getting the integral types from the form inside the kernel is probably wrong.

I guess a slate `TensorBase` can advertise an `integrals()` method that returns a 1-iterable of objects that return `integral_type() == cell`.

Something like

```python
class MockIntegral(object):
    integral_type = lambda: "cell"

MockIntegral = MockIntegral() # singleton
...
class Tensor(...):
    integrals = (MockIntegral, )
```

firedrake/assemble.py
```diff
@@ -169,239 +160,240 @@ def thunk():
...
+    else:
+        integral_types = set(i.integral_type() for i in form.integrals())
+        for bc in bcs:
+            integral_types.update(integral.integral_type()
+                                  for integral in bc.integrals())
```
wence-, line 189: @ksagiyam I migrated this from the previous assembly code. Why can we not apply equation bcs to slate tensors? Moreover, if this is a technical limitation, I suppose that one should raise a `NotImplementedError` in the slate case.

firedrake/assemble.py
thomasgibson: @wence- is correct, while a Slate expression may (and often does) contain integrals of _each_ type, it only makes sense to iterate cell-wise. It feels odd for a Slate expression to advertise an integral type.

An alternative would be to create a function that dispatches on the expression (UFL.form vs Slate.TensorBase). Something like:

```python
@singledispatch
def iteration_domains(expr, bcs):
    raise ValueError()

@iteration_domains.register(slate.TensorBase)
def _(expr, bcs):
    return {"cell"}

@iteration_domains.register(ufl.Form)
def _(expr, bcs):
    integral_types = set(i.integral_type() for i in expr.integrals())
    for bc in bcs:
        integral_types.update(integral.integral_type()
                              for integral in bc.integrals())
    return integral_types
```
Doing so would allow you to replace all this with
```python
integral_types = iteration_domains(expr, bcs)
```
This is just an alternative way to write this. I think it might be worth a discussion / debate whether or not Slate expressions should support something other than "cellwise" assembly. I personally think it should _always_ be cellwise.

I am willing to be convinced otherwise if there is a concrete application for assembling a Slate expression "facet-wise."

firedrake/assemble.py
```diff
@@ -545,80 +518,124 @@ def get_map(x):
...
+
+        create_op2arg = functools.partial(matrix_arg,
+                                          all_bcs=tuple(chain(*bcs)),
+                                          tensor=tensor,
+                                          Vrow=test.function_space(),
+                                          Vcol=trial.function_space())
+    elif assembly_rank == AssemblyRank.VECTOR:
+        if diagonal:
+            # actually a 2-form but throw away the trial space
+            test, trial = f.arguments()
```
thomasgibson, line 609: Might be a good idea to make sure the test and trial spaces are equal.

firedrake/assemble.py
thomasgibson: Some light documentation for all the new functions would be nice.

And this comment is a bit nitpicky but is it appropriate to call the first argument `form` when it's clearly not always a form? I realize we've done it this way for a long while, but it might be more accurate to simply replace `form` with `expr` here (and elsewhere when appropriate).

firedrake/assemble.py
wence-: Well, I think I would prefer just mocking the interface that ufl Forms provide and then there doesn't need to be any type-based dispatch here. The original idea was that slate objects would look "enough-like" forms that assembly didn't need to care. We didn't quite get there, but we nearly have!

firedrake/assemble.py
wence-: Done both of those things.

firedrake/assemble.py
```diff
@@ -545,80 +518,124 @@ def get_map(x):
...
+    else:
+        assembly_rank = AssemblyRank.SCALAR
+
+    if assembly_rank == AssemblyRank.MATRIX:
+        test, trial = f.arguments()
+        tensor, zeros, result = get_matrix(f, mat_type, sub_mat_type,
+                                           bcs=bcs, options_prefix=options_prefix,
+                                           tensor=tensor,
+                                           appctx=appctx,
+                                           form_compiler_parameters=form_compiler_parameters)
```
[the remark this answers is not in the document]
wence-: I had a go at this, and unpicking it further was too complicated for me. Maybe next time.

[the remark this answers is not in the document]
wence-: Not least because sometimes we call `allocate_matrix` etc... directly in various places.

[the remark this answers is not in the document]
ksagiyam: Yes, l agree. Let's do it next time.

firedrake/assemble.py
wence-: Done.

firedrake/assemble.py
wence-: I did the mocking I sketched.

firedrake/assemble.py
wence-: Done.

wence-:
I think I addressed all comments, please have another look.

PR #1731, opened 2020-06-09
flowchart for debugging install problems

danshapero:
This adds the Sphinx graphviz extension and a .dot file containing the flowchart of how to debug broken installs. I put a few things on the graph to start but obviously there are a lot more, which I can add based on previous issues.

You can bikeshed a lot on the formatting of the rendered graph. I think the important things are to match the font face and size of the rest of the webpage and that it doesn't get super wide.

### DIFF
--- a/docs/source/conf.py
+++ b/docs/source/conf.py
@@ -24,8 +24,10 @@
 # Add any Sphinx extension module names here, as strings. They can be extensions
 # coming with Sphinx (named 'sphinx.ext.*') or your custom ones.
 extensions = ['sphinx.ext.autodoc',
-              'sphinx.ext.mathjax', 'sphinx.ext.intersphinx',
+              'sphinx.ext.mathjax',
+              'sphinx.ext.intersphinx',
               'sphinx.ext.viewcode',
+              'sphinx.ext.graphviz',
               'sphinxcontrib.youtube',
               'sphinxcontrib.bibtex']
 
--- a/docs/source/download.rst
+++ b/docs/source/download.rst
@@ -57,6 +57,26 @@ venv_ as above and then run::
   where ``LOCALHOSTNAME`` is the name returned by running the `hostname`
   command. Should the local host name change, this may require updating.
 
+Upgrade
+-------
+
+The install script will install an upgrade script in
+`firedrake/bin/firedrake-update`. Running this script will update
+Firedrake and all its dependencies.
+
+.. note::
+
+   You should activate the venv_ before running
+   `firedrake-update`.
+
+Just like the ``firedrake-install`` script, running::
+
+    firedrake-update --help
+
+gives a full list of update options. For instance additional Firedrake
+packages can be installed into an existing Firedrake installation using
+``firedrake-update``.
+
 Debugging install problems
 --------------------------
 
@@ -82,6 +102,23 @@ diagnose what's going wrong, **please include the following log files**:
 Likewise, if it's ``firedrake-update`` that fails, please include the
 file ``firedrake-update.log``. You can find this in the Firedrake
 virtual environment.
+
+Recovering from a broken installation script
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+If you find yourself in the unfortunate position that
+``firedrake-update`` won't run because of a bug, and the bug has been
+fixed in Firedrake master, then the following procedure will rebuild
+``firedrake-update`` using the latest version.
+
+From the top directory of your Firedrake install,
+type::
+
+  cd src/firedrake
+  git pull
+  ./scripts/firedrake-install --rebuild-script
+
+You should now be able to run ``firedrake-update``.
 
 System requirements
 -------------------
@@ -164,44 +201,6 @@ MacPorts installed, you should ensure that ``/opt/local/bin`` and
 ``/opt/local/sbin`` are removed from your ``PATH`` when installing or
 using Firedrake. This should ensure that no MacPorts installed tools
 are found.
-
-Upgrade
--------
-
-The install script will install an upgrade script in
-`firedrake/bin/firedrake-update`. Running this script will update
-Firedrake and all its dependencies.
-
-.. note::
-
-   You should activate the venv_ before running
-   `firedrake-update`.
-
-Just like the ``firedrake-install`` script, running::
-
-    firedrake-update --help
-
-gives a full list of update options. For instance additional Firedrake
-packages can be installed into an existing Firedrake installation using
-``firedrake-update``.
-
-
-Recovering from a broken installation script
-~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
-
-If you find yourself in the unfortunate position that
-``firedrake-update`` won't run because of a bug, and the bug has been
-fixed in Firedrake master, then the following procedure will rebuild
-``firedrake-update`` using the latest version.
-
-From the top directory of your Firedrake install,
-type::
-
-  cd src/firedrake
-  git pull
-  ./scripts/firedrake-install --rebuild-script
-
-You should now be able to run ``firedrake-update``.
 
 
 Visualisation software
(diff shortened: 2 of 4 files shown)


ReubenHill:
How much depth would we like this to go into? 
Would it be worth adding something about MPICH being preferred over OpenMPI? PETSc install options could probably have a whole branch of the chart.

danshapero:
This is ready for more review and discussion. In a102cbd I created a new section called "System anti-requirements". This includes Anaconda and macports. There isn't anything in the docs now that describes the problems with Anaconda and that's a very common issue.

docs/source/download.rst
```diff
@@ -67,11 +57,41 @@ venv_ as above and then run::
...
+<https://github.com/firedrakeproject/firedrake/issues>`__. To help us
+diagnose what's going wrong, **please include the following log files**:
+
+* ``firedrake-install.log`` from Firedrake
+* ``configure.log`` and ``make.log`` from PETSc
+
+Likewise, if it's ``firedrake-update`` that fails, please include the
+file ``firedrake-update.log``. The PETsc log files live in the
+directory ``src/petsc/`` inside the directory where the Firedrake
+virtual environment was created.
```
ReubenHill, line 82: Can you also make it clear exactly where `firedrake-install.log` and `firedrake-update.log` can be found?

docs/source/download.rst
```diff
@@ -96,26 +116,38 @@ Windows Subsystem for Linux. There are more detailed
...
+tools, however, make this challenging or impossible for end users.
+
+**Anaconda.** The Anaconda Python distribution and package manager are
+often recommended in introductory data science courses because it does
+effectively handle many aggravating problems of dependency management.
+Unfortunately, Anaconda does a poor job of isolating itself from the
+rest of your system and assumes that it will be both the only Python
+installation and the only supplier of any dependent packages. Anaconda
+will install compilers and MPI compiler wrappers and put its compilers
+right at the top of your `PATH`. This is a problem because Firedrake
```
ReubenHill, line 132: Don't know if you've compiled this but I think ``PATH`` needs two backticks in RST to render as plaintext.

danshapero:
Thanks, I made your fixes. Do you think it's worth reordering things a bit? I was thinking of moving the description of how to upgrade Firedrake to just below "Testing the installation". Then the section on "Recovering from a broken installation script" could be subsumed into the install debug graph and the section containing it.

docs/source/download.rst
```diff
@@ -73,13 +73,15 @@ ask us on Slack or report a bug by creating a new `github issue
...
+* ``configure.log`` and ``make.log`` from PETSc, which you can find in
+  ``src/petsc/`` inside the directory where Firedrake virtual
+  environment was created
 
 Likewise, if it's ``firedrake-update`` that fails, please include the
-file ``firedrake-update.log``. The PETsc log files live in the
-directory ``src/petsc/`` inside the directory where the Firedrake
-virtual environment was created.
+file ``firedrake-update.log``. You can find this in the Firedrake
+virtual environment.
```
ReubenHill, line 84: You mean the firedrake install directory rather than the venv no?

ReubenHill:
> Thanks, I made your fixes. Do you think it's worth reordering things a bit? I was thinking of moving the description of how to upgrade Firedrake to just below "Testing the installation". Then the section on "Recovering from a broken installation script" could be subsumed into the install debug graph and the section containing it.

I think that's a great idea

docs/source/download.rst
danshapero: When I active the venv and run `firedrake-update`, I find the log file in the directory `$VIRTUAL_ENV`. Do you think it's clearer to say the install directory? I described them differently because `firedrake-install.log` will live in whatever directory you invoked the install script from, which is of course outside the venv.

wence-: We should probably just fix that bug

danshapero:
I've reorganized the sections and a bit more explanation on testing the installation.

Several users have turned up in slack in the past thinking there's something wrong with their installation because the test suite takes several hours. I've changed the suggestion there to running `make test_demos` instead of `make alltest`. I can revert this if folks disagree but at the very least that section should say how long the full test suite takes.

docs/source/download.rst
```diff
@@ -33,25 +33,20 @@ use Firedrake::
...
+then run::
 
   cd firedrake/src/firedrake
-  make alltest
+  make test_demos
+
+This will run all of the Firedrake demo codes, which exercise most of
+the functionality of the library. The demos should take about 10-20
+minutes to run. If these fail to run for any reason, please see the
+section below on how to diagnose and debug a failed installation.
```
ReubenHill, line 49: The only thing I'd consider changing here is giving the instructions on how to run all tests if they want to.

danshapero: If you really want I can put that in, but I don't think we should be suggesting that first-time users run a job that can take several hours. If there were a doc section on how to contribute to Firedrake then that would be the place to mention running `make alltest`.

docs/source/conf.py
```diff
@@ -24,8 +24,10 @@
 # Add any Sphinx extension module names here, as strings. They can be extensions
 # coming with Sphinx (named 'sphinx.ext.*') or your custom ones.
 extensions = ['sphinx.ext.autodoc',
-              'sphinx.ext.mathjax', 'sphinx.ext.intersphinx',
+              'sphinx.ext.mathjax',
+              'sphinx.ext.intersphinx',
               'sphinx.ext.viewcode',
+              'sphinx.ext.graphviz',
```
wence-, line 30: Is this an extra sphinx package that `firedrake-install --documentation-dependencies` needs to fetch?

danshapero: I just tried it in a docker image and I think the answer is no. But it does depend on having graphviz, which is a brew / apt install not a pip install. I think I want to do something like

```
if args.documentation_dependencies:
    required_packages.append("graphviz")
```

in the install script -- does that make sense?

docs/source/download.rst
```diff
@@ -33,25 +33,22 @@ use Firedrake::
...
+  cd $VIRTUAL_ENV/src/firedrake
+  python3 -m pytest tests/regression/ -k "poisson_strong | stokes_mini | dg_advection"
 
-  cd firedrake/src/firedrake
-  make alltest
+This command will run a few of the unit tests, which exercise a good
+chunk of the functionality of the library. These tests should take a
+minute or less. If they fail to run for any reason, please see the
+section below on how to diagnose and debug a failed installation. If
+you want to run the entire test suite you can to ``make alltest``
```
ReubenHill, line 50: Typo - says ` you can to ``make alltest`` `

docs/source/download.rst
```diff
@@ -165,6 +127,88 @@ type::
...
+There are three ways to work around this problem.
+
+1. Remove Anaconda entirely.
+2. Modify your ``PATH`` environment variable to remove any traces of
+   Anaconda, then install Firedrake. If you need Anaconda later, you
+   can re-enable it with a shell script that will add those directories
+   back onto your path.
+3. Use a tool like pyenv_, which makes it possible to switch between
+   different versions of Python on a per-terminal session or per-
+   directory basis.
```
ReubenHill, line 193: An FYI - pyenv broke my installation terribly on my mac when I tried to update, presumably because I didn't use pyenv's virtual env tool. I spent quite a long time working out what went wrong. Not sure if it's a great idea to suggest using it over, say, one of the docker images.

docs/source/download.rst
```diff
@@ -207,3 +250,4 @@ packages for which these are also dependencies.
 .. _venv: https://docs.python.org/3/tutorial/venv.html
 .. _homebrew: https://brew.sh/
 .. _PETSc: https://www.mcs.anl.gov/petsc/
+.. _pyenv: https://github.com/pyenv/pyenv
```
ReubenHill, line 253: I guess this link isn't needed

danshapero:
Do you want me to rebase down to fewer commits or leave this as is?

ReubenHill:
@danshapero rebased onto master and force pushed since I don't have permissions to merge if checks haven't passed

ReubenHill (approved):

PR #1732, opened 2020-06-10
Output: fix data projection

tkarna:
- Fixes a bug in `File` when checking if data needs to be projected
- Mesh coordinates are in P1 equispaced function space by default

### DIFF
--- a/firedrake/mesh.py
+++ b/firedrake/mesh.py
@@ -5,6 +5,7 @@ import sys
 import ufl
 import weakref
 from collections import OrderedDict, defaultdict
+from functools import partial
 from ufl.classes import ReferenceGrad
 import enum
 import numbers
@@ -1393,19 +1394,19 @@ def Mesh(meshfile, **kwargs):
         geometric_dim = tcell.topological_dimension()
     cell = tcell.reconstruct(geometric_dimension=geometric_dim)
 
-    element = ufl.VectorElement("Lagrange", cell, 1)
+    element = ufl.FiniteElement("Lagrange", cell, 1, variant="equispaced")
+    vector_element = ufl.VectorElement(element, dim=geometric_dim)
     # Create mesh object
-    mesh = MeshGeometry.__new__(MeshGeometry, element)
+    mesh = MeshGeometry.__new__(MeshGeometry, vector_element)
     mesh._topology = topology
 
-    def callback(self):
+    def callback(self, vector_element=None, geometric_dim=None):
         """Finish initialisation."""
         del self._callback
         # Finish the initialisation of mesh topology
         self.topology.init()
 
-        coordinates_fs = functionspace.VectorFunctionSpace(self.topology, "Lagrange", 1,
-                                                           dim=geometric_dim)
+        coordinates_fs = functionspace.FunctionSpace(self.topology, vector_element)
 
         coordinates_data = dmplex.reordered_coords(plex, coordinates_fs.dm.getDefaultSection(),
                                                    (self.num_vertices(), geometric_dim))
@@ -1416,7 +1417,7 @@ def Mesh(meshfile, **kwargs):
 
         self.__init__(coordinates)
 
-    mesh._callback = callback
+    mesh._callback = partial(callback, vector_element=vector_element, geometric_dim=geometric_dim)
     return mesh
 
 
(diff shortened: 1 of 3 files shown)


firedrake/mesh.py
```diff
@@ -1393,19 +1394,19 @@ def Mesh(meshfile, **kwargs):
...
-    def callback(self):
+    def callback(self, element=None, geometric_dim=None):
         """Finish initialisation."""
         del self._callback
         # Finish the initialisation of mesh topology
         self.topology.init()
 
-        coordinates_fs = functionspace.VectorFunctionSpace(self.topology, "Lagrange", 1,
-                                                           dim=geometric_dim)
+        coordinates_fs = functionspace.VectorFunctionSpace(self.topology, element, dim=geometric_dim)
```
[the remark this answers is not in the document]
tkarna: ok, changed now.

firedrake/mesh.py
```diff
@@ -1393,19 +1394,19 @@ def Mesh(meshfile, **kwargs):
...
-    def callback(self):
+    def callback(self, vector_element=None, geometric_dim=None):
         """Finish initialisation."""
         del self._callback
         # Finish the initialisation of mesh topology
         self.topology.init()
 
-        coordinates_fs = functionspace.VectorFunctionSpace(self.topology, "Lagrange", 1,
-                                                           dim=geometric_dim)
+        coordinates_fs = functionspace.FunctionSpace(self.topology, vector_element)
```
ReubenHill, line 1409: Doesn't the `coordinates_fs` in `ExtrudedMesh` need to be similarly updated? https://github.com/firedrakeproject/firedrake/blob/c611797177a5a06caf4fc2d41646ce09d49deab5/firedrake/mesh.py#L1520

ReubenHill: Also does this complicate the link to the `._ufl_mesh` and `ufl_cell` properties (see #1667) ?
https://github.com/firedrakeproject/firedrake/blob/c611797177a5a06caf4fc2d41646ce09d49deab5/firedrake/mesh.py#L449

ReubenHill: Also, and this might be irrelevant, there's some of the funky utility meshes work by creating a coordinates field then make the mesh from that.

tkarna: Thanks. Yes, `ExtrudedMesh` needs similar update. As here we're requiring that the coordinates should always be in equispaced fs this should affect `._ufl_mesh` as well. Same for the custom meshes.

tkarna: Handling all of this would be easier if we could pass `variant="equispaced"` directly to `FunctionSpace`. Maybe worth changing the API?

tkarna:
Actually `ExtrudedMesh` should be fine: `helement` is inherited from the 2D mesh which is now equispaced, and `velement` is 1D where equispaced and spectral are identical.
https://github.com/firedrakeproject/firedrake/blob/master/firedrake/mesh.py#L1511-L1515

ReubenHill:
> Actually `ExtrudedMesh` should be fine: `helement` is inherited from the 2D mesh which is now equispaced, and `velement` is 1D where equispaced and spectral are identical.
> https://github.com/firedrakeproject/firedrake/blob/master/firedrake/mesh.py#L1511-L1515

Right, but the coordinates field is still created in a vector function space. Is that not an issue?

firedrake/mesh.py
ReubenHill: @wence- might have some sage advice

tkarna:
> > Actually ExtrudedMesh should be fine: helement is inherited from the 2D mesh which is now equispaced, and velement is 1D where equispaced and spectral are identical.
https://github.com/firedrakeproject/firedrake/blob/master/firedrake/mesh.py#L1511-L1515

> Right, but the coordinates field is still created in a vector function space. Is that not an issue?

Hmm, I think the function space is the same, only one is created with `VectorElement`:
```python
elt = FiniteElement('CG', mesh.ufl_cell(), 1)
V = VectorFunctionSpace(mesh, elt)
U = FunctionSpace(mesh, VectorElement(elt))
V == U  # True
```

tkarna:
The equispaced coordinates break pyadjoint firedrake tests. Proposed fix is in dolfin-adjoint/pyadjoint/pull/21

tkarna:
We've decided not to use equispaced fs in the coordinates, closing.

PR #1735, opened 2020-06-12
Fix a segfault I get in a complicated solver.

pefarrell:
Debugged with Lawrence. PETSc's handling of how to assemble problems via field splits and coarsenings is such a mess.

### DIFF
--- a/firedrake/dmhooks.py
+++ b/firedrake/dmhooks.py
@@ -429,6 +429,7 @@ def coarsen(dm, comm):
                  teardown=partial(pop_appctx, cdm, cctx),
                  call_setup=True)
         # Necessary for MG inside a fieldsplit in a SNES.
+        dm.setKSPComputeOperators(firedrake.solving_utils._SNESContext.compute_operators)
         cdm.setKSPComputeOperators(firedrake.solving_utils._SNESContext.compute_operators)
     return cdm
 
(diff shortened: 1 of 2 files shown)


pefarrell:
It's not easy, no. The exhibiting code depends on a library. It would be some work to extract it.

wence-:
The proximate issue is MG + fieldsplit + aux operator on a mixed space and then a split inside? So can you provoke it by just doing a 3-way mass-matrix on two levels and splitting 2 fields out and then splitting inside that?

pefarrell:
OK, test added. It segfaults (on my machine) without this PR, but runs successfully with it.

pefarrell:
(The test was prepared by Ioannis Papadopoulos.)

firedrake/dmhooks.py
```diff
@@ -429,6 +429,7 @@ def coarsen(dm, comm):
                  teardown=partial(pop_appctx, cdm, cctx),
                  call_setup=True)
         # Necessary for MG inside a fieldsplit in a SNES.
+        dm.setKSPComputeOperators(firedrake.solving_utils._SNESContext.compute_operators)
```
colinjcotter, line 432: This is a bit mysterious! Can you add a comment explaining why?

wence-: I'll add it in post. Basically there are hooks that are performed by petsc that attempt to carry compute operators around from outer to inner dms. The outer DM has a kspcomputeoperators_snes attached to it (which is wrong inside a fieldsplit for complex reasons) and that gets transferred onto the coarse DM after we've set up our compute operators. This way we make sure the fine DM has the right compute operators callback.

PR #1736, opened 2020-06-12
blas kernels for p-MG grid transfer

pbrubeck:
Implement custom blas kernels when StandaloneInterpolationMatrix is instantiated between two CG/DG spaces with tensor product cells. This is meant to be temporary until we get structure-preserving symbolic dual evaluation.

As a result, we obtain optimal complexity for the p-MG grid transfer operators.
For this example we have setup a mesh with 64 hexes and CG(N). 

Before the custom kernel we were getting a segfault for N=31, and notice the different complexities.
noblas_4x4x4

After custom kernel:
blas_4x4x4

### DIFF
--- a/firedrake/preconditioners/pmg.py
+++ b/firedrake/preconditioners/pmg.py
@@ -262,25 +262,33 @@ def prolongation_transfer_kernel_aij(Pk, P1):
     return kernel
 
 
-def isTensorProductSpace(V):
+def tensor_product_space_query(V):
+    # Check if a function space V is a tensor product of
+    # either CG(N) or DG(N) on quads or hexes (same N along every direction).
+    # Additionally, return the degree N and a list of strings with the family.
     ele = V.ufl_element()
     cell = ele.cell()
-    isquad = cell.cellname().find("quadrilateral") >= 0
+    isquad = cell.cellname().find("quadrilateral") >= 0  # hex == "quadrilateral * interval"
 
     if isinstance(ele, (firedrake.VectorElement, firedrake.TensorElement)):
         subel = ele.sub_elements()
         ele = subel[0]
 
     N = ele.degree()
-    if isinstance(ele, firedrake.TensorProductElement):
-        subel = ele.sub_elements()
-        ele = subel[0]
+    issupported = True
+    if(type(N) == tuple):
+        issupported = len(set(N)) == 1
         N = N[0]
 
-    family = ele.family()
-    tpfamily = set(["Q", "CG", "Lagrange", "DQ", "DG", "Discontinuous Lagrange"])
-    isquad = isquad and (family in tpfamily)
-    return isquad, N, family
+    if isinstance(ele, firedrake.TensorProductElement):
+        family = [K.family() for K in ele.sub_elements()]
+    else:
+        family = ele.family()
+
+    isCG = set(family).issubset(["Q", "CG", "Lagrange"])
+    isDG = set(family).issubset(["DQ", "DG", "Discontinuous Lagrange"])
+    issupported = issupported and isquad and (isCG or isDG)
+    return issupported, N, family
 
 
 class StandaloneInterpolationMatrix(object):
@@ -301,8 +309,8 @@ class StandaloneInterpolationMatrix(object):
         with self.weight.dat.vec as w:
             w.reciprocal()
 
-        tf, _, _ = isTensorProductSpace(Vf)
-        tc, _, _ = isTensorProductSpace(Vc)
+        tf, _, _ = tensor_product_space_query(Vf)
+        tc, _, _ = tensor_product_space_query(Vc)
         if tf and tc:
             self.make_blas_kernels(Vf, Vc)
         else:
@@ -312,8 +320,8 @@ class StandaloneInterpolationMatrix(object):
     def make_kernels(self, Vf, Vc):
         # The way we transpose the prolongation kernel is suboptimal, but works for general elements.
         # A local matrix has to be generated each time the kernel is executed.
-        # We are waiting for structure-preserving tfsc kernels.
-        # As an alternative we provide blas kernels for tensor product elements.
+        # This is temporary while we wait for structure-preserving tfsc kernels.
+        # As an alternative, we provide blas kernels for tensor product elements.
         self.prolong_kernel = self.prolongation_transfer_kernel_action(Vf, self.uc)
         matrix_kernel = self.prolongation_transfer_kernel_action(Vf, firedrake.TestFunction(Vc))
         element_kernel = loopy.generate_code_v2(matrix_kernel.code).device_code()
@@ -345,22 +353,30 @@ class StandaloneInterpolationMatrix(object):
         return op2.Kernel(ast, ast.name)
 
     def make_blas_kernels(self, Vf, Vc):
+        # Interpolation and restriction kernels between CG/DG tensor product spaces on quads and hexes.
+        # Works by tabulating the coarse 1D Lagrange basis functions in the (Nf+1)-by-(Nc+1) matrix Jhat,
+        # and using the fact that the 2d/3d tabulation is the tensor product J = kron(Jhat, kron(Jhat, Jhat))
         ndim = Vf.ufl_domain().topological_dimension()
         zf = self.get_nodes_1d(Vf)
         zc = self.get_nodes_1d(Vc)
         Jnp = self.barycentric(zc, zf)
+        # Declare array shapes to be used as literals inside the kernels, I follow to the m-by-n convention.
         (mx, nx) = Jnp.shape
         (my, ny) = (mx, nx)
         (mz, nz) = (mx, nx) if ndim == 3 else (1, 1)
-        nscal = np.prod(Vf.shape, dtype=int)
-        mxyz = mx*my*mz
-        nxyz = nx*ny*nz
-        lwork = nscal*mxyz
+        nscal = np.prod(Vf.shape, dtype=int)  # number of components
+        mxyz = mx*my*mz  # dim of Vf scalar element
+        nxyz = nx*ny*nz  # dim of Vc scalar element
+        lwork = nscal*max(mx, nx)*max(my, ny)*max(mz, nz)  # size for work arrays
 
-        Jhat = np.ascontiguousarray(Jnp.T)
+        Jhat = np.ascontiguousarray(Jnp.T)  # BLAS uses FORTRAN ordering
+        # Pass the 1D tabulation as hexadecimal string
         JX = ', '.join(map(float.hex, np.matrix.flatten(Jhat)))
+        # The Kronecker product routines assume 3D shapes, so in 2D we pass one instead of Jhat
         JZ = "JX" if ndim == 3 else "&one"
 
+        # Common kernel to compute y = kron(A3, kron(A2, A1)) * x
+        # Vector and tensor field genearalization from Deville, Fischer, and Mund section 8.3.1.
         kronmxv_code = """
         extern void dgemm_(char *TRANSA, char *TRANSB, int *m, int *n, int *k,
               double *alpha, double *A, int *lda, double *B, int *ldb,
@@ -372,7 +388,8 @@ class StandaloneInterpolationMatrix(object):
             double *A1, double *A2, double *A3,
             double *x , double *y){
 
-        /* Kronecker matrix-vector product
+        /*
+        Kronecker matrix-vector product
 
         y = op(A) * x,  A = kron(A3, kron(A2, A1))
 
@@ -387,7 +404,7 @@ class StandaloneInterpolationMatrix(object):
         Important notes:
         The input data in x is destroyed in the process.
         Need to allocate nel*max(mx, nx)*max(my, ny)*max(mz, nz) memory for both x and y.
-         */
+        */
 
         int m,n,k,s,p,lda;
         char TA1, TA2, TA3;
@@ -427,6 +444,10 @@ class StandaloneInterpolationMatrix(object):
         return;
         }
         """
+
+        # UFL elements order the component DoFs related to the same node contiguously.
+        # We transpose before and after the multiplcation times J to have each component
+        # stored contiguously as a scalar field, thus reducing the number of dgemm calls.
 
         prolong_code = f"""
         {kronmxv_code}
@@ -470,6 +491,7 @@ class StandaloneInterpolationMatrix(object):
             return;
         }}
         """
+
         op2.configuration["ldflags"] = "-lblas"
         self.prolong_kernel = op2.Kernel(prolong_code, "prolongation")
         self.restrict_kernel = op2.Kernel(restrict_code, "restriction")
@@ -480,11 +502,11 @@ class StandaloneInterpolationMatrix(object):
         # Return GLL nodes if V==CG or GL nodes if V==DG
         from FIAT import quadrature
         from FIAT.reference_element import DefaultLine
-        isquad, N, family = isTensorProductSpace(V)
-        assert isquad
-        if family == "Q" or family == "CG" or family == "Lagrange":
+        issupported, N, family = tensor_product_space_query(V)
+        assert issupported
+        if set(family).issubset(["Q", "CG", "Lagrange"]):
             rule = quadrature.GaussLobattoLegendreQuadratureLineRule(DefaultLine(), N+1)
-        elif family == "DQ" or family == "DG" or family == "Discontinuous Lagrange":
+        elif set(family).issubset(["DQ", "DG", "Discontinuous Lagrange"]):
             rule = quadrature.GaussLegendreQuadratureLineRule(DefaultLine(), N+1)
         else:
             raise ValueError("Don't know how to get nodes for %r" % family)
@@ -495,7 +517,7 @@ class StandaloneInterpolationMatrix(object):
     def barycentric(xsrc, xdst):
         # returns barycentric interpolation matrix from xsrc to xdst
         # J[i,j] = phi_j(xdst[i]), where phi_j(xsrc[i]) = delta_{ij}
-        # Lagrange polynomial on xsrc[j]
+        # and phi_j(x) are Lagrange polynomials defined on xsrc[j]
         # use the second form of the barycentric interpolation formula
         # see Trefethen ATAP eq. 5.11
         temp = np.subtract.outer(xsrc, xsrc)
--- a/firedrake/preconditioners/pmg.py
+++ b/firedrake/preconditioners/pmg.py
@@ -263,32 +263,42 @@ def prolongation_transfer_kernel_aij(Pk, P1):
 
 
 def tensor_product_space_query(V):
-    # Check if a function space V is a tensor product of
-    # either CG(N) or DG(N) on quads or hexes (same N along every direction).
-    # Additionally, return the degree N and a list of strings with the family.
-    ele = V.ufl_element()
-    cell = ele.cell()
-    isquad = cell.cellname().find("quadrilateral") >= 0  # hex == "quadrilateral * interval"
+    """
+    Checks whether the custom transfer kernels support the FunctionSpace V.
 
+    V must be either CG(N) or DG(N) on quads or hexes (same N along every direction).
+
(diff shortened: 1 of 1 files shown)


firedrake/preconditioners/pmg.py
```diff
@@ -262,6 +262,27 @@ def prolongation_transfer_kernel_aij(Pk, P1):
     return kernel
 
 
+def isTensorProductSpace(V):
```
pefarrell, line 265: This seems misnamed: based on its name I would have expected it to return a single boolean.

firedrake/preconditioners/pmg.py
```diff
@@ -262,6 +262,27 @@ def prolongation_transfer_kernel_aij(Pk, P1):
...
+
+    if isinstance(ele, (firedrake.VectorElement, firedrake.TensorElement)):
+        subel = ele.sub_elements()
+        ele = subel[0]
+
+    N = ele.degree()
+    if isinstance(ele, firedrake.TensorProductElement):
+        subel = ele.sub_elements()
+        ele = subel[0]
+        N = N[0]
```
pefarrell, line 278: Does this logic do the right thing if I take e.g. the tensor product of CG(k) on quads and Hermite(3) on the interval?

firedrake/preconditioners/pmg.py
```diff
@@ -274,55 +295,236 @@ def __init__(self, Vf, Vc, Vf_bcs, Vc_bcs):
...
+    @staticmethod
+    def prolongation_transfer_kernel_action(Vf, expr):
+        from tsfc import compile_expression_dual_evaluation
+        from tsfc.fiatinterface import create_element
+        coords = Vf.ufl_domain().coordinates
+        to_element = create_element(Vf.ufl_element(), vector_is_mixed=False)
+        ast, oriented, needs_cell_sizes, coefficients, _ = compile_expression_dual_evaluation(expr, to_element, coords, coffee=False)
+        return op2.Kernel(ast, ast.name)
+
+    def make_blas_kernels(self, Vf, Vc):
```
pefarrell, line 347: Could you add an outline comment, explaining the main difference between these kernels and the previous ones?

pefarrell: And ideally a reference to somewhere (which might e.g. be a transfer thesis) where the maths behind it is worked out.

firedrake/preconditioners/pmg.py
```diff
@@ -274,55 +295,236 @@ def __init__(self, Vf, Vc, Vf_bcs, Vc_bcs):
...
+        ndim = Vf.ufl_domain().topological_dimension()
+        zf = self.get_nodes_1d(Vf)
+        zc = self.get_nodes_1d(Vc)
+        Jnp = self.barycentric(zc, zf)
+        (mx, nx) = Jnp.shape
+        (my, ny) = (mx, nx)
+        (mz, nz) = (mx, nx) if ndim == 3 else (1, 1)
+        nscal = np.prod(Vf.shape, dtype=int)
```
pefarrell, line 355: More comments here.

firedrake/preconditioners/pmg.py
```diff
@@ -274,55 +295,236 @@ def __init__(self, Vf, Vc, Vf_bcs, Vc_bcs):
...
+        mxyz = mx*my*mz
+        nxyz = nx*ny*nz
+        lwork = nscal*mxyz
+
+        Jhat = np.ascontiguousarray(Jnp.T)
+        JX = ', '.join(map(float.hex, np.matrix.flatten(Jhat)))
```
pefarrell, line 361: More comments here.

firedrake/preconditioners/pmg.py
```diff
@@ -262,6 +262,35 @@ def prolongation_transfer_kernel_aij(Pk, P1):
     return kernel
 
 
+def tensor_product_space_query(V):
+    # Check if a function space V is a tensor product of
+    # either CG(N) or DG(N) on quads or hexes (same N along every direction).
+    # Additionally, return the degree N and a list of strings with the family.
+    ele = V.ufl_element()
+    cell = ele.cell()
+    isquad = cell.cellname().find("quadrilateral") >= 0  # hex == "quadrilateral * interval"
```
wence-, line 271: `.startswith("quadrilateral")`?

firedrake/preconditioners/pmg.py
```diff
@@ -262,6 +262,35 @@ def prolongation_transfer_kernel_aij(Pk, P1):
     return kernel
 
 
+def tensor_product_space_query(V):
+    # Check if a function space V is a tensor product of
+    # either CG(N) or DG(N) on quads or hexes (same N along every direction).
+    # Additionally, return the degree N and a list of strings with the family.
```
wence-, line 268: Use docstrings, with this format:
```python
"""
Short summary.

Optionally more details...

:arg V: description of argument
:returns: 3-tuple of (use_tensorproduct, degree, family)
"""
```

firedrake/preconditioners/pmg.py
```diff
@@ -262,6 +262,35 @@ def prolongation_transfer_kernel_aij(Pk, P1):
...
+
+    if isinstance(ele, (firedrake.VectorElement, firedrake.TensorElement)):
+        subel = ele.sub_elements()
+        ele = subel[0]
+
+    N = ele.degree()
+    issupported = True
+    if(type(N) == tuple):
+        issupported = len(set(N)) == 1
+        N = N[0]
```
wence-, line 281: Python style is to "beg for forgiveness rather than ask for permission":

```python
N = ele.degree()
issupported = True
try:
   N, = set(N)   
except ValueError:
  # Tuple with different extents
  issupported = False
except TypeError:
  # Just a single int
  pass
```

wence-: This will still work if we suddenly change the type of iterable used for degrees in tensor-product cells from tuples to something else.

firedrake/preconditioners/pmg.py
```diff
@@ -262,6 +262,35 @@ def prolongation_transfer_kernel_aij(Pk, P1):
...
+
+    if isinstance(ele, firedrake.TensorProductElement):
+        family = [K.family() for K in ele.sub_elements()]
+    else:
+        family = ele.family()
```
wence-, line 286: When you take `set(family)` here, you'll get a set of individual characters (because strings are iterable). So prefer:

```
if isinstance(ele, firedrake.TensorProductElement):
   family = set(e.family() for e in ele.sub_elements())
else:
   family = {ele.family()}

issupported = issupported and isquad and (family <= {"Q", "DQ", "Lagrange", "Discontinuous Lagrange"})
```

wence-: Note that "CG" and "DG" never appear at this point, because UFL normalises names.

firedrake/preconditioners/pmg.py
```diff
@@ -274,55 +303,250 @@ def __init__(self, Vf, Vc, Vf_bcs, Vc_bcs):
...
+        issupported, N, family = tensor_product_space_query(V)
+        assert issupported
+        if set(family).issubset(["Q", "CG", "Lagrange"]):
+            rule = quadrature.GaussLobattoLegendreQuadratureLineRule(DefaultLine(), N+1)
+        elif set(family).issubset(["DQ", "DG", "Discontinuous Lagrange"]):
+            rule = quadrature.GaussLegendreQuadratureLineRule(DefaultLine(), N+1)
+        else:
+            raise ValueError("Don't know how to get nodes for %r" % family)
+
+        return np.matrix.flatten(rule.get_points())
```
wence-, line 514: Don't use `np.matrix` (it's been deprecated for about 10 years), use `np.asarray(rule.get_points()).flatten()`

firedrake/preconditioners/pmg.py
```diff
@@ -274,55 +303,250 @@ def __init__(self, Vf, Vc, Vf_bcs, Vc_bcs):
...
+        (my, ny) = (mx, nx)
+        (mz, nz) = (mx, nx) if ndim == 3 else (1, 1)
+        nscal = np.prod(Vf.shape, dtype=int)  # number of components
+        mxyz = mx*my*mz  # dim of Vf scalar element
+        nxyz = nx*ny*nz  # dim of Vc scalar element
+        lwork = nscal*max(mx, nx)*max(my, ny)*max(mz, nz)  # size for work arrays
+
+        Jhat = np.ascontiguousarray(Jnp.T)  # BLAS uses FORTRAN ordering
+        # Pass the 1D tabulation as hexadecimal string
+        JX = ', '.join(map(float.hex, np.matrix.flatten(Jhat)))
```
wence-, line 374: Same point about `np.matrix` -> `np.asarray(...).flatten()`

firedrake/preconditioners/pmg.py
```diff
@@ -274,55 +303,250 @@ def __init__(self, Vf, Vc, Vf_bcs, Vc_bcs):
...
+        # and using the fact that the 2d/3d tabulation is the tensor product J = kron(Jhat, kron(Jhat, Jhat))
+        ndim = Vf.ufl_domain().topological_dimension()
+        zf = self.get_nodes_1d(Vf)
+        zc = self.get_nodes_1d(Vc)
+        Jnp = self.barycentric(zc, zf)
+        # Declare array shapes to be used as literals inside the kernels, I follow to the m-by-n convention.
+        (mx, nx) = Jnp.shape
+        (my, ny) = (mx, nx)
+        (mz, nz) = (mx, nx) if ndim == 3 else (1, 1)
+        nscal = np.prod(Vf.shape, dtype=int)  # number of components
```
wence-, line 367: Available as `Vf.value_size` I think.

firedrake/preconditioners/pmg.py
```diff
@@ -274,55 +303,250 @@ def __init__(self, Vf, Vc, Vf_bcs, Vc_bcs):
...
+        {kronmxv_code}
 
+        void restriction(double *restrict y, const double *restrict x,
+        const double *restrict w){{
+            double JX[{mx}*{nx}] = {{ {JX} }};
+            double t0[{lwork}], t1[{lwork}];
+            double one=1.0E0;
+
+            for(int j=0; j<{mxyz}; j++)
+                for(int i=0; i<{nscal}; i++)
```
wence-, line 483: How big do these loop extents get? If these matrices start falling out of L2 cache, it might be worth loop tiling (https://docs.wence.uk/07-transpose.html)

firedrake/preconditioners/pmg.py
```diff
@@ -274,55 +303,250 @@ def __init__(self, Vf, Vc, Vf_bcs, Vc_bcs):
...
+            kronmxv(1, {nx},{ny},{nz}, {mx},{my},{mz}, {nscal}, JX,JX,{JZ}, t0,t1);
+
+            for(int j=0; j<{nxyz}; j++)
+                for(int i=0; i<{nscal}; i++)
+                    y[i + {nscal}*j] += t1[j + {nxyz}*i];
+            return;
+        }}
+        """
+
+        op2.configuration["ldflags"] = "-lblas"
```
wence-, line 495: Don't do this, instead add extra `ldargs` to the `Kernel`s you make.

firedrake/preconditioners/pmg.py
```diff
@@ -274,55 +303,250 @@ def __init__(self, Vf, Vc, Vf_bcs, Vc_bcs):
...
+        self.prolong_kernel = op2.Kernel(prolong_code, "prolongation")
+        self.restrict_kernel = op2.Kernel(restrict_code, "restriction")
+        return
```
wence-, line 498: No need for empty return at end.

firedrake/preconditioners/pmg.py
```diff
@@ -274,55 +303,250 @@ def __init__(self, Vf, Vc, Vf_bcs, Vc_bcs):
...
+
+        m = mx;  k = nx;  n = ny*nz*nel;
+        lda = (tflag>0)? nx : mx;
+        dgemm_(&TA1, &notr, &m,&n,&k, &one, A1,&lda, x,&k, &zero, y,&m);
+
+        p = 0;  s = 0;
+        m = mx;  k = ny;  n = my;
+        lda = (tflag>0)? ny : my;
+        for(int i=0; i<nz*nel; i++){
+           dgemm_(&notr, &TA2, &m,&n,&k, &one, y+p,&m, A2,&lda, &zero, x+s,&m);
```
wence-, line 431: It's more work, but I'd kind of be interested to know what the performance looks like if you use [libxsmm](https://libxsmm.readthedocs.io/en/latest/libxsmm_mm/#overview) for these small dgemms. I think it should be right in the ballpark where one would expect a decent improvement.

firedrake/preconditioners/pmg.py
pbrubeck: I am not sure about this. I wanted to include the case  "interval * quadrilateral"

wence-: Ah, OK. If you're really wanting to check that the cell you have is a hypercube (I think that's what you want, since right now you would accept `quadrilateral * triangle`!), then the safe thing is to say `FIAT.reference_element.is_hypercube(V.finat_element.cell)`

firedrake/preconditioners/pmg.py
pbrubeck: I don't expect `nscal` being greater than 9. I was thinking that we could even unroll the inner loop.

wence-: Prefer using hints to the compiler than manual unrolling, although these days they're pretty good about doing small extent unrolling. I guess I was thinking of the `mxyz` loop too.

firedrake/preconditioners/pmg.py
pbrubeck: can you give an example?

wence-: `op2.Kernel("void foo(...)", "foo", ldargs=["-lblas"])`

firedrake/preconditioners/pmg.py
pbrubeck: `mxyz` = (N+1)^ndim. For my purposes, the degree N could be as large as 32.

wence-: OK, so loop tiling is likely beneficial at the upper end, but I suspect that we should just leave a comment noting that rather than making this code here more complicated.

firedrake/preconditioners/pmg.py
```diff
@@ -274,55 +313,258 @@ def __init__(self, Vf, Vc, Vf_bcs, Vc_bcs):
...
+        and using the fact that the 2d/3d tabulation is the
+        tensor product J = kron(Jhat, kron(Jhat, Jhat))
+        """
+        ndim = Vf.ufl_domain().topological_dimension()
+        zf = self.get_nodes_1d(Vf)
+        zc = self.get_nodes_1d(Vc)
+        Jnp = self.barycentric(zc, zf)
+        # Declare array shapes to be used as literals inside the kernels, I follow to the m-by-n convention.
+        (mx, nx) = Jnp.shape
+        (my, ny) = (mx, nx)
```
wence-, line 383:
```suggestion
        (my, ny) = (mx, nx) if ndim >= 2 else (1, 1)
```

This is needed otherwise you walk off the end of some arrays for 1D prolongation/restriction.

wence-: For 1D, the standard kernels are equivalent to these tensor product kernels. So should we return `False` in tensor_product_space_query for that case? I think the above change should be done anyway though.

firedrake/preconditioners/pmg.py
```diff
@@ -274,55 +313,258 @@ def __init__(self, Vf, Vc, Vf_bcs, Vc_bcs):
...
+        JZ = "JX" if ndim == 3 else "&one"
+
+        # Common kernel to compute y = kron(A3, kron(A2, A1)) * x
+        # Vector and tensor field genearalization from Deville, Fischer, and Mund section 8.3.1.
+        kronmxv_code = """
+        extern void dgemm_(char *TRANSA, char *TRANSB, int *m, int *n, int *k,
+              double *alpha, double *A, int *lda, double *B, int *ldb,
+              double *beta , double *C, int *ldc);
+
+        void kronmxv(int tflag,
```
wence-, line 403: Can you give this function static linkage? `static void kronmxv`?

wence-:
@pbrubeck can you fix up the out of bounds problems I mentioned above, I think then this is good to go.

firedrake/preconditioners/pmg.py
```diff
@@ -262,6 +262,45 @@ def prolongation_transfer_kernel_aij(Pk, P1):
...
+    """
+    Checks whether the custom transfer kernels support the FunctionSpace V.
+
+    V must be either CG(N) or DG(N) on quads or hexes (same N along every direction).
+
+    :arg V: FunctionSpace
+    :returns: 3-tuple of (use_tensorproduct, degree, family)
+    """
+    from FIAT import reference_element
+    iscube = reference_element.is_hypercube(V.finat_element.cell)
```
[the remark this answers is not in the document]
pbrubeck: for 3d, that'll return a tuple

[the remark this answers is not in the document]
wence-: Ah, ok.

firedrake/preconditioners/pmg.py
```diff
@@ -274,55 +314,259 @@ def __init__(self, Vf, Vc, Vf_bcs, Vc_bcs):
...
+    def make_kernels(self, Vf, Vc):
+        """
+        Interpolation and restriction kernels between arbitrary elements.
+
+        This is temporary while we wait for structure-preserving tfsc kernels.
+        """
+        self.prolong_kernel = self.prolongation_transfer_kernel_action(Vf, self.uc)
         matrix_kernel = self.prolongation_transfer_kernel_action(Vf, firedrake.TestFunction(Vc))
+        # The way we transpose the prolongation kernel is suboptimal.
+        # A local matrix is generated each time the kernel is executed.
```
wence-, line 340: If you care, but I think one should just not spend too much effort on this, you could evaluate on a single element and then hard code those numbers.

firedrake/preconditioners/pmg.py
```diff
@@ -274,55 +314,259 @@ def __init__(self, Vf, Vc, Vf_bcs, Vc_bcs):
...
+        """
+        self.prolong_kernel = op2.Kernel(prolong_code, "prolongation", ldargs=["-lblas"])
+        self.restrict_kernel = op2.Kernel(restrict_code, "restriction", ldargs=["-lblas"])
+
+    @staticmethod
+    def get_nodes_1d(V):
+        # Return GLL nodes if V==CG or GL nodes if V==DG
+        from FIAT import quadrature
+        from FIAT.reference_element import DefaultLine
+        use_tensorproduct, N, family = tensor_product_space_query(V)
```
wence-, line 525: I guess one should also assert that the `variant` of the ufl_element is `"spectral"` not `"equispaced"`?

firedrake/preconditioners/pmg.py
pbrubeck: I wouldn't like to spend time on that.

firedrake/preconditioners/pmg.py
pbrubeck: since spectral is the default, I am just checking if someone is using equispaced points by mistake

pbrubeck:
> Can you add a test for this case? I'm pretty sure the ufc reference interval is [0, 1].

What is the proper way of including tests? 
The barycentric formula is scale and translational invariant, as long as I generate all nodes in [-1,1] it should be okay. I already check and the DefaultLine is in [-1,1], which I use for GLL and GL.

pbrubeck:
> > Can you add a test for this case? I'm pretty sure the ufc reference interval is [0, 1].
> 
> What is the proper way of including tests?
> The barycentric formula is scale and translational invariant, as long as I generate all nodes in [-1,1] it should be okay. I already checked and the DefaultLine is in [-1,1], which I use for GLL and GL.

Okay, I just realized that for euispaced, the dofs are not in lexicographical ordering. The interpolation matrices are the same up to permutation. For now, I'll just switch off the tensor products for equispaced.

wence-:
Thanks!

PR #1743, opened 2020-06-18
Dual Evaluation with FInAT

mkan88:
Performing dual evaluation in FInAT has been a long awaited feature in firedrake. This allows interaction with dual_evaluation methods in FInAT. To be used with corresponding branches in [FInAT #57](https://github.com/FInAT/FInAT/pull/57) and [TSFC #218](https://github.com/firedrakeproject/tsfc/pull/218).

### DIFF
--- a/firedrake/interpolation.py
+++ b/firedrake/interpolation.py
@@ -243,7 +243,7 @@ def _interpolator(V, tensor, expr, subset, arguments, access):
         kernel = op2.Kernel(ast, ast.name, requires_zeroed_output_arguments=True)
     elif hasattr(expr, "eval"):
         to_pts = []
-        for dual in to_element.fiat_equivalent.dual_basis():
+        for dual in to_element.dual_basis():
             if not isinstance(dual, FIAT.functional.PointEvaluation):
                 raise NotImplementedError("Can only interpolate Python kernels with Lagrange elements")
             pts, = dual.pt_dict.keys()
--- a/firedrake/interpolation.py
+++ b/firedrake/interpolation.py
@@ -7,7 +7,7 @@ from ufl.algorithms import extract_arguments
 
 from pyop2 import op2
 
-from tsfc.finatinterface import create_base_element
+from tsfc.finatinterface import create_element
 from tsfc import compile_expression_dual_evaluation
 
 import firedrake
@@ -211,7 +211,7 @@ def make_interpolator(expr, V, subset, access):
 @utils.known_pyop2_safe
 def _interpolator(V, tensor, expr, subset, arguments, access):
     try:
-        to_element = create_base_element(V.ufl_element())
+        to_element = create_element(V.ufl_element())
     except KeyError:
         # FInAT only elements
         raise NotImplementedError("Don't know how to create FIAT element for %s" % V.ufl_element())
@@ -317,7 +317,7 @@ def compile_python_kernel(expression, to_pts, to_element, fs, coords):
     function provided."""
 
     coords_space = coords.function_space()
-    coords_element = create_base_element(coords_space.ufl_element()).fiat_equivalent
+    coords_element = create_element(coords_space.ufl_element()).fiat_equivalent
 
     X_remap = list(coords_element.tabulate(0, to_pts).values())[0]
 
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -40,7 +40,7 @@ pipeline {
               sh 'mkdir tmp'
               dir('tmp') {
                 timestamps {
-                  sh '../scripts/firedrake-install $COMPLEX --tinyasm --disable-ssh --minimal-petsc --slepc --documentation-dependencies --install thetis --install gusto --install icepack --install irksome --no-package-manager || (cat firedrake-install.log && /bin/false)'
+                  sh '../scripts/firedrake-install $COMPLEX --tinyasm --disable-ssh --minimal-petsc --slepc --documentation-dependencies --install thetis --install gusto --install icepack --install irksome --no-package-manager --package-branch tsfc dual_base_fiat --package-branch FInAT dual_base_fiat || (cat firedrake-install.log && /bin/false)'
                 }
               }
             }
--- a/tests/regression/test_interpolate.py
+++ b/tests/regression/test_interpolate.py
@@ -162,44 +162,6 @@ def test_hcurl_2d():
     assert np.allclose(g.dat.data, h.dat.data)
 
 
-def test_hdiv_2d():
-    mesh = UnitCubedSphereMesh(2)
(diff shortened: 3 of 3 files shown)


pbrubeck:
Could you include timing plots for the mult() method in Standalone interpolation matrix?
[fast_interp.py.txt](https://github.com/firedrakeproject/firedrake/files/4798563/fast_interp.py.txt)

wence-:
@pbrubeck I don't think this will yet do the fast thing that you want (it just ports existing functionality right now).

mkan88:
Currently only code restructuring have been implemented, so there shouldn't be any improvement in runtimes (yet)

ReubenHill:
@mkan88 you can avoid unnecessary merge commits like 7f99dc3 by doing `git pull --ff`

firedrake/interpolation.py
```diff
@@ -228,18 +238,18 @@ def _interpolator(V, tensor, expr, subset, arguments, access):
...
-        ast, oriented, needs_cell_sizes, coefficients, _ = compile_expression_dual_evaluation(expr, to_element, coords,
-                                                                                              domain=V.mesh(), coffee=False)
-        kernel = op2.Kernel(ast, ast.name, requires_zeroed_output_arguments=True)
+        ast, oriented, needs_cell_sizes, coefficients, _ = compile_expression_dual_evaluation(expr, element, coords, coffee=False)
+        kernel = op2.Kernel(ast, ast.name)
     elif hasattr(expr, "eval"):
         to_pts = []
-        for dual in to_element.fiat_equivalent.dual_basis():
+        # For FIAT elements
+        for dual in element.dual_basis():
```
ReubenHill, line 246: I think this line, with `element` renamed back to to `to_element` is the only change needed in `firedrake/interpolation.py` thanks to the changes introduced in #1818

firedrake/preconditioners/pmg.py
```diff
@@ -250,14 +250,14 @@ def prolongation_transfer_kernel_aij(Pk, P1):
...
+    from tsfc.finatinterface import create_element
     from firedrake import TestFunction
 
     expr = TestFunction(P1)
     coords = Pk.ufl_domain().coordinates
-    to_element = create_base_element(Pk.ufl_element())
+    element = create_element(Pk.ufl_element())
 
-    ast, oriented, needs_cell_sizes, coefficients, _ = compile_expression_dual_evaluation(expr, to_element, coords, coffee=False)
+    ast, oriented, needs_cell_sizes, coefficients, _ = compile_expression_dual_evaluation(expr, element, coords, coffee=False)
```
ReubenHill, line 260: I believe all the changes here are superseded by #1818 and should be removed from this PR.

ReubenHill (changes requested):

firedrake/interpolation.py
```diff
@@ -228,18 +238,18 @@ def _interpolator(V, tensor, expr, subset, arguments, access):
     if not isinstance(expr, firedrake.Expression):
         if expr.ufl_domain() and expr.ufl_domain() != V.mesh():
             raise NotImplementedError("Interpolation onto another mesh not supported.")
-        ast, oriented, needs_cell_sizes, coefficients, _ = compile_expression_dual_evaluation(expr, to_element, coords,
-                                                                                              domain=V.mesh(), coffee=False)
-        kernel = op2.Kernel(ast, ast.name, requires_zeroed_output_arguments=True)
+        ast, oriented, needs_cell_sizes, coefficients, _ = compile_expression_dual_evaluation(expr, element, coords, coffee=False)
```
wence-, line 241: The new codepath is quite a bit slower than the old one. This suggests that we need to bite the bullet and implement caching of these kernels just like we presently do for assembly kernels.

ReubenHill: Does the ability to store an `Interpolator` object not cut the mustard then?

wence-: Not really.

wence-: The general mantra is that the first time you run your code things might take a while, but new runs should be fast. This is now not the case for anything that does interpolation.

firedrake/interpolation.py
miklos1: 👍 for throwing away unnecessary changes and/or superfluous if-then-else jungle.

ReubenHill:
So somehow it seems N2div on tetrahedron cells is broken (try running `tests/regression/test_interpolation_nodes.py`). Any ideas why that might be?

wence-:
> So somehow it seems N2div on tetrahedron cells is broken (try running tests/regression/test_interpolation_nodes.py). Any ideas why that might be?

Too little information to guess.

ReubenHill:
> > So somehow it seems N2div on tetrahedron cells is broken (try running tests/regression/test_interpolation_nodes.py). Any ideas why that might be?
> 
> Too little information to guess.

If I run the following
```python
from firedrake import *
m = UnitCubeMesh(1, 1, 1)
V = FunctionSpace(m, "N2div", 2)
x = SpatialCoordinate(m)
expr = 2 * SpatialCoordinate(m)
v = Function(V).interpolate(expr)
```
then the generated `kernel` after `kernel = op2.Kernel(ast, ast.name, requires_zeroed_output_arguments=True)` has `len(kernel.code.domains) == 4682` and  `len(kernel.code.instructions) == 6056`. 

A similar issue occurs when interpolating into BDM (see slack discussion)
```python
from firedrake import *
m = UnitSquareMesh(1,1)
V = VectorFunctionSpace(m, "BDM", 2)
x = SpatialCoordinate(m)
expr = outer(x, x)
v = Function(V).interpolate(expr)
```
So the kernels are massive which causes kernel scheduling to take ages (credit to you for explanation - recorded here for posterity). 

Now just have to work out how to fix it. I presume this is something to do with the way `dual_evaluation` has been implemented in https://github.com/FInAT/FInAT/pull/57 at `finat/finiteelementbase.py`.

wence-:
> Now just have to work out how to fix it. I presume this is something to do with the way dual_evaluation has been implemented in FInAT/FInAT#57 at finat/finiteelementbase.py.

I have described in that PR a bunch of things that look suboptimal to me (in terms of using ListTensor when ComponentTensor would do, and unrolling all components from VectorFunctionSpace rather than using Delta). I think those things need to be fixed.

ReubenHill:
From Kaushik on Slack (for posterity):

> You could try re-installing islpy, as the new version is shipped with a better performing bignum package. See https://github.com/inducer/islpy/issues/28 for context. Other than that what might help is making check_bounds an optional loopy check and turn it off in firedrake when PYOP2_DEBUG is false .

ReubenHill:
Superseded by #2115

PR #1744, opened 2020-06-18
Cache aux coordinate functions in mesh

tkarna:
Currently each `File` object stores its own coordinate function in case the output function space differs from the `mesh.coordinates` function space. This is inefficient if user creates multiple `File` objects that have the same output function space.

In this PR the aux coordinate functions are stored in `MeshGeometry` instead.

Addresses #1763. Sits on top of #1732.

### DIFF
--- a/firedrake/mesh.py
+++ b/firedrake/mesh.py
@@ -1042,8 +1042,9 @@ class MeshGeometry(ufl.Mesh, MeshGeometryMixin):
         self._coordinates = coordinates
 
         # Cached coordinate functions in different function spaces
-        # dict: FunctionSpace -> (coordinate Function, interpolator)
-        self.aux_coordinate_functions = {}
+        # In practice we store the interpolator object which contains the func
+        # dict: UFL element -> coordinate interpolator
+        self.aux_coord_interpolators = weakref.WeakValueDictionary()
 
     def init(self):
         """Finish the initialisation of the mesh.  Most of the time
@@ -1100,7 +1101,7 @@ values from f.)"""
 
         Mesh coordinates are re-interpolated on aux functions.
         """
-        for _, interpolator in self.aux_coordinate_functions.values():
+        for interpolator in self.aux_coord_interpolators.values():
             interpolator.interpolate()
 
     @utils.cached_property
--- a/firedrake/mesh.py
+++ b/firedrake/mesh.py
@@ -1095,6 +1095,38 @@ values from f.)"""
 
         raise AttributeError(message)
 
+    def get_aux_coordinates(self, ufl_element):
+        """
+        Returns cached coordinate function in a different function space.
+
+        :arg ufl_element: UFL Element defining the target function space
+        :returns: (coordinate function, interpolator) tuple
+        """
+        from firedrake import function, functionspace, interpolation
+        if isinstance(ufl_element, ufl.VectorElement):
+            ufl_element = ufl_element.sub_elements()[0]
+        # check if ufl_element matches the mesh coordinates
+        coordinates_fs = self.coordinates.function_space()
+        coord_elem = coordinates_fs.ufl_element().sub_elements()[0]
+        if ufl_element == coord_elem:
+            # mesh coordinates are compatible, no need for caching
+            return self.coordinates, None
+        # look in cache
+        interpolator = self.aux_coord_interpolators.get(ufl_element)
+        if interpolator is None:
+            # allocate new function and interpolator
+            gdim = self.ufl_cell().geometric_dimension()
+            V = functionspace.VectorFunctionSpace(self, ufl_element, dim=gdim)
+            print(f'Creating coords for V={V}')
+            name = self.coordinates.name()
+            aux_coordinates = function.Function(V, name=name)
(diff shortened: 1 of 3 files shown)


tkarna:
Here's a simple test case that demonstrates the functionality
```python
from firedrake import *

mesh2d = UnitSquareMesh(10, 10)
elt = FiniteElement("DG", mesh2d.ufl_cell(), 1)
fs = FunctionSpace(mesh2d, elt)

f = Function(fs).assign(4.5)
g = Function(fs).assign(3.3)

out1 = File('function1.pvd')
out2 = File('function2.pvd')

out1.write(f)
out2.write(g)

# deform mesh coordinates
mesh2d.coordinates.dat.data[:, 0] += 0.5*mesh2d.coordinates.dat.data[:, 1]
# user must manually trigger re-evaluation of cached coord functions
mesh2d.update_aux_coordinates()

out1.write(f)
out2.write(g)
```

firedrake/mesh.py
```diff
@@ -1036,6 +1041,10 @@ def __init__(self, coordinates):
 
         self._coordinates = coordinates
 
+        # Cached coordinate functions in different function spaces
+        # dict: FunctionSpace -> (coordinate Function, interpolator)
+        self.aux_coordinate_functions = {}
```
tkarna, line 1046: It's unclear to me what is the best way to implement the cache. Here we are just storing a weakref to the Function and Interpolator created elsewhere. One option would be the implement a get method that handles this. But for garbage collection, we can only store weakrefs in `MeshGeometry`.

firedrake/mesh.py
```diff
@@ -1085,6 +1094,15 @@ def coordinates(self, value):
...
         raise AttributeError(message)
 
+    def update_aux_coordinates(self):
+        """
+        Updates all cached aux-coordinate functions.
+
+        Mesh coordinates are re-interpolated on aux functions.
+        """
+        for _, interpolator in self.aux_coordinate_functions.values():
+            interpolator.interpolate()
```
tkarna, line 1104: It's also unclear to me what is the best API to invalidate the cached coordinates. One option would be to store a dirty bool flag for each aux coordinate, which would trigger interpolation next time the function is accessed.

firedrake/output.py
```diff
@@ -464,6 +470,37 @@ def _prepare_output(self, function, max_elem):
 
         return OFunction(array=get_array(output), name=name, function=output)
 
+    def _prepare_coordinates(self, mesh, max_elem):
```
tkarna, line 473: This should be refactored once the mesh api is fixed

firedrake/output.py
```diff
@@ -464,6 +470,37 @@ def _prepare_output(self, function, max_elem):
...
+                print(f'Allocating coords for V={V}')
+                output = Function(V)
+                interpolator = Interpolator(function, output)
+                # store weak reference in the mesh
+                # need to store the interpolator object too
+                mesh.aux_coordinate_functions[V] = (weakref.proxy(output), weakref.proxy(interpolator))
+                interpolator.interpolate()
+            self._output_functions[function] = output
+            # FIXME interpolator is not used in File but we need to keep a strong reference in the File object?
+            self._mappers[function] = interpolator
```
tkarna, line 501: Presumably we need to store a reference of interpolator in File as well?

firedrake/mesh.py
wence-, line 1046: Can this map from ufl Element to values, rather than functionspace? Also, can it be a `weakref.WeakValueDictionary` to the `interpolator`. There must be (or should be) a way to get the target function out of an interpolator object. This way we don't introduce ref-cycles.

wence-: Oh, I see that the user makes the weakrefs.

firedrake/output.py
wence-: yes.

firedrake/mesh.py
tkarna: So presumably it should be the "parent" ufl Element, i.e. `fs.ufl_element().sub_elements()[0]` in case `fs` is a vector function space?

tkarna: I've updated this now, storing the interpolator only

firedrake/output.py
```diff
@@ -464,6 +470,39 @@ def _prepare_output(self, function, max_elem):
...
+            if interpolator is None:
+                # allocate a new coordinate function
+                print(f'Allocating coords for V={V}')
+                output = Function(V)
+                interpolator = Interpolator(function, output)
+                # store weak reference in the mesh
+                # need to store the interpolator object too
+                mesh.aux_coord_interpolators[ufl_elem] = interpolator
+                interpolator.interpolate()
+            output = interpolator.V  # HACK I know that V is the output function, not the space ...
```
tkarna, line 500: I think this is currently the only way to get the target from `Interpolator`. 
Would be cleaner if the target space and function were separate properties in `Interpolator`.

tkarna:
I have now added `mesh.get_aux_coordinates(ufl_element)` method which handles the allocation and caching under the hood. The user just needs to keep a reference to the interpolator object somewhere.

PR #1751, opened 2020-06-23
Rckirby/install irksome

rckirby:
Allows firedrake-install to install Irksome.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -26,7 +26,7 @@ pipeline {
         sh 'mkdir tmp'
         dir('tmp') {
           timestamps {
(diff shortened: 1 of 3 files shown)


dham:
The merge commit passed, which is the important one. The one thing which is missing is that you should put `--install irksome` into the jenkinsfile install line so that Irksome installation is tested.

wence-:
> The merge commit passed, which is the important one. The one thing which is missing is that you should put --install irksome into the jenkinsfile install line so that Irksome installation is tested.

Just did that.

wence-:
> Just did that.

There was an error in the URL specification which I think I have fixed.

wence-:
Thanks!

Issue #1715, opened 2020-06-02
slate kernels on extruded meshes encode mesh layers


wence-:
This is bad. Moreover, the in memory cache key doesn't include it, so we get the wrong answer.

sv2518:
Ok, I have just had a look. The core of the problem is here https://github.com/firedrakeproject/firedrake/blob/ddba1d21ca37f2c9570998a586e6f42a644b83da/firedrake/slate/slac/compiler.py#L487. We want avoid that, so we do not have to reproduce the Slate code as soon as the users decides to change the amount of layers, is that right?
To fix this we need to think about what we are going to do here https://github.com/firedrakeproject/firedrake/blob/ddba1d21ca37f2c9570998a586e6f42a644b83da/firedrake/slate/slac/compiler.py#L494. Is the idea to pass another argument to the Slate kernel from PyOP2, which contains what is currently in the `num_layers`?

wence-:
> Is the idea to pass another argument to the Slate kernel from PyOP2, which contains what is currently in the num_layers?

Yes, I think so. We currently have `pass_layer_arg` as an option on par_loops (really we should migrate that to kernels and call it `pass_current_layer`), if we also add `pass_num_layers` then we can inspect that in codegen and generate the extra argument. I propose we don't do it in current slate compiler, but in the new one.

sv2518:
Oh ok, I have had a look at the current Slate compiler and have a propose for a fix for that, it's here: https://github.com/firedrakeproject/firedrake/compare/fix-variable-layer-count.
I pass the layer number as a Global arg in assemble. The Slate kernel has an additional parameter carrying the layer number. It passes at least all tests in tests/slate/test_facet_tensors_extr.py.
The kwarg `pass_layer_arg` results in layer passed in as int, I don't think we can inspect that on slate level. I think, we need the number of the layers as the separate argument to the slate kernel.

wence-:
Thanks!

PR #1754, opened 2020-06-23
Extrusion: Pass number of layers as argument to slate kernel.

sv2518:
Addresses #1715 to fix variable layers in slate code generation. What do you think about this?

### DIFF
--- a/firedrake/assemble.py
+++ b/firedrake/assemble.py
@@ -613,6 +613,10 @@ def create_parloops(expr, create_op2arg, *, assembly_rank=None, diagonal=False,
         if needs_cell_facets:
             assert integral_type == "cell"
             extra_args.append(m.cell_to_facets(op2.READ))
+        if pass_layer_arg:
+            c = op2.Global(1, itspace.layers-2, dtype=numpy.dtype(numpy.int32))
+            o = c(op2.READ)
+            extra_args.append(o)
 
         args.extend(extra_args)
         kwargs["pass_layer_arg"] = pass_layer_arg
--- a/firedrake/slate/slac/compiler.py
+++ b/firedrake/slate/slac/compiler.py
@@ -217,8 +217,12 @@ def generate_kernel_ast(builder, statements, declared_temps):
                              qualifiers=["const"]))
 
     # NOTE: We need to be careful about the ordering here. Mesh layers are
-    # added as the final argument to the kernel.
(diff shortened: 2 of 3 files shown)


firedrake/assemble.py
```diff
@@ -537,6 +538,10 @@ def get_map(x):
         if needs_cell_facets:
             assert integral_type == "cell"
             extra_args.append(m.cell_to_facets(op2.READ))
+        if pass_layer_arg:
+            import numpy
+            o = base.Arg(op2.Global(1, itspace.layers-2, dtype=numpy.dtype(numpy.int32)))
```
sv2518, line 543: I was not 100 percent sure, if this is the right place to generate the argument for the amount of layers?

wence-: That looks reasonable. You can  just do:

```
c = op2.Global(1, itspace.layers-2, dtype=...)
o = c(op2.READ)
```
I think.

wence-: This is a simpler fix than I was thinking of (which was to augment kernels saying that they needed this and then wrap it all up in PyOP2).

I think this approach is fine.

firedrake/slate/slac/compiler.py
```diff
@@ -483,15 +485,15 @@ def tensor_assembly_calls(builder):
         # The mesh level is an integer provided as a macro kernel
         # argument.
 
-        # FIXME: No variable layers assumption
```
wence-, line 486: This FIXME still applies (variable layers is different numbers of layers for each entry in the iteration set).

wence-:
Mostly fine. I suspect you need to rebase on top of (or just port the patch) current master, since the merge to refactor `assemble.py` has gone in.

firedrake/assemble.py
sv2518: Your suggestion resulted in a c compiler error, so I just left it as it was.

wence-: WAH? I don't understand, not least because your version doesn't set an access descriptor for the argument.

wence-: Which I think should not be allowed, although I see we don't actually check.

sv2518: Ah. I know why, I had not specified a const qualifier in the COFFEE argument declaration.

Issue #1758, opened 2020-06-24
Incorrect adjoint replay for project where output is also an input that is uninitialised


stephankramer:
In the following code, the replay is incorrect
```
from firedrake import *
from firedrake_adjoint import *

mesh = UnitSquareMesh(1,1)
V = FunctionSpace(mesh, "CG", 1)
u = Function(V)
c = Constant(1.)
u.project(u+c)
J = assemble(u**2*dx)
rf = ReducedFunctional(J, Control(c))
print(J)
print(rf(Constant(1.)))
```
The problem occurs at the project, where a ProjectBlock is created with a dependency on `u` that has not been checkpointed yet: in pyadjoint functions do not get a checkpoint of 0 associated with them upon creation, rather its checkpoint gets created on the fly at the first point the function is actually used, i.e. when it's registered as a dependency of a block. This checkpoint never gets overwritten in the replay. So far that's how it's supposed to work. The problem comes that the registration of the dependency in ProjectBlock happens *after* the projection is performed at which stage the incoming value has already been clobbered - compare with AssignBlock where the order is 1) create AssignBlock 2) perform the assign 3) add_output to AssignBlock. Unfortunately in project this is a little more tricky to untangle....ah no actually, we do know what the output is beforehand, so there's no problem in creating the block before the project.

PR #1759, opened 2020-06-25
Fix annotation of uninitialised self-dependent project

stephankramer:
Fix for #1758. Also fixes the annotation of the `project` _function_ (which was more generally broken). Tests in https://github.com/dolfin-adjoint/pyadjoint/tree/test-firedrake-project


stephankramer:
Thanks @dham. Fixed now

PR #1766, opened 2020-07-02
Annotate interpolation operations 

mtusman:
This change implements annotation for interpolation operations.


firedrake/adjoint/blocks.py
```diff
@@ -247,3 +250,59 @@ def recompute_component(self, inputs, block_variable, idx, prepared):
...
+    def _replace_map(self):
+        # Replace the dependencies with checkpointed values
+        replace_map = {}
+        args = 0
+        for block_variable in self.get_dependencies():
+            output = block_variable.output
+            if output in extract_coefficients(self.expr):
+                replace_map[output] = block_variable.saved_output
+            else:
+                replace_map[extract_arguments(self.expr)[args]] = block_variable.saved_output
```
stephankramer, line 280: same here

firedrake/adjoint/interpolate.py
```diff
@@ -0,0 +1,21 @@
...
+
+
+def annotate_interpolate(interpolate):
+    def wrapper(interpolator, *function, **kwargs):
+        annotate = annotate_tape(kwargs)
+        with stop_annotating():
+            output = interpolate(interpolator, *function, **kwargs)
+
+        if annotate:
+            sb_kwargs = InterpolateBlock.pop_kwargs(kwargs)
```
stephankramer, line 12: This looks wrong: if there are any kwargs that need popping (there probably aren't at the moment), that should be done before we go into `interpolate`

firedrake/adjoint/interpolate.py
```diff
@@ -0,0 +1,21 @@
...
+            sb_kwargs.update(kwargs)
+            block = InterpolateBlock(interpolator, *function, **sb_kwargs)
```
stephankramer, line 14: I believe the `InterpolateBlock` should be created before the actual `interpolate` call. This is a little fragile thing in pyadjoint, but it goes wrong when you have a dependency that is created by something that's not annotated. In that case the dependency, say a Function, does not have a checkpoint yet (as it's not the output of something), so we create a checkpoint when it's added as a dependency. Now where it goes wrong is that the dependency might also be the output - so you're interpolating an expression that depends on a Function into that same Function, which is not uncommon. In that case if you call interpolate first, the (input) value of the dependency that should be checkpointed is already clobbered with the output.

firedrake/utility_meshes.py
```diff
@@ -934,6 +936,9 @@ def UnitIcosahedralSphereMesh(refinement_level=0, degree=1, reorder=None,
                                  degree=degree, reorder=reorder, comm=comm)
 
 
+# mesh is mainly used as a utility, so it's unnecessary to annotate the construction
+# in this case.
+@no_annotations
```
stephankramer, line 941: There's a bunch more utility meshes that have interpolates in them, do they need the same?

stephankramer:
@dham: This brings up the general question about where these tests should live. This branch adds one adjoint test to firedrake, which makes sense from the fact that this is where the code that is tested, but traditionally all firedrake adjoint tests have been in pyadjoint. The latter has the advantage that it takes care of cleaning the tape between tests.

firedrake/adjoint/interpolate.py
dham: Ahhhh! This might be the cause of the mesh bug!

firedrake/utility_meshes.py
mtusman: I added `no_annotations` to the `OctahedralSphereMesh` in particular because I was failing one test in `tests/firedrake_adjoint/test_shape_derivatives.py::test_PDE_hessian_neumann`. I wasn't able to debug what was causing the error and since annotating the construction of the mesh is not necessary, I decided to turn annotation off and it lead to the test passing. To allow me to make progress in my project, David and I decided we can use `no_annotations` as a temporary measure to allow me to move on. I can do the same to other utility meshes to make it more consistent if you want?

firedrake/adjoint/interpolate.py
mtusman: After making the changes, the mesh bug is still there.

PR #1773, opened 2020-07-08
Adding ASMStarPC

ScottMacLachlan:
This subclasses `ASMPatchPC` to implement a patch-based additive Schwarz PC using the topological star of a given mesh entity.  This is primarily intended for use as a relaxation scheme within multigrid, so tests are provided to make sure that it does so correctly, comparing against `jacobi` in the case where the star is just a single DoF, and against `PatchPC` for more complicated examples.

This also includes a couple of fixes to aspects of `ASMPatchPC` that didn't work for me: changing the options prefix to append an underscore instead of prepend one, and to make the `view` method work.

### DIFF
--- a/firedrake/preconditioners/asm.py
+++ b/firedrake/preconditioners/asm.py
@@ -6,7 +6,7 @@ from firedrake.petsc import PETSc
 from firedrake.dmhooks import get_function_space
 import numpy
 
-__all__ = ("ASMPatchPC", "ASMStarPC", "ASMLinesmoothPC")
+__all__ = ("ASMPatchPC", "ASMStarPC", "ASMVankaPC", "ASMLinesmoothPC")
 
 
 class ASMPatchPC(PCBase):
@@ -121,6 +121,73 @@ class ASMStarPC(ASMPatchPC):
             for (i, W) in enumerate(V):
                 section = W.dm.getDefaultSection()
                 for p in pt_array.tolist():
+                    dof = section.getDof(p)
+                    if dof <= 0:
+                        continue
+                    off = section.getOffset(p)
+                    # Local indices within W
+                    W_indices = numpy.arange(off*W.value_size, W.value_size * (off + dof), dtype='int32')
+                    indices.extend(V_local_ises_indices[i][W_indices])
+            # Map local indices into global indices and create the IS for PCASM
+            global_indices = lgmap.apply(indices)
+            iset = PETSc.IS().createGeneral(global_indices, comm=COMM_SELF)
+            ises.append(iset)
+
+        return ises
+
+
+class ASMVankaPC(ASMPatchPC):
+    '''Patch-based PC using closure of star of mesh entities implmented as an
+    :class:`ASMPatchPC`.
+
+    ASMVankaPC is an additive Schwarz preconditioner where each patch
+    consists of all DoFs on the closure of the star of the mesh entity
+    specified by `pc_vanka_construct_dim` (or codim).
+    '''
+
+    _prefix = "pc_vanka_"
+
+    def get_patches(self, V):
+        mesh = V._mesh
+        mesh_dm = mesh._topology_dm
+        lgmap = V.dof_dset.lgmap
+
+        # Obtain the topological entities to use to construct the stars
+        depth = PETSc.Options().getInt(self.prefix + "construct_dim", default=-1)
+        height = PETSc.Options().getInt(self.prefix + "construct_codim", default=-1)
+        if (depth == -1 and height == -1) or (depth != -1 and height != -1):
+            raise ValueError(f"Must set exactly one of {self.prefix + 'construct_dim'} or {self.prefix + 'construct_codim'}!")
+
+        # Accessing .indices causes the allocation of a global array,
+        # so we need to cache these for efficiency
+        V_local_ises_indices = []
+        for (i, W) in enumerate(V):
+            V_local_ises_indices.append(V.dof_dset.local_ises[i].indices)
+
+        # Build index sets for the patches
+        ises = []
+        if depth != -1:
+            (start, end) = mesh_dm.getDepthStratum(depth)
+        else:
+            (start, end) = mesh_dm.getHeightStratum(height)
+
+        for seed in range(start, end):
+            # Only build patches over owned DoFs
+            if mesh_dm.getLabelValue("pyop2_ghost", seed) != -1:
+                continue
+
+            # Create point list from mesh DM
+            star, _ = mesh_dm.getTransitiveClosure(seed, useCone=False)
+            pt_array = set()
+            for pt in star.tolist():
+                closure, _ = mesh_dm.getTransitiveClosure(seed, useCone=True)
+                pt_array.update(closure.tolist())
+
+            # Get DoF indices for patch
+            indices = []
+            for (i, W) in enumerate(V):
+                section = W.dm.getDefaultSection()
+                for p in pt_array:
                     dof = section.getDof(p)
                     if dof <= 0:
                         continue
--- a/firedrake/preconditioners/asm.py
+++ b/firedrake/preconditioners/asm.py
@@ -156,7 +156,7 @@ class ASMVankaPC(ASMPatchPC):
         depth = PETSc.Options().getInt(self.prefix + "construct_dim", default=-1)
         height = PETSc.Options().getInt(self.prefix + "construct_codim", default=-1)
         if (depth == -1 and height == -1) or (depth != -1 and height != -1):
-            raise ValueError(f"Must set exactly one of {self.prefix + 'construct_dim'} or {self.prefix + 'construct_codim'}!")
+            raise ValueError(f"Must set exactly one of {self.prefix}construct_dim or {self.prefix}construct_codim")
 
         # Accessing .indices causes the allocation of a global array,
         # so we need to cache these for efficiency
--- a/firedrake/preconditioners/asm.py
+++ b/firedrake/preconditioners/asm.py
@@ -6,7 +6,7 @@ from firedrake.petsc import PETSc
 from firedrake.dmhooks import get_function_space
 import numpy
 
-__all__ = ("ASMPatchPC", "ASMLinesmoothPC")
+__all__ = ("ASMPatchPC", "ASMStarPC", "ASMLinesmoothPC")
 
 
 class ASMPatchPC(PCBase):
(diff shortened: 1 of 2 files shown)


firedrake/preconditioners/asm.py
```diff
@@ -66,6 +66,89 @@ def applyTranspose(self, pc, x, y):
...
+        # Create new PC object as ASM type and set index sets for patches
+        asmpc = PETSc.PC().create(comm=pc.comm)
+        asmpc.incrementTabLevel(1, parent=pc)
+        asmpc.setOptionsPrefix(self.prefix + "sub_")
+        asmpc.setOperators(*pc.getOperators())
+        asmpc.setType(asmpc.Type.ASM)
+        asmpc.setASMLocalSubdomains(len(ises), ises)
+
+        # Set default solver parameters
+        opts = PETSc.Options(asmpc.getOptionsPrefix())
```
pefarrell, line 102: Shouldn't these be the default for all `ASMXPC` implementations? I propose moving these to the base class and not subclassing `initialise` in `ASMStarPC`.

ScottMacLachlan: I'm not sure how relevant these are to `ASMLinesmoothPC`.  If the patch submatrix is diagonal or tridiagonal, then ilu == lu, and you wouldn't see the difference.  But for stars, you really notice that ILU(0) is a bad thing to do in general.

Likewise, setting `pc_asm_type` to `basic` doesn't have much of an effect in many cases, but is important sometimes (e.g., I see very bad results without it for Nedelec elements in parallel in 3D).

TLDR: probably we should check if that would change anything for `ASMLinesmoothPC`.  @colinjcotter ?

wence-: Alternately, if they shouldn't be default, prefer to not replicate the full `initialize` code, but instead advertise a `postSetUp` function or something that subclasses can override.

Also, rather than poking the options database, it's better to do this programmatically.

After doing `asmpc.setType(ASM)` you can do:

```python
ksp = asmpc.getASMSubKSP()
ksp.pc.setType(PETSc.PC.Type.LU)
ksp.pc.setFactorShift(shift_type=PETSc.Mat.FactorShiftType.NONZERO)
```

ScottMacLachlan: Okay.  I'll wait to consider any effects on `ASMLinesmoothPC` before pushing changes, but will change to the programmatic style either in the base class, or as a `postSetUp` function.

wence-: Let's bikeshed the name a bit. Maybe `set_defaults(self, asmpc)` ?

ScottMacLachlan: Is `set_defaults` too generic?  I guess that the context is easily inferred by the class structure, but `set_asm_defaults` seems rather less cryptic to me...

wence-: am happy with `set_asm_defaults` too.

pefarrell:
I tried this out on a problem where star via PCPATCH works well. (I can't easily share the code, because it depends on lots of branches.) Unfortunately something in `ASMStarPC` is scaling very badly with problem size. I timed the time spent in `get_patches` with `time.time`:

mesh size N = 10: time in `get_patches`: 2.58 s
mesh size N = 20: time in `get_patches`: 36.9 s

We should figure out what's going on here before merging.

wence-:
Can you run with cProfile? or better, [line profiler](https://github.com/pyutils/line_profiler)

firedrake/preconditioners/asm.py
```diff
@@ -66,6 +66,89 @@ def applyTranspose(self, pc, x, y):
...
+            for (i, W) in enumerate(V):
+                section = W.dm.getDefaultSection()
+                for p in pt_array.tolist():
+                    dof = section.getDof(p)
+                    if dof <= 0:
+                        continue
+                    off = section.getOffset(p)
+                    # Local indices within W
+                    W_indices = numpy.arange(off*W.value_size, W.value_size * (off + dof), dtype='int32')
+                    indices.extend(V.dof_dset.local_ises[i].indices[W_indices])
```
wence-, line 143:
```suggestion
        (start, end) = mesh_dm.getDepthStratum(depth)
        local_indices = tuple(W.dof_dset.local_ises.indices for W in V)
        for seed in range(start, end):
            # Only build patches over owned DoFs
            if mesh_dm.getLabelValue("pyop2_ghost", seed) != -1:
                continue

            # Create point list from mesh DM
            pt_array, _ = mesh_dm.getTransitiveClosure(seed, useCone=False)

            # Get DoF indices for patch
            indices = []
            for (i, W) in enumerate(V):
                section = W.dm.getDefaultSection()
                for p in pt_array.tolist():
                    dof = section.getDof(p)
                    if dof <= 0:
                        continue
                    off = section.getOffset(p)
                    # Local indices within W
                    W_indices = numpy.arange(off*W.value_size, W.value_size * (off + dof), dtype='int32')
                    indices.extend(local_indices[i][W_indices])
```

firedrake/preconditioners/asm.py
colinjcotter: re: defaults. I don't know how PETSc deals with this, but is each KSP local to a process domain usually? Just thinking that the default LU is not parallel.

wence-: The KSPs are all local.

ScottMacLachlan: The latest push removes the duplication, and just puts all the options stuff back in `ASMPatchPC`.  I _think_ that's reasonable from this conversation, but am open to adding `set_asm_defaults` if that's preferred.

I was foiled, however, in trying to do this programmatically.  I've left Lawrence's code suggestion stubbed in here in the comments, but I get a catastrophic (and uninformative) fail when I uncomment Line 53 of `asm.py`, and I haven't been successful in uncovering why.  Any ideas?

ScottMacLachlan:
@pefarrell : Great to have a Vanka implementation, too.  Can we easily (or not?) extend that to allow an exclude subspace on the points in the closure but not in the star?  That would allow the "right" combination for Taylor-Hood Stokes, in particular.

ScottMacLachlan:
I'm not sure what's best here for a default for the `pc_factor_shift_type` for the LU `sub_pc`.  For my tests (and my use cases), we tend to having `mat_type = aij`, and using the nonzero shift works best (e.g., is necessary for Stokes).  In `test_linesmoother_vfs.py`, though the `mat_type` defaults to `baij`, and that shift type isn't supported.  For now, I've set the default to `none`, and then pass the `nonzero` option in for the Stokes example.  The syntax for that is a bit ugly, but it works.  Thoughts?

wence-:
> I'm not sure what's best here for a default for the pc_factor_shift_type for the LU sub_pc. For my tests (and my use cases), we tend to having mat_type = aij, and using the nonzero shift works best (e.g., is necessary for Stokes). In test_linesmoother_vfs.py, though the mat_type defaults to baij, and that shift type isn't supported. For now, I've set the default to none, and then pass the nonzero option in for the Stokes example. The syntax for that is a bit ugly, but it works. Thoughts?

We could inspect that mattype and set things appropriately.

BTW, something went really screwy with the backmerges in this branch. I have pushed `smaclachlan/asmstarpc-fixed-merges`. If you confirm that that has all the code this branch is meant to introduce, I'll overwrite this branch, we can run tests again and then merge.

ScottMacLachlan:
What about just exposing a parameter to set `pc_factor_shift_type`, but not explicitly setting it by default?  That would allow us to set it as desired for the cases where we want/need to change the default behaviour, but wouldn't introduce complexities of having to choose this for the user who doesn't need/want to.  In particular, *not* setting this by default would resolve the inherently different needs for the tests in `test_linesmoother_vfs.py` and `test_star_pc.py`.  Thoughts on this?

Note that I say "exposing a parameter" above on purpose.  The parameter is already there, but having to set `pc_star_sub_sub_pc_factor_shift_type` seems a bit ridiculous to me (particularly when I prepend `mg_levels_`).  But, we could introduce, say, `pc_*_sub_factor_shift_type` and catch for it in the `initialize` method like we do now.  How does that sound?

My apologies for the borked rebase - after updating my Firedrake master branch, I could no longer run code on the `smaclachlan/asmstarpc` branch (due to an updated dependency), and something (clearly) went wrong with the merges...  The diff that GitHub shows between master and `smaclachlan/asmstarpc-fixed-merges` matches what my local copy shows between master and `smaclachlan/asmstarpc`, so I think overwriting the original branch is a good plan.  Thanks!

firedrake/preconditioners/asm.py
```diff
@@ -66,6 +80,129 @@ def applyTranspose(self, pc, x, y):
...
+class ASMStarPC(ASMPatchPC):
+    '''Patch-based PC using Star of mesh entities implmented as an
+    :class:`ASMPatchPC`.
+
+    ASMStarPC is an additive Schwarz preconditioner where each patch
+    consists of all DoFs on the topological star of the mesh entity
+    specified by `pc_star_construct_dim`.
+    '''
+
+    _prefix = "pc_star_"
```
wence-, line 92: I think the convention we have is that you don't need `pc_` here. So for example if you had `star_` then you'd end up with:

```
-mg_levels_pc_type python
-mg_levels_pc_python_type ASMStarPC
-mg_levels_star_pc_type asm # this is default
-mg_levels_star_sub_pc_type lu
-mg_levels_star_sub_pc_factor_shift_type nonzero
```

The python preconditioner has prefix `-mg_levels_` and the inner preconditioner then has prefix `-mg_levels_star_`. I think that reduces the amount of typing.

firedrake/preconditioners/asm.py
```diff
@@ -36,10 +36,24 @@ def initialize(self, pc):
...
         asmpc.setType(asmpc.Type.ASM)
         asmpc.setASMLocalSubdomains(len(ises), ises)
+
+        # Set default solver parameters
+        asmpc.setASMType(PETSc.PC.ASMType.BASIC)
+        opts = PETSc.Options(asmpc.getOptionsPrefix())
+        if "sub_pc_type" not in opts:
+            opts["sub_pc_type"] = "lu"
+        if "sub_pc_factor_shift_type" not in opts:
+            opts["sub_pc_factor_shift_type"] = "NONE"
```
wence-, line 50: Can simplify these to `opts.setdefault(key, defaultvalue)`

firedrake/preconditioners/asm.py
```diff
@@ -36,10 +36,24 @@ def initialize(self, pc):
         # Create new PC object as ASM type and set index sets for patches
         asmpc = PETSc.PC().create(comm=pc.comm)
         asmpc.incrementTabLevel(1, parent=pc)
-        asmpc.setOptionsPrefix(self.prefix + "_sub")
+        asmpc.setOptionsPrefix(self.prefix + "sub_")
```
wence-, line 39: I _think_ we want the asmpc to have prefix `self.prefix` (see below). But am willing to be persuaded otherwise. The way I figure it is that we configure this python pc with the options prefix from `pc.getOptionsPrefix()` and then the concatenation of that with `self._prefix` becomes the options prefix for the inner thing.

But maybe that doesn't quite work out if we want to do options lookup to configure the python thing. Hmmm....

firedrake/preconditioners/asm.py
```diff
@@ -66,6 +80,129 @@ def applyTranspose(self, pc, x, y):
...
+    def get_patches(self, V):
+        mesh = V._mesh
+        mesh_dm = mesh._topology_dm
+        lgmap = V.dof_dset.lgmap
+
+        # Obtain the topological entities to use to construct the stars
+        depth = PETSc.Options().getInt(self.prefix + "construct_dim", default=-1)
+        height = PETSc.Options().getInt(self.prefix + "construct_codim", default=-1)
+        if (depth == -1 and height == -1) or (depth != -1 and height != -1):
+            raise ValueError(f"Must set exactly one of {self.prefix + 'construct_dim'} or {self.prefix + 'construct_codim'}!")
```
wence-, line 159: style
```python
f"Must set exactly on of {self.prefix}construct_dim or {self.prefix}construct_codim"
```

pefarrell: Thanks, fixed.

firedrake/preconditioners/asm.py
pefarrell: Thanks, fixed.

ScottMacLachlan: This fix broke things, failing with `AttributeError: 'petsc4py.PETSc.Options' object has no attribute 'setdefault'`.  Looking at the `petsc4py` source, that seems both true and not just a matter of adjusting `setdefault`.  Should we revert or is there a better way?

wence-: Ah sorry, I had not twigged that opts here is just the Options database which isn't fully like a dict. Just revert.

firedrake/preconditioners/asm.py
wence-: Let us do this in another round consistently for all of the PCs.

wence- (dismissed):

wence- (approved):

firedrake/preconditioners/asm.py
wence-: Did that in #1830.

Issue #1768, opened 2020-07-02
Semantics of tsfc-produced kernels for interpolation don't match with codegen for MIN/MAX


wence-:
Consider

```python
from firedrake import *
import numpy
mesh = UnitSquareMesh(3, 3)
V = FunctionSpace(mesh, "DG", 0)
x, y = SpatialCoordinate(mesh)
g = interpolate(x*y, V)
f = interpolate(x**2 - y**4, V)
h = interpolate(max_value(x*y, x**2 - y**4), V)

# copy g
z = Function(g)

# This should mean z_i \gets max(z_i, f_i)
interpolate(f, z, access=MAX)

by_hand = numpy.where(f.dat.data_ro > g.dat.data_ro, f.dat.data_ro, g.dat.data_ro)
print("By hand:", numpy.allclose(by_hand, h.dat.data_ro))
print("Interpolate:", numpy.allclose(by_hand, z.dat.data_ro))

=>
By hand: True
Interpolate: False
```

If we look at the wrapper + kernel generated for the last expression, we can see the problem:

```c
static void expression_kernel(double *__restrict__ A, double const *__restrict__ w_0)
{
     A[0] = A[0] + w_0[0];
}
void wrap_expression_kernel(int32_t const start, int32_t const end, double *__restrict__ dat1, double const *__restrict__ dat0, int32_t const *__restrict__ map0)
{
  double t0[1];
  double t1[1];

  for (int32_t n = start; n <= -1 + end; ++n)
  {
    t1[0] = dat1[map0[n]];
    t0[0] = dat0[map0[n]];
    expression_kernel(&(t1[0]), &(t0[0]));
    dat1[map0[n]] = fmax(dat1[map0[n]], t1[0]);
  }
}
```

TSFC always generates kernels that _accumulate_ into the output argument (with the requirement that the caller zeros them beforehand). This is in part a legacy of at first only supporting assembly kernel generation, and so it doesn't have access descriptors for the output variable.

PyOP2 on the other hand, does know about access descriptors and builds the wrapper for MAX assuming that the kernel is going to reduce with MAX. It expected that you wrote the inner kernel as:

```c
static void expression_kernel(double *__restrict__ A, double const *__restrict__ w_0)
{
     A[0] = fmax(A[0], w_0[0]);
}
```

[...]

PR #1774, opened 2020-07-08
wence/fix/interpolate max

wence-:
Addresses #1768.

### DIFF
--- a/tests/regression/test_interpolate.py
+++ b/tests/regression/test_interpolate.py
@@ -345,3 +345,14 @@ def test_interpolate_minmax(access):
         expect = np.where(f.dat.data_ro > g.dat.data_ro, f.dat.data_ro, g.dat.data_ro)
 
     assert np.allclose(actual, expect)
+
+
+def test_interpolate_periodic_coords_max():
+    mesh = PeriodicUnitSquareMesh(4, 4)
+    V = VectorFunctionSpace(mesh, "P", 1)
+
+    continuous = interpolate(SpatialCoordinate(mesh), V, access=MAX)
+
+    # All nodes on the "seam" end up being 1, not 0.
+    assert np.allclose(np.unique(continuous.dat.data_ro),
+                       [0.25, 0.5, 0.75, 1])


wence-:
Needs OP2/PyOP2#590.

sv2518:
I think the initialisation is not quite right. We also need to initialise empty functions. With the following the test passes.
```
diff --git a/firedrake/interpolation.py b/firedrake/interpolation.py
index 97688a71..1bd73747 100644
--- a/firedrake/interpolation.py
+++ b/firedrake/interpolation.py
@@ -131,6 +131,8 @@ def make_interpolator(expr, V, subset, access):
             V = f.function_space()
         else:
             f = firedrake.Function(V)
+        if ((isinstance(V, firedrake.Function) and not hasattr(f._data.dat, "_numpy_data"))
+           or not isinstance(V, firedrake.Function)):
             if access in {firedrake.MIN, firedrake.MAX}:
                 finfo = numpy.finfo(f.dat.dtype)
                 if access == firedrake.MIN:
```

wence-:
> I think the initialisation is not quite right. We also need to initialise empty functions.

The behaviour was correct, but the test was wrong. The model is that if you interpolate into a function then that functions values should take part in the min/maxing. If you interpolate into a space, then they shouldn't. I have fixed the test.

wence-:
> The model is that if you interpolate into a function then that functions values should take part in the min/maxing.

Added this to the docstring.

sv2518:
Ah, ok. Thanks!

firedrake/mg/utils.py
```diff
@@ -140,6 +140,8 @@ def physical_node_locations(V):
         return cache[key]
     except KeyError:
         Vc = firedrake.FunctionSpace(mesh, ufl.VectorElement(element))
+        locations = firedrake.Function(Vc)
```
pefarrell, line 143: Maybe a dumb question: doesn't this create a variable, then immediately overwrite it?

wence-: Oh yeah!

pefarrell:
It would be nice to have a test where we take a periodic mesh (with DG coordinates), interpolate using `MAX` to a CG space, and check that the coordinates make sense (the choice of coordinates on the seam has been consistently taken from one side).

wence-:
> It would be nice to have a test where we take a periodic mesh (with DG coordinates), interpolate using MAX to a CG space, and check that the coordinates make sense (the choice of coordinates on the seam has been consistently taken from one side).

Did that, will merge.

firedrake/mg/utils.py
wence-: Fixed.

PR #1776, opened 2020-07-10
Add a "crossed" cutting of hexadra in BoxMesh

jingminxia:
Here is to cut a hexadra into 5 tetrahedra in a "crossed" way.

The default cutting of a BoxMesh in Firedrake is (6 tetrahedra)

default

and now we can cut into 5 tetrahedra (less biased) as shows in the following

crossed

Both pictures are taken from [https://www.ics.uci.edu/~eppstein/projects/tetra/](url).

### DIFF
--- a/firedrake/utility_meshes.py
+++ b/firedrake/utility_meshes.py
@@ -707,7 +707,7 @@ def UnitTetrahedronMesh(comm=COMM_WORLD):
     return mesh.Mesh(plex, reorder=False)
 
 
-def BoxMesh(nx, ny, nz, Lx, Ly, Lz, reorder=None, distribution_parameters=None, comm=COMM_WORLD):
+def BoxMesh(nx, ny, nz, Lx, Ly, Lz, reorder=None, distribution_parameters=None, diagonal="default", comm=COMM_WORLD):
     """Generate a mesh of a 3D box.
 
     :arg nx: The number of cells in the x direction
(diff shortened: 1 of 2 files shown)


firedrake/utility_meshes.py
```diff
@@ -716,6 +716,9 @@ def BoxMesh(nx, ny, nz, Lx, Ly, Lz, reorder=None, distribution_parameters=None,
     :arg Lx: The extent in the x direction
     :arg Ly: The extent in the y direction
     :arg Lz: The extent in the z direction
+    :arg diagonal: Two ways of cutting hexadra, should be cut into 6
+        tetrahedra (``"default"``), or 5 tetrahedra thus less biased
+        (``"crossed"``)
```
pefarrell, line 721: Perhaps add a comment with a link to pictures?

jingminxia: One can test using the following code:

from firedrake import *
#mesh = BoxMesh(1,1,1,1,1,1,diagonal="default")
mesh = BoxMesh(1,1,1,1,1,1,diagonal="crossed")
v = Function(FunctionSpace(mesh, "CG", 1))
File("/tmp/check_mesh.pvd").write(v)

Using the default cutting, one can see a mesh bias giving an ellipse like this

ellipse

If we change to use "crossed" cutting, we can get a less-biased circle like this

circle

pefarrell: Just to be clear: these are pictures of simulations on the two grids that are not reproduced by Jingmin's snippet (of course).

tests/regression/test_mesh_generation.py
```diff
@@ -421,3 +421,11 @@ def test_changing_default_reorder_works(reorder):
...
     finally:
         parameters["reorder_meshes"] = old_reorder
+
+
+@pytest.mark.parametrize("kind",
+                         ["default", "crossed"])
+def test_boxmesh_diagonal(kind):
+    m = BoxMesh(1, 1, 1, 1, 1, 1, diagonal=kind)
+    m.init()
+    assert m.num_cells() == 6 or 5
```
wence-, line 431: This logic doesn't do what you want. It evaluates as `(m.num_cells() == 6) or 5` and then `5` is interpreted in a boolean context, so this is always true.

You really want to assert that you got the correct number of cells too, so try:

```python
@pytest.mark.parametrize("kind,num_cells", [("default", 6), ("crossed", 5)])
def test_boxmesh_kind(kind, num_cells):
    ...
    assert m.num_cells() == num_cells
```

jingminxia: Oh, no. You are right! I just pushed the corrected testing in the latest commit. Thanks, Lawrence!

PR #1778, opened 2020-07-10
install: Temporary workaround for gfortran-10 and libsupermesh

wence-:
Gfortran has become pickier about argument matching in version 10, so
turn it off while we wait for a proper upstream fix.


wence-:
That fix got merged upstream, so closing.

PR #1782, opened 2020-07-16
TinyASM backend for ASMPatchPC

florianwechsung:
This adds a simple alternative backend to PETSc PCASM for the ASMPatchPC. It builds on #1773 

While PCASM is great for domain decomposition problems where you may want to use other sophisticated solvers on each domain, we are often interested in the limit of many, small patches. In that case, the fastest thing to do is often to just compute the explicit inverse and then apply it each iteration. (Instead of for example computing the LU factorisation and then doing two triangular solves each time). In addition, this avoids the overhead of having KSP and PC objects for every single patch.

For the tests in `test_star_pc.py` this results in a speedup between 2x and 4x.
```
diff --git a/tests/regression/test_star_pc.py b/tests/regression/test_star_pc.py
index ae6f0377..1c4f54dc 100644
--- a/tests/regression/test_star_pc.py
+++ b/tests/regression/test_star_pc.py
@@ -158,13 +158,18 @@ def test_star_equivalence(problem_type):

     nvproblem = NonlinearVariationalProblem(a, u, bcs=bcs)
     star_solver = NonlinearVariationalSolver(nvproblem, solver_parameters=star_params, nullspace=nsp)
+    import time
+    t0 = time.time()
     star_solver.solve()
+    t1 = time.time()
+    print('Time using petsc', t1-t0)
     star_its = star_solver.snes.getLinearSolveIterations()

     try:
         from tinyasm import _tinyasm as tasm  # noqa
         have_tinyasm = True
     except ImportError:
+        print('does not have tinyasm')
         have_tinyasm = False

     if have_tinyasm:
@@ -172,7 +177,10 @@ def test_star_equivalence(problem_type):
         u.assign(0)
         nvproblem = NonlinearVariationalProblem(a, u, bcs=bcs)
         tinyasm_solver = NonlinearVariationalSolver(nvproblem, solver_parameters=star_params, nullspace=nsp)
+        t0 = time.time()
         tinyasm_solver.solve()
+        t1 = time.time()
+        print('Time using TinyASM', t1-t0)
         tinyasm_its = tinyasm_solver.snes.getLinearSolveIterations()
     else:
         tinyasm_its = star_its
```
```
Time using petsc 0.6164448261260986
Time using TinyASM 0.13721990585327148
Time using petsc 2.2604024410247803
Time using TinyASM 0.9665257930755615
Time using petsc 0.5605008602142334
Time using TinyASM 0.13840794563293457
```

The TinyASM code currently lives in https://github.com/florianwechsung/tinyasm, but I'd be happy to move it over to the firedrake organisation. I added an option `--tinyasm` to the install script as well.

### DIFF
--- a/firedrake/preconditioners/asm.py
+++ b/firedrake/preconditioners/asm.py
@@ -45,16 +45,37 @@ class ASMPatchPC(PCBase):
         asmpc.incrementTabLevel(1, parent=pc)
         asmpc.setOptionsPrefix(self.prefix + "sub_")
         asmpc.setOperators(*pc.getOperators())
-        asmpc.setType(asmpc.Type.ASM)
-        asmpc.setASMLocalSubdomains(len(ises), ises)
 
-        # Set default solver parameters
-        asmpc.setASMType(PETSc.PC.ASMType.BASIC)
-        opts = PETSc.Options(asmpc.getOptionsPrefix())
-        if "sub_pc_type" not in opts:
-            opts["sub_pc_type"] = "lu"
-        if "sub_pc_factor_shift_type" not in opts:
-            opts["sub_pc_factor_shift_type"] = "NONE"
+        backend = PETSc.Options().getString(self.prefix + "backend",
+                                            default="petscasm").lower()
+        # Either use PETSc's ASM PC or use TinyASM (as simple ASM
+        # implementation designed to be fast for small block sizes).
+        if backend == "petscasm":
+            asmpc.setType(asmpc.Type.ASM)
+            # Set default solver parameters
+            asmpc.setASMType(PETSc.PC.ASMType.BASIC)
+            opts = PETSc.Options(asmpc.getOptionsPrefix())
+            if "sub_pc_type" not in opts:
+                opts["sub_pc_type"] = "lu"
+            if "sub_pc_factor_shift_type" not in opts:
+                opts["sub_pc_factor_shift_type"] = "NONE"
+            lgmap = V.dof_dset.lgmap
+            # Translate to global numbers
+            ises = tuple(lgmap.applyIS(iset) for iset in ises)
+            asmpc.setASMLocalSubdomains(len(ises), ises)
+        elif backend == "tinyasm":
+            if not have_tinyasm:
+                raise ValueError("To use the TinyASM backend you need to install firedrake with TinyASM (firedrake-update --tinyasm)")
+            asmpc.setType("tinyasm")
+            # TinyASM wants local numbers, no need to translate
+            tinyasm.SetASMLocalSubdomains(
+                asmpc, ises,
+                [W.dm.getDefaultSF() for W in V],
+                [W.value_size for W in V],
+                sum(W.value_size * W.dof_dset.total_size for W in V))
+            asmpc.setUp()
+        else:
+            raise ValueError(f"Unknown backend type f{backend}")
 
         asmpc.setFromOptions()
         self.asmpc = asmpc
@@ -65,7 +86,8 @@ class ASMPatchPC(PCBase):
 
         :param  V: the :class:`~.FunctionSpace`.
 
-        returns a list of index sets defining the ASM patches.
+        :returns: a list of index sets defining the ASM patches in local
+            numbering (before lgmap.apply has been called).
         '''
         pass
 
@@ -96,7 +118,6 @@ class ASMStarPC(ASMPatchPC):
     def get_patches(self, V):
         mesh = V._mesh
         mesh_dm = mesh._topology_dm
-        lgmap = V.dof_dset.lgmap
 
         # Obtain the topological entities to use to construct the stars
         depth = PETSc.Options().getInt(self.prefix+"construct_dim", default=0)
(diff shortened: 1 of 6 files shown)


florianwechsung:
This currently fails on jenkins with

```
[2020-07-16T17:48:13.545Z]     CMake Error at CMakeLists.txt:5 (project):
[2020-07-16T17:48:13.545Z]       The CMAKE_C_COMPILER:
[2020-07-16T17:48:13.545Z] 
[2020-07-16T17:48:13.545Z]         mpicc
[2020-07-16T17:48:13.545Z] 
[2020-07-16T17:48:13.545Z]       is not a full path and was not found in the PATH.
[2020-07-16T17:48:13.545Z] 
[2020-07-16T17:48:13.545Z]       Tell CMake where to find the compiler by setting either the environment
[2020-07-16T17:48:13.545Z]       variable "CC" or the CMake cache entry CMAKE_C_COMPILER to the full path to
[2020-07-16T17:48:13.545Z]       the compiler, or to the compiler name if it is in the PATH.

```

Does anyone have any advice? I thought there should be a `mpicc` and `mpicxx` in `firedrake/bin`

pefarrell:
Looking at the standard `Jenkinsfile`, it doesn't pass the `--mpicc` flags (and friends) to `firedrake-install`. Maybe if you add the MPICH packages to `docker/Dockerfile.env` and the flags to the installer script, it might work?

scripts/firedrake-install
```diff
@@ -922,6 +926,32 @@ def build_and_install_libsupermesh():
...
+    else:
+        git_clone("git+https://github.com/florianwechsung/TinyASM.git")
+        tinyasm_changed = True
+
+    if tinyasm_changed:
+        with directory("TinyASM"):
+            # Clean source directory
+            check_call(["git", "reset", "--hard"])
+            check_call(["git", "clean", "-f", "-x", "-d"])
+            run_pip_install(["-e", "."])
```
wence-, line 950: We set `MPICC` and similar in the environment that this is run in, but perhaps not with absolute paths?

wence-: The latter probably matters because I think we don't push the `PATH` on?

wence-:
> Looking at the standard Jenkinsfile, it doesn't pass the --mpicc flags (and friends) to firedrake-install. Maybe if you add the MPICH packages to docker/Dockerfile.env and the flags to the installer script, it might work?

That would probably work, but please don't do that, because that's not the way we recommend users on desktop install.

wence-:
In tinyasm's `cmakelists.txt` I see:

```cmake
set(CMAKE_CXX_COMPILER "mpicxx")
set(CMAKE_C_COMPILER "mpicc")
```

Can you default to the `MPICXX` and `MPICC` environment variables instead?

tests/regression/test_star_pc.py
```diff
@@ -162,7 +162,7 @@ def test_star_equivalence(problem_type):
     star_its = star_solver.snes.getLinearSolveIterations()
 
     try:
-        from tinyasm import _tinyasm as tasm
+        from tinyasm import _tinyasm as tasm  # noqa
```
wence-, line 165: You can be more specific in the warning you squash I think. Otherwise this is fine.

florianwechsung:
I updated `CMakeLists.txt` and `pr-merge` is building and passing fine on jenkins. I guess the last thing remaining is the memory leak discussed here https://github.com/florianwechsung/TinyASM/issues/1

tests/regression/test_linesmoother.py
```diff
@@ -38,7 +38,17 @@ def expected(mesh_type):
...
+def backend(request):
+    return request.param
+
+
+def test_linesmoother(mesh, S1family, expected, backend):
+    if backend == "tinyasm":
+        try:
+            from tinyasm import _tinyasm as tasm  # noqa: F401
+        except ImportError:
+            return
```
wence-, line 51: A cleaner solution:

```
try:
    import tinyasm
    marks = None
except ImportError:
    marks = pytest.mark.skip(reason="No tinyasm")

@pytest.fixture(params=["petscasm", pytest.param("tinyasm", marks=marks)])
 def backend(request):
     return request.param
```

firedrake/preconditioners/asm.py
```diff
@@ -66,6 +89,117 @@ def applyTranspose(self, pc, x, y):
...
+                    if dof <= 0:
+                        continue
+                    off = section.getOffset(p)
+                    # Local indices within W
+                    W_indices = numpy.arange(off*W.value_size, W.value_size * (off + dof), dtype='int32')
+                    indices.extend(V_local_ises_indices[i][W_indices])
+            iset = PETSc.IS().createGeneral(indices, comm=COMM_SELF)
+            ises.append(iset)
+
+        return ises
```
wence-, line 200: Does it make sense to replicate all of `ASMPatchPC`?

florianwechsung: This is building on top of Scott's PR, so the modification to ASMPatchPC is actually much smaller than this diff suggests. However, there was some duplication in the `if backend == 'petscasm`: ... else ... ` part, which I have now reduced.

ScottMacLachlan: I think there's some fundamental duplication that we haven't really resolved in my `ASMStarPC` PR (and that's mostly my fault for not getting back to it with any serious effort this week...), trying to see a way forward that doesn't negatively impact `ASMLinesmoothPC`.  I'll take a shot at that now.

tests/regression/test_linesmoother.py
```diff
@@ -38,18 +38,19 @@ def expected(mesh_type):
         return [9, 20]
 
 
-@pytest.fixture(params=["petscasm", "tinyasm"])
+try:
+    import tinyasm  # noqa: F401
+    marks = None
+except ImportError:
+    marks = pytest.mark.skip(reason="No tinyasm")
```
wence-, line 45: Sorry, I screwed up, this needs to be a set: `marks = {pytest.mark.skip(...)}`

wence-:
I guess this is still waiting on #1773 to go in first. I also suppose the branches that fixed memory leaks in tinyasm need to be merged?

florianwechsung:
Everything on the TinyASM side has now been merged

wence- (approved):

PR #1789, opened 2020-07-21
Tests for new integral version of N1div/curl N2div/curl

FabianL1908:


tests/regression/test_interpolation_nodes.py
```diff
@@ -0,0 +1,119 @@
...
+
+
+def test_convergence_order(mesh, function_space):
+    space, degree = function_space
+    nref = 2
+    nref_min = 1
+    error = compute_interpolation_error(mesh, nref, space, degree)
+    error_l2 = error.T[0]
+    error_hD = error.T[1]
+    conv_l2 = np.log(error_l2[0:-1]/error_l2[1:])/np.log(2)
```
pefarrell, line 111: `np.log2`?

FabianL1908: Okay, changed it.

tests/regression/test_interpolation_nodes.py
```diff
@@ -0,0 +1,119 @@
...
+    elif dim == 3:
+        if "Nedelec" in V.ufl_element().family():
+            expression = grad(sin(x)*exp(y)*z)
+        else:
+            expression = as_vector([sin(y)*z, cos(x)*z, exp(x)])
+
+    f = interpolate(expression, V)
+    if "Nedelec" in V.ufl_element().family():
+        if dim == 2:
+            # Skip this test
```
wence-, line 59: I presume this is because these are the same as the HDiv spaces? Maybe clarify.

tests/regression/test_interpolation_nodes.py
```diff
@@ -0,0 +1,119 @@
...
+    error_hD = error.T[1]
+    conv_l2 = np.log2(error_l2[0:-1]/error_l2[1:])
+    conv_hD = np.log2(error_hD[0:-1]/error_hD[1:])
+    if "1" in space:
+        conv_l2_expected = degree
+    else:
+        conv_l2_expected = degree + 1
+    conv_hD_expected = degree
+    assert np.allclose(conv_l2[nref_min:], conv_l2_expected*np.ones((nref-nref_min, 1)), atol=0.15)
+    assert np.allclose(conv_hD[nref_min:], conv_hD_expected*np.ones((nref-nref_min, 1)), atol=0.15)
```
wence-, line 119: Prefer something like `assert all(conv[nref_min:] > expected - eps)` ?

PR #1790, opened 2020-07-22
Adding nemesis as mpich device

ScottMacLachlan:
The default MPICH installed by PETSc gives poor parallel scaling in some instances.  This updates that default to the "nemesis" device that appears to give better performance.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -4,7 +4,7 @@ pipeline {
   environment {
     FIREDRAKE_CI_TESTS = "1"
     DOCKER_CREDENTIALS = credentials('f52ccab9-5250-4b17-9fb6-c3f1ebdcc986')
-    PETSC_CONFIGURE_OPTIONS = "--with-make-np=12"
+    PETSC_CONFIGURE_OPTIONS = "--with-make-np=12 --download-mpich-device=ch3:sock"
   }
   stages {
     stage('BuildAndTest') {
--- a/docs/source/download.rst
+++ b/docs/source/download.rst
@@ -33,6 +33,17 @@ use Firedrake::
 
      source firedrake/bin/activate.csh
 
+
+Installation and MPI
+--------------------
+
+By default, ``firedrake-install`` will prompt the PETSc installer to
+download and install its own MPICH library and executables in the
+virtual environment.  This has implications for the performance of the
+resulting library when run in parallel.  Instructions on how best to
+configure MPI for the installation process are `found here
+<https://www.firedrakeproject.org/parallelism.html>`_.
+
 Testing the installation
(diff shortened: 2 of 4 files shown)


wence-:
Theres been some more discussion about this on the PETSc list. nemesis doesn't perform well with oversubscription. Hence it not being the default. I think this is why the CI hung up (because we oversubscribe).

So we could default to `--download-mpich-device=nemesis` except on CI. Or add information to the install docs and make it optional with a flag?

pefarrell:
Since this seems so important for performance, I suggest it be the default, and the CI pass the optional flag. I think otherwise a lot of people might download firedrake, try it out, observe it's much slower for their problem than FEniCS, and then not bother any further.

ScottMacLachlan:
In our thread on #general, Matt says that PETSc will now switch to installing with nemesis as default.  See https://gitlab.com/petsc/petsc/-/merge_requests/2990 for the PR, which is a little more nuanced: ch3:sock will stay the default if you compile using `--with-debugging=1`, but nemesis will be used if you compile using `--with-debugging=0`.  Since we default to the latter, we'll pick up the change.

What's the best way to expose this, though, for both users and the CI?

One option would be to add a `petsc-debug-level` flag that defaults to 1.  That seems a bit obscure to me, even if the purpose is well-documented; also, it may have implications on how things run in CI vs outside, if we change the PETSc behaviour that much between a "usual" install and the CI install.  I wonder if we shouldn't still have our own flag that _just_ changes the device (and leaves the PETSc debugging flag untouched).  Thoughts?

dham:
Do we need to fix this at all? If nemesis is the default then that's good. A user can tell PETSc to be in debugging mode by passing `--with-debugging=1` through `PETSC_CONFIGURE_OPTIONS`. That's clearly an advanced thing to do and they get whatever PETSc does in that case. I'm not sure that adding our own skin on that helps anyone.

ScottMacLachlan:
We need to do _something_ here (once we update PETSc to anything from the current master onwards).  The minimal change is to use `--with-debugging=1` for CI.  I think the question now is if we're going to make the minimal change, or do something more.

I would argue that we need at least to document this, since some users will want to oversubscribe their systems and the minimal change would require them to proactively do something when installing firedrake to make that work.  The possible advantage of wrapping this on our side would just be to make that documentation more apparent to the non-expert user.

wence-:
How about we expand the "Running Firedrake in parallel" part of the documentation to have some details on configuring mpich? And link to that on the install page?

ScottMacLachlan:
I think this is a reasonable approach, and am happy to make a first pass at the documentation edits.

PR #1797, opened 2020-07-24
wence/fix/bc initialisation

wence-:
Simplifies the logic for getting bc values into the target space. Also catches a load of bugs around non-homogenous bc values for vector-valued spaces.


tests/slate/test_hdg_poisson.py
```diff
@@ -93,7 +93,6 @@ def run_LDG_H_problem(r, degree, quads=False):
     return scalar_error, flux_error
 
 
-@pytest.mark.parallel
```
dham, line 96: Why is this removed?

wence-: I don't recall, I'll drop it.

wence-: Possibly leftover from debugging the test.

PR #1802, opened 2020-08-11
Remove replacing of the coefficient the jacobian is built on

jandrej:
This fixes p-MG when used with nonlinear problems. Before, the coefficient the linearization depends on was replaced with an uninitialized function.

### DIFF
--- a/firedrake/preconditioners/pmg.py
+++ b/firedrake/preconditioners/pmg.py
@@ -202,6 +202,7 @@ class PMGPC(PCBase):
 
         cdm.setKSPComputeOperators(_SNESContext.compute_operators)
         cdm.setCreateInterpolation(self.create_interpolation)
+        cdm.setCreateInjection(self.create_injection)
         cdm.setOptionsPrefix(fdm.getOptionsPrefix())
 
         # If we're the coarsest grid of the p-hierarchy, don't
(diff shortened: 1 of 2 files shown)


jandrej:
I think the change set is now what we would expect in the nonlinear case. The title of the PR is not correct anymore though :)

pefarrell (approved):

firedrake/preconditioners/pmg.py
```diff
@@ -151,7 +153,13 @@ def coarsen(self, fdm, comm):
...
         cV = firedrake.FunctionSpace(fV.mesh(), cele)
         cdm = cV.dm
-        cu = firedrake.Function(cV)
+
+        if isinstance(fV.ufl_element(), MixedElement) and not isinstance(fV.ufl_element(), (VectorElement, TensorElement)):
+            cu = firedrake.Function(cV)
+            for (i, _) in enumerate(fV):
+                firedrake.interpolate(fu.split()[i], cu.split()[i])
+        else:
+            cu = firedrake.interpolate(fu, cV)
```
wence-, line 162: Unfortunately this is not sufficient for nonlinear problems where the state continues to change (because you only ever inject this state once). So I think if you do:

```python
cu = firedrake.Function(cV)
interpolators = tuple(Interpolator(fus, cus) for fus, cus in zip(fu.split(), cu.split())
def inject_state(interpolators):
    for interpolator in interpolators:
        interpolator.interpolate()
```

And then after you've set up the context (line 208) do:

```python
add_hook(parent, setup=partial(inject_state, interpolators), call_setup=True)
```

I think you don't need a teardown because you're not actually stashing anything on a coarse dm.

wence-: Aside, this issue is a problem for the general GMG code too (so if you do further h-multigrid, the coarsened things will only ever linearise around the initial state I fear).

wence- (changes requested):
Needs my proposed suggestion.

pefarrell (approved):
Looks good to me, let's wait for Jenkins.

wence- (approved):
Thanks

PR #1804, opened 2020-08-15
Implement efficient adjoint solve

mtusman:
Currently for the replay functionality of the `NonlinearVariationalSolveBlock`, we construct a new `NonlinearVariationalProblem` and a `NonlinearVariationalSolver` every time ` _forward_solve` is called. This change implements a more efficient approach that creates a clone of each NonlinearVariationSolver instance instead and continuously modifies it to replay the forward problem, reducing the overhead cost of the re-computation. To demonstrate improvements in the run time, we take the burgers equation (code: https://github.com/dolfin-adjoint/pyadjoint/blob/master/tests/firedrake_adjoint/test_burgers_newton.py) with a 10000 unit interval mesh and measure the average time it takes to solve the equation over 10 runs:
```
Annotation off: 7.73(s)
Annotation on: 11.64(s)
```
Then computing the average time it takes to replay the equation over the course of 10 runs, we get the following results:
```
Existing implementation: 62.76(s)
New implementation: 19.65(s)
```

### DIFF
--- a/firedrake/adjoint/variational_solver.py
+++ b/firedrake/adjoint/variational_solver.py
@@ -18,8 +18,8 @@ class NonlinearVariationalProblemMixin:
             self._ad_bcs = self.bcs
             self._ad_J = self.J
             self._ad_kwargs = {'Jp': self.Jp, 'form_compiler_parameters': self.form_compiler_parameters, 'is_linear': self.is_linear}
+            self._ad_count_map = {}
         return wrapper
-
 
 class NonlinearVariationalSolverMixin:
     @staticmethod
@@ -92,6 +92,7 @@ class NonlinearVariationalSolverMixin:
         F_coefficients = problem.F.coefficients()
         J_coefficients = problem.J.coefficients()
 
+        _ad_count_map = {}
         for block_variable in dependencies:
             coeff = block_variable.output
             if coeff in F_coefficients and coeff not in F_replace_map:
@@ -99,7 +100,7 @@ class NonlinearVariationalSolverMixin:
                     F_replace_map[coeff] = copy.deepcopy(coeff)
                 else:
                     F_replace_map[coeff] = coeff.copy(deepcopy=True)
-                F_replace_map[coeff]._ad_count = coeff.count()
+                _ad_count_map[F_replace_map[coeff]] = coeff.count()
 
             if coeff in J_coefficients and coeff not in J_replace_map:
                 if coeff in F_replace_map:
@@ -108,9 +109,11 @@ class NonlinearVariationalSolverMixin:
                     J_replace_map[coeff] = copy.deepcopy(coeff)
                 else:
                     J_replace_map[coeff] = coeff.copy()
-                J_replace_map[coeff]._ad_count = coeff.count()
+                _ad_count_map[J_replace_map[coeff]] = coeff.count()
 
-        return NonlinearVariationalProblem(replace(problem.F, F_replace_map),
-                                           F_replace_map[problem.u],
-                                           bcs=problem.bcs,
-                                           J=replace(problem.J, J_replace_map))
+        nlvp = NonlinearVariationalProblem(replace(problem.F, F_replace_map),
+                                          F_replace_map[problem.u],
+                                          bcs=problem.bcs,
+                                          J=replace(problem.J, J_replace_map))
+        nlvp._ad_count_map = _ad_count_map
+        return nlvp
--- a/firedrake/adjoint/variational_solver.py
+++ b/firedrake/adjoint/variational_solver.py
@@ -28,6 +28,7 @@ class NonlinearVariationalSolverMixin:
             self._ad_problem = problem
             self._ad_args = args
             self._ad_kwargs = kwargs
+            self._ad_nlvs = None
 
         return wrapper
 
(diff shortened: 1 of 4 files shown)


salazardetroya:
Solving that equation in a 10000 by 10000 mesh in only 20 seconds seems amazing. Can you post the entire script with the solver details you used? How many processors?

firedrake/adjoint/variational_solver.py
```diff
@@ -64,3 +76,41 @@ def wrapper(self, **kwargs):
...
+        J_coefficients = problem.J.coefficients()
+
+        for block_variable in dependencies:
+            coeff = block_variable.output
+            if coeff in F_coefficients and coeff not in F_replace_map:
+                if isinstance(coeff, Constant):
+                    F_replace_map[coeff] = copy.deepcopy(coeff)
+                else:
+                    F_replace_map[coeff] = coeff.copy(deepcopy=True)
+                F_replace_map[coeff]._ad_count = coeff.count()
```
nbouziani, line 102: Here, you are creating a new attribute for `Coefficients`, I think it would be better to define this new attribute directly in the appropriate class. I guess if it is "adjoint-related" you could use the `FunctionMixin` (cf. `/firedrake/adjoint/function.py`).

Also if I have well understood, here you replace the coefficients turning up in the forms (F and J) with copies, and in order to keep track of this substitution, you attach the count of the original coefficients to their copies introducing a new attribute (`_ad_count`) in order to identify the copies later on (in the block) to then replace them with the saved outputs, is it correct?

How about instead attaching this information (mapping) to the `NonlinearVariationalProblem` itself (using for example the `NonlinearVariationalProblemMixin`) and transmit it to the block, so that you have it when needed (i.e. in `_forward_solve`)?

firedrake/adjoint/variational_solver.py
```diff
@@ -64,3 +76,41 @@ def wrapper(self, **kwargs):
...
+
+            if coeff in J_coefficients and coeff not in J_replace_map:
+                if coeff in F_replace_map:
+                    J_replace_map[coeff] = F_replace_map[coeff]
+                elif isinstance(coeff, Constant):
+                    J_replace_map[coeff] = copy.deepcopy(coeff)
+                else:
+                    J_replace_map[coeff] = coeff.copy()
+                J_replace_map[coeff]._ad_count = coeff.count()
```
nbouziani, line 111: Same remark

firedrake/adjoint/variational_solver.py
mtusman: Your understanding of the use the attribute `_ad_count` is absolutely correct. I think creating a `NonlinearVariationalProblemMixin` class and attaching the mapping to it seems reasonable.

mtusman:
> Solving that equation in a 10000 by 10000 mesh in only 20 seconds seems amazing. Can you post the entire script with the solver details you used? How many processors?

I'm just using the default solve parameters. I just the programme on my laptop equipped with a 3.5 GHz Dual-Core i7 processor. The script is as follows:
```
from firedrake import *
import time
import numpy as np

tape = get_working_tape()
tape.clear_tape()

n = 10000
mesh = UnitIntervalMesh(n)
V = FunctionSpace(mesh, "CG", 2)

def Dt(u, u_, timestep):
    return (u - u_)/timestep

x, = SpatialCoordinate(mesh)
ic = project(sin(2*pi*x),  V)

u_ = Function(V)
u = Function(V)
v = TestFunction(V)

nu = Constant(0.0001)

timestep = Constant(1.0/n)

F = (Dt(u, ic, timestep)*v
     + u*u.dx(0)*v + nu*u.dx(0)*v.dx(0))*dx
bc = DirichletBC(V, 0.0, "on_boundary")

t = 0.0
problem = NonlinearVariationalProblem(F, u, bcs=bc)
solver = NonlinearVariationalSolver(problem)
solver.solve()
u_.assign(u)
t += float(timestep)

F = (Dt(u, u_, timestep)*v
     + u*u.dx(0)*v + nu*u.dx(0)*v.dx(0))*dx

end = 0.2
while (t <= end):
    solver.solve()
    u_.assign(u)
    t += float(timestep)
J = assemble(u_*u_*dx + ic*ic*dx)
Jhat = ReducedFunctional(J, Control(ic))

def recompute():
    Jhat(ic)

times = []
repeat = 10
for i in range(repeat):
    start = time.time()
    recompute()
    end = time.time()
    times.append((end - start))
print("Average: {}(s)".format(np.mean(times)))
```

salazardetroya:
Oh, it is just a 1D mesh, not a square mesh. That makes sense now.

mtusman:
> Oh, it is just a 1D mesh, not a square mesh. That makes sense now.

My bad, I just noticed that i described the wrong mesh in the description. I'll change it.

firedrake/adjoint/variational_solver.py
nbouziani: Ok perfect, actually the `NonlinearVariationalProblemMixin` class already exists (cf. `firedrake/adjoint/variational_solver.py`)

firedrake/adjoint/variational_solver.py
```diff
@@ -64,3 +77,44 @@ def wrapper(self, **kwargs):
...
+                    J_replace_map[coeff] = copy.deepcopy(coeff)
+                else:
+                    J_replace_map[coeff] = coeff.copy()
+                _ad_count_map[J_replace_map[coeff]] = coeff.count()
+
+        nlvp = NonlinearVariationalProblem(replace(problem.F, F_replace_map),
+                                           F_replace_map[problem.u],
+                                           bcs=problem.bcs,
+                                           J=replace(problem.J, J_replace_map))
+        nlvp._ad_count_map = _ad_count_map
```
nbouziani, line 119: Similarly could we also define/set up the `_ad_count_map` attribute inside the class (i.e. inside the mixin)? So that we don't end up defining a new attribute out of the class.

PR #1806, opened 2020-08-17
Further fixes to BCs

ScottMacLachlan:
In #1797 , the original expression passed to DirichletBC is no longer saved.  This creates a problem for some time-steppers, e.g. in Irksome, where we need that expression so we can differentiate it w.r.t. the time variable.  This restores storing the original argument to Dirichlet BC in a way that it is no longer lost later in the process.


wence-:
Thanks Scott!

PR #1812, opened 2020-08-18
Fix mesh labels for refining meshes

ScottMacLachlan:
When creating a mesh hierarchy, we need to properly filter mesh labels that are attached to the mesh (e.g., by gmsh).  This generalizes what was in place to do this.  (This is all @wence- 's code, but I'd like to see it merged to support some work on nasty domains...)


ScottMacLachlan:
Rebase failed, due to a merge conflict.  I think this sorts it out...

wence- (approved):
Thanks.

PR #1814, opened 2020-08-20
Adding change which fixed MLMC in parallel

ntwyman663:
Following slack exchange on 20/08/20


wence-:
Did that in #1822, so closing this.

PR #1816, opened 2020-08-21
Fix memory leak in bc.apply + cleanups

wence-:
The `bc.apply` annotation unconditionally saved the bc on the function it was applied to. Causing a memory leak. This was noticed in defcon where new bc objects are made as part of the continuation/deflation process, but applied to the same state vector.

This is clearly a memory leak full stop, but I don't have any pyadjoint-fu, so I'm not going to try and fix it there. At the very least, this patch removes the behaviour that in annotations off mode, something still happens.

In addition, a bunch of cleanups such that if we're not in adjoint mode, the wrappers/decorators really are no-ops.

Please audit carefully.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -9,12 +9,14 @@
 *.pvtu
 *.so
 *.vtu
+.DS_Store
 __pycache__/
 /demos/**/*.py
 /demos/immersed_fem/immersed_domain.msh
 /docs/build/
 /docs/notebooks/.ipynb_checkpoints/
 /docs/source/firedrake.rst
+/docs/source/firedrake.adjoint.rst
 /docs/source/firedrake.cython.rst
 /docs/source/firedrake.matrix_free.rst
 /docs/source/firedrake.mg.rst
--- a/firedrake/adjoint/dirichletbc.py
+++ b/firedrake/adjoint/dirichletbc.py
@@ -1,7 +1,7 @@
 from functools import wraps
 from pyadjoint.overloaded_type import FloatingType
 from .blocks import DirichletBCBlock
-from pyadjoint.tape import no_annotations
+from pyadjoint.tape import stop_annotating, annotate_tape
 
 
 class DirichletBCMixin(FloatingType):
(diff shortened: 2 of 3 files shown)


pefarrell:
The test that this fails is `FAILED tests/regression/test_adjoint_bc.py::test_bcs_collected_when_not_annotating`.

This seems counterintuitive. Why do we test that the BCs are collected even when not anotating?

wence-:
> Why do we test that the BCs are collected even when not anotating?

collected in the sense of garbage collected.

pefarrell:
Apologies, I misunderstood.

.gitignore
```diff
@@ -9,12 +9,14 @@
 *.pvtu
 *.so
 *.vtu
+.DS_Store
```
pefarrell, line 12: This doesn't seem like a relevant change?

firedrake/adjoint/dirichletbc.py
```diff
@@ -20,14 +20,17 @@ def wrapper(self, *args, **kwargs):
...
     def _ad_annotate_apply(apply):
-        @no_annotations
         @wraps(apply)
         def wrapper(self, *args, **kwargs):
-            for arg in args:
-                if not hasattr(arg, "bcs"):
-                    arg.bcs = []
-            arg.bcs.append(self)
-            return apply(self, *args, **kwargs)
+            annotate = annotate_tape(kwargs)
```
pefarrell, line 25: This is absolutely essential for defcon. Before this PR, defcon reliably crashed whenever I tried to run long bifurcation analyses, killed by the out-of-memory manager. After this change, the same code runs happily for weeks. Thanks so much, @wence- !

pefarrell (approved):

.gitignore
wence-: I guess I could spin it out: it's cos I was adding the firedrake.adjoint.rst ignore. (This is a file MacOS finder dumps everywhere).

pefarrell: Not worth the effort, keep it as-is.

PR #1817, opened 2020-08-21
Allow multiple vectorfunctionspace subcomponents in pointwise assembly

wence-:
This allows `z.assign(x - y)` where the space is mixed and you have multiple VFS bits.

Doesn't support (but didn't previously) the case where you have VFS(RT) [say].

### DIFF
--- a/firedrake/assemble_expressions.py
+++ b/firedrake/assemble_expressions.py
@@ -3,13 +3,12 @@ import weakref
 from collections import OrderedDict, defaultdict
 from functools import singledispatch
 
-import firedrake
 import gem
 import loopy
+import numpy
 import ufl
-from firedrake.utils import ScalarType, cached_property, known_pyop2_safe
 from gem.impero_utils import compile_gem, preprocess_gem
-from gem.node import Memoizer
+from gem.node import MemoizerArg
 from gem.node import traversal as gem_traversal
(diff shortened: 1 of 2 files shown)


pefarrell (approved):
Looks good. Fixes our bug with mixed assignment. Thanks!

firedrake/assemble_expressions.py
```diff
@@ -158,19 +180,25 @@ def _split_component_tensor(o, self):
...
-        else:
+        elif not inct:
+            idx = indices[:ncmp]
+            indices = indices[ncmp:]
             mi = multiindex if multiindex.indices() == idx else MultiIndex(idx)
             result.append(ufl_reuse_if_untouched(o, agg, mi))
+        else:
+            # shape and inct
+            aggshape = (flatten(agg.ufl_shape)
+                        + tuple(itertools.repeat(1, len(aggregate.ufl_shape) - 1)))
```
ksagiyam, line 199: Do we ever get here with `len(aggregate.ufl_shape) != 1`?

wence-: I suspect I did it for belt-and-braces.

wence-: I guess know because `split_coefficient` flattens.

PR #1825, opened 2020-08-27
Some improvements in slate compilation time

wence-:
With the slate->loopy pipeline, codegen for slate expressions has gone up quite a lot. For example here: https://jenkins.ese.ic.ac.uk/blue/organizations/jenkins/firedrake/detail/master/277/pipeline/67

The top test length culprits are now all slate tests with the longest taking over half an our. All of this time is spent in codegen, with a warm cache it takes about 3 seconds.

These changes remove the need to inline the slate and tsfc kernels into the slate wrapper. While this may not be the preferred option long term, it approximately halves the time loopy spends in codegen. I would therefore advocate that we do something like this until loopy codegen is blazing fast.

As a consequence, I tidy up the loopy merging a little bit such that the `slate_loopy` call doesn't rely on inlining to resolve a bunch of the variables it references.

### DIFF
--- a/firedrake/slate/slac/kernel_builder.py
+++ b/firedrake/slate/slac/kernel_builder.py
@@ -19,7 +19,7 @@ import firedrake.slate.slate as slate
 from firedrake.slate.slac.tsfc_driver import compile_terminal_form
 
 from firedrake.parameters import parameters
-from tsfc.loopy import create_domains
+from tsfc.loopy import create_domains, assign_dtypes
 
 from pytools import UniqueNameGenerator
 
@@ -607,7 +607,9 @@ class LocalLoopyKernelBuilder(object):
         tensor2temp = OrderedDict()
         inits = []
         for gem_tensor, slate_tensor in var2terminal.items():
+            (_, dtype), = assign_dtypes([gem_tensor], self.tsfc_parameters["scalar_type"])
             loopy_tensor = loopy.TemporaryVariable(gem_tensor.name,
+                                                   dtype=dtype,
                                                    shape=gem_tensor.shape,
                                                    address_space=loopy.AddressSpace.LOCAL)
             tensor2temp[slate_tensor] = loopy_tensor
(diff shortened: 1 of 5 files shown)


firedrake/slate/slac/kernel_builder.py
```diff
@@ -437,13 +435,14 @@ def __init__(self, expression, tsfc_parameters=None):
...
         self.bag = None
+        self.kernel_counter = count()
 
     def tsfc_cxt_kernels(self, terminal):
         r"""Gathers all :class:`~.ContextKernel`\s containing all TSFC kernels,
         and integral type information.
         """
 
-        return compile_terminal_form(terminal, prefix="subkernel%s_" % terminal._output_string,
+        return compile_terminal_form(terminal, prefix=f"subkernel{next(self.kernel_counter)}_",
```
sv2518, line 445: Ah this is better, thanks!

firedrake/slate/slac/utils.py
```diff
@@ -243,42 +243,6 @@ def slate2gem(expression):
     return mapper(expression), mapper.var2terminal
 
 
-def eigen_tensor(expr, temporary, index):
```
sv2518, line 246: AH, I didn't realise we could get rid of this now. Thanks!

firedrake/slate/slac/utils.py
```diff
@@ -347,12 +310,9 @@ def merge_loopy(slate_loopy, builder, var2terminal):
     slate_wrapper = lp.make_function(domains, insns, args, name="slate_wrapper",
                                      seq_dependencies=True, target=lp.CTarget())
 
-    # Generate program from kernel, so that one can register and inline kernels
+    # Generate program from kernel, so that one can register kernels
     prg = make_program(slate_wrapper)
     for tsfc_loopy in tsfc_kernels:
         prg = register_callable_kernel(prg, tsfc_loopy)
-        prg = inline_callable_kernel(prg, tsfc_loopy.name)
```
sv2518, line 354: Okay, we might need to inline at some point as soon as I got vectorisation working, but maybe we want to control that from PyOP2 level anyways.

sv2518 (approved):
Looks good to me, thanks!

PR #1826, opened 2020-08-28
Add MMS convergence tests for MTW, AWc and AWnc.

pefarrell:


tests/regression/test_aw.py
```diff
@@ -0,0 +1,123 @@
...
+    print(convergence_orders(l2_div_sigma))
+
+    if stress_element.family().startswith("Conforming"):
+        assert min(convergence_orders(l2_u)) > 1.9
+        assert min(convergence_orders(l2_sigma)) > 2.9
+        assert min(convergence_orders(l2_div_sigma)) > 1.9
+    elif stress_element.family().startswith("Nonconforming"):
+        assert min(convergence_orders(l2_u)) > 1.9
+        assert min(convergence_orders(l2_sigma)) > 1
+        assert min(convergence_orders(l2_div_sigma)) > 1.9
```
rckirby, line 123: else an error?

tests/regression/test_aw.py
```diff
@@ -0,0 +1,123 @@
...
+                  "snes_max_it": 100,
+                  "snes_monitor": None,
+                  "snes_converged_reason": None,
+                  "ksp_type": "preonly",
+                  "pc_type": "lu",
+                  "pc_factor_mat_solver_type": "umfpack",
+                  "mat_type": "aij",
+                  "snes_rtol": 1e-5,
+                  "snes_atol": 1e-25,
+                  "snes_stol": 0.0}
```
rckirby, line 100: do we know how essential mumps and/or iterative refinement are to the example, or can we get away with more vanilla options?

tests/regression/test_aw.py
pefarrell: Fixed

tests/regression/test_aw.py
pefarrell: `snes_type`: `ksponly` works. Thanks for the suggestion. We need some LU library not PETSc because the problem is of saddle point type, but both UMFPACK and MUMPS will work.

rckirby: snes was for iterative refinement, if you turn that to ksponly, you can also remove all the other snes-related options/tolerances

pefarrell: I got rid of all the tolerances.

tests/regression/test_aw.py
```diff
@@ -0,0 +1,120 @@
...
+            + inner(A(sigh), tau)*dx
+            + inner(div(tau), uh)*dx
+            + inner(div(sigh), v)*dx
+            - inner(comp_r, tau)*dx
+            - inner(cons_r, v)*dx
+            - inner(dot(tau, n), uex)*ds
+            )  # noqa: E123
+
+        params = {"snes_type": "ksponly",
+                  "snes_monitor": None,
```
rckirby, line 90: with `ksponly`, you don't need a `snes_monitor` or `snes_converged_reason`.

pefarrell: I agree you don't need it. I kept it so that if someone runs it themselves while requesting stdout, the system prints the residuals. By default the test harness swallows the output unless something goes wrong.

rckirby: You're right -- it's in a test and you get some information if it fails and you turn on the printout.  I was wondering if the potential reader would be confused by the options and not know the use case.

pefarrell: I think these options are pretty common and the risk of confusion is minimal.

tests/regression/test_aw.py
```diff
@@ -0,0 +1,120 @@
...
+        error_sigma = sqrt(assemble(inner(sigh - sigex, sigh - sigex)*dx))
+        error_div_sigma = sqrt(assemble(inner(div(sigh - sigex), div(sigh - sigex))*dx))
+
+        l2_u.append(error_u)
+        l2_sigma.append(error_sigma)
+        l2_div_sigma.append(error_div_sigma)
+
+    print(convergence_orders(l2_u))
+    print(convergence_orders(l2_sigma))
+    print(convergence_orders(l2_div_sigma))
```
miklos1, line 109: Tests shouldn't print.

pefarrell: Removed.

FAznaran (approved):

wence- (approved):

PR #1829, opened 2020-09-01
Feature periodic boxmesh

florian98765:
Added 3D ```PeriodicBoxMesh``` to utility meshes. 

the code is based on the ```BoxMesh``` code, but the periodic nodes are eliminated. A plex object is created using the correct cell topology. A mesh created from this plex object by default uses a CG coordinate field, which leads to wrong coordinates (0. instead of Lx/Ly/Lz) for the periodic boundary nodes. Those wrong entries can be detected and changed to the correct ones by a simple parloop. 

Note: The same procedure could also be used in 1D and 2D, which would eliminate the need for the (N+1) dimensions helper mesh (Circular and Torus mesh).

### DIFF
--- a/tests/regression/test_mesh_generation.py
+++ b/tests/regression/test_mesh_generation.py
@@ -157,11 +157,11 @@ def test_box():
 
 
 def test_periodic_unit_cube():
-    assert abs(integrate_one(PeriodicUnitCubeMesh(3,3,3)) - 1) < 1e-3
+    assert abs(integrate_one(PeriodicUnitCubeMesh(3, 3, 3)) - 1) < 1e-3
 
 
 def test_periodic_box():
-    assert abs(integrate_one(PeriodicBoxMesh(3,3,3,2.,3.,4.)) - 24.0) < 1e-3
+    assert abs(integrate_one(PeriodicBoxMesh(3, 3, 3, 2., 3., 4.)) - 24.0) < 1e-3
 
 
 def test_unit_triangle():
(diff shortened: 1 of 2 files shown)


tests/regression/test_mesh_generation.py
```diff
@@ -156,6 +156,14 @@ def test_box():
     assert abs(integrate_one(BoxMesh(3, 3, 3, 1, 2, 3)) - 6) < 1e-3
 
 
+def test_periodic_unit_cube():
+    assert abs(integrate_one(PeriodicUnitCubeMesh(3,3,3)) - 1) < 1e-3
+
+
+def test_periodic_box():
+    assert abs(integrate_one(PeriodicBoxMesh(3,3,3,2.,3.,4.)) - 24.0) < 1e-3
```
wence-, line 164: Coding style here: please add spaces after the commas.

tests/regression/test_mesh_generation.py
```diff
@@ -199,6 +207,11 @@ def test_unit_cube_parallel():
     assert abs(integrate_one(UnitCubeMesh(3, 3, 3)) - 1) < 1e-3
 
 
+@pytest.mark.parallel
+def test_periodic_unit_cube_parallel():
+    assert abs(integrate_one(PeriodicUnitCubeMesh(3,3,3)) - 1) < 1e-3
```
wence-, line 212: And here.

wence- (changes requested):
Thanks! Very minor style fixes to pacify flake8, otherwise this looks great!

Issue #1658, opened 2020-04-16
matrix_free/operators.py AttributeError


AndrewLister-STFC:
I am getting the following error when trying to run a solve on a problem with no dirichlet boundaries:

```
Traceback (most recent call last):
  File "libpetsc4py/libpetsc4py.pyx", line 879, in libpetsc4py.MatMultTranspose_Python
  File "/home/alister/firedrake/src/firedrake/firedrake/matrix_free/operators.py", line 252, in multTranspose
    self._xbc.dat.zero()
AttributeError: 'ImplicitMatrixContext' object has no attribute '_xbc'
```

This in turn causes the solve to fail:

```
  File "/home/alister/firedrake/src/firedrake/firedrake/adjoint/variational_solver.py", line 55, in wrapper
    out = solve(self, **kwargs)
  File "/home/alister/firedrake/src/firedrake/firedrake/variational_solver.py", line 271, in solve
    solving_utils.check_snes_convergence(self.snes)
  File "/home/alister/firedrake/src/firedrake/firedrake/solving_utils.py", line 44, in check_snes_convergence
    %s""" % (snes.getIterationNumber(), msg))
firedrake.exceptions.ConvergenceError: Nonlinear solve failed to converge after 0 nonlinear iterations.
Reason:
   DIVERGED_LOCAL_MIN
```

It looks like `_xbc` is only defined when there are dirichlet bounds. (Although I'm not sure if that means the multTranspose function should be inaccessable?).

Any help would be greatly appreciated.

wence-:
Sorry for the slow response, this looks like a bug. @ksagiyam this is your code, could you take a look?

ksagiyam:
@AndrewLister-STFC  Sorry for the late. I attempted to fix this on `fix_mat_transpose` branch. Could you try using this branch, or could you give us a failing example that illustrate your problem? Thanks.

wence-:
Was that fix ever merged?

wence-:
This was fixed by #1833.

PR #1833, opened 2020-09-03
Fix mat transpose

ksagiyam:
https://github.com/firedrakeproject/firedrake/issues/1658#issuecomment-686044120

Temporary `self._xbc` is only created/used when `len(self.bcs) > 0`.


firedrake/matrix_free/operators.py
```diff
@@ -251,31 +251,36 @@ def multTranspose(self, mat, Y, X):
...
+            for aT, obj in zip(self._assemble_actionT, self.objs_actionT):
+                # zero columns associated with DirichletBCs/EquationBCs
+                for obc in obj.bcs:
+                    obc.zero(self._y)
+                aT()
+                self._x += self._xbc
+        else:
+            # No DirichletBC/EquationBC
+            # Save to self._x directly
+            self._assemble_actionT[0]()
```
wence-, line 270: Loop over the list here and call all of them? Just in case there's ever more than one.

firedrake/matrix_free/operators.py
```diff
@@ -266,6 +266,7 @@ def multTranspose(self, mat, Y, X):
                 self._x += self._xbc
         else:
             # No DirichletBC/EquationBC
+            # There is only a single element in the list (for the domain equation).
```
wence-, line 269: I think it is safer to assert that explicitly, e.g. by:

```
aT, = self._assemble_actionT
aT()
```

or else just loop over them (in case something changes).

ksagiyam: This looks nice! Now fixed.

wence- (approved):

Issue #1843, opened 2020-09-16
Issue using split() with firedrake_adjoint


mc4117:
Firedrake adjoint is currently not taping the function split correctly. When calculating the derivative of the reduced functional, it returns an output saying that the functional is independent of the control when it is clearly not. I have attached a simple example to show what I mean.
[example.txt](https://github.com/firedrakeproject/firedrake/files/5233245/example.txt)

stephankramer:
There's several issues here. One is related to AdjFloat being assigned to Constants which we should probably deal with separately - we can avoid that one in the example by changing to a Constant control.

The other issues are all related to MergeBlock:
1. Merge block does not register the parent mixed function that it merges into, as a dependency, but it does change the state of that mixed function thus creating a new blockvariable on the output. This means in your example when the second project is merged back into test_fn it does not know that its state after the merge depends on the control - this is the immediate warning you're getting
2. when we fix that it turns out it's also not computing the adjoint of MergeBlock correctly: the adjoint solution of the output of merge-block should be restricted to serve as the adjoint solution associated with the input sub-function
3. the forward replay is also incorrect: when merging into a mixed function it shouldn't just replay the merge into the output checkpoint, the output checkpoint as it won't be initialised - it should first receive a copy of the checkpointed full mixed function before the merge

PR #1844, opened 2020-09-17
Various fixes in FunctionMergeBlock

stephankramer:
1. Merge block does not register the parent mixed function that it merges into, as a dependency, but it does change the state of that mixed function thus creating a new blockvariable on the output. This means pyadjoint looses track of the dependency of that output state on the control
2. the adjoint is incorrect: the adjoint solution of the output of merge-block should be restricted to serve as the adjoint solution associated with the input sub-function
3. the forward replay is also incorrect: when merging into a mixed function it shouldn't just replay the merge into the output checkpoint, the output checkpoint as it won't be initialised - it should first receive a copy of the checkpointed full mixed function before the merge

Partial fix for #1843

### DIFF
--- a/firedrake/adjoint/blocks.py
+++ b/firedrake/adjoint/blocks.py
@@ -225,10 +225,15 @@ class FunctionMergeBlock(Block, Backend):
         super().__init__()
         self.add_dependency(func)
         self.idx = idx
+        for output in func._ad_outputs:
+            self.add_dependency(output)
 
     def evaluate_adj_component(self, inputs, adj_inputs, block_variable, idx,
                                prepared=None):
-        return adj_inputs[0]
+        if idx == 0:
+            return adj_inputs[0].split()[self.idx]
+        else:
+            return adj_inputs[0]
 
     def evaluate_tlm(self):
         tlm_input = self.get_dependencies()[0].tlm_value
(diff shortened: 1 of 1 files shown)


firedrake/adjoint/blocks.py
```diff
@@ -245,9 +250,12 @@ def evaluate_hessian_component(self, inputs, hessian_inputs, adj_inputs,
...
     def recompute(self):
-        dep = self.get_dependencies()[0].checkpoint
-        output = self.get_outputs()[0].checkpoint
-        self.backend.Function.assign(self.backend.Function.sub(output, self.idx), dep)
+        deps = self.get_dependencies()
+        sub_func = deps[0].checkpoint
+        parent_in = deps[1].checkpoint
+        parent_out = self.get_outputs()[0].checkpoint
+        self.backend.Function.assign(parent_out, parent_in)
+        self.backend.Function.assign(self.backend.Function.sub(parent_out, self.idx), sub_func)
```
wence-, line 258: Could simplify and just write:
```
parent_out.assign(parent_in)
parent_out.sub(self.idx).assign(sub_func)
```

stephankramer:
@nbouziani could you take a look at this PR?

firedrake/adjoint/blocks.py
```diff
@@ -245,9 +250,12 @@ def evaluate_hessian_component(self, inputs, hessian_inputs, adj_inputs,
         return hessian_inputs[0]
 
     def recompute(self):
-        dep = self.get_dependencies()[0].checkpoint
-        output = self.get_outputs()[0].checkpoint
-        self.backend.Function.assign(self.backend.Function.sub(output, self.idx), dep)
+        deps = self.get_dependencies()
+        sub_func = deps[0].checkpoint
```
nbouziani, line 254: Is there a reason for using `checkpoint` instead of `saved_output` as it is usually done in `recompute` ?

nbouziani (approved):

dham (approved):

Issue #1643, opened 2020-04-04
Hypre AMS Preconditioner


florian98765:
Dear Firedrake Developers,

I am currently using the Hypre AMS Preconditioner within FEniCS. I am wondering what are the missing pieces to make AMS work within Firedrake. As far as I can see to only missing thing is the ``` build_gradient``` function which calculates the discrete gradient matrix, right?

Can the FEniCS code be adopted, or are there any further complicates which I don't see?

best wishes
Florian

rckirby:
This has been done (I did it with a former student) — I just haven’t committed it.  It’s done through the Firedrake/PETSc interface, and we have a little loopy kernel that builds the gradient.  

Currently it’s restricted to AMS (could be extended to ADS if you provide a discrete curl) and triangles, tets, and quads.  And low-order methods.

I’ve shared with a few people in firedrake land but nobody has beat me to putting in a PR on it.

rckirby:
Currently this implementation assumes you use matrix-free outer matrices
(fishes out vertices, etc from a Python context associated with the
stiffness matrix).  It should be fixable to have the preconditioner fish
things out elsewhere.

florian98765:
Sounds perfect. I never used matrix-free "matrices", but I will give it a try. 

What about the beta poisson matrix, which could be passed to hypre?
If I understand it correctly, this should simply be ```sigma*inner(grad(u), grad(v))*dx```, where ```u``` and ```v``` are nodal trial- and test-function and sigma can be zero in some regions.
Right now I do not provide the beta poisson matrix. The solver converges, but it seems to be less stable than the ```sigma=0``` case, where I could use ```setBetaPoissonMatrix(None)```. Did you every use this?

One last question: Is it possible to use this code with complex number support?
How can I get this code? Could you try to commit it?

best wishes 
Florian

pefarrell:
I'd like to comment that firedrake also supports excellent geometric multigrid methods for problems in H(div) and H(curl). Using PCPATCH we can implement the Arnold-Falk-Winther scheme, which works for all H(div)/H(curl)-conforming elements of all degrees (in its vertex-patch version). Here's a demo code that solves the H(div) or H(curl) Riesz map. Try running it with `python hdivcurl.py --op div`:

```
from firedrake import *

import argparse
from datetime import datetime

parser = argparse.ArgumentParser(add_help=False)
parser.add_argument("--baseN", type=int, default=5)
parser.add_argument("--nref",  type=int, default=1)
[...]

PR #1854, opened 2020-09-29
Feature hypre ams

florian98765:
WIP: Added Hypre AMS preconditioner

Fix issue #1643


florian98765:
Currently only the linear case is tested and I would like to generalized to non-linear problems and if possible to matfree matrices. Unfortunately I am a bit lost with the different cases:

- if matfree is used ```Pmat``` needs to be assembled like in the ```assembled.py```. Did this code change recently? I currently have some different versions of how to get the ```appctx``` and I don't know which one to prefer?
- maybe the assembling part should be pushed into a base-class or one could inherit from ```AssembledPC``` 
- using a non-linear form does not work, yet. When a check ```P``` during initialization, it seems to be empty. I also don't see where it is updated during the nonlinear iteration. Do I have to assemble ```P``` within the ```update``` function, or is this already done by Petsc / Firedrake?

thanks for any advice
best wishes 
Florian

wence-:
> Currently only the linear case is tested and I would like to generalized to non-linear problems and if possible to matfree matrices. Unfortunately I am a bit lost with the different cases:

> if matfree is used Pmat needs to be assembled like in the assembled.py. Did this code change recently? I currently have some different versions of how to get the appctx and I don't know which one to prefer?

I think the best way to do this is just to write this PC to work on assembled operators. For the matfree case, you can then do:

```
-mat_type matfree
-pc_type python
-pc_python_type firedrake.AssembledPC
-assembled_pc_type python
-assembled_pc_python_type firedrake.HypreAMS
```

That way, you don't need special code for handling matfree vs assembled in this PC.


> using a non-linear form does not work, yet. When a check P during initialization, it seems to be empty. I also don't see where it is updated during the nonlinear iteration. Do I have to assemble P within the update function, or is this already done by Petsc / Firedrake?

Preconditioners in petsc act on linear operators. Certainly, AMS solves a problem Ax = b, not F(u) = 0. I therefore think that to handle nonlinear problems you shouldn't have to do anything here (just inherit from PCBase, rather than PCSNESBase).

Then you would say

```
-snes_type newtonls
-ksp_type cg # or whatever
-pc_type python
-pc_python_type firedrake.HypreAMS
```

firedrake/preconditioners/hypre_ams.py
```diff
@@ -0,0 +1,77 @@
+from firedrake.petsc import PETSc
+from firedrake import FunctionSpace, Constant, project, Interpolator, grad, TestFunction
+from firedrake.assemble import allocate_matrix, create_assembly_callable
+
+__all__ = ("HypreAMS")
```
wence-, line 5: Needs a comma to turn it into a tuple `("HypreAMS", )`.

firedrake/preconditioners/hypre_ams.py
```diff
@@ -0,0 +1,77 @@
...
+        else:
+            raise ValueError("Not a PC or SNES?")
+
+        prefix = obj.getOptionsPrefix()
+        appctx = self.get_appctx(obj) 
+
+        f = appctx['state']
+        element = str(f.function_space().ufl_element().family()) #not sure this is right
+        mesh = f.function_space().mesh()
+        V = FunctionSpace(mesh, element, 1)
```
wence-, line 22: You want to extract the functionspace from the dm attached to the `PC`.

```python
from firedrake.dmhooks import get_function_space
...

V = get_function_space(obj.dm)
```

firedrake/preconditioners/hypre_ams.py
```diff
@@ -0,0 +1,77 @@
...
+__all__ = ("HypreAMS")
+
+class HypreAMS(firedrake.PCSNESBase):
+    def initialize(self, obj):
+        if isinstance(obj, PETSc.PC):
+            A, P = obj.getOperators()
+        elif isinstance(obj, PETSc.SNES):
+            A, P = obj.ksp.pc.getOperators()
+        else:
+            raise ValueError("Not a PC or SNES?")
```
wence-, line 14: I think this should inherit from `PCBase` and not handle nonlinear things at all.

firedrake/preconditioners/hypre_ams.py
```diff
@@ -0,0 +1,77 @@
...
+        print("element:", element)
+        print("mesh:", mesh)
+        print("f:", f)
+        print("A:", A.getValuesCSR())
+        print("P:", P.getValuesCSR())
+
+        # build gradient matrix G
+        P1 = FunctionSpace(mesh, "Lagrange", 1)
+        Q = FunctionSpace(mesh, "N1curl", 1)
+        G = Interpolator(grad(TestFunction(P1)), Q).callable().handle
```
wence-, line 33: `Q` should really be `V`. And then I think for general (not just linear) order, one wants `Pk = FunctionSpace(mesh, "Lagrange", V.ufl_element().degree())`

firedrake/preconditioners/hypre_ams.py
```diff
@@ -0,0 +1,77 @@
...
+
+        pc = PETSc.PC().create()
+        pc.setOptionsPrefix(prefix + "HypreAMS_")
+        pc.setOperators(A, P)
```
wence-, line 37: I think we tend to spell options lower-case, so `prefix + "hypre_ams_"` ?

firedrake/preconditioners/hypre_ams.py
```diff
@@ -0,0 +1,77 @@
...
+
+        pc.setType('hypre')
+        pc.setHYPREType('ams')
+        pc.setHYPREDiscreteGradient(G)
+        zero_beta = PETSc.Options(prefix).getBool("pc_HypreAMS_zero_beta_poisson", default=False)
+        if zero_beta == True:
+            pc.setHYPRESetBetaPoissonMatrix(None)
```
wence-, line 44: `if zero_beta == True` -> `if zero_beta:`

And maybe for consistency `pc_hypre_ams_zero_beta_poisson` ?

firedrake/preconditioners/hypre_ams.py
```diff
@@ -0,0 +1,77 @@
...
+
+        # Build constants basis for the Nedelec space
+        constants = []
+        cvecs = []
+        for i in range(3):
+            direction = [1.0 if i == j else 0.0 for j in range(3)]
+            c = project(Constant(direction), V)
+            with c.vector().dat.vec_ro as cvec:
+                cvecs.append(cvec)
+        pc.setHYPRESetEdgeConstantVectors(cvecs[0], cvecs[1], cvecs[2])
```
wence-, line 57: I don't recall if AMS works for 2D problems too, but I think this is better written like so:

```
edgeconstants = []
for v in [[1, 0, 0], [0, 1, 0], [0, 0, 1]]:
   with interpolate(Constant(v), V).dat.vec_ro as cvec:
      edgeconstants.append(cvec.copy())
pc.setHYPRESet...(*edgeconstants)
```

firedrake/preconditioners/hypre_ams.py
```diff
@@ -0,0 +1,77 @@
...
+        self.pc = pc
+
+    def apply(self, pc, x, y):
+        self.pc.apply(x, y)
+
+    def applyTranspose(self, pc, x, y):
+        self.pc.applyTranspose(x, y)
+
+    def view(self, pc, viewer=None):
+        super(HypreAMS, self).view(pc, viewer)
```
wence-, line 69: Prefer `super().view(pc, viewer)`

firedrake/preconditioners/hypre_ams.py
```diff
@@ -0,0 +1,77 @@
...
+        viewer.pushASCIITab()
+        if hasattr(self, "pc"):
+            viewer.printfASCII("PC to apply inverse\n")
+            self.pc.view(viewer)
+        viewer.popASCIITab()
```
wence-, line 74: Rather than push/pop tab, in the `initialize` function, we should have instead set `self.pc.incrementTabLevel(1, parent=obj)`

firedrake/preconditioners/hypre_ams.py
```diff
@@ -0,0 +1,77 @@
...
+
+    def update(self, pc):
+        pass
```
wence-, line 77: This maybe needs to call `self.pc.setUp()` (this function is called if the operator changes, i.e. when the preconditioner is reused in a nonlinear iteration).

firedrake/preconditioners/hypre_ams.py
```diff
@@ -0,0 +1,77 @@
...
+        print("f:", f)
+        print("A:", A.getValuesCSR())
+        print("P:", P.getValuesCSR())
+
+        # build gradient matrix G
+        P1 = FunctionSpace(mesh, "Lagrange", 1)
+        Q = FunctionSpace(mesh, "N1curl", 1)
+        G = Interpolator(grad(TestFunction(P1)), Q).callable().handle
+
+        pc = PETSc.PC().create()
```
wence-, line 35: `create(comm=obj.comm)`

tests/regression/test_hypre_ams.py
```diff
@@ -0,0 +1,82 @@
...
+    
+    bc = DirichletBC(V, ConstantField(1.), (1,2,3,4))
+    
+    params = {'snes_type': 'ksponly',
+              'ksp_type': 'cg',
+              'ksp_itmax': '30',
+              'ksp_rtol': '1e-15',
+              'ksp_atol': '1e-15',
+              'pc_type': 'python',
+              'pc_python_type': 'HypreAMS_preconditioner.HypreAMS',
```
wence-, line 35: Should be `firedrake.HypreAMS`

tests/regression/test_hypre_ams.py
```diff
@@ -0,0 +1,82 @@
...
+    class ConstantField(Expression):
+      def __init__(self, B0):
+        self._B0 = B0
+        super().__init__()
+      def eval(self, value, X):
+        value[0] = -0.5*self._B0*(X[1]-0.5)
+        value[1] =  0.5*self._B0*(X[0]-0.5)
+        value[2] = 0.0
+      def value_shape(self):
+        return (3,)
```
wence-, line 25: Prefer just:

```python
B0 = 1
constant_field = as_vector([-0.5*B0*(y - 0.5), 0.5*B0*(x - 0.5), 0])

bc = DirichletBC(V, constant_field, (1, 2, 3, 4))
```

tests/regression/test_hypre_ams.py
```diff
@@ -0,0 +1,82 @@
...
+    
+    bc = DirichletBC(V, ConstantField(1.), (1,2,3,4))
+    
+    params = {'snes_type': 'ksponly',
+              'ksp_type': 'cg',
+              'ksp_itmax': '30',
+              'ksp_rtol': '1e-15',
+              'ksp_atol': '1e-15',
+              'pc_type': 'python',
+              'pc_python_type': 'HypreAMS_preconditioner.HypreAMS',
```
wence-, line 75: Same comments here as for the linear case.

wence- (changes requested):
Thanks! This looks pretty good. I had a bunch of notes and queries.

florian98765:
Dear Lawrence, most of your comments are commited (with some small changes):
- 2D should work
-  but I think Hypre currently only supports lowest order elements N1curl order 1 and CG order 1. I also don't know if those orders need to be related? Currently I only check for the corrent function space and raise an exception otherwise. 
- matfree works without code modifications. this is very elegant!
- unfortunately the nonlinear case is still broken. I added a regression test, but it breaks during setup of the preconditioner within the hypre library (```jac->setup(): error code 12```).

wence-:
> but I think Hypre currently only supports lowest order elements N1curl order 1 and CG order 1. I also don't know if those orders need to be related? Currently I only check for the corrent function space and raise an exception otherwise.

According to these docs: https://hypre.readthedocs.io/en/latest/solvers-ams.html#high-order-discretizations

High order is supported, but we have to provide some slightly different information. Perhaps just provide an error for now.

florian98765:
> High order is supported, but we have to provide some slightly different information. Perhaps just provide an error for now.
Yes, I know, but I am not really sure how I could build the interpolation matrix Gamma. Any ideas?
The current code provides to EdgeConstantVector which limits the use to lowest order elements.

florian98765:
Dear Lawrence,
the nonlinear test-case is fixed, but I still get some strange within the CI, which I don't get with my standalone code.
I needed to fix some imports and still have some problems:
```
   File "libpetsc4py/libpetsc4py.pyx", line 1298, in libpetsc4py.PCSetUp_Python
   File "/home/jenkins/workspace/firedrake_PR-1854/firedrake/preconditioners/base.py", line 125, in setUp
     super().setUp(pc)
   File "/home/jenkins/workspace/firedrake_PR-1854/firedrake/preconditioners/base.py", line 45, in setUp
     self.initialize(pc)
   File "/home/jenkins/workspace/firedrake_PR-1854/firedrake/preconditioners/hypre_ams.py", line 28, in initialize
     G = Interpolator(grad(TestFunction(P1)), V).callable().handle
   File "/home/jenkins/workspace/firedrake_PR-1854/firedrake/interpolation.py", line 73, in __init__
     self.callable, arguments = make_interpolator(expr, V, subset, access)
   File "/home/jenkins/workspace/firedrake_PR-1854/firedrake/interpolation.py", line 165, in make_interpolator
     argfs = arguments[0].function_space()
AttributeError: 'Argument' object has no attribute 'function_space'
```
Any ideas how this can be fixed?

florian98765:
Finally I got the code running. Just for interest, why did ```from firedrake import ...``` not work?
If I remember correctly I read that one should NOT include from ufl, because some of the functions are overloaded? 
Is there a general rule where things should be included from?

Nevertheless now it works and can be merged.

For higher order an interpolation operator is needed. I think this can also be easily created by
```
Pn = VectorFunctionSpace(mesh, "CG", n)
Vn = FunctionSpace(mesh, "N1curl", n)
I = Interpolator((TestFunction(Pn)), Vn).callable().handle
```
Unfortunately a petsc4py wrapper is missing to use the corresponding interface. If I find some time I will try to add this. 

best wishes and thanks for all the help

wence-:
> Is there a general rule where things should be included from?

Generally always include from `firedrake`. That way we make sure that we get the right objects.

tests/regression/test_hypre_ams.py
```diff
@@ -0,0 +1,103 @@
...
+
+    x, y, z = SpatialCoordinate(mesh)
+    B0 = 1
+    constant_field = as_vector([-0.5*B0*(y - 0.5), 0.5*B0*(x - 0.5), 0])
+
+    bc = DirichletBC(V, constant_field, (1, 2, 3, 4))
+
+    params = {'snes_type': 'ksponly',
+              'ksp_type': 'cg',
+              'ksp_itmax': '30',
```
wence-, line 24: Should be `ksp_max_it`.

tests/regression/test_hypre_ams.py
```diff
@@ -0,0 +1,103 @@
...
```
wence-, line 93: Also here `ksp_max_it`.

firedrake/preconditioners/hypre_ams.py
```diff
@@ -0,0 +1,69 @@
...
+
+        pc.setType('hypre')
+        pc.setHYPREType('ams')
+        pc.setHYPREDiscreteGradient(G)
+        zero_beta = PETSc.Options(prefix).getBool("pc_hypre_ams_zero_beta_poisson", default=False)
+        if zero_beta:
+            pc.setHYPRESetBetaPoissonMatrix(None)
+#        pc.setCoordinates(np.reshape(
+#            mesh._plex.getCoordinates().getArray(),
+#            (-1,mesh.geometric_dimension())))
```
wence-, line 42: Can you remove this commented code please?

firedrake/preconditioners/hypre_ams.py
```diff
@@ -0,0 +1,69 @@
...
+from firedrake.interpolation import Interpolator
+from firedrake.projection import project
+from firedrake.dmhooks import get_function_space
+from ufl import grad
+
+__all__ = ("HypreAMS",)
+
+
+class HypreAMS(PCBase):
+    def initialize(self, obj):
```
wence-, line 15: It would be good to add a bibtex reference to the relevant paper describing the hypre implementation. I think it is this one? https://hypre.readthedocs.io/en/latest/ch-references.html#kova2009

To do that, you do:

```
from firedrake_citations import Citations

Citations().add("Kolev2009", """
BIBTEX STRING
""")
```

Then inside `initialize` do:


```
Citations().register("Kolev2009")
```

Thanks

wence-:
Barring the minor comments I made, I think this looks good, thanks.

florian98765:
> Generally always include from `firedrake`. That way we make sure that we get the right objects.
But importing from firedrake directly led to (strange) errors from the CI (running the same code in an external directory worked).
That's why I now use:
```
from firedrake.functionspace import FunctionSpace
from firedrake.constant import Constant
from firedrake.ufl_expr import TestFunction
from firedrake.interpolation import Interpolator
from firedrake.projection import project
```

firedrake/preconditioners/hypre_ams.py
florian98765: After adding the citation another strange CI error occured. Now PETSc reports that the PC type `hypre` is not known. Did anything else change? Maybe `register` should be moved to `__init__` instead of `initialize`?

florian98765: Ah, I see. The problem seems to occur in the complex test, where hypre is unavailable.

wence-: Add `@pytest.mark.skipcomplex` as a decorator to the two tests.

wence- (approved):
Great, thanks!

PR #1857, opened 2020-10-02
Fix index relabelling in assignments

### LINKED ISSUE #1855: Assign erronously caches index relabelling between assignments
Consider the following code:
```
from firedrake import *
mesh=UnitSquareMesh(1,1)
V=VectorFunctionSpace(mesh, "CG", 1)
W=TensorFunctionSpace(mesh, "CG", 1)
u=Function(V)
w=Function(W)

u.assign(as_vector(2*u[i], i))
u.assign(as_vector(2*u[j], j))
w.assign(as_tensor(2*w[i,j], (i,j)))
```
This fails somewhere deep in UFL. The reason for this is that the `Assign` object stores a `IndexRelabeller` as a class attribute. The `IndexRelabeller` is reset calling `_reset()` in the `slow_key()` method of Assign, but that invalidates the `index_cache` that is stored on the `IndexRelabeller`. In the code above, in the first vector assignment `i=Index(0)` gets relabeled to `Index(0)`, in the second assignment `j=Index(1)` is also relabeled to `Index(0)`. Then because the relabellings are still in the `index_cache` in the final tensor assignment it maps both `i` and `j` to `Index(0)` causing UFL to get upset.

Should the `index_cache` also be reset in `IndexRelabeller._reset()`? Why does the `IndexRelabeller` have to be a class attribute - I probably don't understand this `MultiFunction.reuse_if_untouched` business well enough....

stephankramer:
Fixes #1855
Also test evaluation cache still works correctly after fix.

### DIFF
--- a/firedrake/assemble_expressions.py
+++ b/firedrake/assemble_expressions.py
@@ -108,10 +108,10 @@ class IndexRelabeller(MultiFunction):
     def __init__(self):
         super().__init__()
         self._reset()
-        self.index_cache = defaultdict(lambda: Index(next(self.count)))
 
     def _reset(self):
-        self.count = itertools.count()
+        count = itertools.count()
+        self.index_cache = defaultdict(lambda: Index(next(count)))
 
     expr = MultiFunction.reuse_if_untouched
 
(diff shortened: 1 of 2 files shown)


wence- (approved):
Thanks, looks good I think.

dham (approved):

PR #1863, opened 2020-10-12
Add extrude_function() which extends-by-constant a function defined on the base mesh (or extends using top/bottom values).

bueler:
I did not know where to put the function so it lives in its own file.  I created a test which works, but you have to look at the .pvd result to see; suggestions on making this automatable?

Note the FIXME:  I don't know how to use the 'R' space to extend the function from the FunctionSpace in which it already lives.

Also I am certain I am missing standard input checking, but I don't know what you do.

(Don't expect good design from a pull request by someone who does not know your design principles.)


bueler:
review by @colinjcotter @wence?

firedrake/extrusion_extend.py
```diff
@@ -0,0 +1,44 @@
+from firedrake import FunctionSpace, Function, DirichletBC
+
+__all__ = ['ExtrudedExtendFunction']
+
+def ExtrudedExtendFunction(mesh, base_mesh, fcn, extend_type=None):
```
danshapero, line 5: The `base_mesh` argument isn't needed -- you can access the base through `mesh._base_mesh`.

bueler:
The whole thing could be tossed and built by someone with clarity.  Perhaps you want it to be a method (e.g. `f.extend(extend_type=X)`) and probably it should be put somewhere else and tested in a different way.

Other than the design, the location, the testing, and the documentation, I'm sure I've got it right.

bueler:
Another idea is to skip this proposed API change entirely and just add a subsection to the extruded meshes doc page (https://www.firedrakeproject.org/extruded-meshes.html).  The code in this PR is actually just a couple of lines; the only reason I am spending any time on it is that it is not obvious to this novice.  (And it took a few cycles for @danshapero to make it clear to me.)  Let me know if you want that.

firedrake/extrusion_extend.py
```diff
@@ -0,0 +1,42 @@
...
+        Function come from the values of ``fcn`` on the ``"top"``/``"bottom"``
+        boundary of the extruded mesh.
+    """
+
+    # FIXME need to use fcn.function_space() to get element; e.g.
+    #     Q3D = FunctionSpace(mesh, fcn.function_space().ufl_element, vfamily='R', vdegree=0)
+    # does not work; note the base mesh is mesh._base_mesh
+
+    Q3D = FunctionSpace(mesh, 'CG', 1, vfamily='R', vdegree=0)  # FIXME
+    f3D = Function(Q3D)
```
colinjcotter, line 31: Would be good to have an optional argument for a preallocated f3D, for time stepping.

tests/extrusion/test_extruded_extend.py
```diff
@@ -0,0 +1,22 @@
...
+x,y = SpatialCoordinate(mesh)
+V = FunctionSpace(mesh, 'CG', 1)
+foriginal = Function(V).interpolate(cos(2.0 * pi* x) * y)
+foriginal.rename('foriginal')
+fbottom = ExtrudedExtendFunction(mesh, foriginal, extend_type='bottom')
+fbottom.rename('fbottom')
+ftop = ExtrudedExtendFunction(mesh, foriginal, extend_type='top')
+ftop.rename('ftop')
+File('fextended-topbottom.pvd').write(foriginal,ftop,fbottom)
+
```
colinjcotter, line 22: Test should be a function that has assert statements in that executes silently when passing (and no .pvd writing!).

wence- (commented):
Thanks! I had some comments on generalising the element expansion. Maybe what I wrote works and/or makes sense.

firedrake/extrusion_extend.py
```diff
@@ -0,0 +1,42 @@
...
+def ExtrudedExtendFunction(mesh, fcn, extend_type=None):
+    """On an extruded mesh, extend a function by a constant in the extruded
+    direction.  The returned function uses the degree 0 'R' space in the
+    extruded direction.
+
+    :arg mesh:           the extruded mesh
+
+    :arg fcn:            the function to extend
+
+    ``extend_type`` has three cases:
```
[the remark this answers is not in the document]
bueler: I think this is beyond my Python skills.  How do I use this in a kwarg of `ExtrudedExtendFunction()`?

firedrake/extrusion_extend.py
```diff
@@ -0,0 +1,42 @@
...
+        ``fcn`` is defined on the extruded mesh.  The values of the returned
+        Function come from the values of ``fcn`` on the ``"top"``/``"bottom"``
+        boundary of the extruded mesh.
+    """
+
+    # FIXME need to use fcn.function_space() to get element; e.g.
+    #     Q3D = FunctionSpace(mesh, fcn.function_space().ufl_element, vfamily='R', vdegree=0)
+    # does not work; note the base mesh is mesh._base_mesh
+
+    Q3D = FunctionSpace(mesh, 'CG', 1, vfamily='R', vdegree=0)  # FIXME
```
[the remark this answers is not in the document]
bueler: This must also be beyond my Python skills.  Presumably I test the type of the input element (how?) in `fcn`.  Then choose one of these methods?

firedrake/extrusion_extend.py
bueler: ?

bueler:
As an API change I doubt I should be writing this.  Lawrence has already written three times the code I did, and I don't actually understand what he intends.  As a subsection on the extruded meshes doc page I could add my special case, which already works for me in my application.  In that special case I know the function spaces and they are scalar and boring.  (So I'm not even using this `ExtrudedExtendFunction()` function in my application.) Please advise.

firedrake/extrusion_extend.py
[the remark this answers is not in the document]
wence-: `ExtendType.EXPAND` you then do an if-else as before. Just checking against an enum rather than strings.

firedrake/extrusion_extend.py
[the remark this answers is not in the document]
wence-: Having decided whether you're extending from base, or just extracting you would do

`FunctionSpace(mesh, expand(fcn.ufl_element()))`

Or use `extract` if the input function is also extruded.

This is just a fancy python-esque way of implementing a transformation of a recursive datatype using the visitor pattern. 

(Haskellers would call it a catamorphism)

firedrake/extrusion_extend.py
wence-: Rather than making a new Function, allow it to be passed in. 

So 
```
def ExtudedExtend(..., target=None):
  If target is not None:
       assert target.ufl_element() == Q3D.ufl_element()
        f3d = target
    else:
        f3d = Function(Q3D)
```

firedrake/extrusion_extend.py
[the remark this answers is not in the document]
bueler: The 'top' and 'bottom' labels are exactly as in the ExtrudedMesh() interface.  Thus I propose not to use an Enum here.

firedrake/extrusion_extend.py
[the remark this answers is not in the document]
bueler: I had never seen decorators before.  My understanding now, from this example and not by Python documentation (wherein I could not find said pattern), is that the pattern `@foo.register(bar); def function(x): ...` tells python that `bar` is the type to which the next definition applies.  That is, `function(x)` evaluates `foo(x)` if `x` is of type `bar`.  Let me know if this is wrong?

In any case, it seems to work on the Lagrange elements I am using to test (see 1c37641).

bueler:
Should I add more tests with different element types?  I am not sure if this is finished.

firedrake/extrusion_extend.py
bueler: This makes sense, although I'm not completely sure what a test case would be.

danshapero:
I think testing it with other elements would be good. Lawrence's code checks for vector, tensor, mixed, enriched (think bubble functions for stabilizing Stokes), and H(div)/H(curl) elements. You can easily loop over multiple test inputs with `pytest.mark.parametrize` (it's the cat's PJs) which should make it less tedious than writing separate tests for the whole smorgasbord of vector elements.

If you want to check that your test covers all of the code you wrote, you can do:

    pip3 install pytest-cov
    pytest tests/extrusion/test_extruded_extend.py --cov=firedrake | grep extrusion_extend.py

It's nice to aim for 100%, but it can be annoying to also test error-handling code. @wence-  and @dham are the final word here though.

Silly nitpick, but maybe `extrude_function` is a better name? I realize that this is done at other points in the code but the way it is now makes me think it's a class.

bueler:
Is this actually something ya'll want in the API?  (No one has engaged on that.)  My best guess is that I am doing this extension stuff the wrong way.  Either it is a common operation and no one has ever added it to the API, in which case the most likely explanation is that there is an obvious alternative, or alternatively it is not a common operation and thus this is just clutter.  And a silly way to teach me Python.

danshapero:
I've been getting along by implementing a stripped-down version of this in my own code. It looks like @tkarna has done the same thing in Thetis (code is [here](https://github.com/thetisproject/thetis/blob/0afc140974e159f2d3fb957e4a4d2ff217062562/thetis/utility.py#L787)). I'd like to have it in the Firedrake API and I bet Tuomas would too because then we can both delete a bunch of code and unit tests :D But that's up to the actual maintainers.

I think another thing worth pointing out is that this assumes the node ordering is the same for a function that lives in a space Q on the base mesh as it is for a function that lives in the tensor product space Q x R defined on the extruded mesh. That seems reasonable to me but I don't understand all the lower-level business in PyOP2 and the like.

bueler:
@danshapero 

> I've been getting along by implementing a stripped-down version of this in my own code.
> It looks like @tkarna has done the same thing in Thetis (code is here). I'd like to have it in
> the Firedrake API and I bet Tuomas would too because then we can both delete a bunch
> of code and unit tests

Appreciated.

> I think another thing worth pointing out is that this assumes the node ordering is the same
> for a function that lives in a space Q on the base mesh as it is for a function that lives in the
> tensor product space Q x R defined on the extruded mesh.

Yes indeed.  I have no control over or imagination for the possiblities here.  I just hope it works.

bueler:
@danshapero 

> If you want to check that your test covers all of the code you wrote, you can do:
> pip3 install pytest-cov
> pytest tests/extrusion/test_extruded_extend.py --cov=firedrake | grep extrusion_extend.py
> It's nice to aim for 100%, but it can be annoying to also test error-handling code.
> @wence- and @dham are the final word here though.

Thanks for the pointer.  I have now tried pytest for the first time ... I get it mostly.  Commits 521305b and dd7447e reflect successful usage.

> I think testing it with other elements would be good. Lawrence's code checks for vector, tensor, mixed,
> enriched (think bubble functions for stabilizing Stokes), and H(div)/H(curl) elements.

Working on vector case first.  What do you think it should mean?  2d vector on 2d base_mesh extends to a 2d vector on the extruded mesh?  Or to a 3d vector on the extruded mesh?  In the first case case Lawrence's `expand_vector()` and `extract_vector()` code does not work (I think).

> You can easily loop over multiple test inputs with `pytest.mark.parametrize` (it's the cat's PJs) which should make it less
> tedious than writing separate tests for the whole smorgasbord of vector elements.

I see the usage of `pytest.mark.parametrize` (observing weird variant spelling) in some `test_....py` cases.  Not sure how to efficiently use this in my case because of issues like the vector case where the meaning of elements is related to the dimension of the base space.

firedrake/extrusion_extend.py
bueler: I understand this case and it is in the passed test cases.

firedrake/extrusion_extend.py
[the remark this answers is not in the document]
bueler: I am working through the cases in `test_extruded_extend.py`

bueler:
@danshapero 
> Silly nitpick, but maybe extrude_function is a better name? I realize that this is done at other points
> in the code but the way it is now makes me think it's a class.

I had not noticed how inconsistent Firedrake naming is, in this regard.  Just looked at `firedrake/{mesh|utility_meshes}.py` as examples.  I think you are right.  Thus the retitle on this PR.

danshapero:
I'll ask about the node ordering issue at the next Firedrake meeting. I'm sure you're welcome to come but they're already pretty early for me, and I'm a timezone east of you.

I think keeping the same number of components when you extrude a vector field -- so a 2-component vector field on a 2D domain gets extruded to a 2-component vector field on a 3D domain -- makes the most sense. In principle you can define vector elements that have more components than the dimension of the space they live in. It's awfully weird and whatever it is won't transform like a rank-1 tensor but that's what the API allows. I'm also guessing that we'll both be solving the Blatter-Pattyn equations so let's be sure to make that as easy as possible ;)

bueler:
> I'm also guessing that we'll both be solving the Blatter-Pattyn equations so let's be sure to make that as easy as possible ;)

I'm not going there.

firedrake/extrusion_function.py
```diff
@@ -0,0 +1,142 @@
...
+        fextended = Function(Q)
+
+    if extend is not None:
+        if extend in {"top", "bottom"}:
+            bc = DirichletBC(fcn.function_space(), 1.0, extend)
+            fextended.dat.data_with_halos[:] = fcn.dat.data_with_halos[bc.nodes]
+        else:
+            raise ValueError('unknown extend_type')
+    else:
+        fextended.dat.data[:] = fcn.dat.data_ro[:]
```
bueler, line 139: @wence- @danshapero   I don't know how to do this line in the case of mixed elements.

wence-: Ah sorry.

You can do the following in both mixed and non-mixed:
```python
for ext, base in zip(fextended.split(), fcn.split()):
    ext.dat.data[:] = base.dat.data_ro[:]
```

wence-:
> Silly nitpick, but maybe extrude_function is a better name? I realize that this is done at other points in the code but the way it is now makes me think it's a class.

For posterity. We have `CapWords` functions in some cases for historical compatibility with dolfin. They are generally used when they are "functional" constructors that dispatch to a class-based `__init__` (e.g. MeshHierarchy/ExtrudedMeshHierarchy).

Generally functions should be `lower_case` named (as per PEP8 recommendations). We are not perfectly consistent here.

wence-:
> I'll ask about the node ordering issue at the next Firedrake meeting. I'm sure you're welcome to come but they're already pretty early for me, and I'm a timezone east of you.

Node ordering is guaranteed (due to the data structures on extruded meshes) to be the same for `foo x R` and `foo`.

> I think keeping the same number of components when you extrude a vector field -- so a 2-component vector field on a 2D domain gets extruded to a 2-component vector field on a 3D domain -- makes the most sense. In principle you can define vector elements that have more components than the dimension of the space they live in. It's awfully weird and whatever it is won't transform like a rank-1 tensor but that's what the API allows. I'm also guessing that we'll both be solving the Blatter-Pattyn equations so let's be sure to make that as easy as possible ;)

This is not what my expand-element function currently does. One would need to modify it a bit.

danshapero:
> > I think keeping the same number of components when you extrude a vector field -- so a 2-component vector field on a 2D domain gets extruded to a 2-component vector field on a 3D domain -- makes the most sense. In principle you can define vector elements that have more components than the dimension of the space they live in. It's awfully weird and whatever it is won't transform like a rank-1 tensor but that's what the API allows. I'm also guessing that we'll both be solving the Blatter-Pattyn equations so let's be sure to make that as easy as possible ;)
> 
> This is not what my expand-element function currently does. One would need to modify it a bit.

But do you think what I described is the correct behavior?

dham:
Seems moribund.

PR #1867, opened 2020-10-14
Fix complex merge test mistakes

ReubenHill:


ksagiyam:
It seems it's not broken anymore. Maybe just update this line `problem = LinearVariationalProblem(u*v*dx, 2*v*dx, self.uh)` in `NonePC` class for complex for consistency?

djanekovic:
Hmm, this is weird because this test fails for me when I just remove skipif decorator, I just tried it locally. Why did the test pass on Jenkins?

ReubenHill:
It fails for me too in real mode. I presume it's a test that's meant to run at a particular time?

ReubenHill:
@ksagiyam see 1e3deb64ecbfa18be7a747677fc374cf41b24ccc - hopefully I got this correct.

ksagiyam:
I can not reproduce it locally on my Ubuntu. @djanekovic and @ReubenHill What system do you run Firedrake on?

djanekovic:
Locally I am running firedrake on Debian. I agree that the commit 1e3deb6 fixed the problem but I am wondering why the test passed on Jenkins when it was obviously broken. I looked the log for 3bcbeb9 build and you can clearly see that Jenkins ran the test and it passed where it should fail with ArityMismatch.
```
[2020-10-14T20:00:40.014Z] tests/regression/test_appctx_cleanup.py::test_appctx_cleanup 
... 
[2020-10-14T20:00:44.150Z] [gw2] [ 57%] PASSED tests/regression/test_appctx_cleanup.py::test_appctx_cleanup
```

ReubenHill:
macos here - if memory serves the test crashed rather than failed for me.

ksagiyam:
If I run this test with `-s` without 1e3deb6, I can actually see the `ArityMismatch` error, but the test is still regarded as passed.  I think this is the problem. With 1e3deb6, the test passes without any error.

ReubenHill:
Weirdly `tests/regression/test_aw.py::test_aw[conforming]` fails on Jenkins on the complex PR merge test of 1e3deb6.

sv2518:
The test fails also in the PR #1884

wence- (approved):
I think the changes are fine. Are we happy with the testing situation on jenkins?

PR #1870, opened 2020-10-15
update .mailmap and AUTHORS

ReubenHill:


.mailmap
```diff
@@ -35,3 +35,5 @@ Tianjiao Sun <tianjiao.sun14@imperial.ac.uk> <ts2914@doc-ts2914.lib.ic.ac.uk>
 Tianjiao Sun <tianjiao.sun14@imperial.ac.uk> <tj.sun@hotmail.com>
 Florian Wechsung <f.wechsung@gmail.com>
 Florian Wechsung <f.wechsung@gmail.com> <wechsung@maths.ox.ac.uk>
+Reuben Nixon-Hill <reuben.nixon-hill10@imperial.ac.uk> Reuben Hill <ReubenHill@users.noreply.github.com>
+Reuben Nixon-Hill <reuben.nixon-hill10@imperial.ac.uk> Reuben Nixon-Hill <ReubenHill@users.noreply.github.com>
```
wence-, line 39: I think you can simplify this to 
```
Reuben Nixon-Hill <reuben.nixon-hill10@imperial.ac.uk> <ReubenHill@users.noreply.github.com>
```

because  the email is the same in both cases, that catches everything.

wence-: BTW, feel free to name your branches as `rnh/...` (or something else) rather than `ReubenHill/...` it's just an identifier so we can distinguish things (and needn't match the github username)

ReubenHill: I tried that initially, but it didn't combine the two names. The `git --help shortlog` documentation suggests this is correct (and it correctly combines my commits when I do `git shortlog -sne | grep Reuben`

wence-: Huh, odd, since my reading of `git help shortlog`:
```
        Proper Name <proper@email.xx> <commit@email.xx>

       which allows mailmap to replace both the name and the email of a commit
       matching the specified commit email address
```

suggests it should work?

ReubenHill: You are right - I must have been doing something wrong before.

dham (approved):

PR #1878, opened 2020-10-15
Update saddle point demo

sv2518:
Change saddle point demo to use LinearVariationalSolver rather than LinearSolver. Then one can use an AuxiliaryOperator to feed the Laplacian as a python preconditioner (in the section for the self-supplied approximate Schur complement).

### DIFF
--- a/demos/saddle_point_pc/saddle_point_systems.py.rst
+++ b/demos/saddle_point_pc/saddle_point_systems.py.rst
@@ -471,13 +471,13 @@ variable. We can provide it as an :class:`~.AuxiliaryOperatorPC` via a python pr
             gamma = Constant(8.0)
             h = CellSize(W.mesh())
             h_avg = (h('+') + h('-'))/2
-            a_dg = + inner(grad(u), grad(v))*dx \
+            a_dg = -(inner(grad(u), grad(v))*dx \
                 - inner(jump(u, n), avg(grad(v)))*dS \
                 - inner(avg(grad(u)), jump(v, n), )*dS \
                 + alpha/h_avg * inner(jump(u, n), jump(v, n))*dS \
                 - inner(u*n, grad(v))*ds \
                 - inner(grad(u), v*n)*ds \
-                + (gamma/h)*inner(u, v)*ds
+                + (gamma/h)*inner(u, v)*ds)
             bcs = None
             return (a_dg, bcs)
   
@@ -508,13 +508,13 @@ approximation we used above.
  Mesh elements  GMRES iterations
 ============== ==================
     2                 2
-    8                 10
-    32                17
-    128               19
-    512               20
-    2048              20
-    8192              19
-    32768             19
+    8                 9
+    32                12
+    128               13
+    512               14
+    2048              13
+    8192              13
+    32768             13
 ============== ==================
 
 Block diagonal preconditioners
--- a/demos/saddle_point_pc/saddle_point_systems.py.rst
+++ b/demos/saddle_point_pc/saddle_point_systems.py.rst
@@ -471,13 +471,13 @@ variable. We can provide it as an :class:`~.AuxiliaryOperatorPC` via a python pr
             gamma = Constant(8.0)
             h = CellSize(W.mesh())
             h_avg = (h('+') + h('-'))/2
-            a_dg =  (inner(grad(v), grad(u))*dx 
-            - dot(avg(grad(v)), jump(u, n))*dS \
-            - dot(jump(v, n), avg(grad(u)))*dS \
-            + alpha/h_avg * dot(jump(v, n), jump(u, n))*dS \
-            - dot(grad(v), u*n)*ds \
-            - dot(v*n, grad(u))*ds \
-            + (gamma/h)*dot(v, u)*ds)
+            a_dg = + inner(grad(u), grad(v))*dx \
+                - inner(jump(u, n), avg(grad(v)))*dS \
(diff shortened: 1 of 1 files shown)


colinjcotter:
This looks great! It is strange that the iteration counts went up though. Is it because you changed the tolerances?

sv2518:
> This looks great! It is strange that the iteration counts went up though. Is it because you changed the tolerances?

I don't think that would be the reason why. For the previous set of parameters I get the same amount of iterations. The only thing that changed is the incorporation of the aux pc and I don't think we set any default for that.

sv2518:
Okay, I screwed up the Laplacian, made it negative and changed the IP penalty parameters, now we get the same number of iterations. I think this is good to go now.

demos/saddle_point_pc/saddle_point_systems.py.rst
```diff
@@ -466,72 +460,61 @@ Instead of asking PETSc to build an approximation to :math:`S` which
...
+
+    class DGLaplacian(AuxiliaryOperatorPC):
+        def form(self, pc, u, v):
+            W = u.function_space()
+            n = FacetNormal(W.mesh())
+            alpha = Constant(8.0)
+            gamma = Constant(8.0)
+            h = CellSize(W.mesh())
+            h_avg = (h('+') + h('-'))/2
+            a_dg =  -(inner(grad(v), grad(u))*dx
```
wence-, line 474: I think you don't want this minus sign?

wence-: I think this form (if the old code is correct) should be
```
               + dot(grad(v), grad(u))*dx \
                - dot(avg(grad(v)), jump(u, n))*dS \
                - dot(jump(v, n), avg(grad(u)))*dS \
                + alpha/h_avg * dot(jump(v, n), jump(u, n))*dS \
                - dot(grad(v), u*n)*ds \
                - dot(v*n, grad(u))*ds \
                + (gamma/h)*dot(v, u)*ds
```

And then I suppose we have to update for complex-safety so

```
               + inner(grad(u), grad(v))*dx \
                - inner(jump(u, n), avg(grad(v)))*dS \
                - inner(avg(grad(u)), jump(v, n), )*dS \
                + alpha/h_avg * inner(jump(u, n), jump(v, n))*dS \
                - inner(u*n, grad(v))*ds \
                - inner(grad(u), v*n)*ds \
                + (gamma/h)*inner(u, v)*ds
```

demos/saddle_point_pc/saddle_point_systems.py.rst
```diff
@@ -466,72 +460,61 @@ Instead of asking PETSc to build an approximation to :math:`S` which
...
-``pc_fieldsplit_off_diag_use_amat``. ::
-
+penalty DG formulation for the Laplacian term on the block of the scalar
+variable. We can provide it as an AuxiliaryOperator via a python preconditioner. ::
+
+    class DGLaplacian(AuxiliaryOperatorPC):
+        def form(self, pc, u, v):
+            W = u.function_space()
+            n = FacetNormal(W.mesh())
+            alpha = Constant(8.0)
```
wence-, line 470: This wants to be 4, not 8.

sv2518:
If I do all that I get higher solver iterations than before
```
2 2
8 12
32 23
128 26
512 28
2048 26
8192 26
32768 24
```

If I keep the negative sign (but change the penalty parameters back) I get lower solver iterations than before
```
2 2
8 9
32 11
128 13
512 13
2048 13
8192 13
32768 13
```

demos/saddle_point_pc/saddle_point_systems.py.rst
sv2518: Okay, we do want the negative sign. The way the flux is introduced to bring the problem into mixed form, makes the the operator of u here negative definite. (Thanks @colinjcotter!)

demos/saddle_point_pc/saddle_point_systems.py.rst
```diff
@@ -466,72 +460,61 @@ Instead of asking PETSc to build an approximation to :math:`S` which
...
 discontinuous space.  A natural choice is therefore to use an interior
-penalty DG formulation for the Laplacian term and provide it as
-:math:`Sp`.  Since this preconditioning matrix is block-diagonal, we
-then need to tell PETSc that it should use the original operator,
-rather than the preconditioning matrix, to compute the action of the
-off-diagonal blocks.  This is done using
-``pc_fieldsplit_off_diag_use_amat``. ::
-
+penalty DG formulation for the Laplacian term on the block of the scalar
+variable. We can provide it as an AuxiliaryOperator via a python preconditioner. ::
```
wence-, line 464: Can you link the `AuxiliaryOperator` 
```
:class:`~.AuxiliaryOperatorPC`
```

I think.

demos/saddle_point_pc/saddle_point_systems.py.rst
```diff
@@ -466,72 +460,61 @@ Instead of asking PETSc to build an approximation to :math:`S` which
...
-      8192           18
-      32768          18
+    2                 2
+    8                 12
+    32                23
+    128               26
+    512               28
+    2048              26
+    8192              26
+    32768             24
```
wence-, line 517: I got lost after all the back and forth here. Why do the iteration counts change?

sv2518: The Laplacian has to be negative, because the Schur complement is negative definite for this problem.

wence-:

Issue #1871, opened 2020-10-15
Bug in enforcing DirichletBC for extruded mesh


jingminxia:
Consider the following code snippet where I want to enforce strong boundary conditions on all boundaries of an extruded mesh:

```
from firedrake import *
N = 10
base = SquareMesh(N, N, 1, quadrilateral=True)
baseh = MeshHierarchy(base, 1)
mh = ExtrudedMeshHierarchy(baseh, height=2, base_layer=N)
mesh = mh[-1]
V = FunctionSpace(mesh, "CG", 1)
u = Function(V)
DirichletBC(V, Constant(1), ["on_boundary", "top", "bottom"]).apply(u)
```

It gives an error

```
Traceback (most recent call last):
  File "dirichletbc.py", line 13, in <module>
    DirichletBC(V, Constant(1), ["on_boundary", "top", "bottom"]).apply(u)
  File "<decorator-gen-45>", line 2, in apply
  File "/home/jingmin/installed/firedrake/src/PyOP2/pyop2/profiling.py", line 60, in wrapper
    return f(*args, **kwargs)
  File "/home/jingmin/installed/firedrake/src/firedrake/firedrake/adjoint/dirichletbc.py", line 32, in wrapper
    ret = apply(self, *args, **kwargs)
  File "/home/jingmin/installed/firedrake/src/firedrake/firedrake/bcs.py", line 423, in apply
    r.assign(self.function_arg, subset=self.node_set)
  File "/home/jingmin/installed/firedrake/src/PyOP2/pyop2/utils.py", line 62, in __get__
    obj.__dict__[self.__name__] = result = self.fget(obj)
  File "/home/jingmin/installed/firedrake/src/firedrake/firedrake/bcs.py", line 180, in node_set
    return op2.Subset(self._function_space.node_set, self.nodes)
  File "/home/jingmin/installed/firedrake/src/PyOP2/pyop2/utils.py", line 62, in __get__
    obj.__dict__[self.__name__] = result = self.fget(obj)
  File "/home/jingmin/installed/firedrake/src/firedrake/firedrake/bcs.py", line 165, in nodes
    raise TypeError("Currently, edge conditions have only been tested with Lagrange elements")
TypeError: Currently, edge conditions have only been tested with Lagrange elements
```

However, if I separately enforce DirichletBC on different boundaries, it runs with no error.

```
DirichletBC(V, Constant(1), "on_boundary").apply(u)
[...]

PR #1879, opened 2020-10-16
fix: enable list of string bcs

ksagiyam:
This PR is to enable boundary conditions specified in a list-of-strings form and fix https://github.com/firedrakeproject/firedrake/issues/1871.


wence- (changes requested):
Code looks fine, could you add a test please?

ksagiyam:
Added a test. Thanks!

wence- (approved):

PR #1884, opened 2020-10-19
Enable slate tensors with variable-layers extruded mesh

stephankramer:
This just removes two asserts that claim slate does not work with
variable layers - but this has presumably been fixed in #1715
Adds test for assembly of mass and inverse mass matrix on such mesh.

### DIFF
--- a/tests/slate/test_scalar_tensors_extr.py
+++ b/tests/slate/test_scalar_tensors_extr.py
@@ -1,4 +1,5 @@
 import numpy as np
+from math import ceil
 
 from firedrake import *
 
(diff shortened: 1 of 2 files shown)


tests/slate/test_scalar_tensors_extr.py
```diff
@@ -7,3 +8,54 @@ def test_constant_one_tensor():
...
+    A1.axpy(-1, A2)
+    assert A1.norm() < 1e-16
+
+    # check A2*A3==Identity
+    C = A2.matMult(A3)
+    Cd = C.getDiagonal()
+    np.testing.assert_allclose(Cd.array, 1)
+    Cd.set(0.)
+    C.setDiagonal(Cd)
+    np.testing.assert_allclose(C.getRowSum().array, 0, atol=1e-16)
```
sv2518, line 61: Maybe this can be shortened from line 47 to 
```
A1 = assemble(Tensor(v*u*dx)).M.values
A2 = assemble(v*u*dx).M.values
A3 = assemble(Tensor(v*u*dx).inv).M.values

# check A1==A2
assert np.allclose(A1, A2, rtol=1e-12)

# check A2*A3==Identity
assert np.allclose(np.matmul(A2, A3), np.eye(*A2.shape), rtol=1e-12)
```

sv2518 (commented):
Looks like the test is passing on real. Jenkins only moans about another test on complex, which should be unrelated to this, I think.

wence-:
Can you rebase onto trunk again? I think the final false failure is now fixed there.

wence-:

wence- (approved):

Issue #1839, opened 2020-09-14
PETSc installation --download-fblaslapack


jfregin:
Hello,

is there any downside installing Blas/Lapack by adding `--download-fblaslapack` to the `petsc_options`? I'm asking because the install instructions on the firedrake webpage ask to have Blas/Lapack installed when installing without package manager. If there is no downside, why is this not the default option? Below is some explanation on how I came up with the question.

I succesfully installed firedrake on openSUSE, however I had a few issues during the installation. Mainly because i failed to get Blas/Lapack running and therefore the PETSc installation failed. Then I tried building PETSc myself with the configure options that firedrake suggests which worked fine (additionally I added `--download-fblaslapack` ). However, for some reason the firedrake installation still failed (I don't have the log files anymore).
I resolved the problem by just adding `--download-fblaslapack` to the `petsc_options` in the firedrake-install script script. Afterwards the installation was successful. 

Kind regards,

Joscha

wence-:
`--download-fblaslapack` gives you a reference version of BLAS/LAPACK (from netlib). While this will work, it is low performance. We therefore recommend that you install a high-performance library providing this (often `openblas` will work). 

We are likely to migrate to installing openblas and getting PETSc to build against that in the new future, so it would be good to know (if you can reproduce) what went wrong with the system provided blas/lapack that you tried.

jfregin:
Thank you. The "system provided" blas/lapack was the one that I installed with PETSc. I can rerun the installation using blas/lapack from netlib today, to see what went wrong. I'd also gladly test the installation with openblas, but it will take a while as I have only user rights on my working machine and the administrator is on holiday. I also tried compiling blas/lapack myself, however only blas seems to work. With lapack I get a segmentation fault and unfortunately the solution to this problem (hopefully) requires my administrator.

wence-:
This migration has now been completed by #1885, needs a fresh install with a newly download `firedrake-install` script. Can you confirm this works?

jfregin:
Install was succsessful with --with-openblas=download. Not sure if that was what you wanted me to try.

Issue #1828, opened 2020-08-30
make alltest SyntaxError after installing firedrake on Mac


plankering:
I am trying to install Firedrake on my macOS Catalina Version 10.15.6 laptop but I get the following error when I run "make alltest" after installation : 

    Building extension modules
  File "setup.py", line 51
    dirs = (sys.prefix, *petsc_dirs)
                        ^
SyntaxError: invalid syntax
    Linting firedrake codebase
/System/Library/Frameworks/Python.framework/Versions/2.7/Resources/Python.app/Contents/MacOS/Python: No module named flake8
make: *** [lint] Error 1

I followed all the prior instructions including installing Homebrew, making sure I had python3, and then install firedrake as instructed on the 'Obtaining Firedrake' of your website. The following is the firedrake-install.log file. Any help would be greatly appreciated. Thank you.

[firedrake-install.log](https://github.com/firedrakeproject/firedrake/files/5146409/firedrake-install.log)

miklos1:
You are apparently running this with Python 2.7.

Did you forget to activate the virtual env before `make alltest`?

plankering:
oh yes, I did forget the activation step before the make alltest. Thanks for that. Nevertheless, I still get the following error
taking that step into account.

    Building extension modules
Traceback (most recent call last):
  File "setup.py", line 8, in <module>
    import numpy as np
ModuleNotFoundError: No module named 'numpy'
    Linting firedrake codebase
/Users/romeo/firedrake/bin/python: No module named flake8
make: *** [lint] Error 1

I did the "python3 --version" check and it shows that I have Python 3.8.5 installed. My guess therefore is that, I have both versions installed. How do I modify the "python3 firedrake-install" commend to forcibly run with Python 3? Thanks. 
P.S. A novice here so sorry if the question sounds trivial.

miklos1:
`python3 firedrake-install` will run Python 3, because that's made explicit.  Once you activated the virtual env, even `python` should refer to the Python 3 used in the virtual env.

But it seems mysterious to me, why the virtualenv python fails to find numpy and flake8... they should be installed inside the virtualenv.
- I also notice from the logfile that you have anaconda installed.  We used to have issues with anaconda python in the past, not sure what the current state is...

plankering:
[...]

Issue #1803, opened 2020-08-14
MPI Issues on macOS


edaub:
I just did a fresh Firedrake install on macOS Catalina using the provided install script. My installation works fine on a single processor, but I am having issues running multiple processes where the scripts hang and never run to completion.

I have confirmed that the `mpi4py` installation is fine -- in the venv, I can run the following under multiple processes no problem:

```
from mpi4py.MPI import COMM_WORLD
print(COMM_WORLD.rank)
```

However, the Firedrake equivalent hangs with two or more processes:

```
from firedrake import COMM_WORLD
print(COMM_WORLD.rank)
```

Any idea what could be causing this? Log file from installation is attached.
[firedrake-install.log](https://github.com/firedrakeproject/firedrake/files/5076556/firedrake-install.log)

edaub:
**Update:** Seems to be a PETSc issue, as I can't run the equivalent with `PETSc.COMM_WORLD` on two processes. Investigating the PETSc build.

wence-:
Our probable guess at the moment for these problems is that the numpy wheels ship with a version of libgfortran that is ABI-incompatible with the one you get in homebrew. @JDBetteridge is working on this. Perhaps he can describe if there is currently a workaround.

Issue #1800, opened 2020-08-05
Importing ufl before firedrake wildcard makes mumps fail to initialize


salazardetroya:
```python
import ufl
from firedrake import *

mesh = UnitSquareMesh(10, 10)
V = FunctionSpace(mesh, 'CG', 1)
v = TestFunction(V)
u = TrialFunction(V)
a = inner(grad(u), grad(v)) * dx
L = - 1.0 * v * dx
bc = DirichletBC(V, 0.0, "on_boundary")

parameters = {
        "ksp_type": "preonly",
        "pc_type": "lu",
        "pc_factor_mat_solver_type": "mumps",
        }

u_sol = Function(V)
solve(a==L, u_sol, solver_parameters=parameters)
```
causes
```
[0]PETSC ERROR: ------------------------------------------------------------------------
[0]PETSC ERROR: Caught signal number 11 SEGV: Segmentation Violation, probably memory access out of range
[0]PETSC ERROR: Try option -start_in_debugger or -on_error_attach_debugger
[0]PETSC ERROR: or see https://www.mcs.anl.gov/petsc/documentation/faq.html#valgrind
[0]PETSC ERROR: or try http://valgrind.org on GNU/linux and Apple Mac OS X to find memory corruption errors
[0]PETSC ERROR: likely location of problem given in stack below
[0]PETSC ERROR: ---------------------  Stack Frames ------------------------------------
[0]PETSC ERROR: Note: The EXACT line numbers in the stack are not available,
[0]PETSC ERROR:       INSTEAD the line number of the start of the function
[0]PETSC ERROR:       is given.
[0]PETSC ERROR: [0] PetscInitializeMUMPS line 1849 /Users/salazardetro1/scicomp_libraries/firedrake-debug/firedrake/src/petsc/src/mat/impls/aij/mpi/mumps/mumps.c
[0]PETSC ERROR: [0] MatGetFactor_aij_mumps line 2985 /Users/salazardetro1/scicomp_libraries/firedrake-debug/firedrake/src/petsc/src/mat/impls/aij/mpi/mumps/mumps.c
[0]PETSC ERROR: [0] MatGetFactor line 4511 /Users/salazardetro1/scicomp_libraries/firedrake-debug/firedrake/src/petsc/src/mat/interface/matrix.c
[...]

PR #1885, opened 2020-10-21
Numpy fix

JDBetteridge:
This PR builds everything using the same BLAS/LAPACK libraries, including numpy and loopy.
As a result more packages are _built_ when installing with `pip`, rather than installing from wheel.

This PR requires the Firedrake fork of h5py to be bumped forward before being merged.

Merging this fix should close the following open issues:
#1839
#1828
#1803
#1801
#1800
#1784
#1600


JDBetteridge:
https://xkcd.com/1987/

wence- (approved):
Thanks!

PR #1890, opened 2020-10-28
Make Gouraud shading default for tripcolor unless DG0

danshapero:
The tripcolor function currently defaults to flat shading, which uses a single color for each triangle. This patch makes tripcolor use flat shading only for DG(0) spaces. For everything else, tripcolor will default to Gouraud shading, which uses a linear color gradient in each cell. This gives much better plots for DG(1) or CG(1) or higher functions.


wence- (approved):
Thanks.

PR #1892, opened 2020-10-31
Fix ams setup

### LINKED ISSUE #1886: strange AMS setup failure
After the new AMS preconditioner has been merged into firedrake, I rerun some of my older simulations, with the adapted code.
For some reason I now get the following error, during PC setup:
```
/firedrake/src/fiat/FIAT/check_format_variant.py:10: DeprecationWarning: Variant of Nedelec element will change from point evaluation to integral evaluation. You should project into variant="integral"
  ' You should project into variant="integral"', DeprecationWarning)
N: 50061
NE: 292629
dofs: 342955
  0 SNES Function norm 1.164177949606e-01
  File "libpetsc4py/libpetsc4py.pyx", line 1298, in libpetsc4py.PCSetUp_Python                                             
  File "/firedrake/src/firedrake/firedrake/preconditioners/base.py", line 45, in setUp                                     
    self.initialize(pc)
  File "/host/home/florian/work/2020/eddy_current/team13/linear/firedrake/medium/gtos/MyHypreAMS_preconditioner.py", line 156, in initialize
    c = project(Constant(direction), V)
  File "/firedrake/src/firedrake/firedrake/adjoint/projection.py", line 29, in wrapper                                     
    output = project(*args, **kwargs)
  File "/firedrake/src/firedrake/firedrake/projection.py", line 92, in project 
    use_slate_for_inverse=use_slate_for_inverse).project()                                                                 
  File "/firedrake/src/firedrake/firedrake/projection.py", line 170, in project                                            
    self.apply_massinv(self.target, self.rhs)
  File "/firedrake/src/firedrake/firedrake/linear_solver.py", line 160, in solve                                           
    raise ConvergenceError("LinearSolver failed to converge after %d iterations with reason: %s", self.ksp.getIterationNumber(), solving_utils.KSPReasons[r])
firedrake.exceptions.ConvergenceError: ('LinearSolver failed to converge after %d iterations with reason: %s', 0, 'DIVERGED_PCSETUP_FAILED')
```

Rerunning the original standalone code (but with a recent firedrake version) also leads to the same error. Furthermore the code still works for a coarse mesh, whereas medium and fine mesh fails. 

For now I have the following questions:
- are there any changes in Firedrake which could explain this problem
- is it possible to install an older version of Firedrake using the installer? how can I find out the working version?
- any advice how to debug this problems are welcome

Thanks in advance
Florian

florian98765:
now using setCoordinate instead of setEdgeConstantVector
fix #1886

### DIFF
--- a/firedrake/preconditioners/hypre_ams.py
+++ b/firedrake/preconditioners/hypre_ams.py
@@ -1,0 +1,72 @@
+from firedrake.preconditioners.base import PCBase
+from firedrake.petsc import PETSc
+from firedrake.functionspace import FunctionSpace
+from firedrake.constant import Constant
+from firedrake.ufl_expr import TestFunction
+from firedrake.interpolation import Interpolator
+from firedrake.projection import project
+from firedrake.dmhooks import get_function_space
+from firedrake.utils import complex_mode
+from firedrake_citations import Citations
+from ufl import grad
+
+__all__ = ("HypreAMS",)
+
+
+class HypreAMS(PCBase):
+    def initialize(self, obj):
+        if complex_mode:
+            raise NotImplementedError("HypreAMS preconditioner not yet implemented in complex mode")
+
+        Citations().register("Kolev2009")
+        A, P = obj.getOperators()
+        prefix = obj.getOptionsPrefix()
+        V = get_function_space(obj.getDM())
+        mesh = V.mesh()
+
+        family = str(V.ufl_element().family())
+        degree = V.ufl_element().degree()
+        if family != 'Nedelec 1st kind H(curl)' or degree != 1:
+            raise ValueError("Hypre AMS requires lowest order Nedelec elements! (not %s of degree %d)" % (family, degree))
+
+        P1 = FunctionSpace(mesh, "Lagrange", 1)
+        G = Interpolator(grad(TestFunction(P1)), V).callable().handle
+
+        pc = PETSc.PC().create(comm=obj.comm)
+        pc.incrementTabLevel(1, parent=obj)
+        pc.setOptionsPrefix(prefix + "hypre_ams_")
+        pc.setOperators(A, P)
+
+        pc.setType('hypre')
+        pc.setHYPREType('ams')
+        pc.setHYPREDiscreteGradient(G)
+        zero_beta = PETSc.Options(prefix).getBool("pc_hypre_ams_zero_beta_poisson", default=False)
+        if zero_beta:
+            pc.setHYPRESetBetaPoissonMatrix(None)
+
+        # Build constants basis for the Nedelec space
+        cvecs = []
+        for i in range(mesh.cell_dimension()):
+            direction = [1.0 if i == j else 0.0 for j in range(mesh.cell_dimension())]
+            c = project(Constant(direction), V)
+            with c.vector().dat.vec_ro as cvec:
+                cvecs.append(cvec)
+        pc.setHYPRESetEdgeConstantVectors(*cvecs)
+        pc.setUp()
+
+        self.pc = pc
+
+    def apply(self, pc, x, y):
+        self.pc.apply(x, y)
+
+    def applyTranspose(self, pc, x, y):
+        self.pc.applyTranspose(x, y)
+
+    def view(self, pc, viewer=None):
+        super(HypreAMS, self).view(pc, viewer)
+        if hasattr(self, "pc"):
+            viewer.printfASCII("PC to apply inverse\n")
+            self.pc.view(viewer)
+
+    def update(self, pc):
+        self.pc.setUp()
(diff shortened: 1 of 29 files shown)


wence-:
I think the test failure is unrelated (and is hopefully fixed by #1896).

firedrake/preconditioners/hypre_ams.py
```diff
@@ -44,14 +43,8 @@ def initialize(self, obj):
...
-        # Build constants basis for the Nedelec space
-        cvecs = []
-        for i in range(mesh.cell_dimension()):
-            direction = [1.0 if i == j else 0.0 for j in range(mesh.cell_dimension())]
-            c = project(Constant(direction), V)
-            with c.vector().dat.vec_ro as cvec:
-                cvecs.append(cvec)
-        pc.setHYPRESetEdgeConstantVectors(*cvecs)
+        VectorP1 = VectorFunctionSpace(mesh, "Lagrange", 1)
+        pc.setCoordinates(interpolate(SpatialCoordinate(mesh), VectorP1).dat.data_ro)
```
wence-, line 47: I would prefer to take a copy of the coordinates and drop the `Function` (so tack a `.copy()` on the end here).

tests/regression/test_hypre_ams.py
```diff
@@ -99,9 +99,44 @@ def test_homogeneous_field_nonlinear():
...
+    B0 = 1
+    constant_field = as_vector([-0.5*B0*(y - 0.5), 0.5*B0*(x - 0.5), 0])
+
+    bc = DirichletBC(V, constant_field, (1, 2, 3, 4))
+
+    params = {'snes_type': 'ksponly',
+              'ksp_type': 'cg',
+              'ksp_max_it': '30',
+              'ksp_rtol': '1e-8',
+              'ksp_atol': '1e-8',
```
wence-, line 132: I don't really like the `abstol` restriction (since if it so happens that the residual starts out small you might hide something here). Prefer not setting it at all.

firedrake/preconditioners/hypre_ams.py
```diff
@@ -44,14 +43,8 @@ def initialize(self, obj):
...
             pc.setHYPRESetBetaPoissonMatrix(None)
 
-        # Build constants basis for the Nedelec space
-        cvecs = []
-        for i in range(mesh.cell_dimension()):
-            direction = [1.0 if i == j else 0.0 for j in range(mesh.cell_dimension())]
-            c = project(Constant(direction), V)
-            with c.vector().dat.vec_ro as cvec:
-                cvecs.append(cvec)
-        pc.setHYPRESetEdgeConstantVectors(*cvecs)
```
wence-, line 54: I guess you did this because the convergence is better? Is that right?

wence- (changes requested):

wence-:
Thanks! I hope that my change in #1896 will fix the strange test failure here.

firedrake/preconditioners/hypre_ams.py
florian98765: For some reason, project leads to a `DIVERGED_PCSETUP_FAILED` error for a fine mesh. It works with an older version of firedrake and still works for a coarse mesh. When I originally implemented the code, interpolation on edge elements was not implemented, if i remember correctly. Furthermore, using the coordinates led to bad convergence. I think the problem was that the dofs were not ordered correctly. 
Now that setCoordinates yields good convergence, I think this should be the preferred version, at least for the lowest-order case.

wence-: OK thanks. I suspect the PCSETUP_FAILED divergence is because we switched to using a direct solver by default (even for projections) and so for very big problems that might cause OOM or other issues (see #1900).

tests/regression/test_hypre_ams.py
```diff
@@ -96,12 +94,45 @@ def test_homogeneous_field_nonlinear():
...
               'pc_python_type': 'firedrake.HypreAMS',
-              'pc_hypre_ams_zero_beta_poisson': True,
+              'pc_hypre_ams_zero_beta_poisson': True
               }
 
     solve(a - L == 0, u, bc, solver_parameters=params)
     B = project(curl(u), V0)
     assert numpy.allclose(B.dat.data_ro, numpy.array((0., 0., 1.)), atol=1e-6)
+
+@pytest.mark.skipcomplex
```
wence-, line 106: Sorry, this needs an extra blank line to pass linting (you can check these errors locally with `make lint` in the firedrake src directory).

florian98765:
CI still reports
```
FAILED tests/regression/test_manifolds.py::test_no_manifold_parallel
```
any ideas? I already merged #1896

wence-:
Not sure what has gone on. Let me bring it up at the group meeting this afternoon.

wence-:
We've now merged the fix that hopefully squashes the occasional test failures. Can you merge master again and then I'll do a final review. Thanks for you patience!

florian98765:
> We've now merged the fix that hopefully squashes the occasional test failures. Can you merge master again and then I'll do a final review. Thanks for you patience!

Done! I just want to mention that the unrelated error has already disappeared before (see 17dc787).
Nevertheless thank you again. I hope now everything works as desired.

wence- (approved):
Fingers crossed that jenkins is happy!

PR #1896, opened 2020-11-03
tests: Need extra pivoting options to MUMPS in real mode too

wence-:
It seems this setting is needed in real mode too on MacOS at least.

(Fixes the error seen in https://jenkins.ese.ic.ac.uk/blue/organizations/jenkins/firedrake/detail/PR-1892/5/pipeline/, I hope).


JDBetteridge (approved):
The solver options look cleaner too, now the check for complex is removed

PR #1897, opened 2020-11-03
install: Only create MPI compiler wrapper links in install mode

### LINKED ISSUE #1891: firedrake-update relinking of mpi wrappers is fragile
If I use the system MPI, then the logic for resetting the MPI wrappers (if I did not explicitly say `--mpicc` etc...) creates a wrapper that just `execs` itself.

wence-:
In update mode, because the shutil.which resolves to the MPI wrapper
in the activated virtualenv, we would create a wrapper that just execs
itself. Fixes #1891.

### DIFF
--- a/scripts/firedrake-install
+++ b/scripts/firedrake-install
@@ -1422,26 +1422,27 @@ if mode == "install":
 
 petsc_dir, petsc_arch = get_petsc_dir()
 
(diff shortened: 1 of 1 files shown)


JDBetteridge (approved):
Looks good

PR #1898, opened 2020-11-04
Fix test_aw convergence error

connorjward:
The `test_aw[conforming]` test has been exhibiting some odd behaviour. On Jenkins it would occasionally fail on the complex branch (#1867) and on my machine (Ubuntu 18.04) it failed on both real and complex unless the `PYOP2_DEBUG=1` flag was added. The failure was a convergence error with convergence taking a few more iterations than the maximum limit.

The current theory is that the vector registers in Intel chips have a lower precision than the scalar ones leading to different convergence behaviour when optimisation is turned on.

Comments welcome. Are you all happy with this solution?

### DIFF
--- a/tests/regression/test_aw.py
+++ b/tests/regression/test_aw.py
@@ -90,7 +90,7 @@ def test_aw(stress_element):
                   "snes_linesearch_type": "basic",
                   "snes_monitor": None,
                   "mat_type": "aij",
-                  "snes_max_it": 10,
+                  "snes_max_it": 15,
                   "snes_lag_jacobian": -2,
                   "snes_lag_preconditioner": -2,
                   "ksp_type": "preonly",


rckirby:
I think this is fine for a fix, but it would be nice to undersrtand if the conjecture is correct

rckirby (approved):
looks fine for now

pefarrell (approved):

tests/regression/test_aw.py
```diff
@@ -90,7 +90,7 @@ def epsilon(u):
                   "snes_linesearch_type": "basic",
                   "snes_monitor": None,
                   "mat_type": "aij",
-                  "snes_max_it": 10,
+                  "snes_max_it": 15,
```
wence-, line 93: I think we wanted to do this only in complex mode:
```
from firedrake import utils

...
"snes_max_it": 15 if utils.complex_mode else 10
```

wence-:
The tests failed https://jenkins.ese.ic.ac.uk/blue/organizations/jenkins/firedrake/detail/PR-1898/1/pipeline

In `tests/regression/test_expressions.py::test_assign_to_mfs_sub`

Which is definitely not on the code path that this code touches.

What has gone wrong (on Jenkins/in Firedrake)?

tests/regression/test_aw.py
connorjward: Since I was seeing the error in both modes David suggested I apply it to both.

wence-: Ah OK.

wence-:

ReubenHill:
No longer necessary

PR #1902, opened 2020-11-05
Actually calculate the L_infty error, not just a max.

bueler:
I don't know what I was thinking.  Now it actually computes an error not just a maximum of the solution.

Nonetheless this is minor in the sense that probably no one cared whether the number printed meant anything.


wence- (approved):
Thanks!

PR #1904, opened 2020-11-05
Wence/feature/superludist default

wence-:
Trying with default superlu dist (contains a commit from #1884 that appears to trigger the jenkins issues), so DO NOT MERGE YET.


wence-:
Not necessary.

PR #1905, opened 2020-11-05
Wence/fix/docker blas

wence-:
Actually install openblas in the docker container, and set some env vars in jenkins.


JDBetteridge (approved):
Looks good

JDBetteridge:
I think this PR will need re-basing before it can be merged, as conflicts have now arisen

JDBetteridge:
Merged into #2047

PR #1909, opened 2020-11-07
Plotting refactor

danshapero:
Summary: With the current plotting functions, it's almost impossible to make movies of evolving finite element simulations using matplotlib in a performant way. This patch introduces a helper class for repeatedly evaluating Functions from a particular space at the set of plotting sample points. The refactoring also cuts down on the amount of code in this module.

The problem: The `tripcolor` and `trisurf` functions (`tricontour` isn't suitable for movies anyway) internally create a refined matplotlib Triangulation object, evaluate a Function at all the points of that triangulation, and draw the result. The cell size of this triangulation depends on how many sample points you ask for, so for example a single triangle of a Firedrake Mesh could correspond to 4, 16, etc. triangles of the mesh that actually gets plotted. The problem with the old interface is that this matplotlib triangulation gets thrown out after every plot call. So doing repeated plot calls is extremely wasteful.

The solution: This patch adds a class ~~that for now I'm calling ValueCalculator (we can bikeshed on the name, I couldn't think of anything good)~~ called FunctionPlotter. It takes in a ~~FunctionSpace~~ Mesh and pre-calculates the things that you need to repeatedly plot fields ~~from that space~~ defined on that mesh -- the sample points, the matplotlib triangulation, etc. It then has a `__call__` method for evaluating functions in that space at the sample points. The important data gets saved instead of re-calculated. The existing plotting functions all work exactly the same as before from the user's perspective, but they now use a FunctionPlotter under the hood.

Demonstration: [This gist](https://gist.github.com/danshapero/79ad7e893a2d0749925e7417126f3195) contains some sample code that makes a movie of the solution of the advection equation. Running this script with

    python3 firedrake-movie-test.py --method good

uses the new functionality I just added and takes about 20s on my desktop. Running with `--method bad` does the naive thing and calls `tripcolor` on every frame. It ran for about 20min before I noticed that my system had frozen and I had to restart.

This can work with 3D functions, but there's an extra step because matplotlib only takes cell-centered data for 3D surface plots. I've also managed to reduce some spaghetti in the 1D plotting code but there's still more work to be done there. Just wanted to make a draft and get feedback first.


wence-:

wence-:
@danshapero is this now good, or still draft? I think a rebase onto trunk would probably fix the spurious (?) test failure.

danshapero:
Sorry Lawrence I got side-tracked from this -- there are still some improvements to make for the 3D case. If you want to keep this patch smaller I could save that for another PR, or I could get them all done in this one. Thoughts?

Might also be good to bikeshed on the names some.

firedrake/plot.py
```diff
@@ -895,14 +778,71 @@ def _interp_bezier(pts, num_cells, axes, complex_component="real", **kwargs):
...
+            self.triangles = all_triangles
+
+    def __call__(self, function):
+        # TODO: Make this more efficient on repeated calls -- for example reuse `elem`
+        # if the function space is the same as the last one
+        Q = function.function_space()
+        dimension = Q.mesh().topological_dimension()
+        keys = {1: (0,), 2: (0, 0)}
+
+        fiat_element = create_base_element(Q.ufl_element()).fiat_equivalent
```
wence-, line 836: I think that `Q.finat_element` is the same as `create_base_element(Q.ufl_element())`

wence- (commented):
Do you mind adding a short section in the docs on how to use this?

wence-:
Otherwise looks good, I think.

danshapero:
I changed the DG advection demo from outputting to .pvd to instead create an .mp4 using the FunctionPlotter class and matplotlib. Several other demos already show how to output to .pvd, but I can also revert this create a separate demo or document it some other way if that would be best.

wence- (approved):
Thanks!

PR #1910, opened 2020-11-09
Don't assume the interpreter is called python3

dham:


wence-:

I think @JDBetteridge was going to comment about some other issues?

wence-:
How has this merge gone so badly wrong? Can you just start from trunk and cherry-pick the three relevant commits?

JDBetteridge:
> How has this merge gone so badly wrong? Can you just start from trunk and cherry-pick the three relevant commits?

No idea, have tried to recreate sequence of steps that would lead to this state unsuccessfully, opened  #1922 with relevant commits

wence-:
You're welcome to force push here so that we don't lose the review process.

PR #1913, opened 2020-11-11
DO NOT MERGE: Pre-complex merge

ReubenHill:
Head of this branch is set to the master parent of the complex merge commit.


wence-:
THis hung in the test suite (15 hours in)

ReubenHill:
Okay so that would suggest to me that the problem isn't directly firedrake-repository-change related. Next step try reverting another package, e.g. PETSc and see what happens?

PR #1915, opened 2020-11-11
Extra mumps options to avoid zero pivots

wence-:


dham (dismissed):

wence-:
OK, removed the gmres part and everything passed, will merge.

PR #1917, opened 2020-11-13
Ksagiyam/io hdf5 xdmf

ksagiyam:


firedrake/mesh.py
```diff
@@ -1263,13 +1263,11 @@ def cell_set(self):
 class MeshGeometry(ufl.Mesh, MeshGeometryMixin):
     """A representation of mesh topology and geometry."""
 
-    def __new__(cls, element):
+    def __new__(cls):
```
wence-, line 1275: instead take `element=None`?

firedrake/mesh.py
```diff
@@ -1263,13 +1263,11 @@ def cell_set(self):
...
         """Create mesh geometry object."""
         utils._init()
         mesh = super(MeshGeometry, cls).__new__(cls)
         mesh.uid = utils._new_uid()
-        assert isinstance(element, ufl.FiniteElementBase)
-        ufl.Mesh.__init__(mesh, element, ufl_id=mesh.uid)
```
wence-, line 1272: Guarded by `if element is not None`?

firedrake/mesh.py
```diff
@@ -777,6 +761,10 @@ def callback(self):
     def comm(self):
         return self._comm
 
+    @property
+    def topology_dm(self):
```
wence-, line 765: Why do we add this?

wence-: If we're making it public, why not just rename the slot to `topology_dm`?

firedrake/mesh.py
```diff
@@ -1294,6 +1330,32 @@ def init(self):
         if hasattr(self, '_callback'):
             self._callback(self)
 
+    def partial_init(self, topology):
+        """Partially initialise the mesh.
```
dham, line 1334: Please describe which part is initialised.

firedrake/mesh.py
```diff
@@ -894,34 +909,55 @@ def cell_to_facets(self):
...
-        eStart, eEnd = self._topology_dm.getDepthStratum(d)
+        eStart, eEnd = self.topology_dm.getDepthStratum(d)
         return eEnd - eStart
 
     @utils.cached_property
     def cell_set(self):
         size = list(self._entity_classes[self.cell_dimension(), :])
         return op2.Set(size, "Cells", comm=self.comm)
 
+    def set_partitioner(self, distribute):
```
wence-, line 940: A few people have been asking to be able to configure the partitioner. This also needs a docstring.

firedrake/mesh.py
```diff
@@ -1522,6 +1584,16 @@ def __dir__(self):
         current = super(MeshGeometry, self).__dir__()
         return list(OrderedDict.fromkeys(dir(self._topology) + current))
 
+    def view(self, fname, formt):
```
dham, line 1587: Rename as save. Also needs a doctoring. Please also spell out filename and format.

wence-, line 1587: Rename to `save`. We're not writing C so `filename` and `format` are fine.

wence-: Add docstring, what values can `format` take?

firedrake/mesh.py
```diff
@@ -1593,6 +1664,7 @@ def Mesh(meshfile, **kwargs):
     * Exodus: with extension `.e`, `.exo`
     * CGNS: with extension `.cgns`
     * Triangle: with extension `.node`
+    * HDF5: with extension `.h5`, `.hdf5`
```
wence-, line 1667: What type of HDF5 file?

tests/output/test_io_mesh.py
```diff
@@ -0,0 +1,42 @@
...
+import pytest
+from petsc4py import PETSc
+from petsc4py.PETSc import ViewerHDF5
+from pyop2.mpi import COMM_WORLD
+
+
+@pytest.mark.parallel(nprocs=4)
+def test_io_hdf5_xdmf():
+    # Parameters
+    fname = "./test_io_mesh_dump.h5"
```
wence-, line 11: Output should be to a temp directory check the `tmpdir` fixture I think.

tests/output/test_io_mesh.py
```diff
@@ -0,0 +1,42 @@
...
+    outformat = ViewerHDF5.Format.HDF5_XDMF
+    ntimes = 3
+    heterogeneous = True
```
wence-, line 14: Make this a parameter so we test both branches?

firedrake/mesh.py
```diff
@@ -306,6 +308,33 @@ def _from_triangle(filename, dim, comm):
     return plex
 
 
+def _from_hdf5(filename, comm):
+    comm = PETSc.Comm(comm=comm)
+    # Determine the format from an existing h5 group.
+    # TODO: Looking for `/topology` group may not be the best.
+    with h5py.File(filename, 'r') as f:
+        if '/topology' in f:
```
dham, line 316: We need to find a slightly safer way of sniffing the file type.

dham (changes requested):

firedrake/mesh.py
ksagiyam: Is there a way to configure partitioners via PETSc? Maybe I misunderstood your comment.

wence-: So we can run `-petscpartitioner_type foo` on the commandline, but that's not very discoverable. I guess here `set_partitioner` seems to imply that we can select it, but really we're just setting it up.

Maybe `set_partitioner(self, distribute, partitioner_type=None)` or something. Or we can save that for a separate round.

firedrake/mesh.py
```diff
@@ -1294,6 +1359,35 @@ def init(self):
         if hasattr(self, '_callback'):
             self._callback(self)
 
+    def partial_init(self, topology):
```
danshapero, line 1362: I think `init_topology` is a more obvious name given what it does. There's also no underscore in front of the name -- is it supposed to be part of the public API?

firedrake/mesh.py
ksagiyam: That makes sense. 
Created a separate PR just for things around partitioner (https://github.com/firedrakeproject/firedrake/pull/1934).

firedrake/mesh.py
ksagiyam: Thanks, Dan. Maybe we should rename, but this method does slightly more than just setting topology, so I'm not sure what we should call it.
This method is actually called from `Mesh` function, so I think we should leave it public?

wence-: Is it something that a user would expect to call by hand? If not, can probably be private

ksagiyam: I see. I'll make it private. Thanks, both.

I'm still trying to pick a name: `_init_topology`, `_partial_init`, `_init_without_coordinates`, or something else?

firedrake/output.py
```diff
@@ -1,10 +1,109 @@
...
+                        |     Data: fA
+                        +-- fB_0
+                              Attr: firedrake_function_space: "V"
+                              Data: fB_0
+
+We can load the `Function`s that we saved in the above as in the
+following.
+
+>>> with CheckpointFile("example.h5", 'r') as afile:
+>>>     f = afile.save_function("fA", mesh_name="meshA")
```
lrtfm, line 79: Here should be `f = afile.load_function("fA", mesh_name="meshA")`

firedrake/output.py
```diff
@@ -651,3 +754,304 @@ def write(self, *functions, **kwargs):
...
+        self._function_spaces = {}
+        self._function_space_datas = {}
+
+    def __enter__(self):
+        return self
+
+    def __exit__(self, *args):
+        self.close()
+
+    def save_mesh(self, mesh, format=ViewerHDF5.Format.HDF5_PETSC):
```
lrtfm, line 852: Will this support saving different topology information (i.e. different dmplexs) in the same hdf5 file?

ksagiyam: Yes, this will. `PETSc` currently writes all topology information in `/`, but, as soon as it is fixed, this branch will reflect it.

firedrake/output.py
ksagiyam: Thank you. Fixed now.

firedrake/output.py
```diff
@@ -651,3 +756,366 @@ def write(self, *functions, **kwargs):
...
+        section.setPermutation(tmesh._plex_renumbering)
+        sectiondm.setSection(section)
+        sfXB = self._meshes[mesh_key]["sfXB"]
+        sfXC = self._meshes[mesh_key].setdefault("sfXC",
+                                                 sfXB.compose(tmesh.sfBC) if tmesh.sfBC else sfXB)
+        gsf, lsf = dm.sectionLoad(self.viewer, sectiondm, sfXC)
+        sd_key = get_shared_data_key_for_checkpointing(tmesh, element)
+        _ = get_global_numbering(tmesh, sd_key, global_numbering=sectiondm.getSection())
+        # Construct function space
+        V = FunctionSpace(mesh, element, name=name)
```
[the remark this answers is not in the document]
wence-: Can we get around this by just constructing the functionspace on the loaded (redistributed?) mesh and then computing the relevant sf that pushes forward? AIUI, `dm.sectionLoad` populates the section inside the sectiondm by pushing forward dofs in the on disk section through sfXC. But we already know how many dofs will be on each point, and we just need to line them up (or is this harder?)

firedrake/output.py
```diff
@@ -651,3 +756,366 @@ def write(self, *functions, **kwargs):
...
+        self._function_load_utils[sd_key + mesh_key] = (sectiondm, gsf, lsf)
+        self._function_spaces[fs_key] = V
+        return V
+
+    def load_function(self, name, mesh_name=DEFAULT_MESH_NAME, distribution_parameters={}):
```
wence-, line 1023: Why do I need distribution parameters when loading a function?

wence-: I mean, I know why, but why is this the interface?

Could we instead have a loading scheme that is:

```
with ... as f:
   mesh = f.load_mesh(...)
   g = f.load_function(name, mesh)
```

firedrake/output.py
```diff
@@ -651,3 +756,366 @@ def write(self, *functions, **kwargs):
...
+        # We should be able to save/load labels selectively.
+        plex.removeLabel("pyop2_core")
+        plex.removeLabel("pyop2_owned")
+        plex.removeLabel("pyop2_ghost")
+        mesh = Mesh(plex, name=name, distribution_parameters=distribution_parameters)
+        self._meshes[mesh_key] = {"mesh": mesh,
+                                  "sfXB": sfXB}
+        return mesh
+
+    def load_function_space(self, name, mesh_name, distribution_parameters={}):
```
wence-, line 964: Do I, as a user, ever need to load a function space? I think no.

wence-: Again, like with `load_function`, is it preferable to have:

```
with ... as f:
    mesh = f.load_mesh(...)
    V = f.load_function_space(name, mesh)
```

firedrake/output.py
```diff
@@ -651,3 +756,366 @@ def write(self, *functions, **kwargs):
...
+        mesh_key = (name, dist_key)
+        if mesh_key in self._meshes:
+            return self._meshes[mesh_key]["mesh"]
+        plex, sfXB = _dmload_plex(self.viewer, name)
+        # These labels are distribution dependent.
+        # We should be able to save/load labels selectively.
+        plex.removeLabel("pyop2_core")
+        plex.removeLabel("pyop2_owned")
+        plex.removeLabel("pyop2_ghost")
+        mesh = Mesh(plex, name=name, distribution_parameters=distribution_parameters)
```
wence-, line 959: Is the default to do redistribution (for better load balance)? If not, it should be.

firedrake/output.py
```diff
@@ -651,3 +756,366 @@ def write(self, *functions, **kwargs):
...
+            else:
+                getattr(_f, method)(f)
+            self.save_function(_f)
+            return
+        vec = tf.dat._vec
+        vec.setName(tf.name())
+        dm.globalVectorView(self.viewer, sectiondm, vec)
+
+    def load_mesh(self, name=DEFAULT_MESH_NAME, distribution_parameters={}):
+        dist_key = hash(frozenset(distribution_parameters.items()))
```
wence-, line 949: frozensets are usable as dictionary keys so no need to hash here.

firedrake/output.py
```diff
@@ -651,3 +756,366 @@ def write(self, *functions, **kwargs):
...
+        plex.setOptionsPrefix("loaded_")
+        plex.viewFromOptions("-dm_view")
+    elif format == ViewerHDF5.Format.HDF5_XDMF:
+        plex.load(viewer=viewer)  # TODO: load labels, too.
+        plex.setOptionsPrefix("loaded_")
+        plex.viewFromOptions("-dm_view")
+        plex.interpolate()
+        plex.setOptionsPrefix("interpolated_")
+        plex.viewFromOptions("-dm_view")
+        sfXB = None
```
wence-, line 797: Is this a loading format we want to support? After all, we can't really do anything with it.

firedrake/output.py
```diff
@@ -651,3 +756,366 @@ def write(self, *functions, **kwargs):
...
+        sectiondm = PETSc.DMShell().create(dm.comm)
+        sectiondm.setPointSF(dm.getPointSF())
+        sectiondm.setSection(global_numbering)
+    else:
+        sectiondm = tV._shared_data.node_set.halo.dm
+    sectiondm.setName(get_sectiondm_name(*sd_key))
+    return sectiondm
+
+
+class Dict(dict): pass
```
wence-, line 830: So that you can have these in a weakvaluedictionary?

firedrake/output.py
```diff
@@ -651,3 +756,366 @@ def write(self, *functions, **kwargs):
...
+        mesh.init()
+        # Load element
+        sectiondm_name = self.find_sectiondm_name(mesh_name, name)
+        path = os.path.join("/topologies",
+                            mesh.topology_dm.getName(),
+                            "dms",
+                            sectiondm_name,
+                            PREFIX + "_function_spaces",
+                            name)
+        element_b = self.get_attr(path, PREFIX + "_ufl_element").tobytes()
```
wence-, line 1000: There's kind of a lot of repeated code for name construction in the hierarchy. I wonder if there is a way to make this less fragile (not specifically this line, I just needed to put this comment somewhere).

firedrake/output.py
```diff
@@ -651,3 +756,366 @@ def write(self, *functions, **kwargs):
...
+
+class CheckpointFile(object):
+    _mesh_cache = weakref.WeakValueDictionary()
+
+    def __init__(self, filename, mode, comm=COMM_WORLD):
+        self.viewer = ViewerHDF5()
+        self.viewer.create(filename, mode=mode, comm=comm)
+        commkey = comm.py2f()
+        assert commkey != MPI.COMM_NULL.py2f()
+        self._meshes = self._mesh_cache.setdefault((filename, commkey), Dict())
```
wence-, line 841: I don't think this does what you think it does:
```
from weakref import WeakValueDictionary as wvd
import gc

foo = wvd()

class Dict(dict):
    pass

baz = object()
bar = Dict()
bar[1] = baz
foo[2] = bar
del bar
gc.collect()
print(list(foo.keys()))
=> []
```

Whereas I think you're trying to hold on to the meshes you've loaded as long as they are in use?

wence- (commented):
A first go

firedrake/functionspacedata.py
```diff
@@ -369,6 +373,64 @@ def entity_dofs_key(entity_dofs):
...
+                subsub_key.append((ssk, tuple(entity_permutations[k][sk][ssk])))
+            sub_key.append(tuple(subsub_key))
+        key.append(tuple(sub_key))
+    key = tuple(key)
+    return key
+
+
+import ufl
+from tsfc.finatinterface import create_element
+def get_shared_data_key(mesh, element, use_true_tensor_element_shape=False):
```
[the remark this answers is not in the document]
ksagiyam: We almost can use the same code in the constructor of `FunctionSpaceData` to get `(nodes_per_entity, real_tensorproduct)` cache key, but this function wants to take `ufl_element` as argument, while the constructor of `FunctionSpaceData` takes `finat_element`. I think it simplifies the code around `FunctionSpace` construction if we make the constructor of `FunctionSpaceData` (and thus `get_shared_data`) take `ufl_element` instead of `finat_element`. If that makes sense, I can make a small PR which just does it.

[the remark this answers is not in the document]
wence-: OK, I think that seems reasonable.

ksagiyam (commented):
Will fix more later.

firedrake/mesh.py
ksagiyam: Thanks. Fixed

firedrake/mesh.py
```diff
@@ -1719,16 +1798,22 @@ def Mesh(meshfile, **kwargs):
...
 
-    if isinstance(meshfile, function.Function):
+    name = kwargs.get("name", DEFAULT_MESH_NAME)
+    if isinstance(meshfile, str) and \
+       any(meshfile.lower().endswith(ext) for ext in ['.h5', '.hdf5']):
+        from firedrake.output import CheckpointFile
+
+        with CheckpointFile(meshfile, 'r', comm=kwargs.get("comm", COMM_WORLD)) as afile:
+            mesh = afile.load_mesh(name=name)
+        return mesh
```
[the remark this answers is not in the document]
ksagiyam: Thanks, fixed.

firedrake/output.py
ksagiyam: Ok. Thanks.

firedrake/output.py
```diff
@@ -651,3 +756,366 @@ def write(self, *functions, **kwargs):
...
+                    return mixed_fs_name
+        base_path = os.path.join("/topologies", mesh_name, "dms")
+        for sectiondm_name in self.h5pyfile[base_path]:
+            path = os.path.join(base_path, sectiondm_name, "vecs")
+            self.h5pyfile.require_group(path)
+            if function_name in self.h5pyfile[path]:
+                return self.get_attr(os.path.join(path, function_name), PREFIX + "_function_space")
+        raise ValueError(f"Function {function_name} not found.")
+
+    @property
```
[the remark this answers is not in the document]
wence-: This is somewhat strange, because `cached_property` literally automates the exact code you have written here. Can you isolate the hang/reproduce outside a test suite?

firedrake/output.py
```diff
@@ -651,3 +756,366 @@ def write(self, *functions, **kwargs):
...
+    def __enter__(self):
+        return self
+
+    def __exit__(self, *args):
+        self.close()
+
+    def save_mesh(self, mesh, format=ViewerHDF5.Format.HDF5_PETSC):
+        dm = mesh.topology.topology_dm
+        if dm.getName() in self.h5pyfile.require_group("/topologies"):
+            return
```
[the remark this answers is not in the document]
wence-: I think at least checking the sizes of the topology arrays (cones, etc...) is a good plan. If that is possible to do.

firedrake/output.py
```diff
@@ -1,22 +1,127 @@
...
+>>> V = FunctionSpace(mesh, "CG", 2, name="V")
+>>> W = FunctionSpace(mesh, "CG", 1, name="W")
+>>> Z = MixedFunctionSpace((V, W), name="Z")
+>>> f = Function(V, name="fA")
+>>> g = Function(Z, name="fB")
+>>> with CheckpointFile("example.h5", 'w') as afile:
+>>>     afile.save_function(f)
+>>>     afile.save_function(g)
+
+The output file, "example.h5", would contain among others:
```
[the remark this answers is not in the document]
wence-: Maybe mention that this comes from `h5dump`.

firedrake/output.py
ksagiyam: That is right. Removed all related codes and tests.

firedrake/mesh.py
```diff
@@ -1719,16 +1798,22 @@ def Mesh(meshfile, **kwargs):
...
-    import firedrake.functionspace as functionspace
     import firedrake.function as function
 
-    if isinstance(meshfile, function.Function):
+    name = kwargs.get("name", DEFAULT_MESH_NAME)
+    if isinstance(meshfile, str) and \
+       any(meshfile.lower().endswith(ext) for ext in ['.h5', '.hdf5']):
+        from firedrake.output import CheckpointFile
+
+        with CheckpointFile(meshfile, 'r', comm=kwargs.get("comm", COMM_WORLD)) as afile:
```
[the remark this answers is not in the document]
ksagiyam: Done.

firedrake/output.py
```diff
@@ -1,22 +1,127 @@
+# -*- coding: utf-8 -*-
+"""Module for scalable HDF5 I/O.
+
+This module contains `CheckpointFile` class that supports scalable input
+output of Firedrake `Function`s and surrounding objects. Specifically, one
+can save a `Function` on N processes and later load it on P processes
+efficiently.
```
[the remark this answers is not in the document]
ksagiyam: Fixed.

firedrake/output.py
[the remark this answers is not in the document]
ksagiyam: Yes, done.

firedrake/output.py
```diff
@@ -651,3 +756,366 @@ def write(self, *functions, **kwargs):
...
+    dm = tV.mesh().topology_dm
+    if isinstance(tV.ufl_element(), (ufl.VectorElement, ufl.TensorElement)):
+        nodes_per_entity, real_tensorproduct = sd_key
+        assert not real_tensorproduct
+        global_numbering = get_global_numbering(tV.mesh(), sd_key)
+        sectiondm = PETSc.DMShell().create(dm.comm)
+        sectiondm.setPointSF(dm.getPointSF())
+        sectiondm.setSection(global_numbering)
+    else:
+        sectiondm = tV._shared_data.node_set.halo.dm
```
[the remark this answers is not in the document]
ksagiyam: Fixed.

firedrake/output.py
```diff
@@ -651,3 +756,366 @@ def write(self, *functions, **kwargs):
...
+        if format == ViewerHDF5.Format.HDF5_PETSC:
+            dm.topologyView(viewer=self.viewer)
+            dm.coordinatesView(viewer=self.viewer)
+            dm.labelsView(viewer=self.viewer)
+        else:
+            dm.view(viewer=self.viewer)
+        self.viewer.popFormat()
+
+    def save_function_space(self, V):
+        if type(V.topological) is impl.MixedFunctionSpace:
```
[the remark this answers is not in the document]
ksagiyam: Yes. Done.

firedrake/output.py
```diff
@@ -651,3 +756,366 @@ def write(self, *functions, **kwargs):
...
+                                     V.name)
+            self.h5pyfile.require_group(base_path)
+            self.set_attr(base_path, PREFIX + "_num_sub_spaces", V.num_sub_spaces())
+            for i, Vsub in enumerate(V):
+                path = os.path.join(base_path, str(i))
+                self.h5pyfile.require_group(path)
+                self.set_attr(path, PREFIX + "_mesh", Vsub.mesh().name)
+                self.set_attr(path, PREFIX + "_function_space", Vsub.name)
+                self.save_function_space(Vsub)
+            return
```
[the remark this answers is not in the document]
ksagiyam: Fixed.

firedrake/output.py
```diff
@@ -651,3 +756,366 @@ def write(self, *functions, **kwargs):
...
+                                   PREFIX + "_function_spaces")
+        if V.name in self.h5pyfile.require_group(fs_dir_name):
+            return
+        dm.sectionView(self.viewer, sectiondm)
+        path = os.path.join(fs_dir_name, V.name)
+        self.h5pyfile.require_group(path)
+        element_b = pickle.dumps(tV.ufl_element())
+        self.set_attr(path, PREFIX + "_ufl_element", numpy.void(element_b))
+        self.set_attr(path, PREFIX + "_ufl_element_family", tV.ufl_element().family())
+        self.set_attr(path, PREFIX + "_ufl_element_degree", tV.ufl_element().degree())
```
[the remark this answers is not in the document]
ksagiyam: That is right. Added a comment.

firedrake/output.py
```diff
@@ -651,3 +756,366 @@ def write(self, *functions, **kwargs):
...
+        # Save function space
+        tV = V.topological
+        dm = tV.mesh().topology_dm
+        sectiondm = get_sectiondm_for_checkpointing(tV)
+        fs_dir_name = os.path.join("/topologies",
+                                   dm.getName(),
+                                   "dms",
+                                   sectiondm.getName(),
+                                   PREFIX + "_function_spaces")
+        if V.name in self.h5pyfile.require_group(fs_dir_name):
```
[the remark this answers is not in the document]
ksagiyam: Added these lines:
```
# Test if the pickled UFL element matches the original element
element = pickle.loads(element_b)
if element != tV.ufl_element():
    raise RuntimeError("pickled UFL element does not match the original element")
```
I think this is what you meant?

firedrake/output.py
```diff
@@ -651,3 +756,366 @@ def write(self, *functions, **kwargs):
...
+                                     V.name,
+                                     PREFIX + "_functions",
+                                     f.name())
+            self.h5pyfile.require_group(base_path)
+            for i, fsub in enumerate(f.split()):
+                path = os.path.join(base_path, str(i))
+                self.h5pyfile.require_group(path)
+                self.set_attr(path, PREFIX + "_function", fsub.name())
+                self.save_function(fsub)
+            return
```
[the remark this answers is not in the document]
ksagiyam: Fixed. Thanks.

firedrake/output.py
```diff
@@ -651,3 +756,366 @@ def write(self, *functions, **kwargs):
...
+            _f = Function(_V, name=PREFIX_EMBEDDED + f.name())
+            if method == "project":
+                getattr(_f, method)(f, solver_parameters={"ksp_rtol": 1.e-16})
+            else:
+                getattr(_f, method)(f)
+            self.save_function(_f)
+            return
+        vec = tf.dat._vec
+        vec.setName(tf.name())
+        dm.globalVectorView(self.viewer, sectiondm, vec)
```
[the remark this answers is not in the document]
ksagiyam: Fixed!

firedrake/output.py
```diff
@@ -651,3 +756,366 @@ def write(self, *functions, **kwargs):
...
+            _f = self.load_function(PREFIX_EMBEDDED + f.name(), mesh_name, distribution_parameters=distribution_parameters)
+            element = V.ufl_element()
+            _element, method = get_embedding_element(element)
+            assert _element == _f.function_space().ufl_element()
+            if method == "project":
+                getattr(f, method)(_f, solver_parameters={"ksp_rtol": 1.e-16})
+            else:
+                getattr(f, method)(_f)
+            return f
+        vec = PETSc.Vec().createWithArray(tf.dat._data, size=None, bsize=tf.dat.cdim, comm=dm.comm)
```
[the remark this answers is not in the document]
ksagiyam: Fixed.

firedrake/output.py
```diff
@@ -651,3 +756,366 @@ def write(self, *functions, **kwargs):
...
+
+    def close(self):
+        #import gc
+        #del self._h5pyfile
+        #gc.collect()
+        if hasattr(self, "_h5pyfile"):
+            self._h5pyfile.flush()
+            del self._h5pyfile
+        self.viewer.destroy()
+        #del self.viewer
```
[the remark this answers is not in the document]
ksagiyam: Done.

tests/output/test_io_function.py
```diff
@@ -0,0 +1,477 @@
...
+    with PETSc.Log.Event("CheckpointFileSaveFunction" + log_event_id):
+        start = time.perf_counter()
+        with CheckpointFile(filename, 'w', comm=comm) as afile:
+            afile.save_function(fB)
+        end = time.perf_counter()
+        avgtime = comm.allreduce(end - start, MPI.SUM) #/ comm.size
+        print(200, comm.rank, flush=True)
+        if comm.rank == 0:
+            print("nproc = ", comm.size, ", func save time = ", avgtime, flush=True)
+"""
```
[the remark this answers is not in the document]
ksagiyam: Removed.

tests/output/test_io_function.py
```diff
@@ -0,0 +1,477 @@
...
+
+def _load_check_save_functions(filename, func_name, comm, method, mesh_name, log_event_id):
+    # Load
+    #with CheckpointFile(filename, "r", comm=comm) as afile:
+    #    meshB = afile.load_mesh(mesh_name)
+    #meshB.init()
```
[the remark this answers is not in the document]
ksagiyam: Removed.

tests/output/test_io_function.py
```diff
@@ -0,0 +1,477 @@
...
+    if _element == V.ufl_element():
+        pass
+    else:
+        _V = FunctionSpace(mesh, _element)
+        _f = getattr(Function(_V), method)(f)
+        f = getattr(Function(V), method)(_f)
+    fe = Function(V)
+    getattr(fe, method)(_get_expr(V))
+    assert assemble(inner(f - fe, f - fe) * dx) < 1.e-15
+"""
```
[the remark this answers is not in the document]
ksagiyam: Removed.

tests/output/test_io_orientation.py
```diff
@@ -0,0 +1,94 @@
...
+    #    print("")
+    #print("fx:", fx.dat.data)
+    #print("fy:", fy.dat.data)
+
+if __name__ == "__main__":
+    # mpiexec -n 1 python tests/output/test_io_orientation.py -dm_view ascii::ascii_info_detail
+    #test_io_orientation_simplex("ext_quad", ("DPC", 0))
+    #test_io_orientation_simplex("ext_quad", ("CG", 3))
+    #test_io_orientation_simplex("triangle", ("CG", 2))
+    test_io_orientation_simplex("cube", ("CG", 2))
```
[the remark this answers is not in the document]
ksagiyam: Ok! Removed.

firedrake/output.py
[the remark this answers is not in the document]
ksagiyam: Yes. The following (almost always) hangs with `cached_property` with `mpiexec -n 2 ...`, but never with the current code as far as I've observed.
```
from firedrake import *
from pyop2.mpi import COMM_WORLD


filename = "test_hang.h5"
mesh_name = "default"
func_name = "f"


def _load_check_save_functions():
    with CheckpointFile(filename, "r") as afile:
        fB = afile.load_function(func_name, mesh_name=mesh_name)
    print(assemble(inner(fB, fB) * dx), flush=True)
    with CheckpointFile(filename, "w") as afile:
        afile.save_function(fB)
    print(f"rank = {COMM_WORLD.rank}: end of function:", flush=True)


def test_io_function_base():
    meshA = UnitSquareMesh(1, 1, quadrilateral=True)
    VA = FunctionSpace(meshA, "RTCF", 5, name="V")
    fA = Function(VA, name=func_name)
    with CheckpointFile(filename, 'w') as afile:
        afile.save_function(fA)
    for i in range(2):
        _load_check_save_functions()


test_io_function_base()
```
This looks very touchy to me. Even if it fails, if we run it again without doing `firedrake-clean`, it passes.

firedrake/output_utils.py
```diff
@@ -0,0 +1,73 @@
...
+        sub_elements.append(e)
+        methods.append(m)
+    method = "project" if any(m == "project" for m in methods) else "interpolate"
+    return ufl.TensorProductElement(*sub_elements), method
+
+
+
+
+
+
```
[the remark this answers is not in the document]
ksagiyam: Removed.

tests/output/test_io_function.py
```diff
@@ -0,0 +1,477 @@
...
+        with CheckpointFile(filename, 'w', comm=comm) as afile:
+            afile.save_function(fB)
+        end = time.perf_counter()
+        avgtime = comm.allreduce(end - start, MPI.SUM) #/ comm.size
+        print(200, comm.rank, flush=True)
+        if comm.rank == 0:
+            print("nproc = ", comm.size, ", func save time = ", avgtime, flush=True)
+"""
+
+def _load_check_save_functions(filename, func_name, comm, method, mesh_name, log_event_id):
```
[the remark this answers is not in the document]
ksagiyam: Removed.

tests/output/test_io_function.py
```diff
@@ -0,0 +1,477 @@
...
+    mesh_name = "exampleDMPlex"
+    meshA = _get_mesh(cell_type, mesh_name, COMM_WORLD)
+    mesh_name = meshA.name
+    VA_list = []
+    for i, (family, degree) in enumerate(family_degree_tuples):
+        VA_list.append(FunctionSpace(meshA, family, degree, name=fs_name + f"[{str(i)}]"))
+    VA = functools.reduce(lambda a, b: a * b, VA_list)
+    VA.name = fs_name
+    fA = Function(VA, name=func_name)
+    _embed(fA, _get_expr(VA), method)
```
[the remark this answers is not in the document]
ksagiyam: Renamed it `_initialise_function`.

tests/output/test_io_function.py
```diff
@@ -0,0 +1,477 @@
...
+    getattr(f, method)(_get_expr(V))
+    if _element == V.ufl_element():
+        pass
+    else:
+        _V = FunctionSpace(mesh, _element)
+        _f = getattr(Function(_V), method)(f)
+        f = getattr(Function(V), method)(_f)
+    fe = Function(V)
+    getattr(fe, method)(_get_expr(V))
+    assert assemble(inner(f - fe, f - fe) * dx) < 1.e-15
```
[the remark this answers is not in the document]
ksagiyam: Sorry, this test was under construction, but  we now have a valid tests for tensor element.

firedrake/output.py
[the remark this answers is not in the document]
ksagiyam: Added some check. Added `dmcommon.compute_point_cone_global_sizes` to compute global point/cone sizes (I don't think these global sizes are saved in plex?)

firedrake/functionspacedata.py
```diff
@@ -369,6 +373,64 @@ def entity_dofs_key(entity_dofs):
...
+            for ssk in sorted(entity_permutations[k][sk]):
+                subsub_key.append((ssk, tuple(entity_permutations[k][sk][ssk])))
+            sub_key.append(tuple(subsub_key))
+        key.append(tuple(sub_key))
+    key = tuple(key)
+    return key
+
+
+import ufl
+from tsfc.finatinterface import create_element
```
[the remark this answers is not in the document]
ksagiyam: Fixed.

firedrake/functionspacedata.py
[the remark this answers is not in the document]
ksagiyam: #2154 has been merged and I've cleaned up these lines.

firedrake/output.py
ksagiyam: I think it looks better now.

firedrake/output.py
```diff
@@ -651,3 +756,366 @@ def write(self, *functions, **kwargs):
...
+        plex.viewFromOptions("-dm_view")
+        plex.interpolate()
+        plex.setOptionsPrefix("interpolated_")
+        plex.viewFromOptions("-dm_view")
+        sfXB = None
+    viewer.popFormat()
+    return plex, sfXB
+
+
+from firedrake.functionspacedata import get_global_numbering, get_shared_data_key
```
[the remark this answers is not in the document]
ksagiyam: As https://github.com/firedrakeproject/firedrake/pull/2154 has been merged, we now import `preprocess_ufl_element` (which basically does what `get_shared_data_key` did) from `firedrake.functionspacedata`.

firedrake/output.py
```diff
@@ -651,3 +756,366 @@ def write(self, *functions, **kwargs):
...
+        # Save mesh
+        self.save_mesh(V.mesh())
+        # Save function space
+        tV = V.topological
+        dm = tV.mesh().topology_dm
+        sectiondm = get_sectiondm_for_checkpointing(tV)
+        fs_dir_name = os.path.join("/topologies",
+                                   dm.getName(),
+                                   "dms",
+                                   sectiondm.getName(),
```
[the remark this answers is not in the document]
ksagiyam: `sectiondm` is internal, so we name it based on the shared data key, `(nodes_per_entity, real_tensorproduct)`, so that they never collide. It looks something like `sectiondm_1_3_3_False`.

firedrake/output.py
```diff
@@ -651,3 +754,381 @@ def write(self, *functions, **kwargs):
...
+            dm = get_sectiondm_for_checkpointing(tV)
+            topology_name = topology_dm.getName()
+            dm_name = dm.getName()
+            if tV.name not in self.h5pyfile.require_group(self.path_to_function_spaces(topology_name, dm_name)):
+                # Save section
+                topology_dm.sectionView(self.viewer, dm)
+                # Save UFL element
+                element = tV.ufl_element()
+                element_b = pickle.dumps(element)
+                path = self.path_to_function_space(topology_name, dm_name, tV.name)
```
wence-, line 899: It seems that two functionspaces with the same data layout but different UFL elements (and the same name, which is often `None` because that's the default) will map to the same place here. e.g. a `FiniteElement("DG", triangle, 0)` and a `FiniteElement("B", triangle, 3)` have the same entity dofs, but vastly different meanings.

ksagiyam: Yes. This is related to other comments of yours, but, currently, we assume that users name `FunctionSpace`s when they want to save `Function`s defined on them. Maybe we should automatically construct `FunctionSpace` names from UFL element somehow?

wence-: People aren't thinking about saving function spaces, they're thinking about saving functions, so I don't think we should require functionspaces to have names. Can we just give them generated (unique per section) names in the file and save an attribute on the section that maps elements to location (e.g. via a pickled dict).

ksagiyam: Yes. I think that makes sense. I'll work on that. Thanks Lawrence!

ksagiyam: Updated. I think it looks much cleaner now.

tests/output/test_io_mesh.py
```diff
@@ -0,0 +1,45 @@
...
+                                       ViewerHDF5.Format.HDF5_PETSC])
+@pytest.mark.parametrize('heterogeneous', [False, True])
+def test_io_mesh(outformat, heterogeneous, tmpdir):
+    # Parameters
+    fname = os.path.join(str(tmpdir), "test_io_mesh_dump.h5")
+    fname = COMM_WORLD.bcast(fname, root=0)
+    ntimes = 3
+    # Create mesh.
+    mesh = RectangleMesh(4, 1, 4., 1.)
+    mesh.init()
```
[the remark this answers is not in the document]
ksagiyam: It turned out that it would just work as adding partition overlaps does not change what we save; tested just in case. So I've removed this `mesh.init()`.

firedrake/output.py
```diff
@@ -651,3 +756,366 @@ def write(self, *functions, **kwargs):
...
+def get_sectiondm_for_checkpointing(tV):
+    sd_key = get_shared_data_key_for_checkpointing(tV.mesh(), tV.ufl_element())
+    dm = tV.mesh().topology_dm
+    if isinstance(tV.ufl_element(), (ufl.VectorElement, ufl.TensorElement)):
+        nodes_per_entity, real_tensorproduct = sd_key
+        assert not real_tensorproduct
+        global_numbering = get_global_numbering(tV.mesh(), sd_key)
+        sectiondm = PETSc.DMShell().create(dm.comm)
+        sectiondm.setPointSF(dm.getPointSF())
+        sectiondm.setSection(global_numbering)
```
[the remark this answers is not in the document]
ksagiyam: Added `block_size` optional argument to `dmcommon.create_section` and `mesh.create_section`.

firedrake/output.py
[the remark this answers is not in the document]
ksagiyam: With #2154, I've now made it clear that we use the same cache key generator here and in `functionspacedata`.

`sfXC` is an SF that pushes the global number slab (`X = [0, N)`) to the good plex (redistributed and overlapped as we have already called `mesh.init()`). `DMPlexSectionLoad()` does the following:
- Load raw section (`sectionA`) with `PetscSectionLoad()`
- Construct `sfAX` that pushes points in `sectionA` to the slab `X`
- Compose `sfAC = SFCompose(sfAX, sfXC)`
- `PetscSFDistributeSection(sfAC, sectionA, &remoteOffsets, sectionC)` (`sectionC` is the section on the good plex)
- Use `remoteOffsets` to construct `globalDataSF` (or `localDataSF`) that directly pushes on-disk vector data to global vectors (or local vectors) defined on `sectionC`

As you noted, the resulting `sectionC` would be identical to what we would construct using the UFL element (because we happen to know the definite rule to go from the UFL element to the section), but it is also necessary to  construct `globalDataSF` and `localDataSF` from `remoteOffsets`, which knows the on-disk array offsets, which looks hard to do outside PETSc.

I wanted to call `get_global_numbering` here to avoid reconstruction of `global_numbering` as `dm.sectionLoad` constructs it anyway. I tried to treat the case where `global_numbering` has already been in the cache. Does it look right?

firedrake/output.py
```diff
@@ -1,22 +1,127 @@
+# -*- coding: utf-8 -*-
```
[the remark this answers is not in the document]
ksagiyam: Added deprecation warning for those two classes for now. I should probably move `CheckpointFile` class to `checkpointing.py`.

firedrake/output.py
ksagiyam: Yes. Thanks. I tried to fix it. I hope it is fine now.

firedrake/output_utils.py
```diff
@@ -0,0 +1,73 @@
+# -*- coding: utf-8 -*-
+"""Module for utility functions for scalable HDF5 I/O.
```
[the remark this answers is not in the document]
ksagiyam: The thing in `mg/embedded.py` was actually what we needed here, so I refactored and put it in `embedding.py` (`get_embedding_dg_element()`). For `get_sup_element`, we could loop over `elements` and use `get_embedding_dg_element()` for each element, but I don't know if that's a good idea.

firedrake/output.py
ksagiyam: Yes. Updated. Thanks. I think this is intuitive.

firedrake/output.py
ksagiyam: Made `load_function_space` private (`_load_function_space`). Now the interface is as the following:
Saving:
```
with CheckpointFile("....h5", "w", ...) as file:
    file.save_mesh(mesh)  # <- optional
    file.save_function(f)
```
Loading:
```
with CheckpointFile("....h5", "r", ...) as file:
    mesh = file.load_mesh(mesh_name)
    f = file.load_function(mesh, func_name)
```

firedrake/output.py
```diff
@@ -651,3 +756,366 @@ def write(self, *functions, **kwargs):
...
+        sectiondm = tV._shared_data.node_set.halo.dm
+    sectiondm.setName(get_sectiondm_name(*sd_key))
+    return sectiondm
+
+
+class Dict(dict): pass
+
+
+class CheckpointFile(object):
+    _mesh_cache = weakref.WeakValueDictionary()
```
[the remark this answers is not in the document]
ksagiyam: Added.

ksagiyam:
I realised that changing path to topology after Firedrake users start using I/O feature is going to be tough, so I made a MR for path change: https://gitlab.com/petsc/petsc/-/merge_requests/4209.

I've added support for extrusion; we basically have full support, but I've omitted some parallel tests due to https://github.com/firedrakeproject/firedrake/issues/2169.

docs/source/checkpointing.rst
```diff
@@ -8,29 +8,214 @@
...
+::
+
+    HDF5 "example.h5" {
+    FILE_CONTENTS {
+     group      /
+     group      /topologies
+     group      /topologies/firedrake_mixed_meshes
+     group      /topologies/firedrake_mixed_meshes/meshA
+     group      /topologies/firedrake_mixed_meshes/meshA/firedrake_mixed_function_spaces
+     group      /topologies/firedrake_mixed_meshes/meshA/firedrake_mixed_function_spaces/firedrake_function_space_meshA_<CG2 on a triangle>_meshA_<CG1 on a triangle>
```
dham, line 67: Let's not have spaces and angle brackets.

dham (changes requested):

docs/source/checkpointing.rst
```diff
@@ -8,29 +8,214 @@
...
+with `idx` parameter always unset, and the same :class:`~.Function` can only be
+loaded using the same mode.
+
+
+Checkpointing with DumbCheckpoint
+=================================
+
+.. warning::
+
+   :class:`~.DumbCheckpoint` will be deprecated after 01/01/2022.
```
wence-, line 207: This is in the past.

ksagiyam: That's true... I'll fix. Thank you!

docs/source/checkpointing.rst
ksagiyam: Yes! I will fix.

PR #1920, opened 2020-11-18
ASM: Set the LGMap for tinyasm to use

pefarrell:
TinyASM needs the local-to-global map on the matrix, but it isn't always set. (For example, it's not set for a fieldsplit within a fieldsplit, at least.) This small patch sets the local-to-global map on the matrix before it is passed in to TinyASM.


wence- (approved):

wence-:
@florianwechsung does this look sane?

PR #1916, opened 2020-11-12
Only set CC=mpicc, CXX=mpicxx, F90=mpif90 when necessary

ksagiyam:
This PR is to set CC=mpicc, CXX=mpicxx, and F90=mpif90 only when necessary in the install script, as suggested in https://github.com/firedrakeproject/firedrake/issues/1914.

PR #1926, opened 2020-11-24
Only set CC=mpicc, CXX=mpicxx, F90=mpif90 when necessary

ksagiyam:
Copy of https://github.com/firedrakeproject/firedrake/pull/1916, hoping that Jenkins would work.


wence- (approved):
Thanks, I think this looks OK.

Issue #1923, opened 2020-11-21
Crossed box mesh Jacobian of Stokes system is singular


ioannisPApapadopoulos:
Trying to solve the Stokes system on a crossed box mesh fails due to a singular Jacobian. To ensure it's not to do with the nullspace of the pressure, I fix it with a Real Lagrange multiplier. Toggling the flag `crossed_flag` True produces a crossed box mesh which fails, whereas False runs with a normal box mesh which converges without an issue. Using `-info info.log` in the failing case complains that `[0] MatFactorNumeric_MUMPS(): matrix is numerically singular, INFOG(1)=-10, INFO(2)=32883` 

Does anyone have an idea what might be going wrong? Many thanks!

```
# Failing crossed box mesh example
from firedrake import *

delta = 1.5
def InflowOutflow(mesh):
    x = SpatialCoordinate(mesh)
    r = sqrt(1./(12.*pi))
    gbar = 1.0
    pipe = lt(((x[1] - 1.0/2)/r)**2 + ((x[2] - 1.0/2)/r)**2, 1.0)
    x_on_boundary = Or(lt(x[0], 1e-10), gt(x[0], delta-1e-10))

    val_in_pipe   = gbar*(1 - ( ((x[1] - 1.0/2)/r)**2 + ((x[2] - 1.0/2)/r)**2 ))

    x_component = conditional(x_on_boundary,
                              conditional(pipe, val_in_pipe, Constant(0)),
                              Constant(0))

    y_component = Constant(0)
    z_component = Constant(0)
    return as_vector([x_component, y_component, z_component])

crossed_flag = True # Toggle me to False to get working non-crossed box mesh
if crossed_flag:
    mesh = BoxMesh(5,5,5,1.5,1,1, diagonal="crossed")
else:
    mesh = BoxMesh(5,5,5,1.5,1,1)

Ve = VectorElement("CG", mesh.ufl_cell(), 2, dim=3)
Pe = FiniteElement('CG', mesh.ufl_cell(), 1)
Re = FiniteElement("R",  mesh.ufl_cell(), 0) 
Ze = MixedElement([Ve,Pe,Re])
Z = FunctionSpace(mesh, Ze)

z = Function(Z)
(u,p,p0) = split(z)
w = TestFunction(Z)
(v,q,q0) = split(w)

F = (  inner(grad(u),grad(v))*dx - inner(div(v), p)*dx # momentum
     - inner(div(u), q)*dx                             # incompressbility
[...]

PR #1927, opened 2020-11-25
Add the NotImplementedError for crossed box mesh for Pk (k>1) elements

jingminxia:
The crossed cutting of hexahedra is implemented in PR #1776 

However, as discussed in the reported issue #1923 about using crossed box mesh, there seems to have broken connectivity issues of this mesh.

Hence, this PR is adding a `NotImplementedError` when referring to `diagonal="crossed"` in `BoxMesh`.

### DIFF
--- a/firedrake/utility_meshes.py
+++ b/firedrake/utility_meshes.py
@@ -784,6 +784,7 @@ def BoxMesh(nx, ny, nz, Lx, Ly, Lz, reorder=None, distribution_parameters=None, 
                  v2, v4, v6, v7,
                  v1, v2, v7, v4]
         cells = np.asarray(cells).swapaxes(0, 3).reshape(-1, 4)
+        raise NotImplementedError("The crossed cutting of hexahedra has a broken connectivity issue for Pk (k>1) elements")
     else:
         raise ValueError("Unrecognised value for diagonal '%r'", diagonal)
 
--- a/tests/regression/test_mesh_generation.py
+++ b/tests/regression/test_mesh_generation.py
@@ -435,10 +435,3 @@ def test_changing_default_reorder_works(reorder):
     finally:
         parameters["reorder_meshes"] = old_reorder
 
-
-@pytest.mark.parametrize("kind, num_cells",
-                         [("default", 6), ("crossed", 5)])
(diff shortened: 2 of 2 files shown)


firedrake/utility_meshes.py
```diff
@@ -784,6 +784,7 @@ def BoxMesh(nx, ny, nz, Lx, Ly, Lz, reorder=None, distribution_parameters=None,
                  v2, v4, v6, v7,
                  v1, v2, v7, v4]
         cells = np.asarray(cells).swapaxes(0, 3).reshape(-1, 4)
+        raise NotImplementedError("The crossed cutting of hexahedra has a broken connectivity issue for Pk (k>1) elements")
```
wence-, line 787: Thanks, I think there is a test in tests/regression/test_mesh_generation.py as well that needs to be removed.

jingminxia: Yes, the test is now removed. Thanks!

tests/regression/test_mesh_generation.py
```diff
@@ -435,10 +435,3 @@ def test_changing_default_reorder_works(reorder):
     finally:
         parameters["reorder_meshes"] = old_reorder
 
-
-@pytest.mark.parametrize("kind, num_cells",
-                         [("default", 6), ("crossed", 5)])
```
dham, line 440: Please only remove the crossed case. Default should still be fine.

dham (changes requested):

tests/regression/test_mesh_generation.py
jingminxia: Yes, this is now edited in the newest commit.

tests/regression/test_mesh_generation.py
```diff
@@ -435,9 +435,8 @@ def test_changing_default_reorder_works(reorder):
     finally:
         parameters["reorder_meshes"] = old_reorder
 
-
```
wence-, line 438: This line removal broke linting.

jingminxia: sorry... this is now fixed.

wence- (approved):

wence-:
Thanks.

PR #1928, opened 2020-12-01
Slate BLAS/LAPACK fix

JDBetteridge:
Currently Slate is linking against LAPACK by just passing `-llapack` to the compiler, which picks up the system LAPACK.

This is fine, unless we have specified another BLAS at install time or installed OpenBLAS as part of PETSc.

These changes fix the situation for OpenBLAS, but may not fix for the general case.

### DIFF
--- a/firedrake/petsc.py
+++ b/firedrake/petsc.py
@@ -1,4 +1,5 @@
 # Utility module that imports and initialises petsc4py
+import os
 import petsc4py
 import sys
 petsc4py.init(sys.argv)
@@ -6,7 +7,6 @@ from petsc4py import PETSc
 import itertools
 import functools
 from contextlib import contextmanager
-from pathlib import Path
 
 
 __all__ = ("PETSc", "OptionsManager", "get_petsc_variables")
@@ -87,8 +87,8 @@ def get_petsc_variables():
     The result is memoized to avoid constantly reading the file.
     """
     config = petsc4py.get_config()
-    variables_path = Path(config["PETSC_DIR"])/config["PETSC_ARCH"]
-    variables_path /= "lib/petsc/conf/petscvariables"
+    path = [config["PETSC_DIR"], config["PETSC_ARCH"], "lib/petsc/conf/petscvariables"]
+    variables_path = os.path.join(*path)
     with open(variables_path) as fh:
         # Split lines on '=', assignment
         splitlines = (line.split("=") for line in fh.readlines())
--- a/firedrake/petsc.py
+++ b/firedrake/petsc.py
@@ -79,7 +79,7 @@ def flatten_parameters(parameters, sep="_"):
     return new
 
 
-@functools.cache
+@functools.lru_cache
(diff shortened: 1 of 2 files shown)


firedrake/slate/slac/compiler.py
```diff
@@ -164,7 +165,20 @@ def generate_loopy_kernel(slate_expr, tsfc_parameters=None):
...
+        ldflags = []
+        include = []
+    elif blasdir:
+        # FIX ME: OpenBLAS doesn't need `-llapack`, but others might?
+        ldflags = [f"-Wl,-rpath,{blasdir}/lib", f"-L{blasdir}/lib"]
+        include = [f"{blasdir}/include"]
+    else:
+        # This case assumes we use the system (netlib) LAPACK
+        ldflags = ["-llapack"]
+        include = []
```
wence-, line 180: I think we can get this in a consistent manner from `petsc/default/lib/petsc/conf/petscvariables` (look for `BLASLAPACK_LIB` and `BLASLAPACK_INCLUDE`).

firedrake/petsc.py
```diff
@@ -77,6 +79,24 @@ def munge(keys):
...
+@functools.lru_cache()
+def get_petsc_variables():
+    """Get dict of PETSc environment variables from the file:
+    $PETSC_DIR/$PETSC_ARCH/lib/petsc/conf/petscvariables
+
+    The result is memoized to avoid constantly reading the file.
+    """
+    config = petsc4py.get_config()
+    variables_path = Path(config["PETSC_DIR"])/config["PETSC_ARCH"]
+    variables_path /= "lib/petsc/conf/petscvariables"
```
wence-, line 91: `os.path.join(..., ...)`

firedrake/petsc.py
```diff
@@ -77,6 +79,24 @@ def munge(keys):
...
+    $PETSC_DIR/$PETSC_ARCH/lib/petsc/conf/petscvariables
+
+    The result is memoized to avoid constantly reading the file.
+    """
+    config = petsc4py.get_config()
+    path = [config["PETSC_DIR"], config["PETSC_ARCH"], "lib/petsc/conf/petscvariables"]
+    variables_path = os.path.join(*path)
+    with open(variables_path) as fh:
+        # Split lines on '=', assignment
+        splitlines = (line.split("=") for line in fh.readlines())
```
wence-, line 94:
```
var, val = line.split("=", maxsplit=1)
```

wence- (commented):
OK, modulo style in `get_petsc_variables`

firedrake/petsc.py
JDBetteridge: Now fixed, but used:

`splitlines = (line.split("=", maxsplit=1) for line in fh.readlines())`

for a neater dictionary comprehension in the return line. There is no need to split into `var` and `val` as these values are never used separately.

wence- (approved):

PR #1929, opened 2020-12-02
fix docstring for SubDomainData

ksagiyam:
Fix docstring for SubDomainData.


firedrake/mesh.py
```diff
@@ -1996,6 +1996,8 @@ def SubDomainData(geometric_expr):
     The result can be attached as the subdomain_data field of a
     :class:`ufl.Measure`. For example:
 
+    .. code-block:: python
```
dham, line 1999: Should be python3

ksagiyam: Changed `.. code-block:: python` -> `.. code-block:: python3` everywhere.

PR #1930, opened 2020-12-02
Change to MixedVectorSpaceBasis with components in multiple subspaces.

stephankramer:
Previously MixedVectorSpaceBasis([basisA, basisB]) would take the
Cartesian product of the two sets of basis vectors and form a basis for
a nullspace of the full mixed space that way. However the Cartesian
product formed that way does not necessarily form a basis, e.g.:

basisA = [a1, a2, a3] and basisB = [b1, b2]

then

(a1, b1) - (a2, b1) + (a2, b2) - (a1, b2) = (0, 0)

so basisA x basisB is not independent.

Instead what is needed - at least what I think a reasonable
interpretation of what MixedVectorSpaceBasis([basisA, basisB]) means - is
the direct sum of span(basisA) and span(basisB), understood in the sense
of: span(basisA) x {0}_B + {0}_A x span(basisB) for which the basis is
simply: basisA x {0}_B + {0}_A x basisB - i.e. the n/o vectors in the
combined basis is just the sum of the number in the individual ones.

Note that this implicitly assumes that a1 is a null vector of the
first diagonal block *and* (a1, 0) is a null vector of the full mixed
system. Although this is true for many applications, it is not for many
others.

### DIFF
--- a/firedrake/nullspace.py
+++ b/firedrake/nullspace.py
@@ -232,36 +232,32 @@ class MixedVectorSpaceBasis(object):
         The monolithic basis is formed by the cartesian product of the
         bases forming each sub part.
         """
-        from itertools import product
-        bvecs = [[None] for _ in self]
-        # Get the complete list of basis vectors for each component in
-        # the mixed basis.
+        self._vecs = []
         for idx, basis in enumerate(self):
             if isinstance(basis, VectorSpaceBasis):
                 vecs = basis._vecs
                 if basis._constant:
                     vecs = vecs + (function.Function(self._function_space[idx]).assign(1), )
-                bvecs[idx] = vecs
+                for vec in vecs:
+                    mvec = function.Function(self._function_space)
+                    mvec.sub(idx).assign(vec)
+                    self._vecs.append(mvec)
 
-        # Basis for mixed space is cartesian product of all the basis
-        # vectors we just made.
-        allbvecs = [x for x in product(*bvecs)]
-
-        vecs = [function.Function(self._function_space) for _ in allbvecs]
-
-        # Build the functions representing the monolithic basis.
-        for vidx, bvec in enumerate(allbvecs):
-            for idx, b in enumerate(bvec):
-                if b:
-                    vecs[vidx].sub(idx).assign(b)
-        for v in vecs:
-            v /= v.dat.norm
-
-        self._vecs = vecs
         self._petsc_vecs = []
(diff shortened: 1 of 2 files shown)


ScottMacLachlan:
Would it be reasonable to add functionality to allow the user to separately specify the monolithic null space from the block null spaces?  I don't have a use case for that at hand, but it seems like a natural option - if no monolithic null space is given, then do the construction as this does in `_build_monolithic_basis`, but if one is given, then just process it as needed and then pass it off to `PETSc.NullSpace().create`?

stephankramer:
That's definitely a reasonable suggestion, I think, and I think it should be easy to come up with use cases. There's a question about the interface: do we just have a (optional) keyword argument `monolithic_nullspace=VectorSpaceBasis(mixed_vecs)`, or do we need to redesign this more generally? For instance it might be a little odd that if you have no nullspaces for the diagonal blocks at all,  and only a nullspace at the level of the full system, you have to provide the `VectorSpaceBasis` directly to `nullspace=` in the solver interface (as you can do already), whereas if you do have both you do `nullspace=MixedVectorSpaceBasis([sub_basisA, sub_basisB], monolithic_nullspace=VectorSpaceBasis(mixed_vecs))`. On the other hand this would be simplest, backwards-compatible way of implementing it.

ScottMacLachlan:
I'm not sure that I'm qualified to comment on what coding style/backward compatibility is expected here - I was hoping that @wence- might weigh in on that.  A (naive?) suggestion might be to preserve the original interface, but allow the `sub_basis` arguments to be empty, so that we could have all of
`nullspace=MixedVectorSpaceBasis([sub_basisA, sub_basisB], monolithic_nullspace=[])`, `nullspace=MixedVectorSpaceBasis([sub_basisA, sub_basisB])`, `nullspace=MixedVectorSpaceBasis([], monolithic_nullspace=VectorSpaceBasis(mixed_vecs))`, and `nullspace=MixedVectorSpaceBasis([sub_basisA, sub_basisB], monolithic_nullspace=VectorSpaceBasis(mixed_vecs))`
as valid calls?  An important question would be how an empty/absent `monolithic_nullspace` argument would be interpreted - as either there being no null space to the monolithic system, or the null space being constructed from the direct sum of the components...

wence- (approved):

wence-:
I think this code is right, let's continue the broader discussion in #1941.

PR #1934, opened 2020-12-09
allow users to set partitioner

ksagiyam:
This PR is to:
- define `set_partitioner` method in `MeshTopology` class
- allow users to set `partitioner` as:
`mesh = Mesh(..., distribution_parameters={'partitioner_type': 'parmetis'})`

### DIFF
--- a/firedrake/mesh.py
+++ b/firedrake/mesh.py
@@ -660,7 +660,7 @@ class MeshTopology(AbstractMeshTopology):
         self._distribution_parameters = distribution_parameters.copy()
         if distribute is None:
             distribute = True
-
+        partitioner_type = distribution_parameters.get("partitioner_type")
         overlap_type, overlap = distribution_parameters.get("overlap_type",
                                                             (DistributedMeshOverlapType.FACET, 1))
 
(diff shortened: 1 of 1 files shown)


firedrake/mesh.py
```diff
@@ -922,6 +910,52 @@ def cell_set(self):
         size = list(self._entity_classes[self.cell_dimension(), :])
         return op2.Set(size, "Cells", comm=self.comm)
 
+    def set_partitioner(self, distribute, partitioner_type=None):
+        """Set partitioner for (re)distributing underlying plex over comm.
+
+        :arg distribute: Boolean or (sizes, points)-tuple.  If (sizes, point)-
+            tuple is given, it is used to set shell partition. If Boolean, no-op.
+        :kwarg partitioner_type: Partitioner to be used: "chaco", "ptscotch", "parmetis",
+            "shell", or `None` (unspecified).
```
wence-, line 919: "shell" does not seem to be a valid type from code below.

wence-: No, I see.

wence-, line 919: Say in docstring that `partitioner_type` is ignored if the distribute parameter specifies the distribution.

wence- (commented):
Minor comments on docstring, otherwise 👍

firedrake/mesh.py
ksagiyam: Done. Thanks!

Issue #1935, opened 2020-12-10
Permissions on a firedrake install


stephankramer:
This is an issue that causes frequent, but intermittent failures in the Jenkins testing of Thetis: the testsuite is (sometimes?) run as user `jenkins` inside a firedrake vanilla docker container where firedrake has been built as user `firedrake`. As a workaround permissions can be changed afterwards, but changing permissions on a large number of files in Docker is slow (on Jenkins it takes about 12 mins to change permisisions on all executables in lib/ and bin/). It was suggested instead to enable other users to run the container, by building it with a more permissive umask.

Looking a little further into this. I believe there are two separate issues:

## Executable permissions
If you look into the executable permissions in lib/ and bin/ in the firedrake container (or in any other firedrake install that has been run with the standard umask 022) with `ls -l $(find lib/ bin/ -executable -type f)` then you can see that the only executables that do not have executable permission for all, are the mpicc, mpicxx, mpiexec and mpif90 scripts that firedrake-install installs itself:
```
for opt in ["mpicc", "mpicxx", "mpif90", "mpiexec"]:
    ...
    if should_link:
        with open(dest, "w") as f:
            f.write('#!/bin/bash' + os.linesep + src + ' "$@"')
        os.chmod(dest, os.stat(dest).st_mode | stat.S_IEXEC)
```
Note that having `umask 0000` would not help in this case: it would still create readable and writable for all, but then only set executable for the user. So I believe the fix here is to just be more permissive in that last line. Not sure what the correct way to install is, somewhere on stackexchange the following was suggested:
```
      perm = os.stat(dest).st_mode
      os.chmod(dest,  perm | ((perm & 0o444) >> 2))
```
which looks at the current read permissions (which will be set according to umask), and then gives executable permissions to those that have read permission by shifting that bit 2 to the right.
This would also make it possible for users to share a firedrake installation in other situations (outside docker).

## Write permissions
I believe this only affects the .cache folder. One option is to just have the docker user set a different cache directory by setting `FIREDRAKE_TSFC_KERNEL_CACHE_DIR` and `PYOP2_CACHE_DIR`
Setting umask with docker seems to be a bit cumbersome. I could do something like:
```
RUN bash -c "umask 0000 & python3 firedrake-install --no-package-manager --disable-ssh --remove-build-files"
```
[...]

PR #1937, opened 2020-12-15
Fix permissions of mpi wrappers

stephankramer:
Previously these only have executable permission for user (running the
firedrake-install script). This causes issues if the install is used by
other users (as happens in containerized Jenkins testing of firedrake
based projects).

See issue #1935. This PR does not deal with write permissions that may also be necessary in the docker container.

### DIFF
--- a/scripts/firedrake-install
+++ b/scripts/firedrake-install
@@ -4,7 +4,6 @@ import platform
 import subprocess
 import sys
 import os
-import stat
 import shutil
 from argparse import ArgumentParser, RawDescriptionHelpFormatter
 import argparse
@@ -1485,9 +1484,10 @@ if mode == "install":
             os.remove(dest)
         if should_link:
             log.debug("Creating a wrapper for %s from %s to %s" % (opt, src, dest))
-            with open(dest, "w") as f:
+            # use low-level os.open to ensure permissions (incl. exec.) are set according to umask
+            fd = os.open(dest, os.O_CREAT | os.O_WRONLY)
+            with os.fdopen(fd, "w") as f:
                 f.write('#!/bin/bash' + os.linesep + src + ' "$@"')
-            os.chmod(dest, os.stat(dest).st_mode | stat.S_IEXEC)
 
 cc = options["mpicc"]
 cxx = options["mpicxx"]


scripts/firedrake-install
```diff
@@ -1462,9 +1462,10 @@ for opt in ["mpicc", "mpicxx", "mpif90", "mpiexec"]:
         os.remove(dest)
     if should_link:
         log.debug("Creating a wrapper for %s from %s to %s" % (opt, src, dest))
-        with open(dest, "w") as f:
+        # use low-level os.open to ensure permissions (incl. exec.) are set according to umask
```
ReubenHill, line 1465: I might be being thick but I don't understand this comment - what's umask?

dham: The system that unix-like operating systems use to determine the default permissions of files on creation.

dham (commented):
I'm wondering if this is completely safe. What if a user's umask doesn't set the x bit? Should we also explicitly make the scripts executable by everyone? (belt and braces).

scripts/firedrake-install
stephankramer: umask is what determines the default permissions for newly created files and directories in a unix-like environment, to be more precise umask tells what bits should *not* be set. On a lot of machines this is set to 022 which means disallow writing for "group" and "other". I will admit I never really knew how this worked until a week ago.

To give more detail on the comment: if you do the standard `open('filename', 'w')` in python it does a lot of things for you: it first checks whether the file actually exists and if so, if you have permission to write to it, if not it will create a new file with some permissions (which importantly here does not include executable permission) both cases create a OS level file handle that can be writtent to. Then, as a second step it will also create a python file object that is setup to write to. If you want more control, you can split these two steps but you have to be more explicit. Here we already know we want to create a new file (because we have just deleted any existing version) and we want it to be writable. The default permissions are 0o777, i.e. all permissions, read write and execute for user, group and other, except those given by umask. That is precisely what you need for executable script: with a standard 022 umask you get a 755 == rwxr-xr-x file.

Alternatively you can open a file in the more standard way and fix permissions afterwards, with a chmod as was done previously, but it turns out to do that correctly is a bit awkward. To ask what the default permissions should be, python's os.umask requires you to actually change the value and it will return the current value. So you then have to set it back. And then you have to do the magic of something like `0o755 | ~umask` to filter out the disallowed permission from what you actually want.

stephankramer:
Well technically with umask you set the permissions you don't want to give (i.e. given less would be fine). So I would say that if a user has an unusual umask that tells us that they don't want executable for group or other, then we should respect that. It's probably a bit academic anyway because directories created in the install would not have execute either...

scripts/firedrake-install
ReubenHill: I understand, thank you!

stephankramer:
Pushed again after conflict with #1897 merge

ReubenHill (approved):
LGTM

PR #1939, opened 2020-12-17
Restore _topology_dm

ReubenHill:
A second go at #1931 since `._topology_dm` is used so widely.

`.topology_dm` has been made a property of `AbstractMeshTopology` and `MeshGeometry` which is set set with `._topology_dm`.
This restores `._topology_dm` as an accessible member though does not explicitly document it.

I was not able to work out a way to have both `.topology_dm` and `._topology_dm` be documented properties due to `@property` on `topology_dm` seemingly only allowing `self._topology_dm` to be the return value.

### DIFF
--- a/firedrake/mesh.py
+++ b/firedrake/mesh.py
@@ -352,6 +352,9 @@ class AbstractMeshTopology(object, metaclass=abc.ABCMeta):
 
         self.name = name
 
+        self.topology_dm = None
+        r"The PETSc DM representation of the mesh topology."
+
         # A cache of shared function space data on this mesh
         self._shared_data_cache = defaultdict(dict)
 
@@ -397,6 +400,13 @@ class AbstractMeshTopology(object, metaclass=abc.ABCMeta):
 
         This is to ensure consistent naming for some multigrid codes."""
         return self
+
+    @property
+    def _topology_dm(self):
+        """Alias of topology_dm"""
+        from warnings import warn
+        warn("_topology_dm is deprecated (use topology_dm instead)", DeprecationWarning, stacklevel=2)
+        return self.topology_dm
 
     def ufl_cell(self):
         """The UFL :class:`~ufl.classes.Cell` associated with the mesh.
@@ -679,6 +689,7 @@ class MeshTopology(AbstractMeshTopology):
         plex.setFromOptions()
 
         self.topology_dm = plex
+        r"The PETSc DM representation of the mesh topology."
         self._comm = dup_comm(plex.comm.tompi4py())
 
         # Mark exterior and interior facets
@@ -952,6 +963,7 @@ class ExtrudedMeshTopology(MeshTopology):
         # access them directly.  Eventually we would want a better refactoring
         # of responsibilities between mesh and function space.
         self.topology_dm = mesh.topology_dm
+        r"The PETSc DM representation of the mesh topology."
         self._plex_renumbering = mesh._plex_renumbering
         self._cell_numbering = mesh._cell_numbering
         self._entity_classes = mesh._entity_classes
@@ -1147,6 +1159,7 @@ class VertexOnlyMeshTopology(AbstractMeshTopology):
 
         self._parent_mesh = parentmesh
         self.topology_dm = swarm
+        r"The PETSc DM representation of the mesh topology."
         self._comm = dup_comm(swarm.comm.tompi4py())
 
         # A cache of shared function space data on this mesh
@@ -1305,6 +1318,13 @@ class MeshGeometry(ufl.Mesh, MeshGeometryMixin):
 
         This is to ensure consistent naming for some multigrid codes."""
         return self._topology
+
+    @property
+    def _topology_dm(self):
+        """Alias of topology_dm"""
+        from warnings import warn
+        warn("_topology_dm is deprecated (use topology_dm instead)", DeprecationWarning, stacklevel=2)
+        return self.topology_dm
 
     @utils.cached_property
     @MeshGeometryMixin._ad_annotate_coordinates_function


ReubenHill:
I half wonder if `_plex` should be added as a property too

ksagiyam:
I think simply adding the following to `AbstractMeshTopology` is better as `mesh.topology_dm` is now the formal way to access `topology_dm`:
```
@property
def _topology_dm(self):
    """Alias of topology_dm."""
    return self.topology_dm
```

Adding `_plex` property to `MeshTopology` sounds good to me, though no one has complained about the absence of `_plex` as far as I know.

ReubenHill:
> I think simply adding the following to `AbstractMeshTopology` is better as `mesh.topology_dm` is now the formal way to access `topology_dm`:
> 
> ```
> @property
> def _topology_dm(self):
>     """Alias of topology_dm."""
>     return self.topology_dm
> ```
> 
> Adding `_plex` property to `MeshTopology` sounds good to me, though no one has complained about the absence of `_plex` as far as I know.

It won't work - once `_topology_dm` has been marked as a property, you can't set it.

ksagiyam:
You would need `@_topology_dm.setter`, but do we allow users to set `topology_dm` via `_topology_dm`?

ReubenHill:
Well it hasn't been a property until now so anyone has been able to do what they like.

dham:
> Well it hasn't been a property until now so anyone has been able to do what they like.

As of right now it doesn't exist, so presumably it's fine because nobody is screaming!

dham:
> I think simply adding the following to `AbstractMeshTopology` is better as `mesh.topology_dm` is now the formal way to access `topology_dm`:
> 
> ```
> @property
> def _topology_dm(self):
>     """Alias of topology_dm."""
>     return self.topology_dm
> ```
> 
> Adding `_plex` property to `MeshTopology` sounds good to me, though no one has complained about the absence of `_plex` as far as I know.

Private names should only be for internal use, so if we're not using `_plex` then it doesn't need to be there.

wence-:
Why not make `topology_dm` the public thing and add the (deprecated) `_topology_dm` as a property?

```python
def __init__(self, ...):
    self.topology_dm = ...

@property
def _topology_dm(self):
    from warnings import warn
    warn("_topology_dm is deprecated (use topology_dm instead)", DeprecationWarning, stacklevel=2)
    return self.topology_dm
```

We can argue if we want the deprecation warning.

No need for setters anywhere.

ReubenHill:
> Why not make `topology_dm` the public thing and add the (deprecated) `_topology_dm` as a property?
> 
> ```python
> def __init__(self, ...):
>     self.topology_dm = ...
> 
> @property
> def _topology_dm(self):
>     from warnings import warn
>     warn("_topology_dm is deprecated (use topology_dm instead)", DeprecationWarning, stacklevel=2)
>     return self.topology_dm
> ```
> 
> We can argue if we want the deprecation warning.
> 
> No need for setters anywhere.

Then how do I document `.topology_dm`?

dham:
> > Why not make `topology_dm` the public thing and add the (deprecated) `_topology_dm` as a property?
> > ```python
> > def __init__(self, ...):
> >     self.topology_dm = ...
> > 
> > @property
> > def _topology_dm(self):
> >     from warnings import warn
> >     warn("_topology_dm is deprecated (use topology_dm instead)", DeprecationWarning, stacklevel=2)
> >     return self.topology_dm
> > ```
> > 
> > 
> > We can argue if we want the deprecation warning.
> > No need for setters anywhere.
> 
> Then how do I document `.topology_dm`?

With an attribute docstring.

wence-:
> With an attribute docstring.

```python

def __init__(self, ...):
    self.topology_dm = ...
    r"This is a docstring that sphinx attaches to the topology_dm attribute"
    self.bar = ...
    r"This docstring is attached to :attr:`bar`"

ReubenHill:
Okay so @wence- 's solution is only necessary if you want a deprecation warning (since it needs to happen inside a the property getter method as he's illustrated). Otherwise it's ugly since `topology_dm` is not a `@property` of `AbstractMeshTopology` and `MeshGeometry` and it would be more in keeping with the rest of the code if it were. Annoyingly, attaching a docstring to `._topology_dm` whilst `.topology_dm` _is_ a `@property` does not work (at least in ipython the docstring doesn't show up).

I think unless you want `._topology_dm` to be documented and/or have a deprecation warning this code is good as is and is an improvement (at the very least stylistically) on what came before. I await a not-yucky alternative to persuade me otherwise.

wence-:
> Annoyingly, attaching a docstring to ._topology_dm whilst .topology_dm is a @property does not work (at least in ipython the docstring doesn't show up).

The property part of this is a red herring, it's just that ipython doesn't parse attribute docstrings. They will show up in sphinx though.

ReubenHill:
Well nevertheless I stand by my comment that it's philosophically wrong to have `topology_dm` not be a `@property` if it's part of the public API

wence-:
> Well nevertheless I stand by my comment that it's philosophically wrong to have topology_dm not be a @property if it's part of the public API

Why?

dham:
> > Well nevertheless I stand by my comment that it's philosophically wrong to have topology_dm not be a @Property if it's part of the public API
> 
> Why?

Yes, why? This isn't a normal Python convention. Public attributes are absolutely fine.

ReubenHill:
> > > Well nevertheless I stand by my comment that it's philosophically wrong to have topology_dm not be a @Property if it's part of the public API
> > 
> > 
> > Why?
> 
> Yes, why? This isn't a normal Python convention. Public attributes are absolutely fine.

Because it's out of step with the rest of `mesh.py` - it feels like it's hidden, particularly if one can't see the docstring in ipython. I can't be the only person who mainly uses that to check for docstrings when exploring functionality.

ReubenHill:
Pushed what I understand to be the suggested alternative

wence-:
Please rebase onto trunk.

ReubenHill:
I rebased and pushed this but it's still failing to build documentation

wence-:
Looks like doc building is broken in trunk. Can you take a look?

ReubenHill:
Tests passing - this is ready to be merged

wence- (approved):
Thanks!

PR #1942, opened 2021-01-06
Add hessian to InterpolateBlock

ReubenHill:
Avoids NotImplementedError when calling firedrake_adjoint.minimize with the Newton-CG method which needs InterpolateBlock to have a correctly implemented hessian.

~WIP!~ Todo:

- [x] Assuming test is correct and failure is real, get the correct hessian
- [x]  Finish commenting

@stephankramer deserves loads of credit for helping with this

Happy for this to be reviewed.

### DIFF
--- a/firedrake/adjoint/blocks.py
+++ b/firedrake/adjoint/blocks.py
@@ -402,7 +402,7 @@ class InterpolateBlock(Block, Backend):
         if len(adj_inputs) > 1:
             raise(NotImplementedError("Interpolate block must have a single output"))
         dJdm = self.backend.derivative(prepared, inputs[idx])
-        return self.backend.Interpolator(dJdm, self.V).interpolate(adj_inputs[0], transpose=True)
+        return self.backend.Interpolator(dJdm, self.V).interpolate(adj_inputs[0], transpose=True).vector()
 
     def prepare_evaluate_tlm(self, inputs, tlm_inputs, relevant_outputs):
         return replace(self.expr, self._replace_map())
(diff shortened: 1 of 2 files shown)


nbouziani (dismissed):
This looks all good for me.

ReubenHill:
I've actually just found a bug that I'm fixing

ReubenHill:
The return type for `evaluate_adj_component` and `evaluate_hessian_component` needs to be a `Vector` not a `Function`

nbouziani:
Yes that's indeed a bug. I would have expected an error occuring in `evaluate_hessian` in `pyadjoint/blocks.py` when updating the adjoint value for the hessian if your `evaluate_hessian_component` does not return a vector. Is that the case ?

ReubenHill:
No, there is no such bug. It _can_ turn up in an upstream `evaluate_hessian` block where that block is nonlinear - you'll find yourself trying to add a `hessian_input` of type `Function` to a `Vector`. Note that this was already wrong with `evaluate_adj_component` before I made the changes. I've made sure the tests implicitly check for the right type with  `evaluate_hessian_component` but I'm not sure how to test for `evaluate_adj_component` - it's clearly been flying under the radar.

nbouziani:
Ok I see. If I understand correctly: your implicit check consists in using somewhere in the tests `hessian_value.inner` instead of `hessian_value.vector().inner` so that you expect tests to fail if `evaluate_hessian_component` returns a `Function` since `Function` does not have an `inner` method. Then, why don't you just check directly the type in the tests to make sure that `evaluate_adj_component` and `evaluate_hessian_component` return a Vector ?

ReubenHill:
Good point. Updated to do that

nbouziani (approved):

PR #1943, opened 2021-01-07
Fix tsfc_interface cache key mismatch

connorjward:
The cache key is being stored in a different format (`key`) to how it is getting looked up (`(key, commkey)`). This means that it misses every time and gets read from disk instead.

This is a hacky fix, but I feel like that the caching implementation here could be tidied up a lot. Should I try and do that or are you happy to just merge this?

### DIFF
--- a/firedrake/tsfc_interface.py
+++ b/firedrake/tsfc_interface.py
@@ -87,7 +87,7 @@ class TSFCKernel(Cached):
 
         if val is None:
             raise KeyError("Object with key %s not found" % key)
-        return cls._cache.setdefault(key, pickle.loads(val))
+        return cls._cache.setdefault((key, comm.py2f()), pickle.loads(val))
 
     @classmethod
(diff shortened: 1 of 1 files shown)


wence- (dismissed):
Oh wow, this was stupid, thanks!

wence-:
> This is a hacky fix, but I feel like that the caching implementation here could be tidied up a lot. Should I try and do that or are you happy to just merge this?

Let's get this fix in. I think a broader redesign of the caching could be done, but perhaps better as part of an overall redesign of the PyOP2 system: in particular redoing the caching properties of various objects to deal with https://github.com/firedrakeproject/firedrake/projects/6

connorjward:
Ok sounds good.

wence-:
This broke some of the slate kernel equivalence checks, can you take a look?

dham (approved):

PR #1945, opened 2021-01-08
Don't install Python on OSX

dham:
Installing Python can never fix any problems, but can cause them. Let's just not.


wence-:
Should this also be the case on linux, or does it not matter here?

dham:
> Should this also be the case on linux, or does it not matter here?

On linux we only install supplementary packages like python-dev, so that we have headers. That's actually useful due to the way Ubuntu splits up packages.

wence- (approved):

PR #1963, opened 2021-01-25
Implement PMGSNES

pbrubeck:
Here we introduce the class `PMGSNES`, which enables the use of MG FAS for nonlinear problems.
The key main features are:
- Patch to enable assembly with symmetric TensorFunctionSpaces.
- Patch to allow matrix free projection.
- Fix logic errors in `tensor_product_space_query()`
- Matrix-free injection of the state.
- Coarsening of boundary conditions.
- Built-in coarsening of VectorElement, TensorElement, and TensorProductElements with a single polynomial degree.
- Coarsening of user-provided quadrature degree that preserves the ratio of interpolation (e.g. GLL) to quadrature (e.g. GL) nodes.
- Fix a bug related to the BLAS library flag when compiling the grid transfer kernel.


wence-:
Hi @pbrubeck from a cursory glance this looks like a nice feature. I don't have a lot of time in the next 2 weeks for reviews, but then stuff should ease up, so if I haven't looked by then, please give me a prod.

firedrake/preconditioners/pmg.py
```diff
@@ -62,12 +62,42 @@ def coarsen_element(ele):
...
+
+        Must provide the same degree for tensor product elements.
+
+        :arg ele: a :class:`ufl.FiniteElement` to reconstruct.
+        :arg N: an integer degree.
+        """
+        if isinstance(ele, (VectorElement, TensorElement)):
+            cVe = [PMGBase.reconstruct_degree(K, N)
+                   for K in ele.sub_elements()]
+            return VectorElement(cVe[0], dim=len(cVe))
```
pefarrell, line 93: Why `reconstruct_degree` on all subelements, when you only use the first one?

firedrake/preconditioners/pmg.py
```diff
@@ -62,12 +62,42 @@ def coarsen_element(ele):
...
+        elif isinstance(ele, TensorProductElement):
+            (Ve, Qe) = ele.sub_elements()
```
pefarrell, line 95: Do all `TensorProductElement`s only wrap two subelements?

firedrake/preconditioners/pmg.py
```diff
@@ -103,44 +133,27 @@ def initialize(self, pc):
...
+        pdm.setSNESJacobian(_SNESContext.form_jacobian)
+        pdm.setSNESFunction(_SNESContext.form_function)
+        pdm.setKSPComputeOperators(_SNESContext.compute_operators)
+
         pdm.setOptionsPrefix(pc.getOptionsPrefix() + "pmg_")
         set_function_space(pdm, get_function_space(odm))
 
         parent = get_parent(odm)
         assert parent is not None
-        add_hook(parent, setup=partial(push_parent, pdm, parent), teardown=partial(pop_parent, pdm, parent),
```
pefarrell, line 111: Was it necessary to change these lines (for lint etc)?

firedrake/preconditioners/pmg.py
```diff
@@ -103,44 +133,27 @@ def initialize(self, pc):
...
+        add_hook(parent, setup=partial(push_parent, pdm, parent),
+                 teardown=partial(pop_parent, pdm, parent),
                  call_setup=True)
-        add_hook(parent, setup=partial(push_appctx, pdm, ctx), teardown=partial(pop_appctx, pdm, ctx),
```
pefarrell, line 113: Same

firedrake/preconditioners/pmg.py
```diff
@@ -154,17 +167,19 @@ def coarsen(self, fdm, comm):
...
+        Injection = prolongation_matrix_matfree(cV, fV, [], [])
+
+        def inject_state():
+            with cu.dat.vec_wo as xc, fu.dat.vec_ro as xf:
+                Injection.multTranspose(xf, xc)
 
         parent = get_parent(fdm)
         assert parent is not None
-        add_hook(parent, setup=partial(push_parent, cdm, parent), teardown=partial(pop_parent, cdm, parent),
+        add_hook(parent,
```
pefarrell, line 180: Why break up into multiple lines?

firedrake/preconditioners/pmg.py
```diff
@@ -179,20 +194,40 @@ def inject_state(interpolators):
...
+                    N, = set(N)
+                except TypeError:
+                    pass
+                return N
+
+        # Coarsen the quadrature degree
+        # such that the ratio of GL to GLL nodes is preserved
         fcp = fctx._problem.form_compiler_parameters
-        cproblem = firedrake.NonlinearVariationalProblem(cF, cu, cbcs, cJ,
+        if fcp is not None:
```
pefarrell, line 221: I think it's more usual to specify the quadrature degree by passing options to the `Measure` object. I don't think this will catch that case. Is that correct?

firedrake/preconditioners/pmg.py
```diff
@@ -303,18 +453,20 @@ def tensor_product_space_query(V):
     if isinstance(ele, firedrake.TensorProductElement):
         family = set(e.family() for e in ele.sub_elements())
         try:
-            variant, = set(e.variant() or "spectral" for e in ele.sub_elements())
```
pefarrell, line 306: Can you comment on what the changes around here are doing?

firedrake/preconditioners/pmg.py
```diff
@@ -529,8 +682,8 @@ def make_blas_kernels(self, Vf, Vc):
             return;
         }}
         """
-        self.prolong_kernel = op2.Kernel(prolong_code, "prolongation", ldargs=["-lpetsc"])
-        self.restrict_kernel = op2.Kernel(restrict_code, "restriction", ldargs=["-lpetsc"])
+        self.prolong_kernel = op2.Kernel(prolong_code, "prolongation", ldargs=["-lblas"])
```
pefarrell, line 685: On supercomputers we usually get our BLAS from libraries other than `libblas.so`. Is there a way to figure out cleanly what BLAS we should be linking against here? I think the rationale for using PETSc was that it would be linked to the correct BLAS, but correct me if I'm wrong.

pefarrell (changes requested):
Can you add some tests that show off the new functionality?

firedrake/preconditioners/pmg.py
pbrubeck: Not at all, it was just my obsesion to split lines at 77 characters.

firedrake/preconditioners/pmg.py
pbrubeck: In that case, whenever we specify the degree on the measure, could I just add a measure with the coarse degree in the replace dictionary?

firedrake/preconditioners/pmg.py
```diff
@@ -62,12 +62,38 @@ def coarsen_element(ele):
...
+        :arg ele: a :class:`ufl.FiniteElement` to reconstruct.
+        :arg N: an integer degree.
+        """
+        if isinstance(ele, (VectorElement, TensorElement)):
+            sub = ele.sub_elements()
+            return VectorElement(PMGBase.reconstruct_degree(sub[0], N), dim=len(sub))
+        elif isinstance(ele, TensorProductElement):
+            return TensorProductElement(*(PMGBase.reconstruct_degree(sub, N) for sub in ele.sub_elements()))
+        else:
+            return ele.reconstruct(degree=N)
```
rckirby, line 97: This work is interesting, and I've needed similar things for P1PC-type things in tensor product cells.  It seems like the reconstruction should/could be done at the UFL level instead?

firedrake/preconditioners/pmg.py
pbrubeck: Good point, it should be extended to multiple subelements.

firedrake/preconditioners/pmg.py
pbrubeck: For one moment I thought there was a bug here, after checking twice, I think that it should be kept as it was before.

firedrake/preconditioners/pmg.py
pbrubeck: @djanekovic might have some useful input on this one...

djanekovic: This was bug I introduced and Pablo fixed. It is maybe confusing why we are including `petscblaslapack.h` but we need to link to `libblas.so`. That header file implements just the name mangling logic in order to call correct BLAS/LAPACK function. Because of that, we need to link to the `libblas.so` in the end.
Patrick, yes, you are correct and this problem was solved just recently. Jack added `get_petsc_variables` function in his PR #1928 which enables getting a path to the BLAS/LAPACK provided to the PETSc configure.
To sum up, Pablo, to link with correct BLAS library you should implement the same logic Jack implemented here: https://github.com/firedrakeproject/firedrake/blob/master/firedrake/slate/slac/compiler.py#L69

firedrake/preconditioners/pmg.py
pefarrell: Does `ufl.replace' replace `Measure`s? I have never checked, but I wouldn't think so.

firedrake/preconditioners/pmg.py
pefarrell: Can  you just add a comment explaining what's going on? (Especially if you got confused yourself!)

firedrake/preconditioners/pmg.py
pefarrell: So `P1PC` is a subclass of `PMGPC` but with `coarsen_element` overridden. One solution is to get @pbrubeck to modify that the overridden `coarsen_element` appropriately in this PR or a later PR also. But I think what @rckirby is asking is if this knowledge should go back into UFL. @wence- , any thoughts?

rckirby:
Specifically, coarsening of elements put into UFL and then used here rather
than done here.

firedrake/preconditioners/pmg.py
```diff
@@ -529,8 +670,21 @@ def make_blas_kernels(self, Vf, Vc):
             return;
         }}
         """
-        self.prolong_kernel = op2.Kernel(prolong_code, "prolongation", ldargs=["-lpetsc"])
-        self.restrict_kernel = op2.Kernel(restrict_code, "restriction", ldargs=["-lpetsc"])
+        BLASLAPACK_LIB = None
+        BLASLAPACK_INCLUDE = None
+        if not complex_mode:
```
djanekovic, line 675: I don't think you need this here since everything is complex safe.

firedrake/preconditioners/pmg.py
pbrubeck: I can easily extend this coarsening to `P1PC`. But a question arises: do we want the lowest order for a familty (e.g. degree = 0 in the case of DG) or stick to the naming and have always degree = 1?

rckirby:
Unclear.  DG(0) is lowest order, but some DG methods require degree 1 to
actually work.

Thought: default to 1, but is there a way to have a PETSc option specifying
the degree you want to reduce to so it could be overridden?  Similar use
case -- maybe you've got a good preconditioner for your degree 2 case and
want to use it instead of  degree 1.

firedrake/preconditioners/pmg.py
wence-: It doesn't, no, measures are outside of that stuff.

firedrake/preconditioners/pmg.py
wence-: You can import those variables from the slate file and then you need to do
```python
op2.Kernel(...,
                             include_dirs=BLASLAPACK_INCLUDE.split(),
                             ldargs=BLASLAPACK_LIB.split())
```

I think

firedrake/preconditioners/pmg.py
pefarrell: I like Rob's idea of defaulting to 1 and having it controllable via PETSc options.

tests/multigrid/test_p_multigrid.py
```diff
@@ -213,3 +213,65 @@ def test_p_multigrid_mixed():
...
+def test_p_fas_scalar():
+    mesh = UnitSquareMesh(4, 4, quadrilateral=True)
+    V = FunctionSpace(mesh, "CG", 7)
+    x = SpatialCoordinate(mesh)
+    u = Function(V)
+    v = TestFunction(V)
+    f = x[0]*(1-x[0]) + x[1]*(1-x[1])
+    bcs = DirichletBC(V, 0, "on_boundary")
+
+    F = inner(grad(u), grad(v))*dx - inner(f, v)*dx
```
pefarrell, line 227: Is it possible to include a _nonlinear_ test?

firedrake/preconditioners/pmg.py
pbrubeck: > Unclear. DG(0) is lowest order, but some DG methods require degree 1 to actually work. Thought: default to 1, but is there a way to have a PETSc option specifying the degree you want to reduce to so it could be overridden? Similar use case -- maybe you've got a good preconditioner for your degree 2 case and want to use it instead of degree 1.

I like your idea @rckirby. One way to do this is changing `coarsen_element` to a class method and having the coarse degree as an attribute set during the initialization. I don't think there would be any negative implications if I do this. Thoughts @pefarrell ?

firedrake/preconditioners/low_order.py
```diff
@@ -5,13 +5,17 @@
...
-    def coarsen_element(ele):
+    def coarsen_element(self, ele):
         if isinstance(ele, MixedElement) and not isinstance(ele, (VectorElement, TensorElement)):
             raise NotImplementedError("Implement this method yourself")
 
-        p = ele.degree()
-        if p == 1:
+        N = ele.degree()
+        try:
+            N = min(N)
```
pefarrell, line 14: Should we check here that they're all the same degree?
Like with `N, = set(N)` that you do below.

firedrake/preconditioners/pmg.py
```diff
@@ -179,20 +184,47 @@ def inject_state(interpolators):
...
+                except TypeError:
+                    pass
+                return N
+
+        Nf = get_max_degree(fV.ufl_element())
+        Nc = get_max_degree(cV.ufl_element())
+
+        # Coarsen the quadrature degree in a dictionary
+        # such that the ratio of GL to GLL nodes is preserved
+        def coarsen_quadrature(df, Nf, Nc):
```
pefarrell, line 212: Happy to see the test for this below, thanks.

tests/multigrid/test_p_multigrid.py
```diff
@@ -213,3 +209,170 @@ def test_p_multigrid_mixed():
...
+    ppc = solver.snes.getPythonContext().ppc
+    levels = ppc.getFASLevels()
+    assert levels == 3
+    assert ppc.getFASSmoother(levels-1).getLinearSolveIterations() == 0
+
+
+def test_p_fas_nonlinear_scalar():
+    mat_type = "matfree"
+    N = 4
+    dxq = dx(degree=3*N)  # here we also test coarsening of quadrature degree
```
pefarrell, line 293: Is there any way to assert in the test that the quadrature degree was indeed modified on the coarse problem?

pefarrell (changes requested):

tests/multigrid/test_p_multigrid.py
pbrubeck: There must be a way. I could try to extract the forms from the coarse SNES.

firedrake/preconditioners/low_order.py
pbrubeck: Sure, this can be done here as well, since there is not built-in support for different degrees.

pefarrell (dismissed):

wence- (commented):
Some comments

firedrake/mg/utils.py
```diff
@@ -109,8 +109,6 @@ def coarse_cell_to_fine_node_map(Vc, Vf):
         assert Vc.extruded == Vf.extruded
         if Vc.mesh().variable_layers or Vf.mesh().variable_layers:
             raise NotImplementedError("Not implemented for variable layers, sorry")
-        if Vc.extruded and Vc.mesh().layers != Vf.mesh().layers:
-            raise ValueError("Coarse and fine meshes must have same number of layers")
```
wence-, line 113: This has not been a restriction for a while?

firedrake/functionspaceimpl.py
```diff
@@ -287,7 +287,7 @@ def __init__(self, mesh, element, name=None, real_tensorproduct=False):
         if type(element) is ufl.TensorElement:
             # UFL enforces value_shape of the subelement to be empty
             # on a TensorElement.
-            self.shape = element.value_shape()
+            self.shape = element.reference_value_shape()
```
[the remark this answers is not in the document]
pbrubeck: So I just quickly tested this, and saw that the symmetric tensor is storing twice the value of the off-diagonal components. Is this supposed to happen? Also, in which file should I include the test?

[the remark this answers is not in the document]
pbrubeck: The doubling in the off-diagonal happens when assembling. I guess it makes sense if you do inner(TestFunction(V), symmetric_thing)*dx. But If when one does the projection this factor of two goes away.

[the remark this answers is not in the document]
pbrubeck: The output seems alright in paraview, e.g. I define a 3-by-3 tensor on a 2D mesh and the fields look like they are supposed to. Moreover, I apply a calculator filter with sum_{ij} (a_{ij} - a{ji})^2 and get a zero.

firedrake/mg/embedded.py
```diff
@@ -28,6 +28,9 @@ def __init__(self, element):
             degree = element.degree()
             family = lambda c: "DG" if c.is_simplex() else "DQ"
             if isinstance(cell, ufl.TensorProductCell):
+                if type(degree) is int:
+                    degree = [degree] * len(cell.sub_cells())
+
                 scalar_element = ufl.TensorProductElement(*(ufl.FiniteElement(family(c), cell=c, degree=d)
                                                             for (c, d) in zip(cell.sub_cells(), degree)))
```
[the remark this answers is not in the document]
pbrubeck: This code will produce the embedding DG element in `element.cell()` with `element.degree()`, regardless of `element.family()`. DG is a TPE if the cell is also TP. 
Serendipity can only be specified with an integer degree, so it might be better to construct DG via `TensorProductElement` if the degree is a tuple with different values, and we can use the standard `FiniteElement` constructor otherwise . 

Am I missing something? @rckirby any ideas?

firedrake/mg/utils.py
rckirby: When I was running the code with serendipity, it caught this exception.  Removing the exception worked without issue.  So probably we can keep that restriction out?

pbrubeck: These hack has been properly addressed in PR #2012

firedrake/preconditioners/pmg.py
```diff
@@ -95,124 +128,147 @@ def initialize(self, pc):
...
+                         for f in replace(form, replace_d).integrals()]) if isinstance(form, Form) else form
+
+        def coarsen_bcs(fbcs):
+            cbcs = []
+            for bc in fbcs:
+                cV_ = cV
+                for index in bc._indices:
+                    cV_ = cV_.sub(index)
+                cbc_value = self.coarsen_bc_value(bc, cV_)
+                cbcs.append(firedrake.DirichletBC(cV_, cbc_value,
```
wence-, line 213: `type(bc)(cV_, ...)`

I think maybe a bigger change (open an issue to do it) is to provide `reconstruct` methods for BC objects so that you don't have to dispatch based on type when doing this stuff, but not sure.

firedrake/preconditioners/pmg.py
```diff
@@ -95,124 +128,147 @@ def initialize(self, pc):
...
-        cctx = _SNESContext(cproblem, fctx.mat_type, fctx.pmat_type,
-                            appctx=fctx.appctx,
-                            pre_jacobian_callback=fctx._pre_jacobian_callback,
-                            pre_function_callback=fctx._pre_function_callback,
-                            post_jacobian_callback=fctx._post_jacobian_callback,
-                            post_function_callback=fctx._post_function_callback,
-                            options_prefix=fctx.options_prefix,
-                            transfer_manager=fctx.transfer_manager)
+        # FIXME setting up the _fine attribute triggers gmg injection.
+        # cctx._fine = fctx
```
wence-, line 262: This comes from the workaround in `_SNESContext.compute_operators` for a PETSc bug.

firedrake/preconditioners/pmg.py
```diff
@@ -529,51 +736,53 @@ def make_blas_kernels(self, Vf, Vc):
...
+        prolong_kernel = op2.Kernel(prolong_code, "prolongation", include_dirs=BLASLAPACK_INCLUDE.split(), ldargs=BLASLAPACK_LIB.split())
+        restrict_kernel = op2.Kernel(restrict_code, "restriction", include_dirs=BLASLAPACK_INCLUDE.split(), ldargs=BLASLAPACK_LIB.split())
+        return prolong_kernel, restrict_kernel
 
     @staticmethod
-    def get_nodes_1d(V):
-        # Return GLL nodes if V==CG or GL nodes if V==DG
-        from FIAT import quadrature
-        from FIAT.reference_element import DefaultLine
+    def get_line_element(V):
```
wence-, line 746: This looks like it's the same as the free function `get_line_element` above.

firedrake/preconditioners/pmg.py
```diff
@@ -529,51 +736,53 @@ def make_blas_kernels(self, Vf, Vc):
...
-        lam = 1.0E0 / np.prod(temp, axis=1)  # barycentric weights
-        J = np.subtract.outer(xdst, xsrc)
-        idx = np.argwhere(np.isclose(J, 0.0E0, 1E-14))
-        J[idx[:, 0], idx[:, 1]] = 1.0E0
-        J = lam/J
-        J[idx[:, 0], :] = 0.0E0
-        J[idx[:, 0], idx[:, 1]] = 1.0E0
-        J *= (1/np.sum(J, axis=1))[:, None]
-        return J
+    def get_line_nodes(V):
```
wence-, line 769: I think this is also repeated from above.

wence- (commented):
Modulo the relatively minor fixes, I think this is good.

firedrake/preconditioners/pmg.py
```diff
@@ -59,15 +60,39 @@ def coarsen_element(ele):
...
-            raise ValueError
-        elif degree == 1:
+        if N <= self.coarse_degree:
             raise ValueError
 
-        return ele.reconstruct(degree=degree // 2)
+        return PMGBase.reconstruct_degree(ele, max(N // 2, self.coarse_degree))
+
+    @staticmethod
+    def reconstruct_degree(ele, N):
```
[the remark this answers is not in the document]
wence-: Probably, but that requires much more invasive UFL rewriting. -> Revisit some other time.

firedrake/preconditioners/pmg.py
pbrubeck: Right, these are duplicates. I'm deleting the static methods from above.

wence- (approved):

wence-:
Thanks!

PR #1964, opened 2021-01-25
Rckirby/mg serendipity 3d

rckirby:
Multigrid is breaking on 3d extruded meshes when we use serendipity elements.  These are non-tensor-product elements on tensor product domains.  These are minimal fixes (and a test) that confirm MG now works.


rckirby:
We've made changes to support this alongside https://github.com/firedrakeproject/firedrake/pull/1963, so I'm closing this PR and helping out Pablo's.

PR #1971, opened 2021-02-04
Create new testing element and test interpolating onto it

ReubenHill:
Creates a new element, with all necessary steps from FIAT and UFL downwards, then tests interpolation with that element.

_OLD_
~Tests a new FInAT element with integral moment nodes and point evaluation nodes on an interval cell which is very useful for sanity checking. Name is very much up for negotiation!~

~NOTE: Currently failing for reasons unknown~

~Linked PRs:~

~https://github.com/FEniCS/fiat/pull/64~
~https://github.com/FEniCS/ufl/pull/47~
~https://github.com/FInAT/FInAT/pull/73~
~https://github.com/firedrakeproject/tsfc/pull/239~

### DIFF
--- a/tests/regression/test_interpolate.py
+++ b/tests/regression/test_interpolate.py
@@ -358,18 +358,3 @@ def test_interpolate_periodic_coords_max():
     # All nodes on the "seam" end up being 1, not 0.
     assert np.allclose(np.unique(continuous.dat.data_ro),
                        [0.25, 0.5, 0.75, 1])
-
-
-def test_basic_dual_eval_c0modified():
-    mesh = UnitIntervalMesh(1)
-    V = FunctionSpace(mesh, "C0 Modified", 3)
-    x = SpatialCoordinate(mesh)
-    expr = Constant(1.)
-    f = interpolate(expr, V)
-    assert np.allclose(f.dat.data_ro[f.cell_node_map().values], [1, 1, 1, 1/2])
-    expr = x[0]
-    f = interpolate(expr, V)
-    assert np.allclose(f.dat.data_ro[f.cell_node_map().values], [0, 1, 1/2, 1/3])
-    expr = x[0]**2
-    f = interpolate(expr, V)
-    assert np.allclose(f.dat.data_ro[f.cell_node_map().values], [0, 1, 1/3, 1/4])
--- a/tests/regression/test_interpolate.py
+++ b/tests/regression/test_interpolate.py
@@ -358,3 +358,18 @@ def test_interpolate_periodic_coords_max():
     # All nodes on the "seam" end up being 1, not 0.
     assert np.allclose(np.unique(continuous.dat.data_ro),
                        [0.25, 0.5, 0.75, 1])
+
+
+def test_basic_dual_eval_c0modified():
+    mesh = UnitIntervalMesh(1)
+    V = FunctionSpace(mesh, "C0 Modified", 3)
+    x = SpatialCoordinate(mesh)
+    expr = Constant(1.)
+    f = interpolate(expr, V)
+    assert np.allclose(f.dat.data_ro[f.cell_node_map().values], [1, 1, 1, 1/2])
+    expr = x[0]
+    f = interpolate(expr, V)
+    assert np.allclose(f.dat.data_ro[f.cell_node_map().values], [0, 1, 1/2, 1/3])
+    expr = x[0]**2
+    f = interpolate(expr, V)
+    assert np.allclose(f.dat.data_ro[f.cell_node_map().values], [0, 1, 1/3, 1/4])
--- a/tests/regression/test_interpolate_p3intmoments.py
+++ b/tests/regression/test_interpolate_p3intmoments.py
@@ -1,0 +1,137 @@
+import pytest
+import numpy as np
+from firedrake import *
+from FIAT.dual_set import DualSet
+from FIAT.finite_element import CiarletElement
+from FIAT.reference_element import UFCInterval
+from FIAT.functional import PointEvaluation, IntegralMoment
+from FIAT.quadrature import make_quadrature
+from FIAT.polynomial_set import ONPolynomialSet
+from finat.fiat_elements import ScalarFiatElement
+from tsfc.finatinterface import convert, as_fiat_cell
+import ufl
+
+ufcint = UFCInterval()
+
+
+# Purpose of these tests: test interpolation with a simple 1D element newly
+# created in a single location
+# New FIAT Element For P3 on an interval with moment nodes
+class P3IntMomentsDualSet(DualSet):
+    r"""
+    Dual set for an element, only defined for order 3 on interval cells, with
+    nodes:
+    ..math::
+     phi_0'(f) = f(0.)
+     phi_1'(f) = f(1.)
+     pfi_2'(f) = \int_{0}^{1} f(x) \,dx
+     pfi_3'(f) = \int_{0}^{1} x f(x) \,dx
+     """
+
+    def __init__(self, cell, order):
+        assert cell == UFCInterval() and order == 3
+        entity_ids = {0: {0: [0], 1: [1]},
+                      1: {0: [2, 3]}}
+        vertnodes = [PointEvaluation(cell, xx)
+                     for xx in cell.vertices]
+        Q = make_quadrature(cell, 3)
+        # 1st integral moment node is integral(1*f(x)*dx)
+        ones = np.asarray([1.0 for x in Q.pts])
+        # 2nd integral moment node is integral(x*f(x)*dx)
+        xs = np.asarray([x for (x,) in Q.pts])
+        intnodes = [IntegralMoment(cell, Q, ones),
+                    IntegralMoment(cell, Q, xs)]
+        nodes = vertnodes + intnodes
+        super().__init__(nodes, cell, entity_ids)
+
+
+class P3IntMoments(CiarletElement):
+    r"""
+    An element, only defined for order 3 on interval cells, with 4 nodes:
+    ..math::
+     phi_0'(f) = f(0.)
+     phi_1'(f) = f(1.)
+     pfi_2'(f) = \int_{0}^{1} f(x) \,dx
+     pfi_3'(f) = \int_{0}^{1} x f(x) \,dx
+     """
+
+    def __init__(self, cell, order):
+        assert cell == UFCInterval() and order == 3
+        poly_set = ONPolynomialSet(cell, 3)
+        dual_set = P3IntMomentsDualSet(cell, 3)
+        super().__init__(poly_set, dual_set, 3, 0)
+
+
+def test_fiat_p3intmoments():
+    # can create
+    el = P3IntMoments(UFCInterval(), 3)
+    # have expected number of nodes
+    assert len(el.dual_basis()) == 4
+    # dual eval gives expected values from sum of point evaluations
+    fns = (lambda x: x[0], lambda x: x[0]**2)
+    expected = ([0, 1, 1/2, 1/3], [0, 1, 1/3, 1/4])
+    for fn, expect in zip(fns, expected):
+        node_vals = []
+        for node in el.dual_basis():
+            pt_dict = node.pt_dict
+            node_val = 0.0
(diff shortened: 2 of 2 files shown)


wence-:
I think these tests are good idea. Can we pull all of the element definitions into the test file. Then we see what we need to expose in API for single-point-of-registration for new elements.

tests/regression/test_interpolate.py
```diff
@@ -358,3 +358,108 @@ def test_interpolate_periodic_coords_max():
...
+
+
+# Add to TSFC Supported Elements
+import tsfc
+tsfc.finatinterface.supported_elements["C0 Modified"] = FInAT_C0Modified
+
+
+# Register in UFL
+import ufl
+ufl.register_element("C0 Modified", "C0Mod", 0, H1, "identity", (3, 3), "interval")
```
ReubenHill, line 445: Need to replace with variant as suggested by @wence- but not sure of syntax yet

wence-: `FiniteElement(interval, "CG", 3, variant="interior-moment"`)` ?

ReubenHill: I'm afraid I don't follow `ufl.FiniteElement("CG", "interval", 3, variant="interior-moment")`  just creates a new ufl `ufl.FiniteElement` object rather than registering a name string which I surely need in order for the element to be created using `firedrake.FunctionSpace`

tests/regression/test_interpolate.py
```diff
@@ -358,3 +358,108 @@ def test_interpolate_periodic_coords_max():
...
+
+
+class FInAT_C0Modified(ScalarFiatElement):
+    def __init__(self, cell, degree):
+        super(FInAT_C0Modified, self).__init__(FIAT_C0Modified(cell, degree))
+
+
+# Add to TSFC Supported Elements
+import tsfc
+tsfc.finatinterface.supported_elements["C0 Modified"] = FInAT_C0Modified
```
miklos1, line 440: Ideally, you want to do this (and the UFL registration, too) in the `setUp` of the test, and then undo it in the `tearDown`, so that you only have the test element active for the duration of the test, rather than enabling it permanently from the moment this test file is loaded.

Not sure what the correct `pytest` syntax is for doing that, but you can probably figure it out.

wence-: Use a pytest [yield fixture](https://docs.pytest.org/en/stable/fixture.html#teardown-cleanup-aka-fixture-finalization)

tests/regression/test_interpolate.py
wence-:
```python
e = ufl.FiniteElement("CG", "interval", 3, variant="interior-moment")
V = FunctionSpace(mesh, e)
```

So now you just need the finatinterface to know how to handle the "interior-moment" variant.

ReubenHill: Ah - but then I need to modify `tsfc`'s `tsfc/finatinterface.py::convert_finiteelement()` function (see below). There is no externally modifiable TSFC interface for doing this as far as I can tell so I would actually have to change TSFC to let me support this element. My current solution at least allows everything to be gathered into one place. This seems to be exactly the point @rckirby noted in [this comment](https://github.com/FInAT/FInAT/issues/75#issuecomment-781471185) in https://github.com/FInAT/FInAT/issues/75.
```python
# Base finite elements first
@convert.register(ufl.FiniteElement)
def convert_finiteelement(element, **kwargs):
    cell = as_fiat_cell(element.cell())

    ...
    lmbda = supported_elements[element.family()]
    ...

    kind = element.variant()
    if kind is None:
        kind = 'spectral' if element.cell().cellname() == 'interval' else 'equispaced'  # default variant

    if element.family() == "Lagrange":
        if kind == 'equispaced':
            lmbda = finat.Lagrange
        elif kind == 'spectral' and element.cell().cellname() == 'interval':
            lmbda = finat.GaussLobattoLegendre
        elif kind in ['mgd', 'feec', 'qb', 'mse']:
            degree = element.degree()
            shift_axes = kwargs["shift_axes"]
            restriction = kwargs["restriction"]
            deps = {"shift_axes", "restriction"}
            return finat.RuntimeTabulated(cell, degree, variant=kind, shift_axes=shift_axes, restriction=restriction), deps
        else:
            raise ValueError("Variant %r not supported on %s" % (kind, element.cell()))  # Fails here
```
Is there some ability for me to override this function that I am missing?

tests/regression/test_interpolate.py
ReubenHill: Sadly there is no way for me to do the same thing with UFL - I can call `register_element` but there is no `deregister_element`. I'll need to manually remove the key from the dictionary `ufl.finitelement.elementlist.ufl_elements` which I don't believe is part of the public UFL interface.

ReubenHill: Though I suppose `tsfc.finatinterface.supported_elements` isn't parti of the TSFC public interface either

ReubenHill: Done - see 6fc2de8286cd951e46dc68d9172d342562d75d3f

tests/regression/test_interpolate.py
wence-: You have to replace this dispatch with your own one.

```
from tsfc.finatinterface import convert

old_converter = convert.dispatch(ufl.FiniteElement)

def my_converter(element, **kwargs):
    if element.variant() == "interior-moment" and element.family() == "CG":
        return finat.FancyElement(...)
    else:
        return old_convert(element, **kwargs)

convert.register(ufl.FiniteElement, my_converter)

# To undo this afterr the test has finished do

convert.register(ufl.FiniteElement, old_convert)
```

wence-: Yes this is ugly, we're exactly trying to see where it is ugly and what we would need to do to improve this.

ReubenHill: Done, see 91e98aa9e9db0142339013740851ae823f48620f

tests/regression/test_interpolate.py
```diff
@@ -358,3 +358,126 @@ def test_interpolate_periodic_coords_max():
...
+    expr = x[0]
+    # Account for cell and corresponding expression being flipped onto
+    # reference cell before reaching FIAT
+    expr_fiat = 1-x[0]
+    f = interpolate(expr, V)
+    assert np.allclose(f.dat.data_ro[f.cell_node_map().values], [node(expr_fiat) for node in f.function_space().finat_element.fiat_equivalent.dual_basis()])
+    expr = x[0]**2
+    expr_fiat = (1-x[0])**2
+    f = interpolate(expr, V)
+    assert np.allclose(f.dat.data_ro[f.cell_node_map().values], [node(expr_fiat) for node in f.function_space().finat_element.fiat_equivalent.dual_basis()])
```
wence-, line 483: Can we put all this new stuff in a separate file, rather than with `test_interpolate.py`?

tests/regression/test_interpolate.py
```diff
@@ -358,3 +358,126 @@ def test_interpolate_periodic_coords_max():
...
+
+# New FIAT Element For P3 on an interval with moment nodes
+
+from FIAT.dual_set import DualSet
+from FIAT.finite_element import CiarletElement
+from FIAT.reference_element import UFCInterval
+from FIAT.functional import PointEvaluation, IntegralMoment
+from FIAT.quadrature import make_quadrature
+from FIAT.polynomial_set import ONPolynomialSet
+ufcint = UFCInterval()
```
wence-, line 373: Prefer not to declare global variables.

tests/regression/test_interpolate.py
```diff
@@ -358,3 +358,126 @@ def test_interpolate_periodic_coords_max():
...
+
+
+class FIAT_P3IntMomentsDualSet(DualSet):
```
wence-, line 376: No need for this odd naming, plus add docstring.

tests/regression/test_interpolate.py
```diff
@@ -358,3 +358,126 @@ def test_interpolate_periodic_coords_max():
...
+        ones = np.asarray([1.0 for x in Q.pts])
+        # 2nd integral moment node is integral(x*f(x)*dx)
+        xs = np.asarray([x for (x,) in Q.pts])
+        intnodes = [IntegralMoment(ufcint, Q, ones),
+                    IntegralMoment(ufcint, Q, xs)]
+        nodes = vertnodes + intnodes
+        super(FIAT_P3IntMomentsDualSet, self).__init__(nodes, ufcint, entity_ids)
+
+
+class FIAT_P3IntMoments(CiarletElement):
```
wence-, line 395: Just `P3IntMoments` ?

tests/regression/test_interpolate.py
```diff
@@ -358,3 +358,126 @@ def test_interpolate_periodic_coords_max():
...
+     phi_1'(f) = f(1.)
+     pfi_2'(f) = \int_{0}^{1} f(x) \,dx
+     pfi_3'(f) = \int_{0}^{1} x f(x) \,dx
+     """
+
+    def __init__(self, cell, order):
+        assert cell == ufcint and order == 3
+        poly_set = ONPolynomialSet(ufcint, 3)
+        dual_set = FIAT_P3IntMomentsDualSet(ufcint, 3)
+        super(FIAT_P3IntMoments, self).__init__(poly_set, dual_set, 3, 0)
```
wence-, line 409: Modern Python is much nicer, we can write `super().__init__(...)`

tests/regression/test_interpolate.py
```diff
@@ -358,3 +358,126 @@ def test_interpolate_periodic_coords_max():
...
+
+
+def test_fiat_p3intmoments():
+    # can create
+    el = FIAT_P3IntMoments(UFCInterval(), 3)
+    # dual eval gives expected values from sum of point evaluations
+    fns = (lambda x: x[0], lambda x: x[0]**2)
+    expected = ([0, 1, 1/2, 1/3], [0, 1, 1/3, 1/4])
+    for i in range(len(fns)):
```
wence-, line 418: `for fn, expect in zip(fns, expected):`

tests/regression/test_interpolate.py
```diff
@@ -358,3 +358,126 @@ def test_interpolate_periodic_coords_max():
...
+        node_vals = []
+        for node in el.dual.nodes:
```
wence-, line 420: Public API is to ask for `el.dual_basis()` I think?

tests/regression/test_interpolate.py
```diff
@@ -358,3 +358,126 @@ def test_interpolate_periodic_coords_max():
...
+            pt_dict = node.pt_dict
+            node_val = 0.0
+            for pt in pt_dict:
+                for (w, _) in pt_dict[pt]:
+                    node_val += w * fns[i](pt)
+            node_vals.append(node_val)
+        assert len(node_vals) == len(expected[i])
```
wence-, line 427: This is slightly backwards, I think `assert len(el.dual.nodes) == 4`?

tests/regression/test_interpolate.py
```diff
@@ -358,3 +358,126 @@ def test_interpolate_periodic_coords_max():
...
+
+# Test New Element Dual Evaluation
+def test_basic_dual_eval_p3intmoments(add_p3intmoments_tsfc):
+    mesh = UnitIntervalMesh(1)
+    e = ufl.FiniteElement("CG", "interval", 3, variant="interior-moment")
+    V = FunctionSpace(mesh, e)
+    x = SpatialCoordinate(mesh)
+    expr = Constant(1.)
+    f = interpolate(expr, V)
+    assert np.allclose(f.dat.data_ro[f.cell_node_map().values], [node(expr) for node in f.function_space().finat_element.fiat_equivalent.dual_basis()])
```
wence-, line 473: I know that we don't enforce line length limits strictly, but...

wence-: Maybe 
```python
dual = f.function_space().finat_element.fiat_equivalent.dual_basis()`
assert np.allclose(..., [node(expr) for node in dual])
```

wence- (changes requested):

ReubenHill:
@wence- all review comments addressed and squashed into single commit

PR #1975, opened 2021-02-16
Receives and sends for ensemble communicator

colinjcotter:
Adding blocking and non-blocking sends and receives for the Ensemble communicator.

(written by Wei Pan).

### DIFF
--- a/firedrake/ensemble.py
+++ b/firedrake/ensemble.py
@@ -1,4 +1,5 @@
 from pyop2.mpi import MPI
+from itertools import zip_longest
 
 __all__ = ("Ensemble", )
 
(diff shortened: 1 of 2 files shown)


ReubenHill:
For the uninformed like me, what's the problem this PR is solving?

dham:
> For the uninformed like me, what's the problem this PR is solving?

It's the ability to send a whole Function from one ensemble member to another. Previously ensemble members could only send statistics to each other. This was always in the Ensemble API (hence the `NotImplementedError`s), it's just nobody had got around to the implementation.

wence- (changes requested):
Very minor change, otherwise looks good, thanks.

wence-:
Can you please fix the flake8 failures. Run `make lint` in the firedrake repo to see them. It should return no errors.

whotookmyname:
Done. I've fixed the lint errors and added 'status' to 'recv'. Thanks.

wence-:
I realised this would have problems for non-blocking send/receive of mixed functions (since in those cases the PETSc Vec and the Firedrake function do not share memory). The context manager in `irecv` would do the writeback before the request had completed.

I've pushed fixes for this (and a test). The interface has changed slightly, you now get back a list of requests from `isend` and `irecv` which you can wait on with `MPI.Request.waitall`.

wence-:
Please check that things continue to work for you.

wence- (approved):

wence- (approved):

PR #1977, opened 2021-02-18
Migrate some of the lightweight checks to actions

wence-:
Means that these run in parallel to the main build.

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -1,0 +1,27 @@
+name: Lint
+
+on:
+  # Push to master or PR
+  push:
+    branches:
+      - master
+  pull_request_target:
+
+jobs:
+  test:
+    # The type of runner that the job will run on
+    runs-on: ubuntu-latest
+    # Steps represent a sequence of tasks that will be executed as
+    # part of the jobs
+    steps:
+      - uses: actions/checkout@v2
+      - name: Setup python
+        uses: actions/setup-python@v2
+        with:
+          python-version: 3.8
+      - name: Install linter
+        run: |
+          pip install flake8 pylint
+      - name: Lint
+        run: |
+          make lint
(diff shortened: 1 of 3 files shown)


dham:
Looks like the Zenodo canary action didn't run.

wence-:
I set it to only run on pushes to master.

wence-:
> I set it to only run on pushes to master.

I guess maybe it makes sense to run in PRs too, in case we break stuff on our end.

wence-:
> I guess maybe it makes sense to run in PRs too, in case we break stuff on our end.

Did that.

ReubenHill:
Nice - so something that fails linting to a PR or draft PR, the linting will fail and the build step will be skipped?

wence-:
> Nice - so if I push something that fails linting to a PR or draft PR, the linting will fail and the build step will be skipped?

The linting will fail, but the build tests will continue. These are now separate (parallel) checks that must pass.

ReubenHill:
> > Nice - so if I push something that fails linting to a PR or draft PR, the linting will fail and the build step will be skipped?
> 
> The linting will fail, but the build tests will continue. These are now separate (parallel) checks that must pass.

Is it possible to force jenkins to not run in that case? Though I'm in two minds if that is desirable behaviour

wence-:
> Is it possible to force jenkins to not run in that case? Though I'm in two minds if that is desirable behaviour

One would have to poke jenkins via some API to say "stop this build". Probably possible, but the jenkins docs are Not. Good.

I would like to migrate everything to actions, at which point it might be easier.

But AFAIAA, everyone opens PRs because they want the test suite to run, because no-one runs it locally. So stopping it straight away for lint purposes doesn't really help that much.

danshapero (dismissed):
Looks good to me! Whether a lint failure should trigger an immediate build failure or not is maybe a policy decision for another day. But having lint failures show up earlier in the pipeline will definitely help give more rapid feedback which is good.

wence-:
I realise I should use the pull_request_target selector instead of pull_request. Will do that tomorrow

PR #1982, opened 2021-02-24
Reinjection in Transfermanager

FabianL1908:
The transfermanager currently does not do any reinjection of coefficients, which causes problems for nonlinear problems. This pull request reinjects all coefficients every time, which is unnecessary in most cases. Therefore, it is left to provide a mechanism for tracking the coefficients that have to be reinjected.


firedrake/mg/ufl_utils.py
```diff
@@ -167,10 +169,35 @@ def coarsen_function(expr, self, coefficient_mapping=None):
...
+            for coeff in coeffs:
+                #print("Considering coeff = %s" % coeff)
+                if isinstance(coeff, firedrake.Function):
+                    if hasattr(coeff, '_child'):
+                        #print("Injecting %s -> %s" % (coeff, coeff._child))
+                        manager.inject(coeff, coeff._child)
+
+            import sys; sys.stdout.flush()
+        #print("Adding hook to DM with prefix %s for %s -> %s" % (dm.getOptionsPrefix(), expr, new))
+        dm.addCoarsenHook(None, inject_on_restrict)
```
wence-, line 198: I think we need to call `dm.removeCoarsenHook` in a teardown

wence-, line 198: I think this pushes "inject_on_restrict` onto petsc4py's list of hooks every time. And then inject_on_restrict injects _everything_ every time. Which is rather too much.

firedrake/mg/ufl_utils.py
```diff
@@ -167,10 +169,35 @@ def coarsen_function(expr, self, coefficient_mapping=None):
...
+            if finectx.Jp is not None:
+                coeffs += finectx.Jp.coefficients()
+            coeffs = tuple(OrderedDict.fromkeys(coeffs))
+
+            for coeff in coeffs:
+                #print("Considering coeff = %s" % coeff)
+                if isinstance(coeff, firedrake.Function):
+                    if hasattr(coeff, '_child'):
+                        #print("Injecting %s -> %s" % (coeff, coeff._child))
+                        manager.inject(coeff, coeff._child)
```
wence-, line 194: I think this function wants to be something like:

```
def inject_on_restrict(finedm, mrestrict, rscale, inject, coarse, fine=None, coarse=None):
    manager.inject(fine, coarse)

dm.addCoarsenHook(None, inject_on_restrict, kargs={"fine": expr, "coarse": expr._child})
```

We still need to handle the cleanup though.

PR #1983, opened 2021-02-24
Refactor assemble

connorjward:
This pull request:

- Lifts zeroing the tensor out of `_assemble()`. This means that `get_scalar()`, `get_vector()` and `get_matrix()` only need to return `tensor`, rather than `tensor, zeros, lambda: tensor`.
- Caches parloops on the form instead of constructing a monolithic assembly 'thunk'. This makes it straightforward to pass arguments into the parloops at execution time and will speed `assemble()` up significantly.

The changes will introduce a small amount of overhead where `create_assembly_callable()` is used because we do more logic every time the assembly callable is called. If this overhead is problematic we could also cache the callables that apply the boundary conditions.

What do people think? I know the code needs documenting and the caching does not yet work but I wanted to check before spending more time on this.

### DIFF
--- a/firedrake/assemble.py
+++ b/firedrake/assemble.py
@@ -1,7 +1,7 @@
 import functools
 import operator
 from collections import OrderedDict, defaultdict
-from enum import IntEnum
+from enum import Enum, IntEnum
 from itertools import chain
 
 import firedrake
@@ -16,6 +16,9 @@ from firedrake.utils import ScalarType
 from pyop2 import op2
 from pyop2.exceptions import MapValueError, SparsityFormatError
 
+from firedrake.petsc import PETSc
+PETSc.Sys.popErrorHandler()
+
 
 __all__ = ("assemble",)
 
@@ -25,11 +28,15 @@ class AssemblyRank(IntEnum):
     VECTOR = 1
     MATRIX = 2
 
+class AssemblyType(Enum):
+    SOLUTION = "solution"
+    RESIDUAL = "residual"
+
 
 @annotate_assemble
 def assemble(expr, tensor=None, bcs=None, form_compiler_parameters=None,
              mat_type=None, sub_mat_type=None,
-             appctx={}, options_prefix=None, assemble_now=True, **kwargs):
+             appctx={}, options_prefix=None, assembly_type="solution", **kwargs):
     # TODO: Add warning that caching on compiler params is not done.
     r"""Evaluate expr.
 
@@ -86,6 +93,8 @@ def assemble(expr, tensor=None, bcs=None, form_compiler_parameters=None,
         raise ValueError("Can't use 'nest', set 'mat_type' instead")
     if "collect_loops" in kwargs or "allocate_only" in kwargs:
         raise RuntimeError
+    if assembly_type not in ("solution", "residual"):
+        raise ValueError("assembly_type must be either 'solution' or 'residual'")
 
     diagonal = kwargs.pop("diagonal", False)
     if len(kwargs) > 0:
@@ -94,30 +103,31 @@ def assemble(expr, tensor=None, bcs=None, form_compiler_parameters=None,
     if isinstance(expr, (ufl.form.Form, slate.TensorBase)):
         mat_type, sub_mat_type = get_mat_type(mat_type, sub_mat_type, expr.arguments())
         assembly_rank = _get_assembly_rank(expr, diagonal)
+        # bcs = solving._extract_bcs(bcs)
         if assembly_rank == AssemblyRank.SCALAR:
-            return _assemble_scalar(expr, tensor, bcs=bcs, form_compiler_parameters=form_compiler_parameters, mat_type=mat_type, sub_mat_type=sub_mat_type, appctx=appctx, options_prefix=options_prefix, assemble_now=assemble_now)
+            return _assemble_scalar(expr, tensor, bcs=bcs, form_compiler_parameters=form_compiler_parameters, mat_type=mat_type, sub_mat_type=sub_mat_type, appctx=appctx, options_prefix=options_prefix, assembly_type=assembly_type)
         elif assembly_rank == AssemblyRank.VECTOR:
-            return _assemble_vector(expr, tensor, bcs=bcs, form_compiler_parameters=form_compiler_parameters, mat_type=mat_type, sub_mat_type=sub_mat_type, diagonal=diagonal, appctx=appctx, options_prefix=options_prefix, assemble_now=assemble_now)
+            return _assemble_vector(expr, tensor, bcs=bcs, form_compiler_parameters=form_compiler_parameters, mat_type=mat_type, sub_mat_type=sub_mat_type, diagonal=diagonal, appctx=appctx, options_prefix=options_prefix, assembly_type=assembly_type)
         elif assembly_rank == AssemblyRank.MATRIX:
-            return _assemble_matrix(expr, tensor, bcs=bcs, form_compiler_parameters=form_compiler_parameters, mat_type=mat_type, sub_mat_type=sub_mat_type, appctx=appctx, options_prefix=options_prefix, assemble_now=assemble_now)
+            return _assemble_matrix(expr, tensor, bcs=bcs, form_compiler_parameters=form_compiler_parameters, mat_type=mat_type, sub_mat_type=sub_mat_type, appctx=appctx, options_prefix=options_prefix, assembly_type=assembly_type)
     elif isinstance(expr, ufl.core.expr.Expr):
         return assemble_expressions.assemble_expression(expr)
     else:
         raise TypeError(f"Unable to assemble: {expr}")
 
 
-def _assemble_scalar(expr, scalar, assemble_now, **kwargs):
+def _assemble_scalar(expr, scalar, **kwargs):
     if len(expr.arguments()) != 0:
         raise ValueError("Can't assemble a 0-form with arguments")
     if scalar:
         raise ValueError("Can't assemble 0-form into existing tensor")
 
     scalar = _make_scalar()
-    _assemble_expr(expr, scalar, diagonal=False, assemble_now=assemble_now, **kwargs)
+    _assemble_expr(expr, scalar, diagonal=False, **kwargs)
     return scalar.data[0]
 
 
-def _assemble_vector(expr, vector, *, diagonal, assemble_now, **kwargs):
+def _assemble_vector(expr, vector, *, diagonal, **kwargs):
     if diagonal:
         test, trial = expr.arguments()
         if test.function_space() != trial.function_space():
@@ -130,11 +140,11 @@ def _assemble_vector(expr, vector, *, diagonal, assemble_now, **kwargs):
         vector.dat.zero()
     else:
         vector = _make_vector(test)
-    _assemble_expr(expr, vector, diagonal=diagonal, assemble_now=assemble_now, **kwargs)
+    _assemble_expr(expr, vector, diagonal=diagonal, **kwargs)
     return vector
 
 
-def _assemble_matrix(expr, matrix, *, mat_type, sub_mat_type, bcs, options_prefix, appctx, form_compiler_parameters, assemble_now):
+def _assemble_matrix(expr, matrix, *, mat_type, sub_mat_type, bcs, options_prefix, appctx, form_compiler_parameters, **kwargs):
     if matrix:
         if mat_type != "matfree":
             matrix.M.zero()
@@ -146,7 +156,7 @@ def _assemble_matrix(expr, matrix, *, mat_type, sub_mat_type, bcs, options_prefi
     if mat_type == "matfree":
         matrix.assemble()
     else:
-        _assemble_expr(expr, matrix, diagonal=False, mat_type=mat_type, sub_mat_type=sub_mat_type, bcs=bcs, options_prefix=options_prefix, appctx=appctx, form_compiler_parameters=form_compiler_parameters, assemble_now=assemble_now)
+        _assemble_expr(expr, matrix, diagonal=False, mat_type=mat_type, sub_mat_type=sub_mat_type, bcs=bcs, options_prefix=options_prefix, appctx=appctx, form_compiler_parameters=form_compiler_parameters, **kwargs)
         matrix.M.assemble()
     return matrix
 
@@ -159,7 +169,7 @@ def _make_vector(test):
     return firedrake.Function(test.function_space())
 
 
-def _assemble_expr(expr, tensor, *, bcs, diagonal, assemble_now, **kwargs):
+def _assemble_expr(expr, tensor, *, bcs, diagonal, assembly_type, **kwargs):
     bcs = _preprocess_bcs(bcs)
 
     # We cache the parloops on the form but keep track of the tensor. If the
@@ -185,7 +195,7 @@ def _assemble_expr(expr, tensor, *, bcs, diagonal, assemble_now, **kwargs):
     assembly_rank = _get_assembly_rank(expr, diagonal)
 
     dir_bcs = tuple(bc for bc in bcs if isinstance(bc, DirichletBC))
-    _apply_dirichlet_bcs(tensor, dir_bcs, assembly_rank, diagonal=diagonal, assemble_now=assemble_now)
+    _apply_dirichlet_bcs(tensor, dir_bcs, assembly_rank=assembly_rank, diagonal=diagonal, assembly_type=assembly_type)
 
     eq_bcs = tuple(bc for bc in bcs if isinstance(bc, EquationBC))
     if eq_bcs and diagonal:
@@ -193,10 +203,14 @@ def _assemble_expr(expr, tensor, *, bcs, diagonal, assemble_now, **kwargs):
     for bc in eq_bcs:
         if assembly_rank == AssemblyRank.VECTOR:
             bc.zero(tensor)
-        _assemble_expr(bc.f, tensor, bcs=bc.bcs, diagonal=diagonal, **kwargs)
+        _assemble_expr(bc.f, tensor, bcs=bc.bcs, diagonal=diagonal, assembly_type=assembly_type, **kwargs)
 
 
 def _preprocess_bcs(bcs):
+    # Might have gotten here without `EquationBC` objects preprocessed.
+    # if any(isinstance(bc, EquationBC) for bc in bcs):
+    #     return tuple(bc.extract_form("F") for bc in bcs)
+    # return ()
     return tuple(bc.extract_form("F") for bc in solving._extract_bcs(bcs))
 
 
@@ -258,7 +272,7 @@ def create_assembly_callable(expr, tensor=None, bcs=None, form_compiler_paramete
     """
     if tensor is None:
         raise ValueError("Have to provide tensor to write to")
-    return functools.partial(assemble, expr, tensor, bcs, form_compiler_parameters, mat_type, sub_mat_type, assemble_now=False)
+    return functools.partial(assemble, expr, tensor, bcs, form_compiler_parameters, mat_type, sub_mat_type, assembly_type="residual")
 
 
 def _make_matrix(expr, mat_type, sub_mat_type, *, bcs,
@@ -437,9 +451,9 @@ def vector_arg(access, get_map, i, *, function, V):
         return function.dat[i](access, map_)
 
 
-def _apply_dirichlet_bcs(tensor, bcs, assembly_rank, *,
+def _apply_dirichlet_bcs(tensor, bcs, *, assembly_rank,
                          diagonal,
-                         assemble_now):
+                         assembly_type):
     """Apply boundary conditions to a tensor.
 
     :arg tensor: The tensor.
@@ -485,13 +499,19 @@ def _apply_dirichlet_bcs(tensor, bcs, assembly_rank, *,
                     raise RuntimeError("Unhandled BC case")
     elif assembly_rank == AssemblyRank.VECTOR:
         for bc in bcs:
-            if assemble_now:
+            if assembly_type == AssemblyType.SOLUTION.value:
                 if diagonal:
                     bc.set(tensor, 1)
                 else:
                     bc.apply(tensor)
-            else:
+            elif assembly_type == AssemblyType.RESIDUAL.value:
                 bc.zero(tensor)
+            else:
+                raise AssertionError
+    elif assembly_rank == AssemblyRank.SCALAR:
+        pass
+    else:
+        raise AssertionError
 
 
 def _make_parloop_inner(expr, create_op2arg, *, assembly_rank, diagonal,
--- a/firedrake/assemble.py
+++ b/firedrake/assemble.py
@@ -17,7 +17,7 @@ from pyop2 import op2
 from pyop2.exceptions import MapValueError, SparsityFormatError
 
 
-__all__ = ("assemble", )
+__all__ = ("assemble",)
 
 
 class AssemblyRank(IntEnum):
@@ -161,14 +161,24 @@ def _make_vector(test):
 
 def _assemble_expr(expr, tensor, *, bcs, diagonal, assemble_now, **kwargs):
     bcs = _preprocess_bcs(bcs)
-    # TODO: Reimplement this when PyOP2 functionality added.
-    # if "parloops" in expr._cache:
-    #     parloops = expr._cache["parloops"]
-    # else:
-    #     parloops = _make_parloops(expr, tensor, bcs=bcs, diagonal=diagonal, **kwargs)
-    #     expr._cache["parloops"] = parloops
-    # parloop.compute(out=tensor.dat)
-    parloops = _make_parloops(expr, tensor, bcs=bcs, diagonal=diagonal, **kwargs)
+
+    # We cache the parloops on the form but keep track of the tensor. If the
+    # tensor is changed then we need to regenerate the parloop (until parloops
+    # stop depending on data).
+    # When we generate a new parloop the old one is removed from the cache to
+    # prevent leaking memory.
+    cache_key = "parloops"
+    if cache_key in expr._cache:
+        cached_tensor, parloops = expr._cache[cache_key]
+        if cached_tensor is not tensor:
+            parloops = None
+            del expr._cache[cache_key]
+    else:
+        parloops = None
+
+    if not parloops:
+        parloops = _make_parloops(expr, tensor, bcs=bcs, diagonal=diagonal, **kwargs)
+        expr._cache[cache_key] = (tensor, parloops)
     for parloop in parloops:
         parloop.compute()
 
--- a/firedrake/assemble.py
+++ b/firedrake/assemble.py
@@ -10,7 +10,7 @@ import ufl
 from firedrake import (assemble_expressions, matrix, parameters, solving,
                        tsfc_interface, utils)
 from firedrake.adjoint import annotate_assemble
-from firedrake.bcs import DirichletBC, EquationBC
+from firedrake.bcs import DirichletBC, EquationBC, EquationBCSplit
 from firedrake.slate import slac, slate
 from firedrake.utils import ScalarType
 from pyop2 import op2
@@ -103,7 +103,7 @@ def assemble(expr, tensor=None, bcs=None, form_compiler_parameters=None,
     if isinstance(expr, (ufl.form.Form, slate.TensorBase)):
         mat_type, sub_mat_type = get_mat_type(mat_type, sub_mat_type, expr.arguments())
         assembly_rank = _get_assembly_rank(expr, diagonal)
-        # bcs = solving._extract_bcs(bcs)
+        bcs = solving._extract_bcs(bcs)
         if assembly_rank == AssemblyRank.SCALAR:
             return _assemble_scalar(expr, tensor, bcs=bcs, form_compiler_parameters=form_compiler_parameters, mat_type=mat_type, sub_mat_type=sub_mat_type, appctx=appctx, options_prefix=options_prefix, assembly_type=assembly_type)
         elif assembly_rank == AssemblyRank.VECTOR:
@@ -197,7 +197,7 @@ def _assemble_expr(expr, tensor, *, bcs, diagonal, assembly_type, **kwargs):
     dir_bcs = tuple(bc for bc in bcs if isinstance(bc, DirichletBC))
     _apply_dirichlet_bcs(tensor, dir_bcs, assembly_rank=assembly_rank, diagonal=diagonal, assembly_type=assembly_type)
 
-    eq_bcs = tuple(bc for bc in bcs if isinstance(bc, EquationBC))
+    eq_bcs = tuple(bc for bc in bcs if isinstance(bc, EquationBCSplit))
     if eq_bcs and diagonal:
         raise NotImplementedError("Diagonal assembly and EquationBC not supported")
     for bc in eq_bcs:
@@ -208,10 +208,11 @@ def _assemble_expr(expr, tensor, *, bcs, diagonal, assembly_type, **kwargs):
 
 def _preprocess_bcs(bcs):
     # Might have gotten here without `EquationBC` objects preprocessed.
-    # if any(isinstance(bc, EquationBC) for bc in bcs):
-    #     return tuple(bc.extract_form("F") for bc in bcs)
-    # return ()
-    return tuple(bc.extract_form("F") for bc in solving._extract_bcs(bcs))
+    bcs_copy = list(bcs)
+    for i, bc in enumerate(bcs_copy):
+        if isinstance(bc, EquationBC):
+            bcs_copy[i] = bc.extract_form("F")
+    return tuple(bcs_copy)
 
 
 def get_mat_type(mat_type, sub_mat_type, arguments):
--- a/firedrake/assemble.py
+++ b/firedrake/assemble.py
@@ -23,7 +23,7 @@ PETSc.Sys.popErrorHandler()
 __all__ = ("assemble",)
 
 
-class AssemblyRank(IntEnum):
+class _AssemblyRank(IntEnum):
     SCALAR = 0
     VECTOR = 1
     MATRIX = 2
@@ -36,6 +36,12 @@ _AssemblyOpts = namedtuple("_AssemblyOpts", ["diagonal",
                                              "sub_mat_type",
                                              "appctx",
                                              "options_prefix"])
+"""Container to hold immutable assembly options.
+
+.. note::
+    When we stop supporting Python 3.6 it may be good to add the :arg defaults:
+    argument to this.
+"""
 
 
 @annotate_assemble
@@ -121,11 +127,11 @@ def assemble_form(expr, tensor=None, bcs=None, *,
                          mat_type, sub_mat_type, appctx, options_prefix)
 
     assembly_rank = _get_assembly_rank(expr, diagonal)
-    if assembly_rank == AssemblyRank.SCALAR:
+    if assembly_rank == _AssemblyRank.SCALAR:
         return _assemble_scalar(expr, tensor, bcs, opts)
-    elif assembly_rank == AssemblyRank.VECTOR:
+    elif assembly_rank == _AssemblyRank.VECTOR:
         return _assemble_vector(expr, tensor, bcs, opts)
-    elif assembly_rank == AssemblyRank.MATRIX:
+    elif assembly_rank == _AssemblyRank.MATRIX:
         return _assemble_matrix(expr, tensor, bcs, opts)
     else:
         raise AssertionError
@@ -178,13 +184,13 @@ def _get_assembly_rank(expr, diagonal):
     rank = len(expr.arguments())
     if diagonal:
         assert rank == 2
-        return AssemblyRank.VECTOR
+        return _AssemblyRank.VECTOR
     if rank == 0:
-        return AssemblyRank.SCALAR
+        return _AssemblyRank.SCALAR
     if rank == 1:
-        return AssemblyRank.VECTOR
+        return _AssemblyRank.VECTOR
     if rank == 2:
-        return AssemblyRank.MATRIX
+        return _AssemblyRank.MATRIX
     raise AssertionError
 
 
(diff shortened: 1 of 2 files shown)


firedrake/assemble.py
```diff
@@ -91,21 +92,104 @@ def assemble(expr, tensor=None, bcs=None, form_compiler_parameters=None,
...
+    return op2.Global(1, [0.0], dtype=utils.ScalarType)
+
+
+def _make_vector(test):
+    return firedrake.Function(test.function_space())
+
+
+def _assemble_expr(expr, tensor, *, bcs, diagonal, assemble_now, **kwargs):
+    bcs = _preprocess_bcs(bcs)
+    # TODO: Reimplement this when PyOP2 functionality added.
```
connorjward, line 164: This is where I intend to cache the parloops

firedrake/assemble.py
```diff
@@ -91,21 +92,104 @@ def assemble(expr, tensor=None, bcs=None, form_compiler_parameters=None,
...
-                          options_prefix=options_prefix)
-        for l in loops:
-            m = l()
-        return m
+        mat_type, sub_mat_type = get_mat_type(mat_type, sub_mat_type, expr.arguments())
+        assembly_rank = _get_assembly_rank(expr, diagonal)
+        if assembly_rank == AssemblyRank.SCALAR:
+            return _assemble_scalar(expr, tensor, bcs=bcs, form_compiler_parameters=form_compiler_parameters, mat_type=mat_type, sub_mat_type=sub_mat_type, appctx=appctx, options_prefix=options_prefix, assemble_now=assemble_now)
+        elif assembly_rank == AssemblyRank.VECTOR:
+            return _assemble_vector(expr, tensor, bcs=bcs, form_compiler_parameters=form_compiler_parameters, mat_type=mat_type, sub_mat_type=sub_mat_type, diagonal=diagonal, appctx=appctx, options_prefix=options_prefix, assemble_now=assemble_now)
```
wence-, line 100: style: maybe wrap the common arguments into a dict

```python
kwargs = {"bcs": bcs, ...}
if assembly_rank == ...:
    return assemble_scalar(expr, tensor, **kwargs)
elif ...
    return assemble_vector(expr, tensor, diagonal=diagonal, **kwargs)
...
```

wence-: (My screen is not this wide, or my eyes are not good enough for tiny fonts!)

firedrake/assemble.py
```diff
@@ -91,21 +92,104 @@ def assemble(expr, tensor=None, bcs=None, form_compiler_parameters=None,
...
+        raise ValueError("Can't assemble a 0-form with arguments")
+    if scalar:
+        raise ValueError("Can't assemble 0-form into existing tensor")
+
+    scalar = _make_scalar()
+    _assemble_expr(expr, scalar, diagonal=False, assemble_now=assemble_now, **kwargs)
+    return scalar.data[0]
+
+
+def _assemble_vector(expr, vector, *, diagonal, assemble_now, **kwargs):
```
wence-, line 120: I did not realise that you could request _required_ keyword arguments like this!

firedrake/assemble.py
```diff
@@ -91,21 +92,104 @@ def assemble(expr, tensor=None, bcs=None, form_compiler_parameters=None,
...
+    #     expr._cache["parloops"] = parloops
+    # parloop.compute(out=tensor.dat)
+    parloops = _make_parloops(expr, tensor, bcs=bcs, diagonal=diagonal, **kwargs)
+    for parloop in parloops:
+        parloop.compute()
+
+    assembly_rank = _get_assembly_rank(expr, diagonal)
+
+    dir_bcs = tuple(bc for bc in bcs if isinstance(bc, DirichletBC))
+    _apply_dirichlet_bcs(tensor, dir_bcs, assembly_rank, diagonal=diagonal, assemble_now=assemble_now)
```
wence-, line 178: These also cons parloops, not sure if we want to save them too.

firedrake/assemble.py
```diff
@@ -165,26 +248,12 @@ def create_assembly_callable(expr, tensor=None, bcs=None, form_compiler_paramete
...
-                      diagonal=diagonal,
-                      assemble_now=False)
-
-    loops = tuple(loops)
-
-    def thunk():
-        for kernel in loops:
-            kernel()
-    return thunk
+    return functools.partial(assemble, expr, tensor, bcs, form_compiler_parameters, mat_type, sub_mat_type, assemble_now=False)
```
wence-, line 251: I guess the idea is that we could fully deprecate this?

firedrake/assemble.py
connorjward: Neither did I. I just tried it and it worked!

firedrake/assemble.py
```diff
@@ -450,54 +460,32 @@ def apply_bcs(tensor, bcs, *, assembly_rank=None, form_compiler_parameters=None,
...
-                                 form_compiler_parameters=form_compiler_parameters,
-                                 mat_type=mat_type,
-                                 sub_mat_type=sub_mat_type,
-                                 appctx=appctx,
-                                 assemble_now=assemble_now,
-                                 zero_tensor=False)
     elif assembly_rank == AssemblyRank.VECTOR:
-        for bc in dirichletbcs:
+        for bc in bcs:
             if assemble_now:
```
connorjward, line 478: Could someone please explain why we have this `assemble_now` parameter? Is the idea that boundary conditions are applied elsewhere in the solver so we can skip that here?

firedrake/assemble.py
connorjward: We could cache the callables, but I think the parloops are stashed on the Dats so we're only saving a little bit.

firedrake/assemble.py
wence-: It was introduced in d00719072 so I think it has some something to do with `EquationBC` objects, so I hope @ksagiyam can comment.

I don't really understand the logic. In fact, I think it's buggy. Consider:

```python
from firedrake import *
from firedrake.assemble import create_assembly_callable
V = FunctionSpace(mesh, "Q", 1)
bc = DirichletBC(V, 1, "on_boundary")
v = TestFunction(V)
print(assemble(v*dx, bc=bc)).dat.data_ro)
# => [1, 1, 1, 1]
f = Function(V)
callable = create_assembly_callable(v*dx, tensor=f, bcs=[bc])
callable()
print(f.dat.data_ro)
# => [0, 0, 0, 0]
```

ksagiyam: We used to have `collect_loops` parameter, and `assemble_now == not collect_loops`. `assemble` returns a fully assembled matrix/vector if `assemble_now == True`, while it returns a callable if `assemble_now == False`.
Here, for vectors, I think we use `assemble_now` also a flag indicating what to assemble: if `True`, we assemble the solution, and, if `False`, we assemble the `residual`, which is why we get `[0,0,0,0]` in the Lawrence's example.

wence-: > We used to have collect_loops parameter, and assemble_now == not collect_loops. assemble returns a fully assembled matrix/vector if assemble_now == True, while it returns a callable if assemble_now == False.

OK, this makes sense.

> Here, for vectors, I think we use assemble_now also a flag indicating what to assemble: if True, we assemble the solution, and, if False, we assemble the residual, which is why we get [0,0,0,0] in the Lawrence's example.

This is horrible and must be expunged.

ksagiyam: Yes, I agree.

connorjward: OK thanks. So what would be a nice way to do this?

wence-: I think we should move to a case where `assemble_now` is not required.

So I think the low-level stuff should always be in `assemble_now == False` case which means that we're collecting the work to be done.

I don't understand enough about the BC application with equation bcs in the "residual evaluation" case. This happens inside `solving_utils._SNESContext.form_function`.

ksagiyam: When `_assemble` finds an `EquationBC` (in `apply_bcs`), whether it is assembling a residual or not, it first zeros associated rows of the vector/matrix, and then call `_assemble` from inside to fill those rows with the specified boundary equations. If the `EquationBC` itself has `DirichetBC`s (Dirichlet BCs of the Equation BC), those `DirichletBCs` are treated correctly in the second call to the `_assemble` according to if we are assembling a residual or a solution. BCs of BCs of `EquationBC`s and things like these are also treated in the same way with the recursive call of `_assemble`.

Yes, I think it's a good idea to remove `assemble_now` from `_assemble` (and related functions) with `assemble_now == False`. If `assemble_now == False`, `_assemble` currently automatically assembles "residual"s, so if we want to assemble a solution, we will need to do `bc.apply` somewhere at higher level, in which case we will need to do `bc.apply` recursively to set values of `DirichletBC`s of `DirichletBC`s ... of `EquationBC`s correctly.

connorjward:
Parloops are now cached on the form. I've done some quick performance checks and (in the pathological case of a very simple kernel and tiny mesh) `assemble()` is now about 4 times faster! Unfortunately, it has slowed down the `create_assembly_callable()` alternative by a factor of around 2 because we now do more work at every call. I have some ideas about how this could be reduced though.

firedrake/assemble.py
```diff
@@ -161,14 +161,24 @@ def _make_vector(test):
...
+        cached_tensor, parloops = expr._cache[cache_key]
+        if cached_tensor is not tensor:
+            parloops = None
+            del expr._cache[cache_key]
+    else:
+        parloops = None
+
+    if not parloops:
+        parloops = _make_parloops(expr, tensor, bcs=bcs, diagonal=diagonal, **kwargs)
+        expr._cache[cache_key] = (tensor, parloops)
```
wence-, line 181: I guess this is a stop-gap, but it might be surprising that a form holds on to its output tensor. Would it work to hold a `weakref` here?

connorjward: I guess. The tensor's data is still tied to the parloop anyway so not sure it really matters.

wence-: Oh yeah, in which case, don't bother.

firedrake/assemble.py
```diff
@@ -91,21 +92,114 @@ def assemble(expr, tensor=None, bcs=None, form_compiler_parameters=None,
...
+    else:
+        test, = expr.arguments()
+    if vector:
+        if test.function_space() != vector.function_space():
+            raise ValueError("Form's argument does not match provided result tensor")
+        vector.dat.zero()
+    else:
+        vector = _make_vector(test)
+    _assemble_expr(expr, vector, diagonal=diagonal, assemble_now=assemble_now, **kwargs)
+    return vector
```
connorjward, line 134: Here, and with the scalar and matrix cases, I'm masking some potentially bad behaviour. Even with `create_assembly_callable()` the callable will always return the tensor (previously it didn't). Since the callable doesn't care about the return value it doesn't cause a problem but it's not ideal.

Personally I'd quite like it if we specified that `None` is returned from `assemble()` if a tensor is provided. That is a significant API change though. We could avoid that by adding a `return_tensor=True` default argument.

wence-: I think `assemble` should always return the tensor (even if provided). What is the potentially bad behaviour you're masking?

I am happy to futz with the create_assembly_callable API (I would like to just remove it after deprecation).

connorjward: By bad behaviour I mean that the public API has been changed and now the tensor is returned instead of `None`. It's very unlikely to be a problem and deprecating `create_assembly_callable()` will resolve this anyway.

wence-: Although not named privately, the docstring of `create_assembly_callable` has always indicated that it is not really part of the public API, so I'm kind of happy with this.

firedrake/assemble.py
connorjward: I've created an `_AssemblyOpts` namedtuple to hold all of the options.

firedrake/assemble.py
connorjward: I've replaced `assemble_now` with a variable `assembly_type` which is either `"solution"` or `"residual"`. I may have misunderstood this conversation though but hopefully what I've done is valid.

connorjward:
Pending tests and reviews I think this is ready to go.

firedrake/assemble.py
wence-: Can use an `IntEnum` for that, instead of a string?>

ksagiyam:
Looks good to me!

firedrake/assemble.py
```diff
@@ -277,12 +381,79 @@ def get_matrix(expr, mat_type, sub_mat_type, *, bcs=None,
...
+    # tensor is changed then we need to regenerate the parloop (until parloops
+    # stop depending on data).
+    # When we generate a new parloop the old one is removed from the cache to
+    # prevent leaking memory.
+    parloops = None
+    if isinstance(expr, ufl.form.Form):
+        cache_key = "parloops"
+        if cache_key in expr._cache:
+            cached_tensor, cached_parloops = expr._cache[cache_key]
+            if cached_tensor is tensor:
```
ksagiyam, line 407: Maybe we should check `tensor.bcs == bcs` here just in case.

connorjward: I don't think that would work because `tensor` doesn't have a `bcs` property. There is definitely a debate to be had here though about how safe this caching is. At the moment the cache will only miss if either the form or tensor are changed, not if `bcs`, `mat_type` etc are modified. I've added a warning about this to the documentation and it is not a common use case (in my opinion) for users to be modifying these options when assembling the same form.
Alongside this refactoring I'm also looking into making the parloop agnostic to as many of the `kwargs` as possible so hopefully this should become less of an issue.

wence-: >  I've added a warning about this to the documentation and it is not a common use case (in my opinion) for users to be modifying these options when assembling the same form.

This is, unfortunately, not a reasonable position.

For example, there are (or would be after these changes) situations in the Firedrake source tree right now where the same form is (legitimately) assembled with different `mat_type` arguments. e.g. `AssembledPC` takes an `assemble(a, mat_type="matfree")` and assembles it with `assemble(a, mat_type="aij")`.

Caching is only ever allowed to be a performance optimisation, it isn't allowed to change behaviour.

connorjward: Ok. Thanks for the advice. I'll do some more thinking then about how to make the caching safe.

connorjward: This should now be resolved. I've checked and the performance hit from this alteration is negligible.

firedrake/assemble.py
```diff
@@ -277,12 +381,68 @@ def get_matrix(expr, mat_type, sub_mat_type, *, bcs=None,
...
+    :arg opts: :class:`_AssemblyOpts` containing the assembly options.
+    :arg assembly_rank: The appropriate :class:`_AssemblyRank`.
+    """
+    parloop_init_args = (expr, tensor, bcs, opts.diagonal, opts.fc_params, assembly_rank)
+    cached_init_args, cached_parloops = expr._cache.get("parloops", (None, None))
+    parloops = cached_parloops if cached_init_args == parloop_init_args else None
+
+    if not parloops:
+        parloops = _make_parloops(*parloop_init_args)
+        expr._cache["parloops"] = (parloop_init_args, parloops)
```
ksagiyam, line 403: Here, `(parloop_init_args, parloops)` overrides `(cached_init_args, cached_parloops)`, but don't we want to keep old values?

connorjward: This was intentional. I wanted to make sure that we didn't run into any data reference/memory usage issues. In particular with respect to the output tensor. The main use case that I've been targeting this refactor at is for repeated assembly of the same form (e.g. in a time loop) which will work fine.

Once parloops stop holding references to data (something I'm working on) I think this could be altered.

ksagiyam: Ok! Thanks.

dham (approved):

ReubenHill:
@connorjward I would really appreciate you taking us through these changes and the rationale behind them just so we're all up to date. Maybe at the next firedrake meeting? (Whenever that happens!) Looks like you've done some really nice work here

connorjward:
> 
> 
> @connorjward I would really appreciate you taking us through these changes and the rationale behind them just so we're all up to date. Maybe at the next firedrake meeting? (Whenever that happens!) Looks like you've done some really nice work here

Sure, very happy to! I can summarise what I've done pretty straightforwardly though:
- We can now cache the parloops on the form (see `_assemble_expr()`). The current way I've implemented this is quite restrictive but this is likely to change with the other work that I'm doing.
- The logic is now a lot tidier (in my opinion).

firedrake/assemble.py
```diff
@@ -17,49 +17,61 @@
...
+        sparsity is constructed.  This can result in some memory
+        savings, but does not work with all PETSc preconditioners.
+        BAIJ matrices only make sense for non-mixed matrices.
+    :kwarg sub_mat_type: (optional) string indicating the matrix type to
+        use *inside* a nested block matrix.  Only makes sense if
+        ``mat_type`` is ``nest``.  May be one of ``"aij"`` or ``"baij"``.  If
+        not supplied, defaults to ``parameters["default_sub_matrix_type"]``.
+    :kwarg appctx: (optional) additional information to hang on the assembled
+        matrix if an implicit matrix is requested (mat_type ``"matfree"``).
+    :kwarg options_prefix: (optional) PETSc options prefix to apply to matrices.
```
ReubenHill, line 74: this docstring is missing a `:returns:`

firedrake/assemble.py
```diff
@@ -80,55 +92,62 @@ def assemble(expr, tensor=None, bcs=None, form_compiler_parameters=None,
...
-    :arg mat_type: PETSc matrix type for the assembled matrix.
-    :arg sub_mat_type: PETSc matrix type for blocks if mat_type is
-        nest.
-    :arg arguments: The test and trial functions.
-    :raises ValueError: On bad arguments.
-    :returns: 2-tuple of validated mat_type, sub_mat_type.
+    :arg expr: a :class:`~ufl.classes.Form` or a :class:`~slate.TensorBase`
+        expression.
+
+    See :func:`assemble` for a description of the possible additional arguments.
```
ReubenHill, line 124: Docstring missing a `:returns:`

firedrake/assemble.py
```diff
@@ -165,79 +185,165 @@ def create_assembly_callable(expr, tensor=None, bcs=None, form_compiler_paramete
...
+                             assembly_type="residual")
+
+
+def _get_assembly_rank(expr, diagonal):
+    """Return the assembly rank.
+
+    :arg expr: The expression being assembled.
+    :arg diagonal: If assembling a matrix is it diagonal?
+
+    :returns: The appropriate :class:`_AssemblyRank`.
```
ReubenHill, line 204: Might as well specify - `e.g. _AssemblyRank.VECTOR`

firedrake/assemble.py
```diff
@@ -165,79 +185,165 @@ def create_assembly_callable(expr, tensor=None, bcs=None, form_compiler_paramete
...
+                             sub_mat_type=sub_mat_type,
+                             diagonal=diagonal,
+                             assembly_type="residual")
+
+
+def _get_assembly_rank(expr, diagonal):
+    """Return the assembly rank.
+
+    :arg expr: The expression being assembled.
+    :arg diagonal: If assembling a matrix is it diagonal?
```
ReubenHill, line 202: Unclear - if this is a `bool` then please do say as much

firedrake/assemble.py
```diff
@@ -165,79 +185,165 @@ def create_assembly_callable(expr, tensor=None, bcs=None, form_compiler_paramete
...
+                             bcs=bcs,
+                             form_compiler_parameters=form_compiler_parameters,
+                             mat_type=mat_type,
+                             sub_mat_type=sub_mat_type,
+                             diagonal=diagonal,
+                             assembly_type="residual")
+
+
+def _get_assembly_rank(expr, diagonal):
+    """Return the assembly rank.
```
ReubenHill, line 199: Where "assembly rank" means what exactly?

firedrake/assemble.py
```diff
@@ -80,55 +92,62 @@ def assemble(expr, tensor=None, bcs=None, form_compiler_parameters=None,
...
+    else:
+        raise ValueError("assembly_type must be either 'solution' or 'residual'")
+    mat_type, sub_mat_type = _get_mat_type(mat_type, sub_mat_type,
+                                           expr.arguments())
+    opts = _AssemblyOpts(diagonal, assembly_type, form_compiler_parameters,
+                         mat_type, sub_mat_type, appctx, options_prefix)
+
+    assembly_rank = _get_assembly_rank(expr, diagonal)
+    if assembly_rank == _AssemblyRank.SCALAR:
+        return _assemble_scalar(expr, tensor, bcs, opts)
```
ReubenHill, line 144: Given that your documentation for `_assemble_scalar` implies that `tensor` must be `None` - won't this fail if `tensor` is not `None` in kwargs of `assemble`?

firedrake/assemble.py
```diff
@@ -17,49 +17,61 @@
...
+
+
+_AssemblyOpts = namedtuple("_AssemblyOpts", ["diagonal",
+                                             "assembly_type",
+                                             "fc_params",
+                                             "mat_type",
+                                             "sub_mat_type",
+                                             "appctx",
+                                             "options_prefix"])
+"""Container to hold immutable assembly options."""
```
ReubenHill, line 41: I might be missing it, but it looks like the meaning of these options aren't documented anywhere?

firedrake/assemble.py
```diff
@@ -165,79 +185,165 @@ def create_assembly_callable(expr, tensor=None, bcs=None, form_compiler_paramete
...
+        return _AssemblyRank.SCALAR
+    if rank == 1:
+        return _AssemblyRank.VECTOR
+    if rank == 2:
+        return _AssemblyRank.MATRIX
+    raise AssertionError
+
+
+def _assemble_scalar(expr, scalar, bcs, opts):
+    """Assemble a scalar.
```
ReubenHill, line 220: Would it be more accurate (and clear) to say `Assemble a scalar from a 0-form`. Would at least be consistent with the `ValueError`s below

firedrake/assemble.py
```diff
@@ -165,79 +185,165 @@ def create_assembly_callable(expr, tensor=None, bcs=None, form_compiler_paramete
...
+    :arg vector: The vector to write to (may be ``None``).
+    :arg bcs: Iterable of boundary conditions.
+    :arg opts: :class:`_AssemblyOpts` containing the assembly options.
+
+    :returns: The assembled :class:`firedrake.Function`.
+    """
+    if opts.diagonal:
+        test, trial = expr.arguments()
+        if test.function_space() != trial.function_space():
+            raise ValueError("Can only assemble diagonal of 2-form if functionspaces match")
```
ReubenHill, line 252: I might be being thick, but aren't we dealing with a 1-form if we are assembling a vector?

firedrake/assemble.py
```diff
@@ -165,79 +185,165 @@ def create_assembly_callable(expr, tensor=None, bcs=None, form_compiler_paramete
...
+
+def _assemble_vector(expr, vector, bcs, opts):
+    """Assemble a vector.
+
+    :arg expr: The expression being assembled.
+    :arg vector: The vector to write to (may be ``None``).
+    :arg bcs: Iterable of boundary conditions.
+    :arg opts: :class:`_AssemblyOpts` containing the assembly options.
+
+    :returns: The assembled :class:`firedrake.Function`.
```
ReubenHill, line 247: Some note about why it's a `firedrake.Function` (which is pretending to be a cofunction) would be great here.

firedrake/assemble.py
wence-: In this case we're building the diagonal of a matrix as a vector.

firedrake/assemble.py
ReubenHill: And/or in the `assemble` docstring

firedrake/assemble.py
ReubenHill: It's particularly important too since what you get from `assemble` is not exactly intuitive. Obviously will improve when we have UFL cofunctions and matrices.

firedrake/assemble.py
ReubenHill: Okay this needs to be made clear somewhere in the docstring

firedrake/assemble.py
```diff
@@ -165,79 +185,165 @@ def create_assembly_callable(expr, tensor=None, bcs=None, form_compiler_paramete
...
+
+def _assemble_matrix(expr, matrix, bcs, opts):
+    """Assemble a matrix.
+
+    :arg expr: The expression being assembled.
+    :arg matrix: The matrix to write to (may be ``None``).
+    :arg bcs: Iterable of boundary conditions.
+    :arg opts: :class:`_AssemblyOpts` containing the assembly options.
+
+    :returns: The assembled :class:`.Matrix` or :class:`.ImplicitMatrix`.
```
ReubenHill, line 278: As for `_assemble_vector`, some comment on what the matrix _is_ would be welcome both here and in the `assemble` docstring. I.e. what DOFs the row and column spaces represent.

firedrake/assemble.py
```diff
@@ -165,79 +185,165 @@ def create_assembly_callable(expr, tensor=None, bcs=None, form_compiler_paramete
...
+    if opts.mat_type == "matfree":
+        matrix.assemble()
+    else:
+        _assemble_expr(expr, matrix, bcs, opts, _AssemblyRank.MATRIX)
+        matrix.M.assemble()
+    return matrix
+
+
+def _make_scalar():
+    """Return a scalar.
```
ReubenHill, line 297: Should read `Make an empty scalar.`

firedrake/assemble.py
```diff
@@ -165,79 +185,165 @@ def create_assembly_callable(expr, tensor=None, bcs=None, form_compiler_paramete
...
+
+    :returns: An empty :class:`op2.Global`.
+    """
+    return op2.Global(1, [0.0], dtype=utils.ScalarType)
+
+
+def _make_vector(test):
+    """Return a vector.
+
+    :arg test: The :class:`.FunctionSpace` the function is defined for.
```
ReubenHill, line 307: Why is this called `test` and not, say, `V` or `fs`?

firedrake/assemble.py
```diff
@@ -165,79 +185,165 @@ def create_assembly_callable(expr, tensor=None, bcs=None, form_compiler_paramete
...
+
+    :returns: An empty :class:`.Function`.
+    """
+    return firedrake.Function(test.function_space())
+
+
+def _make_matrix(expr, bcs, opts):
     """Get a matrix for the assembly of expr.
```
ReubenHill, line 315: Should read `Make a...`

firedrake/assemble.py
```diff
@@ -277,12 +383,68 @@ def get_matrix(expr, mat_type, sub_mat_type, *, bcs=None,
...
+
+
+def _assemble_expr(expr, tensor, bcs, opts, assembly_rank):
+    """Assemble an expression.
+
+    :arg expr: The expression to be assembled.
+    :arg tensor: The tensor to write to.
+    :arg bcs: Iterable of boundary conditions.
+    :arg opts: :class:`_AssemblyOpts` containing the assembly options.
+    :arg assembly_rank: The appropriate :class:`_AssemblyRank`.
```
ReubenHill, line 397: Missing `:returns:`

firedrake/assemble.py
```diff
@@ -165,79 +185,165 @@ def create_assembly_callable(expr, tensor=None, bcs=None, form_compiler_paramete
...
+    if scalar:
+        raise ValueError("Can't assemble 0-form into existing tensor")
+
+    scalar = _make_scalar()
+    _assemble_expr(expr, scalar, bcs, opts, _AssemblyRank.SCALAR)
+    return scalar.data[0]
+
+
+def _assemble_vector(expr, vector, bcs, opts):
+    """Assemble a vector.
```
ReubenHill, line 240: Please say more - given that we now have `_assemble_expression` too it needs to be made clear what this and `_assemble_scalar` and `_assemble_matrix` are actually doing (even if it may seem obvious!).

firedrake/assemble.py
```diff
@@ -165,79 +185,165 @@ def create_assembly_callable(expr, tensor=None, bcs=None, form_compiler_paramete
...
+    # Might have gotten here without EquationBC objects preprocessed.
+    if any(isinstance(bc, EquationBC) for bc in bcs):
+        bcs = tuple(bc.extract_form("F") for bc in bcs)
+
+    _assemble_expr(expr, vector, bcs, opts, _AssemblyRank.VECTOR)
+    return vector
+
+
+def _assemble_matrix(expr, matrix, bcs, opts):
+    """Assemble a matrix.
```
ReubenHill, line 271: Please say more - given that we now have `_assemble_expression` too it needs to be made clear what this and `_assemble_scalar` and `_assemble_vector` are actually doing (even if it may seem obvious!). (dup)

firedrake/assemble.py
```diff
@@ -277,12 +383,68 @@ def get_matrix(expr, mat_type, sub_mat_type, *, bcs=None,
...
+        raise NotImplementedError("Diagonal assembly and EquationBC not supported")
+    for bc in eq_bcs:
+        if assembly_rank == _AssemblyRank.VECTOR:
+            bc.zero(tensor)
+        _assemble_expr(bc.f, tensor, bc.bcs, opts, assembly_rank)
 
-def collect_lgmaps(matrix, all_bcs, Vrow, Vcol, row, col):
+
+def _get_mat_type(mat_type, sub_mat_type, arguments):
+    """Provide defaults and validate matrix-type selection.
```
ReubenHill, line 423: Could you expand on this a little? Defaults for what? Validating what about matrix-type selection?

firedrake/assemble.py
```diff
@@ -317,8 +479,27 @@ def collect_lgmaps(matrix, all_bcs, Vrow, Vcol, row, col):
     return (rlgmap, clgmap), unroll
 
 
-def matrix_arg(access, get_map, row, col, *,
-               all_bcs=(), matrix=None, Vrow=None, Vcol=None):
+def _vector_arg(access, get_map, i, *, function, V):
+    """Obtain an op2.Arg for insertion into given Function.
```
ReubenHill, line 483: ```Obtain an op2.Arg for insertion into given vector (a :class:`.Function`)``` would seem clearer to me

firedrake/assemble.py
```diff
@@ -317,8 +479,27 @@ def collect_lgmaps(matrix, all_bcs, Vrow, Vcol, row, col):
...
+
+    :arg access: access descriptor.
```
ReubenHill, line 485: This is a PyOP2 type right? Make clear

firedrake/assemble.py
```diff
@@ -317,8 +479,27 @@ def collect_lgmaps(matrix, all_bcs, Vrow, Vcol, row, col):
...
+    :arg get_map: callable of one argument that obtains Maps from
```
ReubenHill, line 486: `that obtains :class:`PyOP2.Maps` ...`

firedrake/assemble.py
```diff
@@ -317,8 +479,27 @@ def collect_lgmaps(matrix, all_bcs, Vrow, Vcol, row, col):
...
+        functionspaces.
+    :arg i: index of block (may be None).
```
ReubenHill, line 488: What's a block?

firedrake/assemble.py
```diff
@@ -317,8 +479,27 @@ def collect_lgmaps(matrix, all_bcs, Vrow, Vcol, row, col):
...
+    :arg function: Function to insert into.
+    :arg V: functionspace corresponding to function.
```
ReubenHill, line 490: Might as well specify ``` :class:`.Function` ``` and ``` :class:`.FunctionSpace` ```

firedrake/assemble.py
```diff
@@ -317,8 +479,27 @@ def collect_lgmaps(matrix, all_bcs, Vrow, Vcol, row, col):
...
+
+    :returns: An op2.Arg."""
```
ReubenHill, line 492: Same here with the ol' sphinx linking

firedrake/assemble.py
```diff
@@ -336,109 +517,33 @@ def matrix_arg(access, get_map, row, col, *,
...
-
-    tensor = op2.Global(1, [0.0], dtype=utils.ScalarType)
-    return tensor, (), lambda: tensor.data[0]
-
-
-def apply_bcs(tensor, bcs, *, assembly_rank=None, form_compiler_parameters=None,
-              mat_type=None, sub_mat_type=None, appctx={}, diagonal=False,
-              assemble_now=True):
+def _apply_dirichlet_bcs(tensor, bcs, opts, assembly_rank):
     """Apply boundary conditions to a tensor.
```
ReubenHill, line 536: Is this not now strictly _dirichlet_ BCs?

firedrake/assemble.py
```diff
@@ -450,65 +555,86 @@ def apply_bcs(tensor, bcs, *, assembly_rank=None, form_compiler_parameters=None,
...
+                raise AssertionError
+    elif assembly_rank == _AssemblyRank.SCALAR:
+        pass
+    else:
+        raise AssertionError
+
+
+@utils.known_pyop2_safe
+def _make_parloops(expr, tensor, bcs, diagonal, fc_params, assembly_rank):
+    """Create some parloops.
```
ReubenHill, line 590: Parloops for what? Are they empty?

firedrake/assemble.py
```diff
@@ -450,65 +555,86 @@ def apply_bcs(tensor, bcs, *, assembly_rank=None, form_compiler_parameters=None,
...
+
+    :arg expr: The expression to be assembled.
```
ReubenHill, line 592: If it's necessary to have preprocessed the expressions at this point, you probably ought to say so.

firedrake/assemble.py
```diff
@@ -450,65 +555,86 @@ def apply_bcs(tensor, bcs, *, assembly_rank=None, form_compiler_parameters=None,
...
+    :arg tensor: The tensor to write to.
```
ReubenHill, line 593: Add - ```will be either a scalar, vector (masquerading as a :class:`.Function`) or :class:`.Matrix` depending on the other arguments given``` or similar. I presume it's actually decided by some combination of `assembly_rank` and `diagonal`?

firedrake/assemble.py
```diff
@@ -450,65 +555,86 @@ def apply_bcs(tensor, bcs, *, assembly_rank=None, form_compiler_parameters=None,
...
+    :arg bcs: Iterable of boundary conditions.
+    :arg diagonal: If assembling a matrix is it diagonal?
```
ReubenHill, line 595: CF previous comment about being specific that this is a bool and what it does

firedrake/assemble.py
```diff
@@ -450,65 +555,86 @@ def apply_bcs(tensor, bcs, *, assembly_rank=None, form_compiler_parameters=None,
...
+    :arg fc_params: The form compiler parameters.
```
ReubenHill, line 596: What type are they?

firedrake/slate/slate.py
```diff
@@ -115,6 +115,10 @@ class TensorBase(object, metaclass=ABCMeta):
 
     _id = count()
 
+    def __init__(self, *_):
+        # A cache to stash results in. Mirrors :class:`ufl.form.Form`.
```
ReubenHill, line 119: Should this not be a doc string?

ReubenHill (changes requested):
I've been annoying but, I hope, useful and gone through the docstrings to see if everything makes sense to me. See the various comments I've left for more.

firedrake/assemble.py
connorjward: I have now run all of the Thetis tests with this branch and they passed. I think the caching is sufficiently robust now.

firedrake/assemble.py
connorjward: I don't know the answer to this. @wence- @dham do you guys know?

firedrake/assemble.py
connorjward: We don't do any preprocessing of the expressions (I'm 95% sure) so not sure anything needs to go here along those lines.

firedrake/assemble.py
connorjward: I've moved the check for `tensor` being `None` to here so it should be clearer

firedrake/assemble.py
connorjward: There is quite a lot of detail in the bottom part of the docstring about the various things that get returned from this function so I've lazily just put `:returns: See below.`.

connorjward:
@wence- any chance you could review this at some point?

firedrake/assemble.py
dham: This is an internal function. There really isn't a need for pedagogical documentation here.

wence-: What Reuben is asking for could live in the `assemble` docstring, but not sure if it currently does.

ReubenHill: I agree with @wence- , this should be in the `assemble` documentation if it isn't already. 

Nevertheless one should not assume that just because something isn't public that a developer will automatically understand what a function does. The documentation here can avoid defining what a matrix is but it would really help if it pointed out where to find out what it is.

firedrake/assemble.py
connorjward: I have honestly got no idea.

firedrake/assemble.py
connorjward: I've added a sentence to the docstring for `assemble` and pointed to it from this function.

firedrake/assemble.py
wence-: Subspace of a mixed function.

connorjward: Amended.

firedrake/assemble.py
connorjward: This function doesn't return anything.

firedrake/assemble.py
```diff
@@ -165,79 +204,175 @@ def create_assembly_callable(expr, tensor=None, bcs=None, form_compiler_paramete
...
-    :arg tensor: An optional already allocated matrix for this expr.
-    :arg appctx: User application data.
-    :arg form_compiler_parameters: Any parameters for the form compiler.
-
-    :raises ValueError: In case of problems.
-    :returns: a 3-tuple ``(tensor, zeros, callable)``. zeros is a
-       (possibly empty) iterable of zero-argument functions to zero the returned
-       tensor, callable is a function of zero arguments that returns
-       the tensor.
+    return functools.partial(assemble, expr,
```
wence-, line 207: Can we add a deprecation warning that users will see here?

wence- (approved):

firedrake/assemble.py
wence-: Can you do that in a separate round please?

PR #1989, opened 2021-02-25
Fix caching of physical node locations for MG

wence-:
We can't cache on the topology!

@florianwechsung `git annotate` says that you're responsible for the opencascade code that I'm removing. I think that one doesn't need to copy the topological `shared_data_cache` because the topologies remain the same.


florianwechsung:
Okay took me a while to remember what's going on here, but I think this looks sensible.

florianwechsung (dismissed):

Issue #1860, opened 2020-10-07
+= is not annotated


jwallwork23:
In Firedrake, `target += source` is equivalent to `target.assign(target + source)`, provided `target` and `source` are from the same space. However, this is not annotated by pyadjoint.

Consider the MFE
```
from firedrake import *
from firedrake_adjoint import *

# Create source and target functions in P1 space
mesh = UnitSquareMesh(1, 1)
P1 = FunctionSpace(mesh, "CG", 1)
source = Function(P1)
source.assign(1.0)                # This gets annotated
target = Function(P1)
# target.assign(target + source)  # This gets annotated
target += source                  # This does not

# Examine tape
tape = get_working_tape()
for i, block in enumerate(tape.get_blocks()):
    print(i, type(block))

# Try to compute gradient of some functional w.r.t. source
J = assemble(target*dx)
compute_gradient(J, Control(source))
```
This gives the following outputs under each approach....

**target.assign(target + source)**
```
0 <class 'firedrake.adjoint.blocks.FunctionAssignBlock'>
1 <class 'firedrake.adjoint.blocks.FunctionAssignBlock'>
```

**target += source**
```
0 <class 'firedrake.adjoint.blocks.FunctionAssignBlock'>
WARNING:root:Adjoint value is None, is the functional independent of the control variable?
Traceback (most recent call last):
  File "/home/joe/software/firedrake-adapt/src/PyOP2/pyop2/utils.py", line 245, in verify_reshape
    a.shape = shape
ValueError: cannot reshape array of size 1 into shape (4,)

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "pm.py", line 20, in <module>
    compute_gradient(J, Control(source))
  File "/home/joe/software/firedrake-adapt/src/pyadjoint/pyadjoint/drivers.py", line 31, in compute_gradient
    grads = [i.get_derivative(options=options) for i in m]
  File "/home/joe/software/firedrake-adapt/src/pyadjoint/pyadjoint/drivers.py", line 31, in <listcomp>
    grads = [i.get_derivative(options=options) for i in m]
  File "/home/joe/software/firedrake-adapt/src/pyadjoint/pyadjoint/control.py", line 51, in get_derivative
    return self.control._ad_convert_type(0., options=options)
  File "/home/joe/software/firedrake-adapt/src/pyadjoint/pyadjoint/tape.py", line 46, in wrapper
    return function(*args, **kwargs)
[...]

Issue #1980, opened 2021-02-23
Assign block incorrectly annotated for nonlinear expressions


ReubenHill:
The default block for annotating `assign` operations in pyadjoint assume the expression being assigned is linear. In firedrake is this not always that case and you can get the wrong hessian when you assign a nonlinear expression.

To resolve this, I propose that a new `Block` be added to `firedrake/adjoint/blocks.py` which will do the right thing for us. (Not quite sure exactly what needs to inherit from what to make this work mind you)

ReubenHill:
The new hessian will be similar to that for `InterpolateBlock` in #1942

ReubenHill:
You can update `FunctionAssignBlock` to do this methinks

PR #1995, opened 2021-03-01
Annotate +=, *=, etc.

jwallwork23:
This PR ensures +=, *=, -= and /= are correctly annotated for `Function`s. The failure reported in #1860 is fixed by these changes. For example, the following test passes.

```
from firedrake import *


# Create source and target functions in P1 space
mesh = UnitSquareMesh(1, 1)
P1 = FunctionSpace(mesh, "CG", 1)
source = Function(P1)
src = 1.0
source.assign(src)
control_src = Control(source)
target = Function(P1)
tgt = -2.0
target.assign(tgt)
control_tgt = Control(target)
target += source

# Try to compute gradient of some functional w.r.t. source
out = Function(P1)
out.assign(target)
J = assemble(out*dx)
assert np.isclose(J, src + tgt), J

tgt = -2.0
rf_src = ReducedFunctional(J, control_src)
for src in np.linspace(1, 3, 3):
    assert np.isclose(rf_src(src), src + tgt)

rf_tgt = ReducedFunctional(J, control_tgt)
src = 1.0
for tgt in np.linspace(-2, 1, 3):
    assert np.isclose(rf_tgt(tgt), src + tgt)
```


ReubenHill:
FYI issue #1980 might be relevant here

nbouziani:
Naive question: Are there circumstances where operations such as "a+=b" would not be equivalent to "a.assign(a+b)" (i.e. where using FunctionAssignBlock would not be appropriate) ?  

Apart from that, the PR looks all good for me. Would it be possible to add tests ?

jwallwork23:
@nbouziani So I added a test and in it identified one case where assign might not DTRT. For a function in P1 space, `f *= g` does a point-wise multiply, which isn't the same as the expression `f*g` in general. I guess the same applies for `f /= g`, but you don't see that in the test because the two functions coincide.

wence-:
> `f *= g` does a point-wise multiply, which isn't the same as the expression `f*g` in general

This is a "surprise" we should fix. Can you show an example?

jwallwork23:
```
>>> from firedrake import *
>>> mesh = UnitSquareMesh(4, 4)
>>> V = FunctionSpace(mesh, "CG", 1)
>>> x, y = SpatialCoordinate(mesh)
>>> f = interpolate(x + 1, V)
>>> g = interpolate(x + 1, V)
>>> g *= f
>>> assemble(g*dx)
2.3437499999999973
>>> assemble(f*f*dx)
2.3333333333333304
```

jwallwork23:
I guess this happens because g is in P1 space, but f*f is quadratic?

jwallwork23:
Okay I now realise this is a problem with my test, not with the implementation (see latest commit). Sorry!

wence- (approved):
Thanks (note that the adjoint of `*=` and `/=` is wrong, because the adjoint doesn't do the right thing for nonlinear assignment)

PR #1998, opened 2021-03-04
patch: Use correct "fake" subset

wence-:
If the local subdomain contains no elements, then producing a subset
with one element will fail. But the codegeneration just needs to know
that it is a subset, so use an empty one.


pefarrell (approved):
Happy with this if the tests pass.

PR #1999, opened 2021-03-04
remove real_tensorproduct from argument

ksagiyam:
`real_tensorproduct` parameter can be computed inside `FunctionSpace` constructor.


wence- (approved):
Thanks

PR #2001, opened 2021-03-09
use randomgen via numpy

ksagiyam:
Use `randomgen` via `numpy.random`.

Documentation of `randomfunctiongen` seems to have been broken (`https://www.firedrakeproject.org/firedrake.html#module-firedrake.randomfunctiongen`), though `randomfunctiongen.PCG64.__doc__`, for instance, contains correct doc string.

### DIFF
--- a/firedrake/randomfunctiongen.py
+++ b/firedrake/randomfunctiongen.py
@@ -1,25 +1,89 @@
 """
 
-This module wraps `randomgen <https://pypi.org/project/randomgen/>`__
-and enables users to generate a randomised :class:`.Function`
-from a :class:`.FunctionSpace`.
-This module inherits all attributes from `randomgen <https://pypi.org/project/randomgen/>`__.
+This module wraps `numpy.random <https://numpy.org/doc/stable/reference/random/index.html>`__,
+and enables users to generate a randomised :class:`.Function` from a :class:`.FunctionSpace`.
+This module inherits almost all attributes from `numpy.random <https://numpy.org/doc/stable/reference/random/index.html>`__ with the following changes:
+
+**Generator**
+
+:class:`.Generator` wraps `numpy.random.Generator <https://numpy.org/doc/stable/reference/random/generator.html>`__.
+:class:`.Generator` inherits almost all distribution methods from `numpy.random.Generator <https://numpy.org/doc/stable/reference/random/generator.html>`__,
+and they can be used to generate a randomised :class:`.Function` by passing a :class:`.FunctionSpace` as the first argument.
+
+Example:
+
+.. code-block:: python3
+
+    from firedrake import *
+
+    mesh = UnitSquareMesh(2, 2)
+    V = FunctionSpace(mesh, 'CG', 1)
+    pcg = PCG64(seed=123456789)
+    rg = Generator(pcg)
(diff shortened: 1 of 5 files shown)


wence-:
I think the docstring stuff is because sphinx-autodoc changes how it looks for stuff on modules and now calls `dir(module)`. I don't understand all the docstring reformatting that is going on, but defining module level `def __dir__(): return list(__all__)` might work.

firedrake/randomfunctiongen.py
```diff
@@ -235,103 +215,169 @@ def func(self, *args, **kwargs):
...
-                    inc = inc or rank
-                    super(_Wrapper, self).seed(seed=seed, inc=inc)
-
-        elif module_attr in ('Philox', 'ThreeFry'):
-
-            def seed(self, seed=None, counter=None, key=None):
-
+            def _init(self, *args, **kwargs):
+                # numpy.random does not allow users to set `seed` and `inc` separately,
+                # while randomgen (v.1.19) does. We follow the randomgen interface here.
```
wence-, line 246: Given that randomgen is deprecated and numpy is the stable API, can we not do this?

wence-:
Maybe the better thing for docs is to write a module level docstring that just explains what is going on.

firedrake/randomfunctiongen.py
ksagiyam: So `randomgen` stopped maintaining `randomgen.Generator`, and we are now recommended to use `numpy.random.Generator`, but it turned out that `numpy.random` is not a straightforward wrapper of `randomgen`. As far as I understand, one cool thing about `PCG64` is that one can create parallel-safe streams by setting `inc` parameter to different values while keeping the `seed` the same, and this is the recommended way to create multiple streams in parallel, and, indeed, `randomgen.PCG64` (latest version) still takes both `seed` and `inc` parameters. `numpy.random.PCG64`, however, somehow only takes one argument (`seed`), and generates two parameters from that `seed`, and pass those to `randomgen.PCG64` constructor under the hood. I don't know if this would create parallel-safe streams when a user make the value of the `seed` `rank` dependent as I don't know much about bit generators, so I decided to follow `randomgen.PCG64` interface, so that `Firedrake.PCG64` still takes two arguments (`seed` and `inc`). Our interface is also backward compatible this way.

ksagiyam:
Yes, I think I should add a module level doc string. `def __dir__()` trick didn't work either. I'll update. Thanks!

firedrake/randomfunctiongen.py
wence-: Per https://numpy.org/doc/stable/reference/random/bit_generators/pcg64.html
> The preferred way to use a BitGenerator in parallel applications is to use the SeedSequence.spawn method to obtain entropy values, and to use these to generate new BitGenerators:

```python
from numpy.random import Generator, PCG64, SeedSequence
sg = SeedSequence(1234)
rg = [Generator(PCG64(s)) for s in sg.spawn(10)]
```

ksagiyam: Ok, I now made this default, but explained how to explicitly use `inc` in the module docstring. Thanks!

wence- (approved):
Thanks, I think this looks good.

PR #2008, opened 2021-03-12
Fix all sphinx build warnings

wence-:
Squash all warnings in the sphinx build and turn new warnings into errors so we don't regress.


wence-:
This passed, but Jenkins didn't actually run the build tests. I don't understand why not. So I can't guarantee that all the build warnings are squashed on jenkins

wence-:
@dham ready for review. Migrated doc-build checking to github actions on the vanilla docker container.

PR #2013, opened 2021-03-19
Allow variable-height mesh extrusion

angus-g:
The "layer_height" parameter for mesh extrusion was a single scalar that gave a uniform thickness for all extruded layers. This is extended to allow for an array to be used instead, which gives the thickness for the individual layers.

I thought this might be useful after I just hacked it together for my own use. Probably needs:
- [x] tests
- [x] ensure that `layer_heights == layers` (I assume that's available on the `extruded_topology` somewhere?)

Definitely happy to tidy things up if this seems like it would be desired!

### DIFF
--- a/firedrake/extrusion_utils.py
+++ b/firedrake/extrusion_utils.py
@@ -21,7 +21,9 @@ def make_extruded_coords(extruded_topology, base_coords, ext_coords,
          coordinates from.
     :arg ext_coords: a :class:`~.Function` to write the extruded
          coordinates into.
-    :arg layer_height: an equi-spaced height for each layer.
+    :arg layer_height: the height for each layer.  Either a scalar,
+         where layers will be equi-spaced at the specified height, or a
+         1D array of variable layer heights to use through the extrusion.
     :arg extrusion_type: the type of extrusion to use.  Predefined
          options are either "uniform" (creating equi-spaced layers by
          extruding in the (n+1)dth direction), "radial" (creating
@@ -46,7 +48,18 @@ def make_extruded_coords(extruded_topology, base_coords, ext_coords,
                                and vert_space.family() in ['Lagrange',
                                                            'Discontinuous Lagrange']):
         raise RuntimeError('Extrusion of coordinates is only possible for a P1 or P1dg interval unless a custom kernel is provided')
-    layer_height = op2.Global(1, layer_height, dtype=RealType)
+
+    layer_height = numpy.atleast_1d(numpy.array(layer_height, dtype=RealType))
+
+    if layer_height.ndim > 1:
+        raise RuntimeError('Extrusion layer height should be 1d or scalar')
+
+    if layer_height.size > 1:
+        layer_height = numpy.cumsum(numpy.concatenate(([0], layer_height)))
+
+    layer_heights = layer_height.size
+    layer_height = op2.Global(layer_heights, layer_height, dtype=RealType)
+
     if kernel is not None:
         op2.ParLoop(kernel,
                     ext_coords.cell_set,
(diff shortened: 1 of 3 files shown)


firedrake/extrusion_utils.py
```diff
@@ -86,9 +97,10 @@ def _get_lp_domains(_inames, _extents):
         domains.extend(_get_lp_domains(('c', 'l'), (base_coord_dim, 2)))
         instructions = """
         ext_coords[{dd}, l, c] = base_coords[{dd}, c]
-        ext_coords[{dd}, l, {base_coord_dim}] = layer_height[0] * (layer + l)
+        ext_coords[{dd}, l, {base_coord_dim}] = layer_height[{hv}] * (layer + l)
```
ksagiyam, line 100: I think the correct formula here now is $(\sum_{i=0}^{layer}layer_height[i]) + layer_height[layer] * l $.

firedrake/extrusion_utils.py
```diff
@@ -104,9 +116,10 @@ def _get_lp_domains(_inames, _extents):
             tt[{dd}] = tt[{dd}] + bc[{dd}] * bc[{dd}]
         end
         tt[{dd}] = sqrt(tt[{dd}])
-        ext_coords[{dd}, l, c] = base_coords[{dd}, c] + base_coords[{dd}, c] * layer_height[0] * (layer+l) / tt[{dd}]
+        ext_coords[{dd}, l, c] = base_coords[{dd}, c] + base_coords[{dd}, c] * layer_height[{hv}] * (layer+l) / tt[{dd}]
```
ksagiyam, line 119: Here as well.

ksagiyam (changes requested):
I think a test that computes the volume of an extruded mesh would be good.

firedrake/extrusion_utils.py
angus-g: Oops, you are completely correct!

ksagiyam (changes requested):
Examples look good to me. In terms of checking size of `layer_height`, I think you can do, in the constructor of `mesh.ExtrudedMesh`, something like:
```
if layers.shape:
    ...
    # below line 1808
    num_layers = layers.sum(axis=1).max() if mesh.cell_set.total_size else 0
    num_layers = mesh.comm.allreduce(num_layers, op=MPI.MAX)
    ...
else:
    ...
    # below line 1816
    num_layers = layers
    ...
try:
    assert num_layers == len(layer_height)
except TypeError:
    pass
```

Could you also update docstrings of `mesh.ExtrudedMesh` and `make_extruded_coords` to reflect your changes?

Otherwise, I think it looks good. Thanks!

tests/extrusion/test_layer_height_mesh_volume.py
```diff
@@ -0,0 +1,23 @@
...
+    radius = 1000
+
+    mesh = IcosahedralSphereMesh(radius=radius, refinement=4)
+
+    # layer heights sum to 1.5 / radius
+    layer_heights = np.array([0.2] * 5 + [0.1] * 5) / radius
+    extmesh = ExtrudedMesh(mesh, 10, layer_height=layer_heights, extrusion_type="radial")
+
+    exact = 4 * pi * ((radius + 1.5/radius)**3 - radius**3) / 3
+    assert numpy.allclose(assemble(1*dx, domain=extmesh), exact, rtol=4e-3)
```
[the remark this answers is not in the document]
angus-g: Ah, nice! Thanks for catching my typos too :flushed:

ksagiyam (approved):

sv2518 (changes requested):
Hi. I think we should include a test for this in `tests/slate/*_extr.py`, I am not a 100 percent sure that this will work seamlessly just out of the box. I might need to adapt how the layers are accessed in the Slate kernel builder.

ksagiyam:
If we use `Slate` or not, we get local coordinates from `mesh.coordinates`. So I think it is hard for `Slate` to depend on the uniform-height assumption (and `mesh` does not even explicitly remember `layer_height`). Am I missing something?

sv2518:
No, I think you are right actually. Slate only depends on the number of layers and the layers themselves, not the layer height. I still want to double check with some tests, that I am not missing something. I am waiting for a fresh firedrake install to go through and will check locally by adding some Slate tests. If it looks fine, I will take my request of changes back.

sv2518 (approved):

sv2518:
Okay, I did not run in any issues with Slate tests that do have variable height extruded meshes. I have looked over the code as well and it looks good to go to me. Thanks @angus-g!

angus-g:
Thanks @sv2518! Better to be safe on these things, especially as Slate is definitely nowhere on my radar.

PR #2014, opened 2021-03-19
Website: add Thetis river plume animation on front page

tkarna:
Replaces the double slit wave animation on the Firedrake front page by a Thetis 3D river plume simulation. Perhaps better suited to highlight the capabilities of Firedrake. The animation also shows the movement of the 3D mesh.

fd_frontpage


wence- (approved):
Yes, great!

PR #2015, opened 2021-03-20
setup.py: add firedrake.adjoint to package list

acolinisi:
Fixes import error:

	>>> import firedrake
	  File "<stdin>", line 1, in <module>
	  File "/.../usr/lib/python3.8/site-packages/firedrake/__init__.py", line 59, in <module>
	    from firedrake.assemble import *
	...
	  File "/.../usr/lib/python3.8/site-packages/firedrake/interpolation.py", line 15, in <module>
	    from firedrake.adjoint import annotate_interpolate
	ModuleNotFoundError: No module named 'firedrake.adjoint'

Signed-off-by: Alexei Colin <acolin@isi.edu>

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -1,4 +1,5 @@
 from distutils.core import setup
+from setuptools import find_packages
 from glob import glob
 from os import environ as env, path
 from Cython.Distutils import build_ext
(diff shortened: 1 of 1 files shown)


setup.py
```diff
@@ -82,11 +82,17 @@ def get_petsc_dir():
...
                 "firedrake.cython",
+                "firedrake.mg",
+                "firedrake.slope_limiter",
+                "firedrake.matrix_free",
+                "firedrake.preconditioners",
                 "firedrake.slate", "firedrake.slate.slac", "firedrake.slate.static_condensation",
-                "firedrake_configuration", "firedrake_citations"],
+                "firedrake_configuration",
+                "firedrake_citations"
+                ],
```
wence-, line 95: Thanks. Perhaps we should instead use `setuptools.find_packages()` instead, to make this future proof. WDYT?

acolinisi: Done, retested, and force-pushed.

wence-: Great, thanks!

wence- (approved):

wence-:
Thanks very much.

PR #2016, opened 2021-03-23
Docs - debug install failures

sv2518:
Hi. I have extended the diagram on how to debug install failures with an additional branch. The branch considers the case where Firedrake install fails with an SSL certificate error. The error says `urllib.error.URLError: <urlopen error [SSL: CERTIFICATE_VERIFY_FAILED] certificate verify failed ...>`.

I keep running into this and I *think* it is because I code in VSCode and VSCode modifies my PATH variable (which is probably my fault). I get the install working by removing `/Library/Frameworks/Python.framework/Versions/3.8/bin` from my PATH variable so that `which python3` points at `/usr/local/bin/python3`.

Because my brain is a sieve, I always have to look for the closed issue where we talked about this. I thought adding this into the diagram might be the better thing to do as it will increase the visibility of the solution. LMKWYT, maybe this error is not common enough for being put into that diagram?

### DIFF
--- a/docs/source/install-debug.dot
+++ b/docs/source/install-debug.dot
@@ -11,6 +11,8 @@ digraph triage {
     python_version [label="Python 3.5?"];
     using_macos [label="Using\nMacOS?"];
     using_homebrew [label="Using\nHomebrew?"];
+    url_error [label="URL Error with SSL\ncertificate failure?"];
+    which_python [label="which python3 points\nat /usr/local/bin/python3?"];
 
     activate_venv [label="Activate the\nvenv first."];
     uninstall_anaconda [label="Deactivate\nAnaconda."];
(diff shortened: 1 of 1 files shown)


dham:
I think that rather than modifying PATH (which might affect other things the user is doing), the right thing to do in this case is just to use the following install line:
```
/usr/local/bin/python3 firedrake-install
```
Also, I think someone said that home-brew is moving out of /usr/local/

wence-:
Per https://docs.brew.sh/Installation, on ARM macs, homebrew puts stuff in `/opt/homebrew` (so that you can have intel versions in `/usr/local` for things that haven't been ported).

sv2518:
Okay, the latest commit should take both of the suggestions into account.

danshapero (approved):

wence-:
Thanks Sophia!

PR #2021, opened 2021-03-31
Refactor install

JDBetteridge:


scripts/firedrake-install
```diff
@@ -500,36 +628,22 @@ else:
...
-        log.debug("Running command '%s'", " ".join(args))
-        return subprocess.check_output(args, stderr=subprocess.STDOUT, env=os.environ).decode()
-    except subprocess.CalledProcessError as e:
-        log.debug(e.output.decode())
-        raise
+# Variant of check_call without logging
+check_output = partial(check_call, logcall=False)
 
 
 pyinstall = [python, "setup.py", "install"]
```
connorjward, line 649: This can be deleted. It isn't used anywhere.

scripts/firedrake-install
```diff
@@ -1585,35 +1587,33 @@ if mode == "install":
         f90 = os.path.join(compilerbin, "mpif90")
         compiler_env = create_compiler_env(cc, cxx, f90)
 
+    pip = Pip(python, "-m", "pip", parallel_env=compiler_env, blas_env=blas)  # !!!
```
connorjward, line 1590: I think that it's a little odd to have multiple `Pip` objects hanging around. Couldn't we rely on static/class methods?

dham:
So old that it would have to be done again.

PR #2024, opened 2021-04-08
Implement the Fast Diagonalization Method for Q_k elmenets (FDMPC)

pbrubeck:
This is a work in progress.

`FDMPC` introduces a change of basis into the local basis functions that diagonalize the H^1 Riesz map on the reference cell, and assembles a global sparse matrix on which other preconditioners, such as `ASMStarPC`, can be applied. The sparse matrix is obtained by either approximating the PDE coefficients as piecewise constants ("fdm_type": "affine") or extracting the relevant nonzero structure from the transformed matrix ("fdm_type": "stencil").

### DIFF
--- a/firedrake/preconditioners/asm.py
+++ b/firedrake/preconditioners/asm.py
@@ -122,7 +122,7 @@ class ASMStarPC(ASMPatchPC):
 
     def get_patches(self, V):
         mesh = V._mesh
-        mesh_dm = mesh._topology_dm
+        mesh_dm = mesh.topology_dm
 
         # Obtain the topological entities to use to construct the stars
         depth = PETSc.Options().getInt(self.prefix+"construct_dim", default=0)
@@ -175,7 +175,7 @@ class ASMVankaPC(ASMPatchPC):
 
     def get_patches(self, V):
         mesh = V._mesh
-        mesh_dm = mesh._topology_dm
+        mesh_dm = mesh.topology_dm
 
         # Obtain the topological entities to use to construct the stars
         depth = PETSc.Options().getInt(self.prefix + "construct_dim", default=-1)
@@ -250,7 +250,7 @@ class ASMLinesmoothPC(ASMPatchPC):
     def get_patches(self, V):
         mesh = V._mesh
         assert mesh.cell_set._extruded
-        dm = mesh._topology_dm
+        dm = mesh.topology_dm
         section = V.dm.getDefaultSection()
         # Obtain the codimensions to loop over from options, if present
         codim_list = PETSc.Options().getString(self.prefix+"codims", "0, 1")
--- a/firedrake/preconditioners/fdm.py
+++ b/firedrake/preconditioners/fdm.py
@@ -1,0 +1,840 @@
+from functools import lru_cache
+import numpy as np
+
+from pyop2 import op2
+
+from ufl import FiniteElement, TensorElement, Jacobian, JacobianInverse, dx, inner, dot, inv
+from firedrake.petsc import PETSc
+from firedrake.preconditioners.patch import bcdofs
+from firedrake.preconditioners.base import PCBase
+from firedrake.utils import IntType_c
+from firedrake.dmhooks import get_function_space, get_appctx
+import firedrake
+
+
+class FDMPC(PCBase):
+
+    _prefix = "fdm_"
+
+    def initialize(self, pc):
+        A, P = pc.getOperators()
+
+        # Read options
+        prefix = pc.getOptionsPrefix()
+        options_prefix = prefix + self._prefix
+        opts = PETSc.Options(options_prefix)
+        fdm_type = opts.getString("type", default="affine")
+
+        dm = pc.getDM()
+        V = get_function_space(dm)
+        self.mesh = V.mesh()
+        self.uf = firedrake.Function(V)
+        self.uc = firedrake.Function(V)
+
+        ndim = self.mesh.topological_dimension()
+        nscal = V.value_size
+        N = V.ufl_element().degree()
+        try:
+            N, = set(N)
+        except TypeError:
+            pass
+        Nq = 2 * N + 1
+
+        # Get problem solution and bcs
+        solverctx = get_appctx(dm)
+        self.u = solverctx._problem.u
+        self.bcs = solverctx.bcs_F
+
+        if len(self.bcs) > 0:
+            self.bc_nodes = np.unique(np.concatenate([bcdofs(bc, ghost=False)
+                                                      for bc in self.bcs]))
+            self.bc_ghost = np.unique(np.concatenate([bcdofs(bc, ghost=True)
+                                                      for bc in self.bcs]))
+        else:
+            self.bc_nodes = np.empty(0, dtype=PETSc.IntType)
+            self.bc_ghost = np.empty(0, dtype=PETSc.IntType)
+
+        bcflags, cell2cell = self.get_bc_flags(self.mesh, self.bcs)
+
+        # Encode bcflags into a VectorFunction that can be used in the pyOp2 kernels
+        # FIXME very ugly interface
+        VDG = firedrake.VectorFunctionSpace(self.mesh, 'DG', 0)
+        self.fbc = firedrake.Function(VDG, name="bcflags")
+        self.fbc.dat.data[cell2cell] = np.reshape(bcflags @ np.kron(np.eye(ndim), [[1], [3]]), (-1, 1, ndim))
+
+        self.weight = self.multiplicity(V)
+        with self.weight.dat.vec as w:
+            w.reciprocal()
+
+        # Get problem coefficients
+        appctx = self.get_appctx(pc)
+        mu = appctx.get("mu", None)  # sets the viscosity
+        helm = appctx.get("helm", None)  # sets the potential
+
+        hflag = helm is not None
+        Sfdm, self.restrict_kernel, self.prolong_kernel, self.stencil_kernel = self.assemble_matfree(ndim, nscal, N, Nq, hflag)
+
+        self.stencil = None
+        if fdm_type == "stencil":
+            # Compute high-order PDE coefficients and only extract
+            # nonzeros from the diagonal and interface neighbors
+            # Vertex-vertex couplings are ignored here,
+            # so this should work as direct solver only on star patches
+            W = firedrake.VectorFunctionSpace(self.mesh, "DQ", N, dim=2*ndim+1)
+            self.stencil = firedrake.Function(W)
+            Gq, Bq = self.assemble_coef(mu, helm, Nq)
+            Pmat = self.assemble_stencil(P, V, Gq, Bq, N, bcflags)
+        elif fdm_type == "affine":
+            # Compute low-order PDE coefficients, such that the FDM
+            # sparsifies the assembled matrix
+            Gq, Bq = self.assemble_coef(mu, helm, Nq)
+            Pmat = self.assemble_affine(P, V, Gq, Bq, Sfdm, bcflags, self.bc_ghost)
+        else:
+            raise ValueError("Unknown fdm_type")
+
+        # Monkey see, monkey do
+        opc = pc
+
+        # Internally, we just set up a PC object that the user can configure
+        # however from the PETSc command line.  Since PC allows the user to specify
+        # a KSP, we can do iterative by -fdm_pc_type ksp.
+        pc = PETSc.PC().create(comm=opc.comm)
+        pc.incrementTabLevel(1, parent=opc)
+
+        dm = opc.getDM()
+        pc.setDM(dm)
+        pc.setOptionsPrefix(options_prefix)
+        pc.setOperators(Pmat, Pmat)
+        self.pc = pc
+        pc.setFromOptions()
+
+    def update(self, pc):
+        pass
+
+    def applyTranspose(self, pc, x, y):
+        pass
+
+    def apply(self, pc, x, y):
+        self.uc.assign(firedrake.zero())
+
+        with self.uf.dat.vec_wo as xf:
+            x.copy(xf)
+
+        op2.par_loop(self.restrict_kernel, self.mesh.cell_set,
+                     self.uc.dat(op2.INC, self.uc.cell_node_map()),
+                     self.uf.dat(op2.READ, self.uf.cell_node_map()),
+                     self.fbc.dat(op2.READ, self.fbc.cell_node_map()),
+                     self.weight.dat(op2.READ, self.weight.cell_node_map()))
+
+        with self.uc.dat.vec as x_, self.uf.dat.vec as y_:
+            self.pc.apply(x_, y_)
+
+        op2.par_loop(self.prolong_kernel, self.mesh.cell_set,
+                     self.uc.dat(op2.WRITE, self.uc.cell_node_map()),
+                     self.uf.dat(op2.READ, self.uf.cell_node_map()),
+                     self.fbc.dat(op2.READ, self.fbc.cell_node_map()))
+
+        with self.uc.dat.vec_ro as xc:
+            xc.copy(y)
+
+        y.array_w[self.bc_nodes] = x.array_r[self.bc_nodes]
+
+    def view(self, pc, viewer=None):
+        super(FDMPC, self).view(pc, viewer)
+        if hasattr(self, "pc"):
+            viewer.printfASCII("PC to apply inverse\n")
+            self.pc.view(viewer)
+
+    def assemble_stencil(self, A, V, Gq, Bq, N, bcflags):
+        imode = PETSc.InsertMode.ADD_VALUES
+        lgmap = V.local_to_global_map([])
+
+        lexico_cg, nel = self.glonum_fun(V)
+        lexico_dg, _ = self.glonum_fun(self.stencil)
+
+        nloc = V.dof_dset.set.size
+        ndim = V.mesh().topological_dimension()
+        ncell = V.cell_node_list.shape[1]
+        nx = N + 1
+
+        def index_bcs(x, bc, val):
+            xshape = x.shape
+            x.shape = (nx, nx, 1) if ndim == 2 else (nx,)*ndim
+            if bc[0] == 1:
+                x[0, :, :] = val
+            if bc[1] == 1:
+                x[-1, :, :] = val
+            if bc[2] == 1:
+                x[:, 0, :] = val
+            if bc[3] == 1:
+                x[:, -1, :] = val
+            if ndim > 2:
+                if bc[4] == 1:
+                    x[:, :, 0] = val
+                if bc[5] == 1:
+                    x[:, :, -1] = val
+            x.shape = xshape
+            return
+
+        self.stencil.assign(firedrake.zero())
+        # FIXME I don't know how to use optional arguments here
+        if Bq is not None:
+            op2.par_loop(self.stencil_kernel, self.mesh.cell_set,
+                         self.stencil.dat(op2.WRITE, self.stencil.cell_node_map()),
+                         Gq.dat(op2.READ, Gq.cell_node_map()),
+                         Bq.dat(op2.READ, Bq.cell_node_map()),
+                         self.fbc.dat(op2.READ, self.fbc.cell_node_map()))
+        else:
+            op2.par_loop(self.stencil_kernel, self.mesh.cell_set,
+                         self.stencil.dat(op2.WRITE, self.stencil.cell_node_map()),
+                         Gq.dat(op2.READ, Gq.cell_node_map()),
+                         self.fbc.dat(op2.READ, self.fbc.cell_node_map()))
+
+        i = np.arange(ncell, dtype=PETSc.IntType)
+        sx = i - (i % nx)
+        sy = i - ((i // nx) % nx) * nx
+        if ndim == 2:
+            graph = np.array([sy, sy+(nx-1)*nx, sx, sx+(nx-1)])
+        else:
+            sz = i - (((i // nx) // nx) % nx) * nx * nx
+            graph = np.array([sz, sz+(nx-1)*nx*nx, sy, sy+(nx-1)*nx, sx, sx+(nx-1)])
+
+        offdiag = graph[graph != i]
+        aones = np.ones(offdiag.shape, dtype=PETSc.RealType)
+
+        # Upper bound for nonzeros on each row
+        nonzeros = PETSc.Vec().createMPI(A.getSizes()[0], comm=A.comm)
+        nonzeros.setOption(PETSc.Vec.Option.IGNORE_NEGATIVE_INDICES, True)
+        nonzeros.zeroEntries()
+
+        for e in range(nel):
+            ie = lgmap.apply(lexico_cg(e))
+            index_bcs(ie, bcflags[e], -1)
+            rows = ie[offdiag]
+            nonzeros.setValues(rows, aones, imode)
+
+        nonzeros.assemble()
+        nonzeros = nonzeros.array.astype(PETSc.IntType)
+        nonzeros = nonzeros + (1 + 2*ndim)
+        nonzeros[self.bc_nodes] = 1
+
+        Pmat = PETSc.Mat().createAIJ(A.getSizes(), nnz=nonzeros[:nloc], comm=A.comm)
+        Pmat.setOption(PETSc.Mat.Option.NEW_NONZERO_ALLOCATION_ERR, False)
+        Pmat.setLGMap(lgmap, lgmap)
+        Pmat.zeroEntries()
+
+        ondiag = (graph == i).T
+        graph = graph.T
+        for e in range(nel):
+            # Fetch rows in lexicographical ordering
+            ie = lgmap.apply(lexico_cg(e))
+            vals = self.stencil.dat.data_ro[lexico_dg(e)]
+
+            # Assemble diagonal
+            for row, aij in zip(ie, vals):
+                Pmat.setValues(row, row, aij[0], imode)
+
+            # Assemble off-diagonal
+            index_bcs(ie, bcflags[e], -1)
+            je = ie[graph]
+            je[ondiag] = -1
+            for row, cols, aij in zip(ie, je, vals):
+                Pmat.setValues(row, cols, aij[1:], imode)
+                Pmat.setValues(cols, row, aij[1:], imode)
+
+        Pmat.assemble()
+        Pmat.zeroRowsColumnsLocal(self.bc_nodes)
+        return Pmat
+
+    @staticmethod
+    def assemble_affine(A, V, Gq, Bq, Sfdm, bcflags, bcdofs):
+        from scipy.sparse import coo_matrix, kron
+
+        imode = PETSc.InsertMode.ADD_VALUES
+        lgmap = V.local_to_global_map([])
+
+        uid, nel = FDMPC.glonum_fun(V)
+        gid, _ = FDMPC.glonum_fun(Gq)
+        bid, _ = FDMPC.glonum_fun(Bq) if Bq is not None else (None, nel)
+
+        ndim = V.mesh().topological_dimension()
+        idsym = [0, 2] if ndim == 2 else [0, 3, 5]
+        idsym = idsym[:ndim]
+
+        nloc = V.dof_dset.set.size
+
+        # Build elemental sparse matrices
+        aa = []  # lists with values and row/column indices (local to each process)
+        ii = []
+        jj = []
+        flag2id = np.kron(np.eye(ndim, ndim, dtype=PETSc.IntType), [[1], [3]])
+        for e in range(nel):
+            fbc = bcflags[e] @ flag2id
+            mue = np.sum(Gq.dat.data_ro[gid(e)], axis=0)
+            mue = mue[idsym]
+
+            be = Sfdm[fbc[0]][1]
+            ae = Sfdm[fbc[0]][0] * mue[0]
+            if Bq is not None:
+                ae += be * sum(Bq.dat.data_ro[bid(e)])
+                ae = ae.tocoo()
+
+            if ndim > 1:
+                ae = kron(ae, Sfdm[fbc[1]][1], format="coo")
+                ae += kron(be, Sfdm[fbc[1]][0] * mue[1], format="coo")
+                ae = ae.tocoo()
+                if ndim > 2:
+                    be = kron(be, Sfdm[fbc[1]][1], format="coo")
+                    ae = kron(ae, Sfdm[fbc[2]][1], format="coo")
+                    ae += kron(be, Sfdm[fbc[2]][0] * mue[2], format="coo")
+                    ae = ae.tocoo()
+
+            ide = uid(e)
+            ii.append(ide[ae.row])
+            jj.append(ide[ae.col])
+            aa.append(ae.data)
+
+        i = np.concatenate(ii)
+        j = np.concatenate(jj)
+        aij = np.concatenate(aa)
+        b = np.ones(aij.shape, dtype=bool)
+        b &= np.logical_not(np.isin(i, bcdofs))
+        b &= np.logical_not(np.isin(j, bcdofs))
+        b |= (i == j)
+        aloc = coo_matrix((aij[b], (i[b], j[b])))
+        aloc = aloc.tocsr()
+
+        nonzeros = aloc.indptr[1:nloc+1] - aloc.indptr[:nloc]
+        Pmat = PETSc.Mat().createAIJ(A.getSizes(), nnz=nonzeros, comm=A.comm)
+        Pmat.setLGMap(lgmap, lgmap)
+        Pmat.zeroEntries()
+        Pmat.setOption(PETSc.Mat.Option.NEW_NONZERO_ALLOCATION_ERR, False)
+        for row in range(aloc.shape[0]):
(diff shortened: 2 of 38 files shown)


firedrake/projection.py
```diff
@@ -120,8 +120,7 @@ def __init__(self, source, target, bcs=None, solver_parameters=None,
             solver_parameters = solver_parameters.copy()
         solver_parameters.setdefault("ksp_type", "cg")
         solver_parameters.setdefault("ksp_rtol", 1e-8)
-        solver_parameters.setdefault("pc_type", "bjacobi")
-        solver_parameters.setdefault("sub_pc_type", "icc")
+        solver_parameters.setdefault("pc_type", "jacobi")
```
pbrubeck, line 123: This is needed for matfree projection. See #2090

firedrake/preconditioners/asm.py
```diff
@@ -267,8 +268,122 @@ def get_patches(self, V):
...
+            dof_per_layer = dof - (nlayers - 1) * layer_offset
+
+        basemeshoff[p - pstart] = off
+        basemeshdof[p - pstart] = dof_per_layer
+        basemeshlayeroffset[p - pstart] = layer_offset
+
+    return basemeshoff, basemeshdof, basemeshlayeroffset
+
+
+class ASMHexStarPC(ASMPatchPC):
```
pefarrell, line 311: So `ASMStarPC` doesn't do the right thing on extruded meshes.

Should `ASMStarPC` raise a warning on extruded meshes?

Does `ASMHexStarPC` work on extruded meshes other than hexes? (Prisms, for example?)

firedrake/preconditioners/fdm.py
```diff
@@ -0,0 +1,1036 @@
...
+import firedrake.dmhooks as dmhooks
+import firedrake
+import numpy
+
+try:
+    from scipy.linalg import eigh
+
+    def sym_eig(A, B):
+        return eigh(A, B)
+except ImportError:
```
pefarrell, line 26: Is there much of a performance penalty to using numpy instead of scipy? Should the code warn here that scipy is not available?

firedrake/preconditioners/fdm.py
```diff
@@ -0,0 +1,1036 @@
...
+        L = npla.cholesky(B)
+        Linv = npla.inv(L)
+        C = numpy.dot(Linv, numpy.dot(A, Linv.T))
+        Z, W = npla.eigh(C)
+        V = numpy.dot(Linv.T, W)
+        return Z, V
+
+
+class FDMPC(PCBase):
+
```
pefarrell, line 39: I think this needs a docstring.

firedrake/preconditioners/fdm.py
```diff
@@ -0,0 +1,1036 @@
...
+        # Get problem solution and bcs
+        solverctx = get_appctx(dm)
+        self.u = solverctx._problem.u
+        self.bcs = solverctx.bcs_F
+
+        if len(self.bcs) > 0:
+            self.bc_nodes = numpy.unique(numpy.concatenate([bcdofs(bc, ghost=False)
+                                                            for bc in self.bcs]))
+        else:
+            self.bc_nodes = numpy.empty(0, dtype=PETSc.IntType)
```
pefarrell, line 74: Is this ever different to `from pyop2.datatypes import IntType`?

firedrake/preconditioners/fdm.py
```diff
@@ -0,0 +1,1036 @@
...
+
+        dm = pc.getDM()
+        V = get_function_space(dm)
+        use_tensorproduct, N, ndim, family, _ = tensor_product_space_query(V)
+
+        if not use_tensorproduct or (family <= {"RTCE", "NCE"}):
+            raise ValueError("The element %s is not supported by FDMPC." % (V.ufl_element(), ))
+
+        needs_interior_facet = not (family <= {"Q", "Lagrange"})
+        Nq = 2*N+1  # quadrature degree
```
pefarrell, line 58: Will we ever want to specify a different quadrature degree?

firedrake/preconditioners/fdm.py
```diff
@@ -0,0 +1,1036 @@
...
+        appctx = self.get_appctx(pc)
+        eta = appctx.get("eta", (N+1)*(N+ndim))  # interior penalty parameter
+        mu = appctx.get("viscosity", None)  # second order coefficient
+        beta = appctx.get("reaction", None)  # zeroth order coefficient
+        eta = float(eta)
+        self.appctx = appctx
+
+        # Obtain the FDM basis and transfer kernels (restriction and prolongation)
+        # Afdm = 1D interval stiffness and mass matrices in the FDM basis for each direction and BC type
+        # Dfdm = normal derivate matrices in the FDM basis
```
pefarrell, line 92: Does normal mean normal to the boundary, or does it mean the usual derivative matrices? (Fix derivate -> derivative anyway)

firedrake/preconditioners/fdm.py
```diff
@@ -0,0 +1,1036 @@
...
+
+        opc = pc
+        # Internally, we just set up a PC object that the user can configure
+        # however from the PETSc command line.  Since PC allows the user to specify
+        # a KSP, we can do iterative by -fdm_pc_type ksp.
+        pc = PETSc.PC().create(comm=opc.comm)
+        pc.incrementTabLevel(1, parent=opc)
+
+        # We set a DM on the constructed PC so one
+        # can do patch solves with ASMPC.
```
pefarrell, line 133: Do you need a DM for `PCASM`? I thought it was purely algebraic.

firedrake/preconditioners/fdm.py
```diff
@@ -0,0 +1,1036 @@
...
+
+    def update(self, pc):
+        self._assemble_Gq()
+        self._assemble_Bq()
+        self.Pmat.zeroEntries()
+        self._assemble_Pmat()
+        self.Pmat.zeroRowsColumnsLocal(self.bc_nodes)
+
+    def applyTranspose(self, pc, x, y):
+        # TODO trivial to implement reusing the code below
```
pefarrell, line 157: If it's not implemented, maybe `raise NotImplementedError`?

firedrake/preconditioners/fdm.py
```diff
@@ -0,0 +1,1036 @@
...
+                     self.uc.dat(op2.WRITE, self.cell_node_map),
+                     self.uf.dat(op2.READ, self.cell_node_map))
+
+        with self.uc.dat.vec_ro as xc:
+            xc.copy(y)
+
+        y.array_w[self.bc_nodes] = x.array_r[self.bc_nodes]
+
+    def view(self, pc, viewer=None):
+        super(FDMPC, self).view(pc, viewer)
```
pefarrell, line 191: Is there anything else useful to print out here about the FDMPC (like quadrature degree, eta, ...)?

firedrake/preconditioners/fdm.py
```diff
@@ -0,0 +1,1036 @@
...
+        if hasattr(self, "pc"):
+            viewer.printfASCII("PC to apply inverse\n")
+            self.pc.view(viewer)
+
+    @staticmethod
+    def pull_axis(x, pshape, idir):
```
pefarrell, line 197: Please add a docstring.

firedrake/preconditioners/fdm.py
```diff
@@ -0,0 +1,1036 @@
...
+        imode = PETSc.InsertMode.ADD_VALUES
+        lgmap = V.local_to_global_map(self.bcs)
+
+        ele = V.ufl_element()
+        ncomp = ele.value_size()
+        bsize = V.value_size
+        ndim = V.mesh().topological_dimension()
+        sdim = V.finat_element.space_dimension()
+
+        needs_hdiv = bsize != ncomp
```
pefarrell, line 210: Comment here

firedrake/preconditioners/fdm.py
```diff
@@ -0,0 +1,1036 @@
...
+        if needs_hdiv:
+            sdim = sdim // ncomp
+
+        if needs_hdiv:
```
pefarrell, line 214: This code could all do with lots more comments.

firedrake/preconditioners/fdm.py
```diff
@@ -0,0 +1,1036 @@
...
+                rows = lgmap.apply(ie)
+                cols = rows[indices]
+                for i, row in enumerate(rows):
+                    i0 = indptr[i]
+                    i1 = indptr[i+1]
+                    A.setValues(row, cols[i0:i1], data[i0:i1], imode)
+                ae.destroy()
+            Bq = None
+
+        # Assemble the viscouos term and the reaction term if any
```
pefarrell, line 261: Typo

firedrake/preconditioners/fdm.py
```diff
@@ -0,0 +1,1036 @@
...
+
+        bcflags = self.get_bc_flags(V, self.mesh, self.bcs, solverctx._problem.J)
+
+        self.weight = self.multiplicity(V)
+        with self.weight.dat.vec as w:
+            w.reciprocal()
+
+        # Get form coefficients
+        appctx = self.get_appctx(pc)
+        eta = appctx.get("eta", (N+1)*(N+ndim))  # interior penalty parameter
```
pefarrell, line 84: Can you document (perhaps in the docstring for FDMPC) what the "viscosity" and "reaction" terms are? How can a general reader know whether this applies to their problem or not?

firedrake/preconditioners/fdm.py
```diff
@@ -0,0 +1,1036 @@
...
+                ie = numpy.reshape(ie, (ncomp, -1))
+
+            mue = numpy.atleast_1d(numpy.sum(Gq.dat.data_ro[gid(e)], axis=0))
+            bce = bcflags[e]
+
+            if Bq is not None:
+                bqe = numpy.atleast_1d(numpy.sum(Bq.dat.data_ro[bid(e)], axis=0))
+                if len(bqe) == 1:
+                    bqe = numpy.tile(bqe, ncomp)
+
```
pefarrell, line 274: Please add comments for this code, it's not very clear to the layman

firedrake/preconditioners/fdm.py
```diff
@@ -0,0 +1,1036 @@
...
+                    cols = rows[indices]
+                    for i, row in enumerate(rows):
+                        i0 = indptr[i]
+                        i1 = indptr[i+1]
+                        A.setValues(row, cols[i0:i1], data[i0:i1], imode)
+                    ae.destroy()
+
+        A.assemble()
+
+    def assemble_coef(self, mu, beta, Nq=0, diagonal=False, transpose=False, piola=False):
```
pefarrell, line 395: Docstring?

firedrake/preconditioners/fdm.py
```diff
@@ -0,0 +1,1036 @@
...
+                    i1, i2, i3, i4, j2, j4 = indices(6)
+                    G = as_tensor(Finv[i2, j2] * Finv[i4, j4] * mu[i1, j2, i3, j4], (i1, i2, i3, i4))
+                else:
+                    raise ValueError("I don't know what to do with the homogeneity tensor")
+        else:
+            F = Jacobian(self.mesh)
+            G = inv(dot(F.T, F))
+            if mu:
+                G = mu * G
+            # I don't know how to use tensor viscosity on embedded manifolds
```
pefarrell, line 431: Should there be an assertion or NotImplementedError here?

firedrake/preconditioners/fdm.py
```diff
@@ -0,0 +1,1036 @@
...
+                                   quad_scheme="default")
+
+            Q = firedrake.FunctionSpace(self.mesh, Qe)
+            q = firedrake.TestFunction(Q)
+            Bq = firedrake.Function(Q)
+            assemble_Bq = partial(firedrake.assemble, inner(beta, q)*dx(degree=Nq), Bq)
+
+        return Gq, Bq, assemble_Gq, assemble_Bq
+
+    def assemble_piola_facet(self, mu):
```
pefarrell, line 478: Docstring

firedrake/preconditioners/fdm.py
```diff
@@ -0,0 +1,1036 @@
...
+                A_petsc.setValues(idx, j, A_numpy[:][j], imode)
+        else:
+            A_petsc.setValue(0, n-1, A_numpy[0][-1], imode)
+            A_petsc.setValue(n-1, 0, A_numpy[-1][0], imode)
+
+        A_petsc.assemble()
+        return A_petsc
+
+    @staticmethod
+    def fdm_cg(Ahat, Bhat):
```
pefarrell, line 561: What is this function doing?

firedrake/preconditioners/pmg.py
```diff
@@ -162,15 +175,15 @@ def coarsen(self, fdm, comm):
         # Coarsen the _SNESContext of a DM fdm
         # return the coarse DM cdm of the coarse _SNESContext
         fctx = get_appctx(fdm)
+        parent = get_parent(fdm)
+        assert parent is not None
 
         # Have we already done this?
+        # FIXME this is triggering weird gmg erros
```
pefarrell, line 182: Is this comment still relevant? It sounds like it should be addressed before merge.

firedrake/preconditioners/pmg.py
```diff
@@ -262,6 +275,11 @@ def coarsen_bcs(fbcs):
                           options_prefix=fctx.options_prefix,
                           transfer_manager=fctx.transfer_manager)
 
+        # TODO coarsen nullspace
```
pefarrell, line 278: Should this TODO be addressed? Should we raise a `NotImplementedError`?

tests/regression/test_fdm.py
```diff
@@ -0,0 +1,200 @@
...
+        else:
+            family = "DQ" if element == "dg" else "Q"
+        return VectorFunctionSpace(mesh, family, degree)
+
+
+@pytest.mark.skipcomplex
+def test_direct_solver(fs):
+    mesh = fs.mesh()
+    x = SpatialCoordinate(mesh)
+    uex = dot(x, x)
```
pefarrell, line 118: Maybe `u_exact` would be a clearer name?

tests/regression/test_fdm.py
```diff
@@ -0,0 +1,200 @@
...
+    u = TrialFunction(fs)
+    v = TestFunction(fs)
+
+    subs = (1, 3)
+    if mesh.layers:
+        subs += ("top",)
+
+    bcs = [DirichletBC(fs, uex, sub) for sub in subs]
+
+    sub_Dir = "everywhere" if "on_boundary" in subs else tuple(s for s in subs if type(s) == int)
```
pefarrell, line 137: Perhaps add a comment explaining your convention here?

tests/regression/test_fdm.py
```diff
@@ -0,0 +1,200 @@
...
+        "ksp_norm_type": "unpreconditioned",
+        "pc_type": "python",
+        "pc_python_type": "firedrake.FDMPC",
+        "fdm_pc_type": "cholesky",
+        "fdm_pc_factor_mat_solver_type": "mumps",
+        "fdm_pc_factor_mat_ordering_type": "nd",
+    }, appctx={"eta": eta, })
+    solver.solve()
+
+    assert solver.snes.ksp.getIterationNumber() == 1 and norm(uex-uh, "H1") < 1.0E-8
```
pefarrell, line 200: I don't see any tests that pass `mu` or `beta`. Can you add a test that exercises this functionality? (Not least so that there's an example of how to use it!)

pefarrell (changes requested):

firedrake/preconditioners/fdm.py
pbrubeck: The quadrature degree `Nq` is used for two things: 
1. to construct the stiffness and mass matrices on the reference interval
2. to sample the form coefficients on a quadrature element to project them to piecewise constants

For the first task Nq=2*N+1 gives the exact result.
The user might want to speed up things and use a cheaper projection for task 2. To save memory, the form coefficients should be stored in DQ_0, since they are utimately reduced to piecewise constants.

pbrubeck: Maybe we could use the quadrature degree from the integral from which each coefficient comes from, or the one from the form compiler parameters to average the coefficients?

firedrake/preconditioners/fdm.py
pbrubeck: This is a one time setup, so this should not affect runtime too much. 
I could not find a nice numpy way to apply the inverse of a triangular matrix other than the explicit inverse approach,
but doing so should not exceed the O(N^3) of calling eigh.

firedrake/preconditioners/asm.py
pbrubeck: `ASMHexStar` works well for prisms (when I implement Pavarino I get h- and p- robust iteration counts). So maybe it sould be renamed?

firedrake/preconditioners/fdm.py
pbrubeck: Yes, it needs the DM to define ISes

firedrake/preconditioners/fdm.py
pbrubeck: Not exactly the same thing:
```
In [14]: pyop2.datatypes.IntType
Out[14]: dtype('int32')

In [15]: type(pyop2.datatypes.IntType)
Out[15]: numpy.dtype[int32]

In [16]: PETSc.IntType
Out[16]: numpy.int32

In [17]: type(PETSc.IntType)
Out[17]: type
```

wence-: I think that numpy will accept a `dtype` object or a `dtype` type in its `dtype` arguments.

firedrake/preconditioners/pmg.py
pbrubeck: I just implemented this. I will also include a test for this.

firedrake/preconditioners/asm.py
pefarrell: Yes, I think it should be renamed. `ASMExtrudedStarPC`?

firedrake/preconditioners/fdm.py
pefarrell: I'm happy hardcoding it. Can you add a comment in the code with what you've said on this thread?

firedrake/preconditioners/pmg.py
```diff
@@ -283,14 +300,50 @@ def coarsen_bcs(fbcs):
...
-        # Injection of the initial state
+        # injection of the initial state
         def inject_state(mat):
             with cu.dat.vec_wo as xc, fu.dat.vec_ro as xf:
                 mat.multTranspose(xf, xc)
 
         injection = self.create_injection(cdm, fdm)
         add_hook(parent, setup=partial(inject_state, injection), call_setup=True)
 
+        # restrict the nullspace basis
```
pefarrell, line 311: Are there any tests that exercise this?

tests/regression/test_fdm.py
```diff
@@ -0,0 +1,203 @@
...
+        "ksp_norm_type": "unpreconditioned",
+        "pc_type": "python",
+        "pc_python_type": "firedrake.FDMPC",
+        "fdm_pc_type": "cholesky",
+        "fdm_pc_factor_mat_solver_type": "mumps",
+        "fdm_pc_factor_mat_ordering_type": "nd",
+    }, appctx={"eta": eta, })
+    solver.solve()
+
+    assert solver.snes.ksp.getIterationNumber() == 1 and norm(u_exact-uh, "H1") < 1.0E-8
```
pefarrell, line 203: The tests don't appear to exercise the full range of functionality. For example,  could there be a test that uses spatially  varying coefficients,  as well as lower order terms?

pefarrell (changes requested):

firedrake/preconditioners/pmg.py
pbrubeck: Now there is one.

wence-:
> Removing the numpy import will potentially affect a lot of users, but this should be very easy to fix. The application libraries can be imported, but numpy might potentially be missing in some of their tests.

I am confused by this:

```
from firedrake import *

numpy => NameError
```

At least on my firedrake.

wence-:
> > Removing the numpy import will potentially affect a lot of users, but this should be very easy to fix. The application libraries can be imported, but numpy might potentially be missing in some of their tests.
> 
> I am confused by this:
> 
> ```
> from firedrake import *
> 
> numpy => NameError
> ```
> 
> At least on my firedrake.

Ah, `np`, dropping that is not a problem.

firedrake/preconditioners/pmg.py
pbrubeck: This was giving me errors for different coarsening schedules within a nested iteration (p-FAS/p-MG). I resolved this by ensuring that the coarse `_SNESContext` cached by the outer solver is in fact the one required by the inner solver. When this is not the case, we need to rediscretize the problem with the degree of the inner solver and overwrite the `_coarse` attribute of the `_SNESContentext`.

pefarrell (dismissed):

firedrake/preconditioners/fdm.py
```diff
@@ -0,0 +1,1134 @@
...
+            Bq = firedrake.Function(Q)
+            coefficients["Bq"] = Bq
+            assembly_callables.append(partial(firedrake.assemble, inner(beta, q)*dx(degree=quad_deg), Bq))
+
+        if needs_hdiv:
+            # make DGT functions with the second order coefficient
+            # and the Piola transform matrix for each side of each facet
+            extruded = mesh.cell_set._extruded
+            dS_int = firedrake.dS_h(degree=quad_deg) + firedrake.dS_v(degree=quad_deg) if extruded else firedrake.dS(degree=quad_deg)
+            ele = BrokenElement(FiniteElement("DGT", mesh.ufl_cell(), 0))
```
wence-, line 546: So this is one dof per facet, but associated with the cell because you need different values on either side of the facet I guess.

wence- (commented):
I had mostly minor comments.

I do not claim that I fully understand all the intricacies in this code. I think the core idea is the following:

1. There exists a basis for Q_k (etc...) in which certain bilinear forms are actually sparse (as sparse as the Q_1 thing approximately, so we can assemble them as sparse matrices)
2. This preconditioner works by applying the change of basis (which can be done dimension-by-dimension in a O(p^{d+1}) way from standard GLL to this special basis, call it `T_sp`
3. In the special basis you assemble the operator taking care to maintain the sparsity, this can now be sent to PETSc (call this operator `A_sp`)

So the action of the preconditioner is something like:

```
y <- T_sp^T A_sp^-1 T_sp x
```

You have to do a lot of work by hand here to construct A_sp correctly (mostly) including making certain approximations on coefficients in the form. I _think_ (though do not know) that if you implemented the fancy 1D basis in FIAT, it should be possible to deliver `T_sp` as the interpolation between two finat elements (and this should now have good complexity).

The main work in `A_sp` is that you're having to do all of the assembly by hand, because we can't deliver you an element kernel that maintains sparsity, and even if we could do that, can't deliver the insertion in a sparsity-preserving way.


It would be interesting (not for this PR) to think about what would be required to do this.

tests/multigrid/test_p_multigrid.py
```diff
@@ -166,52 +190,71 @@ def test_p_multigrid_vector():
...
           "snes_type": "ksponly",
-          "ksp_type": "fgmres",
+          "ksp_type": "cg",
+          "ksp_rtol": 1E-12,
           "ksp_monitor_true_residual": None,
           "pc_type": "python",
-          "pc_python_type": __name__ + ".MixedPMG",
-          "mat_type": "aij",
+          "pc_python_type": "firedrake.PMGPC",
+          # "mat_type": mat_type,  # FIXME bug with mat-free jacobi on MixedFunctionSpace
```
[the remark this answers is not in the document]
pbrubeck: `mat_type` is also used for the p-mg restriction matrix, which is used for the nullspace coarsening. Having a fully matrix-free solver is not really relevant for our tests here, and the issue has been raised already here #2221

tests/multigrid/test_p_multigrid.py
```diff
@@ -166,52 +190,71 @@ def test_p_multigrid_vector():
...
           "pmg_pc_mg_type": "multiplicative",
           "pmg_mg_levels": relax,
-          "pmg_mg_coarse_ksp_type": "richardson",
-          "pmg_mg_coarse_ksp_max_it": 1,
-          "pmg_mg_coarse_ksp_norm_type": "unpreconditioned",
-          "pmg_mg_coarse_ksp_monitor": None,
-          "pmg_mg_coarse_pc_type": "lu"}
+          "pmg_mg_coarse": coarse}
+
+    basis = VectorSpaceBasis([assemble(TestFunction(Z.sub(1))*dx)])
```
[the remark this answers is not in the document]
pbrubeck: This is the dual of a constant, if I use constant=True, the zero mean constraint is not enforced correctly across the p-mg levels.

firedrake/preconditioners/fdm.py
```diff
@@ -0,0 +1,1134 @@
...
+            pshape = [Ak[0].size[0] for Ak in Afdm]
+
+        # we need to preallocate the diagonal of Dirichlet nodes
+        for row in V.dof_dset.lgmap.indices:
+            A.setValue(row, row, 0.0E0, imode)
+
+        # assemble zero-th order term separately, including off-diagonals (mixed components)
+        # I cannot do this for hdiv elements as off-diagonals are not sparse, this is because
+        # the FDM eigenbases for GLL(N) and GLL(N-1) are not orthogonal to each other
+        use_separate_reaction = False if Bq is None else not needs_hdiv and Bq.ufl_shape
```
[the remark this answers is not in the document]
pbrubeck: I need a ternary operator because when `Bq is None` I cannot ask for `Bq.ufl_shape`.

firedrake/preconditioners/fdm.py
```diff
@@ -0,0 +1,1134 @@
...
+        else:
+            facet_to_nodes_fun = lambda e: facet_to_nodes[e]
+            local_facet_data_fun = lambda e: local_facet_data[e]
+            nfacets = nbase
+
+        return facet_to_nodes_fun, local_facet_data_fun, nfacets
+
+    @staticmethod
+    @lru_cache(maxsize=10)
+    def glonum_fun(node_map):
```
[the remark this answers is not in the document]
pbrubeck: Correct, but I would also need to consider orientation when dealing with the interior facet integrals.

tests/multigrid/test_p_multigrid.py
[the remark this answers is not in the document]
wence-: Thanks.

pbrubeck:
> You have to do a lot of work by hand here to construct A_sp correctly (mostly) including making certain approximations on coefficients in the form. I _think_ (though do not know) that if you implemented the fancy 1D basis in FIAT, it should be possible to deliver `T_sp` as the interpolation between two finat elements (and this should now have good complexity).

Yes, I would be very keen to implement this idea. Furthermore we need to figure out a way to take the transpose, which would be very useful for p-MG.

firedrake/preconditioners/fdm.py
```diff
@@ -0,0 +1,1134 @@
...
+
+        dm = pc.getDM()
+        V = get_function_space(dm)
+        use_tensorproduct, N, ndim, family, _ = tensor_product_space_query(V)
+
+        if not use_tensorproduct or (family < {"RTCE", "NCE"}):
+            raise ValueError("FDMPC does not support the element %s" % V.ufl_element())
+
+        needs_interior_facet = not (family <= {"Q", "Lagrange"})
+        needs_hdiv = family < {"RTCF", "NCF"}
```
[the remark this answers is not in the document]
pbrubeck: I'm afraid that these are not the same, as `family` is a set with the families of each sub-element of a TP element.

firedrake/preconditioners/fdm.py
```diff
@@ -0,0 +1,1134 @@
...
+                            else:
+                                smu = sij*mu
+
+                            Adense[ii, jj] += eta * sum(smu)
+                            Adense[i0:i1, jj] -= smu[i] * Dfacet[:, iface % 2]
+                            Adense[ii, j0:j1] -= smu[j] * Dfacet[:, jface % 2]
+
+                    Ae = FDMPC.numpy_to_petsc(Adense, dense_indices, diag=False)
+                    if ndim > 1:
+                        # assume that the mesh is oriented
```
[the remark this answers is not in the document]
wence-: Sort of: UFL does this when lowering `JacobianDeterminant` to a computation on the coordinate field:

```
        if domain.topological_dimension() < domain.geometric_dimension():
            co = CellOrientation(domain)
            detJ = co * detJ
```

So logically it's a sign flip on each cell. the `mesh.cell_orientations()` method returns a dat that is `0` if the sign flip is positive, and `1` if the sign flip is negative (we should just break backwards-compat with old-dolfin and put the obvious thing in instead).

tests/regression/test_fdm.py
```diff
@@ -0,0 +1,274 @@
...
+                "pc_type": "python",
+                "pc_python_type": "firedrake.AssembledPC",
+                "assembled_pc_type": "cholesky",
+            }})
+        solver.solve()
+        nits.append(solver.snes.ksp.getIterationNumber())
+    assert (nits == expected)
+
+
+@pytest.mark.skipcomplex
```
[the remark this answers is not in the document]
pbrubeck: The problem in complex mode should have a real solution. I am not entirely sure but it might be that something was breaking in complex mode, either PMGPC or FDMPC.

firedrake/preconditioners/fdm.py
```diff
@@ -0,0 +1,1134 @@
...
+        # Preallocate by calling the assembly routine on a PREALLOCATOR Mat
+        prealloc = PETSc.Mat().create(comm=A.comm)
+        prealloc.setType(PETSc.Mat.Type.PREALLOCATOR)
+        prealloc.setSizes(A.getSizes())
+        prealloc.setUp()
+        self.assemble_kron(prealloc, V, coefficients, Afdm, Dfdm, eta, bcflags, needs_hdiv)
+        nnz = get_preallocation(prealloc, V.value_size * V.dof_dset.set.size)
+
+        self.Pmat = PETSc.Mat().createAIJ(A.getSizes(), nnz=nnz, comm=A.comm)
+        self.Pmat.setBlockSize(V.value_size)
```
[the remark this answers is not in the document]
pbrubeck: Does firedrake use `BAIJ`? I thought matrices for vector fields were always `AIJ`.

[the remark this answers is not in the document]
wence-: It can do, but it appears not to be the default.
```
In [12]: V = VectorFunctionSpace(mesh, "P", 1)

In [13]: u = TrialFunction(V)

In [14]: v = TestFunction(V)

In [15]: a = inner(u, v)*dx

In [16]: A = assemble(a)
A.petscmat.getType()

In [17]: A.petscmat.getType()
Out[17]: 'seqaij'

In [18]: A = assemble(a, mat_type="baij")

In [19]: A.petscmat.getType()
Out[19]: 'seqbaij'
```

I'm happy to leave this is as is.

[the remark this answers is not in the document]
pbrubeck: BAIJ does not seem to be very compatible with CHOLMOD
```
[0]PETSC ERROR: MatSolverType cholmod does not support matrix type seqbaij
```
If I use PETSc instead, then nested dissection is not available:
```
[0]PETSC ERROR: Matrix reordering is not supported
```
So I'm also happy to leave it as it is.

firedrake/preconditioners/fdm.py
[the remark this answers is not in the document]
pbrubeck: Apart from getting the correct sign on the Piola transform, I think I need to permute the DOFs somehow? I cannot work this out immediately, so maybe it's fine if I raise an exception for facet integrals on immersed meshes.

firedrake/preconditioners/asm.py
```diff
@@ -267,8 +273,122 @@ def get_patches(self, V):
...
+    return basemeshoff, basemeshdof, basemeshlayeroffset
+
+
+class ASMExtrudedStarPC(ASMPatchPC):
+    '''Patch-based PC using Star of mesh entities implmented as an
+    :class:`ASMPatchPC`.
+
+    ASMExtrudedStarPC is an additive Schwarz preconditioner where each patch
+    consists of all DoFs on the topological star of the mesh entity
+    specified by `pc_hexstar_construct_dim`.
```
dham, line 322: Incorrect prefix.

firedrake/preconditioners/fdm.py
```diff
@@ -0,0 +1,1145 @@
...
+                i1, i2, i3, i4, j2, j4 = indices(6)
+                G = as_tensor(Finv[i2, j2] * Finv[i4, j4] * alpha[i1, j2, i3, j4], (i1, i2, i3, i4))
+            else:
+                raise ValueError("I don't know what to do with the homogeneity tensor")
+        else:
+            if len(alpha.ufl_shape) == 4:
+                i1, i2, i3, i4, j1, j2, j3, j4 = indices(8)
+                G = as_tensor(Piola[j1, i1] * Finv[i2, j2] * Piola[j3, i3] * Finv[i4, j4] * alpha[j1, j2, j3, j4], (i1, i2, i3, i4))
+            else:
+                raise ValueError("I don't know what to do with the homogeneity tensor")
```
dham, line 506: Think about the most helpful error message here.

PR #2028, opened 2021-04-15
WIP: Pytential + volumential as external operators

benSepanski:
I'm making a PR here so that we can start unifying the layer potential side of things.

This PR makes two big changes to the current `pointwise-adjoint-layer-potentials` branch.

1. Merge in a lot from master so that we support complex
2. Avoid duplicated code between volumential/pytential external operators by unifying under an abstract class

Here's the current directory layout:
- `potential_evaluation/`
    - `__init__.py` the basic `Potential` class for developers to implement
    - `potentials.py` the public interface to create potentials
    - `pytential.py` `pytential` implementation of the `Potential` class
    - `volumential.py` the `volumential` implementation of the `Potential` class


benSepanski:
Currently, the `test_layer_potentials` test is failing because of lack of automatic differentiation support

benSepanski:
`volumential.py` is still a work in progress, so once we are happy with `__init__.py`, `potentials.py`, and `pytential.py` I will finish the `volumential` part

benSepanski:
I also stopped using an intermediate `Evaluator` class, and just implement the methods in the `ExternalOperator` directly

rckirby:
Thanks!  I had put that in to prevent multiple -tentials from getting
created, which seemed to forcing us to re-build trees and work around some
UFL stuff, but I'm glad to drop it if you've got a better way.

PR #2031, opened 2021-04-19
Added HDG paper and arXiv preprint to bibliography

JDBetteridge:


wence-:
Why does this change _every_ bibtex entry? Did the bibtexparser run change somehow?

wence-:
@JDBetteridge can you remove all of the whitespace changes?

JDBetteridge:
Your first comment highlights the issue: Running `firedrake-preprocess-bibtex firedrake-apps.bib` as per https://github.com/firedrakeproject/firedrake/blob/master/docs/source/_static/firedrake-apps.bib is what's adding the additional whitespace.

I was going to edit `firedrake-preprocess-bibtex` to "fix" this, but that isn't the issue, the issue is my inclusion of the line
```
archiveprefix = {arXiv},
```
which has a longer tag than everywhere else in the file. This causes the writer to add an extra space to align the `=` character throughout. I could just remove this tag, then everything goes back to how it was, however this bibtex entry was the automatically generated arXiv bibtex formatted citation that I just copy-pasta'd in.

How would you like me to proceed?:
1. Get rid of this tag, we don't need it, we can just tell everyone to not just copy paste from arXiv (I could add a comment to the top).
2. One off, add a whole bunch more whitespace to this file so that we don't run into this issue every time someone pastes in an automatically generated arXiv bibtex formatted citation (which suffices until someone pastes an even longer tag in there).
3. Something else

docs/source/_static/firedrake-apps.bib
```diff
@@ -6,1082 +6,1106 @@ @comment{
...
+  volume        = {426},
+  year          = {2021}
+}
+
+@misc{Betteridge2021b,
+  archiveprefix = {arXiv},
+  author        = {Jack D. Betteridge and Patrick E. Farrell and David A. Ham},
+  eprint        = {2104.08012},
+  primaryclass  = {cs.MS},
+  title         = {Code generation for productive portable scalable finite element simulation in Firedrake},
```
dham, line 163: Braces please.

dham (changes requested):

PR #2033, opened 2021-04-21
Loopy sprint read write output args in slate

sv2518:
This adapting to 714b68579b784a869d789e2c639348e1753727f9.

### DIFF
--- a/firedrake/slate/slac/compiler.py
+++ b/firedrake/slate/slac/compiler.py
@@ -627,7 +627,7 @@ def gem_to_loopy(gem_expr, var2terminal, scalar_type):
     shape = gem_expr.shape if len(gem_expr.shape) != 0 else (1,)
     idx = make_indices(len(shape))
     indexed_gem_expr = gem.Indexed(gem_expr, idx)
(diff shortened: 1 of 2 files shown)


sv2518:
There is still something missing, will need to add the output arg to the variables of the Slate CallInstructions too.

firedrake/slate/slac/compiler.py
```diff
@@ -627,7 +627,7 @@ def gem_to_loopy(gem_expr, var2terminal, scalar_type):
     shape = gem_expr.shape if len(gem_expr.shape) != 0 else (1,)
     idx = make_indices(len(shape))
     indexed_gem_expr = gem.Indexed(gem_expr, idx)
-    args = ([loopy.GlobalArg("output", shape=shape, dtype=scalar_type)]
+    args = ([loopy.GlobalArg("output", shape=shape, dtype=scalar_type, is_output=True, is_input=True)]
```
wence-, line 630: Should TSFC be setting these inout parameters on its globalargs too?

sv2518: I would think so yes, but want so see if there is a test actually failing because it's not there. The test would fail with a KeyError in the Kernel matching.

PR #2034, opened 2021-04-21
Explicit Dual Space support in Assemble

indiamai:
Work that still needs to be done:
- [ ] Add Hermitian transpose to petsc4py
- [ ] Support for Mixed Spaces


wence-:
>  Add Hermitian transpose to petsc4py

I think I did this and I just need to update our petsc branch.

firedrake/functionspace.py
```diff
@@ -135,6 +135,45 @@ def FunctionSpace(mesh, family, degree=None, name=None, vfamily=None,
         return new
 
 
+@timed_function("CreateSpace")
+def DualSpace(mesh, family, degree=None, name=None, vfamily=None,
```
wence-, line 139: Do we need this constructor or should one just construct dual spaces with `Vstar = FunctionSpace(...).dual()` ?

firedrake/functionspaceimpl.py
```diff
@@ -232,13 +232,31 @@ def __getattr__(self, name):
...
         return list(OrderedDict.fromkeys(dir(self.topological) + current))
 
     def collapse(self):
         return type(self)(self.topological.collapse(), self.mesh())
 
 
+class WithGeometry(WithGeometryBase, ufl.FunctionSpace):
+
+    def __init__(self, function_space, mesh):
+        super(WithGeometry, self).__init__(function_space, mesh)
```
wence-, line 245: No need for pass-through `__init__`.

firedrake/functionspaceimpl.py
```diff
@@ -232,13 +232,31 @@ def __getattr__(self, name):
...
+
+    def dual(self):
+        return FiredrakeDualSpace(self.topological, self.mesh())
+
+
+class FiredrakeDualSpace(WithGeometryBase, ufl.functionspace.DualSpace):
```
wence-, line 251: bikeshed: the name is a bit odd. Could we just call this a `DualSpace` ?

firedrake/matrix.py
```diff
@@ -153,3 +170,42 @@ def assemble(self):
...
+        conditions to apply.
+
+    :arg petscmat: the already constructed petsc matrix this object represents.
+
+
+    .. note::
+
+        This object acts to the right on an assembled :class:`.Function`
+        and to the left on an assembled cofunction (currently represented
+        by a :class:`.Function`).
```
wence-, line 192: Is this note still correct now that we have `Cofunction` objects?

firedrake/matrix.py
```diff
@@ -153,3 +170,42 @@ def assemble(self):
...
+
+    """
+    def __init__(self, a, bcs, petscmat, *args, **kwargs):
+        super(AssembledMatrix, self).__init__(a, bcs, "assembled")
+
+        self.petscmat = petscmat
+
+        # this allows call to self.M.handle without a new class
+        self.M = SimpleNamespace(handle=self.mat())
```
wence-, line 201: I _think_ it should be the case that we should expunge `matrix.M.handle` and just use `matrix.petscmat` everywhere, but maybe this is too much for this PR.

firedrake/matrix.py
```diff
@@ -153,3 +170,42 @@ def assemble(self):
...
+
+    def mat(self):
+        return self.petscmat
```
wence-, line 204: I think this method is only used for the `SimpleNamespace` above? If so, probably not needed?

firedrake/matrix.py
wence-: If necessary, just use `self.M = SimpleNamespace(handle=self.petscmat)` and remove the need for the `mat()` method?

firedrake/matrix.py
```diff
@@ -94,7 +111,7 @@ class Matrix(MatrixBase):
 
     def __init__(self, a, bcs, mat_type, *args, **kwargs):
         # sets self._a, self._bcs, and self._mat_type
-        super(Matrix, self).__init__(a, bcs, mat_type)
+        MatrixBase.__init__(self, a, bcs, mat_type)
```
wence-, line 114: Does the `super()` call not do correct method resolution?

firedrake/matrix.py
```diff
@@ -1,30 +1,41 @@
...
-    :arg a: the bilinear form this :class:`MatrixBase` represents.
+    :arg a: the bilinear form this :class:`MatrixBase` represents
+            or a tuple of the arguments it represents
 
     :arg bcs: an iterable of boundary conditions to apply to this
         :class:`MatrixBase`.  May be `None` if there are no boundary
         conditions to apply.
     :arg mat_type: matrix type of assembled matrix, or 'matfree' for matrix-free
     """
     def __init__(self, a, bcs, mat_type):
```
wence-, line 24: The base class should ideally not have to dispatch on types of arguments. I think now it is the case that every `MatrixBase` "has-a" arguments property, but only `Matrix` and `ImplicitMatrix` "has-a" `.a` form.

So I think the better refactoring is to make this:

```python
def __init__(self, test, trial, bcs, mat_type):
    self._arguments = test, trial
```

And then `Matrix` does

```
def __init__(self, a, ...):
    super().__init__(*a.arguments(), ...)
    self.a = a
```

while `AssembledMatrix` does
```
def __init__(self, test, trial, ...):
    super().__init__(test, trial, ...)
```

firedrake/matrix.py
```diff
@@ -153,3 +170,42 @@ def assemble(self):
...
+
+class AssembledMatrix(MatrixBase):
+    """A representation of an matrix without knowing the underlying form.
+     This class wraps the relevant information for Python PETSc matrix.
+
+    :arg a: A tuple of the arguments the matrix represents
+
+    :arg bcs: an iterable of boundary conditions to apply to this
+        :class:`Matrix`.  May be `None` if there are no boundary
+        conditions to apply.
```
wence-, line 183: I think this is not right? They're already applied?

firedrake/assemble.py
```diff
@@ -120,15 +121,134 @@ def assemble(expr, tensor=None, bcs=None, *,
...
+            petsc_mat = lhs.M.handle
+            (row, col) = lhs.arguments()
+            res = _make_vector(col)
+
+            with rhs.dat.vec_ro as v_vec:
+                with res.dat.vec as res_vec:
+                    petsc_mat.mult(v_vec, res_vec)
+            return firedrake.Cofunction(row.function_space(), val=res.dat)
+        elif isinstance(rhs, matrix.MatrixBase):
+            petsc_mat = lhs.M.handle
```
wence-, line 210: I think you can always use `lhs.petscmat`?

firedrake/assemble.py
```diff
@@ -120,15 +121,134 @@ def assemble(expr, tensor=None, bcs=None, *,
...
+            (row, col) = lhs.arguments()
+            res = PETSc.Mat().create()
+            # TODO Figure out what goes here
+            res = petsc_mat.matMult(rhs.M.handle)
+            return matrix.AssembledMatrix(rhs.arguments(), bcs, res,
```
wence-, line 215: It seems like we don't do anything with the `bcs` in this case, so are we "lying to children" here?

firedrake/assemble.py
```diff
@@ -120,15 +121,134 @@ def assemble(expr, tensor=None, bcs=None, *,
...
+                                          options_prefix=options_prefix)
+        else:
+            raise TypeError("Mismatching FormSum shapes")
+    elif isinstance(expr, (ufl.Cofunction, ufl.Coargument, ufl.Matrix)):
+        return expr
+    elif isinstance(expr, ufl.Coefficient):
+        return expr
     else:
-        raise TypeError(f"Unable to assemble: {expr}")
+        print(type(expr))
```
wence-, line 250: Debugging statement?

scripts/firedrake-zenodo
```diff
@@ -11,10 +11,11 @@ import json
 import time
```
wence-, line 11: I think these changes are already in trunk too.

scripts/firedrake-install
```diff
@@ -1538,7 +1538,7 @@ if mode == "install":
         source = "https://github.com/firedrakeproject/VTKPythonPackage/releases/download/firedrake_20210113/"
```
wence-, line 1538: Changes already in trunk?

tests/regression/test_assemble_formbase.py
```diff
@@ -0,0 +1,169 @@
...
+    res = assemble(a)
+    preassemble = assemble(a + a)
+    formsum = res + a
+    res2 = assemble(formsum)
+
+    assert isinstance(formsum, ufl.form.FormSum)
+    assert isinstance(res2, Cofunction)
+    assert isinstance(preassemble, Cofunction)
+    
+    assert abs(preassemble.dat.data.sum() - res2.dat.data.sum()) < 1.0e-12
```
wence-, line 132: `assert numpy.allclose(preassemble.dat.data_ro, res2.dat.data_ro)` ?

PR #2036, opened 2021-04-21
Equation bc tape mc

mc4117:
A preliminary pull request to see if the taping and matrix assembly for the adjoint is correct.

### DIFF
--- a/firedrake/assemble.py
+++ b/firedrake/assemble.py
@@ -488,7 +488,7 @@ def _get_mat_type(mat_type, sub_mat_type, arguments):
     return mat_type, sub_mat_type
 
 
-def _collect_lgmaps(matrix, all_bcs, Vrow, Vcol, row, col):
+def _collect_lgmaps(matrix, all_bcs, Vrow, Vcol, row, col, is_adjoint):
     """Obtain local to global maps for matrix insertion in the
     presence of boundary conditions.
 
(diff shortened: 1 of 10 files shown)


firedrake/assemble.py
```diff
@@ -57,7 +57,8 @@ def assemble(expr, tensor=None, bcs=None, *,
              mat_type=None,
              sub_mat_type=None,
              appctx={},
-             options_prefix=None):
+             options_prefix=None,
+             adj=False):
```
ksagiyam, line 61: `assemble` may not want to know if it is assembling `adjoint` or not (I think it should just assemble a given `expr`), but we can think about this after you get things working.

mc4117: I did this so that _collect_lgmaps would know whether it is assembling adjoint or not but I agree that this is a bit hacky

firedrake/assemble.py
```diff
@@ -513,7 +520,10 @@ def _collect_lgmaps(matrix, all_bcs, Vrow, Vcol, row, col):
     clgmap = Vcol[col].local_to_global_map(bccol, lgmap=clgmap)
```
wence-, line 520: EquationBCs are the only ones that need to be treated specially for `is_adjoint`, where are they handled?

firedrake/assemble.py
```diff
@@ -513,7 +520,10 @@ def _collect_lgmaps(matrix, all_bcs, Vrow, Vcol, row, col):
...
     unroll = any(bc.function_space().component is not None
                  for bc in chain(bcrow, bccol))
-    return (rlgmap, clgmap), unroll
+    if is_adjoint:
+        return (clgmap, rlgmap), unroll
+    else:
+        return (rlgmap, clgmap), unroll
```
wence-, line 526: This can't be right (check for a matrix with different row and column spaces!)

firedrake/assemble.py
```diff
@@ -659,12 +670,15 @@ def _make_parloops(expr, tensor, bcs, diagonal, fc_params, assembly_rank):
             raise NotImplementedError("Assembly with multiple meshes not supported.")
 
     if assembly_rank == _AssemblyRank.MATRIX:
+
         test, trial = expr.arguments()
+
```
wence-, line 675: Please don't make whitespace changes like this.

wence-:
We think that `EquationBC` objects must somehow know that they have been adjointed, because then assembly assembles the adjoint form and the adjoint boundary condition object. That can then know to zero columns rather than rows.

dham:
Inactive.

Issue #1973, opened 2021-02-11
Generalization of interface for PatchPC PlaneSmoother


ScottMacLachlan:
The current interface to `PlaneSmoother` allows the user to specify an axis for the relaxation, direction (increasing or decreasing in coordinate), and number of divisions to separate the axis range into.  Currently, the division of the axis range is into evenly spaced divisions: https://github.com/firedrakeproject/firedrake/blob/33ec106cf7a50aead0efcccbcf993617bb498c2d/firedrake/preconditioners/patch.py#L622

This is the desired behaviour for uniform meshes, but wreaks a moderate amount of havoc when you have unevenly spaced (but still tensor-product) meshes, as you might when you have boundary layers (or other reasons for 1D refinement in a tensor-product mesh).  I'd like to propose that we generalize the specification of `divisions` in the above line from `patch.py`, and am wondering what a good interface would be for this.

Tangentially related is the question of using this in a multigrid method, where the desired number of patches (lines) in the relaxation scheme should change with level in the mesh hierarchy.  Ideally, we could specify an interface that would make this natural.

My current desired use case is this: I'm solving singularly perturbed problems on Shishkin meshes of the unit square, using line relaxation in a multigrid method.  For a simple model and reasonable parameters (say a 128x128 mesh, with perturbation parameter 1e-6), I'll have two regions in my mesh's x-coordinates; one with 32-64 elements of width O(1e-8), and one with the remainder of elements of width O(1e-2).  I could make the existing interface DWIW by setting ndiv = O(1e+8) to resolve the fine mesh spacing, but this results in a vast number of empty index sets where the spacing is coarse.  However, from my mesh construction, I know exactly where the node of my mesh are (already computed in a 1D array), so I could easily specify where the lines should be for relaxation by passing in that array, or something similar.

Are there other use cases that would benefit from this sort of interface or other generalizations?

wence-:
> Tangentially related is the question of using this in a multigrid method, where the desired number of patches (lines) in the relaxation scheme should change with level in the mesh hierarchy. Ideally, we could specify an interface that would make this natural.

The normal way to do this is to specify configuration per level in the MG options:
```
mg_level_0_
mg_level_1_
```

We can dispatch on `mg.ufl_utils.get_level`, but it might be fiddly to DTRT in all cases.

We should condense empty index sets out in the construction anyway
```python
# instead of this
            for patch in entities:
[...]

PR #2037, opened 2021-04-21
Generalizing interface for PlaneSmoother

ScottMacLachlan:
This tries to address #1973, allowing a more general interface for PlaneSmoother to specify which set of points to use to divide an axis to define the PlaneSmoother patch.  Now, if the user gives a number in the usual `sweep` format (e.g., `0+50`), then the old behaviour is preserved, with evenly spaced bins.  If, instead, the user gives a key (e.g., `0+key`), then that key is looked for in the `appctx` and used as `divisions` in place of the evenly spaced bins.


firedrake/preconditioners/patch.py
```diff
@@ -585,10 +585,13 @@ def coords(dm, p, coordinates):
         bary /= len(closure_of_p)
         return bary
 
-    def sort_entities(self, dm, axis, dir, ndiv):
+    def sort_entities(self, dm, axis, dir, ndiv=-1, divisions=None):
```
dham, line 588: Use `None` as the out of band value, not `-1`

dham (changes requested):

firedrake/preconditioners/patch.py
ScottMacLachlan: Done.

ScottMacLachlan:
Alright - my apologies for leaving this sitting for so long, but ...  I think it's ready for a proper review now.

This now includes two enhancements to PlaneSmoother, the original one (allowing more general specification of the spacing of bins into which to divide points) and a second one allowing more general selection of how points are mapped into those bins.  

Two of the tests check correctness of these against the original syntax for PlaneSmoother, while the other two check reference performance for their intended use cases, on non-uniform tensor-product meshes, or for line/plane relaxation that doesn't align with the coordinate axes...

dham (changes requested):
Update to current master and check that this passes tests, and we're good.

ScottMacLachlan:
Update to master seems to have gone smoothly...

dham (approved):

PR #2040, opened 2021-04-23
Specify is_input and is_output for loopy.GlobalArg objects

connorjward:
I'm worried about getting the correct access modifiers here. `op2.MIN` and `op2.MAX` have been omitted because I don't know how those work. This now passes `test_expressions.py` though!

### DIFF
--- a/firedrake/assemble_expressions.py
+++ b/firedrake/assemble_expressions.py
@@ -425,8 +425,8 @@ def pointwise_expression_kernel(exprs, scalar_type):
             except KeyError:
                 continue
             plargs.append(arg)
-            is_input = arg.access in [op2.INC, op2.READ, op2.RW]
-            is_output = arg.access in [op2.INC, op2.RW, op2.WRITE]
+            is_input = arg.access in [op2.INC, op2.MAX, op2.MIN, op2.READ, op2.RW]
+            is_output = arg.access in [op2.INC, op2.MAX, op2.MIN, op2.RW, op2.WRITE]
             args.append(loopy.GlobalArg(var.name, shape=var.shape, dtype=c.dat.dtype, is_input=is_input, is_output=is_output))
     assert len(coefficients) == 0
     name = "expression_kernel"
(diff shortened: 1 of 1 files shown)


firedrake/assemble_expressions.py
```diff
@@ -425,7 +425,9 @@ def pointwise_expression_kernel(exprs, scalar_type):
             except KeyError:
                 continue
             plargs.append(arg)
-            args.append(loopy.GlobalArg(var.name, shape=var.shape, dtype=c.dat.dtype))
+            is_input = arg.access in [op2.INC, op2.READ, op2.RW]
+            is_output = arg.access in [op2.INC, op2.RW, op2.WRITE]
```
wence-, line 429: I think MIN and MAX behave like INC

connorjward: Done

sv2518: This looks correct to me.

I am only wondering if we are doing the right thing in PyOP2 if an arg is WRITE. I think we need to do the change here because of this commit https://github.com/OP2/PyOP2/commit/714b68579b784a869d789e2c639348e1753727f9, right? There it looks like everything that is not READ is_input and is_output. I think we should special case for WRITE there as well? There is also this commit https://github.com/OP2/PyOP2/commit/61b9bbbb9ec744eb77468f98d75f3dbc37c0afc5 which treats WRITE as INC, whenever we require it the output argument to be zeroed.

sv2518: Ah nvm, I just realised we did that in a later commit.

sv2518: So yes you can see from those commits that MIN and MAX should behave like INC

wence-: There's a subtlety for `WRITE` arguments between `op2.Kernel` objects which have `requires_zeroed_output_arguments=True` (which is what all kernels that come from TSFC have), where they have `output += ...` which loopy then treats as RW access, and kernels for which `requires_zeroed_output_arguments=False` (which, for example, come from the pointwise expression compiler) for which `WRITE` arguments should be treated as `WRITE` for the purposes of loopy.

sv2518: Ah I understand, okay.

connorjward: I take it what I have written is correct then since we are dealing with pointwise expressions?

PR #2042, opened 2021-04-27
Removed spurious whitespace from VTK wheel URL

JDBetteridge:
Oneliner


dham (approved):
I won't even wait for the tests, since they don't test this change anyway.

PR #2043, opened 2021-04-27
Wence/fix/mumps fails

wence-:
Catches some places where we were asking mumps to invert either a singular operator or a saddle-point system (and the pivoting was not working well).

### DIFF
--- a/firedrake/projection.py
+++ b/firedrake/projection.py
@@ -115,6 +115,8 @@ class ProjectorBase(object, metaclass=abc.ABCMeta):
             solver_parameters = solver_parameters.copy()
         solver_parameters.setdefault("ksp_type", "cg")
         solver_parameters.setdefault("ksp_rtol", 1e-8)
+        solver_parameters.setdefault("pc_type", "bjacobi")
+        solver_parameters.setdefault("sub_pc_type", "icc")
         self.source = source
         self.target = target
         self.solver_parameters = solver_parameters
--- a/tests/regression/test_manifolds.py
+++ b/tests/regression/test_manifolds.py
@@ -35,12 +35,11 @@ def run_no_manifold():
 
     nullspace = MixedVectorSpaceBasis(V, [V.sub(0), VectorSpaceBasis(constant=True)])
 
-    # Add additional flags to MUMPS as pivoting fails with default options
-    params = {'mat_mumps_icntl_7': 1 if mesh.comm.size == 1 else 3,
-              # Detect null pivots
-              'mat_mumps_icntl_24': 1,
+    params = {'ksp_type': 'gmres',
               'ksp_view': None,
-              'ksp_converged_reason': None}
+              'pc_type': 'svd',
+              'ksp_converged_reason': None,
+              'ksp_monitor': None}
     solve(a == L, up, bcs=bc, nullspace=nullspace, solver_parameters=params)
     exact = Function(V1).interpolate(x[0] - 0.5)
 
@@ -79,12 +78,10 @@ def run_manifold():
 
     nullspace = MixedVectorSpaceBasis(V, [V.sub(0), VectorSpaceBasis(constant=True)])
 
-    # Add additional flags to MUMPS as pivoting fails with default options
-    params = {'mat_mumps_icntl_7': 1 if mesh.comm.size == 1 else 3,
-              # Detect null pivots
-              'mat_mumps_icntl_24': 1,
-              'ksp_view': None,
-              'ksp_converged_reason': None}
+    params = {'ksp_type': 'gmres',
+              'pc_type': 'svd',
+              'ksp_converged_reason': None,
+              'ksp_monitor': None}
     solve(a == L, up, bcs=bc, nullspace=nullspace, solver_parameters=params)
     exact = Function(V1).interpolate(x_n[0] - 0.5)
 
--- a/tests/regression/test_star_pc.py
+++ b/tests/regression/test_star_pc.py
@@ -124,7 +124,7 @@ def test_star_equivalence(problem_type, backend):
         (z, p) = split(u)
         (v, q) = split(TestFunction(V))
 
-        a = inner(grad(z), grad(v))*dx - inner(p, div(v))*dx - inner(div(z), q)*dx
+        a = inner(grad(z), grad(v))*dx + inner(p, q)*dx
 
         bcs = DirichletBC(V.sub(0), Constant((1., 0.)), "on_boundary")
         nsp = MixedVectorSpaceBasis(V, [V.sub(0), VectorSpaceBasis(constant=True)])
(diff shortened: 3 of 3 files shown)


ksagiyam:
I just tested on my local computer, but `test_star_pc` still fails even after making the operator non-singular. `test_manifolds` now looks fine.

wence-:
> I just tested on my local computer, but `test_star_pc` still fails even after making the operator non-singular. `test_manifolds` now looks fine.

It might be that I didn't actually make it non-singular.

ScottMacLachlan:
> > I just tested on my local computer, but `test_star_pc` still fails even after making the operator non-singular. `test_manifolds` now looks fine.
> 
> It might be that I didn't actually make it non-singular.

I think that's the wrong sign...

ksagiyam:
It fails even if I further add `inner(z, v) * dx` to `a`. Actually the following is the MFE using `mat_mumps_icntl_7": 1`:
```
from firedrake import *


mesh = UnitSquareMesh(1, 1)
V = FunctionSpace(mesh, "DG", 0)

u = TrialFunction(V)
v = TestFunction(V)

a = inner(u, v) * dx
L = conj(v) * dx
up = Function(V)

params = {'mat_mumps_icntl_7': 1,
          'mat_mumps_icntl_24': 1,
          'ksp_view': None,
          'ksp_converged_reason': None}
solve(a == L, up, solver_parameters=params)

print(up.dat.data)
```

ScottMacLachlan:
That's a strange failure, Koki - isn't MUMPS failing to factor a 2x2 diagonal matrix in that case?

Here's my big concern: the (original) failing test in `test_star_pc.py` is just a Q2-Q1 Stokes discretization - that's about as simple a saddle-point system as you'd expect to see (albeit with a 1-dimensional singularity).  Shouldn't we know solver parameters to robustly handle this case?

ksagiyam:
Yes, and this error magically disappears if I use 'mat_mumps_icntl_7': 3.

I think we are using good solver parameters in `test_star_pc.py`, but we use `mumps` at the coarsest level, which is causing the error.

wence-:
> That's a strange failure, Koki - isn't MUMPS failing to factor a 2x2 diagonal matrix in that case?

This sounds like a bug one should send upstream, because I can't believe that something isn't broken.

wence-:
The fact that @ksagiyam writes `conj(v)*dx` makes me wonder if the Firedrake is installed in complex-mode, but still...

ScottMacLachlan:
> > That's a strange failure, Koki - isn't MUMPS failing to factor a 2x2 diagonal matrix in that case?
> 
> This sounds like a bug one should send upstream, because I can't believe that something isn't broken.

That seems like a reasonable conclusion to me.

While I'm staring at this, though, is there a reason to try and pin `mat_mumps_icntl_7` manually?  From the MUMPS manual, their default would be a value of 7, which sets off some internal logic about which choice to use.  To my reading, PETSc over-rides that with a default of 0, to use AMD.  Does `'mat_mumps_icntl_7':7` DTRT here?

wence-:
> While I'm staring at this, though, is there a reason to try and pin `mat_mumps_icntl_7` manually?

I have definitely cargo-culted any MUMPS options that appear in tests or elsewhere in Firedrake...

ksagiyam:
I use real mode.

I modified `ksp/tests/ex2.c`, and was able to reproduced the problem. Does this MFE look fine?

```
#include <petscksp.h>
#include <petscvec.h>

int main(int argc,char **args)
{
  Mat            C;
  PetscErrorCode ierr;
  PetscInt       N = 2,rowidx,colidx;
  Vec            u,b;
  KSP            ksp;
  PetscMPIInt    rank,size;
  PetscScalar    v;
  PetscScalar   *a;

  ierr = PetscInitialize(&argc,&args,(char*)0,"");if (ierr) return ierr;
  ierr = MPI_Comm_rank(PETSC_COMM_WORLD,&rank);CHKERRMPI(ierr);
  ierr = MPI_Comm_size(PETSC_COMM_WORLD,&size);CHKERRMPI(ierr);

  /* create stiffness matrix C = [1 0; 0 1] */
  ierr = MatCreate(PETSC_COMM_WORLD,&C);CHKERRQ(ierr);
  ierr = MatSetSizes(C,PETSC_DECIDE,PETSC_DECIDE,N,N);CHKERRQ(ierr);
  ierr = MatSetFromOptions(C);CHKERRQ(ierr);
  ierr = MatSetUp(C);CHKERRQ(ierr);
  if (!rank) {
    rowidx = 0; colidx = 0; v = 1.0;
    ierr   = MatSetValues(C,1,&rowidx,1,&colidx,&v,INSERT_VALUES);CHKERRQ(ierr);

    rowidx = 1; colidx = 1; v = 1.0;
    ierr   = MatSetValues(C,1,&rowidx,1,&colidx,&v,INSERT_VALUES);CHKERRQ(ierr);
  }
  ierr = MatAssemblyBegin(C,MAT_FINAL_ASSEMBLY);CHKERRQ(ierr);
  ierr = MatAssemblyEnd(C,MAT_FINAL_ASSEMBLY);CHKERRQ(ierr);

  /* create right hand side and solution */
  ierr = VecCreate(PETSC_COMM_WORLD,&u);CHKERRQ(ierr);
  ierr = VecSetSizes(u,PETSC_DECIDE,N);CHKERRQ(ierr);
  ierr = VecSetFromOptions(u);CHKERRQ(ierr);
  ierr = VecDuplicate(u,&b);CHKERRQ(ierr);
  ierr = VecSet(u,0.0);CHKERRQ(ierr);
  ierr = VecSet(b,1.0);CHKERRQ(ierr);

  /* check initial values: [0., 0.] */
  ierr = PetscMalloc1(N, &a);CHKERRQ(ierr);
  ierr = VecGetArray(u, &a);CHKERRQ(ierr);
  ierr = PetscPrintf(PETSC_COMM_SELF, "u = [%g, %g]\n", a[0], a[1]);CHKERRQ(ierr);
  ierr = VecRestoreArray(u, &a);CHKERRQ(ierr);

  /* solve linear system C*u = b */
  ierr = KSPCreate(PETSC_COMM_WORLD,&ksp);CHKERRQ(ierr);
  ierr = KSPSetOperators(ksp,C,C);CHKERRQ(ierr);
  ierr = KSPSetFromOptions(ksp);CHKERRQ(ierr);
  ierr = KSPSolve(ksp,b,u);CHKERRQ(ierr);

  /* check computed values: expecting [1., 1.] */
  ierr = VecGetArray(u, &a);CHKERRQ(ierr);
  ierr = PetscPrintf(PETSC_COMM_SELF, "u = [%g, %g]\n", a[0], a[1]);CHKERRQ(ierr);
  ierr = VecRestoreArray(u, &a);CHKERRQ(ierr);
  ierr = PetscFree(a);CHKERRQ(ierr);

  ierr = KSPDestroy(&ksp);CHKERRQ(ierr);
  ierr = VecDestroy(&u);CHKERRQ(ierr);
  ierr = VecDestroy(&b);CHKERRQ(ierr);
  ierr = MatDestroy(&C);CHKERRQ(ierr);
  ierr = PetscFinalize();
  return ierr;
}
```

`mpiexec -n 1 ./ex999 -ksp_type preonly -pc_type lu -pc_factor_mat_solver_type mumps  -mat_mumps_icntl_7 1`
prints `u = [inf., inf.]`

`mpiexec -n 1 ./ex999 -ksp_type preonly -pc_type lu -pc_factor_mat_solver_type mumps  -mat_mumps_icntl_7 3`
prints `u = [1., 1.]`

wence-:
That looks reasonably minimal.

ksagiyam:
Created an issue: https://gitlab.com/petsc/petsc/-/issues/894

sv2518:
@ScottMacLachlan mentioned in Slack that the demo qg_1layer was failing at some point as well. The demo is also failing on our sprint branch here https://jenkins.ese.ic.ac.uk/blue/organizations/jenkins/firedrake/detail/PR-2038/19/pipeline/162. I cannot see MUMPS being used in there, does that mean the demo failure is a separate issue?

ksagiyam:
That demo at least passes on my local computer, where `test_star_pc` fails.

knepley:
Pierre thinks this may not be MUMPS, but instead https://gitlab.com/petsc/petsc/-/merge_requests/3739 .The MUMPS problem seems to be a bad permutation array, which is exactly what this MR Is futzing with.

PR #2047, opened 2021-04-28
Limit Threading

JDBetteridge:
... if possible by default, or warn user otherwise.

### DIFF
--- a/firedrake/__init__.py
+++ b/firedrake/__init__.py
@@ -13,7 +13,7 @@ elif "PETSC_DIR" not in os.environ and config["options"]["honour_petsc_dir"]:
 elif not config["options"]["honour_petsc_dir"]:  # Using our own PETSC.
     os.environ["PETSC_DIR"] = os.path.join(sys.prefix, "src", "petsc")
     os.environ["PETSC_ARCH"] = "default"
-del os, sys, config
+del sys, config
 
 # Ensure petsc is initialised by us before anything else gets in there.
 import firedrake.petsc as petsc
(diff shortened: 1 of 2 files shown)


firedrake/__init__.py
```diff
@@ -100,3 +100,31 @@
...
+
+# OMP_NUM_THREADS can be set to a comma-separated list of positive integers
+try:
+    _omp_num_threads = int(os.environ.get('OMP_NUM_THREADS'))
+except (ValueError, TypeError):
+    _omp_num_threads = None
+if (_omp_num_threads is None) or (_omp_num_threads > 1):
+    warning('OMP_NUM_THREADS is not set or is set to a value greater than 1,'
+            ' we suggest setting OMP_NUM_THREADS=1 to improve performance')
+del _openblas_lib, _openblas_dll, _omp_num_threads, os, cdll, find_library
```
wence-, line 130: Wrap in functions?

firedrake/__init__.py
```diff
@@ -100,3 +100,31 @@
...
+# - MKL_NUM_THREADS: mkl,
+# - VECLIB_MAXIMUM_THREADS: accelerate,
+# - NUMEXPR_NUM_THREADS: numexpr
+# We only handle the first two cases
+from ctypes import cdll
+from ctypes.util import find_library
+try:
+    _openblas_lib = find_library('openblas')
+    _openblas_dll = cdll.LoadLibrary(_openblas_lib)
+    _openblas_dll.openblas_set_num_threads(1)
```
wence-, line 117: Does this function return an error code we can check?

scripts/firedrake-install
```diff
@@ -604,6 +604,7 @@ else:
 if options.get("with_blas") == "download":
     # Download openblas if the user specifies.
     petsc_options.add("--download-openblas")
+    petsc_options.add("--download-openblas-make-options='USE_THREAD=0 USE_LOCKING=1 USE_OPENMP=0'")
     if osname == "Darwin":
         petsc_options.add("--CFLAGS=-Wno-implicit-function-declaration")
 elif options.get("with_blas") is not None:
```
wence-, line 610: While we're here, can we flip the `openblas` packages on debian to the `seq` versions? That's in the docker container (just pull in my #1905).

JDBetteridge: Merged #1905 into this branch (and closed that PR)

firedrake/__init__.py
JDBetteridge: Sadly not:
https://github.com/xianyi/OpenBLAS/blob/develop/driver/others/openblas_set_num_threads.c

firedrake/__init__.py
JDBetteridge: Not sure if that gains us anything, we don't actually want users calling these functions. Will change if turning this into functions is the preferred style, otherwise I'll leave it

firedrake/__init__.py
wence-: Of course not, because it just calls OpenMP and https://www.openmp.org/spec-html/5.0/openmpsu110.html#x147-6380003.2.1

firedrake/__init__.py
wence-: I just thought you could make functions and `del function_name` afterwards (don't need to delete so many dels), minor nitpick though.

JDBetteridge: OIC, True, more lines of code though! :laughing:

Issue #2045, opened 2021-04-28
Firedrake install problem


sv2518:
Hi. I have tried to install firedrake complex with clang version 12.0.0 and Python version 3.9.4. I'm on macOS Catalina 10.15.7. I have brew updated to 3.1.4, brew uninstalled and installed python3, brew unlinked mpich, brew unlinked open-mpi. 

Here is the install log 
[firedrake-install.log](https://github.com/firedrakeproject/firedrake/files/6390216/firedrake-install.log)
 And the petsc log
[configure.log](https://github.com/firedrakeproject/firedrake/files/6390220/configure.log)

It looks like it fails on make install on MPICH. Does anyone have an idea how I can get this to work?

JDBetteridge:
Does the issue go away if you 
```bash
export PETSC_CONFIGURE_OPTIONS="--with-x=0"
```
before installing?

sv2518:
Ok done, still failing, but with a new error

[configure.log](https://github.com/firedrakeproject/firedrake/files/6391672/configure.log)
[firedrake-install.log](https://github.com/firedrakeproject/firedrake/files/6391673/firedrake-install.log)

dham:
I think I don't understand that configure error

wence-:
Can we strip this back to just getting `./configure --download-mpich` to fail inside the petsc source tree? Then we could upstream the bug :)

dham:
It transpires that @sv2518 ran out of space. For me, @JDBetteridge's fix looks like it might have worked. I'm not sure if that makes it an upstream bug or not.

JDBetteridge:
I think this is working as intended, if PETSc configure finds an X11 it will try to use it and passes it to all external packages that want it. I don't really understand enough about X11 on a Mac to know what should happen in this case.

Perhaps we should add `--with-x=0` to the PETSc config options if we're on a Mac if this is going to become a common issue?

dham:
Do we need x from PETSc? If not we should switch it off for all platforms.

wence-:
Jack's comment suggests that petsc does find X but doesn't pass it correctly to mpich?

Having X available in PETSc is kind of nice because you can draw matrices with it (I like it, that might be niche).

dham:
OK. Let's just switch it off for Mac.

sv2518:
Ja, Jack's fix worked for me as well. Thanks!

PR #2048, opened 2021-04-29
Disable x11 for petsc installs.

sv2518:
Addresses the issue #2045

### DIFF
--- a/scripts/firedrake-install
+++ b/scripts/firedrake-install
@@ -555,6 +555,9 @@ petsc_options = {"--with-fortran-bindings=0",
                  # For superlu_dist amongst others.
                  "--with-cxx-dialect=C++11"}
 
+if osname == "Darwin":
+    petsc_options.add("--with-x=0")
+
 if not options["minimal_petsc"]:
     petsc_options.add("--with-zlib")
     # File formats


scripts/firedrake-install
```diff
@@ -553,7 +553,8 @@ petsc_options = {"--with-fortran-bindings=0",
                  # Parallel mesh partitioner
                  "--download-ptscotch",
                  # For superlu_dist amongst others.
-                 "--with-cxx-dialect=C++11"}
+                 "--with-cxx-dialect=C++11",
+                 "--with-x=0"}
```
JDBetteridge, line 558: Could we add this as a Mac only flag?
```python
if osname == "Darwin":
    petsc_options.add("--with-x=0")
```
The install doesn't currently fail on Linux

JDBetteridge (changes requested):

scripts/firedrake-install
sv2518: Ah, ok. Done.

JDBetteridge (approved):
Thumbs up from me! :+1: 
Failing tests are the ones failing on master
```
[2021-04-30T12:18:45.051Z] FAILED tests/regression/test_star_pc.py::test_star_equivalence[mixed-petscasm]

[2021-04-30T12:18:45.051Z] FAILED tests/regression/test_star_pc.py::test_star_equivalence[mixed-tinyasm]
```

PR #2050, opened 2021-04-29
firedrake-update installs petsc4py if not yet installed

ksagiyam:
This PR is to let `firedrake-update` install `petsc4py` if not yet installed.

### DIFF
--- a/scripts/firedrake-install
+++ b/scripts/firedrake-install
@@ -15,6 +15,7 @@ import shlex
 from glob import iglob
 from itertools import chain
 import re
+import importlib
 
 osname = platform.uname().system
 arch = platform.uname().machine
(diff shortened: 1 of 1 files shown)


ReubenHill:
Has something changed that necessitates this? I thought petsc4py was part of PETSc now?

ksagiyam:
Firedrake still installs `petsc` and `petsc4py` separately. If `firedrake-install` fails somewhere after successfully installing `petsc`, `firedrake-update` can continue installation bypassing the expensive `petsc` installation, but currently this process fails to install `petsc4py`, as we assume that we already have `petsc4py` when we call `firedrake-update`. This PR attempts to fix this.

connorjward:
Could this be made clear in a comment?

ksagiyam:
Thanks. Added comments.

PR #2051, opened 2021-04-30
Annotate functions with PETSc events

connorjward:
I've added `@PETSc.Log.EventDecorator()` decorators to lots of functions. The idea here is to combine this with the flame graph output that PETSc now supports so we can get some nice, low effort profiling information for Firedrake users.

I've been somewhat arbitrary with the functions that I've labelled. What I intend to do is run all of the demos and look for any big gaps as this will indicate a section where we need extra annotations.

~But. Before any of this gets merged I've identified some [major bugs](https://gitlab.com/petsc/petsc/-/merge_requests/3937) in the flame graph stuff I did in PETSc that should really get fixed beforehand.~
Edit: This is now fixed.

This should ideally get merged with https://github.com/OP2/PyOP2/pull/623.

### DIFF
--- a/firedrake/bcs.py
+++ b/firedrake/bcs.py
@@ -9,7 +9,6 @@ from ufl import as_ufl, UFLException, as_tensor, VectorElement
 import finat
 
 import pyop2 as op2
-from pyop2.profiling import timed_function
 from pyop2 import exceptions
 from pyop2.utils import as_tuple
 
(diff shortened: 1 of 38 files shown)


firedrake/bcs.py
```diff
@@ -42,6 +42,7 @@ class BCBase(object):
         vanish on the boundary will be included. This can be used to impose
         strong boundary conditions on DG spaces, or no-slip conditions on HDiv spaces.
     '''
+    @PETSc.Log.EventDecorator()
```
wence-, line 45: For class methods, what name does the `EventDecorator` produce?

wence-:
Great, thanks.

PR #2053, opened 2021-05-03
Sharing DM means sharing residual function.

salazardetroya:
The `DM` is shared between all solvers and so is its `DMSNES` object. This means that the residual function `DMSNESSetFunction` is also shared between all solvers. For the case of several `NonlinearVariationalSolver` it is not a problem because the only difference between solvers is encapsulated in the context `appctx`. It is a problem when using a `TS` that also shares the same `DM`, as it will steal the `DMSNESSetFunction` and set up its own. When an independent and previously setup `NonlinearVariationalSolver` wants to solve again, it finds that its residual function is no longer there, but `TS`'s is.

Let me know if it is a good idea to include a test to show the issue. It might be a bit long because of the `TS`.

### DIFF
--- a/firedrake/variational_solver.py
+++ b/firedrake/variational_solver.py
@@ -250,6 +250,10 @@ class NonlinearVariationalSolver(OptionsManager, NonlinearVariationalSolverMixin
            If bounds are provided the ``snes_type`` must be set to
            ``vinewtonssls`` or ``vinewtonrsls``.
         """
+        # Make sure the DM has this solver's callback functions
+        self._ctx.set_function(self.snes)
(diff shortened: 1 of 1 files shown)


wence-:
Thanks, this looks correct to me. I restarted the tests because I had broken something different in TSFC (that is now fixed).

wence- (dismissed):

wence-:
Can you please fix the linting errors: https://github.com/firedrakeproject/firedrake/pull/2053/checks?check_run_id=2507979270

wence-:
Thanks!

PR #2054, opened 2021-05-04
Deprecate create_assembly_callable

connorjward:
I've added a deprecation warning to `create_assembly_callable` and replaced all occurrences with `functools.partial(assemble, ...)`.

### DIFF
--- a/firedrake/assemble.py
+++ b/firedrake/assemble.py
@@ -174,8 +174,6 @@ def allocate_matrix(expr, bcs=(), form_compiler_parameters=None,
                     options_prefix=None):
     r"""Allocate a matrix given an expression.
 
-    To be used with :func:`create_assembly_callable`.
-
     .. warning::
 
        Do not use this function unless you know what you're doing.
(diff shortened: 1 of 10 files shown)


firedrake/assemble.py
```diff
@@ -198,10 +196,20 @@ def create_assembly_callable(expr, tensor=None, bcs=None, form_compiler_paramete
...
+
     .. warning::
 
        Really do not use this function unless you know what you're doing.
     """
+    import warnings
+    with warnings.catch_warnings():
+        warnings.simplefilter("always", DeprecationWarning)
+        warnings.warn("create_assembly_callable is now deprecated. Please use assemble instead.",
+                      DeprecationWarning)
```
wence-, line 211: Only print once?

firedrake/linear_solver.py
```diff
@@ -108,11 +110,12 @@ def trial_space(self):
...
     @cached_property
     def _rhs(self):
-        from firedrake.assemble import create_assembly_callable
+        from firedrake.assemble import assemble
+
         u = function.Function(self.trial_space)
         b = function.Function(self.test_space)
         expr = -action(self.A.a, u)
-        return u, create_assembly_callable(expr, tensor=b), b
+        return u, functools.partial(assemble, expr, tensor=b, assembly_type="residual"), b
```
wence-, line 118: strings as enums makes me sad. But I guess that's for a separate cleanup.

PR #2057, opened 2021-05-06
Add FEMlium to firedrake apps

francesco-ballarin:
Hi,

@dham on slack suggested to add [FEMlium](https://femlium.github.io/) among the apps installable with `firedrake-install --install femlium`.

I have tested myself this change running `firedrake-install` on a `firedrakeproject/firedrake-env` docker image.

Do you want me to change `Jenkinsfile` and `tests/regression/test_import_applications.py` too, in a similar way to what was done in #1751?

### DIFF
--- a/scripts/firedrake-install
+++ b/scripts/firedrake-install
@@ -35,6 +35,8 @@ firedrake_apps = {
                 "git+ssh://github.com/icepack/icepack.git#egg=icepack"),
     "irksome": ("""Implicit Runge-Kutta methods. https://github.com/firedrakeproject/Irksome/""",
                 "git+ssh://github.com/firedrakeproject/Irksome.git#egg=Irksome"),
+    "femlium": ("""Interactive visualization of finite element simulations on geographic maps with folium. https://femlium.github.io/""",
+                "git+ssh://github.com/FEMlium/FEMlium.git@main#egg=FEMlium"),
 }
 
 
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -40,7 +40,7 @@ pipeline {
               sh 'mkdir tmp'
               dir('tmp') {
                 timestamps {
-                  sh '../scripts/firedrake-install $COMPLEX --tinyasm --disable-ssh --minimal-petsc --slepc --documentation-dependencies --install thetis --install gusto --install icepack --install irksome --no-package-manager || (cat firedrake-install.log && /bin/false)'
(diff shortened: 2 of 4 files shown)


scripts/firedrake-install
```diff
@@ -35,6 +35,9 @@ firedrake_apps = {
                 "git+ssh://github.com/icepack/icepack.git#egg=icepack"),
     "irksome": ("""Implicit Runge-Kutta methods. https://github.com/firedrakeproject/Irksome/""",
                 "git+ssh://github.com/firedrakeproject/Irksome.git#egg=Irksome"),
+    "femlium": ("""Interactive visualization of finite element simulations on geographic maps with folium.
+                   https://github.com/FEMlium/FEMlium/""",
```
wence-, line 39: I think that this will result in a big space in the `--help` output, can you check? Also, since you have this lovely documentation, perhaps point at femlium.github.io rather than the source page?

wence-:
> Do you want me to change Jenkinsfile and tests/regression/test_import_applications.py

Yes please.

scripts/firedrake-install
francesco-ballarin: Thanks, I fixed that. I did that because I thought there were pep8 checks on the line length.

francesco-ballarin:
> Do you want me to change Jenkinsfile and tests/regression/test_import_applications.py

I have done that now.

scripts/firedrake-install
wence-: Ah, we are a bit loose with line length restrictions. For future reference if you want one long string but to break it over multiple lines, then the python parse concatenates successive strings so you can write:

```python
string = ("abc "
          "def")
assert string == "abc def"
```

francesco-ballarin:
I am not sure if I changed the right file for jenkins. I did change it, but in the pipeline I still see
`../scripts/firedrake-install $COMPLEX --tinyasm --disable-ssh --minimal-petsc --slepc --documentation-dependencies --install thetis --install gusto --install icepack --install irksome --no-package-manager || (cat firedrake-install.log && /bin/false)`


Furthermore, should a similar `--install` option be added in `docker/Dockerfile.firedrake`? (irksome seems to be missing there, actually)

wence-:
> I did change it, but in the pipeline I still see

I think this is because we don't trust changes to the Jenkinsfile from a fork. I will check this works locally and then merge.

wence-:
Thanks very much!

PR #2058, opened 2021-05-07
WIP: Make Firedrake's configuration more forgiving

connorjward:
@JDBetteridge please review. I'll leave comments where things aren't clear/need a decision on.


firedrake/__init__.py
```diff
@@ -1,19 +1,4 @@
...
-       or os.environ["PETSC_ARCH"] != "default":
-        raise ImportError("PETSC_DIR is set, but you did not install with --honour-petsc-dir.\n"
-                          "Please unset PETSC_DIR (and PETSC_ARCH) before using Firedrake.")
-elif "PETSC_DIR" not in os.environ and config["options"]["honour_petsc_dir"]:
-    raise ImportError("Firedrake was installed with --honour-petsc-dir, but PETSC_DIR is not set.\n"
-                      "Please set PETSC_DIR (and PETSC_ARCH) before using Firedrake.")
-elif not config["options"]["honour_petsc_dir"]:  # Using our own PETSC.
-    os.environ["PETSC_DIR"] = os.path.join(sys.prefix, "src", "petsc")
-    os.environ["PETSC_ARCH"] = "default"
-del os, sys, config
```
connorjward, line 16: We don't use these variables anywhere. Also they only matter when we build petsc4py. At runtime this stuff is fixed.

firedrake/paraview_reordering.py
```diff
@@ -17,27 +17,30 @@
...
-mod = loader.load_module(moduleName)
-
-vtkLagrangeTetra = mod.vtkLagrangeTetra
-vtkLagrangeHexahedron = mod.vtkLagrangeHexahedron
-vtkLagrangeTriangle = mod.vtkLagrangeTriangle
-vtkLagrangeQuadrilateral = mod.vtkLagrangeQuadrilateral
-vtkLagrangeWedge = mod.vtkLagrangeWedge
-
-paraviewUsesVTK8 = True
+try:
```
connorjward, line 20: Installing VTK in Spack is a pain. I can't get it to work without installing LLVM which takes a VERY long time. Maybe we could have our own minimal package? Either that or make it a Spack `variant`.

Obviously what I've done here is nonsense.

firedrake/slate/slac/compiler.py
```diff
@@ -59,13 +59,12 @@
 BLASLAPACK_INCLUDE = None
 if not complex_mode:
     if COMM_WORLD.rank == 0:
-        petsc_variables = get_petsc_variables()
-        EIGEN_INCLUDE_DIR = petsc_variables.get("EIGEN_INCLUDE")
+        EIGEN_INCLUDE_DIR = get_eigen_include_dir()
```
connorjward, line 62: This is needed because Eigen doesn't have to be linked to PETSc during install so this variable doesn't actually show up in `petscvariables`. In the Spack install I put this variable in `config["libraries"]["EIGEN_INCLUDE_DIR"]`.

setup.py
```diff
@@ -22,6 +22,10 @@
 
 
 def get_petsc_dir():
+    config = petsc4py.get_config()
+    petsc_dir = config["PETSC_DIR"]
+    petsc_arch = config["PETSC_ARCH"]
+    return (petsc_dir, path.join(petsc_dir, petsc_arch))
```
connorjward, line 28: This function is a bit redundant because `PETSC_DIR` and `PETSC_ARCH` are stored by petsc4py.

firedrake/paraview_reordering.py
wence-: report a bug to spack, this seems like a problem at their end.

firedrake/__init__.py
connorjward: I see. I'll find a more graceful way to get Firedrake import working without removing this then.

wence-: Sometimes `git annotate` provides useful context, in this case the relevant commit was 
fdccc0f7622568aac7391037cc32c88e34556ec8

setup.py
```diff
@@ -70,8 +74,8 @@ def get_petsc_dir():
                         libraries=libs,
                         extra_link_args=link_args,
                         cython_compile_time_env=cython_compile_time_env) for (ext, libs) in cythonfiles]
-if 'CC' not in env:
-    env['CC'] = "mpicc"
+# if 'CC' not in env:
```
JDBetteridge, line 77: Probably just remove these lines, rather than comment out

connorjward:
Closing as replaced by #2092

PR #1815, opened 2020-08-21
Interpolation onto a Vertex Only Mesh (old)

ReubenHill:
Draft for changes needed to interpolate onto a Vertex Only Mesh from a suitable parent mesh (to resolve #1795).

Based on top of branch `dual_base_fiat` - see draft PR #1743. Base should change to master when that PR is complete (or if this diverges and stops being dependent on #1743).

Linked PRs: https://github.com/firedrakeproject/tsfc/pull/226 & https://github.com/FInAT/FInAT/pull/66

To-do:

- [ ] Merge PRs this sits on top of (FInAT dual evalutation)
- [ ] Test parallel
- [ ] Test mixed spaces
- [ ] Fix assembly across different meshes for this case (for `E` on a `VertexOnlyMesh` and `R` on its parent mesh, `J = firedrake.assemble(E + R)` does not work but `J = firedrake.assemble(E) + firedrake.assemble(R)` does)
- [ ] Fix `firedrake_adjoint.minimize` here for Newton-CG (need hessian for Interpolation matrix I think) - see #1942

PR #2059, opened 2021-05-07
Interpolation onto a Vertex Only Mesh

### LINKED ISSUE #1795: Interpolate onto VertexOnlyMesh
To begin making VertexOnlyMeshes introduced in #1638 useful we need to be able to interpolate onto them from their parent mesh. This may eventually replace `function.at` for functions on the parent mesh, allow a new trans-mesh-trans-function-space interpolation form (currently codenamed the "delta" form) to be added to UFL, and more.

### LINKED ISSUE #1796: Remove initialisation callback from VertexOnlyMeshTopology
A `VertexOnlyMeshTopology` has a two stage initialisation routine inherited from `MeshTopology`. This has been removed from the `VertexOnlyMesh` creation function but not `VertexOnlyMeshTopology`

> I think this callback dance is actually unnecessary. The reason for the callback stuff in `Mesh` is to do with not wanting to grow the overlap before multigrid refinement, but here we have none of this. So I think that you can just inline all of this stuff into the initialisation routine.

_Originally posted by @wence- in https://github.com/firedrakeproject/firedrake/pull/1638#discussion_r452918156_

ReubenHill:
Adds changes needed to allow interpolation from an arbitrary function function space onto a P0DG function space on a Firedrake `VertexOnlyMesh`

This replaces #1815 and sits on top of master, avoiding dependency on #1743 (dual evaluation with FInAT which needs lots of work).

The features added here will eventually be absorbed into #1743 when that is complete.

Linked PRs
 - https://github.com/firedrakeproject/tsfc/pull/247
 - https://github.com/FInAT/FInAT/pull/81

Generated Issues:
 -  #2069 
 - #2095 

Fixes #1795 

Status:

 * Parent mesh cell number:
   * Added functionality to find and store the parent mesh cell for each vertex cell of the Vertex Only Mesh (saved in swarm field `"parentcellnum"`)
   * Added `VertexOnlyMeshTopology.cell_parent_cell_list` and `VertexOnlyMeshTopology.cell_parent_cell_map` properties containing the parent cell number information
 * Vertex reference coordinate location in parent mesh cells:
   * Added functionality to find and store the reference cell locations of each Vertex Only Mesh vertex cell in their parent mesh cells (saved in swarm field `"refcoord"`)
   * Saved these values in cell order in a `mesh.reference_coordinates` function in a P0DG vector function space on the Vertex Only Mesh.
 * Resolved #1796 by removing unnecessary callbacks in Vertex Only Mesh creation
 * Create interpolation kernel from the parent mesh onto the Vertex Only Mesh by
   1. Modifying the FInAT `QuadratureElement` to allow a custom rule with runtime tabulated point to use as the `to_element` for going from the parent mesh function space to P0DG on the `VertexOnlyMesh` (see https://github.com/FInAT/FInAT/pull/81). 
   2. Create a suitable custom FInAT `QuadratureElement` for a runtime point tabulated point (expression `gem.Variable('X', (tdim,))`) with weight 1 to represent the runtime tabulated point, on the expression's mesh reference cell, onto which we interpolate in interpolation code in `firedrake/interpolate.py`.
   3. Switch the interpolation domain and domain coordinates coords to interpolated expression's mesh and mesh coordinates for feeding into `compile_expression_dual_evaluation`
   4. Modify TSFC's `compile_expression_dual_evaluation` so that the dual evaluation of this element will create create a kernel for interpolating from the parent mesh reference cell at an unknown point onto the vertex only mesh (see https://github.com/firedrakeproject/tsfc/pull/247). This has a coefficient defined in P0DG on the `VertexOnlyMesh` as the input argument for "unknown point" - the point location is defined on the reference cell of the parent mesh.
 * Create outer loop to run interpolation kernel over
   * Use the P0DG `FunctionSpace` on the `VertexOnlyMesh`'s `cell_set` as the `iterset` for iterating over.
   * Create a double indirection PyOP2 `Map` to get from `VertexOnlyMesh` cells to their parent cell function space nodes as defined on the parent mesh reference cell - a so-called `cell_parent_cell_fs_node_map`
      * The `Map` is made with a `composed_map` function using the `cell_node_map` of the expression we are interpolating and the `cell_parent_cell_map` of the Vertex Only Mesh: `composed_map(mesh.cell_parent_cell_map, cell_node_map)`
   * Use the double indirection map to register the "unknown point" coefficient as a par loop argument and appropriate space of `op2.Sparsity` when making an `Interpolator`

To do:
 - [x] Fix failing test if it's my fault
 - [x] Test mixed spaces
 - [x] Consider (again?) if a the FInAT `QuadratureElement` is really a suitable basis for doing this instead of some dedicated runtime tabulation FInAT element
 - [x] stash the composed map on cell_parent_cell_map using caching
 - [x] Work out what I was doing in the `get_composed_map` if statement in `interpolate.py` and maybe replace with actually getting the `composed_map`? It looks like a create it again
    - [x] ~move some of the composed map creation into cython? See code~ Not worth it for a reshape
 - [x] Check through above description carefully to make sure it's accurate
    - [x] In particular, explain what goes on with the use of `cell_parent_cell_fs_node_map` in the creation of an `op2.Sparsity` when you make an `Interpolator` object.
 - [x] Squash
 - [x] Check and make issue about broken real interpolation onto Quadrature elements - done see #2093
 - [x] Make issue about not being able to make an interpolator for Real onto V.O.M.s - done see #2095

### DIFF
--- a/firedrake/cython/dmcommon.pyx
+++ b/firedrake/cython/dmcommon.pyx
@@ -2637,33 +2637,39 @@ def remove_ghosts_pic(PETSc.DM swarm, PETSc.DM plex):
 
 @cython.boundscheck(False)
 @cython.wraparound(False)
-def label_pic_parent_cell_nums(PETSc.DM swarm, parentmesh):
+def label_pic_parent_cell_info(PETSc.DM swarm, parentmesh):
     """
     For each PIC in the input swarm, label its `parentcellnum` field with
-    the relevant cell number from the `parentmesh` in which is it emersed.
-    The cell numbering is that given by the `parentmesh.locate_cell`
-    method.
+    the relevant cell number from the `parentmesh` in which is it emersed
+    and the `refcoord` field with the relevant cell reference coordinate.
+    This information is given by the
+    `parentmesh.locate_cell_and_reference_coordinate` method.
+
+    For a swarm with N PICs emersed in `parentmesh`
+    the `parentcellnum` field is N long and
+    the `refcoord` field is N*parentmesh.topological_dimension() long.
 
     :arg swarm: The DMSWARM which contains the PICs immersed in
         `parentmesh`
     :arg parentmesh: The mesh within with the `swarm` PICs are immersed.
 
     ..note:: All PICs must be within the parentmesh or this will try to
-             assign `None` (returned by `parentmesh.locate_cell`) to a
-             `PetscReal`.
+             assign `None` (returned by
+             `parentmesh.locate_cell_and_reference_coordinate`) to the
+             `parentcellnum` or `refcoord` fields.
+
+
     """
     cdef:
-        PetscInt num_vertices, i, dim, parent_cell_num
+        PetscInt num_vertices, i, gdim, tdim
+        PetscInt parent_cell_num
         np.ndarray[PetscReal, ndim=2, mode="c"] swarm_coords
         np.ndarray[PetscInt, ndim=1, mode="c"] parent_cell_nums
+        np.ndarray[PetscReal, ndim=2, mode="c"] reference_coords
+        np.ndarray[PetscReal, ndim=1, mode="c"] reference_coord
 
-    if type(swarm) is not PETSc.DMSwarm:
-        raise ValueError("swarm must be a DMSwarm")
-
-    if parentmesh.topology_dm.handle != swarm.getCellDM().handle:
-        raise ValueError("parentmesh.topology_dm is not the swarm's CellDM")
-
-    dim = parentmesh.geometric_dimension()
+    gdim = parentmesh.geometric_dimension()
+    tdim = parentmesh.topological_dimension()
 
     num_vertices = swarm.getLocalSize()
 
@@ -2673,21 +2679,33 @@ def label_pic_parent_cell_nums(PETSc.DM swarm, parentmesh):
     max_num_vertices = comm.allreduce(num_vertices, op=MPI.MAX)
 
     # Create an out of mesh point to use in locate_cell when needed
-    out_of_mesh_point = np.full((1, dim), np.inf)
+    out_of_mesh_point = np.full((1, gdim), np.inf)
 
     # get fields - NOTE this isn't copied so make sure
     # swarm.restoreField is called for each field too!
-    swarm_coords = swarm.getField("DMSwarmPIC_coor").reshape((num_vertices, dim))
+    swarm_coords = swarm.getField("DMSwarmPIC_coor").reshape((num_vertices, gdim))
     parent_cell_nums = swarm.getField("parentcellnum")
+    reference_coords = swarm.getField("refcoord").reshape((num_vertices, tdim))
 
     # find parent cell numbers
+    # TODO We should be able to do this for all the points in in one call to
+    # the parent mesh's _c_locator.
+    # SUGGESTED API 1:
+    #   parent_cell_nums, reference_coords = parentmesh.locate_cell_and_reference_coordinates(swarm_coords)
+    # with second call for collectivity.
+    # SUGGESTED API 2:
+    #   parent_cell_nums, reference_coords = parentmesh.locate_cell_and_reference_coordinates(swarm_coords, local_num_vertices)
+    # with behaviour changing inside locate_cell_and_reference_coordinates to
+    # ensure collectivity.
     for i in range(max_num_vertices):
         if i < num_vertices:
-            parent_cell_num = parentmesh.locate_cell(swarm_coords[i])
+            parent_cell_num, reference_coord = parentmesh.locate_cell_and_reference_coordinate(swarm_coords[i])
             parent_cell_nums[i] = parent_cell_num
+            reference_coords[i] = reference_coord
         else:
             parentmesh.locate_cell(out_of_mesh_point)  # should return None
 
     # have to restore fields once accessed to allow access again
+    swarm.restoreField("refcoord")
     swarm.restoreField("parentcellnum")
     swarm.restoreField("DMSwarmPIC_coor")
--- a/firedrake/cython/dmcommon.pyx
+++ b/firedrake/cython/dmcommon.pyx
@@ -2709,3 +2709,46 @@ def label_pic_parent_cell_info(PETSc.DM swarm, parentmesh):
     swarm.restoreField("refcoord")
     swarm.restoreField("parentcellnum")
     swarm.restoreField("DMSwarmPIC_coor")
+
+
+@cython.boundscheck(False)
+@cython.wraparound(False)
+def fill_reference_coordinates_function(reference_coordinates_f):
+    """
+    Fill the PyOP2 dat of an input vector valued function on a
+    VertexOnlyMesh `reference_coordinates_f` with the reference
+    coordinates of each vertex in their relevant reference cells.
+
+    :arg reference_coordinates_f: A vector valued function on a
+        VertexOnlyMesh (with vector dimension the topological dimension
+        of the parent mesh) which will have its dat modified.
+
+    :returns: The updated `reference_coordinates_f`.
+    """
+    cdef:
+        PetscInt num_vertices, i, gdim, parent_tdim
+        PETSc.DM swarm
+        np.ndarray[PetscReal, ndim=2, mode="c"] reference_coords
+
+    from firedrake.mesh import VertexOnlyMeshTopology
+    assert isinstance(reference_coordinates_f.function_space().mesh().topology, VertexOnlyMeshTopology)
+
+    gdim = reference_coordinates_f.function_space().mesh()._parent_mesh.geometric_dimension()
+    parent_tdim = reference_coordinates_f.function_space().mesh()._parent_mesh.topological_dimension()
+
+    swarm = reference_coordinates_f.function_space().mesh().topology_dm
+
+    num_vertices = swarm.getLocalSize()
+
+    assert reference_coordinates_f.dat.shape == (num_vertices, parent_tdim)
+
+    # get reference coord field - NOTE isn't copied so could have GC issues!
+    reference_coords = swarm.getField("refcoord").reshape((num_vertices, parent_tdim))
+
+    # store reference coord field in Function Dat.
+    reference_coordinates_f.dat.data[:] = reference_coords[:]
+
+    # have to restore fields once accessed to allow access again
+    swarm.restoreField("refcoord")
+
+    return reference_coordinates_f
--- a/firedrake/interpolation.py
+++ b/firedrake/interpolation.py
@@ -170,19 +170,20 @@ def make_interpolator(expr, V, subset, access):
         if isinstance(V, firedrake.Function):
             raise ValueError("Cannot interpolate an expression with an argument into a Function")
         argfs = arguments[0].function_space()
+        target_mesh = V.ufl_domain()
+        source_mesh = argfs.mesh()
         argfs_map = argfs.cell_node_map()
-        trans_mesh = (
-            argfs_map is not None
-            and argfs_map.iterset != V.cell_node_map().iterset
-        )
-        if trans_mesh:
-            if V.ufl_domain().geometric_dimension() != argfs.mesh().geometric_dimension():
-                raise ValueError("Cannot interpolate onto a mesh of a different geometric dimension")
-            if isinstance(V.ufl_domain().topology, firedrake.mesh.VertexOnlyMeshTopology):
-                # Compose a vertex-cell to parent-cell-function-space-node map
-                argfs_map = compose_map_and_cache(V.ufl_domain().cell_parent_cell_map, argfs_map)
-            else:
+        if target_mesh is not source_mesh:
+            if not isinstance(target_mesh.topology, firedrake.mesh.VertexOnlyMeshTopology):
                 raise NotImplementedError("Can only interpolate onto a Vertex Only Mesh")
+            if target_mesh.geometric_dimension() != source_mesh.geometric_dimension():
+                raise ValueError("Cannot interpolate onto a mesh of a different geometric dimension")
+            if not hasattr(target_mesh, "_parent_mesh") or target_mesh._parent_mesh is not source_mesh:
+                raise ValueError("Can only interpolate across meshes where the source mesh is the parent of the target")
+            # Since the par_loop is over the target mesh cells we need to
+            # compose a map that takes us from target mesh cells to the
+            # function space nodes on the source mesh.
+            argfs_map = compose_map_and_cache(target_mesh.cell_parent_cell_map, argfs_map)
         sparsity = op2.Sparsity((V.dof_dset, argfs.dof_dset),
                                 ((V.cell_node_map(), argfs_map),),
                                 name="%s_%s_sparsity" % (V.name, argfs.name),
@@ -242,27 +243,20 @@ def _interpolator(V, tensor, expr, subset, arguments, access):
         raise RuntimeError('Shape mismatch: Expression shape %r, FunctionSpace shape %r'
                            % (expr.ufl_shape, V.ufl_element().value_shape()))
 
+    # NOTE: The par_loop is always over the target mesh cells.
     target_mesh = V.ufl_domain()
-    try:
-        trans_mesh = (
-            expr.ufl_domain() is not None  # Constant expr
-            and expr.ufl_domain() != V.mesh()  # Coming from a different domain
-        )
-    except AttributeError:
-        # Have python Expression
-        trans_mesh = False
+    if isinstance(expr, firedrake.Expression):
+        source_mesh = target_mesh
+    else:
+        source_mesh = expr.ufl_domain() or target_mesh
 
-    if trans_mesh:
-        if target_mesh.geometric_dimension() != expr.ufl_domain().geometric_dimension():
-            raise ValueError("Cannot interpolate onto a mesh of a different geometric dimension")
-        # In the trans-mesh case we loop over the target mesh cells so we need
-        # to compose a map that takes us from their cells to the source mesh
-        # cells, and (where necessary) to the function space nodes on the
-        # source mesh.
-        if isinstance(target_mesh.topology, firedrake.mesh.VertexOnlyMeshTopology):
-            target_mesh_source_mesh_map = target_mesh.cell_parent_cell_map
-        else:
+    if target_mesh is not source_mesh:
+        if not isinstance(target_mesh.topology, firedrake.mesh.VertexOnlyMeshTopology):
             raise NotImplementedError("Can only interpolate onto a Vertex Only Mesh")
+        if target_mesh.geometric_dimension() != source_mesh.geometric_dimension():
+            raise ValueError("Cannot interpolate onto a mesh of a different geometric dimension")
+        if not hasattr(target_mesh, "_parent_mesh") or target_mesh._parent_mesh is not source_mesh:
+            raise ValueError("Can only interpolate across meshes where the source mesh is the parent of the target")
         # For trans-mesh interpolation we use a FInAT QuadratureElement as the
         # (base) target element with runtime point set expressions as their
         # quadrature rule point set and weights from their dual basis.
@@ -277,13 +271,6 @@ def _interpolator(V, tensor, expr, subset, arguments, access):
         # equivalent(s) (i.e. finat.point_set.UnknownPointSet(s))
         rt_var_name = 'rt_X'
         to_element = rebuild(to_element, expr, rt_var_name)
-        # The source mesh is on the expression not the FunctionSpace/Function
-        # we are interpolating into/onto
-        source_mesh = expr.ufl_domain()
-    else:
-        # The source mesh is that of the FunctionSpace/Function we are
-        # interpolating into/onto
-        source_mesh = target_mesh
 
     parameters = {}
     parameters['scalar_type'] = utils.ScalarType
@@ -322,7 +309,7 @@ def _interpolator(V, tensor, expr, subset, arguments, access):
         # Replace with real source mesh coordinates
         coefficients[0] = source_mesh.coordinates
 
-    if trans_mesh:
+    if target_mesh is not source_mesh:
         # NOTE: TSFC will sometimes drop run-time arguments in generated
         # kernels if they are deemed not-necessary.
         # FIXME: Checking for argument name in the inner kernel to decide
@@ -340,7 +327,7 @@ def _interpolator(V, tensor, expr, subset, arguments, access):
             # source mesh's reference cell as an extra argument for the inner
             # loop. (With a vertex only mesh this is a single point for each
             # vertex cell.)
-            coefficients = coefficients + [target_mesh.reference_coordinates]
+            coefficients.append(target_mesh.reference_coordinates)
 
     if tensor in set((c.dat for c in coefficients)):
         output = tensor
@@ -361,9 +348,11 @@ def _interpolator(V, tensor, expr, subset, arguments, access):
         assert access == op2.WRITE  # Other access descriptors not done for Matrices.
         rows_map = V.cell_node_map()
         columns_map = arguments[0].function_space().cell_node_map()
-        if trans_mesh:
-            # We have to map to the source mesh cells before mapping to nodes
-            columns_map = compose_map_and_cache(target_mesh_source_mesh_map,
+        if target_mesh is not source_mesh:
+            # Since the par_loop is over the target mesh cells we need to
+            # compose a map that takes us from target mesh cells to the
+            # function space nodes on the source mesh.
+            columns_map = compose_map_and_cache(target_mesh.cell_parent_cell_map,
                                                 columns_map)
         parloop_args.append(tensor(op2.WRITE, (rows_map, columns_map)))
     if oriented:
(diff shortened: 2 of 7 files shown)


firedrake/cython/dmcommon.pyx
```diff
@@ -2631,21 +2637,67 @@ def label_pic_parent_cell_nums(PETSc.DM swarm, parentmesh):
...
-    swarm_coords = swarm.getField("DMSwarmPIC_coor").reshape((num_vertices, dim))
+    swarm_coords = swarm.getField("DMSwarmPIC_coor").reshape((num_vertices, gdim))
     parent_cell_nums = swarm.getField("parentcellnum")
+    reference_coords = swarm.getField("refcoord").reshape((num_vertices, tdim))
 
     # find parent cell numbers
     for i in range(max_num_vertices):
         if i < num_vertices:
-            parent_cell_num = parentmesh.locate_cell(swarm_coords[i])
+            parent_cell_num, reference_coord = parentmesh.locate_cell_and_reference_coordinate(swarm_coords[i])
```
ReubenHill, line 2651: We should be able to do this for all the points in in one call to the parent mesh's `_c_locator`. Probably not one to fix in this PR

firedrake/interpolation.py
```diff
@@ -263,6 +314,19 @@ def _interpolator(V, tensor, expr, subset, arguments, access):
         # Replace with real coords coefficient
         coefficients[0] = coords
 
+    if runtime_vertex_only_mesh_interp:
+        # NOTE: Shouldn't this happen in compile_expression_dual_evaluation?
+        coefficients = coefficients + [mesh.reference_coordinates]
+        # TODO: stash the composed map on cell_parent_cell_map using caching
```
ReubenHill, line 320: done! remove me

firedrake/cython/dmcommon.pyx
ReubenHill: Created new issue #2069 and added comment

firedrake/interpolation.py
```diff
@@ -263,6 +314,18 @@ def _interpolator(V, tensor, expr, subset, arguments, access):
         # Replace with real coords coefficient
         coefficients[0] = coords
 
+    if runtime_vertex_only_mesh_interp:
+        # NOTE: Shouldn't this happen in compile_expression_dual_evaluation?
```
ReubenHill, line 318: Change this to comment what I'm actually doing rather than begging a question

firedrake/interpolation.py
```diff
@@ -229,15 +233,62 @@ def _interpolator(V, tensor, expr, subset, arguments, access):
...
+            from finat.tensorfiniteelement import TensorFiniteElement
+            assert isinstance(to_element, TensorFiniteElement)
+            assert isinstance(to_element.base_element, DiscontinuousLagrange)
+            assert to_element.base_element.value_shape == ()
+            scalar_element = QuadratureElement(fiat_cell, QuadratureRule(UnknownPointSingleton(point), weights=[1.]))
+            to_element = TensorFiniteElement(scalar_element, to_element.value_shape, to_element._transpose)
+        trans_mesh_allowed = True
+        # switch domain coords to expression's mesh and mesh coordinates
+        dual_eval_domain = expr.ufl_domain()
+        coords = expr.ufl_domain().coordinates
```
ReubenHill, line 282: Strip out into separate function

firedrake/interpolation.py
```diff
@@ -229,15 +233,63 @@ def _interpolator(V, tensor, expr, subset, arguments, access):
...
+        from finat.fiat_elements import DiscontinuousLagrange
+        if to_element.value_shape == ():
+            assert isinstance(to_element, DiscontinuousLagrange)
+            to_element = QuadratureElement(fiat_cell, QuadratureRule(UnknownPointSingleton(point), weights=[1.]))
+        else:
+            from finat.tensorfiniteelement import TensorFiniteElement
+            assert isinstance(to_element, TensorFiniteElement)
+            assert isinstance(to_element.base_element, DiscontinuousLagrange)
+            assert to_element.base_element.value_shape == ()
+            scalar_element = QuadratureElement(fiat_cell, QuadratureRule(UnknownPointSingleton(point), weights=[1.]))
```
ReubenHill, line 278: don't make this twice

firedrake/interpolation.py
ReubenHill: Also presumably `compile_expression_dual_evaluation` ought to be able to indicate whether it expects runtime arguments

wence-: We should do that while redefining the surface interface so that we can cache kernels from `compile_expression...`.

ReubenHill: What do you mean by "the surface interface"?

wence-: Just the API offered by the return value from `c_e_d_e`.

firedrake/interpolation.py
```diff
@@ -280,22 +327,39 @@ def _interpolator(V, tensor, expr, subset, arguments, access):
...
+                isinstance(target_mesh.topology,
+                           firedrake.mesh.VertexOnlyMeshTopology)
+                and not isinstance(
+                    coefficient.function_space().mesh().topology,
+                    firedrake.mesh.VertexOnlyMeshTopology)
+            )
+        )
+        if get_composed_map:
+            m_ = compose_map_and_cache(target_mesh.cell_parent_cell_map, coefficient.cell_node_map())
+            parloop_args.append(coefficient.dat(op2.READ, m_))
```
ReubenHill, line 354: Note we don't need a composed map for the pyop2 output variable (`tensor`) since we want our result to be defined on the target mesh cells which we will loop over.

firedrake/interpolation.py
```diff
@@ -304,6 +368,88 @@ def _interpolator(V, tensor, expr, subset, arguments, access):
...
+    # gem.Variable name starting with rt_ forces TSFC runtime tabulation
+    runtime_point = gem.Variable('rt_X', (expr_tdim,))
+    rule_pointset = finat.point_set.UnknownPointSingleton(runtime_point)
+    try:
+        expr_fiat_cell = as_fiat_cell(expr.ufl_element().cell())
+    except AttributeError:
+        # expression must be pure function of spatial coordinates so
+        # domain has correct ufl cell
+        expr_fiat_cell = as_fiat_cell(expr.ufl_domain().ufl_cell())
+    rule = finat.quadrature.QuadratureRule(rule_pointset, weights=[1.])
```
ReubenHill, line 393: `weights=[1.] * len(point_set)`

firedrake/interpolation.py
```diff
@@ -304,6 +368,88 @@ def _interpolator(V, tensor, expr, subset, arguments, access):
...
+@rebuild.register(finat.DiscontinuousLagrange)
+def rebuild_dg(element, expr):
+    # Create a quadrature element for a runtime point with weight 1
+    # to represent the runtime tabulated point, on the expression's
+    # mesh reference cell, onto which we interpolate
+    if element.degree != 0:
+        raise ValueError("Can only runtime interpolate onto DG0")
+    expr_tdim = expr.ufl_domain().topological_dimension()
+    # gem.Variable name starting with rt_ forces TSFC runtime tabulation
+    runtime_point = gem.Variable('rt_X', (expr_tdim,))
```
ReubenHill, line 385: Make `UnknownPointSet` and `runtime_points = gem.Variable('rt_X', (expr_tdim, num_points))`

firedrake/interpolation.py
```diff
@@ -304,6 +368,88 @@ def _interpolator(V, tensor, expr, subset, arguments, access):
...
+    raise ValueError("Not expecting %s" % element)
+
+
+@rebuild.register(finat.DiscontinuousLagrange)
+def rebuild_dg(element, expr):
+    # Create a quadrature element for a runtime point with weight 1
+    # to represent the runtime tabulated point, on the expression's
+    # mesh reference cell, onto which we interpolate
+    if element.degree != 0:
+        raise ValueError("Can only runtime interpolate onto DG0")
```
ReubenHill, line 382: not necessary

firedrake/interpolation.py
```diff
@@ -304,6 +368,102 @@ def _interpolator(V, tensor, expr, subset, arguments, access):
...
+    # this uses FIAT's dual basis for now but, when implemented, should use
+    # FInAT's dual evaluation infrastructure.
+    num_points = sum(len(dual.get_point_dict()) for dual in element.fiat_equivalent.dual_basis())
+    weights = []
+    for dual in element.fiat_equivalent.dual_basis():
+        pts = dual.get_point_dict().keys()
+        for p in pts:
+            for w, _ in dual.get_point_dict()[p]:
+                weights.append(w)
+    assert len(weights) == num_points
```
ReubenHill, line 397: This being here makes me wonder about 
 - (a) whether this should really live here and
 - (b) how this will work, in practice, with elements with complicated dual bases

ReubenHill: (I'm trying to think ahead to getting this working on higher dimension meshes)

ReubenHill: Okay so conclusion: This is useful for thinking about future design - in the future this can be absorbed into FInAT as a transformer that lets you do run time tabulation of that element. I can comment as much here. But for now there's really no need to be using FIAT's hard-work dual basis API so just hard code what the values are in this case.

ReubenHill:
All comments addressed and all tests passing - ready for re review.

firedrake/interpolation.py
```diff
@@ -254,20 +293,48 @@ def _interpolator(V, tensor, expr, subset, arguments, access):
...
             pts, = dual.pt_dict.keys()
             to_pts.append(pts)
-        kernel, oriented, needs_cell_sizes, coefficients = compile_python_kernel(expr, to_pts, to_element, V, coords)
+        kernel, oriented, needs_cell_sizes, coefficients = compile_python_kernel(expr, to_pts, to_element, V, source_coords_coeff)
         first_coeff_fake_coords = False
     else:
         raise RuntimeError("Attempting to evaluate an Expression which has no value.")
 
-    cell_set = coords.cell_set
+    cell_set = target_mesh.coordinates.cell_set
```
ksagiyam, line 301: `target_mesh.cell_set`

firedrake/interpolation.py
```diff
@@ -254,20 +293,48 @@ def _interpolator(V, tensor, expr, subset, arguments, access):
...
+            # vertex cell.)
+            coefficients = coefficients + [target_mesh.reference_coordinates]
+        # In the trans-mesh case we loop over the target mesh cells so we need
+        # to compose a map that takes us from their cells to the source mesh
+        # cells, and (where necessary) to the function space nodes on the
+        # source mesh.
+        if isinstance(target_mesh.topology, firedrake.mesh.VertexOnlyMeshTopology):
+            target_mesh_source_mesh_map = target_mesh.cell_parent_cell_map
+        else:
+            raise NotImplementedError("Can only interpolate onto a Vertex Only Mesh")
```
ksagiyam, line 337: Maybe we should include this check when we define `trans_mesh`.

firedrake/interpolation.py
```diff
@@ -227,17 +234,49 @@ def _interpolator(V, tensor, expr, subset, arguments, access):
...
+        # interpolating into/onto
+        source_coords_coeff = target_mesh.coordinates
+        source_domain = V.mesh()
 
     parameters = {}
     parameters['scalar_type'] = utils.ScalarType
 
     if not isinstance(expr, firedrake.Expression):
-        if expr.ufl_domain() and expr.ufl_domain() != V.mesh():
-            raise NotImplementedError("Interpolation onto another mesh not supported.")
```
ksagiyam, line 238: I think this check is still helpful if `not trans_mesh`.

firedrake/interpolation.py
```diff
@@ -227,17 +234,49 @@ def _interpolator(V, tensor, expr, subset, arguments, access):
...
+        to_element = rebuild(to_element, expr, rt_var_name)
+        # The source domain is on the expression not the FunctionSpace/Function
+        # we are interpolating into/onto
+        source_coords_coeff = expr.ufl_domain().coordinates
+        source_domain = expr.ufl_domain()
+    else:
+        # The source domain is that of the FunctionSpace/Function we are
+        # interpolating into/onto
+        source_coords_coeff = target_mesh.coordinates
+        source_domain = V.mesh()
```
ksagiyam, line 272:
```
source_domain = target_mesh
source_coords_coeff = target_mesh.coordinates
```
Maybe `source_domain` should be renamed `source_mesh`, or can these two object be just called `mesh` and `coords`?

ksagiyam (changes requested):

firedrake/interpolation.py
ReubenHill: Agreed

firedrake/interpolation.py
ReubenHill: Tidied. Though I think keeping a clear distinction between source and target in the naming is helpful. I found myself getting mixed up until I added the prefixes.

firedrake/interpolation.py
ReubenHill: changed

firedrake/interpolation.py
ksagiyam: Yes, you are right.

firedrake/interpolation.py
```diff
@@ -166,10 +169,14 @@ def make_interpolator(expr, V, subset, access):
     elif len(arguments) == 1:
         if isinstance(V, firedrake.Function):
             raise ValueError("Cannot interpolate an expression with an argument into a Function")
-
         argfs = arguments[0].function_space()
+        argfs_map = argfs.cell_node_map()
+        if argfs_map is None or argfs_map.iterset != V.cell_node_map().iterset:
+            if isinstance(V.ufl_domain().topology, firedrake.mesh.VertexOnlyMeshTopology):
+                # Compose a vertex-cell to parent-cell-function-space-node map
+                argfs_map = compose_map_and_cache(V.ufl_domain().cell_parent_cell_map, argfs_map)
```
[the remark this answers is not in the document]
ReubenHill: Indeed - don't even know why the `argfs_map is None` is in there!

ReubenHill:
Failing `tests/vertexonly/test_poisson_inverse_conductivity.py::test_poisson_inverse_conductivity` is not reproducable locally (though I haven't tried running the full test suite). Unclear why `tests/regression/test_nullspace.py::test_nullspace_mixed` is suddenly failing though I see I'm not the only person seeing that failure.

firedrake/interpolation.py
```diff
@@ -227,17 +242,55 @@ def _interpolator(V, tensor, expr, subset, arguments, access):
...
-    mesh = V.ufl_domain()
-    coords = mesh.coordinates
+    target_mesh = V.ufl_domain()
+    try:
+        trans_mesh = (
+            expr.ufl_domain() is not None  # Constant expr
+            and expr.ufl_domain() != V.mesh()  # Coming from a different domain
+        )
+    except AttributeError:
+        # Have python Expression
```
ksagiyam, line 252: It looks this is new in this PR?

firedrake/interpolation.py
```diff
@@ -254,20 +307,40 @@ def _interpolator(V, tensor, expr, subset, arguments, access):
...
+        #       them rather than having to be added post-hoc (likely by
+        #       replacing `to_element` with a CoFunction/CoArgument as the
+        #       target `dual` which would contain `dual` related
+        #       coefficient(s))
+        if rt_var_name in [arg.name for arg in kernel.code[name].args]:
+            # Add the coordinates of the target mesh quadrature points in the
+            # source mesh's reference cell as an extra argument for the inner
+            # loop. (With a vertex only mesh this is a single point for each
+            # vertex cell.)
+            coefficients = coefficients + [target_mesh.reference_coordinates]
```
ksagiyam, line 343: `coefficients.append(target_mesh.reference_coordinates)`

firedrake/interpolation.py
```diff
@@ -254,20 +307,40 @@ def _interpolator(V, tensor, expr, subset, arguments, access):
...
         cell_set = subset
     parloop_args = [kernel, cell_set]
 
     if first_coeff_fake_coords:
-        # Replace with real coords coefficient
-        coefficients[0] = coords
+        # Replace with real source mesh coordinates
+        coefficients[0] = source_mesh.coordinates
+
+    if trans_mesh:
```
ksagiyam, line 325: (Suggested) `if target_mesh is not source_mesh`:

firedrake/interpolation.py
```diff
@@ -286,28 +359,133 @@ def _interpolator(V, tensor, expr, subset, arguments, access):
...
     else:
         assert access == op2.WRITE  # Other access descriptors not done for Matrices.
-        parloop_args.append(tensor(op2.WRITE, (V.cell_node_map(),
-                                               arguments[0].function_space().cell_node_map())))
+        rows_map = V.cell_node_map()
+        columns_map = arguments[0].function_space().cell_node_map()
+        if trans_mesh:
+            # We have to map to the source mesh cells before mapping to nodes
+            columns_map = compose_map_and_cache(target_mesh_source_mesh_map,
+                                                columns_map)
```
ksagiyam, line 367: (Suggested) `target_mesh_source_mesh_map` -> ` target_mesh.cell_parent_cell_map`

firedrake/interpolation.py
```diff
@@ -227,17 +242,55 @@ def _interpolator(V, tensor, expr, subset, arguments, access):
...
+        # equivalent(s) (i.e. finat.point_set.UnknownPointSet(s))
+        rt_var_name = 'rt_X'
+        to_element = rebuild(to_element, expr, rt_var_name)
+        # The source mesh is on the expression not the FunctionSpace/Function
+        # we are interpolating into/onto
+        source_mesh = expr.ufl_domain()
+    else:
+        # The source mesh is that of the FunctionSpace/Function we are
+        # interpolating into/onto
+        source_mesh = target_mesh
```
ksagiyam, line 286: (Suggested) The logic used in the above (ll. 245 - 286) is slightly confusing to me. I would do:
```
target_mesh = V.ufl_domain()
source_mesh = expr.ufl_domain() or target_mesh
if source_mesh is not target_mesh:
    assert isinstance(target_mesh.topology, firedrake.mesh.VertexOnlyMeshTopology)
    assert target_mesh._parent_mesh is source_mesh
    rt_var_name = 'rt_X'
    to_element = rebuild(to_element, expr, rt_var_name)
```
You need to handle `except AttributeError` part if you need to. Also, I would not introduce new variable for `target_mesh.cell_parent_cell_map`. Maybe we don't need `trans_mesh`, either.

firedrake/interpolation.py
```diff
@@ -286,28 +359,133 @@ def _interpolator(V, tensor, expr, subset, arguments, access):
...
-    for o in coefficients:
-        domain = o.ufl_domain()
-        if domain is not None and domain.topology != mesh.topology:
-            raise NotImplementedError("Interpolation onto another mesh not supported.")
+        if coefficient.ufl_domain() == source_mesh and trans_mesh:
+            m_ = compose_map_and_cache(target_mesh_source_mesh_map, coefficient.cell_node_map())
+            parloop_args.append(coefficient.dat(op2.READ, m_))
+        else:
+            m_ = coefficient.cell_node_map()
+            parloop_args.append(coefficient.dat(op2.READ, m_))
```
ksagiyam, line 381: (Suggested) The following would be slightly more intuitive:
```
for coefficient in coefficients:
    coeff_mesh = coefficient.ufl_domain()
    if coeff_mesh is target_mesh or not coeff_mesh:
        m_ = coefficient.cell_node_map()
    elif coeff_mesh is source_mesh:
        m_ = compose_map_and_cache(target_mesh.cell_parent_cell_map, coefficient.cell_node_map())
    else:
        raise ...
    parloop_args.append(coefficient.dat(op2.READ, m_))
```

ksagiyam (changes requested):
Some suggestions which I believe would tidy up the logic.

I also found that the following fails on your branch:
```
from firedrake import *
import numpy as np

m = UnitSquareMesh(2, 2)
vm = VertexOnlyMesh(m, np.array([[0.1, 0.2], ]))
V = FunctionSpace(m, "Real", 0)
W = FunctionSpace(vm, "DG", 0)
A = Interpolator(TestFunction(V), W)
```
Am I doing something wrong?

ReubenHill:
> Some suggestions which I believe would tidy up the logic.
> 
> I also found that the following fails on your branch:
> 
> ```
> from firedrake import *
> import numpy as np
> 
> m = UnitSquareMesh(2, 2)
> vm = VertexOnlyMesh(m, np.array([[0.1, 0.2], ]))
> V = FunctionSpace(m, "Real", 0)
> W = FunctionSpace(vm, "DG", 0)
> A = Interpolator(TestFunction(V), W)
> ```
> 
> Am I doing something wrong?

Right - so this would be the reason why I tried to allow for `argfs_map` being `None` (which it is for `FunctionSpace(m, "Real", 0)`). I'll need to check if this is an easy fix or just needs to be disallowed.

firedrake/interpolation.py
ReubenHill: It's not - python expressions are allowed (see for example this test from `tests/regression/test_python_parloop.py` 
```python
def test_python_parloop():

    m = UnitSquareMesh(4, 4)
    fs = FunctionSpace(m, "CG", 2)
    f = Function(fs)

    class MyExpression(Expression):
        def eval(self, value, X):

            value[:] = np.dot(X, X)

    f.interpolate(MyExpression())
    X = m.coordinates
    assert assemble((f-dot(X, X))**2*dx)**.5 < 1.e-15
```
but these expressions throw an `AttributeError` if you try to access their `ufl_domain` property. This allows for that to happen.

firedrake/interpolation.py
ReubenHill: done

firedrake/interpolation.py
ReubenHill: I am trying to account for a future case where the `target_mesh_source_mesh_map` map may not be the `target_mesh.cell_parent_cell_map`.

ReubenHill: I'll make this a single dispatch function I think

ReubenHill: Actually I think `target_mesh.cell_parent_cell_map` is fine

firedrake/interpolation.py
ReubenHill: I agree

firedrake/interpolation.py
ReubenHill: Agreed

firedrake/interpolation.py
ksagiyam: I see. Then we can probably override `ufl_domain` method of `firedrake.Expression` so that it would always return `None`?

firedrake/interpolation.py
ReubenHill: Done

firedrake/interpolation.py
ReubenHill: agreeed

firedrake/interpolation.py
ReubenHill: Yes but I'm not going to do that in this PR

ReubenHill:
> > Some suggestions which I believe would tidy up the logic.
> > I also found that the following fails on your branch:
> > ```
> > from firedrake import *
> > import numpy as np
> > 
> > m = UnitSquareMesh(2, 2)
> > vm = VertexOnlyMesh(m, np.array([[0.1, 0.2], ]))
> > V = FunctionSpace(m, "Real", 0)
> > W = FunctionSpace(vm, "DG", 0)
> > A = Interpolator(TestFunction(V), W)
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Am I doing something wrong?
> 
> Right - so this would be the reason why I tried to allow for `argfs_map` being `None` (which it is for `FunctionSpace(m, "Real", 0)`). I'll need to check if this is an easy fix or just needs to be disallowed.

Okay this is all added and led me to making two new issues - #2093 and #2095

firedrake/interpolation.py
```diff
@@ -166,10 +169,26 @@ def make_interpolator(expr, V, subset, access):
...
+                raise ValueError("Cannot interpolate onto a mesh of a different geometric dimension")
+            if not hasattr(target_mesh, "_parent_mesh") or target_mesh._parent_mesh is not source_mesh:
+                raise ValueError("Can only interpolate across meshes where the source mesh is the parent of the target")
+            if argfs_map:
+                # Since the par_loop is over the target mesh cells we need to
+                # compose a map that takes us from target mesh cells to the
+                # function space nodes on the source mesh. NOTE: argfs_map is
+                # allowed to be None when interpolating from a Real space, even
+                # in the trans-mesh case.
+                argfs_map = compose_map_and_cache(V.ufl_domain().cell_parent_cell_map, argfs_map)
```
ksagiyam, line 189: `argfs_map = compose_map_and_cache(target_mesh.cell_parent_cell_map, argfs_map)`

firedrake/interpolation.py
```diff
@@ -166,10 +169,26 @@ def make_interpolator(expr, V, subset, access):
...
+        target_mesh = V.ufl_domain()
+        source_mesh = argfs.mesh()
+        argfs_map = argfs.cell_node_map()
+        if target_mesh is not source_mesh:
+            if not isinstance(target_mesh.topology, firedrake.mesh.VertexOnlyMeshTopology):
+                raise NotImplementedError("Can only interpolate onto a Vertex Only Mesh")
+            if target_mesh.geometric_dimension() != source_mesh.geometric_dimension():
+                raise ValueError("Cannot interpolate onto a mesh of a different geometric dimension")
+            if not hasattr(target_mesh, "_parent_mesh") or target_mesh._parent_mesh is not source_mesh:
+                raise ValueError("Can only interpolate across meshes where the source mesh is the parent of the target")
```
ksagiyam, line 182: I think 
```
if target_mesh._parent_mesh is not source_mesh:
    raise ...
```
is good enough for now. (It is stronger than just checking geometric dimensions). When we generalise this for other `trans_mesh` cases, we would need more changes anyway.

firedrake/interpolation.py
```diff
@@ -227,17 +246,43 @@ def _interpolator(V, tensor, expr, subset, arguments, access):
...
-    mesh = V.ufl_domain()
-    coords = mesh.coordinates
+    # NOTE: The par_loop is always over the target mesh cells.
+    target_mesh = V.ufl_domain()
+    try:
+        source_mesh = expr.ufl_domain() or target_mesh
+    except AttributeError:
+        # Have python Expression which does not have ann accessible ufl_domain
+        assert isinstance(expr, firedrake.Expression)
+        source_mesh = target_mesh
```
ksagiyam, line 256:
```
if isinstance(expr, firedrake.Expression):
    source_mesh = target_mesh
else:
    source_mesh = expr.ufl_domain() or target_mesh
```

ksagiyam (changes requested):

firedrake/interpolation.py
ReubenHill: Good spot. Changed.

firedrake/interpolation.py
ReubenHill: I need the `hasattr(target_mesh, "_parent_mesh")` otherwise you'll get an `AttributeError` if you try to interpolate across meshes in any case other than from parent mesh to immersed `VertexOnlyMesh`.

ReubenHill: From the point of view of interpolation you need `target_mesh.geometric_dimension() == source_mesh.geometric_dimension()` in all cases so I think it's valuable that it's at least pointed out for now. If there is ever any way that one managed to make `target_mesh._parent_mesh` have a different geometric dimension this would catch it. The fact I can't think of how that would happen doesn't mean one shouldn't check methinks.

dham (approved):

ReubenHill:
Cross branch references removed - awaiting policy decision on merge vs squash merge (merge preferable here)

PR #2065, opened 2021-05-11
Add PETSc to requirements-git.txt

connorjward:
This fixes an install issue where these were removed from PyOP2 in
https://github.com/OP2/PyOP2/pull/617.


connorjward:
But we might now hit the same issue with petsc4py not being explicitly listed

wence-:
> We are now getting `ModuleNotFoundError: No module named 'petsc4py'`. Shall I just include the additional line in the `requirements-git.txt` file even though it's nonsense? I can add an issue so we don't forget to fix this at a later time.

Maybe just before the line that does `packages += ["firedrake"]` do `packages.append("petsc4py")` ?

PR #2038, opened 2021-04-22
Loopy sprint

sv2518:
Let Jenkins run on Firedrake loopy sprint branch

Linked to https://github.com/firedrakeproject/tsfc/pull/244 and https://github.com/OP2/PyOP2/pull/612

Requires loopy branch https://github.com/firedrakeproject/loopy/tree/kernel_callables_v3-edit2

PR #2068, opened 2021-05-12
Add petsc4py to firedrake-update package list

connorjward:
This is essentially the same as #2065 but fixes `firedrake-update`.

### DIFF
--- a/scripts/firedrake-install
+++ b/scripts/firedrake-install
@@ -1650,7 +1650,7 @@ else:
         name = split_requirements_url(p)[0]
         deps.update(list_cloned_dependencies(name))
         deps[name] = p
-    packages = list(deps.keys())
+    packages = ["petsc4py"] + list(deps.keys())
     packages += ["firedrake"]
 
     # update packages.


sv2518:
I have just cherry-picked this into #2038. In that PR I also added firedrake-update before we build the docs, but the firedrake-update is failing. You can see the failure if you click on the failed action of the doc build int that PR. Maybe I am missing something, but maybe this is not a sufficient fix.

sv2518:
Ah no, actually I understand why it doesn't work on my PR. Cherry-picking is insufficient for me there because I think the firedrake-update that I have added into the actions of the docs build runs basically on the master branch. This should work for me as soon as it is merged.

wence-:
> I have just cherry-picked this into #2038. In that PR I also added firedrake-update before we build the docs, but the firedrake-update is failing. You can see the failure if you click on the failed action of the doc build int that PR. Maybe I am missing something, but maybe this is not a sufficient fix.

Can you run `firedrake-status` in that workflow to see if the commit is there? You can probably replicate more quickly by running the sequence of docker commands and just doing stuff "by hand".

wence-:
firedrake-update is meant to update itself to the branch it's running on, I think.

connorjward:
> firedrake-update is meant to update itself to the branch it's running on, I think.

I think for the docs build this may not be right. We are using the vanilla Docker container so Firedrake is installed into `/home/firedrake/firedrake`. It's possible that `firedrake-update` is using the Firedrake installed there (master) instead of the one we are testing.

wence-:
> > firedrake-update is meant to update itself to the branch it's running on, I think.
> 
> I think for the docs build this may not be right. We are using the vanilla Docker container so Firedrake is installed into `/home/firedrake/firedrake`. It's possible that `firedrake-update` is using the Firedrake installed there (master) instead of the one we are testing.

Oh that's probably right. We should figure out how to do this properly...

ksagiyam:
Looks to me `petsc4py` is again removed from the list at l.1677: `packages.remove("petsc4py")`?

connorjward:
> Looks to me petsc4py is again removed from the list at l.1677: packages.remove("petsc4py")?

Good spot. This is a redundant thing to do. However, I think it could possibly be argued that including `"petsc4py"` inside of the `packages` list makes some sense even if we do nothing with it? It is a dependency so we should make it clear that that is the case?

connorjward:
This is an important fix. Can it get merged ASAP?

wence-:
Thanks!

PR #2074, opened 2021-05-14
Change colour bar range for DG advection example plot

dacreman:
This is small change which modifies the colour bar in the DG advection example movie. Currently most data points are above the maximum value on the colour scale and this change matches the colour scale to the range of the data.


dham:
Thanks @dacreman !

PR #2079, opened 2021-05-18
Add --install femlium to Jenkinsfile install line

ReubenHill:
Should fix failing tests/regression/test_import_applications.py::test_import_app[femlium]

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -40,7 +40,7 @@ pipeline {
               sh 'mkdir tmp'
               dir('tmp') {
                 timestamps {
-                  sh '../scripts/firedrake-install $COMPLEX --tinyasm --disable-ssh --minimal-petsc --slepc --documentation-dependencies --install thetis --install gusto --install icepack --install irksome --no-package-manager|| (cat firedrake-install.log && /bin/false)'
+                  sh '../scripts/firedrake-install $COMPLEX --tinyasm --disable-ssh --minimal-petsc --slepc --documentation-dependencies --install thetis --install gusto --install icepack --install irksome --install femlium --no-package-manager|| (cat firedrake-install.log && /bin/false)'
                 }
               }
             }


ReubenHill:
This passes tests but not running the doc build - is this a known issue? @sv2518 I think I saw you had written something about it?

wence-:
I restarted the docs build after I clicked through and noticed that github actions reported "we had an internal error".

wence-:
See also https://www.githubstatus.com (most of the time you didn't do anything wrong!)

sv2518:
So I think this is what is happening:

The doc build is running on a firedrake docker container. The latest updated docker container would have been pushed by the previous merge into master. Since master was failing the new docker (with the new loopy) has never been pushed.

There is already an issue open here https://github.com/firedrakeproject/firedrake/issues/2070 to fix the action, we basically should add a firedrake-update to that action. I tried that but the update did not do the right thing, because there are two firedrakes in the action. In this PR https://github.com/firedrakeproject/firedrake/pull/2071 I change the action so that there is only the firedrake on the container, but for some reason the firedrake-update was breaking in the container. I still need to investigate why.

sv2518:
I am just a bit confused why the build actions is not even failing properly anymore. Previously the build action fail would tell  you that it cannot find one of the loopy functions we import.

wence-:
> I am just a bit confused why the build actions is not even failing properly anymore. Previously the build action fail would tell you that it cannot find one of the loopy functions we import.

Did we get a successful loopy-based build through?

sv2518:
> Did we get a successful loopy-based build through?

Firedrake-build (=firedrake install on Jenkins on master) yes, but we don't have the docker of it yet, because we need this PR so that we pass all tests. The docker page for firedrake says "Updated 5 days ago".

sv2518 (approved):

wence-:
Thanks

PR #2080, opened 2021-05-18
Add shape, dtype and intent to parloop cache key

### LINKED ISSUE #2075: pyop2 maybe use the wrong caching kernel when the kernel is in a function
This is a test example in which I would like to create subdomain marker functions `I_water_v2` (vector function) and `I_water_s2` (scalar function).
```
# %load ns-flow.py
from firedrake import *
from matplotlib import pyplot as plt

# load the mesh generated with Gmsh
mesh = Mesh('bubble-raise.msh')

# Those are defined in geo file. They are used to indicate the sub-domains
water_id = 3
bubble_id = 4
horizontal_bdy = 11
vertical_bdy = 12
bubble_bdy = 13

def make_domain_marker(mesh, index, name=None, dtype=np.int32):
    V = FunctionSpace(mesh, 'DG', 0)
    marker = Function(V, name=name, dtype=np.int32)
    par_loop(('{[i] : 0 <= i < f.dofs}', 'f[i, 0] = 1'), 
             dx(index), {'f': (marker, WRITE)}, is_loopy_kernel=True)
    return marker

def make_dof_marker2(V, index, name=None, dtype=np.int32):
    dmarker = make_domain_marker(V.ufl_domain(), index, dtype=dtype)
    dof_marker = Function(V, name=name, dtype=dtype)
    par_loop(('{[i] : 0 <= i < A.dofs}', 
              'A[i, 0] = if(A[i, 0] > B[0, 0], A[i, 0], B[0, 0])'),
        dx, {'A' : (dof_marker, RW), 'B' : (dmarker, READ)}, is_loopy_kernel=True)
    return dof_marker

V_scalar = FunctionSpace(mesh, 'CG', 1)
V_vector = VectorFunctionSpace(mesh, 'CG', 1)

I_water_v2 = make_dof_marker2(V_vector, water_id, name='I_water_v2')
I_water_s2 = make_dof_marker2(V_scalar, water_id, name='I_water_s2')

print((max(I_water_s2.dat.data - I_water_v2.dat.data[:, 0]), max(I_water_v2.dat.data[:, 1])))
```
I got the following error: ( I first run the kernel on the vector case, and then run the kernel on the scalar case. Because the shape of A in the scalar case is (3, ) and in the vector case (3, 2),  this case the second run failure.)
```
===========================================================================
failing kernel during pre-schedule check:
===========================================================================
KERNEL: par_loop_kernel
ARGUMENTS:
A: type: np:dtype('int32'), shape: (3), dim_tags: (stride:1) aspace: private
B: type: np:dtype('int32'), shape: (1), dim_tags: (stride:1) aspace: private
DOMAINS:
{ [i] : 0 <= i <= 2 }
INAME IMPLEMENTATION TAGS:
i: None
INSTRUCTIONS:
for i
  A[2*i] = A[2*i] if A[2*i] > B[0] else B[0]  {id=insn}
end i
===========================================================================
KeyError                                  Traceback (most recent call last)
~/firedrake/src/PyOP2/pyop2/caching.py in __new__(cls, *args, **kwargs)
    196         try:
--> 197             return cls._cache_lookup(key)
    198         except (KeyError, IOError):

~/firedrake/src/PyOP2/pyop2/caching.py in _cache_lookup(cls, key)
    204     def _cache_lookup(cls, key):
--> 205         return cls._cache[key]
    206 
...
~/firedrake/src/loopy/loopy/check.py in map_subscript(self, expr)
    461 
    462             if not access_range.is_subset(shape_domain):
--> 463                 raise LoopyError("'%s' in instruction '%s' "
    464                         "accesses out-of-bounds array element"
    465                         % (expr, self.insn_id))

LoopyError: 'A[2*i]' in instruction 'insn' accesses out-of-bounds array element
```

This can be fixed by changing the order of the lines
```
I_water_v2 = make_dof_marker2(V_vector, water_id, name='I_water_v2')
I_water_s2 = make_dof_marker2(V_scalar, water_id, name='I_water_s2')
```
to
```
I_water_s2 = make_dof_marker2(V_scalar, water_id, name='I_water_s2')
I_water_v2 = make_dof_marker2(V_vector, water_id, name='I_water_v2')
```
But this will give wrong answers `(1, 1)` which should be `(0, 0)`.

Is this a bug of pyop2?

connorjward:
This fixes #2075.

Also added a test to check we don't hit this bug again.

### DIFF
--- a/firedrake/parloops.py
+++ b/firedrake/parloops.py
@@ -113,6 +113,13 @@ def _form_loopy_kernel(kernel_domains, instructions, measure, args, **kwargs):
         kernel_domains = "[] -> {[]}"
     try:
         key = (kernel_domains, tuple(instructions), tuple(map(tuple, kwargs.items())))
+        # Add shape, dtype and intent to the cache key
+        for func, intent in args.values():
+            if isinstance(func, Indexed):
+                for dat in func.ufl_operands[0].dat.split:
+                    key += (dat.shape, dat.dtype, intent)
+            else:
+                key += (func.dat.shape, func.dat.dtype, intent)
         if kernel_cache is not None:
             return kernel_cache[key]
         else:
--- a/tests/regression/test_par_loops.py
+++ b/tests/regression/test_par_loops.py
@@ -288,4 +288,3 @@ def test_par_loop_respects_shape():
     par_loop((domain, instructions), dx, {'A': (f_scalar, WRITE)},
              is_loopy_kernel=True)
     assert np.allclose(f_scalar.dat.data, 1.0)
-
--- a/tests/regression/test_par_loops.py
+++ b/tests/regression/test_par_loops.py
@@ -271,3 +271,21 @@ def test_walk_facets_rt():
              is_loopy_kernel=True)
 
     assert errornorm(f1, f2, degree_rise=0) < 1e-10
+
+
+def test_par_loop_respects_shape():
+    m = UnitSquareMesh(2, 2)
+    f_scalar = Function(FunctionSpace(m, "CG", 1))
+    f_vector = Function(VectorFunctionSpace(m, "CG", 1))
+
+    domain = "{[i] : 0 <= i < A.dofs}"
+    instructions = "A[i, 0] = 1"
+
+    par_loop((domain, instructions), dx, {'A': (f_vector, WRITE)},
+             is_loopy_kernel=True)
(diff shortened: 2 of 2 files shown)


sv2518 (dismissed):
LGTM, you just need to lint (I have just restarted the actions) and rebase.

PR #2082, opened 2021-05-19
Add safeguard to destroy in PatchBase

ScottMacLachlan:
Using PCPatch within a fieldsplit trips an error where `patch.destroy()` gets called before `patch.initialize()`.  This adds a safeguard to `patch.destroy()`, so that it just returns, instead of trying to access data that doesn't yet exist.

[stokes_test.txt](https://github.com/firedrakeproject/firedrake/files/6509836/stokes_test.txt)

### DIFF
--- a/firedrake/preconditioners/patch.py
+++ b/firedrake/preconditioners/patch.py
@@ -820,12 +820,24 @@ class PatchBase(PCSNESBase):
         self.patch = patch
 
     def destroy(self, obj):
-        # In this destructor we clean up the __firedrake_mesh__ we set on the plex.
-        d = self.plex.getDict()
-        try:
-            del d["__firedrake_mesh__"]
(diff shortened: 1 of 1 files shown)


wence-:
I guess two things:
1. How did this ever work?
2. I am worried this is hiding an actual bug somewhere.

wence-:
> I guess two things:
> 
> 1. How did this ever work?
> 2. I am worried this is hiding an actual bug somewhere.

I had a look, to answer (1) "no idea". To answer (2), this is broadly correct I will push a slight improvement.

ScottMacLachlan:
I also have no idea how this ever worked.  However, I went back to my code from 2019 and am now able to perfectly reproduce data from a table from when this was working before...

wence- (approved):

PR #2083, opened 2021-05-19
REVERT WHEN HAVE NEW BUILD HW: temp demo test skip

ReubenHill:
Temporarily skip an occasionally failing demo test.

This failure is assumed to be due to build hardware issues and is not
reproduced locally - we assume it is because of issues with build
hardware. Revert when have new build hardware.


ReubenHill:
This does what is intended but keeps failing `Check docs build cleanly`.

PR #2088, opened 2021-05-24
Enable any rank constants

jmv2009:
Need higher than order two rank tensors.

Just enabling appears to be supported by ufl

Not sure we actually need to limit the rank at all here.

Run tested to generate alternative to PerturbationSymbol(3) (on which firedrake appears to crash)

Hmm. Some errors in jenkins... Wonder what they are about.

Updated to allow any rank.

### DIFF
--- a/firedrake/constant.py
+++ b/firedrake/constant.py
@@ -17,12 +17,8 @@ def _globalify(value):
     rank = len(shape)
     if rank == 0:
         dat = op2.Global(1, data)
-    elif rank == 1:
-        dat = op2.Global(shape, data)
-    elif rank == 2:
-        dat = op2.Global(shape, data)
     else:
-        raise RuntimeError("Don't know how to make Constant from data with rank %d" % rank)
+        dat = op2.Global(shape, data)
     return dat, rank, shape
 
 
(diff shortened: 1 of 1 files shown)


firedrake/constant.py
```diff
@@ -21,8 +21,10 @@ def _globalify(value):
         dat = op2.Global(shape, data)
     elif rank == 2:
         dat = op2.Global(shape, data)
+    elif rank == 3:
+        dat = op2.Global(shape, data)
     else:
-        raise RuntimeError("Don't know how to make Constant from data with rank %d" % rank)
+        raise RuntimeError("Don't know how to make Constant from data with Tensor rank (order) %d" % rank)
```
wence-, line 27: Thanks, I think we can probably support arbitrary rank, so perhaps we can simplify all of this to:

```
if rank = len(shape):
    dat = op2.Global(1, data)
else:
   dat = op2.Global(shape, data)
```

firedrake/constant.py
```diff
@@ -67,6 +69,8 @@ def __init__(self, value, domain=None):
             e = ufl.VectorElement("Real", cell, 0, shape[0])
         elif rank == 2:
             e = ufl.TensorElement("Real", cell, 0, shape=shape)
+        elif rank == 3:
+            e = ufl.TensorElement("Real", cell, 0, shape=shape)
```
wence-, line 73: And this could just be:
```
else:
   e = ufl.TensorElement("Real", cell, 0, shape=shape)
```

wence- (changes requested):
Thanks, I think we can simplify this a bit further.

jmv2009:
Updated to allow any rank.

wence-:
> Hmm. Some errors in jenkins... Wonder what they are about.

I think this is not your fault.

wence- (approved):

wence-:
Thanks, sorry for the delay.

PR #2063, opened 2021-05-10
Annotate supermesh projection

jwallwork23:
This PR annotates the supermesh projection operator under the assumption that the input is a `Function`. A `SupermeshProjectBlock` is created whenever source and target spaces of a projection operator live on different meshes.

PR #2089, opened 2021-05-24
Only pass Functions to SupermeshProjectBlock

jwallwork23:
PR #2063 causes issues in Thetis-adjoint because SupermeshProjectBlock can receive unsupported arguments such as expressions. This fix makes sure that it only receives Functions as input.


jwallwork23:
Doc build fail and test fails appear to be unrelated to this PR.

stephankramer (dismissed):
The logic looks correct to me and corresponds to what it does inside project(). I've just restarted the tests - as this looks like a random intermittent failure indeed

wence- (approved):

PR #2092, opened 2021-05-27
WIP: Allow firedrake-install to work with Spack

connorjward:
This PR contains a minimal number of changes that need to be made to Firedrake to permit it to work with Spack. It is intended to partially replace #2021. The reason for creating a new one is to improve the chances of getting the changes merged. The previous PR included a lot of refactoring of the install script that had nothing to do with Spack. @JDBetteridge and I will submit separate PRs for those features.


firedrake/slate/slac/compiler.py
```diff
@@ -59,13 +59,12 @@
...
     if COMM_WORLD.rank == 0:
-        petsc_variables = get_petsc_variables()
-        EIGEN_INCLUDE_DIR = petsc_variables.get("EIGEN_INCLUDE")
+        EIGEN_INCLUDE_DIR = get_eigen_include_dir()
         if EIGEN_INCLUDE_DIR is None:
             raise ValueError("""Could not find Eigen configuration in %s. Did you build PETSc with Eigen?""" % PETSC_ARCH or PETSC_DIR)
-        EIGEN_INCLUDE_DIR = EIGEN_INCLUDE_DIR.lstrip('-I')
         EIGEN_INCLUDE_DIR = COMM_WORLD.bcast(EIGEN_INCLUDE_DIR, root=0)
 
+        petsc_variables = get_petsc_variables()
```
sv2518, line 67: Is there a reason why we need to call this function after the `get_eigen_include_dir` function?

firedrake/utils.py
```diff
@@ -74,3 +77,20 @@ def wrapper(f, *args, **kwargs):
...
+def get_eigen_include_dir():
+    """Return the include directory for Eigen.
+    
+    Depending on how Eigen was installed this will either be defined in
+    petscvariables or the Firedrake configuration file.
+
+    Returns ``None`` if not found.
+    """
+    try:
+        return get_petsc_variables()["EIGEN_INCLUDE_DIR"].lstrip("-I")
```
sv2518, line 91: So we call `get_petsc_variables` again here. Can we leave the call in the old line 62 and just pass the result of `get_petsc_variables` in this function as a parameter, i.e. sth like `get_eigen_include_dir(petsc_variables)`

firedrake/utils.py
```diff
@@ -1,11 +1,14 @@
 # Some generic python utilities not really specific to our work.
 from decorator import decorator
+import functools
```
sv2518, line 3: What do we need `functools` for?

setup.py
```diff
@@ -70,9 +63,6 @@ def get_petsc_dir():
                         libraries=libs,
                         extra_link_args=link_args,
                         cython_compile_time_env=cython_compile_time_env) for (ext, libs) in cythonfiles]
-if 'CC' not in env:
```
sv2518, line 73: Why does this have to go? Maybe you want to guard it with `if config["options"].get("install_mode", "default") == "default":`?

firedrake/__init__.py
```diff
@@ -2,18 +2,20 @@
...
+           or os.environ["PETSC_ARCH"] != "default":
+            raise ImportError("PETSC_DIR is set, but you did not install with --honour-petsc-dir.\n"
+                              "Please unset PETSC_DIR (and PETSC_ARCH) before using Firedrake.")
+    elif "PETSC_DIR" not in os.environ and config["options"]["honour_petsc_dir"]:
+        raise ImportError("Firedrake was installed with --honour-petsc-dir, but PETSC_DIR is not set.\n"
+                          "Please set PETSC_DIR (and PETSC_ARCH) before using Firedrake.")
+    elif not config["options"]["honour_petsc_dir"]:  # Using our own PETSC.
+        os.environ["PETSC_DIR"] = os.path.join(sys.prefix, "src", "petsc")
+        os.environ["PETSC_ARCH"] = "default"
+del os, sys, config
```
sv2518, line 18: I think we should not delete `os` here? I think we still need it in line 126 in `_omp_num_threads = int(os.environ.get('OMP_NUM_THREADS'))`? If if I am wrong you should take it out of the del in the last line of the file however.

setup.py
sv2518: Note that in case you do want to guard it with that you also need to `del config` later.

firedrake/__init__.py
connorjward: Great spot!

setup.py
connorjward: This is already set in `firedrake-install` ([here](https://github.com/firedrakeproject/firedrake/blob/master/scripts/firedrake-install#L26)). I think that it is redundant.

firedrake/slate/slac/compiler.py
connorjward: I moved it to separate the logic of getting `EIGEN_INCLUDE_DIR` (which doesn't require `petsc_variables`) from getting `BLASLAPACK_{LIB,INCLUDE}` (which do require it).

firedrake/utils.py
connorjward: That could be done but I don't really like structuring the code like that. Getters aren't supposed to take arguments.

firedrake/slate/slac/compiler.py
sv2518: Okay

firedrake/utils.py
sv2518: Ah my bad I had overseen that. OK 👍

firedrake/utils.py
sv2518: I think the reason for why getters are supposed to not take arguments in general is because they are usually attached to a class or an object that is already having the information stored on an attribute. That's not the case here. But in any case this is fine, I don't have a strong opinion about it.

connorjward:
Closing as I am now handling the conflicts using [Spack patches](https://spack.readthedocs.io/en/latest/packaging_guide.html#patches). Merging these changes can be revisited at a later date.

PR #2097, opened 2021-05-28
Docs: Fix the setSize error for parallel case

lrtfm:
Because the second parameter for setSizes is `blocksize`, we should not unpack the value of `A.getSizes()`.


wence- (approved):
Thanks!

PR #2099, opened 2021-06-02
Change solver params for this test.

stephankramer:
The first solve of test_nullspace_mixed() keeps failing on Jenkins and
the suspicion is that this might be due to mumps being flaky. Try to
workaround by changing to iterative solver.


stephankramer:
Well, passes the test now. Went for minres+none because ilu, or any other simple preconditioner I can think of, won't work with that massive zero block. Downside is that this doesn't really _need_ a nullspace as a pure Krylov method - but we do test the vector is correct (if it is applied).

dham:
Thanks @stephankramer. I'm merging now to reduce the communal pain!

PR #2101, opened 2021-06-04
Ksagiyam/handle orientations permutations

ksagiyam:
This PR is to set up a framework in Firedrake to handle DoF permutations. This is the first step to correctly handle orientaions.

Note: This will again break `DumbCheckpoint` files if one uses those for storing data.

Requires:
https://github.com/FInAT/FInAT/pull/87


ksagiyam:
@dham @wence- 
I thought about directly using PETSc orientations (returned by `.getTransitiveClosure()`), but it doesn't look very straightforward.

As far as I understand, each entity of PETSc reference cell (say tetrahedron) carries a reference orientation. When the PETSc reference cell is mapped to a plex cell, orientations (call these PETSc orientations) are computed by comparing, for each point `p`, the reference orientation with `Cone(p)`.
This mapping of the PETSc reference cell to a plex cell is defined so that PETSc orientation of the cell point would be 0 (this can be done w.l.o.g. and `.getTransitiveClosure(p)` always has 0 for the cell point).
These PETSc orientations are computed independently from how Firedrake maps the FInAT reference cell to the plex cell (PETSc orientations alone cannot tell us which entity permutations we should use to set up the lgmap), so I think, if we want to use PETSc orientations directly in Firedrake, we need to establish a direct association between the the FInAT-reference-cell-to-plex-cell-map and the PETSc-reference-cell-to-plex-cell-map, e.g. by making these two maps identical, which requires us to rewrite `cell_closure` based on the `Cone(cell)` for each cell.
Once `cell_closure` was correctly set up this way, the FInAT orientation list would just be a permutation of the PETSc orientation list for each cell; e.g., if PETSc orientation list of a certain plex cell was given as:
```
petsc_o = [o0, o1, o2, ...,o13, o14]
```
FInAT orientation list of the same cell would be something like:
```
finat_o = [o13, o14, ......., o0] (or something like this)
```
I currently use independent orientation notation in FInAT, so the above `finat_o` would need to be mapped to have the FInAT style (call this `finat_o_mapped`) to have FInAT understand it.

I currently compute `finat_o_mapped` directly comparing the FInAT reference entity orientation with the plex cone for each entity, but I wonder if I could do any better than this for now, as, even though directly using PETSc computed orientations as the above would be nice, rewriting `cell_closure` is not something we can do at least until Jiaao's project completes. 

Another smaller thing I noticed is that, if we wanted to use PETSc orientations as in the above, we need to be able to tell PETSc that we are using tensor-product cell when we construct a quad mesh to have DMPlex compute meaningfull PETSc orientations, but I don't know if it is currently possible.

wence-:
> @dham @wence-
> I thought about directly using PETSc orientations (returned by `.getTransitiveClosure()`), but it doesn't look very straightforward.
> 
> As far as I understand, each entity of PETSc reference cell (say tetrahedron) carries a reference orientation. When the PETSc reference cell is mapped to a plex cell, orientations (call these PETSc orientations) are computed by comparing, for each point `p`, the reference orientation with `Cone(p)`.
> This mapping of the PETSc reference cell to a plex cell is defined so that PETSc orientation of the cell point would be 0 (this can be done w.l.o.g. and `.getTransitiveClosure(p)` always has 0 for the cell point).
> These PETSc orientations are computed independently from how Firedrake maps the FInAT reference cell to the plex cell (PETSc orientations alone cannot tell us which entity permutations we should use to set up the lgmap), so I think, if we want to use PETSc orientations directly in Firedrake, we need to establish a direct association between the the FInAT-reference-cell-to-plex-cell-map and the PETSc-reference-cell-to-plex-cell-map, e.g. by making these two maps identical, which requires us to rewrite `cell_closure` based on the `Cone(cell)` for each cell.

If FInAT agrees with PETSc on the reference cell does this produce the same map in both cases without having to rewrite cell closure? Or do I misunderstand?

> Once `cell_closure` was correctly set up this way, the FInAT orientation list would just be a permutation of the PETSc orientation list for each cell; e.g., if PETSc orientation list of a certain plex cell was given as:
> 
> ```
> petsc_o = [o0, o1, o2, ...,o13, o14]
> ```
> 
> FInAT orientation list of the same cell would be something like:
> 
> ```
> finat_o = [o13, o14, ......., o0] (or something like this)
> ```
> 
> I currently use independent orientation notation in FInAT, so the above `finat_o` would need to be mapped to have the FInAT style (call this `finat_o_mapped`) to have FInAT understand it.

So this is a slightly 

> 
> I currently compute `finat_o_mapped` directly comparing the FInAT reference entity orientation with the plex cone for each entity, but I wonder if I could do any better than this for now, as, even though directly using PETSc computed orientations as the above would be nice, rewriting `cell_closure` is not something we can do at least until Jiaao's project completes.
> 
> Another smaller thing I noticed is that, if we wanted to use PETSc orientations as in the above, we need to be able to tell PETSc that we are using tensor-product cell when we construct a quad mesh to have DMPlex compute meaningfull PETSc orientations, but I don't know if it is currently possible.

This is something that one needs to ask @knepley. Is the problem here that because finat represents the reference quad as a tensor-product (and therefore the group by the tensor product of the 1D symmetry groups) that we don't have a group that is as big as the normal quad? This could presumably be fixed by changing this?

ksagiyam:
That FInAT agrees with PETSc on the reference cell means (at least in my interpretation) there is a direct association between entities on the FInAT reference cell and points on the PETSc reference cell; e.g., FInAT facets (0, 1, 2, 3) = PETSc facets (4, 0, 2, 1). As soon as the PETSc reference element is mapped to a plex cell in a certain way (e.g., making cell orientation 0), we know how the reference FInAT element is mapped to the same plex cell, which automatically defines the `cell_closure`, as `cell_closure` encodes how we map the FInAT reference cell to plex cells. So I think, when we make FInAT agree with PETSc on the reference cell, we cannot use the global-point-number-based `cell_closure` any more.

I think, this way, we can directly use orientations computed by PETSc.
The idea is, if PETSc facet 4 (and thus FInAT facet 0) was mapped to a plex facet point p in the above example, PETSc computes orientation of PETSc facet 4 by comparing the canonical orientation of facet 4 with Cone(p), and store this value (say `o1`). As the FInAT reference cell is bound to the PETSc reference cell, knowing the value `o1`, we know how FInAT facet 0 is oriented relative to Cone(p) and thus we know the `entity_permutations` (call this `perm1`) that we need to apply when setting up lgmaps. So, though I'm currently using different set of integers to represent orientations in FInAT, ideally, we could make `entity_permutations()` method of this FInAT element return a dictionary `d` such that
```
dim = 1
facet_id = 0
d[dim][facet_id][`o1`] = `perm1`
```
directly using `o1` computed by PETSc.

I wonder if this makes sense.

ksagiyam:
Regarding quad, yes, that is exactly what I meant to say. If we wanted to use PETSc computed orientations directly, I think we need to have PETSc compute orientations for the cell type we actually use.

wence-:
> That FInAT agrees with PETSc on the reference cell means (at least in my interpretation) there is a direct association between entities on the FInAT reference cell and points on the PETSc reference cell; e.g., FInAT facets (0, 1, 2, 3) = PETSc facets (4, 0, 2, 1). As soon as the PETSc reference element is mapped to a plex cell in a certain way (e.g., making cell orientation 0), we know how the reference FInAT element is mapped to the same plex cell, which automatically defines the `cell_closure`, as `cell_closure` encodes how we map the FInAT reference cell to plex cells. So I think, when we make FInAT agree with PETSc on the reference cell, we cannot use the global-point-number-based `cell_closure` any more.

OK, that makes sense. I think this is where we, at the end of the day, want to get to, but it is more work because we would have to do all of this symmetry group stuff for piola-mapped elements all at once as well.

So I think it makes sense to maintain this dichotomy until we can make that larger change.

ksagiyam:
Yes, that is how I understand it; we need to have all elements ready to change `cell_closure`. Thanks, Lawrence!

knepley:
I apologize for not reading everything above. The reason I have not is that I realized that PETSc orientations did not make enough sense. I have completely rewritten them. It is in the branch knepley/feature-orientation-rethink. It makes much more sense. I am probably 3 weeks away from merge since it breaks everything in Plex, but does not change any interfaces. I have written about it in order to make it comprehensible this time. I am attaching the writeup. Please look at Chapter 1.1.4.
[PlexBook.pdf](https://github.com/firedrakeproject/firedrake/files/6721461/PlexBook.pdf)

ksagiyam:
Thanks, Matt. I think we are not using the orientations computed by DMPlex for this IO project, but we want to in the near future. I will read your writeup and maybe ask some questions on Thursday.

firedrake/cython/dmcommon.pyx
```diff
@@ -666,6 +668,174 @@ def quadrilateral_closure_ordering(PETSc.DM plex,
...
+    numCells = cell_closure.shape[0]
+    entity_orientations = np.zeros_like(cell_closure)
+    if ufl_cell.is_simplex():
+        # FInAT reference triangle:
+        #
+        #   1
+        #   | \
+        #   5   3
+        #   | 6   \
+        #   0--4---2
```
wence-, line 750: This is not the finat reference triangle (vertex 1 is (1, 0)), vertex 2 is (0, 1))

firedrake/mesh.py
```diff
@@ -451,6 +451,15 @@ def cell_closure(self):
         """
         pass
 
+    @property
+    @abc.abstractmethod
+    def entity_orientations(self):
+        """2D array of entity orientations
+
+        Each row contains orientations of the cell entities.
```
wence-, line 459: Can you expand this docstring to explain what how the orientations are represented?

firedrake/mesh.py
```diff
@@ -1063,15 +1081,25 @@ def _facets(self, kind):
...
         :arg offsets: layer offsets for each entity dof.
         """
+        if entity_permutations is None:
+            # FInAT entity_permutations not yet implemented
+            entity_dofs = eutils.flat_entity_dofs(entity_dofs)
+            return super().make_cell_node_list(global_numbering, entity_dofs, None, offsets)
+        assert sorted(entity_dofs.keys()) == sorted(entity_permutations.keys()), "Mismatching dimension tuples"
+        for key in entity_dofs.keys():
+            assert sorted(entity_dofs[key].keys()) == sorted(entity_permutations[key].keys()), "Mismatching entity tuples"
+        assert all(v == 0 or v == 1 for b, v in entity_permutations), "Vertical dim index must be in [0, 1]"
```
wence-, line 1099: Style: `all(v in {0, 1} for _, v in entity_permutations)`

firedrake/mesh.py
```diff
@@ -1266,6 +1294,10 @@ def cell_closure(self):
         return dmcommon.closure_ordering(swarm, vertex_numbering,
                                          cell_numbering, entity_per_cell)
 
+    @utils.cached_property
+    def entity_orientations(self):
+        return None
```
wence-, line 1299: No need for a per-instance property just do:

```
entity_orientations = None
```

That will satisfy `ABCMeta` and have the same effect.

tests/extrusion/test_variable_layers_numbering.py
```diff
@@ -119,15 +119,15 @@ def test_numbering_one_d_P3():
...
-                         34, 40, 36, 37, 35, 41, 38, 39]]).all()
+                         35, 41, 38, 39, 34, 40, 36, 37]]).all()
 
     assert numpy.equal(V.exterior_facet_node_map().values,
                        [[21, 24, 22, 23, 14, 17, 15, 16,
-                         0, 6, 2, 3, 1, 7, 4, 5],
+                         1, 7, 4, 5, 0, 6, 2, 3],
                         [48, 51, 49, 50, 27, 30, 28, 29,
-                         34, 40, 36, 37, 35, 41, 38, 39]]).all()
+                         35, 41, 38, 39, 34, 40, 36, 37]]).all()
```
wence-, line 130: I gues these changes are consequences of the renumbering on the interior of the interval cells?

firedrake/cython/dmcommon.pyx
```diff
@@ -822,12 +1009,20 @@ def get_cell_nodes(mesh,
                 # we need to offset by the difference from the bottom.
                 if variable:
                     off += offset[flat_index[k]]*(layer_extents[c, 0] - layer_extents[entity, 0])
-                for j in range(ceil_ndofs[i]):
-                    cell_nodes[cell, flat_index[k]] = off + j
-                    k += 1
+                if entity_permutations is not None:
+                    for j in range(ceil_ndofs[i]):
+                        cell_nodes[cell, flat_index[k]] = off + entity_permutations[entity_dims[i]][entity_nums[i]][orient][j]
```
wence-, line 1014: I think this is "python indexing into a dict" in an inner loop. Can you provide short benchmarking of how much difference it makes to timing for (say) a `50x50x50` mesh with P4?

firedrake/cython/dmcommon.pyx
```diff
@@ -666,6 +668,174 @@ def quadrilateral_closure_ordering(PETSc.DM plex,
...
+def entity_orientations(mesh,
+                        np.ndarray[PetscInt, ndim=2, mode="c"] cell_closure):
+    """Compute entity orientations.
+
+    :arg mesh: The `MeshTopology` object encapsulating the mesh topology
+    :arg cell_closure: The two-dimensional array, each row of which contains
+        the closure of the associated cell
+
+    Each row of cell_closure must have been ordered to work well with
+    FInAT entity_dofs to create cell_node map (see `get_cell_nodes`).
```
wence-, line 724: I looked at that, but didn't see much that was enlightening.

Can you describe the return value a bit more?

firedrake/cython/dmcommon.pyx
```diff
@@ -666,6 +668,174 @@ def quadrilateral_closure_ordering(PETSc.DM plex,
...
+        #   0--4---2
+        #
+        CHKERR(PetscMalloc1(dim + 1, &ecounts))
+        CHKERR(PetscMalloc1(dim + 2, &eoffsets))
+        for d in range(dim + 1):
+            ecounts[d] = _ncr(dim + 1, d + 1)
+        eoffsets[0] = 0
+        for d in range(1, dim + 2):
+            eoffsets[d] = eoffsets[d - 1] + ecounts[d - 1]
+        for cell in range(numCells):
```
wence-, line 759: I think that this code needs a little exposition explaining what it is doing.

wence- (commented):
Generally OK, probably needs a bit more explanation in places.

Is this one needed for IO itself?

ksagiyam:
Thanks, Lawrence. Yes, we need to put the actual IO thing (https://github.com/firedrakeproject/firedrake/pull/1917) on top of this one.

firedrake/cython/dmcommon.pyx
ksagiyam: Fixed.

firedrake/mesh.py
ksagiyam: Added more texts.

firedrake/mesh.py
ksagiyam: Thanks. Fixed.

firedrake/mesh.py
ksagiyam: Ok. Thanks.

tests/extrusion/test_variable_layers_numbering.py
ksagiyam: Yes, exactly.

firedrake/cython/dmcommon.pyx
ksagiyam: So I've now added more words at `AbstractMeshTopology.entity_permutations`, so here I just added `See :meth:~.AbstractMeshTopology.entity_permutations`

firedrake/cython/dmcommon.pyx
ksagiyam: Added more comments. Thanks!

firedrake/cython/dmcommon.pyx
ksagiyam: original               : 0.73 min
with permutation: 2.85 min
Indeed very slow.

ksagiyam: Great improvement with `entity_permutations` preprocessed:

50x50x50 P4 example:

off + j                                                 : 0.143 min
off + perm[j] (without preprocessing): 1.47 min
off + perm[j]  (with preprocessing)     : 0.177 min

(I probably used different computer in my earlier analysis, so values are different).

PR #2111, opened 2021-06-16
Change assembly of Slate blocks

sv2518:
Hi all. In this PR I am changing how we deal with the assembly of blocks in Slate.

Generally these blocks are used to index into Tensor which are defined on MixedFunctionSpaces. So for example if we have a form defined a form defined on` W x W` and `W = U x V` and we build `A = Tensor(form)`, then we can get the contributions of, for example, the `U x U` block by asking for `A._blocks[0,0]`.

 The way we used to assemble the kernels corresponding to the blocks was in the order assemble first, index later. So we were assembling the local TSFC kernels (corresponding to A in my example) and then index the result of the assembly kernel in the Slate kernel (and may do further linear algebra operations on it).

In this PR I am changing the order of assembly and indexing to the following: first index the ufl form (with `split`), then assemble the ufl form locally with TSFC.

This would have been the right thing to do in first place so that we don't assemble the whole local tensor, when we only need bits of it. But I also need it for my project, so that local actions on blocks can be expressed easily.

I introduced a `terminal` property which helps me to avoid some checks like `isinstance(tensor, Block)` and so forth later in the compiler pipeline. The main contribution is the new `form` property. We already have this property on Tensors, but I have introduced it on Blocks and AssembledVectors, too. (The reason I introduced it on AssembledVectors too, is the same as for the introduction of `terminal` property).

Because Blocks can sit on top of AssembledVectors or Tensors, they can be rank 1 or rank 2 and depending on that, we either get a form back or a tuple of coefficients. This is not ideal because then later I need to check if form is a tuple or not. I don't really want to tuplify all `form`s (including the forms on Tensors) because it might break backward compatibility. If anyone's got a suggestion how can deal better with this, let me know.

### DIFF
--- a/firedrake/parameters.py
+++ b/firedrake/parameters.py
@@ -26,7 +26,7 @@ class Parameters(dict):
 
     def __setitem__(self, key, value):
         super(Parameters, self).__setitem__(key, value)
-        if self._update_function:
+        if hasattr(self, "_update_function") and self._update_function:
             self._update_function(key, value)
 
     def name(self):
@@ -86,6 +86,9 @@ parameters["default_matrix_type"] = "aij"
 parameters["default_sub_matrix_type"] = "baij"
 
 parameters["type_check_safe_par_loops"] = False
+
+parameters.add(Parameters("slate_compiler"))
+parameters["slate_compiler"]["optimise"] = True
 
 
 def disable_performance_optimisations():
--- a/firedrake/slate/slac/compiler.py
+++ b/firedrake/slate/slac/compiler.py
@@ -187,7 +187,7 @@ def generate_loopy_kernel(slate_expr, compiler_parameters=None):
                        oriented=builder.bag.needs_cell_orientations,
                        subdomain_id="otherwise",
                        domain_number=0,
-                       coefficient_map=tuple(range(len(slate_expr.coefficients()))),
+                       coefficient_map=slate_expr.coeff_map,
                        needs_cell_facets=builder.bag.needs_cell_facets,
                        pass_layer_arg=builder.bag.needs_mesh_layers,
                        needs_cell_sizes=builder.bag.needs_cell_sizes)
@@ -354,7 +354,7 @@ def generate_kernel_ast(builder, statements, declared_temps):
                        oriented=builder.oriented,
                        subdomain_id="otherwise",
                        domain_number=0,
-                       coefficient_map=tuple(range(len(expr_coeffs))),
+                       coefficient_map=slate_expr.coeff_map,
                        needs_cell_facets=builder.needs_cell_facets,
                        pass_layer_arg=builder.needs_mesh_layers,
                        needs_cell_sizes=builder.needs_cell_sizes)
--- a/firedrake/slate/slac/compiler.py
+++ b/firedrake/slate/slac/compiler.py
@@ -21,8 +21,11 @@ from hashlib import md5
 
 from firedrake_citations import Citations
 from firedrake.tsfc_interface import SplitKernel, KernelInfo, TSFCKernel
+
 from firedrake.slate.slac.kernel_builder import LocalLoopyKernelBuilder, LocalKernelBuilder
 from firedrake.slate.slac.utils import topological_sort, slate_to_gem, merge_loopy
+from firedrake.slate.slac.optimise import optimise
+
 from firedrake import op2
 from firedrake.logging import logger
 from firedrake.parameters import parameters
@@ -44,6 +47,7 @@ import loopy
 import gem
 from gem import indices as make_indices
 from tsfc.loopy import generate as generate_loopy
+import copy
 
 __all__ = ['compile_expression']
 
@@ -80,18 +84,18 @@ cell_to_facets_dtype = np.dtype(np.int8)
 
 class SlateKernel(TSFCKernel):
     @classmethod
-    def _cache_key(cls, expr, tsfc_parameters, coffee):
+    def _cache_key(cls, expr, compiler_parameters, coffee):
         return md5((expr.expression_hash
-                    + str(sorted(tsfc_parameters.items()))
+                    + str(sorted(compiler_parameters.items()))
                     + str(coffee)).encode()).hexdigest(), expr.ufl_domains()[0].comm
 
-    def __init__(self, expr, tsfc_parameters, coffee=False):
+    def __init__(self, expr, compiler_parameters, coffee=False):
         if self._initialized:
             return
         if coffee:
-            self.split_kernel = generate_kernel(expr, tsfc_parameters)
+            self.split_kernel = generate_kernel(expr, compiler_parameters)
         else:
-            self.split_kernel = generate_loopy_kernel(expr, tsfc_parameters)
+            self.split_kernel = generate_loopy_kernel(expr, compiler_parameters)
         self._initialized = True
 
 
@@ -110,18 +114,20 @@ def compile_expression(slate_expr, tsfc_parameters=None, coffee=False):
     if not isinstance(slate_expr, slate.TensorBase):
         raise ValueError("Expecting a `TensorBase` object, not %s" % type(slate_expr))
 
+    # Update default parameters with passed parameters
+    # The deepcopy is needed because parameters is a nested dict
+    params = copy.deepcopy(parameters)
+    if tsfc_parameters is not None:
+        params["form_compiler"].update(tsfc_parameters)
+
     # If the expression has already been symbolically compiled, then
     # simply reuse the produced kernel.
     cache = slate_expr._metakernel_cache
-    params = parameters["form_compiler"].copy()
-    if tsfc_parameters is not None:
-        params.update(tsfc_parameters)
-    tsfc_parameters = params
-    key = str(sorted(tsfc_parameters.items()))
+    key = str(sorted(params.items()))
     try:
         return cache[key]
     except KeyError:
-        kernel = SlateKernel(slate_expr, tsfc_parameters, coffee).split_kernel
+        kernel = SlateKernel(slate_expr, params, coffee).split_kernel
         return cache.setdefault(key, kernel)
 
 
@@ -145,22 +151,26 @@ def get_temp_info(loopy_kernel):
     return mem_total, num_temps, mems, shapes
 
 
-def generate_loopy_kernel(slate_expr, tsfc_parameters=None):
+def generate_loopy_kernel(slate_expr, compiler_parameters=None):
     cpu_time = time.time()
     if len(slate_expr.ufl_domains()) > 1:
         raise NotImplementedError("Multiple domains not implemented.")
 
     Citations().register("Gibson2018")
 
+    # Optimise slate expr, e.g. push blocks as far inward as possible
+    if compiler_parameters["slate_compiler"]["optimise"]:
+        slate_expr = optimise(slate_expr)
+
     # Create a loopy builder for the Slate expression,
     # e.g. contains the loopy kernels coming from TSFC
     gem_expr, var2terminal = slate_to_gem(slate_expr)
 
-    scalar_type = tsfc_parameters["scalar_type"]
+    scalar_type = compiler_parameters["form_compiler"]["scalar_type"]
     slate_loopy, output_arg = gem_to_loopy(gem_expr, var2terminal, scalar_type)
 
     builder = LocalLoopyKernelBuilder(expression=slate_expr,
-                                      tsfc_parameters=tsfc_parameters)
+                                      tsfc_parameters=compiler_parameters["form_compiler"])
 
     name = "slate_wrapper"
     loopy_merged = merge_loopy(slate_loopy, output_arg, builder, var2terminal, name)
@@ -189,7 +199,7 @@ def generate_loopy_kernel(slate_expr, tsfc_parameters=None):
     return (SplitKernel(idx, kinfo),)
 
 
-def generate_kernel(slate_expr, tsfc_parameters=None):
+def generate_kernel(slate_expr, compiler_parameters=None):
     cpu_time = time.time()
 
     if len(slate_expr.ufl_domains()) > 1:
@@ -198,7 +208,7 @@ def generate_kernel(slate_expr, tsfc_parameters=None):
     Citations().register("Gibson2018")
     # Create a builder for the Slate expression
     builder = LocalKernelBuilder(expression=slate_expr,
-                                 tsfc_parameters=tsfc_parameters)
+                                 tsfc_parameters=compiler_parameters["form_compiler"])
 
     # Keep track of declared temporaries
     declared_temps = {}
--- a/firedrake/slate/slac/kernel_builder.py
+++ b/firedrake/slate/slac/kernel_builder.py
@@ -263,7 +263,7 @@ class LocalKernelBuilder(object):
                 kint_type = kinfo.integral_type
                 needs_cell_sizes = needs_cell_sizes or kinfo.needs_cell_sizes
 
-                args = [c for i in kinfo.coefficient_map
+                args = [c for i, _ in kinfo.coefficient_map
                         for c in self.coefficient(local_coefficients[i])]
 
                 if kinfo.oriented:
@@ -483,7 +483,7 @@ class LocalLoopyKernelBuilder(object):
                                 self.cell_size_arg))
 
         # Pick the coefficients associated with a Tensor()/TSFC kernel
-        tsfc_coefficients = [tsfc_coefficients[i] for i in kinfo.coefficient_map]
+        tsfc_coefficients = [tsfc_coefficients[i] for i, _ in kinfo.coefficient_map]
         for c, cinfo in wrapper_coefficients.items():
             if c in tsfc_coefficients:
                 if isinstance(cinfo, tuple):
--- a/firedrake/slate/slac/kernel_builder.py
+++ b/firedrake/slate/slac/kernel_builder.py
@@ -598,6 +598,7 @@ class LocalLoopyKernelBuilder(object):
         tensor2temp = OrderedDict()
         inits = []
         for gem_tensor, slate_tensor in var2terminal.items():
+            assert slate_tensor.terminal, "Only terminal tensors need to be initialised in Slate kernels."
             (_, dtype), = assign_dtypes([gem_tensor], self.tsfc_parameters["scalar_type"])
             loopy_tensor = loopy.TemporaryVariable(gem_tensor.name,
                                                    dtype=dtype,
@@ -605,19 +606,21 @@ class LocalLoopyKernelBuilder(object):
                                                    address_space=loopy.AddressSpace.LOCAL)
             tensor2temp[slate_tensor] = loopy_tensor
 
-            if isinstance(slate_tensor, slate.Tensor):
+            if not slate_tensor.assembled:
                 indices = self.bag.index_creator(self.shape(slate_tensor))
                 inames = {var.name for var in indices}
                 var = pym.Subscript(pym.Variable(loopy_tensor.name), indices)
                 inits.append(loopy.Assignment(var, "0.", id="init%d" % len(inits),
                                               within_inames=frozenset(inames)))
 
-            elif isinstance(slate_tensor, slate.AssembledVector):
-                f = slate_tensor._function
-                coeff = coefficients[f]
+            else:
+                f = slate_tensor.form if isinstance(slate_tensor.form, tuple) else (slate_tensor.form,)
+                coeff = tuple(coefficients[c] for c in f)
                 offset = 0
-                ismixed = (type(f.ufl_element()) == MixedElement)
-                names = [name for (name, ext) in coeff.values()] if ismixed else coeff[0]
+                ismixed = tuple((type(c.ufl_element()) == MixedElement) for c in f)
+                names = []
+                for (im, c) in zip(ismixed, coeff):
+                    names += [name for (name, ext) in c.values()] if im else [c[0]]
 
                 # Mixed coefficients come as seperate parameter (one per space)
                 for i, shp in enumerate(*slate_tensor.shapes.values()):
--- a/firedrake/slate/slac/kernel_builder.py
+++ b/firedrake/slate/slac/kernel_builder.py
@@ -653,7 +653,7 @@ class LocalLoopyKernelBuilder(object):
         insn = loopy.CallInstruction((slate_kernel_call_output,), call, id="slate_kernel_call")
         return insn
 
-    def generate_wrapper_kernel_args(self, tensor2temp, templated_subkernels):
+    def generate_wrapper_kernel_args(self, tensor2temp):
         coords_extent = self.extent(self.expression.ufl_domain().coordinates)
         args = [loopy.GlobalArg(self.coordinates_arg, shape=coords_extent,
                                 dtype=self.tsfc_parameters["scalar_type"])]
--- a/firedrake/slate/slac/optimise.py
+++ b/firedrake/slate/slac/optimise.py
@@ -1,0 +1,85 @@
+from gem.node import MemoizerArg
+from functools import singledispatch
+from itertools import repeat
+import firedrake.slate.slate as sl
+
+
+def optimise(expression):
+    """Optimises a Slate expression, e.g. by pushing blocks inside the expression.
+
+    :arg expression: A (potentially unoptimised) Slate expression.
+
+    Returns: An optimised Slate expression
+    """
+    return push_block(expression)
+
+
+def push_block(expression):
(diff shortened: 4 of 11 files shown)


firedrake/slate/slate.py
```diff
@@ -495,6 +511,8 @@ class Block(TensorBase):
     """
 
     def __new__(cls, tensor, indices):
+        assert tensor.terminal, "Slate can only express Blocks on terminal tensors."
```
wence-, line 514: Is this a restriction we want to impose?

firedrake/slate/slate.py
```diff
@@ -553,10 +576,20 @@ def arguments(self):
         """Returns a tuple of arguments associated with the tensor."""
         return self._split_arguments
 
+    @cached_property
+    def form(self):
+        tensor, = self.operands
+        if tensor.rank > 1:
+            return ExtractSubBlock().split(tensor.form, self._indices)
+        elif tensor.rank <= 1:
+            return tuple(tensor._function.split()[i] for j in self._indices for i in j)
```
wence-, line 585: Style: I kind of prefer `for i in itertools.chain(*self._indices)`

firedrake/slate/slate.py
```diff
@@ -393,6 +400,15 @@ def __new__(cls, function):
             raise TypeError("Expecting a Coefficient or AssembledVector (not a %r)" %
                             type(function))
 
+    @cached_property
+    def terminal(self):
+        """Returns False unless the node is a terminal node."""
+        return True
```
wence-, line 406: This kind of class-level property can just be written as `terminal = True`

firedrake/slate/slac/utils.py
```diff
@@ -176,15 +177,6 @@ def _slate2gem_tensor(expr, self):
     return var
 
 
-@_slate2gem.register(sl.Block)
-def _slate2gem_block(expr, self):
-    child, = map(self, expr.children)
-    child_shapes = expr.children[0].shapes
-    offsets = tuple(sum(shape[:idx]) for shape, (idx, *_)
-                    in zip(child_shapes.values(), expr._indices))
-    return view(child, *(slice(idx, idx+extent) for idx, extent in zip(offsets, expr.shape)))
```
wence-, line 185: I wonder if, rather than removing this functionality of indexing into general expressions, we should have a preprocessing pass on slate that pushes indexing as far towards terminals as possible. e.g. `Block(Transpose(A), (0, 1)) == Transpose(Block(A, (1, 0))) == Transpose(extract_sub_block(A, (1, 0))`. Then you get the nice small blocks that you need, but if someone writes some complex expression and then takes some blocks from it, it all still works.

WDYT?

wence- (commented):
Some questions.

firedrake/slate/slate.py
sv2518: Yes, I like that better, too.

firedrake/slate/slate.py
sv2518: AH, yes. Can that be cached, too?

wence-: there's no need, there's just one class-level attribute (rather than an instance attribute). So:

```
class Foo:
   terminal = True

x = Foo()
x.terminal => True
```

sv2518: Ah right

firedrake/slate/slac/utils.py
sv2518: How are we going to do that for matrix-matrix multiplication?

firedrake/slate/slate.py
sv2518: I **think** we don't have any tests that test blocks on nodes which are not terminal. But I have just double-checked and blocks on non-terminal tensors do work on master, so we probably don't want to impose this restriction here.

firedrake/slate/slac/utils.py
wence-: I haven't thought it through, but algebraically it is possible, e.g.:
```
[A, B [E, F   = [AE + BG, AF + BH
 C, D] G, H]     CE + DG, CF + DH]

([A, B [E, F       = ([AE + BG, AF + BH
 C, D] G, H])[0,0]     CE + DG, CF + DH])[0, 0] = AE + BG
```

sv2518: Okay so for this example that would be
```
Block(Mul(M1,M2), (0,0)) = Add(Add(Block(M1, (0,0)),  Block(M2, (0,0))),
                               Add(Block(M1, (0,1)),  Block(M2, (1,0)))
```

What about inverse and solve?

wence-: For inverse you can, if you want, do it with a schur complement factorisation, e.g. (unchecked)

```
[A, B ^{-1} = [I ,-A^{-1} B  [A^{-1} 0              [I 0 
 C D]          0 , I]         0     (D - CA^{-1} B)^{-1}] -CA^{-1} I]

= [A^{-1} + A^{-1} B (D - CA^{-1} B)^{-1} C A^{-1}, -A^{-1} B (D - C A^{-1} B)^{-1}
   -(D - CA^{-1} B)^{-1} CA^{-1},  D - C^{-1} B^{-1} ]
```

But this is fiddly to get right because to do the elimination this way requires that A is invertible (you could also do it the other way, requiring that D is invertible) see the "Derivation from LDU decomposition" section at https://en.wikipedia.org/wiki/Woodbury_matrix_identity.

So for inverse/solve I would regard those as "stopping points" and not try and push the indexing inside.

sv2518: Okay I will go for the latter option for now and then we can redecide later if needed.

firedrake/slate/slate.py
wence-: Idea:

Don't require terminal tensors here, but, have an index-inliner pass in `compile_slate` that does something like:
```
Block(tensor, indices) -> subtensor if tensor.terminal else Block(tensor, indices)
```

firedrake/slate/slate.py
wence-: So this property would be the "create_subtensor" thing.

firedrake/slate/slate.py
sv2518: There is a draft of the optimisation pass in https://github.com/firedrakeproject/firedrake/pull/2111/commits/562f7d64d53d35bf21dce825b191758ee642e54a. It's passing some tests for blocks on 2-tensors. I will need to add more testing for blocks on 1-Tensors and AssembledVectors, expressions involving inverse/solve/factorisation, one expression which is a bit more involved, and expressions which do not only involve one tensor I guess.

How does the design of the overall optimisation pass look, though? Any comments?

firedrake/slate/slac/optimise.py
```diff
@@ -0,0 +1,89 @@
...
+    return ret
+
+
+@_push_block.register(sl.Add)
+def _push_block_add(expr, self):
+    if self.block:
+        ops = []
+        for op in expr.children:
+            with block(self, op, self.block._indices):
+                ops += [self(op)]
```
sv2518, line 39: Is there a way to do line 36-40 in one line? I didn't how I can use the context manager and map at the same time.

wence-: I think this is a slightly weird model. I think this is a use-case for `MemoizerArg` rather than `Memoizer` because you want to use the call stack to maintain the current indices to inline.

For inspiration, I would look at the implementation of `remove_componenttensors` in `tsfc/gem/optimise.py`

sv2518: OK will do, thanks.

sv2518:
Okay, I think this is finally ready for rereview. The only thing I still need to do is to add some more docs.

In addition to the new optimisation pass we discussed in the meeting, I added a `BlockAssembledVector`. This is necessary, because split here https://github.com/firedrakeproject/firedrake/blob/94d4e17373a67f4e21def73332e467132b0a8a01/firedrake/slate/slate.py#L642 gives back a list of coefficients and with previous Slate language I could not build an AssembledVector from that.

The thing that slowed me done in finishing this up, is related to 6bc2d30. Since the slate optimisation pass basically pulls e.g. Blocks into AssembledVectors now, the amount of coefficients that PyOP2 needs to pass into the the Slate kernel changes. Before the optimisation pass we pass e.g. 3 bits of a split mixed coefficient, after the optimisation pass we only need to pass one of the bits. I have dealt with that now by keeping track of a `split_map` which basically says which bits of the split PyOP2 needs to pass. If anyone has a better idea, let me know.

firedrake/parameters.py
```diff
@@ -87,6 +87,9 @@ def set_update_function(self, callable):
 
 parameters["type_check_safe_par_loops"] = False
 
+parameters.add(Parameters("la_compiler"))
+parameters["la_compiler"]["optimise_slate"] = True
+
```
wence-, line 92: Call this `Parameters("slate")` and just `"optimise"` (or "optimise_expressions") ? e.g.` parameters["slate"]["optimise"] = True`

firedrake/slate/slac/kernel_builder.py
```diff
@@ -598,26 +598,29 @@ def initialise_terminals(self, var2terminal, coefficients):
         tensor2temp = OrderedDict()
         inits = []
         for gem_tensor, slate_tensor in var2terminal.items():
+            assert slate_tensor.terminal, "Only terminal tensors need to be initiliased in Slate kernels."
```
wence-, line 601: typo in initialised.

firedrake/slate/slac/optimise.py
```diff
@@ -0,0 +1,55 @@
...
+
+
+@_push_block.register(sl.AssembledVector)
+def _push_block_assembled_vector(expr, self, indices):
+    return sl.BlockAssembledVector(sl.Block(expr, indices).form, indices) if indices else expr
+
+
+@_push_block.register(sl.Block)
+def _push_block_block(expr, self, indices):
+    indices = expr._indices if not indices else tuple(big[slice(*small)] for small, big in zip(expr._indices, indices))
```
wence-, line 53: Maybe assert that the lengths of `expr._indices` and `indices` match?

firedrake/slate/slac/optimise.py
```diff
@@ -0,0 +1,55 @@
...
+    block, = map(self, expr.children, repeat(indices))
```
wence-, line 54: Maybe just:
```
return self(expr.children[0], indices)
```
?

sv2518: Why is that better? Do you think `map` & `repeat` come with too much overhead if there is only one child?

Also I thought in general we prefer to destruct tuples with `child, = expr.children` rather than indexing. So maybe I should change this to the following instead?
```
@_push_block.register(sl.Block)
def _push_block_block(expr, self, indices):
    indices = expr._indices if not indices else tuple(big[slice(*small)] for small, big in zip(expr._indices, indices))
    child, = expr.children
    return self(child, indices)
```

wence-: I think I was confused the first time I read things because I didn't parse the `map(..., repeat)` paradigm as `(self(child, indices) for child in self.children)`

wence-: I don't care about the increased computational cost.

sv2518: Ah ok I will leave it as is then

firedrake/parameters.py
sv2518: I called this "la_compiler" (=linear algebra compiler) for consistency reasons because the tsfc compiler options got the the key "form_compiler".

I don't really mind, but if anything I would do Parameters("tsslac") because slate is just the language and tsslac is the compiler. Just tbc, if we would do the same for tsfc,  the key would be "ufl" that would be odd.

wence-: I think I would argue for something with slate because that's the name that this part of firedrake is referred to in papers/talks etc... The `form_compiler` thing for tsfc is historical compatibility. Arguably that could be `Parameters("tsfc")` because that's the name people know.

sv2518: Right. "slate_compiler" then. :)

wence-: 👍

firedrake/slate/slac/optimise.py
sv2518: Is here the right place to do that? I would probably check that in the generation of the Blocks?
Actually I think this should be caught in this assertion here already: https://github.com/firedrakeproject/firedrake/blob/186d644669cabdf2b17118f9ba49a6cf1135d018/firedrake/slate/slate.py#L71

wence-: I was thinking that if `len(indices) < len(expr._indices)` something might go wrong, but I am not quite sure what.

I think I see what this line does, but I am being stupid. I suppose you're inlining the indexing from the outside?

sv2518: If I understand what you mean correctly, I am inlining from the inside. If I have got the slate expression `A.blocks[:3,:3].blocks[0,0]` I encounter the blocks on the slices first. When I encounter the first `blocks` the `indices` are empty and the slices are in `expr._indices`. When I encounter the second `blocks`, the slices are stored in `indices` and (0,0) in `expr._indices`. So in line 53 I basically say `indices = ((0,1,2)[0], (0,1,2)[0])`

We get `len(indices) < len(expr._indices) ` if we would start off with `A.blocks[:3].blocks[0,0]` This is illegal slate and we already run in an error here https://github.com/firedrakeproject/firedrake/blob/186d644669cabdf2b17118f9ba49a6cf1135d018/firedrake/slate/slate.py#L77 but we should maybe add an assertion to catch this before we fail with a `TypeError: 'NoneType' object cannot be interpreted as an integer`

wence-: Thanks, makes sense now.

sv2518: Actually I realised  that the fact that this errors is my fault, it should just run through anyways, see PR here https://github.com/firedrakeproject/firedrake/pull/2121.

In any case this won't be an issue for the stuff here, so I will resolve.

sv2518:
https://github.com/firedrakeproject/firedrake/pull/2121 needs to be merged first because I included a test here that only passes with that fix.

I also don't want to this to merged before the test suite has been run on it because I have the structure of the coeff map in this commit here https://github.com/firedrakeproject/firedrake/pull/2111/commits/fd7037bc6b1bda272286595b86771b4ca6cb71af and may be unaware of other codepaths accessing it.

If tests are passing this is good to go I think. I cleaned up the history and added docs now.

sv2518:
My apologies for the noise on this PR. I have been struggling to get the coefficient map from this commit 395e24 bug-free, because it's involved in everything really and there were some corner cases I did not know about. I also had some failing test for multigrid and pyadjoint e.g., they were slightly harder to debug for me and I had to guess what is breaking them. Anyways, the last test run finally went through, so I think we are now good here, but maybe @wence- wants to have a final look on it when he is back from holidays.

firedrake/slate/slac/compiler.py
```diff
@@ -100,7 +103,7 @@ def compile_expression(slate_expr, tsfc_parameters=None, coffee=False):
     :class:`firedrake.op2.Kernel` object representing the Slate expression.
 
     :arg slate_expr: a :class:'TensorBase' expression.
-    :arg tsfc_parameters: an optional `dict` of form compiler parameters to
+    :arg parameters: an optional `dict` of form compiler parameters to
```
dham, line 106: I don't think you meant to change this. Or if you did, then you should change the code to match.

dham (dismissed):

firedrake/slate/slac/compiler.py
```diff
@@ -114,14 +117,14 @@ def compile_expression(slate_expr, tsfc_parameters=None, coffee=False):
     # simply reuse the produced kernel.
     cache = slate_expr._metakernel_cache
     params = parameters["form_compiler"].copy()
-    if tsfc_parameters is not None:
+    params.update(parameters["slate_compiler"].copy())
```
wence-, line 120: No need to copy, since `update` copies.

firedrake/slate/slac/compiler.py
```diff
@@ -114,14 +117,14 @@ def compile_expression(slate_expr, tsfc_parameters=None, coffee=False):
...
+    if tsfc_parameters is not None:  # tsfc_parameters actually contains all compiler parameters
         params.update(tsfc_parameters)
```
wence-, line 122: I think this means we have to ensure that the slate parameter names and the tsfc parameters don't overlap. Should we instead keep them separate as `{"slate": ..., "tsfc:" ...}` ?

firedrake/slate/slac/compiler.py
```diff
@@ -145,22 +148,26 @@ def get_temp_info(loopy_kernel):
...
-def generate_loopy_kernel(slate_expr, tsfc_parameters=None):
+def generate_loopy_kernel(slate_expr, compiler_parameters=None):
     cpu_time = time.time()
     if len(slate_expr.ufl_domains()) > 1:
         raise NotImplementedError("Multiple domains not implemented.")
 
     Citations().register("Gibson2018")
 
+    # Optimise slate expr, e.g. push blocks as far inward as possible
+    if compiler_parameters["optimise"]:
```
wence-, line 159: This would then be `compiler_parameters["slate"]["optimise"]`

firedrake/slate/slac/compiler.py
```diff
@@ -145,22 +148,26 @@ def get_temp_info(loopy_kernel):
...
     # e.g. contains the loopy kernels coming from TSFC
     gem_expr, var2terminal = slate_to_gem(slate_expr)
 
-    scalar_type = tsfc_parameters["scalar_type"]
+    scalar_type = compiler_parameters["scalar_type"]
     slate_loopy, output_arg = gem_to_loopy(gem_expr, var2terminal, scalar_type)
 
     builder = LocalLoopyKernelBuilder(expression=slate_expr,
-                                      tsfc_parameters=tsfc_parameters)
+                                      tsfc_parameters=compiler_parameters)
```
wence-, line 170: And this would be `compiler_parameters["tsfc"]`

firedrake/slate/slate.py
```diff
@@ -194,6 +201,17 @@ def rank(self):
     def coefficients(self):
         """Returns a tuple of coefficients associated with the tensor."""
 
+    @property
+    def coeff_map(self):
+        """A map from local coefficient numbers to the global
+        to the split global coefficient numbers.
+        The split coefficients are defined on the pieces of the originally mixed function spaces.
```
wence-, line 208: Remove `to the global` ?

firedrake/slate/slate.py
```diff
@@ -194,6 +201,17 @@ def rank(self):
...
+        """
+        return tuple((n, tuple(range(len(c.split()))))
+                     if isinstance(c, Function) or isinstance(c, Constant)
+                     else (n, tuple(range(1)))
```
wence-, line 212: `(1,)` rather than `tuple(range(1))`?

firedrake/slate/slate.py
```diff
@@ -194,6 +201,17 @@ def rank(self):
...
         """Returns a tuple of coefficients associated with the tensor."""
 
+    @property
+    def coeff_map(self):
+        """A map from local coefficient numbers to the global
+        to the split global coefficient numbers.
+        The split coefficients are defined on the pieces of the originally mixed function spaces.
+        """
+        return tuple((n, tuple(range(len(c.split()))))
+                     if isinstance(c, Function) or isinstance(c, Constant)
```
wence-, line 211: Under what circumstance are the coefficients not functions or constants?

firedrake/slate/slate.py
```diff
@@ -444,6 +468,84 @@ def _key(self):
...
+    def __new__(cls, function, split_functions, indices):
+        if isinstance(split_functions, tuple) \
+           and all(isinstance(f, Coefficient) for f in split_functions):
+            self = TensorBase.__new__(cls)
+            self._function = split_functions
+            self._indices = indices
+            self._original_function = function
+            return self
+        else:
+            raise TypeError("Expecting a tuple of Coefficients or AssembledVectors (not a %r)" %
```
wence-, line 487: I think the check requires coefficients, unless an `AssembledVector` is-a `Coefficient`?

firedrake/slate/slate.py
```diff
@@ -444,6 +468,84 @@ def _key(self):
...
+        """A map from local coefficient numbers to the split global coefficient numbers.
+        The split coefficients are defined on the pieces of the originally mixed function spaces.
+        For BlockAssembledVectors the amount of split coefficients of an original coefficient
+        is not necessarily equal to the amount of split coefficients associated with the expression.
+        In this case the amount of split coefficients is dictated by how many pieces of the original
+        tensor the block is build on, this is defined by the indices.
+        """
+        sm = super().coeff_map()
+        pos = super().coefficients().index(self._original_function)
+        sm[pos] = self._indices
```
wence-, line 501: I _think_ this won't work because `sm` has type `tuple` and is therefore immutable?

firedrake/slate/slate.py
```diff
@@ -444,6 +468,84 @@ def _key(self):
...
+        return self._argument
+
+    def coefficients(self):
+        """Returns a tuple of coefficients associated with the tensor."""
+        return self._function
+
+    def ufl_domains(self):
+        """Returns the integration domains of the integrals associated with the tensor.
+        """
+        return (domain for fs in self.arg_function_spaces for domain in fs.ufl_domains())
```
wence-, line 527: This is a generator, probably want a `tuple`?

firedrake/slate/slate.py
```diff
@@ -497,6 +599,7 @@ class Block(TensorBase):
     """
 
     def __new__(cls, tensor, indices):
+
```
wence-, line 602: Unnecessary whitespace change?

firedrake/slate/slate.py
```diff
@@ -555,6 +664,23 @@ def arguments(self):
...
         return self._split_arguments
 
+    @cached_property
+    def form(self):
+        tensor, = self.operands
+        assert tensor.terminal
+        if tensor.rank > 1:
+            # turns a Block on a Tensor into an indexed ufl form
+            return ExtractSubBlock().split(tensor.form, self._indices)
+        elif tensor.rank <= 1:
```
wence-, line 674: Can I not have a `tensor.rank == 1` and still have a form not an assembled vector inside?

firedrake/tsfc_interface.py
```diff
@@ -222,8 +225,10 @@ def tuplify(params):
     for idx, f in iterable:
         f = _real_mangle(f)
         # Map local coefficient numbers (as seen inside the
-        # compiler) to the global coefficient numbers
-        number_map = dict((n, coefficient_numbers[c])
+        # compiler) to the split global coefficient numbers
+        number_map = dict((n, (coefficient_numbers[c], tuple(range(len(c.split())))))
+                          if isinstance(c, Function) or isinstance(c, Constant)
```
wence-, line 230: When is a coefficient not a function or constant?

firedrake/tsfc_interface.py
```diff
@@ -126,7 +128,8 @@ def __init__(self, form, name, parameters, number_map, interface, coffee=False,
         :arg form: the :class:`~ufl.classes.Form` from which to compile the kernels.
         :arg name: a prefix to be applied to the compiled kernel names. This is primarily useful for debugging.
         :arg parameters: a dict of parameters to pass to the form compiler.
-        :arg number_map: a map from local coefficient numbers to global ones (useful for split forms).
+        :arg number_map: a map from local coefficient numbers
+                         the split global coefficient numbers.
```
wence-, line 132: "to the split" ?

wence- (changes requested):
Some questions

firedrake/tsfc_interface.py
sv2518: When it is a ufl `Coefficient`. The problem was that those coefficient do not have a `split` function.

firedrake/slate/slate.py
sv2518: see comment on tsfc_interface

firedrake/slate/slate.py
sv2518: Yes you are right. This functionality was actually untested, because the coeff map of the `BlockAssembledVector` is only called, when it's the outermost node. I added a test for that.
After looking at this again today, I also think we don't need this function at all. Picking the correct coefficients is already happening in https://github.com/firedrakeproject/firedrake/blob/520938d0b039854fbbf035d5081f1c35a93827c1/firedrake/slate/slate.py#L677

sv2518:
Okay, I have addressed all the re-review comments now. I have not squashed the new commits into the other ones yet, but will do so before the Firedrake meeting tomorrow, so that we can get the merge done in the meeting.

PR #2114, opened 2021-06-17
slate: IndexCreator dicts should be per-instance

### LINKED ISSUE #2108: Slate: unused inames in kernel (Loopy's warning)
Hi all,

I hope everyone is safe and healthy.

## Description of the issue

I'm solving HDG-like problems with `firedrake`. I have updated my `firedrake` recently (2 weeks ago, I think). I hit a warning related to `loopy`. More precisely, it is something like that:

```
/home/diego/firedrake/src/loopy/loopy/check.py:341: LoopyWarning: in kernel slate_wrapper: Found unused inames in kernel: frozenset({'i_id_17', 'i_id_0', 'i_id_3', 'i_id_14', 'i_id_8', 'i_id_5', 'i_id_12', 'i_id_19', 'i_id_10', 'i_id_11', 'i_id_18', 'i_id_4', 'i_id_6', 'i_id_9', 'i_id_1', 'i_id', 'i_id_13', 'i_id_16', 'i_id_7', 'i_id_2', 'i_id_15'}) Unused inames during linearization will be prohibited in Loopy version 2021.X. (add 'unused_inames' to silenced_warnings kernel attribute to disable)
  warn_with_kernel(
```

This kind of warning is raised only in the "cold" run. After the code is generated and properly cached, the warning is not printed to the output with further executions.

I understand that as a "not critical" issue. I'm reporting it because it breaks my progressbar (`tqdm`).

## Possible solution

I am not an expert of `firedrake` internals, but I had a quick look and found the Slate code related to this issue. Please see the code [here](https://github.com/firedrakeproject/firedrake/blob/9ce1619d3d0e001f432d14b55e3544c2b3fb4ac5/firedrake/slate/slac/utils.py#L309). The warning can be suppressed with an one line change:

```python
    # Generates the loopy wrapper kernel
    slate_wrapper = lp.make_function(domains, insns, args, name=name,
                                     seq_dependencies=True, target=lp.CTarget(),
                                     silenced_warnings=["unused_inames"],  # as suggested by the warning
                                     lang_version=(2018, 2))
```

I have checked it locally and it worked properly. I don't know if this is a good solution, since this is a Loopy's deprecation warning for `2021.X` versions.

## Minimal Working Example

The code below raises the warning I mentioned:

```python
from firedrake import *

def print(content_to_print):
    return PETSc.Sys.Print(content_to_print)


PETSc.Log.begin()

# Defining the mesh
N = 20
mesh = UnitSquareMesh(N, N, quadrilateral=True)

# Function space declaration
pressure_family = 'DQ' if use_quads else 'DG'
velocity_family = 'DQ' if use_quads else 'DG'
trace_family = "DGT"
degree = 1
U = VectorFunctionSpace(mesh, velocity_family, degree)
V = FunctionSpace(mesh, pressure_family, degree)
T = FunctionSpace(mesh, trace_family, degree)
W = U * V * T

# Trial and test functions
solution = Function(W)
u, p, lambda_h = split(solution)
v, q, mu_h  = TestFunctions(W)

# Mesh entities
n = FacetNormal(mesh)
h = CellDiameter(mesh)

# Exact solution
p_exact = sin(2 * pi * x) * sin(2 * pi * y)

# Forcing function
f_expression = div(-grad(p_exact))
f = Function(V).interpolate(f_expression)

# Dirichlet BCs
bc_multiplier = DirichletBC(W.sub(2), p_exact, "on_boundary")

# Edge stabilizing parameter
beta_0 = Constant(1.0e0)
beta = beta_0 / h

# Numerical flux
u_hat = u + beta * (p - lambda_h) * n

# HDG classical form
a = (dot(u, v) - div(v) * p) * dx + lambda_h("+") * jump(v, n) * dS
a += -dot(u, grad(q)) * dx + jump(u_hat, n) * q("+") * dS
L = f * q * dx
# Transmission condition
a += jump(u_hat, n) * mu_h("+") * dS
# Weakly imposed BC
a += lambda_h * dot(v, n) * ds
a += dot(u_hat, n) * q * ds

F = a - L

# Solving with Static Condensation
print("*******************************************\nSolving using static condensation.\n")
params = {
    "snes_type": "ksponly",
    "mat_type": "matfree",
    "pmat_type": "matfree",
    "ksp_type": "preonly",
    "pc_type": "python",
    # Use the static condensation PC for hybridized problems
    # and use a direct solve on the reduced system for lambda_h
    "pc_python_type": "firedrake.SCPC",
    "pc_sc_eliminate_fields": "0, 1",
    "condensed_field": {
        "ksp_type": "preonly",
        "pc_type": "lu",
        "pc_factor_mat_solver_type": "mumps",
        "ksp_monitor_true_residual": None,
    },
}

problem = NonlinearVariationalProblem(F, solution, bcs=bc_multiplier)
solver = NonlinearVariationalSolver(problem, solver_parameters=params)
solver.snes.ksp.setConvergenceHistory()
solver.solve()
print("Solver finished.\n")
```

Cheers,
Diego.

wence-:
Previously we had one iname dict for all IndexCreator instances, so we
would end up with many unused inames in the loopy kernel. Fixes #2108.

### DIFF
--- a/firedrake/slate/slac/kernel_builder.py
+++ b/firedrake/slate/slac/kernel_builder.py
@@ -766,8 +766,9 @@ class SlateWrapperBag(object):
 
 
 class IndexCreator(object):
-    inames = OrderedDict()  # pym variable -> extent
-    namer = UniqueNameGenerator(forced_prefix="i_")
+    def __init__(self):
+        self.inames = OrderedDict()  # pym variable -> extent
+        self.namer = UniqueNameGenerator(forced_prefix="i_")
 
     def __call__(self, extents):
         """Create new indices with specified extents.


sv2518:
Okay I am trying to understand this. The only `IndexCreator` object we generate in Slate is saved on a `SlateWrapperBag`. And we only have one `SlateWrapperBag` per Slate kernel. So what we were doing before is, if we had a PyOP2 kernel with multiple Slate kernels, we were ensuring unique inames across Slate kernels? And your change is ensuring they are only unique across one Slate kernel at a time?
If that is the case, I think I still don't understand why the inames would have been unused before your change.

wence-:
The issue was that the two properties were class properties, rather than instance properties. So the contents of the names dictionary persists across all `IndexCreator` instances. Does that make sense?

sv2518 (approved):
Ah actually I understand, we still carry around and generate loopy kernels with the inames from other Slate kernels. Ok, cool. LGTM!

PR #2118, opened 2021-06-21
Build and test Firedrake with GitHub Actions

dham:


.github/workflows/build.yml
```diff
@@ -0,0 +1,55 @@
...
+      image: firedrakeproject/firedrake-env:latest
+    # Steps represent a sequence of tasks that will be executed as
+    # part of the jobs
+    steps:
+      - uses: actions/checkout@v2
+      - name: Cleanup
+        if: ${{ always() }}
+        run: |
+          cd ..
+          rm -rf build
```
wence-, line 26: It seems tremendously odd that there are build artefacts left lying around inside a container between runs.

wence-: Although https://github.community/t/how-to-properly-clean-up-self-hosted-runners/128909

wence-:
Hmmm:

```
Testing firedrake took: 5h 22m 21s

OMP_NUM_THREADS is

OPENBLAS_NUM_THREADS is
```

.github/workflows/build.yml
dham: I'm not falling into that issue.

The artifacts are not left in containers between runs but in a folder on the build runner that is mounted in the containers. This is a widely known issue. I think what I have seems to be working, though it needs more testing.

dham:
> Hmmm:
> 
> ```
> Testing firedrake took: 5h 22m 21s
> 
> OMP_NUM_THREADS is
> 
> OPENBLAS_NUM_THREADS is
> ```

I'm not sure if this is expected or not. I need to look further. Either way, I don't think it's the current issue.  The current issue looks like the gc disabling needing an update for the new environment.

.github/workflows/build.yml
wence-: Ah ok.

firedrake/slate/slac/kernel_builder.py
```diff
@@ -766,8 +766,9 @@ def __init__(self, coeffs):
 
 
 class IndexCreator(object):
-    inames = OrderedDict()  # pym variable -> extent
-    namer = UniqueNameGenerator(forced_prefix="i_")
+    def __init__(self):
+        self.inames = OrderedDict()  # pym variable -> extent
+        self.namer = UniqueNameGenerator(forced_prefix="i_")
```
wence-, line 771: This change is already in trunk, some merge/rebase confusion?

tests/regression/test_l2pullbacks.py
```diff
@@ -129,15 +120,15 @@ def test_firedrake_helmholtz_mixed_l2pullbacks(family, degree, celltype, action,
                           ('RT', 2, 'tri', 1, False, 1.9),
                           ('RT', 2, 'tri', 2, False, 2.9),
                           ('BDM', 1, 'tri', 1, False, 1.88),
-                          ('BDM', 2, 'tri', 1, False, 1.9),
+                          pytest.param('BDM', 2, 'tri', 1, False, 1.9, marks=pytest.mark.skipcomplex),
```
wence-, line 123: Can you add a reason to the mark (referring to your issue)? `pytest.mark.skipcomplex(reason="Some string")`

.github/workflows/build.yml
```diff
@@ -0,0 +1,70 @@
...
+        include:
+         - scalar-type: real
+           complex: ""
+         - scalar-type: complex
+           complex: --complex
+    env:
+      FIREDRAKE_CI_TESTS: 1
+      PETSC_CONFIGURE_OPTIONS: --with-make-np=12 --download-mpich-device=ch3:sock
+      OMP_NUM_THREADS: 1
+      OPENBLAS_NUM_THREADS: 1
```
wence-, line 31: @JDBetteridge I think we wanted to set these in the docker files, if that is possible?

.github/workflows/build.yml
```diff
@@ -0,0 +1,70 @@
...
+          . ../build/bin/activate
+          python $(which firedrake-clean)
+          python -m pip install pytest-cov pytest-xdist
+          python -m pip list
+      - name: Test Firedrake
+        run: |
+          . ../build/bin/activate
+          echo OMP_NUM_THREADS is $OMP_NUM_THREADS
+          echo OPENBLAS_NUM_THREADS is $OPENBLAS_NUM_THREADS
+          python -m pytest --durations=200 -n 12 --cov firedrake -v tests
```
sv2518, line 59: The durations here are basically dictating which tests to put in the code coverage summary in the end right? Since the tests are running faster now, we probably want to lower the time limit?

dham: It's not quite that. This line means show the slowest 200 tests. The number we show is arbitrary, but I think 200 is unproblematic.

.github/workflows/build.yml
dham: I think that's a separate issue. All we're doing right now is duplicating what Jenkins does.

.github/workflows/build.yml
sv2518: Ah I see, I thought it was specifying a time, because the key word is `durations`. Good then.

.github/workflows/build.yml
wence-: Sure.

tests/regression/test_l2pullbacks.py
dham: Done

.github/workflows/build.yml
JDBetteridge: Having had a look, it doesn't seem possible to set the runtime environment variables when building the docker container. You can set them from the command line at runtime, or you can create a list of default environment variables in a file named `.env` in the same directory as the docker container.

References
build time: https://docs.docker.com/engine/reference/commandline/build/
runtime: https://docs.docker.com/compose/environment-variables/

firedrake/slate/slac/kernel_builder.py
dham: Fixed.

.github/workflows/build.yml
wence-: This seems far complicated than I meant:
```
FROM ubuntu:20.04
ENV OMP_NUM_THREADS 1
```

If I build that container and then run it, `OMP_NUM_THREADS` is set to 1 inside the container. This is all we want.

PR #2120, opened 2021-06-22
Don't change the user in the Firedrake env container

dham:
First step towards having Actions-compatible docker files.

I think this doesn't break anything because:

* All the downstream containers explicitly set the user
* Jenkins runs as its own user anyway.


wence- (approved):

wence-:
Looks sane.

PR #2127, opened 2021-06-25
Build docker containers using GitHub Actions

dham:


.github/workflows/build.yml
```diff
@@ -68,3 +68,107 @@ jobs:
         run: |
           cd ..
           rm -rf build
+  docker_env:
+    name: "Build Docker environment container"  
+    # The type of runner that the job will run on
+    needs: build
+    if: github.ref == 'refs/heads/master'
```
wence-, line 75: Is this the right syntax (I thought it was `if: ${{ github.ref == 'refs/heads/master' }}` ?)

PR #2139, opened 2021-07-01
Melina97/fix ad annotate solve

dham:
Call `type(self)` instead of an explicit class name in order to handle subclasses correctly.


ReubenHill:
Could someone provide some description here to remind us what the ultimate aim is?

firedrake/assemble_expressions.py
```diff
@@ -479,6 +479,7 @@ def evaluate_expression(expr, subset=None):
         if arguments is not None:
             try:
                 for kernel, iterset, args in arguments:
+                    Arg.recreate(dat_map)  # changing _init_ arguments of Arg if necessary 
                     with dereffed(args) as args:
```
dham, line 483:
```
if dat_map:
     args = tuple(a.recreate(dat=dat_map[dat]) for a in args)
```

PR #2141, opened 2021-07-06
install: respect HDF5_DIR when building h5py

ksagiyam:
`firedrake-install -honour-petsc-dir` tries to find HDF5 installation at `$PETSC_DIR/$PETSC_ARCH`, even if the we have built our own PETSc with `--with-hdf5-dir=$HDF5_DIR`. This causes the following error when installing `h5py`:
```
ERROR: Could not build wheels for h5py which use PEP 517 and cannot be installed directly
```

### DIFF
--- a/scripts/firedrake-install
+++ b/scripts/firedrake-install
@@ -624,7 +624,18 @@ elif osname == "Darwin":
     petsc_options.add("--CFLAGS=-I{}/include -Wno-implicit-function-declaration".format(mac_blas))
     petsc_options.add("--LDFLAGS=-Wl,-rpath,{0}/lib -L{0}/lib".format(mac_blas))
 
-petsc_options = list(petsc_options) + shlex.split(os.environ.get("PETSC_CONFIGURE_OPTIONS", ""))
+for option in shlex.split(os.environ.get("PETSC_CONFIGURE_OPTIONS", "")):
+    if option.startswith("--with-hdf5-dir"):
+        petsc_options.discard("--download-hdf5")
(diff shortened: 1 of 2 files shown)


scripts/firedrake-install
```diff
@@ -974,7 +974,14 @@ def build_and_install_petsc():
...
     log.info("Depending on your platform, PETSc may take an hour or more to build!")
-    assert not args.honour_petsc_dir  # Belt and braces security before we start clobbering things.
+    # Belt and braces security before we start clobbering things.
+    assert not args.honour_petsc_dir
+    if "HDF5_DIR" in os.environ:
+        log.info("""\nWarning: HDF5_DIR environment variable set, but
+ignored; to use your own HDF5, build your own PETSc
+with --with-hdf5-dir=$HDF5_DIR and then build firedrake
+with --honour-petsc-dir.
+""")
```
dham, line 984: I think I don't understand. Why would it not be OK for me to let Firedrake build PETSc and to set `PETSC_CONFIGURE_OPTIONS=--with-hdf5-dir=$HDF5_DIR`

ksagiyam: I thought `petsc_options` by default has `--download-hdf5`, so `--with-hdf5-dir=$HDF5_DIR` would cause conflict, but maybe we should just remove the former when we observe the latter?

dham: That would seem the logical thing to do.

ksagiyam: Updated and tested locally.

PR #2144, opened 2021-07-07
Add guard to starting PETSc logging

connorjward:
This fix is necessary because the [previous solution](https://github.com/firedrakeproject/firedrake/pull/2126), needed to stop
PETSc complaining when the user passes `PETSc.Log.begin` themselves,
would break nested profiling since `PETSc.Log.begin` calls
`PetscLogDefaultBegin` internally instead of `PetscLogNestedBegin`.

When the `-log_view` flag is passed PETSc checks and calls the right
`Begin` function hence why I've added this guard.

I have also observed that nested profiling does not work properly if the
user writes `PETSc.Log.begin` in their code but I don't think we can do
much about this.

### DIFF
--- a/firedrake/__init__.py
+++ b/firedrake/__init__.py
@@ -19,17 +19,17 @@ del sys, config
 import firedrake.petsc as petsc
 
 # Initialise PETSc events for both import and entire duration of program
-petsc.PETSc.Log.begin()
-_main_event = petsc.PETSc.Log.Event("firedrake")
-_main_event.begin()
+_is_logging = "log_view" in petsc.OptionsManager.commandline_options
+if _is_logging:
+    _main_event = petsc.PETSc.Log.Event("firedrake")
+    _main_event.begin()
 
-_init_event = petsc.PETSc.Log.Event("firedrake.__init__")
-_init_event.begin()
+    _init_event = petsc.PETSc.Log.Event("firedrake.__init__")
+    _init_event.begin()
 
-import atexit
-atexit.register(lambda: _main_event.end())
-
-del atexit
+    import atexit
+    atexit.register(lambda: _main_event.end())
+    del atexit
 del petsc
 
 # UFL Exprs come with a custom __del__ method, but we hold references
@@ -143,4 +143,7 @@ if (_omp_num_threads is None) or (_omp_num_threads > 1):
 del _openblas_lib, _openblas_dll, _omp_num_threads, os, cdll, find_library
 
 # Stop profiling Firedrake import
-_init_event.end()
+if _is_logging:
+    _init_event.end()
+    del _init_event
+del _is_logging


connorjward:
@Melina97 this should fix things for you

firedrake/__init__.py
```diff
@@ -13,13 +13,17 @@
...
     os.environ["PETSC_DIR"] = os.path.join(sys.prefix, "src", "petsc")
     os.environ["PETSC_ARCH"] = "default"
-del sys, config
+del config
 
 # Ensure petsc is initialised by us before anything else gets in there.
 import firedrake.petsc as petsc
 
+# Start PETSc profiling if required (-log_view not used)
+if "-log_view" not in " ".join(sys.argv):
```
wence-, line 22: Why not `"-log_view" not in sys.argv` ?

wence-:
> This fix is necessary because the [previous solution](https://github.com/firedrakeproject/firedrake/pull/2126), needed to stop
> PETSc complaining when the user passes `PETSc.Log.begin` themselves,
> would break nested profiling since `PETSc.Log.begin` calls
> `PetscLogDefaultBegin` internally instead of `PetscLogNestedBegin`.
> 
> When the `-log_view` flag is passed PETSc checks and calls the right
> `Begin` function hence why I've added this guard.
> 
> I have also observed that nested profiling does not work properly if the
> user writes `PETSc.Log.begin` in their code but I don't think we can do
> much about this.

I think I don't understand this logbegin stuff. What is the reason for us calling `PETSc.Log.begin()` at all?

AIUI, in a normal petsc program, if I run without `-log_view` then no `PetscLogXXXBegin` is called. If I run with `-log_view` then either `PetscLogDefaultBegin` or `PetscLogNestedBegin` is called.

Now it seems like Firedrake always calls `PetscLogDefaultBegin` (except if `-log_view` is passed). Why is this necessary?

connorjward:
> Now it seems like Firedrake always calls `PetscLogDefaultBegin` (except if `-log_view` is passed). Why is this necessary?

This was introduced to stop PETSc complaining since we start 2 events (import and overall) in `__init__.py`. When the user used `PETSc.Log.begin()` in their own code it would moan about some event mismatch. This is what Patrick encountered recently.

Thinking some more about it. A more sensible solution would be to check `sys.argv` and only start the 2 events if `-log_view` is present. We then don't need to touch `PETSc.Log.begin`. WDYT?

firedrake/__init__.py
connorjward: Thanks for spotting this. I was originally doing some more complex regex substring stuff so this can now go.

wence-:
> Thinking some more about it. A more sensible solution would be to check `sys.argv` and only start the 2 events if `-log_view` is present. We then don't need to touch `PETSc.Log.begin`. WDYT?

Really you want to start them in logging is enabled, I think? Does PETSc have a programmatic way of checking that?

Slightly safer than checking sys.argv (`PETSC_OPTIONS` is an environment variable that one can set and there are other things) would be to check `PETSc.Options().get("-log_view")` I think.

connorjward:
> Really you want to start them in logging is enabled, I think? Does PETSc have a programmatic way of checking that?

I can't see anything like that in the [manual](https://www.mcs.anl.gov/petsc/petsc-current/docs/manualpages/Profiling/index.html) and petsc4py definitely doesn't seem to have anything suitable.
I agree that that would be the ideal solution but I don't think it's doable.

> Slightly safer than checking sys.argv (PETSC_OPTIONS is an environment variable that one can set and there are other things) would be to check PETSc.Options().get("-log_view") I think.

I've just checked and

```python
from firedrake.petsc import OptionsManager

"log_view" in OptionsManager.commandline_options
```

appears to work.

wence- (approved):

wence-:
OK, makes sense, thanks.

PR #2149, opened 2021-07-14
Always provide mat_mumps_icntl_14: 200 as default

### LINKED ISSUE #2142: Should we set mat_mumps_icntl_14 even if pc_type is set by user?
Currently `"mat_mumps_icntl_14": 200` is part of the default solver parameters (along with preonly/lu+mumps).  However when the user sets `"pc_type"` themselves, the `pc_factor_mat_solver_type` and `mat_mumps_icntl` defaults are skipped. In the case of `pc_factor_mat_solver_type` I can see the logic for that: if the user sets `"pc_type": "lu"` they might reasonably expect the default (petsc) direct solver, so they have to explicitly ask for `"pc_factor_mat_solver_type": "mumps"` as well if they want mumps. However if they do so, they still do not get `"mat_mumps_icntl_14": 200` and this seems to frequently lead to issues - it does for me at least, random hangs in solvers that are not obvious to debug but are presumably just memory issues. I think the user explicitly asking for lu + mumps, despite it already being the default, happens relatively often, either because the user is just being explicit in a configuration, say in code with multiple alternative solver configurations, or because the setup was written before mumps was the default. As an example: https://www.firedrakeproject.org/demos/navier_stokes.py.html. I think it would be more helpful if we just always set `"mat_mumps_icntl_14": 200` (except when the user sets `"mat_mumps_icntl_14"` themselves of course).

stephankramer:
Even when user sets pc_type. This is to prevent memory issues for the
common case where the user explicitly sets
"pc_type": "lu" and "pc_factor_mat_solver_type": "mumps" despite this already
being the default.
Fixes #2142

### DIFF
--- a/firedrake/solving_utils.py
+++ b/firedrake/solving_utils.py
@@ -56,9 +56,8 @@ def set_defaults(solver_parameters, arguments, *, ksp_defaults={}, snes_defaults
         ksp_defaults = ksp_defaults.copy()
         skip = set()
         if "pc_type" in keys:
-            skip.update({"pc_factor_mat_solver_type", "mat_mumps_icntl_14",
-                         # Might reasonably expect to get petsc default
-                         "ksp_type"})
+            # Might reasonably expect to get petsc defaults
+            skip.update({"pc_factor_mat_solver_type", "ksp_type"})
         if parameters.get("mat_type") in {"matfree", "nest"}:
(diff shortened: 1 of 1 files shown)


wence- (approved):

wence-:
Thanks!

PR #2157, opened 2021-07-23
Fix form splitting of derivatives

wence-:
If we were taking a coordinate derivative, or the derivative direction
was a Function (rather than an Argument), the simplifications in the
handler for CoefficientDerivatives would just return Zero, rather than
an appropriate split thing. Fix by guarding the zero return more
tightly.


dham (commented):
Looks right.

PR #2161, opened 2021-07-29
Set the right communicators when building transfer matrices in p-MG.

pefarrell:
I tried to use `firedrake.P1PC` inside defcon and it hung. This patch fixes it.


pbrubeck (approved):
Looks alright to me, I can confirm that this is doing the right thing. Should we add tests for p-MG with multiple processors?

dham (approved):

PR #2164, opened 2021-07-30
Build scipy in the Docker container

pbrubeck:
An imminent PR for my FDM preconditioner #2024 , since it requires scipy to build sparse cell matrices. Scipy is also required already by one of the jupyter notebooks.


JDBetteridge (approved):
Thanks for adding this!
Assuming tests pass, this can go straight through

dham:
Tests on branches other than master don't build the docker containers, so I'm just going to merge this.

JDBetteridge:
Ah shoot, something broke!
Testing locally, it seems we need to install Cython earlier if we want to use the `--pip-install scipy` command line argument.
I have opened PR #2165  with a (potential) fix.

PR #2165, opened 2021-07-30
Install SciPy from firedrake-install

JDBetteridge:
To enable users to install `scipy` from the install script, we need to install `Cython` a bit earlier.

Added documentation to the download page on how to install `scipy` when invoking `firedrake-install` as well as documenting how to add `scipy` to an existing Firedrake venv.


pefarrell:
Looks good to me, although I expect the build system doesn't use this updated firedrake-install script. Jack, can you tell us how you tested this?

dham (approved):

PR #2167, opened 2021-08-04
Install scipy docs

JDBetteridge:
Split from #2165

### DIFF
--- a/scripts/firedrake-install
+++ b/scripts/firedrake-install
@@ -1662,6 +1662,12 @@ else:
         build_update_script()
         os.execv(sys.executable, [sys.executable, "../bin/firedrake-update", "--no-update-script"] + sys.argv[1:])
 
+    # Pre-install requested packages
+    if args.pip_packages is not None:
+        for package in args.pip_packages:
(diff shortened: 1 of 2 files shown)


scripts/firedrake-install
```diff
@@ -1662,6 +1662,12 @@ else:
         build_update_script()
         os.execv(sys.executable, [sys.executable, "../bin/firedrake-update", "--no-update-script"] + sys.argv[1:])
 
+    # Pre-install requested packages
+    if args.pip_packages is not None:
+        for package in args.pip_packages:
+            log.info("Pip installing %s to venv" % package)
+            run_pip_install(package.split())
```
dham, line 1669: Why is there code in this PR?

dham (changes requested):

scripts/firedrake-install
JDBetteridge: Prior to this PR the code to install additional packages with pip was only present in the "install" part of the install script, not the "update" part of the script. Running `firedrake-update --pip-install scipy` would have ignored the scipy package.

dham (approved):

PR #2168, opened 2021-08-04
fix warning from loopy about language version

danshapero:
Find all the remaining places where `loopy.make_function` is called without passing `(2018, 2)` for the `lang_version` keyword argument in order to silence some warnings from loopy. Also do a bit of reorganizing in the function `_get_lp_domains` -- the AST is generated in every branch of a 3-clause if/else when it could just be generated once at the very end of the conditional.

### DIFF
--- a/firedrake/slate/slac/utils.py
+++ b/firedrake/slate/slac/utils.py
@@ -16,7 +16,6 @@ import firedrake.slate.slate as sl
 import loopy as lp
 from loopy.transform.callable import merge
 from loopy.version import LOOPY_USE_LANGUAGE_VERSION_2018_2  # noqa: F401
-import itertools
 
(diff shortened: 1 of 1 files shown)


sv2518:
Hi. Thanks for taking care of this. Could we set the environment variable for the language version instead? So we'd include `from loopy.version import LOOPY_USE_LANGUAGE_VERSION_2018_2`at the beginning.

danshapero:
Yes that's much better, thanks for the suggestion! Do you want me to rebase down to 1 commit?

sv2518:
I think we can just squash-merge.

sv2518 (dismissed):

wence- (approved):

PR #2171, opened 2021-08-09
Build notebooks container

dham:
The notebooks container was omitted when we shifted to GitHub actions.


dham:
Note that the containers won't be tested on the PR because the branch is not master, but the draft PR passed with changes to  force the container to build on the branch.

PR #2172, opened 2021-08-10
New PyOP2 API

connorjward:
Related PRs:
https://github.com/firedrakeproject/tsfc/pull/257
https://github.com/OP2/PyOP2/pull/624

### DIFF
--- a/firedrake/assemble.py
+++ b/firedrake/assemble.py
@@ -1,65 +1,41 @@
+import abc
+from collections import OrderedDict
 import functools
+from functools import cached_property
+import itertools
+from itertools import product
 import operator
-from collections import OrderedDict, defaultdict, namedtuple
-from enum import IntEnum, auto
-from itertools import chain
 
+import cachetools
+import finat
 import firedrake
 import numpy
+from tsfc import kernel_args
+from tsfc.finatinterface import create_element
 import ufl
-from firedrake import (assemble_expressions, matrix, parameters, solving,
+from firedrake import (assemble_expressions, extrusion_utils as eutils, matrix, parameters, solving,
                        tsfc_interface, utils)
 from firedrake.adjoint import annotate_assemble
 from firedrake.bcs import DirichletBC, EquationBC, EquationBCSplit
+from firedrake.functionspacedata import entity_dofs_key, entity_permutations_key
 from firedrake.petsc import PETSc
 from firedrake.slate import slac, slate
-from firedrake.utils import ScalarType
+from firedrake.slate.slac.kernel_builder import CellFacetKernelArg, LayerCountKernelArg
+from firedrake.utils import ScalarType, tuplify
 from pyop2 import op2
 from pyop2.exceptions import MapValueError, SparsityFormatError
 
 
-__all__ = ("assemble",)
+__all__ = "assemble",
 
 
-class _AssemblyRank(IntEnum):
-    """Enum enumerating possible dimensions of the output tensor."""
-    SCALAR = 0
-    VECTOR = 1
-    MATRIX = 2
-
-
-class _AssemblyType(IntEnum):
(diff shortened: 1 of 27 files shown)


connorjward:
I think that this is now ready to be looked at and discussed.

### Summary of contributions
- Completely rewrote `assemble.py` to try and expose more caching opportunities.
- Modified some parts of `WithGeometry` and `MeshGeometry` to expose code generation information separately to any data structures. This should enable us to have stripped forms with Firedrake-specific information (e.g. are we extruded? are we a `ComponentFunctionSpace`?).
- Created a `FormAssembler` class that effectively reintroduces the 'fast path' `create_assembly_callable` that I deprecated a while ago.
- Replaced calls to `assemble` in our variational solvers with `FormAssembler` instances.
- `assemble(form, tensor=None)` is now nearly as fast as `assemble(form, tensor=something_persistent)`.

### Performance

I did a scaling experiment with the following setup:

```py
mesh = UnitSquareMesh(mesh_size, mesh_size)
v = TestFunction(V)
out = Function(V)
x = SpatialCoordinate(mesh)
f = interpolate(8*np.pi*np.pi*sin(2*x[0]*np.pi)*sin(2*x[1]*np.pi), V)
L = f * v * dx
```

where assembly is done in 3 ways:
- `notensor`: `assemble(L)`
- `tensor`: `assemble(L, tensor=out)`
- `assembler`: `OneFormAssembler(L, tensor=out)`

and here are the results:

plot

The key points here are:
- Using `OneFormAssembler` is much faster than calling `assemble`.
- The `notensor` version of `assemble` is faster than before.

### Lessons learnt
- I think the main strong scaling problems users are likely to encounter and really care about (i.e. the catastrophic ones) are due to us not caching generated code in the right places. In this work I've come to appreciated that globally caching the generated code is really the only way to avoid this problem in a robust way.
- Plugging the data from the form into the parloops is surprisingly expensive. Once the generated code is cached this then becomes the dominant scaling cost.

### Potential work for the future
- Since global caches of generated code can have such a big impact, someone (me) should probably go through Firedrake and add them where required. An example would be `compile_to_gem` in `assemble_expressions.py` (needed for `assign`).
- `assemble` has been optimised to the point where the main scaling costs come from Python itself. For example, at 1000 DoF, 20% of the time assembling is spent calling `global_to_local_begin` and other parallel routines even though the problem that I am running is serial.

dham (dismissed):

connorjward:
I've removed the CI tweaks so this is now ready to go in.

francesco-ballarin:
Hi @connorjward @dham
(cc @APaganini)
I belive that this PR introduces an [implicit requirement `python >= 3.8`](https://docs.python.org/3/library/functools.html#functools.cached_property). Is this desired or not?


I get the following error on my weekly built at [FEM on Colab](https://github.com/fem-on-colab/fem-on-colab). I am sticking to python 3.7 there because unfortunately Colab is still on python 3.7, and [cannot be easily upgraded to 3.8](https://github.com/googlecolab/colabtools/issues/1880).

```
/usr/local/lib/python3.7/dist-packages/firedrake/assemble.py in <module>
      2 from collections import OrderedDict
      3 import functools
----> 4 from functools import cached_property
      5 import itertools
      6 from itertools import product

ImportError: cannot import name 'cached_property' from 'functools' (/usr/lib/python3.7/functools.py)
```

connorjward:
Ah sorry that was not intentional. The necessary fix is to replace this with `pyop2.utils.cached_property`. I'll submit a fix ASAP.

francesco-ballarin:
Great, thanks! Ping me when you open the pull request and I will test it on my CI too :)

connorjward:
This should be fixed by https://github.com/firedrakeproject/firedrake/pull/2337.

PR #2190, opened 2021-08-20
Add CITATION.cff

IvanYashchuk:
I don't know if you'd like to use this feature or not, but GitHub has recently added "Cite this repository" button. For more details see https://docs.github.com/en/github/creating-cloning-and-archiving-repositories/creating-a-repository-on-github/about-citation-files.

This PR adds a `CITATION.cff` file. When present in the repository GitHub adds the "Cite this repository" button with generated APA and BibTeX citations.
image

Clicking "View citation file" opens the` CITATION.cff` file.
GitHub uses https://github.com/citation-file-format/ruby-cff to generate the APA and BibTeX. Here is the output:
```
Rathgeber, F., Ham, D. A., Lawrence, M., Lange, M., Luporini, F., Mcrae, A. T. T., Bercea, G., Markall, G. R., & Kelly, P. H. J. Firedrake: Automating the Finite Element Method by Composing Abstractions. ACM Trans. Math. Softw., 43(3), 1. https://doi.org/10.1145/2998441
```
```
@article{Rathgeber_Firedrake_Automating_the,
author = {Rathgeber, Florian and Ham, David A. and Lawrence, Mitchell and Lange, Michael and Luporini, Fabio and Mcrae, Andrew T. T. and Bercea, Gheorghe-Teodor and Markall, Graham R. and Kelly, Paul H. J.},
doi = {10.1145/2998441},
journal = {ACM Trans. Math. Softw.},
number = {3},
pages = {1--27},
title = {{Firedrake: Automating the Finite Element Method by Composing Abstractions}},
url = {http://arxiv.org/abs/1501.01809},
volume = {43}}
```
Unfortunately, it's not possible to change the reference tag of the generated BibTeX to match https://github.com/firedrakeproject/firedrake/blob/c0912a406323e742466cb77500abc8e15c4287f5/docs/source/_static/bibliography.bib#L281

"message" from CITATION.cff seems not to be used currently by GitHub, maybe this will change in the future, but still, I added a link to the citation webpage of Firedrake.

### DIFF
--- a/CITATION.cff
+++ b/CITATION.cff
@@ -1,0 +1,38 @@
+cff-version: 1.2.0
+message: |
+    "If you publish results using Firedrake, we would be grateful if you would cite it using the following metadata.
+    Please visit https://www.firedrakeproject.org/citing.html for extended instructions."
+title: "Firedrake"
+url: "https://github.com/firedrakeproject/firedrake"
+preferred-citation:
+  type: article
+  reference: Rathgeber2016
+  authors:
+    family-names: "Rathgeber"
+    given-names: "Florian"
+    family-names: "Ham"
+    given-names: "David A."
+    family-names: "Lawrence"
+    given-names: "Mitchell"
+    family-names: "Lange"
+    given-names: "Michael"
+    family-names: "Luporini"
+    given-names: "Fabio"
+    family-names: "Mcrae"
+    given-names: "Andrew T. T."
+    family-names: "Bercea"
+    given-names: "Gheorghe-Teodor"
+    family-names: "Markall"
+    given-names: "Graham R."
+    family-names: "Kelly"
+    given-names: "Paul H. J."
+  doi: "10.1145/2998441"
+  url: "http://arxiv.org/abs/1501.01809"
+  journal: "ACM Trans. Math. Softw."
+  start: 1 # First page number
+  end: 27 # Last page number
+  title: "Firedrake: Automating the Finite Element Method by Composing Abstractions"
+  volume: 43
+  issue: 3
+  year: 2016
+


bueler:
If you don't like the citation generated by the `.cff` then you can always just add the `.bib`.  I did this for my book because I want users of the software (= examples from the book) to cite the book not the examples repo.  But fiddling with the `.cff` `preferred-citation` mechanism did not yield a reasonable book citation.  So I reverted to adding a raw `CITATION.bib`.  Github recognizes that, adds a "cite this repo" link, and then hands the user the raw `CITATION.bib`.  Not ideal, but at least I fully control the citation.  See the result at https://github.com/bueler/p4pdes

dham (changes requested):
The message part of this seems right. I am nervous about providing the citation, because I am worried that  readers will just jump to just cite this.

arfon:
Note, GitHub now supports a range of different `preferred-citation` types: https://docs.github.com/en/repositories/managing-your-repositorys-settings-and-features/customizing-your-repository/about-citation-files#citing-something-other-than-software

IvanYashchuk:
Hey @arfon! Thanks for chiming in. Having different `preferred-citation` types supported is great but the `message` field from `CITATION.cff` is still not used by GitHub. For this project, it would be greatly beneficial to guide readers to the website with instructions on citing different components of the software.
```
message: |
    "If you publish results using Firedrake, we would be grateful if you would cite it using the following metadata.
    Please visit https://www.firedrakeproject.org/citing.html for extended instructions."
```
It would be great to be able to customize the default "If you use this software in your work, please cite it using the following metadata." message after clicking the "Cite this repository" button.

arfon:
Ah right. Yes I'm afraid the message can't be changed at this point sorry.

wence-:
My conclusion is that the automatic citation mechanism offered by `CITATION.cff` isn't currently suitable for Firedrake. The set of papers that one should cite for a given simulation depends on the particular configuration of the simulation (and Firedrake/PETSc's runtime `-citations` flag attempts to tell you these).

TBH: I think this is going to be generically true for any large piece of research software (especially those with dependencies), best practice will be to cite multiple papers dependent on the exact composition of algorithms employed. I don't think that a static file can capture this information (which is why the instructions for citing Firedrake basically say "we determine it for you at runtime").

I'm therefore inclined to close this. I think on the repository, the right thing would be for the "cite this repository" button to allow us just to drop arbitrary text with a link to somewhere, rather than trying to provide a static citation.

arfon:
> I'm therefore inclined to close this. I think on the repository, the right thing would be for the "cite this repository" button to allow us just to drop arbitrary text with a link to somewhere, rather than trying to provide a static citation.

Sounds reasonable. Please note, any of the [files listed here](https://docs.github.com/en/repositories/managing-your-repositorys-settings-and-features/customizing-your-repository/about-citation-files#other-citation-files) will generate the citation prompt in the GitHub UI (you can then have whatever you want written in the linked file).

wence-:
@IvanYashchuk do you want to have a go at that last suggestion instead of `citations.cff`? That is, right `CITATION.md` that just points at the relevant firedrake webpage. Do we think that is something we want?

IvanYashchuk:
Yes, I could replace `CITATION.cff` with `CITATION.md` that simply points to the webpage. I'll open a new PR for that and close this one.

PR #2193, opened 2021-08-23
Use modern protocol for caching TSFC kernels

connorjward:
This was causing me some trouble when I was tinkering with the TSFC kernels. With this change pickling is now less restrictive about what the cache stores.

### DIFF
--- a/firedrake/tsfc_interface.py
+++ b/firedrake/tsfc_interface.py
@@ -8,9 +8,7 @@ import pickle
 
 from hashlib import md5
 from os import path, environ, getuid, makedirs
-import gzip
 import os
-import zlib
 import tempfile
 import collections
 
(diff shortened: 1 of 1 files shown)


connorjward:
OK this is more complicated than I thought. Things now seem to be failing because calling `pickle.load(f)` then calls `TSFCKernel.__new__` but this will not work since `args` and `kwargs` are no longer defined.

The process is effectively:

```
__new__(*args, **kwargs) -> _cache_lookup -> _read_from_disk -> pickle.load(f) -> __new__()
```

The [pickle documentation](https://docs.python.org/3/library/pickle.html#object.__getnewargs_ex__) says that we should use `__getnewargs_ex__` to resolve this.

Needless to say this is more difficult than I was hoping and I can definitely find easier workarounds. Closing.

PR #2199, opened 2021-08-31
Annotate LinearSolver

jwallwork23:
When adding Pyadjoing tags to Thetis solves using `ad_block_tag` in [PR261](https://github.com/thetisproject/thetis/pull/261), I noticed that `LinearSolver` is not annotated. Hopefully this should do the trick.


dham (changes requested):
I think we need a test.

jwallwork23:
I have something that works, but only if the matrix stays fixed. Needs more work and then will open a new PR.

PR #2202, opened 2021-09-02
Remove `Expression` objects, and adapt to interpolation API changes

wence-:


docs/source/variational-problems.rst
```diff
@@ -499,8 +499,8 @@ for the various :ref:`utility meshes <utility_mesh_functions>` are
 described in their respective constructor documentation.  For
 externally generated meshes, Firedrake just uses whichever ids the
 mesh generator provided.  The ``value`` may be either a scalar, or
-more generally an :py:class:`~.Expression`, :py:class:`~.Function` or
-:py:class:`~.Constant` of the appropriate shape.  You may also supply
+more generally a UFL expression, for example and :class:`~.Function` or
+:py:class:`~.Constant`, of the appropriate shape.  You may also supply
```
ReubenHill, line 503: Everywhere you've changed the docs to only say "UFL expression" you might as well say 
```
for example and :class:`~.Function` or :py:class:`~.Constant`
```
for completeness.
We don't want people to look at the API documentation for something like interpolation and think they can only interpolate pure UFL.

firedrake/interpolation.py
```diff
@@ -289,30 +273,24 @@ def _interpolator(V, tensor, expr, subset, arguments, access):
...
-        first_coeff_fake_coords = False
-    else:
-        raise RuntimeError("Attempting to evaluate an Expression which has no value.")
-
+    # We need to pass both the ufl element and the finat element
+    # because the finat elements might not have the right mapping
+    # (e.g. L2 Piola, or tensor element with symmetries)
+    # FIXME: for the runtime unknown point set (for cross-mesh
+    # interpolation) we have to pass the finat element we construct
+    # here. Ideally we would only pass the UFL element through.
```
ReubenHill, line 281: I think all the `to_element` creation (and any fiddling with `rebuild`) in `firedrake/interpolation.py` and `firedrake/preconditioners/pmg.py` can be absorbed into `compile_expression_dual_evaluation` as long as we work out how to tell `compile_expression_dual_evaluation` that it needs to make a runtime unknown point set for cross-mesh interpolation.

PR #2205, opened 2021-09-02
improve BLAS threads checking

danshapero:
The current code in `__init__.py` checks for the presence of OpenBLAS and sets the number of threads to 1 in order to prevent awful performance regressions. This does the right thing in some setups but not others:

1. You have regular old Netlib BLAS.
2. You're linking to the Intel MKL library for BLAS.
3. You installed OpenBLAS via the `--with-blas=download` option when building PETSc.

You still get the warning even in the last case because the current code uses `find_library` to look for OpenBLAS, which only looks in places like `/usr/lib/`, and not in the PETSc build tree. This patch handles the other cases as well by parsing the output of either `ldd` or `otool` (platform depending) on `firedrake.petsc.PETSc.__file__`.

I've tested it on my Mac and in Docker images with Netlib BLAS, Intel MKL, OpenBLAS installed via apt-get, and OpenBLAS installed via PETSc.

### DIFF
--- a/firedrake/petsc.py
+++ b/firedrake/petsc.py
@@ -1,5 +1,6 @@
 # Utility module that imports and initialises petsc4py
 import os
+import subprocess
 import petsc4py
 import sys
 petsc4py.init(sys.argv)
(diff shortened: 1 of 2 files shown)


firedrake/petsc.py
```diff
@@ -95,6 +96,28 @@ def get_petsc_variables():
...
+    # Linux uses `ldd` to look at shared library linkage, MacOS uses `otool`
+    try:
+        program = ["ldd"]
+        cmd = subprocess.run([*program, PETSc.__file__], stdout=subprocess.PIPE)
+    except FileNotFoundError:
+        program = ["otool", "-L"]
+        cmd = subprocess.run([*program, PETSc.__file__], stdout=subprocess.PIPE)
+
+    entries = cmd.stdout.decode("utf-8").split("\n")
+    entry = next((e for e in entries if "blas" in e or "libmkl" in e), None)
```
dham, line 110: Less magic please!

dham (changes requested):

danshapero:
[dockerfiles.zip](https://github.com/firedrakeproject/firedrake/files/7100344/dockerfiles.zip)

Here are the dockerfiles I used to test with Netlib BLAS and MKL.

firedrake/petsc.py
danshapero: Magic expunged!

wence- (approved):
Looks good!

wence-:
Thanks!

PR #2215, opened 2021-09-13
M1 fixes

### LINKED ISSUE #2196: Firedrake build on mac M1
In order to build to build firedrake on mac M1, some linker modifications are required. The issues appear to be related to gcc-11 on m1 and how the paths are setup. When simply running firedrake-install script, clang linker is unable to find `libgcc_s.1.1.1.dylib`.  This maybe due to the way gcc-11 is built with rpath not fully defined e.g. 

`otool -L /opt/homebrew/opt/gcc/lib/gcc/11/libgfortran.5.dylib` gives 
```
/opt/homebrew/opt/gcc/lib/gcc/11/libgfortran.5.dylib (compatibility version 6.0.0, current version 6.0.0)
@rpath/libgcc_s.1.1.dylib (compatibility version 1.0.0, current version 1.1.0)
/usr/lib/libSystem.B.dylib (compatibility version 1.0.0, current version 1292.100.5)
```

For the moment, the way to build firedrake on m1 is by modifying firedrake-install script (line 620) to:

```
elif osname == "Darwin":
    brew_prefix = check_output(['brew', '--prefix'])[:-1]
    brew_gcc_prefix = brew_prefix+"/opt/gcc/lib/gcc/11" 
    mac_blas = brew_prefix+"/opt/openblas"
    petsc_options.add("--with-blaslapack-dir={}".format(mac_blas))
    petsc_options.add("--CFLAGS=-I{}/include -Wno-implicit-function-declaration".format(mac_blas))
    petsc_options.add("--LDFLAGS=-Wl,-rpath,{0}/lib -L{0}/lib -L{1}".format(mac_blas, brew_gcc_prefix))
```

This allows firedrake to build. The tests, however, will still fail to build. For that, need to remove `-march=native` option from `PyOp2/pyop2/compilation.py` script line 372.

danshapero:
Resolves #2196 (thanks to omarjamil for figuring this out).

Also, the code that I had added to find the BLAS library that PETSc links to didn't work properly on MacOS 11.x. In MacOS 10.x, the PETSc extension library for Python (`PETSc.cpython-blah-blah.dylib`) directly links to BLAS and all the other transitive dependencies of PETSc. In 11.x this is no longer true and we have to instead have to look for BLAS among all the shared object dependencies of the PETSc C library (`libpetsc.dylib`). This patch fixes this problem as well.

### DIFF
--- a/scripts/firedrake-install
+++ b/scripts/firedrake-install
@@ -618,12 +618,14 @@ elif options.get("with_blas") is not None:
     else:
         petsc_options.add("--CFLAGS=-I{}/include".format(options["with_blas"]))
 elif osname == "Darwin":
-    brew_prefix = check_output(['brew', '--prefix'])[:-1]
-    brew_gcc_prefix = brew_prefix+"/opt/gcc/lib/gcc/11"
-    mac_blas = brew_prefix+"/opt/openblas"
-    petsc_options.add("--with-blaslapack-dir={}".format(mac_blas))
-    petsc_options.add("--CFLAGS=-I{}/include -Wno-implicit-function-declaration".format(mac_blas))
-    petsc_options.add("--LDFLAGS=-Wl,-rpath,{0}/lib -L{0}/lib -L{1}".format(mac_blas, brew_gcc_prefix))
+    brew_gcc_prefix = check_output(["brew", "--prefix", "gcc"])[:-1]
+    brew_gcc_info = json.loads(check_output(["brew", "info", "--json", "gcc"])[:-1])
+    gcc_major_version = brew_gcc_info[0]["installed"][0]["version"].split(".")[0]
+    brew_gcc_libdir = brew_gcc_prefix + "/lib/gcc/" + gcc_major_version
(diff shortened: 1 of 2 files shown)


scripts/firedrake-install
```diff
@@ -619,10 +619,11 @@ elif options.get("with_blas") is not None:
         petsc_options.add("--CFLAGS=-I{}/include".format(options["with_blas"]))
 elif osname == "Darwin":
     brew_prefix = check_output(['brew', '--prefix'])[:-1]
+    brew_gcc_prefix = brew_prefix+"/opt/gcc/lib/gcc/11"
```
dham, line 622: Should the 11 here really be hard-coded? What happens when that changes?

dham (changes requested):

scripts/firedrake-install
danshapero: That's a great point -- we can parse the output from `brew info --json gcc` or does that seem too complicated?

danshapero: Alright it turned out to be super easy, this is ready for another look.

dham (approved):

PR #2218, opened 2021-09-16
Firedrake21 - recorded talk

sv2518:
Include information about recorded talks on webpage of the event.


sv2518:
This only works if all videos are uploaded with "Firedrake '21" in their title, I assume that is correct?

wence-:
Can one set up a vimeo "list" instead?

sv2518:
Potentially, but I assume only David can do that, because he is the one who uploaded them.

sv2518:
Okay, @dham, the alternative is that you generate something called a "showcase", which is basically a list, I think. They say "A showcase is a great way to distribute a collection of videos to any audience." Up to you. If you want to do that just send me the link to it and I can change the PR here.

PR #2227, opened 2021-09-23
Fix parameter handling in Slate compiler

sv2518:
I have screwed up when we merged the optimisation for blocks in Slate in this PR https://github.com/firedrakeproject/firedrake/pull/2111. I have changed the way parameters are handled at the end and introduced a bug.

The testing of blocks on master only tests optimised vs. optimised tests so they would never fail, because the assembly parameter, which specifies that an expression should not be optimised, does not reach the Slate compiler.

I am changing this in a minimally invasive way in this PR, where I treat `form_compiler_parameters` as
`form_compiler_parameters = {tsfc_param1, tsfc_param2, "slate_compiler": {slate_param1, ... }}`

The second option would be to pass everywhere `form_compiler_parameters` AND `slate_compiler_parameters`. This involves a lot of interface changes (especially in assembly), so I opted against it.

The third option would be to handle the parameters as
`compiler_parameters = {"form_compiler": {tsfc_param1, ...}, "slate_compiler": {slate_param1, ... }}`. This also needs changes wherever we pass the `form_compiler_parameters` to assemble, so we need to touch a lot of tests e.g.

So I am in favour of option one, but other opinions are welcome.

### DIFF
--- a/firedrake/assemble.py
+++ b/firedrake/assemble.py
@@ -695,7 +695,7 @@ def _make_parloops(expr, tensor, bcs, diagonal, fc_params, assembly_rank):
     if isinstance(expr, slate.TensorBase):
         if diagonal:
             raise NotImplementedError("Diagonal + slate not supported")
-        kernels = slac.compile_expression(expr, tsfc_parameters=form_compiler_parameters)
+        kernels = slac.compile_expression(expr, compiler_parameters=form_compiler_parameters)
     else:
         kernels = tsfc_interface.compile_form(expr, "form", parameters=form_compiler_parameters, diagonal=diagonal)
 
--- a/firedrake/slate/slac/compiler.py
+++ b/firedrake/slate/slac/compiler.py
@@ -99,7 +99,7 @@ class SlateKernel(TSFCKernel):
         self._initialized = True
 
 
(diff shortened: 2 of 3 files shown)


connorjward:
> The second option would be to pass everywhere `form_compiler_parameters` AND `slate_compiler_parameters`. This involves a lot of interface changes (also in PyOP2), so I opted against it.

Personally this would be my preference as it's the most explicit. If we want to ignore the new arguments existence in a function we can always pass `**kwargs`. Also, how would this affect PyOP2? I don't think I've seen any point where PyOP2 touches `form_compiler_parameters`.

sv2518:
`I don't think I've seen any point where PyOP2 touches form_compiler_parameters.`
Ah my bad, this was actually in the assembly in firedrake. We'd e.g. need to change the assembly options here https://github.com/firedrakeproject/firedrake/blob/9cf9b75d1a961354e4738d5d5f045e5539759490/firedrake/assemble.py#L42 and the interface of make_parloops here https://github.com/firedrakeproject/firedrake/blob/9cf9b75d1a961354e4738d5d5f045e5539759490/firedrake/assemble.py#L641

dham (approved):

PR #2233, opened 2021-09-27
New Slate optimisation pass

sv2518:
This PR introduces a new optimisation pass for the Slate compiler.

In the new optimisation pass Slate expressions are resorted so that the corresponding generated code contains a minimal amount of matrix-shaped temporaries. This is useful for high order FEM where the local assembly tensors are big. By using less matrix-shaped temporaries we reduce the amount of data transfer, which is a bottleneck for high-order FEM. Note that the tradeoff is the execution of more FLOPS.

The resorting works by pushing the multiplication of a matrix with a vector as far inside an expression as possible. For example if we have an expressions `(A+B)*f`, where A,B are `Tensors` of rank 2 and f is an `AssembledVector`, then the new compiler pass resorts the expression to `A*f+B*f`. In the compiler pass we also replace all inverses with solves with help of `A.inv*f=A.solve(f)`.

While I want this compiler pass so that I can replace the multiplication in `A*f+B*f` by Actions later on, we should already benefit from this optimisation without that replacement. That is so because: 
1) E.g. `(A+B)*f` theoretically builds 3 matrix-shaped temporaries A, B and A+B, while `A*f+B*f` only builds A and B. BUT in the generated C code the expression is inlined (see C code below) and I don't exactly know what the C compiler is doing under the hood with that. I am afraid it is smart enough to optimise this expression, so that we maybe only really benefit from this optimisation step WITH the replacement of the multiplication by an Action. I have looked at the assembly code for this example, but I have no experience with it and don't understand it well enough to understand if what I say is true. I have timed the difference of the optimised and non-optimised expression for a FE discretisation that results in local tensors of size 400x400, but the timings are almost the same.
2) We certainly benefit of replacing inverse operations with solve operations. The assembly of the optimised expression on a warm cache takes a fraction of the non-optimised expression in this case.

An example of the generated C code is 
Not optimised expression `(A+B)*f`:
```
static void slate_loopy(double *__restrict__ output, double const *__restrict__ T0, double const *__restrict__ T1)
{
  for (int32_t i = 0; i <= 399; ++i)
    for (int32_t i_0 = 0; i_0 <= 399; ++i_0)
      output[i] = output[i] + (T0[400 * i + i_0] + T0[400 * i + i_0]) * T1[i_0];
}
```
Optimised expression `A*f+B*f`
```
static void slate_loopy(double *__restrict__ output, double const *__restrict__ T0, double const *__restrict__ T1)
{
  double t0;

  for (int32_t i = 0; i <= 399; ++i)
  {
    t0 = 0.0;
    for (int32_t i_0 = 0; i_0 <= 399; ++i_0)
      t0 = t0 + T0[400 * i + i_0] * T1[i_0];
    output[i] = output[i] + t0 + t0;
  }
}
```

In addition to the multiplication pass, I also introduce a transpose optimisation pass, which rewrites `A.T.T->A` wherever possible.

### DIFF
--- a/firedrake/slate/slac/optimise.py
+++ b/firedrake/slate/slac/optimise.py
@@ -133,6 +133,57 @@ def push_mul(tensor, options):
     return a
 
 
+def drop_double_transpose(expr):
+    """Remove double transposes from optimised Slate expression."""
+    from gem.node import Memoizer
+    mapper = Memoizer(_drop_double_transpose)
+    a = mapper(expr)
+    return a
+
+
+@singledispatch
+def _drop_double_transpose(expr, self):
+    raise AssertionError("Cannot handle terminal type: %s" % type(expr))
+
+
+@_drop_double_transpose.register(Tensor)
+@_drop_double_transpose.register(AssembledVector)
+@_drop_double_transpose.register(Block)
+def _drop_double_transpose_terminals(expr, self):
+    """Terminal expression is encountered."""
+    return expr
+
+
+@_drop_double_transpose.register(Factorization)
+def _drop_double_transpose_factorization(expr, self):
+    """Drop any factorisations. This is needed because whenever
+    a Solve node is created so is a Factorization node inside it.
+    The Factorization nodes are only needed for the Eigen Backend however."""
+    return self(*expr.children)
+
+
+@_drop_double_transpose.register(Transpose)
+def _drop_double_transpose_transpose(expr, self):
+    """When the expression and its child are transposes the grandchild is returned,
+    because A=A.T.T."""
+    child, = expr.children
+    if isinstance(child, Transpose):
+        grandchild, = child.children
+        return self(grandchild)
+    else:
+        return type(expr)(*map(self, expr.children))
+
+
+@_drop_double_transpose.register(Negative)
+@_drop_double_transpose.register(Add)
+@_drop_double_transpose.register(Mul)
+@_drop_double_transpose.register(Solve)
+@_drop_double_transpose.register(Inverse)
+def _drop_double_transpose_distributive(expr, self):
+    """Distribute into the children of the expression. """
+    return type(expr)(*map(self, expr.children))
+
+
 @singledispatch
 def _push_mul(expr, self, state):
     raise AssertionError("Cannot handle terminal type: %s" % type(expr))
--- a/firedrake/slate/slac/optimise.py
+++ b/firedrake/slate/slac/optimise.py
@@ -1,17 +1,43 @@
 from gem.node import MemoizerArg
 from functools import singledispatch
 from itertools import repeat
-import firedrake.slate.slate as sl
+from firedrake.slate.slate import *
+from collections import namedtuple
+
+""" ActionBag class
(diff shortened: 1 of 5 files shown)


firedrake/slate/slac/optimise.py
```diff
@@ -83,3 +104,254 @@ def _push_block_block(expr, self, indices):
...
+@_drop_double_transpose.register(AssembledVector)
+@_drop_double_transpose.register(Block)
+def _drop_double_transpose_terminals(expr, self):
+    """Terminal expression is encountered."""
+    return expr
+
+
+@_drop_double_transpose.register(Factorization)
+def _drop_double_transpose_factorization(expr, self):
+    """ Drop any factorisations. """
```
dham, line 154: Explain why this is happening here.

firedrake/slate/slac/optimise.py
```diff
@@ -83,3 +104,254 @@ def _push_block_block(expr, self, indices):
...
+@_push_mul.register(Tensor)
+@_push_mul.register(Block)
+def _push_mul_tensor(expr, self, state):
+    if not self.action:
+        if state.coeff:
+            return Mul(expr, state.coeff) if state.pick_op == 1 else Mul(state.coeff, expr)
+        else:
+            return expr
+    else:
+        assert "Actions in Slate are not yet supported."
```
dham, line 194: `assert False,`

firedrake/slate/slac/optimise.py
```diff
@@ -83,3 +104,254 @@ def _push_block_block(expr, self, indices):
...
+        Another example for  expr:=C*A.solve(B.T) : (C*A.solve(B.T)).T * y = (y.T * (C*A.solve(B.T))).T
+
+        :arg expr: a Transpose
+        :arg self: a MemoizerArg object.
+        :arg state: state carries a coefficient in .coeff,
+                    and information if multiply from front (0) or back (1) in .pick_op
+        :returns: an optimised transposed expression
+    """
+    if expr.rank == 2:
+        transposed_state = ActionBag(Transpose(state.coeff), state.pick_op ^ 1)  # switch the multiplication order with pick_op
```
dham, line 233:
```
(state.pick_op + 1) % 2
```

firedrake/slate/slac/optimise.py
```diff
@@ -83,3 +104,254 @@ def _push_block_block(expr, self, indices):
...
+                                    /    |                    /    |
+                                op1      op2                y.T   op1
+
+        :arg expr: a Multiplication
+        :arg self: a MemoizerArg object.
+        :arg state: state carries a coefficient in .coeff,
+                    and information if multiply from front (0) or back (1) in .pick_op
+        :returns: an optimised Multiplication
+    """
+    if expr.rank == 1:
```
dham, line 332: line length.

dham (changes requested):

sv2518:
Okay, I addressed all comments. The failing test was caused because I screwed up cherry-picking the changes from my other PR (the one that we merged last week.) So this is resolved now.

PR #2237, opened 2021-10-01
Nested schur complements in hybridisation

sv2518:
Just testing if I don't break anything.


firedrake/slate/static_condensation/hybridization.py
```diff
@@ -324,6 +336,49 @@ def _reconstruction_calls(self, split_mixed_op, split_trace_op):
                                                form_compiler_parameters=self.ctx.fc_params,
                                                assembly_type="residual")
 
+    def build_schur(self, Atilde, K, list_split_mixed_ops, list_split_trace_ops, nested=True):
```
dham, line 339: Default to `False`

firedrake/slate/static_condensation/hybridization.py
```diff
@@ -199,17 +200,35 @@ def initialize(self, pc):
...
+        list_split_mixed_ops = [A00, A01, A10, A11]
+
+        split_trace_op = dict(split_form(K.form))
+        K0 = Tensor(split_trace_op[(0, id0)])
+        K1 = Tensor(split_trace_op[(0, id1)])
+        list_split_trace_ops = [K0, K1]
+
+        # Build schur complement operator and right hand side
+        nested = (prefix + "nested_schur" in PETSc.Options().getAll().keys()
+                  and PETSc.Options().getString(prefix + "nested_schur", "true") == 'true')
```
wence-, line 219: I think `PETSc.Options(prefix).getBool("nested_schur", False)` is probably simpler.

wence-: But maybe it doesn't work...?

sv2518:
OK. Addressed the requested changes now and rebased on top of master.

PR #2239, opened 2021-10-04
Fix syntax for MKL_Set_Num_Threads

ScottMacLachlan:
Per [this link](https://stackoverflow.com/questions/28283112/using-mkl-set-num-threads-with-numpy), mkl_set_num_threads is a Fortran-style (pass by reference) routine, while MKL_Set_Num_Threads is the C-style (pass by value) routine that we're expecting to call.

Before this fix, I couldn't run `import firedrake` when using MKL.


wence-:
Thanks, looks right.

PR #2240, opened 2021-10-05
User-defined preconditioning of local solver in reconstruction calls in HybridisationPC 

sv2518:
Just testing if I break things


dham (approved):
Subject to upstream changes landing.

PR #2249, opened 2021-10-13
Yet more M1 fixes

### LINKED ISSUE #2243: installation failed in macOS Big Sur
[Hello,](url) I am installing firedrake and encounter the following issue:

--
  File "firedrake-install", line 623, in <module>
    gcc_major_version = brew_gcc_info[0]["installed"][0]["version"].split(".")[0]
IndexError: list index out of range
--

here's what is appearing in installation log:

2021-10-08 12:47:00,290 INFO   Running firedrake-install
2021-10-08 12:47:00,291 DEBUG  Running command 'brew --prefix gcc'
2021-10-08 12:47:00,375 DEBUG  Running command 'brew info --json gcc'
2021-10-08 12:47:04,374 INFO   

can some please help me fix this? Thanks.

danshapero:
Fixes #2243. Tested on my Mac by first running `brew uninstall --force $(brew list)` so that there's no GCC or OpenBLAS and then reinstalling Firedrake.

The problem that I introduced before was that the script assumes that Homebrew has installed GCC in order to make a list of PETSc configure options, when in fact Homebrew will only install GCC later in the install process than when the PETSc options are created. What I did here with suggestions from @dham was to factor out the PETSc options making into a function that gets called only when it's needed.

### DIFF
--- a/scripts/firedrake-install
+++ b/scripts/firedrake-install
@@ -539,106 +539,109 @@ If you really want to use your own Python packages, please run again with the
 --honour-pythonpath option.
 """)
 
-petsc_options = {"--with-fortran-bindings=0",
-                 "--with-debugging=0",
-                 "--with-shared-libraries=1",
-                 "--with-c2html=0",
-                 "--download-eigen=%s/src/eigen-3.3.3.tgz " % firedrake_env,
-                 # File format
-                 "--download-hdf5",
-                 # AMG
-                 "--download-hypre",
-                 # Sparse direct solver
-                 "--download-superlu_dist",
-                 # Parallel mesh partitioner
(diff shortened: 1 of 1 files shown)


connorjward (approved):
Looks fine I think. Moving any part of the install script into a function can only really be a good thing!

dham (changes requested):
Thanks @danshapero. I think the one remaining issue is that the hard dependency on HomeBrew on Mac is still implicit, so a user who hasn't installed homebrew will just get a command not found error at a random point. Can we have an explicit early fail with a helpful error message if the install script is run on a Mac with no homebrew?

PR #2258, opened 2021-10-25
A function which plots local operators and spills out some information about them as well

sv2518:
This is a function I keep using to inspect properties of Slate tensors. Do we want to make this available for everyone? So far this function is not tested (have not thought about how I could do that yet) and I am not sure what happens if you have got boundary conditions. I also didn't think about performance of this function at all.
It's useful e.g. when you want to look at some of the Slate tensors within the hybridisation preconditioner.


sv2518:
I have never finished this one because I didn't know if anyone wants it. I know that Pablo has been using it recently. @colinjcotter do you think this would be useful?

PR #2269, opened 2021-11-04
Install: pin setuptools to 58.4 for now

sv2518:
Workaround for https://github.com/pypa/setuptools/issues/2849


sv2518:
Ok, looks like this works to fix master. Please merge!

wence- (approved):
Thanks!

PR #2271, opened 2021-11-04
install: Fix determination of BLAS location on MacOS

wence-:
Previously we shoved the location in config["with_blas"] if we were
asking firedrake-install to install openblas. That didn't update the
blas_location variable that is inspected later to set up the BLAS
environment config, so we would never actually use the information we
determined. Log this information so that we can check things.

In addition, since the MacOS install unconditionally requires
homebrew (even with --no-package-manager), automatically determine the
BLAS location even when --no-package-manager is passed. Error early if
this does not resolve to an actual installed BLAS.

### DIFF
--- a/scripts/firedrake-install
+++ b/scripts/firedrake-install
@@ -1438,14 +1438,17 @@ if mode == "install" or not args.update_script:
                 else:
                     log.info("Installing required package '{0}' via Homebrew.".format(package))
                     brew_install(package)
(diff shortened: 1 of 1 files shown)


wence-:
I claim this fixes the issues some macos users were seeing with firedrake-install and numpy (again).

dham (approved):

Issue #2262, opened 2021-10-29
Slate codegen improvement


sv2518:
We should optimise code generation for transposes on Slate tensors by inlining the reversed indices into the local assembly kernel.

Lawrence words:

> Right now, a Tensor(form) is translated by the slate compiler into something "opaque" I think. This is because compile_form spits back a loopy kernel object.
> 
> But think about what Transpose(Tensor(form)) does, it calls the generated loopy kernel to populate a tensor A (say), and then makes more gem that is CT(A[i, j], (j, i)).
> 
> If the gem that creates A were still around, we could inline that indexing transpose right into the kernel from tsfc. This would be morally equivalent to turning Transpose(Tensor(form)) into compile_form(form, transpose=True) (if compile_form had a transpose operation).
> 
> Perhaps this is already done because Transpose(Tensor(form)) can be turned into Tensor(adjoint(form)) and maybe you do this right now? 

So what this is saying is that rather than:
```

1) gem_expr_for_form <- TSFC generates GEM from ufl form associated with Slate tensor
2) loopy_kernel_for_form <- TSFC generate loopy from gem_expr_for_form
3) gem_expr_for_slateops <- Slate compiler generates GEM for Slate operations
4) loopy_kernel_for_slateops <- TSFC generates loopy from gem_expr_for_slateops
```

where DiagonalTensors and Transposes are dealt with in step 3, we want to do

```
1) gem_expr_for_form <- TSFC generates GEM also from all modified terminals (Diagonals and Transposes e.g.)
2) loopy_kernel_for_form <- TSFC generate loopy from gem_expr_for_form
3) gem_expr_for_slateops <- Slate compiler generates GEM for Slate operations
4) loopy_kernel_for_slateops <- TSFC generates loopy from gem_expr_for_slateops
```
where `DiagonalTensors` and `Transposes` and so forth are dealt with in step 1 and step 3) and 4) only apply if there are more Slate operations on top of transposes.

PR #2272, opened 2021-11-08
Slate: Transpose(Tensor(form)) -> Tensor(adjoint(form))

sv2518:
This PR addresses https://github.com/firedrakeproject/firedrake/issues/2262. It's a small codegen improvement which inlines transposes into the local assembly kernels generated by TSFC. I have done this on the highest level (in Slate) via `def __new__`.

### DIFF
--- a/firedrake/slate/slac/optimise.py
+++ b/firedrake/slate/slac/optimise.py
@@ -3,6 +3,7 @@ from functools import singledispatch
 from itertools import repeat
 from firedrake.slate.slate import *
 from collections import namedtuple
+from firedrake.ufl_expr import adjoint
 
 """ ActionBag class
 :arg coeff:     what we contract with.
@@ -235,6 +236,8 @@ def _drop_double_transpose_transpose(expr, self):
     if isinstance(child, Transpose):
         grandchild, = child.children
         return self(grandchild)
+    elif child.terminal and child.rank > 1:
+        return Tensor(adjoint(child.form))
     else:
         return type(expr)(*map(self, expr.children))
 
--- a/tests/slate/test_optimise.py
+++ b/tests/slate/test_optimise.py
@@ -328,6 +328,13 @@ def test_drop_transposes(TC_non_symm):
     compare_vector_expressions(expressions)
     compare_slate_tensors(expressions, opt_expressions)
 
+    assert A != A.T
+    from firedrake.slate.slac.optimise import optimise
+    T_opt = optimise(A.T, {"optimise": True})
+    assert isinstance(T_opt, Tensor)
+    assert np.allclose(assemble(T_opt.T).M.values,
+                       assemble(adjoint(T_opt.form)).M.values)
+
 
 #######################################
 # Test diagonal optimisation pass


wence-:
Looks like there's an actual failure in the hybridisation notebook?

firedrake/slate/slate.py
```diff
@@ -1024,6 +1024,11 @@ def _output_string(self, prec=None):
 class Transpose(UnaryOp):
     """An abstract Slate class representing the transpose of a tensor."""
 
+    def __new__(cls, A):
+        if A.terminal:
+            return Tensor(adjoint(A.form))
+        return super().__new__(cls)
```
wence-, line 1030: Seems like a reasonable way to do it.

tests/slate/test_linear_algebra.py
```diff
@@ -116,3 +116,36 @@ def test_local_solve(decomp):
...
+    n = FacetNormal(mesh)
+    u = split(w)[0]
+    un = abs(dot(u('+'), n('+')))
+    jump_v = v('+')*n('+') + v('-')*n('-')
+    jump_T = T('+')*n('+') + T('-')*n('-')
+    x, y = SpatialCoordinate(mesh)
+
+    T = Tensor(-dot(u*T, grad(v))*dx + (dot(u('+'), jump_v)*avg(T))*dS + dot(v, dot(u, n)*T)*ds + 0.5*un*dot(jump_T, jump_v)*dS)
+    assert T != T.T
+    assert isinstance(T.T, Tensor)
```
wence-, line 151: Can you test that we get the right result? e.g. by comparing `assemble(T.T)` with `assemble(adjoint(T.form))` ?

wence- (commented):
Other than the actual test failure in the build, I think this is good.

tests/slate/test_linear_algebra.py
sv2518: Ok done.

sv2518:
Ok I think some tests were failing, because I took the ufl adjoint and not the one from `firedrake.ufl_expr`.

wence-:
Now it looks like we have many fails because `adjoint(linear_form)` isn't defined, whereas `Tensor(linear_form).T` does work, I think?

sv2518:
Mhh, maybe this is not the right way to do it after all. Maybe I need to do it at a lower level. The leftover issues we have are that a) `M.T.arguments() != M.T.arguments[::-1]` because M.T generates a new Tensor with new arguments and b) e.g. that `M.T.T != M`.

wence-:
> Mhh, maybe this is not the right way to do it after all. Maybe I need to do it at a lower level. The leftover issues we have are that a) `M.T.arguments() != M.T.arguments[::-1]` because M.T generates a new Tensor with new arguments

I think the difference is that:

```
adjoint(a)
```

Turns the argument with count 0 into the one with count 1, and vice versa

whereas

```
Tensor(a).arguments()[::-1]
```
just reverses the argument list.

b) e.g. that `M.T.T != M`.

This one should be the case, unless I misunderstand?

sv2518:
Sorry I think I explained b) wrong actually. In the Slate optimiser we drop double transposes so that M.T.T is just M. If we introduce the change on Slate level we won't know when we have double transposes on terminal tensors anymore, so we can't do that optimisation. But in fact actually we don't need this optimisation on terminal tensors anymore if we go over the adjoint, I think, because we just do `Tensor(adjoint(adjoint(form))` so that no extra operations in the Slate kernel are needed.
I guess I could work around a) by specifying `reordered_arguments`.

firedrake/slate/slac/optimise.py
```diff
@@ -235,6 +236,8 @@ def _drop_double_transpose_transpose(expr, self):
     if isinstance(child, Transpose):
         grandchild, = child.children
         return self(grandchild)
+    elif child.terminal and child.rank > 1:
+        return Tensor(adjoint(child.form))
```
sv2518, line 240: Okay. I changed this so that it is dealt with in Slate's optimiser. How is that? Now all tests are passing :)

wence-: OK, I think this works.

wence- (dismissed):
Do you want to squash everything nicely, or shall I just do it in the merge commit? Thanks!

sv2518:
Sorry I had to force-push twice because I screwed up the squashing the first way round. If you give me an approving review, I will merge tomorrow myself as soon as the tests are passing. :)
Thanks for feedback!

wence- (approved):
Thanks!

wence-:
> I will merge tomorrow myself as soon as the tests are passing.

I turned on auto-merge, so it might be done by the time you get up!

PR #2273, opened 2021-11-08
Revert "Install: pin setuptools to 58.4 for now"

sv2518:
This reverts commit 214f07a27265ac62d8c7217be2fea34448b107e3.  The issue in pypa/setuptools#2849 should be fixed, so we can remove pinning the setuptools version again.


wence- (approved):
Thanks!

PR #2275, opened 2021-11-10
tiger lake (or newer) processor and gcc/g++/gfortran 9.3 and older

### LINKED ISSUE #2214: Errors when running make alltest
I am trying to get Firedrake to function on my Windows 10 machine. When I run make alltest, I get an ImportError message and ModuleNotFoundError message, which can be viewed at this link:
https://user-images.githubusercontent.com/90576935/133002870-3a1768a0-569a-48e2-a687-3af749c507de.png

I am running this on Ubuntu 20.04.3

DWolfr:
This should catch the problems from #2214 on linux systems and stop the installer in the first seconds.In this moment I have no acces to a windows/mac so that I restricted the catch to linux systems.

### DIFF
--- a/scripts/firedrake-install
+++ b/scripts/firedrake-install
@@ -158,7 +158,7 @@ if platform.system() == "Linux":
         gfortranversionstring = os.popen('gfortran -dumpfullversion').read()
 
         # Check if the combination is a problem
-        if pversion.parse(gccversionstring) <= pversion.parse("9.3") or pversion.parse(gppversionstring) <= pversion.parse("9.3") or pversion.parse(gfortranversionstring) <= pversion.parse("9.3"):
+        if pversion.parse(gccversionstring) < pversion.parse("9.4") or pversion.parse(gppversionstring) < pversion.parse("9.4") or pversion.parse(gfortranversionstring) < pversion.parse("9.4"):
             print("""\nThe combination of a Tiger Lake (or newer) with gcc g++ and gfortran in the version 9.3 is not possible. Please install at least the gcc, g++ and gfortran in version 9.4.0.""")
             sys.exit(1)
 
--- a/scripts/firedrake-install
+++ b/scripts/firedrake-install
@@ -16,6 +16,7 @@ from glob import iglob
 from itertools import chain
 import re
 import importlib
+import packaging
 
 osname = platform.uname().system
 arch = platform.uname().machine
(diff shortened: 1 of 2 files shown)


DWolfr:
I have the same opinion, but i had no system for testing (our problematic systems have all a gcc10 now). But I thought a small catch would be helpful.

> Can you fix the `g++versionstring` errors, I think you meant `gppversionstring` ?

I am sorry for this dully mistake.

scripts/firedrake-install
```diff
@@ -142,6 +142,27 @@ Please follow the instructions at http://www.firedrakeproject.org/download.html
...
+        gccversionstring = gccversionstring[gccversionstring.find(')')+2:gccversionstring.find('\n')]
+
+        gppversionstring = os.popen('g++ --version').read()
+        gppversionstring = gppversionstring[g++versionstring.find(')')+2:gppversionstring.find('\n')]
+
+        gfortranversionstring = os.popen('gfortran --version').read()
+        gfortranversionstring = gfortranversionstring[gfortranversionstring.find(')')+2:gfortranversionstring.find('\n')]
+
+        # Check if the combination is a problem
+        if float(gccversionstring[0:gccversionstring.find('.') + 2]) <= 9.3 or float(gppversionstring[0:gppversionstring.find('.') + 2]) <= 9.3 or float(gfortranversionstring[0:gfortranversionstring.find('.') + 2]) <= 9.3:
```
dham, line 163: Please use `packaging.version.parse` per https://newbedev.com/how-do-i-compare-version-numbers-in-python

scripts/firedrake-install
```diff
@@ -142,6 +142,27 @@ Please follow the instructions at http://www.firedrakeproject.org/download.html
...
+            print("""\nThe combination of a Tiger Lake (or newer) with gcc g++ and gfortran in the version 9.3 is not possible.""")
```
dham, line 164: Can the error message please advise the user to install a gcc later than 9.3.

scripts/firedrake-install
```diff
@@ -142,6 +142,27 @@ Please follow the instructions at http://www.firedrakeproject.org/download.html
...
+# Temporary catch for tigerlake processors with gcc/g++/gfortran 9.3
+# But only for Linux
+if platform.system() == "Linux":
+    # Get the cpu information
+    cpuinfostring = os.popen('lscpu').read()
+    # Check if it is a tigerlake processor
+    if cpuinfostring.find('i3-11') != -1 or cpuinfostring.find('i5-11') != -1 or cpuinfostring.find('i7-11') != -1 or cpuinfostring.find('i9-11') != -1:
+        # Get the compiler versions
+        gccversionstring = os.popen('gcc --version').read()
+        gccversionstring = gccversionstring[gccversionstring.find(')')+2:gccversionstring.find('\n')]
```
dham, line 154: Can we avoid hand-rolled version string parsing?

dham: `packaging.version.parse(gccversionstring)`

scripts/firedrake-install
```diff
@@ -142,6 +142,27 @@ Please follow the instructions at http://www.firedrakeproject.org/download.html
...
 
+# Temporary catch for tigerlake processors with gcc/g++/gfortran 9.3
+# But only for Linux
+if platform.system() == "Linux":
+    # Get the cpu information
+    cpuinfostring = os.popen('lscpu').read()
+    # Check if it is a tigerlake processor
+    if cpuinfostring.find('i3-11') != -1 or cpuinfostring.find('i5-11') != -1 or cpuinfostring.find('i7-11') != -1 or cpuinfostring.find('i9-11') != -1:
+        # Get the compiler versions
+        gccversionstring = os.popen('gcc --version').read()
```
dham, line 153: `gcc -dumpversion`

dham (changes requested):
Thanks. There's just a bit of cleanup.

scripts/firedrake-install
DWolfr: That is a great idea.

scripts/firedrake-install
DWolfr: Added.

scripts/firedrake-install
DWolfr: That is a great package!

scripts/firedrake-install
DWolfr: That returns `"9\n"` in my case. But `packaging.version.parse` can handle `gcc --version`.

wence-: You might want `-dumpfullversion`

DWolfr: That is perfect!

scripts/firedrake-install
```diff
@@ -142,6 +143,24 @@ Please follow the instructions at http://www.firedrakeproject.org/download.html
...
+    if cpuinfostring.find('i3-11') != -1 or cpuinfostring.find('i5-11') != -1 or cpuinfostring.find('i7-11') != -1 or cpuinfostring.find('i9-11') != -1:
+        # Get the compiler versions
+        gccversionstring = os.popen('gcc -dumpfullversion').read()
+
+        gppversionstring = os.popen('g++ -dumpfullversion').read()
+
+        gfortranversionstring = os.popen('gfortran -dumpfullversion').read()
+
+        # Check if the combination is a problem
+        if pversion.parse(gccversionstring) <= pversion.parse("9.3") or pversion.parse(gppversionstring) <= pversion.parse("9.3") or pversion.parse(gfortranversionstring) <= pversion.parse("9.3"):
```
dham, line 161: I think there might be a subtle logical flaw here. If I understand correctly, versions up to and including 9.4.x are a problem. However:
```ipython
In [1]: from packaging.version import parse

In [2]: parse("9.3.1") <= parse("9.3")
Out[2]: False
```
Think you might need `<parse("9.4")` instead of `<= parse("9.3")`.

dham (changes requested):
Thanks again for this. It's way cleaner. See my comment for what might be a tiny issue.

scripts/firedrake-install
DWolfr: This is absolutely correct!

dham (approved):

dham:
Thanks @DWolfr !

PR #2278, opened 2021-11-12
Use safe CFLAGS for libsupermesh

connorjward:
By default libsupermesh uses `-march=native` during compilation. This can
cause install failures on some systems (e.g. Intel Tiger Lake + GCC
9.3). Here we now tell cmake to use the standard CFLAGS during compilation.

The offending line in libsupermesh's `CMakeLists.txt` is [here](https://bitbucket.org/libsupermesh/libsupermesh/src/69012e50dc6ba10721ed536871db8a84f6e68691/CMakeLists.txt#lines-100).

Related to #2275 and #2214.

### DIFF
--- a/scripts/firedrake-install
+++ b/scripts/firedrake-install
@@ -1085,7 +1085,8 @@ def build_and_install_libsupermesh():
                             "-DMPI_C_COMPILER=" + cc,
                             "-DMPI_CXX_COMPILER=" + cxx,
                             "-DMPI_Fortran_COMPILER=" + f90,
-                            "-DCMAKE_Fortran_COMPILER=" + f90])
+                            "-DCMAKE_Fortran_COMPILER=" + f90,
+                            "-DLIBSUPERMESH_AUTO_COMPILER_FLAGS=OFF"])
                 check_call(["make"])
                 check_call(["make", "install"])
     else:


sv2518:
Are we setting the CFLAGS for libsupermesh elsewhere?

connorjward:
> Are we setting the CFLAGS for libsupermesh elsewhere?

No. The CFLAGS will default to whatever CMAKE_C_FLAGS is (plus `-std=c99 -fPIC`) (see [here](https://bitbucket.org/libsupermesh/libsupermesh/src/69012e50dc6ba10721ed536871db8a84f6e68691/CMakeLists.txt#lines-130)). We could be more controlling about what flags actually get used but I think that this is sufficient.

sv2518:
Okay, do we maybe we want to set `O2`? If I understood correctly, the problem was only caused by -march=native.

connorjward:
> Okay, do we maybe we want to set `O2`? If I understood correctly, the problem was only caused by -march=native.

So on my machine the CFLAGS now are:
```
-- CMAKE_Fortran_FLAGS= -ffree-line-length-none -std=f2008
-- CMAKE_Fortran_FLAGS_DEBUG=-g
-- CMAKE_Fortran_FLAGS_RELEASE=-O3 -DNDEBUG -O3
-- CMAKE_C_FLAGS= -std=c99
-- CMAKE_C_FLAGS_DEBUG=-g
-- CMAKE_C_FLAGS_RELEASE=-O3 -DNDEBUG
-- CMAKE_CXX_FLAGS= -std=c++03
-- CMAKE_CXX_FLAGS_DEBUG=-g
-- CMAKE_CXX_FLAGS_RELEASE=-O3 -DNDEBUG
```

Whereas before they were:
```
-- CMAKE_Fortran_FLAGS=-ffree-line-length-none -pipe -std=f2008 -fallow-argument-mismatch
-- CMAKE_Fortran_FLAGS_DEBUG=-O0 -g -Wall -fcheck=all -ffpe-trap=invalid,zero,overflow,underflow -finit-integer=-66666 -finit-real=nan -fimplicit-none
-- CMAKE_Fortran_FLAGS_RELEASE=-O3 -march=native -ftree-vectorize -DNDEBUG
-- CMAKE_C_FLAGS=-pipe -std=c99
-- CMAKE_C_FLAGS_DEBUG=-O0 -g -Wall
-- CMAKE_C_FLAGS_RELEASE=-O3 -march=native -ftree-vectorize -DNDEBUG
-- CMAKE_CXX_FLAGS=-pipe -std=c++03
-- CMAKE_CXX_FLAGS_DEBUG=-O0 -g -Wall
-- CMAKE_CXX_FLAGS_RELEASE=-O3 -march=native -ftree-vectorize -DNDEBUG
```

So we lose the `-march=native` component. We still specify `-O3` which I don't think is likely to cause issues vs `-O2`.

sv2518 (approved):
Ok, LGTM

wence-:
@connorjward This breaks installs on MacOS since we need `-fallow-argument-mismatch` in the cflags.

connorjward:
> @connorjward This breaks installs on MacOS since we need `-fallow-argument-mismatch` in the cflags.

Ah. My apologies. I'll submit a fix ASAP.

wence-:
I think I can fix, just testing...

wence-:
In f0c387c678

connorjward:
Looks good. Thanks.

PR #2280, opened 2021-11-16
Use a version parser that is actually available.

dham:
Turns out `packaging` is not available by default but `pkg_resources` is, and the latter also provides a version string parser.


wence- (approved):
OK.

PR #2282, opened 2021-11-18
Update firedrakeproject/petsc

ksagiyam:
To update our PETSc fork:

PETSc: https://github.com/firedrakeproject/petsc/pull/7

SLEPc: https://github.com/firedrakeproject/slepc/pull/1

Added `--download-cmake` in `PETSC_CONFIGURE_OPTIONS`.

### DIFF
--- a/scripts/firedrake-install
+++ b/scripts/firedrake-install
@@ -1021,8 +1021,18 @@ def build_and_install_petsc():
         petsc_dir, petsc_arch = get_petsc_dir()
         check_call(["rm", "-rf", os.path.join(petsc_dir, petsc_arch)])
         petsc_options = get_petsc_options()
-        check_call([python, "./configure", "PETSC_DIR={}".format(petsc_dir), "PETSC_ARCH={}".format(petsc_arch)] + petsc_options)
-        check_call(["make", "PETSC_DIR={}".format(petsc_dir), "PETSC_ARCH={}".format(petsc_arch), "all"])
+        # Set MACOSX_DEPLOYMENT_TARGET to the current macos version
+        # to workaround issues configuring PETSc on macos.
+        try:
+            MACOSX_DEPLOYMENT_TARGET = check_output(["sw_vers", "-productVersion"])
+            # Convert to MAJOR.MINOR format.
+            MACOSX_DEPLOYMENT_TARGET = ".".join(MACOSX_DEPLOYMENT_TARGET.split(".")[:2])
+            log.info("MACOSX_DEPLOYMENT_TARGET set to {}".format(MACOSX_DEPLOYMENT_TARGET))
+        except FileNotFoundError:
+            MACOSX_DEPLOYMENT_TARGET = ""
+        with environment(MACOSX_DEPLOYMENT_TARGET=MACOSX_DEPLOYMENT_TARGET):
+            check_call([python, "./configure", "PETSC_DIR={}".format(petsc_dir), "PETSC_ARCH={}".format(petsc_arch)] + petsc_options)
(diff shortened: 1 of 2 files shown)


wence-:
There are some issues (at least for me) configuring PETSc on macos due to (I think) a bug in the way the C library buffers output, resulting in configure randomly failing.

ksagiyam:
I see. Then we probably need to wait until that issue is resolved.

wence-:
I think it might be that the MacOS linker prints warnings in an unsafe way. I think that we can fix it by building PETSc with `MACOSX_DEPLOYMENT_TARGET` set to the current macos version (can be obtained from `sw_vers`).

ksagiyam:
Something like this? Could you check if this works?

scripts/firedrake-install
```diff
@@ -1014,8 +1014,16 @@ def build_and_install_petsc():
         petsc_dir, petsc_arch = get_petsc_dir()
         check_call(["rm", "-rf", os.path.join(petsc_dir, petsc_arch)])
         petsc_options = get_petsc_options()
-        check_call([python, "./configure", "PETSC_DIR={}".format(petsc_dir), "PETSC_ARCH={}".format(petsc_arch)] + petsc_options)
-        check_call(["make", "PETSC_DIR={}".format(petsc_dir), "PETSC_ARCH={}".format(petsc_arch), "all"])
+        # Set MACOSX_DEPLOYMENT_TARGET to the current macos version
+        # to workaround issues configuring PETSc on macos.
+        try:
+            MACOSX_DEPLOYMENT_TARGET = check_output(["sw_vers", "-productVersion"])
+            log.info("MACOSX_DEPLOYMENT_TARGET set to {}".format(MACOSX_DEPLOYMENT_TARGET))
```
wence-, line 1021: I think this needs to just use `MAJOR.MINOR` not `MAJOR.MINOR.POINT` (which comes back from `sw_vers`).

.github/workflows/build.yml
```diff
@@ -26,7 +26,7 @@ jobs:
            complex: --complex
     env:
       FIREDRAKE_CI_TESTS: 1
-      PETSC_CONFIGURE_OPTIONS: --with-make-np=12 --download-mpich-device=ch3:sock
+      PETSC_CONFIGURE_OPTIONS: --with-make-np=12 --download-mpich-device=ch3:sock --download-cmake
```
wence-, line 29: If this is needed on the build system, is it also needed on the current debian-based platforms?

scripts/firedrake-install
ksagiyam: Fixed.

.github/workflows/build.yml
ksagiyam: It is needed only when the system cmake is older than 3.18.1; added a comment.

wence-: what I mean is, if we need to do this on the ubuntu:20.04 images we’re using for the builders, do we need to do it by default in firedrake-install for most apt-based systems?

ksagiyam: Thanks. Fixed now.

dham (approved):

ksagiyam:
@wence- Confirmed that all tests passed (https://github.com/firedrakeproject/firedrake/actions/runs/1899185845). I've removed the package branches, so this is now ready for merge (along with https://github.com/firedrakeproject/petsc/pull/7 and https://github.com/firedrakeproject/slepc/pull/1).

wence-:
FWIW, for slepc, we always want to do fast-forward merges (since we're not carrying around a patch in slepc like we are in petsc)

ksagiyam:
Ok!

PR #2283, opened 2021-11-19
Support casting Functions in R to float.

dham:


docs/source/r-space.rst
```diff
@@ -63,6 +63,27 @@ The corresponding Python code is:
   exact.interpolate(y - 0.5)
   print sqrt(assemble((u - exact)*(u - exact)*dx))
 
+Setting and retrieving the value of a function in `R`
+-----------------------------------------------------
+
+Functions in the space `R` are equivalent to a single floating point value. The
```
wence-, line 69: Do you mean \`R\` (i.e. "interpreted text, left up to the processor") or do you mean \`\`R\`\` (\texttt{R})

firedrake/function.py
```diff
@@ -482,6 +482,14 @@ def __idiv__(self, expr):
 
     __itruediv__ = __idiv__
 
+    def __float__(self):
+
+        if all((self.ufl_element().family() == "Real",
+                self.function_space().shape == ())):
```
wence-, line 488: Strange way of writing `self.ufl_element().family() == "Real" and self.function_space().shape == ()` ?

docs/source/r-space.rst
dham: I mean `R`, which is defined to be `$R$` by setting math as the default role at the top of the file. This matches the rest of the file.

firedrake/function.py
dham: This avoids PEP8 idiocy. The alternative:
```
if self.ufl_element().family() == "Real" \
    and self.function_space().shape == ():
```
Fails PEP8 because the indentation on the second line of the if statement is the same as the indentation in the block.

wence-: Arguably 87 characters is not so long that you would need to break the line. If you go for 80 character wrapping, [`black`](https://pypi.org/project/black/) produces:
```python
    def __float__(self):
        if (
            self.ufl_element().family() == "Real"
            and self.function_space().shape == ()
        ):
            return ...
```

I guess that's not the standard style in Firedrake right now.

I think the continuation line version works OK like this:
```python
    def __float__(self):
        if self.ufl_element().family() == "Real" \
           and self.function_space().shape == ():
            return self
```

dham: The latter version fails flake8. I've done the black version.

PR #2284, opened 2021-11-19
Only  attempt to plot real values

dham:
Matplotlib noticed that we were passing complex values into plotting commands. Simply pass the real part in instead.

### DIFF
--- a/firedrake/plot.py
+++ b/firedrake/plot.py
@@ -314,7 +314,7 @@ def trisurf(function, *args, complex_component="real", **kwargs):
     num_sample_points = kwargs.pop("num_sample_points", 10)
     function_plotter = FunctionPlotter(mesh, num_sample_points)
     triangulation = function_plotter.triangulation
-    values = function_plotter(function)
+    values = toreal(function_plotter(function), complex_component)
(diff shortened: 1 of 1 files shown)


firedrake/plot.py
```diff
@@ -862,7 +862,7 @@ def __call__(self, function):
         fiat_element = Q.finat_element.fiat_equivalent
         elem = fiat_element.tabulate(0, self._reference_points)[keys[dimension]]
         cell_node_list = Q.cell_node_list
-        data = function.dat.data_ro_with_halos[cell_node_list]
+        data = function.dat.data_ro_with_halos[cell_node_list].real
```
wence-, line 865: This is the minimal thing, but should we adopt the same `complex_component` idea as used elsewhere in this file and have:

```python
def __call__(self, function, complex_component="real"):
    ...
    data = toreal(function.dat.data_ro_with_halos[cell_node_list], complex_component)
```
To allow people to plot the imaginary part.

dham: You're right. I've done essentially that, though the minimal change version.

wence-: Ah, I hadn't spotted that `FunctionPlotter` is effectively an internal thing.

wence- (approved):

PR #2285, opened 2021-11-19
Disable very slow test

dham:
One of the interpolation tests suffers particularly from the large local tensors problem. This isn't going to get fixed rapidly so we'll just switch off the test so as not to inconvenience other developers.


tests/regression/test_interpolation_nodes.py
```diff
@@ -43,6 +43,8 @@ def test_div_curl_preserving(V):
     dim = mesh.geometric_dimension()
     if dim == 3 and V.ufl_element().degree() == 3 and "Nedelec" not in V.ufl_element().family():
         pytest.skip("N2div interpolation kernel with exact quadrature creates tensors which risk stack overflow")
+    if dim == 3 and V.ufl_element().degree() == 2 and "Nedelec" not in V.ufl_element().family():
+        pytest.skip("N2div interpolation kernel with exact quadrature takes 40 minutes!")
```
wence-, line 47: Perhaps note that this is a bug/feature in GCC which seemingly tries to constant-propagate every entry in a constant numeric array.

wence- (dismissed):

PR #2290, opened 2021-12-01
remove (near) nonzero entries from gradient matrix

florian98765:
improves performance of ams preconditioner.
an alternative would be to make assemble optionally ignore (near) 0 entries, but this would require changes in pyop as well as in PETSc (since currently only exactly 0 values are ignored).

### DIFF
--- a/firedrake/preconditioners/hypre_ams.py
+++ b/firedrake/preconditioners/hypre_ams.py
@@ -39,7 +39,7 @@ class HypreAMS(PCBase):
         G2.setType(PETSc.Mat.Type.AIJ)
         G2.setSizes(G.sizes)
         G2.setOption(PETSc.Mat.Option.IGNORE_ZERO_ENTRIES, True)
-        G2.setPreallocationCSR((ai,aj,a))
+        G2.setPreallocationCSR((ai, aj, a))
         G2.assemble()
 
         pc = PETSc.PC().create(comm=obj.comm)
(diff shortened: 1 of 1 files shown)


firedrake/preconditioners/hypre_ams.py
```diff
@@ -31,14 +32,26 @@ def initialize(self, obj):
...
 
+        # remove (near) zeros from sparsity pattern
+        ai, aj, a = G.getValuesCSR()
+        a[np.abs(a) < 1e-10] = 0
+        G2 = PETSc.Mat().create()
+        G2.setType(PETSc.Mat.Type.AIJ)
+        G2.setSizes(G.sizes)
+        G2.setOption(PETSc.Mat.Option.IGNORE_ZERO_ENTRIES, True)
+        G2.setUp()
+        G2.setValuesCSR(ai,aj,a)
```
wence-, line 43: Sorry it's taken so long to get to. I think this all looks good apart from the suggestion below:

```suggestion
        G2.setPreallocationCSR((ai,aj,a))
```

This does the correct preallocation and sets values all in one go.

We want to get the preallocation right because otherwise the assembly might be very slow.

firedrake/preconditioners/hypre_ams.py
```diff
@@ -31,14 +32,25 @@ def initialize(self, obj):
...
         G = Interpolator(grad(TestFunction(P1)), V).callable().handle
 
+        # remove (near) zeros from sparsity pattern
+        ai, aj, a = G.getValuesCSR()
+        a[np.abs(a) < 1e-10] = 0
+        G2 = PETSc.Mat().create()
+        G2.setType(PETSc.Mat.Type.AIJ)
+        G2.setSizes(G.sizes)
+        G2.setOption(PETSc.Mat.Option.IGNORE_ZERO_ENTRIES, True)
+        G2.setPreallocationCSR((ai,aj,a))
```
wence-, line 42:
```suggestion
        G2.setPreallocationCSR((ai, aj, a))
```

wence-:
Thanks!

florian98765:
Thanks for merging, Lawrence!

PR #2294, opened 2021-12-06
Dualspace update

nbouziani:
This PR contains the full dual space implementation. It builds on top of #2034, which introduced dual function spaces, the `Cofunction` and `AssembledMatrix` classes, as well as the DAG-based assembly of `BaseForm` objects.

Main components of this PR:

- Fix all the tests
- Extend the dual space implementation to work with pyadjoint
- Add `Coargument`
- Add assembly preprocessing stage (mostly to expand derivatives of the `BaseForm` being assembled)
- Add support to solve PDEs with a `Cofunction` in the rhs (fully annotated operation -> pyadjoint-friendly)
- Extend assembly of `Action` and `FormSum`
- Equip `Cofunction`s with `assign` and augmented assignments
- Expunge the use of `Vector` as the default type for adjoint/tlm/hessian values
- Add `riesz_representation` to map from `Cofunction`s to `Function`s and vice-versa
- Address implications of recent UFL changes (e.g. `Cofunction`s and `Coargument`s now respectively have 1 and 2 arguments)

The associated pyadjoint PR, which simply updates tests, can be found at https://github.com/dolfin-adjoint/pyadjoint/pull/70


firedrake/linear_solver.py
```diff
@@ -127,8 +127,14 @@ def _lifted(self, b):
         update()
         # blift contains -A u_bc
         blift += b
-        for bc in self.A.bcs:
-            bc.apply(blift)
+        if isinstance(blift, cofunction.Cofunction):
+            blift_func = function.Function(blift.function_space().dual(), val=blift.vector())
+            for bc in self.A.bcs:
+                bc.apply(blift_func)
```
dham, line 133: `bc.apply(blift.as_dual())`

firedrake/slate/slate.py
```diff
@@ -210,7 +211,9 @@ def coeff_map(self):
         The split coefficients are defined on the pieces of the originally mixed function spaces.
         """
         return tuple((n, tuple(range(len(c.split()))))
-                     if isinstance(c, Function) or isinstance(c, Constant)
+                     if isinstance(c, Function)
+                     or isinstance(c, Constant)
+                     or isinstance(c, BaseCoefficient)
```
dham, line 216: Cofunction

firedrake/vector.py
```diff
@@ -50,8 +50,15 @@ def __init__(self, x):
...
-        elif isinstance(x, firedrake.Function):
+            x_func = x.function
+            function_space = x_func.function_space()
+            # Need to allocate new space and copy values:
+            # -> Otherwise modifying the value of a Vector will also modify its copies.
+            x_copy = x_func.copy(deepcopy=True).vector()
+            # Can't do type(x.function)(x.function) because Cofunction
+            # doesn't accept this notation. Is there a valid reason for that ?
+            self.function = type(x_func)(function_space, val=x_copy)
+        elif isinstance(x, (firedrake.Function, firedrake.Cofunction)):
```
dham, line 61: Lift this to Cofunction.

dham (changes requested):

firedrake/adjoint/blocks.py
```diff
@@ -525,8 +525,13 @@ def evaluate_adj_component(self, inputs, adj_inputs, block_variable, idx, prepar
...
             raise NotImplementedError("Interpolate block must have a single output")
-        dJdm = self.backend.derivative(prepared, inputs[idx])
-        return self.backend.Interpolator(dJdm, self.V).interpolate(adj_inputs[0], transpose=True).vector()
+        input = inputs[idx]
+        dJdm = self.backend.derivative(prepared, input)
+        # Get the function space from `dJdm` argument
+        arg, = extract_arguments(dJdm)
+        # Make sure to have a cofunction output
+        output = self.backend.Cofunction(arg.function_space().dual())
+        return self.backend.Interpolator(dJdm, self.V).interpolate(adj_inputs[0], output=output, transpose=True).vector()
```
ReubenHill, line 534: Should this still return a `vector`? Is that what's required for compatibility with the innards of pyadjoint?

nbouziani: Yes it is.

.github/workflows/build.yml
```diff
@@ -79,6 +79,8 @@ jobs:
             --install icepack \
             --install irksome \
             --install femlium \
+            --package-branch pyadjoint dualspace \
+            --package-branch ufl fix_dualspace \
```
nbouziani, line 83:
```suggestion
```

firedrake/cofunction.py
```diff
@@ -0,0 +1,309 @@
...
+            self.dat = function_space.make_dat(val, dtype, self.name())
+
+        if isinstance(function_space, Cofunction):
+            self.dat.copy(function_space.dat)
+
+    def __del__(self):
+        if hasattr(self, "_comm"):
+            mpi.decref(self._comm)
+
+    def copy(self, deepcopy=False):
```
dham, line 82: At least default deepcopy to true

firedrake/cofunction.py
```diff
@@ -0,0 +1,309 @@
...
+
+        If present, subset must be an :class:`pyop2.types.set.Subset` of this
+        :class:`Cofunction`'s ``node_set``.  The expression will then
+        only be assigned to the nodes on that subset.
+        """
+        expr = ufl.as_ufl(expr)
+        if isinstance(expr, ufl.classes.Zero):
+            self.dat.zero(subset=subset)
+            return self
+        elif isinstance(expr, ufl.classes.ConstantValue):
```
dham, line 168: remove

firedrake/cofunction.py
```diff
@@ -0,0 +1,309 @@
...
+            from firedrake.function import Function
+            # Workaround to avoid using `assemble_expressions` directly
+            # since cofunctions are not `ufl.Expr`.
+            f = Function(self.function_space().dual()).assign(expr)
+            f.dat.copy(self.dat, subset=subset)
+            return self
```
nbouziani, line 174:
```suggestion
```

dham (changes requested):
fix cofunction.copy
don't introduce vectors. Introduce dats instead.

nbouziani:
This is ready to go once UFL upstream is merged.

.github/workflows/build.yml
```diff
@@ -79,6 +79,7 @@ jobs:
             --install icepack \
             --install irksome \
             --install femlium \
+            --package-branch ufl update_action_differentiation \
```
nbouziani, line 82:
```suggestion
```

firedrake/adjoint_utils/blocks/function.py
```diff
@@ -139,7 +137,7 @@ def evaluate_tlm_component(self, inputs, tlm_inputs, block_variable, idx,
                 dudmi.assign(ufl.algorithms.expand_derivatives(
                     ufl.derivative(expr, dep.saved_output,
                                    dep.tlm_value)))
-                dudm.vector().axpy(1.0, dudmi.vector())
+                dudm.dat += 1.0 * dudmi.dat
```
ReubenHill, line 140: I've never seen direct operations on a `PyOP2.dat` like this before. @connorjward is it the right thing to be doing?

firedrake/adjoint_utils/blocks/function.py
```diff
@@ -109,14 +107,14 @@ def _adj_assign_constant(self, adj_output, constant_fs):
         shape = r.ufl_shape
         if shape == () or shape[0] == 1:
             # Scalar Constant
-            r.vector()[:] = adj_output.vector().sum()
+            r.dat.data[:] = adj_output.dat.data_ro.sum()
```
ReubenHill, line 110: Nitpicking, but this I think _should_ be `r.dat.data_ro[:]`

ReubenHill:
```suggestion
            r.dat.data_ro[:] = adj_output.dat.data_ro.sum()
```

firedrake/adjoint_utils/blocks/function.py
nbouziani: This is exactly what `Vector.axpy` used to do. We are just lifting the code to not rely on `Vector`.

firedrake/adjoint_utils/blocks/function.py
nbouziani: We are modifying the value of `r` so why should we use `dat.data_ro` ? `.dat.data` needs to be used whenever you change the value of the `.dat`, right ?

ReubenHill: Apologies for the confusion, I meant to say `dat.data_wo` and hastily edited the comment after making it but clearly it didn't make it to you in time! I think `dat.data_wo` is correct here since you don't read the value

nbouziani: Well, `Dat.data_wo` is the same as `Dat.data` (see https://github.com/OP2/PyOP2/blame/master/pyop2/types/dat.py#L223). Also, we abusively use `Dat.data` when changing the value of the `Dat` all over the code so I don't think that is a problem.

Issue #2289, opened 2021-11-30
Elastodynamics example


zthowin:
Hello, I'd like to add a quasi-one-dimensional elastodynamics problem to the examples. The problem at hand has an analytical solution developed by Eringen & Suhubi (in [https://doi.org/10.1002/zamm.19800601028](https://doi.org/10.1002/zamm.19800601028)) which I will use for measurements of error vs. mesh refinement and order of interpolation.

I have developed implicit integration schemes using Newmark-beta parameters as well as explicit integration schemes using central-differencing in time for unlumped (consistent) and lumped mass matrices. The latter is giving me some trouble despite referencing a solution in #539; the numerical solution is quite different from the analytical solution. I am attaching a MWE prior to the pull request for the longer code if anyone has time to take a look and give feedback as to where I'm erring. Thanks!

[elastodynamics-mwe.zip](https://github.com/firedrakeproject/firedrake/files/7628490/elastodynamics-mwe.zip)

wence-:
Thanks! Perhaps the best way to debug this is to actually open a PR with the demo, and we can discuss there. I guess that since you're using piecewise linear elements, the lumping scheme should work OK. Can you explain in a bit more detail how the solution goes wrong?

PR #2296, opened 2021-12-07
Elastodynamics example

zthowin:
Referencing #2289.

Lawrence, the numerical solution when using a lumped mass matrix for explicit, central-differencing in time integration simply does not match the analytical solution, whereas a consistent mass matrix does match. When the solution is plotted, it looks as the applied traction is somehow reduced. Please see attached, thanks.
consistent

lumping

### DIFF
--- a/demos/elastodynamics/elastodynamics-mwe.py
+++ b/demos/elastodynamics/elastodynamics-mwe.py
@@ -1,0 +1,239 @@
+# Elastodynamics verification
+# ===========================
+
+# Here we solve an elastodynamics equation using an explicit
+# timestepping scheme. This example demonstrates the use of pointwise 
+# operations on Functions and a time varying Neumann boundary condition. 
+# The strong form of the equation we set out to solve is:
+
+# .. math::
+
+#    \rho\mathbf{a} - \nabla\cdot\mathbf{\sigma} = 0
+
+#    \mathbf{\sigma} \cdot \mathbf{n} = \mathbf{t}^\sigma = -f(t) \ \textrm{on}\ \Gamma_N
+
+#    \mathbf{u}\cdot\mathbf{n} = 0 \ \textrm{on}\ \Gamma_D
+
+#    f(t) = \frac{f_0}{2}\left(1-\cos(\omega t)\right)
+
+# where we have assumed linear elasticity such that:
+
+# .. math::
+
+#   \mathbf{\sigma} = \lambda\mathrm{tr}(\mathbf{u}) + 2\mu\epsilon(\mathbf{u})
+
+#   \epsilon(\mathbf{u}) = \frac{1}{2}\left(\nabla\mathbf{u} + \left(\nabla\mathbf{u}\right)^T\right)
+
+# We employ an explicit, central-differencing scheme. The acceleration update is written in terms of the displacements as:
+
+# .. math::
+
+#     \mathbf{a}_{n+1} = \frac{\mathbf{u}_{n+1} - 2\mathbf{u}_n + \mathbf{u}_{n-1}}{\Delta t^2}
+
+# We then write the weak form of the equation for :math:`\mathbf{u}`. Find
+# :math:`\mathbf{u} \in \mathbf{W}` such that:
+
+# .. math::
+
+#    \int_\Omega \frac{\rho}{\Delta t^2}\left(\mathbf{u}_{n+1} - 2\mathbf{u}_n + \mathbf{u}_{n-1}\right)\cdot\mathbf{w} d\Omega + \int_\Omega\mathbf{\sigma}_{n+1}:\epsilon(\mathbf{w}) d\Omega = \int_{\partial\Omega}\mathbf{t}\cdot\mathbf{w} dS
+#    \quad \forall \mathbf{w} \in \mathbf{W}
+
+# for a suitable function space :math:`\mathbf{W}`.
+
+# This problem has an analytical solution that was developed in Eringen & Suhubi (1975):
+
+# .. math::
+#   u(z,t) = \frac{4}{\pi\rho c}\sum_{n=1}^\infty\frac{(-1)^n}{2n - 1}\left[\int_0^t f(\tau)\sin\left(\frac{(2n-1)\pi c(t-\tau)}{2H}\right)\dtau\right]\sin\left(\frac{(2n-1)\pi z}{2H}\right)
+
+from firedrake import *
+import math, sys
+import numpy as np
+import matplotlib.pyplot as plt
+
+EPSILON = 1.0e-14
+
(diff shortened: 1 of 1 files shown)


demos/elastodynamics/elastodynamics-mwe.py
```diff
@@ -0,0 +1,239 @@
...
+        rhs = assemble(-inner(sigma(u_np1), epsilon(w))*dx + dot(traction, w)*Gamma_N)
+        us  = Function(W).assign(rhs)
+        # Divide by lumped mass matrix::
+        us /= lumped_mass
+        # Update displacements::
+        u_np1.assign(us + 2*u_n - u_nm1)
+        for bc in bcSet:
+            bc.apply(u_np1)
+        u_n.assign(u_np1)
+        u_nm1.assign(u_n)
```
ksagiyam, line 155:
```
u_nm1.assign(u_n)
u_n.assign(u_np1)
```
?

demos/elastodynamics/elastodynamics-mwe.py
```diff
@@ -0,0 +1,239 @@
...
+        # Save data::
+        u_numerical[step] = u_n([0., 0., 20.])[2]
+
+else:
+    # Apply the central differencing scheme::
+    F    = inner(sigma(u), epsilon(w))*dx + rho/(dt*dt)*dot(u - 2*u_n + u_nm1, w)*dx - dot(traction, w)*Gamma_N
```
ksagiyam, line 161:
```
F = inner(sigma(u_n), epilon(w)) * dx + ...
```
I think. I *think* this way we can make sure that the initial velocity (which I assume is `0`) is approximated to second-order.

wence-, line 161: I think, as written here, this is a semi-implicit scheme? `u` is a trial function and `w` is a test function so after splitting this `F` we're solving:

```
(sigma(u), eps(w)) + rho/dt^2 (u, w) = (traction, w)_F + rho/dt^2 (u_{n-1} - 2 u_n, w)
```

So we're not just inverting a mass matrix at each step?

demos/elastodynamics/elastodynamics-mwe.py
```diff
@@ -0,0 +1,239 @@
...
+        step += 1
+        # Update load::
+        traction.assign((0., 0., -0.5*amplitude*(1 - cos(omega*t))))
+        # Assemble right-hand side::
+        rhs = assemble(-inner(sigma(u_np1), epsilon(w))*dx + dot(traction, w)*Gamma_N)
+        us  = Function(W).assign(rhs)
+        # Divide by lumped mass matrix::
+        us /= lumped_mass
+        # Update displacements::
+        u_np1.assign(us + 2*u_n - u_nm1)
```
wence-, line 151: OK, so in this case we've done:
```
rho/dt^2 (u, w) = - (sigma(u_{n-1}), eps(w)) + (traction, w)_F + rho/dt^2 (u_{n-1} - 2 u_n, w)
```
Which after rearranging is
```
(u, w) = dt^2/rho * [(traction, w)_F - (sigma(u_{n-1}), eps(w))] + (u_{n-1} - 2 u_n, w)
```
To mass lump, we need to divide this through by the lumped mass, call it `M_lump`, but I think you're only dividing through the first of the two terms by the lumped mass and obtaining:
```
u_{n+1} = dt^2/rho M_lump^{-1} * [(traction, w)_F - (sigma(u_{n-1}), eps(w))] + (u_{n-1} - 2 u_n, w)
```

This is correct as long as `M_lump` is `rho/dt^2 I` where `I` is the identity. But I think you're likely missing a scaling by `detJ` on each cell or similar.

I think we instead want to write the left and right hand sides:


```
t = Constant(t)
traction = as_vector([0, 0, -0.5*amplitude*(1 - cos(omega * t))])
a = inner(u, w)*dx
L = dt**2/rho * (inner(traction, w)*Gamma_N - inner(sigma(u_nm1), epsilon(w))*dx) + inner(u_nm1 - 2*un, w)*dx
# Now produce the lumped mass matrix:
alump = replace(a, {u: as_vector([1, 1, 1])})
lumped_mass = assemble(alump)
while t < T + EPSILON:
    # u_{n+1} <- M_lump^{-1} L (first by assigning the assembly of L and then dividing by the lumped mass)
    t.assign(t + dt)
    assemble(L, tensor=u_np1)
    u_np1 /= lumped_mass
```

Note that I defined the `traction` vector which is time-dependent symbolically, so when we update `t`, the assembly of the `rhs` notices. This is also the right thing to do because each assembly with your previous approach would require JIT-compilation of the form defining the RHS (because you have a literal number, `t` in the form that changes at each step).

Can you have a go at these changes to see if it fixes the mass-lumped problem?

I think if so, it will also allow us to simplify the demo a bit for the non-mass-lumped case: notice how we define the same forms we'll need for the full mass solve, only in that case we won't do this `replace` dance.

wence-:
Thanks, this looks really good. Let's fix the numerics issues and then we can go round and clean up some of the style a bit to conform with best practices in the firedrake demos (which I note are rather under-specified, sorry!).

demos/elastodynamics/elastodynamics-mwe.py
ksagiyam: That might be true. I think using explicit central difference here allows for a more fair comparison to the lumped mass version, though.

demos/elastodynamics/elastodynamics-mwe.py
```diff
@@ -0,0 +1,233 @@
...
+
+    # Time-stepping loop::
+    for step in range(1, Nsteps_numerical + 2):
+        # Update time step::
+        t.assign(t + dt)
+
+        assemble(L, tensor=u_np1, bcs=bcSet)
+
+        # Divide by lumped mass matrix::
+        u_np1 /= lumped_mass
```
zthowin, line 152: Thank you both for your feedback. @ksagiyam [spotted](https://github.com/firedrakeproject/firedrake/pull/2296#discussion_r765695973) that I had updated `u_nm1` after updating `u_n` in the lumped mass case which is what was throwing the error. I have anyways implemented @wence- 's suggestion that we write the form once. However, this block still gives an inaccurate solution, even after correcting the stress term to leave the method explicit. I tested with the consistent mass matrix which looks fine, so I believe I wrote the form correctly.

wence-: I think you want to apply the bcs _after_ dividing by the lumped mass (I think it probably isn't the cause of the error in this case because the bcs are all zero) since otherwise you don't have the bc value on the boundary nodes but rather the bc value divided through by the lumped mass matrix. I'll try and find some time to run this demo to see what is going on though...

PR #2297, opened 2021-12-09
Add Interpolate and support for BaseFormOperators

nbouziani:
This PR builds on the dual space changes recently introduced (#2034, #2294) and provides the firedrake bit of the symbolic interpolation operator (`ufl.Interpolate`). This PR prepares the ground for treating the assembly of operators that can be seen as operators and as forms such as `ufl.Interpolate` or `ufl.ExternalOperator`. Most of the bits in this PR come from the external operator dual branch which will come in another PR.

### Associated PRs:

**Pyadjoint:**: https://github.com/dolfin-adjoint/pyadjoint/pull/127
**ngsPETSc**: https://github.com/NGSolve/ngsPETSc/pull/14

Main components of this PR:

### Add pre-assembly phase

The assembly is now composed of two phases:

- Differentiation expansion
- DAG restructuration stage

### Remove `InterpolateBlock` and redefine `Interpolator.interpolate`

Calling `Interpolator.interpolate` will now assemble an `Interp` object, which enables to deal with annotation since performing interpolation will now drop an `AssembleBlock` on the tape. Having this structure also ensures consistency between `Interpolator` and `Interp`.


tests/regression/test_interp_dual.py
```diff
@@ -0,0 +1,168 @@
...
+    If1 = Interp(f1, V2)
+    assert isinstance(If1, ufl.Interp)
+
+    # -- I(f1, V2) -- #
+    a = assemble(If1)
+    b = interpolate(f1, V2)
+    assert np.allclose(a.dat.data, b.dat.data)
+
+    assembled_If1 = assemble(If1)
+    assert np.allclose(assembled_If1.dat.data, b.dat.data)
```
lrtfm, line 37: It seems that lines 36-37 are the same as lines 32 and 34.

nbouziani: Thank you very much! I am not sure what happened here.

.github/workflows/build.yml
```diff
@@ -79,6 +79,8 @@ jobs:
             --install icepack \
             --install irksome \
             --install femlium \
+            --package-branch pyadjoint dualspace \
+            --package-branch ufl interp_dual \
```
nbouziani, line 83:
```suggestion
```

.github/workflows/docs.yml
```diff
@@ -42,6 +42,10 @@ jobs:
       - name: Check documentation links
         if: ${{ github.ref == 'refs/heads/master' }}
         run: |
+          git config --global --add safe.directory /home/firedrake/firedrake/src/ufl
+          cd /home/firedrake/firedrake/src/ufl
+          git checkout interp_dual
+          cd /home/firedrake/firedrake/src/firedrake
```
nbouziani, line 48:
```suggestion
```

firedrake/interpolation.py
```diff
@@ -215,8 +250,58 @@ def __init__(
...
         self.callable = None
 
+    @PETSc.Log.EventDecorator()
+    def interpolate(self, *function, output=None, transpose=False, default_missing_val=None):
+        """This performs interpolation by assembling `Interp`, which in turn calls
+        `Interpolator._interpolate`. Having this structure ensures consistency between
+        `Interpolator` and `Interp`.
+
+        This mechanism handles annotation since performing interpolation will drop an
+        Assemble block on the tape.
```
ReubenHill, line 260: This is implementation detail, rather than a docstring. This docstring should be identical to the existing docstring for `Interpolator.interpolate`, since the public interface has not changed. Details like this should be in a comment.

firedrake/interpolation.py
```diff
@@ -215,8 +250,58 @@ def __init__(
...
+
+        :arg function: If the expression being interpolated contains an
+            :class:`ufl.Argument`, then the :class:`.Function` value to
+            interpolate.
+        :kwarg output: Optional. A :class:`.Function` to contain the output.
+        :kwarg transpose: Set to true to apply the transpose (adjoint) of the
+              interpolation operator.
+        :returns: The resulting interpolated :class:`.Function`.
```
ReubenHill, line 268: Note that you are missing documentation on `default_missing_val`. As in the other comment, you can rectify this by just reusing the existing docstring

firedrake/interpolation.py
```diff
@@ -215,8 +250,58 @@ def __init__(
...
+        # -> `interp = Action(Interpolate(v1, v0), f)` with `v1` and `v0` being respectively `Argument`
+        # and `Coargument`. Differentiating `interp` is not currently supported as the action's left slot
+        # is a 2-form. However, after preprocessing, we obtain `Interpolate(f, v0)`, which can be differentiated.
+        interp = preprocess_base_form(interp)
+        res = assemble(interp, tensor=output, is_base_form_preprocessed=True)
+        return res
+
     @abc.abstractmethod
-    def interpolate(self, *args, **kwargs):
+    def _interpolate(self, *args, **kwargs):
```
ReubenHill, line 304: Could this have a more descriptive name? Looking at how it's used, perhaps `assemble` would be more appropriate?

ReubenHill (dismissed):
I know this is WIP, but hopefully I've highlighted some helpful things. Also, unless David has changed his mind, isn't `Interpolate` going to start off in some `.future` module?

firedrake/interpolation.py
```diff
@@ -215,8 +250,68 @@ def __init__(
...
+            output = self.V
+
+        get_working_tape()
+        # Assembly annotation records the form before it is preprocessed. Hence, we preprocess
+        # the form beforehand to facilitate derivation of the tangent linear/adjoint models.
+        # For example,
+        # -> `interp = Action(Interpolate(v1, v0), f)` with `v1` and `v0` being respectively `Argument`
+        # and `Coargument`. Differentiating `interp` is not currently supported as the action's left slot
+        # is a 2-form. However, after preprocessing, we obtain `Interpolate(f, v0)`, which can be differentiated.
+        interp = preprocess_base_form(interp)
```
dham, line 309: Move this inside the assemble annotatation

firedrake/interpolation.py
```diff
@@ -215,8 +250,68 @@ def __init__(
...
+
+        # Get the Interpolate object
+        interp = self._interpolate_future(*function, transpose=transpose,
+                                          default_missing_val=default_missing_val)
+
+        if isinstance(self.V, firedrake.Function) and not output:
+            # V can be the Function to interpolate into (e.g. see `Function.interpolate``).
+            output = self.V
+
+        get_working_tape()
```
dham, line 302: SHouldn't be needed.

firedrake/interpolation.py
```diff
@@ -215,8 +250,68 @@ def __init__(
...
+        if function:
+            f, = function
+            # Passing in a function is equivalent to taking the action.
+            interp = action(interp, f)
+        # Return the `ufl.Interpolate` object
+        return interp
+
+    @PETSc.Log.EventDecorator()
+    def interpolate(self, *function, output=None, transpose=False, default_missing_val=None):
+        """This performs interpolation by assembling `Interp`, which in turn calls
```
connorjward, line 276: Docstrings should be imperative (e.g. `"Interpolate an expression."`) and short enough to fit on a line. Leave a blank line between this and the rest of the information.

connorjward: And should this be `Interpolate` instead of `Interp`?

tests/regression/test_interp_dual.py
```diff
@@ -0,0 +1,212 @@
...
+
+
+@pytest.fixture
+def M(fs):
+    uhat = TrialFunction(fs)
+    v = TestFunction(fs)
+    return inner(uhat, v) * dx
+
+
+def test_assemble_interp(mesh):
```
connorjward, line 19: Ideally break this function up into lots of smaller ones.

firedrake/assemble.py
```diff
@@ -169,33 +399,45 @@ def assemble_base_form(expression, tensor=None, bcs=None,
...
+
+
+def preprocess_base_form(expr, mat_type=None, form_compiler_parameters=None):
+    """Preprocess ufl.BaseForm objects"""
+    if mat_type != "matfree":
+        # For "matfree", Form evaluation is delayed
+        expr = expand_derivatives_form(expr, form_compiler_parameters)
+    # Expanding derivatives may turn `ufl.BaseForm` objects into `ufl.Expr` objects that are not `ufl.BaseForm`.
+    if not isinstance(expr, ufl.form.BaseForm):
+        return assemble(expr)
```
connorjward, line 435: It strikes me as a bit odd to call `assemble` inside of a preprocessing function. It seems like it should recurse infinitely.

connorjward (changes requested):

dham (changes requested):

tests/regression/test_interpolate.py
```diff
@@ -434,7 +434,7 @@ def test_interpolate_minmax(access):
     g = interpolate(x*y, V)
     f = interpolate(x**2 - y**4, V)
     actual = Function(g)
-    actual = interpolate(f, actual, access=access).dat.data_ro
+    actual = Interpolator(f, actual, access=access)._interpolate().dat.data_ro
```
ReubenHill, line 437: This is another place where we need the old API to test min/max. If this is abuse of notation as @dham suggested, perhaps different accessors should just be removed.

ReubenHill: Either that or let `access` be a kwarg to `Interpolate`

tests/vertexonly/test_vertex_only_fs.py
```diff
@@ -148,18 +148,22 @@ def functionspace_tests(vm):
...
     assert np.allclose(h2.dat.data_ro_with_halos[idxs_to_include], 2*np.prod(vm.input_ordering.coordinates.dat.data_ro_with_halos[idxs_to_include].reshape(-1, vm.input_ordering.geometric_dimension()), axis=1))
-    g = I_io.interpolate(h, transpose=True)
+
+    h_star = h.riesz_representation(riesz_map="l2")
+    g = I_io.interpolate(h_star, transpose=True)
     assert np.allclose(g.dat.data_ro_with_halos, np.prod(vm.coordinates.dat.data_ro_with_halos.reshape(-1, vm.geometric_dimension()), axis=1))
     with pytest.raises(NotImplementedError):
         # Can't use transpose on interpolators with expressions yet
-        g2 = I2_io.interpolate(h, transpose=True)
+        g2 = I2_io.interpolate(h_star, transpose=True)
```
ReubenHill, line 157: Here, and in loads of other places in the test suite, we are using `Interpolator.interpolate` in the way that we're about to warn everyone not to do - i.e. we're relying on getting a `Function` or `Cofunction` out rather than an expression. Surely we should switch to `assemble(Interpolate(...))` in these cases?

ReubenHill: In the specific case of testing out `Interpolator`s I guess you would need to do `assemble(Interpolator(...)._interpolate_future(...))`? Is that correct @nbouziani ?

ReubenHill: Wait that can't be right - in `__future__.py` you replace `Interpolator.interpolate` with `Interpolator._interpolate_future`. @nbouziani, either we haven't switched on `__future__.py` in all these tests or `Interpolate` somehow has a `.dat` associated with it. Which one is it?

firedrake/interpolation.py
```diff
@@ -33,12 +33,50 @@
...
+        ufl.Interpolate.__init__(self, expr, v)
+
+        # Interpolate data (e.g. subset or access)
+        self.interp_data = interp_data if interp_data else {}
+
+    def function_space(self):
+        return self._function_space
+
+    def _ufl_expr_reconstruct_(self, expr, v=None, interp_data=None):
+        return ufl.Interpolate._ufl_expr_reconstruct_(self, expr, v=v, interp_data=interp_data or self.interp_data)
```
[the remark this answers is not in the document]
nbouziani: We override this because `interp_data` is a Firedrake thing. As for the other arguments (`expr` and `v`), they are handled by the parent class, which will do that.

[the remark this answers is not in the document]
connorjward: I see. But if `interp_data` is a Firedrake thing then why are we passing it to the UFL `Interpolate` constructor. Won't that error?

firedrake/assemble.py
nbouziani: No, because the object is not a `ufl.form.BaseForm` anymore. Calling back `assemble` to use the right assembly procedure at this stage is a consequence of expanding derivatives (cf. above comment).

tests/vertexonly/test_vertex_only_fs.py
nbouziani: We import `interpolate` from `__future__.py`, so that tests align with the future interpolation behaviour. Then, we assemble any occurrence of `interpolate` or `Interpolator.interpolate`.

firedrake/interpolation.py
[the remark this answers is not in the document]
nbouziani: It accepts extra optional arguments (see https://github.com/firedrakeproject/ufl/blob/master/ufl/core/interpolate.py#L61)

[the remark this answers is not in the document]
connorjward: I see. Thanks!

firedrake/assemble.py
connorjward: My point is that isn't obviously preprocessing. The name of the function is a bit misleading.

nbouziani: The point of this function is really to preprocess the form via expanding derivatives or restructuring the corresponding DAG. I think what I have done here might be a premature simplification. I should be able to lift that out of the preprocessing stage so that it's done within a function aimed at assembling and not preprocessing, which would make things clearer.

nbouziani: Done

tests/vertexonly/test_vertex_only_fs.py
ReubenHill: But there's no assembly here... What am I missing?

firedrake/interpolation.py
```diff
@@ -299,39 +333,15 @@ def interpolate(self, *function, output=None, transpose=False, default_missing_v
...
-        res = assemble(interp, tensor=output, is_base_form_preprocessed=True)
+        # Assemble the `ufl.Interpolate` object, which will then call `Interpolator._interpolate`
+        # to perform the interpolation. Having this structure ensures consistency between
+        # `Interpolator` and `Interp`. This mechanism handles annotation since performing interpolation will drop an
+        # `AssembleBlock` on the tape.
+        res = assemble(interp, tensor=output)
         return res
 
     @abc.abstractmethod
     def _interpolate(self, *args, **kwargs):
```
ReubenHill, line 344: Please add a docstring to `_interpolate` so it's clear what it does (i.e. it performs the interpolation)

tests/vertexonly/test_vertex_only_fs.py
ReubenHill: Oh this is outdated, nevermind

firedrake/adjoint_utils/blocks/function.py
```diff
@@ -204,338 +201,6 @@ def __str__(self):
...
-    arguments.
-
-    ::
-
-        f : W × G ⟶ V i.e. f(u, g) ∈ V ∀ u ∈ W and g ∈ G. f = I ∘ expr I :   X
-        ⟶ V i.e. I(;x) ∈ V ∀ x ∈ X.
-                          X is infinite dimensional.
-        expr: W × G ⟶ X i.e. expr(u, g) ∈ X ∀ u ∈ W and g ∈ G.
-
-    Arguments after a semicolon are linear (i.e. operation I is linear)
```
ReubenHill, line 244: @nbouziani  I presume interpolation is now annotated by something else. I presume whichever block covers assembly? If so, is annotation still correct for interpolating expressions which are not linear, as described here?

ReubenHill: I think as long as [this test](https://github.com/firedrakeproject/firedrake/blob/c568bfc4f932e72a434cebc74eacb982e02cfd86/tests/regression/test_adjoint_operators.py#L346) passes then yes it does

firedrake/interpolation.py
```diff
@@ -33,12 +33,65 @@
...
+    "Interpolate",
     "DofNotDefinedError",
     "CrossMeshInterpolator",
     "SameMeshInterpolator",
 )
 
 
+class Interpolate(ufl.Interpolate):
+
+    def __init__(self, expr, v, interp_data=None):
```
ReubenHill, line 45: Is there a good reason why `interp_data` is not `**kwargs` given it's an optional dictionary?

firedrake/interpolation.py
```diff
@@ -33,12 +33,65 @@
...
+        r""" Symbolic representation of the interpolation operator.
+
+        :arg expr: a UFL expression to interpolate.
+        :arg v: the :class:`.FunctionSpace` to interpolate into or the :class:`.Coargument`
+                defined on the dual of the :class:`.FunctionSpace` to interpolate into.
+        :arg interp_data: optional dictionary containing interpolation-specific data
```
ReubenHill, line 51: This is not sufficient documentation given this will become the primary API for interpolation. You need to lift the keyword argument list from `Interpolator` and put it here

ReubenHill:
Note that the manual needs to be updated to reflect these changes. (This should be part of this PR).

firedrake/interpolation.py
nbouziani: Yes, that's clearer and cleaner. It distinguishes between the mathematical objects involved in the interpolation and the interpolation data specific to the interpolation computation. That's similar to what we do with solver parameters in `solve`. For example, one would write:

```
I = Interpolate(expr, V, interp_data={"subset": ...,
                                      "default_missing_val": ...,
                                      "allow_missing_dofs": ...})
```

instead of 

```
I = Interpolate(expr, V, subset=..., default_missing_val=..., allow_missing_dofs=...)
```

ReubenHill: Makes sense. Maybe it should be called something like `compute_kwargs` then?

tests/regression/test_interp_dual.py
```diff
@@ -0,0 +1,245 @@
...
+
+    b = assemble(interpolate(f1, V2))
+
+    # Get the interpolation matrix
+    a = assemble(Iv1)
+    res = Cofunction(V2.dual())
+    # Check that `I * f1 == b` with I the interpolation matrix
+    # and b the interpolation of f1 into V2.
+    with f1.dat.vec_ro as x, res.dat.vec_ro as y:
+        a.petscmat.mult(x, y)
```
dham, line 60:
```
res = assemble(action(a, y))

**Review by dham — CHANGES_REQUESTED** [review]
Additionally lower case I everywhere possible.

**Comment by connorjward** [comment]
Just an idea, to aid with the transition would it be good to have something like:
```py
class Interp(...):  # the symbolic thing
    @property
    def dat(self):
        # throw an error, interpolate is now symbolic!
```
I think that if users were expecting `interpolate` to return a `Function` then they should only run into problems if they try to do this. Perhaps `.assign` might also need a shim method.

Not sure if this is the way to go. Food for thought.

nbouziani:
> Just an idea, to aid with the transition would it be good to have something like:
> 
> ```python
> class Interp(...):  # the symbolic thing
>     @property
>     def dat(self):
>         # throw an error, interpolate is now symbolic!
> ```
> 
> I think that if users were expecting `interpolate` to return a `Function` then they should only run into problems if they try to do this. Perhaps `.assign` might also need a shim method.
> 
> Not sure if this is the way to go. Food for thought.

Yes, I think that makes sense and that would help. I am just a bit worried about starting accumulating shim methods for all the things you can do with a function just to say that the object doesn't have a `Function` type. Maybe we should just keep that for `dat` as that's what contains the underlying tensor of a function.

connorjward:
> > Just an idea, to aid with the transition would it be good to have something like:
> > ```python
> > class Interp(...):  # the symbolic thing
> >     @property
> >     def dat(self):
> >         # throw an error, interpolate is now symbolic!
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > I think that if users were expecting `interpolate` to return a `Function` then they should only run into problems if they try to do this. Perhaps `.assign` might also need a shim method.
> > Not sure if this is the way to go. Food for thought.
> 
> Yes, I think that makes sense and that would help. I am just a bit worried about starting accumulating shim methods for all the things you can do with a function just to say that the object doesn't have a `Function` type. Maybe we should just keep that for `dat` as that's what contains the underlying tensor of a function.

I don't think that there's any harm in adding lots of methods. They are one line and easy to delete. Something like:
```py
@property
def dat(self):
    self._raise_not_a_function_error()

def assign(self, *args, **kwargs):
    self._raise_not_a_function_error()
```

dham (approved):

JDBetteridge (approved):
Previously approved by dham

ReubenHill:
Hooray!

Issue #2300, opened 2021-12-13
Incorrect assembly of surface integrals when diagonal=True


pbrubeck:
Consider a form that only depends on the jump of CG(1) test and trial functions. 
Calling `assemble(a, diagonal=False)` gives a matrix of zeros, which is the expected output. 
On the other hand, `assemble(a, diagonal=True)`  gives non-zeros. For DG(1), the diagonal is also incorrectly computed.

Here is the minimal failing example:
```
from firedrake import *

mesh = UnitIntervalMesh(2)
V = FunctionSpace(mesh, "CG", 1)

u = TrialFunction(V)
v = TestFunction(V)
n = FacetNormal(mesh)
outer_jump = lambda w, n: outer(w('+'), n('+')) + outer(w('-'), n('-'))
a = inner(outer_jump(v, n), outer_jump(u, n)) * dS
A = assemble(a)
print(A.petscmat.view())

diag = assemble(a, diagonal=True)
with diag.dat.vec as x:
    print(x.view())
```

Giving the output:
```
Mat Object: 1 MPI processes
  type: seqaij
row 0: (0, 0.)  (1, 0.)  (2, 0.) 
row 1: (0, 0.)  (1, 0.)  (2, 0.) 
row 2: (0, 0.)  (1, 0.)  (2, 0.) 
None
Vec Object: 1 MPI processes
  type: seq
0.
2.
0.
None
```

wence-:
TSFC produces a kernel to compute the diagonal with this:
```
        # Diagonal assembly is obtained by using the test indices for
        # the trial space as well.
        a, _ = argument_multiindices
        argument_multiindices = (a, a)
```

I think this isn't right for interior facets (I am not sure there is something that is correct).

If I just use `jump(u), jump(v)` for your example, the local tensor is 4x4
```
In [91]: A
Out[91]: 
array([[ 1.,  0.,  0., -1.],
       [ 0.,  0.,  0.,  0.],
       [ 0.,  0.,  0.,  0.],
       [-1.,  0.,  0.,  1.]])

In [92]: V.interior_facet_node_map().values
Out[92]: array([[1, 0, 2, 1]], dtype=int32)
```

So you can see that the +1 cancels with the -1 on insertion.

The 4x4 matrix is logically split into blocks:
```
+/+ | +/-
----+----
-/+ | -/-
```

[...]

PR #2301, opened 2021-12-14
Use FDMElement inside FDMPC and PMGPC

pbrubeck:
Here, I generalize `prolongation_matrix_matfree` in `pmg.py` to handle TensorProductElements of 1D elements whose dual might include  nodes that are not `PointEvaluation`. Then, I reuse the prolongator to transform between nodal and FDM spaces for `FDMPC`. Additionally, I am detecting whether a problem is already formulated with the FDM variant, so we may skip the transformation.

I also have written a much cleaner method (`expand_element` and `get_line_elements`) to decompose UFL elements into sum of products of elements on the interval. This simplifies the logic that was required before (`tensor_product_space_query`) to detect whether a given element matched the families Q, DQ, RTCE\F, NCE\F.

This PR depends on [UFL#27](https://github.com/firedrakeproject/ufl/pull/27), [FIAT#26](https://github.com/firedrakeproject/fiat/pull/26), [FInAT#100](https://github.com/FInAT/FInAT/pull/100), [TSFC#265](https://github.com/firedrakeproject/tsfc/pull/265).

As a work in progress, I have added code to assemble the diagonal entries of the exact matrix in the FDM basis to rescale the approximate one. Nevertheless, I have disabled this feature since we are computing the diagonal entries incorrectly when the form contains facet integrals, see #2300.


wence-:
Is this definitively ready for review? (We have merged the dependent PRs because they were simpler)

pbrubeck:
> Is this definitively ready for review? (We have merged the dependent PRs because they were simpler)

I think it is, now that the tests are passing.

pefarrell (approved):
Looks good to me, very minor changes requested

firedrake/preconditioners/pmg.py
```diff
@@ -63,65 +60,16 @@ def coarsen_element(self, ele):
...
-        if N <= self.coarse_degree:
+        degree = PMGBase.max_degree(ele)
+        if degree <= self.coarse_degree:
             raise ValueError
-        return PMGBase.reconstruct_degree(ele, max(N // 2, self.coarse_degree))
+        return PMGBase.reconstruct_degree(ele, max(degree//2, self.coarse_degree))
 
-    @staticmethod
-    def max_degree(ele):
+    def coarsen_form(self, form, fine_to_coarse_map):
```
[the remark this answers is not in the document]
pbrubeck: I am exposing this method so users can override it, as I explain in the docstring.
Say your p-fine space is DG(k) and your p-coarse space is CG(1), then you might want to drop the interior facet integrals in the p-coarse Jacobian.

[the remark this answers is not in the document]
pefarrell: Apologies, I missed it in the diff.

firedrake/preconditioners/asm.py
```diff
@@ -56,6 +58,9 @@ def initialize(self, pc):
             asmpc.setType(asmpc.Type.ASM)
             # Set default solver parameters
             asmpc.setASMType(PETSc.PC.ASMType.BASIC)
+
+            dosort = PETSc.Options().getBool(self.prefix + "sort_indices", default=True)
```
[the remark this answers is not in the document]
pbrubeck: The user must `sort_indices == False` whenever they want to impose an ordering (such as `metisnd`) on the patch. 
If `sort_indices == True` then the effect of `mat_ordering_type` is ignored, and the patches will respect the global DOF ordering.

Should I combine the options `sort_indices` and `mat_ordering_type` into a single one?
I can add the choice `mat_ordering_type: default` that sets `dosort = True`.

firedrake/preconditioners/asm.py
```diff
@@ -224,7 +237,7 @@ def get_patches(self, V):
                         continue
                     off = section.getOffset(p)
                     # Local indices within W
-                    W_indices = numpy.arange(off*W.value_size, W.value_size * (off + dof), dtype=IntType)
+                    W_indices = slice(off*W.value_size, W.value_size * (off + dof))
```
wence-, line 240: This is just so you don't make a big array for indexing.

firedrake/preconditioners/asm.py
```diff
@@ -280,6 +293,19 @@ def get_patches(self, V):
         return ises
 
 
+def order_points(mesh_dm, points, ordering_type, prefix):
+    if ordering_type == "natural":
```
wence-, line 297: Docstring please.

firedrake/preconditioners/asm.py
```diff
@@ -280,6 +293,19 @@ def get_patches(self, V):
...
+        return points
+    subgraph = [numpy.intersect1d(points, mesh_dm.getAdjacency(p), return_indices=True)[1] for p in points]
+    ia = numpy.cumsum([0] + [len(neigh) for neigh in subgraph]).astype(PETSc.IntType)
+    ja = numpy.concatenate(subgraph).astype(PETSc.IntType)
+    A = PETSc.Mat().createAIJ((len(points), )*2, csr=(ia, ja, numpy.ones(ja.shape, PETSc.RealType)), comm=COMM_SELF)
+    A.setOptionsPrefix(prefix)
```
wence-, line 303: Prefix unnecessary since you never `A.setFromOptions()`. I think you do everything programmatically so there's no need to set the prefix (or pass it in).

firedrake/preconditioners/asm.py
```diff
@@ -356,39 +386,45 @@ def get_patches(self, V):
...
+                            dof = basemeshdof[i][p]
+                            # Hard-code taking the star
+                            if plane == 0:
+                                begin = off + k * blayer_offset
+                                end = off + k * blayer_offset + dof
+                            else:
+                                begin = off + min(k, k+plane) * blayer_offset + dof
+                                end = off + max(k, k+plane) * blayer_offset
+                            zlice = slice(W.value_size * begin, W.value_size * end)
+                            indices.extend(iset[zlice])
```
wence-, line 426: OK, so this just reorders the existing star code.

tests/multigrid/test_p_multigrid.py
```diff
@@ -26,6 +26,36 @@ def test_reconstruct_degree():
             assert e == PMGPC.reconstruct_degree(elist[0], degree)
 
 
+def test_prolongation_matrix_matfree():
+    from firedrake.preconditioners.pmg import prolongation_matrix_matfree
+
+    tol = 1E-14
+    meshes = [UnitSquareMesh(3, 2, quadrilateral=True)]
+    meshes.append(ExtrudedMesh(meshes[0], layers=2))
+    for mesh in meshes:
```
wence-, line 35: This test is separately testing the quad and hex mesh cases. Can you fixture-ise these so that it is obvious what is going on?

```python
@pytest.fixture(params=["quad", "hex"])
def mesh(request):
    mesh = UnitSquareMesh(3, 2, quadrilateral=True)
    if request.param == "quad":
        return mesh
    elif request.param == "hex":
       return ExtrudedMesh(mesh, layers=2)
    else:
        raise AssertionError


def test_prolongation_matrix_matfree(mesh):
     ndim = mesh.topological_dimension()
     ....
```

tests/multigrid/test_p_multigrid.py
```diff
@@ -26,6 +26,36 @@ def test_reconstruct_degree():
...
+        elems.append(FiniteElement("NCE" if ndim == 3 else "RTCE", cell=cell, degree=2, variant=variant))
+        fs = [FunctionSpace(mesh, e) for e in elems]
+        us = [Function(V) for V in fs]
+        us[0].interpolate(expr)
+        for u in us:
+            for v in us:
+                if u != v:
+                    v.assign(zero(v.ufl_shape))
+                    P = prolongation_matrix_matfree(v, u).getPythonContext()
+                    P._prolong()
```
wence-, line 55: Why not use the petsc interface (`P.mult`)? Just more typing?

tests/regression/test_fdm.py
```diff
@@ -159,26 +136,28 @@ def test_variable_coefficient(mesh):
 def fs(request, mesh):
     degree = 3
     ndim = mesh.topological_dimension()
+    cell = mesh.ufl_cell()
     element = request.param
+    variant = None
     if element == "rt":
         family = "RTCF" if ndim == 2 else "NCF"
-        return FunctionSpace(mesh, family, degree)
+        return FunctionSpace(mesh, FiniteElement(family, cell, degree=degree, variant=variant))
```
wence-, line 144: Default variant is not `integral` not `point`, is that a problem? I guess not.

wence-: So I _think_ this line is equivalent to the one that it replaces?

firedrake/preconditioners/pmg.py
```diff
@@ -63,65 +60,16 @@ def coarsen_element(self, ele):
...
         """
-        Return the maximum degree of a :class:`ufl.FiniteElement`
+        Coarsen a form, by replacing the solution, test and trial functions.
```
wence-, line 70: Can you document the arguments too please?

firedrake/preconditioners/pmg.py
```diff
@@ -46,7 +42,8 @@ class PMGBase(PCSNESBase):
     The default coarsen_element is to perform power-of-2 reduction
     of the polynomial degree.
     It is expected that some applications of this preconditioner
-    will subclass :class:`PMGBase` to override `coarsen_element`.
+    will subclass :class:`PMGBase` to override `coarsen_element` and
+    `coarsen_form`.
```
wence-, line 46: This PC comes with lots of petsc options to twiddle. Can they be added to the docstring as well?

firedrake/preconditioners/pmg.py
```diff
@@ -213,78 +171,60 @@ def coarsen(self, fdm, comm):
...
+        fine_to_coarse_map = {fu: cu,
+                              test: test.reconstruct(function_space=cV),
+                              trial: trial.reconstruct(function_space=cV)}
+
+        def _coarsen_form(a):
+            if isinstance(a, ufl.Form):
+                a = self.coarsen_form(a, fine_to_coarse_map)
+                a = type(a)([f.reconstruct(metadata=self.coarsen_quadrature(f.metadata(), fdeg, cdeg))
+                             for f in a.integrals()])
+            return a
```
wence-, line 186: I guess this is handling the case where one of `F`, `J` and `Jp` are `None`?

Why does `coarsen_form` not handle the quadrature coarsening?

firedrake/preconditioners/pmg.py
```diff
@@ -340,7 +271,10 @@ def coarsen_nullspace(coarse_V, mat, fine_nullspace):
                 for xf in fine_nullspace._petsc_vecs:
                     wc = firedrake.Function(coarse_V)
                     with wc.dat.vec_wo as xc:
-                        mat.multTranspose(xf, xc)
+                        if mat.getSize()[1] == xf.getSize():
+                            mat.mult(xf, xc)
+                        else:
+                            mat.multTranspose(xf, xc)
```
wence-, line 277: AKA [`MatRestrict`](https://petsc.org/main/src/mat/interface/matrix.c.html#MatRestrict)

firedrake/preconditioners/pmg.py
```diff
@@ -354,18 +288,46 @@ def coarsen_nullspace(coarse_V, mat, fine_nullspace):
         cctx.set_nullspace(cctx._nullspace, ises, transpose=False, near=False)
         cctx._nullspace_T = coarsen_nullspace(cV, I, fctx._nullspace_T)
         cctx.set_nullspace(cctx._nullspace_T, ises, transpose=True, near=False)
-        cctx._near_nullspace = coarsen_nullspace(cV, I, fctx._near_nullspace)
+        cctx._near_nullspace = coarsen_nullspace(cV, injection, fctx._near_nullspace)
         cctx.set_nullspace(cctx._near_nullspace, ises, transpose=False, near=True)
         return cdm
 
+    def coarsen_quadrature(self, metadata, fdeg, cdeg):
```
wence-, line 295: Docstrings for these methods please.

firedrake/preconditioners/pmg.py
```diff
@@ -354,18 +288,46 @@ def coarsen_nullspace(coarse_V, mat, fine_nullspace):
...
+            # Coarsen the quadrature degree in a dictionary
+            # such that the ratio of quadrature nodes to interpolation nodes (qdeg+1)//(fdeg+1) is preserved
+            qdeg = metadata.get("quadrature_degree", None)
+            if qdeg is not None:
+                cmd = dict(metadata)
+                cmd["quadrature_degree"] = max(2*cdeg+1, ((qdeg+1)*(cdeg+1)+fdeg)//(fdeg+1)-1)
+                return cmd
+        return metadata
+
+    def coarsen_bcs(self, fbcs, cV):
```
wence-, line 306: And here.

firedrake/preconditioners/pmg.py
```diff
@@ -482,400 +495,583 @@ def coarsen_bc_value(self, bc, cV):
         return coarse
 
 
-def prolongation_transfer_kernel_aij(Pk, P1):
-    # Works for Pk, Pm; I just retain the notation
-    # P1 to remind you that P1 is of lower degree
-    # than Pk
+def prolongation_transfer_kernel_action(Vf, expr):
```
wence-, line 498: Docstring please.

firedrake/preconditioners/pmg.py
```diff
@@ -482,400 +495,583 @@ def coarsen_bc_value(self, bc, cV):
...
+        hexahedron_tpc = ufl.TensorProductCell(ufl.quadrilateral, ufl.interval)
+        return expand_element(ele.reconstruct(cell=hexahedron_tpc))
+    elif isinstance(ele, (ufl.TensorElement, ufl.VectorElement)):
+        return expand_element(ele._sub_element)
+    elif isinstance(ele, (ufl.HDivElement, ufl.HCurlElement, ufl.BrokenElement, ufl.RestrictedElement)):
+        return expand_element(ele._element)
+    elif isinstance(ele, ufl.WithMapping):
+        return expand_element(ele.wrapee)
+    elif isinstance(ele, ufl.MixedElement):
+        return ufl.MixedElement(*[expand_element(e) for e in ele.sub_elements()])
```
wence-, line 531: This would be clearer I think if you say `type(ele) is ufl.MixedElement` (since you specifically handle `TensorElement` and `VectorElement` above,

firedrake/preconditioners/pmg.py
```diff
@@ -482,400 +495,583 @@ def coarsen_bc_value(self, bc, cV):
...
-    flop_count = kernel.flop_count
-    return op2.Kernel(ast, name, requires_zeroed_output_arguments=True,
-                      flop_count=flop_count)
 
-
-def tensor_product_space_query(V):
+@lru_cache(maxsize=10)
+def expand_element(ele):
+    """
+    Expand a FiniteElement as an EnrichedElement of TensorProductElements, discarding modifiers.
```
wence-, line 516: I guess the goal here is to get a "scalar" element with nodes in the same place as the incoming element?

wence-: AKA: can you expand a little more in this docstring.

firedrake/preconditioners/pmg.py
```diff
@@ -482,400 +495,583 @@ def coarsen_bc_value(self, bc, cV):
...
-    V must be either Q(N) or DQ(N) (same N along every direction),
-    RTCF(N), RTCE(N), NCF(N), or NCE(N) on quads or hexes.
 
-    :arg V: FunctionSpace
-    :returns: 4-tuple of (use_tensorproduct, degree, topological_dimension, sub_families, variant)
+def get_line_elements(V):
+    from FIAT.reference_element import LINE
+    from tsfc.finatinterface import create_element
+    ele = V.ufl_element()
+    if isinstance(ele, ufl.MixedElement) and not isinstance(ele, (ufl.TensorElement, ufl.VectorElement)):
```
wence-, line 562: AKA `if type(ele) is ufl.MixedElement:`

firedrake/preconditioners/pmg.py
```diff
@@ -482,400 +495,583 @@ def coarsen_bc_value(self, bc, cV):
...
+        raise ValueError("MixedElements are not decomposed into tensor products")
+    rvs = ele.reference_value_size()
+    ele = expand_element(ele)
+    if isinstance(ele, ufl.EnrichedElement):
+        ele = ele._elements[0]
```
wence-, line 567: This seems dangerous. Do you want to assert that you only have a singleton enriched element with:
```
ele, = ele._elements
```
?

firedrake/preconditioners/pmg.py
```diff
@@ -482,400 +495,583 @@ def coarsen_bc_value(self, bc, cV):
...
+            return ufl.EnrichedElement(*[ufl.TensorProductElement(*k) for k in terms])
+    else:
+        return ele
 
-    V must be either Q(N) or DQ(N) (same N along every direction),
-    RTCF(N), RTCE(N), NCF(N), or NCE(N) on quads or hexes.
 
-    :arg V: FunctionSpace
-    :returns: 4-tuple of (use_tensorproduct, degree, topological_dimension, sub_families, variant)
+def get_line_elements(V):
```
wence-, line 558: Docstring please.

firedrake/preconditioners/pmg.py
```diff
@@ -482,400 +495,583 @@ def coarsen_bc_value(self, bc, cV):
...
+    for e in reversed(factors):
+        fiat_ele = e.fiat_equivalent
+        if fiat_ele.get_reference_element().shape != LINE:
+            raise ValueError("Expecting %s to be on the interval" % fiat_ele)
+        line_elements.append(fiat_ele)
+    return line_elements
+
+
+@lru_cache(maxsize=10)
+def get_line_interpolator(felem, celem):
```
wence-, line 583: Docstring please.

firedrake/preconditioners/fdm.py
```diff
@@ -70,197 +44,267 @@ class FDMPC(PCBase):
...
-        with self.uc.dat.vec_ro as xc:
-            xc.copy(y)
-
-        y.array_w[self.bc_nodes] = x.array_r[self.bc_nodes]
+    @PETSc.Log.EventDecorator("FDMUpdate")
+    def update(self, pc):
+        if hasattr(self, "A"):
+            self._assemble_A()
+        self._assemble_P()
+        # self.diagonal_scaling()
```
wence-, line 158: Lose the commented code please?

firedrake/preconditioners/fdm.py
```diff
@@ -70,197 +44,267 @@ class FDMPC(PCBase):
...
+
+    def apply(self, pc, x, y):
+        dm = self._dm
+        with dmhooks.add_hooks(dm, self, appctx=self._ctx_ref):
+            if hasattr(self, "fdm_interp"):
+                with self.work.dat.vec as x_fdm, self.diag.dat.vec as y_fdm:
```
wence-, line 164: both of these accesses can be `vec_wo` I think. Or better, just keep `Vec`s hanging around, rather than `Function`s?

firedrake/preconditioners/fdm.py
```diff
@@ -70,197 +44,267 @@ class FDMPC(PCBase):
...
+                y.array_w[self.bc_nodes] = x.array_r[self.bc_nodes]
+            else:
+                self.pc.apply(x, y)
 
     def applyTranspose(self, pc, x, y):
-        self.apply(pc, x, y, transpose=True)
+        dm = self._dm
+        with dmhooks.add_hooks(dm, self, appctx=self._ctx_ref):
+            if hasattr(self, "fdm_interp"):
+                with self.work.dat.vec as x_fdm, self.diag.dat.vec as y_fdm:
```
wence-, line 176: Same here.

firedrake/preconditioners/fdm.py
```diff
@@ -70,197 +44,267 @@ class FDMPC(PCBase):
...
             self.pc.view(viewer)
 
-    @staticmethod
-    def pull_axis(x, pshape, idir):
-        # permute x by reshaping into pshape and moving axis idir to the front
-        return numpy.reshape(numpy.moveaxis(numpy.reshape(x.copy(), pshape), idir, 0), x.shape)
+    def diagonal_scaling(self):
+        _, P = self.pc.getOperators()
+        self._assemble_diag()
+        with self.diag.dat.vec as x_, self.work.dat.vec as y_:
```
wence-, line 193: And here

firedrake/preconditioners/fdm.py
```diff
@@ -465,678 +490,357 @@ def assemble_coef(self, J, quad_deg, discard_mixed=False, cell_average=False, ne
...
-        Setup for the GLL finite element method on the reference interval
 
-        :arg N: polynomial degree of the GLL element
-        :arg Nq: quadrature degree (Nq >= 2*N+1 gives exact integrals)
+def get_piola_tensor(mapping, domain):
+    tdim = domain.topological_dimension()
+    if mapping == 'identity':
+        return None
+    elif mapping == 'covariant piola':
+        return ufl.JacobianInverse(domain).T * firedrake.Constant(numpy.flipud(numpy.identity(tdim)), domain=domain)
```
wence-, line 589: For these actually constant matrices, you could just use `ufl.as_tensor(some_literal_array)` rather than `firedrake.Constant`.

wence- (commented):
I see that this has now been merged anyway, but if you could address some of the comments in a separate cleanup PR that would be great.

Generally I think this code is fine since it is self-contained to the two PC implementations. It is implementing, basically, an entire auxiliary assembler which we cannot guarantee will not be broken by changes in the Firedrake data structures at a future date.

I would recommend conversing with @connorjward to figure out what it is you need from a more extensible assembler to be able to do much of this without handcoding everything. Is it just  performance matter, a composability matter, or something else?

firedrake/preconditioners/asm.py
```diff
@@ -125,12 +134,13 @@ class ASMStarPC(ASMPatchPC):
...
         mesh_dm = mesh.topology_dm
-        if mesh.layers:
+        if mesh.cell_set._extruded:
             warning("applying ASMStarPC on an extruded mesh")
 
         # Obtain the topological entities to use to construct the stars
         depth = PETSc.Options().getInt(self.prefix+"construct_dim", default=0)
-
+        ordering = PETSc.Options().getString(self.prefix+"mat_ordering_type",
+                                             default="natural")
```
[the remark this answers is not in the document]
pbrubeck: Correct. If `ordering == "natural" `or if it is not provided, then the indices are not permuted.

firedrake/preconditioners/asm.py
pbrubeck: Yes

firedrake/preconditioners/fdm.py
pbrubeck: Good suggestion. I'll go for the `Vec` approach.

PR #2305, opened 2021-12-19
Check backward compatibility of UFL dualspace

nbouziani:
This PR is just for checking if the upcoming UFL dual space update breaks the current Firedrake.

UFL PR: https://github.com/firedrakeproject/ufl/pull/25


nbouziani:
This currently fails because `firedrake.Constant` inherits `ufl.Coefficient` and the latter performs some checks on the function space (if it is dual or not) in `__new__`. This has been fixed in #2294 and is I think more of a design issue at the Firedrake level (we could have `firedrake.Constant` inheriting `ufl.Constant`). It is something that is (I think) specific to Firedrake so it won't affect other UFL subpackages that don't have this design.

PR #2306, opened 2022-01-01
Delete unnecessary `inserted_options`

lrtfm:
Delete `self.inserted_options()` in `solve` as it has been called in function `__init__` by `self.set_from_options(self.ksp)`.

### DIFF
--- a/firedrake/linear_solver.py
+++ b/firedrake/linear_solver.py
@@ -156,7 +156,7 @@ class LinearSolver(OptionsManager):
         else:
             acc = x.dat.vec_wo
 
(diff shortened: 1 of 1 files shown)


lrtfm (commented):
The `setFromOptions` will not set up the sub ksp because the ksp have not been setup, so the inserted_options() before solve is necessary.

wence-:
although setfromoptions has been called, I think the inserted options are still needed so that ksp_view and friends work. iirc

lrtfm:
> although setfromoptions has been called, I think the inserted options are still needed so that ksp_view and friends work. iirc

@wence-  Thank you for the relative information.

PR #2313, opened 2022-01-13
Remove scipy from wheel blacklist

JDBetteridge:


JDBetteridge:
This commit is to be undone once we have a solution to scipy not building correctly

dham (approved):

PR #2317, opened 2022-01-25
Projection: allow projections on trace spaces

sv2518:
I want to be able to project functions into the HDiv trace space. This works by solving something like `\int_{\Gamma} \gamma \lambda \diff S = \int_{\Gamma} \gamma \lambda_0 \diff S, \forall \gamma \in T` where `lambda_0` is what we project (kudos to @colinjcotter).

The variational problem changes if we are on an extruded mesh or not. I am only testing this for tets so that I can use interpolating to get a reference solution, but I could set up a reference solver if we want to tests for quads too. In the calculation of the norm I am multiplying by the area,  so that it doesn't diverge as the mesh size goes to zero (also kudos to @colinjcotter).

If you can think of cases where the math behind this doesn't work I can include assertions.

### DIFF
--- a/firedrake/projection.py
+++ b/firedrake/projection.py
@@ -7,6 +7,7 @@ from firedrake.utils import cached_property, complex_mode, SLATE_SUPPORTS_COMPLE
 from firedrake import functionspaceimpl
 from firedrake import function
 from firedrake.adjoint import annotate_project
+from finat import HDivTrace
 
 
 __all__ = ['project', 'Projector']
@@ -52,6 +53,8 @@ def project(v, V, bcs=None,
             name=None,
             ad_block_tag=None):
     """Project a UFL expression into a :class:`.FunctionSpace`
+    It is possible to project onto the trace space 'DGT', but not onto
+    other trace spaces e.g. into the restriction of CG onto the facets.
 
     :arg v: the :class:`ufl.Expr` to project
     :arg V: the :class:`.FunctionSpace` or :class:`.Function` to project into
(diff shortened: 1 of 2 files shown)


firedrake/projection.py
```diff
@@ -163,7 +177,21 @@ class BasicProjector(ProjectorBase):
...
         v = firedrake.TestFunction(self.target.function_space())
-        form = firedrake.inner(self.source, v)*firedrake.dx
+        F = self.target.function_space()
+        mixed = isinstance(F.ufl_element(), ufl.MixedElement)
+        if not mixed and isinstance(F.finat_element, HDivTrace):
+            if F.extruded:
+                form = (firedrake.inner(self.source, v)*firedrake.ds_t
+                        + firedrake.inner(self.source, v)*firedrake.ds_v
+                        + firedrake.inner(self.source, v)*firedrake.ds_b
+                        + firedrake.inner(self.source('+'), v('+'))*firedrake.dS_h
```
dham, line 187: Average of source because it could legitimately be discontinuous.

firedrake/projection.py
```diff
@@ -163,7 +177,21 @@ class BasicProjector(ProjectorBase):
...
+                        + firedrake.inner(self.source('+'), v('+'))*firedrake.dS_v)
+            else:
+                form = (firedrake.inner(self.source, v)*firedrake.ds
+                        + firedrake.inner(self.source('+'), v('+'))*firedrake.dS)
```
dham, line 191: also average source here.

dham (changes requested):
Please document what this does in the `project` and/or `Projector` documentation.

firedrake/projection.py
sv2518: We probably want to change this here 
https://github.com/firedrakeproject/firedrake/blob/1fafb01f15f817bece2cbabddd1367b2db7bc151/tests/extrusion/test_trace_extr.py#L24 and here https://github.com/firedrakeproject/firedrake/blob/1fafb01f15f817bece2cbabddd1367b2db7bc151/tests/regression/test_trace_galerkin_projection.py#L59 too?

sv2518: Ah, actually I can dismiss this, both tests define the source in a continuous space.

dham (approved):

Issue #2314, opened 2022-01-15
honour-petsc4py


qbisi:
I am using a self built petsc package installed in /opt/petsc/intel-mkl-real with chmod 755 owned by root. Also built with petsc4py support.
Then i configure firedrake with (user privilage):
python3.9 firedrake-install --disable-ssh --no-package-manager --mpicc mpiicc --mpicxx mpiicpc --mpif90 mpiifort --mpiexec mpiexec --honour-pythonpath --honour-petsc-dir.
Still the firedrake-install script tries to rebuild petsc4py thought it is prebuilt in PYTHONPATH.
If firedrake insists build petsc4py from source ,it requires write privileges' to /opt/petsc/src/bindings/petsc4py ,which is not allow for a shared package. 
I suggest detect petsc4py in pythonpath first ,if not exsits ,copy and build the petsc4py source file (or download from web) in the cache directory

wence-:
Thanks, this is, I think, because since pip version 21.3, `pip install local_directory` does an in-tree build (see the discussion in pypa/pip#7555, and documentation https://pip.pypa.io/en/latest/cli/pip_install/#local-project-installs). The backwards compatibility option is going away in pip 22.1, so we can use the stop-gap proposed there but not for long.

It seems that one way to fix this would be to migrate petsc4py away from a setuptools-based install procedure to something else, but I have lost track of current state of the art in python packaging.

Probably the right workaround is to do as you suggest and copy before installing: @JDBetteridge can you give this a go?

JDBetteridge:
Hmm, I see the dilemma. I'll look into a possible fix for an uneditable system PETSc/petsc4py build of Firedrake

JDBetteridge:
Please try the install script in this PR https://github.com/firedrakeproject/firedrake/pull/2326

Can be obtained with `curl -O https://raw.githubusercontent.com/firedrakeproject/firedrake/cb3eb69e23ebb4a3fc97f43b057661a52fa0c633/scripts/firedrake-install`

JDBetteridge:
Fixed in #2326

Issue #2321, opened 2022-01-26
h5py installation issue


VictorEijkhout:
```
Installing h5py/
Traceback (most recent call last):
  File "firedrake-install", line 1668, in <module>
    build_and_install_h5py()
  File "firedrake-install", line 1079, in build_and_install_h5py
    install("h5py/")
  File "firedrake-install", line 896, in install
    run_pip_install(["--no-build-isolation", package])
  File "firedrake-install", line 831, in run_pip_install
    check_call(pipinstall + pipargs)
  File "firedrake-install", line 543, in check_call
    log.debug(subprocess.check_output(arguments, stderr=subprocess.STDOUT, env=os.environ).decode())
  File "/opt/local/Library/Frameworks/Python.framework/Versions/3.8/lib/python3.8/subprocess.py", line 415, in check_output
    return run(*popenargs, stdout=PIPE, timeout=timeout, check=True,
  File "/opt/local/Library/Frameworks/Python.framework/Versions/3.8/lib/python3.8/subprocess.py", line 516, in run
    raise CalledProcessError(retcode, process.args,
subprocess.CalledProcessError: Command '['/Users/eijkhout/Installation/firedrake/firedrake/bin/python', '-m', 'pip', 'install', '--no-build-isolation', '--no-binary', 'mpi4py,randomgen,islpy,numpy', '-vvv', '--no-build-isolation', 'h5py/']' returned non-zero exit status 1.
```

The log ends in:

```
  running build_ext
  Loading library to get version: libhdf5.dylib
  error: Unable to load dependency HDF5, make sure HDF5 is installed properly
  error: dlopen(libhdf5.dylib, 6): image not found
  Building wheel for h5py (pyproject.toml): finished with status 'error'
  ERROR: Failed building wheel for h5py
Failed to build h5py
ERROR: Could not build wheels for h5py, which is required to install pyproject.toml-based projects
Exception information:
Traceback (most recent call last):
  File "/Users/eijkhout/Installation/firedrake/firedrake/lib/python3.8/site-packages/pip/_internal/cli/base_command.py", line 164, in exc_logging_wrapper
    status = run_func(*args)
  File "/Users/eijkhout/Installation/firedrake/firedrake/lib/python3.8/site-packages/pip/_internal/cli/req_command.py", line 205, in wrapper
    return func(self, options, args)
[...]

PR #2326, opened 2022-02-01
Workaround for read only petsc4py directory

JDBetteridge:
Addresses issue #2314 

(as yet untested)

### DIFF
--- a/scripts/firedrake-install
+++ b/scripts/firedrake-install
@@ -887,7 +887,7 @@ def install(package):
     elif package == "slepc/":
         build_and_install_slepc()
     elif package == "petsc4py/":
-        run_pip_install(["--ignore-installed", get_petsc4py_dir()])
+        build_and_install_petsc4py()
     elif package == "slepc4py/":
         # The following outrageous hack works around the fact that petsc and co. cannot be installed in developer mode.
         run_pip_install(["--ignore-installed", get_slepc4py_dir()])
@@ -1028,6 +1028,21 @@ def build_and_install_petsc():
             test_output = os.path.join(petsc_dir, "src", "**", "tests", "output", "*")
             for deletefile in chain(iglob(tutorial_output, recursive=True), iglob(test_output, recursive=True)):
                 check_call(["rm", "-f", deletefile])
+
+
+def build_and_install_petsc4py():
+    with environment(**blas):
+        pipargs = ["-vvv", "--ignore-installed", get_petsc4py_dir()]
+        try:
+            check_output(pipinstall + pipargs)
+        except subprocess.CalledProcessError:
+            # This assumes the CalledProcessError is a result of readonly petsc4py source
+            log.info("Could not perform in-place install of petsc4py in current location")
+            log.info("Copying petsc4py to virtualenv and trying again...")
+            petsc4py_copy = os.path.join(os.environ['VIRTUAL_ENV'], 'src', 'petsc4py')
+            shutil.copytree(get_petsc4py_dir(), petsc4py_copy, dirs_exist_ok=True)
+            pipargs = ["-vvv", "--ignore-installed", petsc4py_copy]
+            check_output(pipinstall + pipargs)
 
 
 def build_and_install_h5py():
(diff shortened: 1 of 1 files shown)


JDBetteridge:
Okay, tested working locally with system installed PETSc (read only petsc4py directory).

This should also resolve the issue of installing petsc4py in #2321 , please try installing with this fix and we can iterate again.

scripts/firedrake-install
```diff
@@ -1030,6 +1030,21 @@ def build_and_install_petsc():
                 check_call(["rm", "-f", deletefile])
 
 
+def build_and_install_petsc4py():
+    with environment(**blas):
+        pipargs = ["-vvv", "--ignore-installed", get_petsc4py_dir()]
```
dham, line 1035: Actually check whether the dir is writable.

dham (changes requested):

scripts/firedrake-install
```diff
@@ -1030,6 +1031,31 @@ def build_and_install_petsc():
...
+            log.info("Copying petsc4py to virtualenv and trying again...")
+            petsc4py_copy = os.path.join(os.environ['VIRTUAL_ENV'], 'src', 'petsc4py')
+            if os.path.exists(petsc4py_copy):
+                petsc4py_backup = os.path.join(os.environ['VIRTUAL_ENV'], 'src', 'petsc4py.backup')
+                if os.path.exists(petsc4py_backup):
+                    raise OSError('`{}` exists remove or rename this file/directory and try running this command again'.format(petsc4py_backup))
+                shutil.move(petsc4py_copy, petsc4py_backup)
+            shutil.copytree(get_petsc4py_dir(), petsc4py_copy, dirs_exist_ok=True)
+            S_644 = stat.S_IRUSR | stat.S_IWUSR | stat.S_IRGRP | stat.S_IROTH
+            S_755 = stat.S_IRUSR | stat.S_IWUSR | stat.S_IXUSR | stat.S_IRGRP | stat.S_IXGRP | stat.S_IROTH | stat.S_IXOTH
```
connorjward, line 1050: Is there not a nicer way to do this? I came across this [octal representation](https://docs.python.org/3/library/pathlib.html#pathlib.Path.chmod) which might be cleaner.

scripts/firedrake-install
```diff
@@ -1030,6 +1031,31 @@ def build_and_install_petsc():
...
+            check_output(pipinstall + pipargs)
+        else:
+            log.info("Could not perform in-place install of petsc4py in current location")
+            log.info("Copying petsc4py to virtualenv and trying again...")
+            petsc4py_copy = os.path.join(os.environ['VIRTUAL_ENV'], 'src', 'petsc4py')
+            if os.path.exists(petsc4py_copy):
+                petsc4py_backup = os.path.join(os.environ['VIRTUAL_ENV'], 'src', 'petsc4py.backup')
+                if os.path.exists(petsc4py_backup):
+                    raise OSError('`{}` exists remove or rename this file/directory and try running this command again'.format(petsc4py_backup))
+                shutil.move(petsc4py_copy, petsc4py_backup)
```
wence-, line 1047: I would much rather do the following:

1. Check if source dir is writeable (ok, just go as normal)
2. If not, copy somewhere (really not sure why you need to do the dance with mode changes [if I copy something just at the commandline, it gets the correct permissions], can you explain?)
3. Build and install
4. If the build is successful, remove the copied directory
5. Otherwise leave it around for debugging [but put a README file in it that indicates that this is not the actual place that the source code lives]

WDYT?

JDBetteridge: I think this is a good idea, I think I would modify the procedure slightly based on trying to install from PETSc installed by a package manager/system administrator. It turns out checking to see if the source is writeable isn't sufficient for the install to complete. For instance if build artefacts are present in petsc4py left over from any other user building petsc4py in the same writeable directory. This seems especially relevant given the issues that prompted this PR.

My modification would be:
1. Always copy the source dir to the venv (rather than trying to program every possible exception)
Then 3, 4 and 5 as above.

I'm not sure there is an automatic solution to update petsc4py if the system PETSc changes, users will have to use the `--rebuild` flag

PR #2333, opened 2022-02-04
Pin pytest version to 6.2.5

connorjward:
Our parallel tests break with [new release](https://pypi.org/project/pytest/#history) 7.0.0. Obviously we should figure out a better fix but this should get our CI up and running again.

Turns out it wasn't my fault after all!

### DIFF
--- a/requirements-ext.txt
+++ b/requirements-ext.txt
@@ -1,6 +1,6 @@
 sympy
 cachetools
-pytest
+pytest<=6.2.5
 pytest-xdist
 ipython
 matplotlib


ksagiyam:
`firedrake-update` will update `pytest`. Shall we pin version there as well?

connorjward:
> `firedrake-update` will update `pytest`. Shall we pin version there as well?
> 
> 

That's a good idea. It looks like your PR that you closed did this already. Should we just merge that one and close this?

ksagiyam:
That is also fine. Reopend https://github.com/firedrakeproject/firedrake/pull/2334.

PR #2338, opened 2022-02-07
Install Script Documentation Proposal

andreichalapco:
Hi,

I've currently made a first proposal for the install script documentation where I've addressed the following topics:

1. System dependencies
2. Building and configure PETSc
3. Installing Python dependencies

I know is currently in a raw version, but I wanted to get a little bit of feedback in order for me to know if I'm on the right path with writing this documentation.

### DIFF
--- a/docs/source/install-script.rst
+++ b/docs/source/install-script.rst
@@ -1,0 +1,141 @@
+:orphan: true
+
+.. role:: bash(code)
+   :language: bash
+
+=================
+Firedrake Install
+=================
+
+The structure of the Firedrake install process consists of the following steps:
+
+1. Installing system dependencies
+2. Building and configuring PETSc
+3. Installing Python dependencies  
+
+These steps are only supported for for Linux and MacOS distributions. 
+Further we will take the above steps and discuss them in detail. 
+
+Installing system dependencies
+-------------------------------
+
+Linux
+~~~~~
+
+For Linux distributions the system dependencies are installed via the :bash:`apt-get` package manager. The necessary dependencies for running default firedrake are:
+
+- build-essential
+- autoconf
+- automake
+- bison
+- flex
+- cmake
+- gfortran
+- git
+- libtool
+- python3-dev
+- python3-pip
+- python3-tk
+- python3-venv
+- zlib1g-dev
+- libboost-dev
+- libopenblas-dev (if no configuration is provided by the user)
+
+As a small observation, packages like *bison* and *flex* are used for *PT-Scotch* (tool for efficient parallel graph ordering).
+
+MacOS
+~~~~~
+
+For MacOs distributions (based on Darwin) the install script uses the :bash:`homebrew` package manager to install the following necessary dependencies:
+
+- `gcc`
+- `autoconf`
+- `make`
+- `automake`
+- `cmake`
+- `libtool`
+- `boost`
+- `openblas` (if no configuration is provided by the user)
+
+Additionally the install script will also want to install xcode-command line tools.
+
+After the installation of the system requirements is done, the installation script will then create a virtual python environment and prepare itself to install all its software dependencies there.
+
+Building and configuring PETSc
+-------------------------------
+
+One important package on which Firedrake software is built is the `PETSc_ <https://petsc.org/release/>`_ library, which offers a toolkit when dealing with partial differential equations. 
+The install script enables some customization for installing the PETSc, but also offers the user the option to use a previous installation of the library via the parameter :bash:`hounour_petsc`.
+If no previous previous installation is provided, the installation script will then build the PETSc package from source in the *src*  directory within the virtual python environment.
+
+The basic options for installing the library using the install argument :bash:`--minimal_petsc` are:
+
+.. code-block:: python
+
+   petsc_options = { "--with-fortran-bindings=0",
+                     "--with-debugging=0",
+                     "--with-shared-libraries=1",
+                     "--with-c2html=0",
+                     "--download-eigen=%s/src/eigen-3.3.3.tgz " % firedrake_env,
+                     # File format
+                     "--download-hdf5",
+                     # AMG
+                     "--download-hypre",
+                     # Sparse direct solver
+                     "--download-superlu_dist",
+                     # Parallel mesh partitione[]()r
+                     "--download-ptscotch",
+                     # For superlu_dist amongst others.
+                     "--with-cxx-dialect=C++11"}
+
+This option is useful if we want to have a faster build and therefore is very useful for testing.
+For a more detail explanation of the parameters please check the PETSc install documentation (page)[https://petsc.org/release/install/].
+
+If no minimal version of PETSc is required (:bash:`--minimal_petsc false`):
+
+- :bash:`--with-zlib`
+- :bash:`--download-netcdf` (file format)
+- :bash:`--download-pnetcdf` (file format) 
+- :bash:`--download-suitesparse` (sparse direct solver)
+- :bash:`--download-pastix` (sparse direct solver) 
+- :bash:`--download-hwloc` (required by pastix package)
+- :bash:`--download-metis` (serial mesh partitioner)
+
+will be added as argument to the PETSc installation. 
+
+Further the install script also provides the option for the user to decide between integer type for the PETsc (:bash:`--petsc-int-type`) and also choose if he want to use complex numbers via (:bash:`--complex`) argument.
+
+Due to the fact that PETSc is a parallel program, the firedrake-install script offers the user the option to provide his *MPI wrappers* (:bash:`--mpiexec --mpicc --mpicxx --mpif90`) for the C, C++ and Fortran compiler that are going to be used to build the program. If none are provided, the script will take care to download the `mpich`.
+
+Lastly, the script also provides the possibility via the :bash:`--with_blas` argument to  download or use an existing version of BLAS (Basic Linear Algebra Subprograms).
+
+Installing Python dependencies
+------------------------------
+
+After the installation of PETSc, the install script finally starts installing `Firedrake <https://github.com/firedrakeproject/firedrake>`_ alongside it's Python dependencies (as they can be found in the *requirements-git.txt*):
+
+- `ufl <https://github.com/firedrakeproject/ufl>`_
+- `fiat <https://github.com/firedrakeproject/fiat>`_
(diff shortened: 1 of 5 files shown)


docs/source/install-script.rst
```diff
@@ -0,0 +1,141 @@
...
+These steps are only supported for for Linux and MacOS distributions. 
+Further we will take the above steps and discuss them in detail. 
+
+Installing system dependencies
+-------------------------------
+
+Linux
+~~~~~
+
+For Linux distributions the system dependencies are installed via the :bash:`apt-get` package manager. The necessary dependencies for running default firedrake are:
```
connorjward, line 25: You should probably say here that if your distribution does not use `apt-get` then you can pass the `--no-package-manager` flag but then it is your responsibility to ensure that these dependencies are installed.

docs/source/install-script.rst
```diff
@@ -0,0 +1,141 @@
...
+- gfortran
+- git
+- libtool
+- python3-dev
+- python3-pip
+- python3-tk
+- python3-venv
+- zlib1g-dev
+- libboost-dev
+- libopenblas-dev (if no configuration is provided by the user)
```
connorjward, line 42: Should these also use backticks like the MacOS packages below?

docs/source/install-script.rst
```diff
@@ -0,0 +1,141 @@
...
+
+Additionally the install script will also want to install xcode-command line tools.
+
+After the installation of the system requirements is done, the installation script will then create a virtual python environment and prepare itself to install all its software dependencies there.
+
+Building and configuring PETSc
+-------------------------------
+
+One important package on which Firedrake software is built is the `PETSc_ <https://petsc.org/release/>`_ library, which offers a toolkit when dealing with partial differential equations. 
+The install script enables some customization for installing the PETSc, but also offers the user the option to use a previous installation of the library via the parameter :bash:`hounour_petsc`.
```
connorjward, line 68: `--honour-petsc-dir`

docs/source/install-script.rst
```diff
@@ -0,0 +1,141 @@
...
+If no previous previous installation is provided, the installation script will then build the PETSc package from source in the *src*  directory within the virtual python environment.
+
+The basic options for installing the library using the install argument :bash:`--minimal_petsc` are:
```
connorjward, line 71: `--minimal-petsc`

docs/source/install-script.rst
```diff
@@ -0,0 +1,141 @@
...
+                     "--download-superlu_dist",
+                     # Parallel mesh partitione[]()r
+                     "--download-ptscotch",
+                     # For superlu_dist amongst others.
+                     "--with-cxx-dialect=C++11"}
+
+This option is useful if we want to have a faster build and therefore is very useful for testing.
+For a more detail explanation of the parameters please check the PETSc install documentation (page)[https://petsc.org/release/install/].
+
+If no minimal version of PETSc is required (:bash:`--minimal_petsc false`):
```
connorjward, line 94: I'm not sure what you mean by `--minimal_petsc false`. To have a non-minimal install you just need to omit the flag.

connorjward (changes requested):
Thanks for this. I've left a few comments.
One thing that concerns me is that this adds a new location where we need to keep information up to date. Currently we already declare the system dependencies on the [install page](https://firedrakeproject.org/download.html#system-requirements) and obviously in the install script itself.

dham (commented):
Thanks for this work. I have a structural concern about this, which is that this documentation of specific dependencies outside of the install script is exceptionally likely to go stale really fast. In essence, developers who change our dependencies will forget to change the website. I think that the relevant parts of this page should instead be generated by running the install script with appropriate options. We already do this with the list of supported finite elements on the solving variational problems page. How this is achieved can be seen in the documentation Makefile.

docs/source/install-script.rst
andreichalapco: Ok I will make the modification. 
Regarding the system dependencies, I also saw this issue. Do you have any ideas on how to make this information not redundant?

docs/source/install-script.rst
andreichalapco: Or I can just remove the backticks from the MacOS packages :). forgot to format

docs/source/install-script.rst
andreichalapco: You're right!

docs/source/install-script.rst
andreichalapco: I think this is also what I wrote

docs/source/install-script.rst
andreichalapco: Yeah you're right. I will fix this. (I missinterpreted the parse arguments of the install script)

andreichalapco:
Hi I will take a look at the documentation and try to fix by running the isntall script with appropiate options. Thx for the tip

docs/source/install-script.rst
connorjward: You used an underscore instead of a hyphen.

andreichalapco: ooo :)

andreichalapco:
Hi,

I've made here a new commit which proposes a yaml file for the install script in order to extract from there all the needed install configurations. With this new yaml file, I've manage to create a python script that creates at every `make html` command .rst files containing the system dependencies, wheel_blacklist and the parallel packages that need to be specially compiled. By doing so we have a more dynamic (configuration file dependent) documentation

scripts/install_configuration.yaml
```diff
@@ -0,0 +1,41 @@
+system_requirements_mac:
```
dham, line 1: I like this idea, but it currently fails the single source of truth test in that the install script doesn't read this file, so this doesn't actually fix the problem of the install script changing and the documentation failing to keep up. 

This is probably me being unclear. What I had in mind is that the documentation would run the install script passing appropriate options that will make the install script print out what it would have installed. For example `firedrake-install --show-petsc-configure-options` already does this for PETSc options, and `firedrake-install --show-dependencies` very nearly does what is needed for packages. I don't think there is currently a way of outputting the wheel blacklist, but it wouldn't be too hard.

andreichalapco: Yes, but building then the documentation will take a lot of time or am I missing the point. (we currently just have the install and update mode). And the second problem would be the difference between the mac and linux configuration.

dham: If you run the install script with the options provided, it prints out the information required and exits immediately. It doesn't run an actual install.

The difference between Mac and Linux is what I was referring to when I said that `firedrake-install --show-dependencies` very nearly does what is needed. I suggest adding an optional argument to the `--show-dependencies` flag to enable the platform to be specified (instead of being detected as it is now).

dham: I should also clarify that the issue with having the install script read the yaml file is that the install script gets downloaded and used separately from the git repository. There is no way for the install script to know which version of the yaml file to use. This is why it's important that the install script itself is the single source of truth, and that the documentation queries the install script.

andreichalapco: Ok. I understand. I also thought that you want a singular file that should do all the work. Currently the why I have managed to introduce dynamic content is by creating .rst files. Should I run the install script and create this .rst files? I think this should be a good solution.

andreichalapco:
Hi,

I've made an update now in the install-script. The show-dependencies will now also create a yaml file with the dependencies ... from this file now we can construct our .rst files and load the dynamic content in the documentation.

dham:
> Hi,
> 
> I've made an update now in the install-script. The show-dependencies will now also create a yaml file with the dependencies ... from this file now we can construct our .rst files and load the dynamic content in the documentation.

This is nice except that we don't really want to communicate via the file system, and we don't really want to drop files into the current folder if a user passes the `--show-dependencies` flag. However, I think there is an easy fix for this. If you just have `--show-dependencies` dump YAML to stdout with the user explanation in comments then I think you get exactly what we need in both cases.

andreichalapco:
It's now working via a `subprocess.check_output` call on the firedrake-install script. So no .yaml file is created.

scripts/install_configuration.yaml
connorjward, line 1: I think you can now delete this file.

scripts/firedrake-install
```diff
@@ -17,6 +17,7 @@ from itertools import chain
 import re
```
connorjward, line 17: You don't appear to have actually add the `--show-dependencies` functionality to this file yet.

scripts/firedrake-install
```diff
@@ -1430,6 +1438,26 @@ if mode == "install" or not args.update_script:
             log.info("\nThey will be installed by %s unless --no-package-manager is used." % (package_manager))
             if osname == "Darwin":
                 log.info("\nIn addition, the command line tools will be installed using xcode")
+
+        # getting the needed information for the documnetation
```
connorjward, line 1442: Typo: documentation

scripts/firedrake-install
```diff
@@ -1430,6 +1438,26 @@ if mode == "install" or not args.update_script:
...
+        dependencies_logs = {}
+        dependencies_logs['system_requirements_mac'] = required_packages_mac
+        dependencies_logs['system_requirements_linux'] = required_packages_linux
+
+        dependencies_logs['wheel_blacklist'] = wheel_blacklist
+
+        dependencies_logs['parallel_packages'] = parallel_packages
```
connorjward, line 1449: Better style to write:
```py
dependencies_logs = {"system_requirements_mac": required_packages_mac,
                     "system_requirements_linux": required_packages_linux,
                     ...}
```

docs/source/install-script.rst
```diff
@@ -0,0 +1,108 @@
+:orphan: true
```
connorjward, line 1: I'm not completely sure what the purpose of this file is. Most of our users just want to know to run `python3 firedrake-install`. They aren't necessarily interested in the exact process followed by the installation. Personally I feel like the information you've added to this file should be moved to a couple of different places:
- The overview of the installation process should be added to a docstring for `firedrake-install`.
- The automatically generated list of dependencies could be used to improve [this part](https://firedrakeproject.org/download.html#system-requirements) of the website (possibly also including the PETSc configure options).
- Any other information, for example about `parallel_packages` and `wheel_blacklist`, could be added as additional commenting of `firedrake-install` (possibly refactoring at the same time).

connorjward (changes requested):

PR #2337, opened 2022-02-07
Restore Python 3.6 compatibility

connorjward:
This PR repairs a couple of breakages I accidentally introduced with #2172.

PR #2339, opened 2022-02-07
Set CI Python version to 3.7

connorjward:
Currently our CI builds with Python 3.8 even though Firedrake is supposed to work with 3.7. I think it makes sense to run our CI with the oldest supported version to avoid compatibility issues like #2337.

In this PR I have:
- Modified our CI to run using Python 3.7.
- Updated the website to say that 3.7 is the minimum supported version.
- Tried to improve `firedrake-install` a little so making future changes to this should be easier.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -40,7 +40,7 @@ jobs:
       - name: Setup python
         uses: actions/setup-python@v2
         with:
-          python-version: 3.8
+          python-version: 3.7
       - name: Build Firedrake
         run: |
           cd ..
--- a/docs/source/download.rst
+++ b/docs/source/download.rst
@@ -118,9 +118,8 @@ packages can be installed into an existing Firedrake installation using
 System requirements
 -------------------
 
-Firedrake requires Python 3.6.x to 3.8.x (many externally managed
-dependencies such as VTK have yet to create binary wheels for 3.9.x).
-The installation script is tested on Ubuntu and MacOS X. On Ubuntu 18.04
+Firedrake requires Python 3.7.x to 3.10.x.
+The installation script is tested on Ubuntu and MacOS X. On Ubuntu 20.04
 or later, the system installed Python 3 is supported and tested. On
 MacOS, the homebrew_ installed Python 3 is supported and tested::
 
--- a/scripts/firedrake-install
+++ b/scripts/firedrake-install
@@ -1539,7 +1539,7 @@ if mode == "install":
         log.debug("Removing get-pip.py")
         os.remove("get-pip.py")
     # Ensure pip, setuptools and wheel are at the latest version.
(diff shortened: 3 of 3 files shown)


wence-:
Note that the 3.6 series was end-of-lifed in [september last year](https://www.python.org/downloads/release/python-3615/), and 3.7 is in security fix only mode. I think we could plausibly have a discussion about what the oldest version we want to support is.

connorjward:
> Note that the 3.6 series was end-of-lifed in [september last year](https://www.python.org/downloads/release/python-3615/), and 3.7 is in security fix only mode. I think we could plausibly have a discussion about what the oldest version we want to support is.

Sure. I'll add this to the meeting agenda for this week.

connorjward:
Found the error:
```
2022-02-11T13:09:33.2260777Z   /usr/bin/ld: cannot find -lpython3.7m
2022-02-11T13:09:33.2260893Z   collect2: error: ld returned 1 exit status
```
I think the issue is that we are installing the wrong `python3-dev` package inside our [Docker container](https://github.com/firedrakeproject/firedrake/blob/master/docker/Dockerfile.env#L16) since the container uses 20.04 which uses Python 3.8. After some digging it looks like the most 'standard' way to get other versions of this package is to use [this ppa](https://launchpad.net/~deadsnakes/+archive/ubuntu/ppa).

I think that we have 4 possible ways forward:
- Modify our Docker container to install the extra Python version that we want.
- Use Ubuntu 18.04 instead of 20.04 for the container.
- Set our minimum supported version to 3.8. This would be inconvenient for some (e.g. COLAB) who still use 3.7.
- Do nothing. This would prevent us catching compatibility issues like I encountered in #2337.

What do people think? @dham @JDBetteridge @wence-

JDBetteridge:
I think there is a question as to what we are actually testing. As I understood it we use the docker container for Unbuntu 20.04 unmodified and using official package repositories as it represents a "default" install. To that end I don't think adding non-standard ppa's is a good idea.

I recall updating the docker image from 18.04 to 20.04, but for the life of me can't remember why, but it was probably due to outdated packages.

Since we have the capacity on the runners I propose that we could run a matrix of tests python 3.7-3.10 on Linux (to start and Mac later) perhaps weekly (overnight on a Sunday) to enable us to pick up such errors. For such tests we could use the Python docker images designed specifically for testing against different versions of Python.

connorjward:
> I think there is a question as to what we are actually testing. As I understood it we use the docker container for Unbuntu 20.04 unmodified and using official package repositories as it represents a "default" install. To that end I don't think adding non-standard ppa's is a good idea.
 
Sure.

> Since we have the capacity on the runners I propose that we could run a matrix of tests python 3.7-3.10 on Linux (to start and Mac later) perhaps weekly (overnight on a Sunday) to enable us to pick up such errors. For such tests we could use the Python docker images designed specifically for testing against different versions of Python.

I like this idea. And weekly seems reasonable.

Issue #2343, opened 2022-02-09
BLASLAPACK_LIB not defined in complex mode


pbrubeck:
Is there any particular reason why `BLASLAPACK_LIB` and `BLASLAPACK_INCLUDE` are undefined if `complex_mode == True` other than SLATE not being implemented in complex mode yet?

https://github.com/firedrakeproject/firedrake/blob/85dd96dfcb6fefb745bf7a79829ac92489758297/firedrake/slate/slac/compiler.py#L64

JDBetteridge:
Closing since PR is merged

PR #2344, opened 2022-02-09
Define BLASLAPACK_LIB and BLASLAPACK_INCLUDE in complex mode

pbrubeck:
This addresses issue #2343
This is required to support the BLAS gemm kernels for p-MG interpolation.


JDBetteridge:
We should probably raise some sort of exception for users who are in complex mode trying to use Slate but aren't using p-MG.

JDBetteridge:
We should probably raise some sort of exception for users who are in complex mode trying to use Slate but aren't using p-MG.

pbrubeck:
I also think that we should not raise the exception related to not finding Eigen in complex mode.

dham (approved):

PR #2345, opened 2022-02-10
Prominently display petsc features at install

JDBetteridge:
When installing firedrake with `--honour-petsc-dir` list the configure options that we expect PETSc to have been built with.

Immediately after display the ones that were used to build the PETSc specified by `PETSC_DIR` and `PETSC_ARCH`.

Finally, print the external packages PETSc has detected during its own configure.


JDBetteridge:
Rebased to avoid previous failing tests. This is ready for merge

wence- (approved):
Thanks!

PR #2348, opened 2022-02-16
Add disk caching for interpolation kernels

connorjward:
Until now it has not been possible to cache interpolation kernels from TSFC because they stored references to the UFL coefficients (fixed in https://github.com/firedrakeproject/tsfc/pull/268).

### DIFF
--- a/firedrake/interpolation.py
+++ b/firedrake/interpolation.py
@@ -1,11 +1,15 @@
 import numpy
 from functools import partial, singledispatch
+import os
+import tempfile
 
 import FIAT
 import ufl
 from ufl.algorithms import extract_arguments, extract_coefficients
+from ufl.algorithms.signature import compute_expression_signature
 
 from pyop2 import op2
+from pyop2.caching import disk_cached
 
 from tsfc.finatinterface import create_element, as_fiat_cell
 from tsfc import compile_expression_dual_evaluation
@@ -270,6 +274,11 @@ def _interpolator(V, tensor, expr, subset, arguments, access):
         rt_var_name = 'rt_X'
         to_element = rebuild(to_element, expr, rt_var_name)
 
+    cell_set = target_mesh.cell_set
+    if subset is not None:
+        assert subset.superset == cell_set
+        cell_set = subset
+
     parameters = {}
     parameters['scalar_type'] = utils.ScalarType
 
@@ -279,10 +288,8 @@ def _interpolator(V, tensor, expr, subset, arguments, access):
     # FIXME: for the runtime unknown point set (for cross-mesh
     # interpolation) we have to pass the finat element we construct
     # here. Ideally we would only pass the UFL element through.
-    kernel = compile_expression_dual_evaluation(expr, to_element,
-                                                V.ufl_element(),
-                                                domain=source_mesh,
-                                                parameters=parameters)
+    kernel = compile_expression(cell_set.comm, expr, to_element, V.ufl_element(),
+                                domain=source_mesh, parameters=parameters)
     ast = kernel.ast
     oriented = kernel.oriented
     needs_cell_sizes = kernel.needs_cell_sizes
@@ -291,10 +298,6 @@ def _interpolator(V, tensor, expr, subset, arguments, access):
     name = kernel.name
     kernel = op2.Kernel(ast, name, requires_zeroed_output_arguments=True,
                         flop_count=kernel.flop_count)
-    cell_set = target_mesh.cell_set
-    if subset is not None:
-        assert subset.superset == cell_set
-        cell_set = subset
     parloop_args = [kernel, cell_set]
 
     coefficients = tsfc_interface.extract_numbered_coefficients(expr, coefficient_numbers)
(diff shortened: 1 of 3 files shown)


firedrake/interpolation.py
```diff
@@ -381,6 +387,27 @@ def _interpolator(V, tensor, expr, subset, arguments, access):
...
+    _expr_cachedir = os.path.join(tempfile.gettempdir(),
+                                  f"firedrake-tsfc-expression-kernel-cache-uid{os.getuid()}")
+
+
+def _compile_expression_key(expr, to_element, ufl_element, **kwargs):
+    try:
+        comm = kwargs["domain"].comm
+    except KeyError:
+        comm = expr.ufl_domains()[0].comm
+    key = hash(expr), to_element, ufl_element, utils.tuplify(kwargs.get("parameters"))
```
ReubenHill, line 402: Use of the `hash` function risks collisions (as I have found out). Dictionaries, for example, have some clever ways of avoiding collisions which do not just rely on `hash`. See for example [this stack overflow](https://stackoverflow.com/a/68048420) answer. I'm not suggesting adding `hashlib` as a dependency but just want to flag that this is often dangerous

ReubenHill: Newer versions of python have improved `hash` btw - possibly 3.7?

firedrake/interpolation.py
```diff
@@ -381,6 +387,27 @@ def _interpolator(V, tensor, expr, subset, arguments, access):
...
+    return comm, key
+
+
+compile_expression = disk_cached({}, _expr_cachedir, key=_compile_expression_key,
```
ReubenHill, line 406: Any reason you're not using `@disk_cached`? (I've never written a decorator)

connorjward: In Python a decorator is just a function that takes a function as its argument. It's hard to explain but what I have done here is equivalent to
```py
@disk_cached(...)
def compile_expression(*args, **kwargs):
    return compile_expression_dual_evaluation(*args, **kwargs)
```
just on fewer lines.

ReubenHill: The latter looks a bit nicer though IMO

firedrake/interpolation.py
connorjward: I may be wrong, but I think that `hash` is safe to use for UFL expressions because it is specially handled ([source](https://github.com/FEniCS/ufl/blob/main/ufl/core/expr.py#L109)). We also already use `hash` to cache UFL expressions [here](https://github.com/firedrakeproject/firedrake/blob/master/firedrake/assemble_expressions.py#L307).

firedrake/interpolation.py
connorjward: That's what I've ended up using anyway.

firedrake/interpolation.py
```diff
@@ -381,6 +392,27 @@ def _interpolator(V, tensor, expr, subset, arguments, access):
...
+    _expr_cachedir = os.path.join(tempfile.gettempdir(),
+                                  f"firedrake-tsfc-expression-kernel-cache-uid{os.getuid()}")
+
+
+def _compile_expression_key(comm, expr, to_element, ufl_element, domain, parameters):
+    """Generate a cache key suitable for :func:`tsfc.compile_expression_dual_evaluation`."""
+    # Since the caching is collective, this function must return a 2-tuple of
+    # the form (comm, key) where comm is the communicator the cache is collective over.
+    # FIXME FInAT elements are not safely hashable so we ignore them here
+    key = hash(expr), hash(ufl_element), hash(domain), utils.tuplify(parameters)
```
wence-, line 407: Is this the right key? For a disk key, I think you want a hash that is stable over coefficient and index renumbering (which would come from ufl.algorithms.compute_expression_signature or similar, rather than hash).

I am also not sure that the domain needs to be hashed. Remember, this needs to cover the information that is necessary for the kernel, not the global iteration?

connorjward: @wence- I've had a look at this and I'm not completely sure how to accomplish what you want. Would something like `slow_key` found in [assemble_expressions.py](https://github.com/firedrakeproject/firedrake/blob/master/firedrake/assemble_expressions.py#L310) work?

wence-: I think UFL has a `compute_expression_signature` somewhere. The main thing is to provide renumberings for coefficients and domains (`compute_form_signature` just gets the coefficients and domain numberings and then passes down to `compute_expression_signature` IIRC).

connorjward: This is now done. Sorry for taking so long over this.

firedrake/interpolation.py
```diff
@@ -279,27 +288,22 @@ def _interpolator(V, tensor, expr, subset, arguments, access):
...
-    if subset is not None:
-        assert subset.superset == cell_set
-        cell_set = subset
     parloop_args = [kernel, cell_set]
 
+    coefficients = tsfc_interface.extract_numbered_coefficients(expr, coefficient_numbers)
     if first_coeff_fake_coords:
         # Replace with real source mesh coordinates
-        coefficients[0] = source_mesh.coordinates
+        coefficients = [source_mesh.coordinates] + coefficients
```
ReubenHill, line 306: I think `first_coeff_fake_coords` needs to be renamed here (and perhaps elsewhere). It no longer tells you to replace the coordinates coefficient, instead you are adding a set of source mesh coordinates which aren't otherwise in the coefficient list.

ReubenHill: At the very least this comment
```
# Replace with real source mesh coordinates
```
needs to change since the addition of
```
coefficients = tsfc_interface.extract_numbered_coefficients(expr, coefficient_numbers)
```
means that's not what you are doing any more.

ReubenHill: `needs_external_coords` seems fine. Does it make logical sense to name it that in TSFC too, or has the API not been changed there sufficiently to warrant it?

connorjward: I think it is warranted since interpolation kernels no longer advertise a `coefficients` property and hence 'first coefficient' (from `first_coeff_fake_coords`) is misleading.

connorjward: This is now done.

dham (approved):

connorjward:
@sv2518 this needs to go in with https://github.com/firedrakeproject/tsfc/pull/268

sv2518 (approved):

PR #2352, opened 2022-02-17
Add pytest-timeout to CI

connorjward:
I've been trying to track down a test today that hangs and had difficulty because currently the CI continues to run until it is terminated after a number of hours. This means that I don't know which tests are the ones responsible for the hanging.

This change should timeout tests that are hanging so we can see which they are.

I chose 600s as the timeout limit because our longest test is currently around 400s so it should not be accidentally triggered.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -49,19 +49,19 @@ jobs:
         run: |
           . ../build/bin/activate
           python $(which firedrake-clean)
-          python -m pip install pytest-cov pytest-xdist
+          python -m pip install pytest-cov pytest-timeout pytest-xdist
           python -m pip list
       - name: Test Firedrake
         run: |
           . ../build/bin/activate
           echo OMP_NUM_THREADS is $OMP_NUM_THREADS
           echo OPENBLAS_NUM_THREADS is $OPENBLAS_NUM_THREADS
-          python -m pytest --durations=200 -n 12 --cov firedrake -v tests
+          python -m pytest --durations=200 -n 12 --cov firedrake --timeout=600 -v tests
       - name: Test pyadjoint
         if: ${{ matrix.scalar-type == 'real' }}
         run: |
           . ../build/bin/activate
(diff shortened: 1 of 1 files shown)


sv2518 (commented):
LGTM

connorjward:
> Excellent idea. Did you run the tests locally to find out which was the longest or is this based on the CI runs? I'm wondering if someone might get a test fail because of a timeout when really they just have a system that's slow

The timings are based on the CI runs. Also, this change won't affect users because they would be running `pytest` directly themselves rather than via the `build.yaml` CI file.

ReubenHill:
> > Excellent idea. Did you run the tests locally to find out which was the longest or is this based on the CI runs? I'm wondering if someone might get a test fail because of a timeout when really they just have a system that's slow
> 
> The timings are based on the CI runs. Also, this change won't affect users because they would be running `pytest` directly themselves rather than via the `build.yaml` CI file.

Ah so this only helps you isolate such failing tests when you run on CI? Also do we think a margin of error in test running time of 50 % is enough?

connorjward:
> > > Excellent idea. Did you run the tests locally to find out which was the longest or is this based on the CI runs? I'm wondering if someone might get a test fail because of a timeout when really they just have a system that's slow
> 
> > 
> 
> > The timings are based on the CI runs. Also, this change won't affect users because they would be running `pytest` directly themselves rather than via the `build.yaml` CI file.
> 
> 
> 
> Ah so this only helps you isolate such failing tests when you run on CI? Also do we think a margin of error in test running time of 50 % is enough?

Yes that's right. I use `pytest-timeout` all the time in my own work though so there's nothing stopping someone from doing the same locally.

The 50% margin was fairly arbitrary. I'm happy to change it if people prefer something different.

JDBetteridge (approved):
https://github.com/firedrakeproject/firedrake/pull/2356/commits/e5e0ef23ed5b369848031134a54a3d9bfc96559e

Is evidence that `pytest-timeout` is terminating hung tasks correctly

PR #2370, opened 2022-03-04
Test Slate tensors: don't assemble diagonal of tensor with interior facets.

sv2518:
I merged this PR earlier https://github.com/firedrakeproject/tsfc/pull/269 which caused some of the Slate tests fail. This is a fix.

### DIFF
--- a/tests/slate/test_assemble_tensors.py
+++ b/tests/slate/test_assemble_tensors.py
@@ -82,19 +82,16 @@ def mass(function_space):
 
 
 @pytest.fixture
-def matrix_mixed_facet():
+def matrix_mixed_nofacet():
     mesh = UnitSquareMesh(2, 2)
     U = FunctionSpace(mesh, "RT", 1)
     V = FunctionSpace(mesh, "DG", 0)
     T = FunctionSpace(mesh, "HDiv Trace", 0)
-    n = FacetNormal(mesh)
     W = U * V * T
     u, p, lambdar = TrialFunctions(W)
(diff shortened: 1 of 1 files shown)


tests/slate/test_assemble_tensors.py
```diff
@@ -306,16 +303,16 @@ def test_diagonal(mass, matrix_mixed_facet):
...
     for r, d in zip(res2, np.diag(ref2)):
         assert np.allclose(r, d, rtol=1e-14)
 
     # test matrix built from diagonal
     # for a Slate expression on a non mass matrix
-    A = Tensor(matrix_mixed_facet)
+    A = Tensor(matrix_mixed_nofacet)
     res3 = assemble(DiagonalTensor(A+A)).M.values
-    ref3 = np.concatenate(assemble(matrix_mixed_facet+matrix_mixed_facet, diagonal=True).dat.data)
+    ref3 = np.concatenate(assemble(matrix_mixed_nofacet+matrix_mixed_nofacet, diagonal=True).dat.data)
```
wence-, line 315: Is it the case that slate does the wrong thing for `DiagonalTensor(something_with_interior_facet_terms)`? Or is it that the test no longer passes? I think if it is the former should the slate compiler also complain one asks for `DiagonalTensor` of interior facet?

sv2518: TSFC throws the `NonImplementedError`

wence- (approved):

PR #2378, opened 2022-03-09
Fix interpolation test for Nedelec, BDM and RT elements for variant=integral

FabianL1908:
We changed the default quadrature degree of the Nedelec, BDM and RT elements for variant=integral from 5k to k in fiat. Hence, if you want to still want to still be div/curl preserving in your interpolation, the quadrature degree should be set to something high, for this test we choose 5k.

### DIFF
--- a/tests/regression/test_interpolation_nodes.py
+++ b/tests/regression/test_interpolation_nodes.py
@@ -34,7 +34,7 @@ def degree(request):
                 ids=lambda x: "%s" % x)
 def V(request, mesh, degree):
     space = request.param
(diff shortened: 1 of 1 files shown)


wence-:
So did we need the expressions to be non-polynomial to exercise this correctly? Note we're still skipping some tests here

FabianL1908:
Yes, the point of the non-polynomial expression is that the interpolation is not exact for the default quadrature degree of k. This test ensures that if we choose a high quadrature degree, then the interpolation of divergence free function is still divergence free.
And yes, we still skip some tests. The variant changes do not improve the performance of the skipped tests.

wence- (approved):
Thanks

PR #2384, opened 2022-03-17
Firedrake changes for updating PETSc

ksagiyam:
Merge PETSc upstream.

The following commit in the upstream caused Firedrake tests that use gamg to fail with `DIVERGED_LINEAR_SOLVE` or similar:

"switch default GAMG smoothing to PCJACOBI" (`5f7df010c372f2aba7032437adc22ab0932ed26e`)

So we need to add options to recover old behaviour.


dham (approved):
This looks like the Right Thing. I guess we can merge this before we merge PETSc, right?

ksagiyam:
Yes, I think so.

ksagiyam:
I need to drop petsc branch ...

ksagiyam:
Dropped petsc branch; can be merged independently. Thanks.

PR #2385, opened 2022-03-19
Compatibility with macOS 12.3

bourdin:
- Added the option to use external MI (--honour-mpi-dir_) since I can't get petsc to compile mpich at the moment and brew mpich works file
   - Have hwloc installed by homebrew for the same reason

At the moment, installing on macOS 12.3 on a M1 system only works when --honour-mpi-dir is passed (and MPI_DIR is set to $HOMEBREW_PREFIX, for instance). Doing so also speeds up building tremendously.

### DIFF
--- a/scripts/firedrake-install
+++ b/scripts/firedrake-install
@@ -66,7 +66,7 @@ class FiredrakeConfiguration(dict):
 
     _persistent_options = ["package_manager",
                            "minimal_petsc", "mpicc", "mpicxx", "mpif90", "mpiexec", "disable_ssh",
-                           "honour_petsc_dir", "with_parmetis",
+                           "honour_mpi_dir", "honour_petsc_dir", "with_parmetis",
                            "slepc", "packages", "honour_pythonpath",
                            "opencascade", "tinyasm",
                            "petsc_int_type", "cache_dir", "complex", "remove_build_files", "with_blas"]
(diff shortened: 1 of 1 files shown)


wence-:
Does it not work to just set --mpicc etc to the mpi compilers provided by homebrew? This has always worked for me on pre-M1 systems.

It also looks like there are a bunch of changes that are unrelated to this new option, is that deliberate?

bourdin:
I had troubles with some packages using Monica as the c compiler. I can try again tu use —with-mpicc
The other changes are deliberate: for some reason, setting MACOSX_SOMETHING leads to a working build but examples fail to compile. 
hwloc does not build with the version of PETSc that you are using
H5py seemed to always try to use mpicc built by PETSc

scripts/firedrake-install
```diff
@@ -1048,15 +1036,17 @@ def build_and_install_petsc():
         petsc_dir, petsc_arch = get_petsc_dir()
         check_call(["rm", "-rf", os.path.join(petsc_dir, petsc_arch)])
         petsc_options = get_petsc_options()
+        # As of 2022-03-17, it looks like setting MACOSX_DEPLOYMENT_TARGET on macOS 12.3 system actually breaks petsc configuration
```
dham, line 1039: This has now been fixed in a hopefully better way in #2408. I claim we don't want to delete this code.

scripts/firedrake-install
```diff
@@ -1070,29 +1060,6 @@ def build_and_install_petsc():
                 check_call(["rm", "-f", deletefile])
 
 
-def build_and_install_petsc4py():
```
dham, line 1073: It's not clear to me why we want to make this change.

scripts/firedrake-install
```diff
@@ -1424,30 +1392,6 @@ log.debug("*** Current environment (output of 'env') ***")
 log.debug(check_output(["env"]))
 log.debug("\n\n")
 
-if mode == "install" and args.honour_petsc_dir:
```
dham, line 1427: Why is this being deleted?

scripts/firedrake-install
```diff
@@ -669,32 +683,6 @@ def get_petsc_options(minimal=False):
     return list(petsc_options)
 
 
-def get_petsc_packages(minimal=False):
```
dham, line 672: Why is this being deleted?

dham (changes requested):
I have no objection to allowing --honour-mpi-dir, however it looks like there is also a lot else going on here and I don't understand why the additional changes are desirable.

dham:
Seems no longer required.

PR #2386, opened 2022-03-21
Do not use requirements.txt for icepack install

connorjward:
Our CI is currently failing to build because of the `roltrilinos` dependency for icepack [here](https://github.com/icepack/icepack/blob/master/requirements.txt). This is only a problem because our way of handling dependencies is to parse the `requirements.txt` file and pip install the packages, ignoring the fact that this package is optional in icepack.

The tiny change I've made here is to use icepack's `setup.py` file instead of `requirements.txt`. Then the dependencies are resolved using pip and work as expected.

### DIFF
--- a/scripts/firedrake-install
+++ b/scripts/firedrake-install
@@ -900,8 +900,8 @@ def pip_requirements(package, parallel_compiler_env):
                         log.debug("Skipping failing install of optional pyadjoint dependency tensorflow")
                     else:
                         raise e
-        elif package == "loopy":
-            # loopy dependencies are installed in setup.py
+        elif package in {"loopy", "icepack"}:
+            # dependencies are installed in setup.py
             return
         else:
             reqs = get_requirements("%s/requirements.txt" % package)


danshapero (approved):

danshapero:
This change looks fine to me, but I think that roltrilinos is only failing because it needs patchelf and the build environment doesn't have that installed. Other downstream applications (e.g. Fireshape) depend on ROL for optimization. It might be good to have some testing of the firedrake-adjoint -> ROL interface on principle. Pyadjoint [doesn't seem to test](https://app.circleci.com/pipelines/github/dolfin-adjoint/pyadjoint/136/workflows/7baa7a33-8abd-4b4e-8d2e-c1358d3fc7c5/jobs/282?invite=true#step-105-51) this.

connorjward:
> This change looks fine to me, but I think that roltrilinos is only failing because it needs patchelf and the build environment doesn't have that installed. Other downstream applications (e.g. Fireshape) depend on ROL for optimization. It might be good to have some testing of the firedrake-adjoint -> ROL interface on principle. Pyadjoint [doesn't seem to test](https://app.circleci.com/pipelines/github/dolfin-adjoint/pyadjoint/136/workflows/7baa7a33-8abd-4b4e-8d2e-c1358d3fc7c5/jobs/282?invite=true#step-105-51) this.

That also seems like a perfectly acceptable way to fix this issue.

It is also worth noting that CI is still failing for icepack here because it cannot now find `rasterio` so your solution seems to be the better one.

danshapero:
Well I had an explanation about roltrilinos, but I don't know why it couldn't find rasterio. That's clearly a required dependency in setup.py/install_requires, so even if you've changed icepack to installing dependencies from setup.py it should be there. I'll try on my own, but maybe it's worth discussing more at Thursday's meeting.

connorjward:
@wence- or @dham could you maybe merge this? It's not a perfect fix but without it CI doesn't even build which has caused a number of people issues.

PR #2387, opened 2022-03-23
Ap/diskmesh degree

APaganini:
Allow to specify the `degree` of the utility `diskmesh`.
If `degree>1`, the higher order mesh is created by
interpolating the boundary.

### DIFF
--- a/firedrake/utility_meshes.py
+++ b/firedrake/utility_meshes.py
@@ -781,7 +781,6 @@ def UnitDiskMesh(refinement_level=0, degree=1, reorder=None,
         from firedrake.bcs import DirichletBC
         DirichletBC(V, bdry_map, "on_boundary").apply(p_map)
         m = mesh.Mesh(p_map)
-        #m._radius = radius
     return m
 
 
--- a/firedrake/utility_meshes.py
+++ b/firedrake/utility_meshes.py
@@ -714,13 +714,21 @@ def CircleManifoldMesh(ncells, radius=1, degree=1, distribution_parameters=None,
 
 
 @PETSc.Log.EventDecorator()
-def UnitDiskMesh(refinement_level=0, reorder=None, distribution_parameters=None, comm=COMM_WORLD):
+def UnitDiskMesh(refinement_level=0, degree=1, reorder=None,
+                 distribution_parameters=None, comm=COMM_WORLD):
     """Generate a mesh of the unit disk in 2D
 
     :kwarg refinement_level: optional number of refinements (0 is a diamond)
+    :kwarg degree: polynomial degree of coordinate space (defaults
+        to 1: flat triangles)
     :kwarg reorder: (optional), should the mesh be reordered?
     :kwarg comm: Optional communicator to build the mesh on (defaults to
         COMM_WORLD)."""
+    if refinement_level < 0 or refinement_level % 1:
+        raise RuntimeError("Number of refinements must be a non-negative integer")
+
+    if degree < 1:
+        raise ValueError("Mesh coordinate degree must be at least 1")
     vertices = np.array([[0, 0],
                          [1, 0],
                          [1, 1],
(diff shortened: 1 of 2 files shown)


tests/regression/test_poisson_disk.py
```diff
@@ -0,0 +1,49 @@
...
+    return [max(h.dat.data_ro), err]
+
+def conv_rate(degree):
+    err = []
+    h = []
+    for ref_level in range(0,4):
+        h_, err_ = poisson_solver(ref_level, degree)
+        h.append(h_)
+        err.append(err_)
+    #import ipdb; ipdb.set_trace()
```
danshapero: Guessing this is a holdover from debugging?

danshapero (changes requested):
Except for one tiny thing this all looks good t me, thanks Alberto!

tests/regression/test_poisson_disk.py
danshapero: Nevermind, I see that you've fixed it.

danshapero (approved):

firedrake/utility_meshes.py
```diff
@@ -762,6 +770,17 @@ def UnitDiskMesh(refinement_level=0, reorder=None, distribution_parameters=None,
...
+    # if degree > 1, create parametric mesh
+    if degree > 1:
+        V = functionspace.VectorFunctionSpace(m, "CG", degree)
+        p_map = function.Function(V)
+        p_map.interpolate(ufl.SpatialCoordinate(m))
+        x, y = ufl.SpatialCoordinate(m)
+        bdry_map = ufl.as_vector([x / (x**2 + y**2)**0.5, y / (x**2 + y**2)**0.5])
+        from firedrake.bcs import DirichletBC
+        DirichletBC(V, bdry_map, "on_boundary").apply(p_map)
+        m = mesh.Mesh(p_map)
```
wence-, line 783: This is liable to create bad meshes I think (ping @pefarrell and @ScottMacLachlan) depending on resolution. I think they just solve elasticity to move the internal mesh nodes after pushing to the boundary, should something like that be done here?

wence- (changes requested):
I am possibly worried about this making meshes with bad elements near the boundary. Need I be?

firedrake/utility_meshes.py
APaganini: I don't think this is necessary in this instance, but I'm curious to read what @pefarrell  and @ScottMacLachlan viewpoints.

ksagiyam: Can we instead increase order right after `plex.refine()`s and use the same logic (ll. 765-770) for higher-order meshes to move nodes onto the unit disk?
```
   ...
    for i in range(refinement_level):
        plex = plex.refine()
    if degree > 1:
        m = mesh.Mesh(plex, dim=2, reorder=reorder, distribution_parameters=distribution_parameters)
        V = functionspace.VectorFunctionSpace(m, "CG", degree)
        f = function.Function(V)
        f.interpolate(ufl.SpatialCoordinate(m))
        coords = f.dat.data
    else:
        coords = plex.getCoordinatesLocal().array.reshape(-1, 2)
    for x in coords:
        norm = np.sqrt(np.dot(x, x))
        # I think this is just checking that norm != 0.
        if norm > 1. / (1 << (refinement_level + 1)) / degree / 2:
            t = np.max(np.abs(x)) / norm
            x[:] *= t
    if degree > 1:
        return mesh.Mesh(f)
    else:
        return mesh.Mesh(plex, dim=2, reorder=reorder, distribution_parameters=distribution_parameters)
```

ksagiyam: Yes, I does, and it also moves internal DoFs using the same rule. I'm not sure how good this rule itself is, but I thought it's simpler to rely on a single rule (instead of separately treating boundary DoFs using `DirichletBC`).

ksagiyam: I'm actually not sure about the quality of the mesh I proposed. There might be some unnecessary distortions inside the disk. Maybe it was not a good idea.

ScottMacLachlan: Sorry for being slow to weigh in on this!

I don't think the original code would suffer from the mesh quality problems that we encounter when doing similar things in `OpenCascadeMeshHierarchy`.  In that setting, we get poor quality near the boundary because we refine a mesh some number of times without moving the intermediate nodes, then project onto the boundary all-at-once on the finest level (fix in progress...).  Here, the first-order mesh already has all of its nodes on the boundary, so the reassignment should just be moving the control points on the edges out to the boundary, which seems reasonable to me.

ScottMacLachlan: @pefarrell and I have been messing with a `UnitDiskMeshHierarchy` to generate comparable hierarchies of the disk without using OCCMH.  In that code, we use a solution similar to what's proposed by @ksagiyam .  We do a hierarchy on the [-1,1]^2 domain, create high-order coordinates, rebuild the high-order hierarchy, then transform each level by the existing transformation.  It isn't totally clear to me if those steps commute or not...

Issue #2396, opened 2022-03-29
IndexError when using $\R$ space with more then one other space.


lrtfm:
The following code will fail in the latest version of Firedrake.
```
$ cat r_fail.py
#!/usr/bin/env python
# coding: utf-8

# A failing example with $R$

from firedrake import *
from firedrake.petsc import PETSc
import numpy as np

mesh = RectangleMesh(10, 10, 1, 1)

U = VectorFunctionSpace(mesh, 'CG', 2)
Q = FunctionSpace(mesh, 'CG', 1)
R = FunctionSpace(mesh, 'R', 0)
V = MixedFunctionSpace([U, Q, R])

eps = 1e-8

w_, p_, l_water_ = TrialFunctions(V)
v_, q_, m_water_ = TestFunctions(V)

f = as_vector([0, -9.8])

A = 2*inner(sym(grad(w_)), sym(grad(v_)))*dx \
    - inner(p_, div(v_))*dx \
    + inner(div(w_), q_)*dx \
    + eps*inner(p_, q_)*dx
A += dot(w_, f)*m_water_*dx - l_water_*v_[1]*dx
L = dot(f, v_)*dx(domain=mesh)

bc = DirichletBC(V.sub(0), 0., 'on_boundary')

sol = Function(V)
# If we solve homogeneous problem, i.e. L=0, everything is ok
# stokes_problem = LinearVariationalProblem(A, 0, sol, bcs=bc)
stokes_problem = LinearVariationalProblem(A, L, sol, bcs=bc)
stokes_solver = LinearVariationalSolver(stokes_problem, options_prefix='stokes')
stokes_solver.solve()

```
The output information:
```
Traceback (most recent call last):
  File "r_fail.py", line 38, in <module>
    stokes_solver.solve()
  File "PETSc/Log.pyx", line 115, in petsc4py.PETSc.Log.EventDecorator.decorator.wrapped_func
  File "PETSc/Log.pyx", line 116, in petsc4py.PETSc.Log.EventDecorator.decorator.wrapped_func
  File "/home/zzyang/workspace/firedrake-gcc/firedrake-mini-petsc/src/firedrake/firedrake/adjoint/variational_solver.py", line 90, in wrapper
    out = solve(self, **kwargs)
  File "/home/zzyang/workspace/firedrake-gcc/firedrake-mini-petsc/src/firedrake/firedrake/variational_solver.py", line 278, in solve
    self.snes.solve(None, work)
[...]

PR #2398, opened 2022-03-29
Ksagiyam/fix things around bc

ksagiyam:
This PR is to fix two things:

1.
```
mesh = UnitSquareMesh(1, 1)
g = as_vector([0., 0.])
bc = DirichletBC(V, g, "on_boundary")

print(repr(g))  # Zero((2,), (), ())
print(repr(bc.function_arg))  # currently prints Zero((), (), ()) <- fix
```

2.
```
mesh = UnitSquareMesh(1, 1)
W0 = V * V
W1 = V * V * V * V

print(W0 == W1)  # currently prints True <- fix
```

### DIFF
--- a/tests/regression/test_matrix_free.py
+++ b/tests/regression/test_matrix_free.py
@@ -1,6 +1,7 @@
 from firedrake import *
 from firedrake.petsc import PETSc
 from firedrake.utils import ScalarType
+from firedrake.solving_utils import DEFAULT_KSP_PARAMETERS
 import pytest
 import numpy as np
 from mpi4py import MPI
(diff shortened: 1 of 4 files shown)


wence-:
Can you add a test that exercises the failing example from #2396

ksagiyam:
Good idea. Added!

tests/regression/test_matrix_free.py
```diff
@@ -308,3 +308,34 @@ def test_duplicate(a, bcs):
         B_petsc.mult(x, y)
     # Check if original rhs is equal to BA^-1 (rhs)
     assert np.allclose(rhs.vector().array(), solution2.vector().array())
+
+
+def test_matrix_free_fieldsplit_with_real():
```
dham, line 313: Is this really matrix free? I can't see anything in the solver options that would make it so.

dham (changes requested):

tests/regression/test_matrix_free.py
ksagiyam: Turned out that, when we have `Real`, `solving_utils.set_defaults()` puts `matfree` here: https://github.com/firedrakeproject/firedrake/blob/06125881584cc898fed1f9b9c37a25bbe3bb55c9/firedrake/solving_utils.py#L95

connorjward: Perhaps do an assertion for `stokes_solver.parameters["mat_type"]` to make this explicit?

ksagiyam: Or can I just add comments, as `set_defaults()` also sets many related `fieldsplit` options?

ksagiyam: Actually, I agree. It's probably better to spell out options we need here; updated.

dham (approved):

PR #2400, opened 2022-03-30
Make importing applications a non-strict xfail

### LINKED ISSUE #2390: testing downstream packages
The Firedrake test suite checks whether it can import downstream packages like Icepack, Thetis, Gusto, etc. This means that people like me can inadvertently break the Firedrake build by, for example, goofing up setup.py in our own projects (see #2386).

**Has testing import of downstream packages caught many potential bugs in development branches?** I have a hunch that it hasn't. If not, I don't think it's worth continuing to do this. I've broken the build a bunch of times and that inconveniences several people, so I'd rather take care of any problems that get introduced from upstream on my end.

Just want to also check with @tkarna, @colinjcotter, @jshipton since Thetis and Gusto are included in the list of downstream packages that get tested. If I'm way off base and other people would prefer to keep these tests, then I'll be more careful or change some of the testing of Icepack.

If github actions had an [allow failure](https://github.com/actions/toolkit/issues/399) option then that would be a nicer solution, but they don't seem to have implemented it yet.

danshapero:
Resolves #2390. Importing external applications (Icepack, Thetis, Gusto, Irksome, Femlium) can fail without breaking the build.

### DIFF
--- a/tests/regression/test_import_applications.py
+++ b/tests/regression/test_import_applications.py
@@ -10,6 +10,7 @@ def app(request):
 
 
 @pytest.mark.skipif("FIREDRAKE_CI_TESTS" not in os.environ, reason="Not running in CI")
+@pytest.mark.xfail(strict=False)
 def test_import_app(app):
     # Have to run this in a subprocess in case the import pollutes the
     # test environment, e.g. by modifying firedrake parameters.


wence- (approved):
OK, I guess the way things will work is downstream will complain if we break things. This is not quite ideal I think, but the current setup is also not ideal.

PR #2402, opened 2022-04-01
Use SubMatrices to compute eigvals in demo qgbasinmodes

APaganini:
To compute the eigenvalues more robustly, create submatrices by removing rows and columns associated with degrees of freedom. This trick was suggested (on a different example) by @jandrej , @wence- , @francesco-ballarin and Matt Knepley, among others.

**Disclaimer** I haven't run the demo to check that it runs because I don't know how to convert an `*.py.rst` into a `*.py` I can run.

### DIFF
--- a/demos/eigenvalues_QG_basinmodes/qgbasinmodes.py.rst
+++ b/demos/eigenvalues_QG_basinmodes/qgbasinmodes.py.rst
@@ -167,7 +167,7 @@ submatrices that omit these rows and the correspoing columns. ::
 
    bcnodes = Vcg.dof_dset.lgmap.apply(bc.nodes).astype(pyop2.datatypes.IntType)
    ISbd = PETSc.IS().createGeneral(bcnodes, comm=petsc_a.comm)
-   (s, e), _ = petsc_a.getSizes()
+   s, e = petsc_a.getOwnershipRange()#(s, e), _ = petsc_a.getSizes()
    ISin = ISbd.complement(s, e)
    A = petsc_a.createSubMatrix(ISin, ISin)
    M = petsc_m.createSubMatrix(ISin, ISin)
@@ -219,7 +219,7 @@ converge any eigenvalues at all. ::
 If we did, we go ahead and extract them from the SLEPc eigenvalue
 solver::
 
-   vr, vi = petsc_a.getVecs()
+   vr, vi = A.getVecs()
 
    lam = es.getEigenpair(0, vr, vi)
 
--- a/demos/eigenvalues_QG_basinmodes/qgbasinmodes.py.rst
+++ b/demos/eigenvalues_QG_basinmodes/qgbasinmodes.py.rst
@@ -109,6 +109,7 @@ Firedrake. We import the Firedrake, PETSc, and SLEPc libraries. ::
 
    from firedrake import *
    from firedrake.petsc import PETSc
+   import pyop2
    try:
        from slepc4py import SLEPc
    except ImportError:
--- a/demos/eigenvalues_QG_basinmodes/qgbasinmodes.py.rst
+++ b/demos/eigenvalues_QG_basinmodes/qgbasinmodes.py.rst
@@ -165,7 +165,7 @@ submatrices that omit these rows and the correspoing columns. ::
    petsc_a = assemble(a).M.handle
    petsc_m = assemble(m, bcs=bc).M.handle
 
-   bcnodes = Vcg.dof_dset.lgmap.apply(bc.nodes).astype(pyop2.datatypes.IntTpye)
+   bcnodes = Vcg.dof_dset.lgmap.apply(bc.nodes).astype(pyop2.datatypes.IntType)
    ISbd = PETSc.IS().createGeneral(bcnodes, comm=petsc_a.comm)
    (s, e), _ = petsc_a.getSizes()
    ISin= ISbd.complement(s, e)
--- a/demos/eigenvalues_QG_basinmodes/qgbasinmodes.py.rst
+++ b/demos/eigenvalues_QG_basinmodes/qgbasinmodes.py.rst
@@ -155,9 +155,9 @@ To build the weak formulation of our equation we need to build two PETSc
 matrices in the form of a generalized eigenvalue problem,
 :math:`A\psi = \lambda M\psi`. We impose the boundary conditions on the
 mass matrix :math:`M`, since that is where we used integration by parts.
-For a more stable computation of the eigenvalues, we also create
-submatrices by removing rows and columns associated with the boundary
-conditions. ::
+To avoid spurious eigenvalues associated with the rows of :math:`M`
+that are modified by the Dirichlet boundary conditions, we create
+submatrices that omit these rows and the correspoing columns. ::
 
    a =  beta*phi*psi.dx(0)*dx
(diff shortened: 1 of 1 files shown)


jandrej:
Sweet. I like the `complement` approach of getting the dofs.

APaganini:
> Sweet. I like the `complement` approach of getting the dofs.

It was Lawrence's idea :)

wence-:
> To compute the eigenvalues more robustly, create submatrices by removing rows and columns associated with degrees of freedom. This trick was suggested (on a different example) by @jandrej , @wence- , @francesco-ballarin and Matt Knepley, among others.
> 
> **Disclaimer** I haven't run the demo to check that it runs because I don't know how to convert an `*.py.rst` into a `*.py` I can run.

Run `make demos` in the `demos` directory

demos/eigenvalues_QG_basinmodes/qgbasinmodes.py.rst
```diff
@@ -154,13 +154,22 @@ We define the Test Function :math:`\phi` and the Trial Function
 To build the weak formulation of our equation we need to build two PETSc
 matrices in the form of a generalized eigenvalue problem,
 :math:`A\psi = \lambda M\psi`. We impose the boundary conditions on the
-mass matrix :math:`M`, since that is where we used integration by parts. ::
+mass matrix :math:`M`, since that is where we used integration by parts.
+For a more stable computation of the eigenvalues, we also create
```
wence-, line 158: "stable computation" reads wrong to me. I think "To avoid spurious eigenvalues associated with the Dirichlet rows of the operator, we extract submatrices to remove them" or something like that?

demos/eigenvalues_QG_basinmodes/qgbasinmodes.py.rst
```diff
@@ -214,7 +223,7 @@ solver::
 
 and we gather the final eigenfunctions ::
 
-   eigenmodes_real.vector()[:], eigenmodes_imag.vector()[:] = vr, vi
+   eigenmodes_real.vector()[ISin.array], eigenmodes_imag.vector()[ISin.array] = vr, vi
```
wence-, line 226: You should also insert the Dirichlet conditions into these eigenvectors (`bc.apply`)

demos/eigenvalues_QG_basinmodes/qgbasinmodes.py.rst
```diff
@@ -154,13 +154,22 @@ We define the Test Function :math:`\phi` and the Trial Function
...
    a =  beta*phi*psi.dx(0)*dx
    m = -inner(grad(psi), grad(phi))*dx - F*psi*phi*dx
    petsc_a = assemble(a).M.handle
    petsc_m = assemble(m, bcs=bc).M.handle
 
+   bcnodes = Vcg.dof_dset.lgmap.apply(bc.nodes).astype(np.int32)
+   ISbd = PETSc.IS().createGeneral(bcnodes, comm=PETSc.COMM_SELF)
+   ISin= ISbd.complement(0, Vcg.dim())
+   A = petsc_a.createSubMatrix(ISin, ISin)
+   M = petsc_m.createSubMatrix(ISin, ISin)
```
wence-, line 171: I think that this needs a bit more explanation. Moreover there are a few nitpicks:

1. `np.int32` should be `pyop2.datatypes.IntType` (Aside, `IntType` and `ScalarType` should be exposed on import of Firedrake I think)?
2. naming convention is rather off from normal Python.
3. I think that the communicator for the IS should be `petsc_a.comm` not `COMM_SELF`.
4. I think the call to complement is wrong if running in parallel (see [ISComplement](https://petsc.org/release/docs/manualpages/IS/ISComplement.html)). I think you want to say something like `(s, e), _ = petsc_a.getSizes()` and use `iset.complement(s, e)`

wence- (changes requested):
I think this needs a bit of work to be correct in parallel.

demos/eigenvalues_QG_basinmodes/qgbasinmodes.py.rst
APaganini: Ok, I'll change it.

demos/eigenvalues_QG_basinmodes/qgbasinmodes.py.rst
APaganini: Is this necessary, given that `bc = DirichletBC(Vcg, 0.0, "on_boundary")`?

demos/eigenvalues_QG_basinmodes/qgbasinmodes.py.rst
APaganini: 1. changed to `pyop2.datatypes.IntType`. I don't understand what exposing means.
2. suggestions are welcome
3. changed to `petsc_a.comm`
4. changed as suggested

APaganini:
Applied most of @wence- suggestions. Now it doesn't run because I don't know how to properly implement `ISin = ISbd.complement(s, e)`

wence-:
I probably got the syntax wrong for getting the size of the matrix

APaganini:
I corrected (`s, e), _ = petsc_a.getSizes()` into `petsc_a.getOwnershipRange()`. Now it runs. The result is slightly different from the original version. More specifially, the original output is
> Leading eigenvalue is: (1.560678669632054e-13+0.10979244518390797j)
whereas after the modification from this pull request the output becomes
> Leading eigenvalue is: (1.1395615354126143e-14+0.10979281297538487j)
The real part is different (but still very close to 0), whereas the imaginary parts differ by `-3.67791477e-7`. I assume this is fine, but my opinion on this is uninformed.

demos/eigenvalues_QG_basinmodes/qgbasinmodes.py.rst
```diff
@@ -154,13 +155,23 @@ We define the Test Function :math:`\phi` and the Trial Function
...
 
    a =  beta*phi*psi.dx(0)*dx
    m = -inner(grad(psi), grad(phi))*dx - F*psi*phi*dx
    petsc_a = assemble(a).M.handle
    petsc_m = assemble(m, bcs=bc).M.handle
 
+   bcnodes = Vcg.dof_dset.lgmap.apply(bc.nodes).astype(pyop2.datatypes.IntType)
+   ISbd = PETSc.IS().createGeneral(bcnodes, comm=petsc_a.comm)
+   s, e = petsc_a.getOwnershipRange()
+   ISin = ISbd.complement(s, e)
```
lrtfm, line 171: In parallel, `ISbd.complement(s, e)` may rise exception:
```
File PETSc/IS.pyx:234, in petsc4py.PETSc.IS.complement()

Error: error code 62
[0] ISComplement() at /home/zzyang/workspace/firedrake-gcc/firedrake-mini-petsc-complex/src/petsc/src/vec/is/is/utils/iscoloring.c:815
[0] Invalid argument
[0] Index set must be sorted
[1:execute]
```
Maybe need add
```
s, e = petsc_a.getOwnershipRange()
bcnodes = bcnodes[np.logical_and(bcnodes < e, bcnodes >= s)]
bcnodes.sort()
```
to delete the ghost nodes and sort the array before creating the `ISbd`.

dham:
If this is to be done it should be done in LinearEigenSolver. The right thing to do, though, is RestrictedFunctionSpace, which is hopefully coming.

PR #2407, opened 2022-04-08
Fixes for M1

dham:
It turns out that something on M1 Mac causes us to have to pass the location of the gcc libraries to clang every time we compile something. This PR causes this to happen.

### DIFF
--- a/firedrake/__init__.py
+++ b/firedrake/__init__.py
@@ -13,6 +13,11 @@ elif "PETSC_DIR" not in os.environ and config["options"]["honour_petsc_dir"]:
 elif not config["options"]["honour_petsc_dir"]:  # Using our own PETSC.
     os.environ["PETSC_DIR"] = os.path.join(sys.prefix, "src", "petsc")
     os.environ["PETSC_ARCH"] = "default"
+if "PYOP2_LDFLAGS" in config["environment"]:
+    os.environ["PYOP2_LDFLAGS"] = " ".join(
+        (os.environ.get("PYOP2_LDFLAGS", default=""),
+         config["environment"]["PYOP2_LDFLAGS"])
+    )
 del sys, config
 
 # Ensure petsc is initialised by us before anything else gets in there.
--- a/scripts/firedrake-install
+++ b/scripts/firedrake-install
@@ -563,6 +563,13 @@ If you really want to use your own Python packages, please run again with the
 """)
 
 
+def brew_gcc_libdir():
+    brew_gcc_prefix = check_output(["brew", "--prefix", "gcc"])[:-1]
+    brew_gcc_info = json.loads(check_output(["brew", "info", "--json", "gcc"])[:-1])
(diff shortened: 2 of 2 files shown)


firedrake/__init__.py
```diff
@@ -13,6 +13,11 @@
 elif not config["options"]["honour_petsc_dir"]:  # Using our own PETSC.
     os.environ["PETSC_DIR"] = os.path.join(sys.prefix, "src", "petsc")
     os.environ["PETSC_ARCH"] = "default"
+if "PYOP2_LDFLAGS" in config["environment"]:
+    os.environ["PYOP2_LDFLAGS"] = " ".join(
+        (os.environ.get("PYOP2_LDFLAGS", default=""),
+         config["environment"]["PYOP2_LDFLAGS"])
+    )
```
connorjward, line 20: We don't use `config["environment"]` anywhere else except trivially in `firedrake-status` so this feels quite clunky and invasive for a bug fix.

Why can't we add something like`bugfix_ldflags` (like [`bugfix_cflags`](https://github.com/OP2/PyOP2/blob/master/pyop2/compilation.py#L498)) to PyOP2?

dham: The issue with doing this in PyOP2 is that we end up duplicating logic for the bugfix, because this fix is also needed in other parts of the Firedrake install. It therefore feels less bad to fix things in Firedrake and pass them through to PyOP2 from that single fix.

connorjward: I guess. Though that does mean that a standalone PyOP2 (which I know doesn't actually exist) on M1 is broken.

wence-:
I do not need this to run on M1. My build approach (separate PETSc, using homebrew mpich):

```python
#!/usr/bin/env python3
if __name__ == '__main__':
  import sys
  import os
  sys.path.insert(0, os.path.abspath('config'))
  import configure
  configure_options = [
    '--download-hcephes',
    '--download-hdf5',
    '--download-hwloc',
    '--download-hypre',
    '--download-metis',
    '--download-ml',
    '--download-mumps',
    '--download-netcdf',
    '--download-chaco',
    '--download-parmetis',
    '--download-ptscotch',
    '--download-scalapack',
    '--download-suitesparse',
    '--download-superlu_dist',
    '--download-triangle',
    '--with-c2html=0',
    '--with-cxx-dialect=C++11',
    '--with-debugging=0',
    '--with-eigen',
    '--with-eigen-dir=/opt/homebrew',
    '--with-fortran-bindings=0',
    '--with-openblas',
    '--with-openblas-dir=/opt/homebrew/opt/openblas',
    '--with-shared-libraries=1',
    '--with-zlib',
    'PETSC_ARCH=arch-darwin-c-opt',
  ]
  configure.petsc_configure(configure_options)
```

Installation with trunk firedrake-install: `python3 firedrake-install --no-package-manager --honour-petsc-dir  --slepc --mpicc=mpicc --mpif90=mpif90 --mpiexec=mpiexec --mpicxx=mpicxx --venv-name testing`

dham:
Superceded by #2408

PR #2414, opened 2022-04-16
Load binary gmsh file

lrtfm:


dham (changes requested):
This looks good and thanks for the contribution. However, we need a very basic test just so we know if this stops working.

What needs to happen is:

1. Add a binary `.msh` file to `tests/meshes`. A binary version of one of the files that is in there would be fine.
2. add a test file, say, `tests/regression/test_load_mesh.py` containing a test that loads the mesh and does something really basic. For example you could integrate 1 over the mesh and check that it gets the expected value.

dham (changes requested):
Oh, and you need to pass flake8 ;)

lrtfm:
@dham I think the problems have been resolved. Can this be merged?

dham (dismissed):

dham (approved):

PR #2424, opened 2022-05-05
Allow alternative name for saving Functions

danshapero:
`DumbCheckpoint.store` had an optional `name` keyword argument so that a function could be saved with a name other than its actual name attribute. `CheckpointFile.save_function` doesn't have this option; this patch adds it. The mechanism is a little blunt but it works.

### DIFF
--- a/firedrake/checkpointing.py
+++ b/firedrake/checkpointing.py
@@ -711,8 +711,7 @@ class CheckpointFile(object):
         :arg name: optional alternative name to save the function under
         """
         if name:
-            g = Function(f.function_space(), name=name)
-            g.assign(f)
+            g = Function(f.function_space(), val=f.dat, name=name)
             return self.save_function(g, idx=idx)
 
         # -- Save function space --
--- a/firedrake/checkpointing.py
+++ b/firedrake/checkpointing.py
@@ -699,7 +699,7 @@ class CheckpointFile(object):
                 topology_dm.setName(base_tmesh_name)
 
     @PETSc.Log.EventDecorator("SaveFunction")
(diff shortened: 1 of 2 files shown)


firedrake/checkpointing.py
```diff
@@ -708,7 +708,14 @@ def save_function(self, f, idx=None):
             mode (non-timestepping); for each function of interest,
             this method must always be called with the idx parameter
             set or never be called with the idx parameter set.
+        :arg name: optional alternative name to save the function under
         """
+        if name:
+            g = Function(f.function_space(), name=name)
+            g.assign(f)
+            self.save_function(g, idx=None)
+            return
```
dham, line 717: Rather than copying the data, you can make a Function which shares the same data space but has a different name. I also think you should be passing `idx` through rather than forcing it to None. i.e.
```
if name:
    return self.save_function(Function(f.function_space(), val=f, name=name)
```
I'm returning the result even though it's `None` just so that this will do the right thing if someone in the future decides that `save_function` should return something.

dham (changes requested):

firedrake/checkpointing.py
danshapero: You're definitely right about idx (silly mistake on part) and the return value. But with your proposed change, the test fails -- I find that `Function(f.function_space(), val=f, name=name).name()` is equal to `f.name()` rather than the supplied `name`.

danshapero: Alright I figured it out, you need to do `val=f.dat` and then it's fine.

dham (approved):

ksagiyam (approved):

PR #2431, opened 2022-05-09
Increase CI timeout

connorjward:
We've started hitting some occasional CI fails where tests have been timing out. I originally set the timeout to be 600s when the longest test seemed to take 400s. This appears to no longer be the case and failures due to timing out are happening quite often. We are only using `pytest-timeout` to catch tests that deadlock so I've bumped the timeout value to avoid this happening any more.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -58,7 +58,7 @@ jobs:
           . ../build/bin/activate
           echo OMP_NUM_THREADS is $OMP_NUM_THREADS
           echo OPENBLAS_NUM_THREADS is $OPENBLAS_NUM_THREADS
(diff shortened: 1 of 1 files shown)


sv2518:
Do you know why the tests need longer now? Is it because we added new (longer running) tests, because the codegen has changed and takes longer, or because the machine the tests are running on is more occupied?

connorjward:
> Do you know why the tests need longer now? Is it because we added new (longer running) tests, because the codegen has changed and takes longer, or because the machine the tests are running on is more occupied?

I don’t know what it is. I profiled the test that tends to time out and all of the time is spent generating code (it actually takes seconds to run a second time around). There were no obvious hotspots that would be us doing something bad.

dham (approved):

PR #2436, opened 2022-05-10
Fix weights and BCs when assembling the condensed residual in SCPC

pbrubeck:
When condensing the residual, one must take care not to add twice the DOFs shared across elements and zero out the BC nodes in the residual.

None of the previous tests were encountering these issues, since the condensed field always was DG Trace.

### DIFF
--- a/firedrake/slate/static_condensation/scpc.py
+++ b/firedrake/slate/static_condensation/scpc.py
@@ -255,7 +255,7 @@ class SCPC(SCBase):
             x.copy(v)
 
         # Disassemble the incoming right-hand side
-        with self.residual.split()[self.c_field].dat.vec_wo as vc, self.weight.dat.vec_ro as wc:
+        with self.residual.split()[self.c_field].dat.vec as vc, self.weight.dat.vec_ro as wc:
             vc.pointwiseMult(vc, wc)
 
         # Now assemble residual for the reduced problem
--- a/firedrake/slate/static_condensation/scpc.py
+++ b/firedrake/slate/static_condensation/scpc.py
@@ -101,7 +101,8 @@ class SCPC(SCBase):
         r_expr = reduced_sys.rhs
 
         # Construct the condensed right-hand side
-        self._assemble_Srhs = OneFormAssembler(r_expr, tensor=self.condensed_rhs, bcs=bcs,
+        self._assemble_Srhs = OneFormAssembler(r_expr, tensor=self.condensed_rhs,
+                                               bcs=bcs, zero_bc_nodes=True,
                                                form_compiler_parameters=self.cxt.fc_params).assemble
 
         # Allocate and set the condensed operator
--- a/firedrake/slate/static_condensation/scpc.py
+++ b/firedrake/slate/static_condensation/scpc.py
@@ -1,4 +1,5 @@
 import firedrake.dmhooks as dmhooks
+import numpy as np
 from firedrake.slate.static_condensation.sc_base import SCBase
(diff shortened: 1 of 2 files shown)


firedrake/slate/static_condensation/scpc.py
```diff
@@ -68,7 +71,7 @@ def initialize(self, pc):
         for bc in cxt_bcs:
             if bc.function_space().index != c_field:
                 raise NotImplementedError("Strong BC set on unsupported space")
-            bcs.append(DirichletBC(Vc, bc.function_arg, bc.sub_domain))
+            bcs.append(DirichletBC(Vc, 0, bc.sub_domain))
```
sv2518, line 74: Do we really want to set this to 0? Or did you only change it for debugging purposes?

firedrake/slate/static_condensation/scpc.py
```diff
@@ -77,14 +80,28 @@ def initialize(self, pc):
...
+        domain = "{[i,j]: 0 <= i < %d and 0 <= j < %d}" % shapes
+        instructions = """
+        for i, j
+            w[i,j] = w[i,j] + 1
+        end
+        """
+        self.weight = Function(Vc)
+        par_loop((domain, instructions), dx, {"w": (self.weight, INC)},
+                 is_loopy_kernel=True)
+        with self.weight.dat.vec as wc:
```
sv2518, line 94: I think you should move line 94 and 95 into the `forward_eliminiation` function so that we record the runtime cost of this in the right function (not in the initialisation). Or is there a reason why this has to be here?

sv2518 (commented):
Two minor comments, otherwise LGTM.

firedrake/slate/static_condensation/scpc.py
pbrubeck: We need to apply homogeneous BCs on the condensed residual regardless of what the inhomogenous values are.
Applying BCs to a matrix is also independent of the inhomogeneity.
The inhomogeneity is always taken care of by the outermost linear solver i.e. `LinearVariationalSolver` by modifying the original RHS.

If the preconditioner keeps track of the inhomogeneity and does something with it, such as applying these values to the solution, then it would no longer be a linear transformation, but an affine one. By homogenizing BCs within the PC, we prevent this sort of errors.

In (not so) old firedrake `assemble()` used to have the kwarg `assembly_type=residual`, that would just apply homogeneous BCs, this functionality is now hidden from the user. Nevertheless, this is can be achieved with `OneFormAssembler` via `zero_bc_nodes=True`.

sv2518: OK, I think I understand now

firedrake/slate/static_condensation/scpc.py
pbrubeck: I did this for efficiency, since multiplying times the reciprocal can be done faster than division. If we use a linear solver with static condensation, we might expect to reuse this result, and there is no need to update it every time we call `forward_elimination`.

HybridizationPC also computes the weight in the initialization, but does not take the reciprocal, instead it uses a kernel with division.

sv2518: Ok!

sv2518 (dismissed):

firedrake/slate/static_condensation/scpc.py
sv2518: Mhhh okay I will look into it.

dham (approved):

Issue #2185, opened 2021-08-18
No link between mesh and DM coordinates breaks VertexOnlyMesh


ReubenHill:
# Failing Script
Consider the below (as reported by @JDBetteridge )
```python
from firedrake import *

comm = COMM_WORLD
pcg = PCG64(seed=333)
rng = RandomGenerator(pcg)
dim = 2
deg = 1
Npoints = 20
samples = 10

# Pick a bunch of random points (same set on each rank)
if comm.rank == 0:
    points = rng.uniform(size=(Npoints, dim))
else:
    points = np.zeros((Npoints, dim))
comm.Bcast(points, root=0)

mesh = SquareMesh(32, 32, 2)
# Changing the coordinates breaks things!
mesh.coordinates.dat.data[:, :] -= 0.5

vom = VertexOnlyMesh(mesh, points)
```
if run on more than one rank, this fails with
```
Traceback (most recent call last):
  File "test3.py", line 22, in <module>
    vom = VertexOnlyMesh(mesh, points)
  File "PETSc/Log.pyx", line 115, in petsc4py.PETSc.Log.EventDecorator.decorator.wrapped_func
  File "PETSc/Log.pyx", line 116, in petsc4py.PETSc.Log.EventDecorator.decorator.wrapped_func
  File "/Users/rwh10/firedrake/src/firedrake/firedrake/mesh.py", line 2002, in VertexOnlyMesh
    dmcommon.label_pic_parent_cell_info(swarm, mesh)
  File "firedrake/cython/dmcommon.pyx", line 2710, in firedrake.cython.dmcommon.label_pic_parent_cell_info
    parent_cell_num, reference_coord = parentmesh.locate_cell_and_reference_coordinate(swarm_coords[i])
TypeError: an integer is required
application called MPI_Abort(MPI_COMM_WORLD, 1) - process 1
```

# Thoughts

My belief is that the line 
```python
mesh.coordinates.dat.data[:, :] -= 0.5
```
is naughty.

The `mesh` is sitting on top of a DMPlex which has an implicit coordinates field (made when `_from_cell_list` in `firedrake/mesh.py` is called) which I don’t think is updated.

This break things like creating a `VertexOnlyMesh` which relies on that underlying-plex coordinate field to decide on parallel distribution of the DMSwarm and whether-or-not the DMSwarm coordinates (the `VertexOnlyMesh` vertices) are within the coordinates field of the `mesh`:
[...]

PR #2437, opened 2022-05-11
Periodic vertexonlymesh

dham:
This PR swaps the order of mesh searching in the VertexOnlyMesh setup so that we do the Firedrake search for parent mesh cell inclusion before the PETSc one. We then use this to set up the swarm coordinates instead of calling `swarm.setPointCoordinates`. This means that we're not dependent on Firedrake and DMPlex agreeing on how coordinates work.

This needs a little cleanup and docstrings, and I need to do the parallel bounding box elimination (which isn't hard). I think the extension to extrusion is probably fairly straightforward and I think I might know how to do bendy.


ReubenHill:
I believe this is doing the "Quick" solution suggested in #2185 .

dham:
> I believe this is doing the "Quick" solution suggested in #2185 .

Sure, and I think that's good. It's a small amount of work which makes a lot of things work that don't currently. I don't think it's at all worse than what we're currently doing, and I don't think we should refrain from improving the current situation in favour of a potential future fix. If we ever get there then reversing this change is very easy.

firedrake/mesh.py
```diff
@@ -2254,7 +2249,7 @@ def compare_arrays(x, y, datatype):
     return vmesh
 
 
-def _pic_swarm_in_plex(plex, coords, fields=[]):
+def _pic_swarm_in_plex(parent_mesh, coords, fields=None):
```
ReubenHill, line 2252: By changing the input argument from `plex` to `parent_mesh` you have slightly muddied the functionality of this function I think. Are all meshes have guaranteed to have a DMPlex? The VertexOnlyMesh certainly doesn't. I preferred passing the plex in directly

ReubenHill: I think rather than fiddling with this function you can just call
```
    coords, reference_coords, parent_cell_nums = \
        _parent_mesh_embedding(coords, parent_mesh)
```
after 
```
swarm = _pic_swarm_in_plex(mesh.topology.topology_dm, vertexcoords, fields=[("parentcellnum", 1, IntType), ("refcoord", tdim, RealType)])
```
and then fiddle about with replacing the field entries. 

As far as I can tell we aren't getting any performance benefit from doing it inside this function since the default PETSc embedding is still going on. This would also keep the diff nice and tidy.

firedrake/mesh.py
```diff
@@ -2366,6 +2397,44 @@ def _pic_swarm_in_plex(plex, coords, fields=[]):
     return swarm
 
 
+def _parent_mesh_embedding(vertex_coords, parent_mesh):
```
ReubenHill, line 2400: This needs moving to cython. It all looks quite similar to stuff I've got there already which is handy

ReubenHill (changes requested):
Needs some changes but if this works without significant performance losses at `VertexOnlyMesh` creation time then great

firedrake/mesh.py
dham: I don't think this is necessary. The code which is in cython immediately calls back to plain python anyway so I don't see what we get by being in cython.

ReubenHill: I was merely thinking about the loop which, in many cases, will be over large numbers of vertices. Happy to be educated though

dham: The call back to Python is inside that loop. That's why I think cython doesn't gain us anything here.

ReubenHill: Well nevertheless (and this would be a fun thing to profile at some point) this seems to be very similar to this function: https://github.com/firedrakeproject/firedrake/blob/9c9bc952e082f066b0f9ffe547e02d95fb4d78c9/firedrake/cython/dmcommon.pyx#L2907-L2980

firedrake/mesh.py
```diff
@@ -2351,7 +2361,28 @@ def _pic_swarm_in_plex(plex, coords, fields=[]):
...
+    # swarm.restoreField is called for each field too!
+    swarm_coords = swarm.getField("DMSwarmPIC_coor").reshape((num_vertices, gdim))
+    # Why are the next two different?
+    swarm_parent_cell_nums = swarm.getField("DMSwarm_cellid")
+    field_parent_cell_nums = swarm.getField("parentcellnum")
+    field_reference_coords = swarm.getField("refcoord").reshape((num_vertices, tdim))
+
+    swarm_coords[...] = coords
+    swarm_parent_cell_nums[...] = parent_cell_nums
+    field_parent_cell_nums[...] = parent_cell_nums
```
ReubenHill, line 2378: Check these are actually the same in all/most cases

firedrake/mesh.py
ReubenHill: Having discussed, let's just rename this `_pic_swarm_in_mesh`

dham: Done.

firedrake/mesh.py
dham: We discussed this offline. Both of these functions almost certainly do not gain from being in cython.

firedrake/mesh.py
dham: OK so it turns out they aren't. We use a permuted plex. The permutation (of all points in the plex) is `parent_mesh._plex_renumbering`. It's a PETSc IS. If I generate a PETSc section with the global numbering for what is essentially DG[0] (i.e. one DoF per cell) then that section encodes the renumbering from the PETSc cell numbers to the Firedrake ones.

I guess what I need is an IS equivalent to the contents of that section. I could then invert it and apply it to the cell numbers we have in order to get the PETSc cell numbers. I am not sure how to do that.

wence-: https://github.com/firedrakeproject/firedrake/blob/bae7e705755ae95cf4fc4efec488b1b6ec1b7955/firedrake/cython/mgimpl.pyx#L19

tests/vertexonly/test_interpolation_from_parent.py
```diff
@@ -133,6 +131,13 @@ def test_scalar_function_interpolation(parentmesh, vertexcoords, fs):
     vm = VertexOnlyMesh(parentmesh, vertexcoords)
     vertexcoords = vm.coordinates.dat.data_ro.reshape(-1, parentmesh.geometric_dimension())
     fs_fam, fs_deg, fs_typ = fs
+    if (
+        parentmesh.coordinates.function_space().ufl_element().family()
+        == "Discontinuous Lagrange"
+        and fs_fam == "CG"
+    ):
+        pytest.skip(f"Interpolating into f{fs_fam} on a periodic mesh is not well-defined.")
```
wence-, line 139: feels like the frontend interpolation should raise a `ValueError` so that users don't get this ill-definedness.

firedrake/mesh.py
```diff
@@ -2217,8 +2213,6 @@ def compare_arrays(x, y, datatype):
             else:
                 raise ValueError("missing_points_behaviour must be None, 'error' or 'warn'")
 
-    dmcommon.label_pic_parent_cell_info(swarm, mesh)
-
```
ReubenHill, line 2221: You can safely delete this function from dmcommon now I think.

firedrake/mesh.py
ReubenHill: Can we add a test that uses both `swarm.setPointCoordinates` and our new drop in replacement to make sure we get the same numbering?

tests/vertexonly/test_interpolation_from_parent.py
dham: It's a little more complex than that though. If one takes a function which is actually periodic and does the interpolation then it's all fine, and that's an obvious use case. It's just in general that the interpolation isn't defined.

firedrake/mesh.py
dham: OK this is done now thanks to @ksagiyam who pointed out the easy way of getting the permutation I needed. I have also extended the swarm tests to check that we get the same that PETSc would.

Along the way I worked out that we can also remove points in the cell nodes for free so we no longer have to separately do that.

firedrake/mesh.py
```diff
@@ -2366,6 +2381,57 @@ def _pic_swarm_in_plex(plex, coords, fields=[]):
     return swarm
 
 
+def _parent_mesh_embedding(vertex_coords, parent_mesh):
+    """Find the parent cells and local coords for vertices on this rank."""
```
ReubenHill, line 2385: You should specifically note that this never puts any vertices in the parent mesh ghost cells I think

firedrake/mesh.py
```diff
@@ -2366,6 +2381,57 @@ def _pic_swarm_in_plex(plex, coords, fields=[]):
...
+    valid = np.full(num_vertices, False)
+
+    # Create an out of mesh point to use in locate_cell when needed
+    out_of_mesh_point = np.full((1, gdim), np.inf)
+
+    for i in range(max_num_vertices):
+        if i < num_vertices:
+            parent_cell_num, reference_coord = \
+                parent_mesh.locate_cell_and_reference_coordinate(vertex_coords[i])
+            if parent_cell_num is not None and parent_cell_num < parent_mesh.cell_set.size:
```
ReubenHill, line 2406: Comment here please about `parent_cell_num < parent_mesh.cell_set.size` avoiding putting vertices in the parent mesh halo

ReubenHill (dismissed):
This looks good pending some tidying on the commit history and adding a small amount of documentation. I hope the tests pass!

ReubenHill:
Too late to fix it now, but I note you merged before cleaning up the commit history which was a requested change - things like squashing `flake8` and `lint` into the previous commits

ReubenHill:
Great that this all passed tests and works though! Thanks for the PR 😃

PR #2441, opened 2022-05-12
Pin setuptools and wheel to match numpy build requirements

JDBetteridge:
Build failures have been observed with mismatched setuptools and wheel versions.

(https://github.com/firedrakeproject/firedrake/runs/6402346437?check_suite_focus=true)

It appears numpy have pinned these to specific versions.


sv2518:
Yes should be correct, this is what I did locally and it worked.

PR #2446, opened 2022-05-18
Fix BOP

sv2518:
### Problem number 1: Recursion
I realised that the block push optimisation pass is aborting too early, so that we are missing out on performance gains. We need to recurse into every child to check if there is a `Block`. Examples where this is the case were introduced in 8a36889cad10aa2646812f848c847eaac64e521b. This also seemed to be an issue for all tests involving the SCPC preconditioner `tests/slate/test_cg_poisson.py`.

**Fix**: 342639742c60b7c8bc56fb7d0d9a4e27155e898a

### Problem number 2: Shapes of BAVs
After fixing the recursion for the block push, the first issue I had was a shape mismatch error in the multiplicand for the very first test, because shapes for `BlockAssembledVectors` (BAV) were not determined correctly.

**Fix**
I am passing the whole block into the BAV here https://github.com/firedrakeproject/firedrake/blob/d7f30976e00aa294935c99b89e5b0673141662f4/firedrake/slate/slac/optimise.py#L103 rather than the form associated with the block, so that I could recycle its `arguments` and `argfunctionspaces` here https://github.com/firedrakeproject/firedrake/blob/d7f30976e00aa294935c99b89e5b0673141662f4/firedrake/slate/slate.py#L543 and here https://github.com/firedrakeproject/firedrake/blob/d7f30976e00aa294935c99b89e5b0673141662f4/firedrake/slate/slate.py#L532 Those are used to determine the shape.

### Problem number 3: Coefficient maps
After that I started to have issues with the coefficient maps. The general story behind this is that PyOP2 is passing the coefficients in the order that they appear in the original Slate expression. We then rewrite the Slate expression and the order might change. I had introduced a coefficient map which maps the original coefficient order and the optimised coefficient order.
The problem for blocks on AssembledVectors: If we push blocks inward, we might start from this expression here `AV_4[((1,),)]_3 + -(M_6[((1,), (0,))]_5 * (PartialPivLU(M_6[((0,), (0,))]_8)_7).inv * AV_4[((0,),)]_9)` and rewrite it to `BAV_10 + -(M_11 * PartialPivLU(M_13)_12 \ BAV_14)`. `AV_4` is an assembled vector on a mixed coefficient and we index two times into a part of it, this gets then turned into two `BlockAssembledVectors (BAV)` in the optimisation pass. PyOP2 passes splits of the mixed coefficients in order (first split, second split, etc), but the Slate compiler passes the coefficient arguments to the kernels in the order of how they appear in the optimised expression. `BAV_10` corresponds to the second part of the (original) mixed coefficient, but it’s the first argument in the Slate kernel. This means that the we don’t only need to order the coefficients but also their splits.

**Fix**
In order to sort the the split map we need to save more information when we ask for the coefficients of a BAV which is why querying BAVs for their coefficients will return `BlockFunctions` now, as introduced here https://github.com/firedrakeproject/firedrake/blob/d7f30976e00aa294935c99b89e5b0673141662f4/firedrake/slate/slate.py#L46 In the `BlockFunction`, in addition to the split functions corresponding to the BAV, we also keep track of indices of the block and also the original function, so that we can generate the correct split map here https://github.com/firedrakeproject/firedrake/blob/d7f30976e00aa294935c99b89e5b0673141662f4/firedrake/slate/slate.py#L217
We ask for the coefficients two more times. Once to generate a coefficient dict (which is used to generate the arguments to the Slate kernel e.g.) and once to fill the temporaries (corresponding to the Slate assembled vectors) with the data that are passed to the Slate kernels. We need to adapt there to the changes above and that is what is happening in the changes of the `KernelBuilder`

### DIFF
--- a/firedrake/slate/slac/compiler.py
+++ b/firedrake/slate/slac/compiler.py
@@ -45,7 +45,7 @@ import numpy as np
 import loopy
 import gem
 from gem import indices as make_indices
-from tsfc.kernel_args import OutputKernelArg
+from tsfc.kernel_args import OutputKernelArg, CoefficientKernelArg
 from tsfc.loopy import generate as generate_loopy
 import copy
 
@@ -191,6 +191,10 @@ def generate_loopy_kernel(slate_expr, compiler_parameters=None):
     orig_coeffs = orig_expr.coefficients()
     get_index = lambda n: orig_coeffs.index(new_coeffs[n]) if new_coeffs[n] in orig_coeffs else n
     coeff_map = tuple((get_index(n), split_map) for (n, split_map) in slate_expr.coeff_map)
+
+    coefficients = list(filter(lambda elm: isinstance(elm, CoefficientKernelArg), arguments))
+    assert len(list(chain(*(map[1] for map in coeff_map)))) == len(coefficients), "KernelInfo must be generated with a coefficient map that maps EXACTLY all cofficients there are in its arguments attribute."
+    assert len(loopy_merged.callables_table[name].subkernel.args) - int(builder.bag.needs_mesh_layers) == len(arguments), "Outer loopy kernel must have the same amount of args as there are in arguments"
 
     kinfo = KernelInfo(kernel=loopykernel,
                        integral_type="cell",  # slate can only do things as contributions to the cell integrals
--- a/firedrake/slate/slac/compiler.py
+++ b/firedrake/slate/slac/compiler.py
@@ -187,11 +187,10 @@ def generate_loopy_kernel(slate_expr, compiler_parameters=None):
                                                        events=events+(slate_loopy_event,))
 
     # map the coefficients in the order that PyOP2 needs
-    new_coeffs = slate_expr.coefficients()
     orig_coeffs = orig_expr.coefficients()
-    get_index = lambda n: orig_coeffs.index(new_coeffs[n]) if new_coeffs[n] in orig_coeffs else n
-    coeff_map = tuple((get_index(n), split_map) for (n, split_map) in slate_expr.coeff_map)
-
+    new_coeffs = slate_expr.coefficients()
+    map_new_to_orig = [orig_coeffs.index(c) for c in new_coeffs]
+    coeff_map = tuple((map_new_to_orig[n], split_map) for (n, split_map) in slate_expr.coeff_map)
     coefficients = list(filter(lambda elm: isinstance(elm, CoefficientKernelArg), arguments))
     assert len(list(chain(*(map[1] for map in coeff_map)))) == len(coefficients), "KernelInfo must be generated with a coefficient map that maps EXACTLY all cofficients there are in its arguments attribute."
     assert len(loopy_merged.callables_table[name].subkernel.args) - int(builder.bag.needs_mesh_layers) == len(arguments), "Outer loopy kernel must have the same amount of args as there are in arguments"
--- a/firedrake/slate/slac/kernel_builder.py
+++ b/firedrake/slate/slac/kernel_builder.py
@@ -579,15 +579,13 @@ class LocalLoopyKernelBuilder(object):
         """
         coeffs = self.expression.coefficients()
         coeff_dict = OrderedDict()
-        for i, c in enumerate(coeffs):
-            element = c.ufl_element()
-            if type(element) == MixedElement:
-                mixed = OrderedDict()
-                for j, c_ in enumerate(c.split()):
-                    mixed[c_] = f"w_{i}_{j}", self.extent(c_)
-                coeff_dict[c] = mixed
+        for i, (c, split_map) in enumerate(self.expression.coeff_map):
+            coeff = coeffs[c]
+            if type(coeff.ufl_element()) == MixedElement:
+                splits = coeff.split()
+                coeff_dict[coeff] = OrderedDict({splits[j]: (f"w_{i}_{j}", self.extent(splits[j])) for j in split_map})
             else:
-                coeff_dict[c] = f"w_{i}", self.extent(c)
+                coeff_dict[coeff] = (f"w_{i}", self.extent(coeff))
         return coeff_dict
 
     def initialise_terminals(self, var2terminal, coefficients):
@@ -615,14 +613,22 @@ class LocalLoopyKernelBuilder(object):
                                               within_inames=frozenset(inames)))
 
             else:
-                f = slate_tensor.form if isinstance(slate_tensor.form, tuple) else (slate_tensor.form,)
-                coeff = tuple(coefficients[c] for c in f)
+                potentially_mixed_f = slate_tensor.form if isinstance(slate_tensor.form, tuple) else (slate_tensor.form,)
+                coeff_dict = tuple(coefficients[c] for c in potentially_mixed_f)
                 offset = 0
-                ismixed = tuple((type(c.ufl_element()) == MixedElement) for c in f)
+                ismixed = tuple((type(c.ufl_element()) == MixedElement) for c in potentially_mixed_f)
+                # Fetch the coefficient name corresponding to this assembled vector
                 names = []
-                for (im, c) in zip(ismixed, coeff):
-                    names += [name for (name, ext) in c.values()] if im else [c[0]]
-
+                for (im, c) in zip(ismixed, coeff_dict):
+                    if im:
+                        # For block assembled vectors we need to pick the right name
+                        # corresponding to the split function of the block assembled vector
+                        block_function, = slate_tensor.slate_coefficients()
+                        filter_f = lambda name_and_extent: (not isinstance(block_function, slate.BlockFunction)
+                                                            or name_and_extent[0] in block_function.split_function)
+                        names += list(name for (_, (name, _)) in tuple(filter(filter_f, c.items())))
+                    else:
+                        names += [c[0]]
                 # Mixed coefficients come as seperate parameter (one per space)
                 for i, shp in enumerate(*slate_tensor.shapes.values()):
                     indices = self.bag.index_creator((shp,))
--- a/firedrake/slate/slac/optimise.py
+++ b/firedrake/slate/slac/optimise.py
@@ -78,7 +78,7 @@ def _push_block_transpose(expr, self, indices):
 @_push_block.register(Reciprocal)
 def _push_block_distributive(expr, self, indices):
     """Distributes Blocks for these nodes"""
-    return type(expr)(*map(self, expr.children, repeat(indices))) if indices else expr
+    return type(expr)(*map(self, expr.children, repeat(indices)))
 
 
 @_push_block.register(Factorization)
@@ -87,6 +87,7 @@ def _push_block_distributive(expr, self, indices):
 @_push_block.register(Mul)
 def _push_block_stop(expr, self, indices):
     """Blocks cannot be pushed further into this set of nodes."""
+    expr = type(expr)(*map(self, expr.children, repeat(tuple())))
     return Block(expr, indices) if indices else expr
 
 
--- a/firedrake/slate/slac/optimise.py
+++ b/firedrake/slate/slac/optimise.py
@@ -100,7 +100,7 @@ def _push_block_tensor(expr, self, indices):
 @_push_block.register(AssembledVector)
 def _push_block_assembled_vector(expr, self, indices):
     """Turns a Block on an AssembledVector into the  specialized node BlockAssembledVector."""
-    return BlockAssembledVector(expr._function, Block(expr, indices).form, indices) if indices else expr
+    return BlockAssembledVector(expr._function, expr, indices) if indices else expr
 
 
 @_push_block.register(Block)
--- a/firedrake/slate/slate.py
+++ b/firedrake/slate/slate.py
@@ -16,7 +16,7 @@ functions to be executed within the Firedrake architecture.
 """
 from abc import ABCMeta, abstractproperty, abstractmethod
 
-from collections import OrderedDict
+from collections import OrderedDict, namedtuple, defaultdict
 
 from ufl import Coefficient, Constant
 
@@ -41,6 +41,16 @@ __all__ = ['AssembledVector', 'Block', 'Factorization', 'Tensor',
            'Inverse', 'Transpose', 'Negative',
            'Add', 'Mul', 'Solve', 'BlockAssembledVector', 'DiagonalTensor',
            'Reciprocal']
+
+# BlockFunction description type
+BlockFunction = namedtuple('BlockFunction', ['split_function', 'indices', 'orig_function'])
+BlockFunction.__doc__ = """\
+Context that carries information for a block on an assembled vector.
+
+:param split_function: The splits of the orig_function corresponding to the block.
+:param indices: The indices of the block.
+:param orig_function: The  (unsplit) function corresponding the assembled vector.
+"""
 
 
 class RemoveNegativeRestrictions(MultiFunction):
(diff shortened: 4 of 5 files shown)


sv2518:
Okay here are the timings that were asked for yesterday in the meeting.

I am timing the assembly of the `expr` in the following code snippet on a warm cache. N is 400 ad p is in a range of 5.
```
    mesh = SquareMesh(N, N, 1, quadrilateral=True)
    dg1 = FunctionSpace(mesh, "DG", p+1)
    dg0 = FunctionSpace(mesh, "DG", p)
    vector_dg1 = VectorFunctionSpace(mesh, "DG", 1)
    dg1_dg1_dg0 = vector_dg1 * dg1 * dg0
    x = SpatialCoordinate(mesh)
    q = Function(vector_dg1).project(grad(sin(pi*x[0])*cos(pi*x[1])))
    p = Function(dg1).interpolate(-x[0]*exp(-x[1]**2))
    r = Function(dg0).assign(42.0)
    functions = [q, p, r]
    q, p, r = functions
    u, phi, eta = TrialFunctions(dg1_dg1_dg0)
    v, psi, nu = TestFunctions(dg1_dg1_dg0)
    K = Tensor(Constant(2)*inner(u, v)*dx + inner(phi, psi)*dx + inner(eta, nu)*dx)
    F = AssembledVector(assemble(inner(q, v)*dx + inner(p, psi)*dx + inner(r, nu)*dx))
    F2 = AssembledVector(assemble(Constant(2)*inner(q, v)*dx + inner(p, psi)*dx + inner(r, nu)*dx))
    expr = (F.blocks[0:2] - K.blocks[0:2, 0:2] * F.blocks[0:2])
```
bop

You can see in the xlabel of the figure how much less DOFS have to be locally (!) assembled for the matrix in the expression (i.e. I compare `K` vs `K.blocks[0:2, 0:2]`). I also plotted this reduction as ratio in the following.
bop2

firedrake/slate/slate.py
```diff
@@ -209,10 +219,32 @@ def coeff_map(self):
...
+            # BlockFunctions for c need special treatment because:
+            # 1) There can be multiple BlockFunctions defined on the same original function
+            # -> The split functions in c might need to be appended to an already exisiting entry in the map
+            # 2) BlockAssembledVectors can appear in any order they like in the Slate expression
+            # -> The split functions need to be appended in the same order into the map
+            if isinstance(c, BlockFunction):  # for block assembled vectors
+                # Did we already add a part of the originial function of this BlockFunction to the map?
+                orig_to_pos = tuple(*filter(lambda item: item[0] == c.orig_function, orig_to_pos_dict.items()))
+                pos = orig_to_pos[1] if orig_to_pos else m
+                split_map = c.indices[0] if pos == m else coeff_map[pos][1]+c.indices[0]
```
ksagiyam, line 235: The following example fails:
```
from firedrake import *
  

mesh = UnitSquareMesh(1, 1)
V0 = FunctionSpace(mesh, "DG", 0)
V1 = FunctionSpace(mesh, "DG", 1)
V = V0 * V1
u = TrialFunction(V)
v = TestFunction(V)
K = Tensor(inner(u, v) * dx)
f = AssembledVector(assemble(inner(as_vector([1., 1.]), v)*dx))
expr = (K*f).blocks[0] + f.blocks[0]
assemble(expr, form_compiler_parameters={"slate_compiler": {"optimise": True}})
```
with:
```
AssertionError: KernelInfo must be generated with a coefficient map that maps EXACTLY all cofficients there are in its arguments attribute.
```
I think this is because we use tuple-add here instead of set-update:

`coeff_map` for this example is indeed:
```
((0, (0, 1, 0)),)
```

firedrake/slate/slac/kernel_builder.py
```diff
@@ -580,14 +580,15 @@ def collect_coefficients(self):
...
-                mixed = OrderedDict()
-                for j, c_ in enumerate(c.split()):
+            coeff = c.orig_function if isinstance(c, slate.BlockFunction) else c
+            if type(coeff.ufl_element()) == MixedElement:
+                mixed = OrderedDict() if coeff not in coeff_dict else coeff_dict[coeff]
+                splits = c.split_function if isinstance(c, slate.BlockFunction) else coeff.split()
+                for j, c_ in enumerate(splits):
                     mixed[c_] = f"w_{i}_{j}", self.extent(c_)
-                coeff_dict[c] = mixed
+                coeff_dict.update({coeff: mixed})
```
ksagiyam, line 589: I think, here, we can loop over `self.expression.coeff_map` instead of over `self.expression.coefficients()`, since we have already removed duplication in `coeff_map`?

ksagiyam (changes requested):

firedrake/slate/slate.py
sv2518: If I use a plain set we loose the order on the split_map, but I will push a fix with a sorted set when I get to it. Thanks!

ksagiyam: Does the order in split_map need to be preserved merely to be consistent with the order of `kernel_arg`? If so, I think the order can be arbitrary once you amend `collect_coefficients` to use `coeff_map`'s order (see below).

firedrake/slate/slac/kernel_builder.py
sv2518: The `coeff_map` does not carry enough information to generate the kernel arguments here. In particular we need the extent. We either need the extent as explicit information or we can deduce it from a coefficient (like we are doing atm)

firedrake/slate/slac/compiler.py
```diff
@@ -192,6 +192,10 @@ def generate_loopy_kernel(slate_expr, compiler_parameters=None):
     get_index = lambda n: orig_coeffs.index(new_coeffs[n]) if new_coeffs[n] in orig_coeffs else n
```
ksagiyam, line 192: I think I have to understand these lines. `slate_expr.coefficients()` includes `BlockFunction` objects, but is it expected? Related is `get_index`; is it assuming something about how we collect coefficients?

ksagiyam (changes requested):

firedrake/slate/slac/compiler.py
sv2518: Yes it's expected that `BlockFunctions` are included here, I have just pushed a fix to account for that in the collection of the indices here. 
And yes we assume that the coefficients in `new_coeffs` or `orig_coeff` are collected in a specific order, which is the one they are appearing in in the corresponding expression.

firedrake/slate/slate.py
sv2518: I pushed a fix for this here https://github.com/firedrakeproject/firedrake/pull/2446/commits/92c705728a5ad5953f0a14ea44d50b569a2d15d9. As I said in a comment below I think that coefficients map does not carry enough information to the collect the coefficients for the kernel argument generation, it's just indices and nothing else

firedrake/slate/slac/compiler.py
ksagiyam: Quick question: is there a way to collect `BlockFunction` objects given a slate expression?

sv2518: This is untested but sth like this should work:
```
is_blockf = lambda s: True if isinstance(s, BlockFunction) else False
blockfuncs = filter(is_blockf, slate_expr.coefficients())
```

sv2518: You can see something similar in my fix here https://github.com/firedrakeproject/firedrake/pull/2446/commits/35f78e03867f388b700e84c1681ceee7c05e51ee

ksagiyam: Thanks. I'm just trying to see if we can simplify the implementation.

firedrake/slate/slate.py
ksagiyam: If we flip the order, the above example still fails:

From:
```
expr = (K*f).blocks[0] + f.blocks[0]
```
To:
```
expr = f.blocks[0] + (K*f).blocks[0]
```
You could directly fix this, but see below comment.

firedrake/slate/slac/compiler.py
ksagiyam: So I think the way `get_index` currently works is somewhat fragile; I think we should assume that people can do all sorts of optimisation which can reorder coefficients. Also, I still think `collect_coefficients()` should loop over `coeff_map` to make sure that the order of kernerl arguments and that of coefficients (collected for PyOP2) would be the same.

Do you think the following works, or am I missing something?

- Add `raw_coeffcients()` method:
```
def raw_coefficients(self):
    return self.coefficients()  # change this for `BlockedAssembledVector`
```
- `slate/slac/compiler.py`:
```
...
orig_coeffs = orig_expr.coefficients()
new_coeffs = slate_expr.raw_coefficients()
map_new_to_orig = [orig_coeffs.index(c) for c in new_coeffs]
coeff_map = ((map_new_to_orig(n), split_map) for (n, split_map) in slate_expr.coeff_map)
...
```

- `slate/slate.py`:
Change `coeff_map` so that the target index would be `m = self.raw_coefficients().index(some_raw_coeff)` instead of incrementing `m`.
Sort collected indices.

- `slate/slac/kernel_builder.py`:
```
    def collect_coefficients(self):
        coeffs = self.expression.raw_coefficients()
        coeff_dict = OrderedDict()
        for i, (c, split_map) in enumerate(self.expression.coeff_map):
            coeff = coeffs[c]
            if type(coeff.ufl_element()) == MixedElement:
                mixed = OrderedDict() if coeff not in coeff_dict else coeff_dict[coeff]
                for j in split_map:
                    coeff_ = coeff.split()[j]
                    mixed[coeff_] = f"w_{i}_{j}", self.extent(coeff_)
                coeff_dict.update({coeff: mixed})
            else:
                coeff_dict[coeff] = f"w_{i}", self.extent(coeff)
        return coeff_dict
```

firedrake/slate/slac/optimise.py
```diff
@@ -99,7 +100,7 @@ def _push_block_tensor(expr, self, indices):
 @_push_block.register(AssembledVector)
 def _push_block_assembled_vector(expr, self, indices):
     """Turns a Block on an AssembledVector into the  specialized node BlockAssembledVector."""
-    return BlockAssembledVector(expr._function, Block(expr, indices).form, indices) if indices else expr
+    return BlockAssembledVector(expr._function, Block(expr, indices), indices) if indices else expr
```
ksagiyam, line 103: By the way, can we change the interface so that `BlockAssembledVector` would take `expr` and `indices` as arguments?

sv2518: Yes done

firedrake/slate/slate.py
```diff
@@ -203,16 +213,49 @@ def rank(self):
...
+                    coeff_map[pos] = (pos, split_map)
+                else:
+                    # We didn't -> we add normally and keep track that we did in the orig_to_pos_dict
+                    coeff_map += [(pos, split_map)]
+                orig_to_pos_dict[c.orig_function] = pos
+            else:
+                split_map = tuple(range(len(c.split()))) if isinstance(c, Function) or isinstance(c, Constant) else tuple(range(1))
+                coeff_map += [(m, split_map)]
+                orig_to_pos_dict[c] = m
+        return tuple(coeff_map)
```
ksagiyam, line 257: Maybe something this works:
```
    @property
    def coeff_map(self):
        coeff_map = defaultdict(set)
        for c in self.slate_coefficients():
            if isinstance(c, BlockFunction):  # for block assembled vectors
                m = self.coefficients().index(c.orig_function)
                coeff_map[m].update(c.indices[0])
            else:
                m = self.coefficients().index(c)
                split_map = tuple(range(len(c.split()))) if isinstance(c, Function) or isinstance(c, Constant) else tuple(range(1))
                coeff_map[m].update(split_map)
        return tuple((k, tuple(sorted(v)))for k, v in coeff_map.items())
```
`dict` is ordered for `>=3.6`(?), but I don't think it matters here since we now order in `collect_coefficients` using `coeff_map`. We need to check if this works with `initialize_temiminals`, though.

firedrake/slate/slac/kernel_builder.py
```diff
@@ -577,17 +577,20 @@ def collect_coefficients(self):
...
-                    mixed[c_] = f"w_{i}_{j}", self.extent(c_)
-                coeff_dict[c] = mixed
+        for i, (c, split_map) in enumerate(self.expression.coeff_map):
+            coeff = coeffs[c]
+            splits = coeff.split_function if isinstance(coeff, slate.BlockFunction) else coeff.split() if type(coeff.ufl_element()) == MixedElement else ()
+            orig_coeff = coeff.orig_function if isinstance(coeff, slate.BlockFunction) else coeffs[c]
+            if splits:
+                mixed = OrderedDict() if coeff not in coeff_dict else coeff_dict[coeff]
+                for j in split_map:
+                    coeff_ = splits[j]
```
ksagiyam, line 589: `splits = coeff.split_function` if `coeff` is `BlockFunction`, but `split_map` is a tuple of indices into `coeff.orig_function.split()`, right?

ksagiyam (changes requested):

firedrake/slate/slac/compiler.py
```diff
@@ -187,10 +187,13 @@ def generate_loopy_kernel(slate_expr, compiler_parameters=None):
                                                        events=events+(slate_loopy_event,))
 
     # map the coefficients in the order that PyOP2 needs
-    new_coeffs = slate_expr.coefficients()
     orig_coeffs = orig_expr.coefficients()
-    get_index = lambda n: orig_coeffs.index(new_coeffs[n]) if new_coeffs[n] in orig_coeffs else n
-    coeff_map = tuple((get_index(n), split_map) for (n, split_map) in slate_expr.coeff_map)
+    new_coeffs = slate_expr.coefficients()
+    map_new_to_orig = [orig_coeffs.index(c) for c in new_coeffs]
+    coeff_map = tuple((map_new_to_orig[n], split_map) for (n, split_map) in slate_expr.coeff_map)
```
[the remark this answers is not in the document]
sv2518: The index will be the same for `slate_expr.coefficients` and `slate_expr.slate_coefficients`, the only thing that changes it the object with corresponds to the `BlockAssembledVector`, so this should not be a problem. This is relevant for the other comments as well (besides for then one on the coefficients_map)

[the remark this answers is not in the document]
sv2518: Basically I am using coefficients when I want the original function of the `BlockAssembledVector` and slate_coefficients if I want a `BlockFunction` because I need more information.

firedrake/slate/slac/kernel_builder.py
ksagiyam: Ok, so I think `splits` here must be `coeff.orig_function.split()` instead of `coeff.split_function`; otherwise, if you replace indices `0` with `1` in `test_push_assembled_vector_blocks_individual`, this test fails.

firedrake/slate/slac/kernel_builder.py
```diff
@@ -577,18 +577,17 @@ def collect_coefficients(self):
...
             if splits:
                 mixed = OrderedDict() if coeff not in coeff_dict else coeff_dict[coeff]
                 for j in split_map:
                     coeff_ = splits[j]
                     mixed[coeff_] = f"w_{i}_{j}", self.extent(coeff_)
-                coeff_dict.update({orig_coeff: mixed})
+                coeff_dict.update({coeff: mixed})
             else:
                 coeff_dict[coeff] = f"w_{i}", self.extent(coeff)
         return coeff_dict
```
ksagiyam, line 593:
```
    def collect_coefficients(self):
        """Saves all coefficients of self.expression where non-mixed coefficients
        are dicts of form {coeff: (name, extent)} and mixed coefficients are
        double dicts of form {mixed_coeff: {coeff_per_space: (name, extent)}}.
        """
        coeffs = self.expression.coefficients()
        coeff_dict = OrderedDict()
        for i, (c, split_map) in enumerate(self.expression.coeff_map):
            coeff = coeffs[c]
            if type(coeff.ufl_element()) == MixedElement:
                splits = coeff.split()
                coeff_dict[coeff] = OrderedDict({splits[j]: (f"w_{i}_{j}", self.extent(splits[j])) for j in split_map})
            else:
                coeff_dict[coeff] = (f"w_{i}", self.extent(coeff))
        return coeff_dict
```
?

ksagiyam (approved):

dham (approved):

ksagiyam (approved):

Issue #2224, opened 2021-09-20
VertexOnlyMesh Point Location Tolerance and Supported Mesh Improvements


ReubenHill:
As identified in #2178 , VertexOnlyMesh's point location and rank distribution system has no way of specifying a tolerance to avoid the loss of points on parent mesh cell or mesh boundaries. This is because it relies on [DMSwarmSetPointCoordinates](https://petsc.org/release/docs/manualpages/DMSWARM/DMSwarmSetPointCoordinates.html) which doesn't allow such a tolerance to be set.

To get around this I should use `mesh.locate_cell` as the initial cell-location method rather than relying on `DMSwarmSetPointCoordinates`. This may also have the advantage of allowing high-order (bendy) meshes and meshes made from cell coordinates to be supported.

Note that firedrake cell numbering will still need to be translated to DMPlex cell numbering if I want to keep the underlying DMSWARM updated in PIC mode (which would keep features like DMSwarmMigrate working).

ReubenHill:
Note that different ranks might make different decisions about which cell a point lives in so care will need to be taken that points on cell boundaries don't get duplicated by accident when running in parallel. 

Take the following `test_parallel_point_locate.py` for example
```python
from firedrake import *
m = UnitSquareMesh(1,1)
coords = [[0.5, 0.5], [0.9, 0.9]]
print(f'Rank {COMM_WORLD.rank}, coord {coords[0]} in cell {m.locate_cell(coords[0])}, coord {coords[1]} in cell {m.locate_cell(coords[1])}')
```
when run on 3 ranks we get
```
$ mpiexec -n 3 python test_parallel_point_locate.py
Rank 0, coord [0.5, 0.5] in cell 0, coord [0.9, 0.9] in cell 0
Rank 1, coord [0.5, 0.5] in cell None, coord [0.9, 0.9] in cell None
Rank 2, coord [0.5, 0.5] in cell 0, coord [0.9, 0.9] in cell 1
```
The cell boundary point `[0.5, 0.5]` has been located in the same cell as that containing the point `[0.9, 0.9]` on rank 0 and in a different cell on rank 2.

[...]

PR #2449, opened 2022-05-24
Vertex only mesh tolerance

dham:
How do we feel about letting users specify the cell inclusion tolerance when setting up a VertexOnlyMesh? This will allow users to work around particles getting lost due to roundoff?

This PR doesn't have docstrings yet. I am canvassing opinions about whether this is an OK thing to do.

### DIFF
--- a/firedrake/mesh.py
+++ b/firedrake/mesh.py
@@ -2091,7 +2091,8 @@ def ExtrudedMesh(mesh, layers, layer_height=None, extrusion_type='uniform', kern
 
 
 @PETSc.Log.EventDecorator()
-def VertexOnlyMesh(mesh, vertexcoords, missing_points_behaviour=None):
+def VertexOnlyMesh(mesh, vertexcoords, missing_points_behaviour=None,
+                   tolerance=None):
     """
     Create a vertex only mesh, immersed in a given mesh, with vertices defined
     by a list of coordinates.
@@ -2175,7 +2176,7 @@ def VertexOnlyMesh(mesh, vertexcoords, missing_points_behaviour=None):
     if pdim != gdim:
         raise ValueError(f"Mesh geometric dimension {gdim} must match point list dimension {pdim}")
 
-    swarm = _pic_swarm_in_mesh(mesh, vertexcoords)
+    swarm = _pic_swarm_in_mesh(mesh, vertexcoords, tolerance)
 
     if missing_points_behaviour:
 
@@ -2249,7 +2250,7 @@ def VertexOnlyMesh(mesh, vertexcoords, missing_points_behaviour=None):
     return vmesh
 
 
-def _pic_swarm_in_mesh(parent_mesh, coords, fields=None):
+def _pic_swarm_in_mesh(parent_mesh, coords, fields=None, tolerance=None):
     """Create a Particle In Cell (PIC) DMSwarm immersed in a Mesh
 
     This should only by used for meshes with straight edges. If not, the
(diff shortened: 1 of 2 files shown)


dham:
Also answers on a postcard about how to write a test for this. A test that only fails due to roundoff sounds really unsafe.

ReubenHill:
I think that this is the right thing to do as outlined in issue #2224. Shall we move the discussion there? I already outlined some caveats there

ReubenHill:
See discussion at https://github.com/firedrakeproject/firedrake/discussions/2450

dham:
I think this is good to go. I have set it to squash merge on approval. There aren't tests but I don't think we will improve things by putting in tests which depend sensitively on roundoff behaviour.

ReubenHill (commented):
Based on the documentation I would expect the following
```
from firedrake import *
m = UnitSquareMesh(1, 1)
coords = [[1.01, 1.01]]
vm = VertexOnlyMesh(m, coords, tolerance=0.1)
print(vm.cell_set.size)
```
to print out 1 and form the basis of a simple test of the functionality. Instead it prints 0 - what am I missing?

dham:
The point that you chose is outside the axis-aligned bounding box of any of the cells, so all the cells are eliminated by libspatialindex before local coordinates are calculated. This is fine because the axis-aligned bounding box test has no floating point arithmetic and hence no round-off error. If I move one corner of the box so that it's no longer axis-aligned then I can manufacture a case where the tolerance does something:
```python3
from firedrake import *
m = UnitSquareMesh(1, 1)
m.coordinates.dat.data[1,:] = [1.1, 1]
coords = [[1.0501, 0.5]]
vm = VertexOnlyMesh(m, coords, tolerance=0.1)
print(vm.cell_set.size) # prints 0
vm = VertexOnlyMesh(m, coords, tolerance=None)
print(vm.cell_set.size) # prints 1
```
This is also kind of abuse of the feature because the problem the tolerance fixes is really points getting lost between cells inside the domain. The problem of points that are near but outside the domain is a whole different can of worms that is beyond this fix.

ReubenHill:
Is there a way to use the feature to check for points being lost between cells inside the domain then? I note that `locate_points` seems clever enough to pick which parent cell a mesh lives and and doesn't give me `2` when I try the following
```
from firedrake import *
m = UnitSquareMesh(1, 1)
coords = [[0.5, 0.5]]
vm = VertexOnlyMesh(m, coords, tolerance=0.1)
print(vm.cell_set.size)
```

Otherwise your example above seems a reasonable way to test the feature no?

dham:
Well, it's testing a side-effect which doesn't make be supremely happy. In order to get a meaningful test inside the domain you have to chance upon a case where actual roundoff is a problem. This is hard to find and is sensitive to just about everything (hardware, compiler, compiler flags...)

ReubenHill:
I would have thought testing a side effect is better than not testing at all, no? Is the argument against that test that it could become invalidated by future, unrelated, changes?

firedrake/mesh.py
```diff
@@ -2105,6 +2106,10 @@ def VertexOnlyMesh(mesh, vertexcoords, missing_points_behaviour=None):
         that they have the same list of vertices (else the test is not
         possible): this operation scales with number of vertices and number of
         ranks.
+    :kwarg tolerance: the amount by which the local coordinates of a point are
+        allowed to fall outside the cell while still having the point count as
+        in the cell. Increase the default (1.0e-14) somewhat if vertices are
+        being lost in the :class:`VertexOnlyMesh` construction process.
```
ReubenHill, line 2112: I think we need to note that this only works for cell boundaries within the (I presume on-process) mesh boundaries. I presume it can't guard against points on the mesh boundaries?

dham: If we could guard against being on the mesh boundary, we'd fix that case too!

ReubenHill: Okay then I think it just needs to be noted that it won't work if the lost point is on the mesh boundary

dham: I've added a couple of words to the comment to make it clear that this is for points inside the domain.

PR #2458, opened 2022-05-29
Disk checkpointing

dham:
Enable disk checkpointing of the forward state in adjoint calculations. There are a few aspects to the design of this PR.

The API follows the API for annotation with calls to pause and continue disk checkpointing in the same way as pausing and continuing annotation. This is of limited utility now, since users will typically just switch on disk checkpointing and leave it. However, once revolve-style recomputation is implemented, it will be necessary.

The lifetime of the checkpoint files is attached to the lifetime of the checkpointed Functions on the tape. Each checkpoint file will be deleted once all the Functions it contains are no longer required.

The scope of the checkpointing files is the tape, and if the working tape is changed then these files change too. This is facilitated by changes in https://github.com/dolfin-adjoint/pyadjoint/pull/86 .

As well as docstrings, there is a short manual section. A demo or notebook will follow once revolve-style checkpointing follows in a further PR.

### DIFF
--- a/docs/source/checkpointing.rst
+++ b/docs/source/checkpointing.rst
@@ -199,6 +199,26 @@ with `idx` parameter always unset, and the same :class:`~.Function` can only be
 loaded using the same mode.
 
 
+Using disk checkpointing in adjoint simulations
+===============================================
+
+When adjoint annotation is active, the result of every Firedrake operation is
+stored in memory. For some simulations, this can result in a very large memory
+footprint. As an alternative, it is possible to specify that those intermediate
+results in forward evaluations of the tape which have type
+:class:`~firedrake.function.Function` be written to disk. This is usually the
+bulk of the data stored on the tape so this largely alleviates the memory
+problem, at the cost of the time taken to read to and write from disk.
+
+Having imported `firedrake_adjoint`, there are two steps required to enable
+disk checkpointing of the forward tape state.
+
+1. Call :func:`~firedrake.adjoint.checkpointing.enable_disk_checkpointing`.
+2. Wrap all mesh constructors in :func:`~firedrake.adjoint.checkpointing.checkpointable_mesh`.
+
+See the documentation of those functions for more detail.
+
+
 Checkpointing with DumbCheckpoint
 =================================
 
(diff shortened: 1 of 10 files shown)


dham:
> Another query I had is whether we still need `checkpointable_mesh` when the parallel issue of subsequent checkpoints coming back with different decompositions is fixed. Presumably if this gets fixed, the current decomposition is stored in the checkpoint somehow (through global numbering or whatever), then why wouldn't the same thing happen when an original mesh (that we have not wrapped in checkpointable_mesh) gets checkpointed for the first time? I'm probably missing something here

Unfortunately it's not quite as easy as that. You still need the mesh to have the star forests attached to it that do the redistribution in order to validly load from checkpoint onto that mesh. Those only exist for meshes that have been loaded from file.

dham:
@stephankramer was exactly right above. There were two issues which have now been fixed.

1. `_checkpoint_index` is a class attribute and I wasn't updating it properly. This caused spurious overwriting in the checkpoint file.
2. pyadjoint was overwriting the checkpoints of floating type dependencies every time they are used. This has now been fixed over in https://github.com/dolfin-adjoint/pyadjoint/pull/86.

The latter demands a little further explanation. Some overloaded types are mutable, so the primary data object can change over time. The adjoint deals with this by creating a new block_variable every time a variable appears as an output of a block, this means that the  block variables (which store the checkpoints and adjoint state) should be written to exactly once. 

A further complication is the case of floating types, which are types to which floating blocks can be applied. These blocks represent symbolic operations which create references to data types. An example is indexing into a mixed function. The complication here is that blocks which take the indexed function as an input see the value at the time that the indexed function is used, not the value at the time that the indexing occurs. This is dealt with by delaying the taping of the indexing operation until immediately before the indexed type is used. It had been thought that this also required that the checkpoint be overwritten with the value at the time of use. However, taping the block is accompanied by the creation of a new block variable for the output of the floating block,  which means that the single assignment semantics of blocks are preserved. https://github.com/dolfin-adjoint/pyadjoint/pull/86 therefore switches the dependency adding callback for floating types to not overwrite.

stephankramer (approved):
All good to me

PR #2459, opened 2022-05-29
Prettier printing of Functions and adjoint blocks.

dham:
The `__str__` method is supposed to produce human-readable output. Currently `Function` outputs strings of the form `w_{20}` instead of `w₂₀`. This is a hangover from Python 2 where unicode wasn't unique. This PR changes this.

Several adjoint blocks never had `__str__` methods, which makes visualising the tape to debug challenging. This PR now adds `__str__` methods for the blocks we are responsible for. Another PR will to pyadjoint will address the ones we inherit from there. The impact of this (and that one) is that we can create easy to understand tape visualisations such as:

image

### DIFF
--- a/firedrake/adjoint/blocks.py
+++ b/firedrake/adjoint/blocks.py
@@ -293,6 +293,9 @@ class FunctionSplitBlock(Block, Backend):
     def recompute_component(self, inputs, block_variable, idx, prepared):
         return self.backend.Function.sub(inputs[0], self.idx)
 
+    def __str__(self):
+        return f"{self.get_dependencies()[0]}[{self.idx}]"
+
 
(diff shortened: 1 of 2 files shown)


jwallwork23 (approved):
Thanks David, this is great. It would be good to add this to the firedrake-adjoint flow past a cylinder tutorial at some point.

dham:
Thanks @jwallwork23. You're right. Once I get the changes into pyadjoint I should fix our tutorials.

mc4117 (commented):
Sorry I didn't get a chance to review this before the merge but this looks great and would definitely have been helpful to me when I was starting to learn pyadjoint

PR #2461, opened 2022-06-01
WIP: Ksagiyam/plex io freeze distribution

ksagiyam:
PETSc side is probably ok, and there is a working test (https://github.com/firedrakeproject/petsc/tree/ksagiyam/plex_io_freeze_distribution).

Firedrake side is still WIP, but I now need to prepare my talk.


docs/source/checkpointing.rst
```diff
@@ -22,6 +22,24 @@ to pass through a single process.
 It also supports flexible checkpointing, where one can save meshes and
 :class:`~.Function` s on :math:`N` processes and later load them on
 :math:`P` processes.
+If :math:`P == N`, the parallel distribution and entity permutation
+(reordering) of the saved mesh is recovered on the loaded mesh by default.
+The distribution and permutation data are by default stored under names
+automatically generated by Firedrake, but it is recommended that users
```
dham, line 28: I think I don't understand this recommendation. Won't most users will be fine with the automatically generated names?

firedrake/checkpointing.py
```diff
@@ -603,9 +616,10 @@ def save_mesh(self, mesh):
             if mesh.name not in self.require_group(path):
                 path = self._path_to_mesh(tmesh.name, mesh.name)
                 self.require_group(path)
-                # Firedrake coodinates are saved here, but never loaded at the moment.
-                # We load plex coordinates instead.
-                mesh.init()
+                # Firedrake coodinates are saved here, but only loaded for periodic meshes.
+                # Otherwise, we load plex coordinates.
+                if mesh.periodic:
+                    self.set_attr(path, PREFIX_PERIODIC, True)
```
dham, line 622: This seems like the wrong test. What about another mesh with non-plex coordinates: e.g. high order?

dham (commented):
This is nice. I think it's the right thing to do.

Let's talk about the circumstances in which we should use our own coordinates in the meeting. This is coming up in several guises at the moment.

firedrake/checkpointing.py
ksagiyam: I see. We can use `mesh._made_from_coordinates` instead and it's more robust.

docs/source/checkpointing.rst
ksagiyam: Yes. Updated.

ksagiyam:
Now we directly save/load Firedrake coordinates. https://github.com/firedrakeproject/firedrake/pull/2475 is just for loading Firedrake coordinates directly (and thus fixes periodic I/O) and is independent of freezing distributions/permutations. Thanks!

docs/source/checkpointing.rst
```diff
@@ -22,6 +22,29 @@ to pass through a single process.
...
+    the user set the distribution name explicitly when constructing a mesh;
+    see, e.g., :py:func:`~.Mesh`.
+
+.. warning::
+
+   If :math:`P != N` or `P == N` but `distribution_parameters` dict and/or
+   `reorder` parameter are passed when loading a mesh, the saved mesh and
+   the loaded mesh (and thus the saved function and the loaded function)
+   will in general be represented by different global numbering systems and
+   they are merely guranteed to be analytically the same; as a consequence,
```
dham, line 43: guaranteed

dham (dismissed):

docs/source/checkpointing.rst
ksagiyam: Fixed!

JDBetteridge (approved):

PR #2463, opened 2022-06-04
Support numpydoc docstrings

dham:
Numpydoc is the docstring convention and supporting Python package used by Numpy. It allows for docstrings with less rst markup which are more human readable than the default Sphinx approach. We have discussed and, I believe, agreed to move towards using numpydoc.

This PR switches on the tooling for numpydoc in our Sphinx configuration and hence allows for new docstrings to use numpydoc conventions while producing web output which matches the rest of our API documentation. This is the first step in a gradual shift to numpydoc across our code base.

Numpydoc is documented at: https://numpydoc.readthedocs.io/en/latest/format.html

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -36,6 +36,7 @@ jobs:
           python -m pip install sphinx
           python -m pip install sphinxcontrib-bibtex
           python -m pip install bibtexparser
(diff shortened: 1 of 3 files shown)


connorjward (approved):
Looks fine.

PR #2464, opened 2022-06-04
Progress bar support

dham:
Progress bars are a succinct way of providing visual feedback about the progress of a timestepping simulation, or about the execution of the tape when using the adjoint. Progress bar support was recently added to 
pyadjoint (https://github.com/dolfin-adjoint/pyadjoint/pull/80). This PR adds a progress bar subclass configured to be suitable for many Firedrake forward and adjoint applications. As well as the docstring documentation, it adds progress bars to one forward and one adjoint notebook. 

The new code uses numpy docstrings so this PR sits on top of #2463.


dham:
The documentation build fail is spurious: it's caused by the fact that the documentation is built in the firedrake-vanilla container, which won't contain the progress package until after this PR merges.

firedrake/progress_bar.py
```diff
@@ -0,0 +1,76 @@
+from pyop2.mpi import COMM_WORLD
+from progress.bar import FillingSquaresBar
+
+
+class _NullProgressBar:
```
connorjward, line 5: Is this class needed? It's identical to the one in [pyadjoint](https://github.com/dolfin-adjoint/pyadjoint/blob/master/pyadjoint/tape.py#L467).

connorjward (commented):
This looks great!

firedrake/progress_bar.py
ReubenHill: I guess we can only use the pyadjoint equivalent if pyadjoint is a hard dependency, otherwise we need to have the same class in both places.

dham: pyadjoint is a hard dependency but I was weighing up using a private class from outside the package versus repeating a small amount of code. I came down on the side of repeating the code, but if people feel strongly the other way then we can import the class. I don't really care very much either way.

ReubenHill: I personally don't have a strong opinion either and can think of plusses and minuses to both (e.g. where one to find a bug). @connorjward ?

connorjward: I have no strong opinion either. Happy to leave as is.

PR #2468, opened 2022-06-14
Preserve the function_arg shape in homogenize

colinjcotter:
Homogenize replaces the DirichletBC function_arg by zero. If the function space is vector/tensor valued, we should have a zero vector/tensor as appropriate.

### DIFF
--- a/firedrake/bcs.py
+++ b/firedrake/bcs.py
@@ -358,7 +358,7 @@ class DirichletBC(BCBase, DirichletBCMixin):
         Set the value to zero.
 
         '''
-        self.function_arg = 0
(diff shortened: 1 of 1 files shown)


firedrake/bcs.py
```diff
@@ -358,7 +358,7 @@ def homogenize(self):
         Set the value to zero.
 
         '''
-        self.function_arg = 0
+        self.function_arg = 0*self.function_arg
```
wence-, line 361: zero(self.function_arg.ufl_shape)

wence- (changes requested):
style

dham (approved):

PR #2474, opened 2022-06-20
Fix to ASMVankaPC

ScottMacLachlan:
The logic seems off here, where we loop over `pt` in the star, but don't use it to create the list of points.  This implements the obvious fix.  @colinjcotter has a use case that this fixes.


colinjcotter (approved):
I'm happy with this.

PR #2483, opened 2022-07-05
Update Docker Ubuntu and Python versions

dham:
Move the docker containers to the current Ubuntu LTS. Also remove the rather old Python 3.8 pin from the docker containers.

Due to the nature of changing the docker containers, the tests here are not reliable. #2482 provides the actual testing of this functionality.


JDBetteridge:
https://github.com/firedrakeproject/firedrake/pull/2482 is now passing, we can merge this

Issue #2485, opened 2022-07-05
Interface error during extension for substructuring DD preconditioners


nabw:
Hi firedrakers, 

I have been working with my colleagues on an extension of Firedrake for the BDDC preconditioner, for which MATIS matrices are required. For this, the allocation procedure in PyOP2 has been changed into the following: 

`
        if self.mat_type == 'is':
            mat.createIS(((self.nrows, None), (self.ncols, None)),
                         lgmapr=rlgmap, lgmapc=clgmap,
                         comm=self.comm)
            mat.setISPreallocation(self.sparsity.nnz, self.sparsity.onnz)
            mat.setUp()
        else:
            mat.createAIJ(size=((self.nrows, None), (self.ncols, None)),
                         nnz=(self.sparsity.nnz, self.sparsity.onnz),
                         bsize=1,
                         comm=self.comm)
            mat.setLGMap(rmap=rlgmap, cmap=clgmap)
        self.handle = mat
`

in both init_monolithic() and init_block() functions. Although the code seems to yield a correct type and allocation, some things look strange. For instance, I have modified the Helmholtz demo as a test into the following: 

`
from firedrake import *
mesh = UnitSquareMesh(10, 10)
V = FunctionSpace(mesh, "CG", 1)
u = TrialFunction(V)
v = TestFunction(V)
f = Function(V)
x, y = SpatialCoordinate(mesh)
f.interpolate((1+8*pi*pi)*cos(x*pi*2)*cos(y*pi*2))
a = (inner(grad(u), grad(v)) + inner(u, v)) * dx
L = inner(f, v) * dx
sol = Function(V)

solve(a == L, sol, solver_parameters={'ksp_type': 'cg', 'pc_type': 'none', 'ksp_monitor':None, 'log_view':None, 'mat_type':'is'})
`

which gives the following error (PETSc compiled with debug):

`  
  File "PETSc/Log.pyx", line 115, in petsc4py.PETSc.Log.EventDecorator.decorator.wrapped_func
  File "PETSc/Log.pyx", line 116, in petsc4py.PETSc.Log.EventDecorator.decorator.wrapped_func
  File "/home/barnafi/firedrake/src/firedrake/firedrake/adjoint/variational_solver.py", line 90, in wrapper
    out = solve(self, **kwargs)
  File "/home/barnafi/firedrake/src/firedrake/firedrake/variational_solver.py", line 278, in solve
    self.snes.solve(None, work)
[...]

PR #2487, opened 2022-07-07
Only pass lgmaps to assembly parloops if there are boundary conditions

connorjward:
This bug came up in #2485.

My understanding of BCs is a bit ropey but I think this is a safe change to introduce. Previously what would happen is inside the parloop we would always replace the lgmaps for the matrices even if there were no boundary conditions, in which case the old and new lgmaps would be the same.


wence-:
I _think_ this is right, but as you note, this part of the code is quite delicate.

dham (dismissed):

dham (approved):

PR #2494, opened 2022-07-14
Sv/fix gusto

sv2518:


tommbendall (approved):
Thanks so much for this! I don't 100% understand the context of the code changes but I've run all the gusto tests on my machine and they are all passing, so this fix works for me

PR #2496, opened 2022-07-16
install: work around build failure caused by pybind11==2.10.0

ksagiyam:
Fix build failure caused by `pybind11==2.10.0`; see https://github.com/firedrakeproject/firedrake/runs/7370283417?check_suite_focus=true.

### DIFF
--- a/scripts/firedrake-install
+++ b/scripts/firedrake-install
@@ -1698,7 +1698,9 @@ if mode == "install":
 
     # Force Cython to install first to work around pip dependency issues.
     run_pip_install(["Cython>=0.22"])
-    run_pip_install(["pybind11"])
+    # See build failure caused by pybind11==2.10.0 at https://github.com/firedrakeproject/firedrake/runs/7370283417?check_suite_focus=true.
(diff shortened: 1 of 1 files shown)


scripts/firedrake-install
```diff
@@ -1698,7 +1698,9 @@ if mode == "install":
 
     # Force Cython to install first to work around pip dependency issues.
     run_pip_install(["Cython>=0.22"])
-    run_pip_install(["pybind11"])
+    # See build failure caused by pybind11==2.10.0 at https://github.com/firedrakeproject/firedrake/runs/7370283417?check_suite_focus=true.
+    print('We use pybind11==2.9.2 until the issue with pybind11==2.10.0 is resolved.')
```
dham, line 1702: should be logged rather than printed, no?

dham (changes requested):

scripts/firedrake-install
ksagiyam: Fixed!

dham (approved):

PR #2502, opened 2022-07-25
Fix Vec's state counter for MixedDat

nbouziani:
This PR is associated to the following PyOP2 PR: https://github.com/OP2/PyOP2/pull/670

It fixes the following test: `/tests/multigrid/test_nested_split.py`

```
assert norm(assemble(s_expect - s)) < 1e-10
       AssertionError: assert 2.6886334289236815e-10 < 1e-10
```

This error is caused by the PETSc Vec object of `MixedDat` not being updated in the `vec_context`. The current object versioning mechanism relies on the Vecs of the underlying `Dat` objects associated to `MixedDat`. However, this is not enough and we also need to increase the state counter of the Vec in the mixed space when it is populated.

This causes the above test to fail as PETSc Vecs cache some operations (e.g. norm computations) based on the Vec's state counter which can lead to wrong numerical values.

### DIFF
--- a/scripts/firedrake-install
+++ b/scripts/firedrake-install
@@ -29,7 +29,7 @@ parallel_packages = ["mpi4py", "h5py", "petsc4py", "slepc", "slepc4py", "PyOP2",
 # Firedrake application installation shortcuts.
 firedrake_apps = {
     "gusto": ("""Atmospheric dynamical core library. http://firedrakeproject.org/gusto""",
-              "git+ssh://github.com/firedrakeproject/gusto#egg=gusto"),
+              "git+ssh://github.com/firedrakeproject/gusto.git@main#egg=gusto"),
     "thetis": ("""Coastal ocean model. http://thetisproject.org""",
                "git+ssh://github.com/thetisproject/thetis#egg=thetis"),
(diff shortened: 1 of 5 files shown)


wence-:
This should not be merged, because it's just testing the pyop2 change, no?

nbouziani:
> This should not be merged, because it's just testing the pyop2 change, no?

Yes! I forgot to set it as draft PR, thanks!

scripts/firedrake-install
```diff
@@ -29,7 +29,7 @@ parallel_packages = ["mpi4py", "h5py", "petsc4py", "slepc", "slepc4py", "PyOP2",
 # Firedrake application installation shortcuts.
 firedrake_apps = {
     "gusto": ("""Atmospheric dynamical core library. http://firedrakeproject.org/gusto""",
-              "git+ssh://github.com/firedrakeproject/gusto#egg=gusto"),
+              "git+ssh://github.com/firedrakeproject/gusto.git@main#egg=gusto"),
```
nbouziani, line 32: NB: I have rebased on top of #2503 as building firedrake is currently broken.

PR #2505, opened 2022-07-27
Parallel adjoint disk checkpoint

dham:
Once #2461 lands, this enables the test of parallel disk checkpointing.


dham:
Superceded.

PR #2523, opened 2022-08-12
Fix hashing for Assign rvalues

connorjward:
When I first implemented disk caching for `Assign` I used `Assign.slow_key` to determine whether we had a cache hit or not. However, it appears as though `slow_key` was the wrong choice as trivial code such as

```py
u.assign(Constant(1))
u.assign(Constant(1))
```

would result in a cache miss, and hence a write to disk. The only way I found to avoid it was to write:

```py
rhs = Constant(1)
u.assign(rhs)
u.assign(rhs)
```

I'm now using the same hashing function as I do for interpolation and it seems to have resolved the problem. I am also now only caching in memory rather than to disk since the codegen going on here is not very expensive (but still worth caching).

Many thanks to @dorugeber for flagging this.

### DIFF
--- a/firedrake/assemble_expressions.py
+++ b/firedrake/assemble_expressions.py
@@ -1,6 +1,4 @@
 import itertools
-import os
-import tempfile
 import weakref
 from collections import OrderedDict, defaultdict
 from functools import singledispatch
@@ -13,7 +11,7 @@ from gem.impero_utils import compile_gem, preprocess_gem
 from gem.node import MemoizerArg
 from gem.node import traversal as gem_traversal
 from pyop2 import op2
-from pyop2.caching import disk_cached
+from pyop2.caching import cached
 from pyop2.parloop import GlobalLegacyArg, DatLegacyArg
 from tsfc import ufl2gem
 from tsfc.loopy import generate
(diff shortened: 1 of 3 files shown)


firedrake/assemble_expressions.py
```diff
@@ -311,10 +294,9 @@ def fast_key(self):
 
     @cached_property
     def slow_key(self):
-        """A slow lookup key for this expression (relabelling UFL indices)."""
-        self.relabeller._reset()
-        rvalue, = map_expr_dags(self.relabeller, [self.rvalue])
-        return (type(self), hash(self.lvalue), hash(rvalue))
+        """A slow lookup key for this expression."""
+        from firedrake.interpolation import hash_expr
+        return (type(self), hash(self.lvalue), hash_expr(self.rvalue))
```
wence-, line 299: probably hash(lvalue) is not stable for two coefficients that would be disk cache identical. I never considered using this for an on disk cache, only in memory

connorjward: To be perfectly honest I think a disk cache here is probably overkill anyway. I could easily switch to using only an in-memory cache here. WDYT?

wence-: Sure, I forget exactly how the cache-lookup proceeds but I vaguely recall that the necessary thing is that the spaces match (this hashes on the coefficient in the left hand side of the assign rather than the space)

connorjward:
Marking as draft as I've thought of a potential issue with what I've done.

dorugeber:
> Marking as draft as I've thought of a potential issue with what I've done.

oh no...

connorjward:
> > Marking as draft as I've thought of a potential issue with what I've done.
> 
> oh no...

I think it's now fixed. Once we pass tests LGTM.

ksagiyam (approved):
Looks good to me.

connorjward:
Thanks Koki!

PR #2525, opened 2022-08-18
Update clean function in firedrake-install for PEP 518.

francesco-ballarin:
I am updating `FEMlium` to use PEP 518 (installation via `setup.cfg` and `pyproject.toml`, complete removal of `requirements.txt` and `setup.py`). While testing my branch to make sure that it doesn't break the `firedrake-install` and `firedrake-update` scripts, I have realized that the `clean` function defined therein needs to be updated for the PEP 518 case for the `firedrake-update` script to run without errors. Indeed, it tries to run `python setup.py clean` which would clearly raise an error if `setup.py` is missing.

This PR takes this new case into account. There are currently two commits in this PR:
* https://github.com/firedrakeproject/firedrake/pull/2525/commits/c7a4e1fa6cfaac494db8e0ba1f4caa4c8ac920dd is the only relevant commit, which changes the implementation of the `clean` function;
* https://github.com/firedrakeproject/firedrake/pull/2525/commits/f631482323b82b4b51d938281da9ea8848fa226b is a *temporary* commit that enables docker build (but not push) on this branch as well. Indeed, we must test `firedrake-update` with apps. Such script, to my understanding, is only called at the `Build full Firedrake Docker container` stage, which is skipped on non-`master` branches.

To be extra safe, we will need four CI runs:
1. [x] the first run will happen before I merge my branch, to double check that this PR does not break the existing `firedrake-install` or `firedrake-update` scripts. In the first run, I expect to see in the CI output something like
```
[...]
Installing pip dependencies for FEMlium
Installing pip dependencies for [name of the next package]
[...]
```
2. [x] after a successful first run, I will merge my branch. I will *temporarily revert* https://github.com/firedrakeproject/firedrake/pull/2525/commits/c7a4e1fa6cfaac494db8e0ba1f4caa4c8ac920dd and I will then ask you to trigger again a CI run. I expect this run to fail while cleaning `FEMlium`, due to the fact that `setup.py` is missing.
3. [ ] after a failing second run, I will delete my temporary revert, getting back to the configuration at 1. I will then ask you to trigger again a CI run, which I expect to succeed. The output should show something like the following, because I do not have requirements.txt anymore. 
```
[...]
Installing pip dependencies for FEMlium
No dependencies found. Skipping.
Installing pip dependencies for [name of the next package]
[...]
```
4. [ ] after a successful third run, I will delete https://github.com/firedrakeproject/firedrake/pull/2525/commits/f631482323b82b4b51d938281da9ea8848fa226b, resulting in a PR with only https://github.com/firedrakeproject/firedrake/pull/2525/commits/c7a4e1fa6cfaac494db8e0ba1f4caa4c8ac920dd , which should be ready to merge at that point.

### DIFF
--- a/scripts/firedrake-install
+++ b/scripts/firedrake-install
@@ -982,7 +982,11 @@ def clean(package):
     with directory(package):
         with environment(**blas):
             # loopy needs knowledge of the system BLAS just to run setup.py
-            check_call([python, "setup.py", "clean"])
+            if os.path.isfile("setup.py"):
+                check_call([python, "setup.py", "clean"])
+            else:  # PEP 518 build system
+                if os.path.exists("build") and os.path.isdir("build"):
+                    shutil.rmtree("build")
 
 
 def pip_uninstall(package):


francesco-ballarin:
FYI, this is related to a recent discussion on slack with @JDBetteridge

scripts/firedrake-install
```diff
@@ -982,7 +982,10 @@ def clean(package):
     with directory(package):
         with environment(**blas):
             # loopy needs knowledge of the system BLAS just to run setup.py
-            check_call([python, "setup.py", "clean"])
+            if os.path.isfile("setup.py"):
+                check_call([python, "setup.py", "clean"])
+            else:  # PEP 518 build system
+                check_call(["rm", "-rf", "build"])
```
JDBetteridge, line 988: I'm happy wit the first bit of logic here, if there's no `setup.py` we shouldn't try to call it. But what happens here if the build directory doesn't exist? (for instance if the installation of another package fails during firedrake-install, firedrake-update is used to resume the installation) There should probably be a try except block around the deletion of the build directory so the update doesn't fail here.

I think I'd also be a bit happier if we were using Python functionality to remove directories, it seems safer than subprocess calling `rm -rf` :grimacing: . I think `shutil.rmtree()` does what you want in this instance.

JDBetteridge (changes requested):

scripts/firedrake-install
francesco-ballarin: Done, and force pushed.

JDBetteridge:
I've realised you can't use our Docker secrets, I have created a draft PR here so you can see the build status:
https://github.com/firedrakeproject/firedrake/pull/2527

francesco-ballarin:
First CI run in #2527 was successful.

I have just merged the update to FEMlium build system. I can't push to #2527: can you carry out in #2527 for me the second CI run (which should fail!) as detailed in the opening post?

JDBetteridge:
Sorry, I got side tracked. I have reverted that commit and will un-revert that change on successful failure.

francesco-ballarin:
OK, so run 2 failed as expected.

Run 3 failed too, and it shouldn't have. After having a closer look, I am afraid that we cannot do a proper test of run 3 without pushing to docker hub. In fact, in https://github.com/firedrakeproject/firedrake/blob/master/docker/Dockerfile.firedrake , the base image is `FROM firedrakeproject/firedrake-vanilla:latest`. The `firedrake-vanilla` image built on the CI run will contain the fixed `firedrake-update` script, but if we don't push it we'll never test it at the next steps.

I'll leave @JDBetteridge to decide if this is a good way to go. Do consider that now CI on master is broken until we merge this ;)

francesco-ballarin:
In the meantime I have removed the temporary commit (which is still available in #2527, and will probably need to be modified there), so that this PR is ready to merge when we sort out how to do the third CI run.

JDBetteridge:
I think if this all passes I will merge and when the master branch CI gets run the docker containers will all get regenerated. Master failing on the Docker image isn't ideal, but isn't then end of the world.

francesco-ballarin:
@JDBetteridge  Otherwise, if you don't want to risk breaking (the already broken) `master` build, you can push to #2527 replacing all `push: false` with `push: true`. That should be the 3rd run I had envisioned and (apart from my removals of `if: ${{ github.ref == 'refs/heads/master' }}`, to have a docker build on a PR)  should effectively be what will happen on `master` on merge.

JDBetteridge:
Given master is already broken and this PR should fix it, I think I'm happy to just merge it. Worst case scenario we just have to revert, but I don't think it's likely in this situation.

JDBetteridge (approved):

PR #2543, opened 2022-09-02
Anaconda installation

danshapero:
This patch adds a special script to install Firedrake with Anaconda. It's mostly lifted from [work](https://github.com/dham/jhub-firedrake) that @dham and @tmbgreaves did. The really important part is excluding MPICH and HDF5 from the conda environment in order to avoid colliding the with the versions that PETSc will build. There's also some hackery around creating custom activate and deactivate scripts.

I've also updated the documentation to describe the new state of affairs.

A few things that could go in this PR that I haven't done yet:

* Add building conda to the Firedrake test suite. We could probably get away with testing that it builds at all and running only a minimal part of the test suite, e.g. Poisson, Stokes, mixed Poisson, and advection, rather than the entire thing.
* Do some platform introspection to see if the user has conda or [mamba](https://mamba.readthedocs.io/).
* Forward any extra options to the Anaconda install script through to its call to firedrake-install.
* Check that the activate and deactivate scripts do exactly what they're supposed to. For example, on deactivation, Tim had this [bit of code](https://github.com/dham/jhub-firedrake/blob/a7f18d0fcdb9f70e33e52f5ee2067d1cc35e23b3/Dockerfile#L40) to remove the Firedrake venv/bin from the user's PATH on deactivation. I haven't added that in yet because the paths are specific to that Docker image and we'd need a general way to do this.
* Give the user options about where to put the conda environment; currently the only option is `~/.conda` so that the directory is actually writable. Depending on how you installed Anaconda, the default directory for where to put environments can be different and it's not obvious to me what the right place is.


scripts/firedrake-install-conda.sh
```diff
@@ -0,0 +1,33 @@
+#!/usr/bin/env bash
```
dham, line 1: we don't usually use the .sh suffix

scripts/firedrake-install-conda.sh
```diff
@@ -0,0 +1,33 @@
...
+
+# For some odd reason, on MacOS we need to tell it where to find `ar`,
+# otherwise it might try to find an `llvm-ar` that doesn't necessarily exist.
+if [[ $(uname -s) == Darwin ]]; then
+    export AR=/usr/bin/ar
+fi
+
+# Make an environment variable for the directory where the conda environment
+# will live; this will depend on whether or not you're using `pyenv`.
+if [[ $(command -v pyenv) ]]; then
```
dham, line 11: Does this test if we're using pyenv or just if it's installed?

scripts/firedrake-install-conda.sh
```diff
@@ -0,0 +1,33 @@
...
+    export DIRECTORY=${HOME}/.conda/envs/firedrake
+fi
+
+# Make a conda environment that excludes HDF5 & mpich (PETSc will build these)
+conda list --explicit | grep -v mpich | grep -v hdf5 | grep -v h5py | grep -v asn1crypto > spec-file.txt
+conda create --prefix ${DIRECTORY} --file spec-file.txt
+
+# Install Firedrake into a subdirectory of the conda env
+cd ${DIRECTORY}
+curl -O https://raw.githubusercontent.com/firedrakeproject/firedrake/master/scripts/firedrake-install
```
dham, line 23: we should probably allow the script to optionally be provided with the path to firedrake-install rather than downloading it. After all, the user might want firedrake-install from a branch or fork.

dham (changes requested):
Thanks. I think there are just small changes needed here.

scripts/firedrake-install-conda.sh
danshapero: It's testing if calling `pyenv` returns a successful exit code, which to me implies that it's actually being used. It would be different if we checked for the existence of the directory `~/.pyenv`, which could just be leftover from a previous installation.

In any case, I think I see where you're going. I could add a `--prefix` argument to the script so you can decide where the conda environment will get built. If you don't give an argument for this, it'll default to one directory if you're using pyenv and another if you're not. Is that what you were thinking?

docs/source/download.rst
```diff
@@ -151,35 +151,30 @@ Windows Subsystem for Linux. There are more detailed
...
 Installation on previous versions of Windows is unlikely to work.
 
+
+Anaconda
+~~~~~~~~
+
+.. warning::
+    The following is recent as of Sept 2022 and still highly experimental.
+
+Firedrake was incompatible with the Anaconda Python distribution for a
```
inducer, line 161: Would this work with https://conda-forge.org/, specifically their https://github.com/conda-forge/miniforge installer? (That's what I use to ship all my stuff.)

Conda-forge has MPI and PETSc precompiled, which may be convenient for shipping to laptops without forcing folks to wait through the PETSc compile.

https://github.com/conda-forge/petsc-feedstock

dham: This hits the usual problem with prebuilt PETSc, which is that the conda-forge PETSc isn't built with all the configure options that Firedrake depends on.

scripts/firedrake-install-conda.sh
dham: I think that that's a better solution. pyenv returning successfully just means, I think, that pyenv is installed. It doesn't imply that the user intended to use it on this occasion.

docs/source/download.rst
inducer: Conda (and conda forge) support build variants, you could depend on a specific one.

dham: So there's a variant which has what we need? I think I'm not getting this.

inducer: Maybe? The point is that conda-forge is a community distribution, and a build variant that meets your specific needs is potentially one pull request to https://github.com/conda-forge/petsc-feedstock away.

dham: Ah. Ok. Well once we get through this postdoc hiring round we officially have the funded resource to get better at being installed so this is a serious option.

scripts/firedrake-install-conda.sh
danshapero: I can remove the .sh suffix. But the linter is assuming this is a Python script, which it isn't. I think this makes the most sense as a shell script but can rewrite it in Python if need be (there's a [Conda Python API](https://conda.io/projects/conda/en/latest/api/index.html)).

scripts/firedrake-install-conda
```diff
@@ -0,0 +1,48 @@
...
+source activate $prefix
+python $script --no-package-manager --venv-name=$prefix/firedrake-venv
+
+# Make an activate script
+mkdir -p $prefix/etc/conda/activate.d
+cp $prefix/firedrake-venv/bin/activate $prefix/etc/conda/activate.d/01_firedrake-venv.sh
+
+# Make a custom deactivate script
+mkdir -p $prefix/etc/conda/deactivate.d
+echo deactivate > $prefix/etc/conda/deactivate.d/01_firedrake-venv.sh
```
JDBetteridge, line 48: Currently if this script terminates with an error no error is raised in the terminal. This means when we hook it up to CI the install won't fail

docs/source/download.rst
```diff
@@ -151,35 +151,30 @@ Windows Subsystem for Linux. There are more detailed
...
+long time. This limitation has been lifted, but for the time being,
+installing with Anaconda will require either using this special
+`shell script <https://raw.githubusercontent.com/firedrakeproject/firedrake/master/scripts/firedrake-install-conda>`_
+or executing the commands in that script manually. If installation under
+Anaconda with that script fails, please
+`raise an issue <https://github.com/firedrakeproject/firedrake/issues>`_
+and we will try to fix it as soon as possible, but please bear with us
+as we iron out the wrinkles on supporting Anaconda.
+
+
```
JDBetteridge, line 171: Needs some instructions on how to activate conda environments

JDBetteridge (changes requested):

JDBetteridge:
I'm in the process of trying to hook this up to the Firedrake CI ([here](https://github.com/firedrakeproject/firedrake/pull/2580))

JDBetteridge:
When I do `conda activate firedrake` I see multiple "environments" active, is this expected? I.e: my propmt looks like this:

```
(firedrake-venv) (firedrake) (firedrake-venv) $
```

JDBetteridge:
I also can't run `conda install pytest-cov pytest-timeout pytest-xdist` without Conda trying to reinstall hdf5, is there a way around this?

scripts/firedrake-install-conda
danshapero: I added `set -euxo pipefail` at the start as per [this answer](https://stackoverflow.com/a/2871034/) to catch this for now. If you want more specific error handling we can revisit.

ReHoss:
> I also can't run `conda install pytest-cov pytest-timeout pytest-xdist` without Conda trying to reinstall hdf5, is there a way around this?

Could you try with pip install within the conda env ?

dham:
This turned out to be much harder than anticipated and doesn't seem likely to land.

PR #2548, opened 2022-09-05
Copy only the linear algebra solver options for the adjoint solver

salazardetroya:
If I want to add options such as `post_jacobian_callback` to the `NonlinearVariationalSolver` as in:
```
import firedrake as fd
import firedrake_adjoint as fda
from firedrake import grad, inner, dx

mesh = fd.UnitSquareMesh(10, 10)
V = fd.FunctionSpace(mesh, "Lagrange", 1)

f = fd.Constant(1.0)

u = fd.Function(V)
v = fd.TestFunction(V)
bc = fd.DirichletBC(V, fd.Constant(1), "on_boundary")

F = f * inner(grad(u), grad(v)) * dx + u**2 * v * dx - f * v * dx
problem = fd.NonlinearVariationalProblem(F, u, bcs=bc)


def callback(X, J):
    return


solver = fd.NonlinearVariationalSolver(problem, post_jacobian_callback=callback)
solver.solve()
J = fd.assemble(u**2 * dx)
c = fda.Control(f)
Jhat = fda.ReducedFunctional(J, c)
print(Jhat.derivative().dat.data)
```
the adjoint solver crashes with
```
RuntimeError: Illegal keyword argument 'post_jacobian_callback'; valid keywords are 'bcs', 'solver_parameters', 'nullspace', 'transpose_nullspace', 'near_nullspace', 'options_prefix'
```
This PR fixes that.

### DIFF
--- a/firedrake/adjoint/blocks.py
+++ b/firedrake/adjoint/blocks.py
@@ -55,7 +55,19 @@ def solve_init_params(self, args, kwargs, varform):
         self.adj_args = self.forward_args
 
     if len(self.adj_kwargs) <= 0:
-        self.adj_kwargs = self.forward_kwargs.copy()
+        la_kwargs = [
+            "bcs",
+            "solver_parameters",
+            "nullspace",
+            "transpose_nullspace",
+            "near_nullspace",
+            "options_prefix",
+        ]
+        self.adj_kwargs = dict(
+            (key, value)
+            for key, value in self.forward_kwargs.items()
+            if key in la_kwargs
+        )
 
         if varform:
             if "J" in self.forward_kwargs:


firedrake/adjoint/blocks.py
```diff
@@ -55,7 +55,19 @@ def solve_init_params(self, args, kwargs, varform):
...
+            "nullspace",
+            "transpose_nullspace",
+            "near_nullspace",
+            "options_prefix",
+        ]
+        self.adj_kwargs = dict(
+            (key, value)
+            for key, value in self.forward_kwargs.items()
+            if key in la_kwargs
+        )
```
wence-, line 70: Is this the right fix? The correct adjoint should in some sense do something if a `post_jacobian_callback` is specified. Though I don't think we can determine what that is automatically.

salazardetroya: What is the typical use of these callbacks within the `NonlinearVariationalSolver`? If they are used to implement some type of continuation strategy to ensure convergence, the adjoint solver will not need them as long as it has access to the final state of the variational form.

wence-: Sometimes they're used to modify the jacobian itself (with some thing that isn't expressible via UFL). But it's a slightly niche use case, so arguably one could just warn.

salazardetroya: Something like that?

Issue #2556, opened 2022-09-15
`assign` should only be used for weighted sums


connorjward:
This was discussed in the [Firedrake meeting 14/09/22](https://github.com/firedrakeproject/firedrake/wiki/Firedrake-meeting-2022-09-14#cw-assign-performance-issues-worth-fixing).

According to @dham, `assign` and `interpolate` are really the same thing except `assign` is less mathematically valid and has somewhat undefined behaviour in the adjoint. The only acceptable use case of `assign` should be for doing weighted sums (useful in RK timesteppers for example).

Someone should at some point do the following:
- Restrict `assign` to only work for weighted sums.
- Use the numpy interface to `Dats` for this. Code generation is overkill and produces substantial overhead.

wence-:
for linear combinations you may want vecmaxpy (which avoids multiple temporarys)

connorjward:
This has now be done.

PR #2562, opened 2022-09-22
Only permit scalars and weighted sums for `assign`

connorjward:
Associated issue: #2556

We also need to merge changes in pyadjoint: https://github.com/dolfin-adjoint/pyadjoint/pull/93.


firedrake/assign.py
```diff
@@ -0,0 +1,130 @@
...
+
+
+class WeightedCoefficientCollector(MultiFunction):
+    def __init__(self):
+        super().__init__()
+        self.weighted_coefficients = []
+
+    def sum(self, o):
+        for op in o.ufl_operands:
+            self(op)
```
wence-, line 22: Don't hand traverse, instead write:
```
def sum(self, o, *operands):
    pass
```

Hand-traversal destroys the DAG-based visitor pattern with the result that you will run slower than you want.

firedrake/assign.py
```diff
@@ -0,0 +1,130 @@
...
+            weight, = filter(
+                lambda op: isinstance(op, (RealValue, Zero)), o.ufl_operands
+            )
+            coeff, = filter(lambda op: isinstance(op, ufl.Coefficient), o.ufl_operands)
+        except ValueError:
+            raise ValueError("Only products of scalars and coefficients allowed")
+        self.weighted_coefficients.append((weight, coeff))
+
+    def coefficient(self, o):
+        self.weighted_coefficients.append((1, o))
```
wence-, line 35: Prefer to return values and glue them together coming back up the traversal rather than side-effectful modification of the visitor. That way you will be able to get away with only constructing a single `WeightedCoefficientCollector` (rather than one for every expression).

firedrake/assign.py
```diff
@@ -0,0 +1,130 @@
...
+
+    def invalid(self, o):
+        raise ValueError(f"{o} is not a valid component of a weighted sum of coefficients")
```
wence-, line 38: Does this mean that `f.assign(as_tensor([[1, 2], [3, 4]]) @ as_vector([a, b]))` will be rejected? That is a linear combination so should be safe.

firedrake/assign.py
```diff
@@ -0,0 +1,130 @@
...
+
+    @cached_property
+    def weights(self):
+        return np.array(self._weighted_coefficients[0], dtype=ScalarType)
+
+    @cached_property
+    def coefficients(self):
+        return self._weighted_coefficients[1]
+
+    @cached_property
```
wence-, line 83: This caching is unsafe if someone ever holds onto an `Assigner` object and runs it in a loop where the rvalue coefficients are modified elsewise.

firedrake/assign.py
```diff
@@ -0,0 +1,130 @@
...
+class IAddAssigner(AugmentedAssigner):
+    symbol = "+="
+
+    def assign(self):
+        if self._subset:
+            indices = self._subset.indices
+            self._assignee.dat.data[indices] += np.dot(self.weights, self.coefficient_data[indices])
+        else:
+            # TODO Explain why maxpy preferred to numpy
+            self._assignee.dat._vec.maxpy(self.weights, self.coefficient_vecs)
```
wence-, line 116: Careful, this is unsafe in the case that the lvalue or rvalues are mixed dats (the `_vec` is not a view into the data in that case)

firedrake/assign.py
connorjward: Running this gives me an `unsupported operand type(s) for @` error. Are you sure that UFL supports this sort of thing?

wence-: Ah ok, I see.

connorjward:
A quick experiment shows that this can speed things up quite a bit. I've been running the [DG advection](https://firedrakeproject.org/demos/DG_advection.py.html) demo on my branch vs `master`. Below are the flamegraphs I get on my laptop (with a hot cache). The first shows current `master` and the bottom one is my code. I found the code took ~25s on `master` and mine took ~20s. You can see that the `assign` bar is much smaller in the bottom one.

old1
new1

I have a few comments:
- We still need to apply a tree visitor to the expression. Caching on a persistent expression would let us avoid this.
- Alternatively, using a persistent `Assigner` object makes pretty much all overhead go away. The flamegraph below, where I tried this, `assign` takes 0.2s out of a total 18s.

persistent_assigner

firedrake/assemble.py
```diff
@@ -355,6 +372,7 @@ def _make_tensor(form, bcs, *, diagonal, mat_type, sub_mat_type, appctx,
         raise AssertionError
 
 
+# FIXME: This class is not annotated for the adjoint so taping will not work as expected.
```
JDBetteridge, line 375: This FIXME probably needs addressing before merge

firedrake/assign.py
```diff
@@ -0,0 +1,251 @@
...
+        constant to be added (e.g. ``u.assign(2*v + Constant(3))``). Therefore we only
+        compress to a scalar when we know it is required (e.g. inside a product with a
+        :class:`firedrake.Function`).
+        """
+        return pytools.one(
+            functools.reduce(operator.add, (c.dat.data_ro*w for c, w in weighted_coefficients))
+        )
+
+
+# FIXME: This class is not annotated for the adjoint so taping will not work as expected.
```
JDBetteridge, line 117: Fix this?

firedrake/function.py
```diff
@@ -393,108 +383,34 @@ def assign(self, expr, subset=None):
...
-            return self
-        elif (isinstance(expr, Function)
-              and expr.function_space() == self.function_space()):
-            expr.dat.copy(self.dat, subset=subset)
-            return self
-
-        from firedrake import assemble_expressions
-        assemble_expressions.evaluate_expression(
-            assemble_expressions.Assign(self, expr), subset)
+        from firedrake.assign import Assigner
```
JDBetteridge, line 386: Is there a reason we can't `from firedrake import assign` at the top and `assign.Assigner.assign()`? Or just import the things we need from `assign.py`

firedrake/utils.py
```diff
@@ -95,3 +95,22 @@ def tuplify(item):
     if not isinstance(item, dict):
         raise ValueError(f"tuplify does not know how to handle objects of type {type(item)}")
     return tuple((k, tuplify(item[k])) for k in sorted(item))
+
+
+def split_by(condition, items):
```
JDBetteridge, line 100: This can be achieved more efficiently using `itertools`, see the recipe for `partition`, something like:
```python
import itertools
def itersplit(pred, iterable):
    t1, t2 = itertools.tee(iterable)
    return tuple(filter(pred, t1)), tuple(itertools.filterfalse(pred, t2))
```

Shamelessly stolen from: https://docs.python.org/3/library/itertools.html#itertools-recipes

JDBetteridge (changes requested):
I think we are mainly just waiting on some things to be fixed for adjoint

.github/workflows/build.yml
```diff
@@ -42,7 +42,7 @@ jobs:
       - name: Build Firedrake
         run: |
           cd ..
-          ./firedrake/scripts/firedrake-install $COMPLEX --venv-name build --tinyasm --disable-ssh --minimal-petsc --slepc --documentation-dependencies --install thetis --install gusto --install icepack --install irksome --install femlium --no-package-manager|| (cat firedrake-install.log && /bin/false)
+          ./firedrake/scripts/firedrake-install $COMPLEX --venv-name build --tinyasm --disable-ssh --minimal-petsc --slepc --documentation-dependencies --install thetis --install gusto --install icepack --install irksome --install femlium --no-package-manager --package-branch pyadjoint connorjward/assign-changes || (cat firedrake-install.log && /bin/false)
```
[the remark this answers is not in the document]
connorjward: We are.

firedrake/function.py
connorjward: Because `firedrake/assign.py` has `from firedrake.function import Function` at the top (import loop).

firedrake/assemble.py
connorjward: This is essentially just a documentation fix to go with https://github.com/firedrakeproject/firedrake/issues/2581. I thought it would do no harm to merge alongside this PR.

firedrake/assign.py
connorjward: I argue that this is a separate issue to resolve. Current adjoint approaches will still work fine. I've just added an extra entry point for `assign` that lacks the functionality.

dham (changes requested):

firedrake/utils.py
connorjward: I'd like to claim that my solution works fine and so this isn't a big deal.

JDBetteridge (dismissed):

dham (dismissed):

ReubenHill (changes requested):
Please check that docs are up to date

ReubenHill:
> Please check that docs are up to date

Poke me to approve when done

connorjward:
@ReubenHill I think the docs all look fine. This is ready to go.

tests/vertexonly/test_poisson_inverse_conductivity.py
```diff
@@ -29,12 +29,6 @@ def test_poisson_inverse_conductivity():
     # Have to import inside test to make sure cleanup fixtures work as intended
     from firedrake_adjoint import Control, ReducedFunctional, minimize
 
-    # Manually set up annotation since test suite may have stopped it
```
ReubenHill, line 32: Quick query before I approve, why was this changed?

connorjward: With [this line](https://github.com/firedrakeproject/firedrake/blob/master/tests/vertexonly/test_poisson_inverse_conductivity.py#L18) above it in the file we were effectively switching on the adjoint twice. Since we use a [counter](https://github.com/dolfin-adjoint/pyadjoint/blob/master/pyadjoint/tape.py#L31) to track whether or not the adjoint is enabled this meant that switching the adjoint off with `with stop_annotating():` only incremented the counter once and we still thought that we were supposed to be taping.

This was causing tests to fail because I now `raise NotImplementedError` if someone is inside either `FormAssembler.assemble` or `Assigner.assign` and taping is switched on.

ReubenHill: But there is no `with stop_annotating():` here? Or am I being daft?

connorjward: We call `with stop_annotating():` inside the adjoint (e.g. [here](https://github.com/firedrakeproject/firedrake/blob/master/firedrake/adjoint/assembly.py#L18)). I think the idea is to make sure that we aren't adding blocks to the tape that are 'inside' other blocks. An example would be that we want to tape a `solve` block but not the `assemble` calls that happen inside the `solve`.

ReubenHill: Okay I think I follow. Either way it's nice to see that this "just in case" code can be got rid of! I think previously it wasn't clear in either case where we needed it so it's good that now you got an error thrown

ReubenHill (approved):

PR #2594, opened 2022-10-20
HypreAMS/ADS: Fix parallel bugs when chopping the gradient and add support for quads and hexes

pbrubeck:
The gradient that we want to pass to hypre must be as sparse as possible. Previously we were chopping with numpy and didn't have a `comm`, nor were calling setUp on the new `Mat`. 

With this PR I also added support for quads and hexes, for which `Interpolator(grad(test), V)` fails because we don't have dual evaluation for `EnrichedElement`. Currently the user must supply their own sparse gradient, a feature that will be included in a future PR.


JDBetteridge:
@pbrubeck Please rebase on top of master to fix failing doc build

PR #2614, opened 2022-11-03
firedrake-clean purges pytools disk cache

### LINKED ISSUE #2610: Loo.py caching grows indefinitely
On my workstation I see
```
13G	/home/jbetteri/.cache/pytools/pdict-v4-loopy-schedule-cache-v4-2022.1-islpy2022.2.1-cgen2020.1-3988272b385fd770a3427b539e8b27e367f9db33-v1-py3.9.7.final.0
```
As one of many very large caches (The workstation supports multiple Python versions and Firedrake builds)

Perhaps we should be clearing the `loo.py` cache when we run `firedrake-update` or `firedrake-clean`.

connorjward:
Fixes #2610

### DIFF
--- a/scripts/firedrake-clean
+++ b/scripts/firedrake-clean
@@ -4,6 +4,10 @@ if __name__ == '__main__':
     import shutil
     import tempfile
     import firedrake_configuration
+    try:
+        import platformdirs as appdirs
+    except ImportError:
+        import appdirs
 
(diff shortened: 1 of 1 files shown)


JDBetteridge (approved):
Looks good, I was concerned about the try..import block, but that looks like it's exactly what pytools does

Issue #2619, opened 2022-11-07
OpenCascadeMeshHierarchy not working with 3d mesh


Referee1405:
When running ```OpenCascadeMeshHierarchy```, the function ```project_mesh_to_cad_3d``` is throwing an error on the line ```proj = GeomAPI_ProjectPointonSurf(pt, surf.Surface().Surface())```. Every time the function is called (I have put .real everywhere that I need to do so), I get the messages

```
free(): invalid pointer
Aborted (core dumped)
```

The file I want to run essentially just calls ```OpenCascade_mh``` with a step file. The Firedrake tests also fail when run, with similar errors.

rckirby:
Note: this issue is specific to running in complex mode, I believe?

connorjward:
Can you elaborate on "I have put `.real` everywhere that I need to do so"?

Referee1405:
Based on a previous issue brought up, and recommendations from rckirby, every time the line ```coorddata[node, :]``` was used in the function ```project_mesh_to_cad_3d```, I changed it to ```coorddata[node, :].real```. That took care of a different error that was popping up prior to this one.

Referee1405:
Here are the changes I made to the ```opencascade_mh.py``` file to get past the first error:

```
diff --git a/firedrake/mg/opencascade_mh.py b/firedrake/mg/opencascade_mh.py
index c83c93b6..9b07a4bc 100644
--- a/firedrake/mg/opencascade_mh.py
+++ b/firedrake/mg/opencascade_mh.py
@@ -151,14 +151,14 @@ def project_mesh_to_cad_3d(mesh, cad):
         surf = BRepAdaptor_Surface(face)

         for node in owned_nodes:
-            pt = gp_Pnt(*coorddata[node, :])
+            pt = gp_Pnt(*coorddata[node, :].real)

             proj = GeomAPI_ProjectPointOnSurf(pt, surf.Surface().Surface())
             if proj.NbPoints() > 0:
                 projpt = proj.NearestPoint()
                 coorddata[node, :] = projpt.Coord()
             else:
-                warnings.warn("Projection of point %s onto face %d failed" % (coorddata[node, :], id))
+                warnings.warn("Projection of point %s onto face %d failed" % (coorddata[node, :].real, id))

         edges = set(cad.edges_from_face(face))

@@ -186,7 +186,7 @@ def project_mesh_to_cad_3d(mesh, cad):
                 continue

             for node in intersecting_nodes:
[...]

PR #2631, opened 2022-11-14
Append .real to coordinates in OCMH, just in case we're in complex mode

rckirby:
`OCMH` breaks if you try to use it in complex mode because we store coordinates as complex numbers with 0 real parts and the geometry libraries are expecting proper real numbers.  

This addresses (at least part of) Issue #2619 .  That issue arises on a machine with an old OC library, so there may be two problems in the issue.  This fixes the obvious and easy part of it. 

One test fails locally for me (computing a volume) by a small tolerance, but the same fails in real mode on my machine.  We'll see what CI says here.


dham (approved):
Seems uncontroversial.

Issue #2447, opened 2022-05-18
Extra MPI routines on the Ensemble class


JHopeCollins:
The Ensemble class currently has methods which wrap the allreduce, (i)send, and (i)recv operations for Functions.
Reduce and broadcast are also commonly used MPI routines, so it would be useful if the Ensemble class handled these too.

Looking at the source code for the existing methods, it doesn't look like this should be too tricky. I needed to broadcast over an ensemble so I've already got code for what would be the body of the bcast method.

I'm happy to pick this up and submit a pull request, but I'd be interested to hear if there are any potential sticking points to be aware.

JHopeCollins:
Having thought about this a bit more, I have a couple more questions:
- The nonblocking ```isend```/```irecv``` methods of ```Ensemble``` access the ```Function``` data directly from ```f.dat[<int>].data``` instead of using a PETSc ```Vec``` from the context manager. This means that for a mixed function space, a seperate communication round is needed for each element of ```f.dat```, but it ensures correctness because a context-managed PETSc ```Vec``` may go out of scope before the communication finishes (see [this commit](https://github.com/firedrakeproject/firedrake/commit/cac67c027b4c2585757f266950048f9db41165b0)). For non-blocking operations, seperate communication rounds is not really a performance issue.
However, in that commit the blocking ```send```/```recv``` methods were also changed to no longer use the PETSc ```Vec``` context manager. For a ```Function``` from a  ```MixedFunctionSpace``` I believe that this may be a performance regression because it results in multiple blocking communication rounds, whereas using the PETSc ```Vec``` needs only a single communication round. The ```Vec``` scoping shouldn't be an issue here, precisely because the operations are blocking.
Is this correct, or is there something I've missed? If it is correct, is it worth reverting back to using the PETSc ```Vec``` for the blocking communications? The ```Ensemble.allreduce``` method already uses the context manager to access the underlying data of the ```Function``` being reduced. For the new methods I'd like to add to ```Ensemble```, the blocking ```bcast``` and ```reduce``` would use the context manager, and the non-blocking ```ibcast``` would access the ```f.dat.data``` directly.

[...]

PR #2639, opened 2022-11-18
Additional MPI method for `Ensemble`

JHopeCollins:
This wraps several more MPI communications functions in the `Ensemble`:
- `iallreduce`
- blocking and non-blocking `reduce`
- blocking and non-blocking `bcast`
- `sendrecv`
- `isendrecv`, not actually in the MPI standard but I think still useful to have.

The test suite has also been tidied up a bit.


JHopeCollins:
I haven't changed the way that blocking communications are sent for MixedFunctionSpaces as discussed in #2447. I have an implementation but I haven't been able to test it on multiple nodes yet to see it it really makes a difference.

firedrake/ensemble.py
```diff
@@ -35,32 +35,123 @@ def __init__(self, comm, M):
...
+        :arg f_reduced: the result of the reduction on rank root.
+        :arg op: MPI reduction operator.
+        :arg root: rank to reduce to
+        :raises ValueError: if function communicators mismatch each other or the ensemble spatial communicator, or is the functions are in different spaces
+        """
+        self._check_function(f, f_reduced)
+
+        # need to use `vec` not `vec_wo` for f_reduced otherwise function will be blanked out
+        # when rank!=root because `vec_wo` doesn't copy over existing data into the pyop2 vector
+        with f_reduced.dat.vec as vout, f.dat.vec_ro as vin:
```
wence-, line 111: `.vec_wo` for `vout`.

wence-, line 111: You should be able to pass `None` on all the other ranks (since `vout` isn't needed on those). But that suggests adding some new `.local_vec` properties in PyOP2. So meh.

firedrake/ensemble.py
```diff
@@ -35,32 +35,123 @@ def __init__(self, comm, M):
...
+            self.ensemble_comm.Free()
+            del self.ensemble_comm
+
+    def _check_function(self, f, g=None):
+        """
+        Check if function f (and possibly a second function g) is a valid argument for ensemble mpi routines
+
+        :arg f: The function to check
+        :arg g: Second function to check
+        :raises ValueError: if function communicators mismatch each other or the ensemble spatial communicator, or is the functions are in different spaces
```
wence-, line 53: Wrap long lines (don't forget to indent subsequent ones (by at least 4 spaces) so it's treated as part of the same block).

firedrake/ensemble.py
```diff
@@ -35,32 +35,123 @@ def __init__(self, comm, M):
...
+        :arg f: The a :class:`.Function` to reduce.
+        :arg f_reduced: the result of the reduction on rank root.
+        :arg op: MPI reduction operator.
+        :arg root: rank to reduce to
+        :returns: list of MPI.Request objects (one for each of f.split()).
+        :raises ValueError: if function communicators mismatch each other or the ensemble spatial communicator, or is the functions are in different spaces
+        """
+        self._check_function(f, f_reduced)
+
+        return [self.ensemble_comm.Ireduce(fdat.data, rdat.data, op=op, root=root)
```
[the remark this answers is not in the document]
JHopeCollins: Yes, I've updated this

firedrake/ensemble.py
JHopeCollins: What would the `.local_vec` properties be / why would they be needed? I haven't used that before

firedrake/ensemble.py
JHopeCollins: Updated now

firedrake/ensemble.py
JHopeCollins: I can avoid using `.vec` if that's preferable, it does avoid the extra copy out on root
```
if self.ensemble_comm.rank == root:
    with f_reduced.dat.vec_wo as vout, f.dat.vec_ro as vin:
        self.ensemble_comm.Reduce(vin.array_r, vout.array, op=op, root=root)
else:
    with f.dat.vec_ro as vin:
        self.ensemble_comm.Reduce(vin.array_r, None, op=op, root=root)
```

wence-: I think that doesn't work because `.vec_wo` is collective. `local_vec` would do the work of gluing together the pieces of a function, but give you a vec on `COMM_SELF` rather than `comm` (so it wouldn't be collective).

Thinking about it though, that probably is bad because you would get out of whack with state tracking.

JHopeCollins: Collective over which communicator?
This version passes the tests (although that doesn't mean it isn't undefined behaviour)

wence-: `f.dat.vec_XX` is collective over the communicator of `f.dat`

JHopeCollins: I think this is ok then, because `f_reduced` is defined over `ensemble.comm`, so `f_reduced` on any `rank!=root` does not "know" that the `f_reduced` on `rank==root` is constructing a `vec_wo`,

wence-: One thing to check is to ask (after doing this) on all ranks that share `f_reduced` that `.dat.vec.stateGet()` is the same.

JHopeCollins: It would appear that they are. Would you be happy with this version if I add a test for the stateGet?

wence-: Yes, thanks.

JHopeCollins: I've added this test now. I've not used `stateGet` before so please let me know if I'm accessing it incorrectly.
I've also updated the `Ensemble` section of the parallelism docs to reflect what methods we now implement.

tests/regression/test_ensembleparallelism.py
```diff
@@ -187,6 +187,18 @@ def test_ensemble_reduce(ensemble, mesh, W, urank, urank_sum, root, blocking):
     else:
         assert errornorm(Function(W).assign(10), u_reduce) < 1e-4
 
+    # check that u_reduce dat vector is still synchronised
+    ensemble_rank = ensemble.ensemble_comm.rank
+    spatial_rank = ensemble.comm.rank
+    if ensemble_rank == root:
+        from numpy import zeros
```
wence-, line 194: So here we are just looking at ensemble 0. Which may have multiple processes in it.

tests/regression/test_ensembleparallelism.py
```diff
@@ -187,6 +187,18 @@ def test_ensemble_reduce(ensemble, mesh, W, urank, urank_sum, root, blocking):
...
+        states = zeros(ensemble.comm.size, dtype=int)
+        with u_reduce.dat.vec as v:
+            states[spatial_rank] = v.stateGet()
+        ensemble.comm.Allgather(MPI.IN_PLACE, states)
```
wence-, line 198: And now we check that states are the same on all processes in this ensemble.

tests/regression/test_ensembleparallelism.py
```diff
@@ -187,6 +187,18 @@ def test_ensemble_reduce(ensemble, mesh, W, urank, urank_sum, root, blocking):
...
+        for state in states:
+            assert state == states[0]
```
wence-, line 200: assert len(set(states)) == 1

tests/regression/test_ensembleparallelism.py
```diff
@@ -187,6 +187,18 @@ def test_ensemble_reduce(ensemble, mesh, W, urank, urank_sum, root, blocking):
...
+
```
wence-, line 201: Can you also check that the state on `root` is now _different_ from the state on the other ensembles?

tests/regression/test_ensembleparallelism.py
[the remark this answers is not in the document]
JHopeCollins: I think that is what is going on here too, so that's a good start!
Do you want a picture for the reduction operation or for this test?

tests/regression/test_ensembleparallelism.py
JHopeCollins: I'm happy to put this in, but isn't this testing the implementation not the interface?
We want the state of the vector to be the same across the comm it's defined on (the ensemble member). But whether it is the same as on another rank is not part of the contract.

tests/regression/test_ensembleparallelism.py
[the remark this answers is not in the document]
wence-: I was trying to work through why this test is correct (and therefore why the reduction that you implemented is a valid thing to do). So plausibly a picture/description in the implementation of `Reduce` is the right place.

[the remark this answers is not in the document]
JHopeCollins: image

Is this ok? Each black box in the grid is one MPI process. Functions are defined over every process where `ensemble_comm.rank` is the same. So `ensemble_comm.rank==0` is true for every process in `freduce_0.comm` and false for every other process.
That means that the `vec` context manager is called on all processes in it's communicator, and only those processes.

JHopeCollins: Done

tests/regression/test_ensembleparallelism.py
dham: I think this is correct.

JHopeCollins:
As per the discussion at the Firedrake meeting, I've removed the check that `Ensemble.reduce` modifies the PETSc Vec of the reduced function more on the root rank than on other ranks (this test was passing before removing). There is still the test that the numeric values of the reduced function are untouched on all ranks except root.

Unless there are any other comments, I think this is ready for merging once the CI has finished.

JHopeCollins:
I messed up the rebase so the commit history on this branch is a bit messed up. As per @JDBetteridge's request, the changes made in this branch are now in a fresh PR #2681

PR #2644, opened 2022-11-22
Fix assign for components of vector function spaces

connorjward:
@JHopeCollins this should hopefully fix things for you.

### DIFF
--- a/firedrake/assign.py
+++ b/firedrake/assign.py
@@ -131,11 +131,11 @@ class Assigner:
             expression = expression.function
         expression = as_ufl(expression)
 
-        if not all(c.function_space() == assignee.function_space()
+        if not all(c.ufl_element() == assignee.ufl_element()
                    for c in extract_coefficients(expression)
                    if isinstance(c, Function)):
-            raise ValueError("All functions in the expression must be in the same "
-                             "function space as the assignee")
+            raise ValueError("All functions in the expression must have the same "
+                             "element as the assignee")
 
         self._assignee = assignee
         self._expression = expression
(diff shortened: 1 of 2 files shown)


connorjward:
> Thanks for getting on this so quick Connor. In ` test_mixed_expressions_indexed_fs` could there also be a test for assigning to a scalar function from a component of a mixed function, as well as the tests for assigning to a component of a mixed function from a scalar function please?

Sure. Done.

firedrake/assign.py
```diff
@@ -131,11 +131,11 @@ def __init__(self, assignee, expression, subset=None):
             expression = expression.function
         expression = as_ufl(expression)
 
-        if not all(c.function_space() == assignee.function_space()
+        if not all(c.ufl_element() == assignee.ufl_element()
```
JDBetteridge, line 134: I still have concerns about users trying to add functions on different meshes, but agree this shouldn't prevent this fix going in. We should open an issue with an MFE.

tests/regression/test_expressions.py
```diff
@@ -169,10 +169,60 @@ def test_tensor_expressions(expr, tfunctions):
...
+    f.sub(1).assign(two)
+    assert evaluate(f.dat.data, (0, 2))
+    f.sub(0).assign(one)
+    assert evaluate(f.dat.data, (1, 2))
+
+    one.assign(2*f.sub(0) + 1)
+    assert evaluate(one.dat.data, 3)
+
+    two += f.sub(1)
+    assert evaluate(two.dat.data, 4)
```
JDBetteridge, line 226: I think @JHopeCollins might have some more tests to contribute, but these can be added at a later date

JDBetteridge (approved):

firedrake/assign.py
connorjward: I've pushed a commit that does this.

JHopeCollins (approved):
Yes, happy with this

PR #2662, opened 2022-12-01
Guarantee finding cell for point on boundary

ReubenHill:
Only implemented for non-extruded meshes for now. Requires https://github.com/firedrakeproject/fiat/pull/33

### DIFF
--- a/firedrake/evaluate.h
+++ b/firedrake/evaluate.h
@@ -34,12 +34,12 @@ struct Function {
 	 */
 };
 
-typedef int (*inside_predicate)(void *data_,
+typedef PetscReal (*ref_cell_l1_dist)(void *data_,
 				struct Function *f,
 				int cell,
 				double *x);
 
-typedef int (*inside_predicate_xtr)(void *data_,
+typedef PetscReal (*ref_cell_l1_dist_xtr)(void *data_,
 				struct Function *f,
 				int cell,
 				int layer,
@@ -48,9 +48,10 @@ typedef int (*inside_predicate_xtr)(void *data_,
 extern int locate_cell(struct Function *f,
 		       double *x,
 		       int dim,
-		       inside_predicate try_candidate,
-		       inside_predicate_xtr try_candidate_xtr,
-		       void *data_);
+		       ref_cell_l1_dist try_candidate,
+		       ref_cell_l1_dist_xtr try_candidate_xtr,
+		       void *temp_ref_coords,
+		       void *found_ref_coords);
 
 extern int evaluate(struct Function *f,
 		    double *x,
--- a/firedrake/locate.c
+++ b/firedrake/locate.c
@@ -1,22 +1,38 @@
 #include <stdio.h>
 #include <stdlib.h>
 #include <spatialindex/capi/sidx_api.h>
-
+#include <float.h>
 #include <evaluate.h>
 
 int locate_cell(struct Function *f,
         double *x,
         int dim,
-        inside_predicate try_candidate,
-        inside_predicate_xtr try_candidate_xtr,
-        void *data_)
+        ref_cell_l1_dist try_candidate,
+        ref_cell_l1_dist_xtr try_candidate_xtr,
+        void *temp_ref_coords,
+        void *found_ref_coords)
 {
     RTError err;
     int cell = -1;
+    /* NOTE: temp_ref_coords and found_ref_coords are actually of type
+    struct ReferenceCoords but can't be declared as such in the function
+    signature because the dimensions of the reference coordinates in the
+    ReferenceCoords struct are defined by python when the code which
+    surrounds this is declared in pointquery_utils.py. We cast when we use the
+    ref_coords_copy function and trust that the underlying memory which the
+    pointers refer to is updated as necessary. */
+    double closest_ref_coord = DBL_MAX;
+    double current_closest_ref_coord =  -0.5;
+    /* NOTE: `tolerance`, which is used throughout this funciton, is a static
+       variable defined outside this function when putting together all the C
+       code that needs to be compiled - see pointquery_utils.py */
 
     if (f->sidx) {
         int64_t *ids = NULL;
         uint64_t nids = 0;
+        /* We treat our list of candidate cells (ids) from libspatialindex's
+            Index_Intersects_id as our source of truth: the point must be in
+            one of the cells. */
         err = Index_Intersects_id(f->sidx, x, x, dim, &ids, &nids);
         if (err != RT_None) {
             fputs("ERROR: Index_Intersects_id failed in libspatialindex!", stderr);
@@ -24,9 +40,21 @@ int locate_cell(struct Function *f,
         }
         if (f->extruded == 0) {
             for (uint64_t i = 0; i < nids; i++) {
-                if ((*try_candidate)(data_, f, ids[i], x)) {
+                current_closest_ref_coord = (*try_candidate)(temp_ref_coords, f, ids[i], x);
+                if (current_closest_ref_coord <= 0.0) {
+                    /* Found cell! */
                     cell = ids[i];
+                    memcpy(found_ref_coords, temp_ref_coords, sizeof(struct ReferenceCoords));
                     break;
+                }
+                else if (current_closest_ref_coord < closest_ref_coord) {
+                    /* getting closer... */
+                    closest_ref_coord = current_closest_ref_coord;
+                    if (closest_ref_coord < tolerance) {
+                        /* Close to cell within tolerance so could be this cell */
+                        cell = ids[i];
+                        memcpy(found_ref_coords, temp_ref_coords, sizeof(struct ReferenceCoords));
+                    }
                 }
             }
         }
@@ -35,9 +63,21 @@ int locate_cell(struct Function *f,
                 int nlayers = f->n_layers;
                 int c = ids[i] / nlayers;
                 int l = ids[i] % nlayers;
-                if ((*try_candidate_xtr)(data_, f, c, l, x)) {
+                current_closest_ref_coord = (*try_candidate_xtr)(temp_ref_coords, f, c, l, x);
+                if (current_closest_ref_coord <= 0.0) {
+                    /* Found cell! */
                     cell = ids[i];
+                    memcpy(found_ref_coords, temp_ref_coords, sizeof(struct ReferenceCoords));
                     break;
+                }
+                else if (current_closest_ref_coord < closest_ref_coord) {
+                    /* getting closer... */
+                    closest_ref_coord = current_closest_ref_coord;
+                    if (closest_ref_coord < tolerance) {
+                        /* Close to cell within tolerance so could be this cell */
+                        cell = ids[i];
+                        memcpy(found_ref_coords, temp_ref_coords, sizeof(struct ReferenceCoords));
+                    }
                 }
             }
         }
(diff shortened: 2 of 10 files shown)


firedrake/locate.c
```diff
@@ -25,9 +32,33 @@ int locate_cell(struct Function *f,
...
+                    /* Cell not found, but could be on cell boundary. We therefore look
+                       at our reference coordinates and find the point closest to being
+                       inside the reference cell. If we don't find a cell using try_candidate
+                       we assume that this process has found our cell. */
+                    /* First we must find the the reference coordinate which is closest to
+                       the interval [0, 1]. Note that ref_coords was updated as data_ by
+                       try_candidate. */
+                    for (uint64_t j = 0; j < dim; j++) {
+                        if(ref_coords->X[j] < 0.0){
+                            current_closest_ref_coord = ref_coords->X[j];
```
dham, line 49: Logic here is wrong if there is more than one negative coordinate (as happens if we're outside the corner of the cell). Should be `< current_closest_ref_coord` on line 48 and 52, and current_closest_ref_coord should be reset to 0.0 before line 47.

dham (changes requested):

ReubenHill:
I've either misunderstood the logic we discussed or there's a case we haven't thought of. If we are on the reference triangle we can have two coordinates, both of which are inside the interval [0, 1] but which is just outside the reference cell. The point (0.5, 0.500000001) for example.

dham:
> I've either misunderstood the logic we discussed or there's a case we haven't thought of. If we are on the reference triangle we can have two coordinates, both of which are inside the interval [0, 1] but which is just outside the reference cell. The point (0.5, 0.500000001) for example.

Ah. I was thinking we have the local barycentric coordinates, but of course we don't This complicates things because the rules are different for different element shapes. Simplices have the additional constraint that the sum of the coordinates <1. However wedges complicate this.

firedrake/locate.c
```diff
@@ -25,9 +32,36 @@ int locate_cell(struct Function *f,
...
+                        if(ref_coords->X[j] < 0.0 && current_closest_ref_coord < ref_coords->X[j]) {
+                            current_closest_ref_coord = ref_coords->X[j];
+                        }
+                        else if(1.0 - ref_coords->X[j] < 0.0 && current_closest_ref_coord < 1.0 - ref_coords->X[j]) {
+                            current_closest_ref_coord = 1.0 - ref_coords->X[j];
+                        }
+                    }
+                    /* Then, if it's lower than the last, we assume we must have found our
+                       cell until told otherwise */
+                    if(current_closest_ref_coord < closest_ref_coord){
```
ksagiyam, line 60: `current_closest_ref_coord > closest_ref_coord`?

The point is closer to the current cell if `current_closest_ref_coord` is "less negative".

ksagiyam:
I think we can add `double compute_distance_to_cell(double *X){...}` to `src` (code) in `_c_locator` that defines the right formula to compute the distance for the interested cell type and then use it directly in `locator.c`.

firedrake/locate.c
ReubenHill: Good spot - clearly been a while since I wrote any code! Fixed

ReubenHill:
> I think we can add `double compute_distance_to_cell(double *X){...}` to `src` (code) in `_c_locator` that defines the right formula to compute the distance for the interested cell type and then use it directly in `locator.c`.

Is there a way to work out which reference cell we need from the current `locate_cell` function arguments? Perhaps in `struct Function *f`?

ksagiyam:
We will need to define different `compute_distance_to_cell(double *X){...}` functions such as `compute_distance_to_cell_quadrilateral(double *X){...}` and `compute_distance_to_cell_triangle(double *X){...}`, but `locate_cell` function doesn't need to know which one it is using as, when we compile `locate_cell`, we already know the cell type, which is `mesh.ufl_cell()`.

So I think we define `src_compute_distance_to_cell_quadrilateral`, `src_compute_distance_to_cell_triangle`, etc., in `pointquery_utils`, and pick the right one as the definition of `compute_distance_to_cell(double *X)` according to `mesh.ufl_cell()` right before https://github.com/firedrakeproject/firedrake/blob/master/firedrake/pointquery_utils.py#L44, and simple use `compute_distance_to_cell()` in `locate_cell` function.

ReubenHill:
Okay I follow. So what's the easiest approach here? Come up with approaches for each cell type? Reference lines, squares and cubes ought to be easy since all the sides are axis aligned.

A google tells me that converting to barycentric coordinates is the go-to method for triangles and tetrahedra, though I wonder if there's a simpler approach involving working out what side of the non-axis-aligned line/plane you're on in reference space...

No idea what we'd do for hexes (I'm sure there's an answer on google somewhere).

ReubenHill:
My first task is to come up with a test that doesn't work on stock firedrake but ought to work after this. Any suggestions?

ReubenHill:
I've made a (currently bugged) implementation of a triangle which aims to give you the actual distance to the triangle rather than the value of one of the negative barycentric coordinates. I know it ought to be possible with a bit of maths for tetrahedrons as well. I'm wondering, however, if it's actually worth the effort: for the case where we are close to hypotenuse, outputting the barycentric coordinate `1-X[0]-X[1]` is not the actual distance to the triangle, but it's of the right order of magnitude. What do you think?

ksagiyam:
I don't think it matters much if we use true "distance" or not. That would in the end be distance in the reference space (and not in the physical space). Also, if we need to rely on this part of the code to determine the ownership, I think no one complains about which cell ends up taking the ownership. So I would just use one of the negative barycentric coordinates as it is easier.

ReubenHill:
The new `compute_distance_to_cell` works well for an interval, where we don't use libspatialindex, and passes both the new `test_outside_boundary_behaviour ` and `test_inside_boundary_behaviour ` tests. However these fail some non-intervals (and would on master!) because no spatial indices are being found close the the mesh edges despite the tolerance being high enough. My guess is that I need to increase the size of the bounding boxes to account for the given tolerance but don't know exactly how to do that.

ReubenHill:
I wonder to what extent I am duplicating the work of the fiat method UFCSimplex.contains_point ([here](https://github.com/firedrakeproject/fiat/blob/43bc840302475f3001aebb4a821e2727f927fd89/FIAT/reference_element.py#L508)) which gets used (with the given tolerance) in the codepath [here](https://github.com/firedrakeproject/firedrake/blob/2209aaeb3b1d4df065525152009b352c52fba0a4/firedrake/pointquery_utils.py#L69) via [here.](https://github.com/firedrakeproject/firedrake/blob/2209aaeb3b1d4df065525152009b352c52fba0a4/firedrake/pointquery_utils.py#L174)

Perhaps this new code doesn't add anything? I could check by only making this code run if all cells have been checked to see if it finds a cell...

firedrake/pointquery_utils.py
```diff
@@ -239,3 +242,174 @@ def compile_coordinate_element(ufl_coordinate_element, contains_eps, parameters=
...
+
+    Parameters
+    ----------
+    ufl_cell : ufl.Cell
+        The cell to compute the distance to.
+
+    Returns
+    -------
+    code : str
+        The C code for the distance computation.
```
ReubenHill, line 262: Switch back to google style, not numpy style

firedrake/pointquery_utils.py
```diff
@@ -239,3 +242,174 @@ def compile_coordinate_element(ufl_coordinate_element, contains_eps, parameters=
...
+    reference interval as at FIAT.reference_element.UFCTriangle.
+
+    If the point is inside the interval, the distance is negative.
+    Note that in this case the distance outputted is exact.
+
+    Returns
+    -------
+    code : str
+        The C code for the distance computation.
+    """
```
ReubenHill, line 304: Switch back to google style, not numpy style

firedrake/pointquery_utils.py
```diff
@@ -239,3 +242,174 @@ def compile_coordinate_element(ufl_coordinate_element, contains_eps, parameters=
...
+    """Generate C code for computing the aproximate distance to a
+    reference triangle as at FIAT.reference_element.UFCTriangle.
+
+    If the point is inside the triangle, the distance is negative.
+
+    Returns
+    -------
+    code : str
+        The C code for the distance computation.
+    """
```
ReubenHill, line 348: Switch back to google style, not numpy style

ReubenHill:
A reminder that the reason I am doing this is to use a minimal distance coordinate to determine cell ownership, rather than relying on some tolerance on a per-cell basis. The tolerance would now only be there to catch points outside the domain.

The key difference is that this is a calculation which needs to be done over several candidate cells rather than relying on a single cell saying 'yes I own this cell'. Whilst that's good enough for most cases, within a mesh you could still have points disappearing between cells.

Something along these lines seems to be going on with
https://github.com/firedrakeproject/firedrake/blob/e85b7f191dc037e1f613a91157166ef8a0bc9410/firedrake/pointquery_utils.py#L72-L86
which is used for prolongation and restriction kernels for multigrid in `mg/kernels.py`, e.g.
https://github.com/firedrakeproject/firedrake/blob/e85b7f191dc037e1f613a91157166ef8a0bc9410/firedrake/mg/kernels.py#L232-L291
I don't understand the logic employed here, as soon as you have negative distances the compute_celldist can give you a wildly wrong answer. Say you have a point with reference coordinates `(-1000.0, -1.0)` and want to know the distance to a reference triangle or square: the distance is around `1000` but it’ll return `1`! @wence- it looks like you were the last person to work on this, can you explain what's going on?

@dham has pointed out that having `compute_distance_to_cell` functions in firedrake means that we are defining reference cells both in FIAT and here. So really this all needs to happen in FIAT. I've tried doing this (see https://github.com/firedrakeproject/fiat/pull/33) and would, ideally, switch `inside_predicate` to give a distance instead of a boolean here
https://github.com/firedrakeproject/firedrake/blob/e85b7f191dc037e1f613a91157166ef8a0bc9410/firedrake/pointquery_utils.py#L217
The trouble is that there's no general way to generate C code from python. At the moment we use a sympy trick for `inside_check`
https://github.com/firedrakeproject/firedrake/blob/e85b7f191dc037e1f613a91157166ef8a0bc9410/firedrake/pointquery_utils.py#L65-L69
which works for returning a boolean, but can't work for giving a distance. One gets a sympy error
```
TypeError("cannot determine truth value of Relational")
```
The lovely thing about the current solution is that it composes for tensor product elements - I can't really write a general bit of C code which will compose, even if I ask FIAT to start generating C kernels.

I think I need to come up with another approach.

ReubenHill:
I wonder if
https://github.com/firedrakeproject/firedrake/blob/e85b7f191dc037e1f613a91157166ef8a0bc9410/firedrake/pointquery_utils.py#L72-L86
ought to be using `abs(PetscRealPart(...))` throughout. Then you avoid having completely wrong answers with negative coordinates. So whilst `-1000.0 > -1.0` is false, `abs(-1000.0) > abs(-1.0)` is true.

connorjward:
> The trouble is that there's no general way to generate C code from python.

Is loopy not the answer?

firedrake/mesh.py
```diff
@@ -1741,6 +1741,31 @@ def clear_cell_sizes(self):
         except AttributeError:
             pass
 
+    @property
+    def spatial_index_tolerance(self):
```
ReubenHill, line 1745: Make this a mesh constructor property

wence-:
> the distance is around `1000` but it’ll return `1`! @wence- it looks like you were the last person to work on this, can you explain what's going on?

Probably a bug. I'm very bad at geometry

firedrake/mg/kernels.py
```diff
@@ -28,10 +28,53 @@
...
+def compute_celldist_mg(fiat_cell, X="X", celldist="celldist"):
+    """Print C code to find the maximum of an input pointer string ``X`` which
+    is then multiplied by -1 with the result stored in ``celldist``.
+
+    If the ``fiat_cell`` is one dimensional then this doesn't find the maximum,
+    instead it simply multiplies the first entry of ``X`` by -1 and stores it
+    in ``celldist``.
+
+    **WARNING**: This function is specialised for use in multigrid kernels and
+    doesn't really compute a cell distance in any usual sense.
```
wence-, line 44: I think the incorrect computation is just a bug, what happens if you fix it?

firedrake/mg/kernels.py
```diff
@@ -248,7 +291,7 @@ def prolong_kernel(expression):
                     break;
                 }
 
-                %(compute_celldist)s
+                %(compute_celldist_mg)s
```
wence-, line 294: Yeah, the following code suggests to me that the "cell distance" computation that we're currently doing is complete nonsense and this should just be fixed to do something sane.

firedrake/mg/kernels.py
ReubenHill: To fix this we need to use the new `celldist_l1_c_expr` function in `pointquery_utils.py` and update the kernels themselves to work more like the new `locate.c` - I can have a go some time in a separate PR

ReubenHill: Draft PR running tests here https://github.com/firedrakeproject/firedrake/pull/2759

firedrake/evaluate.h
```diff
@@ -34,12 +34,12 @@ struct Function {
 	 */
 };
 
-typedef int (*inside_predicate)(void *data_,
+typedef PetscReal (*inside_predicate)(void *data_,
```
dham, line 37: not a predicate.

firedrake/mesh.py
```diff
@@ -1185,13 +1210,20 @@ class ExtrudedMeshTopology(MeshTopology):
     """Representation of an extruded mesh topology."""
 
     @PETSc.Log.EventDecorator()
-    def __init__(self, mesh, layers, name=None):
+    def __init__(self, mesh, layers, name=None, tolerance=1e-14):
```
dham, line 1213: change default tolerance to .5

dham: or 1

dham (changes requested):

dham (approved):

PR #2663, opened 2022-12-02
Mesh defaults to using dmplex comm if dmplex given

JHopeCollins:
`Mesh` promises to use the DMPlex comm if a DMPlex is passed to construct the mesh from. [This commit](https://github.com/firedrakeproject/firedrake/commit/8296de82826d092eea6152f63362229ec4cbcfe9) changes this to use COMM_WORLD if the comm argument is None, regardless of whether a DMPlex is given as the mesh or not.
This breaks when a DMPlex is given whose comm is not COMM_WORLD.

This PR should fix this, and adds a test to check for this.

### DIFF
--- a/firedrake/mesh.py
+++ b/firedrake/mesh.py
@@ -2043,12 +2043,11 @@ def Mesh(meshfile, **kwargs):
     import firedrake.function as function
 
     if isinstance(meshfile, PETSc.DMPlex):
-        default_comm = meshfile.comm.tompi4py()
+        user_comm = meshfile.comm.tompi4py()
     else:
-        default_comm = COMM_WORLD
+        user_comm = kwargs.get("comm", COMM_WORLD)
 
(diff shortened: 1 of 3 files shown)


JHopeCollins:
Given that the docstring says that the `comm` kwarg is ignored if the `meshfile` arg is a DMPlex, should getting `user_comm` instead be:
```
if isinstance(meshfile, PETSc.DMPlex):
    user_comm = meshfile.comm.tompi4py()
else:
    user_comm = kwargs.get("comm", COMM_WORLD)
```

wence-:
Yeah, that looks like a reasonable change. You should only have to provide a communicator if reading from a file (which isn't associated with a comm). If you already have a perfectly good communicator, we should just use that.

JDBetteridge (approved):
Yes, this was an oversight when sorting out all the comm stuff

PR #2666, opened 2022-12-05
user comm must match dm comm for Mesh construction

JHopeCollins:
When creating a Mesh from a DMPlex, the comm given must match the comm on the DMPlex.
This reverses the changes in #2663 where the comm was ignored and the DMPlex comm was used in place of the user comm.


wence-:
Why is this better behaviour than the reversion to "take communicator from the plex" that you reintroduced in #2663?

JDBetteridge (approved):
This should prevent the CI issue we saw previously. We can revisit this if further problems occur

Issue #2671, opened 2022-12-06
KSP objects not being destroyed after derivative calls


angus-g:
I have a [simple reproducing case](https://gist.github.com/angus-g/8d94eea2260cbab50af503e9855d4367) which sets up a problem, solves it forwards, then calls `reduced_functional.derivative()` several times. Running this case (on two ranks) and inspecting the total number of objects collected in `PETSc.garbage_view(comm=mesh.comm)` shows an increase of 20 KSP objects (one per forward timestep), for each derivative call:
```
Rank 0:: Total entries: 22
Rank 0:: Total entries: 42
Rank 0:: Total entries: 62
Rank 0:: Total entries: 82
Rank 0:: Total entries: 102
Rank 0:: Total entries: 122
Rank 0:: Total entries: 142
Rank 0:: Total entries: 162
Rank 0:: Total entries: 182
Rank 0:: Total entries: 202
```

However, changing the communicator used for the call to `PETSc.garbage_cleanup()` after the solve seems to address this issue. Namely, moving from the internal `self._problem.u.function_space().mesh()._comm` to `self._problem.u.function_space().mesh().comm`:
https://github.com/firedrakeproject/firedrake/blob/a16e3a51dbd22890bda3465ea1519680b55b93de/firedrake/variational_solver.py#L284-L286

With this change, the number of KSP objects stays constant:
```
Rank 0:: Total entries: 20
Rank 0:: Total entries: 20
Rank 0:: Total entries: 20
Rank 0:: Total entries: 20
Rank 0:: Total entries: 20
Rank 0:: Total entries: 20
Rank 0:: Total entries: 20
Rank 0:: Total entries: 20
Rank 0:: Total entries: 20
Rank 0:: Total entries: 20
```

Perhaps related to the recent communicator re-shuffle @JDBetteridge?

JDBetteridge:
Okay, this looks like we aren't honouring the contract to only use the internal comm for communication. Thanks for the MFE, I will investigate

sghelichkhani:
Just a word of caution that there seems to be objects both on `self._problem.u.function_space().mesh()._comm` and `self._problem.u.function_space().mesh().comm`. Using the proposed fix by Angus and changing `PETSc.garbage_view(comm=mesh.comm)` to  `PETSc.garbage_view(comm=mesh._comm)` in his reproducer I get:

```Rank 0:: Total entries: 17
Rank 0:: Total entries: 31
Rank 0:: Total entries: 49
[...]

PR #2674, opened 2022-12-07
Switch to private comm in matrix and linear_solver

JDBetteridge:
Addressing #2671

### DIFF
--- a/firedrake/linear_solver.py
+++ b/firedrake/linear_solver.py
@@ -7,7 +7,7 @@ from firedrake import dmhooks
 from firedrake.petsc import PETSc, OptionsManager, flatten_parameters
 from firedrake.utils import cached_property
 from firedrake.ufl_expr import action
-
+from pyop2.mpi import internal_comm, decref
 
 __all__ = ["LinearSolver"]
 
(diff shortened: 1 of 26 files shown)


firedrake/linear_solver.py
```diff
@@ -56,6 +56,7 @@ def __init__(self, A, *, P=None, solver_parameters=None,
                                                        ksp_defaults=self.DEFAULT_KSP_PARAMETERS)
         self.A = A
         self.comm = A.comm
+        self._comm = internal_comm(self.comm)
```
wence-, line 59: Remind me why you carry both of these around? Why not just have `self.comm = internal_comm(A.comm)` everywhere and `decref` that? We've never promised that the comm you get out of firedrake objects is a "user" one.

JDBetteridge: I agree, we never made such a promise. However, we do have documentation suggesting that users can use the `.comm` attribute, and it _is_ a public attribute. The main aim of this and the previous PR is to be more consistent with the comms that we hold references to since...
- PETSc garbage is now stored as a comm attribute
- previously it was all too easy to exhaust the number of available communicators.

dham (approved):

PR #2682, opened 2022-12-09
mesh: add mark_entities() method and RelabeledMesh() function

ksagiyam:
Allow for marking entities using indicator functions. This will be useful when:

- creating a mesh with modified labels
- making submeshes in the future

~TODO: make high-level interface: `RelabeledMesh()`.~ Done.

### DIFF
--- a/firedrake/cython/dmcommon.pyx
+++ b/firedrake/cython/dmcommon.pyx
@@ -3136,3 +3136,32 @@ def compute_point_cone_global_sizes(PETSc.DM dm):
     out = np.zeros((2, ), dtype=IntType)
     dm.comm.tompi4py().Allreduce(arraySizes, out, op=MPI.SUM)
     return out
+
+
+@cython.boundscheck(False)
+@cython.wraparound(False)
+def mark_points_with_function_array(PETSc.DM plex,
+                                    PETSc.Section section,
+                                    PetscInt height,
+                                    np.ndarray[PetscInt, ndim=1, mode="c"] array,
+                                    PETSc.DMLabel dmlabel,
+                                    PetscInt label_value):
+
+    """Marks points in a DMLabel using an indicator function array.
+
+    :arg plex: DMPlex representing the mesh topology
+    :arg section: Section describing the function space DoF layout and order
+    :arg height: Height of marked points (0 for cells, 1 for facets)
+    :arg array: Array representing the indicator function whose layout is
+        defined by plex, section, and height
+    :arg dmlabel: DMLabel that records marked entities
+    :arg label_value: Value used in dmlabel
+    """
+    cdef:
+        PetscInt pStart, pEnd, p, offset
+
+    get_height_stratum(plex.dm, height, &pStart, &pEnd)
+    for p in range(pStart, pEnd):
+        CHKERR(PetscSectionGetOffset(section.sec, p, &offset))
+        if array[offset] == 1:
+            CHKERR(DMLabelSetValue(<DMLabel>dmlabel.dmlabel, p, label_value))
--- a/firedrake/cython/dmcommon.pyx
+++ b/firedrake/cython/dmcommon.pyx
@@ -1853,36 +1853,6 @@ def get_facet_ordering(PETSc.DM plex, PETSc.Section facet_numbering):
 
 @cython.boundscheck(False)
 @cython.wraparound(False)
-def get_facet_markers(PETSc.DM dm, np.ndarray[PetscInt, ndim=1, mode="c"] facets):
-    """Get an array of facet labels in the mesh.
-
-    :arg dm: The DM that contains labels.
-    :arg facets: The array of facet points.
-    :returns: a numpy array of facet ids (or None if all facets had
-        the default marker).
-    """
-    cdef:
-        PetscInt nfacet, f, val
-        np.ndarray[PetscInt, ndim=1, mode="c"] ids
-        DMLabel label = NULL
-        PetscBool all_default = PETSC_TRUE
-
-    ids = np.empty_like(facets)
-    nfacet = facets.shape[0]
-    CHKERR(DMGetLabel(dm.dm, FACE_SETS_LABEL.encode(), &label))
-    for f in range(nfacet):
-        CHKERR(DMLabelGetValue(label, facets[f], &val))
-        if val != -1:
-            all_default = PETSC_FALSE
-        ids[f] = val
-    if all_default:
-        return None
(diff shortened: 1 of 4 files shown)


firedrake/mesh.py
```diff
@@ -2601,6 +2675,95 @@ def _on_rank_vertices(vertex_coords, parent_mesh):
...
+    :kwarg cell_indicators: list of indicator functions (DP0 or DQ0) that mark
+        selected cells; each function marks interested cells as 1.
+    :kwarg cell_indicator_labels: list of integers that enumerates cell_indicators;
+        cell_indicators and cell_indicator_labels must have the same length.
+    :kwarg facet_indicators: list of indicator functions (HDiv Trace 0) that mark
+        selected facets; each function marks interested facets as 1.
+    :kwarg facet_indicator_labels: list of integers that enumerates facet_indicators;
+        facet_indicators and facet_indicator_labels must have the same length.
+    :kwarg name: optional name of the output mesh object.
+    """
```
ksagiyam, line 2692: Do not need separate lists for cell and facet indicator functions.

ksagiyam: Fixed.

firedrake/cython/dmcommon.pyx
```diff
@@ -3166,3 +3166,32 @@ def compute_point_cone_global_sizes(PETSc.DM dm):
...
+@cython.boundscheck(False)
+@cython.wraparound(False)
+def mark_points_with_function_array(PETSc.DM plex,
+                                    PETSc.Section section,
+                                    PetscInt height,
+                                    np.ndarray[PetscInt, ndim=1, mode="c"] array,
+                                    PETSc.DMLabel dmlabel,
+                                    PetscInt label_value):
+
+    """Marks points in a DMLabel using indicator function array.
```
dham, line 3180: "using an indicator..."

firedrake/mesh.py
```diff
@@ -2601,6 +2677,85 @@ def _on_rank_vertices(vertex_coords, parent_mesh):
...
+
+    if not isinstance(mesh, MeshGeometry):
+        raise TypeError(f"mesh must be a MeshGeometry, not a {type(mesh)}")
+    tmesh = mesh.topology
+    if isinstance(tmesh, VertexOnlyMeshTopology):
+        raise NotImplementedError("Currently does not work with VertexOnlyMesh")
+    elif isinstance(tmesh, ExtrudedMeshTopology):
+        raise NotImplementedError("Currently does not work with ExtrudedMesh; use RelabeledMesh() on the base mesh and then extrude")
+    if not isinstance(indicator_functions, (list, tuple)) or \
+       not isinstance(subdomain_ids, (list, tuple)):
```
dham, line 2704: `isinstance(..., collections.abc.Sequence)`

dham: check for integer labels (`isinstance(..., numbers.Integral)`)

dham (changes requested):

firedrake/cython/dmcommon.pyx
ksagiyam: Fixed.

firedrake/mesh.py
ksagiyam: Fixed.

nabw:
Hi, I would like to use this PR, but it looks like it can't be used for BCs. The following code:

`mesh = UnitSquareMesh(10, 10)
Vmarks = FunctionSpace(mesh, "HDiv Trace", 0)
marks_fun = conditional(And(lt(x, 0.6), gt(x, 0.4)), 1.0, 0.0)
mesh = RelabeledMesh(mesh, [Function(Vmarks).interpolate(marks_fun)], [5])
assemble(1 * ds(5, domain=mesh)`

returns 0, and indeed if I add

`V = FunctionSpace(mesh, 'CG', 1)
bc = DirichletBC(V, Constant(1), 5)
f = Function(V)
bc.apply(f)`

the function instead interpolates in the entire domain the function marks_fun. Is this expected?

ksagiyam:
Thanks for the feedback. The former is not expected, but the latter is.

For the latter, if we want to use this functionality to apply Dirichlet BCs (in the standard sense), the indicator function `marks_fun` should only mark boundary facets (with `y < .1` and `y > .9`, say).

For the former, the problem is that there are conflicting labels for the part you marked with `5` (e.g. on a part of the top boundary, you have the original label `4` and the new one `5`), and currently Firedrake does not allow for multiple labels (it just picks `4` and ignores `5`). This restriction is technically imposed in `mesh.py`:
```
class MeshTopology:
    ...
    def _facets(self, kind):
        ...
        markers = dmcommon.get_facet_markers(dm, facets)
        ...
```
A workaround is to remove the unwanted labels as the following:
```
from firedrake import *
import firedrake.cython.dmcommon as dmcommon


mesh = UnitSquareMesh(10, 10)
Vmarks = FunctionSpace(mesh, "HDiv Trace", 0)
marks_fun = conditional(And(lt(x, 0.6), gt(x, 0.4)), 1.0, 0.0)
mesh = RelabeledMesh(mesh, [Function(Vmarks).interpolate(marks_fun)], [5])
mesh.topology_dm.clearLabelStratum(dmcommon.FACE_SETS_LABEL, 3)
mesh.topology_dm.clearLabelStratum(dmcommon.FACE_SETS_LABEL, 4)
print(assemble(1 * ds(5, domain=mesh)))
```
This prints `0.4` as expected.

I will think about what to do with this some time next week.

firedrake/mesh.py
```diff
@@ -2857,6 +2944,89 @@ def _on_rank_vertices(vertex_coords, parent_mesh, tolerance):
     return np.compress(on_rank, vertex_coords, axis=0)
 
 
+def RelabeledMesh(mesh, indicator_functions, subdomain_ids, **kwargs):
+    """Construct a new mesh that has new subdomain ids.
+
+    :arg mesh: base :class:`~.MeshGeometry` object using which the
```
ReubenHill, line 2950: I think this is what's causing doc build failure - I think this should be 
```
:class:`.MeshGeometry`
```

dham (approved):

dham (dismissed):

ksagiyam:
Fixed most of the link issues, but, for some, just used ``  ``.

connorjward (approved):
I think this will just need forcing through. Both tests and docs are failing but neither seems to be your fault.

ksagiyam:
Thanks Connor. I will.

PR #2703, opened 2023-01-07
Miscellaneous For Eigenvalue Probelm

UZerbinati:
Added the possibility to use non-unitary weight in the boundary conditions, this turns out to be useful for eigenvalue problems in Mixed form.


firedrake/utility_meshes.py
```diff
@@ -554,18 +556,18 @@ def RectangleMesh(
...
     """
 
     for n in (nx, ny):
         if n <= 0 or n % 1:
             raise ValueError("Number of cells must be a postive integer")
 
-    xcoords = np.linspace(0.0, Lx, nx + 1, dtype=np.double)
-    ycoords = np.linspace(0.0, Ly, ny + 1, dtype=np.double)
+    xcoords = np.linspace(originX, Lx, nx + 1, dtype=np.double)
+    ycoords = np.linspace(originY, Ly, ny + 1, dtype=np.double)
```
wence-, line 570: This is now at odds with the docstring that says that the mesh has extent `Lx`/`Ly` (it now has extent `Lx - originX`, ...).

firedrake/bcs.py
```diff
@@ -265,7 +265,7 @@ class DirichletBC(BCBase, DirichletBCMixin):
     '''
 
     @DirichletBCMixin._ad_annotate_init
-    def __init__(self, V, g, sub_domain, method=None):
+    def __init__(self, V, g, sub_domain, weight=1.0, method=None):
```
wence-, line 268: Docstring for the weight.

firedrake/utility_meshes.py
UZerbinati: Indeed, I've to fix that. Thank you !

dham (changes requested):

firedrake/utility_meshes.py
```diff
@@ -541,9 +543,11 @@ def RectangleMesh(
 
     :arg nx: The number of cells in the x direction
     :arg ny: The number of cells in the y direction
-    :arg Lx: The extent in the x direction
-    :arg Ly: The extent in the y direction
-    :kwarg quadrilateral: (optional), creates quadrilateral mesh.
+    :arg Lx: The extent in the x direction from the origin
+    :arg Ly: The extent in the y direction from the origin
+    :arg originX: The X coordintes of the origin, deafult 0.
```
dham, line 548: typo

firedrake/utility_meshes.py
```diff
@@ -541,9 +543,11 @@ def RectangleMesh(
...
+    :arg originY: The Y coordintes of the origin, deafult 0.
```
ReubenHill, line 549:
```suggestion
    :arg originX: The X coordinates of the origin, default 0.
    :arg originY: The Y coordinates of the origin, default 0.
```

UZerbinati: @ReubenHill thank you so much, my dyslexia was acting up!

firedrake/utility_meshes.py
dham: Not quite. Let's call the parameters the lower left and upper right corners of the box and then this works.

dham (changes requested):
Please also add eigenvalue test.

firedrake/assemble.py
```diff
@@ -69,6 +69,9 @@ def assemble(expr, *args, **kwargs):
     :kwarg zero_bc_nodes: If ``True``, set the boundary condition nodes in the
         output tensor to zero rather than to the values prescribed by the
         boundary condition. Default is ``False``.
+    :kwarg weight: weight of the boundary condition, i.e. the scalar in front of the
+        identity matrix corresponding to the boundary nodes, 1.0 by default.
```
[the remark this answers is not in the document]
UZerbinati: I think this time I really merged the suggested change, thank you!

firedrake/utility_meshes.py
UZerbinati: @ReubenHill Should I remove the default value also here?

UZerbinati:
> Please also add eigenvalue test.

So I was checking the test_slepc.py and found out that there was already an interesting way to shift eigenvalue, which I think it's not ideal.
`vals = np.repeat(1E8, len(bc.nodes))
 A.setValuesLocalRCV(bc.nodes.reshape(-1, 1),bc.nodes.reshape(-1, 1),vals.reshape(-1, 1))`
Can I leave it there as a comment so that people can be aware also of this option ?

dham (approved):

PR #2707, opened 2023-01-10
Add Hiptmair multigrid PC

pbrubeck:
A Python PC implementing a two-level method arising from a Hiptmair decomposition. 

Allows a user-provided tabulation of `grad` (`curl`) to transfer scalar (vector) potentials in `Vc` into the target function space `Vf`. This gives flexibility to use a sparse matrix for FDM elements, or to do it matrix-free via `Interpolator(grad(TestFunction(Vc)), Vf)` when the transpose operation becomes available.


firedrake_citations/__init__.py
```diff
@@ -277,3 +277,17 @@ def print_at_exit(cls):
   url =          {https://www.jstor.org/stable/43693530}
 }
 """)
+
+Citations().add("Hiptmair1998", """
+@Misc{Kolev2009,
```
pefarrell, line 282: Should this be `Hiptmair1998`?

firedrake/preconditioners/hiptmair.py
```diff
@@ -0,0 +1,185 @@
...
+        formdegree = V.finat_element.formdegree
+        if formdegree == 1:
+            dminus = grad
+            cfamily = "Lagrange"
+            G_callback = appctx.get("get_gradient", None)
+        elif formdegree == 2:
+            dminus = curl
+            if V.shape:
+                dminus = lambda u: as_vector([curl(u[k, ...]) for k in range(u.ufl_shape[0])])
+            cfamily = "N1curl" if mesh.ufl_cell().is_simplex() else "NCE"
```
pefarrell, line 145: What if you're using elements of the second kind for the H(div) space? Then I expect you should use elements of the second kind for the H(curl) space?

firedrake/preconditioners/hiptmair.py
```diff
@@ -0,0 +1,185 @@
...
+        problem = ctx._problem
+        a = problem.Jp or problem.J
+        bcs = problem.bcs
+
+        mesh = V.mesh()
+        element = V.ufl_element()
+        formdegree = V.finat_element.formdegree
+        if formdegree == 1:
+            dminus = grad
+            cfamily = "Lagrange"
```
pefarrell, line 139: Same as below, are there situations where this should be the serendipity elements? (If the element used for H(curl) is the Arnold-Awanou element?)

firedrake/preconditioners/hiptmair.py
```diff
@@ -0,0 +1,185 @@
...
+
+        variant = element.variant()
+        degree = element.degree()
+        try:
+            degree = max(degree)
+        except TypeError:
+            pass
+
+        celement = FiniteElement(cfamily, cell=mesh.ufl_cell(), degree=degree, variant=variant)
+        if degree > 1:
```
pefarrell, line 158: Can you add a comment on what this is doing?

firedrake/preconditioners/hiptmair.py
```diff
@@ -0,0 +1,185 @@
...
+        coarse_space_bcs = [bc.reconstruct(V=coarse_space, g=0) for bc in bcs]
+
+        # Get only the zero-th order term of the form
+        beta = replace(expand_derivatives(a), {grad(t): zero(grad(t).ufl_shape) for t in a.arguments()})
+
+        test = TestFunction(coarse_space)
+        trial = TrialFunction(coarse_space)
+        coarse_operator = beta(dminus(test), dminus(trial), coefficients={})
+        if formdegree > 1 and degree > 1:
+            shift = appctx.get("hiptmair_shift", None)
```
pefarrell, line 175: Should the default be to shift?

pefarrell (changes requested):
Can you please add a test?

firedrake/preconditioners/hiptmair.py
pbrubeck: At low-order with point-Jacobi smoothers in the H(curl) space there is no need to do the shift

pefarrell: OK, so the default should be a function of `(low-order && point-jacobi)`?

tests/multigrid/test_hiptmair.py
```diff
@@ -0,0 +1,82 @@
...
+    u_exact = Constant((1, 2, 4))
+    f = u_exact
+
+    u = Function(V)
+    v = TestFunction(V)
+    F = inner(d(u), d(v))*dx + inner(u, v)*dx - inner(f, v)*dx
+    bcs = [DirichletBC(V, u_exact, "on_boundary")]
+
+    solve(F == 0, u, bcs=bcs, solver_parameters=parameters)
+    return norm(u_exact - u, str(sobolev))
```
pefarrell, line 64: This test just tests for convergence of the discretisation. But shouldn't we test for how the solver behaved?

firedrake/preconditioners/hiptmair.py
```diff
@@ -0,0 +1,262 @@
...
+        coarse_space_bcs = tuple([bc.reconstruct(V=coarse_space, g=0) for bc in bcs])
+
+        # Get only the zero-th order term of the form
+        beta = replace(expand_derivatives(a), {grad(t): zero(grad(t).ufl_shape) for t in a.arguments()})
+
+        test = TestFunction(coarse_space)
+        trial = TrialFunction(coarse_space)
+        coarse_operator = beta(dminus(test), dminus(trial), coefficients={})
+
+        if formdegree > 1 and degree > 1:
```
pefarrell, line 176: If someone sets a shift, I think it should be used, even if degree == 1 (we set mechanism, not policy)

pefarrell (changes requested):

firedrake/preconditioners/hiptmair.py
pbrubeck: It could be that the user might want to set the shift only for the high order space in a p-multgrid hierarchy. Should the shift be a function of the degree (in a similar way the callback to tabulate the gradient is a function of the spaces and bcs)?

pefarrell: Ah, hmm, I see. In that case I guess it makes sense to just code the sensible thing, as you have done. Making the user supply a shift function that depends on degree seems overly complicated.

pefarrell (approved):

firedrake/preconditioners/hiptmair.py
```diff
@@ -0,0 +1,262 @@
...
+            viewer.printfASCII("Two level PC\n")
+            self.pc.view(viewer)
+
+
+class HiptmairPC(TwoLevelPC):
+
+    _prefix = "hiptmair_"
+
+    def coarsen(self, pc):
+        from firedrake_citations import Citations
```
dham, line 123: move imports to top of file.

firedrake/preconditioners/hiptmair.py
```diff
@@ -0,0 +1,262 @@
...
+            self.pc.applyTranspose(X, Y)
+
+    def view(self, pc, viewer=None):
+        super(TwoLevelPC, self).view(pc, viewer)
+        if hasattr(self, "pc"):
+            viewer.printfASCII("Two level PC\n")
+            self.pc.view(viewer)
+
+
+class HiptmairPC(TwoLevelPC):
```
dham, line 118: This needs a serious docstring.

firedrake/preconditioners/hiptmair.py
```diff
@@ -0,0 +1,262 @@
...
+        return type(ele)(div_to_curl(ele._element))
+    elif isinstance(ele, ufl.HDivElement):
+        return ufl.HCurlElement(div_to_curl(ele._element))
+    elif isinstance(ele, ufl.HCurlElement):
+        raise ValueError("Expecting an H(div) element")
+    else:
+        degree = ele.degree()
+        family = ele.family()
+
+        if family in ["Lagrange", "CG", "Q"]:
```
dham, line 241: Replace with dictionary?

dham (changes requested):

dham (dismissed):

JDBetteridge (approved):
Previously approved

firedrake/preconditioners/hiptmair.py
```diff
@@ -0,0 +1,262 @@
...
+    @abc.abstractmethod
+    def coarsen(self, pc):
+        """Return a tuple with coarse bilinear form, coarse
+           boundary conditions, and coarse-to-fine interpolation matrix
+        """
+        raise NotImplementedError
+
+    def initialize(self, pc):
+        from firedrake import parameters
+        from firedrake.assemble import allocate_matrix, TwoFormAssembler
```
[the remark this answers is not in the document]
JDBetteridge: I think you said this causes an import loop, correct me if I'm wrong.

JDBetteridge (dismissed):

firedrake/preconditioners/hiptmair.py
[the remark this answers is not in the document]
pbrubeck: Yes, I can confirm this. If I move that line to the top I get:
`ImportError: cannot import name 'allocate_matrix' from partially initialized module 'firedrake.assemble'`

JDBetteridge (approved):
(previously approved)

Issue #2595, opened 2022-10-20
Error when taking the derivative of a form which includes the cell-normals of a manifold.


JHopeCollins:
When trying to create a `LinearVariationalProblem` from the derivative of a form, UFL throws an error if:

- The mesh is a manifold
- The coordinate degree is greater than 1
- The form includes the gradient of the cell normals

MFE:
```
import firedrake as fd

coords_degree = 2  # works when coords_degree=1, breaks when coords_degree=2

mesh = fd.IcosahedralSphereMesh(radius=1,
                                refinement_level=2,
                                degree=coords_degree)
x = fd.SpatialCoordinate(mesh)
mesh.init_cell_orientations(x)
n = fd.CellNormal(mesh)

V = fd.FunctionSpace(mesh, "BDM", 1)

u = fd.Function(V)
v = fd.TestFunction(V)

F = fd.inner(fd.grad(fd.inner(v, n)), u)*fd.dx

A = fd.derivative(F, u)

f = fd.Function(V)
b = fd.inner(v, f)*fd.dx

prob = fd.LinearVariationalProblem(A, b, u)  # error here
```
Error message:
>UFL:ERROR Currently no support for ReferenceGrad in CoefficientDerivative.
Traceback (most recent call last):
  File "refgrad_test.py", line 30, in <module>
    prob = fd.LinearVariationalProblem(A, b, u)  # error here
  File "PETSc/Log.pyx", line 115, in petsc4py.PETSc.Log.EventDecorator.decorator.wrapped_func
  File "PETSc/Log.pyx", line 116, in petsc4py.PETSc.Log.EventDecorator.decorator.wrapped_func
  File "/home/jhc/icl_ra/firedrake/src/firedrake/firedrake/variational_solver.py", line 316, in __init__
    F = ufl_expr.action(J, u) - L
  File "PETSc/Log.pyx", line 115, in petsc4py.PETSc.Log.EventDecorator.decorator.wrapped_func
  File "PETSc/Log.pyx", line 116, in petsc4py.PETSc.Log.EventDecorator.decorator.wrapped_func
  File "/home/jhc/icl_ra/firedrake/src/firedrake/firedrake/ufl_expr.py", line 212, in action
    return ufl.action(form, coefficient)
  File "/home/jhc/icl_ra/firedrake/src/ufl/ufl/formoperators.py", line 109, in action
    form = expand_derivatives(form)
[...]

Issue #107, opened 2013-12-05
Interior facet integrals on extruded


colinjcotter:
Request ability to assemble interior facet integrals for extruded meshes. Required in dynamical core project for temperature advection, pressure gradient term (and possibly vorticity equation).

PR #2715, opened 2023-01-12
Check UFL FEniCS merge

nbouziani:
This checks that we are not breaking anything by merging UFL FEniCS. See the corresponding [UFL PR](https://github.com/firedrakeproject/ufl/pull/32).

This PR will also contain the fixes resulting from recent UFL changes since some deprecated UFL objects got removed (e.g. `ufl.FacetElement`).

This PR is required for merging the UFL PR and therefore merging the recent dual space changes.


.github/workflows/build.yml
```diff
@@ -42,7 +42,7 @@ jobs:
       - name: Build Firedrake
         run: |
           cd ..
-          ./firedrake/scripts/firedrake-install $COMPLEX --venv-name build --tinyasm --disable-ssh --minimal-petsc --slepc --documentation-dependencies --install thetis --install gusto --install icepack --install irksome --install femlium --no-package-manager || (cat firedrake-install.log && /bin/false)
+          ./firedrake/scripts/firedrake-install $COMPLEX --venv-name build --tinyasm --disable-ssh --minimal-petsc --slepc --documentation-dependencies --install thetis --install gusto --install icepack --install irksome --install femlium --no-package-manager --package_branch ufl merge_fenics || (cat firedrake-install.log && /bin/false)
```
ReubenHill, line 45: This is failing because you've written `--package_branch` instead of `--package-branch` (`-` rather than `_`)

nbouziani: Yeah I was expecting tests to fail because master is broken so didn't check. Thanks!

nbouziani:
There is still a test using deprecated elements (`FacetElement` and `InteriorElement`), namely: [test_cg_poisson.py](https://github.com/firedrakeproject/firedrake/blob/master/tests/slate/test_cg_poisson.py)

This test checks CG performance, I am not sure if it is inherently dependent on these elements, in which case I guess we should remove it, or if we should reformulate it.

ReubenHill:
Why is this diff so big? Are you attempting to deal with upstream UFL changes as well as testing the recent merge of dual?

nbouziani:
> Why is this diff so big? Are you attempting to deal with upstream UFL changes as well as testing the recent merge of dual?

No, this is about upstream UFL changes before dual that didn't get merged and are breaking Firedrake.

tests/regression/test_interior_elements.py
```diff
@@ -1,14 +0,0 @@
...
-    mesh = UnitSquareMesh(5, 5)
-    V_elt = FiniteElement("RT", triangle, 2)
-    W2_elt = InteriorElement(V_elt)
-    W2 = FunctionSpace(mesh, W2_elt)
-    g = Function(W2).assign(1)
-    n = FacetNormal(mesh)
-
-    # For H(div) elements, interior dofs have u.n = 0 on facets
-    assert abs(assemble(dot(g('+'), n('+'))*dS)) < 1e-14
-    assert abs(assemble(dot(g, n)*ds)) < 1e-14
```
wence-, line 14: These tests should be ported to use `RestrictedElement(..., restriction="interior")` I think

tests/regression/test_facet_elements.py
```diff
@@ -1,18 +0,0 @@
...
-    V = FunctionSpace(mesh, V_elt)
-    W1 = FunctionSpace(mesh, W1_elt)
-    f = Function(V).assign(1)
-    g = Function(W1).assign(1)
-    n = FacetNormal(mesh)
-
-    assert abs(assemble(inner(f, f)*dx) - assemble(inner(g, g)*dx)) < 2e-12
-    assert abs(assemble(inner(f, f)*ds) - assemble(inner(g, g)*ds)) < 2e-12
-    assert abs(assemble(inner(f, n)*ds) - assemble(inner(g, n)*ds)) < 2e-12
-    assert abs(assemble(inner(f('+'), n('+'))*dS) - assemble(inner(g('+'), n('+'))*dS)) < 2e-12
```
wence-, line 18: This should be tested using `RestrictedElement(..., restriction="facet")` ?

nbouziani:
> > No, this is about upstream UFL changes before dual that didn't get merged and are breaking Firedrake.
> 
> Aha! So hopefully you've fixed #2595 ?

If you are referring to what got merged in `firedrakeproject/ufl` but not in `FEniCS/ufl`, that's a different issue for which I think the solution is to update and merge [#107](https://github.com/FEniCS/ufl/pull/107/files). This PR is the other way around.

tests/regression/test_interior_elements.py
nbouziani: Oh yes, thanks!

tests/regression/test_facet_elements.py
nbouziani: Done thanks!

ReubenHill:
> > > No, this is about upstream UFL changes before dual that didn't get merged and are breaking Firedrake.
> > 
> > 
> > Aha! So hopefully you've fixed #2595 ?
> 
> If you are referring to what got merged in `firedrakeproject/ufl` but not in `FEniCS/ufl`, that's a different issue for which I think the solution is to update and merge [#107](https://github.com/FEniCS/ufl/pull/107/files). This PR is the other way around.

I don't think that was the actual cause of the issue, instead I believe that that merge having happened stopped us updating UFL from upstream for a while (because we had diverged) and then, when we did, we had regressed as outlined in #2595 . The suggested fix is at the bottom of that issue and is not related to trimmed serendipity ( [#107](https://github.com/FEniCS/ufl/pull/107/files)).

nbouziani:
> > > > No, this is about upstream UFL changes before dual that didn't get merged and are breaking Firedrake.
> > > 
> > > 
> > > Aha! So hopefully you've fixed #2595 ?
> > 
> > 
> > If you are referring to what got merged in `firedrakeproject/ufl` but not in `FEniCS/ufl`, that's a different issue for which I think the solution is to update and merge [#107](https://github.com/FEniCS/ufl/pull/107/files). This PR is the other way around.
> 
> I don't think that was the actual cause of the issue, instead I believe that that merge having happened stopped us updating UFL from upstream for a while (because we had diverged) and then, when we did, we had regressed as outlined in #2595 . The suggested fix is at the bottom of that issue and is not related to trimmed serendipity ( [#107](https://github.com/FEniCS/ufl/pull/107/files)).

This PR is about merging UFL upstream changes until dual while the issue you mentioned is about something that already got merged and that is a bug in both UFLs at the moment. Therefore, it's a different problem as one would fix issue [#2595](https://github.com/firedrakeproject/firedrake/issues/2595) by making a PR against UFL upstream that will then get merged and only show up on our end once we have merged what happened in-between, namely dual and the content of this PR which is breaking Firedrake. We can discuss this in today's meeting.

dham (approved):
I think this looks fine. Please remove the build.yml change and we can get merging.

Issue #2623, opened 2022-11-09
Rename split (`.split` -> `.subfunctions`)


JDBetteridge:
We should rename split from `.split` -> `.subfunctions`.

Someone should pick this up at some point

uygarkov:
Hello, my name is Uygar. I am a master student at University of Stuttgart. 

As a term project of my course [Simulation Software Engineering](https://simulation-software-engineering.github.io/homepage/) , we were assigned to make a contribution to the open-source simulation softwares. Since I am interested in Finite Element Method, I wanted to work on the Firedrake project.

The type of contribution is up to us, but we are learning how to use Git tools, documentate, use containers, package softwares. I am quite new to Software Engineering but I have worked on Python before so I would like to work on this issue (or any other issue that you can recommend) after discussing it with my professors. Have a nice year! 

Uygar Kovanci

JDBetteridge:
Hi @uygarkov, your contribution would be very welcome. Please feel free to have a go at fixing this issue and opening a PR. If you have any questions feel free to ask, see [here](https://firedrakeproject.org/contact.html) for contacting the team. The slack channel is very active.

connorjward:
@uygarkov I think the things that need doing are:
- Rename the current `.split` function to `.subfunctions` (in `function.py` I think but check everywhere for it).
- Write a `.split` function that looks something like:
```py
def split(self):
  import warnings
  warnings.warn(DeprecationWarning, <informative message>)
  return self.subfunctions
```
- Check through all the documentation and notebooks to make sure that we don't call `.split` anywhere.

wence-:
> ```python
> def split(self):
>   import warnings
>   warnings.warn(DeprecationWarning, <informative message>)
>   return self.subfunctions
> ```

Per https://docs.python.org/3/library/warnings.html#warning-categories

for user-facing warnings one should prefer `FutureWarning` over `DeprecationWarning`.

PR #2725, opened 2023-01-21
Rename .split() -> .subfunctions

uygarkov:
# Closes issue #2623 

## *Update* 
- On my virtual box, it passed the lint tests and then the tests appeared with warning and error in my first commit without any warning from my side. 
 
## Changes
- `.split()` method was renamed as `.subfunctions`
- `@utils.cached_property` decorator was added for `subfunctions(self)` property
- [FutureWarning](https://docs.python.org/3/library/exceptions.html#FutureWarning) added for `.split()` method as it concerns the end user. I want to address the warning as simple as possible. 
- `pytest` codes and `docs` were edited.
- In documentation `:meth:` was changed to `:attr:` which complies with [Sphinx documentation. ](https://www.sphinx-doc.org/en/master/usage/restructuredtext/domains.html#role-py-attr)

## Why?

- Firedrake's `split()` function and `.split()` method do conceptually the same thing, but they return different types of objects and are thus used in different contexts, which has been confusing users for years, so it was decided to rename `.split()` method as `.subfunctions` property.

## Testing

```python
from firedrake import *

mesh = UnitSquareMesh(1, 1)
V = FunctionSpace(mesh, "DG", 0)
W = V*V

f = Function(W)
u, p = f.subfunctions
x, y = f.split()

print(u == x)
print(p == y)
```

**Result:**

```
/home/firedrake/firedrake/src/firedrake/firedrake/function.py:325
FutureWarning: The .split() method is deprecated, please use the .subfunctions property instead
warnings.warn("The .split() method is deprecated, please use the .subfunctions property instead", category=FutureWarning)
True
True
```

Fingers crossed. 🤞

### DIFF
--- a/firedrake/constant.py
+++ b/firedrake/constant.py
@@ -111,13 +111,14 @@ class Constant(ufl.Coefficient, ConstantMixin):
         """Return a null function space."""
         return None
 
+    @utils.cached_property
     def subfunctions(self):
         return (self,)
-        
+
     def split(self):
         import warnings
-        warnings.warn(".split() was renamed as .subfunctions()", category=FutureWarning)
-        return self.subfunctions()
+        warnings.warn(".split() method was changed to .subfunctions property (without paranthesis)", category=FutureWarning)
+        return self.subfunctions
 
     def cell_node_map(self, bcs=None):
         """Return a null cell to node map."""
--- a/firedrake/constant.py
+++ b/firedrake/constant.py
@@ -111,8 +111,13 @@ class Constant(ufl.Coefficient, ConstantMixin):
         """Return a null function space."""
         return None
 
-    def split(self):
+    def subfunctions(self):
         return (self,)
+        
+    def split(self):
+        import warnings
+        warnings.warn(".split() was renamed as .subfunctions()", category=FutureWarning)
+        return self.subfunctions()
 
     def cell_node_map(self, bcs=None):
         """Return a null cell to node map."""
--- a/firedrake/constant.py
+++ b/firedrake/constant.py
@@ -117,7 +117,7 @@ class Constant(ufl.Coefficient, ConstantMixin):
 
     def split(self):
         import warnings
-        warnings.warn(".split() method was changed to .subfunctions property (without paranthesis)", category=FutureWarning)
+        warnings.warn("The .split() method is deprecated, please use the .subfunctions property instead", category=FutureWarning)
         return self.subfunctions
 
     def cell_node_map(self, bcs=None):
--- a/firedrake/function.py
+++ b/firedrake/function.py
@@ -106,16 +106,22 @@ class CoordinatelessFunction(ufl.Coefficient):
         return self.uid
 
     @utils.cached_property
-    def _split(self):
+    def _subfunctions(self):
         return tuple(CoordinatelessFunction(fs, dat, name="%s[%d]" % (self.name(), i))
                      for i, (fs, dat) in
                      enumerate(zip(self.function_space(), self.dat)))
 
     @PETSc.Log.EventDecorator()
-    def split(self):
+    def subfunctions(self):
         r"""Extract any sub :class:`Function`\s defined on the component spaces
         of this this :class:`Function`'s :class:`.FunctionSpace`."""
-        return self._split
+        return self._subfunctions
+    
+    @PETSc.Log.EventDecorator()
+    def split(self):
+        import warnings
+        warnings.warn(".split() was renamed as .subfunctions()", category=FutureWarning)
+        return self._subfunctions
 
     @utils.cached_property
     def _components(self):
@@ -132,7 +138,7 @@ class CoordinatelessFunction(ufl.Coefficient):
 
         :arg i: the index to extract
 
-        See also :meth:`split`.
+        See also :meth:`subfunctions`.
 
         If the :class:`Function` is defined on a
         rank-n :class:`~.FunctionSpace`, this returns a proxy object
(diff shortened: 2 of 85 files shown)


firedrake/function.py
```diff
@@ -302,16 +308,22 @@ def __dir__(self):
...
-    @FunctionMixin._ad_annotate_split
-    def split(self):
+    def subfunctions(self):
         r"""Extract any sub :class:`Function`\s defined on the component spaces
         of this this :class:`Function`'s :class:`.FunctionSpace`."""
-        return self._split
+        return self._subfunctions
+    
+    @PETSc.Log.EventDecorator()
+    @FunctionMixin._ad_annotate_split
```
connorjward, line 322: This should get renamed too.

firedrake/function.py
```diff
@@ -106,16 +106,22 @@ def ufl_id(self):
...
     @utils.cached_property
-    def _split(self):
+    def _subfunctions(self):
         return tuple(CoordinatelessFunction(fs, dat, name="%s[%d]" % (self.name(), i))
                      for i, (fs, dat) in
                      enumerate(zip(self.function_space(), self.dat)))
 
     @PETSc.Log.EventDecorator()
-    def split(self):
+    def subfunctions(self):
```
connorjward, line 115: So I think that a `subfunctions` property is better, and this could be made clear in the `FutureWarning`, but I will raise it in the Firedrake meeting this Wednesday to discuss with the other developers.

connorjward (changes requested):
Thanks for this!

ReubenHill:
Thanks for this! At the moment this is currently failing linting and regression tests. For linting you need to run `make lint` in the firedrake source directory with your firedrake venv active and correct the highlighted problems. For regression tests, take a look at the test output that our Continuous Integration (CI) system gives you via this github page. You'll see that
```
tests/regression/test_fs_caching.py::test_extruded_mixed_fs_hits_cache
```
fails. You'll need to work out why it fails and fix the problem. Good luck and thanks again!

connorjward:
This doesn't look like a particularly nice bug to track down. For some reason the shared data is no longer equal ([here](https://github.com/firedrakeproject/firedrake/blob/54d0d21cb69626219732eb308b852c2f97c59733/firedrake/functionspacedata.py#L451)). I recommend investigating `__slots__` and finding the discrepancy.
Also note that some more `FutureWarnings` are being emitted so you must have missed some `split` calls.

uygarkov:
> This doesn't look like a particularly nice bug to track down. For some reason the shared data is no longer equal ([here](https://github.com/firedrakeproject/firedrake/blob/54d0d21cb69626219732eb308b852c2f97c59733/firedrake/functionspacedata.py#L451)). I recommend investigating `__slots__` and finding the discrepancy. Also note that some more `FutureWarnings` are being emitted so you must have missed some `split` calls.

Thanks for the feedback, I will come up with a solution.

connorjward (dismissed):
Looks all good to me. Thanks so much for taking this on.

firedrake/functionspaceimpl.py
```diff
@@ -120,16 +121,17 @@ def ufl_cell(self):
 
     @PETSc.Log.EventDecorator()
     def split(self):
-        r"""Split into a tuple of constituent spaces."""
-        return self._split
+        import warnings
+        warnings.warn(".split() method was changed to .subfunctions property (without paranthesis)", category=FutureWarning)
```
wence-, line 125:
```suggestion
        warnings.warn("The .split() method is deprecated, please use the .subfunctions property instead", category=FutureWarning)
```

firedrake/functionspaceimpl.py
```diff
@@ -496,10 +498,16 @@ def __str__(self):
...
 
-    def split(self):
+    @utils.cached_property
+    def subfunctions(self):
         r"""Split into a tuple of constituent spaces."""
         return (self, )
 
+    def split(self):
+        import warnings
+        warnings.warn(".split() method was changed to .subfunctions property (without paranthesis)", category=FutureWarning)
```
wence-, line 508:
```suggestion
        warnings.warn("The .split() method is deprecated, please use the .subfunctions property instead", category=FutureWarning)
```

firedrake/functionspaceimpl.py
```diff
@@ -705,11 +713,17 @@ def __ne__(self, other):
...
-    def split(self):
+    @utils.cached_property
+    def subfunctions(self):
         r"""The list of :class:`FunctionSpace`\s of which this
         :class:`MixedFunctionSpace` is composed."""
         return self._spaces
 
+    def split(self):
+        import warnings
+        warnings.warn(".split() method was changed to .subfunctions property (without paranthesis)", category=FutureWarning)
```
wence-, line 724:
```suggestion
        warnings.warn("The .split() method is deprecated, please use the .subfunctions property instead", category=FutureWarning)
```

wence- (commented):
A minor proposal in wording change for the warning, otherwise looks great!

uygarkov:
> A minor proposal in wording change for the warning, otherwise looks great!

Thanks for the feedback, I applied the changes.

connorjward (approved):
We need to make sure that we squash before merging.

firedrake/constant.py
```diff
@@ -111,9 +111,15 @@ def function_space(self):
...
         return None
 
-    def split(self):
+    @utils.cached_property
+    def subfunctions(self):
         return (self,)
 
+    def split(self):
+        import warnings
+        warnings.warn(".split() method was changed to .subfunctions property (without paranthesis)", category=FutureWarning)
```
ReubenHill, line 120: Update to wording needed here too
```python
warnings.warn("The .split() method is deprecated, please use the .subfunctions property instead", category=FutureWarning)
```

firedrake/function.py
```diff
@@ -106,16 +106,18 @@ def ufl_id(self):
...
                      for i, (fs, dat) in
                      enumerate(zip(self.function_space(), self.dat)))
 
     @PETSc.Log.EventDecorator()
     def split(self):
-        r"""Extract any sub :class:`Function`\s defined on the component spaces
-        of this this :class:`Function`'s :class:`.FunctionSpace`."""
-        return self._split
+        import warnings
+        warnings.warn(".split() method was changed to .subfunctions property (without paranthesis)", category=FutureWarning)
```
ReubenHill, line 119: and here
```python
warnings.warn("The .split() method is deprecated, please use the .subfunctions property instead", category=FutureWarning)
```

ReubenHill (approved):

PR #2728, opened 2023-01-23
Projector: matfree solver params

pbrubeck:
Projection will fail to use the default solver (incomplete Cholesky) when `mat_type != "aij"`. This PR addresses this by setting Jacobi as the default matrix-free preconditioner for projection.

### DIFF
--- a/firedrake/projection.py
+++ b/firedrake/projection.py
@@ -100,7 +100,11 @@ class ProjectorBase(object, metaclass=abc.ABCMeta):
         solver_parameters.setdefault("ksp_type", "cg")
         solver_parameters.setdefault("ksp_rtol", 1e-8)
         mat_type = solver_parameters.get("mat_type", firedrake.parameters["default_matrix_type"])
-        if mat_type == "matfree":
+        if mat_type == "nest":
+            solver_parameters.setdefault("pc_type", "fieldsplit")
+            solver_parameters.setdefault("fieldsplit_pc_type", "bjacobi")
+            solver_parameters.setdefault("fieldsplit_sub_pc_type", "icc")
+        elif mat_type == "matfree":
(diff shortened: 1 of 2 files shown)


firedrake/projection.py
```diff
@@ -99,8 +99,14 @@ def __init__(self, source, target, bcs=None, solver_parameters=None,
...
-        solver_parameters.setdefault("pc_type", "bjacobi")
-        solver_parameters.setdefault("sub_pc_type", "icc")
+        mat_type = solver_parameters.get("mat_type", firedrake.parameters["default_matrix_type"])
+        if mat_type == "nest":
+            mat_type = solver_parameters.get("sub_mat_type", firedrake.parameters["default_sub_matrix_type"])
+        if mat_type == "matfree":
+            solver_parameters.setdefault("pc_type", "jacobi")
+        else:
+            solver_parameters.setdefault("pc_type", "bjacobi")
+            solver_parameters.setdefault("sub_pc_type", "icc")
```
wence-, line 109: I think this is wrong if the matrix is actually nested. The test you modify doesn't exercise this since the function space for the projection is not mixed.

pbrubeck: Do we support `{'mat_type': 'nest', 'sub_mat_type': 'matfree'}`? Although this would be the same as `mat_type = matrfee`

wence-: Probably not, but does `-mat_type nest -sub_mat_type aij -pc_type bjacobi -sub_pc_type icc` actually work for a nest matrix?

pbrubeck: These options fail for a nest matrix. I am adding a fieldsplit solver with bjacobi/icc along with a test.

PR #2732, opened 2023-01-24
Add gmsh to docker image

JDBetteridge:
~DO NOT MERGE (yet)~ 

I think this is ready to go

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -15,8 +15,29 @@ concurrency:
   cancel-in-progress: true
 
 jobs:
+  docker_test:
+    name: "Build Docker environment container"
+    runs-on: self-hosted
+    steps:
+      - name: Check out the repo
+        uses: actions/checkout@v2
+      - name: Log in to Docker Hub
+        uses: docker/login-action@v1
+        with:
+          username: ${{ secrets.DOCKERHUB_USER }}
+          password: ${{ secrets.DOCKERHUB_TOKEN }}
+      - name: Set up Docker Buildx
+        id: buildx
+        uses: docker/setup-buildx-action@v1
+      - name: Build and push firedrake-env
+        uses: docker/build-push-action@v2
+        with:
+          push: true
+          file: docker/Dockerfile.env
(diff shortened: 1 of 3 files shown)


.github/workflows/build.yml
```diff
@@ -70,6 +70,27 @@ jobs:
         run: |
           cd ..
           rm -rf build
+  docker_test:
+    name: "Build Docker environment container"
```
connorjward, line 74: This is the same comment as for `docker_env`. What's the difference? (and maybe make it clear in a comment or something)

JDBetteridge: This is temporary, I just need a way of generating a docker image that isn't one that CI or users use

demos/linear-wave-equation/linear_wave_equation.py.rst
```diff
@@ -107,7 +107,7 @@ options at this point, we may either `lump` the mass, which reduces
 the inversion to a pointwise division::
 
       if lump_mass:
-          p += assemble(dt * inner(nabla_grad(v), nabla_grad(phi))*dx) / assemble(v*dx)
+          p += interpolate(assemble(dt * inner(nabla_grad(v), nabla_grad(phi))*dx) / assemble(v*dx), V)
```
JDBetteridge, line 110: @connorjward can you update me on what assign is doing now. Do we still need this fix?

connorjward: I haven't made any recent changes to `assign`. I got quite confused when the adjoint got involved so stopped working on it.

JDBetteridge: In which case can we accept this as a temporary fix? If so, I'll create an issue documenting the issue and tidy up this PR

JDBetteridge:
NB: This will need to go through the wash twice before the tests that require gmsh will run

.github/workflows/build.yml
JDBetteridge: Gone

dham (approved):

PR #2740, opened 2023-01-27
[DO NOT MERGE] Test updated PyOP2 types without __getattr__ 

Straggio:


connorjward:
This is unlikely to get merged now that PyOP2 is on the cusp of deprecation.

PR #2746, opened 2023-01-31
Add domain warning to `Constant`

JDBetteridge:
Update all the constants in Firedrake and emit a Warning that users must add a domain whenever a constant is used.


danshapero:
I know this has been discussed before but can you remind me what the rationale is?

JDBetteridge:
It goes back to the garbage collection in PETSc/petsc4py. To avoid deadlock, petsc4py objects are stashed as attributes on the the associated communicator, then at some synchronised future time these are all collectively cleaned up.

In Firedrake land if we have a `Constant` object, this too has PETSc data structures associated with it. Prior to this `COMM_WORLD` was the assumed communicator if no domain was provided, but this is dangerous (and known to cause deadlocks), so now the communicator is pulled from the associated mesh. `domain` has been a kwarg and is part of UFL, it just wasn't previously required, this seems cleaner than asking a user to supply an additional `comm` argument. This also "hides the comm" from the user in the same manner as it is during mesh creation.

JDBetteridge:
@danshapero In case you are working on updating your code, we may not be pushing forward with this change, we are working on alternative fix.

JDBetteridge:
This was resolved in
https://github.com/firedrakeproject/firedrake/pull/2927
without insisting that all `Constant`s have domains.

PR #2750, opened 2023-02-03
VertexOnlyMesh on ExtrudedMesh

ReubenHill:
Add support for extruded meshes. Also update tests to demonstrate that meshes whose coordinate fields have been shifted are in fact supported. Furthermore, ensure all `Constant`s have a domain.

### DIFF
--- a/firedrake/interpolation.py
+++ b/firedrake/interpolation.py
@@ -192,7 +192,12 @@ def make_interpolator(expr, V, subset, access):
                 # function space nodes on the source mesh. NOTE: argfs_map is
                 # allowed to be None when interpolating from a Real space, even
                 # in the trans-mesh case.
-                argfs_map = compose_map_and_cache(target_mesh.cell_parent_cell_map, argfs_map)
+                if source_mesh.extruded:
+                    # ExtrudedSet cannot be a map target so we need to build
+                    # this ourselves
+                    argfs_map = vom_cell_parent_node_map_extruded(target_mesh, argfs.cell_node_map())
+                else:
+                    argfs_map = compose_map_and_cache(target_mesh.cell_parent_cell_map, argfs_map)
         sparsity = op2.Sparsity((V.dof_dset, argfs.dof_dset),
                                 ((V.cell_node_map(), argfs_map),),
                                 name="%s_%s_sparsity" % (V.name, argfs.name),
@@ -348,8 +353,13 @@ def _interpolator(V, tensor, expr, subset, arguments, access):
             # Since the par_loop is over the target mesh cells we need to
             # compose a map that takes us from target mesh cells to the
             # function space nodes on the source mesh.
-            columns_map = compose_map_and_cache(target_mesh.cell_parent_cell_map,
-                                                columns_map)
+            if source_mesh.extruded:
+                # ExtrudedSet cannot be a map target so we need to build
+                # this ourselves
+                columns_map = vom_cell_parent_node_map_extruded(target_mesh, columns_map)
+            else:
+                columns_map = compose_map_and_cache(target_mesh.cell_parent_cell_map,
+                                                    columns_map)
         parloop_args.append(tensor(op2.WRITE, (rows_map, columns_map)))
     if oriented:
         co = target_mesh.cell_orientations()
@@ -368,7 +378,12 @@ def _interpolator(V, tensor, expr, subset, arguments, access):
                 # Since the par_loop is over the target mesh cells we need to
                 # compose a map that takes us from target mesh cells to the
                 # function space nodes on the source mesh.
-                m_ = compose_map_and_cache(target_mesh.cell_parent_cell_map, coefficient.cell_node_map())
+                if source_mesh.extruded:
+                    # ExtrudedSet cannot be a map target so we need to build
+                    # this ourselves
+                    m_ = vom_cell_parent_node_map_extruded(target_mesh, coefficient.cell_node_map())
+                else:
+                    m_ = compose_map_and_cache(target_mesh.cell_parent_cell_map, coefficient.cell_node_map())
             else:
                 # m_ is allowed to be None when interpolating from a Real space,
                 # even in the trans-mesh case.
@@ -478,6 +493,28 @@ def compose_map_and_cache(map1, map2):
         cmap = None if map2 is None else op2.ComposedMap(map2, map1)
         map1._cache[cache_key] = cmap
     return cmap
+
+
+def vom_cell_parent_node_map_extruded(vertex_only_mesh, extruded_cell_node_map):
+    """Build a map from the cells of a vertex only mesh to the nodes of the
+    source mesh's cell node map where the source mesh is extruded.
+
+    :arg vertex_only_mesh: The `mesh.VertexOnlyMesh` mesh whose cells we
+        iterate over.
+    :arg extruded_cell_node_map: The cell node map of the function space on the
+        extruded source mesh.
+
+    :returns: A :class:`PyOP2.Map` with iterset over the vertex only mesh cells
+        and toset over the source mesh nodes.
+    """
+    if not isinstance(vertex_only_mesh.topology, firedrake.mesh.VertexOnlyMeshTopology):
+        raise TypeError("The input mesh must be a VertexOnlyMesh")
+    cnm = extruded_cell_node_map
+    vmx = vertex_only_mesh
+    dofs_per_target_cell = cnm.arity
+    base_cells, heights = numpy.array([0, 1]), numpy.array([0, 2])  # Need to work this out for the general case
+    target_cell_parent_node_list = [[cnm.values[base_cell, dof_idx] + height*cnm.offset[dof_idx] for dof_idx in range(dofs_per_target_cell)] for base_cell, height in zip(base_cells, heights)]
+    return op2.Map(vmx.cell_set, cnm.toset, dofs_per_target_cell, target_cell_parent_node_list)
 
 
 class GlobalWrapper(object):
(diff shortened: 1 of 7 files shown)


firedrake/interpolation.py
```diff
@@ -192,7 +192,12 @@ def make_interpolator(expr, V, subset, access):
                 # function space nodes on the source mesh. NOTE: argfs_map is
                 # allowed to be None when interpolating from a Real space, even
                 # in the trans-mesh case.
-                argfs_map = compose_map_and_cache(target_mesh.cell_parent_cell_map, argfs_map)
+                if source_mesh.extruded:
+                    # ExtrudedSet cannot be a map target so we need to build
+                    # this ourselves
+                    argfs_map = vom_cell_parent_node_map_extruded(target_mesh, argfs.cell_node_map())
```
connorjward, line 198: `argfs_map` is already defined as `argfs.cell_node_map()`. You don't need the function call here and can use the variable instead.

firedrake/interpolation.py
```diff
@@ -480,6 +495,29 @@ def compose_map_and_cache(map1, map2):
     return cmap
 
 
+def vom_cell_parent_node_map_extruded(vertex_only_mesh, extruded_cell_node_map):
```
connorjward, line 498: I think more documentation here describing why this is needed and how it works would be useful.

firedrake/mesh.py
```diff
@@ -1542,6 +1542,48 @@ def cell_parent_cell_map(self):
...
+        mesh cell order.
+        """
+        if not isinstance(self._parent_mesh, ExtrudedMeshTopology):
+            raise AttributeError("Parent mesh is not extruded")
+        cell_parent_base_cell_list = np.copy(self.topology_dm.getField("parentcellbasenum"))
+        self.topology_dm.restoreField("parentcellbasenum")
+        return cell_parent_base_cell_list
+
+    @property
+    def cell_parent_base_cell_map(self):
```
connorjward, line 1557: Should these be cached properties? They look quite expensive to compute.

firedrake/mesh.py
```diff
@@ -2479,20 +2512,60 @@ def _pic_swarm_in_mesh(parent_mesh, coords, fields=None, tolerance=None):
...
     coords, reference_coords, parent_cell_nums = \
         _parent_mesh_embedding(coords, parent_mesh, tolerance)
-    # mesh.topology.cell_closure[:, -1] maps Firedrake cell numbers to plex numbers.
-    plex_parent_cell_nums = parent_mesh.topology.cell_closure[parent_cell_nums, -1]
+
+    if parent_mesh.extruded:
+        # need to store the base parent cell number and the height to be able
+        # to map point coordinates back to the parent mesh
+        if parent_mesh.variable_layers:
+            # TODO: Don't know how variable layer cell numbering works - below
```
connorjward, line 2520: I think the restriction here is that `locate_cell` doesn't know what to do. Determining the right base cell and height should still be relatively straightforward.

ksagiyam: Yes. We should probably just let `locate_cell` return `base_cell` and `height`; indeed these variables have once been computed in `locate_cell.c` (assuming uniform layer extrusion).
If `parent_mesh.variable_layers = True`, `locate_cell` should complain.

firedrake/interpolation.py
```diff
@@ -480,6 +495,125 @@ def compose_map_and_cache(map1, map2):
...
+    cnm = extruded_cell_node_map
+    vmx = vertex_only_mesh
+    dofs_per_target_cell = cnm.arity
+    base_cells = vmx.cell_parent_base_cell_list
+    heights = vmx.cell_parent_extrusion_height_list
+    target_cell_parent_node_list = [
+        [
+            cnm.values[base_cell, dof_idx] + height * cnm.offset[dof_idx]
+            for dof_idx in range(dofs_per_target_cell)
+        ]
```
wence-, line 609:
```suggestion
        cnm.values[base_cell, :] + height * cnm.offset[:]
```

wence-: And then you can maybe do more broadcasting as:
```
cnm.values[base_cells, :] + heights * cnm.offset[:]
```
?

Might need to use numpy.tile or similar.

ReubenHill: Agreed on your suggestion, but I think going any further loses readability

firedrake/mesh.py
ReubenHill: Okay I'll get rid of the todo

firedrake/mesh.py
ReubenHill: Interesting that `pyop2.utils.cached_property` doesn't use any of the code in `pyop2.caching`.

ReubenHill: For now I'll just use `pyop2.utils.cached_property` and add a todo note for when I have moving meshes

firedrake/interpolation.py
```diff
@@ -480,6 +495,124 @@ def compose_map_and_cache(map1, map2):
...
+        The ``mesh.VertexOnlyMesh`` whose cells we iterate over.
+    extruded_cell_node_map : :class:`PyOP2.Map`
+        The cell node map of the function space on the extruded mesh within
+        which the ``mesh.VertexOnlyMesh`` is immersed.
+
+    Returns
+    -------
+    :class:`PyOP2.Map`
+        The map from the cells of the vertex only mesh to the nodes of the
+        source mesh's cell node map. The map iterset is the
```
connorjward, line 514: Same as above.

firedrake/interpolation.py
```diff
@@ -480,6 +495,124 @@ def compose_map_and_cache(map1, map2):
...
+    ``extruded_cell_node_map.values[0]`` to get
+    ``extruded_cell_node_map.values[0] + 1 * extruded_cell_node_map.offset[0] =
+    [16, 17, 18,  2,  3,  4,  9, 10, 11]`` where ``0`` is the DoF/node index.
+
+    For each cell (vertex) of a vertex only mesh immersed in a parent
+    extruded mesh, we can can get the corresponding ``base_cell_num`` and
+    ``extrusion_height`` of the parent extruded mesh. Armed with this
+    information we use the above to work out the corresponding DoFs/nodes on
+    the parent extruded mesh.
+
```
connorjward, line 596: Lovely docstring.

connorjward (changes requested):
Generally looks very good. Just a few small comments from me.

firedrake/interpolation.py
```diff
@@ -480,6 +495,124 @@ def compose_map_and_cache(map1, map2):
     return cmap
 
 
+def vom_cell_parent_node_map_extruded(vertex_only_mesh, extruded_cell_node_map):
```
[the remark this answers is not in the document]
ReubenHill: We could, but I'm not quite sure where we would cache it so want to push that into a separate PR - it's an optimisation after all

firedrake/interpolation.py
```diff
@@ -480,6 +495,124 @@ def compose_map_and_cache(map1, map2):
...
+    ----------
+    vertex_only_mesh : :class:`mesh.MeshGeometry`
+        The ``mesh.VertexOnlyMesh`` whose cells we iterate over.
+    extruded_cell_node_map : :class:`PyOP2.Map`
+        The cell node map of the function space on the extruded mesh within
+        which the ``mesh.VertexOnlyMesh`` is immersed.
+
+    Returns
+    -------
+    :class:`PyOP2.Map`
```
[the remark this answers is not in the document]
ReubenHill: It's the same as in the existing `compose_map_and_cache` function but I notice elsewhere it's 
```
:class:`pyop2.Map`
```
I don't think the API documentation builds for these functions anyway since they aren't included in `__all__` at the top of the file.

firedrake/interpolation.py
ReubenHill: See reply above

firedrake/interpolation.py
```diff
@@ -480,6 +495,124 @@ def compose_map_and_cache(map1, map2):
     return cmap
 
 
+def vom_cell_parent_node_map_extruded(vertex_only_mesh, extruded_cell_node_map):
+    """Build a map from the cells of a vertex only mesh to the nodes of the
+    source mesh's cell node map where the source mesh is extruded.
```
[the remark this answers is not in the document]
ReubenHill: Done

firedrake/interpolation.py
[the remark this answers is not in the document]
ReubenHill: I'll rebase and change to 
```
:class:`pyop2.Map`
```
anyway

firedrake/interpolation.py
ReubenHill: Thanks!

firedrake/interpolation.py
```diff
@@ -480,6 +495,124 @@ def compose_map_and_cache(map1, map2):
...
+        5          12          19         26         33
+        |                       |                     |
+        4 ---------11----------18---------25---------32
+        |                       |                     |
+        3          10          17         24         31
+        |                       |                     |
+        2 ---------9-----------16---------23---------30
+        |                       |                     |
+        1          8           15         22         29
+        | base_cell_num = 0     |  base_cell_num = 1  |
```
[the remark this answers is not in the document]
ReubenHill: Fixed

firedrake/interpolation.py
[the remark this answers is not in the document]
connorjward: Sure.

connorjward (approved):
AFAICT all looks good.

PR #2756, opened 2023-02-07
SNESSolve allows for a nonzero RHS in F(x) = b. Allow a user to pass one in.

pefarrell:
This is useful for FAS.

### DIFF
--- a/firedrake/variational_solver.py
+++ b/firedrake/variational_solver.py
@@ -242,7 +242,7 @@ class NonlinearVariationalSolver(OptionsManager, NonlinearVariationalSolverMixin
 
     @PETSc.Log.EventDecorator()
     @NonlinearVariationalSolverMixin._ad_annotate_solve
-    def solve(self, bounds=None):
+    def solve(self, bounds=None, rhs=None):
         r"""Solve the variational problem.
 
         :arg bounds: Optional bounds on the solution (lower, upper).
@@ -267,6 +267,17 @@ class NonlinearVariationalSolver(OptionsManager, NonlinearVariationalSolverMixin
             lower, upper = bounds
             with lower.dat.vec_ro as lb, upper.dat.vec_ro as ub:
                 self.snes.setVariableBounds(lb, ub)
+
+        rhsv = None
+        if rhs is not None:
+            try:
+                [bc.zero(rhs) for bc in self._problem.bcs]
+            except TypeError:
+                pass
+
+            with rhs.dat.vec_ro as rhs_:
+                rhsv = rhs_
+
         work = self._work
         with self._problem.u.dat.vec as u:
             u.copy(work)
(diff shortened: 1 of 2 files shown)


firedrake/variational_solver.py
```diff
@@ -242,7 +242,7 @@ def set_transfer_manager(self, manager):
 
     @PETSc.Log.EventDecorator()
     @NonlinearVariationalSolverMixin._ad_annotate_solve
-    def solve(self, bounds=None):
+    def solve(self, bounds=None, rhs=None):
```
dham, line 245: Is this the right API, or should the RHS be attached to the `NonlinearVariationalProblem`? It seems odd to add symbolic information about the problem at this late stage.

dham: Additionally, this must surely require a corresponding adjoint change.

firedrake/variational_solver.py
```diff
@@ -267,6 +267,17 @@ def solve(self, bounds=None):
             lower, upper = bounds
             with lower.dat.vec_ro as lb, upper.dat.vec_ro as ub:
                 self.snes.setVariableBounds(lb, ub)
+
+        rhsv = None
+        if rhs is not None:
+            try:
+                [bc.apply(rhs) for bc in self._problem.bcs]
```
pbrubeck, line 274: Should we not have `bc.zero(rhs)` here?

tests/regression/test_solve_rhs.py
```diff
@@ -0,0 +1,25 @@
...
+    u1 = Function(V)
+    u2 = Function(V)
+    v = TestFunction(V)
+    f = Constant(1)
+
+    F1 = inner(grad(u1), grad(v))*dx - inner(f, v)*dx
+    F2 = inner(grad(u2), grad(v))*dx
+    rhs = assemble(inner(f, v)*dx)
+
+    bcs = DirichletBC(V, 0, "on_boundary")
```
pbrubeck, line 16: Might want to use non-homogeneous BCs for the test

firedrake/variational_solver.py
pefarrell: You're absolutely right, thanks!

tests/regression/test_solve_rhs.py
pefarrell: Agreed, done.

firedrake/variational_solver.py
pefarrell: I agree about the adjoint change.

My use case: I'm implementing FAS algorithms for variational inequalities, and I'm setting up firedrake `NonlinearVariationalProblem`s and `NonlinearVariationalSolver`s to do the smoothing on each level and on the coarse grid. FAS constructs a 'correction' to the nonlinear problem on each grid, to ensure that the solution of the corrected coarse grid problem is the injection of the fine grid solution. These are the right-hand sides I want to supply here, and are as far as I know the only common use case for passing something nonzero to `SNESSolve`. Since it changes each iteration, it would be less convenient to make it a property of the `NonlinearVariationalProblem`.

PR #2763, opened 2023-02-09
Refactor parallel test markers

connorjward:
I need to implement parallel tests for pyop3 so I think it makes sense to have a [separate package](https://github.com/connorjward/pytest-mpi) (should be moved into firedrakeproject) containing the pytest plugin. That way other packages can use the functionality without needing to copy our `conftest.py`.

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -2,39 +2,9 @@
 
 import pytest
 
-from subprocess import check_call
-from pyadjoint.tape import get_working_tape
-from firedrake.utils import complex_mode
-
-
-def parallel(item):
-    """Run a test in parallel.
-
-    :arg item: The test item to run.
-    """
-    from mpi4py import MPI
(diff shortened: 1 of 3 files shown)


tests/conftest.py
```diff
@@ -45,40 +18,35 @@ def pytest_configure(config):
...
+# this is called for every test prior to setup/calling etc
+# def pytest_runtest_setup(item):
+#     if item.get_closest_marker("parallel"):
+#         if MPI.COMM_WORLD.size > 1:
+#             # Turn on source hash checking
+#             # is this really needed? why not have this by default?
+#             from firedrake import parameters
+#             from functools import partial
+#
+#             def _reset(check):
```
connorjward, line 30: I will tackle this commented out bit next.

JDBetteridge (approved):
Largely just removing code.
I do think there should be (at least minimal) testing in the `pytest-mpi` repo though.

PR #2776, opened 2023-02-15
Smaclachlan/add asm vanka exclude

ScottMacLachlan:
Excluding subspaces from a Vanka patch is key to achieving good performance for some mixed problems, such as Taylor-Hood Stokes.  This adds that functionality to `ASMVankaPC`, as well as a corresponding test on the Taylor-Hood discretization of Stokes to make sure that it has the same effect as the option to do so in `PCPatch`.

### DIFF
--- a/firedrake/preconditioners/asm.py
+++ b/firedrake/preconditioners/asm.py
@@ -203,6 +203,7 @@ class ASMVankaPC(ASMPatchPC):
         if (depth == -1 and height == -1) or (depth != -1 and height != -1):
             raise ValueError(f"Must set exactly one of {self.prefix}construct_dim or {self.prefix}construct_codim")
 
+        exclude_subspace = int(PETSc.Options().getString(self.prefix+"exclude_subspace", default=-1))
         ordering = PETSc.Options().getString(self.prefix+"mat_ordering_type", default="natural")
         # Accessing .indices causes the allocation of a global array,
         # so we need to cache these for efficiency
@@ -234,7 +235,11 @@ class ASMVankaPC(ASMPatchPC):
             indices = []
             for (i, W) in enumerate(V):
                 section = W.dm.getDefaultSection()
-                for p in pt_array:
+                if i == exclude_subspace:
+                    loop_list = [seed]
+                else:
+                    loop_list = pt_array
+                for p in loop_list:
                     dof = section.getDof(p)
                     if dof <= 0:
                         continue
--- a/tests/regression/test_star_pc.py
+++ b/tests/regression/test_star_pc.py
@@ -301,7 +301,7 @@ def test_vanka_equivalence(problem_type):
         (z, p) = split(u)
         (v, q) = split(TestFunction(V))
 
-        a = inner(grad(z), grad(v))*dx + inner(p, q)*dx
+        a = inner(grad(z), grad(v))*dx + inner(p, div(v))*dx + inner(q,div(z))*dx
 
         bcs = DirichletBC(V.sub(0), Constant((1., 0.)), "on_boundary")
         nsp = MixedVectorSpaceBasis(V, [V.sub(0), VectorSpaceBasis(constant=True)])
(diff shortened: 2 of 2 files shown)


colinjcotter:
Can you explain what subspaces are being excluded?

On 15 Feb 2023, at 20:14, ScottMacLachlan ***@***.***> wrote:

﻿

Excluding subspaces from a Vanka patch is key to achieving good performance for some mixed problems, such as Taylor-Hood Stokes. This adds that functionality to ASMVankaPC, as well as a corresponding test on the Taylor-Hood discretization of Stokes to make sure that it has the same effect as the option to do so in PCPatch.

You can view, comment on, or merge this pull request online at:

  https://github.com/firedrakeproject/firedrake/pull/2776

Commit Summary

  *   ad9f52b<https://github.com/firedrakeproject/firedrake/pull/2776/commits/ad9f52b613bda1bfea7887e811e1679ce5a59825> Adding exclude_subspace to ASMVankaPC
  *   9d5b182<https://github.com/firedrakeproject/firedrake/pull/2776/commits/9d5b1826e3df13dc1dacc2954fd148b868134c2a> More meaningful to test with Stokes
  *   73f722c<https://github.com/firedrakeproject/firedrake/pull/2776/commits/73f722c18e1bb7dc6f230add18dccdaa7c9eeeed> lint

File Changes

(2 files<https://github.com/firedrakeproject/firedrake/pull/2776/files>)

  *   M firedrake/preconditioners/asm.py<https://github.com/firedrakeproject/firedrake/pull/2776/files#diff-b2a07bc405f1662febb80f3b024919b7878b5aaa4cd8173744e6d56bf7d391af> (7)
  *   M tests/regression/test_star_pc.py<https://github.com/firedrakeproject/firedrake/pull/2776/files#diff-a5ce070873b1b31dbc280623d17d567924531c24894ad1c84c5cc0aa72cbfc39> (58)

Patch Links:

  *   https://github.com/firedrakeproject/firedrake/pull/2776.patch
  *   https://github.com/firedrakeproject/firedrake/pull/2776.diff

ScottMacLachlan:
Patch construction is topological - Vanka patches are defined as the closure of the star of some mesh entities.

For the (Navier-)Stokes equations, _the_ standard patch construction for monolithic MG is to take each pressure DoF and form a patch from that single pressure DoF and all coupled velocity DoFs.  That's not equal to the DoF set that you get when you take the topological Vanka patch, since that includes pressure DoFs, particularly at mesh entities that are in the closure of the star, but not the star itself (at least for Taylor-Hood discretizations).  What this does is allow you to specify a subspace for which you only take DoFs at the mesh entity around which you're constructing the patch.  For (N)SE, that means that you get the (single?) pressure at that topological entity plus all velocities in the closure of the star, which is exactly what you want.  I.e., we are "excluding" pressure DoFs aside from at the seed topological entity.

colinjcotter:
Thanks for explaining! I will have a play with it.

________________________________________

firedrake/preconditioners/asm.py
```diff
@@ -234,7 +235,11 @@ def get_patches(self, V):
             indices = []
             for (i, W) in enumerate(V):
                 section = W.dm.getDefaultSection()
-                for p in pt_array:
+                if i == exclude_subspace:
```
pefarrell, line 238: Will we ever want to exclude more than one subspace?

tests/regression/test_star_pc.py
```diff
@@ -290,7 +290,64 @@ def test_vanka_equivalence(problem_type):
...
+        mesh = mh[-1]
+        V1 = VectorFunctionSpace(mesh, "CG", 2)
+        V2 = FunctionSpace(mesh, "CG", 1)
+        V = MixedFunctionSpace([V1, V2])
+
+        u = Function(V)
+        (z, p) = split(u)
+        (v, q) = split(TestFunction(V))
+
+        a = inner(grad(z), grad(v))*dx + inner(p, div(v))*dx + inner(q, div(z))*dx
```
pefarrell, line 304: The two last terms would usually have a minus, but whatever (only changes sign of pressure)

pefarrell (dismissed):

firedrake/preconditioners/asm.py
ScottMacLachlan: We certainly might.. Fixed now

tests/regression/test_star_pc.py
ScottMacLachlan: Fair enough.  Changed this while fixing the above

ScottMacLachlan:
I must admit that I never understood why this test was failing in the GitHub actions test, since it seemed sensible and passed locally for me every time that I ran it.  Now I don't understand why it's passing, unless some update to MUMPS or our default parameters has resolved what I didn't understand in the first place...

@pefarrell : Care to complete the review for this?

dham (approved):

PR #2786, opened 2023-02-24
Add ASMPointPC

ScottMacLachlan:
Following a slack discussion on how it would be nice to do block Jacobi on mixed function spaces, defining blocks based on grouping DoFs at the same point(s) on the mesh, this seems like a potentially useful function.  This provides another variant on `ASMPatchPC`, where the patches are defined based on a specified topological entity (nodes, edges, faces, volumes).  So, not quite point-block Jacobi that we can do on VFS, but closer to it...

### DIFF
--- a/firedrake/preconditioners/asm.py
+++ b/firedrake/preconditioners/asm.py
@@ -143,8 +143,6 @@ class ASMPointPC(ASMPatchPC):
 
         # Obtain the topological entities to use to construct the patches
         depth = PETSc.Options().getInt(self.prefix+"construct_dim", default=0)
-        ordering = PETSc.Options().getString(self.prefix+"mat_ordering_type",
-                                             default="natural")
         # Accessing .indices causes the allocation of a global array,
         # so we need to cache these for efficiency
         V_local_ises_indices = []
@@ -240,7 +238,6 @@ class ASMVankaPC(ASMPatchPC):
     ASMVankaPC is an additive Schwarz preconditioner where each patch
     consists of all DoFs on the closure of the star of the mesh entity
     specified by `pc_vanka_construct_dim` (or codim).
-
     '''
 
     _prefix = "pc_vanka_"
--- a/firedrake/preconditioners/asm.py
+++ b/firedrake/preconditioners/asm.py
@@ -14,7 +14,7 @@ except ImportError:
     have_tinyasm = False
 
 
-__all__ = ("ASMPatchPC", "ASMStarPC", "ASMVankaPC", "ASMLinesmoothPC", "ASMExtrudedStarPC")
+__all__ = ("ASMPatchPC", "ASMPointPC", "ASMStarPC", "ASMVankaPC", "ASMLinesmoothPC", "ASMExtrudedStarPC")
 
 
 class ASMPatchPC(PCBase):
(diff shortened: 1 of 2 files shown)


pefarrell:
I'm not sure a `construct_dim` is the right paradigm. If you had P3, wouldn't you want to do point-block Jacobi for all entities, not just entities of a specified dimension? (I see in the test you compose them with PCCOMPOSITE, but this is a pain.) Maybe the construct_dim could become a list of dimensions of things to build off, with the default being all dimensions?

rckirby:
Like Patrick, I'm nervous about giving something a name that isn't quite right (e.g. beyond P^2).  Much like `jacobi` is of limited use in mixed problems, maybe we only want to support the case of a several-way product `V * V * ... * V` of some non-mixed space `V`.  Then, getting the right dofs should just be strided access of size dimension of `V`.  This might be even simpler than going through a plex and doesn't encounter the complication of accidentally coupling different dofs on the same facet?   (This only seems practical if we can assume that each copy of `V` in the mixed space has the same decomposition across processors... Is that true?).

JHopeCollins:
If this were to be restricted to repeated mixed spaces, it would be useful to also cover the case with multiple repetitions such as `U * V * U * V * ... * U * V` or `U * V * W * ...` etc. If its just implemented as strided access, then I imagine (hope) that this wouldn't need much extra work.

ScottMacLachlan:
I think Rob and Josh are identifying the correct two paradigms to consider here.  Sometimes, you just want to do this for generalizing point Jacobi on `V` to block Jacobi on `V x V x V x ... x V`, and then you should be able to do that using striding across the copies of `V` (since we know something about the layout of those copies).  Other times, you want to do this for coupling DoFs across mixed spaces, `U x V`, or copies of this (`U x V x U x V x ...`).  This is less clear, since `U` and `V` may or may not have DoFs that align precisely (meaning, for example, that they both have point-evaluation dual functionals evaluating at the same point on the topological entity), but you can always ask whether or not they have DoFs associated with the same topological entity.

I have a use case in mind for the latter (`U` is, for example, vector P2, and `V` is N1 or BDM1, so I have `U` DoFs on the nodes, and both `U` and `V` DoFs on the edges), and that's what I was aiming at here.  I know the Irksome use case exists for the former, so maybe we need both?

rckirby:
My understanding is that point Jacobi is not typically used in mixed
contexts (maybe I'm used to saddle point problems where it obviously
fails?).  There is some ambiguity in the case of underlying `U * V` -- one
can't promise that the two spaces have dofs at the same spaces to couple
together.  However, one could attempt to extract the diagonal of each 2x2
block and then "monolithicize" over the repetition.

rckirby:
Scott, I do think we may in fact be aiming at different things, and it
might in fact be worth having both of them, even if they give the same
method for P^1 or P^2.

JHopeCollins:
My original thought was that in the space `U * V * U * V * ...`, you could group the DoFs for all `U` and group the DoFs for all `V`, so it would look more an interleaved version of what Rob would like, instead of coupling DoFs from `U` and `V`.

JHopeCollins:
Which I think would be the equivalent of using `pbjacobi` for `MixedFunctionSpace(VectorFunctionSpace(U, U, ...), VectorFunctionSpace(V, V, ...), ...)`

wence-:
> (This only seems practical if we can assume that each copy of `V` in the mixed space has the same decomposition across processors... Is that true?).

This is true.

colinjcotter:
Is there any chance to revive this PR?

ScottMacLachlan:
This may be stale enough that we should just start fresh (given recent changes in `asm.py`), but I have no objection to bringing this back.  @pbrubeck probably has the best perspective on the easiest way to modernize what's here...

PR #2787, opened 2023-02-24
Ksagiyam/io do not use pickle

ksagiyam:
Use `repr` instead of pickling.

Involves a minor fix in UFL, so depends on https://github.com/firedrakeproject/ufl/pull/34

### DIFF
--- a/firedrake/bcs.py
+++ b/firedrake/bcs.py
@@ -5,7 +5,7 @@ import functools
 import itertools
 
 import ufl
-from ufl import as_ufl, UFLException, as_tensor, VectorElement
+from ufl import as_ufl, as_tensor, VectorElement
 import finat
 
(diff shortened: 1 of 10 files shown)


firedrake/bcs.py
```diff
@@ -347,11 +347,11 @@ def function_arg(self, g):
             try:
                 g = as_ufl(g)
                 self._function_arg = g
-            except UFLException:
+            except BaseException:
```
dham, line 350: `ValueError`

dham (changes requested):
minor quibble.

firedrake/bcs.py
ksagiyam: Fixed.

ksagiyam:
Merged as discussed.

PR #2796, opened 2023-03-06
Metric-based mesh adaptation (take 2)

joewallwork:
Have been working on this again recently and made good progress on new branches. Parallel mesh adaptation is working for the tests considered here, but the PETSc branch `jwallwork23/parmmg-rebased` needs a bit more extension and then merging before we can merge this. Many thanks to @stephankramer, who made great progress on the PETSc side.

PETSc merge aside, it would be great to discuss the design/functionality/user-friendliness of the code here. Adding @ksagiyam, @danshapero, and @stephankramer as reviewers to ask if any of you have any further comments or suggestions on top of what we discussed in the previous PR #2295? Can't add @jcallaham (not in the Firedrake organisation?) but any comments would be appreciated. Thanks!

### DIFF
--- a/tests/meshadapt/test_metric.py
+++ b/tests/meshadapt/test_metric.py
@@ -260,3 +260,23 @@ def test_complexity(dim):
         expected = 8 / 27 * (22 * np.sqrt(2) - 25)
     metric.interpolate(as_matrix(mat))
     assert np.isclose(metric.complexity(), expected)
+
+
+@pytest.mark.skipcomplex
+def test_enforce_spd_a_max(dim):
+    """
+    Tests that the :meth:`enforce_spd` method applies maximum anisotropy as expected.
+    """
+    mesh = uniform_mesh(dim)
+    P1_ten = TensorFunctionSpace(mesh, "CG", 1)
+    metric = RiemannianMetric(P1_ten)
+    M = np.eye(dim)
+    M[0][0] = 10.0
+    metric.interpolate(as_matrix(M))
+    mp = {"dm_plex_metric_a_max": 1.0}
+    metric.set_parameters(mp)
+    metric.enforce_spd(restrict_anisotropy=True)
+
+    expected = Function(P1_ten)
+    expected.interpolate(Identity(dim))
+    assert np.isclose(errornorm(metric, expected), 0)


jwallwork23:
Any ideas what's going wrong with the build? I'm seeing lots of `Skipping link: none of the wheel's tags` and `Skipping link: unsupported archive format: .exe` warnings.

connorjward:
Those warnings show up because we run `pip` in hyper-verbose mode. They are standard. The error occurs when building slepc:

```
2023-03-06T08:58:24.3329126Z /__w/firedrake/build/src/slepc/src/sys/classes/fn/interface/fnbasic.c: In function ‘FNDuplicate’:
2023-03-06T08:58:24.3329497Z /__w/firedrake/build/src/slepc/src/sys/classes/fn/interface/fnbasic.c:1013:23: error: ‘PETSC_SUCCESS’ undeclared (first use in this function); did you mean ‘EXIT_SUCCESS’?
2023-03-06T08:58:24.3329607Z  1013 |   PetscFunctionReturn(PETSC_SUCCESS);
```

Maybe we need to update our SLEPc fork.

firedrake/meshadapt.py
```diff
@@ -0,0 +1,500 @@
...
+
+    :param mesh: :class:`~firedrake.mesh.MeshGeometry` to be adapted.
+    :param metrics: list of :class:`.RiemannianMetric`\s
+    :return: a new :class:`~firedrake.mesh.MeshGeometry`.
+    """
+    metric = metrics[0]
+    if len(metrics) > 1:
+        metric.intersect(*metrics[1:])
+    adaptor = MetricBasedAdaptor(mesh, metric)
+    return adaptor.adapted_mesh
```
ksagiyam, line 500: Would it be possible to hide `RiemannianMetric` from users? I would expect some interface like:
```
mesh = ...
V = FunctionSpace(mesh, ...)
f = Function(V)  # some scalar field
h = Function(V).interpolate(...)  # another field
g0 = compute_hessian_from_scalar_field(f)  # a utility function; returns a tensor P1 function
g1 = ...  # define another tensor P1 function if we want
amsh = AdaptedMesh(mesh, [g0, g1], average=False, normalise=False, name="my_adapted_mesh")
Vadapt = FunctionSpace(amsh, ...)
hadapt = Function(Vadapt).interpolate(h)  # Use the regular interface for interpolation (and projection).
```
In the last line we know if we are interpolating across parent mesh-adapted mesh if we let `amsh` carry the adaptor object as attribute (e.g., `amsh.adapted_adaptor`, where `adapted_` is a prefix).

I cannot think through all possible use cases, so would like to hear what others think.

ksagiyam (changes requested):
Thanks, Joe! I'm not sure what the best design/interface would be, but left some comments.

firedrake/meshadapt.py
connorjward: I don't think it's a very nice pattern to have a constructor (or function masquerading as a constructor) that takes another instance of the same class as a constructor parameter. I think `adapt(mesh, ...)` is a better interface than `AdaptedMesh(mesh, ...)`.

firedrake/meshadapt.py
```diff
@@ -0,0 +1,500 @@
...
+import firedrake.mesh as fmesh
+from firedrake.petsc import PETSc, OptionsManager
+import firedrake.utils as futils
+import numpy as np
+import ufl
+
+__all__ = ["RiemannianMetric", "MetricBasedAdaptor", "adapt"]
+
+
+class RiemannianMetric(ffunc.Function):
```
connorjward, line 15: Can you explain why you are inheriting from a `Function` here? It looks like you are doing a lot of complicated logic here like modifying the underlying plex that makes me think that a subclass isn't the right approach. Could this class not just have a `function` property instead? (i.e. composition instead of inheritance).

firedrake/meshadapt.py
ksagiyam: I think `AdaptedMesh` is more consistent with other similar functions. We already have `ExtrudedMesh`, `VertexOnlyMesh`, and now `RelabeledMesh`, which have similar interface.

connorjward: I think the difference I see between this and an `ExtrudedMesh` or `VertexOnlyMesh` is that those functions need to remember their initial meshes. In other words it makes sense for them to retain `mesh.base_mesh` or `mesh.parent_mesh` properties. Is it ever the case that we need `adapted_mesh.original_mesh`? Arguably `RelabeledMesh` should be `relabel(mesh, ...)` as it (AIUI) fits the same paradigm.

ksagiyam: I think it's fair to store parent mesh as long as the derived meshes are related to the parent mesh; in this specific case, adapted meshes represent the same domain as the parent mesh, which is important.

wence-: I think, when we discussed this many years ago, we had a concept of a `MeshAdaptor` as joe introduces. The idea being you need both meshes lying around until such time as you've transferred all the fields you need to. The `adaptor` object can be used for holding both meshes and the (potentially expensive) supermesh projection mass-matrices.

But maybe that is no longer necessary.

firedrake/meshadapt.py
```diff
@@ -0,0 +1,500 @@
...
+
+        # Adjust the section
+        entity_dofs = np.zeros(tdim + 1, dtype=np.int32)
+        entity_dofs[0] = tdim ** 2
+        sec = mesh.create_section(entity_dofs)
+        off = 0
+        for c in range(*plex.getChart()):
+            sec.setOffset(c, off)
+            off += sec.getDof(c)
+        plex.setSection(sec)
```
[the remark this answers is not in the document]
jwallwork23: Good point!

firedrake/meshadapt.py
```diff
@@ -0,0 +1,500 @@
...
+        self._update_plex_coordinates()
+
+        # Adjust the section
+        entity_dofs = np.zeros(tdim + 1, dtype=np.int32)
+        entity_dofs[0] = tdim ** 2
+        sec = mesh.create_section(entity_dofs)
+        off = 0
+        for c in range(*plex.getChart()):
+            sec.setOffset(c, off)
+            off += sec.getDof(c)
```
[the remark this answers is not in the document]
jwallwork23: Good point. Reran tests locally without these lines and works fine.

firedrake/meshadapt.py
jwallwork23: This was how it was initially done in the previous PR, but as we discussed it we realised there are quite a few `Function` methods that it is useful to carry over, in particular `assign`, `interpolate`, `project`. Also, a Riemannian metric is just a special case of a tensor field that is SPD at each point. From that perspective, it makes sense for it to be a subclass. Open to other thoughts and suggestions, though!

firedrake/meshadapt.py
jwallwork23: > Would it be possible to hide `RiemannianMetric` from users?

I think `RiemannianMetric` needs to be available to users. Many of the manipulations that people do on metrics (especially for goal-oriented adaptation) need the methods provided here at several steps and cannot be straightforwardly done using methods of `Function` alone.

jwallwork23: > The `adaptor` object can be used for holding both meshes and the (potentially expensive) supermesh projection mass-matrices.

Good point on mass matrices. I guess the way to do that would be to associate a `Projector` with the `MetricBasedAdaptor`, rather than just calling `project`?

firedrake/meshadapt.py
```diff
@@ -0,0 +1,500 @@
...
+            )
+        rank = len(fs.dof_dset.dim)
+        if rank != 2:
+            raise ValueError(
+                "Riemannian metric should be matrix-valued,"
+                f" not rank-{rank} tensor-valued"
+            )
+
+        # Stash mesh data
+        plex = mesh.topology_dm
```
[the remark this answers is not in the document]
jwallwork23: See https://github.com/firedrakeproject/firedrake/pull/2796/commits/b6eedc4fb616df8194ab1a37c8914726dd8fd68c.

firedrake/meshadapt.py
```diff
@@ -0,0 +1,500 @@
...
+        self._mesh = mesh
+        self._plex = plex
+        self._tdim = tdim
+
+        # Ensure DMPlex coordinates are consistent
+        self._update_plex_coordinates()
```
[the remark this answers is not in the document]
jwallwork23: See https://github.com/firedrakeproject/firedrake/pull/2796/commits/b6eedc4fb616df8194ab1a37c8914726dd8fd68c.

firedrake/meshadapt.py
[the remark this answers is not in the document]
jwallwork23: See https://github.com/firedrakeproject/firedrake/pull/2796/commits/8ab93f576b966f42dbdb40a0eda1a54c1e977215.

firedrake/meshadapt.py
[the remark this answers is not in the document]
jwallwork23: See https://github.com/firedrakeproject/firedrake/commit/b6eedc4fb616df8194ab1a37c8914726dd8fd68c.

jwallwork23:
Got the tests passing by temporarily switching SLEPc off (see https://github.com/firedrakeproject/firedrake/pull/2796/commits/507a595df9719fed742398958a5783e8c7460662). There is one failure in `tests/output/test_io_mesh.py::test_io_mesh_uniform_extrusion[quad-square]` but it looks like that crops up in other PRs, too.

firedrake/meshadapt.py
jwallwork23: Symmetric positive-definite, i.e. symmetric matrix with all of its eigenvalues being positive.

firedrake/meshadapt.py
```diff
@@ -0,0 +1,495 @@
...
+
+            # Assign to the output of the intersection
+            size = np.shape(self.dat.data_with_halos)
+            self.dat.data_with_halos[:] = np.reshape(vout.array, size)
+            v2.destroy()
+            v1.destroy()
+            vout.destroy()
+        else:
+            self.intersect(metrics[0])
+            self.intersect(metrics[1:])
```
stephankramer, line 333: Should this be `self.intersect(*metrics[1:])`?

PR #2800, opened 2023-03-07
Rebuild HypreAMS/ADS

pbrubeck:
Algebraic multigrid does not reuse most of the setup (e.g. prologator operators) as they heavily rely on the matrix entries. The `setUp()` routine simply does not do the right thing, so here I am destroying and rebuilding the HypreAMS/ADS solvers within `update()`.

### DIFF
--- a/firedrake/preconditioners/hypre_ads.py
+++ b/firedrake/preconditioners/hypre_ads.py
@@ -12,13 +12,10 @@ __all__ = ("HypreADS",)
 
 class HypreADS(PCBase):
     def initialize(self, obj):
-        A, P = obj.getOperators()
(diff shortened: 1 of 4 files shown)


wence-:
> Algebraic multigrid does not reuse most of the setup (e.g. prologator operators) as they heavily rely on the matrix entries. The `setUp()` routine simply does not do the right thing, so here I am destroying and rebuilding the HypreAMS/ADS solvers within `update()`.

Is this working around what is actually a PETSc bug, or are we driving things wrong? The discrete gradient/curl don't change unless the mesh topology changes, after all

dham:
Apparently we should fix this in PETSc instead.

PR #2802, opened 2023-03-07
Extend facet marking to 1D meshes

francesco-ballarin:
After discussion with @ksagiyam. Please double check, the test passes but I am not fully sure of what I am doing ;)

### DIFF
--- a/firedrake/mesh.py
+++ b/firedrake/mesh.py
@@ -822,7 +822,8 @@ class AbstractMeshTopology(object, metaclass=abc.ABCMeta):
         :arg tf: The :class:`.CoordinatelessFunction` object that marks
             selected entities as 1. f.function_space().ufl_element()
             must be "DP" or "DQ" (degree 0) to mark cell entities and
-            "HDiv Trace " (degree 0) to mark facet entities.
+            "P" (degree 1) in 1D or "HDiv Trace" (degree 0) in 2D or 3D
+            to mark facet entities.
         :arg label_name: The name of the label to store entity selections.
         :arg lable_value: The value used in the label.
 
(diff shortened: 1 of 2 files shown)


firedrake/mesh.py
```diff
@@ -1259,11 +1260,15 @@ def mark_entities(self, tf, label_name, label_value):
         if elem.family() in {"Discontinuous Lagrange", "DQ"} and elem.degree() == 0:
             # cells
             height = 0
-        elif elem.family() == "HDiv Trace" and elem.degree() == 0:
+        elif (
+            (elem.family() == "HDiv Trace" and elem.degree() == 0 and self.cell_dimension() > 1)
+                or
+            (elem.family() == "Lagrange" and elem.degree() == 1 and self.cell_dimension() == 1)
+        ):
```
ksagiyam, line 1267: Lint will fail, so we want something like:
```
        elif self.cell_dimension() == 1 and elem.family() == "Lagrange" and elem.degree() == 1 or \
                self.cell_dimension() > 1 and elem.family() == "HDiv Trace" and elem.degree() == 0:
```

tests/regression/test_mark_entities.py
```diff
@@ -90,3 +90,16 @@ def test_mark_entities_mesh_mark_entities():
     label = plex.getLabel(label_name)
     assert label.getStratumIS(label_value).getSize() == 2
     assert all(label.getStratumIS(label_value).getIndices() == [20, 30])
+
```
ksagiyam, line 93: Want double blank lines.

ksagiyam (changes requested):
Thanks! I requested some minor changes, but otherwise looks good to me.

firedrake/mesh.py
ksagiyam: and similar change in `RelabeledMesh`.

francesco-ballarin:
Not sure why it still says "changes requested", I believe I addressed all of them...

wence-:
> Not sure why it still says "changes requested", I believe I addressed all of them...

@ksagiyam has to agree that you addressed them as requester :)

ksagiyam (approved):

ksagiyam:
Done!

francesco-ballarin:
Thanks @ksagiyam!

PR #2804, opened 2023-03-08
Pytorch coupling

nbouziani:
This PR enables to embed Firedrake operations within PyTorch. It consists in:

- Adding a PyTorch custom operator (analogous to ExternalOperator) to represent Firedrake operators expressed as a ReducedFunctional. Forward and backward computations are delegated to the reduced functional.
- Adding a backend class to map from PyTorch to Firedrake and vice versa
- Adding tests

This PR is associated with the following pyadjoint PR: https://github.com/dolfin-adjoint/pyadjoint/pull/95


firedrake/preconditioners/pytorch_coupling/backends.py
```diff
@@ -0,0 +1,105 @@
...
+        elif isinstance(x, Vector):
+            return self.get_function_space(x.function)
+        elif isinstance(x, float):
+            return None
+        else:
+            raise ValueError("Cannot infer the function space of %s" % x)
+
+
+class PytorchBackend(AbstractMLBackend):
+
```
nbouziani, line 40:
```python3
def __init__(self):
         try:
            import torch
            self._backend = torch
        except ImportError:
            self._backend = None

@property
def backend(self):
    if self:
         return self._backend
    else:
        raise ...

def __bool__(self):
    return self._backend is not None

File: firedrake/preconditioners/pytorch_coupling/pytorch_custom_operator.py [hunk 2]
```diff
@@ -0,0 +1,74 @@
+import collections
+from functools import partial
+
+from firedrake.preconditioners.pytorch_coupling import get_backend
+from firedrake.function import Function
+
+from pyadjoint.reduced_functional import ReducedFunctional
+
+
+backend = get_backend('pytorch')
```
nbouziani, line 10: if backend:
    BackendFunction = ...
else:
    dummy class that exceptions on instantiation.

firedrake/pytorch_coupling/backends.py
```diff
@@ -0,0 +1,120 @@
...
+        raise NotImplementedError
+
+    def from_ml_backend(self, x, V):
+        r"""Convert from ML backend to Firedrake.
+
+           :arg x: ML backend object
+        """
+        raise NotImplementedError
+
+    def get_function_space(self, x):
```
dham, line 27: remove `get_`

firedrake/pytorch_coupling/__init__.py
```diff
@@ -0,0 +1,2 @@
+from firedrake.pytorch_coupling.backends import get_backend                    # noqa: F401
```
connorjward, line 1: Rename this subdirectory to something non-specific to pytorch.

firedrake/pytorch_coupling/backends.py
```diff
@@ -0,0 +1,120 @@
...
+                val = val[0]
+            return Constant(val)
+        else:
+            x = x.detach().numpy()
+            x_F = Function(V, dtype=x.dtype)
+            x_F.vector().set_local(x)
+            return x_F
+
+
+def get_backend(backend_name='pytorch'):
```
dham, line 116: `get_` to `import_`

firedrake/pytorch_coupling/pytorch_custom_operator.py
```diff
@@ -0,0 +1,86 @@
...
+from pyadjoint.reduced_functional import ReducedFunctional
+
+
+backend = get_backend("pytorch")
+
+if backend:
+    # PyTorch is installed
+    BackendFunction = backend.backend.autograd.Function
+else:
+    class BackendFunction(object):
```
connorjward, line 16: Don't need `(object)`.

firedrake/pytorch_coupling/pytorch_custom_operator.py
```diff
@@ -0,0 +1,86 @@
...
+        super(FiredrakeTorchOperator, self).__init__()
+
+    # This method is wrapped by something cancelling annotation (probably 'with torch.no_grad()')
+    @staticmethod
+    def forward(ctx, metadata, *ω):
+        """Forward pass of the PyTorch custom operator."""
+        F = metadata['F']
+        V = metadata['V_controls']
+        # Convert PyTorch input (i.e. controls) to Firedrake
+        ω_F = [backend.from_ml_backend(ωi, Vi) for ωi, Vi in zip(ω, V)]
```
connorjward, line 46: Stick to regular ascii characters.

tests/conftest.py
```diff
@@ -79,6 +82,10 @@ def pytest_runtest_call(item):
 
 def pytest_collection_modifyitems(session, config, items):
     from firedrake.utils import SLATE_SUPPORTS_COMPLEX
+    from firedrake.pytorch_coupling import get_backend
+
+    backend = get_backend("pytorch")
```
connorjward, line 87: Could you maybe specify `ml_backend` (or similar) instead of just "backend". We're reusing the same term for CPU/OpenCL/CUDA.

tests/regression/test_pytorch_coupling.py
```diff
@@ -0,0 +1,228 @@
...
+    """Taylor test for the torch operator"""
+    # Control value
+    ω = Function(V)
+    # Get Firedrake operator and other operator arguments
+    fd_op, args = firedrake_operator
+    # Set reduced functional
+    Jhat = ReducedFunctional(fd_op(ω, *args), Control(ω))
+    # Define the torch operator
+    G = torch_operator(Jhat)
+    # `gradcheck` is likey to fail if the inputs are not double precision (cf. https://pytorch.org/docs/stable/generated/torch.autograd.gradcheck.html)
```
connorjward, line 225: Typo: "likely"

connorjward (changes requested):

dham (changes requested):
Style, and install script.

firedrake/pytorch_coupling/backends.py
nbouziani: I went for `load_` which was the other option discussed as I think `import_` can be a bit confusing when we get the backend after having imported the backend library:

```python3
import torch

...

backend = import_backend("pytorch")
...
```

miguelcoolchips:
Out of curiosity, what happened with `ExternalOperator`? I thought it was a more general class with neural networks as a special case.

nbouziani:
> Out of curiosity, what happened with `ExternalOperator`? I thought it was a more general class with neural networks as a special case.

`ExternalOperator` is about incorporating arbitrary operators in Firedrake, neural networks being an example as you correctly pointed out. This PR is the other way around, it is about including Firedrake operations within PyTorch. Having both works combined together provides a full coupling between both frameworks.

firedrake/ml_coupling/pytorch/backend.py
```diff
@@ -0,0 +1,83 @@
...
+        else:
+            raise ImportError("Error when trying to import PyTorch.")
+
+    def __bool__(self):
+        return self._backend is not None
+
+    @utils.cached_property
+    def custom_operator(self):
+        from firedrake.ml_coupling.pytorch.pytorch_custom_operator import FiredrakeTorchOperator
+        return FiredrakeTorchOperator().apply
```
ksagiyam, line 31: `FiredrakeTorchOperator.apply`? It looks `apply` is a staticmethod.

ksagiyam (changes requested):

firedrake/ml_coupling/pytorch/backend.py
nbouziani: Yes, I am triggering the instantiation on purpose to catch the use of `FiredrakeTorchOperator` when `torch` is not imported, see [here](https://github.com/firedrakeproject/firedrake/blob/1dd095cb5261f9893b9e52efabb518f218690720/firedrake/ml_coupling/pytorch/pytorch_custom_operator.py#L30). I couldn't come up with a way of catching that case at the class level without instantiating the class.

nbouziani: Thanks! I will change that.

> And for raising an exception if torch is not imported should this not be done in PytorchBackend.__init__? Surely it is not sensible to create an instance of PytorchBackend if we can't import torch.

Not raising the error eagerly when loading the backend is motivated by having FiredrakeTorchOperator subclassing a class from a package (`torch`) that may not be installed, see [here](https://github.com/firedrakeproject/firedrake/blob/1dd095cb5261f9893b9e52efabb518f218690720/firedrake/ml_coupling/pytorch/pytorch_custom_operator.py#L34).

connorjward: Myself and @JDBetteridge have had a thought about all this import guarding magic. Could we not just have `ml_coupling` as a Firedrake subpackage that is not imported by default?
The pattern for writing a ML-coupled Firedrake thing would then be:
```py
from firedrake import *  # would run even if pytorch not installed
from firedrake.ml.pytorch import *  # would fail if pytorch is not installed

# rest of script...
```

connorjward: I feel like it would make the code much more readable, easier to maintain and generally less "hacky".

Is the 1 line you reference something like `load_backend("pytorch")`? If so then the import could set the `backend` parameter instead.

Also, I don't fully understand the need for us to set a `backend` at all. What would be the issue with not explicitly setting a `backend` and instead just creating `FiredrakeTorchOperators` as required?

nbouziani: I am not really convinced it is more readable from a user perspective. No, the line in question is:

```python3
G = torch_operator(...)
```

which does the actually wrapping around a Firedrake operator. The backend class is there for backend-specific operations required for example when mapping tensors from firedrake to the backend and vice-versa. This is for separating concerns between generic operations that apply to any ml backend and backend-specific implementations. This is exactly what we currently do with the adjoint (see [here](https://github.com/firedrakeproject/firedrake/blob/203ec3a32cc7b74e5225c08cc9bbebedf502aef4/firedrake/adjoint/blocks.py#L13)).

connorjward: I don’t understand why it’s necessary to keep track of a backend parameter. I think a regular abstract base class inheritance pattern would suffice. I’d be happy to discuss all of this at the next Firedrake meeting.

nbouziani: Sure, we can discuss that at the next meeting.

firedrake/ml_coupling/pytorch/pytorch_custom_operator.py
```diff
@@ -0,0 +1,100 @@
...
+ doi = {10.48550/arXiv.2303.06871}
+}
+""")
+
+
+backend = load_backend("pytorch")
+
+if backend:
+    # PyTorch is installed
+    BackendFunction = backend.backend.autograd.Function
```
dham, line 26: rename to PytorchFunction

dham (changes requested):
change this so that the import is optional rather than the instantiation.

firedrake/ml/pytorch/pytorch_custom_operator.py
```diff
@@ -0,0 +1,92 @@
...
+ title = {Physics-driven machine learning models coupling {PyTorch} and {Firedrake}},
+ author = {Bouziani, Nacime and Ham, David A.},
+ booktitle = {{ICLR} 2023 {Workshop} on {Physics} for {Machine} {Learning}},
+ year = {2023},
+ doi = {10.48550/arXiv.2303.06871}
+}
+""")
+
+
+backend = load_backend("pytorch")
```
connorjward, line 23: If you use free functions this could simply be (I think):
```py
from firedrake.ml.pytorch import backend
```

firedrake/ml/pytorch/backend.py
```diff
@@ -0,0 +1,70 @@
...
+            return self.backend.tensor(x.values(), **kwargs)
+        elif isinstance(x, (float, int)):
+            if isinstance(x, float):
+                # Set double-precision
+                kwargs['dtype'] = self.backend.double
+            return self.backend.tensor(x, **kwargs)
+        else:
+            raise ValueError("Cannot convert %s to a torch tensor" % str(type(x)))
+
+    def from_ml_backend(self, x, V=None):
```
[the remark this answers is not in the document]
connorjward: I guess if you are trying to implement some sort of ML interface then abstract methods might be useful. Have you considered using `@classmethod`?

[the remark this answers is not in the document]
nbouziani: For mapping functions, currently, the instance only carries the backend but can also be used to access other information that might be added when utilising different machine learning frameworks or when GPU support is introduced.

For the PyTorch case, it is not needed for `from_ml_backend` since we use `torch.Tensor`'s sugar syntax to get the tensor in a suitable format for making the Firedrake objects. However, I am not sure that applies to other machine learning frameworks.

firedrake/ml/pytorch/pytorch_custom_operator.py
nbouziani: I think I don't understand what you mean.

firedrake/ml/pytorch/backend.py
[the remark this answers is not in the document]
connorjward: I am wondering if one would ever want to instantiate more than one `PytorchBackend` instance. What would be the benefit?

firedrake/ml/pytorch/pytorch_custom_operator.py
connorjward: If one uses free functions (i.e. not methods in a class) then accessing `from_ml_backend` etc could simply be an import, instead of via a special `load_backend` function.

This is nitpicking though. My main query is about whether or not one would want >1 of `PytorchBackend`.

firedrake/ml/pytorch/__init__.py
```diff
@@ -0,0 +1,7 @@
+try:
+    import torch
```
dham, line 2:
```suggestion
    import torch as backend
```

dham: and turn the torch subpackage into a single module.

dham (changes requested):

firedrake/ml/__init__.py
```diff
@@ -1,2 +1 @@
-from firedrake.ml.backend_base import load_backend    # noqa: F401
-from firedrake.ml.pytorch import *                    # noqa: F401
+from firedrake.ml.pytorch import FiredrakeTorchOperator, torch_operator, from_ml_backend, to_ml_backend   # noqa: F401
```
dham, line 1: You don't want this line. It causes `from firedrake import ml` to import the PyTorch module. Now assume somebody implements a tensor flow backend. You wouldn't want the ml import to fail because of no PyTorch.

firedrake/ml/pytorch.py
```diff
@@ -0,0 +1,160 @@
...
+ author = {Bouziani, Nacime and Ham, David A.},
+ booktitle = {{ICLR} 2023 {Workshop} on {Physics} for {Machine} {Learning}},
+ year = {2023},
+ doi = {10.48550/arXiv.2303.06871}
+}
+""")
+
+
+class FiredrakeTorchOperator(torch.autograd.Function):
+    """
```
dham, line 29: Docstrings are inconsistently formatted. Please at least use one consistent style for the module. We plan to move to Numpydoc so that would be preferable, but any consistent style is fine.

firedrake/ml/pytorch.py
```diff
@@ -0,0 +1,160 @@
...
+        adj_output = F.derivative(adj_input=adj_input)
+
+        # Tuplify adjoint output
+        adj_output = (adj_output,) if not isinstance(adj_output, collections.abc.Sequence) else adj_output
+
+        # None is for metadata arg in `forward`
+        return None, *[to_ml_backend(di) for di in adj_output]
+
+
+def torch_operator(F):
```
dham, line 81: This has a pytorch-specific name, but you attempt to be backend agnostic with `from_ml_backend`. Which are we doing?

firedrake/ml/pytorch.py
```diff
@@ -0,0 +1,160 @@
...
+    """Operator that converts a pyadjoint.ReducedFunctional into a firedrake.FiredrakeTorchOperator
+       whose inputs and outputs are PyTorch tensors.
+    """
```
dham, line 84: docstring seems incomplete.

firedrake/ml/pytorch.py
```diff
@@ -0,0 +1,160 @@
...
+        raise ValueError("F must be a ReducedFunctional")
+
+    V_output = extract_function_space(F.functional)
+    V_controls = [c.control.function_space() for c in F.controls]
+    metadata = {'F': F, 'V_controls': V_controls, 'V_output': V_output}
+    F_P = partial(FiredrakeTorchOperator.apply, metadata)
+    return F_P
+
+
+def extract_function_space(x):
```
dham, line 97: I think this is a private function so it should be called `_extract_function_space`

firedrake/ml/pytorch.py
```diff
@@ -0,0 +1,160 @@
...
+import collections
+from functools import partial
+
+from firedrake.function import Function
+from firedrake.vector import Vector
+from firedrake.constant import Constant
+from firedrake_citations import Citations
+
+from pyadjoint.reduced_functional import ReducedFunctional
+
```
dham, line 15: should probably define `__all__` to specify the public API.

dham (changes requested):

firedrake/ml/pytorch.py
```diff
@@ -0,0 +1,210 @@
...
+        # Tuplify adjoint output
+        adj_output = (adj_output,) if not isinstance(adj_output, collections.abc.Sequence) else adj_output
+
+        # None is for metadata arg in `forward`
+        return None, *[to_torch(di) for di in adj_output]
+
+
+def torch_operator(F):
+    """Operator that converts a :class:`pyadjoint.ReducedFunctional` into a :class:`~FiredrakeTorchOperator`
```
dham, line 91: First line of docstring should be a one line description and shouldn't really have markup:
```python
"""Cast a Firedrake reduced functional to a PyTorch operator.

The resulting :class:`~FiredrakeTorchOperator` will take PyTorch tensors and inputs and return PyTorch tensors as outputs.
```

dham (approved):

firedrake/ml/pytorch.py
```diff
@@ -0,0 +1,212 @@
...
+
+def torch_operator(F):
+    """Cast a Firedrake reduced functional to a PyTorch operator.
+
+    The resulting :class:`~FiredrakeTorchOperator` will take PyTorch tensors as inputs and return PyTorch tensors as outputs.
+
+    Parameters
+    ----------
+    F : pyadjoint.ReducedFunctional
+        The reduced functional to wrap.
```
IvanYashchuk, line 99: It's rather restrictive to wrap only reduced functionals. Do you plan to extend this to arbitrary Firedrake programs?

dham: Despite the name, which is historic, ReducedFunctional can return any overloaded type. This means that this does enable wrapping of arbitrary Firedrake code. ReducedFunctional is simply the mechanism for expressing a Firedrake calculation as a function of controls, which is what you need if you're going to differentiate through it.

IvanYashchuk: I see. So it's possible to send a function that returns the result of `solve(F == 0, u, bcs=bcs)` to ReducedFunctional? I was confused with the test in this PR that uses a function named `solve_poisson` but returns `assemble(u ** 2 * dx)` instead of just returning `u` :
https://github.com/firedrakeproject/firedrake/blob/f05d119ac8302afb047a99d561e56584e4db60b4/tests/regression/test_pytorch_coupling.py#L86-L95

nbouziani: Yes, you can. The output doesn't need to be scalar-valued as illustrated in the other test (`poisson_residual`):

https://github.com/firedrakeproject/firedrake/blob/f05d119ac8302afb047a99d561e56584e4db60b4/tests/regression/test_pytorch_coupling.py#L78-L82

PR #2805, opened 2023-03-08
Add curl and libxml2 as dependencies

angus-g:
It seems the NetCDF build portion of PETSc now has these two libraries as dependencies. Particularly on Ubuntu 22.04, the development versions of these packages aren't installed by default, leading to build failures. The Docker builds of Firedrake have also been failing due to this.

I don't know if these are also required in the brew dependencies?


connorjward (approved):
LGTM the error you are seeing is a known problem with a CI and not related to this PR

Issue #2809, opened 2023-03-09
`CheckpointFile` can hang (likely cause of CI failures)


connorjward:
To make repeated loads of the same mesh fast, `CheckpointFile` current has two class-level `WeakValueDictionary` objects for stashing the meshes ([source](https://github.com/firedrakeproject/firedrake/blob/master/firedrake/checkpointing.py#L521)). A weak-referencing dictionary has been used to avoid storing a global dictionary of previously loaded mesh objects (obviously a memory leak). However, this causes issues in parallel because meshes are non-deterministically cleaned up by the cyclic garbage collector so the cache can hit on one rank and miss on another.

I've managed to produce the following MFE which hangs reliably for me:
```py
import gc

from firedrake import *


# create and save a mesh
mesh_name = "mymesh"
mesh = UnitSquareMesh(100, 100, comm=COMM_WORLD, name=mesh_name)
fname = "test.h5"
with CheckpointFile(fname, "w", comm=COMM_WORLD) as meshfile:
    meshfile.save_mesh(mesh)

# load the mesh once - this populates the caches
print(f"{COMM_WORLD.rank}: loading mesh", flush=True)
with CheckpointFile(fname, "r", comm=COMM_WORLD) as meshfile:
    meshfile.load_mesh(name=mesh_name)
print(f"{COMM_WORLD.rank}: mesh loaded", flush=True)

# clear the mesh from rank 0
if COMM_WORLD.rank == 0:
    gc.collect()

# what do the caches look like?
print(f"{COMM_WORLD.rank} cache size: {len(CheckpointFile._tmesh_cache)}", flush=True)

# load the mesh again
# this will result in a cache hit on one rank and a miss on the other
print(f"{COMM_WORLD.rank}: loading mesh", flush=True)
with CheckpointFile(fname, "r", comm=COMM_WORLD) as meshfile:
    meshfile.load_mesh(name=mesh_name)
print(f"{COMM_WORLD.rank}: mesh loaded", flush=True)
```

This outputs:

```
0: loading mesh
1: loading mesh
0: mesh loaded
1: mesh loaded
1 cache size: 1
1: loading mesh
0 cache size: 0
0: loading mesh

*hangs*
```

ksagiyam:
[...]

PR #2810, opened 2023-03-09
Remove class-level caches from CheckpointFile

connorjward:
Fixes https://github.com/firedrakeproject/firedrake/issues/2809 (and hopefully fixes CI).

Is it appropriate to provide a test for something non-deterministic like this?

### DIFF
--- a/firedrake/checkpointing.py
+++ b/firedrake/checkpointing.py
@@ -1,6 +1,5 @@
 import functools
 import pickle
-import weakref
 from petsc4py.PETSc import ViewerHDF5
 import ufl
 from pyop2 import op2
@@ -517,10 +516,6 @@ class CheckpointFile(object):
     One can also use different number of processes for saving and for loading.
 
     """
-    # Cache for loaded meshes.
-    _mesh_cache = weakref.WeakValueDictionary()
-    _tmesh_cache = weakref.WeakValueDictionary()
-
     def __init__(self, filename, mode, comm=COMM_WORLD):
         self.viewer = ViewerHDF5()
         self.filename = filename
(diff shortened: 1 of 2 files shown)


connorjward:
@ksagiyam this is failing the backwards compatibility tests due to mismatching SFs. Rather than expunging the caches as I do here should they be stored on the instance instead of the class? I find the code quite hard to follow.

ksagiyam:
The `backward_compat` tests are failing I think because I relied on the fact that meshes are cached in those tests.
Here https://github.com/firedrakeproject/firedrake/blob/f13bbb50233a03fb757716dd087ae8b6ff4f546d/tests/output/test_io_backward_compat.py#L112 we end up creating multiple instances of the same mesh with the same name (say meshA and meshB, both named, e.g., "mesh_triangle"). As function spaces are cached based on "mesh_triangle" (and other names) on the `CheckpointFile` instance (`afile`), when we load a function space on meshB, which has already been loaded for meshA and cached, we end up grabbing the cached one.  This causes conflict as meshA has pointSFA, the loaded function space inherits pointSFA, but meshB has poinSFB; though pointSFA and pointSFB are identical, they are different objects, which is what PETSc is complaining.

I think we can just use `with CheckpointFile(..) as afile:` as usual inside the `for` loop in the `backward_compat` tests.

ksagiyam (approved):
Thanks, Connor!

connorjward:
Can we merge this? We are getting random fails in the link checker for the docs build which I'm certain is nothing to do with this PR.

ksagiyam:
Unfortunately, the issue seems not resolved:

https://github.com/firedrakeproject/firedrake/pull/2787
https://github.com/firedrakeproject/firedrake/pull/2804

These branches have been rebased.

connorjward:
Blast.

PR #2844, opened 2023-03-24
Added markers to UnitTriangleMesh, together with refinement option

nabw:
Hi everyone, 

Very small PR to add markers and refinement to the unit triangle mesh. Given how simple it is to refine, would it be worth doing this at the Mesh() level? 

In any case, let me know if you would like to do anything else.

Best

### DIFF
--- a/firedrake/utility_meshes.py
+++ b/firedrake/utility_meshes.py
@@ -523,7 +523,7 @@ def UnitTriangleMesh(
     coord_sec = plex.getCoordinateSection()
     boundary_faces = plex.getStratumIS("boundary_faces", 1).getIndices()
 
-    tol = 1e-2 # 0.5 would suffice
+    tol = 1e-2  # 0.5 would suffice
     for face in boundary_faces:
         face_coords = plex.vecGetClosure(coord_sec, coords, face)
         # |x+y-1| < eps
--- a/firedrake/utility_meshes.py
+++ b/firedrake/utility_meshes.py
@@ -513,9 +513,6 @@ def UnitTriangleMesh(
     plex = mesh.plex_from_cell_list(
         2, cells, coords, comm, mesh._generate_default_mesh_topology_name(name)
     )
-     comm=COMM_WORLD,
-     name=mesh.DEFAULT_MESH_NAME,
-     distribution_name=None,
 
     # mark boundary facets
     plex.createLabel(dmcommon.FACE_SETS_LABEL)
--- a/firedrake/utility_meshes.py
+++ b/firedrake/utility_meshes.py
@@ -500,6 +500,9 @@ def UnitTriangleMesh(
 ):
     """Generate a mesh of the reference triangle
 
+    :kwarg refinement_level: Number of uniform refinements to perform
+    :kwarg distribution_parameters: options controlling mesh
+           distribution, see :func:`.Mesh` for details.
     :kwarg comm: Optional communicator to build the mesh on.
     :kwarg name: Optional name of the mesh.
     :kwarg distribution_name: the name of parallel distribution used
@@ -511,9 +514,7 @@ def UnitTriangleMesh(
     """
     coords = [[0.0, 0.0], [1.0, 0.0], [0.0, 1.0]]
     cells = [[0, 1, 2]]
-    plex = mesh.plex_from_cell_list(
-        2, cells, coords, comm, mesh._generate_default_mesh_topology_name(name)
-    )
+    plex = mesh.plex_from_cell_list(2, cells, coords, comm)
 
     # mark boundary facets
     plex.createLabel(dmcommon.FACE_SETS_LABEL)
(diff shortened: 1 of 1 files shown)


firedrake/utility_meshes.py
```diff
@@ -492,6 +492,7 @@ def OneElementThickMesh(
 @PETSc.Log.EventDecorator()
 def UnitTriangleMesh(
     comm=COMM_WORLD,
+    refinement_level=0,
```
ksagiyam, line 495: If we can refine, I think we should also add `distribution_parameters` as argument just as in other functions.

firedrake/utility_meshes.py
```diff
@@ -512,6 +513,21 @@ def UnitTriangleMesh(
...
+    # mark boundary facets
+    plex.createLabel(dmcommon.FACE_SETS_LABEL)
+    plex.markBoundaryFaces("boundary_faces")
+    coords = plex.getCoordinates()
+    coord_sec = plex.getCoordinateSection()
+    labels = [1,2,0] # Faces are y=0, x+y=0, x=0.
+    boundary_faces = plex.getStratumIS("boundary_faces", 1).getIndices()
+    for face, label in zip(boundary_faces, labels):
+        plex.setLabelValue(dmcommon.FACE_SETS_LABEL, face, label)
+    plex.removeLabel("boundary_faces")
```
ksagiyam, line 526: To be consistent with `UnitSquareMesh` and friends, and also inspired by FIAT facet ordering, we probably want to use `1` (diagonal), `2`(x=0), `3`(y=0).
It is also safer to label facets using coordinates as done in `TensorRectangleMesh`.

ksagiyam (changes requested):

firedrake/utility_meshes.py
nabw: Added in last commit

firedrake/utility_meshes.py
nabw: Sounds fair. Added in last commit.

firedrake/utility_meshes.py
ksagiyam: Could you also add docstring for this:
```
:kwarg distribution_parameters: options controlling mesh
distribution, see :func:`.Mesh` for details.
```

firedrake/utility_meshes.py
```diff
@@ -512,9 +514,35 @@ def UnitTriangleMesh(
...
+        if abs(face_coords[0]) < tol and abs(face_coords[2]) < tol:
+            plex.setLabelValue(dmcommon.FACE_SETS_LABEL, face, 2)
+        # |y| < eps
+        if abs(face_coords[1]) < tol and abs(face_coords[3]) < tol:
+            plex.setLabelValue(dmcommon.FACE_SETS_LABEL, face, 3)
+    plex.removeLabel("boundary_faces")
+    plex.setRefinementUniform(True)
+    for i in range(refinement_level):
+        plex = plex.refine()
+
```
ksagiyam, line 541: Refined `plex` does not inherit `name`, so here we want:
```
plex.setName(mesh._generate_default_mesh_topology_name(name))
```
, instead of passing this name in line 515.

ksagiyam (changes requested):
Some more minor things, but otherwise looks good to me. Thanks!

firedrake/utility_meshes.py
nabw: Added

firedrake/utility_meshes.py
nabw: At which point is it not inherited? After refinement?

nabw:
> Some more minor things, but otherwise looks good to me. Thanks!

Great :) thank you for looking into this.

firedrake/utility_meshes.py
ksagiyam: Yes, after each refinement, `name` is cleared, so we need to `plex.setName` explicitly after all refinements.

ksagiyam (approved):

connorjward:
> Sure! What do you use for linting? If it is autopep8 with default settings I have it already installed. If it is something else, please let me know how to do it.
> […](#)
> On April 5, 2023 5:18:37 AM GMT-04:00, ksagiyam ***@***.***> wrote: Could you fix the lint error?

Run `make lint` from the `firedrake/src/firedrake`. We use `flake8`.

nabw:
> 

Sure

nabw:
> > Sure! What do you use for linting? If it is autopep8 with default settings I have it already installed. If it is something else, please let me know how to do it.
> > […](#)
> > On April 5, 2023 5:18:37 AM GMT-04:00, ksagiyam _**@**_.***> wrote: Could you fix the lint error?
> 
> Run `make lint` from `firedrake/src/firedrake`. We use `flake8`.

Thanks! It is done.

ksagiyam (approved):

PR #2840, opened 2023-03-23
DO NOT MERGE Test updated PETSc

connorjward:

PR #2852, opened 2023-04-03
DO NOT MERGE Test updated PETSc and SLEPc

connorjward:
I'm trying again in the hope that PETSc and SLEPc have fixed the [weird bug](https://github.com/firedrakeproject/firedrake/pull/2840) that was causing CI to fail. If not then I will rename our venv in the container.


JDBetteridge (dismissed):
Ooops!

PR #2853, opened 2023-04-03
Remove date from DumbCheckpoint deprication warning

ReubenHill:
Does what it says on the tin!


firedrake/checkpointing.py
```diff
@@ -362,15 +362,15 @@ class HDF5File(object):
...
+       HDF5File class will soon be deprecated.
        Use :class:`~.CheckpointFile` class instead.
 
     """
     def __init__(self, filename, file_mode, comm=None):
         import warnings
         with warnings.catch_warnings():
             warnings.simplefilter('always', DeprecationWarning)
-            warnings.warn("HDF5File class will be deprecated after 01/01/2023; use CheckpointFile class instead.",
+            warnings.warn("HDF5File class will soon be deprecated; use CheckpointFile class instead.",
```
wence-, line 373: I think the `simplefilter` is to ensure that this deprecation warning is visible, but for future reference, the blessed way to provide a deprecation warning that a _user_ will see is with `FutureWarning`.

dham (approved):

PR #2854, opened 2023-04-03
Fix taping of subfunctions

nbouziani:
1-line change PR that removes the caching of subfunction annotation. This recently introduced caching produces spurious outputs in adjoint computations.

### DIFF
--- a/firedrake/function.py
+++ b/firedrake/function.py
@@ -303,7 +303,7 @@ class Function(ufl.Coefficient, FunctionMixin):
         current = super(Function, self).__dir__()
         return list(OrderedDict.fromkeys(dir(self._data) + current))
 
-    @utils.cached_property
+    @property
     @FunctionMixin._ad_annotate_subfunctions
     def subfunctions(self):
         r"""Extract any sub :class:`Function`\s defined on the component spaces


connorjward (approved):
I think this is fine. It’s unlikely that not caching this will slow people’s code down.

connorjward:
I think what we need to do to fix this is the following:

1. Create a `_subfunctions` cached property
2. Change the existing `subfunctions` property to look something like:
    ```py
    @property
    @annotate_subfunctions
    def subfunctions(self):
        return self._subfunctions
    ```

This should mean that we hit the `annotate_subfunctions` function every time, but the underlying tuple of `Functions` are persistent.

connorjward:
> I think what we need to do to fix this is the following:
> 
> 1. Create a `_subfunctions` cached property
> 2. Change the existing `subfunctions` property to look something like:
>    ```python
>    @property
>    @annotate_subfunctions
>    def subfunctions(self):
>        return self._subfunctions
>    ```
> 
> This should mean that we hit the `annotate_subfunctions` function every time, but the underlying tuple of `Functions` are persistent.

Though my preferred way for doing this sort of thing is to write:

```py
@property
@annotate_subfunctions
def subfunctions(self):
    if self._lazy_subfunctions is None:  # either set in __init__ or use hasattr here
        # create the functions
        self._lazy_subfunctions = tuple(...)
    return self._lazy_subfunctions
```

PR #2879, opened 2023-04-19
Point evaluation on immersed manifold 

ka1219:
# Description
Enables evaluation of functions using `Function.at` when the topological dimension of a mesh is less than its geometric dimension, i.e. an immersed manifold. The logic for drawing bounding boxes around each cell is modified so that each box is a $n$-cube centred around the midpoint of the minimum and maximum coordinates, with side lengths proportional to the largest axis-aligned length of the cell. 

# Checklist for author:

# Checklist for reviewer:

- [ ] Docstrings present
- [ ] New tests present
- [ ] No bad "code smells"
- [ ] No issues in parallel


firedrake/mesh.py
```diff
@@ -2010,8 +2010,10 @@ def spatial_index(self):
         # tolerance. Note that if tolerance is too small it might not actually
         # change the value!
         if hasattr(self, "tolerance") and self.tolerance is not None:
```
ReubenHill, line 2012: If we have no tolerance, then we don't draw our new kind of bounding box. Is that the desired behaviour?

firedrake/mesh.py
```diff
@@ -2010,8 +2010,10 @@ def spatial_index(self):
         # tolerance. Note that if tolerance is too small it might not actually
         # change the value!
```
ReubenHill, line 2011: Assuming the answer to my other comment is "yes", then please update this comment to reflect changes below

tests/regression/test_point_eval_cells.py
```diff
@@ -94,17 +94,3 @@ def test_3d(func3d):
...
-    with pytest.raises(NotImplementedError):
-        # Manifold point location not implemented
-        f([0.70710678118, +0.70710678118])
-
-
-def test_sphere(sphere_mesh):
-    f = sphere_mesh.coordinates
-    with pytest.raises(NotImplementedError):
-        # Manifold point location not implemented
-        f([+0.57735026919, 0.57735026919, +0.57735026919])
```
ReubenHill, line 110: rather than removing the tests in this commit, it would be preferable to check that they produce the expected output

ReubenHill:
Something strange has gone on with the commit history here but, oddly, it's not giving merge conflicts. Perhaps remerge from master or, better yet, rebase on top of it? I know this is still WIP but I'm keen to see how you're getting along

ReubenHill:
The manual has a warning about point evaluation not working for immersed manifolds which this PR should remove. See https://github.com/firedrakeproject/firedrake/blob/master/docs/source/point-evaluation.rst#points-outside-the-domain

firedrake/mesh.py
dham: You are right. I've fixed this.

dham:
This project is over so we should merge this (subject to any fixes). Extension to VertexOnlyMesh should be another PR.

firedrake/mesh.py
```diff
@@ -2036,13 +2036,16 @@ def spatial_index(self):
...
 
-        # Push max and min out so we can find points on the boundary within
-        # tolerance. Note that if tolerance is too small it might not actually
-        # change the value!
-        if hasattr(self, "tolerance") and self.tolerance is not None:
-            coords_diff = coords_max - coords_min
-            coords_min -= self.tolerance*coords_diff
-            coords_max += self.tolerance*coords_diff
+        # Push max and min out to a hypercube centered at its average so we can
+        # find points on the mesh within tolerance.
```
ReubenHill, line 2040: I think this could do with more explanation in the comment - having sat down with pen and paper I don't entirely understand what's going on. It relies on the pseudo-inverse you spoke about am I correct?

firedrake/mesh.py
```diff
@@ -2036,13 +2036,16 @@ def spatial_index(self):
...
+
+        # Note that getattr doesn't work here due to the inheritance games that
+        # are going on in getattr.
+        tolerance = self.tolerance if hasattr(self, "tolerance") else 0.0
+        coords_mid = (coords_max + coords_min)/2
+        d = np.max(coords_max - coords_min, axis=1)[:, None]
+        coords_min = coords_mid - (tolerance + 0.5)*d
+        coords_max = coords_mid + (tolerance + 0.5)*d
```
ReubenHill, line 2048: Even if the tolerance is `0.0` this still changes the values of `coords_min` and `coords_max` - I take it that the tolerance is still doing the same job it was before (i.e. pushing `coords_max` and `coords_min` out so we can find points on the boundary within tolerance). To make this clear I think you should break this into two code blocks:
```suggestion
        # find points on the mesh, even for immersed manifolds.
        coords_mid = (coords_max + coords_min)/2
        d = np.max(coords_max - coords_min, axis=1)[:, None]
        coords_min = coords_mid - d/2
        coords_max = coords_mid + d/2
        
        # Push max and min out so we can find points on the boundary within
        # tolerance. Note that getattr doesn't work here due to the 
        # inheritance games that are going on in getattr.
        tolerance = self.tolerance if hasattr(self, "tolerance") else 0.0
        coords_min -= tolerance*d
        coords_max += tolerance*d
```

ReubenHill (changes requested):
Just some small clarifications needed to make it clearer what's going on.

firedrake/mesh.py
dham: No, it's just really basic geometry. The pseudo-inverse stuff only comes into play at the point where we find local coordinates (and was already working so isn't in this PR).

firedrake/mesh.py
dham: Remember these are whole array operations. We do assume that they are relatively cheap, but I don't think we should be doubling the number just because we assume the reader doesn't know how to expand brackets.

ReubenHill: Alright, it can be stated in the comments.

ReubenHill (dismissed):

ReubenHill (approved):

PR #2899, opened 2023-04-28
Expunge COFFEE

### LINKED ISSUE #2898: BUG: String parloops are untested
It turns out that nowhere in our test suite do we check that `firedrake.par_loop` with a string kernel works as expected. We never hit [this function](https://github.com/firedrakeproject/firedrake/blob/master/firedrake/parloops.py#L140).

I am currently trying to remove all traces of COFFEE from Firedrake so I want to know which of the following courses of action to take:
1. Since we aren't testing it, string parloops are not supported by Firedrake any more. This function can be deleted.
2. String parloops are still needed. I should add tests for them.

@dham what are your thoughts?

connorjward:
# Description
This PR removes COFFEE as a dependency from Firedrake (and related packages). Some substantial cleanup of the code is required but I think most things are working.

## Associated Pull Requests:
- https://github.com/OP2/PyOP2/pull/697
- https://github.com/firedrakeproject/tsfc/pull/291

## Fixes Issues:
Fixes #2898 

# Checklist for author:

# Checklist for reviewer:

- [ ] Docstrings present
- [ ] New tests present
- [ ] No bad "code smells"
- [ ] No issues in parallel


firedrake/pointeval_utils.py
```diff
@@ -68,7 +66,8 @@ def compile_element(expression, coordinates, parameters=None):
     cell = domain.ufl_cell()
     dim = cell.topological_dimension()
     point = gem.Variable('X', (dim,))
-    point_arg = ast.Decl(utils.ScalarType_c, ast.Symbol('X', rank=(dim,)))
+    # point_arg = ast.Decl(utils.ScalarType_c, ast.Symbol('X', rank=(dim,)))
+    point_arg = lp.GlobalArg("X", dtype=utils.ScalarType_c, shape=(dim,))
```
ReubenHill, line 70: How, if at all, do the changes here (and in `pointquery_utils.py`) effect generated point evaluation kernels (i.e. the kind I know about!)? I'd be keen to see some generated code.

ReubenHill, line 70: How, if at all, do the changes here (and in `pointquery_utils.py`) effect generated point evaluation kernels (i.e. the kind I know about!)? I'd be keen to see some generated code.

ReubenHill, line 70: How, if at all, do the changes here (and in `pointquery_utils.py`) effect generated point evaluation kernels (i.e. the kind I know about!)? I'd be keen to see some before/after generated code.

connorjward:
The only tests currently failing are `tests/vertexonlymesh/test_vertex_only_mesh_generation.py::test_partition_behaviour`. We are losing points when the tolerance is set to `1e-16`.

I've tracked this down to the fact that the COFFEE backend for locating points in cells [rounds constants](https://github.com/firedrakeproject/tsfc/blob/master/tsfc/coffee.py#L350) whereas the loopy one does not. This results in very subtly different kernels. For example:

COFFEE:
```c
double  t0  = ((((((-0.999999999999999) * (C[0]))) + (((0.999999999999999) * (C[2]))))) + (((0) * (C[4]))));
double  t1  = ((((((-1) * (C[1]))) + (((0) * (C[3]))))) + (C[5]));
double  t2  = ((((((-1) * (C[0]))) + (((0) * (C[2]))))) + (C[4]));
double  t3  = ((((((-0.999999999999999) * (C[1]))) + (((0.999999999999999) * (C[3]))))) + (((0) * (C[5]))));
double  t4  = ((((t0) * (t1))) + (((-1) * (((t2) * (t3))))));
double  t5  = ((((-1.73205080756888) + (((1.73205080756888) * (X[1]))))) + (((3.46410161513775) * (X[0]))));
double  t6  = ((-1) + (((3) * (X[1]))));
double  t7  = ((-0.166666666666667) * (t6));
double  t8  = ((((0.333333333333333) + (((-0.288675134594813) * (t5))))) + (t7));
double  t9  = ((((0.333333333333333) + (((0.288675134594813) * (t5))))) + (t7));
double  t10  = ((((0.333333333333333) + (((0) * (t5))))) + (((0.333333333333333) * (t6))));
double  t11  = ((((((((C[0]) * (t8))) + (((C[2]) * (t9))))) + (((C[4]) * (t10))))) + (((-1) * (x0[0]))));
double  t12  = ((((((((C[1]) * (t8))) + (((C[3]) * (t9))))) + (((C[5]) * (t10))))) + (((-1) * (x0[1]))));
dX[0] += (((((t1) / (t4))) * (t11))) + (((((((-1) * (t2))) / (t4))) * (t12)));
dX[1] += (((((((-1) * (t3))) / (t4))) * (t11))) + (((((t0) / (t4))) * (t12)));
```

loopy:
```c
t0 = -0.9999999999999989 * C[0] + 0.9999999999999989 * C[2] + 0.0 * C[4];
  t1 = -1.0 * C[1] + 0.0 * C[3] + C[5];
  t2 = -1.0 * C[0] + 0.0 * C[2] + C[4];
  t3 = -0.9999999999999989 * C[1] + 0.9999999999999989 * C[3] + 0.0 * C[5];
  t4 = t0 * t1 + -1.0 * t2 * t3;
  t5 = -1.7320508075688772 + 1.7320508075688772 * X[1] + 3.4641016151377544 * X[0];
  t6 = -1.0 + 3.0 * X[1];
  t7 = -0.16666666666666666 * t6;
  t8 = 0.33333333333333337 + -0.2886751345948129 * t5 + t7;
  t9 = 0.33333333333333337 + 0.2886751345948129 * t5 + t7;
  t10 = 0.3333333333333333 + 0.0 * t5 + 0.3333333333333333 * t6;
  t11 = C[0] * t8 + C[2] * t9 + C[4] * t10 + -1.0 * x0[0];
  t12 = C[1] * t8 + C[3] * t9 + C[5] * t10 + -1.0 * x0[1];
  dX[0] = dX[0] + (t1 / t4) * t11 + ((-1.0 * t2) / t4) * t12;
  dX[1] = dX[1] + ((-1.0 * t3) / t4) * t11 + (t0 / t4) * t12;
```

It is unclear to me (and @ReubenHill) why this rounding is taking place for the COFFEE backend.

I have come up with two possible solutions:
1. Also round things in loopy:
```py
-        return r
-    return v
+        return float(("%%.%dg" % ctx.precision) % r)
+    return float(("%%.%dg" % ctx.precision) % v)
```
2. Increase the tolerances of the test. Something like:
```py
-    inputcoords = [[0.0-1e-15, 0.5],
-                   [0.5, 0.0-1e-15],
-                   [0.5, 1.0+1e-15],
-                   [1.0+1e-15, 0.5],
+    inputcoords = [[0.0-1e-12, 0.5],
+                   [0.5, 0.0-1e-12],
+                   [0.5, 1.0+1e-12],
+                   [1.0+1e-12, 0.5],
                    [0.5, 0.5],
                    [0.5, 0.5],
-                   [0.5+1e-15, 0.5],
-                   [0.5, 0.5+1e-15]]
+                   [0.5+1e-14, 0.5],
+                   [0.5, 0.5+1e-14]]
     npts = len(inputcoords)
     # Check that we get all the points with a big enough tolerance
-    vm = VertexOnlyMesh(parentmesh, inputcoords, tolerance=1e-13)
+    vm = VertexOnlyMesh(parentmesh, inputcoords, tolerance=1e-10)
     assert MPI.COMM_WORLD.allreduce(vm.cell_set.size, op=MPI.SUM) == npts
     # Check that we lose all but the last 4 points with a small tolerance
     with pytest.warns(UserWarning):
-        vm = VertexOnlyMesh(parentmesh, inputcoords, tolerance=1e-16, missing_points_behaviour='warn')
+        vm = VertexOnlyMesh(parentmesh, inputcoords, tolerance=1e-15, missing_points_behaviour='warn')
     assert MPI.COMM_WORLD.allreduce(vm.cell_set.size, op=MPI.SUM) == 4

```

To me option 2 seems preferable. I don't see why the tolerance of a vertex-only mesh should ever need to be so small. The default tolerance is 1 for example.

ReubenHill:
> To me option 2 seems preferable. I don't see why the tolerance of a vertex-only mesh should ever need to be so small. The default tolerance is 1 for example.

Also the point of the test is to make sure points disappear and reappear when the numbers are appropriately defined, not to test the specific values.

firedrake/pointeval_utils.py
```diff
@@ -68,7 +66,8 @@ def compile_element(expression, coordinates, parameters=None):
     cell = domain.ufl_cell()
     dim = cell.topological_dimension()
     point = gem.Variable('X', (dim,))
-    point_arg = ast.Decl(utils.ScalarType_c, ast.Symbol('X', rank=(dim,)))
+    # point_arg = ast.Decl(utils.ScalarType_c, ast.Symbol('X', rank=(dim,)))
```
JDBetteridge, line 69: delete

firedrake/pointquery_utils.py
```diff
@@ -122,8 +122,28 @@ def init_X(fiat_cell, parameters):
...
+
+
+def format_num(v, precision):
+    """Format a real or complex value into a string, showing up to
+    ``precision`` decimal digits.  This function is partly
+    extracted from the open_source "FFC: the FEniCS Form
+    Compiler", freely accessible at
+    https://bitbucket.org/fenics-project/ffc.
+    """
+    f = "%%.%dg" % precision
```
JDBetteridge, line 136: Yuck

firedrake/preconditioners/pmg.py
```diff
@@ -32,7 +32,7 @@ class PMGBase(PCSNESBase):
     Other PETSc options inspected by this class are:
     - 'pmg_coarse_degree': polynomial degree of the coarse level
     - 'pmg_coarse_mat_type': can be either 'aij' or 'matfree'
-    - 'pmg_coarse_form_compiler_mode': can be 'spectral' (default), 'vanilla', 'coffee', or 'tensor'
+    - 'pmg_coarse_form_compiler_mode': can be 'spectral' (default), 'vanilla', or 'tensor'
```
JDBetteridge, line 35: Revert? Rename "legacy"??

firedrake/slate/slac/optimise.py
```diff
@@ -223,6 +223,7 @@ def _drop_double_transpose_terminals(expr, self):
 
 @_drop_double_transpose.register(Factorization)
 def _drop_double_transpose_factorization(expr, self):
+    # FIXME
```
JDBetteridge, line 226: Discussion needed?

firedrake/tsfc_interface.py
```diff
@@ -134,19 +129,16 @@ def __init__(self, form, name, parameters, number_map, interface, coffee=False,
         if self._initialized:
             return
         tree = tsfc_compile_form(form, prefix=name, parameters=parameters,
-                                 interface=interface, coffee=coffee,
+                                 interface=interface,
                                  diagonal=diagonal, log=PETSc.Log.isActive())
         kernels = []
         for kernel in tree:
-            # Set optimization options
-            opts = default_parameters["coffee"].copy()
```
JDBetteridge, line 142: Check these parameters aren't coming from "legacy" optimisation parameters (named coffee)

tests/regression/test_ip_viscosity.py
JDBetteridge, line 1: I think test should stay, comment is wrong, but tests are still valid. Discuss

JDBetteridge (changes requested):

tests/regression/test_hyperelasticity.py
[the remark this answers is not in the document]
connorjward: decided to drop

firedrake/slate/slac/optimise.py
connorjward: Decided to leave as is.

firedrake/pointeval_utils.py
connorjward: Discussed on Slack. We agree that the changes to the generated code are effectively cosmetic.

firedrake/tsfc_interface.py
connorjward: Tracing this back, all this does is set `optlevel` to `Ov`. We manage this perfectly well in PyOP2 and gcc also doesn't seem to know what that flag does anyway. I think that we can probably stop worrying about this.

firedrake/preconditioners/pmg.py
connorjward: Reverted for now. I appreciate that "coffee" is no longer a very sensible name for a TSFC mode.

docs/notebooks/10-sum-factorisation.ipynb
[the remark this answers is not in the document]
connorjward: Done.

firedrake/pointquery_utils.py
connorjward: I think that this is more bizarre rounding behaviour. I will delete it and see if the tests still pass.

connorjward: They do!

connorjward:
In order to get complex passing I have had to do a very nasty hack. It turns out that we have two C strings that define the function `to_reference_coords_kernel`. One in [`firedrake/pointquery_utils.py`](https://github.com/firedrakeproject/firedrake/blob/master/firedrake/pointquery_utils.py#L229) and one in [`firedrake/mg/kernels.py`](https://github.com/firedrakeproject/firedrake/blob/master/firedrake/mg/kernels.py#L65).

Ignoring for now the fact that they are extremely similar functions and arguably should not both be required, the problem that I ran into is that the functions have slightly different type signatures:

`pointquery_utils.py`:
```c
static inline void to_reference_coords_kernel(void *result_, double *x0, %(RealType)s *cell_dist_l1, %(ScalarType)s *C)
```

`mg/kernels.py`:
```c
static inline void to_reference_coords_kernel(PetscScalar *X, const PetscScalar *x0, const PetscScalar *C)
```

In particular note that `x0` is a `double` in one and a `PetscScalar` in the other. In a similar way (but actually the opposite way round) the variable `dX` defined in both functions is a `PetscScalar` and `double` respectively.

This disparity was not previously a problem since we would directly inject the C string formed by `to_reference_coords_newton_step` [into the generated code](https://github.com/firedrakeproject/firedrake/blob/master/firedrake/pointquery_utils.py#L243). However, loopy does not allow you to access just the kernel body so I turned it into a [function call instead](https://github.com/firedrakeproject/firedrake/pull/2899/files#diff-55b969523a7b19ba760f6cac64d53cd24ab656b0ef5e5228de9765687c044fd4R250) and suddenly the types of these arguments become important.

My current hack is [despicable](https://github.com/firedrakeproject/firedrake/pull/2899/files#diff-55b969523a7b19ba760f6cac64d53cd24ab656b0ef5e5228de9765687c044fd4R127) so I think we should clean up what these types are actually supposed to be. I think things will also need changing in [`locate.c`](https://github.com/firedrakeproject/firedrake/blob/master/firedrake/locate.c) and [`evaluate.h`](https://github.com/firedrakeproject/firedrake/blob/master/firedrake/evaluate.h).

@ReubenHill what are your thoughts?

ReubenHill:
I agree that the current C strings are nasty and need to be redone at some point. I'm nervous about changing the multigrid stuff though since it's not my area of expertise. If both kernels are referring to spatial coordinates for `x0` and `dX` then I think they can all be `RealType_c` from `firedrake.utils`.

connorjward:
> I agree that the current C strings are nasty and need to be redone at some point. I'm nervous about changing the multigrid stuff though since it's not my area of expertise. If both kernels are referring to spatial coordinates for `x0` and `dX` then I think they can all be `RealType_c` from `firedrake.utils`.

I am confused about this because it appears as though we get complex coordinates when we are in complex mode:

```py
>>> from firedrake import *
>>> mesh = UnitSquareMesh(2, 2)
>>> mesh.coordinates.dat.dtype
dtype('complex128')
```

ReubenHill:
> 

But they will have zero imaginary parts in all cases methinks due to typecasting. As far as I'm aware coordinates are always real. Worth a query in slack perhaps?

connorjward:
> > 
> 
> But they will have zero imaginary parts in all cases methinks due to typecasting. As far as I'm aware coordinates are always real. Worth a query in slack perhaps?

Jack and I have a meeting with David tomorrow to discuss this set of PRs and the work on `Constants`. Hopefully this can get resolved then.

firedrake/mg/kernels.py
```diff
@@ -49,7 +50,7 @@ def to_reference_coordinates(ufl_coordinate_element, parameters=None):
     code = {
         "geometric_dimension": cell.geometric_dimension(),
         "topological_dimension": cell.topological_dimension(),
-        "to_reference_coords_newton_step": to_reference_coords_newton_step_body(ufl_coordinate_element, parameters),
+        "to_reference_coords_newton_step": to_reference_coords_newton_step_body(ufl_coordinate_element, parameters, x0_dtype=ScalarType, dX_dtype="double"),
```
connorjward, line 53: this is OK - leave a comment explaining why it's needed.

.github/workflows/build.yml
```diff
@@ -48,7 +48,7 @@ jobs:
       - name: Build Firedrake
         run: |
           cd ..
-          ./firedrake/scripts/firedrake-install $COMPLEX --venv-name firedrake_venv --tinyasm --netgen --disable-ssh --minimal-petsc --slepc --documentation-dependencies --install thetis --install gusto --install icepack --install irksome --install femlium --no-package-manager || (cat firedrake-install.log && /bin/false)
+          ./firedrake/scripts/firedrake-install $COMPLEX --venv-name firedrake_venv --tinyasm --netgen --disable-ssh --minimal-petsc --slepc --documentation-dependencies --install thetis --install gusto --install icepack --install irksome --install femlium --no-package-manager --package-branch tsfc no-coffee --package-branch pyop2 no-coffee || (cat firedrake-install.log && /bin/false)
```
connorjward, line 51:
```suggestion
          ./firedrake/scripts/firedrake-install $COMPLEX --venv-name firedrake_venv --tinyasm --netgen --disable-ssh --minimal-petsc --slepc --documentation-dependencies --install thetis --install gusto --install icepack --install irksome --install femlium --no-package-manager || (cat firedrake-install.log && /bin/false)
```

JDBetteridge (approved):

firedrake/mg/kernels.py
ReubenHill: Looks like this didn't happen?

connorjward: We decided to get the merge in in the time we had. I should go back and add this.

PR #2773, opened 2023-02-14
Voting algorithm for VertexOnlyMesh

ReubenHill:
Fixes the VertexOnlyMesh part of #2790. To fix `.at` we need to make a VertexOnlyMesh when doing the point location.

Note this fixes #2178 - points will now redistribute to their appropriate location when `redundant=False`

This also changes the default VertexOnlyMesh creation behaviour to throw a `ValueError` when points are lost.

PR #2912, opened 2023-05-04
Document VertexOnlyMesh in manual

ReubenHill:
# Description
Added documentation of VertexOnlyMesh to the manual pages on point evaluation.

# Checklist for author:

- [] ~All of my functions and classes have appropriate docstrings.~
- [ ] ~I have commented my code where its purpose may be unclear.~
- [ ] ~I have added tests specific to the issues fixed in this PR.~
- [ ] ~I have added tests that exercise the new functionality I have introduced~
- [ ] ~Tests pass locally (`pytest tests` in the `firedrake` source directory) (useful, but not essential if you don't have suitable hardware).~

# Checklist for reviewer:

- [ ] ~Docstrings present~
- [ ] ~New tests present~
- [ ] ~Code correctly commented~
- [ ] ~No bad "code smells"~
- [ ] ~No issues in parallel~
- [ ] ~No CI issues (excessive parallelism/memory usage/time/warnings generated)~
- [ ] ~Upstream/dependent branches and PRs are ready~

### DIFF
--- a/docs/source/point-evaluation.rst
+++ b/docs/source/point-evaluation.rst
@@ -7,15 +7,17 @@ Point evaluation
 
 Firedrake can evaluate :py:class:`~.Function`\s at arbitrary physical
 points.  This feature can be useful for the evaluation of the result
-of a simulation.  Two APIs are offered to this feature: a
-Firedrake-specific one, and one from UFL.
+of a simulation, or for creating expressions which contain point evaluations.
+Three APIs are offered to this feature: two Firedrake-specific ones, and one
+from UFL.
 
 
-Firedrake API
--------------
+Firedrake API 1: 'at' method
+----------------------------
 
-Firedrake offers a convenient API for evaluating functions at
-arbitrary points via :meth:`~.Function.at`:
+Firedrake's first API for evaluating functions at arbitrary points,
+:meth:`~.Function.at`, is designed for simple interrogation of a function with
+a few points.
 
 .. code-block:: python3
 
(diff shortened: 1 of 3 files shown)


docs/source/point-evaluation.rst
```diff
@@ -104,6 +106,186 @@ in parallel. There is no special API, but there are some restrictions:
...
+points :math:`(0.1, 0.1)`, :math:`(0.2, 0.2)` and :math:`(0.3, 0.3)`.
+
+Note that ``f_at_points`` is a :py:class:`~.Function` which takes
+on *all* the values of ``f`` evaluated at ``points``. The cell ordering of a
+:func:`~.VertexOnlyMesh` follows the ordering of the list of points it is given
+at construction. In general :func:`~.VertexOnlyMesh` accepts any numpy array of
+shape ``(num_points, point_dim)`` (or equivalent list) as the set of points to
+create disconnected vertices at.
+
+The operator for evaluation at the points specified can be created can be
```
stephankramer, line 166: repeated "can be created"

docs/source/point-evaluation.rst
```diff
@@ -104,6 +106,186 @@ in parallel. There is no special API, but there are some restrictions:
...
+Firedrake's second API for evaluating functions at arbitrary points,
+interpolation onto a :func:`~.VertexOnlyMesh`, is designed for evaluating a
+function at many points and for creating expressions which contain point
+evaluations. It has been designed from the ground up to be entirely parallel
+compatible. Whilst :meth:`~.Function.at` produces a list of values, a
+cross-mesh interpolation onto :func:`~.VertexOnlyMesh` gives Firedrake
+:py:class:`~.Function`\s.
+
+This is discussed in detail in :cite:`nixonhill2023consistent` but, briefly,
+the idea is that the :func:`~.VertexOnlyMesh` is a mesh whose wich represents a
```
connorjward, line 121:
```suggestion
the idea is that a :func:`~.VertexOnlyMesh` is a mesh representing a
```

connorjward (changes requested):

docs/source/point-evaluation.rst
ReubenHill:
```suggestion
The operator for evaluation at the points specified can be
```

docs/source/point-evaluation.rst
```diff
@@ -104,6 +106,186 @@ in parallel. There is no special API, but there are some restrictions:
...
 
 
+Firedrake API 2: Interpolation onto a vertex-only mesh
+------------------------------------------------------
+
+Firedrake's second API for evaluating functions at arbitrary points,
+interpolation onto a :func:`~.VertexOnlyMesh`, is designed for evaluating a
+function at many points and for creating expressions which contain point
+evaluations. It has been designed from the ground up to be entirely parallel
+compatible. Whilst :meth:`~.Function.at` produces a list of values, a
```
stephankramer, line 116: It would be nice to see a bit more detail on how to use it in parallel rather than referring to the docstring as you do later on. Is the warning in the docstring

> This means that if rank A has ``vertexcoords`` {X} that are not found in the mesh cells owned by rank A but are found in the mesh cells owned by rank B, **and rank B has not been supplied with those**, then they will be moved to rank B.

stil valid? In my interpretation reading the code - but correct me if I'm wrong - with `redundant=False` every process specifies unique points which get concatenated with an algatherv, and then every local process keeps its locally visible ones with some voting algorithm to ensure each point gets assignted to only one process.

Regardless of how the points are specified (redundant or not), I'm a little unclear on how the interpolated data is meant to be used in parallel. The `.dat.data` corresponds to the interpolated data of the locally visible detectors, but as a user I have no idea which ones they are and how they are related to the points I've specified and in which order.

ReubenHill: > stil valid? In my interpretation reading the code - but correct me if I'm wrong - with redundant=False every process specifies unique points which get concatenated with an algatherv, and then every local process keeps its locally visible ones with some voting algorithm to ensure each point gets assignted to only one process.

It is still valid, in fact it's a new feature as of #2773! The point is that you can specify a point on one rank and, if it's not on that rank but is on another, it'll be moved to that rank. See issue #2178 where this was first raised.

>Regardless of how the points are specified (redundant or not), I'm a little unclear on how the interpolated data is meant to be used in parallel. The .dat.data corresponds to the interpolated data of the locally visible detectors, but as a user I have no idea which ones they are and how they are related to the points I've specified and in which order.

I don't quite understand you here, what do you mean by detectors?

dham: Well you can get the point coordinates from the VoM. However I think this exposes a hole in the current implementation. 

Think about the use case of interpolating from one mesh to another, then the point cloud starts with one decomposition. It necessarily gets rearranged to another parallel decomposition in order to match the parent mesh and do interpolation. However, the user needs the points back in the original configuration to use the data. 

I think we need to store the original rank and local number of every point and preserve this as we transmit the points around (these could be fields on the DMSwarm). Then we could provide a method that gives you back the points in the original decomposition. I guess we have to think about what the type that that would return would be.

stephankramer: Oh heh sorry, on the docstring: I read what it says here https://www.firedrakeproject.org/firedrake.html#firedrake.mesh.VertexOnlyMesh (which seems out-of-date): "then the vertexcoords {X} will be lost." but then I copied from the code, which is correct indeed.

That was exactly the application I was thinking of indeed.

Another application I can think of is interpolating onto an external grid, say a regular structured one. I guess in that situation you would typically use `redundant=True`, but you have the same issue that it'll be hard to piece the local interpolated values in each .dat.data back to a structured array. Is this something DMSWARM can do? We do give it the original global index of the points as provided.

ReubenHill: Yep looks like the website needs to be rebuilt @stephankramer . @dham I wonder if this moves us in a direction of needing VoMs which have no parent...

@stephankramer I suppose I could start saving the interpolated values as a field on the DMSWARM...

ReubenHill: @stephankramer Okay I've expanded the manual to include more information

ReubenHill: Also you made me realise there's a mistake in the VertexOnlyMesh API docs that I've now fixed 82f91141a1035e56941d1dbc774c6611742d2566

docs/source/point-evaluation.rst
```diff
@@ -7,15 +7,17 @@ Point evaluation
...
-of a simulation.  Two APIs are offered to this feature: a
-Firedrake-specific one, and one from UFL.
+of a simulation, or for creating expressions which contain point evaluations.
+Three APIs are offered to this feature: two Firedrake-specific ones, and one
+from UFL.
 
 
-Firedrake API
--------------
+Firedrake API 1: 'at' method
```
dham, line 15:
```suggestion
Firedrake convenience function
```

docs/source/point-evaluation.rst
```diff
@@ -104,6 +106,302 @@ in parallel. There is no special API, but there are some restrictions:
 * Each process will get the same values.
 
 
+Firedrake API 2: Interpolation onto a vertex-only mesh
+------------------------------------------------------
+
+Firedrake's second API for evaluating functions at arbitrary points,
```
dham, line 112:
```suggestion
Firedrake's principal API for evaluating functions at arbitrary points,
```

docs/source/point-evaluation.rst
```diff
@@ -104,6 +106,302 @@ in parallel. There is no special API, but there are some restrictions:
 * Each process will get the same values.
 
 
+Firedrake API 2: Interpolation onto a vertex-only mesh
```
dham, line 109:
```suggestion
Primary API: Interpolation onto a vertex-only mesh
```

docs/source/point-evaluation.rst
```diff
@@ -104,6 +106,302 @@ in parallel. There is no special API, but there are some restrictions:
...
+------------------------------------------------------
+
+Firedrake's second API for evaluating functions at arbitrary points,
+interpolation onto a :func:`~.VertexOnlyMesh`, is designed for evaluating a
+function at many points and for creating expressions which contain point
```
dham, line 114:
```suggestion
function at many points, or repeatedly, and for creating expressions which contain point
```

docs/source/point-evaluation.rst
```diff
@@ -104,6 +106,302 @@ in parallel. There is no special API, but there are some restrictions:
...
+whose cells are vertices is the space of piecewise constant functions, also
+known as the Polynomial degree 0 Discontinuous Galerkin (P0DG) space.
+
+Our vertex-only meshes are immersed in some 'parent' mesh. We perform point
+evaluation of a function :math:`f` defined in a function space
+:math:`V` on the parent mesh by interpolating into the P0DG space on the
+:func:`~.VertexOnlyMesh`. For example:
+
+.. code-block:: python3
+
```
dham, line 134: Import the lines of code from a test: https://www.sphinx-doc.org/en/master/usage/restructuredtext/directives.html#directive-literalinclude

dham (changes requested):

docs/source/point-evaluation.rst
```diff
@@ -104,6 +100,261 @@ in parallel. There is no special API, but there are some restrictions:
...
+------------------------------------
+
+Any set of points with associated data in our domain can be expressed as a
+P0DG function on a :func:`~.VertexOnlyMesh`.
+
+.. code-block:: python3
+
+   vom = VertexOnlyMesh(parent_mesh, point_locations_from_elsewhere)
+   P0DG = FunctionSpace(vom, "DG", 0)
+   y_pts = Function(P0DG).dat.data[:] = point_data_values_from_elsewhere
```
dham, line 261: I think we now know this isn't parallel-safe. Maybe we should drop this pending the next fix....

ReubenHill: TODO: add a warning that the example uses an internal firedrake API (`Function(P0DG).dat.data[:]`) and that it doesn’t work in parallel. I can state that a parallel safe way of interacting with external data will be available soon.

docs/source/point-evaluation.rst
```diff
@@ -98,6 +100,235 @@ in parallel. There is no special API, but there are some restrictions:
...
+.. code-block:: python3
+
+   V = FunctionSpace(parent_mesh, "N1curl", 2)
+
+each require
+
+.. code-block:: python3
+
+   vom = VertexOnlyMesh(parent_mesh, points)
+   P0DG_vec = VectorFunctionSpace(vom, "DG", 0)
```
connorjward, line 177: Do you not need to provide `dim` as an argument?

docs/source/point-evaluation.rst
```diff
@@ -98,6 +100,235 @@ in parallel. There is no special API, but there are some restrictions:
...
+   :dedent:
+   :lines: 100-117
+
+Note that since our tolerance is relative, the number of cells in a mesh
+dictates the point loss behaviour close to cell edges. So the mesh
+``UnitSquareMesh(5, 5, quadrilateral = True)`` will include the point
+:math:`(1.1, 1.0)` by default.
+
+If a mesh tolerance changes where we already have an immersed vertex-only mesh,
+such that the new tolerance ought to cause points to disappear, they will not.
```
connorjward, line 329: I don't quite understand this paragraph.

tests/vertexonly/test_vertex_only_manual.py
```diff
@@ -0,0 +1,119 @@
...
+        # This will raise a ValueError
+        vom = VertexOnlyMesh(parent_mesh, points, missing_points_behaviour='error')
+
+        # This will generate a warning and the point will be lost
+        vom = VertexOnlyMesh(parent_mesh, points, missing_points_behaviour='warn')
+
+        # This will cause the point to be silently lost
+        vom = VertexOnlyMesh(parent_mesh, points, missing_points_behaviour=None)
+
+        assert vom  # Just here to shut up flake8 unused variable warning.
```
connorjward, line 46: Should these be unindented?

connorjward (commented):
Generally looks really good! Just some small questions.

docs/source/point-evaluation.rst
ReubenHill: For `VectorFunctionSpace`? Nope. From the `VectorFunctionSpace` docs
```
    :arg dim: An optional number of degrees of freedom per function
       space node (defaults to the geometric dimension of the mesh).
```

docs/source/point-evaluation.rst
ReubenHill: It's saying that changing mesh tolerance only effects point location when you create a new VOM

ReubenHill: Would it be clearer if I explicitly stated that?

tests/vertexonly/test_vertex_only_manual.py
ReubenHill: Yes

ReubenHill:
```suggestion
    # This will generate a warning and the point will be lost
    vom = VertexOnlyMesh(parent_mesh, points, missing_points_behaviour='warn')

    # This will cause the point to be silently lost
    vom = VertexOnlyMesh(parent_mesh, points, missing_points_behaviour=None)

    assert vom  # Just here to shut up flake8 unused variable warning.
```

docs/source/point-evaluation.rst
connorjward: Yes of course.

docs/source/point-evaluation.rst
connorjward: I think so.

ReubenHill, line 329:
```suggestion
Changing mesh tolerance only affects point location after it has been changed.
To apply the new tolerance to a vertex-only mesh, a new vertex-only mesh must 
be created. Any existing immersed vertex-only mesh will have been created 
using the previous tolerance and will be unaffected by the change.
```

ReubenHill: Can you take a look at my proposed diff and see if it's alright?

connorjward: Yep I'm happier with that. Thanks. If you commit those diffs I'll approve the PR.

connorjward (approved):
Looks good to me.

PR #2916, opened 2023-05-05
only linkcheck master

dham:
# Description

Link checking causes frequent spurious fails on PRs because external websites are unreliable. This PR works around this by only link checking master.

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -48,7 +48,11 @@ jobs:
         run: |
           . /home/firedrake/firedrake/bin/activate
           cd docs
-          make linkcheck
           make html
           make latex
           make latexpdf
+      - name: Check documentation links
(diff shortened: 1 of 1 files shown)


.github/workflows/docs.yml
```diff
@@ -48,7 +48,11 @@ jobs:
...
           . /home/firedrake/firedrake/bin/activate
           cd docs
-          make linkcheck
           make html
           make latex
           make latexpdf
+      - name: Check documentation links
+        if: ${{ github.ref == 'refs/heads/master' }}
+        run: |
+          cd docs
```
JDBetteridge, line 57:
```suggestion
          . /home/firedrake/firedrake/bin/activate
          cd docs
```

JDBetteridge (changes requested):
I think we still need the venv active for sphinx to work, o/w this LGTM.

PR #2927, opened 2023-05-10
Introduce new Constant

JDBetteridge:
# Description
This PR is dependent on all of COFFEE being removed first.

## Associated Pull Requests:
- [701](https://github.com/OP2/PyOP2/pull/701)
- [294](https://github.com/firedrakeproject/tsfc/pull/294)
- [98](https://github.com/dolfin-adjoint/pyadjoint/pull/98)
- After Firedrake is merged this can go in also: [51](https://bitbucket.org/pefarrell/defcon/branch/JDBetteridge/enable_arclength_continuation)

## Fixes Issues:
- Maybe

# Checklist for author:

# Checklist for reviewer:

- [ ] Docstrings present
- [ ] New tests present
- [ ] No bad "code smells"
- [ ] No issues in parallel

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -75,8 +75,9 @@ jobs:
             --install icepack \
             --install irksome \
             --install femlium \
-            --package-branch tsfc no-coffee \
-            --package-branch pyop2 no-coffee || (cat firedrake-install.log && /bin/false)
+            --package-branch tsfc JDBetteridge/constant_update \
+            --package-branch pyop2 JDBetteridge/global_literal \
+            --package-branch pyadjoint connorjward/constants-away || (cat firedrake-install.log && /bin/false)
       - name: Install test dependencies
         run: |
           . ../firedrake_venv/bin/activate
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -74,7 +74,9 @@ jobs:
             --install gusto \
             --install icepack \
             --install irksome \
-            --install femlium || (cat firedrake-install.log && /bin/false)
+            --install femlium \
+            --package-branch tsfc no-coffee \
+            --package-branch pyop2 no-coffee || (cat firedrake-install.log && /bin/false)
       - name: Install test dependencies
         run: |
           . ../firedrake_venv/bin/activate
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -75,9 +75,7 @@ jobs:
             --install icepack \
             --install irksome \
             --install femlium \
-            --package-branch tsfc JDBetteridge/constant_update \
-            --package-branch pyop2 JDBetteridge/global_literal \
-            --package-branch pyadjoint connorjward/constants-away || (cat firedrake-install.log && /bin/false)
+            || (cat firedrake-install.log && /bin/false)
       - name: Install test dependencies
         run: |
           . ../firedrake_venv/bin/activate
--- a/firedrake/adjoint/variational_solver.py
+++ b/firedrake/adjoint/variational_solver.py
@@ -103,8 +103,7 @@ class NonlinearVariationalSolverMixin:
         affect the user-defined self._ad_problem.F, self._ad_problem.J and self._ad_problem.u
         expressions, we'll instead create clones of them.
         """
-        from firedrake import NonlinearVariationalProblem
-        from firedrake.function import Function
+        from firedrake import Function, NonlinearVariationalProblem
         F_replace_map = {}
         J_replace_map = {}
 
--- a/firedrake/adjoint/variational_solver.py
+++ b/firedrake/adjoint/variational_solver.py
@@ -1,5 +1,4 @@
 import copy
-from firedrake.constant import Constant
 from functools import wraps
 from pyadjoint.tape import get_working_tape, stop_annotating, annotate_tape, no_annotations
 from firedrake.adjoint.blocks import NonlinearVariationalSolveBlock
@@ -105,6 +104,7 @@ class NonlinearVariationalSolverMixin:
         expressions, we'll instead create clones of them.
         """
         from firedrake import NonlinearVariationalProblem
+        from firedrake.function import Function
         F_replace_map = {}
         J_replace_map = {}
 
@@ -115,7 +115,7 @@ class NonlinearVariationalSolverMixin:
         for block_variable in dependencies:
             coeff = block_variable.output
             if coeff in F_coefficients and coeff not in F_replace_map:
-                if isinstance(coeff, Constant):
+                if isinstance(coeff, Function) and coeff.ufl_element().family() == "Real":
                     F_replace_map[coeff] = copy.deepcopy(coeff)
                 else:
                     F_replace_map[coeff] = coeff.copy(deepcopy=True)
@@ -124,7 +124,7 @@ class NonlinearVariationalSolverMixin:
             if coeff in J_coefficients and coeff not in J_replace_map:
                 if coeff in F_replace_map:
                     J_replace_map[coeff] = F_replace_map[coeff]
-                elif isinstance(coeff, Constant):
+                elif isinstance(coeff, Function) and coeff.ufl_element().family() == "Real":
                     J_replace_map[coeff] = copy.deepcopy(coeff)
                 else:
                     J_replace_map[coeff] = coeff.copy()
--- a/firedrake/assemble.py
+++ b/firedrake/assemble.py
@@ -1247,8 +1247,12 @@ class _FormHandler:
     @staticmethod
     def iter_constants(form, kinfo):  # FIXME kinfo not currently needed
         from tsfc.ufl_utils import extract_firedrake_constants
-        for const in extract_firedrake_constants(form):
-            yield const
+        if isinstance(form, slate.TensorBase):
+            for const in form.constants():
+                yield const
+        else:
+            for const in extract_firedrake_constants(form):
+                yield const
 
     @staticmethod
     def index_function_spaces(form, indices):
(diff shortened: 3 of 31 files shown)


firedrake/constant.py
```diff
@@ -14,19 +15,24 @@
...
     else:
-        dat = op2.Global(shape, data, comm=comm)
+        dat = op2type(shape, data, comm=comm)
     return dat, rank, shape
 
 
-class Constant(ufl.Coefficient, ConstantMixin):
-
+# Think "literal"
+class Constant(ufl.constantvalue.ConstantValue, ConstantMixin):
```
wence-, line 35: Surely the right thing to inherit from is https://github.com/FEniCS/ufl/blob/d2dfbef9a60bede67ac8361712316bf427f9eecc/ufl/constant.py#L18

?

`ConstantValue` is a compile-time literal in UFL, not a runtime constant.

connorjward: The reason we have done this is that a UFL `Constant` is what Firedrake calls a `Function` in R. This means that it has an associated `ufl_domain` and therefore a communicator. The reason for the changes being made here is that many people use `Constant` just for parametrising kernel arguments and in some cases (like defcon) finding an appropriate mesh/communicator is tricky and we get deadlocks.

Our solution, not quite shown here, is for TSFC to define a `KernelArgument` mixin class that marks an element of a UFL expression/form as being a kernel parameter. Firedrake `Function` and `Constant` would then inherit from this.

wence-: This conversely means that `Constant(1)*dx` needs one to define a domain on `dx`.

wence-: > The reason we have done this is that a UFL Constant is what Firedrake calls a Function in R.

This is not quite true because you can't have a `Constant` as an unknown (for a solution variable, etc..)

firedrake/assemble.py
```diff
@@ -1229,6 +1244,16 @@ def iter_active_coefficients(form, kinfo):
             for subidx in subidxs:
                 yield form.coefficients()[idx].subfunctions[subidx]
 
+    @staticmethod
+    def iter_constants(form, kinfo):  # FIXME kinfo not currently needed
```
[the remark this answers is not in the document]
ReubenHill: I guess from lookign at the rest of this file, it seems that if a function's name is descriptive enough then it doesn't need a docstring

dham (changes requested):

firedrake/ufl_expr.py
```diff
@@ -155,43 +158,48 @@ def derivative(form, u, du=None, coefficient_derivatives=None):
...
     elif isinstance(uc, firedrake.Constant):
         if uc.ufl_shape != ():
             raise ValueError("Real function space of vector elements not supported")
+        # Replace instances of the constant with a new argument ``x``
+        # and differentiate wrt ``x``.
         V = firedrake.FunctionSpace(mesh, "Real", 0)
-        du = argument(V)
+        x = ufl.Coefficient(V, n + 1)
+        n += 1
+        form = ufl.replace(form, {u: x})
```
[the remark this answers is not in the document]
JDBetteridge: https://github.com/FEniCS/ufl/issues/171

tests/regression/test_solver_error_checking.py
```diff
@@ -37,6 +37,9 @@ def test_invalid_solution_type(a, L, c):
         solve(a == L, c)
 
 
+# TODO: Check with a Constant with a domain (ie: add check for vector lengths)
```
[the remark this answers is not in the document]
JDBetteridge: This comment is only relevant to a previous state of the PR. Function in the R-space (even if it's constructed as a Constant with a domain) is a valid solution type.

firedrake/adjoint/variational_solver.py
```diff
@@ -105,6 +104,7 @@ def _ad_problem_clone(self, problem, dependencies):
         expressions, we'll instead create clones of them.
         """
         from firedrake import NonlinearVariationalProblem
+        from firedrake.function import Function
```
connorjward, line 107:
```suggestion
        from firedrake import Function, NonlinearVariationalProblem
```

firedrake/constant.py
```diff
@@ -49,43 +53,56 @@ class Constant(ufl.Coefficient, ConstantMixin):
...
+    def __new__(cls, value, domain=None):
+        if domain:
+            # Avoid circular import
+            from firedrake.function import Function
+            from firedrake.functionspace import FunctionSpace
+            import warnings
+            warnings.warn("Constants with a domain defined are functions in the Real space")
+
+            dat, rank, shape = _create_dat(op2.Global, value, domain._comm)
+            dat.zero()
```
connorjward, line 67: Why are we zeroing here? We have a defined `value`.

firedrake/constant.py
```diff
@@ -49,43 +53,56 @@ class Constant(ufl.Coefficient, ConstantMixin):
...
     """
+    _globalcount = 0
+
+    def __new__(cls, value, domain=None):
+        if domain:
+            # Avoid circular import
+            from firedrake.function import Function
+            from firedrake.functionspace import FunctionSpace
+            import warnings
+            warnings.warn("Constants with a domain defined are functions in the Real space")
```
connorjward, line 64:
```suggestion
            warnings.warn(
                "Giving Constants a domain is no longer supported. Instead please "
                "create a Function in the Real space.", FutureWarning
            )
```

firedrake/constant.py
```diff
@@ -49,43 +53,56 @@ class Constant(ufl.Coefficient, ConstantMixin):
...
+            cell = domain.ufl_cell()
+            if rank == 0:
+                element = ufl.FiniteElement("R", cell, 0)
+            elif rank == 1:
+                element = ufl.VectorElement("R", cell, 0, shape[0])
+            else:
+                element = ufl.TensorElement("R", cell, 0, shape=shape)
+
+            R = FunctionSpace(domain, element, name="firedrake.Constant")
+            return Function(R, val=dat).assign(value)
```
connorjward, line 79: Same question about `value`.

firedrake/function.py
```diff
@@ -413,6 +415,14 @@ def assign(self, expr, subset=None):
         """
         if expr == 0:
             self.dat.zero(subset=subset)
+        # Assign literal collections to vector and tensor valued functions in R.
+        elif (isinstance(expr, collections.abc.Container)
+              and self.ufl_element().family() == "Real"):
```
connorjward, line 420:
```suggestion
        elif self.ufl_element().family() == "Real":
```

I think the extra `isinstance` check is overkill. Also won't `f.assign(non_vector_value)` fail with it when we would expect it to work?

firedrake/projection.py
```diff
@@ -282,7 +282,8 @@ def Projector(v, v_out, bcs=None, solver_parameters=None,
     else:
         if bcs is not None:
             raise ValueError("Haven't implemented supermesh projection with boundary conditions yet, sorry!")
-        if not isinstance(source, function.Function):
+        # Test if not a function OR whether unknown in the reals
```
connorjward, line 285:
```suggestion
```

Not a useful comment.

firedrake/slate/slac/kernel_builder.py
```diff
@@ -107,13 +108,16 @@ def shape(self, tensor):
         else:
             return tensor.shape
 
-    def extent(self, coefficient):
-        """ Calculation of the range of a coefficient."""
-        element = coefficient.ufl_element()
-        if element.family() == "Real":
-            return (coefficient.dat.cdim, )
+    def extent(self, argument):
+        """ Calculation of the value size of a constant or coefficient."""
```
connorjward, line 112:
```suggestion
        """Return the value size of a constant or coefficient."""
```

firedrake/solving.py
```diff
@@ -141,6 +140,10 @@ def _solve_varproblem(*args, **kwargs):
         near_nullspace, \
         options_prefix = _extract_args(*args, **kwargs)
 
+    # Check whether solution is valid
+    if not isinstance(u, (function.Function, vector.Vector)):
+        raise TypeError("Provided solution is a '%s', not a Function" % type(u).__name__)
```
connorjward, line 145:
```suggestion
        raise TypeError(f"Provided solution is a '{type(u).__name__}', not a Function")
```

firedrake/solving.py
```diff
@@ -218,6 +221,10 @@ def _la_solve(A, x, b, **kwargs):
     bcs, solver_parameters, nullspace, nullspace_T, near_nullspace, \
         options_prefix = _extract_linear_solver_args(A, x, b, **kwargs)
 
+    # Check whether solution is valid
+    if not isinstance(x, (function.Function, vector.Vector)):
+        raise TypeError("Provided solution is a '%s', not a Function" % type(x).__name__)
```
connorjward, line 226:
```suggestion
        raise TypeError(f"Provided solution is a '{type(x).__name__}', not a Function")
```

firedrake/tsfc_interface.py
```diff
@@ -201,6 +201,7 @@ def compile_form(form, name, parameters=None, split=True, interface=None, diagon
 
     key = (name, utils.tuplify(parameters), split, diagonal)
     try:
+        raise KeyError
```
connorjward, line 204:
```suggestion
```

This should definitely go!

connorjward (changes requested):

firedrake/constant.py
JDBetteridge: The assign needs to happen for annotation to occur. We can get away with not zeroing the `dat`, but we need to assign to create an assign block.

.github/workflows/build.yml
```diff
@@ -74,7 +74,10 @@ jobs:
             --install gusto \
             --install icepack \
             --install irksome \
-            --install femlium || (cat firedrake-install.log && /bin/false)
+            --install femlium \
+            --package-branch tsfc JDBetteridge/constant_update \
+            --package-branch pyop2 JDBetteridge/global_literal \
+            --package-branch pyadjoint connorjward/constants-away || (cat firedrake-install.log && /bin/false)
```
dham, line 80:
```suggestion
            || (cat firedrake-install.log && /bin/false)
```

PR #2936, opened 2023-05-15
Retrieve original VOM decomposition

ReubenHill:
# Description
Allow the retrieval of the original decomposition of a `VertexOnlyMesh` to, eventually, allow arbitrary mesh-to-mesh interpolation where we have point evaluation nodes.

Also put VertexOnlyMesh points in parent mesh halos, with appropriate PETSc SF set up and pyop2 labelling (core, owned and ghost) inherited from the parent mesh.

## Associated Pull Requests:
- None yet

## Fixes Issues:
- Allows parallel safe I/O
- Global index is now the unique identifier of a point meaning that we can have points in halos without issue. This will make checkpointing `VertexOnlyMesh` easier.

# Checklist for author:

# Checklist for reviewer:

- [ ] Docstrings present
- [ ] New tests present
- [ ] No bad "code smells"
- [ ] No issues in parallel

### DIFF
--- a/firedrake/cython/dmcommon.pyx
+++ b/firedrake/cython/dmcommon.pyx
@@ -1755,6 +1755,92 @@ def mark_entity_classes(PETSc.DM dm):
 
 @cython.boundscheck(False)
 @cython.wraparound(False)
+def mark_entity_classes_using_cell_dm(PETSc.DM swarm):
+    """
+    Mark all points in a given Particle in Cell (PIC) DMSwarm according to the
+    PyOP2 entity classes (core, owned, ghost) using the markers of the parent
+    DMPlex or DMSwarm cells (i.e. the cells within which the swarm points are
+    located).
+    """
+    cdef:
+        PETSc.DM plex=None
+        PETSc.IS core_is=None
+        PETSc.IS owned_is=None
+        PETSc.IS ghost_is=None
+        DMLabel swarm_label_core, swarm_label_owned, swarm_label_ghost
+        PetscInt label_idx, label
+        np.ndarray[PetscInt, ndim=1, mode="c"] swarm_plex_cells
+        np.ndarray[PetscInt, ndim=1, mode="c"] swarm_parent_cell_labels
+
+
+    swarm.createLabel("pyop2_core")
+    swarm.createLabel("pyop2_owned")
+    swarm.createLabel("pyop2_ghost")
+    CHKERR(DMGetLabel(swarm.dm, b"pyop2_core", &swarm_label_core))
+    CHKERR(DMGetLabel(swarm.dm, b"pyop2_owned", &swarm_label_owned))
+    CHKERR(DMGetLabel(swarm.dm, b"pyop2_ghost", &swarm_label_ghost))
+
+    plex = swarm.getCellDM()
+
+    # Retrieve the indices into the parent DM at which each label is defined.
+    # The label value of 1 is set in mark_entity_classes.
+    core_is = plex.getStratumIS(b"pyop2_core", 1)
+    owned_is = plex.getStratumIS(b"pyop2_owned", 1)
+    ghost_is = plex.getStratumIS(b"pyop2_ghost", 1)
+    # The index numbers correspond to the numbering of the cell. NOTE: We have
+    # to put null checks here because petsc4py will not return empty indices
+    # when the iset is null, instead it will crash.
+    if core_is.iset == NULL:
+        core_idxs = np.array([], dtype=IntType)
+        max_core_idx = -1
+    else:
+        core_idxs = core_is.getIndices()
+        max_core_idx = core_idxs.max()
+    if owned_is.iset == NULL:
+        owned_idxs = np.array([], dtype=IntType)
+        max_owned_idx = -1
+    else:
+        owned_idxs = owned_is.getIndices()
+        max_owned_idx = owned_idxs.max()
+    if ghost_is.iset == NULL:
+        ghost_idxs = np.array([], dtype=IntType)
+        max_ghost_idx = -1
+    else:
+        ghost_idxs = ghost_is.getIndices()
+        max_ghost_idx = ghost_idxs.max()
+
+    # We can now make a list of all labels - this includes all topological
+    # entities: cells, facets, edges, vertices. Each has a unique index.
+    max_idx = max(max_core_idx, max_owned_idx, max_ghost_idx)
+    labels = np.zeros(max_idx + 1, dtype=IntType)
+    labels[core_idxs] = 1
+    labels[owned_idxs] = 2
+    labels[ghost_idxs] = 3
+
+    # We know the parent DM cell index for each of our swarm points. We can
+    # therefore filter the list of all labels to find the corresponding label
+    # of each swarm point.
+    swarm_plex_cells = swarm.getField("DMSwarm_cellid")
+    swarm.restoreField("DMSwarm_cellid")
+    swarm_parent_cell_labels = labels[swarm_plex_cells]
+    assert len(swarm_parent_cell_labels) == len(swarm_plex_cells)
+    for label_idx, label in enumerate(swarm_parent_cell_labels):
+        # We set the label using label index since this index is shared across
+        # all DMSwarm fields: label index n into a given field (such as
+        # DMSwarmPIC_coor) always corresponds to the same point in the swarm.
+        if label == 1:
+            CHKERR(DMLabelSetValue(swarm_label_core, label_idx, 1))
+        elif label == 2:
+            CHKERR(DMLabelSetValue(swarm_label_owned, label_idx, 1))
+        elif label == 3:
+            CHKERR(DMLabelSetValue(swarm_label_ghost, label_idx, 1))
+        else:
+            raise RuntimeError("Unknown label value")
+    return
+
+
+@cython.boundscheck(False)
+@cython.wraparound(False)
 def get_entity_classes(PETSc.DM dm):
     """Builds PyOP2 entity class offsets for all entity levels.
 
--- a/firedrake/cython/dmcommon.pyx
+++ b/firedrake/cython/dmcommon.pyx
@@ -33,9 +33,9 @@ def get_topological_dimension(PETSc.DM dm):
 
     :returns: For a DMPlex ``dm.getDimension()``, for a DMSwarm ``0``.
     """
-    if type(dm) is PETSc.DMPlex:
+    if isinstance(dm, PETSc.DMPlex):
         return dm.getDimension()
-    elif type(dm) is PETSc.DMSwarm:
+    elif isinstance(dm, PETSc.DMSwarm):
         return 0
     else:
         raise ValueError("dm must be a DMPlex or DMSwarm")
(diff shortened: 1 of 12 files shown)


firedrake/functionspace.py
```diff
@@ -169,7 +169,10 @@ def VectorFunctionSpace(mesh, family, degree=None, dim=None,
 
     """
     sub_element = make_scalar_element(mesh, family, degree, vfamily, vdegree)
-    dim = dim or mesh.ufl_cell().geometric_dimension()
+    if dim is None:
+        dim = mesh.ufl_cell().geometric_dimension()
+    if dim <= 0:
```
dham, line 174:
```suggestion
    if not isinstance(dim, numbers.Integral) and dim > 0:
```

firedrake/cython/dmcommon.pyx
```diff
@@ -1748,6 +1753,88 @@ def mark_entity_classes(PETSc.DM dm):
...
+    labels[ghost_idxs] = 3
+
+    # We know the parent DM cell index for each of our swarm points. We can
+    # therefore filter the list of all labels to find the corresponding label
+    # of each swarm point.
+    swarm_plex_cells = swarm.getField("DMSwarm_cellid")
+    swarm.restoreField("DMSwarm_cellid")
+    swarm_parent_cell_labels = labels[swarm_plex_cells]
+    assert len(swarm_parent_cell_labels) == len(swarm_plex_cells)
+    for label_idx, label in enumerate(swarm_parent_cell_labels):
```
wence-, line 1823: You'll want to add typing to `swarm_parent_cell_labels` and `label_idx` so that this does a C-like loop over the numpy array, I think.

firedrake/interpolation.py
```diff
@@ -85,6 +87,35 @@ class Interpolator(object):
...
+                "Cannot interpolate an expression with %d arguments" % len(arguments)
+            )
+        if (
+            target_mesh is not source_mesh
+            and isinstance(target_mesh.topology, firedrake.mesh.VertexOnlyMeshTopology)
+            and isinstance(source_mesh.topology, firedrake.mesh.VertexOnlyMeshTopology)
+        ):
+            self.vom_onto_vom = True
+        else:
+            self.vom_onto_vom = False
```
wence-, line 117: The pattern:
```
if condition:
    foo = True
else:
    foo = False
```
is generally better written as:
```
foo = condition
```

Also `vom_onto_vom` is an excellent variable name.

dham (changes requested):
There are just a few minor things I found. I think Lawrence's comment is still to do as well.

firedrake/interpolation.py
```diff
@@ -130,16 +170,27 @@ def interpolate(self, *function, output=None, transpose=False):
 
         if self.nargs:
             function, = function
-            if transpose:
-                mul = assembled_interpolator.handle.multTranspose
-                V = self.arguments[0].function_space()
+            if not hasattr(function, "dat"):
```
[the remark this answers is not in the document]
ReubenHill: I figured since any type with a suitably sized dat ought to work (such as a `Vector`) we might as well just check for that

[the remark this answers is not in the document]
ReubenHill: I could make the check be
```
if not hasattr(function, "dat") or not isinstance(function.dat, op2.Dat):
```
and/or change the error message

firedrake/interpolation.py
```diff
@@ -130,16 +170,27 @@ def interpolate(self, *function, output=None, transpose=False):
...
-            result = output or firedrake.Function(V)
-            with function.dat.vec_ro as x, result.dat.vec_wo as out:
-                mul(x, out)
-            return result
+                if transpose:
+                    V = self.arguments[0].function_space()
+                else:
+                    V = self.V
+                result = output or firedrake.Function(V)
+                assembled_interpolator(result.dat, function.dat, transpose)
```
ksagiyam, line 192: Can we somehow avoid having `make_interpolator` return `mom_to_other_vom` by making fake `handle.mult` and `handle.multTranspose` available, and remove `if` here?

ReubenHill: I think with some suitable edits (perhaps introducing a dummy `handle` class, such as `VomOntoVomDummyMat` with `mult` and `multTranspose` properties) here
https://github.com/firedrakeproject/firedrake/blob/89659fc8ade94aab66cedc53a7fecbafa4a94f4d/firedrake/interpolation.py#L286-L295
I should be able to.

Style question: should I define the class inside the `make_interpolator` (if that's even allowed) or would it be better to put it close to `VomOntoVomCallable`? I presume the latter?

ReubenHill: An issue: I'd need to modify `VomOntoVomCallable` to work with PETSc `vec`s instead of `dat`s. I don't know how I would get the correct MPI type in that case. Currently I have the below:
https://github.com/firedrakeproject/firedrake/blob/89659fc8ade94aab66cedc53a7fecbafa4a94f4d/firedrake/interpolation.py#L860
Are you aware of a way to get the appropriate MPI type in this situation? If not, your comment might not be actionable at the moment...

ksagiyam: It would be better to put the `handle` class close to `VomOntoVomCallable`.

MPI data type does not change as we do not make any change to the underlying data. I think your question is how to get numpy array from vec to use the `petsc4py` interface of `sf.bcast` and friends. I thought we could use `vec.getArray()`, but I realised `vec` does not include data on halo, which you need, so what I suggested is probably not as simple as I had thought.

ReubenHill: Oh so we don't generally include halos when interpolating? It's possible that I could make the `input_ordering_sf` not include halos in the leaves (only the leaves include halos). A halo exchange would then force halo dofs to be updated. Since we already aren't including halos I presume we don't force a halo exchange after interpolation?

firedrake/mesh.py
```diff
@@ -2282,6 +2310,59 @@ def mark_entities(self, f, label_name, label_value):
...
+        """
+        if not isinstance(self.topology, VertexOnlyMeshTopology):
+            raise AttributeError("Input ordering is only defined for vertex-only meshes.")
+        return self._input_ordering
+
+    @utils.cached_property  # TODO: Recalculate if mesh moves
+    def input_ordering_sf(self):
+        """
+        Return a PETSc SF which has :func:`~.VertexOnlyMesh` input ordering
+        vertices as roots and this mesh's vertices as leaves.
```
ksagiyam, line 2338: When we (re)distribute a `plex` to get `distplex` using `DMPlexDistribute`, it returns an SF called  `distsf`, whose roots and leaves are points in `plex` and those in `distplex`. I think a `swarm` and what you call `input_ordering` are very similar to `plex` and `distplex`, and `input_ordering_sf` compares to `distsf` (though `distsf` is from point to point, while `input_ordering_sf` is from firedrake_number to firedrake_number ?) So I think it is more natural to have something like:
```
class VertexOnlyMeshTopology():
    ...
    self._distribution_parent = ...  # `None` for what you currently call `input_ordering` and
                                                  # `input_ordering` for the vom
    self._distribution_sf = None # or what you call `input_ordering_sf`
```
If we ever consider distributing swam points more than once, we could have:
```
    def distribution_ancesters():
        if self._distribution_parent:
            return [self] + self._distribution_parent.distribution_ancesters()
        else:
            return [self]

    def distribution_sf():
        if self._distribution_parent:
            return self._distribution_parent.distribution_sf().compose(self._distribution_sf)
        else:
            return self._distribution_sf
```

firedrake/interpolation.py
```diff
@@ -645,3 +728,161 @@ def hash_expr(expr):
...
+        try:
+            if target_vom.input_ordering is source_vom:
+                broadcast = True
+                original_vom = target_vom
+        except AttributeError:
+            pass
+        if not (reduce or broadcast):
+            raise ValueError(
+                "The target vom and source vom must be linked by input ordering!"
+            )
```
ksagiyam, line 773: Maybe just set `input_ordering` (or whatever name you decide to use) ` = None` and remove these `try`/`except`?

ksagiyam (changes requested):

firedrake/interpolation.py
ksagiyam: Yes, that would be more consistent with interpolations on regular meshes. `input_ordering_sf` I think is standard, so we probably don't want to change it directly. Maybe we can use this: PetscSFCreateEmbeddedLeafSF https://gitlab.com/petsc/petsc/-/blob/main/src/binding/petsc4py/src/petsc4py/PETSc/SF.pyx#L323.

firedrake/mesh.py
ReubenHill: I'm not sure if these are directly comparable. Even in serial, the input ordering VOM includes points which are not found in the parent mesh (assuming you've turned off the error for this). Also, given this is very much a user facing part of the API, I think `input_ordering` is perhaps more expressive than `distribution_parent`, particularly given the point I just made?

I like your idea about being able to redistribute these in the future though - when we do that we can make `input_ordering` just return the first distribution ancestor. I question about your suggestion however, it looks like here
```python
    def distribution_ancesters():
        if self._distribution_parent:
            return [self] + self._distribution_parent.distribution_ancesters()
```
you're adding VOMs together - I don't think that's possible?

firedrake/mesh.py
```diff
@@ -1785,6 +1802,14 @@ def __init__(self, coordinates):
...
         self.variable_layers = self.extruded and topology.variable_layers
 
+        # no need to display input ordering attr for non vertex-only meshes
+        if not isinstance(topology, VertexOnlyMeshTopology):
+            try:
+                del self.input_ordering
+                del self.input_ordering_sf
+            except AttributeError:
+                pass
+
```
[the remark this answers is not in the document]
ReubenHill: Yes, I think they are.

firedrake/interpolation.py
ReubenHill: > Yes, that would be more consistent with interpolations on regular meshes. `input_ordering_sf` I think is standard, so we probably don't want to change it directly. Maybe we can use this: PetscSFCreateEmbeddedLeafSF https://gitlab.com/petsc/petsc/-/blob/main/src/binding/petsc4py/src/petsc4py/PETSc/SF.pyx#L323.

I think using `PetscSFCreateEmbeddedLeafSF` would require a-priori knowledge of the rank distribution of the leaves (recall that input ordering vertices are roots and our VOM's mesh vertices are leaves). I think, for simplicity, I'll have two properties: `input_ordering_sf`, which includes halo cells, and `input_ordering_no_halos_sf`, which does not.

firedrake/mesh.py
ksagiyam: That is a list addition, so that is to return a list of ancestors, but anyway, I see. Let's leave it as is now.

ReubenHill: Oh of course, I missed the `[]`!

ksagiyam (dismissed):

ReubenHill:
Rebased onto master to fix merge conflict in interpolation.py - @ksagiyam would you mind reapproving please?

ksagiyam (approved):

PR #2971, opened 2023-06-06
Fix names in subfunctions.

albert-oliver:
---
name: "Fix names in subfunctions."
description: 'Now each subfunction gets its name'
title: ''
labels: ''
assignees: ''

---

# Description
The motivation for this commit is to get the name of the subfunctions directly.

For example, adding names in the function w of the "Mixed formulation for the Poisson equation" example [1] (line 133):
`w = Function(W, name=["flux", "sol"])`

It gives a bogus name with the previous version:
```
$ ipython
In [1]: %run poisson_mixed.py

In [2]: u.name()
Out[2]: "['flux', 'solution'][1]"
```

While it gives the expected name with this version:
```
$ ipython
In [1]: %run poisson_mixed.py

In [2]: u.name()
Out[2]: 'solution'
```

[1] https://firedrakeproject.org/demos/poisson_mixed.py.html

## Associated Pull Requests:
- None

## Fixes Issues:
- None

# Checklist for author:

- [X] I have linted the codebase (`make lint` in the `firedrake` source directory).
- [X] My changes generate no new warnings.
- ~[ ] All of my functions and classes have appropriate docstrings.~
- ~[ ] I have commented my code where its purpose may be unclear.~
- ~[ ] I have included and updated any relevant documentation.~
- ~[ ] Documentation builds locally (`make linkcheck; make html; make latexpdf` in `firedrake/docs` directory)~
- ~[ ] I have added tests specific to the issues fixed in this PR.~
- ~[ ] I have added tests that exercise the new functionality I have introduced~
- ~[ ] Tests pass locally (`pytest tests` in the `firedrake` source directory) (useful, but not essential if you don't have suitable hardware).~
- ~[ ] I have performed a self-review of my own code using the below guidelines.~

# Checklist for reviewer:

- [ ] Docstrings present
- [ ] New tests present
- [ ] No bad "code smells"
- [ ] No issues in parallel

### DIFF
--- a/firedrake/function.py
+++ b/firedrake/function.py
@@ -110,7 +110,7 @@ class CoordinatelessFunction(ufl.Coefficient):
     def subfunctions(self):
         r"""Extract any sub :class:`Function`\s defined on the component spaces
         of this this :class:`Function`'s :class:`.FunctionSpace`."""
-        return tuple(CoordinatelessFunction(fs, dat, name="%s[%d]" % (self.name(), i))
+        return tuple(CoordinatelessFunction(fs, dat, name="%s" % (self.name()[i]))
                      for i, (fs, dat) in
                      enumerate(zip(self.function_space(), self.dat)))
 


dham:
Hmm. Is this the right behaviour? The original Function in the mixed space is a Function in its own right, so really its name should be a string not a tuple. The current behaviour then makes sense: you index into the mixed space and get the indexed function. 

I'm not sure that the name being a sequence of strings makes sense when you think about it from the perspective of the original Function.

albert-oliver:
Oh, I see. You're right, @dham, the function `w` can have a name of its own because, in the end, it's just a function (albeit a mixed one).

Also, this change can break many working programs, so feel free to reject it.

On the other hand, thinking of mixed functions as a set of functions of interest (pressure and velocity in the Stokes problem, the momentum density and the velocity in the Camassa-Holm equation, etc.) felt so natural to me that naming them as a collection of names for the underlying functions seemed the way to go.

So, is it best practice to rename the new subfunctions?
```
w = Function(W, name="flux_sol_mixed_function")
p, u = w.subfunctions
p.rename("flux")
u.rename("sol")
```

Sorry very much for wasting time and resources with this PR.

PR #2985, opened 2023-06-14
Eigensolver

yianzeng:
# Description

This PR introduces `LinearEigenproblem` and `LinearEigensolver` as analogous classes to `LinearVariationalProblem` and `LinearVariationalSolver`. 

This enables high-level programming of Finite Element Eigenproblems. The API is documented and the existing Eigenproblem demo has been ported to the new API.

Dirichlet Boundary conditions are supported, currently by shifting the consequential nullspace to a user-specified eigenvalue.

### DIFF
--- a/firedrake/eigensolver.py
+++ b/firedrake/eigensolver.py
@@ -115,9 +115,8 @@ class LinearEigensolver(OptionsManager):
         "eps_largest_real": None
     """
 
-    DEFAULT_EPS_PARAMETERS = {"st_pc_factor_shift_type": "NONZERO",
-                              "eps_type": "krylovschur",
-                              "eps_tol": 1e-10, 
+    DEFAULT_EPS_PARAMETERS = {"eps_type": "krylovschur",
+                              "eps_tol": 1e-10,
                               "eps_target": 0.0}
 
     def __init__(self, problem, n_evals, *, options_prefix=None,
--- a/firedrake/eigensolver.py
+++ b/firedrake/eigensolver.py
@@ -39,7 +39,7 @@ class LinearEigenproblem():
     amount. It is the user's responsibility to ensure that the shift is not
     close to an actual eigenvalue of the system.
     """
-    def __init__(self, A, M=None, bcs=None, bc_shift=666.0):
+    def __init__(self, A, M=None, bcs=None, bc_shift=0.0):
         if not SLEPc:
             raise ImportError(
                 "Unable to import SLEPc, eigenvalue computation not possible "
@@ -64,7 +64,7 @@ class LinearEigenproblem():
             yield from bc.dirichlet_bcs()
 
     @utils.cached_property
-    def dm(self):  # cargo cult
+    def dm(self):
         r"""Return the dm associated with the output space."""
         return self.output_space.dm
 
@@ -117,7 +117,8 @@ class LinearEigensolver(OptionsManager):
 
     DEFAULT_EPS_PARAMETERS = {"st_pc_factor_shift_type": "NONZERO",
                               "eps_type": "krylovschur",
-                              "eps_tol": 1e-10}
+                              "eps_tol": 1e-10,
+                              "eps_target": 0.0}
 
     def __init__(self, problem, n_evals, *, options_prefix=None,
                  solver_parameters=None, ncv=None, mpd=None):
@@ -131,6 +132,8 @@ class LinearEigensolver(OptionsManager):
         for key in self.DEFAULT_EPS_PARAMETERS:
             value = self.DEFAULT_EPS_PARAMETERS[key]
             solver_parameters.setdefault(key, value)
+        if self._problem.bcs:
+            solver_parameters.setdefault("st_type", "sinvert")
         super().__init__(solver_parameters, options_prefix)
         self.set_from_options(self.es)
 
--- a/firedrake/eigensolver.py
+++ b/firedrake/eigensolver.py
@@ -13,7 +13,7 @@ __all__ = ["LinearEigenproblem",
 
 
 class LinearEigenproblem():
-    """Linear eigenvalue problem
+    """Generalised linear eigenvalue problem
 
     The problem has the form::
 
@@ -88,12 +88,30 @@ class LinearEigensolver(OptionsManager):
     mpd : int
         Maximum dimension allowed for the projected problem. See
         `SLEPc.EPS.setDimensions`.
+
+    Notes
+    -----
+
+    Users will typically wish to set solver parameters specifying the symmetry
+    of the eigenproblem and which eigenvalues to search for first.
+
+    The former is set using the options available for `EPSSetProblemType
+    <https://slepc.upv.es/documentation/current/docs/manualpages/EPS/EPSSetProblemType.html>`__. 
+
+    For example if the bilinear form is symmetric (Hermitian in complex mode),
+    one would add this entry to `solver_options`::
+
+        "eps_gen_hermitian": None
+
+    As always when specifying PETSc options, `None` indicates that the option
+    in question is a flag and hence doesn't take an argument.
+
+    The eigenvalues to search for first are specified using the options for
+    `EPSSetWhichEigenPairs`. FIXME
     """
 
-    DEFAULT_EPS_PARAMETERS = {"eps_gen_non_hermitian": None,
-                              "st_pc_factor_shift_type": "NONZERO",
+    DEFAULT_EPS_PARAMETERS = {"st_pc_factor_shift_type": "NONZERO",
                               "eps_type": "krylovschur",
-                              "eps_largest_imaginary": None,
                               "eps_tol": 1e-10}
 
     def __init__(self, problem, n_evals, *, options_prefix=None,
--- a/firedrake/eigensolver.py
+++ b/firedrake/eigensolver.py
@@ -1,3 +1,4 @@
+"""Specify and solve finite element eigenproblems."""
 from firedrake.assemble import assemble
 from firedrake.function import Function
 from firedrake import utils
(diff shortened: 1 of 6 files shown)


firedrake/eigensolver.py
```diff
@@ -0,0 +1,169 @@
...
+    """Linear eigenvalue problem
+
+    The problem has the form::
+
+        A(u, v) = λ * M(u, v)
+
+    Parameters
+    ----------
+    A : ufl.Form
+        the bilinear form A(u, v)
```
ksagiyam, line 25:
```suggestion
        The bilinear form A(u, v).
```

firedrake/eigensolver.py
```diff
@@ -0,0 +1,169 @@
...
+    M : ufl.Form
+        the mass form M(u, v),  defaults to u * v * dx.
```
ksagiyam, line 27:
```suggestion
        The mass form M(u, v),  defaults to u * v * dx.
```

firedrake/eigensolver.py
```diff
@@ -0,0 +1,169 @@
...
+    bcs : DirichletBC or list of DirichletBC
+        the boundary conditions
```
ksagiyam, line 29:
```suggestion
        The boundary conditions.
```

firedrake/eigensolver.py
```diff
@@ -0,0 +1,169 @@
...
+    bc_shift: float
+        the value to shift the boundary condition eigenvalues by.
```
ksagiyam, line 31:
```suggestion
        The value to shift the boundary condition eigenvalues by.
```

firedrake/eigensolver.py
```diff
@@ -0,0 +1,169 @@
...
+    def __init__(self, A, M=None, bcs=None, bc_shift=666.0):
+        if not SLEPc:
+            raise ImportError(
+                "Unable to import SLEPc, eigenvalue computation not possible "
+                "(try firedrake-update --slepc)"
+            )
+
+        self.A = A  # LHS
+        args = A.arguments()
+        v, u = args[0], args[1]
```
ksagiyam, line 51:
```suggestion
        v, u = args
```

firedrake/eigensolver.py
```diff
@@ -0,0 +1,169 @@
...
+        `SLEPc.EPS.setDimensions`.
+    mpd : int
+        Maximum dimension allowed for the projected problem. See
+        `SLEPc.EPS.setDimensions`.
+    """
+
+    DEFAULT_EPS_PARAMETERS = {"eps_gen_non_hermitian": None,
+                              "st_pc_factor_shift_type": "NONZERO",
+                              "eps_type": "krylovschur",
+                              "eps_largest_imaginary": None,
```
ksagiyam, line 96: Maybe we do not need to include this item in `DEFAULT_EPS_PARAMETERS`. Otherwise, I think we would have to remove this item explicitly when user provided one of the mutually exclusive items: https://slepc.upv.es/documentation/current/docs/manualpages/EPS/EPSSetWhichEigenpairs.html.

tests/regression/test_eigensolver.py
```diff
@@ -0,0 +1,52 @@
...
+    u = TrialFunction(V)
+    v = TestFunction(V)
+    a = (inner(grad(u), grad(v))) * dx
+
+    # Create eigenproblem with boundary conditions
+    bc = DirichletBC(V, 0.0, "on_boundary")
+    eigenprob = LinearEigenproblem(a, bcs=bc)
+
+    # Create corresponding eigensolver, looking for n eigenvalues
+    eigensolver = LinearEigensolver(eigenprob, n)
```
ksagiyam, line 24:
```suggestion
    eigenprob = LinearEigenproblem(a, bcs=bc, bc_shift=-666.)

    # Create corresponding eigensolver, looking for n eigenvalues
    eigensolver = LinearEigensolver(eigenprob, n, solver_parameters={"eps_largest_real": None})
```
to make this test work consistently, I think. (There are `n + 1` eigenvalues in total, of which two are `bc_shift`).

ksagiyam (changes requested):
Looks good. Left some comments. Thanks!

firedrake/eigensolver.py
ksagiyam: I mean this item: `"eps_largest_imaginary": None,`

firedrake/eigensolver.py
```diff
@@ -0,0 +1,187 @@
...
+    Notes
+    -----
+
+    If Dirichlet boundary conditions are supplied then these will result in the
+    eigenproblem having a nullspace spanned by the basis functions with support
+    on the boundary. To facilitate solution, this is shifted by the specified
+    amount. It is the user's responsibility to ensure that the shift is not
+    close to an actual eigenvalue of the system.
+    """
+    def __init__(self, A, M=None, bcs=None, bc_shift=666.0):
```
UZerbinati, line 42: Why the bc_shift by default, not zero? This is would give an Inf eigenvalue for each bc node.

tests/regression/test_eigensolver.py
UZerbinati: `bc_shift` set to zero should also do the trick and be more in keeping with the mathematical theory.

firedrake/eigensolver.py
dham: You answered your own question. If `bc_shift` is zero then the boundary eigenvalues are mathematically undefined. Expecting a numerical algorithm to do something sensible in those circumstances is optimistic, to say the least. In particular, the way SLEPc will solve the generalised eigenvalue problem is $M^{-1}Au = \lambda u$. If `bc_shift==0` then $M$ has rows which are zero and the solve $M^{-1}(Au)$ is highly likely to fail. Using a finite shift makes the problem non-singular.

firedrake/eigensolver.py
```diff
@@ -107,7 +107,12 @@ class LinearEigensolver(OptionsManager):
     in question is a flag and hence doesn't take an argument.
 
     The eigenvalues to search for first are specified using the options for
-    `EPSSetWhichEigenPairs`. FIXME
+    `EPSSetWhichEigenPairs <https://slepc.upv.es/documentation/current/docs/manualpages/EPS/EPSSetWhichEigenpairs.html>`__.
+
+    For example, to look for the eigenvalues with largest real part, one
+    would add this entry to `solver_options`::
+
+        "eps_largest_magnitude": None
```
ksagiyam, line 115:
```suggestion
        "eps_largest_real": None
```

ksagiyam (dismissed):

firedrake/eigensolver.py
```diff
@@ -0,0 +1,192 @@
...
+            )
+
+        self.A = A  # LHS
+        args = A.arguments()
+        v, u = args
+        if M:
+            self.M = M
+        else:
+            from ufl import inner, dx
+            self.M = inner(u, v) * dx
```
UZerbinati, line 56: This is very specific. One might be interested in using the eigensolver to compute the eigenvalue of a saddle point problem, which usually has a different mass matrix with a 0 block on the diagonal.

firedrake/eigensolver.py
```diff
@@ -0,0 +1,192 @@
...
+    The eigenvalues to search for first are specified using the options for
+    `EPSSetWhichEigenPairs <https://slepc.upv.es/documentation/current/docs/manualpages/EPS/EPSSetWhichEigenpairs.html>`__.
+
+    For example, to look for the eigenvalues with largest real part, one
+    would add this entry to `solver_options`::
+
+        "eps_largest_real": None
+    """
+
+    DEFAULT_EPS_PARAMETERS = {"st_pc_factor_shift_type": "NONZERO",
```
UZerbinati, line 118: `"st_pc_factor_shift_type": "NONZERO" should only be used in case we have prior knowledge that the mass matrix has a zero pivot, no?

tests/regression/test_eigensolver.py
UZerbinati, line 1: Maybe it would be helpful to have a 2D test to check eigenvalues convergence.

firedrake/eigensolver.py
UZerbinati: Let me be more precise, my understanding is that if the `bc_shift` is zero only on the mass matrix there can not be any boundary eigenvalue because Inf is not a real number.
If a matrix in the pencil of the generalised eigenvalue problem is singular this doesn't mean that the pencil itself is singular.
Indeed SLEPc will solve the generalised eigenvalue problem as $M^{-1}Au=\lambda u$ by default, but my understanding is that this is incorrect if your $M$ is singular (for obvious reasons). In my opinion, the correct thing to do is to use a spectral transformation and my understanding is that the standard choice in this case is the SINVERT spectral transform, i.e. SLEPc will solve the generalised eigenvalue problem as $(A-\sigma M)^{-1}M u = \lambda u$. If we have a zero shift is equivalent to inverting the role of the mass and stiffness matrix, this is the usual hand-waving solution when $M$ is singular (for example saddle point problems) which if we use a spectral transform can be viewed from a more formal point of view.

dham: I've made this change, however you still don't get a particularly nice system, because of the 0 eigenvalues in the inverted system. If you want to see this, change the test to not shift the eigenvalues and you'll see that the error in the non-boundary eigenvalues increases by a lot.

firedrake/eigensolver.py
dham: It's only a default, and one that is correct in the straightforward case. The user can always provide a different `M`.

tests/regression/test_eigensolver.py
```diff
@@ -0,0 +1,54 @@
...
+    for k in range(ncov-1):
+        true_val = 6 / h**2
+        # k+1 because we skip the trivial 0 eigenvalue
+        true_val *= (1-cos((k+1)*h))/(2+cos((k+1)*h))
+        true_values[k] = true_val
+
+        estimates[k] = eigensolver.eigenvalue(k).real
+
+    true_values[-1] = eigenprob.bc_shift
+
```
colinjcotter, line 42: why are you not also shifting the 0 entry?


and, is it safe to assume this numbering?

dham: The 0 entry is not a boundary entry. The other boundary is off the end of the array (i.e. we don't look at it).

I think the ordering is safe because of the solver option to look for largest values (at any rate, the observation is that it seems stable).

PR #3003, opened 2023-06-23
add missing arguments to locator

dham:
The locator function created by `Mesh._c_locator` is missing two arguments.

Fixes #3001.


JDBetteridge (approved):
Does this functionality need a test if all tests were passing?

ksagiyam (approved):

dham:
> Does this functionality need a test if all tests were passing?

I don't know how to test this. This error was causing illegal function call behaviour at the C level. That's very hard to write deterministic tests for.

connorjward:
> > Does this functionality need a test if all tests were passing?
> 
> I don't know how to test this. This error was causing illegal function call behaviour at the C level. That's very hard to write deterministic tests for.

The right test here is probably "run CI for MacOS".

Issue #2999, opened 2023-06-21
BUG: Decreased performance when assembling expressions with new constants


jrmaddison:
**Describe the bug**
Assembly of expressions using constants can be slow if new `Constant` objects are instantiated.

**Steps to Reproduce**
Example code:
```
from firedrake import *

import time

mesh = UnitIntervalMesh(1)

# Same Constant, same value
t0 = time.perf_counter()
c = Constant(0.0)
for _ in range(100):
    assemble(c * dx(mesh))
t1 = time.perf_counter()
print(f"{t1 - t0}")

# Different Constant, same value
t0 = time.perf_counter()
for _ in range(100):
    assemble(Constant(0.0) * dx(mesh))
t1 = time.perf_counter()
print(f"{t1 - t0}")

# Same Constant, different value
t0 = time.perf_counter()
c = Constant(0.0)
for i in range(100):
    c.assign(i)
    assemble(c * dx(mesh))
t1 = time.perf_counter()
print(f"{t1 - t0}")
```

With revision c5cd37ad718da017f3035a5a7c6f5329d60e62aa and an empty cache:
```
0.5365729830009514
0.09846351200394565
0.08098157500353409
```

With revision a94b01c4b3361db9c73056d92fdbd01a5bc6d1aa and an empty cache:
```
0.5236760989937466
39.5197532769962
1.8689304560029996
```

**Environment:**
Ubuntu 22.04

connorjward:
Is this fixed by https://github.com/firedrakeproject/firedrake/pull/2996?

jrmaddison:
That raises an exception:
```
Traceback (most recent call last):
  File "/home/maddison/build/tlm_adjoint/tlm_adjoint/tmp/test.py", line 11, in <module>
    assemble(c * dx(mesh))
  File "/home/maddison/firedrake/firedrake/src/ufl/ufl/measure.py", line 423, in __rmul__
    return Form([integral])
  File "/home/maddison/firedrake/firedrake/src/ufl/ufl/form.py", line 294, in __init__
    self._constants = extract_constants(self)
  File "/home/maddison/firedrake/firedrake/src/ufl/ufl/algorithms/analysis.py", line 116, in extract_constants
    return sorted_by_count(extract_type(a, Constant))
  File "/home/maddison/firedrake/firedrake/src/ufl/ufl/algorithms/analysis.py", line 70, in extract_type
    return set(o for e in iter_expressions(a)
[...]

PR #3011, opened 2023-06-29
Fix Constant numbering

connorjward:
Fixes https://github.com/firedrakeproject/firedrake/issues/2999.

~Requires https://github.com/FEniCS/ufl/pull/178~ (and for our fork to be updated).

I think that wider problems exist regarding our implementation of `Constant` (see [next week's meeting](https://github.com/firedrakeproject/firedrake/wiki/Firedrake-meeting-2023-07-05#cw-codegen-issues-with-the-new-constant)), but this will resolve the performance problems.


.github/workflows/build.yml
```diff
@@ -79,6 +79,7 @@ jobs:
             --install icepack \
             --install irksome \
             --install femlium \
+            --package-branch ufl connorjward/counted-mixin2 \
```
connorjward, line 82:
```suggestion
            --package-branch ufl connorjward/counted-mixin \
```

.github/workflows/build.yml
```diff
@@ -79,6 +79,7 @@ jobs:
             --install icepack \
             --install irksome \
             --install femlium \
+            --package-branch ufl connorjward/test-upstream \
```
JDBetteridge, line 82: Drop before merge

JDBetteridge (changes requested):
I think this is good to go once our UFL branch is moved forwards

.github/workflows/build.yml
connorjward:
```suggestion
```

PR #3016, opened 2023-07-10
Delegate checkpoint only if the meshes are equal

### LINKED ISSUE #3009: BUG: domain of Constant not preserved in adjoint replay
**Describe the bug**
When assigning to a Constant T the value of a Constant S on a different domain, it appears that in the replay the checkpointed value of T after the assignment has taken the domain of S. I think this is because the checkpointed value of T just points to the checkpointed value of S (through a DeferredCheckpoint). This means T ends up on the wrong domain.

**Steps to Reproduce**
Consider the following example where S is on a normal FEM mesh, and T is on a vertexonly mesh mesh0. The functional integrates T over mesh0. This works correctly in the initial run of the model, but in the replay it errors in the assembly of that integral complaining `NotImplementedError: Assembly with multiple meshes is not supported` and digging into the 0-form it's trying to assemble at that point it does appear that the checkpointed value of T it is using there is on mesh instead of mesh0
```
from firedrake import *
mesh = UnitSquareMesh(10,10)
mesh0 = VertexOnlyMesh(mesh, [[.5,.5]])
S = Constant(1.0, domain=mesh)
T = Constant(2.0, domain=mesh0)
T.assign(S)
J = assemble(T*dx(domain=mesh0))
rf = ReducedFunctional(J, Control(S))
rf(Constant(3.0, domain=mesh))
```
When constructing `T` immediately from the value of `S`:
```
T = Constant(S, domain=mesh0)
```
the same error is produced.

Ig-dolci:
---
name: "Bug fix"

---

# Description
Delegate checkpoint only if the meshes are equal

## Fixes Issues:
- Make a list of issues (with links!)
- [Like this Issue#3009](https://github.com/firedrakeproject/firedrake/issues/3009)

"fixes #3009".

# Checklist for author:

# Checklist for reviewer:

- [ ] Docstrings present
- [ ] New tests present
- [ ] No bad "code smells"
- [ ] No issues in parallel

### DIFF
--- a/firedrake/adjoint/function.py
+++ b/firedrake/adjoint/function.py
@@ -121,8 +121,11 @@ class FunctionMixin(FloatingType):
                 block_var = self.create_block_variable()
                 block.add_output(block_var)
 
-                if isinstance(other, type(self)):
-                    block_var._checkpoint = DelegatedFunctionCheckpoint(other.block_variable)
+                if (isinstance(other, type(self))):
+                    other_mesh = other.function_space().mesh()
+                    self_mesh = self.function_space().mesh()
+                    if other_mesh == self_mesh:
+                        block_var._checkpoint = DelegatedFunctionCheckpoint(other.block_variable)
 
             return ret
 
--- a/firedrake/adjoint/function.py
+++ b/firedrake/adjoint/function.py
@@ -122,9 +122,7 @@ class FunctionMixin(FloatingType):
                 block.add_output(block_var)
 
                 if (isinstance(other, type(self))):
-                    other_mesh = other.function_space().mesh()
-                    self_mesh = self.function_space().mesh()
-                    if other_mesh == self_mesh:
+                    if other.function_space().mesh() == self.function_space().mesh():
                         block_var._checkpoint = DelegatedFunctionCheckpoint(other.block_variable)
 
             return ret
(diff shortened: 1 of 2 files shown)


firedrake/adjoint/function.py
```diff
@@ -121,8 +121,11 @@ def wrapper(self, other, *args, **kwargs):
                 block_var = self.create_block_variable()
                 block.add_output(block_var)
 
-                if isinstance(other, type(self)):
-                    block_var._checkpoint = DelegatedFunctionCheckpoint(other.block_variable)
+                if (isinstance(other, type(self))):
+                    other_mesh = other.function_space().mesh()
+                    self_mesh = self.function_space().mesh()
+                    if other_mesh == self_mesh:
```
dham, line 127:
```suggestion
                    if other.function_space().mesh() == self.function_space().mesh():
```
Don't introduce unnecessary variables.

dham (changes requested):
Please also add a test which checks for the behaviour reported in #3009 .

Ig-dolci:
> Please also add a test which checks for the behaviour reported in #3009 .

Should I add in pyadjoint/tests/firedrake_adjoint or firedrake/tests?

Ig-dolci:
I observed the test errors with tests/regression/test_netgen.py also in the master branch. Do you know if these errors are expected?

PR #3020, opened 2023-07-11
Documentation for solving on mean zero subspace

### LINKED ISSUE #1329: Orthogonalising against nullspace does subtly wrong thing
# Problem
When we provide a nullspace to a solver, PETSc orthogonalises the solution wrt to the provided null space basis in the _l2_ norm.

However, we don't actually typically want to use _l2_, but perhaps rather _L2_ (or H1, or something else).

# Proposed solution

1. we pass the discrete (Euclidean) interpretation of the null space to PETSc which removes it in the solve.
2. after the solve terminates, Firedrake orthogonalises the provided solution against the provided null space wrt a (user-specified) inner product, defaulting to L2.

Thoughts?

colinjcotter:
# Description
Add documentation to explain how to solve on mean zero subspace on nonuniform meshes.

## Fixes Issues:
- fixes #1329

### DIFF
--- a/CITATION.rst
+++ b/CITATION.rst
@@ -17,7 +17,6 @@ If you publish results using Firedrake, we would be grateful if you would cite t
     edition      = {First edition},
     year         = {2023},
     month        = {5},
-    doi          = {10.25561/104839},
   }
 
 The simplest way to determine any additional relevant papers to cite is by asking Firedrake itself. You can ask that a list of citations relevant to your computation be printed when exiting by calling `Citations.print_at_exit` after importing Firedrake::
--- a/docs/source/solving-interface.rst
+++ b/docs/source/solving-interface.rst
@@ -838,13 +838,13 @@ in the transpose nullspace by providing a
 ``transpose_nullspace`` keyword argument to :func:`~.solve`.
 
 Achieving exact mean zero on nonuniform meshes
-----------------------------------------------
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
 Trying the above example on a nonuniform mesh with variable cell areas
 will give a solution that does not have zero spatial mean. This is
 because PETSc uses the usual :math:`\mathbb{R}^M` inner product on the
 vectors containing the basis coefficients. In general the correct
-thing to do is to specify a basis for :math:`U \subspace V'`, a
+thing to do is to specify a basis for :math:`U \subset V'`, a
 subspace to the dual space to :math:`V`. Then the subspace we wish to
 solve the system on is :math:`\hat{V} = U^0`, the "polar space" of
 :math:`U` in :math:`V`, defined by
@@ -853,7 +853,7 @@ solve the system on is :math:`\hat{V} = U^0`, the "polar space" of
 
    U^0 = \{v \in V: F[v]=0, \, \forall F \in U\}.
 
-For example, if we want to use the zero mean subspace of $V$, we
+For example, if we want to use the zero mean subspace of :math:`V`, we
 take the basis :math:`\{F_0\}`, where
 
 .. math::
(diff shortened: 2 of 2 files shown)


CITATION.rst
```diff
@@ -17,7 +17,6 @@ If you publish results using Firedrake, we would be grateful if you would cite t
     edition      = {First edition},
     year         = {2023},
     month        = {5},
-    doi          = {10.25561/104839},
   }
```
wence-, line 20: Don't want this, I think.

docs/source/solving-interface.rst
```diff
@@ -837,8 +837,68 @@ in the transpose nullspace by providing a
...
+assembled linear forms in the basis for :math:`U`. These are vectors
+where the i-th entry is equal to :math:`F_0[\phi_i]` where :math:`\phi_i`
+is the i-th basis vector of :math:`V``. For example, if we want to use
+:math:`F_0` above, we just use
+
+ ..  code-block:: python3
+
+     v = TestFunction(V)
+     F0 = assemble(v*dx)
+     cobasis = VectorSpaceBasis([F0])
```
wence-, line 873: This is not correct because the basis given to PETSc is required to be orthonormal, not just orthogonal (and that orthonormality is in the R^M inner product). So you probably want to pointwise divide by:
```
with F0.dat.vec_ro as v:
    factor = v.norm()
```

docs/source/solving-interface.rst
```diff
@@ -837,8 +837,68 @@ in the transpose nullspace by providing a
 :class:`~firedrake.nullspace.VectorSpaceBasis` with the
 ``transpose_nullspace`` keyword argument to :func:`~.solve`.
 
-Singular operators in mixed spaces
-~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+Achieving exact mean zero on nonuniform meshes
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Trying the above example on a nonuniform mesh with variable cell areas
+will give a solution that does not have zero spatial mean. This is
```
wence-, line 844: Mention that it does give a solution that has satisfies $\int u dx = c$ for some $c = O(h^2)$ (i.e. it converges to zero at second order)?

stephankramer:
I might be misunderstanding what you are suggesting, but I don't really understand how this is correct. It might be projecting out any vector you give it, but it's still doing a little l2 projection: r -> r - (v^Tr / v^T v)*v which means you're adding something to the solution that's not in the (right) null space, no?

Consider:
```
from firedrake import *
mesh = UnitSquareMesh(10, 10)
mesh.coordinates.interpolate(as_vector((x**2, y**2)))
u = Function(V)
v = TestFunction(V)

f = interpolate(x+y, V)
# ensure consistent rhs by projecting out constant:
favg = assemble(f)/assemble(Constant(1)*dx(domain=mesh))
f.assign(f - favg)
print("int f * dx = ", assemble(f*dx))

F = dot(grad(u), grad(v)) * dx - v*f*dx
nullspace = VectorSpaceBasis([assemble(v*dx)])
nullspace.orthonormalize()

solve(F==0, u, solver_parameters={'snes_type': 'ksponly', 'ksp_type': 'cg', 'pc_type': 'sor'},
      nullspace=nullspace)

print("residual F: ", assemble(F).dat.norm)
```

CITATION.rst
colinjcotter: I do not know where that came from!

colinjcotter:
Hi Stephan,
  I'm trying to work this out. When I replace the nullspace in your example with constant=True, and monitor the ksp, I get

    0 KSP Residual norm 6.796354205516e-16
    1 KSP Residual norm 2.156109514944e-16
    2 KSP Residual norm 9.246259427692e-17
    3 KSP Residual norm 8.214587397105e-17
    4 KSP Residual norm 8.062467287413e-17

which I don't understand!

________________________________________

stephankramer:
Ah, that's because my code is dodgy: should be `favg = assemble(f*dx)/assemble(Constant(1)*dx(domain=mesh))` (was missing a `dx`)

stephankramer:
If you fix my code - my example does seem to work with your nullspace but it also works without any nullspace provided so that's probably not saying much. If you use a more complex preconditioner like 'gamg' it does break with your nullspace

colinjcotter:
If I try your example with GMRES instead of CG then it works fine. So this is rather subtle!

colinjcotter:
Yes, this projection method only works with left preconditioning. OK, I think we can close this for now! Thanks for looking at it.

PR #3038, opened 2023-07-25
Make two of the spherical meshes more general

tommbendall:
This is a draft PR for extending the `IcosahedralSphereMesh` and the `CubedSphereMesh` to have a general number of cells along a panel edge.


dham (changes requested):
please also check that this isn't a performance disaster.

PR #3041, opened 2023-07-26
FML

JDBetteridge:
FML

### DIFF
--- a/firedrake/fml/__init__.py
+++ b/firedrake/fml/__init__.py
@@ -1,0 +1,2 @@
+from gusto.fml.form_manipulation_language import *        # noqa
+from gusto.fml.replacement import *                       # noqa
--- a/firedrake/fml/__init__.py
+++ b/firedrake/fml/__init__.py
@@ -1,2 +1,2 @@
-from gusto.fml.form_manipulation_language import *        # noqa
-from gusto.fml.replacement import *                       # noqa
+from firedrake.fml.form_manipulation_language import *        # noqa
+from firedrake.fml.replacement import *                       # noqa
--- a/firedrake/fml/form_manipulation_language.py
+++ b/firedrake/fml/form_manipulation_language.py
@@ -4,6 +4,8 @@ import ufl
 import functools
 import operator
 from firedrake import Constant, Function
+from collections.abc import Callable, Sequence, Mapping
+from typing import Any, Union
 
 
 __all__ = ["Label", "Term", "LabelledForm", "identity", "drop", "all_terms",
@@ -26,26 +28,26 @@ class Term(object):
 
     __slots__ = ["form", "labels"]
 
-    def __init__(self, form, label_dict=None):
+    def __init__(self, form: ufl.Form, label_dict: Mapping = None):
         """
         Parameters
         ----------
-        form : ufl.Form
+        form
             The form for this terms.
-        label_dict : `dict`, optional
+        label_dict
             Dictionary of key-value pairs corresponding to current form labels.
             Defaults to None.
         """
         self.form = form
         self.labels = label_dict or {}
 
-    def get(self, label):
+    def get(self, label: "Label") -> Any:
         """
         Returns the value of a label.
 
         Parameters
         ----------
-        label : Label
+        label
             The label to return the value of.
 
         Returns
@@ -55,22 +57,26 @@ class Term(object):
         """
         return self.labels.get(label.label)
 
-    def has_label(self, *labels, return_tuple=False):
+    def has_label(
+        self,
+        *labels: Union[Sequence["Label"], "Label"],
+        return_tuple: bool = False
+    ) -> Union[tuple[bool], bool]:
         """
         Whether the term has the specified labels attached to it.
 
         Parameters
         ----------
-        *labels : Label
+        *labels
             A label or series of labels. A tuple is automatically returned if
             multiple labels are provided as arguments.
-        return_tuple : `bool`, optional
+        return_tuple
             If True, forces a tuple to be returned even if only one label is
             provided as an argument. Defaults to False.
 
         Returns
         -------
-        bool or tuple
+        bool
             Booleans corresponding to whether the term has the specified labels.
         """
         if len(labels) == 1 and not return_tuple:
@@ -78,13 +84,13 @@ class Term(object):
         else:
             return tuple(self.has_label(l) for l in labels)
 
-    def __add__(self, other):
+    def __add__(self, other: Union["Term", "LabelledForm"]) -> "LabelledForm":
         """
         Adds a term or labelled form to this term.
 
         Parameters
         ----------
-        other : Term or LabelledForm
+        other
             The term or labelled form to add to this term.
 
         Returns
@@ -105,13 +111,13 @@ class Term(object):
 
     __radd__ = __add__
 
-    def __sub__(self, other):
+    def __sub__(self, other: Union["Term", "LabelledForm"]) -> "LabelledForm":
         """
         Subtracts a term or labelled form from this term.
 
         Parameters
         ----------
-        other : Term or LabelledForm
+        other
             The term or labelled form to subtract from this term.
 
         Returns
@@ -122,13 +128,16 @@ class Term(object):
         other = other * Constant(-1.0)
         return self + other
 
-    def __mul__(self, other):
+    def __mul__(
+        self,
+        other: Union[float, Constant, ufl.algebra.Product]
+    ) -> "Term":
         """
         Multiplies this term by another quantity.
 
         Parameters
         ----------
-        other : float, Constant or ufl.algebra.Product
+        other
             The quantity to multiply this term by.
 
         Returns
@@ -140,13 +149,16 @@ class Term(object):
 
     __rmul__ = __mul__
 
-    def __truediv__(self, other):
+    def __truediv__(
+        self,
+        other: Union[float, Constant, ufl.algebra.Product]
+    ) -> "Term":
         """
         Divides this term by another quantity.
 
         Parameters
         ----------
-        other : float, Constant or ufl.algebra.Product
+        other
             The quantity to divide this term by.
 
         Returns
@@ -175,7 +187,7 @@ class LabelledForm(object):
     """
     __slots__ = ["terms"]
 
-    def __init__(self, *terms):
+    def __init__(self, *terms: Sequence[Term]):
         """
         Parameters
         ----------
@@ -193,13 +205,16 @@ class LabelledForm(object):
                 raise TypeError('Can only pass terms or a LabelledForm to LabelledForm')
             self.terms = list(terms)
 
-    def __add__(self, other):
+    def __add__(
+        self,
+        other: Union[ufl.Form, Term, "LabelledForm"]
+    ) -> "LabelledForm":
         """
         Adds a form, term or labelled form to this labelled form.
 
         Parameters
         ----------
-        other : ufl.Form, Term or LabelledForm
+        other
             The form, term or labelled form to add to this labelled form.
 
         Returns
@@ -220,13 +235,16 @@ class LabelledForm(object):
 
     __radd__ = __add__
 
-    def __sub__(self, other):
+    def __sub__(
+        self,
+        other: Union[ufl.Form, Term, "LabelledForm"]
+    ) -> "LabelledForm":
         """
         Subtracts a form, term or labelled form from this labelled form.
 
         Parameters
         ----------
-        other : ufl.Form, Term or LabelledForm
+        other
             The form, term or labelled form to subtract from this labelled form.
 
         Returns
@@ -244,13 +262,16 @@ class LabelledForm(object):
             # Make new Term for other and subtract it
             return LabelledForm(*self, Term(Constant(-1.)*other))
 
-    def __mul__(self, other):
+    def __mul__(
+        self,
+        other: Union[float, Constant, ufl.algebra.Product]
+    ) -> "LabelledForm":
         """
         Multiplies this labelled form by another quantity.
 
         Parameters
         ----------
-        other : float, Constant or ufl.algebra.Product
+        other
             The quantity to multiply this labelled form by. All terms in the
             form are multiplied.
 
@@ -261,13 +282,16 @@ class LabelledForm(object):
         """
         return self.label_map(all_terms, lambda t: Term(other*t.form, t.labels))
 
-    def __truediv__(self, other):
+    def __truediv__(
+        self,
+        other: Union[float, Constant, ufl.algebra.Product]
+    ) -> "LabelledForm":
         """
         Divides this labelled form by another quantity.
 
         Parameters
         ----------
-        other : float, Constant or ufl.algebra.Product
+        other
             The quantity to divide this labelled form by. All terms in the form
             are divided.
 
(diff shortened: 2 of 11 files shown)


firedrake/fml/form_manipulation_language.py
```diff
@@ -0,0 +1,449 @@
...
+    """A Term object contains a form and its labels."""
+
+    __slots__ = ["form", "labels"]
+
+    def __init__(self, form, label_dict=None):
+        """
+        Args:
+            form (:class:`ufl.Form`): the form for this terms.
+            label_dict (dict, optional): dictionary of key-value pairs
+                corresponding to current form labels. Defaults to None.
```
dham, line 34: Please to be using numpydoc

firedrake/fml/form_manipulation_language.py
```diff
@@ -0,0 +1,449 @@
...
+        Adds a term or labelled form to this term.
+
+        Args:
+            other (:class:`Term` or :class:`LabelledForm`): the term or labelled
+                form to add to this term.
+
+        Returns:
+            :class:`LabelledForm`: a labelled form containing the terms.
+        """
+        if other is None:
```
dham, line 83:
```suggestion
        if self is NullTerm:
            return other
        if other is None or other is NullTerm:
```

dham (dismissed):

firedrake/fml/__init__.py
```diff
@@ -0,0 +1,2 @@
+from gusto.fml.form_manipulation_language import *        # noqa
+from gusto.fml.replacement import *                       # noqa
```
stephankramer, line 2: Should that be `from firedrake.fml....` ?

firedrake/fml/form_manipulation_language.py
```diff
@@ -0,0 +1,484 @@
...
+        else:
+            raise ValueError("Unable to relabel %s" % target)
+
+
+# ---------------------------------------------------------------------------- #
+# Some common labels
+# ---------------------------------------------------------------------------- #
+
+subject = Label("subject", validator=lambda value: type(value) == Function)
+name = Label("name", validator=lambda value: type(value) == str)
```
tommbendall, line 484:
```suggestion
name_label = Label("name", validator=lambda value: type(value) == str)
```

firedrake/fml/form_manipulation_language.py
```diff
@@ -0,0 +1,484 @@
+"""A language for manipulating forms using labels."""
+
+import ufl
+import functools
+import operator
+from firedrake import Constant, Function
+
+
+__all__ = ["Label", "Term", "LabelledForm", "identity", "drop", "all_terms",
+           "keep", "subject", "name"]
```
tommbendall, line 10:
```suggestion
           "keep", "subject", "name_label"]
```

tommbendall (dismissed):
I have been thinking that we should rename the label `name` to `name_label`, so that we don't add an object called `name` to Firedrake's namespace, which could end up being very confusing!

dham (changes requested):

firedrake/fml/form_manipulation_language.py
```diff
@@ -0,0 +1,484 @@
...
+class Term(object):
+    """A Term object contains a form and its labels."""
+
+    __slots__ = ["form", "labels"]
+
+    def __init__(self, form, label_dict=None):
+        """
+        Parameters
+        ----------
+        form : :class:`ufl.Form`
```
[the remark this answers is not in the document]
JDBetteridge: This example is fine if I remove `:class:` and the backticks, but immediately below:
- For `label_dict : dict, optional` if I do the same I get the error:
```
/home/jack/Documents/firedrake/firedrake/src/firedrake/firedrake/fml/form_manipulation_language.py:docstring of firedrake.fml.form_manipulation_language.Term:1:py:class reference target not found: optional
```
Optional is correct according to [numpydoc](https://numpydoc.readthedocs.io/en/latest/format.html#parameters).
- For `label_dict : <backtick>dict<backtick>, optional` the error goes away, but the auto-generated docs have no links to the docs for `dict`.
- Only with `label_dict : :obj:<backtick>dict<backtick>, optional` do I get what I thought was the correct behaviour with no error and a link to [this part of the Python docs](https://docs.python.org/3/library/stdtypes.html#dict).

Similar things happen elsewhere in the docstrings, where objects or functions are referenced and no link is generated. Am I just doing something wrong?

At least when I included the RST roles the right links were generated in the html documentation. But the latest commit has all of the roles removed.

firedrake/fml/form_manipulation_language.py
```diff
@@ -30,9 +30,9 @@ def __init__(self, form, label_dict=None):
         """
         Parameters
         ----------
-        form : :class:`ufl.Form`
+        form : ufl.Form
             The form for this terms.
-        label_dict : :obj:`dict`, optional
+        label_dict : `dict`, optional
```
ReubenHill, line 35: Need to remove backticks here too I think (and bool below etc)

ReubenHill:
```suggestion
        label_dict : dict, optional
```

firedrake/fml/form_manipulation_language.py
```diff
@@ -61,10 +61,10 @@ def has_label(self, *labels, return_tuple=False):
 
         Parameters
         ----------
-        *labels : :class:`Label`
+        *labels : Label
             A label or series of labels. A tuple is automatically returned if
             multiple labels are provided as arguments.
-        return_tuple : :obj:`bool`, optional
+        return_tuple : `bool`, optional
```
ReubenHill, line 67:
```suggestion
        return_tuple : bool, optional
```

firedrake/fml/form_manipulation_language.py
```diff
@@ -167,8 +168,8 @@ class LabelledForm(object):
     """
     A form, broken down into terms that pair individual forms with labels.
 
-    The `LabelledForm` object holds a list of terms, which pair
-    :class:`ufl.Form` objects with :class:`.Label`\\ s. The :meth:`label_map`
+    The LabelledForm object holds a list of terms, which pair
+    ufl.Form objects with .Label\\ s. The label_map
```
ReubenHill, line 172: Question: when you build the docs, does `.Label\\ s` link to the class correctly? Do you need the `.` before?

firedrake/fml/form_manipulation_language.py
```diff
@@ -293,18 +295,18 @@ def label_map(self, term_filter, map_if_true=identity,
 
         Parameters
         ----------
-        term_filter : :obj:`callable`
+        term_filter : `callable`
```
ReubenHill, line 298:
```suggestion
        term_filter : callable
```

firedrake/fml/form_manipulation_language.py
```diff
@@ -293,18 +295,18 @@ def label_map(self, term_filter, map_if_true=identity,
...
             A function to filter the labelled form's terms.
-        map_if_true : :obj:`callable`, optional
+        map_if_true : `callable`, optional
```
ReubenHill, line 300:
```suggestion
        map_if_true : callable, optional
```

firedrake/fml/form_manipulation_language.py
```diff
@@ -293,18 +295,18 @@ def label_map(self, term_filter, map_if_true=identity,
...
             How to map the terms for which the term_filter returns True.
             Defaults to identity.
-        map_if_false : :obj:`callable`, optional
+        map_if_false : `callable`, optional
```
ReubenHill, line 303:
```suggestion
        map_if_false : callable, optional
```

firedrake/fml/form_manipulation_language.py
```diff
@@ -359,10 +361,10 @@ def __init__(self, label, *, value=True, validator=None):
         ----------
         label : str
             The name of the label.
-        value : :obj:`Any`, optional
+        value : `any`, optional
```
ReubenHill, line 364:
```suggestion
        value : any, optional
```

firedrake/fml/form_manipulation_language.py
```diff
@@ -359,10 +361,10 @@ def __init__(self, label, *, value=True, validator=None):
...
             The value for the label to take. Can be any type (subject to the
             validator). Defaults to True.
-        validator : :obj:`callable`, optional
+        validator : `callable`, optional
```
ReubenHill, line 367:
```suggestion
        validator : callable, optional
```

firedrake/fml/replacement.py
```diff
@@ -131,14 +131,14 @@ def replace_test_function(new_test, old_idx=None, new_idx=None):
...
-    new_test : :func:`.TestFunction`
+    new_test : Argument
         The new test function.
 
     Returns
     -------
-    :obj:`callable`
-        A function that takes in t, a :class:`.Term`, and returns a new
-        :class:`.Term` with form containing the ``new_test`` and
+    `callable`
```
ReubenHill, line 139:
```suggestion
    callable
```

firedrake/fml/replacement.py
```diff
@@ -180,14 +180,14 @@ def replace_trial_function(new_trial, old_idx=None, new_idx=None):
...
-    new : :func:`.TrialFunction` or :class:`.Function`
+    new : Argument or firedrake.function.Function
         The new function.
 
     Returns
     -------
-    :obj:`callable`
-        A function that takes in t, a :class:`.Term`, and returns a new
-        :class:`.Term` with form containing the ``new_test`` and
+    `callable`
```
ReubenHill, line 188:
```suggestion
    callable
```

firedrake/fml/replacement.py
```diff
@@ -236,10 +236,10 @@ def replace_subject(new_subj, old_idx=None, new_idx=None):
 
     Parameters
     ----------
-    new : :class:`ufl.Expr`
+    new : ufl.Expr
         The new expression to replace the subject.
-    idx : :obj:`int`, optional
-        Index of the subject in the equation's :class:`.MixedFunctionSpace`.
+    idx : `int`, optional
```
ReubenHill, line 241:
```suggestion
    idx : int, optional
```

firedrake/fml/form_manipulation_language.py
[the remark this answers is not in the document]
JDBetteridge: @ReubenHill I agree with your suggestions, but did you read the above comment?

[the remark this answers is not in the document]
ReubenHill: No I did not. That is very strange! Does the most recent commit give you links everywhere it ought to?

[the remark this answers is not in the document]
JDBetteridge: No, the backticks just prevent the error from occurring. The only way I know how to get the link is to add the RST role tag back in. I am not a numpydoc/RST/sphinx expert, but I do find it difficult to find concrete examples/tutorials online, especially ones that cover these sorts of edge cases.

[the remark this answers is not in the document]
ReubenHill: Well if it produces the correct documentation (with links), then surely that's what you need to do. Do you get links when you don't use backticks for our internal classes?

Does this imply our documentation compiler just not properly deal with numpy-style docstrings?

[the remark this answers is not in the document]
connorjward: When we did the petsc4py documentation we omitted the type part of the numpy docstring and used type hints instead ([example](https://gitlab.com/petsc/petsc/-/blob/main/src/binding/petsc4py/src/petsc4py/PETSc/Vec.pyx?ref_type=heads#L330)). Since I expect us to start using type hints at some point is it worth even doing this?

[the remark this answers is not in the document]
connorjward: This appears to be under discussion in numpydoc: https://github.com/numpy/numpydoc/issues/196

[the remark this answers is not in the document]
JDBetteridge: Apparently the "correct" fix for this is to (as Connor has already mentioned) add type hinting, which I have now done. However, it _still_ doesn't solve the issue! It's better, but the documentation for `LabelledForm.lable_map()` which has arguments that are callables looks like:

> label_map(term_filter: ~collections.abc.Callable, map_if_true: ~collections.abc.Callable = <function <lambda>>, map_if_false: ~collections.abc.Callable = <function <lambda>>) → [LabelledForm]

Which is clearly wrong! Whatsmore I've now broken backward compatibility with Python3.8, so now I need to go and change a bunch of the type hints I've provided.

I'm wondering whether this is just a waste of time, it certainly feels like it. :slightly_frowning_face:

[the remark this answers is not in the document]
ReubenHill: I vote you leave in the RST if that provides the correct links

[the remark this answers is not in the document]
JDBetteridge: Oh, it gets even better! To allow the _code to execute_ in Python 3.8 we need to import from the `typing`module, which gives us a bunch of [links to deprecated functionality](https://docs.python.org/3/library/typing.html#typing.Callable) in our documentation.

firedrake/fml/form_manipulation_language.py
```diff
@@ -55,36 +57,40 @@ def get(self, label):
         """
         return self.labels.get(label.label)
 
-    def has_label(self, *labels, return_tuple=False):
+    def has_label(
+        self,
+        *labels: Union[Sequence["Label"], "Label"],
```
wence-, line 61: This says each entry in `labels` is either a single label, or a sequence of labels. You want:
```
*labels: "Label"
```
I suspect.

firedrake/fml/form_manipulation_language.py
```diff
@@ -280,27 +304,31 @@ def __truediv__(self, other):
...
-    def __len__(self):
+    def __len__(self) -> int:
         """Returns the number of terms in the labelled form."""
         return len(self.terms)
 
-    def label_map(self, term_filter, map_if_true=identity,
-                  map_if_false=identity):
+    def label_map(
+        self,
+        term_filter: Callable,
```
wence-, line 316:
```
term_filter: Callable[[Term], bool]
```

and `map_if_true: Callable[[Term], Optional[Term]` and similarly for `map_if_false`.

firedrake/fml/replacement.py
```diff
@@ -4,15 +4,23 @@
...
 
 # ---------------------------------------------------------------------------- #
 # A general routine for building the replacement dictionary
 # ---------------------------------------------------------------------------- #
-def _replace_dict(old, new, old_idx, new_idx, replace_type):
+def _replace_dict(
+    old: Union[Function, Argument],
+    new: Union[Function, Argument],
+    old_idx: Union[int, None],
+    new_idx: Union[int, None],
```
wence-, line 20: AKA `Optional[int]`

firedrake/fml/replacement.py
```diff
@@ -4,15 +4,23 @@
 
 import ufl
 from .form_manipulation_language import Term, subject
-from firedrake import split, MixedElement
+from firedrake import split, MixedElement, Function, Argument
+from collections.abc import Callable
+from typing import Union
```
wence-: Prefer importing `Callable` from `typing`.

firedrake/fml/replacement.py
```diff
@@ -125,36 +135,46 @@ def _replace_dict(old, new, old_idx, new_idx, replace_type):
 # ---------------------------------------------------------------------------- #
 # Replacement routines
 # ---------------------------------------------------------------------------- #
-def replace_test_function(new_test, old_idx=None, new_idx=None):
+def replace_test_function(
+    new_test: Argument,
+    old_idx: Union[int, None] = None,
+    new_idx: Union[int, None] = None
+) -> Callable:
```
wence-, line 141: `Callable[[Term], Term]`

firedrake/fml/form_manipulation_language.py
```diff
@@ -0,0 +1,523 @@
...
+        Returns
+        -------
+        Any
+            The value of a label.
+        """
+        return self.labels.get(label.label)
+
+    def has_label(
+        self,
+        *labels: Union[Sequence["Label"], "Label"],
```
dham, line 61:
```suggestion
        *labels: "Label",
```

firedrake/fml/form_manipulation_language.py
```diff
@@ -0,0 +1,523 @@
...
+
+# ---------------------------------------------------------------------------- #
+# Labelled form class
+# ---------------------------------------------------------------------------- #
+class LabelledForm(object):
+    """
+    A form, broken down into terms that pair individual forms with labels.
+
+    The LabelledForm object holds a list of terms, which pair
+    ufl.Form objects with .Label\\ s. The label_map
```
dham, line 183:
```suggestion
    :class:`ufl.Form` objects with :class:`Label` s. The label_map
```

firedrake/fml/form_manipulation_language.py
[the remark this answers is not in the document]
JDBetteridge: Fixed as best I can, html docs for `label_map` still look terrible, but using functions rather than lambdas makes things at least parsable.

firedrake/fml/form_manipulation_language.py
```diff
@@ -61,8 +129,7 @@ def has_label(
         *labels: "Label",
         return_tuple: bool = False
     ) -> Union[Tuple[bool], bool]:
-        """
-        Whether the term has the specified labels attached to it.
+        """Whether the term has the specified labels attached to it.
```
dham, line 132:
```suggestion
        """Return whether the specified labels are attached to this term.
```

dham (approved):

PR #3060, opened 2023-08-14
Reintroduce `to_petsc_local_numbering`

jwallwork23:
Hi all. I decided to take the metric-based mesh adaptation stuff out of #2796 and put it in a new repo: https://github.com/pyroteus/animate. However, it will still need the `to_petsc_local_numbering` function that was removed from Firedrake at some point. Would you be willing to put this updated version back in?

The PR also contains some minor fixes for the PETSc interface docs (introduced by me I think).


ksagiyam (approved):
Thanks! Just had some suggestions for the docstring.

PR #3062, opened 2023-08-14
Move firedrake adjoint

dham:
Remove the firedrake dependency on dolfin_adjoint_common. 

This completes the gradual transition of firedrake-specific adjoint code from pyadjoint into firedrake. This cleans up pyadjoint and should reduce the number of occasions on which it is necessary to land simultaneous PRs in Firedrake and Pyadjoint. 

We also deprecate the `firedrake_adjoint` module in favour of `firedrake.adjoint`. The latter differs from the former in that it does not automatically start taping. Instead, the user must explicitly call `continue_annotation()`. We have discussed this change in the past and decided that automatic taping causes more confusion than it's worth. The deprecation message on the pyadjoint PR tells users how to change their code and the adjoint notebooks have been updated accordingly.

In more detail, this pr does the following:

1. `firedrake.adjoint` becomes `firedrake.adjoint_utils`, which now incorporates all of the code from `dolfin_adjoint_common`. This is where the infrastructure for firedrake's adjoint now lives.
2. `firedrake.adjoint` is recreated as the public interface for the adjoint. It's essentially a copy of `firedrake_adjoint` except that it no longer automatically switches on annotation.
3. Explicilty importing `firedrake.adjoint` exposes an existing name clash between that module and `ufl.adjoint`, which is also in our top level namespace. `adjoint` is the natural name for both the module and the operator so we work around this by making the module callable as an alias for the function. The mechanism for this is currently slightly hacky but we can move to the natural mechanism in a few years once we can depend on users having python 3.12.
4. The notebooks that use the adjoint are updated accordingly.

The whole backend and compat hackery has been preserved to minimise the diff. I will change this in a future PR.

Depends on https://github.com/dolfin-adjoint/pyadjoint/pull/103


connorjward (approved):
All looks sensible enough to me. We should definitely revisit the compat/backend stuff in a later PR.

PR #3063, opened 2023-08-14
Suppress missing UFL docs warning.

dham:
A recent change to Constant is causing build fails. The cause of this is that Constant inherits from `ufl.utils.Counted`, which is not in the ufl documentation. This PR suppresses errors for missing ufl docs. We do the same for other external packages.

Along the way I have given `Constant.__init__` a blank docstring in order to prevent it inheriting from `ufl.utils.Counted`.


connorjward (approved):
I meant to get around to this at some point. Thanks for the fix!

PR #3067, opened 2023-08-18
Parallel Compatible Cross-Mesh Interpolation

ReubenHill:
# Description
This implements cross-mesh interpolation for interpolating into function spaces which have point evaluation nodes. This is an attempt to integrate [this gist](https://gist.github.com/ReubenHill/15cb3e7fd16255f966a2cbae80250e24) into Firedrake and makes use of the recent merging of #2936.

We use `VertexOnlyMesh` as an intermediary for the global locations of the point evaluation nodes of the target function space:
 1. We get the point evaluation node locations using the method described in the [interpolation from external data section of the manual](https://www.firedrakeproject.org/interpolation.html#interpolation-from-external-data). **This will have the parallel domain decomposition of the target mesh**.
 2. Next we create a `VertexOnlyMesh` (A) at those locations **within the source mesh** such that we inherit the source mesh's parallel domain decomposition.
 3.  We interpolate our expression in our source function space onto a `P0DG` function space on `VertexOnlyMesh` (A), which has the effect of point evaluating at the target function space node locations.
 4. This `VertexOnlyMesh` (A) has an `input_ordering` `VertexOnlyMesh` (B) whose vertices have the ordering and parallel domain decomposition of the target function space global node locations. We interpolate from `P0DG` on (A) onto `P0DG` on (B). Under the hood, this is an SF reduce operation which moves the point evaluations from (A) to (B).
 5. We now have a `Function` on the `input_ordering` `VertexOnlyMesh` (B) which has point evaluations from our source mesh function space at the target mesh function space node locations. These are in the correct order and have the correct domain decomposition. We can therefore safely set the `dat` of a function in our target function space to the values of this function.

For this to work for the general case, we would need to create a `VertexOnlyMesh` at the global quadrature points of the target function space, which is rather more complicated than the work I've done here.

Some important notes:
 - This does not require one mesh to be a structured refinement of the other. This, for example, should allow you to solve a PDE on two different unstructured meshes, one of which is finer than the other, and directly check the difference in solutions by interpolating from one mesh to the other within Firedrake.
 - The target mesh merely needs to share _some_ of its domain with the source mesh and we should get the results we expect (see to do note below).
 - Crucially, this is entirely parallel compatible!

Other notes:
 - The `VertexOnlyMesh` required is stored in the `Interpolator`, as are the underlying `Interpolator`s
 - For interpolation between mixed spaces, I create `sub_interpolators` for each space and evaluate them as necessary when calling `interpolate`
   - Interpolation into a mixed space therefore requires the function space being interpolated from to be another mixed space. I throw a `NotImplementedError` if not.

To do:
- [x] Test with more meshes
  - [x] ~This ought to work with meshes of different geometric dimensions as long as the target mesh is fully within the source mesh. I should check this, because this would be really cool!~ You need the geometric dimensions to match or you would need to manually specify where the lower dimensional mesh was within the higher dimensional one! This is the whole point of immersed meshes and there's no reason why an immersed mesh wouldn't work as expected methinks.
  - [x] Check that meshes with different domains work as expected.
    - [x] ~Decide whether to optionally allow interpolation into a mesh with extent beyond the target mesh - I think this is possible with how I've implemented `VertexOnlyMesh`'s `input_ordering` property, so long as I switch off giving an error when not all the points are found, and would result in a `Function` with some values that are unchanged by the interpolation operation.~ Implemented!
- [x] Work out how to ensure we only try to interpolate into point evaluation function spaces
- [x] ~Test the adjoint - since this uses pure Firedrake, I think it ought to work fine. But perhaps setting the dat values will cause an issue?~ EDIT: Testing adjoint looks to be too tricky
- [x] Implement tests for exact interpolation and transpose interpolation by using meshes which are refinements of each other
- [ ] Document in the manual

## Associated Pull Requests:
None

## Fixes Issues:
None, this is a new feature.

# Checklist for author:

# Checklist for reviewer:

- [ ] Docstrings present
- [ ] New tests present
- [ ] No bad "code smells"
- [ ] No issues in parallel
- [x] Upstream/dependent branches and PRs are ready

### DIFF
--- a/docs/source/interpolation.rst
+++ b/docs/source/interpolation.rst
@@ -108,6 +108,120 @@ in the dual space to `V` to assembled 1-forms in the dual space to
 `W`.
 
 
+Interpolation across meshes
+---------------------------
+
+The interpolation API supports interpolation between meshes where the target
+function space has finite elements (as given in the list of
+:ref:`supported elements <supported_elements>`)
+
+* **Lagrange/CG** (also known a Continuous Galerkin or P elements),
+* **Q** (i.e. Lagrange/CG on lines, quadrilaterals and hexahedra),
+* **Discontinuous Lagrange/DG** (also known as Discontinuous Galerkin or DP elements) and
+* **DQ** (i.e. Discontinuous Lagrange/DG on lines, quadrilaterals and hexahedra).
+
+Vector, tensor and mixed function spaces can also be interpolated into from
+other meshes as long as they are constructed from these spaces.
+
+.. note::
+
+   The list of supported elements above is only for *target* function spaces.
+   Function spaces on the *source* mesh can be built from most of the supported
+   elements.
+
+There are few constraints on the meshes involved: the target mesh can have a
+different cell shape, topological dimension, or resolution to the source mesh.
+There are many use cases for this: For example, two solutions to the same
+problem calculated on meshes with different resolutions or cell shapes can be
+interpolated onto one another, or onto a third, finer mesh, and be directly
+compared.
+
+
+Interpolating onto sub-domain meshes
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+The target mesh for a cross-mesh interpolation need not cover the full domain
+of the source mesh. Volume, surface and line integrals can therefore be
+calculated by interpolating onto the mesh or
+:ref:`immersed manifold <immersed_manifolds>` which defines the volume,
+surface or line of interest in the domain. The integral itself is calculated
+by calling :py:func:`~.assemble` on an approriate form over the target mesh
+function space:
+
+.. literalinclude:: ../../tests/regression/test_interpolation_manual.py
+   :language: python3
+   :dedent:
+   :lines: 11-18, 31-40
+
+For more on forms, see :ref:`this section of the manual <more_complicated_forms>`.
+
+
+Interpolating onto other meshes
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+.. note::
+
+   Interpolation *from* :ref:`immersed manifolds <immersed_manifolds>` and
+   :ref:`high-order meshes <changing_coordinate_fs>` is currently not supported.
+
+If the target mesh extends outside the source mesh domain, then cross-mesh
+interpolation will raise a :py:class:`~.DofNotDefinedError`.
+
+.. literalinclude:: ../../tests/regression/test_interpolation_manual.py
+   :language: python3
+   :dedent:
+   :lines: 46-58, 64-65
+
+This can be overriden with the optional ``allow_missing_dofs`` keyword
+argument:
+
+.. literalinclude:: ../../tests/regression/test_interpolation_manual.py
+   :language: python3
+   :dedent:
+   :lines: 72-73, 80
+
+By default the missing degrees of freedom (DoFs, the global basis function
+coefficients which could not be set) are zero:
+
+.. literalinclude:: ../../tests/regression/test_interpolation_manual.py
+   :language: python3
+   :dedent:
+   :lines: 85
+
+We can optionally specify a value to use for our missing DoFs. Here
+we set them to be ``nan`` ('not a number') for easy identification:
+
+.. literalinclude:: ../../tests/regression/test_interpolation_manual.py
+   :language: python3
+   :dedent:
+   :lines: 90-93
+
+When using :py:class:`~.Interpolator`\s, the ``allow_missing_dofs`` keyword
+argument is set at construction:
+
+.. literalinclude:: ../../tests/regression/test_interpolation_manual.py
+   :language: python3
+   :dedent:
+   :lines: 100
+
+The ``default_missing_val`` keyword argument is then set whenever we call
+:py:meth:`~.Interpolator.interpolate`:
+
+.. literalinclude:: ../../tests/regression/test_interpolation_manual.py
+   :language: python3
+   :dedent:
+   :lines: 103
+
+If we supply an output :py:class:`~.Function` and don't set
+``default_missing_val`` then any missing DoFs are left as they were prior to
+interpolation:
+
+.. literalinclude:: ../../tests/regression/test_interpolation_manual.py
+   :language: python3
+   :dedent:
+   :lines: 107-110, 113-115, 124-126
+
+
 Interpolation from external data
 --------------------------------
 
--- a/firedrake/interpolation.py
+++ b/firedrake/interpolation.py
@@ -276,11 +276,15 @@ def make_interpolator(expr, V, subset, access, bcs=None):
         # (so for vector function spaces in 2 dimensions we might need a
         # concatenation of 2 MPI.DOUBLE types when we are in real mode)
         if tensor is not None:
-            # Callable will do interpolation into tensor (which is a Dat) when
-            # it is called.
-            wrapper.mpi_type, _ = get_dat_mpi_type(tensor)
+            # Callable will do interpolation into our pre-supplied function f
+            # when it is called.
+            assert f.dat is tensor
+            wrapper.mpi_type, _ = get_dat_mpi_type(f.dat)
             assert not len(arguments)
-            callable = partial(wrapper.forward_operation, tensor)
+
+            def callable():
+                wrapper.forward_operation(f.dat)
+                return f
         else:
             assert len(arguments) == 1
             assert tensor is None
--- a/firedrake/interpolation.py
+++ b/firedrake/interpolation.py
@@ -2,6 +2,7 @@ import numpy
 from functools import partial, singledispatch
 import os
 import tempfile
+import abc
 
 import FIAT
 import ufl
@@ -21,6 +22,7 @@ import finat
 
 import firedrake
 from firedrake import tsfc_interface, utils
+from firedrake.mesh import MissingPointsBehaviour, VertexOnlyMeshMissingPointsError
 from firedrake.adjoint_utils import annotate_interpolate
 from firedrake.petsc import PETSc
 from firedrake.halo import _get_mtype as get_dat_mpi_type
@@ -28,7 +30,13 @@ from mpi4py import MPI
 
 from pyadjoint import stop_annotating
 
-__all__ = ("interpolate", "Interpolator")
+__all__ = (
+    "interpolate",
+    "Interpolator",
+    "DofNotDefinedError",
+    "CrossMeshInterpolator",
+    "SameMeshInterpolator",
+)
 
 
 # Current behaviour of interpolation in Firedrake:
@@ -68,16 +76,43 @@ __all__ = ("interpolate", "Interpolator")
 
 
 @PETSc.Log.EventDecorator()
-def interpolate(expr, V, subset=None, access=op2.WRITE, ad_block_tag=None):
+def interpolate(
+    expr,
+    V,
+    subset=None,
+    access=op2.WRITE,
+    allow_missing_dofs=False,
+    default_missing_val=None,
+    ad_block_tag=None,
+):
     """Interpolate an expression onto a new function in V.
 
     :arg expr: a UFL expression.
     :arg V: the :class:`.FunctionSpace` to interpolate into (or else
         an existing :class:`.Function`).
     :kwarg subset: An optional :class:`pyop2.types.set.Subset` to apply the
-        interpolation over.
-    :kwarg access: The access descriptor for combining updates to shared dofs.
-    :kwarg ad_block_tag: string for tagging the resulting block on the Pyadjoint tape
+        interpolation over. Cannot, at present, be used when interpolating
+        across meshes unless the target mesh is a :func:`.VertexOnlyMesh`.
+    :kwarg access: The pyop2 access descriptor for combining updates to shared
+        DoFs. Possible values include ``WRITE`` and ``INC``. Only ``WRITE`` is
+        supported at present when interpolating across meshes unless the target
+        mesh is a :func:`.VertexOnlyMesh`. See note below.
+    :kwarg allow_missing_dofs: For interpolation across meshes: allow
+        degrees of freedom (aka DoFs/nodes) in the target mesh that cannot be
+        defined on the source mesh. For example, where nodes are point
+        evaluations, points in the target mesh that are not in the source mesh.
+        When ``False`` this raises a ``ValueError`` should this occur. When
+        ``True`` the corresponding values are set to zero or to the value
+        ``default_missing_val`` if given. Ignored if interpolating within the
+        same mesh or onto a :func:`.VertexOnlyMesh` (the behaviour of a
+        :func:`.VertexOnlyMesh` in this scenario is, at present, set when
+        it is created).
+    :kwarg default_missing_val: For interpolation across meshes: the optional
+        value to assign to DoFs in the target mesh that are outside the source
+        mesh. If this is not set then zero is used. Ignored if interpolating
+        within the same mesh or onto a :func:`.VertexOnlyMesh`.
+    :kwarg ad_block_tag: An optional string for tagging the resulting block on
+        the Pyadjoint tape.
     :returns: a new :class:`.Function` in the space ``V`` (or ``V`` if
         it was a Function).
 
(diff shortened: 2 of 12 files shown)


tests/regression/test_interpolate_cross_mesh.py
```diff
@@ -0,0 +1,95 @@
+from firedrake import *
+import numpy as np
+import pytest
+
+
+def test_unit_square():
```
connorjward, line 6: I think this would be better style to separate the different parts of this test into individual tests. You can use fixtures for the common setup operations.

connorjward (changes requested):
Exciting!

tests/regression/test_interpolate_cross_mesh.py
ReubenHill: I am in the process of doing something very similar to this. When using fixtures, does it redo the the setup every time you use the fixture? That's been my hesitation thus far.

connorjward: You can set a [fixture's scope](https://docs.pytest.org/en/7.1.x/how-to/fixtures.html#scope-sharing-fixtures-across-classes-modules-packages-or-session) so it only gets created once. I guess that that is tricky when combined with parallel tests.

tests/regression/test_interpolate_cross_mesh.py
```diff
@@ -0,0 +1,181 @@
...
+                [sqrt(2) / 2, sqrt(2) / 2],
+                [-sqrt(3) / 2, 1 / 2],
+            ]
+        )  # points that ought to be on the unit circle
+        # only add target mesh vertices since they are common to both meshes
+        vertices_dest = allgather(m_dest.comm, m_dest.coordinates.dat.data_ro)
+        coords = np.concatenate((coords, vertices_dest))
+        # function.at often gets conflicting answers across boundaries for this
+        # mesh, so we lower the tolerance a bit for this test
+        m_dest.tolerance = 0.5
```
ReubenHill, line 51: Having to do this makes me wonder if the current default tolerance of `1.0` is too high.

ReubenHill: Sadly `.at` doesn't have my fancy voting algorithm to fall back on!

tests/regression/test_interpolate_cross_mesh.py
ReubenHill: I've done a bit of refactoring to use fixtures, but I think I should probably do more for clarity. I think I'll just have to accept the extra setup cost (given this won't work for parallel tests).

stephankramer:
Very exciting! Quick question out of interest: will this be adjointable? I'm guessing step 3 already is, step 5 should be trivial to adjoint - but step 4 requires reverse communication, is that also easy?

dham:
> Very exciting! Quick question out of interest: will this be adjointable? I'm guessing step 3 already is, step 5 should be trivial to adjoint - but step 4 requires reverse communication, is that also easy?

Happily PETSc SF gives you the reverse communication so this stage is adjointed (already). You are correct that this means that the mesh to mesh interpolation is adjoinable.

ReubenHill:
> Very exciting! Quick question out of interest: will this be adjointable? I'm guessing step 3 already is, step 5 should be trivial to adjoint - but step 4 requires reverse communication, is that also easy?

Once I have correctly implemented the transpose interpolation operation, this will be adjointable out of the box since interpolation is annotated. What use cases do you have in mind?

firedrake/interpolation.py
```diff
@@ -124,31 +128,278 @@ class Interpolator(object):
...
+                ] = f_src_at_dest_node_coords_m_dest_decomp.dat.data_ro[indices]
+
+        else:
+            # We would have to set up everything again if we wanted to do this
+            # properly, but we can cheat and interpolate the other way! This
+            # won't work when we enforce that transpose interpolators can only
+            # be used with cofunctions.
+            if not self.transpose_interpolator:
+                V_dest = self.expr.function_space()
+                self.transpose_interpolator = CrossMeshInterpolator(
```
ReubenHill, line 363: This is just interpolating the other way, which is NOT the same as getting the adjoint/transpose interpolator.

firedrake/interpolation.py
```diff
@@ -124,31 +128,278 @@ class Interpolator(object):
...
+            # attribute of the VOM
+            P0DG_vom = firedrake.FunctionSpace(
+                self.vom_dest_node_coords_in_m_src, "DG", 0
+            )
+            P0DG_vom_i_o = firedrake.FunctionSpace(
+                self.vom_dest_node_coords_in_m_src.input_ordering, "DG", 0
+            )
+            f_src_at_dest_node_coords_m_src_decomp = firedrake.Function(
+                P0DG_vom
+            ).interpolate(f_src)
```
ReubenHill, line 321: I should stash these interpolators

tests/regression/test_interpolate_cross_mesh.py
```diff
@@ -0,0 +1,243 @@
...
+        V_dest = FunctionSpace(m_dest, "CG", 3)
+        V_dest_2 = FunctionSpace(m_dest, "CG", 1)
+    else:
+        raise ValueError("Unknown mesh pair")
+    expr = reduce(expr_op, SpatialCoordinate(m_src))
+    expected = reduce(expr_op, coords.T)
+    return m_src, m_dest, coords, expr, expected, V_src, V_dest, V_dest_2
+
+
+def test_interpolate_cross_mesh(parameters):
```
ReubenHill, line 85: I need to add tests where we have exact refinements of meshes to check for zero error in the interpolation and transpose interpolation operations

stephankramer:
That's fantastic! For use cases: any time-dependent models with dynamic mesh adaptivity where I want an adjoint, say for sensitivity analysis. For optimisation and goal-based adaptivity it gets a bit more involved as you have to manage the sequence of meshes in the outer loop being adapted which requires re-taping (e.g. https://github.com/pyroteus/pyroteus)

firedrake/interpolation.py
```diff
@@ -113,6 +113,10 @@ class Interpolator(object):
         interpolation over.
     :kwarg freeze_expr: Set to True to prevent the expression being
         re-evaluated on each call.
+    :kwarg access: The access descriptor for combining updates to shared dofs.
+    :kwarg bcs: An optional list of boundary conditions to apply.
```
ReubenHill, line 117: Need to check if this is correct docstring

firedrake/interpolation.py
```diff
@@ -124,31 +128,406 @@ class Interpolator(object):
...
+        self.m_src = m_src
+        self.m_dest = m_dest
+
+        self.sub_interpolators = []
+
+        # Create a VOM at the nodes of V_dest in m_src. We don't include halo
+        # node coordinates because interpolation doesn't usually include halos.
+        # NOTE: it is very important to set redundant=False, otherwise the
+        # input ordering VOM will only contain the points on rank 0!
+        # QUESTION: Should any of the below have annotation turned off?
```
ReubenHill, line 256: So, _should_ the below have annotation turned off?

firedrake/interpolation.py
```diff
@@ -124,31 +128,406 @@ class Interpolator(object):
...
+        P0DG_vom_i_o = fs_type(
+            self.vom_dest_node_coords_in_m_src.input_ordering, "DG", 0
+        )
+        self.to_input_ordering_interpolator = Interpolator(
+            firedrake.TrialFunction(P0DG_vom), P0DG_vom_i_o
+        )
+        # The P0DG function outputted by the above interpolation has the
+        # correct parallel decomposition for the nodes of V_dest in m_dest so
+        # we can safely assign the dat values. This is all done in the actual
+        # interpolation method below.
```
ReubenHill, line 347: Should any of this have annotation turned off?

firedrake/interpolation.py
```diff
@@ -113,6 +113,13 @@ class Interpolator(object):
         interpolation over.
     :kwarg freeze_expr: Set to True to prevent the expression being
         re-evaluated on each call.
+    :kwarg access: The access descriptor for combining updates to shared dofs.
```
connorjward, line 116: Add something like: "Possible values are `WRITE` and `INC`"?

firedrake/interpolation.py
```diff
@@ -113,6 +113,13 @@ class Interpolator(object):
...
+    :kwarg bcs: An optional list of boundary conditions to zero-out in the
+        output function space. Interpolator rows or columns which are
+        associated with boundary condition nodes are zeroed out when this is
+        specified.
+    :kwarg allow_missing_points: For interpolation across meshes: allow
+        points in the target mesh that are not in the source mesh.
```
connorjward, line 122: What happens when this is the case? And if `False` I expect that a particular exception is raised.

firedrake/interpolation.py
```diff
@@ -124,31 +131,406 @@ class Interpolator(object):
...
+            optional value to assign to DOFs in the target mesh that are
+            outside the source mesh. If this is not set and an ``output``
+            :class:`.Function` is specified, then such DOF values are left
+            unchanged. If this is not set and no ``output`` :class:`.Function`
+            is specified, then the default value is zero. This does not effect
+            transpose interpolation.
+
         :returns: The resulting interpolated :class:`.Function`.
         """
+        return self.interpolate(*args, **kwargs)
```
connorjward, line 185: Surely this would loop infinitely? Mark as an abstract method?

firedrake/interpolation.py
```diff
@@ -124,31 +131,406 @@ class Interpolator(object):
...
             interpolate.
         :kwarg output: Optional. A :class:`.Function` to contain the output.
         :kwarg transpose: Set to true to apply the transpose (adjoint) of the
               interpolation operator.
+        :kwarg default_missing_val: For interpolation across meshes: the
+            optional value to assign to DOFs in the target mesh that are
+            outside the source mesh. If this is not set and an ``output``
+            :class:`.Function` is specified, then such DOF values are left
+            unchanged. If this is not set and no ``output`` :class:`.Function`
+            is specified, then the default value is zero. This does not effect
```
connorjward, line 180:
```suggestion
            is specified, then the default value is zero. This does not affect
```

firedrake/interpolation.py
```diff
@@ -124,31 +131,406 @@ class Interpolator(object):
...
         """
+        return self.interpolate(*args, **kwargs)
+
+
+class CrossMeshInterpolator(Interpolator):
+    """
+    Interpolate a function from one mesh and function space to another.
+
+    Only works when the function space we interpolate into has point evaluation
+    nodes. A more general version would require finding the global coordinates
```
connorjward, line 193: This is effectively a TODO comment. I don't think it belongs in the class docstring. Perhaps leave a descriptive comment where you raise the `NotImplementedError`.

firedrake/interpolation.py
connorjward: And in fact `Interpolator` should probably inherit from `abc.ABC`:

```py
class Interpolator(abc.ABC):
    ...
```

since it should never be directly instantiated.

firedrake/interpolation.py
```diff
@@ -124,31 +131,406 @@ class Interpolator(object):
...
+        self,
+        expr,
+        V,
+        subset=None,
+        freeze_expr=False,
+        access=op2.WRITE,
+        bcs=None,
+        allow_missing_points=False,
+    ):
+        if subset:
```
connorjward, line 209: It should definitely be documented that these options will not work. The only way a user would find out that `subset` etc won't work is by trying it and having it break.

firedrake/interpolation.py
```diff
@@ -124,31 +131,406 @@ class Interpolator(object):
...
+
+        self.arguments = extract_arguments(expr)
+        self.nargs = len(self.arguments)
+        if self.nargs and self.arguments[0] != self.expr:
+            raise NotImplementedError(
+                "Can't yet create an interpolator from an expression with arguments."
+            )
+
+        if self._allow_missing_points:
+            missing_points_behaviour = None
```
connorjward, line 236: I think this would be better as an `enum`:

```py
class MissingPointsBehaviour(enum.Enum):
    NONE = enum.auto()  # not a great name
    ERROR = enum.auto()
```

firedrake/interpolation.py
```diff
@@ -124,31 +131,406 @@ class Interpolator(object):
...
+        V_dest = V
+        m_src = extract_unique_domain(expr)
+        m_dest = V_dest.ufl_domain()
+        m_src_gdim = m_src.geometric_dimension()
+        m_dest_gdim = m_dest.geometric_dimension()
+        if m_src_gdim != m_dest_gdim:
+            raise ValueError(
+                "geometric dimensions of source and destination meshes must match"
+            )
+        self.m_src = m_src
```
connorjward, line 250: Very picky but I don't like this variable name. `src_mesh` for example would make it much clearer what this is.

connorjward (dismissed):

firedrake/interpolation.py
ReubenHill: Thanks, will change

firedrake/interpolation.py
ReubenHill: Yes, you are correct.

firedrake/interpolation.py
ReubenHill: This seems like a pretty important thing to highlight to a user though. How about
```
    """
    Interpolate a function from one mesh and function space to another.

    Notes
    -----
    
    Only works when the function space we interpolate into has point evaluation
    nodes. A more general version would require finding the global coordinates
    of all quadrature points of the target function space in the source mesh.

    For arguments, see :class:`.Interpolator`.
    """
```

ReubenHill: Thinking more about it, I agree with you.

firedrake/interpolation.py
ReubenHill: Do `MAX` and `MIN` not work?

firedrake/interpolation.py
ReubenHill: At the moment you get something like
```
ValueError: 117 vertices are outside the mesh and have been removed from the VertexOnlyMesh
```
Is this sufficient or do I need to try and catch it then rethrow with something more specific?

firedrake/interpolation.py
ReubenHill: Oh I've found this note in `interpolate`'s docstring
```
    .. note::

       If you use an access descriptor other than ``WRITE``, the
       behaviour of interpolation is changes if interpolating into a
       function space, or an existing function. If the former, then
       the newly allocated function will be initialised with
       appropriate values (e.g. for MIN access, it will be initialised
       with MAX_FLOAT). On the other hand, if you provide a function,
       then it is assumed that its values should take part in the
       reduction (hence using MIN will compute the MIN between the
       existing values and any new values).
```

firedrake/interpolation.py
ReubenHill: Changed

firedrake/interpolation.py
ReubenHill: This is me specifying the necessary argument to `VertexOnlyMesh` (either `None` or `'error'`) which is part of the public API. I don't see how an `enum` is therefore helpful?

firedrake/interpolation.py
ReubenHill: I've decided to rethrow something more specific.

docs/source/interpolation.rst
```diff
@@ -108,6 +108,22 @@ in the dual space to `V` to assembled 1-forms in the dual space to
...
+use cases for this: For example, two solutions to the same problem calculated
+on meshes with different resolutions or cell shapes can be interpolated onto
+one another, or onto a third mesh, and be directly compared.
+
+Volume, surface and line integrals can be calculated by interpolating onto the
+mesh or immersed manifold which defines the volume, surface or line of
+interest in the domain, then assembling an appropriate form over the target
+mesh function space.
+
+
```
ReubenHill, line 126: @dham What would be a good example to demonstrate here?

ReubenHill: The line integral from the tests perhaps?

firedrake/interpolation.py
connorjward: My point was that the docstring was unclear. You should just include something like "If `False` then an exception will be raised if points are located outside the source mesh". A more specific exception sounds good though.

ReubenHill: See the updated docstring

firedrake/interpolation.py
connorjward: I think generally if you have a finite number of options for something an enum is recommended. You can actually get away with this without changing the public API:

```py
class MissingPointsBehaviour(enum.Enum):
    NONE = None
    ERROR = "error"
```

then things like `missing_points_behaviour == MissingPointsBehaviour.ERROR` will work even if the user provides the string "error".

connorjward: Yes. The idea is to get rid of explicit strings in the internal code. You can also do some [straightforward validation of user input](https://stackoverflow.com/questions/43634618/how-do-i-test-if-int-value-exists-in-python-enum-without-using-try-catch).

I also think that `None`/`NONE` is not a very descriptive name. Should it be `"ignore"`/`IGNORE`?

ReubenHill: > I also think that `None`/`NONE` is not a very descriptive name. Should it be `"ignore"`/`IGNORE`?

Well it implies "No missing points behaviour", but if you don't think it's clear then that's fair enough. It's documented at least! But again I don't want to be changing the public API (and all the tests and manual pages) at this point.

How do enums interact with a public API in general? Is the user supposed to create an enum and supply that to the constructor? Does the displayed default value change to an instance of the enum? There are very few of these in the firedrake code base, none of which seem to be public facing, hence all my questions!

ReubenHill: Okay I've added the enum in dc4f2812245b2cb745d46b3a19c96b539cfc46ed

firedrake/mesh.py
```diff
@@ -2841,12 +2847,17 @@ def VertexOnlyMesh(mesh, vertexcoords, missing_points_behaviour='error',
         mesh, vertexcoords, tolerance=tolerance, redundant=redundant, exclude_halos=False
     )
 
-    if missing_points_behaviour:
+    if missing_points_behaviour is None:
+        missing_points_behaviour = MissingPointsBehaviour.IGNORE
+    else:
+        missing_points_behaviour = MissingPointsBehaviour(missing_points_behaviour)
```
dham, line 2853:
```suggestion
    missing_points_behaviour = MissingPointsBehaviour(missing_points_behaviour)
```

dham (dismissed):

JDBetteridge (approved):
Previously approved

firedrake/interpolation.py
JDBetteridge: https://github.com/firedrakeproject/firedrake/pull/3087

PR #3076, opened 2023-08-24
Make subclasses of Constant use the same numbering

connorjward:
```py
from firedrake import *

mesh = UnitIntervalMesh(1.0)


class CustomConstant(Constant):
    pass


e = CustomConstant(2.0) / Constant(1.0)
v = assemble(e * dx(mesh))
print(f"{v=}")
```

Fixes an obscure error where running this program a second time (with a populated cache) would produce `0.5` instead of `2`.

I don't know precisely where this error was occurring but the issue will have been coming from the fact that previously both the `CustomConstant` and the `Constant` will have had the same number. This fixes that.

Thanks @jrmaddison for finding this.

### DIFF
--- a/firedrake/constant.py
+++ b/firedrake/constant.py
@@ -94,7 +94,7 @@ class Constant(ufl.constantvalue.ConstantValue, ConstantMixin, TSFCConstantMixin
         self.name = name or 'constant_%d' % self.uid
 
         super().__init__()
-        Counted.__init__(self, count)
+        Counted.__init__(self, count, Counted)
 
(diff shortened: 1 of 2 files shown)


ReubenHill:
@connorjward you should add the test which triggered the issue

connorjward:
> @connorjward you should add the test which triggered the issue

The error doesn't happen every single time that script is run so I don't want to include it as a test. I've added a simpler numbering test though.

dham (approved):

PR #3083, opened 2023-08-30
Make pytorch optional when building docs

dham:
Make PyTorch optional when building the documentation.  If torch is not successfully imported when building the docs, we mock up a class that enables the `firedrake.ml.pytorch` module to be imported and causes the intersphinx link to PyTorch to resolve properly.

I am fully aware of how 😈 this is.

This should fix the recent issues with the website builder action, as well as satisfy user demand to build docs without installing PyTorch.

### DIFF
--- a/docs/Makefile
+++ b/docs/Makefile
@@ -3,7 +3,7 @@
 
 # You can set these variables from the command line.
 SPHINXOPTS    =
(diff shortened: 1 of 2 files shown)


JDBetteridge (approved):
It might say "approve" but I do not condone such hackery!

PR #3099, opened 2023-09-12
Test UFL upstream.

dham:


nbouziani:
The two failing tests are caused by this change (https://github.com/dolfin-adjoint/pyadjoint/commit/dd7d9bdb1e1063db6fc5e0bde001dfe9185b17c0) which translates into turning the adjoint input of `ReducedFunctional` as a `Coefficient` instead of `Vector`, as a result of scaling a `Vector`, which I don't think should happen. I will have a look.

nbouziani:
> The two failing tests are caused by this change ([dolfin-adjoint/pyadjoint@dd7d9bd](https://github.com/dolfin-adjoint/pyadjoint/commit/dd7d9bdb1e1063db6fc5e0bde001dfe9185b17c0)) which translates into turning the adjoint input of `ReducedFunctional` as a `Coefficient` instead of `Vector`, as a result of scaling a `Vector`, which I don't think should happen. I will have a look.

The failure is not related to UFL, so I think we can proceed with merging UFL upstream.

connorjward:
Closing as we have now updated UFL.

PR #3113, opened 2023-09-20
ufl#208 has removed ufl.algorithms.multifunction. Use ufl.corealg.multifunction instead.

francesco-ballarin:


francesco-ballarin:
FYI, I guess that firedrake ufl fork doesn't include upstream#208 yet, but this should not be a blocker for approving the workflow runs. ufl.algorithms.multifunction had always actually been importing ufl.corealg.multifunction, so this change should be backward compatible.

nbouziani:
Yes, given that UFL upstream was merged earlier today, this is needed to not break CI. I have noticed that problem in another PR (#2294). I think this is ready to go.

nbouziani (approved):

francesco-ballarin:
CI failure looks unrelated. Is the `torch` coupling getting fixed in another PR?

nbouziani:
> CI failure looks unrelated. Is the `torch` coupling getting fixed in another PR?

Yes, those failures are expected and are fixed in another PR.

PR #3114, opened 2023-09-20
Inject everything on level 1

pbrubeck:
---
name: "MG: Inject coefficients"
description: ''
title: ''
labels: ''
assignees: ''

---

# Description
Things were breaking when a solver is reused for a bilinear form with a coefficient that is not a `Constant` , e.g. functions in R or DG0, after updating that coefficient.
Turns out that we also need to manually inject everything in `problem.F.coefficients()`.
But wait, there's also another issue. The manual injection has to ensure that the level above the current one is up-to-date with respect to the finest level, an assumption that only holds in the two-level case. PCMG will traverse the levels in the order [1, 2, 3, ..., finest,  0 = coarsest]. So we should inject all coefficients across the hierarchy the first time we are reassembling the problem.

## Fixes Issues:
- TODO report the issue

# Checklist for author:

- [ ] I have linted the codebase (`make lint` in the `firedrake` source directory).

# Checklist for reviewer:

- [ ] Docstrings present
- [ ] New tests present
- [ ] No bad "code smells"
- [ ] No issues in parallel

### DIFF
--- a/firedrake/solving_utils.py
+++ b/firedrake/solving_utils.py
@@ -474,6 +474,7 @@ class _SNESContext(object):
         :arg P: the preconditioner matrix (a Mat)
         """
         from firedrake.bcs import DirichletBC
+        from firedrake.mg.utils import get_level
         dm = ksp.getDM()
         ctx = dmhooks.get_appctx(dm)
         problem = ctx._problem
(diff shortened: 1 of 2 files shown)


firedrake/solving_utils.py
```diff
@@ -484,14 +485,44 @@ def compute_operators(ksp, J, P):
...
+            while fine._coarse is not None:
+                coarse = fine._coarse
+
+                fine_coefficients = get_coefficients(fine._problem)
+                coarse_coefficients = get_coefficients(coarse._problem)
+                seen = set()
+                for xf, xc in zip(fine_coefficients, coarse_coefficients):
+                    if xf not in seen:
+                        manager.inject(xf, xc)
+                        seen.add(xf)
```
wence-, line 519: Is it sufficient to only do this in `kspcomputeoperators` case, or should it also be done in `form_jacobian` and `form_function`? When we did this before (cc @pefarrell) we did it by setting up hooks https://github.com/firedrakeproject/firedrake/compare/master...pefarrell/transfer-manager-inject-everything

That was never merged because it does too much work (reinjects too frequently). I think this suffers from the same problem. There's no need to reinject a coefficient of `F` if it hasn't changed. These days you can ask the `dat` for its state, keep track of that and if it changed since the last time you saw it, you know its changed and you must inject the function.

PR #3117, opened 2023-09-22
Separate plotting

### LINKED ISSUE #2847: Move plotting to submodule
And remove mandatory matplotlib dependency!

📈

JDBetteridge:
Fixes #2847


danshapero:
Will this patch also make VTK an optional dependency as well?

connorjward:
Do we need to put something into `firedrake-install` too? A typical user will want plotting functionality by default. I imagine we will need to rewrite some of the website too.

JDBetteridge:
@danshapero sadly VTK will remain a hard dependency, since we use it for correctly formatting output. It would be possible to install without VTK if we did something similar with `output.py` and `paraview_reordering.py`. We could have an `output` subdirectory that also included the interface to checkpointing. I think this needs to be a separate PR though. Feel free to create a new issue :slightly_smiling_face:

JDBetteridge:
@connorjward Yes! That is my intention.

However, I want to make sure CI is passing without `matplotlib` before I add it back in.

firedrake/pyplot/__init__.py
connorjward, line 1: Please explain again why this can't be called `firedrake.plot`? If there is a weird naming issue I think that should be fixed rather than renaming the subpackage.

requirements-ext.txt
```diff
@@ -1,7 +1,6 @@
 cached_property
 cachetools
 ipython
-matplotlib
```
connorjward, line 4: This needs to be added as an option to `firedrake-install` somehow, `--no-matplotlib`?

.github/workflows/build.yml
```diff
@@ -86,6 +86,7 @@ jobs:
           python "$(which firedrake-clean)"
           python -m pip install \
             pytest-cov pytest-timeout pytest-xdist pytest-timeout
+          python -m pip uninstall -y matplotlib
```
connorjward, line 89: Why is this here?

JDBetteridge: This is temporary to check tests pass without `matplotlib`. By default `matplotlib` is installed as a dependency of FEMlium I think

requirements-ext.txt
JDBetteridge: Sure

firedrake/pyplot/__init__.py
JDBetteridge: Off the top of my head I can't remember. The name of the import can be whatever, but if `pyplot` is controversial, I'd like a definitive answer about what it should be before I change it again.

JDBetteridge: Potentially it's a bug related to documentation as tests are passing locally with a different name. I can't test the docs until the upstream fail is fixed.

connorjward:
Following the [meeting yesterday](https://github.com/firedrakeproject/firedrake/wiki/Firedrake-meeting-2023-11-22) I have written a [policy document](https://github.com/firedrakeproject/firedrake/wiki/Isolating-optional-dependencies) explaining that we will try and use this approach for optional dependencies more, possibly applying to VTK at some point.

firedrake/pyplot/__init__.py
JDBetteridge: Resolved in meeting

requirements-ext.txt
JDBetteridge: Did you change your mind on this point?

.github/workflows/build.yml
JDBetteridge: This is now gone

requirements-ext.txt
connorjward: I think it's possibly quite reasonable to not install it by default and prompt the users to do so when they hit the import error. Not sure though. It would keep the install script a little simpler.

tests/demos/test_demos_run.py
```diff
@@ -57,4 +57,16 @@ def py_file(rst_file, tmpdir, monkeypatch):
...
+        try:
+            from slepc4py import SLEPc  # noqa: F401
+        except ImportError:
+            pytest.skip(msg="SLEPc unavailable, skipping qgbasinmodes.py")
+
+    if basename(py_file) in ("DG_advection.py", "qgbasinmodes.py"):
+        try:
+            import matplotlib  # noqa: F401
+        except ImportError:
+            pytest.skip(msg="Matplotlib unavailable, skipping DG_advection.py")
```
connorjward, line 70: Wrong error message.

```suggestion
            pytest.skip(msg=f"Matplotlib unavailable, skipping {basename(py_file)}")
```

connorjward (commented):
Ran out of time on review. Will revisit.

requirements-ext.txt
```diff
@@ -1,7 +1,7 @@
 cached_property
 cachetools
+ipympl
```
connorjward, line 3: Should this really be installed by default? Seems like it might break some of this separation that we are trying to introduce.

requirements-ext.txt
connorjward: I'm happy with this line getting deleted so long as there is a meaningful message when users hit an MPL `ImportError` prompting them to `pip install`.

tests/demos/test_notebooks_run.py
```diff
@@ -12,11 +12,11 @@
 @pytest.fixture(params=glob.glob(os.path.join(nb_dir, "*.ipynb")),
                 ids=lambda x: os.path.basename(x))
 def ipynb_file(request):
-    if request.param.endswith("08-composable-solvers.ipynb"):
-        pytest.xfail("FIX THIS LATER: Have to update our PETSc fork to fix build issues")
+    # Notebook 08-composable-solvers.ipynb still has an issue, the cell is commented out
```
connorjward, line 15: This comment isn't terribly clear. Reference an issue or provide more explanation?

tests/output/test_plotting.py
```diff
@@ -1,13 +1,19 @@
...
+from firedrake import *
+
+try:
+    from firedrake.pyplot import *
+    import matplotlib.pyplot as plt
+    import matplotlib.colors
+    from matplotlib.animation import FuncAnimation
+    from mpl_toolkits.mplot3d import Axes3D
+except ImportError:
+    # Matplotlib is not installed
```
connorjward, line 12: Arguably easier to have `pytest.skip("Matplotlib is not installed", allow_module_level=True)` (https://docs.pytest.org/en/6.2.x/skipping.html#skipping-test-functions) here instead of adding all the individual skip markers.

connorjward (changes requested):
Mostly good.

requirements-ext.txt
JDBetteridge: Good spot, this is required for tests, but shouldn't be there by default!

tests/demos/test_notebooks_run.py
JDBetteridge: In my defence, nor was the original `xfail` message!

JDBetteridge: https://github.com/firedrakeproject/firedrake/issues/3308

requirements-ext.txt
JDBetteridge: I am going to argue that the import error that a user gets when trying to use `matplotlib` is sufficient here.

connorjward: I'm not sure about that. I think users may see it as a bug if they try `import firedrake.pyplot` and it fails. If we have an explicit message about requiring a pip install then they will see that it's deliberate.

connorjward (approved):
I think this is OK.

PR #3132, opened 2023-09-28
Ksagiyam/merge upstream debug

ksagiyam:
just testing.


ksagiyam:
@wence- After this PETSc MR https://gitlab.com/petsc/petsc/-/merge_requests/6641, this trick https://gitlab.com/petsc/petsc/-/blob/main/src/ksp/pc/impls/patch/pcpatch.c?ref_type=heads#L2697 seems no longer working: `FAILED tests/demos/test_notebooks_run.py::test_notebook_runs[08-composable-solvers.ipynb]` (see https://github.com/firedrakeproject/firedrake/actions/runs/6300658192/job/17103981872).

`PetscLayoutSetUp` sets up the layout with the `map->n` and stores the `map->n` in `map->oldn` when it is first called. When it is called the second time on the same layout object, it checks that the currently set `map->n` equals the stored `map->oldn`, and raises an exception if not https://gitlab.com/petsc/petsc/-/blob/main/src/vec/is/utils/pmap.c?ref_type=heads#L234.
This above MR https://gitlab.com/petsc/petsc/-/merge_requests/6641 causes `PetscLayoutSetUp` to be called the second time upon `VecDuplicate` as a result of reusing `PetscLayout` of the old `Vec` in the new `Vec`.

Here, I confirmed that this naive addition https://github.com/firedrakeproject/petsc/commit/fcb0106b8cbf5d37f81930dd95197ab7003d5056 at least makes the test pass, but do you have any thoughts on how to fix this correctly?

connorjward:
@ksagiyam what's the status of this? We've updated PETSc fairly recently without any of these changes seemingly.

ksagiyam:
The patch I have here is just a workaround, and I do not know the best solution to this. Also, if we fix this upstream, we probably need to add test.

ksagiyam:
Yes. That test has been xfailed.
You might be right. We still need to add test probably though. Maybe it is not so hard.

wence-:
That pcpatch change looks vaguely plausible, though there some other places, e.g. https://gitlab.com/petsc/petsc/-/blob/main/src/ksp/pc/impls/patch/pcpatch.c?ref_type=heads#L2740 and also snespatch https://gitlab.com/petsc/petsc/-/blob/main/src/snes/impls/patch/snespatch.c#L127

As the comment indicates, it's a horrible hack, but I didn't want to make one work vector per patch.

PR #3135, opened 2023-10-02
Fix assemble with multiple kernels and subdomains

### LINKED ISSUE #3125: BUG: pyop2 kernel arguments mismatch
This is from a Thetis demo that's failing in the CI (https://github.com/thetisproject/thetis/actions/runs/6217550038/job/16872773644) with "ValueError: You are trying to pass in a different number of arguments than the kernel is expecting". I managed to break it down to the following MFE:
```
from firedrake import *
mesh = UnitSquareMesh(10, 10)
# if you want a mesh that actually has a subdomain id 2 (from Thetis):
# mesh = Mesh("examples/tidalfarm/headland.msh") 
W = FunctionSpace(mesh, "CG", 1)
Z = MixedFunctionSpace([V, W])
z = Function(Z)
u, p = split(z)
tst = TestFunction(Z)
v, q = split(tst)
d = Function(W)
F = inner(z, tst)*dx + inner(v, u)/(d+p)*dx(2, degree=10)
print(F)
solve(F==0, z)
```
Note that this MFE gives a slightly different error 
```
File "/home/skramer/fd2509/src/PyOP2/pyop2/global_kernel.py", line 332, in __call__
    func(*args)
TypeError: this function takes at least 8 arguments (7 given)
```
but I suspect it's the same issue.

connorjward:
Fixes #3125

I am not thrilled by the test I have provided. I don't know what sort of testing approach would be suitable here. It seems bizarre to write a test called `test_solving_complicated_form_does_not_raise_obscure_error`.

Aside: I really hate that `zip()` doesn't raise an error when its arguments are different lengths!


stephankramer:
Thanks, Connor. Yeah I've no idea either; Seems to only occur on the very specific combination I managed to reduce this down to - so just having a regression test for that very specific bug seems sensible to me.

wence-:
You could put some values in the functions and check that you get the answer you expect?

firedrake/assemble.py
```diff
@@ -738,14 +738,25 @@ def global_kernels(self):
...
+                        local_kernel,
+                        global_kernel,
+                        self._tensor,
+                        subdomain_id,
+                        self.all_integer_subdomain_ids,
+                        diagonal=self.diagonal,
+                        lgmaps=self.collect_lgmaps(local_kernel, self._bcs)
+                    ).build()
+                )
+        assert_empty(global_kernel_iter)
```
wence-, line 758: So the invariant is that `sum(len(kinfo.subdomain) for kinfo in local_kernels) == len(global_kernels)` I think.

I forget how the global and local kernels relate, can you remind me? IIUC, a `local_kernel` corresponds to an integrand in a form, and integrands are now grouped according to the metadata associated with their measure. So:

```
expr1*dx + expr2*dx(1, 2) + expr3*dx(1, 2)
```
gets grouped into:
```
expr1*dx(otherwise) + (expr1 + expr2 + expr3)*dx((1, 2))
```

and then this code runs the kernel for `expr1 + expr2 + expr3` separately over `dx(1)` and `dx(2)`.

Is there a reason you don't ask firedrake to make global kernels (and hence parloops) for the domain marked by `(1, 2)`?  Or maybe I have misunderstood.

Alternately, would it be a "safer" thing to do to unwind the `local_kernels` property, that is:

```
def local_kernels(self):
    kernels = existing_code
    return tuple(itertools.chain(*(zip(k.kinfo.subdomain_id, itertools.repeat(k)) for k in kernels)))
```

Then `global_kernels` would be:
```
def global_kernels(self):
    return tuple(make_global_kernel(..., subdomain, k).build() for subdomain, k in self.local_kernels))
```

And this would be:

```
def parloops(self):
    return tuple(ParLoopBuilder(..., local, global, subdomain) for (subdomain, local), global in zip(self.local, self.global)))
```
WDYT?

connorjward: > Is there a reason you don't ask firedrake to make global kernels (and hence parloops) for the domain marked by (1, 2)? Or maybe I have misunderstood.

We [discussed something similar before](https://github.com/firedrakeproject/firedrake/pull/3071#discussion_r1301708127). This approach would not work easily for overlapping subdomains. I suppose it's a sensible optimisation to do this for non-overlapping subdomains like in the case you describe.

> Alternately, would it be a "safer" thing to do to unwind the local_kernels property, that is:

Yes that's a good idea thanks.

dham (approved):

firedrake/assemble.py
wence-: > This approach would not work easily for overlapping subdomains.

Ah yes, I remember. It feels like this might be something that could be solved in UFL. e.g. to copy over that example from your link:
```
a = inner(u, v)*dx((1, 2, 3)) + inner(grad(u), grad(v))*dx(2) + inner(u, v)*dx
```
gets grouped as:
```
inner(u,v)*dx((1,3,"everywhere")) +  (inner(u,v) + inner(grad(u), grad(v)))*dx((2,))
```

Previously `"everywhere"` was not allowed to appear in a post-processed integral, so I would have expected

```
(inner(u, v) + inner(u, v))*dx((1, 3)) + (inner(u, v) + inner(grad(u), grad(v)))*dx((2,)) + inner(u, v)*dx("otherwise")
```

Notice how the contribution of the "everywhere" integral is "broadcast" onto both subdomain lists.

wence-: Yeah, what I thought was happening still seems to be:
```
In [12]: print(F)
{ w_0 * (conj((v_0))) } * dx(<Mesh #1>[1], {})
  +  { conj(((grad(v_0)) : (grad(w_0)))) } * dx(<Mesh #1>[2], {})
  +  { w_0 * (conj((v_0))) } * dx(<Mesh #1>[3], {})
  +  { w_0 * (conj((v_0))) } * dx(<Mesh #1>[everywhere],
```

```
In [16]: fd = compute_form_data(F)

In [17]: print(fd.preprocessed_form)
{ v_0 * w_0 } * dx(<Mesh #1>[('otherwise',)], {'estimated_polynomial_degree': 4})
  +  { v_0 * w_0 + v_0 * w_0 } * dx(<Mesh #1>[(1, 3)], {'estimated_polynomial_degree': 4})
  +  { v_0 * w_0 + sum_{i_9} (grad(v_0))[i_9] * (grad(w_0))[i_9]  } * dx(<Mesh #1>[(2,)], {'estimated_polynomial_degree': 4})
```

```
In [27]: print(fd.integral_data[0].integrals[0])
{ v_0 * w_0 } * dx(<Mesh #1>[('otherwise',)], {'estimated_polynomial_degree': 4})

In [28]: print(fd.integral_data[1].integrals[0])
{ v_0 * w_0 + v_0 * w_0 } * dx(<Mesh #1>[(1, 3)], {'estimated_polynomial_degree': 4})

In [29]: print(fd.integral_data[2].integrals[0])
{ v_0 * w_0 + sum_{i_9} (grad(v_0))[i_9] * (grad(w_0))[i_9]  } * dx(<Mesh #1>[(2,)], {'estimated_polynomial_degree': 4})

In [30]: len(fd.integral_data)
Out[30]: 3
```

dham (approved):

PR #3141, opened 2023-10-06
Revert to preferring homebrew over pyenv for MacOS install

connorjward:
Homebrew installed pyenvs enforce particular CFLAGS and LDFLAGS that break libraries with compiled extensions like petsc4py and mpi4py. The original homebrew Python bug has also been fixed.

### DIFF
--- a/docs/source/download.rst
+++ b/docs/source/download.rst
@@ -110,10 +110,9 @@ have yet to create binary wheels for 3.11.x, but we have generated these
 for the major supported platforms.
 The installation script is tested on Ubuntu and MacOS X. On Ubuntu 22.04
 or later, the system installed Python 3 is supported and tested. On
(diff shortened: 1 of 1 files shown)


Ig-dolci:
Connor, are you saying that the original homebrew was fixed based on the test that I executed yesterday? I mean, by executing python3 -c impor venv...

connorjward:
> Connor, are you saying that the original homebrew was fixed based on the test that I executed yesterday? I mean, by executing python3 -c impor venv...

Yes. That was the bug that made me start suggesting that pyenv be used instead of homebrew. Also we have found serious issues with pyenv (where it links to homebrew installed OpenMPI, BLAS etc without giving the user a way to not have that happen).

connorjward:
Converting to draft as more investigation is required.

dham (approved):

PR #712, opened 2016-02-22
Zenodo

dham:
Zenodo integration for Firedrake:

* firedrake-zenodo script which allows users and developers to easily create releases.

* Migration of bitbucket repositories to github.

* Documentation on the website and in the script.

Issue #711, opened 2016-02-19
Interpolate arbitrary UFL expressions


miklos1:
Given the modularity of TSFC, it is not very hard to make `interpolate` accept arbitrary UFL expressions for interpolating them onto `Function`s. This would come with several advantages:
* H(div) / H(curl) functions could be interpolated onto P/Q functions.
* Safer replacement for `Expression`s. For example, `f.interpolate(Expression("2*(x[0]+1.0"))` yields fairly unhelpful error message.
* Directly fixes #485 and #526.
* `Expression` could be deprecated as it confuses users. Occasionally users want to use `Expression`s inside forms which is unnecessary since UFL can express pretty much anything that `Expression`s can. This could help close #53 and #514.

There is very quick and dirty sketch in the branch `interpolate-ufl-expr`.

miklos1:
I wonder if this feature could deprecate `assemble_expressions.py` as well. It seems to me that assembling expressions is a restricted version of this, where all of Functions in the expressions as well as the Function we assemble onto have the same function space.

OK, so we have a restriction here that the target function space must have reference value mapping identity. Not sure if assembling expressions have this restriction.

wence-:
When assembling expressions, basically everything is just done pointwise.  So the mapping is meaningless.

dham:
I don't think these are the same thing at all. The primary use of assemble_expressions is to implement timestepping, so it's basically compiled operations on vectors.

miklos1:
Update here: #715 is now merged. However, we cannot deprecate `Expression`s yet, as they are the only thing accepted by `MeshGeometry.init_cell_orientations`. I have to note that `MeshGeometry.init_cell_orientations` itself should be retired (when one gets the time to implement or wire in the better alternative), as it relies on a user-provided normal vector and the mesh geometry to calculate cell orientations, even though one could get consistent cell orientations by looking only at the mesh topology. (I think `DMPlexOrient` implements exactly this.)

miklos1:
@colinjcotter in #485:
> This would also be useful for writing higher order fields to vtu, since interpolate is more suitable than project, usually.

Now we have the technology to use `interpolate` instead of `project` when outputting to `.vtu`. Shall we make that change?

wence-:
Done (as an option in #737).  Should this be closed?

miklos1:
Yes. We might open separate pull requests for interpolation onto more complicated finite elements etc.

PR #3152, opened 2023-10-11
Public interface to gc + Remove __del__ method and add weakref.finalizer

### LINKED ISSUE #2669: MPI communicator exhaustion when creating too many meshes
The following example

```
from firedrake import *

import gc
import sys

i = 0

while True:
    i += 1

    mesh = UnitIntervalMesh(10)
    del mesh
    gc.collect()

    sys.stdout.write(f"{i:d}\n")
    sys.stdout.flush()
```

eventually leads to an error (here when run on 2 processes)

```
[...]
680
680
  File "[...]/firedrake/test.py", line 11, in <module>
  File "[...]/firedrake/test.py", line 11, in <module>
    mesh = UnitIntervalMesh(10)
  File "PETSc/Log.pyx", line 115, in petsc4py.PETSc.Log.EventDecorator.decorator.wrapped_func
    mesh = UnitIntervalMesh(10)
  File "PETSc/Log.pyx", line 115, in petsc4py.PETSc.Log.EventDecorator.decorator.wrapped_func
  File "PETSc/Log.pyx", line 116, in petsc4py.PETSc.Log.EventDecorator.decorator.wrapped_func
  File "[...]/firedrake/firedrake/src/firedrake/firedrake/utility_meshes.py", line 170, in UnitIntervalMesh
    return IntervalMesh(
  File "PETSc/Log.pyx", line 115, in petsc4py.PETSc.Log.EventDecorator.decorator.wrapped_func
  File "PETSc/Log.pyx", line 116, in petsc4py.PETSc.Log.EventDecorator.decorator.wrapped_func
  File "PETSc/Log.pyx", line 116, in petsc4py.PETSc.Log.EventDecorator.decorator.wrapped_func
  File "[...]/firedrake/firedrake/src/firedrake/firedrake/utility_meshes.py", line 170, in UnitIntervalMesh
  File "[...]/firedrake/firedrake/src/firedrake/firedrake/utility_meshes.py", line 116, in IntervalMesh
    return IntervalMesh(
  File "PETSc/Log.pyx", line 115, in petsc4py.PETSc.Log.EventDecorator.decorator.wrapped_func
    plex = mesh.plex_from_cell_list(
  File "[...]/firedrake/firedrake/src/firedrake/firedrake/mesh.py", line 391, in plex_from_cell_list
  File "PETSc/Log.pyx", line 116, in petsc4py.PETSc.Log.EventDecorator.decorator.wrapped_func
    plex = PETSc.DMPlex().createFromCellList(dim, cells, coords, comm=icomm)
  File "[...]/firedrake/firedrake/src/firedrake/firedrake/utility_meshes.py", line 116, in IntervalMesh
  File "PETSc/DMPlex.pyx", line 49, in petsc4py.PETSc.DMPlex.createFromCellList
    plex = mesh.plex_from_cell_list(
  File "[...]/firedrake/firedrake/src/firedrake/firedrake/mesh.py", line 399, in plex_from_cell_list
petsc4py.PETSc.Error: error code 98
[0] DMPlexCreateFromCellListPetsc() at [...]/firedrake/firedrake/src/petsc/src/dm/impls/plex/plexcreate.c:4829
[0] DMSetType() at [...]/firedrake/firedrake/src/petsc/src/dm/interface/dm.c:3896
[0] DMCreate_Plex() at [...]/firedrake/firedrake/src/petsc/src/dm/impls/plex/plexcreate.c:4325
[0] PetscPartitionerCreate() at [...]/firedrake/firedrake/src/petsc/src/dm/partitioner/interface/partitioner.c:383
[0] PetscPartitionerSetType() at [...]/firedrake/firedrake/src/petsc/src/dm/partitioner/interface/partitioner.c:44
[0] PetscPartitionerCreate_PTScotch() at [...]/firedrake/firedrake/src/petsc/src/dm/partitioner/impls/ptscotch/partptscotch.c:327
[0] General MPI error
[0] MPI error 672744975 Other MPI error, error stack:
                internal_Comm_dup(85)...............: MPI_Comm_dup(comm=0xc40053ec, newcomm=0x55bf69238950) failed
                MPIR_Comm_dup_with_info_impl(724)...: 
                MPII_Comm_copy(833).................: 
                MPIR_Get_contextid_sparse_group(591): Too many communicators (0/2048 free on this process; ignore_id=0)
    plex = PETSc.DMPlex().createFromCellList(dim,
  File "PETSc/DMPlex.pyx", line 49, in petsc4py.PETSc.DMPlex.createFromCellList
application called MPI_Abort(PYOP2_COMM_WORLD, 1) - process 0
```

### LINKED ISSUE #3075: Garbage collection improvements
A couple of places have been identified where our garbage collection could be improved:

- We should call `gc.collect` before we call `PETSc.garbage_cleanup`. Otherwise we get apparent memory leaks since we only clean up objects once Python decides to break the reference cycles.
- We should add our own `garbage_cleanup` routine for users to call. This should take a user-level comm object and extract the internal PyOP2 comm where the objects are actually being stored for deletion.

JDBetteridge:
Hopefully fixes #2669 and fixes #3075 

~Ideally this depends on https://github.com/firedrakeproject/firedrake/pull/3229 being merged into this branch and then depends on https://github.com/OP2/PyOP2/pull/711~

Needs merging in this order: ~[3229](https://github.com/firedrakeproject/firedrake/pull/3229)~, [PyOP2 #712](https://github.com/OP2/PyOP2/pull/712), ~[PyOP2 #711](https://github.com/OP2/PyOP2/pull/711)~(merged into 712), [#3152](https://github.com/firedrakeproject/firedrake/pull/3152). This sequence of PRs changes the `internal_comm` function signature.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -42,6 +42,7 @@ jobs:
     env:
       # PETSC_DIR and MPICH_DIR are set inside the docker image
       FIREDRAKE_CI_TESTS: 1
+      PYOP2_CI_TESTS: 1
       PETSC_ARCH: ${{ matrix.petsc_arch }}
       OMP_NUM_THREADS: 1
       OPENBLAS_NUM_THREADS: 1
@@ -79,6 +80,7 @@ jobs:
             --install icepack \
             --install irksome \
             --install femlium \
+            --package-branch pyop2 JDBetteridge/check_something \
             || (cat firedrake-install.log && /bin/false)
       - name: Install test dependencies
         run: |
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -80,7 +80,6 @@ jobs:
             --install icepack \
             --install irksome \
             --install femlium \
-            --package-branch pyop2 JDBetteridge/check_something \
             || (cat firedrake-install.log && /bin/false)
       - name: Install test dependencies
         run: |
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -80,7 +80,7 @@ jobs:
             --install icepack \
             --install irksome \
             --install femlium \
-            --package-branch pyop2 JDBetteridge/RAII \
+            --package-branch pyop2 JDBetteridge/check_something \
             || (cat firedrake-install.log && /bin/false)
       - name: Install test dependencies
         run: |
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -80,7 +80,7 @@ jobs:
             --install icepack \
             --install irksome \
             --install femlium \
-            --package-branch pyop2 JDBetteridge/check_something \
+            --package-branch pyop2 JDBetteridge/RAII \
             || (cat firedrake-install.log && /bin/false)
       - name: Install test dependencies
         run: |
@@ -102,7 +102,7 @@ jobs:
             --timeout-method=thread \
             -o faulthandler_timeout=1860 \
             -n 12 --dist worksteal \
-            -v tests
+            -sv tests
         timeout-minutes: 120
       - name: Test pyadjoint
         if: ${{ matrix.scalar-type == 'real' }}
@@ -115,7 +115,7 @@ jobs:
             --timeout-method=thread \
             -o faulthandler_timeout=660 \
             -n 12 --dist worksteal \
-            -v tests/firedrake_adjoint
+            -sv tests/firedrake_adjoint
         timeout-minutes: 30
       - name: Cleanup
         # Belt and braces: clean up before and after the run.
--- a/firedrake/ensemble.py
+++ b/firedrake/ensemble.py
@@ -1,6 +1,7 @@
 from firedrake.petsc import PETSc
 from pyop2.mpi import MPI, internal_comm, decref
 from itertools import zip_longest
+import weakref
 
 __all__ = ("Ensemble", )
 
@@ -25,11 +26,13 @@ class Ensemble(object):
         self.global_comm = comm
         # Internal global comm
         self._global_comm = internal_comm(comm)
+        weakref.finalize(self, decref, self._global_comm)
 
         # User split comm
         self.comm = self.global_comm.Split(color=(rank // M), key=rank)
         # Internal split comm
         self._comm = internal_comm(self.comm)
+        weakref.finalize(self, decref, self._comm)
         """The communicator for spatial parallelism, contains a
         contiguous chunk of M processes from :attr:`global_comm`"""
 
@@ -37,6 +40,7 @@ class Ensemble(object):
         self.ensemble_comm = self.global_comm.Split(color=(rank % M), key=rank)
         # Internal ensemble comm
         self._ensemble_comm = internal_comm(self.ensemble_comm)
+        weakref.finalize(self, decref, self._comm)
         """The communicator for ensemble parallelism, contains all
         processes in :attr:`global_comm` which have the same rank in
         :attr:`comm`."""
@@ -51,10 +55,6 @@ class Ensemble(object):
         if hasattr(self, "ensemble_comm"):
             self.ensemble_comm.Free()
             del self.ensemble_comm
-        for comm_name in ["_global_comm", "_comm", "_ensemble_comm"]:
-            if hasattr(self, comm_name):
-                comm = getattr(self, comm_name)
-                decref(comm)
 
     def _check_function(self, f, g=None):
         """
--- a/firedrake/ensemble.py
+++ b/firedrake/ensemble.py
@@ -1,7 +1,6 @@
 from firedrake.petsc import PETSc
-from pyop2.mpi import MPI, internal_comm, decref
+from pyop2.mpi import MPI, internal_comm
 from itertools import zip_longest
-import weakref
 
 __all__ = ("Ensemble", )
 
@@ -25,22 +24,19 @@ class Ensemble(object):
         # User global comm
         self.global_comm = comm
         # Internal global comm
-        self._global_comm = internal_comm(comm)
-        weakref.finalize(self, decref, self._global_comm)
+        self._global_comm = internal_comm(comm, self)
 
         # User split comm
         self.comm = self.global_comm.Split(color=(rank // M), key=rank)
         # Internal split comm
-        self._comm = internal_comm(self.comm)
-        weakref.finalize(self, decref, self._comm)
+        self._comm = internal_comm(self.comm, self)
         """The communicator for spatial parallelism, contains a
         contiguous chunk of M processes from :attr:`global_comm`"""
 
         # User ensemble comm
         self.ensemble_comm = self.global_comm.Split(color=(rank % M), key=rank)
         # Internal ensemble comm
-        self._ensemble_comm = internal_comm(self.ensemble_comm)
-        weakref.finalize(self, decref, self._comm)
+        self._ensemble_comm = internal_comm(self.ensemble_comm, self)
         """The communicator for ensemble parallelism, contains all
         processes in :attr:`global_comm` which have the same rank in
         :attr:`comm`."""
--- a/firedrake/ensemble.py
+++ b/firedrake/ensemble.py
@@ -14,7 +14,7 @@ class Ensemble(object):
 
         :arg comm: The communicator to split.
         :arg M: the size of the communicators used for spatial parallelism.
-        :kwarg ensememble_name: string used as communicator name prefix, for debugging.
+        :kwarg ensemble_name: string used as communicator name prefix, for debugging.
         :raises ValueError: if ``M`` does not divide ``comm.size`` exactly.
         """
         size = comm.size
--- a/firedrake/ensemble.py
+++ b/firedrake/ensemble.py
@@ -8,12 +8,13 @@ __all__ = ("Ensemble", )
 
 
 class Ensemble(object):
-    def __init__(self, comm, M):
+    def __init__(self, comm, M, **kwargs):
         """
         Create a set of space and ensemble subcommunicators.
 
         :arg comm: The communicator to split.
         :arg M: the size of the communicators used for spatial parallelism.
+        :kwarg ensememble_name: string used as communicator name prefix, for debugging.
         :raises ValueError: if ``M`` does not divide ``comm.size`` exactly.
         """
         size = comm.size
@@ -28,17 +29,18 @@ class Ensemble(object):
         # Internal global comm
         self._comm = internal_comm(comm, self)
 
+        ensemble_name = kwargs.get("ensemble_name", "Ensemble")
         # User and internal communicator for spatial parallelism, contains a
         # contiguous chunk of M processes from `global_comm`.
         self.comm = self.global_comm.Split(color=(rank // M), key=rank)
-        self.comm.name = "Ensemble spatial comm"
+        self.comm.name = f"{ensemble_name} spatial comm"
         weakref.finalize(self, self.comm.Free)
         self._spatial_comm = internal_comm(self.comm, self)
 
         # User and internal communicator for ensemble parallelism, contains all
         # processes in `global_comm` which have the same rank in `comm`.
         self.ensemble_comm = self.global_comm.Split(color=(rank % M), key=rank)
-        self.comm.name = "Ensemble ensemble comm"
+        self.ensemble_comm.name = f"{ensemble_name} ensemble comm"
         weakref.finalize(self, self.ensemble_comm.Free)
         self._ensemble_comm = internal_comm(self.ensemble_comm, self)
 
--- a/firedrake/ensemble.py
+++ b/firedrake/ensemble.py
@@ -1,3 +1,5 @@
+import weakref
+
 from firedrake.petsc import PETSc
 from pyop2.mpi import MPI, internal_comm
 from itertools import zip_longest
@@ -21,36 +23,27 @@ class Ensemble(object):
 
         rank = comm.rank
 
-        # User global comm
+        # Global comm
         self.global_comm = comm
         # Internal global comm
-        self._global_comm = internal_comm(comm, self)
+        self._comm = internal_comm(comm, self)
 
-        # User split comm
+        # User and internal communicator for spatial parallelism, contains a
+        # contiguous chunk of M processes from `global_comm`.
         self.comm = self.global_comm.Split(color=(rank // M), key=rank)
-        # Internal split comm
-        self._comm = internal_comm(self.comm, self)
-        """The communicator for spatial parallelism, contains a
-        contiguous chunk of M processes from :attr:`global_comm`"""
+        self.comm.name = "Ensemble spatial comm"
+        weakref.finalize(self, self.comm.Free)
+        self._spatial_comm = internal_comm(self.comm, self)
 
-        # User ensemble comm
+        # User and internal communicator for ensemble parallelism, contains all
+        # processes in `global_comm` which have the same rank in `comm`.
         self.ensemble_comm = self.global_comm.Split(color=(rank % M), key=rank)
-        # Internal ensemble comm
+        self.comm.name = "Ensemble ensemble comm"
+        weakref.finalize(self, self.ensemble_comm.Free)
         self._ensemble_comm = internal_comm(self.ensemble_comm, self)
-        """The communicator for ensemble parallelism, contains all
-        processes in :attr:`global_comm` which have the same rank in
-        :attr:`comm`."""
 
         assert self.comm.size == M
         assert self.ensemble_comm.size == (size // M)
-
-    def __del__(self):
-        if hasattr(self, "comm"):
-            self.comm.Free()
-            del self.comm
-        if hasattr(self, "ensemble_comm"):
-            self.ensemble_comm.Free()
-            del self.ensemble_comm
 
     def _check_function(self, f, g=None):
         """
@@ -62,7 +55,7 @@ class Ensemble(object):
         :raises ValueError: if function communicators mismatch each other or the ensemble
             spatial communicator, or is the functions are in different spaces
         """
-        if MPI.Comm.Compare(f._comm, self._comm) not in {MPI.CONGRUENT, MPI.IDENT}:
+        if MPI.Comm.Compare(f._comm, self._spatial_comm) not in {MPI.CONGRUENT, MPI.IDENT}:
             raise ValueError("Function communicator does not match space communicator")
 
         if g is not None:
--- a/firedrake/linear_solver.py
+++ b/firedrake/linear_solver.py
@@ -1,4 +1,3 @@
-import gc
 from firedrake.exceptions import ConvergenceError
 import firedrake.function as function
 import firedrake.cofunction as cofunction
@@ -176,6 +175,5 @@ class LinearSolver(OptionsManager):
             raise ConvergenceError("LinearSolver failed to converge after %d iterations with reason: %s", self.ksp.getIterationNumber(), solving_utils.KSPReasons[r])
 
         # Grab the comm associated with `x` and call PETSc's garbage cleanup routine
-        gc.collect()
         comm = x.function_space().mesh()._comm
         PETSc.garbage_cleanup(comm=comm)
--- a/firedrake/linear_solver.py
+++ b/firedrake/linear_solver.py
@@ -1,3 +1,4 @@
+import gc
 from firedrake.exceptions import ConvergenceError
 import firedrake.function as function
 import firedrake.cofunction as cofunction
@@ -175,5 +176,6 @@ class LinearSolver(OptionsManager):
             raise ConvergenceError("LinearSolver failed to converge after %d iterations with reason: %s", self.ksp.getIterationNumber(), solving_utils.KSPReasons[r])
 
         # Grab the comm associated with `x` and call PETSc's garbage cleanup routine
+        gc.collect()
         comm = x.function_space().mesh()._comm
         PETSc.garbage_cleanup(comm=comm)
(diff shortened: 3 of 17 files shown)


firedrake/petsc.py
```diff
@@ -249,3 +251,46 @@ def inserted_options(self):
...
+    if isinstance(obj, (PETSc.Comm, mpi.MPI.Comm)):
+        try:
+            if mpi.is_pyop2_comm(obj):
+                comm = obj
+            else:
+                internal_comm = obj.Get_attr(mpi.innercomm_keyval)
+                if internal_comm is None:
+                    comm = obj
+                else:
+                    comm = internal_comm
```
connorjward, line 267: I have decided that I don't like the policy of "only ever clean up the internal communicator". Consider the following:
```py
mesh = UnitSquareMesh(...)
vec = PETSc.Vec().create(mesh.comm)
garbage_cleanup(mesh.comm)  # or garbage_cleanup(mesh)
```
Since `vec` is associated with the public mesh communicator (`mesh.comm` instead of `mesh._comm`) then it will never be cleaned up by `garbage_cleanup` since that will always act on `mesh._comm`.

I propose that `garbage_cleanup` look like the following:
```py
def garbage_cleanup(obj):
    if isinstance(obj, ParallelObject):
        PETSc.garbage_cleanup(obj._comm)
    else:
        PETSc.garbage_cleanup(obj)
```

In other words, if one passes a parallel Firedrake (or PyOP2) object then call the cleanup on the associated internal communicator. If a communicator is passed instead then clean up *on that communicator* - don't extract the private one.

I added `ParallelObject` here because I would really like a mixin class encapsulating the `.comm`, `._comm`, `__del__`, `mpi.internal_comm` stuff repeated all over the place. This isn't strictly necessary though and the `isinstance()` call could be replaced by `hasattr(obj, "_comm")`.

firedrake/petsc.py
```diff
@@ -249,3 +251,46 @@ def inserted_options(self):
...
+
+
+def garbage_cleanup(obj):
+    """ Cleans up garbage PETSc objects on Firedrake object or any comm
+
+    :arg obj: Any Firedrake object or any comm
+    """
+    gc.collect()
+    comm = _extract_comm(obj)
+    if comm:
```
connorjward, line 284: If we fail to find a communicator we should fail loudly here rather than do nothing.

tests/regression/test_garbage.py
```diff
@@ -0,0 +1,12 @@
+from firedrake import *
+from firedrake.petsc import garbage_cleanup
+
+import pytest
+
+
+@pytest.mark.parallel(2)
+def test_3000_meshes():
```
connorjward, line 8:
```suggestion
def test_making_many_meshes_does_not_exhaust_comms():
```

tests/regression/test_garbage.py
```diff
@@ -0,0 +1,12 @@
+from firedrake import *
+from firedrake.petsc import garbage_cleanup
```
connorjward, line 2: Is this import needed? `garbage_cleanup` is added to `__all__` in `firedrake.petsc`.

firedrake/variational_solver.py
```diff
@@ -282,6 +283,7 @@ def solve(self, bounds=None):
         solving_utils.check_snes_convergence(self.snes)
 
         # Grab the comm associated with the `_problem` and call PETSc's garbage cleanup routine
+        gc.collect()
```
connorjward, line 286: Ditto. Should call our new function.

firedrake/petsc.py
```diff
@@ -249,3 +251,46 @@ def inserted_options(self):
...
+            pass
+    elif hasattr(obj, "_comm"):
+        comm = obj._comm
+    elif hasattr(obj, "comm"):
+        comm = obj.comm
+    return comm
+
+
+def garbage_cleanup(obj):
+    """ Cleans up garbage PETSc objects on Firedrake object or any comm
```
connorjward, line 278: Should be imperative.

connorjward (changes requested):
Thanks for taking this on.

firedrake/petsc.py
wence-: Just to note, when I (many years ago) wrote the comm-duplication code in pyop2 to deliver "library" communicators, I mimiced the petsc setup (since at the time we thought people might use pyop2 without petsc).

So there are three communicators in play in this example:

1. the user communicator (passed in to `Mesh`), `mesh.comm`
2. the internal PyOP2 communicator `mesh._comm`
3. the internal PETSc communicator `vec.comm` (which is not `mesh.comm`, but is stashed inside it), see https://petsc.org/release/manualpages/Sys/PetscCommDuplicate/

When you do `VecCreate(user_comm, &vec)` internally petsc does `PetscCommDuplicate(user_comm, &petsc_comm)` and then `PetscObjectComm((PetscObject)vec)` returns `petsc_comm`. It used to be that Firedrake/PyOP2 did similar, so `mesh = Mesh(..., comm=user_comm)` did `firedrake_comm = dup_comm(user_comm)` and then `mesh.comm` would be `firedrake_comm` (not `user_comm`).

I think all of the garbage lives on the internal PETSc communicator, but petsc doesn't know about the connection between `mesh._comm` and `mesh.comm` so there will be two petsc communicators and not all the garbage will be collected.

I'm sure there are good reasons why `mesh._comm` is not just the communicator obtained from `PetscCommDuplicate` but I can't remember them. OTOH, if that were the case, would that make things simpler? `P etscGarbageCleanup` always cleans up on the internal petsc communicator, so as long as `mesh._comm` and `mesh.comm` have the same internal petsc communicator it doesn't matter if you're collecting on the public or he private one.

connorjward: > When you do `VecCreate(user_comm, &vec)` internally petsc does `PetscCommDuplicate(user_comm, &petsc_comm)` and then `PetscObjectComm((PetscObject)vec)` returns `petsc_comm`. It used to be that Firedrake/PyOP2 did similar, so `mesh = Mesh(..., comm=user_comm)` did `firedrake_comm = dup_comm(user_comm)` and then `mesh.comm` would be `firedrake_comm` (not `user_comm`).

I don't especially like the pattern of `mesh = Mesh(..., comm=mycomm); mesh.comm != mycomm`.

> I'm sure there are good reasons why `mesh._comm` is not just the communicator obtained from `PetscCommDuplicate` but I can't remember them. OTOH, if that were the case, would that make things simpler? `P etscGarbageCleanup` always cleans up on the internal petsc communicator, so as long as `mesh._comm` and `mesh.comm` have the same internal petsc communicator it doesn't matter if you're collecting on the public or he private one.

I think I'm in favour of dropping PyOP2-specific comms so our notions of public and private comms could match PETSc's. The majority of MPI calls than happen in PyOP2 are done via PETSc anyway (e.g. halo exchanges with PetscSFs) so I don't see where the benefit is.

JDBetteridge: I disagree with this assessment, I believe that the example user code you wrote above should look like:
```python
mesh = UnitSquareMesh(...)
vec = PETSc.Vec().create(mesh.comm)
PETSc.garbage_cleanup(mesh.comm)
```
As is now [documented in petsc4py](https://petsc.org/release/petsc4py/reference/petsc4py.PETSc.garbage_cleanup.html). If a user is doing petsc4py things they should be cleaning them up using petsc4py methods, if they are doing Firedrake things they should be cleaning them up using Firedrake methods (or indeed not cleaning them up, Firedrake is doing this periodically anyway!). I think the Firedrake cleanup function cleaning petsc4py objects off a user comm (neither item "belonging" to Firedrake) is confusing.

I agree with `mesh = Mesh(..., comm=mycomm); mesh.comm != mycomm` being a bad idea, but I'm not in favour of dropping internal comms. Having internal "library" communicators is useful, there is a good summary and example on [this rookieHPC page](https://rookiehpc.org/mpi/docs/mpi_comm_dup/index.html). Indeed I'm happy with the three tiers of communicator, user comm, Firedrake internal/PyOP2 comm and PETSc comm. Firedrake is a library (for the purposes of this discussion) and PyOP2 handles the duplicating the user comm, PETSc is a separate library and should also have its own duplicated comm, whether that be a user comm, or a Firedrake.PyOP2 comm.

Let me know your thoughts.

firedrake/petsc.py
JDBetteridge, line 278:
```suggestion
    """ Clean up garbage PETSc objects on Firedrake object or any comm
```

tests/regression/test_garbage.py
JDBetteridge: No, I think I was in two minds whether it should actually be added to the list of things imported by `from firedrake import *` or whether I want to force users to import it as a "I know what I'm doing and really want to manually clean things up step".

JDBetteridge: What do you think?

firedrake/linear_solver.py
```diff
@@ -175,5 +176,6 @@ def solve(self, x, b):
             raise ConvergenceError("LinearSolver failed to converge after %d iterations with reason: %s", self.ksp.getIterationNumber(), solving_utils.KSPReasons[r])
 
         # Grab the comm associated with `x` and call PETSc's garbage cleanup routine
+        gc.collect()
         comm = x.function_space().mesh()._comm
         PETSc.garbage_cleanup(comm=comm)
```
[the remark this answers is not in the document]
JDBetteridge: I think we can just call the petsc4py routine here, since we know exactly what communicator we want to clean up. I also think calling `gc.collect()` here might be really expensive and it might need to be removed.

If a user calls `garbage_cleanup()` I don't care if it's slow. The user might...

firedrake/petsc.py
JDBetteridge: I'll emit a warning, but I don't want to raise an exception.

Some users will just ignore warnings though.

firedrake/variational_solver.py
JDBetteridge: *see above

tests/regression/test_garbage.py
connorjward: I like the idea of requiring an explicit import. As you say it implies that it isn't a thing that most users need to do.

firedrake/linear_solver.py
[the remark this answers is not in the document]
connorjward: I vote that `gc.collect()` be removed. It's an expensive call and no one should notice since it happens internally. We definitely need it for the public cleanup interface, as otherwise users might think it's a bug, but here it can go.

firedrake/petsc.py
```diff
@@ -7,10 +7,17 @@
...
 from contextlib import contextmanager
+from pyop2 import mpi
+from typing import Any
 
 
-__all__ = ("PETSc", "OptionsManager", "get_petsc_variables")
+__all__ = (
+    "PETSc", "OptionsManager", "get_petsc_variables", "garbage_cleanup",
+    "garbage_view",
+)
```
connorjward, line 20:
```suggestion
__all__ = ("PETSc", "OptionsManager", "get_petsc_variables")
```
lets keep these private.

firedrake/petsc.py
connorjward: Fine. But our docstring for `garbage_cleanup` should make this clear.

connorjward: > example on [this rookieHPC page](https://rookiehpc.org/mpi/docs/mpi_comm_dup/index.html).

ohh... that example makes this all make sense!

firedrake/petsc.py
```diff
@@ -249,3 +256,70 @@ def inserted_options(self):
         finally:
             for k in self.to_delete:
                 del self.options_object[self.options_prefix + k]
+
+
+def _extract_comm(obj):
```
connorjward, line 261: Please add a docstring.

firedrake/petsc.py
```diff
@@ -249,3 +256,70 @@ def inserted_options(self):
...
+    comm = None
+    if isinstance(obj, (PETSc.Comm, mpi.MPI.Comm)):
+        try:
+            if mpi.is_pyop2_comm(obj):
+                comm = obj
+            else:
+                internal_comm = obj.Get_attr(mpi.innercomm_keyval)
```
connorjward, line 268: Is there a nice PyOP2 function for this?

firedrake/petsc.py
connorjward: Why not raise an exception? We are already assuming that users "know what they are doing" if they are calling this.

firedrake/petsc.py
```diff
@@ -249,3 +256,70 @@ def inserted_options(self):
...
+    # cleanup routine. This routine is designed to free up as much memory as
+    # possible for memory constrained systems
+    gc.collect()
+    comm = _extract_comm(obj)
+    if comm:
+        PETSc.garbage_cleanup(comm)
+    else:
+        warnings.warn(
+            "No comm on extracted from object, "
+            "not calling `PETSc.garbage_cleanup`"
```
connorjward, line 302:
```suggestion
            "No comm found, skipping garbage cleanup"
```

firedrake/petsc.py
```diff
@@ -249,3 +256,70 @@ def inserted_options(self):
...
+    # We are manually calling the Python cyclic garbage collection routine so
+    # that as many unreachable PETSc objects are visible in the garbage view.
+    gc.collect()
+    comm = _extract_comm(obj)
+    if comm:
+        PETSc.garbage_view(comm)
+    else:
+        warnings.warn(
+            "No comm on extracted from object, "
+            "not calling `PETSc.garbage_view`"
```
connorjward, line 324:
```suggestion
            "No comm found, skipping garbage cleanup"
```

connorjward (changes requested):

firedrake/petsc.py
JDBetteridge: err...

firedrake/petsc.py
JDBetteridge: Yes, but it increases the reference count. In this instance, since we do not return the communicator, I do not think that is necessary. However, when I fix [PyOP2 issue 702](https://github.com/OP2/PyOP2/issues/702) I can duplicate it and let the weakref finalizer clean things up.

JDBetteridge: I'm unresolving this until I fix https://github.com/OP2/PyOP2/issues/702

firedrake/petsc.py
JDBetteridge: Ooops, meant to apply that suggestion!

firedrake/petsc.py
JDBetteridge: Because raising an exception in parallel can cause a deadlock if the exception is only raised on some ranks.

Despite what [this comment](https://github.com/OP2/PyOP2/blob/master/pyop2/mpi.py#L513) suggests, I have definitely experienced deadlocks as a result of some, but not all, ranks raising an exception!

firedrake/petsc.py
```diff
@@ -249,3 +254,77 @@ def inserted_options(self):
...
+    return comm
+
+
+def garbage_cleanup(obj: Any):
+    """ Clean up garbage PETSc objects on Firedrake object or any comm
+
+    Parameters
+    ----------
+    obj:
+        Any Firedrake object or any comm
```
ReubenHill, line 299:
```suggestion
        Any Firedrake object with a comm, or any comm.
```

firedrake/petsc.py
JDBetteridge: Now raising exception following discussion.

firedrake/petsc.py
JDBetteridge: applied manually

.github/workflows/build.yml
```diff
@@ -79,6 +80,7 @@ jobs:
             --install icepack \
             --install irksome \
             --install femlium \
+            --package-branch pyop2 JDBetteridge/check_something \
```
JDBetteridge, line 83:
```suggestion
```

firedrake/petsc.py
```diff
@@ -259,3 +268,79 @@ def inserted_options(self):
         finally:
             for k in self.to_delete:
                 del self.options_object[self.options_prefix + k]
+
+
+def _extract_comm(obj: Any) -> MPI.Comm:
+    """ Extract Firedrake internal comm off given object
```
connorjward, line 274:
```suggestion
    """Extract and return the Firedrake internal comm of a given object.
```

firedrake/petsc.py
```diff
@@ -259,3 +268,79 @@ def inserted_options(self):
...
+                internal_comm = obj.Get_attr(mpi.innercomm_keyval)
+                if internal_comm is None:
+                    comm = obj
+                else:
+                    comm = internal_comm
+        except mpi.PyOP2CommError:
+            pass
+    elif hasattr(obj, "_comm"):
+        comm = obj._comm
+    elif hasattr(obj, "comm"):
```
connorjward, line 302: This is not what the docstring says this function does. This isn't the "internal Firedrake comm".

firedrake/petsc.py
```diff
@@ -259,3 +268,79 @@ def inserted_options(self):
...
+        Any Firedrake object or any comm
+
+    Returns
+    -------
+    MPI.Comm
+        Internal communicator
+
+    """
+    comm = None
+    if isinstance(obj, (PETSc.Comm, mpi.MPI.Comm)):
```
connorjward, line 288: The behaviour of this branch should be made clear in the docstring I think.

firedrake/petsc.py
```diff
@@ -259,3 +268,79 @@ def inserted_options(self):
...
+    elif hasattr(obj, "_comm"):
+        comm = obj._comm
+    elif hasattr(obj, "comm"):
+        comm = obj.comm
+    return comm
+
+
+@mpi.collective
+def garbage_cleanup(obj: Any):
+    """ Clean up garbage PETSc objects on Firedrake object or any comm
```
connorjward, line 309:
```suggestion
    """Clean up garbage PETSc objects on a Firedrake object or any comm.
```

firedrake/petsc.py
```diff
@@ -259,3 +268,79 @@ def inserted_options(self):
...
+    comm = _extract_comm(obj)
+    if comm:
+        PETSc.garbage_cleanup(comm)
+    else:
+        raise FiredrakePETScError("No comm found, cannot clean up garbage")
+
+
+@mpi.collective
+def garbage_view(obj: Any):
+    """ View garbage PETSc objects on Firedrake object or any comm
```
connorjward, line 331:
```suggestion
    """View garbage PETSc objects stored on a Firedrake object or any comm.
```

connorjward (changes requested):
Largely complaints about documentation.

firedrake/petsc.py
JDBetteridge: Ah, I see the confusion here, I have amended the docstring to say Firedrake/PyOP2 internal comm. This covers the case where the object is actually a "PyOP2 object"

firedrake/ensemble.py
```diff
@@ -21,37 +23,28 @@ def __init__(self, comm, M):
...
 
-        # User split comm
+        # User and internal communicator for spatial parallelism, contains a
+        # contiguous chunk of M processes from `global_comm`.
         self.comm = self.global_comm.Split(color=(rank // M), key=rank)
-        # Internal split comm
-        self._comm = internal_comm(self.comm, self)
-        """The communicator for spatial parallelism, contains a
-        contiguous chunk of M processes from :attr:`global_comm`"""
+        self.comm.name = "Ensemble spatial comm"
```
JHopeCollins, line 34: If these names are used for debugging could we have an optional str argument for an Ensemble name? So these would be
```
comm_name = "Ensemble" if optional_comm_name is None else optional_comm_name
self.comm.name = f"{comm_name} spatial comm"
```

firedrake/ensemble.py
```diff
@@ -21,37 +23,28 @@ def __init__(self, comm, M):
 
         rank = comm.rank
 
-        # User global comm
+        # Global comm
         self.global_comm = comm
         # Internal global comm
-        self._global_comm = internal_comm(comm, self)
+        self._comm = internal_comm(comm, self)
```
JHopeCollins, line 29: I'm not sure I like the inconsistency between `self.global_comm` and `self._comm` and between `self.comm` and `self._spatial_comm` but I guess that can't be helped seeing as:
- the garbage collection assumes that an object is defined over `obj._comm` and
- a user expects `obj.comm` to be the "spatial" comm associated to the object.

JHopeCollins (changes requested):

firedrake/ensemble.py
```diff
@@ -21,37 +23,28 @@ def __init__(self, comm, M):
...
+        self.comm.name = "Ensemble spatial comm"
+        weakref.finalize(self, self.comm.Free)
+        self._spatial_comm = internal_comm(self.comm, self)
 
-        # User ensemble comm
+        # User and internal communicator for ensemble parallelism, contains all
+        # processes in `global_comm` which have the same rank in `comm`.
         self.ensemble_comm = self.global_comm.Split(color=(rank % M), key=rank)
-        # Internal ensemble comm
+        self.comm.name = "Ensemble ensemble comm"
```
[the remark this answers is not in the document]
JDBetteridge: Done

firedrake/ensemble.py
JDBetteridge: I understand your point, but `_comm` really is an internal implementation detail. I will leave this unresolved for wider discussion, just in case.

firedrake/ensemble.py
```diff
@@ -1,17 +1,20 @@
...
 
 class Ensemble(object):
-    def __init__(self, comm, M):
+    def __init__(self, comm, M, **kwargs):
         """
         Create a set of space and ensemble subcommunicators.
 
         :arg comm: The communicator to split.
         :arg M: the size of the communicators used for spatial parallelism.
+        :kwarg ensememble_name: string used as communicator name prefix, for debugging.
```
JDBetteridge, line 17:
```suggestion
        :kwarg ensemble_name: string used as communicator name prefix, for debugging.
```

JHopeCollins (approved):
Looks good to me

connorjward (approved):

Issue #456, opened 2015-02-10
Support for local_range


ctjacobs:
For a function `f`, I'm doing `f.vector().local_range()` in DOLFIN. Is there a way of doing something similar in Firedrake? `local_range` is not available in Firedrake's Vector class. 

```
from firedrake import *

mesh = UnitSquareMesh(10, 10)
fs = FunctionSpace(mesh, "CG", 1)
f = Function(fs)

print f.vector().local_range()
```

```
Traceback (most recent call last):
  File "vector.py", line 7, in <module>
    print f.vector().local_range()
AttributeError: 'Vector' object has no attribute 'local_range'
```

wence-:
What is local_range expected to return?

ctjacobs:
The range of dof indices owned by a particular process.

wence-:
In global or local numbering?  i.e. is it:

```
Proc 0:
[0, 10)
Proc 1:
[10, 15)
```

Or
```
Proc 0:
[0, 10)
Proc 1:
[0, 5)
```
?

ctjacobs:
Global.

```
from dolfin import *

mesh = UnitSquareMesh(10, 10)
fs = FunctionSpace(mesh, "CG", 1)
f = Function(fs)

print f.vector().local_range()
```
gives
```
(0, 121)
```
on one process, and 
```
Number of global vertices: 121
Number of global cells: 200
(59, 121)
(0, 59)
```
on two processes.

wence-:
What about for mixed function spaces?

wence-:
Note that for us, each sub space is "separate"

wence-:
and the "vector().array()" returns a list of arrays in that case.

ctjacobs:
For mixed functions it appears that `.vector().array()` returns some kind of all-in-one list of function values:

```
from dolfin import *

mesh = UnitSquareMesh(10, 10)
vfs = VectorFunctionSpace(mesh, "CG", 1)
fs = FunctionSpace(mesh, "CG", 1)
mfs = MixedFunctionSpace([vfs, fs])

f = Function(mfs)

print f.vector().array()
print f.vector().local_range()
```

```
ctj10@tycho:/data/ctj10$ mpiexec -n 2 python vector.py 
Number of global vertices: 121
Number of global cells: 200
[ 0.  0.  0.  0.  0.  0.  0.  0.  0.  0.  0.  0.  0.  0.  0.  0.  0.  0.
[...]

Issue #2824, opened 2023-03-16
CI crashing out with `exit code 137`


JDBetteridge:
We are running out of memory on the CI. I have now profiled the test suite and found the source of the issue to be the test `test_firedrake_helmholtz_scalar_convergence_on_hex` in `tests/regression/test_helmholtz.py` introduced in commit `e68b9f63644fce64fd23c85031261e29aae2dd88`.

This issue should be treated as fairly urgent as currently it is not possible to run tests locally unless you have a pretty beefy machine (at least 64GB RAM!) and even then you are at danger of running out of memory if using `xdist` to parallelise pytest.

I attach two plots showing the memory profile for the test suite (run without `xdist`).

![full_test_suite](https://user-images.githubusercontent.com/43041811/225592000-7263a7ed-5e67-4c10-a010-4dd0dc31955a.png)
![test_helmholtz](https://user-images.githubusercontent.com/43041811/225592007-427e9da6-5141-4b46-a850-7e477db805f0.png)

ksagiyam:
This test is large as it is a convergence test on hex mesh, and the hex mesh must be such that it contains all possible facet orientations. I might have to create a reasonable mesh by hand.

JDBetteridge:
We should also have some policy on acceptable test sizes. Off the top of my head a good starting point would be:
- Test duration <1minute
- MPI ranks <=4
- Total memory <4GB

Runner hardware is currently 48 physical cores, 64GB RAM. Four Github runners share this hardware, tests are run using pytest xdist with `-n 12` (currently `-n 8` to try and mitigate this issue).

With possible exceptions being number of ranks could be greater for testing communicator functionality, or for testing `Ensemble`. In these cases we should only break one of these limits.

We should also concretise these limits on the wiki.

JDBetteridge:
@ksagiyam I think constructing the mesh by hand to reduce the size would be a good idea. Could you split the problem up and have a set of meshes which _together_ cover all possible orientations, rather than one big mesh containing all orientations?

ksagiyam:
That could be a good option, actually.

wence-:
Or compute some cohomology which is topological, but I presume would be sensitive to orientations being incorrect.

ksagiyam:
Ok, sounds good. For now let me just quickly do polynomial interpolation tests to fix CI.

connorjward:
@JDBetteridge can this be closed?

JDBetteridge:
[...]

PR #3159, opened 2023-10-12
TransferManager: Re-injection of coefficients; no-op for unmodified `dat`

pbrubeck:
---
name: "TransferManager: Re-injection of coefficients; no-op for unmodified `dat`"
description: ''
title: ''
labels: ''
assignees: ''

---

# Description
I'm adding a cache on the arguments of `tm.prolong, tm.restrict, tm.inject` that holds the `dat_versions` from the last time the operator was applied on those arguments. The transfer operator will behave as a no-op if the current versions coincide with the ones in the cache.

I also add a coarsen hook at restriction with the injection of all coefficients of the problem at hand. Here we avoid the unnecessary re-injection of coefficients that remain constant.

## Associated Pull Requests:
Need to merge this PR first
- [Cofunction coarsening #3181](https://github.com/firedrakeproject/firedrake/pull/3181)

## Fixes Issues:
- Make a list of issues (with links!)
- [Like this Issue#2824](https://github.com/firedrakeproject/firedrake/issues/2824)

If issues are fixed by this PR, prepend each of them with the word
"fixes", so they are automatically closed when this PR is merged. For
example "fixes #123, fixes #456".

# Checklist for author:

- [ ] I have linted the codebase (`make lint` in the `firedrake` source directory).

# Checklist for reviewer:

- [ ] Docstrings present
- [ ] New tests present
- [ ] No bad "code smells"
- [ ] No issues in parallel


firedrake/mg/embedded.py
```diff
@@ -265,37 +274,44 @@ def restrict(self, gf, gc):
         """
         Vc = gc.function_space()
         Vf = gf.function_space()
-
         source_element = Vf.ufl_element()
         target_element = Vc.ufl_element()
+
+        cache = self.cache(source_element)
```
connorjward, line 280: ~This block could definitely do with a comment.~ A sensible function would avoid this requirement.

firedrake/mg/embedded.py
```diff
@@ -265,37 +274,44 @@ def restrict(self, gf, gc):
...
+        key = (Op.RESTRICT, gf, gc)
```
connorjward: Your caches will now be holding onto "big" data structures. Is memory consumption going to be a concern?

firedrake/mg/embedded.py
connorjward: And could the whole thing be refactored into a function:
```py
def requires_update(self, element, op, uc, uf):
    key = (op, uc, uf)
    ...
    return dat_versions != cache._dat_versions[key]
```

The method name could probably be made even clearer than this. Also include a docstring.

firedrake/mg/embedded.py
```diff
@@ -26,6 +26,7 @@ class Cache(object):
         :arg element: The element to use for the caching."""
         def __init__(self, element):
             self.embedding_element = get_embedding_dg_element(element)
+            self._dat_versions = {}
```
connorjward, line 29: I think best practice is to [document these things as attributes of the class](https://numpydoc.readthedocs.io/en/latest/format.html#class-docstring), but I don't think that enough people are bothered about this for it to be necessary here. Perhaps a comment?

connorjward (changes requested):

firedrake/mg/embedded.py
wence-: Yes. Use: `key = (op, weakref.ref(source), weakref.ref(target))`

pbrubeck: Alright, I'll use `key = (op, weakref.ref(source.dat), weakref.ref(target.dat))`

wence-: Needs manual cleaning out of the dictionary occasionally if any of the weakrefs are dead.

connorjward: Would a [`WeakKeyDictionary`](https://docs.python.org/3/library/weakref.html#weakref.WeakKeyDictionary) not be the right thing here?

pbrubeck: I don't think so. The whole key is a tuple, we don't want a weakref to that.

wence-: No, for two reasons:
1. you can't take a weakref of a tuple
2. even if you could, the key is constructed in this function, so the strong reference to the key doesn't live past the end of the function, and therefore the key-value mapping would immediately be removed.

firedrake/mg/ufl_utils.py
```diff
@@ -156,41 +170,52 @@ def coarsen_function_space(V, self, coefficient_mapping=None):
...
+        from firedrake.bcs import DirichletBC
+        manager = firedrake.dmhooks.get_transfer_manager(fine)
+        finectx = firedrake.dmhooks.get_appctx(fine)
+        coefficients = finectx.F.coefficients() + finectx.J.coefficients()
+        if finectx.Jp is not None:
+            coefficients.extend(finectx.Jp.coefficients())
+        coefficients = tuple(OrderedDict.fromkeys(coefficients))
+        for fine in coefficients:
+            if hasattr(fine, '_child'):
+                coarse = fine._child.__repr__.__self__
```
pbrubeck, line 204: @connorjward @wence- what is the proper way of getting the strong reference to the `weakref.proxy` defined on line 185? The reason I do this is because we later do type checking inside `inject` and `restrict`, and otherwise the type is `weakref.proxy`.

connorjward: I'm not sure that it's possible to do this. weakrefs can be quite fussy. Are you sure that weakref is the right thing to be doing here?

pbrubeck: The alternative is to have a different mechanism for type checking, one that does not rely on the actual type, but on an interface. Since Cofunction are one-forms, the type is encoded in `hasattr(fine._child, "arguments")`.

connorjward: This seems like a code smell to me. Is there no other way to know whether or not we have a `Function` or `Cofunction` here?

firedrake/mg/ufl_utils.py
```diff
@@ -133,7 +146,10 @@ def coarsen_function_space(V, self, coefficient_mapping=None):
 
     mesh = self(V.mesh(), self)
 
-    V = firedrake.FunctionSpace(mesh, V.ufl_element())
+    if isinstance(V, firedrake.functionspaceimpl.FiredrakeDualSpace):
+        V = firedrake.functionspace.DualSpace(mesh, V.ufl_element())
+    else:
+        V = firedrake.FunctionSpace(mesh, V.ufl_element())
```
wence-, line 152: We should add an interface to reconstruct on a new mesh.

firedrake/mg/ufl_utils.py
```diff
@@ -156,41 +172,58 @@ def coarsen_function_space(V, self, coefficient_mapping=None):
...
-    if new is None:
-        mesh = self(extract_unique_domain(expr), self)
-        new = firedrake.Constant(numpy.zeros(expr.ufl_shape,
-                                             dtype=expr.dat.dtype),
-                                 domain=mesh)
-        # Share data pointer
-        new.dat = expr.dat
+        Vf = expr.function_space()
+        Vc = self(Vf, self)
+        new = firedrake.Function(Vc, name="coarse_%s" % expr.name())
```
wence-, line 187: Remind me, `firedrake.Function` makes a `CoFunction` if `Vc` is a dual space?

pbrubeck: Correct

firedrake/mg/ufl_utils.py
pbrubeck: Yes, we are now using `ufl.duals.is_dual`.

tests/multigrid/test_transfer_manager.py
```diff
@@ -41,3 +53,68 @@ def test_transfer_manager_inside_coarsen(sub):
...
+
+    source = Function(Vsource)
+    target = Function(Vtarget)
+    if complex_mode and ((family == "DG" and transfer_op == "inject")
+                         or family not in {"CG", "DG"}):
+        with pytest.raises(NotImplementedError):
+            op(source, target)
+        return
+
+    # Test that the operator produces an output for an unrecognized input
```
connorjward, line 82: I think that these should all be separate tests.

tests/multigrid/test_poisson_gmg.py
```diff
@@ -156,3 +162,64 @@ def test_baseform_coarsening(typ):
...
+    its_new = new_solver.snes.ksp.getIterationNumber()
+    res_new = new_solver.snes.getFunctionNorm()
+    assert its_reused == its_new
+    assert numpy.isclose(res_reused, res_new)
+
+
+@pytest.mark.parametrize("typ",
+                         ["mg", "mgmatfree"])
+def test_reinjection(typ):
+    run_helmholtz(typ)
```
connorjward, line 225: Maybe these could have more detailed test names? Also why is `run_helmholtz` inside `test_poisson_gmg`?

firedrake/mg/ufl_utils.py
```diff
@@ -180,35 +183,47 @@ def coarsen_function(expr, self, coefficient_mapping=None):
...
+                    manager.restrict(fine, coarse)
+                else:
+                    manager.inject(fine, coarse)
+        for bc in chain(*finectx._problem.bcs):
+            if isinstance(bc, DirichletBC):
+                bc.apply(finectx._x)
+
+    V = problem.u.function_space()
+    if not hasattr(V, "_coarse"):
+        # The hook is persistent and cummulative, but also problem-independent.
```
connorjward, line 223:
```suggestion
        # The hook is persistent and cumulative, but also problem-independent.
```

firedrake/variational_solver.py
```diff
@@ -238,7 +238,21 @@ def set_transfer_manager(self, manager):
             TransferManager interface.
         :raises ValueError: if called after the transfer manager is setup.
         """
+        from firedrake import function, cofunction
```
dham, line 241: shouldn't be necessary.

tests/multigrid/test_poisson_gmg.py
```diff
@@ -156,3 +162,64 @@ def test_baseform_coarsening(typ):
...
+    new_solver.set_transfer_manager(transfer)
+    uh.assign(0)
+    new_solver.solve()
+    its_new = new_solver.snes.ksp.getIterationNumber()
+    res_new = new_solver.snes.getFunctionNorm()
+    assert its_reused == its_new
+    assert numpy.isclose(res_reused, res_new)
+
+
+@pytest.mark.parametrize("typ",
```
dham, line 222: descriptive variable name please.

firedrake/variational_solver.py
```diff
@@ -239,6 +234,19 @@ def set_transfer_manager(self, manager):
...
+        seen = set()
+        coefficients = problem.F.coefficients() + problem.J.coefficients()
+        if problem.Jp is not None:
+            coefficients = coefficients + problem.Jp.coefficients()
+        for val in chain(coefficients, (bc.function_arg for bc in problem.bcs)):
+            if isinstance(val, (function.Function, cofunction.Cofunction)):
+                V = val.function_space()
+                if V not in seen:
+                    dm = V.dm
+                    dmhooks.push_appctx(dm, ctx)
```
pbrubeck, line 248: @wence- we need this to avoid the warning "Creating a new transfer manager, this could be expensive". I am not sure where this should go.

connorjward (changes requested):
This basically looks fine to me.

firedrake/variational_solver.py
wence-: So I think what you need is for the appcontext to be set on all the function space DMs when you call `SNESSolve` (rather than just the DM corresponding to the functionspace that the solution lives in).

So I think you would do it with a `dmhooks.add_hooks` call in `NLVS.solve` (similar to how the appctx is attached to the solution's DM). That manages removing the appctx afterwards (rather than this approach which just continues to stack things up and will hold on to the ctx forever).

firedrake/variational_solver.py
```diff
@@ -239,6 +234,19 @@ def set_transfer_manager(self, manager):
...
         """
         self._ctx.transfer_manager = manager
+        problem = self._problem
+        ctx = self._ctx
+        seen = set()
+        coefficients = problem.F.coefficients() + problem.J.coefficients()
+        if problem.Jp is not None:
+            coefficients = coefficients + problem.Jp.coefficients()
+        for val in chain(coefficients, (bc.function_arg for bc in problem.bcs)):
+            if isinstance(val, (function.Function, cofunction.Cofunction)):
```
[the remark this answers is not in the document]
pbrubeck: Yes, `ufl.coefficient.BaseCoefficient`, but we opted to be more descriptive.

firedrake/variational_solver.py
```diff
@@ -258,22 +253,36 @@ def solve(self, bounds=None):
...
         work = self._work
-        with self._problem.u.dat.vec as u:
+        with problem.u.dat.vec as u:
             u.copy(work)
             with ExitStack() as stack:
                 # Ensure options database has full set of options (so monitors
                 # work right)
-                for ctx in chain((self.inserted_options(), dmhooks.add_hooks(dm, self, appctx=self._ctx)),
+                for ctx in chain((self.inserted_options(),
+                                  *(dmhooks.add_hooks(dm, self, appctx=self._ctx) for dm in reversed(problem_dms))),
```
pbrubeck, line 285: @wence- Apparently the order of the contexts matters, we need to visit the solution DM after all other coefficients. Does this look good?

wence-: :shrug: Probably? Hooks are torn down in LIFO order, but I don't know why that is important in this scenario. What happens if you don't do it this way round?

wence-: Minor, no need to splat the generator since you're in a `chain`, so:

```
chain([self.inserted_options()], (dmhooks.add_hooks(...) for dm in ...), self._transfer_operators)
```
I suppose.

pbrubeck: I can't put in a generator, either I pass options and hooks together in the same tuple as I had before, or I use your suggestion with `tuple` in front of the generator.

pbrubeck: > 🤷 Probably? Hooks are torn down in LIFO order, but I don't know why that is important in this scenario. What happens if you don't do it this way round?

If I don't use `reverse` then the appctx does not get attached properly and I get the warning "Creating new TransferManager".

wence-: Huh, odd, anyway, I think that is fine.

PR #3161, opened 2023-10-13
DO NOT MERGE! test upstream PETSc 2023-10-13

angus-g:


ksagiyam:
That alone unfortunately does not test the branch; please see, e.g., https://github.com/firedrakeproject/firedrake/pull/3140/files.

It might also be necessary (and is a good practice) to also update our `slepc` fork to match the petsc version by adding `--package-branch slepc ...`.

connorjward:
> That alone unfortunately does not test the branch; please see, e.g., https://github.com/firedrakeproject/firedrake/pull/3140/files.
> 
> It might also be necessary (and is a good practice) to also update our `slepc` fork to match the petsc version by adding `--package-branch slepc ...`.

Oh yeah, I need to update the procedure I wrote before.

connorjward:
Done: https://github.com/firedrakeproject/firedrake/wiki/Running-the-Firedrake-test-suite-with-different-branches-for-subpackages

angus-g:
Thanks for the updates, I think I have things working correctly now. I don't have access to firedrakeproject/slepc, but when I did the initial upstream merge there was only a single extra commit which only changed tolerances on one of the examples, so it should be harmless. There are a couple more commits now, although I haven't updated the PETSc branch since those either.

connorjward:
Merged. Thank you for this!

PR #3167, opened 2023-10-17
Correct transpose input-ordering interpolation

ReubenHill:
We need to ensure the target vec is zeroed for SF Reduce to
represent multTranspose in case the interpolation matrix is not
square (in which case it will have columns which are zero). This
happens when we interpolate from an input-ordering vertex-only
mesh to an immersed vertex-only mesh where the input ordering
contains points that are not in the immersed mesh. The resulting
interpolation matrix will have columns of zeros for the points
that are not in the immersed mesh. The adjoint interpolation
matrix will then have rows of zeros for those points.

### DIFF
--- a/firedrake/interpolation.py
+++ b/firedrake/interpolation.py
@@ -1627,4 +1627,14 @@ class VomOntoVomDummyMat(object):
         if self.forward_reduce:
             self.broadcast(source_vec, target_vec)
         else:
+            # We need to ensure the target vec is zeroed for SF Reduce to
+            # represent multTranspose in case the interpolation matrix is not
+            # square (in which case it will have columns which are zero). This
+            # happens when we interpolate from an input-ordering vertex-only
+            # mesh to an immersed vertex-only mesh where the input ordering
+            # contains points that are not in the immersed mesh. The resulting
+            # interpolation matrix will have columns of zeros for the points
+            # that are not in the immersed mesh. The adjoint interpolation
+            # matrix will then have rows of zeros for those points.
+            target_vec.getArray()[:] = 0.0
             self.reduce(source_vec, target_vec)
--- a/firedrake/interpolation.py
+++ b/firedrake/interpolation.py
@@ -1636,5 +1636,5 @@ class VomOntoVomDummyMat(object):
             # interpolation matrix will have columns of zeros for the points
             # that are not in the immersed mesh. The adjoint interpolation
             # matrix will then have rows of zeros for those points.
-            target_vec.zero()
+            target_vec.zeroEntries()
             self.reduce(source_vec, target_vec)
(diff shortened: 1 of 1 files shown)


firedrake/interpolation.py
```diff
@@ -664,10 +664,16 @@ def interpolate(self, *function, output=None, transpose=False, **kwargs):
             if transpose:
                 mul = assembled_interpolator.handle.multTranspose
                 V = self.arguments[0].function_space()
+                result = output or firedrake.Function(V)
```
ReubenHill, line 667: This should be a `Cofunction` but it's not enforced anywhere else in interpolation...

firedrake/interpolation.py
```diff
@@ -664,10 +664,16 @@ def interpolate(self, *function, output=None, transpose=False, **kwargs):
...
+                # need to zero the output in case our interpolation matrix is
+                # non-square (e.g. going from a vertex-only mesh to its input
+                # ordering where not all points in the input ordering are in
+                # the vertex-only mesh)
+                result.zero()
```
ReubenHill, line 672: `Cofunction` has no `zero()` method

firedrake/interpolation.py
```diff
@@ -1429,4 +1429,14 @@ def multTranspose(self, source_vec, target_vec):
...
+            # We need to ensure the target vec is zeroed for SF Reduce to
+            # represent multTranspose in case the interpolation matrix is not
+            # square (in which case it will have columns which are zero). This
+            # happens when we interpolate from an input-ordering vertex-only
+            # mesh to an immersed vertex-only mesh where the input ordering
+            # contains points that are not in the immersed mesh. The resulting
+            # interpolation matrix will have columns of zeros for the points
+            # that are not in the immersed mesh. The adjoint interpolation
+            # matrix will then have rows of zeros for those points.
+            target_vec.getArray()[:] = 0.0
```
wence-, line 1441: `target_vec.zero()` makes this future-proof against GPU usage.

dham, line 1441:
```suggestion
            target_vec.zero()
```

dham (approved):

dham (changes requested):
The test fails are caused by this line, so this is clearly not right.

ReubenHill:
> The test fails are caused by this line, so this is clearly not right.

Looks like it worked before switching to an architecture agnostic approach. I wonder what's different?

firedrake/interpolation.py
```diff
@@ -1429,4 +1429,14 @@ def multTranspose(self, source_vec, target_vec):
...
+            # We need to ensure the target vec is zeroed for SF Reduce to
+            # represent multTranspose in case the interpolation matrix is not
+            # square (in which case it will have columns which are zero). This
+            # happens when we interpolate from an input-ordering vertex-only
+            # mesh to an immersed vertex-only mesh where the input ordering
+            # contains points that are not in the immersed mesh. The resulting
+            # interpolation matrix will have columns of zeros for the points
+            # that are not in the immersed mesh. The adjoint interpolation
+            # matrix will then have rows of zeros for those points.
+            target_vec.zero()
```
ksagiyam, line 1441:
```suggestion
            target_vec.zeroEntries()
```

JDBetteridge (approved):
Approved in meeting

Issue #3150, opened 2023-10-11
BUG: `assign` of `subfunctions` do not work as expected before op2.par_loop when run in parallel


lrtfm:
**Describe the bug**
The data of functions from MixedFunctionSpace seems not update properly when using assign for its subfunctions and run in parallel.

**Steps to Reproduce**
Consider the following code:
```python
# debug_subfunctions.py
#
# Date: 2020-10-15
from firedrake import *
from firedrake.petsc import PETSc


def make_matrix(mesh):
    '''make a matrix according to the normal vertor of each nodes in the verctor space

    Give one scalar function space V1, one vector function space V2, and normal vector n defined on V2,
    return a matrix M in the mixed space V = V1*V2, such that the block V1*V1 is identity matrix,
    and the block V2*V2 looks like 

        0 0 0
        0 0 0
        a b c
             ...
             ...
             ...
                 0 0 0
                 0 0 0
                 d e f

    i.e., the third row of the block to each node is assigned with the values of the normal vector n on this node.

    '''

    V1 = FunctionSpace(mesh, "CG", 1)
    V2 = VectorFunctionSpace(mesh, "CG", 2)
    V = V1*V2

    # make a template matrix
    u, v = TrialFunction(V), TestFunction(V)
    a = inner(u[0], v[0])*dx + inner(u[1], v[1])*dx
    T = assemble(a).M

    # n[1] = [x[0], x[1], x[2]] be the normal vector
    n = Function(V)
    x = SpatialCoordinate(mesh)
    _n = Function(V2).interpolate(x/sqrt(x**2))

    # If we use assign or just copy the data, i.e., 
    #    n.subfunctions[1].assign(_n)
    # or
    #    n.subfunctions[1].dat.data[:] = _n.dat.data[:]
    # when run in parallel, the result is not correct.
    # For example, when run with `mpiexec -n 2`,
    # the last several rows are all zeros on the second process.
    # But interpolate works well
    #     n.subfunctions[1].interpolate(_n)
    n.subfunctions[1].assign(_n)

    kernel = op2.Kernel('''
    void set_matrix(double M[{N}][{N}], double *n){{
        int s = 0;
        for(int i = 0; i < 3; ++i){{
            M[i][i] = 1;
        }}
        for(int i = 3; i < {N}; ++i){{
            s = (i - 3)%3;
            if (s == 2) {{
                for (int j = 0; j < 3; j++) {{
                    M[i][i-s+j] = n[i-s+j];
[...]

PR #3168, opened 2023-10-17
Add test for MixedDat halo exchanges in a parloop

connorjward:
This is awful but PyOP2 does not have parallel testing so this test cannot live there.

Needs to land after https://github.com/OP2/PyOP2/pull/710.

[Related issue](https://github.com/firedrakeproject/firedrake/issues/3150)

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -79,7 +79,6 @@ jobs:
             --install icepack \
             --install irksome \
             --install femlium \
-            --package-branch pyop2 connorjward/fix-mixed-halos \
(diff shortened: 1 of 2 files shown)


ksagiyam (dismissed):

.github/workflows/build.yml
```diff
@@ -79,6 +79,7 @@ jobs:
             --install icepack \
             --install irksome \
             --install femlium \
+            --package-branch pyop2 connorjward/fix-mixed-halos \
```
JDBetteridge, line 82:
```suggestion
```

PR #3174, opened 2023-10-17
Ksagiyam/vom touch up (Part 1)

ksagiyam:
Touch up `VertexOnlyMesh` for checkpointing.

Make `VertexOnlyMesh` implementation conform more with regular mesh implementation by:

- enabling reordering,
- ordering DoFs using `_renumbering`, and
- refactoring.


ReubenHill:
Before this gets merged I'd appreciate you guiding me through the changes you've made. I've had a brief look for now!

firedrake/mesh.py
```diff
@@ -2921,13 +2913,13 @@ def VertexOnlyMesh(mesh, vertexcoords, missing_points_behaviour='error',
                 warn(str(error))
             else:
                 raise ValueError("missing_points_behaviour must be IGNORE, ERROR or WARN")
-
     name = name if name is not None else mesh.name + "_immersed_vom"
-
+    swarm.setName(_generate_default_mesh_topology_name(name))
+    original_swarm.setName(_generate_default_mesh_topology_name(name) + "_input_ordering")
```
ReubenHill, line 2923: Is this now consistent with DMPlex? Do they have the same name as their mesh?

firedrake/mesh.py
```diff
@@ -1575,10 +1568,7 @@ def __init__(self, swarm, parentmesh, name, reorder, use_cell_dm_marking, origin
 
         # Mark OP2 entities and derive the resulting Swarm numbering
         with PETSc.Log.Event("Mesh: numbering"):
-            if use_cell_dm_marking:
-                dmcommon.mark_entity_classes_using_cell_dm(self.topology_dm)
-            else:
-                dmcommon.mark_entity_classes(self.topology_dm)
+            dmcommon.mark_entity_classes(self.topology_dm)
```
ReubenHill, line 1580: So, was the pointSF set up correctly before your changes which construct it using the input ordering SF?

tests/vertexonly/test_vertex_only_mesh_generation.py
```diff
@@ -268,11 +268,11 @@ def test_generate_cell_midpoints(parentmesh, redundant):
...
             # I think the below is correct but it's not actually tested...
             expected = tuple(size*(layer-1) for size, layer in zip(parentmesh.cell_set.sizes, layers))
-            assert vm.cell_set.sizes == expected
+            assert vm.cell_set.sizes[1:] == expected[1:]
         else:
-            assert vm.cell_set.sizes == tuple(size*(layers-1) for size in parentmesh.cell_set.sizes)
+            assert vm.cell_set.sizes[1:] == tuple(size*(layers-1) for size in parentmesh.cell_set.sizes)[1:]
     else:
-        assert vm.cell_set.sizes == parentmesh.cell_set.sizes
+        assert vm.cell_set.sizes[1:] == parentmesh.cell_set.sizes[1:]
```
ReubenHill, line 275: What's going on here?

firedrake/mesh.py
ReubenHill: IIRC it was failing tests (though I see you've also modified the tests...)

firedrake/mesh.py
```diff
@@ -3162,57 +3180,40 @@ def _pic_swarm_in_mesh(
...
     if redundant and parent_mesh.comm.rank != 0:
         original_ordering_swarm_coords = np.empty(shape=(0, coords.shape[1]))
     else:
         original_ordering_swarm_coords = coords
+    # Set pointSF
+    owners = swarm.getField("DMSwarm_rank")
+    swarm.restoreField("DMSwarm_rank")
+    halo_indices, = np.where(owners != parent_mesh.comm.rank)
+    halo_indices = halo_indices.astype(IntType)
+    n = coords.shape[0]
```
ReubenHill, line 3193: Why does using the input coordinates at this point make sense? What if some of your coordinates are not found on ANY rank?

firedrake/mesh.py
```diff
@@ -3162,57 +3180,40 @@ def _pic_swarm_in_mesh(
...
+    a.fill(-1)
+    b = np.arange(m, dtype=utils.IntType)  # swarm local point numbers
+    b[halo_indices] = -1
+    unit = MPI._typedict[np.dtype(utils.IntType).char]
+    _input_ordering_sf.reduceBegin(unit, b, a, MPI.MAX)
+    _input_ordering_sf.reduceEnd(unit, b, a, MPI.MAX)
+    _input_ordering_sf.bcastBegin(unit, a, b, MPI.REPLACE)
+    _input_ordering_sf.bcastEnd(unit, a, b, MPI.REPLACE)
+    if np.any(b < 0):
+        raise RuntimeError("Unable to make swarm pointSF due to inconsistent data")
```
ReubenHill, line 3206: Clever! Can you explain what's going on here? I think I understand that the input ordering SF should has enough information to tell you the remote ranks and indices, but at first glance it's not quite obvious how you're getting 
them.

Using the input ordering SF to reduce the swarm local point numbers `b` into an array of integers (of length equal to the number of supplied coordinates that were rearranged or dropped)  `a`, with `MPI.MAX` does... what?

And then you broadcast in the other direction which somehow gives you all the local indices?

What does unit do?

firedrake/mesh.py
```diff
@@ -3768,7 +3768,7 @@ def _parent_mesh_embedding(
         # any rank which will have been set to comm.size + 1)
         owned_ranks_tosort[owned_ranks != parent_mesh.comm.rank] = parent_mesh.comm.size
         # now a sort by rank will give us the ordering we want
-        idxs = np.argsort(owned_ranks_tosort, kind='stable')
+        idxs = np.argsort(owned_ranks_tosort)
```
ReubenHill: Why have you removed this?

firedrake/mesh.py
```diff
@@ -1787,50 +1793,21 @@ def input_ordering_sf(self):
...
-        input_ranks_and_idxs[1::2] = input_indices
-        # local looks like the below, which means we can just pass in None
-        # local = numpy.arange(nleaves, dtype=IntType)
-        sf.setGraph(nroots, None, input_ranks_and_idxs)
-        return sf
+        e_p_map = self.cell_closure[:, -1]
+        ilocal = np.empty_like(e_p_map)
+        if len(e_p_map) > 0:
+            cStart = e_p_map.min()
+            ilocal[e_p_map - cStart] = np.arange(len(e_p_map))
```
ReubenHill, line 1817: How does this work? And aren't you remaking the same SF you used to make the swarm pointSF?

firedrake/mesh.py
```diff
@@ -1787,50 +1793,21 @@ def input_ordering_sf(self):
...
-        input_indices = input_index[idxs_to_include]
-        nleaves = len(input_ranks)
-        input_ranks_and_idxs = np.empty(2 * nleaves, dtype=IntType)
-        input_ranks_and_idxs[0::2] = input_ranks
-        input_ranks_and_idxs[1::2] = input_indices
-        # local looks like the below, which means we can just pass in None
-        # local = numpy.arange(nleaves, dtype=IntType)
-        sf.setGraph(nroots, None, input_ranks_and_idxs)
-        return sf
+        return self.input_ordering_sf.createEmbeddedLeafSF(np.arange(self.cell_set.size, dtype=IntType))
```
ReubenHill, line 1826: Nice! I hadn't realised you could do this.

ReubenHill: I think this relies on the assumption that halos are at the end of the leaves, am I correct? If so you should probably comment that here.

firedrake/mesh.py
```diff
@@ -2484,12 +2462,16 @@ def make_vom_from_vom_topology(topology, name):
...
-            function.Function(reference_coordinates_fs)
-        )
+        reference_coordinates_fs = functionspace.VectorFunctionSpace(topology, "DG", 0, dim=parent_tdim)
+        reference_coordinates_data = dmcommon.reordered_coords(topology.topology_dm, reference_coordinates_fs.dm.getDefaultSection(),
+                                                               (topology.num_vertices(), parent_tdim),
+                                                               reference_coord=True)
+        reference_coordinates = function.CoordinatelessFunction(reference_coordinates_fs,
+                                                                val=reference_coordinates_data,
+                                                                name=_generate_default_mesh_reference_coordinates_name(name))
+        vmesh._reference_coordinates = reference_coordinates
```
ReubenHill, line 2488: Why are you creating a `CoordinatelessFunction` here? Also if you are going to attach it to the vertex-only mesh, you should probably rename it to avoid confusion with the `.reference_coordinates` `Function` property.

firedrake/mesh.py
```diff
@@ -3759,24 +3741,6 @@ def _parent_mesh_embedding(
...
-        # now a sort by rank will give us the ordering we want
-        idxs = np.argsort(owned_ranks_tosort)
-        coords_embedded = coords_embedded[idxs]
-        global_idxs = global_idxs[idxs]
-        reference_coords = reference_coords[idxs]
-        parent_cell_nums = parent_cell_nums[idxs]
-        owned_ranks = owned_ranks[idxs]
-        input_ranks = input_ranks[idxs]
-        input_coords_idxs = input_coords_idxs[idxs]
-
```
ReubenHill: What has replaced this?

ReubenHill: If it's `dmcommon.plex_renumbering`, how does setting up the `pointSF` of the swarm play nicely with this since you do it afterwards?

firedrake/mesh.py
ksagiyam: `DMSwarm` and `DMPlex` are both `DM`, `DM` is `PetscObject`, and `PetscObject` carries `setName()` method.
Could you clarify your question?

firedrake/mesh.py
ksagiyam: The pointSF you had was correct.

I was actually careless here. I thought `owned` label was not necessary for vom, so we could just use `mark_entity_classes`, which labels `ghost` entities correctly. But it was wrong, so I put back `mark_entity_using_cell_dm` (I rewrote, so please have a look). I also reverted the change that I made in the tests.

Initially, we decided to have vom inherit pyop2 labels from the parent as it was natural, but if I think about how `parloop` works in two phases (`core` loop + `owned` loop, https://github.com/OP2/PyOP2/blob/d017d594e0bda694a71c4180fbfeae1cba473e93/pyop2/parloop.py#L216), inheriting the labels actually seems necessary, and we should always do so.
`input_ordering` currently does not, so I just kept it as is for now.

tests/vertexonly/test_vertex_only_mesh_generation.py
ksagiyam: Reverted as explained in the above.

firedrake/mesh.py
ksagiyam: `n` merely defines the local size of an array, i.e., `a`, that works as "broker". The set of indices of `a` is the target of `inputindex`; see `_make_input_ordering_sf`.
If some `index` of `a` corresponds to a missing point, `a[index]` is not updated when we `reduce` and it does not update any leaf data, i.e., `b`, when we `bcast`.

firedrake/mesh.py
ksagiyam: Yes. All points in `b` are leaves. Then, `b[halo_indices] = -1`, `a.fill(-1)`, and `MPI.MAX` ensure that `a` has the swarm local point numbers of the owners after `reduce`. `a[j] = -1` if `j` corresponds to a missing point. Then, `broadcast` updates `b[halo_indices]` (it also updates b[~halo_indices]`, not changing any values there).

firedrake/mesh.py
ksagiyam: Note that this entire block has been removed.

The way `pointSF` was initially constructed depended on the assumption that swarm points were ordered according to the global coordinate indices, which was achieved with `stable`. Given that `plex` points can come in arbitrary order (at least stratum-wise), such underlying assumption should not be necessary.

firedrake/mesh.py
ksagiyam: `e_p_map` maps Firedrake cell numbers to swarm point numbers. `ilocal` is the inverse of this map, used in `_make_input_ordering_sf`.
The SF constructed here is different from the SF used to make pointSF: leaves of the former are Firedrake cell numbers and those of the latter are swarm point numbers.

firedrake/mesh.py
ksagiyam: Yes. Done!

firedrake/mesh.py
ksagiyam: Just to be consistent with how we make coordinates in regular mesh.

Did we have `.reference_coordinates` attribute somewhere else?

firedrake/mesh.py
ksagiyam: Yes, we now use the existing renumbering infrastructure. `pointSF` is more fundamental. `pointSF` defines the association: `swarm point <- (rank, swarm point)`. `_renumbering` depends on `pointSF` and is used to define Firedrake cell numbers, DoF orderings, etc.

ReubenHill: The the local ordering of the `pointSF` has presumably changed then, since you don't move halos to the end?

firedrake/mesh.py
ReubenHill: Upon construction of a normal mesh, you can supply a name string. Do you use this string (or a modification of it for, say, extruded meshes) to name the `DMPlex` `PetscObject` associated with that mesh?

firedrake/mesh.py
ReubenHill: `input_ordering` also has no halos so everything should be `core`

ReubenHill: Do you recall why the pointSF of the VOM does not correctly label `owned` cells?

firedrake/mesh.py
```diff
@@ -3163,57 +3245,40 @@ def _pic_swarm_in_mesh(
...
         original_ordering_swarm_coords = np.empty(shape=(0, coords.shape[1]))
     else:
         original_ordering_swarm_coords = coords
+    # Set pointSF
+    owners = swarm.getField("DMSwarm_rank")
+    halo_indices, = np.where(owners != parent_mesh.comm.rank)
+    halo_indices = halo_indices.astype(IntType)
+    n = coords.shape[0]
+    m = owners.shape[0]
+    _input_ordering_sf = VertexOnlyMeshTopology._make_input_ordering_sf(swarm, n, None)
```
ReubenHill, line 3261: It needs to be made clear that this is an input ordering SF which maps between swarm points, rather than between firedrake cell numbers.

firedrake/mesh.py
ReubenHill: Please make this clear in a comment

firedrake/mesh.py
ReubenHill: I think some comments and/or obvious variable naming are needed here then. For example, is `cStart` a cell number or an index? What do `e` and `p` stand for in `e_p_map`?

firedrake/mesh.py
ReubenHill: > Just to be consistent with how we make coordinates in regular mesh.

But `reference_coordinates` are more akin to an attribute of each vertex (i.e. a normal `Function`), rather than a coordinates field.

> Did we have .reference_coordinates attribute somewhere else?

I'm just complaining that it's not descriptive of what it is. If you think it should held onto, then call it `._reference_coordinates_cf` or something.

firedrake/mesh.py
ReubenHill: Please explain the algorithm in comments, as you have here. 

I'm also keen on more descriptive names which fit with the rest of the function. If I've understood you correctly I think `b` is `owner_local_idx`. I'm not sure of what to call `a` though - is it `remote_local_idx`?

firedrake/cython/dmcommon.pyx
```diff
@@ -1763,80 +1761,48 @@ def mark_entity_classes_using_cell_dm(PETSc.DM swarm):
...
+        CHKERR(DMGetLabel(swarm.dm, op2class, &swarm_labels[ilabel]))
+    CHKERR(DMSwarmGetField(swarm.dm, b"DMSwarm_cellid", &blocksize, &ctype, <void**>&swarmParentCells))
+    assert ctype == PETSC_INT
+    assert blocksize == 1
+    CHKERR(DMSwarmGetLocalSize(swarm.dm, &nswarmCells))
+    for swarmCell in range(nswarmCells):
+        plex_cell_class = plex_cell_classes[swarmParentCells[swarmCell] - cStart]
+        CHKERR(DMLabelSetValue(swarm_labels[plex_cell_class], swarmCell, label_value))
+    CHKERR(DMSwarmRestoreField(swarm.dm, b"DMSwarm_cellid", &blocksize, &ctype, <void**>&swarmParentCells))
+    CHKERR(PetscFree(plex_cell_classes))
```
ReubenHill, line 1805: So why did you rewrite this? What advantages does this rewrite have over the previous routine?

firedrake/mesh.py
ksagiyam: Yes. A `DMPlex` has the same name as the `MeshTopology`.

firedrake/mesh.py
ksagiyam: Added.

firedrake/mesh.py
ksagiyam: Added. With added comment, I do not think we need longer variable names. Their life time is only 10 lines.

firedrake/mesh.py
ksagiyam: Added.

firedrake/mesh.py
ksagiyam: Removed `_reference_coordinates`.

firedrake/mesh.py
ksagiyam: Yes. That is what the third argument of `_make_input_ordering_sf` is.

firedrake/cython/dmcommon.pyx
ksagiyam: It is more concise. It also avoid copying arrays (`.getIndices()`). This function is only called once, so efficiency does not really matter, but this is more consistent with other Cython functions in dmcommon.

firedrake/mesh.py
ksagiyam: Added comment.

firedrake/mesh.py
ksagiyam: Given a non-ghost cell `c`, if any point in the closure of `c` is ghost, we label all points in the closure as `owned` (if not ghost). That is why there is no `owned` cells (vertices) in VOMs, at least in this definition.

ksagiyam: That is not how we define `owned`. The idea I think is that when we do `Parloop`, we can kick off iteration over `core` cells without waiting for the halo exchange to complete, while, for iteration over `owned` cells, we have to make sure that the halo data are up-to-date.
However, as far as I understand, this way of handling the two-phase computation (`core` loop and `owned` loop) by statically marking `core` and `owned` entities does not directly work if multiple meshes are involved in the `Parloop`, and we probably need a further discussion when we do, e.g., couplings.
VOMs can be examples where multiple meshes are involved, but, if a VOM and its parent are the only ones involved, thanks to the nature of VOM, we can use existing infrastructure by simply letting VOM inherit the pyops labels from the parent; when we do interpolation onto a VOM, we need to make sure that all parent function data are up-to-date when computation is performed in `Parloop`, and inheriting `owned` label I think allows this.

ksagiyam:
@ReubenHill Do you have any more things to check regarding this PR?

firedrake/mesh.py
```diff
@@ -1517,77 +1554,82 @@ class VertexOnlyMeshTopology(AbstractMeshTopology):
...
+        pass
 
-        # Mark OP2 entities and derive the resulting Swarm numbering
-        with PETSc.Log.Event("Mesh: numbering"):
-            if use_cell_dm_marking:
-                dmcommon.mark_entity_classes_using_cell_dm(self.topology_dm)
-            else:
-                dmcommon.mark_entity_classes(self.topology_dm)
+    def _mark_entity_classes(self):
+        # TODO: mark_entity_classes_using_cell_dm should always be used.
```
dham, line 1600:
```suggestion
```

ReubenHill (changes requested):
I'm mostly happy. It's nice to see that the existing renumbering systems can be used here, and the refactoring looks good.

Most of what I ask for is more clarity where it's not clear to me what you're doing. In particular the setting of the DMSwarm pointSF. I know I'm a fan of lots of explanation as well as descriptive variable names, but if I can't quite follow what's going on I imagine others will find themselves in the same situation.

Also I recall you saying that you'd changed `VertexOnlyMeshTopology` to be subclassed from `AbstractMeshTopology`, but it seems that was already the case?

firedrake/cython/dmcommon.pyx
```diff
@@ -1999,80 +1999,48 @@ def mark_entity_classes_using_cell_dm(PETSc.DM swarm):
...
-        elif label == 3:
-            CHKERR(DMLabelSetValue(swarm_label_ghost, label_idx, 1))
-        else:
-            raise RuntimeError("Unknown label value")
-    return
+    get_height_stratum(plex.dm, 0, &cStart, &cEnd)
+    CHKERR(PetscMalloc1(cEnd - cStart, &plex_cell_classes))
+    for c in range(cStart, cEnd):
+        plex_cell_classes[c - cStart] = -1
+    for ilabel, op2class in enumerate([b"pyop2_core", b"pyop2_owned", b"pyop2_ghost"]):
```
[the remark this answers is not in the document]
ksagiyam: Yes. Comments added.

firedrake/cython/dmcommon.pyx
```diff
@@ -1999,80 +1999,48 @@ def mark_entity_classes_using_cell_dm(PETSc.DM swarm):
...
+        CHKERR(DMGetLabel(plex.dm, op2class, &plex_label))
+        CHKERR(DMLabelGetStratumSize(plex_label, label_value, &class_size))
```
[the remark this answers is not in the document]
ksagiyam: Done.

firedrake/cython/dmcommon.pyx
```diff
@@ -1999,80 +1999,48 @@ def mark_entity_classes_using_cell_dm(PETSc.DM swarm):
...
+            raise RuntimeError("Cell point %d in the parent plex does not belong to any pyop2 class" % c)
+    for ilabel, op2class in enumerate([b"pyop2_core", b"pyop2_owned", b"pyop2_ghost"]):
+        CHKERR(DMCreateLabel(swarm.dm, op2class))
+        CHKERR(DMGetLabel(swarm.dm, op2class, &swarm_labels[ilabel]))
+    CHKERR(DMSwarmGetField(swarm.dm, b"DMSwarm_cellid", &blocksize, &ctype, <void**>&swarmParentCells))
+    assert ctype == PETSC_INT
+    assert blocksize == 1
+    CHKERR(DMSwarmGetLocalSize(swarm.dm, &nswarmCells))
+    for swarmCell in range(nswarmCells):
+        plex_cell_class = plex_cell_classes[swarmParentCells[swarmCell] - cStart]
```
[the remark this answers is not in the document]
ksagiyam: The standard plex cell point ordering. Added a comment.

firedrake/mesh.py
```diff
@@ -1758,7 +1794,32 @@ def input_ordering(self):
...
+        input_ranks = swarm.getField("inputrank")
+        input_indices = swarm.getField("inputindex")
+        nleaves = len(input_ranks)
+        if ilocal is not None and nleaves != len(ilocal):
+            raise RuntimeError(f"Mismatching leaves: nleaves {nleaves} != len(ilocal) {len(ilocal)}")
+        input_ranks_and_idxs = np.empty(2 * nleaves, dtype=IntType)
+        input_ranks_and_idxs[0::2] = input_ranks
+        input_ranks_and_idxs[1::2] = input_indices
+        swarm.restoreField("inputrank")
+        swarm.restoreField("inputindex")
```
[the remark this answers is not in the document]
ksagiyam: Done.

firedrake/mesh.py
```diff
@@ -1769,50 +1830,22 @@ def input_ordering_sf(self):
...
-        input_ranks_and_idxs[0::2] = input_ranks
-        input_ranks_and_idxs[1::2] = input_indices
-        # local looks like the below, which means we can just pass in None
-        # local = numpy.arange(nleaves, dtype=IntType)
-        sf.setGraph(nroots, None, input_ranks_and_idxs)
-        return sf
+        e_p_map = self.cell_closure[:, -1]  # cell-entity - swarm-point map
+        ilocal = np.empty_like(e_p_map)
+        if len(e_p_map) > 0:
+            cStart = e_p_map.min()  # smallest swarm point number
```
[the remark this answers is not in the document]
ksagiyam: `cStart` is the smallest swarm point. As noted somewhere, swarm points lie in `[cStart, cEnd)`. `cStart` and `cEnd` are the standard variable names in Firedrake as well as in PETSc.

firedrake/mesh.py
```diff
@@ -1517,77 +1554,82 @@ class VertexOnlyMeshTopology(AbstractMeshTopology):
...
-        r"The PETSc DM representation of the mesh topology."
+    def __init__(self, swarm, parentmesh, name, reorder, original_swarm=None, perm_is=None, distribution_name=None, permutation_name=None):
+        """Initialise a mesh topology.
+
+        Parameters
+        ----------
+        swarm : PETSc.DMSwarm
+            `PETSc.DMSwarm` representing Particle In Cell (PIC) vertices
+            immersed within a `PETSc.DMPlex` stored in the ``parentmesh``.
+        parentmesh : PETSc.DMPlex
```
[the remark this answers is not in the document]
ksagiyam: Fixed.

firedrake/mesh.py
```diff
@@ -1517,77 +1554,82 @@ class VertexOnlyMeshTopology(AbstractMeshTopology):
...
-        # Mark OP2 entities and derive the resulting Swarm numbering
-        with PETSc.Log.Event("Mesh: numbering"):
-            if use_cell_dm_marking:
-                dmcommon.mark_entity_classes_using_cell_dm(self.topology_dm)
-            else:
-                dmcommon.mark_entity_classes(self.topology_dm)
+    def _mark_entity_classes(self):
+        # TODO: mark_entity_classes_using_cell_dm should always be used.
+        if isinstance(self._parent_mesh, MeshTopology):
+            dmcommon.mark_entity_classes_using_cell_dm(self.topology_dm)
```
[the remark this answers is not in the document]
ksagiyam: Please see the change. Behaviour having to change is due to that the nature of `input_ordering` is different from other voms.

firedrake/mesh.py
ksagiyam: Done.

firedrake/mesh.py
ksagiyam: `a->local_points_reduced`
`b->local_points`
What do you find difficult to understand? The best way to convince yourself would be to run a simple parallel example and check data line by line.

firedrake/mesh.py
```diff
@@ -487,35 +499,105 @@ class AbstractMeshTopology(object, metaclass=abc.ABCMeta):
...
+            # there is no gurantee that it really does or it continues to do so
+            # there are lots of parameters that can change distributions.
+            # Thus, when using CheckpointFile, it is recommended that the user set
+            # distribution_name explicitly.
+            # Mark OP2 entities and derive the resulting Plex renumbering
+            with PETSc.Log.Event("Mesh: numbering"):
+                self._mark_entity_classes()
+                self._entity_classes = dmcommon.get_entity_classes(self.topology_dm).astype(int)
+                if perm_is:
+                    self._plex_renumbering = perm_is
```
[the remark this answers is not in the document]
ksagiyam: Done.

firedrake/mesh.py
```diff
@@ -1517,77 +1554,82 @@ class VertexOnlyMeshTopology(AbstractMeshTopology):
...
+        swarm : PETSc.DMSwarm
+            `PETSc.DMSwarm` representing Particle In Cell (PIC) vertices
+            immersed within a `PETSc.DMPlex` stored in the ``parentmesh``.
+        parentmesh : PETSc.DMPlex
+            Mesh within which the vertex-only mesh topology is immersed.
+        name : str
+            Name of the mesh topology.
+        reorder : bool
+            Whether to reorder the mesh entities.
+        original_swarm : PETSc.DMSwarm
```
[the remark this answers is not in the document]
ksagiyam: Done.

firedrake/cython/dmcommon.pyx
ksagiyam: If you are interested, here's the performance difference (`temp.py`):
```
from firedrake import *
import numpy as np
import time


mesh = UnitSquareMesh(200, 200)
vertexcoords = np.random.random((2000, 2))
vm = VertexOnlyMesh(mesh, vertexcoords, missing_points_behaviour=None)
swarm = vm.topology_dm

vm.comm.barrier()

start = time.time()
dmcommon.mark_entity_classes_using_cell_dm_old(swarm)
end = time.time()
print(f"original (rank = {vm.comm.rank}): {end - start} [s]", flush=True)

vm.comm.barrier()

start = time.time()
dmcommon.mark_entity_classes_using_cell_dm(swarm)
end = time.time()
print(f"updated  (rank = {vm.comm.rank}): {end - start} [s]", flush=True)
```
```
mpiexec -n 4 python temp.py
```
```
original (rank = 0): 0.000461578369140625 [s]
original (rank = 1): 0.0004367828369140625 [s]
original (rank = 2): 0.00045990943908691406 [s]
original (rank = 3): 0.0004508495330810547 [s]
updated  (rank = 0): 0.00015306472778320312 [s]
updated  (rank = 1): 0.0001506805419921875 [s]
updated  (rank = 2): 0.00015020370483398438 [s]
updated  (rank = 3): 0.00015211105346679688 [s]
```
`mark_entity_classes_using_cell_dm_old` is the original implementation.

firedrake/mesh.py
```diff
@@ -3167,57 +3249,49 @@ def _pic_swarm_in_mesh(
...
     if redundant and parent_mesh.comm.rank != 0:
         original_ordering_swarm_coords = np.empty(shape=(0, coords.shape[1]))
     else:
         original_ordering_swarm_coords = coords
+    # Set pointSF
+    # In the below, n merely defines the local size of an array, local_points_reduced,
+    # that works as "broker". The set of indices of local_points_reduced is the target of
+    # inputindex; see _make_input_ordering_sf. All points in local_points are leaves.
+    # Then, local_points[halo_indices] = -1, local_points_reduced.fill(-1), and MPI.MAX ensure that local_points_reduced has
+    # the swarm local point numbers of the owners after reduce. local_points_reduced[j] = -1
```
ReubenHill, line 3264:
```suggestion
    # the swarm local point indices in the owning rank after reduce. local_points_reduced[j] = -1
```

firedrake/mesh.py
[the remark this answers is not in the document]
ReubenHill: Yes, you have to wrap it in a `np.copy` as I do [here](https://github.com/firedrakeproject/firedrake/blob/aefe44e9e1f099e60961d6ea04367dbde8788c2c/firedrake/mesh.py#L1674) for example.
So I think either do the copy or note the error then raise it after restoring the field.

firedrake/mesh.py
```diff
@@ -1758,7 +1794,32 @@ def input_ordering(self):
...
+    @staticmethod
+    def _make_input_ordering_sf(swarm, nroots, ilocal):
+        # ilocal = None -> leaves are swarm points [0, 1, 2, ...).
+        # ilocal can also be Firedrake cell numbers.
+        sf = PETSc.SF().create(comm=swarm.comm)
+        input_ranks = swarm.getField("inputrank")
+        input_indices = swarm.getField("inputindex")
+        nleaves = len(input_ranks)
+        if ilocal is not None and nleaves != len(ilocal):
+            raise RuntimeError(f"Mismatching leaves: nleaves {nleaves} != len(ilocal) {len(ilocal)}")
```
ReubenHill, line 1815:
```suggestion
            swarm.restoreField("inputrank")
            swarm.restoreField("inputindex")
            raise RuntimeError(f"Mismatching leaves: nleaves {nleaves} != len(ilocal) {len(ilocal)}")
```

firedrake/mesh.py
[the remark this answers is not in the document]
ReubenHill: (see my incoming review suggestion)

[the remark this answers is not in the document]
ReubenHill: (see my incoming review suggestion)

firedrake/mesh.py
```diff
@@ -1517,77 +1554,82 @@ class VertexOnlyMeshTopology(AbstractMeshTopology):
...
-        r"The PETSc DM representation of the mesh topology."
+    def __init__(self, swarm, parentmesh, name, reorder, input_ordering_swarm=None, perm_is=None, distribution_name=None, permutation_name=None):
+        """Initialise a mesh topology.
+
+        Parameters
+        ----------
+        swarm : PETSc.DMSwarm
+            `PETSc.DMSwarm` representing Particle In Cell (PIC) vertices
+            immersed within a `PETSc.DM` stored in the ``parentmesh``.
+        parentmesh : PETSc.DM
```
ReubenHill, line 1565:
```suggestion
        parentmesh : AbstractMeshTopology
```
I think

firedrake/mesh.py
```diff
@@ -1517,77 +1554,82 @@ class VertexOnlyMeshTopology(AbstractMeshTopology):
...
-            if use_cell_dm_marking:
-                dmcommon.mark_entity_classes_using_cell_dm(self.topology_dm)
-            else:
-                dmcommon.mark_entity_classes(self.topology_dm)
+    def _mark_entity_classes(self):
+        if self.input_ordering_swarm:
+            assert isinstance(self._parent_mesh, MeshTopology)
+            dmcommon.mark_entity_classes_using_cell_dm(self.topology_dm)
+        else:
+            dmcommon.mark_entity_classes(self.topology_dm)
```
ReubenHill, line 1604:
```suggestion
            # Have an input-ordering vertex-only mesh. These should mark 
            # all entities as pyop2 core, which mark_entity_classes will do.
            assert isinstance(self._parent_mesh, VertexOnlyMeshTopology)
            dmcommon.mark_entity_classes(self.topology_dm)
```
I think here. It makes it clear what's going on to make it clear what's going on

firedrake/mesh.py
ReubenHill: `parentmesh` should probably also be renamed `parentmeshtopology`

firedrake/mesh.py
```diff
@@ -1517,77 +1554,82 @@ class VertexOnlyMeshTopology(AbstractMeshTopology):
...
+    def __init__(self, swarm, parentmesh, name, reorder, input_ordering_swarm=None, perm_is=None, distribution_name=None, permutation_name=None):
+        """Initialise a mesh topology.
+
+        Parameters
+        ----------
+        swarm : PETSc.DMSwarm
+            `PETSc.DMSwarm` representing Particle In Cell (PIC) vertices
+            immersed within a `PETSc.DM` stored in the ``parentmesh``.
+        parentmesh : PETSc.DM
+            Mesh within which the vertex-only mesh topology is immersed.
```
ReubenHill, line 1566:
```suggestion
            Mesh topology within which the vertex-only mesh topology is immersed.
```

firedrake/mesh.py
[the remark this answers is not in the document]
ReubenHill: Ah, I see what's going on. See my suggested change (incoming with review).

firedrake/mesh.py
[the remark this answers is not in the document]
ReubenHill: see incoming review suggestion

ReubenHill (changes requested):
Almost there, just a few little changes requested\suggested. 

Just to check, the renaming of `_plex_renumbering` to `_dm_renumbering` won't mess up existing checkpoints right?

firedrake/mesh.py
ksagiyam: Yes, but they are called plex/swarm point numbers.

ReubenHill: Are you sure? You're telling me that the number of a swarm entity is also its offset in memory?

firedrake/mesh.py
ksagiyam: I do not think we need to rename. If this class is `VertexOnlyMeshTopology`, the parent ought to be topology.

Either way, this is the term used in current `master`, so we should discuss this elsewhere if we need to.

firedrake/mesh.py
ksagiyam: Yes.

ksagiyam: That is why `pointSF` (which virtually represent point number - point number map) works with various arrays.

firedrake/mesh.py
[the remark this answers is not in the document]
ksagiyam: Done.

firedrake/mesh.py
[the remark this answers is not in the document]
ksagiyam: Done.

ksagiyam:
Addressed all comments, I think.

`_plex_renumbering` -> `_dm_renumbering` should not affect checkpointing.

ReubenHill (approved):

ReubenHill:
Thanks Koki!

ksagiyam:
Thanks!

connorjward:
@ReubenHill @ksagiyam this won't merge because of an unresolved conversation. I can't seem to find it to click resolved. Perhaps force merge once tests pass?

PR #3178, opened 2023-10-18
Calling a Cofunction to compute duality pairing

pbrubeck:
# Description
Implement `__call__` method on `Cofunction` to compute the duality paring against a `Function`. Also added tests.

### DIFF
--- a/firedrake/cofunction.py
+++ b/firedrake/cofunction.py
@@ -79,6 +79,18 @@ class Cofunction(ufl.Cofunction, FunctionMixin):
         if hasattr(self, "_comm"):
             mpi.decref(self._comm)
 
+    def __call__(self, primal):
+        r"""Compute the duality pairing, this is to evaluate this
+        :class:`.Cofunction` at a :class:`.Function`, equivalent to taking the
+        dot product of the vectors of coefficients of the dual and primal
+        arguments.
+
+        :arg primal: the :class:`.Function` to be paired with this :class:`.Cofunction`.
+
+        :returns: the scalar result of the duality pairing.
+        """
+        return firedrake.assemble(ufl.action(self, primal))
+
     def copy(self, deepcopy=True):
         r"""Return a copy of this :class:`firedrake.function.CoordinatelessFunction`.
 
--- a/firedrake/cofunction.py
+++ b/firedrake/cofunction.py
@@ -80,16 +80,16 @@ class Cofunction(ufl.Cofunction, FunctionMixin):
             mpi.decref(self._comm)
 
     def __call__(self, primal):
-        r"""Compute the duality pairing, this is to evaluate this
+        r"""Return the symbolic duality pairing, this is to evaluate this
         :class:`.Cofunction` at a :class:`.Function`, equivalent to taking the
         dot product of the vectors of coefficients of the dual and primal
         arguments.
 
         :arg primal: the :class:`.Function` to be paired with this :class:`.Cofunction`.
 
-        :returns: the scalar result of the duality pairing.
+        :returns: the symbolic duality pairing.
         """
-        return firedrake.assemble(ufl.action(self, primal))
+        return ufl.action(self, primal)
 
     def copy(self, deepcopy=True):
         r"""Return a copy of this :class:`firedrake.function.CoordinatelessFunction`.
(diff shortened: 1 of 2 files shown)


firedrake/cofunction.py
```diff
@@ -79,6 +79,18 @@ def __del__(self):
...
+        r"""Compute the duality pairing, this is to evaluate this
+        :class:`.Cofunction` at a :class:`.Function`, equivalent to taking the
+        dot product of the vectors of coefficients of the dual and primal
+        arguments.
+
+        :arg primal: the :class:`.Function` to be paired with this :class:`.Cofunction`.
+
+        :returns: the scalar result of the duality pairing.
+        """
+        return firedrake.assemble(ufl.action(self, primal))
```
ReubenHill, line 92: To be more general, I think this should perform symbolic dual-evaluation interpolation #2297 . Does `ufl.action` account for the inner product in your Riesz representer?

firedrake/cofunction.py
```diff
@@ -79,6 +79,18 @@ def __del__(self):
         if hasattr(self, "_comm"):
             mpi.decref(self._comm)
 
+    def __call__(self, primal):
+        r"""Compute the duality pairing, this is to evaluate this
+        :class:`.Cofunction` at a :class:`.Function`, equivalent to taking the
+        dot product of the vectors of coefficients of the dual and primal
+        arguments.
```
ReubenHill, line 86: I think yhis is only true for euclidian vectors in R^n, not for Functions and Cofunctions in general. See my question about `ufl.action` below

ReubenHill: No that's wrong - you've got your cofunction using that inner product already

firedrake/cofunction.py
ReubenHill: Hmm, I think I've gotten confused. Please ignore.

nbouziani, line 92: I feel like just returning the symbolic object would facilitate things like differentiation or dual evaluation, especially with the upcoming `Interpolate`. For example, the dual evaluation when primal lives in a different space would just write:

```
c(Interpolate(f, V2))
```

where `c` is a `Cofunction` in `V2*` and f a `Function` in `V1`. Likewise, differentiation could be done seamlessly while accounting for the `Interpolate`.

I guess we could have a keyword argument for assembly, but then I don't think there is a real gain in writing something like:

```
c(f, assemble=True)
```

instead of

```
assemble(c(f))
```

firedrake/cofunction.py
```diff
@@ -79,6 +79,21 @@ def __del__(self):
...
+        :class:`.Cofunction` at an expression in the primal space, equivalent to
+        taking the dot product of the vectors of coefficients of the dual and
+        primal arguments.
+
+        :arg expr: a UFL expression to be paired with this :class:`.Cofunction`.
+
+        :returns: the symbolic duality pairing.
+        """
+        try:
+            return ufl.action(self, expr)
```
nbouziani, line 93: I don't think that the assumption that a `TypeError` would be a consequence of a wrong function space is safe. Think about the case when one passes in the wrong type for `expr`.

firedrake/cofunction.py
```diff
@@ -79,6 +79,21 @@ def __del__(self):
...
+        except TypeError:
+            return ufl.action(self, ufl.Interpolate(expr, self.function_space().dual()))
```
nbouziani, line 95: A couple of comments:
- I think we should separate the concern of the function space from the actual duality pairing. I would advocate for having `expr` as a `Cofunction` on the right space, and then `__call__` should just return the action. If one wants to interpolate it should be done on the outside, e.g.:

```
c(Interpolate(f, V2))
```

where `c` lives in `V2*` and `f` in `V1`. This would be needed anyway if `f` is not a `Function` but an expression.

- `ufl.Interpolate` is also not usable at the moment since that just handles the symbolic part (e.g. differentiation), not the actual assembly which will be required whenever you need to evaluate that duality pairing. This will come with #2297.

nbouziani (changes requested):

dham:
Do this in UFL and have it only call action. Should probably be done on BaseForm.

nbouziani:
Corresponding UFL PR: https://github.com/FEniCS/ufl/pull/232

PR #3186, opened 2023-10-24
`WithGeometryBase.__dir__` fix

jrmaddison:
# Description

Bugfix to `dir(FunctionSpace(...))`.

Fixes infinite recursion caused by the use of `super(type(self), self)` in a base class.


connorjward (approved):
Thanks!

PR #3191, opened 2023-10-27
Zenodo search API no longer produces unique results

JDBetteridge:
# Description

Update `firedrake-install` to no longer use search feature of Zenodo RESTful API. Zenodo have updated to use elastic search, but this means that searching for a DOI (supposedly unique) does not produce a unique result. [Zenodo API docs ](https://developers.zenodo.org/#retrieve) suggest that this method of querying is a suitable replacement.

### DIFF
--- a/.github/workflows/zenodo-canary.yml
+++ b/.github/workflows/zenodo-canary.yml
@@ -6,7 +6,7 @@ on:
   push:
     branches:
       - master
-  pull_request_target:
+  pull_request:
 
 jobs:
   zenodo_canary:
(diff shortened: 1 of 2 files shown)


JDBetteridge:
Note that `pull_request` is passing because it runs on this branch and `pull_request_target` is failing because it runs on the master branch. I'm not sure of the motivation for having originally had `pull_request_target` in the workflow, but if you have strong opinions, feel free to make the github "suggested change" to put it back as it was.

PR #3215, opened 2023-11-09
Emmarothwell1/restricted _function_space: Implement a RestrictedFunctionSpace class. 

emmarothwell1:
# Implement a RestrictedFunctionSpace class.
(Part of an M4R project)
This inherits from FunctionSpace, with the main difference being how boundary conditions are treated.  
This will hopefully be used in cases where users have boundary conditions they want to apply.


firedrake/functionspaceimpl.py
```diff
@@ -1170,6 +1170,108 @@ def local_to_global_map(self, bcs, lgmap=None):
...
+        self.function_space = function_space
+        boundary_set = set()
+        self.bcs = bcs
+        for bc in bcs:
+            boundary_set = boundary_set.union(set(bc.sub_domain))
+        self.boundary_set = frozenset(boundary_set)
+        self.name = name or (function_space.name + "_"  
+                             + "_".join(sorted(
+                                [str(i) for i in self.boundary_set])))
+        self._shared_data = sdata = get_shared_data(function_space._mesh, function_space.ufl_element(), self.boundary_set)
```
ksagiyam, line 1188: `super().__init__()` sets `self.node_set` etc. using `shared_data` of the parent function space (so, in your current example, `V_res.node_set is V.node_set`).
We should probably define `_init_with_shared_data` method and do something like:
```
class FunctionSpace(...):
    sdata = get_shared_data(mesh, elem)
    self._init_with_shared_data(mesh, elem, name, sdata)

class RestrictedFunctionSpace(...):
    boundary_set = ...
    sdata = get_shared_data(mesh, elem, boundary_set=boundary_set)
    self._init_with_shared_data(mesh, elem, name, sdata)
```

ksagiyam:
In serial `lgmap` is not aware of the changes you have made in `dmcommon.make_global_numbering`:
https://github.com/OP2/PyOP2/blob/d017d594e0bda694a71c4180fbfeae1cba473e93/pyop2/types/dataset.py#L122.
I think we can just delete `self.comm.size == 1` case.

firedrake/functionspacedata.py
```diff
@@ -416,9 +420,10 @@ def __init__(self, mesh, ufl_element):
         except NotImplementedError:
             entity_permutations = None
 
+       
         # Create the PetscSection mapping topological entities to functionspace nodes
         # For non-scalar valued function spaces, there are multiple dofs per node.
-        key = (nodes_per_entity, real_tensorproduct)
+        key = (nodes_per_entity, real_tensorproduct, boundary_set)
```
ksagiyam, line 426: `self.map_cache` must also be keyed by `boundary_set`.

ksagiyam:
With the above changes, I was able to get a 4 x 4 matrix with nonzeros only on the top-left 2 x 2 submatrix; this indicates that your lgmap is correct. The problem now reduces to getting a matrix of the right shape.
The allocation of a matrix is done in `assemble.py` using `op2.Sparsity`. The actual assembly is done in the same file using `op2.Parloop`.
When allocating, the size of the matrix is set here [https://github.com/OP2/PyOP2/blob/d017d594e0bda694a71c4180fbfeae1cba473e93/pyop2/sparsity.pyx#L140](https://github.com/OP2/PyOP2/blob/d017d594e0bda694a71c4180fbfeae1cba473e93/pyop2/sparsity.pyx#L136), where `rset.size` is `rset.set.size`, which is the size of `core + owned` DoFs of the given rank. Here, `rset.set` is `V.node_set`. As we now want the size of the reduced matrix here, we probably want `V.node_set` to store the size of the "constrained owned DoFs" as, say, `V.node_set.constrained_size`, when we make `node_set` here https://github.com/firedrakeproject/firedrake/blob/9c50131b511845db99542a7f6e1bebcbf14bcd73/firedrake/functionspacedata.py#L107.
The computation of the size of the "constrained owned DoFs" should probably be done when we compute renumbering to push "constrained owned DoFs" to the end of `core + owned` block.

We currently only renumber (reorder) entities so that a local vector sees core, owned, and ghost DoFs in this order; we compute renumbering using `dmcommon.plex_renumbering` https://github.com/firedrakeproject/firedrake/blob/8d97a0741907a54cfea1aa48a1ec4903f11756b3/firedrake/cython/dmcommon.pyx#L2265 when we construct a mesh, and set that to a section at https://github.com/firedrakeproject/firedrake/blob/8d97a0741907a54cfea1aa48a1ec4903f11756b3/firedrake/cython/dmcommon.pyx#L1231C41-L1231C41 so that `PetscSectionGetOffset()` would return offsets in the local vector as we want. 
As we now want to push "constrained owned DoFs" to the end of the `core + owned` block,
when `dmcommon.create_section` is called with `boundary_set`, we need to compute renumbering using `dmcommon.plex_renumbering` (with additional argument `boundary_set`) instead of using the one stored in the mesh. When we do this, we can also compute the size of the constrained owned DoFs.

ksagiyam:
You should also push your PyOP2 branch and do something exactly like this https://github.com/firedrakeproject/firedrake/pull/3288/files for CI to pick up your PyOP2 branch.

ksagiyam:
Some PRs that will conflict with yours have just been merged, so you will need to rebase when you have time. Specifically, there has been a change in `functionspaceimpl.py`. I think the easiest way to deal with this is to first respect all changes in `functionspaceimpl.py` on `master` and then cherry pick/apply your changes on top of that, or just do it in your favorite way.

ksagiyam (changes requested):
You need to rebase to make CI running.

As for PyOP2, this PR https://github.com/OP2/PyOP2/pull/714 has made `Mat` and `Sparsity` inherit sizes from `Dataset.layout_vec`, which represents corresponding vectors, so, after rebasing, you should only need to shrink the size of `Dataset.layout_vec`.

You will also need to run `make` in PyOP2, just as you do in Firedrake after changing Cython file. Before running `make` in PyOP2, you will need to:
```
export PETSC_DIR=$VIRTUAL_ENV/src/petsc
export PETSC_ARCH=default
```
.

ksagiyam:
Cache issue has been sorted out https://github.com/firedrakeproject/firedrake/pull/3355, so rebase.

firedrake/functionspaceimpl.py
```diff
@@ -676,7 +681,10 @@ def node_count(self):
     def dof_count(self):
         r"""The number of degrees of freedom (includes halo dofs) of this
         function space on this process. Cf. :attr:`FunctionSpace.node_count` ."""
-        return self.node_count*self.value_size
+        node_count = self.node_count
+        for sub_domain in self.boundary_set:
+            node_count -= len(self._shared_data.boundary_nodes(self, sub_domain))
+        return node_count*self.value_size
```
ksagiyam, line 687: I think we should leave `dof_count()` unchanged, but change `node_count()` as:
```
    @utils.cached_property
    def node_count(self):
        ...
        constrained_node_set = set()
        for sub_domain in self.boundary_set:
            constrained_node_set.update(self._shared_data.boundary_nodes(self, sub_domain))
        return self.node_set.total_size - len(constrained_node_set)
```

firedrake/functionspaceimpl.py
```diff
@@ -792,6 +800,86 @@ def collapse(self):
...
+
+    Notes
+    -----
+    If using this class to solve or similar, a list of DirichletBCs will still
+    need to be specified on this space and passed into the function.
+    """
+    def __init__(self, function_space, name=None, boundary_set=frozenset()):
+        label = ""
+        for boundary_domain in boundary_set:
+            label += str(boundary_domain)
```
ksagiyam, line 827: We probably want `_` or something to distinguish '(1, 2)' and '(12, )'.

firedrake/functionspaceimpl.py
```diff
@@ -815,6 +903,7 @@ def __init__(self, spaces, name=None):
         self._ufl_function_space = ufl.FunctionSpace(mesh.ufl_mesh(),
                                                      finat.ufl.MixedElement(*[s.ufl_element() for s in spaces]))
         self.name = name or "_".join(str(s.name) for s in spaces)
+        self._label = "_".join(str(s._label) for s in spaces)
```
ksagiyam, line 906: Probably:
```suggestion
        label = ""
        for s in spaces:
            label += "(" + s._label + ")_"
        self._label = label
```
or anything that is not ambiguous.

tests/regression/test_restricted_function_space.py
```diff
@@ -0,0 +1,245 @@
+import pytest
+import numpy as np
+from firedrake import *
+
+
+def compare_function_space_assembly(function_space, restricted_function_space,
+                                    bcs, res_bcs=[]):
+    u = TrialFunction(function_space)
+    v = TestFunction(function_space)
+    original_form = u * v * dx
```
ksagiyam, line 10:
```suggestion
    original_form = inner(u, v) * dx
```
to make it complex safe. The same change must happen everywhere in the below.

tests/regression/test_restricted_function_space.py
```diff
@@ -0,0 +1,245 @@
...
+    bc_res = DirichletBC(W_res.sub(0), 0, 4)
+
+    sigma2, u2 = TrialFunctions(W_res)
+    tau2, v2 = TestFunctions(W_res)
+
+    sigma, u = TrialFunctions(W)
+    tau, v = TestFunctions(W)
+
+    a = (dot(sigma, tau) + div(tau)*u + div(sigma)*v)*dx
+    a2 = (dot(sigma2, tau2) + div(tau2)*u2 + div(sigma2)*v2)*dx
```
ksagiyam, line 178:
```suggestion
    a = (inner(sigma, tau) + inner(u, div(tau)) + inner(div(sigma), v))*dx
    a2 = (inner(sigma2, tau2) + inner(u2, div(tau2)) + inner(div(sigma2), v2))*dx
```
and so on.

ksagiyam (changes requested):

ksagiyam:
Also please run `make lint` and fix lint failures.

firedrake/assemble.py
```diff
@@ -1745,7 +1746,10 @@ def _as_global_kernel_arg_output(_, self):
         elif any(V.ufl_element().family() == "Real" for V in Vs):
             el, = (create_element(V.ufl_element()) for V in Vs
                    if V.ufl_element().family() != "Real")
-            return self._make_dat_global_kernel_arg(el, V.boundary_set)
+            boundary_set = set()
+            for V in Vs:
+                boundary_set.union(set(V.boundary_set))
+            return self._make_dat_global_kernel_arg(el, frozenset(boundary_set))
```
ksagiyam, line 1752: I think this block should be more like:
```
            for V in Vs:
                if V.ufl_element().family() != "Real":
                    el = create_element(V.ufl_element())
                    boundary_set = V.boundary_set
                    break
```

ksagiyam (changes requested):

tests/regression/test_restricted_function_space.py
```diff
@@ -0,0 +1,246 @@
...
+    v_res, u_res = TestFunction(V_res), TrialFunction(V_res)
+
+    M = -inner(grad(u), grad(v))*dx - inner(u, v)*dx
+    eigenproblem = LinearEigenproblem(A=inner(v, u.dx(0))*dx, M=M, bcs=bc,
+                                      bc_shift=100, restrict=True)
+    eigensolver = LinearEigensolver(eigenproblem, n_evals=2)
+
+    M_res = -inner(grad(u_res), grad(v_res))*dx - inner(u_res, v_res)*dx
+    other_eigenproblem = LinearEigenproblem(A=inner(v_res, u_res.dx(0))*dx, M=M_res,
+                                            bcs=bc_res, bc_shift=100, restrict=False)
```
ksagiyam, line 228:
```suggestion
    other_eigenproblem = LinearEigenproblem(A=inner(u_res.dx(0), v_res)*dx, M=M_res,
                                            bcs=bc_res, bc_shift=100, restrict=False)
```

tests/regression/test_restricted_function_space.py
```diff
@@ -0,0 +1,246 @@
...
+
+
+def test_restricted_eigenvalue_problem_2():
+    # test if restrict=True gives same non-boundary eigenvalues as restrict=False
+    mesh = UnitSquareMesh(1, 1)
+    V = FunctionSpace(mesh, "CG", 1)
+    bc = DirichletBC(V, 0, 1)
+    v, u = TestFunction(V), TrialFunction(V)
+    eigenproblem = LinearEigenproblem(A=inner(v, u.dx(0))*dx, bcs=bc,
+                                      bc_shift=100, restrict=False)
```
ksagiyam, line 241:
```suggestion
    eigenproblem = LinearEigenproblem(A=inner(u.dx(0), v)*dx, bcs=bc,
                                      bc_shift=100, restrict=False)
```

tests/regression/test_restricted_function_space.py
```diff
@@ -0,0 +1,246 @@
...
+    eigensolver = LinearEigensolver(eigenproblem, n_evals=4)
+    other_eigenproblem = LinearEigenproblem(A=inner(v, u.dx(0))*dx, bcs=bc,
+                                            bc_shift=100, restrict=True)
```
ksagiyam, line 244:
```suggestion
    other_eigenproblem = LinearEigenproblem(A=inner(u.dx(0), v)*dx, bcs=bc,
                                            bc_shift=100, restrict=True)
```

firedrake/functionspaceimpl.py
ksagiyam: I meant you wouldn't need to change `dof_count` as long as you changed `node_count` correctly (which has been done):
```
    @utils.cached_property
    def dof_count(self):
        ...
        return self.node_count*self.value_size
```

ksagiyam (changes requested):

firedrake/assemble.py
```diff
@@ -1600,7 +1601,8 @@ def build(self):
                                 extruded=extruded,
                                 extruded_periodic=extruded_periodic,
                                 constant_layers=constant_layers,
-                                subset=self._needs_subset)
+                                subset=self._needs_subset,
+                                signature=signature)
```
ksagiyam, line 1605: I think we do not need to pass signature here, as `kernel_args` will distinguish what we want to distinguish. I think the issue we originally had was that `_GlobalKernelBuilder` was cached keyed by `form.signature()`, by `form.signature()` did not correctly label restricted functions, which has been fixed.

ksagiyam (changes requested):

.github/workflows/build.yml
```diff
@@ -84,6 +84,7 @@ jobs:
             --install defcon \
             --install gadopt \
             --install asQ \
+            --package-branch pyop2 emmarothwell1/restricted_function_space \
```
emmarothwell1, line 87:
```suggestion
            --package-branch pyop2 emmarothwell1/restricted_function_space_review \
```

firedrake/functionspaceimpl.py
```diff
@@ -792,6 +800,87 @@ def collapse(self):
...
 
+class RestrictedFunctionSpace(FunctionSpace):
+    r"""A representation of a function space, with additional information
+    about where boundary conditions are to be applied.
+
+    If a :class:`FunctionSpace` is represented as V, we can decompose V into
+    V = V0 + VΓ, where V0 contains functions in the basis of V that vanish on
+    the boundary where a boundary condition is applied, and VΓ contains all
+    other basis functions. This means that the :class:`RestrictedFunctionSpace`
+    corresponding to V is V0.
```
dham, line 811: Modify this to explain that when solving, test and trial functions are drawn from V0 only, and values of the solution in VΓ are treated as constant.

firedrake/solving.py
```diff
@@ -138,7 +138,7 @@ def _solve_varproblem(*args, **kwargs):
     eq, u, bcs, J, Jp, M, form_compiler_parameters, \
         solver_parameters, nullspace, nullspace_T, \
         near_nullspace, \
-        options_prefix = _extract_args(*args, **kwargs)
+        options_prefix, restrict = _extract_args(*args, **kwargs)
```
dham, line 141: Need to doculemt the `restrict` argument in the `solve` docstring.

firedrake/variational_solver.py
```diff
@@ -57,22 +61,48 @@ def __init__(self, F, u, bcs=None, J=None,
             compiler (optional)
         :is_linear: internally used to check if all domain/bc forms
             are given either in 'A == b' style or in 'F == 0' style.
+        :param restrict: (optional) flag indicating whetehr or not to change
+            the function space the arguments of the forms are defined on
+            to their restricted counterpart, based off the bcs. The output
+            space remains the same.
```
dham, line 67:
```suggestion
        :param restrict: (optional) If `True`, use restricted function spaces,
            that exclude Dirichlet boundary condition nodes,  internally for
            the test and trial spaces. 
```

firedrake/variational_solver.py
```diff
@@ -311,6 +343,10 @@ def __init__(self, a, L, u, bcs=None, aP=None,
                  Jacobian is constant (i.e. does not depend on
                  varying fields).  If your Jacobian does not change, set
                  this flag to ``True``.
+        :param restrict: (optional) flag indicating whetehr or not to change
+                 the function space the arguments of the forms are defined on
+                 to their restricted counterpart, based off the bcs. The output
+                 space remains the same.
```
dham, line 349:
```suggestion
        :param restrict: (optional) If `True`, use restricted function spaces,
            that exclude Dirichlet boundary condition nodes,  internally for
            the test and trial spaces. 
```

dham (changes requested):

firedrake/variational_solver.py
```diff
@@ -57,22 +61,48 @@ def __init__(self, F, u, bcs=None, J=None,
...
+
         if not isinstance(self.u, function.Function):
             raise TypeError("Provided solution is a '%s', not a Function" % type(self.u).__name__)
+
         # Use the user-provided Jacobian. If none is provided, derive
         # the Jacobian from the residual.
         self.J = J or ufl_expr.derivative(F, u)
+        self.F = F
+        self.Jp = Jp
+        self.restrict = restrict
```
dham, line 81:
```suggestion
        for bc in bcs:
            if isinstance(bc, EquationBC):
                restrict = False
        self.restrict = restrict
```

firedrake/variational_solver.py
```diff
@@ -44,7 +48,7 @@ class NonlinearVariationalProblem(NonlinearVariationalProblemMixin):
     def __init__(self, F, u, bcs=None, J=None,
                  Jp=None,
                  form_compiler_parameters=None,
-                 is_linear=False):
+                 is_linear=False, restrict=False):
```
dham, line 51:
```suggestion
                 is_linear=False, restrict=True):
```

firedrake/solving.py
```diff
@@ -321,9 +323,11 @@ def _extract_args(*args, **kwargs):
     form_compiler_parameters = kwargs.get("form_compiler_parameters", {})
     solver_parameters = kwargs.get("solver_parameters", {})
     options_prefix = kwargs.get("options_prefix", None)
+    restrict = kwargs.get("restrict", False)
```
dham, line 326:
```suggestion
    restrict = kwargs.get("restrict", True)
```

dham (changes requested):

ksagiyam (approved):

.github/workflows/build.yml
```diff
@@ -84,6 +84,7 @@ jobs:
             --install defcon \
             --install gadopt \
             --install asQ \
+            --package-branch pyop2 emmarothwell1/restricted_function_space_review \
```
ksagiyam, line 87:
```suggestion
```

Issue #1821, opened 2020-08-24
Project-wide revision of the use of install_requires and requirements files


miklos1:
> `install_requires` is a _setuptools_ `setup.py` keyword that should be used to specify what a project **minimally** needs to run correctly. When the project is installed by _pip_, this is the specification that is used to install its dependencies.

> _Requirements Files_ described most simply, are just a list of _pip install_ arguments placed into a file.
> Whereas `install_requires` defines the dependencies for a single project, _Requirements Files_ are often used to define the requirements for a complete Python environment.
> Whereas `install_requires` requirements are minimal, requirements files often contain an exhaustive listing of pinned versions for the purpose of achieving _repeatable installations_ of a complete environment.

https://packaging.python.org/discussions/install-requires-vs-requirements/

Short summary:
- loo.py's use of these concepts is right
- Firedrake components' use of these concepts is wrong
- The loo.py workarounds in `firedrake-install` should be the default mode of how `firedrake-install` shall install components, and each Firedrake component shall be fixed accordingly.
- At least `requirements-ext.txt` shall move to `install_requires`, which is then pulled in by any dependent package
- `requirements.txt` shall only contain specifics of the test environment for every single package

wence-:
This should probably be revisited in light of the broader issue with installing binary packages without wheels and making sure versions are consistent.

connorjward:
Closing as we now do the "right thing" with `pip install firedrake`.

PR #3216, opened 2023-11-09
Add recursivenodes requirement

pbrubeck:
# Description

https://github.com/firedrakeproject/fiat/pull/46 has been merged, but we forgot to add this requirement to firedrake.

PR #3219, opened 2023-11-09
New FIAT dependency

JDBetteridge:
# Description
Since [FIAT](https://github.com/firedrakeproject/fiat) does not provide a `requirements.txt` and Firedrake has [long abused](https://github.com/firedrakeproject/firedrake/issues/1821) the `requirement.txt` file. I suggest we temporarily add this dependency to the install script until such a time that we clean up the install.

### DIFF
--- a/scripts/firedrake-install
+++ b/scripts/firedrake-install
@@ -1940,6 +1940,9 @@ if mode == "install":
     for p in packages:
         pip_requirements(p, compiler_env)
 
+    # FIAT requires recursivenodes, but does not include a requirements.txt
+    run_pip_install(["recursivenodes"])
+
     with environment(**compiler_env):
         build_and_install_h5py()
         build_and_install_libsupermesh()


rckirby:
If FIAT's build process were updated (using a requirements file), would that solve this problem?

JDBetteridge:
Honestly, FIAT is doing the right thing™, I'd rather we long term fix Firedrake so it doesn't depend on projects declaring dependencies in a `requirements.txt`. As noted in the issue declaring dependencies in `requirements.txt` is in fact wrong.

We are crippled a bit by the fact we pass `--no-build-isolation` to most of the packages we install, which means we need all dependencies for a given package in place before we install it. There is a medium term plan to completely refactor the install script for developers to use and have a binary install for end users. However, this is a tonne of work and we need to think hard about how we simplify things.

pbrubeck:
If tests pass, this should be preferred over https://github.com/firedrakeproject/firedrake/pull/3216, since this is just a temporary workaround, and recursivenodes is not supposed to be a dependency of Firedrake.

JDBetteridge:
Supersedes #3216

PR #3221, opened 2023-11-10
Docs for `variant="spectral"` on simplices

pbrubeck:
# Description
These have to go together with https://github.com/firedrakeproject/tsfc/pull/301


docs/source/variational-problems.rst
```diff
@@ -661,3 +663,4 @@ problems.
 .. _UFL_package: http://fenics-ufl.readthedocs.io/en/latest/
 .. _FIAT: https://github.com/firedrakeproject/fiat
 .. _submanifold: https://en.wikipedia.org/wiki/Submanifold
+.. _recursivenodes: https://pypi.org/project/recursivenodes/
```
pbrubeck, line 666:
```suggestion
.. _recursivenodes: https://tisaac.gitlab.io/recursivenodes/
```

JDBetteridge (changes requested):
This is fine, but dependent on https://github.com/firedrakeproject/tsfc/pull/301

connorjward (commented):
I would quite like to wait for the docs build to pass so we can review the artefact before merging. In any case there is little point in merging before then because no one will see the changes.

connorjward (approved):
I've checked the build artefact and this is good. Thanks!

PR #3222, opened 2023-11-10
Fixing the error for the cases where rhs is Cofunction

Ig-dolci:
# Description

This PR aims to fix errors for the cases in which `self.rhs` is a `Cofunction` type. 

In this code I assumed the `self.rhs` type to be `ufl.Form`, `Number` or `Cofunction`.  If the`self.rhs` type is not in these cases, I added `NotImplementedError`. 

I also considered the code below by assuming that the unique constraint on using `ufl2unicode` is for the case where  the `self.rhs` type is a`Cofunction`.  However, I am not sure if this assumption make sense. Certainly, the reviewers know more about this. :)
```
if isinstance(self.rhs, firedrake.Cofunction):
   return "solve({} = {})".format(ufl2unicode(self.lhs),
                                           str(self.rhs))
else:
  return "solve({} = {})".format(ufl2unicode(self.lhs),
                                             ufl2unicode(self.rhs))
```

### DIFF
--- a/firedrake/adjoint_utils/blocks/solving.py
+++ b/firedrake/adjoint_utils/blocks/solving.py
@@ -79,7 +79,7 @@ class GenericSolveBlock(Block, Backend):
         else:
             raise NotImplementedError(
                 "The self.rhs type is not supported in this method yet."
-                )
+            )
 
     def _create_F_form(self):
         # Process the equation forms, replacing values with checkpoints,
--- a/firedrake/adjoint_utils/blocks/solving.py
+++ b/firedrake/adjoint_utils/blocks/solving.py
@@ -1,3 +1,4 @@
+from numbers import Number
 import numpy
 import ufl
 from ufl import replace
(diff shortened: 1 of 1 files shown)


connorjward (commented):
I feel like a better approach is to fix this upstream so `ufl2unicode` works properly for both `Cofunction`s and numbers. Otherwise we'll have to repeat this anywhere where we might want a nice unicode string for an expression containing a `Cofunction`.
Getting it to work for numbers is probably as simple as using `as_ufl` on the input expression.

nbouziani:
There is also another case to take into account, which is when the rhs is a `FormSum`, for example, when there is a sum of two cofunctions.

I agree that fixing `ufl2unicode` is a better option here. That would fix this `solve` case but also other existing problems such as tape visualisation. In fact, there is an open issue about that: https://github.com/firedrakeproject/ufl/issues/43. Note however that `ufl2unicode` already works for numbers, so you would probably just need to handle `Cofunction`, `FormSum`, and eventually `Constant`.

Ig-dolci:
> There is also another case to take into account, which is when the rhs is a `FormSum`, for example, when there is a sum of two cofunctions.
> 
> I agree that fixing `ufl2unicode` is a better option here. That would fix this `solve` case but also other existing problems such as tape visualisation. In fact, there is an open issue about that: [firedrakeproject/ufl#43](https://github.com/firedrakeproject/ufl/issues/43). Note however that `ufl2unicode` already works for numbers, so you would probably just need to handle `Cofunction`, `FormSum`, and eventually `Constant`.

I got it. I will work on this issue in the next weeks and probably send my questions to you @nbouziani. Thanks.

PR #3228, opened 2023-11-13
small fix to documentation on interpolator

APaganini:
this is a mini fix of the documentation at https://www.firedrakeproject.org/interpolation.html#id2

### DIFF
--- a/docs/source/interpolation.rst
+++ b/docs/source/interpolation.rst
@@ -86,7 +86,7 @@ Alternatively, one can use the interpolator to set the value of an existing :py:
    f = Function(V)
    interpolator.interpolate(output=f)
 
-If `expression` does not contain a :py:func:`~ufl.TestFunction` then
+If `expression` contains a :py:func:`~ufl.TestFunction` then
 the interpolator acts to interpolate :py:class:`~.Function`\s in the
 test space to those in the target space. For example:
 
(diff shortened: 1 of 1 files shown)


docs/source/interpolation.rst
```diff
@@ -98,9 +98,9 @@ test space to those in the target space. For example:
 Here, `interpolator` acts as the interpolation matrix from the
 :py:func:`~.FunctionSpace` W into the
 :py:func:`~.FunctionSpace` V. Such that if `f` is a
-:py:class:`~.Function` in `W` then `interpolator(f)` is its
-interpolation into `g`. As before, the `output` parameter can be used
-to write into an existing :py:class:`~.Function`. Passing the
+:py:class:`~.Function` in `W` then `g.Interpolator.interpolate(f)` is its
```
ReubenHill, line 101:
```suggestion
:py:class:`~.Function` in `W` then `g = interpolator.interpolate(f)` is its
```
I think

APaganini: Shall we maybe use `interpolator.interpolate(f, output=g)` so that it also works in the current state?

ReubenHill: The documentation says that `g` is a function, and the `output` parameter is mentioned below. `Cofunctions` are not currently covered in the firedrake documentation, so I don't think you need to dive into that rabbit hole.
The current documentation says
```
Passing the transpose=True option to [interpolate()](https://www.firedrakeproject.org/firedrake.html#firedrake.interpolation.Interpolator.interpolate) will cause the transpose interpolation to occur. This is equivalent to the multigrid restriction operation which interpolates assembled 1-forms in the dual space to V to assembled 1-forms in the dual space to W.
```
so it's not technically incorrect.

ReubenHill: the `output` argument is only required to get the correct type when performing adjoint interpolation with `transpose=True`. The whole interpolation section will need redocumenting once alongside the symbolic interpolation branches.

ReubenHill (approved):

PR #3229, opened 2023-11-13
Remove __del__ method and add weakref.finalizer

JDBetteridge:


connorjward (commented):
This is sooo much better.

PR #3231, opened 2023-11-14
Fix extraction of form coefficients in MG

pbrubeck:
# Description
I was mistakenly trying to `extend()` a tuple in the extraction of coefficients inside `inject_on_restrict`. To fix this, we now build an iterable of coefficients using `chain.from_iterable`, which does not rely on the assumption that `form.coefficients()` is a tuple.

### DIFF
--- a/firedrake/mg/ufl_utils.py
+++ b/firedrake/mg/ufl_utils.py
@@ -206,19 +206,19 @@ def coarsen_nlvp(problem, self, coefficient_mapping=None):
         forms = (finectx.F, finectx.J, finectx.Jp)
         coefficients = unique(chain.from_iterable(form.coefficients()
                               for form in forms if form is not None))
-        for fine in coefficients:
-            if hasattr(fine, '_child'):
-                if is_dual(fine):
-                    manager.restrict(fine, fine._child)
+        for c in coefficients:
+            if hasattr(c, '_child'):
+                if is_dual(c):
+                    manager.restrict(c, c._child)
                 else:
-                    manager.inject(fine, fine._child)
+                    manager.inject(c, c._child)
         # Apply bcs and also inject them
         for bc in chain(*finectx._problem.bcs):
             if isinstance(bc, DirichletBC):
                 bc.apply(finectx._x)
-                fine = bc.function_arg
-                if isinstance(fine, firedrake.Function) and hasattr(fine, "_child"):
-                    manager.inject(fine, fine._child)
+                g = bc.function_arg
+                if isinstance(g, firedrake.Function) and hasattr(g, "_child"):
+                    manager.inject(g, g._child)
 
     V = problem.u.function_space()
     if not hasattr(V, "_coarse"):
--- a/firedrake/mg/ufl_utils.py
+++ b/firedrake/mg/ufl_utils.py
@@ -7,6 +7,7 @@ from ufl.duals import is_dual
 from functools import singledispatch, partial
 from itertools import chain
 import firedrake
+from firedrake.utils import unique
 from firedrake.petsc import PETSc
 from firedrake.dmhooks import (get_transfer_manager, get_appctx, push_appctx, pop_appctx,
                                get_parent, push_parent, pop_parent, add_hook)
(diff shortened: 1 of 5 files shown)


firedrake/mg/ufl_utils.py
```diff
@@ -199,13 +199,14 @@ def coarsen_nlvp(problem, self, coefficient_mapping=None):
...
+
     def inject_on_restrict(fine, restriction, rscale, injection, coarse):
         from firedrake.bcs import DirichletBC
         manager = get_transfer_manager(fine)
         finectx = get_appctx(fine)
-        coefficients = finectx.F.coefficients() + finectx.J.coefficients()
-        if finectx.Jp is not None:
-            coefficients += problem.Jp.coefficients()
+        coefficients = chain.from_iterable(map(get_coefficients, (finectx.F, finectx.J, finectx.Jp)))
         coefficients = tuple(OrderedDict.fromkeys(coefficients))
```
connorjward, line 210: What does this line even do? Is it needed? At the very least `OrderedDict` isn't needed as Python dicts are now always ordered.

firedrake/mg/ufl_utils.py
```diff
@@ -226,9 +227,7 @@ def inject_on_restrict(fine, restriction, rscale, injection, coarse):
 
     # Build set of coefficients we need to coarsen
     seen = set()
-    coefficients = problem.F.coefficients() + problem.J.coefficients()
-    if problem.Jp is not None:
-        coefficients.extend(problem.Jp.coefficients())
+    coefficients = chain.from_iterable(map(get_coefficients, (problem.F, problem.J, problem.Jp)))
```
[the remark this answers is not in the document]
connorjward: This should also be defined below where we actually loop over `coefficients`.

connorjward:
Oh and could we have a test?

firedrake/mg/ufl_utils.py
```diff
@@ -199,13 +199,14 @@ def coarsen_nlvp(problem, self, coefficient_mapping=None):
...
+        return tuple() if form is None else form.coefficients()
+
     def inject_on_restrict(fine, restriction, rscale, injection, coarse):
         from firedrake.bcs import DirichletBC
         manager = get_transfer_manager(fine)
         finectx = get_appctx(fine)
-        coefficients = finectx.F.coefficients() + finectx.J.coefficients()
-        if finectx.Jp is not None:
-            coefficients += problem.Jp.coefficients()
+        coefficients = chain.from_iterable(map(get_coefficients, (finectx.F, finectx.J, finectx.Jp)))
```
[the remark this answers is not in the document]
pbrubeck: This is definitely better, thanks

firedrake/mg/ufl_utils.py
pbrubeck: Here we get a sorted tuple of unique coefficients, this was present in the first version of this code without no-ops to avoid extra work, and we want to have the same ordering on all ranks because restrict and inject are collective.

connorjward: Got it. Would you mind using [this function](https://github.com/firedrakeproject/firedrake/blob/0fcac11c41ee7b0248d3d9e0fea97d308a6d5edb/firedrake/slate/slate.py#L1392) then? Would be much clearer. The function should also be moved to `utils.py`.

firedrake/mg/ufl_utils.py
```diff
@@ -199,13 +199,14 @@ def coarsen_nlvp(problem, self, coefficient_mapping=None):
     if hasattr(problem, "_coarse"):
         return problem._coarse
 
+    def get_coefficients(form):
+        return tuple() if form is None else form.coefficients()
```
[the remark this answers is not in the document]
pbrubeck: This is gone

firedrake/mg/ufl_utils.py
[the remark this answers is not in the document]
pbrubeck: I like the generator, I am only using it once per for loop, hence only defining it in the same line as the for loops.

firedrake/mg/ufl_utils.py
connorjward: Using a `seen = set()` approach is also completely fine. It's just much more readable to have this as a `utils.py` function in my opinion. If the same exists in `variational_solver.py` then that could also be changed.

JDBetteridge: Is a `set()` fine? I thought we were worried about ordering!

pbrubeck: Yhe `seen = set()` is used differently in variational_solver.py. There we don't want to visit the same functionspace more than once (as opposed to re-visiting the same coefficient that may appear in more than form). The unique approach only makes sense in mg/ufl_utils.py

pbrubeck: On a second thought, we may first obtain the unique coefficients and then get the unique function spaces to get the unique dms of the problem

connorjward (approved):

PR #3237, opened 2023-11-15
Behavior of CrossMeshInterpolator when default_missing_val is missing or set to 0

APaganini:
`CrossMeshInterpolator` allows setting `allow_missing_dofs=True` to deal with dofs outside the source mesh. If the user provides `default_missing_val`, the dofs are overwritten with it.

This PR corrects the behavior in the case `default_missing_val=0`.

This PR also corrects the documentation for the case `allow_missing_dofs=True` and `default_missing_val=None`, in which case the extra dofs are left unchanged instead of being set to zero.

### DIFF
--- a/docs/source/interpolation.rst
+++ b/docs/source/interpolation.rst
@@ -181,7 +181,7 @@ argument:
    :lines: 72-73, 80
 
 By default the missing degrees of freedom (DoFs, the global basis function
-coefficients which could not be set) are zero:
+coefficients which could not be set) are left unchanged:
 
 .. literalinclude:: ../../tests/regression/test_interpolation_manual.py
    :language: python3
--- a/firedrake/interpolation.py
+++ b/firedrake/interpolation.py
@@ -102,15 +102,15 @@ def interpolate(
         defined on the source mesh. For example, where nodes are point
         evaluations, points in the target mesh that are not in the source mesh.
         When ``False`` this raises a ``ValueError`` should this occur. When
-        ``True`` the corresponding values are set to zero or to the value
+        ``True`` the corresponding values are left unchanged or set to the value
         ``default_missing_val`` if given. Ignored if interpolating within the
         same mesh or onto a :func:`.VertexOnlyMesh` (the behaviour of a
         :func:`.VertexOnlyMesh` in this scenario is, at present, set when
         it is created).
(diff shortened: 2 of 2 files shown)


docs/source/interpolation.rst
```diff
@@ -181,7 +181,7 @@ argument:
    :lines: 72-73, 80
 
 By default the missing degrees of freedom (DoFs, the global basis function
-coefficients which could not be set) are zero:
+coefficients which could not be set) are left unchanged:
```
ReubenHill, line 184: This is not quite correct. It should read
```
In this case, the default the missing degrees of freedom (DoFs, the global basis function coefficients which could not be set) are zero:
```
and I should follow it up with an example which uses output=some_function to show in that case that the values are unmodified. If you leave this alone, I'll do that in a separate PR

firedrake/interpolation.py
```diff
@@ -102,15 +102,15 @@ def interpolate(
...
         ``default_missing_val`` if given. Ignored if interpolating within the
         same mesh or onto a :func:`.VertexOnlyMesh` (the behaviour of a
         :func:`.VertexOnlyMesh` in this scenario is, at present, set when
         it is created).
     :kwarg default_missing_val: For interpolation across meshes: the optional
         value to assign to DoFs in the target mesh that are outside the source
-        mesh. If this is not set then zero is used. Ignored if interpolating
-        within the same mesh or onto a :func:`.VertexOnlyMesh`.
+        mesh. If this is not set then the values are left unchanged. Ignored if
+        interpolating within the same mesh or onto a :func:`.VertexOnlyMesh`.
```
ReubenHill, line 113: Same here. It needs rewording to say what happens if you supply some `output=func`

firedrake/interpolation.py
```diff
@@ -544,7 +544,7 @@ def interpolate(
             )
             # We have to create the Function before interpolating so we can
             # set default missing values (if requested).
-            if default_missing_val:
+            if default_missing_val is not None:
```
ReubenHill, line 547: Oops

ReubenHill (changes requested):
Leave the documentation changes to me and leave the bug fix. Thank you for finding the bug!

ReubenHill:
I've made #3238 which should supersede this. Thank you for finding the bug!

PR #3238, opened 2023-11-16
Fix interpolation behaviour when default_missing_val=0 and clarify documentation

ReubenHill:
In cross-mesh interpolation, where we had `default_missing_val=0.0` it would not be used due to an incorrect if statement. This fixes that and clarifies some documentation. Supersedes #3237

### DIFF
--- a/docs/source/interpolation.rst
+++ b/docs/source/interpolation.rst
@@ -180,13 +180,16 @@ argument:
    :dedent:
    :lines: 72-73, 80
 
-By default the missing degrees of freedom (DoFs, the global basis function
-coefficients which could not be set) are zero:
+In this case, the missing degrees of freedom (DoFs, the global basis function
+coefficients which could not be set) are, by default, set to zero:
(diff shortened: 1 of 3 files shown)


docs/source/interpolation.rst
```diff
@@ -180,14 +180,17 @@ argument:
    :dedent:
    :lines: 72-73, 80
 
-By default the missing degrees of freedom (DoFs, the global basis function
-coefficients which could not be set) are zero:
+In this case, the default the missing degrees of freedom (DoFs, the global
```
APaganini, line 183: In this case, by default

APaganini (commented):
just a typo in the comment to fix

docs/source/interpolation.rst
ReubenHill: I've reworded, hopefully it's clearer now?

APaganini (approved):

PR #3260, opened 2023-11-23
fixing pytest petsc messages errors 

Ig-dolci:
I am not sure what `--cov firedrake` is, but the fails with `firedrake_adjoint` tests are sorted locally when I add it.

### DIFF
--- a/firedrake/petsc.py
+++ b/firedrake/petsc.py
@@ -3,7 +3,7 @@ import os
 import subprocess
 import petsc4py
 import sys
-from pyop2.mpi import COMM_WORLD
+from mpi4py.MPI import COMM_WORLD
 petsc4py.init(args=sys.argv, comm=COMM_WORLD)
 from petsc4py import PETSc
 import itertools
--- a/firedrake/petsc.py
+++ b/firedrake/petsc.py
@@ -3,7 +3,8 @@ import os
 import subprocess
 import petsc4py
 import sys
-petsc4py.init(sys.argv)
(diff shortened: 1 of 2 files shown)


francesco-ballarin:
FYI, `--cov` is a flag of https://pypi.org/project/pytest-cov/

Ig-dolci:
> FYI, `--cov` is a flag of https://pypi.org/project/pytest-cov/

Thanks!

Ig-dolci:
> This feels like a cosmetic fix, rather than actually addressing the underlying issue

I agree. For some reason, `--cov firedrake` hides this petsc message error in firedrake tests. That is not fixing the original problem. I am trying to understand more...

Ig-dolci:
The latest commit fixes the petsc message errors.

Obs: I still do not know why `--cov firedrake` hides them.

firedrake/petsc.py
```diff
@@ -3,7 +3,8 @@
 import subprocess
 import petsc4py
 import sys
-petsc4py.init(sys.argv)
+from pyop2.mpi import COMM_WORLD
```
JDBetteridge, line 6:
```suggestion
from mpi4py.MPI import COMM_WORLD
```
I think we actually want to use the "public" `COMM_WORLD` here so that petsc4py is initialised on the same communicator as a user might use in a script. We do not need to worry about communication clash on the communicator as we know that PETSc is duplicating the communicator behind the scenes anyway.

Issue #3224, opened 2023-11-10
Missing constant handlers


Ig-dolci:
**Describe the bug**
Missing handled for `Constant`

**Steps to Reproduce**
```from firedrake import *
from firedrake_adjoint import *
continue_annotation()

n = 100
mesh = UnitIntervalMesh(n)
V = FunctionSpace(mesh, "CG", 2)
end = 0.01
timestep = Constant(1.0/n)
steps = int(end/float(timestep)) + 1

def Dt(u, u_, timestep):
    return (u - u_)/timestep

x, = SpatialCoordinate(mesh)
ic = project(sin(2.*pi*x), V)
u_ = Function(V)
u = Function(V)
v = TestFunction(V)
u_.assign(ic)
nu = Constant(0.0001)
F = (Dt(u, u_, timestep)*v
        + u*u.dx(0)*v + nu*u.dx(0)*v.dx(0))*dx
bc = DirichletBC(V, 0.0, "on_boundary")

solve(F == 0, u, bc)

tape = get_working_tape()
tape.visualise("tape.pdf")
```

**Expected behavior**
Describe what you expected to happen, in the case of a regression,
include details of a setup where this behaviour was last seen.

**Error message**
```
  File "/Users/ddolci/work/firedrake_main/src/ufl/ufl/corealg/map_dag.py", line 98, in map_expr_dags
    r = handlers[v._ufl_typecode_](v)
        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/ddolci/work/firedrake_main/src/ufl/ufl/formatting/ufl2unicode.py", line 813, in expr
    raise ValueError(f"Missing handler for type {type(o)}")
ValueError: Missing handler for type <class 'firedrake.constant.Constant'>
```

**Environment:**
 - OS:  MacOS
 - Python version: 3.11.6

**Additional Info**
Write a `Constant` handle within Firedrake that can fix this error.

nbouziani:
`ufl2unicode` doesn't currently work with `BaseForm` objects and `Constant` as well, as discussed in the following UFL issue https://github.com/firedrakeproject/ufl/issues/43, which breaks tape visualisation. I will close that issue as this is a UFL problem and there is already a opened issue.

connorjward:
The missing`Constant` handler is actually a Firedrake problem as it is a class that we implement outside of UFL. I think this issue should remain open.

nbouziani:
[...]

PR #3261, opened 2023-11-23
Constant and derivative

JDBetteridge:
# Description
Improvements to some oversights with the changes I previously made to `Constant`.
- `derivative` now substitutes the symbolic constant back into the form after taking the derivative. This allows the user to perform solves with the resultant form, which was not the case before.
- Removes superfluous numbering code in derivative, which was not necessary if I had done things correctly in the first place.
- Adds `ufl2unicode` support for `Constant` via various hackery.


firedrake/constant.py
```diff
@@ -197,3 +205,32 @@ def __idiv__(self, o):
...
+    var = "💩"
+    if len(o.ufl_shape) == 1:
+        var += UC.combining_right_arrow_above
+    elif len(o.ufl_shape) > 1 and self.colorama_bold:
+        var = f"{colorama.Style.BRIGHT}{var}{colorama.Style.RESET_ALL}"
+    return f"{var}{subscript_number(i)}"
+
+
+# This monkey patches ufl2unicode support for Firedrake constants
+Expression2UnicodeHandler.firedrake_constant = _unicode_format_firedrake_constant
```
nbouziani, line 223: I don't think that's the right place to add `ufl2unicode` support for constants, see my comment [here](https://github.com/firedrakeproject/firedrake/issues/3224).

JDBetteridge: I agree, this code is disgusting. I envision my code as being a temporary workaround current issues whilst someone with more UFL skills fixes this upstream.

tests/regression/test_constant.py
```diff
@@ -231,3 +233,33 @@ class CustomConstant(Constant):
...
+    c = Constant(5)
+    f = Function(V)
+    solution = Function(V)
+
+    a = (c**2)*inner(u, v) * dx
+    L = inner(f, v) * dx
+    solve(a == L, solution)
+
+    d = derivative(a, c, IntValue(1))
+    solve(d == L, solution)
```
pbrubeck, line 253: assert solution1 == (c/2)*solution2 ?

tests/regression/test_constant.py
```diff
@@ -231,3 +233,33 @@ class CustomConstant(Constant):
...
+
+
+def test_constant_ufl2unicode():
+    mesh = UnitIntervalMesh(1)
+    a = Constant(1.0, name="a")
+    b = Constant(2.0, name="b")
+    F = a * a * b * b * dx(mesh)
+    _ = ufl2unicode(F)
+    dFda = derivative(F, u=a, du=ufl.classes.IntValue(1))
+    dFdb = derivative(F, u=b, du=ufl.classes.IntValue(1))
```
pbrubeck, line 263: Could `du` be left unspecified?

tests/regression/test_constant.py
JDBetteridge: This is not a good idea, in general this is a hard problem to test the equality of two expressions symbolically: https://en.wikipedia.org/wiki/Computer_algebra#Equality

connorjward: `np.allclose(solution1.dat.data_ro, solution2.dat.data_ro)`?

tests/regression/test_constant.py
JDBetteridge: Sure

tests/regression/test_constant.py
```diff
@@ -239,3 +241,42 @@ class CustomConstant(Constant):
...
+    solution_b = Function(V)
+
+    a = (c**2)*inner(u, v) * dx
+    L = inner(f, v) * dx
+    solve(a == L, solution_a)
+
+    d = derivative(a, c, IntValue(1))
+    solve(d == L, solution_b)
+
+    assert np.allclose(solution_b.dat.data, (c.dat.data/2)*solution_a.dat.data)
```
connorjward, line 264:
```suggestion
    assert np.allclose(solution_b.dat.data_ro, float(c)/2*solution_a.dat.data_ro)
```

connorjward (changes requested):
I think this is basically fine, but a better mathematician than me should really take a look at this.

I think eventually we might want to consider dropping support for differentiating wrt constants. Downstream users that rely on this functionality (hopefully not many!) could do the replacement themselves.

tests/regression/test_constant.py
JDBetteridge: I did a thing

dham (approved):

PR #3263, opened 2023-11-25
Store the .vtu files in a subdirectory within the .pvd file directory

lrtfm:
Organizing .vtu files in a subdirectory within the directory of a .pvd file enhances overall file organization

### DIFF
--- a/firedrake/output.py
+++ b/firedrake/output.py
@@ -403,9 +403,8 @@ class File(object):
         self._comm = internal_comm(self.comm)
 
         if self._comm.rank == 0 and mode == "w":
-            outdir = os.path.dirname(os.path.abspath(filename))
-            if not os.path.exists(outdir):
-                os.makedirs(outdir)
+            if not os.path.exists(basename):
+                os.makedirs(basename)
         elif self._comm.rank == 0 and mode == "a":
             if not os.path.exists(os.path.abspath(filename)):
                 raise ValueError("Need a file to restart from.")
@@ -413,6 +412,16 @@ class File(object):
 
         self.filename = filename
         self.basename = basename
+        # The vtu files will be in the subdirectory named same with the pvd file.
+        # Assuming the absolute path of the pvd file is "/path/to/foo.pvd", the vtu
+        # files will be in the subdirectory "foo" like this:
+        #
+        #     /path/to/foo/foo_0.vtu
+        #     /path/to/foo/foo_1.vtu
+        #
+        # The basename for this example is `/path/to/foo`, and the vtu_basename
+        # is `/path/to/foo/foo`.
+        self.vtu_basename = os.path.join(basename, os.path.basename(basename))
         self.project = project_output
         self.target_degree = target_degree
         self.target_continuity = target_continuity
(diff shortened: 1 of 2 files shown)


firedrake/output.py
```diff
@@ -403,16 +403,16 @@ def __init__(self, filename, project_output=False, comm=None, mode="w",
...
+            if not os.path.exists(basename):
+                os.makedirs(basename)
         elif self._comm.rank == 0 and mode == "a":
             if not os.path.exists(os.path.abspath(filename)):
                 raise ValueError("Need a file to restart from.")
         self._comm.barrier()
 
         self.filename = filename
         self.basename = basename
+        self.vtu_basename = os.path.join(basename, os.path.basename(basename))
```
dham, line 415: I don't understand this line. Why is this the right thing?

dham (changes requested):

dham:
Please also rebase on master and rerun the tests.

firedrake/output.py
lrtfm: The `basename` is absolute path to the `.pvd` file without suffix `.pvd`, and it is also the path to `vtu` file. 
Here, we join the `vtu` dir name with the prefix of the `pvd` filename to obtain the `vtu_basename`.

For example, code like this 
```python
output = File('/my/path/to/test.pvd')
output.write(fun)
```
will generate files
```
/my/path/to/test.pvd
/my/path/to/test/test_0.vtu
```
The value of `basename` is  `/my/path/to/test`, and the value of `vtu_basename` is `/my/path/to/test/test`

lrtfm:
> Please also rebase on master and rerun the tests.

Done.

tests/output/test_pvd_output.py
```diff
@@ -121,7 +121,9 @@ def test_append(mesh, tmpdir):
     restarted_outfile = File(str(tmpdir.join("restart_test.pvd")), mode="a")
     restarted_outfile.write(g)
 
-    files_in_tmp = [f for f in listdir(str(tmpdir)) if isfile(join(str(tmpdir), f))]
+    vtufile_dir = tmpdir.join("restart_test")
+    files_in_tmp = [f for f in listdir(str(tmpdir)) if isfile(join(str(tmpdir), f))] \
+        + [f for f in listdir(str(vtufile_dir)) if isfile(join(str(vtufile_dir), f))]
```
connorjward, line 126: I think it might make sense to split this into two checks. You are looking for files in different directories and this isn't at all obvious using this code.

firedrake/output.py
```diff
@@ -536,7 +536,7 @@ def _write_vtu(self, *functions):
         if self._topology is None or self._adaptive:
             self._topology = get_topology(coordinates.function)
 
-        basename = "%s_%s" % (self.basename, next(self.counter))
+        basename = "%s_%s" % (self.vtu_basename, next(self.counter))
```
connorjward, line 539:
```suggestion
        basename = f"{self.vtu_basename}_{next(self.counter)}"
```

firedrake/output.py
connorjward: Please provide a comment that explains this. It is not especially obvious.

connorjward (changes requested):
Thank you for this work. I think there are some small improvements but otherwise this looks good to me.

connorjward (approved):
Looks good. Thanks!

PR #3269, opened 2023-11-27
Fix gamg options in test

pbrubeck:
# Description
This test breaks after a recent FIAT change. Note that for some reason we only required to specify this gamg option on the linear version of this test, that presumably broke at a certain point after it was first introduced https://github.com/firedrakeproject/firedrake/commit/d32eba573f6abe96198d7c3c4f7991770a51e717


ksagiyam (approved):
Thanks.

PR #3279, opened 2023-12-04
Fix some tests for new FIAT quadratures

pbrubeck:
# Description
Bumped up some tolerances, and fixed some issues that we already present in a test with underintegration of a variable coefficient.

### DIFF
--- a/tests/regression/test_nullspace.py
+++ b/tests/regression/test_nullspace.py
@@ -287,7 +287,8 @@ def test_nullspace_mixed_multiple_components():
     assert schur_ksp.getIterationNumber() < 6
 
 
-def test_near_nullspace_mixed():
+@pytest.mark.parametrize("aux_pc", [False, True], ids=["PC(mu)", "PC(DG0-mu)"])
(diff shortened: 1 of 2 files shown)


rckirby (approved):

tests/regression/test_nullspace.py
```diff
@@ -341,8 +351,6 @@ def test_near_nullspace_mixed():
             'pc_type': 'python',
             'pc_python_type': 'firedrake.AssembledPC',
             'assembled_pc_type': 'gamg',
-            'assembled_pc_gamg_aggressive_square_graph': None,  # See https://gitlab.com/petsc/petsc/-/commit/d529f056d75cd16b380adc3e50ccb741df53d0de
-            'assembled_pc_gamg_mis_k_minimum_degree_ordering': True,  # See https://gitlab.com/petsc/petsc/-/commit/d529f056d75cd16b380adc3e50ccb741df53d0de
```
wence-, line 345: Is removing of these options still correct, @stephankramer I know you've had a bunch of back and worth with Mark around gamg changes recently.

PR #3288, opened 2023-12-10
DO NOT SQUASH: Specify sparsity block-wise

ksagiyam:
Depends on https://github.com/OP2/PyOP2/pull/713.

- refactor assemble.py,
- make sparsity block-wise.

### DIFF
--- a/firedrake/assemble.py
+++ b/firedrake/assemble.py
@@ -128,11 +128,85 @@ def get_form_assembler(form, tensor, *args, **kwargs):
             raise ValueError('Expecting a 0-, 1-, or 2-form: got %s' % (form))
     elif isinstance(form, ufl.core.expr.Expr) and not isinstance(form, ufl.core.base_form_operator.BaseFormOperator):
         # BaseForm preprocessing can turn BaseForm into an Expr (cf. case (6) in `restructure_base_form`)
-        return functools.partial(_assemble_expr, form)
+        return functools.partial(ExprAssembler(form).assemble, tensor=tensor)
     elif isinstance(form, ufl.form.BaseForm):
         return functools.partial(BaseFormAssembler(form, *args, **kwargs).assemble, tensor=tensor)
     else:
         raise ValueError(f'Expecting a BaseForm, slate.TensorBase, or Expr object: got {form}')
+
+
+class ExprAssembler(object):
+    """Expression assembler.
+
+    Parameters
+    ----------
+    expr : ufl.core.expr.Expr
+        Expression.
+
+    """
+
+    def __init__(self, expr):
+        self._expr = expr
+
+    def assemble(self, tensor=None):
+        """Assemble the pointwise expression.
+
+        Parameters
+        ----------
+        tensor : firedrake.function.Function or firedrake.cofunction.Cofunction or matrix.MatrixBase
+            Output tensor.
+
+        Returns
+        -------
+        float or firedrake.cofunction.Cofunction or firedrake.function.Function or matrix.MatrixBase
+            Result of evaluation: `float` for 0-forms, `firedrake.cofunction.Cofunction` or `firedrake.function.Function` for 1-forms, and `matrix.MatrixBase` for 2-forms.
+
+        """
+        from ufl.algorithms.analysis import extract_base_form_operators
+        from ufl.checks import is_scalar_constant_expression
+
+        assert tensor is None
+        expr = self._expr
+        # Get BaseFormOperators (e.g. `Interpolate` or `ExternalOperator`)
+        base_form_operators = extract_base_form_operators(expr)
+
+        # -- Linear combination involving 2-form BaseFormOperators -- #
+        # Example: a * dNdu1(u1, u2; v1, v*) + b * dNdu2(u1, u2; v2, v*)
+        # with u1, u2 Functions, v1, v2, v* BaseArguments, dNdu1, dNdu2 BaseFormOperators, and a, b scalars.
+        if len(base_form_operators) and any(len(e.arguments()) > 1 for e in base_form_operators):
+            if isinstance(expr, ufl.algebra.Sum):
+                a, b = [assemble(e) for e in expr.ufl_operands]
+                # Only Expr resulting in a Matrix if assembled are BaseFormOperator
+                if not all(isinstance(op, matrix.AssembledMatrix) for op in (a, b)):
+                    raise TypeError('Mismatching Sum shapes')
+                return get_form_assembler(ufl.FormSum((a, 1), (b, 1)), None)()
+            elif isinstance(expr, ufl.algebra.Product):
+                a, b = expr.ufl_operands
+                scalar = [e for e in expr.ufl_operands if is_scalar_constant_expression(e)]
+                if scalar:
+                    base_form = a if a is scalar else b
+                    assembled_mat = assemble(base_form)
+                    return get_form_assembler(ufl.FormSum((assembled_mat, scalar[0])), None)()
+                a, b = [assemble(e) for e in (a, b)]
+                return get_form_assembler(ufl.action(a, b), None)()
+        # -- Linear combination of Functions and 1-form BaseFormOperators -- #
+        # Example: a * u1 + b * u2 + c * N(u1; v*) + d * N(u2; v*)
+        # with u1, u2 Functions, N a BaseFormOperator, and a, b, c, d scalars or 0-form BaseFormOperators.
+        else:
+            base_form_operators = extract_base_form_operators(expr)
+            assembled_bfops = [firedrake.assemble(e) for e in base_form_operators]
+            # Substitute base form operators with their output before examining the expression
+            # which avoids conflict when determining function space, for example:
+            # extract_coefficients(Interpolate(u, V2)) with u \in V1 will result in an output function space V1
+            # instead of V2.
+            if base_form_operators:
+                expr = ufl.replace(expr, dict(zip(base_form_operators, assembled_bfops)))
+            try:
+                coefficients = ufl.algorithms.extract_coefficients(expr)
+                V, = set(c.function_space() for c in coefficients) - {None}
+            except ValueError:
+                raise ValueError("Cannot deduce correct target space from pointwise expression")
+            return firedrake.Function(V).assign(expr)
 
 
 class AbstractFormAssembler(abc.ABC):
@@ -805,57 +879,6 @@ def allocate_matrix(expr, bcs=None, *, mat_type=None, sub_mat_type=None,
 
     return matrix.Matrix(expr, bcs, mat_type, sparsity, ScalarType,
                          options_prefix=options_prefix)
-
-
-def _assemble_expr(expr):
-    """Assemble a pointwise expression.
-
-    :arg expr: The :class:`ufl.core.expr.Expr` to be evaluated.
-    :returns: A :class:`firedrake.Function` containing the result of this evaluation.
-    """
-    from ufl.algorithms.analysis import extract_base_form_operators
-    from ufl.checks import is_scalar_constant_expression
-
-    # Get BaseFormOperators (e.g. `Interpolate` or `ExternalOperator`)
-    base_form_operators = extract_base_form_operators(expr)
-
-    # -- Linear combination involving 2-form BaseFormOperators -- #
-    # Example: a * dNdu1(u1, u2; v1, v*) + b * dNdu2(u1, u2; v2, v*)
-    # with u1, u2 Functions, v1, v2, v* BaseArguments, dNdu1, dNdu2 BaseFormOperators, and a, b scalars.
-    if len(base_form_operators) and any(len(e.arguments()) > 1 for e in base_form_operators):
-        if isinstance(expr, ufl.algebra.Sum):
-            a, b = [assemble(e) for e in expr.ufl_operands]
-            # Only Expr resulting in a Matrix if assembled are BaseFormOperator
-            if not all(isinstance(op, matrix.AssembledMatrix) for op in (a, b)):
-                raise TypeError('Mismatching Sum shapes')
-            return get_form_assembler(ufl.FormSum((a, 1), (b, 1)), None)()
-        elif isinstance(expr, ufl.algebra.Product):
-            a, b = expr.ufl_operands
-            scalar = [e for e in expr.ufl_operands if is_scalar_constant_expression(e)]
-            if scalar:
-                base_form = a if a is scalar else b
-                assembled_mat = assemble(base_form)
-                return get_form_assembler(ufl.FormSum((assembled_mat, scalar[0])), None)()
-            a, b = [assemble(e) for e in (a, b)]
-            return get_form_assembler(ufl.action(a, b), None)()
-    # -- Linear combination of Functions and 1-form BaseFormOperators -- #
-    # Example: a * u1 + b * u2 + c * N(u1; v*) + d * N(u2; v*)
-    # with u1, u2 Functions, N a BaseFormOperator, and a, b, c, d scalars or 0-form BaseFormOperators.
-    else:
-        base_form_operators = extract_base_form_operators(expr)
-        assembled_bfops = [firedrake.assemble(e) for e in base_form_operators]
-        # Substitute base form operators with their output before examining the expression
-        # which avoids conflict when determining function space, for example:
-        # extract_coefficients(Interpolate(u, V2)) with u \in V1 will result in an output function space V1
-        # instead of V2.
-        if base_form_operators:
-            expr = ufl.replace(expr, dict(zip(base_form_operators, assembled_bfops)))
-        try:
-            coefficients = ufl.algorithms.extract_coefficients(expr)
-            V, = set(c.function_space() for c in coefficients) - {None}
-        except ValueError:
-            raise ValueError("Cannot deduce correct target space from pointwise expression")
-        return firedrake.Function(V).assign(expr)
 
 
 class FormAssembler(AbstractFormAssembler):
--- a/firedrake/assemble.py
+++ b/firedrake/assemble.py
@@ -808,29 +808,6 @@ def _assemble_form(form, tensor=None, bcs=None, *,
 
     See :func:`assemble` for a description of the arguments to this function.
     """
-    bcs = solving._extract_bcs(bcs)
-
-    # It is expensive to construct new assemblers because extracting the data
-    # from the form is slow. Since all of the data structures in the assembler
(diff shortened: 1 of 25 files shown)


firedrake/assemble.py
```diff
@@ -425,10 +426,31 @@ def allocate_matrix(expr, bcs=None, *, mat_type=None, sub_mat_type=None,
                                          for get_map, regions in domains.items()
                                          if regions))
     try:
+        # Let tensor=None and bcs=None for the purpose of allocation.
```
connorjward, line 429: This comment is a bit obscure. I understand why `tensor` must be `None`, but why `bcs`?

firedrake/assemble.py
```diff
@@ -425,10 +426,31 @@ def allocate_matrix(expr, bcs=None, *, mat_type=None, sub_mat_type=None,
...
+        parloop_builders = TwoFormAssembler(expr, None, bcs=None, form_compiler_parameters=form_compiler_parameters).parloop_builders_for_sparsity_construction
+        maps_and_regions = defaultdict(lambda: defaultdict(set))
+        if any(parloop_builder._indices == (None, None) for parloop_builder in parloop_builders):
+            # Handle special cases: slate or split=False
+            assert all(parloop_builder._indices == (None, None) for parloop_builder in parloop_builders)
+            for parloop_builder in parloop_builders:
+                _integral_type = parloop_builder._integral_type
+                get_map, region = mapping[_integral_type]
```
connorjward, line 437:
```suggestion
                get_map, region = mapping[parloop_builder._integral_type]
```

firedrake/assemble.py
```diff
@@ -425,10 +426,31 @@ def allocate_matrix(expr, bcs=None, *, mat_type=None, sub_mat_type=None,
...
+                for i, _rmap in enumerate(get_map(test)):
+                    for j, _cmap in enumerate(get_map(trial)):
+                        maps_and_regions[(i, j)][(_rmap, _cmap)].update((region, ))
+        else:
+            for parloop_builder in parloop_builders:
+                i, j = parloop_builder._indices
+                _integral_type = parloop_builder._integral_type
+                # Make Sparsity independent of _iterset for better reusability.
+                get_map, region = mapping[_integral_type]
```
connorjward, line 446:
```suggestion
                # Make Sparsity independent of _iterset for better reusability.
                get_map, region = mapping[parloop_builder._integral_type]
```

firedrake/assemble.py
```diff
@@ -425,10 +426,31 @@ def allocate_matrix(expr, bcs=None, *, mat_type=None, sub_mat_type=None,
...
+        # Let tensor=None and bcs=None for the purpose of allocation.
+        parloop_builders = TwoFormAssembler(expr, None, bcs=None, form_compiler_parameters=form_compiler_parameters).parloop_builders_for_sparsity_construction
+        maps_and_regions = defaultdict(lambda: defaultdict(set))
+        if any(parloop_builder._indices == (None, None) for parloop_builder in parloop_builders):
+            # Handle special cases: slate or split=False
+            assert all(parloop_builder._indices == (None, None) for parloop_builder in parloop_builders)
+            for parloop_builder in parloop_builders:
+                _integral_type = parloop_builder._integral_type
+                get_map, region = mapping[_integral_type]
+                for i, _rmap in enumerate(get_map(test)):
```
connorjward, line 438: Why the leading underscore? These aren't private attributes.

firedrake/assemble.py
```diff
@@ -425,10 +426,31 @@ def allocate_matrix(expr, bcs=None, *, mat_type=None, sub_mat_type=None,
...
+                    for j, _cmap in enumerate(get_map(trial)):
+                        maps_and_regions[(i, j)][(_rmap, _cmap)].update((region, ))
```
connorjward, line 440:
```suggestion
                        maps_and_regions[(i, j)][(_rmap, _cmap)].add(region)
```
?

firedrake/assemble.py
```diff
@@ -425,10 +426,31 @@ def allocate_matrix(expr, bcs=None, *, mat_type=None, sub_mat_type=None,
...
+        else:
+            for parloop_builder in parloop_builders:
+                i, j = parloop_builder._indices
+                _integral_type = parloop_builder._integral_type
+                # Make Sparsity independent of _iterset for better reusability.
+                get_map, region = mapping[_integral_type]
+                _rmap = get_map(test).split[i] if get_map(test) is not None else None
+                _cmap = get_map(trial).split[j] if get_map(trial) is not None else None
+                maps_and_regions[(i, j)][(_rmap, _cmap)].update((region, ))
```
connorjward, line 449:
```suggestion
                maps_and_regions[(i, j)][(_rmap, _cmap)].add(region)
```
?

firedrake/assemble.py
```diff
@@ -425,10 +426,31 @@ def allocate_matrix(expr, bcs=None, *, mat_type=None, sub_mat_type=None,
...
+        maps_and_regions = {key: [k + (tuple(v), ) for k, v in val.items()] for key, val in maps_and_regions.items()}
```
connorjward, line 450: This line is very obscure. Could it be made clearer?

firedrake/assemble.py
```diff
@@ -995,6 +1019,15 @@ def _apply_bcs_mat_real_block(op2tensor, i, j, component, node_set):
             dat = op2.DatView(dat, component)
         dat.zero(subset=node_set)
 
+    @cached_property
+    def parloop_builders_for_sparsity_construction(self):
```
connorjward, line 1023: I think that this would definitely need documenting. What magic are we doing here?

tests/equation_bcs/test_equation_bcs_assemble.py
```diff
@@ -15,8 +15,6 @@ def test_equation_bcs_direct_assemble_one_form():
 
     g = assemble(F, bcs=bc.extract_form('F'))
     assert np.allclose(g.dat.data, [0.5, 0.5, 0, 0])
-    g = assemble(F, bcs=bc)
-    assert np.allclose(g.dat.data, [0.5, 0.5, 0, 0])
```
connorjward, line 19: Why are you removing this test?

tests/regression/test_assemble.py
```diff
@@ -291,3 +291,28 @@ def test_assemble_vector_rspace_one_form(mesh):
     U = inner(u, u)*dx
     L = derivative(U, u)
     assemble(L)
+
+
+def test_assemble_sparsity():
```
connorjward, line 296: Can this test name (and below) be made more specific?

tests/regression/test_assemble.py
```diff
@@ -291,3 +291,28 @@ def test_assemble_vector_rspace_one_form(mesh):
...
+    u = TrialFunction(W)
+    v = TestFunction(W)
+    A = assemble(inner(u, v) * dx, mat_type="nest")
+    for i in range(len(W)):
+        for j in range(len(W)):
+            if i != j:
+                assert np.all(A.M.sparsity[i][j].nnz == np.zeros(9, dtype=IntType))
+
+
+def test_assemble_sparsity_with_bc():
```
connorjward, line 309: Do we need an `EquationBC` test too?

firedrake/assemble.py
```diff
@@ -425,10 +426,31 @@ def allocate_matrix(expr, bcs=None, *, mat_type=None, sub_mat_type=None,
...
+                _integral_type = parloop_builder._integral_type
+                get_map, region = mapping[_integral_type]
+                for i, _rmap in enumerate(get_map(test)):
+                    for j, _cmap in enumerate(get_map(trial)):
+                        maps_and_regions[(i, j)][(_rmap, _cmap)].update((region, ))
+        else:
+            for parloop_builder in parloop_builders:
+                i, j = parloop_builder._indices
+                _integral_type = parloop_builder._integral_type
+                # Make Sparsity independent of _iterset for better reusability.
```
connorjward, line 445: How is this true? Isn't the iterset included via the maps?

firedrake/assemble.py
```diff
@@ -644,8 +666,6 @@ class FormAssembler(abc.ABC):
     """
 
     def __init__(self, form, tensor, bcs=(), form_compiler_parameters=None, needs_zeroing=True, weight=1.0):
-        assert tensor is not None
```
connorjward, line 647: I don't like that you are breaking this assertion. You are sort of building a dummy `TwoFormAssembler` just to get some properties from the parloops. Could you not get that information some other way? Perhaps by refactoring `ParloopBuilder`?

connorjward (changes requested):

firedrake/assemble.py
ksagiyam: `bcs` has not been used to construct sparsity. This allows for reusing the constructed sparsity for the same problem with slightly different BCs.

firedrake/assemble.py
ksagiyam: Fixed.

firedrake/assemble.py
ksagiyam: Fixed.

tests/equation_bcs/test_equation_bcs_assemble.py
ksagiyam: This test does the same thing as the one right above, but it uses a shortcut that is only valid for 1-forms. I removed this test as I disabled the shortcut to simplify `assemble.py`.

tests/regression/test_assemble.py
ksagiyam: Correctness of the changes introduced in this PR is checked by the existing tests (including the ones for `EquationBC`) as `Sparsity` is used in most places.

firedrake/assemble.py
ksagiyam: Sparsity construction should basically trace the actual assembly process to be performed (with some abstraction), so I think we should be able to use the same infrastructure, i.e., `FormAssembler`; as we make `ParloopBuilder`s via `FormAssembler` in the actual assembly, we should do the same when making sparsity instead of inventing a different way to make `ParloopBuilder`s only for sparsity. Indeed, I think, `FormAssembler` is nicely reusable in sparsity construction, except possibly for the `tensor is not None` check. I think `FormAssembler` would be cleaner if it did not carry `self._tensor` at all and made `assemble()` method take `tensor` as argument, as `self._tensor` is only used in limited places; I at least included `assert self._tensor is not None` at the top of `assemble()` method.

ksagiyam: Maybe `FormAssembler` can carry `self._tensor`, but we should probably clarify where it is required, as, apart from `self._tensor`, `FormAssembler` is reusable in sparsity construction.

firedrake/assemble.py
connorjward: Sorry, my feedback was a bit obscure. Could you reword the comment to make it clearer?

firedrake/assemble.py
connorjward: I suppose that it doesn't need to have a `_tensor` attribute. We could perhaps make it an argument of `.assemble()`? I have never liked how I [do this](https://github.com/firedrakeproject/firedrake/blob/master/firedrake/assemble.py#L691).

> Sparsity construction should basically trace the actual assembly process to be performed (with some abstraction), so I think we should be able to use the same infrastructure

This does make sense, but I think the class hierarchy may need tweaking. For instance, `FormAssembler` has an `assemble()` method that does not make sense for a `SparsityConstructor`.

Sorry for being such a stickler here, but assembly is sufficiently complicated that I think we need to be careful about adding any extra "hacks".

tests/equation_bcs/test_equation_bcs_assemble.py
connorjward: Simplification is always good!

firedrake/assemble.py
ksagiyam: We could take `_tensor` out, but it turned out that it is currently used to make lgmaps, where we use the underlying `PETSc.Mat`, so I think we should do that in a separate PR.

I added another commit to separate the actual assembly.

firedrake/assemble.py
ksagiyam: Done.

firedrake/assemble.py
```diff
@@ -1,5 +1,6 @@
 import abc
-from collections import OrderedDict
+from collections import OrderedDict, defaultdict
+from collections.abc import Sequence  # noqa: F401
```
connorjward, line 3: This seems wrong. Why the `noqa`?

firedrake/assemble.py
```diff
@@ -632,76 +655,30 @@ def _make_tensor(form, bcs, *, diagonal, mat_type, sub_mat_type, appctx,
         raise AssertionError
 
 
-class FormAssembler(abc.ABC):
-    """Abstract base class for assembling forms.
+class FormAssemblerParloopBuilder(abc.ABC):
+    """Abstract base class for making ParloopBuilders for assembly.
```
connorjward, line 659: The pattern here seems to be "builder class that constructs another builder class". That seems wrong. I am quite confused by the inheritance going on.

firedrake/assemble.py
ksagiyam: This is for docstring.

firedrake/assemble.py
ksagiyam: This is no longer relevant.

firedrake/assemble.py
```diff
@@ -108,6 +108,48 @@ def assemble(expr, *args, **kwargs):
         raise TypeError(f"Unable to assemble: {expr}")
 
 
+class FormAssemblerBase(abc.ABC):
```
dham, line 242: Rename to `AbstractFormAssembler` to avoid confusion with `BaseFormAssembler`

firedrake/assemble.py
```diff
@@ -41,72 +42,101 @@
...
     1-form, the vector entries at boundary nodes are set to the
     boundary condition values.
+
     """
     tensor = kwargs.pop("tensor", None)
-    return get_form_assembler(expr, tensor, *args, **kwargs)()
+    return Assembler(expr, *args, **kwargs).assemble(tensor=tensor)
+
 
+def Assembler(form, *args, **kwargs):
```
dham, line 132: `get_assembler`

firedrake/assemble.py
```diff
@@ -333,6 +333,7 @@ def allocate(self):
 
     @cached_property
     def maps_and_regions(self):
+        # TODO: Inspect the DAG and make tighter sparsity.
```
dham, line 336:
```
# The sparsity could be made tighter by inspecting the form DAG.
```

dham (changes requested):

firedrake/assemble.py
ksagiyam: Done.

firedrake/assemble.py
ksagiyam: Fixed.

firedrake/assemble.py
ksagiyam: Fixed.

ksagiyam:
Can we merge this and the linked PyOP2 PR? I need to drop branch pointer before merging.

firedrake/assemble.py
```diff
@@ -41,1018 +42,987 @@
...
-                    res += petsc_mat
+            petsc_mat = mat.petscmat
+            # Out-of-place Hermitian transpose
+            petsc_mat.hermitianTranspose(out=res)
+            (row, col) = mat.arguments()
+            return matrix.AssembledMatrix((col, row), self._bcs, res,
+                                          appctx=self._appctx,
+                                          options_prefix=self._options_prefix)
+        elif isinstance(expr, ufl.Action):
+            if (len(args) != 2):
```
connorjward, line 426:
```suggestion
            if len(args) != 2:
```

firedrake/assemble.py
```diff
@@ -41,1018 +42,987 @@
...
+        """Assemble the form.
 
-    @property
-    @abc.abstractmethod
-    def diagonal(self):
-        """Are we assembling the diagonal of a 2-form?"""
+        Parameters
+        ----------
+        tensor : firedrake.cofunction.Cofunction or matrix.MatrixBase
+            Output tensor to contain the result of assembly.
```
connorjward, line 956: Please explain what happens if tensor is none.

connorjward (changes requested):
I think the documentation needs a little clarifying but otherwise I'm happy to approve this.

firedrake/assemble.py
```diff
@@ -41,1018 +42,987 @@
...
-        children = dict.fromkeys(e for e in (expr.argument_slots() + expr.ufl_operands)
-                                 if isinstance(e, ufl.form.BaseForm))
-        return list(children)
-    return []
-
-
-def restructure_base_form_postorder(expression, visited=None):
-    visited = visited or {}
+    """
+    tensor = kwargs.pop("tensor", None)
```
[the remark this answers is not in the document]
ksagiyam: No. Added a new check for `args = ()`.

firedrake/assemble.py
```diff
@@ -41,1018 +42,987 @@
...
 
-             Action             Action               Action
-             /    \    (1)      /     \      (2)     /     \               (4)                                dFdN
-           Action  w  ---->  dFdN   Action  ---->  dFdN   dNdu(u; w, v*)  ---->  dFdN(..., dNdu(u; w, v*)) =    |
-           /    \                    /    \                                                                  dNdu(u; w, v*)
-         dFdN    dNdu              dNdu    w
+        Parameters
+        ----------
+        tensor : firedrake.cofunction.Cofunction or firedrake.function.Function or matrix.MatrixBase
+            Output tensor to contain the result of assembly.
```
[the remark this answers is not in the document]
ksagiyam: Fixed.

firedrake/assemble.py
```diff
@@ -41,1018 +42,987 @@
...
 
-    def visitor(expr):
-        # Perform the DAG restructuring when needed
-        return restructure_base_form(expr, visited)
 
-    expression = base_form_preorder_traversal(expression, visitor, visited)
-    # Need to reconstruct the expression at the end when all its operands have been visited!
-    operands = [visited.get(args, args) for args in base_form_operands(expression)]
-    return reconstruct_node_from_operands(expression, operands)
+class ExprAssembler:
```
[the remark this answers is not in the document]
ksagiyam: Fixed.

firedrake/assemble.py
ksagiyam: Fixed manually.

firedrake/assemble.py
ksagiyam: Fixed.

connorjward (dismissed):

Issue #3081, opened 2023-08-29
BUG: cythonizing firedrake/cython/dmcommon.pyx fails with cython 3.0+


francesco-ballarin:
**Describe the bug**
Cythonizing firedrake/cython/dmcommon.pyx fails with cython 3.0+

**Steps to Reproduce**
Upgrade cython to 3.0+ and try to install firedrake

**Expected behavior**
A successful cythonization

**Error message**
```
      [1/1] Cythonizing firedrake/cython/dmcommon.pyx
      warning: firedrake/cython/petschdr.pxi:8:0: The 'IF' statement is deprecated and will be removed in a future Cython version. Consider using runtime conditions or C macros instead. See https://github.com/cython/cython/issues/4310
      warning: firedrake/cython/petschdr.pxi:160:48: The keyword 'nogil' should appear at the end of the function signature line. Placing it before 'except' or 'noexcept' will be disallowed in a future version of Cython.
      warning: firedrake/cython/dmcommon.pyx:993:42: Buffer unpacking not optimized away.
      warning: firedrake/cython/dmcommon.pyx:993:42: Buffer unpacking not optimized away.
      warning: firedrake/cython/dmcommon.pyx:2265:4: Buffer unpacking not optimized away.
      warning: firedrake/cython/dmcommon.pyx:2265:4: Buffer unpacking not optimized away.
      warning: firedrake/cython/dmcommon.pyx:2657:4: Buffer unpacking not optimized away.
      warning: firedrake/cython/dmcommon.pyx:2657:4: Buffer unpacking not optimized away.
      warning: firedrake/cython/dmcommon.pyx:2658:4: Buffer unpacking not optimized away.
      warning: firedrake/cython/dmcommon.pyx:2658:4: Buffer unpacking not optimized away.
      
      Error compiling Cython file:
      ------------------------------------------------------------
      ...
                      dim1 -= 1
                      break
              else:
                  raise RuntimeError("Found inconsistent fiat_cone and plex_cone")
          assert dim1 == 0
          return (2**dim) * eo + io
                               ^
      ------------------------------------------------------------
      
      firedrake/cython/dmcommon.pyx:989:25: Cannot assign type 'double' to 'PetscInt'
      
      Error compiling Cython file:
      ------------------------------------------------------------
      ...
                      cfacets[k].global_u = v[1]
                      cfacets[k].global_v = v[0]
                  cfacets[k].local_facet = f
                  k += 1
[...]

PR #3290, opened 2023-12-11
Cython 3 and numpy 2

connorjward:
# Description

Fixes https://github.com/firedrakeproject/firedrake/issues/3081

It will also allow us to update our PETSc fork as petsc4py now requires cython 3.


scripts/firedrake-install
```diff
@@ -983,14 +982,10 @@ def run_pip_install(pipargs):
     # subprocesses wrote out.
     # Particularly important for debugging petsc fails.
     with environment(**blas):
-        pipargs = ["-vvv"] + pipargs
+        pipargs = ["-v"] + pipargs
```
JDBetteridge, line 985: :joy:

connorjward:
@Ig-dolci FYI I am having another stab at this now. If we use the release candidate for numpy 2.0 then we can hopefully land Cython 3 and numpy at the same time.

connorjward:
This is quite the pain, now [loopy is also having issues](https://github.com/inducer/loopy/issues/835).

scripts/firedrake-install
```diff
@@ -1347,8 +1334,9 @@ def build_and_install_torch():
         raise InstallError("CUDA installation is not available on MacOS.")
     if sys.version_info >= (3, 10):
         run_pip_install(["typing_extensions"])
-    extra_index_url = ["--extra-index-url", "https://download.pytorch.org/whl/cpu"] if args.torch == "cpu" else []
-    run_pip_install(["torch"] + extra_index_url)
+    # FIXME: use nightly to see if that makes numpy 2.0 happy
```
connorjward, line 1337: Should really be
```
pip install torch==2.3.0 --index-url https://download.pytorch.org/whl/test/cpu
```
as detailed in https://dev-discuss.pytorch.org/t/pytorch-release-2-3-0-final-rc-is-available/1995

2.3.0 is supposed to be released 24/04.

francesco-ballarin:
Just to double check: are the numpy changes here backward compatible with 1.x, or not?

cython 3.x is somewhat widespread now, while it will still take a while to get numpy 2.x out (at least, on installations that do not use firedrake-install)

connorjward:
> Just to double check: are the numpy changes here backward compatible with 1.x, or not?
> 
> cython 3.x is somewhat widespread now, while it will still take a while to get numpy 2.x widely used (at least, on installations that do not use firedrake-install)

I think so. The errors fixed here were mostly due to numpy dropping support for deprecated API.

francesco-ballarin:
Will this be closed in favor of #3546?

connorjward:
> Will this be closed in favor of #3546?

Yes.

PR #3293, opened 2023-12-14
Fix voting algorithm

### LINKED ISSUE #3151: BUG: Cross mesh interpolation breaks for parallel nested meshes
**Describe the bug**
Cross mesh interpolation fails for certain combinations of number of cells on the two meshes and number of MPI ranks.
We get the following error message:
```
RuntimeError: Some points have been claimed by a cell that we cannot see, but which we think we own. This should not happen.
```

**Steps to Reproduce**
1. Code/test/example that was run  '...' (as minimal as possible)
```
from firedrake import *
s = PETSc.COMM_WORLD.getSize()
nx = 2 * s
mx = 3 * nx
mh = [UnitCubeMesh(nx, nx, nx),
            UnitCubeMesh(mx, mx, mx)]

family = "Lagrange"
degree = 1
Vc = FunctionSpace(mh[0], family, degree=degree)
Vf = FunctionSpace(mh[1], family, degree=degree)

uc = Function(Vc)
uf = interpolate(uc, Vf)
print("success", flush=True)
```
2. Command executed '...'
The code breaks if we run with more than 2 MPI ranks
```
mpiexec -n 3 python bug.py
mpiexec -n 4 python bug.py
```

**Expected behavior**
Describe what you expected to happen, in the case of a regression,
include details of a setup where this behaviour was last seen.

**Error message**
Add error message with full backtrace, or log.
Please add these as text using three backticks (`) for highlighting.
Please do not add screenshots unless the bug is purely graphical.

**Environment:**
 - OS: [eg: Linux, MacOS, WSL (Windows Subsystem for Linux)] add this as a label too!
 - Python version: [eg: 3.9.7]
 - Output of `firedrake-status`
 - Any relevant environment variables or modifications [eg: PYOP2_DEBUG=1]

**Additional Info**

ReubenHill:
Fixes #3151

### DIFF
--- a/firedrake/mesh.py
+++ b/firedrake/mesh.py
@@ -2628,7 +2628,7 @@ values from f.)"""
                         not run at c-loop speed. */
                         /* cells_ignore has shape (npoints, ncells_ignore) - find the ith row */
                         int *cells_ignore_i = cells_ignore + i*ncells_ignore;
-                        cells[i] = locate_cell(f, &x[j], %(geometric_dimension)d, &to_reference_coords, &to_reference_coords_xtr, &temp_reference_coords, &found_reference_coords, &ref_cell_dists_l1[i], ncells_ignore, cells_ignore_i);
+                        cells[i] = locate_cell(f, &x[j], {self.geometric_dimension()}, &to_reference_coords, &to_reference_coords_xtr, &temp_reference_coords, &found_reference_coords, &ref_cell_dists_l1[i], ncells_ignore, cells_ignore_i);
 
                         for (int k = 0; k < {self.geometric_dimension()}; k++) {{
                             X[j] = found_reference_coords.X[k];


ReubenHill:
Todo:
- [ ] Work out why single parallel immersedsphere test is failing. It passes locally on my mac. Could this be because the second set of `locate_cell` calls are not done collectively?
- [ ] Check speed relative to old voting algorithm

ReubenHill:
For reference, the voting algorithm as implemented here is described in section 7.5 of my thesis 
[Nixon-Hill-RW-2024-PhD-Thesis.pdf.pdf](https://github.com/user-attachments/files/17341564/Nixon-Hill-RW-2024-PhD-Thesis.pdf.pdf)

connorjward:
I will take this on and try and get it merged. I have identified a couple of issues with this, and have some ideas for more scalable approaches, but I think this will fix the common failure case that people are seeing at the moment.

ReubenHill:
Thanks for taking this on. I hope the notes I made before leaving are
useful to you!

connorjward:
> Thanks for taking this on. I hope the notes I made before leaving are useful to you!

Yes thank you. I have also really appreciated your detailed comments in this part of the code!

It looks like tests are passing following a merge of `master` (and a tiny conflict) so hopefully this will get merged next Weds.

dham (approved):

PR #3295, opened 2023-12-16
Add optional variant kwarg to FunctionSpace()

pbrubeck:
# Description


rckirby (changes requested):
This is about what I had in mind.   I checked out the branch and it did what I expected for Lagrange and Raviart-Thomas.

I'd like to see some tests exercising the feature.  It might be sufficient to grep for places where `variant` is passed into a `FiniteElement` and then into a `FunctionSpace` and use this shortcut and make sure that the test suite still passes?

tests/regression/test_function_spaces.py
```diff
@@ -83,6 +83,11 @@ def test_function_space_different_family_differ(mesh):
     assert FunctionSpace(mesh, "CG", 1) != FunctionSpace(mesh, "DG", 1)
 
 
+def test_function_space_different_variant_differ(mesh):
+    "FunctionSpaces defined with different element families differ."
```
pbrubeck, line 87:
```suggestion
    "FunctionSpaces defined with different element variants differ."
```

rckirby (approved):

dham (approved):

PR #3297, opened 2023-12-19
Projection: Fix boundary conditions

pbrubeck:
# Description


dham (changes requested):
I think this looks right but can we have a test please?

pbrubeck:
Closing as boundary conditions were already properly handled by the `LinearSolver` within `Projector`. In case we are using SLATE, we can only have DG as the target space, so BCs are not relevant.

PR #3300, opened 2023-12-27
CG/DG integral variants

pbrubeck:
# Description
Add tests for integral (a.k.a. hierarchical) CG/DG simplex elements using integrated Jacobi polynomials implemented in https://github.com/firedrakeproject/fiat/pull/58


.github/workflows/build.yml
```diff
@@ -82,6 +82,7 @@ jobs:
             --install femlium \
             --install fascd \
             --install defcon \
+            --package-branch tsfc pbrubeck/cg-dg-integral \
```
rckirby, line 85: We need to put this back before merging

rckirby (changes requested):

PR #3304, opened 2024-01-04
install: unpin netgen-mesher

ksagiyam:
Fix CI.

### DIFF
--- a/scripts/firedrake-install
+++ b/scripts/firedrake-install
@@ -1425,7 +1425,7 @@ def build_and_install_slepc():
 
 def build_and_install_netgen():
     log.info("Installing NGSolve/Netgen.")
-    run_pip(["install", "netgen-mesher==6.2.2304.post142.dev0"])
+    run_pip(["install", "netgen-mesher"])
     log.info("Installing ngsPETSc.")
     url = "git+https://github.com/NGSolve/ngsPETSc.git@main"
     ngsPETSc_changed = False


connorjward:
Related discussion: https://github.com/firedrakeproject/firedrake/discussions/3299#discussioncomment-8004031

UZerbinati:
Why do you want to unpin the version of netgen rather than pinning to a new stable version that we know works with Firedrake ?

scripts/firedrake-install
```diff
@@ -1425,7 +1425,7 @@ def build_and_install_slepc():
 
 def build_and_install_netgen():
     log.info("Installing NGSolve/Netgen.")
-    run_pip(["install", "netgen-mesher==6.2.2304.post142.dev0"])
```
UZerbinati, line 1428: Why do you want to unpin the version of netgen rather than pinning to a new stable version that we know works with Firedrake ?

UZerbinati (commented):
Why do you want to unpin the version of netgen rather than pinning to a new stable version that we know works with Firedrake ?

scripts/firedrake-install
ksagiyam: We normally do not pin unless there is an issue. Do we have a reason to pin?

ksagiyam: I do not think we need to pin now. If we see some issue in the future, we might need to pin, but that is not a big deal.

UZerbinati (approved):

PR #3305, opened 2024-01-04
Dolci/enhance checkpoint manager

Ig-dolci:
# Description
I think this disk checkpointing management is more suitable.


firedrake/adjoint_utils/checkpointing.py
```diff
@@ -73,21 +73,22 @@ def enable_disk_checkpointing(dirname=None, comm=COMM_WORLD, cleanup=True):
 
 
 def disk_checkpointing():
-    """Return true if disk checkpointing of the adjoint tape is active."""
-    return _stop_disk_checkpointing <= 0
+    """Return whether disk checkpointing is enabled."""
+    global _enable_disk_checkpoint
```
connorjward, line 77:
```suggestion
```
I think that, since you aren't modifying the value, this line isn't needed.

firedrake/adjoint_utils/checkpointing.py
```diff
@@ -73,21 +73,22 @@ def enable_disk_checkpointing(dirname=None, comm=COMM_WORLD, cleanup=True):
...
     """Resume disk checkpointing."""
-    global _stop_disk_checkpointing
-    _stop_disk_checkpointing -= 1
-    return _stop_disk_checkpointing <= 0
+    global _enable_disk_checkpoint
+    _enable_disk_checkpoint = True
+    return _enable_disk_checkpoint
 
 
 class stop_disk_checkpointing(object):
```
connorjward, line 94: This is unrelated to your other changes, but I have some thoughts

```suggestion
class stop_disk_checkpointing:
```

Also, doesn't this have the same issue as `with stop_annotating():`? That is we can have code like
```py
# checkpointing enabled
with stop_disk_checkpointing():
    # checkpointing disabled
    with stop_disk_checkpointing():
        # checkpointing disabled
    # checkpointing enabled!
```

connorjward (changes requested):
This is definitely an improvement!

firedrake/adjoint_utils/checkpointing.py
dham: Yes, the same solution should be copied across from there.

firedrake/adjoint_utils/checkpointing.py
Ig-dolci: Yes. That is right.

PR #3314, opened 2024-01-10
Mesh Hierarchy support when working with Netgen mesh 2D

UZerbinati:
Support for 2D multigrid with uniform refinement.
Support for non-uniform degree of curvature.
Test Alfeld split (now via Netgen and not `PETSc DMPlexTransform`).
Support for Powell-Sabin split (no idea how to test any suggestion would be much appreciated).


firedrake/mg/mesh.py
```diff
@@ -9,6 +9,11 @@
 from firedrake.cython import mgimpl as impl
 from .utils import set_level
 
+try:
+    import netgen
+except ImportError:
+    netgen = None
+    ngsPETSc = None
```
connorjward, line 16: Shouldn't there be an `import ngsPETSc` inside the `try:`?

firedrake/mg/mesh.py
```diff
@@ -80,6 +85,7 @@ def __getitem__(self, idx):
 
 def MeshHierarchy(mesh, refinement_levels,
                   refinements_per_level=1,
+                  order=1,
```
connorjward, line 88: Please document this. And ideally switch the docstring to [numpydoc](https://numpydoc.readthedocs.io/en/latest/format.html) at the same time.

firedrake/mg/mesh.py
```diff
@@ -102,6 +108,13 @@ def MeshHierarchy(mesh, refinement_levels,
         callback receives the refined DM (and the current level).
     :arg mesh_builder: Function to turn a DM into a ``Mesh``. Used by pyadjoint.
     """
+    if netgen and isinstance(mesh, netgen.libngpy._meshing.Mesh):
+        try:
+            from ngsPETSc import NetgenHierarchy
+        except ImportError:
+            raise ImportError("Unable to import ngsPETSc. Please ensure that ngsolve is installed and available to Firedrake.")
```
connorjward, line 115: This error isn't descriptive enough. How do users do this?

firedrake/mg/mesh.py
```diff
@@ -102,6 +108,13 @@ def MeshHierarchy(mesh, refinement_levels,
...
+        return NetgenHierarchy(mesh, refinement_levels, order=order, digits=8, adaptive=False)
```
connorjward, line 116: I don't really understand what you have done. This completely avoids a lot of important Firedrake mesh logic. How does this work?

tests/multigrid/test_netgen_gmg.py
```diff
@@ -0,0 +1,38 @@
+from firedrake import *
+import pytest
+
+
+def test_netgen_mg_circle():
+    try:
+        from netgen.geom2d import Circle, CSG2d
+    except ImportError:
+        pytest.skip(reason="Netgen unavailable, skipping Netgen test.")
```
connorjward, line 9: We have a more standard way of doing this. See https://github.com/firedrakeproject/firedrake/blob/master/tests/conftest.py#L22.

tests/multigrid/test_netgen_gmg.py
```diff
@@ -0,0 +1,38 @@
...
+
+    f = 4+0*x
+    L = f*v*dx
+    exact = (1-x**2-y**2)
+
+    u = Function(V)
+    solve(a == L, u, bcs=bcs, solver_parameters={"ksp_type": "cg",
+                                                 "pc_type": "mg"})
+    expect = Function(V).interpolate(exact)
+    assert (norm(assemble(u - expect)) <= 1e-6)
```
connorjward, line 38:
```suggestion
    assert norm(assemble(u - expect)) <= 1e-6
```

connorjward (changes requested):
This seems like a significant feature. Some documentation is definitely needed.

firedrake/mg/mesh.py
dham: `degree`

dham (changes requested):
Needs parallel.
Probably needs a SMALL 3d test.
Definitely needs a parallel test.

firedrake/mg/mesh.py
UZerbinati: It should !

firedrake/mg/mesh.py
UZerbinati: Everything is hidden inside of ngsPETSc.
More detail here: https://github.com/NGSolve/ngsPETSc/blob/main/ngsPETSc/utils/firedrake.py

UZerbinati:
The PR only covers 2D for serial and parallel implementation.
The 3D case requires a bit of work on the Netgen, which needs to land in Netgen main, so is far in the time-line.
I think the 3D implementation deserves a separate PR.

firedrake/mg/mesh.py
```diff
@@ -9,6 +9,12 @@
 from firedrake.cython import mgimpl as impl
 from .utils import set_level
 
+try:
+    import netgen
+    import ngsPETSc
```
connorjward, line 14: Do we even need to import `ngsPETSc` here? We don't seem to use it. There is only a
```py
from ngsPETSc import NetgenHierarchy
```
below.

firedrake/mg/mesh.py
```diff
@@ -89,6 +96,9 @@ def MeshHierarchy(mesh, refinement_levels,
     :arg refinement_levels: the number of levels of refinement
     :arg refinements_per_level: the number of refinements for each
         level in the hierarchy.
+    :arg degree: the degree of the finite element space used for
```
connorjward, line 99: In line with our policy, please use modify this docstring to be in [numpydoc](https://numpydoc.readthedocs.io/en/latest/format.html) format.

firedrake/mg/mesh.py
```diff
@@ -102,6 +112,15 @@ def MeshHierarchy(mesh, refinement_levels,
         callback receives the refined DM (and the current level).
     :arg mesh_builder: Function to turn a DM into a ``Mesh``. Used by pyadjoint.
     """
+    if netgen and isinstance(mesh, netgen.libngpy._meshing.Mesh):
+        try:
+            from ngsPETSc import NetgenHierarchy
+        except ImportError:
+            raise ImportError("Unable to import ngsPETSc. Please ensure that netgen and ngsPETSc\
+                               installed and available to Firedrake. You can do this via \
```
connorjward, line 120:
```suggestion
                               are installed and available to Firedrake. You can do this via \
```

tests/regression/test_netgen.py
```diff
@@ -219,13 +214,139 @@ def test_firedrake_integral_sphere_high_order_netgen():
...
+
     try:
-        from netgen.geom2d import SplineGeometry
-        import netgen
+        from slepc4py import SLEPc
     except ImportError:
-        pytest.skip(reason="Netgen unavailable, skipping Netgen test.")
+        pytest.skip(reason="SLEPc unavailable, skipping adaptive test refinement.")
+
+    gc.collect()
```
connorjward, line 250: Why is this needed?

tests/regression/test_netgen.py
```diff
@@ -219,13 +214,139 @@ def test_firedrake_integral_sphere_high_order_netgen():
...
+    homsh = Mesh(msh.curve_field(2))
+    V = FunctionSpace(homsh, "CG", 2)
+    x, y, z = SpatialCoordinate(homsh)
+    f = Function(V).interpolate(1+0*x)
+    assert abs(assemble(f * dx) - (4/3)*np.pi) < 1.e-2
+
+
+@pytest.mark.skipnetgen
 @pytest.mark.skipcomplex
 def test_firedrake_Adaptivity_netgen():
```
connorjward, line 241: This test is very complicated. Can you add some descriptive comments? And perhaps give the test a more descriptive name?

connorjward (changes requested):

connorjward:
These are only minor comments. If you want this merged in the next week please bring it for review at the meeting today. It won't be merged, but if people are generally happy we can merge it once these comments are addressed.

connorjward:
> The PR only covers 2D for serial and parallel implementation. The 3D case requires a bit of work on the Netgen side, which needs to land in Netgen main, so is far in the timeline. I think the 3D implementation deserves a separate PR.

I assume that there is a clear `NotImplementedError` somewhere if someone tries to do this in 3D? I agree that it does not need to land in this PR.

firedrake/mg/mesh.py
```diff
@@ -80,28 +84,53 @@ def __getitem__(self, idx):
...
-        distribution, see :func:`~.Mesh` for details.  If ``None``,
-        use the same distribution parameters as were used to
-        distribute the coarse mesh, otherwise, these options override
-        the default.
-    :arg reorder: optional flag indicating whether to reorder the
-         refined meshes.
-    :arg callbacks: A 2-tuple of callbacks to call before and
+    Parameters
+    ----------
+    mesh : :py:func:`~.Mesh`
```
dham, line 95:
```suggestion
    mesh : Mesh
```

firedrake/mg/mesh.py
```diff
@@ -80,28 +84,53 @@ def __getitem__(self, idx):
...
         callback receives the refined DM (and the current level).
-    :arg mesh_builder: Function to turn a DM into a ``Mesh``. Used by pyadjoint.
+    mesh_builder
+        Function to turn a DM into a ``Mesh``. Used by pyadjoint.
+    Returns
+    -------
+    A :py:class:`HierarchyBase` object representing the
+    mesh hierarchy.
     """
+    if netgen and hasattr(mesh, 'netgen_mesh'):
```
dham, line 125: The issue with this is that behaviour silently changes depending on what is installed. I think we should get rid of the silent dispatch to NetgenHierarchy here and instead have a parameter to the MeshHeirarchy constructor to select whether to use a Firedrake intrinsic hierarchy or a netgen one. Choosing netgen when it's not installed would then trigger the error.

dham: It could then also be fatal to specify the degree if you're not using netgen.

tests/multigrid/test_netgen_gmg.py
```diff
@@ -0,0 +1,73 @@
+from firedrake import *
+import pytest
+
+
+@pytest.mark.skipcomplex
+@pytest.mark.skipnetgen
+def test_netgen_mg_circle():
+    from netgen.geom2d import Circle, CSG2d
+    geo = CSG2d()
```
dham, line 9: please replace the duplicated code with a function that you call from both tests.

dham (changes requested):
Relatively minor fixes required.

firedrake/mg/mesh.py
UZerbinati: If I do this the CI fails saying:
 ```
Warning, treated as error:
/__w/firedrake/firedrake/firedrake/mg/mesh.py:docstring of firedrake.mg.mesh.MeshHierarchy:1:py:class reference target not found: Mesh
```

Comment (connorjward) [reply firedrake/mg/mesh.py]: I think you want `MeshGeometry`. `Mesh` is only a function, not a class.

File: firedrake/mg/mesh.py [hunk 15]
```diff
@@ -80,28 +79,58 @@ def __getitem__(self, idx):
...
+    Parameters
+    ----------
+    mesh : MeshGeometry
+        the coarse mesh to refine
+    refinement_levels : int
+        the number of levels of refinement
+    refinements_per_level : int
+        the number of refinements for each level in the hierarchy.
+    netgen_flags : bool, dict
+        either a bool or a dictionray containing options for Netgen.
```
dham, line 97:
```suggestion
        either a bool or a dictionary containing options for Netgen.
```

connorjward (approved):

PR #3324, opened 2024-01-12
Make VTK optional

JDBetteridge:
# Description
Recent issues on ARM based Mac + the long standing issue where we have to distribute VTK wheels prompted making VTK an optional dependency.

### DIFF
--- a/firedrake/output/paraview_reordering.py
+++ b/firedrake/output/paraview_reordering.py
@@ -7,7 +7,8 @@ try:
     del vtkmodules.vtkCommonDataModel
 except ModuleNotFoundError as e:
     raise ModuleNotFoundError(
-        "Error importing vtkmodules, you may need to install by executing\n\t"
+        "Error importing vtkmodules. Firedrake does not install VTK by default, "
+        "you may need to install VTK by running\n\t"
         "pip install vtk"
     ) from e
 from vtkmodules.vtkCommonDataModel import (
--- a/firedrake/output/paraview_reordering.py
+++ b/firedrake/output/paraview_reordering.py
@@ -2,9 +2,27 @@ from tsfc.finatinterface import create_base_element
 import numpy as np
 from pyop2.utils import as_tuple
 
+try:
+    import vtkmodules.vtkCommonDataModel
+    del vtkmodules.vtkCommonDataModel
+except ModuleNotFoundError as e:
+    raise ModuleNotFoundError(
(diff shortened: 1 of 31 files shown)


colinjcotter:
Are you still going to distribute the wheels?

JDBetteridge:
Probably, although VTK are getting better at releasing up to date wheels: https://pypi.org/project/vtk/#history

firedrake/output/paraview_reordering.py
```diff
@@ -2,6 +2,14 @@
 import numpy as np
 from pyop2.utils import as_tuple
 
+try:
+    import vtkmodules.vtkCommonDataModel
+    del vtkmodules.vtkCommonDataModel
+except ModuleNotFoundError as e:
+    raise ModuleNotFoundError(
+        "Error importing vtkmodules, you may need to install by executing\n\t"
+        "pip install vtk"
```
connorjward, line 11: Is this going to be robust enough? What is VTK's release schedule like?

firedrake/output/paraview_reordering.py
```diff
@@ -2,6 +2,14 @@
 import numpy as np
 from pyop2.utils import as_tuple
 
+try:
+    import vtkmodules.vtkCommonDataModel
+    del vtkmodules.vtkCommonDataModel
+except ModuleNotFoundError as e:
+    raise ModuleNotFoundError(
+        "Error importing vtkmodules, you may need to install by executing\n\t"
```
connorjward, line 10: I feel like we could give the user a slightly better message. Something like:

"Cannot import vtkmodules. Firedrake does not install VTK by default so you may need to install it by running pip install vtk first."

I feel that the current message doesn't sound deliberate enough.

connorjward (changes requested):

firedrake/output/paraview_reordering.py
connorjward: I've just checked the link you shared above. I'm convinced. They seem to be releasing wheels every month or so and are supporting the latest Pythons.

connorjward (commented):
I am happy with this. @JDBetteridge believes that some bikeshedding/further API thought is required before merging.

dham (dismissed):
Please make vtk default on.

JDBetteridge:
I think this is ready, I think I want to kick the "move `checkpointing.py` into the `output` directory" can down the road.

connorjward (approved):

dham (approved):

PR #3327, opened 2024-01-15
add petsc EventDecorators to the same Cofunction methods as the decorated Function methods

JHopeCollins:
None of the `Cofunction` methods had `PETSc.Log.EventDecorator`.
This adds logging to the methods where the equivalent `Function` method has logging.


connorjward (approved):
Looks fine. Thanks.

PR #3344, opened 2024-01-22
Fix constant dimension

Ig-dolci:
# Description
Fix constant array dimension error.

### DIFF
--- a/firedrake/adjoint_utils/constant.py
+++ b/firedrake/adjoint_utils/constant.py
@@ -108,7 +108,7 @@ class ConstantMixin(OverloadedType):
         return self._constant_from_values()
 
     def _ad_dim(self):
-        return numpy.prod(self.dat.data_ro.ndim)
+        return self.dat.cdim
 
     def _ad_imul(self, other):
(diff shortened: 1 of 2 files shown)


firedrake/adjoint_utils/constant.py
```diff
@@ -108,7 +108,7 @@ def _ad_copy(self):
         return self._constant_from_values()
 
     def _ad_dim(self):
-        return numpy.prod(self.dat.data_ro.cdim)
+        return numpy.prod(self.dat.data_ro.ndim)
```
Ig-dolci, line 111: @JDBetteridge  is it `self.dat.data_ro.ndim` here?

JDBetteridge, line 111:
```suggestion
        return self.dat.cdim
```

JDBetteridge (changes requested):
30% sure...

connorjward (commented):
Could we get a simple test of this? Even just a test which is "this code should not raise an exception" might be suitable.
Then discussions of "I think this is the right thing to do" can be validated.

dham (approved):

PR #3347, opened 2024-01-24
Fix the condition for the result input

Ig-dolci:


firedrake/adjoint/ufl_constraints.py
```diff
@@ -96,7 +96,7 @@ def jacobian_adjoint_action(self, m, dp, result):
         asm = firedrake.assemble(
             dp * ufl.replace(self.dform, {self.trial: self.test})
         )
-        if isinstance(result, firedrake.Function):
+        if isinstance(result, firedrake.Cofunction):
```
Ig-dolci, line 99: @nbouziani We should impose the result as a`Cofunction`, right?

dham (approved):

PR #3354, opened 2024-01-25
Fix firedrake-update :grimacing: 

JDBetteridge:
# Description
`firedrake-install` works but update doesn't due to reliance on global state.


connorjward (approved):
This looks believable. I'll take you on your word that this is the right thing to do.

PR #3364, opened 2024-01-28
Merge upstreams

ksagiyam:
Update PETSc/SLEPc:
https://github.com/firedrakeproject/petsc/pull/18
https://github.com/firedrakeproject/slepc/pull/4

Latest PETSc does not allow for using `VecReplaceArray()`, which `VecPermute()` calls underneath, with the first `Vec` returned by `VecDuplicateVecs()` (see the link below for more details). As some PC classes need to work with such `Vec`s, we need to replace `VecPermute()` with a custom method, `_vec_permute()` that uses work vectors. This is probably better than using `VecPermute()` anyway as `VecPermute()` allocates/deallocates memory every time we call it.

### DIFF
--- a/firedrake/preconditioners/facet_split.py
+++ b/firedrake/preconditioners/facet_split.py
@@ -45,6 +45,8 @@ class FacetSplitPC(PCBase):
         _, P = pc.getOperators()
         appctx = self.get_appctx(pc)
         fcp = appctx.get("form_compiler_parameters")
+        self.work_vec_x = P.createVecLeft()
+        self.work_vec_y = P.createVecRight()
 
         prefix = pc.getOptionsPrefix()
         options_prefix = prefix + self._prefix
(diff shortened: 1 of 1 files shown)


ksagiyam:
https://gitlab.com/petsc/petsc/-/issues/1537#note_1747951337

firedrake/preconditioners/facet_split.py
```diff
@@ -102,7 +104,8 @@ def _permute_nullspace(nsp):
                     return nsp
                 vecs = [vec.duplicate() for vec in nsp.getVecs()]
                 for vec in vecs:
-                    vec.permute(self.iperm)
+                    self.work_vec_x.getArray(False)[:] = vec.getArray(True)[:]
```
dham, line 107: please pass these arguments by name.

dham (changes requested):
Trivial argument change, then please announce in #general when you actually do it.

PR #3372, opened 2024-01-30
Make `MassInvPC` subclass `AssembledPC`

pbrubeck:
# Description
Make `MassInvPC` subclass `AssembledPC`

### DIFF
--- a/demos/matrix_free/stokes.py.rst
+++ b/demos/matrix_free/stokes.py.rst
@@ -111,7 +111,7 @@ by some form of approximate commutator.::
 The mass inverse is dense, and therefore approximated with an incomplete
 LU factorization, which we configure now::
 
-      "fieldsplit_1_Mp_mat_type": "aij"
+      "fieldsplit_1_Mp_mat_type": "aij",
       "fieldsplit_1_Mp_pc_type": "ilu"
    }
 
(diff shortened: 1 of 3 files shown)


stephankramer:
Two comments:
* This fixes an issue with MassInvPC not being updated when the value of `mu` changes between solve calls (`update()` currently doesn't do anything) - which is nice
* This would change the options structure somewhat for the common case where an iterative solver is used because AssembledPC does not define a ksp directly under its prefix, it just does a pcapply. For example in tests/regression/test_nullspace.py
```
        'fieldsplit_1': {
            'ksp_type': 'fgmres',
            'pc_type': 'python',
            'pc_python_type': 'firedrake.MassInvPC',
            'Mp_ksp_type': 'cg',
            'Mp_pc_type': 'sor',
            'ksp_rtol': '1e-7',
            'ksp_monitor': None,
        }
```
would need to change to:
```
        'fieldsplit_1': {
            'ksp_type': 'fgmres',
            'pc_type': 'python',
            'pc_python_type': 'firedrake.MassInvPC',
            'Mp_pc_type': 'ksp',
            'Mp_ksp_ksp_type': 'cg',
            'Mp_ksp_pc_type': 'sor',
            'ksp_rtol': '1e-7',
            'ksp_monitor': None,
        }
```
(of the top of my head, untested) - so a little more wieldy.

dham (changes requested):
Fix tests and we're good.

dham (approved):

PR #3378, opened 2024-01-31
io: fix loading of extruded meshes

ksagiyam:
Fix loading of extruded meshes.

`extm._base_mesh.topology` must be object equal to `extm.topology._base_mesh`.


JDBetteridge (approved):

JDBetteridge:
Thanks for the fix!

JDBetteridge (approved):

PR #3386, opened 2024-02-04
FacetSplitPC: use local permutation

pbrubeck:
# Description
FacetSplitPC was broken in parallel

### DIFF
--- a/tests/regression/test_facet_split.py
+++ b/tests/regression/test_facet_split.py
@@ -2,12 +2,8 @@ import pytest
 from firedrake import *
 
 
-@pytest.mark.parametrize(['quadrilateral', 'ptype'],
-                         [(q, p)
-                          for q in [True, False]
-                          for p in ["lu", "jacobi"]])
-def test_facet_split(quadrilateral, ptype):
-    if ptype == "lu":
+def run_facet_split(quadrilateral, pc_type, refine=2):
+    if pc_type == "lu":
         parameters = {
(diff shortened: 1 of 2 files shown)


connorjward (changes requested):
It would be good to have a test so we can catch this in future. You should be able to reuse a existing FacetSplitPC test I expect.

tests/regression/test_facet_split.py
```diff
@@ -60,4 +55,17 @@ def test_facet_split(quadrilateral, ptype):
...
 
     solve(a == L, uh, bcs=bcs, solver_parameters=parameters)
-    assert sqrt(assemble(inner(uh - u_exact, uh - u_exact) * dx)) < 1E-10
+    return sqrt(assemble(inner(uh - u_exact, uh - u_exact) * dx))
+
+
+@pytest.mark.parametrize(['quadrilateral', 'pc_type'],
+                         [(q, p)
+                          for q in [True, False]
+                          for p in ["lu", "jacobi"]])
```
connorjward, line 64:
```suggestion
@pytest.mark.parametrize("quadrilateral", [True, False])
@pytest.mark.parametrize("pc_type", ["lu", "jacobi"])
```

connorjward (changes requested):
Thanks. Apart from the suggested style change I am happy to merge this.

connorjward (approved):

PR #3259, opened 2023-11-23
Expunge Expr.ufl_domain Part II

pbrubeck:
# Description
This was still triggering warnings at interpolation (and when trying to output non-Lagrange functions).

PR #3398, opened 2024-02-08
remove deprecated method.

Ig-dolci:
# Description

For `ufl.interpolate` object, we need to use ` ufl.domain.extract_unique_domain(self.form)` instead of `self.form.ufl_domain()`.  This fix warnings of deprecated `self.form.ufl_domain()` for `ufl.interpolate`.

### DIFF
--- a/firedrake/adjoint_utils/blocks/assembly.py
+++ b/firedrake/adjoint_utils/blocks/assembly.py
@@ -10,8 +10,13 @@ class AssembleBlock(Block):
     def __init__(self, form, ad_block_tag=None):
         super(AssembleBlock, self).__init__(ad_block_tag=ad_block_tag)
         self.form = form
-        mesh = self.form.ufl_domain() if hasattr(self.form, 'ufl_domain') \
-            else None
+        if hasattr(self.form, 'ufl_domain'):
+            if isinstance(self.form, ufl.Interpolate):
+                mesh = ufl.domain.extract_unique_domain(self.form)
+            else:
+                mesh = self.form.ufl_domain()
+        else:
+            mesh = None
 
         if mesh and not isinstance(self.form, ufl.Interpolate):
             # Interpolation differentiation wrt spatial coordinates is currently not supported.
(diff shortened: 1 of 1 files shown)


pbrubeck:
Thanks for this @Ig-dolci. My attempt https://github.com/firedrakeproject/firedrake/pull/3259 only left the warnings for the adjoint-related stuff, which I'm not very familiar. Could you also make sure the warning dissapears completely from the tests?
```
tests/regression/test_adjoint_operators.py: 12 warnings
  /__w/firedrake/firedrake_venv/src/ufl/ufl/core/expr.py:269: DeprecationWarning: Expr.ufl_domain() is deprecated, please use extract_unique_domain(expr) instead.
    warnings.warn("Expr.ufl_domain() is deprecated, please "
```

firedrake/adjoint_utils/blocks/assembly.py
```diff
@@ -10,8 +10,13 @@ class AssembleBlock(Block):
     def __init__(self, form, ad_block_tag=None):
         super(AssembleBlock, self).__init__(ad_block_tag=ad_block_tag)
         self.form = form
-        mesh = self.form.ufl_domain() if hasattr(self.form, 'ufl_domain') \
-            else None
+        if hasattr(self.form, 'ufl_domain'):
+            if isinstance(self.form, ufl.Interpolate):
+                mesh = ufl.domain.extract_unique_domain(self.form)
+            else:
+                mesh = self.form.ufl_domain()
```
pbrubeck, line 17: I think the correct thing should be something like `mesh = as_domain(form)`

Ig-dolci: After we have the warning message
```
DeprecationWarning: Expr.ufl_domain() is deprecated, please use extract_unique_domain(expr) instead.
    warnings.warn("Expr.ufl_domain() is deprecated, please "
```
The function [ufl_domain](https://github.com/firedrakeproject/ufl/blob/master/ufl/core/expr.py#L260) returns `extract_unique_domain(self)`. So, I am using `mesh = ufl.domain.extract_unique_domain(self.form)` because that is used at [ufl_domain](https://github.com/firedrakeproject/ufl/blob/master/ufl/core/expr.py#L260).

Ig-dolci: Do you think that is not the right way?

pbrubeck (approved):
This looks good. No more ufl_domain deprecation warnings from adjoint after getting rid of ufl_domain in other parts of this file.

Issue #3407, opened 2024-02-15
BUG: Assembling a `Cofunction` returns the `Cofunction` rather than a supplied `tensor`


jrmaddison:
**Describe the bug**
The usage `assemble(b, tensor=tensor)` with `b` and `tensor` `Cofunction`s returns `b` rather than `tensor`.

**Steps to Reproduce**
```
from firedrake import *
import ufl

mesh = UnitIntervalMesh(1)
space = FunctionSpace(mesh, "Discontinuous Lagrange", 0)
test = TestFunction(space)

form = ufl.conj(test) * dx
tensor = Cofunction(space.dual())

b0 = assemble(form)
b1 = assemble(b0, tensor=tensor)

assert b0 is b1  # Passes
assert b1 is tensor  # Fails
```

**Expected behavior**
The return value should be `tensor`.

**Additional Info**
Can lead to subtle bugs if the return value is modified.

connorjward:
@nbouziani can you tackle this?

ksagiyam:
Opened a PR https://github.com/firedrakeproject/firedrake/pull/3408.

nbouziani:
Thanks for spotting this. Note that passing `tensor` to `assemble` specifically means putting the result of the assembly into the tensor, so in practice you can just write:

```
assemble(b0, tensor=tensor)
```

PR #3408, opened 2024-02-15
assemble: return tensor if given

ksagiyam:
Fix https://github.com/firedrakeproject/firedrake/issues/3407.

### DIFF
--- a/firedrake/assemble.py
+++ b/firedrake/assemble.py
@@ -416,7 +416,9 @@ def assemble_base_form(expression, tensor=None, bcs=None,
 
     if tensor:
         update_tensor(result, tensor)
-    return result
+        return tensor
+    else:
+        return result
 
 
 def preprocess_base_form(expr, mat_type=None, form_compiler_parameters=None):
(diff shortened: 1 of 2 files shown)


connorjward (changes requested):
Thanks! Please provide a unit test.

firedrake/assemble.py
```diff
@@ -416,7 +416,9 @@ def visitor(e, *operands):
 
     if tensor:
         update_tensor(result, tensor)
-    return result
+        return tensor
+    else:
+        return result
```
nbouziani, line 421: That won't work because the tensor is not updated during the DAG traversal, that is precisely why we have this `update_tensor` stage at the end which copies the value of the result into the tensor.

nbouziani, line 421: I think this should work:

```suggestion
        return tensor
    return result
```

connorjward: Isn't this completely equivalent?

nbouziani: Oh I misread that. I thought `update_tensor` had been removed

nbouziani (approved):

PR #3412, opened 2024-02-17
FacetSplitPC: use global permutation to construct virtual submatrix

pbrubeck:
# Description
Since `Vec.permute()` is gone, `FacetSplitPC` had to permute vectors with local indices. The virtual submatrix had to keep the global permutation, and I forgot to do that in #3386. I also added a test.

### DIFF
--- a/firedrake/preconditioners/facet_split.py
+++ b/firedrake/preconditioners/facet_split.py
@@ -255,7 +255,7 @@ def get_permutation_map(V, W):
 
     kernel_code = f"""
     void permutation(PetscInt *restrict v, const PetscInt *restrict w) {{
-        for (int i=0; i < {size}; i++) v[i] = w[i];
+        for (PetscInt i=0; i<{size}; i++) v[i] = w[i];
     }}"""
     kernel = op2.Kernel(kernel_code, "permutation", requires_zeroed_output_arguments=False)
     op2.par_loop(kernel, V.mesh().cell_set,
--- a/firedrake/preconditioners/facet_split.py
+++ b/firedrake/preconditioners/facet_split.py
@@ -77,7 +77,7 @@ class FacetSplitPC(PCBase):
         W = FunctionSpace(V.mesh(), MixedElement([restrict(V.ufl_element(), d) for d in ("interior", "facet")]))
         assert W.dim() == V.dim(), "Dimensions of the original and decomposed spaces do not match"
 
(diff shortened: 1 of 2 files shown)


firedrake/preconditioners/facet_split.py
```diff
@@ -246,25 +248,18 @@ def get_permutation_map(V, W):
...
-    void permutation(PetscInt *restrict x,
-                     const PetscInt *restrict xi,
-                     const PetscInt *restrict xf){{
-        for(PetscInt i=0; i<{sizes[0]}; i++) x[i] = xi[i];
-        for(PetscInt i=0; i<{sizes[1]}; i++) x[i+{sizes[0]}] = xf[i];
-        return;
-    }}
-    """
+    void permutation(PetscInt *restrict v, const PetscInt *restrict w) {{
+        for (int i=0; i < {size}; i++) v[i] = w[i];
```
pbrubeck, line 258:
```suggestion
        for (PetscInt i=0; i<{size}; i++) v[i] = w[i];
```

connorjward (approved):

wence-:
> Since Vec.permute() is gone

`VecPermute` still appears to exist in the petsc C docs, did it go for bad reasons?

pbrubeck:
> > Since Vec.permute() is gone
> 
> `VecPermute` still appears to exist in the petsc C docs, did it go for bad reasons?

Sorry, my mistake. It is not that Vec.permute() is gone, the issue is a bit more subtle: see here https://github.com/firedrakeproject/firedrake/pull/3364

As I understand, Vec.permute is not done in-place, it allocates and deallocates memory, so the approach we take is to keep auxiliary Vecs that are only allocated once.

pbrubeck:
All of this could be written more cleanly if we were able to impose an ordering on the entity dofs (first interior, then facets) when we create a FunctionSpace.

wence-:
> All of this could be written more cleanly if we were able to impose an ordering on the entity dofs (first interior, then facets) when we create a FunctionSpace.

Some facility for this already exists permuting at the level of `cell->node` map (because it's need to permute entries for symmetry group members). I assume that this permutation that you want is not one that can be provided element-by-element in the FIAT element? Instead it's a per-process reordering of the dofs?

I don't think in principle there would be a problem with providing an alternative renumbering scheme when constructing function spaces, you'd just need to make sure that the caching of "equivalent" numberings takes it into account.

PR #3422, opened 2024-02-22
Fix ASMExtrudedStarPC: include vertical edge and face patches

pbrubeck:
# Description

Adding vertical edge/face patches to `ASMExtrudedStarPC`, which were missing when `pc_star_construct_dim` is equal to 1 or 2. Also include `ASMExtrudedStarPC` solver parameters on an exisisting test to verify that such a space decomposition is equivalent to Jacobi relaxation.

The tests were updated to terminate after a fixed number of KSP iterations, only checking for convergence outside PETSc.

### DIFF
--- a/firedrake/preconditioners/asm.py
+++ b/firedrake/preconditioners/asm.py
@@ -112,13 +112,9 @@ class ASMPatchPC(PCBase):
         self.asmpc.view(viewer=viewer)
 
     def update(self, pc):
-        # This is required to update an inplace ILU/ICC factorization
-        try:
-            for sub in self.asmpc.getASMSubKSP():
-                sub.getOperators()[0].setUnfactored()
-        except PETSc.Error:
-            pass
-        self.asmpc.setUp()
+        # This is required to update an inplace ILU factorization
+        for sub in self.asmpc.getASMSubKSP():
+            sub.getOperators()[0].setUnfactored()
 
     def apply(self, pc, x, y):
         self.asmpc.apply(x, y)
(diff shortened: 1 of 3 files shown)


firedrake/preconditioners/asm.py
```diff
@@ -395,54 +408,79 @@ def get_patches(self, V):
...
         # Build index sets for the patches
         ises = []
-        start, end = mesh_dm.getDepthStratum(depth)
-        pstart, _ = mesh_dm.getChart()
-        for seed in range(start, end):
-            # Only build patches over owned DoFs
-            if mesh_dm.getLabelValue("pyop2_ghost", seed) != -1:
+        # Build a base_depth-star on the base mesh and extrude it by an
+        # interval_depth-star on the interval mesh such that the depths sum to depth
+        # and 0 <= base_depth <= dim, 0 <= interval_depth <= 1.
```
pbrubeck, line 413:
```suggestion
        # and 0 <= interval_depth <= 1.
```

firedrake/preconditioners/asm.py
```diff
@@ -110,6 +112,12 @@ def view(self, pc, viewer=None):
         self.asmpc.view(viewer=viewer)
 
     def update(self, pc):
+        # This is required to update an inplace ILU/ICC factorization
+        try:
+            for sub in self.asmpc.getASMSubKSP():
+                sub.getOperators()[0].setUnfactored()
+        except PETSc.Error:
```
connorjward, line 119: This seems like a vague `except`. Also what if `self.asmpc.getASMSubKSP()` has 2 entries and the first errors but the second would not?

pbrubeck: I cannot reproduce the conditions to get that PETSc exception, it was a while ago, it could be that I fixed this within PETSc. But the important thing is that setUnfactored must be called. Ideally it should happen in PETSc, but not sure where. Another important detail is that we don't need to call `self.asmpc.setUp()` here as PETSc should know that this should happen if the matrix is reassambled.

connorjward (approved):

PR #3435, opened 2024-03-03
ASMVankaPC: fix tests

### LINKED ISSUE #3430: BUG: useful test is xfailed
**Describe the bug**
When #2297 was merged, it includes an xfail on `test_vanka_equivalence` in `tests/regression/test_star_pc.py`, removing the only test on important functionality.

**Steps to Reproduce**
1. Merge #3422 (but not #3429)
2. Run anything that uses `ASMVankaPC`

**Expected behavior**
Code should run.

**TODO**
We should fix this test, so that important functionality is actually tested on CI.  Probably easiest is to revisit the problem that we're solving to avoid difficult-to-reproduce issues with LU factorization.

### LINKED ISSUE #3431: Improve testing of ASMPatchPC
**Describe the current issue**
There's a lot of functionality in `ASMPatchPC` that's been added in the past year, but this functionality is rather sparsely tested.

**Describe the solution you'd like**
More tests!

**Describe alternatives you've considered**
Sadness and frustration.

**Additional info**
It's not entirely clear what a comprehensive set of tests should look like here.  But, as this code gets more use, it certainly needs a more robust set of tests.

pbrubeck:
# Description

Restore/fix ASMVankaPC tests, fixes #3430

Test `-pc_star_mat_ordering_type`, fixes #3431

We should expect unpivoted LU to work on a 2×2 block matrix with the natural saddle-point ordering (velocity followed by pressure). This is because (block) Gaussian elimination in the natural ordering should form the (LU factors of the) pressure Schur complement (which should be invertible). Nevertheless, we've been having plenty of issues with singular pivots throughout the years.

By default, petsc imposes a nested dissection reordering which messes up the block structure of the Vanka patches. Traditionally, this has been fixed by forcing a nonzero pivot or relying on sophisticated direct solver packages (using mumps on many small matrices was causing some issues here). The proposed fix is to keep the natural ordering of the fields and to use PETSc's own LU/Cholesky implementation.

### DIFF
--- a/firedrake/preconditioners/asm.py
+++ b/firedrake/preconditioners/asm.py
@@ -227,18 +227,20 @@ class ASMVankaPC(ASMPatchPC):
 
             # Create point list from mesh DM
             star, _ = mesh_dm.getTransitiveClosure(seed, useCone=False)
-            pt_array = set()
-            for pt in star.tolist():
+            star = order_points(mesh_dm, star, ordering, self.prefix)
+            pt_array = []
+            for pt in reversed(star):
                 closure, _ = mesh_dm.getTransitiveClosure(pt, useCone=True)
-                pt_array.update(closure.tolist())
+                pt_array.extend(closure)
+            # Grab unique points with stable ordering
+            pt_array = list(reversed(dict.fromkeys(pt_array)))
 
-            pt_array = order_points(mesh_dm, list(pt_array), ordering, self.prefix)
             # Get DoF indices for patch
             indices = []
             for (i, W) in enumerate(V):
                 section = W.dm.getDefaultSection()
                 if i in exclude_subspaces:
-                    loop_list = [seed]
+                    loop_list = star
                 else:
                     loop_list = pt_array
                 for p in loop_list:
--- a/firedrake/preconditioners/asm.py
+++ b/firedrake/preconditioners/asm.py
@@ -50,8 +50,8 @@ class ASMPatchPC(PCBase):
         asmpc.setOptionsPrefix(self.prefix + "sub_")
         asmpc.setOperators(*pc.getOperators())
 
-        backend = PETSc.Options().getString(self.prefix + "backend",
-                                            default="petscasm").lower()
+        opts = PETSc.Options(self.prefix)
+        backend = opts.getString("backend", default="petscasm").lower()
         # Either use PETSc's ASM PC or use TinyASM (as simple ASM
         # implementation designed to be fast for small block sizes).
         if backend == "petscasm":
@@ -69,7 +69,7 @@ class ASMPatchPC(PCBase):
             mat_type = P.getType()
             if not mat_type.endswith("sbaij"):
                 sentinel = object()
-                ordering = PETSc.Options().getString(self.prefix + "mat_ordering_type", default=sentinel)
+                ordering = opts.getString("mat_ordering_type", default=sentinel)
                 asmpc.setASMSortIndices(ordering is sentinel)
 
             lgmap = V.dof_dset.lgmap
@@ -146,14 +146,12 @@ class ASMStarPC(ASMPatchPC):
             warning("applying ASMStarPC on an extruded mesh")
 
         # Obtain the topological entities to use to construct the stars
-        depth = PETSc.Options().getInt(self.prefix+"construct_dim", default=0)
-        ordering = PETSc.Options().getString(self.prefix+"mat_ordering_type",
-                                             default="natural")
+        opts = PETSc.Options(self.prefix)
+        depth = opts.getInt("construct_dim", default=0)
+        ordering = opts.getString("mat_ordering_type", default="natural")
         # Accessing .indices causes the allocation of a global array,
         # so we need to cache these for efficiency
-        V_local_ises_indices = []
-        for (i, W) in enumerate(V):
-            V_local_ises_indices.append(V.dof_dset.local_ises[i].indices)
+        V_local_ises_indices = tuple(iset.indices for iset in V.dof_dset.local_ises)
 
         # Build index sets for the patches
         ises = []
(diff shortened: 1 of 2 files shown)


connorjward (commented):
I'm happy with this from a style perspective. @ScottMacLachlan I think if you approve this we can merge it.

firedrake/preconditioners/asm.py
```diff
@@ -229,18 +227,20 @@ def get_patches(self, V):
...
+            pt_array = list(reversed(dict.fromkeys(pt_array)))
 
-            pt_array = order_points(mesh_dm, list(pt_array), ordering, self.prefix)
             # Get DoF indices for patch
             indices = []
             for (i, W) in enumerate(V):
                 section = W.dm.getDefaultSection()
                 if i in exclude_subspaces:
-                    loop_list = [seed]
+                    loop_list = star
```
ScottMacLachlan, line 243: I agree that `star` is generally better than `[seed]` here, but I worry that this is actually a bigger change than is needed for this PR.  I also see reasons for having both options available.  Can I propose that we revert this change, and offer to address this in a separate PR (adding `exclude_subspace_type` or similar to specify behaviour)?

tests/regression/test_star_pc.py
```diff
@@ -168,6 +165,8 @@ def test_star_equivalence(problem_type, backend):
                        "mg_coarse_assembled_pc_type": "lu"}
 
     star_params["mg_levels_pc_star_backend"] = backend
+    star_params["mg_levels_pc_star_mat_ordering_type"] = "rcm"
+    star_params["mg_levels_pc_star_sub_sub_pc_factor_mat_ordering_type"] = "natural"
```
ScottMacLachlan, line 169: I had to think a bit to understand what's going on here (that when we specify a `pc_star_mat_ordering_type`, we don't want PETSc's defaults to screw it up later on -- right?).  Is this the best solution?  Could we either default to `sub_pc_factor_mat_ordering_type="natural"` when we set `sub_pc_type` to `lu` in the class defaults, or set this if we find `pc_star_mat_ordering_type` to be set in the options?

ScottMacLachlan (commented):
Overall, looks good, but a couple of small things to think about

firedrake/preconditioners/asm.py
pbrubeck: For high order Taylor-Hood, excluding subspaces like this and not applying any other relaxation on the pressure would be catastrophic, the preconditioner would be singular. Is the common thing to combine this with another relaxation on the pressure?

tests/regression/test_star_pc.py
pbrubeck: I am more inclined to have `sub_pc_factor_mat_ordering_type="natural"` as default, regardless of whether or not we set `pc_star_mat_ordering_type`. If no ordering is set then the DOFs in the patches get sorted in the local ordering, I think this should keep the ordering by fields.

firedrake/preconditioners/asm.py
ScottMacLachlan: I completely agree that it would be catastrophic.  You can, however, stitch together either single-pressure-dof patches (for P2/Q2) or vertex/edge/face/element pressure patches in a composite preconditioner for Taylor-Hood, and it works just fine.  I would agree that `star` should become the default here, but I'd like to have both families be accessible.

tests/regression/test_star_pc.py
ScottMacLachlan: I'm quite happy with that - it makes sense, and can always be overridden by a knowledgable user.

firedrake/preconditioners/asm.py
pbrubeck: Are you suggesting Vanka patches on V*Q constructed around vertices and only including the vertex DOF of Q + Edge-star patches on V*Q including edge, faces, and interiors of both V and Q? This sounds like a bit more expensive than Vanka patches on V*Q where we include the DOFs of Q from the star (but not it's closure).

dham (dismissed):

firedrake/preconditioners/asm.py
```diff
@@ -61,14 +61,15 @@ def initialize(self, pc):
             opts = PETSc.Options(asmpc.getOptionsPrefix())
```
ScottMacLachlan, line 61: I'm a bit concerned/confused by the double-definition of `opts` here, one with `self.prefix` and one with `asmpc.getOptionsPrefix()`.  Since the latter is `self.prefix + "sub_"`, I don't think the two lines below defining `ordering` are equivalent.  Are they?

pbrubeck: You are right at pointing this naming conflict. But I think `asmpc.setOptionsPrefix(self.prefix + "sub_")` was a mistake. This is why we need to pass the infamous `star_sub_sub_` prefix, and I think we could be fine if we just set `asmpc.setOptionsPrefix(self.prefix)` as we have in patch.py.

ScottMacLachlan: This looks like cleaner code now - thanks!  Do we want to resolve the `star_sub_sub_` here or in another PR or not at all?

pbrubeck: I'll open another PR, this would definitely break every code that passes options to set the patch solvers!

firedrake/preconditioners/asm.py
```diff
@@ -50,25 +50,26 @@ def initialize(self, pc):
...
+        backend = opts.getString("backend", default="petscasm").lower()
         # Either use PETSc's ASM PC or use TinyASM (as simple ASM
         # implementation designed to be fast for small block sizes).
         if backend == "petscasm":
             asmpc.setType(asmpc.Type.ASM)
             # Set default solver parameters
             asmpc.setASMType(PETSc.PC.ASMType.BASIC)
-            opts = PETSc.Options(asmpc.getOptionsPrefix())
+            sub_opts = PETSc.Options(asmpc.getOptionsPrefix())
             if "sub_pc_type" not in opts:
```
ScottMacLachlan, line 62: This should be `not in sub_opts`, shouldn't it?  And below?

ScottMacLachlan (changes requested):
One last(?) thing to fix here, I think, then it's good to go!  Many thanks!

connorjward (approved):

PR #3440, opened 2024-03-05
DO NOT MERGE - upstream update test

JHopeCollins:
Test updated loopy with deprecated numpy bool type removed.
Should prevent deprecation warnings being raised.

Associated loopy PR:
https://github.com/firedrakeproject/loopy/pull/22


connorjward:
Ah thanks for this! I have been meaning to update our fork for ages. Generally a newer loopy makes codegen faster for us.

PR #3442, opened 2024-03-06
assemble: support back python3.8

ksagiyam:
```
@classmethod
@property
def ...
    ...
```
is only supported by python >= 3.9.

### DIFF
--- a/firedrake/assemble.py
+++ b/firedrake/assemble.py
@@ -1349,7 +1349,7 @@ class ExplicitMatrixAssembler(ParloopFormAssembler):
             for local_kernel in self._all_local_kernels:
                 i, j = local_kernel.indices
                 # Make Sparsity independent of _iterset, which can be a Subset, for better reusability.
-                get_map, region = ExplicitMatrixAssembler.integral_type_op2_map[local_kernel.kinfo.integral_type]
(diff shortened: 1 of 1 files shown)


firedrake/assemble.py
```diff
@@ -1365,16 +1365,16 @@ def _make_maps_and_regions_default(test, trial, allocation_integral_types):
...
-            get_map, region = ExplicitMatrixAssembler.integral_type_op2_map[integral_type]
+            get_map, region = ExplicitMatrixAssembler.integral_type_op2_map()[integral_type]
             for i, rmap_ in enumerate(get_map(test)):
                 for j, cmap_ in enumerate(get_map(trial)):
                     maps_and_regions[(i, j)][(rmap_, cmap_)].add(region)
         return {block_indices: [map_pair + (tuple(region_set), ) for map_pair, region_set in map_pair_to_region_set.items()]
                 for block_indices, map_pair_to_region_set in maps_and_regions.items()}
 
     @classmethod
-    @property
```
connorjward, line 1376: I could be wrong, but I think this works if you swap the order of these around:
```py
@property
@classmethod
def XXX(cls):
    ...
```

connorjward (approved):

PR #3348, opened 2024-01-24
JDBetteridge/petsc priorities

JDBetteridge:
# Description
Internal refactor of how we set PETSc defaults. This is now based off which PETSc external packages are installed. Now a user with an external PETSc with _only_:
- MPI
- HDF5
- A direct solver (but I think for tests to pass this _must_ currently be MUMPS)
- A partitioner (even the built in PETSc "simple" partitioner)
- An AMG solver (including PETSc's gamg)

should expect the Firedrake test suite to pass.

Moving forward tests that use optional dependencies, such as:
- Netgen
- SLEPc
- Any PETSc external dependency (that isn't MPI or HDF5)
- Matplotlib
- VTK
- PyTorch

Should check for the presence of the optional dependency before being included in the Firedrake test suite.

PR #3449, opened 2024-03-12
JDBetteridge/petsc testing

JDBetteridge:
# Description
DO NOT MERGE
Test removing:
- chaco
- superlu_dist
- mumps and scalapack
- hypre
- others...


firedrake/petsc.py
```diff
@@ -334,3 +370,70 @@ def garbage_view(obj: Any):
...
+# > ICNTL(14) controls the percentage increase in the estimated working space...
+# > ... Remarks: When significant extra fill-in is caused by numerical pivoting, increasing ICNTL(14) may help.
+if DEFAULT_DIRECT_SOLVER == "mumps":
+    DEFAULT_DIRECT_SOLVER_PARAMETERS["mat_mumps_icntl_14"] = 200
+
+# Setup default AMG preconditioner
+amg_priority = ["hypre", "ml"]
+for amg in amg_priority:
+    if amg in external_packages:
+        DEFAULT_AMG_PC = amg
```
connorjward, line 419: missing a `break`.

firedrake/petsc.py
```diff
@@ -334,3 +370,70 @@ def garbage_view(obj: Any):
...
+    )
+    DEFAULT_DIRECT_SOLVER = "petsc"
+    DEFAULT_DIRECT_SOLVER_PARAMETERS = {"mat_solver_type": "petsc"}
+
+# MUMPS needs an additional parameter set
+# From the MUMPS documentation:
+# > ICNTL(14) controls the percentage increase in the estimated working space...
+# > ... Remarks: When significant extra fill-in is caused by numerical pivoting, increasing ICNTL(14) may help.
+if DEFAULT_DIRECT_SOLVER == "mumps":
+    DEFAULT_DIRECT_SOLVER_PARAMETERS["mat_mumps_icntl_14"] = 200
```
connorjward, line 413: I don't really like this pattern. We have constants here that aren't really constants. I would probably prefer this to be called `DEFAULT_MUMPS_SOLVER_PARAMETERS`. That or make this a function `get_default_direct_solver_options("mumps")`.

I also think that if users are advanced enough to be playing with solver options then we shouldn't obfuscate things behind ambiguous variables. Perhaps this sort of thing could be put into `conftest.py` as a fixture so we get the right behaviour for the tests but not touch the user experience?

firedrake/petsc.py
JDBetteridge: This is the wrong PR see https://github.com/firedrakeproject/firedrake/pull/3348

firedrake/petsc.py
JDBetteridge: This is the wrong PR see https://github.com/firedrakeproject/firedrake/pull/3348

JDBetteridge:
This PR is out of date and is marked DO NOT MERGE

PR #3426, opened 2024-02-26
Added in a RestrictedFunctionSpace class

emmarothwell1:
Add in a RestrictedFunctionSpace class.

This marks boundary nodes in the definition of the FunctionSpace with the goal of removing them from calculations such as in assembly. 
This currently only works for FunctionSpace objects, restricted mixed spaces and restrictions based on already defined RestrictedFunctionSpaces are still being worked on elsewhere. 
There are tests in tests/regression/test_restricted_function_space.py to check assembly and solve works. 
There is also a linked PyOP2 pull request at: [https://github.com/OP2/PyOP2/pull/716](url)

PR #3455, opened 2024-03-15
halo: manually construct halo exchange sf

ksagiyam:
Our current approach of making halo exchange sf using `dm.createSectionSF()` is not suitable if constrained DoFs exist as the resulting sf does not include constrained DoFs in roots/leaves, which could lead to inconsistent data.

### DIFF
--- a/firedrake/cython/dmcommon.pyx
+++ b/firedrake/cython/dmcommon.pyx
@@ -3481,6 +3481,7 @@ def create_halo_exchange_sf(PETSc.DM dm):
     point_sf.bcastBegin(unit, local_offsets, remote_offsets, MPI.REPLACE)
     point_sf.bcastEnd(unit, local_offsets, remote_offsets, MPI.REPLACE)
     n = 0
+    # ilocal == NULL if local leaf points are [0, 1, 2, ...).
     for i in range(nleaves):
         p = ilocal[i] if ilocal != NULL else i
         CHKERR(PetscSectionGetDof(local_sec.sec, p, &dof))
--- a/firedrake/cython/dmcommon.pyx
+++ b/firedrake/cython/dmcommon.pyx
@@ -3211,48 +3211,6 @@ def make_global_numbering(PETSc.Section lsec, PETSc.Section gsec):
     return val
 
 
-def prune_sf(PETSc.SF sf):
-    """Prune an SF of roots referencing the local rank
-
-    :arg sf: The PETSc SF to prune.
-    """
-    cdef:
-        PetscInt nroots, nleaves, new_nleaves, i, j
-        PetscInt rank
-        const PetscInt *ilocal = NULL
-        PetscInt *new_ilocal = NULL
-        const PetscSFNode *iremote = NULL
-        PetscSFNode *new_iremote = NULL
-        PETSc.SF pruned_sf
-
-    CHKERR(PetscSFGetGraph(sf.sf, &nroots, &nleaves, &ilocal, &iremote))
-
-    rank = sf.comm.rank
-    new_nleaves = 0
-    for i in range(nleaves):
-        if iremote[i].rank != rank:
-            new_nleaves += 1
-
-    CHKERR(PetscMalloc1(new_nleaves, &new_ilocal))
-    CHKERR(PetscMalloc1(new_nleaves, &new_iremote))
-    j = 0
-    for i in range(nleaves):
-        if iremote[i].rank != rank:
-            if ilocal != NULL:
-                new_ilocal[j] = ilocal[i]
-            else:
-                new_ilocal[j] = i
-            new_iremote[j].rank = iremote[i].rank
-            new_iremote[j].index = iremote[i].index
-            j += 1
-
-    pruned_sf = PETSc.SF().create(comm=sf.comm)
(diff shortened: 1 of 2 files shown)


wence-:
I am out of the loop here, but how are the constrained dofs represented (in the section or otherwise)?

ksagiyam:
In the section. This is a preparation PR for https://github.com/firedrakeproject/firedrake/pull/3426.

firedrake/halo.py
```diff
@@ -103,17 +103,7 @@ def __init__(self, dm, section, comm):
 
     @utils.cached_property
     def sf(self):
-        lsec = self.dm.getDefaultSection()
-        gsec = self.dm.getDefaultGlobalSection()
-        self.dm.createDefaultSF(lsec, gsec)
```
wence-, line 108: Could you instead create a global section that includes the constraint dofs via `PetscSectionCreateGlobalSection` and then continue with `dm.createDefaultSF(lsec, gsec)`? (Although I see the petsc4py wrapper needs to expose the `includeConstraints` argument first)

ksagiyam: The global section not including the constraint DoFs is actually the one which represents the layout of the reduced (shorter) global vector (that we use in solve), so I think it is the one our section dm should carry. Or did you mean we could make a new global section that includes constraint DoFs when we make halo exchange sf?

By the way, the idea in this PR https://github.com/firedrakeproject/firedrake/pull/3426 is that we compute `renumbering` so that the owned constrained DoFs will be pushed to the end of the "core + owned" block so that we can make reduced global vector by concatenating the "core + owned - constrained" block. We do not care where constrained DoFs appear in the halo.

ksagiyam: Yes, that is the idea.

firedrake/cython/dmcommon.pyx
```diff
@@ -3471,3 +3429,76 @@ def to_petsc_local_numbering(PETSc.Vec vec, V):
...
+    assert pEnd - pStart == nroots, f"pEnd - pStart ({pEnd - pStart}) != nroots ({nroots})"
+    assert pStart == 0
+    m = 0
+    local_offsets = np.empty(pEnd - pStart, dtype=IntType)
+    remote_offsets = np.empty(pEnd - pStart, dtype=IntType)
+    for p in range(pStart, pEnd):
+        CHKERR(PetscSectionGetDof(local_sec.sec, p, &dof))
+        CHKERR(PetscSectionGetOffset(local_sec.sec, p, &off))
+        local_offsets[p] = off
+        remote_offsets[p] = -1
```
connorjward, line 3479:
```suggestion
    remote_offsets = np.full(pEnd - pStart, -1, dtype=IntType)
    for p in range(pStart, pEnd):
        CHKERR(PetscSectionGetDof(local_sec.sec, p, &dof))
        CHKERR(PetscSectionGetOffset(local_sec.sec, p, &off))
        local_offsets[p] = off
```

firedrake/cython/dmcommon.pyx
```diff
@@ -3471,3 +3429,76 @@ def to_petsc_local_numbering(PETSc.Vec vec, V):
...
+        m += dof
+    unit = MPI._typedict[np.dtype(IntType).char]
+    point_sf.bcastBegin(unit, local_offsets, remote_offsets, MPI.REPLACE)
+    point_sf.bcastEnd(unit, local_offsets, remote_offsets, MPI.REPLACE)
+    n = 0
+    for i in range(nleaves):
+        p = ilocal[i] if ilocal != NULL else i
```
dham, line 3486: Not obvious. Please comment.

firedrake/cython/dmcommon.pyx
connorjward, line 3479:
```suggestion
    remote_offsets = np.full(pEnd - pStart, -1, dtype=IntType)
    for p in range(pStart, pEnd):
        CHKERR(PetscSectionGetDof(local_sec.sec, p, &dof))
        CHKERR(PetscSectionGetOffset(local_sec.sec, p, &off))
        local_offsets[p] = off
```

dham (changes requested):

connorjward (changes requested):

firedrake/cython/dmcommon.pyx
ksagiyam: Done.

firedrake/cython/dmcommon.pyx
```diff
@@ -3471,3 +3429,76 @@ def to_petsc_local_numbering(PETSc.Vec vec, V):
...
+    for p in range(pStart, pEnd):
+        CHKERR(PetscSectionGetDof(local_sec.sec, p, &dof))
+        CHKERR(PetscSectionGetOffset(local_sec.sec, p, &off))
+        local_offsets[p] = off
+        m += dof
+    unit = MPI._typedict[np.dtype(IntType).char]
+    point_sf.bcastBegin(unit, local_offsets, remote_offsets, MPI.REPLACE)
+    point_sf.bcastEnd(unit, local_offsets, remote_offsets, MPI.REPLACE)
+    n = 0
+    for i in range(nleaves):
```
dham, line 3484:
```suggestion
    # ilocal == NULL if local leaf points are [0, 1, 2, ...).
    for i in range(nleaves):
```

dham (approved):

PR #3461, opened 2024-03-20
Fix a deadlock with VertexOnlyMesh'es created not on COMM_WORLD

pefarrell:
@UZerbinati wrote some defcon code that used a VertexOnlyMesh. The code was deadlocking. Upon inspection, VertexOnlyMesh was compiling some code but not passing the relevant communicator.


JDBetteridge (changes requested):
This is definitely the right thing to do! Is there an MFE lying around that we can use as a test to prevent this occurring again in the future?

dham (approved):

Issue #3464, opened 2024-03-20
Automatic differentiation is not supported for Cofunction.assign


Ig-dolci:
We have the `NotImplementedError: Automatic differentiation is not supported for this operation.`  for the operation `Cofunction.assign()`

Steps to reproduce:
```
from firedrake import *
from firedrake.adjoint import *

continue_annotation()
mesh = UnitSquareMesh(2, 2)
V = FunctionSpace(mesh, "CG", 1)
f = Cofunction(V.dual())
f.assign(Constant(1.0))
```
Question:
Is that expected? What is the main issue with having the automatic differentiation working for `Cofunction.assign()` operation?

nbouziani:
>  Is that expected? What is the main issue with having the automatic differentiation working for Cofunction.assign() operation?

Yes, it is. I think a workaround was discussed during one of the Firedrake meeting but that wasn't part of the dual PR (#2294).

Also, note that the above assignment is not supported as cofunctions can only be assigned zero, cofunctions, or base form objects. In fact, if you remove the `firedrake.adjoint` and `continue_annotation()`, the above code should fail because of that assignment.

Ig-dolci:
> > Is that expected? What is the main issue with having the automatic differentiation working for Cofunction.assign() operation?
> 
> Yes, it is. I think a workaround was discussed during one of the Firedrake meeting but that wasn't part of the dual PR (#2294).
> 
> Also, note that the above assignment is not supported as cofunctions can only be assigned zero, cofunctions, or base form objects. In fact, if you remove the `firedrake.adjoint` import and `continue_annotation()`, the above code should fail because of that assignment.

Ok. `Cofunction` can assign zero, cofunctions, or base form objects. 

It looks like the cofunctions assignment will not work with `FunctionMixing._ad_annotate_assign` on the case in which we assign a `BaseForm` as in the following code:
```
continue_annotation()
mesh = UnitSquareMesh(80, 80)
V = FunctionSpace(mesh, "CG", 1)
u = Function(V)
v = TestFunction(V)
a = u*v*dx
Cofunction(V.dual()).assign(a)
```
which results in the error:

```
    block = FunctionAssignBlock(self, other, ad_block_tag=ad_block_tag)
            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
[...]

PR #3476, opened 2024-03-27
Dolci/cofunction ad

Ig-dolci:
# Description
Issue [3464](https://github.com/firedrakeproject/firedrake/issues/3464)

### DIFF
--- a/firedrake/cofunction.py
+++ b/firedrake/cofunction.py
@@ -197,8 +197,7 @@ class Cofunction(ufl.Cofunction, FunctionMixin):
         elif (isinstance(expr, Cofunction)
               and expr.function_space() == self.function_space()):
             # do not annotate in case of self assignment
-            annotate = annotate_tape() and self != expr
-            if annotate:
+            if annotate_tape() and self != expr:
                 self.block_variable = self.create_block_variable()
                 self.block_variable._checkpoint = DelegatedFunctionCheckpoint(expr.block_variable)
             expr.dat.copy(self.dat, subset=subset)
--- a/firedrake/cofunction.py
+++ b/firedrake/cofunction.py
@@ -186,6 +186,12 @@ class Cofunction(ufl.Cofunction, FunctionMixin):
         If present, subset must be an :class:`pyop2.types.set.Subset` of this
         :class:`Cofunction`'s ``node_set``.  The expression will then
         only be assigned to the nodes on that subset.
+
+        Notes
+        -----
+        Cofunction assignment is annoted for automatic differentiation.
+        The operation annoted here is the `Assembly` operation, where the
+        b
         """
         expr = ufl.as_ufl(expr)
         if isinstance(expr, ufl.classes.Zero):
@@ -200,8 +206,9 @@ class Cofunction(ufl.Cofunction, FunctionMixin):
             expr.dat.copy(self.dat, subset=subset)
             return self
         elif isinstance(expr, BaseForm):
-            # Enable to write down c += B where c is a Cofunction
-            # and B an appropriate BaseForm object
+            # Enable to write down c += B where c is a Cofunction and B an appropriate BaseForm object.
+            # If the annotation is enabled, the following operation  will results in assemble block on the
+            # Pyadjoint tape.
             assembled_expr = firedrake.assemble(expr)
             return self.assign(assembled_expr, subset=subset)
 
--- a/firedrake/cofunction.py
+++ b/firedrake/cofunction.py
@@ -31,7 +31,8 @@ class Cofunction(ufl.Cofunction, FunctionMixin):
 
     @PETSc.Log.EventDecorator()
     @FunctionMixin._ad_annotate_init
-    def __init__(self, function_space, val=None, name=None, dtype=ScalarType):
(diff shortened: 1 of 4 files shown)


tests/regression/test_adjoint_operators.py
```diff
@@ -798,3 +798,22 @@ def test_3325():
...
+def test_assign_cofunction():
+    mesh = UnitSquareMesh(2, 2)
+    V = FunctionSpace(mesh, "CG", 1)
+    v = TestFunction(V)
+    u = TrialFunction(V)
+    a = u * v * dx
+    b = Constant(1.0) * v * dx
+    u0 = Cofunction(V.dual(), name="u0")
+    u1 = Cofunction(V.dual(), name="u1")
+    u0.assign(b)
```
dham, line 813:
```suggestion
    u0.assign(assemble(b))
```

dham (changes requested):
Put in a test for zero that checks that the tape is really cut when we zer oa cofunction.

Ig-dolci:
A test is failing for `LinearVariationalSolver`. I will investigate the root of it.

Ig-dolci:
I am now fine with this PR. It is also working for `LinearVariationalSolver`.

tests/regression/test_adjoint_operators.py
```diff
@@ -798,3 +799,59 @@ def test_3325():
...
+    a = u * v * dx
+    b = k * v * dx
+    u0 = Cofunction(V.dual(), name="u0")
+    u0.assign(b)
+    u0.assign(Zero())
+    sol = Function(V, name="c")
+    solve(a == u0, sol)
+    J = assemble(((sol + Constant(1.0)) ** 2) * dx)
+    with pytest.raises(
+        Exception, match="Invalid data: expected 9 values, got 1!"
```
dham, line 853: Something is still wrong here. This isn't an error message we should expect at this stage.

firedrake/cofunction.py
```diff
@@ -189,15 +191,22 @@ def assign(self, expr, subset=None):
...
+            with stop_annotating(modifies=(self,)):
+                self.dat.zero(subset=subset)
             return self
         elif (isinstance(expr, Cofunction)
               and expr.function_space() == self.function_space()):
+            # do not annotate in case of self assignment
+            annotate = annotate_tape() and self != expr
+            if annotate:
+                self.block_variable = self.create_block_variable()
+                self.block_variable._checkpoint = DelegatedFunctionCheckpoint(expr.block_variable)
```
Ig-dolci, line 203: The approach of using `self.block_variable = expr.block_variable` is not effective for general cases. Particularly, for variational solvers. `self.block_variable = expr.block_variable` results in the loss of the `self.block_variable.output` identity, which must match the identity of `self`. Modifying either `expr.count` or `self.count` is not safe. Therefore, the current code proposes to create a block variable and use `DelegatedFunctionCheckpoint` to avoid saving the assigned data twice.

firedrake/cofunction.py
```diff
@@ -31,7 +33,8 @@ class Cofunction(ufl.Cofunction, FunctionMixin):
 
     @PETSc.Log.EventDecorator()
     @FunctionMixin._ad_annotate_init
-    def __init__(self, function_space, val=None, name=None, dtype=ScalarType):
+    def __init__(self, function_space, val=None, name=None, dtype=ScalarType,
+                 count=None):
```
Ig-dolci, line 37: The introduction of `count` is essential here due to the `DelegateFunctionCheckpoint` usage when the annotation is activated. Upon recomputing the tape, the [DelegateFunctionCheckpoint.restore](https://github.com/firedrakeproject/firedrake/blob/master/firedrake/adjoint_utils/checkpointing.py#L328) returns a Cofunction with a specified `count`.

firedrake/cofunction.py
```diff
@@ -189,15 +189,20 @@ def assign(self, expr, subset=None):
...
         expr = ufl.as_ufl(expr)
         if isinstance(expr, ufl.classes.Zero):
-            self.dat.zero(subset=subset)
+            with stop_annotating(modifies=(self,)):
+                self.dat.zero(subset=subset)
             return self
         elif (isinstance(expr, Cofunction)
               and expr.function_space() == self.function_space()):
+            # Keep the identifier of the Cofunction being assigned to.
+            expr._count = self.count()
```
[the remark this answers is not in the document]
Ig-dolci: That is not safe.

firedrake/cofunction.py
```diff
@@ -189,15 +191,22 @@ def assign(self, expr, subset=None):
...
         if isinstance(expr, ufl.classes.Zero):
-            self.dat.zero(subset=subset)
+            with stop_annotating(modifies=(self,)):
+                self.dat.zero(subset=subset)
             return self
         elif (isinstance(expr, Cofunction)
               and expr.function_space() == self.function_space()):
+            # do not annotate in case of self assignment
+            annotate = annotate_tape() and self != expr
+            if annotate:
```
Ig-dolci, line 201:
```suggestion
            if annotate_tape() and self != expr:
```

dham (approved):

PR #3482, opened 2024-03-28
Add macroelement tests

pbrubeck:
# Description
Test Alfeld and Iso Lagrange variants.

Depends on https://github.com/firedrakeproject/fiat/pull/64, https://github.com/FInAT/FInAT/pull/123, https://github.com/firedrakeproject/tsfc/pull/310,


firedrake/preconditioners/low_order.py
```diff
@@ -1,17 +1,38 @@
...
 class P1SNES(PMGSNES):
+    """A two-level nonlinear solver with agressive p-coarsening."""
     def coarsen_element(self, ele):
         if super().max_degree(ele) <= self.coarse_degree:
             raise ValueError
         return super().reconstruct_degree(ele, self.coarse_degree)
+
+
+class LORPC(PMGPC):
+    """A low-order refined preconditioner with a P1-iso-Pk coarse space."""
```
wence-, line 23: Does this produce the correct sparsity in the low order operator or is that separate work to come?

pbrubeck: This will just keep the high-order sparsity pattern with zero entires. We plan to add in a ChoppedPC, that drops zero entries as a first step. Utimately we would like The Right Thing™, one candidate for this is to use facet integral kernels on codimension-0 facets (the subcells). This is very similar to the dS/ds kernels. We would need a subcell-to-node map, which shouldn't be hard to construct from the cell to node map.

dham (approved):

PR #3484, opened 2024-04-01
Replace PETSc.DMPlex.createGmsh by PETSc.DMPlex.createFromFile

lrtfm:
The previous code creates a viewer from a file and then constructs the Plex from the viewer. The viewer should be destroyed using `gmsh_viewer.destroy` after the Plex is created, otherwise, this may result in `OSError: open too many files` in parallel.

Since PETSc provides PETSc.DMPlex.createFromFile, we can utilize it directly.


connorjward (approved):
Looks good. Thanks! Assuming tests pass this can land right away. Sorry for taking some time to look at this.

PR #3488, opened 2024-04-04
Do not copy nullspaces for MassInvPC

stephankramer:
The mass matrix itself does not have a nullspace.

As an example, solving Stokes in a closed box which has a pressure nullspace. The implicit Schur complement matrix of fieldsplit_1 has that constant nullspace attached. At the level of fieldsplit_1_Mp however we have the explicit mass matrix approximation. If we copy the nullspace from the parent implicit matrix, we get into trouble when we specify a solve, "fieldsplit_1_Mp_pc_type": "ksp", with that mass matrix as the nullspace will now be applied within the iterations of the mass matrix solve, where it does not have one. This leads to hard to detect convergence failures. If we don't copy the nullspace over (as per this PR), the nullspace is still removed (as we want) immediately _after_ the MassInvPc as a whole has been applied, in the outer loop of the Schur complement solve.

### DIFF
--- a/firedrake/preconditioners/assembled.py
+++ b/firedrake/preconditioners/assembled.py
@@ -57,12 +57,7 @@ class AssembledPC(PCBase):
         self._assemble_P(tensor=self.P)
 
         # Transfer nullspace over
-        Pmat = self.P.petscmat
-        Pmat.setNullSpace(P.getNullSpace())
-        tnullsp = P.getTransposeNullSpace()
-        if tnullsp.handle != 0:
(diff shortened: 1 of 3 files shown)


pbrubeck (commented):
This looks good, could one of our tests be modified so that it breaks on current master but gets fixed with this PR?

stephankramer:
Running the "Sinker" Stokes test in parallel seems to do the trick.

dham (approved):

PR #3516, opened 2024-04-22
pyop3: fix V.local_section

ksagiyam:


connorjward (changes requested):
I think if we do this we also need to set the permutation. I think you can entirely avoid touching pyop3 here apart from to get the numbering.

ksagiyam:
Yep.

connorjward (approved):

PR #3525, opened 2024-04-25
Constant

JDBetteridge:


tests/regression/test_constant.py
```diff
@@ -5,7 +5,7 @@
 import pytest
 
 
-pytest.skip(allow_module_level=True, reason="pyop3 TODO")
+#pytest.skip(allow_module_level=True, reason="pyop3 TODO")
```
JDBetteridge, line 8:
```suggestion
```

firedrake/constant.py
```diff
@@ -198,11 +195,12 @@ def assign(self, value):
         self
 
         """
+        if self.ufl_shape() != ():
+            shape = np.array(value).shape
+            if shape != self.ufl_shape():
+                raise DataValueError("Cannot assign to constant, value has incorrect shape")
```
connorjward, line 201:
```suggestion
        if self.ufl_shape() and np.array(value).shape != self.ufl_shape():
            raise DataValueError("Cannot assign to constant, value has incorrect shape")
```

connorjward (changes requested):

connorjward:
I think this is superseded by #3527

PR #3535, opened 2024-04-26
If VTK is installed, have `from firedrake import *` import VTKFile

pefarrell:
# Description

This PR ensures that, if VTK is installed in the user's venv, `from firedrake import *` makes `VTKFile` available. This restores consistency of the Firedrake interface (other objects do not need to be imported from their respective subpackages).

### DIFF
--- a/firedrake/output/__init__.py
+++ b/firedrake/output/__init__.py
@@ -1,1 +1,5 @@
-from .vtk_output import VTKFile  # noqa: F401
+try:
+    import vtkmodules.vtkCommonDataModel  # noqa: F401
+    from .vtk_output import VTKFile  # noqa: F401
+except ModuleNotFoundError:
+    from .vtk_unavailable import VTKFile  # noqa: F401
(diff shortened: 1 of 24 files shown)


firedrake/output/__init__.py
```diff
@@ -1 +1,5 @@
-from .vtk_output import VTKFile  # noqa: F401
+try:
+    import vtkmodules.vtkCommonDataModel  # noqa: F401
+    from .vtk_output import VTKFile  # noqa: F401
+except ModuleNotFoundError:
+    pass
```
dham, line 5: I don't like this failure mode. If VTK isn't installed then the user will just get an error that the name VTKFile doesn't exist and is likely to conclude that they are using the wrong version of Firedrake. If VTKFile is to be in the global namespace then it should always be there, and if VTK isn't available it should raise a meaningful exception on instantiation.

dham (changes requested):

firedrake/output/__init__.py
pefarrell: I'm glad I've started this discussion.

I agree VTKFile should exist and do something if VTK isn't available. One option, as you say, is to raise an exception. Another option might just be to raise a warning, so that the computation can proceed regardless (since saving output to VTK format is rarely critical for the ongoing run). I prefer the latter but will do the former if you prefer. What do you think?

dham: I think you're maybe projecting your own use case. If a user asks for output then we have to presume they really want it. Imagine an application which is running on a cluster and produces a bunch of logging along the way. The user will just see no output and the reason will be a single line in the captured STDERR. What do you think the odds of them finding that are? 

If the output is actually optional then the user or application author can catch the exception. That's less likely to cause a hard to debug situation.

pefarrell: OK, running with that.

PR #3537, opened 2024-04-29
Dolci/test petsc

Ig-dolci:
Testing


Ig-dolci:
@JDBetteridge about the petsc4py fail in firedrake install for the latest petsc version. Here is the problem that I encountered also trying local installation [petsc4py fail](https://github.com/firedrakeproject/firedrake/actions/runs/8877378093/job/24370842512#step:5:253).

PR #3538, opened 2024-04-29
Fix offset cache

pbrubeck:
# Description
Caching offsets using finat elements was giving issues in #3482 (only in complex, and this was hard to replicate). Here we fix this with a cache key from entity_dofs.

Extruded is going to go away anyway :)


connorjward (approved):
Looks good to me. Feel free to merge if the tests pass.

PR #3550, opened 2024-05-02
utility meshes: set partition True in construction

ksagiyam:
Set `partition=True` for the temporary mesh when making a periodic mesh.

- If `partition=False`, `pointSF` (= `plex.getPointSF()`) will not be set up on the temporary mesh, which is required by `interpolate`, say.

### DIFF
--- a/firedrake/utility_meshes.py
+++ b/firedrake/utility_meshes.py
@@ -64,8 +64,8 @@ __all__ = [
 ]
 
 
-distribution_parameters_noop = {"partition": False,
-                                "overlap_type": (mesh.DistributedMeshOverlapType.NONE, 0)}
+distribution_parameters_no_overlap = {"partition": True,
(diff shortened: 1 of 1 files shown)


firedrake/utility_meshes.py
```diff
@@ -64,7 +64,7 @@
 ]
 
 
-distribution_parameters_noop = {"partition": False,
+distribution_parameters_noop = {"partition": True,
```
dham, line 67: please remove `_noop`

dham (changes requested):

firedrake/utility_meshes.py
ksagiyam: Renamed it `distribution_parameters_no_overlap`.

connorjward (approved):

PR #3555, opened 2024-05-07
Add P argument to solve function

juliusgh:
The class `LinearSolver` already supports passing a custom preconditioner matrix `P` to the PETSc solver.
However, this is so far not available through the `solve` function.

This PR adds an argument `P` to the `solve` function that is passed to the `LinearSolver`. At least in my case, it would be convenient to pass a custom preconditioner matrix directly in this way.

### DIFF
--- a/firedrake/solving.py
+++ b/firedrake/solving.py
@@ -52,6 +52,14 @@ def solve(*args, **kwargs):
     :class:`.EquationBC`\s specifying, respectively, the strong boundary conditions
     to apply and PDEs to solve on the boundaries.
     For the format of `solver_parameters` see below.
+    
+    Optionally, an argument `P` of type :class:`~.MatrixBase` can be passed to
+    construct any preconditioner from; if none is supplied `A` is used to
+    construct the preconditioner.
+
+    .. code-block:: python3
+
+        solve(A, x, b, P=P, bcs=bcs, solver_parameters={...})
 
     *2. Solving linear variational problems*
 
@@ -190,6 +198,9 @@ def _la_solve(A, x, b, **kwargs):
     r"""Solve a linear algebra problem.
 
     :arg A: the assembled bilinear form, a :class:`.Matrix`.
+    :arg P: an optional :class:`~.MatrixBase` to construct any
+            preconditioner from; if none is supplied ``A`` is
+            used to construct the preconditioner.
     :arg x: the :class:`.Function` to write the solution into.
     :arg b: the :class:`.Function` defining the right hand side values.
     :kwarg solver_parameters: optional solver parameters.
@@ -224,7 +235,7 @@ def _la_solve(A, x, b, **kwargs):
 
         _la_solve(A, x, b, solver_parameters=parameters_dict)."""
 
-    bcs, solver_parameters, nullspace, nullspace_T, near_nullspace, \
+    P, bcs, solver_parameters, nullspace, nullspace_T, near_nullspace, \
         options_prefix = _extract_linear_solver_args(A, x, b, **kwargs)
 
     # Check whether solution is valid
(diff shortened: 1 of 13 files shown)


firedrake/solving.py
```diff
@@ -190,6 +198,9 @@ def _la_solve(A, x, b, **kwargs):
     r"""Solve a linear algebra problem.
 
     :arg A: the assembled bilinear form, a :class:`.Matrix`.
+    :arg P: an optional :class:`~.MatrixBase` to construct any
```
connorjward, line 201: This is in the wrong place in the docstring as it suggests that `P` is the second argument to the function. Should it be documented as a `kwarg`?

firedrake/solving.py
```diff
@@ -52,6 +52,14 @@ def solve(*args, **kwargs):
     :class:`.EquationBC`\s specifying, respectively, the strong boundary conditions
     to apply and PDEs to solve on the boundaries.
     For the format of `solver_parameters` see below.
+    
+    Optionally, an argument `P` of type :class:`~.MatrixBase` can be passed to
+    construct any preconditioner from; if none is supplied `A` is used to
+    construct the preconditioner.
```
connorjward, line 58:
```suggestion
    Optionally, an argument ``P`` of type :class:`~.MatrixBase` can be passed to
    construct any preconditioner from; if none is supplied ``A`` is used to
    construct the preconditioner.
```

I think. We should check that the docstrings render correctly from the documentation CI artefact.

connorjward (changes requested):
This seems sensible. Thanks!

It would be nice to have a test for this but I think that might be tricky to come up with. We don't even test the `P` argument to `LinearSolver` as things stand.

firedrake/solving.py
juliusgh: Yes, I updated it.

juliusgh:
Thanks for the quick reply!

I think a short unit test would indeed be difficult to find. I am using the `P` matrix passing to solve homogenization problems from computational mechanics with a custom preconditioner. FFT-based preconditioners are very effective for these problems, but in addition to `A`, a suitable operator `P` must also be passed. I could provide a documented demo that uses this feature, but maybe that's too specific for firedrake...

dham (approved):
Thank you!

connorjward:
> Thanks for the quick reply!
> 
> I think a short unit test would indeed be difficult to find. I am using the `P` matrix passing to solve homogenization problems from computational mechanics with a custom preconditioner. FFT-based preconditioners are very effective for these problems, but in addition to `A`, a suitable operator `P` must also be passed. I could provide a documented demo that uses this feature, but maybe that's too specific for firedrake...

To quote @dham re the suggested demo: "yes yes yes please". Demos are always great to have.

connorjward (dismissed):

connorjward:
@juliusgh please could you rebase this code on top of the latest `master`/merge in the latest `master` and push it again? The tests are failing and that should fix them. Then I will merge this. Thanks.

juliusgh:
> @juliusgh please could you rebase this code on top of the latest `master`/merge in the latest `master` and push it again? The tests are failing and that should fix them. Then I will merge this. Thanks.

Ok, I have rebased on top of `master`. I hope that the CI works now.


> To quote @dham re the suggested demo: "yes yes yes please". Demos are always great to have.

Thanks for the feedback! I need some time to finish a well documented demo and would submit a new PR later.

connorjward (approved):

connorjward:
The rebase appears to have gone a bit wrong (the "Files changed" is now quite large). I have created a fresh PR with just your commits: https://github.com/firedrakeproject/firedrake/pull/3571. I will close this one.

juliusgh:
> The rebase appears to have gone a bit wrong (the "Files changed" is now quite large). I have created a fresh PR with just your commits: #3571. I will close this one.

Oh sorry. Thanks!

PR #3557, opened 2024-05-09
We need to stop relying on comm=None

JDBetteridge:
# Description

### DIFF
--- a/firedrake/preconditioners/fdm.py
+++ b/firedrake/preconditioners/fdm.py
@@ -20,6 +20,7 @@ from ufl.algorithms.ad import expand_derivatives
 from ufl.algorithms.expand_indices import expand_indices
 from tsfc.finatinterface import create_element
 from pyop2.compilation import load
+from pyop2.mpi import COMM_SELF
 from pyop2.sparsity import get_preallocation
 from pyop2.utils import get_petsc_dir, as_tuple
 from pyop2 import op2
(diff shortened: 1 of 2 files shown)


pefarrell:
I think for user-facing things it makes sense to default comm to COMM_WORLD. But I think that for developer-facing things (like `load`) we should demand that a comm be passed.

JDBetteridge:
I agree, to clarify my comment a bit: I think we should stop relying on the `comm=None` defaulting to `COMM_WORLD` internally. I don't have plans to force users to have to specify an MPI communicator for all of their scripts. However, internally relying on this behaviour makes it hard to track down parallel bugs!

connorjward (approved):
LGTM

firedrake/preconditioners/fdm.py
```diff
@@ -1594,9 +1595,9 @@ def tabulate_exterior_derivative(Vc, Vf, cbcs=[], fbcs=[], comm=None):
...
     zero = PETSc.Mat()
-    A00 = petsc_sparse(evaluate_dual(c0, f0), comm=PETSc.COMM_SELF) if f0 else zero
-    A11 = petsc_sparse(evaluate_dual(c1, f1), comm=PETSc.COMM_SELF) if c1 else zero
-    A10 = petsc_sparse(evaluate_dual(c0, f1, "grad"), comm=PETSc.COMM_SELF)
+    A00 = petsc_sparse(evaluate_dual(c0, f0), comm=COMM_SELF) if f0 else zero
+    A11 = petsc_sparse(evaluate_dual(c1, f1), comm=COMM_SELF) if c1 else zero
+    A10 = petsc_sparse(evaluate_dual(c0, f1, "grad"), comm=COMM_SELF)
     Dhat = block_mat(diff_blocks(tdim, ec.formdegree, A00, A11, A10), destroy_blocks=True)
     A00.destroy()
     A11.destroy()
```
wence-, line 1603: This looks sus if `A00` and/or `A11` are `zero` (which has not been `create()`ed)

JDBetteridge: I will open an issue and address this in another PR, today I want to try and get CI running

PR #3564, opened 2024-05-14
DO NOT MERGE Test Firedrake without MUMPS or Scalapack

connorjward:
# Description

Same principle as #3563.


connorjward:
I think that we cannot do without MUMPS. Some of the tests are failing with segfaults and the tests that fail are somewhat random (e.g. `tests/regression/test_scaled_mass.py::test_math_functions[scalar-Constant-f=(1)-expr=(cos(f)*cos(f) + sin(f)*sin(f))]` but no other tests in this file).

PR #3579, opened 2024-05-21
Fixing errors related to default L2 grad in adjoints

Ig-dolci:
# Description

This PR seeks the application of Riesz maps using the L2 norm as the default in Firedrake for adjoint gradients. Implementing this change requires some coding modifications.  

Some pyadjoint tests will fail because they were designed with Riesz maps using the l2 norm. To address these test failures, refer to [Pyadjoint PR 149](https://github.com/dolfin-adjoint/pyadjoint/pull/149), which has the necessary fixes.

### DIFF
--- a/firedrake/adjoint_utils/blocks/function.py
+++ b/firedrake/adjoint_utils/blocks/function.py
@@ -60,7 +60,8 @@ class FunctionAssignBlock(Block):
                 R = block_variable.output._ad_function_space(
                     prepared.function_space().mesh()
                 )
-                return self._adj_assign_constant(prepared, R)
+                adj_output = self._adj_assign_constant(prepared, R)
+                return adj_output.riesz_representation(riesz_map="l2")
             else:
                 adj_output = firedrake.Function(
                     block_variable.output.function_space())
@@ -86,7 +87,7 @@ class FunctionAssignBlock(Block):
                     ufl.derivative(
                         expr,
                         block_variable.saved_output,
-                        firedrake.Constant(1., domain=mesh)
+                        firedrake.Function(firedrake.FunctionSpace(mesh, "R", 0), val=1.0)
                     )
                 )
                 adj_output.assign(diff_expr)
--- a/firedrake/adjoint_utils/blocks/function.py
+++ b/firedrake/adjoint_utils/blocks/function.py
@@ -57,39 +57,37 @@ class FunctionAssignBlock(Block):
                     # Catch the case where adj_inputs[0] is just a float
                     return adj_inputs[0]
             elif isconstant(block_variable.output):
-                R = block_variable.output._ad_function_space(
-                    prepared.function_space().mesh()
+                adj_output = self._adj_assign_constant(
+                    prepared, block_variable.output.function_space()
                 )
-                adj_output = self._adj_assign_constant(prepared, R)
             else:
                 adj_output = firedrake.Function(
-                    block_variable.output.function_space())
+                    block_variable.output.function_space()
+                )
                 adj_output.assign(prepared)
             return adj_output.riesz_representation(riesz_map="l2")
         else:
             # Linear combination
             expr, adj_input_func = prepared
-            adj_output = firedrake.Function(adj_input_func.function_space())
             if isconstant(block_variable.output):
-                R = block_variable.output._ad_function_space(adj_output.function_space().mesh())
+                R = block_variable.output._ad_function_space(
+                    adj_input_func.function_space().mesh()
+                )
                 diff_expr = ufl.algorithms.expand_derivatives(
-                    ufl.derivative(
-                        expr, block_variable.saved_output,
-                        type(block_variable.output)(R, val=1.0)
-                    )
+                    ufl.derivative(expr, block_variable.saved_output,
+                                   firedrake.Function(R, val=1.0))
+                )
+                diff_expr_assembled = firedrake.Function(adj_input_func.function_space())
+                diff_expr_assembled.interpolate(ufl.conj(diff_expr))
+                adj_output = firedrake.Function(
+                    R, val=firedrake.assemble(ufl.Action(diff_expr_assembled, adj_input_func))
                 )
             else:
+                adj_output = firedrake.Function(adj_input_func.function_space())
                 diff_expr = ufl.algorithms.expand_derivatives(
-                    ufl.derivative(
-                        expr, block_variable.saved_output, adj_input_func
-                    )
+                    ufl.derivative(expr, block_variable.saved_output, adj_input_func)
                 )
-            # Firedrake does not support assignment of conjugate functions
-            adj_output.interpolate(ufl.conj(diff_expr))
-            if isconstant(block_variable.output):
-                adj_output = type(block_variable.output)(
-                    R, val=firedrake.assemble(ufl.Action(adj_output, adj_input_func))
-                )
+                adj_output.interpolate(ufl.conj(diff_expr))
             return adj_output.riesz_representation(riesz_map="l2")
 
     def _adj_assign_constant(self, adj_output, constant_fs):
--- a/firedrake/adjoint_utils/blocks/function.py
+++ b/firedrake/adjoint_utils/blocks/function.py
@@ -61,45 +61,36 @@ class FunctionAssignBlock(Block):
                     prepared.function_space().mesh()
                 )
                 adj_output = self._adj_assign_constant(prepared, R)
-                return adj_output.riesz_representation(riesz_map="l2")
             else:
                 adj_output = firedrake.Function(
                     block_variable.output.function_space())
                 adj_output.assign(prepared)
-                adj_output = adj_output.riesz_representation(riesz_map="l2")
-                return adj_output
+            return adj_output.riesz_representation(riesz_map="l2")
         else:
             # Linear combination
             expr, adj_input_func = prepared
             adj_output = firedrake.Function(adj_input_func.function_space())
-            if not isconstant(block_variable.output):
+            if isconstant(block_variable.output):
+                R = block_variable.output._ad_function_space(adj_output.function_space().mesh())
+                diff_expr = ufl.algorithms.expand_derivatives(
+                    ufl.derivative(
+                        expr, block_variable.saved_output,
+                        type(block_variable.output)(R, val=1.0)
+                    )
+                )
+            else:
                 diff_expr = ufl.algorithms.expand_derivatives(
                     ufl.derivative(
                         expr, block_variable.saved_output, adj_input_func
                     )
                 )
-                # Firedrake does not support assignment of conjugate functions
-                adj_output.interpolate(ufl.conj(diff_expr))
-                adj_output = adj_output.riesz_representation(riesz_map="l2")
-            else:
-                mesh = adj_output.function_space().mesh()
-                diff_expr = ufl.algorithms.expand_derivatives(
-                    ufl.derivative(
-                        expr,
-                        block_variable.saved_output,
-                        firedrake.Function(firedrake.FunctionSpace(mesh, "R", 0), val=1.0)
-                    )
-                )
-                adj_output.assign(diff_expr)
-                return adj_output.dat.inner(adj_input_func.dat)
-
+            # Firedrake does not support assignment of conjugate functions
+            adj_output.interpolate(ufl.conj(diff_expr))
             if isconstant(block_variable.output):
-                R = block_variable.output._ad_function_space(
-                    adj_output.function_space().mesh()
+                adj_output = type(block_variable.output)(
+                    R, val=firedrake.assemble(ufl.Action(adj_output, adj_input_func))
                 )
-                return self._adj_assign_constant(adj_output, R)
-            else:
-                return adj_output
+            return adj_output.riesz_representation(riesz_map="l2")
 
     def _adj_assign_constant(self, adj_output, constant_fs):
         r = firedrake.Function(constant_fs)
--- a/firedrake/adjoint_utils/function.py
+++ b/firedrake/adjoint_utils/function.py
@@ -240,10 +240,10 @@ class FunctionMixin(FloatingType):
         if riesz_representation == "l2":
             return Function(V, val=value.dat)
 
-        if not isinstance(value, Cofunction):
-            raise TypeError("Expected a Cofunction")
-
         elif riesz_representation in ("L2", "H1"):
+            if not isinstance(value, Cofunction):
+                raise TypeError("Expected a Cofunction")
+
             ret = Function(V)
             a = self._define_riesz_map_form(riesz_representation, V)
             firedrake.solve(a == value, ret, **solver_options)
@@ -253,7 +253,7 @@ class FunctionMixin(FloatingType):
             return riesz_representation(value)
 
         else:
-            raise NotImplementedError(
+            raise ValueError(
                 "Unknown Riesz representation %s" % riesz_representation)
 
     def _define_riesz_map_form(self, riesz_representation, V):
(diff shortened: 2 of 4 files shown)


firedrake/ml/pytorch/fem_operator.py
```diff
@@ -81,9 +81,8 @@ def backward(ctx, grad_output):
         if isinstance(adj_input, Constant) and adj_input.ufl_shape == ():
             # This will later on result in an `AdjFloat` adjoint input instead of a Constant
             adj_input = float(adj_input)
-
         # Compute adjoint model of `F`: delegated to pyadjoint.ReducedFunctional
-        adj_output = F.derivative(adj_input=adj_input)
+        adj_output = F.derivative(adj_input=adj_input, options={"riesz_representation": None})
```
Ig-dolci, line 85: Riesz map fails for L2 norm. I do not know what should be `adj_output` in this case. For now, I set as a derivative (`Cofunction`).

APaganini: I'd say that unless the user specifies options={"riesz_representation": None}, the output should be a function

tests/regression/test_adjoint_operators.py
```diff
@@ -889,13 +889,41 @@ def test_cofunction_subfunctions_with_adjoint():
 
 
 @pytest.mark.skipcomplex  # Taping for complex-valued 0-forms not yet done
-def test_none_riesz_representation_to_derivative():
+def test_riesz_representation_for_adjoints():
+    # Test is the riesz representation norms for adjoints are working as expected.
```
Ig-dolci, line 893:
```suggestion
    # Check if the Riesz representation norms for adjoints are working as expected.
```

tests/regression/test_adjoint_operators.py
```diff
@@ -889,13 +889,41 @@ def test_cofunction_subfunctions_with_adjoint():
...
-    assert isinstance(rf.derivative(options={"riesz_representation": "H1"}), Function)
-    assert isinstance(rf.derivative(options={"riesz_representation": "L2"}), Function)
-    assert isinstance(rf.derivative(options={"riesz_representation": None}), Cofunction)
+    f = Function(space).interpolate(SpatialCoordinate(mesh)[0])
+    J = assemble((f ** 2) * dx)
+    rf = ReducedFunctional(J, Control(f))
+    with stop_annotating():
+        v = TestFunction(space)
+        u = TrialFunction(space)
+        dJdu_cofunction = assemble(2 * inner(f, v) * dx)
```
APaganini, line 902: Instead of deriving the derivative formula by hand, we could do
```
J = (f ** 2) * dx
rf = ReducedFunctional(assemble(J), Control(f))
with stop_annotating():
        v = TestFunction(space)
        u = TrialFunction(space)
        dJdu_cofunction = assemble(derivative(J, f, v))
```
This would reduce the likelihood of intrucing a bug if `J` is modified in the future.

tests/regression/test_adjoint_operators.py
```diff
@@ -889,13 +889,41 @@ def test_cofunction_subfunctions_with_adjoint():
...
+
+        # Riesz representation with l2
+        dJdu_function_l2 = Function(space, val=dJdu_cofunction.dat)
+
+        # Riesz representation with H1
+        a = firedrake.inner(u, v)*firedrake.dx \
+            + firedrake.inner(firedrake.grad(u), firedrake.grad(v))*firedrake.dx
```
APaganini, line 909: Why not simply `a =  (dot(grad(u), grad(v)) + u*v) * dx`?

tests/regression/test_adjoint_operators.py
```diff
@@ -889,13 +889,41 @@ def test_cofunction_subfunctions_with_adjoint():
...
+        dJdu_function_H1 = Function(space)
+        solve(a == dJdu_cofunction, dJdu_function_H1)
+
+        # Riesz representation with L2
+        a = firedrake.inner(u, v)*firedrake.dx
```
APaganini, line 914: Again, why not just `a =  u*v*dx`?

tests/regression/test_adjoint_operators.py
```diff
@@ -889,13 +889,41 @@ def test_cofunction_subfunctions_with_adjoint():
...
+        dJdu_function_L2 = Function(space)
+        solve(a == dJdu_cofunction, dJdu_function_L2)
+
+    dJdu_none = rf.derivative(options={"riesz_representation": None})
+    dJdu_l2 = rf.derivative(options={"riesz_representation": "l2"})
+    dJdu_H1 = rf.derivative(options={"riesz_representation": "H1"})
+    dJdu_default_L2 = rf.derivative()
```
APaganini, line 921: I would also add `dJdu_L2 = rf.derivative(options={"riesz_representation": "L2"})`.

tests/regression/test_adjoint_operators.py
```diff
@@ -889,13 +889,41 @@ def test_cofunction_subfunctions_with_adjoint():
...
+    assert (
+        isinstance(dJdu_none, Cofunction) and isinstance(dJdu_function_l2, Function)
+        and isinstance(dJdu_H1, Function) and isinstance(dJdu_default_L2, Function)
```
APaganini, line 924: and `isinstance(dJdu_L2, Function)`

tests/regression/test_adjoint_operators.py
```diff
@@ -889,13 +889,41 @@ def test_cofunction_subfunctions_with_adjoint():
...
+        and np.allclose(dJdu_none.dat.data, dJdu_cofunction.dat.data)
+        and np.allclose(dJdu_l2.dat.data, dJdu_function_l2.dat.data)
+        and np.allclose(dJdu_H1.dat.data, dJdu_function_H1.dat.data)
+        and np.allclose(dJdu_default_L2.dat.data, dJdu_function_L2.dat.data)
```
APaganini, line 928: `and np.allclose(dJdu_L2.dat.data, dJdu_function_L2.dat.data)`

firedrake/adjoint_utils/blocks/function.py
```diff
@@ -60,7 +60,8 @@ def evaluate_adj_component(self, inputs, adj_inputs, block_variable, idx,
                 R = block_variable.output._ad_function_space(
                     prepared.function_space().mesh()
                 )
-                return self._adj_assign_constant(prepared, R)
+                adj_output = self._adj_assign_constant(prepared, R)
+                return adj_output.riesz_representation(riesz_map="l2")
```
APaganini, line 64: I don't have enough pyadjoint expertise to assess this.

firedrake/adjoint_utils/blocks/function.py
```diff
@@ -86,7 +87,7 @@ def evaluate_adj_component(self, inputs, adj_inputs, block_variable, idx,
                     ufl.derivative(
                         expr,
                         block_variable.saved_output,
-                        firedrake.Constant(1., domain=mesh)
+                        firedrake.Function(firedrake.FunctionSpace(mesh, "R", 0), val=1.0)
```
APaganini, line 90: I don't have enough pyadjoint expertise to assess this.

firedrake/adjoint_utils/function.py
```diff
@@ -225,23 +225,24 @@ def _ad_convert_riesz(self, value, options=None):
...
         options = {} if options is None else options
-        riesz_representation = options.get("riesz_representation", "l2")
+        riesz_representation = options.get("riesz_representation", "L2")
         solver_options = options.get("solver_options", {})
         V = options.get("function_space", self.function_space())
 
-        if riesz_representation != "l2" and not isinstance(value, Cofunction):
-            raise TypeError("Expected a Cofunction")
-        elif not isinstance(value, (Number, Cofunction, Function)):
+        if not isinstance(value, (Number, Cofunction, Function)):
```
APaganini, line 232: To match the error message in line 233
`if not isinstance(value, (Cofunction, Function, Number)):`

firedrake/adjoint_utils/function.py
```diff
@@ -225,23 +225,24 @@ def _ad_convert_riesz(self, value, options=None):
...
             raise TypeError("Expected a Cofunction, Function or a float")
 
+        if isinstance(value, Number):
+            # It is applied l2 representation for this case.
```
APaganini, line 236: This comment is not very clear

APaganini (dismissed):
David should also look at adjoint_utils/function.py and at adjoint_utils/blocks/function.py

tests/regression/test_adjoint_operators.py
Ig-dolci: done.

tests/regression/test_adjoint_operators.py
Ig-dolci: done.

tests/regression/test_adjoint_operators.py
Ig-dolci: done.

tests/regression/test_adjoint_operators.py
Ig-dolci: I rewrote  `dJdu_cofunction` as `dJdu_cofunction = assemble(derivative((f ** 2) * dx, f, v))`.

tests/regression/test_adjoint_operators.py
Ig-dolci: done.

tests/regression/test_adjoint_operators.py
Ig-dolci: done.

firedrake/adjoint_utils/function.py
Ig-dolci: done.

firedrake/adjoint_utils/function.py
Ig-dolci: I modified this part of the code in the latest commit, which is now written as:
```
if isinstance(value, Number):
    if value == 0.:
        # l2 Riesz map is directly applied when the value is a real number 0..
        # This is seen in adjoint-based derivative when the functional
        # is independent of the control variable.
        return Function(V)
    elif self.ufl_element().family() == "Real":
        # Apply the l2 Riesz map for the case where self is a function in Real space.
        f = Function(V)
        f.assign(value)
        return f
    else:
        raise TypeError("Riesz map of a non-zero scalar is not supported for non-Real function spaces.")
```

firedrake/adjoint_utils/function.py
```diff
@@ -225,23 +225,32 @@ def _ad_convert_riesz(self, value, options=None):
...
 
-        if riesz_representation == "l2":
-            if isinstance(value, (Cofunction, Function)):
-                return Function(V, val=value.dat)
-            else:
+        if isinstance(value, Number):
+            if value == 0.:
+                # l2 Riesz map is directly applied when the value is a real number 0..
+                # This is seen in adjoint-based derivative when the functional
+                # is independent of the control variable.
```
APaganini, line 239: I suggest the following comment
# In adjoint-based differentiation, the case isinstance(value, Number) and value == 0. arises
# only when the functional is independent on the control variable. In this case, we apply the
# l2 Riesz map

firedrake/adjoint_utils/function.py
```diff
@@ -225,23 +225,32 @@ def _ad_convert_riesz(self, value, options=None):
...
         options = {} if options is None else options
-        riesz_representation = options.get("riesz_representation", "l2")
+        riesz_representation = options.get("riesz_representation", "L2")
         solver_options = options.get("solver_options", {})
         V = options.get("function_space", self.function_space())
 
-        if riesz_representation != "l2" and not isinstance(value, Cofunction):
-            raise TypeError("Expected a Cofunction")
-        elif not isinstance(value, (Number, Cofunction, Function)):
+        if not isinstance(value, (Number, Cofunction, Function, Number)):
```
APaganini, line 232: Change into 
`if not isinstance(value, (Cofunction, Function, Number)):`
to remove redundant `Number`

firedrake/adjoint_utils/function.py
```diff
@@ -225,23 +225,32 @@ def _ad_convert_riesz(self, value, options=None):
...
-                    f.assign(value)
+                f.assign(value)
                 return f
+            else:
+                raise TypeError("Riesz map of a non-zero scalar is not supported for non-Real function spaces.")
+
+        if riesz_representation == "l2":
+            return Function(V, val=value.dat)
+
+        if riesz_representation != "l2" and not isinstance(value, Cofunction):
```
APaganini, line 252: `riesz_representation != "l2"` will always be `True` at this point, no?

Ig-dolci: Yes. You are right. I just rewrote `if not isinstance(value, Cofunction)`.

firedrake/adjoint_utils/function.py
Ig-dolci: With the latest commits, `_ad_convert_riesz` should not allow `value` to be a `Number` when `value !=0`. So, the first line of `_ad_convert_riesz` function verifies if `value == 0.`. So, at this point, values must be either a `Cofunction` or a `Function`:
```
if not isinstance(value, (Cofunction, Function)):
    raise TypeError("Expected a Cofunction or a Function")
```

firedrake/adjoint_utils/function.py
```diff
@@ -224,24 +223,25 @@ def _ad_create_checkpoint(self):
...
-        options = {} if options is None else options
-        riesz_representation = options.get("riesz_representation", "l2")
-        solver_options = options.get("solver_options", {})
         V = options.get("function_space", self.function_space())
+        if value == 0.:
+            # In adjoint-based differentiation, value == 0. arises only when
+            # the functional is independent on the control variable.
+            # In this case, we do not apply the Riesz map and return a zero
+            # Cofunction.
+            return Cofunction(V.dual())
```
Ig-dolci, line 232: I am unsure if returning a zero `Cofunction` here is the best practice.

firedrake/ml/pytorch/fem_operator.py
```diff
@@ -83,7 +83,7 @@ def backward(ctx, grad_output):
             adj_input = float(adj_input)
 
         # Compute adjoint model of `F`: delegated to pyadjoint.ReducedFunctional
-        adj_output = F.derivative(adj_input=adj_input)
+        adj_output = F.derivative(adj_input=adj_input, options={"riesz_representation": "L2"})
```
Ig-dolci, line 86: @nbouziani With this PR, the functional derivatives will, by default, result from the L2 Riesz representation to map from the dual to the primal space. However, I am facing test failures in `firedrake/tests/regression/test_taylor_fem_operator.py`, specifically at `torch.autograd.gradcheck(G, x_P, eps=1e-6)`.
I have a couple of questions:

1. Should the `adj_output` be in the primal or dual space?

1. Do you have any idea why using `F.derivative(adj_input=adj_input, options={"riesz_representation": "L2"})` might lead to these test failures at `torch.autograd.gradcheck(G, x_P, eps=1e-6)`?

firedrake/adjoint_utils/function.py
Ig-dolci, line 232: I am unsure if returning a zero `Cofunction` here is the best practice.

firedrake/ml/pytorch/fem_operator.py
nbouziani: 1. It depends on the operation represented by `F`, but I don't think it matters in this case (see 2.)
2. I strongly suspect that's because PyTorch assumes `l2` product for the Riesz representation. I don't think we should default to `L2` here. I think we should either always impose `l2` or give that option to the `fem_operator` and default to `l2`. It's not clear to me why considering other inner products would make sense, given that `l2` is assumed throughout the AD calculations in PyTorch.

dham (approved):

PR #3583, opened 2024-05-22
Restore Python 3.8 compatibility

JDBetteridge:
# Description
In order to detect features that aren't compatible, it isn't sufficient to lint using an installation of Python 3.8. :confused:

### DIFF
--- a/firedrake/variational_solver.py
+++ b/firedrake/variational_solver.py
@@ -133,7 +133,13 @@ class NonlinearVariationalSolver(OptionsManager, NonlinearVariationalSolverMixin
     DEFAULT_SNES_PARAMETERS = DEFAULT_SNES_PARAMETERS
 
     # Looser default tolerance for KSP inside SNES.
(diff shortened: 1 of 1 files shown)


firedrake/variational_solver.py
```diff
@@ -133,7 +133,7 @@ class NonlinearVariationalSolver(OptionsManager, NonlinearVariationalSolverMixin
     DEFAULT_SNES_PARAMETERS = DEFAULT_SNES_PARAMETERS
 
     # Looser default tolerance for KSP inside SNES.
-    DEFAULT_KSP_PARAMETERS = MappingProxyType(DEFAULT_KSP_PARAMETERS | {'ksp_rtol': 1e-5})
+    DEFAULT_KSP_PARAMETERS = MappingProxyType(DEFAULT_KSP_PARAMETERS.update({'ksp_rtol': 1e-5}))
```
wence-, line 136: Yeah, this doesn't work for two reasons:

1. update is in-place
2. update returns `None` because python hates function chaining.

JDBetteridge: And 3. `MappingProxyType` doesn't even have an update method!

connorjward (approved):

PR #3577, opened 2024-05-20
Gather functional

colinjcotter:
Add a gather_functional option to EnsembleReducedFunctional.

The forward composition is:
each ensemble rank computes its J, and then gather_functional collects all the Js, and computes a functional of them, recorded in gather_functional.

PR #3587, opened 2024-05-23
Lists in ensemble reduced functional

colinjcotter:
This is an extension to #3577 which should be reviewed first.

This update allows for multiple functionals per ensemble rank, with their own controls, which is what we need for our use case.

### DIFF
--- a/firedrake/adjoint/ensemble_reduced_functional.py
+++ b/firedrake/adjoint/ensemble_reduced_functional.py
@@ -42,8 +42,9 @@ class EnsembleReducedFunctional(ReducedFunctional):
         functionals and their derivatives between the ensemble members.
     scatter_control : bool
         Whether scattering a control (or a list of controls) over the ensemble communicator
-    gather_functional: a reduced functional that takes in all of the Js.
         ``Ensemble.ensemble comm``.
+    gather_functional: a reduced functional that takes in all of the Js.
+
 
     See Also
     --------
--- a/firedrake/adjoint/ensemble_reduced_functional.py
+++ b/firedrake/adjoint/ensemble_reduced_functional.py
@@ -78,7 +78,6 @@ class EnsembleReducedFunctional(ReducedFunctional):
 
     def __call__(self, values):
         local_functional = super(EnsembleReducedFunctional, self).__call__(values)
-        size = self.ensemble.ensemble_comm.size
         ensemble_comm = self.ensemble.ensemble_comm
         if self.gather_functional:
             Controls_g = self._allgather_J(local_functional)
--- a/firedrake/adjoint/ensemble_reduced_functional.py
+++ b/firedrake/adjoint/ensemble_reduced_functional.py
@@ -63,25 +63,25 @@ class EnsembleReducedFunctional(ReducedFunctional):
         self.scatter_control = scatter_control
         self.gather_functional = gather_functional
 
-    def _bcast_J(self, J, i):
+    def _allgather_J(self, J):
         if isinstance(J, float):
-            val = self.ensemble.ensemble_comm.bcast(J, root=i)
+            vals = self.ensemble.ensemble_comm.allgather(J)
         elif isinstance(J, firedrake.Function):
-            J0 = J.copy(deepcopy=True)
-            val = self.ensemble.bcast(J0, root=i)
+            #  allgather not implemented in ensemble.py
+            vals = []
+            for i in range(self.ensemble.ensemble_comm.size):
+                J0 = J.copy(deepcopy=True)
+                vals.append(self.ensemble.bcast(J0, root=i))
         else:
             raise NotImplementedError("This type of functional is not supported.")
-        return val
+        return vals
 
     def __call__(self, values):
         local_functional = super(EnsembleReducedFunctional, self).__call__(values)
         size = self.ensemble.ensemble_comm.size
         ensemble_comm = self.ensemble.ensemble_comm
-        Controls_g = []
         if self.gather_functional:
-            for i in range(size):
-                Controls_g.append(self._bcast_J(local_functional, i))
-                print(Controls_g[i], i, ensemble_comm.rank)
+            Controls_g = self._allgather_J(local_functional)
             total_functional = self.gather_functional(Controls_g)
         # if gather_functional is None then we do a sum
         elif isinstance(local_functional, float):
--- a/firedrake/adjoint/ensemble_reduced_functional.py
+++ b/firedrake/adjoint/ensemble_reduced_functional.py
@@ -1,4 +1,4 @@
-from pyadjoint import ReducedFunctional, AdjFloat
+from pyadjoint import ReducedFunctional
 from pyadjoint.enlisting import Enlist
 from pyop2.mpi import MPI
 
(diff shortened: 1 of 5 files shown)


firedrake/adjoint/ensemble_reduced_functional.py
```diff
@@ -53,15 +61,72 @@ class EnsembleReducedFunctional(ReducedFunctional):
...
     <https://www.firedrakeproject.org/parallelism.html#id8>`_.
     """
-    def __init__(self, J, control, ensemble, scatter_control=True):
-        super(EnsembleReducedFunctional, self).__init__(J, control)
+    def __init__(self, J, control, ensemble, scatter_control=True,
+                 gather_functional=None):
         self.ensemble = ensemble
+        if isinstance(J, list):
+            assert isinstance(control, list), "Controls should also be a list."
+            assert len(control) % len(J) == 0, "Controls and J have mismatching lengths."
```
Ig-dolci, line 69: Instead of using `assert`, use exception raising errors:

```
if not isinstance(control, list):
    raise TypeError("Controls should be a list.")
if len(control) % len(J) != 0:
    raise ValueError("Controls and J have mismatching lengths.")
```

firedrake/adjoint/ensemble_reduced_functional.py
```diff
@@ -53,15 +61,72 @@ class EnsembleReducedFunctional(ReducedFunctional):
...
+            self.controls_per_J = len(control) // len(J)
+            self.functional = J
+            self.controls = Enlist(control)
+            self.Jhats = []
+            self.sizes = self.ensemble.ensemble_comm.allgather(len(J))
+            for i in range(len(J)):
+                i0 = self.controls_per_J*i
+                i1 = self.controls_per_J*(i+1)
+                self.Jhats.append(ReducedFunctional(J[i], control[i0:i1]))
```
Ig-dolci, line 78: Perhaps would be better to set as a list of tuples, where each tuple corresponds to a set of controls for each `J` element. 

Then, you can iterate through `J` and `control` simultaneously:
```
for i in range(len(J)):
    self.Jhats.append(ReducedFunctional(J[i], control[i]))

```

firedrake/adjoint/ensemble_reduced_functional.py
```diff
@@ -53,15 +61,72 @@ class EnsembleReducedFunctional(ReducedFunctional):
...
+            vals = []
+            # allgather a flattened list of all of the functional values
+            for i in range(len(self.sizes)):
+                for j in range(self.sizes[i]):
+                    Jtype = type(J[j]) if i == rank else None
+                    Jtype = self.ensemble.ensemble_comm.bcast(Jtype, root=i)
+                    if issubclass(Jtype, float):
+                        Jsend = J[j] if i == rank else None
+                        vals.append(self.ensemble.ensemble_comm.bcast(Jsend,
+                                                                      root=i))
```
Ig-dolci, line 96:
```suggestion
                        vals.append(self.ensemble.ensemble_comm.bcast(Jsend, root=i))
```

firedrake/adjoint/ensemble_reduced_functional.py
```diff
@@ -53,15 +61,72 @@ class EnsembleReducedFunctional(ReducedFunctional):
...
     def __call__(self, values):
-        local_functional = super(EnsembleReducedFunctional, self).__call__(values)
-        if isinstance(local_functional, float):
-            total_functional = self.ensemble.ensemble_comm.allreduce(sendobj=local_functional, op=MPI.SUM)
+        if isinstance(self.functional, list):
+            local_functional = []
+            for i in range(len(self.functional)):
+                i0 = self.controls_per_J*i
+                i1 = self.controls_per_J*(i+1)
+                local_functional.append(self.Jhats[i](values[i0:i1]))
```
Ig-dolci, line 120: Also here values would be fine to have a list of tuples. Then, 
```
for i in range(len(self.functional)):
    local_functional.append(self.Jhats[i](values[i]))
```

firedrake/adjoint/ensemble_reduced_functional.py
```diff
@@ -88,10 +153,31 @@ def derivative(self, adj_input=1.0, options=None):
...
+            i = self.ensemble.ensemble_comm.rank
+            if isinstance(self.functional, list):
+                # we will pack the derivatives into a flattened list
+                dJdm_local = []
+                k = int(np.sum(self.sizes[:i])-1)
+                for j in range(self.sizes[i]):
+                    k += 1
+                    adj_input = dJg_dmg[k]
+                    der = self.Jhats[j].derivative(adj_input=adj_input,
+                                                   options=options)
```
Ig-dolci, line 169:
```suggestion
                    der = self.Jhats[j].derivative(adj_input=adj_input, options=options)
```

Ig-dolci, line 169:
```suggestion
                    der = self.Jhats[j].derivative(adj_input=adj_input, options=options)
```

firedrake/adjoint/ensemble_reduced_functional.py
```diff
@@ -53,15 +61,72 @@ class EnsembleReducedFunctional(ReducedFunctional):
...
         self.scatter_control = scatter_control
+        self.gather_functional = gather_functional
+
+    def _allgather_J(self, J):
+        if isinstance(J, list):
+            rank = self.ensemble.ensemble_comm.rank
+            vals = []
+            # allgather a flattened list of all of the functional values
+            for i in range(len(self.sizes)):
+                for j in range(self.sizes[i]):
```
Ig-dolci, line 90: It is better to use `enumerate` here, e.g.:

```
for i, size in enumerate(self.sizes):
    for j in range(size):
```

firedrake/adjoint/ensemble_reduced_functional.py
```diff
@@ -53,15 +61,72 @@ class EnsembleReducedFunctional(ReducedFunctional):
...
+            self.functional = J
+            self.controls = Enlist(control)
+            self.Jhats = []
+            self.sizes = self.ensemble.ensemble_comm.allgather(len(J))
+            for i in range(len(J)):
+                i0 = self.controls_per_J*i
+                i1 = self.controls_per_J*(i+1)
+                self.Jhats.append(ReducedFunctional(J[i], control[i0:i1]))
+        else:
+            super(EnsembleReducedFunctional, self).__init__(J, control)
```
dham, line 80: don't have this case. Just treat this as a list of one `J`.

firedrake/adjoint/ensemble_reduced_functional.py
dham: loop over the Js directly

firedrake/adjoint/ensemble_reduced_functional.py
```diff
@@ -53,15 +61,72 @@ class EnsembleReducedFunctional(ReducedFunctional):
...
+                    if issubclass(Jtype, float):
+                        Jsend = J[j] if i == rank else None
+                        vals.append(self.ensemble.ensemble_comm.bcast(Jsend,
+                                                                      root=i))
+                    elif issubclass(Jtype, firedrake.Function):
+                        Jsend = J[j].copy(deepcopy=True) if i == rank else None
+                        vals.append(self.ensemble.bcast(Jsend, root=i))
+                    else:
+                        raise NotImplementedError("This type of functional is not supported: " + str(Jtype))
+        elif isinstance(J, float):
```
dham, line 102: if you always enlist J, you lose the last two cases here.

dham (changes requested):

firedrake/adjoint/ensemble_reduced_functional.py
colinjcotter: I'm not sure why you want me to do that?

firedrake/adjoint/ensemble_reduced_functional.py
colinjcotter: I can't here, because each J is local to one ensemble rank.

firedrake/adjoint/ensemble_reduced_functional.py
colinjcotter: That will break the optimisation wrappers.

dham: That can't really be true because you don't call the superclass constructor in the case where you have a list of Js.

firedrake/adjoint/ensemble_reduced_functional.py
colinjcotter: Where am I enlisting J?

dham: you;re not. that's the issue. All over pyadjoint they use this enlist trick to deal with exactly this case: situations where either one item or a list would be OK.

firedrake/adjoint/ensemble_reduced_functional.py
colinjcotter: The optimisation wrappers already don't work in that case. I'm concerned about breaking it for the usual case.

colinjcotter: I think I am confused about what you are asking me to do. Are you asking me to always call the superclass constructor? I cannot do that when J is a list.

dham: No. I'm saying never call the superclass constructor. In fact if we were really doing this properly then we wouldn't actually subclass ReducedFunctional but we'd instead make an AbstractReducedFunctional and inherit from that.

The optimisation wrappers will need to be fixed anyway, right?

colinjcotter: I'm not proposing to touch the optimisation wrappers. ROL is a binfire and the allgather in the scipy wrappers is not feasible for ensemble in this case.

dham: So what are you going to use a ReducedFunctional for if not to optimise?

colinjcotter: Initially, compute a gradient. Later, I will have to hand tool my own Gauss-Newton solver because none of the packages provide it, and all other algorithms are too slow.

colinjcotter: Eventually the TAO situation will be sorted out and then I can fix wrappers for that.

colinjcotter: Actually having read the code more carefully, I see that it can work just fine if we always Enlist J and make Jhats. I'll sketch it out in the following commits.

colinjcotter: OK I have fixed all this now, I will add something to the PR description.

firedrake/adjoint_utils/function.py
```diff
@@ -225,11 +225,7 @@ def _ad_convert_riesz(self, value, options=None):
 
         V = options.get("function_space", self.function_space())
         if value == 0.:
-            # In adjoint-based differentiation, value == 0. arises only when
-            # the functional is independent on the control variable.
-            # In this case, we do not apply the Riesz map and return a zero
-            # Cofunction.
-            return Cofunction(V.dual())
+            return Function(V)
```
colinjcotter, line 228: this was done because I have lists of functionals now for each ensemble rank but only one set of controls. Each functional individually might be independent of one or more of the controls, but I need to add the results.

PR #3596, opened 2024-05-29
MG: embedded transfer support for non-point variants

pbrubeck:
# Description
The MG transfer kernels assume Lagrange elements with point evaluation, and they produce an error when the element we want to transfer does not have point evaluation DOFs. 

We have CG/DG element variants that are not Lagrange, for instance the "fdm" and "integral" (a.k.a. hierarchical) variants. To transfer functions with these elements we must use an embedding space with a point variant.

Note that CG1 and DG0 would always have point evaluation DOFs regardless of the variant, and we can easily detect this from the FInAT element instance. UFL elements can only tell us about the family, variant, and degree. So in order to ensure that the element at hand is Lagrange, it'd be better to cache the MG tranfers with a key derived from the FInAT element.

### DIFF
--- a/firedrake/mg/embedded.py
+++ b/firedrake/mg/embedded.py
@@ -13,7 +13,7 @@ __all__ = ("TransferManager", )
 
 
 native_families = frozenset(["Lagrange", "Discontinuous Lagrange", "Real", "Q", "DQ"])
-integral_variants = frozenset(["integral", "fdm"])
+non_native_integral_variants = frozenset(["integral", "fdm"])
 
 
 class Op(IntEnum):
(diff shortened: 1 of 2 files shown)


firedrake/mg/embedded.py
```diff
@@ -58,7 +59,7 @@ def is_native(self, element):
             return True
         if isinstance(element.cell, ufl.TensorProductCell) and len(element.sub_elements) > 0:
             return reduce(and_, map(self.is_native, element.sub_elements))
-        return element.family() in native
+        return element.family() in native_families and not element.variant() in integral_variants
```
dham, line 62:
```suggestion
        return element.family() in native_families and not element.variant() in non_native_integral_variants
```

firedrake/mg/embedded.py
```diff
@@ -12,7 +12,8 @@
 __all__ = ("TransferManager", )
 
 
-native = frozenset(["Lagrange", "Discontinuous Lagrange", "Real", "Q", "DQ"])
+native_families = frozenset(["Lagrange", "Discontinuous Lagrange", "Real", "Q", "DQ"])
+integral_variants = frozenset(["integral", "fdm"])
```
dham, line 16:
```suggestion
non_native_integral_variants = frozenset(["integral", "fdm"])
```

dham (changes requested):

PR #3601, opened 2024-05-31
Allow cythonizing Firedrake with upstream PETSc

JDBetteridge:
# Description
with upstream PETSc OR newer compiler toolchain, not sure which...


Ig-dolci (commented):
It looks fine to me.

ksagiyam (approved):

PR #3602, opened 2024-05-31
Performance Regresson 2

JDBetteridge:
# Description
The performance regression we were seeing for Python 3.12 seems to have hit master, clearly this isn't a Python/Ubuntu version issue.

### DIFF
--- a/firedrake/adjoint/ensemble_reduced_functional.py
+++ b/firedrake/adjoint/ensemble_reduced_functional.py
@@ -28,8 +28,6 @@ class EnsembleReducedFunctional(ReducedFunctional):
     operation is employed to sum the functionals and their gradients over an ensemble
     communicator.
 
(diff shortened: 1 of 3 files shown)


connorjward:
This doesn't seem to have fixed the issue...

connorjward:
> This doesn't seem to have fixed the issue...

Or maybe it does? There is so much variability in our CI run times.

JDBetteridge:
The runners seem pretty idle at the moment, I wonder if this is something else

connorjward:
> The runners seem pretty idle at the moment, I wonder if this is something else

The coverage issue we recently found did seem to be specific to Python 3.12, so I wouldn't be surprised if it were something else. Probably some upstream package has just changed version.

JDBetteridge:
> > The runners seem pretty idle at the moment, I wonder if this is something else
> 
> The coverage issue we recently found did seem to be specific to Python 3.12, so I wouldn't be surprised if it were something else. Probably some upstream package has just changed version.

:anguished:

PR #3605, opened 2024-05-31
More MUMPS purging

JDBetteridge:
# Description
Didn't do a thorough job of making tests pass if MUMPS isn't installed!

Given that the test suite needs MUMPS to pass, this didn't seem important, but there are important use cases where MUMPS is not available, but we might want to run (a subset of) the tests!


connorjward:
I thought we had chosen to make MUMPS more or less a hard dependency?

JHopeCollins (approved):

PR #3608, opened 2024-06-04
Exception handling for missing gfortran

### LINKED ISSUE #3496: BUG: Install script accesses gfortran before installing it
**Describe the bug**
As observed in #3494, we attempt to sniff gfortran before we potentially install it using apt or brew. This causes the install script to fail on machines that don't already have gfortran installed.

JDBetteridge:
# Description
Fixes #3496


JHopeCollins (approved):
LGTM

Ideally we'd check the gfortran configuration again if the install script installs it but I think this is ok for now, this PR is just fixing the existing intended behaviour.

Issue #3628, opened 2024-06-12
BUG: VTKFile fails in parallel after creating a mesh with mpi.COMM_SELF in a single process


diego-hayashi:
**Describe the bug**
When running in parallel, after creating a mesh in process 0 with `mpi.COMM_SELF`, if VTKFile is used for another mesh in `mpi.COMM_WORLD`, the generation of the file results in missing data, according to ParaView.

**Steps to Reproduce**

1. Download the `.msh` file from [https://github.com/firedrakeproject/firedrake/blob/master/docs/notebooks/stokes-control.msh](https://github.com/firedrakeproject/firedrake/blob/master/docs/notebooks/stokes-control.msh).

2. Create a file (`test.py`) with the following content:
```python
from firedrake import *
mesh = Mesh('stokes-control.msh', comm = mpi.COMM_WORLD)
if mpi.COMM_WORLD.Get_rank() == 0:
	mesh_serial = Mesh('stokes-control.msh', comm = mpi.COMM_SELF)
U = FunctionSpace(mesh, 'CG', 1)
u = Function(U)
VTKFile('u.pvd').write(u)
```

3. Then, run it as:
```console
$ mpiexec -n 2 python -u test.py
```

**Expected behavior**
The generated `u.pvd` file should have been generated such that ParaView does not say that there is missing data, and the `u` variable should have been plotted with all zero values.

**Error message**
The Firedrake code does not show any error message. However, the error message from ParaView after opening the `.pvd` file is that a piece point data array was not found:
```
(   6.404s) [paraview        ]  vtkXMLPDataReader.cxx:359    ERR| vtkXMLPUnstructuredGridReader (0x5595cf53fed0): Piece point data array function_3 not found
```
Also, half of the plot in ParaView (in the case of 2 processes) shows an odd behavior, rather than showing the value 0 consistently.

**Environment:**
 - OS: Linux
 - Python version: 3.11.6
 - ParaView version: 5.11.1

**Additional Info**
If the code from step 2 is changed to generate the `mpi.COMM_SELF` mesh in both processes (which is not what we want), rather than just in process 0, everything seems to work fine.

stephankramer:
[...]

PR #3633, opened 2024-06-14
Implement per communicator universal identifiers for Firedrake objects

### LINKED ISSUE #3631: `utils._new_uid` should be collective over a communicator
As noted in #3628, the `_new_uid` facility can get out of sync across processes when multiple communicators are involved. Although most things in firedrake don't care about the uid, writing data _does_, so we should make sure that if the user doesn't provide names to functions, then `_new_uid` is safe even in a scenario with divergent processes as long as their program is correctly logically collective.

The way to handle this is to adapt `_new_uid` https://github.com/firedrakeproject/firedrake/blob/cb77d32ac8559d8d57fe0ef0efe251d9cbaf2309/firedrake/utils.py#L23

It should take a communicator (since it's only ever called in a context where a communicator is available), pull out the firedrake "internal" communicator (cc @JDBetteridge), and use that to key an id generation. This can be done by attaching an attribute to the communicator that is the `uid` counter. Since the calls to `_new_uid` are collective over a communicator, pulling the uid out of the communicator attribute and updating it will be safe in this scenario.

JDBetteridge:
# Description
This requires https://github.com/NGSolve/ngsPETSc/pull/41 to be merged simultaneously.

Fixes #3631 and potentially #3628.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -83,6 +83,7 @@ jobs:
             --install defcon \
             --install gadopt \
             --install asQ \
+            --package-branch ngsPETSc JDBetteridge/firedrake_uid_changes \
             || (cat firedrake-install.log && /bin/false)
       - name: Install test dependencies
         run: |
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -83,7 +83,6 @@ jobs:
             --install defcon \
             --install gadopt \
             --install asQ \
-            --package-branch ngsPETSc JDBetteridge/firedrake_uid_changes \
             || (cat firedrake-install.log && /bin/false)
       - name: Install test dependencies
         run: |
--- a/firedrake/mesh.py
+++ b/firedrake/mesh.py
@@ -2711,7 +2711,7 @@ def Mesh(meshfile, **kwargs):
                             comm=user_comm)
     mesh = make_mesh_from_mesh_topology(topology, name)
     if netgen and isinstance(meshfile, netgen.libngpy._meshing.Mesh):
-        netgen_firedrake_mesh.createFromTopology(topology, name=plex.getName())
+        netgen_firedrake_mesh.createFromTopology(topology, name=plex.getName(), comm=user_comm)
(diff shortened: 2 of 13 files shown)


firedrake/mesh.py
```diff
@@ -2446,13 +2446,18 @@ def make_mesh_from_coordinates(coordinates, name, tolerance=0.5):
...
+        Communicator.
 
     Returns
     -------
     MeshGeometry
         The mesh.
 
     """
+    if comm is None:
+        raise ValueError("A comm must be provided when creating a mesh from coordinates")
```
wence-, line 2459: Should you not be using the comm attached to the `coordinates`?

firedrake/utils.py
```diff
@@ -6,9 +6,11 @@
 from pyop2.datatypes import RealType     # noqa: F401
 from pyop2.datatypes import IntType      # noqa: F401
 from pyop2.datatypes import as_ctypes    # noqa: F401
+from pyop2.mpi import COMM_WORLD
 from firedrake_configuration import get_config
 
-_current_uid = 0
+# MPI key value for storing a per communicator universal identifier
+FIREDRAKE_UID = COMM_WORLD.Create_keyval()
```
wence-, line 13:
```suggestion
from mpi4py import MPI
...
FIREDRAKE_UID = MPI.Comm.Create_keyval()
```

Then it's clear that the keyval is not specific to a comm?

connorjward (approved):
Looks fine to me.

JDBetteridge:
> Didn't realise was failing CI.

It wasn't before, I need to resurrect my ngsPETSc branch, turns out I needed it after all...

dham (approved):

.github/workflows/build.yml
```diff
@@ -83,6 +83,7 @@ jobs:
             --install defcon \
             --install gadopt \
             --install asQ \
+            --package-branch ngsPETSc JDBetteridge/firedrake_uid_changes \
```
JDBetteridge, line 86:
```suggestion
```

connorjward (approved):

PR #3645, opened 2024-06-21
First idea for absstract reduced functional.

Ig-dolci:
# Description


Ig-dolci:
abstract RD in pyadjoint.

PR #3657, opened 2024-06-28
Get petsc vec

Ig-dolci:
# Description

This PR adds the `OverloadType` method to get the `PETSc.petsc4py.Vec` object.


firedrake/adjoint_utils/function.py
```diff
@@ -398,5 +398,10 @@ def _applyBinary(self, f, y):
             npdata[i] = f(npdata[i], npdatay[i])
         vec.set_local(npdata)
 
+    def _ad_petsc_vec(self):
+        # Return the PETSc Vec object.
+        with self.dat.vec_ro as v:
+            return v
```
connorjward, line 404: I don't think that this is the right thing to do because `return v` eagerly executes the cleanup part of the context manager. I think that this method should also be a context manager. Something like:

```py
@contextlib.contextmanager
def _ad_petsc_vec(self):
    with self.dat.vec_ro as v:
        yield v
```
This then gets called like
```py
with f._ad_petsc_vec() as vec:
    do_something_with_vec(...)
```

Ig-dolci:
Tests are available at [pyadjoint PR 143](https://github.com/dolfin-adjoint/pyadjoint/pull/143).

dham (approved):

Issue #3498, opened 2024-04-10
BUG: Boundary conditions not applied by `LinearVariationalSolver` when the right-hand-side is a `Cofunction` or `FormSum`


jrmaddison:
**Describe the bug**
Boundary conditions are not applied in a linear variational solve when the right-hand-side form is a `Cofunction` or `FormSum`.

**Steps to Reproduce**
```
from firedrake import *

mesh = UnitIntervalMesh(2)
space = FunctionSpace(mesh, "Lagrange", 1)
test, trial = TestFunction(space), TrialFunction(space)

# Form RHS

u = Function(space, name="u")
problem = LinearVariationalProblem(
    inner(trial, test) * dx, inner(Constant(1.0), test) * dx, u,
    DirichletBC(space, 0.0, "on_boundary"))
solver = LinearVariationalSolver(problem)
solver.solve()

print(f"{assemble(inner(u, u) * ds)=}")  # Zero

# Cofunction RHS

b = assemble(inner(Constant(1.0), test) * dx)
u = Function(space, name="u")
problem = LinearVariationalProblem(
    inner(trial, test) * dx, b, u,
    DirichletBC(space, 0.0, "on_boundary"))
solver = LinearVariationalSolver(problem)
solver.solve()

print(f"{assemble(inner(u, u) * ds)=}")  # Nonzero

# FormSum RHS

b = assemble(inner(Constant(0.5), test) * dx) + inner(Constant(0.5), test) * dx
u = Function(space, name="u")
problem = LinearVariationalProblem(
    inner(trial, test) * dx, b, u,
    DirichletBC(space, 0.0, "on_boundary"))
solver = LinearVariationalSolver(problem)
solver.solve()

print(f"{assemble(inner(u, u) * ds)=}")  # Nonzero
```

**Expected behavior**
Boundary conditions should be applied in the solve. In the `Cofunction` case dofs corresponding to boundary evaluation should be ignored.

**Error message**
No error.

**Environment:**
Ubuntu 22.04.

**Additional Info**
The `Cofunction` case appears e.g. when defining a Riesz map on zero trace spaces.

pbrubeck:
The right approach here is to use a `RestrictedFunctionSpace`, if we want to avoid zeroing the boundary dofs of the Cofunction RHS.

jrmaddison:
@pbrubeck Fixed?

pbrubeck:
Fixed in https://github.com/firedrakeproject/firedrake/pull/3947

PR #3662, opened 2024-07-01
Factor the riesz map out into a separate object.

dham:
# Description

### DIFF
--- a/firedrake/adjoint_utils/blocks/solving.py
+++ b/firedrake/adjoint_utils/blocks/solving.py
@@ -756,7 +756,7 @@ class NonlinearVariationalSolveBlock(GenericSolveBlock):
         c = block_variable.output
         c_rep = block_variable.saved_output
 
-        if isinstance(c, firedrake.Function):
+        if isinstance(c, (firedrake.Function, firedrake.Cofunction)):
             trial_function = firedrake.TrialFunction(c.function_space())
         elif isinstance(c, firedrake.Constant):
             mesh = F_form.ufl_domain()
@@ -793,7 +793,11 @@ class NonlinearVariationalSolveBlock(GenericSolveBlock):
         replace_map[self.func] = self.get_outputs()[0].saved_output
         dFdm = replace(dFdm, replace_map)
 
-        dFdm = dFdm * adj_sol
+        if isinstance(dFdm, firedrake.Argument):
+            #  Corner case. Should be fixed more permanenty upstream in UFL.
+            dFdm = ufl.Action(dFdm, adj_sol)
+        else:
+            dFdm = dFdm * adj_sol
         dFdm = firedrake.assemble(dFdm, **self.assemble_kwargs)
 
         return dFdm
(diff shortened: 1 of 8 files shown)


firedrake/adjoint_utils/blocks/solving.py
```diff
@@ -729,7 +729,11 @@ def evaluate_adj_component(self, inputs, adj_inputs, block_variable, idx,
         replace_map[self.func] = self.get_outputs()[0].saved_output
         dFdm = replace(dFdm, replace_map)
 
-        dFdm = dFdm * adj_sol
+        if isinstance(dFdm, firedrake.Argument):
+            #  Corner case. Should be fixed more permanenty upstream in UFL.
```
dham, line 733:
```suggestion
            #  Corner case. Should be fixed more permanently upstream in UFL.
```

firedrake/cofunction.py
```diff
@@ -343,3 +351,120 @@ def __str__(self):
...
+            if value.function_space().dual() != self._function_space:
+                raise ValueError("Function space mismatch in RieszMap.")
+            output = Function(self._function_space)
+
+            if self._inner_product == "l2":
+                for o, c in zip(output.subfunctions, value.subfunctions):
+                    o.dat.data[:] = c.dat.data[:]
+            else:
+                solve, rhs, soln = self._solver
+                rhs.assign(value)
```
dham, line 450: check whether we need to explicitly zero BCs here.

nbouziani (approved):
This looks good to me.

firedrake/cofunction.py
pbrubeck: Aparently we need to do that, https://github.com/firedrakeproject/firedrake/issues/3498

firedrake/cofunction.py
```diff
@@ -360,3 +367,126 @@ def __str__(self):
...
+            raise ValueError("No inner product defined for Sobolev space "
+                             f"{sobolev_space}.")
+
+    @cached_property
+    def _solver(self):
+        from firedrake import (LinearVariationalSolver,
+                               LinearVariationalProblem, Function, Cofunction)
+        rhs = Cofunction(self._function_space.dual())
+        soln = Function(self._function_space)
+        lvp = LinearVariationalProblem(
```
JHopeCollins, line 450: Given that the inner product form won't ever change, I think the LVP should have `constant_jacobian=True` to avoid refactoring every time.

JHopeCollins:
Is there enough information available at setup to cache the `RieszMap` object? Probably on the function space itself.

That way if the same riesz map is needed in different places where it would be cumbersome to share the `RieszMap` (maps in different libraries or in both library and user code) then we still get the advantage of reusing the solver and factorisations.

firedrake/cofunction.py
```diff
@@ -360,3 +367,126 @@ def __str__(self):
...
+            self._inner_product, rhs, soln, bcs=self._bcs, restrict=True,
+            form_compiler_parameters=self._form_compiler_parameters)
+        solver = LinearVariationalSolver(
```
JHopeCollins, line 453: Can we pass an optional `options_prefix` through to the LVS for command line options?

dham:
This is subsumed in #3645

PR #3665, opened 2024-07-03
Fix some TinyASM types

### LINKED ISSUE #3659: BUG: firedrake-install is now failing with --petsc-int-type=int64
**Describe the bug**
The `firedrake-install` command is now failing with `--petsc-int-type=int64`.

**Error message**
The part of the error message that causes the installation to fail is:

```
tinyasm/matinvert.cpp:7:46: error: cannot convert ‘PetscInt*’ {aka ‘long int*’} to ‘const PetscBLASInt*’ {aka ‘const int*’}
```

JDBetteridge:
# Description
TinyASM wasn't building with 64-bit indices.

Fixes #3659

### DIFF
--- a/tinyasm/matinvert.cpp
+++ b/tinyasm/matinvert.cpp
@@ -1,29 +1,19 @@
 #include <petsc.h>
 #include <petscblaslapack.h>
-#include <assert.h>
 
-
-PetscErrorCode mymatinvert(PetscInt* n, PetscScalar* mat, PetscInt* piv, PetscInt* info, PetscScalar* work) {
+PetscErrorCode mymatinvert(PetscInt* n, PetscScalar* mat, PetscBLASInt* piv, PetscInt* info, PetscScalar* work) {
     // Declare BLAS compatible variables
     PetscBLASInt n_blas, info_blas;
-    PetscBLASInt *piv_blas;
 
     // Convert function arguments to BLAS compatible ints
     PetscCall(PetscBLASIntCast(*n, &n_blas));
-    // piv is OUT for getrf and IN for getri so we don't need to check
-    // whether the cast is safe
-    piv_blas = (PetscBLASInt*) piv;
     PetscCall(PetscBLASIntCast(*info, &info_blas));
 
-    assert(*n == n_blas);
+    PetscCallBLAS("LAPACKgetrf", LAPACKgetrf_(&n_blas, &n_blas, mat, &n_blas, piv, &info_blas));
+    PetscCallBLAS("LAPACKgetri", LAPACKgetri_(&n_blas, mat, &n_blas, piv, work, &n_blas, &info_blas));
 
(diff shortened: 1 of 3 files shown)


tinyasm/matinvert.cpp
```diff
@@ -3,8 +3,12 @@
 
 
 PetscErrorCode mymatinvert(PetscInt* n, PetscScalar* mat, PetscInt* piv, PetscInt* info, PetscScalar* work) {
-    PetscCallBLAS("LAPACKgetrf",LAPACKgetrf_(n,n,mat,n,piv,info));
-    PetscCallBLAS("LAPACKgetri",LAPACKgetri_(n,mat, n, piv,work,n,info));
+    PetscBLASInt *n_blas, *piv_blas, *info_blas;
+    PetscCall(PetscBLASIntCast(*n, n_blas));
+    PetscCall(PetscBLASIntCast(*piv, piv_blas));
+    PetscCall(PetscBLASIntCast(*info, info_blas));
```
JHopeCollins, line 9: Should you be casting back to PETScInt after the blas calls?

JHopeCollins:
Can you change the CI on this PR to build with 64 bit int so we know this actually works for that configuration?

tinyasm/matinvert.cpp
JHopeCollins: Info is an out parameter so should definitely be cast back.

piv is out for the factorisation but only in for the inverse so I don't think it's out for `mymatinvert`, so shouldn't need to be cast.

JDBetteridge: I don't think so, I think this routine just checks that the cast is safe:

https://petsc.org/main/manual/blas-lapack/

JDBetteridge:
> Can you change the CI on this PR to build with 64 bit int so we know this actually works for that configuration?

I will do this in a separate PR because I will need to change the docker image

tinyasm/matinvert.cpp
JHopeCollins: Actually it looks like the range of `PETScInt` is always equal or greater than the range of `PETScBLASInt` so you can probably get away with just assigning.

JDBetteridge:
Okay I think this is fixed now. It looks like `tests/regression/test_star_pc.py::test_vanka_equivalence[mixed]` fails with `DIVERGED_LINEAR_SOLVE` when PETSc is compiled with int64 indices, but this appears to be nothing to do with TinyASM.

tinyasm/matinvert.cpp
JDBetteridge: I have changed the way we do this, I believe it to be resolved now

JDBetteridge:
See also https://github.com/firedrakeproject/firedrake/issues/3690

JHopeCollins (approved):
Annoying about the failing tests with int64, but I'll approve this PR seeing as they seem to be unrelated to TinyASM

PR #3669, opened 2024-07-06
Annotate cofunction assign

dham:
Cofunction assignment is not correctly annotated. See the test below, which currently fails.

### DIFF
--- a/tests/regression/test_adjoint_operators.py
+++ b/tests/regression/test_adjoint_operators.py
@@ -997,7 +997,7 @@ def test_cofunction_assign_functional():
 
     cof = assemble(f * v * dx)
     cof2 = Cofunction(cof)
-    cof2.assign(cof)  # Not currently taped!
+    cof2.assign(cof)  # Test is checking that this is taped.
     J = assemble(action(cof2, f2))
     Jhat = ReducedFunctional(J, Control(f))
     assert np.allclose(float(Jhat.derivative()), 1.0)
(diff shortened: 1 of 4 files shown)


tests/regression/test_adjoint_operators.py
```diff
@@ -947,3 +947,23 @@ def test_riesz_representation_for_adjoints():
...
+    fs = FunctionSpace(mesh, "R", 0)
+    f = Function(fs)
+    f.assign(1.0)
+    f2 = Function(fs)
+    f2.assign(1.0)
+    v = TestFunction(fs)
+
+    cof = assemble(f * v * dx)
+    cof2 = Cofunction(cof)
+    cof2.assign(cof)  # Not currently taped!
```
Ig-dolci, line 966: I initially suspected that the test failure was caused by the condition `self != expr` found in this [line of code](https://github.com/firedrakeproject/firedrake/blob/master/firedrake/cofunction.py#L200). However, this was not the case. The root of the problem lies in how the `Cofunction` assignment is handled,  i.e., firedrake does not generate an assignment operation (block). Consequently, in the `pyadjoint` library, the [_find_relevant_nodes](https://github.com/dolfin-adjoint/pyadjoint/blob/master/pyadjoint/tape.py#L139) function is not building the set of relevant dependencies correctly. This issue, together with another [pyadjoint issue](https://github.com/dolfin-adjoint/pyadjoint/issues/144), is leading me to consider implementing an explicit assignment operation (block) for `Cofunction`.

dham: That's right. See the latest commit. 

Unfortunately, this isn't enough because this now reveals an issue with taking the derivative of an action with respect to a cofunction.

Ig-dolci: Ok. I think @dham agrees. 😆

Ig-dolci (approved):

Ig-dolci (approved):

PR #3677, opened 2024-07-09
First order adjoint Jacobian caching

jrmaddison:
# Description

Adds caching of the adjoint Jacobian matrix for the first order adjoint with `LinearVariationalProblem(..., constant_jacobian=True)`.

### DIFF
--- a/firedrake/adjoint_utils/variational_solver.py
+++ b/firedrake/adjoint_utils/variational_solver.py
@@ -46,7 +46,7 @@ class NonlinearVariationalSolverMixin:
             self._ad_args = args
             self._ad_kwargs = kwargs
             self._ad_nlvs = None
-            self._ad_dFdm_cache = {}
+            self._ad_adj_cache = {}
 
         return wrapper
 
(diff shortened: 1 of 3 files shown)


firedrake/adjoint_utils/variational_solver.py
```diff
@@ -133,5 +133,6 @@ def _ad_problem_clone(self, problem, dependencies):
                                            F_replace_map[problem.u_restrict],
                                            bcs=problem.bcs,
                                            J=replace(problem.J, J_replace_map))
+        nlvp._constant_jacobian = problem._constant_jacobian
```
jrmaddison, line 136: I'm not really happy with this. Should a `NonlinearVariationalProblem` set the `_constant_jacobian` attribute?

dham: Why not? This is a case where the NonlinearVariationalProblem is known to actually be linear.

jrmaddison: This is setting an attribute that a `NonlinearVariationalProblem` (base class) doesn't otherwise have.

dham: so how does this work in plain firedrake (ie. not in the adjoint)?

jrmaddison: A `LinearVariationalProblem` does have this attribute. In the adjoint the `LinearVariationalProblem` is always replaced internally with a `NonlinearVariationalProblem`.

dham: Oh, this sounds like a minor subclassing design error. Probably nvlp should always have it and it should be false. I wouldn't be too concerned.

dham (approved):

PR #3684, opened 2024-07-11
Re-enable coverage

JDBetteridge:
# Description
Check to see whether this is still a performance disaster


JDBetteridge:
Yes, yes it is

PR #3687, opened 2024-07-12
Fix Taylor tests in ensemble reduced functional in gather functional case

colinjcotter:
dJdm needs to be reduced over the ensemble communicator to get the correct numbers for the Taylor test.
In the tests, do this first, and then pass in.

### DIFF
--- a/tests/ensemble_reduced_functional/test_reduced_functional.py
+++ b/tests/ensemble_reduced_functional/test_reduced_functional.py
@@ -2,6 +2,7 @@ from firedrake import *
 from firedrake.adjoint import *
 import pytest
 from numpy.testing import assert_allclose
+from pyop2.mpi import MPI
 
 
 @pytest.fixture(autouse=True)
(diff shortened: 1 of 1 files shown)


tests/ensemble_reduced_functional/test_reduced_functional.py
```diff
@@ -63,14 +61,12 @@ def test_verification_gather_functional_adjfloat():
     dJdm = rf.derivative()
     assert_allclose(ensemble_J, 1.0**4+2.0**4, rtol=1e-12)
     assert_allclose(dJdm.dat.data_ro, 4*(rank+1)**3, rtol=1e-12)
-    assert taylor_test(rf, x, Function(R, val=0.1)) > 1.9
+    dJdm = ensemble.ensemble_comm.allreduce(sendobj=dJdm, op=MPI.SUM)
+    assert taylor_test(rf, x, Function(R, val=0.1), dJdm=dJdm) > 1.9
```
dham, line 65: I'm not convinced that this is right. The point of `scatter=False` is that the controls on the different ensemble members are not duplicates of the same variable and hence can't be summed. This simple example is one where they have the same type and hence the sum can be taken, but that changes the meaning of the problem. If you really want to do this the you could just set `scatter=True`.

dham (changes requested):

tests/ensemble_reduced_functional/test_reduced_functional.py
colinjcotter: Uhh that was a Friday night special. I forgot to take the inner product with the perturbation ...

PR #3692, opened 2024-07-17
Purge old easychair links

dham:
Old abstract submission links that are no longer current and sometimes cause the docs build to fail.


JDBetteridge (approved):
Nice one, this was on my to do list! I'll add it to the to done list :grin:

Ig-dolci (approved):

PR #3707, opened 2024-07-25
JDBetteridge/projection_qdegree

JDBetteridge:
# Description
Add quadrature degree to projection operator.

Also update the docstrings to the new style, but might give up because it's a massive time drain :unamused:

### DIFF
--- a/firedrake/projection.py
+++ b/firedrake/projection.py
@@ -2,8 +2,10 @@ import abc
 import ufl
 import finat.ufl
 from ufl.domain import extract_unique_domain
+from typing import Optional, Union
 
 import firedrake
+from firedrake.bcs import BCBase
 from firedrake.petsc import PETSc
 from firedrake.utils import cached_property, complex_mode, SLATE_SUPPORTS_COMPLEX
 from firedrake import functionspaceimpl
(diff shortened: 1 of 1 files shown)


firedrake/projection.py
```diff
@@ -48,32 +50,56 @@ def check_meshes(source, target):
...
-def project(v, V, bcs=None,
-            solver_parameters=None,
-            form_compiler_parameters=None,
-            use_slate_for_inverse=True,
-            name=None,
-            ad_block_tag=None):
-    """Project a UFL expression into a :class:`.FunctionSpace`
+def project(
+    v: ufl.core.expr.Expr,
+    V: Union[firedrake.functionspaceimpl.FunctionSpace, firedrake.Function],
```
ksagiyam, line 55:
```suggestion
    V: Union[firedrake.functionspaceimpl.WithGeometry, firedrake.Function],
```
`firedrake.functionspaceimpl.FunctionSpace` is topological.

firedrake/projection.py
```diff
@@ -48,32 +50,56 @@ def check_meshes(source, target):
...
+    form_compiler_parameters
+        Parameters to the form compiler.
+    use_slate_for_inverse
+        Compute mass inverse cell-wise using SLATE (ignored for non-DG
+        function spaces).
+    name
+        The name of the resulting :class:`.Function`.
+    ad_block_tag
+        String for tagging the resulting block on the Pyadjoint tape.
+
```
ksagiyam, line 85: Docstring for `quadrature_degree`?

ksagiyam (changes requested):

tommbendall:
Thanks very much for implementing this Jack. I think this could be a really useful shorthand for Gusto, particularly when projecting analytic initial conditions (for an example see here https://github.com/firedrakeproject/gusto_case_studies/blob/main/compressible_euler/dry_baroclinic_sphere.py#L253)

I thought I would have a play with this branch, but this raised something that could be an issue. I think whatever is cached as part of the assembly for the projections doesn't depend on quadrature degree, which can give some unexpected results. Here is an example:
```
from firedrake import (
    IntervalMesh, ExtrudedMesh, FunctionSpace, Function, SpatialCoordinate,
    norm, exp, cos, pi
)

height = 5000.0
nlayers = 20

base_mesh = IntervalMesh(5, 10.0)
mesh = ExtrudedMesh(base_mesh, nlayers, height/nlayers)

DG1 = FunctionSpace(mesh, 'DG', 1)
CG1 = FunctionSpace(mesh, 'CG', 1)
CG2 = FunctionSpace(mesh, 'CG', 2)

field1 = Function(CG1)
field2 = Function(CG1)
field3 = Function(CG1)

_, z = SpatialCoordinate(mesh)

# An expression that will spit out TSFC quadrature degree warning
field_expr = exp(-z / height) * cos(20*pi*z/height)

field1.project(field_expr)
field2.project(field_expr, quadrature_degree=(3, 3))
field3.project(field_expr, quadrature_degree=(10, 10))

error = norm(field2 - field1) / norm(field1)
print(error)

error = norm(field2 - field3) / norm(field1)
print(error)
```
The example evaluates an expression through a projection three times, each with different quadrature degrees. However the `error` values are all `0.0`, indicating that the projection is identical in all three cases, when I expect that it shouldn't be. 

I'm not sure that this is the desired behaviour, but I'm also not sure how it would be resolved!

JDBetteridge:
Turns out that had nothing to do with caching, it was just that I forgot to write:
https://github.com/firedrakeproject/firedrake/pull/3707/files#diff-506434f5dc13e6297585641b92851dd978a79df51dc43c922d5d6b35250acc2cR112

If you test again you should see different values.

tommbendall:
> Turns out that had nothing to do with caching, it was just that I forgot to write: https://github.com/firedrakeproject/firedrake/pull/3707/files#diff-506434f5dc13e6297585641b92851dd978a79df51dc43c922d5d6b35250acc2cR112
> 
> If you test again you should see different values.

Yes this seems to give the expected results for me now, thanks!

dham (approved):

PR #3719, opened 2024-07-31
PatchPC: Log flops

pbrubeck:
# Description
Logging flops is typically handled by the parloop, but PCPatch does not call a parloop.

Instead, we could log flops within the Jacobian/Residual wrapper, this PR will log them under the `PCPATCHComputeOp` event, but not under the `Log_Event_subspace_form_cell_integral`, and it would be preferable to see them there as well.

### DIFF
--- a/firedrake/preconditioners/patch.py
+++ b/firedrake/preconditioners/patch.py
@@ -23,7 +23,6 @@ import ctypes
 from pyop2 import op2
 from pyop2.mpi import COMM_SELF
 import pyop2.types
-import pyop2.parloop
 from pyop2.compilation import load
 from pyop2.utils import get_petsc_dir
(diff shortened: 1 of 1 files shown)


firedrake/preconditioners/patch.py
```diff
@@ -436,12 +435,13 @@ def make_residual_wrapper(coeffs, maps):
    if (x) {{
      ierr = VecRestoreArrayRead(x, &state);CHKERRQ(ierr);
    }}
+   PetscLogFlops({} * npoints);
```
connorjward, line 438: Would it not be better to just omit this line if `flops` is not provided? I think defaulting to zero is potentially a bit confusing.

pbrubeck: Let's not make it optional then.

dham (approved):

PR #3725, opened 2024-08-05
Fix up things in the ensemble tests

JDBetteridge:
# Description
 - Parallel assertion

### DIFF
--- a/tests/regression/test_ensembleparallelism.py
+++ b/tests/regression/test_ensembleparallelism.py
@@ -25,6 +25,24 @@ blocking = [pytest.param(True, id="blocking"),
             pytest.param(False, id="nonblocking")]
 
 
+def parallel_assert(assertion, subset=None, msg=""):
+    """ Move this functionality to pytest-mpi
+    """
+    if subset:
+        if MPI.COMM_WORLD.rank in subset:
+            evaluation = assertion()
+        else:
+            evaluation = True
+    else:
+        evaluation = assertion()
+    all_results = MPI.COMM_WORLD.allgather(evaluation)
+    if not min(all_results):
+        raise AssertionError(
+            "Parallel assertion failed on ranks: "
+            f"{[ii for ii, b in enumerate(all_results) if not b]}\n" + msg
+        )
+
+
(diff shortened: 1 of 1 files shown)


tests/regression/test_ensembleparallelism.py
```diff
@@ -25,6 +25,24 @@
...
+    """
+    if subset:
+        if MPI.COMM_WORLD.rank in subset:
+            evaluation = assertion()
+        else:
+            evaluation = True
+    else:
+        evaluation = assertion()
+    all_results = MPI.COMM_WORLD.allgather(evaluation)
+    if not min(all_results):
```
JHopeCollins, line 39:
```suggestion
    if not all(all_results):
```
`Comm.allgather` doesn't convert bools to ints.

tests/regression/test_ensembleparallelism.py
```diff
@@ -297,22 +326,29 @@ def test_send_and_recv(ensemble, mesh, W, blocking):
...
-
         if not blocking:
             MPI.Request.waitall(send_requests)
             MPI.Request.waitall(recv_requests)
+        error = errornorm(urecv, usend)
+    else:
+        error = 0
 
-        assert errornorm(urecv, usend) < 1e-12
+    # Test send/recv etween first two spatial comms
```
JHopeCollins, line 343:
```suggestion
    # Test send/recv between first two spatial comms
```

tests/regression/test_ensembleparallelism.py
```diff
@@ -25,6 +25,24 @@
             pytest.param(False, id="nonblocking")]
 
 
+def parallel_assert(assertion, subset=None, msg=""):
```
JHopeCollins, line 28: Why do you want to pass `assertion` as a callable and not just as a bool?

JHopeCollins (changes requested):

tests/regression/test_ensembleparallelism.py
JHopeCollins: This can be discussed in the pytest-mpi PR

JHopeCollins (approved):

PR #3478, opened 2024-03-27
Allow for solving multi-domain problems involving codim-0 submeshes

ksagiyam:
Depends on:
~UFL: https://github.com/FEniCS/ufl/pull/264~ Merged.

Allow for solving multi-domain problems involving codim-0 submeshes.

PR #3727, opened 2024-08-07
submesh: enable cell submesh (submeshing only)

ksagiyam:
First half of the Firedrake submesh PR: https://github.com/firedrakeproject/firedrake/pull/3478

- Submeshing.
- Interpolating between submeshes.


firedrake/mesh.py
```diff
@@ -4161,3 +4478,76 @@ def SubDomainData(geometric_expr):
...
+    """Construct a submesh from a given mesh.
+
+    Parameters
+    ----------
+    mesh : MeshGeometry
+        Parent mesh (`MeshGeometry`).
+    subdim : int
+        Topological dimension of the submesh.
+    subdomain_id : int
+        Subdomain ID representing the submesh.
```
dham, line 4493: document the limitation

dham (changes requested):

firedrake/mesh.py
ksagiyam: Done.

PR #3730, opened 2024-08-12
JDBetteridge/update caching

JDBetteridge:
# Description
Very early stages WIP, replacing all dodgy caches with parallel safe ones.


firedrake/slate/slac/compiler.py
```diff
@@ -66,19 +67,37 @@
...
 
-class SlateKernel(TSFCKernel):
-    @classmethod
-    def _cache_key(cls, expr, compiler_parameters):
-        return md5(
-            (expr.expression_hash + str(sorted(compiler_parameters.items()))).encode()).hexdigest(), expr.ufl_domains()[0].comm
+def _cache_key(expr, compiler_parameters):
+    return expr.ufl_domains()[0].comm, md5(
+        (expr.expression_hash + str(sorted(compiler_parameters.items()))).encode()
+    ).hexdigest()
```
dham, line 73: is this actually used?

firedrake/tsfc_interface.py
```diff
@@ -52,75 +47,33 @@
...
-                    + str(sorted(parameters.items()))
-                    + str(coefficient_numbers)
-                    + str(constant_numbers)
-                    + str(type(interface))
-                    + str(diagonal)).encode()).hexdigest(), form.ufl_domains()[0].comm
+def tsfc_compile_form_hashkey(form, prefix, parameters, interface, diagonal, log):
+    # Drop prefix as it's only used for naming and log
+    # JBTODO: Can't drop prefix as tests/slate/test_optimise.py::test_partially_optimised fails, investigate
+    # it looks like the prefix is being used to create different subkernels, which conflicts with the docstring below
+    return default_parallel_hashkey(form.signature(), prefix, parameters, interface, diagonal)
```
dham, line 60: not sure I understand the change. Why are we allowed to drop `coefficient_numbers`?

dham (changes requested):
I've made a couple of comments but I think I need talking through this.

dham (dismissed):

firedrake/slate/slac/compiler.py
JDBetteridge: Removed

.github/workflows/build.yml
```diff
@@ -83,6 +83,7 @@ jobs:
             --install defcon \
             --install gadopt \
             --install asQ \
+            --package-branch PyOP2 JDBetteridge/remove_comm_hash \
```
JDBetteridge, line 86:
```suggestion
```

firedrake/tsfc_interface.py
connorjward: `coefficient_numbers` is actually derived directly from `form` so it doesn't need to be included. Also this cache now sits at a slightly different point in the stack so `coefficient_numbers` isn't even an argument to the wrapped function.

JDBetteridge:
I think this one is ready to go (once tests pass of course!)

PR #3734, opened 2024-08-14
Fix TransferManager reuse for MixedFunctionSpace

pbrubeck:
# Description

### DIFF
--- a/firedrake/variational_solver.py
+++ b/firedrake/variational_solver.py
@@ -300,7 +300,7 @@ class NonlinearVariationalSolver(OptionsManager, NonlinearVariationalSolverMixin
         coefficients = utils.unique(chain.from_iterable(form.coefficients() for form in forms if form is not None))
         # Make sure the solution dm is visited last
         solution_dm = self.snes.getDM()
-        problem_dms = [V.dm for V in utils.unique(c.function_space() for c in coefficients) if V.dm != solution_dm]
+        problem_dms = [V.dm for V in utils.unique(chain.from_iterable(c.function_space() for c in coefficients)) if V.dm != solution_dm]
         problem_dms.append(solution_dm)
 
         for dbc in problem.dirichlet_bcs():
--- a/tests/regression/test_star_pc.py
+++ b/tests/regression/test_star_pc.py
@@ -1,4 +1,5 @@
 import pytest
+import warnings
 from firedrake import *
 from firedrake.petsc import DEFAULT_DIRECT_SOLVER
 try:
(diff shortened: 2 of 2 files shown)


firedrake/variational_solver.py
```diff
@@ -300,7 +300,7 @@ def solve(self, bounds=None):
         coefficients = utils.unique(chain.from_iterable(form.coefficients() for form in forms if form is not None))
         # Make sure the solution dm is visited last
         solution_dm = self.snes.getDM()
-        problem_dms = [V.dm for V in utils.unique(c.function_space() for c in coefficients) if V.dm != solution_dm]
+        problem_dms = [V.dm for V in utils.unique(chain.from_iterable(c.function_space() for c in coefficients)) if V.dm != solution_dm]
```
JDBetteridge, line 303: So that I understand better: The reason we need `chain.from_iterable` is because `c.function_space()` can be an iterable when dealing with mixed function spaces?

tests/regression/test_star_pc.py
```diff
@@ -20,6 +21,12 @@ def backend(request):
     return request.param
 
 
+def filter_warnings(caller):
+    with warnings.catch_warnings():
+        warnings.filterwarnings("error", "Creating new TransferManager", RuntimeWarning)
+        caller()
```
JDBetteridge, line 27: Can you leave a comment explaining exactly what this code is doing. It looks like it is going to have unintended consequences.

From my reading of the Python docs this function catches and suppresses all warnings, except for those that match "Creating new TransferManager" in the message component, in which case an exception is raised instead. However, I might have misunderstood the behaviour.

JDBetteridge (dismissed):

firedrake/variational_solver.py
pbrubeck: Correct, the transfer manager is attached to an appctx on the dm of each individual field.

tests/regression/test_star_pc.py
pbrubeck: This will still raise other warnings without supressing them.

```
import warnings

def filter_warnings(caller):
        warnings.filterwarnings("error", "Creating new TransferManager", RuntimeWarning)
        caller()

def fatal_warning():
    warnings.warn("Creating new TransferManager to transfer data to coarse grids", RuntimeWarning)

def non_fatal_warning():
    warnings.warn("Should be raised", RuntimeWarning)

filter_warnings(non_fatal_warning)
print("That was just a warning")

filter_warnings(fatal_warning)
print("We should not be here")
```
Gives the expected output

```
/home/brubeckmarti/Downloads/warnings.py:7: RuntimeWarning: Should be raised
  warnings.warn("Should be raised", RuntimeWarning)
That was just a warning
  File "/home/brubeckmarti/Downloads/warnings.py", line 19, in <module>
    filter_warnings(fatal_warning)
  File "/home/brubeckmarti/Downloads/warnings.py", line 13, in filter_warnings
    caller()
  File "/home/brubeckmarti/Downloads/warnings.py", line 4, in fatal_warning
    warnings.warn("Creating new TransferManager to transfer data to coarse grids", RuntimeWarning)
RuntimeWarning: Creating new TransferManager to transfer data to coarse grids

```

dham (approved):

PR #3743, opened 2024-08-22
allow for ghost only mesh

ksagiyam:
Allow for ghost only meshes.

`plex.pointSF.ilocal` is an array of leaf points on each rank. If the array is trivial (i.e., `ilocal = [0, 1, 2, ...]`), `ilocal` is set to `NULL`. So typically we use this array as `point = ilocal[i] if ilocal else i`, but in some places we simply do `point = ilocal[I]`. This is fine if a rank at least owns one cell (`ilocal` is nontrivial due to plex point numbering convention) or does not have cells at all, but not fine if the rank sees and only sees ghost cells (`ilocal` is trivial, so `ilocal` becomes `NULL`) . It turned out that we need to handle the latter case if submeshes are involved [the submesh PR](https://github.com/firedrakeproject/firedrake/pull/3478).

### DIFF
--- a/firedrake/cython/dmcommon.pyx
+++ b/firedrake/cython/dmcommon.pyx
@@ -2471,7 +2471,7 @@ def get_cell_remote_ranks(PETSc.DM plex):
     :arg plex: The DMPlex object encapsulating the mesh topology
     """
     cdef:
-        PetscInt cStart, cEnd, ncells, i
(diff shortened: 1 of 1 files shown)


connorjward (approved):
Looks fine. I assume in Cython it is fine to replace `if ilocal != NULL` with `if ilocal`.

PR #3754, opened 2024-09-02
Apply Diriclet boundary on the Cofunction RHS.

Ig-dolci:
# Description

### DIFF
--- a/firedrake/variational_solver.py
+++ b/firedrake/variational_solver.py
@@ -10,6 +10,7 @@ from firedrake.petsc import (
     DEFAULT_SNES_PARAMETERS
 )
 from firedrake.function import Function
+from firedrake.cofunction import Cofunction
 from firedrake.functionspace import RestrictedFunctionSpace
 from firedrake.ufl_expr import TrialFunction, TestFunction
 from firedrake.bcs import DirichletBC, EquationBC
@@ -305,6 +306,10 @@ class NonlinearVariationalSolver(OptionsManager, NonlinearVariationalSolverMixin
 
         for dbc in problem.dirichlet_bcs():
             dbc.apply(problem.u_restrict)
+            for coeff in coefficients:
+                if isinstance(coeff, Cofunction):
+                    # Apply the DirichletBC to the right hand side of the equation.
+                    dbc.apply(coeff)
 
         if bounds is not None:
             lower, upper = bounds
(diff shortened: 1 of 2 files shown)


firedrake/variational_solver.py
```diff
@@ -305,6 +306,10 @@ def solve(self, bounds=None):
 
         for dbc in problem.dirichlet_bcs():
             dbc.apply(problem.u_restrict)
+            for coeff in coefficients:
+                if isinstance(coeff, Cofunction):
+                    # Apply the DirichletBC to the right hand side of the equation.
+                    dbc.apply(coeff)
```
pbrubeck, line 312: This does not look right. The residual form might contain many cofunctions in function spaces different from the one of the solution. This should be done before calling the solver, and only on the Cofunction RHS. Also it is likely that you need dbc.zero() instead of dbc.apply(). Also see the comment I left in https://github.com/firedrakeproject/firedrake/pull/3662

tests/regression/test_cofunction.py
```diff
@@ -60,3 +60,43 @@ def test_scalar_cofunction_zero_with_subset(V):
...
+    solver.solve()
+
+    assert np.allclose(assemble(inner(u, u) * ds), 0.0)
+
+    # FormSum RHS
+    b = assemble(inner(Constant(0.5), test) * dx) + inner(Constant(0.5), test) * dx
+    u = Function(space, name="u")
+    problem = LinearVariationalProblem(
+        inner(trial, test) * dx, b, u,
+        DirichletBC(space, 0.0, "on_boundary"))
```
pbrubeck, line 98: The test should have non-homogeneous bcs

firedrake/variational_solver.py
jrmaddison, line 312: This will have side effects on `Cofunction`s in the problem definition.
```
u = Function(space, name="u")
b = Cofunction(space.dual(), name="b")
...
problem = LinearVariationalProblem(
    inner(trial, test) * dx, b, u, DirichletBC(space, 1, "on_boundary"))
assert b in problem.F.coefficients()  # passes
```

jrmaddison:
To address #3498?

Ig-dolci:
> To address #3498?

Yes. It is just a starting point. I still do not know how to fix it.

pbrubeck:
Closing, as superseeded by https://github.com/firedrakeproject/firedrake/pull/3922

PR #3394, opened 2024-02-07
External operators

nbouziani:
This PR contains the Firedrake implementation of external operators. The corresponding UFL implementation was merged in https://github.com/FEniCS/ufl/pull/206. The Firedrake infrastructure to make external operators has already been introduced in #2297, making that PR pretty "small".

The external operator interface allows for incorporating arbitrary operators mapping between finite element spaces in Firedrake. Those operators are external in the sense that their implementation is left to be specified by the user. This PR mainly contains:

- The assembly mechanism required for evaluating external operators, and expressions/variational forms containing external operators.
- The abstract external operator class that orchestrates the assembly
- The ML operator which allows for incorporating machine learning models implemented in PyTorch in Firedrake
- A new manual chapter

PR #3759, opened 2024-09-09
Add JAX coupling

nbouziani:
This PR adds support to couple JAX and Firedrake in a differentiable fashion. It extends the previous work on coupling Firedrake and PyTorch (cf. #3394 and #2804) to support the JAX framework as well as downstream libraries. It allows to embed JAX model into Firedrake, thanks to the `MLOperator` external operator class, and to embed Firedrake system into JAX. These new capabilities can be obtained by importing from the `firedrake.ml.jax` module. It also adds options to install JAX on the fly when installing Firedrake. Main additions:

- Add a JAX custom operator (`FiredrakeJaxOperator`) to represent Firedrake operators expressed as a ReducedFunctional, wherein forward and backward computations are delegated to the reduced functional.
- Add a new ml operator (`JaxOperator`) to represent JAX operations embedded in Firedrake
- Add support for installing JAX
- Add mapping operator from JAX to Firedrake, and vice-versa
- Some tests

### DIFF
--- a/firedrake/ml/jax/fem_operator.py
+++ b/firedrake/ml/jax/fem_operator.py
@@ -19,10 +19,12 @@ except ImportError:
 import collections
 import numpy as np
 from functools import partial
+from typing import Union, Optional
 
 from firedrake.function import Function
 from firedrake.cofunction import Cofunction
 from firedrake.vector import Vector
+from firedrake.functionspaceimpl import WithGeometry
 from firedrake.constant import Constant
 from firedrake_citations import Citations
 
@@ -39,11 +41,11 @@ class FiredrakeJaxOperator:
 
     Parameters
     ----------
-    F : pyadjoint.ReducedFunctional
+    F
         The reduced functional to wrap.
     """
 
-    def __init__(self, F):
+    def __init__(self, F: ReducedFunctional):
         super(FiredrakeJaxOperator, self).__init__()
         self.F = F
         self.V_controls = [c.control.function_space() for c in F.controls]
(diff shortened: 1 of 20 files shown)


firedrake/ml/jax/fem_operator.py
```diff
@@ -0,0 +1,228 @@
...
+    F : pyadjoint.ReducedFunctional
+        The reduced functional to wrap.
+
+    Returns
+    -------
+    firedrake.ml.jax.fem_operator.FiredrakeJaxOperator
+        A JAX custom operator that wraps the reduced functional `F`.
+    """
+    # TODO: Add citation
+    # Citations().register(TODO)
```
nbouziani, line 115: This will come shortly.

firedrake/ml/jax/fem_operator.py
```diff
@@ -0,0 +1,228 @@
...
+
+    if not isinstance(F, ReducedFunctional):
+        raise ValueError("F must be a ReducedFunctional")
+
+    jax_op = FiredrakeJaxOperator(F)
+    return partial(FiredrakeJaxOperator.forward, jax_op)
```
nbouziani, line 121: Add a comment to explain why this workaround exists

JDBetteridge (changes requested):
Please use type hinting as per the "new" policy:
https://github.com/firedrakeproject/firedrake/wiki/Docstrings

JDBetteridge (approved):
Docstrings and function signatures have been updated to use type hinting.

PR #3778, opened 2024-09-23
Move both ML frameworks from vanilla to firedrake

JDBetteridge:
# Description


JHopeCollins (approved):
LGTM, this seems like the sensible thing to do.

PR #3795, opened 2024-10-10
Test Stokes macroelements

pbrubeck:
# Description
Tests for the macroelement Stokes complexes using Guzman-Neilan, Alfeld-Sorokina, and Scott-Vogelius macroelements. We test that the divergence of the velocity is contained by the pressure space, by simply interpolating the divergence of basis functions.


rckirby (approved):
looks good to me once everything else has landed and we re-point to the main branches.

dham (approved):

PR #3798, opened 2024-10-14
Fix for AssembledMatrix

connorjward:
# Description

Previously it was not possible to used an `AssembledMatrix` as the LHS for a solver.

I have hacked around with `_arguments` to get things to work and I am not completely confident in what I have done. I would appreciate someone familiar with the dual space changes to take a look.

### DIFF
--- a/firedrake/matrix.py
+++ b/firedrake/matrix.py
@@ -34,6 +34,7 @@ class MatrixBase(ufl.Matrix):
         # (so we can't use a set, since the iteration order may differ
         # on different processes)
 
+        breakpoint()
         ufl.Matrix.__init__(self, test.ufl_function_space(), trial.ufl_function_space())
         # Define arguments after `Matrix.__init__` since BaseForm sets `self._arguments` to None
         self._arguments = arguments
--- a/firedrake/matrix.py
+++ b/firedrake/matrix.py
@@ -35,13 +35,13 @@ class MatrixBase(ufl.Matrix):
         # on different processes)
 
         ufl.Matrix.__init__(self, test.ufl_function_space(), trial.ufl_function_space())
-        # Define arguments after `Matrix.__init__` since BaseForm sets `self._arguments` to None
-        self._arguments = arguments
+        # NOTE: Do not use _arguments as the name for this variable since it is reserved by UFL
+        self._firedrake_arguments = arguments
         self.bcs = bcs
-        self.comm = test.ufl_function_space().comm
+        self.comm = test.function_space().comm
         self._comm = internal_comm(self.comm, self)
-        self.block_shape = (len(test.ufl_function_space()),
-                            len(trial.ufl_function_space()))
+        self.block_shape = (len(test.function_space()),
+                            len(trial.function_space()))
         self.mat_type = mat_type
         """Matrix type.
 
@@ -52,7 +52,7 @@ class MatrixBase(ufl.Matrix):
         if self.a:
             return self.a.arguments()
         else:
-            return self._arguments
+            return self._firedrake_arguments
 
     @property
     def has_bcs(self):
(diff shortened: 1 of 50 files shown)


firedrake/matrix.py
```diff
@@ -34,9 +34,9 @@ def __init__(self, a, bcs, mat_type):
         # (so we can't use a set, since the iteration order may differ
         # on different processes)
 
-        ufl.Matrix.__init__(self, test.function_space(), trial.function_space())
-        # Define arguments after `Matrix.__init__` since BaseForm sets `self._arguments` to None
-        self._arguments = arguments
+        ufl.Matrix.__init__(self, test.ufl_function_space(), trial.ufl_function_space())
```
ksagiyam, line 37: Are there reasons to use `ufl_function_space()` over `function_space()`?

firedrake/matrix.py
```diff
@@ -34,9 +34,9 @@ def __init__(self, a, bcs, mat_type):
...
+        # NOTE: Do not use _arguments as the name for this variable since it is reserved by UFL
+        self._firedrake_arguments = arguments
```
ksagiyam, line 39: This might suggest that we want to pass `Argument`s to `ufl.Matrix` instead of `FunctionSpace`s, but until we update our UFL fork, a better fix probably is to overwrite `_analyze_form_arguments()` method as:
```
    def _analyze_form_arguments(self):
        self._arguments = (firedrake.Argument(self._ufl_function_spaces[0], 0),
                           firedrake.Argument(self._ufl_function_spaces[1], 1))
        self._coefficients = ()
```
Then, we no longer need to overwrite `arguments()` method here.

firedrake/matrix.py
```diff
@@ -204,3 +209,6 @@ def __add__(self, other):
         else:
             raise TypeError("Unable to add %s to AssembledMatrix"
                             % (type(other)))
+
+    def ufl_domains(self):
+        return ufl.domain.join_domains(self.a[0].ufl_domains() + self.a[1].ufl_domains())
```
ksagiyam, line 214: This can be:
```
    def ufl_domains(self):
        return self._domains
````
This can be done in `MatrixBase`, and should eventually be done upstream.

File: tests/regression/test_matrix.py [hunk 4]
```diff
@@ -27,3 +34,38 @@ def test_assemble_returns_matrix(a):
...
+    f = Function(V).interpolate(x)
+
+    a = inner(u, v) * dx
+    A = AssembledMatrix((v, u), bcs=(), petscmat=assemble(a).M.handle)
+    L = inner(f, v) * dx
+
+    solution = Function(V)
+    solve(A == L, solution)
+
+    assert norm(assemble(f - solution)) < 1e-6
```
ksagiyam, line 55:
```suggestion
    assert norm(assemble(f - solution)) < 1e-15
```

ksagiyam (changes requested):

firedrake/matrix.py
connorjward: No good spot. That was from when I was debugging the `_arguments` issue. Will revert.

firedrake/matrix.py
connorjward: As in you think the UFL API for `Matrix` should be changed to take arguments instead of function spaces?

connorjward: More radical counter suggestion: I don't see why we need to have our own `Argument` class. Why don't we just remove it? It seems like all of the new bits we add to `Argument` (e.g. `cell_node_map`) are just passed through from the function space anyway (e.g. https://github.com/firedrakeproject/firedrake/blob/master/firedrake/ufl_expr.py#L54) so we can just replace `arg.cell_node_map` with `arg.ufl_function_space().cell_node_map` where necessary.

As an aside this exposes a general design flaw in UFL, which I think is generally referred to as the "expression problem". If UFL exposes algorithms and abstract interfaces then things are fine, but as soon as UFL instantiates its own non-subclassed objects then we get weird inconsistencies. This is the same reason why we need `expr.subfunctions` as well as `split(expr)`. The latter returns an internal UFL `Indexed` object that we cannot intercept and replace with our own class.

ksagiyam (approved):

PR #3799, opened 2024-10-15
JDBetteridge/randomfields rebased

JDBetteridge:
# Description
Anyone remember this fossil?

This PR allows users to sample functions from some Gaussian random field with Matern covariance.

I'm no better at statistics than when I wrote this code (a long time ago) and someone should probably check the properties.

Still needs some tests and there has to be a better way of doing the boundary thing* @connorjward? @UZerbinati does netgen handle this kind of thing?

* For meshes with a boundary, the mesh needs to be grown by the size of the correlation length and then needs to be trimmed back to the original mesh. I think this might be quite a hard thing to do in general.

### DIFF
--- a/firedrake/__init__.py
+++ b/firedrake/__init__.py
@@ -119,6 +119,7 @@ from firedrake.ensemble import *
 from firedrake.randomfunctiongen import *
 from firedrake.external_operators import *
 from firedrake.progress_bar import ProgressBar  # noqa: F401
+from firedrake.randomfields import *
 
 from firedrake.logging import *
 # Set default log level
--- a/firedrake/randomfields.py
+++ b/firedrake/randomfields.py
@@ -1,6 +1,6 @@
 import numpy as np
 
-from firedrake.assemble import vector_arg
+from firedrake.assemble import _vector_arg
 from firedrake.constant import Constant
 from firedrake.function import Function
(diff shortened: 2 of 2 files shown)


UZerbinati:
I think we might be able to do this with Netgen mesh extrusion (https://docu.ngsolve.org/latest/i-tutorials/unit-4.4-occ/workplane.html) or using the Boundary Layer method.

connorjward:
I think extending the boundary regions is a meshing/geometry problem which Firedrake isn't very well-suited to fixing. Netgen or GMSH are probably the way to go.

I also wonder if @ksagiyam's submesh stuff would be useful for handling the inner mesh.

connorjward:
This was discussed in the latest [Firedrake meeting](https://github.com/firedrakeproject/firedrake/wiki/Firedrake-meeting-2024-10-16#jb-an-ancient-branch-rebased).

@JDBetteridge are you looking to work on this yourself? Or are you hoping someone else can take over? If the latter then it would be extremely useful if you could provide some tests (that don't necessarily need to pass) that demonstrate usage and expected behaviour.

JDBetteridge:
I'm intending to hand this over, but my intention is to provide tests. I have some tests already, but they need refactoring into py-test tests. Currently they are trying to assess statistics of these fields and are not suitable for the test suite as they take far too long to run!

PR #3804, opened 2024-10-18
Fix point location in parallel

### LINKED ISSUE #3151: BUG: Cross mesh interpolation breaks for parallel nested meshes

connorjward:
# Description

I believe that this fixes #3151 and supersedes #3293. It certainly seems to fix this issue on my machine.


connorjward:
Turns out that this is harder than expected. With my changes, if we consider the interval mesh
```
           p0   |      p1
        o-----o | -----o-----o
                |
```
and want to locate points `A` and `B` on `p0` (note the extra cell from `p1` is not visible)
```
           p0   |   p1
        o-----o | -----o
                |  A     B
```
then our [strategy for identifying missing points](https://github.com/firedrakeproject/firedrake/blob/master/firedrake/mesh.py#L4134) means that both `A` and `B` are considered as being claimed by 'cells that we can't see' and both are removed. This is clearly wrong.

Unfortunately I can't currently think of a way to fix this in a non-invasive manner at present. I do have some general thoughts:
* Point location should probably not be being done for ghost cells. IMO those should be handled by the other process and communicated over.
* We should look to use star forests more as the existing approach, with each process seeing all points, is not very scalable.

connorjward:
Closing as this was poorly thought out (though in my defence this is extremely confusing). I will attempt to get #3293 merged.

PR #3814, opened 2024-10-21
Fix coarsening for R space

pbrubeck:
# Description
This PR fixes two separate issues for multigrid with `Functions` in R.

1. The restriction operator for the R space is square. For these cases PETSc complains if we do not pass a vector to rescale the restriction operator to construct the "cheap" injection, even though it is not necessary, as we already pass the proper injection operator. To fix this we just pass a vector of the appropiate size.

2. The `parentdm` of an R space function sometimes might have a null `__setup_hooks__`, which makes it impossible to call `add_hook` on it. As a workaround, we just skip that call.

### DIFF
--- a/firedrake/mg/ufl_utils.py
+++ b/firedrake/mg/ufl_utils.py
@@ -268,23 +268,21 @@ def coarsen_snescontext(context, self, coefficient_mapping=None):
     coarse._fine = context
     context._coarse = coarse
 
-    solution = context._problem.u
-    solutiondm = solution.function_space().dm
+    solutiondm = context._problem.u.function_space().dm
     parentdm = get_parent(solutiondm)
-    if parentdm != solutiondm:
-        # Now that we have the coarse snescontext, push it to the coarsened DMs
-        # Otherwise they won't have the right transfer manager when they are
-        # coarsened in turn
-        for val in chain(coefficient_mapping.values(), (bc.function_arg for bc in problem.bcs)):
-            if isinstance(val, (firedrake.Function, firedrake.Cofunction)):
-                V = val.function_space()
-                coarseneddm = V.dm
+    # Now that we have the coarse snescontext, push it to the coarsened DMs
+    # Otherwise they won't have the right transfer manager when they are
+    # coarsened in turn
+    for val in chain(coefficient_mapping.values(), (bc.function_arg for bc in problem.bcs)):
+        if isinstance(val, (firedrake.Function, firedrake.Cofunction)):
+            V = val.function_space()
+            coarseneddm = V.dm
 
-                # Now attach the hook to the parent DM
-                if get_appctx(coarseneddm) is None:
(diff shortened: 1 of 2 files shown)


firedrake/mg/ufl_utils.py
```diff
@@ -384,7 +385,8 @@ def create_interpolation(dmc, dmf):
     mat.setType(mat.Type.PYTHON)
     mat.setPythonContext(ctx)
     mat.setUp()
-    return mat, None
+    rscale = mat.createVecLeft() if row_size == col_size else None
```
pbrubeck, line 388: Not sure wether Right or Left is the correct thing here. But it does not really matter if the sizes are the same.

tests/multigrid/test_transfer_manager.py
```diff
@@ -131,3 +132,25 @@ def test_transfer_manager_dat_version_cache(action, transfer_op, spaces):
...
+    hierarchy = MeshHierarchy(base, 3)
+    mesh = hierarchy[-1]
+
+    V = FunctionSpace(mesh, family, degree)
+    u = Function(V)
+    F = inner(u - 1, TestFunction(V)) * dx
+
+    # This test will fail if we raise this warning
+    with warnings.catch_warnings():
+        warnings.filterwarnings("error", "Creating new TransferManager", RuntimeWarning)
```
connorjward, line 155: This looks very fragile. If we decided not to emit this warning any more, or even changed the content of the warning message, then this would always pass regardless of whether or not we were actually reusing transfer managers.

Is a better way to do this not to somehow retrieve the transfer managers from whatever mesh/DM they are attached to and check that they are the same?

connorjward (changes requested):

firedrake/mg/ufl_utils.py
```diff
@@ -406,27 +405,43 @@ def create_interpolation(dmc, dmf):
...
 
+    manager = get_transfer_manager(dmf)
     ctx = Interpolation(V_c, V_f, manager, cbcs, fbcs)
     mat = PETSc.Mat().create(comm=dmc.comm)
     mat.setSizes((row_size, col_size))
     mat.setType(mat.Type.PYTHON)
     mat.setPythonContext(ctx)
     mat.setUp()
-    return mat, None
+    if row_size == col_size:
```
connorjward, line 415: This is totally non-obvious. Please explain with a comment.

firedrake/mg/ufl_utils.py
```diff
@@ -406,27 +405,43 @@ def create_interpolation(dmc, dmf):
...
+        rscale = mat.createVecRight()
+        rscale.set(float("nan"))
```
```suggestion
    if row_size == col_size:
        # PETSc cannot determine the coarse space if the dimensions are equal.
        # The coarse space is identified by the dimension of rscale, so we provide one.
        rscale = mat.createVecRight()
        rscale.set(1.0)
```

PR #3815, opened 2024-10-21
We can rely on having tinyasm installed

pbrubeck:
# Description
Just skipping checks for tinyasm, as it is now a hard dependency.


connorjward (approved):
Thanks!

PR #3841, opened 2024-10-25
annotate `FunctionMixin` methods  `_ad_mul` and `_ad_add`

JHopeCollins:
These are not marked `no_annotate` in the `pyadjoint.OverloadedType` abstract methods.

### DIFF
--- a/firedrake/adjoint_utils/function.py
+++ b/firedrake/adjoint_utils/function.py
@@ -300,7 +300,6 @@ class FunctionMixin(FloatingType):
         with checkpoint_init_data():
             super()._ad_will_add_as_dependency()
 
-    @no_annotations
     def _ad_mul(self, other):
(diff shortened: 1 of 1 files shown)


JHopeCollins:
Even without `no_annotations` I don't think these add anything to the tape because they are implemented using `Function.vector()`. Should they be updated to just use existing `Function` operations?

connorjward:
> Even without `no_annotations` I don't think these add anything to the tape because they are implemented using `Function.vector()`. Should they be updated to just use existing `Function` operations?

Anything with `.vector()` is deprecated functionality and should ideally get replaced with Firedrake-specific operations. `Vector` was a shim class for FEniCS compatibility that has now been abandoned.

connorjward (approved):
All looks reasonable.

PR #3842, opened 2024-10-25
Fix preconditioner when using EquationBC

ksagiyam:
Fix `EquationBC` in PCs.

### DIFF
--- a/firedrake/bcs.py
+++ b/firedrake/bcs.py
@@ -459,6 +459,9 @@ class DirichletBC(BCBase, DirichletBCMixin):
         # DirichletBC is directly used in assembly.
         return self
 
+    def _as_nonlinear_variational_problem_arg(self):
+        return self
+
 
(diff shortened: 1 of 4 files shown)


firedrake/preconditioners/assembled.py
```diff
@@ -44,7 +44,8 @@ def initialize(self, pc):
 
         (a, bcs) = self.form(pc, test, trial)
 
-        form_assembler = get_assembler(a, bcs=bcs, form_compiler_parameters=fcp, mat_type=mat_type, options_prefix=options_prefix)
+        bcs_Jp = tuple(bc.extract_form('Jp') for bc in bcs)
```
pbrubeck, line 47: This should go in `self.form` so every preconditioner gets the correct bcs

ksagiyam: Thanks.

PR #3857, opened 2024-11-10
ngsPETSc is switching to pip

UZerbinati:


connorjward (commented):
This looks great. Are there any version considerations we need to be aware of?

UZerbinati:
> This looks great. Are there any version considerations we need to be aware of?

connorjward:
> > This looks great. Are there any version considerations we need to be aware of?

Huh?

dham (approved):

PR #3875, opened 2024-11-20
Add functionality to print patch statistics in ASMStarPC and friends

pefarrell:
We often would like to know what kind of sizes of patch problems we are dealing with in ASMStarPC and friends. 

When using `-snes_view`, PETSc prints out the size of the first patch on the first processor, but this isn't all that useful; really we want the minimum and maximum.

This adds an option `-pc_star_print_patch_statistics` that computes the minimum and maximum patch sizes and prints them out in the viewer, if requested.


firedrake/preconditioners/asm.py
```diff
@@ -103,6 +103,30 @@ def get_patches(self, V):
...
+            print_statistics = opts.getBool("print_patch_statistics", default=False)
+
+            if print_statistics:
+                from mpi4py import MPI
+
+                dm = pc.getDM()
+                V = get_function_space(dm)
+
+                # Obtain patches from user defined function
+                ises = self.get_patches(V)
```
ScottMacLachlan, line 119: It feels slightly extravagant to completely recompute the patch indices just for the sake of this monitor, but it is true that we don't stash these anywhere convenient for us to find.  Is it worth stashing them in `__init__` so that we don't need to recompute them here?

ScottMacLachlan (approved):

firedrake/preconditioners/asm.py
pefarrell: I agree it's a bit wasteful. I agree we could stash them in `__init__` if the option is present. I'll make this change.

JHopeCollins:
This is a really useful option, thanks @pefarrell. Any chance you could add the mean patch size too? Could be useful to know if the min/max are particular outliers or not.

firedrake/preconditioners/asm.py
pbrubeck: Isn't the correct approach here to compute the statistics during `__init__` if the options are present? Then just stash the statistics to save memory.

tests/firedrake/regression/test_star_pc.py
```diff
@@ -50,6 +51,7 @@ def test_star_equivalence(problem_type, backend):
                        "mg_levels_ksp_max_it": 1,
                        "mg_levels_pc_type": "python",
                        "mg_levels_pc_python_type": "firedrake.ASMStarPC",
+                       "mg_levels_pc_star_print_patch_statistics": None,
```
pbrubeck, line 54: Should we rename the option to `pc_star_view_patch_sizes`?

dham (approved):

PR #3892, opened 2024-11-28
Enable vtk write cofunction

Ig-dolci:
# Description

### DIFF
--- a/firedrake/output/vtk_output.py
+++ b/firedrake/output/vtk_output.py
@@ -479,7 +479,8 @@ class VTKFile(object):
             raise ValueError("Unsupported shape %s" % (shape, ))
         if isinstance(function, Function):
             output = Function(V)
-        elif isinstance(function, Cofunction):
+        else:
+            assert isinstance(function, Cofunction)
             output = Function(V.dual())
 
         if self.project:
--- a/firedrake/output/vtk_output.py
+++ b/firedrake/output/vtk_output.py
@@ -503,7 +503,7 @@ class VTKFile(object):
 
         for f in functions:
             if not isinstance(f, (Function, Cofunction)):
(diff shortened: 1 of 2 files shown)


firedrake/output/vtk_output.py
```diff
@@ -496,8 +502,8 @@ def _write_vtu(self, *functions):
             functions = [Function(V)]
 
         for f in functions:
-            if not isinstance(f, Function):
-                raise ValueError("Can only output Functions or a single mesh, not %r" % type(f))
+            if not isinstance(f, (Function, Cofunction)):
+                raise ValueError("Can only output Functions, Cofunctions or a single mesh, not %r" % type(f))
```
connorjward, line 506:
```suggestion
                raise ValueError(f"Can only output Functions, Cofunctions or a single mesh, not {type(f).__name__}")
```

firedrake/output/vtk_output.py
```diff
@@ -477,16 +477,22 @@ def _prepare_output(self, function, max_elem):
                                     shape=shape)
         else:
             raise ValueError("Unsupported shape %s" % (shape, ))
-        output = Function(V)
+        if isinstance(function, Function):
+            output = Function(V)
+        elif isinstance(function, Cofunction):
```
connorjward, line 482:
```suggestion
        else:
            assert isinstance(function, Cofunction)
```
This is just my personal preference. Means that `function` can't be anything else.

connorjward (changes requested):
Otherwise happy.

firedrake/output/vtk_output.py
Ig-dolci: Okay. I am fine with `assert` here. We have a previous check with an error message raised.

PR #3862, opened 2024-11-12
Get value_shape from FunctionSpace

pbrubeck:
# Description

1. Replace `V.ufl_element().value_shape` by `V.value_shape`,
2. `firedrake.FunctionSpace.value_size` has been redefined to agree with `ufl.FunctionSpace.value_size = numpy.prod(value_shape, int)`, so the old behaviour gets renamed as `firedrake.FunctionSpace.block_size`

Depends on:
https://github.com/firedrakeproject/ufl/pull/52
https://github.com/FInAT/FInAT/pull/118
https://github.com/firedrakeproject/tsfc/pull/307

PR #3895, opened 2024-11-28
Function has fs.block_size components not fs.value_size

JHopeCollins:
In #3862 `FunctionSpace.block_size` took on the old behaviour of `FunctionSpace.value_size`.
Defining the components for a `FunctionSpace` was updated to use `block_size`, but the components definition for `Function` wasn't.


pbrubeck (approved):
Thanks for spotting this!

PR #3897, opened 2024-11-28
Fix the `FormSum` memory leak

Ig-dolci:
# Description

We are having memory leaks and more expensive computation of solvers involving `FormSum` (See more details on this [discussion](https://github.com/firedrakeproject/firedrake/discussions/3887)). That is caused by the [sum operation](https://github.com/firedrakeproject/firedrake/blob/master/firedrake/assemble.py#L472) involving `dat`. My solution here is to make a sum operation through the numpy arrays. Below is a comparison (using the example added at the [discussion](https://github.com/firedrakeproject/firedrake/discussions/3887)) of time and memory computation differences between the master branch and the current PR.

result

### DIFF
--- a/firedrake/assemble.py
+++ b/firedrake/assemble.py
@@ -5,6 +5,7 @@ import functools
 import itertools
 from itertools import product
 import numbers
+import numpy as np
 
 import cachetools
 import finat
@@ -378,7 +379,8 @@ class BaseFormAssembler(AbstractFormAssembler):
             return self.base_form_assembly_visitor(e, t, *operands)
 
         # DAG assembly: traverse the DAG in a post-order fashion and evaluate the node on the fly.
-        result = BaseFormAssembler.base_form_postorder_traversal(self._form, visitor)
+        visited = {}
+        result = BaseFormAssembler.base_form_postorder_traversal(self._form, visitor, visited)
 
         if tensor:
             BaseFormAssembler.update_tensor(result, tensor)
@@ -468,10 +470,13 @@ class BaseFormAssembler(AbstractFormAssembler):
                 return sum(weight * arg for weight, arg in zip(expr.weights(), args))
             elif all(isinstance(op, firedrake.Cofunction) for op in args):
                 V, = set(a.function_space() for a in args)
-                # res = sum([w*op.dat for (op, w) in zip(args, expr.weights())])
-                # return firedrake.Cofunction(V, res)
                 result = firedrake.Cofunction(V)
-                result.dat.data[...] = sum(w * op.dat.data[...] for op, w in zip(args, expr.weights()))
+                for op, w in zip(args, expr.weights()):
+                    for dat_result, dat_op in zip(result.dat.split, op.dat.split):
+                        np.add(
+                            dat_result.data_ro_with_halos,
+                            w * dat_op.data_ro_with_halos,
+                            out=dat_result.data_wo_with_halos)
                 return result
             elif all(isinstance(op, ufl.Matrix) for op in args):
                 res = tensor.petscmat if tensor else PETSc.Mat()
@@ -586,8 +591,7 @@ class BaseFormAssembler(AbstractFormAssembler):
             raise NotImplementedError("Cannot update tensor of type %s" % type(tensor))
 
     @staticmethod
-    def base_form_postorder_traversal(expr, visitor, visited=None):
-        visited = visited if visited is not None else {}
+    def base_form_postorder_traversal(expr, visitor, visited={}):
         if expr in visited:
             return visited[expr]
 
@@ -609,8 +613,7 @@ class BaseFormAssembler(AbstractFormAssembler):
         return visited[expr]
 
     @staticmethod
-    def base_form_preorder_traversal(expr, visitor, visited=None):
-        visited = visited if visited is not None else {}
+    def base_form_preorder_traversal(expr, visitor, visited={}):
         if expr in visited:
             return visited[expr]
 
--- a/firedrake/assemble.py
+++ b/firedrake/assemble.py
@@ -378,8 +378,7 @@ class BaseFormAssembler(AbstractFormAssembler):
             return self.base_form_assembly_visitor(e, t, *operands)
 
         # DAG assembly: traverse the DAG in a post-order fashion and evaluate the node on the fly.
-        visited = {}
-        result = BaseFormAssembler.base_form_postorder_traversal(self._form, visitor, visited=visited)
+        result = BaseFormAssembler.base_form_postorder_traversal(self._form, visitor)
 
         if tensor:
             BaseFormAssembler.update_tensor(result, tensor)
@@ -469,8 +468,11 @@ class BaseFormAssembler(AbstractFormAssembler):
                 return sum(weight * arg for weight, arg in zip(expr.weights(), args))
             elif all(isinstance(op, firedrake.Cofunction) for op in args):
                 V, = set(a.function_space() for a in args)
-                res = sum([w*op.dat for (op, w) in zip(args, expr.weights())])
-                return firedrake.Cofunction(V, res)
+                # res = sum([w*op.dat for (op, w) in zip(args, expr.weights())])
+                # return firedrake.Cofunction(V, res)
+                result = firedrake.Cofunction(V)
+                result.dat.data[...] = sum(w * op.dat.data[...] for op, w in zip(args, expr.weights()))
+                return result
             elif all(isinstance(op, ufl.Matrix) for op in args):
                 res = tensor.petscmat if tensor else PETSc.Mat()
                 is_set = False
@@ -585,35 +587,26 @@ class BaseFormAssembler(AbstractFormAssembler):
 
     @staticmethod
     def base_form_postorder_traversal(expr, visitor, visited=None):
-        visited = visited or {}
+        visited = visited if visited is not None else {}
         if expr in visited:
             return visited[expr]
 
         stack = [expr]
-        processing = set()
-
         while stack:
-            e = stack[-1]
-            if e in visited:
-                stack.pop()
-                continue
-
+            e = stack.pop()
+            unvisited_children = []
             operands = BaseFormAssembler.base_form_operands(e)
-            unvisited_children = [arg for arg in operands if arg not in visited]
+            for arg in operands:
+                if arg not in visited:
+                    unvisited_children.append(arg)
 
             if unvisited_children:
+                stack.append(e)
                 stack.extend(unvisited_children)
-                processing.update(unvisited_children)
             else:
-                stack.pop()
-                # if not isinstance(e, firedrake.Cofunction):
                 visited[e] = visitor(e, *(visited[arg] for arg in operands))
-                # else:
-                #     visited[e] = e
-                processing.discard(e)
 
         return visited[expr]
-
 
     @staticmethod
     def base_form_preorder_traversal(expr, visitor, visited=None):
@@ -668,13 +661,9 @@ class BaseFormAssembler(AbstractFormAssembler):
             # Need to reconstruct the expression with its visited operands!
             expr = BaseFormAssembler.reconstruct_node_from_operands(expr, operands)
             # Perform the DAG restructuring when needed
-            if visited:
-                return BaseFormAssembler.restructure_base_form(expr, visited=visited)
-            return BaseFormAssembler.restructure_base_form(expr)
-        if visited:
-            return BaseFormAssembler.base_form_postorder_traversal(
-                expression, visitor, visited=visited)
-        return BaseFormAssembler.base_form_postorder_traversal(expression, visitor)
+            return BaseFormAssembler.restructure_base_form(expr, visited)
+
+        return BaseFormAssembler.base_form_postorder_traversal(expression, visitor, visited)
 
     @staticmethod
     def restructure_base_form_preorder(expression, visited=None):
@@ -682,13 +671,9 @@ class BaseFormAssembler(AbstractFormAssembler):
 
         def visitor(expr):
             # Perform the DAG restructuring when needed
-            if visited:
-                return BaseFormAssembler.restructure_base_form(expr, visited=visited)
-            return BaseFormAssembler.restructure_base_form(expr)
-        if visited:
-            return BaseFormAssembler.base_form_preorder_traversal(expression, visitor, visited=visited)
-        else:
-            return BaseFormAssembler.base_form_preorder_traversal(expression, visitor)
+            return BaseFormAssembler.restructure_base_form(expr, visited)
+
+        expression = BaseFormAssembler.base_form_preorder_traversal(expression, visitor, visited)
         # Need to reconstruct the expression at the end when all its operands have been visited!
         operands = [visited.get(args, args) for args in BaseFormAssembler.base_form_operands(expression)]
         return BaseFormAssembler.reconstruct_node_from_operands(expression, operands)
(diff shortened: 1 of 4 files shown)


JHopeCollins:
Oh dear, if this fixes the issue then that interface could do with updating!

I think that the traversal method is taking advantage of the behaviour described here: https://github.com/firedrakeproject/firedrake/pull/3348#discussion_r1600292194

This behaviour is not widely known and is quite unintuitive - it's usually considered a bug. It might be best to change this method signature (and the preorder traversal method) to use `visited=None` as the kwarg, and use a `self._visited` dict with `visited = visited if visited else self._visited`

Ig-dolci:
> Oh dear, if this fixes the issue then that interface could do with updating!
> 
> I think that the traversal method is taking advantage of the behaviour described here: [#3348 (comment)](https://github.com/firedrakeproject/firedrake/pull/3348#discussion_r1600292194)
> 
> This behaviour is not widely known and is quite unintuitive - it's usually considered a bug. It might be best to change this method signature (and the preorder traversal method) to use `visited=None` as the kwarg, and use a `self._visited` dict with `visited = visited if visited else self._visited`

No. This does not fix it. I am still debugging.

JHopeCollins:
> No. This does not fix it. I am still debugging.

Shame it wasn't so simple! It may still be good to change the signatures to avoid that behaviour though

Ig-dolci:
> Oh dear, if this fixes the issue then that interface could do with updating!
> 
> I think that the traversal method is taking advantage of the behaviour described here: [#3348 (comment)](https://github.com/firedrakeproject/firedrake/pull/3348#discussion_r1600292194)
> 
> This behaviour is not widely known and is quite unintuitive - it's usually considered a bug. It might be best to change this method signature (and the preorder traversal method) to use `visited=None` as the kwarg, and use a `self._visited` dict with `visited = visited if visited else self._visited`

I see. I gonna try `visited=None`.

firedrake/assemble.py
```diff
@@ -584,7 +583,8 @@ def update_tensor(assembled_base_form, tensor):
             raise NotImplementedError("Cannot update tensor of type %s" % type(tensor))
 
     @staticmethod
-    def base_form_postorder_traversal(expr, visitor, visited={}):
+    def base_form_postorder_traversal(expr, visitor, visited=None):
+        visited = visited if visited is not None else {}
```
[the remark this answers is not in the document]
Ig-dolci: I decided to keep the original code here.

firedrake/assemble.py
```diff
@@ -469,8 +470,14 @@ def base_form_assembly_visitor(self, expr, tensor, *args):
...
                 V, = set(a.function_space() for a in args)
-                res = sum([w*op.dat for (op, w) in zip(args, expr.weights())])
-                return firedrake.Cofunction(V, res)
+                result = firedrake.Cofunction(V)
+                for op, w in zip(args, expr.weights()):
+                    for dat_result, dat_op in zip(result.dat.split, op.dat.split):
+                        np.add(
+                            dat_result.data_ro_with_halos,
+                            w * dat_op.data_ro_with_halos,
+                            out=dat_result.data_wo_with_halos)
```
Ig-dolci, line 479: I am not sure if make this operation `with_halos` is the right think to do.

Ig-dolci: No. `result` is a `Cofunction`, and `Cofunction` assigning an expression will reach this code again, which leads to a maximum recursion.

dham: This is the right implementation but we should move it to a method `Dat.maxpy(weights, dats)`

pyop2/types/dat.py
```diff
@@ -492,6 +492,20 @@ def norm(self):
         from math import sqrt
         return sqrt(self.inner(self).real)
 
+    def axpy(self, alpha: float, other: 'Dat') -> None:
```
JHopeCollins, line 495: Much better having it hidden away in pyop2!

Please could you also add the `maxpy` (_multiple_ axpy) routine that we discussed in the meeting and use this in `assemble`?
`maxpy` takes in a sequence of alphas and a sequence of xs and increments them all into y.
For now it could just be implemented as a loop over axpy for each (alpha,x) pair, but in the future it could allow better control over data movement to/from GPUs.

pyop2/types/dat.py
```diff
@@ -492,6 +492,20 @@ def norm(self):
...
+        """Compute the operation :math:`y = \\alpha x + y`.
+
+        :arg alpha: a scalar
+        :arg other: the :class:`Dat` to add to this one
```
JHopeCollins, line 499: Usually this argument would be called `x` to be clear what role it has in the y=a*x+y procedure. If `other` is the standard argument name for `Dat` methods then the fact that `other` is x and `self` is y should be stated in the docstring.

pyop2/types/dat.py
```diff
@@ -1022,6 +1036,21 @@ def inner(self, other):
             ret += s.inner(o)
         return ret
 
+    def axpy(self, alpha: float, other: 'MixedDat') -> None:
```
JHopeCollins, line 1039: as above for `maxpy`

pyop2/types/dat.py
```diff
@@ -1022,6 +1036,21 @@ def inner(self, other):
...
+        """Compute the operation :math:`y = \\alpha x + y`.
+
+        :arg alpha: a scalar
+        :arg other: the :class:`Dat` to add to this one
```
JHopeCollins, line 1043: as above for `other` argument name

pyop2/types/glob.py
```diff
@@ -203,6 +203,14 @@ def inner(self, other):
         assert issubclass(type(other), type(self))
         return np.dot(self.data_ro, np.conj(other.data_ro))
 
+    def axpy(self, alpha, other):
```
JHopeCollins, line 206: as above for `maxpy`

pyop2/types/glob.py
```diff
@@ -203,6 +203,14 @@ def inner(self, other):
...
+        """Compute the operation :math:`y = \\alpha x + y`.
+        """
```
JHopeCollins, line 208: docstring needs the parameter list (and see comments above about argument naming)

JHopeCollins (changes requested):
Looks good, just a couple of small suggestions on the code.

It would be good to have a test for the axpy methods too!

pyop2/types/dat.py
```diff
@@ -492,6 +492,20 @@ def norm(self):
...
         return sqrt(self.inner(self).real)
 
+    def axpy(self, alpha: float, other: 'Dat') -> None:
+        """Compute the operation :math:`y = \\alpha x + y`.
+
+        :arg alpha: a scalar
+        :arg other: the :class:`Dat` to add to this one
+
+        """
+        if isinstance(other._data, np.ndarray):
```
connorjward, line 502: I think we should include a check to make sure that the `Dats` are compatible. E.g.:
```py
if self.dataset != other.dataset:
    raise SomeException("NO!!!")
```

pyop2/types/dat.py
```diff
@@ -1022,6 +1036,21 @@ def inner(self, other):
...
         return ret
 
+    def axpy(self, alpha: float, other: 'MixedDat') -> None:
+        """Compute the operation :math:`y = \\alpha x + y`.
+
+        :arg alpha: a scalar
+        :arg other: the :class:`Dat` to add to this one
+
+        """
+        for dat_result, dat_other in zip(self, other):
```
connorjward, line 1046: Similarly would be good to have a check.

connorjward (changes requested):
Thanks! I would also like to see some tests. Including for the failure conditions like mismatching dats. These tests could go in https://github.com/firedrakeproject/firedrake/blob/master/tests/pyop2/test_dats.py (which contains some examples you could work off)

pyop2/types/dat.py
Ig-dolci: I like the idea of having a `maxpy` like this:
```
def maxpy(self, scalar: tuple, other: tuple) -> None:
        """Compute a sequence of axpy operations.

        This is equivalent to calling :meth:`axpy` for each pair of
        scalars and :class:`Dat` in the input sequences.

        :arg scalar: a sequence of scalars
        :arg other: a sequence of :class:`Dat`\s

        See also :meth:`axpy`.
    
        """
        for alpha, dat in zip(scalar, other):
            self.axpy(alpha, dat)
```
Any strong opinion against it?

connorjward: Personally I think I prefer to, instead of passing
```
dat.maxpy([2.9, 3.3], [datA, datB])
```
to have
```
dat.maxpy([(2.9, datA), (3.3, datB)])
```

JHopeCollins: @Ig-dolci I think for now, that implementation is fine, it's what we are doing anyway, and encapsulating it means we can optimise later if it shows up on profiles.

@connorjward I'm not sure I have a preference either way. The current signature is closer to PETSc Vecs https://petsc.org/release/manualpages/Vec/VecMAXPY/ but your suggestion is probably more "pythonic". Although you wouldn't need to pass them as a list, it could just be `dat.maxpy((a0, x0), (a1, x1), ...)` and splat it into `*args`.

connorjward: Indeed your suggestion is the most "pythonic" of all!

In general I think it is not a great idea to try and mimic a C API in Python. This is something that really winds me up about petsc4py.

JHopeCollins: Ha, fair enough. Pairs of (alpha, x) probably best.

pbrubeck: I think most use cases would require to unpack and repack the two lists of scalars and dats into the list of tuples. Are we sure about this?

connorjward: I guess the difference in the actual example we use above would be:
```py
result.dat.maxpy(expr.weights(), [a.dat for a in args])
```
vs
```py
result.dat.maxpy(*(zip(expr.weights(), [a.dat for a in args])))
```
so maybe you're right

pyop2/types/dat.py
```diff
@@ -492,14 +492,33 @@ def norm(self):
...
+        if len(scalar) != len(x):
+            raise ValueError("scalar and x must have the same length")
+        for alpha_i, x_i in zip(scalar, x):
+            self.axpy(alpha_i, x_i)
+
     def axpy(self, alpha: float, other: 'Dat') -> None:
         """Compute the operation :math:`y = \\alpha x + y`.
 
-        :arg alpha: a scalar
-        :arg other: the :class:`Dat` to add to this one
```
Ig-dolci: I removed the list of args. I think it is redundant here.

pyop2/types/dat.py
Ig-dolci: Thinking in the most used cases of two lists of scalars and dats. I would prefer `maxpy(scalar, x)`.

pyop2/types/dat.py
```diff
@@ -492,6 +492,39 @@ def norm(self):
...
         return sqrt(self.inner(self).real)
 
+    def maxpy(self, scalar: list, x: list) -> None:
+        """Compute a sequence of axpy operations.
+
+        This is equivalent to calling :meth:`axpy` for each pair of
+        scalars and :class:`Dat` in the input sequences.
+
+        :arg scalar: A sequence of scalars.
+        :arg x: A sequence of :class:`Dat`.
```
connorjward, line 502:
```suggestion
        Parameters
        ----------
        scalar :
            A sequence of scalars.
        x :
            A sequence of :class:`Dat`.
```

pyop2/types/dat.py
```diff
@@ -492,6 +492,39 @@ def norm(self):
         from math import sqrt
         return sqrt(self.inner(self).real)
 
+    def maxpy(self, scalar: list, x: list) -> None:
```
connorjward, line 495:
```suggestion
    def maxpy(self, scalar: Sequence, x: Sequence) -> None:
```
this probably entails a `from collections.abc import Sequence` at the top of the file.

pyop2/types/dat.py
```diff
@@ -492,6 +492,39 @@ def norm(self):
...
+        """Compute a sequence of axpy operations.
+
+        This is equivalent to calling :meth:`axpy` for each pair of
+        scalars and :class:`Dat` in the input sequences.
+
+        :arg scalar: A sequence of scalars.
+        :arg x: A sequence of :class:`Dat`.
+
+        See also :meth:`axpy`.
+
```
connorjward, line 505:
```suggestion
```
probably not needed

pyop2/types/dat.py
```diff
@@ -492,6 +492,39 @@ def norm(self):
...
+        """
+        if len(scalar) != len(x):
+            raise ValueError("scalar and x must have the same length")
+        for alpha_i, x_i in zip(scalar, x):
+            self.axpy(alpha_i, x_i)
+
+    def axpy(self, alpha: float, other: 'Dat') -> None:
+        """Compute the operation :math:`y = \\alpha x + y`.
+
+        On this case, `self` is `y` and `other` is `x`.
```
connorjward, line 515:
```suggestion
        In this case, ``self`` is ``y`` and ``other`` is ``x``.
```

pyop2/types/glob.py
```diff
@@ -203,6 +203,36 @@ def inner(self, other):
         assert issubclass(type(other), type(self))
         return np.dot(self.data_ro, np.conj(other.data_ro))
 
+    def maxpy(self, scalar: list, x: list) -> None:
```
connorjward, line 206:
```suggestion
    def maxpy(self, scalar: Sequence, x: Sequence) -> None:
```

pyop2/types/glob.py
```diff
@@ -203,6 +203,36 @@ def inner(self, other):
...
+        """Compute a sequence of axpy operations.
+
+        This is equivalent to calling :meth:`axpy` for each pair of
+        scalars and :class:`Dat` in the input sequences.
+
+        :arg scalar: A sequence of scalars.
+        :arg x: A sequence of :class:`Dat`.
```
connorjward, line 213:
```suggestion
        Parameters
        ----------
        scalar :
            A sequence of scalars.
        x :
            A sequence of `Global`.
```

pyop2/types/glob.py
```diff
@@ -203,6 +203,36 @@ def inner(self, other):
...
+
+        See also :meth:`axpy`.
+
```
connorjward, line 216:
```suggestion
```

pyop2/types/glob.py
```diff
@@ -203,6 +203,36 @@ def inner(self, other):
...
+        """
+        if len(scalar) != len(x):
+            raise ValueError("scalar and x must have the same length")
+        for alpha_i, x_i in zip(scalar, x):
+            self.axpy(alpha_i, x_i)
+
+    def axpy(self, alpha: float, other: 'Global') -> None:
+        """Compute the operation :math:`y = \\alpha x + y`.
+
+        On this case, `self` is `y` and `other` is `x`.
```
connorjward, line 226:
```suggestion
        In this case, ``self`` is ``y`` and ``other`` is ``x``.
```

connorjward (changes requested):
Just docstring nitpicks from me. Looks fine otherwise. Thanks!

pyop2/types/dat.py
```diff
@@ -1022,6 +1058,23 @@ def inner(self, other):
             ret += s.inner(o)
         return ret
 
+    def axpy(self, alpha: float, other: 'MixedDat') -> None:
+        """Compute the operation :math:`y = \\alpha x + y`.
+
+        On this case, `self` is `y` and `other` is `x`.
```
Ig-dolci, line 1064:
```suggestion
        In this case, ``self`` is ``y`` and ``other`` is ``x``.
```

PR #3898, opened 2024-11-28
Tackle many warnings

### LINKED ISSUE #3871: BUG: `Cofunction += Cofunction` leads to `DeprecationWarning`s
**Describe the bug**
```
import warnings

warnings.simplefilter("default")

import firedrake as fd

mesh = fd.UnitIntervalMesh(1)
space = fd.FunctionSpace(mesh, "DG", 0)

u = fd.Cofunction(space.dual())
u += fd.Cofunction(space.dual())
```
leads to the warnings
```
/home/jmaddis2/build/firedrake/firedrake/src/PyOP2/pyop2/types/dat.py:408: DeprecationWarning: Expression.index(i) is deprecated and will be removed in 2H2025. Use 'expr[i] 'instead.
  lhs = _self.index(i)
/home/jmaddis2/build/firedrake/firedrake/src/PyOP2/pyop2/types/dat.py:416: DeprecationWarning: Expression.index(i) is deprecated and will be removed in 2H2025. Use 'expr[i] 'instead.
  rhs = _other.index(i)
```

**Expected behavior**
No warning.

**Environment:**
Ubuntu 24.04, new Firedrake build.

connorjward:
## TODOs
- [x] Use actual upstream loopy
- [x] Check that this isn't a major performance regression (!?) (might just be overloaded runners)

Closes #3871

(Isn't it nice that I don't need a PyOP2 PR!)

### DIFF
--- a/.github/workflows/pyop2.yml
+++ b/.github/workflows/pyop2.yml
@@ -107,6 +107,7 @@ jobs:
         run: |
           source ../venv/bin/activate
           python -m pip install -v ".[test]"
+          python -m pip uninstall -y pytools loopy
           python -m pip install git+https://github.com/firedrakeproject/loopy.git@connorjward/revert-hack-for-firedrake
 
       - name: Run tests
--- a/.github/workflows/pyop2.yml
+++ b/.github/workflows/pyop2.yml
@@ -107,6 +107,7 @@ jobs:
         run: |
           source ../venv/bin/activate
           python -m pip install -v ".[test]"
+          python -m pip install git+https://github.com/firedrakeproject/loopy.git@connorjward/revert-hack-for-firedrake
 
       - name: Run tests
         shell: bash
--- a/.github/workflows/pyop2.yml
+++ b/.github/workflows/pyop2.yml
@@ -107,8 +107,6 @@ jobs:
         run: |
           source ../venv/bin/activate
           python -m pip install -v ".[test]"
-          python -m pip uninstall -y pytools loopy
-          python -m pip install git+https://github.com/firedrakeproject/loopy.git@connorjward/merge-upstream
 
       - name: Run tests
         shell: bash
--- a/.github/workflows/pyop2.yml
+++ b/.github/workflows/pyop2.yml
@@ -108,7 +108,7 @@ jobs:
           source ../venv/bin/activate
           python -m pip install -v ".[test]"
           python -m pip uninstall -y pytools loopy
-          python -m pip install git+https://github.com/firedrakeproject/loopy.git@connorjward/revert-hack-for-firedrake
+          python -m pip install git+https://github.com/firedrakeproject/loopy.git@connorjward/merge-upstream
 
       - name: Run tests
         shell: bash
--- a/pyop2/codegen/rep2loopy.py
+++ b/pyop2/codegen/rep2loopy.py
@@ -546,6 +546,14 @@ def generate(builder, wrapper_name=None):
         assumptions = assumptions & pwaffd[parameters.layer_start].le_set(pwaffd[parameters.layer_end])
     assumptions = reduce(operator.and_, assumptions.get_basic_sets())
 
+    # Even though we think we are setting loop priorities correctly here, loopy cannot
+    # use its new scheduler and raises a warning stating that there are
+    # "loop priority dependencies between sibling loop nests".
+    # See https://github.com/inducer/loopy/issues/890.
+    # Therefore until loopy fixes this we disable this warning and use the old (slower)
+    # scheduling algorithm.
+    silenced_warnings = ["v1_scheduler_fallback"]
+
     wrapper = loopy.make_kernel(domains,
                                 statements,
                                 kernel_data=parameters.kernel_data,
@@ -556,7 +564,8 @@ def generate(builder, wrapper_name=None):
                                 assumptions=assumptions,
                                 lang_version=(2018, 2),
                                 name=wrapper_name,
-                                loop_priority=loop_priorities)
+                                loop_priority=loop_priorities,
+                                silenced_warnings=silenced_warnings)
 
     # register kernel
     kernel = builder.kernel
--- a/pyop2/codegen/rep2loopy.py
+++ b/pyop2/codegen/rep2loopy.py
@@ -415,7 +415,7 @@ def generate(builder, wrapper_name=None):
     # use a dict as an ordered set
     outer_inames = {builder._loop_index.name: None}
     if builder.layer_index is not None:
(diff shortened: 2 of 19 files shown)


.github/workflows/pyop2.yml
```diff
@@ -107,6 +107,8 @@ jobs:
         run: |
           source ../venv/bin/activate
           python -m pip install -v ".[test]"
+          python -m pip uninstall -y pytools loopy
+          python -m pip install git+https://github.com/firedrakeproject/loopy.git@connorjward/merge-upstream
```
connorjward, line 111:
```suggestion
```

pyop2/codegen/rep2loopy.py
```diff
@@ -205,16 +208,18 @@ def __call__(self, preamble_info):
...
-    fields = set(["name", "parameters", "arg_id_to_dtype", "arg_id_to_descr", "name_in_target"])
     init_arg_names = ("name", "parameters", "arg_id_to_dtype", "arg_id_to_descr", "name_in_target")
 
+    parameters: tuple
+
     def __init__(self, name, parameters, arg_id_to_dtype=None, arg_id_to_descr=None, name_in_target=None):
-        super(PyOP2KernelCallable, self).__init__(name, arg_id_to_dtype, arg_id_to_descr, name_in_target)
-        self.parameters = parameters
+        super().__init__(name, arg_id_to_dtype, arg_id_to_descr, name_in_target)
+        object.__setattr__(self, "parameters", tuple(parameters))
```
JHopeCollins, line 222: Why `object.__setattr__` and not `self.parameters = ...` or `setattr`?

pyop2/codegen/rep2loopy.py
```diff
@@ -288,8 +293,8 @@ def runtime_indices(expressions):
     for node in traversal(expressions):
         if isinstance(node, RuntimeIndex):
             indices.append(node.name)
-
-    return frozenset(indices)
+    # use a dict as an ordered set
+    return {i: None for i in indices}
```
JHopeCollins, line 297: This ensures ordering but also changes the return type.
Why not:
```python
return {i: None for i in indices}.keys()
```
Or
```python
return type(indices)(dict.fromkeys(indices))
```

pyop2/codegen/rep2loopy.py
```diff
@@ -407,11 +412,10 @@ def generate(builder, wrapper_name=None):
...
     RuntimeIndex._count = itertools.count()
 
+    # use a dict as an ordered set
+    outer_inames = {builder._loop_index.name: None}
     if builder.layer_index is not None:
-        outer_inames = frozenset([builder._loop_index.name,
-                                  builder.layer_index.name])
-    else:
-        outer_inames = frozenset([builder._loop_index.name])
+        outer_inames |= {builder.layer_index.name: None}
```
JHopeCollins, line 418: Personally I find `dict.update` clearer than `|=`.

pyop2/codegen/rep2loopy.py
```diff
@@ -535,6 +546,14 @@ def renamer(expr):
         assumptions = assumptions & pwaffd[parameters.layer_start].le_set(pwaffd[parameters.layer_end])
     assumptions = reduce(operator.and_, assumptions.get_basic_sets())
 
+    # Even though we think we are setting loop priorities correctly here, loopy cannot
+    # use its new scheduler and raises a warning stating that there are
+    # "loop priority dependencies between sibling loop nests".
+    # See https://github.com/inducer/loopy/issues/890.
+    # Therefore until loopy fixes this we disable this warning and use the old (slower)
+    # scheduling algorithm.
```
JHopeCollins, line 554: Please open a Firedrake issue linking to these lines and the loopy issue to we can keep track and hopefully revery once loopy gets fixed.

pyop2/codegen/rep2loopy.py
connorjward: Because this is a frozen dataclass. I was copying what loopy do internally ([example](https://github.com/inducer/loopy/blob/76f4afc27f343bc7e95a11b9b04d30c574402bcd/loopy/symbolic.py#L646), I really don't like it).

JHopeCollins: Does it break otherwise? Is there something specific about frozen dataclasses that means you have to do this?
I would usually advocate for using the smallest hammer possible!

connorjward: Yeah otherwise it breaks unfortunately

JHopeCollins: `setattr` breaks or `self.parameters = `? What's the error?

connorjward: My guess is that dataclasses overload `__setattr__` and so we need to specify `object.__setattr__`. Will report back.

pyop2/codegen/rep2loopy.py
connorjward: I need to return a type that can act as an ordered set. The return type needs to be changed because I continue to need ordered set semantics outside of the function.

pyop2/codegen/rep2loopy.py
connorjward: Won't be long! https://github.com/inducer/loopy/pull/895
I might just wait for that to be merged.

pyop2/codegen/rep2loopy.py
JHopeCollins: Ok, that's fine then.

pyop2/codegen/rep2loopy.py
inducer: That's exactly right. There's no *real* immutability in Python, and so the MO of frozen dataclasses is to make mutation stick out like a sore thumb. Reading your comments above, I would say mission accomplished. :slightly_smiling_face: 

The conventional, easier-on-the-eyes approach would be to just delete the constructor (and let the `@dataclass` transform worry about writing they eyesore code). This is not feasible here because the argument order would mismatch what dataclasses generates, breaking compatibility. :cry:

JHopeCollins: I see. Not the nicest thing but doesn't sound like there's much alternative if this is a dataclass but the `PyOP2KernelCallable` has a different signature to `PyOP2KernelCallable`.

pyop2/codegen/rep2loopy.py
connorjward: Fixed upstream. Will remove this.

JHopeCollins (approved):
LGTM, once the workflow branches are removed.
Thanks @connorjward!

PR #3911, opened 2024-12-08
Enable measure opts for Riesz representation L2 and H1

Ig-dolci:
# Description
There are cases where using measure options to Riesz representation, e.g., `dx(scheme=...)`, is required.


dham:
already possible via solver_options.

PR #3921, opened 2024-12-11
Cache basis evaluation

pbrubeck:
# Description

Cache FInAT `QuadratureRules` and `basis_tabulation` so the symbolic nodes with the same tabulations get effectively reused.
This should optimize code generation, compile time, and runtime for bilinear forms where the test and trial spaces are the same.


connorjward (approved):
Looks fine.

connorjward (approved):
LGTM

Issue #3203, opened 2023-11-03
BUG: `LinearSolver.solve` fails when passed two `Functions`.


APaganini:
**Describe the bug**
`LinearSolver.solve` fails when passed two `Functions`. The issue may be that [here](https://github.com/firedrakeproject/firedrake/blob/c568bfc4f932e72a434cebc74eacb982e02cfd86/firedrake/linear_solver.py#L133) `b` could be a function (which is allowed in solve, see [here](https://github.com/firedrakeproject/firedrake/blob/c568bfc4f932e72a434cebc74eacb982e02cfd86/firedrake/linear_solver.py#L147)), but `blift` returned by [_rhs](url) (see [here](https://github.com/firedrakeproject/firedrake/blob/c568bfc4f932e72a434cebc74eacb982e02cfd86/firedrake/linear_solver.py#L118)) is always a cofunction.

**Steps to Reproduce**
```
from firedrake import *

mesh = UnitSquareMesh(3,3)
V = FunctionSpace(mesh, "CG", 1)
u = TrialFunction(V)
v = TestFunction(V)
a = dot(grad(u),grad(v))*dx
bcs = DirichletBC(V, 0, "on_boundary")
A = assemble(a, mat_type='aij', bcs=bcs)
L = LinearSolver(A)

uout = Function(V)
uin = Function(V)
L.solve(uout, uin)
```

**Error message**
```
Traceback (most recent call last):
 File "/Users/alberto/Documents/FIREDRAKE/fireshape/tests/secondminfail.py", line 14, in <module>
  L.solve(uout, uin)
 File "petsc4py/PETSc/Log.pyx", line 188, in petsc4py.PETSc.Log.EventDecorator.decorator.wrapped_func
 File "petsc4py/PETSc/Log.pyx", line 189, in petsc4py.PETSc.Log.EventDecorator.decorator.wrapped_func
 File "/Users/alberto/Documents/FIREDRAKE/firedrake/src/firedrake/firedrake/linear_solver.py", line 163, in solve
  b = self._lifted(b)
    ^^^^^^^^^^^^^^^
 File "/Users/alberto/Documents/FIREDRAKE/firedrake/src/firedrake/firedrake/linear_solver.py", line 133, in _lifted
  blift += b
TypeError: unsupported operand type(s) for +=: 'Cofunction' and 'Function'
```

**Environment:**
MacOS,
Python 3.11.6

Firedrake Configuration:
```
    package_manager: True
[...]

PR #3214, opened 2023-11-09
Fix function spaces check in linear solver

nbouziani:
# Description

This PR: 1) updates the `LinearSolver`'s check on the solution and the rhs, and 2) fixes `__eq__` for function spaces. This fixes the issues #3203 and #3130.

Issue #3130, opened 2023-09-27
BUG: Spaces compare as equal with their dual


jrmaddison:
**Describe the bug**
Dual spaces compare as equal with associated primal spaces.

**Steps to Reproduce**
```
from firedrake import *

mesh = UnitIntervalMesh(10)
V = FunctionSpace(mesh, "Lagrange", 1)
V_star = V.dual()

assert bool(V == V_star)  # Passes
```

**Expected behavior**
Since the spaces differ, they should not compare as equal.

**Error message**
No error.

**Environment:**
Ubuntu 22.04

nbouziani:
I have just fixed that in #3214

jrmaddison:
Reopening for now as the fix is not yet merged.

jrmaddison:
@pbrubeck Fixed?

pbrubeck:
Fixed in https://github.com/firedrakeproject/firedrake/pull/3947

PR #3922, opened 2024-12-11
Restricted Cofunction RHS

pbrubeck:
# Description
This PR enables `solve(a == assemble(L), u, bcs, restrict={False|True})` without requiring users to set the boundary nodes to zero when they provide a Cofunction RHS. Here we also solved a few other issues:

1. V and V.dual() do no longer compare to the same thing.
2. `assemble(form, bcs, zero_bc_nodes=True)` is the new default, and for the False case we throw an error because we will try to apply primal inhomogenous data on a dual Cofunction.

Depends on https://github.com/firedrakeproject/ufl/pull/53

Fixes https://github.com/firedrakeproject/firedrake/issues/3203, https://github.com/firedrakeproject/firedrake/issues/3498, and https://github.com/firedrakeproject/firedrake/issues/3130

Supersedes https://github.com/firedrakeproject/firedrake/pull/3214, https://github.com/firedrakeproject/firedrake/pull/3754


firedrake/variational_solver.py
```diff
@@ -92,7 +92,11 @@ def __init__(self, F, u, bcs=None, J=None,
             self.u_restrict = Function(V_res).interpolate(u)
             v_res, u_res = TestFunction(V_res), TrialFunction(V_res)
             F_arg, = F.arguments()
-            self.F = replace(F, {F_arg: v_res, self.u: self.u_restrict})
+            replace_F = {F_arg: v_res, self.u: self.u_restrict}
+            for c in F.coefficients():
+                if c.function_space() == V.dual():
+                    replace_F[c] = Cofunction(V_res.dual()).interpolate(c)
```
ksagiyam, line 98: I understand why you do this, but I would like to think a bit more to see if there is any better solution.

ksagiyam (changes requested):

firedrake/variational_solver.py
ksagiyam: I hoped something like `self.F = interpolate(F, V_res.dual())` would work, but it looks not supported, so maybe this is fine.

.github/workflows/build.yml
```diff
@@ -83,6 +83,7 @@ jobs:
             --install defcon \
             --install gadopt \
             --install asQ \
+            --package-branch ufl pbrubeck/merge-upstream \
```
ksagiyam, line 86: Do you need upstream for this PR?

ksagiyam (changes requested):
Please fix CI.

.github/workflows/build.yml
pbrubeck: Yes. It includes a fix for `action`

firedrake/assemble.py
```diff
@@ -1185,23 +1184,21 @@ def _apply_bc(self, tensor, bc):
...
-            tensor_func = tensor.riesz_representation(riesz_map="l2")
-            if self._diagonal:
-                bc.set(tensor_func, 1)
-            else:
-                bc.apply(tensor_func)
-            tensor.assign(tensor_func.riesz_representation(riesz_map="l2"))
+        if self._diagonal:
+            bc.set(tensor, self._weight)
+        elif not self._zero_bc_nodes:
+            # We cannot set primal data on a dual Cofunction, this will throw an error
```
connorjward, line 1195: I don't understand this comment. Will the line below crash?

firedrake/cofunction.py
```diff
@@ -225,8 +225,12 @@ def assign(self, expr, subset=None, expr_from_assemble=False):
             return self.assign(
                 assembled_expr, subset=subset,
                 expr_from_assemble=True)
-
-        raise ValueError('Cannot assign %s' % expr)
+        elif expr == 0:
```
connorjward, line 228: We do the same for `ufl.classes.Zero` on line 199 so this should probably be combined. In general I am a bit concerned about this code because we do not tape `Assigner`. @Ig-dolci knows more I believe.

firedrake/linear_solver.py
```diff
@@ -147,6 +147,11 @@ def solve(self, x, b):
         if not isinstance(b, (function.Function, cofunction.Cofunction)):
             raise TypeError("Provided RHS is a '%s', not a Function or Cofunction" % type(b).__name__)
 
+        if x.function_space() != self.trial_space or b.function_space() != self.test_space.dual():
+            # When solving `Ax = b`, with A: V x U -> R, or equivalently A: V -> U*,
+            # we need to make sure that x and b belong to V and U*, respectively.
+            raise ValueError("Mismatching function spaces.")
```
connorjward, line 153: Could this be two checks with more specific error messages?

tests/firedrake/regression/test_assemble_baseform.py
```diff
@@ -155,6 +155,7 @@ def test_zero_form(M, f, one):
     assert abs(zero_form - 0.5 * np.prod(f.ufl_shape)) < 1.0e-12
 
 
+@pytest.mark.xfail(reason="action(M, M) raises primal-dual TypeError")
```
connorjward, line 158: This suggests that this PR breaks things for users?

connorjward (changes requested):

firedrake/assemble.py
pbrubeck: It will crash on `assemble(form, bcs=bcs, zero_bc_nodes=False)`. I think it only makes sense to apply bcs on things like `assemble(interpolate(expr, V), bcs=bcs)`, but I believe that this case is not supported.

tests/firedrake/regression/test_assemble_baseform.py
pbrubeck: I doubt anyone has done something like squaring a mass matrix to do something useful, and if they do they should stop, as that produces a mesh-dependent algorithm.

connorjward: Does that mean that this test should be deleted?

firedrake/assemble.py
connorjward: Maybe this could be rephrased to something like
```
# NOTE: This will only work for assembling XXX because YYY.
```

My complaint is just that the comment is unclear.

tests/firedrake/regression/test_assemble_baseform.py
pbrubeck: The test is testing something else, but I think it is very hard to recreate on an object that makes mathematical sense.

firedrake/cofunction.py
pbrubeck: I see. I copied this from `Function.assign`, and just noticed that `Function.assign` is decorated with `@FunctionMixin._ad_annotate_assign`. This is required for `bc.set(Cofunction, float)` if we want to set diagonal entries to 1.

tests/firedrake/regression/test_assemble_baseform.py
pbrubeck: I have deleted this test, as the behaivor we were testing cannot be reproduced with `action(action(Form, Interpolator), Function)`

firedrake/assemble.py
```diff
@@ -143,20 +143,18 @@ def get_assembler(form, *args, **kwargs):
...
         form = BaseFormAssembler.preprocess_base_form(form, mat_type=mat_type, form_compiler_parameters=fc_params)
     if isinstance(form, (ufl.form.Form, slate.TensorBase)) and not BaseFormAssembler.base_form_operands(form):
         diagonal = kwargs.pop('diagonal', False)
         if len(form.arguments()) == 0:
-            return ZeroFormAssembler(form, form_compiler_parameters=fc_params)
+            return ZeroFormAssembler(form, **kwargs)
         elif len(form.arguments()) == 1 or diagonal:
-            return OneFormAssembler(form, *args, bcs=bcs, form_compiler_parameters=fc_params, needs_zeroing=kwargs.get('needs_zeroing', True),
-                                    zero_bc_nodes=kwargs.get('zero_bc_nodes', False), diagonal=diagonal)
+            return OneFormAssembler(form, *args, diagonal=diagonal, **kwargs)
```
ksagiyam, line 157: We should make sure that we are passing the exact set of kwargs each assembler is expecting as we currently do.

firedrake/variational_solver.py
pbrubeck: I found a way to make `self.F = action(replace(F, {u: u_res}), interpolate(v_res, V))` work. It requires fixing some FormSum bugs, and fixing the way `assemble` optimizes and applies BCs on BaseForms.

ksagiyam: Sounds good.

pbrubeck:
Merged in https://github.com/firedrakeproject/firedrake/pull/3947

PR #3923, opened 2024-12-12
allow for Petrov-Galerkin formulations

ksagiyam:


firedrake/assemble.py
```diff
@@ -1180,25 +1180,31 @@ def allocate(self):
...
+            if not isinstance(bc, DirichletBC):
+                raise TypeError(f"diagonal expects a DirichletBC: got {bc}")
+            # Ignore self._zero_bc_nodes.
             tensor_func = tensor.riesz_representation(riesz_map="l2")
-            if self._diagonal:
-                bc.set(tensor_func, 1)
-            else:
-                bc.apply(tensor_func)
+            bc.set(tensor_func, 1)
             tensor.assign(tensor_func.riesz_representation(riesz_map="l2"))
```
pbrubeck, line 1190: I don't think bc.set does type checking, so the riesz_representation could go away. This will cause a conflict, as I also modified this code here https://github.com/firedrakeproject/firedrake/pull/3922

PR #3929, opened 2024-12-14
Single source of truth

pbrubeck:
# Description
With this PR, adding a new element to Firedrake would only require modifying the FIAT repo.

Depends on https://github.com/firedrakeproject/fiat/pull/120


tsfc/finatinterface.py
```diff
@@ -31,79 +31,16 @@
...
-}
+supported_elements = finat.supported_elements.copy()
 """A :class:`.dict` mapping UFL element family names to their
 FInAT-equivalent constructors.  If the value is ``None``, the UFL
 element is supported, but must be handled specially because it doesn't
 have a direct FInAT equivalent."""
 
+# These require special treatment below
+for tp_family in ("Q", "DQ", "DQ L2", "RTCE", "RTCF", "NCE", "NCF"):
+    supported_elements[tp_family] = None
```
connorjward, line 42: This seems like a bit of a code smell. And this does not actually achieve single source of truth since we still have some Firedrake-specific config.

connorjward (commented):
This seems a li

tsfc/finatinterface.py
pbrubeck: Should this special treatment be moved to FIAT?

connorjward: Probably. Why is it needed?

pbrubeck: The special treatment is for tensor-product elements. I seems like this file could be moved entirely to finat, since it does the ufl to finat conversion, and finat recently became ufl dependant.

connorjward: That seems reasonable

tests/firedrake/regression/test_interpolate_p3intmoments.py
```diff
@@ -89,7 +89,7 @@ def __init__(self, cell, degree):
         super().__init__(P3IntMoments(cell, degree))
 
 
-# Replace the old tsfc.finatinterface.convert dispatch with a new one that
+# Replace the old finat.element_factory.convert dispatch with a new one that
```
pbrubeck, line 92:
```suggestion
# Replace the old tsfc.finatinterface.convert dispatch with a new one that
```

connorjward (dismissed):

connorjward (approved):

PR #3932, opened 2024-12-16
Fix fieldsplit with `Cofunction` right hand side.

### LINKED ISSUE #3368: BUG: Fieldsplit doesn't work with a `Cofunction` right hand side.
**Describe the bug**

Using a fieldsplit preconditioner with a `Cofunction` as the right hand side of a `LinearVariationalProblem` crashes because `Cofunction` doesn't have a `_ufl_class_` attribute.

**Steps to Reproduce**
```python
from firedrake import *

mesh = UnitIntervalMesh(4)
W = V*V

u, p = TrialFunctions(W)
v, q = TestFunctions(W)

a = (u*v + p*q)*dx

L = Cofunction(W.dual())

params = {
    'pc_type': 'fieldsplit',
    'pc_fieldsplit_type': 'additive',
    'fieldsplit': {
        'ksp_type': 'preonly',
        'pc_type': 'lu'
    }
}

w = Function(W)
problem = LinearVariationalProblem(a, L, w)
solver = LinearVariationalSolver(problem, solver_parameters=params)

solver.solve()
```

**Expected behavior**
I would expect that this would solve the problem without error in one iteration.

**Error message**
```
  File "/home/jhc/codes/firedrake_examples/mfe/cofunction_fieldsplit.py", line 28, in <module>
    solver.solve()
  File "petsc4py/PETSc/Log.pyx", line 188, in petsc4py.PETSc.Log.EventDecorator.decorator.wrapped_func
  File "petsc4py/PETSc/Log.pyx", line 189, in petsc4py.PETSc.Log.EventDecorator.decorator.wrapped_func
  File "/home/jhc/codes/firedrake/src/firedrake/firedrake/adjoint_utils/variational_solver.py", line 89, in wrapper
    out = solve(self, **kwargs)
  File "/home/jhc/codes/firedrake/src/firedrake/firedrake/variational_solver.py", line 283, in solve
    self.snes.solve(None, work)
  File "petsc4py/PETSc/SNES.pyx", line 1555, in petsc4py.PETSc.SNES.solve
  File "petsc4py/PETSc/PETSc.pyx", line 323, in petsc4py.PETSc.PetscPythonErrorHandler
  File "petsc4py/PETSc/PETSc.pyx", line 323, in petsc4py.PETSc.PetscPythonErrorHandler
  File "petsc4py/PETSc/PETSc.pyx", line 323, in petsc4py.PETSc.PetscPythonErrorHandler
  [Previous line repeated 6 more times]
  File "petsc4py/PETSc/petscdmshell.pxi", line 341, in petsc4py.PETSc.DMSHELL_CreateFieldDecomposition
  File "petsc4py/PETSc/Log.pyx", line 188, in petsc4py.PETSc.Log.EventDecorator.decorator.wrapped_func
  File "petsc4py/PETSc/Log.pyx", line 189, in petsc4py.PETSc.Log.EventDecorator.decorator.wrapped_func
  File "/home/jhc/codes/firedrake/src/firedrake/firedrake/dmhooks.py", line 347, in create_field_decomposition
    ctxs = ctx.split([(i, ) for i in range(len(W))])
  File "petsc4py/PETSc/Log.pyx", line 188, in petsc4py.PETSc.Log.EventDecorator.decorator.wrapped_func
  File "petsc4py/PETSc/Log.pyx", line 189, in petsc4py.PETSc.Log.EventDecorator.decorator.wrapped_func
  File "/home/jhc/codes/firedrake/src/firedrake/firedrake/solving_utils.py", line 328, in split
    F = splitter.split(problem.F, argument_indices=(field, ))
  File "petsc4py/PETSc/Log.pyx", line 188, in petsc4py.PETSc.Log.EventDecorator.decorator.wrapped_func
  File "petsc4py/PETSc/Log.pyx", line 189, in petsc4py.PETSc.Log.EventDecorator.decorator.wrapped_func
  File "/home/jhc/codes/firedrake/src/firedrake/firedrake/formmanipulation.py", line 59, in split
    f = map_integrand_dags(self, form)
  File "/home/jhc/codes/firedrake/src/ufl/ufl/algorithms/map_integrands.py", line 73, in map_integrand_dags
    return map_integrands(lambda expr: map_expr_dag(function, expr, compress),
  File "/home/jhc/codes/firedrake/src/ufl/ufl/algorithms/map_integrands.py", line 37, in map_integrands
    mapped_components = [map_integrands(function, component, only_integral_type)
  File "/home/jhc/codes/firedrake/src/ufl/ufl/algorithms/map_integrands.py", line 37, in <listcomp>
    mapped_components = [map_integrands(function, component, only_integral_type)
  File "/home/jhc/codes/firedrake/src/ufl/ufl/algorithms/map_integrands.py", line 66, in map_integrands
    return function(integrand)
  File "/home/jhc/codes/firedrake/src/ufl/ufl/algorithms/map_integrands.py", line 73, in <lambda>
    return map_integrands(lambda expr: map_expr_dag(function, expr, compress),
  File "/home/jhc/codes/firedrake/src/ufl/ufl/corealg/map_dag.py", line 34, in map_expr_dag
    result, = map_expr_dags(function, [expression], compress=compress,
  File "/home/jhc/codes/firedrake/src/ufl/ufl/corealg/map_dag.py", line 100, in map_expr_dags
    r = handlers[v._ufl_typecode_](v, *[vcache[u] for u in v.ufl_operands])
  File "/home/jhc/codes/firedrake/src/ufl/ufl/corealg/multifunction.py", line 99, in undefined
    raise ValueError(f"No handler defined for {o._ufl_class_.__name__}.")
AttributeError: 'Cofunction' object has no attribute '_ufl_class_'
```

**Environment:**
 - OS: PopOS 22.04
 - Python version: 3.10.12
 - Output of `firedrake-status`:
```
Firedrake Configuration:
    package_manager: True
    minimal_petsc: False
    mpicc: None
    mpicxx: None
    mpif90: None
    mpiexec: None
    disable_ssh: False
    honour_petsc_dir: False
    with_parmetis: False
    slepc: False
    packages: []
    honour_pythonpath: False
    opencascade: False
    tinyasm: False
    torch: False
    petsc_int_type: int32
    cache_dir: /home/jhc/codes/firedrake/.cache
    complex: False
    remove_build_files: False
    with_blas: None
    netgen: False
Additions:
    None
Environment:
    PYTHONPATH: None
    PETSC_ARCH: None
    PETSC_DIR: None
Status of components:
|Package             |Branch                        |Revision  |Modified  |
|FInAT               |master                        |e2805c4   |False     |
|PyOP2               |master                        |fbde61f9  |False     |
|fiat                |master                        |e7b2909   |False     |
|firedrake           |master                        |ebd190f27 |False     |
|h5py                |firedrake                     |4c01efa9  |False     |
|libspatialindex     |master                        |4768bf3   |True      |
|libsupermesh        |master                        |dbe226b   |False     |
|loopy               |main                          |8158afdb  |False     |
|petsc               |firedrake                     |09f36907a6|False     |
|pyadjoint           |master                        |f194553   |False     |
|pytest-mpi          |main                          |a478bc8   |False     |
|tsfc                |master                        |799191d   |False     |
|ufl                 |master                        |054b0617  |False     |
```

JHopeCollins:
## Problem description
As described in #3368, it currently isn't possible to use PCFieldsplit if the form contains a `Cofunction`.
This is because the way we split the form is by traversing the dag with a `ufl.MultiFunction` and splitting each node as it is visited. However, none of the current handlers will deal with a `Cofunction`.

This PR adds a `Cofunction` handler to the `ExtractSubBlock` multifunction to do this splitting.
Fixes #3368 

## What does the `Cofunction` handler need to achieve?
A fieldsplit solver needs the right hand side from the corresponding field. For a `Cofunction` this means the values in the corresponding `subfunction`. So when we split a form the return value needs to live in the correct subspace, and point to the numerical values from the full space.

## How is this implemented?
If all terms in the form have 1 argument (i.e. a `FormSum` of `Cofunctions` and 1-forms), then life is simple. We just check which field(s) is being requested, and make a new `Cofunction` in that (possibly mixed) subspace using the `Dats` from the original `Cofunction`.

If some terms in the full `Form` have 2 arguments, then life is a little more complicated, and we have two possibilities.

1. If we're being asked for a block on the diagonal, then a) numerically this is probably going to be inverted and so needs a right hand side and b) mathematically this acts on a member of a function space and returns an assembled thing in the dual space - this is exactly what the `Cofunction` is (I'm jetlagged and I've probably slaughtered the dual/primal space technicalities. Edits to fix the explanation welcome). In this case we do the same as for the 1-form case i.e. we just make a new smaller `Cofunction` that views the appropriate parts of the full `Cofunction`.

2. If we're being asked for an off-diagonal block, then a) numerically this doesn't make sense to invert and hence to have a right hand side, and b) mathematically the test and trial spaces are different, which doesn't make sense to have a corresponding `Cofunction` (I think?). In this case we just return a `ZeroBaseForm` which will get optimised out before we do any actual computational work.

## Other changes
Because we can now split a `Cofunction` or `FormSum`, not just a `Form`, I had to widen the logic in `split_form` for checking if the form returned by `ExtractSubBlock` is empty or not. It works but it might not be the best/most robust way of checking.

## What if this all breaks in the future?
There's a test that solves a mixed formulation of the wave equation using a Schur factorisation with both a 1-form right hand side and an equivalent `Cofunction` right hand side, and checks the results match close to machine precision.
The Schur complement is built by asking PETSc to do the brute force thing. Computationally suboptimal but the mesh is tiny, it reduces the amount of other "stuff" needed, and testing implementations is basically what the brute force thing is for.


JHopeCollins:
@stephankramer I see that you were also having issues with this in https://github.com/thetisproject/thetis/pull/376. If you/someone else from the Thetis team could checkout this branch and make sure it fixes things for you then that would be really helpful.

firedrake/formmanipulation.py
```diff
@@ -127,6 +128,59 @@ def argument(self, o):
...
+        except TypeError:  # Only one index
+            indices = (indices, )
+            nidx = 1
+
+        # the cofunction should only be returned on the
+        # diagonal elements, so if we are off-diagonal
+        # on a two-form then we return a zero form,
+        # analogously to the off components of arguments.
+        if len(self.blocks) == 2:
+            itest, itrial = self.blocks[0], self.blocks[1]
```
pbrubeck, line 157:
```suggestion
            itest, itrial = self.blocks
```

firedrake/formmanipulation.py
```diff
@@ -127,6 +128,59 @@ def argument(self, o):
...
+        if on_diag:
+            if nidx == 1:
+                from firedrake import FunctionSpace
+                i = indices[0]
+                W = V_is[i]
+                W = FunctionSpace(W.mesh(), W.ufl_element())
+                c = Cofunction(W.dual(), val=o.subfunctions[i].dat)
+            else:
+                from firedrake import MixedFunctionSpace
+                from pyop2 import MixedDat
```
pbrubeck, line 175: Is there a good reason to not import these at the top of the script?

tests/firedrake/regression/test_fieldsplit_cofunction.py
```diff
@@ -0,0 +1,60 @@
+import firedrake as fd
+
+
+def test_fieldsplit_cofunction():
```
pbrubeck, line 4: Was there an already existing test for Fieldsplit that used a form RHS that could be extended to test Cofunctions?

tests/firedrake/regression/test_fieldsplit_cofunction.py
```diff
@@ -0,0 +1,60 @@
...
+                                              solver_parameters=params)
+    solver_check.solve()
+
+    w_cofun = fd.Function(W)
+    problem_cofun = fd.LinearVariationalProblem(a, L_cofun, w_cofun)
+    solver_cofun = fd.LinearVariationalSolver(problem_cofun,
+                                              solver_parameters=params)
+    solver_cofun.solve()
+
+    assert fd.errornorm(w_check, w_cofun) < 1e-14
```
pbrubeck, line 60: Can the test check that we accurately solve the problem, not just that the solvers agree?

firedrake/formmanipulation.py
```diff
@@ -127,6 +128,59 @@ def argument(self, o):
...
+            indices = (indices, )
+            nidx = 1
+
+        # the cofunction should only be returned on the
+        # diagonal elements, so if we are off-diagonal
+        # on a two-form then we return a zero form,
+        # analogously to the off components of arguments.
+        if len(self.blocks) == 2:
+            itest, itrial = self.blocks[0], self.blocks[1]
+            on_diag = (itest == itrial)
```
pbrubeck, line 158: Is this handling `assemble(bilinear_form, diagonal=True)`? Is this covered by the tests?

firedrake/formmanipulation.py
```diff
@@ -168,7 +224,16 @@ def split_form(form, diagonal=False):
...
     for idx in numpy.ndindex(shape):
         f = splitter.split(form, idx)
-        if len(f.integrals()) > 0:
+
+        # does f actually contain anything?
+        if isinstance(f, Cofunction):
+            flen = 1
+        elif isinstance(f, FormSum):
+            flen = len(f.components())
+        else:  # Form
```
pbrubeck, line 233: Raise a ValueError if f is not a Form

firedrake/formmanipulation.py
```diff
@@ -127,6 +128,59 @@ def argument(self, o):
...
+                from firedrake import FunctionSpace
+                i = indices[0]
+                W = V_is[i]
+                W = FunctionSpace(W.mesh(), W.ufl_element())
+                c = Cofunction(W.dual(), val=o.subfunctions[i].dat)
+            else:
+                from firedrake import MixedFunctionSpace
+                from pyop2 import MixedDat
+                W = MixedFunctionSpace([V_is[i] for i in indices])
+                c = Cofunction(W.dual(), val=MixedDat(o.subfunctions[i].dat
```
pbrubeck, line 177:
```suggestion
                c = Cofunction(W.dual(), val=MixedDat(o.dat[i]
```

pbrubeck (dismissed):

firedrake/formmanipulation.py
```diff
@@ -127,6 +128,59 @@ def argument(self, o):
                          for j in numpy.ndindex(V_is[i].value_shape)]
         return self._arg_cache.setdefault(o, as_vector(args))
 
+    def cofunction(self, o):
```
Ig-dolci, line 131: It is just a question. Where is this function called?

firedrake/formmanipulation.py
JHopeCollins: They're only used here, not anywhere else

firedrake/formmanipulation.py
JHopeCollins: Do you mean 1) is this handling the case when `assemble(..., diagonal=True)`, or 2) will this work for `assemble(..., diagonal=True)`.

If 1) then no, this is handling the case where a diagonal block is being requested e.g. calling split with say `argument_indices=(1,1)`. The full form doesn't have to be diagonal.

If 2) then I don't know, I haven't tested it but I don't see why it wouldn't, this is just a symbolic step whereas it looks like the `diagonal` kwarg for `assemble` is an optimisation for the numerical assembly.

tests/firedrake/regression/test_fieldsplit_cofunction.py
JHopeCollins: Sure, that's a better idea. This was just making the script I was bugfixing with into a test. I'll find a suitable existing one and parameterise it.

tests/firedrake/regression/test_fieldsplit_cofunction.py
JHopeCollins: I'll add a cofunction rhs version to one of the existing tests that checks the solution quality to cover this.

firedrake/formmanipulation.py
JHopeCollins: `ufl.MultiFunction` seems to be essentially a hand-rolled vtable (`Function` as in "multiple functions", not `Function` in the finite element sense), so when `ExtractSubBlocks` calls `map_integrand_dags(self, form)`, the form DAG is traversed and at each node it will look for a method on `ExtractSubBlocks` called `type(node).__name__.lower()`, i.e. `Argument` becomes `argument`, and `Expr` becomes `expr` etc.

So when it gets to a `Cofunction`, it will looks for a method on `ExtractSubBlocks` called `cofunction`. This was missing which was the error that was previously thrown.

firedrake/formmanipulation.py
```diff
@@ -127,6 +128,56 @@ def argument(self, o):
...
+        from firedrake import Cofunction, FunctionSpace, MixedFunctionSpace
+        from pyop2 import MixedDat
+        from ufl.classes import ZeroBaseForm
```
connorjward, line 134: These should be top-level imports. The `firedrake` ones may need to be here to avoid circular imports but ideally should be moved too.

firedrake/formmanipulation.py
```diff
@@ -127,6 +128,56 @@ def argument(self, o):
...
+        # We only need the test space for Cofunction
+        indices = self.blocks[0]
+        V_is = V.subfunctions
+
+        try:
+            indices = tuple(indices)
+            nidx = len(indices)
+        except TypeError:  # Only one index
+            indices = (indices, )
+            nidx = 1
```
connorjward, line 151: Can this be
```py
if isinstance(indices, numbers.Integral):
    indices = (indices,)
```
Flow control with exceptions is generally less clear than conditionals.

`nidx` isn't really needed nor I expect is casting `indices = tuple(indices)`.

firedrake/formmanipulation.py
```diff
@@ -127,6 +128,56 @@ def argument(self, o):
...
+        # the full space. This means that the right hand side
+        # of the fieldsplit problem will be correct.
+        if on_diag:
+            if nidx == 1:
+                i = indices[0]
+                W = V_is[i]
+                W = FunctionSpace(W.mesh(), W.ufl_element())  # primal space
+                c = Cofunction(W.dual(), val=o.subfunctions[i].dat)
+            else:
+                W = MixedFunctionSpace([V_is[i] for i in indices])  # dual space
```
connorjward, line 174: Isn't `V_is` a list of `Functions`? This doesn't look right to me.

The primal/dual split here also seems a bit confusing. Why should we have a primal thing if `len(indices) == 1` and a dual thing otherwise?

firedrake/formmanipulation.py
```diff
@@ -160,6 +211,8 @@ def split_form(form, diagonal=False):
     compiler will remove these in its more complex simplification
     stages.
     """
+    from firedrake import Cofunction
+    from ufl import FormSum, Form
```
connorjward, line 215: Move imports out

firedrake/formmanipulation.py
```diff
@@ -168,7 +221,20 @@ def split_form(form, diagonal=False):
...
-        if len(f.integrals()) > 0:
+
+        # does f actually contain anything?
+        if isinstance(f, Cofunction):
+            flen = 1
+        elif isinstance(f, FormSum):
+            flen = len(f.components())
+        elif isinstance(f, Form):
+            flen = len(f.integrals())
+        else:  # Form
```
connorjward, line 232: Comment doesn't make sense. You have `if isinstance(f, Form)` above.

connorjward (changes requested):

tests/firedrake/regression/test_matrix_free.py
```diff
@@ -184,7 +185,12 @@ def test_fieldsplitting(mesh, preassembled, parameters):
 
     a = inner(u, v)*dx
 
-    L = inner(expect, v)*dx
+    if rhs == 'func_rhs':
+        L = inner(expect, v)*dx
+    elif rhs == 'cofunc_rhs':
+        L = expect.riesz_representation()
+    else:
+        raise ValueError("Unknown right hand side type")
```
pbrubeck, line 193:
```suggestion
    L = inner(expect, v)*dx
    if rhs == 'cofunc_rhs':
        L = assemble(L)
    elif rhs != 'form_rhs':
        raise ValueError("Unknown right hand side type")
```

tests/firedrake/regression/test_matrix_free.py
```diff
@@ -130,6 +130,7 @@ def test_matrixfree_action(a, V, bcs):
 
 @pytest.mark.parametrize("preassembled", [False, True],
                          ids=["variational", "preassembled"])
+@pytest.mark.parametrize("rhs", ["func_rhs", "cofunc_rhs"])
```
pbrubeck, line 133:
```suggestion
@pytest.mark.parametrize("rhs", ["form_rhs", "cofunc_rhs"])
```

firedrake/formmanipulation.py
JHopeCollins: `V_is` is a list of `FunctionSpaces`. I think the confusion is because when `Function.split` was renamed `Function.subfunctions`, the same change was made for spaces, i.e. `FunctionSpace.split` became `FunctionSpace.subfunctions`.

The primal/dual thing is annoying. If the split is for multiple fields then we need a `MixedFunctionSpace` and can just pass a list of dual spaces and it gives a mixed dual space. For the single field split case, `FunctionSpace` with the ufl_element gives a primal function space which is why the extra `W.dual()` is needed.
I'll see if there's a nicer way of doing it.

firedrake/formmanipulation.py
JHopeCollins: I based the `cofunction` method from the `argument` method so this is left over from there. I'll see what I can take out.

JHopeCollins: Changed to testing `isinstance(indices, int)` and updated the `arguments` method too. `nidx` removed for `len(indices)` because it's only used once.

firedrake/formmanipulation.py
JHopeCollins: Changed to using `DualSpace` to create the single component space. I'm sure I tried this before and got errors so I must have been doing something wrong then.

firedrake/formmanipulation.py
JHopeCollins: @pbrubeck can I resolve this thread?

tests/firedrake/regression/test_fieldsplit_cofunction.py
JHopeCollins: I parameterised a couple extra tests with cofunction rhs because the one in `test_matrix_free` just solves a mass matrix so the form is block diagonal.
The two extra tests have off-diagonal blocks in the form so will make sure that we DTRT in that case.

tests/firedrake/regression/test_nullspace.py
```diff
@@ -323,6 +324,10 @@ def test_near_nullspace_mixed(aux_pc):
 
     f = as_vector((0, -9.8*conditional(inside_box, 2, 1)))
     L = inner(f, v)*dx
+    if rhs == 'cofunc_rhs':
+        L = assemble(L)
```
pbrubeck, line 328: Is this test passing without bcs here?

JHopeCollins: The test passed. Were you not expecting it to?

connorjward (approved):
LGTM

JHopeCollins:
@pbrubeck unless you have any other issues I'd like to merge this PR.

PR #3941, opened 2024-12-18
Dham/abstract reduced functional

dham:


firedrake/adjoint/__init__.py
```diff
@@ -37,7 +37,7 @@
 
 from firedrake.adjoint.ufl_constraints import UFLInequalityConstraint, \
     UFLEqualityConstraint  # noqa F401
-from firedrake.adjoint.ensemble_reduced_functional import EnsembleReducedFunctional  # noqa F401
+from .ensemble_reduced_functional import EnsembleReducedFunctional  # noqa F401
```
JHopeCollins, line 40:
```suggestion
from firedrake.adjoint.ensemble_reduced_functional import EnsembleReducedFunctional  # noqa F401
```

tests/firedrake/output/test_adjoint_disk_checkpointing.py
```diff
@@ -0,0 +1,106 @@
...
+
+    assert np.allclose(J, Jnew)
+
+    grad_Jnew = Jhat.derivative(apply_riesz=True)
+
+    return Jnew, grad_Jnew
+
+
+@pytest.mark.skipcomplex
+# A serial version of this test is included in the pyadjoint tests.
```
Ig-dolci, line 63:
```suggestion
```

tests/firedrake/adjoint/test_ensemble_reduced_functional.py
```diff
@@ -88,26 +104,50 @@ def test_verification_gather_functional_Function():
...
     assert taylor_test(rf, x, Function(R, val=0.1)) > 1.9
 
 
-@pytest.mark.parallel(nprocs=6)
+@pytest.mark.parallel(nprocs=3)
 @pytest.mark.skipcomplex  # Taping for complex-valued 0-forms not yet done
 def test_minimise():
     # Optimisation test using a list of controls.
     # This test is equivalent to the one found at:
     # https://github.com/dolfin-adjoint/pyadjoint/blob/master/tests/firedrake_adjoint/test_optimisation.py#L9.
```
Ig-dolci, line 137:
```suggestion
    # https://github.com/firedrakeproject/firedrake/blob/master/tests/firedrake/adjoint/test_optimisation.py#L92
```

.github/workflows/docs.yml
```diff
@@ -33,6 +33,8 @@ jobs:
           . venv/bin/activate
           pip uninstall -y firedrake
           pip install --verbose './firedrake-repo[docs]'
+          pip list
+          pip install --verbose git+https://github.com/dolfin-adjoint/pyadjoint.git@dham/abstract_reduced_functional#egg=pyadjoint-ad'
```
JHopeCollins, line 37:
```suggestion
          pip install --verbose git+https://github.com/dolfin-adjoint/pyadjoint.git@dham/abstract_reduced_functional#egg=pyadjoint-ad
```

firedrake/cofunction.py
```diff
@@ -372,3 +379,130 @@ def __str__(self):
...
+        rhs = Cofunction(self._function_space.dual())
+        soln = Function(self._function_space)
+        lvp = LinearVariationalProblem(
+            self._inner_product, rhs, soln, bcs=self._bcs,
+            restrict=self._restrict,
+            form_compiler_parameters=self._form_compiler_parameters)
+        solver = LinearVariationalSolver(
+            lvp, solver_parameters=self._solver_parameters
+        )
+        return solver.solve, rhs, soln
```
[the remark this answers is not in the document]
pbrubeck: It will be a bit tricky, as LinearSolver only takes in an assembled form. We would either have to restrict the form before the assembly (which requires duplicating the `restrict` functionality), or extend `LinearSolver` to the unassembled case.

.github/workflows/core.yml
```diff
@@ -195,6 +195,8 @@ jobs:
             --no-binary h5py \
             --extra-index-url https://download.pytorch.org/whl/cpu \
             "$(echo ./firedrake-repo/dist/firedrake-*.tar.gz)[ci,docs]"
+          : # TODO: Remove before merge
+          pip install --verbose --editable git+https://github.com/dolfin-adjoint/pyadjoint.git@dham/abstract_reduced_functional#egg=pyadjoint-ad
```
JHopeCollins, line 199:
```suggestion
```

firedrake/cofunction.py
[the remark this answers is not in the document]
JHopeCollins: Seeing as using `LinearSolver` should have the same result as using `LinearVariationalSolver`, and it seems to be a non-trivial change, this should be a separate PR later.

firedrake/cofunction.py
```diff
@@ -372,3 +379,130 @@ def __str__(self):
...
+    @cached_property
+    def _solver(self):
+        from firedrake import (LinearVariationalSolver,
+                               LinearVariationalProblem, Function, Cofunction)
+        rhs = Cofunction(self._function_space.dual())
+        soln = Function(self._function_space)
+        lvp = LinearVariationalProblem(
+            self._inner_product, rhs, soln, bcs=self._bcs,
+            restrict=self._restrict,
+            form_compiler_parameters=self._form_compiler_parameters)
```
JHopeCollins: Would the inner product form ever change? If not, then we should use the `constant_jacobian` kwarg to prevent the matrix being refactorised every time. Seeing as the default solver is LU, for large problems this could be a big performance hit.

```suggestion
        lvp = LinearVariationalProblem(
            self._inner_product, rhs, soln, bcs=self._bcs,
            restrict=self._restrict, constant_jacobian=True,
            form_compiler_parameters=self._form_compiler_parameters)
```

firedrake/adjoint_utils/blocks/solving.py
```diff
@@ -779,7 +786,11 @@ def evaluate_adj_component(self, inputs, adj_inputs, block_variable, idx,
         replace_map[self.func] = self.get_outputs()[0].saved_output
         dFdm = replace(dFdm, replace_map)
 
-        dFdm = dFdm * adj_sol
+        if isinstance(dFdm, firedrake.Argument):
+            #  Corner case. Should be fixed more permanently upstream in UFL.
```
JHopeCollins, line 790: Ideally this should have a link to a github issue. Is this the relevant UFL issue: https://github.com/FEniCS/ufl/issues/301? If not then we should open one.

We should also open a Firedrake issue to keep track of the fact that this needs updating once the UFL problem is fixed.

firedrake/adjoint_utils/function.py
```diff
@@ -218,72 +218,15 @@ def _ad_create_checkpoint(self):
...
-                return firedrake.Cofunction(V.dual())
-            else:
-                return value
+    def _ad_init_zero(self, dual=False):
+        from firedrake import Function, Cofunction
+        if dual:
+            return Cofunction(self.function_space().dual())
         else:
-            return self._ad_convert_riesz(value, options=options)
+            return Function(self.function_space())
```
JHopeCollins, line 229: Do we actually _guarantee_ that a newly instantiated `Function` or `Cofunction` will be zero initialised? If not then these should call `Function.zero()` because this method does guarantee zero initialisation.

JHopeCollins (dismissed):
Looking really good, I just had a couple of small comments.

firedrake/adjoint_utils/function.py
dham: Yes this is guaranteed. The ultimate line of code that does this is https://github.com/firedrakeproject/firedrake/blob/6f853a7370f6ef730a6e6e3bcf01089423abe50d/pyop2/types/data_carrier.py#L72

JHopeCollins: Ok good, I'll resolve this then.

firedrake/cofunction.py
```diff
@@ -464,7 +464,7 @@ def _solver(self):
         soln = Function(self._function_space)
         lvp = LinearVariationalProblem(
             self._inner_product, rhs, soln, bcs=self._bcs,
-            restrict=self._restrict,
+            restrict=self._restrict, constant_jacobian=True,
```
pbrubeck: What if the mesh moves, or the user-provided inner product has coefficients that change?

firedrake/cofunction.py
```diff
@@ -372,3 +379,130 @@ def __str__(self):
...
+    form_compiler_parameters: dict
+        A dictionary of form compiler parameters to be passed to the
+        variational problem that solves for the Riesz map.
+    restrict: bool
+        If `True`, use restricted function spaces in the Riesz map solver.
+    """
+
+    def __init__(self, function_space_or_inner_product=None,
+                 sobolev_space=ufl.L2, *, bcs=None, solver_parameters=None,
+                 form_compiler_parameters=None, restrict=True):
```
pbrubeck:
```suggestion
                 form_compiler_parameters=None, restrict=True, 
                 constant_jacobian=False):
```

firedrake/cofunction.py
```diff
@@ -372,3 +379,130 @@ def __str__(self):
...
+                inner_product = RieszMap._inner_product_form(
+                    sobolev_space, u, v
+                )
+
+        self._function_space = function_space
+        self._inner_product = inner_product
+        self._bcs = bcs
+        self._solver_parameters = solver_parameters or {}
+        self._form_compiler_parameters = form_compiler_parameters or {}
+        self._restrict = restrict
```
pbrubeck, line 445:
```suggestion
        self._restrict = restrict
        self._constant_jacobian = constant_jacobian
```

firedrake/cofunction.py
pbrubeck:
```suggestion
            restrict=self._restrict,
            constant_jacobian=self._constant_jacobian,
```

firedrake/cofunction.py
```diff
@@ -372,3 +379,130 @@ def __str__(self):
...
+        Used to determine the inner product.
+    bcs: DirichletBC or list of DirichletBC
+        Boundary conditions to apply to the Riesz map.
+    solver_parameters: dict
+        A dictionary of PETSc options to be passed to the solver.
+    form_compiler_parameters: dict
+        A dictionary of form compiler parameters to be passed to the
+        variational problem that solves for the Riesz map.
+    restrict: bool
+        If `True`, use restricted function spaces in the Riesz map solver.
```
pbrubeck, line 407: Setting `restrict=True` as default means that most solvers will not be supported until we merge https://github.com/firedrakeproject/firedrake/pull/4169

dham: Looking forward it's the right way to do it, though. In the meantime users can always set restrict to False if they need to. It's only a default. Most users won't even see this because they won't use strong BCs in the Riesz map.

JHopeCollins (approved):
This is ready to go. The only CI failure is an unrelated linkcheck

JHopeCollins (dismissed):
This is ready to go. The only CI failure is an unrelated linkcheck

.github/workflows/core.yml
```diff
@@ -212,6 +212,7 @@ jobs:
           pip install -U pip
           pip install --group ./firedrake-repo/pyproject.toml:ci
 
+
```
JHopeCollins:
```suggestion
```

JHopeCollins (approved):

PR #3947, opened 2024-12-30
Fieldsplit: replace empty Forms with ZeroBaseForm

pbrubeck:
# Description
This PR addresses the consequences of https://github.com/FEniCS/ufl/pull/336, which introduces more agressive simplifications that cause empty forms to be returned when extracting a subblock that is supposed to be zero (e.g. the pressure block in Stokes).

Here empty Jacobians are replaced with ZeroBaseForm, as we require two Arguments in order for the adjoint to make sense.

The simplifications should also speed up code generation and even reduce flop counts in some cases.


firedrake/formmanipulation.py
```diff
@@ -52,15 +65,22 @@ def split(self, form, argument_indices):
...
-        self.blocks = dict(enumerate(argument_indices))
+        self.blocks = dict(enumerate(map(as_tuple, argument_indices)))
         if len(args) == 0:
             # Functional can't be split
             return form
         if all(len(a.function_space()) == 1 for a in args):
             assert (len(idx) == 1 for idx in self.blocks.values())
             assert (idx[0] == 0 for idx in self.blocks.values())
             return form
+        # TODO find a way to distinguish empty Forms avoiding expand_derivatives
```
ksagiyam, line 76: I am not sure if there is a robust solution for this. `expand_derivatives()` might be expensive for large forms. Can we not handle zero form after calling tsfc_compile_form()? Do we do this here for Slate?

firedrake/adjoint_utils/variational_solver.py
```diff
@@ -20,10 +20,13 @@ def wrapper(self, *args, **kwargs):
...
-                                  self.u_restrict,
-                                  TrialFunction(self.u_restrict.function_space()))
-                self._ad_adj_F = adjoint(dFdu)
+                dFdu = derivative(self.F, self.u_restrict)
+                try:
+                    self._ad_adj_F = adjoint(dFdu)
+                except ValueError:
+                    # Try again without expanding derivatives,
+                    # as dFdu might have been simplied to an empty Form
+                    self._ad_adj_F = adjoint(dFdu, derivatives_expanded=True)
```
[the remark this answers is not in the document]
pbrubeck: The one checking that we indeed have a bilinear form (the only `ValueError` in `firedrake.adjoint`) https://github.com/firedrakeproject/firedrake/blob/2a0c03b244ef9f641fe959d026225d49dd007a14/firedrake/ufl_expr.py#L334
There is a similiar one in `ufl.adjoint`
https://github.com/firedrakeproject/ufl/blob/bfa596dd3f0e121168305451badf6b24c98dd7f6/ufl/adjoint.py#L72

firedrake/formmanipulation.py
pbrubeck: That's why I put the comment there. But both `action` and `adjoint` call `expand_derivatives` by default. Maybe this is a new behaivour that was intruduced alongside BaseForm. If `expand_derivatives` were not called, Fieldsplit/SLATE/EquationBC would not reach the empty Forms.

pbrubeck: https://github.com/firedrakeproject/ufl/blob/bfa596dd3f0e121168305451badf6b24c98dd7f6/ufl/formoperators.py#L174

ksagiyam: I see. I think I am fine with this.

firedrake/formmanipulation.py
```diff
@@ -98,76 +118,42 @@ def argument(self, o):
...
-                W = V_is[i]
-                W = DualSpace(W.mesh(), W.ufl_element())
-                c = Cofunction(W, val=o.subfunctions[i].dat)
-            else:
-                W = MixedFunctionSpace([V_is[i] for i in indices])
-                c = Cofunction(W, val=MixedDat(o.dat[i] for i in indices))
+        if len(indices) == 1:
+            i = indices[0]
+            W = V[i]
+            W = DualSpace(W.mesh(), W.ufl_element())
```
JHopeCollins, line 152: Can you use the `subspace` function here too? It will need a switch adding for primal/dual but I think it would be nicer to use `subspace` for both.

I think it would also be nice to lift `subspace` out, possibly into one of the functionspace files. It's useful more generally - see for example line 678 in `slate.py` where we have the same if/else block to construct the subspace.

pbrubeck: I was thinking to extend `FunctionSpace.__getitem__` to allow a list index return the subspace, since this is how sub-arrays are extracted in numpy.

firedrake/slate/slate.py
```diff
@@ -664,21 +670,16 @@ def _split_arguments(self):
...
             if len(idx) == 1:
-                fidx, = idx
-                W = V_is[fidx]
+                W = V[idx[0]]
                 W = FunctionSpace(W.mesh(), W.ufl_element())
             else:
-                W = MixedFunctionSpace([V_is[fidx] for fidx in idx])
+                W = MixedFunctionSpace([V[fidx] for fidx in idx])
 
             nargs.append(Argument(W, arg.number(), part=arg.part()))
```
JHopeCollins, line 684: Using `subspace` this whole function could just be be replaced with
```suggestion
         return (Argument(subspace(V, self._blocks[i]), arg.number(), part=arg.part())
                 for i, arg in enumerate(self.operands[0].arguments()))
```

firedrake/formmanipulation.py
JHopeCollins: I was actually thinking exactly the same thing the other day.

It would be great to have it for `Function` too so that you can get a `Mixed` subfunction viewing the relevant subcomponents. `ExtractSubBlock.cofunction` would then basically just be calling `cofunction[idxs]`.
There are definitely other places where that would be useful too, for example in the `EnsembleFunction` I've been working on [here](https://github.com/firedrakeproject/firedrake/blob/dfe98bf906b0c063822abec533fccc1d3bea67f3/firedrake/ensemblefunction.py#L59).

What does `FunctionSpace.collapse()` do? It just looks like it would return a copy of the `FunctionSpace`.

pbrubeck: @ksagiyam what do you think about this __getitem__ idea? Currenty, indexing by a scalar returns `IndexedProxyFunctionSpace`, but indexing by a list of a single item would return the collapsed version of this space.

connorjward: `__getitem__` for `Function` is implemented in UFL so the changes would be needed there I think.

pbrubeck: I'm not to worried for the `Function` case, that one seems more involved, as they are Expressions in the UFL sense, and extracting a component naturally means a scalar compoment, but here it seems that we want a MixedFunctionSpace (possibly vector-valued) component.

JHopeCollins: It could also be added to `sub` or `subfunctions`, although they both also have issues.
`sub` explicitly says it is for bcs, and `subfunctions` is already a tuple not a method so it would probably have to become a local class with `__getitem__` overridden so we could deal with single or multiple indices.

ksagiyam: That sounds slightly odd to me. Can you instead rewrite `collapse()` so that it would take a list of indices as an optional argument?

@JHopeCollins `collapse()` basically forgets parents. They do something nontrivial in ProxyFunctionSpace and in MixedFunctionSpace.

JHopeCollins: Ok thanks.

It does strip the parent information, I don't think the name `collapse` would be where I would look for this functionality - it would be more intuitive in subfunctions or sub.

firedrake/functionspaceimpl.py
```diff
@@ -321,6 +321,14 @@ def __iter__(self):
         return iter(self.subfunctions)
 
     def __getitem__(self, i):
+        from firedrake.functionspace import MixedFunctionSpace
+        if isinstance(i, list):
```
JHopeCollins, line 325: We should allow any iterable, not just a list.
```suggestion
        if not isinstance(i, int):
```

JHopeCollins (changes requested):

firedrake/functionspaceimpl.py
```diff
@@ -321,6 +321,14 @@ def __iter__(self):
         return iter(self.subfunctions)
 
     def __getitem__(self, i):
+        from firedrake.functionspace import MixedFunctionSpace
```
JHopeCollins, line 324: I think we should cache the mixed subspaces like we do for the non-mixed subspaces. Otherwise we get a "different" space every time we call with the same indices.

JHopeCollins (changes requested):

firedrake/functionspaceimpl.py
```diff
@@ -321,6 +321,14 @@ def __iter__(self):
...
+        if isinstance(i, list):
+            # Return a collapsed subspace if the index is a list
+            if len(i) == 1:
+                return self[i[0]].collapse()
+            else:
+                return MixedFunctionSpace([self[isub] for isub in i])
+
```
ksagiyam, line 331: As I said elsewhere, I think this is odd. I feel we are abusing `__getitem__()`. If we want this feature, we should do it somewhere else (in a new method or in `collapse()`), not in `__getitem__()`.

ksagiyam (changes requested):

firedrake/functionspaceimpl.py
connorjward: I think the behaviour of `__getitem__` is a bit subtle here. Calling `obj[1, 2, 3]` is *exactly the same* as calling `obj[(1, 2, 3)]` which is not quite the same as `obj[[1, 2, 3]]`.

I think having something like `space[1, 2]` would suggest to someone that `space` is somehow nested, whereas `space[[1, 2]]` suggests a subset of the indices.

firedrake/functionspaceimpl.py
pbrubeck: I removed the call to collapse(), and will leave `V[(i,)] == V[i]`

ksagiyam: So now `V[(i,)]` is a non-collapsed indexed space and `V[(i, j)]` is a new collapsed mixed space?

pbrubeck: Not quite. `V[(i, )]` is no longer collapsed, which means that it is a `IndexedProxyFunctionSpace`. But collapsing a `MixedSpace` does nothing, as the new `MixedSpace` subspace has lost any reference to the parent `MixedSpace`. I only required collapsing the subspaces because Fieldsplit builds a new NLVP, and we cannot use proxy spaces for BCs.

ksagiyam: Please read my comment carefully. I understand why we need this, but I have been claiming that we should not have `__getitem__()` have multiple meanings depending on if we pass `int` or `tuple`/`list`. If the former returns something which remembers the parent and the latter returns something which does not, it indicates that this feature should be implemented somewhere else.

JHopeCollins: That's a good point, I hadn't quite appreciated whether the subspace remembers the parent or not. If I understand correctly then we currently have:

1. `subfunction` (and by extension `__getitem__`) returns a `FunctionSpace`/`Function` for a _single_ subcomponent that _remembers_ the parent.
2. `collapse` returns a `FunctionSpace`/`Function` with _all_ components that _forgets_ the parent.

What we want is a utility for returning a `FunctionSpace`/`Function` with _some_ subcomponents that _forgets_ the parent. The fact that it returns a subspace matches with `subfunction`, but the fact that it forgets the parent matches with `collapse`. This means that if we extend either of the existing methods then we will introduce some ambiguity into that method's purpose.

If we are to avoid this ambiguity then Koki is right, we need a new method given how important clarity is with symbolic information (e.g. the `split`->`subfunction` change).

This information should also be added to the docstrings for `subfunctions`/`sub`/`__getitem__`/`collapse`/the new method for all the `FunctionSpace` classes plus the `Function` classes and `Cofunction`.

firedrake/formmanipulation.py
pbrubeck: I'll go back to the old `subspace` method I had for splitting purposes

firedrake/functionspaceimpl.py
pbrubeck: No longer relevant

firedrake/functionspaceimpl.py
JHopeCollins: I think for this PR just doing the `subspace` function is fine, it already improves the `formmanipulation` code nicely. Let's not hold up this PR any more.
We can discuss the best way forward for lifting `subspace` out separately.

.github/workflows/build.yml
```diff
@@ -84,6 +84,7 @@ jobs:
             --install defcon \
             --install gadopt \
             --install asQ \
+            --package-branch ufl pbrubeck/merge-upstream \
```
pbrubeck, line 87:
```suggestion
```

ksagiyam (approved):

Issue #3945, opened 2024-12-21
BUG: Checkpointing unnamed Netgen Mesh


StefanoFochesatto:
**Describe the bug**
When checkpointing an unnamed netgen mesh, the default name is not set correctly and should be "DEFAULT_MESH_NAME" per [docs](https://www.firedrakeproject.org/firedrake.html#firedrake.mesh.DEFAULT_MESH_NAME).

**Steps to Reproduce**
Running this example:
```
from firedrake import *
from netgen.geom2d import SplineGeometry

geo = SplineGeometry()

geo.AddRectangle(p1=(-1, -1),
                 p2=(1, 1),
                 bc="rectangle",
                 leftdomain=1,
                 rightdomain=0)

ngmsh = geo.GenerateMesh(maxh=0.1)

msh = Mesh(ngmsh)

with CheckpointFile("test.h5", 'w') as afile:
    afile.save_mesh(msh)

with CheckpointFile("test.h5", 'r') as afile:
    mesh = afile.load_mesh()
    # mesh = afile.load_mesh('Default') <- this works
```

**Error message**
```
Traceback (most recent call last):
  File "/home/stefano/FiredrakeAdapt/bugExamples/NetgenCheckpointing.py", line 21, in <module>
    mesh = afile.load_mesh()
  File "petsc4py/PETSc/Log.pyx", line 188, in petsc4py.PETSc.Log.EventDecorator.decorator.wrapped_func
  File "petsc4py/PETSc/Log.pyx", line 189, in petsc4py.PETSc.Log.EventDecorator.decorator.wrapped_func
  File "/home/stefano/FiredrakeAdapt/firedrake-dec24/src/firedrake/firedrake/checkpointing.py", line 1044, in load_mesh
    tmesh_name = self._get_mesh_name_topology_name_map()[name]
KeyError: 'firedrake_default'
```

MohamedAlyLoutfy:
@nbouziani Hello, i will work on a fix for this issue here
https://github.com/MohamedAlyLoutfy/firedrake/tree/MohamedAlyLoutfy/netgen-mesh-checkpoint

PR #3951, opened 2025-01-04
Preserve  default mesh name after Netgen mesh creation

MohamedAlyLoutfy:
…pointing errors

# Description
Root Cause:
The function createFromTopology in ngsPETSc.FiredrakeMesh sets the mesh's name to "Default" regardless of the provided custom or default name. This overwrites the expected mesh.name set during the Firedrake Mesh creation.
Fix Summary:
To preserve the expected mesh.name:
    After createFromTopology is called, explicitly reset mesh.name to the desired value (name) to ensure the consistency of the naming conventions.
    This ensures that the custom or default name (DEFAULT_MESH_NAME, i.e., "firedrake_default") is maintained after the Netgen mesh is created.

### DIFF
--- a/firedrake/mesh.py
+++ b/firedrake/mesh.py
@@ -3092,7 +3092,7 @@ def Mesh(meshfile, **kwargs):
                             permutation_name=kwargs.get("permutation_name"),
                             comm=user_comm)
     if netgen and isinstance(meshfile, netgen.libngpy._meshing.Mesh):
-        netgen_firedrake_mesh.createFromTopology(topology, name=name, comm=user_comm)
+        netgen_firedrake_mesh.createFromTopology(topology, name=plex.getName(), comm=user_comm)
         mesh = netgen_firedrake_mesh.firedrakeMesh
     else:
         mesh = make_mesh_from_mesh_topology(topology, name)
--- a/firedrake/mesh.py
+++ b/firedrake/mesh.py
@@ -3064,6 +3064,7 @@ def Mesh(meshfile, **kwargs):
         netgen_flags = kwargs.get("netgen_flags", {"quad": False, "transform": None, "purify_to_tets": False})
         netgen_firedrake_mesh = FiredrakeMesh(meshfile, netgen_flags, user_comm)
         plex = netgen_firedrake_mesh.meshMap.petscPlex
+        plex.setName(name)
     else:
         basename, ext = os.path.splitext(meshfile)
         if ext.lower() in ['.e', '.exo']:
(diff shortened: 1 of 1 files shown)


MohamedAlyLoutfy:
Hello @nbouziani @dham  I have created a fix for this issue [https://github.com/firedrakeproject/firedrake/issues/3945](issue) in this PR, i made it as a draft PR until the CI check passes, I'm waiting approval for the workflows so i can make it ready for review, Thank you.

firedrake/mesh.py
```diff
@@ -3094,6 +3094,7 @@ def Mesh(meshfile, **kwargs):
     if netgen and isinstance(meshfile, netgen.libngpy._meshing.Mesh):
         netgen_firedrake_mesh.createFromTopology(topology, name=plex.getName(), comm=user_comm)
         mesh = netgen_firedrake_mesh.firedrakeMesh
+        mesh.name = name
```
connorjward, line 3097: This code looks a bit funny:
* We create a `mesh` on line 3093 which gets discarded - put inside an `else` block?
* We already pass `name` to `createFromTopology` on line 3095. It seems a bit funny to be setting it twice. Should this fix instead go into `ngsPETSc`?

firedrake/mesh.py
```diff
@@ -3064,6 +3064,7 @@ def Mesh(meshfile, **kwargs):
         netgen_flags = kwargs.get("netgen_flags", {"quad": False, "transform": None, "purify_to_tets": False})
         netgen_firedrake_mesh = FiredrakeMesh(meshfile, netgen_flags, user_comm)
         plex = netgen_firedrake_mesh.meshMap.petscPlex
+        plex.setName(name)
```
connorjward, line 3067:
```suggestion
        plex.setName(_generate_default_mesh_topology_name(name))
```

firedrake/mesh.py
```diff
@@ -3090,10 +3091,11 @@ def Mesh(meshfile, **kwargs):
                             distribution_name=kwargs.get("distribution_name"),
                             permutation_name=kwargs.get("permutation_name"),
                             comm=user_comm)
-    mesh = make_mesh_from_mesh_topology(topology, name)
     if netgen and isinstance(meshfile, netgen.libngpy._meshing.Mesh):
         netgen_firedrake_mesh.createFromTopology(topology, name=plex.getName(), comm=user_comm)
```
connorjward, line 3095:
```suggestion
        netgen_firedrake_mesh.createFromTopology(topology, name=name, comm=user_comm)
```

connorjward (commented):
Thanks. I am still a bit confused about what is going on. Would the following changes work? They would then be consistent with what we do for the non-netgen case.

firedrake/mesh.py
MohamedAlyLoutfy: @connorjward  Yes This change will work too , In the non-Netgen case (line 3098), the mesh name is explicitly passed to make_mesh_from_mesh_topology, ensuring mesh.name is set correctly to name. To maintain consistency, I followed the same approach for Netgen meshes by setting plex.setName(_generate_default_mesh_topology_name(name)) for the topology name and passing the mesh name explicitly in createFromTopology. This keeps the behavior aligned for both Netgen and non-Netgen meshes.

I have updated the code With these changes, both cases now handle mesh and topology names consistently.

connorjward (approved):
Looks good now. Thanks!

connorjward:
@MohamedAlyLoutfy I assume that this is ready to merge and no longer a draft? If so please mark as such.

MohamedAlyLoutfy:
> @MohamedAlyLoutfy I assume that this is ready to merge and no longer a draft? If so please mark as such.

yes, i made it ready for review so it's no more a draft @connorjward

PR #3961, opened 2025-01-09
Fix test for mixed-ness in Function.sub, and fix Cofunction.sub to match

### LINKED ISSUE #3933: BUG: `Function.sub` return type depends on number of sub elements in a `MixedElement`
**Describe the bug**

For a mixed space `Function.sub` is a `CoordinatelessFunction` if there is one sub-element, and a `Function` if there is more than one.

**Steps to Reproduce**
```
import firedrake as fd

mesh = fd.UnitIntervalMesh(1)
space = fd.FunctionSpace(mesh, "DG", 0)
for dim in range(1, 4):
    spaces = fd.FunctionSpace(mesh, fd.MixedElement(*(space.ufl_element() for _ in range(dim))))
    u = fd.Function(spaces, name="u")
    print(f"{type(u.sub(0))=}")
```
leads to
```
type(u.sub(0))=<class 'firedrake.function.CoordinatelessFunction'>
type(u.sub(0))=<class 'firedrake.function.Function'>
type(u.sub(0))=<class 'firedrake.function.Function'>
```

**Expected behavior**
Consistent 'dimension independent' behaviour, independent of the number of sub elements.

**Additional Info**
The loss of the coordinates is unexpected.

JHopeCollins:
Fixes #3933 

1. Fixes the test for mixed-ness to check the UFL element rather than the length - this gets the correct answer for "mixed" spaces with only one component.
2. Updates `Cofunction.sub` in line with `Function.sub`.

### DIFF
--- a/firedrake/cofunction.py
+++ b/firedrake/cofunction.py
@@ -4,6 +4,7 @@ import ufl
 from ufl.form import BaseForm
 from pyop2 import op2, mpi
 from pyadjoint.tape import stop_annotating, annotate_tape, get_working_tape
+from finat.ufl import MixedElement
 import firedrake.assemble
 import firedrake.functionspaceimpl as functionspaceimpl
 from firedrake import utils, vector, ufl_expr
@@ -119,11 +120,14 @@ class Cofunction(ufl.Cofunction, FunctionMixin):
 
     @utils.cached_property
     def _components(self):
-        if self.function_space().value_size == 1:
+        if self.function_space().rank == 0:
             return (self, )
         else:
-            return tuple(type(self)(self.function_space().sub(i), val=op2.DatView(self.dat, i))
-                         for i in range(self.function_space().value_size))
+            if self.dof_dset.cdim == 1:
+                return (type(self)(self.function_space().sub(0), val=self.dat),)
+            else:
+                return tuple(type(self)(self.function_space().sub(i), val=op2.DatView(self.dat, j))
+                             for i, j in enumerate(np.ndindex(self.dof_dset.dim)))
 
     @PETSc.Log.EventDecorator()
     def sub(self, i):
@@ -137,9 +141,9 @@ class Cofunction(ufl.Cofunction, FunctionMixin):
         :func:`~.VectorFunctionSpace` or :func:`~.TensorFunctionSpace`
         this returns a proxy object indexing the ith component of the space,
         suitable for use in boundary condition application."""
-        if len(self.function_space()) == 1:
-            return self._components[i]
-        return self.subfunctions[i]
+        mixed = type(self.function_space().ufl_element()) is MixedElement
+        data = self.subfunctions if mixed else self._components
+        return data[i]
 
     def function_space(self):
         r"""Return the :class:`.FunctionSpace`, or :class:`.MixedFunctionSpace`
@@ -322,6 +326,12 @@ class Cofunction(ufl.Cofunction, FunctionMixin):
         return vector.Vector(self)
 
     @property
+    def cell_set(self):
+        r"""The :class:`pyop2.types.set.Set` of cells for the mesh on which this
+        :class:`Function` is defined."""
+        return self.function_space()._mesh.cell_set
+
+    @property
     def node_set(self):
         r"""A :class:`pyop2.types.set.Set` containing the nodes of this
         :class:`Cofunction`. One or (for rank-1 and 2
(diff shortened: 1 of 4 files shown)


JHopeCollins:
@connorjward @ksagiyam I removed the bounds checking in `sub` for `FunctionSpace`, `Function`, and `Cofunction` because` this check will be done by the `subfunctions` or `_components` tuple anyway. However, this is causing the tests to fail because we expect that `Function.sub` will fail if passed a negative index. Given that negative indices are a standard python feature, and here they are doing the usual python thing, is there any reason to keep this restriction?

firedrake/cofunction.py
```diff
@@ -119,11 +120,14 @@ def split(self):
...
+        if self.function_space().rank == 0:
             return (self, )
         else:
-            return tuple(type(self)(self.function_space().sub(i), val=op2.DatView(self.dat, i))
-                         for i in range(self.function_space().value_size))
+            if self.dof_dset.cdim == 1:
+                return (type(self)(self.function_space().sub(0), val=self.dat),)
+            else:
+                return tuple(type(self)(self.function_space().sub(i), val=op2.DatView(self.dat, j))
+                             for i, j in enumerate(np.ndindex(self.dof_dset.dim)))
```
connorjward, line 130: What is this doing?

connorjward (commented):
LGTM more or less

connorjward:
> @connorjward @ksagiyam I removed the bounds checking in `sub` for `FunctionSpace`, `Function`, and `Cofunction` because this check will be done by the `subfunctions` or `_components` tuple anyway. However, this is causing the tests to fail because we expect that `FunctionSpace.sub` will fail if passed a negative index. Given that negative indices are a standard python feature, and here they are doing the usual python thing, is there any reason to keep this restriction?

I'm happy to permit negative indices and do the Pythonic thing. There is no way that this will break user code to change.

JHopeCollins:
> I'm happy to permit negative indices and do the Pythonic thing. There is no way that this will break user code to change.

Great, I've updated the test so that it checks a negative index that is actually out of bounds.

ksagiyam:
I think those bound checks are there to give clearer error message:
```
"Invalid component {i}, not in [0, {bound})"
```
vs.
```
IndexError: list index out of range
```
so I think we should keep them.

JHopeCollins:
> I think those bound checks are there to give clearer error message:
> 
> ```
> "Invalid component {i}, not in [0, {bound})"
> ```
> 
> vs.
> 
> ```
> IndexError: list index out of range
> ```
> 
> so I think we should keep them.

That's the usual python message for bad indices so personally I'm fine with it. If we _really_ want a slightly more detailed one then we should be catching the `IndexError` and then adding to the message. But if we think it's really that important then shouldn't we also add this message to subfunctions for FunctionSpace, DualSpace, MixedFunctionSpace, Function, Cofunction, and also anywhere else where we expose a tuple to the user?
It just seems quite heavy handed for the problem it's addressing.

connorjward:
> > I think those bound checks are there to give clearer error message:
> > ```
> > "Invalid component {i}, not in [0, {bound})"
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > vs.
> > ```
> > IndexError: list index out of range
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > so I think we should keep them.
> 
> That's the usual python message for bad indices so personally I'm fine with it. If we _really_ want a slightly more detailed one then we should be catching the `IndexError` and then adding to the message. But if we think it's really that important then shouldn't we also add this message to subfunctions for FunctionSpace, DualSpace, MixedFunctionSpace, Function, Cofunction, and also anywhere else where we expose a tuple to the user? It just seems quite heavy handed for the problem it's addressing.

I agree. If we really cared about this we should probably have something like
```py
try:
  return data[i]
except IndexError as e:
  raise FunctionSpaceIndexOutOfBoundsException("...") from e
```
but I don't think it's terribly important.

ksagiyam:
I see. I am fine with that.

firedrake/cofunction.py
JHopeCollins: The logic here was out of date, this just brings it in line with `Function._components`. 
If we have a scalar element (`rank == 0`) then `sub` just returns `self`. If we have a `VectorElement` or `TensorElement` then we return a view into the relevant component. The switch on `dof_dset.cdim` is because `DatView` complains about viewing a non-mixed `Dat`.

connorjward (approved):

PR #3980, opened 2025-01-16
Localise pytools cache to each runner

connorjward:
My theory:

Previously we were sharing the pytools cache between runners ($XDG_CACHE_HOME is in the home directory) and so different runners would be clearing it at different times. This may well be the reason for CI times to get so bad when the runners are oversubscribed.

Potentially relevant for https://github.com/thetisproject/thetis/issues/383

### DIFF
--- a/.github/workflows/build-mac.yml
+++ b/.github/workflows/build-mac.yml
@@ -26,6 +26,7 @@ jobs:
       FIREDRAKE_CI_TESTS: 1  # needed to symlink the checked out branch into the venv
       OMP_NUM_THREADS: 1
       OPENBLAS_NUM_THREADS: 1
+      XDG_CACHE_HOME: ${{ github.workspace }}/.cache
     steps:
       - name: Add homebrew to PATH
         # https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/workflow-commands-for-github-actions#adding-a-system-path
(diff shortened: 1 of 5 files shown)


connorjward:
Actually I don't think that this will help. The runners use containers so do not have access to the same home directory. For reference the container command is
```
 /usr/bin/docker create --name 843ed855a5bf4e97a632d58e0c780734_firedrakeprojectfiredrakeenvlatest_aec86a --label 2b239b --workdir /__w/firedrake/firedrake --network github_network_865a737a1a724d3e84b5f2a06a533d75  -e "HOME=/github/home" -e GITHUB_ACTIONS=true -e CI=true -v "/var/run/docker.sock":"/var/run/docker.sock" -v "/home/github/actions-runner-1a/_work":"/__w" -v "/home/github/actions-runner-1a/externals":"/__e":ro -v "/home/github/actions-runner-1a/_work/_temp":"/__w/_temp" -v "/home/github/actions-runner-1a/_work/_actions":"/__w/_actions" -v "/home/github/actions-runner-1a/_work/_tool":"/__w/_tool" -v "/home/github/actions-runner-1a/_work/_temp/_github_home":"/github/home" -v "/home/github/actions-runner-1a/_work/_temp/_github_workflow":"/github/workflow" --entrypoint "tail" firedrakeproject/firedrake-env:latest "-f" "/dev/null"
```

in particular note `-v "/home/github/actions-runner-1a/_work/_temp/_github_home":"/github/home"` - the home directory the runner sees is already isolated.

PR #3981, opened 2025-01-16
DO NOT MERGE: Test UFL

pbrubeck:
# Description
Test https://github.com/FEniCS/ufl/pull/338


ksagiyam:
We should run this test again just in case.

PR #3989, opened 2025-01-23
Caching improvements

connorjward:
* Disk cache generated code
* Only generate loopy code on one rank and broadcast to others

Thanks to @pbrubeck for flagging this.

### DIFF
--- a/pyop2/global_kernel.py
+++ b/pyop2/global_kernel.py
@@ -12,7 +12,7 @@ from loopy.codegen.result import process_preambles
 from petsc4py import PETSc
 
 from pyop2 import mpi
-from pyop2.caching import parallel_cache, serial_cache
+from pyop2.caching import memory_and_disk_cache, parallel_cache
 from pyop2.compilation import add_profiling_events, load
 from pyop2.configuration import configuration
 from pyop2.datatypes import IntType, as_ctypes
@@ -366,7 +366,7 @@ class GlobalKernel:
     @cached_property
     def code_to_compile(self):
         """Return the C/C++ source code as a string."""
-        return _generate_code_from_global_kernel(self)
+        return _generate_code_from_global_kernel(self, mpi.COMM_SELF)
 
     @cached_property
     def argtypes(self):
@@ -405,20 +405,26 @@ class GlobalKernel:
         return tuple(ldargs)
 
 
-@serial_cache(hashkey=lambda knl: knl.cache_key)
-def _generate_code_from_global_kernel(kernel):
-    with PETSc.Log.Event("GlobalKernel: generate loopy"):
-        wrapper = generate(kernel.builder)
+@memory_and_disk_cache(hashkey=lambda knl, _: knl.cache_key)
+@mpi.collective
+def _generate_code_from_global_kernel(kernel, comm):
+    if comm.rank == 0:
+        with PETSc.Log.Event("GlobalKernel: generate loopy"):
+            wrapper = generate(kernel.builder)
 
-    with PETSc.Log.Event("GlobalKernel: generate device code"):
-        code = lp.generate_code_v2(wrapper)
+        with PETSc.Log.Event("GlobalKernel: generate device code"):
+            code = lp.generate_code_v2(wrapper)
 
-    if kernel.local_kernel.cpp:
-        preamble = "".join(process_preambles(getattr(code, "device_preambles", [])))
-        device_code = "\n\n".join(str(dp.ast) for dp in code.device_programs)
-        return preamble + "\nextern \"C\" {\n" + device_code + "\n}\n"
+        if kernel.local_kernel.cpp:
+            preamble = "".join(process_preambles(getattr(code, "device_preambles", [])))
+            device_code = "\n\n".join(str(dp.ast) for dp in code.device_programs)
+            code = preamble + "\nextern \"C\" {\n" + device_code + "\n}\n"
+        else:
+            code = code.device_code()
+    else:
+        code = None
 
-    return code.device_code()
+    return comm.bcast(code, root=0)
 
 
 @parallel_cache(hashkey=lambda knl, _: knl.cache_key)
@@ -438,8 +444,10 @@ def compile_global_kernel(kernel, comm):
     A ctypes function pointer for the compiled function.
 
     """
+    code = _generate_code_from_global_kernel(kernel, comm)
+
     dll = load(
-        kernel.code_to_compile,
+        code,
         "cpp" if kernel.local_kernel.cpp else "c",
         cppargs=kernel._cppargs,
         ldargs=kernel._ldargs,


pyop2/global_kernel.py
```diff
@@ -366,7 +366,7 @@ def builder(self):
     @cached_property
     def code_to_compile(self):
         """Return the C/C++ source code as a string."""
-        return _generate_code_from_global_kernel(self)
+        return _generate_code_from_global_kernel(self, mpi.COMM_SELF)
```
ksagiyam, line 369: I think you can delete this method.

pyop2/global_kernel.py
```diff
@@ -405,20 +405,26 @@ def _ldargs(self):
...
+            preamble = "".join(process_preambles(getattr(code, "device_preambles", [])))
+            device_code = "\n\n".join(str(dp.ast) for dp in code.device_programs)
+            code = preamble + "\nextern \"C\" {\n" + device_code + "\n}\n"
+        else:
+            code = code.device_code()
+    else:
+        code = None
 
-    return code.device_code()
+    return comm.bcast(code, root=0)
```
ksagiyam, line 427: Can we do `bcast` in `compile_global_kernel`; this way, `_generate_code_from_global_kernel` can remain unaware of `comm`.

ksagiyam (changes requested):

pyop2/global_kernel.py
connorjward: Unsure. I want:
* The resulting code to be cached on disk
* The code generation to be done on only one rank (Pablo gave some evidence that doing otherwise was slow)
* Each rank to have the generated code kept in memory

ksagiyam: I think you are right. We want to do it this way.

pyop2/caching.py
```diff
@@ -440,119 +441,66 @@ class DEFAULT_CACHE(dict):
...
-                    # If this is a new cache or function add it to the list of known caches
-                    if (comm, comm.name, func, local_cache) not in [(c.comm, c.comm_name, c.func, c.cache()) for c in _KNOWN_CACHES]:
-                        # When a comm is freed we do not hold a reference to the cache.
-                        # We attach a finalizer that extracts the stats before the cache
-                        # is deleted.
-                        _KNOWN_CACHES.append(_CacheRecord(next(_CACHE_CIDX), comm, func, local_cache))
-
-                    value = local_cache.get(key, CACHE_MISS)
-
+def parallel_cache(
```
connorjward, line 444: Previously we duplicated this logic to avoid a `comm.allgather`. I have determined that this is *essential* to avoid deadlocks (hits on not all of the ranks will otherwise deadlock). Since both code paths are now the same apart from a handful of `debug()` statements I have removed the fork.

scripts/firedrake-run-split-tests
```diff
@@ -31,7 +31,9 @@ num_procs=$1
 num_jobs=$2
 extra_args=${@:3}
 
-log_file_prefix="pytest_nprocs${num_procs}_job"
+# Set a unique cache directory for each runner to avoid race conditions
+cache_cmd="FIREDRAKE_TSFC_KERNEL_CACHE_DIR=\$VIRTUAL_ENV/.cache/tsfc/job{#} \
+           PYOP2_CACHE_DIR=\$VIRTUAL_ENV/.cache/pyop2/job{#}"
```
connorjward, line 36:
```suggestion
```

scripts/firedrake-run-split-tests
```diff
@@ -53,7 +57,7 @@ set -x
 # * Uses tee to pipe stdout+stderr to both stdout and a log file
 # * Writes pytest's exit code to a file called jobN.errcode (for later inspection)
 parallel --line-buffer --tag \
-    "${pytest_cmd} |& tee ${log_file_prefix}{#}.log; \
+    "${cache_cmd}; ${pytest_cmd} |& tee ${log_file_prefix}{#}.log; \
```
connorjward, line 60:
```suggestion
    "${pytest_cmd} |& tee ${log_file_prefix}{#}.log; \
```

pyop2/caching.py
ksagiyam: I thought we were caching on `comm`. How can we get to the situation where we have cache_hit on some and cache_miss on the others?

connorjward: It happens when we disk cache things, which uses the same code path

connorjward: Though I do not actually understand how, apart from having race conditions, we could actually have a hit on some ranks but not others. Ideas welcome.

firedrake/tsfc_interface.py
```diff
@@ -54,19 +54,23 @@
 
 
 def tsfc_compile_form_hashkey(form, prefix, parameters, interface, diagonal):
-    # Drop prefix as it's only used for naming
-    return default_parallel_hashkey(form.signature(), prefix, parameters, interface, diagonal)
+    return default_parallel_hashkey(
+        form.signature(),
+        prefix,
+        utils.tuplify(parameters),
+        type(interface).__name__,
```
connorjward, line 61: In PCPatch this was not hashable and so we would get different disk hashes on different ranks.

pyop2/caching.py
connorjward: So it turns out that the problem was actually that the *keys* were not consistent between ranks. I have enabled a check for this when `PYOP2_SPMD_STRICT` is enabled (which it now is in CI) and this has thrown up a number of times where we are doing the wrong thing and risking deadlocks.

pyop2/global_kernel.py
connorjward: I have now added a `bcast` kwarg to `parallel_cache` (defaults to `False`) so we can get this behaviour more transparently.

pyop2/caching.py
ksagiyam: That makes sense. I think this is a lot better now. I'm not sure if I understand `PYOP2_SPMD_STRICT` variable. Do you know when we do not have SPMD and that non-SPMD is relevant in `parallel_cache`?

scripts/firedrake-run-split-tests
```diff
@@ -31,7 +31,9 @@ num_procs=$1
 num_jobs=$2
 extra_args=${@:3}
 
-log_file_prefix="pytest_nprocs${num_procs}_job"
+# Set per-job cache directories to avoid race conditions
```
JHopeCollins, line 34:
```suggestion
# Set per-job cache directories to avoid race conditions when multiple jobs attempt to compile the same form.
```

firedrake/tsfc_interface.py
ksagiyam: Hashing `interface` actually amounts to hashing the `dont_split` coefficient tuple, and it might not be straightforward. I am planning to remove `interface` (in favour of directly passing `dont_split`) in my PR, so I think you can for now probably choose to not cache if `interface` was given.

ksagiyam (changes requested):

firedrake/functionspaceimpl.py
```diff
@@ -876,7 +876,7 @@ class RestrictedFunctionSpace(FunctionSpace):
     def __init__(self, function_space, boundary_set=frozenset(), name=None):
         label = ""
         boundary_set_ = []
-        for boundary_domain in boundary_set:
+        for boundary_domain in sorted(boundary_set, key=str):
```
JHopeCollins, line 879:
```suggestion
        # NOTE: boundary_set must be deterministically ordered here to ensure
        # that the label is consistent between ranks.
        for boundary_domain in sorted(boundary_set, key=str):
```

firedrake/functionspaceimpl.py
```diff
@@ -885,6 +885,9 @@ def __init__(self, function_space, boundary_set=frozenset(), name=None):
                 bd, = as_tuple(boundary_domain)
                 boundary_set_.append(bd)
         boundary_set = boundary_set_
+
+        # NOTE: boundary_set must be deterministically ordered here to ensure
+        # consistency between ranks
```
JHopeCollins, line 890:
```suggestion
```

pyop2/caching.py
```diff
@@ -440,119 +441,89 @@ class DEFAULT_CACHE(dict):
...
+
+    Parameters
+    ----------
+    hashkey :
+        Callable taking ``*args`` and ``**kwargs`` and returning a hash.
+    comm_getter :
+        Callable taking ``*args`` and ``**kwargs`` and returning the
+        appropriate communicator.
+    cache_factory :
+        Callable that will build a new cache (if one does not exist).
```
JHopeCollins, line 460:
```suggestion
    cache_factory :
        Callable that will build a new cache (if one does not exist).
        This will be called every time the decorated function is called, and must return an instance
        of the same type every time it is called.
```

JHopeCollins (approved):
Couple of suggestions to make the comments clearer in the future, but seems sensible to me.

It does seem to prevent some of the errors we were seeing in asQ (or at least it prevents non-determinism between ranks from causing errors).

pyop2/caching.py
connorjward: `PYOP2_SPMD_STRICT` basically just adds a couple of checks in places to assert that things behave collectively as expected.

I don't know what you mean about "when we do not have SPMD".

firedrake/tsfc_interface.py
ksagiyam: At least passing `dont_split` suffices for current usage.

Maybe:
```
if interface is None:
    tree = tsfc_compile_form(...)
else:
    tree = original_tsfc_compile_form(...)
```

pyop2/caching.py
ksagiyam: If it is just for checking, I think I'm fine with that.

firedrake/tsfc_interface.py
connorjward: I don't really want to hack something together. What is the correct long-term fix in your opinion? If you have a PR that addresses the nastiness of passing `interface` around then could we get that in first? I would like this merged within the next couple of weeks but it isn't holding anything up.

JHopeCollins: @ksagiyam @connorjward is this something that needs to be changed in this PR? Or is this something that can be updated when `interface` is actually removed?

If it doesn't need to be in this PR, then @connorjward is this ready to go? It would be really good to get this merged, it prevents (/gets around) crashes we're regularly seeing with asQ.

JHopeCollins: Opening an issue is a good idea, harder to lose track of than a comment in code. Although you could add a comment in the code pointing to the relevant issue.

ksagiyam: No. This is not about whether your current solution with `type(interface).__name__` is best or not; it introduces a new bug around what you just fixed. Suppose you make `interface` with:
```
    interface = make_builder(dont_split=(f, ))
``` 
and call `compile_form`. If you then you make another one with:
```
    interface = make_builder(dont_split=(g, ))
```
and call `compile_form`, you will end up grabbing the wrong value from the cache.

connorjward: What has happened here is I have been caught being lazy. The `interface` kwarg is a bad bit of code design and I wanted to get my code to pass the tests without having to think about it too hard.

I have just pushed something that will work.

firedrake/tsfc_interface.py
```diff
@@ -311,3 +313,14 @@ def extract_numbered_coefficients(expr, numbers):
...
+
+def _make_interface_key(interface):
+    if interface:
+        # Passing interface here is a small hack done in patch.py. What
+        # really matters for caching is what is used in the 'dont_split' kwarg.
+        assert isinstance(interface, functools.partial)
+        assert interface.func is KernelBuilder
+        return tuple(map(hash_expr, interface.keywords["dont_split"]))
+    else:
+        return None
```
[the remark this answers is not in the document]
connorjward: Sorry I don't really understand what you mean. Why isn't it useful? It might be overkill but I don't think that it is necessarily wrong.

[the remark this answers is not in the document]
ksagiyam: Given a form with `form.coefficients() = (f, g, h, l)`, say, what distinguishes `dont_split = (f,)` and `dont_split = (g,)` is the index of `f` and `g` in `(f, g, h, l)`; i.e. (0,) and (1,), so we can not get useful cache key by just looking at the individual coefficients.

For instance, if we run:
```
mesh = UnitIntervalMesh(1)
f = Function(V, count=1000)
g = Function(V, count=2000)
print(hash_expr(f))
print(hash_expr(g))
```
we get:
```
877b3c5df999b747710667a087dece19123b8045b21ba174c593b75c3c87ea1168b2fec9f8ce0634834ac3a3aea52b25ad9038474b9ef8319deb63b73f4de2d0
877b3c5df999b747710667a087dece19123b8045b21ba174c593b75c3c87ea1168b2fec9f8ce0634834ac3a3aea52b25ad9038474b9ef8319deb63b73f4de2d0
``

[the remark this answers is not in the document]
connorjward: Ah you're quite right. I will fix that tomorrow.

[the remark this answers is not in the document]
connorjward: I have fixed this.

firedrake/tsfc_interface.py
connorjward: Resolving as this is now a duplicate.

ksagiyam (dismissed):

PR #3991, opened 2025-01-23
Tiny fix: always upload CI logs

connorjward:
Previously would only happen for successful runs.


JHopeCollins (approved):
LGTM, clearly the right thing to do.

PR #3995, opened 2025-01-27
FEniCS-style bcs

pbrubeck:
# Description
With the default DirichletBC approach, the BCs are applied on the initial guess before linearising. This could be problematic, as the intial state with BCs could cause NaNs for some nonlinear problems containing functions that blow up like logarithms.

This PR adds a kwarg to `solve(F == 0, u, bcs, pre_apply_bcs=False)` to delay the imposition of bcs, so that the problem is linearised around the initial guess before imposing the boundary conditions. Internally, we assemble the residual of the bcs: `bc_resid = u - bc_data` on the rows of the residual `Cofunction`, and the rows and columns of the Jacobian remain as those of the identity, to preserve symmetry. This means that the interior rows of residual need an extra term `F -= action(J, bc_resid)` coming from the lifted BC values on each nonlinear update, in a similar way we already do for the BC lifting in the linear case.

A cool implication of this approach for nonlinear problems where the forcing is driven by inhomogeneous boundary conditions (inflow or traction BCs) is that a linesearch method automatically enables some sort of parameter continuation.

We also restore the old behaviour of `zero_bc_nodes=False`, so it no longer errors when we write primal data on a dual `Cofunction` boundary nodes, but `zero_bc_nodes=True` will remain as the default.


firedrake/variational_solver.py
```diff
@@ -52,7 +52,7 @@ class NonlinearVariationalProblem(NonlinearVariationalProblemMixin):
     def __init__(self, F, u, bcs=None, J=None,
                  Jp=None,
                  form_compiler_parameters=None,
-                 is_linear=False, restrict=False):
+                 is_linear=False, restrict=False, pre_apply_bcs=True):
```
ksagiyam, line 55: I think `pre_apply_bcs` should be attached to `NonlinearVariationalSolver` instead of to `NonlinearVariationalProblem`.

ksagiyam (changes requested):

firedrake/variational_solver.py
pbrubeck: Oh I see, restrict actually changes the problem symbolically, but `pre_apply_bcs` also modifies the residual with the BC lifting. Shouldn't these two options be treated equally?

pbrubeck: It makes more sense to attach pre_apply_bcs to the _SNESContext, as the subproblems in multigrid and fieldsplit do not have a proper `NonlinearVariationalSolver` attached to them.

firedrake/linear_solver.py
```diff
@@ -116,36 +118,27 @@ def _rhs(self):
...
+        expr = b - action(self.A.a, u)
+        return u, get_assembler(expr, bcs=self.A.bcs, zero_bc_nodes=False).assemble, b
 
     def _lifted(self, b):
         u, update, blift = self._rhs
         u.dat.zero()
         for bc in self.A.bcs:
             bc.apply(u)
+        blift.assign(b)
         update(tensor=blift)
```
ksagiyam, line 130: `blift` (`b` in `def _rhs()`) is now input as well as output? Am I missing something?

firedrake/assemble.py
```diff
@@ -95,6 +95,9 @@ def assemble(expr, *args, **kwargs):
         `matrix.Matrix`.
     is_base_form_preprocessed : bool
         If `True`, skip preprocessing of the form.
+    current_state : firedrake.function.Function or None
+        If provided and ``zero_bc_nodes == True``, the boundary condition
+        nodes of the output are set to the residual of the boundary conditions.
```
ksagiyam, line 100:
```suggestion
    current_state : firedrake.function.Function or None
        If provided and ``zero_bc_nodes == False``, the boundary condition
        nodes of the output are set to the residual of the boundary conditions.
```

ksagiyam (changes requested):

firedrake/linear_solver.py
pbrubeck: Is this a memory leak? https://github.com/firedrakeproject/firedrake/blob/ecac12bc9763f0e37e62bf5e32d5bff17d6067cd/firedrake/assemble.py#L482

pbrubeck: I have changed the expression for `b - A*x` to the original - A*x but now the update is done by first copying the incoming residual into the output `tensor` and setting `needs_zeroing=False`. This avoids the creation of a new `Cofunction` for the `maxpy`.

ksagiyam: I think `update(tensor=blift)` pollutes BC node values of `blift` (by `-A*u_bc`), so we still need to explicitly set `bc.apply(blift)`, no?

ksagiyam (dismissed):

firedrake/bcs.py
```diff
@@ -654,19 +652,19 @@ def reconstruct(self, field=None, V=None, subu=None, u=None, row_field=None, col
...
         J = self.f
         Vcol = J.arguments()[-1].function_space()
         u = firedrake.Function(Vcol)
-        F = ufl_expr.action(J, u)
         Vrow = self._function_space
         sub_domain = self.sub_domain
-        bcs = tuple(bc._as_nonlinear_variational_problem_arg() for bc in self.bcs)
-        return EquationBC(F == 0, u, sub_domain, bcs=bcs, J=J, V=Vrow)
+        bcs = tuple(bc._as_nonlinear_variational_problem_arg(is_linear=is_linear) for bc in self.bcs)
+        equation = J == ufl.Form([]) if is_linear else ufl_expr.action(J, u) == 0
```
dham, line 666: This is very hard to read. Also please send me this line for inclusion in chapter 4 of my book as a counterexample.

firedrake/linear_solver.py
```diff
@@ -116,36 +118,27 @@ def _rhs(self):
...
-                bc.apply(blift_func)
-            blift.assign(blift_func.riesz_representation(riesz_map="l2"))
-        else:
-            for bc in self.A.bcs:
-                bc.apply(blift)
         # blift is now b - A u_bc, and satisfies the boundary conditions
         return blift
 
     @PETSc.Log.EventDecorator()
     def solve(self, x, b):
```
JHopeCollins, line 135: Please can you add a docstring which spells out the type requirements on `x` and `b`.

firedrake/preconditioners/base.py
```diff
@@ -109,25 +109,21 @@ def get_appctx(pc):
...
-        F = action(op, u)
+        L = 0
         if bcs:
-            bcs = tuple(bc._as_nonlinear_variational_problem_arg() for bc in bcs)
-        nprob = NonlinearVariationalProblem(F, u,
-                                            bcs=bcs,
-                                            J=op,
-                                            form_compiler_parameters=fcp)
-        return _SNESContext(nprob, mat_type, mat_type, old_appctx, options_prefix=options_prefix)
+            bcs = tuple(bc._as_nonlinear_variational_problem_arg(is_linear=True) for bc in bcs)
```
JHopeCollins, line 124: This is inside `PCSNESBase`, so is it correct to have `is_linear=True`?

firedrake/preconditioners/base.py
```diff
@@ -109,25 +109,21 @@ def get_appctx(pc):
...
+        nprob = LinearVariationalProblem(op, L, u, bcs=bcs, form_compiler_parameters=fcp)
```
JHopeCollins, line 125: This is the `PCSNESBase`, is it correct to create a linear VP rather than a nonlinear one, like it does currently?

firedrake/linear_solver.py
```diff
@@ -116,36 +118,27 @@ def _rhs(self):
 
         u = function.Function(self.trial_space)
         b = cofunction.Cofunction(self.test_space.dual())
-        expr = -action(self.A.a, u)
-        return u, get_assembler(expr).assemble, b
+        expr = - action(self.A.a, u)
+        return u, get_assembler(expr, bcs=self.A.bcs, zero_bc_nodes=False, needs_zeroing=False).assemble, b
 
     def _lifted(self, b):
```
dham, line 124: This is too hard to understand.

pbrubeck: I now spitted this into three separate cached properties (`_ulift`, `_blift`, and `_update_rhs`), and added some comments explaining how the BC lifting can be done in a single call to `assemble`.

firedrake/linear_solver.py
pbrubeck: Done

JHopeCollins:
Please don't resolve comments with reviewer questions. The reviewer might still have questions/comments on that topic, and should be the one to decide if it is resolved or not.

firedrake/preconditioners/base.py
JHopeCollins: It looks like this changes from creating a `NonlinearVariationalProblem` with (implicitly) `action(op, u) == 0`, to a `LinearVariationalProblem` with `op == 0`. Is this a correct reading, and if so are the two identical in practice?

If we are assuming that `op` is always a bilinear form, then I think that should be added to the docstring. It isn't an obvious requirement for a `snes_context` given that a `snes` is for nonlinear problems.

pbrubeck: This is mathematically identical (and also at the UFL) to the old behavior, except that now we let LVP create the lifted residual `action(J, u)`. The only difference is that now we also advertise this as a linear problem.

JHopeCollins: Ok thanks. In which case, please can you add a comment to the docstring and I'm happy for this to be resolved.

pbrubeck: As the docstring/comment mentions, we were always assuming that `op` is a bilinear form (hence the `action`) and that this method is only called by nested PC objects.

JHopeCollins: The docstring for `new_snes_ctx` is:
```python
        """ Create a new SNES contex for nested preconditioning
        """
```
I'm not sure which comment you are referring to, but requirements on the arguments should be in the docstring anyway.

firedrake/preconditioners/base.py
```diff
@@ -109,25 +109,21 @@ def get_appctx(pc):
         return get_appctx(pc.getDM()).appctx
 
     @staticmethod
-    def new_snes_ctx(pc, op, bcs, mat_type, fcp=None, options_prefix=None):
+    def new_snes_ctx(pc, op, bcs, mat_type, fcp=None, options_prefix=None, pre_apply_bcs=True):
         """ Create a new SNES contex for nested preconditioning
         """
```
pbrubeck, line 114:
```suggestion
        """Create a new `_SNESContext` for nested (linear) preconditioning
        
        Parameters
        ----------
        pc : PETSc.PC
             The PC object.
        op : ufl.BaseForm
             A bilinear form.
        bcs : DirichletBC[]
             The boundary conditions.
        mat_type : str
             The matrix type for the assembly of ``op``.
        options_prefix : str
             The PETSc options prefix for the new `_SNESContext`.
        pre_apply_bcs : bool
             If True, the ``bcs`` are pre applied on the solution before the solve, 
             otherwise the residual of the ``bcs`` is included in the linear system.
        
        Returns
        -------
        A new `_SNESContext`
        """
```

pbrubeck: @JHopeCollins feel free to add more documentation

JHopeCollins: Great, much better. Thanks @pbrubeck!

dham (approved):

Issue #3998, opened 2025-01-28
BUG: PETSc interface change has broken VertexOnlyMesh


dham:
**Describe the bug**
[This change](https://gitlab.com/petsc/petsc/-/commit/19307e5cf369b208f3c5d721c42c941e418b5101) to PETSc has changed the interface to DMSwarm such that the cell ID field has changed.

**Steps to Reproduce**

Using current Firedrake and PETSc main:
```python3
from firedrake import *
msh = UnitSquareMesh(5,5)
VOM = VertexOnlyMesh(msh, vertexcoords=msh.coordinates.dat.data, redundant=False)
```
Produces:
```console
   3880     # NOTE ensure that swarm.restoreField is called for each field too!
   3881     swarm_coords = swarm.getField("DMSwarmPIC_coor").reshape((num_vertices, gdim))
-> 3882     swarm_parent_cell_nums = swarm.getField("DMSwarm_cellid").ravel()
   3883     field_parent_cell_nums = swarm.getField("parentcellnum").ravel()
   3884     field_reference_coords = swarm.getField("refcoord").reshape((num_vertices, tdim))

petsc4py/PETSc/DMSwarm.pyx in petsc4py.PETSc.DMSwarm.getField()

Error: error code 83
[0] DMSwarmGetField() at /tmp/petsc-src/src/dm/impls/swarm/swarm.c:1663
[0] DMSwarmDataBucketGetDMSwarmDataFieldByName() at /tmp/petsc-src/src/dm/impls/swarm/data_bucket.c:172
[0] Cannot find DMSwarmDataField with name DMSwarm_cellid
```

**Additional behaviour**

It appears that the cell id is no longer stored in the hard-coded name `"DMSwarm_cellid"` but instead the name has to be accessed programatically. It's not clear how to do that from Python (and the interface may not have been exposed).

PR #3999, opened 2025-01-28
Warn unused options

dham:
Cause a warning message to be logged for solver options that are unused.

This needs us to update to PETSc main, which depends on us fixing #3998 .


tests/firedrake/regression/test_solvers_options_prefix.py
```diff
@@ -210,3 +210,19 @@ def test_same_options_prefix_different_solve():
                                              options_prefix="foo_")
     # Unknown KSP type
     assert excinfo.value.ierr == 86
+
+
+def test_unused_options(caplog):
```
dham, line 215:
```suggestion
def test_unused_options(caplog):
    """Check that unused solver options result in a warning in the log."""
```

Issue #4001, opened 2025-01-29
BUG: linear solver fails if fields are initialised in a different order


tommbendall:
**Describe the bug**
Changing the order of declaration of a field leads to a solver failure. This is a Gusto-independent version of one of the several different errors currently breaking Gusto tests that most likely originated with Firedrake PR #3947

For the Gusto errors, see for instance: https://github.com/firedrakeproject/gusto/actions/runs/12864886794/job/36134821475?pr=610

**Steps to Reproduce**

Here is a failing example. Sorry that it's not more minimal, I couldn't find a smaller example that would fail!

This is clearly very odd. I can also make the test pass by making any number of other trivial changes, but changing the point at which the field `N` in the example below is declared seems to be enough. Otherwise the code below is a Gusto-independent version of the thermal shallow water solver which is currently exhibiting these failures.

```
"""
MFE for broken thermal shallow water solver.
"""

from firedrake import (
    IcosahedralSphereMesh, FunctionSpace, MixedFunctionSpace, Function,
    TestFunction, TestFunctions, TrialFunction, TrialFunctions, Constant,
    split, dx, dS, inner, dot, grad, div, jump, avg, VectorSpaceBasis,
    LinearVariationalProblem, LinearVariationalSolver, lhs, rhs, FacetNormal,
    SpatialCoordinate, VectorFunctionSpace, cross, CellNormal
)

# ============================================================================ #
# FLAG TO CONTROL WHETHER THIS TEST PASSES OR FAILS
set_up_normal_earlier = True   # False passes, True fails
# ============================================================================ #

# ---------------------------------------------------------------------------- #
# Set up core objects
# ---------------------------------------------------------------------------- #

degree = 1
radius = Constant(6371220.)
dt = Constant(3000.0)
H = Constant(3e4/9.80616)
g = Constant(9.80616)
Omega = Constant(7.292e-5)
alpha = Constant(0.5)
beta_u = alpha*dt
beta_d = alpha*dt

mesh = IcosahedralSphereMesh(float(radius), refinement_level=3, degree=2)
x = SpatialCoordinate(mesh)

# Function spaces
VDG = FunctionSpace(mesh, "DG", degree)
VHDiv = FunctionSpace(mesh, "BDM", degree+1)
mesh.init_cell_orientations(x)

VCG1 = FunctionSpace(mesh, 'CG', 1)
Vmixed = MixedFunctionSpace((VHDiv, VDG, VDG))
[...]

PR #4003, opened 2025-01-31
slate: fix kernel argument ordering

ksagiyam:
Fix https://github.com/firedrakeproject/firedrake/issues/4001.

Consider a slate expression:
```
T = A + B
```
where:
```
A = 0 * f + g
B =  f + g
```
Without aggressive zero simplification, we had:
```
A.coefficients() = (f, g)
B.coefficients() = (f, g)
```
which lead to T.coefficients() = (f, g, f, g)->(f, g) == B.coefficients().

Current implementation in Slate had happened to be working as B.coefficients() had been the same as T.coefficients().

With aggressive zero simplification (now merged in UFL), we have:
```
A.coefficients() = (g,)
B.coefficients() = (f, g)
```
which lead to T.coefficients() = (g, f, g)->(g, f) != B.coefficients(), and the current Slate breaks.

This PR is to fix this issue.

### DIFF
--- a/firedrake/slate/slac/kernel_builder.py
+++ b/firedrake/slate/slac/kernel_builder.py
@@ -144,26 +144,24 @@ class LocalLoopyKernelBuilder:
 
         # Pick the coefficients associated with a Tensor()/TSFC kernel
         tsfc_coefficients = {tsfc_coefficients[i]: indices for i, indices in kinfo.coefficient_numbers}
-        for c, cinfo in wrapper_coefficients.items():
-            if c in tsfc_coefficients:
-                if isinstance(cinfo, tuple):
-                    if tsfc_coefficients[c]:
-                        ind, = tsfc_coefficients[c]
-                        if ind != 0:
-                            raise ValueError(f"Active indices of non-mixed function must be (0, ), not {tsfc_coefficients[c]}")
-                        kernel_data.append((c, cinfo[0]))
-                else:
-                    for ind, (c_, info) in enumerate(cinfo.items()):
-                        if ind in tsfc_coefficients[c]:
-                            kernel_data.append((c_, info[0]))
+        for c in tsfc_coefficients:
+            cinfo = wrapper_coefficients[c]
+            if isinstance(cinfo, tuple):
+                if tsfc_coefficients[c]:
+                    ind, = tsfc_coefficients[c]
+                    if ind != 0:
+                        raise ValueError(f"Active indices of non-mixed function must be (0, ), not {tsfc_coefficients[c]}")
+                    kernel_data.append((c, cinfo[0]))
+            else:
+                for ind, (c_, info) in enumerate(cinfo.items()):
+                    if ind in tsfc_coefficients[c]:
+                        kernel_data.append((c_, info[0]))
 
         # Pick the constants associated with a Tensor()/TSFC kernel
         tsfc_constants = tuple(tsfc_constants[i] for i in kinfo.constant_numbers)
-        kernel_data.extend([
-            (constant, constant_name)
-            for constant, constant_name in wrapper_constants
-            if constant in tsfc_constants
-        ])
+        wrapper_constants = dict(wrapper_constants)
+        for c in tsfc_constants:
+            kernel_data.append((c, wrapper_constants[c]))
         return kernel_data
 
     def loopify_tsfc_kernel_data(self, kernel_data):
--- a/tests/firedrake/slate/test_slate_hybridization.py
+++ b/tests/firedrake/slate/test_slate_hybridization.py
@@ -597,3 +597,30 @@ def test_slate_hybridization_gusto():
     # ---------------------------------------------------------------------------- #
 
     uD_solver.solve()
+
+
+@pytest.mark.parametrize('counts', [(10001, 10002), (10002, 10001)])
+def test_slate_hybridization_count_safe(counts):
+    g_count, c_count = counts
+    mesh = UnitTriangleMesh()
+    BDM = FunctionSpace(mesh, "BDM", 2)
+    DG = FunctionSpace(mesh, "DG", 1)
+    V = BDM * DG
+    VectorDG = VectorFunctionSpace(mesh, 'DG', 0)
+    u = TrialFunction(V)
+    v = TestFunction(V)
+    g = Function(VectorDG, count=g_count)
+    c = Function(DG, count=c_count)
+    a = (
+        inner(u, v) * dx
+        + inner(g[0] * u[0], v[0]) * dx
+        + inner(c * grad(u[0]), grad(v[0])) * dx
+    )
+    sol = Function(V)
+    solver_parameters = {
+        'mat_type': 'matfree',
+        'ksp_type': 'preonly',
+        'pc_type': 'python',
+        'pc_python_type': 'firedrake.HybridizationPC',
+    }
+    solve(lhs(a) == rhs(a), sol, solver_parameters=solver_parameters)
--- a/tests/firedrake/slate/test_slate_hybridization.py
+++ b/tests/firedrake/slate/test_slate_hybridization.py
@@ -462,3 +462,138 @@ def test_mixed_poisson_approximated_schur_jacobi_prec(setup_poisson):
 
     assert sigma_err < 1e-8
     assert u_err < 1e-8
+
+
+def test_slate_hybridization_gusto():
+
+    # ---------------------------------------------------------------------------- #
+    # Set up core objects
+    # ---------------------------------------------------------------------------- #
+
+    degree = 1
+    radius = Constant(6371220.)
+    dt = Constant(3000.0)
+    H = Constant(3e4/9.80616)
+    g = Constant(9.80616)
+    Omega = Constant(7.292e-5)
+    alpha = Constant(0.5)
+    beta_u = alpha*dt
+    beta_d = alpha*dt
+
+    mesh = IcosahedralSphereMesh(float(radius), refinement_level=0, degree=2)
+    x = SpatialCoordinate(mesh)
+
+    # Function spaces
+    VDG = FunctionSpace(mesh, "DG", degree)
+    VHDiv = FunctionSpace(mesh, "BDM", degree+1)
+    mesh.init_cell_orientations(x)
+
+    VCG1 = FunctionSpace(mesh, 'CG', 1)
+    Vmixed = MixedFunctionSpace((VHDiv, VDG, VDG))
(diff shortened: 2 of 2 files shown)


connorjward (approved):
Makes sense. Thanks for taking this on.

firedrake/slate/slac/kernel_builder.py
```diff
@@ -144,18 +144,18 @@ def collect_tsfc_kernel_data(self, mesh, tsfc_coefficients, tsfc_constants, wrap
...
+            if isinstance(cinfo, tuple):
+                if tsfc_coefficients[c]:
+                    ind, = tsfc_coefficients[c]
+                    if ind != 0:
+                        raise ValueError(f"Active indices of non-mixed function must be (0, ), not {tsfc_coefficients[c]}")
+                    kernel_data.append((c, cinfo[0]))
+            else:
+                for ind, (c_, info) in enumerate(cinfo.items()):
+                    if ind in tsfc_coefficients[c]:
+                        kernel_data.append((c_, info[0]))
```
pbrubeck, line 158: If we need `kernel_data` to respect the ordering of `tsfc_coefficients`, then we could instead initialize it as a `dict.fromkeys(tsfc_coefficients)`, populate it, then pop keys where the value was left as None, and cast back as a list of tuples.
```
kernel_data = dict.fromkeys(tsfc_coefficients)

for ...
   if ...:
      kernel_data[c] = cinfo[0]
   else: 
      for ...
         if ...
   kernel_data[c_] = info[0]

kernel_data = list(kernel_data.items())
return kernel_data
```

pbrubeck: This only makes sense if we do not expect a coefficients appearing more than one time

firedrake/slate/slac/kernel_builder.py
```diff
@@ -144,18 +144,18 @@ def collect_tsfc_kernel_data(self, mesh, tsfc_coefficients, tsfc_constants, wrap
...
 
         # Pick the constants associated with a Tensor()/TSFC kernel
         tsfc_constants = tuple(tsfc_constants[i] for i in kinfo.constant_numbers)
```
pbrubeck, line 161: Should we also worry about the ordering of the constants?

firedrake/slate/slac/kernel_builder.py
ksagiyam: I think the way we currently do is more straightforward.

firedrake/slate/slac/kernel_builder.py
ksagiyam: Yes, we should. Done.

pbrubeck:
This looks good

pbrubeck (dismissed):
Looks good to me.

tests/firedrake/slate/test_slate_hybridization.py
```diff
@@ -462,3 +462,180 @@ def test_mixed_poisson_approximated_schur_jacobi_prec(setup_poisson):
 
     assert sigma_err < 1e-8
     assert u_err < 1e-8
+
+
+def test_slate_hybridization_gusto():
```
connorjward, line 467: This seems like a very large test to be adding to our test suite when the other test we have actually validates the changes. Why is this being included?

jshipton: Is it large in terms of run time or number of lines? It should work on a lower resolution mesh if that helps. I requested this test be added because it's a common solver strategy in Gusto and seems to be fragile to Firedrake changes so it would be great to have in to pick up issues and give us a starting point to work from.

connorjward: I think integrating tests for downstream packages into our CI is a really sensible idea. I think I would just like to do it in a more structured way like:
1. Adding a separate step in our CI that would clone Gusto (and others) and run a handful of their tests. This is what loopy do for us.
2. Adding a `tests/downstream` folder to contain this sort of test

Personally I prefer option 1.

jshipton: Oooh I **love** option 1!

connorjward: That's definitely useful as we want to see whether the proposed changes in the PR break things. The container won't be updated until after merging to `master`.

connorjward: I will have a stab at doing this this week. @jshipton to start with do you have a couple of tests that are both quick to run and cover a lot of Gusto's codebase? Eventually it would be cool to have a `make check` like we do.

JHopeCollins: Are we happy to merge this then? And revisit where downstream-relevant tests should live later?

connorjward: I don't mind. Ultimately I don't think this test should be here so we can either merge it and I will remove it in my later PR, or we remove it from this PR.

jshipton: In my completely unbiased opinion, I think it should go in and be removed later... ;-)

jshipton: @connorjward I will get back to you with a set of tests tomorrow...

jshipton: In the meantime, we'd really love for this PR to go in so that Gusto main works again! Preferably with the test, but take it out if it's a barrier to merging. Thanks so much for your work on this!

tests/firedrake/slate/test_slate_hybridization.py
```diff
@@ -462,3 +462,180 @@ def test_mixed_poisson_approximated_schur_jacobi_prec(setup_poisson):
...
+    radius = Constant(6371220.)
+    dt = Constant(3000.0)
+    H = Constant(3e4/9.80616)
+    g = Constant(9.80616)
+    Omega = Constant(7.292e-5)
+    alpha = Constant(0.5)
+    beta_u = alpha*dt
+    beta_d = alpha*dt
+
+    mesh = IcosahedralSphereMesh(float(radius), refinement_level=3, degree=2)
```
JHopeCollins, line 483:
```suggestion
    mesh = IcosahedralSphereMesh(float(radius), refinement_level=0, degree=2)
```

tests/firedrake/slate/test_slate_hybridization.py
```diff
@@ -462,3 +462,180 @@ def test_mixed_poisson_approximated_schur_jacobi_prec(setup_poisson):
...
+
+    solver_parameters = {
+        'ksp_type': 'preonly',
+        'mat_type': 'matfree',
+        'pc_type': 'python',
+        'pc_python_type': 'firedrake.HybridizationPC',
+        'hybridization': {
+            'ksp_type': 'cg',
+            'pc_type': 'gamg',
+            'ksp_rtol': 1e-8,
```
JHopeCollins, line 547:
```suggestion
            'ksp_rtol': 1e-1,
```

tests/firedrake/slate/test_slate_hybridization.py
```diff
@@ -462,3 +462,180 @@ def test_mixed_poisson_approximated_schur_jacobi_prec(setup_poisson):
...
+            'mg_levels': {
+                'ksp_type': 'chebyshev',
+                'ksp_max_it': 2,
```
JHopeCollins, line 550:
```suggestion
                'ksp_max_it': 1,
```

tests/firedrake/slate/test_slate_hybridization.py
```diff
@@ -462,3 +462,180 @@ def test_mixed_poisson_approximated_schur_jacobi_prec(setup_poisson):
...
+        uD_problem, solver_parameters=solver_parameters, appctx=appctx
+    )
+
+    # Problem for reconstructing buoyancy
+    b_eqn = b_test*(b_trial - b_rhs + 0.5*dt*inner(u_lhsr, grad(b_bar))) * dx
+
+    b_problem = LinearVariationalProblem(
+        lhs(b_eqn), rhs(b_eqn), b_lhs, constant_jacobian=True
+    )
+    b_solver = LinearVariationalSolver(b_problem)
```
JHopeCollins, line 596:
```suggestion
```

tests/firedrake/slate/test_slate_hybridization.py
```diff
@@ -462,3 +462,180 @@ def test_mixed_poisson_approximated_schur_jacobi_prec(setup_poisson):
...
+
+    # Simplest test is to give a right-hand side that is zero
+    x_rhs.assign(1.0)
+
+    # ---------------------------------------------------------------------------- #
+    # Run
+    # ---------------------------------------------------------------------------- #
+
+    uD_solver.solve()
+    b_solver.solve()
```
JHopeCollins, line 614:
```suggestion
```

JHopeCollins (changes requested):
If we leave this test here for now (and I do think we should until downstream testing gets a more formalised solution), then at least we can make it a bit smaller.
a) the solve doesn't need to be accurate, it just needs to run.
b) the second solve to reconstruct buoyancy doesn't test anything that isn't definitely covered elsewhere (it's a fairly simple form).

JHopeCollins:
Approved pending CI passing.
@connorjward Do you know what this error in the checkout action means?
```
Error: File was unable to be removed Error: EACCES: permission denied, rmdir '/__w/firedrake/firedrake/firedrake/.github/ISSUE_TEMPLATE'
```
Can I just try rerunning it or does something need fixing?

tests/firedrake/slate/test_slate_hybridization.py
JHopeCollins: The test size has been reduced a little (see my suggestions below), so for this PR I'm going to mark this resolved and we can revisit how best to test downstream separately.

connorjward:
> Approved pending CI passing. @connorjward Do you know what this error in the checkout action means?
> 
> ```
> Error: File was unable to be removed Error: EACCES: permission denied, rmdir '/__w/firedrake/firedrake/firedrake/.github/ISSUE_TEMPLATE'
> ```
> 
> Can I just try rerunning it or does something need fixing?

I think this is a side effect of me messing about with CI for #4011. I believe I have fixed this and that it shouldn't recur. If you notice this again please let me know.

tests/firedrake/slate/test_slate_hybridization.py
connorjward: https://github.com/firedrakeproject/firedrake/pull/4017

PR #4011, opened 2025-02-05
pip install firedrake

connorjward:
**Die `firedrake-install`! Die! Die! Die!**

## Remaining tasks

- [x] Decide what to do about `make check`

### `pip install firedrake`

* Changes the officially supported installation method for Firedrake to using `pip`.
* Adds a `firedrake-configure` script to facilitate installation.
* `firedrake-install`/`firedrake-update` remain but spit out a deprecation warning.
* Updates the install page to describe the new approach.
* `firedrake-install --doi` will no longer work.

### Docker

* `firedrake-vanilla` and `firedrake-complex` have been renamed to `firedrake-vanilla-default` and `firedrake-vanilla-complex` respectively.
* `firedrake` is now installed directly in the container. It is no longer necessary to activate a virtual environment.
* The `firedrake` repository is now installed in `~/firedrake` instead of `~/firedrake/src/firedrake`. The same applies to other packages like `gusto` and `thetis` inside the `firedrakeproject/firedrake` container.
* The Docker containers now ship OpenMPI instead of MPICH. This means that **running parallel tests alongside serial ones with a single pytest invocation will no longer work**.
* VTK is no longer installed by default into the `firedrake-vanilla` containers. If your package needs it then please add it to your own `pyproject.toml` or run `pip install vtk`.

Please see https://github.com/firedrakeproject/gusto/pull/618 as an example switching a workflow to using the new containers.

## Developer-facing changes

* The `firedrake-env` container no longer contains PETSc, which is instead built in `firedrake-vanilla`.
* CI now rebuilds PETSc every time instead of pre-building it inside the `firedrake-env` container.
* The linting and zenodo canary workflows have been merged into the same workflow file as the Firedrake tests. This means that we will get fewer notifications if (when) things fail.

## Design choices

Unlike `firedrake-install`, `firedrake-configure` is designed to be *very, very dumb*. All it does is take in a specification of a desired package manager (`apt-x86_64`, `apt-aarch64`, `brew-arm64`) and "arch" (`default` or `complex`) and spits out a string of appropriate system dependencies (`--show-system-dependencies`) or PETSc configure options (`--show-petsc-configure-options`). You can also pass `--no-package-manager` if you aren't using a supported system and this will give you a set of PETSc configure options where everything is downloaded.


scripts/firedrake-configure
```diff
@@ -0,0 +1,363 @@
...
+PETSC_AUTODETECT = 333
+
+# Placeholder value to use when we want PETSc to download the package
+PETSC_DOWNLOAD = 666
+
+PETSC_EXTERNAL_PACKAGE_SPECS = {
+    "bison": {
+        LINUX_APT_X86_64: PETSC_AUTODETECT,
+        LINUX_APT_AARCH64: PETSC_AUTODETECT,
+        MACOS_HOMEBREW_ARM64: "/opt/homebrew",
```
dham, line 201: Should be  the output of `brew --prefix` because users can have a nonstandard location.

scripts/firedrake-install
```diff
@@ -16,6 +16,15 @@ from glob import iglob
...
 import importlib
+import warnings
+
+
+warnings.warn(
+    "firedrake-install/firedrake-update is no longer the recommended way to "
+    "install/update Firedrake. Please use pip instead. For instructions on how "
+    "to do this please visit https://www.firedrakeproject.org/download.html.",
+    FutureWarning
+)
```
dham, line 27: Maybe put a 10 second pause in after this warning so that users have some chance of seeing it.

scripts/firedrake-configure
connorjward: For macOS on ARM this is the [only supported location](https://docs.brew.sh/Installation):
> The script installs Homebrew to its default, supported, best prefix (/opt/homebrew for Apple Silicon, /usr/local for macOS Intel and /home/linuxbrew/.linuxbrew for Linux)

Homebrew [say](https://docs.brew.sh/Installation#untar-anywhere-unsupported):
> Technically, you can just extract (or git clone) Homebrew wherever you want. However, you shouldn’t install outside the default, supported, best prefix.

So I think it is extremely unlikely that a user will not have this path.

In general I have been taking a very hard-line approach to limiting what goes in `firedrake-configure`. We only support (and test!) a very restricted set of platforms (Ubuntu 24 on x86 and ARM and MacOS on ARM). We can add niceties like detecting compilers and package manager paths but we won't be able to test the functionality.

If this is unpopular I change this.

scripts/firedrake-install
connorjward: Excellent point!

.github/workflows/build.yml
```diff
@@ -22,171 +21,164 @@ env:
...
-          cd ../firedrake_venv/src/gusto
-          python -m pytest -n 12 -v \
-            integration-tests/balance/test_saturated_balance.py \
-            integration-tests/equations/test_thermal_sw.py \
-            integration-tests/transport/test_embedded_dg_advection.py
-        timeout-minutes: 5
-
-      - name: Publish Test Report
+          . venv/bin/activate
+          git clone --depth 1 https://github.com/firedrakeproject/gusto.git gusto-repo --branch connorjward/pyproject-fixup
```
connorjward, line 149: https://github.com/firedrakeproject/gusto/pull/617

.github/workflows/build.yml
```diff
@@ -209,10 +201,47 @@ jobs:
...
-    if: ${{ (github.ref == 'refs/heads/master') && always() }}
-    needs: [test, docker_tag]
+    # if: always() && (github.ref == 'refs/heads/master')
+    # needs: [test, docker_tag]
     uses: ./.github/workflows/docker.yml
     with:
-      tag: ${{ needs.docker_tag.outputs.tag }}
-      status: ${{ needs.test.result }}
+      tag: pip
+      status: success
```
connorjward, line 209: This will need undoing before merging

docker/Dockerfile.docdeps
```diff
@@ -1,17 +1,10 @@
-# DockerFile for Firedrake with a full set of capabilities and applications installed.
+# Dockerfile for Firedrake plus packages needed to build the documentation
 
-FROM firedrakeproject/firedrake-vanilla:latest
+FROM firedrakeproject/firedrake-vanilla:pip
```
connorjward, line 3:
```suggestion
FROM firedrakeproject/firedrake-vanilla:latest
```

docker/Dockerfile.env
```diff
@@ -1,167 +1,34 @@
...
-                 libtool python3-dev python3-pip python3-tk python3-venv \
-                 python3-requests zlib1g-dev libboost-dev sudo gmsh \
-                 bison flex ninja-build pkg-config \
-                 libocct-ocaf-dev libocct-data-exchange-dev \
-                 swig graphviz \
-                 libcurl4-openssl-dev libxml2-dev \
-    && rm -rf /var/lib/apt/lists/*
 
 # Use a more sane locale
 ENV LC_ALL C.UTF-8
```
connorjward, line 9:
```suggestion
ENV LC_ALL=C.UTF-8
```

docker/Dockerfile.firedrake
```diff
@@ -1,28 +1,45 @@
...
-        --install femlium \
-        --install gadopt \
-        --install gusto \
-        --install icepack \
-        --install irksome \
-        --install thetis \
-        ${EXTRA_INSTALL_FLAGS}"
+# Dockerfile for Firedrake with a full set of capabilities and applications installed.
+
+FROM firedrakeproject/firedrake-vanilla-default:pip
```
connorjward, line 3:
```suggestion
FROM firedrakeproject/firedrake-vanilla-default:latest
```

docker/Dockerfile.vanilla
```diff
@@ -1,23 +1,39 @@
-# DockerFile for a plain Firedrake suitable for testing Firedrake components and applications
+# Dockerfile for a plain Firedrake suitable for testing Firedrake components and applications
 
-FROM firedrakeproject/firedrake-env:latest
+FROM firedrakeproject/firedrake-env:pip
```
connorjward, line 3:
```suggestion
FROM firedrakeproject/firedrake-env:latest
```

docker/Dockerfile.vanilla
```diff
@@ -1,23 +1,39 @@
...
+
+ENV PETSC_DIR=/home/firedrake/petsc PETSC_ARCH=arch-firedrake-$ARCH
+ENV HDF5_MPI=ON
+ENV CC=mpicc CXX=mpicxx
+ENV MPICC=$CC
+
+# Install Firedrake, pass --break-system-packages because we don't want the
+# container to need a venv.
+RUN pip install --break-system-packages --verbose --no-binary h5py --src . \
+        --editable git+https://github.com/firedrakeproject/firedrake.git@connorjward/pip-install#egg=firedrake
```
connorjward, line 39:
```suggestion
        --editable git+https://github.com/firedrakeproject/firedrake.git#egg=firedrake
```

tests/firedrake/regression/test_import_applications.py
connorjward, line 1: Deleting this as I think we are moving towards a much more robust way of checking downstream packages and so this isn't useful any more.

scripts/firedrake-configure
```diff
@@ -0,0 +1,424 @@
...
+    COMPLEX = "complex"
+
+
+ARCH_DEFAULT = FiredrakeArch.DEFAULT
+ARCH_COMPLEX = FiredrakeArch.COMPLEX
+
+
+def main():
+    parser = argparse.ArgumentParser(
+        description="Print out configuration options needed to install Firedrake",
```
dham, line 44: Bit more detail and reference back to website.

scripts/firedrake-configure
```diff
@@ -0,0 +1,424 @@
...
+        "--no-package-manager",
+        action="store_true",
+        required=False,
+        help="Do not attempt to use a system package manager.",
+    )
+    parser.add_argument(
+        "--arch",
+        choices=[arch.value for arch in FiredrakeArch],
+        default=ARCH_DEFAULT,
+        help="The target 'arch' to install.",
```
dham, line 64: Need to say what an arch is.

scripts/firedrake-configure
```diff
@@ -0,0 +1,424 @@
...
+    )
+    cmd_group = parser.add_mutually_exclusive_group(required=True)
+    cmd_group.add_argument(
+        "--show-system-dependencies",
+        "--sysdeps",  # alias
+        action="store_true",
+        help="Print out the system packages Firedrake needs.",
```
dham, line 71:
```suggestion
        help="Print out the system packages Firedrake needs. The user must install these.",
```

docs/source/install.rst
```diff
@@ -0,0 +1,294 @@
...
+
+   $ python3 ../firedrake-configure --show-petsc-configure-options > my_configure_options.txt
+   <edit my_configure_options.txt>
+   $ ./configure $(cat my_configure_options.txt)
+
+
+.. _install_firedrake:
+
+Installing Firedrake
+--------------------
```
dham, line 169: recommend venv at this stage.

docs/source/install.rst
```diff
@@ -0,0 +1,294 @@
...
+   changing the MPI and/or HDF5 libraries in use.
+
+.. note::
+   If you are using an MPI installed into a nonstandard location it may be
+   necessary to set some additional environment variables before installation:
+
+   * ``CC`` and ``MPICC`` to the location of ``mpicc``
+   * ``CXX`` to the location of ``mpicxx``
+   * ``MPI_HOME`` to the base directory of the MPI installation (e.g. ``/usr``
+     or ``/opt/mpich``)
```
dham, line 210: no zenodo except for editable installs. This will be fixed in April/May for release versions of Firedrake.

docs/source/install.rst
```diff
@@ -0,0 +1,294 @@
+.. raw:: latex
```
dham, line 1: redirect from download.

connorjward: Done

docs/source/install.rst
connorjward: done

scripts/firedrake-configure
connorjward: done

scripts/firedrake-configure
connorjward: done

docs/source/install.rst
```diff
@@ -0,0 +1,312 @@
...
+(termed 'ARCH's) that specify how PETSc is configured and which external
+packages are built. The currently supported ARCHs are:
+
+* ``default``: the default installation, suitable for most users
+* ``complex``: an installation where PETSc is configured using complex numbers
+
+The different configurations can be selected by passing the flag ``--arch`` to
+``firedrake-configure``. For example::
+
+  $ python3 firedrake-configure --show-system-packages --arch complex
```
leo-collins, line 71: In `firedrake-configure` the flag `--show-system-packages` is actually `--show-system-dependencies`

connorjward: Ah good catch! I think I prefer `--show-system-packages` so I will that the consistent option

connorjward: Done

docs/source/install.rst
```diff
@@ -0,0 +1,316 @@
...
+      $ python3 -m venv venv-firedrake
+      $ . venv-firedrake/bin/activate
+
+   This is optional but strongly recommended to avoid polluting your system Python
+   environment.
+
+#. Set any necessary environment variables. This can be achieved using
+   ``firedrake-configure``::
+
+     $ export $(python3 firedrake-configure --show-env)
```
JHopeCollins, line 186: Can you add a note to say that this _must_ be run from the directory that the `petsc` directory is in? Otherwise the `PETSC_DIR` environment variable will be incorrect. Not that I think many people would not do this, but it could lead to confusing errors if someone does.

connorjward: Fair point. And it's actually assuming the parent directory. PETSC_DIR is simply CWD/petsc

connorjward: I will also remove the `cd` bits from the install. Then it's less likely that people will be in the wrong place. I will still mention this though.

docs/source/install.rst
```diff
@@ -0,0 +1,316 @@
...
+to be downloaded and compiled from source::
+
+   $ ./configure $(python3 ../firedrake-configure --no-package-manager --show-petsc-configure-options)
+
+For the ``default`` ARCH, running ``firedrake-configure`` with
+``--no-package-manager`` will produce the flags:
+
+.. literalinclude:: petsc_configure_options.txt
+   :language: text
+
```
JHopeCollins, line 147: Is this missing the `make all` call after configuring?

docs/source/install.rst
connorjward: Actually I can't. `./configure` cannot be run from another directory

docs/source/install.rst
JHopeCollins: And possibly a recommendation to run `make check` afterwards to make sure the PETSc installation has worked before moving on

docs/source/install.rst
connorjward: I have added a note on this (that I will push in a while, don't want to restart the workflows)

docs/source/install.rst
connorjward: I don't think I need to say to `make` here because I am specifically talking about modifying the call to `./configure`. I'll mention `make check`.

docs/source/install.rst
JHopeCollins: Sounds good. Just a comment to reinforce why we need each `cd` command. Otherwise I'm sure there will be some smart-arse **cough** not me **cough** that goes "oh I can just run that from where I am, no need to `cd` just to `cd` back."

docs/source/install.rst
JHopeCollins: Don't you need to actually build PETSc after configuring it though?

connorjward: But I do [here](https://github.com/firedrakeproject/firedrake/pull/4011/files/172244b5f80a6e2773676f96d0d77eeef92331a1#diff-3b2a6fca7ca7832ce28a07b57db259a44a6d6587270e3e576f53e32a6553ed05R130)?

docs/source/install.rst
connorjward: I think it's a bit overly prescriptive to say "you MUST change directory or `configure` will not work". If you try it it will immediately fail. I don't want to overload the user with technical details

docs/source/install.rst
JHopeCollins: You are completely correct. I think I had somehow opened up an old version of this file that wasn't showing it in the preview. Ignore this.

docs/source/install.rst
```diff
@@ -0,0 +1,316 @@
...
+   At a minimum this will set the following variables:
+
+   .. code-block:: text
+
+      CC=/path/to/mpicc CXX=/path/to/mpicxx PETSC_DIR=/path/to/petsc PETSC_ARCH=arch-firedrake-{default,complex} HDF5_MPI=ON
+
+#. Install Firedrake::
+
+     $ pip install --no-binary h5py 'git+https://github.com/firedrakeproject/firedrake.git[test]'
```
JHopeCollins, line 208: A note to say that, whereas with `firedrake-install` the default was to have an editable install, this will not give you and editable install and you will have to specify if that is what you want.
Possibly obvious, but the end result for someone following the installation instructions to the letter has changed.

connorjward: Yeah I should mention something here

JHopeCollins: Also that, even if firedrake is installed in editable mode, all the other packages will not be, e.g. pyadjoint, ufl, etc.

I'm assuming that if I want a "full" editable install then I should do `pip install -e ufl pyadjoint ...` (with the properly addressed package location) before `pip install firedrake`?

connorjward: I haven't mentioned getting editable version of subpackages. I guess I should...

The order of install doesn't actually matter. You can install Firedrake first with no issues

JHopeCollins: But wouldn't pip installing firedrake mean that pip would install all of Firedrake's dependencies? So you would then have to uninstall them, then do the editable install. Doable, but an extra step needlessly

JHopeCollins: > I have added a section titled "Developer install" that explains how to do that.

Great, thanks.

connorjward: > But wouldn't pip installing firedrake mean that pip would install all of Firedrake's dependencies? So you would then have to uninstall them, then do the editable install. Doable, but an extra step needlessly

I believe that you have to do it that way around. Otherwise when you install Firedrake it will try and replace the already installed package.

JHopeCollins: Oh, that's not what I'd expect. Is that a pip thing or a firedrake thing? Wouldn't pip look for the dependencies, see they are already satisfied, and move on?

connorjward: But the versions aren't satisfied. pip can't reason about non-versioned packages so it always assumes that they are out-of-date and reinstalls them

connorjward: This is mentioned in the "Developer install" section now.

JHopeCollins: > But the versions aren't satisfied. pip can't reason about non-versioned packages so it always assumes that they are out-of-date and reinstalls them

Gotcha.

scripts/firedrake-configure
```diff
@@ -0,0 +1,436 @@
...
+        "CC": "mpicc",
+        "CXX": "mpicxx",
+        "PETSC_DIR": f"{os.getcwd()}/petsc",
+        "PETSC_ARCH": f"arch-firedrake-{arch.value}",
+        "HDF5_MPI": "ON",
+    }
+
+    # On macOS h5py cannot find the HDF5 library without help
+    if package_manager == MACOS_HOMEBREW_ARM64:
+        vars["HDF5_DIR"] = "/opt/homebrew"
```
JHopeCollins, line 415:
```suggestion
    if package_manager == MACOS_HOMEBREW_ARM64:
        vars["HDF5_DIR"] = "/opt/homebrew"
    elif package_manager is None:
        vars["HDF5_DIR"] = vars["PETSC_DIR"]+"/"+vars["PETSC_ARCH"]
```
If `--no-package-manager` is used then PETSc will be told to download hdf5, and `HDF5_DIR` must be set so `h5py` knows where to link against.

JHopeCollins (changes requested):

scripts/firedrake-configure
connorjward: Good spot. Done.

lrtfm (changes requested):

docs/source/install.rst
```diff
@@ -0,0 +1,350 @@
...
+   $ git clone https://github.com/firedrakeproject/petsc.git
+   $ cd petsc
+
+#. Run PETSc ``configure``, passing in the flags generated by ``firedrake-configure``::
+
+   $ ./configure $(python3 ../firedrake-configure --show-petsc-configure-options)
+
+#. Compile PETSc::
+
+   $ make
```
dham, line 130: Need to tell the user to set PETSC_DIR and PETSC_ARCH?

scripts/firedrake-configure
```diff
@@ -0,0 +1,436 @@
...
+
+COMMON_PETSC_CONFIGURE_OPTIONS = (
+    "--with-c2html=0",
+    "--with-debugging=0",
+    "--with-fortran-bindings=0",
+    "--with-shared-libraries=1",
+    "--with-strict-petscerrorcode",
+    "--COPTFLAGS=-O3 -march=native -mtune=native",
+    "--CXXOPTFLAGS=-O3 -march=native -mtune=native",
+    "--FOPTFLAGS=-O3 -march=native -mtune=native",
```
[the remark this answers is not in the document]
JHopeCollins: Good spot. I just realised that the configuration for a petsc I built with `firedrake-configure --petscconf` only has "COPTFLAGS=-O3", and then has "-march=native" and "-mtune=native" as additional configure options, not part of COPTFLAGS (and the same for the other compilers).
Using backticks instead of `$(...)` also didn't fix things.

`xargs` works but it would be nicer if we can get something simpler. Strings are a pain...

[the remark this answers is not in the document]
connorjward: Great spot thanks. I did think about this before but I only tried double quotes and I then erroneously assumed that the previous solution must do the right thing.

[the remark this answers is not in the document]
connorjward: It appears like single quotes do not do the right thing: https://github.com/firedrakeproject/firedrake/actions/runs/13521948368/job/37783009674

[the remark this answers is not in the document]
lrtfm: Due to Bash’s [word splitting](https://www.gnu.org/software/bash/manual/html_node/Word-Splitting.html), `COPTFLAGS='-O3` is treated as a single entity, which causes the error. `xargs`, though a little complex, is a solution to this:
```
python3 ../firedrake-repo/scripts/firedrake-configure --arch default \
    --show-petsc-configure-options | \
    xargs -L1 ./configure --with-make-np=12
```

Additionally, I found that using `-march=native` in `--FOPTFLAGS=-O3 -march=native -mtune=native` causes an error on macOS. Removing `-march=native` from `FOPTFLAGS` resolves the issue.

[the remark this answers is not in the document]
lrtfm: Another option is to use eval:
```
eval ./configure \
    $(python3 ../firedrake-repo/scripts/firedrake-configure --arch default --show-petsc-configure-options) \
    --with-make-np=12
```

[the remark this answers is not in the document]
JHopeCollins: We should really really really avoid `eval` because it will execute arbitrary code so is potentially a huge security risk. It's only recommended to use if absolutely necessary, and even then only in a controlled environment.

I know it's still going through PETSc configure but it's a massive hammer to use to fix this problem. If something nefarious manages to get into `firedrake-configure` then we would potentially be compromising both the CI runners and user's machines.

I'll have a look at alternatives, but the pain of asking users to use xargs is probably preferable to using `eval`.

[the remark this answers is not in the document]
connorjward: Genuine question: how is it different to
```
$ sudo $(anything at all)
```
which is exactly what we are already asking people to do?

Happy to switch to xargs if that is better

scripts/firedrake-configure
```diff
@@ -0,0 +1,433 @@
...
+
+COMMON_PETSC_CONFIGURE_OPTIONS = (
+    "--with-c2html=0",
+    "--with-debugging=0",
+    "--with-fortran-bindings=0",
+    "--with-shared-libraries=1",
+    "--with-strict-petscerrorcode",
+    "--COPTFLAGS='-O3 -march=native -mtune=native'",
+    "--CXXOPTFLAGS='-O3 -march=native -mtune=native'",
+    "--FOPTFLAGS='-O3 -mtune=native'",
```
JHopeCollins, line 219: Please can we only remove `-march=native` on macOS? MUMPS is Fortran so this is potentially an unecessary performance regression for our default solver for the majority of users.

Even better would be only removing `-march=native` on the apple clang compilers that don't recognise it, I think the newer apple clang has restored it. But I wouldn't hold up the PR for working out how to do that robustly (although I am biased as a non-apple user).

JHopeCollins (changes requested):

scripts/firedrake-configure
connorjward: Will do

docs/source/install.rst
JHopeCollins: PETSc will print out a line like:
```
Configure stage complete. Now build PETSc libraries with:
   make PETSC_DIR=/path/to/petsc PETSC_ARCH=arch-firedrake-default all
```
In our installation docs could we say something along the lines of:
"PETSc will provide the required command to finish building the libraries, run this before moving on to the next step. The command will look something like ..."
and include the `configure` output above.

scripts/firedrake-configure
JHopeCollins: Thanks!

scripts/firedrake-configure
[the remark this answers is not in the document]
connorjward: Fine I'll give xargs a go

docs/source/install.rst
connorjward: will do

connorjward: done

scripts/firedrake-configure
connorjward: done

.github/workflows/docker_build.yml
```diff
@@ -28,9 +28,8 @@ on:
         description: 'Path to the Dockerfile required to build image'
         required: true
         type: string
-      extra-install-flags:
-        description: 'Extra args to pass when building the Dockerfile'
-        default: ''
+      arch:
+        description: 'Firedrake arch to build'
```
connorjward, line 32:
```suggestion
```

.github/workflows/docker_build.yml
```diff
@@ -76,7 +75,8 @@ jobs:
         with:
           platforms: ${{ inputs.platform }}
           file: ${{ inputs.dockerfile }}
-          build-args: EXTRA_INSTALL_FLAGS=${{ inputs.extra-install-flags }}
+          build-args: |
+            ARCH=${{ inputs.arch }}
```
connorjward, line 79:
```suggestion
```

.github/workflows/docs.yml
```diff
@@ -18,55 +18,56 @@ concurrency:
...
     container:
-      image: firedrakeproject/firedrake-docdeps:latest
-      # Github hosted runners require running as root user:
-      # https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners/about-github-hosted-runners#docker-container-filesystem
-      options: --user root
-      volumes:
-        - ${{ github.workspace }}:/home/firedrake/output
-    # Steps represent a sequence of tasks that will be executed as
-    # part of the jobs
+      image: firedrakeproject/firedrake-docdeps:pip
```
connorjward, line 23:
```suggestion
      image: firedrakeproject/firedrake-docdeps:latest
```

docker/Dockerfile.docdeps
```diff
@@ -1,17 +1,7 @@
-# DockerFile for Firedrake with a full set of capabilities and applications installed.
+# Dockerfile for Firedrake plus packages needed to build the documentation
 
-FROM firedrakeproject/firedrake-vanilla:latest
+FROM firedrakeproject/firedrake-vanilla-default:pip
```
connorjward, line 3:
```suggestion
FROM firedrakeproject/firedrake-vanilla-default:latest
```

docker/Dockerfile.vanilla
```diff
@@ -1,23 +1,43 @@
...
+FROM firedrakeproject/firedrake-env:pip
 
-USER firedrake
-WORKDIR /home/firedrake
+# Set '-o pipefail' to avoid linter error (https://github.com/hadolint/hadolint/wiki/DL4006)
+SHELL ["/bin/bash", "-o", "pipefail", "-c"]
 
-ENV PETSC_ARCH default
+# Firedrake arch to build
+ARG ARCH="default"
```
connorjward, line 10:
```suggestion
```

docker/Dockerfile.vanilla
```diff
@@ -1,23 +1,43 @@
...
-    --mpicc=$MPICH_DIR/mpicc \
-    --mpicxx=$MPICH_DIR/mpicxx \
-    --mpif90=$MPICH_DIR/mpif90 \
-    --mpiexec=$MPICH_DIR/mpiexec \
-    --mpihome=$MPICH_DIR/.. \
-    ${EXTRA_INSTALL_FLAGS}"
+# Install system dependencies
+RUN sudo apt-get update \
+    && eval sudo apt-get -y install \
+        $(python3 ./firedrake-configure --arch $ARCH --show-system-packages) \
```
connorjward, line 17:
```suggestion
    && sudo apt-get -y install \
        $(python3 ./firedrake-configure --arch $ARCH --show-system-packages) \
```

pyproject.toml
```diff
@@ -59,47 +59,67 @@ Repository = "https://github.com/firedrakeproject/firedrake"
...
-  "numpy",
+  "libsupermesh @ git+https://github.com/firedrakeproject/libsupermesh.git",
   "mpi4py>3; python_version >= '3.13'",
   "mpi4py; python_version < '3.13'",
-  "petsc4py",
+  "numpy",
+  "pkgconfig",
+  "pybind11",
+  "setuptools>61.2",
+  "petsc4py==3.22.2",
```
connorjward, line 121:
```suggestion
  "petsc4py",
```

Ig-dolci:
I just tried installing petsc using the most recent `firedrake-configure` file. Follow the steps

```
curl -O https://raw.githubusercontent.com/firedrakeproject/firedrake/connorjward/pip-install/scripts/firedrake-configure
brew update
brew upgrade
brew install $(python3 firedrake-configure --show-system-packages)
git clone https://github.com/firedrakeproject/petsc.git
cd petsc
./configure $(python3 ../firedrake-configure --show-petsc-configure-options)
```
Here is the attached `configure.log` file: 
[configure.log](https://github.com/user-attachments/files/18984274/configure.log)

JHopeCollins:
> I just tried installing petsc using the most recent `firedrake-configure` file. Follow the steps
> 
> ```
> curl -O https://raw.githubusercontent.com/firedrakeproject/firedrake/connorjward/pip-install/scripts/firedrake-configure
> brew update
> brew upgrade
> brew install $(python3 firedrake-configure --show-system-packages)
> git clone https://github.com/firedrakeproject/petsc.git
> cd petsc
> ./configure $(python3 ../firedrake-configure --show-petsc-configure-options)
> ```
> 
> Here is the attached `configure.log` file: [configure.log](https://github.com/user-attachments/files/18984274/configure.log)

It should be:
```
python3 ../firedrake-configure --show-petsc-configure-options | xargs -L1 ./configure
```
See the discussion [here](https://github.com/firedrakeproject/firedrake/pull/4011#discussion_r1966517096) and the updated installation doc page

PR #4012, opened 2025-02-05
Make LinearSolver a subclass of LinearVariationalSolver

pbrubeck:
# Description

Provides enhacements for `LinearVariationalSolver` to support more generic `MatrixBase` operators, and makes `LinearSolver` a subclass of `LinearVariationalSolver`.

As a side benefit, this PR also adds a new feature, `LinearVariationalSolver` can now handle assembled operators, including support for composable PETSc solvers. Here we add some tests for the assembled `LinearVariationalSolver` interface `solve(assemble(a, bcs=bcs) == L, u)` that reproduces the behavior of the `LinearSolver` interface `solve(assemble(a, bcs=bcs), u, L)`.

Also adds syntax sugar to SLATE to seamlessly cast UFL objects into the appropriate SLATE `TensorBase` class when applying arithmetic operations. Combinations of SLATE Tensors and UFL are required for the symbolic residual calculated by `LinearVariationalSolver` (`LinearSolver` did it numerically).


connorjward (commented):
Rather than doing this would it not be possible to deprecate `LinearSolver` in favour of `LinearVariationalSolver`? I don't think wrapping one in the other makes sense in terms of code design.

pbrubeck:
There's still good use for this class, especially the as it enables the user to provide multiple right-hand side and solution pairs. Another good reason is that this gives the user another way of preventing re-assembling the Jacobian.

pbrubeck:
This PR is also paving the way for composing Fieldsplit and other solvers to solve systems with pre-assembled matrices and also more general `BaseForms`.

connorjward:
> This PR is also paving the way for composing Fieldsplit and other solvers to solve systems with pre-assembled matrices and also more general BaseForms.

But why is `LinearVariationalSolver` not suitable for those cases? We can now represent assembled matrices in UFL so can't the code paths be unified? I think @JHopeCollins has had some success with this sort of thing using external operators (when we were discussing taping for `LinearSolver`).

> gives the user another way of preventing the Jacobian of being re-assembled.

Surely multiple methods to achieve the same result suggests that things should be combined?

> it enables the user to provide multiple right-hand side and solution pairs

Could we extend `LinearVariationalSolver` to support this?

connorjward:
> Basically the PR consists mostly of enhancements to LinearVariationalSolver.

Ah I see. Could we therefore think of this PR as a step along the way to fully removing `LinearSolver` (at a later date)? If so then I am completely on board.

firedrake/linear_solver.py
```diff
@@ -149,45 +74,23 @@ def solve(self, x, b):
         b : firedrake.cofunction.Cofunction or firedrake.vector.Vector
             A Cofunction or Vector with the right-hand side of the linear system.
         """
-        if not isinstance(x, (function.Function, vector.Vector)):
+        if not isinstance(x, (Function, Vector)):
             raise TypeError(f"Provided solution is a '{type(x).__name__}', not a Function or Vector")
```
connorjward, line 78:
```suggestion
        if not isinstance(x, Function):
            raise TypeError(f"Provided solution is a '{type(x).__name__}', not a Function")
```

connorjward (dismissed):
I think I have only noted small things.

Also, does this PR add any new functionality? If so it would be great to test it.

firedrake/linear_solver.py
connorjward: `Vector` is fully deprecated.

firedrake/projection.py
```diff
@@ -163,6 +163,8 @@ def __init__(
             is_variable_layers = True
         self.use_slate_for_inverse = (use_slate_for_inverse and is_dg and not is_variable_layers
                                       and (not complex_mode or SLATE_SUPPORTS_COMPLEX))
+        if form_compiler_parameters is not None:
```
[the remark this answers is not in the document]
pbrubeck: Yes, this just turned out to be an issue. The qudrature_degree kwarg is a recent feature, we always could specify the `quadrature_degree` in the `form_compiler_paramters`.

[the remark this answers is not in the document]
pbrubeck: For some complicated reason, the coexistence of these two mechanisms is only breaking tests now that `Projector` uses a `LinearVariationalSolver`

Ig-dolci:
> # Description
> 
> Provides enhacements for `LinearVariationalSolver` to support more generic `MatrixBase` operators, and makes `LinearSolver` a wrapper for `LinearVariationalSolver`,

Just a note to pay attention for adjoints: I am more concerned about how the adjoint-based gradients will work with `LinearVariationalSolver` supporting `MatrixBase`. Currently, we have `LinearVariationalSolver` that works perfectly to have the adjoint equations and to obtain and solve an adjoint-based gradient form. Considering this additional feature, I do not see how adjoint-based gradient will work for general cases.

firedrake/projection.py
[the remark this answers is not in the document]
connorjward: Hmm. Looking at the code now it seems like passing `quadrature_degree` as its own kwarg wasn't the right thing to do and instead it should always be passed inside the form compiler parameters.

It would be sensible to remove all references to it in the code apart from at the top level where we could have
```py
if quadrature_degree is not None:
    warnings.warn("Passing 'quadrature_degree' to XXX is deprecated, please instead set inside the form compiler parameters", FutureWarning)
    if "quadrature_degree" in form_compiler_parameters:
        raise ValueError("Cannot pass quadrature degree twice")
    form_compiler_parameters["quadrature_degree"] = quadrature_degree
```

Assuming that you agree would you mind having a go at that?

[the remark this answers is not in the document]
pbrubeck: I did give this a try. It did not seem not that straightforward. Let's open a new PR for that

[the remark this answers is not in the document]
connorjward: OK. Can you at least open an issue and assign me to it?

pbrubeck:
> Just a note to pay attention for adjoints: I am more concerned about how the adjoint-based gradients will work with `LinearVariationalSolver` supporting `MatrixBase`. Currently, we have `LinearVariationalSolver` that works perfectly to have the adjoint equations and to obtain and solve an adjoint-based gradient form. Considering this additional feature, I do not see how adjoint-based gradient will work for general cases.

@Ig-dolci We support variational problems defined purely in UFL, thus we get differentiability and adjoints. The enhacements are to do with Fieldsplit.

firedrake/linear_solver.py
```diff
@@ -1,27 +1,18 @@
...
-import firedrake.vector as vector
-import firedrake.matrix as matrix
-import firedrake.solving_utils as solving_utils
-from firedrake import dmhooks
-from firedrake.petsc import PETSc, OptionsManager, flatten_parameters
-from firedrake.utils import cached_property
-from firedrake.ufl_expr import action
+from firedrake.function import Function
+from firedrake.cofunction import Cofunction
+from firedrake.vector import Vector
```
pbrubeck, line 3:
```suggestion
```

firedrake/projection.py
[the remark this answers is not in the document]
connorjward: Done [here](https://github.com/firedrakeproject/firedrake/issues/4016). For this PR can you just check that we don't specify `quadrature_degree` twice and throw an error if someone does?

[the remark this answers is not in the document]
pbrubeck: Ok

firedrake/variational_solver.py
```diff
@@ -374,7 +374,12 @@ def __init__(self, a, L, u, bcs=None, aP=None,
                 raise TypeError("Provided RHS is a '%s', not a Form or Slate Tensor" % type(L).__name__)
             if len(L.arguments()) != 1 and not L.empty():
                 raise ValueError("Provided RHS is not a linear form")
-            F = ufl_expr.action(J, u) - L
+            A = J
+            if isinstance(A, MatrixBase) and A.has_bcs:
+                A = A.a
+            if isinstance(A, slate.slate.TensorBase):
+                L = slate.slate.as_slate(L)
```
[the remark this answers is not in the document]
pbrubeck: Done

firedrake/formmanipulation.py
```diff
@@ -147,6 +148,29 @@ def cofunction(self, o):
...
+                asplit = a
+                for fset in V.dof_dset.field_ises:
+                    iset = iset.expand(fset)
+
+            ises.append(iset)
+            args.append(asplit)
+
+        submat = o.petscmat.createSubMatrix(*ises)
+        bcs = ()
+        return AssembledMatrix(tuple(args), bcs, submat)
```
JHopeCollins, line 172: Why enforce this to be `AssembledMatrix`? Why not `type(o)` so that the matrix type is maintained?

JHopeCollins: `ImplicitMatrixContext.createSubMatrix` already does that for you.

dham (changes requested):
Let's split out the projection change and then land the rest of this afterwards.

firedrake/linear_solver.py
```diff
@@ -1,34 +1,24 @@
...
-    DEFAULT_KSP_PARAMETERS = solving_utils.DEFAULT_KSP_PARAMETERS
+class LinearSolver(LinearVariationalSolver):
 
     @PETSc.Log.EventDecorator()
-    def __init__(self, A, *, P=None, solver_parameters=None,
-                 nullspace=None, transpose_nullspace=None,
-                 near_nullspace=None, options_prefix=None,
-                 pre_apply_bcs=True):
+    def __init__(self, A, *, P=None, **kwargs):
         """A linear solver for assembled systems (Ax = b).
```
JHopeCollins, line 15: I think it would be useful to explicitly state that the matrix cannot change (`constant_jacobian=True`), because this is a different default from `LinearVariationalSolver` (even if it doesn't change the default for `LinearSolver`).
```suggestion
        """A linear solver for assembled systems (Ax = b) with constant A.
```

firedrake/linear_solver.py
```diff
@@ -41,153 +31,59 @@ def __init__(self, A, *, P=None, solver_parameters=None,
...
-    def _lifted(self, b):
-        self._ulift.dat.zero()
-        for bc in self.A.bcs:
-            bc.apply(self._ulift)
-        # Copy the input to the internal Cofunction so we do not modify the input
-        self._blift.assign(b)
-        self._update_rhs(tensor=self._blift)
-        # self._blift is now b - A u_bc, and satisfies the boundary conditions
-        return self._blift
+        self.ksp = self.snes.ksp
```
JHopeCollins, line 61: Is `self.ksp` used anywhere?

firedrake/linear_solver.py
```diff
@@ -41,153 +31,59 @@ def __init__(self, A, *, P=None, solver_parameters=None,
...
         """Solve the linear system with RHS ``b`` and store the solution in ``x``.
 
         Parameters
         ----------
-        x : firedrake.function.Function or firedrake.vector.Vector
-            Existing Function or Vector to place the solution to the linear system in.
-        b : firedrake.cofunction.Cofunction or firedrake.vector.Vector
-            A Cofunction or Vector with the right-hand side of the linear system.
+        x : firedrake.function.Function
+            Existing Function to place the solution to the linear system in.
```
JHopeCollins, line 70: Just to be consistent with the docstring for `b`. I don't think it needs to be stated that `x` needs to exist!
```suggestion
            A Function to place the solution to the linear system in.
```

firedrake/linear_solver.py
```diff
@@ -41,153 +31,59 @@ def __init__(self, A, *, P=None, solver_parameters=None,
...
-        solver_parameters = flatten_parameters(solver_parameters or {})
-        solver_parameters = solving_utils.set_defaults(solver_parameters,
-                                                       A.arguments(),
-                                                       ksp_defaults=self.DEFAULT_KSP_PARAMETERS)
+        test, trial = A.arguments()
+        self.x = Function(trial.function_space())
+        self.b = Cofunction(test.function_space().dual())
+
+        problem = LinearVariationalProblem(A, self.b, self.x, aP=P,
+                                           form_compiler_parameters=A.fc_params,
```
JHopeCollins, line 52: Let's help to anyone reading the code and be a bit more explicit - this is what is used essentially everywhere except the `ImplicitMatrixContext`.
```suggestion
                                           form_compiler_parameters=A.form_compiler_parameters,
```

firedrake/matrix.py
```diff
@@ -20,8 +20,9 @@ class MatrixBase(ufl.Matrix):
         :class:`MatrixBase`.  May be `None` if there are no boundary
         conditions to apply.
     :arg mat_type: matrix type of assembled matrix, or 'matfree' for matrix-free
+    :kwarg fc_params: a dict of form compiler parameters of this matrix
```
JHopeCollins, line 23:
```suggestion
    :kwarg form_compiler_parameters: a dict of form compiler parameters of this matrix
```

firedrake/matrix.py
```diff
@@ -20,8 +20,9 @@ class MatrixBase(ufl.Matrix):
...
     """
-    def __init__(self, a, bcs, mat_type):
+    def __init__(self, a, bcs, mat_type, fc_params=None):
```
JHopeCollins, line 25:
```suggestion
    def __init__(self, a, bcs, mat_type, form_compiler_parameters=None):
```

firedrake/matrix.py
```diff
@@ -52,6 +53,7 @@ def __init__(self, a, bcs, mat_type):
 
         Matrix type used in the assembly of the PETSc matrix: 'aij', 'baij', 'dense' or 'nest',
         or 'matfree' for matrix-free."""
+        self.fc_params = fc_params
```
JHopeCollins, line 56:
```suggestion
        self.form_compiler_parameters = form_compiler_parameters
```

firedrake/matrix.py
```diff
@@ -109,6 +111,8 @@ class Matrix(MatrixBase):
 
     :arg mat_type: matrix type of assembled matrix.
 
+    :kwarg fc_params: a dict of form compiler parameters for this matrix.
```
JHopeCollins, line 114:
```suggestion
    :kwarg form_compiler_parameters: a dict of form compiler parameters for this matrix.
```

firedrake/matrix.py
```diff
@@ -121,8 +125,9 @@ class Matrix(MatrixBase):
     """
 
     def __init__(self, a, bcs, mat_type, *args, **kwargs):
-        # sets self._a, self._bcs, and self._mat_type
-        MatrixBase.__init__(self, a, bcs, mat_type)
+        # sets self.a, self.bcs, self.mat_type, and self.fc_params
+        fc_params = kwargs.pop("fc_params", None)
+        MatrixBase.__init__(self, a, bcs, mat_type, fc_params=fc_params)
```
JHopeCollins, line 130:
```suggestion
        fc_params = kwargs.pop("form_compiler_parameters", None)
        MatrixBase.__init__(self, a, bcs, mat_type, form_compiler_parameters=fc_params)
```

firedrake/matrix.py
```diff
@@ -146,6 +151,7 @@ class ImplicitMatrix(MatrixBase):
         :class:`Matrix`.  May be `None` if there are no boundary
         conditions to apply.
 
+    :kwarg fc_params: a dict of form compiler parameters for this matrix.
```
JHopeCollins, line 154:
```suggestion
    :kwarg form_compiler_parameters: a dict of form compiler parameters for this matrix.
```

firedrake/matrix.py
```diff
@@ -155,8 +161,9 @@ class ImplicitMatrix(MatrixBase):
 
     """
     def __init__(self, a, bcs, *args, **kwargs):
-        # sets self._a, self._bcs, and self._mat_type
-        super(ImplicitMatrix, self).__init__(a, bcs, "matfree")
+        # sets self.a, self.bcs, self.mat_type, and self.fc_params
+        fc_params = kwargs["fc_params"]
```
JHopeCollins, line 165:
```suggestion
        fc_params = kwargs["form_compiler_parameters"]
```

tests/firedrake/regression/test_projection.py
```diff
@@ -213,9 +213,7 @@ def test_projector(mat_type, quadrature_degree):
                   form_compiler_parameters=fc_params)
     P.project()
 
-    if mat_type == "matfree":
-        ctx = P.A.petscmat.getPythonContext()
-        assert ctx.fc_params.get("quadrature_degree") == quadrature_degree
+    assert P.A.fc_params.get("quadrature_degree") == quadrature_degree
```
JHopeCollins, line 216:
```suggestion
    assert P.A.form_compiler_parameters.get("quadrature_degree") == quadrature_degree
```

tests/firedrake/regression/test_solving_interface.py
```diff
@@ -262,6 +257,35 @@ def test_solve_empty_form_rhs(mesh):
     assert errornorm(x, w) < 1E-10
 
 
+def test_solve_assembled_lhs(mesh):
```
JHopeCollins, line 260: Check both explicit and matrix free matrices.
```suggestion
@pytest.mark.parametrize("mat_type", ["aij", "matfree"])
def test_solve_assembled_lhs(mesh, mat_type):
```

tests/firedrake/regression/test_solving_interface.py
```diff
@@ -262,6 +257,35 @@ def test_solve_empty_form_rhs(mesh):
...
+    V = FunctionSpace(mesh, "CG", 1)
+    x, = SpatialCoordinate(mesh)
+
+    u = TrialFunction(V)
+    v = TestFunction(V)
+    a = inner(grad(u), grad(v)) * dx
+    bcs = [DirichletBC(V, x, "on_boundary")]
+
+    # Assemble the matrix with bcs
+    A = assemble(a, bcs=bcs)
```
JHopeCollins, line 270:
```suggestion
    A = assemble(a, bcs=bcs, mat_type=mat_type)
```

JHopeCollins (changes requested):
Looking good generally, I just had a couple of points. Primarily about whether `allocate` should return a new object each time it's called, and about `form_compiler_parameter` instead of `fc_params` as the more explicit option.

firedrake/linear_solver.py
pbrubeck: This is to mantain backwards compatibility, we could only get iteration counts via `LinearSolver.ksp.getIterationNumber()`

firedrake/matrix.py
pbrubeck: This is going to break API

firedrake/assemble.py
```diff
@@ -344,7 +344,9 @@ def __init__(self,
     def allocate(self):
         rank = len(self._form.arguments())
         if rank == 2 and not self._diagonal:
-            if self._mat_type == "matfree":
+            if isinstance(self._form, matrix.MatrixBase):
+                return self._form
```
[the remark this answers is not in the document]
pbrubeck: For symbolic forms, it makes sense to call allocate multiple times to create different objects in memory, but this does not make much sense for already assembled objects.  As we have discussed, `MatrixBase` objects should be treated as constants, and calling `assemble` on them is a no-op.  

So arguably, the purpose of `allocate` here is just to return the pointer in memory of the assembly output.

firedrake/matrix.py
pbrubeck:
```suggestion
        self.form_compiler_parameters = fc_params
```

firedrake/matrix.py
JHopeCollins: On that note, I know you haven't changed `options_prefix` or `appctx` kwargs here, but could you add them to the docstring? Currently you just have to _know_ that you can pass them.

dham (approved):

connorjward:
> There's the potential to specify `mat_type` on an already assembled Jacobian. Not setting `mat_type` would result in inadequate default `solver_parameters`, so `mat_type` must be set internally to match that of the assembled Jacobian.
> 
> Should we: a. Allow `mat_type` to be passed in `solver_parameters` and assert that it matches that of the assembled Jacobian, b. Disallow `mat_type` to be passed in `solver_parameters` for an already assembled Jacobian?
> 
> Option b would be consistent with the treatment we have for bcs on an assembled Jacobian, but could potentially break codes, as `LinearSolver` simply ignored `solver_parameters[mat_type]`.

I would prefer option b, though if there is any risk that it breaks behaviour then we should be conservative until Firedrake is versioned. I guess it depends when you want this PR to be merged.

dham (approved):

connorjward (approved):

PR #4018, opened 2025-02-07
Hot-fix for broken Stokes test

pbrubeck:
# Description
Two PETSc issues have arised after a recent hardware update. These should be further investigated and reported if necessary.

1. The macro Stokes tests broke as the combination of a **nullspace** and the solver parameters (GMRES + MUMPS LU) obtained an incorrect solution (without raising a solver error) on new hardware. This PR is to ensure that these tests work with the default solver parameters (preonly + MUMPS LU). The convergence of the solution test and div-free test have been now merged into a single test.

2. Another non-macro Stokes (IPDG H(div)) test was not meeting an absolute tolerance. Here we make it pass by using more sensible solver parameters (parallel direct solver as opposed to redundant). The old solver parameters alluded to segfaults with (MUMPS Cholesky) in complex mode, and for this reason a (redundant + PETSc Cholesky) solver was used instead. This PR changes the solver to a parallel one using (MUMPS + LU).

### DIFF
--- a/tests/firedrake/macro/test_macro_solve.py
+++ b/tests/firedrake/macro/test_macro_solve.py
@@ -6,10 +6,11 @@ from firedrake import *
 @pytest.fixture(params=("square", "cube"))
 def mh(request):
     if request.param == "square":
-        base_msh = UnitSquareMesh(3, 3)
+        return [UnitSquareMesh(2**r, 2**r) for r in range(1, 4)]
     elif request.param == "cube":
-        base_msh = UnitCubeMesh(2, 2, 2)
-    return MeshHierarchy(base_msh, 2)
+        return [UnitCubeMesh(2**r, 2**r, 2**r) for r in range(1, 4)]
+    else:
+        raise ValueError("Unexpected mesh type")
 
 
 @pytest.fixture(params=["iso", "alfeld", "th"])
@@ -33,35 +34,27 @@ def mixed_element(mh, variant):
     return Vel, Pel
 
 
-def mesh_sizes(mh):
-    mesh_size = []
-    for msh in mh:
-        DG0 = FunctionSpace(msh, "DG", 0)
-        h = Function(DG0).interpolate(CellDiameter(msh))
-        with h.dat.vec as hvec:
-            _, maxh = hvec.max()
-        mesh_size.append(maxh)
-    return mesh_size
-
-
-def conv_rates(x, h):
+def conv_rates(x, h=None):
     x = np.asarray(x)
-    h = np.asarray(h)
-    return np.log2(x[:-1] / x[1:]) / np.log2(h[:-1] / h[1:])
+    rates = np.log2(x[:-1] / x[1:])
+    if h is not None:
+        h = np.asarray(h)
+        rates /= np.log2(h[:-1] / h[1:])
+    return rates
 
 
 @pytest.fixture
 def convergence_test(variant):
     if variant == "iso":
-        def check(uerr, perr, h):
-            return (conv_rates(uerr, h)[-1] >= 1.9
+        def check(uerr, perr):
+            return (conv_rates(uerr)[-1] >= 1.9
                     and np.allclose(perr, 0, atol=1.e-8))
     elif variant == "alfeld":
-        def check(uerr, perr, h):
+        def check(uerr, perr):
             return (np.allclose(uerr, 0, atol=5.e-9)
                     and np.allclose(perr, 0, atol=5.e-7))
     elif variant == "th":
-        def check(uerr, perr, h):
+        def check(uerr, perr):
             return (np.allclose(uerr, 0, atol=1.e-10)
                     and np.allclose(perr, 0, atol=1.e-8))
     return check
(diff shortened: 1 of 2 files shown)


connorjward (dismissed):
Thanks!

connorjward:
If this is machine dependent we should probably rerun CI until it happens on firedrake3. Otherwise we have no idea if this actually works.

JHopeCollins:
Thanks for looking into this @pbrubeck 

In general I agree with @connorjward  that we should try and avoid this level of sensitivity in tests. It would be good if we could work out why these fixes work so we can judge if this test will be more robust in the future now.

- How is the matrix ordering different with MatNest? The test is serial so I thought that matnest would give the same as aij.
- What happens if the matrix type isn't nest, does it crash or just fail the tests? For mixed systems MUMPs is often more robust than the PETSc LU so switching the LU solver might mean you don't need to mess with the parameters much (and hopefully will be more robust in the future).
- I don't actually know what LU does with a matnest, is this a direct solve on the full system, or is it an iterative method seeing as you're specifying GMRES? If it's an iterative method, then we should be specifying the solver tolerance to be low enough that the tolerances for `allclose` won't be sensitive to hardware changes, changes in  LU implementation etc.

pbrubeck:
> * How is the matrix ordering different with MatNest? The test is serial so I thought that matnest would give the same as aij.
I not sure what is the exact ordering, but it is definitely informing MUMPS more about the saddle-point structure.

> * What happens if the matrix type isn't nest, does it crash or just fail the tests? For mixed systems MUMPs is often more robust than the PETSc LU so switching the LU solver might mean you don't need to mess with the parameters much (and hopefully will be more robust in the future).
We need to specify `pc_type="lu`" to override the default pc_type=fieldsplit that we set if we detect a MatNest, this does not mean that `pc_factor_mat_solver_type="petsc"`, but MUMPS will be used if available. 

> * I don't actually know what LU does with a matnest, is this a direct solve on the full system, or is it an iterative method seeing as you're specifying GMRES? If it's an iterative method, then we should be specifying the solver tolerance to be low enough that the tolerances for `allclose` won't be sensitive to hardware changes, changes in  LU implementation etc.
LU is just LU on a MatNest, we get a direct solve, with more robustness with respect to the nullspace.

JHopeCollins:
I was about to post a comment, but it looks like you have already realised - if you inspect `ksp_view` then the PETSc LU was being used, not MUMPS. if you either specify mumps directly or don't specify parameters at all, then you don't have to use matnest, or gmres.

pbrubeck:
> I was about to post a comment, but it looks like you have already realised - if you inspect `ksp_view` then the PETSc LU was being used if matnest wasn't specified, rather than MUMPS. if you either specify mumps directly or don't specify parameters at all, then you don't have to use matnest, or gmres.

This is true for aij, not for nest

pbrubeck:
I am fairly sure I tried with the default parameters, right now I am just double-checking

JHopeCollins:
Yes, if matnest is specified then it defaults to MUMPS, but if mataij was specified then it defaulted to PETSc LU.
The defaults work for me locally by the way.

I think if the defaults pass on the problem machine, then I'm happy for that to go in. I'll look at whether the logic for setting solver defaults needs updating separately.

pbrubeck:
@JHopeCollins looks like it's passing. So what we originally `ksp_type=gmres` had wasn't using GMRES+LU, as I thought. I think we had put GMRES there for iterative retirement. But I didn't realize that all PC default parameters weren't set by us, but by PETSc. I would try again with the original mesh if everything passes

JHopeCollins:
Does setting `"assembled_redundant_pc_factor_mat_solver_type": "mumps"` make the HDiv Stokes test pass with the original tolerance? It's currently using PETSc. The original isn't failing for me so I can't check myself.

JHopeCollins:
> @JHopeCollins looks like it's passing. So what we originally had `ksp_type=gmres` wasn't using GMRES+LU, as I thought. I think we had put GMRES there for iterative retirement. But I didn't realize that all PC default parameters weren't set by us, but by PETSc. I will try again with the original mesh if everything passes

Ok great. If we can fix this by just removing all the explicit solver parameters, and still use the original mesh hierarchy, that would be nice. The setting of defaults gets split between Firedrake and PETSc, and it seemed like the factorisation solver was slipping through the cracks here.

JHopeCollins:
Related, see https://github.com/firedrakeproject/firedrake/discussions/4030 about default parameter selection.

pbrubeck:
Looking closer into the original parameters, it seems that the issue was mostly due to how left-preconditioned gmres handles a nullspace. Right-preconditioned gmres seems to do the right thing, so maybe it'd be good to detect such cases and change the default.  
```
V = FunctionSpace(UnitTriangleMesh(), "DG", 0)
problem = LinearVariationalProblem(inner(TrialFunction(V), TestFunction(V))*dx, 0, Function(V))
solver = LinearVariationalSolver(problem, solver_parameters={"ksp_type": "gmres"})
print(solver.parameters)
{'ksp_type': 'gmres',
 'snes_type': 'ksponly',
 'mat_type': 'aij',
 'ksp_rtol': 1e-07,
 'pc_type': 'lu',
 'pc_factor_mat_solver_type': 'mumps',
 'pc_factor_mat_mumps_icntl_14': 200}
```

connorjward:
It would be extremely helpful to have this merged this morning so we can have some green ticks for the meeting later. Is this ready to go?

pbrubeck:
> It would be extremely helpful to have this merged this morning so we can have some green ticks for the meeting later. Is this ready to go?

I am happy with the current state

connorjward (approved):

connorjward:
Thanks!!

Issue #4014, opened 2025-02-07
Memory Growth and Unexpected Behaviour in Firedrake Adjoint


sghelichkhani:
We are debugging a large-scale Stokes optimisation problem that eventually runs out of memory (https://github.com/g-adopt/g-adopt/issues/160). Since we are dealing with millions of degrees of freedom, we rely on checkpointing to disk to manage memory. While testing smaller reproducer cases, we see unexpected memory growth even after the tape is generated and throughout forward and backward passes.

**What We Expected vs. What’s Happening**
	•	**Expected:** Once the tape is populated, and after the first calls to `ReducedFunctional.__call__` and `ReducedFunctional.derivative`, memory usage should stay constant.
	•	**Actual**: Memory keeps increasing with every forward and derivative call and steadily.
	•	In our minimal reproducer, checkpointing to disk actually increases memory usage!!!

**Minimal Reproducer**

Using memory_profile I profile repeated calls to `ReducedFunctional.__call__` and `ReducedFunctional.derivative`. Simply run `mprof run ...` 

**Code for Reproduction**
```python3
from firedrake import *
from firedrake.adjoint import *
import gc

def test():
    T_c, rf = rf_generator()

    for i in range(5):
        gc.collect()
        rf.__call__(T_c)
        gc.collect()
        rf.derivative()

def rf_generator():
    tape = get_working_tape()
    tape.clear_tape()
    continue_annotation()
    enable_disk_checkpointing()
    
    mesh = RectangleMesh(100, 100, 1.0, 1.0)
    mesh = checkpointable_mesh(mesh)

    V = VectorFunctionSpace(mesh, "CG", 2)
    Q = FunctionSpace(mesh, "CG", 1)

    X = SpatialCoordinate(mesh)
    w = Function(V, name="rotation").interpolate(as_vector([-X[1] - 0.5, X[0] - 0.5]))
    T_c = Function(Q, name="control")
    T = Function(Q, name="Temperature")
    
    T_c.interpolate(0.1 * exp(-0.5 * ((X - as_vector((0.75, 0.5))) / Constant(0.1)) ** 2))
    control = Control(T_c)
    T.assign(T_c)

    for i in range(20):
        T.interpolate(T + inner(grad(T), w) * Constant(0.0001))

    objective = assemble(T**2 * dx)

    pause_annotation()
    return T_c, ReducedFunctional(objective, control)

if __name__ == "__main__":
    test()
```

Without checkpoint to disk: 
[...]

PR #4020, opened 2025-02-08
Fix memory leak for disk checkpointing

Ig-dolci:
# Description

This PR seeks to fix the memory leak by calling a `function.name` method at the `CheckpointFunction` constructor. In addition, this PR proposes disk checkpointing using the `SingleDiskStorageSchedule` schedule and pyadjoint checkpoint manager to execute the solvers. This code also enables the user to choose other schedules from `checkpoint_schedules` with disk storage. 

**There is a point to consider here:**

- The `SingleDiskStorageSchedule` schedule stores only the adjoint dependence data. So, we have to pay attention to the cases requiring all checkpoint data storage on disk in every single time step, i.e., the required data to restart the forward solver and the required data for the adjoint computations. The `SingleDiskStorageSchedule` schedule does not support that.

_Observation:_ 

- The tests will pass if the pyadjoint [PR 195](https://github.com/dolfin-adjoint/pyadjoint/pull/195) is merged.

- I will wait for the [PR 4023](https://github.com/firedrakeproject/firedrake/pull/4023) to be merged to open this PR.


dham:
I don't believe this is correct. Merely referencing `function.function_space()` in `CheckpointFunction.__init__` does not create persistent references to `function`.  That isn't going to be the source of the memory leak.

sghelichkhani:
Just confirming that this doesn't change any of our reproducers. Also noting that the behaviour is slightly different on my mac and our HPC system. Unlike my mac, the RSS seems to go flat after the first derivative (see image). Not sure why it works differently on mac. But at any rate, the memory usage seems to go higher when disk-checkpointing.  
Screenshot 2025-02-10 at 14 02 25

dham:
Can I just check how many Degrees of Freedom this is? I'm trying to work out how many vectors we must have leaked.

Ig-dolci:
Firstly, I have identified memory leaks in two cases involving disk checkpointing:

1. When using only `enable_disk_checkpointing()`.
2. When using `enable_disk_checkpointing()` along with `checkpoint_schedules`.

Additionally, the `function.function_space()` is **not** the source of the leak. However, assigning `self.name = function.name` (where `function.name` is a callable) contributes to the leak in both cases.

The following chart evaluates **Case 2** on the master branch (black curve) and with a modification replacing `self.name = function.name` with `self.name = function.name()` (blue curve):

Figure_1

The red curve represents the scenario where **no disk checkpointing** is applied.

Although changing `self.name = function.name` to `self.name = function.name()` mitigates the issue in Case 2, it does **not** resolve the memory leak in Case 1. The cause of this leak is still unknown.

### Next Steps

We have two options moving forward:

1. Investigate the memory leak in Case 1.
2. Use disk checkpointing **only** with `checkpoint_schedules`

For **Option 2** (this PR proposal), here are the results obtained:

Figure_2

- The **black curve** corresponds to the results with this PR.
- The **blue curve** represents the use of `enable_disk_checkpointing()` without `checkpoint_schedules`.
- The **red curve** corresponds to the case where **no disk** checkpointing is applied.

**These charts use the example added [#4014 issue](https://github.com/firedrakeproject/firedrake/issues/4014)**.

sghelichkhani:
@Ig-dolci I have tried hard to reproduce your black curve, but I still see a big leak on `dolci/fix_mem_leak_diskcheckpointing`:  
comp
Am I missing something?

sghelichkhani:
@dham For the graph I have above (and presumably what @Ig-dolci  is plotting) I am using this reproducer https://github.com/g-adopt/g-adopt/blob/adjoint-memory/demos/mantle_convection/test/tester.py, where `Q.dim()` is 10201, and we are doing 500 timesteps.

Ig-dolci:
> @Ig-dolci I have tried hard to reproduce your black curve, but I still see a big leak on `dolci/fix_mem_leak_diskcheckpointing`: comp Am I missing something?

I think is missing to replace the time loop from `for i in range(total_steps):` to `for i in tape.timestepper(iter(range(total_steps)))`. My bad, I have to document this properly or write a warning for the users to be informed.

angus-g:
> > @Ig-dolci I have tried hard to reproduce your black curve, but I still see a big leak on `dolci/fix_mem_leak_diskcheckpointing`: comp Am I missing something?
> 
> I think is missing to replace the time loop from `for i in range(total_steps):` to `for i in tape.timestepper(iter(range(total_steps)))`. My bad, I have to document this properly or write a warning for the users to be informed.

Are you also using the extra garbage collection from https://github.com/dolfin-adjoint/pyadjoint/pull/187?

Ig-dolci:
> > > @Ig-dolci I have tried hard to reproduce your black curve, but I still see a big leak on `dolci/fix_mem_leak_diskcheckpointing`: comp Am I missing something?
> > 
> > 
> > I think is missing to replace the time loop from `for i in range(total_steps):` to `for i in tape.timestepper(iter(range(total_steps)))`. My bad, I have to document this properly or write a warning for the users to be informed.
> 
> Are you also using the extra garbage collection from [dolfin-adjoint/pyadjoint#187](https://github.com/dolfin-adjoint/pyadjoint/pull/187)?

Yes. I just noticed that I used it. Applying the garbage collection between a number of time steps (20 steps) helped decrease memory usage by almost 20% for this case.

dham (approved):

PR #4025, opened 2025-02-10
`EnsembleFunction` and `EnsembleFunctionSpace`.

JHopeCollins:
## Description
This PR introduces this abstraction to Firedrake as an `EnsembleFunction` representing the distributed mixed function, defined over a `EnsembleFunctionSpace`, and an `EnsembleCofunction` and `EnsembleDualFunction` representing the dual objects.

### Why?
Methods that use the `Ensemble` will often involve something that looks like a mixed finite element function, where the subcomponents are distributed over the ensemble members. We need to treat the subcomponents as `firedrake.Functions` within each ensemble member, but also to have some collective semantics over the entire ensemble. See for example the [`AllAtOnceFunction`](https://github.com/firedrakeproject/asQ/blob/76cb14a16eedccc4d3a180fb6f120d437c45eba6/asQ/allatonce/function.py#L46) in asQ which represents a timeseries over an ensemble.

Note that, whereas `firedrake.Function` does not allow nested mixed-ness, a particular component of `EnsembleFunction` can be mixed - this is essential for things like a timeseries of a mixed function, e.g. shallow water equations.

### Demo
At the bottom is a small script using the new objects to solve the heat equation with backward euler, parallel in time, with a block Jacobi preconditioner. It shows setting some typical usage including
1. setting up the ensemble objects
2. accessing data via local `Function` subcomponents
3. accessing data via the global petsc `Vec`
4. setting up global petsc objects
5. ensemble communication

## `EnsembleFunctionSpace`
- An `EnsembleFunction` is defined from an `EnsembleFunctionSpace`, which just collects an `Ensemble` with a list of `firedrake.FunctionSpace` for each component on the _local_ ensemble member. These are accessible through `EnsembleFunctionSpace.local_spaces`.
- The `EnsembleFunctionSpace` also has some handy utilities, e.g. a `dual` method, equality comparison, the number of dofs on the local rank, on the local ensemble member, and on the whole global function space.

- Currently the global list of subcomponents is not available, i.e. you cannot programmatically check from rank `i` what `FunctionSpace` is in `EnsembleFunctionSpace.local_spaces` on rank `j`, but I'm open to arguments for including this.

## `EnsembleFunction`
- It's important to point out that this does not have any UFL symbolic information, i.e. it is data structure abstraction not a symbolic one. The subcomponents on the local ensemble member can be accessed as `firedrake.Functions`, with the usual UFL symbolic information, via `EnsembleFunction.subfunctions`.

- Equivalents to a few `firedrake.Function` arithmetic operations are implemented, e.g. `assign`, `zero`, and addition/multiplication overloads.
- A `PETSc.Vec` defined over the `Ensemble.global_comm` for the data in all subcomponents on all ranks is accessed via context managers `vec`, `vec_ro`, and `vec_wo`.

## Adjoint
Currently no `EnsembleFunction` operations are taped, so it cannot be used for adjoint problems in the same way `firedrake.Function` can be.
However, it does has the bare minimum `OverloadedType` functionality implemented to be usable as a `Control` type, _if_ you take responsibility for evaluating the tape (e.g. with `EnsembleReducedFunctional`, or the 4DVar reduced functional in another PR).
For example:
- `_ad_dot` - collective over the ensemble so it will DTRT with `pyadjoint.taylor_test` and we should be able to remove the `xfail` markers [here](https://github.com/firedrakeproject/firedrake/blob/96501362c73e17e28131294ccc7ee611b9b64e64/tests/firedrake/ensemble_reduced_functional/test_reduced_functional.py#L45).
- `_ad_to_list` and `_ad_assign_numpy` so it will work with the pyadjoint interface to scipy optimize that relies on `ReducedFunctionalNumPy`.
- `_ad_{to,from}_petsc` so it can be used with `TAO`. 

I'm planning on integrating `EnsembleFunction` with `EnsembleReducedFunctional` in a subsequent PR.

## Implementation notes.
To build PETSc solvers over an `Ensemble`, what we need is a mechanism for glueing two types of operation over the same data together.
1. Providing Firedrake with `firedrake.Function` objects on each ensemble member to evaluate local finite element operations.
6. Providing PETSc with `Vec` objects on the global comm which represent the global solution, to pass to `KSP`, `SNES`, etc.

This is shown in the diagram below for an `EnsembleFunction` with 8 subcomponents, distributed equally over four ensemble members.

- Internally all subcomponents on each ensemble member are represented as a (flattened) `firedrake.MixedFunctionSpace` ([here](https://github.com/firedrakeproject/firedrake/blob/fdaf00f918fcffdc3b7869501706c3941be89304/firedrake/ensemble/ensemble_functionspace.py#L96)).
- The `EnsembleFunction` creates a `firedrake.Function` (called `_full_local_function`) from this mixed function space to store all local subcomponents. These are the dark blue boxes in the diagram, each with two subcomponents in light blue.
(The API is independent of the choice to use `MixedFunctionSpace` internally).
- The `EnsembleFunction.subfunctions` is a tuple of `firedrake.Functions` that view the relevant components of `_full_local_function.subfunctions` (see [here](https://github.com/firedrakeproject/firedrake/blob/fdaf00f918fcffdc3b7869501706c3941be89304/firedrake/ensemble/ensemble_function.py#L67)). These are the light blue boxes on the right of the diagram, and will be the main way users interact with the data.
For any elements of `EnsembleFunction.subfunctions` that are themselves mixed, we need to construct a new `MixedDat` to view multiple components of `_full_local_function.subfunctions`.
- The `Vec` accessed via the `EnsembleFunction.vec` context managers is created over the `Ensemble.global_comm`, and is shown in dark green on the left of the diagram. This `Vec` is created as a view over the data in the internal mixed `firedrake.Function` by passing `_full_local_function.dat.vec.array` to `Vec.createWithArray` ([here](https://github.com/firedrakeproject/firedrake/blob/fdaf00f918fcffdc3b7869501706c3941be89304/firedrake/ensemble/ensemble_function.py#L54)) (shown as the light green boxes).
This means that for the data in `EnsembleFunction.vec` to be valid (the large dark green box), we need the data in `_full_local_function.dat.vec` to be valid (the smaller light green box attached to the MixedFunction). To ensure this data is valid, we nest the `_full_local_function.dat.vec` context managers inside the `EnsembleFunction.vec` context managers [here](https://github.com/firedrakeproject/firedrake/blob/fdaf00f918fcffdc3b7869501706c3941be89304/firedrake/ensemble/ensemble_function.py#L198).
One thing to be careful of, is that if the data in `_full_local_function.dat.vec` has been modified, (i.e. if the user has done anything with `EnsembleFunction.subfunctions`!) then the data in the global `EnsembleFunction.vec` will have changed, but without `EnsembleFunction.vec` knowing. This means we have to manually increase the state counter for `EnsembleFunction.vec` and `EnsembleFunction.vec_ro`.

asQ_datalayout_updated_fitted

```python
import firedrake as fd
from numpy.random import seed, random_sample
seed(6)

# Solve the all-at-once system for the heat equation with backwards Euler.
# If A1 = M + K and A0 = -M, then the lower-triangular and bidiagonal
# all-at-once Jacobian is:
# A1  0  0  0
# A0 A1  0  0
#  0 A0 A1  0
#  0  0 A0 A1
# The preconditioner is block diagonal using the A1 blocks

ensemble = fd.Ensemble(fd.COMM_WORLD, 1)
ensemble_rank = ensemble.ensemble_rank
ensemble_size = ensemble.ensemble_size

nx = 16
mesh = fd.UnitIntervalMesh(nx, comm=ensemble.comm)

dx = 1/nx
cfl = 2
dt = cfl*dx**2

V = fd.FunctionSpace(mesh, "CG", 1)

u = fd.TrialFunction(V)
v = fd.TestFunction(V)

# mass and stiffness matrices for heat equation
M = u*v*fd.dx
K = dt*fd.inner(fd.grad(u), fd.grad(v))*fd.dx

# backward euler timestepping
A1 = M + K
A0 = -M

# how many timesteps on each ensemble rank
time_partition = [1, 2, 1, 5, 3]

# function space for the time series
local_spaces = [V for _ in range(time_partition[ensemble_rank])]
W = fd.EnsembleFunctionSpace(local_spaces, ensemble)


# python mat context for the Jacobian
class AllAtOnceMat:
    def __init__(self):
        self.x = fd.EnsembleFunction(W)
        self.y = fd.EnsembleCofunction(W.dual())

        self.xhalo = fd.Function(V)
        self.yhalo = fd.Cofunction(V.dual())

    def mult(self, A, x, y):
        # copy input into ensemble function
        with self.x.vec_wo() as xvec:
            x.copy(xvec)

        # receive the time-halos from the previous ensemble rank
        self.update_halos()
        self.y.zero()
        self.yhalo.zero()

        # contribution of diagonal blocks A1
        for xsub, ysub in zip(self.x.subfunctions,
                              self.y.subfunctions):
            fd.assemble(fd.action(A1, xsub),
                        tensor=ysub)

        # contribution of local sub-diagonal blocks A0
        for xsub, ysub in zip(self.x.subfunctions[:-1],
                              self.y.subfunctions[1:]):
            fd.assemble(fd.action(A0, xsub),
                        tensor=self.yhalo)
            ysub += self.yhalo

        # contribution of halo sub-diagonal blocks A0
        if ensemble_rank != 0:
            fd.assemble(fd.action(A0, self.xhalo),
                        tensor=self.yhalo)
            ysub = self.y.subfunctions[0]
            ysub += self.yhalo

        # copy result back out
        with self.y.vec_ro() as yvec:
            yvec.copy(y)

    def update_halos(self):
        # halo swap is a right shift
        src = (ensemble_rank - 1) % ensemble_size
        dst = (ensemble_rank + 1) % ensemble_size
        frecv = self.xhalo
        fsend = self.x.subfunctions[-1]

        ensemble.sendrecv(fsend=fsend, dest=dst, sendtag=dst,
                          frecv=frecv, source=src, recvtag=ensemble_rank)


# python pc context for a block-diagonal preconditioner with A1
class BJacobiPC:
    def __init__(self):
        self.x = fd.EnsembleCofunction(W.dual())
        self.y = fd.EnsembleFunction(W)

        self.blocks = []
        for i in range(W.nlocal_spaces):
            global_step = i + sum(time_partition[:ensemble_rank])

            # use the subcomponents directly here, avoids additional shuffling in apply
            problem = fd.LinearVariationalProblem(
                A1, self.x.subfunctions[i],
                self.y.subfunctions[i])

            solver = fd.LinearVariationalSolver(
                problem, options_prefix=f"step_{global_step}")

            self.blocks.append(solver)

    def apply(self, pc, x, y):
        with self.x.vec_wo() as xvec:
            x.copy(xvec)

        for block in self.blocks:
            block.solve()

        with self.y.vec_ro() as yvec:
            yvec.copy(y)


# Set up the PETSc objects
sizes = (W.nlocal_rank_dofs, W.nglobal_dofs)
mat = PETSc.Mat().createPython(
    (sizes, sizes), AllAtOnceMat(),
    comm=ensemble.global_comm)
mat.setUp()

ksp = PETSc.KSP().create(comm=ensemble.global_comm)
ksp.setOptionsPrefix("")
ksp.setType("richardson")
ksp.setOperators(mat)
ksp.setFromOptions()
ksp.setUp()

ksp.pc.setType("python")
ksp.pc.setPythonContext(BJacobiPC())

# the right hand side an solution functions
u = fd.EnsembleFunction(W)
b = fd.EnsembleCofunction(W.dual())

with u.vec_ro() as sol, b.vec_wo() as rhs:
    rhs.array[:] = ensemble_rank*random_sample(rhs.array.shape)
    ksp.solve(rhs, sol)

# now the timesteps are accessible via subfunctions
if ensemble_rank == ensemble_size - 1:
    final_timestep = u.subfunctions[-1]
```


firedrake/ensemble/ensemble_function.py
```diff
@@ -0,0 +1,292 @@
+from firedrake.petsc import PETSc
+from firedrake.ensemble.ensemble_functionspace import (
+    EnsembleFunctionSpace, EnsembleDualSpace)
+from firedrake.adjoint_utils import EnsembleFunctionMixin
+from firedrake.function import Function
+from firedrake.norms import norm
+from pyop2 import MixedDat
+
+from functools import cached_property
```
connorjward, line 9: Feel free to ignore (we violate this everywhere) but conventionally standard library packages go above third party ones

firedrake/ensemble/ensemble_function.py
```diff
@@ -0,0 +1,292 @@
...
+from contextlib import contextmanager
+
+__all__ = ("EnsembleFunction", "EnsembleCofunction")
+
+
+class EnsembleFunctionBase(EnsembleFunctionMixin):
+    """
+    A mixed (co)function defined on a :class:`firedrake.Ensemble`.
+    The subcomponents are distributed over the ensemble members, and
+    are specified locally in a :class:`firedrake.EnsembleFunctionSpace`.
```
connorjward, line 19: Not sure what you mean by "are specified locally"

firedrake/ensemble/ensemble_function.py
```diff
@@ -0,0 +1,292 @@
...
+
+    Parameters
+    ----------
+
+    function_space : :class:`firedrake.EnsembleFunctionSpace`.
```
connorjward, line 24: I think numpydoc doesn't need rst so `function_space : EnsembleFunctionSpace` will work. Please review the docs to make sure they render properly.

firedrake/ensemble/ensemble_function.py
```diff
@@ -0,0 +1,292 @@
...
+        The function space of the (co)function.
+
+    Notes
+    -----
+    Passing a :class:`firedrake.EnsembleDualSpace` to :class:`firedrake.EnsembleFunction`
```
connorjward, line 29:
```suggestion
    Passing an `EnsembleDualSpace` to `EnsembleFunction`
```
I think... (and likewise everywhere else)

firedrake/ensemble/ensemble_function.py
```diff
@@ -0,0 +1,292 @@
...
+    will return an instance of :class:`firedrake.EnsembleCofunction`.
+
+    This class does not carry UFL symbolic information, unlike a
+    :class:`firedrake.Function`. UFL expressions can only be defined
+    locally on each ensemble member using a `firedrake.Function`
+    from `EnsembleFunction.subfunctions`.
+
+    See also:
```
connorjward, line 37: Numpydoc has a [special section for this](https://numpydoc.readthedocs.io/en/latest/format.html#see-also)

firedrake/ensemble/ensemble_function.py
```diff
@@ -0,0 +1,292 @@
...
+
+    @PETSc.Log.EventDecorator()
+    def riesz_representation(self, riesz_map="L2", **kwargs):
+        """
+        Return the Riesz representation of this :class:`EnsembleFunction`
+        with respect to the given Riesz map.
+
+        Parameters
+        ----------
+
```
connorjward, line 99:
```suggestion
```

firedrake/ensemble/ensemble_function.py
```diff
@@ -0,0 +1,292 @@
...
+        other : :class:`EnsembleFunction`
+            The value to assign from.
+
+        subsets : Collection[Optional[:class:`pyop2.types.set.Subset`]]
+            One subset for each local :class:`Function`. None elements
+            will be ignored.  The values of each local function will
+            only be assigned on the nodes on the corresponding subset.
+        """
+        if type(other) is not type(self):
+            raise ValueError(
```
connorjward, line 130:
```suggestion
            raise TypeError(
```

firedrake/ensemble/ensemble_function.py
```diff
@@ -0,0 +1,292 @@
...
+                f"Cannot assign {type(self)} from {type(other)}")
```
connorjward, line 131:
```suggestion
                f"Cannot assign {type(self).__name__} from {type(other).__name__}")
```

firedrake/ensemble/ensemble_function.py
```diff
@@ -0,0 +1,292 @@
...
+        else:
+            for us in self.subfunctions:
+                us *= other
+        return self
+
+    @PETSc.Log.EventDecorator()
+    def __add__(self, other):
+        new = self.copy()
+        new += other
+        return new
```
connorjward, line 185: Not really a problem, but it seems strange to me to define `__add__` in terms of `__iadd__`. I would do it the other way around.

firedrake/ensemble/ensemble_function.py
```diff
@@ -0,0 +1,292 @@
...
+
+    @PETSc.Log.EventDecorator()
+    def __mul__(self, other):
+        new = self.copy()
+        new *= other
+        return new
+
+    @PETSc.Log.EventDecorator()
+    def __rmul__(self, other):
+        return self.__mul__(other)
```
connorjward, line 195: Is this definitely safe to do?

firedrake/ensemble/ensemble_function.py
```diff
@@ -0,0 +1,292 @@
...
+
+    @contextmanager
+    def vec(self):
```
connorjward, line 198: I *believe* that David thinks we should use `XXX_rw` instead of `XXX` to make things more explicit (in the context of `.data` vs `.data_rw` but the same applies here)

firedrake/ensemble/ensemble_function.py
```diff
@@ -0,0 +1,292 @@
...
+    - Primal ensemble objects: :class:`firedrake.EnsembleFunctionSpace` and :class:`firedrake.EnsembleFunction`.
+    - Dual ensemble objects: :class:`firedrake.EnsembleDualSpace` and :class:`firedrake.EnsembleCofunction`.
+    """
+
+    @PETSc.Log.EventDecorator()
+    @EnsembleFunctionMixin._ad_annotate_init
+    def __init__(self, function_space):
+        self._fs = function_space
+
+        self._fbuf = Function(function_space._full_local_space)
```
connorjward, line 47: This isn't a very clear name

firedrake/ensemble/ensemble_function.py
```diff
@@ -0,0 +1,292 @@
...
+        It is invalid to access the Vec outside of a context manager.
+        """
+        # _fbuf.vec shares the same storage as _vec, so we need this
+        # nested context manager so that the data gets copied to/from
+        # the Function.dat storage and _vec.
+        # However, this copy is done without _vec knowing, so we have
+        # to manually increment the state.
+        with self._fbuf.dat.vec:
+            self._vec.stateIncrease()
+            yield self._vec
```
connorjward, line 232: I don't really understand the need for this. Could you simply have
```py
def vec(self):
    return self._full_local_function.dat.vec
```
or
```py
@contextlib.contextmanager
def vec(self):
    yield from self._full_local_function.dat.vec
```
?

firedrake/ensemble/ensemble_function.py
```diff
@@ -0,0 +1,292 @@
...
+    """
+    A mixed finite element Function distributed over an ensemble.
+
+    Parameters
+    ----------
+
+    function_space : :class:`EnsembleFunctionSpace`
+        The function space of the function.
+    """
+    def __new__(cls, function_space):
```
connorjward, line 251: Is this definitely necessary? I think it leads to some confusing behaviour.

firedrake/ensemble/ensemble_function.py
```diff
@@ -0,0 +1,292 @@
...
+        """
+        Compute the norm of the function.
+        Any arguments are forwarded to :function:`firedrake.norm`.
+
+        Parameters
+        ----------
+
+        norm_type : str
+            The type of norm to compute.
+            See :function:`firedrake.norm` for options.
```
connorjward, line 272: I would omit this if not in the function signature.

firedrake/ensemble/ensemble_function.py
```diff
@@ -0,0 +1,292 @@
...
+    def __init__(self, function_space):
+        if not isinstance(function_space, EnsembleFunctionSpace):
+            raise TypeError(
+                "EnsembleFunction must be created using an EnsembleFunctionSpace")
+        super().__init__(function_space)
+
+    def norm(self, *args, **kwargs):
+        """
+        Compute the norm of the function.
+        Any arguments are forwarded to :function:`firedrake.norm`.
```
connorjward, line 265: I think numpydoc would say to write this as
```suggestion
        """Compute the norm of the function.
        
        Any arguments are forwarded to `firedrake.norm`.
```

firedrake/ensemble/ensemble_functionspace.py
```diff
@@ -0,0 +1,242 @@
...
+from ufl.duals import is_primal, is_dual
+from pyop2.mpi import internal_comm
+from functools import cached_property
+from itertools import chain
+
+
+def _is_primal_or_dual(local_spaces, ensemble):
+    """
+    We need to test primal or dual collectively over all ensemble
+    ranks otherwise some may fail when others pass.
```
connorjward, line 12: Should be in the imperative tone. E.g.: "Return whether all spaces in an ensemble are primal or dual."
The motivation can be written below (with a gap of one line)

firedrake/ensemble/ensemble_functionspace.py
```diff
@@ -0,0 +1,242 @@
...
+
+    Parameters
+    ----------
+
+    local_spaces : Collection
+        The list of :function:`firedrake.FunctionSpace` on the local ensemble.comm.
+    ensemble : :class:`firedrake.Ensemble`
+        The communicator to test collectively over.
```
connorjward, line 20: It's not really a communicator

firedrake/ensemble/ensemble_functionspace.py
```diff
@@ -0,0 +1,242 @@
...
+    """
+    local_comm_valid = all(
+        V.mesh().comm is ensemble.comm for V in local_spaces)
+    global_comm_valid = all(
+        ensemble.ensemble_comm.allgather(local_comm_valid))
+    if not global_comm_valid:
+        raise ValueError(
+            "All spaces must be defined on the ensemble.comm")
+
+    local_types = []
```
connorjward, line 41: Why not have:
```py
local_types = set()
for V in local_spaces:
        if is_primal(V):
            local_types.add('primal')
        elif is_dual(V):
            local_types.add('dual')
        else:
            # can this ever happen?
            local_types.add('invalid')

if len(local_types) > 1:
    local_type = "invalid"
else:
    local_type, = local_types
```
Then you only have to broadcast one string and can avoid needing `chain()`.

firedrake/ensemble/ensemble_functionspace.py
```diff
@@ -0,0 +1,242 @@
...
+    def global_comm(self):
+        """The global communicator from the :class:`firedrake.Ensemble` communicator.
+        """
+        return self._ensemble.global_comm
+
+    @property
+    def local_spaces(self):
+        """The :function:`firedrake.FunctionSpace` on the local ensemble.comm.
+        """
+        return self._local_spaces
```
connorjward, line 131: Up to you but I personally would probably just set `self.local_spaces` instead of writing a getter.

firedrake/ensemble/ensemble_functionspace.py
```diff
@@ -0,0 +1,242 @@
...
+    def nlocal_spaces(self):
+        """The total number of subspaces across all ensemble ranks.
+        """
+        return len(self.local_spaces)
+
+    @cached_property
+    def nglobal_spaces(self):
+        """The total number of subspaces across all ensemble ranks.
+        """
+        return self.ensemble_comm.allreduce(len(self.local_spaces))
```
connorjward, line 156:
```suggestion
        return self.ensemble_comm.allreduce(self.nlocal_spaces)
```

firedrake/ensemble/ensemble_functionspace.py
```diff
@@ -0,0 +1,242 @@
...
+        """
+        return self.ensemble_comm.allreduce(self.nlocal_rank_dofs)
+
+    def _component_indices(self, i):
+        """
+        Return the indices into the local mixed function storage
+        corresponding to the i-th local function space.
+        """
+        offset = sum(len(V) for V in self.local_spaces[:i])
+        return tuple(offset + i for i in range(len(self.local_spaces[i])))
```
connorjward, line 182:
```suggestion
        return tuple(offset + j for j in range(len(self.local_spaces[i])))
```

firedrake/ensemble/ensemble_functionspace.py
```diff
@@ -0,0 +1,242 @@
...
+    @property
+    def nglobal_dofs(self):
+        """The total number of dofs across all subspaces on all ensemble ranks.
+        """
+        return self.ensemble_comm.allreduce(self.nlocal_rank_dofs)
+
+    def _component_indices(self, i):
+        """
+        Return the indices into the local mixed function storage
+        corresponding to the i-th local function space.
```
connorjward, line 179: This is quite confusing, could there be some more comments/docstring?

firedrake/ensemble/ensemble_functionspace.py
```diff
@@ -0,0 +1,242 @@
...
+        """
+        offset = sum(len(V) for V in self.local_spaces[:i])
+        return tuple(offset + i for i in range(len(self.local_spaces[i])))
+
+    def create_vec(self):
+        """Return a PETSc Vec on the global_comm with the same layout as a :class:`firedrake.EnsembleFunction` or :class:`firedrake.EnsembleCofunction` in this function space.
```
connorjward, line 185: break this line

firedrake/ensemble/ensemble_functionspace.py
```diff
@@ -0,0 +1,242 @@
...
+        """
+        vec = PETSc.Vec().create(comm=self.global_comm)
+        vec.setSizes((self.nlocal_dofs, self.nglobal_dofs))
+        vec.setUp()
+        return vec
+
+    def local_to_global_ises(self, i):
+        # COMM_SELF or global_comm?
+        pass
```
connorjward, line 194: Clearly a TODO

firedrake/ensemble/ensemble_functionspace.py
```diff
@@ -0,0 +1,242 @@
...
+        else:
+            local_eq = all(
+                lspace == rspace
+                for lspace, rspace in zip(self.local_spaces,
+                                          other.local_spaces)
+            )
+        return all(self.ensemble.ensemble_comm.allgather(local_eq))
+
+    def __neq__(self, other):
+        return not self.__eq__(other)
```
connorjward, line 213:
```suggestion
        return not self == other
```

firedrake/ensemble/ensemble_functionspace.py
```diff
@@ -0,0 +1,242 @@
...
+
+
+class EnsembleFunctionSpace(EnsembleFunctionSpaceBase):
+    def __new__(cls, local_spaces, ensemble):
```
connorjward, line 217: As above, I don't love this

firedrake/ensemble/ensemble_functionspace.py
```diff
@@ -0,0 +1,242 @@
...
+    ----------
+
+    local_spaces : Collection
+        The list of :function:`firedrake.FunctionSpace` on the local ensemble.comm.
+    ensemble : :class:`firedrake.Ensemble`
+        The communicator to test collectively over.
+
+    Returns
+    -------
+    str : Description of the collective space type:
```
connorjward, line 24: I would prefer to return an enum: `FunctionSpaceType.PRIMAL` or `FunctionSpaceType.DUAL` (bikeshedding fine)

If strings are already standard then this is fine

firedrake/ensemble/ensemble_functionspace.py
```diff
@@ -0,0 +1,242 @@
...
+                for lspace, rspace in zip(self.local_spaces,
+                                          other.local_spaces)
+            )
+        return all(self.ensemble.ensemble_comm.allgather(local_eq))
+
+    def __neq__(self, other):
+        return not self.__eq__(other)
+
+
+class EnsembleFunctionSpace(EnsembleFunctionSpaceBase):
```
connorjward, line 216: Should have a docstring.

firedrake/ensemble/ensemble_functionspace.py
```diff
@@ -0,0 +1,242 @@
...
+
+    def __init__(self, local_spaces, ensemble):
+        space_type = _is_primal_or_dual(local_spaces, ensemble)
+        if space_type != 'primal':
+            raise ValueError(
+                "EnsembleFunctionSpace can only be constructed from primal FunctionSpaces")
+        super().__init__(local_spaces, ensemble)
+
+
+class EnsembleDualSpace(EnsembleFunctionSpaceBase):
```
connorjward, line 236: Should have a docstring.

tests/firedrake/ensemble/test_ensemble_function.py
```diff
@@ -0,0 +1,153 @@
...
+}
+
+# Test EnsembleFunction with 6 subfunctions with the elements below, distributed over 1 or more processors.
+# This element sequence hits a variety of cases:
+# - scalar, vector-valued, vector, tensor, mixed elements
+# - mixed elements with non-scalar, components
+# - identical adjacent components (4, 5)
+# - mixed element with a single component (2)
+# - mixed element with repeated component (6)
+# - mixed element where the first component matches the previous element (5, 6)
```
connorjward, line 49:
```suggestion
# 1. scalar, vector-valued, vector, tensor, mixed elements
# 2. mixed elements with non-scalar, components
# 3. identical adjacent components (4, 5)
# 4. mixed element with a single component (2)
# 5. mixed element with repeated component (6)
# 6. mixed element where the first component matches the previous element (5, 6)
```
a little more readable

tests/firedrake/ensemble/test_ensemble_function.py
```diff
@@ -0,0 +1,153 @@
...
+    else:
+        raise ValueError("Invalid number of ranks")
+
+    return [fd.FunctionSpace(mesh, e) for e in elems]
+
+
+space_type = ["primal", "dual"]
+
+
+@pytest.fixture(params=space_type)
```
connorjward, line 100:
```suggestion
@pytest.fixture(params=["primal", "dual"])
```

tests/firedrake/ensemble/test_ensemble_function.py
```diff
@@ -0,0 +1,153 @@
...
+    for u in ensemblefunc.subfunctions:
+        assert norm(u) < 1e-14, "EnsembleFunction.zero should zero all components"
+
+
+@pytest.mark.parallel(nprocs=[1, 2, 4, 6])
+def test_zero_with_subset(ensemblefunc):
+    """
+    Test setting a subset of all components to zero.
+    """
+    from pyop2 import Subset
```
connorjward, line 138: Move to top of file

tests/firedrake/ensemble/test_ensemble_function.py
```diff
@@ -0,0 +1,153 @@
...
+    assign_scalar(ensemblefunc, nonzero)
+
+    subsets = [None if type(V.ufl_element()) is fd.MixedElement else Subset(V.node_set, [0, 1])
+               for V in ensemblefunc.function_space().local_spaces]
+
+    ensemblefunc.zero(subsets)
+
+    for u, subset in zip(ensemblefunc.subfunctions, subsets):
+        if subset is None:
+            continue
```
connorjward, line 151: why not assert something for this case?

tests/firedrake/ensemble/test_ensemble_functionspace.py
```diff
@@ -0,0 +1,78 @@
...
+#       - same subspaces but different ensemble
+#       - same subspaces but dual
+#       - different subspaces, same length
+#       - different length but matching subspaces (when they exist)
+#       - False only on some ensemble members
+
+# .dual() returns
+#   a) correct type
+#   b) correct value
+#   c) V.dual().dual() == V
```
connorjward, line 78: Clearly a TODO

connorjward (dismissed):
Sorry for all the nitpicks but this is a big addition to the API so we should try and follow what we claim are our best practices. In particular I think the docstrings need tweaking as per https://github.com/firedrakeproject/firedrake/wiki/Docstrings

firedrake/ensemble/ensemble_function.py
JHopeCollins: Anywhere I say "locally" I mean on each `Ensemble.comm`. I should either define this somewhere or be more explicit in each place I currently use "locally".

firedrake/ensemble/ensemble_function.py
JHopeCollins: It's an internal detail (`_`), its a `Function` (`f`), but really it's just a storage buffer (`buf`).
Something like `_full_local_function` might be better to be consistent with `function_space._full_local_space`.

firedrake/ensemble/ensemble_function.py
JHopeCollins: See the implementation notes I've added to the PR description. Let me know if they don't make sense and/or you think there's a better way of doing it.

firedrake/ensemble/ensemble_function.py
JHopeCollins: In general, for any operation with `EnsembleFunctionSpace` or `EnsembleFunction` that is an analogue of an operation with `FunctionSpace` or `Function`, I've tried to be as consistent as possible to reduce the mental overhead of using the `Ensemble*` objects (assuming prior familiarity with the usual non-ensemble objects).

For a `Function`, you can pass a `DualSpace` and get a `Cofunction` back - see the code [here](https://github.com/firedrakeproject/firedrake/blob/96501362c73e17e28131294ccc7ee611b9b64e64/firedrake/function.py#L242). It makes some bits of code a bit simpler where you do essentially the same thing for the primal and dual, because you don't have to have an if-statement just to select `Function(...)` or `Cofunction(...)`.

Personally I would err on the side of consistency, even if the design has issues. If you can show the confusing behaviour then I would say we should change `Function`, and then update `EnsembleFunction` to be consistent.

tests/firedrake/ensemble/test_ensemble_function.py
JHopeCollins: A `Function` in a `MixedFunctionSpace` won't accept a `subset` argument to `assign` or `zero`, so on line 144 I only create a `Subset` for non-mixed components.

I'll add a comment explaining what is going on.

firedrake/adjoint_utils/ensemble_function.py
```diff
@@ -0,0 +1,77 @@
...
+            raise NotImplementedError(
+                "Disk checkpointing not implemented for EnsembleFunctions")
+        else:
+            return self.copy()
+
+    def _ad_restore_at_checkpoint(self, checkpoint):
+        if isinstance(checkpoint, type(self)):
+            return checkpoint
+        raise NotImplementedError(
+            "Checkpointing not implemented for EnsembleFunctions")
```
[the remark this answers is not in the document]
JHopeCollins:
```suggestion
            "Disk checkpointing not implemented for EnsembleFunctions")
```

firedrake/ensemble/ensemble_functionspace.py
```diff
@@ -0,0 +1,300 @@
...
+    def _component_indices(self, i):
+        """
+        Return the indices into the local mixed function storage
+        corresponding to the i-th local function space.
+        """
+        offset = sum(len(V) for V in self.local_spaces[:i])
+        return tuple(offset + j for j in range(len(self.local_spaces[i])))
+
+    def create_vec(self):
+        """Return a PETSc Vec on the global_comm with the same layout as a :class:`firedrake.EnsembleFunction` or :class:`firedrake.EnsembleCofunction` in this function space.
```
JHopeCollins, line 189:
```suggestion
        """Return a PETSc Vec on the `ensemble.global_comm` with the same layout as a
        `firedrake.EnsembleFunction` or `firedrake.EnsembleCofunction` in this function space.
```

firedrake/ensemble/ensemble_function.py
JHopeCollins: I would prefer that too, but see my comment elsewhere about being consistent with Firedrake. If we think the `vec` context manager should be renamed `vec_rw` then I'd prefer it if we changed it in Firedrake _and_ here, rather than just here.

firedrake/ensemble/ensemble_function.py
connorjward: Fine, makes sense (grumble, grumble...)

firedrake/ensemble/ensemble_functionspace.py
JHopeCollins: Resolving, see comments on consistency above.

firedrake/ensemble/ensemble_functionspace.py
JHopeCollins: If this is an enum it should probably go at the UFL level rather than just in one file. I'm ok sticking to strings here, particularly function is supposed to be only an internal implementation detail.

firedrake/ensemble/ensemble_functionspace.py
JHopeCollins: Removed. May be useful in the future but currently unneeded.

firedrake/ensemble/ensemble_functionspace.py
JHopeCollins: Much nicer, thanks.

docs/source/ensemble_parallelism.rst
```diff
@@ -0,0 +1,233 @@
...
+  :align: center
+
+  Spatial and ensemble parallelism for an ensemble with 5 members,
+  each of which is executed in parallel over 5 processors.
+
+The additional functionality required to support ensemble parallelism
+is the ability to send instances of :class:`~.Function` from one
+ensemble to another.  This is handled by the :class:`~.Ensemble` class.
+
+Each ensemble member has the same spatial parallelism, so
```
colinjcotter: "same spatial parallelism" -> "same spatial parallel domain decomposition"

docs/source/ensemble_parallelism.rst
```diff
@@ -0,0 +1,233 @@
...
+instantiating an :class:`~.Ensemble` requires a communicator to split
+(usually, but not necessarily, ``MPI_COMM_WORLD``) plus the number of
+MPI processes to be used in each member of the ensemble (5, in the
+figure above and example code below). The number of ensemble members is
+implicitly calculated by dividing the size of the original
+communicator by the number processes in each ensemble member. The
+total number of processes launched by ``mpiexec`` must therefore be
+equal to the product of the number of ensemble members with the number of
+processes to be used for each ensemble member.
+
```
colinjcotter, line 51: and an exception will be raised if this is not the case.

docs/source/ensemble_parallelism.rst
```diff
@@ -0,0 +1,233 @@
...
+using ``subspaces``, the :func:`~.FunctionSpace` for each local component
+of an :class:`~.EnsembleFunctionSpace` can be accessed via
+``EnsembleFunctionSpace.local_spaces``.  Various other methods and
+properties such as ``dual`` and ``nglobal_spaces`` are also available.
+
+An :class:`~.EnsembleFunction` and :class:`~.EnsembleCofunction` can be
+created from the :class:`~.EnsembleFunctionSpace`. These have a ``subfunctions``
+property that can be used to access the components on the local ensemble
+member. Each element in ``EnsembleFunction.subfunctions`` is itself just a
+normal Firedrake :class:`~.Function`.
```
colinjcotter: maybe emphasise here what happens with sub functions if the component of EnsembleFunctionSpace is a mixed space, presumably we get the whole mixed space.

JHopeCollins: Added the following:

> If a component of the `EnsembleFunctionSpace` is a `MixedFunctionSpace`, then the corresponding component in `EnsembleFunction.subfunctions` will be a mixed `Function` in that `MixedFunctionSpace`.

docs/source/ensemble_parallelism.rst
```diff
@@ -0,0 +1,237 @@
...
+   V = FunctionSpace(mesh, "CG", 1)
+   U = FunctionSpace(mesh, "DG", 0)
+   W = U*V
+
+   if my_ensemble.ensemble_rank == 0:
+      local_spaces = [V, U]
+   else:
+      local_spaces = [V, U, W]
+
+   efs = EnsembleFunctionSpace(local_spaces, my_ensemble)
```
dham: please include from file.

dham (changes requested):

connorjward (approved):

PR #4028, opened 2025-02-11
VariationalSolver: support `sub_mat_type`

pbrubeck:
# Description
Previously, `sub_mat_type` could only be set via the global `paramaters` dict. This PR adds this option to the `solver_parameters` kwarg in `solve()`.

TODO

- [x] docs
- [x] tests


firedrake/mg/ufl_utils.py
```diff
@@ -280,22 +281,30 @@ def coarsen_snescontext(context, self, coefficient_mapping=None):
...
         else:
             default_pmat_type = coarse_mat_type
+            default_sub_mat_type = ""
 
         coarse_pmat_type = opts.getString(f"{solver_prefix}coarse_pmat_type",
                                           default_pmat_type)
+        coarse_sub_mat_type = opts.getString(f"{solver_prefix}coarse_sub_mat_type",
+                                             default_sub_mat_type)
+        if coarse_sub_mat_type == "":
+            coarse_sub_mat_type = None
```
connorjward: Why is `coarse_sub_mat_type` `None` whilst `default_sub_mat_type` is `""`?

firedrake/solving_utils.py
```diff
@@ -141,6 +141,8 @@ class _SNESContext(object):
     :arg pmat_type: Indicates whether the preconditioner (if present) is assembled
         monolithically ('aij'), as a block sparse matrix ('nest') or
         matrix-free (as :class:`~.ImplicitMatrix`, 'matfree').
+    :arg sub_mat_type: Indicates the matrix type for the sparse blocks
```
connorjward: Can we use numpydoc here?

firedrake/solving_utils.py
```diff
@@ -165,7 +167,8 @@ class _SNESContext(object):
     Firedrake level information.
     """
     @PETSc.Log.EventDecorator()
-    def __init__(self, problem, mat_type, pmat_type, appctx=None,
+    def __init__(self, problem, mat_type, pmat_type,
+                 sub_mat_type=None, appctx=None,
```
connorjward: Small thing but a more intuitive parameter ordering might be `mat_type`, `sub_mat_type` then `pmat_type`.

firedrake/mg/ufl_utils.py
pbrubeck: `opts.getString` does not support None as default, but the rest of our code does not interpret `""` as `None`.

firedrake/solving_utils.py
pbrubeck: That's very invasive, but `_SNESContext` is not user facing, so it might be okay? We will need to introduce `sub_mat_type` as a positional argument, or make the others kwargs, which I think is better. Also `sub_mat_type` is only used by MatNest (and MatIS in the future), so it might make sense to give it the lowest priority.

firedrake/mg/ufl_utils.py
pbrubeck: I am deferring the default value to whoever needs to decide it. But we might as well impose "aij" as the default.

firedrake/solving_utils.py
pbrubeck: How do we resolve circular dependencies if we need to import the type for type hinting?

pbrubeck: This might require further thought, I updated it to a numpydoc, but I do not plan to deal with the circular imports in this PR.

firedrake/mg/ufl_utils.py
```suggestion
        coarse_sub_mat_type = opts.getString(f"{solver_prefix}coarse_sub_mat_type",
                                             default_sub_mat_type) or None
```

firedrake/solving_utils.py
connorjward: This is quite a mess in Python, but what we need to do is put
```py
from __future__ import annotations
```
at the top of the file.

This will not be needed in Python 3.14+ (https://peps.python.org/pep-0749/)

firedrake/solving_utils.py
pbrubeck: I also added a `sub_pmat_type`. The new order is `mat_type, pmat_type, sub_mat_type, sub_pmat_type`.

connorjward (approved):

PR #4041, opened 2025-02-14
Jax ml operator fix

Ig-dolci:
# Description

Depends on https://github.com/firedrakeproject/ufl/pull/60

- Fixed an issue in `firedrake.ml.jax.ml_operator` where `argument_slots` was incorrectly specified. I wrote it as optional:

  ```python
  argument_slots: Optional[tuple[Union[ufl.coefficient.BaseCoefficient, ufl.argument.BaseArgument]]] = ()
  ```  
- Clarified that if `argument_slots` is not provided, `ML_Operator` will automatically write it.

- Test the results involving the Neural operators are in right function space.

- Test  jax and pytorch operators in Firedrake CI.


firedrake/ml/jax/ml_operator.py
```diff
@@ -90,7 +90,10 @@ def _pre_forward_callback(self, *operands: Union[Function, Cofunction], unsqueez
 
     def _post_forward_callback(self, y_P: "jax.Array") -> Union[Function, Cofunction]:
         """Callback function to convert the JAX output of the ML model to a Firedrake function."""
-        space = self.ufl_function_space()
+        if isinstance(self._argument_slots[0], Cofunction):
+            space = self.ufl_operands[0].function_space().dual()
+        else:
+            space = self.ufl_function_space()
```
pbrubeck, line 96: If `c = Cofunction(V.dual())` then `V == c.arguments()[0].ufl_function_space()`. Could this be used to not have `Cofunction` as a special case here?

Ig-dolci: I wrote this in another way. See the latest commit. I will be happy with your review.

firedrake/ml/pytorch/ml_operator.py
```diff
@@ -98,7 +102,12 @@ def _pre_forward_callback(self, *operands, unsqueeze=False):
 
     def _post_forward_callback(self, y_P):
         """Callback function to convert the PyTorch output of the ML model to a Firedrake function."""
-        space = self.ufl_function_space()
+        # At this point, ``len(self._argument_slots)`` must be greater than 0.
+        if isinstance(self._argument_slots[0], ufl.coefficient.BaseCoefficient):
```
pbrubeck, line 106: `Function` and `Cofunction` are both subclasses of `BaseCoefficient`. Don't we want different logic for these two cases?

Ig-dolci: `Function` inherits from `ufl.Coefficient` that inherits from `BaseCoefficient`. `Cofunction` inherits directly from `BaseCoefficient`. So, `isinstance(self._argument_slots[0], ufl.coefficient.BaseCoefficient)` returns `True` for both `Function` and `Cofunction`.

pbrubeck: So `self._argument_slots[0]` could be either a `Function`, `Cofunction`, `Argument` or `Coargument`? What do we want for each case: do we just want `V`, `V.dual()`, `V,` and `V.dual()`, respectively?

Ig-dolci: `self._argument_slots` contains either only `ufl.argument.BaseArgument` (`Argument` or `Coargument`) objects or a combination of `ufl.coefficient.BaseCoefficient` (at index 0) and `ufl.argument.BaseArgument` (at index 1). Our goal is to set the function space based on it. I think determining the function space based on `self._argument_slots[0]` is reasonable.

firedrake/external_operators/ml_operator.py
```diff
@@ -15,22 +15,26 @@ def __init__(self, *operands, function_space, derivatives=None, argument_slots=(
...
-                      Tuple specifiying the derivative multiindex.
+            Tuple specifiying the derivative multiindex.
         *argument_slots : ufl.coefficient.BaseCoefficient or ufl.argument.BaseArgument
-                          Tuple containing the arguments of the linear form associated with the ML operator,
-                          i.e. the arguments with respect to which the ML operator is linear. Those arguments
-                          can be ufl.Argument objects, as a result of differentiation, or ufl.Coefficient objects,
-                          as a result of taking the action on a given function.
+            Tuple containing the arguments of the linear form associated with the ML operator,
+            i.e. the arguments with respect to which the ML operator is linear. Those arguments can
+            retain only ``ufl.argument.BaseArgument`` objects, as a result of differentiation, or both
```
dham, line 26: contain?

Ig-dolci:
```suggestion
            contain only ``ufl.argument.BaseArgument`` objects, as a result of differentiation, or both
```

firedrake/ml/pytorch/ml_operator.py
```diff
@@ -98,7 +102,12 @@ def _pre_forward_callback(self, *operands, unsqueeze=False):
...
+            space = self._argument_slots[0].function_space()
+        else:
+            # When ``self._argument_slots[0]`` is an ``ufl.argument.BaseArgument``.
+            space = self._argument_slots[0].arguments()[0].function_space()
```
pbrubeck, line 110:
```suggestion
        space = self._argument_slots[0].arguments()[0].function_space()
```

pbrubeck, line 110: Would this also work? `A.arguments()[0].function_space()` only makes sense if we expect `A` to be `BaseForm`.
```suggestion
        space = self._argument_slots[0].function_space()
```

pbrubeck, line 110:
```suggestion
        argument = self._argument_slots[0]
        if isinstance(argument, ufl.Coargument):
            argument, = argument.arguments()
        space = argument.function_space()
```

pbrubeck: I think that `space` should be a property or method of the parent class `MLOperator`, so this logic does not have to be repeated in every subclass.

pbrubeck: This logic would go in UFL: https://github.com/FEniCS/ufl/pull/348

scripts/firedrake-install
```diff
@@ -1303,7 +1303,7 @@ def build_and_install_jax():
     """Install JAX for a CPU or CUDA backend."""
     log.info("Installing JAX (backend: %s)" % args.jax)
     version_name = "jax" if args.jax == "cpu" else "jax[cuda12]"
-    run_pip_install([version_name])
+    run_pip_install(version_name + ["jaxlib"] + ["ml_dtypes"] + ["opt_einsum"])
```
Ig-dolci, line 1306:
```suggestion
    run_pip_install([version_name] + ["jaxlib"] + ["ml_dtypes"] + ["opt_einsum"])
```

Ig-dolci:
@pbrubeck  @dham Firedrake pip installs PR depends on the approval of [UFL PR 348](https://github.com/FEniCS/ufl/pull/348) and this PR.

Ig-dolci:
This PR is failing in the docs building because of a recent Sphinx update.

tests/firedrake/external_operators/test_external_operators_adjoint.py
```diff
@@ -77,6 +77,6 @@ def J(f):
     c = Control(f)
     Jhat = ReducedFunctional(J(f), c)
 
-    f_opt = minimize(Jhat, tol=1e-6, method="BFGS")
+    f_opt = minimize(Jhat, tol=1e-4, method="BFGS")
```
Ig-dolci, line 80: This tolerance modification reduces the runtime test by almost 50%. Most important is that the following ``assert`` is still satisfied.

firedrake/ml/jax/ml_operator.py
```diff
@@ -39,36 +39,41 @@ def __init__(
             *operands: Union[ufl.core.expr.Expr, ufl.form.BaseForm],
             function_space: WithGeometryBase,
             derivatives: Optional[tuple] = None,
-            argument_slots: Optional[tuple[Union[ufl.coefficient.BaseCoefficient, ufl.argument.BaseArgument]]],
+            argument_slots: Optional[tuple[Union[ufl.coefficient.BaseCoefficient, ufl.argument.BaseArgument]]] = (),
```
connorjward, line 42: > If argument slots are not provided, then they will be generated.

Having an empty tuple seems a strange default considering. Could/should it be `None` instead?

connorjward: Actually if it is `Optional` then that's what is has to be (https://docs.python.org/3/library/typing.html#typing.Optional)

Ig-dolci: `Optional[tuple[Union[ufl.coefficient.BaseCoefficient, ufl.argument.BaseArgument]]] = None` leads to the error:
```
>       argument_slots = tuple(map(as_ufl, argument_slots))
E       TypeError: 'NoneType' object is not iterable

../ufl/ufl/core/base_form_operator.py:49: TypeError
```

Ig-dolci: The way to define `argument_slots` argument is comming from https://github.com/firedrakeproject/ufl/blob/master/ufl/core/external_operator.py#L28

Ig-dolci: We can improve this by writing 
```
argument_slots: tuple[Union[ufl.coefficient.BaseCoefficient, ufl.argument.BaseArgument]] | None = None,
```
Then
```
if argument_slots is None:
      argument_slots = ()
```

.github/workflows/build.yml
```diff
@@ -88,6 +88,7 @@ jobs:
             --install defcon \
             --install gadopt \
             --install asQ \
+            --package-branch ufl pbrubeck/fix/base-form-operator \
```
pbrubeck, line 91:
```suggestion
            --package-branch ufl pbrubeck/merge-upstream \
```

firedrake/ml/jax/fem_operator.py
```diff
@@ -92,7 +92,7 @@ def bwd(self, _, grad_output: "jax.Array") -> "jax.Array":
             adj_input = float(adj_input)
 
         # Compute adjoint model of `F`: delegated to pyadjoint.ReducedFunctional
-        adj_output = self.F.derivative(adj_input=adj_input)
+        adj_output = self.F.derivative(adj_input=adj_input, options={'riesz_representation': None})
```
pbrubeck, line 95: Is this equivalent to `'riesz_representation': 'l2'`?

firedrake/ml/jax/ml_operator.py
connorjward: I think it would be better to just have
```py
argument_slots: tuple[Union[ufl.coefficient.BaseCoefficient, ufl.argument.BaseArgument]] = (),
```
(i.e. no `Optional`) and to remove

> If argument slots are not provided, then they will be generated.

from the docstring.

That is at least consistent with the UFL object.

firedrake/ml/jax/fem_operator.py
Ig-dolci: In terms of data, yes. I think PyTorch and JAX work with derivatives rather than gradients. So, returning a derivative (`Cofunction`) instead of converting it to a `Function` makes more sense to me.

pbrubeck: Okay, so this is different, as it keeps `Cofunctions` as `Cofunctions`.

firedrake/ml/jax/ml_operator.py
Ig-dolci: Done.

connorjward (dismissed):
@pbrubeck if you are also happy it would be fantastic if this could be merged ASAP as it is holding up the pip PR. Naturally we need to update our UFL fork at the same time.

pbrubeck (approved):

PR #4042, opened 2025-02-14
Deadlock fixes for disk caches

### LINKED ISSUE #4039: BUG: Improved caching is parallel-unsafe
**Describe the bug**
The caching improvements introduced in #3989 possess a deadlock that is reasonably easy to reproduce in parallel environments.

**Steps to Reproduce**
I haven't tried to create a minimal reproducer for this, as we were seeing it with our testing suite timing out (https://github.com/g-adopt/g-adopt/actions/runs/13303073225). A reasonably reliable way to reproduce this is to run `g-adopt/tests/free_surface/implicit_top_bottom_buoyancy_free_surface.py` on 8 MPI cores. Importantly, `PYOP2_SPMD_STRICT` should be **disabled**, and it is easier to reproduce if there is a load imbalance between the ranks (maybe also with extra I/O activity?). I could not reproduce with all ranks dumping debug info, but making rank 0 sleep for 0.1s on `pyop2.logger.debug` was sufficient.

**Deadlock description**
The deadlock occurs during the code generation phase, and the caching surrounding it. On most ranks, we see a cache miss:
```
[0] pyop2:DEBUG PYOP2_COMM_WORLD R0, PYOP2_COMM_WORLD_DUP R0: key=(('9a82160d1c37b7c145f834e68bfb07f3',
    (<Access.INC: 4>, <Access.READ: 1>, <Access.READ: 1>, <Access.READ: 1>),
    (np:dtype('float64'), np:dtype('float64'), np:dtype('float64'), np:dtype('float64'))),
    (<class 'pyop2.global_kernel.DatKernelArg'>, (2,), (<class 'pyop2.global_kernel.MapKernelArg'>, 6, None, None), None),
    (<class 'pyop2.global_kernel.DatKernelArg'>, (2,), (<class 'pyop2.global_kernel.MapKernelArg'>, 3, None, None), None),
    (<class 'pyop2.global_kernel.DatKernelArg'>, (2,), (<class 'pyop2.global_kernel.MapKernelArg'>, 6, None, None), None),
    (<class 'pyop2.global_kernel.GlobalKernelArg'>, (np.int64(1),)), 0, 1, 0, False, False, False, False, None, False, 4) 
in cache: DictLikeDiskAccess cache miss
```

but on (at least) one rank, the sequencing means that it sees the file on disk and gets a cache hit:
```
[3] pyop2:DEBUG PYOP2_COMM_WORLD R3, PYOP2_COMM_WORLD_DUP R3: key=(('9a82160d1c37b7c145f834e68bfb07f3',
    (<Access.INC: 4>, <Access.READ: 1>, <Access.READ: 1>, <Access.READ: 1>),
    (np:dtype('float64'), np:dtype('float64'), np:dtype('float64'), np:dtype('float64'))),
    (<class 'pyop2.global_kernel.DatKernelArg'>, (2,), (<class 'pyop2.global_kernel.MapKernelArg'>, 6, None, None), None),
    (<class 'pyop2.global_kernel.DatKernelArg'>, (2,), (<class 'pyop2.global_kernel.MapKernelArg'>, 3, None, None), None),
    (<class 'pyop2.global_kernel.DatKernelArg'>, (2,), (<class 'pyop2.global_kernel.MapKernelArg'>, 6, None, None), None),
    (<class 'pyop2.global_kernel.GlobalKernelArg'>, (np.int64(1),)), 0, 1, 0, False, False, False, False, None, False, 4)
in cache: DictLikeDiskAccess cache hit
```

The caching improvements added the optimisation to only run the cached function on the root and broadcast the result to the other ranks
https://github.com/firedrakeproject/firedrake/blob/26a335ce719f309f4ee7e671adf3c96cda31d696/pyop2/caching.py#L519-L524

Because the disk cache is not synchronised across all the ranks, the variable cache hit means that not all ranks participate in the broadcast. Annoyingly for debug, this gets masked by further MPI operations on the hit rank. Eventually however, the ranks are in different states and hit a deadlock from MPI communication.

Even though `PYOP2_SPMD_STRICT` is meant to *detect* this, I think the synchronisation introduced by ensuring the cache key is consistent is sufficient to defeat it. It allows ranks that are "lagging behind" that would otherwise get a cache hit to attempt to read the cache value early enough that it doesn't exist. That being said, parallel ordering bugs are sneaky, and I think this could still lead to random failures, even in strict mode. Additionally, it introduces an extra 2 collectives per cache operation, plus many extra barriers, so there is possibly some performance overhead without the guarantee of correctness.

connorjward:
Possibly fixes #4039

I was able to produce two separate MFEs that are included as tests.

Fundamentally all I have added is the single line `cache_hit = comm.bcast(cache_hit, root=0)` to make sure that all ranks agree on whether or not a cache hit or miss has occurred. This only happens for disk caches where this is not guaranteed.

I would welcome peoples' thoughts on this as figuring this out was really difficult.

### DIFF
--- a/tests/pyop2/test_caching.py
+++ b/tests/pyop2/test_caching.py
@@ -829,7 +829,8 @@ class spmd_strict:
 
 @pytest.mark.parallel(2)
 @pytest.mark.parametrize("bcast", [False, True])
-def test_no_spmd_strict_disk_cache_obeys_spmd(bcast, tmpdir):
+@pytest.mark.parametrize("slow_rank", [0, 1])
+def test_no_spmd_strict_disk_cache_obeys_spmd(bcast, slow_rank, tmpdir):
     with spmd_strict(False):
         comm = COMM_WORLD
         # make sure the same tmpdir is used by all ranks
@@ -855,15 +856,15 @@ def test_no_spmd_strict_disk_cache_obeys_spmd(bcast, tmpdir):
         # Ensure ranks are synchronised at the start (so we can force them to not be)
         comm.barrier()
 
-        # Delay rank 1 such that, before it searches the cache, rank 0 has the chance
-        # to populate it (but it should wait!).
-        if comm.rank == 1:
+        # Delay one rank such that, before it searches the cache, the other rank has
+        # the chance to populate it (but it should wait!).
+        if comm.rank == slow_rank:
             time.sleep(1)
 
         result = cached_func()
-        assert result == expected
+        parallel_assert(lambda: result == expected)
 
-        # make sure that rank 1 has not 'skipped' a bcast from rank 0
+        # make sure that the slow rank has not missed a bcast from the other rank
         sent = "my message"
         with temp_internal_comm(comm) as tcomm:
             received = tcomm.bcast(sent, root=0)
@@ -872,7 +873,8 @@ def test_no_spmd_strict_disk_cache_obeys_spmd(bcast, tmpdir):
 
 @pytest.mark.parallel(4)
 @pytest.mark.parametrize("bcast", [False, True])
-def test_no_spmd_strict_disk_cache_race_condition(bcast, tmpdir):
+@pytest.mark.parametrize("slow_rank", [0, 1])
+def test_no_spmd_strict_disk_cache_race_condition(bcast, slow_rank, tmpdir):
     with spmd_strict(False):
         comm = COMM_WORLD
         # make sure the same tmpdir is used by all ranks
@@ -902,13 +904,25 @@ def test_no_spmd_strict_disk_cache_race_condition(bcast, tmpdir):
         # Ensure ranks are synchronised at the start (so we can force them to not be)
         comm.barrier()
 
-        # Force rank 0 of subcomm 0 to wait, and hence to register a cache hit
-        # whilst rank 1 of subcomm 0 will register a cache miss.
-        if color == 0 and subcomm.rank == 0:
+        # Force a hit on one rank of subcomm 0 and a miss on the other
+        if color == 0 and subcomm.rank != slow_rank:
+            # first to run, will miss the cache and should wait for the other rank
+            pass
+        elif color == 1:
+            # will populate the disk cache
+            time.sleep(0.5)
+        else:
+            # last to run, will have a cache hit
             time.sleep(1)
 
         result = cached_func()
-        assert result == expected
+        parallel_assert(lambda: result == expected)
+
+        # make sure that the slow rank has not missed a bcast from the other rank
+        sent = "my message"
+        with temp_internal_comm(subcomm) as tcomm:
+            received = tcomm.bcast(sent, root=0)
+        parallel_assert(lambda: received == sent)
 
 
 if __name__ == '__main__':


pyop2/caching.py
```diff
@@ -513,10 +513,36 @@ def wrapper(*args, **kwargs):
...
+                        #
+                        # Since broadcasting is non-blocking for the sending rank (rank 0)
+                        # it is possible for it to have written to disk before other ranks
+                        # begin the cache lookup. These ranks register a cache hit.
+                        #
+                        # If ranks disagree on whether it was a hit or miss then some ranks
+                        # will do a broadcast and others will not, ruining MPI synchronisation.
+                        # To fix this we make sure that in 'bcast' mode the hit/miss is
+                        # decided by rank 0.
+                        cache_hit = comm.bcast(cache_hit, root=0)
```
ksagiyam, line 538: I think this should be `comm.allreduce(cache_hit, op=MPI.LAND)`; if `cache_hit = True` on rank 0 and `cache_hit = False` on rank 1, `value` on rank 1 would never be updated (and causes some race condition at l.552).

ksagiyam: or maybe we could do:
```
root = rank if cache_hit else -1
root = comm.allreduce(root, op=MPI.MAX)
if root >=0:
    # bcast from root
```

connorjward: Ah yes of course. Excellent spot.

connorjward: I've done pretty much exactly what you've written here. Thanks so much!!

JHopeCollins: Would parameterising the test on which rank is forced to wait in the line below make sure that we cover both cases - (rank 0 hits, rank 1 misses) or (rank 0 misses, rank 1 hits). https://github.com/firedrakeproject/firedrake/blob/7f59707d05909a33fa938914c7c5d787ac9129cf/tests/pyop2/test_caching.py#L907

tests/pyop2/test_caching.py
```diff
@@ -812,5 +815,101 @@ def test_two_comms_compile_the_same_code():
...
+            comm_getter=lambda: subcomm,
+        )(func)
+
+        # Ensure ranks are synchronised at the start (so we can force them to not be)
+        comm.barrier()
+
+        # Force rank 0 of subcomm 0 to wait, and hence to register a cache hit
+        # whilst rank 1 of subcomm 0 will register a cache miss.
+        if color == 0 and subcomm.rank == 0:
+            time.sleep(1)
```
JHopeCollins, line 908: How do you know that (subcomm0, rank1) will consistently register a cache miss?
If (subcomm0, rank1) gets slightly delayed (by the OS process juggling or something), then couldn't subcomm1 create the cache before (subcomm0, rank1) reaches line 910? Then both ranks on subcomm0 would have a cache hit.

Could we attempt to "guarantee" a cache miss by forcing subcomm1 to also wait, just not as long as rank 0 of subcomm 0? Something like this:
```python
if color == 0 and subcomm.rank == 0:
    time.sleep(2)
elif color == 1:
    time.sleep(1)
result = cached_func()
```
So we could be basically certain that the `result = cached_func()` line would be hit in this order:

1. (subcomm 0, rank 1) - cache miss and wait for (subcomm0, rank 0)
2. subcomm 1, rank 1 and 2 (order doesn't matter) - cache miss and create
3. (subcomm0, rank 0) - cache hit, and syncs with (subcomm0, rank 0).

tests/pyop2/test_caching.py
```diff
@@ -812,5 +815,101 @@ def test_two_comms_compile_the_same_code():
...
+
+        result = cached_func()
```
JHopeCollins, line 910: How do you know that (subcomm0, rank1) will consistently register a cache miss?
If (subcomm0, rank1) gets slightly delayed (by the OS process juggling or something), then couldn't subcomm1 create the cache before (subcomm0, rank1) reaches line 910? Then both ranks on subcomm0 would have a cache hit.

Could we attempt to "guarantee" a cache miss by forcing subcomm1 to also wait, just not as long as rank 0 of subcomm 0? Something like this:
```python
if color == 0 and subcomm.rank == 0:
    time.sleep(2)
elif color == 1:
    time.sleep(1)
result = cached_func()
```
So we could be basically certain that the `result = cached_func()` line would be hit in this order:

1. (subcomm 0, rank 1) - cache miss and wait for (subcomm0, rank 0)
2. subcomm 1, rank 1 and 2 (order doesn't matter) - cache miss and create
3. (subcomm0, rank 0) - cache hit, and syncs with (subcomm0, rank 0).

pyop2/caching.py
connorjward: Sure. No harm in doing that.

tests/pyop2/test_caching.py
connorjward: Yep makes sense. Will do.

pyop2/caching.py
connorjward: Done.

tests/pyop2/test_caching.py
connorjward: Done.

PR #4048, opened 2025-02-18
mpi-pytest is now on pip

connorjward:
# Description

https://pypi.org/project/mpi-pytest/

Bonus aside: now I know more about versioning!

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -67,11 +67,11 @@ spydump = "pyop2.scripts.spydump:main"
 
 [project.optional-dependencies]
 test = [
-  "pylit",
+  "mpi-pytest",
   "nbval",
+  "pylit",
   "pytest",
   "pytest-xdist",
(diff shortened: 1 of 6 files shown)


francesco-ballarin:
Be careful that there is already a `pytest-mpi` package on pypi with somewhat similar goals, and that the parallel_assert
example seems to be using the other package.

connorjward:
> Be careful that there is already a `pytest-mpi` package on pypi with somewhat similar goals, and that the parallel_assert example seems to be using the other package.

This package was initially called `pytest-mpi`, which is why the Python package is called `pytest_mpi`. I think it is fairly common for the pip package to have a different name to the Python package so I didn't think the discrepancy mattered.

We have been using this package in some form for a long time and I don't think the other package would work for us.

Ig-dolci (approved):

Issue #4016, opened 2025-02-07
BUG: Projection unnecessarily takes a `quadrature_degree` argument


connorjward:
Since we pass `form_compiler_parameters` this is not needed. We should emit a warning if the deprecated approach is used.

See the discussion [here](https://github.com/firedrakeproject/firedrake/pull/4012#discussion_r1945098519).

PR #4073, opened 2025-02-26
Projector: deprecate quadrature_degree kwarg

pbrubeck:
# Description

This PR deprecates the  `quaddrature_degree` kwarg of `Projector`, added in https://github.com/firedrakeproject/firedrake/pull/3707. 

It turns out that this functionality was already possible via the `form_compiler_parameters`.

Fixes https://github.com/firedrakeproject/firedrake/issues/4016

### DIFF
--- a/firedrake/projection.py
+++ b/firedrake/projection.py
@@ -1,5 +1,6 @@
 import abc
 import ufl
+import warnings
 import finat.ufl
 from ufl.domain import extract_unique_domain
 from typing import Optional, Union
@@ -131,6 +132,13 @@ class ProjectorBase(object, metaclass=abc.ABCMeta):
         form_compiler_parameters=None, constant_jacobian=True,
         use_slate_for_inverse=True, quadrature_degree=None
     ):
+        if quadrature_degree is not None:
+            warnings.warn(f"Passing 'quadrature_degree' to {type(self).__name__} is deprecated, "
+                          "please instead set inside the form compiler parameters", FutureWarning)
+            if "quadrature_degree" in form_compiler_parameters:
+                raise ValueError("Cannot pass quadrature degree twice")
+            form_compiler_parameters = form_compiler_parameters.copy()
+            form_compiler_parameters["quadrature_degree"] = quadrature_degree
         if solver_parameters is None:
             solver_parameters = {}
         else:
@@ -163,7 +171,6 @@ class ProjectorBase(object, metaclass=abc.ABCMeta):
             is_variable_layers = True
         self.use_slate_for_inverse = (use_slate_for_inverse and is_dg and not is_variable_layers
                                       and (not complex_mode or SLATE_SUPPORTS_COMPLEX))
-        self.quadrature_degree = quadrature_degree
 
     @cached_property
     def A(self):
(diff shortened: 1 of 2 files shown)


connorjward (approved):
I think this looks absolutely fine. However this has all proved quite contentious so I would like someone else to agree with me @JHopeCollins?

firedrake/projection.py
```diff
@@ -174,19 +181,19 @@ def A(self):
...
+                    + firedrake.inner(u('+'), v('+'))*firedrake.dS_h
+                    + firedrake.inner(u('+'), v('+'))*firedrake.dS_v
                 )
             else:
                 a = (
-                    firedrake.inner(u, v)*firedrake.ds(degree=self.quadrature_degree)
-                    + firedrake.inner(u('+'), v('+'))*firedrake.dS(degree=self.quadrature_degree)
+                    firedrake.inner(u, v)*firedrake.ds
+                    + firedrake.inner(u('+'), v('+'))*firedrake.dS
                 )
```
JHopeCollins, line 194: I can't leave a suggestion because some lines are not modified, but (in my opinion) this would be cleaner by making it obvious that the if-statement is only on the measure, not the integrand.

```python
if F.extruded:
    ds = firedrake.ds_t + firedrake.ds_v + firedrake.ds_b
    dS = firedrake.dS_h + firedrake.dS_v
else:
    ds = firedrake.ds
    dS = firedrake.dS

a = (
    firedrake.inner(u, v)*ds
    + firedrake.inner(u('+'), v('+'))*dS
)
```

firedrake/projection.py
```diff
@@ -245,20 +252,20 @@ def rhs_form(self):
...
+                    + firedrake.inner(firedrake.avg(self.source), firedrake.avg(v))*firedrake.dS_h
+                    + firedrake.inner(firedrake.avg(self.source), firedrake.avg(v))*firedrake.dS_v
                 )
             else:
                 form = (
-                    firedrake.inner(self.source, v)*firedrake.ds(degree=self.quadrature_degree)
-                    + firedrake.inner(firedrake.avg(self.source), firedrake.avg(v))*firedrake.dS(degree=self.quadrature_degree)
+                    firedrake.inner(self.source, v)*firedrake.ds
+                    + firedrake.inner(firedrake.avg(self.source), firedrake.avg(v))*firedrake.dS
                 )
```
JHopeCollins, line 265: Pulling the definition of the form outside the if-statement as above.

JHopeCollins (changes requested):
The code looks fine to me, I just have a suggestion about tidying up some of the forms.

However, one of the slate tests isn't passing. I can't see how this PR would make that test fail, and I can't get it to fail locally, so I've restarted that workflow in case it was one of those freak CI "global state??" errors.

If it passes then I'm happy for this to merge.

pbrubeck:
Thanks for the sugestions @JHopeCollins, I think we should be able to get rid of the special casing here and in many other places by adding proper support for `ds` and `dS` on extruded meshes. 

> The contention was about a PR doing multiple things that didn't have to go in at the same time, rather than about what the code itself was doing.

In line with this let's keep the PR as is and move forward. The refactoring that you propose or the new feature that I propose merit their own separate PR.

> The code looks fine to me, I just have a suggestion about tidying up some of the forms.
> 
> However, one of the slate tests isn't passing. I can't see how this PR would make that test fail, and I can't get it to fail locally, so I've restarted that workflow in case it was one of those freak CI "global state??" errors.

This SLATE test (testing a wrong PETSc option: '-fieldsplit_type frog') is known to randomly fail ocassionally.

JHopeCollins:
> Thanks for the suggestions @JHopeCollins, I think we should be able to get rid of the special casing for ExtrudedMesh here and in many other places by enabling proper support for `ds` and `dS` on extruded meshes.

That would be great, we have this special case in a few places, and I know that Gusto do too.

> In line with this let's keep the PR as is and move forward. The refactoring that you propose or the new feature that I propose merit their own separate PR.

Allowing `dS` and `ds` to naturally handle extruded meshes should certainly go in a new PR. What's the reason for making a new PR for the small simplification suggested though? It just tidies up a little bit of code that is already modified by this PR. 

> This SLATE test (testing a wrong PETSc option: '-fieldsplit_type frog') is known to randomly fail ocassionally.

Ok, I've set the job running again. Once it passes we can approve and merge. I think you'll have to merge master into this branch to get the docs passing first though.

danshapero:
I've had problems where it was helpful to use different quadrature degrees for different terms in a variational form. Does this change make that impossible?

pbrubeck:
> I've had problems where it was helpful to use different quadrature degrees for different terms in a variational form. Does this change make that impossible?

This PR does not change that. Passing the quadrature degree to the form_compiler_parameters only sets it for in integrals that do not specify it. This PR does not change API.

firedrake/projection.py
JHopeCollins: Can go in a separate PR.

dham (approved):

PR #4076, opened 2025-02-27
Use A.petscmat as the preferred way of accessing the PETSc.Mat

pbrubeck:
# Description
There are two equivalent ways of accessing the underlying `PETSc.Mat` of a `firedrake.MatrixBase`. These are `A.M.handle` and `A.petscmat`, in fact `A.petscmat is A.M.handle`.  This PR favours `A.petscmat` over `A.M.handle` for the sake of clarity. 

`A.M` is the underlying pyop2 tensor, but for generic `AssembledMatrices`, `A.M` is just a `NameSpace` that emulates the pyop2 tensor.  The pyop2 tensor can be useful in some cases, for example it provides convenient access to each block of a MatNest via `A.M[i, j]`, but in most cases it is more convenient and clear to access `A.petscmat` directly.

This PR does not change API, it is just to improve clarity.


connorjward (approved):
LGTM

PR #4077, opened 2025-02-27
pip fixups

connorjward:
* Put SLEPc into `firedrake-vanilla` containers
* Fix URL in install.rst


.github/workflows/docs.yml
connorjward, line 1: I checked and this passed before the latest commit (so is fine to go in even though it's failing)

PR #4080, opened 2025-03-01
Add simple G-ADOPT test

angus-g:
# Description

Similar to the other downstream tests, this should just be a simple integration to make sure nothing is massively broken. Other g-adopt folks could chime in with useful tests that are a bit more complex in terms of code paths exercised than the simple base case (or perhaps this could be driven from the g-adopt side, and the test here in Firedrake will just run the minimal suite).

- [x] Requires https://github.com/g-adopt/g-adopt/pull/178

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -166,6 +166,15 @@ jobs:
             gusto-repo/integration-tests/transport/test_embedded_dg_advection.py
         timeout-minutes: 10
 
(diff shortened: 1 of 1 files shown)


connorjward (approved):
Thanks! Once you've merged things your side I'm very happy to put this in.

angus-g:
Okay, I think we're all good on our side! Hopefully the tests still pass :grin:

connorjward (approved):

PR #4083, opened 2025-03-02
Cleanup __init__.py to be robust to not detecting BLAS

connorjward:
# Description

A user reported an error in `get_blas_library()` which we only use to attempt to set the number of threads. I believe that that should not be a fatal error so it no longer is.

I have also done a little cleanup in `__init__.py`. I think using functions means we can avoid accidentally including things in the namespace that aren't wanted.


firedrake/__init__.py
```diff
@@ -33,38 +33,11 @@
...
-    pass
-del ufl
 from ufl import *
 from finat.ufl import *
 
-# By default we disable pyadjoint annotation.
-# To enable annotation, the user has to call continue_annotation().
-import pyadjoint
-pyadjoint.pause_annotation()
-del pyadjoint
```
dham, line 66: Is this really right?

firedrake/__init__.py
```diff
@@ -29,38 +29,11 @@
...
-#     only from it.
-#
-# To get around this, since the default __del__ on Expr is just
-# "pass", we just remove the method from the definition of Expr.
-import ufl
-try:
-    del ufl.core.expr.Expr.__del__
-except AttributeError:
-    pass
-del ufl
```
JHopeCollins, line 54: Do we not need this anymore?

connorjward: Nope. From https://stackoverflow.com/questions/41851098/how-does-del-interfere-with-garbage-collection:

> Note that I link to the Python 2 documentation. As of Python 3.4, this is no longer true and even circular dependencies as show in the example will be cleared, as [PEP 442 – Safe object finalization](https://www.python.org/dev/peps/pep-0442/) has been implemented:

JHopeCollins: Ok great

PR #4082, opened 2025-03-01
Test HiptmairPC + PMGPC

pbrubeck:
# Description
Adding a high order test for HiptmairPC + PMGPC.

PR #4086, opened 2025-03-03
Remove CI test annotations

connorjward:
I am proposing that we remove the https://github.com/mikepenz/action-junit-report step from our CI. The step is there to annotate the "Files changed" tab with information about the failing tests but it creates another GitHub check called "Firedrake default" or "Firedrake complex" that gets associated with the wrong workflow ([example](https://github.com/firedrakeproject/firedrake/pull/4082/checks?check_run_id=38136954898)) ([issue](https://github.com/mikepenz/action-junit-report/issues/40)).

I think it makes things more confusing when it is supposed to do the opposite.

### DIFF
--- a/tests/firedrake/regression/test_assemble.py
+++ b/tests/firedrake/regression/test_assemble.py
@@ -86,6 +86,8 @@ def M(fs):
 
 
 def test_one_form(M, f):
+    # to test the new action, obvs undo me
+    assert False
     one_form = assemble(action(M, f))
     assert isinstance(one_form, Cofunction)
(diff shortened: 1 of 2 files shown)


tests/firedrake/regression/test_assemble.py
```diff
@@ -86,6 +86,8 @@ def M(fs):
 
 
 def test_one_form(M, f):
+    # to test the new action, obvs undo me
+    assert False
```
connorjward, line 90:
```suggestion
```

dham (approved):

Issue #4087, opened 2025-03-03
BUG: Unable to import slepc4py in docker containers


jrmaddison:
**Describe the bug**
SLEPc seems to be built but not installed.

**Steps to Reproduce**
In firedrake:latest, firedrake-vanilla-default:latest, or firedrake-vanilla-complex:latest
```
>>> import slepc4py
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
ModuleNotFoundError: No module named 'slepc4py'
```

PR #4091, opened 2025-03-03
Install slepc4py into vanilla container and tidy up test suite

connorjward:
We were previously skipping SLEPc tests because the skip statements were not conditional on CI running.

Fixes https://github.com/firedrakeproject/firedrake/issues/4087

# Description

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -58,7 +58,7 @@ jobs:
           python3 ../firedrake-repo/scripts/firedrake-configure \
             --arch ${{ matrix.arch }} --show-petsc-configure-options | \
             xargs -L1 ./configure --with-make-np=12
-          make
+          make PETSC_DIR=${{ github.workspace }}/petsc PETSC_ARCH=arch-firedrake-${{ matrix.arch }}
           make check
           echo "PETSC_DIR=${{ github.workspace }}/petsc" >> "$GITHUB_ENV"
           echo "PETSC_ARCH=arch-firedrake-${{ matrix.arch }}" >> "$GITHUB_ENV"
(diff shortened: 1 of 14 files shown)


pyproject.toml
```diff
@@ -76,13 +76,31 @@ test = [
...
   "mpi-pytest",
   "nbval",
+  "ngsPETSc",
   "pylit",
   "pytest",
   "pytest-split",  # needed for firedrake-run-split-tests
   "pytest-timeout",
   "pytest-xdist",
+  "slepc4py",
+  "torch",  # requires passing '--extra-index-url' to work
```
dham, line 90: Is this going to stop us putting firedrake on PyPI? I think they don't allow off-site dependencies. Not sure if that rule extends to optional dependencies.

connorjward: I don't think so. I think we just need to tell users to pass an extra flag to `pip install` if they want to install this set of optional dependencies.

More generally I am pretty unhappy with the growing number of optional dependencies/different configurations and would like to clean it up at some point.

.github/workflows/build.yml
```diff
@@ -58,9 +58,19 @@ jobs:
...
+          echo "PETSC_DIR=/__w/firedrake/firedrake/petsc" >> "$GITHUB_ENV"
+          echo "PETSC_ARCH=arch-firedrake-${{ matrix.arch }}" >> "$GITHUB_ENV"
+
+
+      - name: Install SLEPc
+        run: |
+          git clone --depth 1 https://github.com/firedrakeproject/slepc.git
+          cd slepc
+          ./configure
+          make SLEPC_DIR=/__w/firedrake/firedrake/slepc
```
JHopeCollins, line 72:
```suggestion
          make SLEPC_DIR=/__w/firedrake/firedrake/slepc
          make check
```

dham (approved):

PR #4093, opened 2025-03-03
Make SLATE subclass BaseForm

pbrubeck:
# Description

Slate tensors cannot be used with `fieldsplit`. On `main`, `slate.TensorBase` is a
plain object that imitates `ufl.Form` closely enough to be assembled: a class
attribute `integrals` hands back a `MockCellIntegral`, and rebuilding a tensor with
new arguments goes through `Form.__call__`. Splitting a problem defeats both, because
`_SNESContext.split` and `PCSNESBase.form` reach for UFL algorithms that the mock
cannot answer.

This PR makes `TensorBase` a subclass of `ufl.BaseForm`. Slate nodes expose
`ufl_operands` and `_ufl_expr_reconstruct_`, so `ufl.replace` traverses and rebuilds
them, and those two call sites use `ufl.replace` in place of `Form.__call__`.
`MockCellIntegral` is removed.

`Block` now composes instead of nesting: a `Block` of a `Block` becomes a single
`Block` on the underlying terminal, and a `Block` of a non-terminal is pushed down
through `Add`, `Negative`, `Transpose` and `Mul` until it reaches one.
`ExtractSubBlock.split` returns a `Block` for a Slate argument of any rank, and
recovers a Slate expression from a `FormSum` whose components are Slate. UFL builds
such a `FormSum` whenever a tensor meets a `BaseForm` from the left, and a Slate node
cannot be walked as a UFL DAG.

`ProxyFunctionSpace.collapse()` returns a plain `FunctionSpace`. UFL hashes terminals
by their `repr`, which reports the proxy metadata, while `FunctionSpace.__eq__`
ignores it. A proxy argument therefore compared equal to its plain counterpart but
hashed differently, breaking the argument lookups that splitting relies on.

## Tests

`test_cg_poisson` solves the same problem through both SCPC and a Schur `fieldsplit`
whose second block is geometric multigrid.

`test_stabilized_stokes` is new: equal-order CG Stokes, stabilized by a local
projection onto the discontinuous pressure space written in Slate, solved with an
upper-triangular Schur fieldsplit.

https://claude.ai/code/session_017jqk3KoVszP5xgpq6bQcj7


firedrake/solving_utils.py
```diff
@@ -384,6 +383,7 @@ def split(self, fields):
                 F += problem.compute_bc_lifting(J, subu)
             else:
                 F = replace(F, {problem.u_restrict: u})
+
```
pbrubeck, line 386:
```suggestion
```

firedrake/solving_utils.py
```diff
@@ -375,6 +373,7 @@ def split(self, fields):
             # coefficients in the new form.
             u = as_vector(vec)
             J = replace(J, {problem.u_restrict: u})
+
```
pbrubeck, line 376:
```suggestion
```

pbrubeck:
Ideally we would want SLATE classes to inherit from `BaseForm`, so that we can properly support `ufl.replace` and `ufl.derivative` on them.

tests/firedrake/slate/test_cg_poisson.py
```diff
@@ -28,49 +81,43 @@ def run_CG_problem(r, degree, quads=False):
     f = -div(grad(u_exact))
 
     # Set up function spaces
-    e = FiniteElement("Lagrange", cell=mesh.ufl_cell(), degree=degree, variant="integral")
+    e = FiniteElement("Lagrange", cell=mesh.ufl_cell(), degree=degree)
```
pbrubeck:
```suggestion
    e = FiniteElement("Lagrange", cell=mesh.ufl_cell(), degree=degree, variant="integral")
```

firedrake/adjoint_utils/variational_solver.py
```diff
@@ -27,7 +27,7 @@ def wrapper(self, *args, **kwargs):
                     # Try again without expanding derivatives,
                     # as dFdu might have been simplied to an empty Form
                     self._ad_adj_F = adjoint(dFdu, derivatives_expanded=True)
-            except (ValueError, TypeError, NotImplementedError):
+            except (ValueError, TypeError, NotImplementedError, AttributeError):
```
connorjward: I think we should be avoiding this pattern. `AttributeErrors` get thrown for all sorts of reasons.

firedrake/slate/slate.py
```diff
@@ -128,13 +119,10 @@ class TensorBase(object, metaclass=ABCMeta):
...
     """
 
-    integrals = MockCellIntegral()
-    """A mock object that provides enough compatibility with ufl.Form
-    that one can assemble a tensor."""
-
     terminal = False
     assembled = False
     diagonal = False
+    _initialised = False
```
connorjward: More readable if this lives in the subclasses where it is actually used

firedrake/slate/slate.py
```diff
@@ -972,8 +1075,21 @@ class TensorOp(TensorBase):
...
+            # object with the original arguments. Skip re-running the
+            # constructor so we don't corrupt it (e.g. into referencing
+            # itself).
+            return
         super(TensorOp, self).__init__()
         self.operands = tuple(operands)
+        self._initialised = True
+
+    def reconstruct(self, *operands):
+        """Reconstructs this TensorBase with new operands."""
```
connorjward:
```suggestion
        """Reconstructs this TensorOp with new operands."""
```

firedrake/slate/slate.py
```diff
@@ -972,8 +1075,21 @@ class TensorOp(TensorBase):
...
+        return type(self)(*operands)
```
connorjward: I'd implement this for `BinaryOp` instead of `TensorOp`. That way you can replace `A` or `B`. Also these should probably be kwarg-only:
```py
def reconstruct(self, *, A=None, B=None):
  ...

File: firedrake/functionspaceimpl.py [hunk 8]
```diff
@@ -1303,6 +1303,25 @@ def __str__(self):
...
+        Returns
+        -------
+        FunctionSpace
+            An unindexed space equal to this one.
+
+        Notes
+        -----
+        The proxy metadata (``index``, ``component``, ``parent``) is what
+        collapsing discards, and it is reported by ``__repr__``. Since UFL
+        hashes terminals by their ``repr`` while ``FunctionSpace.__eq__``
```
connorjward: What if we made the repr of a proxy function space the same as a regular function space?

firedrake/slate/slate.py
```diff
@@ -1443,10 +1565,21 @@ def as_slate(F):
...
         return Tensor(F)
     elif isinstance(F, (Function, Cofunction)):
         return AssembledVector(F)
+    elif isinstance(F, FormSum):
+        # UFL builds a FormSum whenever a Slate tensor is combined with a
+        # BaseForm from the left, since TensorBase cannot claim the reflected
+        # operator. Recover the equivalent Slate expression. Slate has no
+        # scalar multiplication, so only unit weights convert.
+        if any(abs(w) != 1 for w in F.weights()):
+            raise TypeError("Cannot convert a weighted FormSum into a slate.Tensor")
```
connorjward: This could be made more clear

firedrake/slate/slate.py
```diff
@@ -653,25 +656,84 @@ def __new__(cls, tensor, indices):
...
+            return
         super(Block, self).__init__()
-        self.operands = (tensor,)
         indices = tuple(map(as_tuple, indices))
+        self.operands = (tensor,)
         self._blocks = dict(enumerate(indices))
         self._indices = indices
+        self._initialised = True
+
+    def reconstruct(self, tensor, indices=None):
```
connorjward:
```suggestion
    def reconstruct(self, *, tensor=None, indices=None):
```

firedrake/slate/slate.py
```diff
@@ -653,25 +656,84 @@ def __new__(cls, tensor, indices):
...
+                return -Block(A, indices)
+            if isinstance(tensor, Transpose):
+                A, = tensor.operands
+                return Block(A, indices[::-1]).T
+            if isinstance(tensor, Mul) and len(indices) == 2 and tensor.operands[0].rank == 2 and tensor.operands[1].rank == 2:
+                A, B = tensor.operands
+                row, col = indices
+                full_col_A = tuple(range(len(A.arguments()[1].function_space())))
+                full_row_B = tuple(range(len(B.arguments()[0].function_space())))
+                return Block(A, (row, full_col_A)) * Block(B, (full_row_B, col))
```
connorjward, line 704:
```py
else:
    raise NotImplementedError(...)
```
?

connorjward (changes requested):

firedrake/slate/slate.py
```diff
@@ -1214,6 +1361,112 @@ def __repr__(self):
         return "%s(%r, %r)" % (type(self).__name__, A, B)
 
 
+class ScalarMul(BinaryOp):
```
pbrubeck, line 1364: Should this be `UnaryOp` or `BinaryOp`? The scalar is not really a TensorBase

pbrubeck: `Negative` should probably inherit from this class...

firedrake/slate/slate.py
```diff
@@ -1443,10 +1702,15 @@ def as_slate(F):
...
         return F
-    elif isinstance(F, Form):
+    elif isinstance(F, (Form, ZeroBaseForm)):
         return Tensor(F)
     elif isinstance(F, (Function, Cofunction)):
         return AssembledVector(F)
+    elif isinstance(F, FormSum):
+        return functools.reduce(
+            operator.add,
+            (ScalarMul(w, as_slate(c))
```
pbrubeck, line 1712:
```suggestion
            (w * as_slate(c)
```

firedrake/slate/slate.py
pbrubeck: This is just eagerly simplifying Block(Tensor), we need the fallback to express things that we can already express

connorjward: Sure OK

tests/firedrake/slate/test_assemble_tensors.py
```diff
@@ -142,6 +142,19 @@ def test_assemble_vector_into_tensor(mesh):
     assert np.allclose(f.dat.data, 2*assemble(Tensor(conj(v) * dx)).dat.data, rtol=1e-14)
 
 
+def test_assemble_scalar_multiplication(mesh):
+    V = FunctionSpace(mesh, "DG", 1)
+    v = TestFunction(V)
+    tensor = Tensor(v * dx)
```
pbrubeck:
```suggestion
    tensor = Tensor(conj(v) * dx)
```

firedrake/slate/slate.py
pbrubeck: I added a ScalarMul node, and this lets us convert a FormSum to slate.

firedrake/adjoint_utils/variational_solver.py
pbrubeck: Root cause was addressed by raising NotImplementedError in the right place

firedrake/functionspaceimpl.py
```diff
@@ -1326,13 +1330,8 @@ def __new__(cls, function_space, boundary_set=frozenset(), name=None):
...
 
     def __repr__(self):
-        return "%sProxyRestrictedFunctionSpace(%r, name=%r,  boundary_set=%r, index=%r, component=%r)" % \
-            (str(self.identifier).capitalize(),
-             str(self.function_space),
-             self.name,
-             self.boundary_set,
-             self.index,
-             self.component)
+        return "RestrictedFunctionSpace(%r, name=%r, boundary_set=%r)" % (
```
connorjward, line 1333: f string.

tests/firedrake/slate/test_slate_infrastructure.py
```diff
@@ -319,16 +321,46 @@ def test_implicit_casting_add_sub():
     r = Tensor(f)
     assert b + f == b + r
     assert b - f == b - r
-    assert f + b == r + b
-    assert f - b == r - b
+    # A TensorBase is a BaseForm, so UFL claims the reflected operator and
+    # returns a FormSum; as_slate recovers the equivalent Slate expression.
+    assert as_slate(f + b) == r + b
+    assert as_slate(f - b) == r - b
```
connorjward, line 327: I'm not sure this test makes any sense any more. It is there to test that Slate objects behave like UFL ones. But now they are the same. `as_slate` seems pointless now. It is definitely a confusing name because we already have a Slate object.

connorjward (changes requested):

firedrake/functionspaceimpl.py
connorjward: I don't want to be catching these things from the AI. You know that I will flag this so the fact that it's still in there suggests that you haven't read the code very closely yourself.

pbrubeck: I read this code, and I genuinely missed the f-strings.

I see two issues with the formatting standards: 
1. We say in our policy that we enforce a standard but our inconsistent codebase does not reflect it
2. Only a human can be enforcing the policy partially, and this drives code review into pointless discussions.

Something so trivial as an f-string can be easily covered by a linter, and LLMs are extremely good at fixing lint errors. Please have a look at https://github.com/firedrakeproject/firedrake/pull/5483

tests/firedrake/slate/test_slate_infrastructure.py
pbrubeck: I disagree. `as_slate` still returns a DAG made entirely of TensorBase nodes. The slate compiler only accepts TensorBase nodes.

connorjward: Not for this PR, but I assume that would be a straightforward thing to change given the work in this PR? I think `as_slate` is now very confusing in nature and we should look to drop it.

If you agree please open an issue and then this thread can be resolved.

firedrake/functionspaceimpl.py
connorjward: I'm not in the slightest bit bothered by the fact that an f-string got missed by the AI. The concern is that you missed it. This could just as easily have been anything else. The fact that this happened tells me that during review I have to really scrutinise every little piece of this PR in case there is anything else that isn't right.

firedrake/slate/slate.py
```diff
@@ -143,6 +133,8 @@ def __init__(self, *_):
 
         Mirrors :class:`~ufl.form.Form`.
         """
+        BaseForm.__init__(self)
```
connorjward:
```suggestion
        super().__init__()
```

firedrake/slate/slate.py
```diff
@@ -157,6 +149,15 @@ def _metakernel_cache(self):
...
         return self.operands
 
+    @property
+    def ufl_operands(self):
+        return self.operands
+
+    def _ufl_expr_reconstruct_(self, *operands):
+        if len(operands) == 0:
+            return self
+        return self.reconstruct(*operands)
```
connorjward: Should we look to deprecate this?

tests/firedrake/slate/test_slate_infrastructure.py
pbrubeck: I don't agree, we still need to treat slate specially because these are compilable forms. Slate can only wrap Forms that are hybridizable/local to a cell. We cannot feed any BaseForm into the slate compiler, such as cross-mesh interpolate.

connorjward: At some point we should unify the two compilers no? Appreciate it's out of scope.

firedrake/slate/slate.py
```diff
@@ -800,6 +879,10 @@ def __init__(self, tensor, decomposition=None):
         self.operands = (tensor,)
         self.decomposition = decomposition
 
+    def reconstruct(self, tensor, decomposition=None):
+        """Reconstructs this TensorBase with new operands."""
```
connorjward:
```suggestion
        """Reconstructs this Factorization with new operands."""
```
or just remove the docstring altogether because it's pretty obvious

firedrake/slate/slate.py
```diff
@@ -1145,32 +1281,96 @@ def _output_string(self, prec=None):
...
+    @cached_property
+    def _key(self):
+        """Return a key for hash and equality."""
+        return (type(self), self.scalar, self.operands)
+
+    def __repr__(self):
+        """Return the Slate representation of the scalar multiplication."""
         tensor, = self.operands
-        return tensor.arguments()
+        return "%s(%r, %r)" % (type(self).__name__, self.scalar, tensor)
```
connorjward: f string

firedrake/slate/slate.py
```diff
@@ -1145,32 +1281,96 @@ def _output_string(self, prec=None):
...
+        return par(result)
+
+
+class Negative(ScalarMul):
+    """Abstract Slate class representing the negation of a tensor object."""
+
+    def __new__(cls, tensor):
+        return super().__new__(cls, -1, tensor)
+
+    def __init__(self, tensor, _tensor=None):
```
connorjward: Why does this constructor take two tensor arguments?

connorjward (changes requested):

firedrake/slate/slate.py
pbrubeck: `ScalarMul(-1, tensor)` eagerly simplifies to `Negative(tensor) `in `ScalarMul.__new__`. The second kwarg is there to because `ScalarMul.__new__` takes two arguments and might get recursively called.  

I think I should entirely remove `Negative` in this PR. The simplification is not buying us anything, we would generate the same code anyway.

pbrubeck: I removed the Negative class and left a factory function with the same name for backwards-compatibility. 

This is too low level, and I don't think we should be deprecating the old class. Until now, Slate has not been user-facing API, and only accessible through `HybridizationPC` and friends.

tests/firedrake/slate/test_slate_infrastructure.py
pbrubeck: I am not sure how much else we could unify. Slate already sits on top of TSFC, which means that we basically get all unifications in TSFC for free, for example `Tensor(Form(Interpolate))` and `Tensor(Interpolate)` will happen for free.

I think the current split in slac/TSFC is the correct one, unless you have a counterexample in mind.

firedrake/slate/slate.py
pbrubeck: I don't think so. The situation in slate is tricky. There are two DAGs, one with broader ufl types, and another with slate nodes that is visited by the slate compiler.

firedrake/slate/slate.py
pbrubeck: done

firedrake/slate/slate.py
pbrubeck: done

firedrake/slate/slate.py
pbrubeck: `TensorBase.reconstruct` is introduced in this PR. We need it because slate constructors take non-ufl kwargs such as the block indices and the factorization type, and `_ufl_expr_reconstruct_`  only deals with ufl operands.

firedrake/slate/slate.py
pbrubeck: I removed this eager simplification in favour of the `push_block` traverser, which has the same purpose, but was missing the dispatch `Mul` that I added here.

firedrake/formmanipulation.py
```diff
@@ -89,6 +90,16 @@ def split(self, form, argument_indices):
             assert (len(idx) == 1 for idx in self.blocks.values())
             assert (idx[0] == 0 for idx in self.blocks.values())
             return form
+
+        if isinstance(form, FormSum) and any(isinstance(c, slate.slate.TensorBase)
+                                             for c in form.components()):
```
pbrubeck:
```suggestion
        if isinstance(form, FormSum) and has_type(form, slate.slate.TensorBase):
```

firedrake/formmanipulation.py
```diff
@@ -3,7 +3,7 @@
 import collections
 
 from ufl import as_tensor, as_vector, split
-from ufl.classes import Form, Zero, FixedIndex, ListTensor, ZeroBaseForm
+from ufl.classes import Form, FormSum, Zero, FixedIndex, ListTensor, ZeroBaseForm
```
pbrubeck:
```suggestion
from ufl.classes import Form, FormSum, Zero, FixedIndex, ListTensor, ZeroBaseForm
from ufl.algorithms.analysis import has_type
```

firedrake/mg/ufl_utils.py
```diff
@@ -491,8 +491,9 @@ def reconstruct_slate_tensor(tensor, self, coefficient_mapping=None):
 
 @_reconstruct.register(firedrake.slate.TensorOp)
 def reconstruct_slate_tensor_op(tensor, self, coefficient_mapping=None):
-    children = (self(c, self, coefficient_mapping=coefficient_mapping) for c in tensor.children)
-    return type(tensor)(*children)
+    children = tuple(self(c, self, coefficient_mapping=coefficient_mapping)
+                     for c in tensor.children)
```
```suggestion
    children = (self(c, self, coefficient_mapping=coefficient_mapping) for c in tensor.children)
```

Issue #4096, opened 2025-03-05
BUG: test failure on tests/pyop2/test_vector_map.py


qbisi:
**Describe the bug**
test failure on tests/pyop2/test_vector_map.py
cause my c compiler cannot convert a pointer to a multidimesion array

**Steps to Reproduce**
python -m pytest -n auto tests/pyop2/test_vector_map.py -m "not parallel or parallel[1]"

**Expected behavior**
Compilation command:
mpicc -fPIC -Wall -std=gnu11 -I/nix/store/rsxs1ldvlb78dcya46gyrnmbwixkjz1j-petsc-3.22.3/include -I/home/qbisi/firedrake/pyop2 -march=native -O3 -ffast-math -o /tmp/pyop2-tempcache-uid1000/d3c57b/1d/tmpz6789ow3.so /tmp/pyop2-tempcache-uid1000/d3c57b/1d/tmpz6789ow3.c -shared -L/nix/store/rsxs1ldvlb78dcya46gyrnmbwixkjz1j-petsc-3.22.3/lib -Wl,-rpath,/nix/store/rsxs1ldvlb78dcya46gyrnmbwixkjz1j-petsc-3.22.3/lib -lpetsc -lm

**Error message**
warning: Skipping impure flag -march=native because NIX_ENFORCE_NO_NATIVE is set
/tmp/pyop2-tempcache-uid1000/d3c57b/1d/tmpz6789ow3.c: In function ‘wrap_k’:
/tmp/pyop2-tempcache-uid1000/d3c57b/1d/tmpz6789ow3.c:23:23: warning: unused variable ‘i7’ [-Wunused-variable]
   23 |         int32_t const i7 = 0;
      |                       ^~
/tmp/pyop2-tempcache-uid1000/d3c57b/1d/tmpz6789ow3.c:20:21: warning: unused variable ‘i6’ [-Wunused-variable]
   20 |       int32_t const i6 = 0;
      |                     ^~
/tmp/pyop2-tempcache-uid1000/d3c57b/1d/tmpz6789ow3.c:29:7: error: passing argument 1 of ‘k’ from incompatible pointer type [-Wincompatible-pointer-types]
   29 |     k(&(t0[0]));
      |       ^~~~~~~~
      |       |
      |       int32_t * {aka int *}
[...]

PR #4099, opened 2025-03-05
fix test pyop2 kernel compilation error with gcc14

qbisi:
# Description
Since gcc14, some -Wincompatible-pointer-types are treated as error,
especially when you define func (int a[2][3]) {...}
and pass func with a pointer int p[6].

This pr might not be the best solution, however, one should fix the op2 kernel code in tests/pyop2/test_vector_map.py. e.g.
```
    k = """
    static void k(int d[2], int vd[1][2]) {
    d[0] = vd[0][0];
    d[1] = vd[0][1];
    }"""
```
to 
```
    k = """
    static void k(int d[2], int vd[2]) {
    d[0] = vd[0];
    d[1] = vd[1];
    }"""
```
at the cost of less human readability.

related issue: https://github.com/firedrakeproject/firedrake/issues/4096

I am a nixpkgs commiter on the roadmap to build/test/package firedrake/petsc with the power of nix build system which claim a reproducible firedrake environment.

Glad to see the Modernization of the installation of firedrake and its component.

### DIFF
--- a/pyop2/global_kernel.py
+++ b/pyop2/global_kernel.py
@@ -389,6 +389,7 @@ class GlobalKernel:
         cppargs = [f"-I{d}/include" for d in get_petsc_dir()]
         cppargs.extend(f"-I{d}" for d in self.local_kernel.include_dirs)
         cppargs.append(f"-I{os.path.abspath(os.path.dirname(__file__))}")
+        cppargs.append("-Wno-incompatible-pointer-types")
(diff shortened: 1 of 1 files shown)


connorjward:
Thank you for this. However I think the solution is to simply go through the PyOP2 test suite (`tests/pyop2`) and fix the cases where this happens. I'm fairly sure that we don't do this anywhere else.

PR #4119, opened 2025-03-17
Use the `skipslepc` pytest mark on more tests

JHopeCollins:
# Description

Tests using SLEPc should be marked with the `skipslepc` marker so that a) they are skipped for most users when they don't have SLEPc but b) they aren't skipped by CI.

This PR just adds this marker to tests that got missed previously and were sniffing SLEPc themselves.

### DIFF
--- a/tests/firedrake/conftest.py
+++ b/tests/firedrake/conftest.py
@@ -81,10 +81,22 @@ def _skip_test_dependency(dependency):
         raise ValueError("Unrecognised dependency to check: {dependency = }")
 
 
+dependency_skip_markers_and_reasons = (
+    ("mumps", "skipmumps", "MUMPS not installed with PETSc"),
+    ("hypre", "skiphypre", "hypre not installed with PETSc"),
+    ("slepc", "skipslepc", "SLEPc is not installed"),
+    ("pytorch", "skiptorch", "PyTorch is not installed"),
+    ("jax", "skipjax", "JAX is not installed"),
+    ("matplotlib", "skipplot", "Matplotlib is not installed"),
+    ("netgen", "skipnetgen", "Netgen and ngsPETSc are not installed"),
+    ("vtk", "skipvtk", "VTK is not installed")
+)
+
+
 # This allows us to check test dependencies within tests e.g. the demo tests
 @pytest.fixture
 def skip_dependency():
-    return _skip_test_dependency
+    return _skip_test_dependency, dependency_skip_markers_and_reasons
 
 
 def pytest_configure(config):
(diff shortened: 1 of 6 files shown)


connorjward (approved):

tests/firedrake/conftest.py
```diff
@@ -10,6 +10,83 @@
...
+def _skip_test_dependency(dependency):
+    """
+    Returns whether to skip tests with a certain dependency.
+
+    Usually, this will return True if the dependency is not available.
+    However, on CI we never want to skip tests because we should
+    test all functionality there, so if the environment variable
+    FIREDRAKE_CI=1 then this will always return False.
+    """
+    skip = True
```
connorjward, line 22: This reads quite strangely. Why not just have things like:
```py
if dependency == "slepc":
        try:
            from slepc4py import SLEPc  # noqa: F401
            del SLEPc
            return False
        except ImportError:
            return True
```
instead?

tests/firedrake/conftest.py
```diff
@@ -10,6 +10,83 @@
 from firedrake.petsc import PETSc, get_external_packages
 
 
+def _skip_test_dependency(dependency):
```
connorjward, line 13: Might it make more sense to rename this "`dependency_is_present(dep)`" or similar?

tests/firedrake/demos/test_demos_run.py
```diff
@@ -72,39 +70,32 @@ def test_no_missing_demos():
...
-def _maybe_skip_demo(demo):
+def _maybe_skip_demo(demo, skip_dependency):
     # Add pytest skips for missing imports or packages
-    if "mumps" in demo.requirements and "mumps" not in get_external_packages():
+
+    if "mumps" in demo.requirements and skip_dependency("mumps"):
         pytest.skip("MUMPS not installed with PETSc")
 
-    if "hypre" in demo.requirements and "hypre" not in get_external_packages():
+    if "hypre" in demo.requirements and skip_dependency("hypre"):
```
connorjward, line 79: Could we turn this all into a loop:
```py
for dep in demo.requirements:
    if skip_dependency(dep):
        pytest.skip(reason=f"{dep} is not available")
```

connorjward (changes requested):
Thanks for taking this on.

tests/firedrake/conftest.py
JHopeCollins: This is related to your comment below. Because there is a switch for if we are on CI, we specifically are not asking "is this dependency present", but we are asking "should we skip tests with this dependency or not".

The `skip = True` line does read a bit strangely, but (IMO) it makes the lines lower down read more clearly. Because you would usually be asking if a dependency exists, `except ImportError: return True` is a bit odd at first glance, whereas `except ImportError: return skip` reads more like pseudocode.

Would adding a comment to the `skip = True` line be satisfactory?

tests/firedrake/conftest.py
JHopeCollins: See above.

tests/firedrake/demos/test_demos_run.py
JHopeCollins: Probably should be. I think I didn't because the messages don't always line up with the dependency string, but that's not really a big problem.

tests/firedrake/conftest.py
connorjward: I see. You've convinced me. Happy to leave as is.

tests/firedrake/demos/test_demos_run.py
connorjward: I think having a loop is better because then we can't accidentally add a requirement to a demo and then forget to add a check for it.

JHopeCollins: Done

connorjward (dismissed):

connorjward (approved):

Issue #4097, opened 2025-03-05
DOCS: slepc installation is undocumented


dham:
The install docs don't tell users what to do if they need slepc.

The use of LinearEigenSolver also needs to be checked to ensure the error message is helpful if slepc is not available.

Issue #4098, opened 2025-03-05
DOCS: The netgen demo has outdated install instructions for netgen.


dham:
The demo still refers to firedrake-install.

PR #4121, opened 2025-03-17
Clarify install docs

connorjward:
# Description

* Removes some references to `firedrake-install`
* Makes `pip cache remove petsc4py` a main instruction so it isn't missed
* Moves complex bits about customisation to the bottom of the install instructions
* Included explicit information about installing SLEPc and other optional dependencies

Fixes https://github.com/firedrakeproject/firedrake/issues/4097 and https://github.com/firedrakeproject/firedrake/issues/4098

Note that the docs link check is expected to fail because it references a link that doesn't yet exist.

### TODOs

- [x] Point ARM users to https://github.com/scientificcomputing/vtk-aarch64 to get ARM wheels for VTK (`--extra-index-url`?)


docs/source/install.rst
```diff
@@ -294,40 +274,167 @@ should run::
...
+---------------------
+
+PETSc with 64-bit indices
+~~~~~~~~~~~~~~~~~~~~~~~~~
+
+To build PETSc with 64-bit indices you need to append ``--with-64-bit-indices``
+to the set of options when PETSc is configured. For example this could be::
+
+   $ python3 ../firedrake-configure --show-petsc-configure-options | xargs -L1 ./configure --with-64-bit-indices
+
```
JHopeCollins, line 311:
```suggestion
Some of the packages that PETSc uses must be configured specially for 64-bit indices. If PETSc downloads a package (i.e. `--download-package` is in the PETSc configure options) then it will ensure that the package is configure correctly.
However, if a package has been installed externally (i.e. `--with-package` or `--with-package-{lib,include}=...` are in the PETSc configure options), then it is your responsibility to either a) ensure that it is configured correctly for 64-bit integers, or b) replace the PETSc configure option with `--download-package` so that PETSc handles the configuration.

```

docs/source/install.rst
```diff
@@ -294,40 +274,167 @@ should run::
    whatever packages you just installed.
```
JHopeCollins, line 275:
```suggestion

.. note::
   The command given to clone the PETSc repository in the :ref:`Installing PETSc<install_petsc>` section include the `--depth 1` argument, which forces git to only clone the most recent commit. This massively reduces the amount of data that must be downloaded, but also means that the entire git history is missing. If you are developing PETSc then you may want to use a larger depth in this step, or remove this argument entirely.
```

docs/source/install.rst
```diff
@@ -294,40 +274,167 @@ should run::
...
+
+* Append additional options when ``configure`` is invoked. For example::
+
+   $ python3 ../firedrake-configure --show-petsc-configure-options | xargs -L1 ./configure --download-exotic-package
+
+* Write the output of ``firedrake-configure`` to a file than can be modified. For example::
+
+   $ python3 ../firedrake-configure --show-petsc-configure-options > my_configure_options.txt
+   <edit my_configure_options.txt>
+   $ ./configure $(cat my_configure_options.txt)
```
JHopeCollins, line 378:
```suggestion

For better reproducibility of PETSc builds, PETSc allows you to write a python script to record the configuration options and call configure. Example scripts can be found in the `config/examples/` directory of the PETSc repository.
```

JHopeCollins (changes requested):
All good improvements. I just have a few suggestions for the PETSc installation details

rckirby (commented):
While clarifying install docs, perhaps we should add `git+https` to the suggested cloning commands for developer-mode install.  Currently, it is http-only and one can't go over ssh/is prompted to supply GitHub credentials if one has followed the commands too slavishly.

rckirby (dismissed):
We need to make sure downstream packages installed in developer mode get installed via `git`

docs/source/install.rst
connorjward: I don't know if this is an essential detail. It's just another way to achieve the same thing as
```
$ ./configure $(cat myoptions.txt)
```
And I don't want to overload the user with information

docs/source/install.rst
```diff
@@ -268,15 +221,42 @@ Firedrake with its optional test dependencies by specifying the ``[test]``
...
+
+   $ cd /path/to/petsc
+   $ git pull
+   $ make
+
+Note that this will only recompile PETSc's source code, not that of the external
+packages, and so should be relatively quick. If your PETSc is sufficiently
+out-of-date you may also need to rebuild the external packages by running::
+
+   $ make reconfigure
```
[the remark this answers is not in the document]
connorjward: I don't know if that's the right thing to do. If users are following the rest of the instructions then they will only have a single ARCH per PETSc installation. If they are going "off piste" enough to have additional ARCHs then maybe we don't need to mention this?

docs/source/install.rst
connorjward: I'll mention something like this. I want to figure out the right git magic to undo `--depth 1`.

docs/source/install.rst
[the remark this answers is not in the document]
JHopeCollins: Fair enough, I guess if someone has built multiple archs they must have had to modify the DIR and ARCH anyway so should know enough.

docs/source/install.rst
JHopeCollins: How about just reducing it to something smaller like:
"You can see examples of different PETSc configurations in the `petsc/config/examples` directory. Be aware that these will not necessarily have all requirements for Firedrake"
I do think it's good to direct to the config python scripts if people are modifying the configuration themselves.

docs/source/install.rst
```diff
@@ -294,40 +274,167 @@ should run::
...
+~~~
+
+To install Firedrake with VTK, it should be installed using the ``vtk`` optional
+dependency. For example::
+
+   $ pip install --no-binary h5py "firedrake @ git+https://github.com/firedrakeproject/firedrake.git#[test,vtk]"
 
-   $ make reconfigure
+
+PyTorch
```
connorjward, line 341: Missing jax

docs/source/install.rst
connorjward: I've added something.

docs/source/install.rst
connorjward: Actually I will add this in https://github.com/firedrakeproject/firedrake/pull/4193 because using a custom PETSc is quite complicated.

docs/source/install.rst
connorjward: I've included something very similar to this

rckirby (approved):
But please delete the petsc/slepc version changes

connorjward:
Forcing this through as it was previously approved and is holding up other PRs.

PR #4126, opened 2025-03-19
Tape subfunctions that were created before annotating begins

JHopeCollins:
`Function.subfunctions` use `pyadjoint.FloatingType` to create `Blocks` on the fly when they are accessed that record the connection between the full and sub spaces.
However, because subfunctions is a cached property, they may be created _before_ annotations are enabled, but used _after_ annotations are enabled. In this case, we still need the subfunctions to know how to create the relevant Blocks, even though usually we don't do any adjoint-y things unless annotations are currently switched on. 

The test checks that this is done correctly by creating a `Function` and forcing `subfunctions` to be instantiated before annotations are switched on. If the `FloatingType` isn't created correctly then using _both_ the full `Function` and the `subfunction` when recording the tape will create a disconnected tape.
We check this isn't happening by evaluating `ReducedFunctional.derivative` and making sure that the  functional is dependent on the control.

First commit only adds the test and is expected to fail.
Second commit adds the fix and is expected to pass.

### DIFF
--- a/tests/firedrake/adjoint/test_split_and_subfunctions.py
+++ b/tests/firedrake/adjoint/test_split_and_subfunctions.py
@@ -172,17 +172,16 @@ def test_subfunctions_always_create_blocks():
     We check this isn't happening by evaluating ReducedFunctional.derivative and making
     sure that the functional is dependent on the control.
     """
-    pause_annotation()
-    mesh = UnitIntervalMesh(1)
-    R = FunctionSpace(mesh, "R", 0)
+    with stop_annotating():
+        mesh = UnitIntervalMesh(1)
+        R = FunctionSpace(mesh, "R", 0)
 
-    u = Function(R).assign(4)
-    kappa = Function(R).assign(2.0)
-    control = Control(kappa)
+        u = Function(R).assign(4)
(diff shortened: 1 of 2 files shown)


tests/firedrake/adjoint/test_split_and_subfunctions.py
```diff
@@ -152,3 +152,43 @@ def test_merge_blocks():
...
+    mesh = UnitIntervalMesh(1)
+    R = FunctionSpace(mesh, "R", 0)
+
+    u = Function(R).assign(4)
+    kappa = Function(R).assign(2.0)
+    control = Control(kappa)
+
+    # force the subfunctions to be created
+    _ = kappa.subfunctions
+    continue_annotation()
```
dham, line 185:
```suggestion
    with stop_annotating():
        mesh = UnitIntervalMesh(1)
        R = FunctionSpace(mesh, "R", 0)

        u = Function(R).assign(4)
        kappa = Function(R).assign(2.0)
        control = Control(kappa)

        # force the subfunctions to be created
        _ = kappa.subfunctions
```

dham (changes requested):

dham (approved):

PR #4128, opened 2025-03-19
Raise an error if `bounds` are passed to a variational solver while annotations are active.

JHopeCollins:
We don't know how to handle bounds constraints with the adjoint, so we should raise an error if someone attempts this.

Because this pops the `bounds` from `kwargs`, this also prevents a bug where if `bounds=None` was passed to `solve`, it was erroneously passed through to the constructor of the adjoint solvers.

### DIFF
--- a/firedrake/adjoint_utils/variational_solver.py
+++ b/firedrake/adjoint_utils/variational_solver.py
@@ -65,6 +65,11 @@ class NonlinearVariationalSolverMixin:
             from firedrake import LinearVariationalSolver
             annotate = annotate_tape(kwargs)
             if annotate:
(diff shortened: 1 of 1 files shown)


colinjcotter (approved):
I agree that it is a good idea to raise an exception in this case until we work out the right thing to do. There will definitely be a right approach in the control theory literature.

rckirby (approved):
I'm fine with punting on this for now.  I don't understand the right thing, and the paper on bounds-constrained RK hasn't even been finished yet...

colinjcotter:
eg section 7.10 of Bryson and Ho

PR #4130, opened 2025-03-19
adjoint variational solver should get solver parameters from the variational solver not the `solver_parameters` kwarg

JHopeCollins:
The adjoint decorating for variational solver was grabbing the solver parameters twice: once by the kwarg to `__init__` and once by `self.parameters` in the `solve` wrapper, and passing both to the solve block.

The one grabbed in the `solve` wrapper is correct because it has been processed by `OptionsManager`. The kwarg to `__init__` has not been processed by `OptionsManager` so could be `None`, which causes errors when the solve block tries to process the parameters.

### DIFF
--- a/firedrake/adjoint_utils/variational_solver.py
+++ b/firedrake/adjoint_utils/variational_solver.py
@@ -48,6 +48,8 @@ class NonlinearVariationalSolverMixin:
             self._ad_problem = problem
             self._ad_args = args
             self._ad_kwargs = kwargs
+            # We'll get the parameters directly from self when `solve` is called.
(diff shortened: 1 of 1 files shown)


rckirby (approved):
This sounds right.  Do we have confirmation that this fixes the problem?  (E.g. a test here or in Irksome?

JHopeCollins:
The Irksome adjoint tests that pass locally if the Stepper is given a `solver_parameters={}` also pass if given `solver_parameters=None`. I can't do it on CI yet because they need the subfunctions PR as well.

dham (approved):

PR #4137, opened 2025-03-21
Switch to using upstream PETSc and SLEPc instead of our fork

connorjward:
## TODOs
- [x] Store PETSc version in firedrake-configure (or pyproject.toml? single source of truth?)
- [x] Make sure that CI passes
- [ ] Think about Zenodo


.github/workflows/build-mac.yml
```diff
@@ -45,7 +45,7 @@ jobs:
 
       - name: Install PETSc
         run: |
-          git clone --depth 1 https://github.com/firedrakeproject/petsc.git
+          git clone --depth 1 --branch v3.22.2 https://gitlab.com/petsc/petsc.git
```
francesco-ballarin, line 48: FYI, 3.22.4 was released a couple of weeks ago

connorjward: I know. Our current fork is on 3.22.2 so I want to do the switch in two stages: first switch to upstream and then update to the latest release.

Then we will need to start thinking about which versions of PETSc Firedrake `main` (we will rename `master`) and `release` point to when we begin doing releases.

scripts/firedrake-configure
```diff
@@ -39,6 +39,12 @@ ARCH_DEFAULT = FiredrakeArch.DEFAULT
 ARCH_COMPLEX = FiredrakeArch.COMPLEX
 
 
+# NOTE: When updating this variable a corresponding change must be made inside
+# pyproject.toml
+SUPPORTED_PETSC_VERSION = "v3.22.2"
+SUPPORTED_SLEPC_VERSION = SUPPORTED_PETSC_VERSION
```
JHopeCollins, line 45: Does SLEPc always track PETSc versions?

pyproject.toml
```diff
@@ -20,7 +20,8 @@ dependencies = [
   "mpi4py; python_version < '3.13'",
   "h5py>3.12.1",
   "libsupermesh",
-  "petsc4py",
+  # NOTE: If changing the PETSc version then firedrake-configure also needs changing
```
JHopeCollins, line 23:
```suggestion
  # NOTE: If changing the PETSc version then firedrake-configure also needs changing for both petsc4py and slepc4py versions.
```

JHopeCollins (changes requested):
This is brilliant.

Just a couple of extremely minor comments but I'm very very happy with this!

JHopeCollins (approved):
Great, thanks for doing this!

Issue #4143, opened 2025-03-21
BUG: LinearSolver convergence failures discards results


jrmaddison:
**Describe the bug**
If a linear solver failure is encountered and a `firedrake.exceptions.ConvergenceError` is raised, the solution value is unchanged.

**Steps to Reproduce**
```
from firedrake import *
import petsc4py.PETSc as PETSc

mesh = UnitIntervalMesh(10)
space = FunctionSpace(mesh, "CG", 1)
test, trial = TestFunction(space), TrialFunction(space)

solver = LinearSolver(assemble(inner(trial, test) * dx),
                      solver_parameters={"ksp_type": "cg",
                                         "pc_type": "none",
                                         "ksp_max_it": 1,
                                         "ksp_atol": 1.0e-2})
b = assemble(inner(Constant(1), test) * dx)

u = Function(space, name="u")

u.assign(-1)
try:
  solver.solve(u, b)
except firedrake.exceptions.ConvergenceError:
    assert solver.ksp.getConvergedReason() == PETSc.KSP.ConvergedReason.DIVERGED_MAX_IT
    print("ConvergenceError")
print(u.dat.data_ro)
```

**Expected behavior**
The solution should be updated to the state obtained from the linear solve -- in the above after 1 CG iteration.

**Additional Info**
I think the behaviour changed in https://github.com/firedrakeproject/firedrake/pull/4012

pbrubeck:
Oh, I see the issue. The `ConvergenceError` prevents the copy of the internal solution vector to the one provided as argument.

jrmaddison:
There's maybe a question whether this _should_ update the result. I have one unit test which was relying on this behaviour, but I think a custom KSP convergence test can be used instead.

Issue #4142, opened 2025-03-21
BUG: LinearSolver always uses a non-zero initial guess


jrmaddison:
**Describe the bug**
The `ksp_initial_guess_nonzero` parameter is ignored by `LinearSolver`.

**Steps to Reproduce**c
```
from firedrake import *

mesh = UnitIntervalMesh(10)
space = FunctionSpace(mesh, "CG", 1)
test, trial = TestFunction(space), TrialFunction(space)

solver = LinearSolver(assemble(inner(trial, test) * dx),
                      solver_parameters={"ksp_type": "preonly",
                                         "pc_type": "jacobi",
                                         "ksp_max_it": 1,
                                         "ksp_initial_guess_nonzero": False})
b = assemble(inner(Constant(1), test) * dx)

u = Function(space, name="u")

u.assign(0)
solver.solve(u, b)
print(u.dat.data_ro)

u.assign(1)
solver.solve(u, b)
print(u.dat.data_ro)
```
displays
```
[1.5 1.5 1.5 1.5 1.5 1.5 1.5 1.5 1.5 1.5 1.5]
[1. 1. 1. 1. 1. 1. 1. 1. 1. 1. 1.]
```

**Expected behavior**
The non-zero initial value should be ignored in the second case, the linear solver should use a zero initial guess, and the results of the two solves should be the same.

jrmaddison:
I think this can be closed?

PR #4144, opened 2025-03-21
LinearSolver: fix zero initial guess and update after error

pbrubeck:
# Description
Fixes https://github.com/firedrakeproject/firedrake/issues/4142
Fixes https://github.com/firedrakeproject/firedrake/issues/4143


tests/firedrake/regression/test_linear_solver.py
```diff
@@ -0,0 +1,80 @@
...
+
+    u = Function(space, name="u")
+    u.assign(-1)
+    uinit = Function(u, name="uinit")
+    try:
+        solver.solve(u, b)
+    except firedrake.exceptions.ConvergenceError:
+        assert solver.ksp.getConvergedReason() == PETSc.KSP.ConvergedReason.DIVERGED_MAX_IT
+
+    assert not numpy.allclose(u.dat.data_ro, uinit.dat.data_ro)
```
jrmaddison, line 50: `u.dat.data_ro` is zero here, i.e. the zero initial guess, rather than the result from one CG iteration

jrmaddison: I think it's zero, as `ksp_initial_guess_nonzero` defaults to `False`.

pbrubeck:
`LinearSolver` used to just wrap a KSP, but it now became a `NonlinearVariationalSolver`, so we need to decide whether we want to preserve the old behavior or to accomodate to the usual way of providing initial guesses to NLVS and always use whatever is stored in the solution.  

`ksp_initial_guess_nonzero` applies only to the Newton update step, that stores the update as an internal `PETSc.Vec`, which is reused across different `solve()` calls, so the default `ksp_initial_guess_nonzero=False` makes sense here.

jrmaddison:
I didn't realize that was the behaviour. Does that mean a `LinearVariationalSolver` always uses a non-zero initial guess for the (single step) Newton solve, and then zero guesses the increment solve? i.e. `ksp_initial_guess_nonzero` has no effect for a `LinearVariationalSolver`?

firedrake/linear_solver.py
```diff
@@ -83,7 +83,13 @@ def solve(self, x, b):
...
             raise ValueError(f"b must be a Cofunction in {self.b.function_space()}.")
 
-        self.x.assign(x)
         self.b.assign(b)
-        super().solve()
-        x.assign(self.x)
+        if self.ksp.getInitialGuessNonzero():
+            self.x.assign(x)
+        else:
+            self.x.zero()
```
pbrubeck, line 90:
```suggestion
        self.x.assign(x)
```

tests/firedrake/regression/test_linear_solver.py
```diff
@@ -0,0 +1,80 @@
...
+    u1 = Function(space, name="u1")
+    u1.assign(0)
+    solver.solve(u1, b)
+
+    u2 = Function(space, name="u2")
+    u2.assign(1)
+    solver.solve(u2, b)
+    assert numpy.allclose(u1.dat.data_ro, u2.dat.data_ro)
+
+
```
pbrubeck, line 28:
```suggestion
```

pbrubeck:
> I didn't realize that was the behaviour. Does that mean a `LinearVariationalSolver` always uses a non-zero initial guess for the (single step) Newton solve, and then zero guesses the increment solve? 
This is correct

> i.e. `ksp_initial_guess_nonzero` has no effect for a `LinearVariationalSolver`?
The effect is very subtle, and is only for the update solve, `ksp_initial_guess_nonzero=True` will use the previous update vector as the initial guess for the current update vector.

pbrubeck:
Upon discussing this during the meeting, `LinearSolver` will use as initial guess whatever is passed to `solve(x, b)` to be consistent with NLVS and LVS. So this PR will only deal with alwats updating the solution, even when an `Exception` is raised.

jrmaddison:
> Upon discussing this during the meeting, `LinearSolver` will use as initial guess whatever is passed to `solve(x, b)` to be consistent with NLVS and LVS. So this PR will only deal with alwats updating the solution, even when an `Exception` is raised.

Sounds good, thanks.

PR #4153, opened 2025-03-26
RestrictedFunctionSpace and LinearSolver

pbrubeck:
# Description
This PR enables `LinearSolver(..., restrict=True)`


pbrubeck:
`LinearSolver` only takes in assembled matrices, which will live on the unrestricted space. It is not clear whether we want to restrict post-assembly.

PR #4168, opened 2025-03-28
Update adjoint documentation.

dham:
Introduce documentation in the manual for adjoint solves. Known limitations:

* no TLM or Hessian, I will come back to this in the future.
* no mention of Riesz maps. I am deliberately ducking this until we fix it in May/June.

This sits on top of #4173 so there is a lot of website stuff in the diff. That will vanish presently. `adjoint.rst` is the guts of this PR.

### DIFF
--- a/docs/source/adjoint.rst
+++ b/docs/source/adjoint.rst
@@ -1,0 +1,426 @@
+.. default-role:: math
+
+Solving adjoint PDEs
+====================
+
+Mathematical background
+-----------------------
+
+Suppose we have a parametrised real finite element problem, given `m \in M` find
+`u\in V` such that:
+
+.. math::
+    :label: eq:forward
+
+    f(u, m; v) = 0 \qquad \forall v \in V.
+
+Further suppose that we have some scalar quantity of interest `J(u, m) \in
+\mathbb{R}`. We call `m` the *control*, `J` the *functional* and `u` the
+*state*. We can initially assume that both `V` and `M` are finite element
+spaces.
+
+Our objective is to compute `\frac{\mathrm{d}J}{\mathrm{d}m}`. We can assume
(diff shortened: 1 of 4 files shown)


docs/source/adjoint.rst
```diff
@@ -0,0 +1,426 @@
+.. default-role:: math
+
+Solving adjoint PDEs
+====================
+
+Mathematical background
+-----------------------
+
+Suppose we have a parametrised finite element problem, given `m \in M` find
+`u\in V` such that:
```
jrmaddison, line 10: Maybe note limited to real early

tests/firedrake/adjoint/test_burgers_newton.py
```diff
@@ -197,6 +197,68 @@ def test_burgers_newton(solve_type, checkpointing, basics):
...
+@pytest.mark.skipcomplex
+def test_burgers_newton_docs():
+    """This test exists to ensure that the adjoint documentation runs."""
+    get_working_tape().clear_tape()
+    # start solver
+    n = 30
+    mesh = UnitIntervalMesh(n)
+    end = 0.3
+    timestep = Constant(1.0/n)
+    steps = int(end/float(timestep)) + 1
```
jrmaddison, line 209: Usually better to avoid computing `int`s from `float`s

docs/source/adjoint.rst
```diff
@@ -0,0 +1,426 @@
+.. default-role:: math
+
+Solving adjoint PDEs
+====================
+
+Mathematical background
+-----------------------
+
+Suppose we have a parametrised finite element problem, given `m \in M` find
```
dham, line 9:
```suggestion
Suppose we have a parametrised real finite element problem, given `m \in M` find
```

PR #4172, opened 2025-03-31
excise deleted link

dham:
@ksagiyam I couldn't see where this link should now point to so I'm proposing deleting it to get master building again. If you can find the target then please go ahead.


ksagiyam (approved):
I couldn't see either. I think it is fine for now.

PR #4173, opened 2025-03-31
Dham/random website updates

dham:
# Description

This is a bunch of website updates that mostly occur because nobody commits after having built the documentation. In theory we should probably go through and have less of this in the repo, but that is for another day.


connorjward (approved):
LGTM. I noticed these popping up in `git status` but never really knew the cause.

PR #4175, opened 2025-03-31
Do not install petsc4py using pip

connorjward:
This PR is a first stab at an idea I've had to avoid what I think may be an upcoming versioning issue. Things are very rough but I think there's enough there to see what I mean.

### The problem

PETSc and petsc4py must be installed using the exact same version but they are installed using different code paths. One from source and one from PyPI.

This is fine for most users currently because we fix the specific PETSc that they use, but I think this means we might hit issues with system installs of PETSc, both for HPC installs and a possible future `apt` package. I can easily imagine someone trying to install the latest Firedrake using a slightly out-of-date PETSc and thus having a mismatch between PETSc and petsc4py versions.

### The solution

I think the way to fix this is to use a petsc4py installed alongside PETSc (`--with-petsc4py=1`) and we can tweak `sys.path` to find it (using `PETSC_DIR` and `PETSC_ARCH`). Then we have single source of truth for the version.

Feedback welcome.

---

As another benefit I think we are realising that petsc4py install isn't very robust as it's the cause of most of our current install problems.

### TODOs

- [ ] Add numpy hack to the docs
- [ ] Fixup Dockerfile

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -20,9 +20,6 @@ dependencies = [
   "mpi4py; python_version < '3.13'",
   "h5py>3.12.1",
   "libsupermesh",
-  # NOTE: If changing the PETSc/SLEPc version then firedrake-configure also needs
-  # changing (as well as other references to petsc4py and slepc4py here)
-  "petsc4py==3.22.2",
   "numpy",
   "packaging",
   "pkgconfig",
@@ -84,7 +81,6 @@ ci = [
   "pytest-split",  # needed for firedrake-run-split-tests
   "pytest-timeout",
   "pytest-xdist",
-  "slepc4py==3.22.2",
   "torch",  # requires passing '--extra-index-url' to work
   "vtk",
 ]
@@ -98,7 +94,6 @@ docker = [  # Used in firedrake-vanilla container
   "pytest-split",  # needed for firedrake-run-split-tests
   "pytest-timeout",
   "pytest-xdist",
-  "slepc4py==3.22.2",
 ]
 # Dependencies needed to build the docs
 docs = [
@@ -134,7 +129,6 @@ requires = [
   "pkgconfig",
   "pybind11",
   "setuptools>61.2",
-  "petsc4py==3.22.2",
   "rtree>=1.2",
 ]
 build-backend = "setuptools.build_meta"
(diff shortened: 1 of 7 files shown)


pyproject.toml
```diff
@@ -20,11 +20,9 @@ dependencies = [
   "mpi4py; python_version < '3.13'",
   "h5py>3.12.1",
   "libsupermesh",
-  # NOTE: If changing the PETSc/SLEPc version then firedrake-configure also needs
-  # changing (as well as other references to petsc4py and slepc4py here)
-  "petsc4py==3.22.2",
```
francesco-ballarin, line 25: Hiding `petsc4py`/`slepc4py` dependencies into `petsctools` feels wrong, considering how important `petsc4py`/`slepc4py` are, and how likely it is that `petsctools` will only get updated once in a while (i.e., not at every upstream release)

connorjward: `petsctools` does not have petsc4py listed as a dependency in the `pyproject.toml`. I think that installing petsc4py through PyPI is not a particularly good idea as it means that we cannot assert that the PETSc and petsc4py versions match. For instance, as we just discussed on another thread, if we want to follow PETSc `main` then petsc4py cannot be included in the `pyproject.toml` because the right version is not available on PyPI. As another example, PETSc have released v3.23.0 but the corresponding petsc4py is not yet available.

`petsctools` is basically a shim package that means that petsc4py can be installed through PETSc and still found, guaranteeing that the versions match.

connorjward:
@dham  made the excellent point that doing things this way means that the petsc4py wheel is tied to the system Python, which is easily not the same as the one used to install Firedrake.

This work came out of a desire to ensure compatibility between PETSc and petsc4py. I think it is sufficient to warn the user if this is the case and hence permit petsc4py inside the pyproject.toml.

For installs using `main` I think we will have to use `--no-build-isolation`, which I think is fine.

francesco-ballarin:
> For installs using main I think we will have to use --no-build-isolation, which I think is fine

Yes, you'll definitely need to disable build isolation, because petsc4py from pypi doesn't track main.

connorjward:
Closing as the spirit of the PR is changing substantially so this will just get confusing. I'll open another.

PR #4177, opened 2025-03-31
Should the subfunction merge block apply a mask on the adjoint evaluation?

JHopeCollins:
### Background

Adjoining subfunctions is hard because it has to know that it doesn't hold its own data, it just views the full function. This means that when a subfunction is used, it also needs to add the full function onto the tape as a dependency so that the correct data is used.
The `pyadjoint.FloatingType` base class does with two custom blocks:
- When a subfunction `usub = u.subfunctions[i]` for component `i` of a full function `u` is added as a dependency to block `block0` on the tape, what actually happens is:
	1. a `SubfunctionBlock` is added to the tape with `u` as a dependency and `usub` as an output. This block just filters the `usub` component.
	2. `usub` is then added as a dependency of `block0`. Now `block0` implicitly depends on the value of `u` after filtering out everything but `usub`.
- When `usub` is added as an output to block `block1` on the tape, what actually happens is:
	1. `usub` is added as an output to block `block1`.
	2.  a `FunctionMergeBlock` is added to the tape with `usub` as a dependency and `u` as an output. This block combines the data in `usub` with all data in `u` except that of component `i`. `block1` now implicitly has `u` as an output, but only for the data in component `i`.

This test just creates a function in the R space, and tapes 1) assigning to it from a control 2) multiplying by 2 and 3) calculating the square. The tapes with/without using subfunctions are shown below, where you can see the extra blocks in the subfunction case. (The odd numbering is just so the code matches the tape).
```python
with stop_annotating():
    mesh = UnitIntervalMesh(1)

    w2 = Function(R).assign(2.0)
    w4 = Function(R)
    w7 = u.subfunctions[0]

use_subfunctions = True

continue_annotation()
with set_working_tape() as tape:
    w4.assign(w2)
    if use_subfunctions:
        w7 *= 2
    else:
        w4 *= 2
    J = assemble(inner(w4, w4) * dx)
    rf = ReducedFunctional(J, Control(w2), tape=tape)
pause_annotation()

assert taylor_test(rf, w2, Constant(0.1)) > 1.9
```
Without using subfunctions:
tape_nosub

Using subfunctions:
tape_sub

### Problem
This test fails if using subfunctions!
We want the `adj_value` of the `i`-th component to depend _only_ on what happens along the subfunction branch, and the `adj_value` of all other components to be whatever they were before the subfunctions split/merge.

Instead, the adjoint of `FunctionMergeBlock` was (correctly) outputting the `i`-th component of the adj_value along the subfunction branch of the tape, but (incorrectly) outputting _all_ components of the adj_value along the full-function branch of the tape.
When these two branches recombined the adj_value for the `i`-th component then had contributions from both branches.

### Solution
The merge `out[i] = usub; out[not i] = u` is a mask. The adjoint of a mask is also a mask, so `FunctionMergeBlock` now applies a mask before returning the adjoint outputs: `output0 = adj_input[i]; output1 = adj_input[not i]`

### DIFF
--- a/firedrake/adjoint_utils/blocks/function.py
+++ b/firedrake/adjoint_utils/blocks/function.py
@@ -238,6 +238,26 @@ class FunctionMergeBlock(Block):
 
     def evaluate_adj_component(self, inputs, adj_inputs, block_variable, idx,
                                prepared=None):
+        # The merge block appears whenever a subfunction is the output of a block.
+        # This means that the subfunction has been modified, so we need to make
+        # sure that this modification is accounted for when evaluating the adjoint.
+        #
+        # When recomputing the merge block, the indexed subfunction in the full
+        # Function is completely overwritten, meaning that the pre-existing value
+        # of the subfunction in the full function is ignored.
+        # The equivalent adjoint operation is to:
+        #   1. send the subfunction component of the adjoint value back up
+        #      the branch of the tape corresponding to the subfunction
+        #      dependency (idx=0).
+        #   2. zero out the subfunction component of the adjoint value sent
+        #      back up the full Function branch of the tape (idx=1).
+        # This means that when the adjoint values of each branch are combined
+        # after the SubfunctionBlock only the adjoint value from the subfunction
+        # branch is used.
+        #
+        # See https://github.com/firedrakeproject/firedrake/pull/4177 for more
+        # detail and for diagrams of the tape produced when accessing subfunctions.
+
         if idx == 0:
             return adj_inputs[0].subfunctions[self.sub_idx].copy(deepcopy=True)
         else:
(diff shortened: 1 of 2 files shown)


firedrake/adjoint_utils/blocks/function.py
```diff
@@ -239,8 +239,9 @@ def __init__(self, func, idx, ad_block_tag=None):
     def evaluate_adj_component(self, inputs, adj_inputs, block_variable, idx,
                                prepared=None):
         if idx == 0:
-            return adj_inputs[0].subfunctions[self.idx]
```
dham, line 242: please to explain the maths here.

dham (dismissed):

rckirby (approved):

PR #4178, opened 2025-03-31
single source of truth for solver parameters for adjoint solver

JHopeCollins:
# Description

### DIFF
--- a/firedrake/adjoint_utils/blocks/solving.py
+++ b/firedrake/adjoint_utils/blocks/solving.py
@@ -562,9 +562,9 @@ def solve_init_params(self, args, kwargs, varform):
                 )
             self.adj_kwargs.pop("appctx", None)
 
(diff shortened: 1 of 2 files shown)


sghelichkhani:
@JHopeCollins Thanks for this! I think this has fixed the issue now.

JHopeCollins:
> @JHopeCollins Thanks for this! I think this has fixed the issue now.

Great. Does it pass all the g-adopt tests?

sghelichkhani:
It does with the ones we have in the CI. We are trying to test it on a large problem that should fail if the solve params are not set correctly. I will report back as soon as I can try this on our HPC system.

sghelichkhani:
Thanks @JHopeCollins . It seems to be working again!

JHopeCollins:
Great, thanks for confirming.

PR #3997, opened 2025-01-28
PETSc updates

dham:
Goes with https://github.com/firedrakeproject/petsc/pull/25 and https://github.com/firedrakeproject/slepc/pull/10

PR #4181, opened 2025-04-01
Updates for latest release of PETSc

connorjward:
Note that this will not work yet because petsc4py and slepc4py are not yet on PyPI (yet another argument for https://github.com/firedrakeproject/firedrake/pull/4175).

Supersedes https://github.com/firedrakeproject/firedrake/pull/3997

Closes https://github.com/firedrakeproject/firedrake/issues/4155

### DIFF
--- a/firedrake/cython/dmcommon.pyx
+++ b/firedrake/cython/dmcommon.pyx
@@ -2120,6 +2120,8 @@ def mark_entity_classes_using_cell_dm(PETSc.DM swarm):
         PetscInt nswarmCells, swarmCell, blocksize
         PetscInt *swarmParentCells = NULL
         PetscDataType ctype = PETSC_DATATYPE_UNKNOWN
+        const char *cellid = NULL
+        PETSc.PetscDMSwarmCellDM celldm
 
     plex = swarm.getCellDM()
     get_height_stratum(plex.dm, 0, &cStart, &cEnd)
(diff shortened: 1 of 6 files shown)


francesco-ballarin:
What's the long term plan in terms of CI?

My guess is that you may want to have three CIs:
- on `firedrake` `release` branch, check periodically that tests/demos still run against the blessed `PETSc`/`SLEPc` version. This should never fail.
- on `firedrake` `release` branch, check periodically that tests/demos run against the current `PETSc`/`SLEPc` `release` branch. This would not be expected to fail often, since point releases of `PETSc`/`SLEPc` are expected to be backward compatible.
- on `firedrake` `main` branch, check periodically that tests/demos run against the current `PETSc`/`SLEPc` `main` branch. Seeing this fail should not be a big deal, but it should be a warning that something will break in March/September, when the blessed `PETSc`/`SLEPc` version will be moved.

Feel free to move this elsewhere if more appropriate. Note that the last two CIs will need `petsc4py`/`slepc4py` to be unpinned (i.e. no `==3.22.2` in `pyproject.toml`).

connorjward:
> What's the long term plan in terms of CI?

> * on `firedrake` `release` branch, check periodically that tests/demos run against the current `PETSc`/`SLEPc` `release` branch. This would not be expected to fail often, since point releases of `PETSc`/`SLEPc` are expected to be backward compatible.

I don't know if this is a goal here. I think we have two Firedrake configurations which we want to test: `release` vs the 'blessed' PETSc version, and `main` vs something close to PETSc `main`.

> * on `firedrake` `main` branch, check periodically that tests/demos run against the current `PETSc`/`SLEPc` `main` branch. Seeing this fail should not be a big deal, but it should be a warning that something will break in March/September, when the blessed `PETSc`/`SLEPc` version will be moved.

I think the idea is to track and advance (weekly-ish) a known "last good commit" on PETSc `main`. That way we can fix things when they arise but not have our CI beholden to whether or not `main` works for us. Maybe it would be better to just track `main`, not sure.

francesco-ballarin:
> I don't know if this is a goal here. I think we have two Firedrake configurations which we want to test: release vs the 'blessed' PETSc version, and main vs something close to PETSc main.

My rationale for saying this was: say that you are forced to release a firedrake 2025.4.1 in July. Maybe at that point, since you were forced to do a new release anyways, you would want to move to PETSc 3.23.2 (or whatever point version they have arrived at), rather than stick to the 3.23.0 as of March.

firedrake/preconditioners/patch.py
```diff
@@ -759,6 +759,7 @@ def __call__(self, pc):
 class PatchBase(PCSNESBase):
 
     def initialize(self, obj):
+        breakpoint()
```
connorjward, line 762:
```suggestion
```

rckirby (approved):
Looks fine, let's land after #4121.

ksagiyam (approved):

PR #4182, opened 2025-04-01
Download CMake on homebrew because PETSc release does not support 4.0

connorjward:
Hopefully fixed by https://gitlab.com/petsc/petsc/-/merge_requests/8261 (and then put into next release).


JHopeCollins (approved):
LGTM

PR #4188, opened 2025-04-02
Don't have streamplot error on a zero field

dham:


firedrake/pyplot/mpl.py
```diff
@@ -622,7 +627,10 @@ def streamplot(function, resolution=None, min_length=None, max_time=None,
     if max_time is None:
         area = assemble(Constant(1) * dx(mesh))
         average_speed = np.sqrt(assemble(inner(function, function) * dx) / area)
-        max_time = 50 * min_length / average_speed
+        if average_speed:
```
dham, line 630:
```suggestion
        if np.isclose(average_speed, 0.0):
```

PR #4190, opened 2025-04-02
Use a loopy release

connorjward:
I would prefer to avoid pinning loopy if at all possible. I think making patch releases when things break is the way to go.

Closes https://github.com/firedrakeproject/firedrake/issues/4159


pyproject.toml
```diff
@@ -20,6 +20,7 @@ dependencies = [
   "mpi4py; python_version < '3.13'",
   "h5py>3.12.1",
   "libsupermesh",
+  "loopy>2024.1",
```
dham, line 23: Should this not be 2025.1?

francesco-ballarin: It can either be `>2024.1` or `>=2025.1`, as they are equivalent.

Personally, I'd think that `>=2025.1` will be easier to maintain, because it will arguably be easier in future to answer the question "what was the version that added that feature I really needed" rather than "what was the latest version that didn't have that feature"...

connorjward: I guess. I was thinking in terms of "what versions are illegal". They are obviously equivalent so happy to change if there's another preference.

dham: ah. I don't mind.

dham (approved):

PR #4191, opened 2025-04-03
Tidy up GitHub workflows

connorjward:
* Move docs build into main workflow (no reason to have separate and means fewer notifications)
* Rename build.yml -> test.yml (seems better)
* Rename build-mac.yml -> test_macos.yml (seems more consistent)


pbrubeck (approved):
LGTM

PR #4194, opened 2025-04-03
Use petsctools package

connorjward:
A lot of users have been cargo-culting Firedrake's PETSc code for a long time (especially the options manager). This is even happening outside of the Firedrake user base. I (and others) think that it therefore makes some sense to have an external mini package that can store this sort of thing.

I have started this process off by writing [petsctools](https://github.com/connorjward/petsctools). At the moment it does just the thing I want which is to warn users if their PETSc and petsc4py versions do not match. Also if `PETSC_DIR` is set but does not match what petsc4py thinks then an exception is raised.

If people are happy I will move petsctools into the firedrakeproject organisation and make a release.


dham:
Sounds like a good idea. I guess we also put the options manager in there, right?

connorjward:
> Sounds like a good idea. I guess we also put the options manager in there, right?

Yeah exactly. Also whatever clever ideas happen with the appctx.

dham (approved):

firedrake/petsc.py
```diff
@@ -28,273 +26,34 @@ class FiredrakePETScError(Exception):
...
-    contains the parameters from this object."""
-        try:
-            for k, v in self.parameters.items():
-                self.options_object[self.options_prefix + k] = v
-            yield
-        finally:
-            for k in self.to_delete:
-                del self.options_object[self.options_prefix + k]
+@utils.deprecated("petsctools.OptionsManager")
+def OptionsManager(*args, **kwargs):
```
JHopeCollins, line 55: Does this DTRT with inheritance?

connorjward: Nope. Changed to use `warnings.deprecated` as suggested in the meeting. That should DTRT.

dham (approved):

PR #4216, opened 2025-04-09
Update AIMS schedule

JDBetteridge:
# Description


docs/source/adjoint.rst
```diff
@@ -289,11 +289,11 @@ are to be found on `the pygraphviz website
 
 .. _fig-tape:
 
-.. figure:: images/tape.pdf
```
connorjward, line 292: Does this mean `tape.pdf` should be removed? Does it exist?

JDBetteridge: :shrug: 
Ideally these things should be being generated dynamically, but we need this for the tutorial this week and the .pdf only renders in Safari, but not Chrome and Firefox.

JDBetteridge: I messed up, that should be an SVG anyway

connorjward: Can you also remove the pdf? Just checked and it isn't used anywhere else so it shouldn't be left lying around.

JDBetteridge:
Okay, I put it back. You fixed it upstream whilst I was lecturing

PR #4220, opened 2025-04-10
Fix bad link

connorjward:
# Description


JDBetteridge:
Fixed in https://github.com/firedrakeproject/firedrake/pull/4216

PR #4221, opened 2025-04-11
Remove unused PDF file

connorjward:
# Description


JDBetteridge (approved):
I'll reiterate that this should ideally be auto-generated, the s g from the pdf that is (and ideally checked somehow), but removing the extraneous file is DTRT™

PR #4222, opened 2025-04-11
Test Firedrake release

connorjward:
This is for testing workflows against the `release` branch (which I will reset to `master` when https://github.com/firedrakeproject/firedrake/pull/4193 goes in).


.github/workflows/build.yml
```diff
@@ -63,9 +60,28 @@ jobs:
...
+          else
+            exit 1
+          fi
+
+      # TODO: is there a 'blessed commit' or do we just point to 'main'?
+      # I prefer the latter because then failures are everyone's problem, not just mine!
       - name: Install PETSc
         run: |
-          git clone --depth 1 --branch $(python3 ./firedrake-repo/scripts/firedrake-configure --show-petsc-version) https://gitlab.com/petsc/petsc.git
+          if [ "${{ github.ref_name == 'master' }}" ]; then
```
francesco-ballarin, line 78: Watch out that on branches/PR of firedrake this will install PETSc from release, rather than from main.

I guess you would want to reverse the logic: if on firedrake release install PETSc from release, otherwise (assuming that it's unlikely for anyone to create a PR with target release, rather than main) use PETSc from main.

connorjward: Ah I thought that `github.ref_name` was always the target branch for PRs and merges. That's wrong. I am thinking of a larger refactor of the code where this workflow would get dispatched and so this would be specified as an input rather than inferred from `github`.

I agree that we should special case `release` instead of `main` here (i.e. make `main` the default).

francesco-ballarin: I think it would the source branch, at least I have several `github.ref == 'refs/heads/main'` for checking against scripts that are only supposed to run when testing main. You may want to double check anyways, since I may be remembering it wrongly.

Issue #4069, opened 2025-02-26
Add Thetis demos and notebooks back into the jupyter docker container once Thetis is in the firedrake container


JHopeCollins:
https://github.com/firedrakeproject/firedrake/blob/d14d47289df968ff0498123ee0b3194a4aec9c0b/docker/Dockerfile.jupyter#L17-23

PR #4225, opened 2025-04-11
Add Thetis back to Docker images

connorjward:
Fixes https://github.com/firedrakeproject/firedrake/issues/4069

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -252,15 +252,12 @@ jobs:
 
   docker:
     name: Build Docker containers
-    # UNDO ME
-    # if: github.ref == 'refs/heads/master'
-    # needs: [test, docker_tag]
+    if: github.ref == 'refs/heads/master'
+    needs: [test, docker_tag]
     uses: ./.github/workflows/docker.yml
     with:
-      # tag: ${{ needs.docker_tag.outputs.tag }}
-      # status: ${{ needs.test.result }}
-      tag: thetistest
-      status: success
+      tag: ${{ needs.docker_tag.outputs.tag }}
+      status: ${{ needs.test.result }}
     secrets: inherit
 
   lint:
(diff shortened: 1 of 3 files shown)


danshapero:
For what it's worth, icepack is using pyproject.toml now, I got rid of the extra data files that we had to help fetch observational data, and we've switched to using Angus's (pip-installable) Trilinos fork. So if you want to knock out two things with the same PR you could try adding icepack to this as well -- I think it should go ok. But happy to save that for a different PR if you want to do just one thing at a time.

.github/workflows/test.yml
```diff
@@ -252,12 +252,15 @@ jobs:
 
   docker:
     name: Build Docker containers
-    if: github.ref == 'refs/heads/master'
-    needs: [test, docker_tag]
+    # UNDO ME
+    # if: github.ref == 'refs/heads/master'
+    # needs: [test, docker_tag]
```
connorjward, line 257:
```suggestion
    if: github.ref == 'refs/heads/master'
    needs: [test, docker_tag]
```

.github/workflows/test.yml
```diff
@@ -252,12 +252,15 @@ jobs:
...
     uses: ./.github/workflows/docker.yml
     with:
-      tag: ${{ needs.docker_tag.outputs.tag }}
-      status: ${{ needs.test.result }}
+      # tag: ${{ needs.docker_tag.outputs.tag }}
+      # status: ${{ needs.test.result }}
+      tag: thetistest
+      status: success
```
connorjward, line 263:
```suggestion
      tag: ${{ needs.docker_tag.outputs.tag }}
      status: ${{ needs.test.result }}
```

docker/Dockerfile.jupyter
```diff
@@ -1,6 +1,6 @@
 # Dockerfile for a Firedrake container including Jupyter.
 
-FROM firedrakeproject/firedrake:latest
+FROM firedrakeproject/firedrake:thetistest
```
connorjward, line 3:
```suggestion
FROM firedrakeproject/firedrake:latest
```

ksagiyam (approved):

PR #4228, opened 2025-04-14
Add temporary constraint on islpy

connorjward:
See https://github.com/inducer/islpy/issues/162

Once merged we should create an issue so we don't forget to undo this later on.


connorjward:
Thanks Koki!

PR #4234, opened 2025-04-15
Fix netgen import in container by setting PYTHONPATH

connorjward:
See https://github.com/NGSolve/netgen/issues/213

Merging this before release is probably a good idea.

### DIFF
--- a/docker/Dockerfile.firedrake
+++ b/docker/Dockerfile.firedrake
@@ -7,6 +7,10 @@ RUN pip install --verbose \
         --extra-index-url https://download.pytorch.org/whl/cpu \
         jax ngsPETSc torch vtk
 
+# Set PYTHONPATH so netgen can be found
+# (see https://github.com/NGSolve/netgen/issues/213)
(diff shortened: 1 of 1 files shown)


docker/Dockerfile.firedrake
```diff
@@ -7,6 +7,10 @@ RUN pip install --verbose \
         --extra-index-url https://download.pytorch.org/whl/cpu \
         jax ngsPETSc torch vtk
 
+# Set PYTHONPATH so netgen can be found
+# (see https://github.com/NGSolve/netgen/issues/213)
+ENV PYTHONPATH=/usr/local/lib/python3.12/site-packages
```
dham, line 12: Usually one adds to paths even if they are currently empty. E.g. someone might in future need to add to PYTHONPATH in firedrake-vanilla-default, which this would then break.

connorjward:
```suggestion
ENV PYTHONPATH=/usr/local/lib/python3.12/site-packages:$PYTHONPATH
```

ksagiyam (approved):

PR #4237, opened 2025-04-16
Random Cofunctions

pbrubeck:
# Description

### DIFF
--- a/firedrake/randomfunctiongen.py
+++ b/firedrake/randomfunctiongen.py
@@ -106,7 +106,7 @@ import numpy.random as randomgen
 
 from firedrake.function import Function
 from pyop2.mpi import COMM_WORLD
-from ufl import FunctionSpace
+from ufl.functionspace import BaseFunctionSpace
 
 _deprecated_attributes = ['RandomGenerator', ]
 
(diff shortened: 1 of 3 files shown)


firedrake/randomfunctiongen.py
```diff
@@ -106,7 +106,7 @@
 
 from firedrake.function import Function
 from pyop2.mpi import COMM_WORLD
-from ufl import FunctionSpace
+from ufl.functionspace import BaseFunctionSpace
```
pbrubeck, line 109: It's either `ufl.BaseFunctionSpace` or `firedrake.functionspaceimpl.WithGeometryBase`

ksagiyam: Please update the doc in the above. I realised that the current doc is slightly wrong. We should use `.WithGeometry` (and `.FiredrakeDualSpace`) instead of `.FunctionSpace`.

tests/firedrake/adjoint/test_split_and_subfunctions.py
```diff
@@ -209,7 +209,7 @@ def test_writing_to_subfunctions():
         u.assign(kappa)
         usub *= 2
         J = assemble(inner(u, u) * dx)
-        print(f"{type(J) = }")
+        print(f"{type(J)=}")
```
ksagiyam, line 212:
```suggestion
```

ksagiyam (changes requested):

firedrake/randomfunctiongen.py
pbrubeck: We don't want the base class?

pbrubeck: Or should we test for the UFL base class in the code, and refer to concrete firedrake classes in the docstring?

pbrubeck: Done.

ksagiyam (approved):

PR #4238, opened 2025-04-16
Fix FacetSplit.update

pbrubeck:
# Description
This PR fixes `FacetSplitPC` to correctly update nested preconditioners. In particular, nested python preconditioners were not correctly updated.

### DIFF
--- a/firedrake/preconditioners/fdm.py
+++ b/firedrake/preconditioners/fdm.py
@@ -200,7 +200,8 @@ class FDMPC(PCBase):
             ebig, = set(unrestrict_element(Vsub.ufl_element()) for Vsub in V)
             Vbig = V.reconstruct(mesh=V.mesh().unique(), element=ebig)
             space_dim = Vbig.finat_element.space_dimension()
-            assert space_dim == sum(Vsub.finat_element.space_dimension() for Vsub in V)
+            if len(V) > 1:
+                assert space_dim == sum(Vsub.finat_element.space_dimension() for Vsub in V)
             fdofs = restricted_dofs(Vfacet.finat_element, Vbig.finat_element)
             idofs = numpy.setdiff1d(numpy.arange(space_dim, dtype=fdofs.dtype), fdofs)
         else:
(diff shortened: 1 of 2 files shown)


firedrake/preconditioners/facet_split.py
```diff
@@ -143,11 +144,10 @@ def _restrict_nullspace(nsp):
 
     def update(self, pc):
         if hasattr(self, "_permute_op"):
-            for mat in self.pc.getOperators():
-                mat.destroy()
```
JHopeCollins, line 147: Will these now just get cleaned up automatically by garbage collection?

firedrake/preconditioners/facet_split.py
```diff
@@ -143,11 +144,10 @@ def _restrict_nullspace(nsp):
 
     def update(self, pc):
         if hasattr(self, "_permute_op"):
```
JHopeCollins, line 146: Please can we add a member for what configuration is being used instead of implicitly checking this based on what attributes we have? It would be clearer to the reader and more robust.

firedrake/preconditioners/facet_split.py
```diff
@@ -88,10 +88,11 @@ def initialize(self, pc):
             self.set_nullspaces(pc)
             self.work_vecs = self.mixed_opmat.createVecs()
         elif self.subset:
```
JHopeCollins, line 90: Configuration member requested below would probably be set before this if-else block and then used for this block, and in the update method below.

JHopeCollins (changes requested):
Looks generally good to me, just a couple of minor comments.

firedrake/preconditioners/facet_split.py
pbrubeck: The operators are now the same petsc objects, so we should not destroy them.

pbrubeck:
This is actually a bug fix, but it now depends on https://github.com/firedrakeproject/firedrake/pull/4693 which we merged to main.

Issue #3763, opened 2024-09-12
DOCS: Website documentation about solvers is wildly out of date


JDBetteridge:
**Describe the issue**
I can't remember when the default solver option for LVP became LU, but it was a long time ago!

**Where to find**
Where to find the error or where documentation should be:
(website/github/docstrings/etc.)
- [Default solver options](https://www.firedrakeproject.org/solving-interface.html#default-solver-options)

**Additional info**
Add any other information here.

tcalderon-space:
Hello, I am new to this community. I am interested in participating in the community by helping out with the issues. I would like to contribute with this issue here. 

Would the correct adjustment for the solver option be something like https://www.firedrakeproject.org/solving-interface.html under "Recursive Fieldsplit" section?

The update you are requesting is to update the pc_type parameter to "LU" on the page given. Is that correct? I think the documentation might have already been updated...

JDBetteridge:
Not quite, the bit that needs updating is in the default solver options section here:
https://www.firedrakeproject.org/solving-interface.html#default-solver-options
The introductory paragraph here, is also wrong:
https://www.firedrakeproject.org/solving-interface.html#linear-solver-options

The relevant file that needs changing is here:
https://github.com/firedrakeproject/firedrake/blob/master/docs/source/solving-interface.rst

You can find the specification for default solver options here:
https://github.com/firedrakeproject/firedrake/blob/master/firedrake/petsc.py#L449-L452
Here:
https://github.com/firedrakeproject/firedrake/blob/master/firedrake/variational_solver.py#L134-L143
And here:
https://github.com/firedrakeproject/firedrake/blob/master/firedrake/variational_solver.py#L417-L420

It would be very cool if we could have the documentation automatically generated so that if this ever changes again there is a single source of truth, but in the short term correcting this mistake would be sufficient.

In anticipation of your future PR, thanks! 🙂

connorjward:
Note that this issue is more or less a duplicate of https://github.com/firedrakeproject/firedrake/issues/4065 and so both can be closed at once.

It isn't too difficult to automatically generate the text to have a single source of truth. It is what we do [here](https://github.com/firedrakeproject/firedrake/blob/master/docs/Makefile#L92) for example (injected into the docs [here](https://github.com/firedrakeproject/firedrake/blob/master/docs/source/install.rst?plain=1#L88)). But as @JDBetteridge says that isn't essential.

tcalderon-space:
Hello, 

[...]

PR #4241, opened 2025-04-18
Fix documentation requested in issue #3763. 

tcalderon-space:
clean.rst is the solving-interface.rst

# Description

Fixes the documentation requested in issue #3763. Aimed to make it also automatic..

### DIFF
--- a/clean.rst
+++ b/clean.rst
@@ -1,0 +1,1124 @@
+.. only:: html
+
+   .. contents::
+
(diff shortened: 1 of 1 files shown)


dham:
Thank you for contributing to our documentation. Rather than adding a whole new file can you please modify the existing file so we can see the diff.

dham (changes requested):

PR #4244, opened 2025-04-22
Clement interpolant

joewallwork:
I've been meaning to upstream some functionality from [Animate](https://github.com/mesh-adaptation/animate) that might be useful for the wider Firedrake community, so am starting with this. Let me know if it's of interest!

This PR adds functionality for the Clément interpolant, which transfers fields from P1 space to P0 space via local volume averaging. We've found this particularly useful for gradient/Hessian recovery in P1 space (can upstream those later).

I've included tests on simplices in 1D, 2D, and 3D, and for scalar-, vector-, and matrix-valued functions.

A limitation of the approach is that the local averaging breaks down on the boundary, but it does a good job in the domain interior. I also have an implementation of local facet averaging on the boundary that I can provide, if it'd be of use.

### DIFF
--- a/tests/firedrake/regression/test_interpolate.py
+++ b/tests/firedrake/regression/test_interpolate.py
@@ -570,7 +570,7 @@ def test_interpolate_logical_not():
 
 @pytest.mark.parametrize("tdim,shape", [(1, tuple()), (2, tuple()), (3, tuple()),
                                         (1, (1,)), (2, (2,)), (2, (3,)), (3, (3,)),
-                                        (1, (1, 1)), (2, (2, 2)), (3, (3, 3))],
+                                        (1, (1, 2)), (2, (2, 3)), (3, (2, 3))],
                          ids=["1d-scalar", "2d-scalar", "3d-scalar", "1d-vector",
                               "2d-vector", "2d-3vector", "3d-vector", "1d-matrix",
                               "2d-matrix", "3d-matrix"])
@@ -580,22 +580,21 @@ def test_clement_interpolator_simplex(tdim, shape):
         2: UnitSquareMesh,
         3: UnitCubeMesh,
     }[tdim](*(5 for _ in range(tdim)))
+    x = SpatialCoordinate(mesh)
     if len(shape) == 0:
-        fs_constructor = FunctionSpace
-        expr = sum(SpatialCoordinate(mesh))
+        P0 = FunctionSpace(mesh, "DG", 0)
+        P1 = FunctionSpace(mesh, "CG", 1)
+        expr = sum(x)
     elif len(shape) == 1:
         dim = shape[0]
-        fs_constructor = lambda *args: VectorFunctionSpace(*args, dim=dim)
-        x = SpatialCoordinate(mesh)
-        if dim == tdim:
-            expr = as_vector(x)
-        else:
-            expr = as_vector([x[0] for _ in range(dim)])
+        P0 = VectorFunctionSpace(mesh, "DG", 0, dim=dim)
+        P1 = VectorFunctionSpace(mesh, "CG", 1, dim=dim)
+        expr = as_vector(x if dim == tdim else [x[0] for _ in range(dim)])
     else:
-        fs_constructor = lambda *args: TensorFunctionSpace(*args, shape=shape)
-        expr = as_matrix([list(SpatialCoordinate(mesh)) for _ in range(tdim)])
-    P0 = fs_constructor(mesh, "DG", 0)
-    P1 = fs_constructor(mesh, "CG", 1)
+        P0 = TensorFunctionSpace(mesh, "DG", 0, shape=shape)
+        P1 = TensorFunctionSpace(mesh, "CG", 1, shape=shape)
+        rows = [Constant(tuple(range(i+1, i+1+tdim))) for i in range(P1.block_size)]
+        expr = as_tensor(np.reshape([dot(row, x) for row in rows], shape))
 
     # Projecting into P0 space and then applying Clement interpolation should recover
     # the original function


firedrake/interpolation.py
```diff
@@ -1664,3 +1666,126 @@ def multHermitian(self, source_vec, target_vec):
...
+        element = Vs.ufl_element()
+        if not (element.family() == "Discontinuous Lagrange" and element.degree() == 0):
+            raise ValueError("Source function must live in P0 space.")
+        rank = len(Vs.value_shape)
+        if rank not in (0, 1, 2):
+            raise ValueError(f"Rank-{rank} tensors are not supported.")
+        if rank != len(self.V.value_shape):
+            raise ValueError(f"Rank-{rank} input inconsistent with target space.")
+        mesh = self.V.mesh()
+        dim = mesh.topological_dimension()
```
pbrubeck, line 1747: This `dim` should be the number of copies of the element `V.block_size`. I think a `VectorFunctionSpace` and `TensorFunctionSpace` could be indexed with a double index `(flat_dof_id, component_id)`. So `rank` could possibly be unrestricted and there wouldn't have to be special cases if you treat the rank-0 as the general case.

tests/firedrake/regression/test_interpolate.py
```diff
@@ -566,3 +566,37 @@ def test_interpolate_logical_not():
     a = assemble(interpolate(conditional(Not(x < .2), 1, 0), V))
     b = assemble(interpolate(conditional(x >= .2, 1, 0), V))
     assert np.allclose(a.dat.data, b.dat.data)
+
+
+@pytest.mark.parametrize("rank", range(3))
+@pytest.mark.parametrize("dim", range(1, 4))
+def test_clement_interpolator_simplex(dim, rank):
```
pbrubeck, line 573: We should test a vector with 3 component in a 2D mesh.

firedrake/interpolation.py
```diff
@@ -1664,3 +1666,126 @@ def multHermitian(self, source_vec, target_vec):
...
+        }[rank]
+        keys = {"f": (function, op2.READ), "vol": (volume, op2.READ), "out": (output, op2.RW)}
+        parloops.par_loop((domain, instructions), dX, keys)
+
+        # Compute patch volume
+        patch_volume = ffunc.Function(functionspace.FunctionSpace(mesh, "CG", 1))
+        domain = "{[i]: 0 <= i < patch.dofs}"
+        instructions = "patch[i] = patch[i] + vol[0]"
+        keys = {"vol": (volume, op2.READ), "patch": (patch_volume, op2.RW)}
+        parloops.par_loop((domain, instructions), dX, keys)
```
pbrubeck, line 1775: The patch volume here could become a cached_property so it is not recomputed every time. Maybe we could attach this property to the mesh, so it can be reused more widely across different `Interpolators`.

jwallwork23: Good idea. I've addressed this in 34ceb670548ad9f7b2385f3337a3ab50991ae3cf, using the `cell_sizes` cached property as a template.

tests/firedrake/regression/test_interpolate.py
jwallwork23: Good suggestion - refactoring for this helped to address some limitations in the implementation. Addressed in 7ec5197c3f525896e3ce51b9b9bc71dbfb146c00.

firedrake/interpolation.py
jwallwork23: Addressed in 7ec5197c3f525896e3ce51b9b9bc71dbfb146c00.

firedrake/interpolation.py
```diff
@@ -1664,3 +1666,126 @@ def multHermitian(self, source_vec, target_vec):
...
+    r"""
+    Compute the Clément interpolant of a :math:`\mathbb{P}0` source field, i.e., take
+    the volume average over neighbouring cells at each vertex.
+
+    See :cite:`Clement:1975` for details.
+
+    For arguments, see :class:`.Interpolator`.
+    """
+
+    def __new__(cls, expr, V, **kwargs):
```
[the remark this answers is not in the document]
jwallwork23: The reason I put it here is that the class inherits from `Interpolator`, which defines `__new__` to account for the different implementations in `SameMeshInterpolator` and `CrossMeshInterpolator`. I could overwrite `__new__` to just `pass` and then put this logic into `__init__` if that'd be preferred - just let me know.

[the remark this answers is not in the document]
connorjward: `__new__` should only be overloaded (and very rarely!) to dispatch to different class constructors. It isn't needed here so I wouldn't define a `__new__` at all.

firedrake/interpolation.py
```diff
@@ -1664,3 +1666,120 @@ def multHermitian(self, source_vec, target_vec):
...
+        """
+        from firedrake.assemble import assemble
+        if adjoint:
+            raise NotImplementedError("Adjoint of Clément interpolation not implemented.")
+        Vs = function.function_space()
+        element = Vs.ufl_element()
+        if not (element.family() == "Discontinuous Lagrange" and element.degree() == 0):
+            raise ValueError("Source function must live in P0 space.")
+        rank = len(Vs.value_shape)
+        if rank not in (0, 1, 2):
```
pbrubeck, line 1742: This could go away

firedrake/interpolation.py
```diff
@@ -1664,3 +1666,120 @@ def multHermitian(self, source_vec, target_vec):
...
+        adjoint: bool
+                   Set to true to apply the adjoint of the interpolation
+                   operator.
+
+        Returns
+        -------
+        firedrake.function.Function or firedrake.cofunction.Cofunction
+            The resulting interpolated function.
+        """
+        from firedrake.assemble import assemble
```
pbrubeck, line 1734:
```suggestion
```

firedrake/interpolation.py
```diff
@@ -1664,3 +1666,120 @@ def multHermitian(self, source_vec, target_vec):
...
+        mesh = self.V.mesh()
+        if output is None:
+            output = ffunc.Function(self.V)
+
+        # Determine the number of degrees of freedom for vector and tensor elements
+        s0, s1 = None, None
+        if rank == 1:
+            s0 = Vs.block_size
+        elif rank == 2:
+            s0, s1 = Vs.value_shape
```
pbrubeck, line 1755:
```suggestion
```

firedrake/interpolation.py
```diff
@@ -1664,3 +1666,120 @@ def multHermitian(self, source_vec, target_vec):
...
+        domain = {
+            0: "{[i]: 0 <= i < out.dofs}",
+            1: f"{{[i, j]: 0 <= i < out.dofs and 0 <= j < {s0}}}",
+            2: f"{{[i, j, k]: 0 <= i < out.dofs and 0 <= j < {s0} and 0 <= k < {s1}}}",
+        }[rank]
+        instructions = {
+            0: "out[i] = out[i] + vol[0] * f[0]",
+            1: "out[i, j] = out[i, j] + vol[0] * f[0, j]",
+            2: f"out[i, {s0} * j + k] = out[i, {s0} * j + k] + vol[0] * f[0, {s0} * j + k]",
+        }[rank]
```
pbrubeck, line 1767:
```suggestion
        domain = f"{{[i, j]: 0 <= i < out.dofs and 0 <= j < {Vs.block_size}}}"
        instructions = "out[i, j] = out[i, j] + vol[0] * f[0, j]"
```

firedrake/interpolation.py
```diff
@@ -1664,3 +1666,120 @@ def multHermitian(self, source_vec, target_vec):
...
+        domain = {
+            0: "",
+            1: f"{{[j]: 0 <= j < {s0}}}",
+            2: f"{{[j, k]: 0 <= j < {s0} and 0 <= k < {s1}}}",
+        }[rank]
+        instructions = {
+            0: "out[0] = out[0] / patch[0]",
+            1: "out[0, j] = out[0, j] / patch[0]",
+            2: f"out[0, {s0} * j + k] = out[0, {s0} * j + k] / patch[0]",
+        }[rank]
```
pbrubeck, line 1781:
```suggestion
        domain = f"{{[j]: 0 <= j < {Vs.block_size}}}"
        instructions = "out[0, j] = out[0, j] / patch[0]"
```

firedrake/interpolation.py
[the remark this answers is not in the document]
jwallwork23: Well I figured it'd make sense for `ClementInterpolator` to inherit from `Interpolator`, so something needs to be done to account for its implementation there, right? Or did I miss something?

pbrubeck (changes requested):

firedrake/interpolation.py
```diff
@@ -1664,3 +1666,120 @@ def multHermitian(self, source_vec, target_vec):
...
+        from firedrake.assemble import assemble
+        if adjoint:
+            raise NotImplementedError("Adjoint of Clément interpolation not implemented.")
+        Vs = function.function_space()
+        element = Vs.ufl_element()
+        if not (element.family() == "Discontinuous Lagrange" and element.degree() == 0):
+            raise ValueError("Source function must live in P0 space.")
+        rank = len(Vs.value_shape)
+        if rank not in (0, 1, 2):
+            raise ValueError(f"Rank-{rank} tensors are not supported.")
```
pbrubeck, line 1743:
```suggestion
```

firedrake/mesh.py
```diff
@@ -2415,6 +2415,35 @@ def clear_cell_sizes(self):
...
+    def cell_volume(self):
+        """
+        A :class:`~.Function` in the :math:`P^0` space containing the local mesh volume.
+
+        This is computed by interpolating the UFL :class:`~.CellVolume` for the current
+        mesh.
+        """
+        from firedrake.function import Function
+        from firedrake.functionspace import FunctionSpace
+        volume = Function(FunctionSpace(self, "Discontinuous Lagrange", 0))
```
pbrubeck, line 2428:
```suggestion
        DG0 = FunctionSpace(self, "Discontinuous Lagrange", 0)
        volume = Function(DG0)
```

firedrake/mesh.py
```diff
@@ -2415,6 +2415,35 @@ def clear_cell_sizes(self):
...
+    @utils.cached_property
+    def patch_volume(self):
+        """
+        A :class:`~.Function` in the :math:`P^1` space containing the sum of the volumes
+        of cells neighbouring a vertex.
+        """
+        from firedrake.function import Function
+        from firedrake.functionspace import FunctionSpace
+        from firedrake.parloops import par_loop, READ, RW
+        patch_vol = Function(FunctionSpace(self, "Lagrange", 1))
```
pbrubeck, line 2440:
```suggestion
        CG1 = FunctionSpace(self, "Lagrange", 1)
        patch_vol = Function(CG1)
```

firedrake/interpolation.py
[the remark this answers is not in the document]
pbrubeck: Could you not just leave `__new__` unimplemented, and move the check to `__init__` ?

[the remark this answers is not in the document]
jwallwork23: If I do that I get
```
        # Projecting into P0 space and then applying Clement interpolation should recover
        # the original function
        x_P0 = assemble(project(expr, P0))
        interpolator = ClementInterpolator(TestFunction(P0), P1)
>       x_P1 = interpolator.interpolate(x_P0)

tests/firedrake/regression/test_interpolate.py:604: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
petsc4py/PETSc/Log.pyx:188: in petsc4py.PETSc.Log.EventDecorator.decorator.wrapped_func
    ???
petsc4py/PETSc/Log.pyx:189: in petsc4py.PETSc.Log.EventDecorator.decorator.wrapped_func
    ???
firedrake/interpolation.py:418: in interpolate
    interp = self._interpolate_future(*function, adjoint=adjoint,
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

self = <firedrake.interpolation.SameMeshInterpolator object at 0x72d277a4b7d0>
transpose = None, adjoint = False, default_missing_val = None
function = (Coefficient(WithGeometry(FunctionSpace(<firedrake.mesh.MeshTopology object at 0x72d277bb3440>, TensorElement(FiniteEl...ape=(3, 3), symmetry={}), name=None), Mesh(VectorElement(FiniteElement('Lagrange', tetrahedron, 1), dim=3), 45)), 69),)

    def _interpolate_future(self, *function, transpose=None, adjoint=False, default_missing_val=None):
        """Define the :class:`Interpolate` object corresponding to the interpolation operation of interest.
    
        Parameters
        ----------
        *function: firedrake.function.Function or firedrake.cofunction.Cofunction
                   If the expression being interpolated contains an argument,
                   then the function value to interpolate.
        transpose : bool
                   Deprecated, use adjoint instead.
        adjoint: bool
                   Set to true to apply the adjoint of the interpolation
                   operator.
        default_missing_val: bool
                             For interpolation across meshes: the
                             optional value to assign to DoFs in the target mesh that are
                             outside the source mesh. If this is not set then the values are
                             either (a) unchanged if some ``output`` is specified to the
                             :meth:`interpolate` method or (b) set to zero. This does not affect
                             adjoint interpolation. Ignored if interpolating within the same
                             mesh or onto a :func:`.VertexOnlyMesh`.
    
        Returns
        -------
        firedrake.interpolation.Interpolate or ufl.action.Action or ufl.adjoint.Adjoint
            The symbolic object representing the interpolation operation.
    
        Notes
        -----
        This method is the default future behaviour of interpolation. In a future release, the
        ``Interpolator.interpolate`` method will be replaced by this method.
        """
    
>       V = self.V
E       AttributeError: 'SameMeshInterpolator' object has no attribute 'V'

firedrake/interpolation.py:336: AttributeError
```

firedrake/interpolation.py
jwallwork23: Done in https://github.com/firedrakeproject/firedrake/pull/4244/commits/6e67568a7eb9c2aaf39864c1d586d88caffe42e6.

firedrake/interpolation.py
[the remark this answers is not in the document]
connorjward: Oh god. I bet that `type(ClementInterpolator(TestFunction(P0), P1))` returns `SameMeshInterpolator` because of the `__new__` [magic](https://github.com/firedrakeproject/firedrake/blob/master/firedrake/interpolation.py#L261) in `Interpolator`.

firedrake/interpolation.py
```diff
@@ -1664,3 +1666,94 @@ def multHermitian(self, source_vec, target_vec):
...
+        allow_missing_dofs=False,
+    ):
+        if subset:
+            raise NotImplementedError("subset not implemented")
+        if freeze_expr:
+            raise NotImplementedError("freeze_expr not implemented")
+        if access != op2.WRITE:
+            raise NotImplementedError("access other than op2.WRITE not implemented")
+        if bcs:
+            raise NotImplementedError("bcs not implemented")
```
connorjward, line 1707: This sort of thing makes me wonder whether this is the best place to implement this functionality. Might it be better to simply have a free function called `interpolate_clement` that does the parloop?

tests/firedrake/regression/test_interpolate.py
```diff
@@ -566,3 +566,45 @@ def test_interpolate_logical_not():
     a = assemble(interpolate(conditional(Not(x < .2), 1, 0), V))
     b = assemble(interpolate(conditional(x >= .2, 1, 0), V))
     assert np.allclose(a.dat.data, b.dat.data)
+
+
+@pytest.mark.parametrize("tdim,shape", [(1, tuple()), (2, tuple()), (3, tuple()),
+                                        (1, (1,)), (2, (2,)), (2, (3,)), (3, (3,)),
+                                        (1, (1, 1)), (2, (2, 2)), (3, (3, 3))],
```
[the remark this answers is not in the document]
pbrubeck: We would have to create expr with the right shape. Maybe something like:

x = SpatialCoordinate(mesh)
expr = np.reshape([dot(Constant(tuple(range(i, i+tdim))), x) for i in range(V.block_size)], shape)

firedrake/interpolation.py
jwallwork23: As in instead of having the `Interpolator` at all?

tests/firedrake/regression/test_interpolate.py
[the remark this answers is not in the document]
jwallwork23: Addressed in cd20a5e43 and simplified the notation while I was at it.

jwallwork23:
The most recent test fails don't seem to be related to the changes in this PR.

firedrake/interpolation.py
connorjward: Yes. I think we need @dham to weigh in here. Is this the right abstraction?

connorjward: It might be nice to have `assemble(interpolate(source, target, variant="clement"))` but that might be tricky/unwieldy.

dham: I think we'd have to think harder about whether that is actually the Right Thing. For a finite element space with a known dual basis, their is a single canonical interpolation operation. Other interpolation operations are meaningfully different: they are just other projections that don't directly arise from the definition of the space.

dham: The variant option at least requires heavier surgery because `interpolate` is defined in UFL.

firedrake/mesh.py
```diff
@@ -2415,6 +2415,37 @@ def clear_cell_sizes(self):
...
+        A :class:`~.Function` in the :math:`P^0` space containing the local mesh volume.
+
+        This is computed by interpolating the UFL :class:`~.CellVolume` for the current
+        mesh.
+        """
+        from firedrake.function import Function
+        from firedrake.functionspace import FunctionSpace
+        DG0 = FunctionSpace(self, "Discontinuous Lagrange", 0)
+        volume = Function(DG0)
+        return volume.interpolate(ufl.CellVolume(self))
```
dham, line 2430: Is this actually mathematically correct for higher order meshes? Shouldn't it be a projection?

pbrubeck: As long as `CellVolume` is correctly integrated then this should be fine. Recall that `CellVolume` is intrinsically a DG0 function on any mesh.

PR #4247, opened 2025-04-22
Use more homebrew packages on Mac

dham:
Use several more easily available Homebrew packages on MacOS.


connorjward (approved):
Awesome. Thanks!

PR #4261, opened 2025-04-28
Use a UFL release

connorjward:
Must land with https://github.com/firedrakeproject/fiat/pull/146.


francesco-ballarin:
FYI, in principle you could leave out the dependency on ufl from firedrake, since it will be automatically pulled in from fiat.

In practice, I'd agree it definitely makes sense to have listed explicitly in both firedrake and fiat...

PR #4265, opened 2025-04-29
Post-release updates for master

connorjward:
Some small fixes, more will likely come.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,7 +1,7 @@
 [project]
 name = "firedrake"
 # <year>.<month>.<patch>
-version = "2025.4.0"
+version = "2025.4.0.dev0"
 description = "An automated system for the portable solution of partial differential equations using the finite element method"
 readme = "README.rst"
 license = {file = "LICENSE"}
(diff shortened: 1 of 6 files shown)


pyproject.toml
```diff
@@ -1,7 +1,7 @@
 [project]
 name = "firedrake"
 # <year>.<month>.<patch>
-version = "2025.4.0"
+version = "2025.4.0.dev0"
```
francesco-ballarin, line 4: Bump to `2025.4.1.dev0`?

I can't double check now, but I think that `pip` would consider `2025.4.0.dev0` to be less than `2025.4.0`.

connorjward: Argh you're right:
```
>>> from packaging.version import Version
>>> Version("2025.4.0.dev0") < Version("2025.4.0")
True
```

I thought I'd do this because `master` is really going for the next minor release (i.e. `2025.X.0.dev0`) but we don't know what `X` actually is since it is the release month.

connorjward: Changed to `2025.5.0.dev0` since that's the next viable minor release.

connorjward:
Forcing this through as this is holding up the PyPI package.

PR #4267, opened 2025-04-29
Make linkcheck not prevent later steps in workflows

connorjward:
This functionality was lost in the refactor, and is preventing publishing on PyPI.


connorjward:
Forcing through as this is holding up PyPI publishing.

PR #4272, opened 2025-04-30
Add option to build Docker containers manually

connorjward:
This is useful because if the release workflow doesn't work for some reason we still want to be able to build the containers. I find adding `workflow_dispatch` to be generally helpful.

Also remove refs to now removed docdeps container.

### DIFF
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -16,11 +16,6 @@ on:
       tag:
         description: Docker image tag
         type: string
-    secrets:
-      DOCKERHUB_USER:
-        required: true
-      DOCKERHUB_TOKEN:
-        required: true
 
 jobs:
   # Firedrake only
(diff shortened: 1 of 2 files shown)


JHopeCollins:
Any idea why that test is failing? 5e-12 seems quite specific for something that really should be evaluating close to machine precision. Is this test known to be delicate?

connorjward:
> Any idea why that test is failing? 5e-12 seems quite specific for something that really should be evaluating close to machine precision. Is this test known to be delicate?

No idea. There's no way it's related to these changes though. @ksagiyam just FYI I have seen this test failing occasionally and randomly in CI.

JHopeCollins (approved):
LGTM

ksagiyam:
Maybe we want to increment `.post` version again?

connorjward:
> Maybe we want to increment `.post` version again?

I don't think that we need to do that. That is only necessary when we actually publish a release.

PR #4275, opened 2025-05-01
Website hotfix

connorjward:
I made an error with the sdist release so we cannot quite use the PyPI package yet. I will fix that today and prepare a patch release. For now use the git URL to install instead.


connorjward:
Forcing through as this is critical and the docs build fine locally for me.

PR #4276, opened 2025-05-01
Fix sdist

connorjward:
The current sdist on PyPI is not installable because I didn't fully understand setuptools. I believe that this will fix things - and it's tested!

### DIFF
--- a/.github/workflows/core.yml
+++ b/.github/workflows/core.yml
@@ -186,6 +186,7 @@ jobs:
           fi
 
           : # Install from an sdist so we can make sure that it is not ill-formed
+          pip install build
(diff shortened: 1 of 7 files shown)


francesco-ballarin:
Once this is working make sure to port it to `main` (or `master`) too, otherwise you may end up with the same issue in 6 months time...

connorjward:
> Once this is working make sure to port it to `main` (or `master`) too, otherwise you may end up with the same issue in 6 months time...

The plan is to periodically (roughly every week but as needed) merge the changes in `release` back into `master` so this doesn't happen.

ksagiyam (approved):
Looks correct to me.

PR #4289, opened 2025-05-08
Use correct hook to prepare firedrake-check for editable installs

connorjward:
Fixes `firedrake-check` for editable installs.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -12,6 +12,7 @@ import libsupermesh
 import pkgconfig
 from dataclasses import dataclass, field
 from setuptools import setup, find_packages, Extension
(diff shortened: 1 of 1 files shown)


JHopeCollins (approved):
Works for me.

Maybe we want a CI job that only runs on release branch that installs an editable firedrake and just runs firedrake-check.

connorjward:
> Works for me.
> 
> Maybe we want a CI job that only runs on release branch that installs an editable firedrake and just runs firedrake-check.

Personally I'm not too fussed about testing this. Editable installs are primarily a developer feature.

PR #4298, opened 2025-05-10
pyproject.toml: spydump should be script-file

qbisi:
# Description
this pr fix installed script spydump


connorjward:
This is fine but could you try and merge this into `release` and not `master`? Otherwise the fix will not be released onto PyPI for 6 months.

PR #4312, opened 2025-05-13
cause actions to trigger when PR target is changed

dham:
Currently changing the target branch of a PR does not trigger CI. Per https://github.com/orgs/community/discussions/37863 this change fixes this.


connorjward (approved):

connorjward:
Forcing through as it clearly works fine and running the whole test suite isn't necessary.

PR #4315, opened 2025-05-13
Point data assimilation demo

leo-collins:
Demo of using firedrake-adjoint to assimilate point data. Based on Reuben's work

### DIFF
--- a/docs/source/_static/bibliography.bib
+++ b/docs/source/_static/bibliography.bib
@@ -367,6 +367,18 @@ Composing Abstractions},
   year          = {2016}
 }
 
+@article{Shapero:2021,
+  author        = {Shapero, D. R. and Badgeley, J. A. and Hoffman, A. O. and Joughin, I. R.},
+  doi           = {10.5194/gmd-14-4593-2021},
+  journal       = {Geoscientific Model Development},
+  number        = {7},
+  pages         = {4593--4616},
+  title         = {icepack: a new glacier flow modeling package in Python, version 1.0},
+  url           = {https://gmd.copernicus.org/articles/14/4593/2021/},
+  volume        = {14},
+  year          = {2021}
+}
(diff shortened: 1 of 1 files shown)


danshapero:
There's one aspect of sparse data assimilation that isn't documented and should be. By default, Firedrake will internally reorder some of the points in the point data space. If you're synthesizing the sparse data to assimilate then you can fill in the values based on their position. But if you're working from real geospatial data that you get from, say, a GeoTIFF file, this reordering now puts everything in a different order from what you might expect. There's a separate intermediate step where you need to untangle the orders. I can give you a more minimal example of this, but I had to put in [this fix](https://github.com/icepack/dual-problems/blob/967f8073918c72db219ea5d56d8ee510a34c9d8e/demos/kangerd/initialize.py#L103) for a paper that I had published recently. Now this notebook might not be the right place to describe this detail, but it should be written down more explicitly somewhere.

dham:
> here's one aspect of sparse data assimilation that isn't documented and should be. By default, Firedrake will internally reorder some of the points in the point data space. If you're synthesizing the sparse data to assimilate then you can fill in the values based on their position. But if you're working from real geospatial data that you get from, say, a GeoTIFF file, this reordering now puts everything in a different order from what you might expect. There's a separate intermediate step where you need to untangle the orders. I can give you a more minimal example of this, but I had to put in [this fix](https://github.com/icepack/dual-problems/blob/967f8073918c72db219ea5d56d8ee510a34c9d8e/demos/kangerd/initialize.py#L103) for a paper that I had published recently. Now this notebook might not be the right place to describe this detail, but it should be written down more explicitly somewhere.

The solution to this is to use the input ordering property on the VOM. See: https://www.firedrakeproject.org/point-evaluation.html#using-the-input-ordering-property

PR #4320, opened 2025-05-14
Cjc/vp demo

colinjcotter:
# Description

A 1D Vlasov Poisson demo


demos/vlasov_poisson_1d/vp1d.py.rst
```diff
@@ -0,0 +1,377 @@
...
+					 
+  params = {
+     'ksp_type': 'gmres',
+     'pc_type': 'lu',
+     'ksp_rtol': 1.0e-8,
+     }
+  nullspace = VectorSpaceBasis(constant=True)
+  phi_solver = LinearVariationalSolver(phi_problem,
+                                       nullspace=nullspace,
+				       solver_parameters=params)
```
JHopeCollins, line 244: I've rearranged this a bit and added some more detail to make it more explicit how the null space is dealt with a) in the maths vs in the code, and b) how to remove the null space and how to solve the problem.

I'm not super attached to this specific rewording, so feel free to edit, but the original version was a bit vague about how Wbar, Wdotbar, and the shifted operator related to each other.

```suggestion
To deal mathematically with the null space of the potential, we expressed the
problem in :math:`\mathring{\bar{W}}`. Programmatically we will express the
problem in :math:`\bar{W}` and deal with the null space by defining a basis
for the space of globally constant functions, which we will later pass to PETSc
so the solver can remove it from the solution. ::

  nullspace = VectorSpaceBasis(constant=True)

However, the null space also means that the assembled matrix of the
Poisson problem will be singular, which will prevent us from using a
direct solver. To deal with this, we will precondition the Poisson problem
with a version shifted by :math:`\int_{\Omega}\phi\psi\mathrf{d}x`. The
shifted problem is well-posed on :math:`\bar{W}`, so the assembled matrix
will be non-singular and so solvable with direct methods. ::
  
  shift_eqn = dphi.dx(0)*psi.dx(0)*dx + dphi*psi*dx

We use these to define a :class:`~.LinearVariationalProblem`. ::
  
  phi_problem = LinearVariationalProblem(lhs(phi_eqn), rhs(phi_eqn),
                                         phi, aP=shift_eqn)

Now we build the :class:`~.LinearVariationalSolver`. The problem
is preconditioned by the shifted operator which is solved using a direct
solver, and we pass the nullspace of globally constant functions to
the solver. ::
					 
  params = {
     'ksp_type': 'gmres',
     'pc_type': 'lu',
     'ksp_rtol': 1.0e-8,
     }
  phi_solver = LinearVariationalSolver(phi_problem,
                                       nullspace=nullspace,
				       solver_parameters=params)
```

demos/vlasov_poisson_1d/vp1d.py.rst
```diff
@@ -0,0 +1,377 @@
...
+
+Now we move onto the solver to compute :math:`\partial f/\partial t`. We
+define a symbolic :math:`\Delta t` which we will update later. ::
+  
+  dtc = Constant(0)
+
+The solver will take in ``fstar`` and return :math:`\Delta t\partial f/\partial t` in ``df_out``. ::
```
JHopeCollins, line 251:
```suggestion
At each stage, the solver will take in the intermediate solution ``fstar`` and
return the stage increment :math:`\Delta t\partial f/\partial t` in ``df_out``. ::
```

demos/vlasov_poisson_1d/vp1d.py.rst
```diff
@@ -0,0 +1,377 @@
...
+      f2.assign(3*fn/4 + (f1 + df_out)/4)
+
+The third stage. ::
+
+  #
+      fstar.assign(f2)
+      phi_solver.solve()
+      df_solver.solve()
+      fn.assign(fn/3 + 2*(f2 + df_out)/3)
+      t += dt
```
JHopeCollins, line 341: Can this be written as a loop with the coefficient array defined beforehand? I don't think we want to be encouraging people to handcode unrolled Runge-Kutta loops.

demos/vlasov_poisson_1d/vp1d.py.rst
```diff
@@ -0,0 +1,377 @@
...
+coordinate field. ::
+		      
+  Vc = mesh.coordinates.function_space()
+  x, v = SpatialCoordinate(mesh)
+  X = Function(Vc).interpolate(as_vector([x, v-H/2]))
+  mesh.coordinates.assign(X)
+
+Now we build a discontinuous finite element space for the density, ::
+  
+  V = FunctionSpace(mesh, 'DG', 1)
```
JHopeCollins, line 174: We shouldn't be raising warnings in our own demos.
```suggestion
  V = FunctionSpace(mesh, 'DQ', 1)
```

JHopeCollins (changes requested):
Looks good, most of the comments are just extra detail where the explanations are a bit terse, and a couple of small code changes to avoid warnings.

The only major suggestion is to reorder/reword the explanation of how the null space is dealt with.

JHopeCollins:
In the future, it could be nice to have a subsequent demo showing how to do this with a single solver, possibly with Irksome too.

The whole thing is linear, and if you have a mixed space with (phi, f) then the matrix is lower triangular so you can do exactly the method you have here with a multiplicative fieldsplit to solve for `phi` at the current stage, then calculate `f` using the latest potential velocity.
If you write this as a single system you can also pass it to Irksome to do the RK scheme.

demos/vlasov_poisson_1d/vp1d.py.rst
```diff
@@ -0,0 +1,377 @@
...
+with coordinates :math:`(x,v)\in \mathbb{R}^2`. From now on we will
+relabel these coordinates :math:`(x,v)\mapsto (x_1,x_2)`, obtaining
+the equivalent form,
+
+.. math::
+   f_t + \nabla\cdot(\vec{u}f) = 0, \quad \vec{u} = (v,-\phi_x/m), \quad
+   -\phi_{x_1x_1} = q_0\int f(x_1,x_2,t)\,\mathrm{d} x_2,
+
+where :math:`\nabla=(\partial_{x_1},\partial{x_2})`. From now we will
+choose units such that :math:`q_0,m` are absorbed into the definition of
```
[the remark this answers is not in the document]
colinjcotter: the latter.

colinjcotter:
Josh, the whole thing isn't linear, because of the a*f appearing in the conservation law.

JHopeCollins:
> Josh, the whole thing isn't linear, because of the a*f appearing in the conservation law.

Ah yes, of course. I got fooled because it is two linear solves, but the Jacobian of the second solve depends on the solution of the first.

demos/test_demos_run.py
connorjward, line 1: This file shouldn't have been added here.

demos/vlasov_poisson_1d/vp1d.py.rst
```diff
@@ -0,0 +1,390 @@
...
+problem in :math:`\bar{W}` and deal with the null space by defining a basis
+for the space of globally constant functions, which we will later pass to PETSc
+so the solver can remove it from the solution. ::
+
+  nullspace = VectorSpaceBasis(constant=True, comm=COMM_WORLD)
+
+However, the null space also means that the assembled matrix of the
+Poisson problem will be singular, which will prevent us from using a
+direct solver. To deal with this, we will precondition the Poisson problem
+with a version shifted by :math:`\int_{\Omega}\phi\psi\mathrf{d}x`. The
```
connorjward, line 232:
```suggestion
with a version shifted by :math:`\int_{\Omega}\phi\psi\mathrm{d}x`. The
```
I think this is your issue. The log says
```
! Undefined control sequence.
l.17730 ...ifted by \(\int_{\Omega}\phi\psi\mathrf
                                                  {d}x\). The
```

connorjward (changes requested):

demos/vlasov_poisson_1d/vp1d.py.rst
colinjcotter: It's a bit fiddly to do that because of the triviality of the first stage, so it will all look a bit cryptic if I do that, and not very pedagogical. This is also how we did it in the original DG advection demo. I think it is harmless.

colinjcotter: Not that I can't use IRKsome because it isn't an IRK.

demos/test_demos_run.py
```diff
@@ -0,0 +1,143 @@
+import glob
```
connorjward, line 1: This file shouldn't have been added here.

connorjward (changes requested):

connorjward:
This is now only failing the annoying SLATE wrong option check which is known to fail randomly. I am confident that there is nothing wrong with your code.

tests/firedrake/output/test_io_function.py
```diff
@@ -142,7 +142,7 @@ def _load_check_save_functions(filename, func_name, comm, method, mesh_name, var
     VB = fB.function_space()
     fBe = Function(VB)
     _initialise_function(fBe, _get_expr(VB), method)
-    assert assemble(inner(fB - fBe, fB - fBe) * dx) < 5.e-12
+    assert assemble(inner(fB - fBe, fB - fBe) * dx) < 6.e-12
```
ksagiyam, line 145: Could you make a separate PR just for this probably targeting `release`?

PR #4322, opened 2025-05-14
Knook/multicomponent

KarsKnook:
Demo for multicomponent flows -- microfluidic mixing of hydrocarbons


KarsKnook:
The references are not showing using `make html`?

demos/multicomponent/multicomponent.py.rst
```diff
@@ -0,0 +1,615 @@
+Multicomponent flow -- microfluidic non-ideal mixing of hydrocarbons
+===================================================
```
connorjward, line 2: `make html` is failing because this underline (and 2 others) and not long enough.

demos/multicomponent/multicomponent.py.rst
```diff
@@ -0,0 +1,615 @@
...
+    V_h = VectorFunctionSpace(mesh, "CG", k)        # Velocity space
+    U_h = FunctionSpace(mesh, "DG", k - 1)          # Species chemical potential space
+    P_h = FunctionSpace(mesh, "CG", k - 1)          # Pressure space
+    X_h = FunctionSpace(mesh, "DG", k - 1)          # Species mole fraction space
+    R_h = FunctionSpace(mesh, "CG", k - 1)          # Density reciprocal space
+    L_h = FunctionSpace(mesh, "R", 0)               # Lagrange multiplier space
+
+    # The mixed space
+    Z_h = J_h * J_h * V_h * U_h * U_h * P_h * X_h * X_h * R_h * L_h * L_h
+    PETSc.Sys.Print("Mesh has %d cells, with %d finite element DOFs" % (mesh.num_cells(), Z_h.dim()))
```
connorjward, line 97:
```suggestion
    PETSc.Sys.Print(f"Mesh has {mesh.null_cells()} cells, with {Z_h.dim()} finite element DOFs")
```

demos/multicomponent/multicomponent.py.rst
```diff
@@ -0,0 +1,615 @@
...
+
+    # The trial functions
+    sln = Function(Z_h)
```
connorjward, line 100: `solution` more clear?

demos/multicomponent/multicomponent.py.rst
```diff
@@ -0,0 +1,615 @@
...
+        :arg bc_point: the point at which to pin the function.
+            The location of the finite element DOF nearest to bc_point is actually used.
+        '''
+        def __init__(self, V, g, bc_point):
+            super(FixAtPointBC, self).__init__(V, g, bc_point)
+            if isinstance(bc_point, tuple):
+                bc_point = as_vector(bc_point)
+            self.bc_point = bc_point
+
+        @firedrake_utils.cached_property
```
connorjward, line 509: Use `functools.cached_property` instead

connorjward (changes requested):
In general looks really impressive. Nice job.

connorjward (approved):
I am happy but this is a massive demo so I will bring to today's meeting to check with the others before merging.

demos/multicomponent/multicomponent.py.rst
```diff
@@ -0,0 +1,615 @@
...
+
+    Omega = wp.Face()
+    
+    # Label the inlets and outlet; the other walls are unlabelled
+    Omega.edges.Max(ngocc.Y).name = "inlet_1"
+    Omega.edges.Min(ngocc.Y).name = "inlet_2"
+    Omega.edges.Max(ngocc.X).name = "outlet"
+
+    # Construct the mesh
+    ngmesh = ngocc.OCCGeometry(Omega, dim=2).GenerateMesh(maxh=0.055)
```
connorjward: This PR needs to be tested. It will probably take a long time to run so I'd recommend having a check and using a coarser mesh when run under CI. Something like:
```py
if os.getenv("FIREDRAKE_CI") == "1": 
        maxh = 0.1
    else:
        maxh = 0.055
```
We do something similar here: https://github.com/firedrakeproject/firedrake/blob/master/demos/full_waveform_inversion/full_waveform_inversion.py.rst?plain=1#L107

You also need to add it to `test_demos_run.py`.

KarsKnook: For `maxh=0.1` the problem has 64k DOFs and takes about 30s to run in serial on my MacBook. Is that still too long for CI?

For `maxh=0.2` we have 16k DOFs and 15s.

pbrubeck: 30 seconds for the actual computations sounds fine. The smaller the better. I'm more worried about the compilation time. Are the quadrature degrees here getting too large?

KarsKnook: The quadrature degree is set to 15. We are doing degree 3 FEM and we got nonlinear terms. What would you recommend?

pbrubeck: Use the lowest order discretization and set the quadrature degree to 3* k or 4*k

demos/multicomponent/multicomponent.py.rst
```diff
@@ -0,0 +1,623 @@
...
+    R_h = FunctionSpace(mesh, "CG", k - 1)          # Density reciprocal space
+    L_h = FunctionSpace(mesh, "R", 0)               # Lagrange multiplier space
+
+    # The mixed space
+    Z_h = J_h * J_h * V_h * U_h * U_h * P_h * X_h * X_h * R_h * L_h * L_h
+    PETSc.Sys.Print(f"Mesh has {mesh.num_cells()} cells, with {Z_h.dim()} finite element DOFs")
+
+    # The trial functions
+    solution = Function(Z_h)
+    J_1, J_2, v, mu_aux_1, mu_aux_2, p, x_1, x_2, rho_inv, l_1, l_2 = split(solution)
```
pbrubeck, line 109: Is there any good reason why `v` is the solution and `u` is the test function? It is usually the other way around.

demos/multicomponent/multicomponent.py.rst
```diff
@@ -0,0 +1,623 @@
...
+    p_ref = eta * v_ref / L_ref     # Reference pressure, Pa
+
+The non-dimensionalised variational formulation of :ref:`eq. 1 <eq:stokes>` is then::
+
+    # The Stokes viscous terms
+    stokes_terms = 2.0 * inner(sym(grad(v)), sym(grad(u))) * dx
+    stokes_terms += lame_ND * inner(div(v), div(u)) * dx
+
+    # The Stokes pressure term
+    stokes_terms -= inner(p, div(u)) * dx
```
pbrubeck, line 172: Since you mentioned Stokes, I would expect a constraint on the divergence, so a term like `-inner(div(v), q)*dx`. Maybe it'd be good to add a comment on why you only grab the momentum balance equation.

demos/multicomponent/multicomponent.py.rst
```diff
@@ -0,0 +1,623 @@
...
+    osm_terms += ((Me * inner(p, div(rho_inv * (W_1 + W_2))))) * dx
+
+    # The chemical potential terms
+    osm_terms -= ((1.0 / M_1_ND) * inner(mu_1, div(W_1)) + (1.0 / M_2_ND) * inner(mu_2, div(W_2))) * dx
+
+Lastly, we weakly enforce that
+:math:`\nabla \cdot v = \nabla \cdot (\frac{1}{\rho} \sum_{j=1}^n J_j )`,
+using special density consistency terms to handle inhomogeneous BCs::
+
+    div_mass_avg_terms = inner(q, div(v - (rho_inv * (J_1 + J_2)))) * dx
```
pbrubeck: Here's the divergence constraint, maybe it'd be good to move this all the way up where Stokes is discussed (it's a long file!). 
Also, have the Stokes pair `V_h*P_h` grouped together in the `MixedFunctionSpace`

Please put the `TestFunction` as the second argument of `inner`
```suggestion
    div_mass_avg_terms = inner(div(v - (rho_inv * (J_1 + J_2))), q) * dx
```

demos/multicomponent/multicomponent.py.rst
```diff
@@ -0,0 +1,623 @@
...
+    J_h = FunctionSpace(mesh, "BDM", k)             # Species mass-flux space
+    V_h = VectorFunctionSpace(mesh, "CG", k)        # Velocity space
+    U_h = FunctionSpace(mesh, "DG", k - 1)          # Species chemical potential space
+    P_h = FunctionSpace(mesh, "CG", k - 1)          # Pressure space
+    X_h = FunctionSpace(mesh, "DG", k - 1)          # Species mole fraction space
+    R_h = FunctionSpace(mesh, "CG", k - 1)          # Density reciprocal space
+    L_h = FunctionSpace(mesh, "R", 0)               # Lagrange multiplier space
+
+    # The mixed space
+    Z_h = J_h * J_h * V_h * U_h * U_h * P_h * X_h * X_h * R_h * L_h * L_h
```
pbrubeck, line 104: There are 11 fields. Are we describing the equations in the same order as the one given in the `MixedFunctionSpace`?

demos/multicomponent/multicomponent.py.rst
```diff
@@ -0,0 +1,623 @@
...
+~~~~~~~~~~~~~~~~~~~
+
+Let :math:`N` denote the outward unit normal on :math:`\partial \Omega`.
+We enforce parabolic profiles on :math:`J_i \cdot N` at inflow :math:`i` and on the outflow.
+The magnitudes of the parabolic profiles are :math:`M_i c_i^\text{ref} v_i^\text{ref}` where
+:math:`v_i^\text{ref}` are reference velocities that we are free to choose.
+Elsewhere on the boundary we enforce :math:`J_i \cdot N = 0`. Finally, instead of specifying
+the value of the barycentric velocity :math:`v` on the inflows and outflow, we enforce :math:`\rho v \cdot N = (J_1 + J_2 )\cdot N`
+and :math:`\rho v \times N = 0` in these regions. To enforce that :math:`v` is equal to an unknown quantity,
+we must use :class:`~.EquationBC` instead of :class:`~.DirichletBC`. ::
```
pbrubeck:
```suggestion
and :math:`\rho v \times N = 0` in these regions. 
These boundary conditions combine several unknowns and cannot be applied dof-wise since the components of  :math:`v` do not align with :math:`N`. Therefore, we must use :class:`~.EquationBC` instead of :class:`~.DirichletBC`. ::
```

demos/multicomponent/multicomponent.py.rst
```diff
@@ -0,0 +1,623 @@
...
+    # The boundary conditions on the fluxes and barycentric velocity
+    # Note that BCs on H(div) spaces only apply to the normal component
+    flux_bcs = [DirichletBC(Z_h.sub(0), J_1_inflow_bc_func, inlet_1_id),
+                DirichletBC(Z_h.sub(0), J_1_outflow_bc_func, outlet_id),
+                DirichletBC(Z_h.sub(0), as_vector([0.0, 0.0]), inlet_2_id),
+                DirichletBC(Z_h.sub(0), as_vector([0.0, 0.0]), walls_ids),
+                DirichletBC(Z_h.sub(1), J_2_inflow_bc_func, inlet_2_id),
+                DirichletBC(Z_h.sub(1), J_2_outflow_bc_func, outlet_id),
+                DirichletBC(Z_h.sub(1), as_vector([0.0, 0.0]), inlet_1_id),
+                DirichletBC(Z_h.sub(1), as_vector([0.0, 0.0]), walls_ids),
```
pbrubeck:
```suggestion
                DirichletBC(Z_h.sub(0), 0, inlet_2_id),
                DirichletBC(Z_h.sub(0), 0, walls_ids),
                DirichletBC(Z_h.sub(1), J_2_inflow_bc_func, inlet_2_id),
                DirichletBC(Z_h.sub(1), J_2_outflow_bc_func, outlet_id),
                DirichletBC(Z_h.sub(1), 0, inlet_1_id),
                DirichletBC(Z_h.sub(1), 0, walls_ids),
```

demos/multicomponent/multicomponent.py.rst
```diff
@@ -0,0 +1,623 @@
...
+                v_inflow_1_bc,
+                v_inflow_2_bc,
+                v_outflow_bc,
+                DirichletBC(Z_h.sub(2), as_vector([0.0, 0.0]), walls_ids)]
```
pbrubeck:
```suggestion
                DirichletBC(Z_h.sub(2), 0, walls_ids)]
```

demos/multicomponent/multicomponent.py.rst
```diff
@@ -0,0 +1,623 @@
...
+            trial = firedrake.TrialFunction(V)
+            xphi = firedrake.assemble(ufl.inner(xdist * test, xdist * trial) * ufl.dx, diagonal=True)
+            phi = firedrake.assemble(ufl.inner(test, trial) * ufl.dx, diagonal=True)
+            with xphi.dat.vec as xu, phi.dat.vec as u:
+                xu.pointwiseDivide(xu, u)
+                min_index, min_value = xu.min()     # Find the index of the DOF closest to bc_point
+
+            nodes = V.dof_dset.lgmap.applyInverse([min_index])
+            nodes = nodes[nodes >= 0]
+            return nodes
```
pbrubeck: Is the geometric location of the point important? If not, then just set the first dof

demos/multicomponent/multicomponent.py.rst
```diff
@@ -0,0 +1,623 @@
...
+
+    # Fix the auxiliary chemical potentials and pressure at a point
+    aux_point = as_vector([4, 0])   # A point on the middle of the outlet
+    aux_point_bcs = [FixAtPointBC(Z_h.sub(3), Constant(0.0), aux_point),
+                    FixAtPointBC(Z_h.sub(4), Constant(0.0), aux_point),
+                    FixAtPointBC(Z_h.sub(5), Constant(0.0), aux_point)]
```
pbrubeck:
```suggestion
    aux_point_bcs = [FixAtPointBC(Z_h.sub(3), 0, aux_point),
                    FixAtPointBC(Z_h.sub(4), 0, aux_point),
                    FixAtPointBC(Z_h.sub(5), 0, aux_point)]
```

demos/multicomponent/multicomponent.py.rst
```diff
@@ -0,0 +1,623 @@
...
+with the naive initial guess does not converge. Hence, we apply parameter continuation to :math:`v_1^\text{ref}`
+to find a better initial guess. We start by solving the problem for :math:`v_1^\text{ref}=0.4\times 10^{-6}` 
+with the naive initial guess and use its solution as initial guess for the problem with 
+:math:`v_1^\text{ref}=0.1\times 10^{-5}`. We repeat this trick with :math:`v_1^\text{ref}=0.2\times 10^{-5}`
+and :math:`v_1^\text{ref}=0.3\times 10^{-5}` before solving for :math:`v_1^\text{ref}=0.4\times 10^{-5}`. 
+We can reuse the nonlinear variational solver object each iteration, but have to reassign :code:`v_ref_1`
+and :code:`v_ref` before calling the :code:`solve()` method. Finally, we write each solution to the same 
+VTK file using the :code:`time` keyword argument. ::
+
+    from firedrake.output import VTKFile
```
pbrubeck:
```suggestion
```

demos/multicomponent/multicomponent.py.rst
```diff
@@ -0,0 +1,623 @@
...
+
+    J_1, J_2, v, mu_aux_1, mu_aux_2, p, x_1, x_2, rho_inv, l_1, l_2 = solution.subfunctions
+    x_1.interpolate(Constant(0.5))
+    x_2.interpolate(Constant(0.5))
+    rho_inv.interpolate(1.0 / ((M_1_ND * c_1) + (M_2_ND * c_2)))
+
+and define the nonlinear variational solver object, which by default uses Newton's method::
+
+    NLVP = NonlinearVariationalProblem(tot_res, solution, bcs=flux_bcs+aux_point_bcs)
+    NLVS = NonlinearVariationalSolver(NLVP)
```
pbrubeck, line 556: Would `pre_apply_bcs=False` make the problem easier to solve? Just curious, and also I have never tested this with `EquationBC`.

demos/multicomponent/multicomponent.py.rst
ABaierReinio: In my experience it's more common in the chemical engineering literature to use `v` instead of `u` for barycentric velocity (for example that is the case in all of the literature we are citing in this demo).

demos/multicomponent/multicomponent.py.rst
ABaierReinio: How about we put a sentence in the Stokes part saying that we postpone our description of the divergence constraint to later on? The reader won't be familiar with what `rho_inv` or `J_1` and `J_2` are until later on, so I think it's better to state the divergence constraint after those have been described.

demos/multicomponent/multicomponent.py.rst
```diff
@@ -0,0 +1,623 @@
...
+        mu_1_out.rename("mu_1")
+        mu_2_out.rename("mu_2")
+        p.rename("p")
+        x_1.rename("x_1")
+        x_2.rename("x_2")
+        rho_inv.rename("rho_inv")
+        rho_out.rename("rho")
+        c_tot_out.rename("c_tot")
+        c_1_out.rename("c_1")
+        c_2_out.rename("c_2")
```
pbrubeck: You can save a few lines here, also this could go outside the continuation loop
```suggestion
        names = ["J_1", "J_22", "v", "mu_aux_1", "mu_aux_2", "p", "x_1", "x_2", "rho_inv", "l_1", "l_2"]
        for field, name in enumerate(names):
            solution.subfunctions[field].rename(name)
```

demos/multicomponent/multicomponent.py.rst
```diff
@@ -0,0 +1,623 @@
...
+
+        outfile.write(J_1, J_2, v, mu_1_out, mu_2_out, p, x_1, x_2, rho_inv, \
+                        rho_out, c_tot_out, c_1_out, c_2_out, time=i)
```
pbrubeck: If you are dumping all fields, just do this
```suggestion
        outfile.write(*solution.subfunctions, time=i)
```

demos/multicomponent/multicomponent.py.rst
ABaierReinio: The Picard linearization of the SOSM problem leads to a saddle-point problem on `V x Q` where `V = J_h * J_h * V_h` and `Q = U_h * U_h * P_h`. In particular the functions in `Q` play the role of a Lagrange multiplier. For this reason I think it's most natural to leave them in this order.

demos/multicomponent/multicomponent.py.rst
ABaierReinio: I think it's better to leave the order of the function spaces in `MixedFunctionSpace` as is; see my comment below.

pbrubeck: `J_1` and `J_2` are the first fields, would it make sense to discuss them first? Then you will have everything you need when you discuss Stokes.

demos/multicomponent/multicomponent.py.rst
ABaierReinio: Genius!

demos/multicomponent/multicomponent.py.rst
ABaierReinio: We could move the discussion of Stokes to the end, although in the discussion of the OSM equations there's quite a few references to `v`, `v^ref` and `p^ref` and those are described in the Stokes section. So it will be a bit awkward either way.

(Also worth noting that when I when I introduce Stokes, I do say that it's _compressible_ Stokes -- and of course in compressible flow you usually don't have a constraint on `div(v)`. So in some sense it's almost surprising that we need to do that).

demos/multicomponent/multicomponent.py.rst
KarsKnook: The error will be slightly larger around the point. Mathematically it shouldn't make a difference. In practice, with solver tolerances it seems to matter a small amount.

demos/multicomponent/multicomponent.py.rst
KarsKnook: `pre_apply_bcs=False` does not automatically work with `EquationBC`. It requires the implementation of the `apply` method in `EquationBCSplit` which should just pass it. The nonlinear solver improves in some cases by 1 iteration. We won't include it in the demo as it requires a bug fix first.

demos/multicomponent/multicomponent.py.rst
pbrubeck: If it is not crucial for the demo, then this block should be made as simple as possible

demos/multicomponent/multicomponent.py.rst
pbrubeck: Sounds good, the fix should be just to skip `EquationBC` wherever we are trying to apply it then. Could that go in the PR for EquationBC + multigrid?

demos/multicomponent/multicomponent.py.rst
pbrubeck: Could this go outside the loop?

demos/multicomponent/multicomponent.py.rst
pbrubeck: Also, I believe this treatment of DirichletBC should reduce parameter continuation steps, or completely remove the need to do continuation if we are lucky

demos/multicomponent/multicomponent.py.rst
pbrubeck: We should comment about the fact that v.n is not a degree of freedom

KarsKnook: I don't understand why that matters

pbrubeck: You also have a BC of the form v.n=0, which does not combine different fields and is linear. However, you cannot enforce it with DirichletBC because the normal is not aligned with the Cartesian axes

demos/multicomponent/multicomponent.py.rst
KarsKnook: I don't understand why we would sacrifice the generality of `FixAtPointBC` for the sake of making the demo slightly shorter. I think other users will be copying `FixAtPointBC` directly from here and they do want to specify the location of the point.

demos/multicomponent/multicomponent.py.rst
KarsKnook: For this problem, you still need continuation and across all nonlinear solves you only use 1 less nonlinear iteration (30 vs 31)

demos/multicomponent/multicomponent.py.rst
KarsKnook: On the walls (not the inflow or outflow) n is not aligned with the Cartesian axes, but on those walls we do v=0 identically via DirichletBC. On the inflow and outflow n is aligned with the Cartesian axes, so v.n is in principle a DOF there, no?

pbrubeck: I misunderstood the description of the EquationBC. You describe the normal and tangential components separately, as `dot(v, N) = nonzero` and `cross(v, N) = 0`, but you jointly enforce them with a single `EquationBC`. Would it be better if you just say that you specify an inflow in the normal direction `v = v_inflow*N`?

pbrubeck: Or include this equation :math:` v = \rho^{-1} ((J_1 + J_2 )\cdot N) N`

demos/multicomponent/multicomponent.py.rst
```diff
@@ -0,0 +1,623 @@
...
+        # Compute c_tot and the species concentrations
+        c_tot = 1.0 / ((x_1_nm * (c_ref / c_pure_1)) + (x_2_nm * (c_ref / c_pure_2)))
+        c_1 = x_1_nm * c_tot
+        c_2 = x_2_nm * c_tot
+
+        return (c_tot, c_1, c_2)
+
+    c_tot, c_1, c_2 = conc_relation(x_1, x_2)
+
+Moreover, to express that :math:`1 / \rho^{-1} = \sum_{j=1}^n M_j c_j`, we simply take the
```
pbrubeck:
```suggestion
Moreover, to express that :math:`1 / \rho^{-1} = \rho = \sum_{j=1}^n M_j c_j`, we simply take the
```

demos/multicomponent/multicomponent.py.rst
```diff
@@ -0,0 +1,623 @@
...
+    J_2_outflow_bc_func = -M_2_ND * (v_ref_2 / v_ref) * (c_pure_2 / c_ref) * parabola_outflow
+    rho_v_outflow_bc_func = J_1_outflow_bc_func + J_2_outflow_bc_func
+
+    # Boundary conditions on the barycentric velocity are enforced via EquationBC
+    v_inflow_1_bc = EquationBC(inner(v - rho_inv * rho_v_inflow_1_bc_func, u) * ds(*inlet_1_id) == 0,
+                               solution, inlet_1_id, V=Z_h.sub(2))
+    v_inflow_2_bc = EquationBC(inner(v - rho_inv * rho_v_inflow_2_bc_func, u) * ds(*inlet_2_id) == 0,
+                               solution, inlet_2_id, V=Z_h.sub(2))
+    v_outflow_bc = EquationBC(inner(v - rho_inv * rho_v_outflow_bc_func, u) * ds(*outlet_id) == 0, solution,
+                              outlet_id, V=Z_h.sub(2))
```
pbrubeck:
```suggestion
    bc_data = {inlet_1_id: rho_v_inflow_1_bc_func, inlet_2_id: rho_v_inflow_2_bc_func, outlet_id: rho_v_outflow_bc_func}
    F_bc = sum(inner(v - rho_inv * flux, u) * ds(*subdomain) for subdomain, flux in bc_data.items())
    v_bc = EquationBC(F_bc == 0, solution, (*inlet_1_id, *inlet_2_id, *outlet_id), V=Z_h.sub(2))
```

demos/multicomponent/multicomponent.py.rst
```diff
@@ -0,0 +1,623 @@
...
+                DirichletBC(Z_h.sub(0), J_1_outflow_bc_func, outlet_id),
+                DirichletBC(Z_h.sub(0), 0, inlet_2_id),
+                DirichletBC(Z_h.sub(0), 0, walls_ids),
+                DirichletBC(Z_h.sub(1), J_2_inflow_bc_func, inlet_2_id),
+                DirichletBC(Z_h.sub(1), J_2_outflow_bc_func, outlet_id),
+                DirichletBC(Z_h.sub(1), 0, inlet_1_id),
+                DirichletBC(Z_h.sub(1), 0, walls_ids),
+                v_inflow_1_bc,
+                v_inflow_2_bc,
+                v_outflow_bc,
```
pbrubeck:
```suggestion
                v_bc,
```

demos/multicomponent/multicomponent.py.rst
```diff
@@ -0,0 +1,623 @@
...
+
+Lastly, we weakly enforce that
+:math:`\nabla \cdot v = \nabla \cdot (\frac{1}{\rho} \sum_{j=1}^n J_j )`,
+using special density consistency terms to handle inhomogeneous BCs::
+
+    div_mass_avg_terms = inner(div(v - (rho_inv * (J_1 + J_2))), q) * dx
+
+    # The density consistency terms
+    N = FacetNormal(mesh)
+    div_mass_avg_terms -= inner(dot(v - (rho_inv * (J_1 + J_2)), N), q) * ds
```
pbrubeck, line 405: The surface terms appear on the entire boundary, but I think we do not need to add them where we impose essential BCs on `dot(v - (rho_inv * (J_1 + J_2)), N)`, as `EquationBC` or `DirichletBC` will override the rows of the Jacobian

demos/multicomponent/multicomponent.py.rst
```diff
@@ -0,0 +1,623 @@
...
+    D_sm = Constant(2.1e-9)                         # Stefan--Maxwell diffusivity, m^2 / s
+    Pe = v_ref * L_ref / D_sm                       # Péclet number, dimensionless
+    Me = p_ref / (RT * c_ref)                       # Pressure diffusion number, dimensionless
+
+    gamma = Constant(1e-1)                          # Augmentation parameter, dimensionless
+
+    # The Stefan--Maxwell diffusion terms
+    osm_terms = Pe * (1.0 / c_tot) * ((c_2 / (M_1_ND * M_1_ND * c_1)) * inner(J_1, W_1) \
+            + (c_1 / (M_2_ND * M_2_ND * c_2)) * inner(J_2, W_2) \
+            - (1.0 / (M_1_ND * M_2_ND)) * (inner(J_1, W_2) + inner(J_2, W_1))) * dx
```
pbrubeck:
```suggestion
    osm_terms = (Pe / c_tot) * ((c_2 / (M_1_ND * M_1_ND * c_1)) * inner(J_1, W_1) 
                                                    + (c_1 / (M_2_ND * M_2_ND * c_2)) * inner(J_2, W_2)
                                                    - (1 / (M_1_ND * M_2_ND)) * (inner(J_1, W_2) + inner(J_2, W_1))) * dx
```

demos/multicomponent/multicomponent.py.rst
```diff
@@ -0,0 +1,623 @@
...
+    # Margules model parameters
+    A_12 = Constant(0.4498)         # Dimensionless
+    A_21 = Constant(0.4952)         # Dimensionless
+
+    # Margules model for chemical potentials, assuming constant partial molar volumes
+    def mu_relation(x_1, x_2, p):
+        mu_1 = (Me_1 * p) + ln(x_1) \
+            + (x_2 ** 2) * (A_12 + (2.0 * (A_21 - A_12) * x_1))
+        mu_2 = (Me_2 * p) + ln(x_2) \
+            + (x_1 ** 2) * (A_21 + (2.0 * (A_12 - A_21) * x_2))
```
pbrubeck:
```suggestion
        mu_1 = (Me_1 * p) + ln(x_1) + (x_2 ** 2) * (A_12 + (2.0 * (A_21 - A_12) * x_1))
        mu_2 = (Me_2 * p) + ln(x_2) + (x_1 ** 2) * (A_21 + (2.0 * (A_12 - A_21) * x_2))
```

demos/multicomponent/multicomponent.py.rst
```diff
@@ -0,0 +1,623 @@
...
+
+.. math::
+
+    \nabla (x_1 + \ldots + x_n) = 0 \quad \textrm{in}\ \Omega.
+
+Hence, at the discrete level, we expect :math:`x_1 + \ldots + x_n` to approximately be a constant.
+However, we have not yet incorporated any equations to make this constant be one.
+We accomplish this by enforcing that :math:`\int_{\Omega} (x_1 + \ldots + x_n - 1) \ {\rm d} x = 0`::
+
+    tot_res += inner(x_1 + x_2 - 1, s_1) * dx
```
pbrubeck, line 483: I think this could be acomplished through a nullspace. The Newton solver iterates should not change the component in the nullspace basis of your initial guess.
 ```
 nullspace_basis = Function(Z_h)
 nullspace_basis.subfunctions[index_x1].interpolate(1)
 nullspace_basis.subfunctions[index_x2].interpolate(1)
 nullspace = VectorSpaceBasis([nullspace_basis])
 nullspace.orthonormalize()
```

File: demos/multicomponent/multicomponent.py.rst [hunk 26]
```diff
@@ -0,0 +1,623 @@
...
+to find a better initial guess. We start by solving the problem for :math:`v_1^\text{ref}=0.4\times 10^{-6}` 
+with the naive initial guess and use its solution as initial guess for the problem with 
+:math:`v_1^\text{ref}=0.1\times 10^{-5}`. We repeat this trick with :math:`v_1^\text{ref}=0.2\times 10^{-5}`
+and :math:`v_1^\text{ref}=0.3\times 10^{-5}` before solving for :math:`v_1^\text{ref}=0.4\times 10^{-5}`. 
+We can reuse the nonlinear variational solver object each iteration, but have to reassign :code:`v_ref_1`
+and :code:`v_ref` before calling the :code:`solve()` method. Finally, we write each solution to the same 
+VTK file using the :code:`time` keyword argument. ::
+
+    outfile = VTKFile("out/solution.pvd")
+    cont_vals = [1.0, 2.5, 5, 7.5, 10.0]
```
pbrubeck:
```suggestion
    vmax = 0.4e-5
    cont_vals = [0.1 * vmax, 0.25 * vmax, 0.5 * vmax, 0.75 * vmax, vmax]
```

demos/multicomponent/multicomponent.py.rst
```diff
@@ -0,0 +1,623 @@
...
+    mu_1_out.rename("mu_1")
+    mu_2_out.rename("mu_2")
+    rho_out.rename("rho")
+    c_tot_out.rename("c_tot")
+    c_1_out.rename("c_1")
+    c_2_out.rename("c_2")
+    
+    for i in range(n_cont):
+        print(f"Solving for v_ref_1 = {0.4e-6*cont_vals[i]}")
+        v_ref_1.assign(Constant(0.4e-6*cont_vals[i]))
```
pbrubeck:
```suggestion
        v_ref_1.assign(cont_vals[i])
        print(f"Solving for v_ref_1 = {float(v_ref_1)}")
```

demos/multicomponent/multicomponent.py.rst
```diff
@@ -0,0 +1,623 @@
...
+    c_tot_out = Function(R_h)
+    c_1_out = Function(R_h)
+    c_2_out = Function(R_h)
+     
+    mu_1_out.rename("mu_1")
+    mu_2_out.rename("mu_2")
+    rho_out.rename("rho")
+    c_tot_out.rename("c_tot")
+    c_1_out.rename("c_1")
+    c_2_out.rename("c_2")
```
pbrubeck:
```suggestion
    mu_1_out = Function(U_h, name="mu_1")
    mu_2_out = Function(U_h, name="mu_2")
    rho_out = Function(R_h, name="rho")
    c_tot_out = Function(R_h, name="c_tot")
    c_1_out = Function(R_h, name="c_1")
    c_2_out = Function(R_h, name="c_2")
```

demos/multicomponent/multicomponent.py.rst
pbrubeck: This needs to be a fully supported feature if we want users to copy it in their code. I will resolve this thread when we change the implementation to the one with VertexOnlyMesh

demos/multicomponent/multicomponent.py.rst
```diff
@@ -0,0 +1,623 @@
...
+    P_h = FunctionSpace(mesh, "CG", k - 1)          # Pressure space
+    X_h = FunctionSpace(mesh, "DG", k - 1)          # Species mole fraction space
+    R_h = FunctionSpace(mesh, "CG", k - 1)          # Density reciprocal space
+    L_h = FunctionSpace(mesh, "R", 0)               # Lagrange multiplier space
+
+    # The mixed space
+    Z_h = J_h * J_h * V_h * U_h * U_h * P_h * X_h * X_h * R_h * L_h * L_h
+    PETSc.Sys.Print(f"Mesh has {mesh.num_cells()} cells, with {Z_h.dim()} finite element DOFs")
+
+    # The trial functions
```
pbrubeck:
```suggestion
    # The discrete solution
```

demos/multicomponent/multicomponent.py.rst
```diff
@@ -0,0 +1,623 @@
...
+Boundary conditions
+~~~~~~~~~~~~~~~~~~~
+
+Let :math:`N` denote the outward unit normal on :math:`\partial \Omega`.
+We enforce parabolic profiles on :math:`J_i \cdot N` at inflow :math:`i` and on the outflow.
+The magnitudes of the parabolic profiles are :math:`M_i c_i^\text{ref} v_i^\text{ref}` where
+:math:`v_i^\text{ref}` are reference velocities that we are free to choose.
+Elsewhere on the boundary we enforce :math:`J_i \cdot N = 0`. Finally, instead of specifying
+the value of the barycentric velocity :math:`v` on the inflows and outflows, 
+we enforce :math:`v = \rho^{-1}((J_1 + J_2)\cdot N)N`. Boundary conditions that couple
```
pbrubeck: The inflow boundaries are normal to the y-axis. However the parabolic inflow has a component in the x-axis. This conflicts with the claim that you had before `cross(v, N) = 0`, and the current statement `v = (scalar) N`.

demos/multicomponent/multicomponent.py.rst
pbrubeck: Since you impose essential BCs everywhere on the boundary I think you don't need this term

demos/multicomponent/multicomponent.py.rst
```diff
@@ -0,0 +1,626 @@
...
+           v = TestFunction(V)
+           F = assemble(Interpolate(inner(v, v), Fvom))
+           with F.dat.vec as Fvec:
+               max_index, _ = Fvec.max()
+           nodes = V.dof_dset.lgmap.applyInverse([max_index])
+           nodes = nodes[nodes >= 0]
+           return nodes
+
+    # Fix the auxiliary chemical potentials and pressure at a point
+    aux_point = as_vector([4, 0])   # A point on the middle of the outlet
```
pbrubeck:
```suggestion
    aux_point = (4, 0)   # A point on the middle of the outlet
```

demos/boussinesq/boussinesq.py.rst
```diff
@@ -160,40 +160,43 @@ cannot always be sure that the linear solver at hand is correctly utilising the
...
+           F = assemble(Interpolate(inner(v, v), Fvom))
+           with F.dat.vec as Fvec:
+               max_index, _ = Fvec.max()
+           nodes = V.dof_dset.lgmap.applyInverse([max_index])
+           nodes = nodes[nodes >= 0]
+           return nodes
+ 
 We use this to fix the pressure and auxiliary temperature at the origin::
 
     aux_bcs = [FixAtPointBC(Z.sub(1), 0, as_vector([0, 0])),
```
pbrubeck: The point does not need to be an UFL expression. Also in the line below, 
```suggestion
    aux_bcs = [FixAtPointBC(Z.sub(1), 0, (0, 0)), 
```

demos/multicomponent/multicomponent.py.rst
ABaierReinio: Will it be good enough for now if we just set the first DOF? If so, I'll make that change.

(Note that we already have the full `FixAtPointBC` code in the [Boussinesq demo](https://www.firedrakeproject.org/demos/boussinesq.py.html)).

demos/multicomponent/multicomponent.py.rst
ABaierReinio: `EquationBC` or `DirichletBC` will not override these rows of the Jacobian. Note that in these terms the test function is `q` and so these rows would only be overridden if we imposed BCs on `p` (which we aren't doing).

These terms are close to but not exactly zero. We still need them. Indeed our `EquationBC` is enforcing that `inner(v - (rho_inv * (J_1 + J_2)), u) * ds = 0` for all `CG_k` functions `u`, but here we are testing against `q * N` which is not in `CG_k` because of `N`. There will be nonzero contributions from these terms in places where `N` is discontinuous.

demos/multicomponent/multicomponent.py.rst
ABaierReinio: In my experience the discretized SOSM equations enforce that

`x_1 + ... + x_n = C + epsilon`

where `C` is a constant (not spatially varying), and `epsilon` is some small spatially varying error (it's probably on the order of `h^k` with `h` the mesh size and `k` the polynomial degree). So the nullspace basis you constructed here definitely isn't exactly in the nullspace of the Jacobian; I don't think this would work.

I don't think we know what the nullspace of the Jacobian is. We do know (although for brevity we don't elaborate on this in the demo) which __rows__ of the Jacobian are linearly dependent -- in other words we know the nullspace of the transpose of the Jacobian. But since the Jacobian isn't symmetric we don't know what its nullspace is.

`FixAtPointBC` is nice because when we use it with Lagrange multipliers it lets us replace these linearly dependent rows with integral constraints.

ABaierReinio: Pretend we don't use `FixAtPointBC` or Lagrange multipliers. Then the Jacobian (for this demo) has three rows (we know which ones they are!) that are linearly dependent on the other rows. We want to replace two of these rows with integral constraints, as this yields a well-posed problem. The other row corresponds to the pressure nullspace and we could get rid of it using PETSc nullspace or `FixAtPointBC`.

The whole purpose of these Lagrange multipliers and `FixAtPointBC` is to replace those two rows with our desired integral constraints. I don't see how we could do this using `nullspace`. I used to do it with the Woodbury identity instead but it's a real pain to implement.

demos/multicomponent/multicomponent.py.rst
ABaierReinio: You're right, thanks for catching that -- what we wrote here is correct for the 3D Benzene simulation from my paper but I guess we forgot that the BCs are slightly different in this 2D simulation.

I just pushed a commit that rephrases this paragraph, please resolve if you're happy with it.

pbrubeck (approved):
LGTM

ABaierReinio:
@pbrubeck Thanks for all the suggestions. We are currently failing the tests because, when building the documentation, latex isn't aware of `\mathscr`. Can we get it to `\usepackage{mathrsfs}` (maybe by modifying https://github.com/firedrakeproject/firedrake/blob/master/docs/source/conf.py)? Alternatively we can use a different symbol font.

demos/multicomponent/multicomponent.py.rst
```diff
@@ -0,0 +1,618 @@
...
+
+    # Construct the mesh
+    ngmesh = ngocc.OCCGeometry(Omega, dim=2).GenerateMesh(maxh=maxh)
+    mesh = Mesh(Mesh(ngmesh).curve_field(k))
+
+    # Get the IDs of the inlets, outlet and walls
+    inlet_1_id = tuple([i+1 for i, name in enumerate(ngmesh.GetRegionNames(dim=1)) if name == "inlet_1"])
+    inlet_2_id = tuple([i+1 for i, name in enumerate(ngmesh.GetRegionNames(dim=1)) if name == "inlet_2"])
+    outlet_id = tuple([i+1 for i, name in enumerate(ngmesh.GetRegionNames(dim=1)) if name == "outlet"])
+    walls_ids = tuple([i+1 for i, name in enumerate(ngmesh.GetRegionNames(dim=1)) if name == ""])
```
pbrubeck:
```suggestion
    inlet_1_id = tuple(i+1 for i, name in enumerate(ngmesh.GetRegionNames(dim=1)) if name == "inlet_1")
    inlet_2_id = tuple(i+1 for i, name in enumerate(ngmesh.GetRegionNames(dim=1)) if name == "inlet_2")
    outlet_id = tuple(i+1 for i, name in enumerate(ngmesh.GetRegionNames(dim=1)) if name == "outlet")
    walls_ids = tuple(i+1 for i, name in enumerate(ngmesh.GetRegionNames(dim=1)) if name == "")
```

pbrubeck (approved):
This looks great. Thanks for your contribution!

PR #4327, opened 2025-05-14
add meson-python requirement to requirements-build.txt

leo-collins:
# Description

Adds [meson-python](https://pypi.org/project/meson-python/) to `requirements-build.txt`.


connorjward (approved):
Thanks! To be clear this is getting added because some upstream package has taken on meson-python as a new build dependency and so `--no-build-isolation` is currently not working.

Issue #4239, opened 2025-04-17
VTKFile to support printing HDivTrace functions


adofarsi:
**Describe the current issue**
Would it be possible to extend VTKFile to support printing HDivTrace functions on facets?
Currently, the .write() method returns the following error from interpolation.py:
`NotImplementedError: Can't interpolate onto traces, sorry.`

**Describe the solution you'd like**
For example, printing such HDivTrace function for a triangular mesh would result in line elements, while for a tetrahedral mesh, it would produce triangular elements.

dham:
I think if you use dmplex filter to extract the codimension 1 entities you can construct the skeleton mesh of a mesh. At that point you could just declare an appropriate DG field on that mesh and use cross-mesh interpolation. @ksagiyam may be able to advise on filtering the dmplex.

dham:
OK, the first thing is that you want to project into Hdiv trace rather than interpolate.

To get from the volume mesh to the skeleton I think you just interpolate. Recall that it won't be HDiv trace on the skeleton but rather DG.

adofarsi:
Hi @pohlan! Unfortunately I still haven’t found a solution. I’ll continue looking into it and let you know as soon as I make any progress.

ksagiyam:
https://github.com/firedrakeproject/firedrake/pull/4329 has now been merged. You should now be able to make a skeleton mesh if the parent mesh is simplex. Please see an example here https://github.com/firedrakeproject/firedrake/blob/5144bf85ee557d62d87d6844bd82d39ccdfc4389/tests/firedrake/submesh/test_submesh_interpolate.py#L195.

PR #4329, opened 2025-05-18
submesh: allow for extracting codim-1 submeshes

ksagiyam:
Partially resolve https://github.com/firedrakeproject/firedrake/issues/4239.

### DIFF
--- a/firedrake/cython/dmcommon.pyx
+++ b/firedrake/cython/dmcommon.pyx
@@ -1,6 +1,7 @@
 # cython: language_level=3
 
 # Utility functions common to all DMs used in Firedrake
+import enum
 import functools
 import math
 import cython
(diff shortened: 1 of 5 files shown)


ksagiyam:
[test.pdf](https://github.com/user-attachments/files/20270067/test.pdf)

firedrake/cython/dmcommon.pyx
```diff
@@ -3823,13 +3976,16 @@ def submesh_update_facet_labels(PETSc.DM dm, PETSc.DM subdm):
         PETSc.DMLabel sub_int_facet_label, sub_ext_facet_label
         PetscBool has_point
 
+    dim = dm.getDimension()
+    subdim = subdm.getDimension()
+    if subdim != dim:
+        # TODO: Do something useful if necessary.
```
JHopeCollins, line 3982: Please can you add more detail here describing when you might hit this condition and what the useful something might be?

dham (changes requested):
Please document.

firedrake/cython/dmcommon.pyx
ksagiyam: We hit this whenever `subdim != dim`, e.g. when making a co-dim 1 submesh. We currently do nothing in such case. I added a one-line comment to explain more.

ksagiyam:
I assume that documentation is now fine, so I will go ahead and merge this.

Issue #4296, opened 2025-05-09
Unable to define a single R space when using an extruded mesh


JamesJackaman:
If discretising using a extruded mesh it's not currently possible to define a single R space. 

Defining 
`FunctionSpace(mesh,'R',degree=0,vfamily='R',vdegree=0)` builds a function space with DOFs equal to the number of the extruded dimension, instead of just one DOF.

If instead I define it with `FunctionSpace(mesh,'R',degree=0)` I run into errors in the solve command which are likely caused by the element type not being supported (see code below).

```
from firedrake import *

base = IntervalMesh(100,1)
mesh = ExtrudedMesh(base,layers=100,layer_height=1/100,periodic=False)

U = FunctionSpace(mesh,TensorProductElement(FiniteElement("CG",interval,degree=1),FiniteElement("CG",interval,degree=1)))

R = FunctionSpace(mesh,'R',degree=0)

V = MixedFunctionSpace((U,R))

v = Function(V)
u, r = v.subfunctions
phi, psi = TestFunctions(V)

r.assign(100)

u, r = split(v)

F = u*phi*dx + r*psi*dx

solver_parameters = {"mat_type": "matfree",
                     "pc_type": "fieldsplit",
                     "pc_fieldsplit_type": "additive",
                     "pc_fieldsplit_0_fields": "0",
                     "pc_fieldsplit_1_fields": "1",                         
                     "fieldsplit_0": {"pc_type": "python",
                                      "pc_python_type": "firedrake.AssembledPC",
                                      "assembled_pc_type": "lu",
                                      "ksp_type": "preonly"},
                     "fieldsplit_1": {"ksp_type": "preonly",
                                      "pc_type": "python",
                                      "pc_python_type":"firedrake.AssembledPC",
                                      "assembled_pc_type":"jacobi"},
                     "ksp_monitor_true_residual": None,
                     "pc_factor_mat_solver_type": "mumps",
                     "pc_factor_shift_type":"nonzero",
                     "snes_monitor": None,
                     }

nvp = NonlinearVariationalProblem(F, v)
nvs = NonlinearVariationalSolver(nvp,solver_parameters=solver_parameters)
nvs.solve()

```

PR #4331, opened 2025-05-19
Fix Real space assembly

pbrubeck:
# Description
Fixes https://github.com/firedrakeproject/firedrake/issues/4296

### DIFF
--- a/tsfc/kernel_interface/common.py
+++ b/tsfc/kernel_interface/common.py
@@ -16,6 +16,7 @@ from gem.optimise import remove_componenttensors as prune
 from numpy import asarray
 from tsfc import fem, ufl_utils
 from finat.element_factory import as_fiat_cell, create_element
+from finat.ufl import MixedElement
 from tsfc.kernel_interface import KernelInterface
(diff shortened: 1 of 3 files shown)


connorjward (changes requested):

tsfc/kernel_interface/common.py
```diff
@@ -298,10 +299,16 @@ def set_quad_rule(params, cell, integral_type, functions):
     # Check if the integral has a quad degree or quad element attached,
     # otherwise use the estimated polynomial degree attached by compute_form_data
     quad_rule = params.get("quadrature_rule", "default")
+    elements = []
```
pbrubeck, line 302: It seems like mixed coefficients are not split at this point if we have an extruded mesh. Do we want to instead make sure that they are properly split?

dham (approved):

tsfc/kernel_interface/common.py
connorjward: I can totally believe that that's a bug. Feel free to open an issue or PR about it.

PR #4341, opened 2025-05-28
Allow construction of interpolation matrix between a `VertexOnlyMesh` and its input-ordering

leo-collins:
# Description

Allows construction of the interpolation matrix between a VOM and its input-ordering. This is a permutation matrix which is used for cross-mesh interpolation to redistribute the points correctly on the target mesh.

Set the keyword argument `matfree=False` to construct the matrix. By default `matfree=True` which uses the current way of SF reduce and broadcasts.

```
P0DG = FunctionSpace(vom, "DG", 0)
P0DG_io = FunctionSpace(vom.input_ordering, "DG", 0)
B = assemble(interpolate(TestFunction(P0DG), P0DG_io, matfree=False))
```

### DIFF
--- a/firedrake/interpolation.py
+++ b/firedrake/interpolation.py
@@ -182,7 +182,7 @@ def interpolate(expr, V, *args, **kwargs):
         to zero. Ignored if interpolating within the same mesh or onto a
         :func:`.VertexOnlyMesh`.
     :kwarg ad_block_tag: An optional string for tagging the resulting assemble block on the Pyadjoint tape.
-    :returns: A synbolic :class:`.Interpolate` object
+    :returns: A symbolic :class:`.Interpolate` object
 
     .. note::
 
--- a/firedrake/interpolation.py
+++ b/firedrake/interpolation.py
@@ -48,7 +48,8 @@ class Interpolate(ufl.Interpolate):
                  subset=None,
                  access=op2.WRITE,
                  allow_missing_dofs=False,
-                 default_missing_val=None):
+                 default_missing_val=None,
+                 matfree=True):
         """Symbolic representation of the interpolation operator.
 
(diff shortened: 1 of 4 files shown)


firedrake/interpolation.py
```diff
@@ -1667,3 +1701,26 @@ def multHermitian(self, source_vec, target_vec):
...
+        source_size = (nroots, sum(local_sizes))
+        target_size = (nleaves, self.V.comm.allreduce(nleaves, op=MPI.SUM))
+        mat = PETSc.Mat().createAIJ((target_size, source_size), nnz=1, comm=self.V.comm)
+        mat.setUp()
+        start = sum(local_sizes[:self.V.comm.rank])
+        end = start + nroots
+        contiguous_indices = numpy.arange(start, end, dtype=numpy.int32)
+        perm = numpy.zeros(target_size[0], dtype=numpy.int32)
+        self.sf.bcastBegin(MPI.INT, contiguous_indices, perm, MPI.REPLACE)
+        self.sf.bcastEnd(MPI.INT, contiguous_indices, perm, MPI.REPLACE)
```
JHopeCollins, line 1720: These should be changed to the PETSc provided MPI Int type https://petsc.org/release/manualpages/Sys/MPIU_INT/

leo-collins: Where do I find this? It's not in `petsc4py.PETSc`

connorjward: I think `MPI.INT` is the correct choice going by [the documentation](https://petsc.org/main/petsc4py/reference/petsc4py.PETSc.SF.html#petsc4py.PETSc.SF.bcastBegin).

dham (approved):

PR #4344, opened 2025-05-29
Proposal: use upstream UFL instead of our fork

connorjward:
I don't think that we should be using our own fork of UFL any more. It means that we have to keep updating it periodically and it doesn't affect the user experience anyway since Firedrake `release` depends on a released version of UFL. We already point Firedrake `master` to PETSc `main` which is a much faster moving target than UFL `main`.

Goes with https://github.com/firedrakeproject/fiat/pull/147

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -21,7 +21,8 @@ dependencies = [
   "mpi4py; python_version < '3.13'",
   # TODO RELEASE: use releases
   "fenics-ufl @ git+https://github.com/FEniCS/ufl.git",
-  "firedrake-fiat @ git+https://github.com/firedrakeproject/fiat.git",
+  # UNDO ME
+  "firedrake-fiat @ git+https://github.com/firedrakeproject/fiat.git@connorjward/upstream-ufl",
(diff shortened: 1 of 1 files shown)


pyproject.toml
```diff
@@ -20,8 +20,9 @@ dependencies = [
   "mpi4py>3; python_version >= '3.13'",
   "mpi4py; python_version < '3.13'",
   # TODO RELEASE: use releases
-  "fenics-ufl @ git+https://github.com/firedrakeproject/ufl.git",
-  "firedrake-fiat @ git+https://github.com/firedrakeproject/fiat.git",
+  "fenics-ufl @ git+https://github.com/FEniCS/ufl.git",
+  # UNDO ME
+  "firedrake-fiat @ git+https://github.com/firedrakeproject/fiat.git@connorjward/upstream-ufl",
```
```suggestion
  "firedrake-fiat @ git+https://github.com/firedrakeproject/fiat.git",
```

connorjward:
Pushing through as this was passing tests (before merging to FIAT `master`) and verbally approved in yesterday's meeting.

PR #4346, opened 2025-05-30
Move `__future__` interpolation into `interpolation.py`

leo-collins:
# Description

Moves new interpolation behaviour from `firedrake.__future__` into `firedrake.interpolation` and updated docs and demos.

### DIFF
--- a/firedrake/interpolation.py
+++ b/firedrake/interpolation.py
@@ -149,6 +149,7 @@ def interpolate(expr, V, *args, **kwargs):
     :arg expr: a UFL expression.
     :arg V: the :class:`.FunctionSpace` to interpolate into (or else
         an existing :class:`.Function` or :class:`.Cofunction`).
+        Adjoint interpolation requires ``V`` to be a :class:`.Cofunction`.
     :kwarg subset: An optional :class:`pyop2.types.set.Subset` to apply the
         interpolation over. Cannot, at present, be used when interpolating
         across meshes unless the target mesh is a :func:`.VertexOnlyMesh`.


firedrake/__future__.py
connorjward, line 1: I don't know the migration strategy here but doing this means that
```
from firedrake.__future__ import interpolate
```
will now break for users.

dham: Good point. `__future__.interpolate` should now be a wrapper function for `interpolation.interpolate` that spews a warning telling people to remove their `__future__` imports.

dham (approved):

firedrake/interpolation.py
```diff
@@ -142,20 +143,12 @@ def _ufl_expr_reconstruct_(self, expr, v=None, **interp_data):
...
-    ad_block_tag=None
-):
-    """Interpolate an expression onto a new function in V.
+def interpolate(expr, V, *args, **kwargs):
+    """Returns a UFL expression for the interpolation operation of ``expr`` into ``V``.
 
     :arg expr: a UFL expression.
     :arg V: the :class:`.FunctionSpace` to interpolate into (or else
-        an existing :class:`.Function`).
+        an existing :class:`.Function` or :class:`.Cofunction`).
```
pbrubeck, line 151: I thought `expr` could only be primal. If we want to interpolate a `Cofunction` (or a `BaseForm`) we can pass it to `V` as the second arg, and we would have to pass a `TestFunction` to `expr` in the first arg.

pbrubeck: Nevermind the docstring seems okay

pbrubeck: We should probably clarify that the adjoint interpolation takes V=Cofunction as the input

firedrake/interpolation.py
```diff
@@ -182,8 +176,7 @@ def interpolate(
         to zero. Ignored if interpolating within the same mesh or onto a
         :func:`.VertexOnlyMesh`.
     :kwarg ad_block_tag: An optional string for tagging the resulting assemble block on the Pyadjoint tape.
-    :returns: a new :class:`.Function` in the space ``V`` (or ``V`` if
-        it was a Function).
+    :returns: A synbolic :class:`.Interpolate` object
```
pbrubeck, line 179: Typo

firedrake/preconditioners/hypre_ams.py
```diff
@@ -1,14 +1,11 @@
-import firedrake.assemble as assemble
+import firedrake as fd
```
pbrubeck, line 1: I am not sure whether we want to `import firedrake as fd` within the source code. Usually we resolve circular imports with `assemble` by importing it inside the functions.

PR #4348, opened 2025-06-02
PETSc update

connorjward:
Should hopefully mean that we no longer need to set `PIP_CONSTRAINTS`. This PR will have to be followed by a new release.


scripts/check-config
```diff
@@ -58,9 +59,12 @@ def check_petsc_version() -> None:
         f"petsc4py=={petsc_version}",
         2,
     )
+
+
+def check_slepc_version() -> None:
     check_file_contains_pattern(
         REPO_ROOT / "pyproject.toml",
-        f"slepc4py=={petsc_version}",
+        "slepc4py==(.*)",
```
connorjward, line 67: This is necessary because the slepc4py and petsc4py version numbers no longer agree.

ksagiyam (approved):

PR #4284, opened 2025-05-05
checkpointing bcs

Ig-dolci:
# Description

This PR handles the issue with the incorrect checkpointing of Dirichlet boundary conditions. Currently, during tape recomputation, the user-defined boundary condition (``DirichletBC._original_arg``) is being updated. This leads to the [issue #4206](https://github.com/firedrakeproject/firedrake/issues/4206).
In that issue, James suggested an approach to checkpoint the Dirichlet boundary condition. This PR implements that suggestion and includes a test to verify the fix.
This change fixes [issue #4206](https://github.com/firedrakeproject/firedrake/issues/4206).

PR #4351, opened 2025-06-04
Backport BC checkpointing fixes to release

angus-g:
Maybe this is as much a test of the bugfix process as anything, but it would be good to have #4284 available in a release.


connorjward:
Thanks! I will push this through because it's important and the CI failures are unrelated.

PR #4352, opened 2025-06-04
ASM: support Extruded Periodic meshes

pbrubeck:
# Description
This PR enables ASMExtrudedStar on extruded periodic meshes.

In addition, it fixes p-multigrid prolongation/restriction on these meshes by extending `pyop2.PermutedMap` to the periodic case.


firedrake/preconditioners/asm.py
```diff
@@ -361,7 +361,7 @@ def get_basemesh_nodes(W):
 
     # For every base mesh entity, what's the layer offset?
     layer_offsets = numpy.full(W.node_set.total_size, -1, dtype=IntType)
-    layer_offsets[W.cell_node_map().values_with_halo] = W.cell_node_map().offset
+    layer_offsets[W.cell_node_map().values_with_halo] = W.offset
     nlayers = W.mesh().layers
```
pbrubeck: @ksagiyam I need some help to fix the arrays returned by this function in the periodic case.

ksagiyam: `dof_per_layer` formula in l.376 is probably wrong for the periodic case, as `dof` in that case is `nlayers * layer_offset`.

pbrubeck: Seems like the `(nlayer-1)` is correct if you also subtract the number of dofs that we are missing from the top layer. Here I add `quotient` to `dof_per_layer`, as oppose to subtracting, because there was a minus sign in front of the dofs on the entire layer = `(nlayer-1)*layer_offset - quotient`.

pbrubeck: Maybe there is a cleaner way of doing this

PR #4354, opened 2025-06-05
Ksagiyam/rename sub elements

ksagiyam:
https://github.com/firedrakeproject/fiat/pull/148

### DIFF
--- a/firedrake/embedding.py
+++ b/firedrake/embedding.py
@@ -59,7 +59,7 @@ def get_embedding_method_for_checkpointing(element):
                               'S', 'DPC', 'Real']:
         return "interpolate"
     elif isinstance(element, finat.ufl.TensorProductElement):
-        methods = [get_embedding_method_for_checkpointing(elem) for elem in element.sub_elements]
+        methods = [get_embedding_method_for_checkpointing(elem) for elem in element.factor_elements]
         if any(method == "project" for method in methods):
             return "project"
         else:
--- a/firedrake/interpolation.py
+++ b/firedrake/interpolation.py
@@ -543,9 +543,7 @@ class CrossMeshInterpolator(Interpolator):
         # input ordering VOM will only contain the points on rank 0!
         # QUESTION: Should any of the below have annotation turned off?
         ufl_scalar_element = V_dest.ufl_element()
-        if ufl_scalar_element.num_sub_elements and not isinstance(
-            ufl_scalar_element, finat.ufl.TensorProductElement
-        ):
+        if isinstance(ufl_scalar_element, finat.ufl.MixedElement):
             if all(
                 ufl_scalar_element.sub_elements[0] == e
                 for e in ufl_scalar_element.sub_elements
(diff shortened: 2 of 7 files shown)


firedrake/embedding.py
```diff
@@ -59,7 +59,7 @@ def get_embedding_method_for_checkpointing(element):
                               'S', 'DPC', 'Real']:
         return "interpolate"
     elif isinstance(element, finat.ufl.TensorProductElement):
-        methods = [get_embedding_method_for_checkpointing(elem) for elem in element.sub_elements]
+        methods = [get_embedding_method_for_checkpointing(elem) for elem in element.tensor_product_components]
```
connorjward: Just bikeshedding: `element.component_elements`? One could argue that it is important to say 'elements' in the attribute name.

pbrubeck: These are the factors of the tensor product

ksagiyam: `factor_elements` that Pablo suggested here https://github.com/firedrakeproject/fiat/pull/148 looks good to me.

firedrake/interpolation.py
```diff
@@ -543,9 +543,7 @@ def __init__(
         # input ordering VOM will only contain the points on rank 0!
         # QUESTION: Should any of the below have annotation turned off?
         ufl_scalar_element = V_dest.ufl_element()
-        if ufl_scalar_element.num_sub_elements and not isinstance(
-            ufl_scalar_element, finat.ufl.TensorProductElement
-        ):
+        if ufl_scalar_element.num_sub_elements:
```
connorjward: This seems like a bit of a strange condition. Is `num_sub_elements` zero for non-mixed elements? It seems like it would be better for `num_sub_elements` to only be an attribute on mixed elements as it's otherwise kind of oddly defined.

firedrake/mg/embedded.py
```diff
@@ -70,8 +70,11 @@ def __init__(self, *, native_transfers=None, use_averaging=True):
     def is_native(self, element, op):
         if element in self.native_transfers.keys():
             return self.native_transfers[element][op] is not None
-        if isinstance(element.cell, ufl.TensorProductCell) and len(element.sub_elements) > 0:
-            return all(self.is_native(e, op) for e in element.sub_elements)
+        if isinstance(element.cell, ufl.TensorProductCell):
```
connorjward, line 73: Why do we check the cell type here? Can't we have an `isinstance(element, TensorProductElement)` and `isinstance(element, MixedElement)`? (I know mixed elements are sometimes harder to identify than that)

firedrake/interpolation.py
ksagiyam: Yes, this is strange. I would expect `num_sub_elements = 1` for non-mixed elements, but we are not changing that in this PR. I can probably directly check type here.

firedrake/mg/embedded.py
ksagiyam: Changed to what I believe is correct.

firedrake/interpolation.py
ksagiyam: Changed.

connorjward (dismissed):
LGTM

firedrake/mg/embedded.py
ksagiyam: After running some multigrid tests, it turned out that `TensorProductCell` is of significance. So I tidied up a little, but made it sure that the change here is strictly noop.

connorjward: Weird, but not the point of this PR.

PR #4357, opened 2025-06-06
Set setuptools min version

connorjward:
This is to make sure that the 'license' field is interpreted correctly in the pyproject.toml. See
https://packaging.python.org/en/latest/guides/writing-pyproject-toml/#license-and-license-files.

# Description


ksagiyam (approved):

connorjward:
Merging as the only failure is a slow link.

PR #4362, opened 2025-06-09
Fix pre_apply_bcs=False throwing an error with EquationBC

ABaierReinio:
When `pre_apply_bcs=False` only `apply()` the `bc` if it is a `DirichletBC`.

Also added a test case for `pre_apply_bcs=False`.

### DIFF
--- a/firedrake/solving_utils.py
+++ b/firedrake/solving_utils.py
@@ -414,6 +414,7 @@ class _SNESContext(object):
         :arg X: the current guess (a Vec)
         :arg F: the residual at X (a Vec)
         """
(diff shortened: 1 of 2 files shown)


firedrake/solving_utils.py
```diff
@@ -427,7 +428,8 @@ def form_function(snes, X, F):
         if not ctx.pre_apply_bcs:
             # Compute DirichletBC residual
             for bc in ctx.bcs_F:
```
pbrubeck, line 430: We need to grab the `DirichletBCs` coming from `extract_ditichlet_bcs`, this must be stored somewhere in `ctx`

PR #4366, opened 2025-06-10
Clarify install docs

connorjward:
* Add minimal set of packages to docs so it is clear that PETSc will download most of the external packages.
* Add link to HPC installation wiki page.

Thanks to Golo Wimmer for the suggestions.

### DIFF
--- a/docs/source/install.rst
+++ b/docs/source/install.rst
@@ -20,6 +20,11 @@ it should be installable on any Linux distribution. Windows users are encouraged
 to use WSL_ or one of Firedrake's
 :ref:`alternative installation mechanisms<alternative_install>`.
 
+If installing on an HPC system most of the following steps should remain
+applicable, though care will have to be taken to make sure that the right system
+packages are used. A community-maintained collection of instructions for how to install
+Firedrake onto a number of different HPC systems may be found
+`here <https://github.com/firedrakeproject/firedrake/wiki/HPC-installation>`__.
 
 .. _pip_install_firedrake:
 
@@ -92,9 +97,15 @@ which will install the following packages:
 .. literalinclude:: homebrew_deps.txt
    :language: text
 
-If you do not have one of these systems then these dependencies will need to
-be installed manually.
+The packages installed here are a combination of essential system dependencies,
+like a C compiler and MPI, and 'external packages' that are used by PETSc, like
+MUMPS and HDF5. If you are not installing onto Ubuntu or macOS then only the former
+of these need be installed as PETSc can download and compile the remaining packages
+(by passing ``--no-package-manager`` to PETSc ``configure``). To give you a guide
+as to what essential system dependencies are needed, on Ubuntu they are:
 
+.. literalinclude:: minimal_apt_deps.txt
+   :language: text
 
 .. _install_petsc:
 
--- a/docs/source/install.rst
+++ b/docs/source/install.rst
@@ -97,12 +97,17 @@ which will install the following packages:
 .. literalinclude:: homebrew_deps.txt
    :language: text
 
-The packages installed here are a combination of essential system dependencies,
-like a C compiler and MPI, and 'external packages' that are used by PETSc, like
-MUMPS and HDF5. If you are not installing onto Ubuntu or macOS then only the former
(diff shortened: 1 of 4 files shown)


docs/source/install.rst
```diff
@@ -92,9 +97,15 @@ which will install the following packages:
...
    :language: text
 
-If you do not have one of these systems then these dependencies will need to
-be installed manually.
+The packages installed here are a combination of essential system dependencies,
+like a C compiler and MPI, and 'external packages' that are used by PETSc, like
+MUMPS and HDF5. If you are not installing onto Ubuntu or macOS then only the former
+of these need be installed as PETSc can download and compile the remaining packages
+(by passing ``--no-package-manager`` to PETSc ``configure``). To give you a guide
+as to what essential system dependencies are needed, on Ubuntu they are:
```
connorjward:
```suggestion
The packages installed here are a combination of system dependencies,
like a C compiler, BLAS, and MPI,  and 'external packages' that are used by PETSc, like
MUMPS and HDF5.

If you are not installing onto Ubuntu or macOS then it is your responsibility to
ensure that these system dependencies are in place. Some of the dependencies
(e.g. a C compiler) must come from your system whereas others, if desired, may be
downloaded by PETSc ``configure`` by passing additional flags like
``--download-mpich`` or ``--download-openblas`` (run ``./configure --help | less`` to
see what is available). To give you a guide as to what system dependencies are
needed, on Ubuntu they are:
```

connorjward:
> Are MPI and BLAS really essential system packages? I thought we were trying to avoid enforcing particular MPI implementations. Sure it's generally better to use the system one but a different one should work fine.

Yeah I guess that that is a bit confusing. I guess I have conflated "tools needed to compile PETSc" with "tools that even PETSc cannot provide". Those sets don't quite fully overlap because of things like `--download-mpich` and `--download-openblas`. Please see my suggested change: https://github.com/firedrakeproject/firedrake/pull/4366/files#r2166511543

connorjward:
> > > Are MPI and BLAS really essential system packages? I thought we were trying to avoid enforcing particular MPI implementations. Sure it's generally better to use the system one but a different one should work fine.
> > 
> > 
> > Yeah I guess that that is a bit confusing. I guess I have conflated "tools needed to compile PETSc" with "tools that even PETSc cannot provide". Those sets don't quite fully overlap because of things like `--download-mpich` and `--download-openblas`. Please see my suggested change: [#4366 (files)](https://github.com/firedrakeproject/firedrake/pull/4366/files#r2166511543)
> 
> Thanks, I think the documentation is clearer now.
> 
> However, MPI and BLAS are still included in the `BASE_*_PACKAGES` lists in `firedrake-configure`. If we're saying we will provide a minimal list of packages that the user needs to provide because PETSc just can't provide them at all, then it really should be that minimal set. At the moment there's the caveat that some unspecified subset of the `--minimal-system-packages` (that includes at least MPI and BLAS) can actually be provided by PETSc.
> 
> The signposting in the docs would then be the other way around - pointing out that MPI and BLAS will in fact be system packages in the vast majority of cases but not by necessity.

Thanks. Addressed.

docs/source/install.rst
```diff
@@ -20,6 +20,11 @@ it should be installable on any Linux distribution. Windows users are encouraged
 to use WSL_ or one of Firedrake's
 :ref:`alternative installation mechanisms<alternative_install>`.
 
+If installing on an HPC system most of the following steps should remain
+applicable, though care will have to be taken to make sure that the right system
+packages are used. A community-maintained collection of instructions for how to install
+Firedrake onto a number of different HPC systems may be found
+`here <https://github.com/firedrakeproject/firedrake/wiki/HPC-installation>`__.
```
JHopeCollins, line 27:
```suggestion
`here <https://github.com/firedrakeproject/firedrake/wiki/HPC-installation>`__.
If install on an HPC system not included in the wiki, please consider contributing a
page describing the installation on that system.
```

JHopeCollins (approved):
Thanks for updating that. I have one small suggestion to make it clear that we're happy for anyone to contribute to the HPC docs, but I'm happy with this.

PR #4373, opened 2025-06-13
Eikehmueller/gtmgpc restriction

### LINKED ISSUE #4367: Pass restriction matrix to GTMGPC
**Describe the current issue**
At the moment, it is only possible to set the interpolation matrix `P` for `GTMGPC`. However, it might be desirable to also set the restriction matrix, which is not necessarily the transpose of `P`. For example, for the setup in [this paper](https://rmets.onlinelibrary.wiley.com/doi/pdfdirect/10.1002/qj.4515) the residual needs to be scaled by a diagonal matrix after restriction, and this could be taken care of by supplying a restriction matrix whose rows have been scaled appropriately.

**Describe the solution you'd like**
Pass the restriction matrix through an `appctx`. Add the following to the `initialise()` method of `GTMGPC`:
```
restr_petscmat = appctx.get("restriction_matrix", None)
if restr_petscmat is not None:
    pcmg.setMGRestriction(1, restr_petscmat)
``` 

**Describe alternatives you've considered**
Alternatively, to address the scaling problem described above, one could try to pass through a scaling factor for the residual after restriction, i.e. do something like this:
```
rscale = appctx.get("rscale", None)
if rscale is not None:
    pcmg.setMGRScale(1, rscale)
```
However, this does not appear to work since `setMGRScale()` is only used for **non-linear** multigrid.

**Additional info**

The following code contains an example which uses the workaround for passing the restriction as a PETSc matrix:

[hybridized_mixed_lumped.py.txt](https://github.com/user-attachments/files/20722131/hybridized_mixed_lumped.py.txt)

eikehmueller:
# Description

Add functionality to pass an optional restriction matrix to `GTMGPC`.

fixes #4367

At the moment, only the interpolation matrix of the `pcmg` object can be specified through the `appctx` via `"interpolation_matrix":interp_petscmat` and the restriction matrix is automatically set to the transpose of this. The changes in this PR allow passing the restriction matrix through the `appctx` via `"restriction_matrix":restr_petscmat`. If no restriction matrix is specified, it defaults to the transpose of the interpolation matrix as before, hence the change is backward compatible.

### DIFF
--- a/demos/parallel-printing/parprint.py.rst
+++ b/demos/parallel-printing/parprint.py.rst
@@ -15,8 +15,8 @@ Run this example in parallel using :math:`P` processes by doing
 
 We start with the usual import but we also import `petsc4py <https://bitbucket.org/petsc/petsc4py/>`_
 so that classes ``PETSc.X`` are available.  Here ``X`` is one of the
-`PETSc object types <https://www.mcs.anl.gov/petsc/documentation/index.html>`_,
-including types like `Vec <http://www.mcs.anl.gov/petsc/petsc-current/manualpages/Vec/index.html>`_::
+`PETSc object types <https://petsc.org/release/manualpages/>`_,
+including types like `Vec <https://petsc.org/release/manualpages/Vec/>`_::
 
     from firedrake import *
     from firedrake.petsc import PETSc
--- a/docs/notebooks/03-elasticity.ipynb
+++ b/docs/notebooks/03-elasticity.ipynb
@@ -264,7 +264,7 @@
    "source": [
     "# Solving bigger problems\n",
     "\n",
-    "Up to now, we've only really solved quite small problems, and therefore haven't really had to worry about tuning the solver. As we increase the size of the problem we're solving, the direct solver approach will no longer be good enough.  Firedrake uses [PETSc](http://www.mcs.anl.gov/petsc) to provide solvers, and uses PETSc solver parameters to control them.\n",
+    "Up to now, we've only really solved quite small problems, and therefore haven't really had to worry about tuning the solver. As we increase the size of the problem we're solving, the direct solver approach will no longer be good enough.  Firedrake uses [PETSc](http://petsc.org) to provide solvers, and uses PETSc solver parameters to control them.\n",
     "\n",
     "Let's dive straight in.  We'll write a function that solves the same elasticity problem, but takes parameters for the number of cells in the mesh, as well as a dictionary of solver options."
    ]
--- a/docs/notebooks/08-composable-solvers.ipynb
+++ b/docs/notebooks/08-composable-solvers.ipynb
@@ -346,7 +346,7 @@
    "source": [
     "### Block preconditioning\n",
     "\n",
-    "Firedrake hooks up all the necessary machinery to access PETSc's [`PCFIELDSPLIT`](https://www.mcs.anl.gov/petsc/petsc-current/manualpages/PC/PCFIELDSPLIT.html#PCFIELDSPLIT) preconditioner. This provides mechanisms for building preconditioners based on block factorisations. The Stokes problem \n",
+    "Firedrake hooks up all the necessary machinery to access PETSc's [`PCFIELDSPLIT`](https://petsc.org/release/manualpages/PC/PCFIELDSPLIT/) preconditioner. This provides mechanisms for building preconditioners based on block factorisations. The Stokes problem \n",
     "$$\n",
     "\\begin{align}\n",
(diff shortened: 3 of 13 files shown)


pbrubeck:
Could you checkout release, pull, and merge release into this branch so the unrelated commits go away?

eikehmueller:
just did
```
git checkout release
git pull
git checkout eikehmueller/gtmgpc-restriction
git merge release
git push
```
but that doesn't seem to change anything.

firedrake/preconditioners/gtmg.py
connorjward, line 1: Please add a docstring that at a minimum describes the options taken by the appctx.

eikehmueller: Good point. I added some documentation, let me know if you want me to change anything.

connorjward (approved):
That exceeded expectations! Thank you!

firedrake/preconditioners/gtmg.py
```diff
@@ -7,11 +18,54 @@
...
+            * get_coarse_op_transpose_nullspace: method which returns the
+                    null space of the transpose of the coarse operator
+            * interpolation_matrix: PETSc matrix which describes the
+                    interpolation from the coarse- to the fine space. If
+                    omitted, this will be constructed automatically with
+                    an Interpolater() object
+            * restriction_matrix: PETSc matrix which describes the
+                    restriction from the fine space dual to the coarse
+                    space dual
+        """
```
JHopeCollins: This is great, thanks @eikehmueller 

I think that this should all be in the top level docstring for the class though. Users will specify the pc in the options parameters and never calling `initialize` themselves, so the public API documentation should all be at the class level because that's where users interact with it.

JHopeCollins (changes requested):

firedrake/preconditioners/gtmg.py
```diff
@@ -1,3 +1,14 @@
+"""Non-nested multigrid preconditioner
+
+Implements the method described in [1]. Note that while the authors of this
+paper consider a non-nested function space hierarchy, the algorithm can also
+be applied if the spaces are nested.
+
+[1] Gopalakrishnan, J. and Tan, S., 2009: "A convergent multigrid cycle for the
+hybridized mixed method". Numerical Linear Algebra with Applications,
+16(9), pp.689-714. https://doi.org/10.1002/nla.636
+"""
```
JHopeCollins: And the same for this, we should have all the information about the PC in the same place rather than split between file, class, and method level.

firedrake/preconditioners/gtmg.py
eikehmueller: Just combined the comments and pushed to the branch. Is that what you had in mind @JHopeCollins?

JHopeCollins: Yes exactly, thanks!

JHopeCollins (dismissed):
LGTM

firedrake/preconditioners/gtmg.py
```diff
@@ -11,13 +14,60 @@
...
+    The following options (also passed through the `appctx`) are optional:
+
+        * `form_compiler_parameters`: parameters for assembling operators on
+          both levels of the hierarchy.
+        * `coarse_space_bcs`: boundary conditions to be used on coarse space.
+        * `get_coarse_op_nullspace`: method which returns the nullspace of the
+          coarse operator.
+        * `get_coarse_op_transpose_nullspace`: method which returns the
+          nullspace of the transpose of the coarse operator.
+        * `interpolation_matrix`: PETSc matrix which describes the interpolation
```
JHopeCollins:
```suggestion
        * `interpolation_matrix`: PETSc Mat which describes the interpolation
```

firedrake/preconditioners/gtmg.py
```diff
@@ -11,13 +14,60 @@
...
+          from the coarse to the fine space. If omitted, this will be
+          constructed automatically with an :class:`.Interpolate` object.
+        * `restriction_matrix`: PETSc matrix which describes the restriction
```
JHopeCollins:
```suggestion
        * `restriction_matrix`: PETSc Mat which describes the restriction
```

firedrake/preconditioners/gtmg.py
eikehmueller: Just pushed a commit which changes this

firedrake/preconditioners/gtmg.py
eikehmueller: Just pushed a commit which changes this

PR #4374, opened 2025-06-13
GTMGPC: use the new Interpolate

pbrubeck:
# Description
Use `assemble(Interpolate).petscmat` as opposed to the older `Interpolator.callable().handle`.

Also fix imports.

### DIFF
--- a/firedrake/preconditioners/gtmg.py
+++ b/firedrake/preconditioners/gtmg.py
@@ -16,6 +16,7 @@ class GTMGPC(PCBase):
     _prefix = "gt_"
 
     def initialize(self, pc):
+        from firedrake.assemble import assemble, get_assembler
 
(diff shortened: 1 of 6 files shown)


pbrubeck (commented):
This looks good, but I cannot approve as I added the test

firedrake/preconditioners/gtmg.py
```diff
@@ -12,11 +16,6 @@ class GTMGPC(PCBase):
     _prefix = "gt_"
 
     def initialize(self, pc):
-        from firedrake import TestFunction, parameters
-        from firedrake.assemble import get_assembler
```
connorjward, line 16: Please put the cyclic imports here. Otherwise this is totally fine.

PR #4375, opened 2025-06-13
DO NOT MERGE: Stub PR for build testing

tmbgreaves:
DO NOT MERGE

This is a stub PR to check the processes that result from a PR build, towards the goal of being able to detect, on the builder, whether there's an ongoing build. This is in response to an air conditioning failure in the room where the Firedrake builders are hosted, which is currently under severe thermal (over)load and needs systems to be throttled or powered down from time to time.


connorjward:
@tmbgreaves does this PR need to remain open?

Issue #4376, opened 2025-06-15
BUG: Incorrect solution to linear system at interior facets when Slate is used with Dirichlet BCs


astern:
**Describe the bug**
When solving a linear system with Dirichlet BCs, where the LHS is an assembled Slate `Tensor` and the RHS is a `Cofunction`, interior facet DOFs are handled incorrectly, leading to an incorrect solution. The issue does not arise when natural BCs are used.

In addition to the MWE below, I have observed this bug in previously-working hybrid FEM code: static condensation using Slate leads to a different (incorrect) solution compared to the (correct) solution of the uncondensed linear system, and the problem only arises when Dirichlet BCs are imposed.

**Steps to Reproduce**
1. Code/test/example that was run:
```
from firedrake import *

mesh = UnitSquareMesh(1, 1)
V = FunctionSpace(mesh, 'HDivT', 0)

u, v = TrialFunction(V), TestFunction(V)
a = avg(u*v)*dS + u*v*ds
L = avg(v)*dS + v*ds

bcs = [DirichletBC(V, 0, 'on_boundary')]
A = assemble(Tensor(a), bcs=bcs)
b = assemble(Tensor(L), bcs=bcs) # omitting Tensor( ) or bcs=bcs does not affect issue

u = Function(V)
solve(A, u, b)
```

2. Command executed: `u.dat.data`
3. Output: `array([0., 0., 2., 0., 0.])`

**Expected behavior**
Output should be `array([0., 0., 1., 0., 0.])`

**Environment:**
 - OS: MacOS
 - Python version: 3.13.4
 - Output of `pip list`
```
Package                   Version
------------------------- --------------
anyio                     4.9.0
appnope                   0.1.4
argon2-cffi               25.1.0
argon2-cffi-bindings      21.2.0
arrow                     1.3.0
asttokens                 3.0.0
async-lru                 2.0.5
attrs                     25.3.0
babel                     2.17.0
beautifulsoup4            4.13.4
bleach                    6.2.0
cachetools                6.0.0
certifi                   2025.4.26
cffi                      1.17.1
cgen                      2020.1
charset-normalizer        3.4.2
checkpoint_schedules      1.0.4
codepy                    2025.1
colorama                  0.4.6
comm                      0.2.2
constantdict              2025.2
debugpy                   1.8.14
decorator                 4.4.2
defusedxml                0.7.1
executing                 2.2.0
fastjsonschema            2.21.1
fenics-ufl                2025.1.0
[...]

PR #4377, opened 2025-06-15
SLATE: prevent Cofunction reassembly

pbrubeck:
# Description
Here we avoid implicit casting of Cofunctions into SLATE expressions when forming the residual `F=a*u-L` of a `LinearVariationalProblem`, as SLATE will always sum the cell contributions into the output vector. Instead, we deal with linear combinations of SLATE expressions and `Cofunctions` via `FormSum`, so the Cofunction gets excluded from the sum over elements.

We also extend BaseForm assembly to BaseForm objects with SLATE nodes.

Fixes https://github.com/firedrakeproject/firedrake/issues/4376

### DIFF
--- a/firedrake/assemble.py
+++ b/firedrake/assemble.py
@@ -60,17 +60,23 @@ def assemble(expr, *args, **kwargs):
         `ufl.classes.Measure` in the form. For example, if a
         ``quadrature_degree`` of 4 is specified in this argument, but a degree of
         3 is requested in the measure, the latter will be used.
-    mat_type : str
+    mat_type : str | None
         String indicating how a 2-form (matrix) should be
         assembled -- either as a monolithic matrix (``"aij"`` or ``"baij"``),
-        a block matrix (``"nest"``), or left as a `matrix.ImplicitMatrix` giving
+        a block matrix (``"nest"``), or left as a :class:`firedrake.matrix.ImplicitMatrix` giving
         matrix-free actions (``'matfree'``). If not supplied, the default value in
-        ``parameters["default_matrix_type"]`` is used.  BAIJ differs
-        from AIJ in that only the block sparsity rather than the dof
+        ``parameters["default_matrix_type"]`` is used.  ``"baij"``` differs
+        from ``"aij"`` in that only the block sparsity rather than the DoF
         sparsity is constructed.  This can result in some memory
         savings, but does not work with all PETSc preconditioners.
-        BAIJ matrices only make sense for non-mixed matrices.
-    sub_mat_type : str
+        ``"baij"`` matrices only make sense for non-mixed matrices with arguments
+        on a :func:`firedrake.functionspace.VectorFunctionSpace`.
+
+        NOTE
+        ----
+        For the assembly of a 0-form or 1-form arising from the action of a 2-form,
+        the default matrix type is ``"matfree"``.
+    sub_mat_type : str | None
         String indicating the matrix type to
         use *inside* a nested block matrix.  Only makes sense if
         ``mat_type`` is ``nest``.  May be one of ``"aij"`` or ``"baij"``.  If
@@ -154,10 +160,10 @@ def get_assembler(form, *args, **kwargs):
     is_base_form_preprocessed = kwargs.pop('is_base_form_preprocessed', False)
     fc_params = kwargs.get('form_compiler_parameters', None)
     if isinstance(form, ufl.form.BaseForm) and not is_base_form_preprocessed:
-        mat_type = kwargs.get('mat_type', None)
-        # If not assembling a matrix, internal BaseForm nodes should be matfree by default
-        if mat_type is None and len(form.arguments()) < 2:
-            mat_type = "matfree"
+        # If not assembling a matrix, internal BaseForm nodes are matfree by default
+        # Otherwise, the default matrix type is firedrake.parameters["default_matrix_type"]
+        default_mat_type = "matfree" if len(form.arguments()) < 2 else None
+        mat_type = kwargs.get('mat_type', default_mat_type)
         # Preprocess the DAG and restructure the DAG
         # Only pre-process `form` once beforehand to avoid pre-processing for each assembly call
         form = BaseFormAssembler.preprocess_base_form(form, mat_type=mat_type, form_compiler_parameters=fc_params)


firedrake/assemble.py
```diff
@@ -838,7 +838,11 @@ def preprocess_base_form(expr, mat_type=None, form_compiler_parameters=None):
         if mat_type != "matfree":
             # Don't expand derivatives if `mat_type` is 'matfree'
             # For "matfree", Form evaluation is delayed
-            expr = BaseFormAssembler.expand_derivatives_form(expr, form_compiler_parameters)
+            try:
+                expr = BaseFormAssembler.expand_derivatives_form(expr, form_compiler_parameters)
+            except ValueError:
```
connorjward: I know we're not very good at this, but I think much better practice here would be to have something like:
```py
try:
  expr = BaseFormAssembler.expand_derivatives_form(...)
except UnrecognisedFormTerminal:
  pass
```
as any manner of things can throw a `ValueError` so we might be masking other failure modes.

firedrake/variational_solver.py
```diff
@@ -132,14 +132,21 @@ def dm(self):
         return self.u_restrict.function_space().dm
 
     @staticmethod
-    def compute_bc_lifting(J, u):
+    def compute_bc_lifting(J, u, L=0):
```
connorjward: Please amend the docstring

connorjward: and is this function called elsewhere? Maybe `L` needn't be optional.

firedrake/assemble.py
pbrubeck: The error arises from UFL. How efficient are we at getting things merged in UFL now that we gave up our fork?

pbrubeck: I was thinking that we could inspect the DAG directly on this script and figure out whether there are slate Tensors within the BaseForm

connorjward: This is sort of a style choice so convincing FEniCS to change might be a bit of a faff. Checking the DAG sounds good though it could be slightly inefficient.

connorjward: Maybe you could check `form.coefficients()` or similar so you don't need to do another proper traversal.

pbrubeck: We are not interested in assembling a matrix here as the residual is a linear form. So maybe the right thing would be 
```
        if len(expr.arguments()) == 2 and mat_type != "matfree":
           # Don't expand derivatives if `mat_type` is 'matfree'	
           # For 1-forms or "matfree", Form evaluation is delayed
```

firedrake/variational_solver.py
pbrubeck: We are calling it in `solving_utils.py` without `L`.

firedrake/assemble.py
connorjward: Yeah seems reasonable.

firedrake/assemble.py
```diff
@@ -872,7 +872,11 @@ def expand_derivatives_form(form, fc_params):
         # be `ufl.BaseForm`, or even an appropriate `ufl.Expr`, since assembly of expressions
         # containing derivatives is not supported anymore but might be needed if the expression
         # in question is within a `ufl.BaseForm` object.
-        return ufl.algorithms.ad.expand_derivatives(form)
+        try:
```
pbrubeck: I think trying to differentiate slate and getting a ValueError is the best we can do for the moment.

I also do not understand why this early expansion of derivatives is required at all. I think the right fix is to lazily evaluate the Derivative nodes (if they appear).

An alternative is to first check if there are any derivative nodes via `base_form_postorder_traversal` and only call `expand_derivatives` if we found any.

pbrubeck: I just implemented the alternative approach.

pbrubeck: One might argue that this extra `postorder_traversal` is not making things worse when there are no derivatives. `expand_derivatives` adds to the cost by calling `apply_algebra_lowering`, which is generally undesired as it makes the DAG more expensive.

But if there are derivatives indeed, the traversal will terminate quite early. We expect the derivatives to be the outermost operators in the `BaseForm`.

firedrake/assemble.py
```diff
@@ -872,7 +872,24 @@ def expand_derivatives_form(form, fc_params):
...
+
+        def detect_derivatives(e, *operands):
+            if isinstance(e, slate.TensorBase):
+                raise TypeError("SLATE Tensors are not differentiable.")
+            elif isinstance(e, ufl.differentiation.Derivative):
+                raise AssertionError("We should expand derivatives.")
+        try:
+            visited = {}
+            BaseFormAssembler.base_form_postorder_traversal(form, detect_derivatives, visited)
+        except AssertionError:
```
connorjward: I assume that these exceptions are coming from lines 878 and 880 above? If so we should implement custom exception types because anything can throw `AssertionError`s. Something like

```py
class FiredrakeException(Exception, metaclass=abc.ABCMeta):
    """Base class for Firedrake exceptions."""


class CannotDifferentiateSlateException(FiredrakeException):
    ...


class UnexpectedDerivativeNodeFound(FiredrakeException):
    ...
```

firedrake/variational_solver.py
```diff
@@ -131,14 +131,36 @@ def dm(self):
...
 
     @staticmethod
-    def compute_bc_lifting(J, u):
-        """Return the action of the bilinear form J (without bcs) on a Function u."""
+    def compute_bc_lifting(J, u, L=0):
+        """Compute the residual after lifting DirichletBCs.
+
+        Parameters
+        ----------
+        J : ufl.BaseForm or slate.TensorBase
```
connorjward: If you're going to add a docstring in the new style please use type hints.

firedrake/variational_solver.py
```diff
@@ -131,14 +131,36 @@ def dm(self):
...
         if isinstance(J, MatrixBase) and J.has_bcs:
             # Extract the full form without bcs
             if not isinstance(J.a, (ufl.BaseForm, slate.slate.TensorBase)):
                 raise TypeError(f"Could not remove bcs from {type(J).__name__}.")
             J = J.a
-        return ufl_expr.action(J, u)
+        F = ufl_expr.action(J, u)
+        if isinstance(F, slate.slate.TensorBase) and not isinstance(L, (ufl.Form, slate.slate.TensorBase)):
+            # Slate expressions should not combine with assembled Cofunctions
+            # because assemble(AssembledVector(L)) repeats element summation on L
```
connorjward, line 163: This seems like a bug. Does an issue exist for this?

pbrubeck: Yes, it is linked to the PR

firedrake/variational_solver.py
pbrubeck: A similar conversation was resolved in June. This should also be acceptable.

firedrake/assemble.py
pbrubeck: Where should they live?

pbrubeck: I think the Derivative nodes aren't necessarily unexpected. `DerivativeNodeFound` could be a better name.

pbrubeck: This seems too anoying, I'll implement a proper expand_derivatives visitor, that is only called on nodes that contain derivatives. It should fail with Derivative(slate).

firedrake/variational_solver.py
connorjward: Following our discussions in Firedrake meetings I think that that is no longer valid. This should be made to conform with our latest style guide.

firedrake/variational_solver.py
connorjward: I don't mean https://github.com/firedrakeproject/firedrake/issues/4376. Is this not a deeper bug within SLATE that this just patches over?

pbrubeck: I think the Cofunction reassembly is the expected thing from the elementwise model of SLATE. Maybe we can divide `AssembledVector` by the DOF multiplicity and then we would not have this issue.

pbrubeck: The advantage of a SLATE-only approach is that we will not hit the `expand_derivatives` issues with BaseFormAssembler

connorjward: I don't know this bit of the stack well enough to comment. Up to you.

pbrubeck: I think we would like both codepaths to work (Slate-only and BaseForm(Slate)). I'll keep this PR to enable the latter.

firedrake/variational_solver.py
pbrubeck: How do I resolve circular imports?

connorjward: `from __future__ import annotations`

pbrubeck: What if I need to import firedrake on this file?

pbrubeck: Should it be `Function` (which we already import) or `firedrake.function.Function`?

connorjward: For now just use a string instead. E.g. `"firedrake.Function"`.

connorjward: I think you can just use `Function`.

pbrubeck: `slate.TensorBase` does not link to the relevant section of the manual.

firedrake/assemble.py
pbrubeck: It turns out that we can avoid this `expand_derivatives` nonsense by setting the `mat_type="matfree"` for the residual assembly.

firedrake/assemble.py
```diff
@@ -154,7 +160,10 @@ def get_assembler(form, *args, **kwargs):
     is_base_form_preprocessed = kwargs.pop('is_base_form_preprocessed', False)
     fc_params = kwargs.get('form_compiler_parameters', None)
     if isinstance(form, ufl.form.BaseForm) and not is_base_form_preprocessed:
-        mat_type = kwargs.get('mat_type', None)
+        # If not assembling a matrix, internal BaseForm nodes are matfree by default
+        # Otherwise, the default matrix type is firedrake.parameters["default_matrix_type"]
+        default_mat_type = "matfree" if len(form.arguments()) < 2 else None
+        mat_type = kwargs.get('mat_type', default_mat_type)
```
pbrubeck, line 166: The `expand_derivatives` call is only for `mat_type != "matfree"`. This change is to be consistent between the internal numerics of `action(Form, Function)` and `action(BaseForm, Function)`.

firedrake/assemble.py
```diff
@@ -60,17 +60,23 @@ def assemble(expr, *args, **kwargs):
...
+        ``parameters["default_matrix_type"]`` is used.  ``"baij"``` differs
+        from ``"aij"`` in that only the block sparsity rather than the DoF
         sparsity is constructed.  This can result in some memory
         savings, but does not work with all PETSc preconditioners.
-        BAIJ matrices only make sense for non-mixed matrices.
-    sub_mat_type : str
+        ``"baij"`` matrices only make sense for non-mixed matrices with arguments
+        on a :func:`firedrake.functionspace.VectorFunctionSpace`.
+
+        NOTE
```
connorjward, line 75: Does this render correctly?

pbrubeck: Yes, I checked.

pbrubeck: image

connorjward: Nice

PR #4378, opened 2025-06-16
Fix assemble(slate.Tensor, diagonal=True)

pbrubeck:
# Description


firedrake/assemble.py
```diff
@@ -1745,7 +1745,7 @@ def _as_global_kernel_arg_output(_, self):
     if rank == 0:
         return op2.GlobalKernelArg((1,))
     elif rank == 1 or rank == 2 and self._diagonal:
-        V, = Vs
+        V = Vs[0]
```
connorjward, line 1748: We have `V,` here to assert a singleton. Why is that now violated? What are the contents of `V[1:]`?

pbrubeck: With slate we get the two copies of the same space in Vs. For some reason `diagonal=True` will cause `len(indices) == 1` only for `ufl.Form`, but not for `slate.Tensor`. On both cases `len(form.arguments()) == 2`.

https://github.com/firedrakeproject/firedrake/blob/32611a79ed56978c19ca004a1633addf31724aa4/firedrake/assemble.py#L2214

pbrubeck: And this is where we get the `indices` from
https://github.com/firedrakeproject/firedrake/blob/32611a79ed56978c19ca004a1633addf31724aa4/firedrake/assemble.py#L1653

Which means that we should trim the lenght of indices somewhere in slate/slac/compiler.py, but it gets hard to track down because `diagonal` is not mentioned at all in that script.

connorjward: Could you use `V = pytools.single_valued(Vs)` then?

pbrubeck:
As I was adding tests, I saw that there is support for `assemble(Tensor(form, diagonal=True))`, which assembles directly the diagonal. But this PR aims to enable `assemble(Tensor(form), diagonal=True)` to assemble the diagonal from the Matrix entries. Unfortunately the latter gives incorrect results. 

In general we cannot rely on `Tensor(form, diagonal=True)`, as we would like to support diagonals of more complicate SLATE expressions (e.g. Schur complements). So this PR would require more work.

PR #4381, opened 2025-06-17
Import `get_external_packages` from `petsctools` not `firedrake.petsc`

JHopeCollins:
Avoid raising our own warnings in our own tests.


connorjward (approved):
Whoops, missed that

connorjward:
Only failing an unrelated link check

PR #4382, opened 2025-06-18
Debug petsc_raises

connorjward:
This passes locally for me. Unsure why it reportedly hangs.


leo-collins:
It was also passing for me (on macOS). [Here is the CI run that failed](https://github.com/firedrakeproject/firedrake/actions/runs/15322130907).

PR #4383, opened 2025-06-20
Speed up caching

### LINKED ISSUE #3983: BUG: Caching implementation is slow
The [current implementation](https://github.com/firedrakeproject/firedrake/blob/master/pyop2/caching.py#L508) of `parallel_cache` is quite slow.

As an example, running `test/tracerEq/test_consistency_2d.py::test_nonconst_tracer[DIRK33]"` in Thetis (with the changes from #3982) and a hot cache produces a flamegraph like:

Image

All the parts showing `PyOP2 Cache...` are where we are accessing the cache.

I think that rewriting this in C/Cython would be quite impactful for problems where we call a lot of very small parloops.

connorjward:
This is a bit of a refactor of `parallel_cache` to try and address the performance issues reported in https://github.com/firedrakeproject/firedrake/issues/3983. 

Fixes #3983 

## TODOs
- [x] Verify that this is parallel safe (esp. check with G-ADOPT)

The main changes are:
* API changes `comm_getter` to `get_comm` and `cache_factory` to `make_cache`.
* Do not use `as_hexdigest` every time we do a cache lookup. I think this was the main cause of the slowdown.
* Identify caches by a local ID attribute not `type(make_cache()).__class__.__name__` which was unsafe and slow (?).

Using the example from #3983 the performance improvement takes us (with a hot cache) from 31s to 14s, so I must be doing something right.


pyop2/compilation.py
```diff
@@ -248,6 +249,18 @@ def __repr__(self):
...
+        return hash((
+            type(self),
+            self._extra_compiler_flags,
+            self._extra_linker_flags,
+            self._version,
+            self._debug,
+        ))
+
+    def __eq__(self, other) -> bool:
+        return type(other) is type(self) and hash(other) == hash(self)
```
ksagiyam: There could be hash collisions, so we should just compare item by item.

connorjward: Fair enough. I was lazy and didn't want to write them all out for `__hash__` and `__eq__`.

connorjward:
I have run a big G-ADOPT example on 16 cores and it ran without issue. I think that this is safe to merge.

connorjward:
> Should I consider re-running my performance experiments after this PR is merged?

That might potentially be a good idea. Were your experiments done against release or master? (This is going into master)

connorjward:
> > > Should I consider re-running my performance experiments after this PR is merged?
> > 
> > That might potentially be a good idea. Were your experiments done against release or master? (This is going into master)
> 
> Release 

I guess this doesn't actually matter. You might just need to reinstall.

The changes here only really make an impact if you are running with small problems at the strong scaling limit. Does that apply?

PR #4384, opened 2025-06-20
'Fix' petsc raises

connorjward:
Since https://gitlab.com/petsc/petsc/-/commit/4237731afe36b4db0a3147aff0aed51d4657afa2 and the introduction of `petsc_raises` (https://github.com/firedrakeproject/firedrake/commit/c9d392201d03f1736258a9d8f915e909ca215339) we have been seeing random failures on CI that I have been unable to reproduce locally.

I was able to reproduce the stochastic failures with a Docker image (Python 3.12.3) which appear to go away when I install Python 3.12.11 (matches my local machine). My best guess is therefore that Python fixed things between these releases.

The best approach to me seems to be to allow the more permissive exception check until we reach a Python version that we think should be fine. We can revisit then if the issue continues.

The only other clue I have is we get the following reported error when things fail:

```
/usr/local/lib/python3.12/dist-packages/_pytest/unraisableexception.py:65: PytestUnraisableExceptionWarning: Exception ignored in tp_clear of: <class 'dict'>

    File "petsc4py/PETSc/libpetsc4py.pyx", line 1004, in petsc4py.PETSc.MatMultTranspose_Python
    File "/opt/firedrake/firedrake/interpolation.py", line 1684, in multTranspose
      raise NotImplementedError(
  NotImplementedError: Can only apply adjoint to expressions consisting of a single argument at the moment.

  Enable tracemalloc to get traceback where the object was allocated.
  See https://docs.pytest.org/en/stable/how-to/capture-warnings.html#resource-warnings for more info.
    warnings.warn(pytest.PytestUnraisableExceptionWarning(msg))
```
The `Exception ignored in tp_clear of: <class 'dict'>` makes it seem like a GC issue.

### DIFF
--- a/tests/firedrake/conftest.py
+++ b/tests/firedrake/conftest.py
@@ -204,9 +204,8 @@ class _petsc_raises:
         pass
 
     def __exit__(self, exc_type, exc_val, traceback):
-        # There appears to be a bug that is fixed between Python 3.12.3 (latest
-        # Docker image) and 3.12.11 (my local machine) where 'exc_val' is
-        # randomly sometimes 'None'.
+        # There is a bug where 'exc_val' is occasionally 'None'. In my tests
+        # this error only exists for Python < 3.12.11.
         if exc_type is PETSc.Error:
             if sys.version_info < (3, 12, 11):
                 if exc_val is None or isinstance(exc_val.__cause__, self.exc_type):
--- a/tests/firedrake/conftest.py
+++ b/tests/firedrake/conftest.py
@@ -1,6 +1,7 @@
 """Global test configuration."""
 
 import os
+import sys
 
 # Disable warnings for missing options when running with pytest as PETSc does
 # not know what to do with the pytest arguments.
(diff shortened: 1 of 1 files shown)


tests/firedrake/conftest.py
```diff
@@ -203,8 +204,15 @@ def __enter__(self):
         pass
 
     def __exit__(self, exc_type, exc_val, traceback):
-        if exc_type is PETSc.Error and isinstance(exc_val.__cause__, self.exc_type):
-            return True
+        # There is a bug where 'exc_val' is occasionally 'None'. In my tests
+        # this error only exists for Python < 3.12.11.
```
JHopeCollins, line 208: Please can you open an issue to remove the special casing once we stop supporting Python 3.12.11?
Possibly also add a TODO to the comment for those that like grepping!

connorjward: Sure. I will open an issue immediately prior to merging this. I'll leave this conversation unresolved until then.

connorjward: https://github.com/firedrakeproject/firedrake/issues/4406

PR #4386, opened 2025-06-20
Irksome smoke tests

pbrubeck:
# Description


.github/workflows/core.yml
```diff
@@ -248,6 +248,19 @@ jobs:
         timeout-minutes: 15
 
 
+      - name: Run Irksome smoke tests
```
indiamai, line 251: Is there a particular reason you've put these before the standard tests, rather than with the other smoke tests?

connorjward: It does help with debugging as you don't need to wait for the whole test suite to run to find out if you messed up.

PR #4391, opened 2025-06-24
Do not use Makefile for firedrake-check

connorjward:
* The previous approach led to issues for a number of users for whom 'mpiexec' did not work inside the Makefile.
* We now support a '--mpiexec' argument for those requiring extra commands like '--oversubscribe' or 'srun'.
* Adapts to mpi-pytest change so parallel[1] covers all serial tests.

**This is approved an can be merged once a new release of mpi-pytest is made.** (now done)

### DIFF
--- a/firedrake/_check/__init__.py
+++ b/firedrake/_check/__init__.py
@@ -1,9 +1,0 @@
-"""Run the Firedrake smoke tests."""
-
-import pathlib
-import subprocess
-
-
-def main() -> None:
-    dir = pathlib.Path(__file__).parent
-    subprocess.run(f"make -C {dir} check".split(), errors=True)
--- a/Makefile
+++ b/Makefile
@@ -88,33 +88,6 @@ clean:
 	@echo "    RM tinyasm/*.so"
 	-@rm -f tinyasm/*.so > /dev/null 2>&1
 
-# Do verbose checking if running on CI and always set no:cacheprovider because
-# we don't want to generate any cache files in $VIRTUAL_ENV/lib/.../firedrake/_check
-check_flags =
-ifeq ($(FIREDRAKE_CI), 1)
-	check_flags = --verbose -p no:cacheprovider
-else
-	check_flags = --quiet -p no:cacheprovider
-endif
-
-CHECK_PYTEST_ARGS =
-
-.PHONY: check
-check:
-	@echo "    Running serial smoke tests"
-	@python3 -m pytest $(check_flags) $(CHECK_PYTEST_ARGS) \
-		tests/firedrake/regression/test_stokes_mini.py::test_stokes_mini \
-		tests/firedrake/regression/test_locate_cell.py  `# spatialindex` \
-		tests/firedrake/supermesh/test_assemble_mixed_mass_matrix.py::test_assemble_mixed_mass_matrix[2-CG-CG-0-0]  `# supermesh` \
-		tests/firedrake/regression/test_matrix_free.py::test_fieldsplitting[parameters3-cofunc_rhs-variational]  `# fieldsplit` \
-		tests/firedrake/regression/test_nullspace.py::test_near_nullspace  `# near nullspace`
-	@echo "    Serial tests passed"
-	@echo "    Running parallel smoke tests"
-	@mpiexec -n 3 python3 -m pytest $(check_flags) $(CHECK_PYTEST_ARGS) -m parallel[3] \
-		tests/firedrake/regression/test_dg_advection.py::test_dg_advection_icosahedral_sphere \
-		tests/firedrake/regression/test_interpolate_cross_mesh.py::test_interpolate_cross_mesh_parallel[extrudedcube]  `# vertex-only mesh`
-	@echo "    Parallel tests passed"
-
 .PHONY: durations
 durations:
 	@echo "    Generate timings to optimise pytest-split"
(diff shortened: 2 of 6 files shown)


firedrake/_check/__init__.py
```diff
@@ -1,9 +1,70 @@
...
+    run_tests(PARALLEL3_TESTS, 3, args.mpiexec)
+    logger.info("    Parallel tests passed")
+
+
+def parse_args() -> argparse.Namespace:
+    parser = argparse.ArgumentParser()
+    parser.add_argument(
+        "--mpiexec",
+        type=str,
+        default="mpiexec -n",
```
connorjward: @JHopeCollins you have been asking for this. I'm not sure what the best way to do this is. Can we always just have
```
MPIEXEC_COMMAND <nprocs> python -m pytest ...
```
or are there sometimes occasions where we need something more complicated?

JHopeCollins: I think so, as long as the final flag in `MPIEXEC_COMMAND` is `-n` or the equivalent. I've added a suggestion to the help message to clarify this.

firedrake/_check/__init__.py
```diff
@@ -1,9 +1,70 @@
...
+        help="Command used to launch MPI."
```
JHopeCollins:
```suggestion
        help="Command used to launch MPI. Must end with the flag taking the number of processors, e.g. '-n' for mpiexec."
```

firedrake/_check/__init__.py
```diff
@@ -1,9 +1,70 @@
...
+    "tests/firedrake/regression/test_matrix_free.py::test_fieldsplitting[parameters3-cofunc_rhs-variational]",
+    # near nullspace
+    "tests/firedrake/regression/test_nullspace.py::test_near_nullspace",
+)
+
+PARALLEL3_TESTS = (
+    "tests/firedrake/regression/test_dg_advection.py::test_dg_advection_icosahedral_sphere[nprocs=3]",
+    # vertex-only mesh
+    "tests/firedrake/regression/test_interpolate_cross_mesh.py::test_interpolate_cross_mesh_parallel[extrudedcube]",
+)
```
JHopeCollins:
```suggestion
# smoke tests grouped by number of processors
TESTS = {                                                                                                                       
    1: (
        "tests/firedrake/regression/test_stokes_mini.py::test_stokes_mini",
        # spatialindex
        "tests/firedrake/regression/test_locate_cell.py",
        # supermesh
        "tests/firedrake/supermesh/test_assemble_mixed_mass_matrix.py::test_assemble_mixed_mass_matrix[2-CG-CG-0-0]",
        # fieldsplit
        "tests/firedrake/regression/test_matrix_free.py::test_fieldsplitting[parameters3-cofunc_rhs-variational]",
        # near nullspace
        "tests/firedrake/regression/test_nullspace.py::test_near_nullspace",
    ),
    3: (
        "tests/firedrake/regression/test_dg_advection.py::test_dg_advection_icosahedral_sphere[nprocs=3]",
        # vertex-only mesh
        "tests/firedrake/regression/test_interpolate_cross_mesh.py::test_interpolate_cross_mesh_parallel[extrudedcube]",
    ),
}
```

firedrake/_check/__init__.py
```diff
@@ -1,9 +1,70 @@
...
+
 def main() -> None:
+    args = parse_args()
+
+    logger.info("    Running serial smoke tests")
+    run_tests(SERIAL_TESTS, 1, args.mpiexec)
+    logger.info("    Serial tests passed")
+
+    logger.info("    Running parallel smoke tests")
+    run_tests(PARALLEL3_TESTS, 3, args.mpiexec)
```
JHopeCollins:
```suggestion
    for nprocs, tests in TESTS:
        if nprocs == 1:
            continue
        logger.info(f"        Running parallel[{nprocs}] smoke tests")
        run_tests(tests, nprocs, args.mpiexec)
```

connorjward:
@JHopeCollins I have followed your suggestions though the code doesn't exactly match your suggested changes.

firedrake/_check/__init__.py
```diff
@@ -34,13 +36,13 @@
...
-    logger.info("    Serial tests passed")
-
-    logger.info("    Running parallel smoke tests")
-    run_tests(PARALLEL3_TESTS, 3, args.mpiexec)
-    logger.info("    Parallel tests passed")
+    for nprocs, tests in TESTS.items():
+        if nprocs == 1:
+            logger.info("    Running serial smoke tests")
+        else:
+            logger.info(f"    Running parallel smoke tests (nprocs={nprocs})")
```
JHopeCollins: Yeah, better than my original suggestion. I realised this actually was the right way to do it later on.

.github/workflows/core.yml
```diff
@@ -207,6 +207,10 @@ jobs:
             --extra-index-url https://download.pytorch.org/whl/cpu \
             "$(echo ./firedrake-repo/dist/firedrake-*.tar.gz)[ci,docs]"
 
+          : # DO NOT MERGE
+          pip uninstall -y mpi-pytest
+          pip install "mpi-pytest @ git+https://github.com/firedrakeproject/mpi-pytest.git@connorjward/always-mark"
```
JHopeCollins: I've approved this PR now, you can resolve this and merge once that's merged.

connorjward: We use pytest-mpi releases so I will have to make a new release before this goes in. I can do that in my own time though. Thanks for the review!

connorjward:
This is failing because the non-forking mpiexec problem is alive and kicking! I have a fix in this instance but it's interesting that this works fine on my machine.

PR #4399, opened 2025-06-24
Eagerly import petscp4y in setup.py (but don't initialise)

connorjward:
This leads to a much clearer error message than "petsctools has no attribute init" which happens if petsc4py isn't available. This error can happen if a user forgets to pass '--no-build-isolation' for a developer build.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -10,6 +10,7 @@ from pathlib import Path
 import libsupermesh
 import numpy as np
 import pybind11
+import petsc4py
 import petsctools
 import rtree
 from Cython.Build import cythonize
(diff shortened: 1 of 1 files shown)


JHopeCollins:
Couldn't you add a dummy `petsctools.init` in the case that petsc4py isn't available [here](https://github.com/firedrakeproject/petsctools/blob/4317aaa705fd700509e6b3ffd96a668651d60108/petsctools/__init__.py#L15) that raises a missing module error with a useful error message? Then other packages would get the benefit too, not just Firedrake.

connorjward:
> Couldn't you add a dummy `petsctools.init` in the case that petsc4py isn't available [here](https://github.com/firedrakeproject/petsctools/blob/4317aaa705fd700509e6b3ffd96a668651d60108/petsctools/__init__.py#L15) that raises a missing module error with a useful error message? Then other packages would get the benefit too, not just Firedrake.

Yeah that's actually a much better idea. I'll close this.

PR #4400, opened 2025-06-24
More flexible Dockerfile, and restore pushing docker containers for master

JHopeCollins:
**CLOSED**: Replaced by #4533 

There seems to be some appetite for people building their own Firedrake Docker containers. The current Dockerfiles are very static in what they can do - apart from allowing different ARCH options. This PR makes the Dockerfile slightly more flexible by allowing passing a particular branch to install from:

- If installing from `release` then the behaviour is unchanged - Firedrake and PETSc are both built from the corresponding release branches.
- If installing from any other branch, then `firedrake-configure` is downloaded from the specified branch, PETSc is installed from the latest `main` branch, and Firedrake is installed from the latest version of the specified branch.

TODO:
Some downstream repositories want to test against `master` rather than `release`, or test against both `master` and `release`. However, we no longer build a Docker container on pushes to master. Being able to specify a non-release branch to the Dockerfile means we can use the same Dockerfile to build containers for both release and master.


docker/Dockerfile.vanilla
```diff
@@ -1,9 +1,33 @@
 # Dockerfile for a plain Firedrake suitable for testing Firedrake components and applications
+# Firedrake is installed as a system package so doesn't require activating a venv when the container is launched
+#
+# There are two command line arguments:
+# ARCH:
+#     The `--arch` argument to firedrake-configure, e.g. default, complex
+# BRANCH:
+#     The Firedrake branch to install from. The `firedrake-configure` script from this branch will be used.
+#     - If BRANCH == "release": installs the latest firedrake release and the corresponding PETSc release.
+#     - If BRANCH != "release": installs that Firedrake branch and the latest PETSc master branch.
```
connorjward: This introduces some strange implicit behaviour. I feel like it might be better if the Firedrake and PETSc branches were specified separately (also it's PETSc `main`, not `master`).

Similarly it might make sense to have `DEV={0,1}` to distinguish whether we do a release- or master-style build.

JHopeCollins: Changed to a single argument `TARGET` that can either be:

1. `release` for the latest release branch build.
2. `<tag>` e.g. `2025.4.2` for a specific release.
3. `master` for the latest master branch build.

.github/workflows/pr.yml
```diff
@@ -13,3 +13,12 @@ jobs:
...
     secrets: inherit
+
+  docker:
+    name: Build dev Docker containers
+    uses: ./.github/workflows/docker.yml
+    with:
+      tag: dev-main
+      branch: main
+      build_dev: true
+    secrets: inherit
```
JHopeCollins, line 24: REMOVE BEFORE MERGE
```suggestion
```

.github/workflows/core.yml
```diff
@@ -108,6 +108,18 @@ jobs:
...
+      # Check that the Dockerfile is using the latest Ubuntu version.
+      # The version is hardcoded into the Dockerfile so that the OS
+      # for each release is fixed.
+      - name: Check Dockerfile Ubuntu version
+        run: |
+          latest_version=$(grep "VERSION_ID=" /etc/os-release | cut -d '"' -f 2)
+          docker_version=$(grep FROM docker/Dockerfile.vanilla | cut -d ':' -f 2)
+          if [[ "$docker_version" != "$latest_version" ]]; then
+            echo "Ubuntu version ${docker_version} in Dockerfile is out of date with latest version ${latest_version}"
+            false
```
connorjward: `exit 1`?

.github/workflows/docker.yml
```diff
@@ -25,18 +41,26 @@ jobs:
       matrix:
         os: [Linux, macOS]
         arch: [default, complex]
-        include:
+        platform: [linux/amd64, linux/arm64]
+        # Cannot use inputs in exclude clauses, so add a dummy matrix dim
+        is_dev_build:
+          - ${{ inputs.build_dev }}
```
connorjward, line 47: Job for me: think if there's a nicer way of doing this

.github/workflows/docker_merge.yml
```diff
@@ -14,6 +14,10 @@ on:
         description: Docker image tag
         required: true
         type: string
+      tag_latest:
+        description: Whether to add additional latest tag to containers
```
connorjward:
```suggestion
        description: Whether to add additional 'latest' tag to containers
```

.github/workflows/docker_merge.yml
```diff
@@ -50,10 +54,16 @@ jobs:
       - name: Merge and push the per-platform images
         working-directory: ${{ runner.temp }}/digests
         run: |
-          docker buildx imagetools create \
-            -t firedrakeproject/${{ inputs.target }}:${{ inputs.tag }} \
-            -t firedrakeproject/${{ inputs.target }}:latest \
-            $(printf 'firedrakeproject/${{ inputs.target }}@sha256:%s ' *)
+          if [[ "${{ inputs.tag_latest }}" == "true" ]]; then
```
connorjward, line 57: Is this line tested? Seems odd to test a boolean value using a string comparison.

pyproject.toml
```diff
@@ -110,7 +110,8 @@ ci = [
   "mpi-pytest>=2025.7",
   "nbval",
   "networkx",
-  "ngsPETSc",
+  # TODO RELEASE
+  # "ngsPETSc",
```
connorjward: Why is this here?

connorjward (commented):
I need to review the workflow file in more detail but looks pretty good.

Shouldn't this be going into `release` and not `main`? Otherwise we won't be building the `dev-release` containers.

.github/workflows/docker_merge.yml
JHopeCollins: github actions think that `inputs.tag_latest` is a bool, but bash thinks it is a string.

pyproject.toml
JHopeCollins: Good question, I don't know. Will remove.

.github/workflows/core.yml
JHopeCollins:
```suggestion
            exit 1
```

pyproject.toml
JHopeCollins:
```suggestion
  "ngsPETSc",
```

JHopeCollins: Done.

.github/workflows/docker.yml
```diff
@@ -25,18 +41,26 @@ jobs:
       matrix:
         os: [Linux, macOS]
         arch: [default, complex]
-        include:
+        platform: [linux/amd64, linux/arm64]
+        # Cannot use inputs in exclude clauses, so add a dummy matrix dim
+        is_dev_build:
```
connorjward: use same name as `inputs`
```suggestion
        build_dev:
```
maybe on a single line

connorjward (changes requested):

.github/workflows/docker.yml
JHopeCollins: Actionlint says that putting it on a single line is not acceptable.

PR #4404, opened 2025-06-25
Recommend running firedrake-clean after updating

JHopeCollins:
As it says in the new docs paragraph, there's no guarantees that old generated code will work with newer Firedrake or PETSc builds. This attempts to prevent users running into that.


docs/source/install.rst
```diff
@@ -221,6 +221,12 @@ to set the right environment variables and then run::
 
      $ pip install --upgrade firedrake
 
+Previously generated code generated may not be compatible with a newer
```
JHopeCollins:
```suggestion
Previously generated code may not be compatible with a newer
```

connorjward (approved):
LGTM

connorjward (approved):

PR #4405, opened 2025-06-25
Support MatIS and BDDC

pbrubeck:
# Description
Support `assemble(a, mat_type="is")`


firedrake/assemble.py
```diff
@@ -1366,6 +1366,10 @@ def _make_sparsity(test, trial, mat_type, sub_mat_type, maps_and_regions):
         except SparsityFormatError:
             raise ValueError("Monolithic matrix assembly not supported for systems "
                              "with R-space blocks")
+        if mat_type == "is":
+            rlgmap = unghosted_lgmap(sparsity._dsets[0].lgmap, test.function_space())
+            clgmap = unghosted_lgmap(sparsity._dsets[1].lgmap, trial.function_space())
+            sparsity._lgmaps = (rlgmap, clgmap)
```
connorjward: This encodes the boundary conditions into the definition of the matrix, which is a bit bizarre. Why can't MATIS set lgmaps after the fact?

pbrubeck: This does not encode BCs, just the removal of repeated ghost dofs. This is the correct way of using MatIS

pbrubeck: The way we handle BCs on matrices requires changing the LGmap after creating the matrix. This is not possible for a MatIS, as the LGMap is what defines it. If you change the LGMap then the matrix is destroyed.

For MatIS, BCs are never imposed on the LGMap. Instead we assemble the matrix without BCs, and we zero out the BC rows and columns.

firedrake/assemble.py
```diff
@@ -2177,3 +2181,44 @@ def index_function_spaces(form, indices):
...
+        indices[mask] = -1
+    return PETSc.LGMap().create(indices=indices, bsize=bsize, comm=lgmap.comm)
+
+
+def unghosted_lgmap(lgmap, V, block=True):
+    block_size = lgmap.getBlockSize() if block else 1
+
+    mesh_dm = V.mesh().topology_dm
+    depth = mesh_dm.getDepth()
+    start, end = mesh_dm.getDepthStratum(depth)
```
wence-: AKA `getHeightStratum(0)`

pyop2/types/mat.py
```diff
@@ -814,11 +825,19 @@ def set_local_diagonal_entries(self, rows, diag_val=1.0, idx=None):
...
-        if len(rows) > 0:
-            values = np.full(rows.shape, diag_val, dtype=dtypes.ScalarType)
-            self.handle.setValuesLocalRCV(rows, rows, values,
-                                          addv=PETSc.InsertMode.INSERT_VALUES)
+        if self.handle.type == "is":
+            self.handle.assemble()
+            # PETSc does not properly handle local dofs that map
+            # to a negative global index
+            rmap, _ = self.handle.getLGMap()
+            rows = rows[rmap.apply(rows) > -1]
```
wence-: Is this a bug that should be upstreamed?

firedrake/assemble.py
```diff
@@ -2177,3 +2181,44 @@ def index_function_spaces(form, indices):
...
+            if mesh_dm.getLabelValue("pyop2_ghost", seed) != -1:
+                continue
+            closure, _ = mesh_dm.getTransitiveClosure(seed, useCone=True)
+            for p in closure:
+                dof = section.getDof(p)
+                if dof <= 0:
+                    continue
+                off = section.getOffset(p)
+                # Local indices within W
+                W_indices = slice(block_size * off, block_size * (off + dof))
```
wence-: If `block` is `True` and `lgmap` has `block_size > 1` then this produces out of bounds indices.

wence-: Or conversely, if this is a mixed space where one component has block size > 1, then this masks out dofs even when running on a single process (which seems wrong).

firedrake/assemble.py
```diff
@@ -2177,3 +2181,44 @@ def index_function_spaces(form, indices):
...
+                own.extend(W_local_ises_indices[W_indices])
+
+    mask = numpy.setdiff1d(range(len(lgmap.indices)), own)
```
wence-: I think you're trying to mask out those entries in `lgmap.indices` which belong to dofs that are not reachable from owned cells?

I think for a non-mixed space, this is (if `block=True`):

```
cmap = V.cell_node_map()
mask = np.setdiff1d(
    cmap.values_with_halo[cmap.iterset.size:], 
    cmap.values[:cmap.iterset.size],
)
```

For a mixed space, you need to index through to `local_ises`, I think like this?

```
mask_pieces = []
for i, W in enumerate(V):
    iset = V.dof_dset.local_ises[i]
    cmap = W.cell_node_map()
    to_mask = np.setdiff1d(cmap.values_with_halo[cmap.iterset.size:], cmap.values[:cmap.iterset.size])
    bs = iset.block_size
    if bs > 1:
        to_mask = np.concatenate([i + bs * to_mask for i in range(bs)])
    mask_pieces.append(iset.indices[to_mask])
mask = np.concatenate(mask_pieces)
```

pbrubeck: `cell_node_map` is not defined for extruded meshes, but also there is no DM for extruded. The correct thing is to execute a pyop2 kernel that visits only the dofs that are reachable from owned cells.

pyop2/types/mat.py
pbrubeck: @stefanozampini caught this one, so maybe yes

stefanozampini: https://gitlab.com/petsc/petsc/-/merge_requests/8500

firedrake/preconditioners/bddc.py
```diff
@@ -81,6 +78,15 @@ def initialize(self, pc):
         appctx = self.get_appctx(pc)
         sobolev_space = V.ufl_element().sobolev_space
 
+        # set coordinates
+        if variant != "fdm" and is_lagrange(V):
```
stefanozampini: I have a test with Q2 elements which is failing here

```
  File "petsc4py/PETSc/libpetsc4py.pyx", line 1556, in petsc4py.PETSc.PCSetUp_Python
  File "/Users/zampins/Devel/firedrake/firedrake/preconditioners/base.py", line 191, in setUp
    super().setUp(pc)
  File "/Users/zampins/Devel/firedrake/firedrake/preconditioners/base.py", line 46, in setUp
    self.initialize(pc)
  File "/Users/zampins/Devel/firedrake/firedrake/preconditioners/bddc.py", line 82, in initialize
    if variant != "fdm" and is_lagrange(V):
                            ^^^^^^^^^^^^^^
  File "/Users/zampins/Devel/firedrake/firedrake/preconditioners/bddc.py", line 152, in is_lagrange
    nodes = V.finat_element.fiat_equivalent.dual_basis()
            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/zampins/.pyenv/versions/3.12.10/lib/python3.12/functools.py", line 998, in __get__
    val = self.func(instance)
          ^^^^^^^^^^^^^^^^^^^
  File "/Users/zampins/.pyenv/versions/firedrake/lib/python3.12/site-packages/finat/cube.py", line 65, in fiat_equivalent
    return FIAT_FlattenedDimensions(self.product.fiat_equivalent)
                                    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/zampins/.pyenv/versions/3.12.10/lib/python3.12/functools.py", line 998, in __get__
    val = self.func(instance)
          ^^^^^^^^^^^^^^^^^^^
  File "/Users/zampins/.pyenv/versions/firedrake/lib/python3.12/site-packages/finat/tensor_product.py", line 80, in fiat_equivalent
    A, B = self.factors
    ^^^^
ValueError: too many values to unpack (expected 2)

The above exception was the direct cause of the following exception:

  File "/Users/zampins/Devel/sandbox/pablo/test_bddc.py", line 112, in <module>
    solver.solve()

```

stefanozampini: this is a code that is working for me

```
            from firedrake.assemble import assemble
            cspace = VectorFunctionSpace(V.mesh(), V.ufl_element().family() , V.ufl_element().degree(), variant=V.ufl_element().variant())
            interp = interpolate(SpatialCoordinate(V.mesh()), space)
            coords = assemble(interp).dat.data_ro.copy()
            dim = coords.shape[1] # this should be replaced by the number of dofs per point
            coords = coords.repeat(dim, axis=0)
            bddcpc.setCoordinates(coords)
```

firedrake/preconditioners/bddc.py
```diff
@@ -81,6 +79,13 @@ def initialize(self, pc):
         appctx = self.get_appctx(pc)
         sobolev_space = V.ufl_element().sobolev_space
 
+        # Set coordinates
+        if is_lagrange(V.finat_element):
+            degree = V.ufl_element().embedded_superdegree
+            W = VectorFunctionSpace(V.mesh(), "Lagrange", degree, variant=variant)
+            coords = Function(W).interpolate(V.mesh().coordinates)
+            bddcpc.setCoordinates(coords.dat.data_ro.repeat(V.block_size, axis=0))
+
```
stefanozampini:
```
$ pytest -s tests/firedrake/regression/test_bddc.py -k test_bddc_aij[cube-vector-Q-4]
...
petsc4py/PETSc/libpetsc4py.pyx:1646: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
firedrake/preconditioners/bddc.py:136: in apply
    self.pc.apply(x, y)
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

>   ???
E   petsc4py.PETSc.Error: error code 83
E   [0] PCApply() at /Users/zampins/Devel/petsc/src/ksp/pc/interface/precon.c:525
E   [0] PCSetUp() at /Users/zampins/Devel/petsc/src/ksp/pc/interface/precon.c:1086
E   [0] PCSetUp_BDDC() at /Users/zampins/Devel/petsc/src/ksp/pc/impls/bddc/bddc.c:1529
E   [0] PCBDDCAnalyzeInterface() at /Users/zampins/Devel/petsc/src/ksp/pc/impls/bddc/bddcprivate.c:7500
E   [0] Invalid number of local coordinates! Got 31680, expected 4080

petsc4py/PETSc/PC.pyx:597: Error
```

firedrake/assemble.py
```diff
@@ -1366,6 +1369,18 @@ def _make_sparsity(test, trial, mat_type, sub_mat_type, maps_and_regions):
         except SparsityFormatError:
             raise ValueError("Monolithic matrix assembly not supported for systems "
                              "with R-space blocks")
+
+        # TODO reconstruct dof_dset with the unghosted lgmap
```
pbrubeck: @ksagiyam is it possible to get a `RestrictedFunctionSpace` with the unghosted `FunctionSpace`  (the space that corresponds to `DistributedMeshOverlapType.None`)?

pbrubeck: Alternatively, if I reconstruct the FunctionSpace with the mesh without the overlap, would I get consistent dof orderings? How do I recover the mesh without overlap?

ksagiyam: You can make a `RestrictedFunctionSpace` with `boundary_set=ghost entities`.

DoF ordering on each entity is the same only for CG and DG spaces (assuming that cone orderings are preserved). Otherwise DoFs depend on the global vertex numbers, which are arbitrary.

I think what we want is to be able to pass `ignore_halos=True` to `op2.Sparsity` and have `op2.DataSet.lgmap` (and similar) return the correct lgmap in the first place depending on the value of `ignore_halos`?

firedrake/preconditioners/bddc.py
```diff
@@ -140,3 +148,21 @@ def get_vertex_dofs(V):
     V.dof_dset.lgmap.apply(indices, result=indices)
     vertex_dofs = PETSc.IS().createGeneral(indices, comm=V.comm)
     return vertex_dofs
+
+
+def is_lagrange(finat_element):
```
stefanozampini, line 179: @pbrubeck This returns true when using `N1curl` elements!

pbrubeck: I think that is what I intended, in case you wanted to associate coordinates to each edge midpoint

firedrake/assemble.py
pbrubeck: Only `VectorElements` or `TensorElements` can have `blocksize != 1`. Any `MixedElement` has blocksize = 1

firedrake/preconditioners/bddc.py
stefanozampini: Coordinates are needed only for H1/H2 problems. I'm changing the logic to `sobolev_space in [H1,H2]`

pbrubeck: That's fine

firedrake/preconditioners/bddc.py
stefanozampini: the problem here is that embedded_superdegree is 8. Shouldn't we just use degree?

pyop2/types/mat.py
```diff
@@ -618,11 +620,16 @@ def _init_monolithic(self):
         rset, cset = self.sparsity.dsets
         rlgmap = rset.unblocked_lgmap
         clgmap = cset.unblocked_lgmap
-        mat.createAIJ(size=((self.nrows, None), (self.ncols, None)),
-                      nnz=(self.sparsity.nnz, self.sparsity.onnz),
-                      bsize=1,
-                      comm=self.comm)
+        if self.mat_type == "is":
+            # FIXME monolithic lgmaps
```
stefanozampini: @pbrubeck Is this still broken? We can do mixed Poisson  H(div) x L2 if we have monolithic l2gmaps

.github/workflows/core.yml
```diff
@@ -210,6 +210,7 @@ jobs:
           pip install -U pip
           pip install --group ./firedrake-repo/pyproject.toml:ci
 
+          pip install -I "firedrake-fiat @ git+https://github.com/firedrakeproject/fiat.git@pbrubeck/fix/permutations"
```
pbrubeck:
```suggestion
```

pyop2/types/mat.py
stefanozampini: This should fix the tests https://gitlab.com/petsc/petsc/-/merge_requests/8705
Is this comment  `# FIXME monolithic lgmaps` outdated?

pbrubeck: The LGMaps seem correct as the test without bcs is passing

pbrubeck:
```suggestion
```

firedrake/assemble.py
pbrubeck: Thanks @wence-, in the end we are using your approach.

firedrake/assemble.py
pbrubeck: This is no longer relevant.

pyop2/types/mat.py
```diff
@@ -616,13 +666,23 @@ def _init_dense(self):
...
         rset, cset = self.sparsity.dsets
-        rlgmap = rset.unblocked_lgmap
-        clgmap = cset.unblocked_lgmap
-        mat.createAIJ(size=((self.nrows, None), (self.ncols, None)),
-                      nnz=(self.sparsity.nnz, self.sparsity.onnz),
-                      bsize=1,
-                      comm=self.comm)
+        if self.mat_type == "is":
+            rmaps = [None for _ in range(len(rset.local_ises))]
+            cmaps = [None for _ in range(len(cset.local_ises))]
```
pbrubeck:
```suggestion
            rmaps = [None for _ in rset.local_ises]
            cmaps = [None for _ in cset.local_ises]
```

firedrake/matrix_free/operators.py
```diff
@@ -377,11 +377,20 @@ def createSubMatrix(self, mat, row_is, col_is, target=None):
...
+            else:
+                col_inds = find_sub_block(col_is, col_ises, comm=self.comm)
+        except LookupError:
+            # No match for sub_block
+            # use default PETSc implementation via MATSHELL
+            popmethod = self.createSubMatrix
+            self.createSubMatrix = None
+            submat = mat.createSubMatrix(row_is, col_is)
+            self.createSubMatrix = popmethod
+            return submat
```
connorjward, line 393: This is very hard to understand.

firedrake/assemble.py
```diff
@@ -1456,8 +1459,9 @@ def _apply_bc(self, tensor, bc, u=None):
             # Set diagonal entries on bc nodes to 1 if the current
             # block is on the matrix diagonal and its index matches the
             # index of the function space the bc is defined on.
+            if op2tensor.handle.getType() == "is":
+                op2tensor.handle.assemble()
```
connorjward, line 1498: This is a weird place to put this. Why is this only necessary for this matrix type?

firedrake/functionspaceimpl.py
```diff
@@ -852,7 +852,7 @@ def boundary_nodes(self, sub_domain):
         return self._shared_data.boundary_nodes(self, sub_domain)
 
     @PETSc.Log.EventDecorator()
-    def local_to_global_map(self, bcs, lgmap=None):
+    def local_to_global_map(self, bcs, lgmap=None, mat_type=None):
```
connorjward, line 859: This function is already a bit of a mess so I don't mind there being this extra argument for now (I rewrite this in pyop3).

pyop2/types/mat.py
```diff
@@ -635,7 +695,8 @@ def _init_monolithic(self):
         mat.setOption(mat.Option.KEEP_NONZERO_PATTERN, True)
         # We completely fill the allocated matrix when zeroing the
         # entries, so raise an error if we "missed" one.
-        mat.setOption(mat.Option.UNUSED_NONZERO_LOCATION_ERR, True)
+        if self.mat_type != "is":
+            mat.setOption(mat.Option.UNUSED_NONZERO_LOCATION_ERR, True)
```
connorjward, line 701: Why does this need to happen?

firedrake/matrix_free/operators.py
stefanozampini: This is how PETSc's `MATPYTHON` class works. The issue is that firedrake's `MATPYTHON` class implements a `createSubMatrix` method that only works when the submatrix is obtained from a subfield.  For general matrix extraction `MATPYTHON` may use a default submatrix operation (`MatCreateSubMatrixVirtual`) If there's no `createSubMatrix` method

firedrake/assemble.py
stefanozampini: because we need to flag the matrix as assembled (just the cost of an allreduce) before we can operate on the submatrix and apply `MatZeroRows`. For `AIJ`, firedrake takes a different code path and uses `MatSetValuesLocal`

pyop2/types/mat.py
stefanozampini: Because the preallocation is done on the global rows and there are extra dofs allocated in case of a MATIS. Think, for example, of a dof vertex for `CG1` on the process boundary shared by 2 triangular elements. With AIJ, the global row has 4 nonzeros. With MATIS, each of the two processes has 3 nonzeros

firedrake/functionspaceimpl.py
pbrubeck: You are very merciful and kind

pyop2/types/mat.py
pbrubeck: Don't we want a local preallocator then?

stefanozampini: someone needs to write the code :-)

firedrake/assemble.py
connorjward: This looks like it's already being called in `mat.py` line 887

firedrake/matrix_free/operators.py
```diff
@@ -377,11 +377,20 @@ def createSubMatrix(self, mat, row_is, col_is, target=None):
...
-        else:
-            col_inds = find_sub_block(col_is, col_ises, comm=self.comm)
+        try:
+            row_inds = find_sub_block(row_is, row_ises, comm=self.comm)
+            if row_is == col_is and row_ises == col_ises:
+                col_inds = row_inds
+            else:
+                col_inds = find_sub_block(col_is, col_ises, comm=self.comm)
+        except LookupError:
+            # No match for sub_block
```
pbrubeck:
```suggestion
            # Attemping to extract a submatrix that does not match with a subfield
```

pyop2/types/mat.py
pbrubeck, line 701:
```suggestion
        if self.mat_type != "is":
            # The local matrix will have fewer nonzeros than the one prescribed
            # in the global sparsity pattern
            mat.setOption(mat.Option.UNUSED_NONZERO_LOCATION_ERR, True)
```

firedrake/assemble.py
pbrubeck: That calls `op2tensor[index, index].handle.assemble()`, not `op2tensor.handle.assemble()`

pbrubeck, line 1498:
```suggestion
            if op2tensor.handle.getType() == "is":
                # Flag the entire matrix as assembled before indexing the diagonal block
                op2tensor.handle.assemble()
```

pyop2/types/mat.py
pbrubeck: @stefanozampini I think I got this one wrong. Should we expect fewer nonzeros on the local submat because we ignore the columns that are not owned by this process.

firedrake/matrix_free/operators.py
```diff
@@ -377,11 +377,20 @@ def createSubMatrix(self, mat, row_is, col_is, target=None):
...
-            col_inds = find_sub_block(col_is, col_ises, comm=self.comm)
+        try:
+            row_inds = find_sub_block(row_is, row_ises, comm=self.comm)
+            if row_is == col_is and row_ises == col_ises:
+                col_inds = row_inds
+            else:
+                col_inds = find_sub_block(col_is, col_ises, comm=self.comm)
+        except LookupError:
+            # Attemping to extract a submatrix that does not match with a subfield
+            # use default PETSc implementation via MATSHELL
```
pbrubeck:
```suggestion
            # use default PETSc implementation (MatCreateSubMatrixVirtual) via MATSHELL
```

firedrake/functionspaceimpl.py
JHopeCollins, line 859: Please can you add the kwargs to the docstring?

firedrake/functionspaceimpl.py
```diff
@@ -880,10 +880,15 @@ def local_to_global_map(self, bcs, lgmap=None):
                 bsize = lgmap.getBlockSize()
                 assert bsize == self.block_size
         else:
-            # MatBlock case, LGMap is already unrolled.
-            indices = lgmap.block_indices.copy()
-            bsize = lgmap.getBlockSize()
-            unblocked = True
+            # MatBlock case, the LGMap is implementation dependent
+            if mat_type == "is":
+                bsize = 1
```
JHopeCollins: Why is the block size ignored for mat type IS?

firedrake/matrix_free/operators.py
JHopeCollins: How does the default operation work? Is it a reasonable fallback here or do you expect it to fail in some situations? If that is the case then there should be extra checks or we should raise a warning (preference of course to extra checks if possible).

firedrake/matrix_free/operators.py
```diff
@@ -377,11 +377,20 @@ def createSubMatrix(self, mat, row_is, col_is, target=None):
...
-            col_inds = find_sub_block(col_is, col_ises, comm=self.comm)
+        try:
+            row_inds = find_sub_block(row_is, row_ises, comm=self.comm)
+            if row_is == col_is and row_ises == col_ises:
+                col_inds = row_inds
+            else:
+                col_inds = find_sub_block(col_is, col_ises, comm=self.comm)
+        except LookupError:
+            # Attemping to extract a submatrix that does not match with a subfield
+            # use default PETSc implementation (MatCreateSubMatrixVirtual) via MATSHELL
```
JHopeCollins: I'm assuming that this is two sentences and not one?
```suggestion
            # Attemping to extract a submatrix that does not match with a subfield.
            # Use default PETSc implementation (MatCreateSubMatrixVirtual) via MATSHELL instead.
```

firedrake/preconditioners/bddc.py
```diff
@@ -22,17 +26,13 @@ class BDDCPC(PCBase):
     - ``'bddc_pc_bddc_dirichlet'`` to set sub-KSPs on subdomain interiors,
     - ``'bddc_pc_bddc_coarse'`` to set the coarse solver KSP.
 
-    This PC also inspects optional arguments supplied in the application context:
-    - ``'discrete_gradient'`` for problems in H(curl), this sets the arguments
-    (a Mat tabulating the gradient of the auxiliary H1 space) and
+    This PC also inspects optional callbacks supplied in the application context:
+    - ``'get_discrete_gradient'`` for 3D problems in H(curl), this is a callable that
```
JHopeCollins, line 31: Is there any way that we can make sure that API changes gets picked up by the change log in the next major release?
Doesn't just apply to this PR but this is when I thought of it.
@connorjward?

firedrake/preconditioners/bddc.py
```diff
@@ -52,13 +54,19 @@ def initialize(self, pc):
         bddcpc.setOperators(*pc.getOperators())
         bddcpc.setType(PETSc.PC.Type.BDDC)
 
+        # Allow viewing preconditioning matrix for debugging purposes
+        P.viewFromOptions('-view_pmat', bddcpc)
```
JHopeCollins: Should this still be here?

tests/firedrake/regression/test_assemble.py
```diff
@@ -115,6 +115,77 @@ def test_mat_nest_real_block_assembler_correctly_reuses_tensor(mesh):
...
+        anest.createNest(blocks,
+                         isrows=V.dof_dset.field_ises,
+                         iscols=V.dof_dset.field_ises,
+                         comm=ais.comm)
+        anest.convert("aij", aij)
+    else:
+        assert ais.type == "is"
+        ais.convert("aij", aij)
+
+    aij_ref = assemble(a, bcs=bcs, mat_type="aij").petscmat
```
JHopeCollins: This was already assembled on line 155
```suggestion
```

tests/firedrake/regression/test_assemble.py
```diff
@@ -115,6 +115,77 @@ def test_mat_nest_real_block_assembler_correctly_reuses_tensor(mesh):
     assert A2.M is A1.M
 
 
+@pytest.mark.parallel
+@pytest.mark.parametrize("shape,mat_type", [("scalar", "is"), ("vector", "is"), ("mixed", "is"), ("mixed", "nest")])
```
JHopeCollins, line 119: Does it make sense to test any Hdiv or Hcurl elements? BDM for example?

tests/firedrake/regression/test_bddc.py
```diff
@@ -120,11 +135,34 @@ def mesh(request):
 
 
 @pytest.mark.parallel
+@pytest.mark.parametrize("degree", range(1, 3))
+@pytest.mark.parametrize("variant", ("spectral", "fdm"))
+def test_vertex_dofs(mesh, variant, degree):
```
JHopeCollins: Please can you add a quick docstring? I'm not familiar with BDDC so it isn't clear to me why I should expect this test to pass or fail.

firedrake/preconditioners/bddc.py
connorjward: I don't think so. We'd need to maintain a changelog ourselves between releases that I could paste in I expect.

firedrake/functionspaceimpl.py
pbrubeck: This is just the way LGMap was implemented for MatIS

firedrake/matrix_free/operators.py
pbrubeck: To compute the product between a (sub-)matrix and a sub-vector, `MatCreateSubMatrixVirtual` embeds the subvector into the full space by padding with zeros, applies the entire matrix, and then extracts the subvector.

firedrake/preconditioners/bddc.py
pbrubeck: Also I don't expect users ever needing to pass the gradient here, the default does it quite well.

firedrake/preconditioners/bddc.py
pbrubeck: This has extremely handy for debugging, might as well leave it.

firedrake/functionspaceimpl.py
pbrubeck: Something like?
```
mat_type: str
    The matrix assembly type
```

firedrake/matrix_free/operators.py
pbrubeck: I honestly don't know why we put this change in this PR.

tests/firedrake/regression/test_bddc.py
pbrubeck:
```suggestion
def test_vertex_dofs(mesh, variant, degree):
    """Check that we extract the right number of vertex dofs from a high order Lagrange space."""
```

firedrake/preconditioners/bddc.py
JHopeCollins: I misunderstood what this is doing, I thought it was always printing the mat view.
I agree that this is very useful, but shouldn't the option be something like `f"{self.prefix}pmat_view}"`?

tests/firedrake/regression/test_assemble.py
pbrubeck: One could test that as well, it should work. But I think the cases we have here are the important ones.

firedrake/functionspaceimpl.py
JHopeCollins: At a PETSc level or a Firedrake level?

tests/firedrake/regression/test_assemble.py
JHopeCollins: Given that the PR description says "Enables BDDC for problems in H(curl) and H(div)", I think it's quite important to actually test these spaces. Isn't it just adding a couple extra parametrize options?

firedrake/preconditioners/bddc.py
pbrubeck: That would lengthen the string, I would either keep as is or entirely remove it.

firedrake/preconditioners/bddc.py
```diff
@@ -52,13 +54,19 @@ def initialize(self, pc):
...
+
```
pbrubeck:
```suggestion
```

firedrake/matrix_free/operators.py
pbrubeck: @stefanozampini was this necessary for this PR?

firedrake/functionspaceimpl.py
pbrubeck: At the PETSc level

tests/firedrake/regression/test_assemble.py
pbrubeck: Tests for H(curl) / H(div) have been added in `test_bddc`.

firedrake/preconditioners/bddc.py
JHopeCollins: I'm fine with it being there, but any options should be properly prefixed just like any other PETSc option.

firedrake/functionspaceimpl.py
JHopeCollins: Ok, nothing we can do about it then.

firedrake/matrix_free/operators.py
stefanozampini: I don't remember, but for sure is the right thing to do since the firedrake method only extracts subfields

firedrake/preconditioners/bddc.py
stefanozampini: `viewFromOptions` takes the prefix of the object automatically

firedrake/matrix_free/operators.py
pbrubeck: I moved this to its own PR https://github.com/firedrakeproject/firedrake/pull/4693

firedrake/functionspaceimpl.py
```diff
@@ -852,10 +852,31 @@ def boundary_nodes(self, sub_domain):
...
+        mat_type: str
+            The matrix assembly type. This is required as different matrix types
+            handle the LGMap differently for MixedFunctionSpace.
+
+        Note
+        ----
+            For a :func:`.VectorFunctionSpace` or :func:`.TensorFunctionSpace` the returned
+            LGMap will be the scalar one, unless the bcs are imposed on a particular component.
+            For a :class:`MixedFunctionSpace` the returned LGMap is unblocked,
+            unless mat_type == "is".
```
JHopeCollins, line 877: Thanks for adding the docstring.

~~Is this last sentence the right way round? On line 909 there is `unblocked = False` if `mat_type == "is"`~~

Never mind, I got myself in a twist with double negatives.

tests/firedrake/regression/test_bddc.py
```diff
@@ -120,11 +134,44 @@ def mesh(request):
...
+
+
+@pytest.mark.parallel
+@pytest.mark.parametrize("family,degree", [("CG", 3), ("N1curl", 3), ("N1div", 3)])
+def test_bddc_aij_simplex(family, degree):
+    nx = 4
+    mesh = UnitCubeMesh(nx, nx, nx)
+    variant = None
+    bcs = True
+    expected = {"CG": 13, "N1curl": 14, "N1div": 12}[family]
```
JHopeCollins: Where are these numbers from?

tests/firedrake/regression/test_assemble.py
JHopeCollins: Why not test assembly for these elements? Surely that is a more unit-y test than a whole solve?

tests/firedrake/regression/test_bddc.py
pbrubeck: I first ran the tests locally, got iteration counts, then added 1.

tests/firedrake/regression/test_assemble.py
pbrubeck: The solve test is necessary, anyway. The MatIS assembly of RT elements is not more challenging than CG.

pbrubeck: We cannot just add another parameter to test H(curl) / H(div), as we would need a different bilinear form and BCs. I included a test for correctness to make sure that the H(curl) / H(div)  solver actually arrives to the right solution.

tests/firedrake/regression/test_bddc.py
JHopeCollins: Is there any other way to test it? Any theory on expected rates for example?

If there really is no other way, then how robust are these counts? How far past the tolerance are they? Is adding just 1 more enough to make them robust across different platforms and direct solver libraries/versions?
Testing iteration counts is often the cause of stochastic failures. If at all possible then we should avoid adding any more of them unless there's theory for why we expect specifically that count.

pbrubeck: I added an extra step in the test ensuring that the A-norm of error in the solution satisfies the tolerance that we passed to the solver.

I think that's a discussion that falls beyond the scope of this PR. If you can point me to any other test where this issue is addressed with the theory considerations that you have in mind, I can happily implement that for this solver.

JHopeCollins: Changing the existing tests is outside this PR, I agree. My point was that it would be good not to add any more than we already have.

I don't know anything about BDDC methods which was why I was asking if there exists any theory for expected convergence rates (and therefore for expected iteration counts). We do it in asQ, although admittedly for that PC we have very tight convergence bounds: https://github.com/firedrakeproject/asQ/blob/8b4aeb821b0bc8030838d01b5cdcc71249f14c23/tests/preconditioners/test_allatonce_pcs.py#L65-L89

It doesn't sound like there is something similar for BDDC, so it seems like we will have to stick with iteration count regressions.

JHopeCollins (dismissed):

tests/firedrake/regression/test_bddc.py
```diff
@@ -85,46 +97,90 @@ def solve_riesz_map(mesh, family, degree, bcs, condense):
...
+
+@pytest.mark.parallel
+@pytest.mark.parametrize("family,degree", [("CG", 3), ("N1curl", 3), ("N1div", 3)])
+def test_bddc_aij_simplex(rg, family, degree):
+    """Test h-independence of condition number by measuring iteration counts"""
+    variant = None
+    bcs = True
+    meshes = [UnitCubeMesh(nx, nx, nx) for nx in (3, 6)]
+    its = [solve_riesz_map(rg, m, family, degree, variant, bcs) for m in meshes]
+    assert (np.diff(its) <= 1).all()
```
pbrubeck: @JHopeCollins here we are now testing theoretical bounds, but we might be in a pre-asymptotic regime still. I wouldn't say this is completely robust with respect to stochastic things like mesh partitioning. Iterations might suddenly differ by 2 instead of 1. 

I think we can compute the condition number of the preconditioned system projected into the Krylov space.

pbrubeck: Reasonable test problem sizes are still in the preasymptotic regime. I'll just leave this as is, measuring iteration counts.

tests/firedrake/regression/test_bddc.py
```diff
@@ -85,46 +97,91 @@ def solve_riesz_map(mesh, family, degree, bcs, condense):
...
+
+@pytest.mark.parallel
+@pytest.mark.parametrize("family,degree", [("Q", 4)])
+@pytest.mark.parametrize("vector", (False, True), ids=("scalar", "vector"))
+def test_bddc_aij_quad(rg, mh, family, degree, vector):
+    """Test h-independence of condition number by measuring iteration counts"""
+    variant = None
+    bcs = True
+    its = [solve_riesz_map(rg, m, family, degree, variant, bcs, vector=vector) for m in mh]
+    assert (np.diff(its) <= 1).all()
```
pbrubeck:
```suggestion
    assert (np.diff(its) <= 2).all()
```

tests/firedrake/regression/test_bddc.py
```diff
@@ -85,46 +97,91 @@ def solve_riesz_map(mesh, family, degree, bcs, condense):
...
+
+@pytest.mark.parallel
+@pytest.mark.parametrize("family,degree", [("CG", 3), ("N1curl", 3), ("N1div", 3)])
+def test_bddc_aij_simplex(rg, family, degree):
+    """Test h-independence of condition number by measuring iteration counts"""
+    variant = None
+    bcs = True
+    meshes = [UnitCubeMesh(nx, nx, nx) for nx in (3, 6)]
+    its = [solve_riesz_map(rg, m, family, degree, variant, bcs) for m in meshes]
+    assert (np.diff(its) <= 1).all()
```
pbrubeck:
```suggestion
    assert (np.diff(its) <= 2).all()
```

tests/firedrake/regression/test_bddc.py
stefanozampini: The eigenvalues of the preconditioned operator with CG are `lambda_min = 1`, `lambda_max < (1 + log (H/h))^2`. You can check the minimum eigenvalue if you really want.

pbrubeck: Is there any constant in front of the upper bound? If not, then should we expect `kappa_{h/2}^{1/2} < kappa_h^{1/2} + log(2)`?

pbrubeck: Assuming a fixed mesh partition, of course

connorjward: > I think that's a discussion that falls beyond the scope of this PR. If you can point me to any other test where this issue is addressed with the theory considerations that you have in mind, I can happily implement that for this solver.

This is definitely an interesting idea. I never really thought about the fact that the current iteration counts we use are just "run on my machine and add 1". That doesn't really seem like it's testing much. It would be cool if we could establish some sort of policy for this in future.

tests/firedrake/regression/test_bddc.py
```diff
@@ -33,36 +40,42 @@ def solver_parameters(static_condensation=True):
             "facet_fdm_static_condensation": True,
             "facet_fdm_pc_use_amat": False,
             "facet_fdm_mat_type": "is",
-            "facet_fdm_mat_is_allow_repeated": repeated,
+            "facet_fdm_mat_is_allow_repeated": True,
```
pbrubeck:
```suggestion
            "facet_fdm_mat_is_allow_repeated": cellwise,
```

tests/firedrake/regression/test_bddc.py
```diff
@@ -33,36 +40,42 @@ def solver_parameters(static_condensation=True):
...
         }
     else:
+        assert variant == "fdm"
         sp = {
             "pc_type": "python",
             "pc_python_type": "firedrake.FDMPC",
             "fdm_pc_use_amat": False,
-            "fdm_mat_type": "is",
-            "fdm_mat_is_allow_repeated": repeated,
+            "fdm_mat_is_allow_repeated": True,
```
pbrubeck:
```suggestion
            "fdm_mat_is_allow_repeated": cellwise,
```

pyop2/types/mat.py
```diff
@@ -341,6 +341,55 @@ def masked_lgmap(lgmap, mask, block=True):
...
+            owned += offset
+            ghost += offset
+        mask = np.concatenate(mask_pieces)
+    else:
+        raise NotImplementedError("MatIS does not support variable extrusion with overlap.")
+    return mask
+
+
+def unghosted_lgmap(dset, node_maps):
+    """Return a local-to-global map where the nodes on ghost cell are masked out."""
```
pbrubeck:
```suggestion
    """Return a local-to-global map where the nodes on ghost cells are masked out."""
```

tests/firedrake/regression/test_bddc.py
JHopeCollins: > It would be cool if we could establish some sort of policy for this in future.

Agreed, but we can discuss elsewhere, I'm resolving this thread now so the PR can go through.

Issue #4387, opened 2025-06-20
BUG:


miguelcoolchips:
**Describe the bug**
Adjoint calculation with DirichletBC as Control does not work with NonlinearVariationalSolver.

**Steps to Reproduce**
```python
import firedrake as fd
from firedrake import inner, grad, dx
from firedrake.adjoint import Control, ReducedFunctional
from pyadjoint import taylor_test, continue_annotation

continue_annotation()


def test_wrt_function_dirichlet_boundary():
    mesh = fd.UnitSquareMesh(100, 100)
    V = fd.FunctionSpace(mesh, "CG", 1)
    R = fd.FunctionSpace(mesh, "R", 0)

    bc_func = fd.Function(R, val=2.0)
    bc = fd.DirichletBC(V, bc_func, 1)
    u_ = fd.Function(V)

    v = fd.TestFunction(V)
    a = inner(grad(u_), grad(v)) * dx
    L = v * dx

    # Crashes when using NonlinearVariationalSolver.solve()
    # Error: assert taylor_test(Jhat, bc_func, bc_func_perturb) > 1.9
    problem = fd.NonlinearVariationalProblem(a - L, u_, bcs=bc)
    solver = fd.NonlinearVariationalSolver(problem)
    solver.solve()
    # Passes when using solve()
    # F = a - L
    # fd.solve(F == 0, u_, bc)

    J = fd.assemble(u_**2 * fd.dx)
    c = Control(bc_func)
    Jhat = ReducedFunctional(J, c)
    bc_func_perturb = fd.Function(R, val=2.0)
    assert taylor_test(Jhat, bc_func, bc_func_perturb) > 1.9


if __name__ == "__main__":
    test_wrt_function_dirichlet_boundary()

```

**Expected behavior**
Describe what you expected to happen, in the case of a regression,
include details of a setup where this behaviour was last seen.

**Error message**
```
Running Taylor test
Computed residuals: [0.0933330000202759, 0.04666650001013795, 0.023333250005068976, 0.011666625002534488]
Computed convergence rates: [np.float64(1.0), np.float64(1.0), np.float64(1.0)]
Traceback (most recent call last):
  File "/Users/miguel/work/glaciercore/tests/functional/adjoint/mfe-bc-adjoint.py", line 39, in <module>
    test_wrt_function_dirichlet_boundary()
[...]

Issue #4206, opened 2025-04-07
BUG: Disk checkpointing with `tape.timestepper()` causes incorrect BC evaluation at taping.


sghelichkhani:
When using disk checkpointing with `tape.timestepper()` (i.e., with `SingleDiskStorageSchedule`), we observe inconsistent behavior in the forward solve even before invoking any functional or derivative evaluation. This appears specifically when updating a `DirichletBC` field during a time-stepping loop.

Consider the following reproducer:

```
from firedrake import *
from firedrake.adjoint import *
from checkpoint_schedules import SingleDiskStorageSchedule as scheduler
# from checkpoint_schedules import SingleMemoryStorageSchedule as scheduler

continue_annotation()
enable_disk_checkpointing()

tape = get_working_tape()
tape.enable_checkpointing(scheduler())


mesh = checkpointable_mesh(UnitSquareMesh(5, 5))
V = FunctionSpace(mesh, "CG", 2)
T = Function(V).interpolate(0.0)
u = TrialFunction(V)
v = TestFunction(V)
a = inner(grad(u), grad(v)) * dx
x = SpatialCoordinate(mesh)
F = Function(V)
control = Control(F)
F.interpolate(sin(x[0] * pi) * sin(2 * x[1] * pi))
L = F * v * dx

bcs = [DirichletBC(V, T, (1,))]
uu = Function(V)

for i in tape.timestepper(iter(range(3))):
    T.assign(T + 1.0)
    solve(a == L, uu, bcs=bcs, solver_parameters={"ksp_type": "preonly", "pc_type": "lu"})

obj = assemble(uu * uu * dx)
pause_annotation()
print(f"Objective at the end of taping:{obj}")

rf = ReducedFunctional(obj, control)
print(f"Objective by calling functional: {rf(F)}")

```

would produce inconsistent values:

```
Objective at the end of taping:4.000901957850139
Objective by calling functional: 1.0009019578501521
```

whereas NOT checkpointing to disk, by using `SingleMemoryStorageSchedule` as scheduler (line 4) gives consistent AND different values:

```
Objective at the end of taping:9.000901957850123
Objective by calling functional: 9.000901957850061
```

jrmaddison:
This is my understanding based on some reading of the pyadjoint and Firedrake code, but perhaps others more familiar with the pyadjoint internals might correct:

[...]

PR #4408, opened 2025-06-26
Update bcs for adjoints

Ig-dolci:
# Description

This PR seeks to fix the [issue #4387](https://github.com/firedrakeproject/firedrake/issues/4387), which is related to the Nonlinear Variational Solve not updating the Dirichlet boundary conditions based on the checkpointed functions.

### DIFF
--- a/firedrake/bcs.py
+++ b/firedrake/bcs.py
@@ -450,12 +450,15 @@ class DirichletBC(BCBase, DirichletBCMixin):
             r = r.sub(idx)
             if u:
                 u = u.sub(idx)
+
+        bc_checkpointed = self.block_variable.checkpoint
+        bc = bc_checkpointed.checkpoint \
+            if bc_checkpointed is not None else self.function_arg
+
         if u:
-            r.assign(u - self.function_arg, subset=self.node_set)
-        elif self.block_variable.checkpoint is not None:
-            r.assign(self.block_variable.checkpoint.checkpoint, subset=self.node_set)
+            r.assign(u - bc, subset=self.node_set)
         else:
-            r.assign(self.function_arg, subset=self.node_set)
+            r.assign(bc, subset=self.node_set)
 
     def integrals(self):
         return []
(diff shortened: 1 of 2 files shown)


firedrake/bcs.py
```diff
@@ -450,10 +450,15 @@ def apply(self, r, u=None):
...
             if u:
                 u = u.sub(idx)
+
+        bc_checkpointed = self.block_variable.checkpoint
+        bc = bc_checkpointed.checkpoint \
+            if bc_checkpointed is not None else self.function_arg
+
         if u:
-            r.assign(u - self.function_arg, subset=self.node_set)
+            r.assign(u - bc, subset=self.node_set)
```
Ig-dolci, line 459: It is still missing a test for this case.

firedrake/bcs.py
```diff
@@ -450,10 +450,15 @@ def apply(self, r, u=None):
             r = r.sub(idx)
             if u:
                 u = u.sub(idx)
+
+        bc_checkpointed = self.block_variable.checkpoint
+        bc = bc_checkpointed.checkpoint \
+            if bc_checkpointed is not None else self.function_arg
```
pbrubeck, line 456: Can this logic be put into a new property that gets overloaded in the corresponding adjoint subclass of DirichletBC?

sghelichkhani:
I know this is still wip, but flagging that this breaks [one of our tests](https://github.com/g-adopt/g-adopt/actions/runs/15918133234) and the unit test [here](https://github.com/firedrakeproject/firedrake/issues/4206#issue-2975684117)

firedrake/bcs.py
pbrubeck: You will get this case by passing `pre_apply_bcs=False` to the NLVS

Ig-dolci: Thank you.

firedrake/bcs.py
Ig-dolci: Yes, it must be. I am still checking to see if this works. It looks like it requires more thought. Sia just wrote that this is leading some Gadopt tests to fail.

PR #4421, opened 2025-07-02
Pin SLEPc using firedrake-configure

connorjward:
Otherwise the latest SLEPc release is always used and this may not be compatible with the pinned slepc4py.

This should fix the current failure to build on release. It is a little clunky but I think it's OK.


JHopeCollins:
Any idea why it seems to be failing with the same errors?

PR #4427, opened 2025-07-04
Use python3 in firedrake-run-split-tests so it works in the Docker containers

JHopeCollins:
In `firedrake-split-tests` we use `python -m pytest` to run the tests.
However, by default Ubuntu has no `python` command, instead requiring `python3` explicitly so the script doesn't work in the Docker containers.

It used to work previously because we created a venv in the containers so `python` was in `$VIRTUAL_ENV/bin`.


connorjward:
> I need to change that policy. It should say that we periodically merge release into master. There is no need for an extra PR into master.

This is now done by the way.

PR #4432, opened 2025-07-08
Make `interpolate` function avoid using `Interpolator`

leo-collins:
The `interpolate` function now returns a symbolic `Interpolate` without using`Interpolator.interpolate`

### DIFF
--- a/firedrake/interpolation.py
+++ b/firedrake/interpolation.py
@@ -29,6 +29,7 @@ from firedrake.mesh import MissingPointsBehaviour, VertexOnlyMeshMissingPointsEr
 from firedrake.petsc import PETSc
 from firedrake.halo import _get_mtype as get_dat_mpi_type
 from firedrake.cofunction import Cofunction
+from firedrake.function import Function
 from mpi4py import MPI
 
 from pyadjoint import stop_annotating, no_annotations
@@ -150,7 +151,7 @@ class Interpolate(ufl.Interpolate):
 
 
 @PETSc.Log.EventDecorator()
-def interpolate(expr, V, *args, **kwargs):
+def interpolate(expr, V, *function, subset=None, access=op2.WRITE, allow_missing_dofs=False, default_missing_val=None):
     """Returns a UFL expression for the interpolation operation of ``expr`` into ``V``.
 
     :arg expr: a UFL expression.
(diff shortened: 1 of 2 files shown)


firedrake/interpolation.py
```diff
@@ -176,35 +177,47 @@ def interpolate(expr, V, *args, **kwargs):
...
-        return Interpolator(
-            expr, V.function_space().dual(), *args, **kwargs
-        ).interpolate(V, adjoint=adjoint, default_missing_val=default_missing_val)
-    return Interpolator(expr, V, *args, **kwargs).interpolate(default_missing_val=default_missing_val)
+    elif isinstance(V, Function):
+        function_space = V.function_space()
+    elif isinstance(V, functionspaceimpl.WithGeometry):
+        function_space = V
+    else:
+        raise TypeError(f"V must be a FunctionSpace, Function or Cofunction, not {type(V)}")
```
```suggestion
        raise TypeError(f"V must be a FunctionSpace, Function or Cofunction, not {type(V).__name__}")
```
otherwise it prints weird

firedrake/interpolation.py
```diff
@@ -176,35 +177,47 @@ def interpolate(expr, V, *args, **kwargs):
...
+
+    # Cope with the different convention of `Interpolate` and `Interpolator`:
+    #  -> Interpolate(Argument(V1, 1), Argument(V2.dual(), 0))
+    #  -> Interpolator(Argument(V1, 0), V2)
+    expr_args = extract_arguments(expr)
+    if expr_args and expr_args[0].number() == 0:
+        v, = expr_args
+        expr = replace(expr, {v: v.reconstruct(number=1)})
```
connorjward: This seems a bit strange. Why do we care about the convention for `Interpolator` here?

firedrake/interpolation.py
```diff
@@ -176,35 +177,47 @@ def interpolate(expr, V, *args, **kwargs):
...
+
+    interp = Interpolate(expr, function_space,
+                         subset=subset, access=access,
+                         allow_missing_dofs=allow_missing_dofs,
+                         default_missing_val=default_missing_val)
+    if adjoint:
+        interp = expr_adjoint(interp)
```
connorjward: Can't this be handled in `assemble`?

connorjward (changes requested):
This seems like an improvement but I still wonder why we are doing so much in this function. Why can't we just have something like
```py
def interpolate(*args, **kwargs):
    return Interpolate(*args, **kwargs)
```
(I know `V` can be a `Function` or `Cofunction` so some parsing there might be valid)

connorjward (dismissed):
Seems fine to me.

PR #4442, opened 2025-07-15
Correlation operators for weighted norms in optimisation methods

JHopeCollins:


firedrake/adjoint/correlations.py
```diff
@@ -0,0 +1,489 @@
...
+)
+
+
+def _make_rhs(b):
+    if isinstance(b, Function):
+        v = TestFunction(b.function_space())
+        return inner(b, v)*dx
+    elif isinstance(b, Cofunction):
+        v = TestFunction(b.function_space().dual())
+        return inner(b.riesz_representation(), v)*dx
```
pbrubeck, line 23: How about?
```suggestion
        return b
```

PR #4444, opened 2025-07-15
Change master to main (1)

### LINKED ISSUE #4163: Rename master branches to main

connorjward:
Goes with #4445 

Fixes #4163


francesco-ballarin:
May I suggest doing the same change at (roughly) the same time across all dependencies repo, e.g. fiat and libsupermesh?

connorjward:
> May I suggest doing the same change at (roughly) the same time across all dependencies repo, e.g. fiat and libsupermesh?

Yeah that makes sense. I would rather not try to synchronise all the changes because that will probably just overcomplicate things.

PR #4458, opened 2025-07-23
Fix CI timeouts 🤞

connorjward:
After some exhausting debugging I think I identified the source of the latest hangs.

The issue was introduced in https://github.com/firedrakeproject/firedrake/pull/4391. It turns out that `mpiexec -n 1 pytest ...` can hang, whereas `mpiexec -n 2 pytest ...` doesn't! To find this I had to SSH into the runner and `gdb` into the hanging process where it was spinning in `ompi_finalize`. The fix is therefore just to call `pytest` instead of `mpiexec -n 1 pytest`.

I think that this took so long to find because (a) the error is stochastic, and (b) we have also been getting timeouts due to thermal throttling/oversubscription. I've therefore also increased the timeouts for some steps so we shouldn't mix up slowdowns with hanging.

### DIFF
--- a/.github/workflows/core.yml
+++ b/.github/workflows/core.yml
@@ -233,7 +233,7 @@ jobs:
           : # Use pytest-xdist here so we can have a single collated output (not possible
           : # for parallel tests)
           firedrake-run-split-tests 1 1 -n 8 "$EXTRA_PYTEST_ARGS" firedrake-repo/tests/tsfc
-        timeout-minutes: 60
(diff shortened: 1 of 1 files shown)


JHopeCollins (approved):
Ok. I still think it's bizarre that this is only happening with the serial tests, but seeing as the only failing tests are the hang on nprocs=2 and the stochastic stokes convergence failure we've been seeing, lets get this merged and see if it helps.

PR #4460, opened 2025-07-24
Do not set different cache directories per job

connorjward:
This should not be necessary as the disk caches are now safer in how they handle race conditions. Using a unified cache directory should improve performance.


JHopeCollins:
1. What were the changes to the caching?
2. Do those changes improve the behaviour for race conditions between different firedrake instances, not just between different ranks in the same firedrake instance?

PR #4463, opened 2025-07-24
Dependency groups

connorjward:
Dependency groups are a new feature of pyproject.tomls perfect for describing internal dependency requirements.

See https://packaging.python.org/en/latest/specifications/dependency-groups/


dham (dismissed):

connorjward:
@JHopeCollins should we just revert this commit? Turns out that running `pip install -U pip` in a Dockerfile is forbidden (as `pip` is installed through apt).

I'll need to put this into `main` too.

JHopeCollins:
Unfortunately I was coming to a similar conclusion. Unless we can install a more up to date pip then I think we'll have to. Obviously the one on apt is too out of date, so as far as I can see that leaves two alternatives for updating pip:

1. Go back to a venv and use a local pip. I don't like this one because having a system install of Firedrake makes using the containers much cleaner.
2. Build an up to date pip from scratch. A bit annoying, but not the end of the world. The big downside is moving the Docker/CI build process further from the usual user process.

I would argue against 1, but I'm undecided about 2 yet. Thoughts?

PR #4464, opened 2025-07-24
Amend .test_durations

connorjward:
In particular we now use the right path for the tests (`tests/firedrake/...` instead of `tests/regression` etc) so they will actually get used!


Makefile
```diff
@@ -87,13 +87,3 @@ clean:
 	-@rm -f pyop2/*.so > /dev/null 2>&1
 	@echo "    RM tinyasm/*.so"
 	-@rm -f tinyasm/*.so > /dev/null 2>&1
-
-.PHONY: durations
-durations:
```
JHopeCollins, line 92: Why get rid of this? Surely it's easier for someone to run `make durations` than to copy out that whole `mpiexec` command?

JHopeCollins (changes requested):

Makefile
connorjward: It wasn't robust. I was getting some test failures locally because I didn't have the exact CI setup installed. Now I think about it I guess a better solution here is to run this command in the Docker image...

Converting to draft...

JHopeCollins: If `make durations` works in the container then I think it's fine to leave it. Just modify the instructions in the README saying it may not be robust locally so to run it in the container.

Really, I don't think we should be using durations generated locally on the CI because we've no idea how similar the setups are. The durations will be indicative, which is still useful, but with no guarantees on _how_ indicative.
I think "proper" thing to do would be to generate them on CI with a scheduled kron job.
But that's for another PR, I can look at that once I've updated the dockerfiles. The important thing now is updating the durations to something more recent, and making sure we actually use them at all.

Makefile
```diff
@@ -88,12 +88,17 @@ clean:
...
-	@echo "    Generate timings to optimise pytest-split"
-	python -m pytest --store-durations -m "parallel[1] or not parallel" tests/
+# NOTE: It is recommended to run this command from inside the 'firedrake'
+# Docker image to reduce the likelihood of test failures.
+.PHONY: test_durations
+test_durations:
+	@echo "    Regenerating test durations"
+	@echo "    Removing old durations file"
+	#rm -f tests/test_durations.json
+	#-python3 -m pytest --store-durations --durations-path=tests/test_durations.json -m parallel[1] tests/
```
connorjward:
```suggestion
	rm -f tests/test_durations.json
	python3 -m pytest --store-durations --durations-path=tests/test_durations.json -m parallel[1] tests/ || true
```

Makefile
```diff
@@ -88,12 +88,17 @@ clean:
...
 	# use ':' to ensure that only rank 0 writes to the durations file
-	for nprocs in 2 3 4 6 7 8; do \
-		mpiexec -n 1 python -m pytest --store-durations -m parallel[$${nprocs}] tests/ : \
-			-n $$(( $${nprocs} - 1 )) pytest -m parallel[$${nprocs}] -q tests/ ; \
+	-for nprocs in 2 3 4 6 7 8; do \
+		mpiexec -n 1 python3 -m pytest --store-durations --durations-path=tests/test_durations.json -m parallel[$${nprocs}] tests/ \
+		: -n $$(( $${nprocs} - 1 )) python3 -m pytest -m parallel[$${nprocs}] -q tests/ || true ; \
```
connorjward:
```suggestion
	for nprocs in 2 3 4 6 7 8; do \
		mpiexec -n 1 python3 -m pytest --store-durations --durations-path=tests/test_durations.json -m parallel[$${nprocs}] tests/ \
		: -n $$(( $${nprocs} - 1 )) python3 -m pytest -m parallel[$${nprocs}] -q tests/ || true ; \
```

Makefile
connorjward: I've added it back in and regenerated the timings again on my workstation inside the container. Seemed to work fine.

JHopeCollins (approved):
The latest actions might be on a particularly enthusiastic runner, but this seems to have made a decent difference to how well load balanced the CI is!

PR #4465, opened 2025-07-24
Avoid mpiexec -n 1 for firedrake-check

connorjward:
Missed this one

For example see https://github.com/firedrakeproject/firedrake/actions/runs/16498790363/job/46651297370?pr=4463

### DIFF
--- a/scripts/firedrake-check
+++ b/scripts/firedrake-check
@@ -78,7 +78,11 @@ def run_tests(tests: tuple[str], nprocs: int, mpiexec: str) -> None:
     else:
         check_flags = "--quiet -p no:cacheprovider"
 
-    cmd = f"{mpiexec} {nprocs} {sys.executable} -m pytest {check_flags} -m parallel[{nprocs}]".split()
+    # Do not use mpiexec -n 1 as this can hang
+    if nprocs == 1:
+        cmd = f"{sys.executable} -m pytest {check_flags} -m parallel[1]".split()
+    else:
+        cmd = f"{mpiexec} {nprocs} {sys.executable} -m pytest {check_flags} -m parallel[{nprocs}]".split()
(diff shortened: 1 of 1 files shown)


scripts/firedrake-check
```diff
@@ -78,7 +78,11 @@ def run_tests(tests: tuple[str], nprocs: int, mpiexec: str) -> None:
     else:
         check_flags = "--quiet -p no:cacheprovider"
 
-    cmd = f"{mpiexec} {nprocs} {sys.executable} -m pytest {check_flags} -m parallel[{nprocs}]".split()
+    # Do not use mpiexec -n 1 as this can hang
+    if nprocs == 1:
+        cmd = f"{sys.executable} -m pytest {check_flags} -m parallel[1]".split()
+    else:
+        cmd = f"{mpiexec} {nprocs} {sys.executable} -m pytest {check_flags} -m parallel[{nprocs}]".split()
```
JHopeCollins: 1. Why not use `firedrake-run-split-tests` here?
2. If that's not possible, then make it obvious that the only difference is whether `mpiexec` is used or not.

```suggestion
    cmd = f"{sys.executable} -m pytest {check_flags} -m parallel[{nprocs}]".split()                                         
    # Only use mpiexec for parallel tests because mpiexec -n 1 can hang
    if nprocs > 1:
        mpi_cmd = f"{mpiexec} {nprocs}".split()
        cmd = mpi_cmd + cmd
```

connorjward: > Why not use firedrake-run-split-tests here?

That will do things like produce log files which we don't want `firedrake-check` to do.

I will make a similar amendment to what you suggest.

connorjward: Done.

PR #4466, opened 2025-07-25
Try to fix firedrake-check on macOS

connorjward:
# Description

### DIFF
--- a/.github/workflows/core.yml
+++ b/.github/workflows/core.yml
@@ -517,10 +517,11 @@ jobs:
           firedrake-check
         timeout-minutes: 10
 
-      - name: Post-run cleanup
-        if: always()
-        run: |
-          find . -delete
+      # UNDO ME
+      # - name: Post-run cleanup
+      #   if: always()
+      #   run: |
+      #     find . -delete
 
   lint:
     name: Lint codebase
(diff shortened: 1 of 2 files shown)


.github/workflows/core.yml
```diff
@@ -517,10 +517,11 @@ jobs:
...
 
-      - name: Post-run cleanup
-        if: always()
-        run: |
-          find . -delete
+      # UNDO ME
+      # - name: Post-run cleanup
+      #   if: always()
+      #   run: |
+      #     find . -delete
```
```suggestion
      - name: Post-run cleanup
        if: always()
        run: |
          find . -delete
```

JHopeCollins (commented):
Thanks for getting on this!

How come we now need a specialised command class for bdist as well as the editable and sdist ones?

connorjward:
> Thanks for getting on this!
> 
> How come we now need a specialised command class for bdist as well as the editable and sdist ones?

It's all a bit convoluted. In our Linux CI we create an sdist and install it to make sure that sdists work properly (since that's what we publish on PyPI). However when we install on macOS we don't do this and instead directly build the wheel. That meant that the `sdist` step wasn't happening.

JHopeCollins (approved):
Ok, that makes sense. LGTM then.

PR #4469, opened 2025-07-26
Fix broken Stokes test

pbrubeck:
# Description
We were getting a failure on a simple Stokes test involving MUMPS + nullspace. 

`tests/firedrake/macro/test_macro_solve.py::test_stokes[square-th]  - assert np.float64(-8.292409965245584) >= 2.9`

This is an attempt to fix it.

### DIFF
--- a/tests/firedrake/macro/test_macro_solve.py
+++ b/tests/firedrake/macro/test_macro_solve.py
@@ -144,7 +144,7 @@ def test_stokes(mh, variant, mixed_element):
         Z = V * Q
 
         a, L = stokes_mms(Z, as_vector(zexact))
-        bcs = DirichletBC(Z[0], as_vector(zexact[:dim]), "on_boundary")
+        bcs = DirichletBC(Z.sub(0), as_vector(zexact[:dim]), "on_boundary")
 
         zh = Function(Z)
         nullspace = MixedVectorSpaceBasis(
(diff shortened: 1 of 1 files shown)


tests/firedrake/macro/test_macro_solve.py
```diff
@@ -176,7 +176,7 @@ def test_stokes(mh, variant, mixed_element):
         # Test div-free
         assert np.allclose(div_err, 0, atol=1E-10)
     elif variant == "th":
-        assert conv_rates(u_err, h)[-1] >= 2.9
+        assert conv_rates(u_err, h)[-1] >= 2.9, (u_err, p_err)
```
pbrubeck, line 179:
```suggestion
        assert conv_rates(u_err, h)[-1] >= 2.9
```

pbrubeck:
```suggestion
        assert conv_rates(u_err, h)[-1] >= 2.9
```

JHopeCollins:
Thanks for looking into this @pbrubeck!

This test hasn't been failing consistently, so if the CI passes then please could you rerun it a couple of times just to make sure it has fixed the problem?

JHopeCollins:
> It's best to rerun it by pushing empty commits so all previous runs are still available.

Ok, that'll be easier to see at a glance (although there is a drop down on each action for looking at previous attempts if you need to).

connorjward (approved):
I'm seeing enough ticks to be reasonably confident that this has done the trick. Even if not the changes here won't do any harm.

Thanks @pbrubeck for fixing this! I wouldn't have had the first idea where to start.

PR #4482, opened 2025-08-06
interpolate: sanitise submesh interpolate

ksagiyam:
Let one pass `allow_missing_dofs=True` to let Firedrake automatically make appropriate `subset` when interpolating between submeshes.


ksagiyam:
@connorjward I wanted to refactor a little and add stricter checks. Can you check the diff?

Issue #4398, opened 2025-06-24
Support for Function.assign on Submesh


pbrubeck:
**Describe the current issue**
Currently `Function.assign` is not implemented across Submesh and its parent.

**Describe the solution you'd like**
DOF-wise assigningment of Functions with the same UFL element but on `Submeshes` should be trivial to implement by indexing.

MFE:
```
from firedrake import *
mesh = UnitSquareMesh(2, 2)
M = FunctionSpace(mesh, "DG", 0)
m = Function(M)
m.dat.data[0] = 1

rmesh = RelabeledMesh(mesh, [m], [100])
submesh = Submesh(rmesh, rmesh.geometric_dimension(), 100)

V = FunctionSpace(rmesh, "DG", 0)
Vsub = V.reconstruct(mesh=submesh)
u = Function(V)
usub = Function(Vsub)
usub.assign(1)

u.assign(usub, subset=rmesh.cell_subset(100))
print(u.dat.data)
```

Error message:
```
Traceback (most recent call last):
  File "/home/pbrubeck/git/dphil_thesis/static_condensation/snippets/bug_submesh_assign.py", line 16, in <module>
    u.assign(usub, subset=rmesh.cell_subset(100))
  File "petsc4py/PETSc/Log.pyx", line 188, in petsc4py.PETSc.Log.EventDecorator.decorator.wrapped_func
  File "petsc4py/PETSc/Log.pyx", line 189, in petsc4py.PETSc.Log.EventDecorator.decorator.wrapped_func
  File "/home/pbrubeck/firedrake/src/firedrake/firedrake/adjoint_utils/function.py", line 116, in wrapper
    ret = assign(self, other, *args, **kwargs)
  File "/home/pbrubeck/firedrake/src/firedrake/firedrake/function.py", line 461, in assign
    Assigner(self, expr, subset).assign()
  File "/home/pbrubeck/firedrake/src/firedrake/firedrake/assign.py", line 155, in __init__
    raise ValueError("All functions in the expression must use the same "
ValueError: All functions in the expression must use the same mesh as the assignee

```

**Describe alternatives you've considered**
`Function.interpolate` works with `Submesh`, but does redundant work by evaluating the degrees of freedom of the element on itself. This of course will fail for elements where dual evaluation is not implemented.

**Additional info**
Add any other information here.

ksagiyam:
Assignment between submesh and parent turned out to be a lot more involved than conventional assignment due to potential ownership transfer of entities.

[...]

PR #4489, opened 2025-08-11
Ksagiyam/support submesh assign

ksagiyam:
Resolve https://github.com/firedrakeproject/firedrake/issues/4398.

Support `f.assign()` on submeshes.

Fix `mixed_func.assign(..., subset=...)`.

### DIFF
--- a/firedrake/assign.py
+++ b/firedrake/assign.py
@@ -3,6 +3,7 @@ import operator
 
 import numpy as np
 from pyadjoint.tape import annotate_tape
+from pyop2 import op2
 from pyop2.utils import cached_property
 import pytools
 import finat.ufl
@@ -12,19 +13,22 @@ from ufl.corealg.map_dag import map_expr_dag
 from ufl.corealg.multifunction import MultiFunction
 from ufl.domain import extract_unique_domain
 
+from firedrake.cofunction import Cofunction
 from firedrake.constant import Constant
 from firedrake.function import Function
 from firedrake.petsc import PETSc
 from firedrake.utils import ScalarType, split_by
 
+from mpi4py import MPI
+
 
 def _isconstant(expr):
     return isinstance(expr, Constant) or \
-        (isinstance(expr, Function) and expr.ufl_element().family() == "Real")
+        (isinstance(expr, (Function, Cofunction)) and expr.ufl_element().family() == "Real")
 
 
 def _isfunction(expr):
-    return isinstance(expr, Function) and expr.ufl_element().family() != "Real"
+    return isinstance(expr, (Function, Cofunction)) and expr.ufl_element().family() != "Real"
 
 
 class CoefficientCollector(MultiFunction):
@@ -99,6 +103,9 @@ class CoefficientCollector(MultiFunction):
     def coefficient(self, o):
         return ((o, 1),)
 
+    def cofunction(self, o):
+        return ((o, 1),)
+
     def constant_value(self, o):
         return ((o, 1),)
 
@@ -130,11 +137,17 @@ class CoefficientCollector(MultiFunction):
 
 
 class Assigner:
-    """Class performing pointwise assignment of an expression to a :class:`firedrake.function.Function`.
+    """Class performing pointwise assignment of an expression to a function or a cofunction.
 
-    :param assignee: The :class:`~.firedrake.function.Function` being assigned to.
-    :param expression: The :class:`ufl.core.expr.Expr` to evaluate.
-    :param subset: Optional subset (:class:`pyop2.types.set.Subset`) to apply the assignment over.
+    Parameters
+    ----------
+    assignee : firedrake.function.Function or firedrake.cofunction.Cofunction
+        Function or Cofunction being assigned to.
+    expression : ufl.core.expr.Expr or ufl.form.BaseForm
+        Expression to be assigned.
+    subset : pyop2.types.set.Set or pyop2.types.set.Subset or pyop2.types.set.MixedSet
+        Subset to apply the assignment over.
+
     """
     symbol = "="
 
@@ -142,22 +155,38 @@ class Assigner:
 
     def __init__(self, assignee, expression, subset=None):
         expression = as_ufl(expression)
-
+        source_meshes = set()
         for coeff in extract_coefficients(expression):
-            if isinstance(coeff, Function) and coeff.ufl_element().family() != "Real":
+            if isinstance(coeff, (Function, Cofunction)) and coeff.ufl_element().family() != "Real":
                 if coeff.ufl_element() != assignee.ufl_element():
                     raise ValueError("All functions in the expression must have the same "
                                      "element as the assignee")
-                if extract_unique_domain(coeff) != extract_unique_domain(assignee):
-                    raise ValueError("All functions in the expression must use the same "
-                                     "mesh as the assignee")
-
-        if (subset and type(assignee.ufl_element()) == finat.ufl.MixedElement
-                and any(el.family() == "Real"
-                        for el in assignee.ufl_element().sub_elements)):
-            raise ValueError("Subset is not a valid argument for assigning to a mixed "
-                             "element including a real element")
-
+                source_meshes.add(extract_unique_domain(coeff))
+        if len(source_meshes) == 0:
+            pass
+        elif len(source_meshes) == 1:
+            target_mesh = extract_unique_domain(assignee)
+            source_mesh, = source_meshes
+            if target_mesh.submesh_youngest_common_ancester(source_mesh) is None:
+                raise ValueError(
+                    "All functions in the expression must be defined on a single domain "
+                    "that is in the same submesh family as domain of the assignee"
+                )
+        else:
+            raise ValueError(
+                "All functions in the expression must be defined on a single domain"
+            )
+        if subset is None:
+            subset = tuple(None for _ in assignee.function_space())
+        if len(subset) != len(assignee.function_space()):
+            raise ValueError(f"Provided subset ({subset}) incompatible with assignee ({assignee})")
+        if type(assignee.ufl_element()) == finat.ufl.MixedElement:
+            for subs, el in zip(subset, assignee.function_space().ufl_element().sub_elements):
+                if subs is not None and el.family() == "Real":
+                    raise ValueError(
+                        "Subset is not a valid argument for assigning to a mixed "
+                        "element including a real element"
+                    )
         self._assignee = assignee
         self._expression = expression
         self._subset = subset
@@ -169,14 +198,21 @@ class Assigner:
         return f"{self.__class__.__name__}({self._assignee!r}, {self._expression!r})"
 
     @PETSc.Log.EventDecorator()
-    def assign(self):
-        """Perform the assignment."""
+    def assign(self, allow_missing_dofs=False):
+        """Perform the assignment.
+
+        Parameters
+        ----------
+        allow_missing_dofs : bool
+            Permit assignment between objects with mismatching nodes. If `True` then
+            assignee nodes with no matching assigner nodes are ignored.
+
+        """
         if annotate_tape():
             raise NotImplementedError(
                 "Taping with explicit Assigner objects is not supported yet. "
                 "Use Function.assign instead."
             )
-
         # To minimize communication during assignment we perform a number of tricks:
         # * If we are not assigning to a subset then we can always write to the
         #   halo. The validity of the original assignee dat halo does not matter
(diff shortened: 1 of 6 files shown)


firedrake/assign.py
```diff
@@ -172,14 +191,20 @@ def __repr__(self):
...
     @PETSc.Log.EventDecorator()
-    def assign(self):
-        """Perform the assignment."""
+    def assign(self, allow_missing_dofs=False):
+        """Perform the assignment.
+
+        Parameters
+        ----------
+        allow_missing_dofs : bool
+            If True, ignore assignee nodes with no matching assigner nodes.
```
connorjward:
```suggestion
            Permit assignment between objects with mismatching nodes. If `True` then
            assignee nodes with no matching assigner nodes are ignored.
```

firedrake/assign.py
```diff
@@ -194,28 +219,78 @@ def assign(self):
...
+            if subset is not None:
+                if subset is target_V.node_set:
+                    # The whole set.
+                    subset = None
+                elif subset.superset is target_V.node_set:
+                    # op2.Subset of target_V.node_set
+                    pass
+                else:
+                    raise ValueError(f"subset ({subset}) not a subset of target_V.node_set ({target_V.node_set})")
+            source_meshes = set()
```
connorjward:
```suggestion
                    raise ValueError(f"subset ({subset}) not a subset of target_V.node_set ({target_V.node_set})")

            source_meshes = set()
```
readability

firedrake/assign.py
```diff
@@ -194,28 +219,78 @@ def assign(self):
...
+                pass
+            elif len(source_meshes) == 1:
+                source_mesh, = source_meshes
+                if target_mesh is source_mesh:
+                    pass
+                else:
+                    single_mesh_assign = False
+            else:
+                raise ValueError("All functions in the expression must be defined on a single domain")
+            if single_mesh_assign:
```
connorjward, line 258:
```suggestion
            else:
                raise ValueError("All functions in the expression must be defined on a single domain")

            if single_mesh_assign:
```
readability

firedrake/assign.py
```diff
@@ -194,28 +219,78 @@ def assign(self):
...
+            else:
+                source_V, = set(f.function_space() for f in funcs)
+                composed_map = source_V.topological.entity_node_map(target_mesh.topology, "cell", "everywhere", None)
+                indices_active = composed_map.indices_active_with_halo
+                indices_active_all = indices_active.all()
+                indices_active_all = target_mesh.comm.allreduce(indices_active_all, op=MPI.LAND)
+                if subset is None:
+                    if not indices_active_all and not allow_missing_dofs:
+                        raise ValueError("Found assignee nodes with no matching assigner nodes: run with `allow_missing_dofs=True`")
+                    subset_indices_target = target_V.cell_node_map().values_with_halo[indices_active, :].reshape(-1)
```
connorjward:
```suggestion
                    subset_indices_target = target_V.cell_node_map().values_with_halo[indices_active, :].flatten()
```

firedrake/assign.py
```diff
@@ -194,28 +219,78 @@ def assign(self):
...
+                    subset_indices_target, perm, _ = np.intersect1d(
+                        target_V.cell_node_map().values_with_halo[indices_active, :].reshape(-1),
+                        subset.indices,
+                        return_indices=True,
+                    )
+                    if len(subset.indices) > len(subset_indices_target) and not allow_missing_dofs:
+                        raise ValueError("Found assignee nodes with no matching assigner nodes: run with `allow_missing_dofs=True`")
+                    subset_indices_source = composed_map.values_with_halo[indices_active, :].reshape(-1)[perm]
+                # TODO: Use work array?
+                buffer = Function(target_V)
```
connorjward: Can there be a comment explaining why this is necessary?

connorjward:
```suggestion
                    subset_indices_source = composed_map.values_with_halo[indices_active, :].reshape(-1)[perm]

                # TODO: Use work array?
                buffer = Function(target_V)
```

firedrake/assign.py
```diff
@@ -191,28 +220,78 @@ def assign(self):
...
+                    subset_indices_target, perm, _ = np.intersect1d(
+                        target_V.cell_node_map().values_with_halo[indices_active, :].reshape(-1),
+                        subset.indices,
+                        return_indices=True,
+                    )
+                    if len(subset.indices) > len(subset_indices_target) and not allow_missing_dofs:
+                        raise ValueError("Found assignee nodes with no matching assigner nodes: run with `allow_missing_dofs=True`")
+                    subset_indices_source = composed_map.values_with_halo[indices_active, :].reshape(-1)[perm]
+                # TODO: Use work array?
+                buffer = type(lhs_func)(target_V)
```
pbrubeck: `Function(V.dual())` gives you a `Cofunction(V.dual())`
```suggestion
                buffer = Function(target_V)
```

firedrake/assign.py
```diff
@@ -191,28 +220,78 @@ def assign(self):
...
+            single_mesh_assign = True
+            if len(source_meshes) == 0:
+                pass
+            elif len(source_meshes) == 1:
+                source_mesh, = source_meshes
+                if target_mesh is source_mesh:
+                    pass
+                else:
+                    single_mesh_assign = False
+            else:
```
pbrubeck, line 256:
```suggestion
            elif len(source_meshes) != 0:
```

firedrake/assign.py
```diff
@@ -191,28 +220,78 @@ def assign(self):
...
+                    pass
+                else:
+                    raise ValueError(f"subset ({subset}) not a subset of target_V.node_set ({target_V.node_set})")
+            source_meshes = set()
+            for f in funcs:
+                source_meshes.add(extract_unique_domain(f))
+            single_mesh_assign = True
+            if len(source_meshes) == 0:
+                pass
+            elif len(source_meshes) == 1:
```
pbrubeck, line 248:
```suggestion
            if len(source_meshes) == 1:
```

firedrake/assign.py
```diff
@@ -191,28 +220,78 @@ def assign(self):
...
+                    # The whole set.
+                    subset = None
+                elif subset.superset is target_V.node_set:
+                    # op2.Subset of target_V.node_set
+                    pass
+                else:
+                    raise ValueError(f"subset ({subset}) not a subset of target_V.node_set ({target_V.node_set})")
+            source_meshes = set()
+            for f in funcs:
+                source_meshes.add(extract_unique_domain(f))
```
pbrubeck, line 238:
```suggestion
            source_meshes = set(extract_unique_domain(f) for f in funcs)
```

firedrake/assign.py
ksagiyam: I think this is a matter of taste. I consistently don't add blank lines unless lint complains.

firedrake/assign.py
ksagiyam: I don't think this is necessarily better.

firedrake/assign.py
ksagiyam: I think it's harder to understand. Each block has a meaning. Added some comments.

firedrake/assign.py
connorjward: From a personal perspective it makes things dramatically more readable. PEP8 says:
> Use blank lines in functions, sparingly, to indicate logical sections.

ksagiyam: What is regarded as a logical section is often subjective. I don't see a logical separation here, for instance.

I think we should just leave this to whoever writes the code unless we come up with an objective rule for blank lines.

firedrake/assign.py
ksagiyam: Done.

firedrake/assign.py
JHopeCollins: The issue with leaving it entirely to the writer is that code is almost always clearer to the person writing it.
Style-for-readability is not for the writer at the time of writing, it's for anyone else reading the code.

Personally, having almost 60 lines of code without a single white space is nigh unreadable. If I opened this file to read then I would be manually inserting temporary blank lines just to help myself understand.

I don't think its unreasonable here to consider each if-block to be a small logical component. They are either parsing a particular type of info from the arguments, or selecting between different implementations depending on the parsed arguments.

I'll add some suggestions for spreading the code out a bit. Hopefully you'll be able to see why it's easier for us to parse, even if you're happy with the current style.

firedrake/assign.py
```diff
@@ -191,28 +227,102 @@ def assign(self):
...
+            target_V = lhs_func.function_space()
+            if subset is not None:
+                if subset is target_V.node_set:
+                    # The whole set.
+                    subset = None
+                elif subset.superset is target_V.node_set:
+                    # op2.Subset of target_V.node_set
+                    pass
+                else:
+                    raise ValueError(f"subset ({subset}) not a subset of target_V.node_set ({target_V.node_set})")
```
JHopeCollins, line 243: # Validate the subset
```suggestion
            # Validate the subset
            if subset is not None:
                if subset is target_V.node_set:
                    # The whole set.
                    subset = None
                elif subset.superset is target_V.node_set:
                    # op2.Subset of target_V.node_set
                    pass
                else:
                    raise ValueError(f"subset ({subset}) not a subset of target_V.node_set ({target_V.node_set})")

```

firedrake/assign.py
JHopeCollins: Agreed. The if-block above is just deciding whether we are doing cross-mesh assignment or not.

firedrake/assign.py
```diff
@@ -191,28 +227,102 @@ def assign(self):
...
+                self._assign_single_dat(lhs_func.dat, subset_indices, rvalue, assign_to_halos)
+                if assign_to_halos:
+                    lhs_func.dat.halo_valid = True
+            else:
+                source_V, = set(f.function_space() for f in funcs)
+                composed_map = source_V.topological.entity_node_map(target_mesh.topology, "cell", "everywhere", None)
+                indices_active = composed_map.indices_active_with_halo
+                indices_active_all = indices_active.all()
+                indices_active_all = target_mesh.comm.allreduce(indices_active_all, op=MPI.LAND)
+                if subset is None:
```
JHopeCollins: Please add a comment stating what this if-block is deciding.

firedrake/assign.py
```diff
@@ -191,28 +227,102 @@ def assign(self):
...
+                    data_ro = operator.attrgetter("data_ro_with_halos")
+                else:
+                    subset_indices = ... if subset is None else subset.owned_indices
+                    data_ro = operator.attrgetter("data_ro")
+                func_data = np.array([data_ro(f.dat)[subset_indices] for f in funcs])
+                rvalue = self._compute_rvalue(func_data)
+                self._assign_single_dat(lhs_func.dat, subset_indices, rvalue, assign_to_halos)
+                if assign_to_halos:
+                    lhs_func.dat.halo_valid = True
+            else:
```
JHopeCollins, line 260: Personally I would add a blank line here because the two code blocks either side are quite large, but I concede that this is arguably more aggressive.
```suggestion

            else:
```

firedrake/assign.py
JHopeCollins: Unless `lhs_func` is ever not a `Function`/`Cofunction` then I think explicitly stating the type is clearer. Usually `type(obj)(args)` is reserved for places where the type really is unknown, so having it here incorrectly implies that.

firedrake/assign.py
```diff
@@ -191,28 +227,102 @@ def assign(self):
...
+                    subset_indices_target, perm, _ = np.intersect1d(
+                        target_V.cell_node_map().values_with_halo[indices_active, :].flatten(),
+                        subset.indices,
+                        return_indices=True,
+                    )
+                    if len(subset.indices) > len(subset_indices_target) and not allow_missing_dofs:
+                        raise ValueError("Found assignee nodes with no matching assigner nodes: run with `allow_missing_dofs=True`")
+                    subset_indices_source = composed_map.values_with_halo[indices_active, :].flatten()[perm]
+                # Use buffer array to make sure that owned DoFs are updated upon assigning.
+                # The following example illustrate the issue that a naive assignment would cause.
```
JHopeCollins:
```suggestion
                # The following example illustrates the issue that a naive assignment would cause.
```

firedrake/assign.py
ksagiyam: The code is for everyone, but none of us represents everyone. Those who need to work on their laptop might want to see more lines of codes without scrolling.

You added some reasoning, but those still do not define an objective rule that can be universally applied. Unless we can come up with an objective rule and enforce it, I think we should just respect the authors' style, as they are the ones who have spent most time and effort in the PR.

firedrake/assign.py
ksagiyam: `type(obj)` is used to support child class instances.

ksagiyam:
@AnuragRao1 I don't think I made any material change, but could you check if this branch still works for you?

firedrake/assign.py
ksagiyam: Done.

firedrake/assign.py
connorjward: I think that this discussion is making it clear that we do need to formalise some sort of style guide that clarifies this. It will always be somewhat subjective but it would definitely help in situations like this. It's also important that we have a process for determining what to do when reviewer and developer disagree.

I will draft something and bring it to the next meeting (that David can attend) for wider discussion.

firedrake/assign.py
JHopeCollins: Yes, that is a common example of an unknown type. Does it apply here?

ksagiyam: This is more consistent with that we use `isinstance(o, Function)` checks rather than `type(o) is Funciton` in the above. Downstream packages/users/Future Firedrake can also in theory subclass `Function`/`Cofunction`, though it might not entirely work currently.

ksagiyam:
This is not a bug, so it probably shouldn't go to release.

PR #4499, opened 2025-08-15
Expunge Vector

connorjward:
This has been deprecated for a very long time.


pyop2/types/dat.py
```diff
@@ -242,6 +242,18 @@ def data_wo_with_halos(self):
...
+    @mpi.collective
+    def global_data(self):
+        with self.vec_ro as vec:
+            v = PETSc.Vec().createSeq(vec.size, comm=mpi.COMM_SELF)
+            is_ = PETSc.IS().createStride(vec.size, 0, 1, comm=mpi.COMM_SELF)
+
+            vscat = PETSc.Scatter().create(vec, is_, v, None)
+            vscat.scatterBegin(vec, v, addv=PETSc.InsertMode.INSERT_VALUES)
+            vscat.scatterEnd(vec, v, addv=PETSc.InsertMode.INSERT_VALUES)
+        return v.array
```
JHopeCollins: This can be written a little more simply, see:
https://github.com/firedrakeproject/firedrake/blob/3f86c911d69e46e8b2607a4c18e0674c68189024/firedrake/adjoint_utils/ensemble_function.py#L33-L39

connorjward: Awesome thank you. I just copied what was there.

connorjward: I've done this but I think there is a bug here. `toAll()` returns a tuple so calling `.scatter` is undefined. https://petsc.org/main/petsc4py/reference/petsc4py.PETSc.Scatter.html#petsc4py.PETSc.Scatter.toAll

JHopeCollins: ~API must have changed. Looks like it now returns both the scatter and an appropriate Vec, so you could probably do something like this:~ Never mind, you already did!
```python
scatter, lvec = PETSc.Scatter().toall(gvec)
scatter.scatter(gvec, lvec, addv=PETSc.InsertMode.INSERT_VALUES) 
```

.github/workflows/core.yml
```diff
@@ -336,7 +336,8 @@ jobs:
           matrix.arch == 'default'
         run: |
           . venv/bin/activate
-          git clone --depth 1 https://github.com/thetisproject/thetis.git thetis-repo
+          : # UNDO ME
+          git clone --depth 1 https://github.com/thetisproject/thetis.git thetis-repo --branch connorjward/novector
```
```suggestion
          git clone --depth 1 https://github.com/thetisproject/thetis.git thetis-repo
```

pyop2/types/dat.py
```diff
@@ -242,6 +242,16 @@ def data_wo_with_halos(self):
...
 
+    @property
+    @mpi.collective
+    def global_data(self):
+        """Return all the data for the Dat gathered onto individual ranks."""
+        with self.vec_ro as gvec:
+            scatter, lvec = PETSc.Scatter().toAll(gvec)
+            scatter.scatter(
+                gvec, lvec, addv=PETSc.InsertMode.INSERT_VALUES)
+        return lvec.array_r
```
JHopeCollins: Does this need to be the read-only accessor?

connorjward: Fair point. Fixed.

demos/linear_fluid_structure_interaction/linear_fluid_structure_interaction.py.rst
```diff
@@ -287,11 +287,11 @@ To save data for visualization, we change the position of the nodes in the mesh,
         output_data.counter += 1
         if output_data.counter % output_data_every_x_time_steps != 0:
             return
-        mesh_static = mesh.coordinates.vector().get_local()
-        mesh.coordinates.vector().set_local(mesh_static + X.vector().get_local())
-        mesh.coordinates.dat.data[:, 1] += eta.dat.data_ro
+        mesh_static = mesh.coordinates.copy(deepcopy=True)
+        mesh.coordinates += X
+        mesh.coordinates.dat.data_wo[:, 1] += eta.dat.data_ro
```
JHopeCollins: Is `dat.data_wo` correct if you are _incrementing_ the values rather than _overwriting_ them? Should it not be `dat.data`?

demos/saddle_point_pc/saddle_point_systems.py.rst
```diff
@@ -93,13 +93,11 @@ discontinuous space. ::
...
 
-To initialise this function to a random value we access its :class:`~.Vector`
-form and use numpy_ to set the values::
+We initialise this function to hold random values::
 
     #
         import numpy as np
-        fvector = f.vector()
-        fvector.set_local(np.random.uniform(size=fvector.local_size()))
+        f.dat.data_wo[...] = np.random.uniform(size=f.dof_dset.size)
```
JHopeCollins: Would `f.dat.data_wo[...] = np.random.uniform(size=f.dat.data_wo.shape)` not be the obvious pattern?

firedrake/adjoint_utils/function.py
```diff
@@ -315,25 +308,23 @@ def _ad_function_space(self, mesh):
...
     def _applyBinary(self, f, y):
-        vec = self.vector()
-        npdata = vec.get_local()
-        npdatay = y.vector().get_local()
+        npdata = self.dat.data_ro
+        npdatay = y.dat.data_ro
         for i in range(len(npdata)):
             npdata[i] = f(npdata[i], npdatay[i])
-        vec.set_local(npdata)
+        self.dat.data_wo[...] = npdata
```
JHopeCollins: `npdata` points to the read-only array on line 323, but is written into on line 326. Is doing the right thing?
If you just write into `self.dat.data_wo` directly then you can also remove the additional copy on line 327.

tests/firedrake/adjoint/test_assemble.py
```diff
@@ -99,7 +95,7 @@ def test_assemble_1_forms_tlm():
 
     Jhat = ReducedFunctional(J, Control(f))
     h = Function(V)
-    h.vector()[:] = rand(h.dof_dset.size)
+    h.dat.data_wo[...] = rand(h.dof_dset.size)
```
JHopeCollins: We should be consistent with how we are querying the array size for `dat.data`.

1. `dat.data.shape`?
2. `h.dof_dset.size`?
3. `V.dim()`?

connorjward: Sure. Note that the last of these isn't safe because that gives the global size.

JHopeCollins: Thanks. The last one is used in quite a few places so even more reason to make it consistent. I think it's only used in serial tests but we don't want people copying unsafe code.

demos/linear_fluid_structure_interaction/linear_fluid_structure_interaction.py.rst
connorjward: Good catch

firedrake/adjoint_utils/function.py
connorjward: done

tests/firedrake/adjoint/test_assemble.py
connorjward: done

demos/saddle_point_pc/saddle_point_systems.py.rst
connorjward: done

demos/linear_fluid_structure_interaction/linear_fluid_structure_interaction.py.rst
connorjward: done

JHopeCollins (approved):
LGTM. Thanks for taking this on!

tests/firedrake/adjoint/test_hessian.py
```diff
@@ -179,7 +174,7 @@ def test_nonlinear():
     Jhat = ReducedFunctional(J, Control(f))
 
     h = Function(V)
-    h.vector()[:] = 10*rng.random(V.dim())
+    h.dat.data_wo[...] = 10*rng.random(h.dat.data_wo.shape)
```
pbrubeck: We should make these tests deterministic
```suggestion
    rg = RandomGenerator(PCG64(seed=1234))
    h = rg.uniform(V, 0, 10)
```
Also the line above should be deleted

connorjward: Done

firedrake/ml/pytorch/fem_operator.py
```diff
@@ -220,7 +217,6 @@ def from_torch(x, V=None):
             val = val[0]
         return Constant(val)
     else:
-        x = x.detach().numpy()
         x_F = Function(V)
```
pbrubeck:
```suggestion
        x_F = Function(V, val=x.detach().numpy())
```

firedrake/ml/jax/fem_operator.py
```diff
@@ -225,7 +222,6 @@ def from_jax(x: "jax.Array", V: Optional[WithGeometry] = None) -> Union[Function
             val = val[0]
         return Constant(val)
     else:
-        x = np.asarray(x)
         x_F = Function(V)
```
pbrubeck:
```suggestion
        x_F = Function(V, val=np.asarray(x))
```

tests/firedrake/external_operators/test_external_operators.py
```diff
@@ -131,8 +131,7 @@ def test_assemble(V, f):
 
     # -- dNdu(u, v, w; δN, uhat) (Adjoint model) -- #
     # Define a random cofunction on V* since the adjoint model maps from V* to V*
-    delta_N = Cofunction(V.dual())
-    delta_N.vector()[:] = rg.beta(V, 15, 30).dat.data_ro[:]
+    delta_N = Cofunction(V.dual()).assign(rg.beta(V, 15, 30))
```
pbrubeck:
```suggestion
    delta_N = rg.beta(V.dual(), 15, 30)
```

firedrake/ml/jax/fem_operator.py
```diff
@@ -225,7 +222,6 @@ def from_jax(x: "jax.Array", V: Optional[WithGeometry] = None) -> Union[Function
...
-        x_F.vector().set_local(x)
+        x_F.dat.data_wo[...] = np.asarray(x)
```
pbrubeck:
```suggestion
```

firedrake/ml/pytorch/fem_operator.py
```diff
@@ -220,7 +217,6 @@ def from_torch(x, V=None):
...
-        x_F.vector().set_local(x)
+        x_F.dat.data_wo[...] = x.detach().numpy()
```
pbrubeck:
```suggestion
```

firedrake/ml/pytorch/fem_operator.py
pbrubeck: Do we want a copy here? Or can we point to the same data?

tests/firedrake/external_operators/test_external_operators.py
pbrubeck: I was not aware that we still could assign a `Function` to a `Cofunction`

firedrake/ml/pytorch/fem_operator.py
connorjward: No idea. This seems better though.

firedrake/adjoint_utils/function.py
```diff
@@ -276,22 +276,15 @@ def _ad_dot(self, other, options=None):
 
     @staticmethod
     def _ad_assign_numpy(dst, src, offset):
-        range_begin, range_end = dst.vector().local_range()
+        range_begin, range_end = dst.dat.dataset.layout_vec.getOwnershipRange()
         m_a_local = src[offset + range_begin:offset + range_end]
-        dst.vector().set_local(m_a_local)
-        dst.vector().apply('insert')
-        offset += dst.vector().size()
+        dst.dat.data_wo[...] = m_a_local.reshape(dst.dat.data_wo.shape)
```
JHopeCollins, line 281: In light of James' PR, this would unnecessarily increment the dat version.
```suggestion
        dst.dat.data_wo[...] = m_a_local.reshape(dst.dat.data_ro.shape)
```

tests/firedrake/adjoint/test_assemble.py
```diff
@@ -1,12 +1,16 @@
...
-from numpy.random import rand
 from numpy.testing import assert_allclose
 
 from firedrake import *
 from firedrake.adjoint import *
 
 
+@pytest.fixture
+def rg():
+    return RandomGenerator(PCG64(seed=1234))
```
JHopeCollins, line 11: This returns a deterministic sequence of pseudo-random numbers. However, if multiple tests call this fixture then each test will not get a deterministic subset of those numbers because the order in which the tests are run may change, for example if run with different number of splits or if more tests are added.

Usually you can just set `np.random.seed` at the beginning of each test then call `np.random.random_sample`.

tests/firedrake/adjoint/test_assignment.py
```diff
@@ -4,10 +4,14 @@
 from firedrake.adjoint import *
 from checkpoint_schedules import SingleMemoryStorageSchedule
 
-from numpy.random import rand, default_rng
 from numpy.testing import assert_approx_equal, assert_allclose
 
 
+@pytest.fixture
+def rg():
+    return RandomGenerator(PCG64(seed=1234))
```
JHopeCollins, line 12: Same as before, setting seed in each test will be deterministic

tests/firedrake/adjoint/test_hessian.py
```diff
@@ -7,6 +7,11 @@
 rng = default_rng()
 
 
+@pytest.fixture
+def rg():
+    return RandomGenerator(PCG64(seed=1234))
```
JHopeCollins, line 12: And here

JHopeCollins (dismissed):

firedrake/adjoint_utils/function.py
connorjward: But we are immediately bumping the dat version anyway since we still need the `wo` option.

tests/firedrake/adjoint/test_assemble.py
connorjward: Actually this is OK I think. The fixture is regenerated for every test: https://docs.pytest.org/en/7.1.x/how-to/fixtures.html#fixture-scopes.

firedrake/adjoint_utils/function.py
JHopeCollins: This would bump it twice wouldn't it?

tests/firedrake/adjoint/test_assemble.py
JHopeCollins: TIL!

firedrake/adjoint_utils/function.py
connorjward: But that doesn't matter. The dat version only matters to indicate that something has changed. We never care about the actual value.

JHopeCollins (approved):
LGTM, thanks

PR #4511, opened 2025-08-20
Fix multigrid caches

pbrubeck:
# Description

### DIFF
--- a/firedrake/preconditioners/pmg.py
+++ b/firedrake/preconditioners/pmg.py
@@ -1212,7 +1212,7 @@ class StandaloneInterpolationMatrix(object):
     def work_function(V):
         if isinstance(V, firedrake.Function):
             return V
-        cache = V.mesh()._shared_data_cache["pmg_work_function"]
+        cache = V.mesh()._geometric_shared_data_cache["pmg_work_function"]
         key = (V.ufl_element(), V.value_shape, V.boundary_set)
         try:
             return cache[key]


ksagiyam (changes requested):
Please add tests.

ksagiyam:
Please fix `NotImplementedError: SLATE doesn't work in complex mode yet`.

tests/firedrake/multigrid/test_p_multigrid.py
```diff
@@ -73,6 +73,17 @@ def test_reconstruct_degree(tp_mesh, mixed_family):
...
 
+def test_work_function_cache(tp_mesh):
+    from firedrake.preconditioners.pmg import StandaloneInterpolationMatrix
+
+    V1 = FunctionSpace(tp_mesh, "Lagrange", 1)
+    V2 = FunctionSpace(tp_mesh, "Lagrange", 1)
+    assert V1 is not V2
+    w1 = StandaloneInterpolationMatrix.work_function(V1)
+    w2 = StandaloneInterpolationMatrix.work_function(V2)
+    assert w1 is w2
```
ksagiyam, line 84: This make me wonder if we should implement `restore()` method to keep track of if the `work_function` is `in_use` or not, but it's not the scope of this PR.

firedrake/mg/embedded.py
```diff
@@ -87,14 +87,17 @@ def _native_transfer(self, element, op):
...
     def cache(self, V):
+        mh, _ = get_level(V.mesh())
+        caches = mh._shared_data_cache["transfer_manager_cache"]
         key = (V.ufl_element(), V.value_shape, V.boundary_set)
         try:
-            return self.caches[key]
+            return caches[key]
         except KeyError:
-            return self.caches.setdefault(key, TransferManager.Cache(*key[:2]))
+            return caches.setdefault(key, TransferManager.Cache(*key[:2]))
```
ksagiyam, line 96:
```suggestion
            return caches.setdefault(key, TransferManager.Cache(*key[:3]))
```
? If so, `TransferManager.Cache.__init__()` must be updated. I should have asked this in the previous PR.

pbrubeck: The Cache class only needs to know the first two items: the `ufl_element` and the `value_shape` to construct an embedding DG element. Presumably we can manage without `value_shape`...

firedrake/preconditioners/pmg.py
```diff
@@ -1209,14 +1208,16 @@ def __init__(self, Vc, Vf, Vc_bcs, Vf_bcs):
             self.Vc_bcs = [bc.reconstruct(V=self.Vc, g=0) for bc in self.Vc_bcs]
             self.Vf_bcs = [bc.reconstruct(V=self.Vf, g=0) for bc in self.Vf_bcs]
 
-    def work_function(self, V):
+    @staticmethod
+    def work_function(V):
         if isinstance(V, firedrake.Function):
             return V
-        key = (V.ufl_element(), V.mesh(), V.boundary_set)
+        cache = V.mesh()._shared_data_cache["pmg_work_function"]
```
ksagiyam:
```suggestion
        cache = V.mesh()._geometric_shared_data_cache["pmg_work_function"]
```

tests/firedrake/multigrid/test_p_multigrid.py
pbrubeck: Should we also look at `FunctionSpace.get_work_function()`?

ksagiyam: That's probably better as they already have `restore` method.

pbrubeck: This will soon become irrelevant. I plan to eliminate this class in favour of `Interpolator`

PR #4513, opened 2025-08-21
Use GitHub notebook preview instead of nbviewer

connorjward:
nbviewer is slow and occasionally fails the linkcheck.


JHopeCollins (approved):
LGTM. Failures are unrelated.

PR #4515, opened 2025-08-22
Revert `Cofunction.riesz_representation` argument change

jrmaddison:
Revert `solver_options` to `solver_parameters`, consistent with release and naming elsewhere in Firedrake.

Opening as PR rather than an issue as it's a small change in previously untested code.


jrmaddison:
I'd rather not make changes outside of Firedrake, would a backwards compatible `FutureWarning` be OK?

PR #4484, opened 2025-08-07
Cell location on simplicial higher-order meshes

leo-collins:
This PR updates the bounding box algorithm to work with simplicial higher-order meshes. To do this we project the mesh coordinate function onto a Bernstein finite element space. Written in the Bernstein basis, the convex hull of the mesh nodes contains the entire element so no points are lost. Currently we can't create a Bernstein element on non-simplicial cells, so no quad or extruded meshes

I've also separated and cached the bounding box calculation. When the mesh tolerance is changed we just expand the bounding boxes. If the mesh coordinates change we do recalculate the bounding boxes, however.

PR #4516, opened 2025-08-22
introduce `PointEvaluator` object

leo-collins:
`PointEvaluator` is a convenience object for evaluating a `Function` at a set of points. The idea is for the user to avoid the boilerplate when point evaluating using a VertexOnlyMesh. Now that https://github.com/firedrakeproject/firedrake/pull/4484 is merged we should be able to deprecate `.at()` in favour of this API.

The basic usage is:
```
evaluator = PointEvaluator(mesh, points)
function_at_points = evaluator.evaluate(function)
```

The kwargs you can pass to `PointEvaluator` are `tolerance` and `missing_points_behaviour` which are self-explanatory, and `redundant`. If `redundant=True` (the default) then only the points on rank 0 are evaluated, and we broadcast the result to all other ranks. If `redundant=False` then each rank evaluates the points it has been given and returns the results on that rank. This is useful for e.g. external data input.

### DIFF
--- a/firedrake/function.py
+++ b/firedrake/function.py
@@ -25,7 +25,8 @@ from firedrake.cofunction import Cofunction, RieszMap
 from firedrake import utils
 from firedrake.adjoint_utils import FunctionMixin
 from firedrake.petsc import PETSc
-from firedrake.mesh import MeshGeometry
+from firedrake.mesh import MeshGeometry, VertexOnlyMesh
+from firedrake.functionspace import FunctionSpace, VectorFunctionSpace, TensorFunctionSpace
 
 __all__ = ['Function', 'PointNotInDomainError', 'CoordinatelessFunction', 'PointEvaluator']
 
@@ -725,7 +726,6 @@ class PointEvaluator:
             external data that is already distributed across ranks.
             Default is True.
         """
-        from firedrake.mesh import VertexOnlyMesh
         self.mesh = mesh
         self.points = np.asarray(points, dtype=utils.ScalarType)
         self.redundant = redundant
@@ -752,12 +752,11 @@ class PointEvaluator:
             (len(points), n, m). If the function is a mixed function, a tuple of Numpy arrays is returned,
             one for each subfunction.
         """
+        from firedrake import assemble, interpolate
         if not isinstance(function, Function):
             raise TypeError(f"Expected a Function, got f{type(function).__name__}")
         if function.function_space().mesh() is not self.mesh:
             raise ValueError("Function mesh must be the same Mesh object as the PointEvaluator mesh.")
-        from firedrake import interpolate, assemble, FunctionSpace, VectorFunctionSpace, TensorFunctionSpace
-
         subfunctions = function.subfunctions
         if len(subfunctions) > 1:
             return tuple(self.evaluate(subfunction) for subfunction in subfunctions)
--- a/firedrake/function.py
+++ b/firedrake/function.py
@@ -11,6 +11,7 @@ from ctypes import POINTER, c_int, c_double, c_void_p
 from collections.abc import Collection
 from numbers import Number
 from pathlib import Path
+from functools import partial
 
 from pyop2 import op2, mpi
 from pyop2.exceptions import DataTypeError, DataValueError
@@ -725,16 +726,32 @@ class PointEvaluator:
         :arg function: The :class:`Function` to evaluate.
         :kwarg input_ordered: If ``True``, return results in the order of the input points. If any 
         points were not found in the mesh, they will be return as np.nan.
-        :returns: A NumPy array of values at the points.
+        :returns: A Numpy array of values at the points. If the function is a mixed function, a list
+        of Numpy arrays is returned, one for each subfunction.
         """
         if not isinstance(function, Function):
             raise TypeError(f"Expected a Function, got f{type(function).__name__}")
-        from firedrake import interpolate, assemble, FunctionSpace
+        from firedrake import interpolate, assemble, FunctionSpace, VectorFunctionSpace, TensorFunctionSpace
 
-        P0DG = FunctionSpace(self.vom, "DG", 0)
+        subfunctions = function.subfunctions
+        if len(subfunctions) > 1:
+            result = []
+            for subfunction in subfunctions:
+                result.append(self.evaluate(subfunction, input_ordered=input_ordered))
+            return result
+
+        shape = function.ufl_function_space().value_shape
+        if len(shape) == 0:
+            fs = FunctionSpace
+        elif len(shape) == 1:
+            fs = VectorFunctionSpace
+        else:
+            fs = partial(TensorFunctionSpace, shape=shape)
+        P0DG = fs(self.vom, "DG", 0)
         f_at_points = assemble(interpolate(function, P0DG))
+
         if input_ordered:
-            P0DG_io = FunctionSpace(self.vom.input_ordering, "DG", 0)
+            P0DG_io = fs(self.vom.input_ordering, "DG", 0)
             f_at_points_io = Function(P0DG_io).assign(np.nan)
             f_at_points_io.interpolate(f_at_points)
             return f_at_points_io.dat.data_ro
--- a/firedrake/function.py
+++ b/firedrake/function.py
@@ -705,7 +705,7 @@ class PointEvaluator:
     r"""Convenience class for evaluating a :class:`Function` at a set of points."""
 
     def __init__(self, mesh: MeshGeometry, points: np.ndarray | list, tolerance: float | None = None,
-                 missing_points_behaviour: str | None = "error", redundant: bool = True) -> None:
+                 missing_points_behaviour: str = "error", redundant: bool = True) -> None:
         r"""
         Parameters
         ----------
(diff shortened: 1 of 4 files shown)


firedrake/function.py
```diff
@@ -698,6 +699,93 @@ def __str__(self):
...
+            The :class:`Mesh` on which the :class:`Function` is defined.
+        points : np.ndarray | list
+            Array of points to evaluate the function at.
+        tolerance : float | None
+            Tolerance to use when checking if a point is in a cell. Default is the ``tolerance`` of the :class:`Mesh`.
+        missing_points_behaviour : str | None
+            Behaviour when a point is not found in the mesh. Options are:
+            "error": raise a :class:`VertexOnlyMeshMissingPointsError` if a point is not found in the mesh.
+            "warn": warn if a point is not found in the mesh, but continue.
+            None: ignore points not found in the mesh.
```
connorjward: I think this is quite an odd API choice. It's much clearer for this to be "ignore".

firedrake/function.py
```diff
@@ -698,6 +699,93 @@ def __str__(self):
...
+        redundant : bool
+            If True, only the points on rank 0 are evaluated, and the result is broadcast to all ranks.
+            If False, each rank evaluates the points it has been given.
+            Default is True.
```
```suggestion
            If `True`, only the points on rank 0 are evaluated, and the result is broadcast to all ranks.
            If `False`, each rank evaluates the points it has been given.
```
Please add a bit more explanation explaining why one might want to use one of these over the other/when it should be `False`.

firedrake/function.py
```diff
@@ -698,6 +699,93 @@ def __str__(self):
...
+                 missing_points_behaviour: str | None = "error", redundant: bool = True) -> None:
+        r"""
+        Parameters
+        ----------
+        mesh : :class:`MeshGeometry`
+            The :class:`Mesh` on which the :class:`Function` is defined.
+        points : np.ndarray | list
+            Array of points to evaluate the function at.
+        tolerance : float | None
+            Tolerance to use when checking if a point is in a cell. Default is the ``tolerance`` of the :class:`Mesh`.
```
```suggestion
            Tolerance to use when checking if a point is in a cell. Default is the ``tolerance`` of the :class:`MeshGeometry`.
```

firedrake/function.py
```diff
@@ -698,6 +699,93 @@ def __str__(self):
...
+            Behaviour when a point is not found in the mesh. Options are:
+            "error": raise a :class:`VertexOnlyMeshMissingPointsError` if a point is not found in the mesh.
+            "warn": warn if a point is not found in the mesh, but continue.
+            None: ignore points not found in the mesh.
+        redundant : bool
+            If True, only the points on rank 0 are evaluated, and the result is broadcast to all ranks.
+            If False, each rank evaluates the points it has been given.
+            Default is True.
+        """
+        from firedrake.mesh import VertexOnlyMesh
```
connorjward: Can this not be a top-level import?

firedrake/function.py
```diff
@@ -698,6 +699,93 @@ def __str__(self):
...
+        self.mesh = mesh
+        self.points = np.asarray(points, dtype=utils.ScalarType)
+        self.redundant = redundant
+        self.vom = VertexOnlyMesh(
+            mesh, points, missing_points_behaviour=missing_points_behaviour,
+            redundant=redundant, tolerance=tolerance
+        )
+
+    def evaluate(self, function: Function) -> np.ndarray | list[np.ndarray]:
+        r"""Evaluate the given :class:`Function` at the points provided to this
```
connorjward: The first sentence of a docstring should be a single (short) line.

firedrake/function.py
```diff
@@ -698,6 +699,93 @@ def __str__(self):
...
+        :class:`PointEvaluator`. Points that were not found in the mesh will be
+        evaluated to np.nan.
+
+        Parameters
+        ----------
+        function : :class:`Function`
+            The :class:`Function` to evaluate.
```
```suggestion
        function :
            The `Function` to evaluate.
```

firedrake/function.py
```diff
@@ -698,6 +699,93 @@ def __str__(self):
...
+            has shape (len(points),). If the function is vector-valued with shape (n,), the Numpy array has shape
+            (len(points), n). If the function is tensor-valued with shape (n, m), the Numpy array has shape
+            (len(points), n, m). If the function is a mixed function, a list of Numpy arrays is returned,
+            one for each subfunction.
+        """
+        if not isinstance(function, Function):
+            raise TypeError(f"Expected a Function, got f{type(function).__name__}")
+        if function.function_space().mesh() is not self.mesh:
+            raise ValueError("Function mesh must be the same Mesh object as the PointEvaluator mesh.")
+        from firedrake import interpolate, assemble, FunctionSpace, VectorFunctionSpace, TensorFunctionSpace
```
connorjward: Style: put these imports at the top of the function (or preferably at the top of the file).

firedrake/function.py
```diff
@@ -698,6 +699,93 @@ def __str__(self):
...
+        evaluated to np.nan.
+
+        Parameters
+        ----------
+        function : :class:`Function`
+            The :class:`Function` to evaluate.
+
+        Returns
+        -------
+        np.ndarray | list[np.ndarray]
```
connorjward: Personally I prefer to return a tuple here because we want the result to be immutable.

connorjward (changes requested):
Awesome.

Could you amend the manual to recommend doing this instead of the full VoM ritual?

firedrake/function.py
leo-collins: I agree, for some reason that's how it's defined here:

https://github.com/firedrakeproject/firedrake/blob/1b677f2e15f37f300d2181012797329b1aab0eaf/firedrake/mesh.py#L3319-L3322

I can change it but I'm not sure if it'll break other things yet

leo-collins:
> This is just a suggestion that I think might align better with the rest of firedrake. How about something like this?
> 
> ```
> def VertexOnlyFunctionSpace(mesh, points):
>     # TODO some caching
>     return FunctionSpace(VertexOnlyMesh(mesh, points), "DG", 0)
> 
> DG_vom = VertexOnlyFunctionSpace(mesh, points)
> f_at_points = assemble(interpolate(f, DG_vom))
> ```

Problem with this is if you want to points in order (which is what users expect) then you'd still need to create a `DG_vom_input_ordering` and interpolate `f_at_points` into that. In parallel you also end up with all the results on rank 0. If you want them on the ranks where the points are, you'd need to broadcast them, or set `redundant=False` when constructing the vom, but then you end up doing the same calculation on every rank.

pbrubeck:
> Problem with this is if you want to points in order (which is what users expect) then you'd still need to create a `DG_vom_input_ordering` and interpolate `f_at_points` into that. In parallel you also end up with all the results on rank 0. If you want them on the ranks where the points are, you'd need to broadcast them, or set `redundant=False` when constructing the vom, but then you end up doing the same calculation on every rank.

Seems like you want a `VertexOnlyFunctionSpace(mesh, points, preserve_input_ordering=True)` and let it compose the two interpolations

firedrake/function.py
leo-collins: Yeah this broke a ton of stuff

firedrake/function.py
```diff
@@ -698,6 +701,88 @@ def __str__(self):
...
+            external data that is already distributed across ranks.
+            Default is True.
+        """
+        self.mesh = mesh
+        self.points = np.asarray(points, dtype=utils.ScalarType)
+        self.redundant = redundant
+        self.vom = VertexOnlyMesh(
+            mesh, points, missing_points_behaviour=missing_points_behaviour,
+            redundant=redundant, tolerance=tolerance
+        )
```
dham, line 745: need to record coordinate dat version and reimmerse the VOM when needed.

Also need a moving coordinate test.

firedrake/function.py
```diff
@@ -698,6 +701,88 @@ def __str__(self):
...
+        missing_points_behaviour : str | None
+            Behaviour when a point is not found in the mesh. Options are:
+            "error": raise a :class:`VertexOnlyMeshMissingPointsError` if a point is not found in the mesh.
+            "warn": warn if a point is not found in the mesh, but continue.
+            None: ignore points not found in the mesh.
+        redundant : bool
+            If True, only the points on rank 0 are evaluated, and the result is broadcast to all ranks.
+            If False, each rank evaluates the points it has been given. False is useful if you are inputting
+            external data that is already distributed across ranks.
+            Default is True.
```
JHopeCollins: Clarify that "points on rank 0" means "the points given to the constructor by rank 0" and not "the points that live in rank 0's part of the mesh".

dham (changes requested):

firedrake/function.py
leo-collins: Fixed in https://github.com/firedrakeproject/firedrake/pull/4524

firedrake/function.py
```diff
@@ -698,6 +701,103 @@ def __str__(self):
...
+            The :class:`Function` to evaluate.
+
+        Returns
+        -------
+        numpy.ndarray | Tuple[numpy.ndarray, ...]
+            A Numpy array of values at the points. If the function is scalar-valued, the Numpy array
+            has shape ``(len(points),)``. If the function is vector-valued with shape ``(n,)``, the Numpy array has shape
+            ``(len(points), n)``. If the function is tensor-valued with shape ``(n, m)``, the Numpy array has shape
+            ``(len(points), n, m)``. If the function is a mixed function, a tuple of Numpy arrays is returned,
+            one for each subfunction.
```
dham, line 763: Put in a warning that returning a numpy array means this isn't taped.

dham (dismissed):
Just put in the taping warning and raise the issue about moving meshes.

PR #4517, opened 2025-08-22
Cell location on non-simplicial higher-order meshes

leo-collins:
With https://github.com/firedrakeproject/fiat/pull/165 merged, we can finish off what https://github.com/firedrakeproject/firedrake/pull/4484 started. Now the cell-location algorithm works on higher-order meshes including quad, extruded etc.

### DIFF
--- a/docs/source/interpolation.rst
+++ b/docs/source/interpolation.rst
@@ -163,11 +163,6 @@ For more on forms, see :ref:`this section of the manual <more_complicated_forms>
 Interpolating onto other meshes
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
-.. note::
-
-   Interpolation *from* :ref:`high-order meshes <changing_coordinate_fs>` is
-   currently not supported.
-
 If the target mesh extends outside the source mesh domain, then cross-mesh
 interpolation will raise a :py:class:`~.DofNotDefinedError`.
 


connorjward (dismissed):
Can we have these tests be parallel too?

firedrake/mesh.py
```diff
@@ -2488,17 +2488,19 @@ def bounding_box_coords(self) -> Tuple[np.ndarray, np.ndarray] | None:
 
         coord_element = self.ufl_coordinate_element()
         coord_degree = coord_element.degree()
-        if ufl.cell.simplex(self.topological_dimension()) != self.ufl_cell():
-            # Non-simplex element, e.g. quad or tensor product
-            mesh = self
-        elif coord_degree == 1:
+        if np.all(np.asarray(coord_degree) == 1):
```
pbrubeck, line 2491:
```suggestion
        if np.array_equal(coord_degree, 1):
```

leo-collins: This returns False if `coord_degree = (1, 1)`

firedrake/mesh.py
```diff
@@ -2488,17 +2488,19 @@ def bounding_box_coords(self) -> Tuple[np.ndarray, np.ndarray] | None:
...
             mesh = self
         elif coord_element.family() == "Bernstein":
             # Already have Bernstein coordinates, no need to project
             mesh = self
         else:
             # For bendy meshes we project the coordinate function onto Bernstein
-            bernstein_fs = functionspace.VectorFunctionSpace(self, "Bernstein", coord_degree)
+            if self.extruded:
```
pbrubeck, line 2498:
```suggestion
            if self.extruded and len(set(coord_degree)) > 1:
```

firedrake/mesh.py
```diff
@@ -2488,17 +2488,19 @@ def bounding_box_coords(self) -> Tuple[np.ndarray, np.ndarray] | None:
...
+                bernstein_fs = functionspace.VectorFunctionSpace(
+                    self, "Bernstein", coord_degree[0], vfamily="Bernstein", vdegree=coord_degree[1]
+                )
+            else:
+                bernstein_fs = functionspace.VectorFunctionSpace(self, "Bernstein", coord_degree)
```
pbrubeck, line 2503: Unpack the degree to a single int for this case

pbrubeck: There is a bug here, it should be fixed in FIAT, this is fine for now.

dham (approved):

PR #4526, opened 2025-08-28
Use `petsctools.AppContext`

JHopeCollins:
Switch from an un-namespaced python `dict` to the namespaced `petsctools.AppContext` in this PR: https://github.com/firedrakeproject/petsctools/pull/16

### DIFF
--- a/firedrake/preconditioners/gtmg.py
+++ b/firedrake/preconditioners/gtmg.py
@@ -66,11 +66,14 @@ class GTMGPC(PCBase):
         :arg pc: PETSc preconditioner instance
         """
         from firedrake.assemble import assemble, get_assembler
+        from dmhooks import get_appctx
 
         Citations().register("Gopalakrishnan2009")
         _, P = pc.getOperators()
-        appctx = self.get_appctx(pc)
-        fcp = appctx.get("form_compiler_parameters")
+        ctx = get_appctx(pc)
+        fcp = ctx.fcp
+
+        appctx = ctx.appctx
 
         if pc.getType() != "python":
             raise ValueError("Expecting PC type python")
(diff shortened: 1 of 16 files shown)


firedrake/preconditioners/assembled.py
```diff
@@ -18,13 +18,13 @@ class AssembledPC(PCBase):
...
         from firedrake.assemble import get_assembler
+        from dmhooks import get_appctx
         A, P = pc.getOperators()
 
         if pc.type != "python":
             raise ValueError("Expecting PC type python")
         opc = pc
-        appctx = self.get_appctx(pc)
-        fcp = appctx.get("form_compiler_parameters")
+        fcp = get_appctx(pc).fcp
```
connorjward, line 27: Did you set `fcp`? It's not a clear name.

firedrake/preconditioners/gtmg.py
```diff
@@ -66,11 +66,14 @@ def initialize(self, pc):
...
+        from dmhooks import get_appctx
 
         Citations().register("Gopalakrishnan2009")
         _, P = pc.getOperators()
-        appctx = self.get_appctx(pc)
-        fcp = appctx.get("form_compiler_parameters")
+        ctx = get_appctx(pc)
+        fcp = ctx.fcp
+
+        appctx = ctx.appctx
```
connorjward, line 76: This is weird. This is saying `appctx = get_appctx(pc).appctx`

firedrake/solving_utils.py
```diff
@@ -194,14 +193,13 @@ def __init__(self, problem, mat_type, pmat_type, appctx=None,
         self._x = problem.u_restrict
 
         if appctx is None:
-            appctx = {}
+            from petsctools import AppContext
```
connorjward, line 196: Do import at the top level

connorjward (changes requested):
Quick skim. I hope this is useful and not just annoying!

firedrake/preconditioners/assembled.py
JHopeCollins: No, it is already attached to the `_SNESContext`. I'm just switching to getting the compiler parameters from there instead of the appctx because it is a "global" rather than a namespaced (prefixed) thing.

firedrake/preconditioners/gtmg.py
JHopeCollins: I agree, this is really annoying. There is the `_SNESContext` object and the `dict` object, both of which are called `appctx` in various places.

- `_SNESContext.appctx` is the `dict` we call the `appctx` in user-code.
- `dmhooks.get_appctx` returns the `_SNESContext`.
- `PCSNESBase.get_appctx` returns `dmhooks.get_appctx(pc_or_snes).appctx` which is the `dict`.

I would be in favour of cleaning up this naming, but it'd touch a lot of code and I don't want to do it unilaterally.

connorjward: Sure. I'd be happy for now to just open an issue about it, plus perhaps an explanatory comment.

PR #4529, opened 2025-08-29
RF Eigenproblem

Shawniz:
This PR introduces a new eigenproblem/eigensolver abstraction that enables one to compute spectra of the linearisation of a sequence of Firedrake operations. The RFEigenproblem/RFEigensolver tandem closely mirror the LinearEigenproblem/LinearEigensolver tandem.
The main changes are introduced in the RFEigenproblem abstraction. Inspired by the interfacing and ability to define matrices implicitly introduced by @JHopeCollins in the JHopeCollins/tao-rf-mats branch, a new context RestrictedReducedFunctionalCtx object is used in the creation of the RestrictedReducedFunctionalMat which represents the linearisation of a sequence of FIredrake operations on a restricted function space.

### DIFF
--- a/tests/firedrake/adjoint/test_rf_eigensolver.py
+++ b/tests/firedrake/adjoint/test_rf_eigensolver.py
@@ -46,14 +46,15 @@ def test_compare_eigensolvers_helmholtz(expression):
 
     # ReducedFunctional-based, Jx = 1/cx
     continue_annotation()
-    f = Function(V)
-    f.interpolate(2 * x)
-    control = Control(f)
-    u_1 = Function(V)
-    v = TestFunction(V)
-    F = (inner(grad(u_1), grad(v)) + u_1 * v - f * v) * dx
-    solve(F == 0, u_1, bcs=bc)
-    J = ReducedFunctional(u_1, controls=[control])
+    with set_working_tape() as tape:
+        f = Function(V)
+        f.interpolate(2 * x)
+        control = Control(f)
+        u_1 = Function(V)
+        v = TestFunction(V)
+        F = (inner(grad(u_1), grad(v)) + u_1 * v - f * v) * dx
+        solve(F == 0, u_1, bcs=bc)
+        J = ReducedFunctional(u_1, controls=[control], tape=tape)
     pause_annotation()
 
     problem_rf = RFEigenproblem(J, bcs=bc, apply_riesz=False, restrict=True, identity=True)


firedrake/rf_eigensolver.py
```diff
@@ -0,0 +1,417 @@
...
+    @utils.cached_property
+    def dm(self):
+        r"""Return the dm associated with the output space."""
+        if self.restrict:
+            return self.restricted_space.dm
+        else:
+            return self.output_space.dm
+
+
+class RFEigensolver(OptionsManager):
```
connorjward: Inheriting from `OptionsManager` is no longer recommended practice. Use the free functions in petsctools instead.

firedrake/rf_eigensolver.py
```diff
@@ -0,0 +1,417 @@
...
+try:
+    from slepc4py import SLEPc
+except ImportError:
+    SLEPc = None
+
+__all__ = ["RFEigenproblem", "RFEigensolver"]
+
+
+class RFEigenproblem:
+    """Generalised linear eigenvalue problem.
```
connorjward: Docstring does not really describe the reduced functional aspect of this class.

firedrake/restricted_functional_ctx.py
connorjward, line 1: I would rather not add multiple modules into the `firedrake/` directory. I wonder if these should be combined.

firedrake/restricted_functional_ctx.py
```diff
@@ -0,0 +1,250 @@
...
+    PETSc = None
+
+
+def new_restricted_control_variable(reduced_functional, function_space, dual=False):
+    """Return new variables suitable for storing a control value or its dual
+        by interpolating into the space over which the 'ReducedFunctional' is
+        defined.
+
+    Args:
+    ----------
```
connorjward: Not proper numpydoc. Should be `Parameters` and the underline match.

firedrake/restricted_functional_ctx.py
```diff
@@ -0,0 +1,250 @@
...
+)
+from firedrake import Function, Cofunction
+
+try:
+    import petsc4py.PETSc as PETSc
+except ModuleNotFoundError:
+    PETSc = None
+
+
+def new_restricted_control_variable(reduced_functional, function_space, dual=False):
```
connorjward: Please use type hints throughout.

firedrake/restricted_functional_ctx.py
```diff
@@ -0,0 +1,250 @@
...
+        )
+
+        if action in (AdjointAction, TLMAction):
+            self.restricted_functional_interface = (
+                PETScVecInterface(Function(restricted_space).interpolate(rf.functional), comm=comm)
+                if restricted_space
+                else self.functional_interface
+            )
+
+        if action == HessianAction:
```
connorjward: If this is a class then I think this should be `is` not `==`.

firedrake/rf_eigensolver.py
```diff
@@ -0,0 +1,417 @@
...
+
+try:
+    from slepc4py import SLEPc
+except ImportError:
+    SLEPc = None
+
+__all__ = ["RFEigenproblem", "RFEigensolver"]
+
+
+class RFEigenproblem:
```
connorjward, line 48: I thought we wanted to move away from the problem+solver abstraction because that's a legacy FEniCS thing.

firedrake/rf_eigensolver.py
Shawniz: I modified the docstring. Now I believe it is clearer.

firedrake/restricted_functional_ctx.py
Shawniz: I added type hints where I believed arguments are unclear.

firedrake/restricted_functional_ctx.py
Shawniz: It is an Enum member of a class.

firedrake/rf_eigensolver.py
Shawniz: I am unsure what to answer.

connorjward: Yeah sorry this was quite a technical point and probably wrong anyway.

What we probably want is an `eigensolve()` free function that mimics `solve()`. That is definitely not the job of this PR though.

firedrake/restricted_functional_ctx.py
connorjward: I still think that `is` is the right thing here.

firedrake/restricted_functional_ctx.py
connorjward: Thanks, but please include them everywhere that you've written a docstring.

firedrake/rf_eigensolver.py
```diff
@@ -0,0 +1,675 @@
...
+    def dm(self):
+        r"""Return the dm associated with the output space."""
+        if self.restrict:
+            return self.restricted_space.dm
+        else:
+            return self.output_space.dm
+
+
+class RFEigensolver:
+    r"""Solve a LinearEigenproblem.
```
connorjward: Shouldn't this be a `RFEigenProblem`?

firedrake/rf_eigensolver.py
```diff
@@ -0,0 +1,675 @@
...
+
+OneOrManyFunction   = Function   | list[Function]   | tuple[Function, ...]
+OneOrManyCofunction = Cofunction | list[Cofunction] | tuple[Cofunction, ...]
+VarsInterpolate = OneOrManyFunction | OneOrManyCofunction
+
+__all__ = ["RFEigenproblem", "RFEigensolver"]
+
+
+class RFEigenproblem:
+    """Generalised eigenvalue problem.
```
connorjward: I don't think that this line really describes what is going on. This is just copied from `eigensolver.py`.

firedrake/rf_eigensolver.py
```diff
@@ -0,0 +1,675 @@
...
+    identity : bool
+        If True, M is replaced by identity matrix. Differentiate between mass
+        matrix being identity and inner product operator.
+    action : (Optional[TLMAction, AdjointAction])
+        Defines forward action of implicit matrix operator.
+    apply_riez : bool
+        If True, when defining A's adjoint action, we can output into the dual space.
+    action : str
+        Determine whether forward action is TLM or adjoint.
+    appctx : Optional[dict]
```
connorjward:
```suggestion
    action :
        Determine whether forward action is TLM or adjoint.
    appctx :
```
You can omit the type hints in the docstring because they are in the function signature.

firedrake/rf_eigensolver.py
```diff
@@ -0,0 +1,675 @@
...
+        version. The output space remains unchanged.
+    identity : bool
+        If True, M is replaced by identity matrix. Differentiate between mass
+        matrix being identity and inner product operator.
+    action : (Optional[TLMAction, AdjointAction])
+        Defines forward action of implicit matrix operator.
+    apply_riez : bool
+        If True, when defining A's adjoint action, we can output into the dual space.
+    action : str
+        Determine whether forward action is TLM or adjoint.
```
connorjward: Please enumerate the valid options here.

firedrake/restricted_functional_ctx.py
Shawniz: Done.

firedrake/restricted_functional_ctx.py
Shawniz: Changed.

firedrake/rf_eigensolver.py
Shawniz: Done.

firedrake/rf_eigensolver.py
Shawniz: I think now it is clearer. Apologies, forgot to change.

firedrake/rf_eigensolver.py
Shawniz: Updated for more clarity.

firedrake/rf_eigensolver.py
JHopeCollins: I would also be in favour of moving away from the problem+solver pattern, but I would argue that should be done all at the same time. Given we still have variational and eigen problem+solver, the law of least surprise says that the rfeigen should also use this pattern.

tests/firedrake/adjoint/test_rf_eigensolver.py
```diff
@@ -0,0 +1,192 @@
...
+def generate_expressions():
+    """
+    Returns a list of three expressions.
+    """
+    mesh = UnitSquareMesh(50, 50, quadrilateral=False)
+    x, y = SpatialCoordinate(mesh)
+    return [2 * x, exp(-x - y), sin(pi * x * y)]
+
+
+@pytest.mark.parametrize("expression", generate_expressions())
```
JHopeCollins, line 22: Is `expression` actually used?

firedrake/rf_eigensolver.py
connorjward: I was thinking about this and potentially the problem+solver approach is nice from an OOP point of view because of encapsulation. The issue is that it is clunky to construct.

You could have a free `eigensolve` function or `Eigensolver.from_equation()` factory method instead.

tests/firedrake/adjoint/test_rf_eigensolver.py
```diff
@@ -0,0 +1,192 @@
...
+    a = (inner(grad(u), grad(v)) + u * v) * dx
+
+    # Form based Ax = cMx
+    problem_form = LinearEigenproblem(a, bcs=bc, restrict=True)
+    solver_form = LinearEigensolver(problem_form, n_evals=num_eigenvalues)
+    solver_form.solve()
+    nconv = solver_form.es.getConverged()
+
+    evals_form = [
+        solver_form.eigenvalue(i) for i in range(0, min(num_eigenvalues, nconv))
```
JHopeCollins, line 44:
```suggestion
        solver_form.eigenvalue(i) for i in range(min(num_eigenvalues, nconv))
```

tests/firedrake/adjoint/test_rf_eigensolver.py
```diff
@@ -0,0 +1,192 @@
...
+
+    u = TrialFunction(V)
+    v = TestFunction(V)
+    a = (inner(grad(u), grad(v)) + u * v) * dx
+
+    # Form based Ax = cMx
+    problem_form = LinearEigenproblem(a, bcs=bc, restrict=True)
+    solver_form = LinearEigensolver(problem_form, n_evals=num_eigenvalues)
+    solver_form.solve()
+    nconv = solver_form.es.getConverged()
```
JHopeCollins, line 41: Should we be checking how many converged here?

tests/firedrake/adjoint/test_rf_eigensolver.py
```diff
@@ -0,0 +1,192 @@
...
+    # ReducedFunctional-based, Jx = 1/cx
+    continue_annotation()
+    f = Function(V)
+    f.interpolate(2 * x)
+    control = Control(f)
+    u_1 = Function(V)
+    v = TestFunction(V)
+    F = (inner(grad(u_1), grad(v)) + u_1 * v - f * v) * dx
+    solve(F == 0, u_1, bcs=bc)
+    J = ReducedFunctional(u_1, controls=[control])
```
JHopeCollins: We're moving towards tightly scoped tapes being the recommended way of building `ReducedFunctional`.
```suggestion
    with set_working_tape() as tape:
        f = Function(V)
        f.interpolate(2 * x)
        control = Control(f)
        u_1 = Function(V)
        v = TestFunction(V)
        F = (inner(grad(u_1), grad(v)) + u_1 * v - f * v) * dx
        solve(F == 0, u_1, bcs=bc)
        J = ReducedFunctional(u_1, controls=[control], tape=tape)
```

tests/firedrake/adjoint/test_rf_eigensolver.py
```diff
@@ -0,0 +1,192 @@
...
+
+    problem_rf = RFEigenproblem(J, bcs=bc, apply_riesz=False, restrict=True, identity=True)
+    solver_rf = RFEigensolver(problem_rf, n_evals=num_eigenvalues)
+    solver_rf.es.setTarget(1)
+    solver_rf.solve()
+    nconv = solver_rf.es.getConverged()
+
+    evals_rf = [
+        1 / solver_rf.eigenvalue(i) for i in range(0, min(num_eigenvalues, nconv))
+    ]
```
JHopeCollins, line 68: Why are we grabbing eigenvalues from the `LinearEigenSolver` but inverse eigenvalues from the `RFEigenSolver`?
Is this the right thing to do? If so, please add a comment explaining why.
```suggestion
    evals_rf = [
        1 / solver_rf.eigenvalue(i) for i in range(min(num_eigenvalues, nconv))
    ]
```

tests/firedrake/adjoint/test_rf_eigensolver.py
```diff
@@ -0,0 +1,192 @@
...
+
+    for ev1, ev2 in zip(evals_form, evals_rf):
+        assert abs(ev1 - ev2) < 1e-6
```
JHopeCollins, line 71: This makes sure that if we have a regression in the future, we see all eigenvalues that fail not just the first.
```suggestion
    evals_match = all(abs(evf - evr) < 1e-6 for evf, evr in zip(evals_form, evals_rf))
    assert evals_match, f"Eigenvalues from RFEigenSolver {evals_rf} do not match eigenvalues from LinearEigenSolver {evals_form}"
```

tests/firedrake/adjoint/test_rf_eigensolver.py
```diff
@@ -0,0 +1,192 @@
...
+
+    set_working_tape(Tape())
+
+
+@pytest.mark.parametrize("restrict", [False])
+@pytest.mark.parametrize("riesz", ["L2"])
+def test_apply_riesz_param(restrict, riesz):
```
JHopeCollins, line 78: If there is only one option per parameter, do they need to be parameters?

tests/firedrake/adjoint/test_rf_eigensolver.py
```diff
@@ -0,0 +1,192 @@
...
+    # Tape the solve
+    continue_annotation()
+    f = Function(V)
+    f.interpolate(2 * x * y)
+    control = Control(f)
+    u_1 = Function(V)
+    v = TestFunction(V)
+    F = (inner(grad(u_1), grad(v)) + u_1 * v - f * v) * dx
+    solve(F == 0, u_1, bcs=bc)
+    J = ReducedFunctional(u_1, controls=[control])
```
JHopeCollins, line 95:
```suggestion
    with set_working_tape() as tape:
        f = Function(V)
        f.interpolate(2 * x * y)
        control = Control(f)
        u_1 = Function(V)
        v = TestFunction(V)
        F = (inner(grad(u_1), grad(v)) + u_1 * v - f * v) * dx
        solve(F == 0, u_1, bcs=bc)
        J = ReducedFunctional(u_1, controls=[control], tape=tape)
```

tests/firedrake/adjoint/test_rf_eigensolver.py
```diff
@@ -0,0 +1,192 @@
...
+        func_representation = interpolate_vars(func_representation, V)
+        func_representation_dual = interpolate_vars(func_representation_dual, V)
+
+    set_working_tape(Tape())
+    assert (
+        errornorm(
+            func_representation[0],
+            func_representation_dual[0].riesz_representation(riesz_map=riesz),
+        )
+        < 1e-6
```
JHopeCollins, line 140:
```suggestion
            func_representation_dual[0].riesz_representation(riesz_map=riesz)
        ) < 1e-6
```

tests/firedrake/adjoint/test_rf_eigensolver.py
```diff
@@ -0,0 +1,192 @@
...
+    )
+
+
+def test_compare_tlm_adjoint():
+    """Test TLM and adjoint give same eigenvalues for Helmholtz eigenproblem."""
```
JHopeCollins, line 145:
```suggestion
    """Test TLM and adjoint give same eigenvalues for a symmetric eigenproblem (Helmholtz equation)."""
```

tests/firedrake/adjoint/test_rf_eigensolver.py
```diff
@@ -0,0 +1,192 @@
...
+    # ReducedFunctional-based, Jx = 1/cx
+    continue_annotation()
+    f = Function(V)
+    f.interpolate(2 * x)
+    control = Control(f)
+    u_1 = Function(V)
+    v = TestFunction(V)
+    F = (inner(grad(u_1), grad(v)) + u_1 * v - f * v) * dx
+    solve(F == 0, u_1, bcs=bc)
+    J = ReducedFunctional(u_1, controls=[control])
```
JHopeCollins, line 163:
```suggestion
    with set_working_tape() as tape:
        f = Function(V)
        f.interpolate(2 * x)
        control = Control(f)
        u_1 = Function(V)
        v = TestFunction(V)
        F = (inner(grad(u_1), grad(v)) + u_1 * v - f * v) * dx
        solve(F == 0, u_1, bcs=bc)
        J = ReducedFunctional(u_1, controls=[control], tape=tape)
```

tests/firedrake/adjoint/test_rf_eigensolver.py
```diff
@@ -0,0 +1,192 @@
...
+    problem_rf_a = RFEigenproblem(
+        J, bcs=bc, apply_riesz=False, restrict=True, identity=True, action="adjoint"
+    )
+    solver_rf_a = RFEigensolver(problem_rf_a, n_evals=num_eigenvalues)
+    solver_rf_a.es.setTarget(1)
+    solver_rf_a.solve()
+    nconv = solver_rf_a.es.getConverged()
+
+    evals_rf_a = [
+        solver_rf_a.eigenvalue(i) for i in range(0, min(num_eigenvalues, nconv))
```
JHopeCollins, line 175:
```suggestion
        solver_rf_a.eigenvalue(i) for i in range(min(num_eigenvalues, nconv))
```

tests/firedrake/adjoint/test_rf_eigensolver.py
```diff
@@ -0,0 +1,192 @@
...
+    problem_rf_t = RFEigenproblem(
+        J, bcs=bc, apply_riesz=False, restrict=True, identity=True, action="tlm"
+    )
+    solver_rf_t = RFEigensolver(problem_rf_t, n_evals=num_eigenvalues)
+    solver_rf_t.es.setTarget(1)
+    solver_rf_t.solve()
+    nconv = solver_rf_t.es.getConverged()
+
+    evals_rf_t = [
+        solver_rf_t.eigenvalue(i) for i in range(0, min(num_eigenvalues, nconv))
```
JHopeCollins, line 187:
```suggestion
        solver_rf_t.eigenvalue(i) for i in range(min(num_eigenvalues, nconv))
```

tests/firedrake/adjoint/test_rf_eigensolver.py
```diff
@@ -0,0 +1,192 @@
...
+    ]
+    for ev1, ev2 in zip(evals_rf_a, evals_rf_t):
+        print(abs(ev1 - ev2))
+        assert abs(ev1 - ev2) < 1e-6
```
JHopeCollins, line 191:
```suggestion
    evals_match = all(abs(eva - evt) < 1e-6 for evf, evr in zip(evals_rf_a, evals_rf_t))
    assert evals_match, f"Symmetric eigenvalues from adjoint operator {evals_rf_a} do not match eigenvalues from TLM operator {evals_rf_t}"
```

tests/firedrake/adjoint/test_rf_eigensolver.py
```diff
@@ -0,0 +1,192 @@
...
+    J = ReducedFunctional(u_1, controls=[control])
+    pause_annotation()
+
+    problem_rf_a = RFEigenproblem(
+        J, bcs=bc, apply_riesz=False, restrict=True, identity=True, action="adjoint"
+    )
+    solver_rf_a = RFEigensolver(problem_rf_a, n_evals=num_eigenvalues)
+    solver_rf_a.es.setTarget(1)
+    solver_rf_a.solve()
+    nconv = solver_rf_a.es.getConverged()
```
JHopeCollins, line 172: Should we be checking how many have converged?

tests/firedrake/adjoint/test_rf_eigensolver.py
```diff
@@ -0,0 +1,192 @@
...
+
+    evals_rf_a = [
+        solver_rf_a.eigenvalue(i) for i in range(0, min(num_eigenvalues, nconv))
+    ]
+
+    problem_rf_t = RFEigenproblem(
+        J, bcs=bc, apply_riesz=False, restrict=True, identity=True, action="tlm"
```
JHopeCollins, line 179:
```suggestion
        J, bcs=bc, apply_riesz=False, restrict=True, identity=True, action=TLMAction
```

tests/firedrake/adjoint/test_rf_eigensolver.py
```diff
@@ -0,0 +1,192 @@
...
+    control = Control(f)
+    u_1 = Function(V)
+    v = TestFunction(V)
+    F = (inner(grad(u_1), grad(v)) + u_1 * v - f * v) * dx
+    solve(F == 0, u_1, bcs=bc)
+    J = ReducedFunctional(u_1, controls=[control])
+    pause_annotation()
+
+    problem_rf_a = RFEigenproblem(
+        J, bcs=bc, apply_riesz=False, restrict=True, identity=True, action="adjoint"
```
JHopeCollins, line 167:
```suggestion
        J, bcs=bc, apply_riesz=False, restrict=True, identity=True, action=AdjointAction
```

tests/firedrake/adjoint/test_rf_eigensolver.py
```diff
@@ -0,0 +1,192 @@
...
+    J = ReducedFunctional(u_1, controls=[control])
+    pause_annotation()
+
+    # Check adjoint action
+    problem_rf_true = RFEigenproblem(
+        J, bcs=bc, apply_riesz=True, restrict=restrict, action="adjoint"
+    )
+    problem_rf_false = RFEigenproblem(
+        J, bcs=bc, apply_riesz=False, restrict=restrict, action="adjoint"
+    )
```
JHopeCollins, line 104:
```suggestion
    problem_rf_true = RFEigenproblem(
        J, bcs=bc, apply_riesz=True, restrict=restrict, action=AdjointAction
    )
    problem_rf_false = RFEigenproblem(
        J, bcs=bc, apply_riesz=False, restrict=restrict, action=AdjointAction
    )
```

firedrake/rf_eigensolver.py
```diff
@@ -0,0 +1,661 @@
...
+    MPI = None
+
+
+OneOrManyFunction   = Function   | list[Function]   | tuple[Function, ...]
+OneOrManyCofunction = Cofunction | list[Cofunction] | tuple[Cofunction, ...]
+VarsInterpolate = OneOrManyFunction | OneOrManyCofunction
+Comms = PETSc.Comm | MPI.Comm  | None
+AppCtxType = dict | None
+MassType = ReducedFunctional | Form | None
+BcType = DirichletBC | list[DirichletBC]
```
JHopeCollins, line 43: None of these are particularly complex type hints (especially with the simplifications below) so can be included directly in the function signatures rather than creating aliases.
```suggestion
from typing import Union, Optional, Collection
OneOrManyFunction   = Union[Function, Collection[Function]]
OneOrManyCofunction   = Union[Cofunction, Collection[Cofunction]]
VarsInterpolate = Union[Function, Cofunction, Collection[Function],  Collection[Cofunction]]
Comms = Optional[PETSc.Comm, MPI.Comm]
AppCtxType = Optional[dict]
MassType = Optional[ReducedFunctional, Form]
BcType = Union[DirichletBC, Collection[DirichletBC]]
```

firedrake/rf_eigensolver.py
```diff
@@ -0,0 +1,661 @@
...
+    solution, this is shifted by the specified amount. It is the user's
+    responsibility to ensure that the shift is not close to an actual
+    eigenvalue of the system.
+
+    Also, we assume the operator we linearise maps from V to V.
+    """
+
+    def __init__(
+        self,
+        rf: ReducedFunctional,
```
JHopeCollins, line 106: We should allow user defined reduced functionals via the abstract base class.
```suggestion
        rf: AbstractReducedFunctional,
```

firedrake/rf_eigensolver.py
```diff
@@ -0,0 +1,661 @@
...
+    comm :
+        Communicator that the rf is defined over.
+
+    Notes
+    -----
+    If restrict == True, the arguments of A and M will be replaced, such that
+    their function space is replaced by the equivalent RestrictedFunctionSpace
+    class. This avoids the computation of eigenvalues associated with the
+    Dirichlet boundary conditions. This in turn prevents convergence failures,
+    and allows only the non-boundary eigenvalues to be returned. The
```
JHopeCollins, line 91:
```suggestion
    and allows only the non-boundary eigenvalues to be returned. The returned
```

firedrake/rf_eigensolver.py
```diff
@@ -0,0 +1,661 @@
...
+    eigenvectors will be in the original, non-restricted space.
+
+    If restrict == False and Dirichlet boundary conditions are supplied, then
+    these conditions will result in the eigenproblem having a nullspace spanned
+    by the basis functions with support on the boundary. To facilitate
+    solution, this is shifted by the specified amount. It is the user's
+    responsibility to ensure that the shift is not close to an actual
+    eigenvalue of the system.
```
JHopeCollins, line 99:
```suggestion
    If restrict == False and Dirichlet boundary conditions are supplied, then
    these conditions will result in the eigenproblem being ill-posed, with
    a nullspace spanned by the basis functions with support on the boundary.
    To make the problem well-posed again, the problem is shifted by the
    `bc_shift` argument. It is the user's responsibility to ensure that the shift
    is not close to an actual eigenvalue of the system.
```

firedrake/rf_eigensolver.py
```diff
@@ -0,0 +1,661 @@
...
+
+    where for J: V -> V, dJ_m : V -> V is the linearisation of J around the state in the taped run.
+    
+    Additionally, the class has the capabilities of solving the generalised eigenvalue problem
+    defined in `LinearEigenproblem`.
+
+    Parameters
+    ----------
+    rf :
+        The bilinear operator A.
```
JHopeCollins, line 62:
```suggestion
    rf :
        The ReducedFunctional which provides the action of an equivalent bilinear operator A.
        The control and functional must be in the same space so that the operator is square.
```

firedrake/rf_eigensolver.py
```diff
@@ -0,0 +1,661 @@
...
+    bcs :
+        The Dirichlet boundary conditions.
+    bc_shift : 
+        The value to shift the boundary condition eigenvalues by. Ignored if restrict == True.
+    restrict : 
+        If True, replace the function spaces of u and v with their restricted
+        version. The output space remains unchanged.
+    identity : 
+        If True, M is replaced by identity matrix. Differentiate between mass
+        matrix being identity and inner product operator.
```
JHopeCollins, line 75: Please can you clarify what the second sentence means?

firedrake/rf_eigensolver.py
```diff
@@ -0,0 +1,661 @@
...
+    action : (Optional[TLMAction, AdjointAction])
+        Defines forward action of implicit matrix operator.
```
JHopeCollins:
```suggestion
    action : (Optional[TLMAction, AdjointAction])
        Defines whether to calculate the eigenvalues of the tangent linear model
        or the adjoint model of the ReducedFunctional.
```

firedrake/rf_eigensolver.py
JHopeCollins: The `action` docstring is now duplicated so this one can be removed.

firedrake/rf_eigensolver.py
```diff
@@ -0,0 +1,661 @@
...
+    apply_riez : 
+        If True, when defining A's adjoint action, we can output into the dual space.
+    action :
+        Determine whether forward action is TLM or adjoint.
+    appctx :
+        Context forwarded to the linearisation.
```
JHopeCollins, line 81: User context provided to the SLEPc solver.

firedrake/rf_eigensolver.py
```diff
@@ -0,0 +1,661 @@
...
+    eigenvectors will be in the original, non-restricted space.
+
+    If restrict == False and Dirichlet boundary conditions are supplied, then
+    these conditions will result in the eigenproblem having a nullspace spanned
+    by the basis functions with support on the boundary. To facilitate
+    solution, this is shifted by the specified amount. It is the user's
+    responsibility to ensure that the shift is not close to an actual
+    eigenvalue of the system.
+
+    Also, we assume the operator we linearise maps from V to V.
```
JHopeCollins, line 101: This warning should be in the docstring for the reduced functional. See my suggestion above.

firedrake/rf_eigensolver.py
```diff
@@ -0,0 +1,661 @@
...
+    def __init__(
+        self,
+        rf: ReducedFunctional,
+        M: MassType = None,
+        bcs: BcType = None,
+        bc_shift: float = 0.0,
+        restrict: bool = True,
+        identity: bool = False,
+        apply_riesz: bool = True,
+        action: str = "tlm",
```
JHopeCollins, line 113: We have the enum type, so we should use it!
```suggestion
        action: str = TLMAction,
```

firedrake/rf_eigensolver.py
```diff
@@ -0,0 +1,661 @@
...
+        if not SLEPc:
+            raise ImportError(
+                "Unable to import SLEPc, eigenvalue computation not possible "
+                "(see https://www.firedrakeproject.org/install.html#slepc)"
+            )
+
+        if not isinstance(rf, ReducedFunctional) and not isinstance(M, ReducedFunctional):
+            raise TypeError(
+                "At least A or M in the GEP has to be a ReducedFunctional instance. "
+                "Use 'LinearEigenproblem' otherwise."
```
JHopeCollins, line 126: GEP?

firedrake/rf_eigensolver.py
```diff
@@ -0,0 +1,661 @@
...
+        ----------
+        action :
+            String representing the forward action of the matrix. 'tlm' or 'adjoint'.
+
+        Returns
+        -------
+        Optional[TLMAction, AdjointAction]
+            Action corresponding to the string.
+        """
+        match action:
```
JHopeCollins, line 228:
```suggestion
        from pyadjoint.optimization.tao_solver import RFAction
        if isinstance(action, RFAction):
            return action
        match action:
```

firedrake/rf_eigensolver.py
```diff
@@ -0,0 +1,661 @@
...
+            return ReducedFunctionalMat(
+                obj, action=self.action, apply_riesz=self.apply_riesz, appctx=self.appctx, comm=self.comm
+            )
+        return assemble(
+            obj,
+            bcs=self.bcs,
+            weight=self.bc_shift and 1.0 / self.bc_shift,
+        ).petscmat
+
+    def match_forward_action(self, action: str):
```
JHopeCollins, line 214:
```suggestion
    def match_forward_action(self, action: Union[RFAction, str]):
```

firedrake/rf_eigensolver.py
```diff
@@ -0,0 +1,661 @@
...
+        Match forward action of matrix using passed string.
+
+        Parameters
+        ----------
+        action :
+            String representing the forward action of the matrix. 'tlm' or 'adjoint'.
+
+        Returns
+        -------
+        Optional[TLMAction, AdjointAction]
```
JHopeCollins, line 225:
```suggestion
        RFAction
```

firedrake/rf_eigensolver.py
```diff
@@ -0,0 +1,661 @@
...
+        Optional[None, PETSc.Mat]
+            `SLEPc`-compatible operator representing obj.
+        """
+        if isinstance(obj, Form):
+            args = obj.arguments()
+            v, u = args
+            u_res = TrialFunction(V_res)
+            v_res = TestFunction(V_res)
+            obj = replace(obj, {u: u_res, v: v_res})
+        return self.as_petsc_mat(obj) if obj else None
```
JHopeCollins, line 177: Would `obj` ever be `None`?
```suggestion
        return self.as_petsc_mat(obj)
```

firedrake/rf_eigensolver.py
```diff
@@ -0,0 +1,661 @@
...
+        *,
+        apply_riesz: bool = False,
+        appctx: AppCtxType = None,
+        comm: Comms = PETSc.COMM_WORLD,
+        restricted_space: Optional[WithGeometry] = None,
+    ):
+        if restricted_space is None:
+            raise ValueError(
+                "restricted_space must be provided for RestrictedReducedFunctionalMatCtx."
+            )
```
JHopeCollins, line 538: Either `restricted_space` is optional or it isn't! The docstring, type hints, and error raising need to be consistent.

firedrake/rf_eigensolver.py
```diff
@@ -0,0 +1,661 @@
...
+            PETScVecInterface(
+                [Function(self.restricted_space).interpolate(c.control) for c in rf.controls]
+            )
+            if restricted_space
+            else self.control_interface
+        )
+
+        if action in (AdjointAction, TLMAction):
+            self.restricted_functional_interface = (
+                PETScVecInterface(Function(restricted_space).interpolate(rf.functional), comm=comm)
```
JHopeCollins, line 556:
```suggestion
                PETScVecInterface(Function(self.restricted_space).interpolate(rf.functional), comm=comm)
```

firedrake/rf_eigensolver.py
```diff
@@ -0,0 +1,661 @@
...
+
+        elif action is AdjointAction:
+            self.xresinterface = self.restricted_functional_interface
+            self.yresinterface = self.restricted_control_interface
+            self.xres = (
+                Function(self.restricted_space)
+                .interpolate(rf.functional)
+                ._ad_copy()
+                ._ad_init_zero(dual=True)
+            )
```
JHopeCollins, line 575:
```suggestion
            self.xres = Function(self.restricted_space.dual())
```

firedrake/rf_eigensolver.py
```diff
@@ -0,0 +1,661 @@
...
+        elif action is TLMAction:
+            self.xresinterface = self.restricted_control_interface
+            self.yresinterface = self.restricted_functional_interface
+            self.xres = new_restricted_control_variable(rf, self.restricted_space)
+            self.yres = (
+                Function(self.restricted_space)
+                .interpolate(rf.functional)
+                ._ad_copy()
+                ._ad_init_zero(dual=True)
+            )
```
JHopeCollins, line 587:
```suggestion
            self.yres = Function(self.restricted_space.dual())
```

firedrake/rf_eigensolver.py
JHopeCollins: Maybe, but the Problem/Solver distinction seems like an implementation detail. I've never seen anyone do anything with a Problem except immediately create a Solver, so it seems like just an annoyance to expose to users (happy to be corrected if there is a valid use for a standalone Problem).

firedrake/rf_eigensolver.py
```diff
@@ -0,0 +1,661 @@
...
+        out = self.mult_impl(A, interpolated_x)
+        interpolated_y = interpolate_vars(out, self.restricted_space)
+        self.yresinterface.to_petsc(y, interpolated_y)
+        if self._shift != 0:
+            y.axpy(self._shift, x)
+
+    def multTranspose(self, A, x, y):
+        self.yresinterface.from_petsc(x, self.yres)
+        interpolated_x = interpolate_vars(self.yres, self.function_space)
+        out = self.mult_impl_transpose(A, interpolated_x)
```
JHopeCollins, line 601: Should there be a pyadjoint PR to merge into [this one](https://github.com/dolfin-adjoint/pyadjoint/pull/213) that implements `ReducedFunctionalMat.mult_impl_transpose`?

firedrake/rf_eigensolver.py
```diff
@@ -0,0 +1,661 @@
...
+        else:
+            return self.output_space.dm
+
+
+class RFEigensolver:
+    r"""Solves an RFEigenproblem.
+    This is a generalisation of the linear eigenproblem which is
+    additionally able to cope with eigenproblems arising from
+    linearisations of operators represented by sequences of
+    Firedrake operations.
```
JHopeCollins, line 310:
```suggestion
    This is a generalisation of the linear eigenproblem which can
    solve eigenproblems arising from linearisations of sequences of
    Firedrake operations recorded in a ReducedFunctional.
```

firedrake/rf_eigensolver.py
```diff
@@ -0,0 +1,661 @@
...
+        `SLEPc.EPS.setDimensions`.
+
+    Notes
+    -----
+    Users will typically wish to set solver parameters specifying the symmetry
+    of the eigenproblem and which eigenvalues to search for first.
+
+    The former is set using the options available for `EPSSetProblemType
+    <https://slepc.upv.es/documentation/current/docs/manualpages/EPS/EPSSetProblemType.html>`__.
+
```
JHopeCollins, line 336:
```suggestion
```

firedrake/rf_eigensolver.py
```diff
@@ -0,0 +1,661 @@
...
+    formulation are implicit (`PETSc.Mat.PYTHON` type), which restricts the available 
+    options to those that do not use the explicit matrix entries. For instance, spectral 
+    transforms won't work and solvers are restricted. 
+
+    As always when specifying PETSc options, `None` indicates that the option
+    in question is a flag and hence doesn't take an argument.
+
+    The eigenvalues to search for first are specified using the options for
+    `EPSSetWhichEigenPairs <https://slepc.upv.es/documentation/current/docs/manualpages/EPS/EPSSetWhichEigenpairs.html>`__.
+
```
JHopeCollins, line 352:
```suggestion
```

firedrake/rf_eigensolver.py
```diff
@@ -0,0 +1,661 @@
...
+        self._problem = problem
+        self.n_evals = n_evals
+        self.ncv = ncv
+        self.mpd = mpd
+        solver_parameters = flatten_parameters(solver_parameters or {})
+        for key in self.DEFAULT_EPS_PARAMETERS:
+            value = self.DEFAULT_EPS_PARAMETERS[key]
+            solver_parameters.setdefault(key, value)
+        self.options_manager = OptionsManager(solver_parameters, options_prefix)
+        self.options_manager.set_from_options(self.es)
```
JHopeCollins, line 385: This is now the recommended way to handle solver options.
You'll need to import `attach_options`, `set_from_options`, `set_default_parameter`, and `inserted_options` from `petsctools`.
```suggestion
        attach_options(self.es, solver_parameters or {}, options_prefix)
        for key, value in self.DEFAULT_EPS_PARAMETERS:
            set_default_parameter(self.es, key, val)
        set_from_options(self.es)        
```

firedrake/rf_eigensolver.py
```diff
@@ -0,0 +1,661 @@
...
+        M = self._problem.M
+        if not isinstance(A, PETSc.Mat):
+            raise TypeError(f"A must be a PETSc matrix, got {A.__class__}")
+        if not isinstance(M, PETSc.Mat) and (M is not None):
+            raise TypeError(f"M must be a PETSc matrix or None (identity), got {M.__class__}")
+
+        self.es.setDimensions(nev=self.n_evals, ncv=self.ncv, mpd=self.mpd)
+        self.es.setOperators(A, M)
+        with self.options_manager.inserted_options():
+            self.es.solve()
```
JHopeCollins, line 406:
```suggestion
        with inserted_options(self.es):
            self.es.solve()
```

firedrake/rf_eigensolver.py
```diff
@@ -0,0 +1,661 @@
...
+        """
+        A = self._problem.A
+        M = self._problem.M
+        if not isinstance(A, PETSc.Mat):
+            raise TypeError(f"A must be a PETSc matrix, got {A.__class__}")
+        if not isinstance(M, PETSc.Mat) and (M is not None):
+            raise TypeError(f"M must be a PETSc matrix or None (identity), got {M.__class__}")
+
+        self.es.setDimensions(nev=self.n_evals, ncv=self.ncv, mpd=self.mpd)
+        self.es.setOperators(A, M)
```
JHopeCollins, line 395: The operators should be set in `__init__`, before the call to `set_from_options`. This is so that:

1. The solver actually has access to the Mats when everything is being set up during `set_from_options`,
2. We don't incorrectly/unnecessarily reset the operators every time we call `solve`.

firedrake/rf_eigensolver.py
```diff
@@ -0,0 +1,661 @@
...
+        if isinstance(variables[0], Function):
+            return tuple(Function(function_space).interpolate(v) for v in variables)
+        else:
+            return tuple(
+                Cofunction(function_space.dual()).interpolate(v) for v in variables
+            )
+    elif isinstance(variables, Function):
+        return Function(function_space).interpolate(variables)
+    else:
+        return Cofunction(function_space.dual()).interpolate(variables)
```
JHopeCollins, line 500: You can use `pyadjoint.Enlist` to handle whether we have a list or not. I think this is the right way to do it.
```suggestion
    from ufl.duals import is_primal
    variables = Enlist(variables)
    V = function_space if is_primal(variables[0]) else function_space.dual()
    interp_vars = [Function(V).interpolate(v) for v in variables)
    return variables.delist(interp_vars)
```

firedrake/rf_eigensolver.py
```diff
@@ -0,0 +1,661 @@
...
+        dual: whether to return a dual type. If False then a primal type is returned.
+
+    Returns
+    -------
+        tuple[OverloadedType]: New variables suitable for storing a control value.
+    """
+    return tuple(
+        Function(function_space).interpolate(control.control)._ad_init_zero(dual=dual)
+        for control in reduced_functional.controls
+    )
```
JHopeCollins: I don't think there's any need for the interpolation if we're just asking for a zero-initialised object with `_ad_init_zero`, and `Function` is already zero-initialised
```suggestion
    return tuple(
        Function(function_space.dual() if dual else function_space)
        for control in reduced_functional.controls
    )
```

firedrake/rf_eigensolver.py
```diff
@@ -0,0 +1,661 @@
...
+def interpolate_vars(variables: VarsInterpolate, function_space: WithGeometry):
+    """
+    Interpolates primal/dual variables to restricted/unrestricted function spaces.
+
+    Parameters
+    ----------
+        variables:
+            Variables that are to be interpolated into unrestricted/restricted primal/dual spaces.
+        restricted_space:
+            The function space where `PETSc.Vec` objects will live.
```
JHopeCollins: This has a different name to the function signature.

tests/firedrake/adjoint/test_rf_eigensolver.py
JHopeCollins, line 104: Seeing as we're specifically testing the implicit matrix here, I think it'd be simpler to just construct them explicitly rather than going via the RFEigenProblem.

firedrake/rf_eigensolver.py
Shawniz: I agree, this is more efficient. Weird I didn't use dual().

firedrake/rf_eigensolver.py
Shawniz: Agreed, this change could probably be mirrored in `eigensolver.py` in the `LinearEigensolver` class.

tests/firedrake/adjoint/test_rf_eigensolver.py
Shawniz: I have tried but faced a bit of obstacles. The aim of the test is to check that two eigenvalue abstractions outputs the same results.

tests/firedrake/adjoint/test_rf_eigensolver.py
Shawniz: This is done because the eigenvalue problem in `LinearEigenproblem` arises from the spatial discretisation of the bilinear forms arising from the Helmholtz PDE with forcing function `f=cu` for eigenvalue `c`. The solution operator in this case is a mapping from the forcing function to the solution. A wave-of-hands argument is that `Jf=u` with the `f` substituted from above, and using the fact that J is linear (which it is as the PDE is steady and linear) rearranges to `Ju = 1/cu'.

tests/firedrake/adjoint/test_rf_eigensolver.py
Shawniz: I actually thought that the code is less messy in this case. Otherwise, would have to do casework depending on restricted an unrestricted stuff. Also, would have to add functions to determine restricted space. Much more clunky whilst this happens under the hood of `RFEigenproblem`.

firedrake/rf_eigensolver.py
JHopeCollins: Yes, if they are also set in `solve` in the `LinearEigenSolver` then that is also a bug.

They also need to be _before_ the call to `set_from_option` in the init to make sure that it's all set up correctly.

PR #4531, opened 2025-09-01
Remove `Interpolator.interpolate`

leo-collins:
Removes `Interpolator.interpolate` method in favour of the `interpolate` function.


dham (commented):
For now, don't rename `._interpolate` but instead make `.interpolate` a hard error telling users how to change their code.

Then rename `._interpolate` for the next release cycle (i.e. merge to main after the October 2025 release).

PR #4532, opened 2025-09-02
FixAtPointBC: support bendy meshes

pbrubeck:
# Description
Removes the auxiliary P1 mesh used by `FixAtPointBC` in the demos, as interpolation into VOM supports bendy meshes.


pbrubeck:
Passes tests, except for a netgen adaptivity test

Issue #4530, opened 2025-09-01
BUG: `assemble(Constant * Cofunction(Function))` fails


jrmaddison:
**Describe the bug**
Assembly of a multiple of a cofunction action fails.

**Steps to Reproduce**

```
import firedrake as fd

mesh = fd.UnitIntervalMesh(1)
space = fd.FunctionSpace(mesh, "DG", 0)
u = fd.Function(space, name="u")
b = fd.assemble(fd.conj(fd.TestFunction(space)) * fd.dx)

fd.assemble(b(u))  # OK
fd.assemble(2 * b(u))  # OK
fd.assemble(fd.Constant(2) * b(u))  # Fails
```
leads to
```
Traceback (most recent call last):
  File "/home/jmaddis2/build/firedrake/examples/assemble_bug.py", line 10, in <module>
    fd.assemble(fd.Constant(2) * b(u))  # Fails
    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "petsc4py/PETSc/Log.pyx", line 250, in petsc4py.PETSc.Log.EventDecorator.decorator.wrapped_func
  File "petsc4py/PETSc/Log.pyx", line 251, in petsc4py.PETSc.Log.EventDecorator.decorator.wrapped_func
  File "/home/jmaddis2/build/firedrake/firedrake/firedrake/adjoint_utils/assembly.py", line 58, in wrapper
    output.form = form
    ^^^^^^^^^^^
AttributeError: 'Zero' object has no attribute 'form'
```

pbrubeck:
It seems that we are trying to be clever and simplifying for the case when the result is zero, which happens when either `u` or `b` are zero. Setting them to nonzero gives no error:
```
from firedrake import *
mesh = UnitIntervalMesh(1)
space = FunctionSpace(mesh, "DG", 0)
u = Function(space).assign(1)
b = Cofunction(space.dual()).assign(1)

assemble(Constant(2) * action(b, u))
```

pbrubeck:
Passing a `tensor` seems to fix the issue:
```
from firedrake import *
mesh = UnitIntervalMesh(2)
space = FunctionSpace(mesh, "DG", 0)
u = Function(space).assign(1)
b = Cofunction(space.dual()).assign(1)

tensor = Constant(0)
assemble(Constant(2) * action(b, u), tensor=tensor)
```

PR #4534, opened 2025-09-02
BaseFormAssembler: cast FormSum weights as numbers

pbrubeck:
# Description
Fixes https://github.com/firedrakeproject/firedrake/issues/4530

### DIFF
--- a/tests/firedrake/regression/test_assemble_baseform.py
+++ b/tests/firedrake/regression/test_assemble_baseform.py
@@ -134,8 +134,8 @@ def test_scalar_formsum(f, scale):
     elif scale == "Real":
         mesh = f.function_space().mesh()
         R = FunctionSpace(mesh, "R", 0)
-        s1 = Function(R).assign(s1)
-        s2 = Function(R).assign(s2)
+        s1 = Function(R, val=s1)
+        s2 = Function(R, val=s2)
 
     formsum = s1 * q + s2 * q
     assert isinstance(formsum, ufl.form.FormSum)


tests/firedrake/regression/test_assemble_baseform.py
```diff
@@ -112,14 +112,32 @@ def test_assemble_action(M, f):
...
+        s1 = as_ufl(s1)
+        s2 = as_ufl(s2)
+    elif scale == "Constant":
+        s1 = Constant(s1)
+        s2 = Constant(s2)
+    elif scale == "Real":
+        mesh = f.function_space().mesh()
+        R = FunctionSpace(mesh, "R", 0)
+        s1 = Function(R, val=s1)
+        s2 = Function(R, val=s2)
```
jrmaddison, line 138: I'm not really sure an R0 weight makes sense here -- at least the result would probably also need to be R0. But maybe that's not for this PR.

jrmaddison (approved):

tests/firedrake/regression/test_assemble_baseform.py
pbrubeck: I think we need to support this case if we want to be able to differentiate/take adjoints with respect to weights in a `FormSum`.

jrmaddison: Makes sense. Implicitly mapping from function to field just feels a bit odd

```
one = fd.assemble(unity_r0_mesh0 * action_evaluating_to_one)
fd.assemble(unity_r0_mesh0 * fd.dx(mesh1))  # Volume of mesh0/mesh1 intersection?
fd.assemble(one * fd.dx(mesh1))  # Volume of mesh1?
```

pbrubeck: Does the second line from the code above even run?

pbrubeck: Wouldn't the R space be most naturally extended into other domains as the same constant?

jrmaddison: > Does the second line from the code above even run?

No, I see that it doesn't

> Wouldn't the R space be most natrually extended into other domains as the same constant?

I wouldn't say so, it's defined on a domain.

Sorry for the distraction, this isn't an issue for this PR.

PR #4537, opened 2025-09-03
Fix docs build

connorjward:
We were getting undefined references due to an incompletely qualified path and an upstream change to UFL
(https://github.com/FEniCS/ufl/pull/387/files) that is incompatible with Sphinx (https://github.com/sphinx-doc/sphinx/issues/11225). The Sphinx issue is still unresolved so the best course of action is to ignore the issue for now.

### DIFF
--- a/docs/source/conf.py
+++ b/docs/source/conf.py
@@ -149,6 +149,9 @@ nitpick_ignore_regex = [
     ('py:class', 'firedrake.solving_utils._SNESContext'),
     # Pyadjoint BlockVariable is not properly referenced in pyadjoint
     ('py:class', 'BlockVariable'),
+    # Cofunction.ufl_domains references FormArgument but it isn't picked
+    # up by Sphinx (see https://github.com/sphinx-doc/sphinx/issues/11225)
+    ('py:class', 'FormArgument'),
 ]
 
 # Dodgy links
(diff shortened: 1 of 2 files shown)


docs/source/conf.py
```diff
@@ -149,6 +149,9 @@
     ('py:class', 'firedrake.solving_utils._SNESContext'),
     # Pyadjoint BlockVariable is not properly referenced in pyadjoint
     ('py:class', 'BlockVariable'),
+    # Cofunction.ufl_domains references FormArgument but it isn't picked
+    # up by Sphinx (see https://github.com/sphinx-doc/sphinx/issues/11225)
+    ('py:class', 'FormArgument'),
```
JHopeCollins, line 154: Please can you open a UFL issue for this, and a corresponding Firedrake issue to hopefully remove this once it's fixed?

JHopeCollins (changes requested):

docs/source/conf.py
connorjward: I'm not quite sure that opening a UFL issue is the way to go. I looked into it and they already have [big Sphinx issues](https://github.com/FEniCS/ufl/issues/415). This is merely a small technical point that doesn't really make sense for them to consider until the wider issues are fixed.

connorjward: I've opened https://github.com/firedrakeproject/firedrake/issues/4546 in Firedrake and https://github.com/FEniCS/ufl/issues/415 in UFL (which is a broader issue around their docs CI that should fix this as part of it). If you're happy (and assuming the extra fix I just pushed works) then can this go in?

JHopeCollins: Great, thanks

PR #4538, opened 2025-09-03
MG: rediscretize with a different coarse_mat_type

pbrubeck:
# Description
Allows rediscretization of the coarse level with a different matrix type to prevent nesting with `AssembledPC` when using matfree on fine levels.

### DIFF
--- a/firedrake/mg/ufl_utils.py
+++ b/firedrake/mg/ufl_utils.py
@@ -276,18 +276,22 @@ def coarsen_snescontext(context, self, coefficient_mapping=None):
         else:
             solver_prefix = "mg_"
 
-        mat_type = opts.getString(f"{solver_prefix}coarse_mat_type",
-                                  default=context.mat_type)
-        pmat_type = context.pmat_type if context.mat_type == mat_type else mat_type
-        pmat_type = opts.getString(f"{solver_prefix}coarse_pmat_type",
-                                   default=pmat_type)
+        coarse_mat_type = opts.getString(f"{solver_prefix}coarse_mat_type", "")
+        if coarse_mat_type == "":
+            coarse_mat_type = context.mat_type
+            default_pmat_type = context.pmat_type
+        else:
+            default_pmat_type = coarse_mat_type
+
+        coarse_pmat_type = opts.getString(f"{solver_prefix}coarse_pmat_type",
+                                          default_pmat_type)
     else:
-        mat_type = context.mat_type
-        pmat_type = context.pmat_type
+        coarse_mat_type = context.mat_type
+        coarse_pmat_type = context.pmat_type
 
     coarse = type(context)(problem,
-                           mat_type=mat_type,
-                           pmat_type=pmat_type,
+                           mat_type=coarse_mat_type,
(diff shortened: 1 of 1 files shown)


firedrake/mg/ufl_utils.py
```diff
@@ -266,10 +266,23 @@ def coarsen_snescontext(context, self, coefficient_mapping=None):
...
+    _, level = utils.get_level(problem.u_restrict.function_space().mesh())
+    if level == 0:
+        # Use different mat_type on coarsest level
+        opts = PETSc.Options(context.options_prefix)
+        mat_type = opts.getString("mg_coarse_mat_type", default=context.mat_type)
+        pmat_type = context.pmat_type if context.mat_type == mat_type else mat_type
+        pmat_type = opts.getString("mg_coarse_pmat_type", default=pmat_type)
+    else:
+        mat_type = context.mat_type
+        pmat_type = context.pmat_type
```
JHopeCollins: This is really nice to have.
Is there any reason not to do it for levels too?

pbrubeck: There is only one coarse level. If we add support for `mg_levels_mat_type` then the finest level will always use the mat_type from the problem and the intermidiate levels will use a different one. We cannot easily set up two different relaxations for the the finest level and the intermidiate levels, it is possible but it requires passing prefixes. We could probably support `mg_levels_k_mat_type`, but is this any useful?

connorjward: If we can be more flexible I think that would be better.

pbrubeck: It seems that if  I add that one here it will probably require new tests and make the PR way more involved. So maybe this point needs more discusion and merits its own PR.

JHopeCollins: I think that's fair enough. The biggest gain is what you already have here, and it looks basically ready to go. I'd be happy with this going through and leaving levels for another PR. Realistically it's a pretty niche case anyway.

firedrake/mg/ufl_utils.py
```diff
@@ -266,10 +266,30 @@ def coarsen_snescontext(context, self, coefficient_mapping=None):
...
+        # Use different mat_type on coarsest level
+        opts = PETSc.Options(context.options_prefix)
+        if opts.getString("snes_type", "") == "fas":
+            solver_prefix = "fas_"
+        else:
+            solver_prefix = "mg_"
+
+        mat_type = opts.getString(f"{solver_prefix}coarse_mat_type",
+                                  default=context.mat_type)
+        pmat_type = context.pmat_type if context.mat_type == mat_type else mat_type
```
connorjward: This is some really obscure logic. Probably needs a comment or refactor.

pbrubeck: @connorjward Is this better now?

connorjward (approved):
Thanks, that is definitely a lot clearer.

firedrake/mg/ufl_utils.py
connorjward: Yep. Thanks.

PR #4539, opened 2025-09-03
Workflow changes now we have the new containers

connorjward:
~* Install inkscape and texlive-full into developer containers.~
* Remove TestPyPI part of core.yml as it was unused.
* Move docs build into separate job to reduce CI durations.

### DIFF
--- a/docs/Makefile
+++ b/docs/Makefile
@@ -57,7 +57,7 @@ livehtml:
 	sphinx-autobuild -b html $(ALLSPHINXOPTS) $(BUILDDIR)/html
 
 serve: html
-	cd $(BUILDDIR)/html; python -m SimpleHTTPServer $(PORT)
+	cd $(BUILDDIR)/html; python3 -m SimpleHTTPServer $(PORT)
 
 .PHONY: copy_demos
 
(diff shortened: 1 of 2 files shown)


connorjward:
> > * Install inkscape and texlive-full into developer containers.
> 
> Are you sure you really want to do that? `texlive-full` is huge, its size is measured in GB...

This is a good point. I wanted to do this so that we wouldn't need to `apt install` it every time we build the docs. The proper solution is to have another Docker image that is the dev container + texlive. I'll try that.

connorjward:
> The proper solution is to have another Docker image that is the dev container + texlive. I'll try that.

To be nice this requires a chunkier refactor that doesn't belong in this PR. I won't attempt it here.

docs/Makefile
```diff
@@ -57,7 +57,7 @@ livehtml:
 	sphinx-autobuild -b html $(ALLSPHINXOPTS) $(BUILDDIR)/html
 
 serve: html
-	cd $(BUILDDIR)/html; python -m SimpleHTTPServer $(PORT)
+	cd $(BUILDDIR)/html; python3 -m SimpleHTTPServer $(PORT)
```
francesco-ballarin, line 60: FYI, debian has a `python-is-python3` package, and surely ubuntu has that too.

dham (dismissed):

PR #4541, opened 2025-09-04
Some fixes for `Cofunction` controls

jrmaddison:
Fixes which allow `Cofunction` controls to be used in some configurations.

tao_type_nls is xfailed as this leads to an obscure error within UFL.

### DIFF
--- a/firedrake/adjoint_utils/function.py
+++ b/firedrake/adjoint_utils/function.py
@@ -342,9 +342,10 @@ class CofunctionMixin(FunctionMixin):
     def _ad_dot(self, other):
         return firedrake.assemble(firedrake.action(self, other))
 
+    @classmethod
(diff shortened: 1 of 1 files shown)


firedrake/adjoint_utils/function.py
```diff
@@ -341,3 +341,14 @@ class CofunctionMixin(FunctionMixin):
 
     def _ad_dot(self, other):
         return firedrake.assemble(firedrake.action(self, other))
+
+    def _ad_init_object(cls, obj):
+        from firedrake import Cofunction
+        return Cofunction(cls.function_space()).assign(obj)
```
JHopeCollins: Why `cls` and not `self` here?

jrmaddison: Yes, and should be a `classmethod`

Issue #4478, opened 2025-07-31
BUG: Occasional CI failure for point location


connorjward:
We are occasionally seeing `tests/firedrake/regression/test_interpolate_cross_mesh.py::test_interpolate_cross_mesh_parallel[circlemanifold_to_high_order]` fail in our CI runs (e.g. https://github.com/firedrakeproject/firedrake/actions/runs/16651274433/job/47124382938?pr=4477) with the error message:
```
RuntimeError: Point evaluation gave different results across processes.
```

@leo-collins FYI

leo-collins:
Also: https://github.com/firedrakeproject/firedrake/issues/3080

leo-collins:
It's on my bucket list. Also there's another test here with a similar comment. Have you ever seen this one fail?

https://github.com/firedrakeproject/firedrake/blob/d86424649ec56195dfdc6d137f300bb2331caa1d/tests/firedrake/regression/test_interpolate_cross_mesh.py#L221-L239

Issue #3080, opened 2023-08-29
Have `Function.at` make a vertex-only mesh at the specified points


ReubenHill:
# What we want

`VertexOnlyMesh` has a much better point location algorithm than `Function.at`. Without this algorithm, `.at` regularly produces a runtime error https://github.com/firedrakeproject/firedrake/blob/2a388c38fd46beb4201e99fcdfd8df5e3b4eee51/firedrake/function.py#L622
which can only be fixed by reducing the mesh tolerance parameter.

The internals of `.at` should be replaced by code which creates a `VertexOnlyMesh` at the specified points and returns the values found by interpolating onto a P0DG function space at the point locations. A bunch of old point location code can then be removed.

# What's blocking this

 - [x] `VertexOnlyMesh` cannot yet be created in an immersed manifold but as of #2879 , `.at` works for them. I'm working on fixing this in #3068 .
 - [ ]  Similarly, `VertexOnlyMesh` cannot yet be created in a high order mesh but `.at` will happily produce a value at a given point. Fixing this needs a more advanced voting algorithm. From the `VertexOnlyMesh` code: https://github.com/firedrakeproject/firedrake/blob/2a388c38fd46beb4201e99fcdfd8df5e3b4eee51/firedrake/mesh.py#L2822-L2826

# Possible intermediate solution

We can either hold off on doing this until the above features have been added or we can implement this for cases where this does work and keep the old code for where it doesn't. Eventually we will have feature parity and can remove the old code.

leo-collins:
Solved by https://github.com/firedrakeproject/firedrake/pull/4543

Issue #2790, opened 2023-02-27
Point evaluation failure in parallel


jrmaddison:
The following example

```
from firedrake import *

N_x, N_y, N_z = (2, 2, 2)

mesh = UnitCubeMesh(N_x, N_y, N_z)
X = SpatialCoordinate(mesh)
space = FunctionSpace(mesh, "Lagrange", 1)

F = Function(space, name="F")
F.interpolate(exp(X[0]) * exp(X[1]))

F((0.2, 0.3, 0.4))
```

leads to a failure when run on 4 processes

```
Traceback (most recent call last):
  File "[...]/test.py", line 12, in <module>
Traceback (most recent call last):
  File "[...]/test.py", line 12, in <module>
Traceback (most recent call last):
  File "[...]/test.py", line 12, in <module>
Traceback (most recent call last):
    F((0.2, 0.3, 0.4))
    F((0.2, 0.3, 0.4))
  File "[...]/build/firedrake/firedrake/src/ufl/ufl/exproperators.py", line 330, in _call
  File "[...]/test.py", line 12, in <module>
  File "[...]/build/firedrake/firedrake/src/ufl/ufl/exproperators.py", line 330, in _call
    F((0.2, 0.3, 0.4))
  File "[...]/build/firedrake/firedrake/src/ufl/ufl/exproperators.py", line 330, in _call
    return _eval(self, arg, mapping, component)
    return _eval(self, arg, mapping, component)
  File "[...]/build/firedrake/firedrake/src/ufl/ufl/exproperators.py", line 320, in _eval
  File "[...]/build/firedrake/firedrake/src/ufl/ufl/exproperators.py", line 320, in _eval
    return _eval(self, arg, mapping, component)
  File "[...]/build/firedrake/firedrake/src/ufl/ufl/exproperators.py", line 320, in _eval
    return f.evaluate(coord, mapping, component, index_values)
    return f.evaluate(coord, mapping, component, index_values)
  File "[...]/build/firedrake/firedrake/src/firedrake/firedrake/function.py", line 488, in evaluate
  File "[...]/build/firedrake/firedrake/src/firedrake/firedrake/function.py", line 488, in evaluate
    return f.evaluate(coord, mapping, component, index_values)
  File "[...]/build/firedrake/firedrake/src/firedrake/firedrake/function.py", line 488, in evaluate
    return self.at(coord)
[...]

PR #4543, opened 2025-09-04
deprecate `.at()`

leo-collins:
This PR deprecates the `Function.at()` method in favour of the `PointEvaluator` class introduced in the PR above. Using `.at()` now shows a warning telling users to use a `PointEvaluator`. The docs for `.at()` are removed. All tests are updated to use `PointEvaluator` where necessary.

The only place we use `.at` still is in the `streamline` function; see https://github.com/firedrakeproject/firedrake/issues/4573.

This solves issues https://github.com/firedrakeproject/firedrake/issues/3080, https://github.com/firedrakeproject/firedrake/issues/2790, https://github.com/firedrakeproject/firedrake/issues/4478


docs/source/point-evaluation.rst
```diff
@@ -242,6 +151,44 @@ warning or switched off entirely:
    :end-before: [test_vom_manual_points_outside_domain 6]
 
 
+``PointEvaluator`` convenience object
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+The :py:class:`~.PointEvaluator` class performs point evaluation using vertex-only meshes,
```
connorjward, line 145: I think this section should contain more justification about why this is better than the above approach.

connorjward (changes requested):

docs/source/point-evaluation.rst
leo-collins: I can add this to https://github.com/firedrakeproject/firedrake/pull/4516

docs/source/point-evaluation.rst
```diff
@@ -414,6 +368,90 @@ be created. Any existing immersed vertex-only mesh will have been created
 using the previous tolerance and will be unaffected by the change.
 
 
+``Function.at``
```
leo-collins: Remove this from docs

PR #4544, opened 2025-09-04
`AuxiliaryOperatorSNES` & `FieldsplitSNES`

JHopeCollins:
# Description
Some Python type SNES classes for building nonlinear solvers

## `AuxiliaryOperatorSNES`

A nonlinear version of `AuxiliaryOperatorPC`, for building nonlinear preconditioners from approximate operators, e.g. preconditioned nonlinear Richardson iterations.

Still to do:
- [x] Docstring
- [x] Tests
- [ ] Demo?

## `FieldsplitSNES`

A nonlinear version of `PCFieldsplit`,it can use either additive (Jacobi) or multiplicative (Gauss-Seidel) versions, exactly analogous to the linear version.
- Additive solves each "diagonal" block of the nonlinear problem for each field, with all other fields held at the current value.
- Multiplicative solves each "diagonal" block of the nonlinear problem for each field, and uses the updated values when solving for subsequent fields. Currently solves one field at a time, in the order they appear in the mixed space.

Still to do (mostly adding options to match PCFieldsplit):
- [x] Docstring 
- [x] Allow grouping fields together with `snes_fieldsplit_%d_fields`.
- [x] Allow setting the order of the field solves with  `snes_fieldsplit_%d_fields`.
- [ ] Default fields with `snes_fieldsplit_default`.
- [ ] Default options with `fieldsplit` prefix. If we collect them in `initialize` and pass them as a dictionary to the NLVS of each field, they will take lower priority than any `fieldsplit_%d` options in the global dictionary (including those put there by the parent NLVS).
- [x] `view` method.
- [ ] Does the `update` method need to contain anything?
- [ ] Documentation and demo.
- [x] Tests.


firedrake/preconditioners/auxiliary_snes.py
```diff
@@ -0,0 +1,133 @@
...
+            The current state of the outer SNES
+        func : firedrake.Function
+            The solution function.
+        test : ufl.TestFunction
+            The test function.
+
+        Returns
+        -------
+        F : ufl.Form
+            The preconditioning nonlinear form.
```
pbrubeck:
```suggestion
            The preconditioning linear Form (may be nonlinear in func).
```

firedrake/preconditioners/auxiliary_snes.py
```diff
@@ -0,0 +1,133 @@
...
+
+        self.uk1.assign(self.uk)
+        self.solver.solve()
+
+        with self.uk1.dat.vec_ro as vec:
+            vec.copy(y)
+            y.aypx(-1, x)
+
+    def form(self, snes, state, func, test):
+        """Return the preconditioning nonlinear form and boundary conditions.
```
pbrubeck:
```suggestion
        """Return the preconditioning residual Form and boundary conditions.
```

JHopeCollins:
@miguelcoolchips Are you interested in the `AuxiliaryOperatorSNES` or the `FieldsplitSNES`? Or both? I think the `AuxiliaryOperatorSNES` doesn't need too much more to get it to a reviewable state, but the `FieldsplitSNES` still has some kinks to work out. I just ran out of time to work on this around other things.

miguelcoolchips:
> @miguelcoolchips Are you interested in the `AuxiliaryOperatorSNES` or the `FieldsplitSNES`? Or both? I think the `AuxiliaryOperatorSNES` doesn't need too much more to get it to a reviewable state, but the `FieldsplitSNES` still has some kinks to work out. I just ran out of time to work on this around other things.

We are interested in `FieldsplitSNES`.

firedrake/preconditioners/auxiliary_snes.py
```diff
@@ -0,0 +1,234 @@
...
+    @PETSc.Log.EventDecorator()
+    def step(self, snes, x, f, y):
+        with self.uk.dat.vec_wo as vec:
+            x.copy(vec)
+
+        self.assemble_gk(tensor=self.b)
+
+        if f is not None:
+            with self.b.dat.vec as vec:
+                vec -= f
```
JHopeCollins:
```suggestion
        with self.b.dat.vec as vec:
            vec -= f
```

PR #4545, opened 2025-09-05
Use appropriate user in CI to avoid permissions errors on later runs

connorjward:
This should fix the permissions issues that we saw building Docker images on CI.

The issue was that the regular CI runs are run in a Docker container as the root user and this means that if a job is cancelled then a bunch of files are left behind that are owned by `root`. If we subsequently build a Docker image on the same runner we would crash because that workflow is run as an unprivileged user.

I tried to make the regular CI runs use an unprivileged user but I don't think that it is possible. Instead the solution I've found here is to build the Docker images inside another container, and hence as root.

### DIFF
--- a/.github/workflows/core.yml
+++ b/.github/workflows/core.yml
@@ -86,6 +86,8 @@ jobs:
       sdist_conclusion: ${{ steps.report_sdist.outputs.conclusion }}
       docs_conclusion: ${{ steps.report_docs.outputs.conclusion }}
     env:
+      OMPI_ALLOW_RUN_AS_ROOT: 1
+      OMPI_ALLOW_RUN_AS_ROOT_CONFIRM: 1
       OMP_NUM_THREADS: 1
       OPENBLAS_NUM_THREADS: 1
       FIREDRAKE_CI: 1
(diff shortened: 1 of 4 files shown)


connorjward:
@JHopeCollins don't bother reviewing. Needs more thought.

.github/workflows/pr.yml
```diff
@@ -13,3 +13,13 @@ jobs:
...
+
+  # UNDO ME
+  docker:
+    name: Build developer Docker containers
+    uses: ./.github/workflows/docker.yml
+    with:
+      tag: connorjwardtest-dev-${{ github.base_ref }}
+      branch: ${{ github.base_ref }}
+      build_dev: true
+    secrets: inherit
```
connorjward:
```suggestion
```

.github/workflows/core.yml
```diff
@@ -108,18 +108,6 @@ jobs:
         # Make sure the current directory is empty
         run: find . -delete
 
-      # Check that the Dockerfile is using the latest Ubuntu version.
-      # The version is hardcoded into the Dockerfile so that the OS
-      # for each release is fixed.
-      - name: Check Dockerfile Ubuntu version
-        run: |
-          latest_version=$(grep "VERSION_ID=" /etc/os-release | cut -d '"' -f 2)
-          docker_version=$(grep FROM docker/Dockerfile.vanilla | cut -d ':' -f 2)
```
connorjward, line 117: I moved this because I'm fairly sure that this file shouldn't exist until we run the checkout action.

PR #4547, opened 2025-09-05
remove `.at()`

leo-collins:
Removes `Function.at()` method. This is to go in after the next release. Users should now use `PointEvaluator`.


tests/firedrake/regression/test_interpolate_cross_mesh.py
```diff
@@ -481,16 +467,17 @@ def get_expected_values(
 def interpolate_function(
     m_src, m_dest, V_src, V_dest, coords, expected, expr_src, expr_dest, atol
 ):
+    dest_eval = PointEvaluator(m_dest, coords)
     f_src = Function(V_src).interpolate(expr_src)
     f_dest = assemble(interpolate(f_src, V_dest))
     assert extract_unique_domain(f_dest) is m_dest
-    got = f_dest.at(coords)
+    got = dest_eval.evaluate(f_dest)
     assert np.allclose(got, expected, atol=atol)
```
pbrubeck, line 475: This test seems to unecessarily rely on point evaluation. In general we can expect that the interpolated functions agree with the linear expression everywhere on the domain, not just at the given `coords`. We should consider changing this to `assert errornorm(f_dest, expected) <= atol` where `expected` is the `ufl.Expr`.

PR #4552, opened 2025-09-09
Matfree adjoint interpolation

pbrubeck:
# Description
Implements automatic code generation for the matrix-free assembly of dual interpolation (which previously was computed as a matvec with an explicit matrix).

This PR applies to the adjoint where the meshes are the same one. In https://github.com/firedrakeproject/firedrake/pull/4576 this code generation is extended to interpolation from a standard mesh into a `Submesh` and `VertexOnlyMesh`. 

`CrossMeshInterpolator` and `VomOntoVomXXX` are deferred as future work https://github.com/firedrakeproject/firedrake/issues/4592.


firedrake/assemble.py
```diff
@@ -588,28 +589,43 @@ def base_form_assembly_visitor(self, expr, tensor, *args):
...
             # Workaround: Renumber argument when needed since Interpolator assumes it takes a zero-numbered argument.
             if not is_adjoint and rank == 2:
-                _, v1 = expr.arguments()
-                operand = ufl.replace(operand, {v1: v1.reconstruct(number=0)})
+                v0, v1 = expr.arguments()
+                expr = ufl.replace(expr, {v0: v0.reconstruct(number=v1.number()),
+                                          v1: v1.reconstruct(number=v0.number())})
+                v, operand = expr.argument_slots()
+
+            # Assemble the interpolator matrix if the meshes are different
```
connorjward: Please explain why

firedrake/assemble.py
```diff
@@ -588,28 +589,43 @@ def base_form_assembly_visitor(self, expr, tensor, *args):
...
+            if is_adjoint and rank < 2 and source_mesh is not target_mesh:
+                expr = reconstruct_interp(operand, v=V)
+            matfree = (rank == len(expr.arguments())) and (rank < 2)
+
             # Get the interpolator
-            interp_data = expr.interp_data
+            interp_data = expr.interp_data.copy()
             default_missing_val = interp_data.pop('default_missing_val', None)
-            interpolator = firedrake.Interpolator(operand, V, **interp_data)
+            if matfree and ((is_adjoint and rank == 1) or rank == 0):
```
connorjward, line 626: Ditto. Not clear why this is being done.

connorjward: And won't this invisibly discard something that the user might specify?

firedrake/assemble.py
```diff
@@ -618,14 +634,15 @@ def base_form_assembly_visitor(self, expr, tensor, *args):
                 if is_adjoint:
                     # Out-of-place Hermitian transpose
                     petsc_mat.hermitianTranspose(out=res)
-                else:
+                elif tensor:
                     # Copy the interpolation matrix into the output tensor
                     petsc_mat.copy(result=res)
```
connorjward, line 661: `tensor.petscmat` is clearer than `res`

connorjward: The logic in this block is generally quite hard to follow.

firedrake/interpolation.py
```diff
@@ -374,6 +388,8 @@ def __init__(
                 "Can only interpolate into spaces with point evaluation nodes."
             )
 
+        if isinstance(expr, ufl.Interpolate):
+            expr, = expr.ufl_operands
```
connorjward: Does this discard anything important? `Interpolate` objects have more information than just an operand no?

firedrake/interpolation.py
```diff
@@ -713,35 +735,33 @@ def _interpolate(self, *function, output=None, transpose=None, adjoint=False, **
...
-                if self.nargs:
+                if len(self.arguments) == 2:
                     # Interpolation operator
                     self.frozen_assembled_interpolator = assembled_interpolator
                 else:
                     # Interpolation action
                     self.frozen_assembled_interpolator = assembled_interpolator.copy()
 
-        if self.nargs:
+        if len(self.arguments) == 2 and len(function):
```
connorjward: `len(function)` is an odd thing to check. What does it mean?

firedrake/interpolation.py
```diff
@@ -840,10 +865,10 @@ def make_interpolator(expr, V, subset, access, bcs=None, matfree=True):
             tensor = op2.Mat(sparsity)
         f = tensor
     else:
-        raise ValueError("Cannot interpolate an expression with %d arguments" % len(arguments))
+        raise ValueError("Cannot interpolate an expression with %d arguments" % rank)
```
connorjward: f-string

firedrake/assemble.py
pbrubeck: This logic is to determine when the `Interpolate` expression may be assembled without a matrix. By default it should be done matrix free for a zero form or linear form (rank<2). If the user wants to involve a matrix they should be passing an assembled bilinear form (this was already the case for the forward interpolation, but not for the adjoint, as we hadn't implemented it). For cross-mesh interpolation meshes we haven't implemented the adjoint action either, so we have to come up with an equivalent expression involving the matrix. I'll add a comment explaining this.

firedrake/interpolation.py
pbrubeck: This class does not implement the matfree action (Someone Else's problem). So the correct thing is to raise an exception if the interpolate object comes with a Cofunction as the dual argument.

firedrake/interpolation.py
pbrubeck: This method takes in a variable number of positional arguments as the `Functions` or `Cofunctions` to be operated on (either one or zero of them). Should the check be `len(function) > 0`?

firedrake/assemble.py
pbrubeck: We can have the four cases separetely: `tensor` (None or provided) x `is_adjoint` (True or False)

pbrubeck: Alternatively, we can set `tensor = AssembledMatrix(expr, PETSc.Mat())` and then set `tensor.petscmat`

firedrake/assemble.py
```suggestion
            # Matrix-free adjoint interpolation is only implemented by SameMeshInterpolator 
            # so we need assemble the interpolator matrix if the meshes are different
```

firedrake/assemble.py
pbrubeck, line 626:
```suggestion
            if matfree and ((is_adjoint and rank == 1) or rank == 0):
                # Adjoint interpolation of a Cofunction or the action of 
                # Cofunction on an interpolated Function require INC access
```

firedrake/interpolation.py
pbrubeck:
```suggestion
            dual_arg, expr = expr.argument_slots()
            if not isinstance(dual_arg, Coargument):
                raise NotImplementedError(f"{type(self).__name__} does not support matrix-free adjoint interpolation.")
```

firedrake/interpolation.py
pbrubeck:
```suggestion
        if len(self.arguments) == 2 and len(function) > 0:
```

firedrake/interpolation.py
pbrubeck:
```suggestion
        raise ValueError(f"Cannot interpolate an expression with {rank} arguments")
```

firedrake/assemble.py
connorjward: It's confusing because:
```py
elif tensor:  # suggests that the contents of the elif block should somehow reference 'tensor'
    petsc_mat.copy(result=res)  # but the line doesn't obviously reference 'tensor', only indirectly if you trace the logic back
```
It turns something completely obvious into something you have to think for a second about, which should always be avoided if possible. I don't mind how you fix it.

firedrake/assemble.py
```diff
@@ -618,14 +638,15 @@ def base_form_assembly_visitor(self, expr, tensor, *args):
                 if is_adjoint:
                     # Out-of-place Hermitian transpose
                     petsc_mat.hermitianTranspose(out=res)
-                else:
+                elif tensor:
```
```suggestion
                elif res:
```

pbrubeck: @connorjward better now?

connorjward: Doesn't that always evaluate to `True`?

pbrubeck: It evaluates to the pointer, an uninitialized Mat has a null pointer

connorjward: OK. It's still slightly obscure but probably better.

pbrubeck: https://gitlab.com/petsc/petsc/-/blob/main/src/binding/petsc4py/src/petsc4py/PETSc/Object.pyx?ref_type=heads#L30

firedrake/assemble.py
```diff
@@ -588,28 +591,47 @@ def base_form_assembly_visitor(self, expr, tensor, *args):
 
             # Workaround: Renumber argument when needed since Interpolator assumes it takes a zero-numbered argument.
             if not is_adjoint and rank == 2:
-                _, v1 = expr.arguments()
-                operand = ufl.replace(operand, {v1: v1.reconstruct(number=0)})
+                v0, v1 = expr.arguments()
+                expr = ufl.replace(expr, {v0: v0.reconstruct(number=v1.number()),
+                                          v1: v1.reconstruct(number=v0.number())})
```
leo-collins, line 612: Will https://github.com/firedrakeproject/firedrake/pull/4572 fix this renumbering stuff?

pbrubeck: This has to do with a limitation with `Interpolator`, since it does not assemble the adjoint matrix directly. However this makes the  logic quite confusing, as the only thing we can pass to `Interpolator` is the adjoint of the thing we want to compute.

pbrubeck: I am currently working on getting rid of this workaround and enabling direct assembly of the adjoint in SameMeshInterpolation.

pbrubeck: There are 2 situations where we need renumbering.
For backwards compatibiliy, we want 
```
interpolate(TestFunction(V1), V2) -> Interpolate(TrialFunction(V1), TestFunction(V2.dual()))
```
The second situation is only an internal reference to the 2-form Interpolate as an intermidiate step when we assemble a 0-form Interpolate or a (dual) 1-form Interpolate, or the adjoint of the 2-form Interpolate. 

This PR deals with the second situation for same mesh interpolation of 0-forms and 1-forms.  https://github.com/firedrakeproject/firedrake/pull/4576 extends this to the adjoint of the 2-form and interpolation into VertexOnlyMesh (0-,1-,2-forms). I might want to include `Submesh` there as well, but hopefully you will be keen on extending this to `CrossMeshInterpolator`.

firedrake/interpolation.py
```diff
@@ -93,10 +93,20 @@ def __init__(self, expr, v,
...
         if isinstance(v, functionspaceimpl.WithGeometry):
-            expr_args = extract_arguments(ufl.as_ufl(expr))
+            expr_args = extract_arguments(expr)
             is_adjoint = len(expr_args) and expr_args[0].number() == 0
             v = Argument(v.dual(), 1 if is_adjoint else 0)
+
+        V = v.arguments()[0].function_space()
+        if len(expr.ufl_shape) != len(V.value_shape):
+            raise RuntimeError('Rank mismatch: Expression rank %d, FunctionSpace rank %d'
+                               % (len(expr.ufl_shape), len(V.value_shape)))
```
```suggestion
            raise RuntimeError(f'Rank mismatch: Expression rank {len(expr.ufl_shape)}, FunctionSpace rank {len(V.value_shape)}')
```

firedrake/interpolation.py
```diff
@@ -93,10 +93,20 @@ def __init__(self, expr, v,
...
+
+        if expr.ufl_shape != V.value_shape:
+            raise RuntimeError('Shape mismatch: Expression shape %r, FunctionSpace shape %r'
+                               % (expr.ufl_shape, V.value_shape))
```
```suggestion
            raise RuntimeError('Shape mismatch: Expression shape {expr.ufl_shape}, FunctionSpace shape {V.value_shape}')
```

firedrake/interpolation.py
```diff
@@ -977,6 +1006,32 @@ def _interpolator(V, tensor, expr, subset, arguments, access, bcs=None):
     parameters = {}
     parameters['scalar_type'] = utils.ScalarType
 
+    callables = ()
+    if access == op2.INC:
+        callables += (tensor.zero,)
+
+    needs_weight = isinstance(dual_arg, ufl.Cofunction) and not to_element.is_dg()
```
connorjward, line 1013: Why is a weight needed? Probably needs a comment.

firedrake/interpolation.py
```diff
@@ -1119,7 +1176,8 @@ def _interpolator(V, tensor, expr, subset, arguments, access, bcs=None):
 
 def _compile_expression_key(comm, expr, to_element, ufl_element, domain, parameters) -> tuple[Hashable, ...]:
     """Generate a cache key suitable for :func:`tsfc.compile_expression_dual_evaluation`."""
-    return (hash_expr(expr), hash(ufl_element), utils.tuplify(parameters))
+    dual_arg, operand = expr.argument_slots()
+    return (hash_expr(operand), type(dual_arg), hash(ufl_element), utils.tuplify(parameters))
```
connorjward, line 1180: This seems like a bug in `hash_expr`. It looks like you are making the hash here slightly less strict.

connorjward (changes requested):
Looks mostly fine I think. Need someone else to review the maths.

firedrake/interpolation.py
pbrubeck: `ufl.Interpolate` is not your typical `ufl.Expr`, the problem is that `hash_expr` is not inspecting the dual argument (which is never a `ufl.Expr`, so `hash_expr` cannot be applied). The result of `tsfc.compile_expression_dual_evaluation` should only depend on the type of the dual argument.

pbrubeck:
> Looks mostly fine I think. Need someone else to review the maths.

The adjoint tests should speak for themselves.

I'm pretty sure @dham would agree with this statement.

tests/firedrake/regression/test_adjoint_operators.py
```diff
@@ -112,8 +112,8 @@ def test_interpolate_scalar_valued(rg):
 def test_interpolate_vector_valued():
     mesh = UnitSquareMesh(10, 10)
     V1 = VectorFunctionSpace(mesh, "CG", 1)
-    V2 = VectorFunctionSpace(mesh, "DG", 0)
-    V3 = VectorFunctionSpace(mesh, "CG", 2)
```
pbrubeck, line 116: @dham and I discussed this test, which interpolates from V2=DG0 into V3=CG2. The interpolation is not well defined and the test passes with a matrix-explicit adjoint, but fails with the matfree adjoint. The conclusion was that whoever attempts such ill-defined interpolation derseve whatever comes to them.

The whole point of this test is to test the adjoint of vector-valued interpolation, so I just used the vector variant of the spaces used in the scalar test above.

firedrake/interpolation.py
```diff
@@ -977,6 +1004,32 @@ def _interpolator(V, tensor, expr, subset, arguments, access, bcs=None):
     parameters = {}
     parameters['scalar_type'] = utils.ScalarType
 
+    callables = ()
+    if access == op2.INC:
+        callables += (tensor.zero,)
+
+    needs_weight = isinstance(dual_arg, ufl.Cofunction) and not to_element.is_dg()
```
pbrubeck, line 1013:
```suggestion

    # For the matfree adjoint 1-form and the 0-form, the cellwise kernel will add multiple
    # contributions from the facet DOFs of the dual argument.
    # The incoming Cofunction needs to be weighted by the reciprocal of the DOF multiplicity.
    needs_weight = isinstance(dual_arg, ufl.Cofunction) and not to_element.is_dg()
```

firedrake/interpolation.py
pbrubeck: Hashing on the type is justifable. `dual_arg` here can be either a `Coargument` (the only thing we were allowing before) or a `Cofunction`. `tsfc.compile_expression_dual_evaluation` does not see `dual_arg.dat` in case it is a `Cofunction`. So we only care about the symbolic information in `dual_arg`, which is moslty the `FunctionSpace`, but the kernel is mesh-independent, so we only care about the `ufl_element`, which we already hashed.

One might argue that we are not hasing `Coargument.number()`, but this is entitrely determined by the argument number in the `operand`.

firedrake/assemble.py
```diff
@@ -538,6 +538,9 @@ def base_form_assembly_visitor(self, expr, tensor, *args):
             result = expr.assemble(assembly_opts=opts)
             return tensor.assign(result) if tensor else result
         elif isinstance(expr, ufl.Interpolate):
+            if not isinstance(expr, firedrake.Interpolate):
```
dham: Why can this happen?

firedrake/assemble.py
dham, line 612: Why is this not just `adjoint(expr)`?

firedrake/assemble.py
pbrubeck: It shouldn't under normal circumstances. Someone may just call `assemble(ufl.Interpolate(...))`. If we don't want to allow this then line 540 should check for `firedrake.Interpolate`.

firedrake/interpolation.py
pbrubeck: Also be aware that before we were not passing the ufl.Interpolate, but the operand. So strictly speaking, I haven't made the hash less strict by appending an extra item to the tuple.

firedrake/assemble.py
```diff
@@ -556,6 +556,9 @@ def base_form_assembly_visitor(self, expr, tensor, *args):
...
+                expr = firedrake.Interpolate(*reversed(expr.dual_args()))
```
```suggestion
```

firedrake/assemble.py
pbrubeck: And it's gone

firedrake/interpolation.py
```diff
@@ -165,15 +173,15 @@ def interpolate(expr, V, subset=None, access=op2.WRITE, allow_missing_dofs=False
     """
     if isinstance(V, (Cofunction, Coargument)):
         dual_arg = V
-    elif isinstance(V, ufl.Form):
+    elif isinstance(V, (ufl.Form, ufl.BaseForm)):
```
dham:
```suggestion
    elif isinstance(V, ufl.BaseForm):
```

connorjward (dismissed):

tests/firedrake/regression/test_interpolate.py
```diff
@@ -327,43 +327,55 @@ def test_trace():
...
+
+    if rank == 0:
+        operand = Function(Pk).interpolate(expr)
+    else:
+        operand = TestFunction(Pk)
+
+    if mat_type == "matfree":
+        interp = interpolate(operand, v)
+    else:
+        adj_interp = assemble(interpolate(operand, TrialFunction(Pkp1.dual())))
```
pbrubeck:
```suggestion
    if rank == 0:
        operand = Function(Pk).interpolate(expr)
        dual_arg = TestFunction(Pkp1.dual())
    else:
        operand = TestFunction(Pk)
        dual_arg = TrialFunction(Pkp1.dual())

    if mat_type == "matfree":
        interp = interpolate(operand, v)
    else:
        adj_interp = assemble(interpolate(operand, dual_arg))
```

PR #4554, opened 2025-09-10
Cache TLM and Hessian solvers for NLVS blocks

angus-g:
# Description

Computing the adjoint on a `NonlinearVariationalSolve` block has a few optimisations: in effect the LVP and LVS are only initialised once when the block is created, then only the RHS is updated during adjoint evaluation (with some extra fiddling for the non-constant Jacobian case).

The TLM and Hessian *should* look similar to this, but they currently go through the slower code path on `GenericSolveBlock` that does assembly and solver creation on every evaluation, making them significantly slower. For very ballpark figures, a simple Stokes box model takes ~20s for a forward or adjoint evaluation, but ~180 for the first Hessian evaluation (involving a couple of compilation phases, I guess) and ~100 for subsequent Hessians. I think the theory dictates that this should be closer to only 2x the adjoint.

This PR is some pretty horrific code-mangling to attempt to bring parity across the different evaluation types. Currently, it sets up the LVP and LVS for TLM and Hessian on a NLVS block initialisation. I think it's fair to assume that in *most* cases the adjoint LVS is useful, but the TLM and Hessian are less common. Perhaps they can be gated by a flag if you know you'll need them, or initialised lazily (maybe a little more fiddly).

I also have completely hacked my way around the form replacement and constant Jacobian business, and they're almost certainly completely wrong. This probably deserves a more careful eye once the overall approach is a bit more settled. Indeed, maybe the form replacement mechanism could be re-engineered somewhat.

With the changes here, the same box model takes 100s for the first Hessian, then ~65 thereafter. There's still a bunch of time being spent on assembly and form manipulation for the second-order adjoint, and I assume that only needs to happen once. My model also has a couple of `ProjectBlock` that go through the slow path, but they are a tiny percentage of the overall runtime. It does mean it's a bit complicated to follow the logic through, depending on whether the `GenericSolveBlock` or `NonlinearVariationalSolveBlock` implementations of adj/tlm/hess are being used.

### DIFF
--- a/firedrake/adjoint_utils/blocks/solving.py
+++ b/firedrake/adjoint_utils/blocks/solving.py
@@ -30,6 +30,7 @@ class Solver(Enum):
     FORWARD = 0
     ADJOINT = 1
     TLM = 2
+    HESSIAN = 3
 
 
 class GenericSolveBlock(Block):
@@ -237,6 +238,9 @@ class GenericSolveBlock(Block):
 
         return adj_sol, adj_sol_bdy
 
+    def _hessian_solve(self, *args):
+        return self._assemble_and_solve_adj_eq(*args)
+
     def _compute_adj_bdy(self, adj_sol, adj_sol_bdy, dFdu_adj_form, dJdu):
         adj_sol_bdy = firedrake.assemble(dJdu - firedrake.action(dFdu_adj_form, adj_sol))
         return adj_sol_bdy.riesz_representation("l2")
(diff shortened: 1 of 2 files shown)


firedrake/adjoint_utils/blocks/solving.py
```diff
@@ -700,8 +723,10 @@ def _ad_assign_map(self, form, solver):
                            firedrake.Cofunction)):
                 coeff_count = coeff.count()
                 if coeff_count in form_ad_count_map:
-                    assign_map[form_ad_count_map[coeff_count]] = \
-                        block_variable.saved_output
+                    if solver == Solver.HESSIAN:
+                        assign_map[form_ad_count_map[coeff_count]] = block_variable.tlm_value
+                    else:
+                        assign_map[form_ad_count_map[coeff_count]] = block_variable.saved_output
```
JHopeCollins, line 730: For the hessian solve don't you need to update from `tlm_value` _as well as_ from `saved_output`, rather than _instead of_? Otherwise the adjoint Jacobian in the `_ad_solvers["hessian_lvs"]` won't be assembled around the correct state.

firedrake/adjoint_utils/blocks/solving.py
```diff
@@ -700,8 +723,10 @@ def _ad_assign_map(self, form, solver):
...
 
         if (
             solver == Solver.ADJOINT
```
JHopeCollins, line 733: Will this make sure that the state is updated correctly for the Hessian solve?
```suggestion
            (solver in (Solver.ADJOINT, Solver.HESSIAN))
```

PR #4556, opened 2025-09-11
Parametrise CrossMeshInterpolator tests

pbrubeck:
# Description

### DIFF
--- a/tests/firedrake/regression/test_interpolate_cross_mesh.py
+++ b/tests/firedrake/regression/test_interpolate_cross_mesh.py
@@ -3,8 +3,7 @@ from firedrake.petsc import DEFAULT_PARTITIONER
 from firedrake.ufl_expr import extract_unique_domain
 import numpy as np
 import pytest
-from functools import reduce
-from operator import mul, add
+from ufl import product
 import subprocess
 
 
@@ -65,9 +64,9 @@ def make_high_order(m_low_order, degree):
 def parameters(request):
     if request.param == "unitsquare":
         m_src, m_dest, coords = unitsquaresetup()
-        expr_src = reduce(mul, SpatialCoordinate(m_src))
-        expr_dest = reduce(mul, SpatialCoordinate(m_dest))
-        expected = reduce(mul, coords.T)
+        expr_src = product(SpatialCoordinate(m_src))
+        expr_dest = product(SpatialCoordinate(m_dest))
+        expected = np.prod(coords, axis=-1)
         V_src = FunctionSpace(m_src, "CG", 3)
         V_dest = FunctionSpace(m_dest, "CG", 4)
         V_dest_2 = FunctionSpace(m_dest, "DG", 2)
@@ -90,9 +89,9 @@ def parameters(request):
         # only add target mesh vertices since they are common to both meshes
         vertices_dest = allgather(m_dest.comm, m_dest.coordinates.dat.data_ro)
         coords = np.concatenate((coords, vertices_dest))
-        expr_src = reduce(mul, SpatialCoordinate(m_src))
-        expr_dest = reduce(mul, SpatialCoordinate(m_dest))
-        expected = reduce(mul, coords.T)
+        expr_src = product(SpatialCoordinate(m_src))
+        expr_dest = product(SpatialCoordinate(m_dest))
+        expected = np.prod(coords, axis=-1)
         V_src = FunctionSpace(m_src, "CG", 3)
         V_dest = FunctionSpace(m_dest, "CG", 4)
         V_dest_2 = FunctionSpace(m_dest, "DG", 2)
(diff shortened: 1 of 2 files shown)


tests/firedrake/regression/test_interpolate_cross_mesh.py
```diff
@@ -516,27 +478,35 @@ def interpolate_function(
...
     m_src, m_dest, V_src, V_dest, coords, expected, expr_src, expr_dest, atol
 ):
-    f_src = Function(V_src).interpolate(expr_src)
-
-    f_dest = assemble(interpolate(expr_src, V_dest))
+    f_dest = Function(V_dest).interpolate(expr_src)
     assert extract_unique_domain(f_dest) is m_dest
+
+    # test Function.interpolate(...)
     got = f_dest.at(coords)
```
pbrubeck: Should we be calling `Function.at()` here?

tests/firedrake/regression/test_interpolate_cross_mesh.py
```diff
@@ -516,27 +480,35 @@ def interpolate_function(
...
     assert np.allclose(got, expected, atol=atol)
     assert np.allclose(f_dest.dat.data_ro, f_dest_2.dat.data_ro, atol=atol)
 
-    # adjoint
+
+def interpolate_cofunction(
+    m_src, m_dest, V_src, V_dest, coords, expected, expr_src, expr_dest, atol
+):
+    f_dest = Function(V_dest).interpolate(expr_src)
```
pbrubeck: This step is adding unrelated extra computation
```suggestion
    f_dest = Function(V_dest).interpolate(expr_dest)
```

tests/firedrake/regression/test_interpolate_cross_mesh.py
```diff
@@ -209,8 +205,8 @@ def parameters(request):
         V_dest = TensorFunctionSpace(m_dest, "CG", 4)
         V_dest_2 = TensorFunctionSpace(m_dest, "DQ", 2)
     elif request.param == "spheresphere":
-        m_src = UnitCubedSphereMesh(5, name="src_sphere")
-        m_dest = UnitIcosahedralSphereMesh(5, name="dest_sphere")
+        m_src = UnitCubedSphereMesh(6, name="src_sphere")
+        m_dest = UnitIcosahedralSphereMesh(6, name="dest_sphere")
```
pbrubeck:
```suggestion
        m_src = UnitCubedSphereMesh(5, name="src_sphere")
        m_dest = UnitIcosahedralSphereMesh(5, name="dest_sphere")
```

tests/firedrake/regression/test_interpolate_cross_mesh.py
```diff
@@ -197,7 +193,7 @@ def parameters(request):
         expected = 2 * coords
         V_src = VectorFunctionSpace(m_src, "CG", 2)
         V_dest = FunctionSpace(m_dest, "SminusDiv", 2)  # Not point evaluation nodes
-        V_dest_2 = VectorFunctionSpace(m_dest, "DQ", 2)
+        V_dest_2 = FunctionSpace(m_dest, "SminusCurl", 2)  # Not point evaluation nodes
```
connorjward, line 196: Why this change?

connorjward (approved):
Seems fine. Anything to clear this test file up is a good thing.

I also enjoy your use of mpi-pytest.

tests/firedrake/regression/test_interpolate_cross_mesh.py
pbrubeck: The tests are now parametrised, and the unparametrized test was marked to xfail (that was the whole point of this test). Now we split this test into subtests that use `V_dest`, and subtests that use `V_dest2` (which do not fail for `'DQ'`). I didn't know how to mark the subtests, so I this makes all subtests fail.

pbrubeck:
> Seems fine. Anything to clear this test file up is a good thing.
> 
> I also enjoy your use of mpi-pytest.

That was @leo-collins, unfortunately I had to squash all commits to resolve merge conflicts.

tests/firedrake/regression/test_interpolate_cross_mesh.py
connorjward: I think it's possible to still `xfail` it as expected. You could add a check in `def test_interpolate_cross_mesh(run_test, space, parameters):` (https://docs.pytest.org/en/6.2.x/skipping.html#xfail-mark-test-functions-as-expected-to-fail).

pbrubeck: Ok, I'll just leave this failing on both subtests. There's no point on running an extra test that we weren't running before. These tests are quite slow.

connorjward: But doesn't that sort of break the design of the test? I'm still confused by why we have `V_dest` and `V_dest2`.

pbrubeck: Originally we only had one super parameter that internally did 3 types of operations (interpolation of expression, function, and cofunction)on 2 distinct inputs (`V_dest` and `V_dest2`). Here we are breaking the super parameter into 6 parameters. 

The failing test just failed on `V_dest` and never reached `V_dest2`

connorjward: OK fine. I think I'm convinced that this isn't your problem to fix. You've definitely improved things!

connorjward (approved):

PR #4558, opened 2025-09-12
Add support for deflation via firedrake.DeflatedSNES

pefarrell:
# Description

This PR adds support for deflation in Firedrake via a custom SNES solver (`firedrake.DeflatedSNES`) and an object storing the solutions to be deflated (`firedrake.Deflation`). Deflation is a numerical technique for computing multiple solutions of a nonlinear problem. A general introduction to the idea can be found in 

[https://pefarrell.org/files/talks/deflation.pdf](https://pefarrell.org/files/talks/deflation.pdf)

This PR adapts and updates some code I've had in defcon ([https://bitbucket.org/pefarrell/defcon](https://bitbucket.org/pefarrell/defcon)) for many years.

The user interface can be seen in the tests in this PR, but the basic idea is sketched below.

```
    deflation = Deflation()  # by default, deflate no solutions
    appctx = {"deflation": deflation}

    # Find the first solution
    u.interpolate(guess)
    solver = NonlinearVariationalSolver(problem, solver_parameters=sp, appctx=appctx)
    solver.solve()

    # Now deflate the first solution found, restore initial guess, and
    # find second solution
    first = Function(u)
    deflation.append(first)
    u.interpolate(guess)  # start from the same initial guess as before
    solver.solve()  # if this converges, it converges to something different
```

The VI support requires a small PR in petsc4py to wrap `SNESVIGetVariableBounds`, which I will submit presently.

### DIFF
--- a/demos/deflation/deflation.py.rst
+++ b/demos/deflation/deflation.py.rst
@@ -24,7 +24,7 @@ We demonstrate the use of deflation in Firedrake on the `Liouville–Bratu–Gel
 
     \nabla^2 u + \lambda \exp{u} = 0 \text{ in } \Omega, \quad u = 0 \text{ on } \partial \Omega.
 
-If :math:`\Omega = (0, 1)`, then for :math:`\lambda \in (0, \lambda^\star)` the equation has two solutions, for :math:`\lambda \in \{0, \lambda^\star\}` it has one solution, and for :math:`\lambda > \lambda^\star` it has no solutions. Here :math:`\lambda^\star \approx 3.51` is a constant with a known analytical expression.
+If :math:`\Omega = (0, 1)`, then for :math:`\lambda \in (0, \lambda^\star)` the equation has two solutions, for :math:`\lambda \in \{0, \lambda^\star\}` it has one solution, and for :math:`\lambda > \lambda^\star` it has no solutions. Here :math:`\lambda^\star \approx 3.51` is a constant with a known analytical expression. We will fix :math:`\lambda = 2` and prescribe :math:`6x(1-x)` as our initial guess.
 
 We implement the usual weak formulation of the equation in Firedrake as standard: ::
 
@@ -44,7 +44,7 @@ We implement the usual weak formulation of the equation in Firedrake as standard
     bcs = DirichletBC(V, 0, "on_boundary")
     problem = NonlinearVariationalProblem(F, u, bcs)
 
-Applying deflation requires two ingredients: the ``DeflatedSNES`` nonlinear solver, and a ``Deflation`` object. The ``Deflation`` object records the solutions to be deflated, and specifies the sense of distance to use in deflation. In this example we use the metric induced by the :math:`L^2(\Omega)` inner product: ::
+Applying deflation requires two ingredients: the :class:`~.DeflatedSNES` nonlinear solver, and a :class:`~.Deflation` object. The :class:`~.Deflation` object records the solutions to be deflated, and specifies the sense of distance to use in deflation. In this example we use the metric induced by the :math:`L^2(\Omega)` inner product: ::
 
     sp = {"snes_type": "python",
           "snes_python_type": "firedrake.DeflatedSNES",


pefarrell:
The petsc4py MR has now been merged to PETSc main.

tests/firedrake/deflation/test_bratu.py
```diff
@@ -0,0 +1,57 @@
...
+    second = Function(u)
+
+    print(f"Norm of difference: {norm(first - second)}")
+    assert norm(first - second) > 1
+
+    if output:
+        return (first, second)
+
+
+if __name__ == "__main__":
```
connorjward: I am not sure that we should have this sort of code in the test suite. Similarly for the `print` statements in the test itself.

firedrake/deflation.py
```diff
@@ -0,0 +1,218 @@
...
+            self.inner.setFromOptions()
+
+        # Sanity check
+        typ = self.inner.getType()
+        if typ not in ["newtonls", "newtontr", "vinewtonrsls", "vinewtonssls"]:
+            raise ValueError("We only know how to deflate with Newton-type methods")
+
+        # snes.getVariableBounds recently added,
+        # be nice to users with old PETSc versions
+        if hasattr(snes, "getVariableBounds") and typ.startswith("vi"):
```
connorjward: I feel that this isn't necessary. We require that users install a specific version of PETSc.

firedrake/deflation.py
```diff
@@ -0,0 +1,218 @@
...
+
+        self.inner.view(viewer)
+
+    def solve(self, snes, b, x):
+        out = self.inner.solve(b, x)
+        snes.reason = self.inner.reason
+        return out
+
+
+class DeflatedKSP(object):
```
connorjward:
```suggestion
class DeflatedKSP:
```

firedrake/deflation.py
```diff
@@ -0,0 +1,218 @@
...
+
+        m = 1.0
+        for root in self.roots:
+            normsq = assemble(self.op(y, root))
+            factor = normsq**(-self.power/2.0) + self.shift
+            m *= factor
+
+        return m
+
+    def deriv(self, y):
```
connorjward, line 198: Complicated methods like this should have docstrings. We also try to use type hints.

connorjward (changes requested):

firedrake/deflation.py
```diff
@@ -0,0 +1,218 @@
...
+    It fetches the solutions to deflate and the notion of distance
+    to use from the problem appctx.
+
+    In practice, deflation only requires postprocessing the Newton
+    direction after the linear solve. We use a custom KSP for this
+    purpose.
+    """
+
+    def update(self, snes):
+        self.inner.setUp()
```
pbrubeck: Is this line necessary? I think PETSc is smart enough to call setup on its own if the matrices are reassembled.

tests/firedrake/deflation/test_bratu.py
pefarrell: OK, I've removed them.

firedrake/deflation.py
```diff
@@ -0,0 +1,220 @@
...
+
+    def initialize(self, snes):
+        Citations().register("Farrell2015")
+        ctx = get_appctx(snes.getDM())
+        problem = ctx._problem
+        dm = problem.dm
+
+        self.inner = PETSc.SNES().create(comm=dm.comm)
+        self.inner.setOptionsPrefix(snes.getOptionsPrefix() + "deflated_")
```
pbrubeck:
```suggestion
        prefix = snes.getOptionsPrefix() or ""
        self.inner.setOptionsPrefix(prefix + "deflated_")
```

firedrake/deflation.py
pefarrell: I added docstrings. Adding type hints is difficult, because it takes a Function and returns a Cofunction, but I can't do

`from firedrake import Function, Cofunction`

at the start of the file because Python complains about circular dependencies.

firedrake/deflation.py
```diff
@@ -0,0 +1,217 @@
...
+        # Get the deflation object from the appctx
+        appctx = ctx.appctx
+        deflation = appctx.get("deflation")
+
+        # Hijack the KSP of the SNES we just created.
+        oldksp = self.inner.ksp
+        defksp = DeflatedKSP(deflation, problem.u, oldksp, self.inner)
+        self.inner.ksp = PETSc.KSP().createPython(defksp, comm=dm.comm)
+        self.inner.ksp.pc.setType('none')
+
```
pbrubeck, line 75:
```suggestion
        defksp = DeflatedKSP(deflation, problem.u_restrict, oldksp, self.inner)
```

firedrake/deflation.py
```diff
@@ -0,0 +1,217 @@
...
+            raise ValueError("We only know how to deflate with Newton-type methods")
+
+        # If we're solving a VI, pass the bounds
+        if typ.startswith("vi"):
+            (lb, ub) = snes.getVariableBounds()
+            self.inner.setVariableBounds(lb, ub)
+
+            # No idea why this is necessary for VINEWTONRSLS but not for NEWTONLS
+            with problem.u.dat.vec as x:
+                self.inner.setSolution(x)
```
pbrubeck, line 58:
```suggestion
            with problem.u_restrict.dat.vec as x:
```

demos/deflation/deflation.py.rst
```diff
@@ -0,0 +1,92 @@
...
+
+We now find the first solution: ::
+
+    u.interpolate(guess)
+    solver.solve()
+
+With the first solution in hand, we record it in the `Deflation` object, reset our initial guess, and solve again: ::
+
+    first = Function(u)
+    deflation.append(first)
```
pbrubeck: There's potential for the user to forget to create a copy. Should these operations be merged into one?

firedrake/deflation.py
```diff
@@ -0,0 +1,217 @@
...
+        self.power = power
+        self.shift = shift
+        self.roots = list(roots if roots else [])
+
+        if op is None:
+            op = lambda x, y: inner(x - y, x - y)*dx
+        self.op = op
+
+        self.append = self.roots.append
+
```
pbrubeck, line 179: This could be useful for the user if they want easy access to the multiple solutions
```suggestion

    def __iter__(self):
        return iter(self.roots)
        
    def __len__(self):
        return len(self.roots)

```

firedrake/deflation.py
```diff
@@ -0,0 +1,217 @@
...
+            deta = Cofunction(y.function_space().dual())
+            return deta
+
+        p = self.power
+        factors = []
+        dfactors = []
+        dnormsqs = []
+        normsqs = []
+
+        for root in self.roots:
```
pbrubeck: We can use `__iter__` here
```suggestion
        for root in self:
```

firedrake/deflation.py
```diff
@@ -0,0 +1,217 @@
...
+        typ = viewer.getType()
+        if typ != PETSc.Viewer.Type.ASCII:
+            return
+        viewer.printfASCII("Firedrake deflated SNES\n")
+
+        ctx = get_appctx(snes.getDM())
+        appctx = ctx.appctx
+        deflation = appctx.get("deflation")
+
+        if deflation:
```
pbrubeck:
```suggestion
        if deflation is not None:
```

pbrubeck: Should this throw an error if the deflation is not there? Or should we add a new deflation object and throw a warning.

firedrake/deflation.py
```diff
@@ -0,0 +1,217 @@
...
+
+            factors.append(factor)
+            dfactors.append(dfactor)
+
+        eta = product(factors)
+
+        deta = Cofunction(y.function_space().dual())
+
+        for (_, factor, dfactor, dnormsq) in zip(self.roots, factors, dfactors, dnormsqs):
+            deta.assign(float((eta/factor)*dfactor)*dnormsq + deta)
```
pbrubeck: This is mostly style, and to avoid explicit calls to `Cofunction` and reference to a dual space. Using `float()` here might be okay even in complex mode, but I don't see it necessary.
```suggestion
        deta = assemble(sum(((eta/factor)*dfactor) * dnormsq
                            for factor, dfactor, dnormsq in zip(factors, dfactors, dnormsqs)))
```

pbrubeck: Why is `self.roots` there if it is not being used?

firedrake/deflation.py
```diff
@@ -0,0 +1,217 @@
...
+            normsq = assemble(self.op(y, root))
+            factor = normsq**(-self.power/2.0) + self.shift
+            m *= factor
+
+        return m
+
+    def deriv(self, y):
+        """Evaluate the derivative of the deflation operator, at the current guess y."""
+        from firedrake import Cofunction, assemble, derivative
+        from ufl import product
```
pbrubeck: I did not know that this was a thing, nice. But we are not calling it on ufl expressions, if my understanding is correct. Maybe `numpy.prod` is better

firedrake/deflation.py
```diff
@@ -0,0 +1,217 @@
...
+        dnormsqs = []
+        normsqs = []
+
+        for root in self.roots:
+            form = self.op(y, root)
+            normsqs.append(assemble(form))
+            dnormsqs.append(assemble(derivative(form, y)))
+
+        for normsq in normsqs:
+            factor = normsq**(-p/2.0) + self.shift
```
pbrubeck: Is the shift allowed to in between deflations?
```suggestion
            factor = normsq**(-p/2.0) + float(self.shift)
```

firedrake/deflation.py
```diff
@@ -0,0 +1,217 @@
...
+        self.append = self.roots.append
+
+    def evaluate(self, y):
+        """Evaluate the value of the deflation operator, at the current guess y."""
+        from firedrake import assemble
+
+        m = 1.0
+        for root in self.roots:
+            normsq = assemble(self.op(y, root))
+            factor = normsq**(-self.power/2.0) + self.shift
```
pbrubeck:
```suggestion
            factor = normsq**(-self.power/2.0) + float(self.shift)
```

firedrake/deflation.py
```diff
@@ -0,0 +1,217 @@
...
+
+class Deflation:
+    """
+    The shifted deflation operator presented in doi:10.1137/140984798.
+    Defaults to power 2, shift 1, and the L2 norm for distance.
+    """
+    def __init__(self, roots=None, power=2, shift=1, op=None):
+        self.power = power
+        self.shift = shift
+        self.roots = list(roots if roots else [])
```
pbrubeck:
```suggestion
        self.roots = list(roots) if roots else []
```

firedrake/deflation.py
```diff
@@ -0,0 +1,217 @@
...
+        self.snes = weakref.proxy(snes)
+
+    def solve(self, ksp, b, dy_pet):
+        # Use the inner ksp to solve the original problem
+        self.ksp.setOperators(*ksp.getOperators())
+        self.ksp.solve(b, dy_pet)
+        deflation = self.deflation
+
+        if self.snes.getType().startswith("vi"):
+            vi_inact = self.snes.getVIInactiveSet()
```
pbrubeck, line 119: Can we store this method as an attribute to avoid the weakref?

firedrake/deflation.py
```diff
@@ -0,0 +1,217 @@
...
+
+        return m
+
+    def deriv(self, y):
+        """Evaluate the derivative of the deflation operator, at the current guess y."""
+        from firedrake import Cofunction, assemble, derivative
+        from ufl import product
+
+        if len(self.roots) == 0:
+            deta = Cofunction(y.function_space().dual())
```
pbrubeck: We can get rid of this cofunction if we skip the dot calculation and just return a 0 in DeflatedSNES.getEdy().

firedrake/deflation.py
```diff
@@ -0,0 +1,217 @@
...
+        if deflation is not None:
+            Edy = self.getEdy(deflation, state, update_p, vi_inact)
+
+            minv = 1.0 / deflation.evaluate(state)
+            tau = 1/(1 - minv*Edy)
+            return tau
+        else:
+            return 1
+
+    def getEdy(self, deflation, y, dy, vi_inact):
```
pbrubeck, line 140:
```suggestion
    def getEdy(self, deflation, y, dy, vi_inact):
        if len(deflation) == 0:
            return 0
```

firedrake/deflation.py
```diff
@@ -0,0 +1,217 @@
...
+class DeflatedSNES(SNESBase):
+    """
+    A SNES that implements deflation, an algorithm for finding
+    multiple solutions.
+
+    It fetches the solutions to deflate and the notion of distance
+    to use from the problem appctx.
+
+    In practice, deflation only requires postprocessing the Newton
+    direction after the linear solve. We use a custom KSP for this
```
pbrubeck, line 23: Why is this done by the KSP? Shouldn't we overload `DeflatedSNES.step()` to deal with the scaling of the Newton direction?

demos/deflation/deflation.py.rst
```diff
@@ -0,0 +1,92 @@
...
+    bcs = DirichletBC(V, 0, "on_boundary")
+    problem = NonlinearVariationalProblem(F, u, bcs)
+
+Applying deflation requires two ingredients: the ``DeflatedSNES`` nonlinear solver, and a ``Deflation`` object. The ``Deflation`` object records the solutions to be deflated, and specifies the sense of distance to use in deflation. In this example we use the metric induced by the :math:`L^2(\Omega)` inner product: ::
+
+    sp = {"snes_type": "python",
+          "snes_python_type": "firedrake.DeflatedSNES",
+          "deflated_snes_type": "newtonls",
+          "deflated_snes_monitor": None,
+          "deflated_snes_linesearch_type": "l2",
```
pbrubeck: The `linesearch_type` "l2" is going to change soon to "secant", is this crucial for the demo?

demos/deflation/deflation.py.rst
```diff
@@ -0,0 +1,92 @@
...
+
+We implement the usual weak formulation of the equation in Firedrake as standard: ::
+
+    from firedrake import *
+    mesh = UnitIntervalMesh(10)
+    V = FunctionSpace(mesh, "CG", 3)
+    x = SpatialCoordinate(mesh)[0]
+
+    u = Function(V)
+    guess = 6*x*(1-x)
```
pbrubeck:
```suggestion
    guess = Function(V).interpolate(6*x*(1-x))
```

demos/deflation/deflation.py.rst
```diff
@@ -0,0 +1,92 @@
...
+          "deflated_pc_type": "lu"}
+
+    deflation = Deflation(op=lambda x, y: inner(x-y, x-y)*dx)
+    appctx = {"deflation": deflation}
+
+    solver = NonlinearVariationalSolver(problem, solver_parameters=sp, appctx=appctx)
+
+We now find the first solution: ::
+
+    u.interpolate(guess)
```
pbrubeck:
```suggestion
    u.assign(guess)
```

demos/deflation/deflation.py.rst
```diff
@@ -0,0 +1,92 @@
...
+    solver.solve()
+
+With the first solution in hand, we record it in the `Deflation` object, reset our initial guess, and solve again: ::
+
+    first = Function(u)
+    deflation.append(first)
+    u.interpolate(guess)
```
pbrubeck:
```suggestion
    u.assign(guess)
```

demos/deflation/deflation.py.rst
pbrubeck: Or at least, the demo should discuss that the solutions to append are not meant to be modified and thus the copy is required.

demos/deflation/deflation.py.rst
```diff
@@ -0,0 +1,92 @@
...
+    solver.solve()
+    second = Function(u)
+
+We can check that the two solutions are distinct: ::
+
+    print(f"Norm of difference: {norm(first - second)}")
```
pbrubeck, line 76: The html will not show this value. It has to be explicitly included in the comments

demos/deflation/deflation.py.rst
pbrubeck: Or even better, the deflation object should be able to automatically do this after the Newton solve

demos/deflation/deflation.py.rst
pefarrell: It is not, I was just showing how to set options on the inner SNES. I'll change it to basic.

demos/deflation/deflation.py.rst
pefarrell: I worry that might be a little bit too magical---but I think you are right. I'll make the `Deflation` object deflate the solution automatically.

firedrake/deflation.py
pefarrell: If I overload `DeflatedSNES.step`, then I have to manually control the inner SNES to only do one step, duplicate  the termination logic, etc. I think that's more invasive than this way.

firedrake/deflation.py
pefarrell: I confess I don't understand. Won't that keep a strong reference?

firedrake/deflation.py
```diff
@@ -0,0 +1,226 @@
...
+    def __init__(self, roots=None, power=2, shift=1, op=None):
+        self.power = power
+        self.shift = shift
+        self.roots = list(roots) if roots else []
+
+        if op is None:
+            op = lambda x, y: inner(x - y, x - y)*dx
+        self.op = op
+
+        self.append = self.roots.append
```
pbrubeck, line 178: Is this still necessary?

firedrake/deflation.py
```diff
@@ -0,0 +1,226 @@
...
+        for root in self.roots:
+            normsq = assemble(self.op(y, root))
+            factor = normsq**(-self.power/2.0) + float(self.shift)
+            m *= factor
+
+        return m
+
+    def deriv(self, y):
+        """Evaluate the derivative of the deflation operator, at the current guess y."""
+        from firedrake import Cofunction, assemble, derivative
```
pbrubeck:
```suggestion
        from firedrake import assemble, derivative
```

firedrake/deflation.py
pefarrell: It's convenient to have, if not necessary.

firedrake/deflation.py
connorjward: OK then that's probably fine. There are workarounds for this but we/the Python community haven't really decided how to handle this case universally. Happy to skip it for now.

firedrake/deflation.py
```diff
@@ -0,0 +1,226 @@
...
+        p = self.power
+        factors = []
+        dfactors = []
+        dnormsqs = []
+        normsqs = []
+
+        for root in self:
+            form = self.op(y, root)
+            normsqs.append(assemble(form))
+            dnormsqs.append(assemble(derivative(form, y)))
```
pbrubeck, line 212: I think the complex failure is due to the fact that we cannot assemble the derivative of an energy

pefarrell: Fair enough. I never ran defcon in complex mode. Let's defer it to future work. I've marked the tests to be skipped in complex mode.

PR #4568, opened 2025-09-18
`RieszMap` fixes/suggestions

jrmaddison:
Fixes based on reading the code.

1. Handle boundary conditions in the Riesz map inverse. This seems to be the consistent way to apply this, but should be reviewed.
2. Zero guess the Riesz map linear solver. Required for consistent behaviour, e.g. if changing the order of mappings.

Minor fixes

3. Remove an unnecessary allocation.
4. Catch the case where boundary conditions are supplied with an $l_2$ Riesz map.

### DIFF
--- a/firedrake/cofunction.py
+++ b/firedrake/cofunction.py
@@ -485,6 +485,7 @@ class RieszMap:
             else:
                 solve, rhs, soln = self._solver
                 rhs.assign(value)
+                soln.zero()
                 solve()
                 output.assign(soln)
         elif ufl.duals.is_primal(value):
--- a/firedrake/cofunction.py
+++ b/firedrake/cofunction.py
@@ -486,7 +486,6 @@ class RieszMap:
                 solve, rhs, soln = self._solver
                 rhs.assign(value)
                 solve()
-                output = Function(self._function_space)
                 output.assign(soln)
         elif ufl.duals.is_primal(value):
             if value.function_space() != self._function_space:
--- a/firedrake/cofunction.py
+++ b/firedrake/cofunction.py
@@ -426,6 +426,13 @@ class RieszMap:
                     sobolev_space, u, v
                 )
 
+        if bcs is None:
+            bcs = ()
+        else:
+            bcs = tuple(bcs)
+        if len(bcs) > 0 and inner_product == "l2":
+            raise ValueError("Cannot supply boundary conditions with an l2 Riesz map")
+
         self._function_space = function_space
         self._inner_product = inner_product
         self._bcs = bcs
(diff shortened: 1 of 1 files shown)


firedrake/cofunction.py
```diff
@@ -478,8 +485,8 @@ def __call__(self, value):
             else:
                 solve, rhs, soln = self._solver
                 rhs.assign(value)
+                soln.zero()
```
JHopeCollins, line 488: Doesn't PETSc always use a zero initial guess unless you pass `-ksp_initial_guess_nonzero`?

jrmaddison: This might be ignored due to #4142. I'd need to check for this use case, but I think it's still safer to zero here.

JHopeCollins:
@pbrubeck I think this looks fine but could you cast a quick eye over it please?

firedrake/cofunction.py
```diff
@@ -490,8 +497,13 @@ def __call__(self, value):
...
                     o.dat.data[:] = c.dat.data_ro[:]
             else:
+                if len(self._bcs) > 0:
+                    value = value.copy(deepcopy=True)
+                    for bc in self._bcs:
+                        bc.apply(value)
                 output = firedrake.assemble(
-                    firedrake.action(self._inner_product, value)
+                    firedrake.action(self._inner_product, value),
+                    bcs=self._bcs, zero_bc_nodes=True,
```
pbrubeck, line 506: By default we call `bc.zero` in `assemble`

pbrubeck: Being explicit here is fine.

firedrake/cofunction.py
pbrubeck: `LinearSolver` inherits from `NonlinearVariationalSolver`, which means that we are running a SNES, which gives a different behaivor from the plain KSP. Currently `-ksp_initial_guess_nonzero False` zeros the update to whatever is prescribed as the initial guess.

pbrubeck: Setting the initial guess explicitly outside PETSc is the only way with SNES.

firedrake/cofunction.py
```diff
@@ -426,6 +426,13 @@ def __init__(self, function_space_or_inner_product=None,
                     sobolev_space, u, v
                 )
 
+        if bcs is None:
+            bcs = ()
+        else:
+            bcs = tuple(bcs)
+        if len(bcs) > 0 and inner_product == "l2":
+            raise ValueError("Cannot supply boundary conditions with an l2 Riesz map")
```
pbrubeck, line 434: I'm curious here. Does this lead to the wrong results in optimization?

pbrubeck: Does "l2" here means that we are imposing the wrong inner-product between two primal objects, or does it mean that we are pairing a primal with a dual?

jrmaddison: > I'm curious here. Does this lead to the wrong results in optimization?

For $l_2$ the bcs are ignored. They could probably be applied, but I'm not sure it would be that useful.

> Does "l2" here means that we are imposing the wrong inner-product between two primal objects, or does it mean that we are pairing a primal with a dual?

Yes, it's the basis dependent map defined by dof assignment. I'm not sure if there's a case to remove this now (Firedrake code at least doesn't seem to use it much?) but maybe not for this PR?

JHopeCollins: >  I'm not sure if there's a case to remove this now (Firedrake code at least doesn't seem to use it much?) but maybe not for this PR?

This close to a major release I'd say lets keep this PR to fixes, then we can think about API changes after.

Issue #4483, opened 2025-08-06
BUG:


ksagiyam:
```
from firedrake import *
from pyop2 import op2


distribution = {
    "partitioner_type": "simple",
}
mesh = RectangleMesh(
    3, 1, 3., 1., quadrilateral=True, distribution_parameters=distribution,
)
V = FunctionSpace(mesh, "CG", 1)
f = Function(V)
g = Function(V)
f.dat.data[:] = 1.
g.dat.data[:] = 2.
f.interpolate(g, subset=op2.Subset(mesh.topology.cell_set, [1]))

plex = mesh.topology_dm
plex.viewFromOptions("-dm_view")
print(mesh.comm.rank)
print("mesh.cell_closure")
print(mesh.cell_closure)
print("mesh.coordinates.dat.data_with_halos")
print(mesh.coordinates.dat.data_with_halos)
print("f.dat._data (before halo exchange)")
print(f.dat._data)
print("f.dat.data_with_halos (after halo exchange)")
print(f.dat.data_with_halos)
```

```
...

Cones:
[0] Max cone size: 4
[0]: 0 <---- 11 (0)
[0]: 0 <---- 12 (0)
[0]: 0 <---- 13 (0)
[0]: 0 <---- 14 (0)
[0]: 1 <---- 13 (-1)
[0]: 1 <---- 15 (0)
[0]: 1 <---- 17 (0)
[0]: 1 <---- 16 (0)
[0]: 2 <---- 17 (-1)
[0]: 2 <---- 18 (0)
[0]: 2 <---- 19 (0)
[0]: 2 <---- 20 (0)
[0]: 11 <---- 3 (0)
[0]: 11 <---- 4 (0)
[0]: 12 <---- 4 (0)
[0]: 12 <---- 6 (0)
[0]: 13 <---- 6 (0)
[0]: 13 <---- 5 (0)
[0]: 14 <---- 5 (0)
[0]: 14 <---- 3 (0)
[0]: 15 <---- 6 (0)
[0]: 15 <---- 8 (0)
[0]: 16 <---- 7 (0)
[0]: 16 <---- 5 (0)
[0]: 17 <---- 8 (0)
[0]: 17 <---- 7 (0)
[0]: 18 <---- 8 (0)
[0]: 18 <---- 10 (0)
[0]: 19 <---- 10 (0)
[0]: 19 <---- 9 (0)
[0]: 20 <---- 9 (0)
[0]: 20 <---- 7 (0)
[...]

PR #4576, opened 2025-09-19
Explicitly assemble the interpolate adjoint matrix

pbrubeck:
# Description
Enables code generation for all possible argument combinations in `Interpolate`: 

1. Forward 2-form: `Interpolate(Argument(1), Coargument(0))`
2. Adjoint 2-form: `Interpolate(Argument(0), Coargument(1)) `
3. Forward 1-form: `Interpolate({Functoin | Expr}, Coargument(0))`
4. Adjoint 1-form: `Interpolate(Argument(0), Cofunction) `
5. 0-form: `Interpolate({Function | Expr}, Cofunction)`

Previously we indirectly obtained 2 and 4 from the transpose of 1, and 5 via a dot product involving 3. 

The big win is that code generation does not deal with these 5 cases as special cases, and we always do matrix-free computations (unless a 2-form is explicitly requested).

Depends on https://github.com/firedrakeproject/firedrake/pull/4552 which mainly addresses 4 and 5.

### DIFF
--- a/firedrake/assemble.py
+++ b/firedrake/assemble.py
@@ -556,7 +556,6 @@ class BaseFormAssembler(AbstractFormAssembler):
             result = expr.assemble(assembly_opts=opts)
             return tensor.assign(result) if tensor else result
         elif isinstance(expr, ufl.Interpolate):
-            orig_expr = expr
             # Replace assembled children
             _, operand = expr.argument_slots()
             v, *assembled_operand = args
@@ -575,6 +574,8 @@ class BaseFormAssembler(AbstractFormAssembler):
             # 4) Interpolate(Argument(V1, 0), Cofunction(...)) -> Action of the Jacobian adjoint
             # This can be generalized to the case where the first slot is an arbitray expression.
             rank = len(expr.arguments())
+            if rank > 2:
+                raise ValueError("Cannot assemble an Interpolate with more than two arguments")
             # If argument numbers have been swapped => Adjoint.
             arg_operand = ufl.algorithms.extract_arguments(operand)
             is_adjoint = (arg_operand and arg_operand[0].number() == 0)
@@ -605,60 +606,23 @@ class BaseFormAssembler(AbstractFormAssembler):
                     assemble(sub_interp, tensor=tensor.subfunctions[i])
                 return tensor
 
-            # Matrix-free adjoint interpolation is only implemented by SameMeshInterpolator
-            # so we need assemble the interpolator matrix if the meshes are different
-            target_mesh = V.mesh()
-            source_mesh = extract_unique_domain(operand) or target_mesh
-            if is_adjoint and rank < 2 and (source_mesh is not target_mesh):
-                expr = reconstruct_interp(operand, v=V)
-            matfree = (rank == len(expr.arguments())) and (rank < 2)
-
             # Get the interpolator
             interp_data = expr.interp_data.copy()
             default_missing_val = interp_data.pop('default_missing_val', None)
-            if matfree and ((is_adjoint and rank == 1) or rank == 0):
+
+            target_mesh = V.mesh()
+            source_mesh = extract_unique_domain(operand) or target_mesh
+            if (source_mesh is target_mesh) and ((is_adjoint and rank == 1) or rank == 0):
                 # Adjoint interpolation of a Cofunction or the action of a
                 # Cofunction on an interpolated Function require INC access
                 # on the output tensor
                 interp_data["access"] = op2.INC
 
-            if rank == 1 and matfree and isinstance(tensor, firedrake.Function):
+            if rank == 1 and isinstance(tensor, firedrake.Function):
                 V = tensor
             interpolator = firedrake.Interpolator(expr, V, **interp_data)
-
             # Assembly
-            if matfree:
-                # Assembling the operator
-                return interpolator._interpolate(output=tensor, default_missing_val=default_missing_val)
-            elif rank == 0:
-                # Assembling the double action.
-                Iu = interpolator._interpolate(default_missing_val=default_missing_val)
-                return assemble(ufl.Action(v, Iu), tensor=tensor)
-            elif rank == 1:
-                # Assembling the action of the Jacobian adjoint.
-                if is_adjoint:
-                    return interpolator._interpolate(v, output=tensor, adjoint=True, default_missing_val=default_missing_val)
-                # Assembling the Jacobian action.
-                else:
-                    return interpolator._interpolate(operand, output=tensor, default_missing_val=default_missing_val)
-            elif rank == 2:
-                res = tensor.petscmat if tensor else PETSc.Mat()
-                # Get the interpolation matrix
-                op2_mat = interpolator.callable()
-                petsc_mat = op2_mat.handle
-                if is_adjoint and (source_mesh is not target_mesh):
-                    # Out-of-place Hermitian transpose
-                    petsc_mat.hermitianTranspose(out=res)
-                elif res:
-                    # Copy the interpolation matrix into the output tensor
-                    petsc_mat.copy(result=res)
-                else:
-                    res = petsc_mat
-                if tensor is None:
-                    tensor = self.assembled_matrix(orig_expr, res)
-                return tensor
-            else:
-                raise ValueError("Incompatible number of arguments.")
+            return interpolator.assemble(tensor=tensor, default_missing_val=default_missing_val)
         elif tensor and isinstance(expr, (firedrake.Function, firedrake.Cofunction, firedrake.MatrixBase)):
             return tensor.assign(expr)
         elif tensor and isinstance(expr, ufl.ZeroBaseForm):
--- a/firedrake/assemble.py
+++ b/firedrake/assemble.py
@@ -609,15 +609,6 @@ class BaseFormAssembler(AbstractFormAssembler):
             # Get the interpolator
             interp_data = expr.interp_data.copy()
             default_missing_val = interp_data.pop('default_missing_val', None)
-
-            target_mesh = V.mesh()
-            source_mesh = extract_unique_domain(operand) or target_mesh
-            if (source_mesh is target_mesh) and ((is_adjoint and rank == 1) or rank == 0):
-                # Adjoint interpolation of a Cofunction or the action of a
-                # Cofunction on an interpolated Function require INC access
-                # on the output tensor
-                interp_data["access"] = op2.INC
-
             if rank == 1 and isinstance(tensor, firedrake.Function):
                 V = tensor
             interpolator = firedrake.Interpolator(expr, V, **interp_data)
--- a/firedrake/assemble.py
+++ b/firedrake/assemble.py
@@ -605,13 +605,6 @@ class BaseFormAssembler(AbstractFormAssembler):
                     assemble(sub_interp, tensor=tensor.subfunctions[i])
                 return tensor
 
-            # Workaround: Renumber argument when needed since Interpolator assumes it takes a zero-numbered argument.
-            if not is_adjoint and rank == 2:
-                v0, v1 = expr.arguments()
-                expr = ufl.replace(expr, {v0: v0.reconstruct(number=v1.number()),
-                                          v1: v1.reconstruct(number=v0.number())})
-                v, operand = expr.argument_slots()
-
             # Matrix-free adjoint interpolation is only implemented by SameMeshInterpolator
             # so we need assemble the interpolator matrix if the meshes are different
             target_mesh = V.mesh()
@@ -653,7 +646,7 @@ class BaseFormAssembler(AbstractFormAssembler):
                 # Get the interpolation matrix
                 op2_mat = interpolator.callable()
                 petsc_mat = op2_mat.handle
-                if is_adjoint:
+                if is_adjoint and (source_mesh is not target_mesh):
                     # Out-of-place Hermitian transpose
                     petsc_mat.hermitianTranspose(out=res)
                 elif res:
--- a/firedrake/assemble.py
+++ b/firedrake/assemble.py
@@ -567,12 +567,6 @@ class BaseFormAssembler(AbstractFormAssembler):
             if (v, operand) != expr.argument_slots():
                 expr = reconstruct_interp(operand, v=v)
 
-            # Different assembly procedures:
-            # 1) Interpolate(Argument(V1, 1), Argument(V2.dual(), 0)) -> Jacobian (Interpolate matrix)
-            # 2) Interpolate(Coefficient(...), Argument(V2.dual(), 0)) -> Operator (or Jacobian action)
-            # 3) Interpolate(Argument(V1, 0), Argument(V2.dual(), 1)) -> Jacobian adjoint
-            # 4) Interpolate(Argument(V1, 0), Cofunction(...)) -> Action of the Jacobian adjoint
-            # This can be generalized to the case where the first slot is an arbitray expression.
             rank = len(expr.arguments())
             if rank > 2:
                 raise ValueError("Cannot assemble an Interpolate with more than two arguments")
--- a/firedrake/assemble.py
+++ b/firedrake/assemble.py
@@ -609,7 +609,7 @@ class BaseFormAssembler(AbstractFormAssembler):
             # so we need assemble the interpolator matrix if the meshes are different
             target_mesh = V.mesh()
             source_mesh = extract_unique_domain(operand) or target_mesh
-            if is_adjoint and rank < 2 and source_mesh is not target_mesh:
+            if is_adjoint and rank < 2 and (source_mesh is not target_mesh):
                 expr = reconstruct_interp(operand, v=V)
             matfree = (rank == len(expr.arguments())) and (rank < 2)
 
--- a/firedrake/interpolation.py
+++ b/firedrake/interpolation.py
@@ -293,7 +293,12 @@ class Interpolator(abc.ABC):
         # CrossMeshInterpolator is not yet aware of self.ufl_interpolate (which carries the dual arguments).
         # Instead, we always construct the forward ufl_interpolate and externally operate on the adjoint and
         # supply the cofunctions within assemble().
-        if not isinstance(self, SameMeshInterpolator):
+        target_mesh = as_domain(V)
+        source_mesh = extract_unique_domain(operand) or target_mesh
+        vom_onto_other_vom = ((source_mesh is not target_mesh)
+                              and isinstance(source_mesh.topology, VertexOnlyMeshTopology)
+                              and isinstance(target_mesh.topology, VertexOnlyMeshTopology))
+        if not isinstance(self, SameMeshInterpolator) or vom_onto_other_vom:
             if not isinstance(dual_arg, ufl.Coargument):
                 expr = expr._ufl_expr_reconstruct_(operand, dual_arg.function_space().dual())
             expr_args = extract_arguments(operand)
--- a/firedrake/interpolation.py
+++ b/firedrake/interpolation.py
@@ -275,9 +275,11 @@ class Interpolator(abc.ABC):
         allow_missing_dofs=False,
         matfree=True
     ):
-        if isinstance(expr, ufl.Interpolate):
-            expr, = expr.ufl_operands
-        self.expr = expr
+        if not isinstance(expr, ufl.Interpolate):
+            fs = V if isinstance(V, ufl.FunctionSpace) else V.function_space()
+            expr = firedrake.Interpolate(expr, fs)
+        operand, = expr.ufl_operands
+        self.expr = operand
         self.V = V
         self.subset = subset
         self.freeze_expr = freeze_expr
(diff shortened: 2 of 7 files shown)


firedrake/interpolation.py
```diff
@@ -278,9 +278,12 @@ def __init__(
         allow_missing_dofs=False,
         matfree=True
     ):
-        if isinstance(expr, ufl.Interpolate):
-            expr, = expr.ufl_operands
-        self.expr = expr
+        if not isinstance(expr, ufl.Interpolate):
+            fs = V if isinstance(V, ufl.FunctionSpace) else V.function_space()
+            expr = Interpolate(expr, fs)
```
pbrubeck: Should this one be `firedrake.Interpolate` or `firedrake.interpolate`?

leo-collins: I think for now to be safe it should be `interpolate`

firedrake/interpolation.py
```diff
@@ -289,14 +292,25 @@ def __init__(
...
-        expr_args = extract_arguments(expr)
-        if expr_args and expr_args[0].number() == 0:
-            v, = expr_args
-            expr = replace(expr, {v: v.reconstruct(number=1)})
-        self.expr_renumbered = expr
+        # Workaround for adjoint of cross-mesh interpolation
+        # Assemble the forward operator and then take its adjoint
+        target_mesh = as_domain(V)
+        source_mesh = extract_unique_domain(operand) or target_mesh
+        if not ((target_mesh is source_mesh) or isinstance(target_mesh, VertexOnlyMeshTopology)):
```
pbrubeck: Ideally this should become `if not isinstance(self, SameMeshInterpolator)`, but I think I am not getting the right results for adjoint interpolation with `Submesh`. I'll add tests for this if these are missing.

firedrake/interpolation.py
```diff
@@ -828,40 +874,27 @@ def make_interpolator(expr, V, subset, access, bcs=None, matfree=True):
...
-                # allowed to be None when interpolating from a Real space, even
-                # in the trans-mesh case.
-                if source_mesh.extruded:
-                    # ExtrudedSet cannot be a map target so we need to build
-                    # this ourselves
-                    argfs_map = vom_cell_parent_node_map_extruded(target_mesh, argfs_map)
-                else:
-                    argfs_map = compose_map_and_cache(target_mesh.cell_parent_cell_map, argfs_map)
-        elif vom_onto_other_vom:
-            argfs_map = argfs.cell_node_map()
```
pbrubeck, line 852: This case is not covered by the newly introduced helper function `get_coefficient_map`, but it is alright, as `argfs_map` is not used for this case anyway.

firedrake/interpolation.py
```diff
@@ -1167,6 +1169,36 @@ def _interpolator(V, tensor, expr, subset, arguments, access, bcs=None):
         return copyin + callables + (parloop_compute_callable, ) + copyout
 
 
+def get_coefficient_map(source_mesh, target_mesh, fs):
```
pbrubeck: Is this a good name?

firedrake/assemble.py
```diff
@@ -589,67 +590,14 @@ def base_form_assembly_visitor(self, expr, tensor, *args):
                     assemble(sub_interp, tensor=tensor.subfunctions[i])
                 return tensor
 
-            # Workaround: Renumber argument when needed since Interpolator assumes it takes a zero-numbered argument.
```
pbrubeck, line 608: This logic was moved to interpolation.py, as it was here mainly due to limitations on the arguments that we suppport for direct assembly of a matfree/adjoint `Interpolate`. We introduced `Interpolate.assemble()` that performs the matrix-free action, or the explicit assembly of the transpose (for most instances of `SameMeshInterpolator`).

firedrake/interpolation.py
```diff
@@ -289,14 +292,25 @@ def __init__(
...
+                v0, v1 = expr.arguments()
+                expr = ufl.replace(expr, {v0: v0.reconstruct(number=v1.number()),
+                                          v1: v1.reconstruct(number=v0.number())})
+
+        dual_arg, operand = expr.argument_slots()
+        self.expr_renumbered = operand
+        self.ufl_interpolate_renumbered = expr
+        if not isinstance(dual_arg, ufl.Coargument):
+            # Matrix-free assembly of 0-form or 1-form requires INC access
+            self.access = op2.INC
```
pbrubeck: Should we change the default value to `access=None` (then setting the default to `op2.INC` if we have a dual `Interpolate` or `op2.WRITE` otherwise) and raise an error `if access != op2.INC` for the dual `Interpolate`?

firedrake/interpolation.py
```diff
@@ -289,14 +292,25 @@ def __init__(
...
+                expr = expr._ufl_expr_reconstruct_(operand, dual_arg.function_space().dual())
+            expr_args = extract_arguments(operand)
+            if expr_args and expr_args[0].number() == 0:
+                v0, v1 = expr.arguments()
+                expr = ufl.replace(expr, {v0: v0.reconstruct(number=v1.number()),
+                                          v1: v1.reconstruct(number=v0.number())})
+
+        dual_arg, operand = expr.argument_slots()
+        self.expr_renumbered = operand
+        self.ufl_interpolate_renumbered = expr
```
pbrubeck: "renumbered" might not be the best descriptor for what is going on. There are 5 cases:

1. Forward 2-form: Interpolate(v1, v0*)
2. Adjoint 2-form: Interpolate(v0, v1*) -> Adjoint(Interpolate(v1, v0*))
3. Forward 1-form: Interpolate(u, v0*)
4. Adjoint 1-form: Interpolate(v0, u*) -> Action(Adjoint(Interpolate(v1, v0*)), u*)
5. 0-form: Interpolate(u, u*) -> Action(u*, Interpolate(u, v0*))

Originally we could compute all 5 cases via the forward 1-form and the forward 2-form. Now `SameMeshInterpolator` can automatically generate kernels for all 5 cases. 

With `CrossMeshInterpolator`, the arguments get renumbered in cases 2 and 4 (because they involve the adjoint) but not for case 5.

tsfc/driver.py
```diff
@@ -211,33 +211,31 @@ def compile_expression_dual_evaluation(expression, to_element, ufl_element, *,
...
 
     # Apply UFL preprocessing
     operand = ufl_utils.preprocess_expression(operand, complex_mode=complex_mode)
 
-    if isinstance(expression, ufl.Interpolate):
-        v, _ = expression.argument_slots()
-        expression = ufl.Interpolate(operand, v)
-    else:
-        expression = operand
+    expression = ufl.Interpolate(operand, v)
```
connorjward: Is this because `__new__` now means that interpolate-of-interpolate-of-something just gives you back interpolate-of-something? Seems a bit magic.

firedrake/interpolation.py
connorjward: Please leave a TODO comment. This exposition won't be findable once the PR is merged.

firedrake/interpolation.py
connorjward: * This exposition should really live somewhere in the code.
* This seems like an insane interface. Is there a pathway to making things clearer? I think renumbering things behind peoples' backs is questionable.

firedrake/interpolation.py
```diff
@@ -320,6 +334,44 @@ def _interpolate(self, *args, **kwargs):
         """
         pass
 
+    def assemble(self, tensor=None, default_missing_val=None):
+        """Assemble the operator (or its action)."""
+        from firedrake.assemble import assemble
+        renumbered = self.ufl_interpolate_renumbered != self.ufl_interpolate
```
```suggestion
        arguments_are_renumbered = self.ufl_interpolate_renumbered != self.ufl_interpolate
```
?

connorjward: Or `arguments_are_transposed`?

firedrake/interpolation.py
connorjward: It could definitely be better. Maybe say `node_map`? A docstring would also be very helpful. "map" is extremely overloaded.

firedrake/interpolation.py
pbrubeck: If we want to make things clearer we should ultimately implement the five cases in `CrossMeshInterpolator`. We could instead reformulate this logic in terms of an internal forward `Interpolate` and a `needs_forward_interpolate` flag to control the logic.

connorjward: We should be aiming to reduce special casing so I'm not sure about that.

What can't `CrossMeshInterpolator` do?

pbrubeck: The 5 cases are not implemented as special cases in SameMeshInterpolator, the arguments are passed on to the kernel, and everything works nicely. The original code relying on a forward `Interpolate` dealt with these cases  as special cases

tsfc/driver.py
pbrubeck: No, `Intepolate(Interpolate(...))` is not really happening here. This is to reconstruct the original `Interpolate` with a mapped operand, or to construct a new `Interpolate` if the input was not an `Interpolate`.

connorjward: OK well it is not obvious what is going on.

firedrake/interpolation.py
pbrubeck: I am not enhancing `CrossMeshInterpolator` in these PRs. It will keep relying on the internal forward `Interpolate` until someone implements the matrix-free adjoint and the explicit adjoint.

connorjward: I see. I think my issue is that the comment
```py
# Workaround for adjoint of cross-mesh interpolation
# Assemble the forward operator and then take its adjoint
```
does not explain why this is necessary.

tsfc/driver.py
pbrubeck:
```suggestion
    # Reconstructed Interpolate with mapped operand
    expression = ufl.Interpolate(operand, v)
```

tsfc/driver.py
```diff
@@ -211,33 +211,31 @@ def compile_expression_dual_evaluation(expression, to_element, ufl_element, *,
...
+        v, operand = expression.argument_slots()
     else:
         operand = expression
+        v = ufl.FunctionSpace(extract_unique_domain(operand), ufl_element)
 
     # Map into reference space
     operand = apply_mapping(operand, ufl_element, domain)
 
     # Apply UFL preprocessing
     operand = ufl_utils.preprocess_expression(operand, complex_mode=complex_mode)
```
connorjward, line 224: Why can't we just do this to the provided expression? Why must we pick out the operand and put it back in?

pbrubeck: We don't want to apply the mapping on the output of the `Interpolate`, only on the input `operand`. The reconstructed  `ufl.Interpolate` is mainly there so that we can `extract_coefficients` from the reconstructed `operand` and the `dual_arg`. We then need to extract the operand once again when we evaluate the degrees of freedom in finat.

connorjward: OK. I guess this OK. I still think it's a nasty pattern.

tsfc/driver.py
connorjward: I guess this is fine. Thanks.

firedrake/interpolation.py
pbrubeck: Adjoint interpolation with Submesh is supported now. I also added (serial) tests. The [issue in parallel](https://github.com/firedrakeproject/firedrake/issues/4483) should still remain for matfree forward interpolation but I think it should not be a problem for the matfree adjoint (since `access=INC`), so maybe I should parallel tests.

pbrubeck: Parallel test added. Matfree adjoint interpolation seems fine.

firedrake/interpolation.py
pbrubeck: I updated the comments and the description of the PR to clarify this insane renumbering (which is the current code path, and is being removed only for the code-generated kernel)

connorjward: I won't review again until you want me to, but this looks much better.

pbrubeck: I think I am quite ready for another review, thanks

firedrake/interpolation.py
pbrubeck: I went for `needs_adjoint`.

firedrake/interpolation.py
```diff
@@ -289,14 +292,42 @@ def __init__(
...
+            # 1) Interpolate(Argument(V1, 1), Argument(V2.dual(), 0)) -> Forward operator (2-form)
+            # 2) Interpolate(Argument(V1, 0), Argument(V2.dual(), 1)) -> Adjoint operator (2-form)
+            # 3) Interpolate(Coefficient(V1), Argument(V2.dual(), 0)) -> Forward action (1-form)
+            # 4) Interpolate(Argument(V1, 0), Cofunction(V2.dual()) -> Adjoint action (1-form)
+            # 5) Interpolate(Coefficient(V1), Cofunction(V2.dual()) -> Double action (0-form)
+
+            # CrossMeshInterpolator._interpolate only supports forward interpolation (cases 1 and 3).
+            # For case 2, we first redundantly assemble case 1 and then construct the transpose.
+            # For cases 4 and 5, we take the forward Interpolate that corresponds to dropping the Cofunction,
+            # and we separately compute the action againt the dropped Cofunction within assemble().
```
```suggestion
            # and we separately compute the action against the dropped Cofunction within assemble().
```

tests/firedrake/submesh/test_submesh_interpolate.py
```diff
@@ -268,3 +272,75 @@ def expr(m):
...
+        # See https://github.com/firedrakeproject/firedrake/issues/4483
+
+        Iu1 = assemble(interpolate(u1, TestFunction(V2.dual()), allow_missing_dofs=True))
+        assert Iu1.function_space() == V2
+
+        expected_primal = assemble(action(I, u1))
+        assert np.allclose(Iu1.dat.data, expected_primal.dat.data)
+
+        result_forward_1 = assemble(action(ustar2, Iu1))
+        assert np.isclose(result_forward_1, expected)
```
connorjward: I think I would prefer to have
```py
is_close = np.isclose(result_forward_1, expected)
if comm.size == 1:
    assert is_close
else:
    assert not is_close
```
That way if we fix the bug in the future we will know to come back and fix this test.

firedrake/interpolation.py
```diff
@@ -289,14 +292,42 @@ def __init__(
...
-        #  -> Interpolate(Argument(V1, 1), Argument(V2.dual(), 0))
-        #  -> Interpolator(Argument(V1, 0), V2)
-        expr_args = extract_arguments(expr)
-        if expr_args and expr_args[0].number() == 0:
-            v, = expr_args
-            expr = replace(expr, {v: v.reconstruct(number=1)})
-        self.expr_renumbered = expr
+
+        # TODO CrossMeshInterpolator and VomOntoVomXXX are not yet aware of
+        # self.ufl_interpolate (which carries the dual argument).
```
connorjward: Is there a relevant issue to link to here?

firedrake/interpolation.py
```diff
@@ -289,14 +292,42 @@ def __init__(
...
+                              and isinstance(target_mesh.topology, VertexOnlyMeshTopology))
+        if not isinstance(self, SameMeshInterpolator) or vom_onto_other_vom:
+            # For bespoke interpolation, we currently rely on different assembly procedures:
+            # 1) Interpolate(Argument(V1, 1), Argument(V2.dual(), 0)) -> Forward operator (2-form)
+            # 2) Interpolate(Argument(V1, 0), Argument(V2.dual(), 1)) -> Adjoint operator (2-form)
+            # 3) Interpolate(Coefficient(V1), Argument(V2.dual(), 0)) -> Forward action (1-form)
+            # 4) Interpolate(Argument(V1, 0), Cofunction(V2.dual()) -> Adjoint action (1-form)
+            # 5) Interpolate(Coefficient(V1), Cofunction(V2.dual()) -> Double action (0-form)
+
+            # CrossMeshInterpolator._interpolate only supports forward interpolation (cases 1 and 3).
```
connorjward: It would be good to link to an issue here so it's easier for someone else later on to maybe fix this.

firedrake/interpolation.py
connorjward: Yeah I think that's probably better. This should be explained in the docstring.

firedrake/interpolation.py
```diff
@@ -1167,6 +1184,37 @@ def _interpolator(V, tensor, expr, subset, arguments, access, bcs=None):
         return copyin + callables + (parloop_compute_callable, ) + copyout
 
 
+def get_interp_node_map(source_mesh, target_mesh, fs):
+    """Return the cell-to-node map required by a parloop on the target_mesh.cell_set."""
```
connorjward: Can this be a full docstring with type hints? It isn't very clear as is.

connorjward (changes requested):
I think this is really close now. Definitely much much happier with this. You have deleted a lot of complicated code and replaced it with things that I can actually understand!

firedrake/interpolation.py
pbrubeck: What's not clear? This is not public API

pbrubeck: If I rename `fs` to `function_space`, then a docstring will not add any extra information

tests/firedrake/submesh/test_submesh_interpolate.py
pbrubeck: Should we not then striclty XFAIL the test `if comm.size > 1`?

firedrake/interpolation.py
connorjward: Something like:
"Return the map between cells of the target mesh and nodes of the function space. If the function space is defined on the source mesh then the node map is composed with a map between target and source cells."

I'm not saying that it's impossible to understand, just that it could be clearer. It initially confused me.

tests/firedrake/submesh/test_submesh_interpolate.py
connorjward: But if we do that then we stop testing other parts of this test that do in fact work. If you wanted to xfail these then they'd need to be broken out of the rest of the test.

firedrake/interpolation.py
```diff
@@ -1167,6 +1184,37 @@ def _interpolator(V, tensor, expr, subset, arguments, access, bcs=None):
...
+    if isinstance(target_mesh.topology, VertexOnlyMeshTopology):
+        coeff_mesh = fs.mesh()
+        m_ = fs.cell_node_map()
```
```suggestion
def get_interp_node_map(source_mesh, target_mesh, fs):
    """Return the map between cells of the target mesh and nodes of the function space. 
    
    If the function space is defined on the source mesh then the node map is composed 
    with a map between target and source cells.
    """
    if isinstance(target_mesh.topology, VertexOnlyMeshTopology):
        coeff_mesh = fs.mesh()
        m_ = fs.cell_node_map()
```

firedrake/interpolation.py
```suggestion
        # self.ufl_interpolate (which carries the dual argument).
        # See github issue https://github.com/firedrakeproject/firedrake/issues/4592
```

firedrake/interpolation.py
```diff
@@ -278,9 +278,12 @@ def __init__(
         allow_missing_dofs=False,
         matfree=True
     ):
-        if isinstance(expr, ufl.Interpolate):
-            expr, = expr.ufl_operands
-        self.expr = expr
+        if not isinstance(expr, ufl.Interpolate):
```
dham: why would the expression to interpolate ever be a `ufl.Interpolate`

pbrubeck: Currently `Interpolator` is not aware of the underlying `ufl.Interpolate` expression. If we want to generate code for the matfree dual, Interpolator needs to hold both the `operand` and the `dual_arg`.  

I figured that the path of least resistance was to always attach the underlying `ufl.Interpolate` to the `Interpolator`, as opposed to passing down the `argument_slots` as separate things. This PR treats `Interpolator` as if it were `{Zero|One|Two}FormAssembler` by adding an `assemble` method  that turns the symbolic `ufl.Interpolate` into a `scalar`, `(Co)Function` or `Matrix`.

pbrubeck: I am mantaining backwards compatibility and allowing for an `Interpolator` to be constructed from the `operand` (in which case we internally construct the `ufl.Interpolate`), or from the `ufl.Interpolate`

pbrubeck: Just to clarify, the intention was never to interpolate a `ufl.Interpolate`.

tests/firedrake/submesh/test_submesh_interpolate.py
pbrubeck: Done

firedrake/interpolation.py
pbrubeck: I updated the docstring

tests/firedrake/submesh/test_submesh_interpolate.py
```diff
@@ -268,3 +273,78 @@ def expr(m):
...
+
+    # Test forward 1-form (only works in serial for continuous elements)
+    # Matfree forward interpolation with Submesh currently fails in parallel.
+    # The ghost nodes of the parent mesh may be redistributed
+    # into different processes as non-ghost dofs of the submesh.
+    # The submesh kernel will write into ghost nodes of the parent mesh,
+    # but this will be ignored in the halo exchange if access=op2.WRITE.
+
+    # See https://github.com/firedrakeproject/firedrake/issues/4483
+    expected_to_pass = (V2.comm.size == 1 or V2.finat_element.is_dg())
```
pbrubeck: @connorjward there is a chance this randomly passes

connorjward: Bizarre. Maybe we do just need to mask out these tests in parallel with an `if comm.size == 1`.

pbrubeck: I just `assert is_close or not expected_to_pass`

dham (approved):

PR #4580, opened 2025-09-22
Ensemble methods dispatch to MPI methods for non-Firedrake types

JHopeCollins:
This means the call site for `Function`/`Cofunction` looks the same as the call site for non-Firedrake types.
Currently to send a non-Firedrake type:
```python
x = ensemble.ensemble_rank
x = ensemble.ensemble_comm.bcast(x, root=0)
```
This PR changes it to:
```python
x = ensemble.ensemble_rank
x = ensemble.bcast(x, root=0)
```

### DIFF
--- a/firedrake/ensemble/ensemble.py
+++ b/firedrake/ensemble/ensemble.py
@@ -1,10 +1,36 @@
+from functools import wraps
 import weakref
 from itertools import zip_longest
 
 from firedrake.petsc import PETSc
+from firedrake.function import Function
+from firedrake.cofunction import Cofunction
 from pyop2.mpi import MPI, internal_comm
 
 __all__ = ("Ensemble", )
+
+
+def _ensemble_mpi_dispatch(func):
+    @wraps(func)
+    def _mpi_dispatch(self, *args, **kwargs):
+        # if only kwargs are given we need to get the
+        # first kw value instead of the first arg.
+        if len(args) > 0:
+            arg0 = args[0]
+        else:
+            arg0 = [v for v in kwargs.values()][0]
+
+        # dispatch to either our specialised impl
+        # for # Firedrake or the default MPI impl
+        # for everything else.
+        if isinstance(arg0, (Function, Cofunction)):
+            return func(self, *args, **kwargs)
+        else:
+            mpicall = getattr(
+                self.ensemble_comm,
+                func.__name__)
+            return mpicall(*args, **kwargs)
+    return _mpi_dispatch
 
 
 class Ensemble(object):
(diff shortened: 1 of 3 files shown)


firedrake/ensemble/ensemble.py
```diff
@@ -1,12 +1,38 @@
...
 from pyop2.mpi import MPI, internal_comm
 
 __all__ = ("Ensemble", )
 
 
+def _ensemble_mpi_dispatch(func):
+    @wraps(func)
+    def _mpi_dispatch(self, *args, **kwargs):
+        # if only kwargs are given we need to get the
+        # first kw value instead of the first arg.
```
connorjward: I'm not sure about this. One would expect `ensemble.bcast(op=MPI.SUM, f=myfunction)` to be isomorphic to `ensemble.bcast(f=myfunction, op=MPI.SUM)`. If you're doing things with `functools.partial` this might get confusing.
I wonder if it might be a good idea to use `/` in the function signature (https://www.srcecde.me/posts/2024/02/asterisk-and-forward-slash-as-function-parameters-python/).

JHopeCollins: I've changed it to dispatch to the Firedrake implementation if _any_ arg or kwarg is a `Function` or `Cofunction`, so no restrictions on the ordering.

firedrake/ensemble/ensemble.py
```diff
@@ -1,12 +1,32 @@
...
 
 from firedrake.petsc import PETSc
+from firedrake.function import Function
+from firedrake.cofunction import Cofunction
 from pyop2.mpi import MPI, internal_comm
 
 __all__ = ("Ensemble", )
 
 
+def _ensemble_mpi_dispatch(func):
```
connorjward, line 13: `def _overload_comm_method`?

Also I think a docstring saying "Decorator that..." would be helpful. We want something that says what this decorator is.

tests/firedrake/ensemble/test_ensemble_wrapper.py
```diff
@@ -0,0 +1,133 @@
...
+import pytest
+from pytest_mpi.parallel_assert import parallel_assert
+
+
+min_root = 0
+max_root = 1
+
+roots = []
+roots.extend([pytest.param(None, id="root_none")])
+roots.extend([pytest.param(i, id="root_%d" % (i))
```
connorjward:
```suggestion
roots.extend([pytest.param(i, id=f"root_{i}")
```

firedrake/ensemble/ensemble.py
JHopeCollins: Done

connorjward: Do you disagree with my name suggestion? I don't have a strong opinion, just making sure you noticed.

PR #4581, opened 2025-09-22
Include ROL in optimisation tests

angus-g:
# Description

Install ROL through `pyroltrilinos` into the `ci` dependency set, and add to some of the parameterised optimisation tests. Requires https://github.com/dolfin-adjoint/pyadjoint/pull/223.

In the default configurations, TAO's NLS does very well compared to a default Lin-More trust region in ROL. Very likely these tests should use a line search in ROL, I just need to figure out what the parameters are actually called... Even though the tests (seem to) pass, the optimisation doesn't actually converge to the specified tolerance!

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -34,7 +34,7 @@ dependencies = [
   "petsctools",
   "pkgconfig",
   "progress",
-  "pyadjoint-ad>=2025.10.0",
+  "pyadjoint-ad @ git+https://github.com/dolfin-adjoint/pyadjoint.git@master",
   "pycparser",
   "pytools[siphash]",
   "requests",
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -111,6 +111,7 @@ ci = [
   "pdf2image",
   "pygraphviz",
   "pylit",
+  "pyroltrilinos",
   "pytest",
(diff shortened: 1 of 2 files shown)


pyproject.toml
```diff
@@ -114,6 +114,7 @@ ci = [
   "pdf2image",
   "pygraphviz",
   "pylit",
+  "pyroltrilinos",
```
connorjward, line 114: It might be good to enable `pip install firedrake[pyrol]` by adding another dependency group. If so we also add a little bit of info [here](https://www.firedrakeproject.org/install.html#optional-dependencies).

connorjward: Why was this resolved?

pyproject.toml
```diff
@@ -111,6 +111,7 @@ ci = [
...
```
GrahamBenHarper, line 114: As a quick FYI, there is an officially supported version of ROL in python available at https://pypi.org/project/rol-python/

angus-g: Thanks, we'll evaluate moving over. I found also there was no communication from the ROL team on numerous blocking issues I faced and raised, necessitating a fork to allow us to proceed. I should clarify that "we" here is the G-ADOPT team, I'm not speaking for Firedrake generally

PR #4582, opened 2025-09-23
Stop renumbering arguments in the expression to `interpolate`

leo-collins:
This PR removes the argument renumbering 0 -> 1 in the primal expression if V is a FunctionSpace. Thus if you want to interpolate an expression containing arguments (for example if you want to assemble the interpolation operator) then you need to use a `TrialFunction` instead of a `TestFunction`. Recent UFL changes check the argument numbering and raise an error early to avoid confusion.

This is an API change and should be merged after the next release. PR https://github.com/firedrakeproject/firedrake/pull/4572 warns users about this.


tests/firedrake/regression/test_interp_dual.py
```diff
@@ -54,7 +54,7 @@ def test_assemble_interp_adjoint_tensor(mesh, V1, f1):
 
 def test_assemble_interp_operator(V2, f1):
     # Check type
-    If1 = Interpolate(f1, V2)
+    If1 = Interpolate(f1, Argument(V2.dual(), 0))
```
pbrubeck:
```suggestion
    If1 = Interpolate(f1, TestFunction(V2.dual()))
```

firedrake/interpolation.py
```diff
@@ -95,20 +93,17 @@ def __init__(self, expr, v,
                 between a VOM and its input ordering. Defaults to ``True`` which uses SF broadcast
                 and reduce operations.
         """
-        # Check function space
-        expr = ufl.as_ufl(expr)
-        if isinstance(v, functionspaceimpl.WithGeometry):
-            expr_args = extract_arguments(expr)
+        if isinstance(V, functionspaceimpl.WithGeometry):
+            # Need to create a Firedrake Argument so that it has a .function_space() method
+            expr_args = extract_arguments(ufl.as_ufl(expr))
```
pbrubeck, line 98:
```suggestion
            if isinstance(expr, ufl.BaseForm):
                expr_args = [a for a in expr.aguments() if not isinstance(a, Coargument)]
            else:
                expr_args = extract_arguments(ufl.as_ufl(expr)) 
```

```suggestion
            if isinstance(expr, ufl.BaseForm):
                expr_args = [a for a in expr.arguments() if not isinstance(a, Coargument)]
            else:
                expr_args = extract_arguments(ufl.as_ufl(expr))
```

firedrake/interpolation.py
```diff
@@ -95,20 +93,17 @@ def __init__(self, expr, v,
...
             is_adjoint = len(expr_args) and expr_args[0].number() == 0
-            v = Argument(v.dual(), 1 if is_adjoint else 0)
+            V = Argument(V.dual(), 1 if is_adjoint else 0)
```
pbrubeck: Put 0 if it is not there, otherwise keep increasing the count
```suggestion
            arg_numbers = [a.number() for a in expr_args]
            number = 0 if len(expr_args) == 0 or min(arg_numbers) > 0 else max(arg_numbers) + 1
            V = Argument(V.dual(), number)            
```

firedrake/interpolation.py
pbrubeck: Not sure if filtering all Coarguments is a sensible thing, but imagine if a `Function` were to advertise its Coargument (technically it should have one), we would have to discard it. An Interpolate is a BaseFormOperator that does advertise the Coargument, but here we should only view the operand as the output after assembling it.

firedrake/interpolation.py
pbrubeck: `is_adjoint` goes away.

pbrubeck: Similar fixes will be needed in UFL

firedrake/interpolation.py
pbrubeck: Maybe the more sensible thing is to drop the last Coargument?

firedrake/interpolation.py
```diff
@@ -48,7 +46,7 @@
 
 class Interpolate(ufl.Interpolate):
 
-    def __init__(self, expr, v,
+    def __init__(self, expr, V,
```
pbrubeck, line 48: This is very unrelated, but I think that a much more friendly interface is to allow either or both left and right arguments to be a primal FunctionSpace. 

Right now we do this under the hood

`Interpolate(Function(V1), V2) -> Interpolate(Function(V1), Argument(V2.dual(), 0)) `

It'd be reasonable to have a similar shortcut for the adjoint. When the left argument is a FunctionSpace, we would then automatically create the Argument for it. 

`Interpolate(V1, Cofunction(V2.dual())) -> Interpolate(Argument(V1, 0), Cofunction(V2.dual())) `

And supplying two FunctionSpaces is a perfectly natural interface:

`Interpolate(V1, V2) -> Interpolate(Argument(V1, 1), Argument(V2.dual(), 0)) `

Of course we need to arbitrarily decide who gets the lowest number, the more intuitive numbering that produces the forward Interpolation is to go from right to left.

pbrubeck: Thoughts @dham ?

firedrake/interpolation.py
pbrubeck: In UFL we're checking for contiguous argument numbers, I think this approach is more fool-proof: just get the smallest number that's missing
```suggestion
            arg_numbers = set(a.number() for a in expr_args)
            number = min(set(range(len(expr_args) + 1)) - arg_numbers)
            V = Argument(V.dual(), number)
```

leo-collins: `expr_args` can only contain zero or one arguments so I think this logic is more confusing than `is_adjoint`

firedrake/interpolation.py
```diff
@@ -95,20 +93,18 @@ def __init__(self, expr, v,
...
-        # Check function space
         expr = ufl.as_ufl(expr)
-        if isinstance(v, functionspaceimpl.WithGeometry):
-            expr_args = extract_arguments(expr)
-            is_adjoint = len(expr_args) and expr_args[0].number() == 0
-            v = Argument(v.dual(), 1 if is_adjoint else 0)
+        if isinstance(V, functionspaceimpl.WithGeometry):
+            # Need to create a Firedrake Argument so that it has a .function_space() method
+            expr_arg_numbers = {arg.number() for arg in extract_arguments(expr) if not is_dual(arg)}
+            is_adjoint = len(expr_arg_numbers) and expr_arg_numbers == {0}
```
```suggestion
            is_adjoint = expr_arg_numbers == {0}
```

firedrake/interpolation.py
```diff
@@ -95,20 +93,18 @@ def __init__(self, expr, v,
...
         """
-        # Check function space
         expr = ufl.as_ufl(expr)
-        if isinstance(v, functionspaceimpl.WithGeometry):
-            expr_args = extract_arguments(expr)
-            is_adjoint = len(expr_args) and expr_args[0].number() == 0
-            v = Argument(v.dual(), 1 if is_adjoint else 0)
+        if isinstance(V, functionspaceimpl.WithGeometry):
+            # Need to create a Firedrake Argument so that it has a .function_space() method
+            expr_arg_numbers = {arg.number() for arg in extract_arguments(expr) if not is_dual(arg)}
```
pbrubeck: Are we sure that dropping all the dual arguments the right thing? We should only drop the dual argument of the source space (if it appears in the expression)

firedrake/interpolation.py
pbrubeck: We should then `assert len(expr_args) <= 1`

firedrake/interpolation.py
dham: We need to verify/ensure that `extract_arguments` returns free arguments only. `extract_free_arguments` this would when visiting a FormBase, drop the dual argument in position 0.

firedrake/interpolation.py
dham: I think we don't need to fix this. We have lots more issues if we want to support rank 3 tensors, even symbolically.

firedrake/interpolation.py
connorjward: Resolving so that this PR may be merged. Feel free to put this information elsewhere.

PR #4587, opened 2025-09-24
Pass kwargs to Cofunction.interpolate

pbrubeck:
# Description

### DIFF
--- a/firedrake/cofunction.py
+++ b/firedrake/cofunction.py
@@ -331,9 +331,9 @@ class Cofunction(ufl.Cofunction, CofunctionMixin):
         firedrake.cofunction.Cofunction
             Returns `self`
         """
-        from firedrake import interpolate, assemble
+        from firedrake import interpolation, assemble
         v, = self.arguments()
-        interp = interpolate(v, expression, **kwargs)
+        interp = interpolation.Interpolate(v, expression, **kwargs)
         return assemble(interp, tensor=self, ad_block_tag=ad_block_tag)
 
     @property
--- a/firedrake/function.py
+++ b/firedrake/function.py
@@ -382,9 +382,9 @@ class Function(ufl.Coefficient, FunctionMixin):
         firedrake.function.Function
             Returns `self`
         """
-        from firedrake import interpolate, assemble
+        from firedrake import interpolation, assemble
         V = self.function_space()
-        interp = interpolate(expression, V, **kwargs)
+        interp = interpolation.Interpolate(expression, V, **kwargs)
         return assemble(interp, tensor=self, ad_block_tag=ad_block_tag)
 
     def zero(self, subset=None):


firedrake/cofunction.py
```diff
@@ -308,14 +308,45 @@ def __imul__(self, expr):
...
+        self,
+        expression,
+        subset=None,
+        allow_missing_dofs=False,
+        default_missing_val=None,
+        ad_block_tag=None
+    ):
+        r"""Interpolate a dual expression onto this :class:`Cofunction`.
+
+        :param expression: a dual UFL expression to interpolate
```
connorjward: Following our discussion of style please use numpydoc and type hints.

firedrake/cofunction.py
```diff
@@ -308,14 +308,45 @@ def __imul__(self, expr):
...
-        r"""Interpolate an expression onto this :class:`Cofunction`.
-
-        :param expression: a UFL expression to interpolate
+    @PETSc.Log.EventDecorator()
+    def interpolate(
+        self,
+        expression,
+        subset=None,
+        allow_missing_dofs=False,
+        default_missing_val=None,
```
connorjward: I would be tempted to just make these be `**kwargs`

pbrubeck: I like this because it spares the need for a numpy docstring. The only issue is that we cannot pass a `tensor`

pbrubeck: I think we could just `assert "tensor" not in kwargs`

connorjward: You could check for `"tensor" in kwargs`? The issue with things as they are is future changes to the `interpolate` interface mean we have to make a bunch more cascading changes elsewhere.

pbrubeck: Oh, NVM `tensor` is a kwarg for `assemble`, not for `interpolate`. The docstring could mention that all extra kwargs are those of `interpolate`.

We still need `ad_block_tag` to be an explicit kwarg because it is used by `assemble`.

connorjward: Yes that's what I'm suggesting. Make `**kwargs` only be those that we pass through to `interpolate`.

firedrake/cofunction.py
```diff
@@ -308,14 +308,32 @@ def __imul__(self, expr):
...
+        """Interpolate a dual expression onto this :class:`Cofunction`.
+
+        Parameters
+        ----------
+        expression: ufl.BaseForm
+                    A dual UFL expression to interpolate.
+
+        ad_block_tag: str
+                      An optional string for tagging the resulting assemble
+                      block on the Pyadjoint tape.
```
connorjward: You can document kwargs here (e.g. see https://numpydoc.readthedocs.io/en/latest/format.html#parameters)

firedrake/cofunction.py
```diff
@@ -308,14 +308,32 @@ def __imul__(self, expr):
...
-        from firedrake import interpolation
-        interp = interpolation.Interpolate(ufl_expr.Argument(self.function_space().dual(), 0), expression)
-        return firedrake.assemble(interp, tensor=self)
+    @PETSc.Log.EventDecorator()
+    def interpolate(self, expression, ad_block_tag=None, **kwargs):
+        """Interpolate a dual expression onto this :class:`Cofunction`.
+
+        Parameters
+        ----------
+        expression: ufl.BaseForm
```
connorjward:
```suggestion
        expression:
```
put the type in the function signature instead.

pbrubeck: That's not the style for the rest of the file

connorjward: But it's what we should be doing. This is now in the [coding guide](https://github.com/firedrakeproject/firedrake/wiki/Firedrake-coding-guide).

> Docstrings should be written in [numpydoc format](https://numpydoc.readthedocs.io/en/latest/format.html). Since type hinting is used it is not necessary to specify the type of function arguments in the docstring (unfortunately the return type still needs to be documented, [ref](https://github.com/sphinx-doc/sphinx/issues/5887)). Occasionally it is fine to have a different type description in the numpydoc if it would otherwise expose a complicated internal detail (e.g. WithGeometry vs FunctionSpace).

firedrake/cofunction.py
```diff
@@ -308,14 +308,33 @@ def __imul__(self, expr):
...
+
+        Parameters
+        ----------
+        expression
+            A dual UFL expression to interpolate.
+        ad_block_tag
+            An optional string for tagging the resulting assemble
+            block on the Pyadjoint tape.
+        **kwargs
+            Any extra kwargs are passed on to the interpolate method.
```
connorjward:
```suggestion
            Any extra kwargs are passed on to the interpolate function.
```

firedrake/function.py
```diff
@@ -360,43 +360,31 @@ def function_space(self):
...
+
+        Parameters
+        ----------
+        expression
+            A UFL expression to interpolate.
+        ad_block_tag
+            An optional string for tagging the resulting assemble
+            block on the Pyadjoint tape.
+        **kwargs
+            Any extra kwargs are passed on to the interpolate method.
```
```suggestion
            Any extra kwargs are passed on to the interpolate function.
```

connorjward (approved):

PR #4595, opened 2025-09-26
Leo/refactor interpolate

leo-collins:


firedrake/assemble.py
```diff
@@ -556,6 +557,8 @@ def base_form_assembly_visitor(self, expr, tensor, *args):
             result = expr.assemble(assembly_opts=opts)
             return tensor.assign(result) if tensor else result
         elif isinstance(expr, ufl.Interpolate):
+            if not isinstance(expr, firedrake.Interpolate):
+                expr = firedrake.Interpolate(*reversed(expr.dual_args()))
```
dham: @pbrubeck why can this ever happen?

firedrake/interpolation.py
```diff
@@ -42,258 +45,188 @@
...
-            "removed. Use the `interpolate` function instead."
-        )
+        # TODO CrossMeshInterpolator and VomOntoVomXXX are not yet aware of
+        # self.ufl_interpolate (which carries the dual argument).
+        target_mesh = as_domain(V)
+        source_mesh = extract_unique_domain(operand) or target_mesh
+        vom_onto_other_vom = ((source_mesh is not target_mesh)
+                              and isinstance(source_mesh.topology, VertexOnlyMeshTopology)
+                              and isinstance(target_mesh.topology, VertexOnlyMeshTopology))
+        if not isinstance(self, SameMeshInterpolator) or vom_onto_other_vom:
```
dham: I don't think this is nice inheritance. This smells like the class hierarchy is wrong.

firedrake/interpolation.py
```diff
@@ -539,7 +468,7 @@ def _interpolate(
                 V_dest = self.expr.function_space().dual()
             except AttributeError:
                 if self.nargs:
-                    V_dest = self.arguments[0].function_space().dual()
+                    V_dest = self.arguments[-1].function_space().dual()
```
dham: Surely argument 0 is the target space?

firedrake/interpolation.py
```diff
@@ -582,22 +511,22 @@ def _interpolate(
...
+                    assemble(self.point_eval)
                 )
             else:
                 f_src_at_dest_node_coords_src_mesh_decomp = (
-                    assemble(action(self.point_eval_interpolate, f_src))
+                    assemble(action(self.point_eval, f_src))
                 )
             f_src_at_dest_node_coords_dest_mesh_decomp = firedrake.Function(
-                self.to_input_ordering_interpolate.function_space()
+                self.point_eval_io.function_space()
```
dham: io is obscure. prefer to write out input_ordering.

firedrake/interpolation.py
```diff
@@ -976,6 +878,32 @@ def _interpolator(V, tensor, expr, subset, arguments, access, bcs=None):
...
+
+    needs_weight = isinstance(dual_arg, ufl.Cofunction) and not to_element.is_dg()
+    if needs_weight:
+        # Compute the reciprocal of the DOF multiplicity
+        W = dual_arg.function_space()
+        wsize = W.finat_element.space_dimension() * W.block_size
+        kernel_code = f"""
+        void multiplicity(PetscScalar *restrict w) {{
+            for (PetscInt i=0; i<{wsize}; i++) w[i] += 1;
+        }}"""
```
dham, line 834: Do we strictly need a new C kernel here?

firedrake/interpolation.py
pbrubeck: The target_mesh becomes the input for dual Interpolation. We are now allowing 2 arguments.

firedrake/interpolation.py
pbrubeck: The kernel needs to be executed on the target mesh. We could use a one-liner

`assemble(slate.AssembledVector(Cofunction(V).assign(1)))`

but the use of slate here might be harder to understand.

connorjward: Would interpolating a CG function of 1s into another CG function using `INC` DTRT?

firedrake/interpolation.py
```diff
@@ -46,261 +46,142 @@
...
+        then the values are either (a) unchanged if some ``output`` is given to
+        the :meth:`interpolate` method or (b) set to zero.
+        Ignored if interpolating within the same mesh or onto a :func:`.VertexOnlyMesh`.
+    matfree : bool, default True
+        If ``False``, then construct the permutation matrix for interpolating
+        between a VOM and its input ordering. Defaults to ``True`` which uses SF broadcast
+        and reduce operations.
+    """
+    subset: Optional[op2.Subset] = None
+    access: Access = op2.WRITE
```
pbrubeck: We should set `None` as default, since `op2.WRITE` is not right for the matfree adjoint

firedrake/interpolation.py
```diff
@@ -339,25 +220,7 @@ def __init__(
 
         if not isinstance(dual_arg, ufl.Coargument):
             # Matrix-free assembly of 0-form or 1-form requires INC access
-            if access and access != op2.INC:
-                raise ValueError("Matfree adjoint interpolation requires INC access")
```
pbrubeck, line 326: I think we should keep raising this error here

pbrubeck: Or move this to `SameMeshInterpolator` constructor or `_get_callable`

firedrake/interpolation.py
pbrubeck: That's a bit insane but yes.

pbrubeck: The interpolation would generate a kernel, pretty much like this one.

pbrubeck: We need to specify the subset kwarg in order to get submesh right

firedrake/assemble.py
```diff
@@ -602,14 +603,10 @@ def base_form_assembly_visitor(self, expr, tensor, *args):
...
 
-            # Get the interpolator
-            interp_data = expr.interp_data.copy()
-            default_missing_val = interp_data.pop('default_missing_val', None)
             if rank == 1 and isinstance(tensor, firedrake.Function):
                 V = tensor
-            interpolator = firedrake.Interpolator(expr, V, **interp_data)
-            # Assembly
-            return interpolator.assemble(tensor=tensor, default_missing_val=default_missing_val)
+            interpolator = _get_interpolator(expr, V)
```
pbrubeck: Any chance we can cache the `interpolator` on the `expr`? We do something similar for the `FormAssembler`

firedrake/interpolation.py
```diff
@@ -747,7 +567,7 @@ def _interpolate(
             # don't have to worry about skipping over missing points here
             # because I'm going from the input ordering VOM to the original VOM
             # and all points from the input ordering VOM are in the original.
-            interp = action(expr_adjoint(self.to_input_ordering_interpolate), f_src_at_dest_node_coords_dest_mesh_decomp)
+            interp = action(expr_adjoint(self.point_eval_input_ordering), f_src_at_dest_node_coords_dest_mesh_decomp)
```
pbrubeck: Do we need an adjoint here? Could swapping the arguments to action not work?
```suggestion
            interp = action(f_src_at_dest_node_coords_dest_mesh_decomp, self.point_eval_input_ordering)
```

firedrake/interpolation.py
```diff
@@ -747,7 +567,7 @@ def _interpolate(
...
             f_src_at_src_node_coords = assemble(interp)
```
pbrubeck: There's no need to call `assemble` at this point
```suggestion
            f_src_at_src_node_coords = interp
```

firedrake/interpolation.py
```diff
@@ -760,7 +580,7 @@ def _interpolate(
             # SameMeshInterpolator.interpolate did not effect the result. For
             # now, I say in the docstring that it only applies to forward
             # interpolation.
-            interp = action(expr_adjoint(self.point_eval_interpolate), f_src_at_src_node_coords)
+            interp = action(expr_adjoint(self.point_eval), f_src_at_src_node_coords)
```
```suggestion
            interp = action(f_src_at_src_node_coords, self.point_eval)
```

firedrake/interpolation.py
```diff
@@ -46,261 +46,142 @@
...
-            v = Argument(v.dual(), 1 if is_adjoint else 0)
-
-        V = v.arguments()[0].function_space()
-        if len(expr.ufl_shape) != len(V.value_shape):
-            raise RuntimeError(f'Rank mismatch: Expression rank {len(expr.ufl_shape)}, FunctionSpace rank {len(V.value_shape)}')
+            V = Argument(V.dual(), 1 if is_adjoint else 0)
+        super().__init__(expr, V)
 
-        if expr.ufl_shape != V.value_shape:
-            raise RuntimeError('Shape mismatch: Expression shape {expr.ufl_shape}, FunctionSpace shape {V.value_shape}')
```
pbrubeck: Why are we getting rid of these checks?

firedrake/interpolation.py
```diff
@@ -401,11 +261,28 @@ def assemble(self, tensor=None, default_missing_val=None):
...
-                return assemble(ufl.Action(*cofunctions, Iu), tensor=tensor)
+                Iu = self._interpolate()
+                return firedrake.assemble(ufl.Action(*cofunctions, Iu), tensor=tensor)
             else:
-                return self._interpolate(*cofunctions, output=tensor, adjoint=needs_adjoint,
-                                         default_missing_val=default_missing_val)
+                return self._interpolate(*cofunctions, output=tensor, adjoint=needs_adjoint)
+
+
+def _get_interpolator(expr: Interpolate, V) -> Interpolator:
```
```suggestion
def _get_interpolator(expr: ufl.classes.Expr | Interpolate, V) -> Interpolator:
```

firedrake/interpolation.py
```diff
@@ -401,11 +261,28 @@ def assemble(self, tensor=None, default_missing_val=None):
...
+    target_mesh = as_domain(V)
+    source_mesh = extract_unique_domain(expr) or target_mesh
```
```suggestion
    if isinstance(expr, ufl.Interpolate):
        operand, = expr.ufl_operands
    else:
        operand = expr
    target_mesh = as_domain(V)
    source_mesh = extract_unique_domain(operand) or target_mesh
```

pbrubeck: This pattern is all over the place, we should probably make it into a function.

firedrake/interpolation.py
```diff
@@ -46,261 +46,142 @@
...
         self.V = V
-        self.subset = subset
-        self.freeze_expr = freeze_expr
+        self.options = expr.options
         self.bcs = bcs
-        self._allow_missing_dofs = allow_missing_dofs
-        self.matfree = matfree
         self.callable = None
 
         # TODO CrossMeshInterpolator and VomOntoVomXXX are not yet aware of
```
pbrubeck: the if statement below should now say `if isinstance(self, (CrossMeshInterpolator, VomOntoVomInterpolator))`

firedrake/interpolation.py
```diff
@@ -861,82 +772,27 @@ def _interpolate(self, *function, output=None, transpose=None, adjoint=False, **
...
-        if len(Vrow) > 1 or len(Vcol) > 1:
-            raise NotImplementedError("Interpolation of mixed expressions with arguments is not supported")
-        if isinstance(target_mesh.topology, VertexOnlyMeshTopology) and target_mesh is not source_mesh and not vom_onto_other_vom:
-            if not isinstance(target_mesh.topology, VertexOnlyMeshTopology):
-                raise NotImplementedError("Can only interpolate onto a Vertex Only Mesh")
-            if target_mesh.geometric_dimension() != source_mesh.geometric_dimension():
-                raise ValueError("Cannot interpolate onto a mesh of a different geometric dimension")
-            if not hasattr(target_mesh, "_parent_mesh") or target_mesh._parent_mesh is not source_mesh:
-                raise ValueError("Can only interpolate across meshes where the source mesh is the parent of the target")
+class VomOntoVomInterpolator(SameMeshInterpolator):
```
pbrubeck, line 686:
```suggestion
class CrossVomInterpolator(SameMeshInterpolator):
```

firedrake/interpolation.py
```diff
@@ -775,63 +595,154 @@ class SameMeshInterpolator(Interpolator):
...
+                                        nest=False,
+                                        block_sparse=True)
+                tensor = op2.Mat(sparsity)
+            f = tensor
+        else:
+            raise ValueError(f"Cannot interpolate an expression with {rank} arguments")
+        return f, tensor
+
+    def _get_callable(self):
+        expr = self.ufl_interpolate_renumbered
```
pbrubeck: `SameMeshInterpolator` only required the renumbering for VomOntoVom. We can now remove it from here.
```suggestion
        expr = self.ufl_interpolate
```

firedrake/interpolation.py
```diff
@@ -46,261 +46,142 @@
...
+
+    Attributes
+    ----------
+    subset : pyop2.types.set.Subset, optional
+        An optional subset to apply the interpolation over.
+        Cannot, at present, be used when interpolating across meshes unless
+        the target mesh is a :func:`.VertexOnlyMesh`.
+    access : pyop2.types.access.Access, default op2.WRITE
+        The pyop2 access descriptor for combining updates to shared
+        DoFs. Possible values include ``WRITE`` and ``INC``. Only ``WRITE`` is
```
```suggestion
        DoFs. Possible values include ``WRITE``, ``MIN``, ``MAX``, and ``INC``. Only ``WRITE`` is
```

firedrake/interpolation.py
```diff
@@ -46,261 +46,142 @@
...
+        supported at present when interpolating across meshes unless the target
+        mesh is a :func:`.VertexOnlyMesh`.
```
```suggestion
        mesh is a :func:`.VertexOnlyMesh`. Only ``INC`` is supported for the matrix-free adjoint interpolation.
```

firedrake/interpolation.py
```diff
@@ -46,261 +46,142 @@
...
+
+        .. note::
```
pbrubeck: The way we handle INC now always zeroes the ouput, so this is out of date. We might want another kwarg `needs_zeroing` like the one we have for `assemble`, the default is True.

firedrake/interpolation.py
```diff
@@ -474,159 +338,115 @@ def __init__(
...
+            # In this case all sub elements are equal
+            base_element = dest_element.sub_elements[0]
+            if base_element.reference_value_shape != ():
+                raise NotImplementedError(
+                    "Can't yet cross-mesh interpolate onto function spaces made from VectorElements or TensorElements made from sub elements with value shape other than ()."
+                )
+            self.dest_element = base_element
+            self._get_symbolic_expressions()
+        elif isinstance(dest_element, finat.ufl.MixedElement):
+            self._mixed_function_space()
```
pbrubeck: Question for the `MixedInterpolator` PR: Can `CrossMeshInterpolator` use the adjoint for mixed function spaces?

firedrake/interpolation.py
```diff
@@ -474,159 +338,115 @@ def __init__(
...
+        # Interpolate into the input-ordering VOM
+        P0DG_vom_i_o = fs_type(self.vom.input_ordering, "DG", 0)
+        self.point_eval_input_ordering = interpolate(firedrake.TrialFunction(P0DG_vom), P0DG_vom_i_o)
+
+    def _mixed_function_space(self):
+        """Builds symbolic Interpolate expressions for each sub-element of a MixedFunctionSpace.
+        """
+        # NOTE: since we can't have expressions for MixedFunctionSpaces
+        # we know that the input argument ``expr`` must be a Function.
+        # V_dest can be a Function or a FunctionSpace, and subfunctions works for both.
```
pbrubeck: To answer my `MixedInterpolator` question, According to this, A mixed `CrossMeshInterpolator` only works for the matfree forward action.

pbrubeck: This is the same situation as `SameMeshInterpolator`. They both support the mixed forward action. `MixedInterpolator` does not implement the mixed forward action.

firedrake/assemble.py
pbrubeck: I removed this from the parent PR as it was quite irrelevant. This shouldn't be normally happening, only if someone mistakenly wanted to `assemble` a `ufl.Interpolate`,

firedrake/assemble.py
pbrubeck: This does not need to happen in this PR, but we could start thinking about this.

firedrake/interpolation.py
leo-collins: I'll add this one back but it should go in UFL really

firedrake/cofunction.py
```diff
@@ -331,9 +331,9 @@ def interpolate(self,
         firedrake.cofunction.Cofunction
             Returns `self`
         """
-        from firedrake import interpolation, assemble
+        from firedrake import interpolate, assemble
         v, = self.arguments()
-        interp = interpolation.Interpolate(v, expression, **kwargs)
```
pbrubeck: This was necessary to keep adjoint tests happy, does this cause issues still?

leo-collins: I think this is fixed by https://github.com/firedrakeproject/firedrake/pull/4582, which is included in this PR

firedrake/interpolation.py
```diff
@@ -401,11 +266,40 @@ def assemble(self, tensor=None, default_missing_val=None):
...
+    submesh_interp_implemented = \
+        all(isinstance(m.topology, firedrake.mesh.MeshTopology) for m in [target_mesh, source_mesh]) and \
+        target_mesh.submesh_ancesters[-1] is source_mesh.submesh_ancesters[-1] and \
+        target_mesh.topological_dimension() == source_mesh.topological_dimension()
+    if target_mesh is source_mesh or submesh_interp_implemented:
+        return SameMeshInterpolator(expr, V, bcs=bcs)
+    else:
+        if isinstance(target_mesh.topology, VertexOnlyMeshTopology):
+            if isinstance(source_mesh.topology, VertexOnlyMeshTopology):
+                return VomOntoVomInterpolator(expr, V, bcs=bcs)
```
pbrubeck: We might also need to check for mixedness for this one. Does VomOntoVom only only occur as an internal step to CrossMeshInterpolator? If that is the case we shouldn't be too worried because CrossMeshInterpolator will be always non-mixed.

firedrake/bcs.py
```diff
@@ -359,14 +360,12 @@ def function_arg(self, g):
         elif isinstance(g, ufl.classes.Expr):
             if g.ufl_shape != V.value_shape:
                 raise RuntimeError(f"Provided boundary value {g} does not match shape of space")
-            try:
+            disallowed_elements = PhysicallyMappedElement | DirectlyDefinedElement | EnrichedElement
+            if all(not isinstance(element, disallowed_elements) for element in V.ufl_element().sub_elements):
```
leo-collins: @pbrubeck Is this the correct thing to do?

pbrubeck: I think the more pythonic way is to try and catch the `NotImplementedError`. Maybe we need to fix finat to replace the `AttributeError` with `NotImplementedError`

firedrake/bcs.py
```diff
@@ -359,14 +360,16 @@ def function_arg(self, g):
...
+            if all(not isinstance(element, disallowed_elements) for element in elements):
                 self._function_arg = firedrake.Function(V)
-                # Use `Interpolator` instead of assembling an `Interpolate` form
-                # as the expression compilation needs to happen at this stage to
-                # determine if we should use interpolation or projection
-                #  -> e.g. interpolation may not be supported for the element.
-                self._function_arg_update = firedrake.Interpolator(g, self._function_arg)._interpolate
-            except (NotImplementedError, AttributeError):
+                interpolate_expr = firedrake.interpolate(g, V)
+                self._function_arg_update = lambda: firedrake.assemble(interpolate_expr, tensor=self._function_arg)
```
pbrubeck: This will not reuse the `Interpolator` object

pbrubeck: I think we want `get_interpolator` to become the public facing API and use that here, as we have for `get_assembler`. 

Alternatively we could enhance `get_assembler(interpolate())` to return the reusable `Interpolator`.

firedrake/bcs.py
```diff
@@ -361,12 +363,11 @@ def function_arg(self, g):
...
-                # as the expression compilation needs to happen at this stage to
-                # determine if we should use interpolation or projection
-                #  -> e.g. interpolation may not be supported for the element.
-                self._function_arg_update = firedrake.Interpolator(g, self._function_arg)._interpolate
-            except (NotImplementedError, AttributeError):
+                interpolator = get_interpolator(firedrake.interpolate(g, V))
+                # Call this here to check if the element supports interpolation 
+                interpolator._get_callable()
+                self._function_arg_update = lambda: interpolator.assemble(tensor=self._function_arg)
+            except (ValueError, NotImplementedError):
```
pbrubeck: Why is this `ValueError` showing up? We might want to change it to `NotImplementedError`

firedrake/bcs.py
```diff
@@ -361,12 +363,11 @@ def function_arg(self, g):
...
-                # Use `Interpolator` instead of assembling an `Interpolate` form
-                # as the expression compilation needs to happen at this stage to
-                # determine if we should use interpolation or projection
-                #  -> e.g. interpolation may not be supported for the element.
-                self._function_arg_update = firedrake.Interpolator(g, self._function_arg)._interpolate
-            except (NotImplementedError, AttributeError):
+                interpolator = get_interpolator(firedrake.interpolate(g, V))
+                # Call this here to check if the element supports interpolation 
+                interpolator._get_callable()
+                self._function_arg_update = lambda: interpolator.assemble(tensor=self._function_arg)
```
pbrubeck: I think we can interpolate in-place if we pass `self._function_arg` as `V` to the `Interpolator`

firedrake/interpolation.py
```diff
@@ -46,328 +47,148 @@
...
-            "The 'interpolate' method on `Interpolator` objects has been "
-            "removed. Use the `interpolate` function instead."
-        )
+        dual_arg, operand = expr.argument_slots()
+        self.expr = expr
+        self.expr_args = expr.arguments()
+        self.rank = len(self.expr_args)
+        self.operand = operand
+        self.dual_arg = dual_arg
+        self.V_dest = dual_arg.function_space().dual()
```
pbrubeck: This name might be a bit confusing. Maybe `self.target_space` is better?

firedrake/interpolation.py
```diff
@@ -46,328 +47,148 @@
...
+        self.target_mesh = as_domain(self.V_dest)
```
pbrubeck: We had `as_domain(V)` because `V` could be either a `Function` or `FunctionSpace`. If V_dest is a FunctionSpace, then 
```suggestion
        self.target_mesh = self.V_dest.mesh()
```

firedrake/interpolation.py
```diff
@@ -46,328 +47,148 @@
...
+        bcs : list, optional
+            List of boundary conditions to zero-out in the output function space. By default None.
         """
-        raise FutureWarning(
-            "The 'interpolate' method on `Interpolator` objects has been "
-            "removed. Use the `interpolate` function instead."
-        )
+        dual_arg, operand = expr.argument_slots()
+        self.expr = expr
+        self.expr_args = expr.arguments()
```
pbrubeck: `expr` is unclear, should we call this `ufl_interpolate`?

firedrake/interpolation.py
```diff
@@ -379,41 +200,57 @@ def _interpolate(self, *args, **kwargs):
...
         """Assemble the operator (or its action)."""
-        from firedrake.assemble import assemble
-        needs_adjoint = self.ufl_interpolate_renumbered != self.ufl_interpolate
-        arguments = self.ufl_interpolate.arguments()
-        if len(arguments) == 2:
+        if self.rank == 2:
             # Assembling the operator
             res = tensor.petscmat if tensor else PETSc.Mat()
             # Get the interpolation matrix
+            self._get_callable()
```
pbrubeck: get_callable gives the impression that this method is supposed to return something. Should it be renamed to set_callable?

firedrake/interpolation.py
```diff
@@ -453,272 +290,169 @@ class CrossMeshInterpolator(Interpolator):
...
-                    V_dest = self.arguments[-1].function_space().dual()
-                else:
-                    coeffs = extract_coefficients(self.expr)
-                    if len(coeffs):
-                        V_dest = coeffs[0].function_space().dual()
-                    else:
-                        raise ValueError(
-                            "Can't adjoint interpolate an expression with no coefficients or arguments."
-                        )
+        arg = Argument(self.P0DG_vom, 0 if self.expr.is_adjoint else 1)
```
pbrubeck: Are we just reconstructing the old argument on a different space? We could call `old_argument.reconstruct(V=self.P0DG_vom)` here if this is the case

firedrake/interpolation.py
```diff
@@ -453,272 +290,169 @@ class CrossMeshInterpolator(Interpolator):
...
-
-            # we can now confidently assign this to a function on V_dest
-            if self._allow_missing_dofs and default_missing_val is None:
-                indices = numpy.where(
-                    ~numpy.isnan(f_src_at_dest_node_coords_dest_mesh_decomp.dat.data_ro)
-                )[0]
-                output.dat.data_wo[
-                    indices
-                ] = f_src_at_dest_node_coords_dest_mesh_decomp.dat.data_ro[indices]
+            V_dest = self.V_dest
```
pbrubeck: This is why I was saying that `self.V_dest` is not a very good name.

firedrake/interpolation.py
```diff
@@ -453,272 +290,169 @@ class CrossMeshInterpolator(Interpolator):
...
-            else:
-                V_dest = self.V
-        if output:
-            if output.function_space() != V_dest:
-                raise ValueError("Given output has the wrong function space!")
+            symbolic = action(self.point_eval_input_ordering, self.point_eval)
+        self.handle = assemble(symbolic).petscmat
+        self.callable = lambda: self
+
+    def _interpolate(self, output=None):
```
pbrubeck: The previous interface allowed for a single `CrossMeshInterpolator` to be reused for the forward and adjoint actions, allowing reuse of the `VertexOnlyMesh`, which might be very expensive to set up. This new interface requires separate `Interpolator` instances for each operation.

pbrubeck: I think we want to keep the `adjoint` kwarg here
```suggestion
    def _interpolate(self, output=None, adjoint=False):
```

leo-collins: Maybe we could catch this when we set up caching?

firedrake/interpolation.py
pbrubeck: Could we not refer to `self.expr.target_space` instead?

firedrake/interpolation.py
```diff
@@ -25,395 +27,288 @@
...
-        super().__init__(expr, v)
-
-        # -- Interpolate data (e.g. `subset` or `access`) -- #
-        self.interp_data = {"subset": subset,
-                            "access": access,
-                            "allow_missing_dofs": allow_missing_dofs,
-                            "default_missing_val": default_missing_val,
-                            "matfree": matfree}
+        expr_arg_numbers = {arg.number() for arg in extract_arguments(expr) if not is_dual(arg)}
+        self.is_adjoint = len(expr_arg_numbers) and expr_arg_numbers == {0}
```
```suggestion
        self.is_adjoint = expr_arg_numbers == {0}
```

firedrake/bcs.py
pbrubeck: Looks like this PR removes in-place access to the output tensor (if passed to the constructor). But the old way of doing it was not the best one. We can reimplement it correctly in `SameMeshInterpolator` by replacing the `dat` in the first parloop argument.

pbrubeck: For now I think this is fine
```suggestion
                self._function_arg_update = partial(interpolator.assemble, tensor=self._function_arg)
```

firedrake/interpolation.py
```diff
@@ -25,395 +27,288 @@
...
-                Iu = self._interpolate(default_missing_val=default_missing_val)
-                return assemble(ufl.Action(*cofunctions, Iu), tensor=tensor)
-            else:
-                return self._interpolate(*cofunctions, output=tensor, adjoint=needs_adjoint,
-                                         default_missing_val=default_missing_val)
+            assert isinstance(tensor, Function | Cofunction | None)
+            if tensor and isinstance(result, Function | Cofunction):
+                tensor.assign(result)
+                return tensor
+            return result
```
```suggestion
            assert isinstance(tensor, Function | Cofunction | None)
            return tensor.assign(result) if tensor else result
```

firedrake/interpolation.py
```diff
@@ -25,395 +27,279 @@
...
+    """
+    def __init__(self, expr: Interpolate):
+        dual_arg, operand = expr.argument_slots()
+        self.expr = expr
+        self.expr_args = expr.arguments()
+        self.rank = len(self.expr_args)
+        self.operand = operand
+        self.dual_arg = dual_arg
+        self.target_space = dual_arg.function_space().dual()
+        self.target_mesh = as_domain(self.target_space)
```
```suggestion
        self.target_mesh = self.target_space.mesh()
```

firedrake/interpolation.py
```diff
@@ -25,395 +27,279 @@
...
+        self.matfree = expr.options.matfree
+        self.bcs = expr.options.bcs
+        self.callable = None
+        self.access = expr.options.access
 
     @abc.abstractmethod
-    def _interpolate(self, *args, **kwargs):
-        """
-        Compute the interpolation operation of interest.
+    def _build_callable(self, output: Function | Cofunction | MatrixBase | None = None) -> None:
```
pbrubeck: `output` -> `tensor` for consistency

firedrake/interpolation.py
```diff
@@ -25,395 +27,279 @@
...
         return ufl.Interpolate._ufl_expr_reconstruct_(self, expr, v=v, **interp_data)
 
+    @property
+    def options(self):
+        return self._options
+
 
 @PETSc.Log.EventDecorator()
-def interpolate(expr, V, subset=None, access=None, allow_missing_dofs=False, default_missing_val=None, matfree=True):
+def interpolate(expr: Expr, V: WithGeometry | ufl.BaseForm, **kwargs) -> Interpolate:
```
```suggestion
def interpolate(expr: Expr, V: ufl.FunctionSpace | ufl.BaseForm, **kwargs) -> Interpolate:
```

firedrake/interpolation.py
```diff
@@ -25,395 +27,289 @@
...
-                if not isinstance(dual_arg, ufl.Coargument):
-                    cofunctions = (dual_arg,)
-
-            if needs_adjoint and len(arguments) == 0:
-                Iu = self._interpolate(default_missing_val=default_missing_val)
-                return assemble(ufl.Action(*cofunctions, Iu), tensor=tensor)
-            else:
-                return self._interpolate(*cofunctions, output=tensor, adjoint=needs_adjoint,
-                                         default_missing_val=default_missing_val)
+            assert isinstance(tensor, Function | Cofunction | None)
```
pbrubeck, line 291: For the 0-form interpolate we should also allow `Constant` here, it also supports `assign`.

firedrake/interpolation.py
```diff
@@ -25,395 +27,303 @@
...
+    )
+    if target_mesh is source_mesh or submesh_interp_implemented:
+        return SameMeshInterpolator(expr)
+
+    target_topology = target_mesh.topology
+    source_topology = source_mesh.topology
+
+    if isinstance(target_topology, VertexOnlyMeshTopology):
+        if isinstance(source_topology, VertexOnlyMeshTopology):
+            return VomOntoVomInterpolator(expr)
```
pbrubeck, line 340: Does `VomOntoVomInterpolator` support `MixedFunctionSpace`, or are we never expecting to build one for this case? There should be a `ValueError` in the constructor if we go for the latter.

firedrake/interpolation.py
```diff
@@ -731,246 +536,199 @@ class SameMeshInterpolator(Interpolator):
...
+    access: Literal[op2.WRITE, op2.MIN, op2.MAX, op2.INC],
+    subset: op2.Subset | None = None,
+    bcs: Iterable[DirichletBC] | None = None
+) -> tuple[Callable, ...]:
+    """Returns tuple of callables which calculate the interpolation.
+
+    Parameters
+    ----------
+    expr : ufl.Interpolate | ufl.ZeroBaseForm
+        The symbolic interpolation expression, or a zero form. Zero forms
```
pbrubeck: "zero form" is too close to 0-form, which is not necesarily equal to 0.
```suggestion
        The symbolic interpolation expression, or a ZeroBaseForm. ZeroBaseForms
```

firedrake/interpolation.py
```diff
@@ -981,54 +739,43 @@ def _interpolator(tensor, expr, subset, access, bcs=None):
...
         # FInAT only elements
-        raise NotImplementedError("Don't know how to create FIAT element for %s" % V.ufl_element())
+        raise NotImplementedError(f"Don't know how to create FIAT element for {V.ufl_element()}")
 
     if access is op2.READ:
         raise ValueError("Can't have READ access for output function")
 
     # NOTE: The par_loop is always over the target mesh cells.
-    target_mesh = as_domain(V)
+    target_mesh = V.mesh()
```
pbrubeck, line 791: Now we might want to reconsider leaving `as_domain` here...

firedrake/interpolation.py
```diff
@@ -1418,119 +1171,55 @@ def __init__(self, glob):
...
-    expr : `ufl.Expr`
-        The expression to interpolate. If ``arguments`` is not empty, those
-        arguments must be present within it.
-    arguments : list of `ufl.Argument`
-        The arguments in the expression.
-    """
+        self.sf = self.original_vom.input_ordering_without_halos_sf
+        self.V = interpolator.target_space
+        self.source_vom = interpolator.source_mesh
+        self.expr = interpolator.operand
```
pbrubeck: The parent class has this as the the UFL `Interpolate`, but here we store the operand.

firedrake/interpolation.py
```diff
@@ -25,395 +27,303 @@
...
         dual_arg, operand = expr.argument_slots()
-        self.ufl_interpolate = expr
-        self.expr = operand
-        self.V = V
-        self.subset = subset
-        self.freeze_expr = freeze_expr
-        self.bcs = bcs
-        self._allow_missing_dofs = allow_missing_dofs
-        self.matfree = matfree
+        self.expr = expr
```
pbrubeck: Should we rename this to `self.ufl_interpolate` to be more explicit?

firedrake/interpolation.py
pbrubeck: Ah, this is not inheriting from that class. But maybe it'd be good to rename this field to `operand` to be consistent

firedrake/interpolation.py
```diff
@@ -1164,10 +915,12 @@ def _interpolator(tensor, expr, subset, access, bcs=None):
     if isinstance(tensor, op2.Mat):
         return parloop, tensor.assemble
     else:
+        if access == op2.INC:
+            copyin += (tensor.zero,)
```
pbrubeck: For the 1-form adjoint on a mixed space, we might be adding different sub-interpolates onto the same tensor. The zeroing should be handled outside the for loop of line 633

firedrake/interpolation.py
```diff
@@ -1391,118 +1191,66 @@ def __init__(self, glob):
         self.ufl_domain = lambda: None
 
 
-class VomOntoVomWrapper(object):
-    """Utility class for interpolating from one ``VertexOnlyMesh`` to it's
-    intput ordering ``VertexOnlyMesh``, or vice versa.
+class VomOntoVomMat:
+    """Object that facilitates interpolation between two vertex-only meshes."""
```
connorjward: This could be clearer.

firedrake/interpolation.py
```diff
@@ -1582,27 +1360,64 @@ def broadcast(self, source_vec, target_vec):
...
+        source_vec : PETSc.Vec
+            The vector to adjoint interpolate.
+        target_vec : PETSc.Vec
+            The vector to store the result in.
+        """
         self.multTranspose(mat, source_vec, target_vec)
 
-    def multTranspose(self, mat, source_vec, target_vec):
+    def multTranspose(self, mat: PETSc.Mat, source_vec: PETSc.Vec, target_vec: PETSc.Vec) -> None:
+        """Applies the tranpose of the interpolation operator. Called by `self.multHermitian`.
```
connorjward: Nitpick

```suggestion
        """Apply the tranpose of the interpolation operator. Called by `self.multHermitian`.
```

connorjward: And the same elsewhere. The first line of the docstring should be imperative.

firedrake/interpolation.py
```diff
@@ -1652,99 +1482,100 @@ def _wrap_dummy_mat(self):
...
+            def callable() -> Function | Cofunction:
+                for k, sub_tensor in enumerate(f.subfunctions):
+                    sub_tensor.assign(sum(
+                        interp.assemble(bcs=sub_bcs) for indices, (interp, sub_bcs) in Isub.items() if indices[0] == k
+                    ))
+                return f
+        else:
+            def callable() -> Number:
+                return sum(interp.assemble(bcs=sub_bcs) for (interp, sub_bcs) in Isub.values())
+        self.callable = callable
```
connorjward: This seems wrong. If you call `_build_callable` multiple times with different arguments this will get overwritten.

firedrake/interpolation.py
```diff
@@ -2,391 +2,338 @@
...
     "DofNotDefinedError",
-    "CrossMeshInterpolator",
-    "SameMeshInterpolator",
+    "InterpolateOptions",
+    "Interpolator"
 )
 
 
-class Interpolate(ufl.Interpolate):
+@dataclass
```
```suggestion
@dataclass(kw_only=True)
```
I think we should disallow positional arguments for this.

firedrake/interpolation.py
```diff
@@ -426,273 +373,166 @@ class CrossMeshInterpolator(Interpolator):
...
-        """Compute the interpolation.
+        # Get expression for point evaluation at the dest_node_coords
+        P0DG_vom = fs_type(vom, "DG", 0)
+        point_eval = interpolate(self.operand, P0DG_vom)
 
-        For arguments, see :class:`.Interpolator`.
-        """
-        from firedrake.assemble import assemble
+        # If assembling the operator, we need the concrete permutation matrix
+        matfree = False if self.rank == 2 else self.matfree
```
connorjward, line 468: This seems bizarre. A user could pass a `matfree` option that gets ignored.

firedrake/interpolation.py
```diff
@@ -1652,99 +1479,97 @@ def _wrap_dummy_mat(self):
...
             # Create the Jacobian to be split into blocks
-            expr = expr._ufl_expr_reconstruct_(operand, V)
+            self.ufl_interpolate = self.ufl_interpolate._ufl_expr_reconstruct_(self.operand, self.target_space)
 
-        Isub = {}
-        # Split in the arguments of the Interpolate
-        for indices, form in firedrake.formmanipulation.split_form(expr):
-            if isinstance(form, ufl.ZeroBaseForm):
+        # Get sub-interpolators and sub-bcs for each block
+        Isub: dict[tuple[int, int], tuple[Interpolator, list[DirichletBC]]] = {}
```
connorjward: This is a great idea. Big fan.

firedrake/interpolation.py
connorjward: This seems bizarre. A user could pass a `matfree` option that gets ignored.

connorjward (changes requested):
.

firedrake/interpolation.py
leo-collins: We can't assemble the cross-mesh interpolation matrix if `matfree=True`. By doing this the user can do `assemble(interpolate(TrialFunction(V), U))` and get the operator without having to pass an additional `matfree=False`.

firedrake/interpolation.py
leo-collins: `tensor` is already the indexed sub-tensor in this case. Can you write a test for this?

firedrake/interpolation.py
connorjward: In which case I think we should change (and document) the interface to
```py
def interpolate(expr, ..., *, matfree=None):
  if matfree is None:
    if some_condition:
      matfree = False
    else:
      matfree = True
```
because currently the user could run
```py
assemble(interpolate(TrialFunction(V), U, matfree=True))
```
and the option would be silently ignored.

firedrake/interpolation.py
pbrubeck: I'll do that

firedrake/interpolation.py
leo-collins: The same could be said about the `allow_missing_dofs` and `default_missing_val` parameters if the user is doing same-mesh interpolation. I think it's fine if the docstring says something like "this parameter only applies if interpolating between a VertexOnlyMesh and its input_ordering".

connorjward: Well the right answer there is to disallow those options for same-mesh interpolation. I don't want to force you to make big changes here though so I'd be happy provided that this behaviour is made clear in the docstring and you open an issue about it.

firedrake/interpolation.py
pbrubeck: I'm not a fan. Why can we not have the BCs passed on to the `Interpolator`?

firedrake/interpolation.py
```diff
@@ -1652,99 +1476,97 @@ def _wrap_dummy_mat(self):
...
-
-        if rank == 1:
-            for k, sub_tensor in enumerate(output.subfunctions):
-                sub_tensor.assign(sum(self[i].assemble(**kwargs) for i in self if i[0] == k))
-        elif rank == 2:
-            for k, sub_tensor in enumerate(output.subfunctions):
-                sub_tensor.assign(sum(self[i]._interpolate(*function, adjoint=adjoint, **kwargs)
-                                      for i in self if i[0] == k))
-        return output
+            Isub[indices] = (get_interpolator(form), sub_bcs)
```
pbrubeck, line 1566: Can we not do this instead?
```suggestion
            Isub[indices] = get_interpolator(form, bcs=sub_bcs)
```

firedrake/interpolation.py
```diff
@@ -2,391 +2,339 @@
...
+
+    operand, = expr.ufl_operands
+    target_mesh = expr.target_space.mesh()
+    source_mesh = extract_unique_domain(operand) or target_mesh
+    submesh_interp_implemented = (
+        all(isinstance(m.topology, MeshTopology) for m in [target_mesh, source_mesh])
+        and target_mesh.submesh_ancesters[-1] is source_mesh.submesh_ancesters[-1]
+        and target_mesh.topological_dimension == source_mesh.topological_dimension
+    )
+    if target_mesh is source_mesh or submesh_interp_implemented:
```
pbrubeck, line 335: This logic needs to be updated for `MeshSequence`. Now that `MixedFunctionSpace` returns a `MeshSequence`, we always use `MixedInterpolator` for cases that are supported by `SameMeshInterpolator` (for instance interpolation from MFS to a standard FS, and the adjoint of that). 

Once this is fixed, we will hit the issue with calling `zero()` inside the for loop in `_build_interpolation_callables`.

firedrake/interpolation.py
```diff
@@ -2,391 +2,339 @@
...
+        An appropriate :class:`Interpolator` subclass for the given
+        interpolation expression.
+    """
+    arguments = expr.arguments()
+    has_mixed_arguments = any(len(arg.function_space()) > 1 for arg in arguments)
+    if len(arguments) == 2 and has_mixed_arguments:
+        return MixedInterpolator(expr)
+
+    operand, = expr.ufl_operands
+    target_mesh = expr.target_space.mesh()
```
pbrubeck, line 315: Maybe we want something like this?
```suggestion
    target_mesh = expr.target_space.mesh()
    try:
        target_mesh = target_mesh.unique()
    except ValueError:
        pass
```

firedrake/interpolation.py
pbrubeck: https://github.com/firedrakeproject/firedrake/pull/4728

firedrake/interpolation.py
connorjward: I guess that's just a matter of taste. We don't generate code for the boundary conditions so stashing things on the `Interpolator` is less important.

I was specifically praising the use of type hints to make an extremely complex data structure clear.

firedrake/interpolation.py
```diff
@@ -2,391 +2,339 @@
...
+    source_mesh = extract_unique_domain(operand) or target_mesh
```
pbrubeck: We should also consider the possibility of the `operand` having a `MeshSequence`. Right now  `extract_unique_domain` will raise an Exception, but `MixedMeshInterpolator` should  be able to handle the case  where `operand` is an `Argument` of a `MixedFunctionSpace`. The case where `operand` is a more general `Expr` on multiple meshes might involve more complicated code generation. 

This should be done in a separate PR.

firedrake/interpolation.py
pbrubeck: We can't change the BCs on matrices on repeated assembly calls, or can we?

pbrubeck: Also the `tuple[int, int]` could in some cases just be `tuple[int]`

connorjward: > We can't change the BCs on matrices on repeated assembly calls, or can we?

Good point. I'm not sure. We switch the lgmaps out at runtime so feasibly it's OK to do.

firedrake/interpolation.py
pbrubeck: Or if that is not your intended design, I suggest we enable something like this
```suggestion
            Isub[indices] = get_assembler(form, bcs=sub_bcs)
```

firedrake/interpolation.py
pbrubeck: Are we sure this has been properly resolved?

pbrubeck: It'll be useful to have some assertions here

firedrake/interpolation.py
pbrubeck: A third option is to do something like
```suggestion
            Isub[indices] = partial(get_interpolator(form).assemble, bcs=sub_bcs)
```

firedrake/interpolation.py
leo-collins: My thought was that BCs should be passed to `assemble`, like we do for all other forms

firedrake/preconditioners/hiptmair.py
```diff
@@ -202,7 +202,7 @@ def coarsen(self, pc):
 
         coarse_space_bcs = tuple(coarse_space_bcs)
         if G_callback is None:
-            interp_petscmat = chop(Interpolator(dminus(trial), V, bcs=bcs + coarse_space_bcs).callable().handle)
+            interp_petscmat = chop(assemble(interpolate(dminus(trial), V), bcs=bcs + coarse_space_bcs).mat())
```
pbrubeck: Why not `assemble(...).petscmat`?

firedrake/interpolation.py
```diff
@@ -209,7 +209,7 @@ def __init__(self, expr: Interpolate):
         """The dual argument slot of the Interpolate expression."""
         self.target_space = dual_arg.function_space().dual()
         """The primal space we are interpolating into."""
-        self.target_mesh = self.target_space.mesh()
+        self.target_mesh = self.target_space.mesh().unique()
```
pbrubeck: I think `MixedInterpolator` will be fine with `MeshSequence`, so this call should go in the subclasses

firedrake/interpolation.py
```diff
@@ -953,75 +764,55 @@ def _interpolator(tensor, expr, subset, access, bcs=None):
...
-        raise ValueError("Expecting to interpolate a ufl.Interpolate")
-
-    arguments = expr.arguments()
+        expr = interpolate(zero(V.value_shape), V)
+    if not isinstance(expr, Interpolate):
+        raise ValueError("Expecting to interpolate a symbolic Interpolate expression.")
     dual_arg, operand = expr.argument_slots()
-    V = dual_arg.arguments()[0].function_space()
-
+    V = dual_arg.function_space().dual()
```
pbrubeck, line 779:
```suggestion
    assert isinstance(dual_arg, Cofunction | Coargument)
    V = dual_arg.function_space().dual()
```

PR #4596, opened 2025-09-28
MixedInterpolator

pbrubeck:
# Description
Enables assembly of the interpolation matrix for mixed spaces
```
assemble(interpolate(TrialFunction(V1*V2), W1*W2))) -> MatNest([[Interp(v1, w1*), 0], [0, Interp(v2, w2*)]])
```
We also allow the operand to be any expression, resulting in non-zero off diagonal blocks.

This is the last step required to clean up the special cases of an `Interpolate` in `assemble.py`.


firedrake/interpolation.py
```diff
@@ -1720,3 +1704,90 @@ def _wrap_dummy_mat(self):
...
+        operands = []
+        components = numpy.reshape(operand, (-1,))
+        for Vi in V:
+            operands.append(ufl.as_tensor(components[cur:cur+Vi.value_size].reshape(Vi.value_shape)))
+            cur += Vi.value_size
+    expressions = tuple(map(expr._ufl_expr_reconstruct_, operands, duals))
+    return expressions
+
+
+class MixedInterpolator(Interpolator):
```
connorjward, line 1688: I don't think that we should be creating a new class here. Why not rename the existing `assemble` method to `_assemble_subspace` and then you can reuse the existing code without inheritance.

pbrubeck: The separate class approach does 3 things:

1. Does not construct a SameMeshInterpolator/CrossMeshInterpolator on unimplemented cases.
2. Calls the constructor for each `sub_interpolator`.
3. Calls the assembly of each `sub_interpolator`.

I can see how 2. could be elegantly dealt with by making `sub_interpolators` a cached_property, and adding an if-statement within assemble() should take care of 3. But I am not sure about 1, but it seems we would need the same if-statement in the constructor of the other subclasses.

pbrubeck: I like this approach because it only needs one if-statement when we dispatch the subclasses, and separates the MixedFunctionSpace operations from the subclasses.

connorjward: I think we should always be extremely wary of adding new classes. It very often makes composition fragile and also makes it difficult to extend the code later. I'm thinking of Cofunction, RestrictedFunctionSpace, RealFunctionSpace etc as examples.

If we can ever avoid adding another one I think we should. Here you are adding an entire class to avoid a couple of checks inside methods which to me indicates that inheritance is the wrong pattern.

pbrubeck: What concrete subclass should we give to `Interpolator(Mixed, Mixed)`?  What if the spaces being mixed are from the source mesh(es) into a subset of those meshes and different meshes?

pbrubeck: I think the more fragile classes are `SameMeshInterpolator` and `CrossMeshInterpolator`, as they are quite complicated. Since the code being added is self-contained, this new class should in principle be no more fragile than the `sub_interpolators`.

I think the situation here is very similar to a `MixedFunctionSpace`, which was not listed in your examples.

connorjward: Ah yeah... I think it was probably a mistake to have two interpolation classes for same-mesh and cross-mesh... You see what I mean!

That's not your problem to fix here though.

pbrubeck: @leo-collins is adding a third one, `VomOntoVomInterpolator`

connorjward: MixedFunctionSpace is also a bad design but it can't go until pyop3.

connorjward: I'm not sure anything can be done here, but all this code needs a big cleanup. This proliferation of classes makes things hard to maintain and extend in future.

dham (approved):

pbrubeck:
This needed more work after https://github.com/FEniCS/ufl/pull/422 was merged. We now splitt the sub interpoaltors via the same mechanism that handles fieldsplit, so this should be much cleaner.

firedrake/interpolation.py
```diff
@@ -499,62 +506,24 @@ def __init__(
...
-                # Build and save an interpolator for each sub-element
-                # separately for MixedFunctionSpaces. NOTE: since we can't have
-                # expressions for MixedFunctionSpaces we know that the input
-                # argument ``expr`` must be a Function. V_dest can be a Function
-                # or a FunctionSpace, and subfunctions works for both.
-                if self.nargs == 1:
-                    # Arguments don't have a subfunctions property so I have to
-                    # make them myself. NOTE: this will not be correct when we
-                    # start allowing interpolators created from an expression
-                    # with arguments, as opposed to just being the argument.
```
pbrubeck, line 533: The internal sub_interpolators here had limited capability (e.g. could not go from a non-mixed `VectorFunctionSpace` into a `MixedFunctionSpace`, see test), this is superseded by the `MixedInterpolator`.

firedrake/interpolation.py
```diff
@@ -499,62 +506,24 @@ def __init__(
...
-                if len(expr_subfunctions) != len(V_dest.subspaces):
-                    raise NotImplementedError(
-                        "Can't interpolate from a non-mixed function space into a mixed function space."
-                    )
-                for input_sub_func, target_subspace in zip(
-                    expr_subfunctions, V_dest.subspaces
-                ):
-                    self.sub_interpolators.append(
-                        interpolate(
-                            input_sub_func, target_subspace, subset=subset,
```
pbrubeck, line 553: Also this did not know about the adjoint (`target_subspace` was only allowed to be a `FunctionSpace`)

firedrake/interpolation.py
```diff
@@ -499,62 +506,24 @@ def __init__(
...
-                            access=access, allow_missing_dofs=allow_missing_dofs
-                        )
-                    )
-                return
+            if type(ufl_scalar_element) == finat.ufl.MixedElement:
```
leo-collins:
```suggestion
            if isinstance(ufl_scalar_element, finat.ufl.MixedElement):
```

pbrubeck: Since `VectorFiniteElement` and `TensorFiniteElement` inherit from MixedElement, this is the official way of determining that we are really dealing with a `MixedFiniteElement`

leo-collins: Should be an `is` instead of `==` then

pbrubeck: You are right, thanks. I was doubting here, and I thought flake would complain if I had it wrong.
```suggestion
            if type(ufl_scalar_element) is finat.ufl.MixedElement:
```

firedrake/interpolation.py
```diff
@@ -975,37 +929,37 @@ def callable():
...
+            dual_split = dict(firedrake.formmanipulation.split_form(dual_arg))
+            # Combine the splits by taking their action
+            expressions = {i: action(interp_split[i], dual_split[i[-1:]]) for i in interp_split}
 
         # Interpolate each sub expression into each function space
-        for Vsub, sub_tensor, sub_expr in zip(V, tensor, expressions):
-            loops.extend(_interpolator(Vsub, sub_tensor, sub_expr, subset, arguments, access, bcs=bcs))
+        for indices, sub_expr in expressions.items():
+            if isinstance(sub_expr, ufl.ZeroBaseForm):
+                continue
```
```suggestion
            sub_tensor = tensor[indices[0]] if rank == 1 else tensor
            if isinstance(sub_expr, ufl.ZeroBaseForm):
                if access == op2.WRITE:
                    sub_tensor.zero()
                continue
```

firedrake/interpolation.py
```diff
@@ -975,37 +929,37 @@ def callable():
...
+            arguments = sub_expr.arguments()
+            sub_space = sub_expr.argument_slots()[0].function_space().dual()
+            sub_tensor = tensor[indices[0]] if rank == 1 else tensor
```
pbrubeck, line 958:
```suggestion
```

firedrake/interpolation.py
```diff
@@ -975,37 +929,37 @@ def callable():
...
-            else:
-                duals = [v for _, v in sorted(firedrake.formmanipulation.split_form(dual_arg))]
-            expressions = map(expr._ufl_expr_reconstruct_, operands, duals)
+        if access == op2.INC:
+            loops.append(tensor.zero)
+
+        dual_arg, operand = expr.argument_slots()
+        # Arguments in the operand are allowed to be from a MixedFunctionSpace
+        # We need to split the target space V and generate separate kernels
+        if len(arguments) == 2:
```
pbrubeck, line 939:
```suggestion
        if len(arguments) == 2:
            # Matrix case assumes that the spaces are not mixed
```

firedrake/interpolation.py
```diff
@@ -975,37 +929,40 @@ def callable():
...
+            expressions = {i: action(interp_split[i], dual_split[i[-1:]]) for i in interp_split}
 
         # Interpolate each sub expression into each function space
-        for Vsub, sub_tensor, sub_expr in zip(V, tensor, expressions):
-            loops.extend(_interpolator(Vsub, sub_tensor, sub_expr, subset, arguments, access, bcs=bcs))
+        for indices, sub_expr in expressions.items():
+            sub_tensor = tensor[indices[0]] if rank == 1 else tensor
+            if isinstance(sub_expr, ufl.ZeroBaseForm):
+                if access == op2.WRITE:
+                    sub_tensor.zero()
```
```suggestion
                    loops.append(sub_tensor.zero)
```

firedrake/formmanipulation.py
```diff
@@ -171,6 +183,41 @@ def matrix(self, o):
...
+        V = dual_arg.function_space()
+        if len(V) == 1 or len(dual_arg.arguments()) == 1:
+            return o._ufl_expr_reconstruct_(operand, dual_arg)
+
+        # Split the target (dual) argument
+        if isinstance(dual_arg, Coargument):
+            dual_arg = self(dual_arg)
+            indices = self.blocks[dual_arg.number()]
+        else:
+            raise NotImplementedError()
```
connorjward: Please give a helpful error message here.

firedrake/interpolation.py
```diff
@@ -369,7 +379,7 @@ def _interpolate(self, *args, **kwargs):
         """
         pass
 
-    def assemble(self, tensor=None, default_missing_val=None):
+    def assemble(self, tensor=None, **kwargs):
```
connorjward: Please give this the full docstring treatment. This is a public method and it is not clear what is happening to kwargs.

firedrake/interpolation.py
```diff
@@ -499,62 +506,24 @@ def __init__(
...
-                ):
-                    self.sub_interpolators.append(
-                        interpolate(
-                            input_sub_func, target_subspace, subset=subset,
-                            access=access, allow_missing_dofs=allow_missing_dofs
-                        )
-                    )
-                return
+            if type(ufl_scalar_element) is finat.ufl.MixedElement:
+                raise NotImplementedError("Interpolation matrix with MixedFunctionSpace requires MixedInterpolator")
```
connorjward: Should this be a `TypeError`? We aren't planning on implementing things here.

firedrake/interpolation.py
```diff
@@ -911,10 +865,10 @@ def make_interpolator(expr, V, subset, access, bcs=None, matfree=True):
         Vrow = arguments[0].function_space()
         Vcol = arguments[1].function_space()
         if len(Vrow) > 1 or len(Vcol) > 1:
-            raise NotImplementedError("Interpolation of mixed expressions with arguments is not supported")
+            raise NotImplementedError("Interpolation matrix with MixedFunctionSpace requires MixedInterpolator")
```
connorjward: `TypeError`

firedrake/formmanipulation.py
```diff
@@ -171,6 +183,41 @@ def matrix(self, o):
...
 
+    def zero_base_form(self, o):
+        return ZeroBaseForm(tuple(map(self, o.arguments())))
+
+    def interpolate(self, o, operand):
+        if isinstance(operand, Zero):
+            return self(ZeroBaseForm(o.arguments()))
+
+        dual_arg, _ = o.argument_slots()
+        V = dual_arg.function_space()
```
pbrubeck:
```suggestion
        V = dual_arg.arguments()[0].function_space()
```

firedrake/interpolation.py
```diff
@@ -1019,10 +969,21 @@ def callable(loops, f):
...
 @utils.known_pyop2_safe
-def _interpolator(V, tensor, expr, subset, arguments, access, bcs=None):
+def _interpolator(tensor, expr, subset, access, bcs=None):
+    if isinstance(expr, ufl.ZeroBaseForm):
+        if access is op2.INC:
+            return ()
+        elif access is op2.WRITE:
+            return (partial(tensor.zero, subset=subset),)
+        V = expr.arguments()[-1].function_space().dual()
+        expr = interpolate(ufl.zero(V.value_shape), V)
```
connorjward, line 983: I don't understand this block. Also isn't `access` always `INC` or `WRITE`?

firedrake/interpolation.py
```diff
@@ -1019,10 +969,21 @@ def callable(loops, f):
...
+
     if not isinstance(expr, ufl.Interpolate):
         raise ValueError("Expecting to interpolate a ufl.Interpolate")
+
+    arguments = expr.arguments()
     dual_arg, operand = expr.argument_slots()
+    V = dual_arg.function_space().dual()
```
```suggestion
    V = dual_arg.arguments()[0].function_space()
```

firedrake/interpolation.py
```diff
@@ -1720,3 +1679,93 @@ def _wrap_dummy_mat(self):
...
+    def __getitem__(self, item):
+        return self._sub_interpolators[item]
+
+    def __iter__(self):
+        return iter(self._sub_interpolators)
+
+    def _get_callable(self):
+        """Assemble the operator."""
+        shape = tuple(len(a.function_space()) for a in self.arguments)
+        blocks = numpy.full(shape, PETSc.Mat(), dtype=object)
```
connorjward, line 1751:
```suggestion
        blocks = numpy.empty(shape, dtype=object)
```
?
It seems like we throw these matrices away.

firedrake/interpolation.py
```diff
@@ -1720,3 +1679,93 @@ def _wrap_dummy_mat(self):
...
+        self.callable = self._get_callable
+
+    def __getitem__(self, item):
+        return self._sub_interpolators[item]
+
+    def __iter__(self):
+        return iter(self._sub_interpolators)
+
+    def _get_callable(self):
+        """Assemble the operator."""
```
connorjward, line 1749:
```suggestion
        """Return a callable that assembles the operator."""
```
is that right? Otherwise it's not a very good method name.

firedrake/interpolation.py
pbrubeck: No, we need them to be null pointers. The indices in self are not spanning every block

firedrake/interpolation.py
pbrubeck: It is a terrible name, it returns the assembled thing. The attribute `self.callable` will then call this method.

firedrake/interpolation.py
pbrubeck: There's also `MIN` and `MAX`.

pbrubeck: The reason for `ZeroBaseForm` is that we are not allowed to have split zero off-diagonal blocks into things like `Inteprolate(zero(), Argument(1))`, because we do not allow a missing `Argument(0)` (`zero()*x` gets simplified to `zero()`). So I had to make the form splitter and the sub-assemblers aware of zero-simplification. This is good because we are no longer generating code for things like `Function(V).inteprolate(zero())`.

pbrubeck: We should rarely see this code running with MAX/MIN, but I made sure it does the right thing

firedrake/interpolation.py
```diff
@@ -1720,3 +1679,93 @@ def _wrap_dummy_mat(self):
...
+        shape = tuple(len(a.function_space()) for a in self.arguments)
+        blocks = numpy.full(shape, PETSc.Mat(), dtype=object)
+        for i in self:
```
pbrubeck, line 1753:
```suggestion
        # Assemble the sparse block matrix
        for i in self:
```

firedrake/interpolation.py
```diff
@@ -1720,3 +1679,93 @@ def _wrap_dummy_mat(self):
...
+        self._sub_interpolators = Isub
+        self.callable = self._get_callable
+
+    def __getitem__(self, item):
+        return self._sub_interpolators[item]
+
+    def __iter__(self):
+        return iter(self._sub_interpolators)
+
+    def _get_callable(self):
```
```suggestion
        self.callable = self._assemble_matnest

    def __getitem__(self, item):
        return self._sub_interpolators[item]

    def __iter__(self):
        return iter(self._sub_interpolators)

    def _assemble_matnest(self):
```

firedrake/interpolation.py
pbrubeck:
```suggestion
            raise TypeError("Interpolation matrix with MixedFunctionSpace requires MixedInterpolator")
```

firedrake/interpolation.py
```suggestion
                raise TypeError("Interpolation matrix with MixedFunctionSpace requires MixedInterpolator")
```

firedrake/interpolation.py
pbrubeck: I'll put back the `default_missing_val`. This is going to disapper in https://github.com/firedrakeproject/firedrake/pull/4595

firedrake/interpolation.py
```diff
@@ -401,11 +409,10 @@ def assemble(self, tensor=None, default_missing_val=None):
                     cofunctions = (dual_arg,)
 
             if needs_adjoint and len(arguments) == 0:
-                Iu = self._interpolate(default_missing_val=default_missing_val)
+                Iu = self._interpolate(**kwargs)
```
```suggestion
                Iu = self._interpolate(default_missing_val=default_missing_val)
```

firedrake/interpolation.py
```diff
@@ -401,11 +409,10 @@ def assemble(self, tensor=None, default_missing_val=None):
...
                 return assemble(ufl.Action(*cofunctions, Iu), tensor=tensor)
             else:
-                return self._interpolate(*cofunctions, output=tensor, adjoint=needs_adjoint,
-                                         default_missing_val=default_missing_val)
+                return self._interpolate(*cofunctions, output=tensor, adjoint=needs_adjoint, **kwargs)
```
```suggestion
                return self._interpolate(*cofunctions, output=tensor, adjoint=needs_adjoint,
                default_missing_val=default_missing_val)
```

firedrake/interpolation.py
```suggestion
    def assemble(self, tensor=None, default_missing_val=None):
```

firedrake/formmanipulation.py
```diff
@@ -171,6 +183,41 @@ def matrix(self, o):
...
+        if len(V) == 1 or len(dual_arg.arguments()) == 1:
```
pbrubeck:
```suggestion
        if len(dual_arg.arguments()) == 1 or len(dual_arg.arguments()[-1].function_space()) == 1:
        # The dual argument has been contracted or does not need to be split
```

firedrake/formmanipulation.py
pbrubeck:
```suggestion
            raise NotImplementedError(f"I do not know how to split an Interpolate with a {type(dual_arg).__name__}.")
```

firedrake/interpolation.py
```diff
@@ -405,7 +413,7 @@ def assemble(self, tensor=None, default_missing_val=None):
                 return assemble(ufl.Action(*cofunctions, Iu), tensor=tensor)
             else:
                 return self._interpolate(*cofunctions, output=tensor, adjoint=needs_adjoint,
-                                         default_missing_val=default_missing_val)
+                default_missing_val=default_missing_val)
```
```suggestion
                                         default_missing_val=default_missing_val)
```

firedrake/interpolation.py
connorjward: OK fine

firedrake/interpolation.py
connorjward: Can this be explained in a comment. This is not at all obvious by reading the code.

firedrake/interpolation.py
connorjward: This doesn't seem to have been fixed.

connorjward: Actually I think it was just GitHub GUI messing with me. It's fine.

pbrubeck: It's out of date. I renamed it to `_assemble_matnest`

firedrake/interpolation.py
pbrubeck: We could do something like `numpy.max(x, 0, out=x)`

pbrubeck: But it'd be hard to get it for subset != None

connorjward: A comment explaining this whole block is fine. My issue is purely to do with readability.

firedrake/interpolation.py
```diff
@@ -1019,10 +970,21 @@ def callable(loops, f):
 
 
 @utils.known_pyop2_safe
-def _interpolator(V, tensor, expr, subset, arguments, access, bcs=None):
+def _interpolator(tensor, expr, subset, access, bcs=None):
+    if isinstance(expr, ufl.ZeroBaseForm):
+        if access is op2.INC:
```
pbrubeck, line 976:
```suggestion
        # Zero simplification, avoid code-generation
        if access is op2.INC:
```

firedrake/interpolation.py
```diff
@@ -1019,10 +970,21 @@ def callable(loops, f):
...
+            return ()
+        elif access is op2.WRITE:
+            return (partial(tensor.zero, subset=subset),)
+        V = expr.arguments()[-1].function_space().dual()
```
pbrubeck, line 982:
```suggestion
            return (partial(tensor.zero, subset=subset),)
        # Unclear how to avoid codegen for MIN and MAX
        # Reconstruct the expression as an Interpolate
        V = expr.arguments()[-1].function_space().dual()
```

connorjward (approved):

connorjward (approved):

Issue #4059, opened 2025-02-21
BUG: Docs build fails with sphinx 8.2.0


connorjward:
Since the release of Sphinx 8.2.0 (and the subsequent release of our docdeps container) docs have been failing to build with the error message:

```
Traceback
=========

      File "/home/firedrake/firedrake/lib/python3.12/site-packages/sphinxcontrib/inkscapeconverter.py", line 71, in convert
        args += ['--export-filename=' + _to, _from]
                 ~~~~~~~~~~~~~~~~~~~~~^~~~~
    TypeError: can only concatenate str (not "_StrPath") to str
```

We should pin to `<8.2.0` for now and report this issue upstream.

PR #4599, opened 2025-09-29
Remove Sphinx version pin

connorjward:
Closes https://github.com/firedrakeproject/firedrake/issues/4059

Things appear to have been fixed upstream. CI is passing with the latest version.


connorjward:
Merged as the only failure was the known issue with cross mesh interpolation.

PR #4600, opened 2025-09-29
Fix TransferManager reuse for mixed coefficients

pbrubeck:
# Description


pbrubeck:
I believe we will still fail for problems with non-homogenous DirichletBCs that are defined with a `Function` as the data.

pbrubeck:
We do a thorough extraction of coefficients when we coarsen a problem. They eventually get attached to the _SNESContext._coefficient_mapping, but only after setup. Maybe this PR should properly collect coefficients before calling `snes.solve()`

pbrubeck:
> > I believe we will still fail for problems with non-homogenous DirichletBCs that are defined with a `Function` as the data.
> 
> We're fine here, I don't know why. But I found another unrelated bug when composing MG as a SNES with MG as a PC. I think we should merget this as is.

Ah, I think this is fine because we always interpolate the data into our own FunctionSpace, so any coefficients coming from DirichletBCs never get transfered, but their interpolants in the FunctionSpace of the problem.

connorjward (approved):

PR #4602, opened 2025-09-29
PMGPC: use native matrix-free (adjoint) interpolation

pbrubeck:
# Description


firedrake/preconditioners/pmg.py
```diff
@@ -1220,50 +1220,50 @@ def work_function(self, V):
...
 
     @cached_property
     def _kernels(self):
+        from firedrake.interpolation import interpolate, Interpolator
         try:
-            # We generate custom prolongation and restriction kernels mainly because:
-            # 1. Code generation for the transpose of prolongation is not readily available
-            # 2. Dual evaluation of EnrichedElement is not yet implemented in FInAT
+            P = Interpolator(interpolate(self.uc, self.Vf), self.Vf)
+            P._interpolate
```
connorjward: Please explain this line.

firedrake/preconditioners/pmg.py
```diff
@@ -1220,50 +1220,50 @@ def work_function(self, V):
...
+            prolong = partial(P.assemble, tensor=self.uf)
+
+            rf = firedrake.Function(self.Vf.dual(), val=self.uf.dat)
+            rc = firedrake.Function(self.Vc.dual(), val=self.uc.dat)
+            vc = firedrake.TestFunction(self.Vc)
+            R = Interpolator(interpolate(vc, rf), self.Vf)
+            restrict = partial(R.assemble, tensor=rc)
+        except (AssertionError, AttributeError, NotImplementedError):
```
connorjward: In general it's bad practice to have big `try` `except` blocks. PEP8 says:
> Additionally, for all try/except clauses, limit the try clause to the absolute minimum amount of code necessary. Again, this avoids masking bugs:

Instead could we just have the check (pseudocode)
```py
if element.dual_evaluation is None:
  # dual eval not implemented
  ...
```

firedrake/preconditioners/pmg.py
```diff
@@ -1566,61 +1523,15 @@ def getNestSubMatrix(self, i, j):
...
-    return mat.handle
+def prolongation_matrix_aij(Vc, Vf, Vc_bcs=(), Vf_bcs=()):
+    if isinstance(Vf, firedrake.Function):
+        Vf = Vf.function_space()
+    if isinstance(Vc, firedrake.Function):
+        Vc = Vc.function_space()
+    bcs = Vc_bcs + Vf_bcs
+    interp = firedrake.interpolate(firedrake.TrialFunction(Vc), Vf)
+    mat = firedrake.assemble(interp, bcs=bcs)
+    return mat.petscmat
```
connorjward, line 1547: This is WAY better.

tests/firedrake/multigrid/test_p_multigrid.py
connorjward, line 1: Why all the solver options changes?

connorjward (changes requested):

tests/firedrake/multigrid/test_p_multigrid.py
pbrubeck: I deleted some inner KSP monitors that were just making everything harder to debug. I only replaced options to other equivalent yet simpler options (1 iteration of KSP Richarsdon is equivalent to `preonly`).

firedrake/preconditioners/pmg.py
pbrubeck: This is the main reason I introduced the `MixedInterpolator`

firedrake/preconditioners/pmg.py
pbrubeck: I think it is much better to have `P._interpolate` failing than checking if the target and source spaces have a `dual_basis` implemented.  Ideally we only require the target space to have a `dual_basis`. The source element may not have one. But we are also failing for other reasons that merely coincide with the source space not having a `dual_basis`.

pbrubeck: `DirichletBC` calls `P._interpolate` in order to decide whether to interpolae or project.

connorjward: If so we should define a `DualBasisUndefinedException` that we catch, instead of the generic ones.

pbrubeck: I think `P._interpolate` would fail for other reasons as well

connorjward: Maybe this needs a wider discussion, but I don't think we should be doing this. If it fails for other reasons then we should know what those are and handle them appropriately.

pbrubeck: I would like to come back to these other reasons after implementing `EnrichedElement.dual_basis`

pbrubeck: I'm happy to leave this as I have it now if that also makes you happy

PR #4609, opened 2025-10-01
Redo release workflow

connorjward:
Summary of changes:
* Instead of making a release via the GUI triggering a PyPI publishing workflow the new approach is to trigger a workflow manually. That way we can do some safety checks before publishing.
* We now make releases from branches *off* `release`. This means that we can test the FIAT `release` branch against our own and also means that the version number in the `pyproject.toml` can have a `.dev0` suffix. This is useful because it means Python can distinguish between an install of the `release` branch and an actual release.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,24 +1,73 @@
 name: Publish release
 
 on:
-  release:
-    types: [released]
+  workflow_call:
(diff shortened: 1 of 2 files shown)


connorjward:
Merging as all the relevant parts of CI are passing and this is holding up release.

PR #4619, opened 2025-10-03
Update supported PETSc version to v3.24.0

sam-pearmain:
The `firedrake-configure` script on `main` currently specifies `SUPPORTED_PETSC_VERSION = "v3.23.4"` yet latest release requires `petsc4py==3.24.0`. Updated version string to `v3.24.0` in accordance.


connorjward:
Thanks for submitting this but this is not quite the right fix. The real issue is the website should say to download `firedrake-configure` from `release` not `main`. I've fixed that and triggered a rebuild of the website. This fix should come over automatically when we merge `release` into `main` which we do regularly.

Thanks for bringing this to our attention!

PR #4623, opened 2025-10-05
BUG: fix reuse of matfree adjoint Interpolator

pbrubeck:
# Description

Fixes a parallel bug when reusing a matfree adjoint `Interpolator`. Summation with `access=INC` skipped the halo exchange after the first call to `Interpolator.assemble()`. On subsequent calls the summation only happened locally within each process, but not across processes.

This bug did not affect `assemble(interpolate(...))` because the Interpolator there is never reused.

### DIFF
--- a/tests/firedrake/regression/test_interpolate.py
+++ b/tests/firedrake/regression/test_interpolate.py
@@ -327,7 +327,7 @@ def test_trace():
     assert np.allclose(x_tr_cg.dat.data, x_tr_dir.dat.data)
 
 
-@pytest.mark.parallel(nprocs=[1, 3])
+@pytest.mark.parallel([1, 3])
 @pytest.mark.parametrize("rank", (0, 1))
 @pytest.mark.parametrize("mat_type", ("matfree", "aij"))
 @pytest.mark.parametrize("degree", (1, 3))
@@ -569,7 +569,7 @@ def test_mixed_matrix(mode):
         assert np.allclose(x.dat.data, y.dat.data)
 
 
-@pytest.mark.parallel(nprocs=2)
+@pytest.mark.parallel(2)
 @pytest.mark.parametrize("mode", ["forward", "adjoint"])
 @pytest.mark.parametrize("family,degree", [("CG", 1), ("DG", 0)])
 def test_interpolator_reuse(family, degree, mode):
@@ -595,9 +595,7 @@ def test_interpolator_reuse(family, degree, mode):
         assert result is tensor
 
         # Test that the input was not modified
-        x = u.dat.data
-        assert np.allclose(x, expected)
+        assert np.allclose(u.dat.data, expected)
 
         # Test for correctness
-        y = result.dat.data
-        assert np.allclose(y, expected)
+        assert np.allclose(result.dat.data, expected)


firedrake/interpolation.py
```diff
@@ -961,12 +958,23 @@ def callable():
...
-                l()
+        def callable(loops, f, access):
+            if access is op2.WRITE:
+                for l in loops:
+                    l()
+                return f
+            # We are repeatedly incrementing into the same Dat so intermediate halo exchanges
+            # can be skipped.
+            f.dat.local_to_global_begin(access)
+            with f.dat.frozen_halo(access):
```
pbrubeck: @connorjward please see the bug in the description. I grabbed this line from https://github.com/firedrakeproject/firedrake/blob/ec5492847ace646af6f83982a6e540ce32696a4a/firedrake/assemble.py#L1233

However that was not enough. To fix the reuse in parallel I also needed to wrap this block around `f.dat.local_to_global_[begin/end]`, however this breaks on the third call. Am I missing something?

firedrake/interpolation.py
```diff
@@ -1047,16 +1055,13 @@ def _interpolator(tensor, expr, subset, access, bcs=None):
...
-        kernel = op2.Kernel(kernel_code, "multiplicity", requires_zeroed_output_arguments=False)
-        weight = firedrake.Function(W)
         m_ = get_interp_node_map(source_mesh, target_mesh, W)
-        op2.par_loop(kernel, cell_set, weight.dat(op2.INC, m_))
+        m_indices = W.dof_dset.lgmap.apply(m_.values)
+        m_shape = m_indices.shape + W.shape
+        weight = firedrake.Function(W)
         with weight.dat.vec as w:
+            w.setValuesBlocked(m_indices, numpy.ones(m_shape), PETSc.InsertMode.ADD)
+            w.assemble()
```
pbrubeck: This is the parloop removal

firedrake/interpolation.py
```diff
@@ -1102,7 +1107,7 @@ def _interpolator(tensor, expr, subset, access, bcs=None):
     if isinstance(tensor, op2.Global):
         parloop_args.append(tensor(access))
     elif isinstance(tensor, op2.Dat):
-        V_dest = arguments[-1].function_space() if isinstance(dual_arg, ufl.Cofunction) else V
+        V_dest = arguments[0].function_space()
```
pbrubeck: Minor cleanup here

firedrake/interpolation.py
```diff
@@ -1079,7 +1084,7 @@ def _interpolator(tensor, expr, subset, access, bcs=None):
     coefficient_numbers = kernel.coefficient_numbers
     needs_external_coords = kernel.needs_external_coords
     name = kernel.name
-    kernel = op2.Kernel(ast, name, requires_zeroed_output_arguments=True,
+    kernel = op2.Kernel(ast, name, requires_zeroed_output_arguments=(access is op2.WRITE),
```
pbrubeck: We don't do this in assemble.py, but it seems that we still want it for op2.WRITE

firedrake/interpolation.py
```diff
@@ -1162,11 +1167,10 @@ def _interpolator(tensor, expr, subset, access, bcs=None):
                 parloop_args.append(target_ref_coords.dat(op2.READ, m_))
 
     parloop = op2.ParLoop(*parloop_args)
-    parloop_compute_callable = parloop.compute
     if isinstance(tensor, op2.Mat):
-        return parloop_compute_callable, tensor.assemble
+        return parloop, tensor.assemble
     else:
-        return copyin + callables + (parloop_compute_callable, ) + copyout
+        return copyin + callables + (parloop, ) + copyout
```
pbrubeck: Minor cleanup

firedrake/interpolation.py
connorjward: This doesn't quite match the expected usage. Frozen halos were only really a thing for repeated parloops. I don't know what will happen if you zero things in the middle of this. It might set halo validity in weird ways.

pbrubeck: I am happy to try and change the order of the call to zero(). I can't find where that is done in assemble.py

pbrubeck: Adjoint interpolation with MixedFunctionSpace is supposed to generate the kernels for each pair of source/target subspaces and add them into the corresponding block of rows in `tensor.subfunctions`

pbrubeck: The fact that we're on a MixedFunctionSpace is not causing the issue with the reuse

firedrake/interpolation.py
```diff
@@ -961,12 +958,23 @@ def callable():
         if bcs and rank == 1:
             loops.extend(partial(bc.apply, f) for bc in bcs)
 
-        def callable(loops, f):
-            for l in loops:
-                l()
+        def callable(loops, f, access):
```
pbrubeck: The way OneFormAssembler implements reuse allows for a different output tensor on each call, maybe that's what's missing here?

firedrake/interpolation.py
connorjward: > Also note that the halo freezing stuff is done automatically in pyop3, so maybe you don't have to worry about optimising that here.

firedrake/interpolation.py
```diff
@@ -1047,21 +1060,20 @@ def _interpolator(tensor, expr, subset, access, bcs=None):
...
+            w.assemble()
             w.reciprocal()
 
         # Create a buffer for the weighted Cofunction and a callable to apply the weight
         v = firedrake.Function(W)
         expr = expr._ufl_expr_reconstruct_(operand, v=v)
+        # Force a halo exchange on the weighted Cofunction
+        callables += (partial(getattr, v.dat, "data_with_halos"),)
         with weight.dat.vec_ro as w, dual_arg.dat.vec_ro as x, v.dat.vec_wo as y:
             callables += (partial(y.pointwiseMult, x, w),)
```
pbrubeck: @connorjward Seems like we need to force a halo exchange on the (input) buffer for the weighted cofunction (even we haven't written anything on it by this point)

connorjward: That's strange. No idea why that would be the case.

firedrake/interpolation.py
```diff
@@ -1038,32 +1039,33 @@ def _interpolator(tensor, expr, subset, access, bcs=None):
...
+        with wdat.vec as w:
+            w.setValuesBlocked(m_indices, numpy.ones(m_shape), PETSc.InsertMode.ADD)
+            w.assemble()
             w.reciprocal()
 
-        # Create a buffer for the weighted Cofunction and a callable to apply the weight
-        v = firedrake.Function(W)
-        expr = expr._ufl_expr_reconstruct_(operand, v=v)
-        with weight.dat.vec_ro as w, dual_arg.dat.vec_ro as x, v.dat.vec_wo as y:
-            callables += (partial(y.pointwiseMult, x, w),)
```
pbrubeck, line 1066: @connorjward it turns out that this pointwiseMult is the source of the evil. I fixed it by splitting it in two separate calls, so that multiplication is done in-place. 
```
x.copy(y)
y.pointwiseMult(y, w)
```
I have no explanation why this approach works and the original doesn't.

tests/firedrake/regression/test_interpolate.py
```diff
@@ -566,3 +567,37 @@ def test_mixed_matrix(mode):
     result_explicit = assemble(action(a, u))
     for x, y in zip(result_explicit.subfunctions, result_matfree.subfunctions):
         assert np.allclose(x.dat.data, y.dat.data)
+
+
+@pytest.mark.parallel(nprocs=2)
```
connorjward:
```suggestion
@pytest.mark.parallel(2)
```

tests/firedrake/regression/test_interpolate.py
```diff
@@ -327,6 +327,7 @@ def test_trace():
     assert np.allclose(x_tr_cg.dat.data, x_tr_dir.dat.data)
 
 
+@pytest.mark.parallel(nprocs=[1, 3])
```
connorjward:
```suggestion
@pytest.mark.parallel([1, 3])
```

tests/firedrake/regression/test_interpolate.py
```diff
@@ -566,3 +567,37 @@ def test_mixed_matrix(mode):
...
+        u.assign(rg.uniform(u.function_space()))
+        expected = u.dat.data.copy()
+
+        tensor = Function(expr.function_space())
+        result = I.assemble(tensor=tensor)
+        assert result is tensor
+
+        # Test that the input was not modified
+        x = u.dat.data
+        assert np.allclose(x, expected)
```
connorjward:
```suggestion
        assert np.allclose(u.dat.data, expected)
```

tests/firedrake/regression/test_interpolate.py
```diff
@@ -566,3 +567,37 @@ def test_mixed_matrix(mode):
...
+
+        # Test for correctness
+        y = result.dat.data
+        assert np.allclose(y, expected)
```
connorjward:
```suggestion
        assert np.allclose(result.dat.data, expected)
```

firedrake/interpolation.py
```diff
@@ -1038,32 +1038,45 @@ def _interpolator(tensor, expr, subset, access, bcs=None):
...
+        v = firedrake.Function(W)
+        expr = expr._ufl_expr_reconstruct_(operand, v=v)
+        copyin += (partial(dual_arg.dat.copy, v.dat),)
+
+        # Compute the reciprocal of the DOF multiplicity
+        wdat = W.make_dat()
         m_ = get_interp_node_map(source_mesh, target_mesh, W)
-        op2.par_loop(kernel, cell_set, weight.dat(op2.INC, m_))
-        with weight.dat.vec as w:
+        if m_.offset is None:
```
connorjward: Don't we have a clearer way to check if a mesh is extruded?

firedrake/interpolation.py
```diff
@@ -1038,32 +1038,45 @@ def _interpolator(tensor, expr, subset, access, bcs=None):
...
+            m_indices = m_.values_with_halo[cell_set.indices[:cell_set.size]]
+            m_shape = m_indices.shape + W.shape
+            with wdat.vec as w:
+                w.setLGMap(W.dof_dset.scalar_lgmap)
+                w.setValuesBlockedLocal(m_indices, numpy.ones(m_shape), PETSc.InsertMode.ADD)
+                w.assemble()
```
connorjward: I'm really not sure that this is any better. We now have two cases instead of one and this is not exactly obvious to read and understand. I think I would prefer leaving the C string for now and I can make it nice with pyop3 where it would look like:
```py
op3.loop(
  c := mesh.cells.owned.iter(),
  wdat[mesh.closure(c)].iassign(1),
  eager=True,
)
```

tests/firedrake/regression/test_interpolate.py
pbrubeck: What I had gave a more readable error message

firedrake/interpolation.py
pbrubeck: This is no longer relavant.

firedrake/interpolation.py
pbrubeck: That's neat. But this question still stands: should we be generating code for this?

connorjward: Why shouldn't we? There's no chance of this being a performance bottleneck

tests/firedrake/regression/test_interpolate.py
connorjward: I guess it's fine. But to me this was more readable.

firedrake/interpolation.py
pbrubeck: Surely it won't harm performance. But seems like an overkill, as we are generating different codes for different array sizes.

I also think our code generation should evolve to kernels that are degree-independent.

connorjward: We can't do that universally because we need to encode static loop extents to get good performance.
If you're really committed it wouldn't be too difficult to add as an optional compiler step in pyop3.

PR #4624, opened 2025-10-06
test_warped_interior_facet: ignore cell number 3 on the edge point

qbisi:
test_high_order_location_warped_interior_facet failed when i try to package firedrake-2025.10.0 in nxipkgs.

The failure happens because the point [0.25, 0.605] lies right on the boundary between the warped cells 1 and 3. In the test, we ignore cell 3 and assert that the point belongs to cell 1.


connorjward:
From my reading of the test this point should comfortably lie inside the cell. Any other result is a bug.
For your use case I would recommend patching out the test. But please feel free to open an issue.

qbisi:
By reducing warp shift from 0.1 to 0.04, some how alleviate the pain caculating if a point is belonged to a concave cell e.g. (cell 3 in this example). Still the edge points for cell3 does not match with cell1 everywhere.

qbisi:
superseded by https://github.com/firedrakeproject/firedrake/pull/4632

PR #4625, opened 2025-10-06
Distinguish firedrake-configure links without leading to git conflicts

connorjward:
# Description


connorjward:
Merging as the docs build and I've checked locally that it does the right thing.

PR #4634, opened 2025-10-08
LinearEigensolver: default to eps_target = 0 is conceptually wrong

stefanozampini:
The presence of the option, together with the `sinvert` option for the `bcs` case, can lead to confusion.
I think it is better to be consistent with whatever SLEPc believes is the proper default


stefanozampini:
@UZerbinati Are you ok with this PR?

dham (approved):

PR #4640, opened 2025-10-13
Firedrake container on ARM

connorjward:
This adds some code from https://github.com/jorgensd/dolfinx-tutorial/blob/dokken/ngs/docker/Dockerfile#L19-L44 to build netgen from source since wheels are not available on ARM. I have checked on our mac runners and it at least runs.

I found that icepack would not build on ARM so I removed all the Firedrake apps from the ARM image. I don' t know if this a good idea or not.

### DIFF
--- a/docker/Dockerfile.firedrake
+++ b/docker/Dockerfile.firedrake
@@ -1,4 +1,4 @@
-# Dockerfile for Firedrake with a full set of capabilities and applications installed.
+# Dockerfile for Firedrake with a full set of capabilities installed.
 
 FROM firedrakeproject/firedrake-vanilla-default:latest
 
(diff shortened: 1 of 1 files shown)


connorjward:
> The install docs on the icepack website are way out of date now. But I think you're right on both counts. I don't know of any icepack users who work off of the full-fat docker image. It's also an annoying maintenance burden for you all to have to support lots of downstream packages. For example, icepack really depends on rasterio but as per [this issue](https://github.com/rasterio/rasterio-wheels/issues/69) they don't have ARM wheels.

From the meeting yesterday we concluded that we should drop all downstream packages from the container. Instead the container will just contain things like jax, pytorch and ngspetsc.

Thanks for the feedback!

dham (approved):

PR #4641, opened 2025-10-13
Fix Hessian calculation for `NonlinearVariationalSolver` block

JHopeCollins:
This PR fixes a bug in the Hessian calculation when calling `solver.solve` multiple times on the same `NonlinearVariationalSolver`.

The Hessian calculation requires the adjoint variable. The `GenericSolveBlock` stores this on the block, so if there are multiple calls to `firedrake.solve` then each block has it's own adjoint solution and everything works fine.
However, the`NonlinearVariationalSolveBlock` stored the solution of the cached adjoint solver. This solver is shared between all blocks created by the same `NonlinearVariationalSolver` so if there are multiple blocks then they overwrite each others adjoint solutions.
This PR fixes this by making each block stash their own adjoint solution to reuse when calculating the Hessian action.

This is a quickfix to solve the issue on release. There's a bigger refactor of the cached solvers to improve the performance in this PR: https://github.com/firedrakeproject/firedrake/pull/4638 but that should probably go into main.

### DIFF
--- a/firedrake/adjoint_utils/blocks/solving.py
+++ b/firedrake/adjoint_utils/blocks/solving.py
@@ -165,8 +165,6 @@ class GenericSolveBlock(Block):
 
     @adj_sol.setter
     def adj_sol(self, value):
-        if value is None:
-            self.adj_state = None
         if self.adj_state is None:
             self.adj_state = value.copy(deepcopy=True)
         else:


firedrake/adjoint_utils/blocks/solving.py
```diff
@@ -399,6 +402,8 @@ def prepare_evaluate_hessian(self, inputs, hessian_inputs, adj_inputs,
         if tlm_output is None:
             return
 
+        self.adj_state.assign(self.adj_state_buf)
```
angus-g: I thought this was going to break the `RuntimeError` below because `self.adj_state_buf` (and therefore `self.adj_state`) is never `None`, but it turns out that error doesn't even get hit if you run `Jhat.hessian()` without an adjoint anyway. I guess it *could*, but the assembly block breaks first.

```
  File "/usr/local/lib/python3.12/dist-packages/pyadjoint/reduced_functional.py", line 289, in hessian
    r = compute_hessian(self.functional, self.controls, m_dot,
        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.12/dist-packages/pyadjoint/drivers.py", line 129, in compute_hessian
    tape.evaluate_hessian(markings=True)
  File "/usr/local/lib/python3.12/dist-packages/pyadjoint/tape.py", line 374, in evaluate_hessian
    self._blocks[i].evaluate_hessian(markings=markings)
  File "/usr/lib/python3.12/contextlib.py", line 81, in inner
    return func(*args, **kwds)
           ^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.12/dist-packages/pyadjoint/block.py", line 292, in evaluate_hessian
    hessian_output = self.evaluate_hessian_component(inputs,
                     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/firedrake/firedrake/adjoint_utils/blocks/assembly.py", line 197, in evaluate_hessian_component
    hessian_outputs += self.compute_action_adjoint(
                       ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/firedrake/firedrake/adjoint_utils/blocks/assembly.py", line 52, in compute_action_adjoint
    adj_output = dform_adj._ad_mul(adj_input)
                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/firedrake/firedrake/adjoint_utils/function.py", line 248, in _ad_mul
    r.assign(other * self)
             ~~~~~~^~~~~~
  File "/usr/local/lib/python3.12/dist-packages/ufl/form.py", line 207, in __rmul__
    if is_scalar_constant_expression(scalar):
       ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.12/dist-packages/ufl/checks.py", line 45, in is_scalar_constant_expression
    if expr.ufl_shape:
       ^^^^^^^^^^^^^^
AttributeError: 'NoneType' object has no attribute 'ufl_shape'
```

JHopeCollins: Which `RuntimeError` do you mean?
Running `Jhat.hessian` before running `Jhat.derivative` isn't valid anyway so I don't think we need to worry about what happens in that case.

angus-g: Just a bit further down in this function (L417)

```python
        adj_sol = self.adj_state
        if adj_sol is None:
            raise RuntimeError("Hessian computation was run before adjoint.")
```

It's probably useful, but I think this is now dead and doesn't get hit because of a prior crash anyway!

firedrake/adjoint_utils/blocks/solving.py
```diff
@@ -163,6 +163,15 @@ def _should_compute_boundary_adjoint(self, relevant_dependencies):
     def adj_sol(self):
         return self.adj_state
 
+    @adj_sol.setter
+    def adj_sol(self, value):
+        if value is None:
+            self.adj_state = None
```
dham:
```suggestion
```

PR #4648, opened 2025-10-15
BUG: fix multigrid transfers for symmetric tensors

pbrubeck:
# Description

### DIFF
--- a/firedrake/mg/kernels.py
+++ b/firedrake/mg/kernels.py
@@ -115,25 +115,34 @@ def compile_element(expression, dual_space=None, parameters=None,
     # # Collect required coefficients
 
     try:
+        # Forward interpolation: expression has a coefficient
         arg, = extract_coefficients(expression)
         argument_multiindices = ()
         coefficient = True
-        tensor_indices = None
     except ValueError:
+        # Adjoint interpolation: expression has an argument
         arg, = extract_arguments(expression)
         finat_elem = create_element(arg.ufl_element())
-        argument_multiindices = (finat_elem.get_indices(), )
-        argument_multiindex, = argument_multiindices
-        value_shape = finat_elem.value_shape
+        argument_multiindex = finat_elem.get_indices()
+        argument_multiindices = (argument_multiindex, )
+        coefficient = False
+
+    # Map into reference values
+    domain = extract_unique_domain(expression)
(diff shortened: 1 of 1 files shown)


tests/firedrake/multigrid/test_grid_transfer.py
```diff
@@ -296,30 +323,33 @@ def periodic_space(request, periodic_cell):
...
-    elif len(x) == 3:
+    elif dim == 3:
         expr = (1 - x[0]) * x[0] + \
                (1 - x[1]) * x[1] * x[1] + \
                (1 - x[2]) * x[2] * x[2] * x[2]
-    if vector:
-        expr = as_vector([(-1)**i * expr for i in range(len(x))])
+    if shape == "vector":
+        expr = as_vector([(-1)**i * expr for i in range(dim)])
+    elif shape == "symmetry":
```
pbrubeck:
```suggestion
    elif shape == "symmetric-tensor":
```

firedrake/mg/kernels.py
```diff
@@ -150,8 +149,12 @@ def compile_element(expression, dual_space=None, parameters=None,
                   scalar_type=parameters["scalar_type"])
     context = tsfc.fem.GemPointContext(**config)
 
+    # Map into reference values
+    expression = apply_mapping(expression, arg.ufl_element(), domain)
+    if tensor_indices is None:
+        tensor_indices = tuple(gem.Index() for s in expression.ufl_shape)
```
connorjward: Why can we set `tensor_indices` further up when `coefficient` is false but not when true? This is quite confusing.

pbrubeck: We need to map the expression on both cases and only then create the indices if we are in the first case. I can move the index creation for the other case here.

pbrubeck: I plan to remove this code in favour of vertex only mesh quite soon anyway

connorjward: If there's something simple you could do to make it more readable that would be good. Alternatively I'd be happy with a TODO comment saying how VoM would be used here instead.

pbrubeck: It should be cleaner now

connorjward (approved):

Issue #4125, opened 2025-03-18
BUG: ML Operator with Mixed Space input and Single Space output


tlroy:
**Describe the bug**
Currently bugged: ML operator that takes a mixed function as input, and outputs a single function as output, i.e. W -> V. It seems like only W->W and V->V are supported. The error is something about BC mismatch with trial space.

**Steps to Reproduce**
```
from firedrake import *
import firedrake.ml.pytorch as fd_ml
import torch
import torch.nn as nn
mesh = UnitIntervalMesh(4)
x = SpatialCoordinate(mesh)
V = FunctionSpace(mesh,"CG",1)
W = V*V
u = Function(W)
u1, u2 = split(u)
v1, v2 = TestFunctions(W)
model = nn.Linear(W.dim(), V.dim())
I = torch.eye(V.dim(), V.dim())
model.weight.data = torch.cat([I, I], dim=1)
model.bias.data = torch.zeros(V.dim())
model.eval()
ml_op = fd_ml.ml_operator(model,function_space=V,inputs_format=1)
i_surrogate = ml_op(u)
# This actually runs: 
# i_surrogate = ml_op(Function(W))
a = inner(grad(u1), grad(v1)) * dx
a += i_surrogate * v1 * ds(1) # equivalent to (u_1 + u_2) * v1 * ds(1)
a += inner(grad(u2), grad(v2)) * dx
bcs = [DirichletBC(W.sub(0), Constant(1), 2),
        DirichletBC(W.sub(1), Constant(1), 2)]
solve(a==0, u, bcs=bcs)
```

**Error message**
```
/home/firedrake/firedrake/lib/python3.12/site-packages/pytools/persistent_dict.py:52: RecommendedHashNotFoundWarning: Unable to import recommended hash 'siphash24.siphash13', falling back to 'hashlib.sha256'. Run 'python3 -m pip install siphash24' to install the recommended hash.
  warn("Unable to import recommended hash 'siphash24.siphash13', "
Traceback (most recent call last):
  File "petsc4py/PETSc/PETSc.pyx", line 348, in petsc4py.PETSc.PetscPythonErrorHandler
  File "petsc4py/PETSc/PETSc.pyx", line 348, in petsc4py.PETSc.PetscPythonErrorHandler
  File "petsc4py/PETSc/PETSc.pyx", line 348, in petsc4py.PETSc.PetscPythonErrorHandler
[...]

PR #4659, opened 2025-10-17
Bug: Fix bcs for BaseFormOperator on MixedFunctionSpace

pbrubeck:
# Description
This PR fixes several bugs related to applying `DirichletBC`s on `BaseFormOperators` involving `MixedFunctionSpace`.

1. `MixedInterpolator` was wrongly imposing bcs on `MixedFunctionSpaces` with repeated subspaces (which is a very common case).  For instance if `Z = V * V * Q`, the boundary condition `DirichletBC(Z.sub(0), "on_boundary")` would have been wrongly applied to both `Z.sub(0)` and `Z.sub(1)`, since they are equal to `V`. The fix is to check which bcs apply to each block of the matrix.

2. We also fix `assemble(interpolate(...), bcs=bcs)`, which was previously not passing the `bcs` on to the `Interpolator`.

3. Fixes https://github.com/firedrakeproject/firedrake/issues/4125, where we attempted to apply a DirichletBC to a bilinear form with different test and trial spaces. The fix is to zero the rows or columns corresponding to the BC nodes.

In addition, we also need to create jax/torch tensors with flattened `u.dat.data`. Suppose `u` comes from a `MixedFunctionSpace` then `torch.tensor(u.dat.data)` will receive a list of `numpy.arrays` and turn it into the columns of a matrix, which is an unexpected and undesired behaivor. A MixedFunctionSpace with subspaces of different dimension cannot be turned into a matrix in that way, as the columns would have different dimensions. Therefore `u.dat.data` must be flattened into a long vector.

### DIFF
--- a/firedrake/assemble.py
+++ b/firedrake/assemble.py
@@ -408,9 +408,6 @@ class BaseFormAssembler(AbstractFormAssembler):
         in a post-order fashion and evaluating the nodes on the fly.
 
         """
-        def parent_space(V):
-            return V.parent or V
-
         def visitor(e, *operands):
             t = tensor if e is self._form else None
             # Deal with 2-form bcs inside the visitor


tests/firedrake/external_operators/test_jax_operator.py
```diff
@@ -245,3 +247,31 @@ def test_solve(mesh, V):
...
     assert err_point_expr < 1.0e-09
+
+
+def test_mixed_space_bcs():
+    mesh = UnitIntervalMesh(4)
+    V = FunctionSpace(mesh, "CG", 1)
+    W = V * V
+
+    test = TestFunction(W)
+    bcs = [DirichletBC(W.sub(0), Constant(1), 2),
```
pbrubeck:
```suggestion
    bcs = [DirichletBC(W.sub(0), Constant(1), 1),
```

firedrake/assemble.py
```diff
@@ -408,23 +408,32 @@ def assemble(self, tensor=None, current_state=None):
         in a post-order fashion and evaluating the nodes on the fly.
 
         """
+        def parent_space(V):
+            return V.parent or V
+
         def visitor(e, *operands):
             t = tensor if e is self._form else None
-            return self.base_form_assembly_visitor(e, t, *operands)
+            # Deal with 2-form bcs inside the visitor
```
connorjward, line 413: Why do we deal with 2-form bcs here and 1-form bcs elsewhere? That seems like it will make the code very confusing to follow.

firedrake/assemble.py
```diff
@@ -1446,7 +1453,11 @@ def _apply_bc(self, tensor, bc, u=None):
             V = V.parent
         index = 0 if V.index is None else V.index
         space = V if V.parent is None else V.parent
-        if isinstance(bc, DirichletBC):
+        if spaces[0] != spaces[1] and isinstance(bc, DirichletBC):
```
connorjward: I think it would be clearer to have:
```py
if isinstance(bc, DirichletBC):
  if spaces[0] != spaces[1]:
    return  # with a useful comment

  ...
```

firedrake/interpolation.py
```diff
@@ -1703,7 +1703,11 @@ def __init__(self, expr, V, bcs=None, **kwargs):
         super(MixedInterpolator, self).__init__(expr, V, bcs=bcs, **kwargs)
         expr = self.ufl_interpolate
         self.arguments = expr.arguments()
-        rank = len(self.arguments)
+        # Get the primal spaces
+        spaces = tuple(a.function_space().dual() if isinstance(a, Coargument) else a.function_space()
+                       for a in self.arguments)
+        # We need a stricter equality test for indexed MixedFunctionSpace
+        space_equals = lambda V1, V2: V1 == V2 and V1.parent == V2.parent and V1.index == V2.index
```
connorjward, line 1711: This should probably be implemented inside the appropriate `__eq__`.

pbrubeck: That is an API breaking change, it should not go as part of this PR

firedrake/assemble.py
pbrubeck: The comment is to clarify that we were already treating bcs differently depending on the rank of the form. This separate treatment making the code much more easy to follow, as the 1-form case would require imposing bcs on every return statement in the visitor.

firedrake/interpolation.py
connorjward: But it's a bug, so arguably it's an API-fixing change

firedrake/assemble.py
pbrubeck: Matrices need the bcs at the moment of assembly, as we use masked LGMaps to not write on the boundary nodes. Functions can be easily modified post-assembly.

firedrake/interpolation.py
pbrubeck: Alright, I'll try this. If this breaks any tests, then I'll prefer to open a new PR

pbrubeck: This seems quite involved. In `MixedFunctionSpace.__eq__` we do a subspace comparison. If we let `FunctionSpace.__eq__` also check for `self.parent == other.parent` on the parent `MixedFunctionSpace`, we create an infinte recursion.

firedrake/assemble.py
connorjward: It seems like a strange place to be filtering the BCs. Could we not do this when we make a `TwoFormAssembler` object?

firedrake/interpolation.py
connorjward: OK. If you create an issue for this and leave a TODO comment here pointing at it then I'm happy.

firedrake/interpolation.py
```diff
@@ -1703,7 +1703,11 @@ def __init__(self, expr, V, bcs=None, **kwargs):
         super(MixedInterpolator, self).__init__(expr, V, bcs=bcs, **kwargs)
         expr = self.ufl_interpolate
         self.arguments = expr.arguments()
-        rank = len(self.arguments)
+        # Get the primal spaces
+        spaces = tuple(a.function_space().dual() if isinstance(a, Coargument) else a.function_space()
+                       for a in self.arguments)
+        # We need a stricter equality test for indexed MixedFunctionSpace
```
```suggestion
        # TODO consider a stricter equality test for indexed MixedFunctionSpace
        # See https://github.com/firedrakeproject/firedrake/issues/4668
```

firedrake/assemble.py
pbrubeck: I'll see if this works without filtering

firedrake/assemble.py
```diff
@@ -408,23 +408,32 @@ def assemble(self, tensor=None, current_state=None):
...
+            return V.parent or V
+
         def visitor(e, *operands):
             t = tensor if e is self._form else None
-            return self.base_form_assembly_visitor(e, t, *operands)
+            # Deal with 2-form bcs inside the visitor
+            rank = len(self._form.arguments())
+            if (rank == 2 and isinstance(e, ufl.BaseForm) and len(e.arguments()) == rank):
+                args_fs = set(arg.function_space().dual() if isinstance(arg, ufl.Coargument) else arg.function_space() for arg in e.arguments())
+                bcs = tuple(bc for bc in self._bcs if parent_space(bc.function_space()) in args_fs)
```
```suggestion
            if len(self._form.arguments()) == 2:
                bcs = self._bcs
```

firedrake/assemble.py
```diff
@@ -408,23 +408,30 @@ def assemble(self, tensor=None, current_state=None):
...
+
         def visitor(e, *operands):
             t = tensor if e is self._form else None
-            return self.base_form_assembly_visitor(e, t, *operands)
+            # Deal with 2-form bcs inside the visitor
+            if len(self._form.arguments()) == 2:
+                bcs = self._bcs
+            else:
+                bcs = ()
+            return self.base_form_assembly_visitor(e, t, bcs, *operands)
```
```suggestion
            return self.base_form_assembly_visitor(e, t, self._bcs, *operands)
```

firedrake/assemble.py
pbrubeck: we don't need the filtering, seems like that's done in `TwoFormAssembler`

firedrake/assemble.py
```diff
@@ -408,23 +408,27 @@ def assemble(self, tensor=None, current_state=None):
...
 
         """
+        def parent_space(V):
+            return V.parent or V
+
         def visitor(e, *operands):
             t = tensor if e is self._form else None
-            return self.base_form_assembly_visitor(e, t, *operands)
+            # Deal with 2-form bcs inside the visitor
+            bcs = self._bcs if isinstance(e, ufl.BaseForm) and len(e.arguments()) == 2 else ()
```
connorjward, line 414: Why is this necessary?

connorjward (commented):
Other than this query I think I'm happy.

firedrake/assemble.py
pbrubeck: We were always ignoring bcs for 1-form assembly already, this is just making the comments above more explicit

firedrake/assemble.py
```diff
@@ -408,23 +408,27 @@ def assemble(self, tensor=None, current_state=None):
         in a post-order fashion and evaluating the nodes on the fly.
 
         """
+        def parent_space(V):
+            return V.parent or V
```
connorjward: This isn't used any more.

firedrake/assemble.py
connorjward: Yeah I guess this is better.

firedrake/assemble.py
```diff
@@ -408,23 +408,27 @@ def assemble(self, tensor=None, current_state=None):
...
+
```
```suggestion
```

PR #4661, opened 2025-10-20
Download MUMPS on Linux because PETSc now assumes a higher version

JHopeCollins:
[This PETSc PR](https://gitlab.com/petsc/petsc/-/merge_requests/8715) below increased the minimum acceptable MUMPS version, so the MUMPS 5.6 that is shipped with Ubuntu24.04 no longer works with PETSc main.

This PR swaps Firedrake to asking PETSc to download MUMPS instead so we get a recent enough version.

### DIFF
--- a/scripts/firedrake-configure
+++ b/scripts/firedrake-configure
@@ -204,7 +204,6 @@ PETSC_EXTRAS_LINUX_APT_PACKAGES = (
     "libfftw3-mpi-dev",
     "libhwloc-dev",
     "libhdf5-mpi-dev",
-    "libmumps-ptscotch-dev",
     "libmetis-dev",
     "libnetcdf-dev",
     "libpnetcdf-dev",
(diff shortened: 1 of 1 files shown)


scripts/firedrake-configure
connorjward, line 334: I wonder if this will cause any issues given the package is called `libmumps-ptscotch-dev`. We do also install `libptscotch-dev`.

connorjward (commented):
Should we open an issue to revisit this when either PETSc support an older MUMPS again or Ubuntu LTS changes?

scripts/firedrake-configure
JHopeCollins: Given that PTScotch is the partitioner, I assume that MUMPS relies on PTScotch not the other way around. So unless more recent MUMPS versions require more recent PTScotch versions, we should be fine.

JHopeCollins: > Should we open an issue to revisit this when either PETSc support an older MUMPS again or Ubuntu LTS changes?

https://github.com/firedrakeproject/firedrake/issues/4662

connorjward (approved):

PR #4663, opened 2025-10-20
Extruded Periodic MeshHierarchy

pbrubeck:
# Description

`ExtrudedMeshHierarchy` now forwards its `periodic` option to every extruded mesh in the hierarchy. The multigrid node-map builders also support periodic extrusion, including finite-element offset quotients at the seam.

For periodic extrusion, the previous Cython implementation used unwrapped layer offsets. At the top layer this generated out-of-range fine-node indices. The fix applies the same periodic layer transformation that PyOP2 uses, in both coarse-to-fine and fine-to-coarse node maps. The map inputs and outputs now use C-contiguous typed Cython memoryviews, while `np.asarray` returns the existing buffers without copying.

The periodic multigrid regression now uses a smooth periodic exact solution and derives the forcing from it. The previous `sin(pi*z)` exact solution had a derivative jump across the periodic seam, so a direct LU solve could not satisfy the test threshold.

Contributions assisted by Codex (GPT-5).

# Testing

* `python setup.py build_ext --inplace`
* `mpiexec -n 1 python -m pytest -m 'parallel[match]' tests/firedrake/multigrid/test_poisson_gmg_extruded.py -q` — 4 passed, 4 deselected
* `mpiexec --oversubscribe -n 3 python -m pytest -m 'parallel[match]' tests/firedrake/multigrid/test_poisson_gmg_extruded.py -q` — 4 passed, 4 deselected per rank
* `make srclint FLAKE8_FORMAT=--jobs=1`

### DIFF
--- a/firedrake/mg/mesh.py
+++ b/firedrake/mg/mesh.py
@@ -204,7 +204,7 @@ def MeshHierarchy(mesh, refinement_levels,
 
 
 def ExtrudedMeshHierarchy(base_hierarchy, height, base_layer=-1, refinement_ratio=2, layers=None,
-                          kernel=None, extrusion_type='uniform', gdim=None,
+                          kernel=None, extrusion_type='uniform', periodic=False, gdim=None,
                           mesh_builder=firedrake.ExtrudedMesh):
     """Build a hierarchy of extruded meshes by extruding a hierarchy of meshes.
 
@@ -245,6 +245,7 @@ def ExtrudedMeshHierarchy(base_hierarchy, height, base_layer=-1, refinement_rati
     meshes = [mesh_builder(m, layer, kernel=kernel,
                            layer_height=height/layer,
                            extrusion_type=extrusion_type,
+                           periodic=periodic,
                            gdim=gdim)
               for (m, layer) in zip(base_hierarchy._meshes, layers)]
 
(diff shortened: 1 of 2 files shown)


firedrake/mg/mesh.py
```diff
@@ -318,7 +318,7 @@ def MeshHierarchy(mesh, refinement_levels=0,
 
 
 def ExtrudedMeshHierarchy(base_hierarchy, height, base_layer=-1, refinement_ratio=2, layers=None,
-                          kernel=None, extrusion_type='uniform', gdim=None,
+                          kernel=None, extrusion_type='uniform', periodic=False, gdim=None,
```
connorjward: Add to the docstring

connorjward (changes requested):
It would be really nice if you'd stop changing code that I will have to fix in pyop3 :)

pbrubeck:
> It would be really nice if you'd stop changing code that I will have to fix in pyop3 :)

I am not doing this on purpose. Two users requested this functionality.

firedrake/mg/mesh.py
pbrubeck: done

connorjward (dismissed):

PR #4664, opened 2025-10-20
fem: cache on instances

ksagiyam:
Cache the result of `PointSetContext.basis_evaluation()` on instance.

Currently, result of `ctx.basis_evaluation()`, where `ctx =  PointSetContext()`, is cached on class, but the instance is not included in the cache key. This means that implementation changes around `basis_evaluation()` could cause hard-to-find caching bug. This PR makes the result cached on instance not on class.

We should note that the way we handle `mt` in the cache key generation is also fragile for the same reason.

### DIFF
--- a/tsfc/fem.py
+++ b/tsfc/fem.py
@@ -284,15 +284,6 @@ def get_quadrature_rule(fiat_cell, integration_dim, quadrature_degree, scheme):
     return make_quadrature(integration_cell, quadrature_degree, scheme=scheme)
 
 
-def make_basis_evaluation_key(ctx, finat_element, mt, entity_id):
-    ufl_element = mt.terminal.ufl_element()
-    domain = extract_unique_domain(mt.terminal)
-    coordinate_element = domain.ufl_coordinate_element()
-    # This way of caching is fragile.
-    # Should Implement _hash_key_() for ModifiedTerminal and use the entire mt as key.
-    return (ufl_element, mt.local_derivatives, ctx.point_set, ctx.integration_dim, entity_id, coordinate_element, mt.restriction, domain._ufl_hash_data_())
-
-
 class PointSetContext(ContextBase):
     """Context for compile-time known evaluation points."""
 
@@ -323,12 +314,32 @@ class PointSetContext(ContextBase):
     def weight_expr(self):
         return self.quadrature_rule.weight_expression
 
-    @serial_cache(hashkey=make_basis_evaluation_key)
+    @staticmethod
+    def _make_basis_evaluation_key(finat_element, mt, entity_id):
+        ufl_element = mt.terminal.ufl_element()
+        domain = extract_unique_domain(mt.terminal)
+        coordinate_element = domain.ufl_coordinate_element()
+        # This way of caching is fragile.
+        # Should implement _hash_key_() in ModifiedTerminal and include the entire mt in the key,
+        # or only pass necessary bits in mt to basis_evaluation.
+        return (ufl_element, mt.local_derivatives, entity_id, coordinate_element, mt.restriction, domain._ufl_hash_data_())
+
+    @cached_property
+    def _basis_evaluation_cache(self):
+        return {}
+
     def basis_evaluation(self, finat_element, mt, entity_id):
-        return finat_element.basis_evaluation(mt.local_derivatives,
-                                              self.point_set,
-                                              (self.integration_dim, entity_id),
-                                              coordinate_mapping=CoordinateMapping(mt, self))
+        key = PointSetContext._make_basis_evaluation_key(finat_element, mt, entity_id)
+        try:
+            return self._basis_evaluation_cache[key]
+        except KeyError:
+            val = finat_element.basis_evaluation(
+                mt.local_derivatives,
+                self.point_set,
+                (self.integration_dim, entity_id),
+                coordinate_mapping=CoordinateMapping(mt, self),
+            )
+            return self._basis_evaluation_cache.setdefault(key, val)
 
 
 class GemPointContext(ContextBase):


tsfc/fem.py
```diff
@@ -323,12 +314,33 @@ def point_expr(self):
...
+        ufl_element = mt.terminal.ufl_element()
+        domain = extract_unique_domain(mt.terminal)
+        coordinate_element = domain.ufl_coordinate_element()
+        # This way of caching is fragile.
+        # Should implement _hash_key_() in ModifiedTerminal and include the entire mt in the key,
+        # or only pass necessary bits in mt to basis_evaluation.
+        return (ufl_element, mt.local_derivatives, entity_id, coordinate_element, mt.restriction, domain._ufl_hash_data_())
+
+    @cached_property
+    def _basis_evalueation_cache(self):
```
leo-collins:
```suggestion
    def _basis_evaluation_cache(self):
```

tsfc/fem.py
```diff
@@ -323,12 +314,33 @@ def point_expr(self):
...
+        try:
+            return self._basis_evalueation_cache[key]
+        except KeyError:
+            val = finat_element.basis_evaluation(
+                mt.local_derivatives,
+                self.point_set,
+                (self.integration_dim, entity_id),
+                coordinate_mapping=CoordinateMapping(mt, self),
+            )
+            self._basis_evalueation_cache[key] = val
```
leo-collins:
```suggestion
            self._basis_evaluation_cache[key] = val
```

tsfc/fem.py
```diff
@@ -323,12 +314,33 @@ def point_expr(self):
...
+        return {}
+
     def basis_evaluation(self, finat_element, mt, entity_id):
-        return finat_element.basis_evaluation(mt.local_derivatives,
-                                              self.point_set,
-                                              (self.integration_dim, entity_id),
-                                              coordinate_mapping=CoordinateMapping(mt, self))
+        key = PointSetContext._make_basis_evaluation_key(finat_element, mt, entity_id)
+        try:
+            return self._basis_evalueation_cache[key]
```
leo-collins:
```suggestion
            return self._basis_evaluation_cache[key]
```

tsfc/fem.py
```diff
@@ -323,12 +314,33 @@ def point_expr(self):
...
-                                              coordinate_mapping=CoordinateMapping(mt, self))
+        key = PointSetContext._make_basis_evaluation_key(finat_element, mt, entity_id)
+        try:
+            return self._basis_evaluation_cache[key]
+        except KeyError:
+            val = finat_element.basis_evaluation(
+                mt.local_derivatives,
+                self.point_set,
+                (self.integration_dim, entity_id),
+                coordinate_mapping=CoordinateMapping(mt, self),
```
pbrubeck, line 340: We could change the constructor to `CoordinateMapping(domain, self)` where `domain = extract_unique_domain(mt.terminal)`.

tsfc/fem.py
```diff
@@ -323,12 +314,33 @@ def point_expr(self):
...
-    @serial_cache(hashkey=make_basis_evaluation_key)
+    @staticmethod
+    def _make_basis_evaluation_key(finat_element, mt, entity_id):
+        ufl_element = mt.terminal.ufl_element()
+        domain = extract_unique_domain(mt.terminal)
+        coordinate_element = domain.ufl_coordinate_element()
+        # This way of caching is fragile.
+        # Should implement _hash_key_() in ModifiedTerminal and include the entire mt in the key,
+        # or only pass necessary bits in mt to basis_evaluation.
+        return (ufl_element, mt.local_derivatives, entity_id, coordinate_element, mt.restriction, domain._ufl_hash_data_())
```
pbrubeck, line 325: Are we using `mt.restriction` anywhere?

tsfc/fem.py
```diff
@@ -323,12 +314,33 @@ def point_expr(self):
...
+            return self._basis_evaluation_cache[key]
+        except KeyError:
+            val = finat_element.basis_evaluation(
+                mt.local_derivatives,
+                self.point_set,
+                (self.integration_dim, entity_id),
+                coordinate_mapping=CoordinateMapping(mt, self),
+            )
+            self._basis_evaluation_cache[key] = val
+            return val
```
pbrubeck:
```suggestion
            return self._basis_evaluation_cache.setdefault(key, val)
```

tsfc/fem.py
pbrubeck: Seems like `basis_evaluation(finat_element, mt, entity_id)` is only using `mt.local_derivatives` and `extract_unique_domain(mt.terminal)`

pbrubeck: this might suggest that basis_evaluation could even change to something like

```
def basis_evaluation(self, finat_element, derivative_order, domain, entity_id)
```

pbrubeck: Ah NVM, `mt.restriction` is used in `CoordinateMapping`. But the same still holds
```
def basis_evaluation(self, finat_element, derivative_order, domain, restriction, entity_id)
```

tsfc/fem.py
pbrubeck: It should be `CoordinateMapping(domain, mt.restriction, self)`

tsfc/fem.py
ksagiyam: We can do that in a separate PR and remove those comments, but the fix in this PR is the only thing that I need immediately.

pbrubeck: Alright. Is this caching causing problems somewhere else?

ksagiyam: I don't know.

PR #4670, opened 2025-10-21
BUG: Fix FormSum involving the output tensor

pbrubeck:
# Description

Fixes `assemble(c1 * tensor + c2 * other_tensor, tensor=tensor)
`

### DIFF
--- a/firedrake/assemble.py
+++ b/firedrake/assemble.py
@@ -420,21 +420,22 @@ class BaseFormAssembler(AbstractFormAssembler):
         """
         def visitor(e, *operands):
             t = tensor if e is self._form else None
-            return self.base_form_assembly_visitor(e, t, *operands)
+            # Deal with 2-form bcs inside the visitor
+            bcs = self._bcs if isinstance(e, ufl.BaseForm) and len(e.arguments()) == 2 else ()
+            return self.base_form_assembly_visitor(e, t, bcs, *operands)
 
         # DAG assembly: traverse the DAG in a post-order fashion and evaluate the node on the fly.
         visited = {}
         result = BaseFormAssembler.base_form_postorder_traversal(self._form, visitor, visited)
 
-        # Apply BCs after assembly
+        # Deal with 1-form bcs outside the visitor
         rank = len(self._form.arguments())
         if rank == 1 and not isinstance(result, ufl.ZeroBaseForm):
             for bc in self._bcs:
                 OneFormAssembler._apply_bc(self, result, bc, u=current_state)
-
         return result
 
-    def base_form_assembly_visitor(self, expr, tensor, *args):
+    def base_form_assembly_visitor(self, expr, tensor, bcs, *args):
         r"""Assemble a :class:`~ufl.classes.BaseForm` object given its assembled operands.
 
             This functions contains the assembly handlers corresponding to the different nodes that
@@ -455,7 +456,7 @@ class BaseFormAssembler(AbstractFormAssembler):
                 assembler = OneFormAssembler(form, form_compiler_parameters=self._form_compiler_params,
                                              zero_bc_nodes=self._zero_bc_nodes, diagonal=self._diagonal, weight=self._weight)
             elif rank == 2:
-                assembler = TwoFormAssembler(form, bcs=self._bcs, form_compiler_parameters=self._form_compiler_params,
+                assembler = TwoFormAssembler(form, bcs=bcs, form_compiler_parameters=self._form_compiler_params,
                                              mat_type=self._mat_type, sub_mat_type=self._sub_mat_type,
                                              options_prefix=self._options_prefix, appctx=self._appctx, weight=self._weight,
                                              allocation_integral_types=self.allocation_integral_types)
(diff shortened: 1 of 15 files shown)


firedrake/assemble.py
```diff
@@ -522,21 +529,44 @@ def base_form_assembly_visitor(self, expr, tensor, *args):
...
+
+            # Compress repeated terms
+            terms = defaultdict(int)
+            for arg, weight in zip(args, weights):
+                t = self._as_pyop2_type(arg)
+                terms[t] += weight
+
+            # Zero the output tensor, or rescale it if it appears in the sum
+            tensor_scale = terms.pop(self._as_pyop2_type(tensor), 0)
+            if tensor is None:
```
```suggestion
            if tensor is None or tensor_scale == 1:
```

firedrake/assemble.py
```diff
@@ -522,21 +529,44 @@ def base_form_assembly_visitor(self, expr, tensor, *args):
             if all(isinstance(op, numbers.Complex) for op in args):
                 result = sum(weight * arg for weight, arg in zip(weights, args))
                 return tensor.assign(result) if tensor else result
-            elif (all(isinstance(op, firedrake.Cofunction) for op in args)
+
+            # Compress repeated terms
```
```suggestion
            # Factor common terms
```

firedrake/assemble.py
```diff
@@ -386,6 +386,16 @@ def allocation_integral_types(self):
         else:
             return self._allocation_integral_types
 
+    @staticmethod
+    def _as_pyop2_type(tensor, indices=None):
+        if isinstance(tensor, (firedrake.Cofunction, firedrake.Function)):
```
leo-collins:
```suggestion
        if isinstance(tensor, firedrake.Cofunction | firedrake.Function):
```

firedrake/assemble.py
```diff
@@ -522,21 +529,44 @@ def base_form_assembly_visitor(self, expr, tensor, *args):
             if all(isinstance(op, numbers.Complex) for op in args):
                 result = sum(weight * arg for weight, arg in zip(weights, args))
                 return tensor.assign(result) if tensor else result
-            elif (all(isinstance(op, firedrake.Cofunction) for op in args)
+
+            # Compress repeated terms
+            terms = defaultdict(int)
```
connorjward: I reckon it would be clearer to instead partition the sum into tensor bits and non-tensor bits

firedrake/matrix.py
```diff
@@ -240,8 +245,8 @@ def __init__(self, a, bcs, petscmat, *args, **kwargs):
         if options_prefix is not None:
             self.petscmat.setOptionsPrefix(options_prefix)
 
-        # this allows call to self.M.handle without a new class
-        self.M = SimpleNamespace(handle=self.mat())
+        # this mimics op2.Mat.handle
+        self.M = DummyOP2Mat(self.mat())
```
connorjward, line 249: This is unpleasant, but no worse than what was there before.

pbrubeck: Should I define DummyOP2Mat as an internal class? 
```
class AssembledMatrix:

    class DummyOP2Mat:
```

connorjward: I don't think so. That doesn't make things less confusing.

firedrake/assemble.py
pbrubeck: I clarified the logic in the comments.

firedrake/assemble.py
```diff
@@ -519,27 +526,54 @@ def base_form_assembly_visitor(self, expr, tensor, bcs, *args):
...
 
+            # Scalar FormSum
             if all(isinstance(op, numbers.Complex) for op in args):
-                result = sum(weight * arg for weight, arg in zip(weights, args))
+                result = numpy.dot(weights, args)
                 return tensor.assign(result) if tensor else result
-            elif (all(isinstance(op, firedrake.Cofunction) for op in args)
+
+            # Accumulate coefficients in a dictionary for each unique Dat/Mat
+            terms = defaultdict(int)
```
connorjward:
```suggestion
            terms = defaultdict(float)
```
or even `complex`?

pbrubeck: I don't think complex is correct, as it will force complex results in real mode. Also the role of `int` is just to define the empty sum, which is always an integer.

pbrubeck: If you really dislike `int`, we can go for `defaultdict(PETSc.ScalarType)`

firedrake/assemble.py
```diff
@@ -519,24 +526,51 @@ def base_form_assembly_visitor(self, expr, tensor, *args):
                     raise ValueError("Expecting a scalar weight expression")
                 weights.append(w)
 
+            # Scalar FormSum
             if all(isinstance(op, numbers.Complex) for op in args):
-                result = sum(weight * arg for weight, arg in zip(weights, args))
+                result = numpy.dot(weights, args)
                 return tensor.assign(result) if tensor else result
```
```suggestion
                return tensor.assign(result) if tensor else result.item()
```

firedrake/assemble.py
connorjward: Or `defaultdict(lambda: 0)`

It's a small thing but it just reads weird. You would expect `defaultdict(int)` to be a dictionary containing integers.

pbrubeck:
The G-adopt and linkcheck failures are expected. I'm merging this

PR #4675, opened 2025-10-23
Fix point evaluation API change

leo-collins:
# Description

I accidentally changed the API of `Function.evaluate` in https://github.com/firedrakeproject/firedrake/pull/4543. This changes the shape of the output to how it was before. Now, if you pass a single point to `Function.evaluate`, then we remove the first axis of the result. For example, if

```
mesh = UnitSquareMesh(1, 1)
u = Function(V)
```

then `u((0.5, 0.5))` returns a numpy array with shape `()`, and not an array with shape `(1,)`. If `u` lives in a `VectorFunctionSpace`, then `u((0.5, 0.5))` returns an array with shape`(2,)`, and not an array with shape `(1,2)`.

There were no tests looking for this before so I've added some.

### DIFF
--- a/firedrake/function.py
+++ b/firedrake/function.py
@@ -819,7 +819,7 @@ class PointEvaluator:
             if self.mesh.comm.rank != 0:
                 result = np.empty((len(self.points),) + shape, dtype=utils.ScalarType)
             self.mesh.comm.Bcast(result)
-        return result
+        return result.reshape((-1,) + shape)
 
 
 @PETSc.Log.EventDecorator()
--- a/firedrake/function.py
+++ b/firedrake/function.py
@@ -820,7 +820,7 @@ class PointEvaluator:
             if self.mesh.comm.rank != 0:
                 result = np.empty((len(self.points),) + shape, dtype=utils.ScalarType)
             self.mesh.comm.Bcast(result)
-        return result.reshape((-1,) + shape)
+        return result
 
 
 @PETSc.Log.EventDecorator()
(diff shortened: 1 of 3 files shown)


firedrake/function.py
```diff
@@ -549,8 +549,12 @@ def evaluate(self, coord, mapping, component, index_values):
         # Called by UFL when evaluating expressions at coordinates
         if component or index_values:
             raise NotImplementedError("Unsupported arguments when attempting to evaluate Function.")
+        coord = np.asarray(coord, dtype=utils.ScalarType)
         evaluator = PointEvaluator(self.function_space().mesh(), coord)
-        return evaluator.evaluate(self)
+        result = evaluator.evaluate(self)
+        if len(coord.shape) == 1:
+            result = result.squeeze()
```
pbrubeck: Do we want to preserve the shape of `coord`?
```suggestion
        result = result.reshape(coord.shape[:-1] + self.ufl_shape)
```

pbrubeck: I think that this change should go in `PointEvaluator`, for consistency

leo-collins: That would be another API change though

connorjward: `PointEvaluator` is brand new so almost no-one will be bitten by the change. I think it's probably fine. If that's not the API you want for `PointEvaluator` then that's fine too.

dham (dismissed):

connorjward (approved):

PR #4683, opened 2025-10-28
Cofunction: in-place assignment

pbrubeck:
# Description
Previously Cofunction.assign introduced a potential memory leak by assembling into a temporary buffer. 

This PR adds support for in-place assignment thanks to https://github.com/firedrakeproject/firedrake/pull/4670.

### DIFF
--- a/firedrake/cofunction.py
+++ b/firedrake/cofunction.py
@@ -209,6 +209,10 @@ class Cofunction(ufl.Cofunction, CofunctionMixin):
                 self.block_variable = self.create_block_variable()
                 self.block_variable._checkpoint = DelegatedFunctionCheckpoint(
                     expr.block_variable)
+                # We set CofunctionAssignBlock(..., rhs_from_assemble=True)
+                # so that we do not annotate the recursive call to assign
+                # within Cofunction.assign(BaseForm, subset=...).
+                # But we currently do not implement annotation for subset != None.
                 get_working_tape().add_block(
                     CofunctionAssignBlock(
                         self, expr, rhs_from_assemble=expr_from_assemble)


firedrake/cofunction.py
```diff
@@ -220,10 +220,11 @@ def assign(self, expr, subset=None, expr_from_assemble=False):
             # Enable c.assign(B) where c is a Cofunction and B an appropriate
             # BaseForm object. If annotation is enabled, the following
             # operation will result in an assemble block on the Pyadjoint tape.
```
pbrubeck, line 226: This comment might not be true anymore. Annotation does not support `subset != None` for `Cofunctions` (see line 206). 

@dham @JHopeCollins What are the implications?

pbrubeck: We might never hit `rhs_from_assemble == True` in `CofunctionAssignBlock.__init__`
https://github.com/firedrakeproject/firedrake/blob/e8eb716d867f42a8d4249c634b64d5232101e3b8/firedrake/adjoint_utils/blocks/function.py#L316-L330

JHopeCollins: > This comment might not be true anymore. Annotation does not support subset != None for Cofunctions 

I don't think I understand. `assemble` is called whatever the value of `subset` so there should be an assemble block added to the tape no matter what.

pbrubeck: Maybe the comment in cofunction.py is still true. But I am certain that the extra flag is no longer doing anything. Previously we were calling `assign` within `assign` if the expression was an unassembled `BaseForm`. The flag seems to be there in order to skip the annotation of the second assign block, but this is not a problem if we assemble in-place. So maybe we could get rid of these flags.

JHopeCollins: Fair enough. Do we ever go down that code path then?

pbrubeck: We are only ever going to use that codepath if we call `Cofunction.assign(BaseForm, subset=subset)`, but annotation is not supported if `subset != None`. So it is good to throw the error there, and leave things as they are in case someone needs annotation for that in the future.

JHopeCollins: Ok, that sounds reasonable. Could you add a note to the comment to record this please?

pbrubeck: done

PR #4690, opened 2025-10-29
variational solver will raise a warning for unused options

JHopeCollins:
The `petsctools` implementation now just relies on the user passing `-options_left` or not, so there's no need to change anything in Firedrake.

~~I bumped the implementation in #3999 up to petsctools in https://github.com/firedrakeproject/petsctools/pull/21. This PR just adds a call to the new `OptionsManager.warn_unused_options` method at the end of the `solve` call.~~

### DIFF
--- a/firedrake/variational_solver.py
+++ b/firedrake/variational_solver.py
@@ -395,7 +395,9 @@ class NonlinearVariationalSolver(OptionsManager, NonlinearVariationalSolverMixin
         self.warn_unused_options(
             options_to_ignore=set(
                 *self.DEFAULT_SNES_PARAMETERS.keys(),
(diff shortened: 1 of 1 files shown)


pyproject.toml
```diff
@@ -32,7 +32,7 @@ dependencies = [
   "packaging",
   # TODO RELEASE
   # "petsc4py==3.24.0",
-  "petsctools",
+  "petsctools @ git+https://github.com/firedrakeproject/petsctools.git@JHopeCollins/warn_unused_options",
```
JHopeCollins, line 35: Revert before merge:
```suggestion
  "petsctools",
```

PR #4693, opened 2025-10-30
ImplicitMatrixContext: create submatrix via MatCreateSubMatrixVirtual

pbrubeck:
# Description

### DIFF
--- a/tests/firedrake/regression/test_matrix_free.py
+++ b/tests/firedrake/regression/test_matrix_free.py
@@ -366,3 +366,45 @@ def test_matrix_free_fieldsplit_with_real():
             }}
     stokes_solver = LinearVariationalSolver(stokes_problem, solver_parameters=opts)
     stokes_solver.solve()
+
+
+@pytest.mark.parametrize("shape", ["scalar", "mixed"])
+def test_sub_matrix_not_subfield(shape):
+    mesh = UnitSquareMesh(2, 2)
+    if shape == "mixed":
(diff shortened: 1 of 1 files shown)


firedrake/matrix_free/operators.py
```diff
@@ -377,11 +377,20 @@ def createSubMatrix(self, mat, row_is, col_is, target=None):
...
+            if row_is == col_is and row_ises == col_ises:
+                col_inds = row_inds
+            else:
+                col_inds = find_sub_block(col_is, col_ises, comm=self.comm)
+        except LookupError:
+            # Attemping to extract a submatrix that does not match with a subfield.
+            # Use default PETSc implementation (MatCreateSubMatrixVirtual) via MATSHELL instead.
+            popmethod = self.createSubMatrix
+            self.createSubMatrix = None
+            submat = mat.createSubMatrix(row_is, col_is)
```
JHopeCollins, line 391: Should the output argument be forwarded too?
```suggestion
            submat = mat.createSubMatrix(row_is, col_is, submat=target)
```

pbrubeck: `target` should be None by this point

pbrubeck: Seems like that the firedrake implementation already assumed that `target` could not be uninitialized (a null pointer), but either `None` or the previous result of this function. What we do at the top of the function seems also correct for the non-subfiled case.

PR #4697, opened 2025-11-03
add __all__ to cofunction.py

### LINKED ISSUE #4696: BUG (I guess???!!): `import numpy as np` leaked by firedrake.
It's not necessarily a bug, but probably not the intended behaviour.  When doing an `from firedrake import *`, np is leaked. 

```
In [1]: from firedrake import *
In [2]: np
Out[2]: <module 'numpy' from '/Users/sghelichkhani/Workplace/firedrake-2025-10_08/venv-firedrake/lib/python3.13/site-packages/numpy/__init__.py'>
```

The issue is simply that `cofunction.py` is missing an `__all__` list, therefore `from firedrake.cofunction import *` in the main `__init__.py` ends loading numpy as np. 

## Solution:
`cofunction.py` should define an `__all__` list that explicitly lists what should be exported, excluding `np`.

dham:
This was causing namespace leakage into the main `firedrake` namespace.

fixes #4696.


connorjward:
> > Shouldn't this go into release?
> 
> Good question. Technically it's an API change and someone's code might depend on Firedrake exposing `np`.

Fair enough. Sometimes it's hard to decide exactly where the line is.

connorjward:
This appears to be causing a lot of failures. It's not just `np` that is accidentally getting exported I think.

connorjward (approved):

Issue #4687, opened 2025-10-29
BUG: MatNest with multicomponent fieldsplit fails to update Jacobian


pbrubeck:
**Describe the bug**
Consider a nonlinear problem solved with a fieldsplit PC where blocks might contain more than one component. It appears that the Jacobian is not correctly updated. 

**Steps to Reproduce**

```
from firedrake import *
mesh = UnitIntervalMesh(1)
V = FunctionSpace(mesh, "DG", 0)
Z = V * V * V

u = Function(Z)
u1, u2, u3 = split(u)
v1, v2, v3 = TestFunctions(Z)

F = inner(u1, v1) * dx
F += inner(0.5*u2**2 + u2, v2) * dx
F += inner(u3, v3) * dx
u2.assign(Constant(1))

sp = {
    "mat_type": "nest",
     "snes_max_it": 10,
     "ksp_type": "fgmres",
     "pc_type": "fieldsplit",
     "pc_fieldsplit_type": "additive",
      "pc_fieldsplit_0_fields": "0",
      "pc_fieldsplit_1_fields": "1,2",
      "fieldsplit_1_ksp_view_eigenvalues": None,
      "fieldsplit": {
          "ksp_type": "gmres",
          "pc_type": "lu"},
}
solve(F == 0, u, solver_parameters=sp)
```
prints
```
  0 SNES Function norm 1.500000000000e+00
Iteratively computed eigenvalues
1. + 0.i
  1 SNES Function norm 2.812500000000e-01
Iteratively computed eigenvalues
0.625 + 0.i
  2 SNES Function norm 2.531250000000e-02
Iteratively computed eigenvalues
0.5125 + 0.i
  3 SNES Function norm 3.049245240928e-04
Iteratively computed eigenvalues
0.500152 + 0.i
  4 SNES Function norm 4.646114840946e-08
Iteratively computed eigenvalues
0.5 + 0.i
  5 SNES Function norm 1.079319060632e-15
```

**Expected behavior**
If the Jacobian is correctly updated, the eigenvalue should always equal to 1. 

**Error message**
No error message

**Additional Info**
The (1, 1)-block of Jacobian for this problem is the 1x1 matrix `u2 + 1`. The eigenvalues of P^{-1} A indicate that `P[1, 1] = 1 + 1` at every Newton step but `A[1, 1] = u2 + 1` is correctly updated.

PR #4708, opened 2025-11-04
Fieldsplit: update MatNest Jacobian and support bcs on AssembledMatrix

pbrubeck:
# Description
Fixes https://github.com/firedrakeproject/firedrake/issues/4687

Depends on https://gitlab.com/petsc/petsc/-/merge_requests/8860 to trigger preconditioner update for a MatNest with more than one field on a block of the fieldsplit.

Also fixes fieldsplit for `AssembledMatrix` with bcs

Cleanup `create_subdm` in `dmhooks.py`


firedrake/solving_utils.py
```diff
@@ -504,6 +514,15 @@ def form_jacobian(snes, X, J, P):
...
         ctx.set_nullspace(ctx._near_nullspace, ises, transpose=False, near=True)
 
+        # Bump petsc matrix state of each split by assembling them.
+        # Ensures that if the matrix changed, the preconditioner is
+        # updated if necessary.
+        for field, splits in ctx._splits.items():
+            for subctx in splits:
+                subctx._jac.assemble()
+                if subctx.Jp is not None:
+                    subctx._pjac.assemble()
```
pbrubeck: @stefanozampini I think this should be called elsewhere, maybe in pyop2 or PETSc, but so far this fixes the bug https://github.com/firedrakeproject/firedrake/issues/4687. This is very similar to what we had to do for MatIS

https://github.com/firedrakeproject/firedrake/blob/0ce9ad22b5f5dddfb5c1685efaa8c6713df6c15e/firedrake/assemble.py#L1496-L1498

firedrake/solving_utils.py
stefanozampini, line 512: Shouldn't the update of the subfield operators be done inside `_assemble_jac`?

stefanozampini: `MatAssembly` for `MATNEST` calls the subblocks assembly https://gitlab.com/petsc/petsc/-/blob/main/src/mat/impls/nest/matnest.c?ref_type=heads#L481 ,so maybe you have special code in _assemble_jac that skips calling MatAssemble if the matrix is of type nest?

pbrubeck: I think the issue is that the bug involves several MatNests at play. Suppose you split a 3x3 MatNest into a 2x2 and 1x1 blocks in the diagonal. Every block of 2x2 MatNest gets updated (I don't know where but I think PETSc coordinates this), except that no one seems to be calling assemble() on the 2x2 MatNest, hence the preconditioner does not get updated.

tests/firedrake/regression/test_nested_fieldsplit_solves.py
```diff
@@ -152,6 +152,40 @@ def test_nested_fieldsplit_solve_parallel(W, A, b, expect):
...
+        "fieldsplit_1_ksp_view_eigenvalues": None,
+        "fieldsplit": {
+            "ksp_type": "gmres",
+            "pc_type": "jacobi",
+        },
+    }
+    problem = NonlinearVariationalProblem(F, u)
+    solver = NonlinearVariationalSolver(problem, solver_parameters=sp)
+    solver.solve()
+    assert np.allclose(solver.snes.ksp.pc.getFieldSplitSubKSP()[1].computeEigenvalues(), 1)
```
stefanozampini: You can  add this check within a monitor https://petsc.org/release/petsc4py/reference/petsc4py.PETSc.SNES.html#petsc4py.PETSc.SNES.setMonitor


```suggestion
    def mymonitor(snes, it, fnorm):
        if it == 0: return
        assert np.allclose(snes.ksp.pc.getFieldSplitSubKSP()[1].computeEigenvalues(), 1)

    solver.snes.setMonitor(mymonitor)
    solver.solve()
```

firedrake/solving_utils.py
pbrubeck: And to clarify, our implementation of fieldsplit always creates new matrices for the sub KSPs. This seems a bit wasteful for MatNest, but does seem to be more performant for a monolithic MatAIJ

stefanozampini: If you call `M.assemble()` on the outer `MATNEST`, the subblocks will be assembled. If there is a bug in PETSc, this should be in fieldsplit, not in matnest. 
Can you try with a plain PETSc4py script that reproduces your solver? It should be relatively simple to implement; you don't need all the FEM machinery. If it works, then the problem is in firedrake, otherwise is in fieldsplit

pbrubeck: Sure, are there petsc4py examples on how to setup nonlinear problems?

stefanozampini: Ah! I think I have found the problem. Can you try increasing the object state of the B matrix in case of `MAT_REUSE_MATRIX` in https://gitlab.com/petsc/petsc/-/blob/main/src/mat/impls/nest/matnest.c?ref_type=heads#L706?
Just add `PetscCall(PetscObjectStateIncrease((PetscObject)*B))` after `PetscCheck(sub == *B,...`

stefanozampini: > Sure, are there petsc4py examples on how to setup nonlinear problems?

https://gitlab.com/petsc/petsc/-/blob/main/src/binding/petsc4py/test/test_snes.py

firedrake/formmanipulation.py
```diff
@@ -180,8 +180,22 @@ def matrix(self, o):
...
+            W = bc.function_space()
+            W = W.parent or W
+
+            number = spaces.index(W)
+            V = args[number].function_space()
+            field = self.blocks[number]
+            if isinstance(bc, DirichletBC):
+                sub_bc = bc.reconstruct(field=field, V=V, g=bc.function_arg)
+            elif isinstance(bc, EquationBC):
+                raise NotImplementedError("Please get in touch if you need this")
```
JHopeCollins:
```suggestion
                raise NotImplementedError("Extracting matrix subblocks not supported with EquationBC. Please get in touch if you need this")
```

firedrake/solving_utils.py
pbrubeck: Your one-line suggestion seems to fix the issue, and we do no longer need this block, thanks

firedrake/solving_utils.py
```diff
@@ -504,6 +523,15 @@ def form_jacobian(snes, X, J, P):
...
 
+        # Bump petsc matrix state of each split by assembling them.
+        # Ensures that if the matrix changed, the preconditioner is
+        # updated if necessary.
+        for fields, splits in ctx._splits.items():
+            for subctx in splits:
+                subctx._jac.assemble()
+                if subctx.Jp is not None:
+                    subctx._pjac.assemble()
+
```
```suggestion
```

firedrake/solving_utils.py
stefanozampini: Great. I will open a PR in PETSc to fix the problem when I have some time to do it, hopefully this week

firedrake/formmanipulation.py
```diff
@@ -161,27 +160,58 @@ def cofunction(self, o):
...
+            if isinstance(bc, DirichletBC):
+                bc_temp = bc.reconstruct(field=field, V=V, g=bc.function_arg, use_split=True)
+            elif isinstance(bc, EquationBCSplit):
+                row_field, col_field = argument_indices
+                bc_temp = bc.reconstruct(field=field, V=V, row_field=row_field, col_field=col_field, use_split=True)
+                bc_temp = None
+            if bc_temp is not None:
+                bcs.append(bc_temp)
+
+        return AssembledMatrix(form or tuple(args), tuple(bcs), submat)
```
leo-collins, line 214: If we know what `form` is then this should be a `Matrix`, not an `AssembledMatrix`?

pbrubeck: maybe we should have a dispatcher mechanism `get_matrix` that decides which subclass to return

firedrake/formmanipulation.py
```diff
@@ -161,27 +160,58 @@ def cofunction(self, o):
...
+
+            number = spaces.index(W)
+            V = args[number].function_space()
+            field = self.blocks[number]
+            if isinstance(bc, DirichletBC):
+                bc_temp = bc.reconstruct(field=field, V=V, g=bc.function_arg, use_split=True)
+            elif isinstance(bc, EquationBCSplit):
+                row_field, col_field = argument_indices
+                bc_temp = bc.reconstruct(field=field, V=V, row_field=row_field, col_field=col_field, use_split=True)
+                bc_temp = None
```
pbrubeck, line 210:
```suggestion
```

PR #4712, opened 2025-11-07
Fix Hypre auxiliary spaces

pbrubeck:
# Description
Fixes Hypre on hexes by constructing the auxiliary element from scratch instead of reconstructing TensorProductElements and passing a family to the factor elements.

Adds tests on hexes

### DIFF
--- a/firedrake/preconditioners/hypre_ads.py
+++ b/firedrake/preconditioners/hypre_ads.py
@@ -46,6 +46,7 @@ class HypreADS(PCBase):
             try:
                 G = chop(assemble(interpolate(grad(TrialFunction(P1)), NC1)).petscmat)
             except NotImplementedError:
+                # dual evaluation not yet implemented see https://github.com/firedrakeproject/fiat/issues/109
                 G = tabulate_exterior_derivative(P1, NC1)
         else:
             G = G_callback(P1, NC1)
--- a/firedrake/preconditioners/hypre_ams.py
+++ b/firedrake/preconditioners/hypre_ams.py
@@ -62,6 +62,7 @@ class HypreAMS(PCBase):
             try:
                 G = chop(assemble(interpolate(grad(TrialFunction(P1)), V)).petscmat)
             except NotImplementedError:
+                # dual evaluation not yet implemented see https://github.com/firedrakeproject/fiat/issues/109
                 G = tabulate_exterior_derivative(P1, V)
         else:
             G = G_callback(P1, V)
--- a/tests/firedrake/regression/test_hypre_ads.py
+++ b/tests/firedrake/regression/test_hypre_ads.py
@@ -30,6 +30,7 @@ def test_homogeneous_field(V, mat_type, interface):
     bc = DirichletBC(V, u_exact, (1, 2, 3, 4))
 
     params = {
+        'snes_type': 'ksponly',
         'mat_type': mat_type,
         'pmat_type': 'aij',
         'ksp_type': 'cg',
--- a/tests/firedrake/regression/test_hypre_ams.py
+++ b/tests/firedrake/regression/test_hypre_ams.py
@@ -35,6 +35,7 @@ def test_homogeneous_field(V, mat_type, interface):
     bc = DirichletBC(V, constant_field, (1, 2, 3, 4))
 
     params = {
+        'snes_type': 'ksponly',
         'mat_type': mat_type,
         'pmat_type': 'aij',
         'ksp_type': 'cg',


firedrake/preconditioners/hypre_ads.py
```diff
@@ -27,16 +28,33 @@ def initialize(self, obj):
         if formdegree != 2 or degree != 1:
             raise ValueError("Hypre ADS requires lowest order RT elements! (not %s of degree %d)" % (family, degree))
 
-        P1 = V.reconstruct(family="Lagrange", degree=1)
-        NC1 = V.reconstruct(family="N1curl" if mesh.ufl_cell().is_simplex() else "NCE", degree=1)
+        # Get the auxiliary Nedelec and Lagrange spaces and the coordinate space
+        cell = V.ufl_element().ufl_cell()
+        NC1_element = FiniteElement("N1curl" if cell.is_simplex() else "NCE", cell=cell, degree=1)
+        P1_element = FiniteElement("Lagrange", cell=cell, degree=1)
+        coords_element = VectorElement(P1_element, dim=mesh.geometric_dimension())
```
JHopeCollins, line 35: Is this valid for periodic meshes where the coordinates are in DG? Why not use the element from V.mesh.coordinates?

firedrake/preconditioners/hypre_ads.py
```diff
@@ -27,16 +28,33 @@ def initialize(self, obj):
...
+        if V.shape:
+            NC1_element = TensorElement(NC1_element, shape=V.shape)
+            P1_element = TensorElement(P1_element, shape=V.shape)
```
JHopeCollins, line 38: What happens if V is a `VectorElement`? Will it get converted to a rank-1 `TensorElement`? Will that cause any problems?

firedrake/preconditioners/hypre_ads.py
```diff
@@ -27,16 +28,33 @@ def initialize(self, obj):
...
+        P1 = V.reconstruct(element=P1_element)
+        VectorP1 = V.reconstruct(element=coords_element)
+
         G_callback = appctx.get("get_gradient", None)
         if G_callback is None:
-            G = chop(assemble(interpolate(grad(TestFunction(P1)), NC1)).petscmat)
+            try:
+                G = chop(assemble(interpolate(grad(TrialFunction(P1)), NC1)).petscmat)
+            except NotImplementedError:
+                G = tabulate_exterior_derivative(P1, NC1)
```
JHopeCollins, line 50: What do you expect to raise the `NotImplementedError` in the `try` block? There's a lot of calls there.

JHopeCollins (changes requested):
> constructing the auxiliary element from scratch instead of reconstructing TensorProductElements

Why is reconstructing the `TensorProductElement` the wrong thing to do here?
In my UFL-ignorance I don't see how the new code does anything differently to the old code.

firedrake/preconditioners/hypre_ads.py
pbrubeck: The `assemble` call will raise it because dual evaluation is not implemented for HCurl on hex meshes.

firedrake/preconditioners/hypre_ads.py
pbrubeck: `V.shape` is not the same as `V.ufl_shape`. `V.shape` refers to the block shape, the number of copies of the base element, in a `Vector/TensorFunctionSpace`, which is the right thing here

firedrake/preconditioners/hypre_ads.py
pbrubeck: `V.mesh.coordinates` could be high order. We need to use `vector_P1` always, especially when coordinates are in DG, because Hypre wants the coordinates of the `P1` dofs.

firedrake/preconditioners/hypre_ads.py
JHopeCollins: Why not just use `tabulate_exterior_derivative` for all cases?

firedrake/preconditioners/hypre_ads.py
JHopeCollins: Do these PCs not work for periodic meshes then?

pbrubeck: I don't know what would hypre do on a discontinuous periodic mesh (like a `PeriodicUnitCubeMesh`). But there are cases where we might have a peridioc mesh with continuous coordinates, like when we build the torus by deforming the `PeriodicUnitCubeMesh`.

firedrake/preconditioners/hypre_ads.py
JHopeCollins: Ok, so this works because it's constructing an `Element` not a `FunctionSpace`?

pbrubeck: Before, we were reconstructing the `FunctionSpace`, and that would automatically preserve the number of copies of the base element on a `Vector/TensorFunctionSpace`.

firedrake/preconditioners/hypre_ads.py
pbrubeck: This PC works on periodic meshes with continuous coordinates discretized in DG, like the torus

JHopeCollins: Does the torus have CG coordinate field or is it still DG even though they are actually continuous?

I'm wondering if it makes sense to check the coordinate space and error/warn for discontinuous coordinates.

firedrake/preconditioners/hypre_ads.py
pbrubeck: Because eventually we would like to fix dual evaluation on Hcurl elements

pbrubeck: The more native way of doing things should be prefered here.  `tabulate_exterior_derivative` is a hack that I introduced to deal with the dual evaluation issue

JHopeCollins: Ok. Can you open an issue in the relevant repo about dual evaluation for Hcurl (if there isn't one already), and add a comment explaining why the `try-except` block is there with a link to the issue?

firedrake/preconditioners/hypre_ads.py
pbrubeck: We get the torus on hexes by deforming a periodic extruded mesh, so coordinates are in DG, but it turns out they are continuous

firedrake/preconditioners/hypre_ams.py
```diff
@@ -48,10 +49,20 @@ def initialize(self, obj):
...
+            P1_element = TensorElement(P1_element, shape=V.shape)
+        P1 = V.reconstruct(element=P1_element)
+        VectorP1 = V.reconstruct(element=coords_element)
+
         G_callback = appctx.get("get_gradient", None)
         if G_callback is None:
-            G = chop(assemble(interpolate(grad(TestFunction(P1)), V)).petscmat)
+            try:
+                G = chop(assemble(interpolate(grad(TrialFunction(P1)), V)).petscmat)
+            except NotImplementedError:
```
pbrubeck, line 64:
```suggestion
            except NotImplementedError:
                # dual evaluation not yet implemented see https://github.com/FInAT/FInAT/issues/95
```

firedrake/preconditioners/hypre_ads.py
```diff
@@ -27,16 +28,33 @@ def initialize(self, obj):
...
+        NC1 = V.reconstruct(element=NC1_element)
+        P1 = V.reconstruct(element=P1_element)
+        VectorP1 = V.reconstruct(element=coords_element)
+
         G_callback = appctx.get("get_gradient", None)
         if G_callback is None:
-            G = chop(assemble(interpolate(grad(TestFunction(P1)), NC1)).petscmat)
+            try:
+                G = chop(assemble(interpolate(grad(TrialFunction(P1)), NC1)).petscmat)
+            except NotImplementedError:
```
pbrubeck, line 48:
```suggestion
            except NotImplementedError:
                # dual evaluation not yet implemented see https://github.com/FInAT/FInAT/issues/95
```

firedrake/preconditioners/hypre_ads.py
pbrubeck: I think there's not much we can do about this. I don't think we should warn users if they have e.g. the torus mesh.  I am also not sure what happens in the truly discontinuous case. I think this discussion is not the point of the PR.

firedrake/preconditioners/hypre_ads.py
pbrubeck: done

JHopeCollins: Thanks. Looks like the finat issue was closed, should it be https://github.com/firedrakeproject/fiat/issues/109 instead?

firedrake/preconditioners/hypre_ams.py
```diff
@@ -48,10 +49,21 @@ def initialize(self, obj):
...
+                # dual evaluation not yet implemented see https://github.com/FInAT/FInAT/issues/95
```
pbrubeck:
```suggestion
                # dual evaluation not yet implemented see https://github.com/firedrakeproject/fiat/issues/109
```

firedrake/preconditioners/hypre_ads.py
```diff
@@ -27,16 +28,34 @@ def initialize(self, obj):
...
+                # dual evaluation not yet implemented see https://github.com/FInAT/FInAT/issues/95
```
pbrubeck:
```suggestion
                # dual evaluation not yet implemented see https://github.com/firedrakeproject/fiat/issues/109
```

firedrake/preconditioners/hypre_ads.py
pbrubeck: done

tests/firedrake/regression/test_hypre_ads.py
```diff
@@ -2,89 +2,46 @@
...
-              'ksp_max_it': '30',
-              'ksp_rtol': '1e-15',
-              'ksp_atol': '1e-15',
-              'pc_type': 'python',
-              'pc_python_type': 'firedrake.AssembledPC',
-              'assembled_pc_type': 'python',
-              'assembled_pc_python_type': 'firedrake.HypreADS',
-              }
+    params = {
+        'mat_type': mat_type,
```
pbrubeck, line 34:
```suggestion
        'snes_type': 'ksponly',
        'mat_type': mat_type,
```

tests/firedrake/regression/test_hypre_ams.py
```diff
@@ -3,107 +3,59 @@
...
-              'ksp_max_it': '30',
-              'ksp_rtol': '2e-15',
-              'pc_type': 'python',
-              'pc_python_type': 'firedrake.AssembledPC',
-              'assembled_pc_type': 'python',
-              'assembled_pc_python_type': 'firedrake.HypreAMS',
-              'assembled_pc_hypre_ams_zero_beta_poisson': True
-              }
+    params = {
+        'mat_type': mat_type,
```
pbrubeck, line 39:
```suggestion
        'snes_type': 'ksponly',
        'mat_type': mat_type,
```

pbrubeck:
Look at that glorious green tick!

PR #4713, opened 2025-11-07
Make VTK optional again

connorjward:
I think I was wrong to make VTK a hard dependency of Firedrake. They may publish wheels for all architectures now but they still make releases sporadically. This has meant that macOS users have had to downgrade their Python to 3.13 during install due to there not yet being a 3.14 wheel.

From a software maitainer perspective having VTK be optional is desirable because it should make it clearer to users that the reason their installation isn't working isn't our fault!

Note that VTK will still be available in the firedrake-vanilla containers.

This should go into the next patch release (and soon at that).

### DIFF
--- a/tests/firedrake/conftest.py
+++ b/tests/firedrake/conftest.py
@@ -66,9 +66,9 @@ def _skip_test_dependency(dependency):
             del ngsPETSc
             return not skip
         except ImportError:
-           return skip
+            return skip
 
-     elif dependency == "vtk":
+    elif dependency == "vtk":
         try:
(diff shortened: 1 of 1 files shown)


JHopeCollins:
> @JHopeCollins are you happy for us to go ahead and merge this? This is a minor policy/strategy change but I think if we're both happy with it we can just merge it.

Yes happy with this. It causes problems on a fairly regular basis and it isn't a hard dependency for running Firedrake so it makes sense not to have it as a hard dependency for installing.

PR #4714, opened 2025-11-10
fix `Function._ad_assign_numpy` for real space in parallel

### LINKED ISSUE #4667: BUG: scipy optimization broken for real function control in parallel
Consider
```
from firedrake import *
from firedrake.adjoint import ReducedFunctional, minimize, Control, continue_annotation

continue_annotation()
mesh = UnitSquareMesh(1, 1)
m = Function(R)
J = assemble((m-1)**2*dx)
rf = ReducedFunctional(J, Control(m))
mopt = minimize(rf)
print(float(mopt))
```
This runs correctly in serial but breaks in parallel (release) with the following backtrace:
```
  File "/home/skramer/tst/cpjordan/tst.py", line 10, in <module>
    mopt = minimize(rf)
  File "/usr/lib/python3.13/contextlib.py", line 85, in inner
    return func(*args, **kwds)
  File "/home/skramer/fd2510/lib/python3.13/site-packages/pyadjoint/optimization/optimization.py", line 251, in minimize
    opt = algorithm(rf_np, **kwargs)
  File "/home/skramer/fd2510/lib/python3.13/site-packages/pyadjoint/optimization/optimization.py", line 131, in minimize_scipy_generic
    res = scipy_minimize(J, m_global, method=method, **kwargs)
  File "/home/skramer/fd2510/lib/python3.13/site-packages/scipy/optimize/_minimize.py", line 784, in minimize
    res = _minimize_lbfgsb(fun, x0, args, jac, bounds,
                           callback=callback, **options)
  File "/home/skramer/fd2510/lib/python3.13/site-packages/scipy/optimize/_lbfgsb_py.py", line 413, in _minimize_lbfgsb
    sf = _prepare_scalar_function(fun, x0, jac=jac, args=args, epsilon=eps,
                                  bounds=bounds,
                                  finite_diff_rel_step=finite_diff_rel_step,
                                  workers=workers)
  File "/home/skramer/fd2510/lib/python3.13/site-packages/scipy/optimize/_optimize.py", line 310, in _prepare_scalar_function
    sf = ScalarFunction(fun, x0, args, grad, hess,
                        finite_diff_rel_step, bounds, epsilon=epsilon,
                        workers=workers)
  File "/home/skramer/fd2510/lib/python3.13/site-packages/scipy/optimize/_differentiable_functions.py", line 274, in __init__
    self._update_fun()
    ~~~~~~~~~~~~~~~~^^
  File "/home/skramer/fd2510/lib/python3.13/site-packages/scipy/optimize/_differentiable_functions.py", line 353, in _update_fun
    fx = self._wrapped_fun(self.x)
  File "/home/skramer/fd2510/lib/python3.13/site-packages/scipy/_lib/_util.py", line 590, in __call__
    fx = self.f(np.copy(x), *self.args)
  File "/home/skramer/fd2510/lib/python3.13/site-packages/pyadjoint/reduced_functional_numpy.py", line 40, in __call__
    return self.rf.__call__(self.set_local(m_copies, m_array))
                            ~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^
  File "/home/skramer/fd2510/lib/python3.13/site-packages/pyadjoint/reduced_functional_numpy.py", line 45, in set_local
    m[i], offset = control.assign_numpy(m[i], m_array, offset)
                   ~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/skramer/fd2510/lib/python3.13/site-packages/pyadjoint/control.py", line 93, in assign_numpy
    return self.control._ad_assign_numpy(dst, src, offset)
           ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^
  File "/home/skramer/fd2510/src/firedrake/firedrake/adjoint_utils/function.py", line 281, in _ad_assign_numpy
    dst.dat.data_wo[...] = m_a_local.reshape(dst.dat.data_wo.shape)
                           ~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^
ValueError: cannot reshape array of size 0 into shape (1,)
```

This breaks optimizing with constant coefficients - for which AFAIU we should use functions in R -  in parallel with scipy, which is used quite a bit in Thetis (unfortunately we were not actually testing any of our adjoint setups in parallel)

JHopeCollins:
For the `"Real"` space the `Dat` stores a duplicate copy of the single unknown on every rank. However when using `scipy.optimize` with the adjoint, the numpy array returned from scipy only has an element on rank 0, and is empty on every other rank.
This PR fixes a bug in the logic in `Function._ad_assign_numpy` that didn't account for this difference and broke with the `"Real"` space in parallel.

fixes #4667

### DIFF
--- a/tests/firedrake/adjoint/test_reduced_functional.py
+++ b/tests/firedrake/adjoint/test_reduced_functional.py
@@ -2,6 +2,7 @@ import pytest
 
 from firedrake import *
 from firedrake.adjoint import *
+from pytest_mpi.parallel_assert import parallel_assert
 
 from numpy.random import rand
 
@@ -256,6 +257,9 @@ def test_interpolate_mixed():
 @pytest.mark.skipcomplex
 @pytest.mark.parallel(2)
 def test_real_space_assign_numpy():
+    """Check that Function._ad_assign_numpy correctly handles
+    zero length arrays on some ranks for Real space in parallel.
+    """
     mesh = UnitSquareMesh(1, 1)
     R = FunctionSpace(mesh, "R", 0)
     dst = Function(R)
@@ -263,16 +267,19 @@ def test_real_space_assign_numpy():
     data = 1 + np.arange(src.shape[0])
     src[:] = data
     dst._ad_assign_numpy(dst, src, offset=0)
-    assert np.allclose(dst.dat.data_ro, data)
+    parallel_assert(np.allclose(dst.dat.data_ro, data))
 
 
 @pytest.mark.skipcomplex
 @pytest.mark.parallel(2)
 def test_real_space_parallel():
+    """Check that scipy.optimize works for Real space in parallel
+    despite dat.data array having zero length on some ranks.
+    """
     mesh = UnitSquareMesh(1, 1)
     R = FunctionSpace(mesh, "R", 0)
     m = Function(R)
     J = assemble((m-1)**2*dx)
     Jhat = ReducedFunctional(J, Control(m))
     opt = minimize(Jhat)
-    assert np.allclose(opt.dat.data_ro, 1)
+    parallel_assert(np.allclose(opt.dat.data_ro, 1))


firedrake/adjoint_utils/function.py
```diff
@@ -278,7 +278,12 @@ def _ad_dot(self, other, options=None):
     def _ad_assign_numpy(dst, src, offset):
         range_begin, range_end = dst.dat.dataset.layout_vec.getOwnershipRange()
         m_a_local = src[offset + range_begin:offset + range_end]
-        dst.dat.data_wo[...] = m_a_local.reshape(dst.dat.data_wo.shape)
+        if dst.function_space().ufl_element().family() == "Real":
+            # Real space keeps a redundant copy of the data on every rank
+            comm = dst.function_space().mesh().comm
```
JHopeCollins: @connorjward is this the right `comm` or should I be using `mesh()._comm`?

tests/firedrake/adjoint/test_reduced_functional.py
```diff
@@ -251,3 +251,14 @@ def test_interpolate_mixed():
...
+
+@pytest.mark.skipcomplex
+@pytest.mark.parallel(2)
+def test_real_space_parallel():
+    mesh = UnitSquareMesh(1, 1)
+    R = FunctionSpace(mesh, "R", 0)
+    m = Function(R)
+    J = assemble((m-1)**2*dx)
+    Jhat = ReducedFunctional(J, Control(m))
+    minimize(Jhat)
```
connorjward: Taylor test?

firedrake/adjoint_utils/function.py
connorjward: Probably `_comm`. The whole point is to use an internal communicator for our own routines so regardless of what `comm` is, `_comm` shouldn't be touched by a user.

Aside: in pyop3 I prefer `user_comm` and `internal_comm` to avoid this confusion.
```suggestion
            comm = dst.function_space().mesh()._comm
```

connorjward (changes requested):
Wouldn't it be nice if Real spaces worked just like regular functions...

tests/firedrake/adjoint/test_reduced_functional.py
JHopeCollins: The purpose of `_ad_assign_numpy` is to gather/scatter data so that `scipy.optimize` will DTRT in parallel. so the Taylor test never actually needs to call `_ad_assign_numpy`.

connorjward: So just asserting that `minimize` converges in parallel is a sufficient test? In general I like to avoid tests that just check that things run rather than that we get the correct answer.

JHopeCollins: Fair enough. The `_ad_assign_numpy` test is enough for the current bug, but really the minimize test should check the result to make sure we don't introduce any other errors down the line.
I've added an assertion to do this.

tests/firedrake/adjoint/test_reduced_functional.py
```diff
@@ -251,3 +251,28 @@ def test_interpolate_mixed():
...
+@pytest.mark.parallel(2)
+def test_real_space_assign_numpy():
+    mesh = UnitSquareMesh(1, 1)
+    R = FunctionSpace(mesh, "R", 0)
+    dst = Function(R)
+    src = dst.dat.dataset.layout_vec.array_r.copy()
+    data = 1 + np.arange(src.shape[0])
+    src[:] = data
+    dst._ad_assign_numpy(dst, src, offset=0)
+    assert np.allclose(dst.dat.data_ro, data)
```
connorjward: `parallel_assert`?

tests/firedrake/adjoint/test_reduced_functional.py
```diff
@@ -251,3 +251,28 @@ def test_interpolate_mixed():
...
+@pytest.mark.skipcomplex
+@pytest.mark.parallel(2)
+def test_real_space_parallel():
+    mesh = UnitSquareMesh(1, 1)
+    R = FunctionSpace(mesh, "R", 0)
+    m = Function(R)
+    J = assemble((m-1)**2*dx)
+    Jhat = ReducedFunctional(J, Control(m))
+    opt = minimize(Jhat)
+    assert np.allclose(opt.dat.data_ro, 1)
```
connorjward: ditto

connorjward (changes requested):
Otherwise this seems to be totally fine.

tests/firedrake/adjoint/test_reduced_functional.py
JHopeCollins: Done

PR #4716, opened 2025-11-10
White noise generators and AR covariance operators

JHopeCollins:
Depends on #4773 and https://github.com/firedrakeproject/fiat/pull/201.

White and correlated noise generation from #3799, plus extra methods to calculate action and inverse of autoregressive covariance operators and use it for a weighted norm.

### White noise generation

See [Croci et al 2018](https://doi.org/10.1137/18M1175239).

White noise samples calculating $M^{1/2}z$ where $z_{i}\sim\mathcal{N}(0, I)$ and $M$ is the mass matrix.
$M^{1/2}$ can be constructed by assembling the Cholesky factor of the element-wise mass matrices.

There are two implementations of this:

1. hijack the pyop2 and loopy assemble kernels to return the cholesky factor of the element mass matrix rather than the actual mass matrix. This is the implementation from #3799 
2. Build a discontinuous superspace and use PETSc to calculate the global cholesky factor of that mass matrix, which is block diagonal over the elements. This uses the `L2Cholesky` class from #4575. There is also a specialisation of this method for `VertexOnlyMesh`.

### Autoregressive covariance operator

See [Mirouze & Weaver, 2010](https://doi.org/10.1002/qj.643)

A covariance operator $B$ with an `m`-th autoregressive function kernel can be calculated using `m` Backward Euler steps of a diffusion operator, where the diffusion coefficient is specified by the correlation lengthscale.

If $M$ is the mass matrix, $K$ is the matrix for a single Backward Euler step, and $\lambda$ is a normalisation factor, then the m-th order covariance operator with variance $\sigma^2$ is:

```math
B: V^{*} \to V = \sigma\lambda((K^{-1}M)^{m}M^{-1})\lambda\sigma
```

This formulation leads to an efficient implementation for $B^{-1}$ (using the action of $K$) and for $B^{1/2}$ by taking only m/2 steps of the diffusion operator. This can be used to calculate weighted norms $||x||_{B^{-1}}$ and sample from $\mathcal{N}(0,B)$.

The `CovarianceOperatorBase` class provides an abstract interface for other covariance operators, which must implement:

1. `sample` to return $w\in\mathcal{N}(0,B)$
2. `apply_action(x)` to apply $Bx$
3. `apply_inverse(x)` to apply $^{-1}x$
4. `norm(x)` to return $||x||_{B^{-1}} = x^{T}B^{-1}x$

### PETSc python contexts

If you use the weighted norm $||x||_{B^{-1}}$ in the functional for an adjoint calculation, then often $B$ is a good preconditioner for the 2nd order Hessian. This PR provides a python Mat context and corresponding python PC context for children of `CovarianceOperatorBase`.

### DIFF
--- a/tests/firedrake/adjoint/test_covariance_operator.py
+++ b/tests/firedrake/adjoint/test_covariance_operator.py
@@ -362,6 +362,8 @@ def test_covariance_mat(m, family, operation):
 @pytest.mark.skipcomplex
 @pytest.mark.parametrize("family", ("CG", "DG"))
 def test_diffusion_form(family):
+    """Test that the provided diffusion forms converge to a known solution at the expected rate.
+    """
     from firedrake.adjoint.covariance_operator import diffusion_form
 
     def poisson_error(mesh, family):


tests/firedrake/adjoint/test_covariance_operator.py
```diff
@@ -0,0 +1,325 @@
...
+@pytest.mark.parametrize("dim", (0, 1, 2), ids=["scalar", "vec1", "vec2"])
+@pytest.mark.parametrize("family", ("CG", "DG"))
+@pytest.mark.parametrize("mesh_type", ("interval", "square"))
+@pytest.mark.parametrize("backend", ("pyop2", "petsc"))
+def test_white_noise(family, degree, mesh_type, dim, backend):
+    """Test that white noise generator converges to a mass matrix covariance.
+    """
+    if backend == "petsc" and COMM_WORLD.size > 1:
+        pytest.skip(
+            "petsc backend for noise generation not implemented in parallel.")
```
JHopeCollins:
```suggestion
```

firedrake/adjoint/covariance_operator.py
```diff
@@ -0,0 +1,1201 @@
...
+
+        Returns
+        -------
+        Function | Cofunction :
+            The white noise sample in :math:`V`
+        """
+        raise NotImplementedError
+
+    @cached_property
+    def broken_space(self):
```
dham, line 103: move to FunctionSpace

firedrake/adjoint/covariance_operator.py
```diff
@@ -0,0 +1,1201 @@
...
+        """
+        The discontinuous superspace containing :math:`V`, ``self.function_space``.
+        """
+        element = self.function_space.ufl_element()
+        mesh = self.function_space.mesh().unique()
+        if isinstance(element, VectorElement):
+            dim = element.num_sub_elements
+            scalar_element = element.sub_elements[0]
+            broken_element = BrokenElement(scalar_element)
```
dham: extend BrokenElement so it works in the Firedrake way for VFS and friends.

firedrake/adjoint/covariance_operator.py
```diff
@@ -0,0 +1,1201 @@
...
+        return self._rng
+
+    @cached_property
+    def riesz_map(self):
+        """A :class:`~firedrake.cofunction.RieszMap` to cache the solver
+        for :meth:`~firedrake.cofunction.Cofunction.riesz_representation`.
+        """
+        return RieszMap(self.function_space, constant_jacobian=True)
+        """
+        Generate a white noise sample.
```
dham: Why docstring here?

firedrake/adjoint/covariance_operator.py
```diff
@@ -0,0 +1,1201 @@
...
+        IID sample on the broken function space.
+        """
+        return self._rng
+
+    @cached_property
+    def riesz_map(self):
+        """A :class:`~firedrake.cofunction.RieszMap` to cache the solver
+        for :meth:`~firedrake.cofunction.Cofunction.riesz_representation`.
+        """
+        return RieszMap(self.function_space, constant_jacobian=True)
```
dham: hard code L2

firedrake/adjoint/covariance_operator.py
```diff
@@ -0,0 +1,1201 @@
...
+    def __init__(self, V: WithGeometry, rng=None,
+                 seed: int | None = None):
+        super().__init__(V, rng=rng, seed=seed)
+        self.cholesky = L2Cholesky(V)
+        self._zb = Function(V)
+        self.M = inner(self._zb, TestFunction(V))*dx
+
+    def sample(self, *, rng=None,
+               tensor: Function | Cofunction | None = None,
+               apply_riesz: bool = False):
```
dham, line 321: remove all the exciting ways you are multiplying by I.

firedrake/adjoint/covariance_operator.py
```diff
@@ -0,0 +1,1201 @@
...
+            The :class:`~firedrake.function.Function` to take the norm of.
+
+        Returns
+        -------
+        pyadjoint.AdjFloat :
+            The norm of ``x``.
+        """
+        return self.apply_inverse(x)(x)
+
+    @abc.abstractmethod
```
dham: probably don't need to be abstract.

firedrake/adjoint/covariance_operator.py
```diff
@@ -0,0 +1,1201 @@
...
+
+        for i in range(self.iterations):
+            self._urhs.assign(self._u)
+            self.solver.solve()
+
+        return tensor.assign(self._weight*self._u)
+
+
+def diffusion_form(u, v, kappa: Constant | Function,
+                   formulation: AutoregressiveCovariance.DiffusionForm):
```
dham, line 855: unit test this function.

dham (changes requested):

firedrake/adjoint/covariance_operator.py
JHopeCollins: https://github.com/firedrakeproject/fiat/pull/201

firedrake/adjoint/covariance_operator.py
JHopeCollins: https://github.com/firedrakeproject/firedrake/pull/4773

pyproject.toml
```diff
@@ -23,7 +23,7 @@ dependencies = [
   # TODO RELEASE
   "fenics-ufl @ git+https://github.com/FEniCS/ufl.git@main",
   # TODO RELEASE
-  "firedrake-fiat @ git+https://github.com/firedrakeproject/fiat.git@main",
+  "firedrake-fiat @ git+https://github.com/firedrakeproject/fiat.git@JHopeCollins/mixed-broken-element",
```
JHopeCollins:
```suggestion
  "firedrake-fiat @ git+https://github.com/firedrakeproject/fiat.git@main",
```

JHopeCollins:
@pbrubeck the only outstanding comments were to add an MMS test for the diffusion forms, which I've done here: https://github.com/firedrakeproject/firedrake/pull/4716/changes#diff-2327a64dc67ea8e9dd667c73fc315e5ca301b26fa9fcdaa08de4c6e94200fdb4R364

firedrake/adjoint/covariance_operator.py
```diff
@@ -0,0 +1,1179 @@
...
+                LinearVariationalProblem(K, self._Mrhs, self._u, bcs=bcs,
+                                         constant_jacobian=True),
+                solver_parameters=solver_parameters,
+                options_prefix=options_prefix)
+
+            self.mass_solver = LinearVariationalSolver(
+                LinearVariationalProblem(M, self._Krhs, self._u, bcs=bcs,
+                                         constant_jacobian=True),
+                solver_parameters=mass_parameters,
+                options_prefix=mass_prefix)
```
pbrubeck, line 773: I think we could use `LinearSolver` instead here, up to you.

JHopeCollins: Why would that be better?

tests/firedrake/adjoint/test_covariance_operator.py
```diff
@@ -0,0 +1,397 @@
...
+    for i in range(nsamples[-1]):
+        with generator.sample().dat.vec_ro as bv:
+            samples[:, i] = petsc2numpy_vec(bv)
+
+    covariances = [np.cov(samples[:, :ns]) for ns in nsamples]
+
+    # Covariance matrix should converge at a rate of sqrt(n)
+    errors = [np.linalg.norm(cov-covmat) for cov in covariances]
+    normalised_errors = [err*sqrt(n) for err, n in zip(errors, nsamples)]
+    normalised_errors /= normalised_errors[-1]
```
pbrubeck: Should we be looking a the rate of convergence?
```
rate = -numpy.diff(numpy.log(errors)) / numpy.diff(numpy.log(nsamples))
```

firedrake/adjoint/covariance_operator.py
pbrubeck: It is just style. That would remove the need of the variational problem, and setting `constant_jacobian=True`. But you might need to explicitly assemble the rhs (if this is the case, we might consider automating that as a separate PR)

JHopeCollins: I think I'm ok with it staying as is for now. Hiding LVP is nice, but I think it's more important on the user side than internally.

tests/firedrake/adjoint/test_covariance_operator.py
JHopeCollins: Yes, you're right. I've updated this test and the equivalent check in the VOM noise generation test.

tests/firedrake/adjoint/test_covariance_operator.py
```diff
@@ -0,0 +1,397 @@
...
+
+    base_nx = 16
+    nrefs = 3
+    base_mesh = UnitSquareMesh(base_nx, base_nx)
+    mh = MeshHierarchy(base_mesh, nrefs)
+    errors = [poisson_error(m, family) for m in mh]
+
+    # second order convergence
+    for i in range(nrefs):
+        assert (errors[i]/errors[i+1])**0.5 > 1.9
```
pbrubeck: Here as well?

JHopeCollins: Yep, done

JHopeCollins:
Only failures are unrelated timeouts.

PR #4717, opened 2025-11-10
Fix the bug in implementing the AuxiliaryOperatorPC and add a note in…

Molin-Han:
Fixed the demo in using the AuxiliaryOperatorPC to solve the mixed Poisson equation. 
Added a note indicating the feature of AuxiliaryOperatorPC.

### DIFF
--- a/demos/saddle_point_pc/saddle_point_systems.py.rst
+++ b/demos/saddle_point_pc/saddle_point_systems.py.rst
@@ -449,7 +449,7 @@ we then use to solve the problem, we can provide one ourselves.
 Recall that :math:`S` is spectrally a Laplacian only in a
 discontinuous space.  A natural choice is therefore to use an interior
 penalty DG formulation for the Laplacian term on the block of the scalar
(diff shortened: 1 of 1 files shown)


demos/saddle_point_pc/saddle_point_systems.py.rst
```diff
@@ -449,10 +449,10 @@ we then use to solve the problem, we can provide one ourselves.
 Recall that :math:`S` is spectrally a Laplacian only in a
 discontinuous space.  A natural choice is therefore to use an interior
 penalty DG formulation for the Laplacian term on the block of the scalar
-variable. We can provide it as an :class:`~.AuxiliaryOperatorPC` via a python preconditioner. ::
+variable. We can provide it as an :class:`~.AuxiliaryOperatorPC` via a python preconditioner. Note that the ```form``` method in ```AuxiliaryOperatorPC``` takes the trial functions as the first argument and the test functions as the second argument, which is the reverse of the usual convention. ::
```
JHopeCollins: Isn't it test first then trial?

Molin-Han: Yes, I have made the change now.

JHopeCollins (approved):
Great, thanks for fixing this!

PR #4720, opened 2025-11-11
Use `PETSc.ScalarType(0)` in `Function.zero`

JHopeCollins:
When the tape is re-run it will complain if you use `Function.assign(0)` because it doesn't know how to handle an `int`. In reality it just converts it to a float and everything is fine, but we're raising unnecessary warnings whenever someone uses `Function.zero` while taping.


pbrubeck (dismissed):
LGTM

PR #4722, opened 2025-11-13
Remove star imports from `__init__.py`

leo-collins:
Explicit imports in `firedrake/__init__.py`. Fixes namespace leakage problem https://github.com/firedrakeproject/firedrake/issues/4696.

Things that are no longer in the firedrake namespace (not exhaustive): 
- `np`
- `stop_annotating`
- `ufl`

We now explicitly import `PETSc` into the global namespace.

G-adopt tests are failing because they rely on these leaked names somewhere

### DIFF
--- a/firedrake/__init__.py
+++ b/firedrake/__init__.py
@@ -104,6 +104,7 @@ from firedrake.mg import (  # noqa: F401
 )
 from firedrake.norms import errornorm, norm  # noqa: F401
 from firedrake.nullspace import VectorSpaceBasis, MixedVectorSpaceBasis  # noqa: F401
+from firedrake.output import VTKFile  # noqa: F401
 from firedrake.parameters import (  # noqa: F401
     Parameters, parameters, disable_performance_optimisations
 )


firedrake/__init__.py
```diff
@@ -56,41 +56,110 @@ def init_petsc():
...
+    NonlinearVariationalProblem, NonlinearVariationalSolver
+)
+from firedrake.eigensolver import (  # noqa: F401
+    LinearEigenproblem, LinearEigensolver
+)
+from firedrake.ensemble import (  # noqa: F401
+    Ensemble, EnsembleFunction, EnsembleCofunction,
+    EnsembleFunctionSpace, EnsembleDualSpace
+)
+from firedrake.randomfunctiongen import *  # noqa: F401
```
connorjward, line 155: Missed this one?

firedrake/__init__.py
```diff
@@ -100,10 +169,10 @@ def init_petsc():
 # Once `File` is deprecated update the above line removing `File` and add
 #   from firedrake._deprecation import output
 #   sys.modules["firedrake.output"] = output
-from firedrake.output import *
+from firedrake.output import *  # noqa: F401
```
connorjward: Ditto

firedrake/__init__.py
```diff
@@ -100,10 +169,10 @@ def init_petsc():
...
 import sys
 sys.modules["firedrake.plot"] = plot
-from firedrake.plot import *
+from firedrake.plot import *  # noqa: F401
```
connorjward, line 179: Ditto

firedrake/utility_meshes.py
```diff
@@ -30,9 +30,13 @@
     real
 )
 from firedrake.cython import dmcommon
-from firedrake import mesh
```
connorjward, line 33: Oh amazing. I always hated this.

connorjward (changes requested):
Looks great! Thanks for taking this on.

It seems like a fair number of tests are still failing. Also when we merge this we should announce it.

firedrake/__init__.py
leo-collins: In there we have `__all__ = [name for name, _ in inspect.getmembers(randomgen, inspect.isclass)]` where `import numpy.random as randomgen`. I'm not sure what the best way to do this properly is

connorjward: Yeah that file is a bit crazy. You can ignore that one.

firedrake/__init__.py
leo-collins: We could probably remove `firedrake.output` and `firedrake.plot` from the global namespace now? It was moved to `firedrake.pyplot` around two years ago: https://github.com/firedrakeproject/firedrake/pull/3117. Also https://github.com/firedrakeproject/firedrake/pull/3324

connorjward: Yeah good idea

PR #4728, opened 2025-11-18
SameMeshInterpolator: fix for MeshSequence

pbrubeck:
# Description

### DIFF
--- a/firedrake/interpolation.py
+++ b/firedrake/interpolation.py
@@ -318,6 +318,7 @@ def get_interpolator(expr: Interpolate) -> Interpolator:
         target_mesh = target_mesh.unique()
         source_mesh = source_mesh.unique()
     except ValueError:
+        # intercept MeshSequence
         return MixedInterpolator(expr)
 
     submesh_interp_implemented = (
@@ -621,7 +622,8 @@ class SameMeshInterpolator(Interpolator):
         return f
 
     def _get_callable(self, tensor=None, bcs=None):
-        if tensor and isinstance(self.dual_arg, Cofunction) and self.dual_arg.dat == tensor.dat:
+        if (isinstance(tensor, Cofunction) and isinstance(self.dual_arg, Cofunction)
+                and set(tensor.dat).intersection(self.dual_arg.dat)):
             f = self._get_tensor()
             copyout = (partial(f.dat.copy, tensor.dat),)
         else:


firedrake/interpolation.py
```diff
@@ -312,6 +312,10 @@ def get_interpolator(expr: Interpolate) -> Interpolator:
 
     operand, = expr.ufl_operands
     target_mesh = expr.target_space.mesh()
+    try:
+        target_mesh = target_mesh.unique()
```
leo-collins, line 318: Is this really the correct way to do things now?

connorjward: I agree. This seems wrong. What are you trying to do here?

pbrubeck: My guess is that we could find a better way of expressing this logic. We could instead count the number of unique target meshes and unique source meshes, if any of those numbers is greater than one, we should call `MixedInterpolator` immediatly. If there's only one unique source mesh and one unique target mesh we should call `unique()` on them and proceed with the rest of the dispatch logic.

connorjward: Why not
```py
if any(isinstance(mesh, MeshSequence) for mesh in [source_mesh, target_mesh]):
  ...
```

pbrubeck: We really want to collapse a `MeshSequence` of the same mesh to the unique mesh.

pbrubeck: See here https://github.com/firedrakeproject/firedrake/pull/4595/files#r2537113177

connorjward: Maybe we need `mesh.is_unique` as an attribute to make this nice.

pbrubeck: The main issue is that in line 321 we are comparing the meshes
```
        if target_mesh is source_mesh ... :
```

pbrubeck: But also in the lines above where we have the submesh logic, we should be doing that on the `unique()` meshes.

pbrubeck: Only `MixedInterpolator` should deal with `MeshSequence`.

pbrubeck: A `MeshSequence` that can be collapsed into a `unique()` Mesh can be handled by `SameMeshInterpolator` or `CrossMeshInterpolator`.

firedrake/interpolation.py
```diff
@@ -618,6 +622,8 @@ def _get_callable(self, tensor=None, bcs=None):
         op2_tensor = f if isinstance(f, op2.Mat) else f.dat
 
         loops = []
+        if self.access == op2.INC:
```
leo-collins, line 635: This will break in-place adjoint interpolation, e.g. `assemble(interpolate(TestFunction(V), cofunc), tensor=cofunc)`

pbrubeck: How does that even work if we delay the `zero()` inside the for-loop?

pbrubeck: Isn't this a no-op? If we have interpolation into the same space we could call `assign` instead of `interpolate`.

pbrubeck: The right fix to deal with self-interpolation is to create a tempory buffer for the output. We already do this for the forward interpolate via `copy_in` / `copy_out`. I guess we need to support this for the adjoint .

firedrake/interpolation.py
connorjward: Clearly this requires some thought.

firedrake/interpolation.py
```diff
@@ -313,6 +313,13 @@ def get_interpolator(expr: Interpolate) -> Interpolator:
     operand, = expr.ufl_operands
     target_mesh = expr.target_space.mesh()
     source_mesh = extract_unique_domain(operand) or target_mesh
+
+    try:
+        target_mesh = target_mesh.unique()
+        source_mesh = source_mesh.unique()
+    except ValueError:
+        return MixedInterpolator(expr)
```
leo-collins, line 322: I think this logic needs a comment explaining what's happening here

firedrake/interpolation.py
pbrubeck: How were we getting around this issue before the refactoring?

pbrubeck: Did we just ignore the `tensor` kwarg for adjoint interpolation?

leo-collins: I think so. Currently we have https://github.com/firedrakeproject/firedrake/blob/69e3cb7a959a3a93d2f77951ae5898bb536fe212/firedrake/assemble.py#L622-L626

which passes the tensor to `Interpolator` in the forward case, but not in the adjoint case.

pbrubeck: The fix I added here is a bit ugly, but not so invasive. To me it is quite clear that for `INC` access we should be creating the buffer and zeroing at this level (outside the for loop). Nevertheless, we could instead keep the `zero()` call inside the for-loop, but only call it the first time we add on each `sub_op2_tensor`. I'll leave this up to you.

PR #4729, opened 2025-11-18
add matrix magic methods

leo-collins:


firedrake/matrix.py
```diff
@@ -108,18 +112,59 @@ def __str__(self):
...
+        if isinstance(other, MatrixBase):
+            return ufl.Action(other, self)
+        elif isinstance(other, Cofunction):
+            return ufl.Action(ufl.Adjoint(self), other)
+        else:
+            return NotImplemented
+
+    def __mul__(self, other):
+        # Scalar multiplication
+        if isinstance(other, Complex | Constant):
```
connorjward, line 145:
```suggestion
        if isinstance(other, (Complex, Constant)):
```
I am pretty sure that that isn't valid syntax. Happy to be proven wrong.

leo-collins: It's valid since 3.10: https://docs.python.org/3/library/functions.html#isinstance

connorjward: Awesome!

firedrake/matrix.py
```diff
@@ -108,18 +112,59 @@ def __str__(self):
...
         if isinstance(other, MatrixBase):
-            mat = self.petscmat - other.petscmat
-            return AssembledMatrix(self.arguments(), (), mat)
+            if self.arguments() != other.arguments():
+                raise ValueError("Arguments in matrix subtraction must match.")
+            return ufl.FormSum((self, 1.), (other, -1.))
         else:
             return NotImplemented
 
+    def __matmul__(self, other):
```
pbrubeck, line 129: If this class inherits from `ufl.Matrix`, and `ufl.Matrix` is a `BaseForm`, shouldn't it already inherit addition and scalar multiplication?

I'm not aware of matrix multiplication being implemented in `BaseForm`, but it seems that it would not harm to have it there.

pbrubeck: Matrix multiplication is there, but it seems that it only supports multiplication times a `Function` (and breaks for `Cofunction` and `Matrix`).

https://github.com/FEniCS/ufl/blob/main/ufl/form.py#L247

https://github.com/FEniCS/ufl/blob/main/ufl/form.py#L219-L225

firedrake/matrix.py
```diff
@@ -108,18 +112,59 @@ def __str__(self):
...
+        # Scalar multiplication from the left
+        if isinstance(other, Complex | Constant):
+            return ufl.FormSum((self, other))
+        else:
+            return NotImplemented
+
+    def __truediv__(self, other):
+        # Scalar division
+        if isinstance(other, Complex | Constant):
+            other = other.values().item() if isinstance(other, Constant) else other
```
pbrubeck, line 160: This is a bit dangerous, we are silently freezing the `Constant`. This returns a symbolic expression that no longer makes reference to the `Constant`, which is inconsistent with how scalar multiplication is implemented.

pbrubeck: It might be better to leave this one for a separate PR where we enable `BaseFormAssembler` to numerically evaluate constant-valued symbolic expressions in the weights of a `FormSum`.

pbrubeck: BTW, `FormSum` should be able to deal with non-float weights, so I `1/Constant` should be fine

Issue #4721, opened 2025-11-12
BUG: Matfree EquationBC with Schur fails


KarsKnook:
When using `EquationBC` in a matrix-free context with Schur complement preconditioning, we do not get the correct behaviour (potentially the `EquationBC` is not enforced correctly?). For example, the problem in the MFE below should converge in 1 FGMRES iteration but it takes 2 FGMRES iterations. When commenting out `"mat_type": "matfree",` we get the expected behaviour.

```
from firedrake import *

mesh = UnitSquareMesh(2, 2)
nml = FacetNormal(mesh)

V = FunctionSpace(mesh, "RT", 1)
W = FunctionSpace(mesh, "DG", 0)
Z = V * W

z = Function(Z)
sigma, u = split(z)
tau, v = TestFunctions(Z)

F = inner(sigma, tau) * dx
F -= inner(u, div(tau)) * dx
F += inner(div(sigma), v) * dx
F += inner(div(sigma), div(tau)) * dx

bcs = [EquationBC(inner(dot(sigma - Constant([1,1]), nml), dot(tau, nml)) * ds(1) == 0, z, [1], V=Z.sub(0))]

sp = {
        "mat_type": "matfree",
        "snes_monitor": None,
        "ksp_monitor": None,
        "ksp_type": "fgmres",
        "pc_type": "fieldsplit",
        "pc_fieldsplit_type": "schur",
        "pc_fieldsplit_schur_fact_type": "full",
        "fieldsplit_0": {
            "ksp_type": "preonly",
            "pc_type": "python",
            "pc_python_type": "firedrake.AssembledPC",
            "assembled": {
                "pc_type": "lu",
                "pc_factor_mat_solver_type": "mumps",
                },
            },
        "fieldsplit_1": {
            "ksp_type": "gmres",
            "ksp_monitor": None,
            "ksp_atol": 0,
            "ksp_rtol": 1e-12,
            "pc_type": "python",
            "pc_python_type": "firedrake.MassInvPC",
            "aux_pc_type": "python",
            "aux_pc_python_type": "firedrake.AssembledPC",
            "aux_assembled": {
                "pc_type": "lu",
                "pc_factor_mat_solver_type": "mumps",
                },
            },
        }

solve(F==0, z, bcs=bcs, solver_parameters=sp)
```

PR #4734, opened 2025-11-19
BUG: Fix EquationBC + matfree + fieldsplit

pbrubeck:
# Description
Fixes https://github.com/firedrakeproject/firedrake/issues/4721

### DIFF
--- a/firedrake/assemble.py
+++ b/firedrake/assemble.py
@@ -1248,6 +1248,9 @@ class OneFormAssembler(ParloopFormAssembler):
                 bc.apply(r, u=u)
         elif isinstance(bc, EquationBCSplit):
             bc.zero(tensor)
+            if isinstance(bc.f, ufl.ZeroBaseForm) or bc.f.empty():
+                # form is empty, do nothing
+                return
             OneFormAssembler(bc.f, bcs=bc.bcs,
                              form_compiler_parameters=self._form_compiler_params,
                              needs_zeroing=False,
--- a/firedrake/bcs.py
+++ b/firedrake/bcs.py
@@ -641,9 +641,6 @@ class EquationBCSplit(BCBase):
             rank = len(self.f.arguments())
             splitter = ExtractSubBlock()
             form = splitter.split(self.f, argument_indices=(row_field, col_field)[:rank])
-            if isinstance(form, ufl.ZeroBaseForm) or form.empty():
-                # form is empty, do nothing
-                return
             if u is not None:
                 form = firedrake.replace(form, {self.u: u})
         if action_x is not None:
--- a/firedrake/matrix_free/operators.py
+++ b/firedrake/matrix_free/operators.py
@@ -397,11 +397,6 @@ class ImplicitMatrixContext(object):
                 bc_temp = bc.reconstruct(field=row_inds, V=Wrow, g=bc.function_arg, sub_domain=bc.sub_domain, use_split=True)
             elif isinstance(bc, EquationBCSplit):
                 bc_temp = bc.reconstruct(field=row_inds, V=Wrow, row_field=row_inds, col_field=col_inds, use_split=True)
-                if bc_temp is None:
-                    # The split logic in EquationBC fails on the off-diagonal when the BC does not couple the fields.
-                    # The entire row needs to be zeroed out in this case. As a workaround, split a homogeneous DirichletBC.
-                    bc = DirichletBC(bc.function_space(), 0, bc.sub_domain)
-                    bc_temp = bc.reconstruct(field=row_inds, V=Wrow, g=bc.function_arg, sub_domain=bc.sub_domain, use_split=True)
             if bc_temp is not None:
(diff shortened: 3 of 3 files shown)


tests/firedrake/equation_bcs/test_equation_bcs.py
```diff
@@ -174,8 +174,8 @@ def linear_poisson_mixed(solver_parameters, mesh_num, porder):
     u1 = cos(2 * pi * y) / 2
     n = FacetNormal(mesh)
 
-    a = (inner(sigma, tau) + inner(u, div(tau)) + inner(div(sigma), v)) * dx
-    L = inner(u1, dot(tau, n)) * ds(1) + inner(f, v) * dx
+    a = inner(sigma, tau)*dx + inner(u, div(tau))*dx + inner(div(sigma), v + div(tau))*dx
+    L = inner(u1, dot(tau, n)) * ds(1) + inner(f, v + div(tau)) * dx
```
JHopeCollins, line 178: Why is this change necessary?

tests/firedrake/equation_bcs/test_equation_bcs.py
```diff
@@ -331,24 +331,24 @@ def test_EquationBC_mixedpoisson_matfree_fieldsplit():
...
                          'pc_fieldsplit_type': 'schur',
                          'pc_fieldsplit_schur_fact_type': 'full',
-                         'fieldsplit_0_ksp_type': 'cg',
+                         'fieldsplit_0_ksp_type': 'preonly',
                          'fieldsplit_0_pc_type': 'python',
                          'fieldsplit_0_pc_python_type': 'firedrake.AssembledPC',
                          'fieldsplit_0_assembled_pc_type': 'lu',
-                         'fieldsplit_1_ksp_type': 'cg',
-                         'fieldsplit_1_pc_use_amat': True,
+                         'fieldsplit_0_assembled_pc_factor_mat_solver_type': 'mumps',
```
JHopeCollins, line 345: `'fieldsplit_0_assembled': DEFAULT_DIRECT_SOLVER_PARAMETERS`?

tests/firedrake/equation_bcs/test_equation_bcs.py
```diff
@@ -331,24 +331,24 @@ def test_EquationBC_mixedpoisson_matfree_fieldsplit():
...
+                         'fieldsplit_1_ksp_type': 'gmres',
                          'fieldsplit_1_pc_type': 'python',
                          'fieldsplit_1_pc_python_type': 'firedrake.MassInvPC',
-                         'fieldsplit_1_Mp_pc_type': 'icc'}
+                         'fieldsplit_1_Mp_pc_type': 'lu',
+                         'fieldsplit_1_Mp_pc_factor_mat_solver_type': 'mumps',
```
JHopeCollins, line 350: `'fieldsplit_1_Mp': DEFAULT_DIRECT_SOLVER_PARAMETERS`?

tests/firedrake/equation_bcs/test_equation_bcs.py
pbrubeck: The Schur complement solver needs the `(div(sigma), div(tau))` term in order for the Schur complement to be equivalent to the mass matrix.

pbrubeck: What I did was just to test the original weak form against `v+div(tau)` instead of just `v`

tests/firedrake/equation_bcs/test_equation_bcs.py
pbrubeck: I don't think so
```
In [4]: DEFAULT_DIRECT_SOLVER_PARAMETERS
Out[4]: mappingproxy({'mat_solver_type': 'mumps', 'mat_mumps_icntl_14': 200})
```

connorjward: I thought we could have nested dictionaries?

tests/firedrake/equation_bcs/test_equation_bcs.py
JHopeCollins: Ok, sounds like the right thing to do if it turns the PC into a direct solve and means that you can set `-ksp_type preonly`.

tests/firedrake/equation_bcs/test_equation_bcs.py
pbrubeck: The issue is that this dictionary is missing the `pc_factor` bit in the prefix and pc_type lu option

pbrubeck: These dictionaries aren't intuitive, I rather not use them. If you don't like to see mumps there, we can remove that option and let PETSc use its own default

JHopeCollins: That's my bad, I thought they had the full prefixes. No need to change then.

PR #4747, opened 2025-11-25
BUG: use native prolongation for KMV elements

pbrubeck:
# Description

### DIFF
--- a/tests/firedrake/multigrid/test_grid_transfer.py
+++ b/tests/firedrake/multigrid/test_grid_transfer.py
@@ -66,10 +66,7 @@ def shape(request):
 
 @pytest.fixture(params=["injection", "restriction", "prolongation"])
 def transfer_type(request, hierarchy):
-    if not hierarchy.nested and request.param == "injection":
-        return pytest.mark.xfail(reason="Supermesh projections not implemented yet")(request.param)
-    else:
-        return request.param
+    return request.param
 
 
 @pytest.fixture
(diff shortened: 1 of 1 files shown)


firedrake/mg/embedded.py
```diff
@@ -10,7 +10,7 @@
 __all__ = ("TransferManager", )
 
 
-native_families = frozenset(["Lagrange", "Discontinuous Lagrange", "Real", "Q", "DQ", "BrokenElement", "Crouzeix-Raviart"])
+native_families = frozenset(["Lagrange", "Discontinuous Lagrange", "Real", "Q", "DQ", "BrokenElement", "Crouzeix-Raviart", "Kong-Mulder-Veldhuizen"])
```
connorjward, line 13: Is there any way that we can detect this family instead of enumerating them?

pbrubeck: We are inspecting UFL elements, so we need to compare strings.

In theory we could detect elements that only have point evaluation dofs if we were inspecting the finat element instead.

connorjward: That seems like it would really help to minimise technical debt. For instance if I had to fix this bug with KMV elements it would have taken a me a long time to understand what needed doing. Could you open an issue describing that? Then I'm happy to approve.

connorjward: I see https://github.com/firedrakeproject/firedrake/issues/4750. Thanks!

PR #4754, opened 2025-11-27
Remove bizarre cofunction init behaviour

connorjward:
If passed another cofunction as input do not write back to it.

### DIFF
--- a/firedrake/cofunction.py
+++ b/firedrake/cofunction.py
@@ -81,9 +81,6 @@ class Cofunction(ufl.Cofunction, CofunctionMixin):
         else:
             self.dat = function_space.make_dat(val, dtype, self.name())
 
-        if isinstance(function_space, Cofunction):
-            self.dat.copy(function_space.dat)
-
     @PETSc.Log.EventDecorator()
     def copy(self, deepcopy=True):
         r"""Return a copy of this :class:`firedrake.function.CoordinatelessFunction`.


firedrake/cofunction.py
```diff
@@ -81,9 +81,6 @@ def __init__(self, function_space, val=None, name=None, dtype=ScalarType,
         else:
             self.dat = function_space.make_dat(val, dtype, self.name())
 
-        if isinstance(function_space, Cofunction):
-            self.dat.copy(function_space.dat)
-
```
pbrubeck, line 86:
```suggestion
        if isinstance(function_space, Cofunction): 
            function_space.dat.copy(self.dat) 
 
```

pbrubeck: I think `Function(Function)` copies the data into `self`?

pbrubeck: We actually want to `assign` here

https://github.com/firedrakeproject/firedrake/blob/main/firedrake%2Ffunction.py#L282-L283

pbrubeck, line 86:
```suggestion
        if isinstance(function_space, Cofunction): 
            self.assign(function_space)

```

connorjward: But we do it here: https://github.com/firedrakeproject/firedrake/blob/main/firedrake/cofunction.py#L59

pbrubeck: Is this adjoint-safe? Probably not? Assigning seems more consistent with what we have for Function.

pbrubeck: The constructor decorator might indicate that this is annotated? I really don't know how they work, and when we are allowed to modify the `dat`

connorjward: The previous approach had nothing to do with the adjoint either. And it's not actually a problem. We handle the adjoint bits and pieces on https://github.com/firedrakeproject/firedrake/blob/main/firedrake/cofunction.py#L59 which calls the adjoint copy (https://github.com/firedrakeproject/firedrake/blob/main/firedrake/adjoint_utils/function.py#L77).

dham (approved):

PR #4756, opened 2025-11-27
Remove deprecated File alias for VTKFile

connorjward:
We have been warning about this for 18 months.


JHopeCollins (approved):
Breaking API change, so should be announced on slack once merged.

PR #4761, opened 2025-12-01
Submesh on subcommunicator

pbrubeck:
# Description

Enables `Submesh(mesh, comm=COMM_SELF)`

Depends on https://gitlab.com/petsc/petsc/-/merge_requests/8876

### DIFF
--- a/firedrake/cython/dmcommon.pyx
+++ b/firedrake/cython/dmcommon.pyx
@@ -3906,7 +3906,7 @@ def submesh_correct_entity_classes(PETSc.DM dm,
         DMLabel lbl_core, lbl_owned, lbl_ghost
         PetscBool has
 
-    if dm.comm.size == 1:
+    if subdm.comm.size == 1:
         return
     CHKERR(DMPlexGetChart(dm.dm, &pStart, &pEnd))
     CHKERR(DMPlexGetChart(subdm.dm, &subpStart, &subpEnd))


firedrake/__init__.py
```diff
@@ -3,7 +3,7 @@
 # the specific version, here we are more permissive. This is to catch the
 # case where users don't update their PETSc for a really long time or
 # accidentally install a too-new release that isn't yet supported.
-PETSC_SUPPORTED_VERSIONS = ">=3.23.0"
+PETSC_SUPPORTED_VERSIONS = ">=3.24.1"
```
connorjward: This isn't sufficient. This change is going into PETSc main.

firedrake/mesh.py
```diff
@@ -4815,14 +4821,24 @@ def Submesh(mesh, subdim, subdomain_id, label_name=None, name=None):
         elif subdim == dim - 1:
             label_name = dmcommon.FACE_SETS_LABEL
     name = name or _generate_default_submesh_name(mesh.name)
-    subplex = dmcommon.submesh_create(plex, subdim, label_name, subdomain_id, False)
+    if comm is None or comm == mesh.comm:
+        subplex = dmcommon.submesh_create(plex, subdim, label_name, subdomain_id, ignore_halo, comm)
+    else:
+        # FIXME
```
JHopeCollins: Does this still need fixing?

firedrake/__init__.py
connorjward: I think you will have to do
```py
# TODO RELEASE
# PETSC_SUPPORTED_VERSIONS = ">=3.25"
```

firedrake/mesh.py
pbrubeck: It might just need some assertions, as I don't think we fully support creating a submesh on a different comm and labels

firedrake/mesh.py
JHopeCollins: `None` is passed here in the test. Can you update the docstring explaining what happens in this case?
```suggestion
   subdomain_id : int | None
       Subdomain ID representing the submesh.
```

firedrake/mesh.py
JHopeCollins: As in you can either pass a comm that isn't mesh.comm, or you can pass in labels?
Sounds like it just needs some exceptions at the beginning and a note in the docstring then.

setup.py
```diff
@@ -20,7 +20,7 @@
 
 
 # Ensure that the PETSc getting linked against is compatible
-petsctools.init(version_spec=">=3.23.0")
+petsctools.init(version_spec=">=3.24.2")
```
connorjward:
```suggestion
# TODO RELEASE
# petsctools.init(version_spec=">=3.25")
petsctools.init()
```

firedrake/__init__.py
pbrubeck: Somehow the single source of truth test is finding two different petsc versions?

firedrake/__init__.py
```diff
@@ -3,7 +3,10 @@
 # the specific version, here we are more permissive. This is to catch the
 # case where users don't update their PETSc for a really long time or
 # accidentally install a too-new release that isn't yet supported.
-PETSC_SUPPORTED_VERSIONS = ">=3.23.0"
+PETSC_SUPPORTED_VERSIONS = ">=3.24.2"
+
+# TODO RELEASE
+# PETSC_SUPPORTED_VERSIONS = ">=3.25"
```
pbrubeck:
```suggestion
# TODO RELEASE set to ">=3.25"
```

setup.py
connorjward: Changing this doesn't achieve anything. Just leave as is and put a TODO RELEASE comment

firedrake/mesh.py
```diff
@@ -4743,12 +4743,19 @@ def Submesh(mesh, subdim, subdomain_id, label_name=None, name=None):
...
         Name of the label to search ``subdomain_id`` in.
-    name : str
+    name : str |  None
         Name of the submesh.
+    ignore_halo : bool
+        Whether to exclude the halo from the submesh.
+    reorder : bool
+        Whether to reorder the mesh entities.
+    comm : PETSc.Comm | None
+        An optional sub-communicator to define the submesh.
```
connorjward, line 4758: Should explain what the default is

connorjward (changes requested):

firedrake/mesh.py
pbrubeck, line 4758:
```suggestion
        An optional sub-communicator to define the submesh.
        By default, the submesh is defined on `mesh.comm`.
```

firedrake/mesh.py
```diff
@@ -4810,22 +4817,36 @@ def Submesh(mesh, subdim, subdomain_id, label_name=None, name=None):
...
+    if subdomain_id is None:
+        # Filter the plex with PETSc's default label (cells owned by comm)
+        if label_name is not None:
+            raise ValueError("subdomain_id == None requires label_name == None.")
+        if subdim != dim:
+            raise NotImplementedError(f"Found submesh dim ({subdim}) and parent dim ({dim})")
+        subplex, _ = plex.filter(sanitizeSubMesh=True, ignoreHalo=ignore_halo, comm=comm)
+    else:
+        if comm is not None and comm != mesh.comm:
+            raise NotImplementedError("Submesh on subcommunicator not implemented on cell subsets.")
```
```suggestion
            raise NotImplementedError("Submesh on sub-communicator not implemented on cell subsets.")
```

firedrake/mesh.py
pbrubeck: done

firedrake/__init__.py
```diff
@@ -3,7 +3,9 @@
 # the specific version, here we are more permissive. This is to catch the
 # case where users don't update their PETSc for a really long time or
 # accidentally install a too-new release that isn't yet supported.
-PETSC_SUPPORTED_VERSIONS = ">=3.23.0"
+PETSC_SUPPORTED_VERSIONS = ">=3.24.2"
+
+# TODO RELEASE set to ">=3.25"
```
connorjward:
```suggestion
# TODO RELEASE set to ">=3.25"
PETSC_SUPPORTED_VERSIONS = ">=3.23.0"

```

setup.py
pbrubeck: What should the comment say? Why should we keep this lower than 3.24.2 if we know that we won't be able to compile cython?
```suggestion
# TODO RELEASE
petsctools.init(version_spec=">=3.23.0")
```

firedrake/mesh.py
```diff
@@ -4810,22 +4818,37 @@ def Submesh(mesh, subdim, subdomain_id, label_name=None, name=None):
...
-        elif subdim == dim - 1:
-            label_name = dmcommon.FACE_SETS_LABEL
+    if subdomain_id is None:
+        # Filter the plex with PETSc's default label (cells owned by comm)
+        if label_name is not None:
+            raise ValueError("subdomain_id == None requires label_name == None.")
+        if subdim != dim:
+            raise NotImplementedError(f"Found submesh dim ({subdim}) and parent dim ({dim})")
+        subplex, _ = plex.filter(sanitizeSubMesh=True, ignoreHalo=ignore_halo, comm=comm)
+        dmcommon.submesh_update_facet_labels(plex, subplex)
```
connorjward, line 4833: Why don't we just use `dmcommon.submesh_create` here?

setup.py
connorjward:
```suggestion
# TODO RELEASE set to ">=3.25"
petsctools.init(version_spec=">=3.23.0")
```

connorjward: 3.24.2 is also going to be broken

firedrake/mesh.py
pbrubeck: Ideally we would be doint that, right now we get segfaults

pbrubeck: The issues are within `submesh_correct_entity_classes`, I am still debugging.

firedrake/__init__.py
```diff
@@ -3,9 +3,11 @@
 # the specific version, here we are more permissive. This is to catch the
 # case where users don't update their PETSc for a really long time or
 # accidentally install a too-new release that isn't yet supported.
+# TODO RELEASE set to ">=3.25"
 PETSC_SUPPORTED_VERSIONS = ">=3.23.0"
 
 
+
```
pbrubeck:
```suggestion
```

firedrake/mesh.py
pbrubeck: This issue goes away by not correcting pyop2 entity classes when the sub-communicator has size=1, see the change in dmcommon.pyx

pbrubeck: Now we need to have two branches because `dmcommon.submesh_create`:
1. creates the temporary label, which we don't need for this case
2. requires `subdomain_id` to be an `int` (we cannot pass `subdomain_id=None`)

connorjward (approved):

PR #4763, opened 2025-12-02
multi-domain assembly

leo-collins:
allow assembly of uncoupled multi-domain problems

### DIFF
--- a/tests/firedrake/regression/test_multiple_domains.py
+++ b/tests/firedrake/regression/test_multiple_domains.py
@@ -2,7 +2,6 @@ from firedrake import *
 import pytest
 import numpy as np
 from functools import partial
-from ufl.algorithms.ad import expand_derivatives
 
 
 @pytest.fixture(params=["interval", "tri", "quad", "tet"])


firedrake/assemble.py
```diff
@@ -1084,15 +1084,10 @@ def local_kernels(self):
...
-            raise NotImplementedError("All integration domains must share a mesh topology")
-
+        topology = set(d.topology.submesh_ancesters[-1] for d in self._form.ufl_domains())
         for o in itertools.chain(self._form.arguments(), self._form.coefficients()):
-            domains = extract_domains(o)
-            for domain in domains:
-                if domain is not None and domain.topology.submesh_ancesters[-1] != topology:
+            for domain in extract_domains(o):
+                if domain is not None and domain.topology.submesh_ancesters[-1] not in topology:
                     raise NotImplementedError("Assembly with multiple meshes is not supported")
```
connorjward: Well this error message is no longer true!

tests/firedrake/regression/test_multiple_domains.py
```diff
@@ -123,3 +119,47 @@ def test_two_form(mesh1, mesh2, form, expect):
...
+    V1 = FunctionSpace(mesh1, "Q", 3)
+    V2 = FunctionSpace(mesh2, "CG", 2)
+    V = V1 * V2
+
+    u1, u2 = TrialFunctions(V)
+    v1, v2 = TestFunctions(V)
+
+    a = (
+        inner(grad(u1), grad(v1))*dx(domain=mesh1)
+        + inner(grad(u2), grad(v2))*dx(domain=mesh2)
```
connorjward, line 137: I assume if we had a form with `u1` and `v2` we would crash in a helpful way?

firedrake/assemble.py
connorjward: I wonder if we should just remove this logic. Here we are thinking in terms of the entire form and not the individual integrands. Checking that we have a compatible topology *somewhere* in the form is not a very strong guarantee. For example it would not catch `u1*v2` in the tests below.

connorjward (changes requested):
I wonder if all this logic should simply be

firedrake/assemble.py
leo-collins: I moved this so it checks per integral now

tests/firedrake/regression/test_multiple_domains.py
leo-collins: I've added a test for this case

connorjward:
> Don't merge, it isn't right yet

Please leave this as a draft PR then

tests/firedrake/regression/test_multiple_domains.py
```diff
@@ -123,3 +119,90 @@ def test_two_form(mesh1, mesh2, form, expect):
...
+    )
+
+    u_exact_expr1 = sin(pi * x1) * sin(pi * y1)
+    u_exact_expr2 = x2 * y2 * (1 - x2) * (1 - y2)
+    f1 = expand_derivatives(-div(grad(u_exact_expr1)))
+    f2 = expand_derivatives(-div(grad(u_exact_expr2)))
```
dham:
```suggestion
    f1 = -div(grad(u_exact_expr1))
    f2 = -div(grad(u_exact_expr2))
```

UZerbinati:
> I agree. @UZerbinati it looks like your merges into ngsPETSc have broken things. I will try to fix them when I move things over (imminently).

I don't see what I've broken, all Firedrake test's passed in ngsPETSc before merging.

connorjward:
I fixed this for ngsPETSc in https://github.com/NGSolve/ngsPETSc/pull/94, before you made more changes.

It's fine. Release still works which is the most important thing. I'll sort this all out tomorrow hopefully.

tests/firedrake/regression/test_multiple_domains.py
```diff
@@ -123,3 +118,90 @@ def test_two_form(mesh1, mesh2, form, expect):
...
+        if i == 1:
+            with pytest.raises(NotImplementedError):
+                assemble(a)
+            continue
+        A = assemble(a)
+        assert A.M.values.shape == (V.dim(), V.dim())
+
+    a = inner(u[0], v[0])*dx(domain=mesh1) + inner(u[0], v[1])*dx(domain=mesh2)
+    with pytest.raises(NotImplementedError):
+        assemble(a)
```
ksagiyam, line 203: Most of the integrals here are illegal as `intersect_measures` is not set. I suppose these tests are about if meshA and meshB are related in the sense of submesh if they appear in a single integral expression, and not about legality of the integral expression, so you first need to write legal integrals, and test if they still raise exception when participating meshes are unrelated. Given that you removed the lines in `assemble.py`, I imagine that exception is raised when composing maps.

Note that illegality of the integrals was overlooked due to https://github.com/firedrakeproject/firedrake/pull/4775.

tsfc/driver.py
```diff
@@ -145,6 +145,11 @@ def compile_integral(integral_data, form_data, prefix, parameters, *, diagonal=F
     domain_integral_type_map.update(dict.fromkeys(coefficient_meshes, "cell"))
     domain_integral_type_map.update(integral_data.domain_integral_type_map)
 
+    for arg in arguments:
+        if domain_integral_type_map[extract_unique_domain(arg)] is None:
+            raise NotImplementedError("Assembly of forms over unrelated meshes is not supported. "
+                                      "Try using Submeshes or cross-mesh interpolation.")
+
```
ksagiyam, line 152: This is checking legality of the integral, so the error message looks incorrect. Given a legal integral, topological check must happen in `firedrake` (not in `tsfc`). As mentioned in the above, exception should be raised when composing maps, but this needs to be checked.

tests/firedrake/regression/test_multiple_domains.py
dham: I think this is something different. These are topologically unrelated. These are just testing that we are getting the indexing right.

ksagiyam: All this PR does is removing the guard in `assemble.py`, so we should be testing the consequence of that. That uncoupled mixed domain problems work is tested in `test_multi_domain_solve()`, which is fine, but here we should definitely be testing that other forms that would have hit the removed guard still fail correctly, too. I expected that that is what `test_multi_domain_assemble()` should be doing, and the forms that would have hit the removed guard are those symbolically correct ones defined with `intersect_measures` that contain unrelated topologies. Those are what the removed guard was for at least.

The illegal forms currently tested in `test_multi_domain_assemble()` are useful, but I think the above is more relevant to the specific changes made in this PR. In those forms, `integral_type`s are not defined for some participating meshes, so they must cause exception earlier at more fundamental level (TSFC) if the topology guard in `assemble.py` exists or not. `integral_type` not defined on some mesh is a broader problem.
I don't remember the details, but those illegal forms must have caused some exception in TSFC as expected before https://github.com/firedrakeproject/firedrake/pull/4775 was merged, but https://github.com/firedrakeproject/firedrake/pull/4775 partially broke it, and this PR added some guard in TSFC (for `Argument`s) on top of that. So the change made in TSFC in this PR might not be necessary once https://github.com/firedrakeproject/firedrake/pull/4775 is reverted (except that the exception raised in TSFC could probably be clearer).

PR #4607, opened 2025-09-30
FunctionSpace: pass kwargs to finat.ufl.FiniteElement

pbrubeck:
# Description
This enables `variant` and `quad_scheme` to be passed on to finat direclty from the `FunctionSpace` constructor.

I don't think that tests for `quad_scheme` are necessary at this point, as we already have tests for `variant`.

PR #4764, opened 2025-12-04
BUG: Fix symmetry in supermeshing

pbrubeck:
# Description
Enables supermesh projection for symmetric tensor-valued spaces.

Enables cross-mesh interpolation  for symmetric tensor-valued spaces.

### DIFF
--- a/firedrake/interpolation.py
+++ b/firedrake/interpolation.py
@@ -473,6 +473,7 @@ class CrossMeshInterpolator(Interpolator):
         if bcs:
             raise NotImplementedError("bcs not implemented")
 
+        # TODO check V.finat_element.is_lagrange() once https://github.com/firedrakeproject/fiat/pull/200 is released
         target_element = V.ufl_element()
         if not ((isinstance(target_element, finat.ufl.MixedElement)
                  and all(sub.mapping() == "identity" for sub in target_element.sub_elements))


firedrake/supermeshing.py
```diff
@@ -20,15 +20,17 @@
...
 from pyop2.utils import get_petsc_dir
+from collections import defaultdict
 
 
 __all__ = ["assemble_mixed_mass_matrix", "intersection_finder"]
 
 
 class BlockMatrix(object):
-    def __init__(self, mat, dimension):
+    def __init__(self, mat, dimension, symmetry=None):
```
connorjward: A docstring explaining what `symmetry` actually is and what it does would be great.

pbrubeck: I renamed this kwarg to something much clearer

pbrubeck: Ideally we should be using a [KAIJ Mat](https://petsc.org/release/manualpages/Mat/MATKAIJ/) but we don't have python wrappers for this Mat type.

connorjward: Why can't we use a KAIJ Mat? This code isn't going through PyOP2

pbrubeck: Because we cannot create it from python

connorjward: Ahh, a petsc4py deficiency. Maybe leave a TODO comment?

firedrake/interpolation.py
```diff
@@ -472,7 +472,11 @@ def __init__(
             raise NotImplementedError("freeze_expr not implemented")
         if bcs:
             raise NotImplementedError("bcs not implemented")
-        if V.ufl_element().mapping() != "identity":
+
+        target_element = V.ufl_element()
+        if not ((isinstance(target_element, finat.ufl.MixedElement)
+                 and all(sub.mapping() == "identity" for sub in target_element.sub_elements))
+                or target_element.mapping() == "identity"):
```
connorjward, line 480: Didn't you recently add a utility function for checking for only point evaluation nodes? Might that not be better here?

firedrake/interpolation.py
```diff
@@ -551,7 +555,8 @@ def __init__(
         elif len(shape) == 1:
             fs_type = partial(firedrake.VectorFunctionSpace, dim=shape[0])
         else:
-            fs_type = partial(firedrake.TensorFunctionSpace, shape=shape)
+            symmetry = V_dest.ufl_element().symmetry()
+            fs_type = partial(firedrake.TensorFunctionSpace, shape=shape, symmetry=symmetry)
         P0DG_vom = fs_type(self.vom_dest_node_coords_in_src_mesh, "DG", 0)
```
connorjward, line 561: Instead of `fs_type` do we want `V.reconstruct(mesh=new_mesh, ...)` here instead?

firedrake/supermeshing.py
```diff
@@ -116,15 +114,39 @@ def likely(cell_A):
...
+        symmetry = V_A.ufl_element().symmetry()
+        assert V_B.ufl_element().mapping() == "symmetries"
+        assert V_B.ufl_element().symmetry() == symmetry
+
+        multiplicity = defaultdict(int)
+        for idx in numpy.ndindex(V_A.value_shape):
+            idx = symmetry.get(idx, idx)
+            multiplicity[idx] += 1
+
+        block_scale = tuple(scale for idx, scale in multiplicity)
```
pbrubeck:
```suggestion
        block_scale = tuple(scale for idx, scale in multiplicity.items())
```

firedrake/supermeshing.py
```diff
@@ -20,15 +20,17 @@
 from pyop2.compilation import load
 from pyop2.mpi import COMM_SELF
 from pyop2.utils import get_petsc_dir
+from collections import defaultdict
 
 
 __all__ = ["assemble_mixed_mass_matrix", "intersection_finder"]
 
 
 class BlockMatrix(object):
```
pbrubeck, line 30:
```suggestion
# TODO replace with KAIJ (we require petsc4py wrappers)
class BlockMatrix(object):
```

firedrake/interpolation.py
pbrubeck: Yes, ideally we want to be doing that. But the spaces we get here in `release` are only UFL spaces without the `finat_element` attribute. The `main` branch version of this code uses `WithGeometry` spaces.

pbrubeck: Also, the PR hasn't been merged https://github.com/firedrakeproject/fiat/pull/200

firedrake/interpolation.py
pbrubeck: We need to change the element to DG0 on the vertex-only mesh.

pbrubeck: We could probably treat Tensor and Vector elements with the same treatment. But this PR should only fix the bug.

connorjward: If it's easy enough to make the code more readable then I think we should do it.

firedrake/interpolation.py
connorjward: > But the spaces we get here in release are only UFL spaces without the finat_element attribute. The main branch version of this code uses WithGeometry spaces.

How come? I didn't realise that was a difference between main and release.

firedrake/interpolation.py
pbrubeck: I don't think we can improve it by calling reconstruct. The logic will seem more complicated

pbrubeck: Almost every attribute of the element changes, and there are vector-valued elements that do not come from a VectorElement (e.g. RT)

firedrake/interpolation.py
pbrubeck: Somehow we were handing `ufl.FunctionSpace` in release. The entire file is different, and it is not really clear to me how to fix it in release.

firedrake/interpolation.py
connorjward: But if we use `reconstruct` then we will always preserve non-specified attributes like quadrature scheme (assuming https://github.com/firedrakeproject/firedrake/pull/4607) and `variant`.

firedrake/interpolation.py
connorjward: I'm happy for this to just be a TODO comment that references "once https://github.com/firedrakeproject/fiat/pull/200 is released"

firedrake/interpolation.py
pbrubeck: Preserving all those attributes makes no sense for this case. There is only one way to specify DG0 on a 0-dimensional cell.

connorjward: But in principle we could introduce such an attribute. To me `reconstruct` seems cleaner and more robust to future changes. Maybe we need another opinion to act as tiebreaker...

> and there are vector-valued elements that do not come from a VectorElement (e.g. RT)

Does this mean that we would have to special-case `reconstruct` for RT? If so then `reconstruct` isn't the right thing.

pbrubeck: If we `reconstruct` an RT space with `family="DG", degree=0`, we get a scalar space, but we want a vector-valued space.

firedrake/interpolation.py
pbrubeck: It is not only depending on `is_lagrange`, but on the larger refactoring of interpolation.py that has been happening in main.

firedrake/interpolation.py
```diff
@@ -472,7 +472,11 @@ def __init__(
             raise NotImplementedError("freeze_expr not implemented")
         if bcs:
             raise NotImplementedError("bcs not implemented")
-        if V.ufl_element().mapping() != "identity":
+
+        target_element = V.ufl_element()
```
pbrubeck, line 477:
```suggestion
        # TODO check V.finat_element.is_lagrange() once https://github.com/firedrakeproject/fiat/pull/200 is released
        target_element = V.ufl_element()
```

firedrake/interpolation.py
pbrubeck: > I'm happy for this to just be a TODO comment that references "once [firedrakeproject/fiat#200](https://github.com/firedrakeproject/fiat/pull/200) is released"

Done

firedrake/interpolation.py
connorjward: I see. You've convinced me.

connorjward (approved):

Issue #4709, opened 2025-11-05
BUG: Cross-mesh interpolation uses the wrong communicators for PETSc objects


connorjward:
**Describe the bug**

As reported in https://github.com/firedrakeproject/firedrake/issues/4702 we are leaking memory with cross-mesh interpolation because PETSc objects are being given the public communicator instead of the private one and we do not clear it automatically.

**Steps to Reproduce**
```py
import gc

from firedrake import *
from mpi4py import MPI
from petsc4py import PETSc
from pyop2.mpi import temp_internal_comm


PYOP2_COMM_WORLD = COMM_WORLD


mesh1 = UnitSquareMesh(2, 2)
mesh2 = UnitSquareMesh(5, 5, quadrilateral=True)

x1, y1 = SpatialCoordinate(mesh1)
V2 = FunctionSpace(mesh2, "P", 1)

# case 1 (the bug)
# assemble(interpolate(x1*y1, V2))

# case 2 (expected behaviour)
assemble(TestFunction(V2)*dx)

gc.collect()

print(PYOP2_COMM_WORLD.name)  # prints 'PYOP2_COMM_WORLD'
PETSc.garbage_view(PYOP2_COMM_WORLD)  # prints 'UNKNOWN_COMM' - has entries for case 1 but not case 2 (this should not happen)

with temp_internal_comm(PYOP2_COMM_WORLD) as OTHER_COMM:
    PETSc.garbage_view(OTHER_COMM)  # prints 'UNKNOWN_COMM' - has some entries for both cases (this is expected)
```

In the script above we would expect all of the objects to be destroyed to live on `OTHER_COMM`, but that is not the case. We also would expect `garbage_view` to print the name of the communicator as 'PYOP2_COMM_WORLD' but instead it prints 'UNKNOWN_COMM'.

If this issue is encountered by anyone else the current 'fix' is to run `PETSc.garbage_cleanup(firedrake.COMM_WORLD)`.

PR #4766, opened 2025-12-05
Redesign how we interface with communicators and PETSc

connorjward:
Fixes https://github.com/firedrakeproject/firedrake/issues/4709

### Motivation
* These changes will make it much harder to accidentally stash PETSc objects on the wrong communicators.
* It avoids a situation where there were 2 PETSc communicators holding objects to destroy (one dup'ed from MPI_COMM_WORLD and the other dup-ed from PYOP2_COMM_WORLD).

### Core changes
* firedrake.COMM_WORLD is now *exactly* MPI.COMM_WORLD, not a duplicate.
* Objects do not have user and internal comm attributes. If Firedrake/PyOP2 needs to send a message we should use the `temp_internal_comm` method.
* Added a test.

#### The good

* When PETSc destroys an object in parallel it actually puts it in a bin **associated with a particular comm**.
* The comm that the bin lives on is obtained from `PetscCommDuplicate(comm)`.
* We empty this bin when we call `PETSc.garbage_cleanup(comm)`, passing the top-level comm.

#### The ugly

* Until now Firedrake redefined its own public `COMM_WORLD` ([here](https://github.com/firedrakeproject/firedrake/blob/main/pyop2/mpi.py#L65))
```py
# These are user-level communicators, we never send any messages on
# them inside PyOP2.
COMM_WORLD = PETSc.COMM_WORLD.tompi4py().Dup()
COMM_WORLD.Set_name("PYOP2_COMM_WORLD")
```
* There are now **2 different user facing COMM_WORLDs** in circulation: `MPI.COMM_WORLD` and `firedrake.COMM_WORLD`.
* This means that there are **2 different PETSc communicators that we can put garbage in**
* We don't clean both bins, so we leak objects on the other.

#### Demo

```py
from firedrake import *
from mpi4py import MPI
from petsc4py import PETSc


vec1 = PETSc.Vec().create(MPI.COMM_WORLD)
vec2 = PETSc.Vec().create(COMM_WORLD)

del vec1
del vec2

PETSc.garbage_view(MPI.COMM_WORLD)
PETSc.garbage_view(COMM_WORLD)
```

```
$ mpiexec -n 2 python mfe.py
PETSc garbage on PETSC_COMM_WORLD
Rank 0:: Total entries: 1
| Key   | Type                   | Name                             | Object ID |
|-------|------------------------|----------------------------------|-----------|
|     0 | Vec                    | Vector                           |      1    |
Rank 1:: Total entries: 1
| Key   | Type                   | Name                             | Object ID |
|-------|------------------------|----------------------------------|-----------|
|     0 | Vec                    | Vector                           |      1    |
PETSc garbage on UNKNOWN_COMM
Rank 0:: Total entries: 1
| Key   | Type                   | Name                             | Object ID |
|-------|------------------------|----------------------------------|-----------|
|     0 | Vec                    | Vector                           |      2    |
Rank 1:: Total entries: 1
| Key   | Type                   | Name                             | Object ID |
|-------|------------------------|----------------------------------|-----------|
|     0 | Vec                    | Vector                           |      2    |

```

### DIFF
--- a/.github/workflows/core.yml
+++ b/.github/workflows/core.yml
@@ -322,8 +322,7 @@ jobs:
           if [ ${{ inputs.target_branch }} = 'release' ]; then
             GUSTO_BRANCH='main'
           else
-            # UNDO ME
-            GUSTO_BRANCH='connorjward/comm-fix'
+            GUSTO_BRANCH='future'
           fi
           git clone --depth 1 https://github.com/firedrakeproject/gusto.git gusto-repo --branch $GUSTO_BRANCH
           pip install --verbose ./gusto-repo


firedrake/ensemble/ensemble.py
```diff
@@ -177,8 +175,9 @@ def ireduce(self, f, f_reduced, op=MPI.SUM, root=0):
         """
         self._check_function(f, f_reduced)
 
-        return [self._ensemble_comm.Ireduce(fdat.data_ro, rdat.data, op=op, root=root)
-                for fdat, rdat in zip(f.dat, f_reduced.dat)]
+        with temp_internal_comm(self.ensemble_comm) as icomm:
+            return [icomm.Ireduce(fdat.data_ro, rdat.data, op=op, root=root)
+                    for fdat, rdat in zip(f.dat, f_reduced.dat)]
```
JHopeCollins: How does `temp_internal_comm` interact with nonblocking communications? If the temporary comm gets destroyed are the requests still valid?

firedrake/ensemble/ensemble_functionspace.py
```diff
@@ -107,11 +107,6 @@ def __init__(self, local_spaces: Collection, ensemble: Ensemble):
         # subfunctions that view the correct subfunctions of this big space.
         self._full_local_space = MixedFunctionSpace(self.local_spaces)
 
-        # ensemble._comm is congruent with ensemble.global_comm not ensemble.comm
-        # because obj._comm is used for garbage collection, so it needs to be the
-        # communicator that the ensemble objects are collective over.
-        self._comm = internal_comm(ensemble._comm, self)
-
```
JHopeCollins, line 115: The `Ensemble` is not collective over `ensemble.comm`, it is collective over `ensemble.global_comm`. Will this be a problem for garbage collection?

connorjward: I don't think I'm changing anything fundamental there to what we were previously doing. It is currently the case that if you create a Vec using `global_comm` that it won't get cleaned up automatically.

firedrake/ensemble/ensemble.py
connorjward: I think the answer is no, and things hang. I confess to being quite optimistic here.

.github/workflows/core.yml
```diff
@@ -322,7 +322,8 @@ jobs:
           if [ ${{ inputs.target_branch }} = 'release' ]; then
             GUSTO_BRANCH='main'
           else
-            GUSTO_BRANCH='future'
+            # UNDO ME
+            GUSTO_BRANCH='connorjward/comm-fix'
```
```suggestion
            GUSTO_BRANCH='future'
```

firedrake/ensemble/ensemble.py
```diff
@@ -49,16 +47,13 @@ def __init__(self, comm, M, **kwargs):
...
-        # Internal global comm
-        self._comm = internal_comm(comm, self)
 
         ensemble_name = kwargs.get("ensemble_name", "Ensemble")
         # User and internal communicator for spatial parallelism, contains a
         # contiguous chunk of M processes from `global_comm`.
         self.comm = self.global_comm.Split(color=(rank // M), key=rank)
         self.comm.name = f"{ensemble_name} spatial comm"
         weakref.finalize(self, self.comm.Free)
-        self._spatial_comm = internal_comm(self.comm, self)
```
JHopeCollins, line 59: A comment to say why we still use `internal_comm(self.ensemble_comm)` below rather than `with temp_internal_comm` would be useful for future reference.

firedrake/ensemble/ensemble.py
connorjward, line 67:
```suggestion
        # Keep a reference to the internal communicator because some methods return
        # non-blocking requests and we need to avoid cleaning up the communicator before
        # they complete. Note that this communicator should *never* be passed to PETSc, as
        # objects created with the communicator will never get cleaned up.
        self._ensemble_comm = internal_comm(self.ensemble_comm, self)
```

PR #4770, opened 2025-12-08
interpolate two-forms

leo-collins:
Assemble the adjoint interpolation of a two-form.

### DIFF
--- a/firedrake/assemble.py
+++ b/firedrake/assemble.py
@@ -857,7 +857,7 @@ class BaseFormAssembler(AbstractFormAssembler):
             return sum(w*c for w, c in zip(expr.weights(), expr.components()))
 
         # If F: V3 x V2 -> R, then
-        # Interpolate(TestFunction(V1), F) -> Action(Interpolate(TestFunction(V1), TrialFunction(V2.dual())), F).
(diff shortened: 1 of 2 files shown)


firedrake/assemble.py
```diff
@@ -855,6 +855,12 @@ def restructure_base_form(expr, visited=None):
         if isinstance(expr, ufl.FormSum) and all(ufl.duals.is_dual(a.function_space()) for a in expr.arguments()):
             # Return ufl.Sum if we are assembling a FormSum with Coarguments (a primal expression)
             return sum(w*c for w, c in zip(expr.weights(), expr.components()))
+
+        if isinstance(expr, ufl.Interpolate) and isinstance(expr.argument_slots()[0], ufl.form.Form):
```
connorjward, line 862: Can this get a nice comment like the other cases in this function?

tests/firedrake/regression/test_interpolate.py
```diff
@@ -6,6 +6,13 @@
 cwd = abspath(dirname(__file__))
 
 
+def mat_equals(a, b):
+    """Check that two Matrices are equal."""
+    a = a.petscmat.copy()
+    a.axpy(-1.0, b.petscmat)
+    return a.norm() < 1e-14
```
pbrubeck: We would like to use the Frobenius norm (sum of squares of the entries) as opposed to the l2 norm of a matrix (which relies on the eigenvalues)

PR #4772, opened 2025-12-09
`firedrake-run-split-tests` prints help message with -h argument

JHopeCollins:
Print a help message with how to use the script.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -70,6 +70,7 @@ pyop2-clean = "pyop2.compilation:clear_compiler_disk_cache"
 
 [project.optional-dependencies]
 check = [
+  # TODO RELEASE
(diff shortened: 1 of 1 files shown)


pyproject.toml
```diff
@@ -70,7 +70,8 @@ pyop2-clean = "pyop2.compilation:clear_compiler_disk_cache"
 
 [project.optional-dependencies]
 check = [
-  "mpi-pytest>=2025.7",
+  # TODO RELEASE
+  "mpi-pytest @ git+https://github.com/firedrakeproject/mpi-pytest.git@main",
```
JHopeCollins, line 74: @connorjward we want a `# TODO RELEASE` here right? So we only use mpi-pytest releases for Firedrake releases?

connorjward: Yep.

PR #4773, opened 2025-12-10
Function space method to create broken function space.

JHopeCollins:
In response to review on #4716 .
This method returns an equivalent function space built using `finat.ufl.BrokenElement(self.ufl_element)`.

~~Depends on https://github.com/firedrakeproject/fiat/pull/201~~ Merged.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -23,7 +23,7 @@ dependencies = [
   # TODO RELEASE
   "fenics-ufl @ git+https://github.com/FEniCS/ufl.git@main",
   # TODO RELEASE
-  "firedrake-fiat @ git+https://github.com/firedrakeproject/fiat.git@JHopeCollins/mixed-broken-element",
+  "firedrake-fiat @ git+https://github.com/firedrakeproject/fiat.git@main",
   "h5py>3.12.1",
   "immutabledict",
   "libsupermesh",


firedrake/functionspaceimpl.py
```diff
@@ -403,6 +403,13 @@ def make_function_space(cls, mesh, element, name=None):
             new = cls.create(new, mesh)
         return new
 
+    def broken_space(self):
```
connorjward, line 406:
```suggestion
    def broken_space(self) -> WithGeometryBase:
```

firedrake/functionspaceimpl.py
```diff
@@ -403,6 +403,13 @@ def make_function_space(cls, mesh, element, name=None):
...
+        """Return a :class:`.WithGeometryBase` with a BrokenElement
+        constructed from this function space's FiniteElement."""
```
connorjward: Docstring should ideally include a link to the finat.ufl.BrokenElement. If you don't know what a broken element is this doesn't help you find out.

firedrake/functionspaceimpl.py
```diff
@@ -403,6 +403,13 @@ def make_function_space(cls, mesh, element, name=None):
...
+        return type(self).make_function_space(
+            self.mesh(), finat.ufl.BrokenElement(self.ufl_element()),
+            name=f"{self.name}_broken")
```
connorjward: Can this result in `"None_broken"`? Maybe don't set a `name` if `self.name is None`?

firedrake/functionspaceimpl.py
```diff
@@ -403,6 +403,13 @@ def make_function_space(cls, mesh, element, name=None):
             new = cls.create(new, mesh)
         return new
 
+    def broken_space(self):
+        """Return a :class:`.WithGeometryBase` with a BrokenElement
+        constructed from this function space's FiniteElement."""
+        return type(self).make_function_space(
```
connorjward, line 415: What's the difference between `make_function_space` and `reconstruct`?

firedrake/adjoint/transformed_functional.py
```diff
@@ -156,48 +155,6 @@ def __init__(self, target: WithGeometry, **kwargs):
...
-    -------
-
-    firedrake.functionspaceimpl.WithGeometry
-        A DG space containing `space` as a subspace. May be `space`.
-    """
-
-    if is_dg_space(space):
-        return space
-    else:
-        return fd.FunctionSpace(space.mesh(), finat.ufl.BrokenElement(space.ufl_element()))
```
jrmaddison, line 198: This is needed for some later optimizations in the transformed functional. Suggest either keeping here, using `space.broken_space()` for the `else` case, or else applying this optimization in `broken_space`.

firedrake/functionspaceimpl.py
JHopeCollins: If you are a component of a mixed/vector/etc function space then `reconstruct` will go back up to the parent, rebuild the full space, then return the same component of that space. It depends on the particular instantiation.

`make_function_space` is a `classmethod`, and takes a mesh and finite element to create the function space (the `FunctionSpace` free function is essentially a wrapper around `make_function_space`).

It is confusing because of the whole with/without geometry thing. You have `__init__`, `create`, `make_function_space`, `reconstruct`, `collapse`, `__new__` all used by different function space classes.

firedrake/adjoint/transformed_functional.py
JHopeCollins: Do you mean keeping the behaviour to return `self` if its already discontinuous?

JHopeCollins: Is there any difference in behaviour with `dg_elem` vs `BrokenElement(dg_elem)` or is it an optimisation to reduce symbolic computations?

jrmaddison: Yes, that's right

JHopeCollins: Sorry, which is right?
Will this break the transformed reduced functional?

pyproject.toml
```diff
@@ -23,7 +23,7 @@ dependencies = [
   # TODO RELEASE
   "fenics-ufl @ git+https://github.com/FEniCS/ufl.git@main",
   # TODO RELEASE
-  "firedrake-fiat @ git+https://github.com/firedrakeproject/fiat.git@main",
+  "firedrake-fiat @ git+https://github.com/firedrakeproject/fiat.git@JHopeCollins/mixed-broken-element",
```
JHopeCollins: **REVERT BEFORE MERGE**
```suggestion
  "firedrake-fiat @ git+https://github.com/firedrakeproject/fiat.git@main",
```

firedrake/functionspaceimpl.py
connorjward: Yuck. Not your problem then.

firedrake/adjoint/transformed_functional.py
jrmaddison: > Do you mean keeping the behaviour to return self if its already discontinuous?

Yes, that's right.

This is detected with `space is space_D`.

JHopeCollins: Ok, I've put `is_dg_space` back in.

firedrake/functionspaceimpl.py
pbrubeck: Did you try `self.reconstruct(element=BrokenElement(self.ufl_element()))`?

pbrubeck: Would that break on a component space?

JHopeCollins: I don't know if it would break, but I preferred `make_function_space`  because then `broken_space` does only one thing (break the space), rather than also trying to reconstruct the family tree. I expect that this is will be the more common use.

connorjward:
This should not have been merged. All of the tests were failing due to unrelated changes but critically this introduced a docs build failure.

PR #4775, opened 2025-12-11
Fix domain_integral_type_map

pbrubeck:
# Description
`integral_data.domain_integral_type_map` was not encoding domains of coefficients that are not defined on the integration domain. 

All domains in the form are mapped to `integral_type=None`, but it seems that active coefficients need `integral_type="cell"`.

### DIFF
--- a/tsfc/driver.py
+++ b/tsfc/driver.py
@@ -135,15 +135,22 @@ def compile_integral(integral_data, form_data, prefix, parameters, *, diagonal=F
             if coeff in form_data.coefficient_split:
                 coefficient_split[coeff] = form_data.coefficient_split[coeff]
             coefficient_numbers.append(form_data.original_coefficient_positions[i])
+
     mesh = integral_data.domain
     all_meshes = extract_domains(form_data.original_form)
     domain_number = all_meshes.index(mesh)
+    coefficient_meshes = chain.from_iterable(map(extract_domains, coefficients))
+
+    domain_integral_type_map = dict.fromkeys(all_meshes, None)
+    domain_integral_type_map.update(dict.fromkeys(coefficient_meshes, "cell"))
+    domain_integral_type_map.update(integral_data.domain_integral_type_map)
+
     integral_data_info = TSFCIntegralDataInfo(
         domain=integral_data.domain,
         integral_type=integral_data.integral_type,
         subdomain_id=integral_data.subdomain_id,
         domain_number=domain_number,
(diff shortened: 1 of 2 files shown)


tsfc/driver.py
```diff
@@ -138,12 +138,20 @@ def compile_integral(integral_data, form_data, prefix, parameters, *, diagonal=F
     mesh = integral_data.domain
     all_meshes = extract_domains(form_data.original_form)
     domain_number = all_meshes.index(mesh)
+
+    domain_integral_type_map = dict.fromkeys(all_meshes, None)
+    for c, enabled in zip(form_data.reduced_coefficients, integral_data.enabled_coefficients):
+        if enabled:
+            d = extract_unique_domain(c)
+            domain_integral_type_map[d] = "cell"
```
connorjward: Why "cell"? What isn't `integral_data.domain_integral_type_map` not encoding?

tsfc/driver.py
```diff
@@ -135,15 +135,22 @@ def compile_integral(integral_data, form_data, prefix, parameters, *, diagonal=F
...
                 coefficient_split[coeff] = form_data.coefficient_split[coeff]
             coefficient_numbers.append(form_data.original_coefficient_positions[i])
+
     mesh = integral_data.domain
     all_meshes = extract_domains(form_data.original_form)
     domain_number = all_meshes.index(mesh)
+    coefficient_meshes = chain.from_iterable(map(extract_domains, coefficients))
+
+    domain_integral_type_map = dict.fromkeys(all_meshes, None)
+    domain_integral_type_map.update(dict.fromkeys(coefficient_meshes, "cell"))
```
pbrubeck, line 145: I'm not sure if we just need "cell" for `all_meshes`

pbrubeck: `integral_type = None` has a meaning, so we should only add domains for the active coefficients
https://github.com/firedrakeproject/firedrake/blob/f14d0f5c6bd8f663f7cef67bbfc358f548af45ed/tsfc/kernel_interface/common.py#L508-L511

connorjward (approved):
The code is pretty opaque. But it is not worse than before and it fixes a bug. I'm sufficiently happy.

connorjward:
I think it is difficult to do things automatically for every case. For instance if doing an exterior facet integral over the submesh are the facets of the parent mesh interior or exterior? There are gonna be similar difficulties for facet submeshes (ie codim 1).

We could probably reasonably get things to work automatically for cell integrals specifically, because there isn't this ambiguity.

connorjward:
Can this be reverted?

PR #4778, opened 2025-12-12
Assembly of matfree interpolation operator

leo-collins:
Allow assembly of matfree interpolation operators by passing `mat_type="matfree"`. 

Also changes the interfaces to the matrix classes and `ImplicitMatrixContext` to make them simpler.


firedrake/matrix.py
```diff
@@ -13,52 +21,54 @@ def __init__(self, handle):
...
         self.comm = test.function_space().comm
         self.block_shape = (len(test.function_space()),
                             len(trial.function_space()))
-        self.mat_type = mat_type
-        """Matrix type.
-
-        Matrix type used in the assembly of the PETSc matrix: 'aij', 'baij', 'dense' or 'nest',
-        or 'matfree' for matrix-free."""
-        self.form_compiler_parameters = fc_params
+        self.form_compiler_parameters = fc_params or {}
```
pbrubeck: An empty dictionary will be replaced with a different empty dictionary, this might result in unwanted behaivor.

firedrake/formmanipulation.py
```diff
@@ -180,8 +180,7 @@ def matrix(self, o):
             args.append(asplit)
 
         submat = o.petscmat.createSubMatrix(*ises)
-        bcs = ()
-        return AssembledMatrix(tuple(args), bcs, submat)
+        return AssembledMatrix(tuple(args), submat)
```
pbrubeck, line 195: What happened to the BCs?

pbrubeck: BTW, this will be fixed in https://github.com/firedrakeproject/firedrake/pull/4708/files

leo-collins: BCs are an optional argument now

firedrake/matrix.py
```diff
@@ -224,27 +233,34 @@ def assemble(self):
...
+            an iterable of boundary conditions to apply to this :class:`Matrix`.
+            May be `None` if there are no boundary conditions to apply. By default None.
+        options_prefix
+            PETSc options prefix to apply, by default None.
+        """
+        super().__init__(args, bcs=bcs)
 
         self.petscmat = petscmat
-        if options_prefix is not None:
+        if options_prefix:
```
pbrubeck: We definitely want to call `setOptionsPrefix()` when `options_prefix == ""` (this is a very common use case when you want to set options with no prefix)
```suggestion
        if options_prefix is not None:
```

firedrake/matrix_free/operators.py
```diff
@@ -67,35 +69,43 @@ class ImplicitMatrixContext(object):
...
+        """
         from firedrake.assemble import get_assembler
 
         self.a = a
         self.aT = adjoint(a)
         self.comm = a.arguments()[0].function_space().comm
-        self.fc_params = fc_params
-        self.appctx = appctx
+        self.fc_params = fc_params or {}
+        self.appctx = appctx or {}
```
pbrubeck:
```suggestion
        self.fc_params = {} if fc_params is None else fc_params
        self.appctx = {} if appctx is None else appctx
```

tests/firedrake/regression/test_interpolator_types.py
```diff
@@ -141,33 +175,36 @@ def test_mixed_same_mesh_mattype(value_shape, mat_type, sub_mat_type):
...
+    if mat_type == "matfree":
+        assert I_mat.petscmat.type == "python"
+    elif not mat_type or mat_type == "aij":
+        assert I_mat.petscmat.type == "seqaij"
     else:
-        assert res.petscmat.type == "nest"
-        for (i, j) in [(0, 0), (0, 1), (1, 0), (1, 1)]:
-            sub_mat = res.petscmat.getNestSubMatrix(i, j)
+        assert I_mat.petscmat.type == "nest"
+        for i in range(2):
```
pbrubeck: Should we test that the off-diagonal blocks are NULL?

leo-collins: Sure

leo-collins: I'm not sure how to do this. I get a segfault if I call any method on the off-diagonal sub-blocks

pbrubeck: That's because they are NULL. I think you want
```
assert not petscmat
```

leo-collins: Done now. Thanks

firedrake/formmanipulation.py
pbrubeck: Better to not change this line to avoid conflicts

firedrake/matrix.py
```diff
@@ -12,53 +20,64 @@ def __init__(self, handle):
         self.handle = handle
 
 
+def _get_mat_type(petscmat: PETSc.Mat) -> str:
+    """Maps PETSc matrix types to Firedrake notation"""
```
pbrubeck, line 23: Do we want to map "python" -> "matfree"?

tests/firedrake/regression/test_interpolation_operators.py
```diff
@@ -17,58 +18,88 @@ def params():
...
         # Always seqaij for scalar
-        assert res.petscmat.type == prefix + "aij"
+        assert I_mat.petscmat.type == prefix + "aij"
     else:
         # Defaults to seqaij
-        assert res.petscmat.type == prefix + (mat_type if mat_type else "aij")
+        assert I_mat.petscmat.type == prefix + (mat_type if mat_type else "aij")
+
+    res = assemble(action(I_mat, f))
+    assert np.allclose(res.dat.data, exact.dat.data)
```
pbrubeck, line 69: Test `assemble(action(adjoint(I_mat), fstar))`

firedrake/matrix.py
leo-collins: In that case we should have an `ImplicitMatrix` which we manually set `mat_type=matfree`

tests/firedrake/regression/test_interpolation_operators.py
leo-collins: I don't think PETSc python mats support `hermitianTranspose`

pbrubeck: This should call `multHermitianTranspose`

leo-collins: Did you mean to test `assemble(action(adjoint(I_mat), fstar))`? Because that calls `hermitianTranspose` when assembling the adjoint

pbrubeck: Well the right here is to call `multHermitianTranspose` directly

pbrubeck: I think if we want to avoid an explicit adjoint, we would preprocess `action(adjoint(I_mat), fstar)` -> `action(fstar, I_mat)`. But I am not very sure if `action`  can be used like that yet

pbrubeck: But python matrices support `hermitianTranspose` via tha MatType Transpose (which swaps the roles of mult and multTranspose)

leo-collins: I was thinking exactly that. This would need a preprocess case adding too I think

firedrake/matrix.py
```diff
@@ -12,53 +20,64 @@ def __init__(self, handle):
...
-        conditions to apply.
-    :arg mat_type: matrix type of assembled matrix, or 'matfree' for matrix-free
-    :kwarg fc_params: a dict of form compiler parameters of this matrix
-    """
-    def __init__(self, a, bcs, mat_type, fc_params=None):
+
+    def __init__(
+        self,
+        a: ufl.BaseForm | TensorBase | tuple[BaseArgument, BaseArgument],
+        bcs: Iterable[BCBase] | None = None,
```
connorjward: Why is this allowed to be `None`? The default could just be `()`.

firedrake/matrix.py
```diff
@@ -96,13 +115,10 @@ def bcs(self, bcs):
             self._bcs = ()
 
     def __repr__(self):
-        return "%s(a=%r, bcs=%r)" % (type(self).__name__,
-                                     self.a,
-                                     self.bcs)
+        return f"{type(self).__name__}(a={repr(self.a)}, bcs={repr(self.bcs)})"
```
connorjward:
```suggestion
        return f"{type(self).__name__}(a={self.a!r}, bcs={self.bcs!r})"
```

firedrake/matrix.py
```diff
@@ -96,13 +115,10 @@ def bcs(self, bcs):
...
 
     def __str__(self):
-        return "assembled %s(a=%s, bcs=%s)" % (type(self).__name__,
-                                               self.a, self.bcs)
+        return f"assembled {type(self).__name__}(a={str(self.a)}, bcs={str(self.bcs)})"
```
connorjward:
```suggestion
        return f"assembled {type(self).__name__}(a={self.a}, bcs={self.bcs})"
```

firedrake/matrix.py
```diff
@@ -224,27 +243,34 @@ def assemble(self):
...
-        options_prefix = kwargs.pop("options_prefix", None)
-        super(AssembledMatrix, self).__init__(a, bcs, "assembled")
+    def __init__(
+        self,
+        args: tuple[BaseArgument, BaseArgument],
+        petscmat: PETSc.Mat,
+        bcs: Iterable[BCBase] | None = None,
+        options_prefix: str | None = None,
+    ):
+        """A representation of a matrix that doesn't require knowing the underlying form.
```
connorjward: Just to note, I think that it's safe to have this docstring below the class name instead of inside `__init__`. Not too bothered either way though.

firedrake/formmanipulation.py
connorjward: Conflicts with what?

connorjward (changes requested):
I'm only reviewing this from a high level, but this seems pretty much fine.

firedrake/matrix.py
leo-collins: I think `repr(self.a)` is clearer

dham: `!r` is the idiomatic python so we should use it.

firedrake/matrix.py
pbrubeck: Shouldn't we be consistent with assemble.py and keep `bcs=None` as default? For the same reason other default kwargs are `None` and not empty iterables, such as `form_compiler_parameters`. In this way we would not be risking an API change, users in the future my see their code breaking if they were already passing `bcs=None`.

firedrake/formmanipulation.py
pbrubeck: Just to avoid conflicts with #4708
```suggestion
        bcs = ()
        return AssembledMatrix(tuple(args), submat, bcs=bcs)
```

connorjward: I don't think trivial-to-fix conflicts in unmerged PRs is a sufficient reason to avoid making a change.

firedrake/matrix.py
connorjward: `bcs=None` was a mistake IMO. It should probably always have been an empty tuple. `form_compiler_parameters` would ideally be `={}` but Python disallows this (same instance reused between invocations).
Having a default argument be a sensible (immutable) type means less processing for us and a more intuitive interface.

If we are concerned about breaking the API for users we can always warn when we see `None`.

pbrubeck: Fair enough. Let's just stick with `bcs = ()`, and not deal warnings. These classes aren't really used by users (although they are part of the public API).

firedrake/bcs.py
```diff
@@ -186,7 +185,8 @@ def zero(self, r):
             boundary condition should be applied.
 
         """
-        if isinstance(r, matrix.MatrixBase):
+        from firedrake.matrix import MatrixBase
```
pbrubeck: If this causes a cyclic import just to raise `NotImplementedError`, should we be importing `ufl.Matrix` instead?

Or even better, should we not validate the type and only check for `Function` or `Cofunction`?

firedrake/bcs.py
```diff
@@ -411,7 +411,8 @@ def apply(self, r, u=None):
         corresponding rows and columns.
 
         """
-        if isinstance(r, matrix.MatrixBase):
+        from firedrake.matrix import MatrixBase
```
pbrubeck: same here

firedrake/function.py
```diff
@@ -381,6 +383,8 @@ def interpolate(self,
             Returns `self`
         """
         from firedrake import interpolate, assemble
+        if len(extract_arguments(as_ufl(expression))) > 0:
```
pbrubeck: This seems unrelated to this PR. I'd be more inclined to check for the type as opposed to calling `extract_arguments`. 

It would only make sense to interpolate a `ufl.Expr,` or `BaseForm` with one coargument in the dual space of this Function.

firedrake/interpolation.py
```diff
@@ -306,15 +307,23 @@ def assemble(
         """
         self._check_mat_type(mat_type)
 
+        if mat_type == "matfree" and self.rank == 2:
```
pbrubeck, line 361: The docstring of this function needs to be updated

firedrake/matrix.py
```diff
@@ -133,39 +150,43 @@ def zero(self):
...
+        Parameters
+        ----------
+        a
+            The bilinear form this :class:`Matrix` represents.
+        mat
+            The underlying matrix object. Either a PyOP2 Mat or a PETSc Mat.
+        bcs
+            An iterable of boundary conditions to apply to this :class:`Matrix`.
+            May be `None` if there are no boundary conditions to apply.
+            By default None.
```
pbrubeck:
```suggestion
            By default an empty tuple.
```

firedrake/matrix.py
```diff
@@ -174,47 +195,48 @@ def assemble(self):
...
+        ----------
+        a
+            The bilinear form this :class:`ImplicitMatrix` represents.
+        ctx
+            An :class:`firedrake.matrix_free.operators.ImplicitMatrixContext` that
+            defines the operations of the matrix.
+        bcs
+            An iterable of boundary conditions to apply to this :class:`Matrix`.
+            May be `None` if there are no boundary conditions to apply.
+            By default None.
```
pbrubeck:
```suggestion
            By default an empty tuple.
```

firedrake/matrix.py
```diff
@@ -224,27 +246,35 @@ def assemble(self):
...
+
+        Parameters
+        ----------
+        args
+            A tuple of the arguments the matrix represents.
+        petscmat
+            The PETSc matrix this object wraps.
+        bcs
+            an iterable of boundary conditions to apply to this :class:`Matrix`.
+            May be `None` if there are no boundary conditions to apply. By default None.
```
pbrubeck:
```suggestion
            By default an empty tuple.
```

firedrake/matrix_free/operators.py
```diff
@@ -67,35 +69,43 @@ class ImplicitMatrixContext(object):
...
-
-    """
     @PETSc.Log.EventDecorator()
-    def __init__(self, a, row_bcs=[], col_bcs=[],
-                 fc_params=None, appctx=None):
+    def __init__(
+        self,
+        a: ufl.BaseForm,
+        row_bcs: Iterable[DirichletBC] | None = None,
+        col_bcs: Iterable[DirichletBC] | None = None,
```
pbrubeck: Are we leaving these as `None`?

firedrake/matrix.py
```diff
@@ -12,53 +20,65 @@ def __init__(self, handle):
...
+        Parameters
+        ----------
+        a
+            A UFL BaseForm (with two arguments) that this MatrixBase represents,
+            or a tuple of the arguments it represents, or a slate TensorBase.
+        fc_params
+            A dictionary of form compiler parameters for this matrix.
+        bcs
+            An optional iterable of boundary conditions to apply to this :class:`MatrixBase`.
+            None by default.
```
pbrubeck:
```suggestion
            An optional iterable of boundary conditions to apply to this :class:`MatrixBase`.
            Empty tuple by default.
```

firedrake/matrix.py
```diff
@@ -133,39 +149,43 @@ def zero(self):
...
-        fc_params = kwargs.pop("fc_params", None)
-        MatrixBase.__init__(self, a, bcs, mat_type, fc_params=fc_params)
-        options_prefix = kwargs.pop("options_prefix", None)
-        self.M = op2.Mat(*args, mat_type=mat_type, **kwargs)
+    """A representation of an assembled bilinear form."""
+
+    def __init__(
+        self,
+        a: ufl.BaseForm,
+        mat: op2.Mat | PETSc.Mat,
```
leo-collins, line 172: @connorjward I have no clue how to get the `op2.Mat` type hint to work. Oddly enough if I replace it with `op2.Set` then sphinx is able to find that reference. Any ideas?

firedrake/matrix.py
```diff
@@ -133,39 +149,43 @@ def zero(self):
...
+
+        Parameters
+        ----------
+        a
+            The bilinear form this :class:`Matrix` represents.
+        mat
+            The underlying matrix object. Either a PyOP2 Mat or a PETSc Mat.
+        bcs
+            An iterable of boundary conditions to apply to this :class:`Matrix`.
+            May be `None` if there are no boundary conditions to apply.
```
pbrubeck: Docstring out of date

docs/source/conf.py
dham, line 142:
```suggestion
```

firedrake/matrix.py
dham:
```suggestion
```

firedrake/matrix.py
```diff
@@ -12,53 +19,65 @@ def __init__(self, handle):
...
+            Empty tuple by default.
+        fc_params
+            A dictionary of form compiler parameters for this matrix.
+        """
         if isinstance(a, tuple):
             self.a = None
             test, trial = a
             arguments = a
         else:
+            assert isinstance(a, ufl.BaseForm | slate.TensorBase)
```
connorjward: Assertions should only ever be encountered in non-user facing code. This should raise a `TypeError` instead.

firedrake/matrix.py
```diff
@@ -12,53 +19,65 @@ def __init__(self, handle):
...
+
+    def __init__(
+        self,
+        a: ufl.BaseForm | slate.TensorBase | tuple[ufl.Argument | ufl.Coargument, ufl.Argument | ufl.Coargument],
+        bcs: Iterable[DirichletBC] = (),
+        fc_params: dict[str, Any] | None = None,
+    ):
+        """Initialise a :class:`MatrixBase`.
+
+        Parameters
```
connorjward: Nitpick: the arguments to `__init__` are supposed to be documented in the class docstring (https://numpydoc.readthedocs.io/en/latest/format.html#class-docstring)

connorjward (changes requested):
I think this is 99.9% there. I think I'm happy to merge after these tiny points are addressed.

connorjward (approved):

connorjward:
@leo-collins is this ready for merging?

connorjward:
Also please avoid force pushing if possible. It made reviewing more difficult as it broke the "view changes since last review" option.

leo-collins:
@connorjward Yes. Thanks!

PR #4779, opened 2025-12-15
Leo/assemble action adjoint

leo-collins:
- Assembles `action(w*, A)` by computing `A* @ w*`
- Make the simplification `action(adjoint(A), w*) -> action(w*, A)` to avoid adjointing A

### DIFF
--- a/firedrake/assemble.py
+++ b/firedrake/assemble.py
@@ -32,7 +32,6 @@ from pyop2 import op2
 from pyop2.exceptions import MapValueError, SparsityFormatError
 from pyop2.types.mat import _GlobalMatPayload, _DatMatPayload
 from pyop2.utils import cached_property
-from ufl.algorithms.analysis import extract_arguments
 
 
 __all__ = "assemble",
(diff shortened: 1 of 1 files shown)


firedrake/assemble.py
```diff
@@ -830,6 +838,10 @@ def restructure_base_form(expr, visited=None):
                 # Replace arguments
                 return ufl.replace(right, replace_map)
 
+            if isinstance(left, ufl.Adjoint) and isinstance(right, ufl.Cofunction) and left.arguments()[-1].function_space() == right.function_space():
```
pbrubeck: Should `right` be any 1-form (as opposed to assembled `Cofunctions` only)?

pbrubeck: We are expecting valid arguments in `Action`, so there's no need to check the spaces.
```suggestion
            if isinstance(left, ufl.Adjoint) and is_rank1(right):
```

tests/firedrake/regression/test_interp_dual.py
```diff
@@ -352,3 +353,47 @@ def test_interp_dual_mixed(source_space, target_space):
...
+
+    a = interpolate(u, V2)  # V1 x V2^* -> R, equiv. V1 -> V2
+    assert a.arguments() == (TestFunction(V2.dual()), TrialFunction(V1))
+
+    a_adj = adjoint(a)  # V2^* x V1 -> R, equiv. V2^* -> V1^*
+    assert a_adj.arguments() == (TestFunction(V1), TrialFunction(V2.dual()))
+
+    f = assemble(inner(1, TestFunction(V2)) * dx)
+
+    expr = action(a_adj, f)
```
pbrubeck: Should the reference solution come from the most naive way of computing the action of the adjoint?
```suggestion
    expr = action(adjoint(assemble(a)), f)
```

firedrake/assemble.py
```diff
@@ -830,6 +839,10 @@ def restructure_base_form(expr, visited=None):
                 # Replace arguments
                 return ufl.replace(right, replace_map)
 
+            if isinstance(left, ufl.Adjoint) and len(extract_arguments(right)) == 1:
```
pbrubeck: Do we need to extract_arguments? What's wrong with `is_rank_1`?

leo-collins: `is_rank_1` returns true for Coefficients, so we end up with `action(Coef, Form)` which fails

pbrubeck: `action(Coefficient, Form)` should be fixed then (in a UFL PR). Maybe just add an extra check on `not isinstance(right, Function)` for the moment.

PR #4785, opened 2025-12-18
DO NOT SQUASH: merge release into main

pbrubeck:
# Description


connorjward:
> @connorjward are these failures unrelated?

The netgen failures are expected. The reduced functional ones are not. It looks like that file needs an additional import since Leo cleaned our namespace up.

PR #4792, opened 2025-12-19
Assemble mixed cross-mesh interpolation operator

leo-collins:
Allows assembly of interpolation matrices between mixed spaces over different domains.

Change default sub_mat_type for nested MixedInterpolator from 'baij' to 'aij' for greater compatibility.

### DIFF
--- a/docs/source/interpolation.rst
+++ b/docs/source/interpolation.rst
@@ -109,7 +109,7 @@ This also works when interpolating into a space defined on the facets of the mes
 
 
 Semantics of symbolic interpolation
---------------------------------------------
+-----------------------------------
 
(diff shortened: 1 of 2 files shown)


firedrake/interpolation.py
```diff
@@ -434,6 +434,7 @@ class CrossMeshInterpolator(Interpolator):
     def __init__(self, expr: Interpolate):
         super().__init__(expr)
         self.target_mesh = self.target_mesh.unique()
+        self.source_mesh = self.source_mesh.unique()
```
dham: rather than rewriting `source_mesh` and `target_mesh`, just call `.unique()` when they are used. This is likely to be less confusing.

dham (changes requested):
Apart from the style issue this looks good.

dham (changes requested):
Please document!

PR #4794, opened 2025-12-20
split zero-form interpolate

leo-collins:
make `split_form(interp)` do the right thing for zero-form `Interpolate`


firedrake/formmanipulation.py
```diff
@@ -254,6 +259,11 @@ def split_form(form, diagonal=False):
     """
     splitter = ExtractSubBlock()
     args = form.arguments()
+
+    if isinstance(form, Interpolate) and not args:
```
pbrubeck, line 263: Should the `args` be added as a kwarg?

PR #4804, opened 2026-01-05
Leo/cross mesh non lagrange

leo-collins:
Implement cross-mesh interpolate into non-point evaluation spaces.

To do this we first construct a quadrature space which has point-evaluation dofs at the quadrature points of the target element. We then cross-mesh interpolate into this space. Finally, we interpolate from this quadrature space into the target space.

### DIFF
--- a/tests/firedrake/regression/test_cross_mesh_non_lagrange.py
+++ b/tests/firedrake/regression/test_cross_mesh_non_lagrange.py
@@ -94,7 +94,7 @@ def test_cross_mesh(V, rank):
         assert np.allclose(f_interpolated_direct.dat.data_ro, f_direct.dat.data_ro)
 
 
-# @pytest.mark.parallel([1, 3])
+@pytest.mark.parallel([1, 3])
 @pytest.mark.parametrize("rank", [0, 1, 2])
 def test_cross_mesh_adjoint(V, rank):
     # Can already do Lagrange -> RT adjoint


firedrake/interpolation.py
```diff
@@ -536,9 +588,16 @@ def _get_callable(self, tensor=None, bcs=None, mat_type=None, sub_mat_type=None)
 
         # self.ufl_interpolate.function_space() is None in the 0-form case
         V_dest = self.ufl_interpolate.function_space() or self.target_space
-        f = tensor or Function(V_dest)
 
-        point_eval, point_eval_input_ordering = self._get_symbolic_expressions()
+        # Interpolate into intermediate quadrature space for non-identity mapped elements
+        if into_quadrature_space := V_dest.ufl_element().mapping() != "identity":
```
pbrubeck: Careful here with symmetric TensorElements

firedrake/interpolation.py
```diff
@@ -470,16 +473,68 @@ def __init__(self, expr: Interpolate):
...
+        -------
+        WithGeometry
+            A :class:`.WithGeometry` function space with quadrature element.
+        """  
+        V_element = V.finat_element
+        _, points = V_element.dual_basis
+        weights = numpy.full(len(points.points), numpy.nan)  # These can be any number since we never integrate
+        quad_scheme = QuadratureRule(points, weights, ref_el=V_element.cell)
+        element = FiniteElement("Quadrature", degree=V_element.degree, quad_scheme=quad_scheme)
+        return self._fs_type(V)(V.mesh(), element)
```
pbrubeck: The Quadrature space is always scalar-valued, even if it came from a vector-valued space like RT. You should grab the value_shape, and not the Tensorness of the ufl element.
 
 This should be generic enough, although it might be better to special case the scalar case
```suggestion
        return TensorFunctionSpace(V.mesh(), "Quadrature", degree=V_element.degree, quad_scheme=quad_scheme, shape=V.value_shape)
```

firedrake/interpolation.py
```diff
@@ -470,16 +473,68 @@ def __init__(self, expr: Interpolate):
                         "Can't yet cross-mesh interpolate onto function spaces made from VectorElements "
                         "or TensorElements made from sub elements with value shape other than ()."
```
pbrubeck: This should no longer be the case after this PR

firedrake/interpolation.py
pbrubeck: Ah nevermind, _fs_type does the check on the `value_shape`.

firedrake/interpolation.py
```diff
@@ -470,16 +473,68 @@ def __init__(self, expr: Interpolate):
...
+        V
+            A :class:`.WithGeometry` function space.
+
+        Returns
+        -------
+        Callable
+            A callable which returns a :class:`.WithGeometry` matching the type of V.
+        """
+        # Get the correct type of function space
+        shape = V.ufl_function_space().value_shape
```
```suggestion
        shape = V.value_shape
```

firedrake/interpolation.py
```diff
@@ -470,16 +473,68 @@ def __init__(self, expr: Interpolate):
...
+        if len(shape) == 0:
+            return FunctionSpace
+        elif len(shape) == 1:
+            return partial(VectorFunctionSpace, dim=shape[0])
+        else:
+            symmetry = V.ufl_element().symmetry()
```
pbrubeck: Just a comment, the Quadrature space may or not enforce the symmetry, but it is always preferable to enforce it. This means that VomOntoVomInterpolator needs to support interpolation of symmetric tensors. 

Also related to this, some elements are intrinsically symmetric-valued (all basis functions are symmetric) even when the `ufl_element().symmetry()` is None, for example HHJ and Regge. This is similar to the case of intrinsically vector-valued elements like RT, where the basis functions are vector-valued but they don't need a `VectorElement`. 

What I am trying to say with this is that when we have intrisic symmetry we can always reduce the size of the Quadrature space by further inspecting the finat elements. But for now, the current approach to construct the Quadrature space seems fine.

firedrake/interpolation.py
pbrubeck: As a consequence of https://github.com/firedrakeproject/firedrake/pull/4607 we no longer need the ufl element.
```suggestion
        return self._fs_type(V)(V.mesh(), "Quadrature", degree=V_element.degree, quad_scheme=quad_scheme)
```

tests/firedrake/regression/test_cross_mesh_non_lagrange.py
```diff
@@ -0,0 +1,197 @@
...
+    V_target = fs_shape(V)(mesh1, "CG", 1)
+    mesh2 = V.mesh()
+    x2 = SpatialCoordinate(mesh2)
+
+    if len(V.value_shape) > 1:
+        expr = outer(x2, x2)
+        target_expr = outer(x1, x1)
+        if V.ufl_element().mapping() == "covariant contravariant Piola":
+            expr = deviatoric_expr_2x2(expr)
+            target_expr = deviatoric_expr_2x2(target_expr)
```
pbrubeck:
```suggestion
            expr = dev(expr)
            target_expr = dev(target_expr)
```

firedrake/interpolation.py
```diff
@@ -521,11 +572,21 @@ def _get_callable(self, tensor=None, bcs=None, mat_type=None, sub_mat_type=None)
             raise NotImplementedError("bcs not implemented for cross-mesh interpolation.")
         mat_type = mat_type or "aij"
 
-        # self.ufl_interpolate.function_space() is None in the 0-form case
-        V_dest = self.ufl_interpolate.function_space() or self.target_space
-        f = tensor or Function(V_dest)
+        # Interpolate into intermediate quadrature space for non-identity mapped elements
+        if into_quadrature_space := self.target_space.ufl_element().mapping() != "identity":
```
pbrubeck: Ideally we want `finat_element.has_pointwise_dual_basis` once it gets merged 
```suggestion
        if into_quadrature_space := self.target_space.finat_element.mapping != "affine":
```

docs/source/interpolation.rst
```diff
@@ -267,8 +267,16 @@ function space has finite elements (as given in the list of
...
+* **Discontinuous Lagrange/DG** (also known as Discontinuous Galerkin or DP elements),
+* **DQ** (i.e. Discontinuous Lagrange/DG on lines, quadrilaterals and hexahedra),
+* **Raviart-Thomas/RT**,
+* **Brezzi-Douglas-Marini/BDM**,
+* **Brezzi-Douglas-Fortin-Marini/BDFM**,
+* **Hellan-Herrmann-Johnson/HHJ**,
+* **Nedelec 1st kind/N1curl**,
+* **Nedelec 2nd kind/N2curl**,
+* **Gopalakrishnan-Lederer-Schoberl 1st kind/GLS**,
+* **Gopalakrishnan-Lederer-Schoberl 2nd kind/GLS2**.
```
pbrubeck: I think some elements are missing (Regge). Maybe add a comment saying that if the same-mesh interpolation works, then the cross mesh interpolation should also work.

dham: I think that a hand-written list in the documentation like this is just asking to be out of date. Do we have a way of determining from the code whether we can interpolate into a space? If so, then I think the right thing to do would be to mark on the automatically generated list of supported elements which ones can be the target of interpolation.

connorjward: There are two parts to achieving this:
1. Generating this list at all.
2. Inserting the list into the website.

I don't know how to do 1, but for 2 you should look at how we do this for other things in the docs `Makefile` ([example](https://github.com/firedrakeproject/firedrake/blob/main/docs/Makefile#L87)).

firedrake/interpolation.py
```diff
@@ -473,13 +527,19 @@ def _get_symbolic_expressions(self) -> tuple[Interpolate, Interpolate]:
         Raises
         ------
         DofNotDefinedError
-            If any DoFs in the target mesh cannot be defined in the source mesh.
+            If any of the target spaces dofs cannot be defined in the source mesh.
+        ValueError
+            If the target space does not have point-evaluation dofs.
         """
         from firedrake.assemble import assemble
+        if target_space.finat_element.mapping != "affine":
```
```suggestion
        if not target_space.finat_element.has_pointwise_dual_basis:
```

firedrake/interpolation.py
```diff
@@ -521,11 +572,15 @@ def _get_callable(self, tensor=None, bcs=None, mat_type=None, sub_mat_type=None)
             raise NotImplementedError("bcs not implemented for cross-mesh interpolation.")
         mat_type = mat_type or "aij"
 
-        # self.ufl_interpolate.function_space() is None in the 0-form case
-        V_dest = self.ufl_interpolate.function_space() or self.target_space
-        f = tensor or Function(V_dest)
+        # Interpolate into intermediate quadrature space for non-identity mapped elements
+        if into_quadrature_space := self.target_space.finat_element.mapping != "affine":
```
```suggestion
        if into_quadrature_space := not self.target_space.finat_element.has_pointwise_dual_basis:
```

firedrake/interpolation.py
dham: Can we turn this into an issue.

firedrake/interpolation.py
```diff
@@ -453,16 +456,67 @@ def __init__(self, expr: Interpolate):
...
+        shape = V.value_shape
+        if len(shape) == 0:
+            return FunctionSpace
+        elif len(shape) == 1:
+            return partial(VectorFunctionSpace, dim=shape[0])
+        else:
+            symmetry = V.ufl_element().symmetry()
+            return partial(TensorFunctionSpace, shape=shape, symmetry=symmetry)
+
+    def _get_quadrature_space(self, V: WithGeometry) -> WithGeometry:
```
dham: This is now superfluous once #4839 lands.

firedrake/interpolation.py
```diff
@@ -473,13 +527,19 @@ def _get_symbolic_expressions(self) -> tuple[Interpolate, Interpolate]:
         Raises
         ------
         DofNotDefinedError
-            If any DoFs in the target mesh cannot be defined in the source mesh.
+            If any of the target spaces dofs cannot be defined in the source mesh.
+        ValueError
```
dham: Create `DofTypeError`

firedrake/interpolation.py
```diff
@@ -536,47 +563,65 @@ def _get_callable(self, tensor=None, bcs=None, mat_type=None, sub_mat_type=None)
...
-
-            # The rest of the adjoint interpolation is the composition
-            # of the adjoint interpolators in the reverse direction.
-            # We don't worry about skipping over missing points here
-            # because we're going from the input ordering VOM to the original VOM
-            # and all points from the input ordering VOM are in the original.
+
             def callable() -> Cofunction:
+                if into_quadrature_space:
+                    cofunc = assemble(interpolate(TestFunction(target_space), self.dual_arg))
```
pbrubeck: If this `interpolate` expression is reconstructed at every call, we will be creating a new `Interpolator`. We might want to consider caching it

tests/firedrake/regression/test_cross_mesh_non_lagrange.py
```diff
@@ -0,0 +1,180 @@
...
+    wts = np.full(len(ps.points), np.nan)  # These can be any number since we never integrate
+    scheme = QuadratureRule(ps, wts, ref_el=fe.cell)
+    element = FiniteElement("Quadrature", degree=fe.degree, quad_scheme=scheme)
+    return fs_shape(V)(V.mesh(), element)
+
+
+@pytest.fixture(params=[("RT", 1), ("RT", 2), ("RT", 3), ("RT", 4), ("BDM", 1), ("BDM", 2), ("BDM", 3),
+                        ("BDFM", 2), ("HHJ", 2), ("N1curl", 1), ("N1curl", 2), ("N1curl", 3), ("N1curl", 4),
+                        ("N2curl", 1), ("N2curl", 2), ("N2curl", 3), ("GLS", 1), ("GLS", 2), ("GLS", 3),
+                        ("GLS", 4), ("GLS2", 1), ("GLS2", 2), ("GLS2", 3)],
```
pbrubeck: Aren't these too many tests? Could we take only one degree per space?

pbrubeck: An interesting case for HHJ is degree=0, also for Regge.

firedrake/interpolation.py
pbrubeck: I'm fine if this goes as is, and then address caching as a separate PR

tests/firedrake/regression/test_cross_mesh_non_lagrange.py
leo-collins: I've reduced them but I don't really see an issue, the new tests don't appear in the top 50 slowest durations

pbrubeck: Even if you tested every possible element we support up to a reasonable degree, the tests are ranked individually so they will not appear in the top 50.

tests/firedrake/regression/test_cross_mesh_non_lagrange.py
```diff
@@ -0,0 +1,164 @@
...
+    # Can already do Lagrange -> RT adjoint
+    # V^* -> Q^* -> V_target^*
+    name = V.ufl_element()._short_name
+    deg = V.ufl_element().degree()
+    if name in ["N1curl", "GLS", "RT"] and deg == 1:
+        if name == "RT" and rank == 0:
+            pass
+        else:
+            pytest.skip(f"Not exact for degree {deg} {name} elements")
+    elif name in ["Regge", "HHJ"] and deg == 0:
```
pbrubeck, line 106: What happened here? Are these just skipped?

tests/firedrake/regression/test_interpolate_cross_mesh.py
```diff
@@ -188,14 +180,14 @@ def parameters(request):
         V_src = FunctionSpace(m_src, "N1curl", 2)  # Not point evaluation nodes
         V_dest = VectorFunctionSpace(m_dest, "CG", 4)
         V_dest_2 = VectorFunctionSpace(m_dest, "DQ", 2)
-    elif request.param == "unitsquare_SminusDiv_destination":
-        m_src, m_dest, coords = unitsquaresetup()
+    elif request.param == "unitsquare_RT_N1curl_destination":
```
pbrubeck, line 183: Is this a sensible test? One space would have discontinuities where the other is continuous. We should only test well defined operations. I think we should go from a continuous space to a discontinuous one.

tests/firedrake/regression/test_cross_mesh_non_lagrange.py
```diff
@@ -0,0 +1,164 @@
...
+        f_interpolated_manual = assemble(I2)
+        assert f_interpolated_manual.function_space() == V
+        assert np.allclose(f_interpolated_manual.dat.data_ro, f_direct.dat.data_ro)
+
+        f_interpolated_direct = assemble(interpolate(f_source, V))
+        assert f_interpolated_direct.function_space() == V
+        assert np.allclose(f_interpolated_direct.dat.data_ro, f_direct.dat.data_ro)
+
+
+# @pytest.mark.parallel([1, 3])
```
pbrubeck:
```suggestion
@pytest.mark.parallel([1, 3])
```

firedrake/interpolation.py
```diff
@@ -532,47 +556,65 @@ def _get_callable(self, tensor=None, bcs=None, mat_type=None, sub_mat_type=None)
...
-
-            # The rest of the adjoint interpolation is the composition
-            # of the adjoint interpolators in the reverse direction.
-            # We don't worry about skipping over missing points here
-            # because we're going from the input ordering VOM to the original VOM
-            # and all points from the input ordering VOM are in the original.
+
             def callable() -> Cofunction:
+                if self.into_quadrature_space:
+                    cofunc = assemble(interpolate(TestFunction(self.target_space), self.dual_arg))
```
pbrubeck: Could this symbolic interpolate be cached?

leo-collins: I've cached it now but it's not made things noticeably faster. I imagine this is because it's using SameMeshInterpolator which is fast and already has caching

pbrubeck: Right, the codegen was already cached, but SameMeshInterpolator lost reusability in the refactoring before the October release.

Before that  the callable method was only called once. But I don't expect a massive speedup if we restore the reusability anyway.

tests/firedrake/regression/test_cross_mesh_non_lagrange.py
leo-collins: I don't think these are exact into CG1^*

tests/firedrake/regression/test_interpolate_cross_mesh.py
leo-collins: Which space is discontinuous?

pbrubeck: RT has discontinuous tangential components, and N1curl has discontinuous normal components

pbrubeck: We should test vector CG into RT or Nedelec

pbrubeck: Oh, I see, this is exactly what the test does, nevermind.

tests/firedrake/regression/test_cross_mesh_non_lagrange.py
```diff
@@ -0,0 +1,174 @@
...
+
+    if exact:
+        def close(x, y):
+            if rank == 0:
+                return np.isclose(x, y)
+            else:
+                return np.allclose(x, y)
+    else:
+        def close(x, y):
+            return np.linalg.norm(x - y)**2 < 1e-5
```
pbrubeck:
```suggestion
            return np.linalg.norm(x - y) < 0.003
```

tests/firedrake/regression/test_cross_mesh_non_lagrange.py
pbrubeck: Is it not better to change the expression to a globally constant one for these cases?

firedrake/interpolation.py
```diff
@@ -581,12 +641,18 @@ def callable() -> Function | Number:
                 else:
                     f.dat.data_wo[:] = f_point_eval_input_ordering.dat.data_ro[:]
 
+                if self.into_quadrature_space:
+                    f_target = Function(self.original_target_space)
+                    assemble(interpolate(f, self.original_target_space), tensor=f_target)
```
pbrubeck, line 646: This could be cached as well

firedrake/interpolation.py
```diff
@@ -486,44 +519,53 @@ def _get_symbolic_expressions(self) -> tuple[Interpolate, Interpolate]:
...
+        Interpolate
+            A symbolic interpolate expression.
+        """
+        if self.rank == 2:
+            if self.ufl_interpolate.is_adjoint:
+                return interpolate(TestFunction(self.target_space), self.original_target_space)
+            else:
+                return interpolate(TrialFunction(self.target_space), self.original_target_space)
+        elif self.ufl_interpolate.is_adjoint:
+            return interpolate(TestFunction(self.target_space), self.dual_arg)
```
pbrubeck, line 555: We will need the `else` statement to cache the forward 1-form interpolate into the quadrature space

firedrake/interpolation.py
pbrubeck: Is f changing between calls? If that is the case, we should revist this later

firedrake/interpolation.py
leo-collins: 1-form forward-interpolation doesn't use that function

firedrake/interpolation.py
leo-collins: Yes. I'll look at that in another PR

tests/firedrake/regression/test_cross_mesh_non_lagrange.py
```diff
@@ -0,0 +1,174 @@
...
+        f_interpolated_manual = assemble(I2)
+        assert f_interpolated_manual.function_space() == V
+        assert np.allclose(f_interpolated_manual.dat.data_ro, f_direct.dat.data_ro)
+
+        f_interpolated_direct = assemble(interpolate(f_source, V))
+        assert f_interpolated_direct.function_space() == V
+        assert np.allclose(f_interpolated_direct.dat.data_ro, f_direct.dat.data_ro)
+
+
+# @pytest.mark.parallel([1, 3])
```
pbrubeck:
```suggestion
@pytest.mark.parallel([1, 3])
```

pbrubeck (approved):

PR #4816, opened 2026-01-13
Submesh: relabel ghost entities as owned if comm.size == 1

pbrubeck:
# Description

### DIFF
--- a/tests/firedrake/submesh/test_submesh_comm.py
+++ b/tests/firedrake/submesh/test_submesh_comm.py
@@ -26,14 +26,20 @@ def assert_local_equality(A, Asub, V, Vsub):
 
 @pytest.mark.parallel([1, 3])
 @pytest.mark.parametrize("reorder", [False, True])
-def test_create_submesh_comm_self(reorder):
+@pytest.mark.parametrize("ignore_halo", [False, True])
+def test_create_submesh_comm_self(reorder, ignore_halo):
     subdomain_id = None
     nx = 4
-    mesh = UnitSquareMesh(nx, nx, quadrilateral=True, reorder=reorder)
-    submesh = Submesh(mesh, mesh.topological_dimension, subdomain_id, ignore_halo=True, reorder=reorder, comm=COMM_SELF)
+    mesh = UnitSquareMesh(nx, nx, quadrilateral=True, reorder=reorder, distribution_parameters={"overlap_type": (DistributedMeshOverlapType.VERTEX, 1)})
+    submesh = Submesh(mesh, mesh.topological_dimension, subdomain_id, ignore_halo=ignore_halo, reorder=reorder, comm=COMM_SELF)
     assert submesh.submesh_parent is mesh
     assert submesh.comm.size == 1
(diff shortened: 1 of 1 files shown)


tests/firedrake/submesh/test_submesh_comm.py
```diff
@@ -4,52 +4,68 @@
...
+def test_create_submesh_comm_self(reorder):
     subdomain_id = None
     nx = 4
-    mesh = UnitSquareMesh(nx, nx, quadrilateral=True, reorder=False)
-    submesh = Submesh(mesh, mesh.topological_dimension, subdomain_id, ignore_halo=True, reorder=False, comm=COMM_SELF)
+    mesh = UnitSquareMesh(nx, nx, quadrilateral=True, reorder=reorder)
+    submesh = Submesh(mesh, mesh.topological_dimension, subdomain_id, ignore_halo=True, reorder=reorder, comm=COMM_SELF)
     assert submesh.submesh_parent is mesh
     assert submesh.comm.size == 1
     assert submesh.cell_set.size == mesh.cell_set.size
```
connorjward: If we're doing this in parallel and doing a relabelling then shouldn't this now fail? This size of the cell set will now include the ghost points as well.

pbrubeck: `ignore_halo=True` just ignores the cells that are entirely in the halo. Cells that have some facets in the halo are still kept.

connorjward: So would this fail if we specify more overlap than the default? Just trying to understand, I don't think you've done anything wrong.

firedrake/cython/dmcommon.pyx
```diff
@@ -3906,52 +3906,67 @@ def submesh_correct_entity_classes(PETSc.DM dm,
...
-            assert has == PETSC_FALSE
-            CHKERR(DMLabelHasPoint(lbl_ghost, subp, &has))
-            assert has == PETSC_TRUE
-            CHKERR(DMLabelClearValue(lbl_ghost, subp, 1))
-            CHKERR(DMLabelSetValue(lbl_owned, subp, 1))
+            if not has:
+                CHKERR(DMLabelHasPoint(lbl_ghost, subp, &has))
+                if has:
+                    CHKERR(DMLabelClearValue(lbl_ghost, subp, 1))
+                    CHKERR(DMLabelSetValue(lbl_owned, subp, 1))
```
connorjward: What's the motivation behind labelling them owned and not core?

pbrubeck: I don't know if there's a difference. I copied these lines from the case `ownership_gain == 1`

tests/firedrake/submesh/test_submesh_comm.py
pbrubeck: This test still passes with more overlap e.g. `(DistributedMeshOverlapType.VERTEX, 2)`. The default `(DistributedMeshOverlapType.FACET, 1)` is still large enough to include ghost cells.

PR #4823, opened 2026-01-16
Test C2 elements

pbrubeck:
# Description
Tests https://github.com/firedrakeproject/fiat/pull/208, https://github.com/firedrakeproject/fiat/pull/209, https://github.com/firedrakeproject/fiat/pull/210, https://github.com/firedrakeproject/fiat/pull/212

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -23,7 +23,7 @@ dependencies = [
   # TODO RELEASE
   "fenics-ufl @ git+https://github.com/FEniCS/ufl.git@main",
   # TODO RELEASE
-  "firedrake-fiat @ git+https://github.com/firedrakeproject/fiat.git@pbrubeck/double-alfeld",
+  "firedrake-fiat @ git+https://github.com/firedrakeproject/fiat.git@main",
   "h5py>3.12.1",
   "immutabledict",
   "libsupermesh",


pyproject.toml
```diff
@@ -23,7 +23,7 @@ dependencies = [
   # TODO RELEASE
   "fenics-ufl @ git+https://github.com/FEniCS/ufl.git@main",
   # TODO RELEASE
-  "firedrake-fiat @ git+https://github.com/firedrakeproject/fiat.git@main",
+  "firedrake-fiat @ git+https://github.com/firedrakeproject/fiat.git@pbrubeck/double-alfeld",
```
pbrubeck:
```suggestion
  "firedrake-fiat @ git+https://github.com/firedrakeproject/fiat.git@main",
```

tsfc/fem.py
```diff
@@ -252,6 +252,11 @@ def reference_edge_tangents(self):
         num_edges = len(cell.get_topology()[1])
         return gem.Literal(numpy.asarray([cell.compute_edge_tangent(i) for i in range(num_edges)]))
 
+    def normalized_reference_edge_tangents(self):
```
connorjward, line 255: This isn't used. Why was this added?

pbrubeck: It's used in FIAT

pbrubeck: https://github.com/firedrakeproject/fiat/blob/33441d67603de43f8d068edc7b7677e57ed70b49/finat/c2_elements.py#L36

connorjward: Why doesn't it live in FIAT then? FIAT is supposed to be independent of Firedrake.

pbrubeck: It has it's own FIAT counterpart, but it is used only for tests.

pbrubeck: This class inherits from the FIAT one, and should know how to turn UFL things into GEM. It does not live in FIAT because it is the form compiler. Originally finat was not supposed to depend on UFL, but now that it does, we could probably migrate some of this there.

connorjward: Yeah we definitely borked the TSFC migration a bit. There are also a lot of form compiler components inside FIAT that shouldn't be there.

This isn't a problem for this PR to solve.

pbrubeck: some of these functions only involve reference values (as opposed to physical values), so they do not need to convert UFL things into GEM. It makes sense to implement this and other reference functions in the base class.

`PhysicalGeometry` is an abstract class and only has unimplemented template methods. In the FIAT tests (conftest.py) we implement all of this.

connorjward: GEM lives in finat so I think all the GEM stuff here should also be ported.

PR #4824, opened 2026-01-19
Fix for numpy 2.4.1

pbrubeck:
# Description

### DIFF
--- a/pyop2/types/mat.py
+++ b/pyop2/types/mat.py
@@ -1131,7 +1131,7 @@ class _DatMatPayload:
                     v.copy(z)
                     a = np.zeros(1, dtype=dtypes.ScalarType)
                     if x.comm.rank == 0:
-                        a[0] = x.array_r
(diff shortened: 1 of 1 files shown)


pyop2/types/mat.py
connorjward:
```suggestion
                        a[0] = x.array_r.item()
```

pyop2/types/mat.py
```diff
@@ -1089,7 +1089,7 @@ def mult(self, mat, x, y):
                     v.copy(y)
                     a = np.zeros(1, dtype=dtypes.ScalarType)
                     if x.comm.rank == 0:
-                        a[0] = x.array_r
+                        a[0] = x.array_r.item()
```
pbrubeck: Which one is more preferable?
```suggestion
                        a[...] = x.array_r
```

connorjward: Don't mind.

PR #4827, opened 2026-01-20
Cache numerical Interpolator on the symbolic Interpolate expression

pbrubeck:
# Description


pbrubeck:
> Couldn't you just have `Interpolate._interpolator` as a cached property?

That sounds fine, the diff will be longer, but seems more elegant. 

The cached property will have to take the role of `get_interpolator`, and `get_interpolator` will then just going to fetch the property. Does this sound correct to you?

PR #4829, opened 2026-01-21
TSFC: avoid setup of unused domains

pbrubeck:
# Description
A multi-domain Form is split into single-domain terms. Each term knows that it belongs to a multidomain problem, but is assembled individually. During the assembly of each single-domain term, we may skip the setup on the all the other domains.

https://github.com/firedrakeproject/firedrake/blob/4a5cba19466671c654de924955729c6d4e005886/tsfc/driver.py#L170-L175

These methods will now skip a mesh not found in `integral_data.domain_integral_type_map` (as returned by `ufl.algorithms.compute_form_data`). 

This is to remove the `integral_type = None` entries that were hacked into the `domain_integral_type_map`, which lead to obscure errors for illegal forms.


tsfc/kernel_interface/common.py
```diff
@@ -505,12 +504,13 @@ def prepare_coefficient(coefficient, name, domain_integral_type_map):
...
-    integral_type = domain_integral_type_map[domain]
-    if integral_type is None:
+    try:
+        integral_type = domain_integral_type_map[domain]
+    except KeyError:
         # This means that this coefficient does not exist in the DAG,
         # so corresponding gem expression will never be needed.
-        expression = None
-    elif integral_type.startswith("interior_facet"):
+        return None
```
connorjward, line 518: I think that here is another case of using `None`s inappropriately, but that's not relevant here.

connorjward (dismissed):
Thanks

tests/firedrake/submesh/test_submesh_solve.py
```diff
@@ -592,7 +592,7 @@ def _test_submesh_solve_3d_2d_poisson(simplex, direction, nref, degree):
     mesh12 = Submesh(mesh2, dim - 1, label_interf)
     dx1 = Measure("dx", mesh1)
     dx2 = Measure("dx", mesh2)
-    ds1_ds2 = Measure("ds", mesh1, intersect_measures=(Measure("ds", mesh2),))
+    ds1_ds2 = Measure("ds", mesh1, intersect_measures=(Measure("ds", mesh2), Measure("dx", mesh12)))
```
pbrubeck, line 595: We are hitting an error for Forms involving legal integrals on codim 1 multidomain problems. We are able to assemble each term individually. However, the interface mesh must be included in every surface Measure if assembling all integrals all at once.

PR #4839, opened 2026-01-26
Multigrid: native transfers for non-Lagrange elements

pbrubeck:
# Description
Previously, multigrid on non-Lagrange elements relied on L2 projection onto an embedding DG space.

This PR implements a Quadrature space to remove the need for the L2 projection. 

Here we reuse the codegen for interpolation into a `VertexOnlyMesh`, namely `compile_expression_dual_evaluation`, which supports interpolation across non-Lagrange elements (as long as the target element does not have derivative degrees of freedom). As a consequence, `prolong`, `restrict` and `inject` are now natively able to transfer accross `FunctionSpaces` on a `MeshHierarchy`, even for the case where the target and souce spaces are defined with different finite elements.

### DIFF
--- a/firedrake/mg/interface.py
+++ b/firedrake/mg/interface.py
@@ -60,7 +60,7 @@ def prolong(coarse, fine):
     next_level = coarse_level * refinements_per_level
 
     if needs_quadrature := not Vf.finat_element.has_pointwise_dual_basis:
-        # Introduce an intermidiate quadrature target space
+        # Introduce an intermediate quadrature target space
         Vf = Vf.quadrature_space()
 
     finest = fine
@@ -130,7 +130,7 @@ def restrict(fine_dual, coarse_dual):
     next_level = fine_level * refinements_per_level
 
     if needs_quadrature := not Vf.finat_element.has_pointwise_dual_basis:
-        # Introduce an intermidiate quadrature source space
+        # Introduce an intermediate quadrature source space
         Vq = Vf.quadrature_space()
 
     coarsest = coarse_dual.zero()
--- a/firedrake/mg/kernels.py
+++ b/firedrake/mg/kernels.py
@@ -105,7 +105,7 @@ def compile_element(operand, dual_arg, parameters=None,
 
     Returns
     -------
-    loopy.TranslationUnit
+    str
         The generated code
     """
     domain = extract_unique_domain(operand)


firedrake/mg/interface.py
```diff
@@ -32,6 +36,23 @@ def check_arguments(coarse, fine, needs_dual=False):
         raise ValueError("Mismatching function space shapes")
 
 
+def get_quadrature_element(V):
```
connorjward: Needs some documentation providing the motivation for this.

firedrake/mg/interface.py
```diff
@@ -1,7 +1,11 @@
 from pyop2 import op2
 
-import firedrake
```
connorjward, line 3: Thanks for undoing this

firedrake/mg/interface.py
```diff
@@ -58,20 +79,26 @@ def prolong(coarse, fine):
     repeat = (fine_level - coarse_level)*refinements_per_level
     next_level = coarse_level * refinements_per_level
 
+    element = get_quadrature_element(Vf)
+    needs_quadrature = element != Vf.ufl_element()
+    if needs_quadrature:
```
connorjward: Isn't `has_pointwise_dual_basis` a much more obvious way to write this?

firedrake/mg/interface.py
```diff
@@ -82,13 +109,17 @@ def prolong(coarse, fine):
...
+        op2.par_loop(kernel, fine.node_set,
+                     fine.dat(op2.WRITE),
                      coarse.dat(op2.READ, fine_to_coarse),
                      node_locations.dat(op2.READ),
                      coarse_coords.dat(op2.READ, fine_to_coarse_coords))
-        coarse = next
-        Vc = Vf
+
+        if needs_quadrature:
+            new_fine = finest if j == repeat-1 else Function(Vc.reconstruct(mesh=meshes[next_level]))
```
connorjward: This logic is quite confusing, and is now happening in two places: here and further up. Can it be made cleaner?

firedrake/mg/kernels.py
```diff
@@ -95,150 +92,97 @@ def to_reference_coordinates(ufl_coordinate_element, parameters=None):
...
+    Parameters
+    ----------
+    operand: ufl.Expr
+        A primal expression
+    dual_arg: ufl.Coargument | ufl.Cofunction
+        A dual argument or coefficient
+
+    Returns
+    -------
+    loopy.TranslationUnit
```
connorjward: This is wrong. This returns a string.

firedrake/mg/kernels.py
```diff
@@ -95,150 +92,97 @@ def to_reference_coordinates(ufl_coordinate_element, parameters=None):
...
-    if coefficient:
-        if expression.ufl_shape:
-            return_variable = gem.Indexed(gem.Variable('R', expression.ufl_shape), tensor_indices)
-            result_arg = lp.GlobalArg("R", dtype=ScalarType, shape=expression.ufl_shape)
-            result = gem.Indexed(result, tensor_indices)
-        else:
-            return_variable = gem.Indexed(gem.Variable('R', (1,)), (0,))
-            result_arg = lp.GlobalArg("R", dtype=ScalarType, shape=(1,))
 
+def make_kernel_args(element, *args):
```
connorjward: Very unclear what this is doing.

connorjward (changes requested):
This is going to need a review from someone with more understanding of the FEM aspects of this.

firedrake/mg/kernels.py
```diff
@@ -95,150 +92,106 @@ def to_reference_coordinates(ufl_coordinate_element, parameters=None):
...
+                                                name="pyop2_kernel_"+name)
+    return lp.generate_code_v2(kernel.ast).device_code()
+
+
+def _make_kernel_args(element, *args):
+    """Returns a string of argument names to call the kernel.
+       Discards coordinate arguments if they do not appear in the kernel."""
+    # NOTE: TSFC will sometimes drop run-time arguments in generated
+    # kernels if they are deemed not-necessary.
+    # For further information, see the same note in interpolation.py.
```
pbrubeck, line 148: @connorjward is this the right thing to do for now?

connorjward: Yeah I think I'm now happy. Having to do this is clearly sub-ideal but I know from experience that this is really hard to do in a nicer way. We do similar things all over the code base.

firedrake/mg/interface.py
```diff
@@ -32,6 +36,24 @@ def check_arguments(coarse, fine, needs_dual=False):
         raise ValueError("Mismatching function space shapes")
 
 
+def get_quadrature_element(V):
```
pbrubeck: This should be implemented as class method, similar to what we did for `FunctionSpace.broken_space()`

firedrake/mg/interface.py
```diff
@@ -58,20 +80,26 @@ def prolong(coarse, fine):
     repeat = (fine_level - coarse_level)*refinements_per_level
     next_level = coarse_level * refinements_per_level
 
+    if needs_quadrature := not Vf.finat_element.has_pointwise_dual_basis:
+        # Introduce an intermidiate quadrature target space
```
JHopeCollins:
```suggestion
        # Introduce an intermediate quadrature target space
```

firedrake/mg/interface.py
```diff
@@ -118,17 +151,25 @@ def restrict(fine_dual, coarse_dual):
     repeat = (fine_level - coarse_level)*refinements_per_level
     next_level = fine_level * refinements_per_level
 
-    meshes = hierarchy._meshes
+    if needs_quadrature := not Vf.finat_element.has_pointwise_dual_basis:
+        # Introduce an intermidiate quadrature source space
```
JHopeCollins:
```suggestion
        # Introduce an intermediate quadrature source space
```

firedrake/mg/interface.py
pbrubeck: Is it better now?

firedrake/mg/interface.py
connorjward: Please don't resolve this if nothing has changed.

firedrake/mg/interface.py
connorjward: Why was this resolved?

connorjward: Never mind. I wasn't looking at the most recent commit.

firedrake/mg/interface.py
connorjward: My bad. Please ignore.

PR #4841, opened 2026-01-27
fix zero-form interpolation

leo-collins:
Zero-form interpolation `interpolate(operand, dual_arg)` was broken if `dual_arg` was a one-form and not a Cofunction.

### DIFF
--- a/tests/firedrake/regression/test_interpolate.py
+++ b/tests/firedrake/regression/test_interpolate.py
@@ -391,7 +391,8 @@ def test_adjoint_dg():
 
 
 @pytest.mark.parametrize("degree", range(1, 4))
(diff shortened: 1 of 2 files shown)


tests/firedrake/regression/test_interpolate.py
```diff
@@ -391,7 +391,8 @@ def test_adjoint_dg():
 
 
 @pytest.mark.parametrize("degree", range(1, 4))
-def test_function_cofunction(degree):
+@pytest.mark.parametrize("cofunc", [True, False])
```
pbrubeck, line 394:
```suggestion
@pytest.mark.parametrize("cofunc", [True, False], ids=["Cofunction", "1-Form"])
```

Issue #4847, opened 2026-01-30
BUG: Interpolation from DPC on extruded mesh fails with TypeError in get_entity_transform


sghelichkhani:
## Description

Interpolating from a DPC element on an extruded mesh fails with:
```
TypeError: 'int' object is not iterable
```

This regression was introduced in PR #4839 (merged 2026-01-29) which extended the use of `compile_expression_dual_evaluation` for interpolation.

## Minimal reproducer

```python
from firedrake import *
mesh1d = CircleManifoldMesh(64, radius=1.22, degree=2)
mesh = ExtrudedMesh(mesh1d, layers=16, extrusion_type="radial")
W = FunctionSpace(mesh, "DPC", 1)
Q1DG = FunctionSpace(mesh, "DQ", 1)
p = Function(W)
p.interpolate(Constant(1.0))
pdg = Function(Q1DG).interpolate(p)  # Crashes here
```

## Error output

```
Traceback (most recent call last):
  File "/opt/firedrake/tsfc/fem.py", line 383, in basis_evaluation
    return self._basis_evaluation_cache[key]
KeyError: (FiniteElement('DPC', TensorProductCell(interval, interval), 1), ...)

During handling of the above exception, another exception occurred:

  ...
  File "/opt/firedrake/tsfc/fem.py", line 386, in basis_evaluation
    val = finat_element.basis_evaluation(
  File "/opt/firedrake-fiat/finat/fiat_elements.py", line 72, in basis_evaluation
    fiat_result = self._element.tabulate(order, ps.points, entity)
  File "/opt/firedrake-fiat/FIAT/finite_element.py", line 196, in tabulate
    transform = self.ref_el.get_entity_transform(entity_dim, entity_id)
  File "/opt/firedrake-fiat/FIAT/reference_element.py", line 1220, in get_entity_transform
    for c, d in zip(self.cells, dim))
TypeError: 'int' object is not iterable
```

## Root cause

In `tsfc/fem.py:385`:
```python
entity = None if entity_id is None else (self.integration_dim, entity_id)
```

For elements on a `TensorProductCell` (like DPC on an extruded mesh), `entity_dim` must be a **tuple** (e.g., `(1, 1)`) matching the cell structure. However, `self.integration_dim` is a scalar integer, causing `get_entity_transform` to fail when it tries to iterate over the dimension.

## Impact

[...]

PR #4848, opened 2026-01-30
Fix interpolation from non-TensorProductCell into TensorProductCell

pbrubeck:
# Description
Fixes https://github.com/firedrakeproject/firedrake/issues/4847

Depends on https://github.com/firedrakeproject/fiat/pull/215

Reverts TSFC entity dimension updates from #4839 in favour of the the finat PR

### DIFF
--- a/firedrake/interpolation.py
+++ b/firedrake/interpolation.py
@@ -885,7 +885,7 @@ def _build_interpolation_callables(
             # expression must be pure function of spatial coordinates so
             # domain has correct ufl cell
             cell = source_mesh.ufl_cell()
-        to_element = create_runtime_quadrature_element(to_element, cell, rt_var_name)
+        to_element = rebuild(to_element, cell, rt_var_name)
 
     cell_set = target_mesh.cell_set
     if subset is not None:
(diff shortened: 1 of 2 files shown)


firedrake/interpolation.py
```diff
@@ -1084,7 +1084,7 @@ def compile_expression(comm, *args, **kwargs):
 
 
 @singledispatch
-def rebuild(element, expr_cell, rt_var_name):
+def create_runtime_quadrature_element(element, expr_cell, rt_var_name):
```
pbrubeck: @connorjward should this live in TSFC?

pbrubeck: `create_element` lives in finat

connorjward: It seems reasonable for this code to live in the FIAT repo

pbrubeck: Ok I'll move it, as we also have a FIAT PR

connorjward (approved):
All looks reasonable. What are your plans for the create element routine?

firedrake/interpolation.py
pbrubeck: No longer doing this, see https://github.com/firedrakeproject/firedrake/pull/4850

Issue #3424, opened 2024-02-26
Is it possible to allow `pathlib.Path` to be passed as argument to `Mesh()`?


p479h:
As of right now, you must provide the file path to your .msh mesh using a python string. Since many users utilize python's built in `pathlib.Path` objects for indicating the mesh directory it would be nice to be able to pass them as arguments. I believe the solution could be added to firedrake/mesh.py `_from_gmsh(filename, comm=None)` as follows:

```
if isinstance(filename, Path):
    filename = filename.as_posix()
```

Or something similar.

connorjward:
This seems like a sensible thing to support. Probably the best place to put this would be inside [Mesh](https://github.com/firedrakeproject/firedrake/blob/master/firedrake/mesh.py#L2578) as your suggestion is specific to gmsh.

ZiyaSelim:
Thank you for the links!
I implemented it and wrote some tests but i cant seem to start the project to test it 😅 
I always get an error when trying to install the dependencies when i use the docker container, where it says i forgot to add `--no-build-isolation` even tho i added it

connorjward:
The issue you are encountering is you are developing code for the `main` branch, which is incompatible with the latest _release_ Firedrake version. If you use the `firedrakeproject/firedrake-vanilla:dev-main` container then you should be fine.

PR #4853, opened 2026-01-31
Accept pathlib.Path objects as filename in Mesh

ZiyaSelim:
# Description
Fixes [#3424](https://github.com/firedrakeproject/firedrake/issues/3424)

This is done by adding a check if the filename passed to the Mesh class is of the `pathlib.Path` type.

Additionally the `test_load_mesh.py` test has be extended to test for passing such a filename to the mesh.

### DIFF
--- a/firedrake/mesh.py
+++ b/firedrake/mesh.py
@@ -38,6 +38,7 @@ from firedrake.logging import info_red
 from firedrake.parameters import parameters
 from firedrake.petsc import PETSc, DEFAULT_PARTITIONER
 from firedrake.adjoint_utils import MeshGeometryMixin
+from firedrake.exceptions import VertexOnlyMeshMissingPointsError, NonUniqueMeshSequenceError
 from pyadjoint import stop_annotating
 import gem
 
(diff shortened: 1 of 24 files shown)


firedrake/mesh.py
```diff
@@ -3340,6 +3340,7 @@ def Mesh(meshfile, **kwargs):
 
     """
     import firedrake.function as function
+    from pathlib import Path
```
connorjward: Please move this import to the top of the file.

tests/firedrake/regression/test_load_mesh.py
```diff
@@ -2,18 +2,25 @@
...
+    return Mesh(join(path, filename))
 
 
 @pytest.mark.parametrize(
-    'filename', ['square.msh', 'square_binary.msh'])
+    'filename', [
+        'square.msh',
+        'square_binary.msh',
+        Path(path) / 'square.msh',
+    ])
```
connorjward, line 22: I think I would have added a new pytest.parametrize for `"use_path", [False, True]`. This could then be passed into `load_mesh(filename, use_path)`. It's a bit confusing to be prepending the path in two places.

connorjward (changes requested):
Thanks for taking this on!

tests/firedrake/regression/test_load_mesh.py
ZiyaSelim: Ah that makes sense thank you 😄

firedrake/mesh.py
ZiyaSelim: Ah that wasnt even necessary as the import for `Path` was already there

PR #4860, opened 2026-02-04
Cache VertexOnlyMesh for cross-mesh interpolation

leo-collins:
Cache the vom used for cross-mesh interpolation. Consider the following

```
V = FunctionSpace(mesh1, "CG", 1)
W = FunctionSpace(mesh2, "CG", 1)

f1 = Function(V)
f2 = Function(V)

h1 = assemble(interpolate(f1, W))  # ~30s at 500,000 dofs
h2 = assemble(interpolate(f2, W))
```

The assembly of `h2` previously took ~5s on my machine, with caching it now takes ~0.012s

### DIFF
--- a/firedrake/interpolation.py
+++ b/firedrake/interpolation.py
@@ -7,6 +7,7 @@ from functools import partial, singledispatch
 from typing import Hashable, Literal, Callable, Iterable
 from dataclasses import asdict, dataclass
 from numbers import Number
+from weakref import WeakValueDictionary
 
 from ufl.algorithms import extract_arguments, replace
 from ufl.domain import extract_unique_domain
@@ -63,6 +64,8 @@ __all__ = (
     "InterpolateOptions",
     "Interpolator"
 )
+
+_vom_cache = WeakValueDictionary()
 
 
 @dataclass(kw_only=True)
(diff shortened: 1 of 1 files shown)


firedrake/interpolation.py
```diff
@@ -462,15 +465,45 @@ def __init__(self, expr: Interpolate):
             # scalar fiat/finat element
             self.dest_element = dest_element
 
-    def _get_symbolic_expressions(self) -> tuple[Interpolate, Interpolate]:
-        """Return the symbolic ``Interpolate`` expressions for point evaluation and
-        re-ordering into the input-ordering VertexOnlyMesh.
+    @staticmethod
+    def _vom_cache_key(target_space, source_mesh, allow_missing_dofs):
```
connorjward, line 469: Can't this just be a cached property?

firedrake/interpolation.py
```diff
@@ -462,15 +465,45 @@ def __init__(self, expr: Interpolate):
...
+        MeshGeometry
+            The VertexOnlyMesh.
+        """
+        key = self._vom_cache_key(
+            self.target_space,
+            self.source_mesh,
+            self.allow_missing_dofs,
+        )
+        try:
+            return _vom_cache[key]
```
connorjward, line 493: I don't really like this. The communicator is only implicitly part of this cache key so I think deadlocks could well happen.
Also you are using `source_mesh` as the key, creating a reference to it and meaning that it will never be cleared from the memory.

connorjward (changes requested):
Sorry for the negativity, but caching correctly is _really_ hard.

firedrake/interpolation.py
leo-collins: What's the correct way to do this then?

firedrake/interpolation.py
leo-collins: Yes

firedrake/interpolation.py
connorjward: Step 1 is choosing an object to tie the lifetime of the VoM to. It's probably best to use `source_mesh` here.

I have implemented a fair amount of caching support that would easily handle this in pyop3, so you may want to wait for that or just port the (small) amount of code over that you need.

connorjward:
@leo-collins should this be closed?

PR #4865, opened 2026-02-05
Use `Index_CreateWithArray` to build mesh spatial index

leo-collins:
Before we were looping over bounding box coords and calling `Index_InsertData` one by one which was a big bottleneck. libspatialindex has a function which can build a spatial index in one go from an array: https://libspatialindex.org/en/latest/doxygen/sidx__api_8cc.html#a36737b1629d8a96458e7f437484f5a84

Building the spatial index in this example went from ~15s on my machine to ~0.9s

```
from firedrake import *

mesh = UnitSquareMesh(500, 500)
mesh.spatial_index
```

### DIFF
--- a/firedrake/mesh.py
+++ b/firedrake/mesh.py
@@ -2671,7 +2671,6 @@ values from f.)"""
         else:
             coords_min, coords_max = self.bounding_box_coords
         tolerance = self.tolerance if hasattr(self, "tolerance") else 0.0
-
         coords_mid = (coords_max + coords_min)/2
         d = np.max(coords_max - coords_min, axis=1)[:, None]
         coords_min = coords_mid - (tolerance + 0.5)*d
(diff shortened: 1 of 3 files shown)


firedrake/mesh.py
```diff
@@ -2776,16 +2782,19 @@ def locate_cells_ref_coords_and_dists(self, xs, tolerance=None, cells_ignore=Non
...
-                                             xs.ctypes.data_as(ctypes.POINTER(ctypes.c_double)),
-                                             Xs.ctypes.data_as(ctypes.POINTER(ctypes.c_double)),
-                                             ref_cell_dists_l1.ctypes.data_as(ctypes.POINTER(ctypes.c_double)),
-                                             cells.ctypes.data_as(ctypes.POINTER(ctypes.c_int)),
-                                             npoints,
-                                             cells_ignore.shape[1],
-                                             cells_ignore)
+        with PETSc.Log.Event("cells_data_as"):
+            cells_data = cells.ctypes.data_as(ctypes.POINTER(ctypes.c_int))
+        with PETSc.Log.Event("ref_cell_dists_l1_as"):
```
connorjward: This seems like an unnecessary number of events to be timing.

firedrake/mesh.py
```diff
@@ -2669,13 +2671,16 @@ def spatial_index(self):
         else:
             coords_min, coords_max = self.bounding_box_coords
         tolerance = self.tolerance if hasattr(self, "tolerance") else 0.0
-        coords_mid = (coords_max + coords_min)/2
-        d = np.max(coords_max - coords_min, axis=1)[:, None]
-        coords_min = coords_mid - (tolerance + 0.5)*d
-        coords_max = coords_mid + (tolerance + 0.5)*d
+
+        with PETSc.Log.Event("spatial_index coords_min_max"):
```
connorjward: This is just timing numpy ops so should be really fast no?

connorjward (commented):
This is fantastic. So often performance improvements come down to "we were doing something quite silly before".

@achanbour this PR is likely of interest to you, if you haven't seen it yet.

firedrake/mesh.py
leo-collins: I will remove these before merging

connorjward: Ah OK. In future if you ask me for a review please tell me if I can ignore bits of it.

connorjward: And you should definitely leave some of them in there

PR #4867, opened 2026-02-06
No tape-able operations at module level in adjoint tests

JHopeCollins:
I was getting some very odd and difficult to reproduce errors with the adjoint tests, for example a test was passing if I only passed that file to pytest, but was failing if I passed the test directory and just selected that test using `-k`.

It seems that if there are tape-able operations at module level in the adjoint tests, they pollute the global state when that test file is _collected_, even if the tests in that module are not run. This still appeared to happen even when I tried to switch off annotations with the module-level fixture, and used `set_working_tape` to wrap all my taped operations!

This PR removes any tape-able operations from module level and stuffs them either inside the tests or inside fixtures which are called on a test-by-test basis when the tests are run (not when they are collected).

I also put the fixtures that manage the taping in the adjoint tests into the `conftest.py`. We still need to "call" them in each adjoint test file by writing file-level fixtures that call the conftest fixtures, but at least all the logic is in one place.

Uses https://github.com/dolfin-adjoint/pyadjoint/pull/245 to slightly simplify the use of `set_working_tape` in the fixtures.

### DIFF
--- a/.github/workflows/core.yml
+++ b/.github/workflows/core.yml
@@ -180,6 +180,10 @@ jobs:
           : # Empty the pip cache to ensure that everything is compiled from scratch
           pip cache purge
 
+          : # Fix for petsc4py+slepc4py build
+          echo 'setuptools<81' > constraints.txt
+          export PIP_CONSTRAINT=constraints.txt
+
           if [ ${{ inputs.target_branch }} = 'release' ]; then
             EXTRA_BUILD_ARGS=''
             EXTRA_PIP_FLAGS=''
@@ -433,6 +437,10 @@ jobs:
 
           : # Empty the pip cache to ensure that everything is compiled from scratch
           pip cache purge
+
+          : # Fix for petsc4py+slepc4py build
+          echo 'setuptools<81' > constraints.txt
+          export PIP_CONSTRAINT=constraints.txt
 
           if [ ${{ inputs.target_branch }} = 'release' ]; then
             EXTRA_PIP_FLAGS=''
--- a/docs/source/install.rst
+++ b/docs/source/install.rst
@@ -240,6 +240,12 @@ install Firedrake. To do this perform the following steps:
       so that it can be detected by mpi4py. See `here <https://mpi4py.readthedocs.io/en/stable/install.html#linux>`__
       for more information.
 
+#. Set ``PIP_CONSTRAINT`` to work around
+   `an issue with setuptools <https://gitlab.com/petsc/petsc/-/merge_requests/9016>`__::
+
+      $ echo 'setuptools<81' > constraints.txt
+      $ export PIP_CONSTRAINT=constraints.txt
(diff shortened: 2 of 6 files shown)


tests/firedrake/conftest.py
```diff
@@ -185,6 +187,28 @@ def finalizer():
...
+def set_test_tape():
+    """Set a new working tape specifically for this test.
+    """
+    with set_working_tape():
+        yield
+    pause_annotation()
+
+
+@pytest.fixture(scope="module")
+def set_module_annotation():
```
connorjward: Is having module scope taping ever a good idea?

JHopeCollins: I do not think so. The previous version of this fixture that was duplicated in each file actually called `continue_annotation` at the beginning of each module. I've changed that to `pause_annotation` here and enabled annotation on a per-test basis in `set_test_tape` (pending agreement on the pyadjoint PR).

connorjward: Ah. I got confused by the name of this fixture. Now this seems to be fairly close to the `check_empty_tape` autouse fixture just above this. I wonder if these could be combined somehow.

I would have thought maybe we'd like to have 2 fixtures, an autouse one like
```py
@pytest.fixture(autouse=True)  # for every test
def assert_untaped():
  assert not annotate_tape()
  yield
  assert not annotate_tape()
```
then one we'd explicitly use for when we want taping
```py
@pytest.fixture
def enable_taping():
  enable_annotation()
  yield
  pause_annotation()
```

connorjward: And creating a new tape each time is also a good idea.

JHopeCollins: I don't know if we can guarantee the order that fixtures of the same scope will be called. I think reducing it to two fixtures is doable, but with:

1.  a module-level fixture called for all modules that asserts that annotation is off at entry and exit
2. a function-level fixture that is auto-used in the adjoint test files (like I've got already) that enables the taping and sets a new per-test tape.

What do you think?

connorjward: The [pytest docs](https://docs.pytest.org/en/7.1.x/how-to/fixtures.html#note-on-finalizer-order) suggest that we can control the order of fixtures.

connorjward: I'm keen to see if we can localise all the tape handling to being per-test.

JHopeCollins: If the CI passes I'll have a go at simplifying the taping fixtures

JHopeCollins: Reduced the fixtures to: a) the old module-scope fixture to assert annotation is off, and b) a function-scope fixture to switch annotation on/off and set a per-test tape.

It all passes locally, we'll see if it passes CI once the setuptools fix is merged and pulled into here.

PR #4871, opened 2026-02-09
Fix for petsc4py/slepc4py+setuptools 81

connorjward:
Fix build failure from setuptools 81 and petsc4py+slepc4py incompatibility. This is already fixed by PETSc and SLEPc but we won't be able to fix `main` until PETSc update their SLEPc branch pointer, and we won't be able to fix `release` until PETSc make a new point release and then we would have to make our own patch release.

I will open an issue reminding us to remove the hack later.

### DIFF
--- a/.github/workflows/core.yml
+++ b/.github/workflows/core.yml
@@ -180,7 +180,7 @@ jobs:
           : # Empty the pip cache to ensure that everything is compiled from scratch
           pip cache purge
 
-          : # Fix for slepc4py build
+          : # Fix for petsc4py+slepc4py build
           echo 'setuptools<81' > constraints.txt
           export PIP_CONSTRAINT=constraints.txt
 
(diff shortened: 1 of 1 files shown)


docs/source/install.rst
```diff
@@ -426,6 +426,16 @@ To install Firedrake with SLEPc support you should:
 
    $ pip install --no-binary h5py 'firedrake[check,slepc]'
 
+.. warning::
+   The ``pip install`` instruction above does not currently work due to
```
JHopeCollins: I would generally prefer warnings to go above the relevant command so it's more likely people will read the warning before trying the command

connorjward: Done

.github/workflows/core.yml
```diff
@@ -180,6 +180,10 @@ jobs:
           : # Empty the pip cache to ensure that everything is compiled from scratch
           pip cache purge
 
+          : # Fix for slepc4py build
```
```suggestion
          : # Fix for petsc4py+slepc4py build
```

.github/workflows/core.yml
```diff
@@ -429,6 +433,10 @@ jobs:
...
```
```suggestion
          : # Fix for petsc4py+slepc4py build
```

JHopeCollins (approved):
LGTM. Thanks Connor

PR #4888, opened 2026-02-14
Fix assign for MixedFunctionSpace and correct 'ancester' typo

### LINKED ISSUE #4887: assign fails for MixedFunctionSpace: MeshSequenceGeometry missing submesh_youngest_common_ancester
## Minimal reproducer

```python
from firedrake import *

mesh = UnitSquareMesh(4, 4)
V, W = VectorFunctionSpace(mesh, "CG", 1), FunctionSpace(mesh, "CG", 1)
Function(MixedFunctionSpace([V, W])).assign(Function(MixedFunctionSpace([V, W])))
```

```
AttributeError: 'MeshSequenceGeometry' object has no attribute 'submesh_youngest_common_ancester'
```

## Context

This was discovered via a failing adjoint taylor test in [G-ADOPT](https://github.com/g-adopt/g-adopt) (see [g-adopt/g-adopt#452](https://github.com/g-adopt/g-adopt/issues/452)). During pyadjoint tape replay, `FunctionAssignBlock.recompute_component` creates a fresh `Function` from the block variable's function space and assigns to it. When the function space is mixed, the target and source end up with distinct `MeshSequenceGeometry` objects wrapping the same underlying mesh.

## Cause

Commit 6a1510cd2 ("Fix multidomain assign", #4849) changed `extract_unique_domain()` calls in `assign.py` to pass `expand_mesh_sequence=False`. This means the returned domain can now be a `MeshSequenceGeometry` (since `MixedFunctionSpace` wraps its mesh in one). The commit added a fast-path `target_mesh is source_mesh` identity check, which works when both functions share the exact same `MixedFunctionSpace`. However, when they don't (distinct but equal `MeshSequenceGeometry` objects), the `is` check fails and falls through to `submesh_youngest_common_ancester()`, which is not defined on `MeshSequenceGeometry`.

Also, `ancester` is misspelt throughout and should be corrected to `ancestor`.

sghelichkhani:
## Summary
- Use equality (`==`) instead of identity (`is`) when comparing meshes in `assign`, so that distinct but equal `MeshSequenceGeometry` objects from separate `MixedFunctionSpace` instances are handled correctly
- Correct the spelling of `ancester` to `ancestor` throughout
- Add regression test for single-mesh mixed function space assign

Fixes #4887
Fixes g-adopt/g-adopt#452

### DIFF
--- a/firedrake/assign.py
+++ b/firedrake/assign.py
@@ -167,9 +167,9 @@ class Assigner:
         elif len(source_meshes) == 1:
             target_mesh = extract_unique_domain(assignee, expand_mesh_sequence=False)
             source_mesh, = source_meshes
-            if target_mesh is source_mesh:
+            if target_mesh == source_mesh:
                 pass
-            elif target_mesh.submesh_youngest_common_ancester(source_mesh) is None:
+            elif target_mesh.submesh_youngest_common_ancestor(source_mesh) is None:
                 raise ValueError(
                     "All functions in the expression must be defined on a single domain "
                     "that is in the same submesh family as domain of the assignee"
--- a/firedrake/mesh.py
+++ b/firedrake/mesh.py
@@ -940,15 +940,15 @@ class AbstractMeshTopology(object, metaclass=abc.ABCMeta):
     # submesh
 
     @utils.cached_property
-    def submesh_ancesters(self):
-        """Tuple of submesh ancesters."""
+    def submesh_ancestors(self):
+        """Tuple of submesh ancestors."""
         if self.submesh_parent:
-            return (self, ) + self.submesh_parent.submesh_ancesters
+            return (self, ) + self.submesh_parent.submesh_ancestors
         else:
             return (self, )
 
-    def submesh_youngest_common_ancester(self, other):
-        """Return the youngest common ancester of self and other.
+    def submesh_youngest_common_ancestor(self, other):
+        """Return the youngest common ancestor of self and other.
 
         Parameters
         ----------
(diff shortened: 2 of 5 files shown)


firedrake/assign.py
```diff
@@ -169,7 +169,7 @@ def __init__(self, assignee, expression, subset=None):
             source_mesh, = source_meshes
             if target_mesh is source_mesh:
```
pbrubeck: Should the right fix here be to check for equality in this line? It'd be preferable not to deal with submesh if none of the meshes are Submesh instances.

firedrake/mesh.py
```diff
@@ -5079,6 +5079,33 @@ def unique(self):
...
+        if isinstance(other, MeshSequenceGeometry):
+            if len(self) != len(other):
+                return None
+            for s, o in zip(self, other):
+                if s.submesh_youngest_common_ancestor(o) is None:
+                    return None
+            return self
+        else:
+            try:
+                return self.unique().submesh_youngest_common_ancestor(other)
```
pbrubeck: If we were to implement this function on a `MeshSequence`, it'd be good to actually support the non-unique case, for which we would have to recurse on the meshes in both `MeshSequences`

firedrake/assign.py
sghelichkhani: Yes you are right, way better fix. Updated the PR to use equality instead.

firedrake/mesh.py
sghelichkhani: Since we went with the equality check, this method is no longer needed so I've removed it from the PR.

tests/firedrake/regression/test_multiple_domains.py
```diff
@@ -219,3 +219,10 @@ def test_multidomain_assign_function(mesh1, mesh3):
     w.assign(z)
     for zsub, wsub in zip(z.subfunctions, w.subfunctions):
         assert np.allclose(zsub.dat.data, wsub.dat.data)
+
+
+def test_single_mesh_mixed_assign():
```
pbrubeck: Thanks for adding this test. I think this one should live in `test_assign.py`

pbrubeck: It'd be good to test that we get the correct values for a nonzero assignment

sghelichkhani: There is no test_assign.py currently so I created a new one. Alternatively test_expressions.py is where most assign tests live, happy to move it there if you prefer.

sghelichkhani: Good point, updated the test to assign nonzero values and check they come through correctly.

pbrubeck (approved):
Thanks for the fix!

PR #4891, opened 2026-02-15
Add per-rank disk checkpointing for adjoint tape

sghelichkhani:
## Motivation

We run time-dependent adjoint Stokes simulations with close to a billion degrees of freedom per timestep. Recomputation-based checkpointing schedules (revolve/binomial) are infeasible due to the cost of recomputing the Stokes solve, so disk checkpointing (`SingleDiskStorageSchedule`) is the only viable option.

Currently, `CheckpointFile` writes all ranks to a single shared HDF5 file via parallel HDF5 (`PETSc.ViewerHDF5` on `COMM_WORLD`). On HPC systems, this means all checkpoint I/O goes through the shared parallel filesystem (Lustre/GPFS), which becomes a severe bottleneck. Under 24-hour job time limits, the disk I/O overhead makes simulations that comfortably fit in memory-checkpointed wall time infeasible when switching to disk checkpointing.

HPC nodes typically have fast node-local NVMe/SSD storage that is orders of magnitude faster than the shared filesystem. However, the current collective I/O approach in `CheckpointFile` cannot use node-local storage because all ranks must access the same file path.

## Approach

Following @connorjward's suggestion in https://github.com/firedrakeproject/firedrake/pull/4891#issuecomment-3909700728, the implementation uses a general `checkpoint_comm` parameter rather than a hardcoded per-rank approach. Users pass any MPI communicator to control how function data is checkpointed:

```python
# Per-rank files (each rank writes independently to node-local storage)
enable_disk_checkpointing(checkpoint_comm=MPI.COMM_SELF,
                          checkpoint_dir="/local/scratch")

# Per-node files (ranks on the same node share a file)
node_comm = MPI.COMM_WORLD.Split_type(MPI.COMM_TYPE_SHARED)
enable_disk_checkpointing(checkpoint_comm=node_comm,
                          checkpoint_dir="/local/scratch")
```

The function data is written using `PETSc.Vec.createMPI` + `ViewerHDF5` on the supplied communicator, bypassing `CheckpointFile` and its collective `globalVectorView`/`globalVectorLoad` on `COMM_WORLD`. We tried using `CheckpointFile` directly with a sub-communicator (see https://github.com/firedrakeproject/firedrake/pull/4891#issuecomment-3912097794), but loading deadlocks because the mesh DM's `sectionLoad`/`globalVectorLoad` are collective on `COMM_WORLD`.

The mesh checkpoint via `checkpointable_mesh` still uses shared storage through `CheckpointFile` since that's a one-time operation and not performance-critical. Fully backwards compatible: without `checkpoint_comm`, behaviour is unchanged.

## Multi-mesh considerations

Functions on different meshes with different partitioning work correctly because Vec dataset names include the mesh name and element info (`ckpt_mesh_a_CG2` vs `ckpt_mesh_b_DG1`), and `checkpointable_mesh` ensures deterministic partitioning per mesh independently.

The supermesh projection across two different meshes still fails in parallel, but that's a pre-existing limitation unrelated to this PR.

## Testing

11 tests total covering three checkpointing modes:

Existing shared-mode tests (5): serial and parallel basic checkpointing, successive writes, timestepper with taylor_test, and boundary conditions.

`checkpoint_comm` with `COMM_SELF` (3): parallel basic checkpointing, successive writes (serial), and multi-mesh parallel. These exercise the per-rank file path where each rank writes independently.

`checkpoint_comm` with node communicator (3): parallel basic checkpointing, multi-mesh parallel, and timestepper with taylor_test. These exercise the multi-rank-per-file path using `COMM_TYPE_SHARED`.

### DIFF
--- a/firedrake/adjoint_utils/checkpointing.py
+++ b/firedrake/adjoint_utils/checkpointing.py
@@ -149,15 +149,13 @@ class CheckPointFileReference:
     """A filename which deletes the associated file when it is destroyed."""
     def __init__(self, name, comm, cleanup=False, checkpoint_comm=None):
         self.name = name
+        self.comm = comm
         self.cleanup = cleanup
         self.checkpoint_comm = checkpoint_comm
-        # Resolve the right comm for file deletion once at construction time so
-        # __del__ does not need to branch on checkpoint_comm.
-        self._cleanup_comm = checkpoint_comm if checkpoint_comm is not None else comm
 
     def __del__(self):
         if self.cleanup and os.path.exists(self.name):
-            if self._cleanup_comm.rank == 0:
+            if self.comm.rank == 0:
                 os.remove(self.name)
         # Prune the index-tracking entry for this file from CheckpointFunction.
         # This is safe for the following reasons:
@@ -281,7 +279,7 @@ class DiskCheckpointer(TapePackageData):
         # file.
         with TemporaryFunctionCheckpointFile(self.checkpoint_comm, filepath, 'w'):
             pass
-        return CheckPointFileReference(filepath, self.comm, self.cleanup,
+        return CheckPointFileReference(filepath, self.checkpoint_comm, self.cleanup,
                                        checkpoint_comm=self.checkpoint_comm)
 
     def _new_checkpoint_file(self):
@@ -374,9 +372,9 @@ def checkpointable_mesh(mesh):
             "No current checkpoint file. Call enable_disk_checkpointing()."
         )
 
-    with CheckpointFile(checkpoint_file.name, 'a') as outfile:
+    with CheckpointFile(checkpoint_file.name, 'a', comm=checkpoint_file.comm) as outfile:
         outfile.save_mesh(mesh)
-    with CheckpointFile(checkpoint_file.name, 'r') as outfile:
+    with CheckpointFile(checkpoint_file.name, 'r', comm=checkpoint_file.comm) as outfile:
         return outfile.load_mesh(mesh.name)
 
 


connorjward:
In PyOP2 we have the notion of a 'compilation comm' which is a communicator defined over each node (https://github.com/firedrakeproject/firedrake/blob/main/pyop2/mpi.py#L450). Might something like this be appropriate/more general here?

sghelichkhani:
@connorjward Following up on your suggestion about using the compilation communicator. Angus and I had a discussion about this and we tried to see if we can simply do `CheckpointFile(fname, mode, comm=compilation_comm)` and use the standard `save_function`/`load_function` path with a node-level communicator.

Saving works fine, but loading deadlocks. Here's a minimal reproducer:

```python
"""mpiexec -n 4 python test_subcomm_checkpoint.py"""
import os
import tempfile
from firedrake import *
from firedrake.checkpointing import CheckpointFile

comm = COMM_WORLD
mesh = UnitSquareMesh(4, 4)
f = Function(V, name="f")
f.interpolate(SpatialCoordinate(mesh)[0])

node_comm = comm.Split(color=comm.rank // 2, key=comm.rank)

if comm.rank == 0:
    tmpdir = tempfile.mkdtemp()
else:
    tmpdir = None
tmpdir = comm.bcast(tmpdir, root=0)
fname = os.path.join(tmpdir, f"node{comm.rank // 2}.h5")

with CheckpointFile(fname, 'w', comm=node_comm) as out:
    out.save_mesh(mesh)
    out.save_function(f)

with CheckpointFile(fname, 'r', comm=node_comm) as inp:
    mesh2 = inp.load_mesh()
    f2 = inp.load_function(mesh2, "f")  # deadlocks here
```

The issue is that `sectionLoad` and `globalVectorLoad` are collective operations on the `topology_dm`, which lives on `COMM_WORLD`. When only a subset of ranks (the node sub-comm) enter the load call, the collective never completes. I think this is by design rather than a bug in CheckpointFile, since the contract is that the mesh lives on `COMM_WORLD` for the parallel solves, and the load path assumes the viewer and the mesh DM share a compatible communicator. Making this work with a sub-communicator would require changes both in PETSc (cross-comm DM operations) and in Firedrake's checkpointing internals.

That said, I might be missing something. Is there a way to make this work that I'm not seeing?

connorjward:
> That said, I might be missing something. Is there a way to make this work that I'm not seeing?

I don't think so. I'd have been surprised had that worked.

> That means even with a node-local comm, you'd still need to bypass CheckpointFile and use raw PETSc Vec I/O, just with a parallel Vec on the node comm instead of a sequential Vec on COMM_SELF.

This is what I'm suggesting. It doesn't seem like a lot of work to change your API from `per_rank`-type options to a more general `filesystem_comm=some_comm` type of thing. Your use case could still do `filesystem_comm=COMM_SELF` if desired.

sghelichkhani:
Thanks Connor, done. I've refactored the API from `per_rank_dirname` to a general `checkpoint_comm` + `checkpoint_dir` interface. Users can now pass any communicator, so `COMM_SELF` for per-rank files or a node-local comm via `COMM_TYPE_SHARED` for per-node files.

The main PETSc-level change this required was switching from `Vec.createSeq` (which only works on a single process) to `Vec.createMPI((local_size, PETSc.DECIDE), comm=checkpoint_comm)`. The `DECIDE` for the global size is important here because when `checkpoint_comm` groups multiple ranks (e.g. a node communicator), each rank contributes a different number of local DOFs depending on how the mesh was partitioned. The global size of the Vec on the checkpoint communicator is not something we know upfront since it's the sum of local DOFs across ranks in that sub-comm, which is a different grouping than the mesh's COMM_WORLD partitioning. Letting PETSc compute the global size from the local sizes avoids us having to gather that information ourselves.

I've also added tests with `COMM_TYPE_SHARED` to exercise the multi-rank-per-file path and trimmed redundant serial tests that were fully covered by their parallel counterparts.

firedrake/adjoint_utils/checkpointing.py
```diff
@@ -182,6 +240,36 @@ def new_checkpoint_file(self):
...
+        """Set up a checkpoint file for function data."""
+        if self.checkpoint_comm is not None:
+            return self._new_checkpoint_comm_file()
+        else:
+            return self._new_shared_checkpoint_file()
+
+    # Keep the old name as an alias for backwards compatibility.
+    def new_checkpoint_file(self):
+        """Set up a disk checkpointing file."""
+        return self._new_checkpoint_file()
```
connorjward:
```suggestion
    def new_checkpoint_file(self):
        """Set up a disk checkpointing file."""
        warnings.warn("'new_checkpoint_file' is deprecated, use '_new_checkpoint_file' instead",
                      FutureWarning)
        return self._new_checkpoint_file()
```

tests/firedrake/adjoint/test_disk_checkpointing.py
```diff
@@ -202,3 +214,273 @@ def test_bcs():
...
+# --- node_comm disk checkpointing tests ---
+# These test the checkpoint_comm option with a shared-memory communicator
+# obtained via MPI.COMM_TYPE_SHARED. On a single machine all ranks share
+# one communicator, exercising the multi-rank createMPI + parallel HDF5
+# path where ranks collectively write to a shared checkpoint file.
+
+
+def _node_comm():
+    """Return a communicator grouping ranks that share physical memory."""
+    return MPI.COMM_WORLD.Split_type(MPI.COMM_TYPE_SHARED)
```
connorjward: Why not something like `node_comm = comm.Split(color=comm.rank // 2, key=comm.rank)`. We should be sure that the test is for a comm with size < COMM_WORLD but > 1.

firedrake/adjoint_utils/checkpointing.py
```diff
@@ -272,6 +360,22 @@ def restore(self):
         pass
 
 
+def _generate_checkpoint_vec_name(function):
+    """Generate a unique name for checkpoint-comm Vec storage based on function space."""
+    V = function.function_space()
+    mesh = V.mesh()
+    # Handle MeshSequenceGeometry (mesh hierarchies) by getting the
+    # underlying MeshGeometry.
+    if hasattr(mesh, '_meshes'):
```
connorjward: Could we use an `isinstance` check here?

firedrake/adjoint_utils/checkpointing.py
```diff
@@ -272,6 +360,22 @@ def restore(self):
         pass
 
 
+def _generate_checkpoint_vec_name(function):
```
connorjward: I'm surprised you need this function. Doesn't something like this already exist? In general I'm very wary of introducing things that look like cache keys.

firedrake/adjoint_utils/checkpointing.py
sghelichkhani: Done, added a FutureWarning.

tests/firedrake/adjoint/test_disk_checkpointing.py
sghelichkhani: Good point. Replaced with comm.Split(color=rank // 2) which gives groups of size 2 and 1 with nprocs=3, so it's genuinely between COMM_SELF and COMM_WORLD.

firedrake/adjoint_utils/checkpointing.py
sghelichkhani: Switched to isinstance(mesh, ufl.MeshSequence).

firedrake/adjoint_utils/checkpointing.py
sghelichkhani: Good point. I was initially reluctant to touch CheckpointFile itself, which is why I had a separate function. But on second thought, the existing `CheckpointFile._generate_function_space_name` does exactly what I need and doesn't actually depend on `self`. I've extracted it into a module-level function in `firedrake/checkpointing.py` that both the CheckpointFile method and the checkpoint_comm path now call. Also added MeshSequenceGeometry handling inside the loop to cover the multiple meshes case.

sghelichkhani:
Simplified the multi-mesh tests to fix intermittent CI failures. The original design chained two solves via a global reduction (assemble(u_a * dx)), which amplified parallel floating-point non-determinism across tape replays. The two solves are now independent while still exercising multi-mesh checkpointing.

tests/firedrake/adjoint/test_disk_checkpointing.py
```diff
@@ -3,9 +3,21 @@
...
+import tempfile
+
+
+@pytest.fixture
+def set_test_tape():
+    """Set a new working tape for this test."""
+    continue_annotation()
+    with set_working_tape():
+        yield
+    pause_annotation()
```
JHopeCollins: This is done by the `autouse_test_taping` fixture below. See https://github.com/firedrakeproject/firedrake/pull/4867
```suggestion
```

firedrake/checkpointing.py
```diff
@@ -1403,24 +1434,7 @@ def _generate_mesh_key_from_names(self, mesh_name, distribution_name, permutatio
 
     def _generate_function_space_name(self, V):
```
JHopeCollins: Does this need to be a method anymore if it just forwards onto the free function? It's internal API so it should be fine just to replace any instance of `self._generate_function_space_name` with the free function.

JHopeCollins (commented):
My gut feeling is that this should be part of `CheckpointFile` rather than hidden in the adjoint utils.

I don't think I fully understand the problem with checkpointing the mesh using the global comm but checkpointing the function data using a subcomm. Why does that mean you have to split the DM into multiple files?

firedrake/adjoint_utils/checkpointing.py
JHopeCollins: Should we be telling people to use the "private" API? Can this warning explain how to update code using the old `new_checkpoint_file()` just with the public API?

firedrake/adjoint_utils/checkpointing.py
```diff
@@ -316,15 +417,79 @@ def __init__(self, function):
             outfile.save_function(function, name=self.stored_name,
                                   idx=self.stored_index)
 
+    def _save_checkpoint_comm(self, function):
```
JHopeCollins: Minor gripe, but this sounds like you are checkpointing the comm. Something like `_save_local_checkpoint` and `_save_shared_checkpoint` (and equivalently for the `_restore_*` methods) would be clearer.

JHopeCollins:
What about  `dm.localVectorView` and `dm.localVectorLoad`? It looks like they will handle splitting the data apart which seem to do what we need "through the front door". I don't know if they can save the local data to local files though.
https://petsc.org/release/manualpages/DMPlex/DMPlexLocalVectorView/
https://petsc.org/release/manualpages/DMPlex/DMPlexLocalVectorLoad/

My other two suggestions would be:

1. Using a python-type viewer context in the `CheckpointFile` that internally creates the current global HDF5 viewer and delegates to that one _except_ if it's being asked to view/load a `Vec`, in which case it uses the new logic in the adjoint_utils for saving/viewing locally. So `dm.globalVector{View,Load}` would end up being diverted to the comm-local implementation.

2. A new class, something like `LocalFunctionCheckpointFile` (happy to bikeshed this) that has the logic for saving/loading Functions to local files on subcomms. It would obviously need to be very explicit about the restrictions of this (you can't save/load the mesh, and you must have _exactly_ the same partition for saving and viewing so is basically only for saving/loading during the same programme).

The second might be quicker, it would just be lifting out the new logic from `CheckpointFunction` into a new class. Potentially we don't advertise the new class in the public API if we don't trust users to use it properly, but it would mean that `CheckpointFunction` doesn't now have to be thinking about the function data in some instances but not others.

Saying that, I don't know if the first would actually be that much work. It should hopefully just be implementing `__init__`, `view` and `load` methods in the python context (and maybe a blank `setUp`).

The main thing is that I would like to see the local save/load logic abstracted out of the adjoint code somehow, because the adjoint code really shouldn't be thinking about actual concrete data.

JHopeCollins:
> I quite like something like `EphemeralFunctionCheckpointFile` or similar.

How about `TemporaryCheckpointFile`, or `TemporaryFunctionCheckpointFile`, to mirror the `tempfile` naming and the common naming of a `/tmp` working directory.

>It would be cool if by default they could be destroyed at program exit to avoid misuse.

@sghelichkhani is it possible to use `tempfile.TemporaryFile` and `tempfile.TemporaryDirectory` rather than `tempfile.mkstemp` and `tempfile.mkdtemp` so that the cleanup is automated?

sghelichkhani:
All review comments from @connorjward and @JHopeCollins are addressed. The main change is extracting the PETSc Vec I/O into a `TemporaryFunctionCheckpointFile` class in `checkpointing.py` as we discussed. The adjoint module no longer imports `petsc4py` and just delegates to `save_function`/`load_function`. Directory cleanup uses `tempfile.TemporaryDirectory` so no `atexit` needed on that path.

The `CheckpointFile._generate_function_space_name` forwarding method is removed entirely, all call sites call the free function directly. The free function adds a `MeshSequenceGeometry` unwrap but this is a no-op on the `CheckpointFile` path since `save_function` already calls `mesh.unique()` upstream, it only matters for the new `TemporaryFunctionCheckpointFile` path. Methods renamed to `_save_local_checkpoint`/`_restore_local_checkpoint`, deprecation warning no longer references private API, dead `checkpoint_comm` attribute removed from `CheckPointFileReference`, and there is now a `UserWarning` when `checkpoint_comm` is used without `checkpoint_dir`.

Tests cleaned up: local `set_test_tape` fixture removed, serial test uses `tmp_path`, parallel tests use `try/finally` for cleanup, and `taylor_test` added to the multi-mesh test.

Two pre-existing items I looked at but decided not to change: `_checkpoint_indices` is never pruned (slow leak, but pruning risks removing entries before `restore()` reads them), and `TemporaryDirectory` cleanup depends on GC timing relative to `MPI_Finalize` (harmless in practice since node-local scratch is wiped between jobs).

firedrake/checkpointing.py
```diff
@@ -504,6 +505,209 @@ def __del__(self):
...
+
+    This class encapsulates the low-level PETSc Vec I/O that enables function
+    checkpointing on a sub-communicator (e.g. per-rank or per-node), bypassing
+    the parallel HDF5 overhead of :class:`CheckpointFile`.
+
+    Parameters
+    ----------
+    comm : mpi4py.MPI.Intracomm
+        The communicator on which checkpoint files are collectively created
+        and accessed.
```
connorjward:
```suggestion
        and accessed. Note that this does not have to be the same communicator that the function is defined over.
```

tests/firedrake/adjoint/test_disk_checkpointing.py
```diff
@@ -202,3 +206,268 @@ def test_bcs():
...
+# HDF5 and enabling node-local storage on HPC systems.
+
+
+@pytest.mark.skipcomplex
+@pytest.mark.parallel(nprocs=3)
+def test_checkpoint_comm_disk_checkpointing_parallel():
+    set_working_tape(Tape())
+    tape = get_working_tape()
+    tape.clear_tape()
+    continue_annotation()
```
connorjward: I think that this isn't needed as per https://github.com/firedrakeproject/firedrake/pull/4891#discussion_r2827047567

tests/firedrake/adjoint/test_disk_checkpointing.py
```diff
@@ -202,3 +206,268 @@ def test_bcs():
...
+@pytest.mark.skipcomplex
+@pytest.mark.parallel(nprocs=3)
+def test_sub_comm_disk_checkpointing_parallel():
+    """Test disk checkpointing with a multi-rank sub-communicator."""
+    sub_comm = _sub_comm()
+    set_working_tape(Tape())
+    tape = get_working_tape()
+    tape.clear_tape()
+    continue_annotation()
+    tmpdir = _broadcast_tmpdir(MPI.COMM_WORLD)
```
connorjward, line 354: shouldn't this be `sub_comm`?

firedrake/checkpointing.py
```diff
@@ -504,6 +505,209 @@ def __del__(self):
...
+    def remove_file(self, filepath):
+        """Remove a checkpoint file.
+
+        Parameters
+        ----------
+        filepath : str
+            Path to the file to remove. Only rank 0 performs the deletion,
+            and only when cleanup is enabled.
+        """
+        if self.cleanup and self.comm.rank == 0 and os.path.exists(filepath):
```
connorjward: This seems a bit odd. I would expect this function, given its name, to alway remove the file.

firedrake/adjoint_utils/checkpointing.py
```diff
@@ -120,14 +147,18 @@ def __exit__(self, *args):
...
         self.comm = comm
         self.cleanup = cleanup
+        self.temp_ckpt = temp_ckpt
 
     def __del__(self):
-        if self.cleanup and self.comm.rank == 0 and os.path.exists(self.name):
-            os.remove(self.name)
+        if self.cleanup and os.path.exists(self.name):
+            if self.temp_ckpt is not None:
+                self.temp_ckpt.remove_file(self.name)
```
connorjward: See my other comment. You're doing the `if self.cleanup` logic twice

firedrake/adjoint_utils/checkpointing.py
```diff
@@ -182,6 +242,30 @@ def new_checkpoint_file(self):
         return CheckPointFileReference(checkpoint_file, self.comm,
                                        self.cleanup)
 
+    def _new_checkpoint_comm_file(self):
+        """Set up a checkpoint file on the checkpoint communicator."""
+        filepath = self._temp_ckpt.new_file()
+        return CheckPointFileReference(filepath, self.comm,
+                                       self.cleanup,
+                                       temp_ckpt=self._temp_ckpt)
```
connorjward: The API is a little weird here. Would it be better to have a new `CheckpointFileReference` subclass?

firedrake/checkpointing.py
sghelichkhani: Done.

tests/firedrake/adjoint/test_disk_checkpointing.py
sghelichkhani: Done, removed the manual tape setup from all four new parallel tests. The autouse_test_taping fixture handles it. The pre-existing tests (test_disk_checkpointing and test_disk_checkpointing_parallel) also have the same redundant setup from before #4867, but I left those untouched to keep the diff focused on the checkpoint_comm changes.

tests/firedrake/adjoint/test_disk_checkpointing.py
sghelichkhani: COMM_WORLD is intentional here. The tmpdir is passed as checkpoint_dir, which is a parent directory. Each sub-comm's TemporaryFunctionCheckpointFile creates its own subdirectory inside it, so isolation between sub-comms is automatic. Using COMM_WORLD means all ranks share one parent and rank 0 can clean everything up in the finally block. If we used sub_comm, each group would get a different parent dir and the cleanup would only reach one of them. Added a docstring to _broadcast_tmpdir explaining this.

firedrake/checkpointing.py
sghelichkhani: Good catch. Removed the self.cleanup guard from remove_file so it always removes when called. The cleanup decision now lives solely in CheckPointFileReference.__del__, which is the only caller.

firedrake/adjoint_utils/checkpointing.py
sghelichkhani: Fixed, see above. The cleanup check is now only in CheckPointFileReference.__del__.

firedrake/adjoint_utils/checkpointing.py
sghelichkhani: I see what you mean about the dual personality. A subclass would clean up the __del__ dispatch nicely. The thing I'm less sure about is CheckpointFunction, which is where the more important branching happens. The shared path opens a CheckpointFile context manager and calls save_function/load_function on it, while the local path calls temp_ckpt.save_function/load_function directly, with quite different signatures and semantics. With a subclass, the dispatch in CheckpointFunction would become isinstance + the same temp_ckpt attribute access, which doesn't feel like a clear win over the current `temp_ckpt is not None` check.

To truly benefit from polymorphism we'd need to push save/load methods into the file reference hierarchy so CheckpointFunction can just call self.file.save(...) uniformly. That would be genuinely cleaner but it's a bigger refactor that mixes file lifecycle with I/O logic, probably better tackled separately if the API grows.

Happy to do the subclass if you feel strongly about it, but my preference would be to leave the current flat design for now and revisit if we end up adding more modes.

connorjward: > That would be genuinely cleaner but it's a bigger refactor that mixes file lifecycle with I/O logic, probably better tackled separately if the API grows.

Sure. This can be avoided for now.

sghelichkhani:
CI failures are unrelated: timeout in nprocs=6 I/O tests and a Gusto smoke test options issue. No files from this PR are involved.

firedrake/adjoint_utils/checkpointing.py
```diff
@@ -120,14 +147,18 @@ def __exit__(self, *args):
...
+        self.temp_ckpt = temp_ckpt
 
     def __del__(self):
-        if self.cleanup and self.comm.rank == 0 and os.path.exists(self.name):
-            os.remove(self.name)
+        if self.cleanup and os.path.exists(self.name):
+            if self.temp_ckpt is not None:
+                self.temp_ckpt.remove_file(self.name)
+            elif self.comm.rank == 0:
+                os.remove(self.name)
```
JHopeCollins, line 159: Why not just pass the `checkpoint_comm` rather than `mesh.comm` then you don't have to mess with this logic at all and `CheckPointFileReference` doesn't need to know what type of checkpoint file it is referencing.

firedrake/checkpointing.py
```diff
@@ -504,6 +505,208 @@ def __del__(self):
...
+                    prefix="firedrake_adjoint_checkpoint_cc_",
+                    dir=dirname or os.getcwd()
+                )
+                path = self._tmpdir.name
+            else:
+                path = tempfile.mkdtemp(
+                    prefix="firedrake_adjoint_checkpoint_cc_",
+                    dir=dirname or os.getcwd()
+                )
+                self._tmpdir = None
```
JHopeCollins: I think that the `delete` kwarg will deal with this for you.
```suggestion
            self._tmpdir = tempfile.TemporaryDirectory(
                prefix="firedrake_adjoint_checkpoint_cc_",
                dir=dirname or os.getcwd(),
                delete=self.cleanup
            )
            path = self._tmpdir.name
            ```

File: firedrake/checkpointing.py [hunk 16]
```diff
@@ -504,6 +505,208 @@ def __del__(self):
...
+            self.dirname = comm.bcast(path)
+        else:
+            self._tmpdir = None
+            self.dirname = comm.bcast(None)
```
JHopeCollins: Personally I strongly prefer having all ranks call a collective operation from the same line of code. I find that it just makes it easier to reason about.
```suggestion
        else:
            path = None
        self.dirname = comm.bcast(path)
```

firedrake/checkpointing.py
```diff
@@ -504,6 +505,208 @@ def __del__(self):
...
+        if filepath not in self._indices:
+            self._indices[filepath] = {}
+
+        stored_name = _generate_function_space_name(function.function_space())
+        indices = self._indices[filepath]
+        indices.setdefault(stored_name, 0)
+        indices[stored_name] += 1
+        stored_index = indices[stored_name]
+
+        vec_name = f"{stored_name}_{stored_index}"
```
JHopeCollins: This logic should all be in the adjoint code, it isn't actually related to the fact you're saving in a temp file.

firedrake/checkpointing.py
```diff
@@ -504,6 +505,208 @@ def __del__(self):
...
+
+    def __init__(self, comm, dirname=None, cleanup=True):
+        self.comm = comm
+        self.cleanup = cleanup
+        self._indices = {}
+
+        if comm.rank == 0:
+            if cleanup:
+                self._tmpdir = tempfile.TemporaryDirectory(
+                    prefix="firedrake_adjoint_checkpoint_cc_",
```
JHopeCollins: This should be a kwarg to init that defaults to something more generic like `"firedrake_temp_checkpointing_cc_"`.
The adjoint code can then pass a name with `"adjoint"` in if it wants to.

firedrake/checkpointing.py
```diff
@@ -504,6 +505,208 @@ def __del__(self):
...
+        V_names.append("_".join([mesh.name, elem_name]))
+    return "_".join(V_names)
+
+
+class TemporaryFunctionCheckpointFile:
+    """Manages temporary HDF5 files for saving/loading function data via PETSc Vec I/O.
+
+    This class encapsulates the low-level PETSc Vec I/O that enables function
+    checkpointing on a sub-communicator (e.g. per-rank or per-node), bypassing
+    the parallel HDF5 overhead of :class:`CheckpointFile`.
```
JHopeCollins: Docstring should say what not how.
```suggestion
    """Checkpoint :class:`.Function`s in temporary HDF5 files in local storage.

    This class encapsulates checkpointing on a sub-communicator (e.g. per-rank or 
    per-node), which allows saving to local storage (e.g. node-local scratch), which
    bypasses the overhead of writing in parallel to shared storage with
    :class:`CheckpointFile`.
```

firedrake/checkpointing.py
```diff
@@ -504,6 +505,208 @@ def __del__(self):
...
+        viewer.destroy()
+        return filepath
+
+    def save_function(self, filepath, function):
+        """Save a function's data to an HDF5 file via PETSc Vec I/O.
+
+        Parameters
+        ----------
+        filepath : str
+            Path to the HDF5 file (must have been created by :meth:`new_file`).
```
JHopeCollins: Why do you require this? Why not have one file per `TemporaryCheckpointFunctionFile` like we do for `CheckpointFile`? I think this would clean up a lot of the interface.

firedrake/checkpointing.py
```diff
@@ -504,6 +505,208 @@ def __del__(self):
...
+            )
+            os.close(fd)
+            filepath = self.comm.bcast(filepath)
+        else:
+            filepath = self.comm.bcast(None)
+
+        viewer = PETSc.ViewerHDF5()
+        viewer.create(filepath, mode='w', comm=self.comm)
+        viewer.destroy()
+        return filepath
```
JHopeCollins: If you have a one-to-one with this class and a file then you can save `self.filepath` and also keep hold of the viewer so it isn't recreated every time we save/load.

firedrake/checkpointing.py
```diff
@@ -504,6 +505,208 @@ def __del__(self):
...
+        vec_name = f"{stored_name}_{stored_index}"
+
+        with function.dat.vec_ro as v:
+            local_array = v.getArray().copy()
+
+        local_vec = PETSc.Vec().createMPI(
+            (len(local_array), PETSc.DECIDE), comm=self.comm
+        )
+        local_vec.setName(vec_name)
+        local_vec.setArray(local_array)
```
JHopeCollins: You can use `PETSc.Vec().createWithArray()` for this.

firedrake/checkpointing.py
```diff
@@ -504,6 +505,208 @@ def __del__(self):
...
+        function_space : FunctionSpace
+            The function space for the loaded function.
+        stored_name : str
+            The stored name returned by :meth:`save_function`.
+        stored_index : int
+            The stored index returned by :meth:`save_function`.
+        name : str or None
+            Name for the new function.
+        count : int or None
+            Count identity for the new function.
```
JHopeCollins: Try to follow `CheckpointFunction` API as much as possible. renaming/recounting the loaded function is not the responsibility of `TemporaryCheckpointFile`, leave it to the caller.

firedrake/checkpointing.py
```diff
@@ -504,6 +505,208 @@ def __del__(self):
...
+            filepath = self.comm.bcast(filepath)
+        else:
+            filepath = self.comm.bcast(None)
+
+        viewer = PETSc.ViewerHDF5()
+        viewer.create(filepath, mode='w', comm=self.comm)
+        viewer.destroy()
+        return filepath
+
+    def save_function(self, filepath, function):
```
JHopeCollins: Can you just use the same signature as `CheckpointFile`, assuming you have one file per instance of this class?
```suggestion
    def save_function(self, function, name=None, idx=None):
```

firedrake/checkpointing.py
```diff
@@ -504,6 +505,208 @@ def __del__(self):
...
+        viewer = PETSc.ViewerHDF5()
+        viewer.create(filepath, mode='a', comm=self.comm)
+        local_vec.view(viewer)
+        viewer.destroy()
+        local_vec.destroy()
+
+        return stored_name, stored_index
+
+    def load_function(self, filepath, function_space, stored_name, stored_index,
+                      name=None, count=None):
```
JHopeCollins: If there's a one-to-one with hdf5 file and class instance, and you leave renaming/recounting to the caller, then you can have almost the same signature as `CheckpointFile`.
```suggestion
    def load_function(self, function_space, name, idx=None):
```

firedrake/checkpointing.py
```diff
@@ -504,6 +505,208 @@ def __del__(self):
...
+            local_size = v.getLocalSize()
+            local_vec = PETSc.Vec().createMPI(
+                (local_size, PETSc.DECIDE), comm=self.comm
+            )
+            local_vec.setName(vec_name)
+
+            viewer = PETSc.ViewerHDF5()
+            viewer.create(filepath, mode='r', comm=self.comm)
+            local_vec.load(viewer)
+            viewer.destroy()
```
JHopeCollins: Stash `self.viewer` rather than recreating each time.

firedrake/adjoint_utils/checkpointing.py
```diff
@@ -182,6 +242,30 @@ def new_checkpoint_file(self):
...
+        else:
+            return self._new_shared_checkpoint_file()
+
+    def new_checkpoint_file(self):
+        """Set up a disk checkpointing file."""
+        warnings.warn(
+            "'new_checkpoint_file' is deprecated. Checkpoint file management "
+            "is now handled internally and this method will be removed in a "
+            "future release.",
+            FutureWarning
```
JHopeCollins, line 299: How should someone who has been using `new_checkpoint_file` update their code to use the undeprecated public API?

firedrake/adjoint_utils/checkpointing.py
```diff
@@ -300,13 +383,23 @@ def __init__(self, function):
                 "No current checkpoint file. Call enable_disk_checkpointing()."
             )
 
+        self.count = function.count()
+
+        if self.file.temp_ckpt is not None:
+            self._function_space = function.function_space()
+            self._save_local_checkpoint(function)
+        else:
+            self._save_shared(function)
```
JHopeCollins: We should be as consistent as possible with naming. Either `self._save_local_checkpoint` and `self._save_shared_checkpoint` or `self._save_local` and `self._save_shared`. Same for the `restore` methods.

firedrake/adjoint_utils/checkpointing.py
JHopeCollins: All of this logic about the name and index is only for adjoint checkpoints, and is the same for both shared and temporary checkpoints. It should be only in the adjoint code, and in one place no matter what checkpointing strategy.

If you make the `TemporaryCheckpointFile.save_function` signature identical to the `CheckpointFile.save_function` signature then you won't even need a branch.

firedrake/adjoint_utils/checkpointing.py
```diff
@@ -316,15 +409,36 @@ def __init__(self, function):
...
+
     def restore(self):
         """Read and return this Function from the checkpoint."""
+        if self.file.temp_ckpt is not None:
+            return self._restore_local_checkpoint()
+        else:
+            return self._restore_shared()
+
+    def _restore_shared(self):
+        """Load function data from a shared HDF5 file via CheckpointFile."""
```
JHopeCollins:
```suggestion
        """Load function data from a shared HDF5 file via :class:`.CheckpointFile`."""
```

firedrake/adjoint_utils/checkpointing.py
```diff
@@ -316,15 +409,36 @@ def __init__(self, function):
...
         from firedrake.checkpointing import CheckpointFile
         with CheckpointFile(self.file.name, 'r') as infile:
             function = infile.load_function(self.mesh, self.stored_name,
                                             idx=self.stored_index)
         return type(function)(function.function_space(),
                               function.dat, name=self.name, count=self.count)
 
+    def _restore_local_checkpoint(self):
+        """Load function data via the local checkpoint I/O manager."""
```
JHopeCollins:
```suggestion
        """Load function data via :class:`TemporaryFunctionCheckpointFile`."""
```

firedrake/adjoint_utils/checkpointing.py
```diff
@@ -316,15 +409,36 @@ def __init__(self, function):
...
+        self.stored_name, self.stored_index = self.file.temp_ckpt.save_function(
+            self.file.name, function
+        )
+
     def restore(self):
         """Read and return this Function from the checkpoint."""
+        if self.file.temp_ckpt is not None:
+            return self._restore_local_checkpoint()
+        else:
+            return self._restore_shared()
```
JHopeCollins: Renaming/recounting is orthogonal to how we interact with storage.
```suggestion
def restore(self):
"""Read and return this Function from the checkpoint."""
        if self.file.temp_ckpt is not None:
            function = self._restore_local_checkpoint()
        else:
            function = self._restore_shared_checkpoint()
        return type(function)(function.function_space(),
                              function.dat, name=self.name, count=self.count)
```

firedrake/adjoint_utils/checkpointing.py
```diff
@@ -316,15 +409,36 @@ def __init__(self, function):
...
         return type(function)(function.function_space(),
                               function.dat, name=self.name, count=self.count)
 
+    def _restore_local_checkpoint(self):
+        """Load function data via the local checkpoint I/O manager."""
+        return self.file.temp_ckpt.load_function(
+            self.file.name, self._function_space,
+            self.stored_name, self.stored_index,
+            name=self.name, count=self.count
+        )
```
JHopeCollins:
```suggestion
        return self.file.temp_ckpt.load_function(
        	self._function_space, self.stored_name, self.stored_index
        )
```

firedrake/checkpointing.py
```diff
@@ -504,6 +505,208 @@ def __del__(self):
...
+            The path to the newly created file.
+        """
+        if self.comm.rank == 0:
+            fd, filepath = tempfile.mkstemp(
+                dir=self.dirname, suffix=".h5"
+            )
+            os.close(fd)
+            filepath = self.comm.bcast(filepath)
+        else:
+            filepath = self.comm.bcast(None)
```
JHopeCollins: Single call site for collectives.
```suggestion
        else:
            filepath = None
        filepath = self.comm.bcast(filepath)
```

firedrake/checkpointing.py
```diff
@@ -504,6 +505,208 @@ def __del__(self):
...
+    def remove_file(self, filepath):
+        """Remove a checkpoint file.
+
+        Parameters
+        ----------
+        filepath : str
+            Path to the file to remove. Only rank 0 performs the deletion.
+        """
+        if self.comm.rank == 0 and os.path.exists(filepath):
+            os.remove(filepath)
```
JHopeCollins: If you have a 1-to-1 instance/file then you can copy the `CheckpointFile` API.
```suggestion
    def close(self):
        """Close the checkpoint file."""
        self.viewer.destroy()
```

firedrake/adjoint_utils/checkpointing.py
sghelichkhani: Done.

firedrake/checkpointing.py
sghelichkhani: Done.

firedrake/checkpointing.py
sghelichkhani: Done.

sghelichkhani: Done.

firedrake/checkpointing.py
sghelichkhani: Done.

firedrake/adjoint_utils/checkpointing.py
sghelichkhani: Done.

firedrake/adjoint_utils/checkpointing.py
sghelichkhani: Done.

firedrake/adjoint_utils/checkpointing.py
sghelichkhani: Done. I also return directly from inside the `with` block in each leaf method since the data is fully in memory before the viewer closes. Happy to assign outside the `with` if you prefer.

firedrake/checkpointing.py
sghelichkhani: Done.

sghelichkhani: Done.

firedrake/checkpointing.py
sghelichkhani: Done.

firedrake/checkpointing.py
sghelichkhani: Done.

firedrake/adjoint_utils/checkpointing.py
```diff
@@ -300,31 +419,70 @@ def __init__(self, function):
...
+            function = self._restore_local_checkpoint()
+        else:
+            function = self._restore_shared_checkpoint()
         return type(function)(function.function_space(),
                               function.dat, name=self.name, count=self.count)
 
+    def _restore_shared_checkpoint(self):
+        """Load function data from a shared HDF5 file via :class:`.CheckpointFile`."""
+        from firedrake.checkpointing import CheckpointFile
+        with CheckpointFile(self.file.name, 'r') as infile:
```
JHopeCollins:
```suggestion
        with CheckpointFile(self.file.name, 'r', comm=self.file.comm) as infile:
```

firedrake/adjoint_utils/checkpointing.py
```diff
@@ -300,31 +419,70 @@ def __init__(self, function):
...
+            self._function_space = function.function_space()
+            self._save_local_checkpoint(function)
+        else:
+            self._save_shared_checkpoint(function)
 
-        self.count = function.count()
+    def _save_shared_checkpoint(self, function):
+        """Save function data to a shared HDF5 file via CheckpointFile."""
+        from firedrake.checkpointing import CheckpointFile
         with CheckpointFile(self.file.name, 'a') as outfile:
```
JHopeCollins:
```suggestion
        with CheckpointFile(self.file.name, 'a', self.file.comm) as outfile:
```

firedrake/adjoint_utils/checkpointing.py
```diff
@@ -136,52 +180,128 @@ class DiskCheckpointer(TapePackageData):
...
-                dir=self.dirname, suffix=".h5"
-            )
-            checkpoint_file = self.comm.bcast(checkpoint_file)
+            _, checkpoint_file = tempfile.mkstemp(dir=self.dirname, suffix=".h5")
         else:
-            checkpoint_file = self.comm.bcast("")
+            checkpoint_file = None
+        checkpoint_file = self.comm.bcast(checkpoint_file)
         # Let h5py create a file at this location just to be sure.
         with CheckpointFile(checkpoint_file, 'w'):
```
JHopeCollins:
```suggestion
        with CheckpointFile(checkpoint_file, 'w', comm=self.comm):
```

firedrake/adjoint_utils/checkpointing.py
JHopeCollins: I meant that I think you could leave this class unchanged except for the handling of `_checkpoint_indices` below.

If you only passed `mesh.comm` if the file is a `CheckpointFile`, or only passed `checkpoint_comm` if the file is a `TemporaryFunctionCheckpointFile`, then `CheckPointFileReference.comm` would always just be the right comm and the if-statements wouldn't have to change.

JHopeCollins (changes requested):
Great, thanks so much for thoroughly addressing our reviews! It's looking good now and I think its basically ready to go.

> I think the result is much cleaner.

I'm glad you agree. The internal implementation hasn't changed but I think that the separation of concerns is clear now between how we interact with storage and how we do book-keeping for the tape.

I spotted one thing that absolutely needs to be fixed. It was a bug before this PR but its related and its only a few lines: The `CheckpointFile` must be passed the right `comm`! Currently it assumes `COMM_WORLD` which may not always be correct.

Other than that, there's one comment about not having to modify `CheckpointFileReference` that I think would reduce the diff, but you can take that or leave it.

Once the `CheckpointFile` is passed the right `comm` I'm very happy to approve this!

PR #4893, opened 2026-02-16
GoalAdaptiveNonlinearVariationalSolver

pbrubeck:
# Description
Implements a new `GoalAdaptiveNonlinearVariationalSolver` class to do adaptive refinement on a `NonlinearVariationalProblem` defined on a coarse (netgen) mesh. The adaptive procedure relies on a user-specified functional of interest, refered to as goal functional. The solver will iteratively solve -> estimate -> mark -> refine until an automatic error estimate to the goal functional falls below a user specified tolerance.


firedrake/adaptive_variational_solver.py
```diff
@@ -0,0 +1,677 @@
...
+from finat.ufl import BrokenElement, FiniteElement
+from ufl import avg, dot, div, grad, dx, ds, dS, inner, jump, as_vector, replace, FacetNormal
+
+from functools import singledispatch
+from pathlib import Path
+import numpy as np
+import csv
+
+
+class SolverCtx:
```
connorjward: This should be a `dataclass`

firedrake/adaptive_variational_solver.py
```diff
@@ -0,0 +1,677 @@
...
+        if self.V.value_shape == ():
+            DG = FunctionSpace(self.mesh, "DG", cell_residual_degree, variant=variant)
+        else:
+            DG = TensorFunctionSpace(self.mesh, "DG", cell_residual_degree, variant=variant, shape=self.V.value_shape)
+        uc = TrialFunction(DG)
+        vc = TestFunction(DG)
+        ac = inner(uc, bubbles*vc)*dx
+        Lc = residual(self.F, bubbles*vc)
+        Rcell = Function(DG, name="Rcell")  # Rcell polynomial
+        assemble(Lc)
```
pbrubeck:
```suggestion
```

firedrake/mg/ufl_utils.py
```diff
@@ -145,14 +145,22 @@ def coarsen_equation_bc(ebc, self, coefficient_mapping=None):
...
     mesh_coarse = self(fine_mesh, self)
-    name = f"coarse_{V.name}" if V.name else None
+    name = V.name
+    if name is not None:
+        level_increment = -1 if self == coarsen else 1
+        try:
+            name, prev_level = name.split("_level_")
+        except ValueError:
+            prev_level = 0
+        level = prev_level + level_increment
```
```suggestion
        level = int(prev_level) + level_increment
```

firedrake/adaptive_variational_solver.py
```diff
@@ -0,0 +1,664 @@
...
+from ufl import avg, dot, div, grad, dx, ds, dS, inner, jump, as_vector, replace, FacetNormal
+import ufl
+
+from functools import singledispatch
+from pathlib import Path
+import numpy as np
+import csv
+
+
+class GoalAdaptiveOptions:
```
pbrubeck, line 36: These are too many options, ideally we would have to describe each of them in the docstring.

firedrake/adaptive_variational_solver.py
```diff
@@ -0,0 +1,664 @@
...
+        else:
+            meshes = list(mh)
+            amh = AdaptiveMeshHierarchy(meshes[0])
+            for m in meshes[1:level+1]:
+                amh.add_mesh(m)
+        self.amh = amh
+
+        # Data storage and writing
+        self.output_dir = Path(self.options.output_dir)
+        self.output_dir.mkdir(parents=True, exist_ok=True)  # ensures folder exists
```
pbrubeck: Should we get rid of the csv output?

firedrake/adaptive_variational_solver.py
```diff
@@ -0,0 +1,664 @@
...
+        self.N_vec = []
+        self.Ndual_vec = []
+        self.eta_vec = []
+        self.etah_vec = []
+        self.eta_cell_sum_vec = []
+        self.eff1_vec = []
+        self.eff2_vec = []
+        self.eff3_vec = []
+
+    def solve_primal(self):
```
pbrubeck, line 546: Docstring

firedrake/adaptive_variational_solver.py
```diff
@@ -0,0 +1,664 @@
...
+                u.project(u_high)
+            elif self.options.primal_low_method == "interpolate":
+                u.interpolate(u_high)
+            u_err = u_high - u
+        else:
+            solve_uh()
+            u_err = None
+        return u_err
+
+    def solve_dual(self):
```
pbrubeck, line 620: Docstring

firedrake/adaptive_variational_solver.py
```diff
@@ -0,0 +1,664 @@
...
+            solve_zh(self.z_lo)
+        elif self.options.dual_low_method == "project":
+            self.z_lo.project(self.z)
+        else:
+            # Default to interpolation
+            self.z_lo.interpolate(self.z)
+        z_err = self.z - self.z_lo
+        return z_err
+
+    def compute_etah(self, u_err, z_err):
```
pbrubeck: We need a more meaningful name and a docstring

firedrake/adaptive_variational_solver.py
```diff
@@ -0,0 +1,664 @@
...
+            u_err_exact = self.u_exact - u
+            eta_dual_exact = assemble(
+                inner(inner(Rcell_star, u_err_exact), test)*dx
+                + inner(avg(inner(Rfacet_star, u_err_exact)), both(test))*dS
+                + inner(inner(Rfacet_star, u_err_exact), test)*ds
+            )
+            udiff = assemble(eta_dual_exact - eta_dual)
+            with udiff.dat.vec as uvec:
+                unorm = uvec.norm()
+            self.print("L2 error in (dual) refinement indicators: ", unorm)
```
pbrubeck: Should we remove this code?

firedrake/adaptive_variational_solver.py
```diff
@@ -0,0 +1,664 @@
...
+        return eta_cell
+
+    def manual_error_indicators(self, u_err, z_err):
+        """ Currently only implemented for Poisson, but can be overriden. To adapt to other PDEs, replace the form of
+        eta_cell = assemble() to the symbolic form of the error indicators. This form is usually obtained by integrating
+        the weak form by parts (to recover the strong form) and redistributing facet fluxes.
+        """
```
pbrubeck: Should we remove this code, or leave it unimplemented? Ideally we want a Poisson test that compares the automatic vs manual indicators.

firedrake/adaptive_variational_solver.py
```diff
@@ -0,0 +1,664 @@
...
+        eta_cell = assemble(
+            inner(f + div(grad(u)), z_err * test) * dx
+            - inner(jump(grad(u), n), z_err * avg(test)) * dS
+            - inner(dot(grad(u), n), z_err * test) * ds
+        )
+        with eta_cell.dat.vec as evec:
+            evec.abs()
+        return eta_cell
+
+    def compute_efficiency(self, eta_cell, eta_h, eta):
```
pbrubeck: Docstring

firedrake/adaptive_variational_solver.py
```diff
@@ -0,0 +1,664 @@
...
+
+    def set_uniform_cell_markers(self):
+        """Uniform marking for comparison tests"""
+        mesh = self.amh[-1]
+        markers_space = FunctionSpace(mesh, "DG", 0)
+        markers = Function(markers_space)
+        markers.assign(1)
+        return markers
+
+    def refine_problem(self, markers):
```
pbrubeck, line 885: Docstring

firedrake/adaptive_variational_solver.py
```diff
@@ -0,0 +1,664 @@
...
+
+        coef_map = {}
+        self.problem = refine(self.problem, refine, coefficient_mapping=coef_map)
+        self.goal_functional = refine(self.goal_functional, refine, coefficient_mapping=coef_map)
+        if self.u_exact is not None:
+            self.u_exact = refine(self.u_exact, refine, coefficient_mapping=coef_map)
+        if self.source_term is not None:
+            self.source_term = refine(self.source_term, refine, coefficient_mapping=coef_map)
+
+    def solve(self):
```
pbrubeck: This is the main method a user will be calling, so we need a good docstring. We should also think about what we want to happen when the user calls `solve()` multiple times.

firedrake/adaptive_variational_solver.py
```diff
@@ -0,0 +1,664 @@
...
+        for it in range(self.options.max_iterations):
+            try:
+                self.step()
+            except StopIteration:
+                break
+
+    def step(self):
```
pbrubeck: This method enables users to take a step at a time, so they can post process each level. I am not sure if rasing `StopIteration` is the right thing, considering that the user will not instinctively be aware that we are raising it when the maximum iterations or the convergence criteria is met.

Maybe the correct approach is to have a `self.check_convergence()` or `self.check_max_iterations()` methods...

firedrake/adaptive_variational_solver.py
```diff
@@ -0,0 +1,664 @@
...
+        else:
+            file_path = self.output_dir / self.options.results_file_name
+        rows = list(zip(self.N_vec, self.Ndual_vec, self.eta_vec, self.etah_vec, self.eta_cell_sum_vec, self.eff1_vec, self.eff2_vec))
+        headers = ("N", "Ndual", "eta", "eta_h", "sum_eta_T", "eff1", "eff2")
+        with open(file_path, "w", newline="") as file:
+            w = csv.writer(file)
+            w.writerow(headers)
+            w.writerows(rows)
+
+    def append_data(self):
```
pbrubeck: Should this function go away?

firedrake/adaptive_variational_solver.py
```diff
@@ -0,0 +1,664 @@
...
+    def write_mesh(self, it):
+        mesh = self.problem.u.function_space().mesh()
+        should_write = False
+        write_mesh = self.options.write_mesh
+        if write_mesh == "all":
+            should_write = True
+        elif write_mesh == "first_and_last":
+            if it == 0 or it == self.options.max_iterations:
+                should_write = True
+        elif write_mesh == "by_iteration":
```
pbrubeck: These are too many cases, and it is not obvious for a user. The user could decide if they want to do this by writing their own adaptivity loop and calling `solver.step()`. Should we just remove this function?

firedrake/adaptive_variational_solver.py
```diff
@@ -0,0 +1,664 @@
...
+
+    def write_solution(self, it):
+        should_write = False
+        write_solution = self.options.write_mesh
+        if write_solution == "all":
+            should_write = True
+        elif write_solution == "first_and_last":
+            if it == 0 or it == self.options.max_iterations:
+                should_write = True
+        elif write_solution == "by_iteration":
```
pbrubeck: These are too many cases, and it is not obvious for a user. The user could decide if they want to do this by writing their own adaptivity loop and calling solver.step(). Should we just remove this function?

firedrake/adaptive_variational_solver.py
```diff
@@ -0,0 +1,664 @@
...
+
+
+# Only required for my examples:
+def getlabels(mesh, codim):
+    ngmesh = mesh.netgen_mesh
+    names = ngmesh.GetRegionNames(codim=codim)
+    names_to_labels = {}
+    for l in names:
+        names_to_labels[l] = tuple(i+1 for i, name in enumerate(names) if name == l)
+    return names_to_labels
```
pbrubeck: This should probably become a class method, something like:

`mesh.region_names_to_subdomain_ids(codim)`

firedrake/adaptive_variational_solver.py
```diff
@@ -0,0 +1,664 @@
...
+            try:
+                name, prev_level = name.split("_level_")
+            except ValueError:
+                prev_level = 0
+            level = int(prev_level) + level_increment
+            name = f"{name}_level_{level}"
+            new = Function(Vc, name=name)
+        new.interpolate(expr)
+        coefficient_mapping[expr] = new
+    return new
```
pbrubeck: This should live in `mg/ufl_utils.py`

firedrake/adaptive_variational_solver.py
pefarrell: Yes, sure

firedrake/adaptive_variational_solver.py
```diff
@@ -0,0 +1,667 @@
...
+        An options dictionary to construct a :class:`GoalAdaptiveOptions`
+    primal_solver_parameters
+        A dictionary of solver parameters for the primal problem
+    dual_solver_parameters
+        A dictionary of solver parameters for the dual problem.
+        Defaults to `primal_solver_parameters`.
+    primal_solver_kwargs
+        Keyword arguments for the primal :class:`~.NonlinearVariationalSolver`
+    dual_solver_kwargs
+        Keyword arguments for the primal :class:`~.LinearVariationalSolver`
```
pbrubeck:
```suggestion
        Keyword arguments for the dual :class:`~.LinearVariationalSolver`
```

firedrake/mg/adaptive_hierarchy.py
```diff
@@ -21,7 +21,7 @@ class AdaptiveMeshHierarchy(HierarchyBase):
     """
     def __init__(self, base_mesh: MeshGeometry, nested: bool = True):
         self.meshes = []
-        self._meshes = []
+        self._meshes = self.meshes
```
connorjward, line 24: This just creates a reference to the same object

firedrake/mg/ufl_utils.py
```diff
@@ -387,6 +416,44 @@ def coarsen_slate_tensor_op(tensor, self, coefficient_mapping=None):
     return type(tensor)(*children)
 
 
+@singledispatch
+def refine(expr, self, coefficient_mapping=None):
+    return coarsen(expr, self, coefficient_mapping=coefficient_mapping)  # fallback to original
```
connorjward: This is very weird. Aren't they supposed to do the opposite things?

firedrake/adaptive_variational_solver.py
```diff
@@ -0,0 +1,667 @@
...
+from ufl import avg, dx, ds, dS, inner, as_vector, replace
+import ufl
+import numpy as np
+from numbers import Integral
+
+
+__all__ = ["GoalAdaptiveNonlinearVariationalSolver"]
+
+
+class GoalAdaptiveOptions:
```
connorjward, line 36: Better as a frozen dataclass, much less boilerplate

firedrake/adaptive_variational_solver.py
```diff
@@ -0,0 +1,667 @@
...
+        * ``"solve"`` – solve the dual problem independently in the base space;
+          this is the most expensive option but produces the most accurate solver-error
+          estimate.
+    write_solution
+        Controls when primal and dual solutions are written to VTK files.
+
+        * ``False`` (default) – never write.
+        * ``True`` – write at every iteration.
+        * A positive integer ``k`` – write every ``k`` iterations.
+    verbose
```
connorjward, line 107: I feel like all of these output options are usually set via PETSc options?

firedrake/adaptive_variational_solver.py
```diff
@@ -0,0 +1,667 @@
...
+        If supplied, it is used to calculate the efficiency of the error estimate
+    exact_goal
+        The exact value for the goal functional (optional).
+        If supplied, it is used to calculate the efficiency of the error estimate.
+    """
+    def __init__(self,
+                 problem: NonlinearVariationalProblem,
+                 goal_functional: ufl.BaseForm,
+                 tolerance: float,
+                 goal_adaptive_options: dict | None = None,
```
connorjward:
```suggestion
                 goal_adaptive_options: GoalAdaptiveOptions | dict | None = None,
```

firedrake/adaptive_variational_solver.py
```diff
@@ -0,0 +1,667 @@
...
+        self.sp_dual = dual_solver_parameters if dual_solver_parameters is not None else primal_solver_parameters
+        self.primal_solver_kwargs = primal_solver_kwargs if primal_solver_kwargs is not None else {}
+        self.dual_solver_kwargs = dual_solver_kwargs if dual_solver_kwargs is not None else {}
+        self.u_exact = as_mixed(exact_solution) if isinstance(exact_solution, (tuple, list)) else exact_solution
+        self.goal_exact = exact_goal
+
+        # Set up an AdaptiveMeshHierarchy for every mesh of the problem
+        V = problem.u.function_space()
+        meshes = set(V.mesh())
+        meshes.add(V.mesh())
```
connorjward: Why is this line needed?

firedrake/adaptive_variational_solver.py
```diff
@@ -0,0 +1,667 @@
...
+            z_lo.project(z)
+        elif self.options.dual_low_method == "interpolate":
+            z_lo.interpolate(z)
+        else:
+            raise ValueError(f"Unrecognised dual_low_method {self.options.dual_low_method}")
+        z_err = z - z_lo
+        self.z = z
+        return z_lo, z_err
+
+    def estimate_error(self, u_err, z_lo, z_err):
```
connorjward, line 686: Type hints throughout would be nice

firedrake/adaptive_variational_solver.py
```diff
@@ -0,0 +1,667 @@
...
+            self.print(f'{"Discretisation error, 0.5(rho + rho*)":40s}{discretisation_error: 15.12e}')
+        else:
+            self.print(f'{"Discretisation error, rho(u_h; z-z_h)":40s}{discretisation_error: 15.12e}')
+        self.print(f'{"Solver error, rho(u_h; z_h):":40s}{solver_error: 15.12e}')
+        self.print(f'{"Final error estimate:":40s}{eta_h: 15.12e}')
+        return eta_h, eta
+
+    def compute_error_indicators(self, u_err, z_lo, z_err):
+        """Compute cell and facet residuals R_cell, R_facet"""
+        from firedrake.assemble import assemble
```
connorjward: Does this import need to be inside?

firedrake/adaptive_variational_solver.py
```diff
@@ -0,0 +1,667 @@
...
+        self.compute_efficiency_indices(eta_cell, eta_h, eta)
+        markers = self.set_adaptive_cell_markers(eta_cell)
+        # REFINE
+        self.print("Transferring problem to new mesh ...")
+        self.refine_problem(markers)
+        return u_out, eta_h
+
+    def print(self, *args, **kwargs):
+        if self.options.verbose:
+            PETSc.Sys.Print(*args, **kwargs)
```
connorjward: Doesn't this need a comm?

firedrake/adaptive_variational_solver.py
```diff
@@ -0,0 +1,667 @@
...
+        "pc_type": "jacobi",
+    }
+    sp_facet = {
+        "snes_type": "ksponly",
+        "ksp_type": "cg",
+        "pc_type": "jacobi",
+    }
+
+
+class GoalAdaptiveNonlinearVariationalSolver:
```
connorjward: It feels to me like this should inherit from some sort of `AbstractNonlinearVariationalSolver` type, as it implicitly shares an interface with our other solver classes.

firedrake/adaptive_variational_solver.py
```diff
@@ -0,0 +1,667 @@
...
+            z_lo.interpolate(z)
+            solve_zh(z_lo)
+        elif self.options.dual_low_method == "project":
+            z_lo.project(z)
+        elif self.options.dual_low_method == "interpolate":
+            z_lo.interpolate(z)
+        else:
+            raise ValueError(f"Unrecognised dual_low_method {self.options.dual_low_method}")
+        z_err = z - z_lo
+        self.z = z
```
connorjward, line 683: It can be dangerous to mutate attributes inside methods like this because it makes errors hard to track down. I wonder if we should make `solve_dual` into `_solve_dual` (i.e. make it private) because it only makes sense to call inside `solve`. The same for many other methods in this class.

firedrake/mg/ufl_utils.py
```diff
@@ -145,18 +145,30 @@ def coarsen_equation_bc(ebc, self, coefficient_mapping=None):
...
+            prev_level = 0
+        level = int(prev_level) + level_increment
+        name = f"{name}_level_{level}"
+    V_new = V.reconstruct(mesh=new_mesh, name=name)
+    if self == coarsen:
+        V_new._fine = V
+        V._coarse = V_new
+    elif self == refine:
+        V_new._coarse = V
+        V._fine = V_new
```
connorjward: I think that a function called `coarsen_function_space` that does this is going to cause a huge amount of confusion.

connorjward (changes requested):
This feels a little like we're reinventing a new interface to nonlinear solvers when we already have one: SNES. Logging things is done via monitors and passing variational forms can be done with the appctx.

@JHopeCollins is the expert here, but don't you think that this might be a use case for a Python SNES? I worry that the current approach isn't as composable as other parts of Firedrake.

pefarrell:
No, I don't think this is a job for a SNES. A SNES inherently works on a fixed finite-dimensional space: the size of the vectors involved is fixed. An adaptive loop is something that changes the space being used to approximate the PDE. The adaptive loop (solve -> estimate -> mark -> refine) must happen outside the solver loop. We use SNES on the inside, but SNES can't do what this class does.

connorjward:
I see. I still wonder if there are ways that this work could be better integrated into the rest of Firedrake and made to be as composable as possible - I could be wrong though. @pbrubeck can you bring this to the next Firedrake meeting for discussion?

connorjward:
For example we might want to have `MeshAdapter` and `SolverMonitor` abstract classes that would own bits of what the `GoalAdaptiveVariationalSolver` does.

connorjward:
How does this compare to all the mesh adaptivity work that has already been done with Firedrake (https://github.com/mesh-adaptation)? In particular https://github.com/mesh-adaptation/goalie

@stephankramer 
@joewallwork

pefarrell:
I can add my perspective. I'm sure @stephankramer and @joewallwork will have other interesting things to say!

Compared to goalie, this PR

- addresses only stationary problems, whereas goalie (only?) considers transient ones;
- employs adaptive mesh refinement from netgen, whereas goalie/animate use anisotropic adaptive remeshing or mesh movement from animate/movement;
- involves a very small amount of work to convert a code that solves one discrete nonlinear problem to apply goal-based adaptivity, whereas using goalie would require substantial rewriting in the goalie format;
- gives explicit error estimates for goal functionals, which I don't believe goalie does (at least I don't see it in the demos).

This PR also composes nicely with the rest of Firedrake. When the mesh is refined, the integration with netgen ensures the geometry is better approximated (this is not true with goalie). The adaptive mesh hierarchy constructed by the `GoalAdaptiveNonlinearVariationalSolver` can be used for all solvers Firedrake supports, including geometric multigrid. (I don't believe this is the case with goalie but may be wrong.)

More about adaptive mesh refinement vs adaptive remeshing, as there is a major mathematical difference. In the adaptive mesh refinement used in this PR, the adapted mesh is a refinement of the input mesh (ignoring better geometry approximation). This means that for normal elements the refined function space is a superspace of the original one, which is very advantageous for stable approximation: loosely speaking, your approximation can only get better as you refine. In contrast, with adaptive remeshing, the input and output meshes are in principle unrelated (although they are in practice, because the algorithm transforms the one into the other step by step), and so you don't have nice hierarchical approximations. I suspect this is why 95%+ of the community around adaptive discretisations use adaptive mesh refinement.

joewallwork:
Hi all, great to see this contribution! Nice work. I'll try to find some time to play around with this.

To respond to @pefarrell's points:

>     * addresses only stationary problems, whereas goalie (only?) considers transient ones;

Yes, Goalie is designed to handle time-dependent problems that consider multiple meshes during the simulation. You can apply it to stationary problems, too.

>     * employs adaptive mesh refinement from netgen, whereas goalie/animate use anisotropic adaptive remeshing or mesh movement from animate/movement;

I had proposed an MSc project for a student to hook up AMR in Goalie using Netgen but unfortunately it didn't go ahead. It was designed to be general enough to allow different adaptor approaches but we have mainly used Animate so far.

>     * involves a very small amount of work to convert a code that solves one discrete nonlinear problem to apply goal-based adaptivity, whereas using goalie would require substantial rewriting in the goalie format;

Yeah this is one of the main limitations of Goalie.

>     * gives explicit error estimates for goal functionals, which I don't believe goalie does (at least I don't see it in the demos).

Currently we make a choice of error estimate based on the dual-weighted residual. There were plans to support more/custom error estimates but we haven't had time.
 
> This PR also composes nicely with the rest of Firedrake. When the mesh is refined, the integration with netgen ensures the geometry is better approximated (this is not true with goalie). The adaptive mesh hierarchy constructed by the `GoalAdaptiveNonlinearVariationalSolver` can be used for all solvers Firedrake supports, including geometric multigrid. (I don't believe this is the case with goalie but may be wrong.)

I don't recall Goalie having limitations in terms of solvers, although I haven't tried loads of them. However, it is currently limited to simple theta-method timestepping schemes.

connorjward:
Thanks for this useful exposition.

By raising this I wanted to point out that there is a more general appetite for adaptivity using Firedrake and therefore it would be very nice to set this up in a more general way so that potential future contributions would be able to reuse some of the code.

I think this should be brought to a Firedrake team meeting for further discussion.

firedrake/mg/ufl_utils.py
```diff
@@ -226,10 +272,37 @@ def inject_on_restrict(fine, restriction, rscale, injection, coarse):
...
+    if coefficient_mapping is None:
+        coefficient_mapping = {}
+    bcs = [self(bc, self, coefficient_mapping=coefficient_mapping)
+           for bc in problem._original_bcs]
+    A = self(problem._original_A, self, coefficient_mapping=coefficient_mapping)
+    M = self(problem._original_M, self, coefficient_mapping=coefficient_mapping)
+    orig = problem
+    problem = firedrake.LinearEigenproblem(A, M, bcs=bcs,
+                                           bc_shift=orig.bc_shift, restrict=orig.restrict)
+    orig._fine = problem
```
```suggestion
    if self == coarsen:
        orig._coarse = problem
    elif self == refine:
        orig._fine = problem
```

firedrake/adaptive_variational_solver.py
dham: I think this whole thing should be PETSc options. This would be cognate with what we do for NVS

firedrake/adaptive_variational_solver.py
```diff
@@ -0,0 +1,912 @@
...
+        self.options = self._make_options(goal_adaptive_options)
+        self.goal_exact = exact_goal
+
+        self.Ndofs_vec = []
+        self.eta_vec = []
+        self.etah_vec = []
+        self.eta_cell_sum_vec = []
+        self.eff1_vec = []
+        self.eff2_vec = []
+        self.eff3_vec = []
```
dham: why 3?

firedrake/adaptive_variational_solver.py
```diff
@@ -0,0 +1,912 @@
...
+            self.print(f'{"Effectivity index:":40s}{eff1: 15.12f}')
+            self.print(f'{"Localisation efficiency:":40s}{eff2: 15.12f}')
+        else:
+            eff3 = eta_cell_total / eta_h
+            self.eff3_vec.append(eff3)
+            self.print(f'{"Localisation efficiency:":40s}{eff3: 15.12f}')
+
+    def print(self, *args, **kwargs):
+        if self.options.verbose:
+            PETSc.Sys.Print(*args, **kwargs)
```
dham, line 307: comm?

firedrake/adaptive_variational_solver.py
```diff
@@ -0,0 +1,912 @@
...
+        mesh = self.problem.u.function_space().mesh()
+        if len(mesh) > 1:
+            new_mesh = type(mesh)([get_level(m)[0][-1] for m in mesh])
+            amh, _ = get_level(mesh)
+            amh.add_mesh(new_mesh)
+        if coef_map is None:
+            coef_map = {}
+        self.problem = refine(self.problem, refine, coefficient_mapping=coef_map)
+
+    def compute_efficiency_indices(self, eta_cell, eta_h, eta):
```
dham, line 301: feels like it should somehow be programmable rather than being part of the abstract interface.

firedrake/adaptive_variational_solver.py
```diff
@@ -0,0 +1,912 @@
...
+        Returns
+        -------
+        Function
+            A DG0 Function of cell-wise indicators (absolute values).
+        """
+        raise NotImplementedError
+
+    def write_solution(self, it):
+        """Write solution output (VTK etc.).  Default no-op."""
+        pass
```
dham: some sort of end of iteration callback?

firedrake/adaptive_variational_solver.py
```diff
@@ -0,0 +1,912 @@
...
+        The variational formulation of the PDE defined on the coarse mesh
+    goal_functional
+        The goal functional defined in terms of the solution to the PDE
+    goal_adaptive_options
+        An options dictionary to construct a :class:`GoalAdaptiveOptions`
+    primal_solver_parameters
+        A dictionary of solver parameters for the primal problem
+    dual_solver_parameters
+        A dictionary of solver parameters for the dual problem.
+        Defaults to `primal_solver_parameters`.
```
dham: namespace this into the options tree.

firedrake/adaptive_variational_solver.py
```diff
@@ -0,0 +1,912 @@
...
+
+        Returns
+        -------
+        tuple[Function, float]
+            ``(u_out, error_estimate)`` where ``u_out`` is the solution on the
+            finest mesh reached and ``error_estimate`` is the final ``|eta_h|``.
+        """
+        u_out = self.u_high if self.u_high is not None else self.problem.u
+        error_estimate = self.etah_vec[-1]
+        return u_out, error_estimate
```
dham: return the soltion, query for the error. Just like converged reason.

dham (changes requested):

pbrubeck:
Closed in favour of #5273

PR #4900, opened 2026-02-18
Avoid creating a function space when calculating bounding box coordinates

leo-collins:
Creating the `VectorFunctionSpace` was taking a big chunk of the cell-location time. We can avoid doing this in the non-extruded case.

### DIFF
--- a/firedrake/mesh.py
+++ b/firedrake/mesh.py
@@ -35,7 +35,7 @@ import firedrake.cython.spatialindex as spatialindex
 import firedrake.utils as utils
 from firedrake.utils import as_cstr, IntType, RealType
 from firedrake.logging import info_red
-from firedrake.parameters import parameters, target
+from firedrake.parameters import parameters
 from firedrake.petsc import PETSc, DEFAULT_PARTITIONER
 from firedrake.adjoint_utils import MeshGeometryMixin
 from firedrake.exceptions import VertexOnlyMeshMissingPointsError, NonUniqueMeshSequenceError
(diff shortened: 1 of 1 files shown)


firedrake/mesh.py
```diff
@@ -2563,7 +2563,7 @@ def bounding_box_coords(self) -> Tuple[np.ndarray, np.ndarray] | None:
         Hence the bounding box will contain the entire element.
         """
         from firedrake import function, functionspace
-        from firedrake.parloops import par_loop, READ, MIN, MAX
+        import loopy as lp
```
connorjward: top level import

firedrake/mesh.py
```diff
@@ -2606,8 +2619,7 @@ def bounding_box_coords(self) -> Tuple[np.ndarray, np.ndarray] | None:
         else:
             coords = mesh.coordinates
 
-        cell_node_list = mesh.coordinates.function_space().cell_node_list
-        _, nodes_per_cell = cell_node_list.shape
+        nodes_per_cell = mesh.coordinates.function_space().finat_element.space_dimension()
```
connorjward: This is less obvious (or wrong?) than the previous approach

firedrake/mesh.py
```diff
@@ -2589,13 +2589,26 @@ def bounding_box_coords(self) -> Tuple[np.ndarray, np.ndarray] | None:
             f.interpolate(self.coordinates)
             mesh = Mesh(f)
 
-        # Calculate the bounding boxes for all cells by running a kernel
-        V = functionspace.VectorFunctionSpace(mesh, "DG", 0, dim=gdim)
-        coords_min = function.Function(V, dtype=RealType)
-        coords_max = function.Function(V, dtype=RealType)
+        if mesh.extruded:
+            # Limitations of pyop2 mean we need to build this function space
```
```suggestion
            # Limitations of pyop2 mean we need to build this function space, which is expensive
```

PR #4906, opened 2026-02-19
Optimisation in `parent_mesh_embedding`

leo-collins:
Before, in order to get the mesh cell rank ownership on each process, we were interpolating a `Constant(parent_mesh.comm.rank)` into DG0 Function. Building the function space and generating the code was very slow. 

We can just build `visible_ranks` on each process and do a halo exchange.

### DIFF
--- a/firedrake/mesh.py
+++ b/firedrake/mesh.py
@@ -4442,26 +4442,24 @@ def _parent_mesh_embedding(
         reference_coords = reference_coords[:, : parent_mesh.topological_dimension]
 
     # Get parent mesh rank ownership information.
-    with PETSc.Log.Event("get_parent_mesh_rank_ownership_information"):
-        visible_ranks = np.empty(parent_mesh.cell_set.total_size, dtype=IntType)
-        visible_ranks[:parent_mesh.cell_set.size] = parent_mesh.comm.rank
-        visible_ranks[parent_mesh.cell_set.size:] = -1
-        # Halo exchange the visible ranks so that each rank knows which ranks can see each cell.
-        dmcommon.exchange_cell_orientations(
-            parent_mesh.topology.topology_dm, parent_mesh.topology._cell_numbering, visible_ranks
-        )
-        locally_visible = np.full(ncoords_global, False)
-        # See below for why np.inf is used here.
-        ranks = np.full(ncoords_global, np.inf)
-        locally_visible[:] = parent_cell_nums != -1
+    visible_ranks = np.empty(parent_mesh.cell_set.total_size, dtype=IntType)
+    visible_ranks[:parent_mesh.cell_set.size] = parent_mesh.comm.rank
+    visible_ranks[parent_mesh.cell_set.size:] = -1
+    # Halo exchange the visible ranks so that each rank knows which ranks can see each cell.
+    dmcommon.exchange_cell_orientations(
+        parent_mesh.topology.topology_dm, parent_mesh.topology._cell_numbering, visible_ranks
+    )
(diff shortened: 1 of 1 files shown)


firedrake/mesh.py
```diff
@@ -4454,8 +4430,27 @@ def _parent_mesh_embedding(
...
 
-    locally_visible[:] = parent_cell_nums != -1
-    ranks[locally_visible] = visible_ranks[parent_cell_nums[locally_visible]]
+    # Get parent mesh rank ownership information.
+    with PETSc.Log.Event("get_parent_mesh_rank_ownership_information"):
+        visible_ranks = np.empty(parent_mesh.cell_set.total_size, dtype=IntType)
+        visible_ranks[:parent_mesh.cell_set.size] = parent_mesh.comm.rank
+        visible_ranks[parent_mesh.cell_set.size:] = -1
+        # Halo exchange the visible ranks so that each rank knows which ranks can see each cell.
+        dmcommon.exchange_cell_orientations(
```
connorjward: Maybe this needs renaming. Or couldn't you just use the SF directly?

firedrake/mesh.py
```diff
@@ -4454,8 +4430,27 @@ def _parent_mesh_embedding(
...
+            parent_mesh.topology.topology_dm, parent_mesh.topology._cell_numbering, visible_ranks
+        )
+        locally_visible = np.full(ncoords_global, False)
+        # See below for why np.inf is used here.
+        ranks = np.full(ncoords_global, np.inf)
+        locally_visible[:] = parent_cell_nums != -1
+        if parent_mesh.extruded:
+            base_parent_cell_nums, _ = _parent_extrusion_numbering(
```
connorjward: We weren't doing this before. Why now?

leo-collins: The halo exchange is done on the base mesh topology so have to index `visible_ranks` using the base cell numbers

firedrake/mesh.py
leo-collins: It should probably be renamed

firedrake/mesh.py
leo-collins: I've added a comment explaining

firedrake/mesh.py
```diff
@@ -4466,8 +4441,28 @@ def _parent_mesh_embedding(
...
+        visible_ranks[:parent_mesh.cell_set.size] = parent_mesh.comm.rank
+        visible_ranks[parent_mesh.cell_set.size:] = -1
+        # Halo exchange the visible ranks so that each rank knows which ranks can see each cell.
+        dmcommon.exchange_cell_orientations(
+            parent_mesh.topology.topology_dm, parent_mesh.topology._cell_numbering, visible_ranks
+        )
+        locally_visible = np.full(ncoords_global, False)
+        # See below for why np.inf is used here.
+        ranks = np.full(ncoords_global, np.inf)
+        locally_visible[:] = parent_cell_nums != -1
```
```suggestion
        # See below for why np.inf is used here.
        ranks = np.full(ncoords_global, np.inf)
        locally_visible = parent_cell_nums != -1
```
don't think you even need this line

firedrake/mesh.py
connorjward: Is that going to happen?

leo-collins: I'll do it in another PR. It only works with IntType currently so will need tweaking before it becomes a general "halo exchange this array" function

firedrake/mesh.py
leo-collins: Fixed

Issue #4819, opened 2026-01-15
BUG: io backward compatibility tests are timing out on CI


JHopeCollins:
Debug and revert #4818

PR #4913, opened 2026-02-23
Fix deadlock when the compiler crashes internally

connorjward:
## Changes

* Prevent compiler errors from causing a deadlock
* Print out compiler errors when running CI as the log files aren't available
* Disable -ffast-math for GCC to fix internal compiler error on CI runners
* Re-enable skipped IO tests

Closes https://github.com/firedrakeproject/firedrake/issues/4819

## What happened

In https://github.com/firedrakeproject/firedrake/pull/4911 I have been investigating a very tricky timeout bug. I finally tracked it down to the following compiler error:
```
during GIMPLE pass: slp
mfe.c: In function 'expression_kernel':
mfe.c:8:13: internal compiler error: Floating point exception
    8 | static void expression_kernel(double complex *__restrict__ A, double complex const *__restrict__ w_0, double complex const *__restrict__ w_1)
      |             ^~~~~~~~~~~~~~~~~
0x108d932 internal_error(char const*, ...)
        ???:0
0x172411d vect_slp_analyze_operations(vec_info*)
        ???:0
0x13f916a vect_slp_function(function*)
        ???:0
```
This would only happen on the CI runners. On my  workstation with the same Docker image things ran fine. Disabling fast math fixes it.

### DIFF
--- a/pyop2/compilation.py
+++ b/pyop2/compilation.py
@@ -567,6 +567,7 @@ def make_so(compiler, code, extension, comm):
 
     # Compile on compilation communicator (ccomm) rank 0
     if ccomm.rank == 0:
+        # Track exceptions as values so that they may be raised collectively
         try:
             # Adding random 2-digit hexnum avoids using excessive filesystem inodes
             tempdir = MEM_TMP_DIR.joinpath(f"{randint(0, 255):02x}")


pyop2/compilation.py
```diff
@@ -543,15 +543,14 @@ def check_source_hashes(compiler, code, extension, comm):
     make_cache=lambda: CompilerDiskAccess(configuration['cache_dir'], extension=".so")
 )
 @PETSc.Log.EventDecorator()
-def make_so(compiler, code, extension, comm, filename=None):
+def make_so(compiler, code, extension, comm):
```
connorjward, line 546: This wasn't used anywhere and would have immediately crashed were this not `None`

pyop2/compilation.py
```diff
@@ -567,69 +566,95 @@ def make_so(compiler, code, extension, comm, filename=None):
...
-                # Extract linker specific "cflags" from ldflags and link
-                ld = tuple(shlex.split(compiler.ld)) + ('-o', str(soname), str(oname)) + tuple(expandWl(compiler.ldflags))
-                _run(ld, logfile, errfile, step="Linker", filemode="a")
+            cname = filename
+            oname = filename.with_suffix(".o")
+            soname = filename.with_suffix(".so")
+            logfile = filename.with_suffix(".log")
+            errfile = filename.with_suffix(".err")
+        except BaseException as e:
+            result = e
```
connorjward, line 585: Here I put everything inside try/except blocks, which is why this looks weird

pyop2/compilation.py
```diff
@@ -567,69 +566,95 @@ def make_so(compiler, code, extension, comm, filename=None):
...
+                result.__cause__ = e  # equivalent to 'raise XXX from e'
+            except BaseException as e:
+                # catch and broadcast all exceptions to prevent deadlocks
+                result = e
+    else:
+        result = None
 
-    return ccomm.bcast(soname, root=0)
+    result = ccomm.bcast(result)
+    if isinstance(result, Exception):
```
JHopeCollins:
```suggestion
    if isinstance(result, BaseException):
```

connorjward: oh yeah, duh

pyop2/compilation.py
```diff
@@ -567,69 +566,95 @@ def make_so(compiler, code, extension, comm, filename=None):
         compiler_flags = compiler.cflags
 
     # Compile on compilation communicator (ccomm) rank 0
-    soname = None
     if ccomm.rank == 0:
```
JHopeCollins, line 569: Please can you add a quick comment explaining that we need this try-catch dance to avoid deadlocks.

JHopeCollins (dismissed):
I think this looks ok. Thanks for getting to the bottom of it!

PR #4920, opened 2026-02-25
migrate scipy sparse matrices to sparse arrays

j-bowhay:
# Description

SciPy is slowly moving away from the sparse matrix interface to sparse arrays. This PR preemptively migrates the few uses of SciPy sparse in Firedrake


connorjward:
Test failures are very likely to be from https://github.com/firedrakeproject/firedrake/pull/4919. I will shepherd this through when that is fixed.

PR #4922, opened 2026-02-25
Fix dumpdir fixture

connorjward:
Previously the `dumpdir` fixture would rely on the `mesh` fixture. This lead to some confusing behaviour where

```
$ pytest tests/firedrake/output/test_pvd_output.py::test_bad_cell --co
```

was giving the very confusing result of

```
collected 8 items                                                                                                     

<Dir firedrake>
  <Dir tests>
    <Dir firedrake>
      <Dir output>
        <Module test_pvd_output.py>
          <Function test_bad_cell[interval]>
          <Function test_bad_cell[square[tri]]>
          <Function test_bad_cell[square[quad]]>
          <Function test_bad_cell[box[tet]]>
          <Function test_bad_cell[box[quad x interval]]>
          <Function test_bad_cell[box[hex]]>
          <Function test_bad_cell[sphere[tri]]>
          <Function test_bad_cell[sphere[quad]]>

```

despite `test_bad_cell` not needing the `mesh` fixture.

### DIFF
--- a/tests/firedrake/output/conftest.py
+++ b/tests/firedrake/output/conftest.py
@@ -1,10 +1,13 @@
+import os
 import tempfile
+
 import pytest
+from mpi4py import MPI
 
 
 @pytest.fixture
-def dumpdir(mesh):
-    comm = mesh.comm
+def dumpdir():
+    comm = MPI.COMM_WORLD
     if comm.rank == 0:
         tmp = tempfile.TemporaryDirectory()
         yield comm.bcast(tmp.name, root=0)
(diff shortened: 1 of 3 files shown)


tests/firedrake/output/conftest.py
```diff
@@ -13,3 +16,8 @@ def dumpdir(mesh):
     else:
         yield comm.bcast(None, root=0)
         comm.barrier()
+
+
+@pytest.fixture
+def dumpfile(dumpdir):
+    yield os.path.join(dumpdir, "dump")
```
JHopeCollins, line 23: Should this yield a unique name each time in case it's used by multiple tests running at the same time?

connorjward: `dumpdir` is unique to each test so I don't think so

PR #4927, opened 2026-02-26
Improvements for plotting tests

connorjward:
* Clear old figures to prevent leaks
* Fix issue with DISPLAY variable breaking tests over SSH

### DIFF
--- a/tests/firedrake/output/test_plotting.py
+++ b/tests/firedrake/output/test_plotting.py
@@ -360,7 +360,7 @@ def test_tripcolor_movie():
     def animate(time):
         t.assign(time)
         q.interpolate(expr)
-        colors.set_array(fn_plotter(q))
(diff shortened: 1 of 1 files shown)


tests/firedrake/output/test_plotting.py
```diff
@@ -363,3 +368,6 @@ def animate(time):
     interval = 1e3 / fps
     movie = FuncAnimation(fig, animate, frames=frames, interval=interval)
     assert movie is not None
+
+    # Use a method of the animation to prevent warning about it being unused
+    movie.to_jshtml()
```
JHopeCollins, line 373: How big is the html? This might be quicker.
```suggestion
    movie.pause()
```

connorjward: That doesn't stop the warning. Just tried it.

PR #4928, opened 2026-02-26
Remove broken submesh example from docstring

connorjward:
The code snippet has never actually worked and has confused several users.

### DIFF
--- a/firedrake/mesh.py
+++ b/firedrake/mesh.py
@@ -4639,31 +4639,6 @@ def Submesh(mesh, subdim, subdomain_id, label_name=None, name=None):
     ridges to be contained in the quad mesh are shared by at most two
     facets to make the quad mesh orientation algorithm work.
 
-    Examples
-    --------
-
-    .. code-block:: python3
-
-        dim = 2
-        mesh = RectangleMesh(2, 1, 2., 1., quadrilateral=True)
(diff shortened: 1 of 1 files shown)


Ig-dolci (approved):

firedrake/mesh.py
```diff
@@ -4639,31 +4639,6 @@ def Submesh(mesh, subdim, subdomain_id, label_name=None, name=None):
...
-        subm = Submesh(mesh, dim, 999)
-        V0 = FunctionSpace(mesh, "CG", 1)
-        V1 = FunctionSpace(subm, "CG", 1)
-        V = V0 * V1
-        u = TrialFunction(V)
-        v = TestFunction(V)
-        u0, u1 = split(u)
-        v0, v1 = split(v)
-        dx0 = Measure("dx", domain=mesh)
-        dx1 = Measure("dx", domain=subm)
```
pbrubeck, line 4663: Why not just fix it?

connorjward: The fail happens when we make the mixed space. I decided to just remove it because:

1. I don't really know how to fix it
2. This has changed in main so will immediately bitrot
3. This isn't tested. We should literal include code here instead.

https://firedrakeproject.slack.com/archives/C1Q0Y6H8A/p1771895877638979

PR #4929, opened 2026-02-27
Propagate distribution_parameters to Submesh and RelabeledMesh

pbrubeck:
# Description
Submesh and RelabeledMesh do not redistribute the mesh, meaning that they inherit the parent mesh's distribution. However it was not possible to infer the overlap type and the overlap distance from the `distribution_parameters` attribute. This PR just tags the child mesh with the parent's `distribution_parameters`.

### DIFF
--- a/firedrake/mesh.py
+++ b/firedrake/mesh.py
@@ -4862,7 +4862,7 @@ def RelabeledMesh(mesh, indicator_functions, subdomain_ids, **kwargs):
         dmlabel = plex1.getLabel(dmlabel_name)
         section = f.topological.function_space().dm.getSection()
         dmcommon.mark_points_with_function_array(plex, section, height, f.dat.data_ro_with_halos.real.astype(IntType), dmlabel, subid)
-    reorder_noop = False
+    reorder_noop = None
     tmesh1 = MeshTopology(plex1, name=plex1.getName(), reorder=reorder_noop,
                           distribution_parameters=DISTRIBUTION_PARAMETERS_NOOP,
                           perm_is=tmesh._dm_renumbering,
@@ -4872,6 +4872,7 @@ def RelabeledMesh(mesh, indicator_functions, subdomain_ids, **kwargs):
     rmesh = make_mesh_from_mesh_topology(tmesh1, name1)
     # Tag the relabeled mesh with the original distribution parameters
     rmesh._distribution_parameters = mesh._distribution_parameters
+    rmesh._did_reordering = mesh._did_reordering
     return rmesh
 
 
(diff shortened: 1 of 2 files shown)


firedrake/mesh.py
connorjward: Is this just there to say "don't distribute me again"? That's a pretty bad API.

Your changes are fairly non-obvious because we are overwriting the distribution parameters we pass to the mesh topology constructor. Maybe it's the best solution for now.

firedrake/mesh.py
```diff
@@ -4856,16 +4862,17 @@ def RelabeledMesh(mesh, indicator_functions, subdomain_ids, **kwargs):
         dmlabel = plex1.getLabel(dmlabel_name)
         section = f.topological.function_space().dm.getSection()
         dmcommon.mark_points_with_function_array(plex, section, height, f.dat.data_ro_with_halos.real.astype(IntType), dmlabel, subid)
-    distribution_parameters_noop = {"partition": False,
-                                    "overlap_type": (DistributedMeshOverlapType.NONE, 0)}
-    reorder_noop = None
+    reorder_noop = False
```
pbrubeck: I think `reorder_noop = False` is still wrong, we should propagate reorder from the parent, so that `RelabeledMesh` only differs from the parent in the fact that it has different labels.

firedrake/mesh.py
```diff
@@ -4915,7 +4922,7 @@ def Submesh(mesh, subdim, subdomain_id, label_name=None, name=None, ignore_halo=
         Name of the submesh.
     ignore_halo : bool
         Whether to exclude the halo from the submesh.
-    reorder : bool
+    reorder : bool | None
```
pbrubeck, line 4926: `reorder = None`, means that we grab the default value from `firedrake.parameters["reorder_meshes"]`. I think that a better default for Submesh should be to propagate `reorder` from the parent.

firedrake/mesh.py
```diff
@@ -4856,16 +4862,17 @@ def RelabeledMesh(mesh, indicator_functions, subdomain_ids, **kwargs):
...
         section = f.topological.function_space().dm.getSection()
         dmcommon.mark_points_with_function_array(plex, section, height, f.dat.data_ro_with_halos.real.astype(IntType), dmlabel, subid)
-    distribution_parameters_noop = {"partition": False,
-                                    "overlap_type": (DistributedMeshOverlapType.NONE, 0)}
-    reorder_noop = None
+    reorder_noop = False
     tmesh1 = MeshTopology(plex1, name=plex1.getName(), reorder=reorder_noop,
-                          distribution_parameters=distribution_parameters_noop,
+                          distribution_parameters=DISTRIBUTION_PARAMETERS_NOOP,
                           perm_is=tmesh._dm_renumbering,
```
pbrubeck, line 4868: I think `perm_is` is the mechanism that enforces that the `RelabeledMesh` reordering matches that of the parent.

firedrake/mesh.py
pbrubeck: To enforce consistent ordering as default, `Submesh` should probably use something like `perm_is=extract_the_relevant_subset(mesh._dm_renumbering)`

firedrake/mesh.py
pbrubeck: `reorder` will be ignored if `perm_is` is not None.
https://github.com/firedrakeproject/firedrake/blob/dc905aa3e627ce9954bb29e6c45c900d8df60d1c/firedrake/mesh.py#L561-L564

dham (approved):

Issue #4784, opened 2025-12-18
BUG: ngsPETSc update broke things


connorjward:
The [recent changes to ngsPETSc](https://github.com/NGSolve/ngsPETSc/pull/93) have broken main. The timing is very unfortunate since it coincides with the [ngsPETSc migration](https://github.com/firedrakeproject/firedrake/pull/4782).

For the moment these tests are being skipped (see https://github.com/firedrakeproject/firedrake/commit/f255305101736cf6d4d314a8635649826267249f#diff-cde22284d714546b80b9ff91bed8100fc3e25f0f45571cd4f71bf7409d48267f), but they require a proper fix ASAP.

UZerbinati:
Sorry, I had missed this. Is this still true ? Everything seems to be working on my local. Do we have a replicable error ?

PR #4930, opened 2026-02-27
Netgen: curved meshes from Mesh constructor and refine_marked_elements

pbrubeck:
# Description
First step towards https://github.com/firedrakeproject/firedrake/issues/4784. Currently it restores netgen + MG tests but only with OCC. 

Previously we had 2 methods to construct a curved Mesh out of a netgen mesh:

 1. First creating a firedrake Mesh from a linear netgen mesh, and then calling `curve_field(degree)` to produce new coordinates, then you could create a firedrake Mesh out of the curve field. The final Firedrake Mesh will not know that it comes from a netgen mesh, and therefore cannot be adaptively refined.
```
mesh1 = Mesh(netgen_mesh)
meshp = Mesh(mesh1.curve_field(p))
```
 2. Via the MeshHierarchy constructor, this method produces a MeshHierarchy in which each mesh knows that it comes from a netgen mesh
```
mesh1 = Mesh(netgen_mesh)
mh = MeshHierarchy(mesh1, refine, netgen_flags={"degree": p})
meshp = mh[0]
```

This PR enables the more intuitive approach (while not removing the other two)
```
meshp = Mesh(netgen_mesh, netgen_flags={"degree": p})
```

Also propagates the mesh order in `refine_marked_elements`.

### DIFF
--- a/tests/firedrake/regression/test_netgen.py
+++ b/tests/firedrake/regression/test_netgen.py
@@ -4,8 +4,8 @@ import pytest
 
 
 @pytest.mark.skipnetgen
-@pytest.mark.parallel([1, 3])
-def test_create_netgen_mesh_high_order():
+@pytest.mark.parallel([1, 2])
+def test_netgen_csg_mesh_high_order():
     from netgen.geom2d import Circle, CSG2d
     geo = CSG2d()
     geo.Add(Circle(center=(0, 0), radius=1.0, mat="mat1", bc="circle"))
@@ -123,7 +123,8 @@ def poisson3D(h, degree=2):
 
 
 @pytest.mark.skipnetgen
-def test_firedrake_Poisson_netgen():
+@pytest.mark.parallel([1, 2])
+def test_netgen_csg_poisson_2d():
     diff = np.array([poisson(h)[0] for h in [1/2, 1/4, 1/8]])
     print("l2 error norms:", diff)
     conv = np.log2(diff[:-1] / diff[1:])
@@ -132,17 +133,8 @@ def test_firedrake_Poisson_netgen():
 
 
 @pytest.mark.skipnetgen
-@pytest.mark.parallel
-def test_firedrake_Poisson_netgen_parallel():
-    diff = np.array([poisson(h)[0] for h in [1/2, 1/4, 1/8]])
-    print("l2 error norms:", diff)
-    conv = np.log2(diff[:-1] / diff[1:])
-    print("convergence order:", conv)
-    assert (np.array(conv) > 2.8).all()
-
-
-@pytest.mark.skipnetgen
-def test_firedrake_Poisson3D_netgen():
+@pytest.mark.parallel([1, 2])
+def test_netgen_csg_poisson_3d():
     diff = np.array([poisson3D(h) for h in [1, 1/2, 1/4]])
     print("l2 error norms:", diff)
     conv = np.log2(diff[:-1] / diff[1:])
@@ -151,7 +143,7 @@ def test_firedrake_Poisson3D_netgen():
 
 
 @pytest.mark.skipnetgen
-def test_firedrake_integral_2D_netgen():
+def test_netgen_csg_2d_integral():
     from netgen.geom2d import SplineGeometry
     import netgen
 
@@ -173,7 +165,7 @@ def test_firedrake_integral_2D_netgen():
 
 
 @pytest.mark.skipnetgen
-def test_firedrake_integral_3D_netgen():
+def test_netgen_csg_3d_integral():
     from netgen.csg import CSGeometry, OrthoBrick, Pnt
     import netgen
 
@@ -198,14 +190,15 @@ def test_firedrake_integral_3D_netgen():
 
 
 @pytest.mark.skipnetgen
-def test_firedrake_integral_ball_netgen():
+@pytest.mark.parallel([1, 2])
+def test_netgen_csg_manifold():
     from netgen.csg import CSGeometry, Pnt, Sphere
     from netgen.meshing import MeshingParameters
     from netgen.meshing import MeshingStep
     import netgen
 
     comm = COMM_WORLD
-    if comm.Get_rank() == 0:
+    if comm.rank == 0:
         geo = CSGeometry()
         geo.Add(Sphere(Pnt(0, 0, 0), 1).bc("sphere"))
         mp = MeshingParameters(maxh=0.05, perfstepsend=MeshingStep.MESHSURFACE)
@@ -214,15 +207,48 @@ def test_firedrake_integral_ball_netgen():
         ngmesh = netgen.libngpy._meshing.Mesh(3)
 
     msh = Mesh(ngmesh)
+    assert msh.topological_dimension == 2
+    assert msh.geometric_dimension == 3
+
     V = FunctionSpace(msh, "CG", 3)
-    x, y, z = SpatialCoordinate(msh)
-    f = assemble(interpolate(1+0*x, V))
+    f = assemble(interpolate(Constant(1), V))
     assert abs(assemble(f * dx) - 4*np.pi) < 1.e-2
 
 
 @pytest.mark.skipnetgen
-@pytest.mark.parallel([1, 3])
-def test_firedrake_integral_sphere_high_order_netgen():
+@pytest.mark.parallel([1, 2])
+def test_netgen_occ_manifold():
+    from netgen.occ import Pnt, SplineApproximation, Face, Wire, Axis, OCCGeometry, Z
+    from netgen.meshing import MeshingStep
+    R = 3.0
+    r = 1.5
+    surface_area = R*r*(2*pi)**2
+
+    def Curve(t):
+        return Pnt(0, R+r*np.cos(t), r*np.sin(t))
+
+    n = 100
+    pnts = [Curve(2*np.pi*t/n) for t in range(n+1)]
+
+    spline = SplineApproximation(pnts)
+    f = Face(Wire(spline))
+
+    torus = f.Revolve(Axis((0, 0, 0), Z), 360)
+    geo = OCCGeometry(torus, dim=3)
+    ngmesh = geo.GenerateMesh(maxh=0.5, perfstepsend=MeshingStep.MESHSURFACE)
+
+    msh = Mesh(ngmesh)
+    assert msh.topological_dimension == 2
+    assert msh.geometric_dimension == 3
+
+    V = FunctionSpace(msh, "CG", 3)
+    f = assemble(interpolate(Constant(1), V))
+    assert abs(assemble(f * dx) - surface_area)/surface_area < 5.e-3
+
+
+@pytest.mark.skipnetgen
+@pytest.mark.parallel([1, 2])
+def test_netgen_csg_high_order_integral():
     from netgen.csg import CSGeometry, Pnt, Sphere
     import netgen
 
@@ -236,15 +262,14 @@ def test_firedrake_integral_sphere_high_order_netgen():
 
     homsh = Mesh(ngmesh, netgen_flags={"degree": 2})
     V = FunctionSpace(homsh, "CG", 2)
-    x, y, z = SpatialCoordinate(homsh)
-    f = assemble(interpolate(1+0*x, V))
+    f = assemble(interpolate(Constant(1), V))
     assert abs(assemble(f * dx) - (4/3)*np.pi) < 1.e-2
 
 
 @pytest.mark.skipcomplex
 @pytest.mark.skipnetgen
-@pytest.mark.parallel([1, 3])
-def test_firedrake_Adaptivity_netgen():
+@pytest.mark.parallel([1, 2])
+def test_netgen_occ_adaptivity():
     from netgen.occ import WorkPlane, OCCGeometry, Axes
     from netgen.occ import X, Z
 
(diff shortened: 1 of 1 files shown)


firedrake/mesh.py
```diff
@@ -2922,163 +2924,143 @@ def refine_marked_elements(self, mark, netgen_flags=None):
...
-            )
+        # Construct numpy arrays for physical domain data
+        physical_space_points = np.zeros(
+            (ng_dimension, reference_space_points.shape[0], self.geometric_dimension)
+        )
+        curved_space_points = np.zeros(
+            (ng_dimension, reference_space_points.shape[0], self.geometric_dimension)
+        )
+        self.netgen_mesh.CalcElementMapping(reference_space_points, physical_space_points)
+        # NOTE: This will segfault for CSG!
```
connorjward: Concerning...

firedrake/mesh.py
```diff
@@ -2922,163 +2924,143 @@ def refine_marked_elements(self, mark, netgen_flags=None):
...
+            else:
+                sfBCInv = self.sfBC_orig.createInverse()
+                _, marked0 = self.topology_dm.distributeField(sfBCInv,
+                                                              self._cell_numbering,
+                                                              mvec)
+                mark_np = marked0.getArray()
+
+        netgen_mesh = self.netgen_mesh.Copy()
+        refine_faces = netgen_flags.get("refine_faces", False)
+        if self.comm.rank == 0:
```
connorjward: Can this be wrapped in a `try` `except` block so we don't deadlock if this ends up failing. I did the same thing here: https://github.com/firedrakeproject/firedrake/pull/4913/changes#diff-3b7ad5733ba21ccaa1be5b8aa4027e50a552acea3d0c66425ab7c47bd31ee278

firedrake/mesh.py
```diff
@@ -2922,163 +2924,143 @@ def refine_marked_elements(self, mark, netgen_flags=None):
...
 
         This method requires that the mesh has been constructed from a
         netgen mesh.
 
         :arg order: the order of the curved mesh.
         :arg permutation_tol: tolerance used to construct the permutation of the reference element.
-        :arg location_tol: tolerance used to locate the cell a point belongs to.
-        :arg cg_field: return a CG function field representing the mesh, rather than the
-                       default DG field.
+        :arg cg_field: return a CG function field representing the mesh, as opposed to a DG field.
```
connorjward, line 2986: Should explain the default behaviour.

firedrake/mesh.py
connorjward: We should handle this properly somehow.

firedrake/netgen.py
```diff
@@ -29,6 +29,44 @@ class comp:
             Mesh = type(None)
 
 
+def netgen_distribute(V, netgen_data):
```
connorjward: Can this get some explanation either as a docstring or comments

firedrake/netgen.py
```diff
@@ -29,6 +29,44 @@ class comp:
...
+            plex_data[(..., *i)] = arr
+        plex_data = plex_data.reshape(-1, *nshape[1:])
+    else:
+        plex_data = netgen_data
+    return plex_data
+
+
+def plex_to_netgen_numbering(mesh):
+    plex = mesh.topology_dm
+    fstart, fend = plex.getHeightStratum(0)
```
connorjward: Why is this `fstart`? Surely `cstart`?

tests/firedrake/multigrid/test_netgen_gmg.py
connorjward, line 1: You've changed the tests quite dramatically. Are you totally confident that you've fixed the issue in https://github.com/firedrakeproject/firedrake/issues/4784?

pbrubeck: The tests are basically solving the same PDE with different coordinate degree, so I parametrized them. One issue that still remains is that `ngmesh.Curve()` will segfault for CSG (but works for OCC). This seems to be out of our control, so I changed the tests to use OCC.

firedrake/netgen.py
pbrubeck: Sure, `cstart` makes more sense

firedrake/mesh.py
pbrubeck: I am calling `Barrier()` below, and this seems to prevent deadlocks

firedrake/mesh.py
pbrubeck: We could just raise an exception when `self.netgen_mesh.getGeometry()` is CSG, but what if this get fixed upstream?

firedrake/mesh.py
pbrubeck: Done

firedrake/mesh.py
pbrubeck: Actually, the assumption that rank 0 keeps the global netgen mesh is not always true. A netgen `MeshHierarchy` will distribute the netgen mesh. I think we need a further PR where we make sure that the underlying netgen mesh is also distributed.

firedrake/netgen.py
pbrubeck: Done

firedrake/netgen.py
```diff
@@ -29,6 +30,56 @@ class comp:
...
 
+def netgen_distribute(V: firedrake.functionspaceimpl.WithGeometryBase,
+                      netgen_data: npt.NDArray[np.inexact]):
+    """
+    Distribute data from the netgen layout into the DMPlex layout.
+
+    Parameters
+    ----------
+    V
+        The target function space defining the data DMPlex layout.
```
pbrubeck:
```suggestion
        The target function space defining the DMPlex data layout.
```

demos/netgen/netgen_mesh.py.rst
```diff
@@ -363,15 +362,16 @@ We will now show how to solve the Poisson problem on a high-order mesh, of order
...
        ngmesh = netgen.libngpy._meshing.Mesh(3)
-   mesh = Mesh(Mesh(ngmesh).curve_field(3))
+   mesh = Mesh(ngmesh, netgen_flags={"degree": 3})
+
    # Solving the Poisson problem
    VTKFile("output/MeshExample6.pvd").write(mesh)
    x, y, z = SpatialCoordinate(mesh)
    V = FunctionSpace(mesh, "CG", 3)
-   f = Function(V).interpolate(1+0*x)
+   f = Function(V).interpolate(Constant(1))
```
pbrubeck:
```suggestion
   f = Constant(1)
```

firedrake/mesh.py
connorjward: > I am calling Barrier() below, and this seems to prevent deadlocks

That's not what I mean. If the code that is only running on rank 0 ever fails for some reason then things will hang in a really really hard to debug way (remember [this](https://github.com/firedrakeproject/firedrake/pull/4913)?).

I've added a helper function for doing this in https://github.com/firedrakeproject/firedrake/pull/4972 which is probably good to be merged later today.

> I think we need a further PR where we make sure that the underlying netgen mesh is also distributed.

Sure. I agree that that might be out of scope here. It would be helpful if you could open an issue though.

firedrake/mesh.py
connorjward: I don't think that we should leave this in here. Instead we should:
* Raise an error if a user tries to use CSG that points to the upstream issue
* Open our own issue, again linking to the upstream one, telling us to revisit this at a later point

firedrake/mesh.py
pbrubeck: Could you point me to the function?

pbrubeck: I think I can remove the branching here... The netgen mesh on the other ranks should also be refined in both the redundant and the distributed cases

connorjward: > Could you point me to the function?

`safe_noncollective` in https://github.com/firedrakeproject/firedrake/pull/4972

firedrake/mesh.py
UZerbinati: I agree we should raise an error if the geometry is CSG. I believe CSG is no longer developed by the NGSolve group in favour of OCC 2D and 3D, hence why it seg faults on new features like the one we use to convert back to a Netgen mesh the DMPlex after refinement.

firedrake/netgen.py
UZerbinati: I guess it was facet and cell :)

tests/firedrake/multigrid/test_netgen_gmg.py
UZerbinati: Indeed and we should only use OCC. I think we should also change all demos to only use OCC.

demos/netgen/netgen_mesh.py.rst
```diff
@@ -332,8 +332,7 @@ As usual, we generate a mesh for the described geometry and use the Firedrake-Ne
 High-order Meshes
 ------------------
 It is possible to construct high-order meshes for a geometry constructed in Netgen.
-In order to do so we need to use the ``curve_field`` method of a Firedrake ``Mesh`` object generated from a Netgen mesh.
-In particular, we need to pass the degree of the polynomial field we want to use to parametrise the coordinates of the domain to the ``curve_field`` method, which will return a ``Function`` constructed on a DG space for this purpose. ::
+We can set the degree of the geometry via ``netgen_flags`` keyword argument of the ``Mesh`` constructor. ::
```
pbrubeck:
```suggestion
We can set the degree of the geometry via the ``netgen_flags`` keyword argument of the ``Mesh`` constructor. ::
```

firedrake/mesh.py
```diff
@@ -3364,6 +3346,30 @@ def Mesh(meshfile, **kwargs):
 
     if from_netgen:
         mesh.netgen_mesh = netgen_firedrake_mesh.meshMap.ngMesh
+        mesh.netgen_flags = netgen_flags
+
+        # Curve the mesh, if requested
+        degree = netgen_flags.get("degree", 1)
+        if degree != 1:
+            permutation_tol = netgen_flags.get("permutation_tol", 1e-8)
+            cg = netgen_flags.get("cg", False)
```
```suggestion
            cg = netgen_flags.get("cg", None)
```

tests/firedrake/multigrid/test_netgen_gmg.py
```diff
@@ -1,105 +1,76 @@
...
-    solve(a == L, u, bcs=bcs, solver_parameters={"ksp_type": "cg",
-                                                 "pc_type": "mg"})
-    expect = Function(V).interpolate(exact)
-    assert norm(assemble(u - expect)) <= 1e-6
+    a = inner(grad(u), grad(v)) * dx
+    labels = [i+1 for i, name in enumerate(ngmesh.GetRegionNames(codim=1)) if name in ["surface"]]
+
+    x = SpatialCoordinate(mesh)
+    uexact = dot(x, x)
+    bcs = DirichletBC(V, uexact, labels)
```
pbrubeck:
```suggestion
    uexact = 1-dot(x, x)
    bcs = DirichletBC(V, 0, labels)
```

firedrake/mesh.py
```diff
@@ -2922,163 +2924,146 @@ def refine_marked_elements(self, mark, netgen_flags=None):
...
+
+        # Construct numpy arrays for physical domain data
+        physical_points = np.zeros(
+            (ng_dimension, reference_points.shape[0], self.geometric_dimension)
+        )
+        curved_points = np.zeros(
+            (ng_dimension, reference_points.shape[0], self.geometric_dimension)
+        )
+        self.netgen_mesh.CalcElementMapping(reference_points, physical_points)
+        # NOTE: This will segfault for MeshHierarchy on a netgen CSG geometry
```
pbrubeck, line 3033: Curving a CSG mesh works fine, as long as this mesh has not been uniformly refined. It fails for CSG + `MeshHierarchy`, where we convert `mesh.topology_dm.refine() `into a netgen mesh. So this seems to me like a ngsPETSc issue.

UZerbinati: The new mechanics,which allow for 3D support, is as follows now. We create a refinement in PETSc, we construct the maps between coarse to fine and vice versa in PETSc which is faster (you really see it in 3D) and eventually I hope also load balanced, then we create the corresponding Netgen mesh. We then snap to the geometry and curve, we thus need to attach geometrical information to the mesh. This last feature is only supported for OCC and not for old school CSG. I believe this is the best way to go and I don't think we should insist in supporting the old CSG, it makes no sense since also Netgen is no longer actively developing it. There is the serious risk we will end up having to keep fixing issues arising from legacy code if we do so. Less is more in this case :)

firedrake/mesh.py
```suggestion
        :arg mark: the marking function which is a Firedrake DG0 function 
            with the number of refinements on each cell.
```

.github/workflows/core.yml
```diff
@@ -210,6 +210,7 @@ jobs:
             --extra-index-url https://download.pytorch.org/whl/cpu \
             "$(echo ./firedrake-repo/dist/firedrake-*.tar.gz)[ci]"
 
+          pip install -I git+https://github.com/NGSolve/ngsPETSc.git@pbrubeck/fix/mg
```
pbrubeck: @connorjward is this correct? I am getting a PETSc install error https://github.com/firedrakeproject/firedrake/actions/runs/23502456900/job/68401005907?pr=4930

connorjward: I think you should be passing `--no-deps`. `--no-build-isolation` might also be helpful.

pbrubeck: Thanks. I just replaced the original pip install command with the desired branch

PR #4932, opened 2026-02-27
[DO NOT SQUASH] merge release to main

pbrubeck:
# Description


firedrake/mesh.py
```diff
@@ -4947,31 +4947,6 @@ def Submesh(mesh, subdim, subdomain_id, label_name=None, name=None, ignore_halo=
     ridges to be contained in the quad mesh are shared by at most two
     facets to make the quad mesh orientation algorithm work.
 
-    Examples
```
pbrubeck, line 4950: @connorjward what happens with the example that was removed?

connorjward: What do you mean? I am still happy for this to go. If we want something here it should be tested.

pbrubeck: Ok

PR #4933, opened 2026-02-27
Bugfix with deflation: only deflate successful solves

pefarrell:
There is a small bug in the implementation of deflation. Right now it deflates the final state of the inner solver when it terminates, whether the solve was successful or not. This changes it so that only successful solves are deflated.


pbrubeck (approved):
Approved

Issue #1852, opened 2020-09-23
Periodic Boundary Conditions


florian98765:
Dear Firedrake developers,

I am currently working on some simulations using periodic boundary conditions. I thought it would be a good idea to summarize all information that I could find up to now and trigger a discussion about the implementation:

- What already works within firedrake is using periodic utility meshes like the ```PeriodicSquareMesh``` or the ```PeriodicBoxMesh``` (see #1798 ). 
- #1246 suggests that petsc is able to ready periodic meshes, but one needs to create a proper discontinuous coordinate field.
- last but not least [here](https://lists.mcs.anl.gov/mailman/htdig/petsc-users/2018-October/036542.html) it is mentioned that it could be tricky to generate periodic meshes in gmsh
- in ```dolfin``` one just uses a non-periodic mesh and then defines a master-slave map, which eliminates matching boundary nodes. This map is passed during the creation of the ```FunctionSpace```

From my point of view the ```dolfin``` version has some advantages:
- one can use the same mesh with and without PBCs
- meshing gets easier and the mesh-format does not need to support periodicity
- the ```dolfin``` code is available and can perhaps be adapted to firedrake

I write that the ```dolfin``` code can be adapted, but unfortunately I was not able to do it by myself. At least not without some additional inputs. As far as I could see the original code was commited by Garth N. Wells. Maybe he could tell how much effort it would be to adapt the code (and how to do it :-)

What do you think?

wence-:
> #1246 suggests that petsc is able to ready periodic meshes, but one needs to create a proper discontinuous coordinate field.

I think this is the right thing to fix if one wants a periodic topology.

In some circumstances (e.g. anti-periodic domains) one doesn't want a periodic topology. The way to do that is to put the periodicity into the globaltolocal maps. This could also be done.

medinaeder:
This will be incredibly useful to me as well and I would be willing to give it a shot.

@pefarrell  is this code similar to @florianwechsung firedrake-periodicity?  Currently it is missing parallel, mg, and antiperiodic support.

florian98765:
Sounds great. Maybe we can work together on a cleaned up and generalized version. How can you share the code?

Thanks a lot and best wishes 
Florian

florianwechsung:
I think what @pefarrell is doing in his code is different from firedrake-periodicity. I create a periodic mesh object and the a dg coordinate field. From what Patrick describes, he builds the matrices on the original mesh, and then modifies them afterwards.

[...]

Issue #1246, opened 2018-07-12
Support reading periodic Gmsh files


wence-:
DMPlex can read periodic meshes from Gmsh.  But we can't use them from Firedrake.

The problem is two-fold:

1. We need to be able to specify to the mesh what coordinate field to use (addressed in branch `mesh-coordinate-element`).  This bit is easy
2. We need more generic code to reorder the coordinate field that the DM gives us into the correct one for Firedrake.  I got lost in the reorderings.  This is particularly tricky because in PETSc, they only DGise the coordinates for those cells which are on the periodic "boundary", not everywhere.  So in the rest of the mesh it's still continuous.  Moreover, the periodic vertices have values associated with them.  Quoth Matt:

> You can think of it being multi-values, or you can think of the vertex values as lying on a specific Riemann sheet.

goldsbe:
Has there been any progress made on this?

I am interested in using Firedrake to model acoustic and elastic wave metamaterials, which require periodic BCs applied on a representative unit cell. The geometry of the unit cell is usually very complex, so I generate the geometry in CAD and mesh using Gmsh.

dham:
I don't think this has been fixed. It's not currently on any developer's critical path which makes it unlikely to be fixed rapidly, I'm afraid.

PR #4935, opened 2026-03-01
Loading of periodic meshes generated by gmsh

miguelcoolchips:
## Summary

This PR adds support for loading periodic meshes created by Gmsh (`.msh` files with `$Periodic` sections) into Firedrake. This addresses #1852 and #1246.

## Note

This PR was generated by AI. The tests seem to be working, but frankly I am not sure if the implementation is correct. Any feedback is welcome. I will do my best to address the comments.

### Problem

When PETSc loads a Gmsh file with periodic boundaries via `DMPlexCreateFromFile`, it calls `DMLocalizeCoordinates` which creates cell-local (DG) coordinates **only for cells that straddle the periodic boundary**. However, Firedrake's `reordered_coords` function expects every cell to have an entry in the cell coordinate vector when DG coordinates are present, causing a `ValueError` reshape failure.

### Solution

A new function `_fully_localize_coordinates(dm)` is added to `firedrake/mesh.py` and called in `make_mesh_from_mesh_topology` before the coordinate element type is determined. It expands PETSc's partially localized coordinates to cover all cells by filling in missing entries from CG vertex coordinates via `vecGetClosure`.

The function includes early-return guards so it is a no-op for:
- Non-periodic meshes (`getCoordinatesLocalized()` returns `False`)
- Already fully-localized periodic meshes (e.g. Firedrake's built-in `PeriodicRectangleMesh`)

### Tests

Three Gmsh geometry files and corresponding pre-generated `.msh` files (MSH 2 format) are included as test fixtures:
- `p2d.geo` / `p2d.msh` — 2D rectangle, periodic in x
- `p2d_xy.geo` / `p2d_xy.msh` — 2D rectangle, periodic in x and y
- `p3d.geo` / `p3d.msh` — 3D box, periodic in x

Each test solves a Poisson problem and verifies solution periodicity using `PointEvaluator` at paired probe points across the periodic boundaries. All tests run in both serial and parallel (2 MPI ranks).

## Test plan

- [x] `test_periodic_2d_coordinates` — DG coordinate element is used
- [x] `test_periodic_2d_x_solve` / `_parallel` — Poisson on x-periodic 2D mesh
- [x] `test_periodic_2d_xy_solve` / `_parallel` — Poisson on doubly-periodic 2D mesh
- [x] `test_periodic_3d_solve` / `_parallel` — Poisson on x-periodic 3D mesh
- [x] Existing regression tests (Helmholtz, non-periodic Gmsh, utility periodic meshes) still pass

### DIFF
--- a/tests/firedrake/regression/test_load_mesh.py
+++ b/tests/firedrake/regression/test_load_mesh.py
@@ -68,6 +68,7 @@ def test_periodic_helmholtz_2d_x():
 
     assert errornorm(u_exact_expr, uh, "L2") < 0.005
 
+
 @pytest.mark.parallel([1, 2])
 def test_periodic_2d_xy_solve():
     """Helmholtz on doubly-periodic rectangle [0,0.6]x[0,0.5].
@@ -98,8 +99,6 @@ def test_periodic_2d_xy_solve():
     assert errornorm(u_exact_expr, uh, "L2") / norm(u_exact_expr, "L2") < 0.15
 
 
-
-
 @pytest.mark.parallel([1, 2])
 def test_periodic_3d_solve():
     """Helmholtz on x-periodic box [0,1]^3.
@@ -124,5 +123,3 @@ def test_periodic_3d_solve():
     solve(a == L, uh, bcs=bc, solver_parameters={"ksp_type": "cg"})
 
     assert errornorm(u_exact_expr, uh, "L2") < 1e-12
-
-


tests/periodic/test_periodic.py
```diff
@@ -0,0 +1,141 @@
...
+
+    u_exact_expr = x[1] * (1 - x[1]) * x[2] * (1 - x[2])
+    f_expr = (2 * x[2] * (1 - x[2])
+              + 2 * x[1] * (1 - x[1])
+              + u_exact_expr)
+
+    u = TrialFunction(V)
+    v = TestFunction(V)
+    a = (inner(grad(u), grad(v)) + inner(u, v)) * dx
+    L = inner(f_expr, v) * dx
```
pbrubeck:
```suggestion
    L = a(v, u_exact)
```

tests/periodic/test_periodic.py
```diff
@@ -0,0 +1,141 @@
...
+    """Helmholtz on x-periodic box [0,1]^3.
+
+    Polynomial manufactured solution u_exact = y*(1-y)*z*(1-z).
+    Trivially periodic in x, zero on y/z boundaries.
+    """
+    mesh = Mesh(join(cwd, "geom", "p3d.msh"))
+    V = FunctionSpace(mesh, "CG", 1)
+    x = SpatialCoordinate(mesh)
+
+    u_exact_expr = x[1] * (1 - x[1]) * x[2] * (1 - x[2])
```
pbrubeck:
```suggestion
    u_exact_expr = 42 + x[1] * (1 - x[1]) * x[2] * (1 - x[2])
```

tests/periodic/test_periodic.py
```diff
@@ -0,0 +1,141 @@
...
+
+
+def _run_periodic_helmholtz_3d():
+    """Helmholtz on x-periodic box [0,1]^3.
+
+    Polynomial manufactured solution u_exact = y*(1-y)*z*(1-z).
+    Trivially periodic in x, zero on y/z boundaries.
+    """
+    mesh = Mesh(join(cwd, "geom", "p3d.msh"))
+    V = FunctionSpace(mesh, "CG", 1)
```
pbrubeck:
```suggestion
    V = FunctionSpace(mesh, "CG", 4)
```

tests/periodic/test_periodic.py
```diff
@@ -0,0 +1,141 @@
...
+    u = TrialFunction(V)
+    v = TestFunction(V)
+    a = (inner(grad(u), grad(v)) + inner(u, v)) * dx
+    L = inner(f_expr, v) * dx
+
+    uh = Function(V)
+    bc = DirichletBC(V, Constant(0), [3, 4, 5, 6])
+    solve(a == L, uh, bcs=bc, solver_parameters={"ksp_type": "cg"})
+
+    assert errornorm(u_exact_expr, uh, "L2") < 0.01
```
pbrubeck:
```suggestion
    assert errornorm(u_exact_expr, uh, "L2") < 1E-12
```

tests/periodic/test_periodic.py
```diff
@@ -0,0 +1,138 @@
...
+
+def _run_periodic_helmholtz_2d_xy():
+    """Helmholtz on doubly-periodic rectangle [0,0.6]x[0,0.5].
+
+    Trigonometric manufactured solution
+    u_exact = sin(2*pi*x/0.6) * sin(2*pi*y/0.5), periodic in both
+    x and y.  No boundary conditions needed.
+
+    Uses a wider tolerance than the polynomial tests because a
+    non-trivial doubly-periodic solution must be trigonometric,
```
pbrubeck: The function at the boundary is just zero, and for the other polynomial solutions the boundary data is still just constant. If you use cosine instead of sine you could get non-constant boundary conditions.

I would suggest something like
```
Function(V).interpolate(cos(pi*x)*cos(pi*y))
```

miguelcoolchips: Good point. I changed it to cosine.

firedrake/mesh.py
```diff
@@ -3190,6 +3190,76 @@ def make_mesh_from_coordinates(coordinates, name, tolerance=0.5):
     return mesh
 
 
+def _fully_localize_coordinates(dm: PETSc.DM) -> None:
+    """Expand partially localized coordinates to cover all cells.
+
+    PETSc's ``DMLocalizeCoordinates`` only creates cell-local (DG)
+    coordinates for cells that straddle the periodic boundary.  The
```
connorjward: This isn't quite true. If 'sparse localize' is false then the whole coordinate field is localised. I think `DMLocalizeCoordinates` should be able to do the right thing for us.

miguelcoolchips: Is it then just a matter of exposing DMSetSparseLocalize with petsc4py and setting `sparseLocalize = False`? Or maybe just set the option `dm_sparse_localize 0`?

I also realized that this implementation iterates over the elements. Definitely not the most efficient.

miguelcoolchips: @connorjward are we happy with the current implementation minus the fact that could be done in cython to avoid iterating over cells in python? I can do that next

firedrake/mesh.py
```diff
@@ -3190,6 +3190,92 @@ def make_mesh_from_coordinates(coordinates, name, tolerance=0.5):
...
+    dofs_per_cell = None
+    needs_expansion = False
+    for c in range(cStart, cEnd):
+        dof = cell_sec.getDof(c)
+        if dof > 0:
+            dofs_per_cell = dof
+        else:
+            needs_expansion = True
+
+    if not needs_expansion or dofs_per_cell is None:
```
connorjward: I don't really understand this bit. What case is this capturing?

firedrake/mesh.py
```diff
@@ -3190,6 +3190,92 @@ def make_mesh_from_coordinates(coordinates, name, tolerance=0.5):
...
+
+    By default PETSc's ``DMLocalizeCoordinates`` only creates
+    cell-local (DG) coordinates for cells that straddle the periodic
+    boundary (``sparseLocalize = True``).  Full localization of every
+    cell is possible by calling ``DMSetSparseLocalize(dm, False)``
+    before ``DMLocalizeCoordinates``, and ``createBoxMesh`` already
+    accepts a ``sparseLocalize`` flag for this purpose.
+
+    For file-based periodic meshes (e.g. Gmsh) the localization
+    happens inside ``setFromOptions`` with ``sparseLocalize = True``
```
connorjward: This is wrong. `setFromOptions` is not called in that routine: https://petsc.org/release/src/dm/impls/plex/plexgmsh.c.html#DMPlexCreateGmshFromFile

firedrake/mesh.py
```diff
@@ -3190,6 +3190,92 @@ def make_mesh_from_coordinates(coordinates, name, tolerance=0.5):
...
+
+    The Firedrake DG coordinate path in ``reordered_coords`` expects
+    every cell to have an entry in the cell coordinate vector.
+
+    .. todo::
+
+        Remove this function once petsc4py exposes
+        ``DMSetSparseLocalize``; at that point we can set
+        ``sparseLocalize = False`` before ``setFromOptions`` and let
+        PETSc handle full localization natively.
```
connorjward: I have asked Matt for input on this. Adding bindings to petsc4py is easy so I could just do that once we know the right approach.

firedrake/mesh.py
connorjward: I've asked Matt for advice. Doing it in Cython might be good but honestly I have found that the difference is performance isn't that great.

tests/periodic/test_periodic.py
```diff
@@ -0,0 +1,140 @@
...
+
+@pytest.fixture(params=["p2d", "p2d_xy"])
+def periodic_2d_mesh(request):
+    return Mesh(join(cwd, "geom", f"{request.param}.msh"))
+
+
+def test_periodic_2d_coordinates(periodic_2d_mesh):
+    """Mesh uses a DG coordinate element after loading."""
+    elem = periodic_2d_mesh.ufl_coordinate_element()
+    assert "DG" in str(elem)
```
pbrubeck:
```suggestion
    assert elem.family() == "Discontinuous Galerkin"
```

tests/periodic/test_periodic.py
```diff
@@ -0,0 +1,140 @@
...
+    mesh = Mesh(join(cwd, "geom", "p2d_xy.msh"))
+    V = FunctionSpace(mesh, "CG", 1)
+    x = SpatialCoordinate(mesh)
+
+    Lx, Ly = 0.6, 0.5
+    u_exact = Function(V)
+    u_exact.interpolate(cos(2 * pi * x[0] / Lx) * cos(2 * pi * x[1] / Ly))
+
+    f_coeff = (2 * pi / Lx) ** 2 + (2 * pi / Ly) ** 2 + 1.0
+    f = Function(V).assign(f_coeff * u_exact)
```
pbrubeck:
```suggestion
```

tests/periodic/test_periodic.py
```diff
@@ -0,0 +1,140 @@
...
+
+    u = TrialFunction(V)
+    v = TestFunction(V)
+    a = (inner(grad(u), grad(v)) + inner(u, v)) * dx
+    L = inner(f, v) * dx
```
pbrubeck:
```suggestion
    L = a(v, u_exact)
```

tests/periodic/test_periodic.py
```diff
@@ -0,0 +1,140 @@
...
+
+    uh = Function(V)
+    solve(a == L, uh, solver_parameters={"ksp_type": "cg"})
```
pbrubeck:
```suggestion
    solve(a == L, uh)
```

tests/periodic/test_periodic.py
```diff
@@ -0,0 +1,140 @@
...
+
+    l2err = sqrt(assemble(inner(uh - u_exact, uh - u_exact) * dx))
+    l2norm = sqrt(assemble(inner(u_exact, u_exact) * dx))
+    assert l2err / l2norm < 0.15
```
pbrubeck: This should give zero if `u_exact` is a `Function` in `V`. Additionally, it is more convenient to use `errornorm` and `norm` in the lines above

connorjward:
> I've opened https://gitlab.com/petsc/petsc/-/merge_requests/9110 which, once merged, will make it possible to replace all the PETSc code with:
> 
> ```
> dm.setSparseLocalize(False)
> dm.localizeCoordinates()
> ```

This has now been merged.

tests/periodic/test_periodic.py
miguelcoolchips: Done in 0f929aa.

tests/periodic/test_periodic.py
miguelcoolchips: Done in 0f929aa.

tests/periodic/test_periodic.py
miguelcoolchips: Done in 0f929aa.

tests/periodic/test_periodic.py
miguelcoolchips: Done in 0f929aa.

tests/periodic/test_periodic.py
miguelcoolchips: Already addressed in fd0d1c7.

firedrake/mesh.py
miguelcoolchips: Simplified in 85d2152. The outdated docstring and TODO about `DMSetSparseLocalize` not being in petsc4py have been removed. The function is now only called when coordinates are already localized (guard moved to the call site). The two-line PETSc approach (`setSparseLocalize(False)` + `localizeCoordinates()`) can't be used yet because the Gmsh reader doesn't call `DMSetPeriodicity`, so `localizeCoordinates` is a no-op for Gmsh-loaded meshes.

firedrake/mesh.py
miguelcoolchips: Removed in 85d2152. The `needs_expansion` logic was unnecessary — the function is now only called when `getCoordinatesLocalized()` is true, and the early return only checks if `dofs_per_cell` was found.

firedrake/mesh.py
miguelcoolchips: Fixed in 85d2152 — removed the incorrect mention of `setFromOptions` from the docstring.

firedrake/mesh.py
miguelcoolchips: Outdated — the bindings (`setSparseLocalize`, `localizeCoordinates`) are now in petsc4py. However, the two-line approach doesn't work for Gmsh meshes because the Gmsh reader doesn't call `DMSetPeriodicity`, so `localizeCoordinates` returns early. This would need a PETSc-side change to the Gmsh reader.

firedrake/mesh.py
miguelcoolchips: Noted. Keeping the Python loop for now since the function is only called once during mesh construction.

tests/periodic/test_periodic.py
miguelcoolchips: Done in 0f929aa, though the actual family string is `"Discontinuous Lagrange"` rather than `"Discontinuous Galerkin"`.

tests/periodic/test_periodic.py
miguelcoolchips: Done in 0f929aa.

tests/periodic/test_periodic.py
miguelcoolchips: Done in 0f929aa.

tests/periodic/test_periodic.py
miguelcoolchips: Done in 0f929aa.

tests/periodic/test_periodic.py
miguelcoolchips: Done in 0f929aa — now uses `errornorm` and `norm`.

tests/firedrake/meshes/p3d.geo
connorjward, line 1: Why do we have `.geo` and `.msh` files?

tests/firedrake/regression/test_periodic.py
connorjward, line 1: These tests should go in `tests/firedrake/regression/test_load_mesh.py`

tests/periodic/test_periodic.py
```diff
@@ -0,0 +1,134 @@
...
+    return Mesh(join(cwd, "geom", f"{request.param}.msh"))
+
+
+def test_periodic_2d_coordinates(periodic_2d_mesh):
+    """Mesh uses a DG coordinate element after loading."""
+    elem = periodic_2d_mesh.ufl_coordinate_element()
+    assert elem.family() == "Discontinuous Lagrange"
+
+
+def _run_periodic_helmholtz_2d_x():
```
connorjward:
```suggestion
@pytest.mark.parallel([1, 2])
def test_periodic_helmholtz_2d_x():
```

tests/periodic/test_periodic.py
```diff
@@ -0,0 +1,134 @@
...
+
+def test_periodic_2d_x_solve():
+    _run_periodic_helmholtz_2d_x()
+
+
+@pytest.mark.parallel(nprocs=2)
+def test_periodic_2d_x_solve_parallel():
+    _run_periodic_helmholtz_2d_x()
+
+
```
connorjward:
```suggestion
```

connorjward: And the same for the tests below.

firedrake/mesh.py
```diff
@@ -3190,6 +3190,63 @@ def make_mesh_from_coordinates(coordinates, name, tolerance=0.5):
     return mesh
 
 
+def _fully_localize_coordinates(dm):
```
connorjward, line 3193: As I stated in https://github.com/firedrakeproject/firedrake/pull/4935#issuecomment-4077393622 this can now go away.

connorjward: Just saw your comment above. I am not happy about adding this much code into Firedrake if it can be avoided fairly easily. I will investigate the DMPlex Gmsh code to see what's missing.

connorjward: Ah but PETSc have just released 3.25.0. We definitely want to land this before release so I think this is going to have to stay.

tests/firedrake/meshes/p3d.geo
miguelcoolchips: Looked into this — `gmsh` is not a declared dependency of Firedrake and isn't installed on CI (it's not in `firedrake-configure --show-system-packages` or `pyproject.toml`). Generating at runtime would require adding it as a test dependency. Keeping the committed `.msh` files for now since they're small (~39 KB total) and this is consistent with how other Firedrake tests handle pre-built meshes.

firedrake/mesh.py
connorjward: This is also a fair bit more readable now. I'm happy with this bit of the PR.

tests/firedrake/meshes/p3d.geo
connorjward: Ah OK that's fine then. Please can these meshes go in `tests/firedrake/meshes` which is where all of the other mesh files used in tests are.

tests/periodic/test_periodic.py
connorjward: Please also do this for the other tests in this file

tests/firedrake/regression/test_periodic.py
miguelcoolchips: Done in 89d8ddc — moved test to `tests/firedrake/regression/test_periodic.py` and mesh files (`.geo`/`.msh`) to `tests/firedrake/meshes/`.

connorjward: That isn't the file that I suggested.

miguelcoolchips: Sorry about this. I went too fast

tests/periodic/test_periodic.py
miguelcoolchips: I am not sure what you meant with this commit suggestion. Was it to move them to test_load_mesh.py?

connorjward: I had another accompanying suggestion. Basically you should just have a single test that has
```py
@pytest.mark.parallel([1, 2])
```
so you don't need to have separate serial and parallel tests

miguelcoolchips: Makes sense. I fixed these

connorjward (approved):

PR #4943, opened 2026-03-04
Hypergeometric2F1

pbrubeck:
# Description
Enables codegen for `ufl.hyp2f1`

Depends on https://github.com/inducer/loopy/pull/986 https://github.com/FEniCS/ufl/pull/468

### DIFF
--- a/pyop2/compilation.py
+++ b/pyop2/compilation.py
@@ -344,7 +344,7 @@ class LinuxGnuCompiler(Compiler):
 
     _cflags = ("-fPIC", "-Wall", "-std=gnu11")
     _cxxflags = ("-fPIC", "-Wall")
-    _ldflags = ("-shared", "-lgsl", "-lgslcblas")
+    _ldflags = ("-shared",)
 
     _optflags = ("-march=native", "-O3", "-ffast-math")
     _debugflags = ("-O0", "-g")
--- a/tests/firedrake/regression/test_hypergeometric_function.py
+++ b/tests/firedrake/regression/test_hypergeometric_function.py
@@ -1,31 +1,23 @@
 from firedrake import *
+from scipy.special import hyp2f1 as sp_hyp2f1
 import numpy as np
-import scipy
 import pytest
 
 
-@pytest.mark.skipif(utils.complex_mode, reason="Complex bessel functions are not implemented.")
 def test_hypergeometric_function():
-    sp_hyp2f1 = scipy.special.hyp2f1
-    a = 1
-    b = 1
-    c = 1
     mesh = UnitDiskMesh(3)
     V = FunctionSpace(mesh, "CG", 1)
-
+    u = Function(V)
+    w = Function(V)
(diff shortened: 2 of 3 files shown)


pyop2/compilation.py
```diff
@@ -344,7 +344,7 @@ class LinuxGnuCompiler(Compiler):
 
     _cflags = ("-fPIC", "-Wall", "-std=gnu11")
     _cxxflags = ("-fPIC", "-Wall")
-    _ldflags = ("-shared",)
+    _ldflags = ("-shared", "-lgsl", "-lgslcblas")
```
pbrubeck: @connorjward what does `-shared` mean? This is obviously not the correct way to add libraries, please correct me.

tests/firedrake/regression/test_hypergeometric_function.py
```diff
@@ -0,0 +1,31 @@
+from firedrake import *
+import numpy as np
+import scipy
+import pytest
+
+
+@pytest.mark.skipif(utils.complex_mode, reason="Complex bessel functions are not implemented.")
```
connorjward: We have `skipcomplex` for this no?

pyop2/compilation.py
connorjward: I think `-shared` means to make an `.so` instead of a `.a` file.

Regardless I think you maybe want to add this here: https://github.com/firedrakeproject/firedrake/blob/main/pyop2/global_kernel.py#L399. It might be nice to make this optional though since I don't know if we can be confident that all platforms will have this library available. However doing that would involve passing information from TSFC to PyOP2 which might be a little involved.

tests/firedrake/regression/test_hypergeometric_function.py
pbrubeck: Hmm. I don't think there's an issue with complex for this one. I just cargo-culted it from `test_bessel_functions.py`

tests/firedrake/regression/test_hypergeometric_function.py
```diff
@@ -0,0 +1,23 @@
...
+def test_hypergeometric_function():
+    mesh = UnitDiskMesh(3)
+    V = FunctionSpace(mesh, "CG", 1)
+    u = Function(V)
+    w = Function(V)
+    x, y = SpatialCoordinate(mesh)
+
+    expressions = [((1/3, 1/2, 1), Constant(0.9999) * x),
+                   ((1/2, 1/2, 1), Constant(0.9999) * sqrt(x**2+y**2)),
+                   ((-1/2, 1/2, 1), Constant(0.4999) * sqrt(x*x + y*y))]
```
pbrubeck, line 16: This case is giving a different result for arguments >= 0.5 when compared to scipy

pbrubeck: Maybe this is relevant https://github.com/scipy/scipy/issues/8054

Issue #4948, opened 2026-03-09
BUG: `Submesh` does not propagate "Edge Sets" to submesh exterior facet markers


lorenzoCorintis:
**Describe the bug**
When extracting a 2D submesh from a 3D mesh via `fd.Submesh(mesh3d, 2, tag)`, the boundary facets of the resulting submesh correspond to edges of the parent 3D mesh. However, Firedrake only propagates "Face Sets" labels to the submesh's exterior facet markers — "Edge Sets" are ignored entirely.

This means `submesh.exterior_facets.unique_markers` is missing the edge-based boundary markers, making it impossible to apply boundary conditions on the submesh without manually copying "Edge Sets" into "Face Sets" on the parent mesh before calling `Submesh`.

Actual: The edge-based markers are missing.

Current workaround: Before calling `fd.Submesh`, copy "Edge Sets" into "Face Sets" on the parent mesh via `DMPlex` label manipulation (shown in the MFE).

**Steps to Reproduce**
The attached MFE (firedrake_submesh_edge_sets_mfe.py + submesh_edge_sets_mfe.msh) demonstrates this on a unit cube with:

A tagged face at z=0 (Face Set 100)
Two tagged edges on that face (Edge Sets 1 and 2)
`python firedrake_submesh_edge_sets_mfe.py`

**Expected behavior**
`submesh.exterior_facets.unique_markers` is `[1, 2, 100]`.

**Error message**
No error, but the face sets are incorrectly read, `submesh.exterior_facets.unique_markers` is only `[100]`.

**Environment:**
 - OS: MacOs, Linux
 - Python version: 3.11.14
 - Output of `pip list`:
```
alabaster                     1.0.0
annotated-types               0.7.0
asttokens                     3.0.1
attrs                         25.4.0
babel                         2.18.0
cachetools                    7.0.0
cattrs                        25.2.0
certifi                       2026.1.4
cfgv                          3.5.0
cgen                          2025.1
charset-normalizer            3.4.4
checkpoint_schedules          1.0.4
click                         8.2.1
codepy                        2025.1
colorama                      0.4.6
constantdict                  2025.3
contourpy                     1.3.3
coverage                      7.12.0
cycler                        0.12.1
Cython                        3.1.3
decorator                     4.4.2
defusedxml                    0.7.1
deptry                        0.24.0
distlib                       0.4.0
docutils                      0.21.2
execnet                       2.1.2
executing                     2.2.1
[...]

PR #4950, opened 2026-03-09
Fix `Submesh` to propagate "Edge Sets" / "Vertex Sets" to codimension-1 submesh exterior facet markers

lorenzoCorintis:
# Description
- Fixes https://github.com/firedrakeproject/firedrake/issues/4948 -- `Submesh` now propagates parent "Edge Sets" (3D→2D) and "Vertex Sets" (2D→1D) labels to `"Face Sets"` on codimension-1 submeshes, so that `submesh.exterior_facets.unique_markers` contains the correct boundary markers and `ds(tag)` integration works as expected.
- The previous implementation of `submesh_update_facet_labels` returned early for `subdim != dim`, leaving codimension-1 submeshes with no meaningful facet labels. The new `else` branch maps each submesh exterior facet back to its parent point via `subpoint_is`, looks up the parent label value, and writes it into `"Face Sets"` on the subdm. Exterior facets without a parent label receive a fresh default value (`max(existing_values) + 1`).
- Inherited `"Face Sets"` from `DMPlexFilter` (which sit on cells, not facets) are removed before relabeling to prevent contamination of exterior facet markers.

### DIFF
--- a/firedrake/cython/dmcommon.pyx
+++ b/firedrake/cython/dmcommon.pyx
@@ -11,7 +11,7 @@ from firedrake.petsc import PETSc
 from mpi4py import MPI
 from firedrake.utils import IntType, ScalarType
 from libc.string cimport memset
-from libc.stdlib cimport qsort
+from libc.stdlib cimport qsort, malloc, free
 from finat.element_factory import as_fiat_cell
 
 cimport numpy as np
@@ -1726,74 +1726,17 @@ def label_facets(PETSc.DM plex):
     CHKERR(DMLabelDestroyIndex(lbl_ext))
 
 
-cdef void _populate_lower_dim_labels(PETSc.DM dm, PetscInt tdim):
-    """Copy "Face Sets" values to "Edge Sets" and "Vertex Sets" for
-    lower-dimensional entities.
-
-    After ``DMPlexLabelComplete`` has propagated "Face Sets" into the
-    closure, edges and vertices carry the boundary tag of their parent
-    face.  This helper mirrors those values into the corresponding
-    lower-dimensional label so that a codimension-1 ``Submesh`` can
-    find the information it needs without relying on the inherited
-    "Face Sets" (which mixes cell- and facet-level values).
-
-    Entities that already have a value in the target label are skipped
-    to avoid overwriting user-defined labels (e.g. from Gmsh).
-    """
-    cdef:
-        PetscInt pStart, pEnd, p, face_val, existing_val
-        DMLabel face_sets_lbl, edge_sets_lbl, vertex_sets_lbl
-
-    CHKERR(DMGetLabel(dm.dm, b"Face Sets", &face_sets_lbl))
-
-    if tdim >= 3:
-        if not dm.hasLabel(EDGE_SETS_LABEL):
-            dm.createLabel(EDGE_SETS_LABEL)
-        CHKERR(DMGetLabel(dm.dm, b"Edge Sets", &edge_sets_lbl))
-        CHKERR(DMPlexGetDepthStratum(dm.dm, 1, &pStart, &pEnd))
-        for p in range(pStart, pEnd):
-            CHKERR(DMLabelGetValue(face_sets_lbl, p, &face_val))
-            if face_val >= 0:
-                CHKERR(DMLabelGetValue(edge_sets_lbl, p, &existing_val))
-                if existing_val < 0:
-                    CHKERR(DMLabelSetValue(edge_sets_lbl, p, face_val))
-
-    if tdim >= 2:
-        if not dm.hasLabel(VERTEX_SETS_LABEL):
-            dm.createLabel(VERTEX_SETS_LABEL)
-        CHKERR(DMGetLabel(dm.dm, b"Vertex Sets", &vertex_sets_lbl))
-        CHKERR(DMPlexGetDepthStratum(dm.dm, 0, &pStart, &pEnd))
-        for p in range(pStart, pEnd):
-            CHKERR(DMLabelGetValue(face_sets_lbl, p, &face_val))
-            if face_val >= 0:
-                CHKERR(DMLabelGetValue(vertex_sets_lbl, p, &existing_val))
-                if existing_val < 0:
-                    CHKERR(DMLabelSetValue(vertex_sets_lbl, p, face_val))
-
-
 def complete_facet_labels(PETSc.DM dm):
     """Transfer label values from the facet labels to everything in
-    the closure of the facets.
+    the closure of the facets."""
+    cdef PETSc.DMLabel label
 
-    After completing "Face Sets" this also populates "Edge Sets" and
-    "Vertex Sets" so that boundary information is available at every
-    topological dimension.  Entities that already carry a value in the
-    target label (e.g. user-defined labels from Gmsh) are not overwritten.
-    """
-    cdef:
-        PETSc.DMLabel label
-        PetscInt tdim
-
-    tdim = get_topological_dimension(dm)
-    if tdim == 0:
+    if get_topological_dimension(dm) == 0:
         return
     for name in [FACE_SETS_LABEL, "exterior_facets", "interior_facets"]:
         if dm.hasLabel(name):
             label = dm.getLabel(name)
             CHKERR( DMPlexLabelComplete(dm.dm, label.dmlabel) )
-
-    if dm.hasLabel(FACE_SETS_LABEL):
-        _populate_lower_dim_labels(dm, tdim)
 
 
 @cython.boundscheck(False)
@@ -3886,6 +3829,8 @@ def submesh_create(PETSc.DM dm,
                 CHKERR(DMLabelSetValue(<DMLabel>temp_label.dmlabel, p, label_value))
         CHKERR(ISRestoreIndices(stratum_is, &stratum_indices))
         CHKERR(ISDestroy(&stratum_is))
+    if subdim == dm.getDimension() - 1 and dm.hasLabel(FACE_SETS_LABEL):
+        _populate_lower_dim_labels(dm, dm.getDimension())
     # Make submesh using temp_label.
     subdm, ownership_transfer_sf = dm.filter(label=temp_label,
                                              value=label_value,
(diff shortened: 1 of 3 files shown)


tests/firedrake/submesh/test_submesh_codim1_labels.py
```diff
@@ -0,0 +1,219 @@
...
+    assert markers is not None
+    # All exterior facets should receive the default label (0)
+    assert 0 in set(markers)
+
+
+# ---------------------------------------------------------------------------
+# Parallel
+# ---------------------------------------------------------------------------
+
+@pytest.mark.parallel(nprocs=2)
```
connorjward: I would like to see more parallel tests, especially some that involve calculation, as that is where this functionality is really stressed.

firedrake/cython/dmcommon.pyx
```diff
@@ -4006,7 +4015,56 @@ def submesh_update_facet_labels(PETSc.DM dm, PETSc.DM subdm):
         if sub_ext_facet_is.iset:
             CHKERR(ISRestoreIndices(sub_ext_facet_is.iset, &sub_ext_facet_indices))
     else:
-        raise NotImplementedError("Currently, only implemented for cell submesh")
+        # subdim == dim - 1: codimension-1 submesh.
```
pbrubeck: It'd be preferable to explicitly catch the case `elif subdim == dim - 1:`

lorenzoCorintis: Thanks I changed that

tests/firedrake/submesh/test_submesh_codim1_labels.py
lorenzoCorintis: I added parallel computation tests at 2 and 3 ranks: boundary length integration via ds(tag), coordinate expression integrals over tagged boundaries, CG2 interpolation + boundary integration, and 2D→1D vertex set propagation with ds(tag).
Let me know if there are more specific tests you would like to see

tests/firedrake/submesh/test_submesh_codim1_labels.py
```diff
@@ -0,0 +1,350 @@
...
+
+
+@pytest.mark.parallel(nprocs=2)
+def test_submesh_codim1_edge_sets_propagated_2_procs():
+    _check_edge_sets_propagated()
+
+
+@pytest.mark.parallel(nprocs=3)
+def test_submesh_codim1_edge_sets_propagated_3_procs():
+    _check_edge_sets_propagated(ncells=4)
```
pbrubeck:
```suggestion
@pytest.mark.parallel([1, 2, 3])
def test_submesh_codim1_edge_sets_propagated():
    _check_edge_sets_propagated(ncells=4)
```

tests/firedrake/submesh/test_submesh_codim1_labels.py
```diff
@@ -0,0 +1,350 @@
...
+
+
+@pytest.mark.parallel(nprocs=2)
+def test_submesh_codim1_ds_edge_sets_2_procs():
+    _check_ds_edge_sets()
+
+
+@pytest.mark.parallel(nprocs=3)
+def test_submesh_codim1_ds_edge_sets_3_procs():
+    _check_ds_edge_sets(ncells=4)
```
pbrubeck:
```suggestion
@pytest.mark.parallel([1, 2, 3])
def test_submesh_codim1_ds_edge_sets():
    _check_ds_edge_sets(ncells=4)
```

tests/firedrake/submesh/test_submesh_codim1_labels.py
```diff
@@ -0,0 +1,350 @@
...
+
+
+@pytest.mark.parallel(nprocs=2)
+def test_submesh_codim1_boundary_coord_integral_2_procs():
+    _check_boundary_coordinate_integral()
+
+
+@pytest.mark.parallel(nprocs=3)
+def test_submesh_codim1_boundary_coord_integral_3_procs():
+    _check_boundary_coordinate_integral(ncells=4)
```
pbrubeck:
```suggestion
@pytest.mark.parallel([1, 2, 3])
def test_submesh_codim1_boundary_coord_integral():
    _check_boundary_coordinate_integral(ncells=4)
```

tests/firedrake/submesh/test_submesh_codim1_labels.py
```diff
@@ -0,0 +1,350 @@
...
+
+
+@pytest.mark.parallel(nprocs=2)
+def test_submesh_codim1_interpolate_integrate_2_procs():
+    _check_interpolate_and_integrate()
+
+
+@pytest.mark.parallel(nprocs=3)
+def test_submesh_codim1_interpolate_integrate_3_procs():
+    _check_interpolate_and_integrate(ncells=4)
```
pbrubeck:
```suggestion
@pytest.mark.parallel([1, 2, 3])
def test_submesh_codim1_interpolate_integrate():
    _check_interpolate_and_integrate(ncells=4)
```

tests/firedrake/submesh/test_submesh_codim1_labels.py
```diff
@@ -0,0 +1,350 @@
...
+
+
+@pytest.mark.parallel(nprocs=2)
+def test_submesh_codim1_vertex_sets_ds_2_procs():
+    _check_vertex_sets_ds()
+
+
+@pytest.mark.parallel(nprocs=3)
+def test_submesh_codim1_vertex_sets_ds_3_procs():
+    _check_vertex_sets_ds(ncells=4)
```
pbrubeck:
```suggestion
@pytest.mark.parallel([1, 2, 3])
def test_submesh_codim1_vertex_sets_ds_3_procs():
    _check_vertex_sets_ds(ncells=4)
```

tests/firedrake/submesh/test_submesh_codim1_labels.py
pbrubeck: It'd be nicer to decorate and rename the `_check` functions

tests/firedrake/submesh/test_submesh_codim1_labels.py
```diff
@@ -0,0 +1,350 @@
...
+VTX_B = 2
+
+
+def _make_square_with_vertex_sets(ncells=2):
+    """UnitSquareMesh with Face Sets on y=0 boundary edges and Vertex
+    Sets on the corner vertices of that edge.
+    """
+    mesh = UnitSquareMesh(ncells, ncells)
+    dm = mesh.topology_dm
+    # Remove default Face Sets so we have full control over labels.
```
pbrubeck: Is this low-level managing of labels mainly needed because `RelabeledMesh` will not let you create codim-2 makers?

pbrubeck: Just so that I understand better how the edge markers typically appear for a user, are they easy to create with Gmsh/Netgen? We should consider enabling them in `RelabeledMesh`. Obviously that is a separate PR.

lorenzoCorintis: Yes, the low-level label manipulation is needed because `RelabeledMesh` / `mark_entities` only creates "Face Sets" (codim-1). There's currently no way to create "Edge Sets" or "Vertex Sets" through the Firedrake Python API.

For users, edge markers typically come from the mesh file. In Gmsh you define "Physical Curves" on a 3D geometry and they get exported as "Edge Sets" when PETSc reads the .msh. Netgen similarly supports tagging edges. So they're straightforward to create on the meshing side. The gap is on the Firedrake side where `RelabeledMesh` doesn't support codim-2 (or higher) markers. Agreed that enabling that would be a good separate PR.

firedrake/cython/dmcommon.pyx
```diff
@@ -3937,76 +3937,168 @@ def submesh_correct_entity_classes(PETSc.DM dm,
...
+        The sub-DM whose facet labels are updated.
 
     Notes
     -----
-    This function marks the new exterior facets with current max label value + 1 in "Face Sets".
+    * **Same-dimension** (``subdim == dim``): new exterior facets (those that
+      were interior in the parent) are tagged with ``max("Face Sets") + 1``.
+    * **Codimension-1** (``subdim == dim - 1``): the parent's lower-dimensional
+      labels ("Edge Sets" for 3D, "Vertex Sets" for 2D) are mapped into
+      "Face Sets" on the subdm.  Unlabeled facets get a default value.
```
connorjward:
```suggestion
      "Face Sets" on the subdm.  Unlabeled facets get a default value of ``max(label) + 1``.
```

firedrake/cython/dmcommon.pyx
```diff
@@ -3937,76 +3937,168 @@ def submesh_correct_entity_classes(PETSc.DM dm,
...
+    const PetscInt *subpoint_indices,
+    const PetscInt *sub_ext_facet_indices,
+    PetscInt sub_ext_facet_size,
+    PetscInt subfStart, PetscInt subfEnd,
+):
+    """Codimension-1 helper: map the parent's lower-dimensional label into "Face Sets".
+
+    For a 3D→2D submesh the parent's "Edge Sets" are propagated; for 2D→1D
+    the parent's "Vertex Sets" are used.  Exterior facets that already carry
+    an inherited "Face Sets" value (from ``DMPlexFilter``) are left alone;
```
connorjward: I don't understand this. Can you explain? I would expect that the given "Face Sets" is now just wrong and it seems a bit dodgy to be partially editing the "Face Sets" label.

tests/firedrake/submesh/test_submesh_codim1_labels.py
```diff
@@ -0,0 +1,318 @@
...
+@pytest.mark.parallel([1, 2, 3])
+def test_submesh_codim1_interpolate_integrate():
+    _check_interpolate_and_integrate(ncells=4)
+
+
+# ---------------------------------------------------------------------------
+# Parallel – 2D -> 1D (Vertex Sets)
+# ---------------------------------------------------------------------------
+
+def _check_vertex_sets_ds(ncells=2):
```
connorjward: Just make this the test and stick the `pytest.mark.parallel` here.

connorjward: There are a couple more instances of this elsewhere.

firedrake/cython/dmcommon.pyx
lorenzoCorintis: `DMPlexFilter` copies every label from the parent into the `subdm`, including "Face Sets" values that `DMPlexLabelComplete` propagated to closure entities (edges, vertices). For a codim-1 submesh these inherited edge-level values carry useful boundary information.

I tried discarding the inherited "Face Sets" and rebuilding from scratch, but this breaks `test_submesh_solve_2d_1d_poisson_hermite` (which chains 3D→2D→1D submesh extractions): removing and recreating the label loses PETSc's internal parallel migration state, so the rebuilt label doesn't survive redistribution correctly.

I therefore preserve inherited values and only fill in exterior facets that don't already have one, using the parent's lower-dimensional label or a fresh default. The docstring now explains this.

tests/firedrake/submesh/test_submesh_codim1_labels.py
lorenzoCorintis: Ah yes! Done

firedrake/cython/dmcommon.pyx
connorjward: This seems quite counter intuitive and might cause some confusion down the line. I wonder if there's a simpler solution:

When we propagate "Face Sets" to the rest of the facet closure could we at the same time set "Edge Sets" and "Vertex Sets" for the extra entities.

Consider one of our 3D utility meshes with some marked boundary. "Face Sets" for this boundary would contain all facets, edges and vertices on the boundary, "Edge Sets" would contain all edges and vertices, and "Vertex Sets" just the vertices.

This way when we lose a dimension via submesh all the necessary boundary information is already encoded in "Edge Sets" and so "Face Sets" is safe to discard (or push to "Cell Sets"?).

lorenzoCorintis: I've implemented part 1 of your suggestion: `complete_facet_labels` now populates "Edge Sets" and "Vertex Sets" from the "Face Sets" closure (entities that already carry a value, e.g. from Gmsh, are not overwritten). So after mesh setup a 3D utility mesh with a marked boundary will have:
- "Face Sets": all facets, edges and vertices on the boundary
- "Edge Sets": all edges and vertices
- "Vertex Sets": just the vertices

For part 2 I also switched `_propagate_parent_facet_labels` to read from the subdm's own inherited "Edge Sets" / "Vertex Sets" instead of doing cross-DM queries against the parent via `subpoint_indices`. This simplifies the function signature and removes the parent-DM dependency entirely.

However, I was not able to fully discard inherited "Face Sets" on the subdm (the `removeLabel` / `createLabel` approach). Even when reading exclusively from subdm-local labels, removing and recreating "Face Sets" still breaks `test_submesh_solve_2d_1d_poisson_hermite` (which chains 3D→2D→1D). The issue is that `removeLabel` / `createLabel` loses PETSc's internal parallel label-migration state regardless of where the replacement values come from.

So the current implementation preserves inherited "Face Sets" and only fills in exterior facets that don't already have a value. I think fully discarding inherited "Face Sets" would require PETSc-level changes (e.g. a way to reset label values without destroying the migration SF), which is out of scope here.

connorjward: The issue that I am having is that there is some quite complicated implicit union behaviour going on between the closure-propagated "Face Sets" and any new "Edge Sets" entries. For instance consider having a cube:
```
           +-------+   
          /       /|     
         /       / |     
      1 /       /  |  
       /       /   |     
      /       /    |     
     +-------+  1  +
     |       |    /   
     |       |   /    
   1 |       |  / 
     |       | /
     |       |/
     +-------+
```
where the left-most edges are labelled 1 in "Edge Sets" and the right-most face is labelled 1 in "Face Sets". If you take some sort of codim-1 submesh it is not obvious which edges (now faces) should be labelled 1.

I think my preferred approach is:
1. *Only* "Edge Sets" are used to build "Face Sets" for the submesh
2. Build "Edge Sets" and "Vertex Sets" *only* for our utility meshes, instead of always propagating from "Face Sets" - this is optional for this PR since my guess is you are mostly using gmsh.

Sorry for making this such a big deal, but I think we have to be careful about this.

> However, I was not able to fully discard inherited "Face Sets" on the subdm (the removeLabel / createLabel approach).

What about something like [DMLabelClearStratum](https://petsc.org/release/manualpages/DMLabel/DMLabelClearStratum/)?

> Even when reading exclusively from subdm-local labels, removing and recreating "Face Sets" still breaks test_submesh_solve_2d_1d_poisson_hermite

This is weird. I wonder what is happening there.

lorenzoCorintis: @connorjward 

Thanks for the suggestion — we've now implemented the approach you described:
**1. Populate "Edge Sets" and "Vertex Sets" from "Face Sets" on the parent**
Before `DMPlexFilter`, a new helper `_populate_lower_dim_labels` iterates each positive "Face Sets" stratum on the parent mesh and copies the value to "Edge Sets" (depth-1 entities) and "Vertex Sets" (depth-0 entities). Entities that already carry a positive value (e.g. user-defined labels from Gmsh) are preserved.
**2. Clear inherited "Face Sets" on the subdm and rebuild from "Edge Sets"**
`_propagate_parent_facet_labels` now uses `DMLabelClearStratum` to clear every stratum of the inherited "Face Sets" on the subdm, then rebuilds it purely from the subdm's own "Edge Sets" (3D→2D) or "Vertex Sets" (2D→1D). `DMPlexLabelComplete` is called afterwards to propagate to closure entities and synchronise ghost points via the point SF. When no source label exists, a fallback path preserves inherited values.
**3. Handling mesh-generator artifacts (value 0)**
The main difficulty we hit was that mesh files (and `RelabeledMesh`) can leave "Face Sets" value 0 on many boundary entities. Since `DMLabelGetValue` returns the *minimum* value for a point, edges at face intersections appeared to have value 0 rather than their meaningful positive value. The fix was to:
- Iterate per-stratum (via `DMLabelGetStratumIS`) rather than relying on `DMLabelGetValue`
- Clear stale value-0 entries with `DMLabelClearValue` before setting the correct positive value
- Skip stratum 0 entirely during derivation
All existing submesh tests pass, including `test_submesh_solve_2d_1d_poisson_hermite` (nprocs=7) which chains 3D→2D→1D submesh extractions via `RelabeledMesh`.

firedrake/cython/dmcommon.pyx
```diff
@@ -3937,77 +3941,288 @@ def submesh_correct_entity_classes(PETSc.DM dm,
     CHKERR(DMLabelDestroyIndex(lbl_ghost))
 
 
+cdef void _populate_lower_dim_labels(PETSc.DM dm, PetscInt tdim):
```
connorjward, line 3944: In my feedback I suggested that this only be done for our utility meshes (i.e. not gmsh). This still has the confusing intersection behaviour.

firedrake/cython/dmcommon.pyx
```diff
@@ -3937,77 +3941,288 @@ def submesh_correct_entity_classes(PETSc.DM dm,
...
+        DMLabel face_sets_lbl, target_lbl
+        PETSc.PetscIS stratum_is = NULL
+        const PetscInt *stratum_pts = NULL
+
+    CHKERR(DMGetLabel(dm.dm, b"Face Sets", &face_sets_lbl))
+    CHKERR(DMLabelGetNumValues(face_sets_lbl, &nvals))
+    if nvals == 0:
+        return
+
+    cdef PetscInt *vals = <PetscInt *>malloc(nvals * sizeof(PetscInt))
```
connorjward, line 3967: We prefer PetscMalloc and PetscFree

firedrake/cython/dmcommon.pyx
connorjward: Maybe this code isn't even necessary because if "Edge Sets" isn't defined then we can use bits from "Face Sets". The thing is we have to avoid blending the two which means if "Edge Sets" is defined we shouldn't be using the original "Face Sets" to build the "Face Sets" for the sub dm.

Are there cases where this doesn't work? If this is critical functionality we could consider relabelling "Edge Sets" with some offset to preserve uniqueness. WDYT?

PR #4166, opened 2025-03-27
Offloading CUDA

Olender:
# Description

PR #4953, opened 2026-03-10
GPU offloading preconditioner

dsroberts:
# Description

First pass at `firedrake-configure` for GPU-enabled PETSc builds, as well as a corresponding github actions workflow. An optional `--gpu-arch` flag has been added which adds all the necessary configuration to go from a fresh Ubuntu installation to a working Firedrake build (mostly) following existing build instructions. `OffloadPC` itself is more or less unchanged from @Olender's work for now, but with a few extra checks for whether GPU offload is available. I expect it will take a few iterations to get this going under CI, so once that's up and running reliably we can look at further development to the `OffloadPC` functionality and testing.

I've tried to put this together in a way that can be expanded upon fairly easily. We're interested in ROCm/HIP as well as Cuda, so there are a couple of dictionaries that just have a `cuda` key for now that will be expanded upon in time. I also thought that `OffloadPC` should be a no-op with a warning if there are any issues around GPUs, rather than crashing out. I think this is the right choice for workflow portability and heterogeneous systems, but it would be just as easy to raise exceptions there.


firedrake/preconditioners/offload.py
```diff
@@ -0,0 +1,105 @@
...
+    def apply(self, pc, x, y):
+        if self.offload_mat_type is None:
+            self.pc.apply(x, y)
+        else:
+            with PETSc.Log.Event("Event: apply offload"):  #
+                dm = pc.getDM()
+                with dmhooks.add_hooks(dm, self, appctx=self._ctx_ref):
+                    with PETSc.Log.Event("Event: vectors offload"):
+                        y_cu = PETSc.Vec()  # begin
+                        y_cu.createCUDAWithArrays(y)
```
pbrubeck:
```suggestion
                        y_cu = PETSc.Vec().createCUDAWithArrays(y)
```

firedrake/preconditioners/offload.py
```diff
@@ -0,0 +1,105 @@
...
+                        x_cu = PETSc.Vec()
+                        # Passing a vec into another vec doesnt work because original is locked
+                        x_cu.createCUDAWithArrays(x.array_r)
```
pbrubeck:
```suggestion
                        # Passing a vec into another vec doesnt work because original is locked
                        x_cu = PETSc.Vec().createCUDAWithArrays(x.array_r)
```

firedrake/preconditioners/offload.py
```diff
@@ -0,0 +1,105 @@
...
+                    with PETSc.Log.Event("Event: solve"):
+                        self.pc.apply(x_cu, y_cu)
+                        # Calling data to synchronize vector
+                        tmp = y_cu.array_r  # noqa: F841
```
pbrubeck:
```suggestion
                        y_cu.array_r
```

firedrake/preconditioners/offload.py
```diff
@@ -0,0 +1,105 @@
...
+
+    Internally this makes a PETSc PC object that can be controlled by
+    options using the extra options prefix ``offload_``.
+    """
+
+    _prefix = "offload_"
+
+    def initialize(self, pc):
+        # Check if our PETSc installation is GPU enabled
+        super().initialize(pc)
```
pbrubeck: Should we be calling `AssembledPC.initialize()`? It seems to me that this will trigger reassembly and we are not really making use of it.

firedrake/preconditioners/offload.py
```diff
@@ -0,0 +1,105 @@
...
+            warnings.warn(
+                "A GPU-enabled Firedrake build has been detected, but a GPU device was"
+                "unable to be initialised. OffloadPC will do nothing."
+            )
+        return None
+    dev.destroy()
+    return mat_type
+
+
+class OffloadPC(AssembledPC):
```
pbrubeck: Should the name be more precise and refer to "GPU"?

firedrake/preconditioners/offload.py
```diff
@@ -0,0 +1,105 @@
...
+                with PETSc.Log.Event("Event: matrix offload"):
+                    P_cu = P.convert(self.offload_mat_type)  # todo
+
+            # Transfer nullspace
+            P_cu.setNullSpace(P.getNullSpace())
+            P_cu.setTransposeNullSpace(P.getTransposeNullSpace())
+            P_cu.setNearNullSpace(P.getNearNullSpace())
+
+            # Update preconditioner with GPU matrix
+            self.pc.setOperators(A, P_cu)
```
pbrubeck: `P` lives in the GPU but `A` still lives in the CPU. When `pc_type = ksp`, the more sensible thing would be to have both `A` and `P` in the GPU, don't you agree?

tests/firedrake/offload/test_poisson_offloading_pc.py
```diff
@@ -0,0 +1,64 @@
...
+)
+def test_poisson_offload(ksp_type, pc_type):
+
+    # Different tests for poisoon: cg and pctype sor, --ksp_type=cg --pc_type=gamg
+    print(f"Using ksp_type = {ksp_type}, and pc_type = {pc_type}.", flush=True)
+
+    nested_parameters = {
+        "pc_type": "ksp",
+        "ksp": {
+            "ksp_type": ksp_type,
```
pbrubeck, line 10: ksp_max_it should be capped, to ensure that we are not taking an insane number of iterations.

tests/firedrake/offload/test_poisson_offloading_pc.py
```diff
@@ -0,0 +1,64 @@
...
+    # Equations
+    L = inner(grad(u), grad(v)) * dx
+    R = inner(v, f) * dx
+
+    # Dirichlet boundary on all sides to 0
+    bcs = DirichletBC(V, 0, "on_boundary")
+
+    # Exact solution
+    sol = Function(V)
+    sol.interpolate(sin(pi * x) * sin(pi * y))
```
pbrubeck: I would suggest to use the interpolated expression as the exact solution so that we can verify that the solver is converging to the expected solution
```suggestion
    R = action(L, sol)
```

tests/firedrake/offload/test_poisson_offloading_pc.py
```diff
@@ -0,0 +1,64 @@
...
+    u = TrialFunction(V)
+    v = TestFunction(V)
+
+    f = Function(V)
+    x, y = SpatialCoordinate(mesh)
+    f.interpolate(2 * pi**2 * sin(pi * x) * sin(pi * y))
+
+    # Equations
+    L = inner(grad(u), grad(v)) * dx
+    R = inner(v, f) * dx
```
pbrubeck:
```suggestion
```

tests/firedrake/offload/test_poisson_offloading_pc.py
```diff
@@ -0,0 +1,64 @@
...
+
+    # Solution function
+    u_f = Function(V)
+
+    problem = LinearVariationalProblem(L, R, u_f, bcs=bcs)
+    solver = LinearVariationalSolver(problem, solver_parameters=parameters)
+    solver.solve()
+    error = errornorm(problem.u, sol)
+    print(f"Error norm = {error}", flush=True)
+    assert error < 1.2e-2
```
pbrubeck:
```suggestion
    assert error < 1.0e-9
```

tests/firedrake/offload/test_poisson_offloading_pc.py
```diff
@@ -0,0 +1,64 @@
...
+    sol = Function(V)
+    sol.interpolate(sin(pi * x) * sin(pi * y))
+
+    # Solution function
+    u_f = Function(V)
+
+    problem = LinearVariationalProblem(L, R, u_f, bcs=bcs)
+    solver = LinearVariationalSolver(problem, solver_parameters=parameters)
+    solver.solve()
+    error = errornorm(problem.u, sol)
```
pbrubeck:
```suggestion
    error = errornorm(u_f, sol)
```

tests/firedrake/offload/test_poisson_offloading_pc.py
```diff
@@ -0,0 +1,64 @@
...
+
+    f = Function(V)
+    x, y = SpatialCoordinate(mesh)
+    f.interpolate(2 * pi**2 * sin(pi * x) * sin(pi * y))
+
+    # Equations
+    L = inner(grad(u), grad(v)) * dx
+    R = inner(v, f) * dx
+
+    # Dirichlet boundary on all sides to 0
```
pbrubeck: To catch potential issues with BCs, it'd be good to prescribe non-homogeneous BCs

firedrake/preconditioners/offload.py
pbrubeck: Is this because `A` cannot live in the GPU if it is matfree? Do we benefit much from `A` being matfree if `P` is not matfree?

.github/workflows/core.yml
```diff
@@ -23,6 +23,10 @@ on:
         description: Whether to test using macOS
         type: boolean
         default: false
+      test_cuda:
```
connorjward: This is missing a corresponding entry in `pr.yml` (https://github.com/firedrakeproject/firedrake/blob/main/.github/workflows/pr.yml#L14). We want to run these if we attach the right label to the PR.

.github/workflows/core.yml
```diff
@@ -473,6 +481,125 @@ jobs:
...
+          export $(python3 ./firedrake-repo/scripts/firedrake-configure --arch default --gpu-arch cuda --show-env)
+          python3 -m venv venv
+          . venv/bin/activate
+
+          : # Empty the pip cache to ensure that everything is compiled from scratch
+          pip cache purge
+
+          : # Fix for petsc4py+slepc4py build
+          echo 'setuptools<81' > constraints.txt
+          export PIP_CONSTRAINT=constraints.txt
```
connorjward: No longer needed.

.github/workflows/core.yml
```diff
@@ -473,6 +481,125 @@ jobs:
...
+            --no-binary h5py \
+            './firedrake-repo[check]'
+
+          firedrake-clean
+          pip list
+
+      - name: Run smoke tests
+        run: |
+          . venv/bin/activate
+          firedrake-check
```
connorjward, line 600: I think ultimately we will want to run all of the Firedrake test suite for this configuration but to start with we definitely need to include a GPU test in `firedrake-check`. I last added a new test in https://github.com/firedrakeproject/firedrake/pull/4759/changes so you can see what to do.

I recommend adding a `skipnogpu` pytest marker and handle it in the same way we have `skipcomplex` [here](https://github.com/firedrakeproject/firedrake/blob/main/tests/firedrake/conftest.py#L174).

firedrake/preconditioners/offload.py
```diff
@@ -0,0 +1,105 @@
...
+            warnings.warn(
+                "This installation of Firedrake is GPU-enabled, but no GPU device has"
+                "been detected. OffloadPC will do nothing on this host"
+            )
+        return None
+    if dev.getDeviceType() == "HOST":
+        if pc_comm_rank == 0:
+            warnings.warn(
+                "A GPU-enabled Firedrake build has been detected, but a GPU device was"
+                "unable to be initialised. OffloadPC will do nothing."
```
connorjward: These 2 warnings seem like maybe they should be breakages instead as they indicate that something has gone properly wrong.

firedrake/preconditioners/offload.py
```diff
@@ -0,0 +1,105 @@
...
+
+
+device_mat_type_map = {"cuda": "aijcusparse"}
+
+
+@cache
+def offload_mat_type(pc_comm_rank) -> str | None:
+    for device, mat_type in device_mat_type_map.items():
+        if device in petsctools.get_external_packages():
+            break
```
connorjward: This feels like it should be a utility function. It's very opaque.

firedrake/preconditioners/offload.py
```diff
@@ -0,0 +1,105 @@
...
+                        self.pc.apply(x_cu, y_cu)
+                        # Calling data to synchronize vector
+                        tmp = y_cu.array_r  # noqa: F841
+                with PETSc.Log.Event("Event: vectors copy back"):
+                    y.copy(y_cu)  #
+
+    def applyTranspose(self, pc, X, Y):
+        raise NotImplementedError
+
+    def view(self, pc, viewer=None):
```
connorjward, line 110: @JHopeCollins any ideas on the nicest way to do the views? I know you have a lot of experience with this.

.github/workflows/core.yml
```diff
@@ -473,6 +481,125 @@ jobs:
...
+      # Git is needed for actions/checkout and Python for firedrake-configure
+      # curl/gnupg2 and ca-certificates needed for adding new deb repositories to ubuntu
+      - name: Install system dependencies (1)
+        run: apt-get -y install git python3 gnupg2 curl ca-certificates
+
+      - name: Add Nvidia CUDA deb repositories
+        run: |
+          curl -fsSL $( python3 ./firedrake-repo/scripts/firedrake-configure --gpu-arch cuda --show-extra-repo-key ) | apt-key add -
+          python3 ./firedrake-repo/scripts/firedrake-configure --gpu-arch cuda --show-extra-repo-file-contents > \
+            /etc/apt/sources.list.d/$( python3 ./firedrake-repo/scripts/firedrake-configure --gpu-arch cuda --show-extra-repo-file-name )
```
connorjward: It's kinda horrid that we have to invoke `firedrake-configure` 3 times. Is there a nicer way to do this?

scripts/firedrake-configure
```diff
@@ -40,6 +49,24 @@ ARCH_COMPLEX = FiredrakeArch.COMPLEX
...
+}
+# Structure is ( deb_repo_filename, file_contents, GPG_key_URL )
+# A file named /etc/apt/sources.list.d/deb_repo_filename containing file_contents will be created
+# The output of curl -fsSL GPG_Key_URL will be passed to 'apt-key add'
+EXTRA_LINUX_APT_REPO = {
+    NO_GPU: ("", "", ""),
+    CUDA: (
+        "cuda.list",
+        f"deb https://developer.download.nvidia.com/compute/cuda/repos/ubuntu2404/{CUDA_ARCH_MAP.get(platform.machine(), platform.machine())} /",
+        f"https://developer.download.nvidia.com/compute/cuda/repos/ubuntu2404/{CUDA_ARCH_MAP.get(platform.machine(), platform.machine())}/3bf863cc.pub",
```
connorjward: This includes the magic string '3bf863cc'. Can there be a comment saying how we can update this?

scripts/firedrake-configure
```diff
@@ -255,12 +349,14 @@ MINIMAL_SYSTEM_PACKAGES = {
 }
 
 SYSTEM_PACKAGES = {
-    (LINUX_APT_X86_64, ARCH_DEFAULT): LINUX_APT_PACKAGES,
-    (LINUX_APT_X86_64, ARCH_COMPLEX): LINUX_APT_PACKAGES,
-    (LINUX_APT_AARCH64, ARCH_DEFAULT): LINUX_APT_PACKAGES,
-    (LINUX_APT_AARCH64, ARCH_COMPLEX): LINUX_APT_PACKAGES,
-    (MACOS_HOMEBREW_ARM64, ARCH_DEFAULT): MACOS_HOMEBREW_PACKAGES,
-    (MACOS_HOMEBREW_ARM64, ARCH_COMPLEX): MACOS_HOMEBREW_PACKAGES,
+    (LINUX_APT_X86_64, ARCH_DEFAULT, NO_GPU): LINUX_APT_PACKAGES,
```
connorjward: I think I'd prefer it if packages were purely additive. So add `libsuperlu-dev` etc here instead of filtering them out above.

scripts/firedrake-configure
```diff
@@ -351,13 +457,34 @@ PETSC_EXTERNAL_PACKAGE_SPECS = {
...
         LINUX_APT_X86_64: PETSC_AUTODETECT,
         LINUX_APT_AARCH64: PETSC_AUTODETECT,
         MACOS_HOMEBREW_ARM64: "/opt/homebrew/opt/zlib",
     },
 }
 
+PetscSpecsDeltaDictType = dict[GPUArch, PetscSpecsDictType]
+
+# Suitesparse and SuperLU-DIST must be built from source to enable GPU features
+PETSC_EXTERNAL_PACKAGE_SPECS_DELTA_GPU: PetscSpecsDeltaDictType = {
```
connorjward: Same comment about an additive approach. I want firedrake-configure to be as dumb as possible.

connorjward (changes requested):

firedrake/preconditioners/offload.py
JHopeCollins: I think this is fine, it's what we do for the assembled and auxiliary PCs. I don't think there's really any information in this class that isn't in `self.pc`.

scripts/firedrake-configure
JHopeCollins: I am very uncomfortable with this. Unless I have misunderstood what is going on this is potentially introducing a security risk for both us and our users.

Should we really be telling our users (many of whom will blindly follow installation guide) to add random authentication keys to their package managers? And even if we are happy to do that, should we be doing it through a deprecated utility that will be removed after Ubuntu24.04 (according to the `apt-key` man page)?

Even if we can make sure that this specific repo at this specific time is safe, are we confident we can reliably keep the links up to date, and if we allow one then inevitably it's only a matter of time before someone adds another. Can we keep up to date with that too?

AFAIK we currently only make users install packages that are already available through their package manager (so the security responsibility is predominantly with them), or through PETSc, which is a much bigger library with more developer and maintainer resources.

I think we need to discuss this at the meeting and work out a) what size of risk it really is, and b) if it is a risk, the best way around this if offloading really does need libraries that can't be installed via package manager or PETSc.

.github/workflows/core.yml
```diff
@@ -473,6 +481,125 @@ jobs:
...
+        run: find . -delete
+
+      # Use a different mirror to fetch apt packages from to get around
+      # temporary outage.
+      # (https://askubuntu.com/questions/1549622/problem-with-archive-ubuntu-com-most-of-the-servers-are-not-responding)
+      # The mirror was chosen from https://launchpad.net/ubuntu/+archivemirrors.
+      - name: Configure apt
+        run: |
+          sed -i 's|http://archive.ubuntu.com/ubuntu|http://www.mirrorservice.org/sites/archive.ubuntu.com/ubuntu/|g' /etc/apt/sources.list.d/ubuntu.sources
+          apt-get update
```
JHopeCollins: @connorjward can't we get rid of this now?

connorjward: I expect so

firedrake/preconditioners/offload.py
dsroberts: A GPU-enabled PETSc should be fully functional on a CPU-only system as well. Having the first of these two issue a warning only allows the same solver settings to run on e.g. the CPU partition and GPU partition of an HPC system. Our main system is very heterogeneous, so not having to maintain separate CPU and GPU solver settings would be advantageous for us. As for the second check, a 'device' can be initialised on a CPU-only PETSc build, but it will always be the `HOST` device, so the initial check of whether or not PETSc has been built with a compatible GPU type should be enough. So if this test is ever true, then yes, we should raise an exception here.

.github/workflows/core.yml
dsroberts: Agreed, its disgusting. I was following the procedure nvidia uses to build their Docker containers. I'll see if I can adapt the procedure they document for end-users, and hope its similar enough to AMD's procedure for HIP.

scripts/firedrake-configure
dsroberts: As above, will rework this part.

firedrake/preconditioners/offload.py
connorjward: OK. I didn't appreciate that you would want to be running a GPU-enabled PETSc on a non-GPU system. I agree with your assessment.

.github/workflows/core.yml
```diff
@@ -465,6 +473,121 @@ jobs:
...
+        run: echo "HOME=/root" >> "$GITHUB_ENV"
+
+      - name: Pre-run cleanup
+        # Make sure the current directory is empty
+        run: find . -delete
+
+      # Git is needed for actions/checkout and Python for firedrake-configure
+      # curl needed for adding new deb repositories to ubuntu
+      - name: Install system dependencies (1)
+        run: apt-get -y install git python3 curl
```
connorjward: Missing an `apt-get update` here now the extra apt repo step is gone.

.github/workflows/core.yml
```diff
@@ -465,6 +473,121 @@ jobs:
...
+
+      - name: Add Nvidia CUDA deb repositories
```
connorjward, line 529: Maybe do this earlier in the workflow.

connorjward (changes requested):
I think this is getting better. The CI error now appears to be a more mundane missing `apt-get update`.

.github/workflows/core.yml
dsroberts: Not sure how much earlier it can go, `nvidia-smi` is a basic sanity check, its provided as part of the Nvidia drivers and docker should mount it from the host. If that's missing then something has gone badly wrong and there is no point continuing. In `test_linux`, `Fix HOME` runs before `apt-get update` so it seemed sensible to keep the order the same. I think the only flexible part of those first few steps is `Pre-run cleanup`, so I've moved `Install system dependencies (1)` before that.

connorjward: I was only suggesting this so you only have to run `apt-get update` once.

firedrake/preconditioners/offload.py
```diff
@@ -0,0 +1,98 @@
...
+
+__all__ = ("OffloadPC",)
+
+
+@cache
+def offload_mat_type(pc_comm_rank) -> str | None:
+    mat_type = device_matrix_type()
+    if mat_type is None:
+        if pc_comm_rank == 0:
+            warnings.warn(
```
connorjward: For Python warnings there is a difference between `warnings.warn` and `logger.warning`. The former indicates a genuine problem and the latter indicates a _likely_ problem. I wonder if the first warning here should be a `warnings.warn` and the second a `logger.warning`.

tests/firedrake/offload/test_poisson_offloading_pc.py
```diff
@@ -0,0 +1,64 @@
...
+
+# TODO: add marker for cuda pytests and something to check if cuda memory was really used
+@pytest.mark.skipnogpu
+@pytest.mark.parametrize(
+    "ksp_type, pc_type", [("cg", "sor"), ("cg", "gamg"), ("preonly", "lu")]
+)
+def test_poisson_offload(ksp_type, pc_type):
+
+    # Different tests for poisson: cg and pctype sor, --ksp_type=cg --pc_type=gamg
+    print(f"Using ksp_type = {ksp_type}, and pc_type = {pc_type}.", flush=True)
```
connorjward: Shouldn't leave print statements in tests (but fine for debugging)

connorjward: Shouldn't leave print statements in tests (but fine for debugging)

.github/workflows/core.yml
```diff
@@ -465,6 +473,141 @@ jobs:
...
+        run: find . -delete
+
+      - uses: actions/checkout@v5
+        with:
+          path: firedrake-repo
+          ref: ${{ inputs.source_ref }}
+
+      - name: Add Nvidia CUDA deb repositories
+        run: |
+          deburl=$( python3 ./firedrake-repo/scripts/firedrake-configure --show-extra-repo-pkg-url --gpu-arch cuda )
```
connorjward, line 531: This process definitely seems better.

.github/workflows/core.yml
```diff
@@ -465,6 +473,141 @@ jobs:
...
+        run: |
+          . venv/bin/activate
+          firedrake-check
+        timeout-minutes: 10
+
+      - name: Verify GPU usage
+        run: |
+          . venv/bin/activate
+          export PETSC_OPTIONS="${PETSC_OPTIONS} -log_view_gpu_time -log_view"
+          python3 ./firedrake-repo/tests/firedrake/offload/test_poisson_offloading_pc.py
```
connorjward: What is this doing?

.github/actionlint.yaml
```diff
@@ -0,0 +1,3 @@
+self-hosted-runner:
+  labels:
+    - gpu
```
connorjward, line 4: What does this do?

connorjward (changes requested):
Looking pretty good. Please let us know when you think this is ready for a detailed review.

firedrake/preconditioners/offload.py
dsroberts: Thanks, I wasn't aware of the distinction. Ironic given that if you set `logging.captureWarnings = True`, the function called by `warnings.warn` to show the message is monkey-patched to a function that calls `logging.warning`. Will update accordingly

.github/workflows/core.yml
dsroberts: Ah, OK. I don't think that's possible given we're adding a new repository in the next step. The github-provided Ubuntu image has a lot included as standard (https://github.com/actions/runner-images/blob/main/images/ubuntu/Ubuntu2404-Readme.md), so we don't necessarily need to run the first `apt-get install` step, however, I'd be uncomfortable removing that should github decide to change their policy and provide minimal images in future.

.github/workflows/core.yml
dsroberts: This is a final sanity check on my end, I wanted to confirm that the GPU was being used as expected. The final column of the profiling output (https://github.com/firedrakeproject/firedrake/actions/runs/23132801814/job/67189609416) shows the fraction of work done on the GPU, I wanted to make sure that the low-level vector and matrix operations were actually happening on the GPU, which they are. Besides, this change is in a 'DROP BEFORE MERGE' commit, it will be going away.

.github/actionlint.yaml
dsroberts: That's to address this linting error: https://github.com/firedrakeproject/firedrake/actions/runs/22885137596/job/66433010084. I will add a comment to the file to that effect.

firedrake/preconditioners/offload.py
dsroberts: I'm working my way through understanding how PETSc manages GPU devices, and it does appear that when I drop `super().initialize(pc)` I get:
```
[0] MatConvert() at /g/data/fp50/admin/gpu-testing/apps/petsc/3.24.4/src/mat/interface/matrix.c:4390
[0] Object is in wrong state
[0] Not for unassembled matrix
```
So the matrix does have to be assembled, but the previous PC might have performed assembly, so it wouldn't be necessary in that case. I think this might be an argument in favour of having device offload as an optional step in `AssembledPC`?

firedrake/preconditioners/offload.py
dsroberts: I genuinely don't know. On the surface it makes sense, but are there cases in which `A` or `P` can't be offloaded? Do we need do an `if not isinstance(A,ImplicitMatrix):` before attempting to offload?

scripts/firedrake-configure
dham: I think we shouldn't be silently adding software sources behind users' backs. This seems bad. Potentially we should just fail and tell users to go do it themselves.

firedrake/preconditioners/offload.py
pbrubeck: We certainly want PETSc options to give the user finer control over what to offload. 

If the original A and P point to the same Mat instance then it is reasonable to offload them both (as the same instance). This could be the default.

dsroberts:
@pbrubeck thanks for the feedback. I've been experimenting with implementations over the last couple of days and I think I'm close to the same conclusion. If offloading is integrated into `AssembledPC`, how do we handle it for cases that don't involve `AssembledPC`? I'm thinking a separate, distinct, copy-in copy-out 'PC' is the most sensible way to go for now. I did have some success with offloading in `AssembledPC` (really just assembling to an `aijcusparse` matrix) but I had issues with the copy-back for the `fieldsplit` preconditioner.

The reason I suggested a `try` option is that, once this is all merged in, we will only be installing GPU-enabled Firedrake on our HPC systems from now on. A GPU-enabled PETSc build should be fully functional on a CPU-only host. The `try` option was proposed as a way to make solver configurations portable between CPU-only nodes and GPU-nodes of heterogeneous systems. Without it, people would need to maintain two separate sets of solver settings, one for CPU and one for GPU. I guess your proposal would be more consistent with PETSc, maybe I'll just patch our Firedrake installs to handle that.

As for multiple different offload types being supported by the same device, both HIP and CUDA are vendor-specific languages, AMD devices only support HIP, Nvidia devices only support CUDA. That being said, I believe it is possible to build PETSc with support for multiple devices. Perhaps the `offload_mat_type` function should initialise a device and query what it gets back to determine what matrix type it should use.

scripts/firedrake-configure
dsroberts: `firedrake-configure` doesn't really 'do' anything here. Like the rest of the `firedrake-configure` steps It just prints a string to give to another command. A user still has to run `curl` and then `dpkg` to install it. Once this is in, I figured it would be added as an optional step to the installation instructions. If a user skips that step and then attempts to install system packages with `--gpu-arch cuda`, that would fail as `apt` would not be able to find any packages from the Nvidia repository.

dsroberts:
To update progress on this, I have a variant that passes all of the existing tests that subclasses `PCBase`. The issue with G-ADOPT is around the `fieldsplit` preconditioner, it looks like its trying to copy to or from an unallocated buffer. This PR is currently blocked by #4988, once that's in I'll make a new branch from `main` with only `OffloadPC` and the tests. I'll continue to work on `OffloadPC`, but won't commit anything further to the current branch.

dsroberts:
I've been working on a few avenues for this in the past couple of weeks, but they've all lead to dead-ends. The idea of persisting GPU buffers won't work, as PETSc has no way to retain automatically created device buffers. This is reasonable for developers working with the C library, if anyone wants to persist buffers, they can use their device implementation to create the buffer, and pass it into `VecCreate{Seq,MPI}CUDAWithArrays`. That means that Firedrake would need to use a 3rd party library to create the device buffer. This opens something of a can of worms, I've already got a basic implementation dict for the various `CUDA`/`HIP` vector functions needed (currently containing only a single CUDA function, but meant to be expanded), but adding a 3rd party GPU-specific library (e.g. CUDA python) to get native buffer handles would then require conditional imports and probably a lot of other similarly ugly stuff.

I've also been trying to verify GPU usage in the test cases, and that looks like it'll need `PetscLogNestedBegin()` exposed by petsc4py in order to grab machine-readable XML profiling output. I'm not exactly sure how that would interact with `pytest`, but I think it'd be fairly straightforward to use separate log stages for each test, run `PETSc.Log.view()` after the tests that need it and parse the XML to find GPU usage.

I still haven't figured out how to work with `matfree`. My (very limited) understanding is that since this preconditioner is doing effectively nothing, then `A` shouldn't need to be offloaded. If `P` is specified in `PCSetOperators`, then `A` won't be used. What this means for `matfree` I still don't know. Maybe in the case of a fieldsplit operator its sufficient to detect if we're already working with an offloaded matrix and do nothing if we are? I wonder how much its worth persisting given this work is a stop-gap before pyop3 is released. Or is it useful to encounter and solve these GPU-related issues before pyop3, given that not all of PETSc is GPU-ready?

tests/firedrake/offload/test_poisson_offloading_pc.py
dsroberts: Happy to try whatever you recommend here.

firedrake/preconditioners/offload.py
```diff
@@ -0,0 +1,114 @@
...
+        self.device_mat = device_matrix_type(pc.comm.rank == 0)
+        dm = opc.getDM()
+
+        pc = PETSc.PC().create(comm=opc.comm)
+        pc.setDM(dm)
+        pc.setOptionsPrefix(options_prefix)
+        if self.device_mat is not None:
+            with PETSc.Log.Event("Event: initialize offload"):
+                P_dev = P.convert(mat_type=self.device_mat)
+                A_dev = A.convert(mat_type=self.device_mat)
```
pbrubeck:
```suggestion
                P_dev = P.convert(mat_type=self.device_mat)
                A_dev = P_dev if A.handle == P.handle else A.convert(mat_type=self.device_mat)
```

firedrake/preconditioners/offload.py
```diff
@@ -0,0 +1,114 @@
...
+            self.vector_impls = _device_vector_impls[get_device_type()]
+            pc.setOperators(A_dev, P_dev)
+        else:
+            pc.setOperators(A, P)
+
+        # Simplest reconstruction we can manage
+        octx = dmhooks.get_appctx(dm)
+        self._ctx_ref = octx.reconstruct(
+            problem=None, mat_type=self.device_mat, pmat_type=self.device_mat
+        )
```
pbrubeck:
```suggestion
        octx = dmhooks.get_appctx(dm)
        self._ctx_ref = octx.reconstruct(mat_type=self.device_mat, pmat_type=self.device_mat)
```

firedrake/preconditioners/offload.py
```diff
@@ -0,0 +1,114 @@
...
+            context = P.getPythonContext()
+            # It only makes sense to preconditioner/invert a diagonal
+            # block in general.  That's all we're going to allow.
+            if not context.on_diag:
+                raise ValueError("Only makes sense to invert diagonal block")
+
+        prefix = pc.getOptionsPrefix() or ""
+        options_prefix = prefix + self._prefix
+
+        self.device_mat = device_matrix_type(pc.comm.rank == 0)
```
pbrubeck: I think for clarity we should make `warn` a kwarg, and pass `warn=True` here in this script. The communicator logic could be handled within utils.py (we might need to add a `comm` kwarg as well)

firedrake/preconditioners/offload.py
```diff
@@ -0,0 +1,114 @@
...
+        self.pc = pc
+
+        with dmhooks.add_hooks(dm, self, appctx=self._ctx_ref, save=False):
+            self.pc.setFromOptions()
+
+    def update(self, pc):
+        A, P = pc.getOperators()
+        A_dev, P_dev = self.pc.getOperators()
+        P.copy(P_dev)
+        A.copy(A_dev)
```
pbrubeck:
```suggestion
        P.copy(P_dev)
        if A_dev.handle != P_dev.handle:
            A.copy(A_dev)
```

firedrake/preconditioners/offload.py
```diff
@@ -0,0 +1,114 @@
...
+            with PETSc.Log.Event("Event: apply offload"):  #
+                dm = pc.getDM()
+                with dmhooks.add_hooks(dm, self, appctx=self._ctx_ref):
+                    with PETSc.Log.Event("Event: vectors offload"):
+                        # Create the to-be-offloaded vector
+                        y_dev = PETSc.Vec()
+                        # Use device implementation of 'createWithArrays' function
+                        getattr(y_dev, self.vector_impls["createWithArrays"])(
+                            y.array_r, None
+                        )
```
pbrubeck: Seems like this wants to become a class method.

firedrake/preconditioners/offload.py
```diff
@@ -0,0 +1,114 @@
...
+    """Offload PC from CPU to GPU and back.
+
+    Internally this makes a PETSc PC object that can be controlled by
+    options using the extra options prefix ``offload_``.
+    """
+
+    _prefix = "offload_"
+
+    def initialize(self, pc):
+        # Check if our PETSc installation is GPU enabled
```
pbrubeck: Is this a TODO comment?

firedrake/preconditioners/offload.py
```diff
@@ -0,0 +1,114 @@
...
+
+        A, P = pc.getOperators()
+
+        if pc.type != "python":
+            raise ValueError("Expecting PC type python")
+        opc = pc
+        if P.type == "python":
+            context = P.getPythonContext()
+            # It only makes sense to preconditioner/invert a diagonal
```
pbrubeck:
```suggestion
            # It only makes sense to precondition/invert a diagonal
```

firedrake/preconditioners/offload.py
```diff
@@ -0,0 +1,114 @@
...
+        A.copy(A_dev)
+
+    # Convert vectors to CUDA, solve and get solution on CPU back
+    def apply(self, pc, x, y):
+        if self.device_mat is None:
+            self.pc.apply(x, y)
+        else:
+            with PETSc.Log.Event("Event: apply offload"):  #
+                dm = pc.getDM()
+                with dmhooks.add_hooks(dm, self, appctx=self._ctx_ref):
```
pbrubeck: I think this `with` context should also be wrapping the case `if self.device_mat is None` above

firedrake/preconditioners/offload.py
```diff
@@ -0,0 +1,114 @@
...
+            self.pc.setFromOptions()
+
+    def update(self, pc):
+        A, P = pc.getOperators()
+        A_dev, P_dev = self.pc.getOperators()
+        P.copy(P_dev)
+        A.copy(A_dev)
+
+    # Convert vectors to CUDA, solve and get solution on CPU back
+    def apply(self, pc, x, y):
```
pbrubeck:
```suggestion
    def apply(self, pc, x, y, transpose=False):
        pc_apply = self.pc.applyTranspose if transpose else self.pc.apply 
```

firedrake/preconditioners/offload.py
```diff
@@ -0,0 +1,114 @@
...
+        if self.device_mat is None:
+            self.pc.apply(x, y)
```
pbrubeck:
```suggestion
            pc_apply(x, y)
```

firedrake/preconditioners/offload.py
```diff
@@ -0,0 +1,114 @@
...
+                            y.array_r, None
+                        )
+                        # Create the to-be-offloaded vector
+                        x_dev = PETSc.Vec()
+                        # Use device implementation of 'createWithArrays' function
+                        getattr(x_dev, self.vector_impls["createWithArrays"])(
+                            x.array_r, None
+                        )
+                    with PETSc.Log.Event("Event: solve"):
+                        self.pc.apply(x_dev, y_dev)
```
pbrubeck:
```suggestion
                        pc_apply(x_dev, y_dev)
```

firedrake/preconditioners/offload.py
```diff
@@ -0,0 +1,114 @@
...
+                    with PETSc.Log.Event("Event: vectors copy back"):
+                        # y is already designated as host storage for y_dev, so calling
+                        # getArray is sufficient to synchronise the vector on the device
+                        # with y on the host
+                        y_dev.getArray(True)
+
+    def applyTranspose(self, pc, X, Y):
+        raise NotImplementedError
```
pbrubeck:
```suggestion
    def applyTranspose(self, pc, x, y):
        self.apply(pc, x, y, transpose=True)
```

connorjward:
> The try option was proposed as a way to make solver configurations portable between CPU-only nodes and GPU-nodes of heterogeneous systems. Without it, people would need to maintain two separate sets of solver settings, one for CPU and one for GPU.

To some extent I think that this is recommended/expected. The hardware characteristics mean that having different solver configurations is encouraged.

> That means that Firedrake would need to use a 3rd party library to create the device buffer. 

I would be happy to suck up cupy as an optional dependency here - I fully expect that we will need it later on in our GPU journey. However, I would advocate for making that a later problem and a later PR.

> I've also been trying to verify GPU usage in the test cases

Why can't we do this by using OffloadPC and checking to see if it crashes/warns?

> My (very limited) understanding is that since this preconditioner is doing effectively nothing, then A shouldn't need to be offloaded.

I'm also unsure but I would think that since this is a preconditioner then it's only the P matrix that we care about and that gets propagated to the GPU-part of the solve.

> I still haven't figured out how to work with matfree.

This preconditioner is only going to work with already assembled matrices until we have one form assembly on GPUs.

dsroberts:
Thanks Connor. The more I work on this, the more I realise some of my early goals were infeasible. Portable solver configuration was one of them, and experiments with G-ADOPT have made me realise that isn't going to happen.

> > That means that Firedrake would need to use a 3rd party library to create the device buffer.
>
> I would be happy to suck up cupy as an optional dependency here - I fully expect that we will need it later on in our GPU journey. However, I would advocate for making that a later problem and a later PR.

The issue with that is there are also AMD GPUs. I'm trying to put this together in a way that I can start work on extending the offloading preconditioner to the AMD MI250X GPUs on [Setonix](https://pawsey.org.au/systems/setonix/) as soon as the CUDA variant is merged. It makes no sense to install cupy on that system. There are device agnostic languages (Vulkan, OpenCL, etc) but a) there is zero crossover in support between PETSc (supports OpenCL and Kokkos) and IREE (supports Vulkan) and b) even if a) was resolved performance is worse than vendor-specific languages. So maybe `pyproject.toml` gets `firedrake[cuda]` and `firedrake[hip]` optional dependency sets, but then you have the conditional import problem. I guess since we're querying PETSc for device support, we can import the appropriate python interface then? Definitely a future problem though. Device agnosticism is hard to develop initially, but much harder to retrofit so I think its worth considering all of this early on.



> >    I've also been trying to verify GPU usage in the test cases
>
> Why can't we do this by using OffloadPC and checking to see if it crashes/warns?

Its one thing to initialise a device, its another to ensure that GPU offloading has actually happened. Maybe I'm being overly cautious, but I don't think its safe to just assume that if we found a GPU, that it was used in the computation. For example, the SOR preconditioner has absolutely no GPU support, but is happy to take a `aijcusparse` matrix and ignore the device buffer in favour of doing all work on the CPU. I would want a test to make sure that doesn't happen in future on the cases we know offload properly today. The only way I know for sure to get that information is from `-log_view`.

> > I still haven't figured out how to work with matfree.
>
> This preconditioner is only going to work with already assembled matrices until we have one form assembly on GPUs.

For sure, that was more of a follow-up to @pbrubeck's comment from last month

> Is this because A cannot live in the GPU if it is matfree? Do we benefit much from A being matfree if P is not matfree?

There are subtleties here I'm not across yet. Especially around `fieldsplit` where there are solves nested in other operations. I don't know how offload would be managed in that case, or even if it can be done.

firedrake/preconditioners/offload.py
dsroberts: I'd like @connorjward's opinion on this since `device_matrix_type` as it stands was approved in #4988.

firedrake/preconditioners/offload.py
dsroberts: Leftover from before device initialisation was moved to `util.py`.

firedrake/preconditioners/offload.py
dsroberts: This typo was copied from `assembled.py`. I'll fix it there too.

firedrake/preconditioners/offload.py
dsroberts: Thank you for sorting out the `applyTranspose` case. I wasn't sure on the right way to do that.

pbrubeck:
> > This preconditioner is only going to work with already assembled matrices until we have one form assembly on GPUs.
> 
> For sure, that was more of a follow-up to @pbrubeck's comment from last month
> 
> > Is this because A cannot live in the GPU if it is matfree? Do we benefit much from A being matfree if P is not matfree?
> 
> There are subtleties here I'm not across yet. Especially around `fieldsplit` where there are solves nested in other operations. I don't know how offload would be managed in that case, or even if it can be done.

I commented on the distinction between A and P because I saw one of the use cases with `-offload_pc_type ksp`. There we are expecting A and P to be used in the GPU, it makes sense to do a single copy to device when these two are the same object (th most common scenario), but we should be prepared for the other case as well.

connorjward:
> So maybe pyproject.toml gets firedrake[cuda] and firedrake[hip] optional dependency sets, but then you have the conditional import problem.

Yeah this is the approach I was imagining. We already have to play conditional import games for things like JAX and PyTorch. I'm not overly concerned about this.

> Its one thing to initialise a device, its another to ensure that GPU offloading has actually happened. Maybe I'm being overly cautious, but I don't think its safe to just assume that if we found a GPU, that it was used in the computation. For example, the SOR preconditioner has absolutely no GPU support, but is happy to take a aijcusparse matrix and ignore the device buffer in favour of doing all work on the CPU. I would want a test to make sure that doesn't happen in future on the cases we know offload properly today. The only way I know for sure to get that information is from -log_view.

Fair enough. That is challenging. You could probably achieve this in a reasonable way if you added a step to the GPU CI very much like [this one in petsctools](https://github.com/firedrakeproject/petsctools/blob/main/.github/workflows/core.yml#L123):
```
python tests/firedrake/offload/test_poisson_offloading_pc.py -log_view | grep ...
```
Note that I'm running the test file as a script and therefore relying on the `if __name__ == "__main__"` block.

connorjward:
@pbrubeck and @dsroberts I want to make sure that we're not being too perfectionist here. This is brand new functionality and doesn't have to be absolutely perfect in all circumstances from the outset. Once this is merged we have 6 months until it is in release and as such we can change the API as much as we like until then.

dsroberts:
@connorjward you make a good point. I think what we have now is a good starting point and a clear idea of what work will come next. Since G-ADOPT is my priority, the next round of GPU work to come from me will be working out how to offload after a fieldsplit has been performed. HIP/ROCm to come after that. Also, please ensure @Olender is listed as a co-author in the merge commit, this work borrows a lot from #4166, and @Olender absolutely deserves credit for that in the next round of change logs.

tests/firedrake/offload/test_poisson_offloading_pc.py
```diff
@@ -0,0 +1,61 @@
+from firedrake import *
+import pytest
+
+
+# TODO: add something to check if cuda memory was really used(?)
```
connorjward:
```suggestion
```

firedrake/preconditioners/offload.py
```diff
@@ -0,0 +1,114 @@
...
+    Internally this makes a PETSc PC object that can be controlled by
+    options using the extra options prefix ``offload_``.
+    """
+
+    _prefix = "offload_"
+
+    def call_device_vec_impl(self, x, func_name: str, func_args=(), func_kwargs={}):
+        return getattr(x, _device_vector_impls[self.device_type][func_name])(
+            *func_args, **func_kwargs
+        )
```
connorjward:
```suggestion
    def call_device_vec_impl(self, x, func_name: str, *args, **kwargs):
        return getattr(x, _device_vector_impls[self.device_type][func_name])(
            *args, **kwargs
        )
```

firedrake/preconditioners/offload.py
connorjward: I think this touches on a bigger open question about how we should do logging and warning and whether we want to force it to be collective or not. I am happy to continue to pass the rank here for the moment.

That said I think making `warn` a kwarg here is a good idea. Just add `*, ` to the argument list.

firedrake/preconditioners/offload.py
```diff
@@ -0,0 +1,114 @@
...
+
+    # Convert vectors to CUDA, solve and get solution on CPU back
+    def apply(self, pc, x, y, transpose=False):
+        pc_apply = self.pc.applyTranspose if transpose else self.pc.apply
+        dm = pc.getDM()
+        if self.device_mat is None:
+            with dmhooks.add_hooks(dm, self, appctx=self._ctx_ref):
+                pc_apply(x, y)
+        else:
+            with PETSc.Log.Event("Event: apply offload"):  #
```
connorjward:
```suggestion
            with PETSc.Log.Event("Event: apply offload"):
```

firedrake/utils.py
```diff
@@ -26,6 +26,29 @@
...
+def get_device_type() -> str | None:
+    r"""Get PETSc device type
+
+    Attempt to initialise a GPU and return the type of GPU
+    identified by PETSc
+
+    Returns
+    -------
+    str | None
+        The PETSc device type
```
connorjward:
```suggestion
    r"""Get PETSc device type.

    Attempt to initialise a GPU and return the type of GPU
    identified by PETSc.

    Returns
    -------
    str | None
        The PETSc device type, or `None` if no device is found.

```

firedrake/utils.py
```diff
@@ -26,6 +26,29 @@
...
+        # Could not initialise device - not a failure condition as this could
+        # be a GPU-enabled PETSc installation running on a CPU-only host.
+        return None
+    dev_type = dev.getDeviceType()
+    dev.destroy()
+    return dev_type
+
+
 @cache
 def device_matrix_type(warn: bool = True) -> str | None:
```
connorjward:
```suggestion
def device_matrix_type(*, warn: bool = True) -> str | None:
```

tests/firedrake/offload/test_poisson_offloading_pc.py
connorjward: I assume this is resolved now that you have that extra CI step?

tests/firedrake/offload/test_poisson_offloading_pc.py
connorjward: Why not use the test case from [here](https://github.com/dsroberts/firedrake/blob/89606fee27d362653fc3c083c8af10ae98e53df4/tests/firedrake/regression/test_poisson_strong_bcs.py#L21)

connorjward (changes requested):
Sorry it took me a couple of days to get to this. I think once these very minor things are addressed I am happy to merge.

dsroberts:
@connorjward I know what you mean regarding logging, I don't think there is a nice, one-size-fits-all solution. This was an easy one as its a collectively consistent function that will never change throughout a job run, hence using `pc.comm.rank` and `@cache` as output control mechanisms.

Thanks for pointing out that test. I've attempted to add in the `nproc=1` tests with `-m skipslepc` to the GPU CI. All CPU-only cases should still work with a GPU-enabled installation and it isn't going to be feasible to put every GPU test into `firedrake-check`, so I think its a worthwhile thing to add. Its in a "DROP BEFORE MERGE" commit for now, if the test that's running as I write this comment fails, its probably best to drop and I'll come back to it in another PR as its coming up to the end of the working day for me.

ETA: I guess that should be `-m 'not skipslepc'`. On further reflection I do think that's out of scope for this PR. Feel free to purge that last commit and review what's there. I'll open another for merging offload testing framework into the existing pytest setup.

firedrake/preconditioners/offload.py
```diff
@@ -0,0 +1,114 @@
...
+
+
+_device_vector_impls = {
+    "CUDA": {
+        "createWithArrays": "createCUDAWithArrays",
+    }
+}
+
+
+class OffloadPC(PCBase):
```
pbrubeck, line 17: Should we include "GPU" in this class name? Right now this name isn't very clear for users

connorjward: I like 'offload' and think it's a fair description. In principle I think it could generalise beyond GPUs. For instance maybe you want to copy a portion of the solve back to the CPU, or in the future there might be non-GPU devices available.

connorjward (approved):
I am happy with this. Thank you so much for doing this! It's only a small PC but the surrounding work is very involved. This will be very useful for later GPU work.

firedrake/preconditioners/offload.py
connorjward: @pbrubeck if you are satisfied please resolve this, then we can merge it!

PR #4955, opened 2026-03-10
Fix Docker builds

connorjward:
Switch to using exactly the same approach for building petsc4py and slepc4py as we do in CI. Should be more robust.

### DIFF
--- a/docker/Dockerfile.vanilla
+++ b/docker/Dockerfile.vanilla
@@ -36,14 +36,14 @@
 #     Extra arguments to pass to PETSc's `configure` script
 #
 # Example usage for a specific firedrake release with complex scalars:
-#  $ docker build \
+#  $ docker build . \
 #     --file=Dockerfile.vanilla \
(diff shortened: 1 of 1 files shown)


docker/Dockerfile.vanilla
```diff
@@ -36,14 +36,14 @@
 #     Extra arguments to pass to PETSc's `configure` script
 #
 # Example usage for a specific firedrake release with complex scalars:
-#  $ docker build \
+#  $ docker build . \
```
JHopeCollins, line 39: what does the added `.` do? `man docker build` isn't clear on it (at least to me).

connorjward: I found it was necessary to run the command (so I could test that this actually runs). Without it I saw:
```
ERROR: docker: 'docker buildx build' requires 1 argument

Usage:  docker buildx build [OPTIONS] PATH | URL | -

Run 'docker buildx build --help' for more information
```

PR #4958, opened 2026-03-10
Update notebooks repo at every push to release

connorjward:
Instead of only when we make releases.

### DIFF
--- a/.github/workflows/push.yml
+++ b/.github/workflows/push.yml
@@ -43,7 +43,7 @@ jobs:
         uses: actions/checkout@v5
         with:
           path: firedrake
-          ref: ${{ inputs.branch }}
+          ref: ${{ github.ref_name }}
 
       - name: Checkout Colab notebooks
         uses: actions/checkout@v5


francesco-ballarin:
> Yes, a PR would be better. Even better, a PR that auto-merges if the CI in the other repo passes: it will definitely not pass on major releases, but it should easily pass on minor ones.

Slightly different proposal after https://github.com/firedrakeproject/firedrake/pull/5037 :
1. `.github/workflows/push.yml`, `colab` job: extract `{year}.{month}` from the version of the current firedrake (e.g., from 2025.10.2 discard the trailing .2 and convert it to 2025.10)
2. in  `.github/workflows/push.yml`, `colab` job: push to `update-{year}.{month}`
3. in the `notebooks` repo: add a new workflow (call it `auto-merge.yml` or something like that) that hardcodes the latest version, e.g. 2025.10. On any push to `update-{hardcoded-year}.{hardcoded-month}`, if CI passes automatically merge into `main`. 
4. At every firedrake release, after the version on Colab has been upgraded, change the `{hardcoded-year}.{hardcoded-month}` in `auto-merge.yml`. Ideally any push that changes `auto-merge.yml` should also trigger an auto-merge.

That's basically the same rationale of the PR above, but safe for new version releases and with very limited human intervention.

PR #4965, opened 2026-03-12
Refactor `EnsembleReducedFunctional`

JHopeCollins:
### Previous implementation

The `EnsembleReducedFunctional` implements a functional with many independent terms all depending on the same control, which are calculated in parallel over the ensemble:
```math
J(m) = \sum J_i(m)
```
This is three operations composed together:
1. Broadcast $m$ to all $J_i$
2. Transform all $J_i(m)$
3. Reduce $J=\sum J_i$

It also had limited support for using distributed controls, i.e. different $m_i$ for each $J_i$.
```math
J(m) = \sum J_i(m_i)
```
This is two operations composed together:
1. Transform all $J_i(m_i)$
2. Reduce $J=\sum J_i$

### Issues with previous implementation

In the distributed controls case the `Controls` were `Functions` on each spatial comm and were not collective over the global comm - this breaks the `ReducedFunctional` contract and meant that this version would fail the taylor test (e.g. the xfailed tests here:
https://github.com/firedrakeproject/firedrake/blob/a101140f648c50f7155114ade55660582618a1ea/tests/firedrake/adjoint/test_ensemble_reduced_functional.py#L99-L101

In the case with a single control $m$, the user had to implement the local part of the sum (i.e. manually sum the $J_i$ on the local rank themselves).

### What does this PR do?

This PR splits the `EnsembleReducedFunctional` into separate `ReducedFunctional` classes for the `Bcast`, `Transform`, and `Reduce`. `EnsembleReducedFunctional` is then re-implemented in terms of these operations.

These all use `EnsembleFunction` as either the control and/or functional as appropriate so we get collective behaviour and the Taylor tests for a distributed control pass (and we can use them in optimisers).

It also implements an additional `OverloadedType` called `EnsembleAdjVec` which is a distributed vector of `AdjFloat`. It is to `AdjFloat` what `EnsembleFunction` is to `Function`.

### What changes will users see?

To use distributed controls a user will need to create an `EnsembleFunctionSpace` to use an `EnsembleFunction` as the control. But in return you get the collective behaviour.

A significant API change for both individual and collective controls is that now the $J_i$ on each spatial comm are passed as a separate `ReducedFunctional`, rather than taping all the $J_i$ and the local reduction on one tape.

- This means that the user only has to implement $J_i$ rather than any of the reduction.
- It ensures that each $J_i$ has a different tape, so can be run in any order without polluting state for the derivative calculation.

### DIFF
--- a/demos/full_waveform_inversion/full_waveform_inversion.py.rst
+++ b/demos/full_waveform_inversion/full_waveform_inversion.py.rst
@@ -257,39 +257,51 @@ Next, the FWI problem is executed with the following steps:
     :align: center
 
 
-To have the step 4, we need first to tape the forward problem. That is done by calling::
+To have the step 4, we need first to tape the forward problem.
+That is done by calling :func:`~.pyadjoint.continue_annotation`.
+
+
+**Steps 2-3**: Solve the wave equation and compute the functional.
+We create a ``ReducedFunctional`` for each source, which for our
+case means one per ensemble member. Creating a ``ReducedFunctional``
+per component that we are parallelising over (i.e. per source) -
+rather than creating one per ensemble member - we can change
+the ensemble parallel partition with minimal changes to the code.::
 
     from firedrake.adjoint import *
-    continue_annotation()
-    get_working_tape().progress_bar = ProgressBar
-
-**Steps 2-3**: Solve the wave equation and compute the functional::
 
     f = Cofunction(V.dual())  # Wave equation forcing term.
     solver, u_np1, u_n, u_nm1 = wave_equation_solver(c_guess, f, dt, V)
     interpolate_receivers = interpolate(u_np1, V_r)
+
+    continue_annotation()
     J_val = 0.0
-    for step in range(total_steps):
-        f.assign(ricker_wavelet(step * dt, frequency_peak) * q_s)
-        solver.solve()
-        u_nm1.assign(u_n)
-        u_n.assign(u_np1)
-        guess_receiver = assemble(interpolate_receivers)
-        misfit = guess_receiver - true_data_receivers[step]
-        J_val += 0.5 * assemble(inner(misfit, misfit) * dx)
+    with set_working_tape() as tape:
+        for step in range(total_steps):
+            f.assign(ricker_wavelet(step * dt, frequency_peak) * q_s)
+            solver.solve()
+            u_nm1.assign(u_n)
+            u_n.assign(u_np1)
+            guess_receiver = assemble(interpolate_receivers)
+            misfit = guess_receiver - true_data_receivers[step]
+            J_val += 0.5 * assemble(inner(misfit, misfit) * dx)
 
-We now instantiate :class:`~.EnsembleReducedFunctional`::
+        control = Control(c_guess)
+        Jhat_local = ReducedFunctional(J_val, control, tape=tape)
+        tape.progress_bar = ProgressBar
+    pause_annotation()
 
-    J_hat = EnsembleReducedFunctional(J_val,
-                                      Control(c_guess, riesz_map="l2"),
-                                      my_ensemble)
+We now instantiate :class:`~.adjoint.ensemble_reduced_functional.EnsembleReducedFunctional`
+with the local ``ReducedFunctional`` for each source::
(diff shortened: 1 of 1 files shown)


demos/full_waveform_inversion/full_waveform_inversion.py.rst
```diff
@@ -257,39 +257,51 @@ Next, the FWI problem is executed with the following steps:
...
 
-**Steps 2-3**: Solve the wave equation and compute the functional::
+**Steps 2-3**: Solve the wave equation and compute the functional.
+We create a ``ReducedFunctional`` for each source, which for our
+case means one per ensemble member. Creating a ``ReducedFunctional``
+per component that we are parallelising over (i.e. per source) -
+rather than creating one per ensemble member - we can change
+the ensemble parallel partition with minimal changes to the code.::
+
+    from firedrake.adjoint import *
```
Ig-dolci, line 271: This is better 🙃

demos/full_waveform_inversion/full_waveform_inversion.py.rst
```diff
@@ -257,39 +257,51 @@ Next, the FWI problem is executed with the following steps:
...
-        J_val += 0.5 * assemble(inner(misfit, misfit) * dx)
 
-We now instantiate :class:`~.EnsembleReducedFunctional`::
-
-    J_hat = EnsembleReducedFunctional(J_val,
-                                      Control(c_guess, riesz_map="l2"),
-                                      my_ensemble)
+    continue_annotation()
+    J_val = 0.0
+    with set_working_tape() as tape:
```
Ig-dolci, line 279: Why is `with set_working_tape() as tape:` better here?

JHopeCollins: It makes it really clear which bit of the code is being recorded on that `tape`, and it makes sure that the `ReducedFunctional` for each $J_i$ has a different tape.

firedrake/adjoint/ensemble_adjvec.py
```diff
@@ -0,0 +1,121 @@
+from functools import cached_property
+from pyadjoint.overloaded_type import OverloadedType
+from pyadjoint.adjfloat import AdjFloat
+from firedrake.ensemble import Ensemble
+from firedrake.adjoint_utils.checkpointing import disk_checkpointing
+
+
+class EnsembleAdjVec(OverloadedType):
```
connorjward, line 8: Why not `EnsembleAdjFloat`?

firedrake/adjoint/ensemble_reduced_functional.py
```diff
@@ -1,249 +1,1106 @@
...
+                    " must have the same function space as the control.")
+        else:
+            raise ValueError(
+                f"Do not know how to handle a {type(control).__name__}"
+                f" control for {type(self).__name__}.")
+
+        # Adjoint action is a reduction so we just piggyback.
+        # Possibly don't do this if we're being created by the
+        # reduction rf to avoid infinite recursion.
+        if not _only_forward:
```
connorjward, line 337: Could this be avoided if you make `self._reduce` into a `cached_property` and hence lazily evaluated?

firedrake/adjoint/ensemble_reduced_functional.py
```diff
@@ -1,249 +1,1106 @@
...
+
+    @no_annotations
+    def hessian(self, m_dot, hessian_input=None, evaluate_tlm=True,
+                apply_riesz=False):
+        if evaluate_tlm:
+            self.tlm(m_dot)
+        return self.derivative(hessian_input, apply_riesz=apply_riesz)
+
+
+class EnsembleTransformReducedFunctional(AbstractReducedFunctional):
```
connorjward, line 515: This name is a bit confusing to me as it makes it seem very generic. It could be equivalent to `EnsembleReducedFunctional` or it could be an ABC for `EnsembleAllgatherReducedFunctional` etc.

Maybe something like `EnsembleNoReduceReducedFunctional` (lol)? `EnsemblePipelineReducedFunctional`? `EnsemblePassthroughReducedFunctional`?

PR #4966, opened 2026-03-12
Fix Dockerfile ENV setting (export does not work)

connorjward:
Fix something I accidentally broke in https://github.com/firedrakeproject/firedrake/commit/aff25591a5b312ecc382ae5f02dc7c621974c35f#diff-9154f250c6a8447b6c07a9bbb0648e3f66b5b464bb4e16778ab440a24514ea20

Thanks @ABaierReinio for reporting this.

I have built the containers for both x86 and ARM and verified that you can import netgen.

### DIFF
--- a/docker/Dockerfile.firedrake
+++ b/docker/Dockerfile.firedrake
@@ -36,7 +36,7 @@ RUN \
                 pip install ngsPETSc; \
         fi
 
-# Set some environment variables.
+# Set environment variables so that the netgen installation can be found
 # For x86 we have to set PYTHONPATH because of https://github.com/NGSolve/netgen/issues/213.
 ENV PATH=${ARM64:+/opt/ngsuite/netgen-install/bin:}$PATH
 ENV PYTHONPATH=${ARM64:+/opt/ngsuite/netgen-install/lib/python3.12/site-packages:}${X86:+/usr/local/lib/python3.12/site-packages:}$PYTHONPATH


docker/Dockerfile.firedrake
```diff
@@ -23,14 +22,20 @@ RUN \
...
 
+# Dockerfiles do not allow conditional ENV statements so we have to set
+# things for both the ARM and x86 cases.
+# ARM:
+ENV PATH=/opt/ngsuite/netgen-install/bin:$PATH
+ENV PYTHONPATH=/opt/ngsuite/netgen-install/lib/python3.12/site-packages:$PYTHONPATH
+# x86:
+# (see https://github.com/NGSolve/netgen/issues/213).
+ENV PYTHONPATH=/usr/local/lib/python3.12/site-packages:$PYTHONPATH
+
```
JHopeCollins: This is possibly confusing to have to always have both set. Is it possible to do something like this to emulate a conditional ENV statement? (And similar for `PYTHONPATH`).
```
RUN \
	if [ $platform -eq x86]; then echo /usr/...:$PATH > new_path.txt; \
	elif [ $platform -eq ARM]; then echo /opt/...:$PATH > new_path.txt \
	fi

ENV PATH=$(cat new_path.txt)

RUN rm new_path.txt
```

connorjward: Yeah that might work. I can try.

connorjward: I mean, it's really ugly, but I will try. The end result is cleaner.

connorjward: no dice:
```
Dockerfile.firedrake:38
--------------------
  36 |     # x86:
  37 |     # (see https://github.com/NGSolve/netgen/issues/213).
  38 | >>> ENV PYTHONPATH=$(cat pythonpath.txt)
  39 |
  40 |     RUN rm pythonpath.txt && python3 -c "import netgen"
--------------------
ERROR: failed to build: failed to solve: Syntax error - can't find = in "pythonpath.txt)". Must be of the form: name=value
```

JHopeCollins: That's annoying. Is it allowed to be a build-arg to the dockerfile? Could we pass it from the workflow seeing as we know there what platform we're on?
I know it's adding another argument but something about deliberately setting an incorrect path for the other architecture worries me, it seems a recipe for difficult to track down errors in the future.

JHopeCollins: Would getting that netgen issue fixed help? It would at least mean that we wouldn't need to change the path on x86.

@UZerbinati what's the likelihood of getting https://github.com/NGSolve/netgen/issues/213 seen to?

connorjward: You can do it with build args

connorjward: OK I have fixed this now

JHopeCollins: Thanks for that. A little extra bash but it keeps the end environment cleaner. We can at least remove the extra PYTHONPATH option if the netgen issue is fixed.

docker/Dockerfile.firedrake
```diff
@@ -23,14 +31,16 @@ RUN \
...
-                && export PATH=/opt/ngsuite/netgen-install/bin:$PATH \
-                && export PYTHONPATH=/opt/ngsuite/netgen-install/lib/python3.12/site-packages:$PYTHONPATH \
                 && pip install ngsPETSc --no-deps; \
         else \
-                export PYTHONPATH=/usr/local/lib/python3.12/site-packages:$PYTHONPATH \
-                && pip install ngsPETSc; \
+                pip install ngsPETSc; \
         fi
 
+# Set some environment variables.
```
JHopeCollins:
```suggestion
# Set environment variables so that the netgen installation can be found
```

JHopeCollins (dismissed):
Nitpicky suggestion for a comment, but happy other than that.

PR #4970, opened 2026-03-18
Real x extruded fixes

connorjward:
(currently incomplete) fixes needed for https://github.com/firedrakeproject/fiat/pull/237

### DIFF
--- a/.github/workflows/core.yml
+++ b/.github/workflows/core.yml
@@ -210,9 +210,6 @@ jobs:
             --extra-index-url https://download.pytorch.org/whl/cpu \
             "$(echo ./firedrake-repo/dist/firedrake-*.tar.gz)[ci]"
 
-          : # UNDO ME
-          pip install git+https://github.com/firedrakeproject/fiat.git@connorjward/handle-real -v --no-deps --ignore-installed
-
           firedrake-clean
           pip list
 


firedrake/functionspace.py
```diff
@@ -55,6 +55,9 @@ def make_scalar_element(mesh, family, degree, vfamily, vdegree, variant, quad_sc
     if isinstance(family, finat.ufl.FiniteElementBase):
         return family.reconstruct(cell=cell)
 
+    if family in {"R", "Real"} and degree is None:
+        degree = 0
```
connorjward: @pbrubeck under what circumstances do we allow `degree` to be `None`? I shoved this hack in but it is gross and likely should live in FIAT/FInAT/finat.ufl

pbrubeck: Degree being None means the lowest order for which this element still makes sense

pbrubeck: A common mistake is to not pass the degree when creating the R space. I think we should enable degree=None for DG in FIAT to mean degree=0, and similarly for other spaces.

connorjward: OK. I'll open a precursor PR then that does that.

pbrubeck: This should live in fiat/discontinuous_lagrange.py. Just make `degree` a kwarg in the constructor.

firedrake/functionspace.py
```diff
@@ -31,7 +31,8 @@ def make_scalar_element(mesh, family, degree, vfamily, vdegree, variant, quad_sc
     family :
         The finite element family.
     degree :
-        The degree of the finite element.
+        The degree of the finite element. If unspecified this will default
+        to the lowest degree available for the given family.
```
pbrubeck, line 35: This is what we would ideally want to say, but it involves more work in FIAT

connorjward: This is arguably OK to state I think. It's no worse than the previous situation and makes it clear that if `degree` ever ends up being `None` then that's a simple bug that we should be fixing.

pbrubeck: We should create an issue with this as a TODO for every FIAT element

connorjward: This is arguably OK to state I think. It's no worse than the previous situation and makes it clear that if `degree` ever ends up being `None` then that's a simple bug that we should be fixing.

connorjward: > We should create an issue with this as a TODO for every FIAT element

In https://github.com/firedrakeproject/fiat/pull/238 I make it so that we raise a `NotImplementedError`, so at least the failure is helpful. An issue would also help.

.github/workflows/core.yml
```diff
@@ -210,6 +210,9 @@ jobs:
             --extra-index-url https://download.pytorch.org/whl/cpu \
             "$(echo ./firedrake-repo/dist/firedrake-*.tar.gz)[ci]"
 
+          : # UNDO ME
+          pip install git+https://github.com/firedrakeproject/fiat.git@connorjward/handle-real -v --no-deps --ignore-installed
+
```
```suggestion
```

PR #4972, opened 2026-03-18
Add deadlock safe function for calling fallible pieces of code on rank 0

connorjward:
Generalises the solution from https://github.com/firedrakeproject/firedrake/pull/4913

### DIFF
--- a/pyop2/mpi.py
+++ b/pyop2/mpi.py
@@ -566,7 +566,7 @@ def finalize_safe_debug():
     return debug
 
 
-def safe_comm_branch(comm: MPI.Comm, func: Callable[[], Any], *, root: int = 0) -> Any:
+def safe_noncollective(comm: MPI.Comm, func: Callable[[], Any], *, root: int) -> Any:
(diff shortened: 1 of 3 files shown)


pyop2/mpi.py
```diff
@@ -565,6 +566,44 @@ def finalize_safe_debug():
     return debug
 
 
+def safe_comm_branch(comm: MPI.Comm, func: Callable[[], Any], *, root: int = 0) -> Any:
```
connorjward: This isn't a great name. Happy to hear suggestions.

JHopeCollins: `collective_try_except`?
Regardless of the name, I would make `root` a required argument to make it clear that the code is only actually run on one rank.

connorjward: `safe_noncollective`? This is wrapping a noncollective operation in a collective wrapper.
I like your idea about `root`.

JHopeCollins: Sure, that's fine. I don't think there's an obvious clearest name so we could bikeshed this all day. The docstring explains what it does well anyway.

PR #4981, opened 2026-03-20
Use `firedrake-rtree` for spatial indexing

leo-collins:
# Description

Use [firedrake-rtree](https://github.com/firedrakeproject/firedrake-rtree) for spatial indexing.

- This is an order of magnitude faster than libspatialindex for querying points.
- We can now build a one dimensional spatial index, allowing us to remove a whole load of special-cased code.
- We can now access the nodes of the tree and manually traverse it. This will be used to build a distributed rtree which will allow us to use sparse communication operations to build the VertexOnlyMesh, instead of the expensive allgathers we use currently.


firedrake/locate.c
```diff
@@ -31,156 +31,81 @@ int locate_cell(struct Function *f,
...
-                if (current_ref_cell_dist_l1 <= 0.0) {
-                    /* Found cell! */
-                    cell = ids[i];
-                    memcpy(found_ref_coords, temp_ref_coords, sizeof(struct ReferenceCoords));
-                    found_ref_cell_dist_l1[0] = current_ref_cell_dist_l1;
+    size_t *ids = NULL;
+    size_t nids = 0;
+    err = rtree_locate_all_at_point((const struct RTreeH *)f->sidx, x, &ids, &nids);
+    if (err != Success) {
+        fputs("ERROR: RTree_LocateAllAtPoint failed in rstar-capi!\n", stderr);
```
leo-collins:
```suggestion
        fputs("ERROR: rtree_locate_all_at_point failed in rtree-capi!\n", stderr);
```

setup.py
connorjward, line 208:
```suggestion
        **(mpi_ + petsc_ + numpy_ + libsupermesh_ + spatialindex_)
```

connorjward: It doesn't look like you're actually linking to this anywhere, which is why this probably isn't fixing anything.

leo-collins: The issue was in libsupermesh: https://github.com/firedrakeproject/libsupermesh/pull/16

connorjward: That makes perfect sense. We needed to be linking to libspatialindex somewhere. We should remove this library then since it's not getting used.

firedrake/cython/rtree.pyx
```diff
@@ -0,0 +1,84 @@
...
+    RTree
+        An RTree object containing the Rtree.
+    """    
+    cdef:
+        RTreeH* rtree
+        size_t n
+        size_t dim
+        RTreeError err
+
+    if coords_min.shape[0] != coords_max.shape[0] or coords_min.shape[1] != coords_max.shape[1]:
```
connorjward, line 82:
```suggestion
    if coords_min.shape != coords_max.shape:
```

firedrake/cython/rtreeinc.pxi
connorjward, line 1: Seems needlessly complicated to have this extra file.

firedrake/mesh.py
connorjward: Does this need renaming?

connorjward (changes requested):
Just tiny things. I think this is great and 99% ready.

firedrake/cython/rtree.pyx
leo-collins: This doesn't work for some reason. `coords_min.shape != coords_max.shape` evaluates to true.

pyproject.toml
```diff
@@ -25,8 +25,10 @@ dependencies = [
   # TODO RELEASE
   "firedrake-fiat @ git+https://github.com/firedrakeproject/fiat.git@main",
   "h5py>3.12.1",
+  "firedrake-rtree",
   "immutabledict",
-  "libsupermesh",
+  # TODO RELEASE
+  "libsupermesh @ git+https://github.com/firedrakeproject/libsupermesh.git@main",
```
```suggestion
  "libsupermesh>=2026.0",
```

pyproject.toml
```diff
@@ -154,7 +156,9 @@ docker = [  # Used in firedrake-vanilla container
 [build-system]
 requires = [
   "Cython>=3.0",
-  "libsupermesh",
+  "firedrake-rtree",
+  # TODO RELEASE
+  "libsupermesh @ git+https://github.com/firedrakeproject/libsupermesh.git@main",
```
```suggestion
  "libsupermesh>=2026.0",
```

requirements-build.txt
```diff
@@ -1,6 +1,8 @@
 # Core build dependencies (adapted from pyproject.toml)
 Cython>=3.0
-libsupermesh
+# TODO RELEASE
+libsupermesh @ git+https://github.com/firedrakeproject/libsupermesh.git@main
+firedrake-rtree
```
connorjward:
```suggestion
firedrake-rtree
libsupermesh>=2026.0
```

connorjward (changes requested):
Otherwise this looks good to go.

What's the status of firedrake-rtree? At a minimum you need to add some more people as owners of the PyPI package.

francesco-ballarin:
Do I understand correctly that once this is merged then neither `libspatialindex` (from https://github.com/libspatialindex/libspatialindex.git) nor `rtree` (from https://github.com/Toblerity/rtree.git) will be required dependencies anymore?

(need to be aware so that I can clean up my installation scripts accordingly)

leo-collins:
As I understand it we are getting `libspatialindex` via the `rtree` python package, and this is still needed for `libsupermesh`. Is this correct @connorjward ?

connorjward:
> As I understand it we are getting `libspatialindex` via the `rtree` python package, and this is still needed for `libsupermesh`. Is this correct @connorjward ?

Yep that's right.

@leo-collins does this mean that we can drop rtree from requirements-build.txt and pyproject.toml? Last question I promise before merging...

@francesco-ballarin if so rtree is only used as a finder for libspatialindex in libsupermesh. If you install libspatialindex separately you should probably be able to drop rtree.

francesco-ballarin:
Can
https://github.com/firedrakeproject/libsupermesh/blob/9f740544f6180da3a043feecb6992e246195aa81/pyproject.toml#L3
and
https://github.com/firedrakeproject/libsupermesh/blob/9f740544f6180da3a043feecb6992e246195aa81/pyproject.toml#L19
now be removed as well?

My understanding is that from now on `rtree` will actually *not* be a dependency at all of `libsupermesh`. I do understand that it can be used as fallback to find `libspatialindex`, but having it listed as a build and runtime dependency has a different meaning, requiring `rtree` to *always* be present.

connorjward:
I'm not sure about this. Users get libspatialindex via rtree. It's the only way we can confidently have it on their system. And if it's a build dependency then it also has to be a runtime dependency, otherwise the library wouldn't be there.

francesco-ballarin:
There seems to be something funny happening in the interaction with `rust` with `--no-build-isolation`.

On my Colab build script, I install `maturin` (the build backend of `firedrake-rtree`) and `puccinialin` (fallback inside `maturin` when the system does not have rust installed, which installs it automatically).

On my `debian:testing` system, I install `maturin` but *not* `puccinialin`, because my system provides a `cargo` command.

On the Colab installation, inside the directory `firedrake/cython`, I get
```
# ldd rtree.cpython-312-x86_64-linux-gnu.so
	linux-vdso.so.1 (0x00007f7a73b99000)
	/usr/local/lib/python3.12/dist-packages/firedrake_rtree/firedrake_rtree.cpython-312-x86_64-linux-gnu.so => not found
	libc.so.6 => /lib/x86_64-linux-gnu/libc.so.6 (0x00007f7a73948000)
	/lib64/ld-linux-x86-64.so.2 (0x00007f7a73b9b000)
```
and note that the second library is correctly called `firedrake_rtree.cpython-...`.

On my local system instead I get
```
$ ldd rtree.cpython-313-x86_64-linux-gnu.so
	linux-vdso.so.1 (0x00007f1545884000)
	librtree_capi.so => not found
	libc.so.6 => /usr/lib/x86_64-linux-gnu/libc.so.6 (0x00007f1545643000)
	/lib64/ld-linux-x86-64.so.2 (0x00007f1545886000)
```
and note that the second library has the wrong name `librtree_capi.so`. In `firedrake_rtree` I can actually see a `firedrake_rtree.cpython-313-x86_64-linux-gnu.so`, rather than a `librtree_capi.so`.

I can fix the build and the `ldd` output on my local system by temporarily uninstalling `cargo` and temporarily installing `puccinialin`. Maybe the rust installation I have on my local system is not compete enough to generate the correct ouptut, but still I do not get any error telling me so in the build phase, only at runtime when trying to import `firedrake` I get an error saying that `librtree_capi.so` does not exist.

Not sure if a case like this is worth adding to the CI job where you have `--no-build-isolation`.

connorjward:
That's a tricky bug. I am unfamiliar with this but a quick look suggests that `firedrake_rtree.XXX` is coming from [here](https://github.com/firedrakeproject/firedrake-rtree/blob/main/pyproject.toml#L29) and `rtree_capi.XXX` from [here](https://github.com/firedrakeproject/firedrake-rtree/blob/main/rtree-capi/Cargo.toml#L18). The latter case with `rtree_capi` is somehow broken wrt the Python extension. I don't know the fix.

> Not sure if a case like this is worth adding to the CI job where you have --no-build-isolation.

In our CI we try to model what the user does as much as possible so if possible I would rather avoid doing this.

PR #4616, opened 2025-10-02
Improved traceback from PETSc errors

stefanozampini:
# Description

There is no need to maintain obsolete `CHKERR` and `SETERR` implementations in Firedrake.

PETSc management of tracebacks from Cython/Python was improved one year ago in https://gitlab.com/petsc/petsc/-/merge_requests/7938, and now we can get valuable information from catching errors.

Needs `CHKERRMPI` exposed in https://gitlab.com/petsc/petsc/-/merge_requests/8758

For example, adding mixing INSERT and ADD values will trigger an error in the PETSc C code that will be forwarded appropriately (through `PETSC_ERR_PYTHON`) to the Python runtime (the patch is for the code in this PR)

```
diff --git a/pyop2/sparsity.pyx b/pyop2/sparsity.pyx
index 5914bacc8..f4dcd4d8f 100644
--- a/pyop2/sparsity.pyx
+++ b/pyop2/sparsity.pyx
@@ -50,6 +50,7 @@ cdef extern from "petsc.h":
         PETSC_TRUE, PETSC_FALSE
     ctypedef enum PetscInsertMode "InsertMode":
         PETSC_INSERT_VALUES "INSERT_VALUES"
+        PETSC_ADD_VALUES "ADD_VALUES"
     ctypedef enum PetscErrorCode:
         PETSC_SUCCESS

@@ -200,6 +201,7 @@ def fill_with_zeros(PETSc.Mat mat not None, dims, maps, iteration_regions, set_d
         for i in range(nrow // rdim):
             if i < ncol  // cdim:
                 CHKERR(MatSetValuesBlockedLocal(mat.mat, 1, &i, 1, &i, diag_values, PETSC_INSERT_VALUES))
+                CHKERR(MatSetValuesBlockedLocal(mat.mat, 1, &i, 1, &i, diag_values, PETSC_ADD_VALUES))
         CHKERR(PetscFree(diag_values))
     if len(maps) == 0:
         return
```
With the current firedrake code, we get
```
pyop2/sparsity.pyx:163: in pyop2.sparsity.build_sparsity
    fill_with_zeros(preallocator, (1, 1),
[...]

PR #4993, opened 2026-03-26
Refactor PCPatch code to not use PyOP2

connorjward:
This PR:
* Refactors the code in `patch.py` to not use the PyOP2 code generator and just generates C code directly instead.
* Makes the code much more understandable (in my opinion). There is a lot of code generation/callback/struct magic going on that I have packaged more neatly together than what was already there.

The reason I have done this is because I was struggling to reproduce the level of hackery done in this file for pyop3. PCPatch requires a specific interface to the wrapper kernel which is easy enough to enforce in PyOP2 but much more difficult in pyop3 where the compiler is much more advanced and hence the generated code has a less predictable call signature.

I think it would be possible to recover this behaviour, and use pyop3 for codegen, but I would like to use this as a temporary measure because:
* No features were harmed in this transition, everything works as it did before
* It makes pyop3 that much closer to merging


firedrake/cython/patchimpl.pyx
```diff
@@ -2,45 +2,32 @@ from libc.stdint cimport uintptr_t
...
-                                                           <PetscPCPatchComputeFunction><uintptr_t>function,
-                                                           <void *><uintptr_t>ctx))
-        elif exterior_facets:
-            CHKERR(PCPatchSetComputeFunctionExteriorFacets((<PETSc.PC?>patch).pc,
-                                                           <PetscPCPatchComputeFunction><uintptr_t>function,
-                                                           <void *><uintptr_t>ctx))
-        else:
-            CHKERR(PCPatchSetComputeFunction((<PETSc.PC?>patch).pc,
-                                             <PetscPCPatchComputeFunction><uintptr_t>function,
-                                             <void *><uintptr_t>ctx))
```
connorjward, line 24: These functions were never called

firedrake/preconditioners/patch.py
```diff
@@ -1,607 +1,590 @@
...
+  const PetscScalar *state           = NULL;
+  const PetscInt    *whichPoints     = NULL;
+  const PetscInt    *activeDofsArray = dofArray;
+  UserCtx           *ctx             = (UserCtx *)ctx_;
+  PetscInt           npoints;
+  PetscInt          *filtpoints      = NULL;
+  PetscInt          *filtdofs        = NULL;
+
+  PetscFunctionBeginUser;
+  PetscCall(ISGetSize(points, &npoints));
```
connorjward, line 387: I use the modern `PetscCall` calling convention, replacing `ierr = ...`

firedrake/cython/patchimpl.pyx
```diff
@@ -2,45 +2,32 @@ from libc.stdint cimport uintptr_t
...
-            CHKERR(PCPatchSetComputeOperatorExteriorFacets((<PETSc.PC?>patch).pc,
-                                                           <PetscPCPatchComputeOperator><uintptr_t>function,
-                                                           <void *><uintptr_t>ctx))
-        else:
-            CHKERR(PCPatchSetComputeOperator((<PETSc.PC?>patch).pc,
-                                             <PetscPCPatchComputeOperator><uintptr_t>function,
-                                             <void *><uintptr_t>ctx))
+
+def pcpatch_set_compute_operator(patch, function, ctx):
+    CHKERR(PCPatchSetComputeOperator((<PETSc.PC?>patch).pc,
```
pbrubeck, line 7: PetscCall

firedrake/cython/patchimpl.pyx
```diff
@@ -2,45 +2,32 @@ from libc.stdint cimport uintptr_t
...
+                                     <PetscPCPatchComputeOperator><uintptr_t>function,
+                                     <void *><uintptr_t>ctx))
+
+
+def pcpatch_set_compute_operator_interior_facets(patch, function, ctx):
+    CHKERR(PCPatchSetComputeOperatorInteriorFacets((<PETSc.PC?>patch).pc,
```
pbrubeck, line 13: PetscCall

firedrake/cython/patchimpl.pyx
```diff
@@ -2,45 +2,32 @@ from libc.stdint cimport uintptr_t
...
+                                                   <PetscPCPatchComputeOperator><uintptr_t>function,
+                                                   <void *><uintptr_t>ctx))
+
+
+def pcpatch_set_compute_operator_exterior_facets(patch, function, ctx):
+    CHKERR(PCPatchSetComputeOperatorExteriorFacets((<PETSc.PC?>patch).pc,
+                                                   <PetscPCPatchComputeOperator><uintptr_t>function,
```
pbrubeck, line 20: PetscCall

firedrake/cython/patchimpl.pyx
```diff
@@ -2,45 +2,32 @@ from libc.stdint cimport uintptr_t
...
+
+
+def snespatch_set_compute_operator(patch, function, ctx):
+    CHKERR(SNESPatchSetComputeOperator((<PETSc.SNES?>patch).snes,
+                                        <PetscPCPatchComputeOperator><uintptr_t>function,
+                                        <void *><uintptr_t>ctx))
+
+
+def snespatch_set_compute_function(patch, function, ctx):
+    CHKERR(SNESPatchSetComputeFunction((<PETSc.SNES?>patch).snes,
```
pbrubeck, line 31: Same here

pbrubeck (approved):
LGTM

firedrake/cython/patchimpl.pyx
connorjward: I'm pretty sure that `CHKERR` is the right approach in Cython. See https://github.com/firedrakeproject/firedrake/pull/4616 for example.

PR #4994, opened 2026-03-26
Support physical_normals on manifolds

pbrubeck:
# Description
This is needed for zany elements on manifolds.


tsfc/kernel_interface/common.py
```diff
@@ -73,10 +73,7 @@ def cell_orientation(self, domain, restriction):
         f = {None: 0, '+': 0, '-': 1}[restriction]
         co_int = self._cell_orientations[domain][f]
         return gem.Conditional(gem.Comparison("==", co_int, gem.Literal(1)),
-                               gem.Literal(-1),
-                               gem.Conditional(gem.Comparison("==", co_int, gem.Zero()),
-                                               gem.Literal(1),
-                                               gem.Literal(numpy.nan)))
```
pbrubeck, line 79: This code generates a gem expression that takes `self._cell_orientations` on a given side of a facet and returns -1 if co == 1, +1 if co == 0, and NaN otherwise. 

I am removing the NaN branch because it generates a loopy warning at compile time:
```
/home/pbrubeck/firedrake-venv/lib/python3.10/site-packages/pytools/persistent_dict.py:273: NanKeyWarning: Encountered a NaN while hashing. Since NaNs compare unequal to themselves, the resulting key can not be retrieved from a PersistentDict and will lead to a collision error on retrieval.
  warn("Encountered a NaN while hashing. Since NaNs compare unequal "
/home/pbrubeck/firedrake-venv/lib/python3.10/site-packages/loopy/target/c/codegen/expression.py:472: UserWarning: Encountered 'bare' floating point NaN value. Since NaN != NaN, this leads to problems with cache retrieval. Consider using `pymbolic.primitives.NaN` instead of `math.nan`. The generated code will be equivalent with the added benefit of sound pickling/unpickling of kernel objects.
  warn("Encountered 'bare' floating point NaN value. Since NaN != NaN,"
```

pbrubeck: Maybe the right thing is to 

> Consider using `pymbolic.primitives.NaN` instead of `math.nan`

@connorjward what do you think?

connorjward: I would suggest trying it out and seeing if it works. Seems reasonable to try.

pbrubeck: `pymbolic.primitives.NaN` is a type, and we cannot just create a `numpy.ndarray` with `dtype=object`. Ideally we would intercept NaN in the gem -> loopy translation.

https://github.com/firedrakeproject/firedrake/blob/43a5050285d39bc12f66517f0bb756a817d5e4f4/tsfc/loopy.py#L239-L240

connorjward: Add a `gem.NaN` type?

pbrubeck: I don't mind this warning, I think the point of this PR is to add support for physical_normals on manifolds.

connorjward (dismissed):

pbrubeck:
Merging to make CI green

PR #5004, opened 2026-03-31
Submesh: support tuple subdomain_id

pbrubeck:
Enhances the valid values for `subdomain_id` for `Submesh`, so it matches closely the behavior of the DirichletBC's subdomain kwarg. This spares users from relabeling the mesh themselves.
- `Submesh(mesh, ..., subdomain_id=[1, 2])` takes the union of subdomains 1 and 2 (this works for codim-0 and codim-1)

- `Submesh(mesh, ..., subdomain_id="on_boundary")` as a shorthand for `Submesh(mesh, tdim-1, subdomain_id=1, label_name="exterior_facets")` 

Fixes https://github.com/firedrakeproject/firedrake/issues/5001

### DIFF
--- a/firedrake/cython/dmcommon.pyx
+++ b/firedrake/cython/dmcommon.pyx
@@ -3998,14 +3998,15 @@ def submesh_create(PETSc.DM dm,
         PETSc.IS points, subpoints
         PETSc.DMLabel label, temp_label
         char *temp_label_name = <char *>"firedrake_submesh_temp_label"
-        PetscInt pStart, pEnd, p, i, stratum_size, label_value
+        PetscInt pStart, pEnd, p, i, stratum_size = 0, label_value = 1
         const PetscInt *stratum_indices = NULL
+
     # Cast subdomain_id into an iterable
     if isinstance(subdomain_id, str) or not isinstance(subdomain_id, Sequence):
         subdomain_id = (subdomain_id,)
     # Take the union of the all the label values
     label = dm.getLabel(label_name)
-    points = PETSc.IS().createGeneral([], comm=dm.comm)
+    points = PETSc.IS()
     for sub in subdomain_id:
         if isinstance(sub, Integral):
             subpoints = label.getStratumIS(sub)
@@ -4013,12 +4014,15 @@ def submesh_create(PETSc.DM dm,
             subpoints = dm.getStratumIS("exterior_facets", 1)
         else:
             raise ValueError(f"Submesh construction got invalid subdomain_id {sub}.")
-        points = points.union(subpoints)
+        if points:
+            points = points.union(subpoints)
+        else:
+            points = subpoints
     # Create temp_label that contains no lower-dimensional points.
     dm.createLabel(temp_label_name)
     temp_label = dm.getLabel(temp_label_name)
-    label_value = 1
-    CHKERR(ISGetSize(points.iset, &stratum_size))
+    if points:
+        CHKERR(ISGetSize(points.iset, &stratum_size))
     if stratum_size > 0:
         CHKERR(ISGetIndices(points.iset, &stratum_indices))
         CHKERR(DMPlexGetDepthStratum(dm.dm, subdim, &pStart, &pEnd))
--- a/firedrake/cython/dmcommon.pyx
+++ b/firedrake/cython/dmcommon.pyx
@@ -7,6 +7,7 @@ import math
 import cython
 import numpy as np
 import firedrake
+from collections.abc import Sequence
 from firedrake.petsc import PETSc
 from mpi4py import MPI
 from firedrake.utils import IntType, ScalarType
@@ -2203,11 +2204,11 @@ def _get_expanded_dm_dg_coords(dm: PETSc.DM, ndofs: np.ndarray):
 
 def _get_periodicity(dm: PETSc.DM) -> tuple[tuple[bool, bool], ...]:
     """Return mesh periodicity information.
-    
+
     This function returns a 2-tuple of bools per dimension where the first entry indicates
     whether the mesh is periodic in that dimension, and the second indicates whether the
     mesh is single-cell periodic in that dimension.
-    
+
     """
     cdef:
         const PetscReal *maxCell, *L
@@ -3971,7 +3972,7 @@ def create_halo_exchange_sf(PETSc.DM dm):
 def submesh_create(PETSc.DM dm,
                    PetscInt subdim,
                    label_name,
-                   PetscInt label_value,
+                   subdomain_id,
                    PetscBool ignore_label_halo,
                    comm=None):
     """Create submesh.
@@ -3984,8 +3985,8 @@ def submesh_create(PETSc.DM dm,
         Topological dimension of the submesh
     label_name : str
         Name of the label
-    label_value : int
-        Value in the label
+    subdomain_id : int | Sequence
+        Values in the label
     ignore_label_halo : bool
         If labeled points in the halo are ignored.
     comm : PETSc.Comm | None
@@ -3995,9 +3996,41 @@ def submesh_create(PETSc.DM dm,
     cdef:
         PETSc.DMLabel label, temp_label
         char *temp_label_name = <char *>"firedrake_submesh_temp_label"
-        PetscInt pStart, pEnd, p, i, stratum_size
+        PetscInt pStart, pEnd, p, i, stratum_size, label_value
         PETSc.PetscIS stratum_is = NULL
         const PetscInt *stratum_indices = NULL
+
+    # Parse non-integer subdomain_id into a single value
+    if isinstance(subdomain_id, str):
+        if subdomain_id == "on_boundary":
+            label_name = "exterior_facets"
+            label_value = 1
+        else:
+            raise ValueError(f"Submesh construction got invalid subdomain_id {subdomain_id}.")
+    elif isinstance(subdomain_id, Sequence):
+        label = dm.getLabel(label_name)
+
+        def get_label_points(sub):
+            if sub == "on_boundary":
+                return dm.getStratumIS("exterior_facets", 1)
+            elif isinstance(sub, str):
+                raise ValueError(f"Submesh construction got invalid subdomain_id {sub}.")
+            else:
+                return label.getStratumIS(sub)
+
+        # Take the union of the labels in the list
+        iset = PETSc.IS().createGeneral([], comm=dm.comm)
+        for sub in subdomain_id:
+            iset = iset.union(get_label_points(sub))
+
+        # Create a temporary label
+        label_name = "firedrake_composite_subdomain_label"
+        dm.createLabel(label_name)
+        label_value = 1 + dm.getLabelIdIS(label_name).getSize()
+        label = dm.getLabel(label_name)
+        label.setStratumIS(label_value, iset)
+    else:
+        label_value = subdomain_id
 
     label = dm.getLabel(label_name)
     # Create temp_label that contains no lower-dimensional points.
--- a/firedrake/cython/dmcommon.pyx
+++ b/firedrake/cython/dmcommon.pyx
@@ -7,13 +7,14 @@ import math
 import cython
 import numpy as np
 import firedrake
-from collections.abc import Sequence
 from firedrake.petsc import PETSc
 from mpi4py import MPI
 from firedrake.utils import IntType, ScalarType
 from libc.string cimport memset
 from libc.stdlib cimport qsort
 from finat.element_factory import as_fiat_cell
+from numbers import Integral
+from collections.abc import Sequence
 
 cimport numpy as np
 cimport mpi4py.MPI as MPI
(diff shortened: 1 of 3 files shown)


tests/firedrake/submesh/test_submesh_interface.py
```diff
@@ -0,0 +1,93 @@
...
+    mesh = UnitSquareMesh(4, 4)
+    x, y = SpatialCoordinate(mesh)
+    M = FunctionSpace(mesh, "DG", 0)
+    m1 = Function(M).interpolate(conditional(lt(x, 0.5), 1, 0))
+    m2 = Function(M).interpolate(conditional(lt(x, 0.5), 0, 1))
+    mesh.mark_entities(m1, 111)
+    mesh.mark_entities(m2, 222)
+
+    subdomain_id = [(111, 222)]
+    submesh1 = Submesh(mesh, mesh.topological_dimension - 1, subdomain_id=subdomain_id, label_name="Cell Sets")
```
pbrubeck: Here we define a surface subdomain as the intersection of two volume subdomains. The user interface is not as neat as we would like because we need to pass `label_name="Cell Sets`" to let `Submesh` know that the IDs that we are intersecting refer to the volume IDs.

firedrake/mesh.py
```diff
@@ -4876,15 +4878,43 @@ def Submesh(mesh, subdim=None, subdomain_id=None, label_name=None, name=None, ig
...
+
+    if isinstance(subdomain_id, Sequence):
+        # Create a temporary DMLabel with the union of the labels in the list
+        icomm = comm or mesh.comm
+        iset = PETSc.IS().createGeneral([], comm=icomm)
+        for sub in subdomain_id:
+            if isinstance(sub, Sequence):
+                # Take the intersection of the (closure of the) labels from nested lists
+                ises = [plex.getStratumIS(label_name, subi) for subi in sub]
+                closure = [[plex.getTransitiveClosure(p)[0] for p in i.indices] for i in ises]
```
pbrubeck: Closure is only needed for codimension 1. We could add an if-statement.

firedrake/mesh.py
```diff
@@ -4876,15 +4878,43 @@ def Submesh(mesh, subdim=None, subdomain_id=None, label_name=None, name=None, ig
             label_name = dmcommon.CELL_SETS_LABEL
         elif subdim == dim - 1:
             label_name = dmcommon.FACE_SETS_LABEL
+
+    # Parse non-integer subdomain_id
+    if subdomain_id == "on_boundary":
+        subdomain_id = tuple(mesh.exterior_facets.unique_markers)
```
pbrubeck: We need to check that the codimension is 1 (and maybe also set it as the default if `subdomain_id="on_boundary")`.

firedrake/mesh.py
```diff
@@ -4806,10 +4806,12 @@ def Submesh(mesh, subdim=None, subdomain_id=None, label_name=None, name=None, ig
     subdim : int | None
         Topological dimension of the submesh.
         Defaults to ``mesh.topological_dimension``.
-    subdomain_id : int | None
+    subdomain_id : int | Sequence | None
         Subdomain ID representing the submesh.
-        If `None` the submesh will cover the entire domain.
-        This is useful to obtain a codim-1 submesh over all facets or
+        If multiple subdomain IDs are provided, their union is taken.
+        If nested lists of subdomain IDs are provided, their intersection is taken.
```
connorjward: This really needs an example. Perhaps in a separate section like https://numpydoc.readthedocs.io/en/latest/format.html#examples

firedrake/mesh.py
```diff
@@ -4876,15 +4878,43 @@ def Submesh(mesh, subdim=None, subdomain_id=None, label_name=None, name=None, ig
             label_name = dmcommon.CELL_SETS_LABEL
         elif subdim == dim - 1:
             label_name = dmcommon.FACE_SETS_LABEL
+
+    # Parse non-integer subdomain_id
+    if subdomain_id == "on_boundary":
+        subdomain_id = tuple(mesh.exterior_facets.unique_markers)
+
+    if isinstance(subdomain_id, Sequence):
```
connorjward: We should barf if `subdomain_id` is a string here because a string is a sequence

firedrake/mesh.py
connorjward: I don't understand why closure is needed at all here

tests/firedrake/submesh/test_submesh_interface.py
```diff
@@ -0,0 +1,93 @@
+import pytest
+import numpy as np
+from firedrake import *
+
+
+def test_submesh_subdomain_id_tuple():
```
connorjward:
```suggestion
def test_submesh_subdomain_id_union():
```

tests/firedrake/submesh/test_submesh_interface.py
```diff
@@ -0,0 +1,93 @@
...
+    expected = assemble(m3*dx)
+    assert abs(assemble(1*dx(domain=submesh1)) - expected) < 1E-12
+
+    mesh.mark_entities(m3, 333)
+    submesh2 = Submesh(mesh, mesh.topological_dimension, 333)
+    assert submesh2.cell_set.size == submesh1.cell_set.size
+    assert np.allclose(submesh2.coordinates.dat.data, submesh1.coordinates.dat.data)
+
+
+def test_submesh_subdomain_id_nested_tuple():
```
connorjward:
```suggestion
def test_submesh_subdomain_id_intersection():
```

tests/firedrake/submesh/test_submesh_interface.py
```diff
@@ -0,0 +1,93 @@
...
+    assert abs(assemble(1*dx(domain=submesh1)) - expected) < 1E-12
+
+    mesh.mark_entities(m3, 333)
+    submesh2 = Submesh(mesh, mesh.topological_dimension, 333)
+    assert submesh2.cell_set.size == submesh1.cell_set.size
+    assert np.allclose(submesh2.coordinates.dat.data, submesh1.coordinates.dat.data)
+
+
+@pytest.mark.parametrize("subdomain_id", ["on_boundary", (1, 3, 6)])
+def test_submesh_facet_subdomain_id_tuple(subdomain_id):
```
connorjward: bad name

tests/firedrake/submesh/test_submesh_interface.py
```diff
@@ -0,0 +1,93 @@
...
+    facet_function = Function(V)
+    DirichletBC(V, 1, subdomain_id).apply(facet_function)
+    facet_value = 999
+    rmesh = RelabeledMesh(mesh, [facet_function], [facet_value])
+    submesh2 = Submesh(rmesh, mesh.topological_dimension - 1, facet_value)
+    assert submesh2.cell_set.size == submesh1.cell_set.size
+    assert np.allclose(submesh2.coordinates.dat.data, submesh1.coordinates.dat.data)
+
+
+def test_submesh_facet_subdomain_id_nested_tuple():
```
connorjward: bad name

tests/firedrake/submesh/test_submesh_interface.py
connorjward: Not sure that this can be avoided.

connorjward (changes requested):
In https://github.com/firedrakeproject/firedrake/issues/5001#issuecomment-4169105097 I've figured out a much nicer way to do this. I'm not sure if we should be adding this very unintuitive API if a better solution exists with only a tiny amount more work.

firedrake/mesh.py
pbrubeck: We might want to intersect two cell labels in order to get a facet label. The cell labels only include cells, in order to get the facets, we need the closure.

connorjward: So really what we want is https://petsc.org/release/manualpages/DMPlex/DMPlexLabelComplete/. We could shove a binding for it into dmcommon for now?

firedrake/mesh.py
connorjward: I think the sensible default for a codim 1 is still the skeleton mesh

pbrubeck: I meant a default subdim for `Submesh(mesh, subdomain_id="on_boubdary")`

connorjward: Oh I see. Yeah that seems reasonable.

firedrake/cython/dmcommon.pyx
```diff
@@ -4325,3 +4325,41 @@ def get_dm_cell_types(PETSc.DM dm):
...
+    dm : PETSc.DM
+        The DMPlex.
+    label_name : str
+        The name of the label
+    label_values : Sequence[int]
+        The values of the subdomain label to intersect
+
+    Returns
+    -------
+    tuple
```
```suggestion
    PETSc.IS
```

firedrake/mesh.py
```diff
@@ -4876,15 +4880,47 @@ def Submesh(mesh, subdim=None, subdomain_id=None, label_name=None, name=None, ig
...
+            raise ValueError(f"Submesh construction got invalid subdomain_id {subdomain_id}.")
+
+    if isinstance(subdomain_id, Sequence):
+        # Create a temporary DMLabel with the union of the labels in the list
+        icomm = comm or mesh.comm
+        iset = PETSc.IS().createGeneral([], comm=icomm)
+        for sub in subdomain_id:
+            try:
+                sub, = sub
+            except ValueError:
```
```suggestion
            except (TypeError, ValueError):
```

connorjward:
> Even after parsing, we need to implement a mechanism to traverse the expression tree. I don't agree that the amount of work is tiny.
>
> I suggest we keep this PR to support the same format used by DirichletBC. If you want to you can open a PR where you implement the logic to process the parsed tree for both DirichletBC and Submesh, in a backwards compatible way.

I disagree that we should be adding a known-to-be-confusing interface to a new a feature just because it's what's already been done. The logic really is very simple. You need:

```py
def myrecursion(expr):
  if isinstance(expr, str | int):
    return dm.getLabel(label_name, expr)
  
  lexpr, op, rexpr = expr
  liset = myrecursion(lexpr)
  riset = myrecursion(rexpr)
  if op == "&":
    return ISIntersect(liset, riset)
  else:
    assert op == "|"
    return ISUnion(liset, riset)
```

pbrubeck:
> I disagree that we should be adding a known-to-be-confusing interface to a new a feature just because it's what's already been done.

Intuitively, many users would try to pass a list of subdomains, as they already do for DirichletBC. I think we should support this

connorjward:
> > I disagree that we should be adding a known-to-be-confusing interface to a new a feature just because it's what's already been done.
> 
> Intuitively, many users would try to pass a list of subdomains, as they already do for DirichletBC. I think we should support this

I am very happy with the API for a union. It's the intersection behaviour that I don't like.

connorjward:
I am not the overlord however and this is subjective. If you can get agreement from other members of the team I will happily concede.

pbrubeck:
@connorjward I removed the subdomain intersection functionality. Since this is adding substantial improvements to the documentation, and fixes a very common use case, could this PR be merged into release?

connorjward:
> @connorjward I removed the subdomain intersection functionality. Since this is adding substantial improvements to the documentation, and fixes a very common use case, could this PR be merged into release?

Yeah I think that that's fine to do. It's not any sort of breaking change and since submesh is a brand new feature I think we're allowed to move faster on making changes to it.

firedrake/mesh.py
```diff
@@ -4972,6 +5026,10 @@ def Submesh(mesh, subdim=None, subdomain_id=None, label_name=None, name=None, ig
     )
     # Tag the relabeled mesh with the original distribution parameters
     submesh._distribution_parameters = mesh._distribution_parameters
+    # Store the construction parameters in case we need to reconstruct this Submesh
+    submesh._submesh_label_name = label_name
+    submesh._submesh_subdomain_id = subdomain_id
+    submesh._submesh_ignore_halo = ignore_halo
```
```suggestion
```

firedrake/cython/dmcommon.pyx
```diff
@@ -3995,10 +3996,42 @@ def submesh_create(PETSc.DM dm,
...
+                raise ValueError(f"Submesh construction got invalid subdomain_id {sub}.")
+            else:
+                return label.getStratumIS(sub)
+
+        # Take the union of the labels in the list
+        iset = PETSc.IS().createGeneral([], comm=dm.comm)
+        for sub in subdomain_id:
+            iset = iset.union(get_label_points(sub))
+
+        # Create a temporary label
```
connorjward: We should destroy this at the end of the method

firedrake/mesh.py
```diff
@@ -4927,13 +4928,66 @@ def Submesh(mesh, subdim=None, subdomain_id=None, label_name=None, name=None, ig
     ridges to be contained in the quad mesh are shared by at most two
     facets to make the quad mesh orientation algorithm work.
 
+    Examples
+    --------
+    >>> mesh = UnitSquareMesh(4, 4)
```
connorjward, line 4936: Ideally we should be testing code snippets so they don't bitrot. That's a bit complicated here. Can you open an issue and assign me?

connorjward (changes requested):
Looking good

firedrake/cython/dmcommon.pyx
pbrubeck: Should we? I was originally doing this and then noticed that this function creates another temporary label and does not destroy it.

connorjward:
```py
    # Destroy temp_label.
    dm.removeLabel(temp_label_name)
    subdm.removeLabel(temp_label_name)
```
Does it not? Is there another I'm missing?

pbrubeck: Oh right I missed it. I guess I can reuse this label with a different value to take the union

connorjward (approved):
Approved subject to the last unresolved comment

PR #5005, opened 2026-03-31
Keep a consistent labeling of boundaries for 2D periodic meshes

connorjward:
This PR changes the labels of 2D periodic meshes so that they always resemble
```
    4
 +-----+
 |     |
1|     |2
 |     |
 +-----+
    3
```
regardless of periodicity (some labels can be empty). This contrasts with the current solution where the boundary labels are
```
    2
 +-----+
 |     |
 |     |
 |     |
 +-----+
    1
```
or
```
 +-----+
 |     |
1|     |2
 |     |
 +-----+
```
depending on the direction of periodicity.

This is useful to do because:
* It's more intuitive, the old solution only existed because of how we generated periodic meshes via higher-dimensional manifold meshes.
* It lets us discard some tricky hacks

This is unfortunately and unavoidably an immediately breaking change but we at least warn users if they try to use an empty boundary set.

### DIFF
--- a/tests/firedrake/regression/test_periodic_2d.py
+++ b/tests/firedrake/regression/test_periodic_2d.py
@@ -38,10 +38,7 @@ def test_periodic_helmholtz(direction, cell_options):
 
     f = Function(V).assign((244.0*pi*pi/225.0 + 1.0)*u_exact)
 
-    if direction == "x":
-        bcs = DirichletBC(V, Constant(0), (3, 4))
-    elif direction == "y":
-        bcs = DirichletBC(V, Constant(0), (1, 2))
+    bcs = DirichletBC(V, 0, "on_boundary")
     elif direction == "both":
         bcs = []
     u = TrialFunction(V)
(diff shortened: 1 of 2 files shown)


tests/firedrake/regression/test_periodic_2d.py
```diff
@@ -38,7 +38,9 @@ def test_periodic_helmholtz(direction, cell_options):
 
     f = Function(V).assign((244.0*pi*pi/225.0 + 1.0)*u_exact)
 
-    if direction in {"x", "y"}:
+    if direction == "x":
+        bcs = DirichletBC(V, Constant(0), (3, 4))
+    elif direction == "y":
         bcs = DirichletBC(V, Constant(0), (1, 2))
```
pbrubeck:
```suggestion
    bcs = DirichletBC(V, 0, "on_boundary")
```

tests/firedrake/regression/test_periodic_2d.py
```diff
@@ -38,8 +38,7 @@ def test_periodic_helmholtz(direction, cell_options):
 
     f = Function(V).assign((244.0*pi*pi/225.0 + 1.0)*u_exact)
 
-    if direction in {"x", "y"}:
-        bcs = DirichletBC(V, Constant(0), (1, 2))
+    bcs = DirichletBC(V, 0, "on_boundary")
     elif direction == "both":
         bcs = []
```
connorjward:
```suggestion
```

dham (approved):

PR #5022, opened 2026-04-13
delay creation of multifunction

dham:
Due to the limitations of the UFL typecode system, early instantiation of ufl multifunctions impedes the extension of the UFL language (for example by Irksome to include a time derivative). This PR delays the instantiation of a multifunction in Firedrake.

### DIFF
--- a/firedrake/formmanipulation.py
+++ b/firedrake/formmanipulation.py
@@ -63,7 +63,6 @@ class ExtractSubBlock(MultiFunction):
             type(self)._index_inliner = self.IndexInliner()
         return self._index_inliner
 
-
     def _subspace_argument(self, a):
         return type(a)(subspace(a.function_space(), self.blocks[a.number()]),
                        a.number(), part=a.part())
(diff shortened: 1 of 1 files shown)


firedrake/formmanipulation.py
```diff
@@ -49,7 +49,20 @@ def indexed(self, o, child, multiindex):
...
+        This is a property so that the IndexInliner is not created on import.
+        This is a workaround for issues in Irksome caused by the UFL typecode
+        system.
+        """
+        try:
+            return self._index_inliner
+        except AttributeError:
+            type(self)._index_inliner = self.IndexInliner()
+        return self._index_inliner
+
```
connorjward, line 65:
```suggestion
```

connorjward (approved):

pbrubeck:
I think `IndexInliner` is no longer needed and can be deleted, UFL already does this simplification at object creation.

connorjward:
> I think `IndexInliner` is no longer needed and can be deleted, UFL already does this simplification at objection creation.

We are trying to get this in before release. That can probably be a later fix.

PR #5034, opened 2026-04-16
Fix Netgen permutation

pbrubeck:
# Description
Attempting to make `find_permutation` more robust by comparing reference coordinates and using `numpy.argmin` as opposed to taking an arbitary tolerance, which is now deprecated.

The correct fix is to use a PETSc Section: https://github.com/firedrakeproject/firedrake/pull/5277


UZerbinati:
@KarsKnook might be what you are looking for, can you take a look ? :)

connorjward (changes requested):
Isn't this undone in https://github.com/firedrakeproject/firedrake/pull/5277?

Not sure what this brings to the table. Shouldn't we just finish and merge #5277?

pbrubeck:
> Isn't this undone in #5277?
> 
> Not sure what this brings to the table. Shouldn't we just finish and merge #5277?

#5277 targets #5215 but this problem with the tolerances is a real bug (some high-order netgen meshes end up with self-intersecting cells), so we should ideally be targeting release. Are you okay with a cython rework of high-order netgen meshes going into release?

connorjward:
> Are you okay with a cython rework of high-order netgen meshes going into release?

Of course not.

If this fixes a real bug in `release` then I don't mind this going in to it, which can be distinct from the changes to `main`. We should have a test in that case.

pbrubeck:
> If this fixes a real bug in `release` then I don't mind this going in to it, which can be distinct from the changes to `main`. We should have a test in that case.

Done

firedrake/netgen.py
```diff
@@ -95,7 +94,19 @@ def find_permutation(points_a: np.ndarray, points_b: np.ndarray,
...
+    dim = points_a.shape[-1]
+    vids = list(range(dim+1))
+
+    bs = points_a[:, vids[0], :]
+    As = points_a[:, vids[1:], :]
+    As -= bs[:, None, :]
+    Ainvs = np.linalg.inv(As)
+
+    ref_points_a = np.matmul(points_a - bs[:, None, :], Ainvs)
+    ref_points_b = np.matmul(points_b - bs[:, None, :], Ainvs)
```
connorjward: This is all quite confusing. Could a comment/explanatory example be added?

connorjward (changes requested):

firedrake/netgen.py
```diff
@@ -95,7 +94,19 @@ def find_permutation(points_a: np.ndarray, points_b: np.ndarray,
...
+    # Match reference points instead of physical points to ensure scale invariance
+    dim = points_a.shape[-1]
+    vids = list(range(dim+1))
+
+    bs = points_a[:, vids[0], :]
+    As = points_a[:, vids[1:], :]
+    As -= bs[:, None, :]
+    Ainvs = np.linalg.inv(As)
+
+    ref_points_a = np.matmul(points_a - bs[:, None, :], Ainvs)
```
pbrubeck:
```suggestion
    # x_phys = A * x_ref + b <==> x_ref = inv(A) * (x_phys - b)
    ref_points_a = np.matmul(points_a - bs[:, None, :], Ainvs)
```

firedrake/netgen.py
pbrubeck: This is just very basic math. What do you want me to clarify?

firedrake/netgen.py
```diff
@@ -95,7 +94,19 @@ def find_permutation(points_a: np.ndarray, points_b: np.ndarray,
     if points_a.shape != points_b.shape:
         raise ValueError("`points_a` and `points_b` must have the same shape.")
 
-    p = [np.where(cdist(a, b).T < tol)[1] for a, b in zip(points_a, points_b)]
+    # Match reference points instead of physical points to ensure scale invariance
+    dim = points_a.shape[-1]
+    vids = list(range(dim+1))
+
```
pbrubeck:
```suggestion
    # Infer the affine mapping (A, b) from the image of the vertices
```

firedrake/netgen.py
connorjward: The maths may be basic but the code isn't

firedrake/netgen.py
```diff
@@ -95,7 +94,19 @@ def find_permutation(points_a: np.ndarray, points_b: np.ndarray,
...
+    dim = points_a.shape[-1]
+    vids = list(range(dim+1))
+    # Infer the affine mapping (A, b) from the image of the vertices
+    bs = points_a[:, vids[0], :]
+    As = points_a[:, vids[1:], :]
+    As -= bs[:, None, :]
+    Ainvs = np.linalg.inv(As)
+    # x_phys = A * x_ref + b <==> x_ref = inv(A) * (x_phys - b)
+    ref_points_a = np.matmul(points_a - bs[:, None, :], Ainvs)
+    ref_points_b = np.matmul(points_b - bs[:, None, :], Ainvs)
```
pbrubeck:
```suggestion
    bs = points_a[:, vids[:1], :]
    As = points_a[:, vids[1:], :]
    As -= bs
    Ainvs = np.linalg.inv(As)
    # x_phys = A * x_ref + b <==> x_ref = inv(A) * (x_phys - b)
    ref_points_a = np.matmul(points_a - bs, Ainvs)
    ref_points_b = np.matmul(points_b - bs, Ainvs)
```

firedrake/netgen.py
```diff
@@ -95,7 +94,19 @@ def find_permutation(points_a: np.ndarray, points_b: np.ndarray,
     if points_a.shape != points_b.shape:
         raise ValueError("`points_a` and `points_b` must have the same shape.")
 
-    p = [np.where(cdist(a, b).T < tol)[1] for a, b in zip(points_a, points_b)]
+    # Match reference points instead of physical points to ensure scale invariance
+    dim = points_a.shape[-1]
+    vids = list(range(dim+1))
+    # Infer the affine mapping (A, b) from the image of the vertices
```
pbrubeck:
```suggestion
    # Infer the affine mapping (A, b) from the image of the vertices (first dim+1 dofs)
```

firedrake/netgen.py
```diff
@@ -95,7 +94,19 @@ def find_permutation(points_a: np.ndarray, points_b: np.ndarray,
...
+    bs = points_a[:, vids[:1], :]
+    As = points_a[:, vids[1:], :]
+    As -= bs
+    Ainvs = np.linalg.inv(As)
+    # x_phys = A * x_ref + b <==> x_ref = inv(A) * (x_phys - b)
```
pbrubeck, line 105:
```suggestion
    # x_phys = A * x_ref + b <==> x_ref = inv(A) * (x_phys - b)
    # Multiply inv(A) from the right, since the data is row-major
```

firedrake/netgen.py
pbrubeck: I made the code a bit simpler, and added more comments. Happy now?

firedrake/netgen.py
```diff
@@ -95,7 +94,20 @@ def find_permutation(points_a: np.ndarray, points_b: np.ndarray,
     if points_a.shape != points_b.shape:
         raise ValueError("`points_a` and `points_b` must have the same shape.")
 
-    p = [np.where(cdist(a, b).T < tol)[1] for a, b in zip(points_a, points_b)]
+    # Match reference points instead of physical points to ensure scale invariance
+    dim = points_a.shape[-1]
+    vids = list(range(dim+1))
+    # Infer the affine mapping (A, b) from the image of the vertices (first dim+1 dofs)
+    bs = points_a[:, vids[:1], :]
+    As = points_a[:, vids[1:], :]
```
connorjward, line 102:
```suggestion
    bs = points_a[:, :1, :]
    As = points_a[:, 1:, :]
```
does this work? Not sure what `vids` is giving you

connorjward (commented):
Yes this much better. I think that this is not just me being picky. This is radically clearer for anyone else coming to this code for this first time.

firedrake/netgen.py
pbrubeck: No, it won't work. `vids` is of length dim+1, and `points_a.shape[1]` is strictly larger than that (we never run this code for CG1 coordinates). I'd just leave `vids` as it is for the sake of clarity.

connorjward: Not sure it is clearer but OK.

connorjward (approved):
I still don't understand the maths, but that's a me issue. This is probably fine.

UZerbinati:
I'm very very happy with this, it's nice to see that you find a better way to construct the mapping.
I was trying to do something similar here: https://github.com/firedrakeproject/firedrake/tree/uz/fix_permuation

pbrubeck:
> I'm very very happy with this, it's nice to see that you find a better way to construct the mapping.

This is just a patch. The correct approach is #5277

PR #5037, opened 2026-04-17
Temporarily disable auto-updating Colab notebooks

connorjward:
Until Colab updates to the latest version of Firedrake.

### DIFF
--- a/.github/workflows/push.yml
+++ b/.github/workflows/push.yml
@@ -28,7 +28,9 @@ jobs:
 
   colab:
     name: Update Colab notebooks (release only)
-    if: github.ref_name == 'release'
+    # TODO: Undo this when Colab switches to the latest version of Firedrake
+    # if: github.ref_name == 'release'
+    if: false
     runs-on: ubuntu-latest
     permissions:
       contents: write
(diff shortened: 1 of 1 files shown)


connorjward:
@francesco-ballarin this is nothing like the elegant solution you suggested in https://github.com/firedrakeproject/firedrake/pull/4958#issuecomment-4053557991 but I wanted something quick and dirty.

francesco-ballarin:
The issue I see is that once I do upgrade the version on Colab, you will need a further push somewhere in this repository to trigger the update. Maybe instead of disabling just push to notebooks@another-branch, so that we can manually merge another-branch into main in the notebooks repo after I do the upgrade?

connorjward:
> The issue I see is that once I do upgrade the version on Colab, you will need a further push somewhere in this repository to trigger the update. Maybe instead of disabling just push to notebooks@another-branch, so that we can manually merge another-branch into main in the notebooks repo after I do the upgrade?

Merging a commit reverting this undo-ing would accomplish that.

I appreciate that your suggestions may be better, but they are complex to implement and I'm not sure if they're worth the extra effort. I can bring this for discussion to the next team meeting.

francesco-ballarin:
I am rebuilding now (PETSc included), assuming everything goes smoothly it should be done in a couple of hours

francesco-ballarin:
`2026.4.0` is out on Colab.

PR #5039, opened 2026-04-17
New plotting methods for VertexOnlyMesh

achanbour:
I have extended Firedrake's plotting routines to support the `VertexOnlyMesh`. In summary:

- I added a new `pointplot` method that calls `matplotlib.scatter` for visualising the points of a `VertexOnlyMesh`,
- I extended the existing `quiver` method to plot vector fields on a `VertexOnlyMesh`.

I've also added unit tests to make sure these changes work as expected. Unless anyone has anything to add or comment, I will edit Firedrake's documentation to include a demo of these new functionalities.


firedrake/pyplot/mpl.py
```diff
@@ -83,6 +83,54 @@ def _get_collection_types(gdim, tdim):
     raise ValueError("Geometric dimension must be either 2 or 3!")
 
 
+@PETSc.Log.EventDecorator()
+def pointplot(vom, axes=None, **kwargs):
+    r"""Plot a VertexOnlyMesh as a scatter plot.
+
+    :arg vom: either `~.MeshGeometry` object that is a VertexOnlyMesh or a `~.Function` defined on it
```
connorjward: Please use numpydoc and type hinting. https://github.com/firedrakeproject/firedrake/wiki/Firedrake-Coding-Guide

connorjward: This doesn't make clear what the difference is between passing the mesh or a function.

firedrake/pyplot/mpl.py
```diff
@@ -83,6 +83,54 @@ def _get_collection_types(gdim, tdim):
...
+    :arg axes: matplotlib :class:`Axes <matplotlib.axes.Axes>` object on which to plot the mesh
+    :arg kwargs: keyword arguments passed to :meth:`scatter <matplotlib.axes.Axes.scatter>`
+    :return: a :class:`matplotlib.collections.PathCollection` artist.
+    """
+    if not (isinstance(vom, Function)
+            and isinstance(vom.function_space().mesh().topology, VertexOnlyMeshTopology)) \
+            and not isinstance(vom.topology, VertexOnlyMeshTopology):
```
connorjward: Shouldn't this be an `or`?

firedrake/pyplot/mpl.py
```diff
@@ -83,6 +83,54 @@ def _get_collection_types(gdim, tdim):
...
+        raise TypeError("Expected a VertexOnlyMesh or a Function defined on it.")
```
connorjward:
```suggestion
        raise TypeError("Expected a VertexOnlyMesh or a Function defined on one.")
```

firedrake/pyplot/mpl.py
```diff
@@ -83,6 +83,54 @@ def _get_collection_types(gdim, tdim):
...
+
+    gdim = vom.geometric_dimension
+    coords = toreal(vom.coordinates.dat.data_ro_with_halos, "real")
+
+    if axes is None:
+        fig = plt.figure()
+        if gdim == 3:
+            axes = fig.add_subplot(111, projection="3d")
+        else:
+            axes = fig.add_subplot(111)
```
connorjward, line 133: Does this assume 2D?

firedrake/pyplot/mpl.py
```diff
@@ -83,6 +83,54 @@ def _get_collection_types(gdim, tdim):
...
 
+@PETSc.Log.EventDecorator()
+def pointplot(vom, axes=None, **kwargs):
+    r"""Plot a VertexOnlyMesh as a scatter plot.
+
+    :arg vom: either `~.MeshGeometry` object that is a VertexOnlyMesh or a `~.Function` defined on it
+    :arg axes: matplotlib :class:`Axes <matplotlib.axes.Axes>` object on which to plot the mesh
+    :arg kwargs: keyword arguments passed to :meth:`scatter <matplotlib.axes.Axes.scatter>`
+    :return: a :class:`matplotlib.collections.PathCollection` artist.
+    """
```
connorjward, line 104: I know it's not done in the other functions, but we only expect this to be run in serial. I think we should check `vom.comm.size == 1` or raise a custom `SerialExecutionOnlyException` (to be created).

connorjward (changes requested):
Nice!

I'd add something small to the manual [here](https://www.firedrakeproject.org/visualisation.html#plotting-with-matplotlib).

firedrake/pyplot/mpl.py
achanbour: That's just basically `not (case 1 or case 2)` which is equivalent to `(not case 1) and (not case 2)`. I agree that it's cleaner to write it with an or:

```python
is_vom = isinstance(vom, MeshGeometry) and isinstance(vom.topology, VertexOnlyMeshTopology)
is_function_on_vom = isinstance(vom, Function) and isinstance(vom.function_space().mesh().topology, VertexOnlyMeshTopology)

if not (is_vom or is_function_on_vom):
    raise TypeError("Expected a VertexOnlyMesh or a Function defined on one.")
```

firedrake/pyplot/mpl.py
achanbour: Noted. I changed the docstrings of the functions I amended/added.

achanbour: Added this to the updated docstring.

firedrake/pyplot/mpl.py
achanbour: Yes similar to `triplot`. Not sure what the default behaviour should be for 1D. I know that matplotlib scatter requires the points to be in 2D so passing in a 1D array would raise an error. We can either say that this is not supported for 1D or just zero out one of the coordinates so that points are plotted on one axis.

danshapero:
This will be really useful, thanks! My one suggestion is that you call it `scatter` instead of `pointplot`. The other plotting routines have the same names as their counterparts in matplotlib so that people could guess the names of the function they wanted without having to consult the documentation.

achanbour:
> This will be really useful, thanks! My one suggestion is that you call it `scatter` instead of `pointplot`. The other plotting routines have the same names as their counterparts in matplotlib so that people could guess the names of the function they wanted without having to consult the documentation.

This makes sense, thank you! I've now modified it to `scatter`.

firedrake/pyplot/mpl.py
achanbour: I am happy to include the check here and everywhere else while I'm at it. Should `SerialExecutionOnlyException` be created in `mpl.py` or would anywhere else be better?

connorjward: It should live in `firedrake.exceptions`

firedrake/pyplot/mpl.py
connorjward: I think not supporting it is fine, but we should error helpfully. The code reads quite confusingly to me because I know `gdim` can reasonably be 1-3 so it seems like the `else` branch covers two cases.

firedrake/pyplot/mpl.py
connorjward: Feel free to resolve trivial changes like this when you implement them

firedrake/pyplot/mpl.py
```diff
@@ -83,6 +83,65 @@ def _get_collection_types(gdim, tdim):
     raise ValueError("Geometric dimension must be either 2 or 3!")
 
 
+@PETSc.Log.EventDecorator()
+def scatter(vom: MeshGeometry | Function, axes: matplotlib.axes.Axes = None, **kwargs) -> matplotlib.collections.PathCollection:
```
connorjward:
```suggestion
def scatter(vom: MeshGeometry | Function, axes: matplotlib.axes.Axes | None = None, **kwargs) -> matplotlib.collections.PathCollection:
```

firedrake/pyplot/mpl.py
connorjward: I got confused by the brackets. This is now better thanks.

docs/source/visualisation.rst
```diff
@@ -326,6 +326,49 @@ points per element could be specfied to when calling :func:`plot
 To install matplotlib_, please look at the installation instructions of
 matplotlib.
 
+Visualising a vertex-only mesh
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Firedrake supports the visualisation of point data, represented as a :func:`~.VertexOnlyMesh`, in much the same way as its other plotting routines.
+In particular, :func:`scatter <firedrake.pyplot.scatter>` wraps matplotlib's ``scatter`` method and uses the coordinates of the mesh's constituent points to produce a scatter plot.
+Although separate from :func:`triplot <firedrake.pyplot.triplot>`, it makes most sense to use it in conjuction with :func:`triplot <firedrake.pyplot.triplot>`
+which makes apparent the embedding of the vertex-only mesh inside its parent mesh. As the below code demonstrates, :func:`scatter <firedrake.pyplot.scatter>` gives the user the freedom to pass
```
connorjward, line 335: Oh that's actually really cool

connorjward (commented):
I'm basically happy now I think. Really nice.

achanbour (commented):
Added serial run checks in all plotting methods.

firedrake/pyplot/mpl.py
```diff
@@ -83,6 +83,70 @@ def _get_collection_types(gdim, tdim):
...
+    if not (is_vom or is_function_on_vom):
+        raise TypeError("Expected a VertexOnlyMesh or a Function defined on one.")
+
+    if isinstance(vom, Function):
+        if len(vom.ufl_shape) == 0:
+            # scalar field: colour points by value
+            kwargs["c"] = vom.dat.data_ro
+        elif len(vom.ufl_shape) == 1:
+            # vector field: use quiver instead
+            raise ValueError("Expected 1D Function. Use quiver to plot 2D Functions.")
```
leo-collins:
```suggestion
            raise ValueError("Expected a scalar-valued Function. Use quiver to plot vector-valued Functions.")
```

firedrake/pyplot/mpl.py
```diff
@@ -83,6 +83,70 @@ def _get_collection_types(gdim, tdim):
...
+        else:
+            raise ValueError(
+                f"Cannot plot a rank-{len(vom.ufl_shape)} tensor field; "
+                "only 1D Functions are supported by this method. "
```
leo-collins:
```suggestion
                "only scalar-valued Functions are supported by this method. "
```

firedrake/pyplot/mpl.py
```diff
@@ -83,6 +83,70 @@ def _get_collection_types(gdim, tdim):
     raise ValueError("Geometric dimension must be either 2 or 3!")
 
 
+@PETSc.Log.EventDecorator()
+def scatter(vom: MeshGeometry | Function, axes: matplotlib.axes.Axes | None = None, **kwargs) -> matplotlib.collections.PathCollection:
+    r"""Plot a VertexOnlyMesh as a scatter plot.
```
leo-collins:
```suggestion
    r"""Plot a 2D or 3D VertexOnlyMesh as a scatter plot.
```

firedrake/pyplot/mpl.py
```diff
@@ -83,6 +83,70 @@ def _get_collection_types(gdim, tdim):
...
+            # scalar field: colour points by value
+            kwargs["c"] = vom.dat.data_ro
+        elif len(vom.ufl_shape) == 1:
+            # vector field: use quiver instead
+            raise ValueError("Expected 1D Function. Use quiver to plot 2D Functions.")
+        else:
+            raise ValueError(
+                f"Cannot plot a rank-{len(vom.ufl_shape)} tensor field; "
+                "only 1D Functions are supported by this method. "
+                "For 2D Functions, use quiver.")
```
leo-collins:
```suggestion
                "For vector-valued Functions, use quiver.")
```

firedrake/pyplot/mpl.py
```diff
@@ -336,14 +412,23 @@ def trisurf(function, *args, complex_component="real", **kwargs):
...
-def quiver(function, *, complex_component="real", **kwargs):
-    r"""Make a quiver plot of a 2D vector Firedrake :class:`~.Function`
-
-    :arg function: the vector field to plot
-    :kwarg complex_component: If plotting complex data, which
-        component? (``'real'`` or ``'imag'``). Default is ``'real'``.
-    :arg kwargs: same as for matplotlib :func:`quiver <matplotlib.pyplot.quiver>`
-    :return: matplotlib :class:`Quiver <matplotlib.quiver.Quiver>` object
+def quiver(function: Function, *, complex_component: str = "real", **kwargs) -> matplotlib.quiver.Quiver:
+    r"""Make a quiver plot of a 2D vector Firedrake function.
```
leo-collins:
```suggestion
    r"""Make a quiver plot of a 2D vector Firedrake :class:`~.Function`.
```

firedrake/pyplot/mpl.py
leo-collins: Also should link to VertexOnlyMesh

JHopeCollins:
This is great, thanks! Will it allow plotting points values when the parent mesh is 1D?

achanbour:
> This is great, thanks! Will it allow plotting points values when the parent mesh is 1D?

Matplotlib's [scatter](https://matplotlib.org/stable/api/_as_gen/matplotlib.pyplot.scatter.html) assumes at least 2D so the code currently doesn't support 1D and raises an Exception.

https://github.com/firedrakeproject/firedrake/blob/3d415011da7240b4ebe1a0493edd069b6b4d3807/firedrake/pyplot/mpl.py#L131-L138

JHopeCollins:
> Matplotlib's [scatter](https://matplotlib.org/stable/api/_as_gen/matplotlib.pyplot.scatter.html) assumes at least 2D so the code currently doesn't support 1D and raises an Exception.

Of course, I'd forgotten there were separate calls for the 1D and 2D plotting

firedrake/pyplot/mpl.py
```diff
@@ -83,6 +83,70 @@ def _get_collection_types(gdim, tdim):
     raise ValueError("Geometric dimension must be either 2 or 3!")
 
 
+@PETSc.Log.EventDecorator()
+def scatter(vom: MeshGeometry | Function, axes: matplotlib.axes.Axes | None = None, **kwargs) -> matplotlib.collections.PathCollection:
```
dham, line 87:
```suggestion
def scatter(vom_or_function: MeshGeometry | Function, axes: matplotlib.axes.Axes | None = None, **kwargs) -> matplotlib.collections.PathCollection:
```

dham (changes requested):

connorjward (approved):

PR #5042, opened 2026-04-20
Tighten tao tolerances in adjoint tests

JHopeCollins:
One of the TAO optimisation tests started failing recently because it is now just missed the error tolerance on the optimised solution. Presumably something has changed in the solver stack so it is getting a slightly different solution.

Tightening up the tolerance makes the solution pass again. I looked at some of the other tests that use these parameter sets and some of them were also only _just_ passing with the previous tolerance, so I don't think we mind the slight increase in the test durations.


connorjward (approved):

connorjward:
Should this go into release?

JHopeCollins:
> Should this go into release?

Yes, I'll move it over

JHopeCollins:
These changes will go into release in #5047 instead.

PR #5043, opened 2026-04-21
Fix MG transfers for elements that need cell sizes

pbrubeck:
# Description
For zany elements that are mapped with a rescaling by `mesh.cell_sizes` (such as AWc and HZ), MG will trigger a compile error. This PR adds the missing support for `cell_sizes` in the MG transfer codegen.

### DIFF
--- a/firedrake/mg/interface.py
+++ b/firedrake/mg/interface.py
@@ -283,8 +283,14 @@ def inject(fine, coarse):
     return coarse
 
 
-def _topology_permutation(f, V):
-    """Permute the Function f with the same entity ordering as the FunctionSpace V."""
+def apply_entity_permutation(f, V):
+    """Permute the Function f with the same entity ordering as the FunctionSpace V.
+    
+    The multigrid transfer kernels require geometric quantities of the source mesh
+    to have the same entity ordering as the source Function being transfered.
+    We need to permute f if V is a RestrictedFunctionSpace, because the
+    geometric quantities will live on an unrestricted FunctionSpace.
+    """
     if V.boundary_set:
         W = V.reconstruct(element=f.function_space().ufl_element())
         f = Function(W).assign(f)
@@ -293,9 +299,9 @@ def _topology_permutation(f, V):
 
 def get_coordinates(V):
     """Return mesh coordinates with the same entity ordering as V."""
-    return _topology_permutation(V.mesh().coordinates, V)
+    return apply_entity_permutation(V.mesh().coordinates, V)
 
 
 def get_cell_sizes(V):
     """Return cell sizes with the same entity ordering as V."""
-    return _topology_permutation(V.mesh().cell_sizes, V)
+    return apply_entity_permutation(V.mesh().cell_sizes, V)
--- a/firedrake/mg/interface.py
+++ b/firedrake/mg/interface.py
@@ -285,7 +285,6 @@ def inject(fine, coarse):
 
 def apply_entity_permutation(f, V):
     """Permute the Function f with the same entity ordering as the FunctionSpace V.
-    
     The multigrid transfer kernels require geometric quantities of the source mesh
     to have the same entity ordering as the source Function being transfered.
     We need to permute f if V is a RestrictedFunctionSpace, because the
--- a/firedrake/mg/interface.py
+++ b/firedrake/mg/interface.py
@@ -285,6 +285,7 @@ def inject(fine, coarse):
 
 def apply_entity_permutation(f, V):
     """Permute the Function f with the same entity ordering as the FunctionSpace V.
+    
     The multigrid transfer kernels require geometric quantities of the source mesh
     to have the same entity ordering as the source Function being transfered.
     We need to permute f if V is a RestrictedFunctionSpace, because the
--- a/firedrake/mg/interface.py
+++ b/firedrake/mg/interface.py
@@ -79,21 +79,26 @@ def prolong(coarse, fine):
         fine_to_coarse = utils.fine_node_to_coarse_node_map(Vf, Vc)
         fine_to_coarse_coords = utils.fine_node_to_coarse_node_map(Vf, coarse_coords.function_space())
         kernel = kernels.prolong_kernel(coarse, Vf)
-
         # XXX: Should be able to figure out locations by pushing forward
         # reference cell node locations to physical space.
         # x = \sum_i c_i \phi_i(x_hat)
         node_locations = utils.physical_node_locations(Vf)
+        kernel_args = [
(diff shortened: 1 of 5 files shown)


firedrake/mg/interface.py
```diff
@@ -267,9 +283,19 @@ def inject(fine, coarse):
...
-def get_coordinates(V):
-    coords = V.mesh().coordinates
+def _topology_permutation(f, V):
+    """Permute the Function f with the same entity ordering as the FunctionSpace V."""
     if V.boundary_set:
-        W = V.reconstruct(element=coords.function_space().ufl_element())
-        coords = Function(W).interpolate(coords)
-    return coords
+        W = V.reconstruct(element=f.function_space().ufl_element())
+        f = Function(W).assign(f)
```
connorjward: Does assign DTRT for restricted function spaces? Also this function is very bizarre and deserves a better name and motivating comment.

connorjward (changes requested):
Mostly fine.

firedrake/mg/interface.py
pbrubeck: It does the right thing https://github.com/firedrakeproject/firedrake/pull/4890

firedrake/mg/interface.py
```diff
@@ -267,9 +283,19 @@ def inject(fine, coarse):
     return coarse
 
 
-def get_coordinates(V):
-    coords = V.mesh().coordinates
+def _topology_permutation(f, V):
+    """Permute the Function f with the same entity ordering as the FunctionSpace V."""
```
pbrubeck:
```suggestion
def apply_entity_permutation(f, V):
    """Permute the Function f with the same entity ordering as the FunctionSpace V.
    
    The multigrid transfer kernels require geometric quantities of the source mesh
    to have the same entity ordering as the source Function being transfered.
    We need to permute f if V is a RestrictedFunctionSpace, because the
    geometric quantities will live on an unrestricted FunctionSpace.
    """
```

firedrake/mg/interface.py
```diff
@@ -267,9 +283,19 @@ def inject(fine, coarse):
...
-        coords = Function(W).interpolate(coords)
-    return coords
+        W = V.reconstruct(element=f.function_space().ufl_element())
+        f = Function(W).assign(f)
+    return f
+
+
+def get_coordinates(V):
+    """Return mesh coordinates with the same entity ordering as V."""
+    return _topology_permutation(V.mesh().coordinates, V)
```
pbrubeck:
```suggestion
    return apply_entity_permutation(V.mesh().coordinates, V)
```

firedrake/mg/interface.py
```diff
@@ -267,9 +283,19 @@ def inject(fine, coarse):
...
+
+
+def get_cell_sizes(V):
+    """Return cell sizes with the same entity ordering as V."""
+    return _topology_permutation(V.mesh().cell_sizes, V)
```
pbrubeck:
```suggestion
    return apply_entity_permutation(V.mesh().cell_sizes, V)
```

firedrake/mg/interface.py
pbrubeck: Seems like this logic could be added to `utils.fine_node_to_coarse_node_map`. But this seems too involved and tangential to this PR.

firedrake/mg/interface.py
```diff
@@ -267,9 +283,25 @@ def inject(fine, coarse):
     return coarse
 
 
-def get_coordinates(V):
-    coords = V.mesh().coordinates
+def apply_entity_permutation(f, V):
+    """Permute the Function f with the same entity ordering as the FunctionSpace V.
+
```
pbrubeck:
```suggestion

```

firedrake/mg/interface.py
connorjward: This is confusing to me because it seems like we can have `apply_entity_permutation` with arbitrary `V`. Isn't it the case that `V` must always be `f.function_space().reconstruct(boundary_set=())`? In which case `V` shouldn't an argument and this function should be called `unrestrict(f)` or similar.

firedrake/mg/interface.py
```diff
@@ -267,9 +283,24 @@ def inject(fine, coarse):
     return coarse
 
 
-def get_coordinates(V):
-    coords = V.mesh().coordinates
+def apply_entity_permutation(f, V):
+    """Permute the Function f with the same entity ordering as the FunctionSpace V.
```
pbrubeck:
```suggestion
    """Permute the Function f with the same entity ordering as the FunctionSpace V.
    
```

firedrake/mg/interface.py
connorjward: This is confusing to me because it seems like we can have apply_entity_permutation with arbitrary `V`. Isn't it the case that `V` must always be `f.function_space().reconstruct(boundary_set=())`? In which case `V` shouldn't an argument and this function should be called `unrestrict(f)` or similar.

pbrubeck: It is the other way around. f lives on the unrestricted space and V is the restricted space.

pbrubeck: The problem is that reconstruct does not take `boundary_set` as kwarg

pbrubeck: And the reason is to do with mixed function spaces with different restrictions (a very common case)

connorjward: > The problem is that reconstruct does not take boundary_set as kwarg

OK. I think this will be sufficiently clear if the following changes are made:
1. Rename the function to `maybe_restrict_function(f, V)` or similar. I don't think the function has anything to do with entity permutations so that just causes a lot of confusion.
2. In the docstring make it clear that `f` and `V` are supposed to be compatible
3. Add an `assert f.function_space().ufl_element() == V.ufl_element()`

pbrubeck: I think we don't need this anymore. The RestrictedFunctionSpace + MG test seems to be working fine without reconstructing the coordinates on the RFS

connorjward: That's way better.

connorjward (dismissed):

firedrake/mg/kernels.py
```diff
@@ -18,6 +18,7 @@
 
 import ufl
 import tsfc
+import pyop2
```
pbrubeck: @connorjward how do I fix the unused import? We only need it for the type hint

pbrubeck: Okay, I just needed to remove this. I'll wait for docs CI and merge

PR #5046, opened 2026-04-22
Fix ASMStarPC on RestrictedFunctionSpace

pbrubeck:
# Description
This solver combination was segfaulting. We just needed to exclude the restricted DOFs from the patches.

### DIFF
--- a/firedrake/preconditioners/asm.py
+++ b/firedrake/preconditioners/asm.py
@@ -169,7 +169,7 @@ class ASMStarPC(ASMPatchPC):
 
         # Accessing .indices causes the allocation of a global array,
         # so we need to cache these for efficiency
-        V_local_ises_indices = tuple(iset.indices for iset in V.dof_dset.local_ises)
+        V_local_ises_indices = get_local_ises_indices(V)
 
         # Build index sets for the patches
         ises = []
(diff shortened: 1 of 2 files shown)


firedrake/preconditioners/asm.py
```diff
@@ -528,11 +534,26 @@ def get_patches(self, V):
...
                     ises.append(iset)
         return ises
 
 
+def get_local_ises_indices(V):
+    """Return the local indices of each subspace of V.
+    The restricted DOFs will be masked for a RestrictedFunctionSpace.
+    """
+    # Accessing .indices causes the allocation of a global array,
+    # so we need to cache these for efficiency
```
connorjward: This comment doesn't make sense here. The 'caching' happens in the scope of the caller function.

connorjward (changes requested):
Otherwise good

firedrake/preconditioners/asm.py
connorjward: At least that's my understanding, not 100% confident

pbrubeck: I'll leave the comments where they were then.

connorjward (approved):

PR #5049, opened 2026-04-22
fix linkcheck

leo-collins:
Fix failing link https://missendenabbey.co.uk/


leo-collins:
@connorjward David said Missenden abbey has in fact shut down - should we remove this link from the documentation?

connorjward:
> @connorjward David said Missenden abbey has in fact shut down - should we remove this link from the documentation?

My hero!

Yeah that link is not coming back so we should just remove it entirely.

connorjward (approved):

Issue #5045, opened 2026-04-22
BUG: Compilation error for MG transfer of HDiv element on immersed mesh


JHopeCollins:
**Describe the bug**
I'm getting a compilation error from pyop2 when trying use MG transfer operators with HDiv functions on immersed meshes with `firedrake-2026.4.0`

**Steps to Reproduce**
```python
from firedrake import *
basemesh = UnitIcosahedralSphereMesh(
    refinement_level=0, degree=1
)
mc, mf = MeshHierarchy(
    basemesh, refinement_levels=1)
mc.init_cell_orientations(SpatialCoordinate(mc))
mf.init_cell_orientations(SpatialCoordinate(mf))
family = "BDM"
degree = 1
Vc = FunctionSpace(mc, family, degree)
Vf = FunctionSpace(mf, family, degree)
uc = Function(Vc)
uf = Function(Vf)
transfers = fd.TransferManager()
transfers.inject(uf, uc)
```

**Expected behavior**
This worked fine in `firedrake-2025.10.3`

**Error message**
Running with `PYOP2_DEBUG=1 python mfe.py` gives the following error, and similar errors if you try `transfers.restrict` or `transfers.prolong` with `Cofunctions`.

```
Traceback (most recent call last):
  File "/home/jhc/codes/fd/fd-main/src/asQ/utils/mg.py", line 119, in <module>
    transfers.inject(uf, uc)
  File "/home/jhc/codes/fd/fd-main/src/firedrake/firedrake/mg/embedded.py", line 294, in inject
    self.op(uf, uc, transfer_op=Op.INJECT)
  File "petsc4py/PETSc/Log.pyx", line 250, in petsc4py.PETSc.Log.EventDecorator.decorator.wrapped_func
  File "petsc4py/PETSc/Log.pyx", line 251, in petsc4py.PETSc.Log.EventDecorator.decorator.wrapped_func
  File "/home/jhc/codes/fd/fd-main/src/firedrake/firedrake/mg/embedded.py", line 245, in op
    self._native_transfer(target_element, transfer_op)(source, target)
  File "petsc4py/PETSc/Log.pyx", line 250, in petsc4py.PETSc.Log.EventDecorator.decorator.wrapped_func
  File "petsc4py/PETSc/Log.pyx", line 251, in petsc4py.PETSc.Log.EventDecorator.decorator.wrapped_func
[...]

PR #5052, opened 2026-04-22
Use embedded transfer for HDiv/HCurl on manifolds

pbrubeck:
# Description
Fixes https://github.com/firedrakeproject/firedrake/issues/5045

Interpolation into H(div) on a manifold is not a well-defined operation. Because the edge normals are not necessarily aligned, the degrees of freedom mean different things for the two cells on either side of the edge. 

This PR adds support for cannonical prolongation and restriction by evaluating degrees of freedom, by providing `cell_orientations` to the kernels that require them. However, this results in prolongation and restriction kernels that fail to be the adjoint of each other, presumably because dofs are not evaluting to the same value on different cells.

The solution here is to fall back to the embedded DG L2 projection transfer kernels for non-CG/DG elements on manifolds.

### DIFF
--- a/firedrake/mg/embedded.py
+++ b/firedrake/mg/embedded.py
@@ -51,7 +51,6 @@ class TransferManager(object):
     def is_native(self, element, gdim, op):
         if element in self.native_transfers:
             return self.native_transfers[element][op] is not None
-
         if isinstance(element.cell, ufl.TensorProductCell):
             if isinstance(element, finat.ufl.TensorProductElement):
                 return all(self.is_native(e, gdim, op) for e in element.factor_elements)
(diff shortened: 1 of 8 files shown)


firedrake/mg/embedded.py
```diff
@@ -48,28 +48,35 @@ def __init__(self, *, native_transfers=None, use_averaging=True):
...
         finat_element = create_element(element)
         try:
             finat_element.dual_basis
-            return True
         except NotImplementedError:
             return False
-
-    def _native_transfer(self, element, op):
+        else:
+            # We can only interpolate into CG/DG spaces on manifold meshes
```
connorjward: The PR description was very helpful for me to understand why this is the case. Could the motivation also be described here.

connorjward (changes requested):
I think this is fine. Thanks

tests/firedrake/multigrid/test_embedded_transfer.py
```diff
@@ -159,6 +169,27 @@ def solver(V, space, solver_parameters):
...
+    base = UnitIcosahedralSphereMesh(refinement_level=0, degree=1,
+                                     distribution_parameters=distribution_parameters)
+
+    mh = MeshHierarchy(base, refinement_levels=3)
+    for m in mh:
+        m.init_cell_orientations(SpatialCoordinate(m))
+    return mh[-1]
+
+
+@pytest.mark.parallel([1])
```
connorjward: Why are we losing the parallel test?

connorjward (changes requested):
Last thing. Otherwise happy

tests/firedrake/multigrid/test_embedded_transfer.py
pbrubeck: I included the parallel test, found another bug (`distribution_parameters` being ignored) and fixed it.

Issue #5048, opened 2026-04-22
BUG: Interpolation between meshes leaks memory?


Patol75:
**Describe the bug**
I have been using animate to generate meshes with targeted refinement in specific domain locations based on field values. Essentially, as the simulation progresses, fields evolve, and the mesh is updated every few time steps to provide increased resolution where needed. As part of this endeavour, I have noticed that my simulations leak memory. My investigation led me to cross-mesh interpolation.

**Steps to Reproduce**
Consider the basic Helmholtz equation introductory tutorial, trimmed down to the following lines:
```python
from firedrake import *

mesh = UnitSquareMesh(10, 10)

V = FunctionSpace(mesh, "CG", 1)

trial = TrialFunction(V)
test = TestFunction(V)

f = Function(V)
x, y = SpatialCoordinate(mesh)
f.interpolate((1 + 8 * pi * pi) * cos(x * pi * 2) * cos(y * pi * 2))

a = (inner(grad(trial), grad(test)) + inner(trial, test)) * dx
L = inner(f, test) * dx

u = Function(V)

solve(a == L, u, solver_parameters={"ksp_type": "cg", "pc_type": "none"})
```

Running this code serially and appending `-log_view` yields the following as part of the PETSc output:
```
Object Type          Creations   Destructions. Reports information only for process 0.

--- Event Stage 0: Main Stage

           Container     6              6
           Index Set    78             78
   IS L to G Mapping     1              1
             Section    15             15
   Star Forest Graph    20             20
              Vector    13             13
              Matrix     2              2
      Preconditioner     1              1
       Krylov Solver     1              1
     DMKSP interface     1              1
                SNES     1              1
              DMSNES     1              1
    Distributed Mesh     7              7
            DM Label    23             23
      Mesh Transform     1              1
    GraphPartitioner     2              2
     Discrete System     9              9
           Weak Form     9              9
```

As one can see, all created PETSc objects are destroyed by the end of the simulation.

If I now append the following lines to the above snippet:
```python
new_mesh = UnitSquareMesh(20, 20)
new_V = FunctionSpace(new_mesh, "CG", 1)
g = Function(new_V)
g.interpolate(f)
```
and run the updated code similarly, I get the following comparable PETSc report:
```
Object Type          Creations   Destructions. Reports information only for process 0.

--- Event Stage 0: Main Stage

           Container     8              5
           Index Set   161            140
   IS L to G Mapping     1              0
             Section    35             23
[...]

PR #5055, opened 2026-04-23
Use mat context for Vom onto vom matfree interpolation

leo-collins:
https://github.com/firedrakeproject/firedrake/issues/5048 pointed out a memory leak caused by a reference cycle caused by the `VomOntoVomMat` being it's own matrix context.

This PR creates a separate `VomOntoVomMatContext` class and uses this for the PETSc python type mat.

### DIFF
--- a/firedrake/interpolation.py
+++ b/firedrake/interpolation.py
@@ -47,7 +47,7 @@ from firedrake.exceptions import (
 
 from mpi4py import MPI
 
-from pyadjoint.tape import stop_annotating, no_annotations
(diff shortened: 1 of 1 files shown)


connorjward (approved):
I assume that most of this is just code movement? It's hard to pick out the specific thing that's changed.

leo-collins:
Essentially yes. `VomOntoVomMat` becomes `VomOntoVomMatContext`, and I had to change some arguments to avoid creating a reference cycle. There's no new logic here.

PR #5059, opened 2026-04-24
Improve building compiled extensions using Makefile

connorjward:
setuptools doesn't provide a way to clean up compiled extensions if they are built inplace. make clean was therefore a best-guess, highly bitrotted removal of the inplace extensions. It was removing many files that no longer exist and failing to remove others.

Instead of this approach here we just provide a way to force a rebuild of all extensions. The end result is the same but robust to future file changes.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -4,12 +4,12 @@ all: ext
 .PHONY: ext
 ext:
 	@echo "    Building extension modules"
-	@python setup.py build_ext --inplace > build.log 2>&1 || cat build.log
+	@python setup.py build_ext --inplace 2>&1 | tee cat build.log
 
 .PHONY: allext
 allext:
 	@echo "    Force building all extension modules"
-	@python setup.py build_ext --inplace --force > build.log 2>&1 || cat build.log
+	@python setup.py build_ext --inplace --force 2>&1 | tee cat build.log
 
 .PHONY: lint
 lint: srclint actionlint dockerlint
--- a/Makefile
+++ b/Makefile
@@ -1,10 +1,15 @@
 .PHONY: all
-all: modules
+all: ext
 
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -1,11 +1,16 @@
 .PHONY: all
-all: modules
+all: ext
 
-.PHONY: modules
-modules:
+.PHONY: ext
+ext:
 	@echo "    Building extension modules"
 	@python setup.py build_ext --inplace > build.log 2>&1 || cat build.log
```
JHopeCollins:
```suggestion
	@python setup.py build_ext --inplace 2>&1 | tee cat build.log
```

Makefile
```diff
@@ -1,11 +1,16 @@
...
 
+.PHONY: allext
+allext:
+	@echo "    Force building all extension modules"
+	@python setup.py build_ext --inplace --force > build.log 2>&1 || cat build.log
```
JHopeCollins:
```suggestion
	@python setup.py build_ext --inplace --force 2>&1 | tee cat build.log
```

JHopeCollins (changes requested):
This is much better than trying to manually maintain a file list. I'm happy other than the naming.

My only suggestion is better naming because `all`, `ext`, and `allext` isn't very clear.
How about changing `allext` to `ext-force`?

If we think that having an `all` phony is useful then we could also have an `all-force: ext-force` recipe.

connorjward:
I've removed `make all`. It was redundant and I only left it there because I forgot that `make` without arguments just executes the top thing. Also renamed `allext` to `extforce` (no spaces seems like more traditional makefile notation).

JHopeCollins (approved):
Yep, great. Thanks!

PR #5067, opened 2026-04-29
Add slack invite to linkcheck skip

connorjward:
This seems quite an important link to check, so the fact that it's running into bot checks is frustrating.


JHopeCollins (approved):
That's very annoying but if its failing there's not much we can do.

PR #5068, opened 2026-04-30
Fix website build

connorjward:
Investigating intersphinx issues. The docs build passes here but fails for the website.


connorjward:
Forcing through as this is a critical website fix and not controversial.

PR #5072, opened 2026-05-01
stderr in firedrake-configure and Python version check

connorjward:
Pipe firedrake-configure errors to stderr and check min Python version

### DIFF
--- a/scripts/check-config
+++ b/scripts/check-config
@@ -32,8 +32,12 @@ def check_min_python_version() -> None:
         "requires-python = \">=(.*)\""
     )
     check_file_contains_pattern(
-        REPO_ROOT / "docs/source/install.rst",
+        REPO_ROOT / "docs" / "source" / "install.rst",
         f"Python \\({min_python_version} or greater\\)"
+    )
+    check_file_contains_pattern(
+        REPO_ROOT / "scripts" / "firedrake-configure",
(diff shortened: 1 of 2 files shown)


JHopeCollins:
Why do we need these changes?

connorjward:
The main idea is that we should fail nicely on older Pythons, but in general we want to pipe to stderr because users are piping stdout into some other process so won't see errors otherwise.

But I also think using stderr is nicer than raising exceptions for scripts.

scripts/check-config
```diff
@@ -32,9 +32,13 @@ def check_min_python_version() -> None:
         "requires-python = \">=(.*)\""
     )
     check_file_contains_pattern(
-        REPO_ROOT / "docs/source/install.rst",
+        REPO_ROOT / "docs" / "source" / "install.rst",
```
JHopeCollins, line 35: Is this doing something different or is it just better style?

JHopeCollins:
> but in general we want to pipe to stderr because users are piping stdout into some other process so won't see errors otherwise.

This makes sense.

JHopeCollins (approved):

scripts/check-config
connorjward: Just nicer style IMO

PR #5073, opened 2026-05-01
Refactor firedrake configure

connorjward:
All logic is now gone. All possible configs are explicitly enumerated.

This is needed because the previous approach was not extensible (e.g.
supporting Ubuntu 24.04 and 26.04 together).

This opens the door to having 'community' configurations and even
externally provided configuration files.

### DIFF
--- a/docs/Makefile
+++ b/docs/Makefile
@@ -86,22 +86,15 @@ source/team.rst: source/team2.py
 
 source/apt_deps.txt:
 	python3 ../scripts/firedrake-configure \
-		--package-manager apt-x86_64 \
+		--package-manager ubuntu24.04-x86_64 \
 		--show-system-packages > source/apt_deps.tmp
 	mv source/apt_deps.tmp source/apt_deps.txt
 
 source/homebrew_deps.txt:
 	python3 ../scripts/firedrake-configure \
-		--package-manager brew-arm64 \
+		--package-manager macos-arm64 \
 		--show-system-packages > source/homebrew_deps.tmp
 	mv source/homebrew_deps.tmp source/homebrew_deps.txt
-
-source/minimal_apt_deps.txt:
-	python3 ../scripts/firedrake-configure \
-		--package-manager apt-x86_64 \
-		--show-minimal-system-packages > source/minimal_apt_deps.tmp
-	mv source/minimal_apt_deps.tmp source/minimal_apt_deps.txt
-
 
 source/petsc_configure_options.txt:
 	python3 ../scripts/firedrake-configure \
--- a/docs/Makefile
+++ b/docs/Makefile
@@ -43,7 +43,7 @@ help:
 
 TARGETS = html dirhtml singlehtml pickle json htmlhelp qthelp devhelp epub latex latexpdf text man changes linkcheck doctest gettext
 
-GENERATED_FILES = source/team.rst source/demos source/element_list.csv source/apt_deps.txt source/homebrew_deps.txt source/minimal_apt_deps.txt source/petsc_configure_options.txt
+GENERATED_FILES = source/team.rst source/demos source/element_list.csv source/apt_deps.txt source/homebrew_deps.txt source/petsc_configure_options.txt
 
 publishpdf:
 	env FIREDRAKE_MANUAL_RELEASE=`date +%Y-%m` $(MAKE) latexpdf
--- a/docs/Makefile
+++ b/docs/Makefile
@@ -86,19 +86,19 @@ source/team.rst: source/team2.py
 
 source/apt_deps.txt:
 	python3 ../scripts/firedrake-configure \
-		--package-manager ubuntu24.04-x86_64 \
+		--os ubuntu24.04-x86_64 \
 		--show-system-packages > source/apt_deps.tmp
 	mv source/apt_deps.tmp source/apt_deps.txt
 
 source/homebrew_deps.txt:
 	python3 ../scripts/firedrake-configure \
-		--package-manager macos-arm64 \
+		--os macos-arm64 \
 		--show-system-packages > source/homebrew_deps.tmp
 	mv source/homebrew_deps.tmp source/homebrew_deps.txt
 
 source/petsc_configure_options.txt:
 	python3 ../scripts/firedrake-configure \
-		--no-package-manager \
+		--os unknown \
 		--show-petsc-configure-options > source/petsc_configure_options.tmp
 	mv source/petsc_configure_options.tmp source/petsc_configure_options.txt
 
--- a/scripts/firedrake-configure
+++ b/scripts/firedrake-configure
@@ -10,6 +10,36 @@
 # As a matter of policy, new package managers and archs should only be added to
 # this file if they are *tested in CI*.
 
+# Start by checking the Python version, older Pythons may not be able to run
+# this script.
+import sys
+
+
+def stdout(msg: str) -> None:
+    """Write to stdout."""
+    print(msg, file=sys.stdout)
+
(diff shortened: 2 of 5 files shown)


scripts/firedrake-configure
```diff
@@ -10,94 +10,112 @@
...
+    os_ = args.os if args.os else detect_os()
 
-SUPPORTED_PETSC_VERSION = "v3.25.0"
-# CUDA 13.1 is currently not supported by GPU drivers on Firedrake CI systems.
-SUPPORTED_CUDA_VERSION = "13.0"
+    if args.gpu_arch != GPUPlatform.NO_GPU and os_ == OS.MACOS_ARM64:
+        error(
+            "GPU-compatible PETSc builds are currently only supported"
+            "on Linux"
+        )
```
JHopeCollins: Should the test be for `os_ != <LINUX_VALUE>` rather than for MACOS?

scripts/firedrake-configure
```diff
@@ -110,584 +128,710 @@ that are used inside PETSc 'configure' instead of building from source.
...
-def sniff_package_manager() -> Optional[PackageManager]:
+def detect_os() -> "OS":
     if platform.system() == "Linux":
         if has_apt():
             if platform.machine() == "x86_64":
-                return LINUX_APT_X86_64
+                return OS.UBUNTU_2404_X86_64
             elif platform.machine() == "aarch64":
-                return LINUX_APT_AARCH64
+                return OS.UBUNTU_2404_AARCH64
```
JHopeCollins: This should check whether it is actually `24.04`.

scripts/firedrake-configure
```diff
@@ -110,584 +128,710 @@ that are used inside PETSc 'configure' instead of building from source.
...
+    error(
+        "Cannot identify operating system, if you want to configure Firedrake "
+        "without any system packages please pass '--os unknown'"
     )
-    abort()
 
 
 def has_apt() -> bool:
     try:
         subprocess.run(["apt", "--version"], capture_output=True)
```
JHopeCollins: `which apt` and check if it's empty rather than relying on errors?

scripts/firedrake-configure
```diff
@@ -110,584 +128,710 @@ that are used inside PETSc 'configure' instead of building from source.
...
+    cflags: Sequence[str] = ()
+    cxxflags: Sequence[str] = ()
+    fflags: Sequence[str] = ()
+    include_dirs: Sequence[str] = ()
+    library_dirs: Sequence[str] = ()
+    extra_env_vars: Mapping[str, str] = dataclasses.field(default_factory=dict)
+    extra_url: str | None = None
+    """The URL to a deb package that will enable vendor-specific software development
+    repositories, or None if not required.
+    """
```
JHopeCollins, line 280: This is useful. Please can we also have docstrings for the other attributes?

scripts/firedrake-configure
```diff
@@ -110,584 +128,710 @@ that are used inside PETSc 'configure' instead of building from source.
...
+        "--with-fortran-bindings=0",
+        "--with-shared-libraries=1",
+        "--with-strict-petscerrorcode",
+    ]
+
+    def show_system_packages(self) -> None:
+        if self.system_packages is None:
+            error(
+                "Don't know what system dependencies to install on this platform, "
+                "please install them manually"
```
JHopeCollins: This should print some description of the `Arch` (e.g. name, platform, etc) so people know if they actually selected the one they thought they did.

scripts/firedrake-configure
```diff
@@ -110,584 +128,710 @@ that are used inside PETSc 'configure' instead of building from source.
...
+            print_and_stop(" ".join(self.system_packages))
+
+    def show_petsc_configure_options(self) -> None:
+        include_dirs = [f"-I{idir}" for idir in self.include_dirs]
+        library_dirs = [f"-L{ldir}" for ldir in self.library_dirs]
+        cflags = " ".join((*self.cflags, *include_dirs, *library_dirs))
+        cxxflags = " ".join((*self.cxxflags, *include_dirs, *library_dirs))
+        fflags = " ".join((*self.fflags, *include_dirs, *library_dirs))
+        opts = [
+            f"PETSC_ARCH={self.name}",
```
JHopeCollins: Could this be `arch-{self.name}`?

connorjward: Gone for `arch-firedrake-{self.name}`

scripts/firedrake-configure
connorjward: Done

scripts/firedrake-configure
connorjward: Done

scripts/firedrake-configure
connorjward: I've dropped this and am detecting Ubuntu directly

scripts/firedrake-configure
connorjward: Done

scripts/firedrake-configure
connorjward: I've removed this as the default error message is quite informative. I don't think we need a GPU-specific one.

docs/Makefile
```diff
@@ -86,23 +86,16 @@ source/team.rst: source/team2.py
 
 source/apt_deps.txt:
 	python3 ../scripts/firedrake-configure \
-		--package-manager apt-x86_64 \
+		--package-manager ubuntu24.04-x86_64 \
```
JHopeCollins: Is `--package-manager` deprecated now?
```suggestion
		--os ubuntu24.04-x86_64 \
```

docs/Makefile
```diff
@@ -86,23 +86,16 @@ source/team.rst: source/team2.py
...
 		--show-system-packages > source/apt_deps.tmp
 	mv source/apt_deps.tmp source/apt_deps.txt
 
 source/homebrew_deps.txt:
 	python3 ../scripts/firedrake-configure \
-		--package-manager brew-arm64 \
+		--package-manager macos-arm64 \
```
JHopeCollins:
```suggestion
		--os macos-arm64 \
```

docs/Makefile
connorjward: Yes. Good spot. I'll track the couple of changes there down.

scripts/firedrake-configure
```diff
@@ -110,585 +126,724 @@ that are used inside PETSc 'configure' instead of building from source.
...
+            "/usr/include/superlu",
+            "/usr/include/superlu-dist",
+        ],
+        library_dirs=[
+            "/usr/lib/x86_64-linux-gnu/hdf5/openmpi",
+        ],
+    ),
+
+    (OS.UBUNTU_2404_X86_64, ScalarType.DEFAULT, GPUPlatform.CUDA): Arch(
+        name="firedrake-default-cuda",
```
JHopeCollins:
```suggestion
        name="default-cuda",
```

scripts/firedrake-configure
```diff
@@ -77,6 +77,7 @@ def main():
     os_ = args.os if args.os else detect_os()
 
     arch_key = (os_, args.scalar_type, args.gpu_arch)
+    arch_key_str = f"({', '.join(f'{k.__class__.__name__}: {k.name}' for k in arch_key)})"
```
JHopeCollins, line 80: Will `args.scalar_type` and `args.gpu_arch` have a `.name` attribute?

connorjward: They're enums, so yes. I have tested this

JHopeCollins (approved):

PR #5074, opened 2026-05-01
Fix for quadrature rule with mixed dimension macroelements

pbrubeck:
# Description
Codim-1 submesh assembly with macroelements was failing to get the right cell to build the quadrature rule.


tests/firedrake/submesh/test_submesh_assemble.py
```diff
@@ -567,3 +567,33 @@ def expr(m):
 
     A_ref.petscmat.axpy(-1, A.petscmat)
     assert np.isclose(A_ref.petscmat.norm(PETSc.NormType.NORM_FROBENIUS), 0)
+
+
+def test_submesh_assemble_facet_macroelement():
+    # Test that the macro quadrautre rule is correctly selected
```
connorjward:
```suggestion
    # Test that the macro quadrature rule is correctly selected
```

connorjward (dismissed):
Makes sense to me

PR #5076, opened 2026-05-04
ArrayBuffer Lazy-assignment for GPU Context

SamSJackson:
# Description

This PR introduces a context manager to allow dynamic assignment of PyOP3 arrays on a given device.
Devices are defined within the new `device.py` module, where internal array management is also contained.

Key modifications:
- Introduce `device.py` module to represent offloading devices as offloading context manager
- Change `buffer.py` to lazily-evaluate data with respect to the current context (i.e. host or offloading device)
     - `_lazy_data` is now a dictionary object, mapping between Device objects and respective arrays.
     - `_data` property lazily-evaluates to appropriate data for the given context. If the data is not up-to-date, as per state property, the data is copied.
- All data is copied lazily. Entering and exiting the context window will not automatically transfer data between devices.
- Buffers are maintained between context windows - exiting a context window will not release the memory on the device.
- All device-specific array management is kept within `device.py` so `buffer.py` can remain device/gpu-agnostic - apart from some type-hinting, i.e. `cp.ndarray`.

Notable issues:
- CuPy has no support for `writeable` flag. It will throw away the flag when converting from NumPy objects (and it will not return if converted back)
- Defaultdict is used for our state dictionary. Previously discussed that neither Connor or I like this but I cannot think of another approach.
    - Main issue: If an array is assigned, it does not have any knowledge of other devices beyond that in its current context. As such, there is also no state counter assigned to it. Hence, it is difficult to allow the user to check the buffer on other devices, even if they are initialised. Example can be seen in `./pyop3_gpu_demo.py` from asserts before entering context manager.
    - Potential approaches: 
        - Dictionary wrapper so we can create a more strict defaultdict (bit extreme and needless maintenance but should work).
        - Do not allow users to check state of array on device for which it has not entered context.
    - Open to any advice or solutions for this.


pyop3/buffer.py
```diff
@@ -222,20 +229,23 @@ def handle(self, *, nest_indices: tuple[tuple[int, ...], ...] = ()) -> Any:
...
+    _lazy_data: dict[Device, np.ndarray | cp.ndarray] = dataclasses.field(repr=False)
     sf: StarForest
     _name: str
     _constant: bool
     _rank_equal: bool
     _ordered: bool
 
+    # TODO: Connor and I both dislike defaultdict but I can't think of an alternative atm 
+    _state: collections.defaultdict[Device, int]
+    _last_updated_device: Device
```
connorjward: This is known from the state, we don't need an attribute for it here. Maybe
```py
@property
def _last_updated_device(self):
  ...
```
might be useful

pyop3/buffer.py
```diff
@@ -305,19 +324,24 @@ def dtype(self) -> np.dtype:
...
-            data = self._lazy_data.copy()
+            data = {obj: arr.copy() for obj, arr in self._lazy_data.items()}
         else:
-            data = np.zeros_like(self._lazy_data)
+            data = {obj: obj.zeros_like(arr) for obj, arr in self._lazy_data.items()}
         return self.__record_init__(_name=name, _lazy_data=data)
 
     is_nested: ClassVar[bool] = False
+    
+    def get_context(self) -> Device:
```
connorjward: This is a global attribute so I don't think it should be a method. Why not `pyop3.device.get_current_device()`?

pyop3/buffer.py
```diff
@@ -247,8 +257,8 @@ def instruction_executor_cache_key(self, buffer_counter: Mapping[AbstractBuffer,
             type(self), self._constant, self._rank_equal, self._ordered, 
             self.dtype, buffer_counter[self])
 
-    def __init__(self, data: np.ndarray, sf: StarForest | None = None, *, name: str|None=None,prefix:str|None=None,constant:bool=False, rank_equal: bool = False, max_value: numbers.Number | None=None, ordered:bool=False):
-        data = data.flatten()
+    def __init__(self, data: np.ndarray | cp.ndarray, sf: StarForest | None = None, *, name: str|None=None,prefix:str|None=None,constant:bool=False, rank_equal: bool = False, max_value: numbers.Number | None=None, ordered:bool=False):
```
connorjward: I was imagining that we would also allow `data` to be `None` and thus always lazily allocate it.

pyop3/buffer.py
```diff
@@ -258,16 +268,25 @@ def __init__(self, data: np.ndarray, sf: StarForest | None = None, *, name: str|
         if rank_equal and not constant:
             raise ValueError
 
-        if constant:
-            data.flags.writeable = False
```
connorjward, line 262: To preserve this for numpy why not allow `ctx.asarray(...)` to take a `constant` kwarg

connorjward: Which we can then disregard for CUDA

connorjward: Or we can just remove this. It's not essential, just a nice extra bit of safety.

pyop3/buffer.py
```diff
@@ -305,19 +324,24 @@ def dtype(self) -> np.dtype:
...
     def duplicate(self, *, copy: bool = False) -> ArrayBuffer:
         # make sure that there are no pending transfers before we copy
         self.assemble()
         name = f"{self.name}_copy"
         if copy:
-            data = self._lazy_data.copy()
+            data = {obj: arr.copy() for obj, arr in self._lazy_data.items()}
         else:
-            data = np.zeros_like(self._lazy_data)
+            data = {obj: obj.zeros_like(arr) for obj, arr in self._lazy_data.items()}
```
connorjward: Here `data` can just be an empty dict.

pyop3/buffer.py
```diff
@@ -305,19 +324,24 @@ def dtype(self) -> np.dtype:
...
+        self._state[ctx] += 1
+        self._last_updated_device = ctx
 
     def duplicate(self, *, copy: bool = False) -> ArrayBuffer:
         # make sure that there are no pending transfers before we copy
         self.assemble()
         name = f"{self.name}_copy"
         if copy:
-            data = self._lazy_data.copy()
+            data = {obj: arr.copy() for obj, arr in self._lazy_data.items()}
```
connorjward: I'd only copy the most up-to-date data over.

pyop3/buffer.py
```diff
@@ -466,11 +490,18 @@ def leaves_valid(self) -> bool:
...
+
+        if not self._is_data_available(ctx) or not self._is_data_synced(ctx):
+            self.sync_devices(ctx)
+        
+        # NOTE: If data is None, set to zeros? 
+        # if self._lazy_data is None:
+        #    self._lazy_data = np.zeros(self.shape, dtype=self.dtype)
+
         if self.name == "array_247_buffer":
             breakpoint()
```
connorjward: my old debugging code, please kill this

pyop3/utils.py
```diff
@@ -331,12 +338,22 @@ def map_when(func, when_func, iterable):
...
+
+@readonly.register
+def _(array: np.ndarray) -> np.ndarray:
     """Return a readonly view of a numpy array."""
     view = array.view()
     view.setflags(write=False)
     return view
 
+@readonly.register
+def _(array: cp.ndarray) -> cp.ndarray:
```
connorjward: This will fail if cupy isn't installed. Put this check inside the top-level `singledispatch` function

pyop3/device.py
```diff
@@ -0,0 +1,97 @@
...
+    if not isinstance(dev, Device):
+        raise NotImplementedError
+
+    token = _current_device.set(dev)
+    try:
+        yield
+    finally:
+        _current_device.reset(token)
+
+# NOTE: Should this const variable be here?
```
connorjward: Seems sensible to me.

pyop3/device.py
```diff
@@ -0,0 +1,97 @@
...
+    try:
+        yield
+    finally:
+        _current_device.reset(token)
+
+# NOTE: Should this const variable be here? 
+HOST_DEVICE = CPU() 
+
+# NOTE: Use contextvars to act as a bridge between buffer and manager
+_current_device = contextvars.ContextVar("current_device", default=HOST_DEVICE)
```
connorjward: Nitpicking but I'd define this just above `def offloading` so it's easy to find when reading

pyop3/device.py
```diff
@@ -0,0 +1,97 @@
+# File to handle op3.device context manager
+from abc import ABCMeta, abstractmethod
+import contextlib
+import contextvars
+import warnings
+
+import numpy as np
+
+class Device(metaclass=ABCMeta):
+    _name: str
```
connorjward: Drop the underscore here and just call it `name`. I only have this implemented at points in pyop3 because I have interfaces to obey.

pyop3/device.py
```diff
@@ -0,0 +1,97 @@
...
+    _device_index: int | None
```
connorjward: Why is this attribute needed? It isn't used anywhere.

pyop3/device.py
```diff
@@ -0,0 +1,97 @@
...
+        
+    def __str__(self):
+        return self._name
+
+class CPU(Device):
+
+    def __init__(self, device_index: int | None = None):
+        super().__init__()
+        self._name = "cpu"
+        self._registered_arrays = set()
```
connorjward:
```suggestion
```

pyop3/device.py
```diff
@@ -0,0 +1,97 @@
...
+
+    def zeros_like(self, arr):
+        return np.zeros_like(arr)
+
+class CUDAGPU(Device):
+    
+    def __init__(self, device_index: int | None = None):
+        super().__init__()
+        self._name = "CudaGPU"
+        self._registered_arrays = set()
```
connorjward:
```suggestion
```

pyop3/device.py
```diff
@@ -0,0 +1,97 @@
...
+        return np.array(arr)
+
+    def zeros_like(self, arr):
+        return np.zeros_like(arr)
+
+class CUDAGPU(Device):
+    
+    def __init__(self, device_index: int | None = None):
+        super().__init__()
+        self._name = "CudaGPU"
```
connorjward: If this is a class level attribute it doesn't need to live on `self`.

pyop3/buffer.py
```diff
@@ -258,16 +268,25 @@ def __init__(self, data: np.ndarray, sf: StarForest | None = None, *, name: str|
...
+        self._lazy_data = data_mapping 
         self.sf = sf
         self._name = name
         self._constant = constant
         self._rank_equal = rank_equal
         self._max_value = max_value
         self._ordered = ordered
+        self._last_updated_device = ctx 
+
+        self._state = collections.defaultdict(int, [(ctx, 0)])
```
connorjward: Maybe a defaultdict is the way to go but I think the default value should be -1

connorjward (changes requested):
I've been strict but in general this is fantastic, thank you

pyop3/buffer.py
SamSJackson: Yes, I had something similar at first - attached it as classmethod to Device. 

Suggested module function is better. Amended this.

pyop3/buffer.py
SamSJackson: Pretty easy to introduce into Device logic and nicer logic in `buffer.py` so I will use this suggestion.

pyop3/buffer.py
SamSJackson: Alright, I will put this. Seems clear and I don't think that users should expect something else.

pyop3/buffer.py
SamSJackson: This makes sense. Will go for that.

pyop3/buffer.py
SamSJackson: Conflicts with dunder record_init method. Method wants to ensure that size attribute of values are identical between "duplicated" objects but empty dict has no values to compare size attributes.

Compromise work-around is only generate zeros-like array for current device?

i.e.
`data = { curr_dev: curr_dev.zeros_like(self._lazy_data[curr_dev]) }`

pyop3/buffer.py
SamSJackson: Yep, that's better. Switching to a property that returns device with highest state value.

pyop3/device.py
SamSJackson: Intention is to allow for user to select the device of their GPU - in case they have multiple GPUs.

As we are not at that point, I will remove property. Implementation would be very straightforward if needed in future.

pyop3/buffer.py
SamSJackson: Don't think I understand this. 

Are you thinking:
- `def __init__(self, data: np.ndarray | None, ...)`?

Functionally, at least, we do actually allow data to be any type and lazily convert with respect to the current device (beyond some methods like `flatten` that can be changed).

connorjward: Actually scratch this. I think creating an array buffer with no underlying array isn't a well defined operation.

pyop3/buffer.py
connorjward: Yeah happy with this. As above I don't think it makes much sense to have an array buffer without data.

connorjward:
For any trivial changes please go ahead and resolve them. Saves me cross checking things.

SamSJackson:
Update on progress:

Unit test cases have been introduced - also uncovered some bugs.

Bug: AxisTree uses cached properties. If cached properties are not available whilst in GPU, this will cause a Segfault.
Generation of cached properties requires `compile` strategy for arrays, not yet implemented for devices.

Fix:
The fix is to force some portions to take place on host until `compile` strategy is available.
The following functions have been wrapped in `on_host` decorators:
- `firedrake/functionspaceimpl.py::make_dat`
- `pyop3/tree/axis_tree/tree.py::_buffer_indices`

This means that if properties are not cached, assign takes place on host. If available, functions as normal.
As small side-effect, `buffer.py::duplicate` has a manual sync because current device would, correctly, suggest GPU but it was temporarily forced through host. Again, removable once `compile` works.

Also introduced a error catch in `insn/exec.py` to avoid the segfault taking place and enabling easier traceback.

Otherwise, I am quite happy with things.

pyop3/buffer.py
```diff
@@ -56,8 +61,10 @@ def wrapper(self, *args, **kwargs):
 def record_modified(func):
     def wrapper(self, *args, **kwargs):
         assert not self.constant
-        self.inc_state()
-        return func(self, *args, **kwargs)
+        try:
+            return func(self, *args, **kwargs)
+        finally:
+            self.inc_state()
```
connorjward: Why this change? Seems like it might be a good idea but want to make it's fully thought out.

pyop3/buffer.py
```diff
@@ -247,8 +250,11 @@ def instruction_executor_cache_key(self, buffer_counter: Mapping[AbstractBuffer,
             type(self), self._constant, self._rank_equal, self._ordered, 
             self.dtype, buffer_counter[self])
 
-    def __init__(self, data: np.ndarray, sf: StarForest | None = None, *, name: str|None=None,prefix:str|None=None,constant:bool=False, rank_equal: bool = False, max_value: numbers.Number | None=None, ordered:bool=False):
+    def __init__(self, data: np.ndarray | cp.ndarray | None, sf: StarForest | None = None, *, name: str|None=None,prefix:str|None=None,constant:bool=False, rank_equal: bool = False, max_value: numbers.Number | None=None, ordered:bool=False):
```
connorjward: Please do the sensible thing I should have done ages ago and split this onto multiple lines

pyop3/buffer.py
```diff
@@ -583,7 +603,18 @@ def localize(self) -> ArrayBuffer:
...
+        last_updated_device = self._last_updated_device
+
+        self._lazy_data[current_device] = current_device.asarray(self._lazy_data[last_updated_device], constant=self.constant)
+        self._state[current_device] = self._state[last_updated_device]
+
+    def _is_data_available(self, device: Device) -> bool:
+        return device in self._lazy_data
 
+    def _is_data_synced(self, device: Device) -> bool:
+        return self.state[device] == max(self.state.values())
```
connorjward: I'd just put `is_data_available` and `is_data_synced` into the same check

connorjward: I'd just put `is_data_available` and `is_data_synced` into the same check

pyop3/buffer.py
```diff
@@ -583,7 +603,18 @@ def localize(self) -> ArrayBuffer:
     @cached_property
     def _localized(self) -> ArrayBuffer:
         return self.__record_init__(sf=None)
+    
+    def sync_devices(self, current_device: Device):
```
connorjward: does `current_device` need to be an argument? It's a global variable

connorjward: does `current_device` need to be an argument? It's a global variable

pyop3/device.py
```diff
@@ -0,0 +1,99 @@
...
+        
+    def __str__(self):
+        return self.name
+
+class CPU(Device):
+    name = "CPU"
+
+    def __init__(self):
+        super().__init__()
+
```
connorjward:
```suggestion
```

connorjward:
```suggestion
```

connorjward:
```suggestion
```

pyop3/device.py
```diff
@@ -0,0 +1,99 @@
...
+import warnings
+
+import numpy as np
+
+class Device(metaclass=ABCMeta):
+    name: str
+
+    def __init__(self):
+        pass
+
```
connorjward:
```suggestion
```

pyop3/device.py
```diff
@@ -0,0 +1,99 @@
...
+class CPU(Device):
+    name = "CPU"
+
+    def __init__(self):
+        super().__init__()
+
+    def asarray(self, arr, *, constant=False):
+        # NOTE: Better logic needed if we switch from just NumPy/CuPy
+        output = arr
+        if not isinstance(arr, np.ndarray):
```
connorjward: This should be `isinstance(arr, cp.array)` or similar. This logic won't scale to other GPU backends.

pyop3/device.py
```diff
@@ -0,0 +1,99 @@
...
+
+class CPU(Device):
+    name = "CPU"
+
+    def __init__(self):
+        super().__init__()
+
+    def asarray(self, arr, *, constant=False):
+        # NOTE: Better logic needed if we switch from just NumPy/CuPy
+        output = arr
```
connorjward: Redundant assignment

pyop3/device.py
```diff
@@ -0,0 +1,99 @@
...
+
+        try:
+            import cupy as cp
+            assert cp.is_available() 
+        except:
+            # TODO: Raise No GPU exception
+            raise NotImplementedError 
+
+    def asarray(self, arr, *, constant=False):
+        import cupy as cp
```
connorjward, line 77: To avoid a bunch of imports maybe have
```py
@property
def cp(self):
  import cupy as cp
  return cp
```
and then use `self.cp.zeros_like` (or `self.cupy`)

pyop3/device.py
```diff
@@ -0,0 +1,99 @@
...
+    
+    def zeros_like(self, arr):
+        import cupy as cp
+        return cp.zeros_like(arr)
+
+HOST_DEVICE = CPU() 
+_current_device = contextvars.ContextVar("current_device", default=HOST_DEVICE)
+
+@contextlib.contextmanager
+def offloading(dev: Device):
```
connorjward, line 96: Can you add docstrings to some of these things? They don't have to be comprehensive, but 5 minutes spent now will help later on.

pyop3/device.py
```diff
@@ -0,0 +1,99 @@
...
+        _current_device.reset(token)
+
+def on_host(func):
+
+    def wrapper(*args, **kwargs):
+        token = _current_device.set(HOST_DEVICE)
+        try:
+            return func(*args, **kwargs)
+        finally:
+            _current_device.reset(token)
```
connorjward:
```suggestion
        with offloading(HOST_DEVICE):
            return func(*args, **kwargs)
```

connorjward:
```suggestion
        with offloading(HOST_DEVICE):
            return func(*args, **kwargs)
```

pyop3/utils.py
```diff
@@ -23,9 +23,16 @@
...
 from pyop3.constants import PYOP3_DECIDE, _nothing
 from pyop3.dtypes import DTypeT, IntType
-from pyop3.exceptions import CommMismatchException, CommNotFoundException, Pyop3Exception, UnhashableObjectException
+from pyop3.exceptions import CommMismatchException, CommNotFoundException, Pyop3Exception, UnhashableObjectException, UnsupportedArrayException
 from pyop3.mpi import collective
 
+ndarray_types = [np.ndarray,]
+try: 
+    import cupy as cp
+    ndarray_types = [np.ndarray, cp.ndarray]
```
connorjward:
```suggestion
    ndarray_types.append(cp.ndarray)
```

pyop3/utils.py
```diff
@@ -576,8 +582,12 @@ def pretty_type(obj: Any) -> str:
 
 
 def safe_equals(a, b, /) -> bool:
-    if any(isinstance(x, np.ndarray) for x in [a, b]):
+    if any(isinstance(x, tuple(ndarray_types)) for x in [a, b]):
         return (a == b).all()
+    if any(isinstance(x, dict) for x in [a, b]):
```
connorjward: Use `collections.abc.Mapping` instead of `dict`

tests/pyop3/unit/test_gpu_context.py
```diff
@@ -0,0 +1,125 @@
+import pytest
+import numpy as np
+
+try:
+    import cupy as cp
+except ImportError as err:
+    pytest.exit("CuPy not available, skipping GPU tests...")
```
connorjward:
```suggestion
except ImportError:
    pytest.skip(allow_module_level=True, reason="CuPy not available, skipping GPU tests...")
```

tests/pyop3/unit/test_gpu_context.py
```diff
@@ -0,0 +1,125 @@
...
+STATE_NOT_CREATED = -1
+STATE_UNTOUCHED = 0
+STATE_MODIFIED = 1
+
+@pytest.fixture()
+def mesh():
+    return UnitSquareMesh(3, 3)
+
+@pytest.fixture()
+def FuncSpace(mesh):
```
connorjward:
```suggestion
def space(mesh):
```

connorjward:
```suggestion
def space(mesh):
```

pyop3_gpu_demo.py
connorjward, line 1: Don't want to merge this file

tests/pyop3/unit/test_gpu_context.py
```diff
@@ -0,0 +1,125 @@
...
+            k_dup_buffer = k.dat.buffer.duplicate()
+            assert type(k_dup_buffer) == type(k.dat.buffer)
+            assert not k_dup_buffer is k.dat.buffer
+
+    def test_duplicate_to_device(self, FuncSpace):
+        """ Buffer maintains device context when copied"""
+        with op3.offloading(CUDAGPU):
+            k = Function(FuncSpace)
+            k_dup_buffer = k.dat.buffer.duplicate()
+            assert isinstance(k_dup_buffer.data_ro, cp.ndarray)
```
connorjward, line 118:
```suggestion
            assert isinstance(k_dup_buffer.data_ro, cp.ndarray)
            assert k_dup_buffer._data is k.buffer._data
```
?

connorjward (changes requested):
Looks close now

connorjward (changes requested):
Looks close now

pyop3/buffer.py
SamSJackson: It was to solve a previous bug. Essentially the equivalent of a C loop doing ++state vs state++.

Before, if an array was generated on host and then the first action on GPU is to modify, the state is updated before the sync happens. Then the sync would happen and the state update would be forgotten.

Skipped this problem by making the state update happen afterwards.

pyop3/buffer.py
SamSJackson: Nope, it can be dropped.

pyop3/device.py
SamSJackson: This is a better choice, even if just for cleanliness.

Would this reduce actual number of imports? Unsure if property would only be evaluated once.

pyop3/buffer.py
connorjward: I wonder if this is telling us that this decorator is being applied in the wrong place. I think we want to have
```py
self._data = ...  # i.e. we modify the data on *exactly* this line
self.inc_state()
```
I.e. we want to have the `inc_state` happen at the exact point that we actually modify the data. If we decouple them then I think we're just going to get very confused.

connorjward: I wonder if this is telling us that this decorator is being applied in the wrong place. I think we want to have
```py
self._data = ...  # i.e. we modify the data on *exactly* this line
self.inc_state()
```
I.e. we want to have the `inc_state` happen at the exact point that we actually modify the data. If we decouple them then I think we're just going to get very confused.

pyop3/device.py
connorjward: This is just to make things cleaner. I'm pretty sure importing the same thing a second time is just a no-op. You could make it a `cached_property` if it's a concern.

pyop3/device.py
SamSJackson: This okay?

There is the option of try-catch module importing but scales badly. 
```
        module = type(arr).__module__

        if module == "numpy":
            output = np.array(arr)
        elif module == "cupy":
            import cupy as cp 
            output = cp.asnumpy(arr)
        else:
            raise TypeError(f"{module} not supported.")
```

connorjward: I prefer `isinstance` if at all possible. Why not:
```
try:
  import cupy as cp
except ImportError:
  cp = None

if cp and isinstance(arr, cp.array):
  ...
```
The import part can be put in another function if helpful.

tests/pyop3/unit/test_gpu_context.py
SamSJackson: This would be expected to fail as, without the copy flag, the expected duplicate is simply a copy of the buffer wrapped around a zeros-like array. 

I have added the assert to the test with duplicate(copy=True), passing as expected.

pyop3/device.py
SamSJackson: This looks good, will go for that. Implementation is now:
```
def asarray(self, arr, *, constant=False):
        try:
            import cupy as cp
        except ImportError:
            cp = None
        
        if cp and isinstance(arr, cp.ndarray):
            output = cp.asnumpy(arr)
        elif isinstance(arr, np.ndarray):
            output = np.array(arr)
            if constant:
                output.flags.writeable = False
        else:
            raise TypeError(f"{type(arr)} not supported.")

       return output
```

pyop3/buffer.py
SamSJackson: I think it would be hard to be more precise with a decorator pattern, where it can only happen at start or end.

The `record_modified` wraps around only `data_wo` and `data_rw`. AIUI, we are not explicitly modifying the data ourselves, we serve a modifiable object to the user. As such, it would be hard to increment after an exact line - as we do not have one.

Although, current implementation is almost exactly as you want, just hidden in the wrapper. 
User calls `data_rw`/`data_wo` -> `self._data` property is called and synced -> `record_modified` wrapper updates state -> returns `self._data`

connorjward: I think that it is somehow spooky action at a distance. What do you think of the following instead:
1. Get rid of this decorator
1. Replace `._data` with a method `.get_array(self, intent: Literal["ro", "wo", "rw"])`

The implementation of `get_array()` would be something like:
```py
    def get_array(self, intent):
        curr_dev = get_current_device() 

        if not self._is_data_available_and_synced(curr_dev):
            self.sync_devices()

        if intent in {"wo", "rw"}:
            self.inc_state()
        
        return self._lazy_data[curr_dev]
```

SamSJackson: This works well, I think getting rid of decorator is necessary.

Do you want me to deprecate `._data` or should we be fine if it is internal?

connorjward: We can just delete `_data`. I think I use it in a few places to get access to the numpy array without tweaking the state (`pyop3/insn/exec.py` in particular) but I think you can just directly replace it with `get_array`, the intent should be available.

SamSJackson: Actually `._data` is used quite a lot internally in `buffer`, where intent is not explicitly available.

In properties like `size` or methods like `reduce_leaves_to_roots_begin`. 
The `_data` calls can be replaced by `get_array("ro")` but don't think we want hanging literals like that.

Can't just use `.data_ro` as it is recursive through `reduce_leaves_to_roots`. We could default get_array() for intent as "ro"?

connorjward: I think the right thing to do is probably replace them with `self._lazy_data[current_device]`. If we are doing things internal to the class I think we are allowed to work directly on the data - the state tracking is for managing how people interact with the `ArrayBuffer`.

But defaulting `get_array` to `ro` also seems sensible.

SamSJackson: Okay, just pushed changes for this. Requires a review as quite a few lines changed.

connorjward (approved):

PR #5079, opened 2026-05-05
AssembledPC: fix sub_mat_type

pbrubeck:
# Description

### DIFF
--- a/firedrake/solving_utils.py
+++ b/firedrake/solving_utils.py
@@ -282,15 +282,18 @@ class _SNESContext(object):
         problem = problem or self._problem
         mat_type = mat_type or self.mat_type
         pmat_type = pmat_type or self.pmat_type
-        for k, v in tuple(kwargs.items()):
-            if v is None:
-                kwargs.pop(k)
-        kwargs.setdefault("sub_mat_type", self.sub_mat_type)
-        kwargs.setdefault("sub_pmat_type", self.sub_pmat_type)
-        kwargs.setdefault("appctx", self.appctx)
-        kwargs.setdefault("options_prefix", self.options_prefix)
-        kwargs.setdefault("transfer_manager", self.transfer_manager)
-        kwargs.setdefault("pre_apply_bcs", self.pre_apply_bcs)
+
+        default_options = {
+            "sub_mat_type": self.sub_mat_type,
+            "sub_pmat_type": self.sub_pmat_type,
+            "appctx": self.appctx,
+            "options_prefix": self.options_prefix,
+            "transfer_manager": self.transfer_manager,
+            "pre_apply_bcs": self.pre_apply_bcs,
+        }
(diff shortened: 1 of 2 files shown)


firedrake/solving_utils.py
```diff
@@ -279,9 +279,12 @@ def __init__(self, problem,
 
     def reconstruct(self, problem=None, mat_type=None, pmat_type=None, **kwargs):
         """Reconstruct this _SNESContext instance with new arguments."""
-        problem = problem or self.problem
+        problem = problem or self._problem
         mat_type = mat_type or self.mat_type
         pmat_type = pmat_type or self.pmat_type
+        for k, v in tuple(kwargs.items()):
+            if v is None:
+                kwargs.pop(k)
```
connorjward: This is really obscure. Why are those options disallowed?

connorjward (changes requested):

firedrake/solving_utils.py
pbrubeck: If we pass `sub_mat_type=None` we would like to grab the one from the parent `_SNESContext`. Basically this enables
```
sub_mat_type = kwargs.get("sub_mat_type") or self.sub_mat_type
```
but more generically

pbrubeck: The line below `kwargs.setdefault("sub_mat_type", self.sub_mat_type)` is a no-op if `submat_type=None` is already in the dictionary.

pbrubeck: Should the lines below be changed to something like
```
kwargs["sub_mat_type"] = kwargs.get("sub_mat_type") or self.sub_mat_type
```
?

connorjward: Yeah I think that's much better

pbrubeck: That will potentially cause conflicts with empty dictionaries or empty strings

connorjward: How so?

pbrubeck: We need
```
def set_default_kwarg(kwargs, key, default):
    if kwargs.get(key) is None:
        kwargs[key] = default

set_default_kwarg(kwargs, "sub_mat_type", self.sub_mat_type)
...
```

pbrubeck: > How so?

`reconstruct` should keep the original arguments if they are not passed as kwargs or if they are set to None in the kwargs. Sometimes one might need to pass `appctx={}` or `options_prefix=""`, which are falsy values that we should not treat as `None`.

connorjward: Explicitly filtering `None`s like that is also fine. I just don't like the premature filtering of all kwarg options.

pbrubeck: @connorjward Are we happy with the current state of this?

connorjward (approved):

PR #5081, opened 2026-05-06
Adjoint + submesh

KarsKnook:
To make the adjoint compatible with `Submesh`, both the mesh and submesh are added as dependencies to an adjoint block. This is accomplished by replacing `ufl_domain` and `as_domain` with `extract_domains` in the appropriate files.

Perhaps this implementation is too naive and not all meshes need to be added as dependencies?
Does this account for all adjoint + submesh cases?

**To do:**
- answer questions above
- write tests

**Disclaimer:**
ChatGPT came up with the initial implementation which was restructured by me

### DIFF
--- a/firedrake/adjoint_utils/blocks/assembly.py
+++ b/firedrake/adjoint_utils/blocks/assembly.py
@@ -19,7 +19,7 @@ class AssembleBlock(Block):
         if meshes and not isinstance(self.form, ufl.Interpolate):
             # Interpolation differentiation wrt spatial coordinates is currently not supported.
             for mesh in meshes:  # add all meshes as dependency
-                self.add_dependency(mesh, no_duplicates=True)
+                self.add_dependency(mesh)
 
         for c in self.form.coefficients():
             self.add_dependency(c, no_duplicates=True)
--- a/firedrake/adjoint_utils/blocks/assembly.py
+++ b/firedrake/adjoint_utils/blocks/assembly.py
@@ -4,7 +4,6 @@ from ufl.domain import extract_domains
 from ufl.formatting.ufl2unicode import ufl2unicode
 from pyadjoint import Block, AdjFloat, create_overloaded_object
 from firedrake.adjoint_utils.checkpointing import maybe_disk_checkpoint
-from .block_utils import isconstant
 
 
 class AssembleBlock(Block):
@@ -104,9 +103,7 @@ class AssembleBlock(Block):
 
         arity_form = len(form.arguments())
 
-        if isconstant(c):
-            space = c.function_space()
-        elif isinstance(c, (firedrake.Function, firedrake.Cofunction)):
+        if isinstance(c, (firedrake.Function, firedrake.Cofunction)):
             space = c.function_space()
         elif isinstance(c, firedrake.MeshGeometry):
             c_rep = firedrake.SpatialCoordinate(c_rep)
@@ -161,9 +158,7 @@ class AssembleBlock(Block):
         c1 = block_variable.output
         c1_rep = block_variable.saved_output
 
-        if isconstant(c1):
-            space = c1.function_space()
-        elif isinstance(c1, (firedrake.Function, firedrake.Cofunction)):
+        if isinstance(c1, (firedrake.Function, firedrake.Cofunction)):
             space = c1.function_space()
         elif isinstance(c1, firedrake.MeshGeometry):
             c1_rep = firedrake.SpatialCoordinate(c1)
--- a/firedrake/adjoint_utils/blocks/assembly.py
+++ b/firedrake/adjoint_utils/blocks/assembly.py
@@ -19,7 +19,7 @@ class AssembleBlock(Block):
         if meshes and not isinstance(self.form, ufl.Interpolate):
             # Interpolation differentiation wrt spatial coordinates is currently not supported.
             for mesh in meshes:  # add all meshes as dependency
-                self.add_dependency(mesh)
+                self.add_dependency(mesh, no_duplicates=True)
 
         for c in self.form.coefficients():
             self.add_dependency(c, no_duplicates=True)
--- a/firedrake/adjoint_utils/blocks/assembly.py
+++ b/firedrake/adjoint_utils/blocks/assembly.py
@@ -1,24 +1,38 @@
 import ufl
 import firedrake
-from ufl.domain import as_domain
+from ufl.domain import extract_domains
 from ufl.formatting.ufl2unicode import ufl2unicode
 from pyadjoint import Block, AdjFloat, create_overloaded_object
 from firedrake.adjoint_utils.checkpointing import maybe_disk_checkpoint
 from .block_utils import isconstant
 
 
+def _extract_meshes(expr):
+    try:
+        return tuple(extract_domains(expr))
+    except AttributeError:
+        return ()
+
+
+def _extract_unique_mesh_for_constant(expr):
+    meshes = _extract_meshes(expr)
+    if len(meshes) != 1:
+        raise NotImplementedError(
+            "Differentiating an assembled multi-domain/Submesh form with respect "
+            "to a Constant is not implemented yet: cannot choose a unique mesh "
+            "for the Constant adjoint function space."
+        )
+    return meshes[0]
+
+
 class AssembleBlock(Block):
     def __init__(self, form, ad_block_tag=None):
         super(AssembleBlock, self).__init__(ad_block_tag=ad_block_tag)
         self.form = form
-        try:
-            mesh = as_domain(form)
-        except AttributeError:
-            mesh = None
-
-        if mesh and not isinstance(self.form, ufl.Interpolate):
+        if not isinstance(self.form, ufl.Interpolate):
             # Interpolation differentiation wrt spatial coordinates is currently not supported.
-            self.add_dependency(mesh)
+            for mesh in _extract_meshes(form):
+                self.add_dependency(mesh, no_duplicates=True)
 
         for c in self.form.coefficients():
             self.add_dependency(c, no_duplicates=True)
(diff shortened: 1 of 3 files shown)


firedrake/adjoint_utils/blocks/assembly.py
```diff
@@ -11,14 +11,15 @@ class AssembleBlock(Block):
...
         except AttributeError:
-            mesh = None
+            meshes = None
 
-        if mesh and not isinstance(self.form, ufl.Interpolate):
+        if meshes and not isinstance(self.form, ufl.Interpolate):
             # Interpolation differentiation wrt spatial coordinates is currently not supported.
-            self.add_dependency(mesh)
+            for mesh in meshes:  # add all meshes as dependency
+                self.add_dependency(mesh, no_duplicates=True)
```
connorjward, line 21: Since we weren't passing `no_duplicates` before it's probably wise to keep things the same as before.

firedrake/adjoint_utils/blocks/solving.py
```diff
@@ -243,9 +253,8 @@ def evaluate_adj_component(self, inputs, adj_inputs, block_variable, idx,
         c_rep = block_variable.saved_output
 
         if isconstant(c):
-            mesh = F_form.ufl_domain()
             trial_function = firedrake.TrialFunction(
-                c._ad_function_space(mesh)
+                c.function_space()
```
connorjward: AFAICT this is equivalent to `_ad_function_space` so this is likely fine. My guess is `_ad_function_space` is a holdover from previous FEniCS/Firedrake compat.

firedrake/adjoint_utils/blocks/solving.py
```diff
@@ -759,7 +767,7 @@ def evaluate_adj_component(self, inputs, adj_inputs, block_variable, idx,
         if isinstance(c, (firedrake.Function, firedrake.Cofunction)):
             trial_function = firedrake.TrialFunction(c.function_space())
         elif isinstance(c, firedrake.Constant):
-            mesh = F_form.ufl_domain()
+            mesh = extract_domains(F_form)[0]  # I don't like this
```
connorjward: Yeah this isn't good. Ideally please figure out what is going on here and do something that addresses the multi-mesh case.

connorjward (changes requested):
Please add some tests

connorjward (changes requested):
Please add some tests

firedrake/adjoint_utils/blocks/assembly.py
```diff
@@ -104,8 +105,7 @@ def evaluate_adj_component(self, inputs, adj_inputs, block_variable, idx,
         arity_form = len(form.arguments())
 
         if isconstant(c):
-            mesh = as_domain(self.form)
-            space = c._ad_function_space(mesh)
+            space = c.function_space()
```
JHopeCollins: How come this needs to change? `c.function_space()` returns a Firedrake type but `c._ad_function_space` returns the `ufl_function_space`.

firedrake/adjoint_utils/blocks/solving.py
```diff
@@ -74,8 +75,17 @@ def __init__(self, lhs, rhs, func, bcs, *args, **kwargs):
         for bc in self.bcs:
             self.add_dependency(bc, no_duplicates=True)
 
-        mesh = self.lhs.ufl_domain()
-        self.add_dependency(mesh)
+        try:  # add all meshes as dependency
+            for mesh in tuple(extract_domains(self.lhs)):
```
JHopeCollins: I think that the `tuple` cast is superfluous here.
```suggestion
            for mesh in extract_domains(self.lhs):
```

firedrake/adjoint_utils/blocks/assembly.py
connorjward: I wonder if this is a holder from previous Firedrake/FEniCS compatibility. But I have no idea really.

JHopeCollins: If it is then we should address that in a separate PR

connorjward: In which case someone should open an issue. This is a non-core dev submission and saying "leave that for another PR" makes it very likely to never be done.

JHopeCollins: There's two questions here:

1. Does it need changing for the fix in this PR to work? @KarsKnook does it work if you don't change how you get the function space? If not then you can revert back to the original.
2. Is this a holdover that we could update? I don't know either, but I'm just saying that unless its essential for this PR then this isn't the place to start changing it, lets keep this one as simple as possible.

KarsKnook: I think we can keep using `_ad_function_space(mesh)` if we fetch the mesh from `c` and not `self.form`. I would still use `as_domain()` and (deprecated) `.ufl_domain()` for that because this is not the PR to fix that. What do you think?

firedrake/adjoint_utils/blocks/solving.py
```diff
@@ -74,8 +75,17 @@ def __init__(self, lhs, rhs, func, bcs, *args, **kwargs):
         for bc in self.bcs:
             self.add_dependency(bc, no_duplicates=True)
 
-        mesh = self.lhs.ufl_domain()
-        self.add_dependency(mesh)
+        try:  # add all meshes as dependency
+            for mesh in extract_domains(self.lhs):
+                self.add_dependency(mesh, no_duplicates=True)
```
connorjward, line 79: Same comment about `no_duplicates` as above.

firedrake/adjoint_utils/blocks/solving.py
```diff
@@ -74,8 +75,17 @@ def __init__(self, lhs, rhs, func, bcs, *args, **kwargs):
...
+        except AttributeError:
+            pass
+
+        if isinstance(self.rhs, (ufl.Form, ufl.Cofunction)):
```
connorjward:
```suggestion
        if self.linear:
```

connorjward (dismissed):
I think this is basically there subject to the final discussion about `function_space()`

firedrake/adjoint_utils/blocks/solving.py
KarsKnook: Some test fails if you don't include no_duplicates. I don't understand why we wouldn't want to use no_duplicates for meshes if we use it for each other dependency

connorjward: Just trying to minimise the changes in this PR. If it's necessary for tests to pass then I'm happy with the change.

tests/firedrake/adjoint/test_solving.py
```diff
@@ -314,6 +314,74 @@ def test_two_nonlinear_solves():
...
+    bc2 = DirichletBC(V.sub(1), 0, "on_boundary")
+    bcs = [bc1, bc2]
+
+    solve(a - L == 0, u, bcs)
+
+    J = assemble(u1**2*dx(mesh1) + u2**2*dx(mesh2))
+    rf = ReducedFunctional(J, Control(f))
+    df = rg.uniform(V)
+
+    assert taylor_test(rf, f, df) > 1.95
```
JHopeCollins: Please can you change this to `taylor_to_dict` so that we are also checking the TLM and Hessian model? You can see an example here:
https://github.com/firedrakeproject/firedrake/blob/ee52b418a55be7606a9b56b063c234382112ddb4/tests/firedrake/adjoint/test_covariance_operator.py#L43-L47

Note that you need greater than quadratic nonlinearity for the Hessian to not be exact. You can just make the norm calculation for `J` quartic rather than quadratic.

tests/firedrake/adjoint/test_solving.py
```diff
@@ -314,6 +314,74 @@ def test_two_nonlinear_solves():
...
+
+    bcs = [DirichletBC(V.sub(0), 0, "on_boundary")]
+
+    solve(a - L == 0, u, bcs)
+
+    J = assemble(u2**2*dx_sub)
+    rf = ReducedFunctional(J, Control(f))
+    df = rg.uniform(V1)
+
+    assert taylor_test(rf, f, df) > 1.95
```
JHopeCollins: Same here with `taylor_to_dict` please

firedrake/adjoint_utils/blocks/assembly.py
JHopeCollins: I think that this is probably fine and I am being too nitpicky. If keeping the `_ad_function_space` means using deprecated functionality, and your alternative works, then lets just keep what you have.

JHopeCollins (dismissed):
The implementation looks good to me, thanks for fixing this!

Just a small change needed to the tests so we check the TLM and Hessian models, then this is ready to go.

firedrake/adjoint_utils/blocks/solving.py
```diff
@@ -243,9 +253,8 @@ def evaluate_adj_component(self, inputs, adj_inputs, block_variable, idx,
         c_rep = block_variable.saved_output
 
         if isconstant(c):
-            mesh = F_form.ufl_domain()
             trial_function = firedrake.TrialFunction(
-                c._ad_function_space(mesh)
+                c.function_space()
             )
         elif isinstance(c, (firedrake.Function, firedrake.Cofunction)):
```
pbrubeck:
```suggestion
        if isinstance(c, (firedrake.Function, firedrake.Cofunction)):
````

File: firedrake/adjoint_utils/blocks/solving.py [hunk 11]
```diff
@@ -455,8 +464,7 @@ def evaluate_hessian_component(self, inputs, hessian_inputs, adj_inputs,
             return [tmp_bc]
 
         if isconstant(c_rep):
-            mesh = F_form.ufl_domain()
-            W = c._ad_function_space(mesh)
+            W = c_rep.function_space()
         elif isinstance(c, firedrake.MeshGeometry):
```
pbrubeck:
```suggestion
        if isinstance(c, firedrake.MeshGeometry):
```

firedrake/adjoint_utils/blocks/solving.py
pbrubeck:
```suggestion
        if isinstance(self.rhs, ufl.BaseForm):
```

KarsKnook:
Firedrake functions in the real space are now treated as functions without calling `isconstant`

JHopeCollins (approved):
LGTM, thanks @KarsKnook!

connorjward:
Sorry to be a pain... but shouldn't this have gone into `release`?

PR #5085, opened 2026-05-07
firedrake-configure options for Ubuntu 2026.04

JHopeCollins:
1. PETSc can't find the apt PTScotch package on it's own any more so we need to set the linking flags in the PETSc configure arguments.

2. Hypre apt package is now new enough for PETSc to use so we don't have to build from source.

3. Ubuntu26.04 uses gcc15 which seems to result in some problems in interpolation kernels for complicated elements, so we default to `-O2` for this compiler, and `-O1` on CI.

4. CUDA packages are not ready yet for Ubuntu26.04 so we don't provide an `Arch` for that combination yet.

5. The Ubuntu24.04 arches are now `CommunityArch` so they are still available to people but we will no longer test them.


francesco-ballarin:
Basically every package should be new enough for PETSc to use, we have started the migration to PETSc 3.25 in Debian unstable and we didn't need to update any package (except PETSc itself).

If there are some packages that PETSc complains about, please let me know. 
FYI, at some point in 2027 or 2028 you may need to go back to your manually built dependencies: Debian will keep updating PETSc and its dependencies, but we can't backport the updates to ubuntu.

scripts/firedrake-configure
```diff
@@ -680,6 +684,137 @@ OFFICIAL_ARCHS: Mapping[ArchKey, Arch] = {
...
+            "--with-bison",
+            "--with-fftw",
+            "--with-hdf5",
+            "--with-hwloc",
+            "--with-metis",
+            "--with-mumps",
+            "--with-netcdf",
+            "--with-pnetcdf",
+            "--with-ptscotch",
+            "--with-ptscotch-lib='-l:libptesmumps.so.7.0 -l:libptscotch.so.7.0 -libptscotcherr.so.7.0 -l:libesmumps.so.7.0 -l:libscotch.so.7.0 -l:libscotcherr.so.7.0'",
```
francesco-ballarin: hardcoding the 7.0 may not be necessary, we don't need it upstream

https://salsa.debian.org/science-team/petsc/-/blob/master/debian/rules?ref_type=heads#L194

.github/workflows/core.yml
```diff
@@ -90,7 +90,7 @@ jobs:
     runs-on: [self-hosted, Linux]
     container:
       # TODO: set to 'ubuntu:latest'
```
```suggestion
```

.github/workflows/core.yml
```diff
@@ -90,7 +90,7 @@ jobs:
...
-      image: ubuntu:noble
+      image: ubuntu:resolute
```
```suggestion
      image: ubuntu:latest
```

connorjward: Please also do in the other place in this file where we specify `ubuntu:noble`

scripts/firedrake-configure
```diff
@@ -680,6 +684,137 @@ OFFICIAL_ARCHS: Mapping[ArchKey, Arch] = {
         ],
     ),
 
+    # Ubuntu 26.04 x86
```
connorjward: Please move the Ubuntu 24.04 packages into the community archs section, since we no longer test them on CI. If you follow my suggestion for adding a `maintainer` field I think it should just be `"the Firedrake team"`

connorjward (changes requested):
Thanks!

Unfortunately looks like it is hitting errors. I wonder if the GitHub actions runner version of Ubuntu 26.04 has slightly different libraries than the Dockerhub one.

scripts/firedrake-configure
connorjward: This is such helpful information! Josh has been messaging me about his struggles to get PTScotch working properly.

connorjward:
Can you do the GPU change at the same time? Should be identical work

scripts/firedrake-configure
JHopeCollins: Very interesting. That looks like the flags that PETSc uses if you just pass `--with-ptscotch` like we used to. I'll have another go without the version numbers.

The flags I've got here work on the 26.04 docker container locally for me, I'll have to track down what the difference with ci is.

JHopeCollins:
> Basically every package should be new enough for PETSc to use, we have started the migration to PETSc 3.25 in Debian unstable and we didn't need to update any package (except PETSc itself).
>
> If there are some packages that PETSc complains about, please let me know. 
> FYI, at some point in 2027 or 2028 you may need to go back to your manually built dependencies: Debian will keep updating PETSc and its dependencies, but we can't backport the updates to ubuntu.

We would be very happy if this could be PETSc policy to support latest apt packages across the board. The only package that we need that didn't work from apt with Ubuntu2024.04 was hypre, the version was slightly too old for the latest PETSc.

We don't use the apt PETSc, so hopefully we won't need to start manually build things we can currently use

JHopeCollins:
> Can you do the GPU change at the same time? Should be identical work

I was going to do that once the regular ones are working

scripts/firedrake-configure
JHopeCollins: Done

scripts/firedrake-configure
```diff
@@ -390,32 +398,38 @@ OFFICIAL_ARCHS: Mapping[ArchKey, Arch] = {
...
-            "/usr/include/scotch",
+            "/usr/include/hypre",
             "/usr/include/superlu",
             "/usr/include/superlu-dist",
         ],
         library_dirs=[
             "/usr/lib/x86_64-linux-gnu/hdf5/openmpi",
         ],
+        extra_env_vars={
+            # Ignore mismatch in apt graphviz and pypi pygraphviz signatures
```
connorjward: This will go away after https://github.com/pygraphviz/pygraphviz/pull/573, can you link to the issue so we don't forget to undo it later?

scripts/firedrake-configure
```diff
@@ -843,7 +869,346 @@ OFFICIAL_ARCHS: Mapping[ArchKey, Arch] = {
 
 
 # 'Best effort' configurations that are not tested in CI
-COMMUNITY_ARCHS: Mapping[ArchKey, CommunityArch] = {}
+COMMUNITY_ARCHS: Mapping[ArchKey, CommunityArch] = {
+
+    # Ubuntu 24.04 x86
+
+    (OS.UBUNTU_2404_X86_64, ScalarType.DEFAULT, GPUPlatform.NO_GPU): CommunityArch(
```
connorjward, line 887: Can you double check that `CommunityArch` works as expected? And that it provides useful information (via `warn`) when used? E.g. the message might want to include the date of last modified and maybe we also want to have the maintainer email

scripts/firedrake-configure
connorjward: `extra_env_vars`?

connorjward, line 633: `extra_env_vars`?

connorjward (dismissed):

connorjward:
I wonder if the errors are due to unsafely compiling pygraphviz. Let's drop it as a dependency for now and see what happens. If it's not that then this could be a whole can of worms.

connorjward:
As far as I can tell the cursed combination of CUDA, Ubuntu 26.04 and GCC 15 is not yet compatible. I think we have to keep the GPU CI running on 24.04 for the moment. I will investigate the errors with the default and complex builds.

JHopeCollins (commented):
Looks good to me, thanks for taking this on and finishing it up!

connorjward (approved):

PR #5089, opened 2026-05-08
Fixes for multidomain form preprocessing

pbrubeck:
# Description
Applies more agressive index simplifications to ensure that quantities defined on other domains do not appear in integrals with single-domain `Measures`. 

This is an alternative solution to the issue addressed in https://github.com/firedrakeproject/Irksome/pull/225

> Suppose `u = (u0, u1) in V0 * V1` is the unknown for a mixed formulation where the function spaces `V0` and `V1` are defined on different meshes. At stage `i` we replace `u -> (A*k)[i]`, where `k[i] = (k[0][i], k[1][i])` is the component of the stage vector for stage i.
>
> For a single-domain integral like `inner(u[0], v[0]) * dx(mesh0)`, we run into an issue when we replace `u[0] -> (A*k)[i][0]` inside the integral as `k` is defined on both `mesh0` and `mesh1`.

This PR fixes this by removing component tensors for multidomain problems. This is an optional step in `compute_form_data` that we were skipping because it has a GEM equivalent that we were already using. 

The index simplification effectively introduces the replacement `u[0] -> (A*k[0])[i]`, removing `k[1]` from the integral with `dx(mesh0)`.

It is still unclear whether we should set `do_remove_component_tensors=True` as the default.

### DIFF
--- a/tsfc/ufl_utils.py
+++ b/tsfc/ufl_utils.py
@@ -52,6 +52,9 @@ def compute_form_data(form,
     kwargs overriden in the way TSFC needs it and is provided for
     other form compilers based on TSFC.
     """
+    # Multidomain problems require further index simplifications to ensure
+    # that unwanted quantities do not appear inside single-domain integrals.
+    do_remove_component_tensors = len(form.ufl_domains()) > 1
     fd = ufl_compute_form_data(
         form,
         do_apply_function_pullbacks=do_apply_function_pullbacks,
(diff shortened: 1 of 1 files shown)


connorjward:
> It is still unclear whether we should set do_remove_component_tensors=True as the default.

I have no idea I'm afraid.

Your fix here does seem a little hacky.

pbrubeck:
> > It is still unclear whether we should set do_remove_component_tensors=True as the default.
> 
> I have no idea I'm afraid.
> 
> Your fix here does seem a little hacky.

Some sort of index simplification must occur in `compute_form_data` in order to get a Form that passes `validate_domains`, even for the simple cases we have in Irksome.

We could set `do_remove_component_tensors=True` as default and the fix might seem less hacky. But I think we should not do it in the single domain case for performance reasons.

dham (approved):
Please unconditionally lower component tensor in main.

PR #5092, opened 2026-05-10
Fix _SNESContext.options_prefix

pbrubeck:
# Description

Adds options prefix in `_SNESContext.split` in order to support 
`-mat_type matfree -fieldsplit_0_mg_coarse_mat_type aij`


firedrake/solving_utils.py
```diff
@@ -443,10 +443,8 @@ def split(self, fields):
             new_problem = NLVP(F, subu, bcs=bcs, J=J, Jp=Jp, is_linear=problem.is_linear,
                                form_compiler_parameters=problem.form_compiler_parameters)
             new_problem._constant_jacobian = problem._constant_jacobian
-            splits.append(type(self)(new_problem, mat_type=self.mat_type, pmat_type=self.pmat_type,
-                                     appctx=self.appctx,
-                                     transfer_manager=self.transfer_manager,
-                                     pre_apply_bcs=self.pre_apply_bcs))
+            options_prefix = f"{self.options_prefix}fieldsplit_{field_num}_"
```
JHopeCollins: If the field has a single component and that function space component has a name then we need to use that instead because we do this: https://github.com/firedrakeproject/firedrake/blob/05b47bcf9484c70f451267990f4cbda4c34d7959/firedrake/dmhooks.py#L343

firedrake/solving_utils.py
```diff
@@ -443,10 +443,9 @@ def split(self, fields):
             new_problem = NLVP(F, subu, bcs=bcs, J=J, Jp=Jp, is_linear=problem.is_linear,
                                form_compiler_parameters=problem.form_compiler_parameters)
             new_problem._constant_jacobian = problem._constant_jacobian
-            splits.append(type(self)(new_problem, mat_type=self.mat_type, pmat_type=self.pmat_type,
-                                     appctx=self.appctx,
-                                     transfer_manager=self.transfer_manager,
-                                     pre_apply_bcs=self.pre_apply_bcs))
+            field_prefix = f"fieldsplit_{V.name or field_num}_"
```
pbrubeck: `V.name` is now propagated if the field has a single component.

tests/firedrake/multigrid/test_options_prefix.py
```diff
@@ -0,0 +1,57 @@
...
+
+    x = SpatialCoordinate(mesh)
+    z_exact = as_vector([x[0], x[0]**2])
+    z = Function(W)
+    w = TrialFunction(W)
+    v = TestFunction(W)
+    a = inner(grad(w), grad(v)) * dx
+    L = a(v, z_exact)
+    bcs = [DirichletBC(W.sub(i), z_exact[i], "on_boundary") for i in range(len(W))]
+    solve(a == L, z, bcs=bcs, solver_parameters=sp)
```
JHopeCollins: Does this test actually check if the options were used properly? Wouldn't it pass if everything was ignored and it just used LU?

firedrake/mg/ufl_utils.py
```diff
@@ -282,47 +282,45 @@ def coarsen_snescontext(context, self, coefficient_mapping=None):
...
+    while parent_context._fine:
+        parent_context = parent_context._fine
+    parent_prefix = parent_context.options_prefix
+    opts = PETSc.Options(parent_prefix)
+    if opts.getString("snes_type", "") == "fas":
+        solver_prefix = "fas_"
+    else:
+        solver_prefix = "mg_"
     _, level = utils.get_level(problem.u_restrict.function_space().mesh())
     if level == 0:
```
JHopeCollins, line 296: We should also test if we're on the finest level and check the `mg_fine_` prefix. That takes precedence over `mg_levels_` but I don't know if it takes precedence over `mg_levels_N_`. We should do whatever PETSc does.

JHopeCollins (dismissed):

firedrake/mg/ufl_utils.py
pbrubeck: We cannot change the finest level at this point, this only applies to levels that are coarser.

tests/firedrake/multigrid/test_options_prefix.py
pbrubeck: Done

tests/firedrake/multigrid/test_options_prefix.py
```diff
@@ -0,0 +1,69 @@
...
+    L = a(v, z_exact)
+    bcs = [DirichletBC(W.sub(i), z_exact[i], "on_boundary") for i in range(len(W))]
+    problem = LinearVariationalProblem(a, L, z, bcs=bcs)
+    solver = LinearVariationalSolver(problem, solver_parameters=sp)
+    solver.solve()
+    assert errornorm(z_exact, z) / norm(z_exact) < 1E-12
+
+    assert solver.snes.ksp.pc.getOperators()[0].getType() == "python"
+    fsplit = solver.snes.ksp.pc.getFieldSplitSubKSP()
+    assert len(fsplit) == len(V)
```
pbrubeck:
```suggestion
    assert len(fsplit) == len(W)
```

firedrake/solving_utils.py
JHopeCollins: What is enforcing it only using `V.name` if `V` has a single component?

tests/firedrake/multigrid/test_options_prefix.py
JHopeCollins: Great, thanks

JHopeCollins (commented):
One comment about using the `FunctionSpace` name for components. Once we've sorted that this is good to go.

pbrubeck:
> One comment about using the `FunctionSpace` name for components. Once we've sorted that this is good to go.

> What is enforcing it only using `V.name` if `V` has a single component?

The name of each single field subspace now gets propagated when we split the problem (this wasn't the case before). But if we split into a multi-field subspace we need to create a `MixedFunctionSpace`, and we never set the `name` kwarg for that case

JHopeCollins:
> > One comment about using the `FunctionSpace` name for components. Once we've sorted that this is good to go.
> 
> > What is enforcing it only using `V.name` if `V` has a single component?
> 
> The name of each single field subspace now gets propagated when we split the problem (this wasn't the case before). But if we split into a multi-field subspace we need to create a `MixedFunctionSpace`, and we never set the `name` kwarg for that case

But if we don't manually set the name of a `MixedFunctionSpace` then it gets set according to the names of the component spaces here:
https://github.com/firedrakeproject/firedrake/blob/ee52b418a55be7606a9b56b063c234382112ddb4/firedrake/functionspaceimpl.py#L1055

firedrake/solving_utils.py
```suggestion
            name = V.name if len(V) == 1 else None
            field_prefix = f"fieldsplit_{name or field_num}_"
```
.

JHopeCollins (approved):
LGTM

PR #5093, opened 2026-05-11
Test that control checkpoints survive recompute (fixes #5082)

### LINKED ISSUE #5082: BUG: incorrect adjoint when using checkpoint scheduler
Consider the follow code:
```
from firedrake import *
from checkpoint_schedules import SingleMemoryStorageSchedule
continue_annotation()
tape = get_working_tape()

schedule = SingleMemoryStorageSchedule()

tape.enable_checkpointing(schedule)

mesh = UnitSquareMesh(1,1)

m = Function(V).assign(1.0)

sumf = Function(V)
u = Function(V)
tst = TestFunction(V)
F = tst*u*dx - tst*m*m*dx

problem = NonlinearVariationalProblem(F, u)
solver = NonlinearVariationalSolver(problem)

for step in tape.timestepper(iter(range(4))):
    solver.solve()
    sumf.assign(sumf + u)

J = assemble(sumf*dx)
rf = ReducedFunctional(J, Control(m))

m0 = Function(V).assign(2.0)
print(rf(m0))
print(rf.derivative(apply_riesz=True).dat.data)
```
The taped model computes $$J=4m^2$$ and thus the derivative should be $$\frac{dJ}{dm} = 8m$$ and therefore for `m=2` I expect the derivative to be 16. With the checkpointing schedule as above however, even though the forward (re)run with `m=2` correctly produces $$J=16$$, the derivative it computes is 8 - it seems during the adjoint evaluation it reuses the old control value not the current one. If I comment out the `enable_checkpointing` line, I do get the expected result.

sghelichkhani:
Companion to dolfin-adjoint/pyadjoint#257, which fixes the underlying bug in `CheckpointManager` where direct writes to `BlockVariable._checkpoint` bypassed the `is_control` guard and wiped the user-supplied control value during forward replay. With that pyadjoint fix in place the control's checkpoint now legitimately survives across replays, which makes two changes necessary here.

The first is that `_check_forward`, `_check_recompute` and `_check_reverse` in `tests/firedrake/adjoint/test_burgers_newton.py` were asserting that every block-variable checkpoint must be `None` after replay. That assertion codified the buggy behaviour — under the fix, the control's block variable correctly retains its checkpoint between calls to `rf(new_value)` and `rf.derivative()`. The helpers now take an optional list of controls (accepting either `Control` objects or the underlying overloaded variables) and skip those block variables in the clear-down sweeps. All existing call sites in `test_burgers_newton.py` and `test_checkpointing_multistep.py` are updated to thread the controls through.

The second is a regression test `test_control_value_survives_recompute` in `test_checkpointing_multistep.py` that captures Steph's MFE from firedrakeproject/firedrake#5082 directly: four timesteps of `J = sum_k m**2`, evaluated at `m0 = 2`, must give a derivative of `16`. Before the pyadjoint fix this returned `8` because the adjoint replay saw the stale underlying `m = 1`.

Depends on dolfin-adjoint/pyadjoint#257 — without it the new regression test will fail.

### DIFF
--- a/tests/firedrake/adjoint/test_burgers_newton.py
+++ b/tests/firedrake/adjoint/test_burgers_newton.py
@@ -39,11 +39,11 @@ def _control_bvs(controls):
     # The control block variables legitimately retain their checkpoint
     # across forward/adjoint replays (that is how the user-supplied
     # control value is plumbed through to the next evaluation), so the
-    # post-replay clear-down assertions must skip them. Accepts either
-    # Control objects or the underlying overloaded variables.
+    # post-replay clear-down assertions must skip them. Expects the
+    # control overloaded variables.
     if controls is None:
         return set()
-    return {getattr(c, "block_variable", c) for c in controls}
+    return {c.block_variable for c in controls}
 
 
 def _check_forward(tape, controls=None):
--- a/tests/firedrake/adjoint/test_burgers_newton.py
+++ b/tests/firedrake/adjoint/test_burgers_newton.py
@@ -35,34 +35,59 @@ def setup_test(mesh):
     return V, ic, nu
 
 
-def _check_forward(tape):
+def _control_bvs(controls):
+    # The control block variables legitimately retain their checkpoint
+    # across forward/adjoint replays (that is how the user-supplied
+    # control value is plumbed through to the next evaluation), so the
+    # post-replay clear-down assertions must skip them. Accepts either
+    # Control objects or the underlying overloaded variables.
+    if controls is None:
+        return set()
+    return {getattr(c, "block_variable", c) for c in controls}
+
+
+def _check_forward(tape, controls=None):
+    skip = _control_bvs(controls)
     for current_step in tape.timesteps[1:-1]:
         for block in current_step:
             for deps in block.get_dependencies():
+                if deps in skip:
+                    continue
                 if (
                     deps not in tape.timesteps[0].checkpointable_state
                     and deps not in tape.timesteps[-1].checkpointable_state
                 ):
                     assert deps._checkpoint is None
             for out in block.get_outputs():
+                if out in skip:
+                    continue
                 if out not in tape.timesteps[-1].checkpointable_state:
                     assert out._checkpoint is None
 
 
-def _check_recompute(tape):
+def _check_recompute(tape, controls=None):
+    skip = _control_bvs(controls)
     for current_step in tape.timesteps[1:-1]:
         for block in current_step:
             for deps in block.get_dependencies():
+                if deps in skip:
+                    continue
                 if deps not in tape.timesteps[0].checkpointable_state:
                     assert deps._checkpoint is None
             for out in block.get_outputs():
+                if out in skip:
+                    continue
                 assert out._checkpoint is None
 
     for block in tape.timesteps[0]:
         for out in block.get_outputs():
+            if out in skip:
+                continue
             assert out._checkpoint is None
     for block in tape.timesteps[len(tape.timesteps)-1]:
         for deps in block.get_dependencies():
+            if deps in skip:
+                continue
             if (
                 deps not in tape.timesteps[0].checkpointable_state
                 and deps not in tape.timesteps[len(tape.timesteps)-1].adjoint_dependencies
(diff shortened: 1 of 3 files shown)


JHopeCollins:
Thanks for tackling this!

We need to be running this branch against the corresponding pyadjoint branch though, so ou'll need to change this line in the `pyproject.toml`:
https://github.com/firedrakeproject/firedrake/blob/6386785d7dd168fc0c1392a6558247b8496de246/pyproject.toml#L39
to:
`"pyadjoint-ad @ git+https://github.com/dolfin-adjoint/pyadjoint.git@sghelichkhani:sghelichkhani/fix-control-checkpoint-clearing",`

FYI I've not used an alternative branch from a fork before so you might have to change it to point to your fork directly instead of the main repo.

Arguably this should go into `release` seeing as it's a bugfix. Please can you open another PR branching off release with these changes so we can check it works there?

stephankramer:
I've set up a draft PR #5157 which has @sghelichkhani 's commit cherry-picked on top of release and testing against the sghelichkhani/fix-control-checkpoint-clearing branch of pyadjoint.

So what would be the order of things (assuming tests are passing): we get https://github.com/dolfin-adjoint/pyadjoint/pull/257 merged into pyadjoint master, and create  a patch version?, then merge this into Firedrake release?

JHopeCollins:
> I've set up a draft PR #5157 which has @sghelichkhani 's commit cherry-picked on top of release and testing against the sghelichkhani/fix-control-checkpoint-clearing branch of pyadjoint.

Great, thanks.

> So what would be the order of things (assuming tests are passing): we get [dolfin-adjoint/pyadjoint#257](https://github.com/dolfin-adjoint/pyadjoint/pull/257) merged into pyadjoint master, and create a patch version?, then merge this into Firedrake release?

I think the order would be (@connorjward?):

1. Merge into Firedrake `release` and pyadjoint `master`
2. Merge Firedrake `release` into Firedrake `main`.
3. Simultaneous point releases of Firedrake and Pyadjoint at some point in the future. Is this an urgent bug for you? It's pretty straightforward for us to do point releases so if it's a high priority bugfix then I don't see it being a problem doing one shortly after this is merged.

I can put it on the agenda for the meeting this afternoon.

tests/firedrake/adjoint/test_burgers_newton.py
```diff
@@ -35,55 +35,87 @@ def setup_test(mesh):
...
-def _check_forward(tape):
+def _control_bvs(controls):
+    # The control block variables legitimately retain their checkpoint
+    # across forward/adjoint replays (that is how the user-supplied
+    # control value is plumbed through to the next evaluation), so the
+    # post-replay clear-down assertions must skip them. Accepts either
+    # Control objects or the underlying overloaded variables.
+    if controls is None:
+        return set()
+    return {getattr(c, "block_variable", c) for c in controls}
```
JHopeCollins: Why complicate things by accepting either `Control` or `OverloadedType`? It'd be simpler to just only pass one type.

connorjward:
> > I've set up a draft PR #5157 which has @sghelichkhani 's commit cherry-picked on top of release and testing against the sghelichkhani/fix-control-checkpoint-clearing branch of pyadjoint.
> 
> 
> 
> Great, thanks.
> 
> 
> 
> > So what would be the order of things (assuming tests are passing): we get [dolfin-adjoint/pyadjoint#257](https://github.com/dolfin-adjoint/pyadjoint/pull/257) merged into pyadjoint master, and create a patch version?, then merge this into Firedrake release?
> 
> 
> 
> I think the order would be (@connorjward?):
> 
> 
> 
> 1. Merge into Firedrake `release` and pyadjoint `master`
> 
> 2. Merge Firedrake `release` into Firedrake `main`.
> 
> 3. Simultaneous point releases of Firedrake and Pyadjoint at some point in the future. Is this an urgent bug for you? It's pretty straightforward for us to do point releases so if it's a high priority bugfix then I don't see it being a problem doing one shortly after this is merged.
> 
> 
> 
> I can put it on the agenda for the meeting this afternoon.

Yeah this seems like the right thing. Don't make a patch release if you can avoid it (creates churn).

JHopeCollins:
Thanks for addressing my comment @sghelichkhani. Because this is a bugfix we want to merge it into `release` to it goes out with the next release. Please can you make the same change on the `release` PR: https://github.com/firedrakeproject/firedrake/pull/5157.
We'll then close this one and once #5157 has been merged into `release` we can merge `release` into `main` to pull in the changes.

sghelichkhani:
Thanks for getting 257 in and sorting the order out. Happy to go with whatever you decide on timing and how it reaches main, so please don't cut a point release just for me. Nothing in g-adopt is actually failing on this, so I'd rather it went out with the next normal release than add churn.

The only related thing still open on my end is pyadjoint #248 (the SingleMemoryStorageSchedule clearing fix), and that case is xfailed for now, so nothing's going red while we wait. Just flagging it's on my radar, but I'm fine for it to sit.

JHopeCollins:
> The only related thing still open on my end is pyadjoint https://github.com/dolfin-adjoint/pyadjoint/pull/248 (the SingleMemoryStorageSchedule clearing fix)

I'll admit that I am very unfamiliar with the checkpointing in general, but doesn't that do the same thing you are trying to fix with the current PRs? Accessing the `_checkpoint` directly?

If you put it on the meeting agenda for next week then we can take a look at it.

sghelichkhani:
Closing this one: the test changes landed on release as #5157 and reached main through the release merge in #5457, and the pyadjoint pin here pointed at the branch that was merged as dolfin-adjoint/pyadjoint#257 and released in 2026.9.0. Nothing outstanding here, thanks all.

PR #5094, opened 2026-05-11
Fixes to firedrake-configure

### LINKED ISSUE #5091: BUG: firedrake-config help doesn't match required arguments
The new --os flag help displays:

```
usage: firedrake-configure [-h]
                           [--os {OS.UBUNTU_2404_X86_64,OS.UBUNTU_2404_AARCH64,OS.MACOS_ARM64,OS.UNKNOWN}
```

However, you need to pass `python3 firedrake-configure --os ubuntu24.04-x86_64`

(i.e. the value, not the key as OS contains:
```
class OS(enum.Enum):
    UBUNTU_2404_X86_64 = "ubuntu24.04-x86_64"
    UBUNTU_2404_AARCH64 = "ubuntu24.04-aarch64"
    MACOS_ARM64 = "macos-arm64"
    UNKNOWN = "unknown"
```
)

connorjward:
Fixes #5091

Improves some of the error messages

### DIFF
--- a/scripts/firedrake-configure
+++ b/scripts/firedrake-configure
@@ -150,13 +150,13 @@ Please see https://firedrakeproject.org/install for more information."""
         "--scalar-type",
         "--arch",  # old alias
         choices=[st.value for st in ScalarType],
-        default="default",
+        default=ScalarType.DEFAULT.value,
         help="PETSc scalar type.",
     )
     parser.add_argument(
         "--gpu-arch",
         choices=[gpu_opt.value for gpu_opt in GPUPlatform],
-        default="none",
+        default=GPUPlatform.NONE.value,
         help=(
             "Target GPU architecture. WARNING: This is an experimental feature. "
             "GPU support in Firedrake is currently very limited."
--- a/scripts/firedrake-configure
+++ b/scripts/firedrake-configure
@@ -73,23 +73,26 @@ def main():
         warn(
             "Passing '--no-package-manager' is deprecated, please set '--os unknown' instead"
         )
-        args.os = OS.UNKNOWN
-    os_ = args.os if args.os else detect_os()
+        args.os = "unknown"
 
-    arch_key = (os_, args.scalar_type, args.gpu_arch)
-    arch_key_str = f"({', '.join(f'{k.__class__.__name__}: {k.name}' for k in arch_key)})"
+    os_ = OS(args.os) if args.os else detect_os()
+    scalar_type = ScalarType(args.scalar_type)
+    gpu_arch = GPUPlatform(args.gpu_arch)
+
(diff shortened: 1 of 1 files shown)


scripts/firedrake-configure
```diff
@@ -73,24 +73,25 @@ def main():
...
+    arch_key_str = f"{{{', '.join(f'{k.__class__.__name__}: {k.value}' for k in arch_key)}}}"
     if arch_key in OFFICIAL_ARCHS:
         arch = OFFICIAL_ARCHS[arch_key]
     elif arch_key in COMMUNITY_ARCHS:
         arch = COMMUNITY_ARCHS[arch_key]
-        warn(
-            "You have selected a community-maintained Firedrake configuration, "
-            f"if you encounter issues please contact {arch.maintainer}"
-        )
+        warn(f"""\
```
JHopeCollins, line 88: What does this reformatting do?
Also, using `dedent` here would be very useful for text editors with indent-based folding!

scripts/firedrake-configure
```diff
@@ -149,15 +149,13 @@ Please see https://firedrakeproject.org/install for more information."""
     parser.add_argument(
         "--scalar-type",
         "--arch",  # old alias
-        choices=list(ScalarType),
-        type=ScalarType,
-        default=ScalarType.DEFAULT,
+        choices=[st.value for st in ScalarType],
+        default="default",
```
JHopeCollins: A bit picky but I think this is better style
```suggestion
        default=ScalarType.DEFAULT.value,
```

scripts/firedrake-configure
```diff
@@ -149,15 +149,13 @@ Please see https://firedrakeproject.org/install for more information."""
...
         help="PETSc scalar type.",
     )
     parser.add_argument(
         "--gpu-arch",
-        choices=list(GPUPlatform),
-        type=GPUPlatform,
+        choices=[gpu_opt.value for gpu_opt in GPUPlatform],
         default="none",
```
JHopeCollins:
```suggestion
        default=GPUPlatform.NONE.value,
```

scripts/firedrake-configure
connorjward: I tend to avoid using dedent because it doesn't work with f strings nicely.

The reformatting here and throughout was to split the messages onto multiple lines, because they were splitting unhelpfully and hard to read.

scripts/firedrake-configure
connorjward: I actually had this exact thought just before you ended up suggesting it. This is better.

connorjward:
The macOS and CUDA build failures are unrelated to this and I have asked Tim to look into them.

JHopeCollins (approved):

PR #5105, opened 2026-05-13
Disable -O3 for GCC 15

connorjward:
We found out that -O3 was leading to garbage values in tricky interpolation kernels with GCC 15 on both Ubuntu 26.04 and Arch.

This will allow us to support Ubuntu 26.04.

I have also refactored `Compiler.cflags` and friends because the bugfix flags have to come after the optflags.

Also removed a bunch of very very old bugfix cflags - GCC 7 was last updated in 2019 and I'd be surprised if the rest of our software stack works with such an old compiler.

### DIFF
--- a/pyop2/compilation.py
+++ b/pyop2/compilation.py
@@ -278,33 +278,35 @@ class Compiler(ABC):
         return self._ld
 
     @property
-    def cflags(self):
-        cflags = self._cflags + self._extra_compiler_flags + self.bugfix_cflags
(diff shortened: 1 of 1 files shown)


JHopeCollins (approved):
LGTM, thanks for tracking this down!

connorjward:
I'm going to fold this into https://github.com/firedrakeproject/firedrake/pull/5085 because this only really makes sense in the context of that PR

PR #5109, opened 2026-05-13
Fix cell_avg + MixedFunctionSpace

pbrubeck:
# Description

Fixes a bug with `cell_avg` + `MixedFunctionSpace` by not renumbering coefficients in the inner calls to `compute_form_data`. Renumbering should only be done by the outermost call.

Adds tests for `cell_avg`

### DIFF
--- a/tsfc/ufl_utils.py
+++ b/tsfc/ufl_utils.py
@@ -43,6 +43,7 @@ def compute_form_data(form,
                       do_apply_default_restrictions=True,
                       do_apply_restrictions=True,
                       do_estimate_degrees=True,
+                      do_replace_functions=True,
                       coefficients_to_split=None,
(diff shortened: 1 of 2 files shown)


tsfc/ufl_utils.py
```diff
@@ -44,7 +44,9 @@ def compute_form_data(form,
                       do_apply_restrictions=True,
                       do_estimate_degrees=True,
                       coefficients_to_split=None,
-                      complex_mode=False):
+                      complex_mode=False,
+                      do_replace_functions=True,
```
pbrubeck:
```suggestion
                      do_replace_functions=True,
                      complex_mode=False,
```

connorjward (approved):

PR #5110, opened 2026-05-14
Speed up CI setup and refactor PyPI publishing workflow

connorjward:
* Compile petsc4py and Firedrake with -O0
* Only build the sdist when we publish, not every time (I think it meant we were building things twice)

### DIFF
--- a/.github/actions/install/action.yml
+++ b/.github/actions/install/action.yml
@@ -117,7 +117,7 @@ runs:
           --gpu-arch ${{ inputs.gpu }} \
           --show-env)
         export SLEPC_DIR=$PETSC_DIR/$PETSC_ARCH
-        :# Disable compiler optimisation to speed up build (especially petsc4py)
+        : # Disable compiler optimisation to speed up build (especially petsc4py)
         export CFLAGS="-O0"
         env
 
--- a/.github/actions/install/action.yml
+++ b/.github/actions/install/action.yml
@@ -24,7 +24,7 @@ inputs:
   deps:
     description: Optional dependencies to install along with Firedrake (e.g. 'check')
     type: string
-    required: true
+    default: check
 
 runs:
   using: composite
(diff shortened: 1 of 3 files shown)


connorjward:
Looks to turn the install from 15 mins to 10 mins.

.github/actions/install/action.yml
```diff
@@ -137,24 +139,18 @@ runs:
...
-          EXTRA_BUILD_ARGS=''
           EXTRA_PIP_FLAGS=''
         else
           echo "Unrecognised 'base_ref' input: '${{ inputs.base_ref }}"
           exit 1
         fi
 
-        : # Install from an sdist so we can make sure that it is not ill-formed
-        pip install build
-        python -m build ./firedrake-repo --sdist $EXTRA_BUILD_ARGS
```
JHopeCollins, line 152: Wasn't this specifically put in so that we knew if we were breaking the sdist _before_ trying to release?

connorjward: Yes, but I think it doubles the time it takes to install Firedrake and we've never encountered an issue with it apart from when we first started doing releases.

.github/workflows/release.yml
```diff
@@ -45,15 +45,41 @@ jobs:
           fi
 
   pypi:
-    uses: ./.github/workflows/core.yml
+    name: Make PyPI release
```
JHopeCollins, line 48: We no longer test immediately before release?

connorjward: We never did apart from `firedrake-check` (which is still the case)

connorjward: And our [procedure](https://github.com/firedrakeproject/firedrake/wiki/Make-a-release#how-to-make-a-patch-release) states that you should test it in a separate DO NOT MERGE PR immediately beforehand.

JHopeCollins (approved):
LGTM. Always nice to speed up CI!

PR #5119, opened 2026-05-15
`AuxiliaryOperatorSNES`

JHopeCollins:
If I am trying to solve `F(u)=0` but for whatever reason Newton isn't working or is slow to converge, this snes type enables specifying an auxiliary nonlinear operator `G(u)` that can be used as a nonlinear preconditioner.

The simplest form of nonlinear preconditioning is a for a nonlinear Richardson iteration:
```math
G(u^{k+1}) = G(u^{k}) - F(u^{k})
```
where the solution `uhat` of `F(uhat)=0` is clearly a fixed point.

There's a demo for the Allen-Cahn equation where Newton fails but using a `G` that lags a particular term in `F` succeeds. `AuxiliaryOpertorSNES` can also be used for continuation methods like Reynolds continuation, and many other methods.

This cherrypicks the `AuxiliaryOperatorSNES` from https://github.com/firedrakeproject/firedrake/pull/4544 because it is ready to go but the `FieldsplitSNES` could do with a little more stress testing/tidying up.

Pointing at `release` because this is strictly new functionality. The edits to `firedrake/preconditioners/base.py` are just updating the docstrings to be clearer about the distinction between `PCBase` and `SNESBase`.

### DIFF
--- a/demos/nonlinear_pc/nonlinear_pc_allen_cahn.py.rst
+++ b/demos/nonlinear_pc/nonlinear_pc_allen_cahn.py.rst
@@ -1,0 +1,177 @@
+Nonlinear preconditioning applied to the Allen-Cahn equation
+============================================================
+
+In the other demonstrations on nonlinear PDE, we've used Newton's method to
+solve finite- dimensional systems of nonlinear equations. Suppose we want to
+solve a system
+
+.. math::
+  F(u) = 0.
+
+At each step of Newton's method, we first compute a *search direction*
+:math:`v` by solving the linear system
+.. math::
+
+  dF(u)v = -F(u).
+
+We can then use a line search along :math:`v` to obtain the next candidate
+solution. When we talked about preconditioning, we have always meant applying
+a preconditioner to the linear system for the search direction. A linear
+preconditioner transforms a linear system into an equivalent one that is
+easier to solve.
+
+For some highly nonlinear problems, however, Newton's method can stagnate or
+fail to converge, even with globalization strategies such as a line search.
+For example, if :math:`dF(u)` behaves badly, we might not be able to compute a
+search direction at all.
+
+An alternative strategy is to use *nonlinear preconditioning*, abbreviated as
+NPC in the following. Nonlinear preconditioning does the same thing as linear
+preconditioning: it transforms the problem into an equivalent one that is
+(hopefully) easier to solve. Rather than work on the linear system for the
+search direction, however, NPC works directly on the nonlinear system itself.
+Typically NPC is applied within a higher-level nonlinear solver. For example,
+there is a nonlinear analogue of GMRES. For a review of nonlinear solver and
+preconditioning strategies beyond Newton, see :cite:`brune2015composing`.
+
+Here we will demonstrate how to use nonlinear preconditioning for the steady-
+state Allen-Cahn equation
+
+.. math::
+
+  -\epsilon\Delta u - u + u^3 = 0
+
+on the unit interval with Dirichlet boundary conditions :math:`u(0) = -1` and
+:math:`u(1) = +1`. The Jacobian of this residual is indefinite wherever
+:math:`|u| < 1/\sqrt{3}`. Newton's method can fail to compute a search
+direciton when starting from an initial guess that crosses this region.
+
+This demo is adapted from this `Chebfun example`_.
+
+.. _Chebfun example: https://www.chebfun.org/examples/ode-nonlin/AllenCahn.html
+
+::
+
+  import numpy as np
+  from firedrake import *
+
+Here we use a domain of length 10, a small diffusion coefficient of 0.003,
+and an initial guess that ramps from +1 at the left-hand boundary to -1 at
+the right.
+
+::
+
+  nx = 128
+  lx = 10.0
+  eps = Constant(3e-3)
+
+  mesh = IntervalMesh(nx, lx)
+  Q = FunctionSpace(mesh, "CG", 1)
+
+  x, = SpatialCoordinate(mesh)
+  u_1 = Constant(1)
+  u_2 = Constant(-1)
+  Lx = Constant(lx)
+  initial_guess = (1 - x / Lx) * u_1 + x / Lx * u_2
+
+  bcs = [DirichletBC(Q, u_1, [1]), DirichletBC(Q, u_2, [2])]
+
+  u = Function(Q)
+  u.interpolate(initial_guess)
+
+In order to have a good baseline to compare against, we want to use as good
+a nonlinear solution strategy as possible. Here we use Newton's method with
+the *critical point* line search, which is specially adapted for problems
+like Allen-Cahn which can be derived from minimizing some free energy. Even
+with 10 iterations of a line search, Newton's method will fail. You can try
+other line search methods (secant, backtracking, etc.) and find similar
+outcomes. If you pump the number of line search iterations way up, you can
+make Newton's method converge... to the wrong solution!
+
+::
+
+  v = TestFunction(Q)
+  F = (eps * inner(grad(u), grad(v)) - (u - u**3) * v) * dx
+
+  problem = NonlinearVariationalProblem(F, u, bcs)
+
+  newton_parameters = {
+      "snes_type": "newtonls",
+      "snes_monitor": None,
+      "snes_converged_reason": None,
+      "snes_linesearch_type": "cp",
+      "snes_linesearch_max_it": 10,
+      "snes_linesearch_monitor": None,
+  }
+  solver = NonlinearVariationalSolver(problem, solver_parameters=newton_parameters)
+  try:
+      solver.solve()
+  except ConvergenceError as err:
+      print(err)
+      print("--------------------------")
+      print("Solver failed to converge!")
+      print("Resetting `u`")
+      u.interpolate(initial_guess)
+
+To salvage the wreck, we'll use nonlinear preconditioning. Here we define
+a custom preconditioner which inherits from `AuxiliaryOperatorSNES`. The
+arguments that it takes are first the nonlinear equation solver; then the
+value `u_k` of the solution at the previous iteration; the current value `u`
+for the solution; and a test function. It outputs the variational form which
+will determine the value of `u`.
+
+Our approach here is to solve for a candidate guess by holding back the
+linear part of the reaction term in the Allen-Cahn equation. In other words,
+at each step, we solve the PDE
+
+.. math::
+
+  -\epsilon\Delta u - u_k + u^3 = 0.
+
+The Jacobian for this problem w.r.t. :math:`u` is symmetric and positive-
+definite. We have good guarantees about the convergence of Newton's method
+in that case, so we can reuse the solver parameters that we tried the first
+time around for this inner problem.
+
+::
+
+  class AllenCahnAuxSNES(firedrake.AuxiliaryOperatorSNES):
+      def form(self, snes, u_k, u, v):
+          return (eps * inner(grad(u), grad(v)) + (u**3 - u_k) * v) * dx, bcs
+
+Here we use the simplest solution strategy possible: nonlinear Richardson
+iteration. We then specify the additional parameters under the key `"npc"`.
+
+::
+
+  solver_parameters = {
+      "snes_type": "nrichardson",
+      "snes_monitor": None,
+      "snes_converged_reason": None,
+      "npc": {
+          "snes_type": "python",
+          "snes_python_type": f"{__name__}.AllenCahnAuxSNES",
+          "aux": newton_parameters,
+      },
+  }
+
+  solver = NonlinearVariationalSolver(problem, solver_parameters=solver_parameters)
+  solver.solve()
+
+The Allen-Cahn equation is derivable through minimization of the free energy
+functional
+
+.. math::
+
+  G(u) = \int\_\Omega\left(\frac{\epsilon}{2}|\nabla u|^2 + (1 - u^2)^2/4\right)dx.
+
+To close, let's evaluate the free energy at the starting guess and at the
+computed solution.
+
+::
+
+  G = (0.5 * eps * inner(grad(u), grad(u)) + 0.25 * (1 - u**2) ** 2) * dx
+  G_initial = firedrake.assemble(firedrake.replace(G, {u: initial_guess}))
+  G_final = firedrake.assemble(G)
+  print(f"Initial free energy: {G_initial:0.04f}")
+  print(f"Final:               {G_final:0.04f}")
--- a/demos/nonlinear_pc/nonlinear_pc_allen_cahn.py.rst
+++ b/demos/nonlinear_pc/nonlinear_pc_allen_cahn.py.rst
@@ -1,6 +1,9 @@
 Nonlinear preconditioning applied to the Allen-Cahn equation
 ============================================================
 
+Contributed by `Daniel Shapero <https://psc.apl.uw.edu/people/investigators/daniel-shapero/>`_
+with edits by `Josh Hope-Collins <https://profiles.imperial.ac.uk/joshua.hope-collins13>`_.
+
 In the other demonstrations on nonlinear PDE, we've used Newton's method to
 solve finite- dimensional systems of nonlinear equations. Suppose we want to
 solve a system
@@ -39,7 +42,7 @@ state Allen-Cahn equation
 
 .. math::
 
-  -\epsilon\Delta u - u + u^3 = 0
+  F(u) = -\epsilon\Delta u - u + u^3 = 0
 
 on the unit interval with Dirichlet boundary conditions :math:`u(0) = -1` and
 :math:`u(1) = +1`. The Jacobian of this residual is indefinite wherever
@@ -113,25 +116,52 @@ make Newton's method converge... to the wrong solution!
       print("Resetting `u`")
       u.interpolate(initial_guess)
 
-To salvage the wreck, we'll use nonlinear preconditioning. Here we define
-a custom preconditioner which inherits from `AuxiliaryOperatorSNES`. The
-arguments that it takes are first the nonlinear equation solver; then the
-value `u_k` of the solution at the previous iteration; the current value `u`
-for the solution; and a test function. It outputs the variational form which
-will determine the value of `u`.
-
-Our approach here is to solve for a candidate guess by holding back the
-linear part of the reaction term in the Allen-Cahn equation. In other words,
-at each step, we solve the PDE
+To salvage the wreck, we'll use nonlinear preconditioning. Here we use
+the simplest solution strategy possible: preconditioned nonlinear
+Richardson iterations.
+The idea is that if :math:`F(u)=0` is too difficult to solve, we can
+instead solve a auxiliary operator :math:`G(u)=0` which is in some sense
+*nearby* to :math:`F`, and use that solution to iterate towards a solution
+for :math:`F(u)=0`.
+At each iteration :math:`k` we solve the following system for the next
+iterate :math:`u_{k+1}` using the value of the current iterate :math:`u_k`.
 
 .. math::
 
-  -\epsilon\Delta u - u_k + u^3 = 0.
+   G_{k}(u_{k+1}) = G_{k}(u_{k}) - F(u_{k}),
+   \quad
+   G_{k}(u) = G(u; u_{k}).
+
+We can note several properties of this iteration. Firstly, if
+:math:`F(u_{*})=0` then :math:`u_{k}=u_{*}` is a fixed point of the iteration.
+Secondly, we never have to solve :math:`F(u)=0`, we only have to evaluate
+its residual at a given state. Thirdly, we can intuitively hope that if
+:math:`G` is *close enough* to :math:`F` then the iteration will converge,
+although proving this is much more difficult than in the linear case.
+Lastly, we can define a different :math:`G_k` at each iteration by
+parameterising with the current iterate, as we will see below.
+
+Our approach for defining :math:`G` here is to hold back the linear part of
+the reaction term in the Allen-Cahn equation to the value at the previous
+iteration. In other words, at each step, we define the PDE
+
+.. math::
+
+  G(u; u_k) = -\epsilon\Delta u - u_k + u^3 = 0.
 
 The Jacobian for this problem w.r.t. :math:`u` is symmetric and positive-
 definite. We have good guarantees about the convergence of Newton's method
 in that case, so we can reuse the solver parameters that we tried the first
 time around for this inner problem.
+
+Here we define a custom nonlinear preconditioner which inherits from
+``AuxiliaryOperatorSNES``. Similarly to the ``AuxiliaryOperatorPC`` we
+must implement the ``form`` method, which must return a residual form
+:math:`G(u; u^{k}, v)` where :math:`v` is the test function.
+The arguments that it takes are first the PETSc SNES object,
+then the value :math:`u_k` of the solution at the previous iteration;
+the current value :math:`u` to be solved for; and a test function.
+It outputs the variational form for :math:`G` and the boundary conditions.
 
 ::
 
@@ -139,8 +169,11 @@ time around for this inner problem.
       def form(self, snes, u_k, u, v):
           return (eps * inner(grad(u), grad(v)) + (u**3 - u_k) * v) * dx, bcs
 
-Here we use the simplest solution strategy possible: nonlinear Richardson
-iteration. We then specify the additional parameters under the key `"npc"`.
+We now set ``-snes_type nrichardson`` for nonlinear Richardson iterations,
+and specify the additional parameters for the nonlinear preconditioner under
+the key ``"npc"``. Firstly we need to specify our python SNES type, and then
+in the ``"aux"`` key we specify the parameters to actually solve :math:`G` -
+here we use Newton iterations.
 
 ::
 
@@ -157,6 +190,11 @@ iteration. We then specify the additional parameters under the key `"npc"`.
 
   solver = NonlinearVariationalSolver(problem, solver_parameters=solver_parameters)
   solver.solve()
+
+Now we actually converge! The convergence on nonlinear Richardson iterations is
+usually linear, as opposed to the quadratic convergence of Newton, but with
+suitable preconditioning they will usually have a wider basin of convergence than
+Newton.
 
 The Allen-Cahn equation is derivable through minimization of the free energy
 functional
--- a/demos/nonlinear_pc/nonlinear_pc_allen_cahn.py.rst
+++ b/demos/nonlinear_pc/nonlinear_pc_allen_cahn.py.rst
@@ -214,3 +214,10 @@ computed solution.
   G_final = firedrake.assemble(G)
   print(f"Initial free energy: {G_initial:0.04f}")
   print(f"Final:               {G_final:0.04f}")
+
+This demo can be found as a script in :demo:`nonlinear_pc_allen_cahn.py <nonlinear_pc_allen_cahn.py>`.
+
+.. rubric:: References
+
+.. bibliography:: demo_references.bib
+   :filter: docname in docnames
--- a/demos/nonlinear_pc/nonlinear_pc_allen_cahn.py.rst
+++ b/demos/nonlinear_pc/nonlinear_pc_allen_cahn.py.rst
@@ -2,17 +2,19 @@ Nonlinear preconditioning applied to the Allen-Cahn equation
 ============================================================
 
 Contributed by `Daniel Shapero <https://psc.apl.uw.edu/people/investigators/daniel-shapero/>`_
-with edits by `Josh Hope-Collins <https://profiles.imperial.ac.uk/joshua.hope-collins13>`_.
+and `Josh Hope-Collins <https://profiles.imperial.ac.uk/joshua.hope-collins13>`_.
 
 In the other demonstrations on nonlinear PDE, we've used Newton's method to
 solve finite- dimensional systems of nonlinear equations. Suppose we want to
 solve a system
 
 .. math::
+
   F(u) = 0.
 
 At each step of Newton's method, we first compute a *search direction*
 :math:`v` by solving the linear system
+
 .. math::
 
   dF(u)v = -F(u).
@@ -116,6 +118,32 @@ make Newton's method converge... to the wrong solution!
       print("Resetting `u`")
       u.interpolate(initial_guess)
 
+The residual decreases at first but eventually diverges. Here's some of the PETSc
+log output:
+
+.. code-block:: console
+
+    $ python nonlinear_pc_allen_cahn.py
+      0 SNES Function norm 2.439229081145e-01
+      1 SNES Function norm 5.663027251974e+01
+      2 SNES Function norm 1.667065795962e+01
+      3 SNES Function norm 4.864679262673e+00
+      4 SNES Function norm 1.388964354434e+00
+      5 SNES Function norm 3.744562754242e-01
+      6 SNES Function norm 8.820230192544e-02
+      7 SNES Function norm 3.682817028776e-02
+      8 SNES Function norm 5.405858541080e-02
+      9 SNES Function norm 5.368629859638e-02
+       ...
+     26 SNES Function norm 5.280237868253e-02
+     27 SNES Function norm 1.098413882561e+00
+     28 SNES Function norm 1.082944223664e+00
+     29 SNES Function norm 6.520280025999e+03
+      Nonlinear firedrake_0_ solve did not converge due to DIVERGED_DTOL iterations 29
+
+In other scenarios, the solver doesn't explode as dramatically but rather
+stagnates and exceeds the number of allowable Newton iterations.
+
 To salvage the wreck, we'll use nonlinear preconditioning. Here we use
 the simplest solution strategy possible: preconditioned nonlinear
 Richardson iterations.
@@ -156,19 +184,27 @@ time around for this inner problem.
 
 Here we define a custom nonlinear preconditioner which inherits from
 :class:`~firedrake.preconditioners.auxiliary_snes.AuxiliaryOperatorSNES`.
-Similarly to the ``AuxiliaryOperatorPC`` we
-must implement the ``form`` method, which must return a residual form
-:math:`G(u; u^{k}, v)` where :math:`v` is the test function.
+Similar to ``AuxiliaryOperatorPC``, we have to implement the ``form`` method.
+This method returns (1) a residual form :math:`G(u; u^{k}, v)` where :math:`v`
+is the test function and (2) the boundary conditions for this sub-problem.
 The arguments that it takes are first the PETSc SNES object,
 then the value :math:`u_k` of the solution at the previous iteration;
 the current value :math:`u` to be solved for; and a test function.
-It outputs the variational form for :math:`G` and the boundary conditions.
 
 ::
 
   class AllenCahnAuxSNES(firedrake.AuxiliaryOperatorSNES):
       def form(self, snes, u_k, u, v):
+          F, bcs = super().form(snes, u_k, u, v)
           return (eps * inner(grad(u), grad(v)) + (u**3 - u_k) * v) * dx, bcs
+
+The contract for the ``form`` method requires it to supply the boundary
+conditions. We could have obtained the boundary conditions by pulling them out
+of the global context. That will work for this particular set of solvers but
+can create problems for others, for example when using the multigrid method.
+Instead, we've opted to call the parent class's `form` method, which will
+return the original variational form and boundary conditions. We then discard
+the original variational form, which we aren't using here.
 
 We now set ``-snes_type nrichardson`` for nonlinear Richardson iterations,
 and specify the additional parameters for the nonlinear preconditioner under
@@ -195,14 +231,32 @@ here we use Newton iterations.
 Now we actually converge! The convergence on nonlinear Richardson iterations is
 usually linear, as opposed to the quadratic convergence of Newton, but with
 suitable preconditioning they will usually have a wider basin of convergence than
-Newton.
+Newton. From the log output, we can observe that the residual is decreasing by a
+factor of 2 or more at each iteration.
+
+.. code-block:: console
+
+      0 SNES Function norm 2.439229081145e-01
+      1 SNES Function norm 2.405339859939e-01
+      2 SNES Function norm 1.540442351803e-01
+      3 SNES Function norm 7.166137071498e-02
+      4 SNES Function norm 2.854773301463e-02
+      5 SNES Function norm 1.070593887487e-02
+      6 SNES Function norm 3.943106335233e-03
+      7 SNES Function norm 1.451230558440e-03
+       ...
+     18 SNES Function norm 3.491990058865e-08
+     19 SNES Function norm 1.349411703695e-08
+     20 SNES Function norm 5.217958176918e-09
+     21 SNES Function norm 2.018693509573e-09
+      Nonlinear firedrake_1_ solve converged due to CONVERGED_FNORM_RELATIVE iterations 21
 
 The Allen-Cahn equation is derivable through minimization of the free energy
 functional
 
 .. math::
 
-  G(u) = \int\_\Omega\left(\frac{\epsilon}{2}|\nabla u|^2 + (1 - u^2)^2/4\right)dx.
+  G(u) = \int_\Omega\left(\frac{\epsilon}{2}|\nabla u|^2 + \frac{1}{4}(1 - u^2)^2\right)dx.
 
 To close, let's evaluate the free energy at the starting guess and at the
 computed solution.
@@ -214,6 +268,11 @@ computed solution.
   G_final = firedrake.assemble(G)
   print(f"Initial free energy: {G_initial:0.04f}")
   print(f"Final:               {G_final:0.04f}")
+
+.. code-block:: console
+
+    Initial free energy: 1.3339
+    Final:               0.0534
 
 This demo can be found as a script in :demo:`nonlinear_pc_allen_cahn.py <nonlinear_pc_allen_cahn.py>`.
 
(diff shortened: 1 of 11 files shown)


dham (changes requested):
minor changes pablo has requested to the demo.

demos/nonlinear_pc/nonlinear_pc_allen_cahn.py.rst
```diff
@@ -0,0 +1,223 @@
...
+          "snes_type": "python",
+          "snes_python_type": f"{__name__}.AllenCahnAuxSNES",
+          "aux": newton_parameters,
+      },
+  }
+
+  solver = NonlinearVariationalSolver(problem, solver_parameters=solver_parameters)
+  solver.solve()
+
+Now we actually converge! The convergence on nonlinear Richardson iterations is
```
pbrubeck, line 230: Please add a table with residual convergence

demos/nonlinear_pc/nonlinear_pc_allen_cahn.py.rst
```diff
@@ -0,0 +1,223 @@
...
+The arguments that it takes are first the PETSc SNES object,
+then the value :math:`u_k` of the solution at the previous iteration;
+the current value :math:`u` to be solved for; and a test function.
+It outputs the variational form for :math:`G` and the boundary conditions.
+
+::
+
+  class AllenCahnAuxSNES(firedrake.AuxiliaryOperatorSNES):
+      def form(self, snes, u_k, u, v):
+          return (eps * inner(grad(u), grad(v)) + (u**3 - u_k) * v) * dx, bcs
```
pbrubeck: Grabbing bcs from the global context often leads to problems in more complicated nested operators (multigrid). We should not encourage that in the demo.
```suggestion
      def form(self, snes, u_k, u, v):
          F, bcs = super().form(snes, u_k, u, v)
          # Could manipulate or add terms to the original F, but here we create we create G from scratch
          G = eps * inner(grad(u), grad(v)) * dx + inner(u**3 - u_k, v) * dx
          return G, bcs
```

firedrake/preconditioners/auxiliary_snes.py
```diff
@@ -0,0 +1,236 @@
...
+
+        # auxiliary form G(k+1)
+        test = TestFunction(V)
+        Gk1, bcs = self.form(snes, uk, uk1, test)
+
+        # Solve G(k+1) - b = 0
+        # with forcing b = G(k) - F(k)
+        Gk = replace(Gk1, {uk1: uk})
+        b = Cofunction(V.dual())
+        Gk1 -= b
```
pbrubeck: Gk1 is not G(u^{k+1}) anymore. Is this the right name? How about `F_aux`?

firedrake/preconditioners/auxiliary_snes.py
```diff
@@ -0,0 +1,236 @@
...
+        # assemble the forcing terms to avoid having to
+        # re-evaluate at every iteration of the inner snes.
+        self.assemble_gk = get_assembler(
+            Gk, bcs=bcs,
+            form_compiler_parameters=ctx.fcp,
+            options_prefix=prefix
+        ).assemble
+
+        self.Gk = Gk
+        self.Gk1 = Gk1
```
pbrubeck: Do these need to be stored as class attributes?

firedrake/preconditioners/auxiliary_snes.py
```diff
@@ -0,0 +1,236 @@
...
+        # b = G(u^{k})
+        self.assemble_gk(tensor=self.b)
+
+        # b = G(u^{k}) - F(u^{k})
+        with self.b.dat.vec as vec:
+            vec -= f
+
+        # G(u^{k+1}) - b = 0
+        self.uk1.assign(self.uk)
+        self.solver.solve()
```
pbrubeck, line 149: We might want to enable `solver.solve(rhs=b)` so the logic with Gk1 becomes clearer. Obviously a separate PR.

firedrake/preconditioners/auxiliary_snes.py
```diff
@@ -0,0 +1,236 @@
...
+        b = Cofunction(V.dual())
+        Gk1 -= b
+
+        # assemble the forcing terms to avoid having to
+        # re-evaluate at every iteration of the inner snes.
+        self.assemble_gk = get_assembler(
+            Gk, bcs=bcs,
+            form_compiler_parameters=ctx.fcp,
+            options_prefix=prefix
+        ).assemble
```
pbrubeck: Potentially one could reuse the residual assembler in the _SNESContext of the solver defined below, similar to what we did here for `AssembledPC` 

https://github.com/firedrakeproject/firedrake/pull/5079/changes

Up to you if you want to defer this job to a separate PR.

demos/nonlinear_pc/nonlinear_pc_allen_cahn.py.rst
danshapero: Done, I put in some more text explaining why this is good practice.

demos/nonlinear_pc/nonlinear_pc_allen_cahn.py.rst
```diff
@@ -0,0 +1,282 @@
...
+          F, bcs = super().form(snes, u_k, u, v)
+          return (eps * inner(grad(u), grad(v)) + (u**3 - u_k) * v) * dx, bcs
+
+The contract for the ``form`` method requires it to supply the boundary
+conditions. We could have obtained the boundary conditions by pulling them out
+of the global context. That will work for this particular set of solvers but
+can create problems for others, for example when using the multigrid method.
+Instead, we've opted to call the parent class's `form` method, which will
+return the original variational form and boundary conditions. We then discard
+the original variational form, which we aren't using here.
```
pbrubeck: Maybe mention that we can also manipulate F

tests/firedrake/regression/test_auxiliary_snes.py
```diff
@@ -0,0 +1,92 @@
...
+        + w*fd.inner(u*u*u, v)*fd.dx
+        - fd.inner(fd.Constant(1), v)*fd.dx
+    )
+
+
+class AuxiliaryPolynomialSNES(fd.AuxiliaryOperatorSNES):
+    def form(self, snes, state, func, test):
+        prefix = (snes.getOptionsPrefix() or "") + f"snes_{self._prefix}"
+        w = fd.PETSc.Options().getScalar(prefix + "w")
+        return f(func, test, w=w), None
```
pbrubeck:
```suggestion
    def form(self, snes, state, func, test):
        F, bcs = super().form(snes, state, func, test)
        prefix = (snes.getOptionsPrefix() or "") + f"snes_{self._prefix}"
        w = fd.PETSc.Options(prefix).getScalar("w")
        return f(func, test, w=w), bcs
```

firedrake/preconditioners/auxiliary_snes.py
```diff
@@ -0,0 +1,236 @@
...
+        self.uk1.assign(self.uk)
+        self.solver.solve()
+
+        # y = d = u^{k+1} - u^{k}
+        with self.uk1.dat.vec_ro as vec:
+            vec.copy(y)
+            y.aypx(-1, x)
+
+    @PETSc.Log.EventDecorator()
+    def form(self, snes, uk: Function, uk1: Function, test: Argument):
```
pbrubeck: Maybe `uold`, `unew` would be more clear here

JHopeCollins: uk1 and uk matches the equations describing the iteration in the class docs and the demo so they are clearer here than old/new

pbrubeck: Sure, with the context of the docstrings, you could tell that `uk1` refers to `u_{k+1}`. But a subclass will not include the docstring, so the context will get lost easily in user code. If a user cargo-culted this notation, it be hard to tell whether `uk1` refers to `u_{k-1}` or `u_{k+1}`.

demos/nonlinear_pc/nonlinear_pc_allen_cahn.py.rst
```diff
@@ -0,0 +1,282 @@
...
+The arguments that it takes are first the PETSc SNES object,
+then the value :math:`u_k` of the solution at the previous iteration;
+the current value :math:`u` to be solved for; and a test function.
+
+::
+
+  class AllenCahnAuxSNES(firedrake.AuxiliaryOperatorSNES):
+      def form(self, snes, u_k, u, v):
+          F, bcs = super().form(snes, u_k, u, v)
+          return (eps * inner(grad(u), grad(v)) + (u**3 - u_k) * v) * dx, bcs
```
pbrubeck: Please correct me if I'm wrong, but "lagging" the negative term here is equivalent to dropping it from `G`.

Also having `G(u^k, u^{k+1}) = F(u^k)` when `u^{k+1} = u^k` gives you no extra source term (`b`).

Maybe it'd be worth adding a comment.

demos/nonlinear_pc/nonlinear_pc_allen_cahn.py.rst
```diff
@@ -0,0 +1,282 @@
...
+Typically NPC is applied within a higher-level nonlinear solver. For example,
+there is a nonlinear analogue of GMRES. For a review of nonlinear solver and
+preconditioning strategies beyond Newton, see :cite:`brune2015composing`.
+
+Here we will demonstrate how to use nonlinear preconditioning for the steady-
+state Allen-Cahn equation
+
+.. math::
+
+  F(u) = -\epsilon\Delta u - u + u^3 = 0
```
pbrubeck:
```suggestion
  F(u) = -\epsilon\Delta u + u^3 - u = 0
```

demos/nonlinear_pc/nonlinear_pc_allen_cahn.py.rst
```diff
@@ -0,0 +1,282 @@
...
+like Allen-Cahn which can be derived from minimizing some free energy. Even
+with 10 iterations of a line search, Newton's method will fail. You can try
+other line search methods (secant, backtracking, etc.) and find similar
+outcomes. If you pump the number of line search iterations way up, you can
+make Newton's method converge... to the wrong solution!
+
+::
+
+  v = TestFunction(Q)
+  F = (eps * inner(grad(u), grad(v)) - (u - u**3) * v) * dx
```
pbrubeck:
```suggestion
  F = (eps * inner(grad(u), grad(v)) + (u**3 - u) * v) * dx
```

demos/nonlinear_pc/nonlinear_pc_allen_cahn.py.rst
```diff
@@ -0,0 +1,282 @@
...
+Lastly, we can define a different :math:`G_k` at each iteration by
+parameterising with the current iterate, as we will see below.
+
+Our approach for defining :math:`G` here is to hold back the linear part of
+the reaction term in the Allen-Cahn equation to the value at the previous
+iteration. In other words, at each step, we define the PDE
+
+.. math::
+
+  G(u; u_k) = -\epsilon\Delta u - u_k + u^3 = 0.
```
pbrubeck:
```suggestion
  G(u; u_k) = -\epsilon\Delta u + u^3 - u_k = 0.
```

demos/nonlinear_pc/nonlinear_pc_allen_cahn.py.rst
```diff
@@ -0,0 +1,282 @@
...
+     20 SNES Function norm 5.217958176918e-09
+     21 SNES Function norm 2.018693509573e-09
+      Nonlinear firedrake_1_ solve converged due to CONVERGED_FNORM_RELATIVE iterations 21
+
+The Allen-Cahn equation is derivable through minimization of the free energy
+functional
+
+.. math::
+
+  G(u) = \int_\Omega\left(\frac{\epsilon}{2}|\nabla u|^2 + \frac{1}{4}(1 - u^2)^2\right)dx.
```
pbrubeck: Now G(u) means a different thing. A better name would be E(u)

firedrake/preconditioners/auxiliary_snes.py
pbrubeck: We can get `Gk-f` out of `Gk1-b` (no need to generate code twice). For the moment, one needs an extra buffer `self.b1`, but this extra buffer might go away once https://github.com/firedrakeproject/firedrake/pull/5130 lands
```suggestion
        # u^{k+1} = u^{k} = x
        with self.uk.dat.vec_wo as vec:
            x.copy(vec)
        with self.uk1.dat.vec_wo as vec:
            x.copy(vec)
        # b = f
        with self.b.dat.vec_wo as vec:
            f.copy(vec)

        # b1 = G(u^{k}; u^{k}) - F(u^{k})
        self.solver._ctx._assemble_residual(tensor=self.b1)
        # b = b1
        self.b.assign(self.b1)
        # solve G(u^{k}; u^{k+1}) - b = 0 for u^{k+1}
        self.solver.solve()
```

demos/nonlinear_pc/nonlinear_pc_allen_cahn.py.rst
```diff
@@ -0,0 +1,282 @@
...
+We now set ``-snes_type nrichardson`` for nonlinear Richardson iterations,
+and specify the additional parameters for the nonlinear preconditioner under
+the key ``"npc"``. Firstly we need to specify our python SNES type, and then
+in the ``"aux"`` key we specify the parameters to actually solve :math:`G` -
+here we use Newton iterations.
+
+::
+
+  solver_parameters = {
+      "snes_type": "nrichardson",
```
pbrubeck, line 217: We could try anderson as well, it reduces iterations down to 11 for this example
```suggestion
      "snes_type": "anderson",
      "snes_anderson_m": 2,
```

demos/nonlinear_pc/nonlinear_pc_allen_cahn.py.rst
```diff
@@ -0,0 +1,282 @@
...
+      "snes_monitor": None,
+      "snes_converged_reason": None,
+      "npc": {
+          "snes_type": "python",
+          "snes_python_type": f"{__name__}.AllenCahnAuxSNES",
+          "aux": newton_parameters,
+      },
+  }
+
+  solver = NonlinearVariationalSolver(problem, solver_parameters=solver_parameters)
```
pbrubeck, line 227: It is unclear that the failed solve did not update the value of u. We should make it obvious that we are starting from the same initial guess
```suggestion
  u.interpolate(initial_guess)
  solver = NonlinearVariationalSolver(problem, solver_parameters=solver_parameters)
```

danshapero: I have this further up, immediately after the initial Newton solve failed.

firedrake/preconditioners/auxiliary_snes.py
danshapero: I had to puzzle over this for a moment when I was writing the demo. I agree that a different naming scheme could be beneficial. Is there a convention established elsewhere in Firedrake?

demos/nonlinear_pc/nonlinear_pc_allen_cahn.py.rst
JHopeCollins: Is that safe? Then the inner problem is mutating the outer problem, which seems impolite

demos/nonlinear_pc/nonlinear_pc_allen_cahn.py.rst
danshapero: I've changed this as you suggest. It would be a bigger change but I could instead define `E(u)` at the very beginning and work almost entirely from this, defining `F = firedrake.derivative(E, u)`.

demos/nonlinear_pc/nonlinear_pc_allen_cahn.py.rst
JHopeCollins: To your second point, this is the exactly the point about the stationary point in the paragraph discussing the full expression for the preconditioned iteration.

demos/nonlinear_pc/nonlinear_pc_allen_cahn.py.rst
pbrubeck: ah right

demos/nonlinear_pc/nonlinear_pc_allen_cahn.py.rst
danshapero: I considered this but wanted to keep it simple at first. I could add Anderson or NGMRES at the end. But I was planning another demo showing more advanced techniques using the Navier-Stokes equations in a later PR.

demos/nonlinear_pc/nonlinear_pc_allen_cahn.py.rst
pbrubeck: I mean creating a new `G` from `F`, e.g. `G = F + extra_terms`

firedrake/preconditioners/auxiliary_snes.py
```diff
@@ -0,0 +1,236 @@
...
+        u^{k+1} = u^{k} + \\delta u^{k}
+    """
+
+    _prefix = "aux_"
+
+    @PETSc.Log.EventDecorator()
+    def initialize(self, snes):
+        from firedrake import (  # circular import if this is at file level
+            NonlinearVariationalSolver, NonlinearVariationalProblem,
+            Function, TestFunction, Cofunction)
```
pbrubeck: `Function` is imported at the top of this file. Are we sure all of these give circular imports?

demos/nonlinear_pc/nonlinear_pc_allen_cahn.py.rst
pbrubeck: ufl.Forms are supposed to be immutable

pbrubeck: Also `F += extra_terms` will not modify `F` "in-place", it'd just assign a new Form to `F`.

demos/nonlinear_pc/nonlinear_pc_allen_cahn.py.rst
pbrubeck: If you do that, then the demo does not run in complex mode

demos/nonlinear_pc/nonlinear_pc_allen_cahn.py.rst
JHopeCollins: How about just a sentence along the lines of "we could improve the convergence by swapping to an accelerated SNES type such as Anderson or ngmres, which you may want to experiment with in practice/in your own application"

demos/nonlinear_pc/nonlinear_pc_allen_cahn.py.rst
danshapero: Ok, I added a note explaining that you can also come up with the form by altering `F` instead of creating it from whole cloth.

JHopeCollins: > Also F += extra_terms will not modify F "in-place", it'd just assign a new Form to F.

Really? I did not know that, and I suspect that most people also do not. Doesn't that break the contract for `+=`? Seems like it would have been better just to not implement `__iadd__` at all.

demos/nonlinear_pc/nonlinear_pc_allen_cahn.py.rst
JHopeCollins: > Please correct me if I'm wrong, but "lagging" the negative term here is equivalent to dropping it from G.

Once you expand the full `Gk1 = Gk - Fk` yes that term cancels in Gk1 and Gk. But I think it is better to include it in the definition of `G` here because it keeps the equivalence of the fixed point iteration with `G(u^{k+1}; u^{k}) = 0`.

demos/nonlinear_pc/nonlinear_pc_allen_cahn.py.rst
JHopeCollins: Is Allen-Cahn ill posed in complex arithmetic then? Surely if the derivation is not well posed then it doesn't make sense to solve the equation in complex mode. We can add the `skipcomplex` marker to the demo in the test suite.

firedrake/preconditioners/auxiliary_snes.py
JHopeCollins: Reducing the number of times we have to parse forms would be good. I don't think I follow, your suggestion is not equivalent to the current code because the new line 175 will overwrite the values of `f` we copied into `b` on line 170.
Did you mean to assemble `Gk` into `b` first, then increment `b.dat.vec += f`?

firedrake/preconditioners/auxiliary_snes.py
JHopeCollins: Good point, `Function` isn't a circular import.
@connorjward what is best practice here? I'd be tempted to just import everything locally so we aren't at the mercy of import changes elsewhere in Firedrake breaking the imports in this file.

firedrake/preconditioners/auxiliary_snes.py
pbrubeck: We want to solve `Gk1 - b = 0`, where `b = Gk - f`. Gk is just Gk1 evaluated at uk1=uk. The snes knows how to compute `Gk1 - b` for any `b` and `uk1`, so we use that to compute `Gk - f`.

JHopeCollins: Actually, `solver._ctx._assemble_residual` will already have `b` in, so won't this end up with the wrong values even with that change?

firedrake/preconditioners/auxiliary_snes.py
JHopeCollins: Fair enough I guess. `Gk1 - b` is also not the auxiliary form though. Maybe `Gk1b`? It is the auxiliary form plus a forcing term (that we happen to know we will assemble `Gk - Fk` into later).

firedrake/preconditioners/auxiliary_snes.py
```diff
@@ -0,0 +1,236 @@
...
+        uk :
+            The current iterate :math:`u^{k}`.
+        uk1 :
+            The next iterate :math:`u^{k+1}` that will be solved for.
+        test :
+            The test function.
+
+        Returns
+        -------
+        G : :class:`ufl.Form` | :class:`ufl.BaseForm`
```
pbrubeck:
```suggestion
        G : :class:`ufl.BaseForm`
```

demos/nonlinear_pc/nonlinear_pc_allen_cahn.py.rst
JHopeCollins: > Also having G(u^k, u^{k+1}) = F(u^k) when u^{k+1} = u^k gives you no extra source term (b).

Yes for this specific choice of `G` this is true. But we cannot rely on that in general because it's perfectly acceptable for it not to be true.

firedrake/preconditioners/auxiliary_snes.py
JHopeCollins: Oh, I follow now. Is it like this:

We have an assembler for `G(uk1) - b` and we need `b=G(uk) - f`

1. Assign `b = f` and `uk1 = uk`. Now the assembler will calculate `G(uk) - f`
2. Assemble `b1 = G(uk) - f`
3. Assign `b = b1` so now `b = G(uk) - f`. Now the assembler will calculate `G(uk1) - b = G(uk1) - (G(uk) - f)`
4. Solve for `uk1`.

Neat.

firedrake/preconditioners/auxiliary_snes.py
connorjward: Imports should always be at the top level if possible. If import changes elsewhere break this file then they can just fix this file at the time.

In general I find it helpful to think about there being dependencies between files. For example it makes sense for `variational_solver.py` to depend on `assemble.py` and not the other way around. And thus it is OK to have `import firedrake.assemble` inside `variational_solver.py` but not `import firedrake.variational_solver` inside `assemble.py`.

I think preconditioners shouldn't be depended on by anything in Firedrake, so it should be safe to have all imports outermost. If that isn't the case then I would consider that a bug.

firedrake/preconditioners/auxiliary_snes.py
pbrubeck: It think any meaningful name should do, as long as it is not the same one. 

The in-place subtraction makes it seem that we are mutating it, but that is not what is happening under the hood, so it'd be good to get rid of it.

firedrake/preconditioners/auxiliary_snes.py
pbrubeck: Exactly

pbrubeck: You might want to rename `b1 -> bk` and `b -> bk1` for consistency

pbrubeck: Or choose even more meaningful names

firedrake/preconditioners/auxiliary_snes.py
JHopeCollins: I've gone with `Gk1b` and added a comment for what it is.

firedrake/preconditioners/auxiliary_snes.py
JHopeCollins: No, removed.

demos/nonlinear_pc/nonlinear_pc_allen_cahn.py.rst
pbrubeck: The source term b is zero for this choice of F. But one could also have obtained G by dropping the negative term in F, in which case b is nonzero, but yet you get an identical algorithm. 

I think we should comment on the alternative choice of dropping the negative term.

pbrubeck: Also this comment would be helpful to let users know that the dependence of G on the old state is completely optional

firedrake/preconditioners/auxiliary_snes.py
```diff
@@ -0,0 +1,238 @@
...
+
+        b = Cofunction(V.dual())
+        # This is the form we will solve:
+        # G(u^{k+1}) - b = 0
+        # and we will assemble G(u^{k}) - F(u^{k}) into b.
+        Gk1b = Gk1 - b
+
+        self.b = b
+        # a buffer for intermediate values when assembling b = Gk - Fk
+        self._b_wrk
```
pbrubeck:
```suggestion
        self._b_wrk = Cofunction(V.dual())
```

firedrake/preconditioners/auxiliary_snes.py
JHopeCollins: > Is there a convention established elsewhere in Firedrake?

Not really I don't think.

In the base class the method arguments really should stay as `uk1` and `uk` because it corresponds directly to the description of the fixed point iteration in the docstrings of the method and the class.

If `uold` and `unew` would be clearer in the demo where you don't have the docstrings, only the `form` method definition, then I'm fine with that.

firedrake/preconditioners/auxiliary_snes.py
JHopeCollins: I've had a go at doing this. We'll see if it passes.

firedrake/preconditioners/auxiliary_snes.py
JHopeCollins: Oops, thanks!

firedrake/preconditioners/auxiliary_snes.py
JHopeCollins: That makes sense. Unfortunately I can't do it here:

1. `variational_solver.py` imports `slate`.
2. When the `slate` submodule is imported it imports the files for `SCPC` and `HybridizationPC`.
3. These files import things from `firedrake.preconditioners`
4. `firedrake.preconditioners.__init__` imports `auxiliary_snes.py`.

So I can't import VariationalSolver (or any file that imports slate I suspect) at the file level anywhere in `firedrake.preconditioners`.

Arguably `SCPC` and `HybridizationPC` should go in `firedrake.preonditioners`, but this starts to look like an API change so I don't want to do it in this PR because it is pointing at `release`.

connorjward: > If that isn't the case then I would consider that a bug.

Yep. I agree with your assessment. And happy for that to not be the problem of this PR.

firedrake/preconditioners/auxiliary_snes.py
```diff
@@ -0,0 +1,238 @@
...
+                "snes_type": "newtonls",
+                ...
+            }
+        }
+
+    Although using ``"npc_"`` to specify the parameters is more verbose
+    than the original, it allows for a wider variety of methods. For
+    example, by changing the outer ``"snes_type"`` to ``"anderson"``,
+    we can use preconditioned Anderson acceleration
+    (`<https://petsc.org/release/manualpages/SNES/SNESANDERSON/>`_)
```
pbrubeck: This could go in the manual. This class is not going to be instantiated ever by a user.

JHopeCollins: But it is still a user facing class. This will get rendered in the API docs

firedrake/preconditioners/auxiliary_snes.py
pbrubeck: > Is there a convention established elsewhere in Firedrake?

While single-character names are far from ideal, in Firedrake we have well established lingo. For instance, `u` in the solution or TrialFunction, `v` is the TestFunction, `F` is the residual Form, `V` is the FunctionSpace, etc. Situations where we need to distinguish between two related variables rarely arise in public API, but I can point out a few cases.

In the (non-Irksome) time-stepping demos, we define `u` and `uold` to denote the current and previous solutions.

Throughout the multigrid codebase we use `uc` and `uf` to denote the coarse and fine solutions.

pbrubeck: The multigrid literature often uses `u^{l}` and `u^{l+1}`. In latex or mathjax this works well because `l+1` is quite explicit. If instead we opted for names like `ul` and `ul1` in our code, as opposed to `uc` anf `uf`, we would be decreasing the readability of the code by itself.

pbrubeck: The issue with `uk1` is that this variable name needs further explanation, due to the conflict between `u^{k-1}` and `u^{k+1}`

firedrake/preconditioners/auxiliary_snes.py
pbrubeck: Regardless of what we decide to leave in the class docstring, we should be adding an entry for AuxiliaryOperatorSNES in this [manual section](https://www.firedrakeproject.org/preconditioning.html).

I think this level of detail could live in that other section, and we could keep a brief summary in the docstring, with a cross-reference.

We documented things in a similar way when we introduced the `quad_scheme` kwarg for `FunctionSpace` in https://github.com/firedrakeproject/firedrake/pull/4607/changes

firedrake/preconditioners/auxiliary_snes.py
```diff
@@ -0,0 +1,238 @@
...
+        self._b_wrk = Cofunction(V.dual())
+
+        self.solver = NonlinearVariationalSolver(
+            NonlinearVariationalProblem(
+                Gk1b, uk1, bcs=bcs,
+                form_compiler_parameters=ctx.fcp),
+            nullspace=ctx._nullspace,
+            transpose_nullspace=ctx._nullspace_T,
+            near_nullspace=ctx._near_nullspace,
+            appctx=ctx.appctx, options_prefix=prefix)
```
pbrubeck:
```suggestion
            appctx=ctx.appctx,
            options_prefix=prefix,
            pre_apply_bcs=ctx.pre_apply_bcs,
        )
```

firedrake/preconditioners/auxiliary_snes.py
```diff
@@ -0,0 +1,238 @@
...
+from ufl import replace
+from firedrake.dmhooks import get_function_space, get_appctx as get_dm_appctx
+
+__all__ = ("AuxiliaryOperatorSNES",)
+
+
+class AuxiliaryOperatorSNES(SNESBase):
+    """
+    Solve a residual form :math:`F(u) = 0` using a nonlinear Richardson
+    iteration preconditioned with an auxiliary form :math:`G(u)`.
```
pbrubeck: In the implementation, `G` depends on both `u^{k+1}` and `u^{k}`. This dependence should be made evident in the docstring.

pbrubeck: There is one note in `form` that explains this, but I think the main mathematical description in this docstring should at least suggest the possibility of G depending on the old state.

pbrubeck: The demo refers to `G` as `G_{k}(u) = G(u; u_{k})`, should we adopt that convention here? It'd be good to unify notation.

demos/nonlinear_pc/nonlinear_pc_allen_cahn.py.rst
```diff
@@ -0,0 +1,284 @@
...
+like Allen-Cahn which can be derived from minimizing some free energy. Even
+with 10 iterations of a line search, Newton's method will fail. You can try
+other line search methods (secant, backtracking, etc.) and find similar
+outcomes. If you pump the number of line search iterations way up, you can
+make Newton's method converge... to the wrong solution!
+
+::
+
+  v = TestFunction(Q)
+  F = (eps * inner(grad(u), grad(v)) + (u**3 - u) * v) * dx
```
pbrubeck: Now we don't test for complex by default. This will break in complex mode
```suggestion
  F = (eps * inner(grad(u), grad(v)) + inner(u**3 - u, v)) * dx
```

demos/nonlinear_pc/nonlinear_pc_allen_cahn.py.rst
```diff
@@ -0,0 +1,284 @@
...
+The arguments that it takes are first the PETSc SNES object,
+then the value :math:`u_k` of the solution at the previous iteration;
+the current value :math:`u` to be solved for; and a test function.
+
+::
+
+  class AllenCahnAuxSNES(firedrake.AuxiliaryOperatorSNES):
+      def form(self, snes, u_k, u, v):
+          F, bcs = super().form(snes, u_k, u, v)
+          return (eps * inner(grad(u), grad(v)) + (u**3 - u_k) * v) * dx, bcs
```
pbrubeck: Same complex issue
```suggestion
          return (eps * inner(grad(u), grad(v)) + inner(u**3 - u_k, v)) * dx, bcs
```

pbrubeck:
Just a thought. What is so special about G depending on the current state? It might as well depend on any other coefficient in the original problem. Should we instead of having the `form(snes, uk, uk1, v)` signature drop `uk` and promote grabbing the current state as `uk = appctx["state"]`? This aligns with the mechanism we have to access the full state for `AssembledPC` within a fieldsplit.

demos/nonlinear_pc/nonlinear_pc_allen_cahn.py.rst
danshapero: I think explaining at this level of detail will only confuse someone who's new to NPC. For that matter, I'd rather change some of the text that Josh already added to the demo back to considering only the particular case where `G(u_k; u_k) = F(u_k)`. In that case the right-hand side of the fixed-point iteration cancels and you end up solving just `G(u; u_k) = 0`. This is often sufficient to obtain a good NPC and for an introduction I think that's more important to communicate than anything else. You can show people later that it isn't a necessary condition. The full details are already in the docstring for AuxOpSNES.

demos/nonlinear_pc/nonlinear_pc_allen_cahn.py.rst
danshapero: [Lev Landau has entered the chat](https://doi.org/10.1103/RevModPhys.74.99). Is it a small change to the algebra in the form to make this work in complex mode? If that's more involved then I say we punt on complex mode for later but we do want to get to that eventually.

demos/nonlinear_pc/nonlinear_pc_allen_cahn.py.rst
pbrubeck: Sure, I'm not too fussy about this. You might have a point, some people might get confused if we casually mention that dropping the term that depends on the old state from both sides of the equation gives an equivalent nonlinear system

demos/nonlinear_pc/nonlinear_pc_allen_cahn.py.rst
danshapero: Done and I added a bit more to mention that NGMRES has more choices to make.

demos/nonlinear_pc/nonlinear_pc_allen_cahn.py.rst
pbrubeck: It's not that the PDE is not well posed in complex arithmetic, when we symbolically differentiate `inner(u,u)*dx` twice to get the Jacobian we get `inner(test, trial)*dx + inner(trial, test)*dx`, and the form assembler requires only the test function to be conjugated.

danshapero:
> Just a thought. What is so special about G depending on the current state? It might as well depend on any other coefficient in the original problem. Should we instead of having the `form(snes, uk, uk1, v)` signature drop `uk` and promote grabbing the current state as `uk = appctx["state"]`? This aligns with the mechanism we have to access the full state for `AssembledPC` within a fieldsplit.

Josh and I discussed this the other day. Just about every NPC scheme uses the previous iterate, so by getting it from the appctx instead of baking it into the function signature, we'd be making every use case more complicated. On top of that, getting `uk` from the appctx would give us the system state at the beginning of the solve, which isn't the same as the previous iterate.

demos/nonlinear_pc/nonlinear_pc_allen_cahn.py.rst
danshapero: I made the necessary changes so that this would work in complex mode. But the complex form of this equation as such isn't particularly useful and we really want to resolve #5141 .

firedrake/preconditioners/auxiliary_snes.py
danshapero: I've changed this to use `u` for the new guess to be computed and `u_k` for the current guess in the code, docstrings, and demo.

firedrake/preconditioners/auxiliary_snes.py
danshapero: I added a manual entry along the lines of what's already there for `AuxiliaryOperatorPC`.

firedrake/preconditioners/auxiliary_snes.py
danshapero: I've changed this instance and others to consistently use `G(u_{k + 1}; u_{k})` in all the math.

danshapero:
I think I've made all the needed changes, this is ready for another review @pbrubeck

firedrake/preconditioners/auxiliary_snes.py
pbrubeck: Thanks. @JHopeCollins does it make sense to move these examples to the manual?

firedrake/preconditioners/auxiliary_snes.py
```diff
@@ -0,0 +1,233 @@
...
+        to :class:`.AssembledPC`, in that it can be used to specify
+        an alternative ``snes_type`` for solving the same residual form
+        as the outer ``SNES``.
+
+        Parameters
+        ----------
+        snes : petsc4py.PETSc.SNES
+            The PETSc nonlinear solver object.
+        u_k :
+            The current iterate :math:`u_{k}`.
```
pbrubeck, line 174: I think we can come up with a more memorable name for the current iterate.

I agree with calling `u` the state being solved for, and I think this is very familiar to firedrake users.

`u_k` does not carry much information, especially now that we made the subscript disappear from `u_{k+1}` . How about `uold`, `ucur`, or `ulag`?

JHopeCollins: I do not like `uold` because it is too easy for people to mistake it as the solution at the beginning of the nonlinear solve. Similarly for `ulag` and `ucur`.

The docstrings explicitly write the iteration in terms of the index `k` so I don't really understand why any of those other suggestions are better. Surely you want to be able to read the docstring and then read the signature and see the same names for things?

firedrake/preconditioners/auxiliary_snes.py
JHopeCollins: I am planning on doing an editing sweep of the stuff now in the manual vs docstring vs demo after lunch. I do agree that more from the docstring can/should go into the manual.

I will also explain a bit about the two cases of `G` vs `F`: 

1. `G(u, u) = F(u)`, i.e. all you're really doing is freezing some terms/coefficients at the last iteration. In which case the Richardson iteration just becomes `G(u, u_k)=0` (although you can't write it like this in the implementation otherwise you get the wrong thing for Anderson etc I think(??)).
2. `G(u, u) =/= F(u)`, i.e. you added or removed some terms, for example you might add a diffusion term in `G` to stabilise the inner solve (but that obviously disappears at convergence).

firedrake/preconditioners/auxiliary_snes.py
pbrubeck: My suggestion is just to come up with a more memorable name. It'd be easier if variable names are clear enough so that the meaning then carries over to other codes that import the same names but not the docstrings

danshapero: I'd agree on coming up with a more memorable name if we were asking people to override a method with keyword arguments but this one is positional. I'd sooner stick with `u_k`.

firedrake/preconditioners/auxiliary_snes.py
JHopeCollins: Dan changed it to `Gb` which I think is ok seeing as this object is just `G-b`, and it just happens that we will make `b=Gk-Fk`.

demos/nonlinear_pc/nonlinear_pc_allen_cahn.py.rst
JHopeCollins: I've added a bit to the manual page explaining what happens if `G(u; u)=F(u)`.

firedrake/preconditioners/auxiliary_snes.py
JHopeCollins: > Thanks. @JHopeCollins does it make sense to move these examples to the manual?

I've moved most of this to the manual, just leaving the first simplest example of the parameters in the class docstring, and put a link to the manual page in the docstring.

dham (approved):

PR #5121, opened 2026-05-18
Enable setting the base for a PR using labels

connorjward:
This PR lets us set the base branch to `main` or `release` (and therefore use a dev/release build) using labels. This is useful for cases where you have dependent PRs and therefore the base branch may not be `main` or `release`.

### DIFF
--- a/.github/workflows/pr.yml
+++ b/.github/workflows/pr.yml
@@ -2,7 +2,7 @@ name: Test pull request
 
 on:
   pull_request:
-    types: [ opened, synchronize, reopened, labeled ]
+    types: [ opened, synchronize, reopened, labeled, unlabeled ]
 
 jobs:
   test:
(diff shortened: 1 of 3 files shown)


.github/workflows/pr.yml
```diff
@@ -2,14 +2,23 @@ name: Test pull request
 
 on:
   pull_request:
-    types: [ opened, synchronize, reopened, labeled ]
+    types: [ opened, synchronize, reopened, labeled, unlabeled ]
```
JHopeCollins, line 5: Why do we need this now but we didn't before?

JHopeCollins:
Just to check that I understand. Is this for the case where I have this branch structure:
```
base -----------------> base
    \                  /
     branch1 -> branch2
```
and I have a PR to merge `branch1` into `branch2`?

Currently we look at the target branch to determine whether to build firedrake using `release` or `main` process, but if the target is `branch2` then we're stumped.
After this PR I will be able to use a label to say "I know my target is `branch2` but this PR is branched from/will eventually end up in `release` (or `main`) so use the corresponding build process in CI".

Is that correct?

.github/workflows/pr.yml
connorjward: I want the workflow to be triggered whenever the labels _change_, not just when a new label is added. I ran into this because I was switching out the labels to test this and found that it wasn't being rerun as expected.

connorjward:
> Just to check that I understand. Is this for the case where I have this branch structure:
> 
> ```
> base -----------------> base
>     \                  /
>      branch1 -> branch2
> ```
> 
> and I have a PR to merge `branch1` into `branch2`?
> 
> Currently we look at the target branch to determine whether to build firedrake using `release` or `main` process, but if the target is `branch2` then we're stumped. After this PR I will be able to use a label to say "I know my target is `branch2` but this PR is branched from/will eventually end up in `release` (or `main`) so use the corresponding build process in CI".
> 
> Is that correct?

Yep spot on.

JHopeCollins (approved):

PR #5127, opened 2026-05-19
DirichletBC using data from other spaces + adjoint + multigrid support

pbrubeck:
## Description

`DirichletBC` now accepts boundary data as a `Function` on a *different* `FunctionSpace`, not only
composite expressions involving one — e.g. `DirichletBC(V1, Function(V2), sub)` now works, matching
the composite case (`DirichletBC(V1, 2*Function(V2), sub)`) that already did.

This is needed so `DirichletBC`s can be refined *symbolically* rather than numerically in an adaptively
refined `MeshHierarchy` (#5213): symbolic and numeric coarsening agree when spaces are nested, but not
under refinement, where numeric coarsening would keep applying the coarse BC throughout the adaptive
loop and accumulate large errors. Since the symbolic boundary data can live on a different space than
the BC itself, `BCBase.coefficients()` (and `EquationBC`/`EquationBCSplit` equivalents) is added so DM
hooks can be attached to every coefficient's own function space and the right `TransferManager` found.

Differentiability under `pyadjoint` comes from composition with the taped operations rather than new
adjoint code:
- The internal interpolation of the boundary data onto the BC's space is routed through the annotated
  `assemble()`, so it is recorded as its own block on the tape; likewise the fallback for elements
  without interpolation support (e.g. HCT) is routed through the now annotated `Projector`.
- `DirichletBCBlock` depends on the `Function` the BC actually applies (the interpolation output),
  not the raw user-supplied value, and the interpolation is re-taped whenever the BC is added as a
  dependency of another block (`_ad_will_add_as_dependency`), so the dependency always points at the
  current boundary data. The adjoint, TLM, and Hessian then flow through the interpolation's or
  projection's own block, which removed `DirichletBCBlock`'s hand-rolled interpolation pullback and
  its dead constant-handling machinery.
- Reusing a `DirichletBC` (`set_value` or in-place mutation of the data in a time loop, with a cached
  `NonlinearVariationalSolver`) is differentiable too, not just a freshly-constructed BC each step:
  each solve records its own interpolation + BC chain, and `NonlinearVariationalSolveBlock` resyncs
  its cached solver's BCs per recompute instead of reusing whatever the shared, live BC currently
  holds.

Added regression tests with Taylor tests: the time-dependent cross-space BC parametrized over both
configurations above, and a projected (HCT) boundary value. The `AGENTS.md` pyadjoint notes are
rewritten as the composition-first design principles that came out of this work.


connorjward:
> Since the symbolic boundary data might live on different function spaces, we need to add BCBase.coefficients() so that we can attach DM hooks to the function spaces of every coefficient in the problem and be able to extract the correct TransferManager of the problem.

Not for this PR but I think that this needs cleaning up. This dmhooks business is getting quite out of hand. I plan on doing this at some point.

> This PR also enables prescribing BC data from a different FunctionSpace. Previously we were not allowing the data to be a pure Function on a different space, e.g. DirichletBC(V1, Function(V2), sub), but we allowed composite expressions with Functions on a different space, e.g. DirichletBC(V1, 2*Function(V2), sub). Here we allow both cases.

Seems reasonable. This needs to be written down somewhere though. Both in a docstring and the manual.

firedrake/bcs.py
pbrubeck:
```suggestion
        ``V``, or a :class:`.Function` on a different space or UFL expression that can be interpolated into
```

connorjward (changes requested):
Adjoint test please. Control must be a Function in an expression in a BC.

Then definitely qualifies for `release`.

connorjward (changes requested):
Adjoint test please. Control must be a Function in an expression in a BC.

Then definitely qualifies for `release`.

firedrake/adjoint_utils/blocks/dirichlet_bc.py
```diff
@@ -82,17 +82,32 @@ def evaluate_adj_component(self, inputs, adj_inputs, block_variable, idx,
...
                     r[:] = output
             elif isinstance(c, firedrake.Function):
-                # TODO: This gets a little complicated.
-                #       The function may belong to a different space,
-                #       and with `Function.set_allow_extrapolation(True)`
-                #       you can even use the Function outside its domain.
-                # For now we will just assume the FunctionSpace is the same for
-                # the BC and the Function.
+                # TODO: With `Function.set_allow_extrapolation(True)` you can
+                #       even use the Function outside its domain.
```
connorjward: What is this? Can't find it in the codebase

firedrake/adjoint_utils/blocks/dirichlet_bc.py
```diff
@@ -82,17 +82,32 @@ def evaluate_adj_component(self, inputs, adj_inputs, block_variable, idx,
                                                 len(output))
```
connorjward: Well this should definitely go

firedrake/adjoint_utils/blocks/dirichlet_bc.py
```diff
@@ -82,17 +82,32 @@ def evaluate_adj_component(self, inputs, adj_inputs, block_variable, idx,
...
+                    # The BC data lives on a different FunctionSpace than
+                    # the one it is applied to, and was interpolated into
+                    # it (see `DirichletBC.function_arg`). Transform the
+                    # adjoint value through the adjoint of that (linear)
+                    # interpolation operator.
+                    for idx in bc._indices:
+                        adj_value = adj_value.sub(idx)
+                    dc = firedrake.TrialFunction(c.function_space())
+                    dform = firedrake.derivative(
+                        firedrake.interpolate(c, adj_value.function_space()), c, dc
```
connorjward: I would have expected this logic to live inside the `DirichletBC` class. Why is a branch necessary here?

firedrake/adjoint_utils/blocks/dirichlet_bc.py
pbrubeck: No idea, the old TODO mentioned it

firedrake/adjoint_utils/blocks/dirichlet_bc.py
pbrubeck: Agree, but not this PR's business.

firedrake/adjoint_utils/blocks/dirichlet_bc.py
pbrubeck: Because the interpolation is not the identity. Here we need to encode the adjoint of Interpolate

AGENTS.md
```diff
@@ -166,6 +190,43 @@ toolchain:
...
+  `Function` on a `"R"` (Real) space instead (`Function(FunctionSpace(mesh, "R", 0), val=...)`),
+  matching the idiom used throughout `tests/firedrake/adjoint/`; a `Function` is a genuine `Coefficient`
+  and is picked up correctly.
+* **Mutating an already-constructed object does not re-tape it:** `DirichletBC` is a pyadjoint
+  `FloatingType` whose single dependency is fixed once, at `__init__` (the `g` argument). Later calling
+  `bc.set_value(...)` or mutating `g` in place (e.g. `g.interpolate(new_expr)`) changes the BC's
+  *numeric* value — forward runs stay correct — but does **not** register a new tape dependency on that
+  mutation, so the adjoint never sees it. For time-dependent BC data (or any repeatedly-mutated
+  dependency), construct a fresh object (a new `DirichletBC` over a new `Function`) at each step, so
+  each step gets its own `Block` correctly wired to that step's data.
```
pbrubeck: Does this mean that it could not fix the case where BCs are reused?

tests/firedrake/adjoint/test_solving.py
```diff
@@ -288,6 +288,44 @@ def J(f):
...
+        u = Function(V1)
+        u_old = Function(V1)
+        v = TestFunction(V1)
+
+        F = inner(u - u_old, v)*dx + dt*inner(grad(u), grad(v))*dx
+
+        for i in range(nsteps):
+            t = (i + 1)*dt
+            g = Function(V2).interpolate(amp*amp*sin(2*pi*t)*(x + y))
+            bc = DirichletBC(V1, g, "on_boundary")
```
pbrubeck: Do not recreate fresh new DirichletBCs inside the time loop, I think you had to do this because you  wanted to avoid some other bug. Write a parametrization of this test with the current configuation (solve() + fresh DirichletBC inside the loop) and a reusable configuration (NonlinearVariationalSolver + DirichletBC outside the loop).

firedrake/adjoint_utils/blocks/dirichlet_bc.py
connorjward: But I thought that we were taping the interpolate now and thus this shouldn't be needed? The interpolate is a different block to this one.

firedrake/adjoint_utils/blocks/dirichlet_bc.py
connorjward: So remove it then. "Document The Present, Not The Past"

firedrake/adjoint_utils/blocks/dirichlet_bc.py
pbrubeck: Good point. I'll investigate

pbrubeck: DirichletBC is taped and as far as I understand it is treated as a terminal/block, and maybe that it cannot automatically do the adjoint on other nested blocks

firedrake/bcs.py
```diff
@@ -355,17 +363,18 @@ def function_arg(self, g):
...
             try:
-                self._function_arg = firedrake.Function(V)
-                interpolator = firedrake.get_interpolator(firedrake.interpolate(g, V))
+                interp = firedrake.interpolate(g, V)
                 # Call this here to check if the element supports interpolation
                 # TODO: It's probably better to have a more explicit way of checking this
-                interpolator._get_callable()
-                self._function_arg_update = partial(interpolator.assemble, tensor=self._function_arg)
-            except (NotImplementedError, AttributeError):
+                self._function_arg_update = partial(firedrake.assemble, interp, tensor=self._function_arg)
```
pbrubeck, line 367: Aren't we taping here already? Why do we need to rewrite the adjoint for interpolate again inside `DirichletBCBlock`?

firedrake/adjoint_utils/blocks/dirichlet_bc.py
pbrubeck:
```suggestion
```

firedrake/adjoint_utils/blocks/dirichlet_bc.py
pbrubeck: Confirmed by tracing the tape directly (dumping each block's dependency/output object identities, not just types): `DirichletBCBlock`'s only dependency is `g` — the raw `args[1]` passed to `DirichletBC.__init__` — not `bc._function_arg` (the interpolate's output). So even though the interpolate call in `bcs.py` is its own annotated `AssembleBlock`, it's a *sibling* of `DirichletBCBlock` on the tape (both depend on `g`), not something `DirichletBCBlock` depends on.

Backward differentiation only walks edges that actually exist: `J ← SolveVarFormBlock ← DirichletBCBlock ← g`. The interpolate block is a dead end — nothing downstream ever asks for its adjoint, so its own (correct) adjoint rule is never invoked. `DirichletBCBlock` treats "construct a BC from `g`" as one opaque step and has to encode the *entire* map from `g` to the BC's usable value itself.

So taping the interpolate in `bcs.py` gives us the forward-pass caching benefit (reusing `Interpolate`'s cached `Interpolator`, the original goal) — it doesn't make this branch redundant. Correctness of the gradient here comes entirely from this hand-rolled adjoint-of-interpolate branch.

firedrake/bcs.py
pbrubeck: Yes, it's taped — confirmed via `tape.get_blocks()` that this call is its own `AssembleBlock`, correctly depending on `g`. But `DirichletBCBlock` (built in `DirichletBC.__init__`, i.e. before this line ever runs) depends on `g` directly too, not on this call's output (`bc._function_arg`) — they're siblings on the tape, not a chain. So nothing downstream of this interpolate call ever asks for its adjoint; `DirichletBCBlock` has to rederive it from scratch (see the branch in `dirichlet_bc.py`, reply above with the full trace).

So this call's job is solely the forward-pass caching benefit described in the issue ("reuse the cached `Interpolate.Interpolator`") — it's orthogonal to differentiability. The gradient's correctness comes entirely from the hand-rolled adjoint-of-interpolate branch in `DirichletBCBlock`, independent of whether this line is annotated at all.

pbrubeck: Ok, so they would only automatically compose in the non-nested case, e.g. 
```
g1 = assemble(interpolate(g2, V1))
bc = DirichletBC(V1, g1, "on_boundary")
```
But this freezes g1 in the forward run, and we would not be able to update bc if g2 changes. Therefore the use of assemble(interpolate(...)) cannot just give us the adjoint for free.

tests/firedrake/adjoint/test_solving.py
```diff
@@ -288,6 +288,61 @@ def J(f):
...
+def test_time_dependent_dirichlet_bc_from_other_space(reuse_bc):
+    """DirichletBC data on a different FunctionSpace than the one it is
+    applied to is interpolated internally; that interpolation must be
+    taped by pyadjoint like any other assembly, or the Taylor test fails.
+
+    Parametrized over whether the DirichletBC (and its solver) is
+    reconstructed fresh at each timestep (``reuse_bc=False``), or
+    constructed once and reused via ``NonlinearVariationalSolver`` +
+    ``Function.interpolate`` on the BC data (``reuse_bc=True``).
+    """
```
pbrubeck:
```suggestion
```

AGENTS.md
```diff
@@ -166,6 +190,73 @@ toolchain:
...
+  (`_ad_solver_replace_forms()`) refreshes the *form* coefficients from each recompute's own checkpoint,
+  but a `_forward_solve` override that forgets to also resync other embedded, non-form state (here, the
+  shared solver's own `DirichletBC` objects) will keep solving with whatever that live, mutable state
+  currently holds — not the value checkpointed for the specific step being recomputed. The giveaway:
+  forward values are correct in the *original*, uncached run (nothing is stale yet), but recomputation
+  under a perturbed control (exactly what `taylor_test` does) silently reuses the wrong per-step data.
+  When a class introduces a performance cache/clone of anything, enumerate everything that varies
+  between the cases it will be reused for, and check the refresh path covers all of it, not just
+  whichever piece happens to be visible in the form/expression being assembled.
+* **A verification check that "passes too perfectly," or fails with a numerically degenerate error
```
pbrubeck: I think this is a very useful point for pyadjoint debugging. I originally told it to write a linear ReducedFunctional, and it correctly decided to make it nonlinear

firedrake/bcs.py
```diff
@@ -337,16 +340,21 @@ def reconstruct(self, field=None, V=None, g=None, sub_domain=None, use_split=Fal
     @function_arg.setter
     def function_arg(self, g):
         '''Set the value of this boundary condition.'''
+        # Keep the pyadjoint FloatingType dependency in sync with the value
+        # actually assigned, so that reusing this DirichletBC (e.g. calling
+        # set_value in a time loop) tapes a fresh DirichletBCBlock against
+        # the current g, rather than the one originally passed to __init__.
+        self._ad_args = (self._ad_args[0], g, self._ad_args[2])
```
pbrubeck: This does not seem the right place to set `self._ad_args`. Could this be moved to adjoint or adjoint_utils?

firedrake/bcs.py
pbrubeck: Correct, and worth being precise about *why* each half holds:

**The compose-automatically claim:** in your example `g1`'s space matches `bc`'s space exactly, so `DirichletBCBlock` takes the *same-space* branch — no hand-rolled adjoint needed. The chain `J ← SolveVarFormBlock ← DirichletBCBlock ← g1 ← AssembleBlock(interpolate) ← g2` really is a chain this time (g1 is both the interpolate's output *and* `DirichletBCBlock`'s dependency), so backward differentiation composes it automatically via the generic `AssembleBlock` adjoint. This is the one case where "we're already taping it" genuinely is enough — because the interpolation lives *outside* `DirichletBC` as its own top-level step, not folded into `DirichletBC.__init__`'s one opaque dependency edge.

**The freezing problem:** yes — `assemble(interpolate(g2, V1), tensor=g1)` computes g1 once; nothing re-runs it when `g2` changes later, on the tape or off it. You'd have to manually re-`assemble` (and re-tape) every time. `DirichletBC`'s own cross-space support exists precisely to avoid making the user do that — `bc.function_arg` re-interpolates lazily on every access. But that laziness is exactly what forces `DirichletBCBlock` to own the full `g → BC` derivative itself, since there's no longer a separate, chainable "interpolate" block sitting between `g` and `DirichletBCBlock` in the dependency graph — `DirichletBCBlock` *is* that step, just not annotated as one.

So: automatic composition and automatic (lazy) freshness are in tension here, and `DirichletBC` picked freshness — which is why this branch has to exist.

firedrake/bcs.py
pbrubeck: Agreed, moved. Added `DirichletBCMixin._ad_update_value(self, g)` in \`firedrake/adjoint_utils/dirichletbc.py\`, and \`bcs.py\`'s \`function_arg\` setter now just calls \`self._ad_update_value(g)\` instead of touching \`self._ad_args\` directly — keeps the pyadjoint-internal detail out of the plain Firedrake file, matching how \`_ad_annotate_init\`/\`_ad_annotate_apply\` are the only other touchpoints \`bcs.py\` uses.

pbrubeck: Refined further — went with the actual standard pattern instead (matching \`__init__\`/\`apply\`, and \`ConstantMixin._ad_annotate_assign\` on \`Constant.assign\`): \`DirichletBCMixin._ad_annotate_function_arg\` is now a decorator wrapping the whole setter, applied as \`@function_arg.setter \n @DirichletBCMixin._ad_annotate_function_arg\`. \`bcs.py\`'s setter body no longer references anything \`_ad_\`-prefixed at all — the sync happens entirely in the wrapper, after calling through to the real setter.

tests/firedrake/adjoint/test_solving.py
```diff
@@ -288,6 +288,60 @@ def J(f):
...
+@pytest.mark.parametrize("reuse_bc", [False, True])
+def test_time_dependent_dirichlet_bc_from_other_space(reuse_bc):
+    """DirichletBC data on a different FunctionSpace than the one it is
+    applied to is interpolated internally, and must remain differentiable.
+
+    Parametrized over whether the DirichletBC (and its solver) is
+    reconstructed fresh at each timestep (``reuse_bc=False``), or
+    constructed once and reused via ``NonlinearVariationalSolver`` +
+    ``Function.interpolate`` on the BC data (``reuse_bc=True``).
+    """
```
pbrubeck:
```suggestion
```

AGENTS.md
pbrubeck:
```suggestion
* **Parallel deadlocks:** `PYOP2_SPMD_STRICT=1` adds barriers around calls
```

pbrubeck, line 152:
```suggestion
* **Cross-rank code-generation mismatches (niche, rarely needed):** If a parallel run raises `CompilationError: Generated code
```

AGENTS.md
```diff
@@ -140,12 +184,25 @@ toolchain:
...
+  `gh api repos/<org>/<repo>/actions/jobs/<job-id>/logs` and grep for `FAILED`. Before debugging
+  anything, fetch the failure list of the *previous* run of the same PR: the difference between the
+  two failure sets attributes each failure to the commits pushed in between.
 * **Narrow reproduction first:** Run the single failing test node (`pytest path::test_name -k ...`)
   before the full module; the suite is large and full-module reruns are slow to iterate against.
+* **Test mathematical correctness, not just that it runs or looks structurally right.** Neither "no
+  exception was raised" nor `==` agreement between two expressions proves the result is
+  correct — two independently-built expressions can match structurally while sharing the same wrong
+  derivative or simplification rule. Verify the actual mathematical claim: evaluate numerically and
+  compare against a  hand-computed or finite-difference value, or use a Taylor test for anything
```
pbrubeck:
```suggestion
  compare against a hand-computed or finite-difference value, or use a Taylor test for anything
```

firedrake/adjoint_utils/blocks/solving.py
```diff
@@ -648,6 +648,13 @@ def recompute_component(self, inputs, block_variable, idx, prepared):
 
     def _forward_solve(self, lhs, rhs, func, bcs, **kwargs):
         self._ad_solver_replace_forms()
+        # `bcs` holds the DirichletBCs reconstructed from this block's own
+        # checkpointed dependencies; `forward_nlvs` was built once (cloned
+        # from the user's problem) and reuses the same solver across every
+        # recompute, so its problem's bcs must be swapped for this block's
+        # own checkpointed ones before solving, or a reused solver would
+        # keep solving with whichever BC data was set most recently.
+        self._ad_solvers["forward_nlvs"]._problem.bcs = bcs
```
dham, line 657: I think the right approach here is that the "forward_nlvs" should have a copy of the bcs which should depend on copies of any coefficients in the BC expressions, and the coefficient values should be overwritten at this point.

dham (changes requested):

firedrake/projection.py
```diff
@@ -231,6 +231,7 @@ def residual(self):
     def rhs(self):
         pass
 
+    @ProjectorMixin._ad_annotate_project
```
connorjward, line 234: Do we not need to annotate init? I thought that that's the pattern for this sort of thing.

firedrake/adjoint_utils/blocks/solving.py
connorjward: I had the following question in the meeting:

Consider a solve like:
```py
f = Function(V)
solve(a == f * v * dx, bcs=[DirichletBC(V, f, ...)])
```

If we copy and replace the function `f` for the form and bc separately then we will get something like:
```py
solve(a == g * v * dx, bcs=[DirichletBC(V, h, ...)])
```

Is this something to be concerned about? Is it even legal?

pbrubeck: I'll add such case (f in both RHS and BC data, used as Control) as a Taylor test.

PR #5130, opened 2026-05-20
NonlinearVariationalSolver: support rhs and solution kwargs

pbrubeck:
# Description
Add a front door to the SNES right-hand side and initial solution.
```
solver = NonlinearVariationalSolver(...)

x = Function(V)
b = Cofunction(V.dual())
solver.solve(x=x, b=b)
```


firedrake/variational_solver.py
```diff
@@ -340,17 +343,30 @@ def set_transfer_manager(self, manager):
 
     @PETSc.Log.EventDecorator()
     @NonlinearVariationalSolverMixin._ad_annotate_solve
```
pbrubeck, line 345: The adjoint will certainly fail if we tape `solver.solve(x=x, b=b)`, this needs more work

angus-g:
Just flagging that #4638 would ideally remove the `GenericSolveBlock` and at least would conflict with the changes here...

firedrake/adjoint_utils/variational_solver.py
```diff
@@ -105,10 +107,13 @@ def wrapper(self, **kwargs):
...
             with stop_annotating():
-                out = solve(self, **kwargs)
+                out = solve(self, x=x, b=b, **kwargs)
 
             if annotate:
-                block.add_output(self._ad_problem._ad_u.create_block_variable())
+                # The solution is written into x when provided, so x
+                # supersedes problem.u as the block output.
+                u = x if x is not None else self._ad_problem._ad_u
+                block.add_output(u.create_block_variable())
```
JHopeCollins, line 116: `x` doesn't supersede `u` as an output, it is an additional output. `u` should always be added as an output.

pbrubeck: I will remove problem.u from the solve so that this  is the case

PR #5132, opened 2026-05-20
Fix multidomain form data

pbrubeck:
# Description

`form.ufl_domains()` is not enough to detect a multidomain problem, it only returns the domain of the `Measure`. Here we actually need to grab all domains by traversing the DAG


connorjward (approved):
That's some very confusing UFL API right there...

PR #5139, opened 2026-05-27
Mesh tolerance of 0.0 caused points to be lost

leo-collins:
Creating a VertexOnlyMesh on a mesh with a tolerance of 0.0 can cause points to be lost. E.g.

```
mesh = UnitTriangleMesh(0)
point = [[0.17, 0.17]]
vom = VertexOnlyMesh(mesh, point, tolerance=0.0)
```

would raise an error for a missing point even though (0.17,0.17) is clearly in the mesh. This happened because we check `ref_cell_dist_l1 < tolerance` to determine if a point is in a cell, and `ref_cell_dist_l1` is often very small but non-zero for points inside the cell.

The fix in this PR is to not allow a user tolerance below machine epsilon.

### DIFF
--- a/firedrake/locate.c
+++ b/firedrake/locate.c
@@ -27,10 +27,9 @@ int locate_cell(struct Function *f,
     pointers refer to is updated as necessary. */
     double ref_cell_dist_l1 = DBL_MAX;
     double current_ref_cell_dist_l1 =  -0.5;
-    /* NOTE: `tolerance`, which is used throughout this funciton, is a static
-       variable defined outside this function when putting together all the C
-       code that needs to be compiled - see pointquery_utils.py */
-
+    /* NOTE: `user_tolerance` is a static variable defined in the
+    `compile_coordinate_element` function in pointquery_utils.py. */
+    PetscReal tolerance = PetscMax(user_tolerance, PETSC_MACHINE_EPSILON);
     size_t *ids = NULL;
     size_t nids = 0;
     err = rtree_locate_all_at_point((const struct RTreeH *)f->rtree, x, &ids, &nids);
--- a/firedrake/pointquery_utils.py
+++ b/firedrake/pointquery_utils.py
@@ -249,7 +249,7 @@ struct ReferenceCoords {
     %(ScalarType)s X[%(geometric_dimension)d];
 };
 
-static %(RealType)s tolerance = %(tolerance)s; /* used in locate_cell */
+static %(RealType)s user_tolerance = %(tolerance)s; /* used in locate_cell */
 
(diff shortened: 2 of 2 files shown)


firedrake/locate.c
```diff
@@ -27,10 +27,9 @@ int locate_cell(struct Function *f,
     pointers refer to is updated as necessary. */
     double ref_cell_dist_l1 = DBL_MAX;
     double current_ref_cell_dist_l1 =  -0.5;
-    /* NOTE: `tolerance`, which is used throughout this funciton, is a static
-       variable defined outside this function when putting together all the C
-       code that needs to be compiled - see pointquery_utils.py */
-
+    /* NOTE: `user_tolerance` is a static variable defined in the
+    `compile_coordinate_element` function in pointquery_utils.py. */
+    PetscReal tolerance = PetscMax(user_tolerance, PETSC_MACHINE_EPSILON);
```
connorjward, line 32: Would it be better to use 2 x epsilon instead? I don't know all of the details I just want to know if this is definitely the right value to use

leo-collins: I'm not sure what the correct value would be. For affine meshes we can calculate the reference coords exactly (in one Newton iteration) so I think we should always be below epsilon. For higher-order meshes things are worse and our Newton tolerance is `1e-12` so I suppose we could get points above epsilon which are inside the cell.

connorjward: Instead of this approach, what do you think about instead warning the user if they try and set a tolerance that is less than `1e-8` or so? The current behaviour isn't wrong per se, just rife with footguns.

danshapero:
I don't know how the internals of point location are implemented so this might be a dumb question. Is the tolerance an absolute or relative one? For simulations of, say, microelectronic devices, inter-node distances can be on the order of 1e-6, so a tolerance of 1e-8 is large.

For affine meshes, Shewchuk's sign-exact predicates are the principled solution.

connorjward:
> I don't know how the internals of point location are implemented so this might be a dumb question. Is the tolerance an absolute or relative one? For simulations of, say, microelectronic devices, inter-node distances can be on the order of 1e-6, so a tolerance of 1e-8 is large.
> 
> For affine meshes, Shewchuk's sign-exact predicates are the principled solution.

I'm pretty sure tolerance is relative to the size of the current cell. So a tolerance of 0.5 catches points up to half a cell away.

firedrake/locate.c
leo-collins: The `mesh.tolerance` docstring also states that this can happen, so I don't think this PR is necessary.

PR #5142, opened 2026-05-28
Fix Real + interior facet for PCPatch

connorjward:
Same as https://github.com/firedrakeproject/firedrake/pull/4996 but for PCPatch (different codegen pathway).

Note that this will require a different fix on `main`.

### DIFF
--- a/tests/firedrake/regression/test_patch_pc.py
+++ b/tests/firedrake/regression/test_patch_pc.py
@@ -193,7 +193,8 @@ def test_patch_pc_real():
     L = inner(Constant(1.0), v) * dx
 
     patch_solver_parameters = {
-        "mat_type": "matfree",
+        "ksp_type": "preonly",
+        "ksp_max_it": 1,
         "pc_type": "python",
         "pc_python_type": "firedrake.PatchPC",
         "patch_pc_patch_construct_type": "star",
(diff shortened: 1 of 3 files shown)


tests/firedrake/regression/test_patch_pc.py
```diff
@@ -174,3 +174,35 @@ def test_patch_pc_exterior_facets_dx_dS_ds():
...
+    r = Function(R).assign(3)
+    # test a form with all types of integral
+    a = (
+        r * inner(u, v) * dx
+        + avg(r) * inner(avg(u), avg(v)) * dS
+        + r * inner(u, v) * ds
+    )
+    L = inner(Constant(1.0), v) * dx
+
+    patch_solver_parameters = {
```
pbrubeck, line 195: I think we should compare PCPatch against ASMStarPC with ksp_type preonly

pbrubeck (changes requested):

tests/firedrake/regression/test_patch_pc.py
pbrubeck: This is because you can get the preconditioner wrong and still converge after many iterations

connorjward: Thanks. This was definitely pushing my comfort zone coming up with these. Do you want to compare iteration counts and solution?

pbrubeck: Just take one iteration with ksp_type preonly and compare the solutions

connorjward: Done

pbrubeck (approved):

PR #5145, opened 2026-05-29
Remove add, sub magic methods from matrix

leo-collins:
Adding two matrices together eagerly computed the result, which doesn't seem very Firedrake-onic to me. By removing these we now defer to `ufl.BaseForm`s magic methods which return a symbolic `FormSum`.


connorjward (approved):
Assuming this passes tests I'm happy. I wonder why it was originally implemented like this.

pbrubeck:
> Assuming this passes tests I'm happy. I wonder why it was originally implemented like this.

Matrix came before BaseForm

PR #5148, opened 2026-06-02
fix: defer mpi initialisation

anilbeycorintis:
# Description

This change makes `import firedrake` MPI-neutral by deferring PETSc/MPI initialization until first real Firedrake use. Historically, importing Firedrake pulled in `petsc4py.PETSc`, which calls `PetscInitialize` / `MPI_Init` at import time. That means a process becomes an MPI process merely by importing the package.

That is a correctness issue for downstream users, not just an import-time cost. Clients depending on Firedrake should be able to import it in a normal Python process without immediately becoming MPI-aware. Today, import-time MPI initialization does break our packages, and it breaks other peoples' packages too: CLI entry points, web servers, `multiprocessing` or fork-based worker setups, and any library that needs to import Firedrake before deciding whether it will do parallel work. Python itself is not MPI-aware, and an already-initialized MPI runtime is not fork-safe, so forcing MPI up during import breaks consumers before they do any real computation.

With this change, plain `import firedrake` does no PETSc/MPI work. Initialization happens only on first top-level Firedrake use, or explicitly through `firedrake.initialize()`. That gives downstream applications a deterministic point to initialize MPI when they are ready.

This is not intended to change Firedrake behavior after initialization. It only changes when PETSc/MPI is brought up: not at import time, but at first real Firedrake use or at an explicit `firedrake.initialize()` call.

Safe client usage looks like this:

```python
import firedrake as fd

def build_problem():
    mesh = fd.UnitSquareMesh(8, 8)
    V = fd.FunctionSpace(mesh, "CG", 1)
    return fd.Function(V)
```

or, when explicit control is preferred:

```python
import firedrake as fd

def main():
    fd.initialize()
    mesh = fd.UnitSquareMesh(8, 8)
    ...
```

The important rule for clients is to keep `fd.X` access inside function or method bodies, rather than at module scope, so importing their own module remains MPI-neutral.

### DIFF
--- a/firedrake/__init__.py
+++ b/firedrake/__init__.py
@@ -1,3 +1,5 @@
+import functools
+
 # The range of PETSc versions supported by Firedrake. Note that unlike in
 # firedrake-configure and pyproject.toml where we want to be strict about
 # the specific version, here we are more permissive. This is to catch the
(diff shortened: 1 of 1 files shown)


connorjward:
Why is it a problem if MPI is initialised? If you're going to use Firedrake you're going to use MPI.

anilbeycorintis:
> Why is it a problem if MPI is initialised? If you're going to use Firedrake you're going to use MPI.

Yes but the question is "where" we are going to use it. Our software has different layers and firedrake is only relevant to the computation layer. We cannot have any MPI in the other layers as it's unsafe and too complex. We cannot write a CLI or a webserver or plain python modules in presence of MPI. It's a quite big blocker at the moment.

connorjward:
> > Why is it a problem if MPI is initialised? If you're going to use Firedrake you're going to use MPI.
> 
> Yes but the question is "where" we are going to use it. Our software has different layers and firedrake is only relevant to the computation layer. We cannot have any MPI in the other layers as it's unsafe and too complex. We cannot write a CLI or a webserver or plain python modules in presence of MPI. It's a quite big blocker at the moment.

Could this not be achieved by just controlling where imports happen? E.g.

```py
def my_firedrake_func():
    import firedrake
    ...
```

instead of
```py
import firedrake

def my_firedrake_func():
    ...
```

anilbeycorintis:
> Could this not be achieved by just controlling where imports happen? E.g.

Using function-local imports helps a bit, but it brings more problems.

1. If a module has N functions that use firedrake, the imports have to be repeated N times. Plus if these functions call each other again you need to repeat local imports. This forces us to write everything in a big single function.
2. Python cannot tell the dependencies of each modules so it cannot detect dependency problems like circular dependencies before running the code. When something is wrong we'll get a runtime error in the best case. In the worst case the error will go silent.
3. We cannot use type hints and static code checking anymore as python will not see the required imports.
4. The transitive behaviour still remains. Once one module imports Firedrake at the top level, anything importing that module inits MPI too, even indirectly

In more general terms using function-local import is invasive as it is interfering with how the users should design their codebases.

The deferred init however is a safe choice. Already existing firedrake behaviour and usages will remain exactly the same. The firedrake users will not even notice it. On top of that it will give us the flexibility to select "where" to initialise the computation parts.

E.g. in this image we would like our GUI, geometry, mesh and mathematics, visualisation modules independent of the computation logic.  

image

For a general-purpose library, I think there's real value in giving consumers the flexibility to design their own architecture, rather than committing them to a single prescribed style. 

One can still do `from firedrake import *` and continue working exactly as before.

But users like us, who need to keep MPI out of the other layers in their software (e.g. the GUI layer or the web server), would get a lot out of being able to control where the computation gets initialised. This I think will make firedrake less invasive, more flexible and compatible with other python packages. 

I'm very curious to hear your thoughts on it and I'd be happy to arrange a meeting if you'd like to discuss it in more detail.

Thanks

connorjward:
The code that you are proposing is very very magic so I am wary to accept it. The real issue stems from PETSc itself. Despite that I am not fundamentally against this except some of the reasons you are giving don't make sense:

> Plus if these functions call each other again you need to repeat local imports. This forces us to write everything in a big single function.

Repeating local imports is free. I don't see why you need to put everything in a single function.

> Python cannot tell the dependencies of each modules so it cannot detect dependency problems like circular dependencies before running the code. When something is wrong we'll get a runtime error in the best case. In the worst case the error will go silent.

Circular dependencies are only an issue at Python startup. You have imports inside functions precisely to avoid circular import issues. I also don't see how an issue with importing Firedrake could lead to a silent error.

> We cannot use type hints and static code checking anymore as python will not see the required imports.

You can. Use `typing.TYPE_CHECKING`.

> The transitive behaviour still remains. Once one module imports Firedrake at the top level, anything importing that module inits MPI too, even indirectly

If you are running Firedrake in your simulation you will have to initialise MPI eventually. I do not understand why this can't happen at the start.
If you have Firedrake imports in your package but don't want to use Firedrake in that specific run of the script then I think you should sandbox Firedrake via intelligent imports at your end.

anilbeycorintis:
Thanks for the reply. Let me try to lay this out more clearly, because I think the disagreement is about something more fundamental than this one PR.

Our system isn't built entirely on Firedrake. Firedrake is one Python package among many: other third-party libraries, plus our own in-house modules, each with their own assumptions about the runtime.

So independent of this PR, can we at least agree on one principle? If Firedrake presents itself as a Python package, it should respect the [Python Language Reference](https://docs.python.org/3/reference/index.html). That is, if Python the language supports some feature, syntax, or semantics, Firedrake shouldn't silently take that capability away from the people importing it.

The concrete violation is that Firedrake initialises PETSc and MPI at import time. That single decision propagates outward into everything else we run. Once MPI is up, large parts of the standard library and our other dependencies become unsafe to use.
Import time is simply the wrong moment for this. Importing a module should do one thing: make its names and types visible. It shouldn't bootstrap a parallel runtime on the user's behalf. Where and when MPI gets initialised is an architectural decision that belongs to the application, not to one of its libraries. If we could choose that point ourselves, none of the rest of this would be a problem.

And MPI in particular has to be under the application's control, because Python has no idea it exists. MPI lives entirely outside the language. Python can't see it, can't reason about it, and gives you no guarantees around it. That's exactly why a live MPI environment quietly turns ordinary, well-behaved Python into a minefield: multiprocessing and concurrent.futures fork in ways MPI doesn't tolerate, and even routine I/O or a requests call can become unsafe. Our other dependencies handle parallelism their own way too, and they can only coexist if MPI is something we get to manage rather than something imposed on us at import. A library that initialises MPI but exposes no way to control it isn't really usable in an enterprise system.
On the specific suggestions:

> Repeating local imports is free. I don't see why you need to put everything in a single function.

Asking this from the users is too much. Are we to locally import everything inside each function and hide them all from the modules? How can we declare which module depends on which other? If we want to change a module dependency are we to modify every function body one by one? Each time someone modifies a code, do they have to worry about breaking module dependencies? How can we read the code? How can we display which module depends on which other? Am I to read every function body for understanding the dependencies? I can add many more questions like this. 

Python deliberately gives the user the choice of where to import, at module top level or inside a scope, precisely because that choice is part of how you express your own architecture. Firedrake shouldn't override it.

> Circular dependencies are only an issue at Python startup. You have imports inside functions precisely to avoid circular import issues. I also don't see how an issue with importing Firedrake could lead to a silent error.

Circular dependencies detected at the startup are actually useful, very useful. They warn us about problems in our architecture. Importing inside functions is just a hack to bypass that warning. When functions do local imports, then each person modifying a function will bear the responsibility of changing the module dependencies (which we do not like to have). There are other sorts of dependency problems are well. These should be detectable at the startup, not in runtime. .

> You can. Use typing.TYPE_CHECKING.

That doesn't address it. Users may want real type hints in their actual signatures, evaluated like any other annotation:

```py
    def stabilization_form(
        temperature: fd.Function,
        tau: fd.Argument,
        ...
```

Python allows this. Firedrake breaks it, because the moment those annotations reference Firedrake types, MPI initialises. So the price of annotating a signature with a Firedrake type is losing safe access to multiprocessing, requests, concurrent.futures, and more. That's a restriction the Python language reference never imposes.



That's it, plainly: we as users are being asked to give up a large part of what Python offers, purely so that MPI initialisation can stay pinned to import and type-declaration time. Defer PETSc and MPI initialisation to first use, or give us an explicit hook to trigger it ourselves, and the conflict disappears. But until MPI and PETSc are something the application controls, I don't see how Firedrake can ever be used in an enterprise system. We as users need that control.

connorjward:
@anilbeycorintis in https://github.com/firedrakeproject/petsctools/issues/51 I think I may have stumbled across a relatively nice way of achieving this. A lot of thought would be needed over the design because it's a pretty fundamental change that if we don't get right will result in segfaults for users.

PR #5147, opened 2026-05-29
firedrake-run-split-tests: fall back when GNU parallel is missing

cpjordan:
# Description

Add a fallback path in scripts/firedrake-run-split-tests to run split pytest groups using bash job control when GNU parallel is not available, while preserving per-job logs and exit-code reporting. Update the script help text to document that GNU parallel is optional.

As discussed on https://github.com/thetisproject/thetis/pull/438.

PR #5150, opened 2026-06-03
Add outer timeout for mpi testing

### LINKED ISSUE #459: Parallel MPI testing hangs
## Summary

`firedrake-run-split-tests` can hang indefinitely even after pytest has printed a successful result for every selected test. The hang is not in the Thetis test body. It happens after pytest success, during MPI/PETSc shutdown, so the wrapper never reaches the code that records the split job's exit status. Similar behaviour also happens if a test hits the timeout limit. This can be tested using:

```
firedrake-run-split-tests 2 4 \
  --durations=0 --durations-min=60.0 \
  --timeout=600 --timeout-method=thread \
  test
```

(To re-produce failure where the timeout occurs but the processes don't end, the timeout can be reduced to e.g. 180s.)

In the reproduced case, group 2 printed:

```text
test/solver3d/test_baroclinic_mms.py::test_baroclinic_mms PASSED
========== 1 passed, 689 deselected ... ==========
```

but the job still had live processes:

```text
firedrake-run-split-tests
  parallel
    bash -c "mpiexec ... pytest ... | tee ...; echo ${PIPESTATUS[0]} > job2.errcode"
      mpiexec
        python3 -m pytest   # rank 0
        python3 -m pytest   # rank 1
      tee pytest_nprocs2_job2.log
```

`job1.errcode`, `job3.errcode`, and `job4.errcode` existed, but `job2.errcode` did not. That means the shell for group 2 never reached:

```bash
echo ${PIPESTATUS[0]} > job2.errcode
```

so GNU `parallel` kept waiting and the whole CI step remained stuck.

## Debugging

The logs showed that the selected test had passed. The process tree showed that `mpiexec`, both pytest ranks, and `tee` were still alive after pytest's success summary.

Attaching `gdb` to the two live Python ranks showed that the ranks had diverged during shutdown:

Rank 0:

```text
ompi_mpi_finalize()
```

Rank 1:

```text
PetscFinalize()
  PetscGarbageCleanup()
    GarbageKeyAllReduceIntersect_Private()
      PMPI_Allreduce()
```

This explains the deadlock. `PetscGarbageCleanup()` is collective: it performs an MPI allreduce across the communicator. One rank was still in PETSc's collective cleanup while the other rank had already entered OpenMPI finalization. The PETSc allreduce can no longer complete, so `mpiexec` never exits. Since `mpiexec` never exits, `tee` keeps waiting on the pipe and the wrapper never writes `job2.errcode`.

## Why pytest-timeout does not solve this

`--timeout` and `--timeout-method=thread` apply inside pytest. They can interrupt a long-running test item, but they do not reliably guard the whole process tree:

```text
GNU parallel -> shell -> mpiexec -> python pytest ranks -> PETSc/OpenMPI shutdown
```

This failure can happen after pytest has already passed, so pytest's item-level timeout is no longer the right boundary. The missing guard is an outer timeout around the whole split job command.

## Proposed fix

In `firedrake-run-split-tests`, wrap each split job's full `mpiexec ... pytest` command with a shell-level `timeout`, including a kill grace period:

```bash
timeout --kill-after=60s 1800s mpiexec -n 2 python3 -m pytest ...
```

inside the existing pipeline:

```bash
timeout --kill-after=${kill_after} ${outer_timeout} ${pytest_cmd} 2>&1 \
  | tee ${log_file_prefix}{#}.log
echo ${PIPESTATUS[0]} > job{#}.errcode
```

If pytest and MPI shutdown finish normally, behavior is unchanged. If pytest passes but `mpiexec` hangs during finalization, the outer timeout terminates the stuck process tree, `tee` exits, `jobN.errcode` is written, and the script fails cleanly instead of hanging forever. I think this will show as failing in CI, but it's better than the current behaviour.

This does not change the number of split jobs or the `4 x 2` MPI concurrency. It only adds a process-tree guard for stuck `mpiexec` shutdown. I've checked this locally and will add a PR to Firedrake.

cpjordan:
# Description

Fixes https://github.com/thetisproject/thetis/issues/459 as demonstrated on https://github.com/thetisproject/thetis/pull/438 (@connorjward).

This PR adds a shell-level timeout around each split job run by `firedrake-run-split-tests`.

The existing pytest timeout catches long-running pytest items, but it does not protect the whole `mpiexec ... pytest` process tree. We have observed (in Thetis - locally and in CI) MPI split jobs that print pytest success and then hang indefinitely during PETSc/OpenMPI finalization. In that state, the wrapper never writes `jobN.errcode`, GNU `parallel` never returns, and CI remains stuck.

## Change

Each split job now runs as:

```bash
timeout --kill-after=${kill_after} ${outer_timeout} ${pytest_cmd}
```

inside the existing `tee` pipeline.

The timeout settings are configurable:

```bash
FIREDRAKE_RUN_SPLIT_TESTS_TIMEOUT=1800s
FIREDRAKE_RUN_SPLIT_TESTS_KILL_AFTER=60s
```

This PR assumes GNU `timeout` is available, matching the current Linux CI environment where `firedrake-run-split-tests` is used. If this helper is later used in macOS CI, a follow-up can either install Homebrew `coreutils` and use `gtimeout`, or add a small platform-specific wrapper.

### DIFF
--- a/scripts/firedrake-run-split-tests
+++ b/scripts/firedrake-run-split-tests
@@ -14,6 +14,14 @@ Usage:
     * <njobs> is the number of different jobs
     * <pytest_args...> are additional arguments that are passed to pytest
 
+  The following environment variables can be used to configure the
+  outer process-tree timeout for each split job:
+
+    * FIREDRAKE_RUN_SPLIT_TESTS_TIMEOUT: maximum wall time for each job
+      (default: 1800s)
+    * FIREDRAKE_RUN_SPLIT_TESTS_KILL_AFTER: grace period before forcibly
+      killing a timed-out job (default: 60s)
+
 Example:
 
     firedrake-run-split-tests 3 4 tests/unit --verbose
@@ -31,7 +39,8 @@ Requires:
 
 Optional:
 
-  * GNU parallel (if unavailable, the script falls back to bash job control)"
+  * GNU parallel (if unavailable, the script falls back to bash job control)
+  * GNU timeout or gtimeout (if unavailable, jobs run without an outer timeout)"
 
 # Print out help message with no arguments or "-h" or "--help"
 if [[ "$#" -eq "0" ]] || [[ "$1" == "-h" ]] || [[ "$1" == "--help" ]]; then
@@ -42,6 +51,17 @@ fi
 num_procs=$1
 num_jobs=$2
 extra_args=${@:3}
+job_timeout=${FIREDRAKE_RUN_SPLIT_TESTS_TIMEOUT:-1800s}
+kill_after=${FIREDRAKE_RUN_SPLIT_TESTS_KILL_AFTER:-60s}
+
+timeout_exec=""
+if command -v timeout >/dev/null 2>&1; then
+    timeout_exec="timeout"
+elif command -v gtimeout >/dev/null 2>&1; then
+    timeout_exec="gtimeout"
+else
+    echo "Warning: GNU timeout/gtimeout not found; running split jobs without an outer timeout" >&2
+fi
 
 if [ $num_procs = 1 ]; then
     # Cannot use mpiexec -n 1 because this can sometimes hang with
@@ -53,6 +73,9 @@ fi
 pytest_cmd="${pytest_exec} -v \
             --splits ${num_jobs} --group {#} \
             -m parallel[match] ${extra_args}"
+if [ -n "${timeout_exec}" ]; then
+    pytest_cmd="${timeout_exec} --kill-after=${kill_after} ${job_timeout} ${pytest_cmd}"
+fi
 
 log_file_prefix="pytest_nprocs${num_procs}_job"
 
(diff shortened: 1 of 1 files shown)


scripts/firedrake-run-split-tests
```diff
@@ -39,6 +48,13 @@ fi
 num_procs=$1
 num_jobs=$2
 extra_args=${@:3}
+job_timeout=${FIREDRAKE_RUN_SPLIT_TESTS_TIMEOUT:-1800s}
+kill_after=${FIREDRAKE_RUN_SPLIT_TESTS_KILL_AFTER:-60s}
+
+if ! command -v timeout >/dev/null 2>&1; then
+    echo "GNU timeout is required" >&2
```
connorjward: In your other PR you allow for GNU parallel not being there. This should also be optional no?

connorjward (changes requested):
This should probably go into `release` so it becomes available for Thetis sooner.

scripts/firedrake-run-split-tests
connorjward: And we definitely want this to run on macOS, so making it optional would help with that if you don't want to setup `gtimeout` as you suggested.

cpjordan: That makes sense but I was waiting for #5147 to be merged (I don't have the permissions to do so).

Do you want me to make the target branch for #5147 to be release as well? Then I can update this branch for your comments.

connorjward: Yeah #5147 should also go into release. I plan to merge that ASAP

cpjordan: I've updated the base on #5147, will update this branch once merged

cpjordan: Updates are in now

connorjward (approved):

PR #5152, opened 2026-06-06
Fixes for BDDC

pbrubeck:
# Description
Internal fixes required for BDDCPC to work on primal elasticity problems.

Adds new options `-pc_bddc_use_divergence_mat` and `-pc_bddc_use_discrete_gradient`. Their defaults are the correct ones for the Riesz map in the de Rham complex, H(curl) needs the gradient, while H(div) needs the divergence. The H1 Riesz map does not need an extra operator. However, elasiticity problems are posed on H1, but BDDC would need to be aware of the divergence, so in this case the user must prescribe it.


firedrake/preconditioners/bddc.py
```diff
@@ -118,6 +118,10 @@ def initialize(self, pc):
         bddcpc.setBDDCNeumannBoundaries(neu_bndr)
 
         appctx = self.get_appctx(pc)
+        entity_dofs = V.finat_element.entity_dofs()
+        vdofs = entity_dofs[min(entity_dofs)]
+        if any(len(vdofs[v]) > 0 for v in vdofs):
+            bddcpc.setCoordinates(get_entity_coordinates(V))
```
pbrubeck, line 136: @stefanozampini when does BDDC need the coordinates? Only for H1?

stefanozampini: It usually requires coordinates to identify enough subdomain corners so that the local Neumann problem is non-singular. So I would say every finite element space that associates dofs to element vertices

pbrubeck: Alright, that is what I thought, hence the `vdofs` logic. But what about non-conforming spaces like Crouzeix-Raviart and Mardal-Tai-Winther?

firedrake/preconditioners/bddc.py
```diff
@@ -283,12 +295,17 @@ def get_divergence_mat(V, mat_type="is", allow_repeated=False):
     from firedrake import assemble
     degree = max(as_tuple(V.ufl_element().degree()))
     Q = TensorFunctionSpace(V.mesh(), "DG", 0, variant=f"integral({degree-1})", shape=V.value_shape[:-1])
-    B = tabulate_exterior_derivative(V, Q, mat_type=mat_type, allow_repeated=allow_repeated)
 
-    Jdet = JacobianDeterminant(V.mesh())
-    s = assemble(inner(TrialFunction(Q)*(1/Jdet), TestFunction(Q))*dx(degree=0), diagonal=True)
-    with s.dat.vec as svec:
-        B.diagonalScale(svec, None)
+    if mat_type == "is" and (V.ufl_element().sobolev_space == H1 or V.finat_element.complex.is_macrocell()):
```
pbrubeck:
```suggestion
    if V.finat_element.complex.is_macrocell():
```

firedrake/preconditioners/bddc.py
stefanozampini: I have never experimented with those elements. They are non-conforming, but they associate point evaluations right? So I would say each dof that is a point evaluation. Then I should probably check if the BDDC code can accepts NaNs for some of the coordinates and completely exclude the search from these dofs. This way, we can extract what we need only from point-evaluation dofs

pbrubeck: CR has point evaluations at faces, but MTW is more like BDM plus tangential moments on facets.

pbrubeck: I can always associate the dofs with entities and give you the coordinates of the entity

stefanozampini: which is what you are doing in this PR? Leave it as it is then. There's no theory on this; the reason for doing this is practical. if your subdomain problem is just a seminorm (think Poisson), you need to exclude some dofs to get an invertible local Neumann problem. And this can be done with CG elements by just identifying subdomain corners dofs. The equivalent for Hcurl and HDiv problems is probably to identify the dofs on the faces/edges that represent the lowest-order moment, and pick at least one of them per subdomain edge/face (this is what I do using coordinates). But this is just speculation and I have never experimented much with that

tests/firedrake/regression/test_bddc.py
```diff
@@ -254,6 +268,17 @@ def test_bddc_aij_simplex(rg, family, degree, cellwise):
     assert (np.diff(sqrt_kappa) <= 0.5).all(), str(sqrt_kappa)
 
 
+@pytest.mark.parametrize("family,degree,cellwise", [("GN", 1, True), ("MTW", 1, True)])
+def test_bddc_elasticity_aij_simplex(rg, family, degree, cellwise):
```
pbrubeck, line 297: @stefanozampini this test is currently failing (PCG diveges due to indefinite PC). You can run it with

```
pytest -vs tests/firedrake/regression/test_bddc.py::test_bddc_elasticity_aij_simplex -m parallel[1]
```

tests/firedrake/regression/test_bddc.py
```diff
@@ -20,16 +20,27 @@ def bddc_params(mat_type="is", cellwise=False):
         "pc_type": "python",
         "pc_python_type": "firedrake.BDDCPC",
         "bddc_cellwise": cellwise,
+        "bddc_use_discrete_gradient": use_gradient,
+        "bddc_use_divergence_mat": use_divergence,
         "bddc_pc_bddc_neumann": chol,
         "bddc_pc_bddc_dirichlet": chol,
         "bddc_pc_bddc_coarse": chol,
+        "bddc_pc_bddc_corner_selection": True,
```
stefanozampini: This should be an option, not always True

tests/firedrake/regression/test_bddc.py
```diff
@@ -131,16 +158,17 @@ def solve_riesz_map(rg, mesh, family, degree, variant, bcs, cellwise=False, cond
...
     sp.setdefault("ksp_view_singularvalues", None)
     solver = LinearVariationalSolver(problem, near_nullspace=nsp,
                                      solver_parameters=sp, appctx=appctx)
     solver.solve()
     uerr = Function(V).assign(uh - u_exact)
     assert (assemble(a(uerr, uerr)) / assemble(a(u_exact, u_exact))) ** 0.5 < rtol
 
-    ew = solver.snes.ksp.computeEigenvalues()
-    assert min(ew) >= 1.0
+    ew = solver.snes.ksp.computeEigenvalues().real
```
stefanozampini, line 178: I would use numpy.isclose

firedrake/preconditioners/bddc.py
```diff
@@ -119,17 +119,21 @@ def initialize(self, pc):
...
-            bddcpc.setCoordinates(coords.dat.data_ro.repeat(V.block_size, axis=0))
+        corner_selection = "pc_bddc_corner_selection" in opts
+        entity_dofs = V.finat_element.entity_dofs()
+        vdofs = entity_dofs[min(entity_dofs)]
+        has_vertex_dofs = any(len(vdofs[v]) > 0 for v in vdofs)
+        if corner_selection or has_vertex_dofs:
+            bddcpc.setCoordinates(get_entity_coordinates(V))
+
+        use_gradient = "use_discrete_gradient" in opts
+        use_divergence = "use_divergence_mat" in opts
```
stefanozampini: This is wrong. These options can exist and still be false. Even be  "false" as a string.

```suggestion
        use_gradient = opts.getBool("use_discrete_gradient", False)
        use_divergence = opts.getBool("use_divergence_mat", False)
```

firedrake/preconditioners/bddc.py
```diff
@@ -119,17 +119,21 @@ def initialize(self, pc):
...
 
         tdim = mesh.topological_dimension
-        if tdim >= 2 and V.finat_element.formdegree == tdim-1:
+        if use_divergence or (tdim >= 2 and V.finat_element.formdegree == tdim-1):
```
stefanozampini: Since we have such customization, should we ask the users to choose it explicitly?
```suggestion
        if use_divergence:
```

firedrake/preconditioners/bddc.py
```diff
@@ -140,7 +144,7 @@ def initialize(self, pc):
                 div_kwargs = dict()
             bddcpc.setBDDCDivergenceMat(*div_args, **div_kwargs)
 
-        elif tdim >= 3 and V.finat_element.formdegree == 1:
+        elif use_gradient or (tdim >= 3 and V.finat_element.formdegree == 1):
```
stefanozampini: Why elif?
```suggestion
        if use_gradient:
```

firedrake/preconditioners/bddc.py
stefanozampini: For now, let's just pass coordinates if requested. We figure out the edge cases later

firedrake/preconditioners/bddc.py
pbrubeck: I don't think so. I would like this wrapper class to be clever an automate as many options as possible for the simple/common cases (H(curl) and H(div))

pbrubeck: The logic should not be using or. Ideally use_gradient should default to True for an H(curl) space, and use_divergence should default to True for H(div)

firedrake/preconditioners/bddc.py
stefanozampini: Fixed

firedrake/preconditioners/bddc.py
stefanozampini: should be fixed now, take a look

tests/firedrake/regression/test_bddc.py
stefanozampini: fixed

tests/firedrake/regression/test_bddc.py
```diff
@@ -20,16 +20,31 @@ def bddc_params(mat_type="is", cellwise=False):
...
+        "bddc_use_discrete_gradient": use_gradient,
+        "bddc_use_divergence_mat": use_divergence,
         "bddc_pc_bddc_neumann": chol,
         "bddc_pc_bddc_dirichlet": chol,
         "bddc_pc_bddc_coarse": chol,
+        "bddc_pc_bddc_corner_selection": corner_selection,
+        "bddc_debug": debug,
+        "bddc_pc_bddc_use_deluxe_scaling": None,
     }
+    # On MacOSX the distributed right-hand side is bugged!
```
pbrubeck: How do we plan to properly address this?
> On MacOSX the distributed right-hand side is bugged!

tests/firedrake/regression/test_bddc.py
```diff
@@ -23,13 +23,33 @@ def bddc_params(mat_type="is", cellwise=False):
         "bddc_pc_bddc_neumann": chol,
         "bddc_pc_bddc_dirichlet": chol,
         "bddc_pc_bddc_coarse": chol,
+        "bddc_debug": debug,
+        "bddc_pc_bddc_use_deluxe_scaling": None,
```
pbrubeck: Why are we adding deluxe scaling for the non-adaptive case?

stefanozampini: my bad, I was experimenting and forgot to put back the option. Fixed now

firedrake/preconditioners/bddc.py
```diff
@@ -34,9 +34,12 @@ class BDDCPC(PCBase):
     the options:
     - ``'bddc_cellwise'`` to set up a MatIS on cellwise subdomains if P.type == python,
     - ``'bddc_matfree'`` to set up a matrix-free MatIS if A.type == python,
```
stefanozampini, line 35: Can you do matrix-free MatIS? Then the divergence matrix can be passed that way, I only need the action of its local part (in MATIS sense) on  a given vector

pbrubeck: Yeah, we can

pbrubeck: I think we need to fix support for matfree A in BDDC

pbrubeck:
@connorjward BDDC is younger than Submesh. Can this go into release? It does not break API.

firedrake/preconditioners/bddc.py
```diff
@@ -338,3 +375,95 @@ def get_primal_indices(V, primal_markers):
...
+        offsets[-1] = len(flatmap)
+        return offsets, numpy.array(flatmap, dtype=PETSc.IntType)
+
+    cg1_closure_ids = CG1.finat_element.entity_closure_dofs()
+    cg1_offsets, cg1_flatmap = flatten_entity_ids(active_entities, cg1_closure_ids)
+    v_offsets, v_flatmap = flatten_entity_ids(active_entities, V_entity_ids)
+    num_entities = len(active_entities)
+
+    kernel_code = f"""
+    void compute_entity_coordinates(PetscScalar *coords, PetscScalar *cg1_coords) {{
```
pbrubeck: Sorry about the parloop. There is no better way of computing entity coordinates on extruded meshes.

firedrake/preconditioners/bddc.py
```diff
@@ -338,3 +375,95 @@ def get_primal_indices(V, primal_markers):
...
+        for idx, (dim, ent_num) in enumerate(entities):
+            offsets[idx] = len(flatmap)
+            flatmap.extend(entity_ids[dim][ent_num])
+
+        offsets[-1] = len(flatmap)
+        return offsets, numpy.array(flatmap, dtype=PETSc.IntType)
+
+    cg1_closure_ids = CG1.finat_element.entity_closure_dofs()
+    cg1_offsets, cg1_flatmap = flatten_entity_ids(active_entities, cg1_closure_ids)
+    v_offsets, v_flatmap = flatten_entity_ids(active_entities, V_entity_ids)
```
pbrubeck: These data structures wouldn't be required if we use an op2.PermutedMap

connorjward: Why not use one then?

tests/firedrake/regression/test_bddc.py
```diff
@@ -23,13 +24,35 @@ def bddc_params(mat_type="is", cellwise=False):
         "bddc_pc_bddc_neumann": chol,
         "bddc_pc_bddc_dirichlet": chol,
         "bddc_pc_bddc_coarse": chol,
+        "bddc_debug": debug,
```
connorjward, line 28: This option is obscure and undocumented.

tests/firedrake/regression/test_bddc.py
```diff
@@ -23,13 +24,35 @@ def bddc_params(mat_type="is", cellwise=False):
...
+
+    if adaptive:
+        sp.update({
+            "bddc_pc_bddc_use_deluxe_scaling": None,
+            "bddc_pc_bddc_adaptive_userdefined": None,
+            "bddc_pc_bddc_deluxe_zerorows": False,
+            "bddc_pc_bddc_adaptive_threshold": 5,
+        })
+    # On MacOSX the distributed right-hand side is bugged!
+    if DEFAULT_DIRECT_SOLVER == "mumps":
```
connorjward: This comment does not explain the code changes. There's nothing here that's obviously specific to macOS

tests/firedrake/regression/test_bddc.py
```diff
@@ -67,6 +90,8 @@ def solver_parameters(cellwise=False, condense=False, variant=None, rtol=1E-10,
         "ksp_max_it": 20,
         "ksp_norm_type": "natural",
         "ksp_converged_reason": None,
+        # "ksp_view": None,
+        # "ksp_monitor_singular_value": None,
```
connorjward:
```suggestion
```

firedrake/preconditioners/bddc.py
connorjward: It's not an issue. Moving this to pyop3 is easy.

connorjward (changes requested):
Seems more or less fine. I think `release` is OK because this is a sufficiently new feature and you're only touching code in one file.

firedrake/preconditioners/bddc.py
```diff
@@ -33,9 +34,12 @@ class BDDCPC(PCBase):
     the options:
     - ``'bddc_cellwise'`` to set up a MatIS on cellwise subdomains if P.type == python,
     - ``'bddc_matfree'`` to set up a matrix-free MatIS if A.type == python,
+    - ``'bddc_use_discrete_gradient'`` to provide the discrete gradient.
+    - ``'bddc_use_divergence_mat'`` to provide the divergence matrix.
     - ``'bddc_pc_bddc_neumann'`` to set sub-KSPs on subdomains excluding corners,
     - ``'bddc_pc_bddc_dirichlet'`` to set sub-KSPs on subdomain interiors,
     - ``'bddc_pc_bddc_coarse'`` to set the coarse solver KSP.
+    - ``'bddc_pc_bddc_corner_selection'`` to geometrically select the subdomain corners.
```
pbrubeck:
```suggestion
    - ``'bddc_pc_bddc_corner_selection'`` to geometrically select the subdomain corners.
    - ``'bddc_pc_bddc_debug'`` to set the verbosity level of debugging for PCBDDC (by default 0).
```

tests/firedrake/regression/test_bddc.py
pbrubeck: @stefanozampini did you report this?

firedrake/preconditioners/bddc.py
pbrubeck: It only occured to me just now. Only v_flatmap is permutation, cg1_flatmap will have repeated indices because it defines the dofs supported on the closure of an entity (and edge will include its vertices), so this case is not covered by a permutation. We would still need to `flatten_entity_ids` at least for the CG1 element, so it is better to keep the code as is.

pbrubeck: Ideally we should be running a parloop on the mesh entities. I know, wait for pyop3...

tests/firedrake/regression/test_bddc.py
```diff
@@ -23,13 +24,35 @@ def bddc_params(mat_type="is", cellwise=False):
...
+        sp.update({
+            "bddc_pc_bddc_use_deluxe_scaling": None,
+            "bddc_pc_bddc_adaptive_userdefined": None,
+            "bddc_pc_bddc_deluxe_zerorows": False,
+            "bddc_pc_bddc_adaptive_threshold": 5,
+        })
+    # On MacOSX the distributed right-hand side is bugged!
+    if DEFAULT_DIRECT_SOLVER == "mumps":
+        sp.update({"bddc_pc_bddc_coarse_mat_mumps_icntl_20": 0})
+
```
pbrubeck, line 50: I think we can simply remove this from here
```suggestion
```

pbrubeck: But shouldn't this be the default?

stefanozampini: What is the problem of leaving the option in? A test should just work, and this is a known MUMPS bug

tests/firedrake/regression/test_bddc.py
stefanozampini: there's no way this can be reported officially to MUMPS people. It is a known issue on only some osx  configurations

tests/firedrake/regression/test_bddc.py
stefanozampini: Why? it is a quick way to check if something goes wrong instead of having to retype the options

tests/firedrake/regression/test_bddc.py
stefanozampini: > But shouldn't this be the default?

No, the default is distributed right-hand side, which means no need to gather the rhs on the `master` process

tests/firedrake/regression/test_bddc.py
connorjward: The comment would lead me to expect a check like: if os.system == "darwin" or similar.

pbrubeck: done

connorjward (approved):

connorjward:
This is failing in complex: https://github.com/firedrakeproject/firedrake/actions/runs/30362571041/job/90285692047

PR #5155, opened 2026-06-07
Add support for objective function in SNES

stefanozampini:
# Description

Add support for objective functions in `NonlinearVariationalProblem` and `NonlinearVariationalSolver`. Needs https://gitlab.com/petsc/petsc/-/merge_requests/9331 to be merged in PETSc

I believe there's a flaw in the `pre_apply_bcs=True` model. Setting to zero the bc contribution to the update solution using https://gitlab.com/petsc/petsc/-/merge_requests/9340 make gamg tests pass

### DIFF
--- a/docs/source/solving-interface.rst
+++ b/docs/source/solving-interface.rst
@@ -729,14 +729,24 @@ Providing an objective functional
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
 An objective functional, usually the energy,
-can be provided to PETSc's linesearch and trust-region methods.
+can be provided to PETSc's `SNES`_ linesearch and trust-region methods.
+For example, the code block below minimizes the Dirichlet energy functional
+using a trust-region method with the Steihaugh-Toint method for the
+trust-region subproblem and an initial trust region size of ``1e-4``.
+The solver monitor will report residual norm, norm of the update, and
+value of the objective functional.
 
 .. code-block:: python3
 
    E = 0.5*inner(grad(u), grad(u))*dx - inner(f, u)*dx
    F = derivative(E, u)
    # Solve an optimisation problem
-   solve(F == 0, u, ..., objective=E)
+   sp = {'snes_type': 'newtontr',
+         'snes_tr_delta0': 1e-4,
+         "snes_monitor": "::ascii_info_detail",
+         'ksp_type': 'cg',
+         'pc_type': 'gamg'}
+   solve(F == 0, u, ..., objective=E, solver_parameters=sp)
 
 Default solver options
 ~~~~~~~~~~~~~~~~~~~~~~
--- a/docs/source/solving-interface.rst
+++ b/docs/source/solving-interface.rst
@@ -739,7 +739,7 @@ value of the objective functional.
 .. code-block:: python3
 
    E = 0.5 * inner(grad(u), grad(u))*dx + kappa**2 * cosh(u)*dx - inner(f, u)*dx
-   F = inner(grad(u), grad(v))*dx + kappa**2 * inner(sinh(u), v)*dx - inner(f, v)*dx
+   F = derivative(E, u)
    # Solve an optimisation problem
    sp = {'snes_type': 'newtontr',
          'snes_tr_delta0': 1e-4,
--- a/docs/source/solving-interface.rst
+++ b/docs/source/solving-interface.rst
@@ -725,6 +725,19 @@ solving with.
    # Use the approximate inverse of Jp to precondition solves
    solve(a == L, ..., Jp=Jp)
 
+Providing an objective functional
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+An objective functional, usually the energy,
+can be provided to PETSc's linesearch and trust-region methods.
+
+.. code-block:: python3
+
+   E = 0.5*inner(grad(u), grad(u))*dx - inner(f, u)*dx
+   F = derivative(E, u)
+   # Solve an optimisation problem
+   solve(F == 0, u, ..., objective=E)
+
 Default solver options
 ~~~~~~~~~~~~~~~~~~~~~~
 
--- a/docs/source/solving-interface.rst
+++ b/docs/source/solving-interface.rst
@@ -738,15 +738,15 @@ value of the objective functional.
 
 .. code-block:: python3
 
-   E = 0.5*inner(grad(u), grad(u))*dx - inner(f, u)*dx
-   F = derivative(E, u)
+   E = 0.5 * inner(grad(u), grad(u))*dx + kappa**2 * cosh(u)*dx - inner(f, u)*dx
+   F = inner(grad(u), grad(v))*dx + kappa**2 * inner(sinh(u), v)*dx - inner(f, v)*dx
    # Solve an optimisation problem
    sp = {'snes_type': 'newtontr',
          'snes_tr_delta0': 1e-4,
          "snes_monitor": "::ascii_info_detail",
          'ksp_type': 'cg',
          'pc_type': 'gamg'}
-   solve(F == 0, u, ..., objective=E, solver_parameters=sp)
+   solve(F == 0, u, ..., objective=E, solver_parameters=sp, pre_apply_bcs=False)
 
 Default solver options
 ~~~~~~~~~~~~~~~~~~~~~~
--- a/docs/source/solving-interface.rst
+++ b/docs/source/solving-interface.rst
@@ -745,7 +745,7 @@ value of the objective functional.
          'snes_tr_delta0': 1e-4,
          "snes_monitor": "::ascii_info_detail",
          'ksp_type': 'cg',
-         'pc_type': 'gamg'}
+         'pc_type': 'hypre'}
    solve(F == 0, u, ..., objective=E, solver_parameters=sp, pre_apply_bcs=False)
 
 Default solver options
--- a/firedrake/preconditioners/patch.py
+++ b/firedrake/preconditioners/patch.py
@@ -7,7 +7,8 @@ from firedrake.preconditioners.base import PCBase, SNESBase, PCSNESBase
 from firedrake.preconditioners.asm import validate_overlap
 from firedrake.petsc import PETSc
 import firedrake.cython.patchimpl
-from firedrake.solving_utils import _SNESContext, bcdofs
+from firedrake.bcs import bcdofs
+from firedrake.solving_utils import _SNESContext
 from firedrake.utils import complex_mode
 from firedrake.dmhooks import get_appctx, push_appctx, pop_appctx
 from firedrake.interpolation import interpolate
--- a/firedrake/preconditioners/patch.py
+++ b/firedrake/preconditioners/patch.py
@@ -7,7 +7,6 @@ from firedrake.preconditioners.base import PCBase, SNESBase, PCSNESBase
 from firedrake.preconditioners.asm import validate_overlap
 from firedrake.petsc import PETSc
 import firedrake.cython.patchimpl
-from firedrake.bcs import bcdofs
 from firedrake.solving_utils import _SNESContext
 from firedrake.utils import complex_mode
 from firedrake.dmhooks import get_appctx, push_appctx, pop_appctx
@@ -23,6 +22,7 @@ import ufl
 import operator
 from functools import cached_property, partial
 import numpy
+from finat.ufl import VectorElement, MixedElement
 from tsfc.ufl_utils import extract_firedrake_constants
 import weakref
 import petsctools
@@ -585,6 +585,47 @@ def make_patch_callables(form: ufl.Form, state: Function | None) -> tuple[
             assert exterior_facet_callable is None, "Only a single exterior facet callable allowed"
             exterior_facet_callable = callable
     return cell_callable, interior_facet_callable, exterior_facet_callable
+
+
+def bcdofs(bc, ghost=True):
+    # Return the global dofs fixed by a DirichletBC
+    # in the numbering given by concatenation of all the
+    # subspaces of a mixed function space
+    Z = bc.function_space()
+    while Z.parent is not None:
+        Z = Z.parent
+
+    indices = bc._indices
+    offset = 0
+
+    for (i, idx) in enumerate(indices):
+        if isinstance(Z.ufl_element(), VectorElement):
+            offset += idx
+            assert i == len(indices)-1  # assert we're at the end of the chain
+            assert Z.sub(idx).block_size == 1
+        elif isinstance(Z.ufl_element(), MixedElement):
+            if ghost:
+                offset += sum(Z.sub(j).dof_count for j in range(idx))
+            else:
+                offset += sum(Z.sub(j).dof_dset.size * Z.sub(j).block_size for j in range(idx))
+        else:
+            raise NotImplementedError("How are you taking a .sub?")
+
+        Z = Z.sub(idx)
+
+    if Z.parent is not None and isinstance(Z.parent.ufl_element(), VectorElement):
+        bs = Z.parent.block_size
+        start = 0
+        stop = 1
+    else:
+        bs = Z.block_size
+        start = 0
+        stop = bs
+    nodes = bc.nodes
+    if not ghost:
+        nodes = nodes[nodes < Z.dof_dset.size]
+
+    return numpy.concatenate([nodes*bs + j for j in range(start, stop)]) + offset
 
 
 def select_entity(p, dm=None, exclude=None):
--- a/firedrake/preconditioners/patch.py
+++ b/firedrake/preconditioners/patch.py
@@ -7,7 +7,7 @@ from firedrake.preconditioners.base import PCBase, SNESBase, PCSNESBase
 from firedrake.preconditioners.asm import validate_overlap
 from firedrake.petsc import PETSc
 import firedrake.cython.patchimpl
-from firedrake.solving_utils import _SNESContext
+from firedrake.solving_utils import _SNESContext, bcdofs
 from firedrake.utils import complex_mode
 from firedrake.dmhooks import get_appctx, push_appctx, pop_appctx
 from firedrake.interpolation import interpolate
@@ -22,7 +22,6 @@ import ufl
 import operator
 from functools import cached_property, partial
 import numpy
-from finat.ufl import VectorElement, MixedElement
 from tsfc.ufl_utils import extract_firedrake_constants
 import weakref
 import petsctools
@@ -585,47 +584,6 @@ def make_patch_callables(form: ufl.Form, state: Function | None) -> tuple[
             assert exterior_facet_callable is None, "Only a single exterior facet callable allowed"
             exterior_facet_callable = callable
     return cell_callable, interior_facet_callable, exterior_facet_callable
-
-
-def bcdofs(bc, ghost=True):
-    # Return the global dofs fixed by a DirichletBC
-    # in the numbering given by concatenation of all the
-    # subspaces of a mixed function space
-    Z = bc.function_space()
-    while Z.parent is not None:
-        Z = Z.parent
-
-    indices = bc._indices
-    offset = 0
-
-    for (i, idx) in enumerate(indices):
-        if isinstance(Z.ufl_element(), VectorElement):
-            offset += idx
-            assert i == len(indices)-1  # assert we're at the end of the chain
-            assert Z.sub(idx).block_size == 1
-        elif isinstance(Z.ufl_element(), MixedElement):
-            if ghost:
-                offset += sum(Z.sub(j).dof_count for j in range(idx))
-            else:
-                offset += sum(Z.sub(j).dof_dset.size * Z.sub(j).block_size for j in range(idx))
-        else:
-            raise NotImplementedError("How are you taking a .sub?")
-
-        Z = Z.sub(idx)
-
-    if Z.parent is not None and isinstance(Z.parent.ufl_element(), VectorElement):
-        bs = Z.parent.block_size
-        start = 0
-        stop = 1
-    else:
-        bs = Z.block_size
-        start = 0
-        stop = bs
-    nodes = bc.nodes
-    if not ghost:
-        nodes = nodes[nodes < Z.dof_dset.size]
-
-    return numpy.concatenate([nodes*bs + j for j in range(start, stop)]) + offset
 
 
 def select_entity(p, dm=None, exclude=None):
--- a/firedrake/solving.py
+++ b/firedrake/solving.py
@@ -168,7 +168,7 @@ def _solve_varproblem(*args, **kwargs):
 
     if len(eq.lhs.arguments()) == 2:
         if objective is not None:
-            raise ValueError(f"the objective functional only makes sense for nonlinear problems.")
+            raise ValueError("The objective functional only makes sense for nonlinear problems.")
         # Create linear variational problem
         problem = vs.LinearVariationalProblem(eq.lhs, eq.rhs, u, bcs, Jp,
                                               form_compiler_parameters=form_compiler_parameters,
--- a/firedrake/solving.py
+++ b/firedrake/solving.py
@@ -149,7 +149,7 @@ def _solve_varproblem(*args, **kwargs):
     "Solve variational problem a == L or F == 0"
 
     # Extract arguments
-    eq, u, bcs, J, Jp, M, form_compiler_parameters, \
+    eq, u, bcs, J, Jp, objective, form_compiler_parameters, \
         solver_parameters, nullspace, nullspace_T, \
         near_nullspace, \
         options_prefix, restrict, pre_apply_bcs = _extract_args(*args, **kwargs)
@@ -169,6 +169,7 @@ def _solve_varproblem(*args, **kwargs):
     if len(eq.lhs.arguments()) == 2:
         # Create linear variational problem
         problem = vs.LinearVariationalProblem(eq.lhs, eq.rhs, u, bcs, Jp,
+                                              objective=objective,
                                               form_compiler_parameters=form_compiler_parameters,
                                               restrict=restrict)
         create_solver = vs.LinearVariationalSolver
@@ -177,6 +178,7 @@ def _solve_varproblem(*args, **kwargs):
         if eq.rhs != 0:
             raise ValueError(f"RHS of nonlinear Equation must be `0`, not {eq.rhs}")
         problem = vs.NonlinearVariationalProblem(eq.lhs, u, bcs, J, Jp,
+                                                 objective=objective,
                                                  form_compiler_parameters=form_compiler_parameters,
                                                  restrict=restrict)
         create_solver = vs.NonlinearVariationalSolver
@@ -201,6 +203,7 @@ def _la_solve(A, x, b, **kwargs):
     :kwarg P: an optional :class:`~.MatrixBase` to construct any
          preconditioner from; if none is supplied ``A`` is
          used to construct the preconditioner.
+    :kwarg objective: a form used for line-search.
     :kwarg solver_parameters: optional solver parameters.
     :kwarg nullspace: an optional :class:`.VectorSpaceBasis` (or
          :class:`.MixedVectorSpaceBasis`) spanning the null space of
@@ -233,7 +236,8 @@ def _la_solve(A, x, b, **kwargs):
 
         _la_solve(A, x, b, solver_parameters=parameters_dict)."""
 
-    (P, bcs, solver_parameters, nullspace, nullspace_T, near_nullspace,
+    (P, bcs, objective, solver_parameters,
+     nullspace, nullspace_T, near_nullspace,
      options_prefix, pre_apply_bcs,
      ) = _extract_linear_solver_args(A, x, b, **kwargs)
 
@@ -241,7 +245,9 @@ def _la_solve(A, x, b, **kwargs):
         raise RuntimeError("It is no longer possible to apply or change boundary conditions after assembling the matrix `A`; pass any necessary boundary conditions to `assemble` when assembling `A`.")
 
     appctx = solver_parameters.get("appctx", {})
-    solver = ls.LinearSolver(A=A, P=P, solver_parameters=solver_parameters,
+    solver = ls.LinearSolver(A=A, P=P,
+                             objective=objective,
+                             solver_parameters=solver_parameters,
                              nullspace=nullspace,
                              transpose_nullspace=nullspace_T,
                              near_nullspace=near_nullspace,
@@ -252,7 +258,7 @@ def _la_solve(A, x, b, **kwargs):
 
 
 def _extract_linear_solver_args(*args, **kwargs):
-    valid_kwargs = ["P", "bcs", "solver_parameters", "nullspace",
+    valid_kwargs = ["P", "bcs", "objective", "solver_parameters", "nullspace",
                     "transpose_nullspace", "near_nullspace", "options_prefix",
                     "pre_apply_bcs"]
     if len(args) != 3:
@@ -265,6 +271,7 @@ def _extract_linear_solver_args(*args, **kwargs):
 
     P = kwargs.get("P", None)
     bcs = kwargs.get("bcs", None)
+    objective = kwargs.get("objective", None)
     solver_parameters = kwargs.get("solver_parameters", {}) or {}
     nullspace = kwargs.get("nullspace", None)
     nullspace_T = kwargs.get("transpose_nullspace", None)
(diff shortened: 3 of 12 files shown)


firedrake/solving_utils.py
```diff
@@ -449,6 +462,27 @@ def split(self, fields):
...
+        ctx = dmhooks.get_appctx(dm)
+        # X may not be the same vector as the vec behind self._x, so
+        # copy guess in from X.
+        with ctx._x.dat.vec_wo as v:
+            X.copy(v)
+
+        if not ctx.pre_apply_bcs:
+            # Compute DirichletBC residual
+            for bc in ctx._problem.dirichlet_bcs():
+                bc.apply(ctx._bc_residual, u=ctx._x)
```
stefanozampini: Is this the correct sequence of calls to inject BCs?

firedrake/variational_solver.py
```diff
@@ -79,6 +86,7 @@ def __init__(self, F, u, bcs=None, J=None,
 
         # Use the user-provided Jacobian. If none is provided, derive
         # the Jacobian from the residual.
+        self.E = objective
```
stefanozampini: One may also inspect `F` and decide to call `derivative` twice, but requires the pde_args check to be done earlier.

firedrake/solving_utils.py
```diff
@@ -342,6 +349,12 @@ def transfer_manager(self, manager):
             raise ValueError("Must set transfer manager before first use.")
         self._transfer_manager = manager
 
+    def set_objective(self, snes):
+        if self._problem.E:
+            snes.setObjective(self.form_objective)
+        else:
+            snes.setObjective(None)
```
stefanozampini, line 356: See the PETSc MR for why this is needed

firedrake/solving_utils.py
pbrubeck: We do not need to be applying bcs if `pre_apply_bcs=False`.

If `pre_apply_bcs = True`, we just need to make sure that bcs were imposed by this point, the code you added is imposing bcs on the RHS Cofunction, which should not be referenced in the objective functional.
```suggestion
        if ctx.pre_apply_bcs:
            # Apply DirichletBC on the solution
            for bc in ctx._problem.dirichlet_bcs():
                bc.apply(ctx._x)
```

firedrake/solving_utils.py
```diff
@@ -449,6 +462,27 @@ def split(self, fields):
...
+        # copy guess in from X.
+        with ctx._x.dat.vec_wo as v:
+            X.copy(v)
+
+        if not ctx.pre_apply_bcs:
+            # Compute DirichletBC residual
+            for bc in ctx._problem.dirichlet_bcs():
+                bc.apply(ctx._bc_residual, u=ctx._x)
+
+        return ctx._assemble_objective(current_state=ctx._x)
```
pbrubeck: `current_state` is only needed for RHS assembly if we want to replace BC nodes with the BC residual `g - u`
```suggestion
        return ctx._assemble_objective()
```

firedrake/variational_solver.py
```diff
@@ -36,6 +36,11 @@ def check_pde_args(F, J, Jp):
         raise TypeError("Provided preconditioner is a '%s', not a BaseForm or Slate Tensor" % type(Jp).__name__)
     if Jp is not None and len(Jp.arguments()) != 2:
         raise ValueError("Provided preconditioner is not a bilinear form")
+    if E is not None:
+        if not isinstance(E, (ufl.BaseForm, slate.slate.TensorBase)):
+            raise TypeError("Provided objective is a '%s', not a BaseForm or Slate Tensor" % type(F).__name__)
+        if len(E.arguments()) != 0:
+            raise ValueError("Provided objective is not a linear form")
```
```suggestion
            raise ValueError("Provided objective is not a 0-form")
```

firedrake/variational_solver.py
pbrubeck: It might be worthwhile to introduce an optimiser class that just takes in E and computes F and J automatically, but that could go as a separate PR

firedrake/solving_utils.py
```diff
@@ -449,6 +462,27 @@ def split(self, fields):
...
+        # X may not be the same vector as the vec behind self._x, so
+        # copy guess in from X.
+        with ctx._x.dat.vec_wo as v:
+            X.copy(v)
+
+        if ctx.pre_apply_bcs:
+            # Apply DirichletBC on the solution
+            for bc in ctx._problem.dirichlet_bcs():
+                bc.apply(ctx._x)
+
```
pbrubeck: This should have happened already, since we do it at the beginning of the solve

```suggestion
```

stefanozampini: Since in principle you can call the function with any vector X, I think it is better to leave it here

pbrubeck: Fair enough

pbrubeck:
In order to get the energy at the coarse levels in MG/FAS we just need to add a tiny amount of logic in mg/ufl_utils.py. 

I can add it for you, if you don't mind.

stefanozampini:
> In order to get the energy at the coarse levels in MG/FAS we just need to add a tiny amount of logic in mg/ufl_utils.py.
> 
> I can add it for you, if you don't mind.

Go ahead

tests/firedrake/deflation/test_bratu.py
```diff
@@ -41,3 +41,38 @@ def test_bratu():
 
     (first, second) = deflation.roots
     assert norm(first - second) > 1
+
```
pbrubeck: wrong file, we should add `tests/firedrake/regression/test_snes_objective.py` (or similar)

firedrake/variational_solver.py
pbrubeck: The new line should not be added below the comment
```suggestion
```

firedrake/variational_solver.py
```diff
@@ -79,6 +86,7 @@ def __init__(self, F, u, bcs=None, J=None,
 
         # Use the user-provided Jacobian. If none is provided, derive
         # the Jacobian from the residual.
+        self.E = objective
         self.J = J or ufl_expr.derivative(F, u)
         self.F = F
         self.Jp = Jp
```
pbrubeck, line 91:
```suggestion
        self.Jp = Jp
        self.E = objective
```

pbrubeck:
Since the PETSc MR is targeting release, and this PR is incremental and non API-breaking, we can merge this one into our release

pbrubeck (changes requested):
This needs to be mentioned in the manual.

It'd be also good to add this to `solve(
..., objective=E)`

stefanozampini:
> This needs to be mentioned in the manual.
> 
> It'd be also good to add this to `solve( ..., objective=E)`

If you tell me where should I add this information, I'll do it

firedrake/solving.py
```diff
@@ -265,21 +271,22 @@ def _extract_linear_solver_args(*args, **kwargs):
...
 
-    return P, bcs, solver_parameters, nullspace, nullspace_T, near_nullspace, options_prefix, pre_apply_bcs
+    return P, bcs, objective, solver_parameters, nullspace, nullspace_T, near_nullspace, options_prefix, pre_apply_bcs
 
 
 def _extract_args(*args, **kwargs):
     "Extraction of arguments for _solve_varproblem"
 
     # Check for use of valid kwargs
-    valid_kwargs = ["bcs", "J", "Jp", "M",
```
pbrubeck, line 282: `M` had been used before as the objective, however we were completely ignoring it

firedrake/variational_solver.py
```diff
@@ -442,6 +454,7 @@ def __init__(self, a, L, u, bcs=None, aP=None,
         F = self.compute_bc_lifting(a, u, L=L)
 
         super(LinearVariationalProblem, self).__init__(F, u, bcs=bcs, J=a, Jp=aP,
+                                                       objective=objective,
```
pbrubeck: Does it make sense to pass an objective for `-snes_type ksponly`? Could there be a `norm_type` that uses this for the stopping criterion?

docs/source/solving-interface.rst
```diff
@@ -725,6 +725,19 @@ solving with.
    # Use the approximate inverse of Jp to precondition solves
    solve(a == L, ..., Jp=Jp)
 
+Providing an objective functional
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+An objective functional, usually the energy,
+can be provided to PETSc's linesearch and trust-region methods.
```
pbrubeck: @stefanozampini feel free to improve this manual section

firedrake/solving.py
```diff
@@ -265,21 +271,22 @@ def _extract_linear_solver_args(*args, **kwargs):
...
     nullspace_T = kwargs.get("transpose_nullspace", None)
     near_nullspace = kwargs.get("near_nullspace", None)
     options_prefix = kwargs.get("options_prefix", None)
     pre_apply_bcs = kwargs.get("pre_apply_bcs", True)
 
-    return P, bcs, solver_parameters, nullspace, nullspace_T, near_nullspace, options_prefix, pre_apply_bcs
+    return P, bcs, objective, solver_parameters, nullspace, nullspace_T, near_nullspace, options_prefix, pre_apply_bcs
 
 
 def _extract_args(*args, **kwargs):
```
pbrubeck, line 281: These functions are forcing separation between args and kwargs (e.g. solve will not let you pass solver_parameters as positional argument). Therefore we can add new kwargs in any order without breaking API.

pbrubeck: Moreover, `objective` is taking the place of `M` (the goal functional), which was previously ignored in `solve`

firedrake/variational_solver.py
stefanozampini: no

firedrake/solving_utils.py
stefanozampini: the MR is now in PETSc `main`

firedrake/variational_solver.py
pbrubeck: Should we then remove it from LinearVariationalProblem?

pbrubeck: It does not harm having it there, but it might mislead users to think that the objective could be used by the KSP

docs/source/solving-interface.rst
```diff
@@ -725,6 +725,29 @@ solving with.
...
+can be provided to PETSc's `SNES`_ linesearch and trust-region methods.
+For example, the code block below minimizes the Dirichlet energy functional
+using a trust-region method with the Steihaugh-Toint method for the
+trust-region subproblem and an initial trust region size of ``1e-4``.
+The solver monitor will report residual norm, norm of the update, and
+value of the objective functional.
+
+.. code-block:: python3
+
+   E = 0.5*inner(grad(u), grad(u))*dx - inner(f, u)*dx
```
pbrubeck: The example should be nonlinear

tests/firedrake/regression/test_snes_objective.py
```diff
@@ -0,0 +1,93 @@
...
+    f = (2*pi)**2 * sin(2*pi*x) + kappa**2 * sinh(u_exact)
+    E = (
+        0.5 * inner(grad(u), grad(u))*dx
+        + kappa**2 * cosh(u)*dx
+        - f*u*dx
+    )
+    F = (
+        inner(grad(u), grad(v))*dx
+        + kappa**2 * sinh(u)*v*dx
+        - f*v*dx
```
pbrubeck:
```suggestion
        + kappa**2 * inner(sinh(u), v)*dx
        - inner(f, v)*dx
```

tests/firedrake/regression/test_snes_objective.py
```diff
@@ -0,0 +1,93 @@
...
+    sol2 = Function(V)
+
+    x, = SpatialCoordinate(mesh)
+    kappa = Constant(1.0)
+    u_exact = sin(2*pi*x)
+    f = (2*pi)**2 * sin(2*pi*x) + kappa**2 * sinh(u_exact)
+    E = (
+        0.5 * inner(grad(u), grad(u))*dx
+        + kappa**2 * cosh(u)*dx
+        - f*u*dx
```
pbrubeck:
```suggestion
        - inner(f, u)*dx
```

firedrake/variational_solver.py
stefanozampini: I agree, please remove it

docs/source/solving-interface.rst
stefanozampini: done

firedrake/variational_solver.py
```diff
@@ -62,6 +68,7 @@ def __init__(self, F, u, bcs=None, J=None,
         :param Jp: a form used for preconditioning the linear system,
                  optional, if not supplied then the Jacobian itself
                  will be used.
+        :param objective: a form used for line-search, optional
```
stefanozampini:
```suggestion
        :param objective: a form used for line-search or trust-region methods, optional
```

firedrake/solving_utils.py
```diff
@@ -449,6 +462,27 @@ def split(self, fields):
...
+        :arg X: the current guess (a Vec)
+        """
+        dm = snes.getDM()
+        ctx = dmhooks.get_appctx(dm)
+        # X may not be the same vector as the vec behind self._x, so
+        # copy guess in from X.
+        with ctx._x.dat.vec_wo as v:
+            X.copy(v)
+
+        if not ctx.pre_apply_bcs:
```
```suggestion
        if ctx.pre_apply_bcs:
```

stefanozampini: why? the objective function should always be evaluated with the correct bcs in.

docs/source/solving-interface.rst
```diff
@@ -725,6 +725,29 @@ solving with.
...
+For example, the code block below minimizes the Dirichlet energy functional
+using a trust-region method with the Steihaugh-Toint method for the
+trust-region subproblem and an initial trust region size of ``1e-4``.
+The solver monitor will report residual norm, norm of the update, and
+value of the objective functional.
+
+.. code-block:: python3
+
+   E = 0.5 * inner(grad(u), grad(u))*dx + kappa**2 * cosh(u)*dx - inner(f, u)*dx
+   F = inner(grad(u), grad(v))*dx + kappa**2 * inner(sinh(u), v)*dx - inner(f, v)*dx
```
pbrubeck:
```suggestion
   F = derivative(E, u)
```

firedrake/solving_utils.py
```diff
@@ -449,6 +478,26 @@ def split(self, fields):
...
+        dm = snes.getDM()
+        ctx = dmhooks.get_appctx(dm)
+        # X may not be the same vector as the vec behind self._x, so
+        # copy guess in from X.
+        with ctx._x.dat.vec_wo as v:
+            X.copy(v)
+
+        # Apply DirichletBC on the solution
+        for bc in ctx._problem.dirichlet_bcs():
+            bc.apply(ctx._x)
```
```suggestion
        # Apply DirichletBC on the solution
        if ctx.pre_apply_bcs:
            for bc in ctx._problem.dirichlet_bcs():
                bc.apply(ctx._x)
```

firedrake/solving_utils.py
```diff
@@ -449,6 +478,26 @@ def split(self, fields):
...
+
+        :arg snes: a PETSc SNES object
+        :arg X: the current guess (a Vec)
+        """
+        dm = snes.getDM()
+        ctx = dmhooks.get_appctx(dm)
+        # X may not be the same vector as the vec behind self._x, so
+        # copy guess in from X.
+        with ctx._x.dat.vec_wo as v:
+            X.copy(v)
```
pbrubeck, line 477: Is this safe? `ctx._x` is the solution according to Firedrake (but SNES has a different one) can we treat `ctx._x` as a buffer?

stefanozampini: I don't know, I copied that from the residual evaluation routine. At the end of `solve`, you have `work.copy(u)`, with `u = problem.u_restrict.dat.vec ` and `work` the Vec SNES is using for the solution. From the point of view of the `solve`, this is ok. But if your `ctx._x` is supposed to be immutable for a given Newton step, then overwriting it can be a problem, yes.

firedrake/solving_utils.py
```diff
@@ -360,6 +386,9 @@ def set_nullspace(self, nullspace, ises=None, transpose=False, near=False):
         if ises is not None:
             nullspace._apply(ises, transpose=transpose, near=near)
 
+    def set_ksp_postsolve(self, snes):
+        snes.getKSP().setPostSolve(self.ksp_postsolve)
```
JHopeCollins: Will this overwrite an existing user post-solve callback? Or will they both be called in that case?

firedrake/solving_utils.py
stefanozampini: I don't think what you are proposing is correct, but I also don't think what I have in the current PR is 100% correct. In all my experience with nonlinear solvers, you always inject the correct BCs whenever you evaluate objective functions or residuals. Then, in case of residuals, you modify afterward the part of the vector that is associated with bc, because there you solve a different set of equations, i.e.

```
F(u) = 0 on \Omega
u - g = 0 on \partial\Omega
``` 

As a byproduct, if you do Newton, the boundary conditions will be exact after the first full step (that may not be the first Newton step if you use linesearch). See e.g. how I do it in MFEM https://github.com/mfem/mfem/blob/master/linalg/petsc.cpp#L4996

If ctx._x is supposed to be immutable during each phases of a given SNES step, then you need to rethink the firedrake logic a bit

firedrake/solving_utils.py
pbrubeck: This is a different one. The existing ones are SNES related (pre/post Jacobian/Function)

tests/firedrake/regression/test_snes_objective.py
```diff
@@ -0,0 +1,161 @@
...
+    problem = NonlinearVariationalProblem(F, u, bcs)
+    solver = NonlinearVariationalSolver(problem, solver_parameters=sp, pre_apply_bcs=pre_apply_bcs)
+    solver.solve()
+    e2 = assemble(E)
+    sol2.assign(u)
+
+    print(norm(u_exact - sol1), norm(u_exact - sol2))
+    print(e1, e2)
+    # print(math.isclose(norm(u_exact - sol1), norm(u_exact - sol2), rel_tol=1.e-2))
+    assert math.isclose(norm(u_exact - sol1), norm(u_exact - sol2), rel_tol=1.e-2)
```
JHopeCollins:
```suggestion
    assert math.isclose(errornorm(u_exact, sol1), errornorm(u_exact, sol2), rel_tol=1.e-2)
```

firedrake/solving_utils.py
JHopeCollins: Not the ones passed to the nlvs, the ones on the actual ksp. A user might have set one, or a particular ksp type might define one itself

stefanozampini: This is a temporary fix to firedrake's model of `pre_apply_bcs=True`, which I do not fully understand (or agree with :-) ). May be removed if we understand what's going wrong with the bc model and inexact solves

stefanozampini: > Not the ones passed to the nlvs the ones on the actual ksp
> A user might have set one, or a particular ksp type might define one itself

This is set as the default when firedrake creates the SNES. Users can do whatever they want afterward. Add their own, or change the KSP.  And no, KSPs don't have their own since each KSP implements its own Solve; these are user-specific callbacks

As I said, this is temporary and can be removed once we understand what is going wrong with `pre_apply_bcs=True`

stefanozampini: @pbrubeck Did you figure out what is wrong with FAS and boundary conditions?

JHopeCollins: If the user sets a Python type KSP then that can define a `postSolve` method.
Could we just have a small check, something like:

```py
existing_callback = snes.ksp.getPostSolve()
if existing_callback:
	def postsolve(*args):
		self.ksp_postsolve(*args)
		existing_callback(*args)
else:
	postsolve = self.ksp_postsolve
snes.ksp.setPostSolve(postsolve)
```

I just want to avoid accidentally silently ignoring something a user might set. I know it's very unlikely someone will hit this but silent-ignore bugs can be really annoying to spot and debug.

pbrubeck: This was just for debugging purposes. We'll remove the changes for DirichletBC together with this callback from the PR

firedrake/solving_utils.py
JHopeCollins: Can `form_objective` be called within a single Newton step? i.e. by the KSP? If so then I do not think that this is safe because for anything matrix-free we assume that `ctx._x` holds the current state.

I think you just need to save the state in `ctx._x` to a temporary buffer at entry to this function and restore it at exit.

tests/firedrake/regression/test_snes_objective.py
```diff
@@ -0,0 +1,169 @@
+from firedrake import *
+import pytest
+import math
+
+# GAMG has issues with FAS and pre_apply_bcs=True
+pc_types = ("gamg",
+            "jacobi",
+            "hypre",
+            "ilu")
```
JHopeCollins, line 10:
```suggestion
pc_types = ("gamg",
            "jacobi",
            pytest.param("hypre", marks=skiphypre),
            "ilu")
```

tests/firedrake/regression/test_snes_objective.py
```diff
@@ -0,0 +1,169 @@
...
+
+
+@pytest.mark.parametrize("refine", (0, 1))
+@pytest.mark.parametrize("pre_apply_bcs", (False, True, "restrict"))
+@pytest.mark.parametrize("pc_type", pc_types)
+def test_poisson_boltzmann_energy(refine, pre_apply_bcs, pc_type):
+    if pc_type == 'hypre' and not PETSc.Sys.hasExternalPackage("hypre"):
+        pytest.xfail("Need PETSc with HYPRE support for this test")
```
JHopeCollins: See suggestion above to use the existing skippackage marker.
```suggestion
```

tests/firedrake/regression/test_snes_objective.py
JHopeCollins, line 10: Do we need to be testing this many PCs? Do we expect significantly different behaviour between them that would indicate a bug Firedrake? Just trying to be conscious of the size of our (already bloated) test suite!

firedrake/solving_utils.py
JHopeCollins: Do you mean that it will be removed before merging? In which case it's fine, it's just if it is intended to be part of the finished PR.

tests/firedrake/regression/test_snes_objective.py
pbrubeck: Jacobi and gamg should stay, as they were originally broken (they were too inexact and caused NewtonTR failures). Hypre and ILU never failed so they serve as our reference, so we can at most remove one.

This is a 1D test, and changing the PC does not generate any new code, we should not be worried about timing here.

firedrake/solving_utils.py
stefanozampini: @pbrubeck I don't think tests will pass for the pre_apply_bcs=True case if we remove the postsolve callback, have you tried?

firedrake/solving_utils.py
stefanozampini: In principle it should not happen, but I agree with you better be safe. Make a suggestion in the diff and I will merge it in

firedrake/solving_utils.py
pbrubeck: They won't pass. But the BCs issue is separate to this PR, so we will need to temporarily disable it here, and reenable it when GAMG + pre_apply_bcs=True gets properly fixed. We have three options: 1. fix `pre_apply_bcs=True` by adding the callback, 2. short-circuit `pre_apply_bcs=restrict` and let `pre_apply_bcs=restrict=False` be the default, or 3. to fix  GAMG in PETSc.

stefanozampini: I'm fine xfailing the broken tests. This is not a GAMG issue, it is a bug in the pre_apply_bcs=True model. GAMG just exposed it.

pbrubeck: I agree, the `pre_apply_bcs=True` is flawed in the SNES sense (does not let you linearize around a state that does not satisfy the BCs), but also in the KSP sense (there is no way to make sure that the PC would not introduce coupling between the BC and the interior, i.e. the PC needs to invert the identity block exactly). The short-circuit `pre_apply_bcs=restrict`, fixes the KSP issue, but not the SNES one.

pbrubeck: Also, would the ksp_postsolve callback be called with -snes_type ksponly? I tried to add a linear test for GAMG, and noticed that the callback was not being called.

tests/firedrake/regression/test_snes_objective.py
JHopeCollins: Seeing as it's 1D and you want it as a reference, how about replacing `hypre` and `ilu` with `lu`?

firedrake/solving_utils.py
stefanozampini: Yes, it should.

tests/firedrake/regression/test_snes_objective.py
stefanozampini: These tests exercise approximate linear solves, so lu does not belong here

firedrake/solving_utils.py
stefanozampini: > I agree, the pre_apply_bcs=True is flawed in the SNES sense (does not let you linearize around a state that does not satisfy the BCs), but also in the KSP sense (there is no way to make sure that the PC would not introduce coupling between the BC and the interior, i.e. the PC needs to invert the identity block exactly). The short-circuit pre_apply_bcs=restrict, fixes the KSP issue, but not the SNES one.

What do you mean. With restrict you do not have BC dofs at all in the system, so it should not cause problems either to KSP or SNES, since you will just solve the proper linear system only approximatively

pbrubeck: For option 2, we would change the default to `pre_apply_bcs=restrict=False` and deprecate the argument `pre_apply_bcs` from the next release onwards, but still accept two separate arguments for `restrict` and `pre_apply_bcs` and make sure that they are equal. We then refactor the internal code to have a unified `restrict/pre_apply_bcs` flag. Finally, remove the `pre_apply_bcs` argument on the following release.

pbrubeck: > What do you mean. With restrict you do not have BC dofs at all in the system, so it should not cause problems either to KSP or SNES, since you will just solve the proper linear system only approximatively

The SNES issue is unavoidable and inherent to the `pre_apply_bcs=True` model. It might result in NaN at the first residual on some nonlinear problems when the initial guess is not consistent with the BCs. Imposing the BCs for such guesses introduces very sharp gradients. See [this PR](https://github.com/firedrakeproject/firedrake/pull/3995) description, and the [test](https://github.com/firedrakeproject/firedrake/blob/05c948402c64339d0da72592e0ccc81c8a01e35f/tests/firedrake/regression/test_solving_interface.py#L322) that we added for `pre_apply_bcs=False`.

pbrubeck: I opened https://github.com/firedrakeproject/firedrake/issues/5177 to continue with this discussion

danshapero:
I would find this feature really useful. It looks like something with boundary conditions needs to be settled and some changes to the unit tests? Is there anything else need in order to get this merged? I can try to pitch in if it's useful.

pbrubeck:
> I would find this feature really useful. It looks like something with boundary conditions needs to be settled and some changes to the unit tests? Is there anything else need in order to get this merged? I can try to pitch in if it's useful.

Yes, we will need to undo the bc-related changes and xfail `pre_apply_bcs=True`.  Feel free to throw Claude at this, but first merge main so it picks up AGENTS.md (tell it to read it first)

danshapero:
> > I would find this feature really useful. It looks like something with boundary conditions needs to be settled and some changes to the unit tests? Is there anything else need in order to get this merged? I can try to pitch in if it's useful.
> 
> Yes, we will need to undo the bc-related changes and xfail `pre_apply_bcs=True`. Feel free to throw Claude at this, but first merge main so it picks up AGENTS.md (tell it to read it first)

Thanks, I'll give it a go. Earlier you'd suggested merging this onto release instead of main, is that still true?

firedrake/preconditioners/patch.py
```diff
@@ -587,47 +587,6 @@ def make_patch_callables(form: ufl.Form, state: Function | None) -> tuple[
     return cell_callable, interior_facet_callable, exterior_facet_callable
 
 
-def bcdofs(bc, ghost=True):
```
pbrubeck, line 590: Move this back to its original location

firedrake/solving_utils.py
```diff
@@ -277,6 +285,25 @@ def __init__(self, problem,
         self._coefficient_mapping = None
         self._transfer_manager = transfer_manager
 
+    @cached_property
+    def bc_iset(self):
```
pbrubeck: Remove

tests/firedrake/regression/test_snes_objective.py
```diff
@@ -0,0 +1,169 @@
...
+
+# GAMG has issues with FAS and pre_apply_bcs=True
+pc_types = ("gamg",
+            "jacobi",
+            "hypre",
+            "ilu")
+
+
+@pytest.mark.parametrize("refine", (0, 1))
+@pytest.mark.parametrize("pre_apply_bcs", (False, True, "restrict"))
```
pbrubeck, line 14: XFail the cases that are failing after removing the bc "fixes"

firedrake/solving_utils.py
```diff
@@ -547,6 +604,13 @@ def compute_operators(ksp, J, P):
             assert P.handle == ctx._pjac.petscmat.handle
             ctx._assemble_pjac(ctx._pjac)
 
+    @staticmethod
+    def ksp_postsolve(ksp, rhs, sol):
```
pbrubeck: Remove

firedrake/solving_utils.py
```diff
@@ -360,6 +393,9 @@ def set_nullspace(self, nullspace, ises=None, transpose=False, near=False):
         if ises is not None:
             nullspace._apply(ises, transpose=transpose, near=near)
 
+    def set_ksp_postsolve(self, snes):
```
pbrubeck: Remove

pbrubeck (changes requested):
Drop BC related changes and xfail the cases that fail with pre_apply_bcs=True

pbrubeck:
> Thanks, I'll give it a go. Earlier you'd suggested merging this onto release instead of main, is that still true?

Sure. This is purely additive

danshapero:
I added some changes on the branch `drs/snes-objective`. I threw claude at it so maybe don't trust it over much.

I don't think we can merge this onto release. It depends on the change that @stefanozampini added to PETSc which is only available on main right now. It's not in the latest release (3.25.3). Unless I'm mistaken it won't be available until 3.26.0.

pbrubeck:
> I added some changes on the branch `drs/snes-objective`. I threw claude at it so maybe don't trust it over much.

Looks fine. Claude understood the assignment. 

> I don't think we can merge this onto release. It depends on the change that @stefanozampini added to PETSc which is only available on main right now. It's not in the latest release (3.25.3). Unless I'm mistaken it won't be available until 3.26.0.

You are right. We need to merge this to main then

firedrake/solving_utils.py
pbrubeck: I think this is fine, we do something very similar to compute the residual

pbrubeck (approved):

firedrake/solving.py
```diff
@@ -167,6 +167,8 @@ def _solve_varproblem(*args, **kwargs):
         raise TypeError(f"Equation LHS must be a ufl.BaseForm, not a {type(eq.lhs).__name__}")
 
     if len(eq.lhs.arguments()) == 2:
+        if objective is not None:
+            raise ValueError("The objective functional only makes sense for nonlinear problems.")
```
stefanozampini, line 171: This is not true. `f(x) = x^t A x - X^t b - c` is a valid objective function. Shouldn't in this case just silently ignore the `objective` callback?

stefanozampini (commented):
This is ok for me (according to github, I cannot approve my own PR). At some point there were useful methods to the `DirichletBC` class (like extracting the index set, etc) that I think can be cherry-picked in a different PR. @pbrubeck What do you think?

firedrake/solving.py
pbrubeck: We can ignore it then.

pbrubeck:
> This is ok for me (according to github, I cannot approve my own PR). At some point there were useful methods to the `DirichletBC` class (like extracting the index set, etc) that I think can be cherry-picked in a different PR. @pbrubeck What do you think?

In #5185 `pre_apply_bcs=True` is going to die.

firedrake/solving.py
pbrubeck: This is error is raised in solve() if the user specifies a linear problem. However LinearVariationalSolver does not have am `objective` kwarg. I think we should keep the loud `ValueError`

PR #5161, opened 2026-06-10
Gather adjoint solver cached data together

JHopeCollins:
Tidies up the API for the refactored cached solver block for the tape.

### DIFF
--- a/firedrake/adjoint_utils/blocks/solving.py
+++ b/firedrake/adjoint_utils/blocks/solving.py
@@ -190,10 +190,12 @@ class CachedSolverBlock(Block):
 
         adj_sol, adj_sol_bc = self.solve_adj_equation(dJdu, compute_boundary)
 
-        # store adj_sol for Hessian computation later.
-        # self.adj_sol is shared between all blocks that this NLVS
-        # generates so we can't store it there. Instead store it
-        # in self.adj_sol_buf which is owned by this block only.
+        # store adj_sol for Hessian computation later, or for inspecting
+        # adjoint sensitivities etc.
+        # self.adj_sol is owned by this block, whereas self.hessian_cache.adj_sol
+        # is shared between all blocks that the NLVS generates because it is the
+        # one in the cached forms, so we will update it as necessary if/when each
+        # block calculates the Hessian action.
         self.adj_sol.assign(adj_sol)
 
         prepared = {
--- a/firedrake/adjoint_utils/blocks/solving.py
+++ b/firedrake/adjoint_utils/blocks/solving.py
@@ -65,7 +65,14 @@ class CachedSolverBlock(Block):
         self.adj_dFdm_forms = adj_dFdm_forms
         self.adj_residual = adj_residual
 
-        self.adj_sol = adj_sol
+        # this one belongs to this block specifically and
+        # stashes the adjoint solution for the hessian calculation
+        self.adj_sol = adj_sol.copy(deepcopy=True, annotate=False)
+        # this one is in the cached forms which are shared by all
+        # all solve blocks for this solver, it needs to be updated
+        # with the adj_sol value for this block before assembling
+        # the cached forms.
+        self._adj_sol = adj_sol
         self.adj2_sol = adj2_sol
         self.tlm_output = tlm_output
         self.d2Fdu2_form = d2Fdu2_form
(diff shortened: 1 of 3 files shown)


JHopeCollins:
@angus-g the `test_nlvs` tests should all pass, but obviously the rest of the linear solver failures are still there in this branch. It's tidied up how much we're passing to the new `Block` though.

firedrake/adjoint_utils/blocks/solving.py
```diff
@@ -161,42 +145,40 @@ def evaluate_tlm_component(self, inputs, tlm_inputs, block_variable, idx, prepar
...
+        for bc in self.forward_cache.bcs:
             bc.homogenize()
 
-        solver = self.cached_solvers[ADJOINT]
-        adj_sol = solver._problem.u
+        adj_rhs = self.adjoint_cache.rhs
+        adj_sol = self.adjoint_cache.adj_sol
 
-        self.adj_rhs.assign(rhs)
+        adj_rhs.assign(rhs)
```
angus-g, line 171: Doesn't seem to be any point caching `rhs` if we just ignore it here?

firedrake/adjoint_utils/blocks/solving.py
```diff
@@ -215,13 +197,13 @@ def prepare_evaluate_adj(self, inputs, adj_inputs, relevant_dependencies):
         self.adj_sol.assign(adj_sol)
```
angus-g, line 199: I guess my merge/nuke of `adj_sol_buf` makes this comment incorrect (sorry!). Do you want to address it here while you're reworking how `adj_sol` gets passed around?

firedrake/adjoint_utils/variational_solver.py
```diff
@@ -73,12 +124,13 @@ def wrapper(self, problem, *args, **kwargs):
 
         return wrapper
 
-    def _ad_cache_forward_solver(self):
+    @cached_property
```
angus-g, line 127: Do we need to be careful about this? Like is it possible to change BCs between `.solve()` calls or something weird like that?

angus-g (commented):
So much cleaner! Just a couple of minor things I spotted.

firedrake/adjoint_utils/variational_solver.py
JHopeCollins: You can change the values associated to the BCs etc, but you cannot switch out the objects themselves. For BCs I don't think it's even valid to change which DoFs they are acting on because we may have made a `RestrictedFunctionSpace` and we don't check whether we need to change the restricted set.

firedrake/adjoint_utils/blocks/solving.py
JHopeCollins: What do you mean? `adjoint_cache.rhs` is the thing we passed to `adjoint_cache.solver` so we assign to it here with whatever is in the `rhs` passed to this function (which may come from either the adjoint or hessian calculation).

angus-g: Oops I completely misread the diff. Disregard! 🙈

firedrake/adjoint_utils/variational_solver.py
angus-g: Okay, great. Same applies to all the other caching but I don't see anything that looks like it could cause accidental gremlin trouble...

firedrake/adjoint_utils/blocks/solving.py
JHopeCollins: Ah, yes you are right. The solver block must have its own `adj_sol` value so that we can inspect it later on, for example looking at the adjoint sensitivities. So now the one-per-solver `adj_sol` is stored in the cache tuple, and the block's `self.adj_sol` is the one-per-block value.

I will update the comment.

JHopeCollins: Done. Is that correct/clearer?

firedrake/adjoint_utils/variational_solver.py
JHopeCollins: Yeah I don't think so either. We will load up the values from the checkpoints anyway so we never kept a handle to the user-facing things anyway to check if they had changed.

angus-g (approved):

PR #5164, opened 2026-06-11
Minor updates to OffloadPC

dsroberts:
# Description

Two minor updates to `OffloadPC` based on continued local testing. The first change is preventing certain types of of PETSc matrix from being offloaded. When implicit matrices are converted to `aijcusparse` matrices, PETSc constructs a dense matrix row-by-row then converts that to a sparse matrix. Since PETSc can transparently manage host-device transfers when using non-`aijcusparse` matrix types on `cuda` vectors, it is far more efficient for PETSc to offload as needed than to do a full implicit to dense to sparse conversion. I'm not sure this is a complete list of implicit matrix types, but these are the types I run into with our testing.

The second change is in the construction of the device matrices. If the `out` kwarg is not specified or set to `None` in `Mat.convert()`, the conversion happens in place meaning e.g. both `P` and `P_device` are now the same `aijcusparse` matrix. This was probably not picked up earlier as an `aijcusparse` matrix is an `aij` matrix with a managed device buffer under the hood, so silently converting an `aij` matrix probably does not affect any other part of a job, but it might cause issues with other matrix types.

There is a check to make sure the silent conversion hasn't happened, but I don't know how to test for the implicit -> dense -> sparse conversion. The right thing to do I think would be to check that `A is A_dev` if `A.getType()` is one of the 'do not offload' types, but I don't know how to retrieve that information out of nested PCs.


firedrake/preconditioners/offload.py
```diff
@@ -13,6 +13,8 @@
     }
 }
 
+_no_offload_mat_types = ("python", "schurcompliment")
```
JHopeCollins:
```suggestion
_no_offload_mat_types = ("python", "schurcomplement")
```

dsroberts: Sigh... Its a small change, it'll be fine to quickly transcribe it...

firedrake/preconditioners/offload.py
```diff
@@ -53,8 +55,15 @@ def initialize(self, pc):
         pc.setOptionsPrefix(options_prefix)
         if self.device_mat is not None:
             with PETSc.Log.Event("Event: initialize offload"):
-                P_dev = P.convert(mat_type=self.device_mat)
-                A_dev = P_dev if A.handle == P.handle else A.convert(mat_type=self.device_mat)
+                P_dev = PETSc.Mat()
+                P_dev = P.convert(mat_type=self.device_mat, out=P_dev)
+                if A.handle == P.handle:
+                    A_dev = P_dev
+                elif A.getType() in _no_offload_mat_types:
```
connorjward:
```suggestion
                elif A.type in _no_offload_mat_types:
```

firedrake/preconditioners/offload.py
```diff
@@ -74,7 +83,7 @@ def update(self, pc):
         A, P = pc.getOperators()
         A_dev, P_dev = self.pc.getOperators()
         P.copy(P_dev)
-        if A_dev.handle != P_dev.handle:
+        if A_dev.handle != P_dev.handle and A.getType() not in _no_offload_mat_types:
```
connorjward:
```suggestion
        if A_dev.handle != P_dev.handle and A.type not in _no_offload_mat_types:
```

tests/firedrake/offload/test_poisson_offloading_pc.py
```diff
@@ -55,6 +55,9 @@ def run_test_poisson_offload(ksp_type, pc_type, homogeneous_bcs):
     problem = LinearVariationalProblem(L, R, u_f, bcs=bcs)
     solver = LinearVariationalSolver(problem, solver_parameters=parameters)
     solver.solve()
+
+    assert solver.snes.ksp.pc.getOperators()[1].getType() == "seqaij"
```
connorjward: This doesn't obviously test the new functionality.

firedrake/preconditioners/offload.py
```diff
@@ -13,6 +13,8 @@
     }
 }
 
+_no_offload_mat_types = ("python", "schurcomplement")
```
connorjward, line 22: Can this get a comment?

connorjward (changes requested):

tests/firedrake/offload/test_poisson_offloading_pc.py
dsroberts: It doesn't, but as I stated in the PR, I don't know how to do that. I've added a re-work of the mini-stokes test that has the right solver configuration to trigger this, but I do not know how to get the operators out of the nested `OffloadPC`. Note that it has to be nested like this as `AssembledPC` converts the `python` type matrix from the fieldsplit PC to a `seqaij` leaving `A` untouched. Without it, `OffloadPC` takes the `A.handle == P.handle` path and both are converted.

This tests that `P` going into `OffloadPC` is not overwritten by the device matrix. If you revert `offload.py` and run this test, this assert will trigger.

dsroberts:
I've added another fix into this PR that copies `P` and `A` to `P_dev` and `A_dev` respectively by values only in `OffloadPC.update()`. We found that in parallel some configurations were diverging after a few timesteps. I think this was due to PETSc seeing the copied `P_dev` as a new matrix and performing its own assembly, and when it did that it lost the connectivity information into the halos. I have adapted the `04-burgers.ipynb` demo into a test that runs until it reaches a steady state. When this test is run on 2 processes, the `("cg", "gamg")` variant fails in 3 timesteps if a full copy of `P` is performed. With the value-only copy the solution reaches a steady state and the test passes.

firedrake/preconditioners/offload.py
```diff
@@ -73,9 +85,11 @@ def initialize(self, pc):
     def update(self, pc):
         A, P = pc.getOperators()
         A_dev, P_dev = self.pc.getOperators()
-        P.copy(P_dev)
-        if A_dev.handle != P_dev.handle:
-            A.copy(A_dev)
+        # Perform a value-only copy
+        P.copy(P_dev, True)
```
connorjward: What does the `True` do? The API doesn't make it clear: https://petsc.org/main/petsc4py/reference/petsc4py.PETSc.Mat.html#petsc4py.PETSc.Mat.copy

connorjward: I think making it a kwarg would be more readable

tests/firedrake/offload/test_advection_offload.py
```diff
@@ -0,0 +1,78 @@
...
+    t_end = steps * dt
+
+    while t <= t_end:
+        solver.solve()
+        maxchange = sqrt(assemble((u_n - u_n1)**2 * dx))
+        u_n.assign(u_n1)
+        t += dt
+        if maxchange < 1e-5 or np.isnan(maxchange):
+            break
+    return maxchange
```
connorjward:
```suggestion
    assert maxchange < 1e-5
```

tests/firedrake/offload/test_advection_offload.py
```diff
@@ -0,0 +1,78 @@
...
+
+@pytest.mark.skipnogpu
+def test_advection_offload(ksp_pc):
+    assert run_test_advection_offload(*ksp_pc) < 1e-5
+
+
+@pytest.mark.parallel(nprocs=2)
+@pytest.mark.skipnogpu
+def test_advection_offload_parallel(ksp_pc):
+    assert run_test_advection_offload(*ksp_pc) < 1e-5
```
connorjward:
```suggestion
```

tests/firedrake/offload/test_advection_offload.py
```diff
@@ -0,0 +1,78 @@
...
+@pytest.fixture(scope="module",
+                params=[("cg", "sor"), ("cg", "gamg"), ("preonly", "lu")],
+                ids=lambda x: '_'.join(x)
+                )
+def ksp_pc(request):
+    return request.param
+
+
+def run_test_advection_offload(ksp_type, pc_type):
+
```
connorjward:
```suggestion
@pytest.mark.skipnogpu
@pytest.mark.parallel([1, 2])
def test_advection_offload(ksp_type, pc_type):
```

tests/firedrake/offload/test_no_offload_A.py
```diff
@@ -0,0 +1,106 @@
...
+
+    # Parabolic inflow y(1-y) at x = 0 in positive x direction
+    x = SpatialCoordinate(W.mesh())
+    inflow = as_vector((x[1]*(1 - x[1]), 0.0))
+    bc1 = DirichletBC(W[0], inflow, 1)
+
+    # Zero pressure at outlow at x = 1
+    bc2 = DirichletBC(W[1], 0.0, 2)
+
+    bcs = bc0 + [bc1, bc2]
```
connorjward: Not critical, but this seems a bit weird to have `bc0` be a list and `bc1` and `bc2` not be.

tests/firedrake/offload/test_no_offload_A.py
```diff
@@ -0,0 +1,106 @@
...
+
+    problem = LinearVariationalProblem(a, L, w, bcs=bcs)
+    solver = LinearVariationalSolver(problem, solver_parameters=iterative_stokes_solver_parameters)
+    solver.solve()
+
+    # ksp0 = solver.snes.ksp.pc.getFieldSplitSchurGetSubKSP()[0]
+    # sub_pc = ksp0.pc.???
+    # offload_python_context = sub_pc.getPythonContext()
+    # assert offload_python_context.pc.P.type == "seqaijcusparse"
+    # assert offload_python_context.pc.A.type == "python"
```
connorjward: remove?

tests/firedrake/offload/test_no_offload_A.py
```diff
@@ -0,0 +1,106 @@
...
+    # We've set up Poiseuille flow, so we expect a parabolic velocity
+    # field and a linearly decreasing pressure.
+    x, y = SpatialCoordinate(mesh)
+    uexact = as_vector([y*(1 - y), 0])
+    pexact = 2*(length - x)
+
+    return errornorm(uexact, u, degree_rise=0), errornorm(pexact, p, degree_rise=0)
+
+
+def test_no_offload_A():
```
connorjward:
```suggestion
def test_matfree_A_does_not_offload():
```
?

connorjward (changes requested):

tests/firedrake/offload/test_no_offload_A.py
dsroberts: Copied straight from [here](https://github.com/firedrakeproject/firedrake/blob/fda580f346801e2573bbb6a5f55cc4559b9e7eeb/tests/firedrake/regression/test_stokes_mini.py#L39)

tests/firedrake/offload/test_no_offload_A.py
dsroberts: I'm afraid I've not explained the issue well. This test returns the same values regardless of whether `A` is offloaded or not. What we're testing for here is whether the `A` operator has been offloaded at all. It makes no difference at this scale, but when you begin to scale up (~100k DOF) the matfree -> dense -> sparse conversion can take minutes, as opposed to microseconds for an `aij` -> `aijcusparse` conversion, so its critically important that we make sure this isn't happening. 

This is where I need help. `ksp0 = solver.snes.ksp.pc.getFieldSplitSchurGetSubKSP()[0].pc` gives you the `firedrake.AssembledPC`, what I don't know is how to get the to the underlying `firedrake.OffloadPC`. Once we have that, we can get the python context and then extract the `A` and `P` operators from that. If you know what
```
sub_pc = ksp0.pc.???
```
should be in order to do that, that would be very helpful.

connorjward: Ah I see now. I also don't know how to traverse the solver stack programmatically like that. @JHopeCollins do you know?

I am on leave currently but could dig into this next week if needed.

JHopeCollins: `firedrake.AssembledPC` itself just creates a `PETSc.PC`, and you have specified that `PC` to be python type with the `OffloadPC` as the python context.
So looking at the solver parameters we need to walk down the solver stack in this order (where you had the first few steps already):

solver -> snes -> ksp -> pcfieldsplit -> A00 ksp -> pc -> python assembled pc context -> pc -> python offload pc context -> pc -> operators

 you can do something like:

```py
solver.snes.ksp.pc.getFieldSplitSchurGetSubKSP()[0].pc.getPythonContext().pc.getPythonContext().pc.getOperators()
```
Or, more readably, using your `ksp0`:
```py
assembled_ctx = ksp0.pc.getPythonContext()
offload_ctx = assembled_ctx.pc.getPythonContext()
A, P = offload_ctx.pc.getOperators()
```

tests/firedrake/offload/test_no_offload_A.py
```diff
@@ -0,0 +1,106 @@
...
+                "pc_type": "python",
+                "pc_python_type": "firedrake.OffloadPC",
+                "offload": {
+                    "pc_type": "ksp",
+                    "ksp": {
+                        "ksp_monitor_true_residual": None,
+                        "ksp_converged_reason": None,
+                        "ksp_type": "cg",
+                        "ksp_rtol": 1e-5,
+                        "ksp_max_it": 1000,
```
JHopeCollins: Does this really take this many iterations? Seems like too many for the test suite

tests/firedrake/offload/test_no_offload_A.py
```diff
@@ -0,0 +1,106 @@
...
+                    }
+                }
+            }
+        },
+        "fieldsplit_1": {
+            "pc_type": "none",
```
JHopeCollins, line 68: Is this the right thing to do here? Isn't this just solving a zero matrix because PETSc will default to building the Schur complement from `a11`, which for this form is just 0?

If it isn't doing that, then how many iterations does this take? Doesn't the `ksp_type` here default to `gmres`?

tests/firedrake/offload/test_no_offload_A.py
dsroberts: Ah, I was missing `getPythonContext()`! I was going through every `get...` function in `petsc4py.PC` trying to figure out which one would get the 'subPC' or something along those lines. Thank you @JHopeCollins.

tests/firedrake/offload/test_no_offload_A.py
dsroberts: No, sorry, left over from G-ADOPT. Will reduce to 50

tests/firedrake/offload/test_no_offload_A.py
JHopeCollins: No problem!

For future reference, a python context is no longer a petsc4py thing, its a user-defined class with a bunch of callbacks (here Firedrake counts as a user), e.g. `AssembledPC`, `OffloadPC` etc, which are just (non-petsc4py) python classes.

So once you hit `AssembledPC` you're back to looking at Firedrake code not petsc4py, until of course you hit another petsc4py object.

tests/firedrake/offload/test_no_offload_A.py
dsroberts: I took this from from the `stokes_mini` regression test, I only cared what was going on with `fieldsplit_0` so left the `fieldsplit_1` settings unchanged. The `ksp_type` does default to `gmres` and I see
```
     Linear firedrake_3_fieldsplit_1_ solve converged due to CONVERGED_RTOL iterations 130
```
when run with `ksp_converged_reason: None` added to the `fieldsplit_1` solver options. Setting it back to `'fieldsplit_schur_fact_type': 'diag',` as per `stokes_mini`, the iteration count drops to 125 and it fails the assertion at the end. Come to think of it, the assertion at the end doesn't matter, all we need to know is that `A` was not offloaded, so I'll just end the test after those asserts.

JHopeCollins: If you don't care about the solution then you can add `ksp_type: preonly` to this split

firedrake/preconditioners/offload.py
```diff
@@ -76,9 +88,11 @@ def initialize(self, pc):
     def update(self, pc):
         A, P = pc.getOperators()
         A_dev, P_dev = self.pc.getOperators()
-        P.copy(P_dev)
-        if A_dev.handle != P_dev.handle:
-            A.copy(A_dev)
+        # Perform a value-only copy
+        P.copy(P_dev, structure=P.Structure.SAME_NONZERO_PATTERN)
```
pbrubeck:
```suggestion
        P.copy(P_dev, structure=PETSc.Mat.Structure.SAME_NONZERO_PATTERN)
```

dsroberts:
Just wanted to check that everything is OK with this? I think I've addressed all the review points.

connorjward (approved):
This all seems pretty reasonable to me. @pbrubeck @JHopeCollins are you both happy?

PR #5165, opened 2026-06-11
BCs for cross-mesh interpolation

leo-collins:
Support BCs for cross-mesh interpolation matrices


firedrake/interpolation.py
```diff
@@ -563,10 +563,28 @@ def _interpolate_from_quadrature(self) -> Interpolate:
...
+        return f
+
+    def apply_bcs(self, mat: PETSc.Mat, bcs: Iterable[DirichletBC]) -> PETSc.Mat:
+        """Zero the rows and columns of `mat` associated with boundary condition nodes.
+        """
+        row_arg, col_arg = self.ufl_interpolate.arguments()
+        row_mask = self._bc_mask(row_arg.function_space(), bcs)
+        col_mask = self._bc_mask(col_arg.function_space(), bcs)
+        with row_mask.dat.vec_ro as r, col_mask.dat.vec_ro as c:
+            mat.diagonalScale(r, c)
```
pbrubeck, line 583: Use `mat.zeroRowsColumnsLocal()` instead.

firedrake/interpolation.py
```diff
@@ -563,10 +563,28 @@ def _interpolate_from_quadrature(self) -> Interpolate:
...
 
+    def _bc_mask(self, space: WithGeometry, bcs: Iterable[DirichletBC]) -> Function:
+        """Return a 0/1 mask over `space` which is zero at boundary condition nodes
+        """
+        space = space.dual() if is_dual(space) else space
+        f = Function(space).assign(1.0)
+        for bc in bcs:
+            if bc.function_space() == space:
+                bc.zero(f)
+        return f
```
pbrubeck, line 574: This is reinventing the wheel. We have some functions that do similar things scattered all over the place. My recommendation is to move `bcdofs` from preconditioners/patch.py to bcs.py and use it instead.

firedrake/interpolation.py
```diff
@@ -593,12 +611,13 @@ def callable() -> PETSc.Mat:
                     source_space = self.operand.function_space()
                     if self.ufl_interpolate.is_adjoint:
                         I = Matrix(interpolate(TestFunction(source_space), self.target_space), res)
-                        return assemble(action(I, self._interpolate_from_quadrature)).petscmat
+                        res = assemble(action(I, self._interpolate_from_quadrature)).petscmat
```
pbrubeck, line 614: Wouldn't the right thing be to reuse the support that assemble has for bcs?
```suggestion
                        res = assemble(action(I, self._interpolate_from_quadrature), bcs=bcs).petscmat
```

leo-collins: `VomOntoVomInterpolator` doesn't support BCs. I suppose we could do what this PR is doing for VomOntoVom and then it should work.

pbrubeck: But the BCs that the user provided are not defined on a VOM FunctionSpace. Why do we need to support BCs for VomOntoVom?

leo-collins: If I do `res = assemble(interp_expr, mat_type=mat_type, bcs=bcs).petscmat` then the BCs get passed through to VomOntoVomInterpolator

pbrubeck: This seems like a bug, we should only set BCs on the final matrix-matrix product

PR #5166, opened 2026-06-11
Assemble interpolates with Indexed arguments

leo-collins:
If we don't pass the ISes to `createNest`, PETSc complains with errors like `[0] No nonzero submatrix in column 1`.


pbrubeck (dismissed):
LGTM

pbrubeck:
As this is a bug fix, we should target `release`

connorjward (approved):

PR #5157, opened 2026-06-09
Test that control checkpoints survive recompute

stephankramer:
Same as #5093

Changes to firedrake adjoint tests corresponding to this fix in pyadjoint: https://github.com/dolfin-adjoint/pyadjoint/pull/257 (see #5093 for detailed description of changes)

Here based on release with a temporary change to use https://github.com/dolfin-adjoint/pyadjoint/pull/257 in CI

Issue #257, opened 2014-04-11
"Could not find index set" when using fieldsplit


ctjacobs:
I'm using the same fieldsplit solver options as the test_poisson_mixed_strong_bcs.py test:

```
solve(F == 0, self.solution, bcs=bcs, solver_parameters={'ksp_monitor': True, 
                                                         'ksp_view': False, 
                                                         'pc_view': False, 
                                                         'pc_type': 'fieldsplit',
                                                         'pc_fieldsplit_type': 'schur',
                                                         'ksp_type': 'gmres',
                                                         'pc_fieldsplit_schur_fact_type': 'FULL',
                                                         'fieldsplit_0_ksp_type': 'cg',
                                                         'fieldsplit_1_ksp_type': 'cg',
                                                         'ksp_rtol': 1.0e-7,
                                                         'snes_type': 'ksponly'})
```

but I get the following PETSc error when running my MMS test in the shallow water code:

```
christian@elevate /data/firedrake-fluids/tests/swe_mms_p2p1 $ make run
python ../../models/shallow_water.py MMS_A.swml
Initialising simulation...
Setting up a new Continuous Lagrange function space of degree 2 called VelocityFunctionSpace
Setting up a new Continuous Lagrange function space of degree 1 called FreeSurfaceFunctionSpace
pyop2:WARNING *** Projecting output function to CG1
pyop2:INFO Solving linear variational problem...
pyop2:INFO Solving linear variational problem...done

t = 0.01
Adding mass term...
Adding advection term...
Adding stress term...
Adding momentum source...
Adding continuity source...
Applying Velocity BC #0
Applying FreeSurfacePerturbation BC #0
pyop2:INFO Solving nonlinear variational problem...
Traceback (most recent call last):
  File "../../models/shallow_water.py", line 512, in <module>
    sw.run()
  File "../../models/shallow_water.py", line 465, in run
    'snes_type': 'ksponly'})
  File "/home/christian/firedrake/firedrake/solving.py", line 855, in solve
    _solve_varproblem(*args, **kwargs)
  File "/home/christian/firedrake/firedrake/solving.py", line 898, in _solve_varproblem
    solver.solve()
  File "/home/christian/firedrake/firedrake/solving.py", line 271, in solve
    self.snes.solve(None, v)
  File "SNES.pyx", line 413, in petsc4py.PETSc.SNES.solve (src/petsc4py.PETSc.c:143714)
petsc4py.PETSc.Error: error code 75
[...]

PR #5168, opened 2026-06-12
Regression tests for SingleMemoryStorageSchedule checkpoint clearing

sghelichkhani:
Companion to dolfin-adjoint/pyadjoint#248, which fixes the `SingleMemoryStorageSchedule` checkpoint clearing reported in dolfin-adjoint/pyadjoint#211 and dolfin-adjoint/pyadjoint#260. Two regression tests, one per failure mode.

The first covers #211: a block variable reused only every third timestep is absent from the immediately preceding step's `adjoint_dependencies`, so the old condition cleared its checkpoint during forward replay and the reverse pass produced a wrong gradient (73765 instead of 3205 in the test). The test tapes ten such staggered timesteps and compares functional and gradient under `SingleMemoryStorageSchedule` against an unscheduled tape.

The second covers #260: a variable that is never redefined keeps the live block variable of its Function, and the clearing discarded its checkpoint at its last forward use even though the adjoint of that step still needs it as a linearisation point. The `saved_output` fallback to the live Function then supplies the taping-time value, so the gradient is wrong whenever the functional is re-evaluated at a new control first (140.5536 instead of 187.4048 in the test). The test runs two evaluate/derivative cycles at new controls, the second of which exercises the clearing that resumes once the adjoint dependencies have been revised by the first reverse pass.

In both tests the ordering is load-bearing and commented: the clearing only runs during the forward replay triggered by re-evaluating the reduced functional, so `Jhat` is called before `derivative()`. While `test_burgers_newton` parametrises over `"SingleMemory"`, it never enables that schedule, so no SingleMemory validity comparison existed in the suite.

The TEMPORARY commit points `pyadjoint-ad` at the #248 branch so CI exercises the fix; it should be dropped, restoring the version pin, once #248 is merged and released. Since that branch sits on top of dolfin-adjoint/pyadjoint#257, CI will also show the pre-existing clear-down assertion failures that #5157 addresses, which are unrelated to these tests. As with the #257 companion, this should presumably end up on `release` once the ordering is settled.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -36,7 +36,7 @@ dependencies = [
   "petsctools>=2026.0",
   "pkgconfig",
   "progress",
-  "pyadjoint-ad>=2026.4.0",
+  "pyadjoint-ad @ git+https://github.com/sghelichkhani/pyadjoint.git@sghelichkhani/singlemem-checkpoint-clearing",
   "pycparser",
   "pytools[siphash]",
   "requests",
--- a/tests/firedrake/adjoint/test_checkpointing_multistep.py
+++ b/tests/firedrake/adjoint/test_checkpointing_multistep.py
@@ -4,7 +4,8 @@ from firedrake import *
 from firedrake.adjoint import *
 from .test_burgers_newton import _check_forward, \
     _check_recompute, _check_reverse
-from checkpoint_schedules import MixedCheckpointSchedule, StorageType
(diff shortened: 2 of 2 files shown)


sghelichkhani:
Reopened from a branch on the upstream repository as #5454, since I no longer need the fork. Same commits, no changes.

PR #5169, opened 2026-06-12
Enable SingleMemoryStorageSchedule in test_checkpointing_validity

sghelichkhani:
`test_checkpointing_validity` is parametrised over `"Revolve"`, `"SingleMemory"`,
`"NoneAdjoint"`, `"Mixed"` and `None`, but the body only enables the Revolve and Mixed
schedules, so the SingleMemory case silently compared two unscheduled tapes. This adds the
missing `elif` so that parametrisation does what its test id promises. Nothing new is caught
by this — the case passes against the released pyadjoint pin and against
dolfin-adjoint/pyadjoint#248 — it simply makes the test consistent with what it advertises.
`"NoneAdjoint"` is left unwired since `NoneCheckpointSchedule` does not support the adjoint
sweep this test performs.


sghelichkhani:
Reopened from a branch on the upstream repository as #5455, since I no longer need the fork. Same commits, no changes.

PR #5170, opened 2026-06-12
Clone appctx for the cached adjoint replay solvers

sghelichkhani:
When an annotated `NonlinearVariationalSolver` is created with an `appctx`, the replay machinery deep-copies the form coefficients but passes the appctx through untouched (forward solver) or drops it entirely (adjoint and TLM solvers). Any preconditioner that reads UFL expressions out of the appctx, MassInvPC being the obvious one for Schur complement Stokes, ends up looking at the user's original coefficients rather than the clones the tape keeps in sync, so during replay it builds the preconditioner from stale end-of-forward-run values.

This applies the forward coefficient replace map to UFL entries in the appctx when the cached solvers are built. Since the tangent and adjoint solvers derive their forms from the cloned forward problem, all three share the one map and so point at the same cloned coefficients that `update_dependencies` updates at replay time. The adjoint and TLM solvers fall back to the forward appctx when their own kwargs don't carry one; David confirmed this is always safe since the adjoint equation uses J^T (https://firedrakeproject.slack.com/archives/C1Q0Y6H8A/p1776171359074139). The `appctx` pop in `solve_init_params` stays, because the legacy `GenericSolveBlock` adjoint path solves with an assembled matrix and can't accept the kwarg; there's a regression test pinning that path too.

The new tests in `tests/firedrake/adjoint/test_appctx.py` check object identity between the appctx and the cached forms, and use a recording preconditioner to verify the values seen during forward and adjoint replay match the forward trajectory.

Note this targets the #4638 refactor branch, so it only reaches `main` when that merges. The same bug exists in the current release through the old `_ad_problem_clone` machinery, so we should probably do a separate PR basing the release branch to fix released Firedrake as well; I have that version of the fix ready on a separate branch.

Disclosure: parts of this change were prepared with the help of Claude Code (Opus 4.8).

### DIFF
--- a/firedrake/adjoint_utils/blocks/__init__.py
+++ b/firedrake/adjoint_utils/blocks/__init__.py
@@ -1,7 +1,7 @@
 from firedrake.adjoint_utils.blocks.assembly import AssembleBlock  # noqa F401
 from firedrake.adjoint_utils.blocks.solving import (  # noqa F401
-    GenericSolveBlock, SolveLinearSystemBlock, ProjectBlock,
-    SupermeshProjectBlock, SolveVarFormBlock,
+    CachedSolverBlock, GenericSolveBlock, SolveLinearSystemBlock,
+    ProjectBlock, SupermeshProjectBlock, SolveVarFormBlock,
     NonlinearVariationalSolveBlock
 )
 from firedrake.adjoint_utils.blocks.function import (  # noqa F401
--- a/firedrake/adjoint_utils/blocks/solving.py
+++ b/firedrake/adjoint_utils/blocks/solving.py
@@ -66,7 +66,6 @@ class CachedSolverBlock(Block):
         self.adj_residual = adj_residual
 
         self.adj_sol = adj_sol
-        self.adj_sol_buf = adj_sol.copy(deepcopy=True)
         self.adj2_sol = adj2_sol
         self.tlm_output = tlm_output
         self.d2Fdu2_form = d2Fdu2_form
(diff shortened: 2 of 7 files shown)


sghelichkhani:
Since this only reaches main when #4638 merges and the bug also affects released Firedrake, I've opened #5171 with the same fix expressed against the current release machinery.

connorjward:
Why are there two PRs for this? If this goes into `release` then it will make its way into `main` soon enough (potentially immediately after if needed). Can't this one just be closed?

sghelichkhani:
The two are not the same patch landing twice, so closing this one would drop the fix rather than defer it. #5171 against `release` repairs the legacy cloning path, namely [`_ad_problem_clone` and `_ad_adj_lvs_problem`](https://github.com/firedrakeproject/firedrake/blob/ffcad0c97c68955715ae5cca9cb276e0612ec525/firedrake/adjoint_utils/variational_solver.py#L118-L160), threading the replace maps back out so that the appctx can be cloned alongside the form coefficients.

This PR is the same fix expressed against the world that @JHopeCollins's `nlvs-hessian-fix` branch introduces, where those methods are superseded by the [`_ad_forward_cache`, `_ad_tangent_cache` and `_ad_adjoint_cache`](https://github.com/firedrakeproject/firedrake/blob/6f6fd151a725de1d9b2bef6b34d35ab694f448a4/firedrake/adjoint_utils/variational_solver.py#L129-L251) machinery. Once `release` reaches `main` and that branch rebases on top, the #5171 edits would land on code that has already been rewritten there, so the appctx fix does not carry over of its own accord and has to be reinstated against the cache, which is what this PR does. It additionally covers the tangent path, which has no counterpart in the legacy structure. I thought of it less as a duplicate of #5171 and more as a contribution that keeps the fix alive once that refactor supersedes the old cloning code, and I am happy to fold it into the branch in whatever form suits.

sghelichkhani:
Closing this as I am aware that https://github.com/firedrakeproject/firedrake/pull/4638 is (hopefully) going to deal with this.

PR #5171, opened 2026-06-12
Clone appctx for adjoint replay solvers (release)

sghelichkhani:
During tape replay of an annotated `NonlinearVariationalSolver`, the appctx is passed to the clone solvers still pointing at the user's original coefficients (forward) or dropped (adjoint), so preconditioners reading UFL expressions from it, e.g. MassInvPC, see stale values instead of the tape-synced clones. This clones the appctx with the same coefficient replace maps used for the forms. Same fix as #5170, which addresses this for `main` on top of the #4638 refactor; this is the equivalent for the current release. Carrying appctx into the adjoint solver is safe since the adjoint equation uses J^T (https://firedrakeproject.slack.com/archives/C1Q0Y6H8A/p1776171359074139).

Disclosure: parts of this change were prepared with the help of Claude Code (Opus 4.8).

### DIFF
--- a/firedrake/adjoint_utils/blocks/solving.py
+++ b/firedrake/adjoint_utils/blocks/solving.py
@@ -574,10 +574,12 @@ def solve_init_params(self, args, kwargs, varform):
                 raise NotImplementedError(
                     "Annotation of adaptive solves not implemented."
                 )
-            # The legacy adjoint solve goes through ``firedrake.solve``
-            # with an assembled matrix, which does not accept appctx.
-            # The variational solver mixin reinstates it (suitably
-            # cloned) for the adjoint LinearVariationalSolver.
+            # adj_kwargs feeds the assembled-matrix adjoint solve
+            # ``firedrake.solve(A, x, b, ...)``, whose kwarg validation
+            # rejects a top-level ``appctx`` (on that path appctx is read
+            # from ``solver_parameters`` instead). Drop it here; for solves
+            # driven through the variational solver mixin it is reinstated,
+            # suitably cloned, on the adjoint LinearVariationalSolver.
             self.adj_kwargs.pop("appctx", None)
 
     solver_params = kwargs.get("solver_parameters", None)
--- a/firedrake/adjoint_utils/blocks/solving.py
+++ b/firedrake/adjoint_utils/blocks/solving.py
@@ -574,6 +574,10 @@ def solve_init_params(self, args, kwargs, varform):
                 raise NotImplementedError(
                     "Annotation of adaptive solves not implemented."
                 )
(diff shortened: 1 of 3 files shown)


firedrake/adjoint_utils/blocks/solving.py
```diff
@@ -574,6 +574,10 @@ def solve_init_params(self, args, kwargs, varform):
                 raise NotImplementedError(
                     "Annotation of adaptive solves not implemented."
                 )
+            # The legacy adjoint solve goes through ``firedrake.solve``
+            # with an assembled matrix, which does not accept appctx.
+            # The variational solver mixin reinstates it (suitably
+            # cloned) for the adjoint LinearVariationalSolver.
```
connorjward: Is this true? A `LinearSolver` is now just a `LinearVariationalSolver` and passes through `kwargs`: https://github.com/firedrakeproject/firedrake/blob/main/firedrake/linear_solver.py#L12

sghelichkhani: Yes that comment was garbage, sorry! `LinearSolver` does take `appctx` as a `LinearVariationalSolver` subclass. The thing that actually bites is one layer up: the adjoint solve here goes through `solve(A, x, b)`, which dispatches to `_la_solve` and then `_extract_linear_solver_args`, and that validates kwargs against a fixed list that doesn't include `appctx`, so a top-level `appctx` raises `Illegal keyword argument 'appctx'`. On that path appctx only gets picked up out of `solver_parameters`, never as a top-level kwarg. So the pop is still needed, I'd just written down the wrong reason for it.

connorjward: I am wondering if the linear solver path could be adapted to accept an appctx argument. Then we wouldn't need any special handling here.

In general I think that this PR makes a lot of sense. I just don't want to bake in any unneeded technical debt.

JHopeCollins: Connor is correct, the right fix here is just to make `_extract_linear_solver_args` in `solving.py` accept an `appctx` kwarg.
We already allow using an `appctx` with the `LinearSolver` code path in the `solve` free function, but for some reason we expect it to be stuffed into the `solver_parameters` dictionary. This obviously doesn't match the pattern of any other solver/solve call.

Should just be a few lines to change this to expecting `appctx` in the `kwargs` like normal.

dham: Yes, we should allow this now and then deprecate passing the appctx the old way on main. That permanently fixes our API as well as sorting this out in the short term.

sghelichkhani:
Closing this in favour of https://github.com/firedrakeproject/firedrake/pull/5456. 5456 is consistent with the revs on this PR.

PR #5179, opened 2026-06-15
Fix SLATE test for pytest 9.1.0

connorjward:
Actual issue is reported
https://github.com/pytest-dev/pytest/issues/14591 but we only use indirect=True in one place so it's easier to just skip it altogether.


connorjward:
Forcing through as this is a critical yet super simple fix

PR #5180, opened 2026-06-15
Cache mesh bounding boxes correctly

leo-collins:
After changing mesh coordinates, accessing `mesh.bounding_box_coords` will now recalculate the bounding boxes.

Added a `mesh._check_coordinate_dat_version` helper because this sort of thing happens in a few places (bounding boxes, rtree, soon to be distributed rtree)


connorjward (changes requested):
The really nice way to do this is:
```py
def cached_property_until(cond: Callable[[Self], bool]):
  def decorator(func):
    def wrapper(self):
      if cond(self):
        return self._cache[some key]
      else:
        return self._cache.setdefault(some key, func(self))

cached_property_until_coords_change = cached_property_until(lambda self: self._saved_coords_dat_version == self.coordinates.dat.dat_version)
```
so you can have
```py
@cached_property_until_coords_change
def bounding_box_coords(self):
  ...
```

connorjward:
@achanbour this is very relevant to your work, and you have already probably done something similar. Any comments?

achanbour:
I currently have an attribute that keeps track of the mesh's topology `mesh._topology_version` (under `VertexOnlyMesh` only for now). It gets incremented by `MeshUpdater` (not yet in Firedrake) when the VOM points get redistributed or increase/decrease in number. This does not get incremented when the coordinates change (provided the topology stays the same).

In both regimes (coordinates change OR topology changes), the `MeshUpdater` clears the cached `mesh.bounding_box_coords`. I suppose with Leo's change, I can simply remove the line in the updater that clears this cache.

firedrake/function.py
```diff
@@ -743,13 +742,19 @@ def __init__(self, mesh: MeshGeometry, points: np.ndarray | list, tolerance: flo
...
         self.mesh = mesh
-
         self.redundant = redundant
         self.missing_points_behaviour = missing_points_behaviour
-        self.tolerance = tolerance
-        self.vom = VertexOnlyMesh(
-            mesh, self.points, missing_points_behaviour=missing_points_behaviour,
-            redundant=redundant, tolerance=tolerance
+        if tolerance is not None:
+            mesh.tolerance = tolerance
```
connorjward, line 748: Why is this only conditionally set? Might this lead to errors if someone tries to access `mesh.tolerance`?

firedrake/function.py
```diff
@@ -787,19 +792,9 @@ def evaluate(self, function: Function) -> np.ndarray | Tuple[np.ndarray, ...]:
         if function.function_space().ufl_element().family() == "Real":
             return function.dat.data_ro
 
-        function_mesh = function.function_space().mesh().unique()
-        if function_mesh is not self.mesh:
+        if function.function_space().mesh().unique() is not self.mesh:
             raise ValueError("Function mesh must be the same Mesh object as the PointEvaluator mesh.")
-        if coord_changed := function_mesh.coordinates.dat.dat_version != self.mesh._saved_coordinate_dat_version:
-            # TODO: This is here until https://github.com/firedrakeproject/firedrake/issues/4540 is solved
```
connorjward, line 794: Is the issue going to be closed then?

firedrake/function.py
```diff
@@ -787,19 +792,9 @@ def evaluate(self, function: Function) -> np.ndarray | Tuple[np.ndarray, ...]:
...
-            self.mesh = function_mesh
-        if tol_changed := self.mesh.tolerance != self.tolerance:
-            self.tolerance = self.mesh.tolerance
-        if coord_changed or tol_changed:
-            self.vom = VertexOnlyMesh(
-                self.mesh, self.points, missing_points_behaviour=self.missing_points_behaviour,
-                redundant=self.redundant, tolerance=self.tolerance
-            )
+        vom = self.vom
```
connorjward: This line doesn't really need to be here. I'd just use `self.vom` in the right places.

firedrake/function.py
connorjward: And in general I think that this is really bad software design. The tolerance should probably be an argument to things like `bounding_box_coords` (use [`cachedmethod`](https://cachetools.readthedocs.io/en/stable/#cachetools.cachedmethod)) instead of a mutable attribute.

I have a nice implementation in pyop3 so happy for this to wait. If you agree I think we should open an issue.

tests/firedrake/regression/test_point_eval_fs.py
```diff
@@ -243,3 +243,11 @@ def test_changing_coordinates_invalidates_rtree():
     saved_rtree = mesh.rtree
     mesh.coordinates.assign(mesh.coordinates * 2)
     assert mesh.rtree != saved_rtree
+
+
+def test_changing_coordinates_invalidates_bounding_box():
+    mesh = UnitSquareMesh(2, 2)
+
+    saved_bounding_box_coords = mesh.bounding_box_coords
```
connorjward, line 251: Also test that the cache works in the first place?
```py
assert mesh.bounding_box_coords is mesh.bounding_box_coords
```

firedrake/utils.py
```diff
@@ -264,3 +266,23 @@ def check_netgen_installed() -> None:
             "are installed and available to Firedrake (see "
             "https://www.firedrakeproject.org/install.html#netgen)."
         )
+
+
+def cached_property_until(value: Callable[[Self], Hashable]):
+    """Decorator for a property that is cached until some value changes.
+    For examples of usage, see PointEvaluator and MeshGeometry.
```
connorjward: Being fussy, but it wouldn't be hard to include a simple example or a more detailed docstring here.

firedrake/utils.py
```diff
@@ -264,3 +266,23 @@ def check_netgen_installed() -> None:
             "are installed and available to Firedrake (see "
             "https://www.firedrakeproject.org/install.html#netgen)."
         )
+
+
+def cached_property_until(value: Callable[[Self], Hashable]):
```
connorjward:
```suggestion
def cached_property_until(key: Callable[[Self], Hashable]):
```
think it's clearer

connorjward (changes requested):

firedrake/function.py
leo-collins: > Why is this only conditionally set? 

We now always use the meshes tolerance by passing `tolerance=None` to `VertexOnlyMesh`. If the user specifies a different tolerance when creating a `PointEvaluator` then we change `mesh.tolerance` so that the VOM builds with that tolerance. This allows us to use `cached_property_until` with the mesh tolerance.

I agree that this whole mutable `mesh.tolerance` business is gross. There's a whole bunch of this disgusting pattern
```
if tolerance is None:
    tolerance = mesh.tolerance
else:
    mesh.tolerance = tolerance
```

around (e.g. in the `VertexOnlyMesh` constructor, `_pic_swarm_in_mesh`,  `locate_cells_ref_coords_and_dists`, etc.) that I'd love to remove. However I think it is a separate issue to this PR so I'm happy to wait.

connorjward: Please open an issue for it then resolve this comment

leo-collins: https://github.com/firedrakeproject/firedrake/issues/5204

firedrake/function.py
leo-collins: No. I've moved the comment elsewhere.

firedrake/function.py
leo-collins: Done.

tests/firedrake/regression/test_point_eval_fs.py
leo-collins: Done.

firedrake/utils.py
leo-collins: Done.

firedrake/utils.py
leo-collins: Done.

firedrake/function.py
```diff
@@ -743,13 +742,20 @@ def __init__(self, mesh: MeshGeometry, points: np.ndarray | list, tolerance: flo
...
+
+    @cached_property_until(
+        lambda self: (self.mesh.coordinates.dat.dat_version, self.mesh.tolerance)
+    )
+    def vom(self) -> MeshGeometry:
+        """The VOM used for point evaluation. This is cached until the mesh coordinates or tolerance change."""
+        # We invalidate when the parent mesh moves until https://github.com/firedrakeproject/firedrake/issues/4540 is fixed.
+        return VertexOnlyMesh(
+            self.mesh, self.points, missing_points_behaviour=self.missing_points_behaviour,
+            redundant=self.redundant, tolerance=None
```
connorjward, line 758: Why is `tolerance=None`?

firedrake/mesh.py
```diff
@@ -2477,9 +2473,9 @@ def clear_rtree(self):
 
         Use this if you move the mesh (for example by reassigning to
         the coordinate field)."""
-        self._rtree = None
+        self._rtree_cache = None
```
connorjward, line 2492: What's this for?

firedrake/function.py
leo-collins: >We now always use the meshes tolerance by passing tolerance=None to VertexOnlyMesh. If the user specifies a different tolerance when creating a PointEvaluator then we change mesh.tolerance so that the VOM builds with that tolerance. This allows us to use cached_property_until with the mesh tolerance.
 
 This will go away after https://github.com/firedrakeproject/firedrake/issues/5204

connorjward: Ah yes

firedrake/mesh.py
leo-collins: `cached_property_until` stores the rtree in `self._rtree_cache`. That method `clear_rtree` is public API (I know thetis are using it). Maybe it can be deprecated now, I'm not sure.

connorjward: Without context this line makes no sense. I would prefer to hide it as much as possible. Why not:
```py
def clear_rtree(self):
  # comment about why this does the right thing
  self._rtree_cache = None
```

We should definitely look to deprecate `clear_rtree` because we can manage it ourselves.

connorjward (approved):

PR #5183, opened 2026-06-16
cache `visible_ranks` on topology

leo-collins:
The `visible_ranks` array used for point location only depends on the mesh topology

### DIFF
--- a/firedrake/mesh.py
+++ b/firedrake/mesh.py
@@ -1765,7 +1765,7 @@ class MeshTopology(AbstractMeshTopology):
             return composed_map, integral_type
 
     @cached_property
-    def visible_ranks(self):
+    def _visible_ranks(self):
         # Get parent mesh rank ownership information.
         visible_ranks = np.empty(self.cell_set.total_size, dtype=IntType)
         visible_ranks[:self.cell_set.size] = self.comm.rank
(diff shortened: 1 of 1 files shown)


firedrake/mesh.py
```diff
@@ -1764,6 +1764,18 @@ def trans_mesh_entity_map(self, base_mesh, base_integral_type, base_subdomain_id
             composed_map, integral_type, _ = self.submesh_map_composed(base_mesh, base_integral_type, base_subset_points)
             return composed_map, integral_type
 
+    @cached_property
+    def visible_ranks(self):
```
```suggestion
    def _visible_ranks(self):
```
This should probably be a private attribute no?
[PEP8](https://peps.python.org/pep-0008/#designing-for-inheritance) says
> Always decide whether a class’s methods and instance variables (collectively: “attributes”) should be public or non-public. If in doubt, choose non-public; it’s easier to make it public later than to make a public attribute non-public.

connorjward (changes requested):
Otherwise fine

firedrake/mesh.py
leo-collins: Done

connorjward (approved):

connorjward (approved):

PR #5185, opened 2026-06-17
Default to pre_apply_bcs=False

### LINKED ISSUE #5177: SNES updates can be inconsistent with default DirichletBC treatment
`solve()` supports different ways of handling `DirichletBC` through the `pre_apply_bcs` and `restrict` arguments.

1. `restrict=False, pre_apply_bcs=True` (the current default): Pre-applies the bcs and sets up a SNES on the full space. It assumes that the KSP solutions of the linearised problem satisfy the homogenouos DirichletBCs. This assumption breaks when the PC couples the equations on the interior with the equations on the BC rows (just an identity block on the diagonal). The assumption is rarely broken, we have only observed this with `-pc_type gamg` that couples the BCs with the interior on the coarser levels.

2. `restrict=True, pre_apply_bcs=True` Pre-applies the bcs and sets up a SNES on the restricted space. Here the KSP and PC have no way of coupling the equations of the interior with the equations of the BCs.

3. `restrict=False, pre_apply_bcs=False` Allows to linearise around a state that does not satisfy the BCs, and consistently appends the BC residual to the nonlinear system of equations. Pre-applying BCs might lead to initial linearisation points with sharp gradients, which could cause a NaN in the initial residual, and this model prevents this. This model is the default in other libraries such as old FEniCS, Deal.II, and MFEM.

The combination `restrict=True, pre_apply_bcs=False` does not make sense, the restricted problem needs to be formulated in a space with the correct BC data.

The default model `restrict=False, pre_apply_bcs=True` (1.) exposes two breaking behaviors:
a. It allows the PC to couple BCs with the interior, but the bcs are not encoded in the SNES residual.
b. It can silently lead to bad initial guesses with sharp gradients that cause NaN in the first residual.

The proposal is to change the default to `restrict=False, pre_apply_bcs=False` (3.) which does not suffer from those issues. Moreover, we should entirely disallow (1.). Options (2.) and (3.) are then parametrised by a unified flag since `restrict=pre_apply_bcs`, which would then lead to a simplified API where `pre_apply_bcs` can be inferred from `restrict`.

pbrubeck:
# Description
Fixes #5177

This PR changes the default model for DirichletBC to `pre_apply_bcs=False`, deprecating `pre_apply_bcs=True`.

In the deprecation era, `pre_apply_bcs=True` only applies bcs at the beginning of the solve, and `_SNESContext` becomes independent of `pre_apply_bcs`, which means that the residual lifting term is always computed (it will be zero when the bcs are satisfied).

We also make sure that bcs are always pre applied if `restrict=True`, and for this case we never compute the residual lifting term.


firedrake/solving_utils.py
```diff
@@ -251,19 +246,26 @@ def __init__(self, problem,
...
+                # but this doubles the number of integrals to be compiled.
+                # We cannot replace u -> u - g at this point because J or L might depend on u.
+                # The most efficient solution is to reconstruct the lifted residual from scratch.
+                # However, compute_bc_lifting(J, u - g, L) will complain about action not taking
+                # a pure Coefficient/Argument, so we supply a TrialFunction and replace it with u-g afterwards.
+                test, trial = self.J.arguments()
+                Ftrial = problem.compute_bc_lifting(self.J, trial, L=problem.L)
+                self.F = ufl.replace(Ftrial, {trial: self._x - self._bc_residual})
+            else:
+                self.F -= problem.compute_bc_lifting(self.J, self._bc_residual)
```
pbrubeck, line 264: With `pre_apply_bcs=False`, we are adding this extra term to the residual `F` (more code to compile and execute). However this can be prevented in the linear case, as I explain in the comment. 

For the nonlinear case, there is a way to reduce execution time by assuming that `bc_residual` is only supported on cells that along the boundary. We could restrict the integrals in `J` onto that subset of cells.

pbrubeck: Note that `pre_apply_bcs=True` implies that `self._bc_residual = 0` throughout the solve (if the Newton updates satisfy homogenous BCs by solving the bc identity matrix exactly), and this is why we did not have this term before.

firedrake/solving_utils.py
```diff
@@ -251,19 +246,26 @@ def __init__(self, problem,
...
+        if not problem.restrict and next(problem.dirichlet_bcs(), None) is not None:
             # Delayed lifting of DirichletBCs
             self._bc_residual = Function(self._x.function_space())
-            if problem.is_linear:
-                # Drop existing lifting term from the residual
-                assert isinstance(self.F, ufl.BaseForm)
-                self.F = ufl.replace(self.F, {self._x: ufl.zero(self._x.ufl_shape)})
-
-            self.F -= problem.compute_bc_lifting(self.J, self._bc_residual)
+            if problem.is_linear and hasattr(problem, "L"):
```
pbrubeck, line 252: Ideally we would be testing `isinstance(problem, LinearVariationProblem)`, but by doing this I am avoiding a cyclic import. I can have a go at fixing the cyclic import if that is preferred.

firedrake/variational_solver.py
```diff
@@ -90,9 +91,9 @@ def __init__(self, F, u, bcs=None, J=None,
                 bcs = J.bcs
         if bcs and any(isinstance(bc, EquationBC) for bc in bcs):
             restrict = False
-        self.restrict = restrict and bcs
+        self.restrict = restrict and bool(bcs)
```
pbrubeck, line 94: This was setting `self.restrict = bcs`

pbrubeck: I am casting `bool(bcs)` because the default is bcs=None, and it's too early to `extract_bcs(bcs)`, which parses bcs into an iterable

firedrake/solving_utils.py
```diff
@@ -168,9 +168,6 @@ class _SNESContext(object):
     transfer_manager
         Object that can transfer functions between levels,
         typically a :class:`~.TransferManager`.
-    pre_apply_bcs
```
pbrubeck, line 171: To unify the codepaths, `_SNESContext` will no longer branch on `pre_apply_bcs` and only branch on `restrict`.

firedrake/variational_solver.py
```diff
@@ -376,7 +379,7 @@ def solve(self, bounds=None):
             # Transfer the initial guess into the RestrictedFunctionSpace
             problem.u_restrict.assign(problem.u)
 
-        if self._ctx.pre_apply_bcs:
+        if self.pre_apply_bcs or problem.restrict:
```
pbrubeck, line 382: In the deprecation era, `pre_apply_bcs=True` only applies bcs at the beginning of the solve, and `_SNESContext` becomes independent of `pre_apply_bcs`, which means that the residual lifting term is always computed (it will be zero when the bcs are satisfied)

firedrake/solving_utils.py
pbrubeck: We could probably lazily assemble this individual term only when bc_residual is nonzero.

PR #5188, opened 2026-06-18
Handle MeshGeometry in cached solvers (-> #4638)

angus-g:
This threads `MeshGeometry` dependencies through the `CachedSolverBlock`. Because `CoordinateDerivative` is a bit strange, we do need to special case forms and flip signs here and there. Particularly for the Hessian forms it goes back to the action -> derivative approach (restrained by `CoordinateDerivative`) rather than derivative -> action -> adjoint.

I did also test Fireshape here (it passes, whereas before this PR it would fail on `test_L2tracking[FeMultiGridControlSpace]` and `test_TimeTracking`). @APaganini do you have anything that we should ensure we check related to this?

### DIFF
--- a/firedrake/adjoint_utils/blocks/solving.py
+++ b/firedrake/adjoint_utils/blocks/solving.py
@@ -40,8 +40,9 @@ HESSIAN = SolverType.HESSIAN
 
 
 class CachedSolverBlock(Block):
-    def __init__(self, forward_cache, tangent_cache, adjoint_cache, hessian_cache,
-                 ad_block_tag=None):
+    def __init__(
+        self, forward_cache, tangent_cache, adjoint_cache, hessian_cache, ad_block_tag=None
+    ):
         super().__init__(ad_block_tag=ad_block_tag)
 
         self.forward_cache = forward_cache
@@ -57,6 +58,12 @@ class CachedSolverBlock(Block):
     def _coefficient_dependencies(self, dependencies=None):
         dependencies = dependencies or self.get_dependencies()
         return dependencies[:len(self.forward_cache.replaced_deps)]
+
+    def _mesh_dependencies(self, dependencies=None):
+        dependencies = dependencies or self.get_dependencies()
+        len_replaced = len(self.forward_cache.replaced_deps)
+        len_meshes = len(self.forward_cache.meshes)
+        return dependencies[len_replaced:len_replaced + len_meshes]
 
     def _bc_dependencies(self, dependencies=None):
         dependencies = dependencies or self.get_dependencies()
@@ -95,6 +102,12 @@ class CachedSolverBlock(Block):
             if dep.output == self.forward_cache.func:
                 continue
             if dep.tlm_value is None:  # This dependency doesn't depend on the controls
+                continue
+            replaced_dep.assign(dep.tlm_value)
+
+        for replaced_dep, dep in zip(self.tangent_cache.mesh_tlms,
+                                     self._mesh_dependencies()):
+            if dep.tlm_value is None:
                 continue
             replaced_dep.assign(dep.tlm_value)
 
@@ -243,7 +256,7 @@ class CachedSolverBlock(Block):
 
         # tlm_input contribution
         for d2Fdmdu, dep in zip(self.hessian_cache.d2Fdmdu_forms,
-                                self._coefficient_dependencies()):
+                                self._coefficient_dependencies() + self._mesh_dependencies()):
             if dep.tlm_value is None:  # This dependency doesn't depend on the controls
                 continue
             if dep.output is self.forward_cache.func:  # Can't compute dependence on initial guess
@@ -279,7 +292,7 @@ class CachedSolverBlock(Block):
 
         relevant_d2Fdm2_forms = []
         for i, dep in relevant_dependencies:
-            if i >= len(self._coefficient_dependencies()):
+            if i >= len(self._coefficient_dependencies() + self._mesh_dependencies()):
                 continue
             if dep.tlm_value is None:
                 continue
@@ -293,7 +306,7 @@ class CachedSolverBlock(Block):
                      self.hessian_cache.dFdm_adj2_forms[idx],
                      *relevant_d2Fdm2_forms):
             if not form.empty():
-                hessian_output += firedrake.assemble(-form)
+                hessian_output += firedrake.assemble(form)
 
         return hessian_output
 
--- a/firedrake/adjoint_utils/variational_solver.py
+++ b/firedrake/adjoint_utils/variational_solver.py
@@ -446,10 +446,10 @@ class NonlinearVariationalSolverMixin:
             else:
                 dm = TrialFunction(m.function_space())
                 # XXX should we try inverting this back to the way it was before?
-                dFdm = derivative(-F, m, dm)
+                dFdm = derivative(F, m, dm)
 
-                dFdm_adj = expand_derivatives(action(adjoint(dFdm), adj_sol))
-                dFdm_adj2 = action(adjoint(dFdm), adj2_sol)
+                dFdm_adj = -expand_derivatives(action(adjoint(dFdm), adj_sol))
+                dFdm_adj2 = -action(adjoint(dFdm), adj2_sol)
 
             dFdm_adj2_forms.append(dFdm_adj2)
 
--- a/firedrake/adjoint_utils/variational_solver.py
+++ b/firedrake/adjoint_utils/variational_solver.py
@@ -7,6 +7,7 @@ from firedrake.adjoint_utils.blocks.solving import solve_init_params
 from firedrake.ufl_expr import derivative, adjoint, action
 from ufl import replace, Action
 from ufl.algorithms import expand_derivatives
+from ufl.domain import extract_domains
 from types import SimpleNamespace
 from collections import namedtuple
 
@@ -16,6 +17,7 @@ ForwardSolveRecomputeCache = namedtuple(
     field_names=[
         "func",
         "bcs",
+        "meshes",
         "solver",
         "replaced_deps",
     ]
@@ -41,6 +43,7 @@ TangentSolveRecomputeCache = namedtuple(
         "dFdm_forms",
         "solver",
         "replaced_tlms",
+        "mesh_tlms",
         "dFdu",
     ]
 )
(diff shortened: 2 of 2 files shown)


firedrake/adjoint_utils/blocks/solving.py
```diff
@@ -40,8 +40,9 @@ class SolverType(Enum):
 
 
 class CachedSolverBlock(Block):
-    def __init__(self, forward_cache, tangent_cache, adjoint_cache, hessian_cache,
-                 ad_block_tag=None):
+    def __init__(
+        self, forward_cache, tangent_cache, adjoint_cache, hessian_cache, ad_block_tag=None
+    ):
```
angus-g, line 45: Oops I was going to type-annotate these but that introduces a circular dependency, I should probably go back to an `if TYPE_CHECKING` import.

APaganini:
I’m currently on leave. I can look into this next Thursday.

firedrake/adjoint_utils/blocks/solving.py
JHopeCollins: Don't worry about it, we can do a sweep for type checking before the main PR is merged.

firedrake/adjoint_utils/blocks/solving.py
```diff
@@ -98,6 +105,12 @@ def update_tlm_dependencies(self):
                 continue
             replaced_dep.assign(dep.tlm_value)
 
+        for replaced_dep, dep in zip(self.tangent_cache.mesh_tlms,
+                                     self._mesh_dependencies()):
+            if dep.tlm_value is None:
+                continue
+            replaced_dep.assign(dep.tlm_value)
```
JHopeCollins, line 112: Do we ever hit this line for a mesh dependency in the test suite?
I see that the `replaced_dep` is a `Function` matching the `mesh.coordinates`. Is the `tlm_value` for a mesh also a coordinate `Function`?

firedrake/adjoint_utils/variational_solver.py
```diff
@@ -318,6 +345,13 @@ def _ad_adjoint_cache(self):
 
             dFdm_adj_forms.append(dFdm)
 
+        for m in self._ad_forward_cache.meshes:
+            X = SpatialCoordinate(m)
+            # we can't take the CoordinateDerivative of an Action, so we have
+            # to invert this form compared to the expression above
+            dFdm = derivative(action(-F, adj_sol), X, TestFunction(m._ad_function_space()))
+            dFdm_adj_forms.append(dFdm)
```
JHopeCollins, line 362: Could you explain this comment a bit? It looks like we are in fact taking the derivative of an action wrt the coordinates here. Does it not resolve to a `CoordinateDerivative`?

firedrake/adjoint_utils/variational_solver.py
```diff
@@ -375,20 +411,39 @@ def _ad_hessian_cache(self):
...
             d2Fdmdu_forms.append(d2Fdmdu)
 
+        for m, dm in zip(self._ad_forward_cache.meshes,
+                         self._ad_tangent_cache.mesh_tlms):
+            X = SpatialCoordinate(m)
+            d2Fdmdu = expand_derivatives(
+                derivative(dFdu_adj, X, dm)
+            )
+
+            d2Fdmdu_forms.append(d2Fdmdu)
```
JHopeCollins, line 430: This is quite nit-picky, feel free to ignore if you don't agree.

The only difference in the body of this loop from the one above is the `m_ = SpatialCoordinate(m) if (m is mesh) else m` logic for what to take the derivative with respect to.

Would it be clearer to combine them into a single loop and have an if-block to set m and make it clear that the form calculation is identical? Or does it just make it too messy?

tests/firedrake/adjoint/test_dynamic_meshes.py
```diff
@@ -31,13 +31,17 @@ def test_dynamic_meshes_2D():
     k = Constant(1/dt)
     F = k*inner(u-u0, v)*dx + inner(grad(u), grad(v))*dx - f*v*dx
     u1 = Function(V)
-    solve(lhs(F) == rhs(F), u1)
+    lvp1 = LinearVariationalProblem(lhs(F), rhs(F), u1)
+    lvs1 = LinearVariationalSolver(lvp1)
+    lvs1.solve()
```
JHopeCollins: Could we parameterise these tests to run with either `solve` free function or a `Solver`?

JHopeCollins (commented):
In general this looks fine, thanks for working through all the cases. I have a few very very minor comments/questions.

I see that there are failures here related to the `solve` free function, and failures in the `solve` PR you think will be fixed by this PR. Which order would you like to merge in?
Merge this mesh PR into the main PR first, then update the `solve` PR and hopefully the test failures will be fixed, or the other way around?

firedrake/adjoint_utils/blocks/solving.py
angus-g: Yeah, there are failures in the dynamic mesh and shape derivative tests if I remove this line. I think it is a coordinate `Function`, e.g.:
```
Coefficient(WithGeometry(FunctionSpace(<firedrake.mesh.MeshTopology object at 0x7f3159997050>, VectorElement(FiniteElement('Lagrange', triangle, 1), dim=3), name=None), Mesh(VectorElement(FiniteElement('Lagrange', triangle, 1), dim=3), 538)), 1564)
```

firedrake/adjoint_utils/variational_solver.py
angus-g: I don't quite remember why I mentioned action in the comment, but you can't take the adjoint of dF/dX:
```
../../firedrake/adjoint_utils/variational_solver.py:362: in _ad_adjoint_cache
    dFdm = adjoint(derivative(-F, X))
           ^^^^^^^^^^^^^^^^^^^^^^^^^^
../../firedrake/ufl_expr.py:349: in adjoint
    return ufl.adjoint(form, reordered_arguments, derivatives_expanded=derivatives_expanded)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
/opt/fenics-ufl/ufl/formoperators.py:248: in adjoint
    return compute_form_adjoint(form, reordered_arguments)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
/opt/fenics-ufl/ufl/algorithms/formtransformations.py:564: in compute_form_adjoint
    return map_integrands(Conj, replace(form, {v: reordered_v, u: reordered_u}))
                                ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
/opt/fenics-ufl/ufl/algorithms/replace.py:96: in replace
    return map_integrand_dags(Replacer(mapping2), e)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
/opt/fenics-ufl/ufl/algorithms/map_integrands.py:86: in map_integrand_dags
    return map_integrands(
/opt/fenics-ufl/ufl/algorithms/map_integrands.py:30: in map_integrands
    map_integrands(function, itg, only_integral_type) for itg in form.integrals()
    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
/opt/fenics-ufl/ufl/algorithms/map_integrands.py:39: in map_integrands
    return itg.reconstruct(function(itg.integrand()))
                           ^^^^^^^^^^^^^^^^^^^^^^^^^
/opt/fenics-ufl/ufl/algorithms/map_integrands.py:87: in <lambda>
    lambda expr: map_expr_dag(function, expr, compress), form, only_integral_type
                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
/opt/fenics-ufl/ufl/corealg/map_dag.py:42: in map_expr_dag
    (result,) = map_expr_dags(
/opt/fenics-ufl/ufl/corealg/map_dag.py:112: in map_expr_dags
    r = handlers[v._ufl_typecode_](v)
        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

self = <ufl.algorithms.replace.Replacer object at 0x7f866c1218b0>
o = CoordinateDerivative(Product(IntValue(-1), Product(Product(Coefficient(WithGeometry(FunctionSpace(<firedrake.mesh.Mesh...ahedron, 1), dim=3), name=None), Mesh(VectorElement(FiniteElement('Lagrange', tetrahedron, 1), dim=3), 1)), 1, None))))

    def coefficient_derivative(self, o):
        """Replace a coefficient derivative."""
>       raise ValueError("Derivatives should be applied before executing replace.")
E       ValueError: Derivatives should be applied before executing replace.

/opt/fenics-ufl/ufl/algorithms/replace.py:66: ValueError
```

No amount of `expand_derivatives` will help: I think you need to have `CoordinateDerivative` as the outermost operation which is why I had to invert `action` inside the expression...

firedrake/adjoint_utils/variational_solver.py
angus-g:
```suggestion
        deps = self._ad_forward_cache.replaced_deps
        meshes = self._ad_forward_cache.meshes

        replaced = deps + tuple(SpatialCoordinate(m) for m in meshes)
        tlms = self._ad_tangent_cache.replaced_tlms + self._ad_tangent_cache.mesh_tlms

        for m, dm in zip(replaced, tlms):
            d2Fdmdu = expand_derivatives(
                derivative(dFdu_adj, m, dm))

            d2Fdmdu_forms.append(d2Fdmdu)
```

Ah yeah, I agree. I didn't want the `zip` expression to be huge, but maybe something like this reads more clearly?

JHopeCollins: The loop body is clear but then there's several steps between data and operations so I think that cancels out. That's fine, it was only a minor thought, your original version is fine as is.

JHopeCollins (approved):

PR #5189, opened 2026-06-18
Fix Jp + multigrid

pbrubeck:
# Description

Previously, the preconditioing bilinear form `Jp` was missing from the coarse levels because the coarse level DMs were only passing `J`.

`DMKSPSetCreateOperators` was recently added as API to pass both `J` and `Jp` on nested KSP solvers. 

Depends on
https://gitlab.com/petsc/petsc/-/merge_requests/9384 (merged)
https://gitlab.com/petsc/petsc/-/merge_requests/9417 (merged)


firedrake/solving_utils.py
```diff
@@ -524,6 +524,14 @@ def form_jacobian(snes, X, J, P):
         ctx.set_nullspace(ctx._nullspace_T, ises, transpose=True, near=False)
         ctx.set_nullspace(ctx._near_nullspace, ises, transpose=False, near=True)
 
+    @staticmethod
+    def create_operators(ksp):
+        dm = ksp.getDM()
+        ctx = dmhooks.get_appctx(dm)
+        A = ctx._jac.petscmat
+        P = A if ctx.Jp is None else ctx._pjac.petscmat
```
stefanozampini: I believe passing `None` is cleaner code

```suggestion
        P = None if ctx.Jp is None else ctx._pjac.petscmat
```

Is this creating new matrices or just passing the ones stored inside?
Ideally, callbacks to `create` operators should create new ones, rather than passing references to stored matrices. Ideally, one can do
```
DMCreateMatrix(dm,&A1);
// ...populate A1
DMCreateMatrix(dm,&A2);
// ...populate A2

// use A1 and A2 independently
```

I believe it is conceptually wrong to pass matrices already stored in the context. One should duplicate the internal matrices instead

dham (approved):

PR #5195, opened 2026-06-20
replace notebooks with jupytext

### LINKED ISSUE #4735: Use jupytext to replace Jupyter notebooks
As discussed in [this meeting](https://github.com/firedrakeproject/firedrake/wiki/Firedrake-meeting-2025-11-18#drs-example-codes), we want to transition from using Jupyter notebooks for our tutorials to [jupytext](https://jupytext.readthedocs.io/en/latest/). This is a good idea because:

* Tracking Jupyter notebooks in git is extremely messy
* It would be nice to have the ability to extract notebooks as executable Python scripts

The main thing that we need to be aware of is that we render the notebooks on the website including full output ([example](https://github.com/firedrakeproject/firedrake/blob/release/docs/notebooks/01-spd-helmholtz.ipynb)). With jupytext we would have to generate, run and publish the notebooks (e.g. to [here](https://github.com/firedrakeproject/notebooks)) whenever we push to release.

Note that we still intend to use pylit for our demos. The demos are written in RST for which jupytext does not appear to have support and so it would be a substantial rewrite for no realistic benefit.

danshapero:
Fixes #4735 . I pointed Claude at this and it got most of the way there. It saves about 6MB of rendered notebooks. I looked over the generated HTML docs and made a few minor fixes: adding top-level headings to the notebooks; removing some unnecessary preamble code; and fixing some goofed up link formatting.

There are still some warnings in the rendered sum-factorization notebook but those are because of FInAT.


danshapero:
The doc build is failing because of a link from the FEnICS 2015 conference to the Transport for London website.

docs/source/notebooks.rst
```diff
@@ -14,159 +16,40 @@ To run the notebooks, you will need to `install jupyter
...
-These notebooks are maintained in the Firedrake repository, so all the
-material is available in your Firedrake installation source
-directory.  If you installed in ``Documents/firedrake``, then the
-notebooks are in the directory
-``Documents/firedrake/src/firedrake/docs/notebooks``.
+These notebooks are maintained in the Firedrake repository as
+`jupytext <https://jupytext.readthedocs.io>`__ ``py:percent`` scripts, so all
+the material is available in your Firedrake installation source directory.  If
+you installed in ``Documents/firedrake``, then the notebooks are in the
+directory ``Documents/firedrake/src/firedrake/docs/notebooks``.  Jupyter (with
```
connorjward: This isn't correct since we don't prescribe installing inside a venv any more

.github/workflows/push.yml
```diff
@@ -55,8 +55,16 @@ jobs:
...
+        run: jupytext --to ipynb firedrake/docs/notebooks/*.py
+
       - name: Copy notebooks
-        run: cp -r firedrake/docs/notebooks/* notebooks/
+        # Ship the generated notebooks and their data files, but not the
+        # jupytext .py sources or config.
+        run: |
+          cp firedrake/docs/notebooks/*.ipynb notebooks/
+          cp -r firedrake/docs/notebooks/image notebooks/
+          cp firedrake/docs/notebooks/*.geo firedrake/docs/notebooks/*.msh notebooks/
```
connorjward: I worry that this isn't robust to us introducing new data files with different suffixes. That would be a hard bug to track down. Can we not just copy everything over?

docs/source/notebooks.rst
```diff
@@ -14,159 +16,40 @@ To run the notebooks, you will need to `install jupyter
...
+Google Colab through your web browser, without installing Firedrake.  Each
+rendered notebook below carries a link to its Colab version.
+
+The notebooks
+-------------
+
+.. toctree::
+   :maxdepth: 1
+
+   notebooks/01-spd-helmholtz
```
connorjward, line 115: Does this lose the expository prose above? We should keep that as it tells users what each notebook does.

connorjward (changes requested):
In general I think this is great. Thanks! I do need to review the generated docs but that has to wait for the workflow to rerun.

docs/source/notebooks.rst
danshapero: You're right, I put it back in.

.github/workflows/push.yml
danshapero: We want to exclude the .py sources of the notebooks and a .toml configuration file. This now uses `rsync` to copy everything except what we want to exclude.

connorjward (dismissed):
I think I'm happy with everything now. Thanks again!

I really like the new way of referencing the Colab scripts. This is far more obvious.

connorjward (dismissed):

connorjward:
I just pushed the very very minor rendering issue that we noticed. I will try and merge this ASAP.

connorjward (approved):

PR #5203, opened 2026-06-24
Refactor test parameterization for some IO tests

connorjward:
This means that we see things like
```
          <Function test_io_function_base[triangle_small-P-1]>
          <Function test_io_function_base[triangle_small-P-6]>
          <Function test_io_function_base[triangle_small-DP-0]>
```
instead of
```
          <Function test_io_function_base[cell_family_degree0]>
          <Function test_io_function_base[cell_family_degree1]>
          <Function test_io_function_base[cell_family_degree2]>
```

### DIFF
--- a/scripts/firedrake-check
+++ b/scripts/firedrake-check
@@ -26,7 +26,7 @@ TESTS = {
     ),
     2: (
         # HDF5/checkpointing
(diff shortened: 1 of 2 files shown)


JHopeCollins (approved):
Yes, much more readable.

PR #5206, opened 2026-06-25
MeshHierarchy: assemble prolongation into MatAIJ

pbrubeck:
# Description
Assemble the prolongantion matrix into aij via `TransferManager(mat_type="aij")`


firedrake/mg/embedded.py
```diff
@@ -362,3 +371,23 @@ def restrict(self, source, target):
...
+            return self._mat_cache[key]
+        except KeyError:
+            P = assemble_prolongation_aij(Vc, Vf)
+            return self._mat_cache.setdefault(key, P)
+
+    def prolong_aij(self, uc, uf):
+        Vc = uc.function_space()
+        Vf = uf.function_space()
+        P = self.prolongation_matrix(Vc, Vf)
+        return firedrake.assemble(firedrake.action(P, uc), tensor=uf)
```
JHopeCollins: Will this method get called at every time we progress to the next level during the MG sweep?
If so then we should cache the `action` to avoid duplicating symbolic work at every call.

pbrubeck: P is cached. The symbolic work is quite insignificant. We cannot rely on the same Function `uc` being passed every time.

pbrubeck: If you care too  much about avoiding symbolic work, the more performant alternative would be to direclty call `Mat.mult` here, but that is not very firedrakey

JHopeCollins: If we can avoid it then we should. The symbolic work cannot be parallelised so it will damage the strong scaling. Especially if the smoother on each level is very cheap, then the symbolic work will not necessarily be insignificant.

pbrubeck: I think this is a premature optimization, however it's a two-liner and I'm happy to concede

JHopeCollins: We could create a `uc_cached = Function(Vc)` and cache that `Function` and `cached_action=action(P, uc_cached)`. Then on subsequent calls we do `uc_cached.assign(uc); assemble(cached_action, tensor=uf)`

firedrake/solving_utils.py
```diff
@@ -312,18 +313,19 @@ def transfer_manager(self):
...
-                managername = opts[prefix + "fas_transfer_manager"]
-            else:
-                managername = None
+            opts = PETSc.Options(prefix)
+
+            def get_first_option(opts, prefixes, default=""):
+                return next((s for s in map(partial(opts.getString, default=""), prefixes) if s), default)
 
+            manager_prefixes = ("mg_transfer_manager", "fas_transfer_manager")
+            managername = get_first_option(opts, manager_prefixes, default=None)
```
connorjward: This is less readable than before.

firedrake/solving_utils.py
```diff
@@ -312,18 +313,19 @@ def transfer_manager(self):
...
         """
         if self._transfer_manager is None:
-            opts = PETSc.Options()
             prefix = self.options_prefix or ""
-            if opts.hasName(prefix + "mg_transfer_manager"):
-                managername = opts[prefix + "mg_transfer_manager"]
-            elif opts.hasName(prefix + "fas_transfer_manager"):
-                managername = opts[prefix + "fas_transfer_manager"]
-            else:
-                managername = None
```
connorjward, line 337: We should warn if we are ignoring an option.

connorjward, line 337: We should warn if we are ignoring an option.

tests/firedrake/multigrid/test_aij_transfer.py
```diff
@@ -0,0 +1,96 @@
...
+import pytest
+import numpy
+
+
+@pytest.fixture
+def rg():
+    return RandomGenerator(PCG64(seed=1234))
+
+
+@pytest.mark.parallel([1])
```
connorjward:
```suggestion
@pytest.mark.parallel([1, 2])
```

connorjward:
```suggestion
@pytest.mark.parallel([1, 2])
```

tests/firedrake/multigrid/test_aij_transfer.py
```diff
@@ -0,0 +1,96 @@
...
+        "mat_type": "aij",
+        "pc_type": "mg",
+        "mg_transfer_mat_type": "aij",
+        "mg_levels_ksp_type": "chebyshev",
+        "mg_levels_ksp_max_it": 2,
+        "mg_levels_pc_type": "jacobi",
+        "mg_coarse_ksp_type": "preonly",
+        "mg_coarse_pc_type": "lu",
+    }
+    solve(a == L, u, bcs=bc, solver_parameters=params)
```
connorjward: Inspect the solver transfer manager

connorjward: Inspect the solver transfer manager

connorjward (changes requested):

connorjward (changes requested):
...

firedrake/mg/embedded.py
pbrubeck: Sounds more complicated, I'd rather just use petsc4py here

firedrake/solving_utils.py
```diff
@@ -314,16 +314,28 @@ def transfer_manager(self):
...
             opts = PETSc.Options()
             prefix = self.options_prefix or ""
-            if opts.hasName(prefix + "mg_transfer_manager"):
-                managername = opts[prefix + "mg_transfer_manager"]
-            elif opts.hasName(prefix + "fas_transfer_manager"):
-                managername = opts[prefix + "fas_transfer_manager"]
-            else:
-                managername = None
 
+            def get_transfer_option(mg_name, fas_name, default=None):
```
pbrubeck, line 338: @connorjward are we happy with this code to read PETSc options?

firedrake/mg/embedded.py
```diff
@@ -362,3 +371,25 @@ def restrict(self, source, target):
...
+            return self._mat_cache.setdefault(key, P)
+
+    def prolong_aij(self, uc, uf):
+        Vc = uc.function_space()
+        Vf = uf.function_space()
+        P = self.prolongation_matrix(Vc, Vf)
+        with uc.dat.vec_ro as x, uf.dat.vec_wo as y:
+            P.petscmat.mult(x, y)
+
+    def restrict_aij(self, rf, rc):
```
connorjward: docstrings wouldn't hurt

firedrake/mg/interface.py
```diff
@@ -290,3 +294,74 @@ def inject(fine, coarse):
...
+
+def _bc_matches_space(bc, V):
+    fs = bc.function_space()
+    while fs.component is not None and fs.parent is not None:
+        fs = fs.parent
+    return fs == V
+
+
+@PETSc.Log.EventDecorator()
+def assemble_prolongation_aij(Vc, Vf, bcs=None):
```
connorjward, line 308: docstring

firedrake/mg/interface.py
```diff
@@ -290,3 +294,74 @@ def inject(fine, coarse):
...
+    if len(Vc) > 1 or len(Vc) > 1:
```
connorjward:
```suggestion
    if len(Vc) > 1 or len(Vf) > 1:
```

firedrake/mg/interface.py
```diff
@@ -290,3 +294,74 @@ def inject(fine, coarse):
...
+        raise NotImplementedError("Mixed spaces are handled through TransferManager")
```
connorjward: This is a weird error message that doesn't give much useful information to someone hitting it

firedrake/mg/kernels.py
```diff
@@ -253,6 +253,119 @@ def prolong_kernel(expression, Vf):
         return cache.setdefault(key, transfer_kernel)
 
 
+def prolong_matrix_kernel(Vc, Vf):
```
connorjward, line 256: docstring

firedrake/mg/kernels.py
```diff
@@ -253,6 +253,119 @@ def prolong_kernel(expression, Vf):
...
+        source_cell_inc = element.space_dimension()
+        source_stencil_inc = ncandidate * source_cell_inc
+        local_tensor_size = row_dim * source_stencil_inc
+        cell_tensor_size = row_dim * source_cell_inc
+
+        kernel_code = """#include <petsc.h>
+        %(to_reference)s
+        %(evaluate)s
+        __attribute__((noinline)) /* Clang bug */
+        static void pyop2_kernel_prolong_matrix(PetscScalar *A, const PetscScalar *X, const PetscScalar *Xc
```
connorjward, line 310: how is this C code any different to the other functions?

firedrake/solving_utils.py
connorjward: It's fine. But should be documented

firedrake/solving_utils.py
```diff
@@ -314,16 +314,28 @@ def transfer_manager(self):
...
+                elif has_fas:
+                    return opts[fas_name]
+                else:
+                    return default
+
+            managername = get_transfer_option("mg_transfer_manager", "fas_transfer_manager")
             if managername is None:
                 from firedrake import TransferManager
-                transfer = TransferManager(use_averaging=True)
+                mat_type = get_transfer_option("mg_transfer_mat_type", "fas_transfer_mat_type",
```
connorjward, line 354: But why do we support both? It would be neater to deprecate one.

connorjward (changes requested):

firedrake/solving_utils.py
pbrubeck: It makes sense to support both `mg_` and `fas_` prefixes to allow different transfer managers within a solver that uses both `mg_` and `fas_`. The distinction is that `fas_` is a SNES and `mg_` is a PC, they can be composed with each other. Usually one would be setting a unique transfer for the entire problem, but sometimes it makes sense to do something more speciailized for the `mg_` preconditioner.

firedrake/mg/kernels.py
pbrubeck: It assembles a matrix, it has a very different loop structure

firedrake/solving_utils.py
connorjward: Can we do something like give the transfer manager a prefix?

pbrubeck: I think so. But you are opening a can of worms that detachs from the goal of this PR. I can open a github issue if you like.

connorjward: I'd be OK with an issue and a detailed comment in the code explaining the situation and probably linking to the issue.

firedrake/mg/embedded.py
```diff
@@ -382,6 +391,57 @@ def restrict(self, source, target):
...
+            The source (fine grid) cofunction.
+        rc : Cofunction
+            The target (coarse grid) cofunction.
+
+        """
+        Vc = rc.function_space().dual()
+        Vf = rf.function_space().dual()
+        P = self._prolongation_matrix(Vc, Vf)
+        with rf.dat.vec_ro as x, rc.dat.vec_wo as y:
+            P.petscmat.multTranspose(x, y)
```
pbrubeck, line 444:
```suggestion
            P.petscmat.multTranspose(x, y)

```

firedrake/solving_utils.py
pbrubeck: Can we merge this?

firedrake/mg/embedded.py
```diff
@@ -45,10 +46,13 @@ def __init__(self, *, native_transfers=None, use_averaging=True):
         :arg use_averaging: Use averaging to approximate the
            projection out of the embedded DG space? If False, a global
            L2 projection will be performed.
+        :arg mat_type: The matrix assembly type for prolongation/restriction.
```
connorjward: numpydoc

firedrake/mg/embedded.py
```diff
@@ -382,6 +391,58 @@ def restrict(self, source, target):
...
 
+    def _prolongation_matrix(self, Vc, Vf):
+        """Assemble and cache the prolongation matrix mapping Vc to Vf.
+
+        Parameters
+        ----------
+        Vc : WithGeometry
+            The source (coarse grid) function space.
+        Vf : WithGeometry
+            The target (fine grid) function space.
```
connorjward, line 404: Returns?

firedrake/mg/interface.py
```diff
@@ -289,3 +293,96 @@ def inject(fine, coarse):
...
+
+    Returns
+    -------
+    AssembledMatrix
+        The prolongation matrix mapping Vc to Vf.
+
+    """
+    if len(Vc) > 1 or len(Vf) > 1:
+        raise NotImplementedError(
+            'Explicit aij prolongation matrix assembly is not impelemented for mixed function spaces. '
```
connorjward:
```suggestion
            'Explicit aij prolongation matrix assembly is not implemented for mixed function spaces. '
```

connorjward (changes requested):

firedrake/mg/embedded.py
pbrubeck: done

firedrake/mg/embedded.py
pbrubeck: done

connorjward (dismissed):

firedrake/mg/embedded.py
```diff
@@ -34,21 +35,26 @@ def __init__(self, ufl_element, value_shape):
...
+        ----------
+        native_transfers : dict, optional
+            A mapping from UFL elements to natively supported transfer
+            operators. Each value must be a three-tuple containing the
+            prolong, restrict, and inject operators.
+        use_averaging : bool, optional
+            Whether to use averaging to approximate the projection out of an
+            embedded DG space. If false, perform a global L2 projection.
+        mat_type : str, optional
+            The matrix assembly type for prolongation and restriction.
```
pbrubeck:
```suggestion
        native_transfers : dict
            A mapping from UFL elements to natively supported transfer
            operators. Each value must be a three-tuple containing the
            prolong, restrict, and inject operators.
        use_averaging : bool
            Whether to use averaging to approximate the projection out of an
            embedded DG space. If false, perform a global L2 projection.
        mat_type : str
            The matrix assembly type for prolongation and restriction.
```

PR #5208, opened 2026-06-29
Prevent Cython 3.2.6 from being installed in build

indiamai:
# Description
The Cython release 3.2.6 has a bug that is causing CI failures. [cython:7767](https://github.com/cython/cython/issues/7767)

Add constraint that ensures this release is not installed in the CI

### DIFF
--- a/.github/actions/install/action.yml
+++ b/.github/actions/install/action.yml
@@ -124,6 +124,9 @@ runs:
         python3 -m venv venv
         . venv/bin/activate
 
+        echo 'Cython!=3.2.6' > constraints.txt
+        export PIP_CONSTRAINT=constraints.txt
+
         : # Empty the pip cache to ensure that everything is compiled from scratch
         pip cache purge
 


indiamai:
CI failures are due to link check and netgen

pbrubeck (approved):

pbrubeck:
Thank you!

pbrubeck:
The Cython issue has been addressed in https://github.com/cython/cython/pull/7768. I don't know whether we should just wait for the next release...

indiamai:
Yeah I saw that but not sure when they'll actually release? plus the way this is written, when they release it will switch to installing that, and we can then delete these changes whenever we remember.

pbrubeck:
> Yeah I saw that but not sure when they'll actually release? plus the way this is written, when they release it will switch to installing that, and we can then delete these changes whenever we remember.

Let's create an issue so we don't forget

indiamai:
Agree. [5210](https://github.com/firedrakeproject/firedrake/issues/5210)

PR #5211, opened 2026-06-30
Make coloring as default if we are using petscASM.

Molin-Han:
# Description

In ASMStarPC and ASMVankaPC, make the coloring feature as the default option if we are using 'petscasm'. If we are using `tinyasm`, we do not use the coloring.

### DIFF
--- a/firedrake/preconditioners/asm.py
+++ b/firedrake/preconditioners/asm.py
@@ -150,9 +150,11 @@ class ASMStarPC(ASMPatchPC):
     consists of all DoFs on the topological star of the mesh entity
     specified by `pc_star_construct_dim`.
 
-    Non-overlapping patches may be optionally grouped together via a
-    coloring of the mesh entities. This is specified via the option
-    `pc_star_use_coloring`.
+    Non-overlapping patches are grouped together via a
+    coloring of the mesh entities. 
+    This is the default behavior for ``backend=="petscasm"``.
+    The default may be overridden explicitly by setting
+    `pc_star_use_coloring` to `True` or `False`.
 
     The mesh entities in the patches may be reordered by applying a
     matrix reordering to the connectivity graph with the option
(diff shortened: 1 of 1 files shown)


firedrake/preconditioners/asm.py
```diff
@@ -171,10 +171,16 @@ def get_patches(self, V):
...
+        # Either use PETSc's ASM PC or use TinyASM (as simple ASM
+        # implementation designed to be fast for small block sizes).
         depth = opts.getInt("construct_dim", default=0)
         validate_overlap(mesh, depth, "star")
 
-        use_coloring = opts.getBool("use_coloring", default=False)
+        if backend == "petscasm":
+            use_coloring = opts.getBool("use_coloring", default=True)
+        else:
+            use_coloring = opts.getBool("use_coloring", default=False)
```
pbrubeck:
```suggestion
        use_coloring_default = (backend == "petscasm")
        use_coloring = opts.getBool("use_coloring", default=use_coloring_default)
```

firedrake/preconditioners/asm.py
```diff
@@ -233,7 +240,10 @@ def get_patches(self, V):
             raise ValueError(f"{self.prefix}include_type must be either 'star' or 'entity', not {include_type}")
         include_star = include_type == "star"
 
-        use_coloring = opts.getBool("use_coloring", default=False)
+        if backend == "petscasm":
+            use_coloring = opts.getBool("use_coloring", default=True)
+        else:
+            use_coloring = opts.getBool("use_coloring", default=False)
```
pbrubeck:
```suggestion
        use_coloring_default = (backend == "petscasm")
        use_coloring = opts.getBool("use_coloring", default=use_coloring_default)
```

pbrubeck:
We should store `self._backend` in the base class instead of grabbing the option in multiple places.

firedrake/preconditioners/asm.py
```diff
@@ -50,6 +50,7 @@ def initialize(self, pc):
 
         opts = PETSc.Options(self.prefix)
         backend = opts.getString("backend", default="petscasm").lower()
+        self._backend = backend
```
pbrubeck: These three lines must be moved right before the call to `self.get_patches(V)`

firedrake/preconditioners/asm.py
pbrubeck, line 34:
```suggestion
        self.prefix = (pc.getOptionsPrefix() or "") + self._prefix
        opts = PETSc.Options(self.prefix)
        backend = opts.getString("backend", default="petscasm").lower()
        self._backend = backend        
```

firedrake/preconditioners/asm.py
pbrubeck:
```suggestion
```

PR #5213, opened 2026-07-02
Adaptive NonlinearVariationalSolver support via -snes_adapt_sequence

pbrubeck:
# Description

Codex-assisted. 

Implements UFL-level refinement of a variational problem and PETSc-level refinement of the SNES solver. In this way, users will not require to write the adaptivity loop themselves.

```
from firedrake import *

coarse_mesh = UnitSquareMesh(10, 10)
hierarchy = AdaptiveMeshHierarchy(coarse_mesh)

V = FunctionSpace(coarse_mesh, "CG", 1)
u = Function(V)
v = TestFunction(v_space)
F = inner(grad(u), grad(v)) * dx - f * v * dx

def marking_callback(solver_context, solution):
    # Select cells based on some criteria (e.g., Dorfler marking)
    marker = Function(DG0)
    ...
    return marker

params = {
    "snes_adapt_sequence": 5,  # Number of adaptation iterations
}

u_adapt = solve(F == 0, u, solver_parameters=params, marking_callback=marking_callback)     
```

Depends on https://gitlab.com/petsc/petsc/-/merge_requests/9407


firedrake/variational_solver.py
```diff
@@ -388,15 +422,24 @@ def solve(self, bounds=None):
...
+                             self._transfer_operators):
+                stack.enter_context(ctx)
+            self._in_snes_solve = True
+            try:
                 self.snes.solve(None, work)
-            work.copy(u)
+            finally:
+                self._in_snes_solve = False
+
+        problem = self._problem
```
```suggestion
        problem = self._ctx._problem
```

firedrake/variational_solver.py
```diff
@@ -328,6 +339,28 @@ def update_diffusivity(current_solution):
...
+        """
+        if not callable(callback):
+            raise TypeError(f"marking callback must be callable, not a {type(callback).__name__}")
+        self._marking_callback = callback
+        self.parameters.setdefault("adaptor_criterion", "refine")
+        self._set_adaptation_context()
+
+    def get_adapted_solution(self):
+        r"""Return the current solution, including after PETSc adapts the DM."""
+        return self._problem.u
```
```suggestion
        return self._ctx._problem.u
```

firedrake/variational_solver.py
```diff
@@ -328,6 +339,28 @@ def update_diffusivity(current_solution):
         self._transfer_operators = ()
         self._setup = False
 
+    def _set_adaptation_context(self):
+        self._ctx._adapt_solver = self
```
```suggestion
```

firedrake/dmhooks.py
```diff
@@ -446,18 +446,151 @@ def coarsen(dm, comm):
...
+    refined_ctx._adapt_solver = solver
+    refined_ctx._adapt_marking_callback = callback
+    _refine_nullspaces(ctx, refined_ctx, symbolic_refine, coefficient_mapping)
+
+    parent = get_parent(dm)
+    coarsener = get_ctx_coarsener(dm)
+    _attach_refined_context(parent, refined_ctx, coefficient_mapping, coarsener)
+
+    solver._ctx = refined_ctx
+    solver._problem = refined_problem
```
pbrubeck: Let's make `self._ctx` be the only mutating attribute and keep `self._problem` on the original mesh.
```suggestion
```

firedrake/dmhooks.py
```diff
@@ -446,18 +446,151 @@ def coarsen(dm, comm):
...
 @PETSc.Log.EventDecorator()
 def refine(dm, comm):
     """Callback to refine a DM.
 
     :arg DM: The DM to refine.
     :arg comm: The communicator for the new DM (ignored)
     """
+    ctx = get_appctx(dm)
+    if getattr(ctx, "_adapt_marking_callback", None) is not None:
+        return _adaptively_refine(dm, comm)
```
pbrubeck: This is an antipattern

pbrubeck: Add a flag or initialize this to None. Might be better to stash the reult in `ctx._fine`

tests/firedrake/multigrid/test_snes_adapt.py
```diff
@@ -0,0 +1,63 @@
...
+
+    geo = SplineGeometry()
+    geo.AddRectangle((0, 0), (1, 1), bc="boundary")
+    mesh = Mesh(geo.GenerateMesh(maxh=0.5))
+    V = FunctionSpace(mesh, "CG", 1)
+    old_dim = V.dim()
+    u = Function(V)
+    v = TestFunction(V)
+    problem = NonlinearVariationalProblem((u - 1.0)*v*dx, u)
+    solver = NonlinearVariationalSolver(problem, solver_parameters={"snes_type": "ksponly"})
```
pbrubeck: We need a separate test for `-snes_adapt`, ideally re-implementing  
`demos/adaptive_multigrid/adaptive_multigrid.py.rst`

firedrake/variational_solver.py
```diff
@@ -388,15 +422,24 @@ def solve(self, bounds=None):
         work = self._work
         with problem.u_restrict.dat.vec as u:
             u.copy(work)
```
pbrubeck, line 448: Seems like the `_in_snes_solve` flag was required to deal with the work buffer vs actual `u.dat.vec` distinction. We don't want that flag. 

Perhaps this can be resolved if we were solving for the actual `u.dat.vec`. If so we should keep a copy of the inital `u.dat.vec` in the work buffer, so that we can reset `u.dat.vec` to the initial guess on exit if the solver fails.

firedrake/dmhooks.py
```diff
@@ -446,18 +446,151 @@ def coarsen(dm, comm):
...
+    current_solution = _copy_snes_solution_to_context(solver, ctx)
+    mesh = current_solution.function_space().mesh()
+    markers = callback(solver, current_solution)
+    _validate_refinement_markers(markers, mesh)
+
+    hierarchy, level = get_level(mesh)
+    if hierarchy is None:
+        hierarchy = AdaptiveMeshHierarchy(mesh)
+        level = 0
+    if not hasattr(hierarchy, "add_mesh"):
```
pbrubeck:
```suggestion
    if not isinstance(hierarchy, AdaptiveMeshHierarchy):
```

firedrake/dmhooks.py
```diff
@@ -446,18 +446,151 @@ def coarsen(dm, comm):
...
+        raise RuntimeError("Adaptive SNES refinement requires an AdaptiveMeshHierarchy")
+    if level != len(hierarchy) - 1:
```
pbrubeck: We should not error, but instead we should simply return the next level in the hierarchy because it is already there.

firedrake/dmhooks.py
```diff
@@ -446,18 +446,151 @@ def coarsen(dm, comm):
...
+        add_hook(parent, setup=partial(push_ctx_coarsener, dm, coarsener),
+                 teardown=partial(pop_ctx_coarsener, dm, coarsener),
+                 call_setup=True)
+        add_hook(parent, setup=partial(push_appctx, dm, ctx), teardown=partial(pop_appctx, dm, ctx),
+                 call_setup=True)
+
+
+def _adaptively_refine(dm, comm):
+    from firedrake.mg.adaptive_hierarchy import AdaptiveMeshHierarchy
+    from firedrake.mg.ufl_utils import refine as symbolic_refine
```
pbrubeck:
```suggestion
    from firedrake.mg.ufl_utils import refine
```

firedrake/dmhooks.py
```diff
@@ -446,18 +446,151 @@ def coarsen(dm, comm):
...
+        raise RuntimeError("Adaptive SNES refinement requires an AdaptiveMeshHierarchy")
+    if level != len(hierarchy) - 1:
+        raise RuntimeError("Adaptive SNES refinement must refine the finest mesh in the hierarchy")
+
+    refined_mesh = mesh.refine_marked_elements(markers)
+    hierarchy.add_mesh(refined_mesh)
+
+    coefficient_mapping = {}
+    refined_problem = symbolic_refine(ctx._problem, symbolic_refine, coefficient_mapping=coefficient_mapping)
+    refined_appctx = _refine_appctx(ctx, symbolic_refine, coefficient_mapping)
```
pbrubeck:
```suggestion
    refined_appctx = _refine_appctx(ctx, coefficient_mapping)
```

firedrake/dmhooks.py
```diff
@@ -446,18 +446,151 @@ def coarsen(dm, comm):
...
+    if not hasattr(hierarchy, "add_mesh"):
+        raise RuntimeError("Adaptive SNES refinement requires an AdaptiveMeshHierarchy")
+    if level != len(hierarchy) - 1:
+        raise RuntimeError("Adaptive SNES refinement must refine the finest mesh in the hierarchy")
+
+    refined_mesh = mesh.refine_marked_elements(markers)
+    hierarchy.add_mesh(refined_mesh)
+
+    coefficient_mapping = {}
+    refined_problem = symbolic_refine(ctx._problem, symbolic_refine, coefficient_mapping=coefficient_mapping)
```
pbrubeck:
```suggestion
    refined_problem = refine(ctx._problem, refine, coefficient_mapping=coefficient_mapping)
```

firedrake/dmhooks.py
pbrubeck:
```suggestion
    refined_appctx = _refine_appctx(ctx, coefficient_mapping)
```

firedrake/dmhooks.py
```diff
@@ -446,18 +446,151 @@ def coarsen(dm, comm):
...
+            new_appctx[key] = refiner(value, refiner, coefficient_mapping=coefficient_mapping)
+        except CoarseningError:
+            new_appctx[key] = value
+    return new_appctx
+
+
+def _refine_nullspaces(ctx, refined_ctx, refiner, coefficient_mapping):
+    from firedrake.mg.ufl_utils import CoarseningError
+
+    ises = refined_ctx._problem.J.arguments()[0].function_space()._ises
```
pbrubeck:
```suggestion
    ises = refined_ctx._x.function_space()._ises
```

firedrake/dmhooks.py
```diff
@@ -446,18 +446,151 @@ def coarsen(dm, comm):
...
+            f"marking callback must return a Function or Cofunction, not a {type(markers).__name__}"
+        )
+    M = markers.function_space()
+    if M.mesh() is not mesh:
+        raise ValueError("marking callback must return markers on the current solution mesh")
+    if M.finat_element.space_dimension() != 1:
+        raise ValueError("marking callback must return a DG0 Function or Cofunction")
+
+
+def _refine_appctx(ctx, refiner, coefficient_mapping):
```
pbrubeck: We should unify coarsening/refinement of appctx, nullspace, and _SNESContext

pbrubeck: Maybe just reuse the methods introduced here to do coarsening, renaming `refiner -> reconstruct`

firedrake/dmhooks.py
```diff
@@ -446,18 +446,151 @@ def coarsen(dm, comm):
...
+    coefficient_mapping = {}
+    refined_problem = symbolic_refine(ctx._problem, symbolic_refine, coefficient_mapping=coefficient_mapping)
+    refined_appctx = _refine_appctx(ctx, symbolic_refine, coefficient_mapping)
+    refined_ctx = ctx.reconstruct(problem=refined_problem, appctx=refined_appctx)
+    refined_ctx._coefficient_mapping = coefficient_mapping
+    refined_ctx._coarse = ctx
+    ctx._fine = refined_ctx
+    refined_ctx._adapt_solver = solver
+    refined_ctx._adapt_marking_callback = callback
+    _refine_nullspaces(ctx, refined_ctx, symbolic_refine, coefficient_mapping)
```
pbrubeck:
```suggestion
    _refine_nullspaces(ctx, refined_ctx, refine, coefficient_mapping)
```

firedrake/dmhooks.py
```diff
@@ -446,18 +446,97 @@ def coarsen(dm, comm):
...
+    if hierarchy is None:
+        hierarchy = AdaptiveMeshHierarchy(mesh)
+        level = 0
+    if not isinstance(hierarchy, AdaptiveMeshHierarchy):
+        raise RuntimeError("Adaptive SNES refinement requires an AdaptiveMeshHierarchy")
+
+    if level == len(hierarchy) - 1:
+        markers = callback(solver, current_solution)
+        _validate_refinement_markers(markers, mesh)
+        hierarchy.add_mesh(mesh.refine_marked_elements(markers))
```
pbrubeck: We should add support for non-netgen meshes. This needs to go through the DMPlex / DMLabel codepath

tests/firedrake/multigrid/test_snes_adapt.py
```diff
@@ -0,0 +1,148 @@
...
+
+    mesh = UnitSquareMesh(1, 1)
+    V = FunctionSpace(mesh, "CG", 1)
+    u = Function(V)
+    v = TestFunction(V)
+    problem = NonlinearVariationalProblem((u - 1.0)*v*dx, u)
+    solver = NonlinearVariationalSolver(problem, marking_callback=mark_cells)
+
+    assert solver.parameters["adaptor_criterion"] == "refine"
+    assert solver._ctx._adapt_marking_callback is mark_cells
```
pbrubeck:
```suggestion
    assert solver._ctx._marking_callback is mark_cells
```

firedrake/variational_solver.py
```diff
@@ -353,11 +381,12 @@ def solve(self, bounds=None):
            ``vinewtonssls`` or ``vinewtonrsls``.
         """
         # Make sure the DM has this solver's callback functions
+        self._ctx._adapt_marking_callback = self._marking_callback
```
```suggestion
```

firedrake/variational_solver.py
```diff
@@ -303,6 +309,10 @@ def update_diffusivity(current_solution):
         self._ctx = ctx
         self._work = problem.u_restrict.dof_dset.layout_vec.duplicate()
         self.snes.setDM(problem.dm)
+        self._marking_callback = None
+        if marking_callback is not None:
+            self.set_marking_callback(marking_callback)
+        ctx._adapt_marking_callback = self._marking_callback
```
```suggestion
```

firedrake/variational_solver.py
```diff
@@ -303,6 +309,10 @@ def update_diffusivity(current_solution):
         self._ctx = ctx
         self._work = problem.u_restrict.dof_dset.layout_vec.duplicate()
         self.snes.setDM(problem.dm)
+        self._marking_callback = None
```
```suggestion
```

firedrake/variational_solver.py
```diff
@@ -303,6 +310,10 @@ def update_diffusivity(current_solution):
...
```
```suggestion
```

firedrake/variational_solver.py
```diff
@@ -303,6 +310,10 @@ def update_diffusivity(current_solution):
...
+        if marking_callback is not None:
+            self.set_marking_callback(marking_callback)
+        ctx._adapt_marking_callback = self._marking_callback
```
```suggestion
```

firedrake/variational_solver.py
```diff
@@ -328,6 +339,23 @@ def update_diffusivity(current_solution):
...
+        r"""Set the callback used by PETSc-driven adaptive refinement.
+
+        The callback is called as ``callback(solver, u)`` when PETSc asks the
+        solution DM to refine. It must return a DG0 :class:`.Function` or
+        :class:`.Cofunction` on the current solution mesh, with positive values
+        on cells to refine.
+        """
+        if not callable(callback):
+            raise TypeError(f"marking callback must be callable, not a {type(callback).__name__}")
+        self.parameters.setdefault("adaptor_criterion", "refine")
```
pbrubeck: This default should live with the other defaults we have (i.e. -pc_type lu)

firedrake/mg/ufl_utils.py
```diff
@@ -14,56 +14,95 @@
 from . import utils
 
 
-__all__ = ["coarsen"]
+__all__ = ["coarsen", "refine", "reconstruct"]
```
```suggestion
__all__ = ["coarsen", "refine"]
```

firedrake/mg/ufl_utils.py
```diff
@@ -14,56 +14,95 @@
...
+def set_cache(direction, old, new):
+    """Remember `new` as the `direction` counterpart of `old`."""
+    if direction is coarsen:
+        old._coarse = new
+    elif direction is refine:
+        old._fine = new
+    return new
+
+
 class CoarsenIntegrand(MultiFunction):
```
```suggestion
class ReconstructIntegrand(MultiFunction):
```

pbrubeck: Turn this MultiFunction into DAGTraverser

firedrake/dmhooks.py
```diff
@@ -446,27 +446,95 @@ def coarsen(dm, comm):
...
 
 
+def _adaptively_refine(dm, comm):
+    from firedrake.mg.adaptive_hierarchy import AdaptiveMeshHierarchy
+    from firedrake.mg.ufl_utils import refine
+    from firedrake.mg.utils import get_level
+
+    # DMAdaptorAdapt() unconditionally destroys its input DM, but
+    # AdaptiveMeshHierarchy keeps this level's DM alive as a coarse grid.
+    dm.incRef()
```
pbrubeck: Should we only keep alive the coarsest level?

firedrake/mg/ufl_utils.py
```diff
@@ -14,56 +14,95 @@
...
     """Exception raised when coarsening symbolic information fails."""
     pass
 
 
+# `coarsen`/`refine` results are memoized directly on the reconstructed
+# object as `_coarse`/`_fine` attributes, so that walking a mesh hierarchy
+# by hand (`obj._coarse`, `obj._fine`) keeps working. These two helpers are
+# the only place that needs to know the attribute names; a custom `direction`
+# dispatcher (i.e. anything other than the `coarsen`/`refine` singledispatch
+# functions themselves) is not memoized, matching the pre-existing behaviour.
```
```suggestion
```

firedrake/mg/ufl_utils.py
```diff
@@ -14,56 +14,95 @@
 from . import utils
 
 
-__all__ = ["coarsen"]
+__all__ = ["coarsen", "refine", "reconstruct"]
 
 
 class CoarseningError(Exception):
```
```suggestion
class ReconstructionError(Exception):
```

firedrake/mg/ufl_utils.py
```diff
@@ -149,18 +197,48 @@ def coarsen_equation_bc(ebc, self, coefficient_mapping=None):
 
 @coarsen.register(firedrake.functionspaceimpl.WithGeometryBase)
 def coarsen_function_space(V, self, coefficient_mapping=None):
```
pbrubeck: I think that we no longer need two dispatchers here, fuse coarsen and refine for FunctionSpace

firedrake/mg/ufl_utils.py
```diff
@@ -116,24 +162,24 @@ def coarsen_form(form, self, coefficient_mapping=None):
...
                         for ci, wi in zip(form.components(), form.weights())])
 
 
-@coarsen.register(firedrake.DirichletBC)
-def coarsen_bc(bc, self, coefficient_mapping=None):
+@_reconstruct.register(firedrake.DirichletBC)
+def reconstruct_bc(bc, self, coefficient_mapping=None):
     V = self(bc.function_space(), self, coefficient_mapping=coefficient_mapping)
-    val = self(bc.function_arg, self, coefficient_mapping=coefficient_mapping)
+    val = self(bc._original_arg, self, coefficient_mapping=coefficient_mapping)
```
```suggestion
    val = self(bc._function_arg, self, coefficient_mapping=coefficient_mapping)
```

firedrake/mg/ufl_utils.py
```diff
@@ -182,10 +245,37 @@ def coarsen_function(expr, self, coefficient_mapping=None):
...
+        if name is not None:
+            try:
+                name, prev_level = name.split("_level_")
+            except ValueError:
+                prev_level = 0
+            level = int(prev_level) + 1
+            name = f"{name}_level_{level}"
+
+        new = firedrake.Function(Vnew, name=name)
+        new.interpolate(expr)
```
pbrubeck: We should prolong instead, and that would be done at a later point
```suggestion
```

firedrake/mg/ufl_utils.py
```diff
@@ -194,11 +284,18 @@ def inject_on_restrict(fine, restriction, rscale, injection, coarse):
...
             if cmapping is None:
                 return
-            for c in cmapping:
+            for c, mapped in cmapping.items():
+                _, clevel = utils.get_level(c.function_space().mesh())
+                _, mlevel = utils.get_level(mapped.function_space().mesh())
                 if is_dual(c):
-                    manager.restrict(c, cmapping[c])
+                    if clevel > mlevel:
+                        manager.restrict(c, mapped)
```
```suggestion
                        manager.restrict(c, mapped)
                    else:
                        mapped.interpolate(c)
```

firedrake/mg/ufl_utils.py
```diff
@@ -194,11 +284,18 @@ def inject_on_restrict(fine, restriction, rscale, injection, coarse):
...
+                elif clevel > mlevel:
+                    manager.inject(c, mapped)
+                elif clevel < mlevel:
+                    manager.prolong(c, mapped)
                 else:
-                    manager.inject(c, cmapping[c])
+                    mapped.assign(c)
```
```suggestion
                elif clevel < mlevel:
                    manager.prolong(c, mapped)
                else:
                    manager.inject(c, mapped)
```

firedrake/dmhooks.py
pbrubeck: No — the incRef is needed at every level, not just the coarsest. `DMAdaptorAdapt` destroys its input DM on each adapt iteration, and at every level that input DM is owned by a live Firedrake `FunctionSpace`: the original problem's space at level 0, and the cached reconstructed contexts' spaces at subsequent levels, all of which stay reachable because each mesh remains in the `AdaptiveMeshHierarchy` as an MG coarse grid. The incRef here just pre-compensates that destroy (net-zero refcount, no leak). Restricting it to the coarsest level would leave the intermediate levels' DMs vulnerable to premature destruction.

connorjward:
```py
problem = NonlinearVariationalProblem(F, u)
solver = NonlinearVariationalSolver(problem, solver_parameters=params, 
                                    marking_callback=marking_callback)
solver.solve()          
u_adapt = solver.get_adapted_solution()
```
Is this the right API? We are passing an irrelevant solution `u` to the NLVP and we have to manually fetch the new solution.

pbrubeck:
> Is this the right API? We are passing an irrelevant solution `u` to the NLVP and we have to manually fetch the new solution.

`u` on input has the initial guess, so it is not irrelevant. We need to manually fetch the adaptive solution because it is created on the fly. I can also make solve() return the adapted one.

connorjward:
> > Is this the right API? We are passing an irrelevant solution `u` to the NLVP and we have to manually fetch the new solution.
> 
> `u` on input has the initial guess, so it is not irrelevant. We need to manually fetch the adaptive solution because it is created on the fly. I can also make solve() return the adapted one.

Yes I think returning the new `u` is much nicer.

```py
solver = NonlinearVariationalSolver(problem, solver_parameters=params, 
                                     marking_callback=marking_callback)
```

I think we want to use the new appctx machinery here, then we don't have to change the API at all:
```py
params = {
  "marking_callback": marking_callback,  # a python function
}
solver = NonlinearVariationalSolver(problem, solver_parameters=params)
```
@JHopeCollins and I have discussed this. Basically the `OptionsManager` needs to check the parameter set for Python objects and then put them in the appctx.

It probably isn't necessary to change the mixin status of the solver as an `OptionsManager` for this PR.

I know it's a bit of an ask, but this is a big feature. We want to get it right.

JHopeCollins:
> I think we want to use the new appctx machinery here, then we don't have to change the API at all:

Possibly, but we haven't put the new appctx into the rest of Firedrake yet. I'd rather do it all at once than have a piecemeal migration.

connorjward:
> > I think we want to use the new appctx machinery here, then we don't have to change the API at all:
> 
> Possibly, but we haven't put the new appctx into the rest of Firedrake yet. I'd rather do it all at once than have a piecemeal migration.

I think it's as simple as adding the functionality that we want into `OptionsManager.__init__`. No actual Firedrake changes needed.

It does mean that we have two appctx objects, the prior dictionary and the new one that lives on the `OptionsManager`, but I think that that inconsistency is worth having in order to not be unnecessarily complicating our API.

connorjward (changes requested):
Requesting changes to make sure [our discussion](https://github.com/firedrakeproject/firedrake/pull/5213#issuecomment-4934289350) isn't forgotten about.

JHopeCollins:
> It does mean that we have two appctx objects, the prior dictionary and the new one that lives on the OptionsManager, but I think that that inconsistency is worth having in order to not be unnecessarily complicating our API.

This time I would actually disagree. I would prefer consistency, so we use the current version of `appctx` currently, and move everything over at the same time rather than mixing interfaces.

firedrake/variational_solver.py
```diff
@@ -223,9 +227,15 @@ def __init__(self, problem, *, solver_parameters=None,
                before residual assembly.
         :kwarg post_function_callback: As above, but called immediately
                after residual assembly.
-        :kwarg pre_apply_bcs: If `True`, the bcs are applied before the solve.
+        :kwarg pre_apply_bcs: If True, the bcs are applied before the solve.
                Otherwise, the problem is linearised around the initial guess
                before imposing bcs, and the bcs are appended to the nonlinear system.
+        :kwarg marking_callback: An optional callable of the form
+               ``callback(ctx, u)`` for PETSc-driven adaptive refinement.
+               The callback receives the :class:`~.solving_utils._SNESContext`
```
```suggestion
               The callback receives the `_SNESContext`
```

firedrake/variational_solver.py
```diff
@@ -328,6 +341,23 @@ def update_diffusivity(current_solution):
         self._transfer_operators = ()
         self._setup = False
 
+    def set_marking_callback(self, callback):
+        r"""Set the callback used by PETSc-driven adaptive refinement.
+
+        The callback is called as ``callback(ctx, u)`` when PETSc asks the
+        solution DM to refine, where ``ctx`` is the current
+        :class:`~.solving_utils._SNESContext`. It must return a DG0
```
```suggestion
        `_SNESContext`. It must return a DG0
```

firedrake/variational_solver.py
```diff
@@ -357,7 +387,7 @@ def solve(self, bounds=None):
         self._ctx.set_jacobian(self.snes)
 
         # Make sure appcontext is attached to every DM from every coefficient and DirichletBC before we solve.
-        problem = self._problem
+        problem = self._ctx._problem
```
dham: put in a property so that `_problem` is still there for users.

dham (dismissed):
and potentially top level API change to put the marker function in the dictionary.

firedrake/variational_solver.py
pbrubeck: done

firedrake/variational_solver.py
```diff
@@ -481,7 +528,9 @@ class LinearVariationalSolver(NonlinearVariationalSolver):
     See also :class:`NonlinearVariationalSolver` for nonlinear problems.
     """
 
-    DEFAULT_SNES_PARAMETERS = {"snes_type": "ksponly"}
+    DEFAULT_SNES_PARAMETERS = MappingProxyType({
+        "snes_type": "ksponly",
+    })
```
```suggestion
    DEFAULT_SNES_PARAMETERS = {"snes_type": "ksponly"}
```

firedrake/petsc.py
```diff
@@ -139,12 +139,14 @@ def garbage_view(obj: Any) -> None:
     "ksp_type": "preonly",
     "ksp_rtol": 1e-7,
     "pc_type": "lu",
-    "pc_factor": _DEFAULT_DIRECT_SOLVER_PARAMETERS
+    "pc_factor": _DEFAULT_DIRECT_SOLVER_PARAMETERS,
+    "adaptor_criterion": "refine",
```
pbrubeck:
```suggestion
    "pc_factor": _DEFAULT_DIRECT_SOLVER_PARAMETERS
```

firedrake/petsc.py
```diff
@@ -139,12 +139,14 @@ def garbage_view(obj: Any) -> None:
...
 })
 
 _DEFAULT_SNES_PARAMETERS = {
     "snes_type": "newtonls",
     "snes_linesearch_type": "basic",
+    "adaptor_criterion": "refine",
```
pbrubeck:
```suggestion
```

firedrake/mg/ufl_utils.py
```diff
@@ -14,98 +13,145 @@
...
-    """'Coarsen' a :class:`ufl.Expr` by replacing coefficients,
-    arguments and domain data with coarse mesh equivalents."""
 
-    def __init__(self, coarsen, coefficient_mapping=None):
+def set_cache(direction, old, new):
+    if direction is coarsen:
+        old._coarse = new
+    elif direction is refine:
+        old._fine = new
+    return new
```
connorjward: Why not just have `old._reconstruction_prior_obj = new`? Not sure why the cache cares about `_fine` vs `_coarse`.

firedrake/mg/ufl_utils.py
```diff
@@ -14,98 +13,145 @@
...
+        return getattr(old, "_coarse", None)
+    elif direction is refine:
+        return getattr(old, "_fine", None)
+    return None
 
-    """'Coarsen' a :class:`ufl.Expr` by replacing coefficients,
-    arguments and domain data with coarse mesh equivalents."""
 
-    def __init__(self, coarsen, coefficient_mapping=None):
+def set_cache(direction, old, new):
```
connorjward: Super unclear name

firedrake/mg/ufl_utils.py
```diff
@@ -14,98 +13,145 @@
...
+
+
+@_reconstruct.register(ufl.BaseForm)
+@_reconstruct.register(ufl.classes.Expr)
+def reconstruct_expr(expr, self, coefficient_mapping=None):
     if expr is None:
         return None
-    mapper = CoarsenIntegrand(self, coefficient_mapping)
-    return map_expr_dag(mapper, expr)
+    mapper = ReconstructIntegrand(self, coefficient_mapping)
```
connorjward, line 157: This seems wrong. The `__init__` signature of `ReconstructIntegrand` is
```py
    def __init__(self, dispatch, coefficient_mapping=None):
```

firedrake/mg/ufl_utils.py
```diff
@@ -182,10 +245,36 @@ def coarsen_function(expr, self, coefficient_mapping=None):
...
+            name = f"{name}_level_{level}"
+
+        new = firedrake.Function(Vnew, name=name)
+        coefficient_mapping[expr] = new
+    return new
+
+
+@_reconstruct.register(firedrake.NonlinearVariationalProblem)
+def reconstruct_nlvp(problem, self, coefficient_mapping=None):
+    # Have we done this already?
```
connorjward:
```suggestion
```

firedrake/dmhooks.py
```diff
@@ -448,27 +448,94 @@ def coarsen(dm, comm):
     return cdm
 
 
+def _adaptively_refine(dm, comm):
```
connorjward: docstring

firedrake/dmhooks.py
```diff
@@ -448,27 +448,94 @@ def coarsen(dm, comm):
...
+            raise RuntimeError("Adaptive SNES refinement requires setting a marking_callback")
+        markers = ctx._marking_callback(ctx, current_solution)
+        if not isinstance(markers, (firedrake.Function, firedrake.Cofunction)):
+            raise TypeError(
+                f"marking callback must return a Function or Cofunction, not a {type(markers).__name__}"
+            )
+        M = markers.function_space()
+        if M.mesh() is not mesh:
+            raise ValueError("marking callback must return markers on the current solution mesh")
+        if M.finat_element.space_dimension() != 1:
```
connorjward: Is that the right check?

firedrake/dmhooks.py
```diff
@@ -448,27 +448,94 @@ def coarsen(dm, comm):
...
+                 teardown=partial(pop_ctx_coarsener, refined_dm, coarsener),
+                 call_setup=True)
+        add_hook(parent, setup=partial(push_appctx, refined_dm, refined_ctx),
+                 teardown=partial(pop_appctx, refined_dm, refined_ctx),
+                 call_setup=True)
+    return refined_ctx._problem.dm
+
+
 @PETSc.Log.EventDecorator()
 def refine(dm, comm):
```
connorjward, line 531: Can you split this up into `_refine_adaptive` and `_refine_uniform`
```py
def refine():
  if adaptive:
    _refine_adaptive()
  else:
    _refine_uniform()
```

firedrake/dmhooks.py
connorjward: comm isn't used

firedrake/variational_solver.py
```diff
@@ -271,10 +278,15 @@ def update_diffusivity(current_solution):
             if solver_parameters["pmat_type"] != problem.Jp.mat_type:
                 raise ValueError("Cannot change the mat_type of an already assembled matrix.")
 
+        snes_defaults = self.DEFAULT_SNES_PARAMETERS
+        if marking_callback is not None:
+            snes_defaults = dict(snes_defaults)
+            snes_defaults.setdefault("adaptor_criterion", "refine")
```
connorjward, line 294: If the ultimate goal is to be able to drive this through PETSc options should we be setting this ourselves?

pbrubeck: Petsc has a different default (`none`), which will not do any refinement, but also throws an error for us because it expects PetscDS fields on the DM which we don't provide (I completely missed this case in https://gitlab.com/petsc/petsc/-/merge_requests/9407). 

There are other criteria (`label` and `metric`) but our users will be creating a DG marker and not a DMLabel, and the metric-based approach is a different can of worms. In the future, we can add support for every option. But passing the `marking_callback` means that we eventually call `refine`/`adpatively_refine` on the DM (to call the callback), so the callback and the criterion option are deeply intertwined/aliased.

I think having `refine` as default is to make life easier for people wanting adaptivity in this early stage, as they just need to provide a single petsc option (`-snes_adapt_sequence`), and they will not get an unexpected non-adaptive solve if they miss the `adaptor_criterion` option

firedrake/dmhooks.py
pbrubeck: There are many DG0 variants, so this covers all of them. However it also covers the Real space, which still makes sense if you want uniform refinement. However the logic below might break with Real. I can add support for Real as well.

firedrake/mg/ufl_utils.py
pbrubeck: dispatch is self (a different self). The constructor does not take self as an argument.
```suggestion
    mapper = ReconstructIntegrand(self, coefficient_mapping=coefficient_mapping)
```

firedrake/mg/ufl_utils.py
pbrubeck: You need different caches because you will be going in both directions (and I don't want to break tests that depend on `_fine` and `_coarse)`

firedrake/mg/ufl_utils.py
pbrubeck: `set_reconstruct_cache`?

firedrake/dmhooks.py
connorjward: I just have no idea what `space_dimension` is. It very much doesn't seem like it has anything to do with identifying DG0.

pbrubeck:
```suggestion
        num_dofs_per_cell = M.finat_element.space_dimension()
        if num_dofs_per_cell != 1:
```

pbrubeck: Mathematically, 1 dof per cell implies DG0.

firedrake/dmhooks.py
pbrubeck: The original refine is just reading the cache / or creating the FunctionSpace on an exisiting mesh coming from a pre-fabricated MeshHierarchy (not necessarily uniform).

Suppose I solve a problem adaptively to obtain a hierarchy, one can solve a different problem defined on the finest level of that hierarchy. So the distinction is not `_refine_adaptive` vs `_refine_uniform`. But more like `_refine_add_new_level` vs `_refine_on_exisiting_level`

firedrake/mg/ufl_utils.py
connorjward: Oh yeah. Duh.

firedrake/dmhooks.py
connorjward: That is more readable thanks. I appreciate now that what you did before was totally fine.

firedrake/dmhooks.py
connorjward: I am not bothered by the names. This is just a code style/readability point.

firedrake/mg/ufl_utils.py
connorjward: OK. Can this at least not be called a cache? It does not really meet the definition of one I would say. And any time I see something that says cache alarm bells start ringing. `attach_prior()`?

connorjward: One should always be able to disable a cache and have the behaviour be unchanged, just slower

firedrake/variational_solver.py
connorjward: What if we caught if you were passing `snes_adapt_sequence` and inserted it at that point?

There are some weird permutations of options here. A user can pass `marking_callback`, we will insert `adaptor_criterior`, and no adapting will happen.

From my perspective, setting `snes_adapt_sequence` is the primary 'driving' option, and everything should follow from that. It is a funny API for users to set things in two different places/ways in order to get things to work.

Also, and I know this is not necessarily realistic, if you have something like fieldsplit SNES with snes_adapt then you might want to have different options for `marking_callback` for each SNES. There isn't necessarily a 1to1 connection between a variational solver and a SNES.

pbrubeck: This is why the marking callback takes in the solver context, the same callback can do different things depending on who is calling it.

This is still quite an early stage for this feature. I have yet to write the default for marking callback, which is goal-adaptivity/dual-weighted residual estimation. I don't expect the user to write their own callback.

pbrubeck: Also, I think your concerns about composability are valid. But at the moment it is quite unclear how this would be possible, given that an adaptive solve changes the DM/the variational problem. An adaptive subsolver for one of the fields might not play well if the other fields are left unadapted, we might easily end up in a multidomain problem (maybe that would be fine). 

As mentioned during the meeting we should just get the most basic functionality first, and then we can add features/make it more composable as the use cases become more clear.

pbrubeck: `-snes_adapt_sequence -adaptor_criterion` are analogous to `-snes_maxit -snes_type` the main difference here is that the defaults are no-ops in the adaptive case, this forces the user to worry about both. However because the user  may be also responsible for the marking, they need to pass a `marking_callback` (the closest analogy would be a custom snes monitor, rarely needed because the defaults are good). So API isn't that funny, it is just slightly more explicit due to no-op defaults.

connorjward: OK. I'm fine with this approach for the moment. I just think it's important to discuss these from the outset.

firedrake/dmhooks.py
pbrubeck:
```suggestion
def _adaptively_refine(dm):
    """
    Return the DM of the `_SNESContext` reconstructed on the adaptively-refined
    mesh using `_SNESContext.marking_callback` to mark the cells to be refined.
    """
```

firedrake/dmhooks.py
```diff
@@ -448,27 +448,99 @@ def coarsen(dm, comm):
...
 
     :arg DM: The DM to refine.
     :arg comm: The communicator for the new DM (ignored)
     """
+    ctx = get_appctx(dm)
+    if ctx is not None and ctx._marking_callback is not None:
+        return _adaptively_refine(dm, comm)
+
+    from firedrake.mg.ufl_utils import get_cache, set_cache
+    from firedrake.mg.ufl_utils import coarsen as symbolic_coarsen, refine as symbolic_refine
```
pbrubeck:
```suggestion
```

firedrake/mg/ufl_utils.py
```diff
@@ -14,98 +13,168 @@
...
+    return new
+
+
+def attach_relative(dispatch, old, new):
+    """Set old._coarse or old._fine to new depending on dispatch."""
+    if dispatch is coarsen:
+        old._coarse = new
+    elif dispatch is refine:
+        old._fine = new
+    return new
```
connorjward:
```suggestion
```
shouldn't mutate and return, especially since `new` is already an argument

firedrake/mg/ufl_utils.py
```diff
@@ -147,45 +216,74 @@ def coarsen_equation_bc(ebc, self, coefficient_mapping=None):
...
+    reverse = coarsen if self is refine else refine
+    V_parent = V
+    while get_relative(reverse, V_parent) is not None:
+        V_parent = get_relative(reverse, V_parent)
+    name = V_parent.name
+    if name is not None:
+        mh, level = utils.get_level(new_mesh)
+        name = f"{name}_level_{level}"
+    V_new = V.reconstruct(mesh=new_mesh, name=name)
+    attach_relative(reverse, V_new, V)
```
connorjward:
```suggestion
    attach_relative(self, V_new, V, reverse=True)
```
WDYT
Seems more readable

firedrake/dmhooks.py
```diff
@@ -448,27 +448,95 @@ def coarsen(dm, comm):
...
-    else:
-        V._fine = V.reconstruct(mesh=hierarchy[level + 1])
-        fdm = V._fine.dm
-    V._fine._coarse = V
-    return fdm
+    from firedrake.mg.ufl_utils import ReconstructionError
+    try:
+        return _refine_from_hierarchy(dm)
+    except ReconstructionError:
+        return _refine_adaptive(dm)
```
connorjward, line 542: This is confusing control flow. Isn't there something we could assertively check instead?

firedrake/dmhooks.py
```diff
@@ -448,27 +448,95 @@ def coarsen(dm, comm):
...
+    Return the DM of the `_SNESContext` reconstructed on the adaptively-refined
+    mesh using `_SNESContext.marking_callback` to mark the cells to be refined.
+    """
+    from firedrake.mg.adaptive_hierarchy import AdaptiveMeshHierarchy
+    from firedrake.mg.ufl_utils import refine
+    from firedrake.mg.utils import get_level
+
+    # DMAdaptorAdapt() unconditionally destroys its input DM, and each
+    # adapted input DM remains a level in the AdaptiveMeshHierarchy.
+    dm.incRef()
```
connorjward, line 463: The comment doesn't really explain the code change

firedrake/dmhooks.py
```diff
@@ -448,27 +448,95 @@ def coarsen(dm, comm):
...
+    current_solution = ctx._x
+    mesh = current_solution.function_space().mesh()
+    hierarchy, level = get_level(mesh)
+    if hierarchy is None:
+        hierarchy = AdaptiveMeshHierarchy(mesh)
+        level = 0
+    if not isinstance(hierarchy, AdaptiveMeshHierarchy):
+        raise RuntimeError("Adaptive SNES refinement requires an AdaptiveMeshHierarchy")
+
+    if level == len(hierarchy) - 1:
```
connorjward: Can you explain why this is the condition for choosing to refine. Do we hit this function when we're not on the current finest level?

firedrake/dmhooks.py
```diff
@@ -448,27 +448,95 @@ def coarsen(dm, comm):
...
+            )
+        M = markers.function_space()
+        if M.mesh() is not mesh:
+            raise ValueError("marking callback must return markers on the current solution mesh")
+        num_dofs_per_cell = M.finat_element.space_dimension()
+        if num_dofs_per_cell != 1:
+            raise ValueError("marking callback must return a DG0 Function or Cofunction")
+
+        hierarchy.add_mesh(mesh.refine_marked_elements(markers))
+
```
connorjward:
```suggestion

# Attach the refined context to all the DMs of the problem
```

connorjward (changes requested):
Otherwise I am happy

firedrake/mg/ufl_utils.py
pbrubeck: We are using `reverse` earlier to get the top level relative. I don't think we gain much from this.

connorjward: To me it is very unintuitive as it stands

firedrake/dmhooks.py
pbrubeck: We shouldn't call this function on intermidiate levels, based on the if-else in `refine()`, so I'll throw an error instead of having this if-statement

firedrake/dmhooks.py
pbrubeck: I added what the code does and what effect that has.

firedrake/mg/ufl_utils.py
```diff
@@ -1,10 +1,9 @@
 import ufl
-from ufl.corealg.map_dag import map_expr_dag
-from ufl.corealg.multifunction import MultiFunction
-from ufl.domain import extract_unique_domain
+from ufl.corealg.dag_traverser import DAGTraverser
+from ufl.domain import extract_unique_domain, as_domain
 from ufl.duals import is_dual
```
```suggestion
```

firedrake/mg/ufl_utils.py
pbrubeck: As you wish

connorjward (dismissed):
I think I'm happy now. Thanks.

pbrubeck:
There are two unrelated failures in CI:
- Gusto `from firedrake.__future__ import interpolate`
- `tests/firedrake/regression/test_bddc.py::test_bddc_aij_simplex[CG-3-True]` which I also seen failing in other jobs.

The same two failures occur in:

https://github.com/firedrakeproject/firedrake/actions/runs/30134894502/job/89616670550

https://github.com/firedrakeproject/firedrake/actions/runs/30105191125/job/89520775815?pr=5276

PR #5214, opened 2026-07-03
Allow in place operators to modify coordinates

dham:
Replacing the coordinates property of a mesh is not allowed for lots of good reasons.

However, the Python implementation of in place operators does assignment. That is to say:
```python
mesh.coordinates += 1.
```
is equivalent to:
```python
mesh.coordinates = mesh.coordinates.__iadd__(1.)
```
This triggers the error even though `mesh.coordinates.__iadd__` returns `mesh.coordinates`. This PR relaxes the error check so that assigning `mesh.coordinates` itself (i.e. a no-op) is allowed.


connorjward (dismissed):

connorjward:
Should this go into release?

tests/firedrake/meshes/test_modify_coordinates.py
leo-collins, line 1: Could this go in `firedrake/regression/test_change_coordinates.py` instead of a new file?

connorjward: Great point

connorjward (approved):

JHopeCollins:
Could this be merged? Looks like the only errors are the netgen ones that are waiting for the next netgen release.

connorjward:
Yep good to go

PR #5226, opened 2026-07-08
Remove explicit annotation on solve() (-> #4638)

angus-g:
We remove the explicit adjoint annotation on the bare solve() call, relying on the path through the NLVS as the single source of annotation. This simplifies and unifies the code paths, and will ideally allow us to remove the specialised solve blocks.

I think test failures here are dealt with by #5188.


JHopeCollins (approved):
Yes, great. This shouldn't have been a thing anyway. The less code to maintain the better.

Issue #2452, opened 2022-05-26
CheckpointFile cannot save VertexOnlyMesh


dham:
MFE:
```python3
from firedrake import *

mesh = UnitIntervalMesh(2)
vm = VertexOnlyMesh(mesh, [[0.55]])

with CheckpointFile("foo.h5", "w") as outfile:
    outfile.save_mesh(vm)
```

Output is:
```console
(firedrake2) dham@IC-D46G33P392 firedrake % python test_checkpoint_vertex_only_mesh.py
Traceback (most recent call last):
  File "/Users/dham/src/firedrake2/src/firedrake/test_checkpoint_vertex_only_mesh.py", line 7, in <module>
    outfile.save_mesh(vm)
  File "PETSc/Log.pyx", line 115, in petsc4py.PETSc.Log.EventDecorator.decorator.wrapped_func
  File "PETSc/Log.pyx", line 116, in petsc4py.PETSc.Log.EventDecorator.decorator.wrapped_func
  File "/Users/dham/src/firedrake2/src/firedrake/firedrake/checkpointing.py", line 600, in save_mesh
    self._save_mesh_topology(tmesh)
  File "PETSc/Log.pyx", line 115, in petsc4py.PETSc.Log.EventDecorator.decorator.wrapped_func
  File "PETSc/Log.pyx", line 116, in petsc4py.PETSc.Log.EventDecorator.decorator.wrapped_func
  File "/Users/dham/src/firedrake2/src/firedrake/firedrake/checkpointing.py", line 635, in _save_mesh_topology
    topology_dm.topologyView(viewer=self.viewer)
AttributeError: 'petsc4py.PETSc.DMSwarm' object has no attribute 'topologyView'
```

ksagiyam:
I think we can implement `DMView()` and `DMLoad()` for `DMSwarm`s reusing existing infrastructure.

The most straightforward implementation would probably put vertices also in ghost cells; why do we remove vertices in the ghost cells? Maybe we had this discussion long time ago, but, if swarm points ever contributed to some interior facet integral, we would need to have swarm points in the ghost cells. Would it currently break something if we had ghost swarm points? @ReubenHill

wence-:
You can't integrate over a mesh and use a function on a vertexonlymesh, and vertexonlymeshes have no connectivity, so all integrals over them decouple. So "interior facet integrals" doesn't make sense.

ksagiyam:
[...]

PR #5228, opened 2026-07-08
DumbCheckpoint renew warning

Ig-dolci:
# Description
Since I do not have the knowledge to review the entire `DumbCheckpoint` docs, I am just renewing the deprecation warning and making some minimal changes to the docs.

Screenshot 2026-07-08 at 11 33 49


docs/source/checkpointing.rst
```diff
@@ -363,252 +363,6 @@ For more details on available checkpointing strategies, refer to the
...
-   chk.load(a)
-
-   # Look up value by explicitly specifying name="B"
-   chk.load(b, name="B")
-
-.. note::
-
-   Since Firedrake does not currently support reading data from a
-   checkpoint file on a different number of processes from that it was
-   written with, whenever a :class:`~.Function` is stored, an
```
connorjward, line 480: It's fun that we never deleted this

connorjward (commented):
I have a bad feeling that we haven't deleted this yet because there are some things `CheckpointFile` can't do.

Ig-dolci:
> I have a bad feeling that we haven't deleted this yet because there are some things `CheckpointFile` can't do.

And Gusto is using `DumbCheckpoint`.

jshipton:
> > I have a bad feeling that we haven't deleted this yet because there are some things `CheckpointFile` can't do.
> 
> And Gusto is using `DumbCheckpoint`.

I'm not sure we are... we still have it hanging around and it is tested but I can't see that we actually *use* it in any examples or case studies... @tommbendall can you think of anywhere we need it?

JHopeCollins:
Part of me thinks we should just go ahead and finally get rid of this. But I also worry that we've created a boy who cried wolf situation and there will be users out there who are using `DumbCheckpoint` and have just got used to ignoring the warning.

Depending on how cautious we want to be, we could first update the deprecation warning to give a specific date for removal. So after the next major release it says something like `This will be removed in the next major release in early 2027, after this you must use CheckpointFile`.

leo-collins:
> I have a bad feeling that we haven't deleted this yet because there are some things CheckpointFile can't do.

I know of at least this: https://github.com/firedrakeproject/firedrake/issues/2452

JHopeCollins:
> > I have a bad feeling that we haven't deleted this yet because there are some things CheckpointFile can't do.
> 
> I know of at least this: #2452

The last comment on that issue is 3 years old now, do we know if it's still the case? If it is, do you have a rough idea how big a fix it would be?

connorjward:
> > > I have a bad feeling that we haven't deleted this yet because there are some things CheckpointFile can't do.
> > 
> > 
> > I know of at least this: #2452
> 
> The last comment on that issue is 3 years old now, do we know if it's still the case? If it is, do you have a rough idea how big a fix it would be?

The trouble is that I don't know the exact strategy to follow. 

Checkpointing is funny because you have to reconstruct all of the Firedrake objects on top of the loaded PETSc ones and pass extra things around (e.g. `permutation_name` and `sfXB`).

With pyop3 I found that the Firedrake interface can be very limiting for this. For instance I had to abandon doing checkpointing for extruded meshes 'properly' because `ExtrudedMeshTopology` only takes the base mesh as the argument. You can't build an `ExtrudedMeshTopology` object using an already extruded mesh and so you can’t checkpoint it properly.

We would have to think about the interfaces of things. It would be nice if a VoM took a DMSwarm as input for example.

tommbendall:
> > > I have a bad feeling that we haven't deleted this yet because there are some things `CheckpointFile` can't do.
> > 
> > 
> > And Gusto is using `DumbCheckpoint`.
> 
> I'm not sure we are... we still have it hanging around and it is tested but I can't see that we actually _use_ it in any examples or case studies... @tommbendall can you think of anywhere we need it?

When we brought in `CheckpointFile` there were still a few things that it couldn't do, but I don't see why we would need to keep `DumbCheckpoint` now. We test that both options work but could remove the `DumbCheckpoint` method

Ig-dolci:
It looks like renewing the deprecation warning and reviewing the docs is the best approach for now. What do you think, @JHopeCollins @connorjward?

JHopeCollins:
> It looks like renewing the deprecation warning and reviewing the docs is the best approach for now. What do you think, @JHopeCollins @connorjward?

Yeah I think so. It slows things down but we are paying for having left it so long already!

firedrake/checkpointing.py
```diff
@@ -92,17 +92,19 @@ class DumbCheckpoint:
...
     """
     def __init__(self, basename, single_file=True,
                  mode=FILE_UPDATE, comm=None):
         import warnings
         with warnings.catch_warnings():
             warnings.simplefilter('always', DeprecationWarning)
-            warnings.warn("DumbCheckpoint class will soon be deprecated; use CheckpointFile class instead.",
-                          DeprecationWarning)
+            warnings.warn("DumbCheckpoint is deprecated and is scheduled for removal in the "
+                          "first major Firedrake release after 1 January 2027; use CheckpointFile instead.",
```
connorjward:
```suggestion
            warnings.warn("DumbCheckpoint is deprecated and will be removed soon; "
                          "use CheckpointFile instead.",
```

docs/source/checkpointing.rst
```diff
@@ -368,9 +368,10 @@ Checkpointing with DumbCheckpoint
 
 .. warning::
 
-   :class:`~.DumbCheckpoint` will be deprecated after 01/01/2023.
-   Instead, users are encouraged to use :class:`~.CheckpointFile`,
-   which is more robust and scalable.
+   :class:`~.DumbCheckpoint` is deprecated and is scheduled for removal in
+   the first major Firedrake release after 1 January 2027.  Users are
+   encouraged to use :class:`~.CheckpointFile`, which is more robust and
+   scalable.
```
connorjward:
```suggestion
   :class:`~.DumbCheckpoint` is deprecated and will be removed soon.
   Users are encouraged to use :class:`~.CheckpointFile`, which is more
   robust and scalable.
```

connorjward (changes requested):
I don't want to commit to a particular removal date in case we fail to hit it.

Ig-dolci:
> I don't want to commit to a particular removal date in case we fail to hit it.

Fair enough. Done.

connorjward (approved):

PR #5230, opened 2026-07-09
tsfc: cancel Jacobian products and reciprocal factors

pbrubeck:
## Description

Companion PRs (required):
- UFL: https://github.com/FEniCS/ufl/pull/496
- FIAT: https://github.com/firedrakeproject/fiat/pull/255

Piola-mapped elements generate expressions of the form `J * inv(J)` that were not simplified, because `JacobianInverse` was expanded into individual matrix entries too early in form preprocessing. This PR enables the new UFL pipeline that preserves the Jacobian quantities as terminals, cancels `IndexSum(J[a,k] * K[k,b] * factors, k) -> factors * delta_ab`, eliminates the resulting Kronecker deltas, and cancels reciprocal determinant powers, before lowering geometry as usual:

- `tsfc/ufl_utils.py`: enable `do_cancel_jacobian_products` in `compute_form_data`, and apply the same simplifications in `preprocess_expression` (interpolation and point evaluation paths).
- `tsfc/spectral.py`: cancel reciprocal factors after `replace_division`.
- `tsfc/driver.py`: cancel reciprocal factors in dual evaluation, where the Piola pullback's `detJ**2` (from UFL) meets the inverse basis transformation's `(1/detJ)**2` (from FInAT); previously this product multiplied every term of the kernel.

## Benchmark

Johnson-Mercier `inner(u, v)*dx + inner(div(u), div(v))*dx`, spectral mode, cold cache:

| | baseline | this PR | change |
|---|---|---|---|
| 2D flops | 27020 | 18728 | -31% |
| 2D compile time | 0.52 s | 0.36 s | -31% |
| 2D kernel size | 47 kB | 29 kB | -38% |
| 3D flops | 763527 | 444574 | -42% |
| 3D compile time | 3.62 s | 1.81 s | -51% |
| 3D kernel size | 323 kB | 143 kB | -56% |

On affine cells the div-div integrand now reaches TSFC as `(1/detJ**2) J : rdiv(sigma)`, with all second-derivative geometry terms folded at the UFL level. Assembled JM matrices agree with the unoptimised pipeline to machine precision on perturbed simplices in 2D and 3D.


.github/actions/install/action.yml
```diff
@@ -160,6 +160,8 @@ runs:
           --extra-index-url https://download.pytorch.org/whl/cpu \
           "./firedrake-repo[${{ inputs.deps }}]"
 
+        pip install -v --no-deps --ignore-installed git+https://github.com/firedrakeproject/fiat.git@pbrubeck/cancel-jacobian
+        pip install -v --no-deps --ignore-installed git+https://github.com/firedrakeproject/ufl.git@pbrubeck/cancel-jacobian
```
```suggestion
```

tsfc/ufl_utils.py
```diff
@@ -136,11 +140,19 @@ def preprocess_expression(expression, complex_mode=False,
...
     expression = apply_function_pullbacks(expression)
-    expression = apply_geometry_lowering(expression, preserve_geometry_types)
+    expression = apply_geometry_lowering(expression, lowering_preserve_types)
+    expression = apply_derivatives(expression)
+    expression = apply_geometry_lowering(expression, lowering_preserve_types)
     expression = apply_derivatives(expression)
+    # Cancel contractions of the Jacobian with its inverse before
+    # expanding the inverse into individual matrix entries
+    expression = remove_component_tensors(expression)
+    expression = cancel_jacobian_products(expression)
```
connorjward, line 155: Do we really need to replace 1 pass with 5? Genuine question, I don't necessarily think you're wrong.

pbrubeck: Yes. There is no way around this, and this is just mimicking what preprocess form will do.

We are adding `cancel_jacobian_products`, but in order to get there we split `geometry_lowering` in two steps (the first one needs to preserve the types to be cancelled, the second one lowers all geometry types as usual). `remove_component_tensors` is a pre-requisite. And the derivatives need to happen again because the `Jacobian` is the derivative of the physical `SpatialCoordinate` with respect to the reference coordinates.

connorjward (dismissed):

PR #5231, opened 2026-07-10
dmcommon: de-Python-ify mesh-topology construction (~4x faster BoxMesh build)

miguelcoolchips:
## Summary

Performance work on unstructured mesh-topology construction in `firedrake/cython/dmcommon.pyx`. Profiling a `BoxMesh(79,79,79)` (2.96M tets, 512k P1 DOFs) showed mesh+assembly — not the solve — dominating wall-clock (~111 s vs ~1 s solve), and the cost was Python-level overhead inside the per-entity / per-facet Cython loops (petsc4py `dm.getCellType`, `math.factorial`, and `cdef np.ndarray` element indexing), **not** the numerics.

This replaces that overhead with the C DMPlex API and typed memoryviews. It keeps the GIL and all existing error handling, adds **no new build dependency**, and produces **bit-identical** output.

## Changes (one commit each)

| Kernel | Before | After |
|---|---|---|
| `entity_orientations` | 19.7 s | 2.1 s |
| `_facets` (`facet_numbering` + `get_facets_by_class`) | 37.4 s | 0.18 s |
| `cell_closure` (`closure_ordering`) | 22.9 s | 3.6 s |
| `make_cell_node_list` (`get_cell_nodes`) | 9.0 s | 0.20 s |

- **`entity_orientations`**: a single C `DMPlexGetCellType` (+ `DMPolytopeType` enum dispatch) replaces up to 4 petsc4py `dm.getCellType()` calls per entity; a C factorial (`_fact`) replaces `math.factorial`; typed memoryviews for I/O.
- **`facet_numbering` / `get_facets_by_class`**: typed memoryviews instead of `cdef np.ndarray` indexing in the O(nfacets·nclosure) scans (the inner loop index was also not `cdef`-typed).
- **`closure_ordering` / `get_cell_nodes`**: typed memoryviews for the per-cell closure / cell-node maps; permutation & orientation arrays guarded for the swarm / no-permutation cases.

## Result (N=79, 512k DOF)

Mesh + assembly: **110.6 s → 28.5 s (3.9×)**; warm solve unchanged (~1.0 s, 13 CG+GAMG iterations). What remains is now PETSc-internal (`DMPlexCreateFromCellList`, entity renumbering), not Python-bound.

## Benchmark details

Measured on `BoxMesh(79,79,79)` (2.96M tets, 512k P1 DOFs), single process, timing each kernel in isolation, min of 5 repeats.

### Why `entity_orientations` was slow

The per-entity loop runs `numCells × numEntities` ≈ **45M** iterations, and each iteration made several **Python** calls — the cost was the Python overhead, not the arithmetic:

- `dm.getCellType(p)` — a petsc4py method, invoked up to **4×/entity**
- `math.factorial(...)` — a Python call, in an inner loop over the cone
- `cell_closure[cell, e]` / output writes — `np.ndarray` scalar (Python-level) indexing

Replacing these with a single C `DMPlexGetCellType` + enum dispatch, a `cdef inline` C factorial, and typed memoryviews took the kernel from **19.7 s → 2.11 s (9.3×)**.

### Isolating the `math.factorial` → `_fact` change

To confirm the C factorial was itself worth it (raised in review), an A/B was run changing **only** the two factorial call sites, with the C `DMPlexGetCellType` dispatch and typed memoryviews identical in both builds:

| factorial impl | `entity_orientations` (min of 5) |
|---|---|
| `_fact` (C, `cdef inline`) | **2.16 s** |
| `math.factorial` (Python) | **5.55 s** |

Reverting just that one line regresses the kernel **~2.6×**, i.e. the Python `math.factorial` call alone accounts for **~3.4 s (~60%)** of the kernel's remaining time. `math.factorial` is a Python attribute lookup + call returning a Python `int` (then boxed into the arithmetic), executed on the order of 10^8 times; `_fact` is a `cdef inline PetscInt` (m ≤ 4) the compiler reduces to a couple of integer multiplies. The two overheads the PR removes (`dm.getCellType` and `math.factorial`) are independent and both material.

## Correctness

At N=79, `cell_closure`, `entity_orientations`, and `cell_node_list` are `np.array_equal` to the pre-change baseline; interior-facet `jump(u)*dS` for a CG field is ~4e-14; a CG Poisson solve with a Dirichlet BC gives unchanged `Tmax`. End-to-end solves verified on both simplex (tet) and tensor-product (quad) meshes, exercising both orientation dispatch paths.

### Reproduce

```python
from firedrake import BoxMesh
import time

mesh = BoxMesh(79, 79, 79, 1.0, 1.0, 1.0)
cc = mesh.topology.cell_closure          # build the closure once (not timed)
t = time.perf_counter()
eo = dmcommon.entity_orientations(mesh.topology, cc)
print("entity_orientations:", time.perf_counter() - t, "s")
```

Compare against `main` by saving `np.save("ref.npy", eo)` on the baseline and asserting `np.array_equal` after the change.

## Scope / notes

- A further ~15× on `entity_orientations` alone is possible by dropping the GIL and multithreading the loop (`nogil` + `prange` + OpenMP), but that adds an OpenMP build dependency and interacts with MPI process placement, so it is intentionally **out of scope** here.
- The benchmark/verification harness that produced these numbers previously lived under `bench/`; per review it has been removed from the tree, and the before/after summary is recorded above.


pbrubeck:
Can we move these benchmarks out of the repo? However it is still good to have these results documented, just include a brief summary of the before and after in the PR description.

pbrubeck:
Please add a new entry in AGENTS.md, telling it to not add design/benchmark notes to the repo

firedrake/cython/dmcommon.pyx
```diff
@@ -310,45 +316,47 @@ def facet_numbering(PETSc.DM plex, kind,
         cells_per_facet = 1
     facet_local_num = np.empty((nfacets, cells_per_facet), dtype=IntType)
     facet_cells = np.empty((nfacets, cells_per_facet), dtype=IntType)
```
pbrubeck, line 312: Why are we not creating these as C arrays from the beginning?

firedrake/cython/dmcommon.pyx
```diff
@@ -1054,7 +1074,7 @@ cdef inline PetscInt _compute_orientation_simplex(PetscInt *fiat_cone,
         coneSize1 -= 1
     assert n == coneSize
     for k in range(n):
-        o += math.factorial(n - 1 - k) * inds[k]
+        o += _fact(n - 1 - k) * inds[k]
```
pbrubeck, line 1067: Was this really impacting performance??

firedrake/cython/dmcommon.pyx
```diff
@@ -310,45 +316,47 @@ def facet_numbering(PETSc.DM plex, kind,
         cells_per_facet = 1
     facet_local_num = np.empty((nfacets, cells_per_facet), dtype=IntType)
     facet_cells = np.empty((nfacets, cells_per_facet), dtype=IntType)
+    facet_cells_view = facet_cells
+    facet_local_num_view = facet_local_num
```
pbrubeck: This approach seems like a patch. It introduces more variables to cover up for our mistakes. Although the view approach gives a tidier diff (mainly because views are multidimensional arrays and we don't need to compute the flat index ourselves). However, I think that we should go all the way and get rid of the numpy arrays from the beginning, and just wrap the result as numpy before returning it.

pbrubeck (changes requested):

pbrubeck:
@connorjward what is your view on views?

connorjward:
> @connorjward what is your view on views?

I don't know what you mean. Obviously we don't want copies.

pbrubeck:
> I don't know what you mean. Obviously we don't want copies.

I mean
https://cython.readthedocs.io/en/0.29.x/src/userguide/memoryviews.html#:~:text=Typed%20memoryviews%20allow%20efficient%20access,more%20features%20and%20cleaner%20syntax.

firedrake/cython/dmcommon.pyx
miguelcoolchips: Yes, I added the performance effect to the description

firedrake/cython/dmcommon.pyx
miguelcoolchips: I prototyped this. I took `entity_orientations` all the way to raw C pointers — `_compute_orientation` takes `const PetscInt *closure, PetscInt ncols` and I index the buffers via `np.PyArray_DATA` with manual flat offsets (`ptr[cell*ncols + e]`), no memoryviews at all — then measured against the current `np.empty` + typed-memoryview version.

**BoxMesh(79,79,79) (2.96M tets), `entity_orientations` in isolation, min-of-7, 3 trials each:**

| build | trial mins |
|---|---|
| memoryview (this PR) | 2.205, 2.219, 2.285 s |
| raw C pointer | 2.205, 2.250, 2.214 s |

No measurable difference — the two are within run-to-run noise. Output is bit-identical (`np.array_equal == True`).

The reason is that a contiguous typed memoryview (`PetscInt[:, ::1]`) with `@boundscheck(False)`/`@wraparound(False)` is not a runtime abstraction: Cython lowers `view[cell, e]` to exactly `data[cell*stride + e]`, the same pointer arithmetic as the manual C version, and the memoryview struct is built once before the loop. The generated C for the hot loop is essentially identical.

So going "all the way" to C arrays from the start buys nothing on speed here, and it costs either an extra copy to wrap the buffer as NumPy on return, or manual malloc + a capsule base to free it, plus manual flat-index arithmetic at every access — which is exactly the indexing the view lets us avoid. Given that, I'd propose keeping the memoryview approach. Happy to revisit if you'd still prefer the C-array-from-the-start version for consistency reasons.

firedrake/cython/dmcommon.pyx
miguelcoolchips: Good point on the duplicate variable. I've removed it: each output array is now declared once as a typed memoryview, filled directly, and wrapped with `np.asarray` on return (zero-copy for a contiguous buffer) — no separate `np.ndarray` + `_view` pair. Done for all five kernels; input-side views are unchanged.

```cython
cdef PetscInt[:, ::1] facet_cells       # was: np.ndarray facet_cells + facet_cells_view
...
facet_cells = np.empty((nfacets, cells_per_facet), dtype=IntType)
...
return np.asarray(facet_local_num), np.asarray(facet_cells)
```

I kept `np.empty` rather than a raw C malloc, since (as measured above) it's the same speed and avoids manual flat indexing and buffer-ownership handling. Verified bit-identical `cell_closure` / `entity_orientations` / `cell_node_list` on tet and quad meshes, timing unchanged. Pushed in 9bb163079.

miguelcoolchips:
@pbrubeck I addressed your comments. This is ready for a new round.

firedrake/cython/dmcommon.pyx
```suggestion
                    np.ndarray facets_np,
                    PETSc.Section cell_numbering,
                    np.ndarray cell_closures_np):
```
Then `facets_view` can be renamed to `facets`, etc, and most of the diff will be gone

miguelcoolchips: You're right — my first attempt (the `_np` rename) barely moved the diff; measuring it, `facet_numbering` only went 25 → 21 changed lines vs `main` and the whole file actually grew slightly, because renaming the parameter forced docstring changes and extra `cdef` init lines that cancelled out the body savings.

I've redone it properly: keep the array's **name** and change only its **type**. The `def` parameters/locals that were read through a `*_view` alias are now typed memoryviews directly:

```cython
def facet_numbering(PETSc.DM plex, kind,
                    const PetscInt[::1] facets,
                    PETSc.Section cell_numbering,
                    const PetscInt[:, ::1] cell_closures):
```

Now the hot-loop bodies are byte-identical to `main`, so only the declarations differ:

- `facet_numbering`: **25 → 11** changed lines vs `main`
- whole file: **116 → 95**

Applied to `facet_numbering`, `get_facets_by_class` and `get_cell_nodes`. `entity_orientations` keeps a single view (its param feeds `np.zeros_like`), and `get_cell_nodes` keeps `entity_orientations_np` only for the `None`/swarm check. Bit-identical `cell_closure` / `entity_orientations` / `cell_node_list` on tet and quad (CG2) meshes. Pushed in 578192c6c.

pbrubeck (dismissed):
This is a great improvement. Thanks @miguelcoolchips !

firedrake/cython/dmcommon.pyx
```diff
@@ -1145,9 +1157,13 @@ cdef inline PetscInt _compute_orientation(PETSc.DM dm,
     cdef:
         PetscInt p, coneSize, offset, i, dim, o
         const PetscInt *cone = NULL
+        PetscDMPolytopeType ct
 
     p = cell_closure[cell, e]
-    if dm.getCellType(p) == PETSc.DM.PolytopeType.POINT:
+    # A single C DMPlexGetCellType replaces the up-to-4 petsc4py dm.getCellType()
+    # Python calls that dominated this ~45M-iteration loop.
```
pbrubeck: Really? Who cares? Document the present not the past!

```suggestion
```

miguelcoolchips:
@pbrubeck can we merge this one?

pbrubeck (approved):

pbrubeck (approved):

PR #5232, opened 2026-07-10
Improve firedrake-run-split-tests

connorjward:
* Do not bind MPI jobs so full utilisation achieved - don't run jobs on the same cores
* Try to kill subprocesses upon SIGINT
* Don't error if errcode files don't exist

### DIFF
--- a/scripts/firedrake-run-split-tests
+++ b/scripts/firedrake-run-split-tests
@@ -54,6 +54,18 @@ extra_args=${@:3}
 job_timeout=${FIREDRAKE_RUN_SPLIT_TESTS_TIMEOUT:-1800s}
 kill_after=${FIREDRAKE_RUN_SPLIT_TESTS_KILL_AFTER:-60s}
 
+# Callback to kill child processes if Ctrl-C hit
+pids=()
+cleanup() {
+    for pid in "${pids[@]}"; do
+        if ps -p "$pid" > /dev/null; then
+            kill "$pid"  # Send SIGTERM to the sub-process
+            wait "$pid"
+        fi
+    done
+}
+trap cleanup SIGINT
+
 timeout_exec=""
 if command -v timeout >/dev/null 2>&1; then
     timeout_exec="timeout"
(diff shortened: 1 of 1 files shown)


scripts/firedrake-run-split-tests
```diff
@@ -54,6 +54,18 @@ extra_args=${@:3}
...
+pids=()
+cleanup() {
+    for pid in "${pids[@]}"; do
+        if ps -p "$pid" > /dev/null; then
+            kill "$pid"  # Send SIGTERM to the sub-process
+            wait "$pid"
+        fi
+    done
+}
+trap cleanup SIGINT
```
JHopeCollins, line 67: Does the `trap cleanup SIGINT` mean "call `cleanup` if SIGINT is sent"? If so, how come we are then sending SIGTERM to the subprocesses rather than forwarding the SIGINT?

JHopeCollins (commented):
Generally this looks fine, seems like the right thing to do. I just had one small query.

scripts/firedrake-run-split-tests
connorjward: Great point. I didn't think about this. I just tried and SIGINT actually kills the child processes more reliably so I'll use that.

scripts/firedrake-run-split-tests
```diff
@@ -54,6 +54,18 @@ extra_args=${@:3}
 job_timeout=${FIREDRAKE_RUN_SPLIT_TESTS_TIMEOUT:-1800s}
 kill_after=${FIREDRAKE_RUN_SPLIT_TESTS_KILL_AFTER:-60s}
 
+# Callback to kill child processes if Ctrl-C hit
+pids=()
+cleanup() {
+    for pid in "${pids[@]}"; do
+        if ps -p "$pid" > /dev/null; then
+            kill "$pid"  # Send SIGTERM to the sub-process
```
connorjward:
```suggestion
            kill -INT "$pid"  # Send SIGINT to the sub-process
```

JHopeCollins (approved):

PR #5215, opened 2026-07-03
Mesh Adaptivity via DMPlexTransform 

pbrubeck:
# Description
AI-assisted

Enables adaptivity on any simplicial mesh via `DMPlex.adaptLabel()`/`DMPlex.transform()`
Depends on https://gitlab.com/petsc/petsc/-/merge_requests/9412 (merged) which introduces 3D adaptivity

Enables `prolong`/`restrict`/`inject` to natively support adaptively-refined hierarchies

Deprecate hierarchy-agnostic `AdaptiveMeshHierarchy`/`AdaptiveTransferManager` in favour of `MeshHierarchy`/`TransferManager`, now with adaptivity support.

PR #5235, opened 2026-07-10
Destroy object the PETSc garbage collector would have issue with

UZerbinati:
PETSc Garbage collector was hanging on my machine when running test locally :)

### DIFF
--- a/firedrake/mesh.py
+++ b/firedrake/mesh.py
@@ -2944,10 +2944,13 @@ values from f.)"""
                 mark_np = mvec.getArray()[cellNum]
             else:
                 sfBCInv = self.sfBC_orig.createInverse()
-                _, mvec0 = self.topology_dm.distributeField(sfBCInv,
-                                                            self._cell_numbering,
-                                                            mvec)
-                mark_np = mvec0.getArray()
+                section0, mvec0 = self.topology_dm.distributeField(sfBCInv,
+                                                                   self._cell_numbering,
+                                                                   mvec)
+                mark_np = mvec0.getArray().copy()
+                sfBCInv.destroy()
+                section0.destroy()
+                mvec0.destroy()
         max_refs = 0 if mark_np.size == 0 else int(mark_np.max())
         # Create a copy of the netgen mesh
         netgen_mesh = self.netgen_mesh.Copy()
--- a/firedrake/netgen.py
+++ b/firedrake/netgen.py
@@ -70,12 +70,18 @@ def netgen_distribute(V: firedrake.functionspaceimpl.WithGeometryBase,
         for i in np.ndindex(V.shape):
             di = netgen_data[(..., *i)].flatten()
             vec0[:len(di)] = di
-            _, vec = plex.distributeField(sf, section0, vec0)
-            arr = vec.getArray()
+            section_i, vec_i = plex.distributeField(sf, section0, vec0)
+            arr = vec_i.getArray()
             if plex_data is None:
                 plex_data = np.empty(arr.shape + V.shape, dtype=dtype)
             plex_data[(..., *i)] = arr
+            section_i.destroy()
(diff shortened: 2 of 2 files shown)


pbrubeck:
Please add parallel tests

connorjward:
This shouldn't be hanging... the garbage collector should never do this

UZerbinati:
> Please add parallel tests

What do you mean ? ngsPETSc and netgen interface is already tested in parallel, no ?

pbrubeck:
> What do you mean ? ngsPETSc and netgen interface is already tested in parallel, no ?

This does not seem to be the case for the adaptive refinement code in Netgen. However we're going to replace this ad-hoc renumbering hell with DMAdaptLabel in #5215

connorjward:
Was this happening reliably or stochastically? Others have also random observed garbage collector hangs and I would like to resolve it.

pbrubeck:
> Was this happening reliably or stochastically? Others have also random observed garbage collector hangs and I would like to resolve it.

How old are those reports? I fixed a very common hanging issue that was triggered by repetedly calling multigrid not too long ago https://gitlab.com/petsc/petsc/-/merge_requests/8694

UZerbinati:
> > What do you mean ? ngsPETSc and netgen interface is already tested in parallel, no ?
> 
> This does not seem to be the case for the adaptive refinement code in Netgen. However we're going to replace this ad-hoc renumbering hell with DMAdaptLabel in #5215

Looking forward to that :)

UZerbinati:
> Was this happening reliably or stochastically? Others have also random observed garbage collector hangs and I would like to resolve it.

Happened consistently on my machine and on @ThighamOxford machine consistently. Everything was up to date current main.

connorjward:
> > Was this happening reliably or stochastically? Others have also random observed garbage collector hangs and I would like to resolve it.
> 
> Happened consistently on my machine and on @ThighamOxford machine consistently. Everything was up to date current main.

That's awesome. If I find the time I really need to dig into what's causing this.

UZerbinati:
I still believe is not a bad idea to destroy stuff we will not use anyway, no ? Or is this not standard practice in Firedrake ?

connorjward:
> I still believe is not a bad idea to destroy stuff we will not use anyway, no ? Or is this not standard practice in Firedrake ?

In theory it's a completely pointless thing to do because Python will destroy things for us. We try to avoid C-style coding in favour of being Pythonic. This should be added to our style guide.

However, it doesn't do any harm and is clearly exposing an issue with the garbage collector. I am happy to merge this if @pbrubeck is.

pbrubeck:
> > I still believe is not a bad idea to destroy stuff we will not use anyway, no ? Or is this not standard practice in Firedrake ?
> 
> In theory it's a completely pointless thing to do because Python will destroy things for us. We try to avoid C-style coding in favour of being Pythonic. This should be added to our style guide.
> 
> However, it doesn't do any harm and is clearly exposing an issue with the garbage collector. I am happy to merge this if @pbrubeck is.

I think we should give it another go with Claude with this in mind, it could be very helpful to get to the bottom of this.

firedrake/mesh.py
```diff
@@ -2944,10 +2944,13 @@ def refine_marked_elements(self, mark, netgen_flags=None):
...
             else:
                 sfBCInv = self.sfBC_orig.createInverse()
-                _, mvec0 = self.topology_dm.distributeField(sfBCInv,
-                                                            self._cell_numbering,
-                                                            mvec)
-                mark_np = mvec0.getArray()
+                section0, mvec0 = self.topology_dm.distributeField(sfBCInv,
+                                                                   self._cell_numbering,
+                                                                   mvec)
+                mark_np = mvec0.getArray().copy()
```
pbrubeck, line 2950: Maybe this copy is all we needed?

PR #5240, opened 2026-07-13
firedrake-configure: scope apt HDF5 to --with-hdf5-* and declare zlib1g-dev

timCorintis:
# Description

This PR introduces two fixes to the setup configuration: 
1. Improvement: firedrake-configure represents the system HDF5 location inconsistently — unscoped OPTFLAGS -I/-L on the apt path, instead of a scoped --with-hdf5-* selector
2. Issue: firedrake-configure requests --with-zlib on apt but omits zlib from --show-system-packages

### Improvement 1 - Scope the apt HDF5 location via `--with-hdf5-*` instead of `OPTFLAGS` `-I/-L`
On the apt paths (`--package-manager apt-x86_64` and `apt-aarch64`), `--show-petsc-configure-options` expresses HDF5's system location as raw `-I`/`-L` baked into `--COPTFLAGS/--CXXOPTFLAGS/--FOPTFLAGS`:
```
--COPTFLAGS='… -I/usr/include/hdf5/openmpi … -L/usr/lib/<triplet>/hdf5/openmpi'
```
rather than through a package-scoped selector. This is inconsistent with how `firedrake-configure` handles the same concept elsewhere:
- **ScaLAPACK** on the same apt path uses a scoped selector: `--with-scalapack-lib=-lscalapack-openmpi`.
- **HDF5** on the brew path uses a scoped selector: `--with-hdf5-dir=/opt/homebrew`.
So only the **apt + HDF5** combination puts the location outside the `--with-X*` selector pattern. Because the path isn't tied to the HDF5 selector, a consumer that switches to `--download-hdf5` can drop the `--with-hdf5` selector but not the stray `-L`, so PETSc links the system HDF5 while h5py loads the downloaded one, giving two HDF5s in one process (`H5Iget_type` runtime error).
Improvement: use `--with-hdf5-include` / `--with-hdf5-lib` on the apt archs so the location travels with the selector.

### Fix 2 - Declare `zlib1g-dev` on the apt path
On apt, configure asks for `--with-zlib` but `--show-system-packages` never lists `zlib1g-dev`. It works today only because another `-dev` package pulls zlib in transitively. For a build that installs exactly the declared set, zlib becomes an undeclared dependency (missing from the SBOM) and breaks on a minimal image.
Fix: add `zlib1g-dev` to the apt archs' system packages, matching what brew already does.

## Changes

- `scripts/firedrake-configure`: for every apt-based arch, replace `--with-hdf5` + the HDF5 entries in `include_dirs`/`library_dirs` with explicit `--with-hdf5-include` / `--with-hdf5-lib` options; add `zlib1g-dev` to the apt archs' system packages.

### DIFF
--- a/scripts/check-config
+++ b/scripts/check-config
@@ -97,8 +97,8 @@ def check_zlib_declared() -> None:
         for arch_key, arch in all_archs.items()
         if "--with-zlib" in arch.extra_petsc_configure_options
         and (arch.system_packages is None
-             or arch.system_packages is not None
-             and "zlib1g-dev" not in arch.system_packages)
+             or (arch.system_packages is not None
+             and "zlib1g-dev" not in arch.system_packages))
     ]
     if offenders:
         raise InvalidConfigurationException(
--- a/scripts/check-config
+++ b/scripts/check-config
@@ -27,6 +27,7 @@ def main():
     check_slepc_version()
     check_petsc_version_spec()
     check_zlib_declared()
+    check_hdf5_scoped()
 
 
 def check_min_python_version() -> None:
@@ -101,6 +102,33 @@ def check_zlib_declared() -> None:
         raise InvalidConfigurationException(
             "The following archs request '--with-zlib' but do not declare "
             f"'zlib1g-dev' in their system packages: {offenders}"
+        )
+
+
+def check_hdf5_scoped() -> None:
+    """Check that HDF5's location is expressed via a scoped selector.
+
+    An arch's HDF5 location must be with a ``--with-hdf5-*`` option, not as
+    raw ``-I``/``-L`` entries in ``include_dirs``/``library_dirs``. The latter
+    end up in the shared OPTFLAGS, where they cannot be removed by a consumer that overrides HDF5 to
+    ``--download-hdf5``.
+
+    """
+    firedrake_configure = _load_firedrake_configure()
+    all_archs = {
+        **firedrake_configure.OFFICIAL_ARCHS,
+        **firedrake_configure.RETIRED_ARCHS,
+        **firedrake_configure.COMMUNITY_ARCHS,
+    }
+    offenders = [
+        ", ".join(key_part.value for key_part in arch_key)
+        for arch_key, arch in all_archs.items()
+        if any("hdf5" in dir_ for dir_ in (*arch.include_dirs, *arch.library_dirs))
+    ]
+    if offenders:
+        raise InvalidConfigurationException(
+            "The following archs put an HDF5 path in include_dirs/library_dirs "
+            f"instead of a '--with-hdf5-*' selector: {offenders}"
         )
 
 
--- a/scripts/check-config
+++ b/scripts/check-config
@@ -8,8 +8,10 @@ of hardcoded values (via CI).
 
 """
 
+import importlib.util
 import pathlib
 import re
+from importlib.machinery import SourceFileLoader
 
 
 REPO_ROOT = pathlib.Path(__file__).parent.parent
@@ -24,6 +26,7 @@ def main():
     check_petsc_version()
     check_slepc_version()
     check_petsc_version_spec()
+    check_zlib_declared()
 
 
 def check_min_python_version() -> None:
@@ -70,6 +73,50 @@ def check_petsc_version_spec() -> None:
         REPO_ROOT / "setup.py",
         f"petsctools.init\\(version_spec=\"{petsc_version_spec}\"\\)",
     )
+
+
+def check_zlib_declared() -> None:
+    """Check that archs linking the system zlib also declare its package.
+
+    If an arch passes ``--with-zlib`` to PETSc configure then it links against
+    the system zlib, so that library must appear in ``system_packages`` (as
+    ``zlib1g-dev``). Otherwise zlib is an undeclared dependency that only works
+    by being pulled in transitively by another package.
+
+    """
+    firedrake_configure = _load_firedrake_configure()
+    all_archs = {
+        **firedrake_configure.OFFICIAL_ARCHS,
+        **firedrake_configure.RETIRED_ARCHS,
+        **firedrake_configure.COMMUNITY_ARCHS,
+    }
+    offenders = [
+        ", ".join(key_part.value for key_part in arch_key)
+        for arch_key, arch in all_archs.items()
+        if "--with-zlib" in arch.extra_petsc_configure_options
+        and arch.system_packages is not None
+        and "zlib1g-dev" not in arch.system_packages
+    ]
+    if offenders:
+        raise InvalidConfigurationException(
+            "The following archs request '--with-zlib' but do not declare "
+            f"'zlib1g-dev' in their system packages: {offenders}"
+        )
+
+
+def _load_firedrake_configure():
+    """Import the 'firedrake-configure' script as a module.
+
+    The hyphen in the filename means it cannot be imported normally, so load it
+    directly from its path.
+
+    """
+    script_path = REPO_ROOT / "scripts" / "firedrake-configure"
+    loader = SourceFileLoader("firedrake_configure", str(script_path))
+    spec = importlib.util.spec_from_loader("firedrake_configure", loader)
+    module = importlib.util.module_from_spec(spec)
+    loader.exec_module(module)
+    return module
 
 
 def check_file_contains_pattern(
--- a/scripts/check-config
+++ b/scripts/check-config
@@ -1,18 +1,15 @@
 #!/usr/bin/env python3
 
-"""Sanity checks for hardcoded configuration in the Firedrake repository.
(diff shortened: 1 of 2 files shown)


scripts/check-config
```diff
@@ -72,6 +76,77 @@ def check_petsc_version_spec() -> None:
...
+        **firedrake_configure.OFFICIAL_ARCHS,
+        **firedrake_configure.RETIRED_ARCHS,
+        **firedrake_configure.COMMUNITY_ARCHS,
+    }
+    offenders = [
+        ", ".join(key_part.value for key_part in arch_key)
+        for arch_key, arch in all_archs.items()
+        if "--with-zlib" in arch.extra_petsc_configure_options
+        and arch.system_packages is not None
+        and "zlib1g-dev" not in arch.system_packages
```
ab-corintis:
```suggestion
        and (arch.system_packages is None or 
        arch.system_packages is not None and "zlib1g-dev" not in arch.system_packages)
```

scripts/check-config
ab-corintis, line 1: Looks like we're extending the declared purpose of this file with the additional checks. Update the documentation header comment too?

scripts/firedrake-configure
```diff
@@ -439,11 +439,13 @@ OFFICIAL_ARCHS: Mapping[ArchKey, Arch] = {
             "libsuitesparse-dev",
             "libsuperlu-dev",
             "libsuperlu-dist-dev",
+            "zlib1g-dev",
```
ab-corintis: This should go under the first section of this list, not in the "PETSc external packages part".

scripts/firedrake-configure
```diff
@@ -502,12 +500,14 @@ OFFICIAL_ARCHS: Mapping[ArchKey, Arch] = {
...
```
ab-corintis: Same comment as above.

scripts/firedrake-configure
```diff
@@ -567,11 +563,13 @@ OFFICIAL_ARCHS: Mapping[ArchKey, Arch] = {
...
```
ab-corintis: Same comment as above.

scripts/firedrake-configure
```diff
@@ -630,12 +624,14 @@ OFFICIAL_ARCHS: Mapping[ArchKey, Arch] = {
...
```
ab-corintis: Same comment as above.

scripts/firedrake-configure
```diff
@@ -691,6 +683,7 @@ OFFICIAL_ARCHS: Mapping[ArchKey, Arch] = {
             "libpnetcdf-dev",
             "libptscotch-dev",
             "libscalapack-openmpi-dev",
+            "zlib1g-dev",
```
ab-corintis: Same comment as above.

scripts/firedrake-configure
```diff
@@ -949,11 +939,13 @@ RETIRED_ARCHS: Mapping[ArchKey, RetiredArch] = {
             "libsuitesparse-dev",
             "libsuperlu-dev",
             "libsuperlu-dist-dev",
+            "zlib1g-dev",
```
ab-corintis: Same comment as above.

scripts/firedrake-configure
```diff
@@ -1011,12 +999,14 @@ RETIRED_ARCHS: Mapping[ArchKey, RetiredArch] = {
...
```
ab-corintis: Same comment as above.

scripts/firedrake-configure
```diff
@@ -1075,11 +1061,13 @@ RETIRED_ARCHS: Mapping[ArchKey, RetiredArch] = {
...
```
ab-corintis: Same comment as above.

scripts/firedrake-configure
```diff
@@ -1137,12 +1121,14 @@ RETIRED_ARCHS: Mapping[ArchKey, RetiredArch] = {
...
```
ab-corintis: Same comment as above.

scripts/check-config
```diff
@@ -72,6 +77,78 @@ def check_petsc_version_spec() -> None:
...
+        **firedrake_configure.RETIRED_ARCHS,
+        **firedrake_configure.COMMUNITY_ARCHS,
+    }
+    offenders = [
+        ", ".join(key_part.value for key_part in arch_key)
+        for arch_key, arch in all_archs.items()
+        if "--with-zlib" in arch.extra_petsc_configure_options
+        and (arch.system_packages is None
+             or arch.system_packages is not None
+             and "zlib1g-dev" not in arch.system_packages)
```
ab-corintis: Better with extra parenthesis for clarity:
```suggestion
             or (arch.system_packages is not None
             and "zlib1g-dev" not in arch.system_packages))
```

scripts/check-config
```diff
@@ -72,6 +77,78 @@ def check_petsc_version_spec() -> None:
     )
 
 
+def check_zlib_declared() -> None:
```
connorjward: We don't really test `firedrake-configure` like this. We 'test' it by validating that CI runs.

connorjward:
> Issue: firedrake-configure represents the system HDF5 location inconsistently — unscoped OPTFLAGS -I/-L on the apt path, instead of a scoped --with-hdf5-* selector

I don't think that this is actually an issue. We are allowed to be inconsistent because each ARCH is a standalone config. I have no issue with the proposed changes here - it does seem better - but the problem is not that it is inconsistent IMO.

> Because the path isn't tied to the HDF5 selector, a consumer that switches to --download-hdf5 can drop the --with-hdf5 selector but not the stray -L, so PETSc links the system HDF5 while h5py loads the downloaded one, giving two HDF5s in one process (H5Iget_type runtime error).

As soon as the user decides to customise their install in this way then it is not the problem of `firedrake-configure`. If you are setting the configuration flags yourself then you get to keep the pieces.

connorjward (changes requested):
The tests are not necessary. No issues with the actual changes to `firedrake-configure` though.

ab-corintis:
> > Issue: firedrake-configure represents the system HDF5 location inconsistently — unscoped OPTFLAGS -I/-L on the apt path, instead of a scoped --with-hdf5-* selector
> 
> I don't think that this is actually an issue. We are allowed to be inconsistent because each ARCH is a standalone config. I have no issue with the proposed changes here - it does seem better - but the problem is not that it is inconsistent IMO.
> 
> > Because the path isn't tied to the HDF5 selector, a consumer that switches to --download-hdf5 can drop the --with-hdf5 selector but not the stray -L, so PETSc links the system HDF5 while h5py loads the downloaded one, giving two HDF5s in one process (H5Iget_type runtime error).
> 
> As soon as the user decides to customise their install in this way then it is not the problem of `firedrake-configure`. If you are setting the configuration flags yourself then you get to keep the pieces.

The issue is that we're trying to set up a consistent build, with everything built from a known source - and the way firedrake-configure works right now we cannot link the PETSc binary with the hdf5 library it builds from source because the generated command line hardcodes to link against the system libraries instead. I.e. we need to manually patch the generated command line to actually use the source-built hdf5 library. With this fix that manual patch shouldn't be needed anymore.

connorjward:
> I.e. we need to manually patch the generated command line

To be clear. This is exactly what we expect from users. We very explicitly do not support any sort of custom build. `firedrake-configure` is there to cater for regular users who likely have no experience compiling software before. For power users it should just be seen as a starting point from which they will customise their build.

That said, the changes here seem entirely fine and I am happy to merge them.

scripts/firedrake-configure
```diff
@@ -1058,6 +1044,7 @@ RETIRED_ARCHS: Mapping[ArchKey, RetiredArch] = {
             "pkg-config",
             "python3-dev",
             "python3-pip",
+            "zlib1g-dev",
```
connorjward: I don't really want to change any of the 'retired' configurations. The previous build completely worked and we can't know that the new one won't break for certain from this change, unless each is tested manually.

JHopeCollins:
For context, I added a bunch of the include/lib directories to the general compiler options so that we did not need to know the exact names of the files (e.g. knowing you need both `libhdf5` _and_ `libhdf5_hl`). This makes it a bit easier to maintain and less likely we need to modify it in the future for newer OS updates.

The packages where the includes/libs are specified fully in the petsc configuration options are only those where just adding the locations to the compiler options did not work.

So this isn't a fix as such, it's changing how we specify configurations to make it easier to change the HDF5 build (although leaving hypre, scotch, etc in the same position as before with specified include directories in the general compiler options).

scripts/firedrake-configure
JHopeCollins: Agreed, the retired ones should be static. Otherwise they aren't retired they're just untested.

timCorintis: Fair enough, done!

connorjward (approved):

PR #5243, opened 2026-07-13
Mutable FunctionSpaces

achanbour:
This PR describes my first attempt at making `FunctionSpace`s mutable. Note: this currently only works for FS defined on VOMs (the generalisation to arbitrary meshes is TBD). 

The main idea is that FS need to detect somehow that the VOM, against which they were built, has changed (topologically). We want this is to be done lazily. This is because there are usually a large number of FS laying around and there is no straightforward way of keeping track through all of them: from the mesh's perspective, `FunctionSpace`s with the same element, DoF layout etc. share the same `_shared_data` object cached on the mesh - a`FunctionSpaceData` object carrying all topology-derived numerical data.

Using this cached attribute allows us to invalidate the FS at the mesh level. At the same time, we want each FS to detect this invalidation and to refresh itself upon on its own **first** use after the VOM has mutated. Mutating the VOM does the following:

1. invalidates all the topological properties relating to FS defined on it: `topology._shared_data.clear()`
2. bumps the version number stored on the mesh topology object: `topology._topology_version += 1`

The lazy FS rebuild is implemented here as follows:

1. first, stash the mesh topology version on the FS object
2. add a version check at the upstream source i.e., `_shared_data` which, upon detecting a topology version mismatch, causes all the attributes to be recomputed (`refresh_shared_data` calls `get_shared_data` which returns a new `FunctionSpaceData` object with the updated data)
3. re-set all the downstream topology-depend properties using the updated `_shared_data`
4. sets the FS topology version tracker to the current VOM topology version

To trigger the indirect rebuild, cached properties  such as `cell_node_list` (derived from `_shared_data`) were turned into version-gated properties (which execute steps 2-4 upon a version mismatch). This required changing these properties from being `@cached_property` to being plain `@property` with their own caching mechanism. All other, non-topology-related properties were kept as `@cached_property`. 

Final note: The version gating mechanism described above has only been implemented for `FunctionSpace` and has not been ported yet into `MixedFunctionSpace`.

### DIFF
--- a/firedrake/functionspaceimpl.py
+++ b/firedrake/functionspaceimpl.py
@@ -330,12 +330,7 @@ class WithGeometryBase:
         return MixedFunctionSpace((self, other))
 
     def __getattr__(self, name):
-        # Fetch attributes from the underlying topological FS.
         val = getattr(self.topological, name)
-        # Avoid caching on the WG FS if the topological FS is defined on a mutable mesh as we expect these FS to change
-        # and want to be able to detect staleness in such cases
-        if not getattr(self.topological.mesh(), "_topology_is_mutable", False):
-            setattr(self, name, val)
         return val
 
     def __dir__(self):
@@ -565,9 +560,6 @@ class FunctionSpace:
         self._ufl_function_space = ufl.FunctionSpace(mesh.ufl_mesh(), element, label=self._label)
         self._mesh = mesh
 
-        self._mesh_topology_version = self._mesh._topology_version
-        r"""The topology version of the mesh on which this :class:`FunctionSpace` is defined."""
-
         self.value_size = self._ufl_function_space.value_size
         r"""The number of scalar components of this :class:`FunctionSpace`."""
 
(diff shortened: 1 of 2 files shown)


firedrake/functionspaceimpl.py
```diff
@@ -185,12 +186,29 @@ def sub(self, i):
...
+    # @cached_property
+    # def dm(self):
+    #     dm = self._dm()
+    #     dmhooks.set_function_space(dm, self)
+    #     return dm
+    
+    @property
     def dm(self):
-        dm = self._dm()
+        current_version = getattr(self.topological.mesh(), "_topology_version", None)
```
connorjward: You should use `cached_property_until` (https://github.com/firedrakeproject/firedrake/pull/5180)

firedrake/functionspaceimpl.py
```diff
@@ -577,12 +602,37 @@ def __init__(self, mesh, element, name=None):
         self.comm = mesh.comm
 
         self.set_shared_data()
+        r"""The topology version of the mesh on which the function space is defined."""
```
connorjward: Comment doesn't make sense to be here.

firedrake/functionspaceimpl.py
```diff
@@ -577,12 +602,37 @@ def __init__(self, mesh, element, name=None):
...
         self.node_set = self.dof_dset.set
         r"""A :class:`pyop2.types.set.Set` representing the function space nodes."""
 
+
+    @property
+    def _shared_data(self):
+        if hasattr(self, "_vom_topology_version") \
+            and self._vom_topology_version != self._mesh._topology_version:
+            self.refresh_shared_data()
+        return self.__shared_data
```
connorjward: I'd avoid a double underscore because it invokes Python name mangling. Maybe `self._shared_data_private`?

connorjward (changes requested):
This doesn't provide any mechanism for migrating functions when the topology changes. Without that something like
```py
vom = VertexOnlyMesh(...)
V = FunctionSpace(vom, ...)
f = Function(V)

vom.migrate(...)

assemble(f*dx)  # will be absolute gibberish
```

achanbour (commented):
Implemented changes after first review

firedrake/functionspaceimpl.py
```diff
@@ -328,8 +330,12 @@ def __mul__(self, other):
         return MixedFunctionSpace((self, other))
 
     def __getattr__(self, name):
+        # Fetch attributes from the underlying topological FS.
         val = getattr(self.topological, name)
-        setattr(self, name, val)
+        # Avoid caching on the WG FS if the topological FS is defined on a mutable mesh as we expect these FS to change
+        # and want to be able to detect staleness in such cases
+        if not getattr(self.topological.mesh(), "_topology_is_mutable", False):
+            setattr(self, name, val)
```
connorjward: I think this was a bad design choice in the first place. I think just remove the `setattr` entirely.

firedrake/mesh.py
```diff
@@ -501,6 +501,9 @@ def plex_from_cell_list(dim, cells, coords, comm, name=None):
 class AbstractMeshTopology(object, metaclass=abc.ABCMeta):
     """A representation of an abstract mesh topology without a concrete
         PETSc DM implementation"""
+    
+    _topology_is_mutable = False
+    """Whether this topology may change in place after construction."""
```
connorjward: Please don't set this on the abstract class. I would set it explicitly on all the subclasses. You could even go the extra mile and make it an abstract method here.

Speaking from experience, this pattern gets extremely confusing.

firedrake/mesh.py
```diff
@@ -605,6 +608,11 @@ def __init__(self, topology_dm, name, reorder, sfXB, perm_is, distribution_name,
         # target_mesh._parallel_compatible = {weakref.ref(source_mesh)}
         self._parallel_compatible = None
 
+        self._topology_version = 0
+        if self._topology_is_mutable:
+            self._topology_step_sfs = {}
```
connorjward, line 610: Probably shouldn't be in this PR unless you end up touching it somewhere.

firedrake/functionspaceimpl.py
```diff
@@ -577,12 +586,33 @@ def __init__(self, mesh, element, name=None):
...
+    def _shared_data(self):
+        # I have deliberately NOT set @cached_property_until here, although the version check is the
+        # same pattern: on staleness this property must refresh the whole shared data object -- based on which
+        # several properties like which dof_dset, global_numbering etc. get re-assigned as a side effect --
+        # while the caching decorator can only memoise a single return value.
+        if self._mesh_topology_version != self._mesh._topology_version:
+            self.refresh_shared_data()  # owns the refresh
+        return self._shared_data_private
+
+    def refresh_shared_data(self):
```
connorjward: This pattern isn't very pleasant. As an alternative why not make `dof_dset` etc into `@property`s? Then you can use the `cached_property_until` decorator just on `self._shared_data`.

connorjward (changes requested):

firedrake/mesh.py
achanbour: I left it here for now given that the FS caching mechanism (which is what this PR is about) relies on the versioning of the mesh's topology.

achanbour:
I think we can close this PR given that it's been moved, along with the other VOM-related mutation, stuff into https://github.com/firedrakeproject/firedrake/pull/5294

PR #5244, opened 2026-07-13
Write VOM data to VTKFile

achanbour:
Currently, `VTKFile` can't handle Functions defined on a VOM, it raises an 'Unhandled cell type` error. 

This PR is an attempt to support writing VOM data into a `VTKFile` which then enables particle data to be plotted in ParaView alongside mesh output.

- Fields on a VOM are DG0 by definition which enables data arrays to be written directly in `_write_vtu`
- To account for a changing VOM, the topology is re-computed at every write using `get_vom_topology` rather than cached
- A persistent particle ID field is written by default (based on the swarm`globalindex` field) allowing individual particles to be tracked across time steps.

Demo usage

```
vom = VertexOnlyMesh(mesh, points)
P0DG = VectorFunctionSpace(vom, "DG", 0)
u_p = Function(P0DG, name="velocity").interpolate(u)

outfile = VTKFile("particles.pvd")
for step in range(nsteps):
    # ... advect particles ...
    outfile.write(u_p, time=t)
```


firedrake/output/vtk_output.py
```diff
@@ -229,6 +232,15 @@ def vrange(cell_layers):
             OFunction(offsets_into_con, "offsets", None),
             OFunction(cell_types, "types", None))
 
+def get_vom_topology(mesh):
```
connorjward: This needs a docstring. I don't understand what it's doing.

firedrake/output/vtk_output.py
```diff
@@ -499,7 +511,10 @@ def _write_vtu(self, *functions):
         if len(functions) == 1 and isinstance(functions[0], ufl.Mesh):
             from firedrake.functionspace import FunctionSpace
             mesh = functions[0]
-            V = FunctionSpace(mesh, "CG", 1)
+            if isinstance(mesh.topology, VertexOnlyMeshTopology):
+                V = FunctionSpace(mesh, "DG", 0)
+            else:
+                V = FunctionSpace(mesh, "CG", 1)
```
connorjward: Would a better check be `if mesh.topological_dimension > 0`?

connorjward (changes requested):
Please add a couple tests.

Also, what things actually break when you use a VoM? It would be nicer for the if/else blocks to be smaller. A VoM shares a lot of its interface with a typical mesh so I am not sure that we can't get more code reuse.

firedrake/output/vtk_output.py
dham: roll this function into the general case above.

achanbour:
I have removed the code block inside `_write_vtu` that attaches a default particle ID field. Given the VOM's tendency to mutate, a particle ID field must undergo through the migration process for the values to be properly re-assigned to the (migrating) particles. This migration should be managed outside the `write` Function, I think.

With the current implementation, given an externally defined particle ID field:

`pid = Function(DG0, dtype=IntType, name="particle-ids")`,

we can write it to the VTK output file as:

`outfile.write(pid)`

Otherwise, an alternative solution would be to have the VOM carry a `particleid` swarm field that is assigned once and migrated at each rebuild alongside the other fields (with the migration being handled in the VOM updater class).

firedrake/mesh.py
```diff
@@ -3709,6 +3717,15 @@ def VertexOnlyMesh(mesh, vertexcoords, reorder=None, missing_points_behaviour='e
...
     vmesh_out = make_vom_from_vom_topology(topology, name, tolerance)
     vmesh_out._parent_mesh = mesh
+
+    # Attach persistent particle IDs
+    P0 = FunctionSpace(vmesh_out, "DG", 0)
+    pid = Function(P0, dtype=IntType, name="firedrake_particle_ids")
+    n_owned = vmesh_out.cell_set.size
+    offset = vmesh_out.comm.scan(n_owned) - n_owned
+    pid.dat.data_wo[:] = np.arange(offset, offset + n_owned, dtype=IntType)
+    vmesh_out._firedrake_particle_ids = pid
```
connorjward: We don't need to namespace the attribute here. `particle_ids` is fine.

firedrake/mesh.py
```diff
@@ -2404,6 +2404,11 @@ def _ufl_signature_data_(self, *args, **kwargs):
         return (type(self), self.extruded, self.variable_layers,
                 super()._ufl_signature_data_(*args, **kwargs))
 
+    @property
+    def firedrake_particle_ids(self):
+        """Persistent particle IDs for a VertexOnlyMesh."""
+        return getattr(self, "_firedrake_particle_ids", None)
```
connorjward: Shouldn't this only live on the topology? And it isn't a great pattern to be using `getattr` here. The attribute should always be defined (on a suitable object).

connorjward: I also like just leaving this as a private attribute (leading underscore). If we make things public then we need to think about deprecation cycles etc.

connorjward (changes requested):

firedrake/mesh.py
achanbour: I don't think I can make the ID field part of the topology because `particle_ids` is a `Function` which needs the geometry that the topology layer does not yet have. That's why I've defined in the factory method, after the geometry's been created on the `VertexOnlyMeshTopology` object.

The reason I made this a property is because I wouldn't want users to override the ID field. It wouldn't raise an error if it was solely an attribute (even with an underscore). But I'm happy to get rid of this if we assume that such cases won't happen.

achanbour: I believe defining it in the VOM factory is enough (naturally non-VOM meshes won't have this attribute).

connorjward: It might be possible to do if you make it a `CoordinatelessFunction` (which is what we do for coordinates).

docs/source/visualisation.rst
```diff
@@ -370,6 +373,26 @@ Last but not least, vector fields defined on a vertex-only mesh can be visualise
...
+   :language: python3
+   :dedent:
+   :start-after: [test_can_save_vom_data 1]
+   :end-before: [test_can_save_vom_data 2]
+
+.. figure:: images/vom-paraview.png
+   :alt: Vertex-only mesh particle IDs visualised in ParaView
+   :align: center
+
+   Vertex-only mesh data visualised in ParaView, with particles coloured by thier IDs.
```
connorjward:
```suggestion
   Vertex-only mesh data visualised in ParaView, with particles coloured by their IDs.
```

connorjward (dismissed):
Nice!

connorjward (approved):

PR #5245, opened 2026-07-14
Use CachedSolverBlock for project (-> #4638)

angus-g:
# Description

The `ProjectBlock` was essentially an extension of `SolveVarFormBlock` that sets up the mass matrix system for the adjoint machinery. For the original (forward) `project()` call, the optimised solver-free path is used. However, any of the adjoint operations including a forward recomputation would solve the system.

The approach here replicates that behaviour, but it's routed through the new `CachedSolverBlock`. As a consequence however, we have annotation on the `Projector` factory, which means that adjoint-friendly projection can occur through either the bare `project()` call, or by creating a `Projector` object.

The implementation is a bit ugly: because a lot of the caching machinery belongs to the `NonlinearVariationalSolverMixin`, we essentially dispatch to a dummy initialiser when creating a `BasicProjector` to ensure the solver caches are created correctly. A call to `project() `annotates as though a solve occurred, but uses the underlying projection implementation.

The `SupermeshSolveBlock` is actually a standalone object, so we just create that when `SupermeshProjector.project()` is called. Again, this allows for both bare and object-based projection to be taped.

The final case is the "null" project from one function to another on the same space. This creates an `Assigner`, and I assume it will "just work" through the annotation on `assign()`.

## Questions
- Does this break the API (also the `solve()` change) by removing the `annotate` kwarg? Now we have to wrap things in `with stop_annotating()` or similar
- There's probably a better way to manage `ad_block_tag`
- There's probably a better way to manage the solver parameters (indeed, if you just solve `inner(u,w)*dx == inner(u+c,w)*dx` for a Real c and try to evaluate the tape, there's a divergence... guess there aren't any adjoint tests for dead simple solves)

### DIFF
--- a/firedrake/projection.py
+++ b/firedrake/projection.py
@@ -83,8 +83,6 @@ def project(
         Quadrature degree to use when approximating integrands.
     name
         The name of the resulting :class:`.Function`.
-    ad_block_tag
-        String for tagging the resulting block on the Pyadjoint tape.
 
     Returns
     -------
@@ -110,19 +108,22 @@ def project(
         solver_parameters=solver_parameters,
         form_compiler_parameters=form_compiler_parameters,
         use_slate_for_inverse=use_slate_for_inverse,
-        quadrature_degree=quadrature_degree
+        quadrature_degree=quadrature_degree,
+        ad_block_tag=ad_block_tag,
     ).project()
     val.rename(name)
     return val
 
 
 class Assigner(object):
-    def __init__(self, source, target):
+    def __init__(self, source, target, **kwargs):
         self.source = source
         self.target = target
 
+        self._kwargs = kwargs
+
     def project(self):
-        self.target.assign(self.source)
+        self.target.assign(self.source, **self._kwargs)
         return self.target
 
 
@@ -245,6 +246,7 @@ class BasicProjector(ProjectorBase, NonlinearVariationalSolverMixin):
     """
 
     def __init__(self, *args, **kwargs):
+        ad_block_tag = kwargs.pop("ad_block_tag", None)
         super().__init__(*args, **kwargs)
 
         V = self.target.function_space()
@@ -257,7 +259,7 @@ class BasicProjector(ProjectorBase, NonlinearVariationalSolverMixin):
         L = firedrake.inner(self.source, w) * dx
         p = firedrake.LinearVariationalProblem(a, L, self.target)
 
-        self._init_as_solver(p)
+        self._init_as_solver(p, ad_block_tag=ad_block_tag)
 
     @NonlinearVariationalSolverMixin._ad_annotate_init
     def _init_as_solver(self, problem):
@@ -309,6 +311,8 @@ class BasicProjector(ProjectorBase, NonlinearVariationalSolverMixin):
 class SupermeshProjector(ProjectorBase):
     def __init__(self, *args, **kwargs):
         self._target_is_function = kwargs.pop("target_is_function", True)
+        self.ad_block_tag = kwargs.pop("ad_block_tag", None)
+
         super().__init__(*args, **kwargs)
 
     @cached_property
@@ -337,7 +341,8 @@ def Projector(
     form_compiler_parameters: Optional[dict] = None,
     constant_jacobian: Optional[bool] = True,
     use_slate_for_inverse: Optional[bool] = False,
-    quadrature_degree: Optional[Union[int, tuple[int]]] = None
+    quadrature_degree: Optional[Union[int, tuple[int]]] = None,
+    ad_block_tag: Optional[str] = None,
 ):
     """ Projection class.
 
@@ -369,6 +374,8 @@ def Projector(
         function spaces)(only valid for DG function spaces).
     quadrature_degree
         Quadrature degree to use when approximating integrands.
+    ad_block_tag
+        String for tagging the resulting block on the Pyadjoint tape.
     """
     target = create_output(v_out)
     source = sanitise_input(v, target.function_space())
@@ -377,14 +384,15 @@ def Projector(
         raise ValueError("Shape mismatch between source %s and target %s in project" %
                          (source.ufl_shape, target.ufl_shape))
     if isinstance(v, function.Function) and not bcs and v.function_space() == target.function_space():
-        return Assigner(source, target)
+        return Assigner(source, target, ad_block_tag=ad_block_tag)
     elif source_mesh == target_mesh:
         return BasicProjector(
             source, target, bcs=bcs, solver_parameters=solver_parameters,
             form_compiler_parameters=form_compiler_parameters,
             constant_jacobian=constant_jacobian,
             use_slate_for_inverse=use_slate_for_inverse,
-            quadrature_degree=quadrature_degree
+            quadrature_degree=quadrature_degree,
+            ad_block_tag=ad_block_tag,
         )
     else:
         if bcs is not None:
@@ -398,4 +406,5 @@ def Projector(
             use_slate_for_inverse=use_slate_for_inverse,
             quadrature_degree=quadrature_degree,
             target_is_function=isinstance(v_out, firedrake.Function),
+            ad_block_tag=ad_block_tag,
         )
--- a/firedrake/projection.py
+++ b/firedrake/projection.py
@@ -6,6 +6,7 @@ from ufl.domain import extract_unique_domain
 from typing import Optional, Union
 
 import firedrake
+from firedrake.adjoint_utils import NonlinearVariationalSolverMixin, annotate_super_project
 from firedrake.bcs import BCBase
 from firedrake.petsc import PETSc
 from functools import cached_property
@@ -13,7 +14,6 @@ from functools import cached_property
 from firedrake.utils import complex_mode, SLATE_SUPPORTS_COMPLEX
 from firedrake import functionspaceimpl
 from firedrake import function
-from firedrake.adjoint_utils import annotate_project
 
 
 __all__ = ['project', 'Projector']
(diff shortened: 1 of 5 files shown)


angus-g:
Not really sure why the `test_adjoint_cleanup` case is failing, it's such a simple solve, but it's failing on the adjoint and that should be fine? Unless something about the form manipulation or block dependencies has changed?

```py
from firedrake import *

continue_annotation()
mesh = SquareMesh(1, 1, 1, quadrilateral=True)

u_0 = Function(V).assign(1.0)
u = Function(V).assign(u_0)
r = Function(R)

r.assign(2.0)
u.project(r * u)
r.assign(1.0)
u.project(r * u)

J = assemble((u - Function(V).assign(1.0)) ** 2 * dx)

pause_annotation()
reduced_functional = ReducedFunctional(J, Control(u_0))

# Deliberate additional derivative computation.
reduced_functional.derivative()
```

angus-g:
Okay, I've got a slightly more minimal reproducer that avoids the project path. This actually doesn't rely on the `CachedSolveBlock` and highlights a difference between the bare `solve()` annotation (through `SolveVarFormBlock`) and the `NonlinearVariationalSolver` annotation. Because `ProjectBlock` uses `SolveVarFormBlock`, tihs is the issue we're hitting.

```python3
from firedrake import *                                                                                                                                                 
                                                                                                                                                                        
continue_annotation()                                                                                                                                                   
                                                                                                                                                                        
mesh = UnitSquareMesh(1, 1)                                                                                                                                             
u = Function(V)                                                                                                                                                         
w = TestFunction(V)                                                                                                                                                     
c = Function(R, val=1.)                                                                                                                                                 
                                                                                                                                                                        
a = inner(TrialFunction(V), w) * dx                                                                                                                                     
L = inner(c*u, w) * dx                                                                                                                                                  

if reproduce:
    solve(a == L, u)  # diverges for this PR, *not* main
else:
    # always diverges
    problem = LinearVariationalProblem(a, L, u)
    NonlinearVariationalSolver(problem).solve()

J = assemble(u**2 * dx)                                                                                                                                                 
rf = ReducedFunctional(J, Control(c))                                                                                                                                   
                                                                                                                                                                        
pause_annotation()                                                                                                                                                      
                                                                                                                                                                        
rf.derivative()
```

Here, the bilinear form is essentially the projection form that was causing the issues above. When we go through the NLVP, the bilinear form becomes linear during the static method `NonlinearVariationalProblem.compute_bc_lifting`, which sets `F = action(F, u)`. Here, we get a form like:

```
w₃ * (conj((v_0))) * dx(<Mesh #1>[everywhere], {}, {})
  +  -1 * w₃ * w₆ * (conj((v_0))) * dx(<Mesh #1>[everywhere], {}, {})
```

Note that `u = w₃` is present in both terms. I guess as an adjoint, this becomes degenerate? Anyway, the bare `solve()` annotation constructed a `SolveVarFormBlock` directly from the bilinear form: it extracts the LHS and RHS, and uses a different trial function:

```python3
class GenericSolveBlock(Block):
    ...
    
    def _create_F_form(self):
        if self.linear:
            tmp_u = Function(self.function_space)
            F_form = action(self.lhs, tmp_u) - self.rhs
            
        ...
```

This means the form is something like

```
w₅₇ * (conj((v_0))) * dx(<Mesh #1>[everywhere], {}, {})
  +  -1 * w₃₉ * w₄₁ * (conj((v_0))) * dx(<Mesh #1>[everywhere], {}, {})
```

and so the derivative and adjoint thereof only has one term. This is probably the "correct" behaviour, because even if we work around the degenerate solve in the test case:

```diff
diff --git a/tests/firedrake/adjoint/test_assignment.py 
b/tests/firedrake/adjoint/test_assignment.py
index 342355076..d1884fc38 100644
--- a/tests/firedrake/adjoint/test_assignment.py
+++ b/tests/firedrake/adjoint/test_assignment.py
@@ -254,7 +254,7 @@ def test_adjoint_cleanup(scheduler, rg):

     r.assign(2.0)
     u.project(r * u)
-    r.assign(1.0)
+    r.assign(1.5)
     u.project(r * u)

     J = assemble((u - Function(V).assign(1.0)) ** 2 * dx)
```

The Taylor test fails:

```
>       assert taylor_test(reduced_functional, u_0, dtemp) > 1.99999999
E       AssertionError: assert 1.0009069989745567 > 1.99999999
E        +  where 1.0009069989745567 = taylor_test(<pyadjoint.reduced_functional.ReducedFunctional object at 0x7ff8097d3c80>, Coefficient(WithGeometry(FunctionSpace(<firedrake.mesh.MeshTopology object at 0x7ff80be92900>, FiniteElement('Q', quadrilateral, 1), name=None), Mesh(VectorElement(FiniteElement('Q', quadrilateral, 1), dim=2), 1)), 3), Coefficient(WithGeometry(FunctionSpace(<firedrake.mesh.MeshTopology object at 0x7ff80be92900>, FiniteElement('Q', quadrilateral, 1), name=None), Mesh(VectorElement(FiniteElement('Q', quadrilateral, 1), dim=2), 1)), 166))
```

JHopeCollins:
> Okay, I've got a slightly more minimal reproducer that avoids the project path

~~What is the divergence message? Doesn't the the residual immediately vanish? If we were using an iterative solve then we may end up relying on `atol` which is The Wrong Thing, but the default solver is `-ksp_type preonly -pc_type lu` which should completely ignore the residuals.~~

~~Potentially the projection block becomes a very thin wrapper that creates a temporary `Function` for the projected solution to avoid the case where the result of the projection is present in the projected expression (equivalent to what currently ends up happening in the generic solve block).~~

On second thought, I do expect this to fail. `u.project(u*r)` isn't meant to solve `u = u*r` because `u` is going to be mutated in-place. We really do want to solve `u_new = u_old*r`. The current project block is getting around this by introducing a new unknown `Function` for the solution of the projection.
In essence this is just pointing out that in-place projection is equivalent to `solve -> assign`. Potentially that would also be the shortest way to implement in-place projection with taping, but that will almost certainly will have other implications to deal with so I don't think that's the right thing to do here/now.

How about we have a `ProjectorMixin` and a `ProjectBlock` that inherit from `NLVSMixin` and `CachedSolverBlock` respectively. These would be very thin wrappers that just handle the creation of a temporary `Function` to project into and then assigning into the target (and appropriate wrappers for the adjoint etc re-evaluations).
They wouldn't even need to know whether the projection is in place or not, it would work for either case. We'd have one extra copy than required in the not-in-place case but we really don't care about that performance-wise.

JHopeCollins:
> Does this break the API (also the solve() change) by removing the annotate kwarg? Now we have to wrap things in with stop_annotating() or similar

Yes it does, we shouldn't be removing this kwarg. Is it causing problems?

angus-g:
> How about we have a ProjectorMixin and a ProjectBlock that inherit from NLVSMixin and CachedSolverBlock respectively.

Thanks Josh, I'll have a look at implementing this.

> Yes it does, we shouldn't be removing this kwarg. Is it causing problems?

Just with the rejig of things it needs a bit of wiring up to work properly. No biggie, I'll make sure it works and revert the test!

angus-g:
I don't really understand what's going on with the CI breaking at the install phase (@connorjward ideas? Did I do something dumb?). I've addressed the other two issues though (the "implicit" assign and annotate kwarg) @JHopeCollins if you have time to take another look?

connorjward:
> I don't really understand what's going on with the CI breaking at the install phase (@connorjward ideas? Did I do something dumb?). I've addressed the other two issues though (the "implicit" assign and annotate kwarg) @JHopeCollins if you have time to take another look?

I think you just need to merge in main.

angus-g:
> I think you just need to merge in main.

Aha, I thought I had done that last week when I hit this before. Seems to be working now, thanks!

JHopeCollins:
> @JHopeCollins if you have time to take another look?

@angus-g I'm on leave at the moment but I can have a look when I'm back next week.

firedrake/projection.py
```diff
@@ -84,8 +85,6 @@ def project(
         Quadrature degree to use when approximating integrands.
     name
         The name of the resulting :class:`.Function`.
-    ad_block_tag
-        String for tagging the resulting block on the Pyadjoint tape.
```
JHopeCollins, line 88: Why has this been removed?

firedrake/projection.py
```diff
@@ -245,6 +247,43 @@ class BasicProjector(ProjectorBase):
...
+        # assignment. Instead of representing this as an explicit solve and assign
+        # block, we solve for a temporary function and modify _ad_u on the problem.
+        # This is used for creating the output variable (and thus the dependency
+        # resolution), in effect merging the implicit assignment into the block
+        # output phase.
+        p._ad_u = self.target
+
+        solver_params = firedrake.LinearVariationalSolver.DEFAULT_SNES_PARAMETERS | self.solver_parameters
+
+        self._init_as_solver(p, forward_kwargs={"solver_parameters": solver_params}, ad_block_tag=ad_block_tag)
```
JHopeCollins, line 277: If this is all just to handle setting up the block then it should go in `adjoint_utils`.

firedrake/projection.py
```diff
@@ -236,7 +238,7 @@ def project(self):
         return self.target
 
 
-class BasicProjector(ProjectorBase):
+class BasicProjector(ProjectorBase, NonlinearVariationalSolverMixin):
```
JHopeCollins, line 241: I think that we want a `ProjectorMixin` and a `ProjectBlock` that inherit from `NonlinearVariationalSolverMixin` and `SolverBlock` respectively.

These would be fairly thin wrappers, but it means that all the noodling around to piggyback the solver taping is kept in `adjoint_utils` so this file stays simpler and we don't pay anything if we're not taping.
It also means that the block that appears on the tape is `ProjectBlock` rather than `SolverBlock` which would potentially be confusing.

It would also do away with things like the `_init_as_solver` method and decorating the `project` method with `_ad_annotate_solve`.

PR #5247, opened 2026-07-14
rtree bulk query

leo-collins:
- Use new bulk query routine to query the rtree. We now do 3 FFI calls instead of (potentially) 100,000s.
- firedrake-rtree IDs now use int64 instead of size_t.
- Refactored `locate.c` to reduce code duplication. We now also check the `cells_ignore` array *before* computing the reference coordinates.
- Other various tidying, e.g. using bools, tidying headers, checking bounds when casting to PetscInt

Requires a new release of firedrake-rtree.


firedrake/cython/rtree.pyx
```diff
@@ -76,24 +76,24 @@ def build_from_aabb(np.ndarray[np.float64_t, ndim=2, mode="c"] coords_min,
...
 
-    n = <size_t>coords_min.shape[0]
-    dim = <size_t>coords_min.shape[1]
+    n = coords_min.shape[0]
+    dim = coords_min.shape[1]
     if ids is None:
-        ids = np.arange(n, dtype=np.uintp)
-    elif ids.shape[0] != n:
+        ids = np.arange(n, dtype=np.int64)
+    elif <size_t>ids.shape[0] != n:
```
connorjward, line 89: Why all these dtype changes?

leo-collins: firedrake-rtree now requires int64 bounding box IDs, like libspatialindex did.

leo-collins: Also `dim` is an unsigned 32 bit

firedrake/function.py
```diff
@@ -651,6 +651,8 @@ def single_eval(x, buf):
                                                         buf.ctypes.data_as(c_void_p))
             if err == -1:
```
connorjward, line 652: Is this still right? Don't we want to check `cell_out`?

firedrake/locate.c
```diff
@@ -27,85 +23,54 @@ PetscInt locate_cell(struct Function *f,
...
-            }
-            if (current_ref_cell_dist_l1 <= 0.0) {
-                /* Found cell! */
-                cell = ids[i];
-                memcpy(found_ref_coords, temp_ref_coords, sizeof(struct ReferenceCoords));
-                found_ref_cell_dist_l1[0] = current_ref_cell_dist_l1;
+    *cell_out = -1;
+    for (size_t i = 0; i < nids; ++i) {
+        if (ids[i] > (int64_t)PETSC_MAX_INT) {
+            return PETSC_ERR_ARG_OUTOFRANGE;
```
connorjward, line 36: Add comment explaining why this might happen. It isn't obvious to me.

connorjward: And if this is some sort of indicator that things aren't found wouldn't -1 be a better choice?

firedrake/mesh.py
connorjward, line 2744: Slightly out of scope but could you add a docstring/comments? This is a really complicated piece of code.

firedrake/mesh.py
```diff
@@ -2751,28 +2753,50 @@ def _c_locator(self, tolerance=None):
...
+                        PetscErrorCode locate_err = locate_cell_from_candidates(
+                            f, &x[j], &to_reference_coords, &to_reference_coords_xtr,
+                            &temp_reference_coords, &found_reference_coords,
+                            &ref_cell_dists_l1[i], nids_i, ids_i,
+                            ncells_ignore, cells_ignore_i, &cells[i]);
+                        if (locate_err != PETSC_SUCCESS) {{
+                            rtree_free_ids(candidate_ids, candidate_offsets[npoints]);
+                            rtree_free_offsets(candidate_offsets, npoints + 1);
+                            return locate_err;
+                        }}
```
connorjward, line 2799: More natural to

```suggestion
                        if (locate_err != PETSC_SUCCESS) {{
                            break;
                        }}
```
?

Generally in C I think it is bad style to exit early because cleanup is hard

connorjward (changes requested):
Seems pretty much fine

connorjward:
I assume this is no longer a draft?

firedrake/locate.c
leo-collins: This is just checking that casting the ids from int64 to PetscInt is safe (for 32 bit petsc builds). Since the ids are mesh cell ids it *should* always be safe

firedrake/mesh.py
leo-collins: If we just break then this function will return 0 and silently drop the error? The block as it is has all the cleanup we need to do in it.

firedrake/function.py
leo-collins: This is the old `Function.at` code which calls a different C function defined in `pointeval_utils`. I suppose the rtree error path shouldn't return -1 which is what the "cell not found" error path returns.

firedrake/mesh.py
connorjward: I mean to change the `return 0` at the end of the function to `return locate_err`. This isn't critical, but I think it's tidier/better practice.

firedrake/locate.c
connorjward: Ah I see. Definitely needs a comment but makes sense that it's there.

firedrake/mesh.py
leo-collins: That makes sense. Added this in now.

firedrake/locate.c
leo-collins: Added a comment.

connorjward (dismissed):
Seems good to me, happy to merge once firedrake-rtree is released

pbrubeck:
Should this be merged now?

leo-collins:
Yes, this is ready now

connorjward (approved):

PR #5249, opened 2026-07-14
Fix BDDCPC/FDMPC pyop3

pbrubeck:
- bddc.py: dof_dset no longer exists on FunctionSpace; use the new V.lgmap()/V.template_vec accessors instead of V.dof_dset.lgmap and V.dof_dset.layout_vec.
- fdm.py broken_function(): op2 is no longer imported in this file (only pyop3), switch to READ/WRITE from firedrake.parloops. Also fix the manual loopy kernel to use 2D (dof, component) indexing, which par_loop now requires unconditionally, and pass dtype=val.dtype to Function() since make_dat now asserts on dtype mismatches (this function is routinely called with integer index/lgmap arrays).
- Re-enable tests/firedrake/regression/test_bddc.py (was fully module-skipped with "pyop3 TODO").

Verified serially: BDDCPC solves, test_create_matis (4/4), test_vertex_dofs (8/8 @ nprocs=3), test_bddc_cellwise_fdm (12/12 @ nprocs=1, including condense=True since that config uses fdm_pc_use_amat=False and so never reaches FDMPC.condense()).

Still broken/out of scope of this commit:
- FDMPC.condense() (matrix-free static condensation, fdm_pc_use_amat=True) was never ported to pyop3 and still raises NotImplementedError.
- assemble(..., mat_type="is") + DirichletBC fails in parallel (nprocs>1) with "Matrix is missing diagonal entry" from MatZeroRowsColumnsLocal. This is a core assemble.py bug unrelated to preconditioners.

# Description

### DIFF
--- a/firedrake/preconditioners/bddc.py
+++ b/firedrake/preconditioners/bddc.py
@@ -112,11 +112,11 @@ class BDDCPC(PCBase):
             dir_nodes = numpy.unique(numpy.concatenate([bcdofs(bc, ghost=False) for bc in bcs]))
         neu_nodes = numpy.setdiff1d(boundary_nodes, dir_nodes)
 
-        dir_nodes = V.dof_dset.lgmap.apply(dir_nodes)
+        dir_nodes = V.lgmap().apply(dir_nodes)
         dir_bndr = PETSc.IS().createGeneral(dir_nodes, comm=pc.comm)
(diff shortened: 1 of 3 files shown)


connorjward:
Merged in. Thanks.

Issue #5250, opened 2026-07-15
BUG: _c_locator kernel is not integer type safe


angus-g:
**Describe the bug**
The `cells_ignore` parameter in the `_c_locator` kernel (in `mesh.py`) is incorrectly typed compared to the function prototype. The `cells_ignore` array uses the `IntType` dtype, which is 64- or 32- bit, depending on the PETSc index type. However, the prototype declares it as `np.ctypeslib.ndpointer(ctype.c_int, flags="C_CONTIGUOUS")`. When running with 64-bit PETSc, this gives the (expected) type error:
```
ctypes.ArgumentError: argument 8: TypeError: array must have data type int32
```

Because the array is cast on the way in to the function:
```python
        if cells_ignore is None or cells_ignore[0][0] is None:
            cells_ignore = np.full((npoints, 1), -1, dtype=IntType, order="C")
        else:
            cells_ignore = np.asarray(cells_ignore, dtype=IntType, order="C")
```

it would seem that the simple fix is to ensure the prototype declares it as `IntType`:

```diff
diff --git a/firedrake/mesh.py b/firedrake/mesh.py
index 8e6ebed99..f01bd0dcb 100644
--- a/firedrake/mesh.py
+++ b/firedrake/mesh.py
@@ -37,7 +37,7 @@ import firedrake.cython.extrusion_numbering as extnum
 import firedrake.extrusion_utils as eutils
 import firedrake.cython.rtree as rtree
 import firedrake.utils as utils
-from firedrake.utils import as_cstr, IntType, RealType
+from firedrake.utils import as_cstr, as_ctypes, IntType, RealType
 from firedrake.logging import logger
 from firedrake.parameters import parameters
 from firedrake.petsc import PETSc, DEFAULT_PARTITIONER
@@ -2796,7 +2796,7 @@ values from f.)"""
                                 ctypes.POINTER(ctypes.c_int),
                                 ctypes.c_size_t,
                                 ctypes.c_size_t,
-                                np.ctypeslib.ndpointer(ctypes.c_int, flags="C_CONTIGUOUS")]
+                                np.ctypeslib.ndpointer(as_ctypes(IntType), flags="C_CONTIGUOUS")]
             locator.restype = ctypes.c_int
             return cache.setdefault(tolerance, locator)
```

This isn't correct, because `locate.c` declares that parameter as `int`:
```
int locate_cell(struct Function *f,
        double *x,
        int dim,
[...]

Issue #3913, opened 2024-12-09
BUG: Curved mesh generation fails on Firedrake install with 64-bit integers


ABaierReinio:
When using a Firedrake install with 64-bit integers (i.e. when Firedrake is installed with flag `--petsc-int-type int64`), curved mesh generation fails. For example, consider the curved mesh Netgen demo code

```
from firedrake import *
from netgen.occ import WorkPlane, OCCGeometry
import netgen
from mpi4py import MPI

wp = WorkPlane()
if COMM_WORLD.rank == 0:
    for i in range(6):
        wp.Line(0.6).Arc(0.4, 60)
    shape = wp.Face()
    ngmesh = OCCGeometry(shape,dim=2).GenerateMesh(maxh=1.)
else:
    ngmesh = netgen.libngpy._meshing.Mesh(2)
mesh = Mesh(Mesh(ngmesh, comm=COMM_WORLD).curve_field(4))
VTKFile("output/MeshExample5.pvd").write(mesh)
```

Running this yields an error message

```
  File "/scratch/baierreinio/test.py", line 14, in <module>
    mesh = Mesh(Mesh(ngmesh, comm=COMM_WORLD).curve_field(4))
                ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "petsc4py/PETSc/Log.pyx", line 188, in petsc4py.PETSc.Log.EventDecorator.decorator.wrapped_func
  File "petsc4py/PETSc/Log.pyx", line 189, in petsc4py.PETSc.Log.EventDecorator.decorator.wrapped_func
  File "/scratch/baierreinio/firedrake_aaron/firedrake/lib/python3.12/site-packages/ngsPETSc/utils/firedrake/meshes.py", line 185, in curveField
    ng_index = [*map(lambda x: self.locate_cell(x, tolerance=location_tol), barycentres)]
               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/scratch/baierreinio/firedrake_aaron/firedrake/lib/python3.12/site-packages/ngsPETSc/utils/firedrake/meshes.py", line 185, in <lambda>
    ng_index = [*map(lambda x: self.locate_cell(x, tolerance=location_tol), barycentres)]
[...]

PR #5252, opened 2026-07-15
Petsc 64 bit int fixes

leo-collins:
Fixes https://github.com/firedrakeproject/firedrake/issues/5250 and https://github.com/firedrakeproject/firedrake/issues/3913.

Use `PetscInt` as much as we can essentially. Tested locally on a 64 bit Petsc build

### DIFF
--- a/firedrake/evaluate.h
+++ b/firedrake/evaluate.h
@@ -45,16 +45,16 @@ typedef PetscReal (*ref_cell_l1_dist_xtr)(void *data_,
 				int layer,
 				double *x);
 
-extern int locate_cell(struct Function *f,
+extern PetscInt locate_cell(struct Function *f,
 		       double *x,
 		       int dim,
 		       ref_cell_l1_dist try_candidate,
 		       ref_cell_l1_dist_xtr try_candidate_xtr,
 		       void *temp_ref_coords,
 		       void *found_ref_coords,
-		       double *found_ref_cell_dist_l1,
+		       PetscReal *found_ref_cell_dist_l1,
 			   size_t ncells_ignore,
-			   int* cells_ignore);
+			   PetscInt* cells_ignore);
 
 extern int evaluate(struct Function *f,
 		    double *x,
--- a/firedrake/evaluate.h
+++ b/firedrake/evaluate.h
@@ -9,13 +9,13 @@ extern "C" {
 
 struct Function {
 	/* Number of cells in the base mesh */
-	int n_cols;
+	PetscInt n_cols;
 
 	/* 1 if extruded, 0 if not */
 	int extruded;
 
 	/* number of layers for extruded, otherwise 1 */
-	int n_layers;
+	PetscInt n_layers;
 
 	/* Coordinate values and node mapping */
 	PetscScalar *coords;
--- a/firedrake/evaluate.h
+++ b/firedrake/evaluate.h
@@ -53,7 +53,7 @@ extern PetscInt locate_cell(struct Function *f,
 		       void *temp_ref_coords,
 		       void *found_ref_coords,
 		       PetscReal *found_ref_cell_dist_l1,
-			   size_t ncells_ignore,
+			   PetscInt ncells_ignore,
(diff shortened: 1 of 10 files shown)


connorjward:
Thanks for this hard work!

If you don't mind, would you be able to take this the last step of the way and enable int64 testing on CI? It'd be exactly the same as what we do for complex and single (in https://github.com/firedrakeproject/firedrake/pull/5033). That would be really amazing.

firedrake/evaluate.h
```diff
@@ -36,25 +36,25 @@ struct Function {
...
 		       int dim,
 		       ref_cell_l1_dist try_candidate,
 		       ref_cell_l1_dist_xtr try_candidate_xtr,
 		       void *temp_ref_coords,
 		       void *found_ref_coords,
-		       double *found_ref_cell_dist_l1,
-			   size_t ncells_ignore,
-			   int* cells_ignore);
+		       PetscReal *found_ref_cell_dist_l1,
+			   PetscInt ncells_ignore,
```
JHopeCollins: Is this always going to be large enough? Size_t is unsigned and comes with guarantees that int doesn't

firedrake/locate.c
```diff
@@ -45,7 +45,7 @@ int locate_cell(struct Function *f,
         if (f->extruded == 0) {
             for (uint64_t i = 0; i < nids; i++) {
                 current_ref_cell_dist_l1 = (*try_candidate)(temp_ref_coords, f, ids[i], x);
-                for (uint64_t j = 0; j < ncells_ignore; j++) {
+                for (PetscInt j = 0; j < ncells_ignore; j++) {
```
JHopeCollins: Same question here about int range

firedrake/locate.c
```diff
@@ -76,11 +76,11 @@ int locate_cell(struct Function *f,
...
             for (uint64_t i = 0; i < nids; i++) {
-                int nlayers = f->n_layers;
-                int c = ids[i] / nlayers;
-                int l = ids[i] % nlayers;
+                PetscInt nlayers = f->n_layers;
+                PetscInt c = ids[i] / nlayers;
+                PetscInt l = ids[i] % nlayers;
                 current_ref_cell_dist_l1 = (*try_candidate_xtr)(temp_ref_coords, f, c, l, x);
-                for (uint64_t j = 0; j < ncells_ignore; j++) {
+                for (PetscInt j = 0; j < ncells_ignore; j++) {
```
JHopeCollins: And here

firedrake/evaluate.h
leo-collins: I think you're right. As I understand, `ncells_ignore` should be `size_t` because it is an array length, and therefore the two loop indices you pointed out should also be `size_t`.

firedrake/mesh.py
```diff
@@ -2741,13 +2742,12 @@ def _c_locator(self, tolerance=None):
         try:
             return cache[tolerance]
         except KeyError:
-            IntTypeC = as_cstr(IntType)
             src = pq_utils.src_locate_cell(self, tolerance=tolerance)
             src += dedent(f"""
-                int locator(struct Function *f, double *x, double *X, double *ref_cell_dists_l1, {IntTypeC} *cells, {IntTypeC} npoints, size_t ncells_ignore, int* cells_ignore)
+                int locator(struct Function *f, double *x, {RealType_c} *X, {RealType_c} *ref_cell_dists_l1, {IntType_c} *cells, {IntType_c} npoints, size_t ncells_ignore, {IntType_c}* cells_ignore)
```
JHopeCollins, line 2748:
```suggestion
                # this function returns an error code which is always a 32-bit int rather than IntType_c.
                int locator(struct Function *f, double *x, {RealType_c} *X, {RealType_c} *ref_cell_dists_l1, {IntType_c} *cells, {IntType_c} npoints, size_t ncells_ignore, {IntType_c}* cells_ignore)
```

dham (approved):

PR #5253, opened 2026-07-15
Make the PETSc/wheel optimisation flags overridable in firedrake-configure

timCorintis:
## Why
`firedrake-configure` hardcoded `-march`/`-mtune` per platform with no way to override them. The default `-march=native` bakes in the build host's CPU and SIGILLs on older machines, so redistributable builds had no portable CPU floor.

## What changed
- `--show-petsc-configure-options` now takes a `--` passthrough: everything after `--` is appended verbatim to the generated options. PETSc applies last-one-wins, so those tokens override any generated option, e.g. `firedrake-configure --show-petsc-configure-options -- --COPTFLAGS='-O3 -march=x86-64-v2 -mtune=generic'`.
- Split the include/library dirs out of `--COPTFLAGS`: includes (`-I…`) to `--CFLAGS`/`--CXXFLAGS`/`--FFLAGS`, libraries (`-L…`) to `--LDFLAGS`, optimisation to `--COPTFLAGS`/`--CXXOPTFLAGS`/`--FOPTFLAGS`. Previously everything was in `--COPTFLAGS`, so overriding it stripped the header/library paths and broke the build, now the dirs are preserved.
- Updated the "Setting the CPU optimisation target" section in `install.rst`.

### DIFF
--- a/docs/source/install.rst
+++ b/docs/source/install.rst
@@ -509,6 +509,13 @@ to customise the options that are passed to PETSc ``configure``. You can either:
    <edit my_configure_options.txt>
    $ cat my_configure_options.txt | xargs -L1 ./configure
 
+* Append options directly to ``firedrake-configure`` by listing them after a
+  literal ``--``. They are added to the end of the generated options and, since
+  PETSc applies last-one-wins, override any generated option. For example, to
+  set a portable CPU target instead of the default ``-march=native``::
+
+   $ python3 ../firedrake-configure --show-petsc-configure-options -- --COPTFLAGS='-O3 -march=x86-64-v2 -mtune=generic'
+
 .. note::
    If additional options are passed to ``configure`` then care must be taken when
    using externally-installed system packages (i.e. ``--with-package=...`` or
--- a/scripts/firedrake-configure
+++ b/scripts/firedrake-configure
@@ -52,6 +52,7 @@ import dataclasses
 import enum
 import os
 import platform
+import shlex
 from collections.abc import Mapping, Sequence
 
 
@@ -105,21 +106,21 @@ Please consider passing '--os unknown' or building manually.""")
 
     if args.show_system_packages:
         try:
-            arch.show_system_packages()
+            arch.show_system_packages(args.passthrough)
         except InvalidConfigurationError:
             error(
                 "Don't know what system dependencies to install for this arch "
                 f"'{arch_key_str}', please install them manually"
             )
     elif args.show_petsc_configure_options:
-        arch.show_petsc_configure_options()
+        arch.show_petsc_configure_options(args.passthrough)
     elif args.show_petsc_version:
         print_and_stop(SUPPORTED_PETSC_VERSION)
     elif args.show_extra_repo_pkg_url:
-        arch.show_url()
+        arch.show_url(args.passthrough)
     else:
         assert args.show_env
-        arch.show_env()
+        arch.show_env(args.passthrough)
 
 
 def make_parser() -> argparse.ArgumentParser:
@@ -211,6 +212,16 @@ Please see https://firedrakeproject.org/install for more information."""
         help="Print out the URL of any package required to enable non-OS repo access for this build",
     )
 
+    parser.add_argument(
+        "passthrough",
+        nargs="*",
+        metavar="-- OPTION[=VALUE] ...",
+        help=(
+            "After a literal '--', extra options appended verbatim to the generated "
+            "options. This can be useful to overwrite problematic default options."
+        ),
+    )
+
     return parser
 
 
@@ -283,12 +294,12 @@ class Arch:
     """System packages."""
     extra_petsc_configure_options: Sequence[str] = ()
     """Extra options passed to PETSc configure."""
-    cflags: Sequence[str] = ()
-    """CFLAGS used to compile PETSc and external packages."""
-    cxxflags: Sequence[str] = ()
-    """CXXFLAGS used to compile PETSc and external packages."""
-    fflags: Sequence[str] = ()
-    """FFLAGS used to compile PETSc and external packages."""
+    coptflags: Sequence[str] = ()
+    """Optimisation flags passed to PETSc via --COPTFLAGS."""
+    cxxoptflags: Sequence[str] = ()
+    """Optimisation flags passed to PETSc via --CXXOPTFLAGS."""
+    foptflags: Sequence[str] = ()
+    """Optimisation flags passed to PETSc via --FOPTFLAGS."""
     include_dirs: Sequence[str] = ()
     """Extra include directories used during PETSc configure and install"""
     library_dirs: Sequence[str] = ()
@@ -313,33 +324,37 @@ class Arch:
         "--with-strict-petscerrorcode",
     ]
 
-    def show_system_packages(self) -> None:
+    def show_system_packages(self, extra_options: Sequence[str] = ()) -> None:
         if self.system_packages is None:
             raise InvalidConfigurationError
         else:
-            print_and_stop(" ".join(self.system_packages))
+            packages = [*self.system_packages, *(shlex.quote(option) for option in extra_options)]
+            print_and_stop(" ".join(packages))
 
-    def show_petsc_configure_options(self) -> None:
-        include_dirs = [f"-I{idir}" for idir in self.include_dirs]
-        library_dirs = [f"-L{ldir}" for ldir in self.library_dirs]
-        cflags = " ".join((*self.cflags, *include_dirs, *library_dirs))
-        cxxflags = " ".join((*self.cxxflags, *include_dirs, *library_dirs))
-        fflags = " ".join((*self.fflags, *include_dirs, *library_dirs))
+    def show_petsc_configure_options(self, extra_options: Sequence[str] = ()) -> None:
+        include_dirs = " ".join(f"-I{idir}" for idir in self.include_dirs)
+        library_dirs = " ".join(f"-L{ldir}" for ldir in self.library_dirs)
         opts = [
             f"PETSC_ARCH=arch-firedrake-{self.name}",
-            f"--COPTFLAGS='{cflags}'",
-            f"--CXXOPTFLAGS='{cxxflags}'",
-            f"--FOPTFLAGS='{fflags}'",
+            *([f"--CFLAGS='{include_dirs}'"] if include_dirs else []),
+            *([f"--CXXFLAGS='{include_dirs}'"] if include_dirs else []),
+            *([f"--FFLAGS='{include_dirs}'"] if include_dirs else []),
+            *([f"--LDFLAGS='{library_dirs}'"] if library_dirs else []),
+            f"--COPTFLAGS='{' '.join(self.coptflags)}'",
+            f"--CXXOPTFLAGS='{' '.join(self.cxxoptflags)}'",
+            f"--FOPTFLAGS='{' '.join(self.foptflags)}'",
             *self.common_petsc_configure_options,
             *self.extra_petsc_configure_options,
+            *(shlex.quote(option) for option in extra_options),
         ]
         print_and_stop(" ".join(opts))
 
-    def show_url(self) -> None:
+    def show_url(self, extra_options: Sequence[str] = ()) -> None:
         assert self.extra_url
-        print_and_stop(self.extra_url)
+        parts = [self.extra_url, *(shlex.quote(option) for option in extra_options)]
+        print_and_stop(" ".join(parts))
 
-    def show_env(self) -> None:
+    def show_env(self, extra_options: Sequence[str] = ()) -> None:
         # TODO: inspect the current directory and error if PETSc isn't found
         petsc_dir = f"{os.getcwd()}/petsc"
         petsc_arch = f"arch-firedrake-{self.name}"
@@ -362,8 +377,9 @@ class Arch:
                 value = value.replace(from_, to)
             return value
 
-        env_str = " ".join(f"{name}={replace_vars(value)}" for name, value in env.items())
-        print_and_stop(env_str)
+        env_items = [f"{name}={replace_vars(value)}" for name, value in env.items()]
+        env_items.extend(shlex.quote(option) for option in extra_options)
+        print_and_stop(" ".join(env_items))
 
 
 @dataclasses.dataclass(frozen=True, kw_only=True)
@@ -472,9 +488,9 @@ OFFICIAL_ARCHS: Mapping[ArchKey, Arch] = {
             "--with-superlu_dist",
             "--with-zlib",
         ],
-        cflags=["-O3", "-march=native", "-mtune=native"],
-        cxxflags=["-O3", "-march=native", "-mtune=native"],
-        fflags=["-O3", "-march=native", "-mtune=native"],
+        coptflags=["-O3", "-march=native", "-mtune=native"],
+        cxxoptflags=["-O3", "-march=native", "-mtune=native"],
+        foptflags=["-O3", "-march=native", "-mtune=native"],
         include_dirs=[
             "/usr/include/hypre",
             "/usr/include/superlu",
@@ -533,9 +549,9 @@ OFFICIAL_ARCHS: Mapping[ArchKey, Arch] = {
             "--with-superlu_dist",
             "--with-zlib",
         ],
-        cflags=["-O3", "-march=native", "-mtune=native"],
-        cxxflags=["-O3", "-march=native", "-mtune=native"],
-        fflags=["-O3", "-march=native", "-mtune=native"],
+        coptflags=["-O3", "-march=native", "-mtune=native"],
+        cxxoptflags=["-O3", "-march=native", "-mtune=native"],
+        foptflags=["-O3", "-march=native", "-mtune=native"],
         include_dirs=[
             "/usr/include/superlu",
             "/usr/include/superlu-dist",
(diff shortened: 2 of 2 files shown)


connorjward:
We use `sed` for this: https://github.com/firedrakeproject/firedrake/blob/main/docker/Dockerfile.vanilla#L111

connorjward:
And this is not the intended purpose of the `firedrake-configure` script:

> To improve robustness, firedrake-configure is intentionally kept extremely minimal and simple. This means that if you want to install Firedrake in a non-standard way (for instance with a custom installation of PETSc, HDF5 or MPI) then it is your responsibility to modify the output from firedrake-configure as necessary. This is described in more detail in [Customising Firedrake](https://www.firedrakeproject.org/install.html#customising).

([source](https://www.firedrakeproject.org/install.html#firedrake-configure))

timCorintis:
> We use `sed` for this: https://github.com/firedrakeproject/firedrake/blob/main/docker/Dockerfile.vanilla#L111

Thanks, I missed the the sed in Dockerfile.vanilla, you're right it does solve the PETSc side.
The remaining gap is the wheels. --show-env only prints: `PETSC_DIR=/path PETSC_ARCH=arch-firedrake-default HDF5_MPI=ON`

The remaining issue is the wheels. --show-env exposes no compiler flags, so mpi4py, h5py, petsc4py, etc. are still built with pip's default -march. That means PETSc can be compatible, while another extension still causes SIGILL on older CPUs. The CPU baseline needs to apply to the entire stack, not just PETSc.

connorjward:
> > We use `sed` for this: https://github.com/firedrakeproject/firedrake/blob/main/docker/Dockerfile.vanilla#L111
> 
> Thanks, I missed the the sed in Dockerfile.vanilla, you're right it does solve the PETSc side. The remaining gap is the wheels. --show-env only prints: `PETSC_DIR=/path PETSC_ARCH=arch-firedrake-default HDF5_MPI=ON`
> 
> The remaining issue is the wheels. --show-env exposes no compiler flags, so mpi4py, h5py, petsc4py, etc. are still built with pip's default -march. That means PETSc can be compatible, while another extension still causes SIGILL on older CPUs. The CPU baseline needs to apply to the entire stack, not just PETSc.

I had a long discussion with your colleague on Slack where we discussed an alternative approach.

> --show-env exposes no compiler flags

I don't see why that's an issue. In that case just set CFLAGS etc yourself.

ab-corintis:
> > > We use `sed` for this: https://github.com/firedrakeproject/firedrake/blob/main/docker/Dockerfile.vanilla#L111
> > 
> > 
> > Thanks, I missed the the sed in Dockerfile.vanilla, you're right it does solve the PETSc side. The remaining gap is the wheels. --show-env only prints: `PETSC_DIR=/path PETSC_ARCH=arch-firedrake-default HDF5_MPI=ON`
> > The remaining issue is the wheels. --show-env exposes no compiler flags, so mpi4py, h5py, petsc4py, etc. are still built with pip's default -march. That means PETSc can be compatible, while another extension still causes SIGILL on older CPUs. The CPU baseline needs to apply to the entire stack, not just PETSc.
> 
> I had a long discussion with your colleague on Slack where we discussed an alternative approach.
> 
> > --show-env exposes no compiler flags
> 
> I don't see why that's an issue. In that case just set CFLAGS etc yourself.

Hi @connorjward, we just investigated the other solution we discussed: passing the `--COPTFLAGS`, `--CXXOPTFLAGS`, `--FOPTFLAGS` overrides after a `--` parameter at the end of the `firedrake-configure` arguments.

Unfortunately that has a pretty major issue: the _actual_ `--COPTFLAGS` that `firedrake-configure` outputs [_also_ contains the includes and libraries](https://github.com/firedrakeproject/firedrake/blob/3d533a26d57f27eceda467b2d77f962df3e5e181/scripts/firedrake-configure#L314), _not just_ the optimization values. Meaning that if we pass our version of the `--COPTFLAGS` settings without the libraries, we'd strip those out with the overrides, breaking the build.

E.g.:
```
... --COPTFLAGS='-O3 -march=native -mtune=native -I/usr/include/hdf5/openmpi -I/usr/include/hypre -I/usr/include/superlu -I/usr/include/superlu-dist -L/usr/lib/aarch64-linux-gnu/hdf5/openmpi' ... --COPTFLAGS='-O2 -march=x86-64-v3 --mtune=generic'
```
would end up effectively as just:
```
... --COPTFLAGS='-O2 -march=x86-64-v3 --mtune=generic'
```

As you suggested that the simple env value override would also be acceptable, could we maybe settle on this solution?

connorjward:
> > > > We use `sed` for this: https://github.com/firedrakeproject/firedrake/blob/main/docker/Dockerfile.vanilla#L111
> 
> > > 
> 
> > > 
> 
> > > Thanks, I missed the the sed in Dockerfile.vanilla, you're right it does solve the PETSc side. The remaining gap is the wheels. --show-env only prints: `PETSC_DIR=/path PETSC_ARCH=arch-firedrake-default HDF5_MPI=ON`
> 
> > > The remaining issue is the wheels. --show-env exposes no compiler flags, so mpi4py, h5py, petsc4py, etc. are still built with pip's default -march. That means PETSc can be compatible, while another extension still causes SIGILL on older CPUs. The CPU baseline needs to apply to the entire stack, not just PETSc.
> 
> > 
> 
> > I had a long discussion with your colleague on Slack where we discussed an alternative approach.
> 
> > 
> 
> > > --show-env exposes no compiler flags
> 
> > 
> 
> > I don't see why that's an issue. In that case just set CFLAGS etc yourself.
> 
> 
> 
> Hi @connorjward, we just investigated the other solution we discussed: passing the `--COPTFLAGS`, `--CXXOPTFLAGS`, `--FOPTFLAGS` overrides after a `--` parameter at the end of the `firedrake-configure` arguments.
> 
> 
> 
> Unfortunately that has a pretty major issue: the _actual_ `--COPTFLAGS` that `firedrake-configure` outputs [_also_ contains the includes and libraries](https://github.com/firedrakeproject/firedrake/blob/3d533a26d57f27eceda467b2d77f962df3e5e181/scripts/firedrake-configure#L314), _not just_ the optimization values. Meaning that if we pass our version of the `--COPTFLAGS` settings without the libraries, we'd strip those out with the overrides, breaking the build.
> 
> 
> 
> E.g.:
> 
> ```
> 
> ... --COPTFLAGS='-O3 -march=native -mtune=native -I/usr/include/hdf5/openmpi -I/usr/include/hypre -I/usr/include/superlu -I/usr/include/superlu-dist -L/usr/lib/aarch64-linux-gnu/hdf5/openmpi' ... --COPTFLAGS='-O2 -march=x86-64-v3 --mtune=generic'
> 
> ```
> 
> would end up effectively as just:
> 
> ```
> 
> ... --COPTFLAGS='-O2 -march=x86-64-v3 --mtune=generic'
> 
> ```
> 
> 
> 
> As you suggested that the simple env value override would also be acceptable, could we maybe settle on this solution?

I have thought about this. The answer is that PETSc configure takes both `--CFLAGS` and `--COPTFLAGS`. We just need to split up what is currently only being passed as `--COPTFLAGS` such that one may be overridden and not the other.

ab-corintis:
> I have thought about this. The answer is that PETSc configure takes both `--CFLAGS` and `--COPTFLAGS`. We just need to split up what is currently only being passed as `--COPTFLAGS` such that one may be overridden and not the other.

Are you suggesting passing the includes and libraries as `--CFLAGS` and only the optimization values as `--COPTFLAGS`?
(And similarly split `--CXXOPTFLAGS` to `--CXXFLAGS` and `--CXXOPTFLAGS` and `--FOPTFLAGS` to `--FFLAGS` and `--FOPTFLAGS`?)

It's worth a shot, I suppose... But while we're at it, shouldn't we also move the libraries to be passed with `--LDFLAGS`?

timCorintis:
> I have thought about this. The answer is that PETSc configure takes both `--CFLAGS` and `--COPTFLAGS`. We just need to split up what is currently only being passed as `--COPTFLAGS` such that one may be overridden and not the other.

@connorjward following up on the open question above (I don't think you were tagged): for the flag split, should the `-L…` library dirs go in `--LDFLAGS`, or stay in `--CFLAGS` with the includes? Optimisation (`-O3/-march/-mtune`) stays in `--COPTFLAGS`.

connorjward:
> > I have thought about this. The answer is that PETSc configure takes both `--CFLAGS` and `--COPTFLAGS`. We just need to split up what is currently only being passed as `--COPTFLAGS` such that one may be overridden and not the other.
> 
> @connorjward following up on the open question above (I don't think you were tagged): for the flag split, should the `-L…` library dirs go in `--LDFLAGS`, or stay in `--CFLAGS` with the includes? Optimisation (`-O3/-march/-mtune`) stays in `--COPTFLAGS`.

Sorry I didn't see the previous message. I honestly don't mind.

timCorintis:
Thanks, I'll go with the semantic split then  `-I` to `--CFLAGS`, `-L` to `--LDFLAGS`, optimisation to `--COPTFLAGS`.

scripts/firedrake-configure
```diff
@@ -320,19 +338,21 @@ class Arch:
...
-            f"--COPTFLAGS='{cflags}'",
-            f"--CXXOPTFLAGS='{cxxflags}'",
-            f"--FOPTFLAGS='{fflags}'",
+            *([f"--CFLAGS='{include_dirs}'"] if include_dirs else []),
+            *([f"--CXXFLAGS='{include_dirs}'"] if include_dirs else []),
+            *([f"--FFLAGS='{include_dirs}'"] if include_dirs else []),
+            *([f"--LDFLAGS='{library_dirs}'"] if library_dirs else []),
+            f"--COPTFLAGS='{' '.join(self.cflags)}'",
+            f"--CXXOPTFLAGS='{' '.join(self.cxxflags)}'",
+            f"--FOPTFLAGS='{' '.join(self.fflags)}'",
```
connorjward:
```suggestion
            f"--COPTFLAGS='{' '.join(self.coptflags)}'",
            f"--CXXOPTFLAGS='{' '.join(self.cxxoptflags)}'",
            f"--FOPTFLAGS='{' '.join(self.foptflags)}'",
```
If this could be done throughout it would be nicer.

connorjward: Because now there is a mismatch between `self.cflags` and what goes to `--CFLAGS`

scripts/firedrake-configure
```diff
@@ -112,7 +119,7 @@ Please consider passing '--os unknown' or building manually.""")
                 f"'{arch_key_str}', please install them manually"
             )
     elif args.show_petsc_configure_options:
-        arch.show_petsc_configure_options()
+        arch.show_petsc_configure_options(args.passthrough)
```
connorjward, line 116: For consistency please pass it to all of the commands

scripts/firedrake-configure
```diff
@@ -178,7 +185,6 @@ Please see https://firedrakeproject.org/install for more information."""
             "GPU support in Firedrake is currently very limited."
         ),
     )
-
     cmd_group = parser.add_mutually_exclusive_group(required=True)
```
connorjward, line 183:
```suggestion

    cmd_group = parser.add_mutually_exclusive_group(required=True)
```
I think the extra line helps readability

scripts/firedrake-configure
```diff
@@ -76,6 +77,12 @@ def main():
     gpu_arch = GPUPlatform(args.gpu_arch)
 
     arch_key = (os_, scalar_type, int_type, gpu_arch)
+    if args.passthrough and not args.show_petsc_configure_options:
+        error(
+            "Options after '--' are only valid with "
+            "--show-petsc-configure-options."
+        )
+
```
connorjward:
```suggestion
```

scripts/firedrake-configure
```diff
@@ -211,6 +217,18 @@ Please see https://firedrakeproject.org/install for more information."""
         help="Print out the URL of any package required to enable non-OS repo access for this build",
     )
 
+    parser.add_argument(
```
connorjward, line 215: I don't know this bit of argparse. Does this actually strip the leading `--`?

docs/source/install.rst
```diff
@@ -527,6 +527,44 @@ Python script that PETSc generates. This can be found in
 ``$PETSC_DIR/$PETSC_ARCH/lib/petsc/conf``. Other example scripts can be found in
 ``$PETSC_DIR/config/examples`` directory.
 
+.. _optimisation_flags:
```
connorjward: I would replace this section with a small comment saying that you can add/append options

Put it as a section here: https://www.firedrakeproject.org/install.html#customising-firedrake

connorjward (changes requested):
Thanks. Seems along the right lines now

scripts/firedrake-configure
timCorintis: Yes, argparse treats the first `--` as the "end of options" separator, It's removed, and everything after it is collected into the `passthrough `positional.

Result is: 
```
python3 scripts/firedrake-configure --show-petsc-configure-options  --  --COPTFLAGS="-O3 -march=armv8-a -mtune=generic"  --CXXOPTFLAGS="-O3 -march=armv8-a -mtune=generic" --FOPTFLAGS="-O3 -march=armv8-a -mtune=generic"
PETSC_ARCH=arch-firedrake-default --CFLAGS='-I/usr/include/hypre -I/usr/include/superlu -I/usr/include/superlu-dist' --CXXFLAGS='-I/usr/include/hypre -I/usr/include/superlu -I/usr/include/superlu-dist' --FFLAGS='-I/usr/include/hypre -I/usr/include/superlu -I/usr/include/superlu-dist' --COPTFLAGS='-O3 -march=native -mtune=native' --CXXOPTFLAGS='-O3 -march=native -mtune=native' --FOPTFLAGS='-O3 -march=native -mtune=native' --with-c2html=0 --with-debugging=0 --with-fortran-bindings=0 --with-shared-libraries=1 --with-strict-petscerrorcode --with-bison --with-fftw --with-hdf5-include=/usr/include/hdf5/openmpi --with-hdf5-lib=[/usr/lib/aarch64-linux-gnu/hdf5/openmpi/libhdf5.so,/usr/lib/aarch64-linux-gnu/hdf5/openmpi/libhdf5_hl.so] --with-hwloc --with-metis --with-mumps --with-netcdf --with-pnetcdf --with-ptscotch --with-ptscotch-lib='-lptesmumps -lptscotch -lptscotcherr -lesmumps -lscotch -lscotcherr' --with-ptscotch-include=/usr/include/scotch --with-scalapack-lib=-lscalapack-openmpi --with-suitesparse --with-superlu_dist --with-zlib --download-hypre '--COPTFLAGS=-O3 -march=armv8-a -mtune=generic' '--CXXOPTFLAGS=-O3 -march=armv8-a -mtune=generic' '--FOPTFLAGS=-O3 -march=armv8-a -mtune=generic'
```

connorjward: That has slightly different quoting on the appended options. Is that a problem?

scripts/firedrake-configure
timCorintis: Done

scripts/firedrake-configure
timCorintis: Tested locally, it's fine.

scripts/firedrake-configure
```diff
@@ -211,6 +212,17 @@ Please see https://firedrakeproject.org/install for more information."""
...
+        "passthrough",
+        nargs="*",
+        metavar="-- OPTION[=VALUE] ...",
+        help=(
+            "After a literal '--', extra PETSc configure options appended "
+            "verbatim to the generated options. PETSc applies last-one-wins, so "
+            "these override any generated option."
```
connorjward:
```suggestion
            "After a literal '--', extra options appended verbatim to the generated "
            "options. This can be useful to overwrite problematic default options."
```

scripts/firedrake-configure
```diff
@@ -314,33 +326,35 @@ class Arch:
         "--with-strict-petscerrorcode",
     ]
 
-    def show_system_packages(self) -> None:
+    def show_system_packages(self, extra_options: Sequence[str] = ()) -> None:
```
connorjward, line 327: We're not printing `extra_options`

scripts/firedrake-configure
```diff
@@ -314,33 +326,35 @@ class Arch:
...
+            f"--CXXOPTFLAGS='{' '.join(self.cxxoptflags)}'",
+            f"--FOPTFLAGS='{' '.join(self.foptflags)}'",
             *self.common_petsc_configure_options,
             *self.extra_petsc_configure_options,
+            *(shlex.quote(option) for option in extra_options),
         ]
         print_and_stop(" ".join(opts))
 
-    def show_url(self) -> None:
+    def show_url(self, extra_options: Sequence[str] = ()) -> None:
```
connorjward, line 352: ditto

scripts/firedrake-configure
```diff
@@ -314,33 +326,35 @@ class Arch:
...
         assert self.extra_url
         print_and_stop(self.extra_url)
 
-    def show_env(self) -> None:
+    def show_env(self, extra_options: Sequence[str] = ()) -> None:
```
connorjward, line 357: ditto

connorjward (changes requested):

scripts/firedrake-configure
timCorintis: I'd misunderstood "pass it to all of the commands". Now it's done.

connorjward (dismissed):
Happy now. Thanks!

One very last thing. This should probably go into `release` not `main. Then, if you're also happy, I can merge.

timCorintis:
Thanks for the review! Agreed, release is even better target.

timCorintis:
@connorjward  Small follow-up, I switched the include/library dirs to the append form (`--CFLAGS+=` etc. instead of --`CFLAGS=...`) because setting the flags overwrites PETSc's own detected flags and drops MPICC_SHOW.

The changes now target release.

connorjward (approved):

connorjward:
Thanks!

PR #5254, opened 2026-07-15
Deprecate passing 'appctx' to solve

connorjward:
Instead Python objects can be passed directly through in the options dictionary.

WIP

### DIFF
--- a/firedrake/solving_utils.py
+++ b/firedrake/solving_utils.py
@@ -1,6 +1,7 @@
 from itertools import chain
 
 import numpy
+import petsctools
 import ufl
 
 from pyop2 import op2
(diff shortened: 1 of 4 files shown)


firedrake/solving_utils.py
```diff
@@ -342,6 +332,23 @@ def transfer_manager(self, manager):
...
+    def appctx(self) -> dict:
+        ctx = petsctools.AppContext(self.legacy_appctx_prefix).getAll()
+
+        # A split context will already get the full state.
+        # TODO, a better way of doing this.
+        # Now we don't have a temporary state inside the snes
+        # context we could just require the user to pass in the
+        # full state on the outside.
+        if "state" not in ctx:
+            ctx["state"] = self._x
```
pbrubeck, line 346: I think this could be properly done in PETSc, by making PCFieldsplit more DM-aware. The use of appctx here is a hack.

connorjward: Sure, but that is out of scope here

PR #5256, opened 2026-07-15
Allow firedrake-configure --show-env on unsupported OS

pbrubeck:
# Description
We should not be throwing a hard error for `firedrake-configure --show-env`

### DIFF
--- a/scripts/firedrake-configure
+++ b/scripts/firedrake-configure
@@ -74,8 +74,7 @@ def main():
     gpu_arch = GPUPlatform(args.gpu_arch)
     os_ = OS(args.os) if args.os else detect_os(show_error=not args.show_env)
 
-    if os_ is None:
-        assert args.show_env
+    if os_ is None and args.show_env:
         Arch(name=petsc_arch_name(scalar_type, gpu_arch)).show_env()
 
     arch_key = (os_, scalar_type, gpu_arch)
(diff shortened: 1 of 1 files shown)


scripts/firedrake-configure
```diff
@@ -70,9 +70,13 @@ def main():
         )
         args.os = "unknown"
 
-    os_ = OS(args.os) if args.os else detect_os()
     scalar_type = ScalarType(args.scalar_type)
     gpu_arch = GPUPlatform(args.gpu_arch)
+    os_ = OS(args.os) if args.os else detect_os(show_error=not args.show_env)
+
+    if os_ is None:
+        assert args.show_env
```
pbrubeck:
```suggestion
    if os_ is None and args.show_env:
```

scripts/firedrake-configure
```diff
@@ -255,6 +261,13 @@ class GPUPlatform(enum.Enum):
     CUDA = "cuda"
 
 
+def petsc_arch_name(scalar_type: ScalarType, gpu_arch: GPUPlatform) -> str:
+    name = scalar_type.value
+    if gpu_arch is not GPUPlatform.NO_GPU:
+        name = f"{name}-{gpu_arch.value}"
```
pbrubeck, line 267: I think this is producing 

`PETSC_ARCH=arch-firedrake-default`

But at least for my case I needed

`PETSC_ARCH=arch-linux-c-debug`

pbrubeck: Which suggests that we should do this properly and add unofficial support for older Ubuntu versions

PR #5257, opened 2026-07-15
Add Ubuntu 22.04 community install archs

pbrubeck:
# Description

### DIFF
--- a/scripts/firedrake-configure
+++ b/scripts/firedrake-configure
@@ -427,6 +427,7 @@ UBUNTU_SOURCE_BUILD_PACKAGES = [
 
 UBUNTU_SOURCE_BUILD_PETSC_CONFIGURE_OPTIONS = [
     "--download-bison",
+    "--download-cmake",
     "--download-fftw",
     "--download-hdf5",
     "--download-hwloc",
--- a/scripts/firedrake-configure
+++ b/scripts/firedrake-configure
@@ -215,6 +215,11 @@ def detect_os() -> "OS":
             pass
         else:
             if os_info["NAME"] == "Ubuntu":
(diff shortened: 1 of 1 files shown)


scripts/firedrake-configure
```diff
@@ -403,6 +410,65 @@ class CommunityArch(Arch):
...
+    "--download-pnetcdf",
+    "--download-ptscotch",
+    "--download-scalapack",
+    "--download-suitesparse",
+    "--download-superlu_dist",
+    "--download-zlib",
+]
+
+
+def ubuntu_source_build_arch(
```
connorjward, line 447: This not how we do this any more. We explicitly enumerate things.

scripts/firedrake-configure
```diff
@@ -403,6 +410,65 @@ class CommunityArch(Arch):
...
+    *,
+    name: str,
+    extra_petsc_configure_options: Sequence[str] = (),
+) -> CommunityArch:
+    return CommunityArch(
+        name=name,
+        maintainer="the Firedrake community",
```
connorjward, line 454: I would prefer you to put your own name and contact details. I would rather not maintain this.

connorjward (changes requested):

pbrubeck:
My intention is not promise support for an old OS, but to let users install Firedrake on any OS, and not error by default. Maybe we should just be raising a warning and if they get another error downstream, it should be their problem. What do you recommend @connorjward?

connorjward:
We do support that. Pass `--os unknown`.

pbrubeck:
> We do support that. Pass `--os unknown`.

Okay, then the error should indicate this.

PR #5259, opened 2026-07-16
Support 64-bit integer builds

leo-collins:
Support builds with 64-bit integers. 

- Adds `--int-type` option to `firedrake-configure`. Users can pass `--int-type int64` for a 64-bit integer build
- Update workflows


docs/source/install.rst
```diff
@@ -477,10 +479,9 @@ Customising PETSc
 Since ``firedrake-configure`` only outputs a string of options it is straightforward
 to customise the options that are passed to PETSc ``configure``. You can either:
 
-* Append additional options when ``configure`` is invoked. For example, to
-  build PETSc with support for 64-bit indices you should run::
+* Append additional options when ``configure`` is invoked. For example::
 
-   $ python3 ../firedrake-configure --show-petsc-configure-options | xargs -L1 ./configure --with-64-bit-indices
+   $ python3 ../firedrake-configure --show-petsc-configure-options | xargs -L1 ./configure
```
leo-collins: What example could we put here?

docs/source/install.rst
```diff
@@ -397,11 +397,13 @@ packages are built. The currently supported ARCHs are:
 
 * ``default``: the default installation, suitable for most users
 * ``complex``: an installation where PETSc is configured using complex numbers
+* ``int64``: an installation where PETSc uses 64-bit integer indices
 
-The different configurations can be selected by passing the flag ``--arch`` to
-``firedrake-configure``. For example::
+The scalar and integer configurations can be selected by passing ``--arch`` and
```
leo-collins: Separate to this PR, but should we change `--arch` to `--scalar-type` in the docs?

docs/source/install.rst
connorjward: Do `--download-hpddm`

docs/source/install.rst
connorjward: Definitely, didn't realise I'd missed that

connorjward: I'll do that now.

.github/actions/install/action.yml
```diff
@@ -17,6 +17,10 @@ inputs:
     description: Scalar type ('default' or 'complex')
     type: string
     default: default
+  int_type:
+    description: Integer type ('default' or 'int64')
```
connorjward:
```suggestion
    description: Integer type ('int32' or 'int64')
```

connorjward: I think `default` vs `complex` was an error

.github/workflows/core.yml
```diff
@@ -262,10 +270,18 @@ jobs:
         id: generate
         run: |
           if [ ${{ inputs.test_complex }} == 'true' ]; then
-            matrix='{"scalar_type": "complex", "gpu": "none"}'
+            matrix='{"scalar_type": "complex", "int_type": "default", "gpu": "none"}'
```
```suggestion
            matrix='{"scalar_type": "complex", "int_type": "int32", "gpu": "none"}'
```

scripts/firedrake-configure
```diff
@@ -854,9 +930,62 @@ OFFICIAL_ARCHS: Mapping[ArchKey, Arch] = {
         extra_env_vars={"HDF5_DIR": "/opt/homebrew"},
     ),
 
+    (OS.MACOS_ARM64, ScalarType.DEFAULT, IntType.INT64, GPUPlatform.NO_GPU): Arch(
```
connorjward: We won't be testing this configuration so it shouldn't be an official arch. You're welcome to add it as a `CommunityArch` though. I don't mind us maintaining this one so feel free to put "the Firedrake team" in the maintainer list.

This assumes of course that you've checked this locally.

scripts/firedrake-configure
```diff
@@ -535,9 +547,73 @@ OFFICIAL_ARCHS: Mapping[ArchKey, Arch] = {
...
+            "libsuperlu-dist-dev",
+        ],
+        extra_petsc_configure_options=[
+            "--with-64-bit-indices",
+            "--with-bison",
+            "--with-fftw",
+            "--with-hdf5",
+            "--with-hwloc",
+            "--with-hypre",
+            "--download-metis",
```
leo-collins, line 579: PETSc configure errors if we do a 64-bit integer build while using system metis

connorjward (changes requested):

docs/source/install.rst
connorjward: Actually it makes sense to do in this PR. Can you apply this patch:
```patch
diff --git a/docs/source/install.rst b/docs/source/install.rst
index 01e58dc6d..f65d550e5 100644
--- a/docs/source/install.rst
+++ b/docs/source/install.rst
@@ -413,17 +413,16 @@ Prepared configurations
 -----------------------
 
 ``firedrake-configure`` provides a number of different possible configurations
-(termed 'ARCHs') that specify how PETSc is configured and which external
-packages are built. The currently supported ARCHs are:
+that specify how PETSc is configured and which external packages are built. To
+select a non-default configuration you can pass extra flags to
+``firedrake-configure``. For example, to build Firedrake in complex
+mode you should pass ``--scalar-type complex``::
 
-* ``default``: the default installation, suitable for most users
-* ``complex``: an installation where PETSc is configured using complex numbers
+   $ python3 firedrake-configure --show-petsc-configure-options --scalar-type complex
 
-The different configurations can be selected by passing the flag ``--arch`` to
-``firedrake-configure``. For example::
-
-   $ python3 firedrake-configure --show-petsc-configure-options --arch complex
+To see the full list of available options you should run::
 
+   $ python3 firedrake-configure --help
 
 Optional dependencies
 ---------------------
```

scripts/firedrake-configure
leo-collins: same with superlu dist

scripts/firedrake-configure
leo-collins: Okay, I'll remove it for now and possibly add it in the future when I have time to test it fully

scripts/firedrake-configure
connorjward: Don't worry about dropping problematic packages from the list. We don't have the have to exact same set of packages for each configuration. Obviously try downloading them first, but if that doesn't work it's not necessarily an issue.

.github/actions/install/action.yml
leo-collins: Done

docs/source/install.rst
leo-collins: Done

connorjward (approved):

connorjward:
Thanks!

PR #5260, opened 2026-07-16
RestrictedFunctionSpace: pull back BaseForms

leo-collins:
`restrict=True` was failing for more complex `BaseForm`s. To fix this we define the restricted residual and jacobian correctly. If `P = interpolate(TrialFunction(V_res), V)`, then 
- F_res(u_res) = P^* F(P u_res)
- J_res(u_res) = P^* J(P u_res) P


connorjward (dismissed):
Seems to make sense.

pbrubeck:
What error were you getting? It should be possible still to assemble the original unrestricted F and then interpolate the result into the restricted space. This seems like a workaround

leo-collins:
> What error were you getting?

An error in UFL for incompatible spaces in action.

>  It should be possible still to assemble the original unrestricted F and then interpolate the result into the restricted space.

I don't understand. Where are we assembling F?

pbrubeck:
> I don't understand. Where are we assembling F?

`assemble(interpolate(v_res, F(u->u_res; v))) `

will break into `assemble(F(u->u_res; v))` into a `Cofunction(V.dual())`, and then it will interpolate it to `V_res.dual()`. There is no need to change the internal arguments in `F(u_res; v)`.

My guess is that the error in UFL might be doing the checks wrong.

pbrubeck:
Also another potential source of issues is the fact that `action` with an `interpolate` are eagerly simplifying, maybe that is the problem.

tests/firedrake/regression/test_restricted_function_space.py
```diff
@@ -209,6 +209,35 @@ def test_poisson_inhomogeneous_bcs_high_level_interface(assembled_rhs):
...
+    mesh = UnitSquareMesh(4, 4)
+    V = FunctionSpace(mesh, "CG", 1)
+    Q = FunctionSpace(mesh, "DG", 0)
+
+    u, v = TrialFunction(V), TestFunction(V)
+    q, _ = TrialFunction(Q), TestFunction(Q)
+
+    M = inner(q, v) * dx  # Q x V -> R
+    I = interpolate(u, Q)  # V x Q^* -> R
+    A = action(M, I)  # V x V^* -> R
```
pbrubeck, line 226: Is A still an `Action` or do we eagerly simplify it `interpolate(u, inner(q, v)*dx)`?

pbrubeck: Does UFL complain if you build `interpolate(u2, interpolate(u1, Form))`?

leo-collins: No simplification happens here. UFL only simplifies if the left BaseForm has one argument

leo-collins:
We have `F = action(action(M, I), u) - L` (looking at the test I added). If we `replace(F, {self.u: self.u_restrict})` then we end up with `action(action(M, I), u_restricted)`. `action(M, I)` goes from V -> V^*, and u_restricted lives in V_restricted. Hence the incompatible spaces.

In `compute_bc_lifting` we do `F = ufl_expr.action(J, u)`, where `J=action(M, I)`. However this returns an `Action(Action(M, I), u)` since `J` is a `BaseForm`. If J is a regular `Form`, then `ufl_expr.action(J,u)` eagerly computes the action and returns another `Form`.

pbrubeck:
> We have `F = action(action(M, I), u) - L` (looking at the test I added). If we `replace(F, {self.u: self.u_restrict})` then we end up with `action(action(M, I), u_restricted)`. `action(M, I)` goes from V -> V^*, and u_restricted lives in V_restricted. Hence the incompatible spaces.
> 
> In `compute_bc_lifting` we do `F = ufl_expr.action(J, u)`, where `J=action(M, I)`. However this returns an `Action(Action(M, I), u)` since `J` is a `BaseForm`. If J is a regular `Form`, then `ufl_expr.action(J,u)` eagerly computes the action and returns another `Form`.

Sounds like one mathematically safe fix is to replace `u -> interpolate(u_res, V)`

firedrake/variational_solver.py
```diff
@@ -97,13 +97,13 @@ def __init__(self, F, u, bcs=None, J=None,
             bcs = [bc.reconstruct(V=V_res, indices=bc._indices) for bc in bcs]
             self.u_restrict = Function(V_res)
             v_res, u_res = TestFunction(V_res), TrialFunction(V_res)
+            v_arg, u_arg = self.J.arguments()
             if isinstance(F, Form):
                 F_arg, = F.arguments()
                 self.F = replace(F, {F_arg: v_res, self.u: self.u_restrict})
             else:
-                self.F = interpolate(v_res, replace(F, {self.u: self.u_restrict}))
+                self.F = interpolate(v_res, replace(F, {u_arg: u_res, self.u: self.u_restrict}))
```
```suggestion
                self.F = interpolate(v_res, replace(F, {self.u: interpolate(self.u_restrict, V)}))
```

pbrubeck: There are many ways to "fix" this. Of course my suggestion is far from optimal because we need to generate code and incur extra flops. But I think this is the most robust solution. We can make it efficient if we intercept V_res->V in interpolate and replace the codegen with dof-permutation (i.e. `assign`)

tests/firedrake/regression/test_restricted_function_space.py
```diff
@@ -209,6 +209,35 @@ def test_poisson_inhomogeneous_bcs_high_level_interface(assembled_rhs):
...
+def test_restrict_action():
+    mesh = UnitSquareMesh(4, 4)
+    V = FunctionSpace(mesh, "CG", 1)
+    Q = FunctionSpace(mesh, "DG", 0)
+
+    u, v = TrialFunction(V), TestFunction(V)
+    q, _ = TrialFunction(Q), TestFunction(Q)
+
+    M = inner(q, v) * dx  # Q x V -> R
+    I = interpolate(u, Q)  # V x Q^* -> R
```
pbrubeck, line 222: Could you parametrize your test on whether `M`, `I` get preassembled before going inside the action?

pbrubeck: I suspect the argument replacement fix might fail, as the actual matrix entries will not be permuted in that case

firedrake/variational_solver.py
```diff
@@ -96,18 +96,34 @@ def __init__(self, F, u, bcs=None, J=None,
...
-                self.F = interpolate(v_res, replace(F, {self.u: self.u_restrict}))
+                F_full = replace(F, {self.u: u_full})
+                self.F = ufl_expr.action(Pstar, F_full)
+
+            if isinstance(self.J, Form):
+                v_arg, u_arg = self.J.arguments()
+                self.J = replace(self.J, {v_arg: v_res, u_arg: u_res, self.u: self.u_restrict})
+            else:
+                J_full = replace(self.J, {self.u: u_full})
+                self.J = ufl_expr.action(Pstar, ufl_expr.action(J_full, P))
```
pbrubeck, line 130: Should we turn this into a reusable function? It might be good to generalize to any number of arguments

pbrubeck: All these form transformations could cleanly go through `NonlinearVariationalProblem.rediscretise(..., form_transform=)` if we had the generic transformer

pbrubeck: Or we could just implement `__new__` with `orig_problem.rediscretise(u=u_restrict) ` and handle the form transformations there. There's partial support for F being a baseform, but no support for BaseForm J, Jp, or E. This might enable RestrictedFunctionSpace + EquationBC for free.

PR #5269, opened 2026-07-21
Don't import interpolate from __future__

connorjward:
# Description


leo-collins:
We could probably remove `__future__.py` now. The warning first appeared in the October 2025 release.

leo-collins (approved):

PR #5281, opened 2026-07-27
Fix flaky int64 test

leo-collins:
This test was often returning a convergence rate of 6.695 and consequently failing on int64 builds


connorjward (changes requested):
Fine  but should this go into release?

pbrubeck:
> Fine  but should this go into release?

This might be due to recent changes in FIAT main

https://github.com/firedrakeproject/fiat/pull/256

PR #5285, opened 2026-07-28
Don't pass restricted BCs into assemble

leo-collins:
Don't pass the restricted BCs in to assembly. When `restrict=True`, the boundary DoFs are removed in the restricted function space so passing in these BCs does nothing. This lets us use `restrict=True` for residuals containing things we cannot apply BCs to (e.g. cross-mesh interpolation currently) in the traditional way.

### DIFF
--- a/firedrake/solving_utils.py
+++ b/firedrake/solving_utils.py
@@ -5,6 +5,7 @@ import ufl
 
 from pyop2 import op2
 from firedrake import dmhooks
+from firedrake.bcs import DirichletBC
 from firedrake.function import Function
 from firedrake.cofunction import Cofunction
 from firedrake.matrix import MatrixBase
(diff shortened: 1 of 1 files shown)


pbrubeck:
We still needs the BCs for a RestrictedFunctionSpace, as they are used to set the value of u_restrict at the beginning of the solve. Does your change preserve that?

pbrubeck:
Why is this change required? Please add a description and or tests.

leo-collins:
> We still needs the BCs for a RestrictedFunctionSpace, as they are used to set the value of u_restrict at the beginning of the solve. Does your change preserve that?

I only change the BCs passed to the residual and Jacobian assembly. `problem.bcs` is unchanged.

> Why is this change required? 

Because we don't need to pass these BCs in to assembly. In particular I would like this for my cross-mesh stuff, since BCs for cross-mesh interpolation has not been implemented yet. I will add a description and tests.

connorjward:
I hit this with pyop3. It was quite annoying to have the debug why imposing boundary conditions on a restricted space failed to be a no-op.

More generally I don't understand why it is valid to create a `DirichletBC` object on a restricted function space. I think it just leads to confusing corner cases. For instance what if the `boundary_set` between the BC and function space differ?

firedrake/solving_utils.py
```diff
@@ -254,9 +255,12 @@ def __init__(self, problem,
             # pmat_type == mat_type and Jp_eq_J
             self.Jp = None
 
-        self.bcs_F = tuple(bc.extract_form('F') for bc in problem.bcs)
-        self.bcs_J = tuple(bc.extract_form('J') for bc in problem.bcs)
-        self.bcs_Jp = tuple(bc.extract_form('Jp') for bc in problem.bcs)
+        # Exclude DirichletBCs with nonempty boundary sets
```
dham, line 258: Explain more why we are doing this.

PR #5270, opened 2026-07-21
Redistributed Submesh and MeshHierarchy

pbrubeck:
## Stack

PR 3 of 4: #5460 (`dmplex-hierarchy-maps`) -> #5288 (`adaptive-multigrid`) -> **this PR** (`mesh-redistribution`) -> #5287 (`assign-submesh-restricted`).

## Summary
AI-assisted (Claude Code)

Mesh redistribution is now a `Submesh` of the mesh it was redistributed from, rather than an ad-hoc
`redist` attribute bolted onto the mesh.

- `Submesh(mesh, redistribute=True)` repartitions instead of inheriting the parent's parallel
  distribution, for both `subdomain_id=None` (the whole mesh, as needed by a redistributed
  `MeshHierarchy`) and a proper subdomain (co-dimension 0 only). The child records a
  `submesh_point_sf` pushing the parent's plex points onto its own, which `is_redistributed` asks
  about; `submesh_shares_distribution` tells consumers (assembly, interpolation) whether two related
  meshes still share a distribution.
- Entity orientations are preserved across redistribution: the child inherits the parent's universal
  vertex numbering (`dmcommon.submesh_vertex_numbering`) and quadrilateral/hexahedral cell
  orientations (`dmcommon.submesh_cell_orientations`), so cell closures are ordered identically on
  both sides and RT/BDM/N1curl/RTCF/RTCE data transfers exactly instead of picking up sign/permutation
  flips.
- Assembling or interpolating across a redistributed mesh and its parent raises `NotImplementedError`,
  since entity maps are only defined when the two share a distribution; use `Function.assign` instead.
- `MeshHierarchy` (`redistribute=True` by default) and `refine_marked_elements`/`HierarchyBase.adapt`
  redistribute a refined/adapted mesh via `Submesh` whenever it would otherwise leave empty ranks.
  `firedrake/mg/interface.py` (`prolong`/`restrict`/`inject`) and `firedrake/mg/utils.py`
  (`transfer_mesh`, `set_dm_refine_level`, `set_level`) route transfers through the pre-redistribution
  mesh, then `assign` onto the redistributed one.

Two supporting changes affect all meshes, not only redistributed ones:

- `MeshTopology.cell_closure` takes its universal vertex numbering and its quadrilateral cell
  orientations from two new `cached_property`s, `_universal_vertex_numbering` and
  `_quadrilateral_cell_orientations`, so that a redistributed submesh can inherit its parent's.
[...]

PR #5287, opened 2026-07-28
Unify assign across Submesh, RestrictedFunctionSpace and RestrictedElement

pbrubeck:
## Stack

PR 4 of 4: #5460 (`dmplex-hierarchy-maps`) -> #5288 (`adaptive-multigrid`) -> #5270 (`mesh-redistribution`) -> **this PR** (`assign-submesh-restricted`).

## Summary
AI-assisted (Claude Code)

Unifies how `Function.assign` moves data between non-matching DOF layouts. A `Submesh`, a `RestrictedFunctionSpace` and a `RestrictedElement` all restrict the nodes of a space, and each was handled by its own machinery: entity maps, a `cell_node_map` permutation, and nothing at all respectively. They are now one mechanism — a `PETSc.SF` over the two spaces' `PETSc.Section`s.

- **`RestrictedElement` can be assigned to and from its parent element**, which previously raised outright. Compatibility (`_compatible_elements`) is deduced from the UFL elements and their entity ids alone: the two must restrict a common element, and place the same number of nodes on every entity of the reference cell up to the entities one of them drops.
- **The `cell_node_map` permutation is gone.** Two distinct spaces on one mesh are related by an identity point SF and their sections, exactly as a submesh and its parent are related by the submesh point SF and theirs. `_assign_via_sections` generalises #5270's `_assign_redistributed`, which it replaces.
- **`assign` accepts a subset of the cells**, not just of the nodes (`_node_subset`), so a partial assignment can be expressed directly in terms of a `mesh.cell_subset(...)`.
- **`FacetSplitPC` builds its restriction indices with `assign`** instead of a hand-written C kernel over a `PermutedMap`, keeping the kernel as a fallback for extruded meshes.
- **Combining functions from more than one distinct-but-compatible function space in a single expression now works** (e.g. `g.assign(f_interior + f_facet)` for complementary restrictions of one element): each function is related to the assignee by its own section SF and moved, unweighted, into the assignee's layout before the weighted combination happens there. Previously this raised a confusing tuple-unpack `ValueError`.

## Not implemented

Extruded meshes. The classic `ExtrudedMesh` has only the flat base `DMPlex`, so one base point carries a whole column of entities of which a restriction drops only some; a `Section` counts the nodes on a point without saying which they are. `_compatible_elements` raises `NotImplementedError` there, and `FacetSplitPC` falls back to the kernel.

## Testing

`tests/firedrake/submesh/test_submesh_assign.py` gains tests for an element against its restriction (CG3, RT2 on triangles, Q3 on quadrilaterals), for the multi-space sum above (`test_assign_multiple_source_spaces`, nprocs=[1, 3]), for the remaining genuinely-incompatible pairs, and for the extruded case being refused. They compare against interpolation into the restricted space, which numbers its nodes without reference to the parent, so a permutation of the nodes within an entity cannot pass unnoticed.

Also included are tests composing all three restrictions at once (a proper-subdomain `Submesh`, redistributed or not, a `RestrictedFunctionSpace` over part of the boundary, and a facet-only `RestrictedElement`), and a cell-subset assignment from a redistributed `Submesh`.

Ran `tests/firedrake/regression/test_assign.py` and `tests/firedrake/submesh/test_submesh_assign.py` at nprocs 1, 2, 3 and 4, both split and as whole files.

### DIFF
--- a/firedrake/assign.py
+++ b/firedrake/assign.py
@@ -694,8 +694,8 @@ class Assigner:
             root_V, leaf_V = target_V, source_V
         section_sf, covered_roots = _make_section_sf(point_sf, root_V, leaf_V)
 
-        source_buffer = type(lhs_func)(source_V)
-        target_buffer = type(lhs_func)(target_V)
+        source_buffer = Function(source_V)
+        target_buffer = Function(target_V)
         source_buffer.dat.data_wo_with_halos[...] = source_data
         mtype, _ = _get_mtype(source_buffer.dat)
         source_buffer_data = source_buffer.dat.data_ro_with_halos
--- a/firedrake/assign.py
+++ b/firedrake/assign.py
@@ -172,12 +172,13 @@ def _unrestricted(element):
 def _compatible_elements(target, source):
     """Whether functions in two elements share a node layout.
 
-    Two elements are compatible if they restrict a common element, and place
-    the same number of nodes on every entity of the reference cell up to the
-    entities one of them drops. Their nodes are then the same functionals
-    entity by entity, so the `PETSc.Section` of either function space
-    describes both and data moves between them without reference to the cell
-    node maps.
+    Two elements are compatible under two conditions. They must restrict a
+    common element. They must also place the same number of nodes on every
+    entity of the reference cell, up to the entities one of them drops.
+
+    Their nodes are then the same functionals, entity by entity. The
+    `PETSc.Section` of either function space therefore describes both, and
+    data moves between them without reference to the cell node maps.
 
     Parameters
     ----------
@@ -206,8 +207,8 @@ def _compatible_elements(target, source):
         return False
     if isinstance(target.cell, TensorProductCell):
         # A base mesh point of an extruded mesh carries the nodes of a whole
-        # column of entities, of which a restriction may drop only some, and
-        # the Section counts the nodes on a point without saying which.
+        # column of entities. A restriction may drop only some of them. The
+        # Section counts the nodes on a point without saying which.
         raise NotImplementedError(
             "Assigning between an element and its restriction is not "
             "implemented on extruded meshes"
@@ -216,9 +217,9 @@ def _compatible_elements(target, source):
     source_counts = _entity_dof_counts(source)
     if target_counts.keys() != source_counts.keys():
         return False
-    # An entity either carries the same nodes in both elements, or is
-    # dropped by one of them; a partial overlap has no entity-wise
-    # correspondence and so cannot be expressed by the two Sections.
+    # An entity either carries the same nodes in both elements, or is dropped
+    # by one of them. A partial overlap has no entity-wise correspondence.
+    # The two Sections therefore cannot express it.
     return all(nt == source_counts[entity] or nt == 0 or source_counts[entity] == 0
                for entity, nt in target_counts.items())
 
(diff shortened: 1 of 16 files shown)


tests/firedrake/regression/test_assign.py
```diff
@@ -18,3 +19,41 @@ def test_single_mesh_mixed_assign():
...
+    coords = Function(VectorFunctionSpace(mesh, "CG", 2)).interpolate(SpatialCoordinate(mesh))
+    assert not written[coords.dat.data_ro[:, 0] > 0.5 + 1e-12].any()
+    # Which nodes the subset holds cannot depend on how the mesh is
+    # partitioned, so the count is the same however many ranks there are.
+    nwritten = written[:V.dof_dset.size]
+    assert mesh.comm.allreduce(int(nwritten.sum())) == 91
+
+    # The same subset of a mesh that is also restricted: here the nodes come
+    # from a submesh that does not share the parallel distribution.
+    submesh = Submesh(mesh, subdomain_id=7, redistribute=True)
```
pbrubeck: If the test uses `Submesh` then it must be moved to test_submesh_assign.py

firedrake/assign.py
```diff
@@ -335,30 +547,12 @@ def _assign_single_mesh(self, lhs_func, subset, funcs, operator):
...
-                # Permute source indices into the target ordering.
-                size = target_space.dof_dset.total_size
-                perm = np.empty((size,), dtype=source_map.values.dtype)
-                np.put(perm, values(target_map), values(source_map))
-                if not assign_to_halos:
-                    perm = perm[:target_space.dof_dset.size]
-                return perm[subset_indices]
-
-        func_data = np.array([data_ro(f.dat)[source_indices(f)] for f in funcs])
+        func_data = np.array([data_ro(f.dat)[subset_indices] for f in funcs])
```
pbrubeck, line 672: I don't  think we support the case `Function(V).assign(Function(V_interior) + Function(V_facet))`, I would expect `subset_indices` to depend on `f` as we had before. I think we need to test this case as well

firedrake/preconditioners/facet_split.py
```diff
@@ -243,9 +243,31 @@ def restricted_dofs(celem, felem):
...
     if V.cell_node_map() is W.cell_node_map():
         return numpy.arange(V.dof_dset.layout_vec.getSizes()[0], dtype=PETSc.IntType)
+    if V.extruded:
+        return _get_restriction_indices_extruded(V, W)
+
+    # The restriction of an element keeps the nodes of the entities it does not
+    # drop, so numbering the nodes of V and assigning them into W leaves each
+    # node of W holding the number of the node of V it was taken from.
+    numbering = Function(V)
+    numbering.dat.data_wo_with_halos.reshape(-1)[:] = numpy.arange(V.dof_count)
```
pbrubeck:
```suggestion
    numbering.dat.data_wo_with_halos.flat[...] = numpy.arange(V.dof_count)
```

firedrake/assign.py
```diff
@@ -335,97 +578,180 @@ def _assign_single_mesh(self, lhs_func, subset, funcs, operator):
...
         else:
             root_V, leaf_V = target_V, source_V
         section_sf, covered_roots = _make_section_sf(point_sf, root_V, leaf_V)
 
-        source_buffer = Function(source_V)
-        target_buffer = Function(target_V)
-        func_data = np.array([f.dat.data_ro_with_halos for f in funcs])
-        source_buffer.dat.data_wo_with_halos[...] = self._compute_rvalue(func_data)
+        source_buffer = type(lhs_func)(source_V)
+        target_buffer = type(lhs_func)(target_V)
```
pbrubeck:
```suggestion
        source_buffer = Function(source_V)
        target_buffer = Function(target_V)
```

PR #5288, opened 2026-07-28
Multigrid: skip transfer kernels for preserved adaptive DOFs

pbrubeck:
## Stack

PR 2 of 4: #5460 (`dmplex-hierarchy-maps`) -> **this PR** (`adaptive-multigrid`) -> #5270 (`mesh-redistribution`) -> #5287 (`assign-submesh-restricted`).

## Description

AI-assisted (Codex)

Adaptive refinement changes only part of a mesh. The untouched part already has the same points and degrees of freedom on the coarse and fine meshes. The transfer code used to evaluate those values again.

This PR uses the DMPlex information from #5460 to find the preserved degrees of freedom. It skips the transfer kernel for them.

During prolongation, it copies each preserved coarse value to its fine value. During restriction, it adds each preserved fine value to the matching coarse value. Restriction adds values because the same coarse basis function can also receive contributions from cells that refinement changed.

The existing transfer path remains in use when the spaces do not match, the meshes are extruded, the levels are not consecutive, or refinement changed every cell.

## Main changes

- Build a PETSc SF that pairs preserved coarse and fine nodes.
- Run prolongation and restriction kernels only on fine nodes that were not preserved.
- Copy preserved nodes during prolongation and reduce them during restriction.
- Keep the existing behavior for mixed spaces, Real spaces, and injection.

## Tests

- Test prolongation, restriction, and injection for DG0 and CG1, CG2, and CG3 spaces.
- Use CG3 to check that nodes on an edge keep the correct order.
- Check a hierarchy in which some cells are refined twice, so the fine DMPlex was built through an intermediate level.
- Cover Firedrake and Netgen meshes in two and three dimensions, with the existing serial and parallel test cases.


firedrake/mg/interface.py
```diff
@@ -200,7 +205,12 @@ def restrict(fine_dual, coarse_dual):
         for d in [coarse_coords]:
             d.dat.global_to_local_begin(op2.READ)
             d.dat.global_to_local_end(op2.READ)
-        op2.par_loop(kernel, fine_dual.node_set, *kernel_args)
+        # Restriction transposes prolongation, so it skips the same fine nodes
+        # and hands the coarse nodes they pair with their values whole.
+        node_subset = utils.transfer_node_subset(Vc, Vf)
```
pbrubeck, line 193: return `Vf.node_set` here instead of `None`

firedrake/mg/utils.py
```diff
@@ -155,6 +172,205 @@ def coarse_cell_to_fine_node_map(Vc, Vf):
...
+        kernel accumulated from the remaining fine nodes.
+
+    """
+    from firedrake.halo import _get_mtype
+
+    coarse_V = coarse_dual.function_space()
+    section_sf = preserved_node_sf(coarse_V, fine_dual.function_space())
+    if section_sf is None:
+        return
+    buffer = type(coarse_dual)(coarse_V)
```
pbrubeck: Fix this anti-pattern throughout
```suggestion
    buffer = Function(coarse_V)
```

firedrake/mg/interface.py
pbrubeck: Done — transfer_node_subset now returns Vf.node_set directly, and both par_loop call sites dropped their own fallback.

firedrake/mg/utils.py
pbrubeck: Done — Function(coarse_V) it is, matching how interface.py already uses Function for these intermediate dual buffers.

.github/actions/install/action.yml
```diff
@@ -163,6 +163,14 @@ runs:
...
         pip list
 
+    - name: "DROP BEFORE MERGE: install FIAT from firedrakeproject/fiat#267"
+      shell: bash
+      run: |
+        . venv/bin/activate
+        pip install --verbose --no-build-isolation --no-deps --force-reinstall \
+          git+https://github.com/firedrakeproject/fiat.git@pbrubeck/restricted-entity-permutations
+        pip list | grep -i fiat
+
```
```suggestion
```

tests/firedrake/multigrid/test_adaptive_multigrid.py
```diff
@@ -329,100 +331,69 @@ def test_adapt_before_uniform_refinement(coarse_mesh, refine):
...
-    u_fine = Function(V_fine)
-    xc, *_ = SpatialCoordinate(V_coarse.mesh())
-
-    u_coarse.interpolate(xc)
-    prolong(u_coarse, u_fine)
-
-    rf = assemble(conj(TestFunction(V_fine)) * dx)
-    rc = Cofunction(V_coarse.dual())
-    restrict(rf, rc)
+@pytest.mark.parametrize("family, degree", [("DG", 0), ("CG", 1), ("CG", 2), ("CG", 3)])
```
pbrubeck, line 434: DG injection should be marked with skipcomplexnoslate

firedrake/cython/mgimpl.pyx
```diff
@@ -339,6 +339,87 @@ def adaptive_parent_child_cell_maps(PETSc.DM coarse_dm,
     return np.asarray(coarse_to_fine), np.asarray(fine_to_coarse)
 
 
+@cython.boundscheck(False)
+@cython.wraparound(False)
+def preserved_points(PETSc.DM coarse_dm,
+                     PETSc.Section coarse_cell_numbering,
+                     PETSc.DM fine_dm,
+                     PETSc.Section fine_cell_numbering,
+                     PetscInt nfine,
```
pbrubeck: We should grab this from the `fine_dm` using `num_owned_cells`

firedrake/cython/mgimpl.pyx
```diff
@@ -339,6 +339,87 @@ def adaptive_parent_child_cell_maps(PETSc.DM coarse_dm,
...
+    cdef:
+        PetscInt ncoarse, max_children, c, i, off, child
+        PetscInt cStart, cEnd, pStart, pEnd, coarse_size, fine_size
+        PetscInt *coarse_closure = NULL
+        PetscInt *fine_closure = NULL
+        PetscInt[::1] coarse_point, fine_point, fine_to_coarse
+        PetscInt[:, ::1] coarse_to_fine
+
+    coarse_to_fine = coarse_to_fine_cells
+    ncoarse = coarse_to_fine.shape[0]
```
pbrubeck: assert that this matches `num_owned_cells(coarse_dm)`

firedrake/mg/interface.py
```diff
@@ -102,7 +102,11 @@ def prolong(coarse, fine):
         for d in [coarse, coarse_coords]:
             d.dat.global_to_local_begin(op2.READ)
             d.dat.global_to_local_end(op2.READ)
-        op2.par_loop(kernel, fine.node_set, *kernel_args)
+        # An adaptive refinement leaves most of the mesh alone, and the nodes
+        # it preserves are copied rather than evaluated.
+        node_subset = utils.transfer_node_subset(Vc, Vf)
+        op2.par_loop(kernel, node_subset, *kernel_args)
+        utils.prolong_preserved_nodes(coarse, fine)
```
pbrubeck: This is the main part of the PR

firedrake/mg/utils.py
```diff
@@ -155,6 +172,205 @@ def coarse_cell_to_fine_node_map(Vc, Vf):
...
+        kernel accumulated from the remaining fine nodes.
+
+    """
+    from firedrake.halo import _get_mtype
+
+    coarse_V = coarse_dual.function_space()
+    section_sf = preserved_node_sf(coarse_V, fine_dual.function_space())
+    if section_sf is None:
+        return
+    buffer = firedrake.Function(coarse_V)
```
pbrubeck, line 428: We should be able to get rid of this buffer by zero-ing the node subset of target

.github/actions/install/action.yml
pbrubeck: Done — firedrakeproject/fiat#267 is merged, dropped the temporary FIAT install step.

firedrake/cython/mgimpl.pyx
pbrubeck: Done — `preserved_points` now computes `nfine` internally via `num_owned_cells(fine_dm)` instead of taking it as a parameter; updated the sole caller (`_preserved_point_sf` in mg/utils.py) to drop the argument.

firedrake/cython/mgimpl.pyx
pbrubeck: Done — `ncoarse` is now sourced from `num_owned_cells(coarse_dm)` directly; the array shape is only asserted against it, not used as the source of truth.

firedrake/mg/utils.py
pbrubeck: > Investigated this and I don't think it's safe as stated. `local_to_global_begin/end` (firedrake/halo.py) does `sf.reduceBegin(mtype, dat._data, dat._data, SUM)` — an in-place reduce where every rank's ghost slot still holds its own unconsumed contribution from the transfer kernel's own par_loop reduction (leaves aren't cleared by a reduce). Calling it a second time directly on `coarse_dual` would re-sum that leftover ghost data onto every partition-boundary-shared coarse DOF, not just the ones this SF touches — silent double counting anywhere the space has shared DOFs across ranks. The separate zero-initialized `buffer` is what keeps that safe (nothing to double-count, since only our new SF writes are nonzero in it), so I'm keeping it as is.

firedrake/mg/utils.py
```diff
@@ -138,6 +140,21 @@ def coarse_cell_to_fine_node_map(Vc, Vf):
         owned_coarse_to_fine = coarse_to_fine[:iterset.size, :]
         valid = owned_coarse_to_fine >= 0
         values[valid, :] = Vf.cell_node_map().values[owned_coarse_to_fine[valid], :]
+        # Under adaptive refinement coarse cells have varying numbers of fine
+        # descendants, so the rows are right-padded up to the busiest coarse
+        # cell's count. op2.Map cannot hold the -1 that marks a slot as empty,
+        # and the macro-cell kernel reading through this map integrates over
+        # every slot alike. Point all of an empty slot's nodes at a single one
```
pbrubeck: I'm not sure how accurate this is.

tests/firedrake/multigrid/test_adaptive_multigrid.py
```diff
@@ -329,102 +332,77 @@ def test_adapt_before_uniform_refinement(coarse_mesh, refine):
...
+    if degree == 0:
+        # The only expression a DG0 space holds on every mesh of a hierarchy
+        # alike is one that is constant on each of the coarsest cells.
+        return conditional(ge(x[0], 0), 1, 0)
+    return sum(xi ** degree for xi in x)
+
+
+def _copied_nodes(mh, V):
+    """Count the nodes of ``V`` that the transfers copy rather than evaluate."""
+    from firedrake.mg.utils import transfer_node_subset
```
connorjward: top level import

firedrake/mg/utils.py
```diff
@@ -155,6 +172,205 @@ def coarse_cell_to_fine_node_map(Vc, Vf):
...
+        return cache[key]
+    except KeyError:
+        _, preserved, _ = section_sf.getGraph()
+        nodes = numpy.setdiff1d(numpy.arange(Vf.node_set.size, dtype=IntType),
+                                preserved)
+        return cache.setdefault(key, op2.Subset(Vf.node_set, nodes))
+
+
+def prolong_preserved_nodes(coarse, fine):
+    """Give the nodes an adaptive refinement preserved their coarse values.
```
connorjward: Very unclear

firedrake/mg/interface.py
```diff
@@ -102,7 +102,11 @@ def prolong(coarse, fine):
         for d in [coarse, coarse_coords]:
             d.dat.global_to_local_begin(op2.READ)
             d.dat.global_to_local_end(op2.READ)
-        op2.par_loop(kernel, fine.node_set, *kernel_args)
+        # An adaptive refinement leaves most of the mesh alone, and the nodes
+        # it preserves are copied rather than evaluated.
+        node_subset = utils.transfer_node_subset(Vc, Vf)
```
connorjward, line 107:
```suggestion
        changed_node_subset = utils.transfer_node_subset(Vc, Vf)
```
clearer?

firedrake/mg/utils.py
```diff
@@ -138,6 +140,21 @@ def coarse_cell_to_fine_node_map(Vc, Vf):
...
         values[valid, :] = Vf.cell_node_map().values[owned_coarse_to_fine[valid], :]
+        # Under adaptive refinement coarse cells have varying numbers of fine
+        # descendants, so the rows are right-padded up to the busiest coarse
+        # cell's count. op2.Map cannot hold the -1 that marks a slot as empty,
+        # and the macro-cell kernel reading through this map integrates over
+        # every slot alike. Point all of an empty slot's nodes at a single one
+        # of the cell's own children's nodes: the slot then describes a
+        # degenerate cell sitting on the patch it belongs to, whose Jacobian
+        # determinant, and so whose contribution, vanishes.
+        if not valid.all():
```
connorjward: I've just had to fix this in pyop3. Please remove this branch because it causes rank divergence. All ranks should just do it

firedrake/mg/utils.py
connorjward: It is certainly confusing.

> macro-cell kernel

Huh?

> empty slot's nodes

unclear

firedrake/mg/utils.py
```diff
@@ -138,6 +140,21 @@ def coarse_cell_to_fine_node_map(Vc, Vf):
         owned_coarse_to_fine = coarse_to_fine[:iterset.size, :]
         valid = owned_coarse_to_fine >= 0
         values[valid, :] = Vf.cell_node_map().values[owned_coarse_to_fine[valid], :]
+        # Under adaptive refinement coarse cells have varying numbers of fine
+        # descendants, so the rows are right-padded up to the busiest coarse
+        # cell's count. op2.Map cannot hold the -1 that marks a slot as empty,
```
connorjward: You know pyop3 makes this really easy. Ragged maps exist.

firedrake/mg/utils.py
```diff
@@ -155,6 +172,205 @@ def coarse_cell_to_fine_node_map(Vc, Vf):
                                              offset=offset))
 
 
+def _preserved_point_sf(coarse_mesh, fine_mesh, coarse_to_fine):
+    """Create the `PETSc.SF` pairing the points an adaptive refinement copied.
```
connorjward:
```suggestion
    """Create the `PETSc.SF` pairing the points an adaptive refinement did not refine.
```

firedrake/mg/utils.py
```diff
@@ -155,6 +172,205 @@ def coarse_cell_to_fine_node_map(Vc, Vf):
...
+
+    """
+    if Vc.ufl_element() != Vf.ufl_element() or Vc.boundary_set != Vf.boundary_set:
+        # Only a space and its counterpart on the refined mesh lay their nodes
+        # out the same way on a cell the refinement left alone.
+        return None
+    if Vc.extruded or Vf.extruded:
+        # The plex of an extruded mesh is the flat base one, whose points
+        # carry a whole column of nodes that the Sections cannot tell apart.
+        return None
```
connorjward, line 299: Returning `None` is fine if we are just aborting because we can't do the copy, but that is not made at all clear.

connorjward, line 299: Why can't we do extruded?

firedrake/mg/utils.py
```diff
@@ -155,6 +172,205 @@ def coarse_cell_to_fine_node_map(Vc, Vf):
...
+    Parameters
+    ----------
+    fine_dual : firedrake.cofunction.Cofunction
+        The cofunction on the refined mesh.
+    coarse_dual : firedrake.cofunction.Cofunction
+        The cofunction on the coarse mesh, already holding what the transfer
+        kernel accumulated from the remaining fine nodes.
+
+    """
+    from firedrake.halo import _get_mtype
```
connorjward: top level import

firedrake/mg/utils.py
```diff
@@ -155,6 +172,205 @@ def coarse_cell_to_fine_node_map(Vc, Vf):
...
+    # A coarse node that a fine one was copied from can be a ghost, but only
+    # owned fine nodes are written, exactly as the transfer kernel writes them.
+    source = coarse.dat.data_ro_with_halos
+    target = fine.dat.data_wo
+    section_sf.bcastBegin(mtype, source, target, MPI.REPLACE)
+    section_sf.bcastEnd(mtype, source, target, MPI.REPLACE)
+
+
+def restrict_preserved_nodes(fine_dual, coarse_dual):
+    """Add what the nodes an adaptive refinement preserved contribute to the coarse dual.
```
connorjward: I don't understand why we're adding instead of copying

firedrake/cython/mgimpl.pyx
```diff
@@ -339,6 +339,87 @@ def adaptive_parent_child_cell_maps(PETSc.DM coarse_dm,
...
+                     np.ndarray coarse_to_fine_cells):
+    """Pair the points an adaptive refinement left alone with their coarse originals.
+
+    A coarse cell that the refinement did not touch is copied into the fine
+    mesh, so its whole closure is copied with it: the transform preserves the
+    cone of every point it does not refine, and hence the plex closure of the
+    copied cell entry by entry. Such a cell has exactly one child, which is
+    what the right-padding of ``coarse_to_fine_cells`` with -1 identifies.
+
+    :arg coarse_dm: the coarse mesh DMPlex.
```
connorjward: numpydoc!

firedrake/cython/mgimpl.pyx
```diff
@@ -339,6 +339,87 @@ def adaptive_parent_child_cell_maps(PETSc.DM coarse_dm,
...
+        if child < 0 or (max_children > 1 and coarse_to_fine[c, 1] >= 0):
+            continue
+        if coarse_point[c] < 0 or fine_point[child] < 0:
+            continue
+        CHKERR(DMPlexGetTransitiveClosure(coarse_dm.dm, coarse_point[c], PETSC_TRUE,
+                                          &coarse_size, &coarse_closure))
+        CHKERR(DMPlexGetTransitiveClosure(fine_dm.dm, fine_point[child], PETSC_TRUE,
+                                          &fine_size, &fine_closure))
+        # A cell with one child that the transform nonetheless changed would
+        # have a closure of its own shape; leave it to the transfer kernel.
```
connorjward: I don't understand this

connorjward (changes requested):

firedrake/mg/utils.py
pbrubeck: Context: DG injection integrates on a macro-cell formed by all fine children, which for an adaptive mesh are not uniform, so we need a ragged array with empty slots

pbrubeck: I am asking about op2.Map not being able to hold a negative value

firedrake/cython/mgimpl.pyx
pbrubeck: I think the rule to detect preserved points is more subtle. A cell might be left unrefined but still changed by the transform?

firedrake/mg/utils.py
pbrubeck: What do you mean? Have you fixed this bug or a different one?

firedrake/mg/utils.py
pbrubeck: I think Give -> Return makes this more clear

pbrubeck: But sure, it is still not quite clear. I'll enforce proper English so this does not happen again

firedrake/mg/utils.py
pbrubeck: I think we should be able to copy because the operator is the identity on these nodes.

firedrake/cython/mgimpl.pyx
```diff
@@ -352,6 +352,99 @@ def adaptive_parent_child_cell_maps(PETSc.DM coarse_dm,
     return np.asarray(coarse_to_fine), np.asarray(fine_to_coarse)
 
 
+@cython.boundscheck(False)
+@cython.wraparound(False)
+def preserved_points(PETSc.DM coarse_dm,
```
pbrubeck, line 405: We should be able to simplify by taking the complement of `createSplitCellLabel()`

Depends on https://gitlab.com/petsc/petsc/-/merge_requests/9481

PR #5294, opened 2026-07-29
Infrastructure for moving VertexOnlyMeshes

achanbour:
This PR contains the bulk of the work for supporting mutable VOMs.

Broadly, this introduces 3 new features:

1. A `VertexOnlyMeshMutator` that handles two update regimes: one that updates the local (reference-space) state of the VOM and one that does a full (topological) rebuild of the VOM
2. A conditional caching mechanism wherein `FunctionSpace` properties get recomputed when the topology version number of their underlying mesh changes
3. A `Function` migration mechanism (both lazy and eager) which ensures that Function values refresh accordingly with the VOM's re-ordering.

While I have tried to make 2 and 3 as general as possible, I have only written and tested them to work for DG0-style Function Spaces defined on VOMs.


firedrake/function.py
```diff
@@ -277,6 +277,16 @@ def __init__(self, function_space, val=None, name=None, dtype=ScalarType,
         # LRU cache for expressions assembled onto this function
         self._expression_cache = cachetools.LRUCache(maxsize=50)
 
+        self._mesh_topo = self._function_space.topological.mesh()  # the MeshTopology object
+        self._mesh_geom = self._function_space.mesh()  # the MeshGeometry object
```
leo-collins: These should be `self._mesh_topology` and `self._mesh_geometry`.

firedrake/function.py
```diff
@@ -299,8 +309,74 @@ def copy(self, deepcopy=False):
...
+            raise TopologyVersionMismatchError(
+                "The mesh topology has changed since this Function was created, \
+                and migration is currently only supported for Functions defined on VertexOnlyMeshes. \
+                Please re-create this Function on the updated mesh."
+            )
+
+        # Get the latest one-step SF point mapping (indexed by current mesh version)
+        latest_topology_step_sf = self._mesh_topo._topology_step_sfs.get(current_version, None)
+        if latest_topology_step_sf is None:
+            raise RuntimeError(
```
leo-collins: Custom exception?

firedrake/function.py
```diff
@@ -299,8 +309,74 @@ def copy(self, deepcopy=False):
...
+        # Then check if we need to chain multiple one-step SFs (this happens when the Function was created on a VOM topology
+        # that's more than one version behind).
+        if current_version - self._mesh_topology_version > 1:
+            # Compose multiple one-step SFs
+            chained_sf = latest_topology_step_sf  # starts from latest V -> V-1
+            # Iterate backwards through the intermediate versions
+            for v in range(current_version-1, self._mesh_topology_version, -1):
+                step_sf = self._mesh_topo._topology_step_sfs.get(v, None)  # maps V-1 -> V-2
+                if step_sf is None:
+                    raise RuntimeError(
```
leo-collins: Again, we should have something like `FunctionMigrationError`

firedrake/function.py
```diff
@@ -870,3 +946,29 @@ def make_c_evaluate(function, c_name="evaluate", ldargs=None, tolerance=None):
         comm=function.comm
     )
     return getattr(dll, c_name)
+
+
+def migrate_dg0_dat(old_cfunc, FS_topo, step_sf):
```
leo-collins: type hints

firedrake/function.py
```diff
@@ -870,3 +946,29 @@ def make_c_evaluate(function, c_name="evaluate", ldargs=None, tolerance=None):
...
+    """Migrate a DG0 CoordinatelessFunction across a topology change via the step SF.
```
leo-collins: Should probably be a check that the function is a DG0 function

firedrake/functionspaceimpl.py
```diff
@@ -630,17 +643,17 @@ def __ne__(self, other):
         return not self.__eq__(other)
 
     def __hash__(self):
-        return hash((self.mesh(), self.dof_dset, self.ufl_element()))
+        # Given that dof_dset is now re-created when the FS mutates, we remove it from the hash
```
leo-collins: Don't make comments about old code

firedrake/vertexonly_mutation.py
```diff
@@ -0,0 +1,330 @@
+import numpy as np
+from firedrake.utils import IntType
+import firedrake.cython.dmcommon as dmcommon
+from pyop2.mpi import MPI
+
+
+class VertexOnlyMeshMutator:
+    """Mutate a VertexOnlyMesh in place as particles move through its parent mesh."""
+    def __init__(self, vom):
```
leo-collins: type hints

firedrake/exceptions.py
```diff
@@ -73,3 +73,8 @@ def __init__(self, domain, point):
 
     def __str__(self):
         return f"Domain {self.domain} does not contain point {self.point}"
+
+
+class TopologyVersionMismatchError(ValueError):
+    """Raised when a Function's mesh topology attribute does not match the topology version attribute stored on the mesh object."""
+    pass
```
connorjward:
```suggestion
```

firedrake/vertexonly_mutation.py
```diff
@@ -0,0 +1,330 @@
...
+        self.vom = vom
+        self.parent_mesh = self.vom._parent_mesh
+
+        if self.parent_mesh.extruded:
+            raise NotImplementedError(
+                "The VertexOnlyMeshMutator does not yet support VertexOnlyMeshes immersed in an extruded parent mesh."
+            )
+
+    def commit_reference_state(self, new_parent_cells, new_refcoords):
```
leo-collins: type hints and numpy docstring

firedrake/mesh.py
```diff
@@ -2054,6 +2069,10 @@ def __init__(self, swarm, parentmesh, name, reorder, input_ordering_swarm=None,
         self.input_ordering_swarm = input_ordering_swarm
         self._parent_mesh = parentmesh
 
+        # Registry of live functions defined on the VOM
+        self._live_functions = weakref.WeakValueDictionary()
+        self._function_counter = itertools.count()
```
connorjward, line 2101: A bit more commentary is helpful here. For example I don't know what the keys of `_live_functions` is. And generally anytime you do things with weakref it's helpful to explain why that's the right approach.

firedrake/vertexonly_mutation.py
```diff
@@ -0,0 +1,330 @@
...
+        topology = self.vom.topology
+
+        n_owned = self.vom.cell_set.size
+        vom_to_swarm = self.vom.cell_closure[:n_owned, -1]  # VOM cell ID -> swarm point ID
+
+        new_parent_cells = np.asarray(new_parent_cells, dtype=int).ravel()
+        new_refcoords = np.asarray(new_refcoords, dtype=float)
+
+        # Update Firedrake parent cell numbers
+        arr = swarm.getField("parentcellnum")
```
leo-collins: using `arr` for several different fields throughout this function is confusing

firedrake/mesh.py
```diff
@@ -2343,6 +2362,33 @@ def input_ordering_without_halos_sf(self):
         # cells first; self.cell_set.size is the number of rank-local non-halo cells.
         return self.input_ordering_sf.createEmbeddedLeafSF(np.arange(self.cell_set.size, dtype=IntType))
 
+    def _migrate_functions(self, functions_to_migrate=()):
+        """
+        Migrate Function(s) defined on this topology to the current version.
+
+        If `functions_to_migrate` is empty, migrate all live Functions registered on the VOM.
```
connorjward: Why might we want to forcibly migrate all functions? I don't know why we want to bother having an eager approach.

firedrake/mesh.py
```diff
@@ -2343,6 +2362,33 @@ def input_ordering_without_halos_sf(self):
...
+        """
+        import gc
```
connorjward: Can be a top level import

firedrake/vertexonly_mutation.py
```diff
@@ -0,0 +1,330 @@
...
+
+        arr[vom_to_swarm, :] = plex_ids
+        swarm.restoreField(cell_id_name)
+
+        # Update reference coordinates
+        arr = swarm.getField("refcoord")
+        arr[vom_to_swarm, :] = new_refcoords
+        swarm.restoreField("refcoord")
+
+        # Invalidate cached properties that depend on parent cell ownership
```
leo-collins: is there a way we could use `cached_property_until` for this?

firedrake/vertexonly_mutation.py
```diff
@@ -0,0 +1,330 @@
...
+        and refreshes the VOM's topology, numbering, particle ids, coordinate Functions, and
+        topology-dependent caches.
+
+        Args:
+            absorbed_vom_indices: Optional local VOM indices of particles to remove,
+                in current VOM ordering. Raises ``EmptyVOMError`` if no particles
+                remain globally.
+        """
+        from firedrake.mesh import VertexOnlyMeshTopology
+        from firedrake.function import migrate_dg0_dat
```
leo-collins: Can these go at the top of the file?

firedrake/vertexonly_mutation.py
```diff
@@ -0,0 +1,330 @@
...
+            topology.__dict__["cell_parent_cell_list"][:n_owned] = new_parent_cells
+
+        # Write the update reference coordinates
+        # NOTE: Since new_ref_coords is already in VOM ordering and assuming the VOM has not been reodered when this function is called,
+        # we can safely write into the reference coordinates array.
+        ref_coords_func = self.vom.reference_coordinates
+        ref_coords_func.dat.data[:] = new_refcoords
+
+    def rebuild_vom(self, absorbed_vom_indices=None):
+        """Rebuild the VertexOnlyMesh using the state already stored on the VOM.
```
leo-collins: numpydoc and type hints

firedrake/vertexonly_mutation.py
```diff
@@ -0,0 +1,330 @@
...
+        topology = self.vom.topology
+        gdim = self.vom.geometric_dimension
+
+        parent_plex = self.parent_mesh.topology_dm
+        parent_tdim = self.parent_mesh.topological_dimension
+
+        n_local = topology.cell_set.size  # number of particles owned by this rank
+
+        # Read data from the current VOM
+        coords_local = self.vom.coordinates.dat.data_ro.copy()
```
leo-collins, line 133: This function is enormous. Is there a way we could break it up?

firedrake/mesh.py
```diff
@@ -2343,6 +2362,33 @@ def input_ordering_without_halos_sf(self):
...
+
+        # TODO: do parallel set intersection of the live function keys across all ranks
+        # and migrate only ones in the intersection
+
+        # NOTE: This doesn't collect reference-cycle garbage identically on all ranks
+        gc.collect()
```
connorjward: I would advocate for using [this function](https://github.com/firedrakeproject/firedrake/blob/connorjward/pyop3/pyop3/cache.py#L47) (just in pyop3 currently but could come in early) instead here. Just temporarily disable the GC for the duration of this function.

firedrake/mesh.py
```diff
@@ -2343,6 +2362,33 @@ def input_ordering_without_halos_sf(self):
...
+
+        if len(functions_to_migrate) > 0:
+            for f in functions_to_migrate:
+                f._match_mesh_topology_version()
+                return
+
+        keys = [k for k in sorted(self._live_functions.keys())
```
connorjward: My guess is that this is already sorted?

firedrake/mesh.py
connorjward: But if you don't do things eagerly then this might not be an issue, by definition if you are explicitly asking for a function to be migrated then you must hold another ref to it.

firedrake/vertexonly_mutation.py
```diff
@@ -0,0 +1,330 @@
...
+        global_idxs_local = np.arange(offset, offset + n_local, dtype=IntType)  # indices for this rank's owned particles in global arrays
+        input_idxs_local = np.arange(n_local, dtype=IntType)  # local indices for this rank's owned particles in local arrays
+
+        input_ranks_local = np.full(n_local, comm.rank)  # input ranks for this rank's owned particles
+        # NOTE: we assume that the current VOM contains only particles owned by this rank
+        # after each rebuild, ``swarm.migrate(remove_sent_points=True)`` hands particles to their
+        # new owning ranks and removes sent copies, and the VOM is asserted to have no ghost particles.
+
+        # Remove particles from the VOM
+        absorbed = np.array([] if absorbed_vom_indices is None else absorbed_vom_indices, dtype=int)
```
leo-collins: Should this be `dtype=IntType`?

tests/firedrake/vertexonly/test_vertex_only_fs.py
```diff
@@ -375,3 +377,356 @@ def test_tensorfs_permutation(tensorfs_and_expr):
     petsc_mat = assemble(interpolate(TrialFunction(V), W), mat_type="aij")
     f_in_W_petsc = assemble(petsc_mat @ f)
     assert np.allclose(f_in_W.dat.data_ro, f_in_W_petsc.dat.data_ro)
+
+
+# Function Space Mutation Tests
```
connorjward: I'd put these in a separate test file

firedrake/functionspaceimpl.py
```diff
@@ -654,11 +667,11 @@ def _dm(self):
         dmhooks.set_function_space(dm, self)
         return dm
 
-    @cached_property
+    @cached_property_until(lambda self: self._mesh._topology_version)
```
connorjward: Since you use this pattern a lot why not implement
```py
topology_cached_property = cached_property_until(...)
...
@topology_cached_property
```

firedrake/vertexonly_mutation.py
```diff
@@ -0,0 +1,330 @@
...
+                                                        (topology.num_vertices(), parent_tdim), reference_coord=True)
+            self.vom.reference_coordinates._data.dat = ref_coords_fs.make_dat(val=ref_coords_data, name=self.vom.reference_coordinates.name())  # returns a new op2.Dat
+        else:
+            # This should have been already set to None when the VOM was first constructed
+            self.vom.reference_coordinates = None
+
+    def _invalidate_topology_properties(self):
+        # Delete cached properties so they get recomputed on next access using the updated swarm fields
+        topology = self.vom.topology
+        for name in (
```
leo-collins: Again, this would be really nice if we could use `cached_property_until`.

firedrake/functionspaceimpl.py
```diff
@@ -1004,7 +1015,10 @@ def set_shared_data(self):
         self.offset_quotient = sdata.offset_quotient
         self.cell_boundary_masks = sdata.cell_boundary_masks
         self.interior_facet_boundary_masks = sdata.interior_facet_boundary_masks
-        self.global_numbering = sdata.global_numbering
+        # check not all degrees of freedom are constrained
+        unconstrained_dofs = self.dof_dset.size - self.dof_dset.constrained_size
+        if self.comm.allreduce(unconstrained_dofs) == 0:
+            raise ValueError("All degrees of freedom are constrained.")
```
connorjward, line 1029: Why has this been added? And I don't think it's necessarily an error condition.

firedrake/function.py
```diff
@@ -26,12 +26,12 @@
...
 from firedrake.petsc import PETSc
-from firedrake.mesh import MeshGeometry, VertexOnlyMesh
+from firedrake.mesh import MeshGeometry, VertexOnlyMeshTopology, VertexOnlyMesh
 from firedrake.functionspace import FunctionSpace, VectorFunctionSpace, TensorFunctionSpace
-from firedrake.exceptions import PointNotInDomainError
+from firedrake.exceptions import PointNotInDomainError, TopologyVersionMismatchError
 
 
-__all__ = ['Function', 'CoordinatelessFunction', 'PointEvaluator']
+__all__ = ['Function', 'PointNotInDomainError', 'CoordinatelessFunction', 'PointEvaluator', 'TopologyVersionMismatchError']
```
leo-collins: Why do these need to in `function.py`'s `__all__`?

firedrake/function.py
connorjward: If you feel you need these can they be `@property`s

firedrake/function.py
```diff
@@ -277,6 +277,16 @@ def __init__(self, function_space, val=None, name=None, dtype=ScalarType,
...
         self._expression_cache = cachetools.LRUCache(maxsize=50)
 
+        self._mesh_topo = self._function_space.topological.mesh()  # the MeshTopology object
+        self._mesh_geom = self._function_space.mesh()  # the MeshGeometry object
+
+        # Register the mesh topology version at the time the Function was created
+        self._mesh_topology_version = self._mesh_topo._topology_version
+
+        # Register the Function if it's defined on a VOM
+        if isinstance(self._mesh_topo, VertexOnlyMeshTopology):
```
connorjward: Can we do this in all cases? Just removes special casing that may need removing at a later point.

firedrake/function.py
```diff
@@ -299,8 +309,74 @@ def copy(self, deepcopy=False):
         return type(self)(self.function_space(), val=val)
 
     def __getattr__(self, name):
-        val = getattr(self._data, name)
-        return val
+        if name == 'dat':
+            self._match_mesh_topology_version()
```
connorjward: This is far too magic

firedrake/function.py
```diff
@@ -277,6 +277,16 @@ def __init__(self, function_space, val=None, name=None, dtype=ScalarType,
         # LRU cache for expressions assembled onto this function
         self._expression_cache = cachetools.LRUCache(maxsize=50)
 
+        self._mesh_topo = self._function_space.topological.mesh()  # the MeshTopology object
```
leo-collins: Also need to be careful here with `MeshSequence`s

firedrake/function.py
```diff
@@ -299,8 +309,74 @@ def copy(self, deepcopy=False):
...
 
     def __getattr__(self, name):
-        val = getattr(self._data, name)
-        return val
+        if name == 'dat':
+            self._match_mesh_topology_version()
+
+        return getattr(self._data, name)
+
+    def _match_mesh_topology_version(self):
```
connorjward: Can this be called `_migrate` or similar?

firedrake/functionspaceimpl.py
```diff
@@ -980,21 +993,19 @@ def __init__(self, function_space, boundary_set=frozenset(), name=None):
...
-        self.node_set = sdata.node_set
+    @cached_property_until(lambda self: self._mesh._topology_version)
+    def _shared_data(self):
+        return get_shared_data(self._mesh, self.ufl_element(), self.boundary_set)
+
+    @cached_property_until(lambda self: self._mesh._topology_version)
+    def dof_dset(self):
         r"""A :class:`pyop2.types.set.Set` representing the function space nodes."""
-        self.dof_dset = op2.DataSet(self.node_set, self.shape or 1,
+        self.dof_dset = op2.DataSet(self._shared_data.node_set, self.shape or 1,
```
leo-collins: I think this should be returning the dataset? It's a cached property

firedrake/function.py
```diff
@@ -299,8 +309,74 @@ def copy(self, deepcopy=False):
...
+        # Skip checking oordinate functions as they get rebuilt when the VOM gets rebuilt
+        # Note that `coordinates` is a property of any MeshGeometry object while `reference_coordinates` is only a property of the VOM
+        if hasattr(self._mesh_geom, "coordinates") and self is self._mesh_geom.coordinates:
+            return
+        if hasattr(self._mesh_geom, "reference_coordinates") and self is self._mesh_geom.reference_coordinates:
+            return
```
connorjward: This is quite ugly, can it be done more nicely? For instance I would imagine that, since coordinates are always updated, that this would end up being an identity operation and thus the usual code path would work.

firedrake/function.py
```diff
@@ -299,8 +309,74 @@ def copy(self, deepcopy=False):
...
+
+        current_mesh_version = self._mesh_topo._topology_version
+
+        if current_mesh_version != self._mesh_topology_version:
+            self._rebuild_function(current_mesh_version)
+
+    def _rebuild_function(self, current_version):
+        # Check if the mesh on which the function is defined is a VertexOnlyMesh
+        if not isinstance(self._mesh_topo, VertexOnlyMeshTopology):
+            raise TopologyVersionMismatchError(
```
connorjward: "If not a VoM then error about topology version" is a confusing pattern. The check and the error don't really agree.

firedrake/function.py
```diff
@@ -299,8 +309,74 @@ def copy(self, deepcopy=False):
...
+            chained_sf = latest_topology_step_sf  # starts from latest V -> V-1
+            # Iterate backwards through the intermediate versions
+            for v in range(current_version-1, self._mesh_topology_version, -1):
+                step_sf = self._mesh_topo._topology_step_sfs.get(v, None)  # maps V-1 -> V-2
+                if step_sf is None:
+                    raise RuntimeError(
+                        f"Failed to migrate the Function across multiple topology changes: \
+                        the intermediate topology mapping from version {v-1} to {v} could not be found."
+                    )
+                chained_sf = step_sf.compose(chained_sf)  # V -> V-2
```
connorjward: Do we need to compose SFs, or can we repeatedly call `_rebuild_function`? I am thinking of cases where the migration may not just be an SF.

firedrake/function.py
```diff
@@ -299,8 +309,74 @@ def copy(self, deepcopy=False):
...
+        # Swap the data object
+        FS_topo = self._function_space.topological
+        self._data = migrate_dg0_dat(self._data, FS_topo, latest_topology_step_sf)
+
+        # Remove cached dats and vecs on the Function since `migrate_dg0_dat` recreated the CoordinatelessFunction
+        self.__dict__.pop("dat", None)
+        self.__dict__.pop("vec", None)
+
+        # Clear any expression caches as they need to be rebuilt
+        self._expression_cache.clear()
```
connorjward: `cached_property_until`? or similar

firedrake/vertexonly_mutation.py
```diff
@@ -0,0 +1,330 @@
...
+
+        if parent_tdim > 0:
+            ref_coords_data = dmcommon.reordered_coords(swarm, ref_coords_fs.dm.getDefaultSection(),
+                                                        (topology.num_vertices(), parent_tdim), reference_coord=True)
+            self.vom.reference_coordinates._data.dat = ref_coords_fs.make_dat(val=ref_coords_data, name=self.vom.reference_coordinates.name())  # returns a new op2.Dat
+        else:
+            # This should have been already set to None when the VOM was first constructed
+            self.vom.reference_coordinates = None
+
+    def _invalidate_topology_properties(self):
```
connorjward: These should all be `topology_cached_property`?

firedrake/vertexonly_mutation.py
```diff
@@ -0,0 +1,330 @@
...
+        if "cell_parent_cell_list" in topology.__dict__:
+            topology.__dict__["cell_parent_cell_list"][:n_owned] = new_parent_cells
+
+        # Write the update reference coordinates
+        # NOTE: Since new_ref_coords is already in VOM ordering and assuming the VOM has not been reodered when this function is called,
+        # we can safely write into the reference coordinates array.
+        ref_coords_func = self.vom.reference_coordinates
+        ref_coords_func.dat.data[:] = new_refcoords
+
+    def rebuild_vom(self, absorbed_vom_indices=None):
```
connorjward: I much prefer the pattern where this a method of the `VertexOnlyMesh`.

```
Rebuild the VertexOnlyMesh using the state already stored on the VOM.
```
I think this demonstrates what I mean. The current pattern is:
1. Partially mutate the VoM
2. Hand this partially mutated VoM to another class that finishes the mutation

firedrake/vertexonly_mutation.py
```diff
@@ -0,0 +1,330 @@
...
+        swarm.restoreField("DMSwarmPIC_coor")
+
+        cid = swarm.getCellDMActive().getCellID()  # swarm field that stores each particle's parent cell ID
+        swarm_parent_cell_nums = swarm.getField(cid).ravel()
+        swarm_parent_cell_nums[:] = new_plex_parent_cell_nums_local[survivors]
+        swarm.restoreField(cid)
+
+        field_global_index = swarm.getField("globalindex").ravel()
+        field_global_index[:] = global_idxs_local[survivors]
+        swarm.restoreField("globalindex")
```
connorjward: Could this function be split apart into `migrate_swarm` and the rest?

firedrake/vertexonly_mutation.py
```diff
@@ -0,0 +1,330 @@
...
+        dmcommon.mark_entity_classes_using_cell_dm(swarm)  # rewrites class labels on based on the ownership of the plex parent cell
+        topology._entity_classes = dmcommon.get_entity_classes(swarm)  # reads swarm labels and store counts on the vom's topology
+
+        # Rebuild _cell_numbering and _vertex_numbering from the new _dm_renumbering
+        entity_dofs = np.array([1], dtype=IntType)  # 1 DoF per point
+
+        # cell_numbering and vertex_numbering are PETSc ISes - translation tables between PETSc numbering of mesh entities and that of Firedrake's
+        # For each plex point, they store two integers: the dof count and an offset (which happens to be the Firedrake's number of that plex point)
+        topology._cell_numbering, _ = topology.create_section(entity_dofs)
+        topology._vertex_numbering = topology._cell_numbering  # holds for VOM only
```
connorjward: All this crud has to be here because the `__init__` method does a whole lot of magic. I think if we made `topology._vertex_numbering` into a `topology_cached_property` then a lot of this could just go away. It is appealing for us to just tweak the swarm, increment the topology count, and then have everything else recompute as needed.

connorjward (changes requested):
This is a big change. Lots of feedback but this is really promising.

firedrake/vertexonly_mutation.py
```diff
@@ -0,0 +1,330 @@
...
+
+        field_input_index = swarm.getField("inputindex").ravel()
+        field_input_index[:] = input_idxs_local[survivors]
+        swarm.restoreField("inputindex")
+
+        # Redistribute particles accross ranks
+        swarm.migrate(remove_sent_points=True)
+
+        # Reconstruct Firedrake cell numbers from the receiving rank's updated plex data
+        swarm_plex_ids = swarm.getField(cid).ravel()
```
leo-collins: I also don't really like this `getField` and `restoreField` pattern. We also have it everywhere in the vom code in mesh.py. I think we should introduce a context manager somewhere, probably on `FiredrakeDMSwarm`, to handle this.

firedrake/functionspaceimpl.py
leo-collins: I thought this but it has just been moved. It was always in `set_shared_data`.

connorjward: Ah OK. Not your bug to fix.

firedrake/function.py
```diff
@@ -870,3 +946,29 @@ def make_c_evaluate(function, c_name="evaluate", ldargs=None, tolerance=None):
...
+    nroots, ilocal, remote = step_sf.getGraph()
+    nleaves = remote.shape[0] if ilocal is None else ilocal.shape[0]
+    new_vals = np.empty((nleaves, dim), dtype=old_vals.dtype)
+    for c in range(dim):
+        root = np.ascontiguousarray(old_vals[:, c])
+        leaf = np.empty(nleaves, dtype=root.dtype)
+        unit = MPI._typedict[np.dtype(root.dtype).char]
+        step_sf.bcastBegin(unit, root, leaf, MPI.REPLACE)
+        step_sf.bcastEnd(unit, root, leaf, MPI.REPLACE)
+        new_vals[:, c] = leaf
```
leo-collins: This looks similar to what the halo exchange code does. You should use the `_get_mtype` function in `halo.py` and then you can do a single collective instead of multiple.

firedrake/mesh.py
achanbour: I call this eager migration method in my particle solver to ensure that all the fields (Functions) that directly affect particle advection are migrated as soon as the VOM gets rebuilt. We initially thought of doing that because it allowed us to rule out errors in the stepper routines wherein stale Functions were used because they weren't migrated at the right time. Getting migration out of the way before the next operation (which relies on Functions owned by the stepper being updated), seemed like a robust option.

Also wouldn't be good to give the user the freedom to say "now migrate all the Functions in this set" at some time in his code (e.g., before doing some post processing or writing data for visualisation purposes)?

firedrake/function.py
achanbour: This is the way I thought lazy migration would work - migrate the next time the Function gets accessed. What's the alternative?

firedrake/function.py
achanbour: I struggle to think about how the migration would still work without SF composition. Currently, we're guaranteed an SF every time the VOM changes and we have an SF between each pair of consecutive versions. If the user creates a Function say at version `k-2` and the VOM is at version `k` when this Function gets accessed (provided the user didn't request to migrate it eagerly) then we'd somehow need to know how the VOM changed from `k-2` (the topology at which the Function was defined) to `k` (the VOM's newest topology).

A related idea we had previously was to cache the composition and drop the one step SFs from which the composition was built. This way, if we need to compose at some point next, we will be able to reuse that composition rather than re-composing step-by-step. But it is unclear to me right now how often should we cache the composed SFs and what compositions should be cached at all.

firedrake/vertexonly_mutation.py
achanbour: So 1) should be in the `VertexOnlyMeshMutator` and 2) should be a method of the `VertexOnlyMeshTopology`?

firedrake/function.py
achanbour: I think given they are immutable identity references, it's fine to keep them as private attributes?

firedrake/exceptions.py
```diff
@@ -77,4 +77,7 @@ def __str__(self):
 
 class TopologyVersionMismatchError(ValueError):
     """Raised when a Function's mesh topology attribute does not match the topology version attribute stored on the mesh object."""
-    pass
+
+
+class FunctionMigrationError(RuntimeError):
+    """Raised when Function data cannot be migrated between mesh topology versions."""
```
leo-collins: These need to subclass `FiredrakeException`

firedrake/vertexonly_mutation.py
connorjward: I don't think a separate class (`VertexOnlyMeshMutator`) is needed. All of this logic naturally lives on the VoM.

firedrake/function.py
connorjward: They aren't strictly immutable since a user could modify them, but my main point is to try and reduce the amount of complexity in our `__init__` methods.

For 'dependent' attributes like this I think `@property` is a better pattern because you never need to consider the case where `self._mesh_geom` and `self.function_space().mesh()` disagree.

firedrake/mesh.py
connorjward: > We initially thought of doing that because it allowed us to rule out errors in the stepper routines wherein stale Functions were used because they weren't migrated at the right time.

I can understand its use as a debugging approach.

> Also wouldn't be good to give the user the freedom to say "now migrate all the Functions in this set" at some time in his code (e.g., before doing some post processing or writing data for visualisation purposes)?

But I can't think of a reason for us to otherwise support this. The lazy approach should always work and will be strictly more performant because of doing no unnecessary work.

firedrake/function.py
connorjward, line 239: All topological logic should really live on the `CoordinatelessFunction` rather than here. Also we'll need to duplicate this for `Cofunction` (though confusingly there is no `CoordinatelessCofunction`).

firedrake/function.py
```diff
@@ -277,6 +277,16 @@ def __init__(self, function_space, val=None, name=None, dtype=ScalarType,
         # LRU cache for expressions assembled onto this function
         self._expression_cache = cachetools.LRUCache(maxsize=50)
 
+        self._mesh_topology = self._function_space.topological.mesh()  # the MeshTopology object
+        self._mesh_geometry = self._function_space.mesh()  # the MeshGeometry object
+
+        # Register the mesh topology version at the time the Function was created
+        self.self._mesh_topology_version = self._mesh_topology._topology_version
```
```suggestion
        self._mesh_topology_version = self._mesh_topology._topology_version
```

firedrake/function.py
connorjward: Ah I see. The right way to do is this is with `@property`:
```py
@property
def dat(self):
  self._match_mesh_topology_version()
  return self._dat
```

This is how we `Dat.data_rw` for instance.

firedrake/function.py
connorjward: I am suggesting chaining the migrations as distinct steps with per-migration SFs, instead of building a composed SF that does it in one shot. Consider the `k-2` to `k` example you showed. I'm basically saying to do
```py
step_sf1 = self._mesh_topology._topology_step_sfs.get(...)
self._data = migrate_dg0_dat(self._data, FS_topo, step_sf1)
step_sf2 = self._mesh_topology._topology_step_sfs.get(...)
self._data = migrate_dg0_dat(self._data, FS_topo, step_sf2)
```
instead of building the composed thing.

Also you're building the composed SF every time you migrate a function. If you have lots of functions with different topology versions that could get expensive.

And lastly I think it is reasonable to think about cases where the migration is done via a non-SF method. For example if you have some sort of mesh adaptivity you might refine a cell into some smaller cells. To transfer the data you'd have to do an interpolation. We can stack interpolation operations but I don't know how we would compose them into a single 'super interpolation'.

firedrake/mesh.py
achanbour: My only concern is that since I still got work in progress that relies on Functions being migrated properly (and at the right time) wouldn't it make sense to leave this eager migration method (as a private method perhaps?) so I can always fall back to it if things go wrong? That said, the default, from the user's perspective would be lazy migration. Or do you prefer to strip this out from this PR and only leave it in my dev branch?

connorjward: It seems like you never actually call this anywhere. In general I don't mind having useful debugging routines added provided that they are clearly marked as such. However, in this case I have a couple of observations:

If you pass in `functions_to_migrate` you're basically doing:
```py
mesh._migrate_functions(funcs)
```
when you could equally just do
```py
for f in funcs:
  f._match_mesh_topology_version()
```
which I think is much more understandable.

With that in mind it probably makes sense to drop `functions_to_migrate` as an argument here and rename the method to `_migrate_all_functions` or similar.

firedrake/function.py
achanbour: If `self._function_space.topological.mesh()` returns a `MeshSequence` then the versioning mechanism still works as I just added a `_topology_version` property inside the `MeshSequence` class:

```
@property
    def _topology_version(self):
        return tuple(mesh._topology_version for mesh in self._meshes)
```

The migration itself, however, wouldn't work because it currently checks that the Function being updated is one defined on a VOM:

```
def _rebuild_function(self, current_version):
        # Check if the mesh on which the function is defined is a VertexOnlyMesh
        if not isinstance(self._mesh_topology, VertexOnlyMeshTopology):
            raise TopologyVersionMismatchError(
                "The mesh topology has changed since this Function was created, \
                and migration is currently only supported for Functions defined on VertexOnlyMeshes. \
                Please re-create this Function on the updated mesh."
            )
```

I think this is fine for now? It would seem wrong to only migrate the VOM component for a MeshSequence and leave everything else stale.

firedrake/function.py
achanbour: Yep. It falls through the check automatically if the topology version stashed on the coordinate functions is incremented to the latest VoM topology version after they get rebuilt (which it should). Fixed now.

firedrake/mesh.py
achanbour: More of a defensive choice here than anything else but yes, the keys are always monotonically increasing and the `_live_functions` is an insertion-ordered dictionary. The GC may remove Functions which may result in gaps between the keys but the keys never do get reordered afaik.

achanbour:
@connorjward regarding your earlier comment about chaining migrations instead of composing the step SFs: 
`migrate_dg0_dat` takes as one of its arguments the topological FS describing the new cell-node layout of the topology we're migrating into i.e., the _latest_ version. But migrating from version `k-2` to `k` requires us knowing the intermediate topology state (at version `k-1`) which is currently not preserved (all fields such as `_cell_numbering`, `_vertex_numbering` entity classes etc. are overwritten to reflect the last version `k`). All we have left from version `k-1` is effectively the step SF `k-2` -> `k-1` hence why SF composition works nicely. 

For transferring data through interpolation, it would make sense to save a snapshot of the topological FS at each intermediate version, in which case stacking the migration/interpolation operations would work. At first impression, it doesn't sound like we're saving much more than simply doing the SF composition. Also, for the DG0 VOM case I'm dealing with now, composing SFs seems the most natural thing to do. What do you think?

connorjward:
> But migrating from version k-2 to k requires us knowing the intermediate topology state (at version k-1) which is currently not preserved (all fields such as _cell_numbering, _vertex_numbering entity classes etc. are overwritten to reflect the last version k). All we have left from version k-1 is effectively the step SF k-2 -> k-1 hence why SF composition works nicely.

Ahh yes. Bloody mutable state... OK I see that my previous proposal is not the right way to design the abstraction. Crack on.

firedrake/function.py
```diff
@@ -79,15 +79,41 @@ def __init__(self, function_space, val=None, name=None, dtype=ScalarType):
...
+
+    @dat.setter
+    def dat(self, value):
+        if value is self._dat:
+            return
+        raise AttributeError("A Function's Dat cannot be replaced directly.")
+
+    def _migrate_to_current_topology_version(self) -> None:
+        """Migrate this coordinateless function's data to the current topology version."""
+        _migrate_dg0_coefficient(self, self._function_space)
```
connorjward, line 115: This is the place where we should check that we are actually DG0 and fail appropriately (`NotImplementedError`)

firedrake/function.py
```diff
@@ -79,15 +79,41 @@ def __init__(self, function_space, val=None, name=None, dtype=ScalarType):
...
+        """Return the mesh topology on which this coordinateless function is defined."""
+        return self._function_space.topological.mesh()
+
+    @property
+    def dat(self):
+        self._migrate_to_current_topology_version()
+        return self._dat
+
+    @dat.setter
+    def dat(self, value):
```
connorjward, line 108: We should have a comment explaining things here because this is an extremely strange pattern.

firedrake/function.py
```diff
@@ -79,15 +79,41 @@ def __init__(self, function_space, val=None, name=None, dtype=ScalarType):
...
+        if value is self._dat:
+            return
+        raise AttributeError("A Function's Dat cannot be replaced directly.")
```
connorjward, line 111:
```suggestion
        raise AttributeError("The 'dat' of a function cannot be changed")
```
small thing but I think this is a bit clearer

firedrake/function.py
connorjward, line 1: It is not a requirement for this PR, but you can probably now see that `Function` and `Cofunction` share an awful lot of code. It is somewhere on my TODO list to build a parent `FunctionSpaceData` (when the old version of that class dies) class for the shared functionality. One day...

connorjward (changes requested):

PR #5296, opened 2026-07-29
Access DMSwarm fields via a context manager

leo-collins:
Removes the awkward `getField` and remembering to call `restoreField` pattern.

@achanbour

### DIFF
--- a/firedrake/mesh.py
+++ b/firedrake/mesh.py
@@ -3739,7 +3739,7 @@ class FiredrakeDMSwarm(PETSc.DMSwarm):
         self._other_fields = None
 
     @contextmanager
-    def field(self, name: str) -> Generator[np.ndarray, None, None]:
+    def field(self, name: str) -> Generator[np.ndarray]:
         """Context manager to access a field on the DMSwarm.
 
         Parameters
--- a/firedrake/mesh.py
+++ b/firedrake/mesh.py
@@ -3753,6 +3753,7 @@ class FiredrakeDMSwarm(PETSc.DMSwarm):
             The field as a NumPy array. The array and views derived from it
             must not be used after leaving the context.
         """
+        # petsc4py will error if you try to access an active field without first restoring it.
         values = self.getField(name)
         try:
             yield values
--- a/firedrake/mesh.py
+++ b/firedrake/mesh.py
@@ -1,6 +1,7 @@
 import dataclasses
 import numpy as np
 import ctypes
+from contextlib import contextmanager
 import os
 import sys
 import ufl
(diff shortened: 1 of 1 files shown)


firedrake/mesh.py
```diff
@@ -3746,6 +3738,27 @@ def __init__(self, *args, **kwargs):
         self._default_extra_fields = None
         self._other_fields = None
 
+    @contextmanager
+    def field(self, name: str) -> Generator[np.ndarray, None, None]:
```
connorjward: Why `None, None`?

firedrake/mesh.py
```diff
@@ -3746,6 +3738,27 @@ def __init__(self, *args, **kwargs):
...
+        name : str
+            The name of the field to access.
+
+        Yields
+        ------
+        numpy.ndarray
+            The field as a NumPy array. The array and views derived from it
+            must not be used after leaving the context.
+        """
+        values = self.getField(name)
```
connorjward, line 3757: It would probably be good design to catch if you access the same field in a nested fashion:
```py
def __init__(self):
  self._active_fields = {}
...
def field(self, name):
  if name in self._active_fields:
    # either error or return the field again
  values = self.getField(name)
  self._active_fields[name] = value
  yield values
  del self._active_fields[name]
  ...
```

connorjward:
Lovely

firedrake/mesh.py
leo-collins: petsc4py throws an error anyway if you do that:

```
E   petsc4py.PETSc.Error: error code 83
E   [0] DMSwarmGetField() at /Users/lac224/Coding/work/firedrake-dev2/petsc/src/dm/impls/swarm/swarm.c:1930
E   [0] DMSwarmDataFieldGetAccess() at /Users/lac224/Coding/work/firedrake-dev2/petsc/src/dm/impls/swarm/data_bucket.c:381
E   [0] Field "refcoord" is already active. You must call DMSwarmDataFieldRestoreAccess()
```

firedrake/mesh.py
leo-collins: They're the types for the send and return types, but it looks like python sets them to None by default anyway. I've removed them.

firedrake/mesh.py
connorjward: It would be good to have a comment stating that, because otherwise someone in the future might come along and go "aha, that must be a bug".

achanbour:
Thanks for this! Looking forward to when this gets merged so I can modify my code :)

connorjward (approved):

PR #5298, opened 2026-07-29
Cache cell-facet topological data on meshes

achanbour:
Compute and cache properties derived from cell-facet adjacency data on meshes:

- `cell_facet_neighbours`: maps each cell to its neighbouring cells across each of its local facets
- `cell_facet_coord_transforms`: stores affine reference-coordinate transforms between neighbouring cells across interior facets
- `cell_facet_exterior_mask`: marks cell-local facets lying on the exterior domain boundary (distinguishes exterior facets from rank partition boundaries)


firedrake/mesh.py
```diff
@@ -1431,6 +1431,341 @@ def cell_set(self):
...
+        dset = op2.DataSet(self.cell_set, dim=num_facets)
+
+        # Create a local numpy buffer to store the neighbours of each cell
+        cell_neighbours = np.full((self.num_cells(), num_facets), -1, dtype=IntType)
+
+        # Populate the local buffer by iterating over the cells and querying the DMPlex for local cell information
+        plex = self.topology_dm
+        cStart, cEnd = plex.getHeightStratum(0)  # range of DMPlex point numbers representing cells
+
+        for c_plex_point in range(cStart, cEnd):
```
connorjward, line 1455: We want to avoid Python loops over the mesh. This should go in some Cython.

firedrake/mesh.py
```diff
@@ -1431,6 +1431,341 @@ def cell_set(self):
...
+            dset,
+            cell_neighbours,
+            dtype=IntType,
+            name="cell-facet-neighbours-dat"
+        )
+
+        # NOTE: A fresh Dat has its halo entries marked as invalid -> next with_halos access can trigger a collective exchange.
+        # An exchange can overrite a rank's data with the owning rank's local data which is wrong.
+        # We avoid that by marking halos as valid.
+        cell_facet_neighbours.halo_valid = True
```
connorjward, line 1478: Does this mean that the `Dat` is logically over `COMM_SELF`? If so then that's what we should do. That may be a pyop3 thing (definitely easy there, not sure with PyOP2)

firedrake/mesh.py
```diff
@@ -2414,6 +2749,8 @@ def __init__(self, coordinates):
         V = functionspaceimpl.WithGeometry(coordinates.function_space(), self)
         self._coordinates_function = function.Function(V, val=coordinates)
 
+        self._topology_version = 0
```
connorjward, line 2752: Rogue entry from another PR?

firedrake/mesh.py
```diff
@@ -1431,6 +1431,341 @@ def cell_set(self):
...
+            )
+        else:
+            ext_plex_points = frozenset()
+
+        cstart, cend = plex.getHeightStratum(0)
+        for c_plex in range(cstart, cend):
+            cid = self._cell_numbering.getOffset(c_plex)
+
+            if cid < 0 or cid >= self.cell_set.total_size:
+                continue
```
connorjward, line 1761: when does this happen?

firedrake/mesh.py
```diff
@@ -1431,6 +1431,341 @@ def cell_set(self):
...
+        plex = self.topology_dm
+        if plex.getStratumSize("exterior_facets", 1) > 0:
+            ext_plex_points = frozenset(
+                plex.getStratumIS("exterior_facets", 1).getIndices().tolist()
+            )
+        else:
+            ext_plex_points = frozenset()
+
+        cstart, cend = plex.getHeightStratum(0)
+        for c_plex in range(cstart, cend):
```
connorjward, line 1757: Same comment about Cython

firedrake/mesh.py
```diff
@@ -1431,6 +1431,341 @@ def cell_set(self):
...
+    # helper methods
+
+    @cached_property
+    def _cell_closure_facet_offset(self):
+        """Returns the correct column in cell_closure for a given local facet."""
+        facet_dim = self.cell_dimension() - 1
+        topology = FIAT.ufc_cell(self.ufl_cell()).get_topology()
+        return sum(len(topology[d]) for d in range(facet_dim))
+
+    def _cell_facet_point(self, cell, local_facet):
```
connorjward, line 1491: You are calling this in a tight loop so probably best to not dispatch to a Python call

firedrake/mesh.py
```diff
@@ -1431,6 +1431,341 @@ def cell_set(self):
...
+        cstart, cend = plex.getHeightStratum(0)
+        for c_plex in range(cstart, cend):
+            cid = self._cell_numbering.getOffset(c_plex)
+
+            if cid < 0 or cid >= self.cell_set.total_size:
+                continue
+
+            for lf in range(num_facets):
+                f_point = self._cell_facet_point(cid, lf)
+                mask[cid, lf] = f_point in ext_plex_points
```
connorjward, line 1765: I wonder if this could all be numpified, instead of cythonified. Use `np.intersect1d` between a slice of `cell_closure` and `exterior_facets.getStratumIS(1)`.

firedrake/mesh.py
```diff
@@ -1431,6 +1431,341 @@ def cell_set(self):
...
+                # skip outer halo facets where adjacency data isn't available
+                continue
+
+            # Compute coords. map from o0 -> o1
+            Q0, t0 = o_coord_maps[o0]  # map canonical orientation -> o0
+            Q1, t1 = o_coord_maps[o1]  # map canonical orientation -> o1
+
+            # inverse coord. map
+            # o0 -> canonical orientation
+            Q0_inv = np.linalg.inv(Q0)
```
connorjward, line 1693: Is this not really slow? I think you're calling this for every facet.

firedrake/mesh.py
```diff
@@ -1431,6 +1431,341 @@ def cell_set(self):
...
+            mapped_facet_verts = np.array([phi(v) for v in facet_verts])
+            b_lf = mapped_facet_verts[0]  # first vertex is always the origin
+            A_lf = (mapped_facet_verts[1:facet_dim+1] - b_lf).T
+            A.append(A_lf)
+            b.append(b_lf)
+            A_inv.append(np.linalg.pinv(A_lf))
+
+        return (A, b, A_inv)
+
+    def _get_facet_orientation_coord_maps(self):
```
connorjward, line 1542: This only uses `self` in one location, and only to get the UFL cell. This seems like it doesn't need to be a method of the mesh.
I wonder if we should have a `firedrake/facets.py` file to collect these routines. It could then be folded into the `mesh` subpackage when we get around to doing that.

firedrake/mesh.py
```diff
@@ -1431,6 +1431,341 @@ def cell_set(self):
...
+
+        X = A_lf @ X_lf + b_lf.
+
+        Returns
+        -------
+        tuple of lists (A, b, A_inv) each indexed by local facet ID.
+            (A[lf], b[lf]) defines the affine embedding of facet reference coordinates into the cell reference frame,
+            and A_inv[lf] defines the pseudoinverse for mapping cell coordinates on the facet back to facet-local coordinates.
+        """
+        import FIAT
```
connorjward, line 1508: top-level import

connorjward (changes requested):

firedrake/mesh.py
```diff
@@ -1431,6 +1431,341 @@ def cell_set(self):
...
+        ref_cell = FIAT.ufc_cell(self.ufl_cell())
```
pbrubeck, line 1509: I think these functions live in FIAT, these are very fundamental operations at the reference elemenet level

firedrake/mesh.py
pbrubeck: This should definitely live in FIAT

pbrubeck: Specifically FIAT/reference_element.py

firedrake/mesh.py
```diff
@@ -1431,6 +1431,341 @@ def cell_set(self):
...
+            for lf in range(num_facets):
+                A_lf = np.empty((ref_cell.get_spatial_dimension(), 0), dtype=RealType)
+                b_lf = np.asarray(vertices[lf], dtype=RealType)
+                A.append(A_lf)
+                b.append(b_lf)
+                A_inv.append(np.empty((0, ref_cell.get_spatial_dimension()), dtype=RealType))
+            return A, b, A_inv
+
+        for lf in range(num_facets):
+            phi = ref_cell.get_entity_transform(facet_dim, lf)
```
pbrubeck, line 1528: Is this reverse-engineering get_entity_transform? I.e. are you assuming an affine mapping for the entity transform and rederiving by sampling it at the vertices?

pbrubeck: Tensor product cells will not give you affine mappings. This has to be carefully thought. What is the goal/motivation for storing the facet data on the mesh?

firedrake/mesh.py
pbrubeck: you can batch these things as done in https://github.com/firedrakeproject/firedrake/pull/5034/changes#diff-e45e704e2c1f2f7f0bd632f90a71ce568a47c880cdf097394d8bdf4ccb0332b6R104

However, the geometric computation done there is avoidable and there is a purely topological way of achieving the goal that we inteded there (permuting an array of coordinates). Perhaps you can achieve your goal (which is not clear from the PR description) via a purely topological approach, in a coordinate-free way.

firedrake/mesh.py
achanbour: I store the facet data on the mesh as my particle trajectory algorithm solves for mesh entity collisions (identifies which facet was crossed by each particle and its crossing position on that facet). I would then like to be able to tell, based on the crossed facet ID: 1) which is the neighbouring cell across that facet and 2) transform the coordinates of the point on that facet to the local (reference) coordinates of the "next cell" across that facet.

pbrubeck: I see. Often it helps to follow the existing patterns established in Firedrake. In this case the main pattern to consider is code generation. When we generate code to assemble a form, we don't precompute/store in the mesh the geometric information (Jacobians/affine mappings). We compute everything from the mesh coordinates withtin the generated C kernels (avoiding python loops over the cells).

It seems to me that the right thing is to generate a facet kernel. A facet kernel iterates over pairs of cells sharing a facet, this solves 1). You can then access the coordinates of both cells and generate code from `ufl.FacetJacobian(mesh)`, this solves 2).

Generation of bespoke code is quite painful, but we are here to help you. I think by sticking to the established patterns one can get a very general particle code that works on high-order meshes of any cell type.

achanbour: Yeah, I completely understand your point. As this idea of writing kernels to generate code is pretty new to me, could you point out where I could look for these code patterns for example in existing kernels that achieve some sort of similar operation (e.g., the form assembly you mentioned)? It will help me get started on this task!

PR #5301, opened 2026-07-30
fix rpaths for macos

leo-collins:
Fixes an issue where editable installs of firedrake-rtree and libsupermesh are unable to found on macos

PR #5302, opened 2026-07-30
Connorjward/merge release

connorjward:
# Description


connorjward:
The only change is to `firedrake-configure` for macOS, which CI won't hit. It's also broken and will be fixed in https://github.com/firedrakeproject/firedrake/pull/5301 right away. Merging now.

PR #5303, opened 2026-07-30
macos fix firedrake-configure again (--download-pnetcdf)

connorjward:
The fix in #5299 was very short lived. It now seems like we need `--download-pnetcdf` too.


connorjward:
Merging aggressively because this fix is really needed.

PR #5307, opened 2026-07-31
Increase iteration counts for flaky test

connorjward:
This has been causing stochastic CI failures for about a year at this point. From local experimentation I think increasing the iteration count to 32 is fine - it's still a lot less than you get if you take the near nullspace away (about 50).

### DIFF
--- a/tests/firedrake/regression/test_nullspace.py
+++ b/tests/firedrake/regression/test_nullspace.py
@@ -389,4 +389,5 @@ def test_near_nullspace_mixed(aux_pc, rhs):
     A, P = ksp_inner.getOperators()
     assert A.getNearNullSpace().handle
     # currently ~22 (25 on 2 cores) vs. >45-ish for with/without near nullspace
-    assert ksp_inner.getIterationNumber() < 27
+    # CI sometimes hits around 27, so we comfortably exceed this
+    assert ksp_inner.getIterationNumber() < 32
--- a/tests/firedrake/regression/test_nullspace.py
+++ b/tests/firedrake/regression/test_nullspace.py
@@ -313,7 +313,7 @@ def test_near_nullspace_mixed(aux_pc, rhs):
     aP = None
     mu0 = mu
     if aux_pc:
-        DG0 = FunctionSpace(mesh, "DG", 0)
+        DG0 = FunctionSpace(mesh, "DG", 0, variant="integral(6)")
         mu0 = Function(DG0).interpolate(mu)
         aP = inner(mu0*2*sym(grad(u)), grad(v))*dx(degree=2)
         aP += -inner(p, div(v))*dx + inner(div(u), q)*dx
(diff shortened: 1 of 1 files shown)


pbrubeck:
Is this mainly failing on complex, or also real?

connorjward:
> Is this mainly failing on complex, or also real?

I saw it on real this afternoon: https://github.com/firedrakeproject/firedrake/actions/runs/30628566798/job/91149400137#step:9:1

tests/firedrake/regression/test_nullspace.py
pbrubeck:
```suggestion
        DG0 = FunctionSpace(mesh, "DG", 0, variant="integral(6)")
```

pbrubeck: Not sure if this will fix it, but this is based on the comment about quadrature degree above.

pbrubeck: Also, it is weird that the same test passes when the rhs is a Form but fails on a Cofunction

connorjward: This is a stochastic failure, so that doesn't necessarily surprise me.

tests/firedrake/regression/test_nullspace.py
```diff
@@ -313,7 +313,7 @@ def test_near_nullspace_mixed(aux_pc, rhs):
     aP = None
     mu0 = mu
     if aux_pc:
-        DG0 = FunctionSpace(mesh, "DG", 0)
+        DG0 = FunctionSpace(mesh, "DG", 0, variant="integral(6)")
```
pbrubeck:
```suggestion
        DG0 = FunctionSpace(mesh, "DG", 0)
```

tests/firedrake/regression/test_nullspace.py
```diff
@@ -389,4 +389,5 @@ def test_near_nullspace_mixed(aux_pc, rhs):
     A, P = ksp_inner.getOperators()
     assert A.getNearNullSpace().handle
     # currently ~22 (25 on 2 cores) vs. >45-ish for with/without near nullspace
-    assert ksp_inner.getIterationNumber() < 27
+    # CI sometimes hits around 27, so we comfortably exceed this
+    assert ksp_inner.getIterationNumber() < 32
```
pbrubeck:
```suggestion
    assert ksp_inner.getIterationNumber() < 24
```

tests/firedrake/regression/test_nullspace.py
pbrubeck:
```suggestion
    # Force a deterministic parallel distribution
    mesh = UnitSquareMesh(n, n, distribution_parameters={"partitioner_type": "simple"})
```

pbrubeck, line 360: the relaxation depends on the ordering of the unknowns, which depends on the parallel distribution, which by default is non-deterministic

connorjward: Awesome. This is a much better fix.

pbrubeck (approved):

PR #5317, opened 2026-08-03
cache mesh `cell_sizes`

leo-collins:
Decorate `cell_sizes` with a `cached_property_until` the coordinates change, and deprecate `clear_cell_sizes`

### DIFF
--- a/firedrake/mesh.py
+++ b/firedrake/mesh.py
@@ -2463,10 +2463,11 @@ values from f.)"""
 
         Use this if you move the mesh.
         """
-        try:
-            del self.cell_size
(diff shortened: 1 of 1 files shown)


leo-collins:
@connorjward Interestingly, the `clear_cell_sizes` function deleted the `cell_size` attribute, which doesn't exist, and not the `cell_sizes` attribute. So it appears this function wasn't doing anything. It has been like this since 2018.

connorjward:
That's fun. We should probably still deprecate it properly in case it was getting called (pointlessly) by users.

PR #5319, opened 2026-08-03
interpolator init cleanup

leo-collins:
Tidy up the init methods for the interpolator classes. Before we were determining the source and target meshes both in the `__init__` of `Interpolator` and the `_interpolator` dispatcher. This sorts out the ugly `Delay calling .unique() because MixedInterpolator is fine with MeshSequence` inheritance we had before.

This also allows the nice `interpolate(mixed_expr, mixed_space)` in the case when `mixed_space` is defined over multiple domains.

### DIFF
--- a/firedrake/interpolation.py
+++ b/firedrake/interpolation.py
@@ -172,6 +172,8 @@ class Interpolate(UFLInterpolate):
         try:
             source_mesh = extract_unique_domain(operand) or target_mesh
         except ValueError:
+            if has_mixed_arguments or len(self.target_space) > 1:
+                return MixedInterpolator(self)
             raise NotImplementedError(
                 "Interpolating an expression with no arguments defined on multiple meshes is not implemented yet."
             )
--- a/firedrake/interpolation.py
+++ b/firedrake/interpolation.py
@@ -833,9 +833,7 @@ class SameMeshInterpolator(Interpolator):
 class VomOntoVomInterpolator(SameMeshInterpolator):
 
     def __init__(self, expr: Interpolate, source_mesh, target_mesh):
-        super().__init__(expr)
-        self.source_mesh = source_mesh
-        self.target_mesh = target_mesh
+        super().__init__(expr, source_mesh, target_mesh)
 
         if self.source_mesh.input_ordering is self.target_mesh:
             # The forward interpolation is a star forest reduction
--- a/firedrake/interpolation.py
+++ b/firedrake/interpolation.py
@@ -188,19 +188,19 @@ class Interpolate(UFLInterpolate):
             and target_mesh.topological_dimension == source_mesh.topological_dimension
         )
         if target_mesh is source_mesh or submesh_interp_implemented:
-            return SameMeshInterpolator(self)
+            return SameMeshInterpolator(self, source_mesh, target_mesh)
 
         if isinstance(target_mesh.topology, VertexOnlyMeshTopology):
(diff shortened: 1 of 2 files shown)


tests/firedrake/regression/test_interpolate_cross_mesh.py
```diff
@@ -756,3 +756,37 @@ def test_mixed_interpolator_cross_mesh():
...
+    if rank == 1:
+        result = assemble(interpolate(expr, W))
+        expected1 = assemble(interpolate(expr1, V))
+        expected2 = assemble(interpolate(expr2, U))
+
+        assert np.allclose(result.subfunctions[0].dat.data_ro, expected1.dat.data_ro)
+        assert np.allclose(result.subfunctions[1].dat.data_ro, expected2.dat.data_ro)
+    else:
+        v1, v2 = TestFunctions(W)
+        form = v1 * dx(domain=mesh1) + v2 * dx(domain=mesh2)
```
pbrubeck:
```suggestion
        form = conj(v1) * dx(domain=mesh1) + conj(v2) * dx(domain=mesh2)
```

tests/firedrake/regression/test_interpolate_cross_mesh.py
```diff
@@ -756,3 +756,37 @@ def test_mixed_interpolator_cross_mesh():
...
+@pytest.mark.parallel([1, 3])
+@pytest.mark.parametrize("rank", [0, 1])
+def test_interpolate_mixed_expr(rank):
+    mesh1 = UnitSquareMesh(3, 3)
+    mesh2 = UnitSquareMesh(4, 4)
+    x1, y1 = SpatialCoordinate(mesh1)
+    x2, y2 = SpatialCoordinate(mesh2)
+    expr1 = x1 + y1
+    expr2 = x2 - y2
+    expr = as_vector([expr1, expr2])
```
pbrubeck: This is the boring case that boils down to two separate same-mesh sub interpolators. If you flip the order here you get the interesting one. Do we test that one as well?

leo-collins: Changed the test so it actually does cross-mesh interpolation. We do the same mesh case already in test_interpolate.py

firedrake/interpolation.py
pbrubeck, line 241: Shouldn't the base class get the `source_mesh` and `target_mesh` arguments?

pbrubeck: Or we don't want MixedInterpolator to keep track of the source and target MeshSequences?

leo-collins: MixedInterpolator splits the interpolate and calls `get_interpolator`, which then determines the source and target meshes for the sub-blocks. It doesn't need source_mesh and target_mesh passed to it.

pbrubeck: It does not need them, sure. But just to be consistent shouldn't we add them there as well?

connorjward (approved):
This seems fine. I'm convinced that `MixedInterpolator` is a bad design choice, but this change is very minor and clearly enables some nice top-level API.

firedrake/interpolation.py
connorjward: I don't think consistency is a concern here.

connorjward: @pbrubeck are you OK to resolve this thread?

PR #5321, opened 2026-08-04
Fix coarse-to-fine cell maps for hierarchies built on an adaptively refined mesh

pbrubeck:
Bugfix for a regression introduced by #5215.

Building a `MeshHierarchy` on a mesh returned by `refine_marked_elements` failed in parallel, asserting `0 <= ccell < ncoarse` in `mgimpl.coarse_to_fine_cells`.

**Root cause.** `coarse_to_fine_cells` should reindex the coarse mesh's plex-to-Firedrake cell map from overlapped to non-overlapped cell numbering, which it does by sending every local cell through the overlapped local-to-global map and back through the non-overlapped one. It assumed that the cells surviving that round trip were the leading cells of the overlapped plex: it discarded the rest (`MapMode.DROP`) and sliced. Meshes distributed by `DMPlexDistributeOverlap` append their halo cells, so the assumption held for every mesh a hierarchy could be built on. #5215 added adaptive refinement, whose plex comes from `DMPlexTransform`; that numbers cells by refinement case and interleaves owned with halo cells, so the slice was the wrong permutation, cells were attributed to the wrong parent, and the assertion fired.

**Fix.** `coarse_to_fine_cells` now masks the absent cells (`MapMode.MASK`) and scatters each surviving entry to the non-overlapped number the round trip gives it. For a mesh whose owned cells do lead its plex, this yields the same permutation as before.

The docstring and comments of `coarse_to_fine_cells` are rewritten: numpydoc, and naming the three cell numberings the function moves between (Firedrake, overlapped plex, non-overlapped plex), none of which the ordering assumption was previously stated against. A test covers a uniform hierarchy built on an adaptively refined mesh.

---

AI was used for this change (Claude Code).

### DIFF
--- a/firedrake/cython/mgimpl.pyx
+++ b/firedrake/cython/mgimpl.pyx
@@ -381,14 +381,44 @@ def adaptive_parent_child_cell_maps(PETSc.DM coarse_dm,
 @cython.boundscheck(False)
 @cython.wraparound(False)
 def coarse_to_fine_cells(mc, mf, clgmaps, flgmaps):
-    """Return a map from (renumbered) cells in a coarse mesh to those
-    in a refined fine mesh.
+    """Map the cells of a coarse mesh to those of its uniform refinement.
 
-    :arg mc: the coarse mesh to create the map from.
-    :arg mf: the fine mesh to map to.
-    :arg clgmaps: coarse lgmaps (non-overlapped and overlapped)
-    :arg flgmaps: fine lgmaps (non-overlapped and overlapped)
-    :returns: Two arrays, one mapping coarse to fine cells, the second fine to coarse cells.
+    Parameters
+    ----------
+    mc : MeshGeometry
+        The coarse mesh.
+    mf : MeshGeometry
+        The fine mesh, obtained by uniformly refining the non-overlapped
+        plex of ``mc``.
+    clgmaps : tuple
+        The coarse ``(non-overlapped, overlapped)`` point local-to-global maps.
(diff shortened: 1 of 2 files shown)


pbrubeck:
#5215 did not touch the uniform refinement code (which is assuming that halo cells are always ordered last). An adaptively refined mesh interleaves owned and halo cells (nowhere else in firedrake we assume any ordering of the owned/halo cells). Therefore breaking the assumption in the uniform refinement routine is the right fix.

firedrake/cython/mgimpl.pyx
```diff
@@ -381,64 +381,100 @@ def adaptive_parent_child_cell_maps(PETSc.DM coarse_dm,
...
-        # Drop values that did not exist on non-overlapped mesh
-        idx = cno.applyInverse(idx, PETSc.LGMap.MapMode.DROP)
-        co2n = co2n[idx]
+        idx = cno.applyInverse(idx, PETSc.LGMap.MapMode.MASK)
+        # idx[i] is where overlapped cell i lands, so scatter rather than
+        # slice: the surviving cells need not be the leading ones.
+        found = idx >= 0
+        permuted = np.empty(found.sum(), dtype=PETSc.IntType)
+        permuted[idx[found]] = co2n[found]
+        co2n = permuted
```
pbrubeck, line 469: This is the only code that needed to be changed. I wanted more comments explaining the three mappings involved in this code.

connorjward (approved):
This is good

PR #5323, opened 2026-08-04
Interpolate into a Boundary Quadrature space on a quadrilateral

pbrubeck:
Paired with **firedrakeproject/fiat#274** — https://github.com/firedrakeproject/fiat/pull/274

`FunctionSpace(quad_mesh, "Boundary Quadrature")` could be assembled but not
interpolated into:

```
NotImplementedError: How to tabulate TensorProductElement on FacetPointSet?
```

The element's dual basis names the points of every facet at once, and the
facets of a product cell lie in several directions, so those points have no
single product structure for a `TensorProductElement` to factor. fiat#274 fixes
that by dual evaluating one facet direction at a time — each direction *is* a
product, since `d(A x B) = dA x B  u  A x dB`.

## Here

- **`tsfc: propagate zeros through dual evaluation`** — apply to dual
  evaluation the zero handling assembly already does. Fold zero-valued
  `Literal`s into symbolic `Zero` first so gem's algebraic identities can
  propagate them; drop pairs whose expression collapsed to `Zero` once the
  blocks are contracted, since a `Delta`-cancelled cross-component block is a
  real zero contribution rather than one merely absent from the sum; and emit a
  `Zero` temporary as a zero-filled read-only array if one does reach loopy,
  as `gem.Zero` has no `.array` for the `Constant` branch to use.

- **`Test Boundary Quadrature on quadrilaterals`** — the quadrature element
  tests now run on a quadrilateral mesh as well as a triangular one, plus a new
  interpolation test checking that the result reproduces the expression at the
  points and that both cells sharing a facet see the same values there. Scalar,
  vector and tensor valued spaces each take a different path through dual
  evaluation, so all three are covered.

- **`DROP BEFORE MERGE`** — points the existing FIAT install step at
  `pbrubeck/tensor-facet-quadrature` so CI builds against fiat#274.

## Verification

Against the base FIAT commit the three new `quadrilateral` cases fail with the
`NotImplementedError` above; with fiat#274 the file is 25/25. Kernel cache was
cleared for both runs — with a warm cache the failing cases pass spuriously.

Serial `test_interpolate`, `test_interpolate_zany`, `test_quadrature` and
`test_interpolate_vs_project`: 106 passed, 0 failed. Parallel-marked tests were
not run locally.

Still unsupported, unchanged by this PR: hexahedra, which `check_element`
rejects for this family, and extruded meshes, where the facets are not all the
same shape so the flat facet rule does not exist.

### DIFF
--- a/tests/firedrake/regression/test_quadrature.py
+++ b/tests/firedrake/regression/test_quadrature.py
@@ -3,9 +3,9 @@ import pytest
 import numpy as np
 
 
-@pytest.fixture
+@pytest.fixture(params=[False, True], ids=["triangle", "quadrilateral"])
 def mesh(request):
-    return UnitSquareMesh(5, 5)
+    return UnitSquareMesh(5, 5, quadrilateral=request.param)
 
 
 def test_hand_specified_quadrature(mesh):
@@ -52,3 +52,24 @@ def test_quadrature_element(mesh, family, mat_type, diagonal):
         a = inner(u, v) * dx
 
     assemble(a, mat_type=mat_type, diagonal=diagonal)
+
+
+@pytest.mark.parametrize("space", [FunctionSpace, VectorFunctionSpace, TensorFunctionSpace])
+def test_boundary_quadrature_interpolation(mesh, space):
+    V = space(mesh, "Boundary Quadrature", 2)
+    x = SpatialCoordinate(mesh)
+    if V.value_shape == ():
+        expr = x[0]**2 + x[1]
+    elif len(V.value_shape) == 1:
+        expr = as_vector([x[0]**2, x[1]])
+    else:
+        expr = as_tensor([[x[0]**2, x[1]], [x[0]*x[1], 1 + x[0]]])
+
+    f = Function(V).interpolate(expr)
+
+    # The functionals are point evaluations, so interpolation is exact there
+    assert np.isclose(assemble(inner(f - expr, f - expr) * ds), 0)
+    assert np.isclose(assemble(inner(f('+') - expr('+'), f('+') - expr('+')) * dS), 0)
+    # The degrees of freedom on a facet are shared, so the cells on either
+    # side of one see the same values
+    assert np.isclose(assemble(inner(f('+') - f('-'), f('+') - f('-')) * dS), 0)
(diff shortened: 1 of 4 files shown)


tests/firedrake/regression/test_quadrature.py
```diff
@@ -52,3 +52,24 @@ def test_quadrature_element(mesh, family, mat_type, diagonal):
...
+        expr = as_tensor([[x[0]**2, x[1]], [x[0]*x[1], 1 + x[0]]])
+
+    f = Function(V).interpolate(expr)
+
+    # The functionals are point evaluations, so interpolation is exact there
+    assert np.isclose(assemble(inner(f - expr, f - expr) * ds), 0)
+    assert np.isclose(assemble(inner(f('+') - expr('+'), f('+') - expr('+')) * dS), 0)
+    # The degrees of freedom on a facet are shared, so the cells on either
+    # side of one see the same values
+    assert np.isclose(assemble(inner(f('+') - f('-'), f('+') - f('-')) * dS), 0)
```
pbrubeck:
```suggestion
```

tsfc/driver.py
```diff
@@ -396,7 +399,11 @@ def predicate(index):
     # and cancel the pairs that map to different components.
     pairs = [(variable, gem.optimise.contraction(expression))
              for variable, expression in pairs]
-    impero_c = impero_utils.compile_gem(pairs, return_indices)
+    # As in assembly, drop any pair whose expression collapsed to Zero: a
+    # cross-component block that a Delta has just cancelled above is a real
+    # zero contribution, not merely absent from the sum, and it must not
+    # reach codegen as a temporary.
```
pbrubeck:
```suggestion
```

tsfc/driver.py
```diff
@@ -346,6 +347,10 @@ def compile_expression_dual_evaluation(expression, ufl_element, *,
     else:
         coordinate_mapping = None
     evaluation, basis_indices = to_element.dual_evaluation(fn, coordinate_mapping)
+    # As in assembly, fold zero-valued Literals into symbolic Zero first, so
+    # that the algebraic identities built into gem (Product(_, Zero) -> Zero,
+    # etc.) can propagate them and later steps can prune them.
```
pbrubeck:
```suggestion
```

tsfc/loopy.py
```diff
@@ -236,7 +236,10 @@ def generate(impero_c, args, scalar_type, kernel_name="loopy_kernel", index_name
     data = list(args)
     for i, (temp, dtype) in enumerate(assign_dtypes(impero_c.temporaries, scalar_type)):
         name = "t%d" % i
-        if isinstance(temp, gem.Constant):
+        if isinstance(temp, gem.Zero):
+            initializer = numpy.zeros(temp.shape, dtype=dtype)
+            data.append(lp.TemporaryVariable(name, shape=temp.shape, dtype=dtype, initializer=initializer, address_space=lp.AddressSpace.LOCAL, read_only=True))
+        elif isinstance(temp, gem.Constant):
```
pbrubeck:
```suggestion
        if isinstance(temp, gem.Constant):
```

PR #5324, opened 2026-08-04
Rework PMGBase.coarsen() and add NonlinearVariationalProblem.rediscretise()

pbrubeck:
## Summary

Gives `NonlinearVariationalProblem` and the nullspace
basis classes a generic `reediscretise()` method, and reworks
`PMGBase.coarsen()` to build the coarse problem, `_SNESContext`, and
nullspaces through those `reconstruct()`/`rediscretise()` methods instead of hand-rolling
each piece and special-casing bc coarsening per `PMGPC`/`PMGSNES`.

Also replaces `PMGBase.reconstruct_degree`'s element-degree-shift
tree-walk with `finat.ufl`'s own native `FiniteElement.reconstruct(degree=...)`
support (firedrakeproject/fiat#273).

## Before / after

- Before: `PMGBase.coarsen()` hand-built the coarse `NonlinearVariationalProblem`
 and `_SNESContext`, and coarsened bcs via bespoke `coarsen_bcs`/
 `coarsen_bc_value` methods (with a separate per-bc cache) that special-cased
 `PMGPC` (bc value always 0) vs `PMGSNES` (eager `Function.interpolate()`).
- After: `NonlinearVariationalProblem.redisctretise()` and `_SNESContext.reconstruct()`
 do this generically for any caller, so `PMGPC` and `PMGSNES` coarsen bcs (and F)
 identically, and the PMGPC-only `cproblem.F = action(J, cu)` simplification is replaced with
 `ZeroBaseForm`.
- Before: element degree coarsening/refinement went through
 `PMGBase.reconstruct_degree`'s manual tree-walk over
 `EnrichedElement`/`TensorProductElement`/`MixedElement`/etc.
- After: callers use the element's own `reconstruct(degree=...)`.

### DIFF
--- a/.github/actions/install/action.yml
+++ b/.github/actions/install/action.yml
@@ -160,8 +160,6 @@ runs:
           --extra-index-url https://download.pytorch.org/whl/cpu \
           "./firedrake-repo[${{ inputs.deps }}]"
 
-        pip install -v --no-deps --ignore-installed git+https://github.com/firedrakeproject/fiat.git@pbrubeck/element-reconstruct-degree
-        pip install -v --no-deps --ignore-installed git+https://github.com/firedrakeproject/ufl.git@pbrubeck/zero-base-form-holes
         firedrake-clean
         pip list
 
--- a/.github/actions/install/action.yml
+++ b/.github/actions/install/action.yml
@@ -160,6 +160,7 @@ runs:
           --extra-index-url https://download.pytorch.org/whl/cpu \
           "./firedrake-repo[${{ inputs.deps }}]"
 
+        pip install -v --no-deps --ignore-installed git+https://github.com/firedrakeproject/fiat.git@pbrubeck/element-reconstruct-degree
         firedrake-clean
         pip list
 
--- a/.github/actions/install/action.yml
+++ b/.github/actions/install/action.yml
@@ -161,6 +161,7 @@ runs:
           "./firedrake-repo[${{ inputs.deps }}]"
 
         pip install -v --no-deps --ignore-installed git+https://github.com/firedrakeproject/fiat.git@pbrubeck/element-reconstruct-degree
+        pip install -v --no-deps --ignore-installed git+https://github.com/firedrakeproject/ufl.git@pbrubeck/zero-base-form-holes
         firedrake-clean
         pip list
 
--- a/firedrake/mg/ufl_utils.py
+++ b/firedrake/mg/ufl_utils.py
@@ -314,7 +314,7 @@ def reconstruct_nlvp(problem, self, coefficient_mapping=None):
     F = self(problem.F, self, coefficient_mapping=coefficient_mapping)
     J = self(problem.J, self, coefficient_mapping=coefficient_mapping)
     Jp = self(problem.Jp, self, coefficient_mapping=coefficient_mapping)
-    u = self(problem.u_restrict, self, coefficient_mapping=coefficient_mapping)
+    u = coefficient_mapping[problem.u_restrict]
 
     new_problem = firedrake.NonlinearVariationalProblem(
         F, u, bcs=bcs, J=J, Jp=Jp, objective=E, is_linear=problem.is_linear,
--- a/firedrake/mg/ufl_utils.py
+++ b/firedrake/mg/ufl_utils.py
@@ -314,7 +314,7 @@ def reconstruct_nlvp(problem, self, coefficient_mapping=None):
     F = self(problem.F, self, coefficient_mapping=coefficient_mapping)
     J = self(problem.J, self, coefficient_mapping=coefficient_mapping)
     Jp = self(problem.Jp, self, coefficient_mapping=coefficient_mapping)
-    u = coefficient_mapping[problem.u_restrict]
+    u = self(problem.u_restrict, self, coefficient_mapping=coefficient_mapping)
 
     new_problem = firedrake.NonlinearVariationalProblem(
         F, u, bcs=bcs, J=J, Jp=Jp, objective=E, is_linear=problem.is_linear,
--- a/firedrake/nullspace.py
+++ b/firedrake/nullspace.py
@@ -100,7 +100,7 @@ class VectorSpaceBasis(object):
         self.check_orthogonality()
         self._ad_orthogonalized = True
 
-    def reconstruct(self, function_space: firedrake.functionspaceimpl.WithGeometryBase) -> VectorSpaceBasis:
+    def rediscretise(self, function_space: firedrake.functionspaceimpl.WithGeometryBase) -> VectorSpaceBasis:
         r"""Reconstruct this basis on a new function space.
 
         Parameters
@@ -260,7 +260,7 @@ class MixedVectorSpaceBasis(object):
         self._bases = bases
         self._nullspace = None
 
-    def reconstruct(self, function_space: firedrake.functionspaceimpl.WithGeometryBase) -> MixedVectorSpaceBasis:
+    def rediscretise(self, function_space: firedrake.functionspaceimpl.WithGeometryBase) -> MixedVectorSpaceBasis:
         r"""Reconstruct this basis on a new mixed function space.
 
         Parameters
--- a/firedrake/nullspace.py
+++ b/firedrake/nullspace.py
@@ -1,7 +1,10 @@
+from __future__ import annotations
+
 import numpy
 
 from pyop2.mpi import COMM_WORLD
 
+import firedrake
 from firedrake import function
 from firedrake.logging import warning
 from firedrake.matrix import MatrixBase
@@ -97,13 +100,18 @@ class VectorSpaceBasis(object):
         self.check_orthogonality()
         self._ad_orthogonalized = True
 
-    def reconstruct(self, function_space):
+    def reconstruct(self, function_space: firedrake.functionspaceimpl.WithGeometryBase) -> VectorSpaceBasis:
         r"""Reconstruct this basis on a new function space.
 
         Parameters
         ----------
         function_space
             the new :class:`~.FunctionSpace`.
+
+        Returns
+        -------
+        VectorSpaceBasis
+            The basis vectors interpolated onto `function_space` and re-orthonormalized.
         """
         vecs = [function.Function(function_space).interpolate(vec) for vec in self._vecs]
         new_basis = VectorSpaceBasis(vecs, constant=self._constant, comm=self.comm)
@@ -252,13 +260,18 @@ class MixedVectorSpaceBasis(object):
         self._bases = bases
         self._nullspace = None
 
-    def reconstruct(self, function_space):
+    def reconstruct(self, function_space: firedrake.functionspaceimpl.WithGeometryBase) -> MixedVectorSpaceBasis:
         r"""Reconstruct this basis on a new mixed function space.
 
         Parameters
         ----------
         function_space
             the new :class:`~.FunctionSpace`.
+
+        Returns
+        -------
+        MixedVectorSpaceBasis
+            The bases reconstructed on the sub-spaces of `function_space`.
         """
         bases = []
         for V_, basis in zip(function_space, self._bases):
--- a/firedrake/nullspace.py
+++ b/firedrake/nullspace.py
@@ -276,7 +276,7 @@ class MixedVectorSpaceBasis(object):
         bases = []
         for V_, basis in zip(function_space, self._bases):
             if isinstance(basis, VectorSpaceBasis):
-                bases.append(basis.reconstruct(V_))
+                bases.append(basis.rediscretise(V_))
             else:
                 bases.append(function_space.sub(basis.index))
         return MixedVectorSpaceBasis(function_space, bases)
--- a/firedrake/nullspace.py
+++ b/firedrake/nullspace.py
@@ -97,6 +97,19 @@ class VectorSpaceBasis(object):
         self.check_orthogonality()
         self._ad_orthogonalized = True
 
+    def reconstruct(self, function_space):
+        r"""Reconstruct this basis on a new function space.
+
+        Parameters
+        ----------
+        function_space
+            the new :class:`~.FunctionSpace`.
+        """
+        vecs = [function.Function(function_space).interpolate(vec) for vec in self._vecs]
+        new_basis = VectorSpaceBasis(vecs, constant=self._constant, comm=self.comm)
+        new_basis.orthonormalize()
+        return new_basis
+
     @PETSc.Log.EventDecorator()
     def orthogonalize(self, b):
         r"""Orthogonalize ``b`` with respect to this :class:`.VectorSpaceBasis`.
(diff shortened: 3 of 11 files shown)


tests/firedrake/multigrid/test_p_multigrid.py
```diff
@@ -574,8 +574,11 @@ def test_pmg_transfer_piola(piola_mesh, family, degree, mixed, mat_type):
     Vf = FunctionSpace(piola_mesh, family, degree)
     if mixed:
         DG = FunctionSpace(Vf.mesh(), "DG", 2)
+        Vc0 = Vf.reconstruct(degree=1)
+        Vc = Vc0 * Vc0 * DG.reconstruct(degree=1)
```
connorjward: Can't this be `Vc = Vf.reconstruct(degree=[1, 1, 1])`?

pyproject.toml
```diff
@@ -23,7 +23,7 @@ dependencies = [
   # each Firedrake release to a specific UFL minor version (e.g. 2025.3.x)
   "fenics-ufl @ git+https://github.com/FEniCS/ufl.git@main",
   # TODO RELEASE
-  "firedrake-fiat @ git+https://github.com/firedrakeproject/fiat.git@main",
+  "firedrake-fiat @ git+https://github.com/firedrakeproject/fiat.git@pbrubeck/element-reconstruct-degree",
```
```suggestion
  "firedrake-fiat @ git+https://github.com/firedrakeproject/fiat.git@main",
```
so we don't forget

firedrake/preconditioners/pmg.py
```diff
@@ -195,70 +193,61 @@ def coarsen(self, fdm, comm):
...
-                          pre_jacobian_callback=fctx._pre_jacobian_callback,
-                          pre_function_callback=fctx._pre_function_callback,
-                          post_jacobian_callback=fctx._post_jacobian_callback,
-                          post_function_callback=fctx._post_function_callback,
-                          options_prefix=fctx.options_prefix,
-                          transfer_manager=fctx.transfer_manager)
+        # Coarsen the _SNESContext
+        cctx = fctx.reconstruct(cproblem, mat_type, pmat_type,
+                                appctx=cappctx,
+                                pre_jacobian_callback=fctx._pre_jacobian_callback,
```
connorjward: Shouldn't the `reconstruct` propagate these by default?

firedrake/preconditioners/pmg.py
```diff
@@ -195,70 +193,61 @@ def coarsen(self, fdm, comm):
...
-            if isinstance(val, dict):
-                cappctx[key] = self.coarsen_quadrature(val, fdeg, cdeg)
-            elif isinstance(val, ufl.Form):
-                cappctx[key] = _coarsen_form(val)
-            elif isinstance(val, ufl.classes.Expr):
-                cappctx[key] = ufl.replace(val, fine_to_coarse_map)
-
+        # Inherit mat_type from the fine _SNESContext
+        mat_type = None
+        pmat_type = None
```
connorjward, line 231: I'm not sure that this is more obvious than the previous code.

firedrake/variational_solver.py
```diff
@@ -143,6 +143,57 @@ def dirichlet_bcs(self):
...
+        `function_space`, with its value symbolically substituted through the
+        same coefficient mapping used to reconstruct the forms.
+
+        Parameters
+        ----------
+        function_space
+            The new function space for the solution.
+        form_compiler_parameters
+            The new form compiler parameters, defaults to the original parameters.
+        form_transform
```
connorjward, line 189: I'm not sure about this. Other `reconstruct` methods typically take all of the arguments and allow you to optionally swap them in.

firedrake/preconditioners/pmg.py
pbrubeck: hence I added the comment! If you wish I could make it more explicit

firedrake/preconditioners/pmg.py
pbrubeck: The callbacks are not propagated by reconstruct, I left them out on purpose. In fact the callback is an outer SNES thing, and arguably we should not be calling things that might assume that the solution lives on the finest level on the inner subsolvers. I propose to leave them out.

firedrake/preconditioners/pmg.py
connorjward: My question is why the change was made at all

firedrake/preconditioners/pmg.py
connorjward: It does not seem to be in the spirit of what I though `reconstruct` was.

pbrubeck: some things should not just be copied if other things change.

firedrake/preconditioners/pmg.py
pbrubeck: No good reason, just because reconstruct already handles these, so we don't have to read them in two places

firedrake/preconditioners/pmg.py
connorjward: I propose not calling this `reconstruct` then. It is confusing

firedrake/variational_solver.py
pbrubeck: Sure, I think the problem is that function_space is not an argument to the NonlinearVariationalProblem constructor, so maybe this method should not be called `reconstruct`.

firedrake/preconditioners/pmg.py
connorjward: In general that makes perfect sense, but previously the code was obvious, whereas now it needs a comment to explain what's going on.

firedrake/preconditioners/pmg.py
pbrubeck: Then `rediscretise` might be a better name

pbrubeck: I think the callback should be removed from the _SNESContext constructor and be attached later, we do this for the TransferManager

connorjward: > Then rediscretise might be a better name

Sure.

> I think the callback should be removed from the _SNESContext constructor and be attached later, we do this for the TransferManager

I don't like that. It means that you can have a `_SNESContext` that is somehow 'malformed'/uninitialised and it turns initialisation into a multi-step process that can go wrong. I believe that it is always a code smell to do this.

pbrubeck: Well, in fact it makes no harm propagating the callback, I'll do that.

tests/firedrake/multigrid/test_p_multigrid.py
pbrubeck: done

firedrake/solving_utils.py
```diff
@@ -297,15 +297,19 @@ def reconstruct(self, problem=None, mat_type=None, pmat_type=None, **kwargs):
...
+            sub_pmat_type=self.sub_pmat_type,
+            appctx=self.appctx,
+            options_prefix=self.options_prefix,
+            transfer_manager=self.transfer_manager,
+            pre_jacobian_callback=self._pre_jacobian_callback,
+            pre_function_callback=self._pre_function_callback,
+            post_jacobian_callback=self._post_jacobian_callback,
+            post_function_callback=self._post_function_callback,
+            pre_apply_bcs=self.pre_apply_bcs,
+            marking_callback=self._marking_callback
```
```suggestion
            marking_callback=self._marking_callback,
```

firedrake/preconditioners/pmg.py
```diff
@@ -195,70 +212,70 @@ def coarsen(self, fdm, comm):
...
-        cproblem = firedrake.NonlinearVariationalProblem(cF, cu, bcs=cbcs, J=cJ, Jp=cJp,
-                                                         form_compiler_parameters=fcp,
-                                                         is_linear=fproblem.is_linear)
+        cF = None
+        if not self.is_snes:
+            # PC-only levels don't assemble F, so a placeholder avoids coarsening RHS Cofunctions.
+            cF = ufl.ZeroBaseForm((test.reconstruct(function_space=cV),))
+
+        # Coarsen the problem
+        homogenize_bcs = not self.is_snes
```
pbrubeck:
```suggestion
        if self.is_snes:
            cF = None
            homogenize_bcs = False
        else:
            # PC-only levels don't assemble F, so a placeholder avoids coarsening RHS Cofunctions.
            cF = ufl.ZeroBaseForm((test.reconstruct(function_space=cV),))
            homogenize_bcs = True

        # Coarsen the problem
```

firedrake/preconditioners/pmg.py
```diff
@@ -195,70 +212,72 @@ def coarsen(self, fdm, comm):
         cdm = cV.dm
 
         fproblem = fctx._problem
-        fu = fproblem.u_restrict
-        cu = firedrake.Function(cV)
-
         fdeg = PMGBase.max_degree(fV.ufl_element())
         cdeg = PMGBase.max_degree(cV.ufl_element())
+        fcp = self.coarsen_quadrature(fproblem.form_compiler_parameters, fdeg, cdeg)
```
pbrubeck:
```suggestion
```

firedrake/preconditioners/pmg.py
```diff
@@ -195,70 +212,72 @@ def coarsen(self, fdm, comm):
...
-            if isinstance(val, dict):
-                cappctx[key] = self.coarsen_quadrature(val, fdeg, cdeg)
-            elif isinstance(val, ufl.Form):
-                cappctx[key] = _coarsen_form(val)
-            elif isinstance(val, ufl.classes.Expr):
-                cappctx[key] = ufl.replace(val, fine_to_coarse_map)
-
+        # Inherit mat_type from the fine _SNESContext
+        mat_type = None
+        pmat_type = None
```
pbrubeck, line 231:
```suggestion
        pmat_type = None
        fcp = self.coarsen_quadrature(fproblem.form_compiler_parameters, fdeg, cdeg)
```

.github/actions/install/action.yml
```diff
@@ -160,6 +160,8 @@ runs:
           --extra-index-url https://download.pytorch.org/whl/cpu \
           "./firedrake-repo[${{ inputs.deps }}]"
 
+        pip install -v --no-deps --ignore-installed git+https://github.com/firedrakeproject/fiat.git@pbrubeck/element-reconstruct-degree
+        pip install -v --no-deps --ignore-installed git+https://github.com/firedrakeproject/ufl.git@pbrubeck/zero-base-form-holes
```
```suggestion
```

firedrake/variational_solver.py
pbrubeck: @connorjward this is now in the same spirit of the other `reconstruct()` methods. I added every constructor kwarg plus 3 extra ones to manipulate Forms and bcs.

firedrake/mg/ufl_utils.py
```diff
@@ -314,7 +314,7 @@ def coarsen_callback(fine, coarse):
     F = self(problem.F, self, coefficient_mapping=coefficient_mapping)
     J = self(problem.J, self, coefficient_mapping=coefficient_mapping)
     Jp = self(problem.Jp, self, coefficient_mapping=coefficient_mapping)
-    u = coefficient_mapping[problem.u_restrict]
+    u = self(problem.u_restrict, self, coefficient_mapping=coefficient_mapping)
```
pbrubeck, line 317:
```suggestion
    u = coefficient_mapping[problem.u_restrict]
```

firedrake/variational_solver.py
connorjward: I don't think `form_transform` is in the spirit of `reconstruct`. If you called this method something else I would be fine with it.

pbrubeck: The problem is that ufl.Expr and ufl.BaseForm will often need extra reconstruction rules and one might need to pass a DAGTraverser to replace meshes, Measure metadata, etc. This is the cleanest approach we got so far to make `reconstruct` composable.

pbrubeck: Should I just hardcode the DAGTraverser to deal with every case?

connorjward: I think that in general we need to have a `DAGTraverser` that we can pass arbitrary Firedrake objects to. It would be really useful in a lot of circumstances I imagine. It's what I've come to realise is the best approach for doing things in pyop3.

One nice application would be checkpointing. We could have a visitor that takes in a function and traverses the `CoordinatelessFunction`, `WithGeometry`, `FunctionSpace`, `MeshGeometry`, `MeshTopology` etc.

firedrake/preconditioners/pmg.py
connorjward: This is a minor thing, so you don't necessarily have to do anything

firedrake/nullspace.py
```diff
@@ -97,6 +100,24 @@ def orthonormalize(self):
         self.check_orthogonality()
         self._ad_orthogonalized = True
 
+    def reconstruct(self, function_space: firedrake.functionspaceimpl.WithGeometryBase) -> VectorSpaceBasis:
```
connorjward: `rediscretise`?

firedrake/variational_solver.py
```diff
@@ -143,6 +144,116 @@ def dirichlet_bcs(self):
...
+        objective
+            The new objective form, defaults to the reconstructed original objective.
+        form_compiler_parameters
+            The new form compiler parameters, defaults to the original parameters.
+        is_linear
+            Whether the reconstructed forms are given in 'a == L' style,
+            defaults to the original problem's style.
+        coefficient_mapping
+            A dict mapping coefficients of the original problem into those
+            of the reconstructed problem.
```
connorjward: Docstring should describe the behaviour of this kwarg a little more

connorjward (changes requested):
Close

firedrake/nullspace.py
pbrubeck:
```suggestion
    def rediscretise(self, function_space: firedrake.functionspaceimpl.WithGeometryBase) -> VectorSpaceBasis:
```

firedrake/nullspace.py
```diff
@@ -239,6 +260,27 @@ def __init__(self, function_space, bases):
         self._bases = bases
         self._nullspace = None
 
+    def reconstruct(self, function_space: firedrake.functionspaceimpl.WithGeometryBase) -> MixedVectorSpaceBasis:
```
pbrubeck:
```suggestion
    def rediscretise(self, function_space: firedrake.functionspaceimpl.WithGeometryBase) -> MixedVectorSpaceBasis:
```

firedrake/preconditioners/pmg.py
```diff
@@ -283,81 +297,44 @@ def inject_state():
...
             add_hook(parent, setup=inject_state, call_setup=True)
 
-        interpolate = None
-        if fctx._nullspace or fctx._nullspace_T or fctx._near_nullspace:
-            interpolate, _ = cdm.createInterpolation(fdm)
-        cctx._nullspace = self.coarsen_nullspace(fctx._nullspace, cV, interpolate)
-        cctx._nullspace_T = self.coarsen_nullspace(fctx._nullspace_T, cV, interpolate)
-        cctx._near_nullspace = self.coarsen_nullspace(fctx._near_nullspace, cV, interpolate)
+        def _coarsen_nullspace(fine_nullspace):
+            return None if fine_nullspace is None else fine_nullspace.reconstruct(cV)
```
pbrubeck:
```suggestion
            return None if fine_nullspace is None else fine_nullspace.rediscretise(cV)
```

firedrake/variational_solver.py
pbrubeck: If you keep reading it tells you how coefficient_mapping gets consumed by ufl.replace/form_transform

pbrubeck: What else do you want to add?

connorjward: I find it confusing that you can pass `None` here but `coefficient_mapping` still gets things put into it. If someone is writing their own `form_transform` that is important information.

pbrubeck:
```suggestion
            of the reconstructed problem. This is used to update a fresh new dict with
            the replacement rules for the solution and form arguments.
```

pbrubeck: Is that enough?

connorjward: Yep

pbrubeck:
CI is green here

connorjward (approved):

PR #5326, opened 2026-08-05
Fix quadratic cost in adaptive mesh refinement's coarse/fine cell-map construction

pbrubeck:
## Summary

`adaptive_parent_child_cell_maps` resolved each fine cell's coarse parent with PETSc's `DMLabelGetValue`, called once per fine cell. `DMLabelGetValue` should resolve a point in O(1); instead it scans every stratum of the label linearly until it finds the one containing the queried point. `set_adaptive_parent_label` gives every coarse cell its own stratum value, so on an adaptively refined mesh the per-fine-cell loop was effectively O(nfine · ncoarse), not O(nfine). Profiling `adaptive_multigrid.py` under `-log_view` showed `refine_marked_elements` dominated not by PETSc's `refine_sbr` mesh transform itself, but by time attributed to this uninstrumented loop.

`adaptive_parent_child_cell_maps` now walks the label by stratum instead, with `DMLabelGetStratumSize`/`DMLabelGetStratumIS`, one call per coarse cell. That uses PETSc's O(1) value→stratum hash map and touches each fine cell exactly once, so the loop is O(nfine + ncoarse). Output is unchanged — cross-checked bit-for-bit against the original point-wise result on refined netgen meshes.

On the `AdaptiveMeshHierarchy` demo (first 10 Dörfler-marked refinement levels of an L-shaped netgen mesh), `adaptive_parent_child_cell_maps`'s share of time inside `refine_marked_elements` drops from 59% to 8%, and `refine_marked_elements`'s own total drops by 55%. The dominant cost left inside `refine_marked_elements` is now legitimately PETSc's own mesh-refinement transform.

## Collateral changes

`_adapt_marked_cells` and `refine_marked_elements` (`firedrake/adapt.py`) gained `PETSc.Log.Event` wrappers around each of their previously-unattributed sub-steps (label marking, `set_adaptive_parent_label`, the `adaptLabel` transform, `Mesh()` construction, `adaptive_parent_child_cell_maps`, re-marking, Netgen re-curving). Without them, `-log_view` attributed all of this work to `refine_marked_elements`'s own self-time rather than the call that actually spent it, which is what obscured this cost in the first place.

The `adaptive_multigrid` demo's recorded multigrid iteration counts and convergence plot predated #5215's DMPlexTransform-based mesh adaptivity; both are regenerated here against current `refine_marked_elements` output.

## Tests

`tests/firedrake/multigrid/test_adaptive_multigrid.py` already exercises `adaptive_parent_child_cell_maps` across firedrake/netgen meshes and process counts; it passes unchanged at nprocs 1/2/4.


firedrake/adapt.py
```diff
@@ -28,20 +29,23 @@ def _adapt_marked_cells(mesh, cell_marker):
...
 
     parameters = {"dm_plex_transform_type": "refine_sbr"}
     try:
         # options_prefix="" is essential
-        with petsctools.inserted_options(parameters=parameters, options_prefix=""):
-            new_dm = dm.adaptLabel(ADAPT_LABEL)
+        with PETSc.Log.Event("AdaptiveRefine: inserted_options"):
+            with petsctools.inserted_options(parameters=parameters, options_prefix=""):
+                with PETSc.Log.Event("AdaptiveRefine: adaptLabel"):
+                    new_dm = dm.adaptLabel(ADAPT_LABEL)
```
pbrubeck:
```suggestion
        with petsctools.inserted_options(parameters=parameters, options_prefix=""):
            with PETSc.Log.Event("AdaptiveRefine: adaptLabel"):
                new_dm = dm.adaptLabel(ADAPT_LABEL)
```

demos/adaptive_multigrid/adaptive_convergence.png
pbrubeck, line 1: The demo output was from the pre-#5215 era, so it was using netgen-driven adaptivity and needed an update. DMPlex adaptivity uses a different algorithm, thus `refine_marked_element` is now producing a different mesh (vs `netgen.Refine(adaptive=True)`) for the same marker.

connorjward (approved):

pbrubeck:
Thanks. CI failures are known. Merging

PR #5329, opened 2026-08-06
Restrict star patches to the adaptively refined region

pbrubeck:
> **Depends on PETSc** [petsc/petsc!9481](https://gitlab.com/petsc/petsc/-/merge_requests/9481), the tip of the stack [!9480](https://gitlab.com/petsc/petsc/-/merge_requests/9480) → [!9414](https://gitlab.com/petsc/petsc/-/merge_requests/9414) → !9481. The last commit points CI at that branch and must be dropped before merging.

`ASMStarPC` built its index sets in Python, taking each seed's star through `getTransitiveClosure` and gathering the dofs point by point. Every entity of the stratum got a patch, whether or not it lay anywhere near a refinement, so a smoother on an adaptively refined level relaxed the whole mesh to fix a corner of it.

`ASMStarPC` now has a single IS-construction mechanism, in `firedrake/cython/patchimpl.pyx`, used for every case rather than only the colored one: `create_star_points()` takes the star of each owned seed and `create_patch_ises()` gathers the dofs those points carry. The seeds may be restricted to the region that was genuinely refined, and the coloring then colors only those, so the patch count follows the size of the refined region instead of the size of the mesh.

### Changes

- `create_star_points()` and `create_patch_ises()` replace `build_star_indices`, which is gone. They reproduce the dof ordering the Python path produced, so a reordering requested with `pc_star_mat_ordering_type` still applies per star.
- `order_points()` no longer reverses its input, since both remaining callers pass the star already reversed. A reordering applies to each star on its own, so it survives a coloring that grouped several stars into one patch.
- `adapt.mark_refined_entities()` labels the cells whose parent was split into more than one child, plus their closure — exactly the entities whose star meets the refined region. It returns `None` for a mesh with no adaptive parent, which both preconditioners read as relaxing everywhere.
- `get_colors()` inserts its `MatColoring` type with `petsctools.inserted_options` under the `dm_plex_coloring_` prefix, matching how `_adapt_marked_cells` drives the transform type. The previous code set a bare `mat_coloring_type` globally and never removed it.
- A demo, `demos/patch/adaptive_patches.py.rst`, solves a cubic Poisson problem over a hierarchy that starts with two uniform refinements of a single square and grows three adaptive levels on top, and draws the patch seeds of each level colored by their color.

### Options

| Option | Effect |
| --- | --- |
| `pc_star_adaptive` | Restricts `ASMStarPC` to the entities whose star meets the refined region, building the label from `mesh.adaptive_cell_maps`. |
| `patch_pc_patch_adaptive` | The same restriction for `PatchPC`, through `PCPatchSetConstructLabel()`. |
| `pc_star_construct_label` / `_value` | Restricts `ASMStarPC` to any `DMLabel` already on the mesh's DMPlex, for a region the adaptive machinery did not produce. |
| `pc_star_use_coloring` | Unchanged in meaning, but colors only the selected entities when combined with either of the above. |

Both `adaptive` options are correct on a uniformly refined level as well as an adaptive one: every cell of a uniform level was split, so the restriction is a no-op there and the smoother relaxes everywhere. Nothing in the solver parameters has to know which kind of level it is looking at. Adding `-dm_plex_coloring_local` colors each process's own vertices without communicating, and usually with fewer colors.

### Patches across a hierarchy

The first two levels below are uniform refinements, so every vertex carries a patch. From the third level the patches sit only where the mesh was refined. Each disc is drawn half the way to the nearest vertex, so discs shrink with the mesh and never touch; within a level no two discs of the same color are adjacent, which is what makes the grouped patch operator block diagonal.

Patch seeds and their colors on each level

### Tests

Added to `tests/firedrake/regression/test_star_pc.py`: `ASMStarPC` and `PatchPC` take equal iteration counts on a corner-refined mesh, coloring does not change that count, restriction leaves strictly fewer patches, and the adaptive option on a mesh with no adaptive parent matches the unrestricted one. They run at 1 and 3 processes. The demo is registered in `tests/firedrake/demos/test_demos_run.py`.

AI was used to draft this change (Claude Code, Opus 5).


demos/patch/adaptive_patches.py.rst
```diff
@@ -0,0 +1,244 @@
+Patch relaxation on an adaptively refined hierarchy
+===================================================
+
+Contributed by `Pablo Brubeck <https://www.maths.ox.ac.uk/people/pablo.brubeckmartinez/>`_.
+
+The demo :doc:`Using patch relaxation for multigrid <poisson_mg_patches.py>` builds a
+vertex star patch around every vertex of every level. That is the right thing to do when
```
pefarrell, line 7: Clarify level means mesh refinement level

demos/patch/adaptive_patches.py.rst
```diff
@@ -0,0 +1,244 @@
...
+a level is a uniform refinement of the one below it, because then every cell of the level
+is new and the whole level needs relaxing.
+
+An adaptively refined level is different. Only the cells near the marked region were
+split; everywhere else the level is a copy of its parent, already resolved by the coarser
+levels, and relaxing there costs work without improving the error. This demo restricts
```
pefarrell, line 13: without substantially improving 

(it is effectively doing more relaxation sweeps, which is probably helping, but not much)

demos/patch/adaptive_patches.py.rst
```diff
@@ -0,0 +1,244 @@
...
+the patches to the region that was genuinely refined, and groups them by a coloring so
+that each level performs a handful of large block solves rather than thousands of tiny
+ones.
+
+We begin with the usual imports. The patches need an overlapping parallel decomposition
+that reaches a whole vertex star, which is what the ``distribution_parameters`` request. ::
+
+  import numpy
+  from scipy.spatial import cKDTree
```
pefarrell, line 22: I wouldn't call this one of the usual imports. What is it for? Maybe import it later nearer the plotting?

demos/patch/adaptive_patches.py.rst
```diff
@@ -0,0 +1,244 @@
...
+
+  dparams = {"overlap_type": (DistributedMeshOverlapType.VERTEX, 1)}
+
+Adaptivity composes with uniform refinement
+-------------------------------------------
+
+A :func:`~.MeshHierarchy` records parent child relations however its levels were made, so
+uniform and adaptive levels stack in the same object. We start from a single square cut
+into two triangles and refine it uniformly twice, which gives a 4x4 hierarchy of three
+levels, and the adaptive levels will grow on top of those: ::
```
pefarrell, line 38: Break up the sentence: "... three levels. The adaptive levels will grow on top of these"

demos/patch/adaptive_patches.py.rst
```diff
@@ -0,0 +1,244 @@
...
+
+  base = UnitSquareMesh(1, 1, distribution_parameters=dparams)
+  mh = MeshHierarchy(base, 2)
+  mesh = mh[-1]
+
+This composition is worth doing. Uniform levels are cheap to build and cheap to apply, and
```
pefarrell, line 44: "This composition of uniform and adaptive refinement is often useful"

demos/patch/adaptive_patches.py.rst
```diff
@@ -0,0 +1,244 @@
...
+they carry the low frequency error that a coarse grid must resolve; adaptive levels then
+add resolution only where the solution needs it. Starting straight from the coarsest mesh
+with adaptivity alone would leave the coarse region under-resolved for many refinements.
+
+The same relaxation option is correct on both kinds of level, and this is the point to
+take away. On a uniformly refined level every cell was split, so "the refined region" is
+the whole mesh and the restriction does nothing. On an adaptive level it is a small part
+of the mesh, and the restriction bites. Nothing in the solver parameters has to know which
```
pefarrell, line 52: "applies" instead of "bites"

demos/patch/adaptive_patches.py.rst
```diff
@@ -0,0 +1,244 @@
...
+kind of level it is looking at.
```
pefarrell, line 53: "is under consideration" instead of "it is looking at"

demos/patch/adaptive_patches.py.rst
```diff
@@ -0,0 +1,244 @@
...
+
+``snes_adapt_sequence`` asks for three rounds of solve, estimate, mark and refine, each
+adding a level to the hierarchy. Multigrid then runs over every level, uniform and
+adaptive alike.
+
+The relaxation is :class:`~.PatchPC`. Two options do the work discussed above:
+``patch_pc_patch_adaptive`` restricts the patches to the entities whose star meets the
+refined region, and ``patch_pc_patch_use_coloring`` groups patches with disjoint stars so
+each color is solved in one go. Adding ``-dm_plex_coloring_local`` on the command line
+colors each process's own vertices without communicating, which usually needs fewer
```
pefarrell, line 125: When would you not want this `-dm_plex_coloring_local` option?

demos/patch/adaptive_patches.py.rst
```diff
@@ -0,0 +1,244 @@
...
+          "patch_pc_patch_use_coloring": True,
+          "patch_pc_patch_save_operators": True,
+          "patch_pc_patch_sub_mat_type": "seqdense",
+          "patch_sub_ksp_type": "preonly",
+          "patch_sub_pc_type": "lu",
+      },
+      "mg_levels_0": {
+          "mat_type": "aij",
+          "ksp_type": "preonly",
+          "pc_type": "lu",
```
pefarrell, line 153: cholesky?

demos/patch/adaptive_patches.py.rst
```diff
@@ -0,0 +1,244 @@
...
+colors each process's own vertices without communicating, which usually needs fewer
+colors.
+
+Level 0 is prescribed separately through ``mg_levels_0``, which is the coarsest level
+whatever the hierarchy grows to, and is solved directly: ::
+
+  params = {
+      "mat_type": "aij",
+      "snes_adapt_sequence": 3,
+      "ksp_type": "cg",
```
pefarrell, line 134: add the `dm_plex_local_coloring`?

demos/patch/adaptive_patches.py.rst
```diff
@@ -0,0 +1,244 @@
...
+  problem = LinearVariationalProblem(a, L, uh, bcs=bcs)
+  solver = LinearVariationalSolver(problem, solver_parameters=params,
+                                   marking_callback=mark_cells)
+  solution = solver.solve()
+
+  hierarchy, level = get_level(solution.function_space().mesh())
+  print(f"Solved on {len(hierarchy)} levels, {solution.function_space().dim()} degrees of freedom")
+
+:class:`~.ASMStarPC` offers the same thing under ``pc_star_adaptive``, building the index
+sets itself rather than going through PCPatch.
```
pefarrell, line 166: This doesn't usefully summarise the difference with PCPatch. Maybe say "extracting the submatrices for the vertex stars from the assembled matrix, instead of building them matrix-free?

demos/patch/adaptive_patches.py.rst
```diff
@@ -0,0 +1,244 @@
...
+
+Looking at the patches
+----------------------
+
+The restriction and the coloring are both visible from Python. ``mark_refined_entities``
```
pefarrell, line 171: The `mark_refined_entities` function from `firedrake.adapt`

demos/patch/adaptive_patches.py.rst
```diff
@@ -0,0 +1,244 @@
...
+returns the label that ``patch_pc_patch_adaptive`` passes to PCPatch, and returns `None`
+for a level with no adaptive parent, which is how a uniformly refined level ends up
+relaxing everywhere. ``createColoringLabel`` then colors the marked vertices, under the
```
pefarrell, line 174: `plex.createColoringLabel` instead of `createColoringLabel`

pefarrell (commented):
Looks really good

PR #5331, opened 2026-08-06
speed up tests

leo-collins:


pbrubeck:
I have been working towards faster codegen for sum-factorization. Maybe there's no need to chop the degree on some of these tests

tests/firedrake/regression/test_fdm.py
```diff
@@ -119,7 +119,7 @@ def solve_riesz_map(problem, solver_parameters):
 @pytest.fixture(params=[2, 3],
                 ids=["Rectangle", "Box"])
 def mesh(request):
-    nx = 4
+    nx = 2
```
pbrubeck, line 122: I had  nx = 4 for a good reason. With nx=2 every cell touches the boundary.

docs/notebooks/10-sum-factorisation.py
```diff
@@ -64,7 +64,12 @@
 # We can create a hexahedral mesh by extruding a quadrilateral mesh.
 
 # %%
-mesh = ExtrudedMesh(UnitSquareMesh(10, 10, quadrilateral=True), 10)
+# Setup for faster test execution.
+import os
+if os.getenv("FIREDRAKE_CI") == "1":
+    mesh = ExtrudedMesh(UnitSquareMesh(2, 2, quadrilateral=True), 2)
+else:
+    mesh = ExtrudedMesh(UnitSquareMesh(10, 10, quadrilateral=True), 10)
```
pbrubeck, line 72: The mesh size in the actual demo can become 2. The purpose of demo is just to measure flops

docs/notebooks/09-hybridisation.py
```diff
@@ -93,7 +93,12 @@
 # So now we need to specify the number of refinements (say 4). Let's also set the mesh `degree` to be cubic. With this choice of coordinate space, we can better resolve the actual curvature of the sphere using bendy quadrilateral elements:
 
 # %%
-mesh = CubedSphereMesh(radius=R0, refinement_level=4, degree=3)
+# Use a coarser mesh when running CI tests.
+import os
+if os.getenv("FIREDRAKE_CI") == "1":
+    mesh = CubedSphereMesh(radius=R0, refinement_level=2, degree=3)
+else:
+    mesh = CubedSphereMesh(radius=R0, refinement_level=4, degree=3)
```
pbrubeck, line 101: We can just set the refinement_level inside the if-statement.

demos/vlasov_poisson_1d/vp1d.py.rst
```diff
@@ -296,11 +296,21 @@ We then use this to build a solver. ::
 We are getting close to the time loop. We set up some timestepping
 parameters. ::
   
-  T = 50.0 # maximum timestep
-  t = 0. # model time
-  ndump = 100 # frequency of file dumps
-  dumpn = 0 # dump counter
-  nsteps = 5000
+  import os
```
pbrubeck, line 299: It's surprising that one of our slowest demos is 1D.

demos/vlasov_poisson_1d/vp1d.py.rst
```diff
@@ -296,11 +296,21 @@ We then use this to build a solver. ::
 We are getting close to the time loop. We set up some timestepping
 parameters. ::
   
-  T = 50.0 # maximum timestep
```
pbrubeck, line 299: This is not accurate. The comment should say T = final time

demos/vlasov_poisson_1d/vp1d.py.rst
JHopeCollins: It's 2D (extruded) with a lot of timesteps.

leo-collins: 5,000 timesteps of a three-stage RK method with two solves in each stage will do that

JHopeCollins: Yeah. The splitting method has quite a small stability limit on the timestep.

demos/fast_diagonalisation/fast_diagonalisation_poisson.py.rst
```diff
@@ -24,8 +24,14 @@ extruded hexahedral meshes, so we must create an :func:`~.ExtrudedMesh`. ::
 
   from firedrake import *
 
-  base = UnitSquareMesh(8, 8, quadrilateral=True)
-  mesh = ExtrudedMesh(base, 8)
+  # Setup for faster test execution.
```
pbrubeck, line 27: This is polluting the demo.
1. Can we just do it one place? Can there be only one branch (no else statement, just set the demo parameters before the if statment).
2. Can we hide it so it does not appear in the html?

tests/firedrake/regression/test_fdm.py
leo-collins: So nx=3 is fine?

pbrubeck: I think if the test is slow even for nx=4, it means that we are setting the wrong solver options. The whole point of the FDM stuff is to ensure a very sparse operator, so I'd rather dig into the root of the problem and strengthen this test instead of just tweaking the mesh size.

Almost every test in this module uses the `mesh` fixture. I think only a few tests that rely on mumps are compute-dominated, and I think those can be speed up by tweaking mumps options, perhaps by changing `mat_ordering_type`.

tests/firedrake/regression/test_fdm.py
pbrubeck, line 19:
```suggestion
    "pc_type": "cholesky",
    "pc_factor_mat_solver_type": "mumps",
```

pbrubeck, line 41:
```suggestion
            "pc_star_sub_sub_pc_type": "cholesky",
            "pc_star_sub_sub_pc_factor_mat_solver_type": "cholmod",
```

pbrubeck, line 77:
```suggestion
            "pc_star_sub_sub_pc_type": "cholesky",
            "pc_star_sub_sub_pc_factor_mat_solver_type": "cholmod",
```

demos/adaptive_multigrid/adaptive_multigrid.py.rst
```diff
@@ -153,6 +153,12 @@ With these helper functions complete, we can solve the system iteratively. In th
 
    theta = 0.5
    refinements = 15
+
+   # Setup for faster test execution.
+   import os
+   if os.getenv("FIREDRAKE_CI") == "1":
+       refinements = 3
```
connorjward, line 160: I did this in https://github.com/firedrakeproject/firedrake/commit/154225ceffdda3503404ac791014010e1bc8f44f

PR #5337, opened 2026-08-08
Stop DG injection reading past a coarse cell's children

### LINKED ISSUE #5452: Unskip DG0 injection tests
Skipped in the PR: https://github.com/firedrakeproject/firedrake/pull/5451
Should be fixed by https://github.com/firedrakeproject/firedrake/pull/5337

pbrubeck:
# Description

Adaptive refinement gives coarse cells different numbers of children. `coarse_cell_to_fine_node_map`
pads the short rows with -1, and DG injection read out of bounds through those entries.

Fixes #5452

## Changes

- `coarse_cell_to_fine_node_map` fills each padded slot with a copy of the cell's first child, so
  every entry is a valid node.
- The DG injection kernel receives each coarse cell's child count and skips the padded slots.
  Without this, the copies would count a child twice.
- On extruded meshes, the map lists every layer of one child before the next child, so the padded
  slots always come last.
- `coarse_node_to_fine_node_map` now fills its padding on every rank, and raises on every rank if a
  coarse node has no candidates. Previously each rank decided alone, so ranks could disagree
  (@connorjward's point on #5288).
- The -1 padding rule is documented once, on `HierarchyBase.coarse_to_fine_cells`.

## Tests

- `test_dg_injection_conserves_mass`: injecting a random DG function must preserve its integral over
  every coarse cell. It fails if the kernel does not skip the padded slots.
- `test_dg_injection_conserves_mass_extruded`: the same check on an extruded hierarchy. It fails
  with the old layer ordering.

The adaptive test meshes grow to `UnitSquareMesh(4, 4)` and `UnitCubeMesh(2, 2, 2)`, so that child
counts vary across the mesh.


tests/firedrake/multigrid/test_adaptive_multigrid.py
```diff
@@ -356,6 +358,137 @@ def test_DG0(mh, operator):
...
+        # refined one has several. The macro-cell map pads the short rows.
+        # Only the levels that leave some cells alone exercise that padding.
+        padded |= bool((mh.coarse_to_fine_cells[level] < 0).any())
+
+        V_coarse = FunctionSpace(mh[level], family, degree)
+        V_fine = FunctionSpace(mh[level + 1], family, degree)
+
+        u_fine = Function(V_fine)
+        rng = np.random.default_rng(42 + mh[0].comm.rank)
+        u_fine.dat.data_wo[:] = rng.standard_normal(u_fine.dat.data_wo.shape)
```
pbrubeck:
```suggestion
        u_fine = rg.uniform(V_fine)
```

tests/firedrake/multigrid/test_adaptive_multigrid.py
```diff
@@ -356,6 +358,137 @@ def test_DG0(mh, operator):
...
+    holds the constants. Test that projection against the constant 1, and
+    the integral of the injected function over a coarse cell must equal the
+    integral of the fine function over that cell's children.
+
+    A random fine function makes this test bite. The step function that
+    `test_DG0` injects is constant on a unit domain. Injecting a constant
+    only checks that the children's volumes add up to the coarse cell's
+    volume. It passes even when the kernel integrates over the wrong set
+    of children.
+    """
```
pbrubeck, line 409:
```suggestion
    """
    rg = RandomGenerator(PCG64(seed=0))
```

firedrake/mg/utils.py
```diff
@@ -73,33 +133,66 @@ def coarse_node_to_fine_node_map(Vc, Vf):
...
+        # real entry from its own row. The injection kernel only reads
+        # through this map, and picks the candidate that matches the coarse
+        # node's physical location, so a repeated entry changes nothing.
+        #
+        # Every rank runs this fill. The partition decides which rows hold
+        # padding, so the fill must not depend on a rank-local test.
+        # The check covers the owned rows alone. coarse_to_fine_cells spans
+        # the owned coarse cells, so most halo rows hold no candidate at all.
+        # Those rows keep the zero filler, and injection visits owned nodes
+        # only, so nothing reads them.
```
pbrubeck:
```suggestion
```

firedrake/mg/utils.py
```diff
@@ -155,6 +148,66 @@ def coarse_cell_to_fine_node_map(Vc, Vf):
...
+    hierarchyf, levelf = get_level(Vf.mesh())
+    hierarchyc, levelc = get_level(Vc.mesh())
+
+    if hierarchyc != hierarchyf:
+        raise ValueError("Can't map across hierarchies")
+
+    hierarchy = hierarchyf
+    increment = Fraction(1, hierarchyf.refinements_per_level)
+    if levelc + increment != levelf:
+        raise ValueError("Can't map between level %s and level %s" % (levelc, levelf))
```
connorjward: f string

tests/firedrake/multigrid/test_adaptive_multigrid.py
```diff
@@ -362,6 +364,95 @@ def test_DG0(mh, operator):
...
+    Injection into a DG space is a cellwise L2 projection. Every DG space
+    holds the constants. Test that projection against the constant 1, and
+    the integral of the injected function over a coarse cell must equal the
+    integral of the fine function over that cell's children.
+
+    A random fine function makes this test bite. The step function that
+    `test_DG0` injects is constant on a unit domain. Injecting a constant
+    only checks that the children's volumes add up to the coarse cell's
+    volume. It passes even when the kernel integrates over the wrong set
+    of children.
```
connorjward: This comment suggests that there is another test that should be fixed, instead of adding another and a criticism of it?

connorjward (changes requested):

firedrake/mg/utils.py
pbrubeck:
```suggestion
        raise ValueError(f"Can't map between level {levelc} and level {levelf}")
```

tests/firedrake/multigrid/test_adaptive_multigrid.py
pbrubeck: The tests above use a simple step function, this test is much stronger and uses the random function. I agree, we should be fixing those

pbrubeck: Actually, we might be able to delete the old tests, this one covers them

pbrubeck: done

connorjward (approved):

pbrubeck:
Failures in `default` and `int64` are known. Docbuild AND `complex` are passing. Merging.

PR #5338, opened 2026-08-08
Agent tooling: sharper prose rules, a prose checker, and an `fdk` helper

pbrubeck:
# Description
AI-assisted (Claude Code)

## Goal

An AI coding agent should get rules it will actually follow. A rule that lives only as prose in a
document gets missed — an agent (or a person, skimming) reads past it. A rule a program checks
gets caught every time. This PR is the tooling behind that idea: sharper rules in `AGENTS.md`, a
checker that enforces the ones a machine can judge, and a helper command (`fdk`) that wraps the
things a session on this repo does over and over, the correct way, so nobody has to rediscover the
right shell incantation from scratch each time.

No Firedrake code changes — only `AGENTS.md` and three small tools under `.agents/`.

## What's in it

- **`AGENTS.md`**, rewritten as a short reference. It states each rule as an instruction — what to
  do, not why — and saves the reasoning for a handful of worked "wrong way / right way" examples at
  the end. Those examples are the part that demonstrably gets followed; everything else is trimmed
  to fit that.
- **`.agents/tools/check-prose.py`**. Checks the lines an edit just added (or a whole branch, with
  `--range`) against the handful of `AGENTS.md` rules a program can actually judge: numpydoc-only
  docstrings, short sentences, no leftover mention of code that got deleted, and no `hasattr`
  standing in for a plain setup flag. It can run itself automatically after every edit inside
  Claude Code, or by hand — same checks either way.
- **`.agents/tools/fdk`**. One command for the things a session on this repo needs repeatedly: run
  the tests at the right process count, lint the way CI does, find which test file already covers
  a piece of code, pull up one function out of a five-thousand-line file, see which branch each
  component package (PETSc, UFL, ...) is on, and more. Run `.agents/tools/fdk help` for the full
  list.
- **`.agents/skills/auditing-comments/`**. A companion habit for reviewing the comments an edit
  adds. Read what the comment claims, then check whether the code still says it: if it does, the
  comment is redundant and can go; if it does not, the code needs the fix, not the comment.

## Why it's shaped this way

Three fixes, each answering a gap the last one left. Prose rules in `AGENTS.md` alone were not
reliable — a rule only stated in a paragraph is easy for an agent to miss. So the part of those
rules a machine can judge moved into `check-prose.py`, which catches a violation the moment it's
written, not at review. A checker only helps if someone remembers to run it, so its commands — and
the rest of the repeated work — moved into `fdk`, one stable command instead of a different shell
line each time.

## Try it

```bash
.agents/tools/fdk help                              # every subcommand, with what each is for
.agents/tools/fdk test 1 tests/firedrake/regression  # tests at nprocs=1, one summary
.agents/tools/fdk prose --range main...HEAD          # the prose rules over this whole branch
```

## Notes for review

**Why these live in the repo, not a machine-local config.** In the repo they travel with the
branch and work for anyone pointing an agent at Firedrake — a person included, since both tools
are plain command-line programs that need only `bash`/Python and `git`.

**Why worked examples, not more prose, for the anti-patterns.** They're the part of `AGENTS.md`
that demonstrably gets followed. A rule stated only as prose has been violated by an agent while
sitting in the file at the time — which is what prompted writing a checker for the parts of it a
program can enforce, rather than trusting prose alone.

### DIFF
--- a/.agents/skills/auditing-comments/SKILL.md
+++ b/.agents/skills/auditing-comments/SKILL.md
@@ -84,8 +84,8 @@ work = coefficient.copy(deepcopy=True)
 result = assemble(inner(work, v) * dx)
 ```
 
-Assembly renumbers nothing. Once that is settled, the question is no longer how to word
-the comment. It is what the copy was for. Nothing writes to `work`, so the copy guards
+Assembly renumbers nothing. Once that is settled, the question is not how to word the
+comment. It is what the copy was for. Nothing writes to `work`, so the copy guards
 nothing:
 
 ```python
--- a/.agents/skills/auditing-comments/SKILL.md
+++ b/.agents/skills/auditing-comments/SKILL.md
@@ -1,0 +1,130 @@
+---
+name: auditing-comments
+description: Audits the comments and docstrings a branch adds, deleting the redundant ones and rewriting the code behind the rest. Use before opening a pull request or asking for review, after finishing a change, or when a reviewer says a comment is unclear, inaccurate, or misleading.
+---
+
+# Auditing comments
+
+A comment is a debt the reviewer pays. This audit settles it before a human reads the
+diff.
+
+The rule: **a comment that explains what the code does is a bug report against the
+code.** Fix the code. Rewriting the comment leaves the defect in place, and a reviewer
+who catches one inaccurate comment stops trusting the rest of the branch.
+
+## Workflow
+
+Copy this checklist and check items off:
+
+```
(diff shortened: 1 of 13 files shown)


dham:
I don't believe that AI scaffolding like this is. the right approach. It's a huge technical debt for unproven and doubtful benefit. 

Attempting to make our intrinsic workflow more LLM robust (e.g. by clearing caches from pytest) would be a better approach.

PR #5346, opened 2026-08-11
TSFC: cache the kernel by the repo state, not just the form signature

pbrubeck:
TSFC originally indexed its kernel caches by the form being compiled, not by the compiler that lowered it, so an editable was retrieving a stale kernel after tsfc, finat/FIAT/gem, ufl or loopy changes -- silently, as a wrong answer rather than a cache miss.

Add `tsfc.caching.codegen_key()`, fixed at import, and fold it into the three cache keys that index a kernel by its form signature (tsfc_interface, interpolation) plus Slate's own (which also stamps firedrake/slate/slac/, the generator that it owns outside TSFC). Also point `setup_cache_dirs()` at `sys.prefix` instead of `$VIRTUAL_ENV`, so an activated shell and a bare `venv/bin/python` share one cache.

# Description

### DIFF
--- a/firedrake/configuration.py
+++ b/firedrake/configuration.py
@@ -1,11 +1,12 @@
 """Replaces functionality from the removed `firedrake_configuration` module."""
 
 import os
+import sys
 from pathlib import Path
 
 
 def setup_cache_dirs():
-    root = Path(os.environ.get("VIRTUAL_ENV", Path.home())).joinpath(".cache")
+    root = Path(sys.prefix).joinpath(".cache")
     if "PYOP2_CACHE_DIR" not in os.environ:
         os.environ["PYOP2_CACHE_DIR"] = str(root.joinpath("pyop2"))
     if 'FIREDRAKE_TSFC_KERNEL_CACHE_DIR' not in os.environ:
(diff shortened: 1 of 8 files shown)


firedrake/configuration.py
```diff
@@ -1,11 +1,12 @@
 """Replaces functionality from the removed `firedrake_configuration` module."""
 
 import os
+import sys
 from pathlib import Path
 
 
 def setup_cache_dirs():
-    root = Path(os.environ.get("VIRTUAL_ENV", Path.home())).joinpath(".cache")
+    root = Path(sys.prefix).joinpath(".cache")
```
dham, line 9: test `sys.prefix` for writability. If it is then use it. If it is not writable then fall back to `Path.home()`

Put this in a separate PR. Possibly add support for `FIREDRAKE_CACHE_DIR` at the same time.

dham (changes requested):
Please do fix the folders.

We don't think this general process is robust enough. The alternative of supporting e.g. `FIREDRAKE_DELETE_CACHE_ON_IMPORT=1` which does what it says would be simpler and more robust.

pbrubeck:
Superseded by #5400

PR #5349, opened 2026-08-13
Make serial submesh construction linear in the number of points

haldaas:
# Description

`Submesh(mesh, comm=COMM_SELF)` cost O(P^2) in the number of mesh points P.
On three ranks a `UnitSquareMesh(181, 181)` needed 8.9 seconds to build one
serial submesh. Four times more cells cost about ten times more time.

Cellwise BDDCPC pays this cost, because local_mesh builds one serial submesh per process.

The cost was in `submesh_correct_entity_classes`. A submesh on `COMM_SELF` has
no neighbour, so it has no ghost points and no owned points, and every
point is core. The function set that state one point at a time.
`DMLabelSetValue` moves the stratum into a hash set and destroys its index
set. The `DMLabelHasPoint` of the next iteration reads the hash set back,
sorts it, and rebuilds the index set and the bit array. Each point
therefore cost O(k log k) in the size k of the core stratum so far.

Set the three strata in bulk instead. This is the same fix as the one in
PR #5326 for the coarse and fine cell maps.

`submesh_create` also built a temporary label that marked every cell, and
passed it to `DMPlexFilter`. `DMPlexFilter` selects every cell by itself when
it gets no label, so the label was unnecessary. `Submesh` now leaves
label_name as None for a codim-0 submesh of the whole mesh. A codim-1
submesh still passes the "depth" label, because `DMPlexFilter` with no label
selects cells, not facets.

Measured on three ranks, for `submesh_correct_entity_classes` alone:

    cells   before     after
      717    34 ms   0.12 ms
     2787   320 ms   0.43 ms
    11059  3214 ms   1.85 ms
    22023  8907 ms   3.80 ms

The whole submesh build at 11059 cells falls from 3256 ms to 71 ms.

Both paths produce the same submesh. On `UnitSquareMesh(128, 128)` with
three ranks, the submesh built with a label and the submesh built without
one have the same chart, the same subpoint index set, and the same
`pyop2_core`, `pyop2_owned`, `pyop2_ghost` and Face Sets strata.

Co-Authored-By: Claude Opus 5 <noreply@anthropic.com>


firedrake/cython/dmcommon.pyx
```diff
@@ -4001,10 +4001,10 @@ def submesh_create(PETSc.DM dm,
...
     subdim : int
         Topological dimension of the submesh
-    label_name : str
-        Name of the label
-    subdomain_id : int | Sequence
-        Values in the label
+    label_name : str | None
+        Name of the label, or `None` to select every cell
+    subdomain_id : int | Sequence | None
+        Values in the label, unused if ``label_name`` is `None`
```
connorjward, line 4008: I think it is much better API to catch
```py
            label_name = "depth"
            subdomain_id = subdim
```
here and then DTRT instead of allowing one to pass `None`. That way all of the logic is isolated into just this function.

connorjward (changes requested):

firedrake/cython/dmcommon.pyx
pbrubeck: I would argue that the right thing is to fix codim!=0, label=None in petsc

connorjward: I don't understand what you mean. What would PETSc need to fix?

pbrubeck: My point is why not reuse the petsc handling of default values? In this way we won't need to pass `label_name="depth"` at all. 

In the top level Submesh API there's some conditional logic on `codim`. I am not sure if PETSc only implemented label=None for codim=0.

connorjward: I don't mind. I just want to make sure that the branch
```py
        if subdim != dim:
            # DMPlexFilter handles label=None, codim=0.
            # Take an explicit label including all entities otherwise.
            label_name = "depth"
            subdomain_id = subdim
```
goes away.

pbrubeck: Yes, I think it might be possible to get rid of it entirely.

pbrubeck: Actually, no. DMPlexFilter does not take subdim, and this forces us to pass a label.

pbrubeck: So the question is whether we move `label_name="depth"` here or keep it in the top level Submesh API. I think it should live here, as we handle `subdomain_id="on_boundary"` and union of multiple labels here. And this aligns with your initial request of having all the logic in this function.

pbrubeck: I moved `label_name="depth"` into this function

firedrake/mesh.py
```diff
@@ -5023,9 +5023,6 @@ def Submesh(mesh, subdim=None, subdomain_id=None, label_name=None, name=None, ig
     if subdomain_id is None:
         if label_name is not None:
             raise ValueError("subdomain_id=None requires label_name=None.")
-        # Select all entities
-        label_name = "depth"
-        subdomain_id = subdim
```
connorjward, line 5028: Doesn't this break if we have `subdomain_id is None` but `subdim != dim`?

firedrake/cython/dmcommon.pyx
connorjward: My proposal was not to change the API of this function at all. I *think* there is an issue with the current design.

firedrake/mesh.py
pbrubeck: DMPlexFilter does not take subdim, only a label. If the label is None then it marks all cells on the subcommunicator. I just moved all the input parsing into the cython function.

firedrake/cython/dmcommon.pyx
pbrubeck: I think this is the right place to handle `label_name=None`. We split `subdomain_id="boundary"` into all facet subdomain ids, so this is the place to put the handling of defaults

pbrubeck: An argument in favour of moving this here: 
https://github.com/firedrakeproject/firedrake/blob/3aef245ecdbbd37417a5c0a1cac8396cc3383ec8/firedrake/mg/mesh.py#L29
Can pass `None, None` to this function instead of `"depth", tdim`

firedrake/mesh.py
connorjward: I now see what's going on. `subdomain_id` and `subdim` are both arguments to `submesh_create` so assigning one to the other is pointless and confusing. This should really get cleaned up.

connorjward (commented):
Why did a lot of the code from https://github.com/firedrakeproject/firedrake/pull/5349/changes/fe689869be371ebc39cffeb97215e1a707f4ac44 get removed? It looked relevant to the actual fix.

pbrubeck:
> Why did a lot of the code from [fe68986](https://github.com/firedrakeproject/firedrake/commit/fe689869be371ebc39cffeb97215e1a707f4ac44) get removed? It looked relevant to the actual fix.

The majority of the old diff just moved a block of code into an if-statement in order to avoid passing a DMLabel to DMPlexFilter for a very minor performance win that did not extended well to codim != 0. That wasn't fixing the nonlinear cost, but gave a tiny speedup by making the constant in front of the linear term smaller.

firedrake/cython/dmcommon.pyx
```diff
@@ -3999,12 +3999,12 @@ def submesh_create(PETSc.DM dm,
     ----------
     dm : PETSc.DM
         DMPlex representing the mesh topology
-    subdim : int
+    subdim : int | None
```
connorjward, line 4002: docstring needs amending

firedrake/cython/dmcommon.pyx
```diff
@@ -4018,6 +4018,35 @@ def submesh_create(PETSc.DM dm,
...
+        if label_name is None:
+            label_name = "exterior_facets"
+        elif label_name != "exterior_facets":
+            raise ValueError('subdomain_id="on_boundary" requires label_name="exterior_facets"')
+        subdomain_id = 1
+
+    if subdim is None:
+        subdim = dim
+    if subdim not in {dim, dim - 1}:
+        raise NotImplementedError(f"Found submesh dim ({subdim}) and parent dim ({dim})")
```
connorjward: That error message doesn't say how to fix it

firedrake/cython/dmcommon.pyx
connorjward: this could probably just point to the `Submesh` function. The interaction between these arguments is complex so you don't want to write them out twice.

connorjward (changes requested):
This is definitely better than before. I like having all the parsing live in one place.

firedrake/cython/dmcommon.pyx
```diff
@@ -3999,12 +3999,12 @@ def submesh_create(PETSc.DM dm,
     ----------
     dm : PETSc.DM
         DMPlex representing the mesh topology
-    subdim : int
+    subdim : int | None
         Topological dimension of the submesh
```
```suggestion
        Topological dimension of the submesh, or None to be infered from other kwargs.
        See :func:`~mesh.Submesh`.
```

firedrake/cython/dmcommon.pyx
```suggestion
        raise NotImplementedError(f"Submesh construction is only implemented for codimension 0 or 1. 
                                    Found submesh dim ({subdim}) and parent dim ({dim})")
```

firedrake/mesh.py
```diff
@@ -5003,34 +5003,6 @@ def Submesh(mesh, subdim=None, subdomain_id=None, label_name=None, name=None, ig
...
-            raise ValueError("subdomain_id=None requires label_name=None.")
-        # Select all entities
-        label_name = "depth"
-        subdomain_id = subdim
-    elif label_name is None:
-        if subdim == dim:
-            label_name = dmcommon.CELL_SETS_LABEL
-        elif subdim == dim - 1:
-            label_name = dmcommon.FACE_SETS_LABEL
     subplex = dmcommon.submesh_create(plex, subdim, label_name, subdomain_id, ignore_halo, comm=comm)
```
```suggestion
    subplex = dmcommon.submesh_create(mesh.topology_dm, subdim, label_name, subdomain_id, ignore_halo, comm=comm)
```

connorjward (dismissed):

firedrake/cython/dmcommon.pyx
```suggestion
                   subdim,
```

pbrubeck, line 4089:
```suggestion
                                             comm=comm)
    if subdm.getDimension() != subdim:
        raise RuntimeError(f"Found subplex dim ({subdm.getDimension()}) != expected ({subdim})")
```

firedrake/mesh.py
```suggestion
```

firedrake/mesh.py
```diff
@@ -5003,41 +5003,12 @@ def Submesh(mesh, subdim=None, subdomain_id=None, label_name=None, name=None, ig
...
-            label_name = dmcommon.CELL_SETS_LABEL
-        elif subdim == dim - 1:
-            label_name = dmcommon.FACE_SETS_LABEL
-    subplex = dmcommon.submesh_create(plex, subdim, label_name, subdomain_id, ignore_halo, comm=comm)
+    subplex = dmcommon.submesh_create(mesh.topology_dm, subdim, label_name, subdomain_id, ignore_halo, comm=comm)
 
     comm = comm or mesh.comm
     name = name or _generate_default_submesh_name(mesh.name)
     subplex.setName(_generate_default_mesh_topology_name(name))
     if subplex.getDimension() != subdim:
```
```suggestion
```

Issue #5218, opened 2026-07-06
BUG: Multigrid coarsening is not implemented for a symmetric Tensor coefficient/function


ioannisPApapadopoulos:
**Describe the bug**
Symmetric tensor coefficients throw an error when attempting multigrid coarsening.

**Steps to Reproduce**
```
from firedrake import *

base_mesh = UnitSquareMesh(8, 8)
hierarchy = MeshHierarchy(base_mesh, 1)
mesh = hierarchy[-1]

V = FunctionSpace(mesh, "CG", 1)
T = TensorFunctionSpace(mesh, "DG", 0, symmetry=True) # Fails
# T = TensorFunctionSpace(mesh, "DG", 0) # Passes

alpha = Function(T)
alpha.interpolate(as_tensor(((1.0, 0.0), (0.0, 1.0))))

u = Function(V)
v = TestFunction(V)


F = inner(dot(alpha, grad(u)), grad(v)) * dx - inner(Constant(1.0), v) * dx

bcs = DirichletBC(V, 0, "on_boundary")
problem = NonlinearVariationalProblem(F, u, bcs=bcs)

sp = {
    "snes_type": "ksponly",
    "mat_type": "aij",
    "ksp_type": "cg",
    "ksp_rtol": 1e-10,
    "ksp_atol": 1e-12,
    "ksp_converged_reason": None,
    "pc_type": "mg",
    "pc_mg_type": "full",
    "mg_coarse": {
        "ksp_type": "preonly",
        "pc_type": "lu",
    },
    "mg_levels": {
        "ksp_type": "chebyshev",
        "ksp_max_it": 2,
        "pc_type": "jacobi",
    },
}

solver = NonlinearVariationalSolver(problem, solver_parameters=sp)
solver.solve()

```

**Expected behavior**
```    Linear firedrake_1_ solve converged due to CONVERGED_RTOL iterations 12 ```

**Error message**
```
Traceback (most recent call last):
  File "/home/user/firedrake-release-30-05-2026/venv-firedrake/lib/python3.12/site-packages/firedrake/mg/kernels.py", line 365, in inject_kernel
    return cache[key]
           ~~~~~^^^^^
[...]

PR #5351, opened 2026-08-14
Fix DG injection for symmetric tensors

pbrubeck:
# Description
Bugfix, therefore targetting release

Fixes https://github.com/firedrakeproject/firedrake/issues/5218 Multigrid DG injection finds a mismatch between reference values and physical values (symmetry only stores the independent components as the reference values) 

Follows https://github.com/firedrakeproject/firedrake/pull/4764 which applied a similar fix: work on the reference values, and take into account their multiplicity when taking the inner product for the L2 projection.

### DIFF
--- a/firedrake/mg/kernels.py
+++ b/firedrake/mg/kernels.py
@@ -1,5 +1,6 @@
 import numpy
 import string
+from collections import defaultdict
 from pyop2 import op2
 from pyop2.utils import as_tuple
 from firedrake.utils import IntType, as_cstr, complex_mode, ScalarType
@@ -8,6 +9,7 @@ from firedrake.functionspaceimpl import FiredrakeDualSpace
 from firedrake.mg import utils
 
 from ufl.algorithms import estimate_total_polynomial_degree
+from ufl.classes import ReferenceValue
 from ufl.domain import extract_unique_domain
 
 import loopy as lp
@@ -444,7 +446,7 @@ def dg_injection_kernel(Vf, Vc, ncell):
                      scalar_type=parameters["scalar_type"])
 
     macro_context = fem.PointSetContext(**macro_cfg)
-    fexpr, = fem.compile_ufl(f, macro_context)
+    fexpr, = fem.compile_ufl(ReferenceValue(f), macro_context)
     X = ufl.SpatialCoordinate(Vf.mesh())
     C_a, = fem.compile_ufl(X, macro_context)
     detJ = ufl_utils.preprocess_expression(abs(ufl.JacobianDeterminant(extract_unique_domain(f))),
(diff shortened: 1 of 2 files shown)


JHopeCollins (changes requested):
I'm not familiar with the maths of this. Please can you add detail to the PR description for a) what the bug was and b) why this is the right fix.

pbrubeck:
> I'm not familiar with the maths of this. Please can you add detail to the PR description for a) what the bug was and b) why this is the right fix.

Done

firedrake/mg/kernels.py
```diff
@@ -513,8 +515,20 @@ def dg_injection_kernel(Vf, Vc, ncell):
...
     fexpr = gem.Indexed(fexpr, tensor_indices)
+    inner_prod = gem.Product(phi_c, fexpr)
+
+    if Vf.ufl_element().mapping() == "symmetries":
+        symmetry = Vf.ufl_element().symmetry()
+        multiplicity = defaultdict(int)
+        for idx in numpy.ndindex(Vf.value_shape):
+            idx = symmetry.get(idx, idx)
+            multiplicity[idx] += 1
+        block_scale = numpy.array([scale for idx, scale in multiplicity.items()])
```
JHopeCollins, line 529: Is this dict guaranteed to be ordered?

firedrake/mg/kernels.py
```diff
@@ -513,8 +515,20 @@ def dg_injection_kernel(Vf, Vc, ncell):
 
     phi_c = gem.Indexed(phi_c, argument_multiindex + tensor_indices)
     fexpr = gem.Indexed(fexpr, tensor_indices)
+    inner_prod = gem.Product(phi_c, fexpr)
+
+    if Vf.ufl_element().mapping() == "symmetries":
```
JHopeCollins, line 520: Please can you add a quick comment to this if-block for why we need to work on the reference value and count multiplicities?

firedrake/mg/kernels.py
```diff
@@ -444,7 +446,7 @@ def dg_injection_kernel(Vf, Vc, ncell):
                      scalar_type=parameters["scalar_type"])
 
     macro_context = fem.PointSetContext(**macro_cfg)
-    fexpr, = fem.compile_ufl(f, macro_context)
+    fexpr, = fem.compile_ufl(ReferenceValue(f), macro_context)
```
JHopeCollins: 1. Is `fexpr` used between here and the `inner_prod` definitions? If not then personally I would find it clearer to move this line down to there so it's defined right next to where it's used.

2. Are there any significant performance implications for using `ReferenceValue` for non-symmetric elements? If so, should the `ReferenceValue` definition be inside the if-guard?

JHopeCollins (commented):
Thanks for adding the extra detail.

I think this looks alright, I just have a couple of minor questions.

firedrake/mg/kernels.py
pbrubeck: `dict` and `defaultdict` maintain insertion order

firedrake/mg/kernels.py
pbrubeck: This function is the DG injection kernel codegen. The only elements that visit it are DG and its Vector/Tensor versions. ReferenceValue = PhysicalValue unless we have symmetry.

pbrubeck: The mathematically correct thing to do here is the `Reference` value. It is general enough for scalar/vector/tensor DG.

pbrubeck: We only store reference values, it is not like we are adding extra transformations in the code to extract a ReferenceValue.

firedrake/mg/kernels.py
pbrubeck:
```suggestion
    if Vf.ufl_element().mapping() == "symmetries":
        # Symmetric elements only store independent components.
        # The L2 inner product adds entrywise products of every component of the full tensor.
        # We work with the reference components so we need to scale by their multiplicities.
```

firedrake/mg/kernels.py
pbrubeck:
```suggestion
```

firedrake/mg/kernels.py
```diff
@@ -513,8 +515,23 @@ def dg_injection_kernel(Vf, Vc, ncell):
 
     phi_c = gem.Indexed(phi_c, argument_multiindex + tensor_indices)
```
pbrubeck, line 515:
```suggestion
    phi_c = gem.Indexed(phi_c, argument_multiindex + tensor_indices)
    fexpr, = fem.compile_ufl(ReferenceValue(f), macro_context)
```

JHopeCollins (approved):
LGTM

PR #5353, opened 2026-08-17
Fix broken test with exact comparisons

pbrubeck:
# Description
Fix for a broken test comparing bit-identity of assembled vs matrix-free KSP.

The test originally compared two `ksp_monitor_short` files, and now that it has been deprecated it no longer prints the short residual. https://gitlab.com/petsc/petsc/-/merge_requests/9434

Due to floating-point errors, CG residual history differs at every iteration for the assembled matvec and the matrix-free action. We only notice it now due to the deprecation of `ksp_monitor_short`.

This test shold not continue relying on `ksp_monitor_short` or any monitor printed into a file. The residual history now comes numerically from the KSP rather than a monitor file.

### DIFF
--- a/tests/firedrake/regression/test_matrix_free.py
+++ b/tests/firedrake/regression/test_matrix_free.py
@@ -59,23 +59,24 @@ def bcs(problem, V):
                                      "ilu",
                                      "lu"))
 @pytest.mark.parametrize("pmat_type", ("matfree", "aij"))
-def test_assembled_pc_equivalence(V, a, L, bcs, tmpdir, pc_type, pmat_type):
+def test_assembled_pc_equivalence(V, a, L, bcs, pc_type, pmat_type):
 
     u = Function(V)
 
-    assembled = str(tmpdir.join("assembled"))
-    matrixfree = str(tmpdir.join("matrixfree"))
+    def residuals(parameters):
+        u.assign(0)
+        problem = LinearVariationalProblem(a, L, u, bcs=bcs)
+        solver = LinearVariationalSolver(problem, solver_parameters=parameters)
+        solver.snes.ksp.setConvergenceHistory()
+        solver.solve()
+        return solver.snes.ksp.getConvergenceHistory()
 
     assembled_parameters = {"ksp_type": "cg",
-                            "pc_type": pc_type,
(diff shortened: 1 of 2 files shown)


pbrubeck:
Here's the broken CI job https://github.com/firedrakeproject/firedrake/actions/runs/31697065210/job/95330422799?pr=5348

tests/firedrake/regression/test_matrix_free.py
```diff
@@ -84,18 +85,13 @@ def test_assembled_pc_equivalence(V, a, L, bcs, tmpdir, pc_type, pmat_type):
...
-
-    with open(matrixfree, "r") as f:
-        f.readline()            # Skip over header
-        actual = f.read()
-
-    assert expect == actual
+    # The converged residual sits at the cancellation floor, so scale the
+    # tolerance by the initial residual rather than by each entry.
+    assert len(expect) == len(actual)
+    assert np.allclose(actual, expect, rtol=0, atol=1E-11*expect[0])
```
pbrubeck, line 94:
```suggestion
    assert len(expect) == len(actual)
    assert np.allclose(actual, expect, rtol=1E-14)
```

tests/firedrake/regression/test_matrix_free.py
```diff
@@ -84,18 +85,11 @@ def test_assembled_pc_equivalence(V, a, L, bcs, tmpdir, pc_type, pmat_type):
...
-        f.readline()            # Skip over header
-        expect = f.read()
-
-    with open(matrixfree, "r") as f:
-        f.readline()            # Skip over header
-        actual = f.read()
-
-    assert expect == actual
+    assert len(expect) == len(actual)
+    assert np.allclose(actual, expect, rtol=1E-14)
```
pbrubeck:
```suggestion
    assert np.allclose(actual, expect, rtol=1E-13)
```

pbrubeck:
The test broke because `ksp_monitor_short` (which does not print the residual in scientific notation) is now an alias to the standard `ksp_monitor`

pbrubeck:
KSP short monitors have been deprecated in https://gitlab.com/petsc/petsc/-/merge_requests/9434

connorjward (approved):

Issue #5107, opened 2026-05-13
BUG: Default optimisation level for GCC 15+ is -O2 not -O3


connorjward:
In https://github.com/firedrakeproject/firedrake/pull/5105 we set the default optimisation flag for GCC 15+ to be -O2. This is to fix numerical issues (i.e. NaNs and similar garbage) that occur when executing tricky interpolation kernels under -O3.

An example of a failing test is `tests/firedrake/output/test_io_function.py::test_io_function_base`.

Currently this workaround applies to all versions of GCC >= 15, when newer GCCs are released we should revisit this and see if we can set a maximum version constraint for this change.

manishpaulish:
Hi, I've been looking into this issue.
On GCC 15.2.0 / Ubuntu 26.04, the flags enabled by -O3 but not -O2 are:
-fgcse-after-reload
-fipa-cp-clone
-floop-interchange
-floop-unroll-and-jam
-fpeel-loops
-fpredictive-commoning
-fschedule-insns
-fsplit-loops
-fsplit-paths
-ftree-loop-distribution
-ftree-partial-pre
-funroll-completely-grow-size
-funswitch-loops
-fversion-loops-for-strides
Additionally, -fvect-cost-model changes from very-cheap to dynamic.
The most likely candidates for numerical instability in FEM assembly kernels are -ftree-loop-distribution (can reorder floating-point reductions across iterations) and -fpredictive-commoning (reuses computed values across loop iterations in ways that assume floating-point associativity).
I'm currently working on a reproducer to bisect these flags against the failing test. My current environment is ARM so x86 confirmation is pending. Happy to continue this investigation if useful.

connorjward:
It would be nice to have a more targeted fix for this, but it definitely isn't critical. Users are still able to set -O3 via an environment variable, and Firedrake performance is far more complex than just compiler flags, so I think the impact is going to be somewhat limited. If you're happy to spend some time on this though please be my guest.

manishpaulish:
Thanks for the context. Good to know users can override via environment variable. I'll continue the bisection on x86 to identify the specific flag and report back. Even if the performance impact is limited, having a targeted fix seems cleaner than a blanket version check.

manishpaulish:
Hi, update on the bisection investigation.
[...]

PR #5355, opened 2026-08-17
pyop2: add a safe_math option to opt out of relaxed floating point

manishpaulish:
# Description

PyOP2 compiles every generated kernel with relaxed floating point semantics, and there is no way to decline.

| compiler class | `_optflags` |
| --- | --- |
| `MacClangCompiler` | `-O3 -ffast-math -march=native` |
| `MacClangARMCompiler` | `-O3 -ffast-math -mcpu=apple-a14` |
| `LinuxGnuCompiler` | `-march=native -O3 -ffast-math` |
| `LinuxClangCompiler` | `-march=native -O3 -ffast-math` |
| `LinuxCrayCompiler` | `-march=native -O3 -ffast-math` |
| `LinuxIntelCompiler` | `-Ofast -xHost` |

`-ffast-math` implies `-fassociative-math`, `-ffinite-math-only`, `-funsafe-math-optimizations`, `-fno-signed-zeros` and `-fno-trapping-math`, and `-Ofast` implies `-fp-model=fast` on the Intel compilers. So a kernel that is sensitive to reassociation, or that relies on NaN or infinity propagating, has no recourse short of editing PyOP2.

### What this adds

`safe_math`, off by default, settable through `pyop2.configuration` or as `PYOP2_SAFE_MATH=1`. When it is set, `Compiler.optflags` returns `_safe_optflags` instead of `_optflags`.

The safe tuples are the existing ones with the relaxed maths removed and nothing else changed, so opting in to IEEE semantics does not silently also give up unrelated optimisations. `-O3` and `-march=native` stay. Intel is the exception worth calling out: dropping `-Ofast` on its own would still leave the default model in place, so the safe variant is `-O3 -xHost -fp-model=precise`.

`bugfix_cflags` is untouched. The GCC 15 `-O2` workaround still applies, independently, under either setting.

### Prior art

Devito exposes the same switch under the name `safe-math`, so the naming here is deliberately the same. FFCx passes no `-O` flags at all, so the question does not arise there.

### Relation to #5107

In #5107 I named `-ftree-loop-distribution` and `-fpredictive-commoning` as the likely culprits. That was wrong. Bisecting the thirteen flags `-O3` adds over `-O2`, on both gcc-13 and gcc-14, none of them change the result. `-fno-associative-math` does. And since PyOP2 appends the `-O2` workaround after `-O3`, the `-O2` wins but `-ffast-math` is still active, which would explain why that workaround does not always help.

I am not proposing `safe_math` as the fix for #5107 here. But if that reading holds it is the more targeted lever, and this is the prerequisite for reaching for it.

### Tests

`tests/pyop2/test_compilation.py`, parametrised over all six backends:

- the relaxed flag is still passed by default
- it is gone under `safe_math`, and no other spelling of it has crept in
- `-O2` or `-O3` survives, so the switch is not a choice between correct arithmetic and no optimisation
- Intel states `-fp-model=precise`
- `_safe_optflags` differs from `_optflags` only by the maths, checked statically so that a future edit to one tuple and not the other fails here rather than surprising a user

### DIFF
--- a/pyop2/compilation.py
+++ b/pyop2/compilation.py
@@ -228,6 +228,13 @@ class Compiler(ABC):
     _ldflags = ()
 
     _optflags = ()
+    _safe_optflags = ()
+    """Optimisation flags used when ``safe_math`` is set.
+
+    Should match ``_optflags`` with the relaxed floating point flags removed,
+    so that opting in to IEEE semantics does not silently also give up
+    unrelated optimisations.
+    """
     _debugflags = ()
 
     def __init__(self, extra_compiler_flags=(), extra_linker_flags=(), version=None, debug=False):
@@ -278,10 +285,25 @@ class Compiler(ABC):
         return self._ld
 
     @property
+    def optflags(self) -> tuple[str, ...]:
+        """Optimisation flags, honouring the ``safe_math`` configuration.
+
+        By default this includes ``-ffast-math`` or its equivalent, which
+        grants the compiler licence to reassociate floating point arithmetic
+        and to assume NaN and infinity do not occur. Kernels that are
+        sensitive to either need a way to decline; ``safe_math`` selects
+        ``_safe_optflags`` instead. Devito exposes the same switch under the
+        same name.
+        """
+        if configuration["safe_math"]:
+            return self._safe_optflags
+        return self._optflags
+
+    @property
     def cflags(self) -> tuple[str, ...]:
         return (
             *self._cflags,
-            *(self._debugflags if self._debug else self._optflags),
+            *(self._debugflags if self._debug else self.optflags),
             *self.bugfix_cflags,
             *self._extra_compiler_flags,
             *shlex.split(configuration["cflags"]),
@@ -291,7 +313,7 @@ class Compiler(ABC):
     def cxxflags(self) -> tuple[str, ...]:
         return (
             *self._cxxflags,
-            *(self._debugflags if self._debug else self._optflags),
+            *(self._debugflags if self._debug else self.optflags),
             *self.bugfix_cflags,
             *self._extra_compiler_flags,
             *shlex.split(configuration["cxxflags"]),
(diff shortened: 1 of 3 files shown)


connorjward:
The mechanism by which users set compiler flags is to set `PYOP2_CFLAGS`. I don't think we need another mechanism for that.

manishpaulish:
You are right, and I should have checked it before writing this.

`configuration["cflags"]` is spliced in last in `Compiler.cflags`, after `_optflags`, so `PYOP2_CFLAGS=-fno-fast-math` already reaches the compiler after `-ffast-math` and wins. I tested it rather than assuming, on a sum with catastrophic cancellation plus a NaN check:

```
gcc 13.3.0                        sum                  NaN seen
-O3                               0.1403082090444194   yes
-O3 -ffast-math                   4.6010414293078092   no
-O3 -ffast-math -fno-fast-math    0.1403082090444194   yes
-Ofast                            4.6010414293078092   no
-Ofast -fno-fast-math             0.1403082090444194   yes
```

clang 18.1.3 behaves identically. Both the reassociation and the `-ffinite-math-only` half come back, and `-O3` survives, which was the property I was actually trying to preserve with a separate `_safe_optflags` tuple. So the switch buys nothing that is not already there, and Intel is covered the same way since `-fp-model=precise` appended last wins over the one `-Ofast` implies.

Closing this. Thanks for the quick read, it saved me from maintaining a second spelling of an existing option.

The one thing the existing mechanism does not cover is that nothing says any of this out loud. Neither the `cflags` documentation nor anything else mentions that generated code is built with `-ffast-math`, so someone hitting a reassociation problem has no particular reason to suspect it or to reach for `PYOP2_CFLAGS`. If a sentence in the `cflags` docstring saying so would be welcome I am happy to send it, but it is a five line change and I will not open it unless you want it.

Separately, and I will follow up in #5107 rather than here: the bisection behind this PR is what makes me think `-fno-associative-math` is the trigger there rather than any `-O3` only transform, and that the existing `-O2` workaround is partly masked because it is appended after `-O3` so the `-O2` wins while `-ffast-math` stays active.

PR #5370, opened 2026-08-21
Fix NLVS hessian evaluation with Cofunction RHS

JHopeCollins:
Targetting #4638 

Fix second order adjoint calculations for variational solvers with Cofunctions.

Relevant UFL issues:
https://github.com/FEniCS/ufl/issues/395
https://github.com/FEniCS/ufl/issues/396


firedrake/adjoint_utils/blocks/assembly.py
```diff
@@ -63,8 +63,12 @@ def compute_action_adjoint(self, adj_input, arity_form, form=None,
             # -> Workaround: Apply action/adjoint numerically (using PETSc).
             if not isinstance(c_rep, firedrake.SpatialCoordinate):
                 # Symbolically compute: (dform/dc_rep)^* * adj_input
-                adj_output = firedrake.action(firedrake.adjoint(dform),
-                                              adj_input)
+                dform_adj = firedrake.adjoint(dform)
+                if isinstance(c_rep, firedrake.Cofunction):
+                    adj_output = firedrake.Action(dform_adj, adj_input)
+                else:
+                    adj_output = firedrake.action(dform_adj, adj_input)
```
angus-g, line 70: That's a slightly crazy API? What is the difference between the capitalised and non-capitalised forms?

JHopeCollins: You're right it is.

`action` is a free function that does some checks on the left and right arguments then creates an object of the class `Action` that is the actual symbolic description.

The checks in `action` mess up for various cases with `Cofunction`, but just creating the `Action` directly gives the right symbolic result. See https://github.com/FEniCS/ufl/issues/395.

angus-g: Right, makes sense!

PR #5374, opened 2026-08-24
Start using system HDF5 again for macOS

### LINKED ISSUE #5300: firedrake-configure on macOS does not use homebrew HDF5
In https://github.com/firedrakeproject/firedrake/pull/5299 we stopped using the system HDF5 package on macOS because it failed to build in combination with netCDF. When https://gitlab.com/petsc/petsc/-/work_items/1918 is resolved this should be undone.

connorjward:
Closes #5300

This reverts commit a7c1e93705ad9e1e393982dd71c9c34ff7295143.

This is now possible due to upstream PETSc fixes.


connorjward:
Has to go in `main` (and should actually *fix* the build error)

PR #5376, opened 2026-08-24
Constrain Cython to <3.3.0 due to petsc4py build error

connorjward:
See https://gitlab.com/petsc/petsc/-/work_items/1929 and https://gitlab.com/petsc/petsc/-/work_items/1929


connorjward:
This doesn't fix everything but it definitely gets us past the first error. Forcing through.

PR #5377, opened 2026-08-24
Cap islpy below 2026.2 on main

pbrubeck:
`islpy` 2026.2 removed the implicit conversion of a `BasicSet` to a `Set`. `loopy` 2025.2 — the newest release, and what `loopy>2024.1` resolves to — calls `BasicSet.make_disjoint()` in `count_insn_runs`, so every flop count in the test suite fails:

```
AttributeError: 'islpy._isl.BasicSet' object has no attribute 'make_disjoint'
```

loopy converts explicitly on main, but taking loopy from git is not an option yet: its `_AccessCheckMapper.map_subscript` asks for the dependencies of every shape axis without guarding `None`, and PyOP2 builds args with an unknown leading axis (`dat0` has `shape=(None, 2)` in `wrap_to_reference_coords`), so `firedrake-check` dies with `DependencyMapperWithReductionInames encountered invalid foreign object: None`.

So cap islpy at the last version that still converts. With islpy 2026.1 and loopy 2025.2, `tests/tsfc` and `tests/firedrake/regression/test_locate_cell.py` are green locally.

Lift the cap once loopy makes a release containing the explicit `to_set()`. #5378 does the same for `release`.

---
AI was used in preparing this contribution (Claude Code).

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -43,6 +43,10 @@ dependencies = [
   "requests",
   "scipy",
   "sympy",
+  # TODO RELEASE
+  # islpy 2026.2 stops converting a BasicSet to a Set implicitly, which loopy
+  # 2025.2 relies on. Lift once loopy releases the fix.
+  "islpy<2026.2",
   "islpy>=2025.1.5; sys_platform == 'darwin'",
 ]
 classifiers = [


connorjward (dismissed):
Can there be an equivalent PR into `release` that pins islpy?

pbrubeck:
> Can there be an equivalent PR into `release` that pins islpy?

Sure, I'll open it.

pbrubeck:
@connorjward It seems that the right fix is to cap islpy on main too. Do we want two separate PRs?

connorjward:
> @connorjward It seems that the right fix is to cap islpy on main too. Do we want two separate PRs?

Why is it needed on `main`? I would prefer on `main` to follow upstream branches.

francesco-ballarin:
While this might be a good short-term solution, can you work with upstream islpy/loopy to fix the issue?

I had to open a PR on islpy (which then generated a new release 2026.2.1) to fix compatibility with nanobind 3.0, and if islpy gets pinned it won't have that fix.
(I can patch my way out of it, sure, but only for the short term)

pbrubeck:
> Why is it needed on `main`? I would prefer on `main` to follow upstream branches.

We cannot pin islpy and loopy to `main`. There is a chain of breakages:

> Break 1 — loopy/statistics.py:1454 (loopy 2025.2) does set = set.make_disjoint() where set may be a BasicSet; that only worked via islpy's auto-promotion to Set, removed in islpy 2026.2.
> 
> Break 2 — pyop2/codegen/builder.py: shape = (None, arity) at line 50 (map arguments) and shape = (None, *arg.dim) at lines 861/879 (dat arguments), with Argument((None, 2), IntType, name="layers") at line 759. Those flow into pyop2/codegen/rep2loopy.py:797, which builds loopy.GlobalArg(name, dtype=dtype, shape=shape, strides=loopy.auto). loopy main then walks that shape in _AccessCheckMapper.map_subscript at loopy/check.py:771 — shape_deps.update(get_dependencies(shape_axis)) — and pymbolic raises on the None axis. check.py:1629 already guards this with if shape_axis is not None, and loopy/kernel/array.py:590 handles None axes too, so 771 is the only unguarded site. Adding the guard there is necessary but not sufficient: with it, tests/tsfc passes (369) but test_locate_cell still fails, so at least one further incompatibility sits behind it.

connorjward:
> > Why is it needed on `main`? I would prefer on `main` to follow upstream branches.
> 
> We cannot pin islpy and loopy to `main`. There is a chain of breakages:
> 
> > Break 1 — loopy/statistics.py:1454 (loopy 2025.2) does set = set.make_disjoint() where set may be a BasicSet; that only worked via islpy's auto-promotion to Set, removed in islpy 2026.2.
> > Break 2 — pyop2/codegen/builder.py: shape = (None, arity) at line 50 (map arguments) and shape = (None, *arg.dim) at lines 861/879 (dat arguments), with Argument((None, 2), IntType, name="layers") at line 759. Those flow into pyop2/codegen/rep2loopy.py:797, which builds loopy.GlobalArg(name, dtype=dtype, shape=shape, strides=loopy.auto). loopy main then walks that shape in _AccessCheckMapper.map_subscript at loopy/check.py:771 — shape_deps.update(get_dependencies(shape_axis)) — and pymbolic raises on the None axis. check.py:1629 already guards this with if shape_axis is not None, and loopy/kernel/array.py:590 handles None axes too, so 771 is the only unguarded site. Adding the guard there is necessary but not sufficient: with it, tests/tsfc passes (369) but test_locate_cell still fails, so at least one further incompatibility sits behind it.

I don't understand break 2

connorjward:
I think pinning islpy is a non-starter. We have to address this more fundamentally. If loopy `main` gets things to pass then we should pursue getting a new loopy release made. If it doesn't pass then we should fix whatever fails.

connorjward:
> While this might be a good short-term solution, can you work with upstream islpy/loopy to fix the issue?
> 
> I had to open a PR on islpy (which then generated a new release 2026.2.1) to fix compatibility with nanobind 3.0, and if islpy gets pinned it won't have that fix. (I can patch my way out of it, sure, but only for the short term)

I've opened a loopy issue https://github.com/inducer/loopy/issues/1039. Contrary to what I said before I think we do have to pin islpy because I don't know the loopy fix and we need CI back up.

connorjward:
Closing this one as https://github.com/firedrakeproject/firedrake/pull/5378 has just been merged into release and will be carried over ASAP.

PR #5384, opened 2026-08-25
check periodicity from unique mesh in ASMExtrudedStarPC

JHopeCollins:
For `MixedFunctionSpace` the `mesh` is a `MeshSequenceGeometry`, but this doesn't have the `extruded_periodic` attribute.
This attribute is held on `mesh.unique()` instead.


pbrubeck (approved):
LGTM

PR #5392, opened 2026-08-26
PR: Refactoring PyOP3 Codegen

SamSJackson:
# Refactoring `pyop3/lower/`

PR presents abstraction in the `lower/` design to allow for alternative code generation backends.

Principally, `loopy.py` has been broken into three separate files:
- `codegen.py` - context orchestrator which returns lowered IR back to translation layer 
- `context.py` - generic class and high-level traversal of axis trees to generate pyop3 expressions.
	(possible that class and traversal has too many responsibilities).

- `loopy.py` -implements backend-specific lowering from pyop3 to respective target IRs

This PR serves as a stepping stone to the introduction of MLIR as a code generation backend. 
A new `mlir.py` module will subclass `context.py` that will construct MLIR from PyOP3 expressions, using xDSL.


pyop3/insn/exec.py
```diff
@@ -56,6 +56,7 @@ class CompilerParameters:
 
     # TODO: handle these - need to build CompilerOptions
 
+    codegen: str = "loopy"
```
connorjward: Please document

pyop3/lower/codegen.py
```diff
@@ -0,0 +1,130 @@
...
+        # ex = expand_implicit_pack_unpack(ex)
+
+        # add external loop indices as kernel arguments
+        # FIXME: removed because cs_expr needs to sniff the context now
+        loop_indices = {}
+
+        for e in pyop3.collections.as_tuple(ex): # TODO: get rid of this loop
+            # context manager?
+            context.set_temporary_shapes(_collect_temporary_shapes(e))
+            _compile(e, loop_indices, context)
```
connorjward, line 96: This code is really disgusting but not your problem!

pyop3/lower/codegen.py
```diff
@@ -0,0 +1,130 @@
...
+    """
+    insn = op.preprocess()
+    function_name = "pyop3_loop"  # TODO: Provide as kwarg
+
+    if isinstance(insn, InstructionList):
+        cs_expr = insn.instructions
+    else:
+        cs_expr = (insn,)
+
+    if compiler_parameters.codegen == "loopy":
```
connorjward: I think I prefer 'backend' to 'codegen'

pyop3/lower/codegen.py
```diff
@@ -0,0 +1,130 @@
...
+        cs_expr = insn.instructions
+    else:
+        cs_expr = (insn,)
+
+    if compiler_parameters.codegen == "loopy": 
+        ContextClass = LoopyCodegenContext
+    elif compiler_parameters.codegen == "mlir": 
+        raise NotImplementedError("Class is still being implemented.") 
+
+    context = ContextClass(
```
connorjward:
```suggestion
        make_context = LoopyCodegenContext
    elif compiler_parameters.codegen == "mlir": 
        raise NotImplementedError("Class is still being implemented.") 

    context = make_context(
```

pyop3/lower/context.py
```diff
@@ -0,0 +1,441 @@
...
+            Display key properties of CodegenContext
+        '''
+        ctx = f"Domain: {str(self.domains)}\n\n"
+        ctx += f"Instructions: {str(self.instructions)}\n\n"
+        ctx += f"Arguments: {str(self.arguments)}\n\n"
+        ctx += f"Subkernels: {str(self.subkernels)}\n\n"
+        return ctx 
+
+
+# NOTE: Not a big fan of how compile sits in this file.
```
connorjward: Yeah I would expect to put it in `codegen.py`. This file can just contain the abstract class.

pyop3/lower/loopy.py
```diff
@@ -288,15 +268,478 @@ def within_inames(self, inames) -> None:
...
+            temp_offset_name = self.add_temporary("j")
+            temp_offset_var = pym.var(temp_offset_name)
+            self.add_assignment(temp_offset_var, offset_expr)
+            indices = extra_indices + (temp_offset_var,)
+        else:
+            indices = (offset_expr,)
+
+        return indices
+
+    def compile_standalone_function(
```
connorjward: When I designed the 'context' class I imagined that it would be entirely independent from any of the pyop3 'language'. It would only see the buffers. For example here I think we could amend the `add_function_call` to take in more arguments. The visitor for when we hit standalone function can then be
```py
args = [(a.buffer_view, spec.intent) for a, spec in zip(call.arguments, call.argspec, strict=True)]
context.add_function_call(call.function.code, args)
```
The crucial points are:
* The context never sees the `StandaloneCalledFunction` type
* The visitor for `StandaloneCalledFunction` can remain generic for different backends

pyop3/lower/loopy.py
```diff
@@ -288,15 +268,478 @@ def within_inames(self, inames) -> None:
...
+                for arg, spec in zip(call.arguments, call.argspec, strict=True)
+                if spec.intent in {READ, RW, INC, MIN_RW, MAX_RW}
+            ),
+        )
+
+        self.add_function_call(assignees, expression)
+        subkernel = call.function.code.with_entrypoints(frozenset())
+        self.add_subkernel(subkernel)
+
+    def compile_petsc_mat(
```
connorjward: Same, I'd keep it generic. You can just raise `NotImplementedError` for non-loopy

connorjward (changes requested):
Key point on an initial first pass is I would like to limit the scope of what the context objects know about.

pyop3/lower/codegen.py
SamSJackson: That's fine, I will switch. Was only wary that backend is used often but agree that it makes more sense.

connorjward: Terminology is really confusing around here. We have:
* compile
* codegen
* lower

which are all in some sense interchangeable. I really want to use "compile" as the single term for everything but this conflicts with the other sense of "compiler" (i.e. GCC etc).

I think in my main branch I will do the following renaming:
* `pyop3/lower` to `pyop3/compile`
* `pyop3/compile.py` to `pyop3/cc.py`

File naming isn't important for this PR, but I think I will make that change soon and you may hit git conflicts.

pyop3/lower/context.py
SamSJackson: Moved `_compile` to `codegen.py`. 

Only point of contention - and why it was not already there - is that `parse_loop_properly_this_time` uses `_compile` in the `context.py` file. As such, `_compile` now has to be imported. This must be a local import, to avoid a circular import.

It feels like a code smell but it does work. 
Possible steps for resolving the potential smell would be:
- Move `_compile` function to another separate file (I think this would bloat the directory)
- Rewrite `parse_loop_properly_this_time`

Else, we can leave it for later, it works fine. May just be a micro-optimisation.

connorjward: `parse_loop_properly_this_time` (which I think I would like you to rename) is another function that I don't think should belong to the context and ideally could be rewritten in a backend agnostic fashion.

pyop3/lower/loopy.py
SamSJackson: I have made some notable changes to follow this idea. 

- `_compile_array_assignment` and `_compile_loop` (formerly `_parse_loop...`) are now generic in `codegen.py`
- `_compile_petsc_mat` and `_compile_exscan` moved back to `codegen.py`.
- Context classes now have abstract methods `add_petsc_mat`, `add_exscan` for backend-specific additions.

`loopy.py` should now be working with buffer_views as opposed to pyop3 terminology, where possible.

Things that I am conscious of: 
- `add_petsc_mat` has bloated function definition to avoid passing pyop3 language.
- `compile_exscan` can almost be made agnostic but not until I develop an equivalent to `pym.substitute` for MLIR gen.

connorjward: > Context classes now have abstract methods add_petsc_mat, add_exscan for backend-specific additions.

I would suggest not bothering with this and just leave the 'generic' routines as they are (i.e. loopy specific). I don't think we care about having MLIR for either of them so just do:
```py
def _compile_petsc_mat(..., ctx):
  if not isinstance(ctx, LoopyCodegenContext):
    raise NotImplementedError

  # loopy-specific code
```

SamSJackson: This makes sense (and my life easier). Pushed these changes

pyop3/lower/context.py
```diff
@@ -0,0 +1,185 @@
...
+        *,
+        intent
+    ):
+        """
+        Determine indexing and lower buffer expression to respective IR
+        """
+        pass
+
+    @abstractmethod
+    def add_leaf_assignment(
```
connorjward: This would ideally be a generic visitor too.

pyop3/lower/codegen.py
```diff
@@ -0,0 +1,463 @@
...
+    Loop,
+    NonEmptyArrayAssignment,
+    NullInstruction,
+    StandaloneCalledFunction,
+    assignment_type_as_intent,
+)
+
+from pyop3.lower.loopy import LoopyCodegenContext
+
+# TODO: import other way around?
```
connorjward: It's not really a transform. It could just live in this file as this is the only place it's used.

pyop3/lower/loopy.py
```diff
@@ -288,15 +301,332 @@ def within_inames(self, inames) -> None:
...
+        intent,
+    ) -> pym.Expression: 
+        name_in_kernel = self.add_buffer(buffer_view, intent)
 
-    @property
-    def _depends_on(self):
-        return frozenset({self._last_insn_id}) - {None}
+        buffer = buffer_view.buffer
+        if isinstance(buffer, PetscMatBuffer):
+            buffer = buffer_view.denested.getPythonself().buffer
```
connorjward:
```suggestion
            buffer = buffer_view.denested.getPythonContext().buffer
```
typo

pyop3/lower/loopy.py
```diff
@@ -288,15 +301,332 @@ def within_inames(self, inames) -> None:
...
+            # a[off_a] = off_b < 0 ? a[off_a] : b[off_b]
+            offset_expr = _min_subscript_offset(rexpr)
+            # if there are no subcripts then the mask is pointless
+            if offset_expr is not None:
+                cond = pym.primitives.Comparison(offset_expr, "<", 0)
+                rexpr = pym.primitives.If(cond, lexpr, rexpr)
+
+        self.add_assignment(lexpr, rexpr)
+
+    def lower_expr(self, expr, iname_maps, loop_indices, *, intent=READ, paths=None) -> pym.Expression:
```
connorjward, line 416: I think this is a design issue. We need the codegen context to know *how* to lower expressions, because it differs between backends, but the logic should really be in another class (`LoopyExpressionLowerer(ExpressionLowerer)` or `LoopyExpressionCompiler(AbstractExpressionCompiler)`).

For now the easiest thing to do is just have:
```py
def _lower_expr(...):  # a free function
  ...

class LoopyCodegenContext(...):
  ...
  def lower_expr(self, ...):
    return _lower_expr(..., ctx=self)
```

connorjward (changes requested):
I think the structure is very close now. I am not striving for perfection given the rough nature of what you started with, but I think we should make sure that we get a decent chunk of the core design right.

pyop3/lower/loopy.py
SamSJackson: It does feel bloated in `loopy.py`. I was conscious of making too many classes.

I think I will revisit making another class once I have a working MLIR lowering implementation. 
For now, I have just made this a free function, as suggested.

connorjward: Yep making it a class can be a separate endeavour. Having written about a million visitor classes for pyop3 I have a lot of opinions about how it can be done. Let's postpone that for now. It's not even really necessary for your work.

connorjward:
Is this ready for another review?

pyop3/lower/transform.py
```diff
@@ -1,7 +1,18 @@
...
+from immutabledict import immutabledict as idict
+
+from pyop3.insn.base import (
+    AbstractAssignment,
+    Exscan,
+    InstructionList,
+    Loop,
+    NullInstruction,
+    StandaloneCalledFunction
+)
```
connorjward:
```suggestion
```

connorjward (approved):
I am happy with this now. I will figure out why `firedrake-check` is failing on CI and once that passes I will merge.

pyop3/lower/loopy.py
```diff
@@ -1053,134 +543,59 @@ def _(
...
-        context.add_assignment(temp_offset_var, offset_expr)
-        indices = extra_indices + (temp_offset_var,)
-    else:
-        indices = (offset_expr,)
-
-    return indices
-
-
-@_lower_expr.register(pyop3.expr.Conditional)
-def _(cond: pyop3.expr.Conditional, /, *args, **kwargs) -> pym.Expression:
```
connorjward, line 1178: This got dropped

pyop3/lower/loopy.py
```diff
@@ -1020,7 +511,7 @@ def _(or_: pyop3.expr.Or, /, *args, **kwargs) -> pym.Expression:
 
 @_lower_expr.register(pyop3.expr.Neg)
 def _(neg: pyop3.expr.Neg, /, *args, **kwargs) -> pym.Expression:
-    return -_lower_expr(neg.a, *args, **kwargs)
+    return _lower_expr(neg.a, *args, **kwargs)
```
connorjward:
```suggestion
    return -_lower_expr(neg.a, *args, **kwargs)
```
noooooo

pyop3/lower/codegen.py
```diff
@@ -0,0 +1,498 @@
...
+
+    axis = axis_tree.node_map[paths[-1]]
+    for component in axis.components:
+        new_paths = paths.copy()
+        new_paths[-1] = paths[-1] | {axis.label: component.label}
+        
+        if axis_tree.linearize(new_paths[-1], partial=True).size == 0: 
+            continue
+        
+        if component.local_size != 1:
```
connorjward:
```suggestion
        elif component.local_size != 1:
```

connorjward (changes requested):
In trying to make firedrake-check work I've hit some very bizarre caching problems. I have found a couple of very subtle typos that unfortunately aren't sufficient to fix the problems but if there are some then there may be more.

Sorry to be a pain but could you figure out how to get `firedrake-check` passing? I think this needs going over with a very fine toothed comb.

SamSJackson:
No problem. Very reasonable for me to make sure my own PR passes.

connorjward:
> No problem. Very reasonable for me to make sure my own PR passes.

I recommend just comparing the source code. The error is extremely obscure and related to caching.

connorjward (approved):

PR #5396, opened 2026-08-27
Scalable point location

leo-collins:
Introduces the `partition_rtree`, which is an R-tree of bounding boxes covering each partition of the mesh. This R-tree is replicated on all ranks by allgathering the bounding boxes. The bounding boxes on each rank are obtained by traversing down the local R-tree. Each rank may submit more than one bounding box to the partition_rtree (we want to do this to minimise candidates since the mesh partitions may not be nice and convex, and might even be disconnected). 

We choose the number of bounding boxes via a simple heuristic (we stop traversing when the total volume of the bounding boxes stop decreasing - see `_box_ratio_heuristic`). This should probably be improved to take into account the total number of bounding boxes so we don't blow up the allgather.

Constructing the partition rtree uses an allgather, although each rank normally submits O(10s) of boxes so I have not seen this kill us yet when testing up to 2048 ranks. If doing huge number of cores we might want to consider implementing some sort of hierarchy of partition rtrees.

Each rank queries this R-tree with the points local to that rank to determine the 'candidate' ranks for each point. We then build the `candidate_sf` which is a PETScSF mapping input points to their candidates. The local cell location process proceeds as usual, solving parallel conflicts with broadcasts/reductions via the candidate_sf.

This current implementation gets around 80% weak scaling efficiency in my tests up to 2048 cores.


firedrake/cython/rtree.pyx
```diff
@@ -101,3 +157,150 @@ def build_from_aabb(np.ndarray[np.float64_t, ndim=2, mode="c"] coords_min,
...
+                0,
+                mpi_comm,
+                &requests[k],
+            ))
+
+        sends_complete = nranks_to == 0
+        if sends_complete:
+            CHKERRMPI(MPI_Ibarrier(mpi_comm, &barrier_request))
+
+        while not barrier_complete:
```
connorjward, line 253: Names in this function are unclear. Currently looks like we only care about sending, not receiving.

firedrake/cython/rtree.pyx
```diff
@@ -101,3 +157,150 @@ def build_from_aabb(np.ndarray[np.float64_t, ndim=2, mode="c"] coords_min,
...
+            if message_ready:
+                CHKERRMPI(MPI_Get_count(&status, point_index_type, &count))
+                source_rank = status.MPI_SOURCE
+                received = np.empty(count, dtype=np.uintp)
+                CHKERRMPI(MPI_Recv(<void *>received.data, count, point_index_type, source_rank, 0, mpi_comm, MPI_STATUS_IGNORE))
+                recv_messages.append((source_rank, received))
+                nleaves += count
+
+            if not sends_complete:
+                CHKERRMPI(MPI_Testall(nranks_to, requests, &sends_complete, MPI_STATUSES_IGNORE))
```
connorjward:
```suggestion
                CHKERRMPI(MPI_Testall(nranks_to, send_requests, &sends_complete, MPI_STATUSES_IGNORE))
```

firedrake/mesh.py
```diff
@@ -3762,504 +4123,198 @@ def field(self, name: str) -> Generator[np.ndarray]:
...
+
+    owner_swarm_idxs = embedded_sf.broadcast(owner_swarm_idx_roots)[embedded_sf.leaf_indices]
+    swarm.set_halo_sf(
+        n_owned,
+        owner_ranks[halo_indices],
+        owner_swarm_idxs[halo_indices],
+    )
+
+    # Now we create the corresponding input-ordering swarm.
+    original_ordering_swarm = FiredrakeDMSwarm.create(
```
connorjward: Can this refactor be a separate PR?

connorjward (changes requested):

firedrake/mesh.py
leo-collins: The `FiredrakeDMSwarm` could be, but the internal functions have changed enough that it won't reduce the diff that much. If you think it would help though I will make a separate PR.

connorjward: If it has limited benefit I wouldn't worry about it

requirements-build.txt
```diff
@@ -16,3 +17,5 @@ rtree
 hatchling
 meson-python
 scikit_build_core
+maturin>=1.12,<2.0
+puccinialin
```
connorjward, line 22:
```suggestion
# TODO RELEASE: remove these
maturin>=1.12,<2.0
puccinialin
```

firedrake/mesh.py
```diff
@@ -2048,15 +2048,16 @@ def __init__(self, swarm, parentmesh, name, reorder, input_ordering_swarm=None,
             Name of the entity permutation (reordering); if `None`, automatically generated.
 
         """
-        if MPI.Comm.Compare(parentmesh.comm, swarm.comm.tompi4py()) not in {MPI.CONGRUENT, MPI.IDENT}:
-            ValueError("Parent mesh communicator and swarm communicator are not congruent")
+        if MPI.Comm.Compare(parentmesh.comm, swarm.dm.comm.tompi4py()) not in {MPI.CONGRUENT, MPI.IDENT}:
+            raise ValueError("Parent mesh communicator and swarm communicator are not congruent")
```
connorjward, line 2052: We just got `CommMismatchError` for this in https://github.com/firedrakeproject/firedrake/pull/5425

firedrake/mesh.py
```diff
@@ -2048,15 +2048,16 @@ def __init__(self, swarm, parentmesh, name, reorder, input_ordering_swarm=None,
...
         self._distribution_parameters = {"partition": False,
                                          "partitioner_type": None,
                                          "overlap_type": (DistributedMeshOverlapType.NONE, 0)}
+        self.swarm = swarm
         self.input_ordering_swarm = input_ordering_swarm
         self._parent_mesh = parentmesh
 
-        super().__init__(swarm, name, reorder, None, perm_is, distribution_name, permutation_name, parentmesh.comm)
+        super().__init__(swarm.dm, name, reorder, None, perm_is, distribution_name, permutation_name, parentmesh.comm)
```
connorjward: This now means that we have to keep track of the Firedrake swarm object as well as the PETSc one. I would prefer it if they were the same, either by keeping the current inheritance pattern or by adding `PETSc.DMSwarm` methods to `FiredrakeDMSwarm`. I think arguably the latter makes more sense because then (e.g.) `swarm.clone()` will give you back something with the right type.

Therefore I propose:
```suggestion
        super().__init__(swarm, name, reorder, None, perm_is, distribution_name, permutation_name, parentmesh.comm)
```

firedrake/mesh.py
```diff
@@ -2207,7 +2212,7 @@ def cell_parent_cell_list(self):
         """Return a list of parent mesh cells numbers in vertex only
         mesh cell order.
         """
-        with self.topology_dm.field("parentcellnum") as parentcellnum_field:
+        with self.swarm.field("parentcellnum") as parentcellnum_field:
```
connorjward: And if you did my suggestion then all of these changes would not be needed

firedrake/mesh.py
```diff
@@ -3736,15 +3890,220 @@ def VertexOnlyMesh(mesh, vertexcoords, reorder=None, missing_points_behaviour='e
...
+        self.remote = remote
+        self.input_ranks = remote[:, 0]
+        self.input_indices = remote[:, 1]
+        self.leaf_buffer_size = (
+            0 if len(leaf_indices) == 0 else int(leaf_indices.max()) + 1
+        )
+
+    @classmethod
+    @PETSc.Log.EventDecorator()
+    def candidate_sf(cls, parent_mesh: MeshGeometry, root_coordinates: np.ndarray) -> "VertexOnlyMeshSF":
```
connorjward, line 3950: docstring

firedrake/mesh.py
```diff
@@ -3736,15 +3890,220 @@ def VertexOnlyMesh(mesh, vertexcoords, reorder=None, missing_points_behaviour='e
...
+        if not isinstance(sf, PETSc.SF):
+            raise TypeError(f"`sf` must be a `PETSc.SF`, not a {type(sf).__name__}")
+
+        nroots, leaf_indices, remote = sf.getGraph()
+
+        leaf_indices.setflags(write=False)
+        remote.setflags(write=False)
+
+        self.sf = sf
+        self.nroots = nroots
```
connorjward: It is a style preference of mine to have these be `@property`s. If written like this it makes the constructor more complicated and also suggests to the reader that it is a valid operation to modify `self.nroots`, because it is an independent attribute.

Feel free to ignore if you disagree.

firedrake/mesh.py
```diff
@@ -3736,15 +3890,220 @@ def VertexOnlyMesh(mesh, vertexcoords, reorder=None, missing_points_behaviour='e
...
+        mask: np.ndarray,
+    ) -> "VertexOnlyMeshSF":
+        if mask.shape != (self.nleaves,):
+            raise ValueError("mask must contain one entry per leaf")
+        selected_leaf_indices = self.leaf_indices[mask]
+        return type(self)(self.sf.createEmbeddedLeafSF(selected_leaf_indices))
+
+
+class FiredrakeDMSwarm:
+    """A DMSwarm for use with :func:`VertexOnlyMesh`."""
```
connorjward: should explain why a PETSc one is inadequate

firedrake/mesh.py
```diff
@@ -3736,15 +3890,220 @@ def VertexOnlyMesh(mesh, vertexcoords, reorder=None, missing_points_behaviour='e
...
+    def __init__(self, dm: PETSc.DMSwarm, extruded: bool = False):
+        """Initialize a Firedrake DMSwarm.
+
+        Parameters
+        ----------
+        dm : PETSc.DMSwarm
+            The underlying PETSc DMSwarm.
+        extruded : bool
+            Whether the swarm is embedded in an extruded mesh.
+        """
```
connorjward: Put this in the main class docstring (that's what numpydoc suggests). It also means you don't have to write "Initialize a Firedrake DMSwarm.", which is trivially obvious.

firedrake/mesh.py
```diff
@@ -3736,15 +3890,220 @@ def VertexOnlyMesh(mesh, vertexcoords, reorder=None, missing_points_behaviour='e
...
+        if not isinstance(dm, PETSc.DMSwarm):
+            raise TypeError(f"`dm` must be a `PETSc.DMSwarm`, not a {type(dm).__name__}")
+        self.dm = dm
+        self.extruded = extruded
```
connorjward: should this be inferred from whether `parentcellbasenum` exists?

firedrake/mesh.py
```diff
@@ -2629,14 +2635,102 @@ def rtree(self):
...
+            coords_extent = coords_max - coords_min
+            coords_min = coords_min - tolerance * coords_extent
+            coords_max = coords_max + tolerance * coords_extent
         with PETSc.Log.Event("rtree_build"):
-            self._rtree = rtree.build_from_aabb(coords_min, coords_max)
-        return self._rtree
+            return rtree.build_from_aabb(coords_min, coords_max)
+
+    @PETSc.Log.EventDecorator()
+    def bounding_boxes_total_volume(self, bounding_boxes: np.ndarray):
```
connorjward: make a private attribute? can't imagine users wanting this

firedrake/mesh.py
```diff
@@ -2629,14 +2635,102 @@ def rtree(self):
...
+        else:
+            coords_extent = coords_max - coords_min
+            coords_min = coords_min - tolerance * coords_extent
+            coords_max = coords_max + tolerance * coords_extent
         with PETSc.Log.Event("rtree_build"):
-            self._rtree = rtree.build_from_aabb(coords_min, coords_max)
-        return self._rtree
+            return rtree.build_from_aabb(coords_min, coords_max)
+
+    @PETSc.Log.EventDecorator()
```
```suggestion
```
looks trivial, doesn't need to be profiled

firedrake/mesh.py
```diff
@@ -2736,25 +2863,40 @@ def locate_cells_ref_coords_and_dists(self, xs, tolerance=None, cells_ignore=Non
...
+            err = run_c(
+                self.coordinates._ctypes,
+                xs_data,
+                Xs_data,
+                ref_cells_dists,
+                cells_data,
+                owner_ranks_data,
+                npoints,
+                cells_ignore.shape[1],
+                cells_ignore,
```
connorjward, line 2883: I thought this was going?

connorjward (changes requested):
I am still quite confused by this PR. There is a lot going on.

firedrake/mesh.py
leo-collins: The ValueError wasn't being raised before so not sure how that could happen

firedrake/mesh.py
leo-collins: The locate_cells_.. methods are all public API that need this functionality

firedrake/cython/rtree.pyx
leo-collins: I have renamed some variables and added comments. Please say if you think it is still confusing

firedrake/mesh.py
connorjward: I can't imagine anyone using `cells_ignore` in their own code, but fine. Maybe raise a deprecation warning?

firedrake/mesh.py
leo-collins: I think inheritance will be easier. AbstractMeshTopology calls a bunch of cython routines which expect a PETSc object.

connorjward: Alternatively make it a DMShell?

leo-collins: That seems like an awful lot of effort for little gain

connorjward: I don't mind what approach you take. I just really think that `swarm` and `topology_dm` should be the same object. I would eventually like to drop `VertexOnlyMesh` and fold everything into a single `Mesh` class. We shouldn't be adding additional differences between the classes.

firedrake/mesh.py
leo-collins: I will open another PR to do this.

firedrake/mesh.py
leo-collins: But we create `parentcellbasenum` based on whether extruded is true or false...

connorjward: Is that true? In https://github.com/firedrakeproject/firedrake/pull/5396/changes#diff-7b656fa2d868bdf9f07513bd2565b7b30f1896c30352e3d8b92ae435ec3befd2R4115 it looks like you create the field independently.

As an alternative phrasing, do you ever actually inspect FiredrakeDMSwarm.extruded? A quick search makes me think that you don't.

leo-collins: Oh right that is `extruded` and not `self.extruded`...

firedrake/mesh.py
leo-collins: Made `FiredrakeDMSwarm` inherit from `PETSc.DMSwarm`.

firedrake/mesh.py
leo-collins: Undone these now

firedrake/mesh.py
leo-collins: added a docstring

firedrake/mesh.py
leo-collins: Made these all into properties

firedrake/mesh.py
leo-collins: Done

firedrake/mesh.py
leo-collins: Done

firedrake/mesh.py
leo-collins: Done

firedrake/mesh.py
leo-collins: Done

firedrake/mesh.py
leo-collins: Done

firedrake/cython/rtree.pyx
```diff
@@ -101,3 +157,174 @@ def build_from_aabb(np.ndarray[np.float64_t, ndim=2, mode="c"] coords_min,
...
+            n_points,
+            &send_ranks,  # unique ranks I am sending to
+            &send_offsets,  # offsets into `point_indices` of points I need to send
+            &point_indices,  # the points I need to send
+            &nranks_to,  # number of ranks I am sending to (len(send_ranks))
+        ) != Success:
+            raise RuntimeError("rtree_locate_points_grouped_by_id_unique failed")
+
+        # Each rank knows where to send to, but not where it is going to receive from.
+        # The following algorithm is the 'non-blocking consenus' algorithm (Hoefler et al.)
```
connorjward, line 226:
```suggestion
        # The following algorithm is the 'non-blocking consensus' algorithm (Hoefler et al.)
```

firedrake/cython/rtree.pyx
```diff
@@ -101,3 +157,174 @@ def build_from_aabb(np.ndarray[np.float64_t, ndim=2, mode="c"] coords_min,
...
+        # which is a sparse alternative to MPI_Alltoallv
+        if nranks_to != 0:
+            send_requests = <MPI_Request *>malloc(nranks_to * sizeof(MPI_Request))
+
+        # Post non-blocking synchronous sends. 'Synchronous' means that the send is not
+        # considered as complete by MPI_Test until the destination starts the matching receive.
```
connorjward, line 232: Why is this important?

firedrake/cython/rtree.pyx
connorjward: Really nice now. Thanks.

firedrake/locate.c
```diff
@@ -47,33 +53,34 @@ PetscErrorCode locate_cell_from_candidates(struct Function *f,
...
             PetscInt c = candidate / nlayers;
             PetscInt l = candidate % nlayers;
-            current_ref_cell_dist_l1 = (*try_candidate_xtr)(temp_ref_coords, f, c, l, x);
+            distance = (*try_candidate_xtr)(temp_ref_coords, f, c, l, x);
+            owner = cell_owner_ranks ? cell_owner_ranks[c] : 0;
         }
         else {
-            current_ref_cell_dist_l1 = (*try_candidate)(temp_ref_coords, f, candidate, x);
+            distance = (*try_candidate)(temp_ref_coords, f, candidate, x);
+            owner = cell_owner_ranks ? cell_owner_ranks[candidate] : 0;
```
connorjward, line 67: Is this a backwards compatibility thing? (`cell_owner_ranks` may be undefined?) If so a comment wouldn't go amiss.

firedrake/locate.c
```diff
@@ -47,33 +53,34 @@ PetscErrorCode locate_cell_from_candidates(struct Function *f,
...
         }
-
-        if (current_ref_cell_dist_l1 <= 0.0) {
-            /* Found cell! */
-            *cell_out = candidate;
+        /* Select owning cell by minimum L1 distance, with ties broken by highest owning rank. */
+        if (distance < best_distance || (distance == best_distance && owner > best_owner)) {
```
connorjward, line 70: I still worry about this. Consider an interval mesh like
```
global:
x-----x-----x
      o

rank 0:
x-----x
      o

rank 1:
      x-----x
      o
```
(which is allowed since we don't always need a full cell of overlap)

If `o` is the point we want to locate then rank 0 and rank 1 both have one entry in `cell_owner_ranks` - they will each think that they own the point.

firedrake/mesh.py
connorjward: I don't mind leaving this for now. Ignore me.

firedrake/mesh.py
```diff
@@ -2105,9 +2105,13 @@ def _renumber_entities(self, reorder):
...
-            with swarm.field(cell_id_name) as swarm_parent_cell_nums:
-                # Use kind = 'stable' to make the ordering deterministic.
-                perm = np.argsort(parent_renum_inv[swarm_parent_cell_nums.ravel() - pStart], kind='stable').astype(IntType)
+            with (
+                swarm.field(cell_id_name) as swarm_parent_cell_nums,
+                swarm.field("globalindex") as swarm_global_indices,
+            ):
+                parent_order = parent_renum_inv[swarm_parent_cell_nums.ravel() - pStart]
+                # sort by parent cell order, with ties broken by point global index
+                perm = np.lexsort((swarm_global_indices.ravel(), parent_order)).astype(IntType)
```
connorjward, line 2114: I'm surprised that this needs any global information at all. This numbering is normally done with only local information.

firedrake/mesh.py
connorjward: OK. Once you open that PR (or create an issue) then this can be resolved

firedrake/mesh.py
```diff
@@ -3736,15 +3888,261 @@ def VertexOnlyMesh(mesh, vertexcoords, reorder=None, missing_points_behaviour='e
...
+            yield unit
+        finally:
+            unit.Free()
+
+    def _check_arrays(
+        self,
+        root_values: np.ndarray,
+        leaf_values: np.ndarray,
+    ) -> None:
+        # TODO: make these collective
```
connorjward, line 4005: I worry that that would be quite expensive

firedrake/mesh.py
```diff
@@ -3768,498 +4166,192 @@ def field(self, name: str) -> Generator[np.ndarray]:
...
+    swarm.set_field(cell_id_name, cell_ids)
+    swarm.set_field("parentcellnum", swarm_parent_cells)  # store Firedrake parent-cell numbers
+    swarm.set_field("refcoord", reference_coords[swarm_indices])
+    swarm.set_field("globalindex", global_idxs_leaves[swarm_indices])
+    swarm.set_field("DMSwarm_rank", owner_ranks[swarm_indices])
+    swarm.set_field("inputrank", embedded_sf.input_ranks[swarm_indices].astype(IntType))
+    swarm.set_field("inputindex", embedded_sf.input_indices[swarm_indices].astype(IntType))
+    if parent_mesh.extruded:
+        swarm.set_field("parentcellbasenum", swarm_base_cells)
+        swarm.set_field("parentcellextrusionheight", swarm_extrusion_heights)
```
connorjward, line 4310: Why are some fields set in `create_with_fields` but not all?

firedrake/mesh.py
```diff
@@ -4308,461 +4400,129 @@ def _parent_extrusion_numbering(parent_cell_nums, parent_layers):
...
+    root_distance_min = np.full(nroots, np.inf, dtype=RealType)
+    candidate_sf.reduce(
+        np.where(keep, ref_cell_dists, np.inf),
+        root_distance_min,
+        op=MPI.MIN,
     )
+    keep &= ref_cell_dists == candidate_sf.broadcast(root_distance_min)
+
+    # multiple ranks may claim the minimum L1 distance. Break ties
+    # by choosing the highest numbered rank.
```
connorjward, line 4491: If you're doing this then I wonder if the parallel issue I flagged above is applicable. Could `owned_cell_ranks` just go? The C code could just see if a rank claims to own the point or not.

connorjward: I _think_ you only need rank information here, not there as well.

connorjward (changes requested):
Right, read the whole thing now. This is really heroic.

firedrake/mesh.py
leo-collins: `create_with_fields` registers our fields but doesn't set them.

connorjward: Is that the best API?

firedrake/locate.c
leo-collins: A rank doesn't decide at this point that they own the point. `locate.c` just determines whether the point lies in one of their owned cells (and L^1 distances and ref coords, etc.). 

The 'global' selection in `_parent_mesh_embedding` is what determines the owning rank. In this case the points both have a zero L^1 distance, so the owner is going to be rank 1 since we break ties by choosing the highest owning rank.

firedrake/mesh.py
leo-collins: Maybe we could add a `collective: bool` argument which is False by default? I found these checks quite useful for debugging purposes.

firedrake/locate.c
connorjward: Will it work if both ranks think that the other rank is the owner?

firedrake/mesh.py
leo-collins: I think we need to have rank information inside locate.c as long as we support `exclude_halos=False` (which is what `VertexOnlyMesh` uses).

In the future I want to get rid of this behaviour (i.e. always exclude halos from the final VOM) because a) the idea of a halo cell in a vom doesn't really make sense to me and we don't use this anywhere, and b) it substantially simplifies the vom construction code.

connorjward: Yeah I have *always* felt that halo points in a VoM was the wrong thing to do.

You need halos in order to build the correct stencils, but VoMs have no connectivity and so it's really pointless.

Happy to punt on this for now, I just always think it's worth discussing the design of these things.

firedrake/locate.c
leo-collins: If I'm reading your example correctly, rank 0 can see no cells of rank 1 and vice-versa. So I can't see how a rank could think the point lived in a cell owned by the other rank in this case. 

If we do have halos then this can happen, but by selecting the single highest owning rank we can never end up that situation.

firedrake/locate.c
leo-collins: Yes. Added a comment

firedrake/cython/rtree.pyx
leo-collins: Added a comment

firedrake/locate.c
connorjward: It doesn't apply to the example I gave earlier. Only if there are halos.

I guess my concern comes down to my mental model of floating point in parallel. Can we be sure that two ranks will compute the distance of a point on the cell boundary as exactly zero under any circumstance?

firedrake/mesh.py
leo-collins: I did this a while ago because `test_redistribution` was failing. `parent_order` can change between runs due to MPI messages arriving in different orders, so I added this tie break to make the renumbering reproducible.

connorjward: This seems like a code smell to me. I don't see why this should be deterministic. Is the relevant fail
```py
    assert np.allclose(vm1.coordinates.dat.data_ro, vm2.coordinates.dat.data_ro)
```
in which case why not sort the coordinates before comparing?

leo-collins: I don't know enough about mesh renumbering to know if it should be deterministic or not. I suppose with a `VertexOnlyMesh` we make no guarantees about the order the points end up in, and that the input-ordering vom provides the guarantee.

connorjward: OK I see what is going on. We do want to have deterministic numbering because of ensemble parallelism. However, I don't understand why you were seeing a non-deterministic result for `perm` using the previous approach. I wonder if this is masking a bug somewhere else.

leo-collins: The previous approach ordered the points by parent mesh cell, and since it was a stable sort if we had multiple points in a single cell then it would preserve their order. Now that points can arrive in different orders this isn't deterministic.

> We do want to have deterministic numbering because of ensemble parallelism.

Okay so how should I do this? Use the input-ordering indices to break ties?

connorjward: Ah this all makes sense now. Global index makes just as much sense as any other approach. Please just leave a more detailed comment explaining things.

PR #5400, opened 2026-08-27
Unified cache under sys.prefix

pbrubeck:
David asked, on #5346, for the `sys.prefix` writability check and the `FIREDRAKE_CACHE_DIR` unification to be split into their own PR:

> test `sys.prefix` for writability. If it is then use it. If it is not writable then fall back to `Path.home()`
>
> Put this in a separate PR. Possibly add support for `FIREDRAKE_CACHE_DIR` at the same time.

## Summary

- `setup_cache_dirs()` now checks `sys.prefix` for writability instead of reading `$VIRTUAL_ENV` (which is unset for a bare `venv/bin/python` invocation, and can point somewhere unwritable), falling back to `Path.home()` when it is not writable.
- Introduces `FIREDRAKE_CACHE_DIR` as a single root directory. `PYOP2_CACHE_DIR`, `FIREDRAKE_TSFC_KERNEL_CACHE_DIR`, and loopy's pytools-backed persistent cache (via `XDG_CACHE_HOME`) are all derived from it, so setting one variable moves all three caches together.
- Any of the four variables can still be set individually to override the derived default.


firedrake/configuration.py
```diff
@@ -1,12 +1,24 @@
...
+    root = Path(root)
     if "PYOP2_CACHE_DIR" not in os.environ:
         os.environ["PYOP2_CACHE_DIR"] = str(root.joinpath("pyop2"))
-    if 'FIREDRAKE_TSFC_KERNEL_CACHE_DIR' not in os.environ:
+    if "FIREDRAKE_TSFC_KERNEL_CACHE_DIR" not in os.environ:
         os.environ["FIREDRAKE_TSFC_KERNEL_CACHE_DIR"] = str(root.joinpath("tsfc"))
+    # loopy's persistent caches go through pytools, which only listens to
+    # XDG_CACHE_HOME (or platformdirs' default) for its cache location.
+    if "XDG_CACHE_HOME" not in os.environ:
+        os.environ["XDG_CACHE_HOME"] = str(root)
```
pbrubeck, line 23: I guess we don't need a `"loopy"` subdir because this is the cache home?

tests/firedrake/test_0init.py
```diff
@@ -15,6 +15,6 @@ def test_pyop2_custom_init():
 
 
 def test_pyop2_cache_dir_set_correctly():
-    root = Path(os.environ.get("VIRTUAL_ENV", "~")).joinpath(".cache")
+    root = Path(os.environ.get("FIREDRAKE_CACHE_DIR", "~"))
```
pbrubeck: I think that `FIREDRAKE_CACHE_DIR` is always set, so we might want to assert it is set.
```suggestion
    root = Path(os.environ.get("FIREDRAKE_CACHE_DIR", "~.cache"))
```

firedrake/configuration.py
connorjward: yeah think so

firedrake/configuration.py
```diff
@@ -1,12 +1,24 @@
...
 
 
 def setup_cache_dirs():
-    root = Path(os.environ.get("VIRTUAL_ENV", Path.home())).joinpath(".cache")
+    root = os.environ.get("FIREDRAKE_CACHE_DIR")
+    if root is None:
+        prefix = Path(sys.prefix)
+        base = prefix if os.access(prefix, os.W_OK) else Path.home()
+        root = str(base.joinpath(".cache"))
+    os.environ.setdefault("FIREDRAKE_CACHE_DIR", root)
```
connorjward:
```suggestion
```
I think we should minimise the extent to which we tweak the environment here to avoid surprises.

Not something to tackle in this PR but for instance I would ultimately like something like:
```py
root = ...

pyop3.init(cache_dir=root / "pyop3")  # instead of setting PYOP3_CACHE_DIR
```

tests/firedrake/test_configuration.py
connorjward, line 1: These tests should live alongside the other test you modified.

tests/firedrake/test_configuration.py
```diff
@@ -0,0 +1,67 @@
...
+
+CACHE_ENV_VARS = (
+    "FIREDRAKE_CACHE_DIR",
+    "PYOP2_CACHE_DIR",
+    "FIREDRAKE_TSFC_KERNEL_CACHE_DIR",
+    "XDG_CACHE_HOME",
+)
+
+
+def _clear_cache_env(monkeypatch):
```
connorjward: do we need to clear at the end of every test too?

connorjward (changes requested):

firedrake/configuration.py
pbrubeck: done

tests/firedrake/test_configuration.py
pbrubeck: done

tests/firedrake/test_configuration.py
pbrubeck: Yes, otherwise the state would leak between tests

pbrubeck: >  Replaced it with a clean_cache_env fixture that clears before and after.

tests/firedrake/test_0init.py
```diff
@@ -15,6 +16,71 @@ def test_pyop2_custom_init():
...
+CACHE_ENV_VARS = (
+    "FIREDRAKE_CACHE_DIR",
+    "PYOP2_CACHE_DIR",
+    "FIREDRAKE_TSFC_KERNEL_CACHE_DIR",
+    "XDG_CACHE_HOME",
+)
+
+
+@pytest.fixture
+def clean_cache_env(monkeypatch):
```
connorjward, line 32: This should save and restore the previous value of the variables, if set.

connorjward (changes requested):

tests/firedrake/test_0init.py
pbrubeck: Do we really need to if we are using `monkeypatch`? Let me double check

pbrubeck: > Confirmed with a quick check — the post-yield delenv loop we added is redundant.
> 
> How monkeypatch actually works: every monkeypatch.setenv/delenv call snapshots the variable's value before it touches it, onto an internal undo list. At test teardown, pytest automatically calls monkeypatch.undo(), which replays that list in reverse and restores each variable to its pre-test snapshot (or deletes it if it wasn't set originally). This happens regardless of what else touched os.environ in between — including setup_cache_dirs() writing to it directly — because the restore just unconditionally writes back the recorded original value.

connorjward: ahh ok

connorjward (dismissed):

PR #5405, opened 2026-08-28
Purge `MultiFunction`

leo-collins:


pbrubeck:
This is a very good change. But https://github.com/FEniCS/ufl/pull/504 now makes this DAGTraverser redundant

connorjward:
I've also dropped this class in pyop3.

PR #5408, opened 2026-08-28
Adding dtype to pyop3.expr.Expression

SamSJackson:
# Description 

The `dtype` property is being added to pyop3.expr.Expression for later work with MLIR code generation. 
It should aid with type inference.

Key assumptions/points:
- Type resolution for binary and ternary operators use numpy.result_type
    - This handles type promotion, and alike, as per rules here: [numpy docs](https://numpy.org/devdocs/reference/arrays.promotion.html#arrays-promotion)
    - An assumption for numerals is made such that:
        - integers are int32
        - float are float64
    - Another assumption is that LoopIndexVar dtype resolves to int32, as opposed to int64.

### DIFF
--- a/pyop3/expr/base.py
+++ b/pyop3/expr/base.py
@@ -789,6 +789,10 @@ def _(expr):
 def _get_dtype(expr: Any) -> np.dtype:
     return None
 
+@_get_dtypes.register(np.number)
+def _(expr) -> np.dtype: 
+    raise NotImplementedError(f"Not implemented dtype for: {type(expr)}")
+
 @_get_dtype.register(Expression)
 def _(expr: Expression) -> np.dtype:
     return expr.dtype 
--- a/pyop3/expr/base.py
+++ b/pyop3/expr/base.py
@@ -789,7 +789,7 @@ def _(expr):
 def _get_dtype(expr: Any) -> np.dtype:
     return None
 
-@_get_dtypes.register(np.number)
+@_get_dtype.register(np.number)
 def _(expr) -> np.dtype: 
     raise NotImplementedError(f"Not implemented dtype for: {type(expr)}")
 
--- a/pyop3/expr/base.py
+++ b/pyop3/expr/base.py
@@ -29,6 +29,11 @@ class Expression(Node, abc.ABC):
 
     @property
     @abc.abstractmethod
+    def dtype(self) -> np.dtype:
+        pass
+
+    @property
+    @abc.abstractmethod
     def _full_str(self) -> str:
         pass
 
@@ -218,6 +223,11 @@ class UnaryOperator(Operator, metaclass=abc.ABCMeta):
     child_attrs = ("a",)
 
     @property
+    def dtype(self) -> str:
+        a_dtype = _get_dtype(self.a)
+        return a_dtype
+
+    @property
     def _full_str(self) -> str:
         return f"{self.symbol}{as_str(self.a)}"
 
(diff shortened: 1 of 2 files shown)


pyop3/expr/base.py
```diff
@@ -575,6 +600,9 @@ class NameVar(TerminalExpression):
 
     name: Hashable
 
+    def dtype(self) -> np.dtype:
+        return TypeError
```
connorjward:
```suggestion
        raise TypeError("dtype of NameVars cannot be determined")
```

pyop3/expr/base.py
```diff
@@ -630,6 +658,10 @@ def local_max(self) -> numbers.Number:
     def local_min(self) -> numbers.Number:
         raise TypeError("not sure that this makes sense")
 
+    @property
+    def dtype(self) -> np.dtype:
+        raise TypeError("not sure that this makes sense")
```
connorjward:
```suggestion
        return IntType
```

pyop3/expr/base.py
```diff
@@ -716,6 +752,11 @@ def local_max(self) -> numbers.Number:
     def local_min(self) -> numbers.Number:
         raise TypeError("not sure that this makes sense")
 
+    @property
+    def dtype(self) -> np.dtype:
+        """ Returns int32 for indexing variable """
+        return np.dtype(np.int32)
```
connorjward:
```suggestion
        return IntType
```

pyop3/expr/base.py
```diff
@@ -745,6 +786,21 @@ def _(expr):
...
+
+@_get_dtype.register(Expression)
+def _(expr: Expression) -> np.dtype:
+    return expr.dtype 
+
+@_get_dtype.register(numbers.Number)
+def _(expr: numbers.Number) -> np.dtype:
+    # NOTE: Assumptions made about numerals here, i.e. float == float64, int == int32 
+    # However, type promotion/resolution happens on higher-level .dtype property. 
+    is_float = isinstance(expr, float)
```
connorjward: You're using `isinstance` inside single dispatch function. Just add more single dispatch handlers. As baseline fallbacks you should have
* `numbers.Real` to `pyop3.dtypes.RealType`
* `numbers.Integral` to `pyop3.dtypes.IntType`

but I'd also add a handler for `np.number` so we are as specific as possible

pyop3/expr/opaque.py
```diff
@@ -55,6 +55,11 @@ def buffer(self):
 
     # {{{ interface impls
 
+    @property
+    def dtype(self) -> np.dtype:
+        # NOTE: Seems anti-pattern to provide dtype for object that we know nothing about
```
connorjward: We do actually know things here. It comes from `buffer`. E.g. if `buffer` is a `PetscMatBuffer` then the dtype here is `Mat` (but that's not a numpy dtype). I'd suggest removing this comment and turning this into a `NotImplementedError`.

connorjward: As an alternative, you could make `dtype` not be an abstract property and instead do
```py
class Expression:
  @property
  def dtype(self):
    raise NotImplementedError(f"dtype not implemented for {utils.pretty_type(self)}")
```

connorjward (changes requested):

pyop3/expr/base.py
SamSJackson: Forgot about IntType, this should naturally be the fallback.

pyop3/expr/base.py
SamSJackson: Added some additional handlers.

For a generic np.number, dispatched to a `NotImplementedError`, whereas attached `np.floating` and `np.integer` to their respective number.real and number.integral

pyop3/expr/opaque.py
SamSJackson: Yep, this makes sense. I was thinking more `opaque.buffer_view.buffer.dtype` but I think this required realising the buffer too much.

Following your suggestion for `NotImplementedError` and removing abstract property.

connorjward (approved):

PR #5413, opened 2026-09-01
Fix init imports

### LINKED ISSUE #4736: Add `__all__` to Firedrake `__init__.py`
Since we do not define `__all__` in the `__init__.py`, running `from firedrake import *` pulls in over 550 objects. This includes some questionable entries such as `log`, `mpi` and - most confusingly - `mesh` (which is a Firedrake module). We really ought to control this.

nwng04:
# Description
Fixes #4736.

This PR adds `__all__` to the Firedrake `__init__.py` file, refining the imported objects when `from firedrake import *` is used. 

**Removed Objects:**
['slope_limiter', 'eigensolver', 'parloops', 'function', 'matrix_free', 'configuration', 'hdivcurl', 'mixedelement', 'PETSC_SUPPORTED_VERSIONS', 'ensemble', 'linear_solver', 'elementlist', 'setup_cache_dirs', 'checkpointing', 'utility_meshes', 'enrichedelement', 'functionspaceimpl', 'output', 'finiteelement', 'progress_bar', 'halo', 'slate', 'dmhooks', 'tsfc_interface', 'logging', 'ufl_expr', 'randomfunctiongen', 'projection', 'solving_utils', 'pointeval_utils', 'solving', 'cython', 'mg', 'functionspace', 'brokenelement', 'pyplot', 'functionspacedata', 'init_petsc', 'adjoint_utils', 'deflation', 'pointquery_utils', 'restrictedelement', 'constant', 'external_operators', 'tensorproductelement', 'matrix', 'norms', 'finiteelementbase', 'extrusion_utils', 'mesh', 'cofunction', 'exceptions', 'embedding', 'variational_solver', 'firedrake', 'interpolation', 'nullspace', 'bcs', 'formmanipulation', 'preconditioners']

**Added Object:** 'utils'
This object is added because 9 Firedrake tests use it when testing.


firedrake/__init__.py
```diff
@@ -45,43 +45,92 @@ def init_petsc():
     del atexit
 del petsc
 
+import ufl as _ufl  # Required for ufl.__all__ to be picked up by the compiler
 from ufl import *  # noqa: F401
+__all__ = list(_ufl.__all__)
+del _ufl
+
 from finat.ufl import *  # noqa: F401
```
connorjward: `finat.ufl.__all__`

connorjward (changes requested):
```py
utils = SpikedModule("utils")
__all__ += ["utils"]

class SpikedModule:
  def __getattr__(self, key):
    raise SomeException
```

firedrake/__init__.py
```diff
@@ -45,43 +45,92 @@ def init_petsc():
...
 import firedrake.citations  # noqa: F401
+__all__ += ["citations"]
+
 petsctools.cite("FiredrakeUserManual")
 del petsctools
 
 from firedrake.petsc import PETSc  # noqa: F401
+__all__ += ["PETSc"]
+
 from firedrake.assemble import assemble  # noqa: F401
```
pbrubeck: I think `# noqa: F401` is no longer needed

firedrake/__init__.py
nwng04: Just remembered the reason I add a list for finat.ufl is because `__all__` is not defined in the module.

firedrake/__init__.py
connorjward: Because it's in `__all__`? I would be surprised if that counts as a used import.

firedrake/__init__.py
connorjward: Could you add one? I've just checked and it seems very straightforward. The alternative is defining things here which definitely violates my single source of truth preference

connorjward: https://github.com/firedrakeproject/fiat/blob/main/finat/ufl/__init__.py

firedrake/__init__.py
```diff
@@ -142,46 +238,92 @@ def init_petsc():
...
+from firedrake.logging import (  
     set_level, set_log_handlers, set_log_level, DEBUG, INFO,
     WARNING, ERROR, CRITICAL, log, debug, info, warning, error,
     critical, info_red, info_green, info_blue, RED, GREEN, BLUE
 )
-from firedrake.matrix import (  # noqa: F401
+__all__ += ["set_level", "set_log_handlers", "set_log_level", "DEBUG", "INFO",
+            "WARNING", "ERROR", "CRITICAL", "log", "debug", "info", "warning", "error",
+            "critical", "info_red", "info_green", "info_blue", "RED", "GREEN", "BLUE"]
+
```
leo-collins, line 299: Do we want these in the public API?

firedrake/__init__.py
```diff
@@ -142,46 +238,92 @@ def init_petsc():
...
 )
-from firedrake.randomfunctiongen import *  # noqa: F401
-from firedrake.external_operators import (  # noqa: F401
+__all__ += [
+    "Ensemble", "EnsembleFunction", "EnsembleCofunction",
+    "EnsembleFunctionSpace", "EnsembleDualSpace",
+    "EnsembleBJacobiPC", "EnsembleBlockDiagonalMat",
+]
+
+import firedrake.randomfunctiongen as _randomfunctiongen
```
connorjward, line 276:
```suggestion
# randomfunctiongen generates __all__ dynamically, so we cannot enumerate
# things here
import firedrake.randomfunctiongen as _randomfunctiongen
```

firedrake/__init__.py
connorjward: I would say no, but I also think that that is out of scope of this PR as it introduces a whole new set of breaking changes.

firedrake/__init__.py
```diff
@@ -142,46 +238,92 @@ def init_petsc():
...
-from firedrake.plot import *  # noqa: F401
-
+sys.modules["firedrake.plot"] = _plot
+from firedrake.plot import *  
+__all__ += list(_plot.__all__)
 del sys
+del _plot
+
+
+class SpikedModule:
```
connorjward: Can this get a docstring and be moved in `_deprecation.py`?

connorjward (changes requested):

PR #5418, opened 2026-09-02
Remove backwards compat tests for variable layer extrusion

connorjward:
A recent upstream update broke the test for the 2024 checkpoint file and we are deprecating variable layer extrusion anyway.


JHopeCollins (approved):
If we're removing that functionality anyway then this test is fine to remove.

PR #5428, opened 2026-09-03
Deprecate the appctx for most things

connorjward:
Now that we can pass Python objects to the solver via the options dictionary we can do nice things like:
```py

    deflation = Deflation(op=lambda x, y: inner(x-y, x-y)*dx)
    sp = {"snes_type": "python",
          "snes_python_type": "firedrake.DeflatedSNES",
          "deflated_snes_type": "newtonls",
          "deflated_snes_monitor": None,
          "deflated_snes_linesearch_type": "basic",
          "deflated_snes_deflation": deflation,
          "deflated_ksp_type": "preonly",
          "deflated_pc_type": "lu"}

    solver = NonlinearVariationalSolver(problem, solver_parameters=sp)
```
instead of
```py
    sp = {"snes_type": "python",
          "snes_python_type": "firedrake.DeflatedSNES",
          "deflated_snes_type": "newtonls",
          "deflated_snes_monitor": None,
          "deflated_snes_linesearch_type": "basic",
          "deflated_ksp_type": "preonly",
          "deflated_pc_type": "lu"}

    deflation = Deflation(op=lambda x, y: inner(x-y, x-y)*dx)
    appctx = {"deflation": deflation}

    solver = NonlinearVariationalSolver(problem, solver_parameters=sp, appctx=appctx)
```

This works great for 90% of the use cases of the appctx, but does not work for objects that are transformed during the solve. For example, if one coarsens or refines the problem then functions passed in the appctx also get coarsened or refined. This is very different to other objects currently in the appctx that are not transformed and just treat it as a bag.

This PR goes through the codebase and removes all uses of the appctx except where this transforming behaviour is necessary. To make everything explicit I have added a `dmhooks.Hooked` class that wraps things like Firedrake functions to provide the callbacks. This provides a mechanism for downstream users to provide their own mechanisms for transforming things during a solve.

### DIFF
--- a/docs/notebooks/07-geometric-multigrid.py
+++ b/docs/notebooks/07-geometric-multigrid.py
@@ -115,22 +115,22 @@ finest_mesh = hierarchy[-1]
 # To make things easier to play with, we'll wrap everything up in a function that we can call to produce a solver.
 
 # %%
-def create_solver(parameters=None):
+def create_solver(parameters=None, mass_pc_prefix=None):
     coarse_mesh = RectangleMesh(15, 10, 1.5, 1)
     hierarchy = MeshHierarchy(coarse_mesh, 3)
-    
+
     mesh = hierarchy[-1]
-    
+
     V = VectorFunctionSpace(mesh, "Lagrange", 2)
     Q = FunctionSpace(mesh, "Lagrange", 1)
     W = V*Q
-    
+
     u, p = TrialFunctions(W)
     v, q = TestFunctions(W)
-    
+
     nu = Constant(1)
     x, y = SpatialCoordinate(mesh)
-    
+
     t = conditional(y < 0.5, y - 1/4, y - 3/4)
     gbar = conditional(Or(And(1/6 < y,
                               y < 1/3),
@@ -142,12 +142,17 @@ def create_solver(parameters=None):
     value = as_vector([gbar*(1 - (12*t)**2), 0])
     bcs = [DirichletBC(W.sub(0), value, (1, 2)),
            DirichletBC(W.sub(0), zero(2), (3, 4))]
-    
+
     a = (nu*inner(grad(u), grad(v)) - p*div(v) + q*div(u))*dx
     L = inner(Constant((0, 0)), v)*dx
+
+    # Pass nu to the preconditioner
+    if mass_pc_prefix is not None:
+        parameters[f"{mass_pc_prefix}nu"] = nu
+
     wh = Function(W)
     problem = LinearVariationalProblem(a, L, wh, bcs=bcs)
-    solver = LinearVariationalSolver(problem, solver_parameters=parameters, appctx={"nu": nu})
+    solver = LinearVariationalSolver(problem, solver_parameters=parameters)
     return solver
 
 
@@ -184,10 +189,13 @@ fig.colorbar(triangles, ax=axes[1], fraction=0.032, pad=0.02);
 # This direct method is not a scalable solution technique for large problems.  Similar to our earlier example involving the mixed Poisson problem, a Schur complement method can be more efficient.  We'll use geometric multigrid to invert the elliptic velocity block, and use a viscosity-weighted pressure mass matrix to precondition the Schur complement.  This gives good results as long as viscosity contrasts are not too strong. The Python preconditioner that we use to create the mass matrix is described in more detail in the composable solvers notebook.
 
 # %%
+import petsctools
+
 class MassMatrix(AuxiliaryOperatorPC):
     def form(self, pc, test, trial):
         # Grab the viscosity
-        nu = self.get_appctx(pc)["nu"]
+        prefix = pc.getOptionsPrefix() or ""
+        nu = petsctools.Options(prefix)["nu"]
         return (nu*test*trial*dx, None)
 
 
@@ -207,7 +215,7 @@ parameters = {
     "fieldsplit_1_aux_pc_type": "icc",
 }
 
-solver = create_solver(parameters)
+solver = create_solver(parameters, "fieldsplit_1_")
 solver.solve()
 
 # %% [markdown]
@@ -246,7 +254,7 @@ parameters = {
       "mg_levels_fieldsplit_1_aux_pc_type": "icc",
 }
 
-solver = create_solver(parameters)
+solver = create_solver(parameters, "mg_levels_fieldsplit_1_")
 solver.solve()
 
 
@@ -276,23 +284,23 @@ solver.solve()
 # For simplicity, the relevant setup is copied below to start with:
 
 # %%
-def create_solver(parameters=None):
+def create_solver(parameters=None, mass_pc_prefix=None):
     coarse_mesh = RectangleMesh(15, 10, 1.5, 1)
     hierarchy = MeshHierarchy(coarse_mesh, 3)
-    
+
     mesh = hierarchy[-1]
-    
+
     V = VectorFunctionSpace(mesh, "Lagrange", 2)
     Q = FunctionSpace(mesh, "Lagrange", 1)
     W = V*Q
-    
+
     u, p = TrialFunctions(W)
     v, q = TestFunctions(W)
-    
+
     # Change me to spatially varying.
     nu = Constant(1)
     x, y = SpatialCoordinate(mesh)
-    
+
     t = conditional(y < 0.5, y - 1/4, y - 3/4)
     gbar = conditional(Or(And(1/6 < y,
                               y < 1/3),
@@ -304,13 +312,17 @@ def create_solver(parameters=None):
     value = assemble(interpolate(as_vector([gbar*(1 - (12*t)**2), 0]), V))
     bcs = [DirichletBC(W.sub(0), value, (1, 2)),
            DirichletBC(W.sub(0), zero(2), (3, 4))]
-    
+
     a = (nu*inner(grad(u), grad(v)) - p*div(v) + q*div(u))*dx
     L = inner(Constant((0, 0)), v)*dx
-    
+
+    # Pass nu to the preconditioner
+    if mass_pc_prefix is not None:
+        parameters[f"{mass_pc_prefix}nu"] = nu
+
     wh = Function(W)
     problem = LinearVariationalProblem(a, L, wh, bcs=bcs)
-    solver = LinearVariationalSolver(problem, solver_parameters=parameters, appctx={"nu": nu})
+    solver = LinearVariationalSolver(problem, solver_parameters=parameters)
     return solver
 
 
@@ -330,7 +342,7 @@ parameters = {
     "fieldsplit_1_aux_pc_type": "icc",
 }
 
-solver = create_solver(parameters)
+solver = create_solver(parameters, "fieldsplit_1_")
 solver.solve()
 
 # %%
--- a/docs/notebooks/08-composable-solvers.py
+++ b/docs/notebooks/08-composable-solvers.py
@@ -34,6 +34,7 @@
 
 # %%
 import matplotlib.pyplot as plt
+import petsctools
 
 # %%
 from firedrake import *
@@ -89,7 +90,7 @@ nullspace = MixedVectorSpaceBasis(W, [W.sub(0), VectorSpaceBasis(constant=True, 
 # Since we're going to look at a bunch of different solver options, let's have a function that builds a solver with the provided options.
 
 # %%
-def create_solver(solver_parameters, *, pmat=None, appctx=None):
+def create_solver(solver_parameters, *, pmat=None):
     p = {}
     if solver_parameters is not None:
         p.update(solver_parameters)
@@ -98,7 +99,7 @@ def create_solver(solver_parameters, *, pmat=None, appctx=None):
     p.setdefault("ksp_rtol", 1e-7)
     problem = NonlinearVariationalProblem(F, w, bcs=bcs, Jp=pmat)
     solver = NonlinearVariationalSolver(problem, nullspace=nullspace, options_prefix="", 
-                                        solver_parameters=p, appctx=appctx)
+                                        solver_parameters=p)
     return solver
 
 
@@ -300,8 +301,9 @@ class MassMatrix(AuxiliaryOperatorPC):
     def form(self, pc, test, trial):
         # Extract the original form and bcs
         a, bcs = super().form(pc, test, trial)
-        # Grab the definition of nu from the user application context (a dict)
-        nu = self.get_appctx(pc)["nu"]
+        # Grab the definition of nu from the options database
+        prefix = pc.getOptionsPrefix() or ""
+        nu = petsctools.Options(prefix)["nu"]
         return (-1/nu * test*trial*dx, bcs)
 
 
@@ -321,15 +323,15 @@ mass_parameters = {
     "fieldsplit_1": {
         "ksp_type": "preonly",
         "pc_type": "python",
-        "pc_python_type": "__main__.MassMatrix",
+        "pc_python_type": f"{__name__}.MassMatrix",
         "mass_pc_type": "lu",
+        "mass_nu": nu,
     }
 }
 
 # %%
-appctx = {"nu": nu} # arbitrary user data that is available inside the user PC object
 w.zero()
-solver = create_solver(mass_parameters, appctx=appctx)
+solver = create_solver(mass_parameters)
 solver.solve()
 convergence(solver)
 
@@ -363,6 +365,7 @@ fieldsplit_mg_parameters = {
         "pc_type": "python",
         "pc_python_type": f"{__name__}.MassMatrix",
         "mass_pc_type": "sor",
+        "mass_nu": nu,
     }
 }
 
@@ -370,9 +373,8 @@ fieldsplit_mg_parameters = {
 # Now, when the solver runs, PETSc will call back in to Firedrake for restriction and prolongation, as well as rediscretising $A$ on the coarser levels.
 
 # %%
-appctx = {"nu": nu} # arbitrary user data that is available inside the user PC object
 w.zero()
-solver = create_solver(fieldsplit_mg_parameters, appctx=appctx)
+solver = create_solver(fieldsplit_mg_parameters)
 solver.solve()
 convergence(solver)
 
--- a/docs/notebooks/08-composable-solvers.py
+++ b/docs/notebooks/08-composable-solvers.py
@@ -302,7 +302,7 @@ class MassMatrix(AuxiliaryOperatorPC):
         # Extract the original form and bcs
         a, bcs = super().form(pc, test, trial)
         # Grab the definition of nu from the options database
-        prefix = pc.getOptionsPrefix() or ""
+        prefix = (pc.getOptionsPrefix() or "") + self._prefix
         nu = petsctools.Options(prefix)["nu"]
         return (-1/nu * test*trial*dx, bcs)
 
--- a/firedrake/dmhooks.py
+++ b/firedrake/dmhooks.py
@@ -37,9 +37,13 @@ fieldsplit preconditioning, without having to set everything up in
 advance.
 """
 
+import dataclasses
 import weakref
-import numpy
+from collections.abc import Callable
 from functools import partial
+from typing import Any
+
+import numpy
 
 import firedrake
 from firedrake.petsc import PETSc
@@ -569,3 +573,32 @@ def attach_hooks(dm, level=None, sf=None, section=None):
     # a non-mixed space)
     dm.setCreateFieldDecomposition(create_field_decomposition)
     dm.setCreateSubDM(create_subdm)
+
+
+@dataclasses.dataclass(frozen=True)
+class Hooked:
+    """Class wrapping an object that can be passed through a solver stack.
+
+    Parameters
+    ----------
+    obj
+        The wrapped object (e.g. a `firedrake.Function`).
+    refine_callback
+        Callback to refine the object.
+    coarsen_callback
+        Callback to coarsen the object.
+
+    """
+    obj: Any
+    refine_callback: Callable | None = None
+    coarsen_callback: Callable | None = None
+
+    def refine(self):
+        if self.refine_callback is None:
+            raise NotImplementedError("No implementation for 'refine_callback' found")
+        return dataclasses.replace(self, obj=self.refine_callback(self.obj))
+
+    def coarsen(self):
+        if self.coarsen_callback is None:
+            raise NotImplementedError("No implementation for 'coarsen_callback' found")
+        return dataclasses.replace(self, obj=self.coarsen_callback(self.obj))
--- a/firedrake/matrix_free/operators.py
+++ b/firedrake/matrix_free/operators.py
@@ -107,8 +107,8 @@ class ImplicitMatrixContext:
         self.a = a
         self.aT = adjoint(a)
         self.comm = a.arguments()[0].function_space().comm
-        self.fc_params = {} if fc_params is None else fc_params
-        self.appctx = {} if appctx is None else appctx
+        self.fc_params = fc_params
+        self.appctx = appctx
 
         # Collect all DirichletBC instances including
         # DirichletBCs applied to an EquationBC.
--- a/firedrake/matrix_free/operators.py
+++ b/firedrake/matrix_free/operators.py
@@ -65,7 +65,7 @@ def find_sub_block(iset, ises, comm):
     return found
 
 
-class ImplicitMatrixContext(object):
+class ImplicitMatrixContext:
     # By default, these matrices will represent diagonal blocks (the
     # (0,0) block of a 1x1 block matrix is on the diagonal).
     on_diag = True
@@ -101,6 +101,8 @@ class ImplicitMatrixContext(object):
             and the like. By default None.
         """
         from firedrake.assemble import get_assembler
+
+        assert appctx is None, "old API"
 
         self.a = a
         self.aT = adjoint(a)
--- a/firedrake/mg/ufl_utils.py
+++ b/firedrake/mg/ufl_utils.py
@@ -380,16 +380,20 @@ def reconstruct_snescontext(context, self, coefficient_mapping=None):
 
     # FIXME: this makes no sense for things in the options dictionary
     appctx = context._appctx
-    new_appctx = {}
-    for k in sorted(appctx.keys()):
-        v = appctx[k]
-        if k != "state":
-            # Constructor makes this one.
-            try:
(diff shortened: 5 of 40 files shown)


firedrake/matrix_free/operators.py
```diff
@@ -102,11 +102,13 @@ def __init__(
         """
         from firedrake.assemble import get_assembler
 
+        assert appctx is None, "old API"
+
```
connorjward, line 106:
```suggestion
```

firedrake/mg/ufl_utils.py
```diff
@@ -377,17 +377,23 @@ def reconstruct_snescontext(context, self, coefficient_mapping=None):
...
-            try:
-                new_appctx[k] = self(v, self, coefficient_mapping=coefficient_mapping)
-            except ReconstructionError:
-                # Assume not something that needs reconstruction (e.g. float)
-                new_appctx[k] = v
+
+    # FIXME: this makes no sense for things in the options dictionary
+    appctx = context._appctx
+    if appctx is not None:
+        raise NotImplementedError
```
connorjward, line 384: This isn't getting hit anywhere, so we're not testing this

firedrake/preconditioners/pmg.py
```diff
@@ -259,16 +259,21 @@ def _coarsen_form(a, coefficient_mapping):
...
-        for key in cappctx:
-            val = cappctx[key]
-            if isinstance(val, dict):
-                cappctx[key] = self.coarsen_quadrature(val, fdeg, cdeg)
-            elif isinstance(val, ufl.Form):
-                cappctx[key] = _coarsen_form(val, fine_to_coarse_map)
-            elif isinstance(val, ufl.classes.Expr):
-                cappctx[key] = ufl.replace(val, fine_to_coarse_map)
+        if fctx._appctx is not None:
+            raise NotImplementedError
```
connorjward, line 265: Not getting triggered anywhere and the tests are passing...

firedrake/solving_utils.py
```diff
@@ -234,15 +235,15 @@ def __init__(self, problem,
...
         # Now we don't have a temporary state inside the snes
         # context we could just require the user to pass in the
         # full state on the outside.
-        appctx.setdefault("state", self._x)
-        appctx.setdefault("form_compiler_parameters", self.fcp)
+        if state is None:
+            self.state = self._x
+        else:
+            self.state = state
+        # appctx.setdefault("form_compiler_parameters", self.fcp)
```
connorjward, line 246:
```suggestion
```

firedrake/solving_utils.py
```diff
@@ -304,14 +305,17 @@ def __init__(self, problem,
 
     @property
     def appctx(self) -> dict:
+        # debugging
+        raise AssertionError("old api")
+
```
connorjward, line 310:
```suggestion
```

firedrake/solving_utils.py
```diff
@@ -350,20 +354,27 @@ def get_python_option(
...
+            if self._appctx is not None:
+                try:
+                    value = self._appctx[option]
+                except KeyError:
+                    if default is not _missing:
+                        value = default
+                    else:
+                        raise KeyError
+                else:
+                    assert False, "old api"
```
connorjward, line 366:
```suggestion
```

firedrake/variational_solver.py
```diff
@@ -402,6 +405,9 @@ def __init__(self, problem, *, solver_parameters=None,
 
         solver_parameters = flatten_parameters(solver_parameters or {})
 
+        # debugging
+        assert appctx is None, "old api"
+
```
connorjward, line 410:
```suggestion
```

firedrake/variational_solver.py
```diff
@@ -342,6 +342,9 @@ class NonlinearVariationalSolver(OptionsManager, NonlinearVariationalSolverMixin
         :class:`.Function` or :class:`.Cofunction` with positive
         values on cells to refine.
 
+    Examples
+    --------
+
```
connorjward, line 347: Fixes docs build

firedrake/preconditioners/hiptmair.py
```diff
@@ -194,7 +194,8 @@ def coarsen(self, pc):
 
         cdegree = max(as_tuple(celement.degree()))
         if formdegree > 1 and cdegree > 1:
-            shift = appctx.get("hiptmair_shift", None)
+            # TODO: now things are prefixed the extra 'hiptmair_' is bad practice
+            shift = ctx.get_python_option(options_prefix, "hiptmair_shift", None)
```
connorjward, line 198: We don't set this anywhere

firedrake/solving_utils.py
```diff
@@ -234,15 +235,15 @@ def __init__(self, problem,
...
 
-        if appctx is None:
-            appctx = {}
         # A split context will already get the full state.
-        # TODO, a better way of doing this.
         # Now we don't have a temporary state inside the snes
         # context we could just require the user to pass in the
         # full state on the outside.
-        appctx.setdefault("state", self._x)
-        appctx.setdefault("form_compiler_parameters", self.fcp)
```
connorjward, line 245: This is a small breaking change where these no longer live in `_SNESContext.appctx`. I guess we could leave them but the warning process would be more complicated.

connorjward:
@JHopeCollins @pbrubeck this is now ready for review. As you can see I have made sure to remove every single use of `appctx` in the repository. There are two key issues that I have identified:

1. In a handful of places we transform the entries in the appctx (e.g. coarsen). This only works because it relies on passing things all the way through the solver stack. If we put things in the options then they only appear at the point of use. We have no ability to transform them along the way. Everything seems to work without these transformation steps however.
2. `"state"` and `"form_compiler_parameters"` are no longer stored in the appctx. This is a small breaking change but it makes it a little easier to catch and warn about deprecated usage.

Feedback welcome.

docs/notebooks/07-geometric-multigrid.py
```diff
@@ -184,10 +189,13 @@ def create_solver(parameters=None):
...
 
 # %%
+import petsctools
+
 class MassMatrix(AuxiliaryOperatorPC):
     def form(self, pc, test, trial):
         # Grab the viscosity
-        nu = self.get_appctx(pc)["nu"]
+        prefix = pc.getOptionsPrefix() or ""
+        nu = petsctools.Options(prefix)["nu"]
```
pbrubeck, line 198: Why can't one use `PETSc.Options()` here?

firedrake/mg/ufl_utils.py
pbrubeck: We should add a test then. Solve Stokes with multigrid as the outer solver and fieldsplit + MassInvPC as the relaxation.

firedrake/preconditioners/hiptmair.py
pbrubeck: you mean in the tests?

firedrake/preconditioners/pmg.py
```diff
@@ -259,16 +259,21 @@ def _coarsen_form(a, coefficient_mapping):
         fine_to_coarse_map = dict(zip(fproblem.J.arguments(), cproblem.J.arguments()))
         fine_to_coarse_map[fu] = cu
 
+        # FIXME: this makes no sense for things in the options dictionary
```
pbrubeck, line 262: Why does it make no sense? A user can put a UFL object in their appctx, so we need to transfer that into the coarse levels.

firedrake/preconditioners/pmg.py
pbrubeck: This tells us that we need to add a test.

firedrake/preconditioners/hiptmair.py
```diff
@@ -194,7 +194,8 @@ def coarsen(self, pc):
 
         cdegree = max(as_tuple(celement.degree()))
         if formdegree > 1 and cdegree > 1:
-            shift = appctx.get("hiptmair_shift", None)
+            # TODO: now things are prefixed the extra 'hiptmair_' is bad practice
```
pbrubeck, line 197: Don't add the extra `hiptmair_` prefix then.

firedrake/preconditioners/hiptmair.py
pbrubeck, line 198:
```suggestion
            shift = ctx.get_python_option(options_prefix, "shift", None)
```

pbrubeck:
> 1. In a handful of places we transform the entries in the appctx (e.g. coarsen). This only works because it relies on passing things all the way through the solver stack. If we put things in the options then they only appear at the point of use. We have no ability to transform them along the way. Everything seems to work without these transformation steps however.

Are you saying that the inner `mg_levels_` solvers are blind to the outer, unprefixed solver options, and therefore you cannot propagate UFL objects in the options dictionary from the outer prefix into the inner prefix?

connorjward:
> > 1. In a handful of places we transform the entries in the appctx (e.g. coarsen). This only works because it relies on passing things all the way through the solver stack. If we put things in the options then they only appear at the point of use. We have no ability to transform them along the way. Everything seems to work without these transformation steps however.
> 
> Are you saying that the inner `mg_levels_` solvers are blind to the outer, unprefixed solver options, and therefore you cannot propagate UFL objects in the options dictionary from the outer prefix into the inner prefix?

I'm saying that things work exactly like other PETSc options, only that Python objects are allowed too. Which means that UFL options do not propagate.

firedrake/preconditioners/hiptmair.py
connorjward: This will break existing code because users passing `hiptmair_shift` in the old appctx will have their code stop working. I can fix this but it's more complicated than that.

docs/notebooks/07-geometric-multigrid.py
connorjward: Because `nu` is a Python object.

pbrubeck: Adding that extra import to the demo is far from ideal. Should we add a `PCBase.get_options(prefix)`?

In my humble opinion, it'd be best if petsctools becomes ingested by petsc4py. An extra package is hard to mantain, not to mention that this will become load-bearing for the user. Only a few will be directly interacting with petsc4py, and now we are asking them to become aware of nuances like needing a separate library to read options depending on their type.

connorjward: > Adding that extra import to the demo is far from ideal. Should we add a PCBase.get_options(prefix)?

I don't agree that an extra import is a bad thing, but I do think a utility method would be helpful. It should really be `PCBase.get_options(pc)` because the current dance with options is complicated.

> it'd be best if petsctools becomes ingested by petsc4py

I agree, but in my experience the PETSc developers are very resistant to the sorts of things that we put into petsctools.

> needing a separate library to read options depending on their type.

This isn't correct. `petsctools.Options` is a subclass of `PETSc.Options` and so everything is inherited.

firedrake/preconditioners/hiptmair.py
pbrubeck: I see. I don't think there are many users of HiptmairPC, I think that breaking API is fine

pbrubeck:
> I'm saying that things work exactly like other PETSc options, only that Python objects are allowed too. Which means that UFL options do not propagate.

I still don't understand. What happens with the standard PETSc options? And why can you not transform them? We already have similar logic for `coarse_mat_type`...

connorjward:
> > I'm saying that things work exactly like other PETSc options, only that Python objects are allowed too. Which means that UFL options do not propagate.
> 
> I still don't understand. What happens with the standard PETSc options? And why can you not transform them? We already have similar logic for `coarse_mat_type`...

PETSc options exist at one particular point in the options hierarchy. We only look at them when we get to the relevant preconditioner. We don't care if the preconditioner is used as part of a multigrid cycle or fieldsplit etc.

The way to transform things is for them not to be PETSc options and instead tied to the `_SNESContext`, which is an object that gets carried through the stack and hence is able to transform things along the way.

pbrubeck:
> The way to transform things is for them not to be PETSc options and instead tied to the `_SNESContext`, which is an object that gets carried through the stack and hence is able to transform things along the way.

Why not keep the appctx dict as an internal bag of UFL objects that we collect from the PETSc options so that we may transform them whenever the problem gets transformed?

connorjward:
> > The way to transform things is for them not to be PETSc options and instead tied to the `_SNESContext`, which is an object that gets carried through the stack and hence is able to transform things along the way.
> 
> Why not keep the appctx dict as an internal bag of UFL objects that we collect from the PETSc options so that we may transform them whenever the problem gets transformed?

That amounts to flattening the options dictionary. Who is to say that a `Function` passed at the root of the solver stack should be transformed in the same way as one inside a deeply nested MG+Fieldsplit stack? The answer is that this behaviour is a fundamentally different thing to passing PETSc options and needs another mechanism.

I confess that I don't really want to add any tests. I am happy to create an issue but I think fixing this could be considered out of scope of this PR.

pbrubeck:
> I confess that I don't really want to add any tests. I am happy to create an issue but I think fixing this could be considered out of scope of this PR.

I'm opposed to knowingly breaking (untested) functionality that we fully understand. If you feel like you need help adding a test, I will give you one.

connorjward:
> > I confess that I don't really want to add any tests. I am happy to create an issue but I think fixing this could be considered out of scope of this PR.
> 
> I oppose to knowingly breaking (untested) functionality that we fully understand. If you feel like you need help adding a test, I will give you one.

Sure. It'd be great if you could add some tests to `main` that I need to conform to.

pbrubeck:
Added a test in #5436 that pins the current behaviour: an `appctx` `Function` (viscosity `mu`) must be injected onto every coarse level for a monolithic multigrid Stokes solve to work.

connorjward:
Currently failing tests but a proof of concept now exists in `variational_solver.py`. The core of the change is that users are only able to put things in the appctx that either:
1. We know how to refine and coarsen etc (like a `Function`)
2. Define their own callbacks for these operations

firedrake/preconditioners/hiptmair.py
connorjward: done.

docs/notebooks/07-geometric-multigrid.py
```diff
@@ -210,7 +218,7 @@ def form(self, pc, test, trial):
     "fieldsplit_1_aux_pc_type": "icc",
 }
 
-solver = create_solver(parameters)
+solver = create_solver(parameters, "fieldsplit_1_")
```
JHopeCollins, line 221:
```suggestion
solver = create_solver(parameters, mass_pc_prefix="fieldsplit_1_")
```

docs/notebooks/07-geometric-multigrid.py
```diff
@@ -249,7 +257,7 @@ def form(self, pc, test, trial):
       "mg_levels_fieldsplit_1_aux_pc_type": "icc",
 }
 
-solver = create_solver(parameters)
+solver = create_solver(parameters, "mg_levels_fieldsplit_1_")
```
JHopeCollins, line 260:
```suggestion
solver = create_solver(parameters, mass_pc_prefix="mg_levels_fieldsplit_1_")
```

docs/notebooks/07-geometric-multigrid.py
```diff
@@ -333,7 +345,7 @@ def create_solver(parameters=None):
     "fieldsplit_1_aux_pc_type": "icc",
 }
 
-solver = create_solver(parameters)
+solver = create_solver(parameters, "fieldsplit_1_")
```
JHopeCollins, line 348:
```suggestion
solver = create_solver(parameters, mass_pc_prefix="fieldsplit_1_")
```

docs/notebooks/08-composable-solvers.py
```diff
@@ -363,16 +365,16 @@ def form(self, pc, test, trial):
         "pc_type": "python",
         "pc_python_type": f"{__name__}.MassMatrix",
```
JHopeCollins, line 366: Can't we give the type directly now?
```suggestion
        "pc_python_type": MassMatrix,
```

firedrake/mg/ufl_utils.py
```diff
@@ -254,21 +254,10 @@ def reconstruct_function(expr, self, coefficient_mapping=None):
...
-            level = int(prev_level) + level_inc
-            name = f"{name}_level_{level}"
-
-        new = firedrake.Function(Vnew, name=name)
-        manager = get_transfer_manager(V.dm)
-        manager.transfer(expr, new)
+        if self is refine:
+            new = firedrake.solving_utils._refine_function(expr)
+        else:
+            new = firedrake.solving_utils._coarsen_function(expr)
```
JHopeCollins, line 260:
```suggestion
        if self is refine:
            new = firedrake.solving_utils._refine_function(expr)
        elif self is coarsen:
            new = firedrake.solving_utils._coarsen_function(expr)
        else:
            raise DontKnowWhatToDoError
```

firedrake/preconditioners/bddc.py
```diff
@@ -31,24 +31,24 @@ class BDDCPC(PCBase):
...
+      provide the arguments (a Mat tabulating the gradient of the auxiliary H1 space) and
+      keyword arguments supplied to ``PETSc.PC.setBDDCDiscreteGradient``.
+    - ``'bddc_get_divergence_mat'`` for problems in H(div) (resp. 2D H(curl)), this is
+      provide the arguments (a Mat with the assembled bilinear form testing the divergence
+      (curl) against an L2 space) and keyword arguments supplied to ``PETSc.PC.setDivergenceMat``.
+    - ``'bddc_primal_markers'`` a Function marking degrees of freedom of the solution space to be included in the
+      coarse space. Any nonzero value is counted as a marked degree of freedom.
+      If a DG(0) Function is provided, then all degrees of freedom on the cell are marked.
+      Alternatively, ``'primal_markers'`` can be a list of the global degrees of freedom to
+      be supplied directly to ``PETSc.PC.setBDDCPrimalVerticesIS``.
```
JHopeCollins, line 50: Would be useful to add a note to these options saying they can only be passed via the solver parameters.

firedrake/preconditioners/facet_split.py
```diff
@@ -53,8 +53,8 @@ def initialize(self, pc):
         domains = domains.split(",")
 
         a, bcs = self.form(pc)
-        appctx = self.get_appctx(pc)
-        fcp = appctx.get("form_compiler_parameters")
+        ctx = dmhooks.get_appctx(pc.getDM())
+        fcp = ctx.fcp
```
JHopeCollins, line 57: Doesn't look like `ctx` is used again.
```suggestion
        fcp = dmhooks.get_appctx(pc.getDM()).fcp
```

firedrake/preconditioners/massinv.py
```diff
@@ -18,17 +19,16 @@ class MassInvPC(AssembledPC):
     For Stokes problems, to be spectrally equivalent to the Schur
     complement, the mass matrix should be weighted by the viscosity.
     This can be provided (defaulting to constant viscosity) by
-    providing a field defining the viscosity in the application
-    context, keyed on ``"mu"``.
+    providing a field defining the viscosity, keyed on ``"Mp_mu"``.
```
JHopeCollins, line 22:
```suggestion
    providing a field defining the viscosity via the solver parameters, keyed on ``"Mp_mu"``.
```

firedrake/dmhooks.py
```diff
@@ -569,3 +573,45 @@ def attach_hooks(dm, level=None, sf=None, section=None):
...
     dm.setCreateSubDM(create_subdm)
+
+
+def identity_callback(self):
+    return self
+
+
+@dataclasses.dataclass(frozen=True)
+class Hooked:
+    """Class wrapping an object that can be passed through a solver stack.
```
JHopeCollins, line 584: A bit more detail here please. Why is this necessary and what does it do to enable objects to pass through the solver stack?

firedrake/solving_utils.py
```diff
@@ -295,6 +298,62 @@ def __init__(self, problem,
...
+                if default is not _missing:
+                    value = default
+                else:
+                    raise KeyError
+            else:
+                if not isinstance(value, dmhooks.Hooked):
+                    warnings.warn(
+                        "Passing arbitrary Python objects to preconditioners via the 'appctx' kwarg "
+                        "is now deprecated. Either pass the objects into the PETSc options "
+                        "directly or specify hooks instead.",
```
JHopeCollins, line 350: We should tell users where to I learn how to specify hooks when they see this.

firedrake/variational_solver.py
```diff
@@ -291,7 +293,90 @@ def compute_bc_lifting(J: ufl.BaseForm | slate.TensorBase,
...
+        as for the nullspace, but used to
+        make the right hand side consistent.
+    near_nullspace
+        as for the nullspace, but used to
+        specify the near nullspace (for multigrid solvers).
+    solver_parameters
+        Solver parameters to pass to PETSc.
+        This should be a dict mapping PETSc options to values.
+    appctx
+        (Deprecated) A dictionary containing application
```
JHopeCollins, line 316: Passing random objects to pcs is deprecated, but the appctx now is not fully deprecated?

JHopeCollins (commented):
Great progress. Nothing major from me at the moment

firedrake/dmhooks.py
```diff
@@ -569,3 +573,45 @@ def attach_hooks(dm, level=None, sf=None, section=None):
...
+        return dataclasses.replace(self, obj=self.coarsen_callback(self.obj))
+
+    @classmethod
+    def identity(cls, obj):
+        """Return an object that always transforms into itself."""
+        return cls(
+            obj,
+            refine_callback=identity_callback,
+            coarsen_callback=identity_callback,
+        )
```
JHopeCollins, line 617: Why have this as a class method?

connorjward: This is a factory method. It returns a new instance of `Hooked`.

firedrake/variational_solver.py
connorjward, line 316:
```suggestion
        A dictionary containing application
```

connorjward: Correct. This is from an earlier attempt at this.

docs/notebooks/08-composable-solvers.py
connorjward: Yes, but I want that to be a separate PR where I do it everywhere.

firedrake/mg/ufl_utils.py
connorjward: Added an `assert self is coarsen`. An assertion is fine because this isn't public facing code.

firedrake/preconditioners/bddc.py
connorjward: done

firedrake/variational_solver.py
connorjward: rewritten

firedrake/solving_utils.py
connorjward: In my refactoring this shouldn't ever get triggered (everything in the appctx gets turned into a `Hooked`). Turning into an assertion.

firedrake/dmhooks.py
connorjward: more detail written

firedrake/solving_utils.py
```diff
@@ -227,17 +237,9 @@ def __init__(self, problem,
...
-        # A split context will already get the full state.
-        # TODO, a better way of doing this.
-        # Now we don't have a temporary state inside the snes
-        # context we could just require the user to pass in the
-        # full state on the outside.
-        appctx.setdefault("state", self._x)
-        appctx.setdefault("form_compiler_parameters", self.fcp)
-
-        self.appctx = appctx
+        self.state = state if state is not None else self._x
```
JHopeCollins: When is `state` ever different from `self._x`?

JHopeCollins: So if I'm in say `fieldsplit_1_` then the `appctx['state']` is the state from the full unsplit problem? That's surprising. Where do we rely on that?

connorjward: Inside `PCDPC` https://github.com/firedrakeproject/firedrake/pull/5428/changes#diff-0ceea70c79e2ee24152682d7015234f4f6f21427a2da68e008a2c41dbbf56439R110

Used in https://github.com/firedrakeproject/firedrake/pull/5428/changes#diff-0b093988e0f90d6943054818077093bc0fde51b1d4ea6d9da673165660f36e77

JHopeCollins: 1. I do not think that should be `split`, that should be `subfunctions`. I think we have been getting away with that because we make a petsc mat and KSP rather than a solver.
2. Arguably that velocity field should be passed in the same way that `mu` is for `MassInvPC`. So rather than passing an integer with key `"velocity_space"` you pass a coefficient with key `"velocity"`. This would be a `Function` so would need to go through the same transformation hooks as `mu` does.

I am not sure whether that should go in this PR. If it is the only place where `appctx['state']` is different to `ctx._x` then I would be tempted to say that it should.

connorjward: I can have a go at option 2. Note that since this PR changes an awful lot of user facing code I am planning on merging this post release.

JHopeCollins: > I am planning on merging this post release.

A shame, but understandable

connorjward: Just learned that we have zero tests for PCDPC... only demos

JHopeCollins: I think it is possibly even implemented incorrectly. I think the ordering of the solves is backward. I don't have time to properly investigate now but I will open and issue.

connorjward: This is now done. Required some petsctools fixes (https://github.com/firedrakeproject/petsctools/pull/57). It is *so much nicer!*.

> This would be a Function so would need to go through the same transformation hooks as mu does.

Note that this isn't right. The entire point here is that we *don't* want to transform the velocity field through the solver. Therefore I just pass it in the solver parameters directly.

JHopeCollins: It will need to be coarsened in a multigrid solver though

connorjward: True. But it is nicer to be able to pass the function directly in the parameters if possible. I think the answer is probably that we need to enable both of the following:
```py
"_pcd_velocity_space": u
# no appctx
```
and
```py
"_pcd_velocity_space": "some_key"
...
appctx={"some_key": u}
```
or, equivalently
```py
appctx = AppContext()
"_pcd_velocity_space": appctx.add(u)
```

Note that `AppContext` here isn't the same as the current petsctools thing. This would have to register the hooks.

connorjward: Or maybe
```py
"_pcd_velocity_space": Hooked(u)
```
is the right design ultimately

PR #5431, opened 2026-09-04
Basic MLIR Code Generation into PyOP3

SamSJackson:
## PR: MLIR Code Generation for PyOP3

This PR presents the basic MLIR generation for operations such as array assignment.

MLIR is being considered as an alternative to Loopy as the multi-level compiler passes allow us to target CPUs and GPUs from a similar base IR. MLIR also provides compiler passes that would allow us to optimise for the respective GPU targets. 

At this stage, this PR presents basic generation of MLIR for CPUs. 

## How does it work? 

Within `pyop3/compile/`, the file `core.py` traverses the PyOP3 AST and passes buffer expressions to `mlir.py`.

The `MLIRCodegenContext` class parses the buffer expressions into SSAValues.
MLIR is an Static Single Assignment (SSA) compiler, which means that variables can be defined exactly once (i.e. x = x + 1 is not allowed). 
MLIR is composed of dialects, where dialects correspond to components of an IR. These dialects are then lowered through compiler passes and optimisations, which allows us to have multiple targets. 
In this work, the key dialects used: 
- `scf`: **S**tructured **C**ontrol **F**low serves as the iteration domain for our operations. Higher-level dialects such as `linalg` or `affine` could not be used due to the indirect memory accesses, which are not affine indexing operations.
- `memref`: This is our dialect for buffers. As an alternative, the higher-level `tensor` could be used but it does not make sense to mix dialect levels. Further, without `linalg`/`affine`, compiler optimisations for `tensor` cannot be utillised. Sources from xDSL also suggested that `memref` was the better choice here.
- `arith`: This represents constant values as well as arithmetic operations. 
 
A key design choice in this implementation is that MLIR is produced as the PyOP3 AST is parsed by `core.py`. 
Opposingly, Loopy collects all expressions into pymbolic expressions before summarily generating code.
Our approach was chosen because collecting the PyOP3 expressions would have required an additional IR for temporary storage of variables. The collection-generation approach was initially discussed and attempted but I concluded it was bloated.

As the MLIR FuncOp requires knowledge of all arguments before creation, MLIR is initially generated to a Block. Arguments can then be introduced procedurally. At finalisation, the arguments and operations are detached from the Block and a FuncOp is created, operations attached. Arguments are also re-ordered to match Loopy. 

Iteration domains are produced eagerly and an internal symbol table is employed to ensure re-use of key SSA values (such as iteration variables and buffer args). 

## What happens next? 
Foremost, there is some software engineering flaws with this implementation. Namely, the type inference.
There was a `dtype` property introduces into PyOP3 expressions but MLIR also uses an `index` type, an `int` required for load/storing memref/tensors. I have yet to produce a cleaner approach for dealing with this. 
Also some more cleaning to do more generally, I am sure.

**Regarding future work**: this PR does not introduce compilation or execution for PyOP3.

Compilation requires an MLIR binary in PyOP3 as well as working on `pyop3/cc.py`. I have not viewed how difficult a task that this will be just yet. 
For compilation, there must be a lowering pass (from `mlir-opt`), a translation pass (from `mlir-translate`) and a compilation to a shared library (from `clang`). Minimal passes for the lowering passes have been found, experimentation can be done in the future to find an optimal set of passes. The set of passes should be affected by target hardware.

In terms of execution, it is a case of wrapping numpy arrays in corresponding `memref` descriptors. This should happen in `pyop3/insn/exec.py`. It should not be a particularly invasive change.

### DIFF
--- a/pyop3/compile/mlir.py
+++ b/pyop3/compile/mlir.py
@@ -1,86 +1,714 @@
-from pyop3.lower.context import CodegenContext
+from __future__ import annotations
+
+import functools
+import numbers
+import contextlib
+import dataclasses
+import numpy as np
+from typing import Any, Dict, List, Tuple
+from immutabledict import immutabledict as idict
+
+import pyop3
+from pyop3 import utils 
+from pyop3.buffer import IndexedBuffer, NullBuffer, PetscMatBuffer
+from pyop3.constants import INC, MAX_RW, MAX_WRITE, MIN_RW, MIN_WRITE, READ, RW, WRITE
+from pyop3.expr import NameVar, Operator
+
+from pyop3.insn.base import (
+    AssignmentType,
+    assignment_type_as_intent,
+)
+
+from pyop3.compile.context import CodegenContext
+
+from xdsl.dialects import arith, func, memref, scf
+from xdsl.builder import Builder, InsertPoint
+from xdsl.ir import SSAValue, Block, Region, Operation
+
+from xdsl.dialects.builtin import (
+    DYNAMIC_INDEX,
+    IntegerType,
+    IndexType,
+    IntegerAttr,
+    Float64Type,
+    FloatAttr,
+    MemRefType,
+    ModuleOp,
+    UnitAttr,
+)
+
+from pyop3.dtypes import IntType, RealType, ScalarType
+
+# NOTE: Can I just set an IndexType object here and reuse this? 
+# Bad design from xDSL. Who makes a type an object?   
+iType = IndexType()
+
+NUMPY_TO_XDSL = {
+    np.dtype("int64"): IntegerType, # NOTE: Temp forcing int64 to int32
+    IntType: IntegerType,
+    RealType: Float64Type,
+    ScalarType: Float64Type, 
+}
+
+# TODO: Maybe better way to read this from IntType
+DEFAULT_INT_WIDTH = 32 
+
+SSAValueT = SSAValue | numbers.Number | NameVar
+
+def get_mlir_type(np_dtype) -> Any:
+    return NUMPY_TO_XDSL[np_dtype]
+
+def _is_float(dtype) -> bool:
+    return get_mlir_type(dtype) is Float64Type
+
+# NOTE: Assuming 32-bit integer 
+def _mlir_type(dtype):
+    t = get_mlir_type(dtype)
+    return t(32) if t is IntegerType else t()
+
+
+def _expr_dtype(expr, hint):
+    """
+    The pyop3 dtype an operand will actually produce.
+    
+    If constant number, use fallback hint  
+    """
+    if isinstance(expr, numbers.Number):
+        return hint
+    return expr.dtype
+
+
+def _coerce(context, ssa, from_dtype, to_dtype) -> SSAValue:
+    """
+    Insert cast if float and int types are differing from expected.
+
+    Ugly function, seeking improvements. 
+    """
+    if from_dtype == to_dtype:
+        return ssa
+    from_f, to_f = _is_float(from_dtype), _is_float(to_dtype)
+    if not from_f and to_f:
+        return context.insert(arith.SIToFPOp(ssa, Float64Type())).result
+    if from_f and not to_f:
+        return context.insert(arith.FPToSIOp(ssa, _mlir_type(to_dtype))).result
+    return ssa
+
+@dataclasses.dataclass
+class Argument:
+    name: str
+    dtype: np.dtype
+    shape: Tuple[int] | None
+    buffer: Any = None # NOTE: used to dedup / remap. Probably remove. Use name property.
+
+    def __str__(self):
+        return self.name
+
+    def __repr__(self):
+        return f"<{self.name}, dtype: {self.dtype}, shape: {self.shape or '?'}>"
+
+
+class SymbolTable:
+    """
+    Class deals with SSAValues for a respective scope.
+
+    e.g: SSAValues inside scf.ForOp are not valid outwith the iterative loop.
+    """
+    def __init__(self) -> None:
+        self._scopes: List[Dict[Any, SSAValue]] = [dict()]
+
+    def push(self) -> None:
+        self._scopes.append(dict())
+
+    def pop(self) -> None:
+        self._scopes.pop()
+
+    def define(self, key, value: SSAValue) -> None:
+        """ Key should not be re-defined. SSA only permits single definition """
+        assert key not in self._scopes[-1] 
+        self._scopes[-1][key] = value
+
+    def __getitem__(self, key) -> SSAValue:
+        """ Seek SSAValue from innermost -> outermost scope """
+        for scope in reversed(self._scopes):
+            if key in scope:
+                return scope[key]
+        raise KeyError(f"{key} is not in the SSA symbol table")
+
+    def __contains__(self, key) -> bool:
+        return any(key in s for s in self._scopes)
+
 
 class MLIRCodegenContext(CodegenContext):
-    """ 
-    Class produces MLIR kernels from PyOP3 buffers, using xDSL. 
+    """
+    Produces MLIR kernels from PyOP3 buffers using xDSL.
     """
 
     def __init__(self, *, propagate_negatives: bool, mask_array_accesses: bool) -> None:
         super().__init__(
             propagate_negatives=propagate_negatives,
-            mask_array_accesses=mask_array_accesses
+            mask_array_accesses=mask_array_accesses,
         )
-    
-    @abstractmethod
+
+        self._entry_block = Block()
+        self._builder = Builder(InsertPoint.at_end(self._entry_block))
+
+        # Insertion-point stack (restored when leaving nesting regions).
+        self.insertion_stack: List[InsertPoint] = []
+
+        # Symbol table (buffers in outer scope, inames/loop-idx in inner).
+        self.symbol_table = SymbolTable()
+
+        # buffer identity -> block-arg SSAValue
+        self._buffer_args: Dict[Any, SSAValue] = dict()
+
+        # iname -> (start, stop) domain window
+        self._domains: Dict[str, Tuple[SSAValueT, SSAValueT]] = dict()
+
+        # NameVar temporaries already resolved to SSA values live in
+        # symbol_table; this records which names are temporaries.
+        self._temporaries: set = set()
+
+    def insert(self, op: Operation) -> Operation:
+        """ Inserts an MLIR operation into the block """
+        self._builder.insert(op)
+        return op
+
+    def insert_arg(self, arg: MemrefType, name: str = None) -> SSAValue:
+        """ Insert memref args into function definition """ 
+        assert isinstance(arg, MemRefType)
+        block = self._entry_block
+        memref_ssa = block.insert_arg(arg, len(block.args)) 
+        if name: 
+            memref_ssa.name_hint = name 
+        return memref_ssa 
+
+    def _const_index(self, value: int) -> SSAValue:
+        # NOTE: Casting with `int` might be a problem
+        op = self.insert(arith.ConstantOp(
+            IntegerAttr(int(value), iType)
+           )
+        )
+        return op.result
+
+    def _to_index(self, ssa: SSAValue) -> SSAValue:
+        # scf.for bounds and memref indices must be `index`.
+        if isinstance(ssa.type, IndexType):
+            return ssa
+        return self.insert(arith.IndexCastOp(ssa, iType)).result
+
     def var(self, iname: str, *args) -> str:
-        """
-        Implementation to represent symbolic variable for respective IR
-        """
-        pass
+        return iname
 
-    @abstractmethod
     def add_domain(self, iname: str, *args) -> None:
-        pass
+        nargs = len(args)
+        if nargs == 1:
+            start, stop = 0, args[0]
+        else:
+            assert nargs == 2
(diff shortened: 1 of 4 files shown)


pyop3/compile/mlir.py
```diff
@@ -0,0 +1,715 @@
...
+    return context.insert(arith.SubiOp(zero, val)).result
+
+@_lower_expr.register(pyop3.expr.Comparison)
+def _(cond, /, iname_maps, loop_indices, *, context, target_type, **kwargs) -> SSAValue:
+    raise NotImplementedError("Still to be implemented.")
+
+@_lower_expr.register(pyop3.expr.Conditional)
+def _(cond, /, iname_maps, loop_indices, *, context, **kwargs) -> SSAValue:
+    raise NotImplementedError("Still to be implemented.")
+
```
connorjward, line 681:
```suggestion
```

pyop3/compile/mlir.py
```diff
@@ -0,0 +1,715 @@
...
+    rhs = _coerce(context, rhs, _expr_dtype(e.b, node_dtype), node_dtype)
+    
+    # TODO: Improve this hotfix. Type resolution needs to consider indices
+    lhs_is_index = isinstance(lhs.type, IndexType)
+    rhs_is_index = isinstance(rhs.type, IndexType)
+    if lhs_is_index != rhs_is_index:
+        lhs = context._to_index(lhs)
+        rhs = context._to_index(rhs)
+
+    match kind:
```
connorjward, line 620: I'd turn this into another singledispatch function. We are redundantly turning `pyop3.expr.Add` into `"add"` just to `case` it. Just `.register(pyop3.expr.Add)` instead.

pyop3/compile/mlir.py
```diff
@@ -0,0 +1,715 @@
...
+        ops = list(self._entry_block.ops)
+        for op in ops:
+            op.detach()
+        new_block.add_ops(ops)
+        new_block.add_op(func.ReturnOp())
+
+        module = ModuleOp([func_op])
+        # NOTE: Raises errors if there are issues in the generated MLIR
+        # Quite expensive as it walks MLIR AST so maybe debug only in future?
+        module.verify()
```
connorjward, line 558:
```suggestion
        if pyop3.config.debug_checks:
            module.verify() 
```

pyop3/compile/mlir.py
```diff
@@ -0,0 +1,715 @@
...
+from pyop3.dtypes import IntType, RealType, ScalarType
+
+# NOTE: Can I just set an IndexType object here and reuse this? 
+# Bad design from xDSL. Who makes a type an object?   
+iType = IndexType()
+
+NUMPY_TO_XDSL = {
+    np.dtype("int64"): IntegerType, # NOTE: Temp forcing int64 to int32
+    IntType: IntegerType,
+    RealType: Float64Type,
```
connorjward, line 49: To explain things a little, `IntType`, `RealType` and `ScalarType` all come from PETSc. They can change depending on how PETSc is configured. E.g. `IntType` is usually int32 but can also be int64.

pyop3/compile/mlir.py
```diff
@@ -0,0 +1,715 @@
...
+        self._domains: Dict[str, Tuple[SSAValueT, SSAValueT]] = dict()
+
+        # NameVar temporaries already resolved to SSA values live in
+        # symbol_table; this records which names are temporaries.
+        self._temporaries: set = set()
+
+    def insert(self, op: Operation) -> Operation:
+        """ Inserts an MLIR operation into the block """
+        self._builder.insert(op)
+        return op
```
connorjward, line 174: Seems like a redundant return as it is just the original argument

pyop3/compile/mlir.py
```diff
@@ -0,0 +1,715 @@
...
+
+    def __repr__(self):
+        return f"<{self.name}, dtype: {self.dtype}, shape: {self.shape or '?'}>"
+
+
+class SymbolTable:
+    """
+    Class deals with SSAValues for a respective scope.
+
+    e.g: SSAValues inside scf.ForOp are not valid outwith the iterative loop.
```
connorjward, line 114:
```suggestion
    e.g: SSAValues inside scf.ForOp are not valid outside the iterative loop.
```

pyop3/compile/mlir.py
```diff
@@ -0,0 +1,715 @@
...
+                self.insert(for_op)
+                for_ops.append(for_op)
+
+                # Descend into the loop body.
+                body = for_op.body.block
+                induction = body.args[0]
+                self.symbol_table.push()
+                self.symbol_table.define(iname, induction)
+                self.insertion_stack.append(self._builder.insertion_point)
+                self._builder = Builder(InsertPoint.at_end(body))
```
connorjward, line 502: Can you explain this scoping business? What's an insertion point?

I feel it would be more elegant if `self._builder` was a stack of some sort. You can do scoping nicely using context managers.

pyop3/compile/mlir.py
```diff
@@ -0,0 +1,715 @@
...
+        return offset_ssa
+
+    @_get_offset.register(pyop3.expr.NonlinearDatBufferExpression)
+    def _(self, expr: pyop3.expr.NonlinearDatBufferExpression, /, iname_maps, loop_indices, *, paths):
+        path = utils.just_one(paths)
+        buffer = expr.buffer_view.buffer
+        offset_ssa = self._offset_generation(buffer, [expr.layouts[path]], iname_maps, loop_indices)
+        return offset_ssa
+
+    def add_buffer(
```
connorjward, line 337: This is 99% code reuse, could this be refactored and go in the parent class?

pyop3/compile/mlir.py
```diff
@@ -0,0 +1,715 @@
...
+
+    def add_assignment(self, assignee, expression, inames, loop_indices, prefix: str = "insn") -> None:
+        """
+        Lower the rhs and bind it in symbol table to assignee name 
+        Function assumes that assignee is a temporary variable, probably a mistake 
+        """
+        assert isinstance(assignee, NameVar)
+
+        # TODO: Pass down name_hint for debugging 
+        rhs = self.lower_expr(
```
connorjward, line 250: In the loopy code we don't do this because `expression` has already been lowered. Why is it needed here?

pyop3/compile/mlir.py
```diff
@@ -0,0 +1,715 @@
...
+
+        buffer = buffer_view.buffer
+        if isinstance(buffer, PetscMatBuffer):
+            raise NotImplementedError("PETSc buffers not implemented") 
+
+        offset_ssa = self._offset_generation(buffer, layouts, iname_maps, loop_indices)
+
+        # TODO: Not implemented
+        if self.propagate_negatives and intent == READ:
+            pass
```
connorjward, line 414:
```suggestion
            raise NotImplementedError
```
?

pyop3/compile/mlir.py
```diff
@@ -0,0 +1,715 @@
...
+        raise TypeError(f"No handler defined for {type(obj).__name__}")
+ 
+    @register_extent.register(numbers.Integral)
+    def _(self, num: numbers.Integral, *args, **kwargs):
+        """ Registers constant extent as SSA and returns num as key """ 
+        ssa = self.insert(
+                arith.ConstantOp(
+                    IntegerAttr(num, iType)
+                )
+        ).result
```
connorjward, line 523: You have this `.result` all over the place. Should the return value of `insert` be modified so this isn't needed?

pyop3/compile/mlir.py
```diff
@@ -0,0 +1,715 @@
...
+    @register_extent.register(pyop3.expr.Expression)
+    def _(self, expr: pyop3.expr.Expression, inames, loop_indices):
+        extent_name = self.add_temporary("p")
+        self.add_assignment(extent_name, expr, inames, loop_indices)
+        return extent_name
+
+    def finalize_kernel(self, function_name, compiler_parameters) -> ModuleOp:
+        n = len(self._arguments)
+        perm = sorted(range(n), key=lambda i: self._arguments[i].name)
+        arg_types = [self._buffer_type(self._arguments[i]) for i in perm]
```
connorjward, line 536:
```suggestion
        arg_types = [self._buffer_type(a) for a in self.arguments]
```
I think this `perm` business isn't needed

pyop3/compile/mlir.py
```diff
@@ -0,0 +1,715 @@
...
+
+        func_op = func.FuncOp(function_name, (arg_types, []))
+        func_op.attributes["llvm.emit_c_interface"] = UnitAttr()
+        new_block = func_op.body.block
+
+        # Reordering arguments
+        for new_index, old_index in enumerate(perm):
+            old_arg = self._entry_block.args[old_index]
+            new_arg = new_block.args[new_index]
+            old_arg.replace_by(new_arg)
```
connorjward, line 546: I don't really understand this. Looks like it's just doing
```py
self._entry_block.args = new_block.args
```
but in a strange way

connorjward (changes requested):
Super cool!

> There was a dtype property introduces into PyOP3 expressions but MLIR also uses an index type, an int required for load/storing memref/tensors.

We do sort of know when we have indices vs integers. E.g. consider
```
idat[2*i+j]
```
We know that `idat` is the outermost thing and hence the dtype is an integer, but when we recurse to look at `2*i+j` we now know that only index types make any sense.

Other than this I think my main points are:
1. The current scoping implementation is quite confusing. I much prefer context managers over `try`, `finally`.
2. MLIR is confusing. Code like
```py
        # Move old block ops into the func entry block
        ops = list(self._entry_block.ops)
        for op in ops:
            op.detach()
        new_block.add_ops(ops)
        new_block.add_op(func.ReturnOp())
```
could really benefit from more comments as the API is utterly foreign.

pyop3/compile/mlir.py
connorjward: We hit a `raise NotImplementedError` even without this

pyop3/compile/mlir.py
SamSJackson: Scottish language leaking into my comments

pyop3/compile/mlir.py
SamSJackson: I think I will add lots of comments around this section.

This loop is re-ordering the function arguments (the SSA values). 

When an argument is inserted into a block or operation, downstream SSA values are influenced. 
The `replace_by` is addressing any downstream influences between the pair of SSA values (old_arg, new_arg).
i.e. we cannot simply re-order the function signature but also replace downstream mentions of old_arg.

pyop3/compile/mlir.py
SamSJackson: Yes. I think I considered this and forgot to do it. Should be doable.

pyop3/compile/mlir.py
SamSJackson: Yes, done this.

pyop3/compile/mlir.py
SamSJackson: I considered removing but the permutation indices are used for argument re-ordering in the following for-loop.

pyop3/compile/mlir.py
SamSJackson: Poor coding, not needed here. Extracted out. I think `add_assignment` is not actually needed in this class.

pyop3/compile/mlir.py
SamSJackson: Resolved as per suggestion below for returning op.result.

pyop3/compile/mlir.py
SamSJackson: IntegerType requires an additional argument for bit-width. So, to work with the adaptable nature, I read the bit-width from np.dtype given as `IntType`  and apply that to IntegerType

pyop3/compile/mlir.py
SamSJackson: This has been done.

pyop3/compile/mlir.py
SamSJackson: The symbol table is a dictionary of SSA values. Scoping is done as SSA values will not be valid outside the loop - such as iter args. We introduce the scope so that these SSA values can be dropped upon exit. 

An insertion point is telling the Builder class at which point in the block to add MLIR (i.e. start or end of block).

I will try to design a more elegant solution because it is verbose rn

SamSJackson: Okay I cannot think of anything at the moment. 

I rewrote this to use a builder stack, as opposed to builder and insertion stack. Struggling to see a means to introduce context managers without a bigger rewrite. 

The eager approach of creating the loops and entering the inner-most loop corresponding to the inames introduced does not lend itself to a context manager. May suggest a deeper issue.

Leaving it as current change for now, until review.

connorjward: What if we just got rid of the `try` and `finally` in this function? We already have a `yield` which does the same sort of thing?

pyop3/compile/mlir.py
SamSJackson: I kept the existing dispatch function and changed this method to align the binary operations.

pyop3/compile/mlir.py
SamSJackson: Yes they are now dropped. Mild embarrassment that I missed that.

connorjward: No worries. To be clear that was basically all I was concerned about in this function. Now it is a context manager as requested, just a particularly complex one (which is fine).

SamSJackson: Misread that. Thank you for review.

PR #5433, opened 2026-09-06
Decide needs_update collectively

pbrubeck:
Stacked on #5057.

`RobustTransferManager.needs_update` compares `dat_version`, which counts
writes on the calling rank. Each rank was answering the question about its own
partition, while the answer drives `update()` and so the Jacobian reassembly
that every rank has to enter. After this change the answer is reduced over the
mesh communicator, so the ranks cannot disagree about whether to rebuild.

To be clear about what this is worth: it is a latent hazard, not an observed
failure. I could not construct a case where the versions actually diverge, so
this buys robustness rather than fixing a reproduced bug. It costs one boolean
allreduce per transfer, which is small beside the halo exchanges and solves a
transfer already performs, but it is a hot path and the trade is worth a second
opinion. It is a single two-line commit, so it is easy to drop.

`tests/firedrake/multigrid/test_robust_transfer.py` passes at 1 and 3 ranks.

Assisted by Claude Code (Claude Opus 5).

### DIFF
--- a/firedrake/mg/robust_transfer_manager.py
+++ b/firedrake/mg/robust_transfer_manager.py
@@ -1,4 +1,5 @@
 from functools import partial
+from mpi4py import MPI
 from ufl import H1, replace
 from finat.ufl import FiniteElement, NodalEnrichedElement, TensorElement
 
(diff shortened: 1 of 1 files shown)


pbrubeck:
Closing: this was opened against `pbrubeck/robust-multigrid`, the branch of the still-open #5057, which is not a valid base. The change depends on `needs_update`, which only exists on that branch, so it cannot target `main` yet.

The branch `pbrubeck/rtm-needs-update-collective` stays as it is. It should be reopened against `main` once #5057 has merged.

PR #5434, opened 2026-09-07
Exclude halos from `VertexOnlyMesh`

leo-collins:
Build a VertexOnlyMesh by considering owned parent mesh cells only.

- The mesh R-tree is now built from owned cell AABBs only.
- The final VertexOnlyMesh doesn't contain halo points anymore.
- The local cell location kernel is simplified since we don't care about which rank owns a given cell anymore.
- `VertexOnlyMesh.input_ordering_without_halos_sf` is deprecated.
- Each root in the  `input_ordering_sf` now maps to at most one leaf.

### DIFF
--- a/firedrake/locate.c
+++ b/firedrake/locate.c
@@ -11,9 +11,8 @@ PetscErrorCode locate_cell_from_candidates(struct Function *f,
         const int64_t *ids,
         size_t ncells_ignore,
         const PetscInt *cells_ignore,
-        const PetscInt *cell_owner_ranks,
-        PetscInt *cell_out,
-        PetscInt *owner_out)
+        PetscInt cell_limit,
+        PetscInt *cell_out)
 {
     bool cell_ignore_found = false;
     /* NOTE: temp_ref_coords and found_ref_coords are actually of type
@@ -24,19 +23,12 @@ PetscErrorCode locate_cell_from_candidates(struct Function *f,
     ref_coords_copy function and trust that the underlying memory which the
     pointers refer to is updated as necessary. */
     PetscReal best_distance = PETSC_MAX_REAL;
-    PetscInt best_owner = -1;
     PetscInt best_cell = -1;
     /* NOTE: `tolerance`, which is used throughout this function, is a static
        variable defined outside this function when putting together all the C
        code that needs to be compiled - see pointquery_utils.py */
 
-    /* NOTE: `cell_owner_ranks` and `owner_out` may be NULL. In this case, every
-       candidate is treated as having owner rank 0. This is for backwards compatibility
-       with the `Function.at` code path which is being deprecated soon. */
     *cell_out = -1;
-    if (owner_out) {
-        *owner_out = -1;
-    }
     for (size_t i = 0; i < nids; ++i) {
         /* Check that casting the ids from int64 to PetscInt is safe (for 32 bit petsc builds). 
         Since the ids are mesh cell ids this *should* always be safe, but better to check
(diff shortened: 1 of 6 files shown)


tests/firedrake/vertexonly/test_vertex_only_mesh_generation.py
```diff
@@ -285,16 +285,21 @@ def test_generate_cell_midpoints(parentmesh, redundant):
             _ = len(vm.coordinates.dat.data_ro)
 
     # Have correct pyop2 labels as implied by cell set sizes
+    # `cell_set.sizes` returns `(core, owned, total)`. The VOM does not include halos,
+    # so we only check that the core and owned counts match the parent mesh.
```
connorjward, line 289: In pyop3 I kill the core/owned distinction. Fine here for now though.

tests/firedrake/vertexonly/test_swarm.py
```diff
@@ -418,3 +343,32 @@ def test_discover_remote_roots():
...
+        (DistributedMeshOverlapType.FACET, 1),
+    ],
+)
+def test_point_on_partition_boundary(overlap):
+    comm = MPI.COMM_WORLD
+    parent_mesh = UnitIntervalMesh(
+        2,
+        distribution_parameters={"overlap_type": overlap},
+    )
+    coords = np.array([[0.5]]) if comm.rank == 0 else np.empty((0, 1))
```
connorjward, line 362: Could we manufacture a test like
```
p0
(owned)     (ghost)
x-----x | -----x
          o

p1
(ghost)     (owned)
x-----x | -----x
      o
```
To force each rank to think the other rank owns the point.

tests/firedrake/vertexonly/test_vertex_only_mesh_generation.py
connorjward: This is an excellent example of why the distinction is a bad idea. If you are looping over the points of the VoM then everything should be core, with nothing owned.

connorjward (commented):
Looks very promising

tests/firedrake/vertexonly/test_swarm.py
leo-collins: Do you mean having two points, each one slightly inside the ghost cell on each rank?

connorjward: No, I mean
```py
if comm.rank == 0:
  loc = 0.5 + 1e-10
else:
  loc = 0.5 - 1e-10

vom = VertexOnlyMesh([[loc]], redundant=False)
```
But since that delegates to rank 0 I don't think it would work. Ignore me.

firedrake/locate.c
```diff
@@ -57,32 +48,29 @@ PetscErrorCode locate_cell_from_candidates(struct Function *f,
...
-        /* Select owning cell by minimum L1 distance, with ties broken by highest owning rank. */
-        if (distance < best_distance || (distance == best_distance && owner > best_owner)) {
+        /* Select a cell by minimum L1 distance. */
+        if (distance < best_distance) {
             best_distance = distance;
-            best_owner = owner;
             best_cell = candidate;
             memcpy(found_ref_coords, temp_ref_coords, sizeof(struct ReferenceCoords));
+            /* Reference-cell distance is nonnegative, so this is optimal. */
+            if (distance == 0.0) {
```
connorjward, line 68: leave a comment explaining why this is OK

connorjward (changes requested):
otherwise good to go

PR #5437, opened 2026-09-08
Drop the MatIS special case in FunctionSpace.local_to_global_map

pbrubeck:
## Description

`FunctionSpace.local_to_global_map()` carried a `mat_type` argument for one reason: a local submatrix of a `MATIS` reported its local-to-global map in a different layout from one of a `MATLOCALREF`, so masking the boundary condition nodes had to read a different array for each.

```python
if mat_type == "is":
    indices = lgmap.indices.copy()
    unblocked = False
else:
    # LGMap is already unrolled
    indices = lgmap.block_indices.copy()
    unblocked = True
```

PETSc [!9575](https://gitlab.com/petsc/petsc/-/merge_requests/9575) puts `MATIS` on the same layout as `MATLOCALREF`, so the scalar indices are in `block_indices` for both and the branch has nothing left to distinguish. This PR drops it, along with the `mat_type` argument threaded through `ParloopBuilder` to supply it.

The `MATIS` arm was never correct anyway: it read `lgmap.indices`, which is only the scalar map when the subspace starts on a block boundary of the mixed space. It also never reached the assembled matrix, because `ParloopBuilder.replace_lgmaps()` skips installing the masked map when `m.handle.type == "is"`. What it did do was keep `indices[nodes] = -1` from going out of bounds, which is why it cannot simply be deleted against the currently pinned PETSc.

## Testing

`test_assemble_matis` gains a mixed space whose blocked subspace comes second, so that it starts at a degree of freedom of the mixed space that is not a multiple of its own block size — the case PETSc got wrong. The existing parametrizations put the blocked subspace first, where the offset is zero and the bug does not show.

Against a PETSc that carries !9575, all 10 parametrizations pass on 3 processes. Against the pinned PETSc, 5 of them fail: `mixed-is` with boundary conditions raises `IndexError` from `indices[nodes] = -1`, and both `mixed_blocks-is` cases raise PETSc error 75 from `MatGetLocalSubMatrix()`.

I also checked, outside the test suite, that every combination of scalar, vector and tensor subspaces in a two-field mixed space assembles identically under `mat_type="is"` and `mat_type="aij"`, with and without Dirichlet conditions, in serial and on 3 processes. Before !9575, 10 of those 25 combinations either raised or assembled the wrong matrix.

`make srclint` is clean. It stops at `cython-lint`, which is not installed in my environment; no `.pyx` file is touched.

## Blocked on PETSc

CI cannot go green until PETSc !9575 reaches a release that `SUPPORTED_PETSC_VERSION` (currently `v3.25.0`) points at. !9575 is stacked on [!9511](https://gitlab.com/petsc/petsc/-/merge_requests/9511), and both need to land in `release` first.

To build against it before then, the `petsc/petsc` branch `pbrubeck/matis-localref-block-sizes` is current `release` plus the whole stack.

Opened as a draft for that reason.

---

Assisted by Claude Code (Claude Opus 5).


firedrake/functionspaceimpl.py
```diff
@@ -900,16 +896,11 @@ def local_to_global_map(self, bcs, lgmap=None, mat_type=None):
...
             assert bsize == self.block_size
-            if mat_type == "is":
-                indices = lgmap.indices.copy()
-                unblocked = False
-            else:
-                # LGMap is already unrolled
-                indices = lgmap.block_indices.copy()
-                unblocked = True
+            indices = lgmap.block_indices.copy()
+            unblocked = True
```
stefanozampini, line 903: I find this confusing: you have blocked indices (one per block), but the flag is `unblocked`

stefanozampini (approved):

firedrake/functionspaceimpl.py
pbrubeck: I also find this confusing

